PR #167, opened 2021-11-02
service/share: Service interface and basic implementation for GetShare

Wondertan:
## Content
* Fully defined interface for `share.Service`
	* (in future we may add more write operations ad-hoc) 
* `GetShare` implementation.
	* `GetShares` are not needed right now and will be implemented together with BlockSync logic
	* `GetSharesByNamespaces` will be implemented by #5
* Reusable testing utilities
* Alias for DAH in `service/header`
* Multiple TODOs for possible data usage optimizations

## Refs
Closes #23
Depends on #158

PR #171, opened 2021-11-05
das: DASer interface and implementation

### LINKED ISSUE #166: das: DASer
## Goal
We need a component that will do data availability sampling automatically for each new header.

## Design
The DASer relies on two services in our codebase: shares and headers. Header service is needed to subscribe for new Headers on the network using `Subscriber` interface, while Share service is needed to request random samples via GetShare() method.

The basic DASer structure should look like:
```go
type DASer struct {
	headers headers.Service
	shares  shares.Service
}
```

It should also contain `Start` and `Stop` methods to control its flow. When it starts it should initiate header subscription and header-by-header execute DASing. On `Stop` the process should be stopped.

Wondertan:
## Refs
Based on #167 
Closes #166


renaynay (changes requested):

das/daser.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// NewDASer creates a new DASer.
+func NewDASer(shares share.Service) DASer {
+	return &daser{
+		shares: shares,
+	}
+}
+
+type daser struct {
```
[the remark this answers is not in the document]
Wondertan: Outdated

das/daser.go
```diff
@@ -0,0 +1,87 @@
...
+	shares share.Service
+}
+
+func (das *daser) DAS(ctx context.Context, dah *header.DataAvailabilityHeader) error {
+	log.Debugw("DAS", "dah", dah.Hash())
```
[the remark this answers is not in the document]
Wondertan: Outdated

service/share/light_availability.go
```diff
@@ -0,0 +1,72 @@
...
+	"context"
+	"errors"
+
+	format "github.com/ipfs/go-ipld-format"
+
+	"github.com/celestiaorg/celestia-node/ipld"
+)
+
+// DefaultSamples sets the default amount of samples to be DASed by light Availability.
+var DefaultSamples = 16
```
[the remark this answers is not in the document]
Wondertan: Ok.

service/share/light_availability.go
```diff
@@ -0,0 +1,72 @@
...
+
+// lightAvailability implements Availability using Data Availability Sampling technic.
```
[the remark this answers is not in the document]
Wondertan: That should have been Technique

service/share/sample.go
```diff
@@ -0,0 +1,76 @@
...
+	return nil
+}
+
+func (ss *squareSampler) samples() []Sample {
+	samples := make([]Sample, 0, len(ss.smpls))
+	for s := range ss.smpls {
+		samples = append(samples, s)
+	}
+	return samples
+}
```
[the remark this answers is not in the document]
Wondertan: To convert map of samples to slice of samples

service/share/sample.go
```diff
@@ -0,0 +1,76 @@
...
+
+func newSquareSampler(squareWidth int, expectedSamples int) *squareSampler {
+	return &squareSampler{
+		squareWidth: squareWidth,
+		smpls:       make(map[Sample]struct{}, expectedSamples),
+	}
+}
+
+func (ss *squareSampler) sample(num int) error {
```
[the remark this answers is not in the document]
Wondertan: Changed to `generateSample` and added some docs. Requests are out of scope of sampler.

service/share/light_availability.go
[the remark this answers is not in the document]
Wondertan: fixed

service/share/light_availability.go
```diff
@@ -0,0 +1,72 @@
...
+type lightAvailability struct {
+	getter format.NodeGetter
+}
+
+// NewLightAvailability creates a new Light DataAvailability.
```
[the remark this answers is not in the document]
Wondertan: fixed in e4015ee

service/share/light_availability.go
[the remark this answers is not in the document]
Wondertan: fixed in e4015ee

service/share/sample.go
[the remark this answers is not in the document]
Wondertan: fixed in e4015ee

service/share/sample.go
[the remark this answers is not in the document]
Wondertan: fixed in e4015ee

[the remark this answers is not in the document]
renaynay: Ah I see bc it's a map in the `squareSampler`.

renaynay (approved):
👍🏻

ipld/write.go
```diff
@@ -13,13 +13,20 @@ import (
 	"github.com/celestiaorg/celestia-core/pkg/wrapper"
 )
 
+// BatchSize defines an amount of IPLD Nodes to be buffered and written at once.
+// This configuration is very database backend specific and the current default(128) is not optimized for the version of
+// Badger we use. We set it to one to avoid test flakiness, as some tests may read for data that was not written yet.
+// TODO(@Wondertan): Find out the perfect value for Badger(e.g. ask PL folks) or migrate to go-car IPLD
+//  storage(preferred).
+const BatchSize = 1
```
adlerjohn, line 21: Is there a reason to set this to `1` when the comment says `128`?

service/share/light_availability.go
```diff
@@ -0,0 +1,75 @@
...
+	return &lightAvailability{
+		getter: get,
+	}
+}
+
+// SharesAvailable randomly samples DefaultSamples amount of Shares committed to the given Root.
+// This way SharesAvailable subjectively verifies that Shares are available.
+func (la *lightAvailability) SharesAvailable(ctx context.Context, dah *Root) error {
+	log.Debugw("Validate availability", "root", dah.Hash())
+	samples, err := SampleSquare(len(dah.ColumnRoots), DefaultSampleAmount)
```
adlerjohn, line 34: Aside: can there be a common size like `dah.size`? It seems a bit hacky to use the length of column roots but not row roots. A reader might think "why?"

More to the point, can you make a TODO for the number of samples here?

service/share/sample.go
```diff
@@ -0,0 +1,77 @@
...
+		if _, ok := ss.smpls[s]; ok {
+			continue
+		}
+
+		done++
+		ss.smpls[s] = struct{}{}
+	}
+
+	return nil
+}
```
adlerjohn, line 60: If picking a large number of samples this could take a while. But a large number of samples would imply a small square size, so short runtime anyways. That being said, this could be a source of non-deterministic timeouts/performance. Would it be better to simply use a standard library function to generate `num` random numbers between `0` and `ss.squareWidth*ss.squareWidth`, then mutate them into 2D?

adlerjohn (changes requested):

ipld/write.go
Wondertan: 128 is set by default, so when we don't set anything it will be 128, but purposely change it to 1 and the purpose is described.

service/share/light_availability.go
Wondertan: I think this is the right point! I will change to RowRoots for consistency. Adding Size method to DAH is also a good option but that will require a PR to core where DAH is defined and another loop of reviews we don't have a time for.

service/share/sample.go
Wondertan: I assume this is not a final solution and the next plan I see here is to do https://github.com/celestiaorg/celestia-node/issues/182, also as mentioned at the top of the file.

Wondertan: Unfortunately, Golang std does not have such func in crypto safe randomization API https://pkg.go.dev/crypto/rand@go1.17.3.

Wondertan: Also, note this code was working fine before for MVP.

Wondertan: I understand your concern entirely and this __may__ loop forever, but this is not realistic for 16 samples we have as a default rn.

Wondertan: @adlerjohn, can we keep this code as it is right now as it proved that it is functional and change it later to random tree walks as in #182?

service/share/light_availability.go
adlerjohn: Can you just make an issue so we don't forget eventually?

service/share/sample.go
adlerjohn: Absolutely we can keep it as-is for now, so long as there's a plan (which it looks like there is).

adlerjohn (dismissed):

service/share/light_availability.go
Wondertan: @adlerjohn, https://github.com/celestiaorg/celestia-core/issues/575

Wondertan:
Extracted flaky test fixing commits into #186

adlerjohn (approved):
lgtm

node/services/service.go
```diff
@@ -19,8 +24,9 @@ func Block(lc fx.Lifecycle, fetcher block.Fetcher, store ipld.DAGService) *block
 }
 
 // Share constructs new share.Service.
-func Share(lc fx.Lifecycle, dag ipld.DAGService) share.Service {
-	service := share.NewService(dag)
+func Share(ctx context.Context, lc fx.Lifecycle, dag ipld.DAGService) share.Service {
+	avail := share.NewLightAvailability(merkledag.NewSession(fxutil.WithLifecycle(ctx, lc), dag))
```
liamsi, line 28: Wondering if we should not directly use the `fxutil.WithLifecycle(ctx, lc)` as a param instead?

renaynay (approved):
lfg

liamsi (commented):
Joining too late to the party. Saw this was already merged 💪🏼

Issue #130, opened 2021-10-11
Further clarify relation of abci-app in the embedded case


liamsi:
### Summary

ADR-001 described the relation between node and core+abci-app. Since then, there was a slight pivot to embed the tendermint process in node directly (ADR-002). ADR 002 does not yet clearly describe how the abci-app fits into the equation.

### Gain

No doubt on the relationship between node and abci-app and which components "drives" which. Also, the implications for the user interface (on the command-line) if any should be clarified. 

### Evidence

https://github.com/celestiaorg/celestia-node/pull/97#discussion_r726020336

### Implementation ideas

Previously the idea was that the app binary would be used to init core and the app. Now that node embeds the core node, it seems to be necessary that the node exposes the same cli as the app. This needs to be done in a way that does not require writing a lot of code (ideally, the corresponding commands are just "passed through" to the cli of node, similar to how the app, exposes the [tendermint node's cli](https://github.com/cosmos/cosmos-sdk/blob/7e6151e93c4d48da813aba37d475df2393ca087b/server/tm_cmds.go#L9) already).

### Urgency

High

liamsi:
Didn't we remove the embedded case for now? can we close this then?

renaynay:
Yes @liamsi

renaynay:
However, we will be revisiting the concept of embedding celestia-app once we think about how full nodes can serve state.

PR #175, opened 2021-11-09
Update to the latest version of celestia-core

### LINKED ISSUE #174: Update to the latest stable version of celestia-core
The current celestia-node is using an old version of celestia-core. We will eventually need to update, and doing so sooner rather than later will hopefully ensure that we do not build on old code. It is also necessary if we wish to use the latest version of the celestia-app #130

evan-forbes:
This PR updates to the latest version of celestia-core. I purposefully limited the changes introduced to be strictly related to upgrading, and did not include any changes related to #130

blocking #130
closes #174 

Note, this ivolved swithcing to a version of celestia-core that declares itself as tendermint [#528](https://github.com/celestiaorg/celestia-core/issues/528)


testutils/blockdata.go
```diff
@@ -0,0 +1,158 @@
+package testutils
```
evan-forbes, line 1: moved this here, as this repo is the only place that it is used. Happy to move it to celestia-core if we think that's a better plan.

ipld/read.go
```diff
@@ -11,9 +11,9 @@ import (
 
 	"github.com/celestiaorg/rsmt2d"
 
-	"github.com/celestiaorg/celestia-core/pkg/da"
-	"github.com/celestiaorg/celestia-core/pkg/wrapper"
 	"github.com/celestiaorg/celestia-node/ipld/plugin"
+	"github.com/tendermint/tendermint/pkg/da"
+	"github.com/tendermint/tendermint/pkg/wrapper"
```
Wondertan, line 16: Hmm. Now I have a concern about importing Tendermint directly because it does not have those packages and likely will never do. 

I feel like this is confusing. It kinda lies that Tendermint has those packages but it is not. For any newcomer who decides to look at the sources, this may end up as a road to nowhere.

Unfortunately, Tendermint is not a black box to do so and now I don't like this idea :(

testutils/blockdata.go
Wondertan: Can you pls point to the place we use those? Right now, in Share Service, we define our own analogs of such helpers.  If it is used in IPLD and absolutely necessary pls move those into `ipld` package, otherwise we don't need them.

Wondertan (changes requested):

testutils/blockdata.go
Wondertan: Oh, ok this is also used in `service/block`. Can you pls put each used testing helper in respective package where it is used? E.g `GenerateRandomBlockData` to be put into `testing.go` in `service/block`.

ipld/read.go
renaynay: Yeah I agree this is kind of confusing :/

evan-forbes: I agree that it is confusing, and as discussed a while ago somewhere we should eventually move those packages to an external repo to avoid this exact thing. Until then, thanks to go modules we have no other option unless we wish change how celestia-core declare's it package (and therefore re-change all the import everywhere and make pulling changes from upstream feel like pulling teeth).

renaynay: Okey :(

Wondertan: Okey ;(

evan-forbes:
forgot to point out that I moved the testutil code to the proper repo 0168e02e73ba6e8ce130c264b3a5af2b9d7ca037

and then fixed the linter issues as well https://github.com/celestiaorg/celestia-node/pull/175/commits/9894a1e6550cbd811682da4df72886de26ad26f5

adlerjohn (approved):

adlerjohn (changes requested):
Ah nvm there's a conflict with default branch, can you resolve @evan-forbes ?

evan-forbes:
> Ah nvm there's a conflict with default branch, can you resolve
:heavy_check_mark:

adlerjohn (dismissed):

liamsi (dismissed):
THANKS 🚀

liamsi:
I messed up a few things after merging in main. Should be resolved now. But I've copied the test helper `generateRandNamespacedRawData` to several places.

ipld/plugin/nmt_test.go
```diff
@@ -41,3 +44,27 @@ func TestNamespaceFromCID(t *testing.T) {
 		})
 	}
 }
+
+// generateRandNamespacedRawData returns random namespaced raw data for testing purposes.
+func generateRandNamespacedRawData(total, nidSize, leafSize uint32) [][]byte {
```
liamsi, line 49: This helper exists now 4 times in tests. I don't really mind (as it's only in tests but maybe this should actually again be moved into a testutil package or export it only once; in a separate PR though IMHO).

liamsi (dismissed):
THANKS 🚀

Wondertan (commented):
@evan-forbes, We should change the reference form celestial-core to tendermint in Makefile, otherwise LGTM

evan-forbes:
good catch. I didn't look into exactly how proto generation is handled here, is this correct?
https://github.com/celestiaorg/celestia-node/pull/175/commits/b476f6b80d19b8b5df1157ddba5fa36448a16ad2

liamsi:
Good catch indeed! I'm not sure if that go list command uses the replace directive or not. If it does, this should be correct.

Wondertan:
> good catch. I didn't look into exactly how proto generation is handled here, is this correct?

Yeah, thanks! Did you try rerunning the proto command to see if it works?

evan-forbes:
I did rerun it, but nothing changed. Would someone mind double checking for me to make sure? The generated file is importing using tendermint/tendermint instead of celestaiorg/celestia-core, so I expect it to be working.

liamsi:
I'll check out the branch locally and try it.

liamsi (approved):

liamsi:
Works as expected.

Wondertan (approved):
If it runs, then it works, no changes should be made

Wondertan:
@renaynay, after merging this PR we will need to rebase our PRs and change import manually there in all the new files we write.

renaynay (approved):
Let's merge.

PR #176, opened 2021-11-09
Use celestia-app when starting up devnode

evan-forbes:
This PR adds code to use celestia-app instead of kvstore when compiling a dev node. 

While this works for the most part, I'm curous for which scenarios this will actually get used. It makes testing more difficult, as the app relies on setting up a configured network or always fails.

Notably, it doesn't start the app's grpc server yet.


Wondertan:
> While this works for the most part, I'm curous for which scenarios this will actually get used. It makes testing more difficult, as the app relies on setting up a configured network or always fails.

By  "this" and "It" do you mean integration implemented in this PR?

evan-forbes:
> By "this" and "It" do you mean integration implemented in this PR?

yeah, will running an embedded full celestia-app only be used while developing/testing, or do we intend on people being able to use a single binary to run both?

Wondertan:
> do we intend on people being able to use a single binary to run both in a real network?

Yeah, I am aiming for this. What concerns do you have about this?

evan-forbes:
> Yeah, I am aiming for this. What concerns do you have about this?

I don't have any concerns, just clarifying :slightly_smiling_face:

evan-forbes:
While this PR does add the ability to embed the celestia-app instead of the kvstore, this functionality is not yet working. This is because in app is built to only function in a viable network. To achieve that, we will essentially have to a similar script to [this one](https://github.com/celestiaorg/lazyledger-app/blob/master/contrib/single-node.sh) before we can actually start the node. #190

Per #189, we are also not yet running the celestia-app's grpc server when embedding it. We will have to do that before we can actually use the embedded app.

evan-forbes:
This is kinda blocked by #175 #189 and #190

we should hopefully have PR for #189 and #190 soon

liamsi (commented):
Needs to be rebased. Otherwise looks good so far. I think the prior is to get the actual mode running and not the devmode but AFAIU this already works via the remote client (instead of embedding)?

liamsi:
If the devmode is going to be remove, this can be closed I assume?

renaynay:
Yeah I think we can close this as it'll be too intense to rebase / refactor depending on our impl of node types / removal of dev mode. cc @Wondertan

Wondertan:
I am closing this for now but keeping it. This might be a good place to return back to

Issue #84, opened 2021-09-23
service: Implement processing of BadEncodingFraudProof and StateFraudProof


musalbas:
Implement processing of BadEncodingFraudProof and StateFraudProof to check if the fraud proof is valid or not.

If a fraud proof is valid, the node should halt (https://github.com/celestiaorg/celestia-node/issues/82). This is because, under Tendermint BFT:
* in order for a fraud proof to be valid it must be in a block that has consensus, which means it must be signed by 2/3 of validators;
* in order to "replace" that invalid block which has consensus, 2/3 of validators must equivocate;
* if 2/3 of validators equivocate, Tendermint halts and kicks into "manual recovery" mode (https://blog.cosmos.network/the-4-classes-of-faults-on-mainnet-bfabfbd2726c).

liamsi:
Full nodes will also halt, we need to think about who will actually be around in the network to notify light nodes
some copy pasta from slack:
> ismail: if everyone immediately stops the node whenever they see a fraud proof, aren’t odds hight that a) the fraud proofs does not gossip the fraud proof fast enough and b) especially for historic headers no one will be left to actually serve that fraud proof?
> hlib: Should sharing the fraud proof to your peers before stopping solve this?

adlerjohn:
By induction, only a single block is needed to be proven invalid, and that block is exactly the one that halts the full nodes. There is no need to produce fraud proofs for blocks that descend from that first invalid block, even if they themselves are invalid. The process of halting on an invalid block MUST also produce a fraud proof, and that fraud proof alone is sufficient for allowing light nodes to orphan all descendant blocks.

renaynay:
Closing in favour of https://github.com/celestiaorg/celestia-node/issues/528

PR #177, opened 2021-11-09
das: implement DASer

### LINKED ISSUE #166: das: DASer

Wondertan:
## Content
* Initial DASer implementation which just listens for headers and DASes them
* Adds DASer and Header Service to the node.

## Refs
Based on #171
Closes #166

### DIFF
--- a/das/daser.go
+++ b/das/daser.go
@@ -70,7 +70,7 @@ func (d *DASer) sampling(ctx context.Context, sub header.Subscription) {
 			}
 			return
 		}
-		
+
 		err = d.da.SharesAvailable(ctx, h.DAH)
 		if err != nil {
 			if err == context.Canceled {
--- a/das/daser.go
+++ b/das/daser.go
@@ -1,0 +1,83 @@
+package das
+
+import (
+	"context"
+	"fmt"
+
+	logging "github.com/ipfs/go-log/v2"
+
+	"github.com/celestiaorg/celestia-node/service/header"
+	"github.com/celestiaorg/celestia-node/service/share"
+)
+
+var log = logging.Logger("das")
+
+// TODO(@Wondertan): Initialization and warm-up.
+// TODO(@Wondertan): Start and Stop is better be thread-safe.
+// DASer continuously validates availability of data committed to headers.
+type DASer struct {
+	da   share.Availability
+	hsub header.Subscriber
+
+	cancel context.CancelFunc
+}
+
+// NewDASer creates a new DASer.
+func NewDASer(da share.Availability, hsub header.Subscriber) *DASer {
+	return &DASer{
+		da:   da,
+		hsub: hsub,
+	}
+}
+
+// Start initiates subscription for new Headers and spawns a sampling routine.
+func (d *DASer) Start() error {
+	if d.cancel != nil {
+		return fmt.Errorf("da: DASer already started")
+	}
+
+	sub, err := d.hsub.Subscribe()
+	if err != nil {
+		return err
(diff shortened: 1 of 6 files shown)


das/daser.go
```diff
@@ -0,0 +1,84 @@
...
+// sampling validates availability for each Header received from header subscription.
+func (d *DASer) sampling(ctx context.Context, sub header.Subscription) {
+	defer sub.Cancel()
+	for {
+		h, err := sub.NextHeader(ctx)
+		if err != nil {
+			if err == context.Canceled {
+				return
+			}
+			log.Errorw("daser failed to get next header", "err", err)
```
renaynay, line 71:
```suggestion
			log.Errorw("DASer failed to get next header", "err", err)
```

das/daser.go
```diff
@@ -0,0 +1,84 @@
...
+			return
+		}
+
+		err = d.da.SharesAvailable(ctx, h.DAH)
```
renaynay, line 75: Do we have an issue open somewhere for error handling when things like `SharesAvailable` fail?

node/services/service.go
```diff
@@ -27,11 +44,20 @@ func Block(lc fx.Lifecycle, fetcher block.Fetcher, store ipld.DAGService) *block
...
+func Share(lc fx.Lifecycle, dag ipld.DAGService, avail share.Availability) share.Service {
+	service := share.NewService(dag, avail)
 	lc.Append(fx.Hook{
 		OnStart: service.Start,
 		OnStop:  service.Stop,
 	})
 	return service
 }
+
+func DASer(avail share.Availability, service *header.Service) *das.DASer {
```
renaynay, line 56:
```suggestion
// DASer constructs a new Data Availability Sampler.
func DASer(avail share.Availability, service *header.Service) *das.DASer {
```

das/daser.go
Wondertan: While we are in dev/test nets environment I would just:
* Retry
* Persist failed retries with #180 

In a real environment, we should probably go with #84. As SharesAvailable should somehow work with BadEncodingFraudProofs at a later stage.

renaynay:
On hold while we add tests.

renaynay (approved):

das/daser.go
```diff
@@ -0,0 +1,81 @@
...
+			if err != context.Canceled {
+				log.Errorw("DASer failed to get next header", "err", err)
+			}
+			return
+		}
+
+		err = d.da.SharesAvailable(ctx, h.DAH)
+		if err != nil {
+			if err == context.Canceled {
+				return
```
mattdf, line 75: Can you make it so that both these if statements follow the same true path? i.e. either both `==` or both `!=`, it's confusing right now.

mattdf: @renaynay  ^

Wondertan: Made things consistent again

mattdf (approved):

renaynay (approved):

Issue #125, opened 2021-10-08
ipld: Change `PutData` to prevent duplicate erasure coding process


renaynay:
`PutData` expects original shares and duplicates the erasure coding process inside of `celestia-node` because it adds a NodeVisitor to the `NewErasuredNamespacedMerkleTree` so that all nodes an be added to the batch during the erasure coding process. We need to fix this so that erasure coding doesn't happen twice in the codebase -- either we have to modify the event loop in `BlockService`, or we modify the IPLD plugin so we can use it when initially erasure coding the block data.

[Ref](https://github.com/celestiaorg/celestia-node/pull/113#discussion_r724983149)

renaynay:
We should also keep in mind that computation is cheaper than storing in mem in Go, so maybe we can run some benchmarks later on to see how to best optimise this..

renaynay:
We also need to check to ensure we're not calling `ValidateBasic()` all over the place as `ValidateBasic()` erasure codes the block to check the DAH (so we're duplicating the process several times).

renaynay:
No longer relevant as `BlockService` doesn't exist anymore.

PR #186, opened 2021-11-09
fix flaky TestEventLoop and TestSubscriber

### LINKED ISSUE #154: tests: TestEventLoop is flaky
In #153 tests failed once from:
```
=== RUN   TestEventLoop
    event_test.go:38: 
        	Error Trace:	event_test.go:38
        	            	failure to retrieve data square: merkledag: not found
        	Test:       	TestEventLoop
```

### LINKED ISSUE #173: tests: `TestSubscriber` can hang / sometimes fails due to `context deadline exceeded`
## Proof
image

## Refs
Found in #171 

## Possible solution
There has to be a small 200 milliseconds timeout after the first Subscription on PubSub Topic. Basically, we need to ensure that peers notified each other about joining a topic before they are sending the msg. Otherwise, msg will never find a recipient.

Wondertan:
Closes #154 
Closes #173

### DIFF
--- a/service/block/event_test.go
+++ b/service/block/event_test.go
@@ -3,6 +3,7 @@ package block
 import (
 	"context"
 	"testing"
+	"time"
 
 	md "github.com/ipfs/go-merkledag/test"
 	"github.com/stretchr/testify/assert"
@@ -31,9 +32,9 @@ func TestEventLoop(t *testing.T) {
 	numBlocks := 3
 	expectedBlocks := mockFetcher.generateBlocks(t, numBlocks)
 
-	ctx, cancel = context.WithCancel(context.Background())
-	t.Cleanup(cancel)
-
+	// as we calculate extended square twice, this may take a lot of time, causing the main thread expect the result
+	// earlier than its ready.
+	time.Sleep(100 * time.Millisecond)
 	for i := 0; i < numBlocks; i++ {
 		block, err := serv.GetBlockData(ctx, expectedBlocks[i].Header().DAH)
 		require.NoError(t, err)
--- a/service/header/subscription_test.go
+++ b/service/header/subscription_test.go
@@ -3,6 +3,7 @@ package header
 import (
 	"context"
 	"testing"
+	"time"
 
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	mocknet "github.com/libp2p/go-libp2p/p2p/net/mock"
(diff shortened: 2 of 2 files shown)


service/block/event_test.go
```diff
@@ -31,9 +32,9 @@ func TestEventLoop(t *testing.T) {
 	numBlocks := 3
 	expectedBlocks := mockFetcher.generateBlocks(t, numBlocks)
 
-	ctx, cancel = context.WithCancel(context.Background())
-	t.Cleanup(cancel)
-
+	// as we calculate extended square twice, this may take a lot of time, causing the main thread expect the result
+	// earlier than its ready.
+	time.Sleep(100 * time.Millisecond)
```
adlerjohn, line 37: This seems a bit...brittle? Instead of sleeping, shouldn't there be some sort of async `await` or callback?

Wondertan: Well, we just shouldn't calculate extended headers twice ;) But that's a whole different story...
Adding callbacks and/or awaiting on chan to API just to make test happier is not a good idea, though I understand where you are coming from.

renaynay:
Looks like it's still broken though :( @Wondertan

Wondertan:
@renaynay, All the checks successful

renaynay (approved):
LGTM thank you

service/block/event_test.go
liamsi: Agree with John on this: I think this kind of timeouts are the reason we experience non-deterministic test-flakiness in tendermint. No need to change anything here though. And as you said the square will only be computed once in the future (if we have an issue for that I would recommend linking this comment or code there).

service/header/subscription_test.go
```diff
@@ -17,33 +18,36 @@ func TestSubscriber(t *testing.T) {
...
 		pubsub.WithMessageSignaturePolicy(pubsub.StrictNoSign))
 	require.NoError(t, err)
-	topic, err := gossub2.Join(ExtendedHeaderSubTopic)
+
+	// create header Service
+	headerServ2 := NewHeaderService(nil, nil, pubsub2)
+	err = headerServ2.Start(context.Background())
 	require.NoError(t, err)
 
+	time.Sleep(2 * time.Second)
```
liamsi, line 44: Similar as above: https://github.com/celestiaorg/celestia-node/pull/186/files#r746171777

liamsi (approved):

service/block/event_test.go
Wondertan: @renaynay, do we have an issue for this? I feel like you created one at some point?

renaynay: i think this https://github.com/celestiaorg/celestia-node/issues/125

service/header/subscription_test.go
renaynay: https://github.com/celestiaorg/celestia-node/issues/125

Wondertan: Note, this is not fixable due to the way pubsub(gossipbsub) works. Pubsub's test themselves are using 2 second timeout. Only in the recent release a new feature appeared to wait here, but it is also brittle. For ref https://github.com/libp2p/go-libp2p-pubsub/pull/452.

renaynay (approved):

adlerjohn (approved):

Issue #172, opened 2021-11-05
`ExtendedHeaderExchange` -- ability for nodes to request extended headers by height / latest height from their peers via libp2p streams


renaynay:
TODO: 
- [ ] Ability to ask for header at a given height `RequestHeader(ctx, height)` and `RequestHeaders(ctx, to, from)`
- [ ] Implement header sync

PR #188, opened 2021-11-10
service/header: Implement `HeaderExchange`

renaynay:
Related to https://github.com/celestiaorg/celestia-node/issues/172.


service/header/exchange.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	host.SetStreamHandler(headerExchangeProtocolID, ex.requestHandler)
+	return ex
+}
+
+// requestHandler handles inbound ExtendedHeaderRequests.
+func (e *exchange) requestHandler(stream network.Stream) {
+	// unmarshal request
+	pbreq := new(header_pb.ExtendedHeaderRequest)
+	_, err := serde.Read(stream, pbreq)
```
liamsi, line 42: I'm surprised to see this used here. Looking at the git commit history: https://github.com/celestiaorg/celestia-node/pull/188/commits/7d6ef08c9344175fbe9be892f9765aad614683e5
it seems like this is only used to length-prefix the messages when writing and using that info when reading? Is that correct? I'm honestly not sure this justifies yet another dependency. I'd prefer to copy the necessary minimal logic into this repo instead. Core/tendermint used to provide a  helper for that (and it already is a dependency) but it got internalized: https://github.com/celestiaorg/celestia-core/tree/master/internal/libs/protoio

But also [another already used dependency](https://github.com/celestiaorg/celestia-node/blob/c2d6a3537259d910bee337017ab90d6d5e82662b/go.mod#L12) provides that same functionality: https://github.com/gogo/protobuf/blob/226206f39bd7276e88ec684ea0028c18ec2c91ae/io/varint.go#L44 and https://github.com/gogo/protobuf/blob/226206f39bd7276e88ec684ea0028c18ec2c91ae/io/varint.go#L94
This exists in gogo/proto as it is common to length prefix to read and write from streams in protobuf: https://developers.google.com/protocol-buffers/docs/techniques#streaming

Also, we know the max size of the extended header (and it's not really large).

liamsi (commented):
First pass

service/header/exchange.go
Wondertan: None of the mentioned utilities provide buffer pooling, but up to ofc...

liamsi: Yeah, I honetsly forgot about the buffer pooling. Maybe you are right and that justifies the additional dependency. Not sure. I would have literally copied that helper code into this repo and made it so that the max message is exactly the max extended header size instead. on another note, do you think it makes sense to let the user choose the max in the serde library instead?

service/header/exchange.go
```diff
@@ -0,0 +1,194 @@
+package header
+
+import (
+	"context"
+	"fmt"
+
+	header_pb "github.com/celestiaorg/celestia-node/service/header/pb"
```
liamsi, line 7: nit: aliases with an underscore seem somewhat uncommon in golang.

service/header/header.go
```diff
@@ -32,10 +33,10 @@ func (eh *ExtendedHeader) MarshalBinary() ([]byte, error) {
 	return MarshalExtendedHeader(eh)
 }
 
-// MarshalBinary unmarshals ExtendedHeader from binary.
+// UnmarshalBinary unmarshals ExtendedHeader from binary.
 func (eh *ExtendedHeader) UnmarshalBinary(data []byte) error {
 	if eh == nil {
-		return fmt.Errorf("header: cannot UnmarshalBinary - nil ExtendedHeader")
+		return fmt.Errorf("expected: cannot UnmarshalBinary - nil ExtendedHeader")
```
liamsi, line 39: The "expected:" seems a bit confusing here. Why not simply:
```suggestion
		return fmt.Errorf("cannot UnmarshalBinary - nil ExtendedHeader")
```

service/header/header.go
```diff
@@ -46,3 +47,36 @@ func (eh *ExtendedHeader) UnmarshalBinary(data []byte) error {
...
+}
+
+// MarshalBinary marshals ExtendedHeaderRequest to binary.
+func (ehr *ExtendedHeaderRequest) MarshalBinary() ([]byte, error) {
+	return MarshalExtendedHeaderRequest(ehr)
+}
+
+func (ehr *ExtendedHeaderRequest) UnmarshalBinary(data []byte) error {
+	if ehr == nil {
+		return fmt.Errorf("expected: cannot UnmarshalBinary - nil ExtendedHeader")
```
liamsi, line 65: same as above

service/header/service.go
```diff
@@ -32,32 +32,32 @@ func NewHeaderService(exchange Exchange, store Store, pubsub *pubsub.PubSub) *Se
 	}
 }
 
-// Start starts the header Service.
+// Start starts the expected Service.
```
liamsi, line 35: Wait, I see a pattern here. Why is every "header" replaced with "expected"?

service/header/exchange.go
Wondertan: > Here, so far, I only see a very simple request of something that won't constantly be requested in large numbers.

It's header sync. Full nodes will serve millions of such msgs per day for light clients and other nodes.

Wondertan: > On another note, do you think it makes sense to let the user choose the max in the serde library instead?
Yes, that is why there is a TODO saying so.

liamsi: >  Full nodes will serve thousands of such msgs per day for light clients and other nodes.

Yeah, not sure that really justifies a pool still. but all good. Let's revisit mem consumption and load once we have a more realistic network or a more realistic testing framework.

service/header/service.go
renaynay: refactoring probably gone wrong...

liamsi (dismissed):
LFGTM

service/header/exchange_test.go
```diff
@@ -0,0 +1,254 @@
...
+		require.NoError(t, err)
+		eh, err := ProtoToExtendedHeader(resp)
+		require.NoError(t, err)
+		// compare
+		assert.Equal(t, store.headers[int(i)].Height, eh.Height)
+		assert.Equal(t, store.headers[int(i)].Hash(), eh.Hash())
+	}
+}
+
+type mockStore struct {
```
Wondertan, line 225: I changed the Store interface a bit in the Store PR, so we should merge it first and slightly change this mock structure.

service/header/exchange.go
```diff
@@ -0,0 +1,194 @@
...
+	if err != nil {
+		log.Errorw("reading header request from stream", "err", err.Error())
+		//nolint:errcheck
+		stream.Reset()
+		return
+	}
+	request := &ExtendedHeaderRequest{
+		Origin: pbreq.Origin,
+		Amount: pbreq.Amount,
+	}
```
Wondertan, line 52: Let's just use the pb definition directly. There is no point of having a separate struct just to be passed inside request handling logic.

service/header/exchange.go
```diff
@@ -0,0 +1,194 @@
...
+	// route depending on amount of headers requested
+	if request.Amount > 1 {
+		e.handleMultipleHeaderRequest(request, stream)
+	} else {
+		e.handleSingleHeaderRequest(request.Origin, stream) // TODO @renaynay: should we parallelise this?
+	}
```
Wondertan, line 58: You don't need this if tbh. You can loop around ‘request.Amount’ the same as you do handleMultipleHeaderRequest, so it there would be generalized logic for single and multiple cases. Really, they are identical.

service/header/exchange.go
```diff
@@ -0,0 +1,194 @@
...
+// handleMultipleHeaderRequest handles responding to an ExtendedHeaderRequest for
+// multiple headers.
+func (e *exchange) handleMultipleHeaderRequest(request *ExtendedHeaderRequest, stream network.Stream) {
+	height := request.Origin
+	amount := uint64(0)
+	for amount < request.Amount {
+		e.handleSingleHeaderRequest(height, stream)
+		height++
+		amount++
+	}
```
Wondertan, line 101: Once you done writing, you must close the stream to free up allocated resources.

Wondertan (changes requested):

service/header/exchange.go
```diff
@@ -0,0 +1,173 @@
...
+// writes it to the stream.
+func (e *exchange) handleRequest(origin uint64, stream network.Stream) {
+	var (
+		header *ExtendedHeader
+		err    error
+	)
+	if origin == uint64(0) {
+		header, err = e.store.Head()
+	} else {
+		header, err = e.store.GetByHeight(context.Background(), origin)
```
Wondertan, line 70: We should use GetRangeByHeight for ranges, pls.

liamsi (dismissed):
LFG 🚀 I'm glad @Wondertan did review this as well and found a bunch of inconsistencies/bugs related to stream lifecycles.

liamsi (approved):
Only got better

Wondertan:
We need to rebase on top of the store before we can merge it.

Wondertan (approved):

liamsi (approved):
:shipit:

PR #193, opened 2021-11-12
add the celestia-app's cli

evan-forbes:
WIP, sitll need to debug and put possibly move commands.

blocked by #176

### DIFF
--- a/cmd/app.go
+++ b/cmd/app.go
@@ -1,0 +1,55 @@
+package cmd
+
+import (
+	"io"
+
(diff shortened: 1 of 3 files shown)


renaynay:
@evan-forbes @Wondertan Do we still want to add this?

Wondertan:
Good question. I would still add this so the planned Bridge/Validator node could be fully functional in the embedded case. This is a low priority though and may remain open for some time.

Wondertan:
Same as with #176. I am closing this for now but keeping the branch. This might be a good place to return back to

PR #195, opened 2021-11-12
service/header: Store implementation 

### LINKED ISSUE #150: `ExtendedHeaderStore` -- ability for both `light` and `full` nodes to store the extended headers

Wondertan:
## Content
* Adjacent Headers verification logic
* Basic Validation for ExtendedHeaders
* Updated and simplified Store interface.
* Actual Store implementation
	* Height indexer

## TODO
- [x] Testing utilities
- [x] Testing
- [x] Docs

## Refs
Closes #150

### DIFF
--- a/service/header/interface.go
+++ b/service/header/interface.go
@@ -2,6 +2,7 @@ package header
 
 import (
 	"context"
+	"errors"
 
 	tmbytes "github.com/tendermint/tendermint/libs/bytes"
 )
@@ -35,24 +36,22 @@ type Exchange interface {
 	RequestHeaders(ctx context.Context, from, to int64) ([]*ExtendedHeader, error)
 }
 
+// ErrNotFound is returned when there is no requested header.
+var ErrNotFound = errors.New("header: not found")
+
 // Store encompasses the behavior necessary to store and retrieve ExtendedHeaders
 // from a node's local storage.
 type Store interface {
 	// Head returns the ExtendedHeader of the chain head.
-	Head() (*ExtendedHeader, error)
-
+	Head(context.Context) (*ExtendedHeader, error)
 	// Get returns the ExtendedHeader corresponding to the given hash.
-	Get(ctx context.Context, hash tmbytes.HexBytes) (*ExtendedHeader, error)
-	// GetMany returns the ExtendedHeaders corresponding to the given hashes.
-	GetMany(ctx context.Context, hashes []tmbytes.HexBytes) ([]*ExtendedHeader, error)
-
+	Get(context.Context, tmbytes.HexBytes) (*ExtendedHeader, error)
 	// GetByHeight returns the ExtendedHeader corresponding to the given block height.
-	GetByHeight(ctx context.Context, height int64) (*ExtendedHeader, error)
+	GetByHeight(context.Context, int64) (*ExtendedHeader, error)
 	// GetRangeByHeight returns the given range of ExtendedHeaders.
 	GetRangeByHeight(ctx context.Context, from, to int64) ([]*ExtendedHeader, error)
-
-	// Put stores the given ExtendedHeader.
-	Put(ctx context.Context, header *ExtendedHeader) error
-	// PutMany stores the given ExtendedHeaders.
-	PutMany(ctx context.Context, headers []*ExtendedHeader) error
+	// Has check whether Header is stored.
+	Has(context.Context, tmbytes.HexBytes) (bool, error)
+	// Append verifies and stores the given ExtendedHeaders.
+	Append(context.Context, ...*ExtendedHeader) error
 }
--- a/service/header/interface.go
+++ b/service/header/interface.go
@@ -50,7 +50,7 @@ type Store interface {
 	GetByHeight(context.Context, int64) (*ExtendedHeader, error)
 	// GetRangeByHeight returns the given range of ExtendedHeaders.
 	GetRangeByHeight(ctx context.Context, from, to int64) ([]*ExtendedHeader, error)
-	// Has check whether Header is stored.
+	// Has checks whether ExtendedHeader is already stored.
 	Has(context.Context, tmbytes.HexBytes) (bool, error)
 	// Append verifies and stores the given ExtendedHeaders.
 	Append(context.Context, ...*ExtendedHeader) error
--- a/service/header/interface.go
+++ b/service/header/interface.go
@@ -66,7 +66,7 @@ type Store interface {
 	// Has checks whether ExtendedHeader is already stored.
 	Has(context.Context, tmbytes.HexBytes) (bool, error)
 
-	// Append stores and verifies given ExtendedHeader.
+	// Append stores and verifies the given ExtendedHeader(s).
 	// It requires them to be adjacent and in ascending order.
 	Append(context.Context, ...*ExtendedHeader) error
 }
--- a/service/header/interface.go
+++ b/service/header/interface.go
@@ -3,6 +3,7 @@ package header
 import (
 	"context"
 	"errors"
+	"fmt"
 
 	tmbytes "github.com/tendermint/tendermint/libs/bytes"
 )
@@ -36,8 +37,13 @@ type Exchange interface {
 	RequestHeaders(ctx context.Context, from, to int64) ([]*ExtendedHeader, error)
 }
 
-// ErrNotFound is returned when there is no requested header.
-var ErrNotFound = errors.New("header: not found")
+var (
+	// ErrNotFound is returned when there is no requested header.
+	ErrNotFound = errors.New("header: not found")
+
+	// ErrNoHead is returned when Store does not contain Head of the chain,
+	ErrNoHead = fmt.Errorf("header/store: no chain head")
+)
 
 // Store encompasses the behavior necessary to store and retrieve ExtendedHeaders
 // from a node's local storage.
--- a/service/header/interface.go
+++ b/service/header/interface.go
@@ -42,6 +42,8 @@ var ErrNotFound = errors.New("header: not found")
 // Store encompasses the behavior necessary to store and retrieve ExtendedHeaders
 // from a node's local storage.
 type Store interface {
+	// Open opens and initializes Store.
+	Open(context.Context) error
 	// Head returns the ExtendedHeader of the chain head.
 	Head(context.Context) (*ExtendedHeader, error)
 	// Get returns the ExtendedHeader corresponding to the given hash.
--- a/service/header/interface.go
+++ b/service/header/interface.go
@@ -50,16 +50,23 @@ var (
 type Store interface {
 	// Open opens and initializes Store.
 	Open(context.Context) error
+
 	// Head returns the ExtendedHeader of the chain head.
 	Head(context.Context) (*ExtendedHeader, error)
+
 	// Get returns the ExtendedHeader corresponding to the given hash.
 	Get(context.Context, tmbytes.HexBytes) (*ExtendedHeader, error)
+
 	// GetByHeight returns the ExtendedHeader corresponding to the given block height.
 	GetByHeight(context.Context, int64) (*ExtendedHeader, error)
-	// GetRangeByHeight returns the given range of ExtendedHeaders.
+
+	// GetRangeByHeight returns the given range [from:to) of ExtendedHeaders.
 	GetRangeByHeight(ctx context.Context, from, to int64) ([]*ExtendedHeader, error)
+
 	// Has checks whether ExtendedHeader is already stored.
 	Has(context.Context, tmbytes.HexBytes) (bool, error)
-	// Append verifies and stores the given ExtendedHeaders.
+
+	// Append stores and verifies given ExtendedHeader.
+	// It requires them to be adjacent and in ascending order.
 	Append(context.Context, ...*ExtendedHeader) error
 }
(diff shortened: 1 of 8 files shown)


Wondertan:
Tests will come soon. Unfortunately, we can reuse testing utilities from Tendermint and have to write our own in a reusable way, which adds time overhead.

adlerjohn (commented):
Why is go 1.17 failing but not 1.16?

Wondertan:
@adlerjohn, that's a https://github.com/celestiaorg/celestia-node/issues/173

service/header/store.go
```diff
@@ -0,0 +1,302 @@
...
+		cache: cache,
+		index: index,
+	}, nil
+}
+
+func (s *store) Start(context.Context) error {
+	err := s.loadHead()
+	if err == datastore.ErrNotFound {
+		return fmt.Errorf("header/store: no chain head")
+	}
```
liamsi, line 79: Won't this always happen if the first constructor is used to init this?
```
func NewStore(ds datastore.Batching) (Store, error) {
	return newStore(ds)
}
```

service/header/store.go
```diff
@@ -0,0 +1,302 @@
...
+
+	return err
+}
+
+func (s *store) Head(ctx context.Context) (*ExtendedHeader, error) {
+	s.headLk.RLock()
+	head := s.head
+	s.headLk.RUnlock()
+	return s.Get(ctx, head)
```
liamsi, line 88: What happens if there is no head yet? error or nil?

service/header/store.go
```diff
@@ -0,0 +1,302 @@
...
+			return nil, ErrNotFound
+		}
+
+		return nil, err
+	}
+
+	return s.Get(ctx, hash)
+}
+
+func (s *store) GetRangeByHeight(ctx context.Context, from, to int64) ([]*ExtendedHeader, error) {
```
liamsi, line 121: Gimme all the height! ;-)

service/header/store.go
```diff
@@ -0,0 +1,302 @@
...
+
+func (s *store) Has(_ context.Context, hash bytes.HexBytes) (bool, error) {
+	if ok := s.cache.Contains(hash); ok {
+		return ok, nil
+	}
+
+	return s.ds.Has(datastore.NewKey(hash.String()))
+}
+
+func (s *store) Append(ctx context.Context, headers ...*ExtendedHeader) error {
```
liamsi, line 149: I think this deserves a comment that the passed-in headers are expected to be adjacent / a chain of consecutive headers.

service/header/store.go
```diff
@@ -0,0 +1,302 @@
...
+	for _, h := range headers {
+		err = VerifyAdjacent(head, h)
+		if err != nil {
+			log.Errorw("invalid header", "head", head.Hash(), "header", h.Hash(), "err", err)
+			break // if some headers are cryptographically valid, why not include them? Exactly, so let's include
+		}
+		verified, head = append(verified, h), h
+	}
+	if len(verified) == 0 {
+		return fmt.Errorf("header/store: no valid headers were given")
```
liamsi, line 170: Just a question because I'm not used to this pattern: did you and @renaynay agree on prefixing each error with the module?

service/header/verify.go
```diff
@@ -0,0 +1,52 @@
+package header
+
+import (
+	"bytes"
+	"fmt"
+	"time"
+)
+
+func VerifyAdjacent(trusted, untrusted *ExtendedHeader) error {
```
liamsi, line 9: Doesn't this logic exist in tendermint and we can use it as a library instead?

service/header/verify.go
```diff
@@ -0,0 +1,52 @@
...
+	if !bytes.Equal(untrusted.ValidatorsHash, trusted.NextValidatorsHash) {
+		return fmt.Errorf("expected old header next validators (%X) to match those from new header (%X)",
+			trusted.NextValidatorsHash,
+			untrusted.ValidatorsHash,
+		)
+	}
+
+	// Ensure that +2/3 of new validators signed correctly.
+	if err := untrusted.ValidatorSet.VerifyCommitLight(trusted.ChainID, untrusted.Commit.BlockID,
+		untrusted.Height, untrusted.Commit); err != nil {
```
liamsi, line 47: nit: I would suggest putting each argument on a new line.

Also: cool, the verify commit logic is re-used from tendermint.

liamsi (commented):
Beautiful PR 👏🏼  Needs some tests though.

service/header/store.go
Wondertan: This is why for the first time you have to use `NewStoreWithHead` and then `NewStore`. Not sure thought how convenient this is.

service/header/store.go
Wondertan: Well, I guess there will be a panic. Fortunately, this case should never happen if Store is properly initialized.

service/header/store.go
Wondertan: *hate* 😄

service/header/store.go
Wondertan: True

Wondertan: Will add to the comment to the interface

service/header/store.go
Wondertan: I don't remember us discussing this, so no.

Wondertan: What would you do instead?

service/header/store.go
```diff
@@ -0,0 +1,302 @@
...
+func (s *store) Head(ctx context.Context) (*ExtendedHeader, error) {
+	s.headLk.RLock()
+	head := s.head
+	s.headLk.RUnlock()
+	return s.Get(ctx, head)
+}
+
+func (s *store) Get(_ context.Context, hash bytes.HexBytes) (*ExtendedHeader, error) {
+	if v, ok := s.cache.Get(hash); ok {
+		return v.(*ExtendedHeader), nil
```
renaynay, line 93: This *could technically* panic

service/header/verify.go
Wondertan: Yeah and I import their code initially, but I decided to reimplement it, as they have a trust period for headers and we don't need them AFAIU, because we need to be able to sync anytime. Also, it looks cleaner when we use our data structure and it us just less code.

One interesting thing though. They also have `VerifyUnadjasent`, where you can verify a new header by a validator set from the past. If there is some overlap then the Header is secure. This is also proof of why backward syncing will work.

service/header/verify.go
Wondertan: Yeah, this part is easy and pleasant to reuse, so why not.

service/header/store.go
liamsi: This is not obvious at all and this is not documented anywhere IIUC.

service/header/store.go
liamsi: I don't think the `header/store:` adds much to the error. But I see that everywhere else the errors are prefixed similarly. So it's OK to keep this consistent for now.

renaynay: I like it

service/header/store.go
```diff
@@ -0,0 +1,302 @@
...
+	err = s.head.UnmarshalJSON(b)
+	if err != nil {
+		return err
+	}
+
+	log.Infow("loaded head", "hash", s.head)
+	return nil
+}
+
+func (s *store) newHead(head bytes.HexBytes) error {
```
renaynay, line 229: pls document helper methods like this

service/header/store.go
```diff
@@ -0,0 +1,302 @@
...
+	log.Infow("new head", "hash", head)
+
+	// at this point Header body of the given 'head' must be already written with put
```
renaynay, line 232: you can mention this in the docs of this method

renaynay (commented):
Initial review, but waiting for tests!

service/header/verify.go
liamsi: >  as they have a trust period for headers and we don't need them AFAIU, 

I don't understand why in our case we would not need a trust period. Same as in the tendermint case a trusted header expires after the trust period.

> If there is some overlap then the Header is secure. This is also proof of why backward syncing will work.

True. But the use-case is slightly different because there the headers in between the trusted height and the new height are skipped because of the overlap of the validators. This is thoroughly analyzed and well within the security assumptions of tendermint. We should do the same if we want to implement backward syncing (cc @musalbas @adlerjohn)

service/header/store.go
liamsi: OK, at some point we should add a contribution guidline document that states that this is expected (and why).

service/header/store.go
Wondertan: Lol, no

service/header/verify.go
Wondertan: > I don't understand why in our case we would not need a trust period. Same as in the tendermint case a trusted header expires after the trust period.

What is trust period then and why do __we__ need? Note that the Store is for both light/full nodes and when you sync from scratch you don't have a trust period for headers you request. Also, we verify headers cryptographically, so if it's valid by the PoS rules, why would we need a trust period?

service/header/store.go
Wondertan: I was confused about this msg, as I remember explaining this with comments, but later realized I didn't commit that part 😅

service/header/store.go
renaynay: https://github.com/celestiaorg/celestia-node/issues/198

service/header/verify.go
liamsi: > What is trust period then and why do **we** need it?

You can go offline as a light client for any duration < trust period without having to be subjectively reinitialized. It's the same use-case as for tendermint.

liamsi: > Also, we verify headers cryptographically, so if it's valid by the PoS rules, why would we need a trust period for them?

https://blog.ethereum.org/2014/11/25/proof-stake-learned-love-weak-subjectivity/

service/header/store.go
Wondertan: Done

service/header/verify.go
Wondertan: @liamsi, Ok, I see why I was wrong. Thank you for the explanation!

Wondertan: I created an issue to track the resolution of long-range attacks https://github.com/celestiaorg/celestia-node/issues/205(currently empty). As discussed in sync, let's not fix the issue before devnet. Also, importing Tendermint implementation of `VerifyAdjacent` would not just help as we should also implement additional logic for unboding period checking.

service/header/store.go
```diff
@@ -0,0 +1,302 @@
...
+	"github.com/ipfs/go-datastore"
+	"github.com/ipfs/go-datastore/namespace"
+
+	"github.com/celestiaorg/celestia-core/libs/bytes"
+)
+
+// TODO(@Wondertan): Those values must be configurable and proper defaults should be set for specific node type.
+const (
+	// DefaultStoreCache defines the amount of max entries allowed in the Header Store cache.
+	DefaultStoreCache = 1024
```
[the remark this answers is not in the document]
Wondertan: Fixed

service/header/store.go
Wondertan: Done

service/header/store.go
Wondertan: Ok

service/header/interface.go
```diff
@@ -50,16 +50,23 @@ var (
...
 	Head(context.Context) (*ExtendedHeader, error)
+
 	// Get returns the ExtendedHeader corresponding to the given hash.
 	Get(context.Context, tmbytes.HexBytes) (*ExtendedHeader, error)
+
 	// GetByHeight returns the ExtendedHeader corresponding to the given block height.
 	GetByHeight(context.Context, int64) (*ExtendedHeader, error)
-	// GetRangeByHeight returns the given range of ExtendedHeaders.
+
+	// GetRangeByHeight returns the given range [from:to) of ExtendedHeaders.
```
renaynay, line 63: `[from:to]`

service/header/interface.go
```diff
@@ -50,16 +50,23 @@ var (
...
 	GetRangeByHeight(ctx context.Context, from, to int64) ([]*ExtendedHeader, error)
+
 	// Has checks whether ExtendedHeader is already stored.
 	Has(context.Context, tmbytes.HexBytes) (bool, error)
-	// Append verifies and stores the given ExtendedHeaders.
+
+	// Append stores and verifies given ExtendedHeader.
+	// It requires them to be adjacent and in ascending order.
```
renaynay, line 70:
```suggestion
	// Append stores and verifies the given ExtendedHeader(s).
	// It requires them to be adjacent and in ascending order.
```

renaynay (commented):
some nits otherwise good + rebase

service/header/store.go
```diff
@@ -0,0 +1,318 @@
...
+
+// NewStore constructs a Store over datastore.
+// The datastore must have a head there otherwise Start will error.
+// For first initialization of Store use NewStoreWithHead.
+func NewStore(ds datastore.Batching) (Store, error) {
+	return newStore(ds)
+}
+
+// NewStoreWithHead initiates a new Store and forcefully sets a given trusted header as head.
+func NewStoreWithHead(ds datastore.Batching, head *ExtendedHeader) (Store, error) {
```
liamsi, line 41: I wish we could have one foolproof constructor instead.

Wondertan: I don't like passing nil into constructors and this would be the case in a unified constructor if a head exists.

Wondertan: But I agree this is not good enough and there might be a better UX for this

Wondertan: For example, we can have one constructor and a helper function that will inject a new Head into Datastore.

liamsi (approved):
🚀 👏🏼 🥇

renaynay (approved):

PR #196, opened 2021-11-14
cmd: add flag to start node on remote core

### LINKED ISSUE #142: Add flag to pass address of remote Core node
This flag should tell Node to use a remote Core node instead of starting an embedded Core node process.

vgonkivs:
* Added flag that tells node to connect to the remote address.

Closes #142 

P.S. Unfortunately, I was not able to test it properly. I have tried to create and run `StartRemoteClient()` from ` core/testing.go ` but got  an error `Error: node: failed to start: client not running`.  Did I do something wrong?

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"net/url"
 	"os/signal"
 	"syscall"
 
@@ -18,7 +19,9 @@ func Start(repoName string, tp node.Type) *cobra.Command {
 	if len(repoName) == 0 && tp != node.Dev { // repository path not necessary for **DEV MODE**
 		panic("parent command must specify a persistent flag name for repository path")
 	}
-	return &cobra.Command{
+
+	const cfgAddress = "core.remote"
+	cmd := &cobra.Command{
 		Use:          "start",
 		Short:        "Starts Node daemon. First stopping signal gracefully stops the Node and second terminates it.",
 		Aliases:      []string{"run", "daemon"},
@@ -43,13 +46,22 @@ func Start(repoName string, tp node.Type) *cobra.Command {
 				return err
 			}
 
-			return start(cmd, tp, repo)
+			u, err := url.Parse(cmd.Flag(cfgAddress).Value.String())
+			opts := make([]node.Options, 0)
+			if err == nil && u.Scheme != "" && u.Host != "" {
+				opts = append(opts, node.WithRemoteClient(u.Scheme, u.Host))
+			}
+
+			return start(cmd, tp, repo, opts...)
 		},
 	}
+
+	cmd.Flags().String(cfgAddress, "", "Indicates node to connect to the remote core")
+	return cmd
 }
 
-func start(cmd *cobra.Command, tp node.Type, repo node.Repository) error {
-	nd, err := node.New(tp, repo)
+func start(cmd *cobra.Command, tp node.Type, repo node.Repository, opts ...node.Options) error {
+	nd, err := node.New(tp, repo, opts...)
 	if err != nil {
 		return err
 	}
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -56,7 +56,7 @@ func Start(repoName string, tp node.Type) *cobra.Command {
 		},
 	}
 
-	cmd.Flags().String(cfgAddress, "", "Indicates node to connect to the remote core")
+	cmd.Flags().String(cfgAddress, "", "Indicates node to connect to the given remote core node")
 	return cmd
 }
 
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -1,6 +1,8 @@
 package cmd
 
 import (
+	"errors"
+	"net"
 	"net/url"
 	"os/signal"
 	"syscall"
@@ -8,6 +10,10 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/celestiaorg/celestia-node/node"
+)
+
+var (
+	errIncorrectAddress = errors.New("incorrect address provided")
 )
 
 // Start constructs a CLI command to start Celestia Node daemon of the given type 'tp'.
@@ -46,10 +52,14 @@ func Start(repoName string, tp node.Type) *cobra.Command {
 				return err
 			}
 
-			u, err := url.Parse(cmd.Flag(cfgAddress).Value.String())
+			address := cmd.Flag(cfgAddress).Value.String()
 			opts := make([]node.Options, 0)
-			if err == nil && u.Scheme != "" && u.Host != "" {
-				opts = append(opts, node.WithRemoteClient(u.Scheme, u.Host))
+			if address != "" {
+				protocol, ip, err := parseAddress(address)
+				if err != nil {
+					return err
+				}
+				opts = append(opts, node.WithRemoteClient(protocol, ip))
 			}
 
 			return start(cmd, tp, repo, opts...)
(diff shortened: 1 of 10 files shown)


Wondertan:
@vgonkivs, this should work over ‘start’ command and not ‘init’. We should start the node with a remote core, even when the configuration is initialized to work with an embedded one. Essentially, flags should override configurations in runtime, not overwrite them on disk.

Wondertan (changes requested):
@vgonkivs, this should work over ‘start’ command and not ‘init’. We should start the node with a remote core, even when the configuration is initialized to work with an embedded one. Essentially, flags should override configurations in runtime, not overwrite them on disk.

vgonkivs:
Thanks for explanation @Wondertan . Do we need to handle it for dev node?

Wondertan:
@vgonkivs, you can omit it this for the dev mode, as it is going to be reworked.

renaynay:
@vgonkivs there is a failing test for starting the remote client - let me know if you can resolve this. Happy to help as well.

vgonkivs:
Hi, @renaynay. Currently investigating what went wrong. Panic happened on newly added test for node starting over remote core.

`github.com/celestiaorg/celestia-core/libs/pubsub.(*Subscription).Out(0x0, 0x0)
/home/runner/go/pkg/mod/github.com/celestiaorg/celestia-core@v0.0.2-0.20210924001615-488ac31b4b3c/libs/pubsub/subscription.go:43 +0x5
github.com/celestiaorg/celestia-core/rpc/core.Subscribe.func1(0x1b25db0, 0x0, 0xc016c2c090, 0x13, 0x1b0a3e0, 0x25bee00, 0xc06dcb9920, 0xc08337ca40, 0xf)
/home/runner/go/pkg/mod/github.com/celestiaorg/celestia-core@v0.0.2-0.20210924001615-488ac31b4b3c/rpc/core/events.go:50 +0x4c
created by github.com/celestiaorg/celestia-core/rpc/core.Subscribe
/home/runner/go/pkg/mod/github.com/celestiaorg/celestia-core@v0.0.2-0.20210924001615-488ac31b4b3c/rpc/core/events.go:47 +0x4bd`

Based on the stack trace, it seems that somehow pubsub subscription was not created...
Strange behaviour, but all tests have been passed several times on my machine.
Will be glad to get help from you.

vgonkivs:
Now all my test have passed but another issue appears on TestSubscriber on pipeline with 1.16

renaynay:
Don't @vgonkivs, there is something wrong with the subscriber stuff on IPFS side I believe, so we're gonna ignore the flakeyness for now.

Bidon15 (dismissed):
Works flawlessly. No hickups so far 👏

cmd/start.go
```diff
@@ -43,13 +46,22 @@ func Start(repoName string, tp node.Type) *cobra.Command {
 				return err
 			}
 
-			return start(cmd, tp, repo)
+			u, err := url.Parse(cmd.Flag(cfgAddress).Value.String())
+			opts := make([]node.Options, 0)
+			if err == nil && u.Scheme != "" && u.Host != "" {
+				opts = append(opts, node.WithRemoteClient(u.Scheme, u.Host))
+			}
+
```
Wondertan, line 54: 👍🏻 for proper `url.Parse` usage. 

Pls extract parsing into a separate small helper function. Also, we should error to the user if he passes a malformed URL and ideally check if port presents inside parsed url.Host.

node/core/core.go
```diff
@@ -26,7 +26,12 @@ func DefaultConfig() Config {
 // Components collects all the components and services related to managing the relationship with the Core node.
 func Components(cfg Config) fx.Option {
 	return fx.Options(
-		fxutil.ProvideIf(cfg.Remote, RemoteClient),
+		fxutil.ProvideIf(cfg.Remote, func() (core.Client, error) {
+			return RemoteClient(cfg)
+		}),
+		fxutil.InvokeIf(cfg.Remote, func(c core.Client) error {
+			return c.Start()
+		}),
```
Wondertan, line 34: Nice catch!

node/fxutil/util.go
```diff
@@ -26,3 +26,11 @@ func ProvideIf(cond bool, ctor interface{}) fx.Option {
...
 	return fx.Options()
 }
+
+// ProvideIf invokes a given function if a condition is met.
+func InvokeIf(cond bool, function interface{}) fx.Option {
+	if cond {
+		return fx.Invoke(function)
+	}
+	return fx.Options()
+}
```
Wondertan, line 36: Awesome! I am so glad that you are catching the pattern here!

Wondertan (commented):
@vgonkivs, man, this is amazing, thank you!

Left a few comments, but otherwise LGTM!

Wondertan:
@vgonkivs, one more thing. We recently made an update in https://github.com/celestiaorg/celestia-node/pull/175, where we changed all the imports from `celestiaorg/celestia-core` to `tendermint/termint`. This change also affects your PR so pls do a rebase on the main branch.

cmd/start.go
```diff
@@ -43,13 +46,22 @@ func Start(repoName string, tp node.Type) *cobra.Command {
...
+			return start(cmd, tp, repo, opts...)
 		},
 	}
+
+	cmd.Flags().String(cfgAddress, "", "Indicates node to connect to the remote core")
```
renaynay, line 59:
```suggestion
	cmd.Flags().String(cfgAddress, "", "Indicates node to connect to the given remote core node")
```

core/testing.go
```diff
@@ -40,6 +40,13 @@ func StartRemoteClient() (*node.Node, Client, error) {
 	return remote, client, err
 }
 
+// StartRemoteCore starting a remote core and returns it's protocol and address
```
renaynay, line 43:
```suggestion
// StartRemoteCore starts a remote core node and returns its protocol and address
```

renaynay (commented):
some nits but otherwise LGTM! and yes we need a rebase before merge :)

vgonkivs:
Please let me know if everything is ok , so I can squash commits into one.

liamsi:
Awesome! Thanks! Did someone try running this? @vgonkivs @Bidon15 ?

vgonkivs:
Hi @liamsi . It works for me with --core.remote on mocked Node and without flag(embedded).

Bidon15:
> Awesome! Thanks! Did someone try running this? @vgonkivs @Bidon15 ?

Yeap. All good for me 👍🔥

cmd/start.go
```diff
@@ -9,6 +12,10 @@ import (
 	"github.com/celestiaorg/celestia-node/node"
 )
 
+var (
+	errIncorrectAddress = errors.New("incorrect address provided")
+)
```
Wondertan, line 17: nit: we don't make errors as a separate var unless there is one more usage for it or at least there should a plan/use-case on how it can be used later.

nit2: it would be nice if the error can report the connection to the remote core. We have multiple addresses, so it may be confusing which address the error is telling about. Appending `for Remote Core` would fix this.

Wondertan (dismissed):
Besides two nits, LGTM

cmd/start.go
Wondertan: @vgonkivs

vgonkivs: @Wondertan , done.

Wondertan (commented):
Awesome! Thanks again @vgonkivs.

Wondertan (approved):

Bidon15 (approved):

PR #203, opened 2021-11-16
service/block, service/header: `BlockService` broadcasts generated `ExtendedHeader`s to network

renaynay:
`block.Service` now receives a `Broadcaster` from the `header.Service` so that it can `Broadcast()` new `ExtendedHeader`s to the network as they are created during the block handling process.

### DIFF
--- a/node/services/service.go
+++ b/node/services/service.go
@@ -3,31 +3,36 @@ package services
 import (
 	"context"
 
+	ipld "github.com/ipfs/go-ipld-format"
 	"github.com/ipfs/go-merkledag"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
+	"go.uber.org/fx"
 
 	"github.com/celestiaorg/celestia-node/das"
 	"github.com/celestiaorg/celestia-node/node/fxutil"
 	"github.com/celestiaorg/celestia-node/service/block"
 	"github.com/celestiaorg/celestia-node/service/header"
 	"github.com/celestiaorg/celestia-node/service/share"
-
-	ipld "github.com/ipfs/go-ipld-format"
-	"go.uber.org/fx"
 )
 
-func Header(lc fx.Lifecycle, ps *pubsub.PubSub) *header.Service {
+// Header constructs a new header.Service.
+func Header(lc fx.Lifecycle, ps *pubsub.PubSub) (*header.Service, header.Broadcaster) {
 	service := header.NewHeaderService(nil, nil, ps)
 	lc.Append(fx.Hook{
 		OnStart: service.Start,
 		OnStop:  service.Stop,
 	})
-	return service
+	return service, service
 }
 
 // Block constructs new block.Service.
-func Block(lc fx.Lifecycle, fetcher block.Fetcher, store ipld.DAGService) *block.Service {
-	service := block.NewBlockService(fetcher, store)
+func Block(
+	lc fx.Lifecycle,
+	fetcher block.Fetcher,
+	store ipld.DAGService,
+	broadcaster header.Broadcaster,
+) *block.Service {
+	service := block.NewBlockService(fetcher, store, broadcaster)
 	lc.Append(fx.Hook{
 		OnStart: service.Start,
 		OnStop:  service.Stop,
--- a/service/block/event_test.go
+++ b/service/block/event_test.go
@@ -5,7 +5,10 @@ import (
 	"testing"
 	"time"
 
+	"github.com/ipfs/go-datastore"
 	md "github.com/ipfs/go-merkledag/test"
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
+	mocknet "github.com/libp2p/go-libp2p/p2p/net/mock"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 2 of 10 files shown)


node/services/service.go
```diff
@@ -2,22 +2,25 @@ package services
...
-	ipld "github.com/ipfs/go-ipld-format"
-	"go.uber.org/fx"
 )
 
 func Header(lc fx.Lifecycle, ps *pubsub.PubSub) *header.Service {
-	service := header.NewHeaderService(nil, nil, ps)
+	service, err := header.NewHeaderService(nil, nil, ps)
+	if err != nil {
+		panic(fmt.Errorf("failed to create new header service: %w", err))
+	}
```
Wondertan, line 23: You can make it return an error, pls no panics

node/services/service.go
```diff
@@ -26,8 +29,13 @@ func Header(lc fx.Lifecycle, ps *pubsub.PubSub) *header.Service {
...
 // Block constructs new block.Service.
-func Block(lc fx.Lifecycle, fetcher block.Fetcher, store ipld.DAGService) *block.Service {
-	service := block.NewBlockService(fetcher, store)
+func Block(
+	lc fx.Lifecycle,
+	fetcher block.Fetcher,
+	store ipld.DAGService,
+	broadcaster header.Broadcaster,
+) *block.Service {
+	service := block.NewBlockService(fetcher, store, broadcaster)
```
Wondertan, line 38: It is ok for now, but I would feed it with the whole header service instead of Broadcaster. The reason is that we won't have multiple Broadcaster and broadcasting will happen only in one specific place, so it seems to me like there is no point in abstraction. I know that you want to make it similar to Subscriber, like if there is a Subscriber then there should be Publisher respectively.

Wondertan: Note, I am ok to leave it like this for now.

renaynay: Yeah I don't think that block service needs access to the whole header service -- it gets only the instance of Broadcaster from the header service btw - it doesn't instantiate its own. Do you think it should and then only get fed pubsub from the DI app?

node/services/service.go
```diff
@@ -2,32 +2,40 @@ package services
...
 	"github.com/celestiaorg/celestia-node/service/header"
 	"github.com/celestiaorg/celestia-node/service/share"
-
-	ipld "github.com/ipfs/go-ipld-format"
-	"go.uber.org/fx"
 )
 
-func Header(lc fx.Lifecycle, ps *pubsub.PubSub) *header.Service {
-	service := header.NewHeaderService(nil, nil, ps)
+func Header(lc fx.Lifecycle, ps *pubsub.PubSub) (*header.Service, error) {
```
adlerjohn, line 19: Nit, but can you document this exported function too?

service/block/event_test.go
```diff
@@ -73,16 +122,43 @@ func (m *mockFetcher) UnsubscribeNewBlockEvent(ctx context.Context) error {
...
+
 // generateBlocks generates new raw blocks and sends them to the mock fetcher,
 // returning the extended blocks generated from the process to compare against.
 func (m *mockFetcher) generateBlocks(t *testing.T, num int) []Block {
 	t.Helper()
 
 	extendedBlocks := make([]Block, num)
 
-	for i := 0; i < num; i++ {
+	for i := 1; i <= num; i++ {
```
adlerjohn, line 158: Why this change? Doesn't seem to have any effects?

liamsi (approved):
Looks good!

service/block/event_test.go
renaynay: relic from when I was debugging something - you're right

liamsi:
Is this new?

```
=== RUN   TestFull_WithRemoteCore
The p2p host is listening on:
*  /ip4/10.1.1.234/tcp/2121/p2p/12D3KooWBCKfb9j5yMQvzQjrzjQMVrpUyszKhoV3zHQB9GPguJc4
*  /ip6/::1/tcp/2121/p2p/12D3KooWBCKfb9j5yMQvzQjrzjQMVrpUyszKhoV3zHQB9GPguJc4
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xb7e780]

goroutine 5508 [running]:
github.com/tendermint/tendermint/libs/pubsub.(*Subscription).Out(0xc00142ba40)
	/home/runner/go/pkg/mod/github.com/celestiaorg/celestia-core@v0.34.14-celestia/libs/pubsub/subscription.go:43
github.com/tendermint/tendermint/rpc/core.Subscribe.func1()
	/home/runner/go/pkg/mod/github.com/celestiaorg/celestia-core@v0.34.14-celestia/rpc/core/events.go:50 +0x9e
created by github.com/tendermint/tendermint/rpc/core.Subscribe
	/home/runner/go/pkg/mod/github.com/celestiaorg/celestia-core@v0.34.14-celestia/rpc/core/events.go:47 +0x485
FAIL	github.com/celestiaorg/celestia-node/node	0.286s
```

node/services/service.go
Wondertan: I just don't think we need Broadcaster. Passing HeaderService to BlockService is temporary. This will be fixed once we start listen for Headers first from Core and not for Blocks

Wondertan:
@liamsi, this is the bug in the core

Wondertan (commented):
Approve as long as we fix some things later https://github.com/celestiaorg/celestia-node/issues/214

Wondertan (approved):

Bidon15 (approved):

Wondertan (approved):

PR #204, opened 2021-11-16
node: add p2p multiaddr log

### LINKED ISSUE #200: node/node.go: Log p2p information of node during `Start()`
Log celestia-node p2p address information so that we can copy and paste it. For help, ask @Wondertan.

Bidon15:
during startup of the node, we log the p2p multiaddr of it
Closes #200

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -102,16 +102,15 @@ func (n *Node) Start(ctx context.Context) error {
 
 	// TODO(@Wondertan): Print useful information about the node:
 	//  * API address
-	//  * Host listening address
 	log.Infof("started %s Node", n.Type)
 
 	addrs, err := peer.AddrInfoToP2pAddrs(host.InfoFromHost(n.Host))
 	if err != nil {
-		log.Errorf("Error while retreiving multiaddress information: %s", err)
+		log.Errorw("Retrieving multiaddress information", "err", err)
 		return err
 	}
 	for _, addr := range addrs {
-		log.Infow("node P2P info", "multiaddress", addr.String())
+		fmt.Println("node P2P info", "multiaddress", addr.String())
 	}
 	return nil
 }
--- a/node/node.go
+++ b/node/node.go
@@ -111,7 +111,7 @@ func (n *Node) Start(ctx context.Context) error {
 		return err
 	}
 	for _, addr := range addrs {
-		log.Infof("P2P multiaddress - %s", addr.String())
+		log.Infow("node P2P info", "multiaddress", addr.String())
 	}
 	return nil
 }
--- a/node/node.go
+++ b/node/node.go
@@ -11,6 +11,7 @@ import (
 	logging "github.com/ipfs/go-log/v2"
 	"github.com/libp2p/go-libp2p-core/connmgr"
 	"github.com/libp2p/go-libp2p-core/host"
+	"github.com/libp2p/go-libp2p-core/peer"
 	"github.com/libp2p/go-libp2p-core/routing"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	"go.uber.org/fx"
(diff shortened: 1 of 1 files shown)


node/node.go
```diff
@@ -101,9 +102,17 @@ func (n *Node) Start(ctx context.Context) error {
 
 	// TODO(@Wondertan): Print useful information about the node:
 	//  * API address
-	//  * Pubkey/PeerID
 	//  * Host listening address
 	log.Infof("started %s Node", n.Type)
+
+	addrs, err := peer.AddrInfoToP2pAddrs(host.InfoFromHost(n.Host))
+	if err != nil {
+		log.Errorf("Error while retreiving multiaddress information: %s", err)
```
renaynay, line 110:
```suggestion
		log.Errorw("Retreiving multiaddress information", "err", err)
```

node/node.go
```diff
@@ -101,9 +102,17 @@ func (n *Node) Start(ctx context.Context) error {
...
+		return err
+	}
+	for _, addr := range addrs {
+		log.Infof("P2P multiaddress - %s", addr.String())
```
renaynay, line 114:
```suggestion
		log.Infow("node P2P info", "multiaddress", addr.String())
```

node/node.go
```diff
@@ -101,9 +102,17 @@ func (n *Node) Start(ctx context.Context) error {
...
 	//  * Host listening address
 	log.Infof("started %s Node", n.Type)
+
+	addrs, err := peer.AddrInfoToP2pAddrs(host.InfoFromHost(n.Host))
+	if err != nil {
+		log.Errorf("Error while retreiving multiaddress information: %s", err)
+		return err
+	}
+	for _, addr := range addrs {
+		log.Infow("node P2P info", "multiaddress", addr.String())
```
Wondertan, line 114: Hmm, I feel like this might be simple `fmt.Println()` to differentiate it from all the logs. It is not a log of events but like something informational that is returned when you exec something in CLI. I don't have a strong opinion here, though. I

node/node.go
```diff
@@ -101,9 +102,17 @@ func (n *Node) Start(ctx context.Context) error {
...
-	//  * Pubkey/PeerID
 	//  * Host listening address
 	log.Infof("started %s Node", n.Type)
+
+	addrs, err := peer.AddrInfoToP2pAddrs(host.InfoFromHost(n.Host))
+	if err != nil {
+		log.Errorf("Error while retreiving multiaddress information: %s", err)
+		return err
+	}
+	for _, addr := range addrs {
```
Wondertan, line 113: Pls, don't print duplicates, so you just need to print the second half `addrs[len(addrs)/2:]`

node/node.go
```diff
@@ -101,9 +102,17 @@ func (n *Node) Start(ctx context.Context) error {
 
 	// TODO(@Wondertan): Print useful information about the node:
 	//  * API address
-	//  * Pubkey/PeerID
 	//  * Host listening address
```
Wondertan, line 105: You can remove this one as well

node/node.go
mattdf: retreiving -> retrieving

node/node.go
Bidon15: Changed to fmt.Println as indeed it can be handy to always see at the beginning the addrs, even if you turn off logging display

node/node.go
```diff
@@ -102,16 +102,15 @@ func (n *Node) Start(ctx context.Context) error {
...
 
 	addrs, err := peer.AddrInfoToP2pAddrs(host.InfoFromHost(n.Host))
 	if err != nil {
-		log.Errorf("Error while retreiving multiaddress information: %s", err)
+		log.Errorw("Retrieving multiaddress information", "err", err)
 		return err
 	}
 	for _, addr := range addrs {
-		log.Infow("node P2P info", "multiaddress", addr.String())
+		fmt.Println("node P2P info", "multiaddress", addr.String())
```
liamsi, line 113: why `fmt.Println`?

Wondertan: Because this information needs to be printed out even if logs are turned off

Wondertan: @Bidon15, Can you make it like: 
```
The p2p Host is listening on:
* Addr
* Addr
```

Wondertan: Users can dump to file anything they want from stdout.

Wondertan: From UX standpoint, I don't thing we need to scary light-node user with tonnes of logs, so by default I would prefer turning them off, but when you start something you need to get a sign that everything works fine and showing a succint information about the address a node listens to - makes sense.

Wondertan (approved):
LGTM!

mattdf (approved):

renaynay (approved):

Issue #69, opened 2021-09-20
docker: Create docker image for Celestia-Node


renaynay:
Create a docker image for `celestia-node` binary. This issue is **blocked** until we have an executable.

sysrex:
I can help with this, and create an initial docker file

Bidon15:
Hey @sysrex, thank you for being pro-active here 🥳 
Please, go ahead with the initial PR for this - this will be awesome

jbowen93:
Hey @sysrex what's your status on this? I just threw together a quick Dockerfile for my own uses. 
```
FROM ubuntu:latest
RUN apt update
RUN apt install -y git wget make gcc

# Golang stuff
RUN wget -P /tmp https://golang.org/dl/go1.17.3.linux-amd64.tar.gz
RUN tar -C /usr/local -xzf /tmp/go1.17.3.linux-amd64.tar.gz
RUN rm /tmp/go1.17.3.linux-amd64.tar.gz
ENV GOPATH /go
ENV PATH $GOPATH/bin:/usr/local/go/bin:$PATH
RUN mkdir -p "$GOPATH/src" "$GOPATH/bin" && chmod -R 777 "$GOPATH"

# Clone from private repository
RUN mkdir -p -m 0600 ~/.ssh && ssh-keyscan github.com >> ~/.ssh/known_hosts
RUN --mount=type=ssh git clone git@github.com:celestiaorg/celestia-node.git

# Celestia node stuff
WORKDIR "/celestia-node"
RUN make install
WORKDIR "/"
CMD celestia dev start
```
Build on M1 Mac: `docker build --ssh default --platform linux/x86_64 .`
Run on M1 Mac: `docker run -it --platform linux/x86_64 <image sha>`

renaynay:
@jbowen93 would you be able to make a PR for this (under new dir `docker` with README.md and this `Dockerfile`) 🙏🏻

jbowen93:
This is resolved by https://github.com/celestiaorg/celestia-node/pull/295

PR #206, opened 2021-11-16
Add docker directory with Dockerfile and README

jbowen93:
Relates to https://github.com/celestiaorg/celestia-node/issues/69


docker/README.md
```diff
@@ -0,0 +1,13 @@
+# Dockerfile for `celestia-node`
+
+To build on a M1 Mac:
+```
+docker build --ssh default --platform linux/x86_64 .
+```
+
+To run on a M1 Mac:
```
liamsi, line 8: Is there a reason why this defaults to a M1 Mac? 

IMO, the instructions should default to Linux (even though I'm a mac user myself). Can someone with a Linux machine please test the instructions? 

Then we should change the default instructions and put the mac instructions below or a separate section.

liamsi (commented):
Dope! Thanks 👍🏼

docker/README.md
jbowen93: This is just because I have a M1 Mac and haven’t gone to the trouble of testing on something else.

mattdf:
I just ran this on my Linux laptop (i use arch btw) and it ran fine. No platform flag needed.

Docker version 20.10.10 for reference.

docker/README.md
```diff
@@ -0,0 +1,13 @@
+# Dockerfile for `celestia-node`
+
```
renaynay, line 2: We should also probably add that this is a dockerfile specifically for a `full` node. Maybe we can name the file itself `full.Dockerfile` and make a separate one for the `light` node since we'll need that for devnet.

docker/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM golang:latest
+
+RUN git clone https://github.com/celestiaorg/celestia-node.git
+
+WORKDIR "/go/celestia-node"
+RUN make install
+RUN celestia full init
+CMD celestia full start
```
Wondertan, line 8: Add a line at the end.

docker/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM golang:latest
```
Wondertan, line 1: We should somehow differentiate in the name that this file is for Full and not Light node. Eventually, we should make one for Light,

docker/README.md
Wondertan: Oh, ok I see) We think the same sometimes ❤️

libs/keystore/keystore.go
```diff
@@ -10,6 +10,7 @@ type (
 	// KeyName represents private key name.
 	KeyName string
 
+	// TODO(Josh): Should this match libp2ps PrivKey type?
```
renaynay:
```suggestion
	// TODO @jbowen93: Should this match libp2ps PrivKey type?
```

node/p2p/identity.go
```diff
@@ -2,16 +2,42 @@ package p2p
...
 	"crypto/rand"
+	"errors"
 
+	"github.com/celestiaorg/celestia-node/libs/keystore"
 	"github.com/libp2p/go-libp2p-core/crypto"
 	"github.com/libp2p/go-libp2p-core/peer"
 	"github.com/libp2p/go-libp2p-core/peerstore"
 )
 
+// TODO(Josh): What should the real name for this be?
```
renaynay, line 13: same thing re TODO structure

renaynay (commented):
small nits but otherwise LGTM!

Two things: 
1. check to make sure lint passes via `make lint`
2. we usually rebase on main to maintain cleaner commit history, but i guess it doesn't really matter much since we squash on merge anyway.

liamsi:
It would be nice if we could get a small local testnet running with docker-compose: https://github.com/celestiaorg/test-infra/issues/17
e.g. 3 validators (with full celestia nodes attached) and 1 light client

jbowen93:
> It would be nice if we could get a small local testnet running with docker-compose: [celestiaorg/test-infra#17](https://github.com/celestiaorg/test-infra/issues/17) e.g. 3 validators (with full celestia nodes attached) and 1 light client

As mentioned in https://github.com/celestiaorg/test-infra/issues/17 my preference would be to use K8s over docker compose.

Also, this PR is extremely out of date, I'm just going to close it and make a new one (probably referencing celestiaorg/test-infra#17]) when I get that working.

PR #207, opened 2021-11-16
fix(node/p2p): fix MutualPeers reading from Config

Wondertan:
A quick fix that I accidentally caught.


liamsi (approved):
👀

mattdf (approved):
nice lol, feels like one of those things that would have been annoying to track down later.

renaynay (approved):
❤️

PR #208, opened 2021-11-17
fix(node/p2p): ensure  DHT bootrapps itself on Start

### LINKED ISSUE #201: node/p2p: Bootstrapper
Create a component that ensures that the node is always connected to *all* of the hardcoded or configured bootstrap nodes.

Need this for devnet.

Wondertan:
## Background
Basically, this just ensures nodes (both light and full) are connected to bootstrap peers and their DHT table is refreshed, I was imagining something more complex, but this should be fine.

## Ref
Close #201


mattdf (approved):
LGTM

renaynay (approved):

PR #213, opened 2021-11-18
Persist generated key to keystore and reuse existing key if present

jbowen93:
identity.go previously created a new key on every run of `celestia full init` resulting in new PeerIDs each time. Persisting the keys allows us to have persistent PeerIDs.

### DIFF
--- a/libs/keystore/fs_keystore.go
+++ b/libs/keystore/fs_keystore.go
@@ -2,11 +2,14 @@ package keystore
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/fs"
 	"os"
 	"path/filepath"
 )
+
+var ErrNotFound = errors.New("keystore: key not found")
 
 // fsKeystore implements persistent Keystore over OS filesystem.
 type fsKeystore struct {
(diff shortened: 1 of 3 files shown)


liamsi (approved):
👏🏼

libs/keystore/fs_keystore.go
```diff
@@ -2,12 +2,15 @@ package keystore
...
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/fs"
 	"os"
 	"path/filepath"
 )
 
+var ErrNotFound = errors.New("keystore: key not found")
```
Wondertan, line 12: A simple comment would not hurt.

node/p2p/identity.go
```diff
@@ -2,21 +2,49 @@ package p2p
...
 
 	"github.com/libp2p/go-libp2p-core/crypto"
 	"github.com/libp2p/go-libp2p-core/peer"
 	"github.com/libp2p/go-libp2p-core/peerstore"
+
+	"github.com/celestiaorg/celestia-node/libs/keystore"
 )
 
+// TODO(Josh): What should the real name for this be?
+const keyName = "test-key"
```
Wondertan, line 15: `p2p-key` is fine or did it cause problems?

libs/keystore/map_keystore.go
```diff
@@ -35,7 +35,7 @@ func (m *mapKeystore) Get(n KeyName) (PrivKey, error) {
 
 	k, ok := m.keys[n]
 	if !ok {
-		return PrivKey{}, fmt.Errorf("keystore: key '%s' not found", n)
+		return PrivKey{}, fmt.Errorf("%w: %s", ErrNotFound, n)
```
Wondertan, line 38: Awesome! You found it.

node/p2p/identity.go
renaynay: @jbowen93 check this when u have time.

renaynay (approved):
leggo

Bidon15 (approved):

PR #216, opened 2021-11-18
service/header: Header Exchange implemented over Core Node

renaynay:
This PR will allow header service to sync headers from the core node in order to sync up to the chain. This is necessary for bootstrap peers.

### DIFF
--- a/service/header/core_exchange.go
+++ b/service/header/core_exchange.go
@@ -1,3 +1,122 @@
 package header
 
+import (
+	"context"
+	"math"
 
+	format "github.com/ipfs/go-ipld-format"
+
+	"github.com/tendermint/tendermint/pkg/da"
+	"github.com/tendermint/tendermint/pkg/wrapper"
+	"github.com/tendermint/tendermint/types"
+
+	"github.com/celestiaorg/celestia-node/core"
+	"github.com/celestiaorg/celestia-node/ipld"
+	"github.com/celestiaorg/nmt"
+	"github.com/celestiaorg/rsmt2d"
+)
+
+type CoreExchange struct {
+	fetcher    *core.BlockFetcher
+	shareStore format.DAGService
+}
+
+func NewCoreExchange(fetcher *core.BlockFetcher, dag format.DAGService) *CoreExchange {
+	return &CoreExchange{
+		fetcher:    fetcher,
+		shareStore: dag,
+	}
+}
+
+func (ce *CoreExchange) RequestHeader(ctx context.Context, height uint64) (*ExtendedHeader, error) {
+	intHeight := int64(height) + 1
+	block, err := ce.fetcher.GetBlock(ctx, &intHeight)
+	if err != nil {
+		return nil, err
+	}
+	return ce.generateExtendedHeaderFromBlock(block)
+}
+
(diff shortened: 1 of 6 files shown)


service/header/core_exchange.go
```diff
@@ -0,0 +1,122 @@
...
+func (ce *CoreExchange) RequestHeader(ctx context.Context, height uint64) (*ExtendedHeader, error) {
+	intHeight := int64(height) + 1
+	block, err := ce.fetcher.GetBlock(ctx, &intHeight)
+	if err != nil {
+		return nil, err
+	}
+	return ce.generateExtendedHeaderFromBlock(block)
+}
+
+func (ce *CoreExchange) RequestHeaders(ctx context.Context, origin, amount uint64) ([]*ExtendedHeader, error) {
```
Wondertan, line 40: This will be so slooow, but it is fine for now. Just for reference, we should use BlockSearch API instead

Wondertan (approved):

liamsi (approved):

service/header/core_exchange.go
```diff
@@ -0,0 +1,122 @@
...
+
+func NewCoreExchange(fetcher *core.BlockFetcher, dag format.DAGService) *CoreExchange {
+	return &CoreExchange{
+		fetcher:    fetcher,
+		shareStore: dag,
+	}
+}
+
+func (ce *CoreExchange) RequestHeader(ctx context.Context, height uint64) (*ExtendedHeader, error) {
+	intHeight := int64(height) + 1
```
adlerjohn, line 32: Are we sure that passing `math.MaxUint64` won't result in some weird behavior in the node?

liamsi: MaxInt64 would already overflow here. Maybe just add a TODO to make sure that conversion is safe.

adlerjohn: Yeah a TODO should suffice. If we're not careful, someone will be able to crash a node by simply asking for a block at a large height, for example, which is pretty bad.

Wondertan: That is like saying: Are we sure that passing `nil` to `reader.Read` won't result in some weird behavior in the node?

I mean this is like an unreal case that should be possible until we reach such height.

adlerjohn: In this particular case, requesting headers is done by an untrusted peer! So you probably don't want that to be able to _crash_ your node. If it were something local and trusted, then it'd be a non-issue. Or is this function only callable internally?

liamsi: It is talking to a trusted core node indeed.

Wondertan: The reason is that we count blocks from `0`, while tendermint from 1

Wondertan: Btw, +1 there is because we count blocks from 0 while Tendermint from 1. What should we stick to?

Wondertan: > In this particular case, requesting headers is done by an untrusted peer! So you probably don't want that to be able to crash your node. If it were something local and trusted, then it'd be a non-issue. Or is this function only callable internally?

Ah, I got your point now. Well, the CoreExchange is not going to be part of a publicly facing API. Though, there might be a case where this happens.

PR #211, opened 2021-11-18
service/header, cmd: Inject head by trusted header hash

renaynay:
TODO: 

- [ ] resolve broken TestFullLifecycle

Issue #169, opened 2021-11-04
[EPIC] Public API / RPC Tracking Issue


Wondertan:
## Current State
Right now, we have implemented an "RPC" server in #124 which can be started /stopped and provides a way to register arbitrary __HTTP__ handlers. Also, we have an unmerged #135 that adds the first message for the "RPC" server - StatusMessage.

## Problems
* As you may notice, I put "RPC" in quotes and the reason is that the current implementation is not an RPC server, but a plain HTTP server using std implementation. 
* Current architecture/vision for Services registering custom HTTP Handlers is prone to cyclic dependencies.
	* As noted [here](https://github.com/celestiaorg/celestia-node/pull/135#discussion_r735408661), `RegisterHandlers` was planned to be extended, so Services could register their own handlers. This is impossible because Services would need to have a pointer to the Node structure, while Node itself contains pointers to all the services. Unfortunately, this wouldn't compile due to mentioned cyclic dependency, which can not be avoided without hacky solutions.
* Simple HTTP server does not provide capabilities of the real RPC server. Even though RPC may and usually runs over HTTP, RPC provides a client/server framework and a protocol to remotely call arbitrary handlers. Using simple HTTP server requires us to:
	* Implement client code separately.
	* Implement a custom framework to work over websockets.
	* Come up with a custom authorization solution.
	* Basically, reinvent multiple wheels.

## Solution
> NOTE: It is not detailed due to current priorities and according to the current devnet attack plan, we won't need an RPC server, as https://github.com/celestiaorg/dalc is going to implement its own GRPC server.

The main idea is to use existing RPC implementation without reimplementing the wheel:
* https://github.com/etclabscore/go-openrpc-reflect
* https://github.com/filecoin-project/go-jsonrpc
* https://github.com/grpc/grpc-go
	* God pls no
* https://github.com/sourcegraph/jsonrpc2

> More details to come once this starts to be a priority again

## Action Items
- [x] #255
- [ ] #256
- [x] #257
- [x] #258
- [x] #259

renaynay:
SGTM - let's use filecoin impl.

renaynay:
@Wondertan let's use this issue as a main issue tracking for RPC in general and then break it down into smaller issues from here.

liamsi:
> God pls no

What is so bad about grpc? 

> SGTM - let's use filecoin impl.

What is the rationale behind this and where can I learn more about filecoin's rpc implementation? There isn't even a Readme in the repo.

[...]

PR #220, opened 2021-11-19
Pre Genesis: The PR to make testnet happen!

Wondertan:
Sorry for this huge PR. Really, don't have time to split into pieces. However, you can look at the history of commits as they are logically separated(the second major part at least) and hope they will make sense to you.


node/full.go
```diff
@@ -1,26 +0,0 @@
...
-		lightComponents(cfg, repo),
-		fxutil.ProvideIf(!cfg.Core.Remote, repo.Core), // provide core repo constructor only in embedded mode.
-		nodecore.Components(cfg.Core),
-		fx.Provide(func(client core.Client) block.Fetcher {
-			return core.NewBlockFetcher(client)
-		}),
-		fx.Provide(rpc.NewServer),
-		fx.Provide(services.Block),
-	)
-}
```
liamsi, line 26: Are these not necessary anymore?

service/header/interface.go
```diff
@@ -33,10 +33,6 @@ type Broadcaster interface {
 // Exchange encompasses the behavior necessary to request ExtendedHeaders
 // from the network.
 type Exchange interface {
-	// Start starts Exchange.
-	Start(context.Context) error
-	// Stop stops Exchange.
-	Stop(context.Context) error
```
liamsi: Are these moved to another interface or just left in the implementation only?

service/header/service_test.go
```diff
@@ -1,42 +0,0 @@
...
-
-	serv := NewHeaderService(ex, emptyStore, ps, expectedHeader.Hash())
-
-	err = serv.Start(context.Background())
-	require.NoError(t, err)
-	// check that head is properly injected
-	head, err := emptyStore.Head(context.Background())
-	require.NoError(t, err)
-	assert.Equal(t, expectedHeader.Height, head.Height)
-}
```
liamsi: Why do we delete the test?

service/header/sync.go
```diff
@@ -57,37 +61,8 @@ func (s *syncer) Sync(ctx context.Context) {
...
-		err = s.store.Append(ctx, headers...)
-		if err != nil {
-			return err
-		}
-
-		start += amount
-	}
-
-	return s.store.Append(ctx, newHead)
-}
```
liamsi, line 87: ~~This is not necessary anymore because we assume that this will always start from genesis now?~~ NVM

service/header/sync.go
```diff
@@ -118,3 +93,56 @@ func (s *syncer) validator(ctx context.Context, p peer.ID, msg *pubsub.Message)
...
+
+	return head, err
+}
+
+// TODO(@Wondertan): Number of headers that can be requested at once. Either make this configurable or,
+// find a proper rationale for constant.
+var requestSize uint64 = 128
+
+// syncDiff requests headers from knownHead up to new head.
+func (s *Syncer) syncDiff(ctx context.Context, knownHead, newHead *ExtendedHeader) error {
```
liamsi, line 126: Ah just moved `syncDiff`.

node/services/config.go
```diff
@@ -1 +1,44 @@
...
+
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/multiformats/go-multiaddr"
+	tmbytes "github.com/tendermint/tendermint/libs/bytes"
+)
+
+type Config struct {
+	// GenesisHash is the Block/Header hash that Nodes use as starting point for header synchronization.
+	// Only affects the node once on initial sync.
+	GenesisHash string
```
liamsi, line 14: Q: Can this be any trusted header hash that the user chooses or does it have to be from genesis? Because the var name makes it sound like this is the genesis hash but the comment just speaks of it as a subjective starting point?

node/full_test.go
```diff
@@ -105,39 +97,6 @@ func TestFull_P2P_Streams(t *testing.T) {
...
-	t.Cleanup(func() {
-		resp.Body.Close()
-	})
-	assert.Equal(t, "pong", string(pong))
-
-	ctx, cancel = context.WithCancel(context.Background())
-	t.Cleanup(cancel)
-
-	require.NoError(t, node.Stop(ctx))
-}
```
liamsi, line 133: Another deleted test? Why? I understand Start/Stop where removed but it would still be good to test that the RPC server runs as expected. Though that could be more of an integration test.

cmd/start.go
```diff
@@ -63,17 +61,17 @@ func Start(repoName string, tp node.Type) *cobra.Command {
...
+				opts = append(opts, node.WithGenesis(genesisHash))
 			}
 
 			return start(cmd, tp, repo, opts...)
 		},
 	}
 
 	cmd.Flags().String(cfgAddress, "", "Indicates node to connect to the given remote core node")
-	cmd.Flags().String(headHash, "", "known header hash with which to initialize the node")
+	cmd.Flags().String(genesis, "", "Hex encoded block hash. Starting point for header synchronization.")
```
liamsi, line 74: this should also be clearer: Is it the genesis or some starting point?

liamsi (commented):
Reviewed commit for commit. Good job on the commit comments and splitting it up this way. 
Left a few questions. Looks good!

node/full.go
Wondertan: Block is still needed, but not the RPC rn, it is not used and needs to be reworked anyway.

service/header/interface.go
Wondertan: In the implementation only

service/header/service_test.go
Wondertan: Because there is another test that has this case covered

service/header/sync.go
Wondertan: K

service/header/sync.go
Wondertan: Yep

node/services/config.go
Wondertan: It's like something in between. It can only be used one time as genesis, but it can go from any height and not specifically 0. Not sure how to name it.

node/full_test.go
Wondertan: We don't use an RPC anywhere. I removed it for now. And it is not an RPC, but just a plain HTTP server. 
#169

cmd/start.go
Wondertan: Let's think of how to name it. It can be genesis and also be just a random height to start syncing from. So technically, it would be even possible for FullNode without a core to sync the chain from height 2 or 2000. This is not a FullNode mb already but almost, because of the border between Full/Partial Storage Node.

Wondertan: `Head` is also a bad name.

renaynay (dismissed):
Approved as #221 is based on this PR and already went through that.

node/services/config.go
liamsi: `TrustedHeaderHash`?

cmd/start.go
liamsi: Suggestions:
- trusted header hash
- checkpoint hash

node/services/config.go
adlerjohn: Second `TrustedHeaderHash`.

adlerjohn:
Is this still under active development or ready for review?

liamsi:
These are effectively 3 PRs I think. this contains the changes of:
- https://github.com/celestiaorg/celestia-node/pull/211 
- then #220 itself contains fixes for #211
- then all sorts of different smaller fixes and refactorings

IMO, this should be split up into at least 2 if not 3 PRs again and merged separately (on Monday and before devnet ideally).

node/services/config.go
Wondertan: Changing this to TrustedHeaderHash

cmd/start.go
Wondertan: trusted header hash

Wondertan:
Closing in favor of #221

PR #224, opened 2021-11-19
service/header: Verify hash matches header retrieved in `RequestByHash()`

### LINKED ISSUE #219: service/header: Exchange should verify that requested hash matches the header
Currently, we don't verify a received Header against the hash used for requesting. We should fix this security breach.

renaynay:
Resolves #219

TODO 
- [x] tests
- [x] implement same check for p2p ex

### DIFF
--- a/service/header/core_exchange.go
+++ b/service/header/core_exchange.go
@@ -1,7 +1,9 @@
 package header
 
 import (
+	"bytes"
 	"context"
+	"fmt"
 	"math"
 
 	format "github.com/ipfs/go-ipld-format"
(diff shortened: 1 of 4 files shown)


liamsi (dismissed):
Looks correct. Will the TODOs in the opening comment be resolved here or in follow up PRs?

liamsi (dismissed):
image

adlerjohn (approved):
ahhh I'm approoooving!

service/header/core_exchange.go
```diff
@@ -56,7 +58,19 @@ func (ce *CoreExchange) RequestByHash(ctx context.Context, hash []byte) (*Extend
...
-	return ce.generateExtendedHeaderFromBlock(block)
+	extHeader, err := ce.generateExtendedHeaderFromBlock(block)
+	if err != nil {
+		return nil, err
+	}
+	// verify hashes match
+	if !hashMatch(hash, extHeader.Hash().Bytes()) {
+		return nil, fmt.Errorf("incorrect hash in header: expected %x, got %x", hash, extHeader.Hash().Bytes())
+	}
+	return extHeader, nil
```
Wondertan, line 69: We also need to have a ValidateBasic on it

Wondertan: So we can see other fields are matching

service/header/exchange.go
```diff
@@ -191,6 +191,9 @@ func (ex *exchange) RequestByHash(ctx context.Context, hash []byte) (*ExtendedHe
 	if err != nil {
 		return nil, err
 	}
+	if !hashMatch(headers[0].Hash().Bytes(), hash) {
+		return nil, fmt.Errorf("incorrect hash in header: expected %x, got %x", hash, headers[0].Hash().Bytes())
```
Wondertan, line 195:
```suggestion
		return nil, fmt.Errorf("header hash mismatch: expected %x, got %x", hash, headers[0].Hash().Bytes())
```

liamsi (dismissed):
Still approoooving anyways!

Wondertan (dismissed):

liamsi (approved):

Wondertan (approved):

adlerjohn (approved):

PR #229, opened 2021-11-23
cmd, service/header: rename `genesis` to `trustedHash` to avoid confusion, extract flags to common var for re-use

renaynay:
This PR:
* renames `genesis` to `trustedHash` in the codebase to avoid confusion
* extracts flags to common variable for re-use between commands `start` and `init`

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -1,0 +1,40 @@
+package cmd
+
+import "flag"
+
+var configFlags = []*flag.Flag{
+	loglevelFlag,
+	nodeConfigFlag,
+	trustedHashFlag,
+	trustedPeerFlag,
+	coreRemoteFlag,
+}
+
+var (
+	loglevelFlag = &flag.Flag{
+		Name:     "log.level",
+		Usage:    "DEBUG, INFO, WARN, ERROR, DPANIC, PANIC, FATAL, and\n// their lower-case forms",
+		DefValue: "info",
+	}
+	nodeConfigFlag = &flag.Flag{
+		Name:     "node.config",
+		Usage:    "Path to a customized Config",
+		DefValue: "",
+	}
+	trustedHashFlag = &flag.Flag{
+		Name:     "headers.trusted-hash",
+		Usage:    "Hex encoded block hash. Starting point for header synchronization",
+		DefValue: "",
+	}
+	trustedPeerFlag = &flag.Flag{
+		Name:     "headers.trusted-peer",
+		Usage:    "Multiaddr of a reliable peer to fetch headers from",
+		DefValue: "",
+	}
+	coreRemoteFlag = &flag.Flag{
+		Name: "core.remote",
+		Usage: "Indicates node to connect to the given remote core node. " +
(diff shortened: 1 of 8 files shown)


node/services/config.go
```diff
@@ -9,19 +9,19 @@ import (
 )
 
 type Config struct {
-	// GenesisHash is the Block/Header hash that Nodes use as starting point for header synchronization.
+	// TrustedHash isg the Block/Header hash that Nodes use as starting point for header synchronization.
```
adlerjohn, line 12:
```suggestion
	// TrustedHash is the Block/Header hash that Nodes use as starting point for header synchronization.
```

adlerjohn (approved):

cmd/flags.go
```diff
@@ -0,0 +1,40 @@
...
+	trustedHashFlag,
+	trustedPeerFlag,
+	coreRemoteFlag,
+}
+
+var (
+	loglevelFlag = &flag.Flag{
+		Name:     "log.level",
+		Usage:    "DEBUG, INFO, WARN, ERROR, DPANIC, PANIC, FATAL, and\n// their lower-case forms",
+		DefValue: "info",
```
Wondertan, line 17: We should stop turning logging by default IMO and make empty DefValue. Can you pls include this in this PR?

cmd/start.go
```diff
@@ -42,7 +35,7 @@ func Start(repoName string, tp node.Type) *cobra.Command {
 		Args:         cobra.NoArgs,
 		SilenceUsage: true,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			level, err := logging.LevelFromString(cmd.Flag(loglevel).Value.String())
+			level, err := logging.LevelFromString(cmd.Flag(loglevelFlag.Name).Value.String())
```
Wondertan, line 38: This code also defaults to INFO by default. To disable this, we should run `logging.LevelFromString` only `if cmd.Flag(loglevelFlag.Name).Value.String() != ""`. Similar to other flags.

Wondertan (commented):
Awesome! Thank you for deduplication

cmd/flags.go
renaynay: wait why? Log info level should always be default?

Wondertan: > Log info level should always be default?

Exactly, why? IMHO logging is mostly for power users. I don't have a strong opinion though on such minor things for now, but for light clients, I would remove logging by default.

renaynay: No I think that's not good. I don't know any other client software that disables logging by default. If the user doesn't want to see the logs, they can run it as a background proc.

Wondertan: Well, how often do you encounter logs while working with any software clients daily? Anyway, I think keeping logging by default is good for now, as things will break and triers will be able to report logs.

Wondertan: We also should pipe those to some file by default mb

Wondertan: https://github.com/celestiaorg/celestia-node/issues/245

Wondertan (approved):

PR #233, opened 2021-11-23
fix: store do not store the head we already have

Wondertan:


renaynay (approved):
Quick fix -- we need to figure out ExtendedHeaderSub message deduplication.

service/header/store.go
```diff
@@ -192,7 +195,7 @@ func (s *store) Append(ctx context.Context, headers ...*ExtendedHeader) error {
 		verified, head = append(verified, h), h
 	}
 	if len(verified) == 0 {
```
renaynay, line 197:
```suggestion
	if len(verified) == 0 {
	log.Debug("header/store: no valid headers were given")
```

adlerjohn (approved):

renaynay (approved):

PR #240, opened 2021-11-23
Mainly remove misleading error on node stop

### LINKED ISSUE #228: node: Implement graceful shutdown
Related to #82.

Reference [this article](https://rudderstack.com/blog/implementing-graceful-shutdown-in-go/) for new ideas.

Wondertan:
Closes #228

### DIFF
--- a/service/header/subscription.go
+++ b/service/header/subscription.go
@@ -37,7 +37,7 @@ func (s *subscription) NextHeader(ctx context.Context) (*ExtendedHeader, error) 
 	var header ExtendedHeader
 	err = header.UnmarshalBinary(msg.Data)
 	if err != nil {
-		log.Errorw("unmarshalling data from message", "err")
+		log.Errorw("unmarshalling data from message", "err", err)
 		return nil, err
 	}
 
--- a/service/header/subscription.go
+++ b/service/header/subscription.go
@@ -30,7 +30,6 @@ func newSubscription(topic *pubsub.Topic) (*subscription, error) {
 func (s *subscription) NextHeader(ctx context.Context) (*ExtendedHeader, error) {
 	msg, err := s.subscription.Next(ctx)
 	if err != nil {
-		log.Errorw("reading next message from subscription", "err", err.Error())
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


service/header/subscription.go
```diff
@@ -30,15 +30,14 @@ func newSubscription(topic *pubsub.Topic) (*subscription, error) {
 func (s *subscription) NextHeader(ctx context.Context) (*ExtendedHeader, error) {
 	msg, err := s.subscription.Next(ctx)
 	if err != nil {
-		log.Errorw("reading next message from subscription", "err", err.Error())
```
Wondertan, line 33: Note for @renaynay. There is no value in logging the error here as the only types of errors we will see here are context-related. And users of NextHeader should handle them accordingly as we do in DASer.

service/header/subscription.go
```diff
@@ -30,15 +30,14 @@ func newSubscription(topic *pubsub.Topic) (*subscription, error) {
...
 		return nil, err
 	}
 	log.Debugw("received message", "topic", msg.Message.GetTopic(), "sender", msg.ReceivedFrom)
 
 	var header ExtendedHeader
 	err = header.UnmarshalBinary(msg.Data)
 	if err != nil {
-		log.Errorw("unmarshaling data from message", "err", err.Error())
+		log.Errorw("unmarshalling data from message", "err")
```
adlerjohn, line 40: Is there a reason to remove `err.Error()` from the string? Now the string is just "unmarshalling data from message err" all the time, no?

renaynay: Yeah, i don't understand why to remove `unmarshaling` as well as the error message?

Wondertan: Oh, ok. Thank you for catching this. I wanted to only remove `.Error()` part.

Wondertan: Fixed

renaynay (approved):

service/header/subscription.go
Wondertan: @adlerjohn, resolved and approved?

adlerjohn (approved):

PR #241, opened 2021-11-24
share: handle rows concurrently

### LINKED ISSUE #184: service/share: Implement parallelisation for `GetSharesByNamespace`
Implement parallelisation for `GetSharesByNamespace` when there are multiple rows that contain the given `namespace.ID`.

vgonkivs:
* Handle multiple rows concurrently;

Closes #184

### DIFF
--- a/service/share/share.go
+++ b/service/share/share.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"math/rand"
-	"sync"
 
 	"github.com/ipfs/go-cid"
 	format "github.com/ipfs/go-ipld-format"
@@ -145,51 +144,35 @@ func (s *service) GetSharesByNamespace(ctx context.Context, root *Root, nID name
 		return nil, ipld.ErrNotFoundInRange
 	}
 
-	namespacedShares := make([]Share, 0)
-	nodeCh := make(chan []format.Node)
-	errCh := make(chan error)
-	wg := sync.WaitGroup{}
-	newCtx, cancel := context.WithCancel(ctx)
+	type res struct {
+		nodes []format.Node
+		err   error
+	}
+	resultCh := make(chan *res)
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
 
 	for _, rootCID := range rowRootCIDs {
-		wg.Add(1)
 		go func(rootCID cid.Cid) {
-			nodes, err := ipld.GetLeavesByNamespace(newCtx, s.dag, rootCID, nID)
-			if err != nil {
-				errCh <- err
-				return
-			}
-			nodeCh <- nodes
+			nodes, err := ipld.GetLeavesByNamespace(ctx, s.dag, rootCID, nID)
+			resultCh <- &res{nodes: nodes, err: err}
 		}(rootCID)
 	}
 
-	go func() {
-		wg.Wait()
-		close(nodeCh)
-		close(errCh)
-		_ = cancel
-	}()
-
-	go func() {
-		for nodes := range nodeCh {
-			for _, node := range nodes {
+	namespacedShares := make([]Share, 0)
+	for i := 0; i < len(rowRootCIDs); i++ {
+		select {
+		case result := <-resultCh:
+			if result.err != nil {
+				return nil, result.err
+			}
+			for _, node := range result.nodes {
 				namespacedShares = append(namespacedShares, node.RawData()[1:])
 			}
-			wg.Done()
+		case <-ctx.Done():
+			return nil, ctx.Err()
 		}
-	}()
-
-	var fetchErr error
-	for err := range errCh {
-		if fetchErr == nil {
-			fetchErr = err
-			cancel()
-		}
-		wg.Done()
-	}
-
-	if fetchErr != nil {
-		return nil, fetchErr
 	}
 
 	return namespacedShares, nil
--- a/service/share/share.go
+++ b/service/share/share.go
@@ -149,11 +149,12 @@ func (s *service) GetSharesByNamespace(ctx context.Context, root *Root, nID name
 	nodeCh := make(chan []format.Node)
 	errCh := make(chan error)
 	wg := sync.WaitGroup{}
+	newCtx, cancel := context.WithCancel(ctx)
 
 	for _, rootCID := range rowRootCIDs {
 		wg.Add(1)
 		go func(rootCID cid.Cid) {
-			nodes, err := ipld.GetLeavesByNamespace(ctx, s.dag, rootCID, nID)
+			nodes, err := ipld.GetLeavesByNamespace(newCtx, s.dag, rootCID, nID)
 			if err != nil {
 				errCh <- err
 				return
@@ -166,6 +167,7 @@ func (s *service) GetSharesByNamespace(ctx context.Context, root *Root, nID name
 		wg.Wait()
 		close(nodeCh)
 		close(errCh)
+		_ = cancel
 	}()
 
 	go func() {
@@ -179,8 +181,9 @@ func (s *service) GetSharesByNamespace(ctx context.Context, root *Root, nID name
 
 	var fetchErr error
 	for err := range errCh {
-		if fetchErr != nil {
+		if fetchErr == nil {
 			fetchErr = err
+			cancel()
 		}
 		wg.Done()
 	}
--- a/service/share/share.go
+++ b/service/share/share.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"math/rand"
+	"sync"
 
 	"github.com/ipfs/go-cid"
 	format "github.com/ipfs/go-ipld-format"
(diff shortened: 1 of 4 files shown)


service/share/share.go
```diff
@@ -134,18 +135,24 @@ func (s *service) GetShares(context.Context, *Root) ([][]Share, error) {
...
-	rowRootCIDs := make([]cid.Cid, 0)
+	rowRootCIDs := make(chan cid.Cid, len(root.RowsRoots))
+	wg := sync.WaitGroup{}
 	for _, row := range root.RowsRoots {
 		if !nID.Less(nmt.MinNamespace(row, nID.Size())) && nID.LessOrEqual(nmt.MaxNamespace(row, nID.Size())) {
-			rowRootCIDs = append(rowRootCIDs, plugin.MustCidFromNamespacedSha256(row))
+			wg.Add(1)
+			go func(row []byte) {
+				rowRootCIDs <- plugin.MustCidFromNamespacedSha256(row)
+			}(row)
```
Wondertan, line 145: Pls look at `plugin.MustCidFromNamespacedSha256`. This is a cheap operation that does not need a parallelization. In fact, new code will be even slower and use more RAM due to goroutines overhead for an operation that can be done sequentially fast enough. Instead, you should run each `ipld.GetLeavesByNamaspace` in a separate routine, as it is a blocking operation that involves a lot of blocking IO.

Wondertan (changes requested):
@vgonkivs, thank you for working on this. Unfortunately, this is wrong and I'll try to explain why.

service/share/share.go
vgonkivs: Thanks for explanations, @Wondertan . Sorry, for not getting what exactly should be done from the first attempt. May I use the same approach for `ipld.GetLeavesByNamaspace`(wg+channels) ?

service/share/share.go
```diff
@@ -145,16 +146,40 @@ func (s *service) GetSharesByNamespace(ctx context.Context, root *Root, nID name
...
+				namespacedShares = append(namespacedShares, node.RawData()[1:])
+			}
+			wg.Done()
 		}
+	}()
+
+	for err := range errCh {
+		return nil, err
 	}
+
```
Wondertan, line 182: By quick look, there is one bug in here and maybe more: if `ipld.GetLeavesByNamespace` returns an error, `wg.Wait` will never unblock(line 165) and both `errCh` and `nodeCh` won't be closed.

Wondertan (changes requested):

service/share/share.go
vgonkivs: Yeah, unfortunately It has one more issue that I haven`t noticed. 
Regarding line 165 I believed that it has to be cleared by GC but it`s not. As a fallback solution I`d like to propose to call `wg.Done` also at line 180 and not return the err but storing it to local variable and return the error. So channels will be closed in this case. The minus is that we have to perform all requests to `ipld.GetLeavesByNamespace` even if error occurred.

renaynay:
@vgonkivs Thank you very much for this PR - would you be able to take a first shot at writing a test for this?

service/share/share.go
```diff
@@ -145,16 +146,49 @@ func (s *service) GetSharesByNamespace(ctx context.Context, root *Root, nID name
...
+			}
+			wg.Done()
 		}
+	}()
 
-		for _, node := range nodes {
-			namespacedShares = append(namespacedShares, node.RawData()[1:])
+	var fetchErr error
+	for err := range errCh {
+		if fetchErr != nil {
```
liamsi, line 182: 🤔 Are you sure this line makes sense?

liamsi (changes requested):
Please revisit this: https://github.com/celestiaorg/celestia-node/pull/241/files#r760551422

I think it would be good to add a benchmark that shows that this parallelization is justified (though it is relatively obvious that it should be faster but still).

service/share/share.go
vgonkivs: Hi @liamsi , yes , you right. The condiiton should be `if fetchErr == nil` to handle the first occurred error. But I`m thinking, do we need here to handle the first error in case of multiple errors will be occurred? Or we can return the last one and in this case we do not need this condition.

liamsi: Good question. I think we could (as before) return on the first caught error. There is no need to continue fetching after an error occurred.

vgonkivs: Added new context to cancel all requests if error occurred.

vgonkivs:
Hi @renaynay , @liamsi . Thanks for the idea to add benchmarks and additional tests. I will try to provide them within next week.

Wondertan:
> I think it would be good to add a benchmark that shows that this parallelization is justified (though it is relatively obvious that it should be faster but still).

@liamsi, the optimization is not CPU bound, i.e. we don't parallelize something calculation intensive, but we parallelize IO bound logic, so parallelization is furthermore a wrong word here, while concurrency is a right one(like it is said in PR name). Thus, we might not see any performance improvements and even performance degradation when we mock out network responses. 

Particularly, here we start traversing and requesting each row concurrently, instead of doing them in order. Imagine sending 10 network requests in order and compare this to spawning 10 concurrent goroutines. In order requires for n+1 request to wait for n response to land first, while sending concurrently allows requests to be sent independently. In the mock environment, where responses are immediate, goroutines may only add overhead, as there are no responses to wait for.

vgonkivs:
Thanks for the explanation @Wondertan. I was a bit confused after I had received the results - the code really works slower on mock.
@liamsi , I have update PR with additional test and benchmark.

service/share/share.go
```diff
@@ -145,16 +146,52 @@ func (s *service) GetSharesByNamespace(ctx context.Context, root *Root, nID name
...
+			}
+			nodeCh <- nodes
+		}(rootCID)
+	}
+
+	go func() {
+		wg.Wait()
+		close(nodeCh)
+		close(errCh)
+		_ = cancel
```
Wondertan, line 170: ?

cancel is a func that should be called. Here you just reassign the func value to nothing.

service/share/share.go
```diff
@@ -145,16 +146,52 @@ func (s *service) GetSharesByNamespace(ctx context.Context, root *Root, nID name
 	}
 
 	namespacedShares := make([]Share, 0)
+	nodeCh := make(chan []format.Node)
+	errCh := make(chan error)
```
Wondertan, line 150: The code can be simplified by making an anonymous `result struct` that has both the value and the error, so only one channel of the type needs to be created and handled. That's a common pattern in such situations.

service/share/share.go
```diff
@@ -145,16 +146,52 @@ func (s *service) GetSharesByNamespace(ctx context.Context, root *Root, nID name
...
+	wg := sync.WaitGroup{}
+	newCtx, cancel := context.WithCancel(ctx)
+
```
Wondertan, line 153: It's ok to have multiple defers in a func and it is a good practice to keep each defer next to the value a defer is closing/canceling/flushing or etc.

Wondertan (changes requested):

service/share/share.go
vgonkivs: done

service/share/share.go
vgonkivs: removed

service/share/share.go
vgonkivs: reworked

Wondertan (approved):
👍🏻

adlerjohn:
Waiting on review from @liamsi

service/share/share.go
```diff
@@ -144,17 +144,37 @@ func (s *service) GetSharesByNamespace(ctx context.Context, root *Root, nID name
...
+			resultCh <- &res{nodes: nodes, err: err}
+		}(rootCID)
+	}
 
-		for _, node := range nodes {
-			namespacedShares = append(namespacedShares, node.RawData()[1:])
+	namespacedShares := make([]Share, 0)
+	for i := 0; i < len(rowRootCIDs); i++ {
+		select {
+		case result := <-resultCh:
```
renaynay, line 166: Should we check for `ok` as well / make sure`result` is never nil? It could technically panic otherwise, no?

service/share/share_test.go
```diff
@@ -56,3 +56,36 @@ func TestService_GetSharesByNamespace(t *testing.T) {
...
+func TestService_GetSharesByNamespaceNotFoundInRange(t *testing.T) {
+	serv, root := RandServiceWithSquare(t, 1)
+	root.RowsRoots = nil
+
+	shares, err := serv.GetSharesByNamespace(context.Background(), root, []byte(""))
+	assert.Len(t, shares, 0)
+	require.Error(t, err, "namespaceID not found in range")
+}
+
+func BenchmarkService_GetSharesByNamespace(b *testing.B) {
```
renaynay, line 69: Is there also a test to check for the correctness of the shares that are returned via this process? If not, can you please add it?

service/share/share.go
Wondertan: `resultCh` is never closed, so that is not an issue

service/share/share_test.go
Wondertan: WDYM by "correctness"?

service/share/share.go
renaynay: and there's no instance where result could ever be nil?

service/share/share_test.go
renaynay: `TestService_GetSharesByNamespace` should be modified so that if `expectedShares > 1`, then check both shares[0] and shares[1]. 

We should also expand testing around this method to cover more cases (like overflow of NID into multiple rows, not just 2)

service/share/share.go
Wondertan: Yes, you can see it from the code.

service/share/share_test.go
Wondertan: Should this be part of this PR?

renaynay: At least the first point, yes bc it's testing for accuracy of *all* shares returned by `GetShares`

Wondertan: @vgonkivs

vgonkivs: @renaynay, I hope I understand you correctly.
Added loop to `TestService_GetSharesByNamespace` to check all returned shares.

renaynay: Perfect.

renaynay (approved):
thanks so much for this @vgonkivs 

The tests are failing on a flakey test, which we have documented here: https://github.com/celestiaorg/celestia-node/issues/173

Wondertan (approved):

liamsi (approved):
Looks good! Thank you @vgonkivs 🙏🏼 

Also thanks @renaynay and @Wondertan for the reviews.

Issue #64, opened 2021-09-16
Add dependabot.yml to keeps deps up to data


liamsi:
This is not urgent and could even happen after devnet.

See for example https://github.com/celestiaorg/optimint/pull/105

Wondertan:
@Bidon15, @jbowen93, We need this already. Can one of you take over this? It is simple as just copying the config from the core.

PR #248, opened 2021-11-29
Copy dependabot.yml from celestiaorg/celestia-core

jbowen93:
Resolves https://github.com/celestiaorg/celestia-node/issues/64

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,27 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: "/"
+    schedule:
+      interval: daily
+      time: "11:00"
+    open-pull-requests-limit: 10
+  - package-ecosystem: npm
+    directory: "/docs"
+    schedule:
+      interval: daily
+      time: "11:00"
+    open-pull-requests-limit: 10
+    reviewers:
+      - fadeev
+  - package-ecosystem: gomod
+    directory: "/"
+    schedule:
+      interval: daily
+      time: "11:00"
+    open-pull-requests-limit: 10
+    reviewers:
+      - melekes
+      - tessr
+    labels:
+      - T:dependencies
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,27 @@
...
+      - fadeev
+  - package-ecosystem: gomod
+    directory: "/"
+    schedule:
+      interval: daily
+      time: "11:00"
+    open-pull-requests-limit: 10
+    reviewers:
+      - melekes
+      - tessr
```
renaynay, line 25:
```suggestion
      - Wondertan
      - renaynay
```

renaynay (changes requested):

.github/dependabot.yml
liamsi: I think the CODEOWNERS overwrites this anyways but yeah doesn't make sense and we should update it.

.github/dependabot.yml
```diff
@@ -0,0 +1,28 @@
...
+    directory: "/"
+    schedule:
+      interval: daily
+      time: "11:00"
+    open-pull-requests-limit: 10
+    reviewers:
+      - Wondertan
+      - renaynay
+    labels:
+      - T:dependencies
```
Wondertan, line 28: Pls new line her

Wondertan (changes requested):
Can we make the schedule weekly pls? I don't want to receive daily notifications pls

.github/dependabot.yml
```diff
@@ -3,13 +3,15 @@ updates:
...
     open-pull-requests-limit: 10
   - package-ecosystem: npm
     directory: "/docs"
     schedule:
-      interval: daily
+      interval: weekly
+      day: monday
       time: "11:00"
     open-pull-requests-limit: 10
     reviewers:
```
Wondertan: One last thing, we don't need a npm

Wondertan (approved):

renaynay (approved):

Issue #250, opened 2021-11-29
node: Clean up node types


Bidon15:
### Summary

After devnet, we need to do hygiene work in our node types

### Gain

This is beneficial for our upcoming tool that will be creating in-memory network and deploy nodes to test against

### Evidence

Code-base node part

### Urgency

Medium

PR #261, opened 2021-12-03
fix(cmd): fix repository path handling for Init and Start commands

Wondertan:
Fixes bug introduced in #229, without the fix `start` does not work.

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -1,6 +1,8 @@
 package cmd
 
 import (
+	"fmt"
+
 	"github.com/spf13/cobra"
 
 	"github.com/celestiaorg/celestia-node/node"
@@ -8,11 +10,11 @@ import (
 
 // Init constructs a CLI command to initialize Celestia Node of the given type 'tp'.
 // It is meant to be used a subcommand and also receive persistent flag name for repository path.
-func Init(repoName string, tp node.Type) *cobra.Command {
+func Init(repoFlagName string, tp node.Type) *cobra.Command {
 	if !tp.IsValid() {
 		panic("cmd: Init: invalid Node Type")
 	}
-	if len(repoName) == 0 {
+	if len(repoFlagName) == 0 {
 		panic("parent command must specify a persistent flag name for repository path")
 	}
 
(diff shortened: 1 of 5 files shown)


cmd/init.go
```diff
@@ -1,18 +1,20 @@
...
 
 // Init constructs a CLI command to initialize Celestia Node of the given type 'tp'.
 // It is meant to be used a subcommand and also receive persistent flag name for repository path.
-func Init(repoName string, tp node.Type) *cobra.Command {
+func Init(repoFlagName string, tp node.Type) *cobra.Command {
 	if !tp.IsValid() {
 		panic("cmd: Init: invalid Node Type")
 	}
-	if len(repoName) == 0 {
+	if len(repoFlagName) == 0 {
```
renaynay, line 17: won't `len(repoFlagName)` never == 0 ? We're removing dev mode anyway which is the only thing that passes `""` to this function.

cmd/start.go
```diff
@@ -35,6 +35,11 @@ func Start(repoName string, tp node.Type) *cobra.Command {
 		Args:         cobra.NoArgs,
 		SilenceUsage: true,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			repoPath := cmd.Flag(repoFlagName).Value.String()
+			if repoPath == "" {
+				return fmt.Errorf("repository path must be speficied")
```
renaynay, line 40:
```suggestion
				return fmt.Errorf("repository path must be specified")
```

cmd/init.go
Wondertan: Just a sanity check, as `Start` is exported and might be used outside. Does not hurt. Will revisit this when will delete Dev mode. Resolved?

Wondertan: @renaynay

liamsi (approved):

cmd/init.go
liamsi: BTW, is there an issue to track removal of dev mode? 
Can this be closed then: https://github.com/celestiaorg/celestia-node/pull/176#issuecomment-985824614

liamsi:
Is this tracked somewhere?
```
TestExtendedHeaderBroadcast
[...]
 event_test.go:100: 
        	Error Trace:	event_test.go:100
        	            	context deadline exceeded
        	Test:       	TestExtendedHeaderBroadcast
```

liamsi (approved):

cmd/init.go
renaynay: Implicitly tracked here: https://github.com/celestiaorg/celestia-node/issues/250

Resolving.

renaynay (approved):

Issue #262, opened 2021-12-03
[Feature Request]: In-memory Test Swamp Tool


Bidon15:
### Summary

Need in fast execution of tests during PR stages

### Impact on users

Having a robust process of testing our implementation during every PR will be beneficial to track the quality and catch bugs as early as possible in scenarios where we are deploying the whole network setup (light-full-bridge). Now, unit tests covers the features in isolation and we need to expand testing envs and cases. 

### Cost of not doing

As we will start to have RC and release stages in CD/CD, the cost of not doing this tool will be costly in finding the root cause of found bugs during e2e testing phases

### Evidence

During devnet, we have found out that most of the bugs could be caught using a more simple testing approach by directly creating components/test scenarios in code base and not opting full into dockerising the components for testing, leaving the latter for more e2e testing approaches. 

### Steps that we need to achieve in order to start development of test scenarios using the tool:

1. Have a code-base cleanup #250 
2. Discuss the architecture design of the tool @Wondertan and @Bidon15 
3. Implementation of the architecture @Bidon15 

### What will be tested with this tool
Already:
- syncing
- dasing

Overall:
- header service
- share service

What is needed: 
- state service 
- networking setups

### What is close to what we want:
- DALC [repo](https://github.com/celestiaorg/dalc)

### Urgency

High

liamsi:
Will this test-swamp also enable developers to quickly run tests locally?

Bidon15:
> Will this test-swamp also enable developers to quickly run tests locally?

Yes. It should be by design to give the same UX as what is already happening with unit tests

Bidon15:
For test swamp to operate we need these points: 
- Initialise the mocknet FullMeshLinked to link all full/bridge/light nodes
- Have `WithHost` and `WithCoreClient` options for node
  - `WithHost` is needed to link+connect nodes between each other
  - `WithCoreClient` is needed to feed bridge(currently it's full) node with an existing coreclient (mockedCoreClient)
 - Place the swamp in the `node/tests` directory
   -  `swamp` dir for the tool itself
   -  `domain_name/*_test.go` 

Ideally, swamp will take care of creating + linking + connecting all nodes and leaving the user to only focus on the testing scenario

Proposals so far:
- we need to migrate some of the existing full/light tests to the `node/tests` dir
- so far, having multiple embedded core client communicating between each other is a limitation, so we will start with testing 1 full/bridge node and 1+ light nodes

PR #264, opened 2021-12-06
cmd, cmd/celestia: Remove `dev` mode

renaynay:
We no longer need dev mode in its current form. Instead, we can test real nodes against small testnets as was done in Devnet and will be automated by @jbowen93, or via an in-memory test suite as will be implemented by @Bidon15 (Ref #262).

We may later introduce a `--dev` flag that runs a modified version of the node type, but that will be scoped out in later iterations.

Related to #250.


liamsi (approved):
image
Easy reviewer times 👌🏼

Wondertan (approved):
Approving, though I have a strange feeling that we may miss something here.

liamsi:
> Approving, though I have a strange feeling that we may miss something here.

What could it be?

Issue #163, opened 2021-10-29
Get Celestia version from git tags in Makefile 


orlandorode97:
### Summary

Get Celestia version from git tags in Makefile 

### Gain

The gain for this change is to get the `celestia` command-line version dynamically from git tags instead of having a hardcoded value.

### Evidence

This article is applying something similar: https://dev.to/eugenebabichenko/generating-pretty-version-strings-including-nightly-with-git-and-makefiles-48p3

### Implementation ideas

Instead of having a hardcoded value for the `celestia` version, we could run a shell command in the Makefile to get the version dynamically from the git tags.

### Urgency

Don't Know

renaynay:
This was fixed by @tzdybal

renaynay:
Wait nevermind @tzdybal only fixed the commit, but we don't have node tags yet, so reopening until we make a release.

PR #269, opened 2021-12-06
feat(Makefile): add tag parsing to inject version from git

Wondertan:
This last bit is required before the release, otherwise binary wouldn't properly show the version number.

Closes: https://github.com/celestiaorg/celestia-node/issues/163


liamsi (approved):
utACK

Wondertan:
Tested locally, works as expected

adlerjohn (approved):

Wondertan:
Output when a tag is aligned with HEAD.
image

Output when a tag is not aligned with HEAD
image

PR #277, opened 2021-12-08
service/block: Remove block fetching capability from `BlockService`

### LINKED ISSUE #251: service/block, service/header: Move `ExtendedHeader` creation from `BlockService` to `HeaderService`
Ref #244 
Related to #214 

TODO: 
- [ ] refactor `BlockService` so that it no longer subscribes to new block events from core
- [ ] refactor `HeaderService` so that it is able to subscribe to new block events from core after it is finished syncing headers (#327)

renaynay:
This PR is the first step necessary to resolve #251 (it resolves the first TODO).


Wondertan:
Btw, there should be a changelog entry for this one.

node/services/service.go
```diff
@@ -78,11 +77,9 @@ func HeaderStore(ds datastore.Batching) (header.Store, error) {
 // BlockService constructs new block.Service.
 func BlockService(
 	lc fx.Lifecycle,
-	fetcher *core.BlockFetcher,
 	store ipld.DAGService,
-	broadcaster header.Broadcaster,
 ) *block.Service {
-	service := block.NewBlockService(fetcher, store, broadcaster)
+	service := block.NewBlockService(store)
```
adlerjohn, line 82: You're removing the fetcher and broadcaster here, but aren't they needed for fetching and broadcasting blocks, not headers?

utils/erasure.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+// ExtendBlock erasure codes the given raw block's data and returns the
+// erasure coded block data upon success.
+func ExtendBlock(block *core.Block, dag format.DAGService) (*rsmt2d.ExtendedDataSquare, error) {
+	namespacedShares, _ := block.Data.ComputeShares()
+	shares := namespacedShares.RawShares()
+
+	// create nmt adder wrapping batch adder
+	batchAdder := ipld.NewNmtNodeAdder(context.Background(), format.NewBatch(context.Background(), dag))
```
adlerjohn, line 24: Why was this changed from `ce.shareStore` to `dag format.DAGService`?

utils/erasure.go
```diff
@@ -0,0 +1,38 @@
...
+
+	// compute extended square
+	return rsmt2d.ComputeExtendedDataSquare(shares, rsmt2d.NewRSGF8Codec(), tree.Constructor)
+}
+
+// squareSize64 computes the square size as uint64 from
+// the given length of shares.
+func squareSize64(length int) uint64 {
+	return uint64(math.Sqrt(float64(length)))
+}
```
adlerjohn, line 38: Is there a reason the TODO to un-dup this was removed?

Wondertan:
@adlerjohn, note that this PR is a part of a bigger goal and changes here might not make sense in isolation

adlerjohn:
@Wondertan I gathered that much 😂. So long as the noted peculiarities are expected I have no issue with them temporarily existing.

node/services/service.go
renaynay: No, block reconstruction will be done by fetching shares via `ShareService` instead and reconstructed / stored within `BlockService`.

utils/erasure.go
renaynay: I've de-duplicated it as much as possible (besides the IPLD package to avoid circular dependency) which will eventually be separated out into its own repo later on.

adlerjohn (approved):

utils/erasure.go
```diff
@@ -0,0 +1,38 @@
+package utils
```
Wondertan, line 1: Why utils? 

I like the approach of making things reusable but in a modular domain-specific way, not a monolithic way. And we are already going with the former.

The latter:
services/
    header_service.go
    share_service.go

The former: 
header/
    service.go
share/
    service.go

Same with utils. Having one big folder for all the utils is monolithic and prone to dependency cycles, as it depends on our data structures.

If this makes sense to you, pls move you the helper func inside a respective module/domain package.

adlerjohn (commented):
conflicts

renaynay:
@adlerjohn closing this PR in favour of #317

Issue #286, opened 2021-12-10
[Bug]: Race Condition in TestFull_P2P_Streams


Bidon15:
### Celestia Node version

a89ef036a260a2b81f64fa508b801c062b7cff32

### OS

Ubuntu 20.04.3 LTS

### Install tools

- git clone https://github.com/celestiaorg/celestia-node.git
- makefile

### Steps to reproduce it

1. Make install
2. go test --race ./...

### Expected result

TestFull_P2P_Streams is passing with no race conditions

### Actual result

TestFull_P2P_Streams is failing with data races

### Relevant log output

```shell
https://gist.github.com/Bidon15/9ec686472ed21ee77662fca18dc5ecf2
```

### Notes

We should create a separate GH Action to catch this during every PR

Issue #283, opened 2021-12-10
[Bug]: Race Condition in TestExtendedHeaderBroadcast


Bidon15:
### Celestia Node version

a89ef036a260a2b81f64fa508b801c062b7cff32

### OS

Ubuntu 20.04.3 LTS

### Install tools

- git clone https://github.com/celestiaorg/celestia-node.git
- makefile

### Steps to reproduce it

1. Make install 
2. go test --race ./...

### Expected result

TestExtendedHeaderBroadcast is passing with no race conditions

### Actual result

TestExtendedHeaderBroadcast is failing with race conditions

### Relevant log output

```shell
==================
WARNING: DATA RACE
Write at 0x00c0011c8de0 by goroutine 70:
  runtime.mapassign_fast64()
      /usr/local/go/src/runtime/map_fast64.go:92 +0x0
  github.com/celestiaorg/celestia-node/service/block.(*mockFetcher).generateBlocksWithValidHeaders()
      /home/appuser/go/src/github.com/celestiaorg/celestia-node/service/block/event_test.go:152 +0x454
  github.com/celestiaorg/celestia-node/service/block.TestExtendedHeaderBroadcast()
      /home/appuser/go/src/github.com/celestiaorg/celestia-node/service/block/event_test.go:95 +0x9a4
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1259 +0x22f
  testing.(*T).Run·dwrap·21()
      /usr/local/go/src/testing/testing.go:1306 +0x47

Previous read at 0x00c0011c8de0 by goroutine 24:
  [failed to restore the stack]

Goroutine 70 (running) created at:
  testing.(*T).Run()
      /usr/local/go/src/testing/testing.go:1306 +0x726
  testing.runTests.func1()
      /usr/local/go/src/testing/testing.go:1598 +0x99
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1259 +0x22f
  testing.runTests()
      /usr/local/go/src/testing/testing.go:1596 +0x7ca
  testing.(*M).Run()
      /usr/local/go/src/testing/testing.go:1504 +0x9d1
  main.main()
      _testmain.go:49 +0x22b

Goroutine 24 (running) created at:
  github.com/celestiaorg/celestia-node/service/block.(*Service).listenForNewBlocks()
      /home/appuser/go/src/github.com/celestiaorg/celestia-node/service/block/event.go:20 +0x264
[...]

Issue #232, opened 2021-11-23
panic: AddrFactory


Wondertan:
![image](https://user-images.githubusercontent.com/22449364/143041890-09e278c1-ce02-436d-b0f7-25ce2a5b5a48.png)

PR #288, opened 2021-12-11
service/block, node/p2p: Fix race conditions in TestExtendedHeaderBroadcast and TestFull_P2P_Streams

jenyasd209:
PR supposed to:
- fix https://github.com/celestiaorg/celestia-node/issues/286;
- fix https://github.com/celestiaorg/celestia-node/issues/283;
- and maybe fix https://github.com/celestiaorg/celestia-node/issues/232

### DIFF
--- a/CHANGELOG-PENDING.md
+++ b/CHANGELOG-PENDING.md
@@ -25,3 +25,4 @@ Month, DD, YYYY
 - [go package] (Link to PR) Description @username
 
 - [header] Added missing `err` value in ErrorW logging calls. @jbowen93
+- [service/block, node/p2p] [Fix race conditions in TestExtendedHeaderBroadcast and TestFull_P2P_Streams.](https://github.com/celestiaorg/celestia-node/pull/288) @jenyasd209
--- a/node/p2p/addrs.go
+++ b/node/p2p/addrs.go
@@ -47,7 +47,9 @@ func AddrsFactory(announce []string, noAnnounce []string) func() (_ p2pconfig.Ad
 
 		return func(maListen []ma.Multiaddr) (out []ma.Multiaddr) {
 			// combine maListen and maAnnounce addresses
-			maAnnounce = append(maAnnounce, maListen...)
+			out = make([]ma.Multiaddr, len(maAnnounce), len(maAnnounce)+len(maListen))
+			copy(out[:len(maAnnounce)], maAnnounce[:])
+
 			// filter out unneeded
 			for _, maddr := range maListen {
 				ok := maNoAnnounce[string(maddr.Bytes())]
--- a/node/p2p/addrs.go
+++ b/node/p2p/addrs.go
@@ -48,7 +48,7 @@ func AddrsFactory(announce []string, noAnnounce []string) func() (_ p2pconfig.Ad
 		return func(maListen []ma.Multiaddr) (out []ma.Multiaddr) {
 			// combine maListen and maAnnounce addresses
 			out = make([]ma.Multiaddr, len(maAnnounce), len(maAnnounce)+len(maListen))
-			copy(out[:len(maAnnounce)], maAnnounce[:])
+			copy(out[:len(maAnnounce)], maAnnounce)
 
 			// filter out unneeded
 			for _, maddr := range maListen {
--- a/node/p2p/addrs.go
+++ b/node/p2p/addrs.go
@@ -45,10 +45,10 @@ func AddrsFactory(announce []string, noAnnounce []string) func() (_ p2pconfig.Ad
 			maNoAnnounce[string(maddr.Bytes())] = true
 		}
 
-		return func(maListen []ma.Multiaddr) (out []ma.Multiaddr) {
-			// combine maListen and maAnnounce addresses
-			out = make([]ma.Multiaddr, len(maAnnounce), len(maAnnounce)+len(maListen))
-			copy(out[:len(maAnnounce)], maAnnounce)
+		return func(maListen []ma.Multiaddr) []ma.Multiaddr {
+			// copy maAnnounce to out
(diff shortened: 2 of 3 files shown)


node/p2p/addrs.go
```diff
@@ -47,7 +47,9 @@ func AddrsFactory(announce []string, noAnnounce []string) func() (_ p2pconfig.Ad
 
 		return func(maListen []ma.Multiaddr) (out []ma.Multiaddr) {
 			// combine maListen and maAnnounce addresses
-			maAnnounce = append(maAnnounce, maListen...)
+			out = make([]ma.Multiaddr, len(maAnnounce), len(maAnnounce)+len(maListen))
+			copy(out[:len(maAnnounce)], maAnnounce[:])
```
jenyasd209, line 51: want to clarify: Were it useless appending or `maAnnounce` should be used in `out`?

Bidon15:
Thank you so much for the PR @jenyasd209 ! 🥳 
It would be great if you could:

1. Update the changelog.md 
2. Fix the lint error on `maAnnounce[:]`
3. Take a look on new error thrown during failure of TestExtendedHeaderBroadcast

jenyasd209:
> 3\. TestExtendedHeaderBroadcast

Updated. The problem is I have no failures locally (go v1.16)

Wondertan:
@jenyasd209, thank you for taking over bugs!
I am fine with the Race fixes, but unsure about #232 being fixed. The bug described there is not reproducible and happened only once in an unusual situation. Furthermore, I am not sure why this bag happened in the first place. There should not be any panics as append(on the line 50) should not produce such errors, due to the way appends works. Also, a lot of things changed from the point the bug appeared.

Let's proceed by removing your fix for #232 in this PR and closing the issue, as there should be no panics on the code as it is right now and the fix does not really fix anything.

Wondertan:
@jenyasd209, also, the error you were facing is test flakiness and can be resolved by re-running all tests using github actions page

jenyasd209:
> @jenyasd209, thank you for taking over bugs!
> 
> I am fine with the Race fixes, but unsure about #232 being fixed. The bug described there is not reproducible and happened only once in an unusual situation. Furthermore, I am not sure why this bag happened in the first place. There should not be any panics as append(on the line 50) should not produce such errors, due to the way appends works. Also, a lot of things changed from the point the bug appeared.
> 
> Let's proceed by removing your fix for #232 in this PR and closing the issue, as there should be no panics on the code as it is right now and the fix does not really fix anything.

I tagged https://github.com/celestiaorg/celestia-node/issues/232 because it looks similar to https://github.com/celestiaorg/celestia-node/issues/286 (logs https://gist.github.com/Bidon15/9ec686472ed21ee77662fca18dc5ecf2). There is race condition at the same line where panic happens. Due to append is not thread safe panic can happen if there are a couple calls of [this func](https://github.com/celestiaorg/celestia-node/pull/288/files#diff-6c92c4208db674253540bd33d7e10e55af978b7dbc0c0104b2b0726d1a5869b1L48).

Wondertan:
@jenyasd209, ooh, I see. Thanks for pointing me to the connection with the races. I made a deeper look and there is clearly a bug/typo. We weren't even using `maAnnounce` after the append, as you can see there. We should append there not to the `maAnnounce, but to `maListen`, i.e: `maListen = append(maListen, maAnnounce...)`. At the first sight, I was even thinking that this happens. Also, this resolves races, as no more appending of the shared slice will happen.

node/p2p/addrs.go
Wondertan: @jenyasd209, Yeah, it's useless, should be `maListen = append(maListen, maAnnounce...)`. Thanks again for finding this out!

jenyasd209: Can addresses from `maAnnounce` exist in `maNoAnnounce`?

Wondertan: @jenyasd209, Technically yes, in practice no one will do this

jenyasd209: discussions can be closed?

Wondertan: I was thinking you are going to change appending to `maListen`, but to out is also fine.

Wondertan: @jenyasd209, On last thing here I care about, is readability. Consider the following:
```go
out = make([]ma.Multiaddr, 0, len(maAnnounce)+len(maListen)) // this can be even removed, but proper preallocation is good
out = append(out, maAnnounce)
```

Wondertan: Does exactly the same as your code, but less bulky.

Wondertan (dismissed):
Awesome, thank you!

CHANGELOG-PENDING.md
```diff
@@ -7,7 +7,7 @@ Month, DD, YYYY
 ### BREAKING CHANGES
 
 - [go package] (Link to PR) Description @username
-[feat(cmd): give a birth to cel-shed and p2p key utilities #281](https://github.com/celestiaorg/celestia-node/pull/281) [@Wondertan](https://github.com/Wondertan)
+- [feat(cmd): give a birth to cel-shed and p2p key utilities #281](https://github.com/celestiaorg/celestia-node/pull/281) [@Wondertan](https://github.com/Wondertan)
```
renaynay, line 10: I guess this got included somehow on accident?

renaynay (dismissed):
Small nit but otherwise good. Unfortunately `TestExtendedHeaderBroadcast` will no longer be needed as header production will be removed from `BlockService` anyway with #277 .

CHANGELOG-PENDING.md
Wondertan: I guess he just added `-` to make it a bullet point.

Wondertan:
@jenyasd209, can you pls make a rebase?

Wondertan (approved):

renaynay (approved):

PR #289, opened 2021-12-14
ci: add race detector job

Bidon15:
it will run in parallel with normal unit test execution
relates to #286


renaynay:
Will this run on every PR? I think it's a bit overkill like @Wondertan mentioned on Friday. I agree that it will be distracting as we won't know whether the tests are flakey or the code is racey. Can we run on every release instead?

Bidon15:
> Will this run on every PR? I think it's a bit overkill like @Wondertan mentioned on Friday. I agree that it will be distracting as we won't know whether the tests are flakey or the code is racey. Can we run on every release instead?

Well, considering that release is not happening that often, I think we can introduce the schedule to be at least once per week (let's say Monday 😅 ) 

I'll upload the updated conf

renaynay:
I think weekly should be fine. cc @Wondertan

Wondertan:
> Don't we want the tests to be non-flaky too though?

@renaynay, @liamsi We want and race flag that will help us find such hidden ones. The only thing I am worried about is a distraction to fix those for past code and constantly re-running tests to finally get green. This is true unless we find time to allocate for fixing all the existing flakiness. Then we will be able to easily put --race flag on every test run and fix flakiness early, along with races. 

I am totally fine with the second option, as this is a technical debt that we have to fulfill anyway.

Bidon15:
So, overall I still left the scheduler of the action to be run every Monday 12:00(24hrs) UTC

Wondertan:
Let's return to this question once @liamsi is back. After thinking for some time, I think I was wrong, and having `-race` being run on each test run is the rightest approach that we should stick to.

cc @renaynay

renaynay:
I still like the idea that we run it once a week. @Wondertan

Wondertan:
I understand why you like it and still would like to encourage you to embrace and perceive the always with `race` flag approach due to its correctness and tech debt "preventness".

Wondertan:
@liamsi, pls give your opinion on this one. You kinda inspired me to have --race flag always.

liamsi:
I think it should alsways run. tests should be written in a way that this will be possible.

Wondertan:
@Bidon15 can you pls update here the workflow to always run things with -race?

renaynay:
Number 1 IMO @Bidon15

renaynay:
Well to clarify, I'd like for it to be a specifically separate action (so we can see `test`, `race`, `lint` etc). I don't mind if it runs in parallel w test, just that it's viewable as a separate action

Wondertan:
Agree. Let's do number 1

Bidon15:
So this PR has caught 2 new data races
https://github.com/celestiaorg/celestia-node/issues/339
https://github.com/celestiaorg/celestia-node/issues/340

Also there is a failure with Adjecent test but I haven't locally caught it, yet  😓

CHANGELOG-PENDING.md
```diff
@@ -28,6 +28,7 @@ Month, DD, YYYY
 - [service/header: Refactor `HeaderService` to only manage its sub-services' lifecycles #317](https://github.com/celestiaorg/celestia-node/pull/317) [@renaynay](https://github.com/renaynay)
 - [docker] Created `docker/` dir with `Dockerfile` and `entrypoint.sh` script. 
 - [chore(share): handle rows concurrently in GetSharesByNamespace #241](https://github.com/celestiaorg/celestia-node/pull/241) [@vgonkivs](https://github.com/vgonkivs)
+- [test.yml](https://github.com/celestiaorg/celestia-node/pull/289) ci: adding data race detector action [@Bidon15](https://github.com/Bidon15)
```
renaynay, line 31:
```suggestion
- [ci: adding data race detector action](https://github.com/celestiaorg/celestia-node/pull/289) [@Bidon15](https://github.com/Bidon15)
```

renaynay (commented):
just changelog nit and then approve otherwise - we can merge with a failing test-race. We need to address all races later. Thanks!

renaynay (approved):

liamsi (approved):

PR #291, opened 2021-12-15
node: remove InitWith

Wondertan:
## About 
This PR removes redundant `node.InitWith` as we have a separate option to pass a Config, so to have similar functionality, one should use `node.Init(node.WithConfig())`.


adlerjohn (approved):

renaynay (approved):
👍🏻

Bidon15 (approved):

renaynay (approved):

PR #290, opened 2021-12-15
node: enhance DI allowing overriding of dependencies

Wondertan:
## About
This is a solution for https://github.com/celestiaorg/celestia-node/issues/285. It is different from the design there, as it's not possible. Finding the other solution took a lot of time, sketching, and learning. The reason why is the limitation of FX described in uber-go/fx#825. This PR solves the problem by extending `fxutil` lib to mimic `fx` functionality + new `OverrideSupply` option described in uber-go/fx#825. In future, the code of `fxutil` might be deleted once the described `fx` feature is implemented and merged.

This PR might be difficult to understand in detail as it involves reflection + understanding of the DI itself, but also a good opportunity to understand how both worlds work for the implementer and for the reviewers. 

## Changes
* `fxutil` mimicing `fx` options with more new additionals
	* `OverrideSupply`
	* `SupplyAs`
	* `ProvideAs`
* Remove `fx.App` from Node
* New option to override p2p priv key(unblocking https://github.com/celestiaorg/celestia-node/issues/284)

PR #292, opened 2021-12-15
feat(node): extract overrides from Config into Settings

Wondertan:
## About 
#290 adds Component overriding options to `node` and applies those onto Config. This is confusing and this PRs aims to fix this by introducing `node.settings`.


node/init.go
```diff
@@ -13,10 +13,10 @@ import (
 // Init initializes the Node FileSystem Repository for the given Node Type 'tp' in the directory under 'path' with
 // default Config. Options are applied over default Config and persisted on disk.
 func Init(path string, tp Type, options ...Option) error {
-	cfg := DefaultConfig(tp)
+	cfg, sets := DefaultConfig(tp), new(settings)
```
adlerjohn, line 16: Not the biggest fan of `sets`. Is there a better name? Maybe even `s`.

node/fxutil/options.go
```diff
@@ -131,6 +131,11 @@ func OverrideSupply(vals ...interface{}) Option {
 				refVal, tp = refVal.Elem(), tp.Elem()
 			}
 
+			if !refVal.IsValid() {
+				// if there was nil - ignore
+				continue
```
adlerjohn, line 136: Is it okay to silently ignore an invalid thing here?

node/init.go
Wondertan: Don't have strong preferences about naming here. Btw, why?

node/fxutil/options.go
Wondertan: Good question. All the __really__ invalid cases are checked and errors for them are thrown above the line. Regarding this case, it is basically `nil` and it is usually fine to allow it to be passed in the optional pattern, e.g  `WithSomething(nil)`, so you don't need to check for `if something != nil { WithSomething(something) }`.

node/init.go
adlerjohn: "sets" makes me initially think of https://en.wikipedia.org/wiki/Set_(mathematics)

node/fxutil/options.go
adlerjohn: Ah okay. Can you document this in a comment?

renaynay (approved):
I'm good with this as long as @adlerjohn 's comments are addressed.

renaynay (approved):

Bidon15 (approved):

PR #293, opened 2021-12-15
feat(cmd):  custom p2p key flag

### LINKED ISSUE #284: cmd: A flag to set custom private key for p2p communications
Right now, the p2p private key is generated and saved on the first run, but it should be possible to set it from the CLI flags to facilitate testing utilities and Node deployment by excluding indeterministic key generation. Instead, the infrastructure should control non-determinism and be able to pre-generate the key using [`cel-shed`](#276) and apply it to the Node.

Wondertan:
## About
Currently, only `start` cmd works with a custom private key.  To make `init` work #291 and #292 have to be merged. 

Closes #284

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -10,6 +10,7 @@ var configFlags = []*pflag.Flag{
 	trustedHashFlag,
 	trustedPeerFlag,
 	coreRemoteFlag,
+	p2pKeyFlag,
 }
 
 var (
@@ -37,6 +38,11 @@ var (
 		Name: "core.remote",
 		Usage: "Indicates node to connect to the given remote core node. " +
 			"Example: <protocol>://<ip>:<port>, tcp://127.0.0.1:26657",
+		DefValue: "",
+	}
+	p2pKeyFlag = &pflag.Flag{
+		Name:     "p2p.key",
+		Usage:    "Custom p2p Ed25519 private key",
 		DefValue: "",
 	}
 )
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -1,8 +1,10 @@
 package cmd
 
-import "flag"
+import (
+	"github.com/spf13/pflag"
+)
 
-var configFlags = []*flag.Flag{
+var configFlags = []*pflag.Flag{
 	loglevelFlag,
 	nodeConfigFlag,
 	trustedHashFlag,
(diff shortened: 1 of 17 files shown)


cmd/flags.go
```diff
@@ -1,40 +1,48 @@
...
-	coreRemoteFlag = &flag.Flag{
+	coreRemoteFlag = &pflag.Flag{
 		Name: "core.remote",
 		Usage: "Indicates node to connect to the given remote core node. " +
 			"Example: <protocol>://<ip>:<port>, tcp://127.0.0.1:26657",
 		DefValue: "",
 	}
+	p2pKeyFlag = &pflag.Flag{
+		Name:     "p2p.key",
+		Usage:    "Custom p2p Ed25519 private key",
```
renaynay, line 45: Is there a reason to pass it directly in the command line rather than as a file?

renaynay: @jbowen93 Is there a reason for this from the docker side? IMO it should be fine to just pass a file path rather than the actual key itself. This common practice and if it's an ENV var issue, then just mount the priv key into the container and set the ENV var to the path.

adlerjohn: Yeah it's a bit sus to get a private key as a non-interactive parameter. If anything we shouldn't even give that option to prevent bad habits.

musalbas: Yes, this should be avoided, because command line history is usually stored on disk to .bash_history, etc. This means an unprivileged attacker could be able to extract the private key, even if the server operator only intended for the keys to be stored in RAM and not on disk.

Wondertan: Good points! To resolve we need input from @jbowen93, he had some thoughts about this and ask me to implement this as a param to facilitate docker setup, IIRC.

One more thing is that this param is supposed to be hidden, meaning that it exists, but is not shown on `help`, so regular users wouldn't know about it unless they read the code or were told by someone.

jbowen93: Apologies for delay on getting back to this. Feedback from Renay, John, and Mustafa is correct, we should be using files for this instead of env vars. That's my mistake requesting this from Hlib.

The current setup for [docker](https://github.com/celestiaorg/devops-test/blob/main/devnet/docker/bridge-docker-compose.yml#L21) and [k8s](https://github.com/celestiaorg/devops-test/blob/main/devnet/k8s/core-nodes/0/core0-deployment.yaml#L93) expects the key to exist at `{repo.path}/keys/OAZHALLLMV4Q`.

jbowen93: https://github.com/celestiaorg/networks/issues/46 is the issue I'm using to track documenting best practices for external validators.

Wondertan:
Ok, closing this PR: 
* Flags should now be implemented and based over #313 
* Approach with the flag is not the way to go, as we figured out.

PR #294, opened 2021-12-15
chore: rename Full node to Bridge node

Wondertan:
Simple renaming.

Closes #250. I decided not to introduce a new Full node type yet as there are still things like #179 and #181 that need to be implemented first and maybe even #223

PR #295, opened 2021-12-16
docker: Create a single Dockerfile that works for full and light `celestia-node`s

jbowen93:
Create a new `Dockerfile` that works for both celestia-node full and light nodes while matching the expectations of both the docker-compose and k8s cluster setups. 

At runtime the a `NODE_TYPE` environment variable matching either `full` or `light` is expected. 

If not provided it defaults to `full`. 

For examples see the [docker-compose config](https://github.com/celestiaorg/devops-test/blob/main/devnet/docker/docker-compose.yml#L84) and the [k8s config](https://github.com/celestiaorg/devops-test/blob/main/devnet/k8s/core-nodes/0/core0-deployment.yaml#L76).

This has been manually tested by building the Docker image
```
docker build -f docker/Dockerfile -t jbowen93/celestia-node:multi
``` 
and then testing it with docker-compose
```
cd devnet
./start-docker.sh
```
and with kubernetes
```
cd devnet
kubectl create -f k8s/core-nodes -R && \
sleep 45s && \
kubectl create -f k8s/light-nodes -R
```

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,24 @@
+FROM golang:1.17 as builder
+RUN apt update && \
+    apt install make bash
+ENV HOME /celestia-node
+COPY / ${HOME}
+WORKDIR ${HOME}
+RUN make build 
+
+FROM ubuntu
+ENV NODE_TYPE full
+ENV SLEEP 0s
+
+WORKDIR $APP
+
+# COPY docker/entrypoint.sh script/entrypoint.sh
+COPY docker/entrypoint.sh /
+
+# Copy in the binary
+COPY --from=builder /celestia-node/celestia /
+
+EXPOSE 2121
+
+ENTRYPOINT ["/entrypoint.sh"]
+CMD ["celestia"]
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -4,16 +4,11 @@ RUN apt update && \
 ENV HOME /celestia-node
 COPY / ${HOME}
 WORKDIR ${HOME}
-RUN make build 
+RUN env GOOS=linux GOARCH=amd64 make build
 
 FROM ubuntu
 ENV NODE_TYPE full
-# Issue/bug #279 
-ENV SLEEP 0s
 
-WORKDIR $APP
-
-# COPY docker/entrypoint.sh script/entrypoint.sh
 COPY docker/entrypoint.sh /
 
 # Copy in the binary
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -8,6 +8,7 @@ RUN make build
 
 FROM ubuntu
 ENV NODE_TYPE full
+# Issue/bug #279 
 ENV SLEEP 0s
 
 WORKDIR $APP
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -20,5 +20,4 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v2
         with:
-          # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
-          version: v1.42
+          version: v1.43
--- a/CHANGELOG-PENDING.md
+++ b/CHANGELOG-PENDING.md
@@ -18,8 +18,6 @@ Month, DD, YYYY
 
 - [docker] Created `docker/` dir with `Dockerfile` and `entrypoint.sh` script. 
 
-- [docker] Created `docker/` dir with `Dockerfile` and `entrypoint.sh` script. 
-
 ### BUG FIXES
 
 - [go package] (Link to PR) Description @username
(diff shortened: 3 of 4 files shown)


docker/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM golang:1.17 as builder
+RUN apt update && \
+    apt install make bash
+ENV HOME /celestia-node
+COPY / ${HOME}
+WORKDIR ${HOME}
+RUN make build 
+
+FROM ubuntu
```
Bidon15, line 9: out of curiosity, is it possible to have alpine linux instead of ubuntu? 🧐 
imho making docker image with smaller footprint will be great

docker/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+ENV NODE_TYPE full
+ENV SLEEP 0s
```
Bidon15, line 11: I suggest we have to mention the issue why we need this variable
```suggestion
# Issue/bug #279 
ENV SLEEP 0s
```

renaynay (commented):
Thank you so much for this @jbowen93 

Can you change the name of the dockerfile to `bridge.Dockerfile` since this will be called a bridge node? Similarly after merging https://github.com/celestiaorg/celestia-node/pull/294, the `NODETYPE` will have to be `bridge`.

docker/Dockerfile
jbowen93: I tried using Alpine but there's something not working in it, probably related to musl libc and some cgo thing. I didn't want to track it down so I punted to Ubuntu.

jbowen93:
> Thank you so much for this @jbowen93
> 
> Can you change the name of the dockerfile to `bridge.Dockerfile` since this will be called a bridge node? Similarly after merging #294, the `NODETYPE` will have to be `bridge`.

I don't believe we should change the name of the Dockerfile. This Dockerfile is intentionally designed to work with *both* full nodes (now bridge nodes) and light nodes, you just pass different environment variables and startup commands. See an example of a [full/bridge node](https://github.com/celestiaorg/devops-test/blob/main/devnet/docker/docker-compose.yml#L84) and a [light node](https://github.com/celestiaorg/devops-test/blob/main/devnet/docker/docker-compose.yml#L172).

Also keep in mind that the env vars set in the Dockerfile are only defaults that should be overwritten later. From [the docs](https://docs.docker.com/compose/environment-variables/):

> When you set the same environment variable in multiple files, here’s the priority used by Compose to choose which value to use:
>
> Compose file
> Shell environment variables
> Environment file
> Dockerfile
> Variable is not defined

jbowen93:
Lint should be resolved after https://github.com/celestiaorg/celestia-node/pull/308 is merged

docker/Dockerfile
marbar3778: what would be the difference in image size if you were to download what is needed in an alpine image vs using ubuntu?

renaynay:
> I don't believe we should change the name of the Dockerfile. This Dockerfile is intentionally designed to work with both full nodes (now bridge nodes) and light nodes, you just pass different environment variables and startup commands. See an example of a full/bridge node and a light node.

@jbowen93 Okay as long as eventually we have automation for pushing three different images to dockerhub on release, one for bridge, full, and light so that people can just pull and run.

renaynay (approved):

Bidon15 (approved):

jbowen93:
> > I don't believe we should change the name of the Dockerfile. This Dockerfile is intentionally designed to work with both full nodes (now bridge nodes) and light nodes, you just pass different environment variables and startup commands. See an example of a full/bridge node and a light node.
> 
> @jbowen93 Okay as long as eventually we have automation for pushing three different images to dockerhub on release, one for bridge, full, and light so that people can just pull and run.

Just clarifying some more. This image supports multiple node types at run time. There is no need to push different images. 

Full node (will rename to bridge)
```
 docker run --network docker_localnet -e NODE_TYPE=full \
 jbowen93/celestia-node:multi celestia full --repo.path /celestia-full start \
 --core.remote tcp://192.167.10.2:26657 \
 --headers.trusted-hash "64D831D041A57A9D93078B445F7B1347024A9B5A8A444086C0B556B65A8A96C7" \
 --p2p.mutual "/ip4/192.167.10.7/tcp/2121/p2p/12D3KooWGGtejCnY3Q9XUUy6muSRm1Whgf8ijTRtwJ8DGA1164EB,/ip4/192.167.10.8/tcp/2121/p2p/12D3KooWQCbe8R4QD1jxnUs5SWWdFg5Kza3mBE94Cz1RJnLTXfNM,/ip4/192.167.10.9/tcp/2121/p2p/12D3KooWA6uovjYCgFvWhAbrcQmnLihFy1gZkdyJt8SrPKZjeWda"
```

light node
```
docker run --network docker_localnet -e NODE_TYPE=light \
jbowen93/celestia-node:multi celestia light --repo.path /celestia-light \
start --headers.trusted-peer "/ip4/192.167.0.2/tcp/2121/p2p/12D3KooWHLnz72q5mmCBooVbkjNcSWTzhEsc6gtSPwCqgervgNnb" \
 --headers.trusted-hash "64D831D041A57A9D93078B445F7B1347024A9B5A8A444086C0B556B65A8A96C7"
```

I did a quick manual test for this:
1.  Set up the core nodes with `docker-compose docker/core-docker-compose.yml up`
2. Run the above full node `docker run` command
3. Retrieve the ip address of the full node by looking through `docker network inspect docker_localnet`
4. Get the peer id for the full node. I ran `docker logs a345973ca120 2> /dev/null` for this
5. Run the above light node `docker run` command

PR #297, opened 2021-12-16
chore: bump deps

### LINKED ISSUE #274: go mod: massive dependecy update

Wondertan:
Kinda closes #274. 

Why kinda? Because we can't update most of libp2p and ipfs deps, until we don't upgrade NMT-IPLD to support conversion to ipld/go-ipld-prime. 

Why? Because there was an interface change in `go-datastore` which affected all the PL repos relying on `go-datastore` roughly 80%. 

Why is that a problem? The IPFS migration to `go-ipld-prime` happened before this massive `go-datastore`(DS) migration, so if you want to use the latest `go-libp2p`, it forces you to rely on the patched `go-datastore`, which forces you to rely on the patched `go-merkledag`, which forces you to use `go-ipld-prime`.

Without updating to the latest libp2p stuff we loosing sometimes important bug fixes and optimizations(mostly in bitswap related). Fortunately, there is one more reason to migrate to `go-ipld-prime` (or ipldv1).


adlerjohn (approved):

renaynay (approved):
Thanks for this @Wondertan , do you mind documenting on #274 what you wrote in the description so we have an easy reference.

PR #308, opened 2021-12-21
workflows/lint: update golangci-lint to v1.43

Wondertan:


.github/workflows/lint.yml
```diff
@@ -20,5 +20,4 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v2
         with:
-          # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
-          version: v1.42
+          version: latest
```
Bidon15, line 23:
```suggestion
          version: v1.43
```

Bidon15 (approved):

.github/workflows/lint.yml
```diff
@@ -20,5 +20,4 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v2
         with:
-          # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
-          version: v1.42
+          version: v1.43
```
adlerjohn, line 23: Nit, but can we keep the documentation string
```
# Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
```
?

Wondertan: It's not correct anymore and has never been. It says we only use the latest version but we only use concrete one instead.

adlerjohn (approved):

renaynay (approved):

Issue #243, opened 2021-11-24
Create Github Action to trigger on PR merge to celestia-node `main` branch


jbowen93:
Create a [Github Action](https://docs.github.com/en/actions) that is triggered when a PR is merged to the [celestia-node](https://github.com/celestiaorg/celestia-node) `main` branch.

This can initially just be a placeholder action.

adlerjohn:
@jbowen93 is this still needed?

ramin:
our GitHub actions have astral traveled through many galaxies, to where i'd safely say "this is done"

Issue #68, opened 2021-09-20
docker: Create CI pipeline to build docker image and push on every release to dockerhub


renaynay:
Create a way for us to build a new docker image on every release and push it to dockerhub for easy pull + testing.

jbowen93:
Let's use Github's built in [Github Container Registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry) for this. It allows us to have all artifacts in a single page. See [devops-test](https://github.com/celestiaorg/devops-test/pkgs/container/devops-test) as an example.

PR #310, opened 2021-12-22
Create GH Action that generates a pre-release w/ latest tag

jbowen93:
This PR creates a Github Action that will create a pre-release titled "Development Build" using the `latest` tag whenever a commit is merged to the `main` branch. If there is are any existing releases utilizing the `latest` tag they will be deleted and replaced with the new release. 

The release is done using [marvinpinto/automatic-releases](https://github.com/marvinpinto/actions/tree/master/packages/automatic-releases). The commit log is automatically created by retrieving all commits that have occurred since the last `latest` tagged release. 

An example run is [here](https://github.com/jbowen93/celestia-node/runs/4602026237?check_suite_focus=true)

An example release is [here](https://github.com/jbowen93/celestia-node/releases/tag/latest)

pre-release-snip

This is a necessary step for #68 

Resolves https://github.com/celestiaorg/celestia-node/issues/243


.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,33 @@
...
+        uses: "actions/checkout@v2"
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.17
+      - name: "Build"
+        run: "make build"
+      - name: "Test"
+        run: "make test"
+      - uses: "marvinpinto/action-automatic-releases@latest"
```
marbar3778, line 26: since this is a go project id recommend using something like https://github.com/goreleaser/goreleaser. Cosmos uses it on tendermint, cosmos-sdk and other repos. It has some use fun functionality

jbowen93: Hey Marko, I've played around with `goreleaser` a bit on [this branch](https://github.com/jbowen93/celestia-node/tree/jbowen93/goreleaser). Unfortunately `goreleaser` is opinionated and will only ever create a release if there's an associated SemVer compatible tag ([docs](https://goreleaser.com/limitations/semver/)). 

See the message on a[ manual run](https://github.com/jbowen93/celestia-node/runs/4694054105?check_suite_focus=true#step:5:15)
```
No tag found for commit '8000227'. Snapshot forced
/opt/hostedtoolcache/goreleaser-action/1.2.5/x64/goreleaser release --rm-dist --release-notes CHANGELOG-PENDING.md --snapshot
```

This GH action is intentionally designed to create a pre-release on all pushes to main/master. There should be a separate GH action that handles SemVer releases. That action should definitely use `goreleaser`.

marbar3778: why are there pre releases on every push?

jbowen93: > why are there pre releases on every push?

The goal workflow I'm trying to get to is that every merge to master results in a new release and docker image, both tagged as `latest`. This docker image is then used to standup an ephemeral cluster in both [docker-compose](https://github.com/celestiaorg/devops-test/tree/main/devnet/docker) and [k8s](https://github.com/celestiaorg/devops-test/tree/main/devnet/k8s) and a basic [smoke test](https://en.wikipedia.org/wiki/Smoke_testing_(software)) is run against it.  Right now the "smoke test" is just checking that all nodes in the ephemeral cluster have the same first block ([code](https://github.com/celestiaorg/devops-test/blob/main/.github/workflows/docker-compose.yml#L19)). 

The purpose of this is to find regressions ASAP while not blocking PRs to individual repos on cross component integration tests. 

If we just test with docker images it's not important to have a release but some validators don't intend to use docker so we may want to extend the above plan to include smoke testing an ephemeral cluster consisting of VMs.

marbar3778: hmm, with docker:latest it makes sense but a release for every push to master is overkill imo. On top of this validators should not update regularly as if there is a breaking change it will halt the chain. Happy to hop on a call to talk through how validator coordination and release works on other chains. 

A consistent release cycle like this is something I generally advise against with teams.

jbowen93: It's worth clarifying that this is purely for development. Obviously validators shouldn't be using the latest release but we shouldn't only perform integration testing on numbered releases. The earlier we catch breakages the easier it will be to find the root cause.

This process will only ever result in a single pre-release with the idea being that you could have a test setup that always grabs the binary via `wget https://github.com/celestiaorg/celestia-node/releases/download/latest/celestia`.

Wondertan (dismissed):
LGTM

adlerjohn (approved):
utACK modulo conflicts

liamsi:
What is the point of cutting/tagging a release for each merge on master? We want monthly releases (or manual ones in between if the dev team decides to do so).

jbowen93:
> What is the point of cutting/tagging a release for each merge on master? We want monthly releases (or manual ones in between if the dev team decides to do so).

I'm fine with closing this issue and instead having a GH action that creates a `latest` tagged image that can be used for testing.

Issue #311, opened 2021-12-22
cmd: Hardcode default repo name strings


renaynay:
This is the default directory for the node's persistent files, right? If we're building something that isn't just a scappy devnet I think this should be a constant instead of a hard-coded string.

_Originally posted by @adlerjohn in https://github.com/celestiaorg/celestia-node/pull/294#r772747161_

PR #313, opened 2021-12-23
cmd: introduce Env - an Environment for CLI commands

Wondertan:
This is a slight refactoring of `cmd` package to:
* Have more granular control over flags
	* So Light node doesn't have Core related flags 
	* Bridge node doesn't have `trustedPeer` flags and etc
* Have a pattern on how we add additional flags and node types
* Have encapsulated shared environment for commands to put common things into
	* Flag parsing
	* (Soon) RPC Client related initializations
	* more 
* Have an ergonomic API for us while developing commands
	
Also, the PR removes `completion` and `help`(use --help instead) commands for simplicity until we really need them.

Commands and flags are tested manually, but still, there are might be some issues, so look carefully.

NOTE for reviewers: I suggest reviewing it not only from GH but also by checking out and looking at the new files structure in `cmd` package to get the whole idea.

Also closes: https://github.com/celestiaorg/celestia-node/issues/311


Wondertan:
Linter is failing for some weird issue. It says there are line duplicates but if you look closely, there are no duplicates, even those lines really look similar, but they just follow the same pattern.

adlerjohn:
Is this one ready for review?

Wondertan:
@adlerjohn, Why not? Not a draft, and all the checks are ok, so yes, is ready

cmd/env.go
```diff
@@ -0,0 +1,49 @@
...
+	return env.opts
+}
+
+func (env *Env) addOption(opt node.Option) {
+	env.opts = append(env.opts, opt)
+}
+
+type envKey struct{}
+
+var env = envKey{}
```
renaynay, line 49: what's the point of this?

cmd/flags_core.go
```diff
@@ -0,0 +1,58 @@
...
+	return nil
+}
+
+// parseAddress parses the given address of the remote core node
+// and checks if it configures correctly
+func parseAddress(address string) (string, string, error) {
+	u, err := url.Parse(address)
+	if err != nil || u.Scheme == "" || u.Host == "" {
+		return "", "", err
+	}
```
renaynay, line 51: if `u.Scheme` is empty but err == nil, it'll still return a nil error even though you want it to return an err right?

cmd/flags_headers.go
```diff
@@ -0,0 +1,59 @@
...
+
+	"github.com/multiformats/go-multiaddr"
+	"github.com/spf13/cobra"
+	flag "github.com/spf13/pflag"
+
+	"github.com/celestiaorg/celestia-node/node"
+)
+
+var (
+	headersTrustedHashF = "headers.trusted-hash"
```
renaynay, line 15: I'm guessing `headersTrustedHashF` is meant to stand for `headersTrustedHashFlag`. I'd prefer it not be F and just be `Flag`.

cmd/flags_headers.go
```diff
@@ -0,0 +1,59 @@
...
+	headersTrustedPeerF = "headers.trusted-peer"
+)
+
+func HeadersFlags() *flag.FlagSet {
+	flags := &flag.FlagSet{}
+
+	flags.String(
+		headersTrustedHashF,
+		"",
+		"Hex encoded block hash. Starting point for header synchronization",
```
renaynay, line 25:
```suggestion
		"Hex-encoded block hash checkpoint for header synchronization",
```

cmd/flags_node.go
```diff
@@ -0,0 +1,49 @@
...
+
+	flags.String(
+		nodeStoreF,
+		fmt.Sprintf("~/.celestia-%s", strings.ToLower(tp.String())),
+		"The path to root/home directory of your Celestia Node Store",
+	)
+	flags.String(
+		nodeConfigF,
+		"",
+		"Path to a customized Node Config",
```
renaynay, line 29:
```suggestion
		"Path to a customized node config TOML file",
```

cmd/flags_node.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+func ParseNodeFlags(cmd *cobra.Command, env *Env) error {
+	env.storePath = cmd.Flag(nodeStoreF).Value.String()
+
+	nodeConfig := cmd.Flag(nodeConfigF).Value.String()
+	if nodeConfig != "" {
+		cfg, err := node.LoadConfig(nodeConfig)
+		if err != nil {
+			return fmt.Errorf("cmd: while parsing '%s': %w", nodeStoreF, err)
```
renaynay, line 42:
```suggestion
			return fmt.Errorf("cmd: while parsing '%s': %w", nodeConfigF, err)
```

cmd/start.go
```diff
@@ -1,138 +1,62 @@
...
+// Start constructs a CLI command to start Celestia Node daemon of any type with the given flags.
+func Start(fsets ...*flag.FlagSet) *cobra.Command {
 	cmd := &cobra.Command{
 		Use: "start",
-		Short: `
-			Starts Node daemon. First stopping signal gracefully stops the Node and second terminates it.
-			Passed actions have effect only ont this start and not persisted. 
-		`,
+		Short: `Starts Node daemon. First stopping signal gracefully stops the Node and second terminates it.
+Passed actions have effect only ont this start and not persisted.`,
```
renaynay, line 18:
```suggestion
Options passed on start override configuration options only on start and are not persisted in config.`,
```

cmd/flags_headers.go
liamsi: Checkpoint actually sounds as if this hash is checked along the way and you start syncing with earlier hashes. Is that the case now? If not, starting point is more accurate, or simply:
```suggestion
		"Hex encoded block hash. Used to subjectively intitialize header synchronization",
```

renaynay: Go with Ismail's suggestion!

cmd/flags_core.go
```diff
@@ -0,0 +1,58 @@
...
+
+var (
+	coreRemoteF = "core.remote"
+)
+
+func CoreFlags() *flag.FlagSet {
+	flags := &flag.FlagSet{}
+
+	flags.String(
+		coreRemoteF,
```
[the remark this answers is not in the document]
Wondertan: Flag

cmd/env.go
Wondertan: To preallocate env, so when `WithEnv` is called it does not allocate.

cmd/flags_core.go
Wondertan: That's the code @vgonkivs wrote to parse an address and its is correct.

cmd/flags_headers.go
Wondertan: Ok. Will change it to Flag

cmd/flags_headers.go
Wondertan: I agree though it requires the reader to be aware of the subjective initialization concept. Starting check/point is more readable imo.

cmd/flags_core.go
Wondertan: That's a code that was originally written by @vgonkivs and I think it is correct enough.

Wondertan: If the correct address is passed but without Scheme or Host we just return. Though we can make it even more restrictive and error out if Host or Scheme are missing, as both are needed.

cmd/flags_headers.go
Wondertan: I also changed block => header in suggestion.

Wondertan: If you still think "subjectively initialize" is more readable I apply suggestion.

cmd/flags_node.go
Wondertan: Good catch

cmd/flags_core.go
vgonkivs: Then I'd suggest to rename it to `validateAddress` since now it's not just parsing but also includes additional conditions for validation.

Wondertan: Errored

cmd/flags_headers.go
renaynay: I think subjectively initialize is good.

Wondertan: mkay

cmd/env.go
liamsi: Isn't this an optimization with negligible gain but which adds cognitive overhead by making it more difficult to reason about this? Especially as `env` is also used a local var name everywhere as well.

Wondertan: We can have another name. Agree that having it named as `env` may add additional cognitive overhead and be confusing. If the name is different, then it should be ok

Wondertan: Changed

cmd/env.go
```diff
@@ -0,0 +1,49 @@
...
+	return env.opts
+}
+
+func (env *Env) addOption(opt node.Option) {
+	env.opts = append(env.opts, opt)
+}
+
+type envCtxKey struct{}
+
+var envKey = envCtxKey{}
```
liamsi, line 49: Please add a comment in the code of why this is done. I still think this is unnecessary even with the clearer name now.

Wondertan: It is unnecessary in the same level as preallocating slice when you can just append. Preallocating things is a mindset.

Wondertan: Added more docs and for this case as well

liamsi: > It is unnecessary in the same level as preallocating slice when you can just append. 

a slice, especially if large, is not comparable an empty struct.

Docs looks good. thank you.

Wondertan: Yeah, lets go ban global vars from Go 😁

(its an anti-pattern but there are cases where are they acceptable. An anti-pattern is more for a shared state, in this case envKey has immutable properties)

renaynay: I agree this addition is mostly confusing more than it is helpful. Can you please remove it? This code is used once during node start-up and is not worth our time.

Wondertan: ...

Wondertan: Removed

renaynay (approved):
approoooooved

liamsi (approved):
Approved :)

cmd/env.go
liamsi: Let's agree that it is at least a code-smell in general. And yes, there are exceptions where it makes sense. But I think we can also agree that this is not one of these exceptions.

PR #314, opened 2021-12-23
ci: increase tokens ratio for dupl to fix false positive scenarios

Bidon15:
In addition, removing nolint in tests


Wondertan (dismissed):
👍🏻

Wondertan (approved):

renaynay (approved):

PR #316, opened 2021-12-27
service/header: Fix `TestSubscriber` flakiness 

renaynay:
This PR removes `time.Sleep` from `TestSubscriber` and instead uses the new pubsub method to wait to publish until there's at least 1 other peer listening on the topic.


liamsi (approved):
I didn't verify the WithReadiness function but the change looks good :)

Wondertan (approved):
I verified the WithReadiness func and it fixes the problem

PR #303, opened 2021-12-20
service/header: Refactor `broadcaster` to only be used within `HeaderService` when core (bridge mode) is enabled

renaynay:
This PR contains the last component necessary for **bridge** nodes to be able to listen to core nodes for new block events and translate them to `ExtendedHeader`s.

A more comprehensive explanation can be found [here](https://github.com/celestiaorg/celestia-node/blob/main/docs/adr/adr-003-march2022-testnet.md#headerservice-becomes-main-component-around-which-most-other-services-are-focused).

Resolves second TODO in https://github.com/celestiaorg/celestia-node/issues/251.

PR #317, opened 2021-12-27
service/header: Refactor `HeaderService` to separate distinct components into clearer sub-components

### LINKED ISSUE #287: service/block: `BlockService` not necessary for `bridge` nodes
Since **bridge** nodes relay all block data from the core consensus network to the DA network, they do not need to perform DAS to reconstruct blocks (as they will already have stored the block data once it is received from Core). 

TODO: 
- [ ] ensure that `BlockService` is not added as a component to **bridge** node during start-up

This issue is blocked by #250 as **bridge** node type has not yet been implemented.

renaynay:
This PR does two things: 
1. refactors `HeaderService` in order to make it agnostic to the details of its sub-services. `HeaderService`'s no longer has any function and only contains pointers to its subcomponents for later use once we scope out the RPC component. All ingredients necessary to construct its sub-services are created during node construction. Sub-service lifecycle management is also delegated to the DI application.
2. splits up sub-services a bit better into more isolated functions. For example, rather than having `P2PExchange` start the server-side component (which is necessary for all nodes), but not need it for its exchange component when in **bridge** mode, the server-side component is split out into a separate lifecycle which is present for every node type as all nodes can serve headers, but not all nodes request headers from the same method (p2p vs. directly from core).

This PR is quite large but its necessary moving forward as `HeaderService` is the largest and most complicated component of celestia-node, and many other services are dependent on it, so keeping it clean and its sub-services nicely self-contained is crucial.

Another thing to mention is that this PR includes commits from #277, rendering that PR unnecessary on its own, and should also precede #303. 

Resolves #287


service/header/pubsub_manager.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/libp2p/go-libp2p-core/peer"
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
+)
+
+// PubSubTopic hardcodes the name of the ExtendedHeader
+// gossipsub topic.
+const PubSubTopic = "header-sub"
+
+// topicValidator represents a topic validator function for a pubsub topic.
+type topicValidator func(ctx context.Context, p peer.ID, msg *pubsub.Message) pubsub.ValidationResult
```
Wondertan, line 16: You don't need to define own type for this as you can import it from pubsub

service/header/pubsub_manager.go
```diff
@@ -0,0 +1,66 @@
...
+
+// PubsubManager manages the lifecycle and relationship of header Service
+// with the "header-sub" gossipsub topic.
+type PubsubManager struct {
```
Wondertan, line 20: Call it a P2PSubscriber, as it implements a Subscriber interface and works on the p2p layer.

renaynay: done

node/services/service.go
```diff
@@ -31,14 +29,32 @@ func HeaderSyncer(cfg Config) func(ex header.Exchange, store header.Store) (*hea
...
+	psManager *header.P2PSubscriber,
+	ex header.Exchange,
+) *header.Service {
+
+	headerLifecycles := []header.Lifecycle{
+		syncer,
+		psManager,
+		ex,
+	}
+	// TODO @renaynay: implement adding listener to headerLifecycles if core is enabled
```
renaynay, line 49: FYI this is waiting on broadcaster to be based on the refactoring so it can be implemented.

node/bridge_test.go
```diff
@@ -99,7 +98,6 @@ func TestBridge_P2P_Streams(t *testing.T) {
 
 func TestBridge_WithRemoteCore(t *testing.T) {
 	// TODO(@Wondertan): Fix core
-	t.Skip("Skip until we fix core")
```
renaynay, line 102: what does "fix core" mean @Wondertan

service/header/p2p_exchange_test.go
```diff
@@ -58,45 +58,6 @@ func TestP2PExchange_RequestHeaders(t *testing.T) {
 	}
 }
 
-// TestP2PExchange_Response_Head tests that the P2PExchange instance can respond
-// to an ExtendedHeaderRequest for the chain head.
-func TestP2PExchange_Response_Head(t *testing.T) {
```
renaynay, line 63: this is already covered with previous tests.

service/header/p2p_exchange_test.go
```diff
@@ -137,96 +101,17 @@ func TestP2PExchange_RequestByHash(t *testing.T) {
 	assert.Equal(t, store.headers[reqHeight].Hash(), eh.Hash())
 }
 
-// TestP2PExchange_Response_Single tests that the P2PExchange instance can respond
-// to a ExtendedHeaderRequest for one ExtendedHeader accurately.
-func TestExchange_Response_Single(t *testing.T) {
```
renaynay, line 142: this test is also covered in previous tests

service/header/sync.go
```diff
@@ -70,10 +85,10 @@ func (s *Syncer) Validate(ctx context.Context, p peer.ID, msg *pubsub.Message) p
...
 	}
 
-	// if syncing is still in progress - just ignore the new Header
-	// Syncer will fetch it after anyway
-	select {
-	case <-s.done:
+	// if syncing is still in progress - just ignore the new header as
+	// Syncer will fetch it after anyway, but if syncer is done, append
+	// the header.
+	if atomic.LoadUint64(s.inProgress) == 0 {
```
renaynay, line 91: cc @Wondertan take a look at this.

node/bridge_test.go
Wondertan: Fix core means update to the latest Tendermint

Wondertan: There was an issue in old Tendermint that could panic occasionally.

Wondertan: It was panicking only on CI though, so we need to check how it goes again, as we now rely on the latest changes.

service/header/sync.go
Wondertan: Mk, its fine

service/block/service.go
```diff
@@ -5,44 +5,32 @@ import (
...
-// Service represents the Block service that can be started / stopped on a `Bridge` node.
-// Service contains 4 main functionalities:
-//		1. Fetching "raw" blocks from either Celestia Core or other Celestia Bridge nodes.
-// 		2. Erasure coding the "raw" blocks and producing a DataAvailabilityHeader + verifying the Data root.
-// 		3. Storing erasure coded blocks.
-// 		4. Serving erasure coded blocks to other nodes.
-// TODO @renaynay: add docs about broadcaster
+// Service represents the Block service that can be started / stopped on a `Full` node.
+// Its main function is to reconstruct full blocks from shares via the `FullAvailability`
+// service.
```
Wondertan, line 12: `FullAvailability` service? 
btw, It can reconstruct blocks from DAGService.

service/header/core_exchange.go
```diff
@@ -31,6 +28,20 @@ func NewCoreExchange(fetcher *core.BlockFetcher, dag format.DAGService) *CoreExc
 	}
 }
 
+// Start starts the CoreExchange by ensuring its core client
+// is running.
+func (ce *CoreExchange) Start(context.Context) error {
+	// ensure the core client is running
+	if !ce.fetcher.IsRunning() {
+		return fmt.Errorf("client not running")
+	}
```
Wondertan, line 37: If `fetcher core.BlockFetcher` is started with DI then DI guarantees that fetcher IsRunning, so this check might be useless.

service/header/p2p_exchange.go
```diff
@@ -41,13 +41,12 @@ func NewP2PExchange(host host.Host, peer *peer.AddrInfo, store Store) *P2PExchan
...
 		connected:   make(chan struct{}),
 	}
-	host.SetStreamHandler(exchangeProtocolID, ex.requestHandler)
 	ex.host.Network().Notify(&network.NotifyBundle{ConnectedF: ex.Connected})
 	return ex
 }
 
 func (ex *P2PExchange) Start(ctx context.Context) error {
-	log.Info("starting p2p exchange")
+	log.Info("p2p: starting p2p exchange")
```
Wondertan, line 49: we might need to consider having sub-packages under `header` for exchange and subcribers

service/header/pubsub_manager.go
```diff
@@ -0,0 +1,62 @@
+package header
```
Wondertan, line 1: Pls change the name accordingly to the struct name

service/header/sync.go
```diff
@@ -14,24 +16,35 @@ type Syncer struct {
 	store    Store
 	trusted  tmbytes.HexBytes
 
-	// done is triggered when syncing is finished
-	// it assumes that Syncer is only used once
-	done chan struct{}
+	// inProgress is set to 1 once syncing commences and
+	// is set to 0 once syncing is either finished or
+	// not currently in progress
+	inProgress *uint64
```
Wondertan, line 22: Atomic values should be stored on top of the strcut def, because of https://golang.org/pkg/sync/atomic/#pkg-note-BUG. Also, you should store it not as a pointer but as a value, then you need to do atomic operations by referencing values e.g. `atomic.StoreUint64(&s.inProgress, 1)`

service/header/sync.go
```diff
@@ -61,6 +73,11 @@ func (s *Syncer) Sync(ctx context.Context) {
 	}
 }
 
+// finish indicates Syncer's sync status as no longer in progress.
+func (s *Syncer) finish() {
+	atomic.StoreUint64(s.inProgress, 0)
+}
```
Wondertan, line 79: We should either have both start and finish or don't have them at all

service/header/p2p_exchange.go
renaynay: hmmmmmm yes. Maybe write an issue for it

service/header/core_exchange.go
renaynay: What if it dies after start-up? How does DI catch that?

Wondertan: What if P2PExchange dies after the start and Syncer does not catch this?

Wondertan: And there are other similar cases, I am just not sure if we need to track if the component you depend on is started or else. IMHO, we should aim for `Start` returning nil guarantee if everything is fine. If not, the error will be returned by accessing the method of started but failed component.

Wondertan: Also, how fetcher can die? Don't know

service/header/p2p_exchange.go
Wondertan: https://github.com/celestiaorg/celestia-node/issues/322

service/header/core_exchange.go
renaynay: Okay that's fair. I'll remove the `IsStarted` method

node/core/core.go
```diff
@@ -53,3 +56,11 @@ func Components(cfg Config, loader core.RepoLoader) fxutil.Option {
 func RemoteClient(cfg Config) (core.Client, error) {
 	return core.NewRemote(cfg.RemoteConfig.Protocol, cfg.RemoteConfig.RemoteAddr)
 }
+
+func HeaderExchangeCore(lc fx.Lifecycle, fetcher *core.BlockFetcher, dag ipld.DAGService) *header.CoreExchange {
+	coreEx := header.NewCoreExchange(fetcher, dag)
+	lc.Append(fx.Hook{
```
Wondertan, line 62: This can be removed

node/services/service.go
```diff
@@ -62,12 +75,15 @@ func HeaderExchangeP2P(cfg Config) func(
 	}
 }
 
-func StartHeaderExchangeP2PServer(lc fx.Lifecycle, host host.Host, store header.Store) {
-	ex := header.NewP2PExchange(host, &peer.AddrInfo{}, store)
+// P2PServer creates a new header.P2PServer.
+func P2PServer(lc fx.Lifecycle, host host.Host, store header.Store) *header.P2PServer {
```
Wondertan, line 79: `P2PExchangeServer` and for the constructor itself?

service/header/interface.go
```diff
@@ -33,6 +33,11 @@ type Broadcaster interface {
 // Exchange encompasses the behavior necessary to request ExtendedHeaders
 // from the network.
 type Exchange interface {
+	// Start starts the Exchange lifecycle.
+	Start(ctx context.Context) error
+	// Stop stops the Exchange lifecycle.
+	Stop(ctx context.Context) error
```
Wondertan, line 39: another nit: this can be omitted to shorten the interface while each implementation can be implemented as they need to, e.g CoreExchange does not need them and some other implementations may have `Close()` error instead.

Wondertan (approved):
Well done!(even when I prefer medium-rare more 😄)

service/header/p2p_exchange.go
```diff
@@ -58,111 +57,20 @@ func (ex *P2PExchange) Start(ctx context.Context) error {
 
 		err := ex.host.Connect(ctx, *ex.trustedPeer)
 		if err != nil {
-			log.Errorw("connecting to trusted peer", "err", err)
-			log.Warn("HEADERS WONT BE SYNCHRONIZED - PLEASE RESTART WITH TRUSTED PEER BEING ONLINE")
+			log.Errorw("p2p: connecting to trusted peer", "err", err)
+			log.Warn("p2p: HEADERS WONT BE SYNCHRONIZED - PLEASE RESTART WITH TRUSTED PEER BEING ONLINE")
```
liamsi, line 61: orthogonal to this PR but please remind me: so in case we come online without a trusted peer, the node does not really do anything because it won't sync headers?

service/header/sync.go
```diff
@@ -70,10 +97,10 @@ func (s *Syncer) Validate(ctx context.Context, p peer.ID, msg *pubsub.Message) p
 		return pubsub.ValidationReject
 	}
 
-	// if syncing is still in progress - just ignore the new Header
-	// Syncer will fetch it after anyway
-	select {
-	case <-s.done:
+	// if syncing is still in progress - just ignore the new header as
+	// Syncer will fetch it after anyway, but if syncer is done, append
+	// the header.
```
liamsi, line 102: This is not trivial to understand but I might be just tired lol. The header will not be appended to the store because the syncer will fetch it again anyways sounds inefficient and confusing at first but I guess it makes sense.

service/header/sync.go
```diff
@@ -14,24 +16,35 @@ type Syncer struct {
...
+		exchange:   exchange,
+		store:      store,
+		trusted:    trusted,
+		inProgress: 0, // syncing is not currently in progress
 	}
 }
 
+// Start starts the syncing routine.
+func (s *Syncer) Start(context.Context) error {
+	go s.Sync(context.TODO()) // TODO @Wondertan: leaving this to you to implement in disconnection toleration PR.
```
liamsi, line 37: Is there an issue for that "disconnection toleration PR"?

liamsi (approved):
This looks like great work! I definitely need to check out the code locally soon and go through each line again to get all the context again!

service/header/p2p_exchange.go
Wondertan: For devnet we defined its is acceptable to do so, but it should change so we can download from any peer and trusted peer will only stay for subjective initialization.

service/header/sync.go
Wondertan: Yes it is. There is a todo for that and we will change it. A lot of sacrifices were made to make devnet run.

service/header/sync.go
Wondertan: Yes, can't find it from phone now

node/bridge_test.go
```diff
@@ -126,26 +127,6 @@ func TestBridge_WithRemoteCore(t *testing.T) {
...
-
-	ctx, cancel := context.WithCancel(context.Background())
-	t.Cleanup(cancel)
-
-	err = node.CoreClient.Stop()
-	require.NoError(t, err)
-
-	err = node.Start(ctx)
-	require.Error(t, err, "node: failed to start: client not running")
-}
```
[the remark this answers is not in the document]
renaynay: yes

service/header/sync.go
renaynay: Side note: @Wondertan and I discussed - validation function should be separated out of syncer because that's not the only component using validation logic.

service/header/p2p_exchange.go
liamsi: I hope this is also tracked in an issue. Resolving now.

service/header/sync.go
liamsi: I think that makes sense.

liamsi (approved):

PR #327, opened 2022-01-06
services/header: Refactor `HeaderService` to be responsible for broadcasting new `ExtendedHeader`s to the gossipsub network

### LINKED ISSUE #214: service/header: change Broadcaster from interface to struct

renaynay:
This PR contains the last component necessary for bridge nodes to be able to listen to core nodes for new block events and translate them to `ExtendedHeader`s.

A more comprehensive explanation can be found [here](https://github.com/celestiaorg/celestia-node/blob/main/docs/adr/adr-003-march2022-testnet.md#headerservice-becomes-main-component-around-which-most-other-services-are-focused).

Resolves second TODO in #251.
Resolves #214


service/header/core_listener.go
```diff
@@ -0,0 +1,113 @@
+package header
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/tendermint/tendermint/types"
+)
+
+// CoreListener TODO @renaynay: document
```
liamsi, line 10: Don't you want to include the documentation as part of this PR?

service/header/service.go
```diff
@@ -19,6 +19,10 @@ type Service struct {
 	p2pServer     *P2PExchangeServer
 }
 
+// TODO @renaynay: how will we register core listener on the header Service? It's a part of header service but
+// 	we can't pass it directly to constructor b/c only Bridge nodes provide CoreListener, otherwise it'll always be nil
+// 	maybe we can make it an interface? Still hacky.
```
liamsi, line 24: This could be an issue describing the high level problem as well.

liamsi (commented):
Change looks good

service/header/service.go
Wondertan: > how will we register core listener on the header Service?

But should we? For now, we can only create and Start CoreListener without passing it anywhere, as you probably did. I also think that we can keep this the way it is rn unless we need to access CoreListener somehow. To me, it's like something hanging in the background without the need to be ever touched in the runtime, so putting it inside header service is not of an issue.

service/header/core_listener.go
```diff
@@ -0,0 +1,113 @@
...
+type CoreListener struct {
+	ex     *CoreExchange
```
Wondertan, line 12: From CoreExchnage you need only a fetcher, why not to rcv it directly?

Wondertan: * Lesser coupling
* Eliminates the need to provide CoreExchange as a ptr

Wondertan:
@renaynay, a tip: if you add changelogs to the top the there won't be any conflicts. Also, it makes the ordering nice where top changes are recent ones

service/header/core_listener.go
renaynay: kk

renaynay: actually no, i need exchange's helper functions to generate ExtendedHeader

Wondertan: Maybe then extract this helper func?

Wondertan: It is gross when two independent structs access each others private funcs

service/header/service.go
renaynay:
```
// TODO @renaynay: how will we register core listener on the header Service? It's a part of header service but
// 	we can't pass it directly to constructor b/c only Bridge nodes provide CoreListener, otherwise it'll always be nil
// 	maybe we can make it an interface? Still hacky.

```

saving this comment here ^ 

---------------------------------------------------------------------------
I think the main issue is that all other header sub-services are passed to the header service and contained there while this sub-service is its own special case and is just delegated to the DI app to manage its lifecycle, and isn't tracked anywhere else.

service/header/core_listener.go
renaynay: okey

renaynay: @Wondertan `generateExtendedHeader` is very much tied to fetcher since it accesses a couple different methods on fetcher to construct the ExtendedHeader

Either I can export this method on `CoreExchange` or I can make the method on fetcher instead of `CoreExchange`. Wdyt?

Wondertan: >  I can make the method on fetcher instead of CoreExchange

This will be a dep cycle.

Wondertan: > generateExtendedHeader is very much tied to fetcher

Do you mean tied to CoreExchange?

renaynay: tied to CoreExchange I mean. Should I just export that method?

liamsi (approved):
Looks good and straightforward.

Wondertan (approved):
Approving. My review was here(https://github.com/renaynay/celestia-node/pull/4)

core/fetcher.go
```diff
@@ -26,6 +26,21 @@ func NewBlockFetcher(client Client) *BlockFetcher {
 	}
 }
 
+// GetBlockInfo queries Core for additional block information, like Commit and ValidatorSet.
+func (f *BlockFetcher) GetBlockInfo(ctx context.Context, height *int64) (*types.Commit, *types.ValidatorSet, error) {
+	commit, err := f.Commit(ctx, height)
+	if err != nil {
+		return nil, nil, fmt.Errorf("core/fetcher: getting commit: %w", err)
```
adlerjohn, line 33: Can these errors be more descriptive? E.g. also include the `height`.

core/fetcher.go
```diff
@@ -26,6 +26,21 @@ func NewBlockFetcher(client Client) *BlockFetcher {
...
+	}
+
+	valSet, err := f.ValidatorSet(ctx, height)
```
adlerjohn, line 36: Is this guaranteed to be the same height as the call to `f.Commit`, if `height` is `nil`?

core/fetcher.go
Wondertan: Height will be logged out in the places using the code.

core/fetcher.go
Wondertan: No guarantees. Good catch!

Wondertan: Should be commit.Height instead.

core/fetcher.go
adlerjohn: That's doesn't surprise me. In general I feel like it would be better if errors were completely self-contained, and didn't depend on spooky logic on other places of the code to log out everything. Might just be a Go thing, so 🤷

core/fetcher.go
liamsi: Does `f.ValidatorSet` error if height is nil?

adlerjohn: IIRC if `height` is `nil` then it means "latest." If that's the case, then I don't see any reason why `f.ValidatorSet()` would error when requesting the latest validator set. Might be missing something though?

Wondertan: Not it does not error if nil. Nil is for the latest as John said

liamsi: > Is this guaranteed to be the same height as the call to f.Commit, if height is nil?

So the difference in heights for commit and valset would only occur in the case that between requesting the latest commit and the latest valset a new block "happened". Yeah, better to use commit.Height as Hlib said then.

renaynay: very good catch thankie @adlerjohn :)

liamsi: Just noting that we always call this method with a concrete `&block.Height` taken from a block that was received via a new block event. So this will never be called with latest/`nil` as far as I understand but always with a concrete height here. If that is actually the expectation of how GetBlockInfo is to be used, it would be even better by making this explicit and to error on `nil` instead.

liamsi (approved):
Aaaand another one 💪🏼 
Looks good!

core/fetcher.go
Wondertan: If don't need to pass nil there, then we don't need to have a ptr to int
> it would be even better by making this explicit and to an error on nil instead.

Wondertan: But for consistency, we can keep a ptr still, but make it work with nil as suggested. If don't need this now, it does not mean we won't need it at all.

Wondertan (approved):

core/fetcher.go
renaynay: https://github.com/celestiaorg/celestia-node/issues/336

resolving

liamsi: > If don't need this now, it does not mean we won't need it at all.

Yeah, but it's better to introduce things as they are needed IMO instead of deciding APIs for some hypothetical future that might or might not ever occur.

PR #329, opened 2022-01-07
changelog: release v0.1.1

Wondertan:


liamsi:
There are merge conflicts. Is this a fix that is compatible to 0.1.0? Do we need to update any documentation in networks for this?

Wondertan:
Yes
> Is this a fix that is compatible to 0.1.0?

Do we need to update any documentation in networks for this?
> No

liamsi (commented):
It is not clear to me if this quickfix fixes the 0.1.0 release only. If there are breaking changes (which I would assume on master) then it's probably better to cut a 0.2.0 release?

liamsi:
Why is there a release: https://github.com/celestiaorg/celestia-node/releases/tag/v0.1.1 before this got merged?

Can we have a somewhat consistent workflow with releases core here please?

Wondertan:
No breaking changes here. Master is not included in 0.1.1, only a specific fix.

Wondertan:
We don't have this for such a case. Where we are far away from the release in the master but need to incorporate a fix in past release and in master together.
> consistent workflow

Wondertan:
I want to discuss the problem I've stumbled upon while doing this. The process is not that straightforward and CHANGELOG.md also creates problems.

liamsi:
> No breaking changes here. Master is not included in 0.1.1, only a specific fix.

So this is a fix for 0.1.0, right? Then the flow could be:
create a `0.1.x` branch, create the fixes you want to include onto 0.1.x (which should have the same base as 0.1.0 first), open a PR, review, merge, release. Add thoe fixes to master as well.

liamsi:
I also think we should clean up all these stale branches. 
There only should be 0.1.x (protected), master (protected) and a bunch of feature branches.

liamsi (approved):

renaynay (approved):

renaynay:
> I also think we should clean up all these stale branches. There only should be v0.1.x (protected) and so on, master (protected) and a bunch of feature branches. For clarity: I mean `v0.1.x` as in https://github.com/tendermint/tendermint/tree/v0.35.x for instance. I do not mean we need a v0.1.1 branch.

This is why I like working off a fork :)

Wondertan:
Also, I sent I did a git push --all and recovered all the deleted branched from local store. 🙈

adlerjohn (approved):

liamsi:
And I was wondering where all these branches were coming from lol 🤣

Wondertan:
deleting them

Wondertan:
That was totally accidentally, done.
Need to build a habit to clean branches locally 😄

Issue #23, opened 2021-08-23
shares: Implement the `SharesService`


renaynay:
Implement `SharesService` in package `shares`. 

Implement: 
- [x] type `Share`
- [x] Constructor function for `SharesService`

The `SharesService` contains the following methods / functionality: 
- [x] `Start`
- [x] `Stop`

For `ShareExchange`, a `light` node only requires the ability to request shares and perform sampling (random share requests across random peers), and store the shares they receive, while a `full` node should be able to serve those share requests as well as perform requests.

`ShareStore` only applies to `light` nodes as `full` nodes store full blocks, so if the node is a `full` node, `ShareStore` should be disabled.

---
```go
type Service interface {
	GetShare(ctx context.Context, dah header.DataAvailabilityHeader, row, col int) (Share, error)
	GetShares(context.Context, DataAvailabilityHeader) ([]Shares, error)
	GetSharesByNamespace(context.Context, DataAvailabilityHeader, namespace.ID) ([]Shares, error)
	
	Start(context.Context) error
	Stop(context.Context) error
}
```

- [x] GetSharesByNamespace depend on https://github.com/celestiaorg/celestia-node/issues/5

renaynay:
Should there be a pruning mechanism / threshold for `light` nodes ShareStore? cc @Wondertan @liamsi

liamsi:
I think full storage nodes certainly need pruning (not necessarily for devnet/testnet but for mainnet). 

Light nodes download even less shares. Hence, pruning for light nodes is even less of a priority. That said, if light nodes get subjectively re-initialized we might consider overwriting the existing header (and shares) store. So that way we would have implicit pruning: whenever the light node gets re-initialized it throws away old headers and corresponding shares.

renaynay:
5 November 2021: 

TODO: 

- [ ] Integrate `GetSharesByNamespace` into `shares.Service` inside #170 
- [ ] Implement `GetShares` (ability to reconstruct whole square) -- this is already implemented somewhere, so mostly copy and paste, but better to refactor it a bit for networking, concurrency, and storage optimisations (but this can be left for post Devnet, just need to document it in an issue).

Wondertan:
Reopen as there is still one simple todo left

renaynay:
TODO: 
- [ ] `GetShares`

Wondertan:
@renaynay, GetShares are required for Full Availability so start with this one

Issue #247, opened 2021-06-22
Improve RetrieveBlockData


Wondertan:
## Summary
Current RetrieveBlockData implementation spawns a goroutine per each share in ExtendedDataSquare. In the worst case with the max block 128X128 that is 16384 goroutines. This causes the race detector to stop working in tests, as it has a limit to 8k routines. Furthermore, this a lot of routines for a single operation, and in fact most of the time they are idle. 

Why routines are idle. Shares in a block are addressed and committed with multiple Merkle Trees, so when we request a whole block we need to walk through all the trees. The walking step is: request blob of data by its hash(1), unpack block(2), check if we unpacked more hashes or a share(3), and proceed with unpacked hashes onto the next step recursively(3). Those steps are executed until no more hashes left and we have all the shares. From this, we can see that every walking step is a full roundtrip - network request, unpack, request again, and so on. Now, imagine those 16384 goroutines which walk down the same trees. In practice, those routines mostly wait for every single roundtrip to finish and then they compete to initiate the next roundtrip. Competing, in this case, is useless and we need to avoid it.

Instead of spawning goroutine per share, we should spawn a routine per tree we need to walk through, particularly per every DAHeader root. This way, we don't have any competing routines and only one routine initiates all ther roundtrips. In numbers, for the largest block that would be 128 routines per Row DAHeader root and 256 routines if want to fetch and store all the inner nodes of trees. 
NOTE: All the shares are addressed and committed twice, with two Row and Column Merkle Trees, thus it is not required to traverse both trees to fetch the shares, but again, if we need to store all inner nodes of both trees(cc @adlerjohn) than both tree should be traversed.

## Implementation Details
// TODO

### Action Items
// TODO

### References
As already mentioned, the original spark of the issue is the race detector complaining about exceeding the limit of 8k routines. The solution for that was to detect if the race detector running and simply skipping the test :grimacing:. Later, after some discussions, https://github.com/celestiaorg/lazyledger-core/issues/357 was created. And the solution for it is being implemented here https://github.com/celestiaorg/lazyledger-core/pull/424.

Also, supersede https://github.com/celestiaorg/lazyledger-core/issues/278

## Appendix
There is a plan to remove DAHeader and use only one root hash to address and commit to all the shares. In such a case, we would not need to spawn any routines at all. The routine calling the RetrieveBlockData needs to be blocking anyway and it can then take care of initiating all the round trips.

renaynay:
[...]

PR #333, opened 2022-01-09
service/share: Implement `FullAvailability`

### LINKED ISSUE #179: share: Full Availability
## Background
Right now we have defined `share.Availability` interface in #171. The PR also contains a Light Availability implementation that uses Data Availability sampling to validate that shares are available. Similar to Light Availability, there should be Full Availability which instead of fetching randomized samples, fetches the whole square of shares.

## Design
The new `fullAvailability` structure should satisfy `share.Availability` interface and be placed in `service/share` directory. The `SharesAvailable` method should implement logic similar to `share.Service.GetShares`, basically loading all shares committed to the `share.Root`. 

## Side Notes
Essentially, this implements Full Node syncing logic, if combined with a DASer(#166). So when the DASer uses light light availability it performs sampling over random shares. Respectively, if the DASer depends on Full Availability it will perform fetching of all the shares, basically doing full sync.

renaynay:
Resolves #179 and https://github.com/celestiaorg/celestia-node/issues/23

### DIFF
--- a/node/components.go
+++ b/node/components.go
@@ -19,7 +19,7 @@ func lightComponents(cfg *Config, store Store) fxutil.Option {
 	)
 }
 
-// bridgeComponents keeps all the components as DI options required to build a Full Node.
+// bridgeComponents keeps all the components as DI options required to build a Bridge Node.
 func bridgeComponents(cfg *Config, store Store) fxutil.Option {
 	return fxutil.Options(
 		fxutil.Supply(Bridge),
--- a/node/components.go
+++ b/node/components.go
@@ -39,7 +39,7 @@ func baseComponents(cfg *Config, store Store) fxutil.Option {
 		fxutil.Provide(services.ShareService),
 		// TODO @renaynay: once full node type is defined, add FullAvailability
 		//  to full node and LightAvailability to light node. Bridge node does
-		//  need Availability.
+		//  not need Availability.
 		fxutil.Provide(services.LightAvailability),
 		fxutil.Provide(services.HeaderService),
 		fxutil.Provide(services.HeaderStore),
--- a/node/components.go
+++ b/node/components.go
@@ -19,7 +19,7 @@ func lightComponents(cfg *Config, store Store) fxutil.Option {
 	)
 }
 
-// fullComponents keeps all the components as DI options required to build a Full Node.
+// bridgeComponents keeps all the components as DI options required to built a Full Node.
 func bridgeComponents(cfg *Config, store Store) fxutil.Option {
 	return fxutil.Options(
 		fxutil.Supply(Bridge),
@@ -37,12 +37,15 @@ func baseComponents(cfg *Config, store Store) fxutil.Option {
 		fxutil.Provide(store.Datastore),
 		fxutil.Provide(store.Keystore),
 		fxutil.Provide(services.ShareService),
+		// TODO @renaynay: once full node type is defined, add FullAvailability
+		//  to full node and LightAvailability to light node. Bridge node does
+		//  need Availability.
+		fxutil.Provide(services.LightAvailability),
 		fxutil.Provide(services.HeaderService),
 		fxutil.Provide(services.HeaderStore),
 		fxutil.Provide(services.HeaderSyncer(cfg.Services)),
 		fxutil.Provide(services.P2PSubscriber),
 		fxutil.Provide(services.HeaderP2PExchangeServer),
-		fxutil.Provide(services.LightAvailability), // TODO(@Wondertan): Move to light once FullAvailability is implemented
 		p2p.Components(cfg.P2P),
 	)
 }
--- a/node/components.go
+++ b/node/components.go
@@ -19,7 +19,7 @@ func lightComponents(cfg *Config, store Store) fxutil.Option {
 	)
 }
 
-// bridgeComponents keeps all the components as DI options required to built a Full Node.
+// bridgeComponents keeps all the components as DI options required to build a Full Node.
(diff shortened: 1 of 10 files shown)


node/components.go
```diff
@@ -24,6 +25,7 @@ func bridgeComponents(cfg *Config, store Store) fxutil.Option {
 	return fxutil.Options(
 		fxutil.Supply(Bridge),
 		baseComponents(cfg, store),
+		fxutil.Provide(services.FullAvailability),
```
Wondertan, line 28: We don't need it for Bridge and only for Full. To do this we need to remove Availability from ShareService.

service/share/full_availability.go
```diff
@@ -0,0 +1,31 @@
...
+func NewFullAvailability(get format.NodeGetter) Availability {
+	return &fullAvailability{
+		getter: get,
+	}
+}
+
+// SharesAvailable reconstructs the data committed to the given Root by requesting
+// enough Shares from the network.
+func (fa *fullAvailability) SharesAvailable(ctx context.Context, root *Root) error {
+	_, err := ipld.RetrieveData(ctx, root, fa.getter, rsmt2d.NewRSGF8Codec())
```
Wondertan, line 29: Eventually, we should rework this https://github.com/celestiaorg/celestia-node/issues/247

service/share/share.go
```diff
@@ -129,8 +130,32 @@ func (s *service) GetShare(ctx context.Context, dah *Root, row, col int) (Share,
...
 
-func (s *service) GetShares(context.Context, *Root) ([][]Share, error) {
-	panic("implement me")
+func (s *service) GetShares(ctx context.Context, root *Root) ([][]Share, error) {
+	eds, err := ipld.RetrieveData(ctx, root, s.dag, rsmt2d.NewRSGF8Codec())
+	if err != nil {
+		return nil, err
+	}
+
+	flattened := ipld.ExtractODSShares(eds)
```
Wondertan, line 139: This does looping and there is a loop below, while we can do the conversion in one loop only. This way we avoid zalooping.

service/share/testing.go
```diff
@@ -79,11 +86,16 @@ func NewDAGNet(ctx context.Context, t *testing.T) *DAGNet {
...
+func (dn *DAGNet) RandLightService(n int) (Service, *Root) {
 	dag, root := dn.RandDAG(n)
 	return NewService(dag, NewLightAvailability(dag)), root
 }
 
+func (dn *DAGNet) RandFullService(n int) (Service, *Root) {
+	dag, root := dn.RandDAG(n)
+	return NewService(dag, NewFullAvailability(dag)), root
+}
+
```
Wondertan, line 98: 👍🏻

Wondertan (commented):
The first pAss looks good.

service/share/full_availability.go
```diff
@@ -0,0 +1,31 @@
...
+// NewFullAvailability creates a new full Availability.
+func NewFullAvailability(get format.NodeGetter) Availability {
+	return &fullAvailability{
+		getter: get,
+	}
+}
+
+// SharesAvailable reconstructs the data committed to the given Root by requesting
+// enough Shares from the network.
+func (fa *fullAvailability) SharesAvailable(ctx context.Context, root *Root) error {
```
liamsi, line 28: nit: Is `SharesAvailable` a good name for a method that reconstructs the block? Is this because both "availability" implementations share the same interface or sth?

renaynay: Yes but we're gonna think about the `Availability` interface a bit and how it interacts w/ ShareService soon. It's for another PR tho.

renaynay: cc @Wondertan

liamsi (approved):

node/components.go
```diff
@@ -37,12 +37,15 @@ func baseComponents(cfg *Config, store Store) fxutil.Option {
 		fxutil.Provide(store.Datastore),
 		fxutil.Provide(store.Keystore),
 		fxutil.Provide(services.ShareService),
+		// TODO @renaynay: once full node type is defined, add FullAvailability
+		//  to full node and LightAvailability to light node. Bridge node does
+		//  need Availability.
```
Wondertan, line 42: does need or does not?

service/share/full_availability.go
Wondertan: `SharesAvailable` checks if shares are available. 'FullAvailability' implies full availability, which involves reconstruction.

Wondertan (approved):

node/components.go
renaynay: does not* whoops.

Wondertan (approved):

node/components.go
```diff
@@ -19,7 +19,7 @@ func lightComponents(cfg *Config, store Store) fxutil.Option {
 	)
 }
 
-// fullComponents keeps all the components as DI options required to build a Full Node.
+// bridgeComponents keeps all the components as DI options required to built a Full Node.
```
liamsi, line 22:
```suggestion
// bridgeComponents keeps all the components as DI options required to build a Full Node.
```

node/components.go
```diff
@@ -19,7 +19,7 @@ func lightComponents(cfg *Config, store Store) fxutil.Option {
 	)
 }
 
-// fullComponents keeps all the components as DI options required to build a Full Node.
+// bridgeComponents keeps all the components as DI options required to build a Full Node.
```
Wondertan, line 22:
```suggestion
// bridgeComponents keeps all the components as DI options required to build a Bridge Node.
```

liamsi (approved):
LGTM

Issue #326, opened 2022-01-05
service/header: Syncer should receive topic so it can register its own validator


renaynay:
A pubsub topic can have several validators so Syncer should receive a topic and register its own validator on it.

Issue #242, opened 2021-11-24
service/header: HeaderSync should tolerate network disconnects


Wondertan:
## What
If a light client goes offline for some time and misses a few headers from the HeaderSub network - it does not fall back to regular sync logic. Currently, only full restart helps because we force the light client to sync up to head on start, but we should somehow catch up on runtime.

Issue #235, opened 2021-11-23
service/header: if Heads in Store are colliding choose the one that has more consensus on it.


Wondertan:

renaynay:
Pls more description @Wondertan

Wondertan:
Closing, as we should halt instead, see #365

PR #334, opened 2022-01-10
header: hardening syncing logic

Wondertan:
Previously, we were syncing only once on `Start` and were catching up the chain by Pubsub topic. This PR changes that we can sync multiple times during runtime and if we missed some new Headers from PubSub topic, that will trigger sync again to catch-up. Also, this PR starts to verify new non-adjacent headers from PubSub and caches them to later be applied adjacently to avoid unnecessary request of already rcvd headers

Also fixes a variety of minor things.

Depends on #327 
Closes: #242 
Closes: #326

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,8 @@
 abci-cli
 addrbook.json
 artifacts/*
+celestia
+cel-shed
 coverage.txt
 docs/.vuepress/dist
 docs/_build
--- a/node/services/service.go
+++ b/node/services/service.go
@@ -46,13 +46,13 @@ func HeaderSyncer(cfg Config) func(
 }
 
 // P2PSubscriber creates a new header.P2PSubscriber.
-func P2PSubscriber(lc fx.Lifecycle, sub *pubsub.PubSub) *header.P2PSubscriber {
+func P2PSubscriber(lc fx.Lifecycle, sub *pubsub.PubSub) (*header.P2PSubscriber, header.Subscriber) {
 	p2pSub := header.NewP2PSubscriber(sub)
 	lc.Append(fx.Hook{
 		OnStart: p2pSub.Start,
 		OnStop:  p2pSub.Stop,
 	})
-	return p2pSub
+	return p2pSub, p2pSub
 }
 
 // HeaderService creates a new header.Service.
--- a/node/services/service.go
+++ b/node/services/service.go
@@ -47,13 +47,13 @@ func HeaderSyncer(cfg Config) func(
 }
 
 // P2PSubscriber creates a new header.P2PSubscriber.
-func P2PSubscriber(lc fx.Lifecycle, sub *pubsub.PubSub) (*header.P2PSubscriber, header.Subscriber) {
+func P2PSubscriber(lc fx.Lifecycle, sub *pubsub.PubSub) (*header.P2PSubscriber, header.Subscriber, header.Broadcaster) {
 	p2pSub := header.NewP2PSubscriber(sub)
 	lc.Append(fx.Hook{
 		OnStart: p2pSub.Start,
 		OnStop:  p2pSub.Stop,
 	})
-	return p2pSub, p2pSub
+	return p2pSub, p2pSub, p2pSub
 }
 
 // HeaderService creates a new header.Service.
--- a/node/services/service.go
+++ b/node/services/service.go
@@ -39,6 +39,7 @@ func HeaderSyncer(cfg Config) func(
 		syncer := header.NewSyncer(ex, store, sub, trustedHash)
 		lc.Append(fx.Hook{
 			OnStart: syncer.Start,
+			OnStop:  syncer.Stop,
 		})
 
 		return syncer, nil
--- a/node/services/service.go
+++ b/node/services/service.go
@@ -19,14 +19,14 @@ import (
 )
 
 // HeaderSyncer creates a new header.Syncer.
-func HeaderSyncer(cfg Config) func(lc fx.Lifecycle, ex header.Exchange, store header.Store) (*header.Syncer, error) {
-	return func(lc fx.Lifecycle, ex header.Exchange, store header.Store) (*header.Syncer, error) {
+func HeaderSyncer(cfg Config) func(lc fx.Lifecycle, ex header.Exchange, store header.Store, sub *header.P2PSubscriber) (*header.Syncer, error) {
+	return func(lc fx.Lifecycle, ex header.Exchange, store header.Store, sub *header.P2PSubscriber) (*header.Syncer, error) {
 		trustedHash, err := cfg.trustedHash()
 		if err != nil {
 			return nil, err
 		}
 
-		syncer := header.NewSyncer(ex, store, trustedHash)
+		syncer := header.NewSyncer(ex, store, sub, trustedHash)
 		lc.Append(fx.Hook{
 			OnStart: syncer.Start,
 		})
@@ -36,8 +36,8 @@ func HeaderSyncer(cfg Config) func(lc fx.Lifecycle, ex header.Exchange, store he
 }
 
 // P2PSubscriber creates a new header.P2PSubscriber.
-func P2PSubscriber(lc fx.Lifecycle, sub *pubsub.PubSub, syncer *header.Syncer) *header.P2PSubscriber {
-	p2pSub := header.NewP2PSubscriber(sub, syncer.Validate)
+func P2PSubscriber(lc fx.Lifecycle, sub *pubsub.PubSub) *header.P2PSubscriber {
+	p2pSub := header.NewP2PSubscriber(sub)
 	lc.Append(fx.Hook{
 		OnStart: p2pSub.Start,
 		OnStop:  p2pSub.Stop,
--- a/node/services/service.go
+++ b/node/services/service.go
@@ -23,13 +23,13 @@ func HeaderSyncer(cfg Config) func(
 	lc fx.Lifecycle,
 	ex header.Exchange,
 	store header.Store,
-	sub *header.P2PSubscriber,
+	sub header.Subscriber,
 ) (*header.Syncer, error) {
 	return func(
 		lc fx.Lifecycle,
 		ex header.Exchange,
 		store header.Store,
-		sub *header.P2PSubscriber,
+		sub header.Subscriber,
 	) (*header.Syncer, error) {
 		trustedHash, err := cfg.trustedHash()
 		if err != nil {
@@ -47,23 +47,23 @@ func HeaderSyncer(cfg Config) func(
 }
 
 // P2PSubscriber creates a new header.P2PSubscriber.
-func P2PSubscriber(lc fx.Lifecycle, sub *pubsub.PubSub) (*header.P2PSubscriber, header.Subscriber, header.Broadcaster) {
+func P2PSubscriber(lc fx.Lifecycle, sub *pubsub.PubSub) (header.Subscriber, header.Broadcaster) {
 	p2pSub := header.NewP2PSubscriber(sub)
 	lc.Append(fx.Hook{
 		OnStart: p2pSub.Start,
 		OnStop:  p2pSub.Stop,
 	})
-	return p2pSub, p2pSub, p2pSub
+	return p2pSub, p2pSub
 }
 
 // HeaderService creates a new header.Service.
 func HeaderService(
 	syncer *header.Syncer,
-	p2pSub *header.P2PSubscriber,
+	sub header.Subscriber,
 	p2pServer *header.P2PExchangeServer,
 	ex header.Exchange,
 ) *header.Service {
-	return header.NewHeaderService(syncer, p2pSub, p2pServer, ex)
+	return header.NewHeaderService(syncer, sub, p2pServer, ex)
 }
 
 // HeaderExchangeP2P constructs new P2PExchange for headers.
--- a/node/services/service.go
+++ b/node/services/service.go
@@ -19,8 +19,18 @@ import (
 )
 
 // HeaderSyncer creates a new header.Syncer.
-func HeaderSyncer(cfg Config) func(lc fx.Lifecycle, ex header.Exchange, store header.Store, sub *header.P2PSubscriber) (*header.Syncer, error) {
-	return func(lc fx.Lifecycle, ex header.Exchange, store header.Store, sub *header.P2PSubscriber) (*header.Syncer, error) {
+func HeaderSyncer(cfg Config) func(
+	lc fx.Lifecycle,
+	ex header.Exchange,
+	store header.Store,
+	sub *header.P2PSubscriber,
+) (*header.Syncer, error) {
+	return func(
+		lc fx.Lifecycle,
+		ex header.Exchange,
+		store header.Store,
+		sub *header.P2PSubscriber,
+	) (*header.Syncer, error) {
 		trustedHash, err := cfg.trustedHash()
 		if err != nil {
 			return nil, err
--- a/node/services/service.go
+++ b/node/services/service.go
@@ -47,13 +47,13 @@ func HeaderSyncer(cfg Config) func(
 }
 
 // P2PSubscriber creates a new header.P2PSubscriber.
-func P2PSubscriber(lc fx.Lifecycle, sub *pubsub.PubSub) (header.Subscriber, header.Broadcaster) {
+func P2PSubscriber(lc fx.Lifecycle, sub *pubsub.PubSub) *header.P2PSubscriber {
 	p2pSub := header.NewP2PSubscriber(sub)
 	lc.Append(fx.Hook{
 		OnStart: p2pSub.Start,
 		OnStop:  p2pSub.Stop,
 	})
-	return p2pSub, p2pSub
+	return p2pSub
 }
 
 // HeaderService creates a new header.Service.
--- a/service/header/core_exchange.go
+++ b/service/header/core_exchange.go
@@ -31,6 +31,10 @@ func (ce *CoreExchange) RequestHeader(ctx context.Context, height uint64) (*Exte
 }
 
 func (ce *CoreExchange) RequestHeaders(ctx context.Context, from, amount uint64) ([]*ExtendedHeader, error) {
+	if amount == 0 {
+		return nil, nil
+	}
+
 	log.Debugw("core: requesting headers", "from", from, "to", from+amount)
 	headers := make([]*ExtendedHeader, amount)
 	for i := range headers {
--- a/service/header/header_verify.go
+++ b/service/header/header_verify.go
@@ -21,24 +21,24 @@ import (
 var TrustingPeriod = 168 * time.Hour
 
 // IsExpired checks if header is expired against trusting period.
-func IsExpired(eh *ExtendedHeader) bool {
+func (eh *ExtendedHeader) IsExpired() bool {
 	expirationTime := eh.Time.Add(TrustingPeriod)
 	return !expirationTime.After(time.Now())
 }
 
-// VerifyNonAdjacent validates non-adjacent untrusted header against trusted.
-func VerifyNonAdjacent(trst, untrst *ExtendedHeader) error {
-	if untrst.ChainID != trst.ChainID {
+// VerifyNonAdjacent validates non-adjacent untrusted header against trusted 'eh'.
+func (eh *ExtendedHeader) VerifyNonAdjacent(untrst *ExtendedHeader) error {
+	if untrst.ChainID != eh.ChainID {
 		return &VerifyError{
 			fmt.Errorf(
-				"header belongs to another chain %q, not %q", untrst.ChainID, trst.ChainID,
+				"header belongs to another chain %q, not %q", untrst.ChainID, eh.ChainID,
 			),
 		}
 	}
 
-	if !untrst.Time.After(trst.Time) {
+	if !untrst.Time.After(eh.Time) {
 		return &VerifyError{
-			fmt.Errorf("expected new header time %v to be after old header time %v", untrst.Time, trst.Time),
+			fmt.Errorf("expected new header time %v to be after old header time %v", untrst.Time, eh.Time),
 		}
 	}
 
@@ -50,7 +50,7 @@ func VerifyNonAdjacent(trst, untrst *ExtendedHeader) error {
 	}
 
 	// Ensure that untrusted commit has enough of trusted commit's power.
-	err := trst.ValidatorSet.VerifyCommitLightTrusting(trst.ChainID, untrst.Commit, light.DefaultTrustLevel)
+	err := eh.ValidatorSet.VerifyCommitLightTrusting(eh.ChainID, untrst.Commit, light.DefaultTrustLevel)
 	if err != nil {
 		return &VerifyError{err}
 	}
@@ -58,21 +58,21 @@ func VerifyNonAdjacent(trst, untrst *ExtendedHeader) error {
 	return nil
 }
 
-// VerifyAdjacent validates adjacent untrusted header against trusted.
-func VerifyAdjacent(trst, untrst *ExtendedHeader) error {
-	if untrst.Height != trst.Height+1 {
+// VerifyAdjacent validates adjacent untrusted header against trusted 'eh'.
+func (eh *ExtendedHeader) VerifyAdjacent(untrst *ExtendedHeader) error {
+	if untrst.Height != eh.Height+1 {
 		return ErrNonAdjacent
 	}
 
-	if untrst.ChainID != trst.ChainID {
+	if untrst.ChainID != eh.ChainID {
 		return &VerifyError{
-			fmt.Errorf("header belongs to another chain %q, not %q", untrst.ChainID, trst.ChainID),
+			fmt.Errorf("header belongs to another chain %q, not %q", untrst.ChainID, eh.ChainID),
 		}
 	}
 
-	if !untrst.Time.After(trst.Time) {
+	if !untrst.Time.After(eh.Time) {
 		return &VerifyError{
-			fmt.Errorf("expected new header time %v to be after old header time %v", untrst.Time, trst.Time),
+			fmt.Errorf("expected new header time %v to be after old header time %v", untrst.Time, eh.Time),
 		}
 	}
 
@@ -84,10 +84,10 @@ func VerifyAdjacent(trst, untrst *ExtendedHeader) error {
 	}
 
 	// Check the validator hashes are the same
-	if !bytes.Equal(untrst.ValidatorsHash, trst.NextValidatorsHash) {
+	if !bytes.Equal(untrst.ValidatorsHash, eh.NextValidatorsHash) {
 		return &VerifyError{
 			fmt.Errorf("expected old header next validators (%X) to match those from new header (%X)",
-				trst.NextValidatorsHash,
+				eh.NextValidatorsHash,
 				untrst.ValidatorsHash,
 			),
 		}
--- a/service/header/interface.go
+++ b/service/header/interface.go
@@ -5,14 +5,25 @@ import (
 	"errors"
 	"fmt"
 
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	tmbytes "github.com/tendermint/tendermint/libs/bytes"
 )
+
+// Validator aliases a func that validates ExtendedHeader.
+type Validator = func(context.Context, *ExtendedHeader) pubsub.ValidationResult
 
 // Subscriber encompasses the behavior necessary to
 // subscribe/unsubscribe from new ExtendedHeader events from the
 // network.
 type Subscriber interface {
+	// Subscribe creates long-living for validated ExtendedHeaders.
+	// Multiple Subscriptions can be created.
 	Subscribe() (Subscription, error)
+	// AddValidator registers a Validator for all Subscriptions.
+	// Registered Validators receive ExtendedHeaders before Subscriptions
+	// and decide whether they will receive ExtendedHeaders or not.
+	// Multiple validators can be registered.
+	AddValidator(Validator) error
 }
 
 // Subscription can retrieve the next ExtendedHeader from the
--- a/service/header/interface.go
+++ b/service/header/interface.go
@@ -16,7 +16,7 @@ type Validator = func(context.Context, *ExtendedHeader) pubsub.ValidationResult
 // subscribe/unsubscribe from new ExtendedHeader events from the
 // network.
 type Subscriber interface {
-	// Subscribe creates long-living for validated ExtendedHeaders.
+	// Subscribe creates long-living Subscription for validated ExtendedHeaders.
 	// Multiple Subscriptions can be created.
 	Subscribe() (Subscription, error)
 	// AddValidator registers a Validator for all Subscriptions.
--- a/service/header/interface.go
+++ b/service/header/interface.go
@@ -20,8 +20,8 @@ type Subscriber interface {
 	// Multiple Subscriptions can be created.
 	Subscribe() (Subscription, error)
 	// AddValidator registers a Validator for all Subscriptions.
-	// Registered Validators receive ExtendedHeaders before Subscriptions
-	// and decide whether they will receive ExtendedHeaders or not.
+	// Registered Validators screen ExtendedHeaders for their validity
+	// before they are sent through Subscriptions.
 	// Multiple validators can be registered.
 	AddValidator(Validator) error
 }
--- a/service/header/interface.go
+++ b/service/header/interface.go
@@ -63,7 +63,7 @@ var (
 	// ErrNotFound is returned when there is no requested header.
 	ErrNotFound = errors.New("header: not found")
 
-	// ErrNoHead is returned when Store does not contain Head of the chain.
+	// ErrNoHead is returned when Store is empty (does not contain any known header).
 	ErrNoHead = fmt.Errorf("header/store: no chain head")
 
 	// ErrNonAdjacent is returned when Store is appended with a header not adjacent to the stored head.
--- a/service/header/interface.go
+++ b/service/header/interface.go
@@ -52,8 +52,11 @@ var (
 	// ErrNotFound is returned when there is no requested header.
 	ErrNotFound = errors.New("header: not found")
 
-	// ErrNoHead is returned when Store does not contain Head of the chain,
+	// ErrNoHead is returned when Store does not contain Head of the chain.
 	ErrNoHead = fmt.Errorf("header/store: no chain head")
+
+	// ErrNonAdjacent is returned when Store is appended with a header not adjacent to the stored head.
+	ErrNonAdjacent = fmt.Errorf("header/store: non-adjacent")
 )
 
 // Store encompasses the behavior necessary to store and retrieve ExtendedHeaders
--- a/service/header/local_exchange.go
+++ b/service/header/local_exchange.go
@@ -35,6 +35,9 @@ func (l *LocalExchange) RequestHeader(ctx context.Context, height uint64) (*Exte
 }
 
 func (l *LocalExchange) RequestHeaders(ctx context.Context, origin, amount uint64) ([]*ExtendedHeader, error) {
+	if amount == 0 {
+		return nil, nil
+	}
 	return l.store.GetRangeByHeight(ctx, origin, origin+amount)
 }
 
--- a/service/header/p2p_exchange.go
+++ b/service/header/p2p_exchange.go
@@ -95,6 +95,9 @@ func (ex *P2PExchange) RequestByHash(ctx context.Context, hash tmbytes.HexBytes)
 }
 
 func (ex *P2PExchange) performRequest(ctx context.Context, req *pb.ExtendedHeaderRequest) ([]*ExtendedHeader, error) {
+	if req.Amount == 0 {
+		return make([]*ExtendedHeader, 0), nil
+	}
 	if ex.trustedPeer == "" {
 		return nil, fmt.Errorf("no trusted peer")
 	}
--- a/service/header/p2p_exchange.go
+++ b/service/header/p2p_exchange.go
@@ -124,12 +124,6 @@ func (ex *P2PExchange) performRequest(ctx context.Context, req *pb.ExtendedHeade
 			stream.Reset() //nolint:errcheck
 			return nil, err
 		}
-		// sanity check the header
-		err = header.ValidateBasic()
-		if err != nil {
-			stream.Reset() //nolint:errcheck
-			return nil, err
-		}
 
 		headers[i] = header
 	}
--- a/service/header/store.go
+++ b/service/header/store.go
@@ -1,6 +1,7 @@
 package header
 
 import (
+	"bytes"
 	"context"
 	"strconv"
 	"sync"
@@ -9,7 +10,7 @@ import (
 	"github.com/ipfs/go-datastore"
 	"github.com/ipfs/go-datastore/namespace"
 
-	"github.com/tendermint/tendermint/libs/bytes"
+	tmbytes "github.com/tendermint/tendermint/libs/bytes"
 )
 
 // TODO(@Wondertan): Those values must be configurable and proper defaults should be set for specific node type.
@@ -26,7 +27,7 @@ type store struct {
 	index *heightIndexer
 
 	headLk sync.RWMutex
-	head   bytes.HexBytes
+	head   tmbytes.HexBytes
 }
 
 // NewStore constructs a Store over datastore.
@@ -95,7 +96,7 @@ func (s *store) Head(ctx context.Context) (*ExtendedHeader, error) {
 	}
 }
 
-func (s *store) Get(_ context.Context, hash bytes.HexBytes) (*ExtendedHeader, error) {
+func (s *store) Get(_ context.Context, hash tmbytes.HexBytes) (*ExtendedHeader, error) {
 	if v, ok := s.cache.Get(hash.String()); ok {
 		return v.(*ExtendedHeader), nil
 	}
@@ -145,7 +146,7 @@ func (s *store) GetRangeByHeight(ctx context.Context, from, to uint64) ([]*Exten
 	return headers, nil
 }
 
-func (s *store) Has(_ context.Context, hash bytes.HexBytes) (bool, error) {
+func (s *store) Has(_ context.Context, hash tmbytes.HexBytes) (bool, error) {
 	if ok := s.cache.Contains(hash.String()); ok {
 		return ok, nil
 	}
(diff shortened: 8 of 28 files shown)


service/header/store.go
```diff
@@ -183,9 +184,13 @@ func (s *store) Append(ctx context.Context, headers ...*ExtendedHeader) error {
...
 	verified := make([]*ExtendedHeader, 0, lh)
 	for _, h := range headers {
-		if head.Height == h.Height {
+		if head.Height == h.Height && bytes.Equal(head.Hash(), h.Hash()) {
+			log.Warnw("duplicate header", "hash", head.Hash())
 			continue
 		}
+		// TODO(@Wondertan): Fork choice rule - if two headers are on the same height, but have different hashes
+		//  choose one who has more consensus(power) over it.
+		//  It is not critical to implement this rn, because we can reliably apply the last header who has +2/3.
```
adlerjohn, line 193: Shouldn't we halt if there's two headers with a Tendermint Commit within the unbonding window? And if it's outside the unbonding window, you should just use the one that's an ancestor of the trusted hash?

service/header/sync.go
```diff
@@ -10,62 +11,119 @@ import (
...
+	sub      *P2PSubscriber
 	exchange Exchange
 	store    Store
 	trusted  tmbytes.HexBytes
 
 	// inProgress is set to 1 once syncing commences and
 	// is set to 0 once syncing is either finished or
 	// not currently in progress
-	inProgress uint64
+	inProgress  uint64
```
adlerjohn, line 34: Is there a reason to not use a `bool` here?

service/header/sync.go
```diff
@@ -10,62 +11,119 @@ import (
...
 	// indicate syncing
 	s.syncInProgress()
 	// when method returns, toggle inProgress off
 	defer s.finishSync()
-	// TODO(@Wondertan): Retry logic
 	for {
-		localHead, err := s.getHead(ctx)
+		if ctx.Err() != nil {
+			return
+		}
```
adlerjohn, line 106: It's not immediately clear when this would happen, and if returning here is the right thing to do. Can you document this logic?

service/header/sync.go
```diff
@@ -88,40 +146,66 @@ func (s *Syncer) finishSync() {
...
+		return pubsub.ValidationReject
+	}
+
+	// we are good to go
+	return pubsub.ValidationAccept
 }
 
-// getHead tries to get head locally and if not exists requests trusted hash.
-func (s *Syncer) getHead(ctx context.Context) (*ExtendedHeader, error) {
+// subjectiveHead tries to get head locally and if not exists requests by trusted hash.
```
adlerjohn, line 207:
```suggestion
// subjectiveHead tries to get head locally and if it does not exist, requests it by trusted hash.
```

service/header/sync.go
renaynay: could use an atomic.Value i guess?

service/header/store.go
Wondertan: Ok, so we need centralized halting management in the node that can be triggered by multiple places: header verification and fraud proofs.

Wondertan: Btw, I created [this issue](https://github.com/celestiaorg/celestia-node/issues/235) during devnet as per @adlerjohn's suggestion. Should I close it now?

Wondertan: [An issue ](https://github.com/celestiaorg/celestia-node/issues/365)for halting.

service/header/sync.go
Wondertan: There is no std atomic bool. Using int as bool is idiomatic.

Wondertan: Some people using https://github.com/uber-go/atomic which provides an atomic bool, which is int underneath

service/header/store.go
adlerjohn: Looks like this second issue supersedes the first, so sure.

service/header/sync.go
```diff
@@ -10,62 +11,119 @@ import (
 	tmbytes "github.com/tendermint/tendermint/libs/bytes"
 )
 
+// TODO:
+//  1. Sync protocol for peers to exchange their local heads on connect
+//  2. If we are far from peers but within an unbonding period - trigger sync automatically
+//  3. If we are beyond the unbonding period - request Local Head + 1 header from trusted and hardcoded peer
+//  automatically and continue sync until know head.
```
liamsi, line 18:
```suggestion
//  automatically and continue sync until known head.
```
What does known mean in this context?

service/header/sync.go
```diff
@@ -88,40 +146,66 @@ func (s *Syncer) finishSync() {
...
+		// our subjective view can be far from objective network head(during sync), so we cannot be sure if header is
+		// 100% invalid due to outdated validator set, thus ValidationIgnore and thus 'possibly malicious'
+		log.Warnw("rcvd possibly malicious header",
+			"err", err, "hash", maybeHead.Hash(), "height", maybeHead.Height, "from", p.ShortString())
+		return pubsub.ValidationIgnore
+	}
+
+	if maybeHead.Height > localHead.Height+1 {
+		// we might be missing some headers, so save verified header to pending cache
+		s.pending.Add(maybeHead)
```
liamsi, line 185: Could this be gamed by sending a node a bunch of headers that just keep growing and growing `pending`?

service/header/sync.go
```diff
@@ -174,3 +268,156 @@ func (s *Syncer) syncDiff(ctx context.Context, knownHead, newHead *ExtendedHeade
 
 	return s.store.Append(ctx, newHead)
 }
+
+// getHeaders gets headers from either remote peers or from local cached of headers rcvd by PubSub
+func (s *Syncer) getHeaders(ctx context.Context, start, amount uint64) ([]*ExtendedHeader, error) {
```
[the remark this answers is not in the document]
liamsi: Good point. So I don't think the private method needs a test necessarily. But there definitely should be a test that covers the logic. I just noticed we do not track test coverage anywhere.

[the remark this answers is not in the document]
liamsi: https://github.com/celestiaorg/celestia-node/issues/368

service/header/verify.go
```diff
@@ -6,6 +6,35 @@ import (
 	"time"
 )
 
+// Verify validates trusted header against untrusted.
+// TODO(@Wondertan): Unbonding period!!!
```
liamsi, line 10: Is there an issue about this?

liamsi (commented):
Generally looks very good. Left some comments and agree with @adlerjohn and most of @renaynay's comments.

Wondertan:
I was changing network interfaces, hibernating, suspending for long, and still, my Light Node was syncing without needing to restart.

service/header/sync.go
Wondertan: Added a TODO explaining why it is needed.

service/header/p2p_exchange.go
```diff
@@ -165,12 +169,6 @@ func (ex *P2PExchange) performRequest(ctx context.Context, req *pb.ExtendedHeade
 			stream.Reset() //nolint:errcheck
 			return nil, err
 		}
-		// sanity check the header
-		err = header.ValidateBasic()
-		if err != nil {
-			stream.Reset() //nolint:errcheck
-			return nil, err
-		}
```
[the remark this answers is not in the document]
Wondertan: Because I moved `ValidateBasic` to be always called in `UnmarshalExtendedHeader`

service/header/sync.go
```diff
@@ -174,3 +268,156 @@ func (s *Syncer) syncDiff(ctx context.Context, knownHead, newHead *ExtendedHeade
...
+		out := rs.ranges[0]
+		if !out.Empty() {
+			return out, true
+		}
+
+		rs.ranges = rs.ranges[1:]
+	}
+}
+
+type _range struct {
```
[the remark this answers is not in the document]
Wondertan: added docs

service/header/sync.go
```diff
@@ -10,62 +11,119 @@ import (
...
 	for {
-		localHead, err := s.getHead(ctx)
+		if ctx.Err() != nil {
+			return
+		}
+
+		sbjHead, err := s.subjectiveHead(ctx)
 		if err != nil {
-			log.Errorw("getting head", "err", err)
+			log.Errorw("getting subjective head", "err", err)
```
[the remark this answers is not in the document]
Wondertan: I would prepend additional prefix only for core-related stuff until we have time to extract some `headers` comps into `subpackages` with own log instances.

service/header/sync.go
```diff
@@ -88,40 +146,66 @@ func (s *Syncer) finishSync() {
...
+// validate implements validation of incoming Headers and stores them if they are good.
+func (s *Syncer) validate(ctx context.Context, p peer.ID, maybeHead *ExtendedHeader) pubsub.ValidationResult {
+	localHead, err := s.store.Head(ctx)
+	if err != nil {
+		log.Errorw("getting local head", "err", err)
+		return pubsub.ValidationIgnore // we don't know if header is invalid so ignore
+	}
+
+	if maybeHead.Height <= localHead.Height {
+		log.Warnw("rcvd known header", "from", p.ShortString())
```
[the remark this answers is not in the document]
Wondertan: We have a lot of them because of duplicates on the network rn. WDYM by network latency? If we rcvd a msg once we should not have it again.

service/header/sync.go
Wondertan: Like this question! No, you can't because Verify above ensures that each header coming to pending should have 2/3 power overlap with the local subjective head. One thing I am also considering rn, is to check against the objective head we get from a trusted peer.

service/header/verify.go
Wondertan: I thought there is one, but I can't find it. Note that we are considered to be safe until we fully rely on the trustedPeer.

service/header/sync.go
Wondertan: TBH, I can't parse this as well from a first glance

Wondertan: Those TODO are more like a personal todos, that I forgot to remove

service/header/sync.go
adlerjohn: Can you document your response?

service/header/sync_ranges.go
```diff
@@ -0,0 +1,164 @@
...
+	if ln == 0 {
+		return nil
+	}
+
+	out := r.headers[ln-1]
+	r.headers = r.headers[:ln-1]
+	return out
+}
+
+// Before cuts off all the headers before height 'end' and returns them.
```
renaynay, line 150:
```suggestion
// Before returns all the headers before height 'end'.
```

service/header/sync_ranges.go
```diff
@@ -0,0 +1,164 @@
+package header
+
+import "sync"
+
+// ranges keep non-overlapping and non-adjacent header ranges which used to cache headers
+// Always in ascending heights order
+type ranges struct {
+	ranges []*_range
```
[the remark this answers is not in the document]
Wondertan: This should have on disk offloading instead if gets big enough. Headers getting there are verified and no need to throw them away if exceeding mem limits. Notably, if you sync years of headers from scratch, ranges can grow to be huge and cause OOM, so I agree there should be a limit. Not as part of this PR though

service/header/sync_ranges.go
Wondertan: I also want to emphasize cutting/dereferencing them, meaning that it does not only return but also cuts them.

Wondertan: What word should we use for that?

Wondertan:
Changes since your last reviewed
* Handling of unbonding/trusting period for headers
	* Look at `trustedHead`
* More refactorings and optimizations
* Better logging 
* Better code comments

Wondertan:
Tests are failing due to known #340.

service/header/sync_ranges.go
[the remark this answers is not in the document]
renaynay: Can you please make an issue, and link it in the codebase please as well as here.. It's a serious consideration.

service/header/sync_ranges.go
renaynay: truncate

service/header/store.go
```diff
@@ -182,23 +191,25 @@ func (s *store) Append(ctx context.Context, headers ...*ExtendedHeader) error {
...
+
+			var verErr *VerifyError
+			if errors.As(err, &verErr) {
+				log.Errorw("invalid header",
+					"current height", head.Height,
+					"height", head.Height,
+					"hash", h.Hash(),
+					"reason", err)
+				break
+			}
```
renaynay, line 209: why use errors.As instead of errors.Is here? you're not using `verErr`

renaynay (commented):
need to pass over it again tomorrow.

service/header/store.go
Wondertan: Oh, thank you for catching this. Should use verErr.Reason instead

service/header/sync_ranges.go
Wondertan: Changed the word to truncate and applied suggestion

service/header/sync.go
```diff
@@ -20,147 +23,263 @@ type Syncer struct {
...
+	// NOTE: Pending cache can't be DOSed as we verify above each header against a trusted one.
+	s.pending.Add(maybeHead)
+	// and trigger sync to catch-up
+	s.wantSync()
+	log.Infow("new pending head",
+		"height", maybeHead.Height,
+		"hash", maybeHead.Hash(),
+		"from", p.ShortString())
+	return pubsub.ValidationAccept
 }
```
liamsi, line 254: The logic/flow of `incoming` isn't particularly easy to follow. the comments help a lot though.

service/header/sync.go
```diff
@@ -20,147 +23,263 @@ type Syncer struct {
...
+	if head.Height == newHead.Height {
+		return nil
+	}
+
+	log.Infow("syncing headers", "from", head.Height, "to", newHead.Height)
+	defer log.Info("synced headers")
+
+	start, end := uint64(head.Height)+1, uint64(newHead.Height)
+	for start <= end {
+		amount := end - start + 1
```
liamsi, line 276: Ah, I see now. This is where amount could end up being 0, right?

service/header/sync_ranges.go
```diff
@@ -0,0 +1,148 @@
...
+	// short-circuit if header is from the past
+	if head != nil && head.Height >= h.Height {
+		// TODO(@Wondertan): Technically, we can still apply the header:
+		//  * Headers here are verified, so we can trust them
+		//  * PubSub does not guarantee the ordering of msgs
+		//    * So there might be a case where ordering is broken
+		//    * Even considering the delay(block time) with which new headers are generated
+		//    * But rarely
+		//  Would be still nice to implement
+		log.Warnf("rcvd headers in wrong order")
```
liamsi, line 40:
```suggestion
		log.Warnf("received headers in wrong order")
```
Maybe even add, where the current head is and what was received. 

Orthogonal question: Is it possible to filter the logs per module/package? (e.g. if I want to only see the header related warnings etc).

service/header/sync_ranges.go
```diff
@@ -0,0 +1,148 @@
...
+		out := rs.ranges[0]
+		if !out.Empty() {
+			return out, true
+		}
+
+		rs.ranges = rs.ranges[1:]
+	}
+}
+
+type _range struct {
```
liamsi, line 94: Can we find a name that doesn't start with an underscore here?

service/header/sync_ranges.go
```diff
@@ -0,0 +1,148 @@
...
+	r.lk.Lock()
+	defer r.lk.Unlock()
+	ln := len(r.headers)
+	if ln == 0 {
+		return nil
+	}
+	return r.headers[ln-1]
+}
+
+// Before truncates all the headers before height 'end'.
```
liamsi, line 132: How does it "truncate"? It just returns the headers before "end", right?

service/header/sync_ranges.go
```diff
@@ -0,0 +1,148 @@
...
+
+	if r.Start >= start && r.Start < end {
+		return r, true
+	}
+
+	return nil, false
+}
+
+// Back provides a first non-empty range, while cleaning up empty ones.
+func (rs *ranges) Back() (*_range, bool) {
```
liamsi, line 76: FirstNonEmptyRange  instead of Back?

service/header/sync_ranges.go
```diff
@@ -0,0 +1,148 @@
...
+		rs.ranges = append(rs.ranges, newRange(h))
+
+		// it is possible to miss a header or few from PubSub, due to quick disconnects or sleep
+		// once we start rcving them again we save those in new range
+		// so 'Syncer.getHeaders' can fetch what was missed
+	}
+}
+
+// BackWithin searches for a range within a given height span (start:end].
+func (rs *ranges) BackWithin(start, end uint64) (*_range, bool) {
```
liamsi, line 62: RangeWithIn instead of BackWithin?

service/header/sync.go
```diff
@@ -20,147 +23,263 @@ type Syncer struct {
...
-		err := s.store.Append(ctx, header)
-		if err != nil {
-			log.Errorw("appending store with header from PubSub",
-				"hash", header.Hash().String(), "height", header.Height, "peer", p.ShortString())
-			// TODO(@Wondertan): We need to be sure that the error is actually validation error.
-			//  Rejecting msgs due to storage error is not good, but for now that's fine.
+	return s.incoming(ctx, p, maybeHead)
+}
+
+// incoming process new incoming Headers, validates them and applies/caches if applicable.
```
renaynay, line 188:
```suggestion
// incoming processes new incoming Headers, validates them and stores/caches if applicable.
```

service/header/sync.go
```diff
@@ -20,147 +23,263 @@ type Syncer struct {
...
+func (s *Syncer) incoming(ctx context.Context, p peer.ID, maybeHead *ExtendedHeader) pubsub.ValidationResult {
+	// 1. Try to append. If header is not adjacent/from future - try it for pending cache below
+	err := s.store.Append(ctx, maybeHead)
+	if err == nil {
+		return pubsub.ValidationAccept
+	} else if err != ErrNonAdjacent {
```
renaynay, line 194: switch on err?

service/header/sync.go
```diff
@@ -20,147 +23,263 @@ type Syncer struct {
...
+
+	if head.Height == newHead.Height {
+		return nil
+	}
+
+	log.Infow("syncing headers", "from", head.Height, "to", newHead.Height)
+	defer log.Info("synced headers")
+
+	start, end := uint64(head.Height)+1, uint64(newHead.Height)
+	for start <= end {
```
renaynay, line 275: shouldn't it be `start < end` since end is `newHead.Height` and we already have that header?

service/header/sync.go
```diff
@@ -169,8 +288,49 @@ func (s *Syncer) syncDiff(ctx context.Context, knownHead, newHead *ExtendedHeade
...
+		if r, ok := s.pending.BackWithin(start, end); ok {
+			// first, request everything between start and found range
+			hs, err := s.exchange.RequestHeaders(ctx, start, r.Start-start)
+			if err != nil {
+				return nil, err
+			}
+			out = append(out, hs...)
+			start += uint64(len(hs))
+
+			// than, apply cached range
```
renaynay, line 316:
```suggestion
			// then, apply cached range
```

node/services/service.go
```diff
@@ -19,30 +19,41 @@ import (
...
-func P2PSubscriber(lc fx.Lifecycle, sub *pubsub.PubSub, syncer *header.Syncer) *header.P2PSubscriber {
-	p2pSub := header.NewP2PSubscriber(sub, syncer.Validate)
+func P2PSubscriber(lc fx.Lifecycle, sub *pubsub.PubSub) (*header.P2PSubscriber, header.Subscriber) {
+	p2pSub := header.NewP2PSubscriber(sub)
 	lc.Append(fx.Hook{
 		OnStart: p2pSub.Start,
 		OnStop:  p2pSub.Stop,
 	})
-	return p2pSub
+	return p2pSub, p2pSub
```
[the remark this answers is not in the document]
Wondertan: It implement two different interfaces and should be provided as both. Look at the returned values.

service/header/core_exchange.go
```diff
@@ -31,6 +31,10 @@ func (ce *CoreExchange) RequestHeader(ctx context.Context, height uint64) (*Exte
 }
 
 func (ce *CoreExchange) RequestHeaders(ctx context.Context, from, amount uint64) ([]*ExtendedHeader, error) {
+	if amount == 0 {
+		return nil, nil
+	}
```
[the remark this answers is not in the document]
Wondertan: If it was not a legit case, I would not add this.  It is valid case for syncer

service/header/header_verify.go
```diff
@@ -0,0 +1,106 @@
...
+//
+// Should be significantly less than the unbonding period (e.g. unbonding
+// period = 3 weeks, trusting period = 2 weeks).
+//
+// More specifically, trusting period + time needed to check headers + time
+// needed to report and punish misbehavior should be less than the unbonding
+// period.
+// TODO(@Wondertan): We should request it from the network's state params
+//  or listen for network params changes to always have a topical value.
+var TrustingPeriod = 168 * time.Hour
```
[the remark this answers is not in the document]
Wondertan: Can you explain the use case for a light client? I don't see a point to allow users to change this to unsafe values. Also, look at my TODO which says that we need to fetch it from network-wide params to be always consistent and thus safe.

service/header/header_verify.go
```diff
@@ -0,0 +1,106 @@
...
+	return nil
+}
+
+// VerifyError is thrown on during VerifyAdjacent and VerifyNonAdjacent if verification fails.
+type VerifyError struct {
+	Reason error
+}
+
+func (vr *VerifyError) Error() string {
+	return fmt.Sprintf("header: verify: %s", vr.Reason.Error())
```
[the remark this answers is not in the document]
Wondertan: Well, we return a string here and not an error.

service/header/local_exchange.go
```diff
@@ -35,6 +35,9 @@ func (l *LocalExchange) RequestHeader(ctx context.Context, height uint64) (*Exte
 }
 
 func (l *LocalExchange) RequestHeaders(ctx context.Context, origin, amount uint64) ([]*ExtendedHeader, error) {
+	if amount == 0 {
```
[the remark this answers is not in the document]
Wondertan: Same as above

service/header/sync_ranges.go
Wondertan: Pls look at the code, it also truncates, line 143.

service/header/sync_ranges.go
Wondertan: Too long. I would prefer to keep one word, max two.

node/services/service.go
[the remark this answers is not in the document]
renaynay: It would be nice if you added an fxutil method to do ProvideAs(interface, interface... ) like provide it as several different values.

service/header/sync_ranges.go
Wondertan: Can be also Tail, Dequeue, Last (no need to mention cleaning of empty ones in the name)

service/header/sync.go
```diff
@@ -20,147 +23,263 @@ type Syncer struct {
...
+		triggerSync: make(chan struct{}, 1), // should be buffered
 	}
 }
 
 // Start starts the syncing routine.
-func (s *Syncer) Start(context.Context) error {
-	go s.Sync(context.TODO()) // TODO @Wondertan: leaving this to you to implement in disconnection toleration PR.
+func (s *Syncer) Start(ctx context.Context) error {
+	if s.cancel != nil {
+		return fmt.Errorf("header: Syncer already started")
```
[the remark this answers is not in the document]
Wondertan: We use fmt.Errof consistently everywhere, so in case we need to wrap an error, we can use %w

service/header/sync.go
```diff
@@ -20,147 +23,263 @@ type Syncer struct {
...
-		if err != nil {
-			log.Errorw("appending store with header from PubSub",
-				"hash", header.Hash().String(), "height", header.Height, "peer", p.ShortString())
-			// TODO(@Wondertan): We need to be sure that the error is actually validation error.
-			//  Rejecting msgs due to storage error is not good, but for now that's fine.
+	return s.incoming(ctx, p, maybeHead)
+}
+
+// incoming process new incoming Headers, validates them and applies/caches if applicable.
+func (s *Syncer) incoming(ctx context.Context, p peer.ID, maybeHead *ExtendedHeader) pubsub.ValidationResult {
```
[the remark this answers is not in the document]
Wondertan: Ok. Can you suggest another name? handleIncoming? tryHead?

service/header/sync.go
Wondertan: It can be 0 on line 288

service/header/sync_ranges.go
Wondertan: > is it possible to filter the logs per module/package?

Yes

service/header/sync_ranges.go
Wondertan: Well, Go's std does that for reserved names internally.

service/header/sync_ranges.go
Wondertan: `RangeWithIn` implies the search through all the ranges, but this method only looks at the tail/back and returns it if it is within the range. We don't do searching here, so this guy is specifically for tail/back.

service/header/sync_ranges.go
renaynay: I agree need other word. Prefer `FirstNonEmptyRange`

node/services/service.go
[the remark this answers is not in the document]
Wondertan: True. Will defer to `fx` for now.

.gitignore
```diff
@@ -23,6 +23,8 @@ abci-cli
 addrbook.json
 artifacts/*
 build/*
+celestia
+cel-shed
```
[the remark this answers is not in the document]
Wondertan: Why? those binaries are files and not directories.

service/header/header_verify.go
```diff
@@ -0,0 +1,106 @@
...
+			fmt.Errorf("expected new header time %v to be after old header time %v", untrst.Time, eh.Time),
+		}
+	}
+
+	now := time.Now()
+	if !untrst.Time.Before(now) {
+		return &VerifyError{
+			fmt.Errorf("new header has a time from the future %v (now: %v)", untrst.Time, now),
+		}
+	}
```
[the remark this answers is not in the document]
Wondertan: No, you can't do it in Go, though I understand where are you coming from.

service/header/header_verify.go
```diff
@@ -0,0 +1,106 @@
...
```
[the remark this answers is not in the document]
Wondertan: same here

service/header/interface.go
```diff
@@ -52,8 +52,11 @@ var (
 	// ErrNotFound is returned when there is no requested header.
 	ErrNotFound = errors.New("header: not found")
 
-	// ErrNoHead is returned when Store does not contain Head of the chain,
+	// ErrNoHead is returned when Store does not contain Head of the chain.
```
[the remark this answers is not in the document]
Wondertan: Errors are part of the top-level interface or API. When you work with the interface in Go, you should also see what errors those interfaces return so you can work and check for those. In other words, errors in go are not the implementation detail, but a part of the public interface, e.g io.EOF

service/header/store.go
```diff
@@ -164,6 +171,8 @@ func (s *store) Append(ctx context.Context, headers ...*ExtendedHeader) error {
 	default:
 		return err
 	case ErrNoHead:
+		// TODO(@Wondertan): Should be a separate Init method instead
```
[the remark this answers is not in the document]
Wondertan: And not sure if we need this refactoring soon.

service/header/sync.go
```diff
@@ -20,147 +23,263 @@ type Syncer struct {
...
+		err := s.sub.AddValidator(s.validateMsg)
+		if err != nil {
+			return err
+		}
+	}
+
+	// TODO(@Wondertan): Ideally, this initialization should be part of Init process
+	err := s.initStore(ctx)
+	if err != nil {
+		log.Error(err)
```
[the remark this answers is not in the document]
Wondertan: Returning error here breaks tests in `node_test.go` because trustedHash there is not provided. Even if we provide it, syncer will try to fetch the header by hash, but there is nowhere it can fetch so this always errors, so I making this more like an optional thing.

service/header/sync.go
Wondertan: Here it works. Makes code better. Thanks for noticing this.

service/header/sync.go
Wondertan: It was like this like ten commits before, but I've changed that, so newHead is getting applied internally and not here.

Wondertan: https://github.com/celestiaorg/celestia-node/pull/334/commits/de047fab16c45d6446cb97aa9def49598eb9fa34

Wondertan:
Since the last reviews, I changed the Subscriber interface to contain the AddValidator method and made a reusable mock for it.

node/services/service.go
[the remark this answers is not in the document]
Wondertan: Well, I managed to make it work now! Now, there is only one value returned.

service/header/sync_ranges.go
Wondertan: Are you kidding me? It's not `First` by the code and by the other options I provided...

service/header/sync_ranges.go
Wondertan: LastRangeWithin?

service/header/sync_ranges.go
Wondertan: `LastRange`? We don't return empty ranges for other methods as well, should I append `NonEmpty` to each? Obviously no, this and for this one, we don't need this.

Wondertan:
Test is failing due to https://github.com/celestiaorg/celestia-node/issues/392

.gitignore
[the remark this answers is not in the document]
renaynay: tru

service/header/sync.go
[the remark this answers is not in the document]
renaynay: Hmmm.... still seems weird though.

service/header/sync.go
```diff
@@ -20,147 +23,263 @@ type Syncer struct {
...
+// IsSyncing returns the current sync status of the Syncer.
+func (s *Syncer) IsSyncing() bool {
+	return atomic.LoadUint64(&s.inProgress) == 1
+}
+
+// init initializes if it's empty
+func (s *Syncer) initStore(ctx context.Context) error {
+	_, err := s.store.Head(ctx)
+	switch err {
+	case ErrNoHead:
```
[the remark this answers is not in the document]
renaynay: @Wondertan

service/header/sync.go
Wondertan: I also added more docs to the Syncer as well

service/header/interface.go
```diff
@@ -5,14 +5,25 @@ import (
...
 // subscribe/unsubscribe from new ExtendedHeader events from the
 // network.
 type Subscriber interface {
+	// Subscribe creates long-living for validated ExtendedHeaders.
+	// Multiple Subscriptions can be created.
 	Subscribe() (Subscription, error)
+	// AddValidator registers a Validator for all Subscriptions.
+	// Registered Validators receive ExtendedHeaders before Subscriptions
+	// and decide whether they will receive ExtendedHeaders or not.
+	// Multiple validators can be registered.
```
renaynay, line 25:
```suggestion
	// AddValidator registers a Validator for all Subscriptions.
	// Registered Validators screen ExtendedHeaders for their validity
	// before they are sent through Subscriptions.
	// Multiple validators can be registered.
```

service/header/interface.go
```diff
@@ -5,14 +5,25 @@ import (
...
 )
 
+// Validator aliases a func that validates ExtendedHeader.
+type Validator = func(context.Context, *ExtendedHeader) pubsub.ValidationResult
+
 // Subscriber encompasses the behavior necessary to
 // subscribe/unsubscribe from new ExtendedHeader events from the
 // network.
 type Subscriber interface {
+	// Subscribe creates long-living for validated ExtendedHeaders.
```
renaynay, line 19:
```suggestion
	// Subscribe creates long-living Subscription for validated ExtendedHeaders.
```

renaynay: why long-living?

service/header/sync.go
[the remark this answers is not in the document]
Wondertan: Yes, this is why this initialization part should live elsewhere not cause such issues, but we can live with this for now

service/header/sync.go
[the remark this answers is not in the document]
Wondertan: I can't do anything about this

service/header/sync.go
Wondertan: `start <= end` tells the `s.getHeaders` to include `end` to be gotten and it gets it(with Before which truncates) from the pending cache.

service/header/sync.go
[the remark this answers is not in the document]
Wondertan: And this is related to the comment above

service/header/sync.go
Wondertan: > Shouldn't you also not apply it here?
So by `<=` we are applying it from the pending cache.

Wondertan: Think of a case where we catch-up. We missed a few headers and after rcvd a new pending one, which becomes a Head of pending cache. Then we trigger sync from [subj:pendHead] including pendHead with `<=`, so it is gotten from pending cache.

Wondertan:
@renaynay, I also added a bigger comment on Syncer, might be worth looking for typos and grammar stuff, as you usually do

service/header/header_verify.go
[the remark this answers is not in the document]
liamsi: The use-case is that you can locally set to a stricter trusting period. Why would the network-wide param change once it is set?

service/header/header_verify.go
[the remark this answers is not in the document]
liamsi: Yes, but why?

service/header/sync.go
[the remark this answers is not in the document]
liamsi: ok

service/header/sync_ranges.go
liamsi: headersRange

service/header/header_verify.go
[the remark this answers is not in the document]
Wondertan: Pls look at the error interface. It's Error method returns a string...

service/header/sync_ranges.go
liamsi: I saw that. truncate still sounds like it does something with the headers directly here (instead of truncating a slice). but it's ok.

service/header/header_verify.go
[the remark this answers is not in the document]
Wondertan: Do we need this in this PR?
>The use-case is that you can locally set to a stricter trusting period.
Not sure exactly, but I assume there are no fixed params for the network and everything can be changed during a chain is live. Or do you say you can change unbonding period only by regenesis?
> Why would the network-wide param change once it is set?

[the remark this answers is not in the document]
liamsi: > Do we need this in this PR?

No. But we should capture this in an issue.

> Not sure exactly, but I assume there are no fixed params for the network and everything can be changed during a chain is live. Or do you say you can change unbonding period only by regenesis?

There are also local params per light node. The trusting_period could be safely set to anything < unbondin_period locally. There will be a global unbonding period but that is unlikely to change (and it would require an upgrade).

service/header/header_verify.go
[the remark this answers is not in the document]
liamsi: > Pls look at the error interface. It's Error method returns a string...

I know. But that does not explain why this struct is necessary.

service/header/sync_ranges.go
liamsi: LastRangeWithin is better.

service/header/header_verify.go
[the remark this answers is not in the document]
Wondertan: Ok: https://github.com/celestiaorg/celestia-node/issues/401
> No. But we should capture this in an issue.

Sure, I just don't see why this is a use case users would want. The fact that Tedermint does this, does not tell me why. I don't see this case giving anything to a user from a security standpoint, but only problems as user can harm himself by highering the value without even knowing what it is. Just look at russian th chat, people aped on every config field.
> The trusting_period could be safely set to anything < unbondin_period locally. 

How those upgrades are done? By that TODO I meant that It won't be safe for users to trust TrustingPeriod constant, as the unbonding period might be changed to even lower values by some upgrade system which does not affect the code. It is optimistic to think that we won't forget to update all such values in code after the update and it is even more optimistic to think that all the users will update instantly to the code with new values, thus runtime fetching/listening makes sense. This vision is against user facing configurations for security params, where we give users an ability to shoot into their own legs. However, that's far from now and I am just explaining my POV and why i don't see user based configuration for TrustingPeriod a valid direction, but until automatic approach, it might make sense to allow changing this param from config. 
> and it would require an upgrade

service/header/header_verify.go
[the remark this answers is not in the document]
Wondertan: Did you ask this or why not use `%w`?
> But that does not explain why this struct is necessary.

Do you ask why not this? Can be like this as well, but does not make a huge diff.
```go
return fmt.Errof("header: verify: %w", vr.Reason).Error()
```

service/header/sync_ranges.go
Wondertan: Oh, true, thanks for catching this. My bad that I confused you.

service/header/store_test.go
```diff
@@ -32,7 +32,7 @@ func TestStore(t *testing.T) {
 	err = store.Append(ctx, in...)
 	require.NoError(t, err)
 
-	out, err := store.GetRangeByHeight(ctx, 1, 11)
+	out, err := store.GetRangeByHeight(ctx, 2, 12)
```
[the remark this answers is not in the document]
Wondertan: Where is it failing though?
>why does this test fail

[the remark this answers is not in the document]
Bidon15: it is failing in the checking of the 1st block's height

service/header/sync_ranges.go
Wondertan: Actually, from another look, it makes more sense to call `First`. Sorry for being defensive here and you were right about First

Wondertan: Changed

service/header/sync_ranges.go
Wondertan: Changed

service/header/sync_ranges.go
Wondertan: I am already got used to `_range`, but I changed it to `headerRange`.

Wondertan:
There are two broken tests:
* port already in use
	* not sure what to do with it
* syncer related one:
	* this is type of flakeyness that will be fixed with younger bro PR, as it allows to wait for stuff instead of sleeps

Bidon15 (dismissed):
Okay, let's go 👍

service/header/interface.go
```diff
@@ -5,14 +5,25 @@ import (
 	"errors"
 	"fmt"
 
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	tmbytes "github.com/tendermint/tendermint/libs/bytes"
 )
 
+// Validator aliases a func that validates ExtendedHeader.
+type Validator = func(context.Context, *ExtendedHeader) pubsub.ValidationResult
```
renaynay, line 13: Remember when we couldn't do this 😂

Wondertan: Not really. We might have some misunderstanding. Good that we are on the same page in the end. 

One thing I would change here is to remove dependency on pubsub type, but this is fine for now to keep as otherwise that would overengineering with no real upsides.

service/header/sync.go
```diff
@@ -169,8 +284,49 @@ func (s *Syncer) syncDiff(ctx context.Context, knownHead, newHead *ExtendedHeade
...
 		}
 
-		start += amount
+		start += uint64(len(headers))
+	}
+
+	return nil
+}
+
+// getHeaders gets headers from either remote peers or from local cached of headers rcvd by PubSub
```
renaynay, line 293:
```suggestion
// getHeaders gets headers from either remote peers or from local cache of headers received by PubSub
```

renaynay (commented):
Almost ready.

service/header/interface.go
```diff
@@ -52,8 +63,11 @@ var (
 	// ErrNotFound is returned when there is no requested header.
 	ErrNotFound = errors.New("header: not found")
 
-	// ErrNoHead is returned when Store does not contain Head of the chain,
+	// ErrNoHead is returned when Store does not contain Head of the chain.
```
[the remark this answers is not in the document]
Wondertan: Yes it is returned when there is no headers at all, because no head means no headers.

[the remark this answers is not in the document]
Wondertan: Happy to accept suggestion for more extensive comment.

Wondertan:
@renaynay, responded

service/header/interface.go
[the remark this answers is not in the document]
renaynay:
```suggestion
	// ErrNoHead is returned when Store is empty (does not contain any known header).
```

renaynay:
@Wondertan sir responded

Wondertan:
@renaynay, time to approve the behemoth beast

Bidon15 (dismissed):

Wondertan:
One of the suggestions broke the linter and I had to fix it, therefore I need reapprovals @Bidon15

Wondertan:
Broken tests is a flakey one, but will be fixed by a younger bro

service/header/interface.go
renaynay: resolved.

renaynay (approved):
LEMS (let's end my suffering)

Wondertan:
Lol, no drama pls

adlerjohn (approved):

liamsi (approved):
Looks behemoth to me :D 

But seriously, Thanks for tackling this, thanks fort the thorough reviews and for addressing all feedback.

I really hope we can split PRs like this into smaller PRs in the future.

PR #342, opened 2022-01-12
Fix start order of mocknet

### LINKED ISSUE #173: tests: `TestSubscriber` can hang / sometimes fails due to `context deadline exceeded`

atoulme:
Fixes #173 by separating the creation of the hosts from connecting them, allowing that we configure the gossip protocol on them, otherwise the host doesn't know to gossip with the other host.

### DIFF
--- a/service/header/subscription_test.go
+++ b/service/header/subscription_test.go
@@ -51,35 +51,22 @@ func TestSubscriber(t *testing.T) {
 	err = p2pSub2.Start(context.Background())
 	require.NoError(t, err)
 
-	done := make(chan bool, 1)
 	sub0, err := net.Hosts()[0].EventBus().Subscribe(&event.EvtPeerIdentificationCompleted{})
 	require.NoError(t, err)
 	sub1, err := net.Hosts()[1].EventBus().Subscribe(&event.EvtPeerIdentificationCompleted{})
 	require.NoError(t, err)
 
-	go func() {
-		foundSub0 := false
-		foundSub1 := false
-		for {
-			select {
-			case <-sub0.Out():
-				foundSub0 = true
-			case <-sub1.Out():
-				foundSub1 = true
-			}
-			if foundSub0 && foundSub1 {
-				close(done)
-				return
-			}
-		}
-	}()
 	err = net.ConnectAllButSelf()
 	require.NoError(t, err)
 
-	select {
-	case <-done:
-	case <-ctx.Done():
-		assert.FailNow(t, "timeout waiting for peers to connect")
+	// wait on both peer identification events
+	for i := 0; i < 2; i++ {
+		select {
+		case <-sub0.Out():
+		case <-sub1.Out():
+		case <-ctx.Done():
+			assert.FailNow(t, "timeout waiting for peers to connect")
+		}
 	}
 
 	// subscribe
--- a/service/header/subscription_test.go
+++ b/service/header/subscription_test.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/ipfs/go-datastore"
+	"github.com/libp2p/go-libp2p-core/event"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	mocknet "github.com/libp2p/go-libp2p/p2p/net/mock"
 	"github.com/stretchr/testify/assert"
@@ -51,23 +52,34 @@ func TestSubscriber(t *testing.T) {
 	require.NoError(t, err)
 
 	done := make(chan bool, 1)
+	sub0, err := net.Hosts()[0].EventBus().Subscribe(&event.EvtPeerIdentificationCompleted{})
+	require.NoError(t, err)
+	sub1, err := net.Hosts()[1].EventBus().Subscribe(&event.EvtPeerIdentificationCompleted{})
+	require.NoError(t, err)
+
 	go func() {
+		foundSub0 := false
+		foundSub1 := false
 		for {
-			if net.Hosts()[1].Peerstore().Peers().Len() == 2 && net.Hosts()[0].Peerstore().Peers().Len() == 2 {
-				close(done)
-				break
+			select {
+			case <-sub0.Out():
+				foundSub0 = true
+			case <-sub1.Out():
+				foundSub1 = true
 			}
-			time.Sleep(100 * time.Millisecond)
+			if foundSub0 && foundSub1 {
+				close(done)
+				return
+			}
 		}
 	}()
-	ticker := time.NewTicker(2 * time.Second)
 	err = net.ConnectAllButSelf()
 	require.NoError(t, err)
 
 	select {
 	case <-done:
-	case <-ticker.C:
-		assert.Fail(t, "timeout waiting for peers to connect")
+	case <-ctx.Done():
+		assert.FailNow(t, "timeout waiting for peers to connect")
 	}
 
 	// subscribe
--- a/service/header/subscription_test.go
+++ b/service/header/subscription_test.go
@@ -36,7 +36,6 @@ func TestSubscriber(t *testing.T) {
 	err = p2pSub1.Start(context.Background())
 	require.NoError(t, err)
 
-
 	// get mock host and create new gossipsub on it
 	pubsub2, err := pubsub.NewGossipSub(ctx, net.Hosts()[1],
 		pubsub.WithMessageSignaturePolicy(pubsub.StrictNoSign))
@@ -51,8 +50,25 @@ func TestSubscriber(t *testing.T) {
 	err = p2pSub2.Start(context.Background())
 	require.NoError(t, err)
 
+	done := make(chan bool, 1)
+	go func() {
+		for {
+			if net.Hosts()[1].Peerstore().Peers().Len() == 2 && net.Hosts()[0].Peerstore().Peers().Len() == 2 {
+				close(done)
+				break
+			}
+			time.Sleep(100 * time.Millisecond)
+		}
+	}()
+	ticker := time.NewTicker(2 * time.Second)
 	err = net.ConnectAllButSelf()
-	time.Sleep(400 * time.Millisecond) // wait for peers to finish their connection.
+	require.NoError(t, err)
+
+	select {
+	case <-done:
+	case <-ticker.C:
+		assert.Fail(t, "timeout waiting for peers to connect")
+	}
 
 	// subscribe
 	_, err = p2pSub2.Subscribe()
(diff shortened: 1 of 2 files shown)


service/header/subscription_test.go
```diff
@@ -54,9 +51,16 @@ func TestSubscriber(t *testing.T) {
 	err = p2pSub2.Start(context.Background())
 	require.NoError(t, err)
 
+	err = net.ConnectAllButSelf()
+	time.Sleep(400 * time.Millisecond) // wait for peers to finish their connection.
```
renaynay, line 55: Is this necessary even though we wait for the method to return?

renaynay, line 55:
```suggestion
        require.NoError(t, err)
	time.Sleep(400 * time.Millisecond) // wait for peers to finish their connection.
```

renaynay (commented):
LFG @atoulme 🚀

service/header/subscription_test.go
liamsi: If it is necessary, is there a way to learn about peers finishing up their conns reliably independent of a sleep? If not, then maybe even increase those milliseconds as this might fail on time to time if there is load on the github servers.

service/header/subscription_test.go
```diff
@@ -36,9 +36,6 @@ func TestSubscriber(t *testing.T) {
 	err = p2pSub1.Start(context.Background())
 	require.NoError(t, err)
 
-	// subscribe
-	subscription, err := p2pSub1.Subscribe()
-	require.NoError(t, err)
```
liamsi, line 39: Remove one newline to please the linter (or run `gofmt -s` on the file):
```
service/header/subscription_test.go#L39
File is not `gofmt`-ed with `-s` (gofmt)
```

liamsi (commented):
Thanks for this @atoulme 🙏🏼 

Looks good! Let's merge this after 
- https://github.com/celestiaorg/celestia-node/pull/342#discussion_r782915222 and 
- https://github.com/celestiaorg/celestia-node/pull/342#discussion_r782932001
are addressed.

Wondertan (changes requested):
Hey @atoulme. Thank you for PR! This is our long-living and annoying issue with PubSub testing and it is hard to fix this. Unfortunately, this fix does not fix the issue entirely. We were actively using sleep here sometime before and it does not guarantee test success. GH CI can be overloaded so that even a second of sleep didn't help. The proper solution would be to follow what [go-legs](https://github.com/filecoin-project/go-legs/blob/main/test/util.go#L33) is doing.

service/header/subscription_test.go
Wondertan: We had sleeps before. They do not fix the issue unless the sleep is ridiculously big.

atoulme: Good catch thanks :)

atoulme: I have removed the need for a sleep by watching the peer store of both peers, and making sure that both have connected. I run the check against a timeout of 2 seconds to acknowledge the wonkiness of github actions.

I have run the test in a loop 1000 times to make sure it was a good fix, and it held well.

service/header/subscription_test.go
```diff
@@ -18,7 +18,7 @@ func TestSubscriber(t *testing.T) {
 	defer cancel()
 
 	// create mock network
-	net, err := mocknet.FullMeshConnected(ctx, 2)
```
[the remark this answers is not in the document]
atoulme: The problem here is that there are transient states where the peers are connected, but they don't have both the protocols set. So they would not know for example to talk over gossip. By turning debug logging on, I found for example that the peer connection emitted an error: `error opening stream protocol not supported`.

By digging a bit, I found that one node was connecting with the pubsub protocols, while the other had just started and didn't have protocols set.
So the first fix was to start both p2psubcribers then get them to connect.

The second problem was to acknowledge that nodes were actually in a race condition to connect. I had missed that mocknet was not only registering the nodes, but also connecting them right away, before setting protocols. Hat tip to Rene for pointing that mocknet was maybe the problem.
So the second fix consisted in separating creating the mocknet, and then starting the peer connections after protocols are set so we eliminate this race condition.

Finally, I still had issues, hence the `time.Sleep(400 * time.Millisecond)` which I just refactored into a check, to make sure peers were well connected.
That third scenario came from the fact that `Subscribe` could execute faster than the peer connection. The `Subscribe` call sends a RPC message to the peer requesting subscribing to a topic. The node will loop over its peers to find ones that have the same gossip topics. As the peers are not connected yet, the node will find no peers matching the gossip topic, and drop the message. That's what the code at https://github.com/libp2p/go-libp2p-pubsub/blob/7ef0669764b94ed11e8b5b0128ee408942ebe5eb/pubsub.go#L860 makes me believe.
To fix that third bug, I made sure to wait that peers are correctly both connected to subscribe.

Hopefully, this rationale makes sense. After all those changes, I ran the test in a loop for 1000 occurrences without issues. I didn't check how go-legs fixes this issue, I hope those changes are useful. Cheers!

service/header/subscription_test.go
atoulme: No problem. Fixed.

service/header/subscription_test.go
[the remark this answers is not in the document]
Wondertan: Ok, so the transient state you are talking about is when IdentifyProtocol yet didn't finish a handshake. Makes sense about it causing the problems.

service/header/subscription_test.go
```diff
@@ -54,9 +50,33 @@ func TestSubscriber(t *testing.T) {
 	err = p2pSub2.Start(context.Background())
 	require.NoError(t, err)
 
+	done := make(chan bool, 1)
+	go func() {
+		for {
+			if net.Hosts()[1].Peerstore().Peers().Len() == 2 && net.Hosts()[0].Peerstore().Peers().Len() == 2 {
+				close(done)
+				break
```
Wondertan, line 58: Return?

service/header/subscription_test.go
```diff
@@ -54,9 +50,33 @@ func TestSubscriber(t *testing.T) {
...
+			}
+			time.Sleep(100 * time.Millisecond)
+		}
+	}()
+	ticker := time.NewTicker(2 * time.Second)
```
Wondertan, line 63: Do we need to tick periodcally? time.Timer ?

service/header/subscription_test.go
```diff
@@ -54,9 +50,33 @@ func TestSubscriber(t *testing.T) {
...
+	err = net.ConnectAllButSelf()
+	require.NoError(t, err)
+
+	select {
+	case <-done:
+	case <-ticker.C:
```
Wondertan, line 69: One more even preferred option would be to rely on the global timeout set in the context by using <-ctx.Done()

service/header/subscription_test.go
```diff
@@ -54,9 +50,33 @@ func TestSubscriber(t *testing.T) {
 	err = p2pSub2.Start(context.Background())
 	require.NoError(t, err)
 
+	done := make(chan bool, 1)
+	go func() {
+		for {
+			if net.Hosts()[1].Peerstore().Peers().Len() == 2 && net.Hosts()[0].Peerstore().Peers().Len() == 2 {
```
Wondertan, line 56: We can make this even better by using Host's EventBus and waiting for `event.EvtPeerIdentificationCompleted`. Would be cool if you can implement this instead to remove sleeps and this loop entirely.

Wondertan (commented):
Thank for the detailed rationale. Pls consider suggestions

service/header/subscription_test.go
liamsi: That is a good idea. I think we should open an issue and use this approach where feasible in general (I'm sure there are more places where this could be applied). Might be reasonable to do this in a followup PR later.

service/header/subscription_test.go
atoulme: D'oh

service/header/subscription_test.go
atoulme: neat, will use that

service/header/subscription_test.go
```diff
@@ -54,9 +50,33 @@ func TestSubscriber(t *testing.T) {
...
+		}
+	}()
+	ticker := time.NewTicker(2 * time.Second)
+	err = net.ConnectAllButSelf()
+	require.NoError(t, err)
+
+	select {
+	case <-done:
+	case <-ticker.C:
+		assert.Fail(t, "timeout waiting for peers to connect")
```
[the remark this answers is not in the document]
atoulme: good point

service/header/subscription_test.go
atoulme: ok sure

service/header/subscription_test.go
atoulme: added it, seems to work nicely.

renaynay (dismissed):
LGTM - thanks for all your work @atoulme and your suggestions @Wondertan !

service/header/subscription_test.go
```diff
@@ -54,9 +51,44 @@ func TestSubscriber(t *testing.T) {
...
+			case <-sub0.Out():
+				foundSub0 = true
+			case <-sub1.Out():
+				foundSub1 = true
+			}
+			if foundSub0 && foundSub1 {
+				close(done)
+				return
+			}
+		}
```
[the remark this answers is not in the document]
Wondertan: It is ok two rely on `2` here as we already access net.Hosts()[0] knowing how many hosts we have.

service/header/subscription_test.go
```diff
@@ -54,9 +51,44 @@ func TestSubscriber(t *testing.T) {
...
+	}()
+	err = net.ConnectAllButSelf()
+	require.NoError(t, err)
+
+	select {
+	case <-done:
+	case <-ctx.Done():
+		assert.FailNow(t, "timeout waiting for peers to connect")
+	}
```
Wondertan, line 83: It is not necessary to wait for events in a separate routine and then wait on custom done chan here. Instead, we can simplify this by blocking on event subscriptions here directly with the same looping twice tip above. So in the end we can eliminate:
* goroutine
* bools 
* select 
* done chan

Wondertan (commented):
One more review round, mostly to simplify code

renaynay (approved):

Wondertan (approved):

PR #344, opened 2022-01-12
core: Retry remote core client three times before failing out

renaynay:
Resolves https://github.com/celestiaorg/celestia-node/pull/289.

### DIFF
--- a/core/client.go
+++ b/core/client.go
@@ -3,6 +3,8 @@ package core
 import (
 	"fmt"
 
+	retryhttp "github.com/hashicorp/go-retryablehttp"
+
 	corenode "github.com/tendermint/tendermint/node"
 	"github.com/tendermint/tendermint/rpc/client"
 	"github.com/tendermint/tendermint/rpc/client/http"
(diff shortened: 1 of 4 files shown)


core/client_test.go
```diff
@@ -102,3 +103,21 @@ func TestRemoteClient_StartBlockSubscription_And_GetBlock(t *testing.T) {
...
+	// deliberately stop remote
+	err := remote.Stop()
+	require.NoError(t, err)
+
+	client, err := NewRemote(protocol, ip)
+	require.NoError(t, err)
+
+	_, err = client.Block(context.Background(), nil)
+	require.Error(t, err)
+	assert.ErrorContains(t, err, "giving up after 3 attempt(s)")
```
liamsi, line 122: Can't this be tested against the error type itself instead of via string? (if it is part of their public API)

liamsi (approved):
To nits:
 - checking against an error string
 - yet another dependency introduced for something simple (but it's OK)

core/client_test.go
renaynay: They don't have err types

core/client.go
```diff
@@ -14,9 +16,13 @@ type Client = client.Client
 
 // NewRemote creates a new Client that communicates with a remote Core endpoint over HTTP.
 func NewRemote(protocol, remoteAddr string) (Client, error) {
-	return http.New(
+	httpClient := retryhttp.NewClient()
+	httpClient.RetryMax = 2
```
Wondertan, line 20: Why not 3? God loves Trinity

Wondertan (approved):
:heart:

@jbowen93, wanna test out that reconnection is now working?

core/client.go
renaynay: 3 actually means 4 tries :) bc it's 2 retries plus 1 try.

Issue #62, opened 2021-09-15
build: Network Types


Wondertan:
## Goal
We should have software-level support for different network types, e.g. `celestia-devent` or `celestia-mainnet`

## Current solution
In upcoming https://github.com/celestiaorg/celestia-node/pull/57 this is solved by simply extending Config with a network type field.

## Desired solution
The network type should be hardcoded in binary, as different network types usually run over different software versions, and allowing users to manually change network type is a source of bugs coming from users joining a network with a software version we were not expecting.

Wondertan:
We will also need to define and formalize a list of running network types with supporting bootstrap peers.

renaynay:
Need a hardcoded list of peers for `DEVNET`.

liamsi:
> In upcoming #57 this is solved by simply extending Config with a network type field.

I still think the easiest is to create a directory to contain the genesis & config file. It is common that these are either in a separate repo (example: https://github.com/osmosis-labs/networks) or in the main repo itself ([example1](https://github.com/umee-network/umee/tree/main/networks), [example2](https://github.com/osmosis-labs/osmosis/tree/main/networks)).

> Need a hardcoded list of peers for DEVNET.

I think a set of bootstrap peers could be part of the above directory.

Wondertan:
@liamsi, It has to be part of the binary so its hash commits to network params

jbowen93:
Based on my investigations over in [devops-test/devnet](https://github.com/celestiaorg/devops-test/tree/main/devnet) I'd recommend we only include the `genesis.json` and `gentx` files and *not* the config. 

See my recent PRs that moved from mounting config files to using a flag in [docker-compose](https://github.com/celestiaorg/devops-test/pull/51) and [k8s](https://github.com/celestiaorg/devops-test/pull/53).

If we wanted to provide a recommendation for initial peer setup I think we should use an `env` file.

Wondertan:
Actually, hard coding genesis hash only would work well enough.

Issue #340, opened 2022-01-11
[Bug]: TestRetrieveBlockData exceeds goroutine limit with -race flag


Bidon15:
### Celestia Node version

af0277b15ec17214965ef15083a2cd2f31b9f93f

### OS

macOS 12.0.1

### Steps to reproduce it

go test -v -race -run TestRetrieveBlockData ./ipld -count 1 

### Expected result

TestRetrieveBlockData is passing

### Actual result

TestRetrieveBlockData is failing

### Relevant log output

```shell
race: limit on 8128 simultaneously alive goroutines is exceeded, dying
```

### Notes

Same is reproducible in github action execution

Wondertan:
It is not a race condition. Unfortunately, the race detector has a limit to the number of goroutines possible in the test and this test spawn goroutines deliberately.

Wondertan:
The proper fix is: https://github.com/celestiaorg/celestia-node/issues/247

Wondertan:
The test is now skipped with #388 and should be unskipped once the solution above is implemented.

renaynay:
Closing as #247 has been merged

PR #346, opened 2022-01-13
params: Define Network Types

Wondertan:
* Introduces `build` package to track global build-related things.
* Ability to get Genesis Hash and Bootstrappers for networks
* Removes .gitignore entry for `build`, we anyway don't use it
* Removes Network field from Config
	* Networks should not be configured, otherwise, a user changing their networks may see weird bugs, like by syncing from one network and then from another in the same Store/Repo.
	* Not a breaking change

The use of it will be shown in the next PR.

Closes: #62

### DIFF
--- a/CHANGELOG-PENDING.md
+++ b/CHANGELOG-PENDING.md
@@ -12,6 +12,7 @@ Month, DD, YYYY
 
 ### FEATURES
 
+- [build: Define Network Types #346](https://github.com/celestiaorg/celestia-node/pull/346) [@Wondertan](https://github.com/Wondertan)
 - [feat(cmd): give a birth to cel-shed and p2p key utilities #281](https://github.com/celestiaorg/celestia-node/pull/281) [@Wondertan](https://github.com/Wondertan)
 - [feat(cmd|node): MutualPeers Node option and CLI flag #280](https://github.com/celestiaorg/celestia-node/pull/280) [@Wondertan](https://github.com/Wondertan)
 - [node: enhance DI allowing overriding of dependencies](https://github.com/celestiaorg/celestia-node/pull/290) [@Wondertan](https://github.com/Wondertan)
--- a/CHANGELOG-PENDING.md
+++ b/CHANGELOG-PENDING.md
@@ -31,6 +31,7 @@ Month, DD, YYYY
 
 ### BUG FIXES
 
+- [fix(header/service): #339 race](https://github.com/celestiaorg/celestia-node/pull/343) [@Wondertan](https://github.com/Wondertan)
 - [Properly fetch Validators from Core and two more fixes #328](https://github.com/celestiaorg/celestia-node/pull/328) [@Wondertan](https://github.com/Wondertan)
 - [header] Added missing `err` value in ErrorW logging calls. @jbowen93
 - [service/block, node/p2p] [Fix race conditions in TestExtendedHeaderBroadcast and TestFull_P2P_Streams.](https://github.com/celestiaorg/celestia-node/pull/288) [@jenyasd209](https://github.com/jenyasd209)
--- a/CHANGELOG-PENDING.md
+++ b/CHANGELOG-PENDING.md
@@ -12,7 +12,7 @@ Month, DD, YYYY
 
 ### FEATURES
 
-- [build: Define Network Types #346](https://github.com/celestiaorg/celestia-node/pull/346) [@Wondertan](https://github.com/Wondertan)
+- [params: Define Network Types #346](https://github.com/celestiaorg/celestia-node/pull/346) [@Wondertan](https://github.com/Wondertan)
 - [feat(cmd): give a birth to cel-shed and p2p key utilities #281](https://github.com/celestiaorg/celestia-node/pull/281) [@Wondertan](https://github.com/Wondertan)
 - [feat(cmd|node): MutualPeers Node option and CLI flag #280](https://github.com/celestiaorg/celestia-node/pull/280) [@Wondertan](https://github.com/Wondertan)
 - [node: enhance DI allowing overriding of dependencies](https://github.com/celestiaorg/celestia-node/pull/290) [@Wondertan](https://github.com/Wondertan)
--- a/build/bootstrap.go
+++ b/build/bootstrap.go
@@ -1,0 +1,15 @@
+package build
+
+// BootstrapFor reports multiaddress of bootstrap peers for a given network.
+func BootstrapFor(net Network) ([]string, error) {
+	if err := net.Validate(); err != nil {
+		return nil, err
+	}
+
+	return bootstrapList[net], nil
+}
+
+// NOTE: Everytime we add a new long-running network, it's genesis hash has to be added it here.
+var bootstrapList = map[Network][]string{
+	DevNet: {}, // TODO(@Wondertan): Set multiaddress once bootstrap peers are set up.
+}
--- a/build/bootstrap.go
+++ b/build/bootstrap.go
@@ -1,6 +1,6 @@
 package build
 
-// Bootstrappers reports multiaddress of bootstrap peers for a current network.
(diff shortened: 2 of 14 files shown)


build/bootstrap.go
```diff
@@ -0,0 +1,20 @@
+package build
+
+// Bootstrappers reports multiaddress of bootstrap peers for a current network.
+func Bootstrappers() []string {
+	return bootstrapList[network] // network is guarantee to be valid
+}
+
+// BootstrappersFor reports multiaddress of bootstrap peers for a given network.
+func BootstrappersFor(net Network) ([]string, error) {
```
renaynay, line 9: Same question here re: how this will be used in practice.

build/genesis.go
```diff
@@ -0,0 +1,20 @@
...
+// GenesisFor reports a hash of a genesis block for a given network.
+func GenesisFor(net Network) (string, error) {
+	if err := net.Validate(); err != nil {
+		return "", err
+	}
+
+	return genesisList[net], nil
+}
+
+// NOTE: Everytime we add a new long-running network, it's genesis hash has to be added it here.
```
renaynay, line 17:
```suggestion
// NOTE: Every time we add a new long-running network, its genesis hash has to be added here.
```

build/networks.go
```diff
@@ -0,0 +1,51 @@
+package build
+
+import "errors"
+
+// GetNetwork returns the network of a current build.
```
renaynay, line 5:
```suggestion
// GetNetwork returns the network of the current build.
```

build/networks.go
```diff
@@ -0,0 +1,51 @@
...
+func GetNetwork() Network {
+	return network
+}
+
+// DefaultNetwork sets a default network for Celestia Node.
+var DefaultNetwork = DevNet
+
+// NOTE: Everytime we add a new long-running network, it has to be added it here.
```
renaynay, line 13:
```suggestion
// NOTE: Every time we add a new long-running network, it has to be added here.
```

build/networks.go
```diff
@@ -0,0 +1,51 @@
...
+
+// Validate the network.
+func (n Network) Validate() error {
+	if _, ok := networksList[n]; !ok {
+		return ErrInvalidNetwork
+	}
+	return nil
+}
+
+// A strict list of networks.
```
renaynay, line 33:
```suggestion
// networksList is a strict list of all known long-standing networks.
```

build/networks.go
```diff
@@ -0,0 +1,51 @@
...
+var networksList = map[Network]struct{}{
+	DevNet: {},
+}
+
+// A used network within a build.
+// Can be set with 'ldflags'
+var network Network
```
renaynay, line 40:
```suggestion
// network is the currently used network within a build.
// It can be set with 'ldflags'.
var network Network
```

renaynay: Idflags?

build/networks.go
```diff
@@ -0,0 +1,51 @@
...
+
+// this init ensures `network` is always set and correct
```
renaynay, line 42:
```suggestion
// init ensures `network` is always set and correct
```

renaynay (changes requested):

build/bootstrap.go
```diff
@@ -0,0 +1,20 @@
...
+	if err := net.Validate(); err != nil {
+		return nil, err
+	}
+
+	return bootstrapList[net], nil
+}
+
+// NOTE: Everytime we add a new long-running network, it's genesis hash has to be added it here.
+var bootstrapList = map[Network][]string{
+	DevNet: {}, // TODO(@Wondertan): Set multiaddress once bootstrap peers are set up.
```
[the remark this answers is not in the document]
Wondertan: * We have to have bootstrap peers for devnet, otherwise, our DA does not exist and all the full/light nodes are living in our own world disconnected from others. This is what is happening rn with Russians. 
* As you know, diff between `devnet` and `testnet` is ephemeral and I doubt that for `testnet` we will set up everything from scratch without reusing devnet's infra. So once we start `testnet`, we will most likely just rename `DevNet` to `TestNet`.

build/networks.go
Wondertan: [This](https://github.com/celestiaorg/celestia-node/blob/main/Makefile#L3) and [this](https://github.com/celestiaorg/celestia-node/blob/main/Makefile#L14). The way we pass build-time flags in Go...

renaynay: can you add: 

```
It can be set with `-ldflags` during build.
```

Wondertan: ldflags implies during build. Dash ahead does make any difference.

build/genesis.go
```diff
@@ -0,0 +1,20 @@
+package build
+
+// Genesis reports a hash of a genesis block for a current network.
+func Genesis() string {
+	return genesisList[network] // network is guarantee to be valid
+}
+
+// GenesisFor reports a hash of a genesis block for a given network.
+func GenesisFor(net Network) (string, error) {
```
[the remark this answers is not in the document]
Wondertan: That's manual access, not a programmatic one.
> Why can't users just look at the genesisList in the codebase? 

TBH, I don't have a `real` reason for this to happen, and I just like making things public if they are :smile:. Does not hurt to have it.

Wondertan:
build -> params

Wondertan:
Tests are failing due to known #340.

node/p2p/p2p.go
```diff
@@ -18,9 +18,6 @@ type Config struct {
 	// NoAnnounceAddresses - Addresses the P2P subsystem may know about, but that should not be announced/advertised,
 	// as undialable from WAN
 	NoAnnounceAddresses []string
-	// TODO(@Wondertan): This should be a built-time parameter. See https://github.com/celestiaorg/celestia-node/issues/62
-	// Networks stands for network name, e.g. celestia-devnet.
-	Network string
 	// TODO(@Wondertan): This should be a built-time parameter. See https://github.com/celestiaorg/celestia-node/issues/63
```
liamsi, line 21:
```suggestion
	// TODO(@Wondertan): This should be a build-time parameter. See https://github.com/celestiaorg/celestia-node/issues/63
```

params/networks.go
```diff
@@ -0,0 +1,51 @@
...
+	return network
+}
+
+// DefaultNetwork sets a default network for Celestia Node.
+var DefaultNetwork = DevNet
+
+// NOTE: Every time we add a new long-running network, it has to be added here.
+const (
+	// DevNet or devnet-2 according to celestiaorg/networks
+	DevNet Network = "devnet-2"
```
liamsi, line 16: IMO, this should rather be in a config file instead of code. Which also means it would be set at run-time. For bridge-nodes the config could be shared between core/app and node (chain-id).

Wondertan: Initialized node can't be easily switched to another network over the same root folder/Store. If we give this opportunity in the config that means we should support it and test it, otherwise users will do this and break/corrupt their Store and Node. Another safe option we discussed with Rene is to pass the network name on init and initialize with it.

CHANGELOG-PENDING.md
```diff
@@ -12,6 +12,7 @@ Month, DD, YYYY
 
 ### FEATURES
 
+- [build: Define Network Types #346](https://github.com/celestiaorg/celestia-node/pull/346) [@Wondertan](https://github.com/Wondertan)
```
renaynay, line 15:
```suggestion
- [params: Define Network Types #346](https://github.com/celestiaorg/celestia-node/pull/346) [@Wondertan](https://github.com/Wondertan)
```

renaynay (dismissed):
already looked through this a few times, LGTM but address @liamsi 's comments.

renaynay (approved):

liamsi (approved):

PR #348, opened 2022-01-13
node/services: use genesis if trusted hash is not set

### LINKED ISSUE #324: node: BridgeNode should get trusted hash automatically
Currently, in **remote mode** we are _manually_ passing a trusted hash from the Core/App client for the Celestia Bridge Node

In case we are creating an **embedded core type** for the Celestia Bridge Node, it makes more to get/pass trusted hash automatically.
This approach can be utilised for the remote type, too.

Wondertan:
Required for Bridge Node, but also works for Light and in future for Full. This won't be safe for them though, once we stop relying on a trusted peer for the whole syncing, but until then it is safe, as we trust a trusted peer.

Based/depends on #346 
Closes #324

### DIFF
--- a/node/services/config.go
+++ b/node/services/config.go
@@ -6,6 +6,8 @@ import (
 	"github.com/libp2p/go-libp2p-core/peer"
 	"github.com/multiformats/go-multiaddr"
 	tmbytes "github.com/tendermint/tendermint/libs/bytes"
+
+	"github.com/celestiaorg/celestia-node/build"
 )
 
 type Config struct {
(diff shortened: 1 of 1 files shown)


node/services/config.go
```diff
@@ -40,5 +42,8 @@ func (cfg *Config) trustedPeer() (*peer.AddrInfo, error) {
 }
 
 func (cfg *Config) trustedHash() (tmbytes.HexBytes, error) {
+	if cfg.TrustedHash == "" {
```
adlerjohn, line 45: Seems generally leaky to compare against `""` (which will be a run-time failure if the implementation of `TrustedHash` changes) instead of using a method on `cfg` or `TrustedHash` (which would be a compile-time failure). But whatever.

adlerjohn (commented):
Is this meant to be merged into `hlib/build`?

node/services/config.go
Wondertan: Wait? It is not leaky to compare string against empty string? If the implementation of `TrustedHash` changes then there will be a compile-time error that you cannot compare against an empty string.

Wondertan:
> Is this meant to be merged into hlib/build?
No, directly to master. I based it on hlib/build to make the diff clear

renaynay (approved):

liamsi (approved):
🤙🏼

Issue #350, opened 2022-01-13
swamp: investigate panic when using both mocknet.Connect and header.Start


Bidon15:
In case ConnectPeers from Swamp.Network and [header.Start](https://github.com/celestiaorg/celestia-node/blob/v0.1.1/service/header/p2p_exchange.go#L55) is utilised to connect nodes between each other will panic due to channel being closed from [dial_sync.go(go-libp2p-swarm lib) in func DialLock](https://github.com/libp2p/go-libp2p-swarm/blob/v0.5.3/dial_sync.go#L120), which is calling ad.decref() that closes the channel

Note: It's not clear why the version was v0.5.3 for libp2p-swarm. Need to understand why it's not latest though

Bidon15:
So on Ubuntu, the panic is reproducible as in macOS, so it's getting even more 🧐 

I've tired bumping up the version for libp2p-swarm from `v0.5.3` to `v0.6.0` and higher and got issues with deps overall, so this is not a solution if we are not planning to upgrade libp2p lib itself in the nearest future. 

Some of the logs for deps upgrade issues
```sh
github.com/ipfs/go-ipfs-blockstore
../../../../pkg/mod/github.com/ipfs/go-ipfs-blockstore@v0.1.6/blockstore.go:123:32: not enough arguments in call to bs.datastore.Get
	have (datastore.Key)
	want (context.Context, datastore.Key)
```

In CI both go 1.16 and 1.17 are working with no hick-ups, so I'll leave CI topic

PR #367, opened 2022-01-19
service/header: remove start/stop from P2PExchange

Bidon15:
Change config to log warnings with missing trusted peer
In HeaderExchangeP2P func we are creating an addr for the host instead of making it in the removed connect
Removed store in p2pExchange

Now, instead of handling connection ourselves, we are delegating this to libp2p itself and only creating NewStream in `performRequest` func in `p2p_exchange.go`

Solves the issue `panic of closing a closed channel` #350 

Co-authored-by: @Wondertan

### DIFF
--- a/CHANGELOG-PENDING.md
+++ b/CHANGELOG-PENDING.md
@@ -18,6 +18,7 @@ Month, DD, YYYY
 
 ### IMPROVEMENTS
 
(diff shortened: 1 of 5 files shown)


renaynay (approved):
utACK but LGTM

PR #369, opened 2022-01-21
docs/adr: State Interaction

renaynay:
This PR introduces a new ADR detailing celestia-node's state interaction with celestia-core/app for the March 2022 testnet.

### DIFF
--- a/docs/adr/adr-004-state-interaction.md
+++ b/docs/adr/adr-004-state-interaction.md
@@ -99,7 +99,7 @@ An idea for how `P2PAccess` could be implemented can be found below.
 ```go
 type P2PAccess struct {
 	host libhost.Host
-    // P2PAccess will still use lens.ChainClient's methods
+	// P2PAccess will still use lens.ChainClient's methods
 	// under the hood, but will use libp2p transport instead
 	// of RPC communication with the Core endpoint.
    cc  *lens.ChainClient 
--- a/docs/adr/adr-004-state-interaction.md
+++ b/docs/adr/adr-004-state-interaction.md
@@ -121,20 +121,20 @@ func (pa *P2PAccess) AccountBalance(ctx context.Context, acct Account) (*Balance
 	if err != nil {
 		return nil, err
     }
-    
-	// write query to stream
+
+    // write query to stream
     err := stream.Write(query)
     if err != nil {
 		return nil, err
     }
 
-	// read resp from stream
+    // read resp from stream
     resp, err := stream.Read(buf)
     if err != nil {
         return nil, err		    
     }	
 
-	// unmarshal balance from response
+    // unmarshal balance from response
 	var bal Balance
     err = bal.Unmarshal(buf)	
 	if err != nil {
@@ -161,13 +161,13 @@ func (pa *P2PAccess) SubmitTx(ctx context.Context, tx Tx) (*TxResponse, error) {
 		return nil, err
     }
 
-	// read response from stream
+    // read response from stream
     resp, err := stream.Read(buf)
     if err != nil {
         return nil, err
     }	
 	
-	// get txResponse from response
+    // get txResponse from response
 	var tx Tx
 	err = tx.Unmarshal(buf)
 	if err != nil {
--- a/docs/adr/adr-004-state-interaction.md
+++ b/docs/adr/adr-004-state-interaction.md
@@ -48,15 +48,14 @@ type StateAccessor interface {
 }
 ```
 
-`StateAccessor` will have 3 separate implementations under the hood: 
+`StateAccessor` will have 3 separate implementations under the hood, the first of which will be outlined in this ADR: 
 1. **CORE**: RPC connection with a celestia-core endpoint handed to the node upon initialisation
 (*required for this iteration*) 
-2. **P2P**: discovery of a **bridge** node or other node that can provide state and submit messages via libp2p to be 
-relayed to celestia-core (*nice-to-have for this iteration*)
+2. **P2P**: discovery of a **bridge** node or other node that is able to provide state (*nice-to-have for this iteration*)
 3. **LOCAL**: eventually, **full** nodes will be able to provide state to the network, and can therefore execute and 
 respond to the queries without relaying queries to celestia-core (*to be scoped out and implemented in later iterations*)
 
-### `CoreAccess`
+### 1. Core Implementation of `StateAccessor`: `CoreAccess`
 
 `CoreAccess` will be the direct RPC connection implementation of `StateAccessor`. It will use the [lens implementation of ChainClient](https://github.com/strangelove-ventures/lens/blob/main/client/chain_client.go#L23)
 under the hood. `lens.ChainClient` provides a nice wrapper around the Cosmos-SDK APIs. 
@@ -92,91 +91,22 @@ func (ca *CoreAccess) SubmitTx(ctx context.Context, tx Tx) (TxResponse, error) {
 }
 ```
 
-### `P2PAccess`
+### 2. P2P Implementation of `StateAccessor`: `P2PAccess`
 
-An idea for how `P2PAccess` could be implemented can be found below.
+While it is not necessary to detail how `P2PAccess` will be implemented in this ADR, it will still conform to the 
+`StateAccessor` interface, but instead of being provided a core endpoint to connect to via RPC and using `lens.ChainClient`
+for state-related queries, `P2PAccess` will perform service discovery of state-providing nodes in the network and perform
+the state queries via libp2p streams.
 
 ```go
 type P2PAccess struct {
-	host libhost.Host
-	// P2PAccess will still use lens.ChainClient's methods
-	// under the hood, but will use libp2p transport instead
-	// of RPC communication with the Core endpoint.
-   cc  *lens.ChainClient 
+    // for discovery and communication
+    // with state-providing nodes
+    host host.Host
+    // for managing keys
+    keybase        keyring.Keyring
+    keyringOptions []keyring.Option
 }
-
-func NewP2PAccess(host libhost.Host, cc *lens.ChainClient) *P2PAccess {
-    return &P2PAccess{
-		host: host,
-		cc: cc,
-    }   	
-}
-
-func (pa *P2PAccess) AccountBalance(ctx context.Context, acct Account) (*Balance, error) {
-    // passing in lens.ChainClient here to help construct the query
-	query := acctBalanceQuery(pa.cc, acct)
-	
-    // open connection with peer who serves State
-    stream, err := pa.ConnectToStateProvider()   
-	if err != nil {
-		return nil, err
-    }
-
-    // write query to stream
-    err := stream.Write(query)
-    if err != nil {
-		return nil, err
-    }
-
-    // read resp from stream
-    resp, err := stream.Read(buf)
-    if err != nil {
-        return nil, err		    
-    }	
-
-    // unmarshal balance from response
-	var bal Balance
-    err = bal.Unmarshal(buf)	
-	if err != nil {
-		return nil, err
-    }
-	
-    return bal, nil
-}
-
-// where Tx implements sdk.Msg
-func (pa *P2PAccess) SubmitTx(ctx context.Context, tx Tx) (*TxResponse, error) {
-    // passing in lens.ChainClient here to help construct the query 
-	query := submitTxQuery(pa.cc, tx)
-	
-    // open connection with peer who serves State
-    stream, err := pa.ConnectToStateProvider()
-    if err != nil {
-        return nil, err
-    }
-    
-    // write query to stream
-    err := stream.Write(query)
-    if err != nil {
-		return nil, err
-    }
-
-    // read response from stream
-    resp, err := stream.Read(buf)
-    if err != nil {
-        return nil, err
-    }	
-	
-    // get txResponse from response
-	var tx Tx
-	err = tx.Unmarshal(buf)
-	if err != nil {
-		return nil, err
-}
-	
-    return TxResponse, nil
-}
-
 ```
 
 ### `StateProvider`
--- a/docs/adr/adr-004-state-interaction.md
+++ b/docs/adr/adr-004-state-interaction.md
@@ -77,7 +77,7 @@ func NewCoreAccess(cc *lens.ChainClient) (*CoreAccess, error) {
 
 // where Account wraps account information and Balance 
 func (ca *CoreAccess) AccountBalance(ctx context.Context, acct Account) (Balance, error) {
-	coins, err := da.cc.QueryBalanceWithAddress(acct.address)
+	coins, err := da.cc.QueryBalanceWithAddress(acct.Address)
 	if err != nil {
 		return err
     }   
--- a/docs/adr/adr-004-state-interaction.md
+++ b/docs/adr/adr-004-state-interaction.md
@@ -111,7 +111,7 @@ type P2PAccess struct {
 }
 ```
 
-### `StateProvider`
+#### `StateProvider`
 
 A **bridge** node will run a `StateProvider` (server-side of `P2PAccessor`). The `StateProvider` will be responsible for
 relaying the state-related queries through to its trusted celestia-core node.
--- a/docs/adr/adr-004-state-interaction.md
+++ b/docs/adr/adr-004-state-interaction.md
@@ -1,0 +1,189 @@
+# ADR #004: Celestia-Node State Interaction for March 2022 Testnet
+
+<hr style="border:3px solid gray"> </hr>
+
+## Authors
+
+@renaynay
+
+## Changelog
+
+* 2022-01-14: initial draft
+
+<hr style="border:2px solid gray"> </hr>
+
+## Context
+
+Currently, celestia-node lacks a way for users to submit a transaction to celestia-core as well as access its own state
+and public state of others.
+
+## Decision
+
+Both celestia **light** and **full** nodes will run `StateService`. 
+`StateService` will be responsible for managing the RPC endpoints provided by `StateAccessor`.
+
+`StateAccessor` will be an interface for interacting with celestia-core in order to retrieve account-related information
+as well as submit transactions.
+
+### `StateService`
+
+`StateService` will expose several higher-level RPC endpoints for users to access the methods provided by the 
+`StateAccessor`.
+
+```go
+type StateService struct {
+    accessor StateAccessor
+}
+``` 
+
+### `StateAccessor`
+
+`StateAccessor` is defined by the two methods listed below and may be expanded in the future to include more methods 
+related to accounts and transactions in later iterations.
+
+```go
+type StateAccessor interface {
+    func AccountBalance(context.Context, Account) (Balance, error)
+    func SubmitTx(context.Context, Tx) error
+}
+```
+
+`StateAccessor` will have 3 separate implementations under the hood: 
+1. **CORE**: RPC connection with a celestia-core endpoint handed to the node upon initialisation
+(*required for this iteration*) 
+2. **P2P**: discovery of a **bridge** node, and then establish direct RPC connection (*nice-to-have for this iteration*)
+3. **LOCAL**: eventually, **full** nodes will be able to provide state to the network, and can execute the
+
+### `CoreAccess`
+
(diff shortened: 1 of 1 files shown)


docs/adr/adr-004-state-interaction.md
```diff
@@ -0,0 +1,173 @@
...
+
+Both celestia **light** and **full** nodes will run `StateService`. 
+`StateService` will be responsible for lifecycle of the given implementation of `StateAccessor`.
+
+`StateAccessor` will be an interface for interacting with celestia-core in order to retrieve account-related information
+as well as submit transactions.
+
+### `StateService`
+
+`StateService` will manage the lifecycle of the `StateAccessor` and also expose several higher-level RPC endpoints for
```
Wondertan, line 29: We now landed upon DI managing lifecycles.

docs/adr/adr-004-state-interaction.md
```diff
@@ -0,0 +1,173 @@
...
+``` 
+
+### `StateAccessor`
+
+`StateAccessor` is defined by the two methods listed below and may be expanded in the future to include more methods 
+related to accounts and transactions in later iterations.
+
+```go
+type StateAccessor interface {
+    func AccountBalance(context.Context, Account) (Balance, error)
```
Wondertan, line 45: We should still think through this. Accessing Celestia's balance is not enough and there are many more fields, like delegation that light clients need to access.

docs/adr/adr-004-state-interaction.md
```diff
@@ -0,0 +1,173 @@
...
+    func SubmitTx(context.Context, Tx) error
+}
+```
+
+`StateAccessor` will have 3 separate implementations under the hood: 
+1. **DIRECT**: direct RPC connection with a celestia-core endpoint handed to the node upon initialisation
+(**required for this iteration**) 
+2. **P2P**: discovery of a **bridge** node, and then establish direct RPC connection (*nice-to-have for this iteration*)
```
Wondertan, line 53: No RPC in p2p impl:
* Unsecure,
	* With RPC you connect without a secured tunnel and proper pubkey handshake, just with an address
* We will need to rewrite it once 'local' is implemented.
	* Because 'local' won't be sharing state over Tendermint's RPC
* Ugly, the code will be super ugly to make this work. Especially addresses convertions. 
* We will waste more time making this work and fixing all the caveats, while simple req/resp with libp2p is straightforward. IDK, why this frightens everybody like something hard when it's not. It won't be any different from HeaderExchange besides discovery which we need anyway

docs/adr/adr-004-state-interaction.md
renaynay: ok ok

docs/adr/adr-004-state-interaction.md
```diff
@@ -0,0 +1,173 @@
...
+3. **LOCAL**: eventually, **full** nodes will be able to provide state to the network, and can execute the
+`StateAccessor`'s locally (will be implemented in later iterations)
+
+### `DirectAccess`
```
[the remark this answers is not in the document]
renaynay: yes core is better name.

docs/adr/adr-004-state-interaction.md
```diff
@@ -0,0 +1,189 @@
...
+    // passing in lens.ChainClient here to help construct the query
+	query := acctBalanceQuery(pa.cc, acct) 
+	
+	// open connection with peer who serves State
+    stream, err := pa.ConnectToStateProvider()   
+	if err != nil {
+		return nil, err
+    }
+    
+	// write query to stream
```
evan-forbes, line 123: formatting comments?

docs/adr/adr-004-state-interaction.md
```diff
@@ -0,0 +1,189 @@
...
+    if err != nil {
+        return nil, err		    
+    }	
+
+	// unmarshal balance from response
+	var bal Balance
+    err = bal.Unmarshal(buf)	
+	if err != nil {
+		return nil, err
+    }
```
evan-forbes, line 140: same formatting issues as above

evan-forbes (approved):
LGTM :+1: 

PS: I love ADRs with lots of code

docs/adr/adr-004-state-interaction.md
```diff
@@ -0,0 +1,189 @@
...
+
+An idea for how `P2PAccess` could be implemented can be found below.
+
+```go
+type P2PAccess struct {
+	host libhost.Host
+    // P2PAccess will still use lens.ChainClient's methods
+	// under the hood, but will use libp2p transport instead
+	// of RPC communication with the Core endpoint.
+   cc  *lens.ChainClient
```
Wondertan, line 103: Pls, provide proof that this is feasible. This is the part of the point which I said will be ugly and has multiple caveats that need to be double-checked before doing the implementation.

docs/adr/adr-004-state-interaction.md
```diff
@@ -0,0 +1,189 @@
...
+    }	
+	
+	// get txResponse from response
+	var tx Tx
+	err = tx.Unmarshal(buf)
+	if err != nil {
+		return nil, err
+}
+	
+    return TxResponse, nil
```
Wondertan, line 175: What's the point of having streams here if you say that we will do everything with the lens methods above, but with hijacked p2p transport?

docs/adr/adr-004-state-interaction.md
Wondertan: Otherwise, we shouldn't use the lens for the p2p approach and only for RPC.

Wondertan (changes requested):

docs/adr/adr-004-state-interaction.md
Wondertan: Ok, so I checked and lens [uses](https://github.com/strangelove-ventures/lens/blob/main/client/chain_client.go#L27) tm's RPCClient or `core.Client` in our codebase, so to make this happen we would need to implement `core.Client` over P2P.

docs/adr/adr-004-state-interaction.md
renaynay: Yes, later iterations please.

docs/adr/adr-004-state-interaction.md
```diff
@@ -0,0 +1,189 @@
...
+    func AccountBalance(context.Context, Account) (Balance, error)
+    func SubmitTx(context.Context, Tx) error
+}
+```
+
+`StateAccessor` will have 3 separate implementations under the hood: 
+1. **CORE**: RPC connection with a celestia-core endpoint handed to the node upon initialisation
+(*required for this iteration*) 
+2. **P2P**: discovery of a **bridge** node, and then establish direct RPC connection (*nice-to-have for this iteration*)
+3. **LOCAL**: eventually, **full** nodes will be able to provide state to the network, and can execute the
```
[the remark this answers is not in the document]
renaynay: sry fat fingered and deleted a whole line

docs/adr/adr-004-state-interaction.md
renaynay: coding in markdown is no fun 😓

docs/adr/adr-004-state-interaction.md
renaynay: Which I think would be better. We don't need to fully implement all methods either, just those we need as we go along

docs/adr/adr-004-state-interaction.md
renaynay: WE can also manually create the queries.

docs/adr/adr-004-state-interaction.md
Wondertan: > Which I think would be better

Any arguments supporting this point?

Wondertan: Note that, I am not completely against this way, I just don't see we should go this way. If we merge this point as part of the ADR, everyone on the team should be aligned with this trajectory so pls explain.

docs/adr/adr-004-state-interaction.md
Wondertan: If we go with lens and core.Client then the P2PAccess implementation won't have such logic, but it will use the lens, only core.Client over p2p will have those and not the P2PAcess, so there is discrepancy within the code. You didn't mention the code as pesudo-code and its seems like your goal is use it further, so It makes sense to check if the code is actually workable. Maybe not compile, but at least is doing the right thing.

docs/adr/adr-004-state-interaction.md
Wondertan: I understand that you want to do less work by reusing lens, but I am afraid about this being even more work and overhead than just using p2p streams directly. Devil is always in detail, especially when you are trying to connect two different worlds into something workable.

docs/adr/adr-004-state-interaction.md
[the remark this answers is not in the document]
renaynay: can i resolve

docs/adr/adr-004-state-interaction.md
renaynay: Okay looking into it now, I think it would be easier to just not use the chain client in p2p, but that would require us doing a lot of other work from scratch instead of shoehorning things around the ChainClient implementation. I guess I just liked the fact all the methods we need are already implemented, including keyring management parts.

docs/adr/adr-004-state-interaction.md
renaynay: resolved?

docs/adr/adr-004-state-interaction.md
renaynay: resolved?

docs/adr/adr-004-state-interaction.md
renaynay: resolved?

docs/adr/adr-004-state-interaction.md
renaynay: resolved?

docs/adr/adr-004-state-interaction.md
renaynay: I honestly don't want to detail the p2p implementation here right now bc we don't need to implement it for this iteration anyway and it might change later on. So should I just remove the code snippets for now?

docs/adr/adr-004-state-interaction.md
```diff
@@ -0,0 +1,191 @@
...
+    func AccountBalance(context.Context, Account) (Balance, error)
+    func SubmitTx(context.Context, Tx) error
+}
+```
+
+`StateAccessor` will have 3 separate implementations under the hood: 
+1. **CORE**: RPC connection with a celestia-core endpoint handed to the node upon initialisation
+(*required for this iteration*) 
+2. **P2P**: discovery of a **bridge** node or other node that can provide state and submit messages via libp2p to be 
+relayed to celestia-core (*nice-to-have for this iteration*)
```
liamsi, line 55: only critique here: I think this is difficult to understand if people did not attend the calls. I think it is fine now but if we start that implementation I think we should also describe this in more detail in a separate ADR. Same with the LOCAL approach below.

docs/adr/adr-004-state-interaction.md
```diff
@@ -0,0 +1,191 @@
...
+### `P2PAccess`
+
+An idea for how `P2PAccess` could be implemented can be found below.
+
+```go
+type P2PAccess struct {
+	host libhost.Host
+    // P2PAccess will still use lens.ChainClient's methods
+	// under the hood, but will use libp2p transport instead
+	// of RPC communication with the Core endpoint.
```
liamsi, line 104:
```suggestion
	// P2PAccess will still use lens.ChainClient's methods
	// under the hood, but will use libp2p transport instead
	// of RPC communication with the Core endpoint.
```

liamsi (commented):
A quick pass with minor comments/questions. Will do a more thorough review later/during the weekend.

docs/adr/adr-004-state-interaction.md
```diff
@@ -0,0 +1,191 @@
...
+	return da.cc.SendMsg(ctx, tx)
+}
+```
+
+### `P2PAccess`
+
+An idea for how `P2PAccess` could be implemented can be found below.
+
+```go
+type P2PAccess struct {
```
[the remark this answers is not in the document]
renaynay: No, the `CoreAccess` implementation is documented above this section.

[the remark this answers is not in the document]
renaynay: I can document it better though.

docs/adr/adr-004-state-interaction.md
renaynay: I will just write in the ADR that we are only outlining the core access impl for htis ADR as the p2p impl is still a bit up in the air. I think outlining the interface that it should adhere to is good enough and the blurb about discovering a state provider. Same w local.

docs/adr/adr-004-state-interaction.md
renaynay: removed the code for the most part, pls check again.

docs/adr/adr-004-state-interaction.md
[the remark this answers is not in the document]
renaynay: Pls check now?

docs/adr/adr-004-state-interaction.md
```diff
@@ -0,0 +1,121 @@
...
+
+### `StateAccessor`
+
+`StateAccessor` is defined by the two methods listed below and may be expanded in the future to include more methods 
+related to accounts and transactions in later iterations.
+
+```go
+type StateAccessor interface {
+    func AccountBalance(context.Context, Account) (Balance, error)
+    func SubmitTx(context.Context, Tx) error
```
liamsi, line 47: It is a bit unclear which type `Tx` refers to here. In the core impl it says `where Tx implements sdk.Msg`. Is that generally the assumption in the iface here too?

docs/adr/adr-004-state-interaction.md
```diff
@@ -0,0 +1,121 @@
...
+
+func NewCoreAccess(cc *lens.ChainClient) (*CoreAccess, error) {
+	return &CoreAccess{
+        cc: cc,		
+    }, nil   
+}
+
+// where Account wraps account information and Balance 
+func (ca *CoreAccess) AccountBalance(ctx context.Context, acct Account) (Balance, error) {
+	coins, err := da.cc.QueryBalanceWithAddress(acct.address)
```
liamsi, line 80: Or `acct.Address()`?
```suggestion
	coins, err := da.cc.QueryBalanceWithAddress(acct.Address)
```

docs/adr/adr-004-state-interaction.md
```diff
@@ -0,0 +1,121 @@
...
+	return da.cc.SendMsg(ctx, tx)
+}
+```
+
+### 2. P2P Implementation of `StateAccessor`: `P2PAccess`
+
+While it is not necessary to detail how `P2PAccess` will be implemented in this ADR, it will still conform to the 
+`StateAccessor` interface, but instead of being provided a core endpoint to connect to via RPC and using `lens.ChainClient`
+for state-related queries, `P2PAccess` will perform service discovery of state-providing nodes in the network and perform
+the state queries via libp2p streams.
```
liamsi, line 99:
```suggestion
the state queries via libp2p streams. More details of the p2p implementation will be described in a separate dedicated ADR.
```

docs/adr/adr-004-state-interaction.md
```diff
@@ -0,0 +1,121 @@
...
+``` 
+
+### `StateAccessor`
+
+`StateAccessor` is defined by the two methods listed below and may be expanded in the future to include more methods 
+related to accounts and transactions in later iterations.
+
+```go
+type StateAccessor interface {
+    func AccountBalance(context.Context, Account) (Balance, error)
```
liamsi, line 46: From a quick look at the lens impl it seems this returns `sdk.Coins` which could be several balances (e.g. IBC coins or if the chain has several coins/tokens). We should consider returning only the Celes balance here.

docs/adr/adr-004-state-interaction.md
renaynay: Yes. I can document it in the interface docs

docs/adr/adr-004-state-interaction.md
renaynay: Like I described here? https://github.com/celestiaorg/celestia-node/pull/369/files#diff-51474eb9a6d4dcb6b29018687c8bfc9f38bda0259cac4562475c3a9a98c1568bR85

liamsi: Makes sense, yes. But even there you are not returning a `Coin`. Not sure if we have a `Balance` type anywhere. Or is that an alias?

liamsi: OK, I'm simply assuming balance is just a type alias or an abstraction for Coin. Ok for me to resolve this thread.

liamsi (approved):
Ismail says LGMT

liamsi (approved):

docs/adr/adr-004-state-interaction.md
renaynay: Yes it will be an alias / abstraction.

adlerjohn (approved):

docs/adr/adr-004-state-interaction.md
Wondertan: I am ok to remove it!

docs/adr/adr-004-state-interaction.md
Wondertan: Okay.

For this, I bet, we would need to implement a custom key management api.
> including keyring management parts.

docs/adr/adr-004-state-interaction.md
Wondertan: Yes

docs/adr/adr-004-state-interaction.md
```diff
@@ -0,0 +1,123 @@
...
+}
+```
+
+### `StateProvider`
+
+A **bridge** node will run a `StateProvider` (server-side of `P2PAccessor`). The `StateProvider` will be responsible for
+relaying the state-related queries through to its trusted celestia-core node.
+
+A new `core.StateFetcher` will be implemented over the `core.Client` in the `core` package. The `core.StateFetcher` will
+contain methods that wrap communication with the `core.Client`.
```
Wondertan, line 120: Is it supposed to be used by `StateProvider`? What methods the `core.StateFetcher` will have? Will it wrap lens methods? I see that you want to follow the existing pattern with `core.Fetcher` and I am ok with it, but in this case, it does not seem necessary as the lens is already a wrapper that we can instantiate once in `node/core` and reuse where needed, like in `StateProvider` without additional wrappers. WDYT?

docs/adr/adr-004-state-interaction.md
```diff
@@ -0,0 +1,123 @@
...
+    // for discovery and communication
+    // with state-providing nodes
+    host host.Host
+    // for managing keys
+    keybase        keyring.Keyring
+    keyringOptions []keyring.Option
+}
+```
+
+### `StateProvider`
```
Wondertan, line 114:
```suggestion
#### `StateProvider`
```

As it seems more like a sub header to `P2PAccess`

docs/adr/adr-004-state-interaction.md
Wondertan: Yes

docs/adr/adr-004-state-interaction.md
renaynay: So you want StateProvider to use chainclient under the hood to communicate with its RPC client? I think that's totally fine 👍🏻  Pushing changes now.

Wondertan: Yes, unless we implement block execution for full nodes.

docs/adr/adr-004-state-interaction.md
Wondertan: Resolved by my side. Resolve it via the button as an ack that you read this.

Wondertan (approved):
_Wonderting_ conclusions are: LGTBM

docs/adr/adr-004-state-interaction.md
renaynay: resolved?

evan-forbes (approved):

liamsi (approved):
Still Approving Anyways

docs/adr/adr-004-state-interaction.md
renaynay: i assume resolved, sorry need to resolve this to merge!

PR #375, opened 2022-01-24
node/config: add WithBootstrapPeers option

### LINKED ISSUE #372: node: add WithBootstrapPeers Option
WithBootstrapPeers option is needed to adjust Bootstrap param to the node.
This will be a handy option for those, who want to adjust their own set of bootstrap peers directly in the node and not through the Config struct.

In addition, It is required in order for the Test Swamp to initialise the light node with 1+ connection to Bridge Nodes

Another way is to use Connect from mocknet, but needs some discussion (I guess) 

The same has been done before 
#280 and #299

Bidon15:
Add tests for WithMutualPeers and WithBootstrapPeers for both bridge/light types
Remove redundancy in WithMockClient test

This is needed for the swamp to initialise the light node with 2+ connections to bridge nodes
Resolves #372

### DIFF
--- a/CHANGELOG-PENDING.md
+++ b/CHANGELOG-PENDING.md
@@ -12,7 +12,6 @@ Month, DD, YYYY
 
 ### FEATURES
 
-- [node/config: add WithBootstrapPeers option #375](https://github.com/celestiaorg/celestia-node/pull/375) [@Bidon15](https://github.com/Bidon15)
 - [feat(cmd): give a birth to cel-shed and p2p key utilities #281](https://github.com/celestiaorg/celestia-node/pull/281) [@Wondertan](https://github.com/Wondertan)
 - [feat(cmd|node): MutualPeers Node option and CLI flag #280](https://github.com/celestiaorg/celestia-node/pull/280) [@Wondertan](https://github.com/Wondertan)
 - [node: enhance DI allowing overriding of dependencies](https://github.com/celestiaorg/celestia-node/pull/290) [@Wondertan](https://github.com/Wondertan)
--- a/CHANGELOG-PENDING.md
+++ b/CHANGELOG-PENDING.md
@@ -12,6 +12,7 @@ Month, DD, YYYY
 
 ### FEATURES
 
+- [node/config: add WithBootstrapPeers option #375](https://github.com/celestiaorg/celestia-node/pull/375) [@Bidon15](https://github.com/Bidon15)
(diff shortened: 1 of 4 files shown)


CHANGELOG-PENDING.md
```diff
@@ -12,6 +12,7 @@ Month, DD, YYYY
 
 ### FEATURES
 
+- [node/config: add WithBootstrapPeers option #375](https://github.com/celestiaorg/celestia-node/pull/375) [@Bidon15](https://github.com/Bidon15)
```
renaynay, line 15: We don't need changelog for this feature, IMO

node/config_opts.go
```diff
@@ -34,9 +34,18 @@ func WithConfig(custom *Config) Option {
 	}
 }
 
+// WithMutualPeers sets MutualPeers[] config.
```
renaynay, line 37:
```suggestion
// WithMutualPeers sets the `MutualPeers` field in the config.
```

node/config_opts.go
```diff
@@ -34,9 +34,18 @@ func WithConfig(custom *Config) Option {
...
 func WithMutualPeers(addrs []string) Option {
 	return func(cfg *Config, _ *settings) (_ error) {
 		cfg.P2P.MutualPeers = addrs
 		return nil
 	}
 }
+
+// WithBootstrapPeers sets BootstrapPeers[] config.
```
renaynay, line 45:
```suggestion
// WithBootstrapPeers sets the `BootstrapPeers` field in the config.
```

CHANGELOG-PENDING.md
Bidon15: It's either we don't include it or move it to the improvements section 😓

renaynay (dismissed):

CHANGELOG-PENDING.md
renaynay: I think we don't need to include every config override feature

Bidon15: Ok. Removing the changelog then

renaynay (approved):

node/light_test.go
```diff
@@ -62,3 +62,16 @@ func TestNewLightWithHost(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, node.Host.ID(), nw.Peers()[0])
 }
+
+func TestLight_WithBootstrapPeers(t *testing.T) {
```
Wondertan, line 66: Why Light is allowed to have tests only for Bootstrapp and not mutual? :sad_boy_cry_png:

Bidon15: >_< Ok 🙈

renaynay (approved):

Wondertan (approved):

Issue #377, opened 2022-01-24
header: race: core_listener.go


Wondertan:
![image](https://user-images.githubusercontent.com/22449364/150794843-8bd497e0-26f2-4350-8237-53b903522b71.png)

liamsi:
This is no bueno because:
 - a screenshot is not searchable
 - the screenshot does not show which test triggers this

PR #378, opened 2022-01-24
header: fix race in core listener

Wondertan:
Closes: #377

### DIFF
--- a/das/daser.go
+++ b/das/daser.go
@@ -14,7 +14,6 @@ import (
 var log = logging.Logger("das")
 
 // DASer continuously validates availability of data committed to headers.
-// TODO(@Wondertan): Start and Stop is better be thread-safe.
 type DASer struct {
 	da   share.Availability
 	hsub header.Subscriber
@@ -51,10 +50,6 @@ func (d *DASer) Start(context.Context) error {
 
 // Stop stops sampling.
 func (d *DASer) Stop(ctx context.Context) error {
-	if d.cancel == nil {
-		return fmt.Errorf("da: DASer already stopped")
-	}
-
 	d.cancel()
 	select {
 	case <-d.done:
--- a/service/header/core_listener.go
+++ b/service/header/core_listener.go
@@ -24,9 +24,6 @@ type CoreListener struct {
 	fetcher *core.BlockFetcher
 	dag     format.DAGService
 
-	blockSub <-chan *types.Block
-
-	ctx    context.Context
 	cancel context.CancelFunc
 }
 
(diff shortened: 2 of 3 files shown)


das/daser.go
```diff
@@ -14,7 +14,6 @@ import (
 var log = logging.Logger("das")
 
 // DASer continuously validates availability of data committed to headers.
-// TODO(@Wondertan): Start and Stop is better be thread-safe.
```
Wondertan, line 17: Actually no, we should never call Start and Stop from different routines and we are doing so, thus this todo is not relevant

das/daser.go
```diff
@@ -51,10 +50,6 @@ func (d *DASer) Start(context.Context) error {
 
 // Stop stops sampling.
 func (d *DASer) Stop(ctx context.Context) error {
-	if d.cancel == nil {
-		return fmt.Errorf("da: DASer already stopped")
-	}
-
```
Wondertan, line 57: Instead, If it is already stopped, there will be a panic. IMO, panic in this case is better as it tells that we are using code incorrectly. Meaning we are Starting once, but stopping twice. This is only relevant for Stop. Maybe this will change once we have restartable stuff during runtime.

service/header/core_listener.go
```diff
@@ -40,7 +37,7 @@ func NewCoreListener(p2pSub *P2PSubscriber, fetcher *core.BlockFetcher, dag form
 
 // Start kicks off the CoreListener listener loop.
 func (cl *CoreListener) Start(ctx context.Context) error {
-	if cl.blockSub != nil {
+	if cl.cancel != nil {
```
Wondertan, line 40: We should always use the cancel func as a way to check the lifecycle state. It won't ever cause us races.

service/header/core_listener.go
```diff
@@ -49,46 +46,45 @@ func (cl *CoreListener) Start(ctx context.Context) error {
...
-	cl.blockSub = nil
+	cl.cancel = nil
 	return cl.fetcher.UnsubscribeNewBlockEvent(ctx)
 }
 
 // listen kicks off a loop, listening for new block events from Core,
 // generating ExtendedHeaders and broadcasting them to the header-sub
 // gossipsub network.
-func (cl *CoreListener) listen() {
+func (cl *CoreListener) listen(ctx context.Context, sub <-chan *types.Block) {
```
Wondertan, line 65: It is cleaner to pass those instead of accessing from cl.

renaynay (dismissed):
LGTM. Thanks.

liamsi:
Unclear which race this fixes. There is still a race in the CI. But that one is a different one?
See: https://github.com/celestiaorg/celestia-node/issues/377#issuecomment-1020416181

Wondertan:
@liamsi, the race on the CI is goroutine limit issue and not a real race. This PR fixes what it states to fix - #377

liamsi (approved):
🟢 ✅ 🟩 🍏 📗 💚

renaynay (approved):

PR #380, opened 2022-01-24
core: suppress logging of retry-client

### LINKED ISSUE #379: core: disable logs of retriable http client

renaynay:
Resolves #379.

### DIFF
--- a/core/client.go
+++ b/core/client.go
@@ -18,6 +18,8 @@ type Client = client.Client
 func NewRemote(protocol, remoteAddr string) (Client, error) {
 	httpClient := retryhttp.NewClient()
 	httpClient.RetryMax = 2
+	// suppress logging
+	httpClient.Logger = nil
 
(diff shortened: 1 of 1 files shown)


Wondertan (approved):
Nice! 
* We can make it using our logger
* Although logs from this retry client are not that important.

core/client.go
```diff
@@ -18,6 +18,8 @@ type Client = client.Client
 func NewRemote(protocol, remoteAddr string) (Client, error) {
 	httpClient := retryhttp.NewClient()
 	httpClient.RetryMax = 2
+	// suppress logging
+	httpClient.Logger = nil
```
liamsi, line 22: I'd also prefer this to use the main logger used in this package instead of nil but this is still better than before.

liamsi (dismissed):

Wondertan (approved):

liamsi (approved):

PR #388, opened 2022-01-24
tests(ipld): temporary skip a test not to see the race detector complaining

Wondertan:


ipld/read_test.go
```diff
@@ -109,6 +109,8 @@ func TestBlockRecovery(t *testing.T) {
 }
 
 func TestRetrieveBlockData(t *testing.T) {
+	t.Skip("skipping as it spawns too many goroutines")
```
liamsi, line 112: Alright. It is not easily possible to only skip the large one, right?

Wondertan: Possible

liamsi: Let's do that then :)

Wondertan: Done

liamsi: Thanks for putting this in a separate PR btw 🙏🏼

ipld/read_test.go
```diff
@@ -123,7 +123,9 @@ func TestRetrieveBlockData(t *testing.T) {
 		{"128x128(max)", MaxSquareSize},
 	}
 	for _, tc := range tests {
-
+		if tc.squareSize == MaxSquareSize {
+			t.Skip("skipping as it spawns too many goroutines")
+		}
```
liamsi, line 128: dopey-dopey

renaynay (approved):

liamsi (approved):
Thanks!

Issue #389, opened 2022-01-24
Bridge node doesn't include header flags on `init` and `start`


jbowen93:
During the refactor here: https://github.com/celestiaorg/celestia-node/commit/b87d4bf36b9d3cbbb18c6efb8686da134693b487#diff-304d1a2647db4cdfdb35931612185b94b8738294c2acbdd8db6c0ab6600c7c1f the Bridge node didn't have the `cmdnode.HeadersFlags()` included. 

This prevents passing `headers.trusted-hash or headers.trusted-peer` through a command line flag which is how the ephemeral cluster is configured.

PR #390, opened 2022-01-24
[Docker | Bridge init/start]: entrypoint.sh uses node.store flag and bridge node includes HeadersFlags

jbowen93:
Update docker/entrypoint.sh to use node.store flag to work with newest main. 

Also add the `cmdnode.HeadersFlags()` to the Bridge Node's `init` and `start` commands.

Resolves https://github.com/celestiaorg/celestia-node/issues/389


CHANGELOG-PENDING.md
```diff
@@ -31,6 +31,7 @@ Month, DD, YYYY
 - [docker: Created `docker/` dir with `Dockerfile` and `entrypoint.sh` script](https://github.com/celestiaorg/celestia-node/pull/295) [@jbowen93](https://github.com/jbowen93)
 - [chore(share): handle rows concurrently in GetSharesByNamespace #241](https://github.com/celestiaorg/celestia-node/pull/241) [@vgonkivs](https://github.com/vgonkivs)
 - [ci: adding data race detector action](https://github.com/celestiaorg/celestia-node/pull/289) [@Bidon15](https://github.com/Bidon15)
+- [node: add the cmdnode.HeadersFlags() to the Bridge Node's init and start commands](https://github.com/celestiaorg/celestia-node/pull/390)
```
renaynay, line 34:
```suggestion
- [node: add the cmdnode.HeadersFlags() to the Bridge Node's init and start commands #390](https://github.com/celestiaorg/celestia-node/pull/390) [@jbowen93](https://github.com/jbowen93)
```

CHANGELOG-PENDING.md
```diff
@@ -41,6 +42,7 @@ Month, DD, YYYY
 - [ci: increase tokens ratio for dupl to fix false positive scenarios](https://github.com/celestiaorg/celestia-node/pull/314) [@Bidon15](https://github.com/Bidon15)
 - [node: Wrap datastore with mutex to prevent data race](https://github.com/celestiaorg/celestia-node/pull/325) [@Bidon15](https://github.com/Bidon15)
 - [ci: update Docker entrypoint.sh to use new `store.path` flag name](https://github.com/celestiaorg/celestia-node/pull/337) [@jbowen93](https://github.com/jbowen93)
+- [ci: update docker/entrypoint.sh to use new `node.store` flag replacing `store.path`](https://github.com/celestiaorg/celestia-node/pull/390)
```
renaynay, line 45:
```suggestion
- [docker: update docker/entrypoint.sh to use new `node.store` flag replacing `store.path` #390](https://github.com/celestiaorg/celestia-node/pull/390) [@jbowen93 ](https://github.com/jbowen93)
```

Wondertan (changes requested):
`HeaderFlags` also bring in the `trusted-peer` flag which the Bridge node does not need and won't ever use. Can you pls decouple `HeadersFlags` into two flags(while still preserving HeadersFlags pointing to the new decoupled ones(name as you wish)), so Bridge can only import `trusted-hash`?

(Note for the question that you will encounter during implementation: yes, it is ok to return whole FlagSet only for one flag) 

Other than that, LGTM

jbowen93:
> `HeaderFlags` also bring in the `trusted-peer` flag which the Bridge node does not need and won't ever use. Can you pls decouple `HeadersFlags` into two flags(while still preserving HeadersFlags pointing to the new decoupled ones(name as you wish)), so Bridge can only import `trusted-hash`?
> 
> (Note for the question that you will encounter during implementation: yes, it is ok to return whole FlagSet only for one flag)
> 
> Other than that, LGTM

ack. Will update tonight or monday.

cmd/celestia/cmd_test.go
```diff
@@ -0,0 +1,96 @@
...
+		rootCmd.SetArgs([]string{
+			"bridge", 
+			"--node.store", ".celestia-light",
+			"init",
+		})
+		err := rootCmd.ExecuteContext(cmdnode.WithEnv(context.Background()))
+		require.NoError(t, err)
+	})
+
+	// Commented out until a dry-run option can be implemented
```
renaynay, line 36: Pls tag your github handle here

```
// TODO @jbowen93: commented out until.... 

```

cmd/celestia/cmd_test.go
```diff
@@ -0,0 +1,96 @@
...
+		rootCmd.SetArgs([]string{
+			"bridge", 
+			"--node.store", ".celestia-bridge",
+			"init",
+		})
+		err := rootCmd.ExecuteContext(cmdnode.WithEnv(context.Background()))
+		require.NoError(t, err)
+	})
+
+	// Commented out until a dry-run option can be implemented
```
renaynay, line 78: same here.

cmd/flags_header.go
```diff
@@ -59,3 +50,61 @@ func ParseHeadersFlags(cmd *cobra.Command, env *Env) error {
 
 	return nil
 }
+
+// TrustedPeerFlags gives a set of hardcoded Header package TrustedPeer flags.
```
renaynay, line 54:
```suggestion
// TrustedPeerFlags returns a set of flags related to configuring a `TrustedPeer`.
```

cmd/flags_header.go
```diff
@@ -59,3 +50,61 @@ func ParseHeadersFlags(cmd *cobra.Command, env *Env) error {
...
+	flags := &flag.FlagSet{}
+
+	flags.String(
+		headersTrustedPeerFlag,
+		"",
+		"Multiaddr of a reliable peer to fetch headers from. (Format: multiformats.io/multiaddr)",
+	)
+
+	return flags
+}
```
renaynay, line 65: I don't understand why this needs to be a function?

Maybe I'm missing something? cc @Wondertan

renaynay (commented):
Can you also run lint please 🙏🏻

cmd/flags_header.go
jbowen93: I was just following the convention that already existed. I briefly considered trying to get [Viper](https://github.com/spf13/viper) following the conventions [here](https://github.com/carolynvs/stingoftheviper) but it took me a bit to wrap my head around the whole config process.

renaynay: Yeah I don't see the purpose in making this a function rather than just a var.

Wondertan: We can't make it a var because then multiple commands will use the same instance of one global flag. Remember that we have multiple node types(bridge and light) and for each of them own flag should be created.

Wondertan (commented):
LGTM and thanks for catching the pattern!

One semantical nit for discussion: should we not append 's' to func returning or parsing one flag only? E.g `TrustedPeerFlags` returns only one flag, so, IMO, it makes sense to remove 's' from the end.

Wondertan:
Regarding the tests. I am really happy about new tests being added to test the flags and CLI in some way. Unfortunately, we didn't have time to write proper tests for CLI commands and are testing such things manually only 😬. For this we would need to develop some CLI testing suite and a simple pattern to test stuff, so your tests here will be reworked at some point.

cmd/flags_header.go
```diff
@@ -59,3 +50,59 @@ func ParseHeadersFlags(cmd *cobra.Command, env *Env) error {
...
+
+// ParseTrustedPeerFlags parses Header package flags from the given cmd and applies values to Env.
+func ParseTrustedPeerFlags(cmd *cobra.Command, env *Env) error {
```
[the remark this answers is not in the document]
Wondertan: Happy to approve after this one is applied

cmd/flags_header.go
renaynay: ack.

cmd/flags_header.go
```diff
@@ -28,25 +28,8 @@ func HeadersFlags() *flag.FlagSet {
...
-	tpeer := cmd.Flag(headersTrustedPeerFlag).Value.String()
-	if tpeer != "" {
-		_, err := multiaddr.NewMultiaddr(tpeer)
-		if err != nil {
-			return fmt.Errorf("cmd: while parsing '%s': %w", headersTrustedPeerFlag, err)
-		}
-
-		env.AddOptions(node.WithTrustedPeer(tpeer))
-	}
+	ParseTrustedHashFlags(cmd, env)
```
renaynay, line 31: maybe check the err here?

cmd/flags_header.go
```diff
@@ -28,25 +28,8 @@ func HeadersFlags() *flag.FlagSet {
...
+	ParseTrustedPeerFlags(cmd, env)
```
renaynay, line 32: and here

cmd/flags_header.go
jbowen93: done

cmd/flags_header.go
jbowen93: done

renaynay (dismissed):

Wondertan (approved):

Wondertan:
cc @renaynay for the test failure(race), it is something new and is related to Core block fetching

renaynay:
made an issue re race in `TestBlockFetcher_GetBlock_and_SubscribeNewBlockEvent`. 

@jbowen93 this unfortunately needs rebasing 😭

renaynay:
I think before @Wondertan

Wondertan (approved):

renaynay (approved):

Issue #392, opened 2022-01-25
swamp: race in NewSwamp and Core


Wondertan:
![image](https://user-images.githubusercontent.com/22449364/150992689-74f52b01-b556-42eb-ac76-d8891412791f.png)

Wondertan:
Found in https://github.com/celestiaorg/celestia-node/pull/334

Bidon15:
So this should be fixed after I finish with https://github.com/celestiaorg/celestia-node/issues/371

Some sample runs of the draft(will be closed) https://github.com/celestiaorg/celestia-node/pull/393

https://github.com/celestiaorg/celestia-node/runs/4938174725?check_suite_focus=true

PR #394, opened 2022-01-25
swamp: separate tendermint node creation from the swamp

### LINKED ISSUE #371: swamp/core: Test-case should set block retention
During test-case setup phase, we have to provide the functionality to the test-case to set the block retention as required by the scenario, not hard coding in the swamp initialisation

Bidon15:
Adapting the sync test to this change

Resolves #371 #392

### DIFF
--- a/node/tests/swamp/config.go
+++ b/node/tests/swamp/config.go
@@ -16,11 +16,11 @@ type Config struct {
 }
 
 // DefaultConfig creates a KvStore with a block retention of 200
-// In addition, the empty block interval is set to 100ms
+// In addition, the empty block interval is set to 200ms
 func DefaultConfig() *Config {
 	app := core.CreateKvStore(200)
 	tnCfg := tn.TestConfig()
-	tnCfg.Consensus.CreateEmptyBlocksInterval = 100 * time.Millisecond
+	tnCfg.Consensus.CreateEmptyBlocksInterval = 200 * time.Millisecond
 	return &Config{
 		App:     app,
 		CoreCfg: tnCfg,
--- a/node/tests/swamp/config.go
+++ b/node/tests/swamp/config.go
@@ -3,9 +3,10 @@ package swamp
 import (
 	"time"
 
-	"github.com/celestiaorg/celestia-node/core"
 	"github.com/tendermint/tendermint/abci/types"
 	tn "github.com/tendermint/tendermint/config"
+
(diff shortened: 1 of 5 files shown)


Bidon15:
I won't open this to be r-4-r as I want #334 to be in main and then re-run this
Currently, I am seeing errors in test execution

node/tests/swamp/config.go
```diff
@@ -0,0 +1,28 @@
...
+
+	"github.com/tendermint/tendermint/abci/types"
+	tn "github.com/tendermint/tendermint/config"
+
+	"github.com/celestiaorg/celestia-node/core"
+)
+
+// Config struct represents a set of pre-requisite attributes from the test scenario
+type Config struct {
+	App     types.Application
```
Wondertan, line 14: Semantically, it is strange to have a whole App under the config, but it might be ok.

node/tests/swamp/swamp.go
```diff
@@ -63,9 +61,30 @@ func NewSwamp(t *testing.T) *Swamp {
...
+	rpctest.RecreateConfig(&opt)
+
+	tn := rpctest.NewTendermint(cfg.App, &opt)
+
+	// rewriting the created config with test's one
+	tn.Config().Consensus = cfg.CoreCfg.Consensus
+
+	err := tn.Start()
+
+	return tn, err
```
Wondertan, line 85: nit: you can directly `return tn, tn.Start()`

Bidon15: Agreed. Done

node/tests/swamp/config.go
Bidon15: tbh, I haven't came up with a more neat idea, yet. I propose we leave this as is and come back as we will start playing with core/app side more that just consensus and kvStores.

Wondertan (approved):

node/tests/swamp/config.go
renaynay, line 14: i find it odd that the `app` is in the `Config` struct but ok. Maybe swamp should have a `InfraComponents` struct or something like that. I think `Config` is just a misnomer.

renaynay (approved):

node/tests/swamp/config.go
Bidon15: Oh, thank you so much for the suggestion 🤩  . I'll change this name to InfraComponents in the upcoming PR

PR #398, opened 2022-01-26
logs(node): add logs to IsInit func to report errors

Wondertan:
I was playing around with some stuff and accidentally broke config formatting and the node didn't tell me any meaningful message and this PR should fix this,


Wondertan:
Logging for such startup errors looks ugly(timestamps and other logs verbosity) and instead we should probably just fmt.Println, but this is other concern.

liamsi (approved):
🟢

renaynay (approved):

PR #399, opened 2022-01-26
feat(params): add bootstrappers

Wondertan:
Just added a list of boostrappers

### DIFF
--- a/CHANGELOG-PENDING.md
+++ b/CHANGELOG-PENDING.md
@@ -21,6 +21,7 @@ Month, DD, YYYY
 
 ### IMPROVEMENTS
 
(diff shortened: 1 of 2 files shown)


Wondertan:
I tested manually by connecting to those addresses.

liamsi (approved):
I'd prefer if these was in a config file instead of hardcoded but that is OK.

renaynay:
@liamsi geth also hardcodes them, this is normal.

liamsi:
> geth also hardcodes them, this is normal.

per network / testnet?

Wondertan:
Moreover, everyone hardcodes them besides tendermint networks and they should do this as well.

PR #55, opened 2021-09-10
core: add tests for remote node

renaynay:
This PR adds test for remote core node. 

#51 depends on this PR.

PR #407, opened 2022-01-31
feat: make cli more portable and implement ability to add custom services

evan-forbes:
The goal of this PR is to explore easy ways to for us to add custom functionality to a celestia-node. It spawned from the `dalc`'s need to access the `header.Store`. Rather than exporting a `header.Store` field in `node.Node`, this PR makes it possible for the `dalc` to act as a plugin, creating a `celestia` binary that incorporates the additional services and initialization processes.

it features a new `node.Plugin` interface, which is passed to functions that create the `rootCmd` cli command, and is used to add additional services to celestia-node.

the dalc uses this new feature to run optimint specific services in [#55](https://github.com/celestiaorg/dalc/pull/55)


node/node.go
```diff
@@ -59,12 +59,16 @@ type Node struct {
 
 	DASer *das.DASer `optional:"true"`
 
+	// PluginOutlet serves as an arbitrary type for plugins to return so that fx
+	// will call them
+	PluginOutlet
```
evan-forbes, line 64: Plugins have to have a service that initializes this `PluginOutlet` types so that fx will call them at all.

node/node.go
```diff
@@ -80,11 +84,16 @@ func New(tp Type, store Store, options ...Option) (*Node, error) {
...
+		pluginComponents[i] = plug.Components(cfg, store)
+	}
+
 	switch tp {
 	case Bridge:
-		return newNode(bridgeComponents(cfg, store), s.overrides())
+		return newNode(bridgeComponents(cfg, store), fxutil.Options(pluginComponents...), s.overrides())
 	case Light:
-		return newNode(lightComponents(cfg, store), s.overrides())
+		return newNode(lightComponents(cfg, store), fxutil.Options(pluginComponents...), s.overrides())
```
evan-forbes, line 96: the components/services that a plugin returns are passed to the node

evan-forbes (commented):
marking this ready for review

evan-forbes:
I'll leave this open for further review, but I think its best to wait until the entire team can look at these changes, as they weren't discussed anywhere previously.

node/init.go
```diff
@@ -12,7 +12,7 @@ import (
 
 // Init initializes the Node FileSystem Store for the given Node Type 'tp' in the directory under 'path' with
 // default Config. Options are applied over default Config and persisted on disk.
-func Init(path string, tp Type, options ...Option) error {
+func Init(path string, tp Type, plugs []Plugin, options ...Option) error {
```
liamsi, line 15: Could the plugs also be Options somehow?

evan-forbes:
> I'm inclined to insist for an ADR that a) motivates this change and b) explains the decision, or, at least some light weight docs that empower devs to use this feature. WDYT?

Yep, that makes perfect sense. Ideally that should have been the process all along, but I didn't really realize that this was needed until actually working on it. I should be able write the ADR pretty quick since the code is already mostly completed.

evan-forbes:
converted to a draft until a decision for #414 is made

liamsi:
IMO: we should merge this unless the node team thinks this introduces too much technical debt. I don't see how this would be the case though. With the removal of fxutils a some of the existing code base has to be overhauled anyways. I don't think this PR would change this massively.

Wondertan:
Wasn't able to review it on the weekends, as I was focusing on release. Once is done, I will start working on Public API for celestia-node and this work will be highly relevant, so I'll review it for sure.

renaynay:
@evan-forbes you were a visionary ❤️

Bidon15:
Closing this in favour of new one. 
https://github.com/celestiaorg/celestia-node/issues/1369

Thank you for the visionary PR :giga-chad: ❤️

PR #237, opened 2021-11-23
node/p2p: Deduplicate PubSub msgs

Wondertan:
Closes #236

Issue #306, opened 2021-12-21
Blocker for updating libp2p/ipfs libs


Wondertan:
Why kinda? Because we can't update most of libp2p and ipfs deps, until we don't upgrade NMT-IPLD to support conversion to ipld/go-ipld-prime.

Why? Because there was an interface change in go-datastore which affected all the PL repos relying on go-datastore, roughly 80%.

Why is that a problem? The IPFS migration to go-ipld-prime happened before this massive go-datastore(DS) migration, so if you want to use the latest go-libp2p, it forces you to rely on the patched go-datastore, which forces you to rely on the patched go-merkledag, which forces you to use go-ipld-prime.

Without updating to the latest libp2p stuff we loosing sometimes important bug fixes and optimizations(mostly in bitswap related) and peer store memory leaks. Fortunately, there is one more reason to migrate to go-ipld-prime (or ipldv1).

renaynay:
![Uploading Screenshot 2022-02-03 at 09.54.49.png…]()

renaynay:
<img width="1012" alt="Screenshot 2022-02-03 at 09 54 49" src="https://user-images.githubusercontent.com/41963722/152310851-cf0a0fc4-dd04-4143-9f5a-5512bba0590a.png">

liamsi:
Can we at least include which versions we are talking about?

Wondertan:
@liamsi, I removed the old confusing description with a new shorter, while, I believe, a more descriptive one and added a version number for the conflicting dependency.

PR #409, opened 2022-02-02
service/header: fix ExtendedHeader message duplicates on the network

Wondertan:
## What
Finally, a fix to solve this longstanding issue. 

## How
By using a new feature in libp2p/go-libp2p-pubsub#468 we can now implement custom per-topic message id generation where for the 'HeaderSub' we avoid Signatures from being hashed.

## TODO
- [ ] An integration test to assert this is exactly true
- [ ] Manually deploy a segregated network of Bridge nodes and assert that light nodes do not receive duplicates occasionally. 

Closes #237

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -45,4 +45,7 @@ require (
 	go.uber.org/zap v1.20.0
 )
 
-replace github.com/tendermint/tendermint v0.34.14 => github.com/celestiaorg/celestia-core v0.34.14-celestia
+replace (
+	github.com/libp2p/go-libp2p-pubsub v0.5.7-0.20211029175501-5c90105738cf => github.com/celestiaorg/go-libp2p-pubsub v0.5.7-0.20220202152246-c33ecdf03b34
+	github.com/tendermint/tendermint v0.34.14 => github.com/celestiaorg/celestia-core v0.34.14-celestia
+)
--- a/service/header/p2p_subscriber.go
+++ b/service/header/p2p_subscriber.go
@@ -6,6 +6,8 @@ import (
 
 	"github.com/libp2p/go-libp2p-core/peer"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
+	pb "github.com/libp2p/go-libp2p-pubsub/pb"
+	"github.com/minio/blake2b-simd"
 )
 
 // PubSubTopic hardcodes the name of the ExtendedHeader
@@ -30,7 +32,7 @@ func NewP2PSubscriber(ps *pubsub.PubSub) *P2PSubscriber {
 // Start starts the P2PSubscriber, registering a topic validator for the "header-sub"
 // topic and joining it.
 func (p *P2PSubscriber) Start(context.Context) (err error) {
-	p.topic, err = p.pubsub.Join(PubSubTopic)
+	p.topic, err = p.pubsub.Join(PubSubTopic, pubsub.WithTopicMessageIdFn(msgID))
 	return err
 }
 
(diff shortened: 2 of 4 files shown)


go.mod
```diff
@@ -45,4 +45,7 @@ require (
 	go.uber.org/zap v1.20.0
 )
 
-replace github.com/tendermint/tendermint v0.34.14 => github.com/celestiaorg/celestia-core v0.34.14-celestia
+replace (
+	github.com/libp2p/go-libp2p-pubsub v0.5.7-0.20211029175501-5c90105738cf => github.com/celestiaorg/go-libp2p-pubsub v0.5.7-0.20220202152246-c33ecdf03b34
```
Wondertan, line 49: A hack to do less work, but still have the newest pubsub features, excluding the commit that breaks dependencies for us(#306)

adlerjohn:
#237 seems like a PR, did you link to the right thing?

service/header/p2p_subscriber.go
```diff
@@ -78,3 +80,36 @@ func (p *P2PSubscriber) Broadcast(ctx context.Context, header *ExtendedHeader) e
...
+		log.Errorw("unmarshalling header while computing msg id", "err", err)
+		return msgID(pmsg.Data)
+	}
+
+	// IMPORTANT NOTE:
+	// Due to the nature of the Tendermint consensus, validators don't collect all commit signatures,
+	// but only the minimum required amount of them(+2/3 of power). Also, signatures are collected asynchronously.
+	// So for, each validator may have a different set of minimally required signatures causing nondeterminism in
+	// the header message gossiped over the network. Subsequently, this causes message duplicates as each Bridge Node,
+	// connected to personal a validator, sends the validaotor's own view of commits of effectively the same header.
```
adlerjohn, line 103:
```suggestion
	// Due to the nature of the Tendermint consensus, validators don't necessarily collect commit signatures from the entire validator set,
	// but only the minimum required amount of them (>2/3 of voting power). In addition, signatures are collected asynchronously.
	// Therefore, each validator may have a different set of signatures that pass the minimum required voting power threshold, causing nondeterminism in
	// the header message gossiped over the network. Subsequently, this causes message duplicates as each Bridge Node,
	// connected to a personal validator, sends the validaotor's own view of commits of effectively the same header.
```

adlerjohn (commented):
Nice fix! Left a comment nit.

service/header/p2p_subscriber.go
```diff
@@ -78,3 +80,36 @@ func (p *P2PSubscriber) Broadcast(ctx context.Context, header *ExtendedHeader) e
...
+	//
+	// To solve the problem above, we exclude nondeterministic value from message id calculation
+	h.Commit.Signatures = nil
+
+	data, err := MarshalExtendedHeader(h)
+	if err != nil {
+		log.Errorw("unmarshalling header while computing msg id", "err", err)
```
renaynay, line 110:
```suggestion
		log.Errorw("marshalling header while computing msg id", "err", err)
```

service/header/p2p_subscriber.go
```diff
@@ -78,3 +80,36 @@ func (p *P2PSubscriber) Broadcast(ctx context.Context, header *ExtendedHeader) e
...
+func msgID(pmsg *pb.Message) string {
+	msgID := func(data []byte) string {
+		hash := blake2b.Sum256(data)
+		return string(hash[:])
+	}
+
+	h, err := UnmarshalExtendedHeader(pmsg.Data)
+	if err != nil {
+		log.Errorw("unmarshalling header while computing msg id", "err", err)
+		return msgID(pmsg.Data)
```
renaynay, line 95: So i guess technically if we run into some marshalling errs in this msgID function, then we *could* have duplicated header messages in the network, but it's unlikely right?

Wondertan:
Thank you for catching this. No, I linked the wrong thing.
> #237 seems like a PR, did you link to the right thing?

Wondertan:
Just tested manually on my node that this works as expected and there are no duplicate header logs anymore!
YAY!!! 🎉

service/header/p2p_subscriber.go
renaynay: ack

renaynay (approved):
LGTM

adlerjohn (approved):

Wondertan:
A quick note that we are now maintaining our own fork of pubsub due to https://github.com/celestiaorg/celestia-node/pull/409#discussion_r797772373

Wondertan:
A quick note that we are now maintaining our own fork of pubsub due to https://github.com/celestiaorg/celestia-node/pull/409#discussion_r797772373

liamsi (commented):
Looks good! No objections from my side. I'd give the inner `msgID` function a different name though.

service/header/p2p_subscriber.go
```diff
@@ -78,3 +80,38 @@ func (p *P2PSubscriber) Broadcast(ctx context.Context, header *ExtendedHeader) e
 	}
 	return p.topic.Publish(ctx, bin)
 }
+
+// msgID computes an id for a pubsub message
+// TODO(@Wondertan): This cause additional allocations per each recvd message in the topic. Find a way to avoid those.
+func msgID(pmsg *pb.Message) string {
+	msgID := func(data []byte) string {
```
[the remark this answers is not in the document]
Wondertan: Ok

go.mod
Wondertan: Yop

Wondertan: Done https://github.com/celestiaorg/celestia-node/issues/436

Wondertan: Note that fixing this would indirectly require an update to IPLDv1, which we should do at some point anyway.

liamsi (approved):

renaynay (approved):
LGTM
+1 to  @liamsi comment re: issue for dep

PR #411, opened 2022-02-04
[DRAFT] `CoreAccess`

renaynay:
DRAFT.


Wondertan:
Why do we need a TrustedCore if we have RemoteCore?

PR #414, opened 2022-02-07
ADR 005 Plugins

evan-forbes:
This ADR has a pending implementation in #407 and an example plugin at [#55](https://github.com/celestiaorg/dalc/pull/55)

[rendered](https://github.com/celestiaorg/celestia-node/blob/abb50637185200b18333e5b42e501c6b36da103e/docs/adr/adr-005-plugins.md)


evan-forbes:
> @evan-forbes pls add changelog :)

wait, should ADRs add to the changelogs? I couldn't find other adrs in the logs, so I added the `no changelog` label, but happy to add a log if that's needed

renaynay:
Oh did we not add any ADRs? Maybe we can make a section specifically for ADRs bc I think they're important to note.

docs/adr/adr-005-plugins.md
```diff
@@ -0,0 +1,275 @@
+# ADR 005: Plugins
+
+## Changelog
+
+- 06-02-2022: init commit
```
adlerjohn, line 5: 🇺🇸
```suggestion
- 2022-02-06: init commit
```

docs/adr/adr-005-plugins.md
```diff
@@ -0,0 +1,275 @@
...
+- it will likely add future engineering efforts if we continue to support this feature
+- perhaps less intuitive for those not familiar with uber/fx
+
+## References
+
+blocking a refactor of dalc [#55](https://github.com/celestiaorg/dalc/pull/55)
+
+first discussed and current implementation [#407](https://github.com/celestiaorg/celestia-node/pull/407)
+
+initial issue [#406](https://github.com/celestiaorg/celestia-node/issues/406)
```
adlerjohn, line 275:
```suggestion
- blocking a refactor of dalc [#55](https://github.com/celestiaorg/dalc/pull/55)
- first discussed and current implementation [#407](https://github.com/celestiaorg/celestia-node/pull/407)
- initial issue [#406](https://github.com/celestiaorg/celestia-node/issues/406)
```

adlerjohn (commented):
Needs changelog

adlerjohn (approved):
utACK. Seems reasonable.

docs/adr/adr-005-plugins.md
```diff
@@ -0,0 +1,273 @@
...
+
+It should be possible to achieve the flexibility that a custom node requires, but it limits the UX for both developers and custom node users. For instance, running the custom node takes two binaries, both of which have to be installed using the associated versions. When creating a custom node, we have to implement redundant functionality, such as basic program structure, clis, and some tests.
+
+It also makes it more difficult for custom nodes to support different types of celestia nodes. With the plugin design, it is possible to run the same plugin on every node type without adding any code.
+
+Lastly, this approach wouldn't make it easy to run custom nodes with multiple modifications, whereas the plugin approach does.
+
+### Expose more services over rpc
+
+This would likely require a lot of overhead, and not eliminate any of the existing complexity.
```
liamsi, line 23: This is not an argument against the plugin approach but: rpc might be required either way though.

liamsi (approved):
Good!

evan-forbes:
There's one option that was not discussed in this ADR and I think would be simpler than this. Instead of supporting plugins, we would merge a slightly more generalized dalc grpc server (replace `optimint.Block`s with `core.Message`s) into celestia-node. Then optimint can copy and paste what ever protobuf definitions it needs instead of importing celestia-core. Then we should be able to archive the dalc repo.

docs/adr/adr-005-plugins.md
```diff
@@ -0,0 +1,273 @@
+# ADR 005: Plugins
+
+## Changelog
+
+- 2022-02-06: init commit
+
+## Context
+
+The modular design of celestia-node allows for the functionality it provides to be extended. The main blocker for creating custom light client applications is having a mechanism to access the internals of a celestia-node. Here we are suggesting that there be some interface that installs custom services into celestia-node, along with making the celestia-node cli more portable, so that light client developers can create their own binaries with the custom services that they create.
```
renaynay, line 9: I guess my only question for this design is what the difference is between us adding in this plugin functionality vs us creating a robust enough `LightNode` or `FullNode` RPC interface that would enable the same kind of custom client? I don't really see the difference between Plugins and providing a super robust `LightNode` RPC interface.

evan-forbes: as mentioned [here](https://github.com/celestiaorg/celestia-node/pull/414#issuecomment-1055871523), I do think that in the short-medium term, we should simply use a simplified rpc and no plugins. 

However, this is only possible because recently we have reduced the scope of how agnostic optimint needs to be, and are willing to copy paste code from celestia-core instead of importing. If we still wanted those features, with a robust RPC we would have to deal with all the UX terribleness of having one extra binary to run, handling yet another cli, redundant testing, and redundant basic program structure. With plugins we don't.

Plugins would also enable entirely new node types, like custom DAS, storage, rpc, and even custom blockchains that have a one way bridge from celestia. These custom node types might be possible over rpc, but they would suffer the same ux issues as the dalc does, and might even be reliant on us adding specific rpc methods that we otherwise would not wish to expose.

again, I think designing for those things so early is a mistake, and we should go with a simplified dalc-esque rpc instead, but I think having in process access to celestia-node internals would prove invaluable to those custom node types.

renaynay (approved):

adlerjohn (approved):

PR #420, opened 2022-02-09
service/state: Provide simple scaffold for `StateService` and implement `CoreAccess`

renaynay:
This PR is the first implementation of `StateAccessor` over core: `CoreAccess`. It will provide a `light node` with the ability to be initialised with a connection to a celestia-core endpoint over which it can query for account balances and submit transactions.

[ADR](https://github.com/celestiaorg/celestia-node/blob/main/docs/adr/adr-004-state-interaction.md) for reference.

Two notes: 
1. Although it was documented in the ADR, I decided *against* using the [lens](https://github.com/strangelove-ventures/lens) dependency as it is still in development  and introduced a bit more overhead / complexity than was necessary for two simple queries against the application.
2. It is extremely difficult to write unit tests that test the full functionality of `CoreAccess` as we lack a good way to mock the celestia-app for testing purposes. This will be discussed in the following week and will hopefully provide more clarity in how we can increase test coverage for this feature.

TODO: 
- [x] doc.go
- [ ] swamp test
- [x] start/stop hooks w/ DI
- [x] construction
- [x] lint
- [x] update ADR


node/services/config.go
```diff
@@ -17,10 +17,13 @@ type Config struct {
...
 	TrustedHash string
-	// TrustedPeer is the peer we trust to fetch headers from.
-	// Note: The trusted does *not* imply Headers are not verified, but trusted as reliable to fetch headers
-	// at any moment.
+	// TrustedPeer is the multiaddr of the peer we trust to fetch headers from.
+	// Note: The trusted does *not* imply Headers are not verified, but trusted
+	// as reliable to fetch headers at any moment.
 	TrustedPeer string
+
+	// TODO @renaynay: document
```
adlerjohn, line 25: Can we document this now?

adlerjohn (commented):
Lint error

node/services/config.go
adlerjohn: Documented in [b3b5707](https://github.com/celestiaorg/celestia-node/pull/420/commits/b3b5707466aa31385127857142f5ad8b834bcb1d)

liamsi (commented):
> Although it was documented in the ADR, I decided against using the [lens](https://github.com/strangelove-ventures/lens) dependency as it is still in development and introduced a bit more overhead / complexity than was necessary for two simple queries against the application

Do you mind quickly updating the ADR accordingly in this PR?

node/state/core.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+func CoreAccessor(
+	keyCfg services.KeyConfig,
+	storePath string,
+	coreEndpoint string,
+) (state.Accessor, error) {
+	encCfg := cosmoscmd.MakeEncodingConfig(app.ModuleBasics)
+
+	ring, err := keyring.New(app.Name, keyring.BackendFile, storePath, nil) // TODO @renaynay: user input?
```
liamsi, line 21: What does the comment refer to?

liamsi (approved):
👏🏼

node/state/core.go
renaynay: I'm passing nil here for `userInput` parameter, but not really sure about it.

service/state/core_access.go
```diff
@@ -0,0 +1,93 @@
...
+		coreEndpoint: endpoint,
+	}, nil
+}
+
+func (ca *CoreAccessor) Start(_ context.Context) error {
+	if ca.coreConn != nil {
+		return fmt.Errorf("core-access: already connected to core endpoint")
+	}
+	// dial given celestia-core endpoint
+	client, err := grpc.Dial(ca.coreEndpoint, grpc.WithTransportCredentials(insecure.NewCredentials()))
```
vgonkivs, line 47: Dial will send error immediately If there even a small chance that server could start later than client. I'd suggest to add [WithBlock](https://pkg.go.dev/google.golang.org/grpc#WithBlock) option + context with Timeout. In this case you need [DialContext](https://pkg.go.dev/google.golang.org/grpc#DialContext) instead of Dial

adlerjohn (commented):
Merge conflicts

renaynay:
Rebased.

renaynay:
FYI the test is currently broken as mock celestia-app isn't working.

adlerjohn (commented):
Lint errors!

service/state/core_access.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+// NewCoreAccessor dials the given celestia-core endpoint and
+// constructs and returns a new CoreAccessor with the active
+// connection.
+func NewCoreAccessor(
+	signer *apptypes.KeyringSigner,
+	encCfg cosmoscmd.EncodingConfig,
+	endpoint string,
+) (*CoreAccessor, error) {
```
adlerjohn, line 32: If this never fails, is there a reason for it to return an `error`?

service/state/core_access.go
```diff
@@ -0,0 +1,91 @@
...
+	// dial given celestia-core endpoint
+	client, err := grpc.DialContext(ctx, ca.coreEndpoint, grpc.WithBlock())
+	if err != nil {
+		return err
+	}
+	ca.coreConn = client
+	return nil
+}
+
+func (ca *CoreAccessor) Stop(_ context.Context) error {
```
adlerjohn, line 53:
```suggestion
func (ca *CoreAccessor) Stop(context.Context) error {
```

service/state/core_access.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	err := ca.coreConn.Close()
+	if err != nil {
+		return err
+	}
+	ca.coreConn = nil
+	return nil
+}
+
+func (ca *CoreAccessor) Balance(ctx context.Context) (*Balance, error) {
```
adlerjohn, line 65: Can you add docstrings to public methods? For example, it's completely non-obvious that `Balance()` returns the balance of a particular address.

renaynay: they're documented in the interface

adlerjohn: Oh I see

renaynay:
Test is failing due to failure in mock celestia-app

adlerjohn (commented):
Is CI expected to fail, or no?

renaynay:
Yes @adlerjohn

adlerjohn (approved):
utACK

node/services/config.go
```diff
@@ -59,3 +64,15 @@ func (cfg *Config) trustedHash() (tmbytes.HexBytes, error) {
 	}
 	return hex.DecodeString(cfg.TrustedHash)
 }
+
+// KeyConfig contains configuration details for construction of the keyring.Signer
+// that will be used to manage the node's account.
+type KeyConfig struct {
+	KeyringAccName string
```
Wondertan, line 71: I guess this is the name for a default keyring account name. Do we really need a way to configure its name? It does not seem to change anything meaningful for users.

To handle a default Node key we can:
* Have a name constant "default" which is the default for everyone and this is the key that is autogenerated on the first start of the node, so users would need to top it up with tokens to use it. This is the most simplest approach
* Have an API that allows users to generate/import a key/account and set it as default with an API. This is an approach we should get to eventually. 

In both cases, I don't see why users would want to change their default keyring name through a config.

Wondertan (changes requested):
image

Mostly looks good and straightforward. Good that the implementation turned out to be quite small even without a lens. T

here are only a few things that require changes, other than that are mostly nits, questions, and discussions on how things should look further.

service/state/core_access.go
renaynay: I am in favour of documenting them only in interface as there are less places for inconsistent docs if we do end up changing them in one place and not the other.

go.mod
```diff
@@ -43,12 +45,16 @@ require (
...
+	github.com/tendermint/spm v0.1.5
 	github.com/tendermint/tendermint v0.34.14
 	go.uber.org/fx v1.16.0
 	go.uber.org/zap v1.21.0
+	google.golang.org/grpc v1.43.0
 )
 
 replace (
+	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v0.44.1-celestia
+	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
```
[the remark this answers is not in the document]
renaynay: Nothing we can do 😭  thanks, cosmos-sdk 😭

docs/adr/adr-004-state-interaction.md
```diff
@@ -19,95 +20,113 @@ and public state of others.
...
 
 ### 2. P2P Implementation of `StateAccessor`: `P2PAccess`
 
-While it is not necessary to detail how `P2PAccess` will be implemented in this ADR, it will still conform to the 
-`StateAccessor` interface, but instead of being provided a core endpoint to connect to via RPC and using `lens.ChainClient`
-for state-related queries, `P2PAccess` will perform service discovery of state-providing nodes in the network and perform
-the state queries via libp2p streams. More details of the p2p implementation will be described in a separate dedicated ADR.
+While it is not necessary to detail how `P2PAccess` will be implemented in this ADR, it will still conform to the
+`StateAccessor` interface, but instead of being provided a core endpoint to connect to via RPC and
+using `lens.ChainClient`
```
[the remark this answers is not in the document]
renaynay: resolved?

service/state/interface.go
```diff
@@ -0,0 +1,22 @@
...
+	// Start starts the state Accessor.
+	Start(context.Context) error
+	// Stop stops the state Accessor.
+	Stop(context.Context) error
+	// Balance retrieves the Celestia coin balance
+	// for the node's account/signer.
+	Balance(ctx context.Context) (*Balance, error)
+	// BalanceForAddress retrieves the Celestia coin balance
+	// for the given address.
+	BalanceForAddress(ctx context.Context, addr string) (*Balance, error)
```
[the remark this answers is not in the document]
renaynay: could use `types.AccAddress` ?

```go
// AccAddress a wrapper around bytes meant to represent an account address.
// When marshaled to a string or JSON, it uses Bech32.
type AccAddress []byte
```

wdyt @Wondertan ?

node/components.go
```diff
@@ -23,6 +25,20 @@ func lightComponents(cfg *Config, store Store) fxutil.Option {
...
+				cfg.Core.RemoteConfig.RemoteAddr)
+			if err != nil {
+				return nil, err
+			}
+			lc.Append(fx.Hook{
+				OnStart: ca.Start,
+				OnStop:  ca.Stop,
+			})
+			return ca, nil
+		}),
```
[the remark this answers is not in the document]
renaynay: Yeah I can do that. The main issue is -- we wanted full node to be able to *provide* state rather than use a core accessor to access it. I think that having that functionality available in light node for now is good. I can also add it to *bridge* node since it already receives a core endpoint anyway. Wouldn't hurt to implement core_access there as well. Wdyt regarding *full* node?

go.sum
```diff
@@ -53,22 +54,40 @@ dmitri.shuralyov.com/gpu/mtl v0.0.0-20190408044501-666a987793e9/go.mod h1:H6x//7
...
 dmitri.shuralyov.com/state v0.0.0-20180228185332-28bcc343414c/go.mod h1:0PRwlb0D6DFvNNtx+9ybjezNCa8XF0xaYcETyp6rHWU=
+filippo.io/edwards25519 v1.0.0-beta.2 h1:/BZRNzm8N4K4eWfK28dL4yescorxtO7YG1yun8fy+pI=
+filippo.io/edwards25519 v1.0.0-beta.2/go.mod h1:X+pm78QAUPtFLi1z9PYIlS/bdDnvbCOGKtZ+ACWEf7o=
 git.apache.org/thrift.git v0.0.0-20180902110319-2566ecd5d999/go.mod h1:fPE2ZNJGynbRyZ4dJvy6G277gSllfV2HJqblrnkyeyg=
+github.com/99designs/keyring v1.1.6 h1:kVDC2uCgVwecxCk+9zoCt2uEL6dt+dfVzMvGgnVcIuM=
+github.com/99designs/keyring v1.1.6/go.mod h1:16e0ds7LGQQcT59QqkTg72Hh5ShM51Byv5PEmW6uoRU=
 github.com/AndreasBriese/bbloom v0.0.0-20180913140656-343706a395b7/go.mod h1:bOvUY6CB00SOBii9/FifXqc0awNKxLFCL/+pkDPuyl8=
 github.com/AndreasBriese/bbloom v0.0.0-20190306092124-e2d15f34fcf9/go.mod h1:bOvUY6CB00SOBii9/FifXqc0awNKxLFCL/+pkDPuyl8=
 github.com/AndreasBriese/bbloom v0.0.0-20190825152654-46b345b51c96/go.mod h1:bOvUY6CB00SOBii9/FifXqc0awNKxLFCL/+pkDPuyl8=
+github.com/Azure/azure-pipeline-go v0.2.1/go.mod h1:UGSo8XybXnIGZ3epmeBw7Jdz+HiUVpqIlpz/HKHylF4=
```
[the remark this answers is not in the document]
renaynay: :/

service/state/core_access.go
```diff
@@ -0,0 +1,92 @@
...
+
+	"github.com/celestiaorg/celestia-app/app"
+	apptypes "github.com/celestiaorg/celestia-app/x/payment/types"
+)
+
+// CoreAccessor implements Accessor over an RPC connection
+// with a celestia-core node.
+type CoreAccessor struct {
+	signer *apptypes.KeyringSigner
+	encCfg cosmoscmd.EncodingConfig
```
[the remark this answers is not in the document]
renaynay: resolved?

service/state/core_access.go
Wondertan: I agree with you, but we should also be on the same page with @liamsi

node/services/config.go
renaynay: Have taken the first approach. Resolved?

service/state/core_access.go
```diff
@@ -0,0 +1,92 @@
...
+	ca.coreConn = nil
+	return nil
+}
+
+func (ca *CoreAccessor) Balance(ctx context.Context) (*Balance, error) {
+	return ca.BalanceForAddress(ctx, ca.signer.GetSignerInfo().GetAddress().String())
+}
+
+func (ca *CoreAccessor) BalanceForAddress(ctx context.Context, addr string) (*Balance, error) {
+	queryCli := banktypes.NewQueryClient(ca.coreConn)
```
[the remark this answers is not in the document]
renaynay: See https://github.com/celestiaorg/celestia-node/pull/420/commits/8959e791e5355100dca4221558943657da1bedfa

service/state/core_access.go
```diff
@@ -0,0 +1,92 @@
...
+
+	balReq := &banktypes.QueryBalanceRequest{
+		Address: addr,
+		Denom:   app.DisplayDenom,
+	}
+
+	balResp, err := queryCli.Balance(ctx, balReq)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
renaynay: i like having req defined separately.

node/components.go
```diff
@@ -23,6 +25,20 @@ func lightComponents(cfg *Config, store Store) fxutil.Option {
 		baseComponents(cfg, store),
 		fxutil.Provide(services.DASer),
 		fxutil.Provide(services.HeaderExchangeP2P(cfg.Services)),
+		// state components
+		fxutil.ProvideIf(cfg.Core.Remote, state.NewService),
+		fxutil.ProvideIf(cfg.Core.Remote, func(lc fx.Lifecycle) (state.Accessor, error) {
```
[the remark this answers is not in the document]
renaynay: Because it's the only way to tell right now whether light node is running with a core connection.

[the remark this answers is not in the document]
renaynay: This is to prevent adding an additional flag.

node/state/core.go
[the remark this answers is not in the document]
renaynay: can we expose the path as a method on `keystore.Keystore` then? I can't just pass a string to `lightComponents`, I'd have to pass the whole keystore bc of fx

service/state/core_access.go
liamsi: Hmm, can you have a look on how they look on godoc? I'd make the decision depending on that: how are the implemented methods displayed and is there a connection from the impl to the interface on godoc.

node/components.go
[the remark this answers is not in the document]
Wondertan: Ok, I see

node/state/core.go
liamsi: I think this is to get the user's password for the file: https://github.com/cosmos/cosmos-sdk/blob/20735e792c85a0d84ca32e97a0a3c814bcc3529f/crypto/keyring/keyring.go#L671

node/components.go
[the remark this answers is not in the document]
Wondertan: Until we don't implement the FullNode state computation there won't be a way for a full node operator to access the state, considering that we don't support running light and full nodes together(ref to a different ports discussion), so I think it is fine for a full node to rely on a Core as long as this is not a requirement.

service/state/core_access.go
```diff
@@ -0,0 +1,92 @@
...
+
+	return balResp.Balance, nil
+}
+
+func (ca *CoreAccessor) SubmitTx(ctx context.Context, tx Tx) (*TxResponse, error) {
+	txResp, err := apptypes.BroadcastTx(ctx, ca.coreConn, sdk_tx.BroadcastMode_BROADCAST_MODE_SYNC, tx)
```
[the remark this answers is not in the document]
Wondertan: Refer to sdk_tx pkg for a description of a difference between modes.  

Regarding sync and async methods: I would keep only one method with options that allow setting the mode(this is what I am doing with PublicAPI)

node/components.go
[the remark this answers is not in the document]
Wondertan: SGTM

service/state/core_access.go
Wondertan: Good point, cc @renaynay

node/state/core.go
liamsi: Can we manually test this before merging? I think in the SDK this is used with Stdin. The userInput is fetched from this context:
https://github.com/cosmos/cosmos-sdk/blob/fece3f0485dd418a2bc54ecc61f17da3303192c7/simapp/simd/cmd/root.go#L46

service/state/core_access.go
liamsi: ref: https://github.com/celestiaorg/celestia-node/issues/426#issue-1131238570

node/state/core.go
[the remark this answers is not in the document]
Wondertan: Yes. I think it is fine to pass the Keystore to fx(I guess we are doing this already) and to state.Accessor contructor

service/state/interface.go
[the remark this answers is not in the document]
Wondertan: I think it is fine. I also checked the SDK and they have an interface that AccAddress supports and I think we should use the interface instead.

service/state/core_access.go
renaynay: Resolved?

node/state/core.go
renaynay: I manually tested it and it does not panic.

renaynay: what is the userInput used for?

liamsi: To get the user's input for the password, which is used to encrypt the file. If it does not panic, it's OK for now. We should probably support all the other backends as well though.

service/state/core_access.go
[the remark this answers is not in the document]
renaynay: I changed default to `MODE_BLOCK`

should I add this method as well

```go
func (ca *CoreAccessor) SubmitTxWithOptions(ctx context.Context, tx Tx, mode sdk_tx.BroadcastMode) (*TxResponse, error) {
	txResp, err := apptypes.BroadcastTx(ctx, ca.coreConn, mode, tx)
	if err != nil {
		return nil, err
	}
	return txResp.TxResponse, nil
}
```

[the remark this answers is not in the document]
renaynay: Pls check @Wondertan: https://github.com/celestiaorg/celestia-node/pull/420/commits/f05d22b8eba7eede502ec4dc270e636db1b8be94

node/state/core.go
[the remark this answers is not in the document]
renaynay: Resolved https://github.com/celestiaorg/celestia-node/pull/420/commits/59d2a7c6091607692aad8506047024ab6b615164 ?

node/components.go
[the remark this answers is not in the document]
renaynay: Resolved https://github.com/celestiaorg/celestia-node/pull/420/commits/46f49d2feec485267a3f995907c9240ef628ef70 ?

service/state/interface.go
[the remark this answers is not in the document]
renaynay: Resolved https://github.com/celestiaorg/celestia-node/pull/420/commits/46f49d2feec485267a3f995907c9240ef628ef70 ?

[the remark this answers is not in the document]
Wondertan: No, I meant [this](https://github.com/celestiaorg/cosmos-sdk/blob/master/types/address.go#L104) interface

node/components.go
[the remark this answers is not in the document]
Wondertan: No. What about Bridge? Don't see a reason for it to be restricted to state access, so let's put this in baseComponents.
> Any node should be able to access the state

service/state/core_access.go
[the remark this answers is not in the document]
Wondertan: Not necessary part pf this PR.
> should I add this method as well

Wondertan:
@renaynay, this PR is fully broken

node/state/core.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/celestiaorg/celestia-node/service/state"
+)
+
+var keyringAccName = "default"
+
+func CoreAccessor(
+	ks keystore.Keystore,
+	coreEndpoint string,
+) (state.Accessor, error) {
+	// TODO @renaynay: Include optioin for setting custom `userInput` parameter with
```
Wondertan, line 19:
```suggestion
	// TODO @renaynay: Include option for setting custom `userInput` parameter with
```

service/state/interface.go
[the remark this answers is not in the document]
Wondertan: Also, pls make a local alias to that interface, like you did for other types.

node/components.go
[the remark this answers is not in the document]
renaynay: https://github.com/celestiaorg/celestia-node/pull/420/commits/9e3e46bc16cfd28172f5927e1a32856380a27eeb

Now?

Wondertan (approved):
Ok lets go

service/state/core_access.go
liamsi: What was the conclusion here?

service/state/core_access_test.go
```diff
@@ -0,0 +1,31 @@
+package state
+
+/*
+The test below has been commented out for now as the mock celestia-app has issues
+initializing the chain, and therefore cannot be used properly in the test yet.
+*/
```
liamsi, line 6: I'd prefer a skipped test instead of commented out. Also an issue that describes the necessary steps to unblock the test.

liamsi (approved):
LFG

adlerjohn (approved):
Issues for the open conversations should be opened, but not going to block on them.

service/state/core_access_test.go
renaynay: Deleted test, tracking [here](https://github.com/celestiaorg/celestia-node/issues/570)

service/state/core_access.go
renaynay: Conclusion: if it implements an interface, those methods do not need to be documented in the implementation (only in the interface) as it reduces surface-area for having discrepancies in documentation. Any methods outside of interface impl on that struct must be documented, though.

PR #428, opened 2022-02-11
feat(service/header): HeightSub

Wondertan:
Improved version of HeightSub adapted from https://github.com/celestiaorg/celestia-node/pull/397 and integrated into the Store, instead of Syncer.

Review commit by commit

Closes: https://github.com/celestiaorg/celestia-node/issues/444

### DIFF
--- a/service/header/store.go
+++ b/service/header/store.go
@@ -110,6 +110,9 @@ func (s *store) Get(_ context.Context, hash tmbytes.HexBytes) (*ExtendedHeader, 
 }
 
 func (s *store) GetByHeight(ctx context.Context, height uint64) (*ExtendedHeader, error) {
+	// here we subscribe to the height only in case current head is before the requested 'height'
+	// and if it is, we expect the Sub to return us either error or the header itself
+	// otherwise, Sub will return us both nils, signaling that we have to get the header from the Store
 	h, err := s.heightSub.Sub(ctx, height)
 	if h != nil || err != nil {
 		return h, err
--- a/service/header/store.go
+++ b/service/header/store.go
@@ -28,6 +28,8 @@ type store struct {
 
 	headLk sync.RWMutex
 	head   tmbytes.HexBytes
+
+	heightSub *heightSub
 }
 
 // NewStore constructs a Store over datastore.
@@ -71,9 +73,10 @@ func newStore(ds datastore.Batching) (*store, error) {
 	}
 
 	return &store{
-		ds:    ds,
-		cache: cache,
-		index: index,
+		ds:        ds,
+		cache:     cache,
+		index:     index,
+		heightSub: newHeightSub(),
 	}, nil
 }
 
@@ -120,6 +123,11 @@ func (s *store) Get(_ context.Context, hash tmbytes.HexBytes) (*ExtendedHeader, 
 }
 
 func (s *store) GetByHeight(ctx context.Context, height uint64) (*ExtendedHeader, error) {
+	h, err := s.heightSub.Sub(ctx, height)
+	if h != nil || err != nil {
+		return h, err
+	}
+
 	hash, err := s.index.HashByHeight(height)
 	if err != nil {
 		if err == datastore.ErrNotFound {
@@ -222,6 +230,7 @@ func (s *store) Append(ctx context.Context, headers ...*ExtendedHeader) error {
 	}
 
 	log.Infow("new head", "height", head.Height, "hash", head.Hash())
+	s.heightSub.Pub(verified...)
 	return nil
 }
 
--- a/service/header/store.go
+++ b/service/header/store.go
@@ -110,13 +110,14 @@ func (s *store) Get(_ context.Context, hash tmbytes.HexBytes) (*ExtendedHeader, 
 }
 
 func (s *store) GetByHeight(ctx context.Context, height uint64) (*ExtendedHeader, error) {
-	// here we subscribe to the height only in case current head is before the requested 'height'
-	// and if it is, we expect the Sub to return us either error or the header itself
-	// otherwise, Sub will return us both nils, signaling that we have to get the header from the Store
+	// if the requested 'height' was not yet published
+	// we subscribe and for it
 	h, err := s.heightSub.Sub(ctx, height)
-	if h != nil || err != nil {
+	if err != errElapsedHeight {
 		return h, err
 	}
+	// otherwise, the errElapsedHeight is thrown,
+	// which means the requested 'height' should be already stored
 
 	hash, err := s.index.HashByHeight(height)
 	if err != nil {
--- a/service/header/store.go
+++ b/service/header/store.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"strconv"
-	"sync"
 
 	lru "github.com/hashicorp/golang-lru"
 	"github.com/ipfs/go-datastore"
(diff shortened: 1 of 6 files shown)


service/header/store_heightsub.go
```diff
@@ -20,16 +20,21 @@ func newHeightSub() *heightSub {
 	}
 }
 
+// Height reports current height.
+func (hs *heightSub) Height() uint64 {
```
renaynay, line 24: Can we call it `LatestAvailableHeight` or something? I'm in favour of higher verbosity.

renaynay (commented):
beautiful

service/header/store_heightsub.go
Wondertan: I don't see where is the boundary between `LatestAvailableHeight` and `LatestAvailableHeightOfSubjectiveCelestiaDABlockchain` tbh

Wondertan: `HeadHeight`as compromise?

service/header/store_heightsub.go
```diff
@@ -0,0 +1,77 @@
...
+// newHeightSub instantiates new heightSub.
+func newHeightSub() *heightSub {
+	return &heightSub{
+		heightReqs: make(map[uint64][]chan *ExtendedHeader),
+	}
+}
+
+// Sub subscribes for a header of a given height.
+// It can return both values as nil, which means a requested header was already provided
+// and caller should get it elsewhere.
```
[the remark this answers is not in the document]
Wondertan: Yes, from the store. In other words, it only subscribes for future heights and for known returns both nils.

Wondertan:
@renaynay, I added one more commit, which stops tracking head on the Store and starts to rely on heightSub.Height. Referencing the highest header in two places is a no-no and now only one in heightSub is present.

service/header/store_heightsub.go
```diff
@@ -0,0 +1,77 @@
...
+
+import (
+	"context"
+	"sync"
+	"sync/atomic"
+)
+
+// heightSub provides a minimalistic mechanism to wait till header for a height becomes available.
+type heightSub struct {
+	height       uint64 // atomic
```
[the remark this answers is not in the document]
renaynay: Okay I guess I don't care what it's called, maybe just add a doc to the field in the struct: 

```
// height refers to the latest locally available block height that has been fully verified and inserted into the chain.
```
? Compromise?

service/header/store_heightsub.go
renaynay: Refer to: 
https://github.com/celestiaorg/celestia-node/pull/428#discussion_r805345176

renaynay (commented):
All LGTM, except check out this err: https://github.com/celestiaorg/celestia-node/runs/5173909128?check_suite_focus=true#step:4:402

service/header/store_heightsub.go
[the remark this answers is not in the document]
Wondertan: Yep, good comment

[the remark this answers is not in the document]
Wondertan: done

renaynay (dismissed):
❤️

service/header/store.go
```diff
@@ -120,6 +110,11 @@ func (s *store) Get(_ context.Context, hash tmbytes.HexBytes) (*ExtendedHeader,
 }
 
 func (s *store) GetByHeight(ctx context.Context, height uint64) (*ExtendedHeader, error) {
+	h, err := s.heightSub.Sub(ctx, height)
+	if h != nil || err != nil {
```
liamsi, line 114: What is `h` and why do we return if it is != nil? 

Ah, I see the extended header. The flow here is not obvious: So we either return if we get and extended header on subscribe (or error), or, we have get it from the store (below)? I think comments would help here.

service/header/store_heightsub.go
```diff
@@ -0,0 +1,84 @@
...
+func (hs *heightSub) Sub(ctx context.Context, height uint64) (*ExtendedHeader, error) {
+	if hs.Height() >= height {
+		return nil, nil
+	}
+
+	hs.heightReqsLk.Lock()
+	if hs.Height() >= height {
+		// This is a rare case we have to account for.
+		// The lock above can park a goroutine long enough for hs.height to change for a requested height,
+		// leaving the request never fulfilled and the goroutine deadlocked.
```
liamsi, line 42: Good comment 👍🏼

service/header/store_heightsub.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+}
+
+// Pub processes all the outstanding subscriptions matching the given headers.
+// Pub is safe to be called from one goroutine.
+func (hs *heightSub) Pub(headers ...*ExtendedHeader) {
+	height := hs.Height()
+	from, to := uint64(headers[0].Height), uint64(headers[len(headers)-1].Height)
+	if height != 0 && height+1 != from {
+		log.Warnf("PLEASE REPORT THE BUG: headers given to the heightSub are in the wrong order")
```
liamsi, line 63: Shouldn't this be log.Errorf if this indicates a bug?

service/header/store_heightsub.go
```diff
@@ -0,0 +1,84 @@
...
+		return
+	}
+	atomic.StoreUint64(&hs.height, to)
+
+	hs.heightReqsLk.Lock()
+	defer hs.heightReqsLk.Unlock()
+
+	// instead of looping over each header in 'headers', we can loop over each request
+	// which will drastically decrease idle iterations, as there will be lesser requests than the headers
```
liamsi, line 72:
```suggestion
	// which will drastically decrease idle iterations, as there will be less requests than headers
```

liamsi (commented):
Needs an issue in the opening comment

service/header/store_heightsub.go
Wondertan: Good point. Also, I think we should go even further and made it a Fatal error that terminates the node, otherwise storage might be unrecoverable

Wondertan: Done.

service/header/store.go
Wondertan: Commented.

Bidon15 (approved):
Works on a drafted swamp test locally (need a clock fix to make it work in CI)

service/header/store_heightsub.go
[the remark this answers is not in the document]
renaynay: ack

service/header/store_heightsub.go
renaynay: ack

service/header/store.go
liamsi: Good comment. I'd usually prefer if stuff like this was expressed via code directly. But I think this is good as is.

service/header/store_heightsub.go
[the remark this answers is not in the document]
liamsi: Agree that this could be more expressive. see also: https://github.com/celestiaorg/celestia-node/pull/428/files#r806313889
```go
if hs.Height() >= height {
	return nil, errElapsedHeight
}
```

liamsi (approved):

service/header/store_heightsub.go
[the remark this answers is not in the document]
Wondertan: Changed to throw an error.

service/header/store.go
Wondertan: Changed to error

renaynay (approved):

Wondertan:
I am resolving all the comments and merging

PR #431, opened 2022-02-12
refactor(service/header): rework on disk writing strategy of the Store

Wondertan:
The PR makes all the writing to Store happen through writeLoop in a separate routine so that
	* Appends are not blocked by long disk IO writes and DB compactions
	* We can batch header writes
	
Also, this PR makes Append writes atomic. Meaning that all the required writes are done in one single batch. Previously, each Append did 3 full io write requests to the disk for:
* Headers themselves
* Highest head reference
* Height indices
Now we make all these writes atomically in one io write additionally to collecting multiple Appends in one big append, leading to a way lesser io writes.
	
Based on #430 
Closes https://github.com/celestiaorg/celestia-node/issues/448

Review commit by commit

### DIFF
--- a/service/header/store.go
+++ b/service/header/store.go
@@ -19,6 +19,9 @@ var (
 	DefaultIndexCacheSize = 2048
 )
 
+// errStoppedStore is returned on operations over stopped store
+var errStoppedStore = errors.New("header: stopped store")
+
 // store implement Store interface for ExtendedHeader over Datastore.
 type store struct {
 	// header storing
@@ -110,6 +113,11 @@ func (s *store) Start(ctx context.Context) error {
 }
 
 func (s *store) Stop(ctx context.Context) error {
+	select {
+	case <-s.writesDn:
+		return errStoppedStore
+	default:
+	}
 	// signal to prevent further writes to Store
 	s.writes <- nil
 	select {
@@ -246,6 +254,8 @@ func (s *store) Append(ctx context.Context, headers ...*ExtendedHeader) error {
 		// we return an error here after writing,
 		// as there might be an invalid header in between of a given range
 		return err
+	case <-s.writesDn:
+		return errStoppedStore
 	case <-ctx.Done():
 		return ctx.Err()
 	}
--- a/service/header/store.go
+++ b/service/header/store.go
@@ -25,7 +25,7 @@ var (
 // errStoppedStore is returned for attempted operations on a stopped store
 var errStoppedStore = errors.New("header: stopped store")
 
-// store implement Store interface for ExtendedHeader over Datastore.
+// store implements the Store interface for ExtendedHeaders over Datastore.
 type store struct {
 	// header storing
 	//
--- a/service/header/store.go
+++ b/service/header/store.go
@@ -46,7 +46,7 @@ type store struct {
 	//
 	// queued of header ranges to be written
 	writes chan []*ExtendedHeader
-	// signals writes being done
+	// signals when writes are finished
 	writesDn chan struct{}
 	// pending keeps headers pending to be written in a batch
 	pending []*ExtendedHeader
--- a/service/header/store.go
+++ b/service/header/store.go
@@ -60,13 +60,13 @@ func NewStore(ds datastore.Batching) (Store, error) {
 }
 
 // NewStoreWithHead initiates a new Store and forcefully sets a given trusted header as head.
-func NewStoreWithHead(ds datastore.Batching, head *ExtendedHeader) (Store, error) {
+func NewStoreWithHead(ctx context.Context, ds datastore.Batching, head *ExtendedHeader) (Store, error) {
 	store, err := newStore(ds)
 	if err != nil {
 		return nil, err
 	}
 
-	return store, store.Init(context.TODO(), head)
+	return store, store.Init(ctx, head)
 }
 
 func newStore(ds datastore.Batching) (*store, error) {
--- a/service/header/store.go
+++ b/service/header/store.go
@@ -17,6 +17,9 @@ var (
 	DefaultStoreCacheSize = 10240
 	// DefaultIndexCacheSize defines the amount of max entries allowed in the Height to Hash index cache.
 	DefaultIndexCacheSize = 2048
+	// DefaultWriteBatchSize defines the size of the batched header write.
+	// Headers are written in batches not to thrash the underlying Datastore with writes.
+	DefaultWriteBatchSize = 2048
 )
 
 // errStoppedStore is returned on operations over stopped store
@@ -45,6 +48,8 @@ type store struct {
 	writes chan []*ExtendedHeader
 	// signals writes being done
 	writesDn chan struct{}
+	// pending keeps headers pending to be written in a batch
+	pending []*ExtendedHeader
 }
 
 // NewStore constructs a Store over datastore.
@@ -81,14 +86,15 @@ func newStore(ds datastore.Batching) (*store, error) {
 		cache:       cache,
 		heightIndex: index,
 		heightSub:   newHeightSub(),
-		writes:      make(chan []*ExtendedHeader, 8),
+		writes:      make(chan []*ExtendedHeader, 16),
 		writesDn:    make(chan struct{}),
+		pending:     make([]*ExtendedHeader, 0, DefaultWriteBatchSize),
 	}, nil
 }
 
 func (s *store) Init(_ context.Context, initial *ExtendedHeader) error {
 	// trust the given header as the initial head
-	err := s.write(initial)
+	err := s.flush(initial)
 	if err != nil {
 		return err
 	}
@@ -108,7 +114,7 @@ func (s *store) Start(ctx context.Context) error {
 		s.heightSub.SetHeight(uint64(head.Height))
 	}
 
-	go s.writeLoop()
+	go s.flushLoop()
 	return nil
 }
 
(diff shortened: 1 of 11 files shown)


service/header/store.go
```diff
@@ -191,28 +241,71 @@ func (s *store) Append(ctx context.Context, headers ...*ExtendedHeader) error {
...
+		err := s.flush(s.pending...)
+		if err != nil {
+			// TODO(@Wondertan): Should this be a fatal error case with os.Exit?
+			from, to := uint64(headers[0].Height), uint64(headers[len(headers)-1].Height)
+			log.Errorw("writing header batch", "from", from, "to", to)
+		}
+		// reset pending
+		s.pending = s.pending[:0]
+
+		if headers == nil {
```
renaynay, line 290: so if the writes queue ever ends up empty for whatever reason, `flushLoop` will exit and never be called at any point again?

service/header/store_heightsub.go
```diff
@@ -25,6 +25,11 @@ func (hs *heightSub) Height() uint64 {
 	return atomic.LoadUint64(&hs.height)
 }
 
+// SetHeight sets the new head height for heightSub.
+func (hs *heightSub) SetHeight(height uint64) {
```
renaynay, line 29: Does this need to be called from anywhere outside the header pkg?

Wondertan: No, same for Pub and Sub and Height.

Wondertan: Even though their names are public, those methods will never be accessed outside the pkg, as the constructor is private. 
I keep methods capitalized even for structs in the same pkg. This ways it's clearer that some code call the method from a dependency structure.

service/header/store.go
renaynay: But could there be a case where `writes` queue can be empty but the node is still technically syncing / trying to stay up to head with the network? What if there's a network issue?

Wondertan: What do you mean by empty writes queue? A nil case?

Wondertan: Look at append. It will only send non empty values to the writed queue

renaynay: No i mean what happens if all `writes` have been flushed and no new `writes` have been added even though the node is still operational?

service/header/store_heightsub.go
renaynay: I mean the only time I ever see private structs with exported methods are when struct is implementing an interface. Is it really necessary to export them when this isn't the case?

service/header/store.go
```diff
@@ -191,28 +241,71 @@ func (s *store) Append(ctx context.Context, headers ...*ExtendedHeader) error {
...
+		if len(s.pending) < DefaultWriteBatchSize && headers != nil {
+			continue
+		}
 
-	log.Infow("new head", "height", head.Height, "hash", head.Hash())
-	s.heightSub.Pub(verified...)
-	return nil
+		err := s.flush(s.pending...)
+		if err != nil {
+			// TODO(@Wondertan): Should this be a fatal error case with os.Exit?
```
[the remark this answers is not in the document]
Wondertan: It's a good thing as this PR makes all the writes atomic. Meaning that all the required writes are done in one single batch. Previously, each Append did 3 full io write requests to the disk for:
* Headers themselves
* New head reference or head hash
* Height indices

Now we make all these writes atomically in one io write. Furthermore, we now collect multiple appends in one big append.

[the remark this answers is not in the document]
Wondertan: So to answer your question, if this one big atomic write fails, then nothing would be written and there won't be any db corruption and issues with header adjacency.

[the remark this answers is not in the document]
Wondertan: This is true though until a case where one random write fails but a subsequent ones not, but AFAIU the underlying datastore won't accept more writes if one of them failed.

service/header/store_heightsub.go
Wondertan: It's not necessary but it is a code style we have throughout the project. IMO, it makes things clearer. The rule is: a structure that is meant to be used privately inside a pkg, should have capitalized methods if those methods are meant to be used by other structures within the same pkg. Non-capitalized methods of a structure are meant to be used only within the structure.

service/header/store.go
Wondertan: Nothing happens. Nothing to flush, nothing to writes, so goroutine just blocked on `writes` chan.

service/header/store.go
```diff
@@ -17,14 +17,39 @@ var (
 	DefaultStoreCacheSize = 1024
 	// DefaultIndexCacheSize defines the amount of max entries allowed in the Height to Hash index cache.
 	DefaultIndexCacheSize = 256
+	// DefaultWriteBatchSize defines the size of the batched header write.
+	// Headers are written in batches not to thrash the underlying Datastore with writes.
+	DefaultWriteBatchSize = 2056
```
liamsi, line 22: 2056 headers? 2056 bytes? Also why 2056???

service/header/store.go
```diff
@@ -17,14 +17,39 @@ var (
...
+	// manages current store head height (1) and
+	// allows callers to wait till header for a height is stored (2)
 	heightSub *heightSub
+
+	// writing to datastore
+	//
+	// queued of header ranges to be written
+	writes chan []*ExtendedHeader
+	// signals when writes are finished
+	writesDn chan struct{}
```
liamsi, line 50: DN 🤣 

Does this stand for done in this case?

service/header/store_heightsub.go
```diff
@@ -55,15 +60,17 @@ func (hs *heightSub) Sub(ctx context.Context, height uint64) (*ExtendedHeader, e
...
-// Pub is safe to be called from one goroutine.
+// Pub is only safe when called from one goroutine.
+// For Pub to work correctly, heightSub has to be initialized with SetHeight
+// so that given headers are contiguous to the height on heightSub.
 func (hs *heightSub) Pub(headers ...*ExtendedHeader) {
 	height := hs.Height()
 	from, to := uint64(headers[0].Height), uint64(headers[len(headers)-1].Height)
-	if height != 0 && height+1 != from {
+	if height+1 != from {
 		log.Fatal("PLEASE REPORT THE BUG: headers given to the heightSub are in the wrong order")
```
liamsi, line 70:
```suggestion
		log.Fatal("PLEASE FILE A BUG REPORT: headers given to the heightSub are in the wrong order")
```

liamsi (commented):
Needs issue in opening comment. Issue should answer: What does rework mean? Why is it necessary.

service/header/store_heightsub.go
renaynay: Agh i really don't like `SetHeight` being exported. I know it's not reachable but still.

service/header/store.go
Wondertan: Yes, previously there was an explicit done word in the commit, but I applied Rene's suggestion which changed it to finished.

service/header/store.go
Wondertan: Headers.

service/header/store_heightsub.go
Wondertan: Why specifically SetHeight and not others though. Is this because it is a setter that feels public while it is not?

service/header/store.go
Wondertan: Hmm, I have no idea why I put 2056 😅. I wanted it to be 2048, lol

Wondertan: Made it 2048.

service/header/store.go
Wondertan: Not sure if this is resolved. Happy to discuss if there is still something unclear.

service/header/store_heightsub.go
renaynay: Yes.

renaynay (commented):
I think all LGTM, except for exported methods on heightSub - I'd prefer for them to be private just for clarity purposes, but since they're not really accessible anyway, I won't *die* on this hill.

service/header/store.go
renaynay: This is utACKed but I do want to have a call about it later at some point because I don't quite understand how it's not possible in practice that a node could (due to network issues) not receive any new headers from the network and then have writes queue flushed out in the process. It seems like, even though it's unlikely, it's still a possibility. I'm fine w merging as is now but can we talk about it later and maybe write an issue if you see merit in my concern?

Wondertan:
> I'd prefer for them to be private just for clarity purposes

@renaynay, look at this from the other perspective as well. Such public methods add more clarity with an internal API and help devs understand the code and and instantly see relations between the components within a package, especially when a package is big like `service/headers` one.

service/header/store.go
Wondertan: @renaynay, If there is no writes being enqueue, the pending cache won't grow beyond `DefaultWriteBatchSize` and no flush will happen. This is true, but there is no problem with this as:
* Those headers are still available in the Store cache and anyone can you use them without accessing disk
* The most important one, If the Store is stopped than the flush is mandatory(this is why there is a special case for nil headers) even when pending cache is not big enough, so there won't be case that we miss a flush or some headers.
> then have writes queue flushed out in the process

renaynay: okay so reading `headers := range writes` will never come up empty? My concern not with the flushing, but with accidental exit of the flushLoop with no way to kickstart it again unless node restart.

renaynay: That's the only question i have ^

Wondertan: There are no accidental exits of the flushLoop, but purposeful, by the Stop only.

Wondertan: Note that I didn't aimed for restartability in this case and there might be issues if Store is getting restarted with Start and Stop multiple times during runtime. I don't see any reason for Store to be restarted tbh, but making this possible would not be hard.

liamsi (approved):
Agree that the exported (but unnecessary) setters are somewhat confusing. But nbd. I'd rather keep the public API smaller tbh.

renaynay (approved):
ок

PR #432, opened 2022-02-12
perf(services/header): increase Store caches sizes

Wondertan:
Note that I have no idea how to calculate a universal value for cache sizes in our case, so this is mostly playing around and experimenting based on a feeling that the initial values are too conservative.


renaynay (approved):
also utACK but please let's make an issue for testing this.

Wondertan:
@adlerjohn, the amount of IOs our bridge does when it helps the light client to sync is too big, while there is still plenty of ram. Also, this will indirectly accelerate future DASing of past headers.

Wondertan:
@renaynay, this increase can not break anything, but will only eat more RAM, so don't think we should test anything here.

PR #435, opened 2022-02-13
fix(service/header): allow some clock drift during verification

Wondertan:
So, I've noticed that sometimes tests can fail due to an invalid header being from the future by tiny fractions of time. I didn't look deeper into why this happens, but here is the solution that is also present in tendermint light client. Similar to them, this PR allows a 10 seconds clock drift from 'now' time during verification. Tendermint also allows configuring this field, just because. In contrast, this PR does not allow configuring this value(besides tests if needed).

The Tendermint's [`maxClockDrift`](https://github.com/tendermint/tendermint/blob/master/light/verifier.go#L30) param. 
> IMO, this param makes sense for testing only and is pointless to users while only opening a door for users to shoot into their own legs while playing with configs.

### DIFF
--- a/service/header/header_verify.go
+++ b/service/header/header_verify.go
@@ -80,7 +80,8 @@ func (eh *ExtendedHeader) verify(untrst *ExtendedHeader) error {
 
 	now := time.Now()
 	if !untrst.Time.Before(now.Add(clockDrift)) {
-		return fmt.Errorf("new untrusted header has a time from the future %v (now: %v)", untrst.Time, now)
+		return fmt.Errorf(
+			"new untrusted header has a time from the future %v (now: %v, clockDrift: %v)", untrst.Time, now, clockDrift)
 	}
 
 	return nil


service/header/header_verify.go
```diff
@@ -96,6 +64,28 @@ func (eh *ExtendedHeader) VerifyAdjacent(untrst *ExtendedHeader) error {
...
+		return fmt.Errorf("new untusted header has different chain %s, not %s", untrst.ChainID, eh.ChainID)
+	}
+
+	if !untrst.Time.After(eh.Time) {
+		return fmt.Errorf("expected new untrusted header time %v to be after old header time %v", untrst.Time, eh.Time)
+	}
+
+	now := time.Now()
+	if !untrst.Time.Before(now.Add(clockDrift)) {
+		return fmt.Errorf("new untusted header has a time from the future %v (now: %v)", untrst.Time, now)
```
adlerjohn, line 83: Can you add the maximum block drift to this error message? Otherwise the user will have to dig into the code to figure out what it is.

adlerjohn (commented):
Can you
1. Make a new issue to make this parameter configurable.
2. Link in this PR (or ideally in the code too) the lines in the Tendermint light client that do this.

?

service/header/header_verify.go
renaynay, line 83:
```suggestion
		return fmt.Errorf("new untrusted header has a time from the future %v (now: %v)", untrst.Time, now)
```

service/header/header_verify.go
```diff
@@ -96,6 +64,28 @@ func (eh *ExtendedHeader) VerifyAdjacent(untrst *ExtendedHeader) error {
...
 }
 
+// clockDrift defines how much new header's time can drift into
+// the future relative to the now time during verification.
+var clockDrift = 10 * time.Second
+
+// verify performs basic verification of untrusted header.
+func (eh *ExtendedHeader) verify(untrst *ExtendedHeader) error {
+	if untrst.ChainID != eh.ChainID {
+		return fmt.Errorf("new untusted header has different chain %s, not %s", untrst.ChainID, eh.ChainID)
```
renaynay, line 74:
```suggestion
		return fmt.Errorf("new untrusted header has different chain %s, not %s", untrst.ChainID, eh.ChainID)
```

service/header/header_verify.go
Wondertan: Ok

Wondertan:
The global issue for making global vars configurable now exists #449 
Linked the Tendermint's code to the PR.

renaynay (approved):

Issue #370, opened 2022-01-21
swamp/tests/sync: Light client continues sync after abrupt stop and then start


Bidon15:
Steps:

1. Bridge node is running
2. Wait until block height reached 30
3. Start the light node (sync is starting too) 
4. Stop the light node 
5. Wait until block height reaches 200
6. Start the light node 

Expectation:
Light node should successfully sync till 200 height

Wondertan:
Was this an issue before hardening sync PR or after? If before, play try it on main.

PR #438, opened 2022-02-14
swamp: Add store method to node creation

Bidon15:
- move test from swamp to tests dir
  - adapt makefile to new changes in dir test placement
- rename config to infracomps
- use coreclient with swamp subscription id instead of blockfetcher

We need store to successfully save the state of node in order to start/stop the existing node

Relates to #370

### DIFF
--- a/node/tests/swamp/config.go
+++ b/node/tests/swamp/config.go
@@ -9,19 +9,19 @@ import (
 	"github.com/celestiaorg/celestia-node/core"
 )
 
-// InfraComponents struct represents a set of pre-requisite attributes from the test scenario
-type InfraComponents struct {
+// Components struct represents a set of pre-requisite attributes from the test scenario
+type Components struct {
 	App     types.Application
 	CoreCfg *tn.Config
 }
 
-// DefaultInfraComponents creates a KvStore with a block retention of 200
+// DefaultComponents creates a KvStore with a block retention of 200
 // In addition, the empty block interval is set to 200ms
-func DefaultInfraComponents() *InfraComponents {
+func DefaultComponents() *Components {
 	app := core.CreateKvStore(200)
 	tnCfg := tn.TestConfig()
 	tnCfg.Consensus.CreateEmptyBlocksInterval = 200 * time.Millisecond
-	return &InfraComponents{
+	return &Components{
 		App:     app,
 		CoreCfg: tnCfg,
 	}
--- a/node/tests/swamp/config.go
+++ b/node/tests/swamp/config.go
@@ -9,19 +9,19 @@ import (
 	"github.com/celestiaorg/celestia-node/core"
 )
 
-// Config struct represents a set of pre-requisite attributes from the test scenario
-type Config struct {
+// InfraComps struct represents a set of pre-requisite attributes from the test scenario
+type InfraComps struct {
 	App     types.Application
 	CoreCfg *tn.Config
 }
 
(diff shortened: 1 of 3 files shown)


node/tests/swamp/config.go
```diff
@@ -9,19 +9,19 @@ import (
 	"github.com/celestiaorg/celestia-node/core"
 )
 
-// Config struct represents a set of pre-requisite attributes from the test scenario
-type Config struct {
+// InfraComps struct represents a set of pre-requisite attributes from the test scenario
+type InfraComps struct {
```
liamsi, line 13: What does `Comps` stand for?

node/tests/swamp/swamp.go
```diff
@@ -146,29 +150,44 @@ func (s *Swamp) createPeer(ks keystore.Keystore) host.Host {
...
+// and a mockstore to the NewLightNodeWithStore method
+func (s *Swamp) NewLightNode(options ...node.Option) *node.Node {
+	cfg := node.DefaultConfig(node.Light)
+	store := node.MockStore(s.t, cfg)
+
+	return s.NewLightNodeWithStore(store, options...)
+}
+
+// NewBridgeNodeWithStore creates a new instance of BridgeNodes with predefined Store.
+// Afterwards, the instance is store in the swamp's BridgeNodes slice.
```
renaynay, line 189:
```suggestion
// NewBridgeNodeWithStore creates a new instance of BridgeNodes with predefined Store.
// Afterwards, the instance is stored in the swamp's BridgeNodes slice.
```

node/tests/swamp/config.go
Bidon15: Thanks for the suggestions. Changed to suggested ones

node/tests/swamp/config.go
```diff
@@ -9,19 +9,19 @@ import (
 	"github.com/celestiaorg/celestia-node/core"
 )
 
-// Config struct represents a set of pre-requisite attributes from the test scenario
-type Config struct {
+// InfraComponents struct represents a set of pre-requisite attributes from the test scenario
+type InfraComponents struct {
```
Wondertan, line 13: You can omit Infra tbh. The full name would be then `swamp.Components` which is descriptive enough, but the idea to rename it good.

node/tests/swamp/swamp.go
```diff
@@ -71,14 +76,14 @@ func NewSwamp(t *testing.T, cfg *Config) *Swamp {
...
+func newTendermintCoreNode(ic *InfraComponents) (*tn.Node, error) {
 	var opt rpctest.Options
 	rpctest.RecreateConfig(&opt)
 
-	tn := rpctest.NewTendermint(cfg.App, &opt)
+	tn := rpctest.NewTendermint(ic.App, &opt)
 
 	// rewriting the created config with test's one
-	tn.Config().Consensus = cfg.CoreCfg.Consensus
+	tn.Config().Consensus = ic.CoreCfg.Consensus
```
Wondertan, line 86: Can you change this field with `opt` var directly and after line 81?

Wondertan: Ok.

Wondertan:
Btw, the test failure is something I've never seen before and it seems like we need an issue with this.
cc @renaynay as it related to CoreListener.

Wondertan (approved):
👍🏻

liamsi (approved):

PR #416, opened 2022-02-09
feat(node): new Full Node type

Wondertan:
Just a new Full Node type with Full availability

PR #441, opened 2022-02-15
swamp: Full Node sync tests

### LINKED ISSUE #417: swamp/tests/sync: Include Full Node type
Currently, we have a Bridge/Light node types in the sync related tests
We need tests with Full Node types, too

Reference: https://github.com/celestiaorg/celestia-node/pull/416

Bidon15:
Introduction of Full Node to the swamp tests 

1. Creation of the full node
2. Sync test for full <-> bridge and light ->full -> bridge

Based on #416 #428 #438
Resolves #417

### DIFF
--- a/node/tests/sync_test.go
+++ b/node/tests/sync_test.go
@@ -113,6 +113,16 @@ func TestSyncStartStopLightWithBridge(t *testing.T) {
 	assert.EqualValues(t, h.Commit.BlockID.Hash, sw.GetCoreBlockHashByHeight(ctx, 40))
 }
 
+/*
+Test-Case: Sync a Full Node with a Bridge Node
+Steps:
+1. Create a Bridge Node(BN)
+2. Start a BN
+3. Check BN is synced to height 20
+4. Create a Full Node(FN) with a trusted peer
+5. Start a FN with a defined connection to the BN
+6. Check FN is synced to height 30
+*/
 func TestSyncFullWithBridge(t *testing.T) {
 	sw := swamp.NewSwamp(t, swamp.DefaultComponents())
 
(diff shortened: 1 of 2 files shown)


node/tests/sync_test.go
```diff
@@ -112,3 +112,104 @@ func TestSyncStartStopLightWithBridge(t *testing.T) {
...
+	h, err = full.HeaderServ.GetByHeight(ctx, 30)
+	require.NoError(t, err)
+
+	assert.EqualValues(t, h.Commit.BlockID.Hash, sw.GetCoreBlockHashByHeight(ctx, 30))
+
+	addrs, err = peer.AddrInfoToP2pAddrs(host.InfoFromHost(full.Host))
+	require.NoError(t, err)
+	light := sw.NewLightNode(node.WithTrustedPeer(addrs[0].String()))
+
+	err = sw.Network.UnlinkPeers(bridge.Host.ID(), light.Host.ID())
```
Bidon15, line 205: This test-case is covering the case, where a Light Node has no linkage to Bridge Node and syncs from a Full Node only

Bidon15:
After this PR, the main focus will be pointed to refactoring the swamp code-base #454 and covering possible cases that can de derived or even be fully covered here https://github.com/celestiaorg/test-infra/issues/21

node/tests/sync_test.go
```diff
@@ -112,3 +112,104 @@ func TestSyncStartStopLightWithBridge(t *testing.T) {
...
 }
+
+/*
+Test-Case: Sync a Full Node with a Bridge Node
+Steps:
+1. Create a Bridge Node(BN)
+2. Start a BN
+3. Check BN is synced to height 20
+4. Create a Full Node(FN) with a trusted peer
+5. Start a FN with a defined connection to the BN
```
liamsi, line 123: If I understand the code correctly here, that same FN has the BN as a trusted peer (and hence has a connection to the BN). The comment makes it sound like it is connected to a trusted peer and additionally to a (different) BN.

Bidon15: True. I'll redo the comments as it sound redundant and confusing later 😓

Bidon15: I've found other mentions like that in older tests, so I'll push the changed test-case comment section in a separate PR to not explode this with off-topic commits 🙏

renaynay (approved):
LGTM - love the light node sync against full node test ❤️

liamsi (approved):
👏🏼

PR #442, opened 2022-02-15
ipld: retrieve data starting from root

### LINKED ISSUE #247: Improve RetrieveBlockData

vgonkivs:
Fixes #247

* fetching data from root instead of particular share;
* removed share counter;

Time duration from start fetching till the end of fetching:
```
=== RUN   TestRetrieveBlockData
=== RUN   TestRetrieveBlockData/1x1(min)
374.436µs
=== RUN   TestRetrieveBlockData/32x32(med)
52.936894ms
=== RUN   TestRetrieveBlockData/128x128(max)
1.298028147s
```

Unfortunately, test  TestRetrieveBlockData/128x128(max) failed by timeout. I guess bottleneck could be [here](https://github.com/celestiaorg/rsmt2d/blob/ac0f1e1a51bf7b5420965fb7c35fa32a56e02292/extendeddatacrossword.go#L77). In the worst case we are  iterating over a slice of 65k of elements and making a pretty heavy operation of copying.

### DIFF
--- a/ipld/read.go
+++ b/ipld/read.go
@@ -3,8 +3,6 @@ package ipld
 import (
 	"context"
 	"errors"
-	"fmt"
-	"math/rand"
 
 	"github.com/ipfs/go-cid"
 	ipld "github.com/ipfs/go-ipld-format"
@@ -20,285 +18,130 @@ import (
 
 var ErrRetrieveTimeout = errors.New("retrieve data timeout")
 
+type totalShares struct {
+	data [][]byte
+	err  chan error
+}
+
 // RetrieveData asynchronously fetches block data using the minimum number
 // of requests to IPFS. It fails if one of the random samples sampled is not available.
 func RetrieveData(
-	ctx context.Context,
+	parentCtx context.Context,
 	dah *da.DataAvailabilityHeader,
 	dag ipld.NodeGetter,
 	codec rsmt2d.Codec,
 ) (*rsmt2d.ExtendedDataSquare, error) {
 	edsWidth := len(dah.RowsRoots)
-	sc := newshareCounter(ctx, uint32(edsWidth))
-	// convert the row and col roots into Cids
 	rowRoots := dah.RowsRoots
 	colRoots := dah.ColumnRoots
-	// sample 1/4 of the total extended square by sampling half of the leaves in
-	// half of the rows.
-	// Sampling is done randomly here in an effort to increase
-	// resilience when the downloading node is also serving the shares back to
-	// the network. // TODO @renaynay: this should probably eventually change to
-	// just downloading the first half.
-	for _, row := range uniqueRandNumbers(edsWidth/2, edsWidth) {
-		for _, col := range uniqueRandNumbers(edsWidth/2, edsWidth) {
-			rootCid, err := plugin.CidFromNamespacedSha256(rowRoots[row])
+	shares := &totalShares{
+		data: make([][]byte, edsWidth*edsWidth),
+		err:  make(chan error),
+	}
+
+	ctx, cancel := context.WithCancel(parentCtx)
+	defer cancel()
+
+	go fillBlockWithData(ctx, rowRoots, dag, true, shares)
+	go fillBlockWithData(ctx, colRoots, dag, false, shares)
+
+	for i := 0; i < len(dah.RowsRoots); i++ {
+		select {
+		case err := <-shares.err:
 			if err != nil {
 				return nil, err
 			}
-			go sc.retrieveShare(rootCid, true, row, col, dag)
+		case <-ctx.Done():
+			return nil, ctx.Err()
 		}
 	}
-	// wait until enough data has been collected, too many errors encountered,
-	// or the timeout is reached
-	err := sc.wait()
-	if err != nil {
-		return nil, err
-	}
-	// flatten the square
-	flattened := sc.flatten()
-	// repair the square
+
 	tree := wrapper.NewErasuredNamespacedMerkleTree(uint64(edsWidth) / 2)
-	return rsmt2d.RepairExtendedDataSquare(rowRoots, colRoots, flattened, codec, tree.Constructor)
+	return rsmt2d.RepairExtendedDataSquare(rowRoots, colRoots, shares.data, codec, tree.Constructor)
 }
 
-func RetrieveDataExt(
+// fillBlockWithData fetches 1/4 of shares for the given root
+func fillBlockWithData(
 	ctx context.Context,
-	dah *da.DataAvailabilityHeader,
-	dag ipld.NodeGetter,
-	codec rsmt2d.Codec,
-) (*rsmt2d.ExtendedDataSquare, error) {
-	edsWidth := len(dah.RowsRoots)
-	rowRoots := dah.RowsRoots
-	colRoots := dah.ColumnRoots
+	data [][]byte, dag ipld.NodeGetter,
+	isRow bool,
+	f *totalShares,
+) {
+	for i := 0; i < len(data)/2; i++ {
+		go func(i int) {
+			rootHash, err := plugin.CidFromNamespacedSha256(data[i])
+			if err != nil {
+				f.err <- err
+				return
+			}
+			subtreeRootHash, err := getHalfTreeNodeHash(ctx, rootHash, dag, true)
+			if err != nil {
+				f.err <- err
+				return
+			}
 
-	sc := newshareCounter(ctx, uint32(edsWidth))
-	go fillTheBlock(ctx, colRoots, false, dag, sc)
-	go fillTheBlock(ctx, rowRoots, true, dag, sc)
-	err := sc.wait()
-	if err != nil {
-		fmt.Println(err)
-		return nil, err
-	}
-	// flatten the square
-	flattened := sc.flatten()
-	fmt.Println(len(flattened))
-	tree := wrapper.NewErasuredNamespacedMerkleTree(uint64(edsWidth) / 2)
-	return rsmt2d.RepairExtendedDataSquare(rowRoots, colRoots, flattened, codec, tree.Constructor)
-}
-
-func fillTheBlock(ctx context.Context, data [][]byte, isRow bool, dag ipld.NodeGetter, sc *shareCounter) {
-	numbers := make([]uint32, 0)
-	if isRow {
-		numbers = uniqueRandNumbers(int(sc.edsWidth/4), int(sc.edsWidth))
-	} else {
-		numbers = uniqueNumbersInRange(int(sc.edsWidth/2), int(sc.edsWidth), int(sc.edsWidth/4))
-	}
-	for _, number := range numbers {
-		go func(number uint32) {
-			rootCid, _ := plugin.CidFromNamespacedSha256(data[number])
-			leafs, _ := GetLeafsExt(ctx, dag, rootCid, int(sc.edsWidth/4))
-			for leafIndex, l := range leafs {
-				data := l.RawData()[1:]
+			leafs, err := GetLeafs(ctx, dag, *subtreeRootHash)
+			if err != nil {
+				f.err <- err
+				return
+			}
+			for leafIdx, leaf := range leafs {
+				shareData := leaf.RawData()[1:]
+				// it's not needed to store data for cols
+				// as we are fetching data from the same share for rows and cols
 				if isRow {
-					sc.shareChan <- indexedShare{data: data[NamespaceSize:], index: index{row: uint32(number), col: uint32(leafIndex)}}
-				} else {
-					sc.shareChan <- indexedShare{data: data[NamespaceSize:], index: index{row: uint32(leafIndex), col: uint32(number)}}
+					f.data[(i*len(data))+leafIdx] = shareData[NamespaceSize:]
 				}
 			}
-		}(number)
+			f.err <- nil
+		}(i)
 	}
 }
 
-func GetLeafsExt(
-	ctx context.Context,
-	dag ipld.NodeGetter,
-	root cid.Cid,
-	maxDepth int,
-) (leafs []ipld.Node, err error) {
+// getHalfTreeNodeHash returns only one subtree - left or right
+func getHalfTreeNodeHash(ctx context.Context, root cid.Cid, dag ipld.NodeGetter, isLeftSubtree bool) (*cid.Cid, error) {
 	nd, err := dag.Get(ctx, root)
 	if err != nil {
 		return nil, err
 	}
-	if maxDepth == 0 {
-		leafs = append(leafs, nd)
-		return
-	}
 	links := nd.Links()
-	if len(links) == 1 {
+	if isLeftSubtree || len(links) == 1 {
+		return &links[0].Cid, nil
+	}
+
+	return &links[1].Cid, nil
+}
+
+// GetLeafs recursively starts going down to find all leafs from the given root
+func GetLeafs(ctx context.Context, dag ipld.NodeGetter, root cid.Cid) (leafs []ipld.Node, err error) {
+	// request the node
+	nd, err := dag.Get(ctx, root)
+	if err != nil {
+		return nil, err
+	}
+
+	// look for links
+	lnks := nd.Links()
+	if len(lnks) == 1 {
+		// in case there is only one we reached tree's bottom, so finally request the leaf.
+		nd, err = dag.Get(ctx, lnks[0].Cid)
+		if err != nil {
+			return nil, err
+		}
 		leafs = append(leafs, nd)
 		return
 	}
-	newDep := maxDepth / 2
-	for _, leaf := range links {
-		l, err := GetLeafsExt(ctx, dag, leaf.Cid, newDep)
+
+	for _, node := range lnks {
+		// recursively walk down through selected children to get the leafs
+		l, err := GetLeafs(ctx, dag, node.Cid)
 		if err != nil {
 			return nil, err
 		}
 
 		leafs = append(leafs, l...)
 	}
-	return leafs, nil
-}
-
-func uniqueNumbersInRange(from, to, count int) []uint32 {
-	samples := make(map[uint32]struct{}, count)
-	for i := 0; i < count; {
-		sample := uint32(rand.Intn(to-from) + from)
-		if _, has := samples[sample]; has {
-			continue
-		}
-		samples[sample] = struct{}{}
-		i++
-	}
-	out := make([]uint32, count)
-	counter := 0
-	for s := range samples {
-		out[counter] = s
-		counter++
-	}
-	return out
-}
-
-// uniqueRandNumbers generates count unique random numbers with a max of max
-func uniqueRandNumbers(count, max int) []uint32 {
-	if count > max {
-		panic(fmt.Sprintf("cannot create %d unique samples from a max of %d", count, max))
-	}
-	samples := make(map[uint32]struct{}, count)
-	for i := 0; i < count; {
-		// nolint:gosec // G404: Use of weak random number generator
-		sample := uint32(rand.Intn(max))
-		if _, has := samples[sample]; has {
-			continue
-		}
-		samples[sample] = struct{}{}
-		i++
-	}
-	out := make([]uint32, count)
-	counter := 0
-	for s := range samples {
-		out[counter] = s
-		counter++
-	}
-	return out
-}
-
-type index struct {
-	row uint32
-	col uint32
-}
-
-type indexedShare struct {
-	data []byte
-	index
-}
-
-// shareCounter is a thread safe tallying mechanism for share retrieval
-type shareCounter struct {
-	// all shares
-	shares map[index][]byte
-	// number of shares successfully collected
-	counter uint32
-	// the width of the extended data square
-	edsWidth uint32
-	// the minimum shares needed to repair the extended data square
-	minSharesNeeded uint32
-
-	shareChan chan indexedShare
-	ctx       context.Context
-	cancel    context.CancelFunc
-	// any errors encountered when attempting to retrieve shares
-	errc chan error
-}
-
-func newshareCounter(parentCtx context.Context, edsWidth uint32) *shareCounter {
-	ctx, cancel := context.WithCancel(parentCtx)
-
-	// calculate the min number of shares needed to repair the square
-	minSharesNeeded := edsWidth * edsWidth / 4
-
-	return &shareCounter{
-		shares:          make(map[index][]byte),
-		edsWidth:        edsWidth,
-		minSharesNeeded: minSharesNeeded,
-		shareChan:       make(chan indexedShare, 512),
-		errc:            make(chan error, 1),
-		ctx:             ctx,
-		cancel:          cancel,
-	}
-}
-
-// retrieveLeaf uses GetLeafData to fetch a single leaf and counts that leaf
-func (sc *shareCounter) retrieveShare(
-	rootCid cid.Cid,
-	isRow bool,
-	axisIdx uint32,
-	idx uint32,
-	dag ipld.NodeGetter,
-) {
-	data, err := GetLeafData(sc.ctx, rootCid, idx, sc.edsWidth, dag)
-	if err != nil {
-		select {
-		case <-sc.ctx.Done():
-		case sc.errc <- err:
-		}
-	}
-
-	if len(data) < plugin.ShareSize {
-		return
-	}
-
-	// switch the row and col indexes if needed
-	rowIdx := idx
-	colIdx := axisIdx
-	if isRow {
-		rowIdx = axisIdx
-		colIdx = idx
-	}
-
-	select {
-	case <-sc.ctx.Done():
-	case sc.shareChan <- indexedShare{data: data[NamespaceSize:], index: index{row: rowIdx, col: colIdx}}:
-	}
-}
-
-// wait until enough data has been collected, the timeout has been reached, or
-// too many errors are encountered
-func (sc *shareCounter) wait() error {
-	defer sc.cancel()
-
-	for {
-		select {
-		case <-sc.ctx.Done():
-			err := sc.ctx.Err()
-			if err == context.DeadlineExceeded {
-				return ErrRetrieveTimeout
-			}
-			return err
-		case share := <-sc.shareChan:
-			_, has := sc.shares[share.index]
-			// add iff it does not already exists
-			if !has {
-				sc.shares[share.index] = share.data
-				sc.counter++
-				// check finishing condition
-				if sc.counter >= sc.minSharesNeeded {
-					return nil
-				}
-			}
-
-		case err := <-sc.errc:
-			return fmt.Errorf("failure to retrieve data square: %w", err)
-		}
-	}
-}
-
-func (sc *shareCounter) flatten() [][]byte {
-	flattended := make([][]byte, sc.edsWidth*sc.edsWidth)
-	for index, data := range sc.shares {
-		flattended[(index.row*sc.edsWidth)+index.col] = data
-	}
-	return flattended
+	return
 }
 
 // GetLeafData fetches and returns the data for leaf leafIndex of root rootCid.
--- a/ipld/read.go
+++ b/ipld/read.go
@@ -29,10 +29,13 @@ func RetrieveData(
 	rowRoots := dah.RowsRoots
 	dataSquare := make([][]byte, edsWidth*edsWidth)
 
-	quadrant := pickRandomQuadrant(rowRoots)
+	quadrant, err := pickRandomQuadrant(rowRoots)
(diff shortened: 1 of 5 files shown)


ipld/read.go
```diff
@@ -20,188 +18,130 @@ import (
...
-		cancel:          cancel,
+			leafs, err := GetLeafs(ctx, dag, *subtreeRootHash)
+			if err != nil {
+				f.err <- err
+				return
+			}
+			for leafIdx, leaf := range leafs {
+				shareData := leaf.RawData()[1:]
+				// it's not needed to store data for cols
+				// as we are fetching data from the same share for rows and cols
```
renaynay, line 91: Can you also add a document as to why we still fetch cols even though we're not storing them in the data slice? cc @Wondertan

ipld/read.go
```diff
@@ -20,188 +18,130 @@ import (
...
-func (sc *shareCounter) retrieveShare(
-	rootCid cid.Cid,
-	isRow bool,
-	axisIdx uint32,
-	idx uint32,
-	dag ipld.NodeGetter,
-) {
-	data, err := GetLeafData(sc.ctx, rootCid, idx, sc.edsWidth, dag)
+// getHalfTreeNodeHash returns only one subtree - left or right
+func getHalfTreeNodeHash(ctx context.Context, root cid.Cid, dag ipld.NodeGetter, isLeftSubtree bool) (*cid.Cid, error) {
```
renaynay, line 102: You're always calling this with `isLeftSubtree` as true, so why not just have this method return the left subtree only instead of passing in the bool?

ipld/read.go
```diff
@@ -20,188 +18,130 @@ import (
...
-			if !has {
-				sc.shares[share.index] = share.data
-				sc.counter++
-				// check finishing condition
-				if sc.counter >= sc.minSharesNeeded {
-					return nil
-				}
-			}
+// GetLeafs recursively starts going down to find all leafs from the given root
+func GetLeafs(ctx context.Context, dag ipld.NodeGetter, root cid.Cid) (leafs []ipld.Node, err error) {
```
renaynay, line 116: Please rename to `GetLeaves`

renaynay (commented):
Initial review - left some comments but looks good in general! Thank you for your work on this :)

ipld/read.go
Wondertan, line 102: Cid's are meant to be copied and passed as values, even it is a struct. Not sure why you kept root CID as value and the returning one as a pointer.

Wondertan, line 102: Let's make this func public and call it `GetSubtreeLeaves` so it also returns the actual leaves of subtree by calling `GetLeaves` internally. This will make the code cleaner.

ipld/read.go
```diff
@@ -20,188 +18,130 @@ import (
...
+			}
+			for leafIdx, leaf := range leafs {
+				shareData := leaf.RawData()[1:]
+				// it's not needed to store data for cols
+				// as we are fetching data from the same share for rows and cols
+				if isRow {
+					f.data[(i*len(data))+leafIdx] = shareData[NamespaceSize:]
+				}
+			}
+			f.err <- nil
```
Wondertan, line 96: This could be a func that returns an error instead of writing to chan in each err != nil

ipld/read.go
```diff
@@ -20,188 +18,130 @@ import (
...
+		select {
+		case err := <-shares.err:
 			if err != nil {
 				return nil, err
 			}
-
-			go sc.retrieveShare(rootCid, true, row, col, dag)
+		case <-ctx.Done():
+			return nil, ctx.Err()
 		}
```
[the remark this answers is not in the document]
Wondertan: This will remove the need to make a custom totalShares types. Let's pls avoid spamming types if we don't have to, so we keep things allocated on the stack and not the heap.

Wondertan (changes requested):
Initial review. LGTM. Learning quick

ipld/read.go
```diff
@@ -20,188 +18,130 @@ import (
...
-// too many errors are encountered
-func (sc *shareCounter) wait() error {
-	defer sc.cancel()
-
-	for {
-		select {
-		case <-sc.ctx.Done():
-			err := sc.ctx.Err()
-			if err == context.DeadlineExceeded {
-				return ErrRetrieveTimeout
```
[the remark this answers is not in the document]
vgonkivs: removed

ipld/read.go
vgonkivs: Renamed

ipld/read.go
```diff
@@ -20,188 +18,130 @@ import (
...
+type totalShares struct {
+	data [][]byte
+	err  chan error
+}
+
 // RetrieveData asynchronously fetches block data using the minimum number
 // of requests to IPFS. It fails if one of the random samples sampled is not available.
 func RetrieveData(
-	ctx context.Context,
+	parentCtx context.Context,
```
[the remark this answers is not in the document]
vgonkivs: parentCtx removed

ipld/read.go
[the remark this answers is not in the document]
vgonkivs: function reworked

ipld/read.go
vgonkivs: changed pointer to value in return

vgonkivs: renamed

ipld/read.go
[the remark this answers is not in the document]
vgonkivs: reworked to be able to work with errgroup.

ipld/read.go
vgonkivs: reworked

ipld/read.go
```diff
@@ -29,179 +25,117 @@ func RetrieveData(
...
-	// wait until enough data has been collected, too many errors encountered,
-	// or the timeout is reached
-	err := sc.wait()
-	if err != nil {
+	errGroup, ctx := errgroup.WithContext(ctx)
+	errGroup.Go(func() error {
+		return fillQuarter(ctx, rowRoots, dag, true, dataSquare)
+	})
+	errGroup.Go(func() error {
+		return fillQuarter(ctx, rowRoots, dag, false, dataSquare)
```
liamsi, line 37: Are these supposed to be row roots in both cases?

ipld/read.go
```diff
@@ -29,179 +25,117 @@ func RetrieveData(
...
-	errc chan error
+	return rsmt2d.RepairExtendedDataSquare(rowRoots, colRoots, dataSquare, codec, tree.Constructor)
 }
 
-func newshareCounter(parentCtx context.Context, edsWidth uint32) *shareCounter {
-	ctx, cancel := context.WithCancel(parentCtx)
+// fillQuarter fetches 1/4 of shares for the given root
+func fillQuarter(
+	ctx context.Context,
+	data [][]byte, dag ipld.NodeGetter,
```
liamsi, line 50: Shouldn't this be called `roots` instead of just `data`.
Also why has ever param its own line while dad and data share one?

ipld/read.go
vgonkivs: Ah, yeah. Nice catch. Thank you. Changed to cols for the second case.

ipld/read.go
vgonkivs: Agree. Renamed to roots.

ipld/read.go
renaynay: Can we simplify it for this PR at least to only have this function do `getLeftHalfTreeNodeHash` ? It's cleaner if this is the only usage *for now* IMO.

renaynay (approved):
LGTM now that @Wondertan 's comments have been addressed.

renaynay:
Needs a rebase though

renaynay:
Unfortunately needs another rebase :(

vgonkivs:
Rebased

service/share/full_availability.go
```diff
@@ -13,19 +13,19 @@ import (
 // recovery technique. It is considered "full" because it is required
 // to download enough shares to fully reconstruct the data square.
 type fullAvailability struct {
-	getter format.NodeGetter
+	service format.DAGService
```
liamsi, line 16: Just a question: what triggered changing this to point to a DAGService instead of a NodeGetter? From reading the changes it seems that everywhere only the the NodeGetter part is used. Or am I missing sth?

liamsi (approved):
👏🏼

CHANGELOG-PENDING.md
```diff
@@ -7,7 +7,7 @@
 ### FEATURES
 
 ### IMPROVEMENTS
-
+- [ipld: retreive data starting from root #442](https://github.com/celestiaorg/celestia-node/pull/442) [@vgonkivs](https://github.com/vgonkivs)
```
adlerjohn, line 10:
```suggestion

- [ipld: retrieve data starting from root #442](https://github.com/celestiaorg/celestia-node/pull/442) [@vgonkivs](https://github.com/vgonkivs)

```

ipld/read.go
```diff
@@ -18,190 +16,131 @@ import (
...
-			// add iff it does not already exists
-			if !has {
-				sc.shares[share.index] = share.data
-				sc.counter++
-				// check finishing condition
-				if sc.counter >= sc.minSharesNeeded {
-					return nil
-				}
-			}
+// getLeaves recursively starts going down to find all leafs from the given root
```
adlerjohn, line 116: What is the difference between `GetLeaves` and `getLeaves`? Looks like the latter is the implementation? They should probably have different names.

service/share/full_availability.go
vgonkivs: Because now, we are committing shares to ipfs after successful repairing. And it requires ipld.NodeAdder.

ipld/read.go
vgonkivs: `GetLeaves` is a wrapper over `getLeaves` but I think they both have correct namings. I've decided to move more complex logic to a separate function.

renaynay (approved):

ipld/read.go
```diff
@@ -18,190 +16,131 @@ import (
...
-	errc chan error
-}
-
-func newshareCounter(parentCtx context.Context, edsWidth uint32) *shareCounter {
-	ctx, cancel := context.WithCancel(parentCtx)
+// fillQuarter fetches 1/4 of shares for the given root
+func fillQuarter(
+	ctx context.Context,
+	roots [][]byte, dag ipld.NodeGetter,
+	isRow bool,
```
adlerjohn, line 56: What's the purpose of `isRow`? It looks like you only ever set it to `true`...

ipld/read.go
```diff
@@ -18,190 +16,131 @@ import (
...
-	ctx       context.Context
-	cancel    context.CancelFunc
-	// any errors encountered when attempting to retrieve shares
-	errc chan error
-}
-
-func newshareCounter(parentCtx context.Context, edsWidth uint32) *shareCounter {
-	ctx, cancel := context.WithCancel(parentCtx)
+// fillQuarter fetches 1/4 of shares for the given root
+func fillQuarter(
```
adlerjohn, line 53: Nit, but I would prefer `Quadrant` instead of `Quarter`

Wondertan: Agree with John

Wondertan:
@vgonkivs, will review after quadrant randomization is implemented.

ipld/read.go
```diff
@@ -18,190 +18,162 @@ import (
...
+	return extended, batchAdder.Commit()
 }
 
-type index struct {
-	row uint32
-	col uint32
+type quadrant struct {
+	isLeftSubtree bool
+	from          int
+	chunk         [][]byte //roots
```
Wondertan, line 53: Rename to Root. Also, I suggest having roots be always as CID's so you convert them in the beginning. This way it's also clear what roots are and not just arbitrary bytes.

ipld/read.go
```diff
@@ -18,190 +18,162 @@ import (
...
-	// number of shares successfully collected
-	counter uint32
-	// the width of the extended data square
-	edsWidth uint32
-	// the minimum shares needed to repair the extended data square
-	minSharesNeeded uint32
+	// define range of shares for sampling
+	if q > 2 {
+		quadrant.from = edsWidth / 2
+		copy(quadrant.chunk, roots[edsWidth/2:])
```
Wondertan: You could also point to the full root slice in the code where you use roots, instead of copying the half.

ipld/read.go
```diff
@@ -18,190 +18,162 @@ import (
...
+				return err
+			}
+			subtreeRootHash, err := GetSubtreeLeaves(ctx, rootHash, dag, roots.isLeftSubtree)
+			if err != nil {
+				return err
+			}
 
-	// calculate the min number of shares needed to repair the square
-	minSharesNeeded := edsWidth * edsWidth / 4
+			leaves, err := GetLeaves(ctx, dag, subtreeRootHash, uint32(len(roots.chunk)))
```
Wondertan, line 101: Should be the size of roots divided by two

ipld/read.go
Wondertan: As per multiple discussionss, no need to fetch cols, but only as retry if fails for rows

ipld/read.go
```diff
@@ -18,190 +18,162 @@ import (
...
-func (sc *shareCounter) retrieveShare(
-	rootCid cid.Cid,
-	isRow bool,
-	axisIdx uint32,
-	idx uint32,
-	dag ipld.NodeGetter,
-) {
-	data, err := GetLeafData(sc.ctx, rootCid, idx, sc.edsWidth, dag)
+// GetSubtreeLeaves returns only one subtree - left or right
+func GetSubtreeLeaves(ctx context.Context, root cid.Cid, dag ipld.NodeGetter, isLeftSubtree bool) (cid.Cid, error) {
```
Wondertan, line 130: Can you make this func to return the actual subtree by calling GetLeaves internally?

ipld/read.go
```diff
@@ -18,190 +17,166 @@ import (
...
-}
+// fillQuadrant fetches 1/4 of shares for the given root
+func fillQuadrant(
+	ctx context.Context,
+	roots *quadrant, dag ipld.NodeGetter,
+	isRow bool,
+	dataSquare [][]byte,
+) error {
+	errGroup, ctx := errgroup.WithContext(ctx)
+	fetcher := func(i int) {
```
Wondertan, line 90: You can avoid defining the func by doing `i := i`

ipld/read.go
```diff
@@ -18,190 +17,166 @@ import (
...
-	ctx       context.Context
-	cancel    context.CancelFunc
-	// any errors encountered when attempting to retrieve shares
-	errc chan error
-}
+// fillQuadrant fetches 1/4 of shares for the given root
+func fillQuadrant(
+	ctx context.Context,
+	roots *quadrant, dag ipld.NodeGetter,
+	isRow bool,
```
Wondertan, line 86: Delete isRow

ipld/read.go
vgonkivs: But then we will add it again in scope of implementing retry mechanism from #508

ipld/read.go
```diff
@@ -18,190 +17,166 @@ import (
...
+				// shares from the right subtree should be
+				// inserted after all shares from the left subtree
+				if !roots.isLeftSubtree {
+					leafIdx += len(leaves)
+				}
+				shareData := leaf.RawData()[1:]
+				// it's not needed to store data for cols
+				// as we are fetching data from the same share for rows and cols
+				if isRow {
+					dataSquare[((i+roots.from)*len(roots.chunk)*2)+leafIdx] = shareData[NamespaceSize:]
```
Wondertan, line 116: `len(roots.chunk)` does not change so can be called once in the func

ipld/read.go
Wondertan: But that is an optional nit

ipld/read.go
vgonkivs: Renamed

ipld/read.go
vgonkivs: removed

ipld/read.go
vgonkivs: renamed to `rootCids`

ipld/read.go
vgonkivs: changed type to `cid.Cid`. So, there is no needs of copying anymore.

ipld/read.go
vgonkivs: done

ipld/read.go
vgonkivs: moved to `GetSubtreeLeaves`

ipld/read.go
vgonkivs: done

ipld/read.go
vgonkivs: removed

ipld/read.go
vgonkivs: done

ipld/read.go
```diff
@@ -18,190 +17,170 @@ import (
...
+	// quadrants 1 and 3 corresponds to left subtree,
+	// 2 and 4 to the right subtree
+	/* | 1 | 2 |
+	   | 3 | 4 |
+	*/
+	// choose subtree
+	if q%2 == 1 {
+		quadrant.isLeftSubtree = true
+	}
+	to := 0
```
renaynay, line 73: I don't understand why `from` is a field inside of `quadrant` but `to` is not, and vice versa. Is there a reason?

ipld/read.go
```diff
@@ -18,190 +17,170 @@ import (
...
-	// any errors encountered when attempting to retrieve shares
-	errc chan error
+	var err error
+	for index, counter := quadrant.from, 0; index < to; index++ {
+		quadrant.rootCids[counter], err = plugin.CidFromNamespacedSha256(roots[index])
+		if err != nil {
+			return nil, err
+		}
+		counter++
+	}
```
renaynay, line 89: if both `counter` and `index` are incrementing +1, why not just use `index`?

ipld/read.go
```diff
@@ -18,190 +17,170 @@ import (
...
-	axisIdx uint32,
-	idx uint32,
+// GetSubtreeLeaves returns only one subtree - left or right
+func GetSubtreeLeaves(
+	ctx context.Context,
+	root cid.Cid,
 	dag ipld.NodeGetter,
-) {
-	data, err := GetLeafData(sc.ctx, rootCid, idx, sc.edsWidth, dag)
+	isLeftSubtree bool, treeSize uint32) ([]ipld.Node, error) {
```
renaynay, line 129:
```suggestion
	isLeftSubtree bool, 
	treeSize uint32,
	) ([]ipld.Node, error) {
```

ipld/read.go
```diff
@@ -18,190 +17,170 @@ import (
...
+}
 
-	select {
-	case <-sc.ctx.Done():
-	case sc.shareChan <- indexedShare{data: data[NamespaceSize:], index: index{row: rowIdx, col: colIdx}}:
+func GetLeaves(ctx context.Context, dag ipld.NodeGetter, root cid.Cid, size uint32) ([]ipld.Node, error) {
+	leaves, err := getLeaves(ctx, dag, root, make([]ipld.Node, 0, size))
+	if err != nil {
+		return nil, err
 	}
```
renaynay, line 151: I think you can just return `getLeaves`, etc.

ipld/read.go
vgonkivs: because `counter` and `index` could not be equal.

renaynay: Oh sorry, I thought I saw 
```
for index, counter := quadrant.from
```

but it's actually

```
for index, counter := quadrant.from, 0
```

Wondertan (approved):
After @renaynay comments are applied this is fine to me. `pickRandomQuadrant` still feels a bit ugly to me, but I cant propose anything better atm and this PR is better to be merged so :shipit:

ipld/read.go
vgonkivs: We need `from` field in fillQuadrant to correctly set index of share. But `to`  var is used to define the last root that we need to store from RowRoots, since the quadrant can have two different ranges:
* [0;edsWidth/2)
* [edsWidth/2;edsWidth)

renaynay: I still don't understand why `to` can't also be a field set within `quadrant` ? it's just a bit hard to follow the code here.

ipld/read.go
```diff
@@ -18,190 +17,170 @@ import (
...
-	ctx       context.Context
-	cancel    context.CancelFunc
-	// any errors encountered when attempting to retrieve shares
-	errc chan error
+	var err error
+	for index, counter := quadrant.from, 0; index < to; index++ {
+		quadrant.rootCids[counter], err = plugin.CidFromNamespacedSha256(roots[index])
+		if err != nil {
+			return nil, err
+		}
```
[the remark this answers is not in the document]
vgonkivs: In the beginning of the RetrieveData we don't know what quadrant will we choose and thus generate cids for each root in row/col. Do we need this?

ipld/read.go
vgonkivs: extended `quadrant` struct with `to` field

renaynay: But neither does `quadrant.from` right?

ipld/read.go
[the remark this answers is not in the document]
renaynay: I'd recommend making an issue out of this discussion, merging as is and then coming back to it later, if it's important.

ipld/read.go
```diff
@@ -18,190 +17,165 @@ import (
...
+}
 
-	// calculate the min number of shares needed to repair the square
-	minSharesNeeded := edsWidth * edsWidth / 4
+// fillQuadrant fetches 1/4 of shares for the given root
+func fillQuadrant(
+	ctx context.Context,
+	roots *quadrant, dag ipld.NodeGetter,
+	dataSquare [][]byte,
+) error {
```
renaynay, line 100:
```suggestion
	roots *quadrant, 
	dag ipld.NodeGetter,
	dataSquare [][]byte,
) error {
```

ipld/read.go
```diff
@@ -18,190 +17,165 @@ import (
...
+	}
+	return quadrant, nil
+}
 
-	// calculate the min number of shares needed to repair the square
-	minSharesNeeded := edsWidth * edsWidth / 4
+// fillQuadrant fetches 1/4 of shares for the given root
+func fillQuadrant(
+	ctx context.Context,
+	roots *quadrant, dag ipld.NodeGetter,
```
renaynay, line 98:
```suggestion
	quadrant *quadrant, dag ipld.NodeGetter,
```

renaynay: let's just continue using this naming convention everywhere.

ipld/read.go
renaynay: (pls also refactor uses of `roots` -> `quadrant` in this function

ipld/read.go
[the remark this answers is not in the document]
vgonkivs: Ok, I will create an issue.

ipld/read.go
```diff
@@ -18,190 +17,165 @@ import (
...
+) error {
+	errGroup, ctx := errgroup.WithContext(ctx)
+	for i := 0; i < len(roots.rootCids); i++ {
+		i := i
+		errGroup.Go(func() error {
+			leaves, err := GetSubtreeLeaves(ctx, roots.rootCids[i], dag, roots.isLeftSubtree, uint32(len(roots.rootCids)/2))
+			if err != nil {
+				return err
+			}
+			length := len(roots.rootCids)
```
renaynay, line 109: could this be called `quadrantWidth` ?

renaynay: Since this is the "length" of all rootCIDs in the quadrant correct?

ipld/read.go
```diff
@@ -18,190 +17,165 @@ import (
...
+			for leafIdx, leaf := range leaves {
+				// shares from the right subtree should be
+				// inserted after all shares from the left subtree
+				if !roots.isLeftSubtree {
+					leafIdx += len(leaves)
+				}
+				shareData := leaf.RawData()[1:]
+				dataSquare[((i+roots.from)*length*2)+leafIdx] = shareData[NamespaceSize:]
```
renaynay, line 117: Can you just document here how you calculate the position/index of where to add the share data inside the `dataSquare` ? Something like: 

```suggestion
                                 // position is calculated by < insert explanation>
                                 position := ((i+roots.from)*length*2)+leafIdx
				dataSquare[position] = shareData[NamespaceSize:]
```

renaynay (commented):
small comments to improve readability IMO.

ipld/read.go
vgonkivs: Yes

ipld/read.go
vgonkivs: It is  used to calculate `position` in `dataSquare`

ipld/read.go
vgonkivs: changed naming

ipld/read.go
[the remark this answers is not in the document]
vgonkivs: https://github.com/celestiaorg/celestia-node/issues/516

renaynay (approved):
all good on my side, waiting on @Wondertan

ipld/read.go
```diff
@@ -18,190 +17,177 @@ import (
...
-	case sc.shareChan <- indexedShare{data: data[NamespaceSize:], index: index{row: rowIdx, col: colIdx}}:
-	}
+	return getLeaves(ctx, dag, subtreeRootHash, make([]ipld.Node, 0, treeSize))
 }
 
-// wait until enough data has been collected, the timeout has been reached, or
-// too many errors are encountered
-func (sc *shareCounter) wait() error {
-	defer sc.cancel()
+func GetLeaves(ctx context.Context, dag ipld.NodeGetter, root cid.Cid, size uint32) ([]ipld.Node, error) {
```
liamsi, line 159: Can this exported function be documented? Also can it have a unit test as well?

liamsi (commented):
I think we should add tests and docs for the new exported functions. Otherwise looks good! Will take a closer look again after tests were added.

ipld/read.go
vgonkivs: removed

ipld/read.go
```diff
@@ -18,190 +17,177 @@ import (
...
 }
 
-// retrieveLeaf uses GetLeafData to fetch a single leaf and counts that leaf
-func (sc *shareCounter) retrieveShare(
-	rootCid cid.Cid,
-	isRow bool,
-	axisIdx uint32,
-	idx uint32,
+// GetSubtreeLeaves returns only one subtree - left or right
+func GetSubtreeLeaves(
```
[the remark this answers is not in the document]
vgonkivs: made as unexported

liamsi (approved):
👏🏼 🙏🏼

PR #443, opened 2022-02-15
swamp: Sync sub test with state sync

### LINKED ISSUE #351: service/header: Sync test needs to know at which specific height should the node be synced
Currently, in our sync test nodes are just waiting to be fully synced, without explicitly checking the height of which we consider this to be correct. This is causing the test to be flaky in CI 
Example of failed : [Run #2](https://github.com/celestiaorg/celestia-node/runs/4809450105?check_suite_focus=true)
Example of passed: [Run #3](https://github.com/celestiaorg/celestia-node/runs/4809482324?check_suite_focus=true)

In both scenarios we are trying to catch up with the very fast block production from the core.Client which is not good. 
To resolve this, we need 2 points to be done:
1. Make block production slower (e.g. locally we can produce around 30-40 blocks and in CI this can be close to 200)
2. Have an API(sub or smt like requestByHeight) from node which can notify the synced height

### LINKED ISSUE #370: swamp/tests/sync: Light client continues sync after abrupt stop and then start

Bidon15:
Depends on #397, #428
Resolves #351 #370

### DIFF
--- a/node/tests/swamp/swamp.go
+++ b/node/tests/swamp/swamp.go
@@ -13,6 +13,7 @@ import (
 	ma "github.com/multiformats/go-multiaddr"
 	"github.com/stretchr/testify/require"
 	tn "github.com/tendermint/tendermint/node"
+	coretypes "github.com/tendermint/tendermint/rpc/core/types"
 	rpctest "github.com/tendermint/tendermint/rpc/test"
 	"github.com/tendermint/tendermint/types"
 
@@ -96,6 +97,13 @@ func (s *Swamp) stopAllNodes(ctx context.Context, allNodes ...[]*node.Node) {
 			require.NoError(s.t, node.Stop(ctx))
 		}
 	}
+}
+
+// GetCoreBlockByHeight returns a tendermint block by height
+func (s *Swamp) GetCoreBlockByHeight(ctx context.Context, height *int64) *coretypes.ResultBlock {
+	b, err := s.CoreClient.Block(ctx, height)
+	require.NoError(s.t, err)
+	return b
 }
 
 // WaitTillHeight holds the test execution until the given amount of blocks
@@ -226,4 +234,38 @@ func (s *Swamp) NewLightNodeWithStore(store node.Store, options ...node.Option) 
 	s.LightNodes = append(s.LightNodes, node)
 
 	return node
+}
+
+// RemoveNode removes a node from the swamp's node slice
+// this allows reusage of the same var in the test scenario
+// if the user needs to stop and start the same node
+func (s *Swamp) RemoveNode(n *node.Node, t node.Type) error {
+	switch t {
+	case node.Light:
+		return s.remove(n, s.LightNodes)
+	case node.Bridge:
+		return s.remove(n, s.BridgeNodes)
+	default:
+		return fmt.Errorf("no such type or node")
+	}
+}
+
+func (s *Swamp) remove(rn *node.Node, sn []*node.Node) error {
+	if len(sn) == 1 {
+		sn = nil
+		return nil
+	}
+
+	initSize := len(sn)
+	for i := 0; i < len(sn); i++ {
+		if sn[i] == rn {
+			sn = append(sn[:i], sn[i+1:]...)
+			i--
+		}
+	}
+
+	if initSize <= len(sn) {
+		return fmt.Errorf("cannot delete the node")
+	}
+	return nil
 }
--- a/node/tests/swamp/swamp.go
+++ b/node/tests/swamp/swamp.go
@@ -144,6 +144,7 @@ func (s *Swamp) createPeer(ks keystore.Keystore) host.Host {
 	host, err := s.Network.AddPeer(key, a)
 	require.NoError(s.t, err)
 
+	require.NoError(s.t, s.Network.LinkAll())
 	return host
 }
 
--- a/node/tests/swamp/swamp.go
+++ b/node/tests/swamp/swamp.go
@@ -12,8 +12,8 @@ import (
 	mocknet "github.com/libp2p/go-libp2p/p2p/net/mock"
 	ma "github.com/multiformats/go-multiaddr"
 	"github.com/stretchr/testify/require"
+	"github.com/tendermint/tendermint/libs/bytes"
 	tn "github.com/tendermint/tendermint/node"
-	coretypes "github.com/tendermint/tendermint/rpc/core/types"
 	rpctest "github.com/tendermint/tendermint/rpc/test"
 	"github.com/tendermint/tendermint/types"
 
@@ -99,11 +99,11 @@ func (s *Swamp) stopAllNodes(ctx context.Context, allNodes ...[]*node.Node) {
 	}
 }
 
-// GetCoreBlockByHeight returns a tendermint block by height
-func (s *Swamp) GetCoreBlockByHeight(ctx context.Context, height *int64) *coretypes.ResultBlock {
-	b, err := s.CoreClient.Block(ctx, height)
+// GetCoreBlockHashByHeight returns a tendermint block's hash by provided height
+func (s *Swamp) GetCoreBlockHashByHeight(ctx context.Context, height int64) bytes.HexBytes {
+	b, err := s.CoreClient.Block(ctx, &height)
 	require.NoError(s.t, err)
-	return b
+	return b.BlockID.Hash
 }
 
 // WaitTillHeight holds the test execution until the given amount of blocks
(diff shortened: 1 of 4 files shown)


Wondertan:
@Bidon15, You can now rebase it on main and undraft it

liamsi (commented):
> Depends on https://github.com/celestiaorg/celestia-node/pull/397

I don't really see the relation to https://github.com/celestiaorg/celestia-node/pull/397
(is one of the SyncState methods used internally by `GetByHeight`? Doesn't look like it)

node/tests/sync_test.go
```diff
@@ -37,17 +37,23 @@ func TestSyncLightWithBridge(t *testing.T) {
 
 	err := bridge.Start(ctx)
 	require.NoError(t, err)
-	time.Sleep(1 * time.Second)
-	assert.False(t, bridge.HeaderServ.IsSyncing())
+	h, err := bridge.HeaderServ.GetByHeight(ctx, 20)
+	require.NoError(t, err)
+	fmt.Println(h.Commit.Hash().String())
```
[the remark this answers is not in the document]
Bidon15: Oops. Removing

Bidon15:
> > Depends on #397
> 
> I don't really see the relation to #397 (is one of the SyncState methods used internally by `GetByHeight`? Doesn't look like it)

Yes. You are totally right about the `GetByHeight`, which is related to 397, but was migrated to a standalone #428 
I'll adapt the description as well as add the test that is covering #370 that I've removed initially

liamsi (approved):
👍🏼

node/tests/sync_test.go
```diff
@@ -37,17 +36,79 @@ func TestSyncLightWithBridge(t *testing.T) {
...
+	state := bridge.CoreClient.IsRunning()
+	require.True(t, state)
+	err := bridge.Start(ctx)
+	require.NoError(t, err)
+
+	h, err := bridge.HeaderServ.GetByHeight(ctx, 20)
+	require.NoError(t, err)
+
+	var ch int64 = 20
+	require.EqualValues(t, h.Commit.BlockID.Hash, sw.GetCoreBlockByHeight(ctx, &ch).BlockID.Hash)
```
Wondertan, line 90: Pls, hide this int ugliness inside the `GetCoreBlockByHeight` method. You can pass here an int and make a pointer to it inside the method. Pls fix this everywhere.

node/tests/sync_test.go
```diff
@@ -37,17 +36,79 @@ func TestSyncLightWithBridge(t *testing.T) {
...
+
+	addrs, err := peer.AddrInfoToP2pAddrs(host.InfoFromHost(sw.Network.Host(bridge.Host.ID())))
```
Wondertan, line 92: You can simply pass `bridge.Host` to `InfoFromHost` without `sw.Network.Host`. Pls remove it everywhere.

node/tests/sync_test.go
```diff
@@ -37,17 +36,79 @@ func TestSyncLightWithBridge(t *testing.T) {
 
 	err := bridge.Start(ctx)
 	require.NoError(t, err)
-	time.Sleep(1 * time.Second)
-	assert.False(t, bridge.HeaderServ.IsSyncing())
+	h, err := bridge.HeaderServ.GetByHeight(ctx, 20)
+	require.NoError(t, err)
```
Wondertan, line 40: Pls make the context here is wrapped with a timeout. This is important to follow in every test

node/tests/swamp/swamp.go
```diff
@@ -226,3 +235,37 @@ func (s *Swamp) NewLightNodeWithStore(store node.Store, options ...node.Option)
...
+func (s *Swamp) remove(rn *node.Node, sn []*node.Node) error {
+	if len(sn) == 1 {
+		sn = nil
+		return nil
+	}
+
+	initSize := len(sn)
+	for i := 0; i < len(sn); i++ {
+		if sn[i] == rn {
+			sn = append(sn[:i], sn[i+1:]...)
```
Wondertan, line 262: Here we have a bug. Append in this case will create a new slice with a newly allocated memory behind it and you only set it to `sn` in the local stack, instead, you should set the result here directly into Swamp.LightNode or Swamp.BridgeNodes. 

Ping @vgonkivs for a more in-depth explanation.

Wondertan (changes requested):

node/tests/swamp/swamp.go
Wondertan: @Bidon15 https://github.com/golang/go/wiki/SliceTricks

node/tests/sync_test.go
Bidon15: Done

node/tests/sync_test.go
Bidon15: Agreed. Changed the func naming to Hash, too

node/tests/sync_test.go
Bidon15: I've made the ctxTimeout for GetByHeight only.
wdyt?

node/tests/swamp/swamp.go
Bidon15: done according to ideas above. Thanks!

service/header/store_heightsub.go
```diff
@@ -40,7 +40,7 @@ func (hs *heightSub) SetHeight(height uint64) {
 // It can return errElapsedHeight, which means a requested header was already provided
 // and caller should get it elsewhere.
 func (hs *heightSub) Sub(ctx context.Context, height uint64) (*ExtendedHeader, error) {
-	if hs.Height() >= height || hs.Height() == 0 { // if height is zero - intentionally avoid subscribing
```
Bidon15, line 43: I don;t know how but this if statement is not letting the light node sub to headers in the specific test 

cc: @Wondertan

service/header/service.go
```diff
@@ -45,7 +45,11 @@ func (s *Service) Stop(context.Context) error {
 	return nil
 }
 
-// TODO(@Bidon15): Issue #351
+// GetByHeight returns the ExtendedHeader by a provided height
+// This is a blocking call until the height is reached or context deadline is exceeded
```
renaynay, line 49:
```suggestion
// GetByHeight returns the ExtendedHeader at the given height, blocking
// until header has been processed by the store or context deadline is exceeded.
```

service/header/service.go
```diff
@@ -45,7 +45,11 @@ func (s *Service) Stop(context.Context) error {
...
+func (s *Service) GetByHeight(ctx context.Context, height uint64) (*ExtendedHeader, error) {
```
renaynay, line 50: How would this fit into your idea for RPC on header service @Wondertan ?

renaynay (changes requested):

node/tests/swamp/swamp.go
```diff
@@ -226,3 +235,39 @@ func (s *Swamp) NewLightNodeWithStore(store node.Store, options ...node.Option)
 
 	return node
 }
+
+// RemoveNode removes a node from the swamp's node slice
+// this allows reusage of the same var in the test scenario
+// if the user needs to stop and start the same node
+func (s *Swamp) RemoveNode(n *node.Node, t node.Type) error {
```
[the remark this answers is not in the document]
Bidon15: No. It is shut down before being removed from the swamp slices in the test scenario

service/header/service.go
Wondertan: This fits well, we made this together in some pair coding session, AFAIR.

node/tests/sync_test.go
Wondertan: I am ok with this as long as there is a global context with a timeout. The global context is needed so we don't ever deadlock while running tests locally or on CI.

Wondertan: The blocking may happen literally everywhere where context is passed and not only for GetByHeight, so the timeout should be on the root context. Further, it won't really make sense to have a special context timeout for GetByHeight as there will be one in the root.

Wondertan (approved):
Approving, considering that the last context timeout issue will be fixed.

node/tests/sync_test.go
Bidon15: Ok, I've changed to global ctx with timeout, so resolving and merging 
thanks for input 🙏

renaynay (approved):

PR #455, opened 2022-02-17
Pass trusted peers to config

### LINKED ISSUE #364: node: Light node can be initialised with multiple trusted peers
### Summary

Light node needs more options to include in trusted peers instead of 1

### Impact on users

In current implementation, a light node can be only initialised with 1 trusted peer. It will be beneficial to allow multiple trusted peers (array of them) for light node initialisation

vgonkivs:
Fixes #364 

Chages:
* multiple trusted peers can be passed on node initialisation;
* trusted peer will be taken randomly on `performRequest`;

### DIFF
--- a/cmd/flags_header.go
+++ b/cmd/flags_header.go
@@ -12,8 +12,8 @@ import (
 )
 
 var (
-	headersTrustedHashFlag = "headers.trusted-hash"
-	headersTrustedPeerFlag = "headers.trusted-peer"
+	headersTrustedHashFlag  = "headers.trusted-hash"
+	headersTrustedPeersFlag = "headers.trusted-peers"
 )
 
 // HeadersFlags gives a set of hardcoded Header package flags.
@@ -27,7 +27,7 @@ func HeadersFlags() *flag.FlagSet {
 	)
 
 	flags.StringSlice(
-		headersTrustedPeerFlag,
+		headersTrustedPeersFlag,
 		make([]string, 0),
 		"Multiaddr of a reliable peer to fetch headers from. (Format: multiformats.io/multiaddr)",
 	)
@@ -47,7 +47,7 @@ func ParseHeadersFlags(cmd *cobra.Command, env *Env) error {
 		env.AddOptions(node.WithTrustedHash(hash))
 	}
 
-	tpeers, err := cmd.Flags().GetStringSlice(headersTrustedPeerFlag)
+	tpeers, err := cmd.Flags().GetStringSlice(headersTrustedPeersFlag)
 	if err != nil {
 		return err
 	}
@@ -55,7 +55,7 @@ func ParseHeadersFlags(cmd *cobra.Command, env *Env) error {
 		for _, tpeer := range tpeers {
 			_, err := multiaddr.NewMultiaddr(tpeer)
 			if err != nil {
-				return fmt.Errorf("cmd: while parsing '%s' with peer addr '%s': %w", headersTrustedPeerFlag, tpeer, err)
+				return fmt.Errorf("cmd: while parsing '%s' with peer addr '%s': %w", headersTrustedPeersFlag, tpeer, err)
 			}
 			env.AddOptions(node.WithTrustedPeer(tpeer))
 		}
--- a/cmd/flags_header.go
+++ b/cmd/flags_header.go
@@ -26,9 +26,9 @@ func HeadersFlags() *flag.FlagSet {
 		"Hex encoded header hash. Used to subjectively initialize header synchronization",
 	)
 
-	flags.String(
+	flags.StringSlice(
 		headersTrustedPeerFlag,
-		"",
+		make([]string, 0),
 		"Multiaddr of a reliable peer to fetch headers from. (Format: multiformats.io/multiaddr)",
 	)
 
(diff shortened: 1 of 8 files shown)


cmd/flags_header.go
```diff
@@ -12,8 +12,8 @@ import (
 )
 
 var (
-	headersTrustedHashFlag = "headers.trusted-hash"
-	headersTrustedPeerFlag = "headers.trusted-peer"
+	headersTrustedHashFlag  = "headers.trusted-hash"
+	headersTrustedPeersFlag = "headers.trusted-peers"
```
Wondertan, line 16: This is a breaking change so we should add the proper entry to changelog

node/services/config.go
```diff
@@ -17,32 +17,40 @@ type Config struct {
 	// TrustedHash is the Block/Header hash that Nodes use as starting point for header synchronization.
 	// Only affects the node once on initial sync.
 	TrustedHash string
-	// TrustedPeer is the peer we trust to fetch headers from.
+	// TrustedPeers is the peer we trust to fetch headers from.
```
Wondertan, line 20: Should also change the comment accordingly to plural form

service/header/p2p_exchange.go
```diff
@@ -24,13 +25,13 @@ type P2PExchange struct {
 
 	// TODO @renaynay: post-Devnet, we need to remove reliance of Exchange on one bootstrap peer
 	// Ref https://github.com/celestiaorg/celestia-node/issues/172#issuecomment-964306823.
-	trustedPeer peer.ID
+	trustedPeers peer.IDSlice
```
Wondertan, line 28: You can remove the TODO above.

service/header/p2p_exchange.go
```diff
@@ -98,11 +99,10 @@ func (ex *P2PExchange) performRequest(ctx context.Context, req *pb.ExtendedHeade
 	if req.Amount == 0 {
 		return make([]*ExtendedHeader, 0), nil
 	}
-	if ex.trustedPeer == "" {
-		return nil, fmt.Errorf("no trusted peer")
-	}
```
Wondertan, line 103: `if len(ex.trustedPeer) == 0` ?

service/header/p2p_exchange_test.go
```diff
@@ -109,17 +110,17 @@ func createMocknet(ctx context.Context, t *testing.T) (libhost.Host, libhost.Hos
...
 	err := serverSideEx.Start(context.Background())
 	require.NoError(t, err)
 
 	t.Cleanup(func() {
 		serverSideEx.Stop(context.Background()) //nolint:errcheck
 	})
-
-	return NewP2PExchange(host, peer.ID()), store
+	ids := []peer.ID{tpeer.ID()}
+	return NewP2PExchange(host, ids), store
```
Wondertan, line 123: Inline those?

Wondertan (commented):
Thanks

service/header/p2p_exchange.go
vgonkivs: I thought now we are using peers by default if nothing were provided...

liamsi (commented):
Hlib's comments are on point. I do not have anything to add :)

cmd/flags_header.go
vgonkivs: done

service/header/p2p_exchange.go
vgonkivs: removed

liamsi (approved):
👍👏🙌👌

node/services/config.go
```diff
@@ -17,32 +17,39 @@ type Config struct {
...
 	}
 }
 
-func (cfg *Config) trustedPeer() (*peer.AddrInfo, error) {
-	if cfg.TrustedPeer == "" {
-		log.Warn("No Trusted Peer provided. Headers won't be synced accordingly")
-		return &peer.AddrInfo{}, nil
+func (cfg *Config) trustedPeers() ([]*peer.AddrInfo, error) {
+	if len(cfg.TrustedPeers) == 0 {
+		cfg.TrustedPeers = params.Bootstrappers()
```
renaynay, line 36: We should log this IMO.

```
No trusted peers given, initializing with default bootstrappers as trusted peers
```

service/header/p2p_exchange.go
```diff
@@ -98,11 +97,14 @@ func (ex *P2PExchange) performRequest(ctx context.Context, req *pb.ExtendedHeade
 	if req.Amount == 0 {
 		return make([]*ExtendedHeader, 0), nil
 	}
-	if ex.trustedPeer == "" {
+
+	if len(ex.trustedPeers) == 0 {
 		return nil, fmt.Errorf("no trusted peer")
```
renaynay, line 102: "no trusted peers"

renaynay (commented):
Awesome! Some small comments.

service/header/p2p_exchange.go
vgonkivs: updated

node/services/config.go
vgonkivs: added log

service/header/p2p_exchange.go
Wondertan: Yes we are, but it is still possible to import P2PExchange from another project and give zero peers.
Think of packages inside ‘service’(will probably get extracted soon) as libraries that eventually will be used by other people and not just by us.

node/services/config.go
Wondertan: @renaynay, explain why. _Not_ passing trusted peers is going to be a default case after this PR, and there is no point to warn user that their are doing something wrong by using defaults

Wondertan: Though Info log is fine, unless it's not a Warning, like it was before

liamsi (approved):

node/services/config.go
renaynay: No, that's what i meant. It should be an INFO log.

renaynay: thank you for adding the log @vgonkivs , this comment thread is now resolved.

service/header/p2p_exchange.go
vgonkivs: condition added

Wondertan (approved):

PR #458, opened 2022-02-17
fix(service/header): lazily load Store head

### LINKED ISSUE #457: [BUG]: store is always initialized
After restart, the header Store of a Node is always reinitialized even if the reference to the new latest head exists.

Wondertan:
Fixes #457 by making store head being lazy loaded and not on Start.

### DIFF
--- a/service/header/store.go
+++ b/service/header/store.go
@@ -99,21 +99,11 @@ func (s *store) Init(_ context.Context, initial *ExtendedHeader) error {
 		return err
 	}
 
-	log.Infow("initial head", "height", initial.Height, "hash", initial.Hash())
+	log.Infow("initialized head", "height", initial.Height, "hash", initial.Hash())
 	return nil
 }
 
-func (s *store) Start(ctx context.Context) error {
-	head, err := s.readHead(ctx)
-	switch err {
-	default:
-		return err
-	case datastore.ErrNotFound:
-		log.Warnf("starting uninitialized store")
-	case nil:
-		s.heightSub.SetHeight(uint64(head.Height))
-	}
-
+func (s *store) Start(context.Context) error {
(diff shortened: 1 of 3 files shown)


service/header/store.go
```diff
@@ -140,10 +130,21 @@ func (s *store) Stop(ctx context.Context) error {
 
 func (s *store) Head(ctx context.Context) (*ExtendedHeader, error) {
 	head, err := s.GetByHeight(ctx, s.heightSub.Height())
-	if err == ErrNotFound {
+	if err == nil {
```
liamsi, line 133: ~~In the case `err != nil` the error might get overwritten by readHead and not handled a few lines below?~~
Ah, it's handled by the switch case.

service/header/store.go
```diff
@@ -140,10 +130,21 @@ func (s *store) Stop(ctx context.Context) error {
...
+		return head, err
```
liamsi, line 134:
```suggestion
		return head, nil
```

liamsi (dismissed):

Wondertan:
The initialization flow:
1 Node creates Store
2 Node initializes Store
	2a Checks if there is any head
	2b If not - requests and writes requested head to Store
3 Node starts Store

The bug was with the fact that step 2a required step 3, as Head was only loaded on start. Now that head is lazy-loaded, 2a can understand if there is head without 3

service/header/store.go
Wondertan: We that it is nil by the if case, but it doesn't hurt to say it is nil explicitly.

renaynay (approved):
MGMT

Wondertan:
@renaynay, yeah, nice band 😄 
> MGMT

Issue #453, opened 2022-02-17
[Bug]: Flakey TestCoreListener


Bidon15:
### Celestia Node version

In main https://github.com/celestiaorg/celestia-node/pull/442

### OS

Github action CI: Ubuntu

### Steps to reproduce it

Trigger CI tests with race detector

### Expected result

TestCoreListener is passing

### Actual result

TestCoreListener is failing

### Relevant log output

```shell
https://gist.github.com/Bidon15/24f4cd075ea5b2ff28e14d05affe6683
```

### Notes

This is not reproducible in local env

PR #459, opened 2022-02-17
service/headers: Fix race in `TestCoreListener`

renaynay:
Attempts fix for #453 

It is not locally reproducible so we will have to run the CI several times, but I applied the same fix as in #342. I am assuming this issue stems from the same issue addressed in that PR.

### DIFF
--- a/service/header/core_listener_test.go
+++ b/service/header/core_listener_test.go
@@ -2,14 +2,14 @@ package header
 
 import (
 	"context"
-	"github.com/libp2p/go-libp2p-core/event"
-	mocknet "github.com/libp2p/go-libp2p/p2p/net/mock"
-	"github.com/stretchr/testify/assert"
 	"testing"
 	"time"
 
 	mdutils "github.com/ipfs/go-merkledag/test"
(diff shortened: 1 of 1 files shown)


Wondertan:
@renaynay Try with
`go test ./service/header/ -count 1000 -run TestCoreListener -v -race`

service/header/core_listener_test.go
```diff
@@ -50,16 +54,40 @@ func TestCoreListener(t *testing.T) {
 }
 
 func createMocknetWithTwoPubsubEndpoints(t *testing.T) (*pubsub.PubSub, *pubsub.PubSub) {
-	host1, host2 := createMocknet(context.Background(), t)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*15)
+	t.Cleanup(cancel)
```
Wondertan, line 58: Can you pls make this context being global and defined in the test? So all the places where context is accepted take the same context and backgrounds.

Wondertan: Yes

Wondertan (approved):

liamsi (approved):

PR #462, opened 2022-02-21
feat(cmd/cel-shed): new header category and store-init cmd

Wondertan:
Implements a utility to force the store to change it is head. Was useful to recover from bug #457.
Very low priority to review

### DIFF
--- a/cmd/cel-shed/header.go
+++ b/cmd/cel-shed/header.go
@@ -36,6 +36,11 @@ Custom store path is not supported yet.`,
 			return fmt.Errorf("invalid node-type")
 		}
 
+		height, err := strconv.Atoi(args[1])
+		if err != nil {
+			return fmt.Errorf("invalid height: %w", err)
+		}
+
 		store, err := node.OpenStore(fmt.Sprintf("~/.celestia-%s", strings.ToLower(tp.String())), tp)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


cmd/cel-shed/header.go
```diff
@@ -0,0 +1,66 @@
...
+
+		hstore, err := header.NewStore(ds)
+		if err != nil {
+			return err
+		}
+
+		height, err := strconv.Atoi(args[1])
+		if err != nil {
+			return fmt.Errorf("ivalid height: %w", err)
+		}
```
liamsi, line 57: I think it would make sense to check the validity of this earlier (before writing to disk). Also:
 ```suggestion
		height, err := strconv.Atoi(args[1])
		if err != nil {
			return fmt.Errorf("invalid height: %w", err)
		}
```

Wondertan: True

Wondertan: Moved

liamsi (approved):

renaynay (approved):

Wondertan:
I tried resolving conflicts here, hoping it won't re-request reviews, but that doesn't work like this.

renaynay (approved):

PR #463, opened 2022-02-21
feat(cmd): cli flag to enable http/pprof handler to capture profiles

Wondertan:
Useful to profile the Node


cmd/flags_misc.go
```diff
@@ -40,5 +50,20 @@ func ParseMiscFlags(cmd *cobra.Command) error {
...
+	if ok, err := cmd.Flags().GetBool(pprofFlag); ok && err == nil {
+		// TODO(@Wondertan): Eventually, this should be registered on http server in RPC
+		go func() {
+			mux := http.NewServeMux()
+			mux.HandleFunc("/debug/pprof/", pprof.Index)
+			mux.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
+			mux.HandleFunc("/debug/pprof/profile", pprof.Profile)
+			mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
+			mux.HandleFunc("/debug/pprof/trace", pprof.Trace)
+			log.Println(http.ListenAndServe("0.0.0.0:6000", mux))
```
liamsi, line 62: Port should be user-controllable IMO.

cmd/flags_misc.go
```diff
@@ -25,6 +29,12 @@ func MiscFlags() *flag.FlagSet {
 and their lower-case forms`,
 	)
 
+	flags.Bool(
+		pprofFlag,
+		false,
+		"Enables std pprof handler to collect profiles",
```
liamsi, line 35:
```suggestion
		"Enables standard profiling handler (pprof) and exposes the profiles on port 6000",
```

liamsi (dismissed):

cmd/flags_misc.go
```diff
@@ -40,5 +50,20 @@ func ParseMiscFlags(cmd *cobra.Command) error {
...
+			mux.HandleFunc("/debug/pprof/profile", pprof.Profile)
+			mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
+			mux.HandleFunc("/debug/pprof/trace", pprof.Trace)
+			log.Println(http.ListenAndServe("0.0.0.0:6000", mux))
+		}()
+	} else if err != nil {
+		return err
+	}
+
 	return nil
```
renaynay, line 68: can we just return err here

renaynay (approved):
👍🏻

cmd/flags_misc.go
Wondertan: True, I would leave this for now and return back once we work through everything that is provided by HTTP. There will be a single port for everything including pprof.

Wondertan: https://github.com/celestiaorg/celestia-node/issues/469

cmd/flags_misc.go
Wondertan: Yes, updated

liamsi (changes requested):
image

liamsi (approved):

PR #464, opened 2022-02-21
perf(ipld/write): increase batch size

Wondertan:
The former TODO was incorrect. Writing each share in a separate IO is a bad idea, this change ensures even the biggest blocks will be written as on operation.

### DIFF
--- a/ipld/write.go
+++ b/ipld/write.go
@@ -14,11 +14,8 @@ import (
 )
 
 // BatchSize defines an amount of IPLD Nodes to be buffered and written at once.
-// This configuration is very database backend specific and the current default(128) is not optimized for the version of
-// Badger we use. We set it to one to avoid test flakiness, as some tests may read for data that was not written yet.
-// TODO(@Wondertan): Find out the perfect value for Badger(e.g. ask PL folks) or migrate to go-car IPLD
-//  storage(preferred).
-const BatchSize = 1
+// TODO(@Wondertan): Should rely on biggest extended block size param instead
+const BatchSize = 16384 // (128*128 so we flush big blocks in one IO write) //
 
 // PutData posts erasured block data to IPFS using the provided ipld.NodeAdder.
 func PutData(ctx context.Context, shares [][]byte, adder ipld.NodeAdder) (*rsmt2d.ExtendedDataSquare, error) {


liamsi (approved):

vgonkivs (approved):

ipld/write.go
```diff
@@ -14,11 +14,8 @@ import (
 )
 
 // BatchSize defines an amount of IPLD Nodes to be buffered and written at once.
-// This configuration is very database backend specific and the current default(128) is not optimized for the version of
-// Badger we use. We set it to one to avoid test flakiness, as some tests may read for data that was not written yet.
-// TODO(@Wondertan): Find out the perfect value for Badger(e.g. ask PL folks) or migrate to go-car IPLD
-//  storage(preferred).
-const BatchSize = 1
+// TODO(@Wondertan): Should rely on biggest extended block size param instead
+const BatchSize = 16384 // (128*128 so we flush big blocks in one IO write) //
```
vgonkivs, line 18: This const variable can be deleted because we can compute anytime we want to send a batch,  by multiplying `(edsWidth/2) * (edsWidth/2)`.

vgonkivs (changes requested):

ipld/write.go
Wondertan: Good idea

Wondertan: One thing, I would make it `edsWidth*edsWidth`, as we will store it's full size including what was repaired

Wondertan: But also, there will be tree nodes, so the amount of ipld.Node getting written is even higher than edsWidth*edsWidth

Wondertan: Not sure though how to calculate the amount of nodes in a binary tree with a formula

Wondertan: @liamsi, do you know?

liamsi: For a perfect binary merkle tree the number of all nodes (including the leaves) is always 2^(n+1) - 1 (where n = number of leaves)

vgonkivs: lgtm

vgonkivs (approved):

ipld/write.go
```diff
@@ -55,3 +48,12 @@ func ExtractODSShares(eds *rsmt2d.ExtendedDataSquare) [][]byte {
...
 	return origShares
 }
+
+// batchSize calculates the amount of nodes that are generated from block of 'squareSizes'
+// to be batched in one write.
+func batchSize(squareSize int) int {
+	// (squareSize*squareSize) - all the shares
+	// (squareSize*2-1) - amount of nodes in a generated binary tree
+	// *squareSize*2 - multiplier for the amount of trees, both over rows and cols
+	return squareSize*squareSize + (squareSize*2-1)*squareSize*2
```
liamsi, line 58: Ideally, this would be exactly the number of all nodes of all trees of one block of total width `squareSize`, right?

Wondertan: Yes, I hope so, but I could make a mistake here.

renaynay (commented):
can you provide a test here?

Wondertan:
> can you provide a test here?
@renaynay, I am fine with having one, not sure who test this though. Do you have an idea?

ipld/write.go
liamsi: ah, I see. Then this is correct!

liamsi: maybe add a comment in the code as well why we count the leaves twice?

liamsi (approved):

ipld/write.go
Wondertan: Good

liamsi: > Yes, but do you remember that in IPLD implementation we have one more link per each leave.


Can you see a reason we would ever remove that property and have the hashes in the leaf nodes directly? I can't think of a good reason at least..

Wondertan: > Can you see a reason we would ever remove that property and have the hashes in the leaf nodes directly?

I think we should remove this property. This only adds additional roundtrips on the network + more block data size amplification. I remember you did this for a reason and I remember that I had a solution for that reason in my mind, but not in detail right now. Also, IIRC, I mentioned this in the storage optimization issue.

liamsi: Hmm, then it will be tricky to remember to change this here as well. If there is an issue for this (the one you mention about optimizing storage) it should probably link to this PR and the above method even IMO.

Wondertan: Just added more info regarding the tree, will also add a link to storage optimization with cross-linking

ipld/write.go
```diff
@@ -13,23 +13,16 @@ import (
...
 // PutData posts erasured block data to IPFS using the provided ipld.NodeAdder.
 func PutData(ctx context.Context, shares [][]byte, adder ipld.NodeAdder) (*rsmt2d.ExtendedDataSquare, error) {
 	if len(shares) == 0 {
 		return nil, fmt.Errorf("empty data") // empty block is not an empty Data
 	}
-	// create nmt adder wrapping batch adder
-	batchAdder := NewNmtNodeAdder(ctx, ipld.NewBatch(ctx, adder, ipld.MaxSizeBatchOption(BatchSize)))
+	squareSize := int(math.Sqrt(float64(len(shares))))
+	// create nmt adder wrapping batch adder with calculated size
+	batchAdder := NewNmtNodeAdder(ctx, ipld.NewBatch(ctx, adder, ipld.MaxSizeBatchOption(batchSize(squareSize))))
```
liamsi, line 23: > `squareSize := int(math.Sqrt(float64(len(shares))))`

These are only the orig. square shares, right? But I thought `batchSize` was supposed to return the number of all nodes in the extended square? 
```suggestion
	squareSize := int(math.Sqrt(float64(len(shares))))
	// create nmt adder wrapping batch adder with calculated size
	bs := batchSize(squareSize * 2)
	batchAdder := NewNmtNodeAdder(ctx, ipld.NewBatch(ctx, adder, ipld.MaxSizeBatchOption(bs)))
```

ipld/write.go
```diff
@@ -55,3 +48,19 @@ func ExtractODSShares(eds *rsmt2d.ExtendedDataSquare) [][]byte {
...
+	// (squareSize*2-1) - amount of nodes in a generated binary tree
+	// *squareSize*2 - multiplier for the amount of trees, both over rows and cols
+	//
+	// Note that our IPLD tree looks like:
+	// ---X
+	// -X---X
+	// X-X-X-X
+	// X-X-X-X
+	// So we count leaves two times here for a reason. https://github.com/celestiaorg/celestia-node/issues/183
+	return squareSize*squareSize + (squareSize*2-1)*squareSize*2
```
liamsi, line 65: OK, so this is really weird but the numbers of keys readable via `bs.AllKeysChan(ctx)` as per test in https://github.com/celestiaorg/celestia-node/pull/471
always matches this formula instead: 
```suggestion
	return (squareSize*2-1)*squareSize*2 - (squareSize * squareSize)
```

Any idea why we can remove the a CID per leaf/share?

Wondertan: No idea right now. Will thing with clear mind tmrw

liamsi: Ah, I think i understand: the shares / leaves yield the same CIDs independent of if we compute row or column trees. But in `(squareSize*2-1)*squareSize*2` we compute all  hashes for leaves for rows and columns too (where the leaves yield the same keys/CIDs). hence we have to subtract these.
I guess the CIDs for the last two layers are the same as well? here:
// X-X-X-X
// X-X-X-X

Wondertan: Ohhhhh, yesss

Wondertan: This explains it

Wondertan: That additional layer is always the same for rows and col.

liamsi (dismissed):

liamsi (approved):
If all tests pass now, we are good IMO.

Wondertan:
Updating the comment?

liamsi:
Ah yeah, we should:
 - [ ] update the code comment(s)
 - [ ] run `go mod tidy`
 - [x] fix typos in the opening comment

liamsi (approved):
now: https://github.com/celestiaorg/celestia-node/pull/464#pullrequestreview-890429275

ipld/write.go
```diff
@@ -55,3 +49,20 @@ func ExtractODSShares(eds *rsmt2d.ExtendedDataSquare) [][]byte {
...
+	// (squareSize*2-1) - amount of nodes in a generated binary tree
+	// squareSize*2 - the total number of trees, both for rows and cols
+	// (squareSize*squareSize) - all the shares
+	//
+	// Note that while our IPLD tree looks like this:
+	// ---X
+	// -X---X
+	// X-X-X-X
+	// X-X-X-X
+	// here we count leaves only once: the CIDs are the same for columns and wows
```
liamsi, line 65: wows :D
```suggestion
	// here we count leaves only once: the CIDs are the same for columns and rows
```

Wondertan:
Approved! Wanna wait until tmrw for others to look or should we merge now?

liamsi:
no need to rush from my side.

liamsi:
` make test-unit-race` seems to fail the freshly introduced test 🤔 
https://github.com/celestiaorg/celestia-node/runs/5295737307?check_suite_focus=true#step:5:108
additionally: TestCoreListener https://github.com/celestiaorg/celestia-node/runs/5295737307?check_suite_focus=true#step:5:257

ipld/read_test.go
```diff
@@ -365,6 +371,43 @@ func TestGetLeavesByNamespace_MultipleRowsContainingSameNamespaceId(t *testing.T
...
+		{"2", 2},
+		{"4", 4},
+		{"8", 8},
+		{"16", 16},
+		{"32", 32},
+		// {"64", 64}, // test case too large for CI with race detector
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ctx, cancel := context.WithTimeout(context.Background(), time.Second*time.Duration(tt.origWidth))
```
liamsi, line 388: 32x32 ODS needs roughly 4 seconds
63x64 over 30s

Wondertan: On ci, I guess, not locally

Wondertan: So are you done with it?

renaynay (approved):
Thanks for adding tests :)

ipld/read_test.go
liamsi: yes, I think it's ok as is

PR #465, opened 2022-02-21
perf(node/store): fine-tune Badgerdb params

Wondertan:
Motivations for each configuration field are described in the comments.
Need to be reviewed ASAP for the release.

### DIFF
--- a/CHANGELOG-PENDING.md
+++ b/CHANGELOG-PENDING.md
@@ -21,6 +21,7 @@ Month, DD, YYYY
 - [swamp: initial structure of the tool](https://github.com/celestiaorg/celestia-node/pull/315) [@Bidon15](https://github.com/Bidon15)
 
 ### IMPROVEMENTS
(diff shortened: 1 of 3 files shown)


liamsi (approved):
I did not verify each option individually. But the comments help a lot (thanks!) 

In general I'm in favour of optimizing everything later (and ideally backed by empirical data). That said, this looks like a no-brainer to me.

Wondertan:
@liamsi, I forgot to mention that I had to dive deeper into this to understand what caused the OOMs during sync, as it was caused during Badger compactions. Also, this tuning fixed the issue and reduced memory requirements a lot, while improving performance. 

My droplet doesn't need more than 3Gb of RAM during compactions anymore and it consumes ~1.8GB while syncing Bridge node and with Validator running.

Wondertan:
For the light node, we can go even further and make it run below 100 MB and even 50mb only, after looking at profiling insights.

renaynay (approved):

PR #466, opened 2022-02-21
feat(node): add go-watchdog to curb OOMs

Wondertan:
Note that the [watchdog](github.com/raulk/go-watchdog) didn't fix the OOM issue I had during synchronization and I am not entirely sure the addition of this utility is fully justified in our case. However, the problem it solves is real and according to PL folks, it helps them to prevent OOMs caused by occasional spikes of memory allocation that Badger made, like in our case. Also, the problem of Golang runtime not being able to execute GC in such cases, preventing the OOM is real and the Golang team confirmed this. To understand the issue you can narrow down through all the issues that watchdogs links to in his README. One more good point of watchdog, which sells it for me, is that it automatically captures heap profiles when heap size exceeds some threshold and I would like to have this feature, as in the future, this can facilitate us with debugging future OOMs. We would just need to analyze pprof files saved on disk by the watchdog and ask folks to provide those profiles if they report an OOM.

### DIFF
--- a/CHANGELOG-PENDING.md
+++ b/CHANGELOG-PENDING.md
@@ -23,6 +23,7 @@ Month, DD, YYYY
 - [swamp: initial structure of the tool](https://github.com/celestiaorg/celestia-node/pull/315) [@Bidon15](https://github.com/Bidon15)
 
 ### IMPROVEMENTS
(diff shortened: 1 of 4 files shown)


liamsi:
Issue was already fixed with https://github.com/celestiaorg/celestia-node/pull/465

liamsi (dismissed):

renaynay (dismissed):
LGTM - will this cause log pollution?

liamsi (approved):

PR #468, opened 2022-02-21
chore(service/header): minor code cleanups

Wondertan:
Last cleanups before the release, also include cache size changes, again.

Review commit by commit.

### DIFF
--- a/service/header/store.go
+++ b/service/header/store.go
@@ -170,7 +170,7 @@ func (s *store) Get(_ context.Context, hash tmbytes.HexBytes) (*ExtendedHeader, 
 
 func (s *store) GetByHeight(ctx context.Context, height uint64) (*ExtendedHeader, error) {
 	// if the requested 'height' was not yet published
-	// we subscribe and for it
+	// we subscribe to it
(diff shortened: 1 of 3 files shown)


renaynay (approved):

service/header/store.go
```diff
@@ -14,9 +14,9 @@ import (
 // TODO(@Wondertan): Those values must be configurable and proper defaults should be set for specific node type.
 var (
 	// DefaultStoreCacheSize defines the amount of max entries allowed in the Header Store cache.
-	DefaultStoreCacheSize = 10240
+	DefaultStoreCacheSize = 4096
 	// DefaultIndexCacheSize defines the amount of max entries allowed in the Height to Hash index cache.
-	DefaultIndexCacheSize = 2048
+	DefaultIndexCacheSize = 16384
```
liamsi, line 19: It's impossible for a reviewer to know why these numbers are better than others. Instead of repeatedly changing and tweaking these now, we should instead come up with metrics that we want to improve. 
Doesn't require any changes in this PR though.

liamsi (approved):
Aprooove!

PR #471, opened 2022-02-22
test(ipld): a test for batchSize

Wondertan:


ipld/read_test.go
```diff
@@ -365,6 +371,27 @@ func TestGetLeavesByNamespace_MultipleRowsContainingSameNamespaceId(t *testing.T
...
+
+	out, err := bs.AllKeysChan(ctx)
+	require.NoError(t, err)
+
+	var count int
+	for range out {
+		count++
+	}
+
+	assert.Equal(t, count, batchSize(64))
```
liamsi, line 392: Even with the correct batch size via https://github.com/celestiaorg/celestia-node/pull/464/files#r812345568
this won't work. The reason is that `out, err := bs.AllKeysChan(ctx)` will only report different CIDs while here 4096 CIDs seem to always be the same. I'm confused about the latter rn. Could this be due to the way the wrapper handles the parity leaves?

ipld/read_test.go
```diff
@@ -365,6 +371,27 @@ func TestGetLeavesByNamespace_MultipleRowsContainingSameNamespaceId(t *testing.T
...
 }
 
+func TestBatchSize(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+
+	bs := blockstore.NewBlockstore(dssync.MutexWrap(ds.NewMapDatastore()))
+	dag := merkledag.NewDAGService(blockservice.New(bs, offline.Exchange(bs)))
+	eds := generateRandEDS(t, 32)
+	_, err := PutData(ctx, ExtractODSShares(eds), dag)
```
liamsi, line 381:
```suggestion
	origWidth := 32
	eds := generateRandEDS(t, origWidth)
	_, err := PutData(ctx, ExtractODSShares(eds), dag)
```

ipld/read_test.go
liamsi, line 392:
```suggestion
	extendedWidth := origWidth * 2
	assert.Equal(t, count, batchSize(extendedWidth))
```

ipld/read_test.go
```diff
@@ -365,6 +371,28 @@ func TestGetLeavesByNamespace_MultipleRowsContainingSameNamespaceId(t *testing.T
...
+
+	out, err := bs.AllKeysChan(ctx)
+	require.NoError(t, err)
+
+	var count int
+	for range out {
+		count++
+	}
+	extendedWidth := origWidth * 2
+	assert.Equal(t, count, batchSize(extendedWidth))
```
liamsi, line 393:
```suggestion
	tests := []struct {
		name      string
		origWidth int
	}{
		{"2", 2},
		{"4", 4},
		{"8", 8},
		{"16", 16},
		{"32", 32},
		{"64", 64},
	}
	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			ctx, cancel := context.WithTimeout(context.Background(), time.Second)
			defer cancel()

			bs := blockstore.NewBlockstore(dssync.MutexWrap(ds.NewMapDatastore()))
			dag := merkledag.NewDAGService(blockservice.New(bs, offline.Exchange(bs)))

			eds := generateRandEDS(t, tt.origWidth)
			_, err := PutData(ctx, ExtractODSShares(eds), dag)
			require.NoError(t, err)

			out, err := bs.AllKeysChan(ctx)
			require.NoError(t, err)

			var count int
			for range out {
				count++
			}
			extendedWidth := tt.origWidth * 2
			assert.Equalf(t, count, batchSize(extendedWidth), "batchSize(%v)", extendedWidth)
		})
	}
```

Issue #181, opened 2021-11-09
das: DASer checkpoints and DASing of past headers


Wondertan:
## Background 
Right now, DASer(#177) only implements continuous sampling over the newest ongoing Headers. Unfortunately, when a Node goes offline it stops listening for Headers, but when it goes back online it should execute DASing for missed Headers.

## Design
To fix the problem above we should introduce checkpointing for DASer with a warm-up logic. Once a Node goes gracefully offline, the DASer should persist a pointer to the latest DASed Header. Once it goes back online, the DASer will see the newest Head and compare it to the checkpoint. If there is a difference, it should execute DASIng for all missed Headers.

Issue #427, opened 2022-02-11
das: Implement DASer public API


Bidon15:
## Context
Currently, our `daser` only exposes Start/Stop.
 
In order to get more detailed information from the `daser`, `DASState` must be implemented.

Ideas for API methods for the `daser`:

* status of the daser(is it active/not active) 
* error handling (atm of writing, we are only logging error messages) 
*  height/hash info
* square width info

Some of the above can be modelled after [`SyncState`](#397).

## TODO

- [ ] Implement `DASState`
- [ ] Implement above suggested methods on `daser`

liamsi:
Is this resolved with #397?

Bidon15:
> Is this resolved with #397?

The reference is what is the example and expectation to be done for DASer 

This is going to be resolved with #473 and upcoming changes (usually I can cover some of the basics during test devs)

renaynay:
Well it won't really be covered by #473 but a subsequent PR that will implement DASState where we can access realtime info on DASer

renaynay:
I'm not sure errors with sampling should be stored in DASState -- but rather CacheAvailability

Issue #180, opened 2021-11-09
share: Cache Availability


Wondertan:
## Background
We have defined `share.Availability` interface which validates the availability of shares for a specific `share.Root`. In case we call `SharesAvailable` multiple times over the same Root, we will execute availability again. Instead, we should not execute expensive validation and perform caching of successful and failed executions. This way Light Availability won't sample shares again over the same Root. 

## Design
To fix the problem above we should introduce another implementation of `share.Availability` which wraps e.g Light Availability and caches the results on disk.

Basically, on `SharesAvailable` Cache Availability checks if there is a cached result on disk for a given Root and if not pass the call to `SharesAvailable` of wrapped `share.Availability`. Once done, Cache Availability saves result on disk for the Root. Further, If someone calls `SharesAvailable` Cache Availability  will return result from cache, without re-executing `SharesAvailable` on a wrapped `share.Availability`.

renaynay:
<img width="824" alt="Screenshot 2022-03-21 at 18 49 55" src="https://user-images.githubusercontent.com/41963722/159333699-6840d1e9-9479-4514-89d5-139187f467cb.png">

- @Wondertan comment on #473

PR #473, opened 2022-02-23
das: DASing of past headers

renaynay:
This PR implements DASing of past headers in a single routine. 

Eventually, this will be parallelised. 

## TODO 
- [x] figure out what to do in the instance that DASer runs and samples new headers [100:150] and DASer is stopped, so sample routine stores latest checkpoint to disk as last DASed height (150) but catchUp routine has only sampled from [1:40] so there is a gap missing from (40: 100)?
- [x] fix constructor
- [x] cleanup 
- [x] lint
- [x] doc.go
- [x] open issue for parallelisation of DASer
- [x] open issue for `DASState`

Resolves https://github.com/celestiaorg/celestia-node/issues/181.

### DIFF
--- a/das/checkpoint_store.go
+++ b/das/checkpoint_store.go
@@ -1,0 +1,51 @@
+package das
+
+import (
+	"encoding/binary"
+
+	"github.com/ipfs/go-datastore"
+	"github.com/ipfs/go-datastore/namespace"
+)
+
+var (
+	storePrefix   = datastore.NewKey("das")
+	checkpointKey = datastore.NewKey("checkpoint")
+)
+
+// NewCheckpointStore wraps the given datastore.Datastore with
+// the `das` prefix.
+func NewCheckpointStore(ds datastore.Datastore) datastore.Datastore {
+	return namespace.Wrap(ds, storePrefix)
+
+}
+
+// loadCheckpoint // TODO @renaynay: document
+//
+// 1. gets checkpoint from disk
+// 1a. if there is no checkpoint, you start from `GetByHeight` of 1
+// 1b. if there is a checkpoint, continue
+func loadCheckpoint(ds datastore.Datastore) (int64, error) {
+	checkpoint, err := ds.Get(checkpointKey)
+	if err != nil {
+		// if no checkpoint was found, return checkpoint as
+		// 0 since DASer begins sampling on checkpoint+1
+		if err == datastore.ErrNotFound {
+			log.Debug("checkpoint not found, starting sampling at block height 1")
+			return 0, nil
+		}
+
+		return 0, err
+	}
+	return int64(binary.BigEndian.Uint64(checkpoint)), err
+}
+
+// TODO @renaynay: document
+
+// storeCheckpoint stores the given DAS checkpoint to disk.
+func storeCheckpoint(ds datastore.Datastore, checkpoint int64) error {
+	buf := make([]byte, 8)
+	binary.BigEndian.PutUint64(buf, uint64(checkpoint))
+
+	return ds.Put(checkpointKey, buf)
+}
+
--- a/das/checkpoint_store.go
+++ b/das/checkpoint_store.go
@@ -16,7 +16,6 @@ var (
 // the `das` prefix.
 func NewCheckpointStore(ds datastore.Datastore) datastore.Datastore {
 	return namespace.Wrap(ds, storePrefix)
-
 }
 
 // loadCheckpoint loads the DAS checkpoint from disk and returns it.
--- a/das/checkpoint_store.go
+++ b/das/checkpoint_store.go
@@ -12,9 +12,11 @@ var (
 	checkpointKey = datastore.NewKey("checkpoint")
 )
 
-// NewCheckpointStore wraps the given datastore.Datastore with
-// the `das` prefix.
-func NewCheckpointStore(ds datastore.Datastore) datastore.Datastore {
+// wrapCheckpointStore wraps the given datastore.Datastore with the `das`
+// prefix. The checkpoint store stores/loads the DASer's checkpoint to/from
+// disk using the checkpointKey. The checkpoint is stored as a uint64
+// representation of the height of the latest successfully DASed header.
+func wrapCheckpointStore(ds datastore.Datastore) datastore.Datastore {
 	return namespace.Wrap(ds, storePrefix)
 }
 
--- a/das/checkpoint_store.go
+++ b/das/checkpoint_store.go
@@ -2,7 +2,6 @@ package das
 
 import (
 	"encoding/binary"
-
 	"github.com/ipfs/go-datastore"
 	"github.com/ipfs/go-datastore/namespace"
 )
@@ -39,8 +38,6 @@ func loadCheckpoint(ds datastore.Datastore) (int64, error) {
 	return int64(binary.BigEndian.Uint64(checkpoint)), err
 }
 
-// TODO @renaynay: document
-
 // storeCheckpoint stores the given DAS checkpoint to disk.
 func storeCheckpoint(ds datastore.Datastore, checkpoint int64) error {
 	buf := make([]byte, 8)
@@ -48,4 +45,3 @@ func storeCheckpoint(ds datastore.Datastore, checkpoint int64) error {
 
 	return ds.Put(checkpointKey, buf)
 }
-
--- a/das/checkpoint_store.go
+++ b/das/checkpoint_store.go
@@ -20,7 +20,7 @@ func wrapCheckpointStore(ds datastore.Datastore) datastore.Datastore {
 	return namespace.Wrap(ds, storePrefix)
 }
 
-// loadCheckpoint loads the DAS checkpoint from disk and returns it.
+// loadCheckpoint loads the DAS checkpoint height from disk and returns it.
 // If there is no known checkpoint, it returns height 0.
 func loadCheckpoint(ds datastore.Datastore) (int64, error) {
 	checkpoint, err := ds.Get(checkpointKey)
--- a/das/checkpoint_store.go
+++ b/das/checkpoint_store.go
@@ -19,11 +19,8 @@ func NewCheckpointStore(ds datastore.Datastore) datastore.Datastore {
 
 }
 
-// loadCheckpoint // TODO @renaynay: document
-//
-// 1. gets checkpoint from disk
-// 1a. if there is no checkpoint, you start from `GetByHeight` of 1
-// 1b. if there is a checkpoint, continue
+// loadCheckpoint loads the DAS checkpoint from disk and returns it.
+// If there is no known checkpoint, it returns height 0.
 func loadCheckpoint(ds datastore.Datastore) (int64, error) {
 	checkpoint, err := ds.Get(checkpointKey)
 	if err != nil {
--- a/das/checkpoint_store.go
+++ b/das/checkpoint_store.go
@@ -2,6 +2,7 @@ package das
 
 import (
 	"encoding/binary"
+
 	"github.com/ipfs/go-datastore"
 	"github.com/ipfs/go-datastore/namespace"
 )
--- a/das/daser.go
+++ b/das/daser.go
@@ -122,7 +122,7 @@ func (d *DASer) sample(ctx context.Context, sub header.Subscription, checkpoint 
 		// to our last DASed header, kick off routine to DAS all headers
 		// between last DASed header and h. This situation could occur
 		// either on start or due to network latency/disconnection.
-		if h.Height > (height + 1) { // TODO @renaynay: what if height = 7 and h.Height = 9 (so only need to catch up on 8
+		if h.Height > (height + 1) {
 			// DAS headers between last DASed height up to the current
 			// header
 			go d.catchUp(ctx, height, h.Height-1)
--- a/das/daser.go
+++ b/das/daser.go
@@ -176,11 +176,11 @@ func (d *DASer) catchUpScheduler(ctx context.Context, checkpoint int64) {
 		if err := storeCheckpoint(d.cstore, checkpoint); err != nil {
 			log.Errorw("storing checkpoint to disk", "height", checkpoint, "err", err)
 		}
-		// signal that all catchUp routines have finished
-		close(d.catchUpDn)
 		// close out jobs channel
 		close(d.jobsCh)
 		d.jobsCh = nil
+		// signal that all catchUp routines have finished
+		close(d.catchUpDn)
 	}()
 
 	for {
--- a/das/daser.go
+++ b/das/daser.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/ipfs/go-datastore"
 	logging "github.com/ipfs/go-log/v2"
 
 	"github.com/celestiaorg/celestia-node/service/header"
@@ -18,21 +19,35 @@ type DASer struct {
 	da   share.Availability
 	hsub header.Subscriber
 
+	// getter allows the DASer to fetch a header at a certain height
+	// and blocks until it becomes available.
+	getter HeaderGetter
+	// checkpoint from disk -- DASStore stores checkpoint w/ DASCheckpoint key
+	// checkpoint = latest successfully DASed header (reference to it, not the header)
+	ds datastore.Datastore
+
 	cancel context.CancelFunc
 	done   chan struct{}
 }
 
 // NewDASer creates a new DASer.
-func NewDASer(da share.Availability, hsub header.Subscriber) *DASer {
+func NewDASer(
+	da share.Availability,
+	hsub header.Subscriber,
+	getter HeaderGetter,
+	ds datastore.Datastore,
+) *DASer {
 	return &DASer{
-		da:   da,
-		hsub: hsub,
-		done: make(chan struct{}),
+		da:     da,
+		hsub:   hsub,
+		getter: getter,
+		ds:     ds,
+		done:   make(chan struct{}),
 	}
 }
 
 // Start initiates subscription for new ExtendedHeaders and spawns a sampling routine.
-func (d *DASer) Start(context.Context) error {
+func (d *DASer) Start(ctx context.Context) error {
 	if d.cancel != nil {
 		return fmt.Errorf("da: DASer already started")
 	}
@@ -42,8 +57,30 @@ func (d *DASer) Start(context.Context) error {
 		return err
 	}
 
-	ctx, cancel := context.WithCancel(context.Background())
-	go d.sampling(ctx, sub)
+	// load latest DASed checkpoint
+	checkpoint, err := loadCheckpoint(d.ds)
+	if err != nil {
+		return err
+	}
+	log.Infow("loaded latest DASed checkpoint", "height", checkpoint)
+
+	// load current network head // TODO @renaynay: do we have to sample over netHead as well or will that be handled by sampleLatest via hsub?
+	netHead, err := d.getter.Head(ctx)
+	if err != nil {
+		return err
+	}
+
+	dasCtx, cancel := context.WithCancel(context.Background())
+
+	// start two separate routines:
+	// 1. samples headers from the latest DASed header to the
+	//    current network head (samples headers from the past)
+	// 2. samples new headers coming through the ExtendedHeader
+	//    gossipsub topic (samples new inbound headers in the
+	//    network)
+	go d.sampleFromCheckpoint(dasCtx, checkpoint, netHead.Height)
+	go d.sampleLatest(dasCtx, sub)
+
 	d.cancel = cancel
 	return nil
 }
@@ -52,6 +89,7 @@ func (d *DASer) Start(context.Context) error {
 func (d *DASer) Stop(ctx context.Context) error {
 	d.cancel()
 	select {
+	// TODO @renaynay: add done case for sampleCheckpoint
 	case <-d.done:
 		d.cancel = nil
 		return nil
@@ -60,8 +98,55 @@ func (d *DASer) Stop(ctx context.Context) error {
 	}
 }
 
-// sampling validates availability for each Header received from header subscription.
-func (d *DASer) sampling(ctx context.Context, sub header.Subscription) {
+// sampleFromCheckpoint starts sampling headers from the last known
+// `checkpoint` (latest/heighest DASed header), and breaks the loop
+// once network head `netHead` is reached, leaving only the `sampleLatest`
+// loop running.
+//
+// 2. Loop:
+// 2a. request by height (checkpoint+1)
+// 2b. DAS on that height
+// 2c. stop when height == network height
+func (d *DASer) sampleFromCheckpoint(ctx context.Context, checkpoint, netHead int64) {
+	// immediately break if DASer is up to speed with network head
+	if checkpoint+1 == netHead { // TODO @renaynay: test this scenario
+		log.Debugw("caught up to network head", "net head", netHead)
+		return
+	}
+
+	// start sampling from the next header after the checkpoint as the
+	// checkpoint has already been successfully DASed.
+	for height := checkpoint + 1; height < netHead; height++ {
+		h, err := d.getter.GetByHeight(ctx, uint64(height))
+		if err != nil {
+			if err == context.Canceled {
+				return
+			}
+
+			log.Errorw("failed to get next header", "height", height, "err", err)
+			continue // TODO @renaynay: should we really continue in this case?
+		}
+
+		startTime := time.Now()
+
+		err = d.da.SharesAvailable(ctx, h.DAH)
+		if err != nil {
+			if err == context.Canceled {
+				return
+			}
+			log.Errorw("sampling failed", "height", h.Height, "hash", h.Hash(),
+				"square width", len(h.DAH.RowsRoots), "data root", h.DAH.Hash(), "err", err)
+			// continue sampling
+		}
+
+		sampleTime := time.Since(startTime)
+		log.Infow("sampling successful", "height", h.Height, "hash", h.Hash(),
+			"square width", len(h.DAH.RowsRoots), "finished (s)", sampleTime.Seconds())
+	}
+}
+
+// sampleLatest validates availability for each Header received from header subscription. // TODO renaynay: rename to sampleLatest
+func (d *DASer) sampleLatest(ctx context.Context, sub header.Subscription) {
 	defer sub.Cancel()
 	defer close(d.done)
 	for {
--- a/das/daser.go
+++ b/das/daser.go
@@ -122,7 +122,7 @@ func (d *DASer) sample(ctx context.Context, sub header.Subscription, checkpoint 
 		// to our last DASed header, kick off routine to DAS all headers
 		// between last DASed header and h. This situation could occur
 		// either on start or due to network latency/disconnection.
-		if h.Height > (height+1) { // TODO @renaynay: what if height = 7 and h.Height = 9 (so only need to catch up on 8
+		if h.Height > (height + 1) { // TODO @renaynay: what if height = 7 and h.Height = 9 (so only need to catch up on 8
 			// DAS headers between last DASed height up to the current
 			// header
 			go d.catchUp(ctx, height, h.Height-1)
@@ -159,9 +159,7 @@ func (d *DASer) catchUp(ctx context.Context, from, to int64) {
 
 	// start sampling from height at checkpoint+1 since the
 	// checkpoint height has already been successfully DASed
-	fmt.Println("from: ", from, " to: ", to)
 	for height := from + 1; height <= to; height++ {
-		fmt.Println("catchup loop height: ", height)
 		h, err := d.getter.GetByHeight(ctx, uint64(height))
 		if err != nil {
 			if err == context.Canceled {
--- a/das/daser.go
+++ b/das/daser.go
@@ -73,6 +73,9 @@ func (d *DASer) Start(context.Context) error {
 	d.cancel = cancel
 	d.sampleDn, d.catchUpDn = make(chan struct{}), make(chan struct{})
 
+	// kick off catch-up routine scheduler
+	go d.catchUpScheduler(dasCtx, checkpoint)
+	// kick off sampling routine for recently received headers
 	go d.sample(dasCtx, sub, checkpoint)
 	return nil
 }
@@ -101,9 +104,6 @@ func (d *DASer) sample(ctx context.Context, sub header.Subscription, checkpoint 
 		sub.Cancel()
 		close(d.sampleDn)
 	}()
-
-	// kick off catchUpScheduler
-	go d.catchUpScheduler(ctx, checkpoint)
 
 	for {
 		h, err := sub.NextHeader(ctx)
--- a/das/daser.go
+++ b/das/daser.go
@@ -26,8 +26,9 @@ type DASer struct {
 	// checkpoint = latest successfully DASed header (reference to it, not the header)
 	ds datastore.Datastore
 
-	cancel context.CancelFunc
-	done   chan struct{}
+	cancel             context.CancelFunc
+	sampleLatestDn     chan struct{} // done signal for sampleLatest loop
+	sampleCheckpointDn chan struct{} // done signal for sampleFromCheckpoint loop
 }
 
 // NewDASer creates a new DASer.
@@ -38,11 +39,12 @@ func NewDASer(
 	ds datastore.Datastore,
 ) *DASer {
 	return &DASer{
-		da:     da,
-		hsub:   hsub,
-		getter: getter,
-		ds:     ds,
-		done:   make(chan struct{}),
+		da:                 da,
+		hsub:               hsub,
+		getter:             getter,
+		ds:                 ds,
+		sampleLatestDn:     make(chan struct{}),
+		sampleCheckpointDn: make(chan struct{}),
 	}
 }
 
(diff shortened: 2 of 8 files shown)


das/daser.go
```diff
@@ -18,21 +19,37 @@ type DASer struct {
 	da   share.Availability
 	hsub header.Subscriber
 
-	cancel context.CancelFunc
-	done   chan struct{}
+	// getter allows the DASer to fetch a header at a certain height
+	// and blocks until it becomes available.
```
liamsi, line 23: @adlerjohn suggested in the past to the DASing and the availability check async. But IMO we could do this blocking  as it's done here first.

das/daser_test.go
```diff
@@ -4,29 +4,153 @@ import (
...
+
+	ctx, cancel := context.WithCancel(context.Background())
+	t.Cleanup(cancel)
+
+	daser := NewDASer(shareServ, nil, mockGet, cstore)
+
+	wg := &sync.WaitGroup{}
+	wg.Add(1)
+	go func(wg *sync.WaitGroup) {
+		daser.catchUp(ctx, 2, mockGet.head) // pick random number as `from` parameter
```
liamsi, line 65: which random number? `2`?

das/daser_test.go
```diff
@@ -4,29 +4,153 @@ import (
...
+// TestDASer_catchUp_oneHeader tests that catchUp works with a from-to
+// difference of 1
+func TestDASer_catchUp_oneHeader(t *testing.T) {
+	ds := ds_sync.MutexWrap(datastore.NewMapDatastore())
+	cstore := NewCheckpointStore(ds)
+
+	mockGet, shareServ, _ := createDASerSubcomponents(t, 6, 0)
+
+	// store checkpoint
+	err := storeCheckpoint(cstore, 5) // pick random header as last checkpoint
```
liamsi, line 80: If I read this comment I think `storeCheckpoint` picks a random number 🤔

das/doc.go
```diff
@@ -0,0 +1,22 @@
...
+sufficiently likely that all block data is available as it is assumed that there
+are enough `light` availability instances active on the network doing sampling over
+the same block to collectively verify its availability.
+
+The central component of this package is the `DASer`. It performs one basic function:
+a sampling loop that performs DAS on new ExtendedHeaders in the network. The DASer kicks
+off this loop by loading its last DASed header (`checkpoint`) and kicking off a `catchUp`
+loop to DAS all headers between the checkpoint and the current network head. It simultaneously
+continues to perform DAS over new ExtendedHeaders received via gossipsub.
+*/
```
liamsi, line 21: Awesome!!!

liamsi (commented):
Left some nits. Overall looks very good.

das/daser.go
adlerjohn: Yeah that would be a cute optimization, but not strictly necessary.

das/doc.go
adlerjohn: +1 damn son

das/checkpoint_store.go
```diff
@@ -0,0 +1,45 @@
...
+	storePrefix   = datastore.NewKey("das")
+	checkpointKey = datastore.NewKey("checkpoint")
+)
+
+// NewCheckpointStore wraps the given datastore.Datastore with
+// the `das` prefix.
+func NewCheckpointStore(ds datastore.Datastore) datastore.Datastore {
+	return namespace.Wrap(ds, storePrefix)
+
+}
```
adlerjohn, line 20: nit
```suggestion
	return namespace.Wrap(ds, storePrefix)
}
```

das/checkpoint_store.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+var (
+	storePrefix   = datastore.NewKey("das")
+	checkpointKey = datastore.NewKey("checkpoint")
+)
+
+// NewCheckpointStore wraps the given datastore.Datastore with
+// the `das` prefix.
+func NewCheckpointStore(ds datastore.Datastore) datastore.Datastore {
```
adlerjohn, line 17: Hmmm. Is this it best name for this? It's not really making a _new_ store if it wraps a given store.

das/checkpoint_store.go
```diff
@@ -0,0 +1,45 @@
...
+	return namespace.Wrap(ds, storePrefix)
+
+}
+
+// loadCheckpoint loads the DAS checkpoint from disk and returns it.
+// If there is no known checkpoint, it returns height 0.
+func loadCheckpoint(ds datastore.Datastore) (int64, error) {
```
adlerjohn, line 24: This function and the next one assume a prefixed store right? Is there a way to enforce this at the type level? Otherwise it's relying on the library user knowing to use a prefix store.

das/daser.go
```diff
@@ -18,21 +19,37 @@ type DASer struct {
...
+	getter HeaderGetter
+	// checkpoint from disk -- DASStore stores checkpoint w/ DASCheckpoint key
+	// checkpoint = latest successfully DASed header (reference to it, not the header)
```
adlerjohn, line 26: Can this documentation for what a checkpoint is be placed somewhere in `checkpoint_store.go`? Checkpoint appears to be a height, not a reference.

das/daser.go
```diff
@@ -18,21 +19,37 @@ type DASer struct {
...
+		getter:    getter,
+		ds:        ds,
+		sampleDn:  make(chan struct{}),
+		catchUpDn: make(chan struct{}),
 	}
 }
 
 // Start initiates subscription for new ExtendedHeaders and spawns a sampling routine.
-func (d *DASer) Start(context.Context) error {
+func (d *DASer) Start(_ context.Context) error {
```
adlerjohn, line 52:
```suggestion
func (d *DASer) Start(context.Context) error {
```

das/daser.go
```diff
@@ -75,6 +118,58 @@ func (d *DASer) sampling(ctx context.Context, sub header.Subscription) {
...
+		log.Infow("sampling successful", "height", h.Height, "hash", h.Hash(),
+			"square width", len(h.DAH.RowsRoots), "finished (s)", sampleTime.Seconds())
+
+		height = h.Height
+	}
+}
+
+// catchUp starts sampling headers from the last known
+// checkpoint (latest/highest DASed header) `from`, and breaks the loop
+// once network head `to` is reached. (from:to]
```
adlerjohn, line 153: This comment refers to a last known checkpoint, but really it's catching up from `from`, which could be anything.

das/daser.go
```diff
@@ -75,6 +118,58 @@ func (d *DASer) sampling(ctx context.Context, sub header.Subscription) {
...
+func (d *DASer) catchUp(ctx context.Context, from, to int64) {
+	defer close(d.catchUpDn)
+
+	routineStartTime := time.Now()
+	log.Infow("starting sample routine", "from", from, "to", to)
+
+	// start sampling from height at checkpoint+1 since the
+	// checkpoint height has already been successfully DASed
```
adlerjohn, line 161: ditto

das/daser.go
renaynay: By `available`, it means the node has the header.

renaynay: That the network is at least @ that height

das/daser_test.go
renaynay: Yes

das/daser.go
```diff
@@ -42,28 +59,54 @@ func (d *DASer) Start(context.Context) error {
...
+// sample validates availability for each Header received from header subscription.
+func (d *DASer) sample(ctx context.Context, sub header.Subscription, checkpoint int64) {
+	height := checkpoint
+
+	defer func() {
+		// store latest DASed checkpoint to disk
+		// TODO @renaynay: what sample DASes [100:150] and
+		//  stores latest checkpoint to disk as network head (150)
+		// 	but catchUp routine has only sampled from [1:40] so there is a gap
+		//  missing from (40: 100)?
```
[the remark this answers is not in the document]
renaynay: this TODO should ideally be resolved in this PR. I need to think about it a bit more because it might require a more hefty solution that would make this way less light-weight 😞  I need to think about it.

das/checkpoint_store.go
renaynay: I could make a checkpoint store type, yes

renaynay: Actually I don't think this is necessary - will push a better fix for this.

das/daser.go
renaynay: yes

das/daser_test.go
liamsi: image
https://3d.xkcd.com/221/

das/daser.go
liamsi: > By available, it means the node has the header.

That is really confusing in this context. Esp. as you also ensure availability of the data behind the header later as well.

liamsi: Also: 

> That the network is at least @ that height

Looking at how this method is used, the network will always be past that height (as you only call it after pubsub gives you some larger height).

das/checkpoint_store.go
renaynay: renamed to `wrapCheckpointStore`

das/daser.go
renaynay: done

das/daser_test.go
liamsi: maybe replace "random" with "arbitrary" or simply remove the comment and introduce a var `someHeight := 2` or sth

das/daser.go
renaynay: > That is really confusing in this context. Esp. as you also ensure availability of the data behind the header later as well.

Okay this is a good point


> Looking at how this method is used, the network will always be past that height (as you only call it after pubsub gives you some larger height).

Yes this is also true but we need to have stored and processed the EH first to be able to DAS over it since we need the DAH. I can document this.

das/checkpoint_store.go
adlerjohn: Fixed in [ee95275](https://github.com/celestiaorg/celestia-node/pull/473/commits/ee952758c5c23f631937be7326e069d8c834721b)

das/daser.go
renaynay: can you check now?

das/daser.go
renaynay: resolved?

das/daser.go
Wondertan: > suggested in the past to the DASing and the availability check async
> Yeah that would be a cute optimization, but not strictly necessary.

WDYM? Isn't the DAS of height is the same as the availability check? Yes, I read through the thread and I understand that @renaynay meant a different thing, but I am curious how you misunderstood this and what optimization you meant. I believe we have all such cute optimizations already.

adlerjohn: I think I understood it as being able to to DAS on multiple blocks simultaneously without blocking, e.g. during initial sync.

renaynay: Yes. We will be able to DAS on blocks during sync, but we can only DAS on headers that are already known to the node, hence why we need the `HeaderGetter`. I'm not sure where the confusion lies anymore.

renaynay:
I think I will address the first TODO in a separate issue:

`
figure out what to do in the instance that DASer runs and samples new headers [100:150] and DASer is stopped, so sample routine stores latest checkpoint to disk as last DASed height (150) but catchUp routine has only sampled from [1:40] so there is a gap missing from (40: 100)?
 `

das/daser.go
```diff
@@ -75,6 +116,57 @@ func (d *DASer) sampling(ctx context.Context, sub header.Subscription) {
...
+			"square width", len(h.DAH.RowsRoots), "finished (s)", sampleTime.Seconds())
+
+		checkpoint = h.Height
+	}
+}
+
+// catchUp starts a sampling routine for headers starting at the `from`
+// height and exits the loop once `to` is reached. (from:to]
+func (d *DASer) catchUp(ctx context.Context, from, to int64) {
+	defer close(d.catchUpDn)
```
Wondertan, line 152: There can be multiple `catchUp` routines during runtime, so this will panic if e.g. I close my laptop lid twice in a row for more than block-time time. 

I see here two solutions:
* Making a `catchUp` struct with its own lifecycle, so `sample` will be able to create `catchUp` many instances and wait until they are all done on `defer`
* Spawning one only `catchUp` routine that communicates with `sample` by a chan, i.e. `sample` sends to `catchUp` structs so the routine process them in the order.

I believe the second approach is better, as it simplifies parallelization and accounting for in-progress catchups to be provided with DASer API. The parallelization can be done with `catchUpWorker`s controlled by the `catchUp` routine, so once the job is rcvd from the `sampler`, `catchUp` routines equally splits according to then N of workers and pushes to another chan that the workers listen to.
The `catchUp` routine would also control the accounting of heights being DASed or `State()`

das/daser.go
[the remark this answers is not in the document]
Wondertan: @renaynay, we discussed on our call that https://github.com/celestiaorg/celestia-node/issues/180 should fix this TODO entirely, but in fact, it should only prevent re-sampling of already sampled heights and if it's implemented, the issue with the gap will still present.

The naive solution to the gap issue would be to store a checkpoint in a single `catchUp` routine that is described in the comment above. Basing on your example, when the `catchUp` is stopped it stores the height 40, so then DASer will start from 41 and resample [100:150] again, which we would like to avoid and where https://github.com/celestiaorg/celestia-node/issues/180 can help us.

Wondertan (changes requested):
Generally looks good, but there are two issues:
* Stopping logic will panic if there were two or more catchups
* A few changes still needed to be done before #180 becomes helpful

das/doc.go
```diff
@@ -0,0 +1,22 @@
+/*
+Package das contains the most important functionality provided by celestia-node.
```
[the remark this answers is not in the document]
renaynay: `da.Sampler` sounds good to me. We can write up an epic-style issue based on this PR later with all desired refactoring-related changes, including better documentation. WDYT?

das/daser.go
```diff
@@ -75,6 +116,57 @@ func (d *DASer) sampling(ctx context.Context, sub header.Subscription) {
 			continue
 		}
 
+		// If the next header coming through gossipsub is not adjacent
+		// to our last DASed header, kick off routine to DAS all headers
+		// between last DASed header and h. This situation could occur
+		// either on start or due to network latency/disconnection.
+		if h.Height > (checkpoint + 1) {
```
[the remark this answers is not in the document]
renaynay: What do you mean by "our listing does not allow this"?

das/doc.go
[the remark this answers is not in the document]
Wondertan: Good

das/daser.go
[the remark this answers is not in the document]
Wondertan: Oh, "linting", a typo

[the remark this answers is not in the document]
renaynay: resolved?

das/daser.go
renaynay: resolved by https://github.com/celestiaorg/celestia-node/pull/473/commits/55d198fe32faf83acab201a5325ddb15a93ba018 ?

Wondertan (commented):
Ok, so the last two issues are fixed. Here is another review with suggestions, thoughts, and one bug catch.

Also, I think I found a drawback in a solution for the last TODO. Think of a common case:
* A Node is started and needs/starts a catch-up
* It finishes catching up while the `sample` routine is and was going for thousands of heights
* The node is stopped along with the DASser and the _very_ old checkpoint is saved
* On startup, DASer will load this old checkpoint and will start dasing from it

You may think that #180 will fix this and this is true to some extent, but #180 is about the new `ShareAvailable` implementation which stores the fact of availability on disk and to check it we need to do an IO per height, so on long height ranges, this can be an IO issue that does it for no reason. I think we can justify this issue on the case of a gap described originally, but for the common case I mentioned above this is not acceptable IMO. Also, the logs will be misleading and false stating, as it would say catching up over already sampled headers. 

The issue can be fixed by checking: if there were no catchUp routines terminated(meaning we fully caught up on everything), then store the checkpoint not from the `catchUpScheduler`, but the one in `sample` like it was before.

das/daser.go
```diff
@@ -75,6 +113,102 @@ func (d *DASer) sampling(ctx context.Context, sub header.Subscription) {
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case job := <-jobsCh:
+			// spawn catchUp routine
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
```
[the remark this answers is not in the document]
Wondertan: Might be worth linking this comment to the parallelization issue.

adlerjohn:
CI won't run with merge conflicts IIRC, so could you fix the conflict as you push?

das/daser.go
renaynay: @adlerjohn @liamsi  resolved?

das/daser.go
```diff
@@ -42,28 +58,50 @@ func (d *DASer) Start(context.Context) error {
...
+// sample validates availability for each Header received from header subscription.
+func (d *DASer) sample(ctx context.Context, sub header.Subscription, checkpoint int64) {
+	defer func() {
+		sub.Cancel()
+		close(d.sampleDn)
+	}()
+
+	// kick off catchUpScheduler
+	jobsCh := make(chan *catchUpJob)
+	go d.catchUpScheduler(ctx, jobsCh, checkpoint)
```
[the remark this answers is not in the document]
renaynay: I'll make `jobsCh` a daser field although I don't think it's necessary. I think jobs ch should be created on start and closed out on stop. Not in the constructor.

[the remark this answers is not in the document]
renaynay: Resolved?

das/daser.go
```diff
@@ -75,6 +117,110 @@ func (d *DASer) sampling(ctx context.Context, sub header.Subscription) {
...
+			// header
+			select {
+			case <-ctx.Done():
+				return
+			default:
+				d.jobsCh <- &catchUpJob{
+					from: checkpoint,
+					to:   h.Height - 1,
+				}
+			}
```
Wondertan, line 135: :nooooo:
The write should be a select case, otherwise, the problem I mentioned still might be the case.

das/daser.go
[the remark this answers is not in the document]
Wondertan: My main point was about `catchUpScheduler` being started and controlled by the constructor, so `sample` and `catchUpScheduler` are hierarchically equal.

das/daser.go
```diff
@@ -75,6 +113,102 @@ func (d *DASer) sampling(ctx context.Context, sub header.Subscription) {
...
+		// to our last DASed header, kick off routine to DAS all headers
+		// between last DASed header and h. This situation could occur
+		// either on start or due to network latency/disconnection.
+		if h.Height > checkpoint+1 {
+			// DAS headers between last DASed height up to the current
+			// header
+			jobsCh <- &catchUpJob{
+				from: checkpoint,
+				to:   h.Height - 1,
+			}
```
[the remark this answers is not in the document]
Wondertan: Nope, see https://github.com/celestiaorg/celestia-node/pull/473#discussion_r831334777

das/daser.go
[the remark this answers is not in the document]
Wondertan: Ok

Wondertan (changes requested):

das/daser.go
```diff
@@ -75,6 +117,110 @@ func (d *DASer) sampling(ctx context.Context, sub header.Subscription) {
...
+		// fully caught up to network head, it will resume DASing from this checkpoint
+		// up to current network head
+		// TODO @renaynay: Implement Share Cache #180 to ensure no duplicate DASing over same
+		//  header
+		if err := storeCheckpoint(d.cstore, checkpoint); err != nil {
+			log.Errorw("storing checkpoint to disk", "height", checkpoint, "err", err)
+		}
+		// close out jobs channel
+		close(d.jobsCh)
+		d.jobsCh = nil
```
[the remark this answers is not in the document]
renaynay: Yes thank you!

das/daser.go
```diff
@@ -90,5 +149,91 @@ func (d *DASer) sampling(ctx context.Context, sub header.Subscription) {
...
+		case <-ctx.Done():
+			return
+		case job := <-d.jobsCh:
+			// spawn catchUp routine
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+
+				d.catchUp(ctx, job)
+				// TODO @renaynay: assumption that each subsequent job is to a higher height than the previous.
```
renaynay, line 193: cc @Wondertan

das/daser.go
```diff
@@ -90,5 +149,91 @@ func (d *DASer) sampling(ctx context.Context, sub header.Subscription) {
...
+	// checkpoint height is DASed by broader sample routine
+	for height := job.from + 1; height <= job.to; height++ {
+		h, err := d.getter.GetByHeight(ctx, uint64(height))
+		if err != nil {
+			if err == context.Canceled {
+				return
+			}
+
+			log.Errorw("failed to get next header", "height", height, "err", err)
+			continue // TODO @renaynay: should we really continue in this case?
```
renaynay, line 217: cc @Wondertan

node/services/service.go
```diff
@@ -140,8 +140,14 @@ func ShareService(lc fx.Lifecycle, dag ipld.DAGService, avail share.Availability
 }
 
 // DASer constructs a new Data Availability Sampler.
-func DASer(lc fx.Lifecycle, avail share.Availability, sub header.Subscriber) *das.DASer {
-	das := das.NewDASer(avail, sub)
+func DASer(
```
Wondertan, line 143: Pls put it at baseComponents

Wondertan, line 143: Pls put it at baseComponents

das/daser_test.go
```diff
@@ -4,29 +4,254 @@ import (
...
 	"github.com/celestiaorg/celestia-node/service/share"
 )
 
-func TestDASer(t *testing.T) {
-	shareServ, dah := share.RandLightServiceWithSquare(t, 16)
+// TestDASerLifecycle tests to ensure every mock block is DASed and
+// the DASer checkpoint is updated to network head.
+func TestDASerLifecycle(t *testing.T) {
+	ds := ds_sync.MutexWrap(datastore.NewMapDatastore())
+	cstore := wrapCheckpointStore(ds)
```
Wondertan, line 24: Do you need to wrap it here if its wrapper inside the constructor

Wondertan, line 24: Do you need to wrap it here if its wrapper inside the constructor

das/daser.go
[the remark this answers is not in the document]
renaynay: resolved?

das/daser.go
```diff
@@ -75,6 +117,110 @@ func (d *DASer) sampling(ctx context.Context, sub header.Subscription) {
...
+		// store latest DASed checkpoint to disk here to ensure that if DASer is not yet
+		// fully caught up to network head, it will resume DASing from this checkpoint
+		// up to current network head
+		// TODO @renaynay: Implement Share Cache #180 to ensure no duplicate DASing over same
+		//  header
+		if err := storeCheckpoint(d.cstore, checkpoint); err != nil {
+			log.Errorw("storing checkpoint to disk", "height", checkpoint, "err", err)
+		}
+		// close out jobs channel
+		close(d.jobsCh)
```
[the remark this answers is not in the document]
renaynay: resolved?

das/daser_test.go
renaynay: resolved?

das/daser.go
renaynay: resolved via https://github.com/celestiaorg/celestia-node/pull/473/commits/1ed26f5d47b9495e0f0d650148cae5fb0378b7d7 ?

renaynay:
Tested against devnet 💪🏻

das/daser.go
```diff
@@ -42,28 +62,49 @@ func (d *DASer) Start(context.Context) error {
...
-	defer close(d.done)
+// sample validates availability for each Header received from header subscription.
+func (d *DASer) sample(ctx context.Context, sub header.Subscription, checkpoint int64) {
+	defer func() {
+		sub.Cancel()
+		close(d.sampleDn)
+	}()
+
+	// kick off catchUpScheduler
+	go d.catchUpScheduler(ctx, checkpoint)
```
Wondertan, line 106: Sample does not control the lifecycle of this routine, so pls move it to the constructor

das/daser.go
```diff
@@ -90,5 +149,91 @@ func (d *DASer) sampling(ctx context.Context, sub header.Subscription) {
...
+		case job := <-d.jobsCh:
+			// spawn catchUp routine
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+
+				d.catchUp(ctx, job)
+				// TODO @renaynay: assumption that each subsequent job is to a higher height than the previous.
+				//  I don't see why that is *not* the case, though.
+				checkpoint = job.to
```
Wondertan, line 195: The assumption is correct. The problem is that we cannot guarantee the order in which catchups are finished. Probably the last bug I see in the code. If there are multiple catchUp routines spawned on loc 189. All of them will conflict on writing on this line. This is also racy and if we tested a case with multiple catchUp routines with race detector on, then it will complain.

das/daser.go
Wondertan: If the error happens here then something is really wrong, so I would propose full return here, so node has to be restarted to start catching up here again.

Wondertan (changes requested):

das/daser.go
liamsi: Alternative would be to retry the header a few times before giving up. But could be done in a separate PR.

Wondertan: Although, there is another argument - is pruning from the future. If some headers were pruned, the daser should be ok with the `header.ErrNotFound`, but:
* It should be handled in a specifical case
* We should handle pruning in a way that won't involve ranging over each unavailable header. Meaning that DASer should be somehow aware of pruned headers, so he won't waste any resources on them.

So making a return here as default should be ok for the reason in the comment above.

liamsi: Despite pruning there is the assumption that all headers will be _somehow_ downloadable from the network. Note: this is only relevant for nodes that sync from genesis anyways. I think it is safe to assume that in between the the unbonding period headers are available without any further gymnastics (besides maybe asking a max a hand full of nodes).

das/daser.go
renaynay: https://github.com/celestiaorg/celestia-node/pull/473/commits/a999cf1bf646fdd672732de603ac363256db3af4

Resolved?

das/daser.go
Wondertan: True. Right now, syncing from genesis is the only default, so if there are no headers, something is really wrong and stopping adding is the simplest default we can do tight now.

Regarding asking for headers if nothing is locally. I want this to be a case for our API, so if the pruning is on and someone asks a node for a very old header, it should take if from p2p network and provide to the api.

renaynay: tracked: https://github.com/celestiaorg/celestia-node/issues/547

i implemented a full return for now.
can i resolve this?

das/daser.go
renaynay: resolved?

das/daser.go
renaynay: We could probably solve this by introducing catchUp job IDs. Do you think this needs to be solved in this PR or can we solve it in a PR that implements job IDs + solves this race condition?

das/daser.go
Wondertan: Yes.

Wondertan: But retries here are useless.

Wondertan: See https://github.com/celestiaorg/celestia-node/issues/547#issuecomment-1077661008

das/daser.go
```diff
@@ -90,5 +148,97 @@ func (d *DASer) sampling(ctx context.Context, sub header.Subscription) {
 		sampleTime := time.Since(startTime)
 		log.Infow("sampling successful", "height", h.Height, "hash", h.Hash(),
 			"square width", len(h.DAH.RowsRoots), "finished (s)", sampleTime.Seconds())
+
+		checkpoint = h.Height
```
Wondertan, line 152: nit: I would rename this checkpoint to something like `height` or `sampleHeight` not mix it up with real catching up checkpoint

das/daser.go
```diff
@@ -42,28 +61,49 @@ func (d *DASer) Start(context.Context) error {
...
+	// load latest DASed checkpoint
+	checkpoint, err := loadCheckpoint(d.cstore)
+	if err != nil {
+		return err
+	}
+	log.Infow("loaded checkpoint", "height", checkpoint)
+
+	dasCtx, cancel := context.WithCancel(context.Background())
 	d.cancel = cancel
+	d.sampleDn, d.catchUpDn = make(chan struct{}), make(chan struct{})
```
Wondertan, line 73: nit: you can set those in constructor and there is also no need to set those to nil in Stop. Setting cancel to nil to check if DASer was already started or not is enough. Also, we need to check if DASer was started or not actually, otherwise there are possible issues with those channels and cancel being set to nil.

das/daser.go
```diff
@@ -90,5 +148,97 @@ func (d *DASer) sampling(ctx context.Context, sub header.Subscription) {
...
+	defer func() {
+		// store latest DASed checkpoint to disk here to ensure that if DASer is not yet
+		// fully caught up to network head, it will resume DASing from this checkpoint
+		// up to current network head
+		// TODO @renaynay: Implement Share Cache #180 to ensure no duplicate DASing over same
+		//  header
+		if err := storeCheckpoint(d.cstore, checkpoint); err != nil {
+			log.Errorw("storing checkpoint to disk", "height", checkpoint, "err", err)
+		}
+		// signal that all catchUp routines have finished
```
Wondertan, line 173: doc nit: we now how only one catchUp routine

das/daser.go
```diff
@@ -90,5 +148,97 @@ func (d *DASer) sampling(ctx context.Context, sub header.Subscription) {
...
+		}
+
+		sampleTime := time.Since(startTime)
+		log.Debugw("sampled past header", "height", h.Height, "hash", h.Hash(),
+			"square width", len(h.DAH.RowsRoots), "finished (s)", sampleTime.Seconds())
 	}
+
+	log.Infow("successfully caught up", "from", job.from,
+		"to", job.to, "finished (s)", time.Since(routineStartTime))
+	// report successful result
```
Wondertan, line 242: log nit: a bit of incosistency with the starting log and success log. Should be something like:
`successfully sampled past headers`.

das/daser.go
```diff
@@ -90,5 +148,97 @@ func (d *DASer) sampling(ctx context.Context, sub header.Subscription) {
...
+		err = d.da.SharesAvailable(ctx, h.DAH)
+		if err != nil {
+			if err == context.Canceled {
+				// report previous height as the last successfully sampled height and
+				// error as nil since the routine was ordered to stop
+				return height - 1, nil
+			}
+			log.Errorw("sampling failed", "height", h.Height, "hash", h.Hash(),
+				"square width", len(h.DAH.RowsRoots), "data root", h.DAH.Hash(), "err", err)
+			// continue sampling
```
Wondertan, line 232: For some reason I thought you PR makes the code to return here, especially with warning that tells that you need to restart the node to continue sampling, but the return happens only in case the header is not available. I think it is ok to continue here and all the retries and etc should happen inside SharesAvailable. 

Also BEFP handling should probably happen inside. So like FullAvailability generates it and LighAvailbaility rcvs it and signals to DASer via special error returned from SharesAvailable or something like this. This can also help with case of restarting of LightNode, so once DASer asks LightAvailablity, it will check if there are any known BEFP(either newly rcvs or stored) and return an error on which DASer will terminate.

Wondertan (commented):
No more serious issues I am aware of, only nits left. Almost there.

das/daser.go
renaynay: We do check if DASer was started though by checking to see if `d.cancel != nil` in `Start`

Wondertan: Ok, true, not in this PR.

das/daser.go
renaynay: resolved? https://github.com/celestiaorg/celestia-node/pull/473/commits/ad2975d10455ed92488bdbd81d5c30cf59e3a0bf

das/daser.go
renaynay: resolved? https://github.com/celestiaorg/celestia-node/pull/473/commits/ad2975d10455ed92488bdbd81d5c30cf59e3a0bf

das/daser.go
renaynay: Resolved https://github.com/celestiaorg/celestia-node/pull/473/commits/ad2975d10455ed92488bdbd81d5c30cf59e3a0bf ?

das/daser.go
renaynay: resolved? https://github.com/celestiaorg/celestia-node/pull/473/commits/ad2975d10455ed92488bdbd81d5c30cf59e3a0bf

Wondertan (approved):
:shipit:

liamsi (commented):
What is with the 2 TODOs mentioned in the opening comment?

das/checkpoint_store.go
```diff
@@ -0,0 +1,46 @@
...
+	return int64(binary.BigEndian.Uint64(checkpoint)), err
+}
+
+// storeCheckpoint stores the given DAS checkpoint to disk.
+func storeCheckpoint(ds datastore.Datastore, checkpoint int64) error {
+	buf := make([]byte, 8)
+	binary.BigEndian.PutUint64(buf, uint64(checkpoint))
+
+	return ds.Put(checkpointKey, buf)
+}
```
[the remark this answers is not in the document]
renaynay: Yes

renaynay:
@liamsi the 2 TODOs mentioned initially in the comment of this PR are related to an issue that can only be resolved once https://github.com/celestiaorg/celestia-node/issues/427 is implemented.

liamsi (approved):
I find the naming confusing but the docs clarify it. So LGTM

PR #477, opened 2022-02-23
tests(service/header): fix flakey TestCoreListener

### LINKED ISSUE #453: [Bug]: Flakey TestCoreListener

Wondertan:
One more attempt to close #453

### DIFF
--- a/service/header/core_listener_test.go
+++ b/service/header/core_listener_test.go
@@ -32,19 +32,18 @@ func TestCoreListener(t *testing.T) {
 	fetcher := createCoreFetcher(t)
 
 	// create CoreListener and start listening
-	cl := createCoreListener(t, fetcher, ps0)
+	cl := createCoreListener(ctx, t, fetcher, ps0)
 	err = cl.Start(ctx)
 	require.NoError(t, err)
 
 	// ensure headers are getting broadcasted to the gossipsub topic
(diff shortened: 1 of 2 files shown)


adlerjohn (commented):
Could you explain the change here? It's not obvious why passing in the `ctx` vs using `context.Background()` solves the issue.

renaynay (approved):
yeah I guess let's try this

renaynay:
> Could you explain the change here? It's not obvious why passing in the `ctx` vs using `context.Background()` solves the issue.

@adlerjohn this is because @Wondertan had a suspicion that the code is blocking on an instance of context here and we're not catching it bc we're passing in a new context every time. Regardless, it is cleaner within the scope of a test to pass in the same context.

adlerjohn (approved):

Issue #472, opened 2022-02-22
[Change Request]: remove 1.16 from github actions


liamsi:
### Summary

https://github.com/celestiaorg/celestia-node/blob/bf8c028560a9689ba0f95c726ed280f1cdb49a55/.github/workflows/test.yml#L15

### Gain

Example: After official release of ABC lib, it has proven in many benchmarks to outperform existing XYZ lib by N%

### Evidence

Example: https:/some links

### Implementation ideas

Example: We can start with wrapping existing N code snippet to make implementation easier for feature ABC. 
Still the other option is re-writing this code snippet to achieve neater code design.

### Urgency

Low

rootulp:
We probably also want to bump 1.16 -> 1.17 in:
1. https://github.com/celestiaorg/celestia-node/blame/main/README.md#L8
2. https://github.com/celestiaorg/celestia-node/blob/main/go.mod#L3

PR #479, opened 2022-02-24
chore: remove Go 1.16 from GH actions

rootulp:
Fixes https://github.com/celestiaorg/celestia-node/issues/472

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go: ["1.16", "1.17"]
+        go: ["1.17"]
     steps:
       - uses: actions/checkout@v2
       - name: set up go


liamsi (commented):
Thanks for this! I do think it makes sense to also update the other places you've mentioned in https://github.com/celestiaorg/celestia-node/issues/472#issuecomment-1049584665 but I would leave it up to the core team to review this and decide on this.

renaynay:
We should update go.mod to 1.17 as well as `libs/fslock/lock_unix.go` to remove the `// +build` directive as the new build directive for 1.17 is `go:build`

README.md
```diff
@@ -1,14 +1,16 @@
...
 
 ### Minimum requirements
 
 | Requirement | Notes            |
 |-------------|------------------|
-| Go version  | 1.16 or higher   |
+| Go version  | 1.17 or higher   |
 
 ### Installation
+
```
rootulp, line 12: whitespace changes were caused by my editor auto fixing [markdownlint](https://github.com/markdownlint/markdownlint) issues. Can revert if reviewers request.

go.mod
```diff
@@ -48,6 +48,155 @@ require (
 	go.uber.org/zap v1.21.0
 )
 
+require (
```
rootulp, line 51: changes caused by `go mod tiny` after increasing `go 1.16` to `go 1.17`

libs/fslock/lock_unix.go
```diff
@@ -1,6 +1,3 @@
-//go:build darwin || freebsd || linux
```
renaynay, line 1: we still need this build directive

rootulp: Sorry I blindly removed this b/c I thought both build directives were no longer needed. I looked at the previous commits:
1. https://github.com/rootulp/celestia-node/commit/b15571ca1bfb3264d4e4861a67e48aad68502016
2. https://github.com/rootulp/celestia-node/commit/1d807a4aa171604ba18732a75972ed14d18dcfa8

and learned that `//go:build` is the preferred build directive for [Go 1.17+](https://go.dev/doc/go1.17#language): 

> The go command now understands //go:build lines and prefers them over // +build lines. 

---

Based on [this article](https://www.digitalocean.com/community/tutorials/building-go-applications-for-different-operating-systems-and-architectures) I think the `_unix.go` file suffix isn't sufficient to inform the Go compiler to only build this file for Unix operating systems:

> Because unix is not a valid GOOS, the _unix.go suffix has no meaning to the Go compiler. It does, however, convey the intended purpose of the file. Like the os/path_unix.go file, your path_unix.go file still needs to use build tags, so keep that file unchanged.

so I think you are right and we still need one of these. I will add it back.

renaynay: Thank you ❤️

Wondertan (approved):
Awesome! The v1.17's [lazyloading feature](https://go.dev/ref/mod#lazy-loading) is something I was waiting for a long time in Go and excited to finally have it in this repo. Thank you @rootulp. LGTM.

Wondertan:
@liamsi, as you added the `no_changelog` tag, why do you think change does not deserve a changelog entry, even if it is a small change it has a bigger impact.

adlerjohn:
+1 this is a breaking change and should have a changelog entry

rootulp:
@liamsi added no_changelog when this PR just removed 1.16 from GH actions. Sorry this PR has increased in scope since then so I added a changelog. Please let me know if there's anything else I can do.

Wondertan:
@rootulp, can you pls remove your merge commit? We always rebase and avoid those

renaynay (approved):
Approved, thanks for your contribution @rootulp 

We will squash merge the PR :)

PR #480, opened 2022-02-25
header: skip empty blocks

### LINKED ISSUE #470: service/header: bridge node should short-circuit storing of empty blocks
Currently, we process empty blocks like any other blocks, but we should account for this special by excluding them from being written, again and again. 

Empty blocks are always equal, and there can be many of them through the life of the chain, but there is no point in writing the same data many times and wasting io.

vgonkivs:
Fixes #470

### DIFF
--- a/CHANGELOG-PENDING.md
+++ b/CHANGELOG-PENDING.md
@@ -7,7 +7,7 @@
 ### FEATURES
 
 ### IMPROVEMENTS
-
+- [header: bridge node should short-circuit storing of empty blocks #480](https://github.com/celestiaorg/celestia-node/pull/480) [@vgonkivs](https://github.com/vgonkivs)
 ### BUG FIXES
 
 ### MISCELLANEOUS
--- a/CHANGELOG-PENDING.md
+++ b/CHANGELOG-PENDING.md
@@ -7,7 +7,7 @@
 ### FEATURES
 
 ### IMPROVEMENTS
-- [header: bridge node should short-circuit storing of empty blocks #480](https://github.com/celestiaorg/celestia-node/pull/480) [@vgonkivs](https://github.com/vgonkivs)
+- [header: bridge node avoids duplicate storing of empty blocks #480](https://github.com/celestiaorg/celestia-node/pull/480) [@vgonkivs](https://github.com/vgonkivs)
 ### BUG FIXES
 
 ### MISCELLANEOUS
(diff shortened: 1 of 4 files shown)


service/header/core_listener.go
```diff
@@ -79,6 +79,11 @@ func (cl *CoreListener) listen(ctx context.Context, sub <-chan *types.Block) {
 
 			eh, err := MakeExtendedHeader(ctx, b, comm, vals, cl.dag)
 			if err != nil {
+				if err == errEmptyBlock {
+					log.Debugw("core-listener: ", err)
```
renaynay, line 83: We should probably log more details about the block here (height, time received, etc.)

adlerjohn (commented):
Is this actually needed? Without looking into the implementation, shouldn't `ComputeShares()`, `ipld.PutData()`, and `NewDataAvailabilityHeader()` have some sort of fast-path for empty blocks already?

adlerjohn:
Need a fix to the pending changelog.

renaynay (commented):
👍🏻 

Can you also add a test for this?

vgonkivs:
> Is this actually needed? Without looking into the implementation, shouldn't ComputeShares(), ipld.PutData(), and NewDataAvailabilityHeader() have some sort of fast-path for empty blocks already?

`ipld.PutData()` returns an error for empty block. But do we need an extra call for  `ComputeShares()` and   `ipld.PutData()` if we can rely on txs length?

@adlerjohn

vgonkivs:
Added test.

CHANGELOG-PENDING.md
```diff
@@ -7,7 +7,7 @@
 ### FEATURES
 
 ### IMPROVEMENTS
-
+- [header: bridge node should short-circuit storing of empty blocks #480](https://github.com/celestiaorg/celestia-node/pull/480) [@vgonkivs](https://github.com/vgonkivs)
```
renaynay, line 10:
```suggestion
- [header: bridge node avoids duplicate storing of empty blocks #480](https://github.com/celestiaorg/celestia-node/pull/480) [@vgonkivs](https://github.com/vgonkivs)
```

renaynay (approved):
❤️

adlerjohn (approved):

Wondertan (approved):
Sweet! `b.Txs` is a good solution

PR #491, opened 2022-03-03
core: Fix flakiness in tests using mock core node

renaynay:
Attempt at fixing [1](https://github.com/celestiaorg/celestia-node/runs/5273396388?check_suite_focus=true#step:4:374) and [2](https://github.com/celestiaorg/celestia-node/runs/5378065062?check_suite_focus=true#step:4:420) by waiting until mock core node is running to proceed with test.

### DIFF
--- a/service/header/core_exchange_test.go
+++ b/service/header/core_exchange_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"testing"
+	"time"
 
 	mdutils "github.com/ipfs/go-merkledag/test"
 	"github.com/stretchr/testify/assert"
@@ -13,14 +14,18 @@ import (
 )
 
 func TestCoreExchange_RequestHeaders(t *testing.T) {
-	fetcher := createCoreFetcher(t)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*15)
+	t.Cleanup(cancel)
+
+	fetcher, err := createCoreFetcher(ctx, t)
+	require.NoError(t, err)
 	store := mdutils.Mock()
 
 	// generate 10 blocks
 	generateBlocks(t, fetcher)
 
 	ce := NewCoreExchange(fetcher, store)
-	headers, err := ce.RequestHeaders(context.Background(), 1, 10)
+	headers, err := ce.RequestHeaders(ctx, 1, 10)
 	require.NoError(t, err)
 
 	assert.Equal(t, 10, len(headers))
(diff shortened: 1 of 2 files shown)


adlerjohn (commented):
lgtm, but needs changelog

service/header/core_exchange_test.go
```diff
@@ -13,14 +14,18 @@ import (
...
-	fetcher := createCoreFetcher(t)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*15)
+	t.Cleanup(cancel)
+
+	fetcher, err := createCoreFetcher(ctx, t)
+	require.NoError(t, err)
 	store := mdutils.Mock()
 
 	// generate 10 blocks
 	generateBlocks(t, fetcher)
```
Wondertan, line 25: The referenced failures are on line 24, which is after generateBlocks. The generateBlocks awaits for 10 blocks which also includes waiting for the `IsRunning` to become true, so we were already indirectly waiting for mock to become running and I doubt this will fix issues on line 24, but it does not hurt to have this,

Wondertan (commented):
The change itself is fine but I doubt it will fix anything. Btw, pls try to reproduce this by rerunning the test multiple times with `-count 100 -race`. Also, add to the top of the test the line `runtime.GOMAXPROCS(1)` which allows simulation of a single-threaded environment like on CI

service/header/core_exchange_test.go
renaynay: Very weird then that this happens. It's likely a problem with the mock core node itself then?

Wondertan: > It's likely a problem with the mock core node itself then?

I think so

Wondertan: @renaynay, we can either debug the issue for them and contribute the fix or just embrace and live with it

renaynay: 😞  I don't want to embrace and live with a flakey ci pipeline. Let's at least open an issue and then see if they address it.

Wondertan: Agreed. One thing though. They might not have it at all rn btw. Their code is constantly changing and we are not updating it at all.

evan-forbes: > P.S. If Only Tendermint Was A Library (IOTWAL)

lol IOTWAL

We won't be able to update to the latest version of tendermint until at least after testnet, and even then we will have to wait until the cosmos-sdk uses some version of v0.36.x

I am planning on working on a better way to test the application and the rpc for it this week. We need something that doesn't require running a full tendermint node in order for the CI to not be so dang flaky. 

I'm out of the loop on this, do we have a test suite that runs multiple tests around a single tendermint node, or do we boot one up for each test?

Wondertan:
Closing. The current plan is to update to the latest tendermint and see how this changes the flakeyness story. Also, the change does fixes anything as per info above

Issue #349, opened 2022-01-13
cmd: Allow users to set network type on init


Wondertan:

0xJepsen:
Does this still need to be done? What are the available network types we want to allow users to set? This logic is in the cmd package right?

PR #494, opened 2022-03-03
chore(node): always use `param` defined bootstrappers and remove Bootstrappers config field

### LINKED ISSUE #488: [BUG] Node doesn't automatically connect to bootstrap peers if it fully synced
The problem is that we don't get bootstrap peers defined `param` pkg in [here](https://github.com/celestiaorg/celestia-node/blob/1fcf0c0bb5d5a4e18b51cf12440ce86a84cf7a72/node/p2p/p2p.go#L79).

### LINKED ISSUE #489: node/p2p: Remove `BootstrapPeers` config field
Bootstrap peers are not something user-configurable, but a network-wide configuration that is now defined in `param` pkg. Originally, this field was created as a quick workaround solution, but now it is not needed anymore

Wondertan:
Currently, we allow users to configure bootstrap peers, while this set of peers is constant and defined by the network infrastructure providers, so this is something that users shouldn't be able to change deliberately. The only case where users would want to change their bootstrappers is when they want to join a different network, but this is something that will be done automatically when we start to support multiple networks (#349). Originally, we had this field to be able to quickly configure bootstrappers, but it was clear then that configuring this is not a good idea. Also, putting wrong values there may break network connectivity so altogether it makes sense to remove this field.

> This is what I decide to do together while fixing #488. Something simple to get into working mode again from a mild depressive state.

Closes #488, closes #489

### DIFF
--- a/node/p2p/routing.go
+++ b/node/p2p/routing.go
@@ -25,14 +25,11 @@ func ContentRouting() routing.ContentRouting {
 // Basically, this provides a way to discover peer addresses by respecting public keys.
 func PeerRouting(cfg Config) func(routingParams) (routing.PeerRouting, error) {
 	return func(params routingParams) (routing.PeerRouting, error) {
-		bpeers, err := cfg.bootstrapPeers()
-		if err != nil {
-			return nil, err
-		}
-
+		bpeers := nparams.BootstrappersInfos()
 		prefix := fmt.Sprintf("/celestia/%s", nparams.GetNetwork())
 		mode := dht.ModeAuto
 		if cfg.Bootstrapper {
+			bpeers = nil // no bootstrappers for a bootstrapper ¯\_(ツ)_/¯
 			mode = dht.ModeServer
 		}
 
--- a/node/p2p/routing.go
+++ b/node/p2p/routing.go
@@ -25,27 +25,27 @@ func ContentRouting() routing.ContentRouting {
 // Basically, this provides a way to discover peer addresses by respecting public keys.
 func PeerRouting(cfg Config) func(routingParams) (routing.PeerRouting, error) {
 	return func(params routingParams) (routing.PeerRouting, error) {
-		bpeers := nparams.BootstrappersInfos()
-		prefix := fmt.Sprintf("/celestia/%s", nparams.GetNetwork())
-		mode := dht.ModeAuto
-		if cfg.Bootstrapper {
-			bpeers = nil // no bootstrappers for a bootstrapper ¯\_(ツ)_/¯
-			mode = dht.ModeServer
-		}
-
-		d, err := dht.New(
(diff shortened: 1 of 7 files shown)


node/p2p/routing.go
```diff
@@ -25,14 +25,11 @@ func ContentRouting() routing.ContentRouting {
...
-		bpeers, err := cfg.bootstrapPeers()
-		if err != nil {
-			return nil, err
-		}
-
+		bpeers := nparams.BootstrappersInfos()
 		prefix := fmt.Sprintf("/celestia/%s", nparams.GetNetwork())
 		mode := dht.ModeAuto
 		if cfg.Bootstrapper {
+			bpeers = nil // no bootstrappers for a bootstrapper ¯\_(ツ)_/¯
```
adlerjohn, line 32: 1. lol at the comment
2. Is it okay to `nil`ify `bpeers` here, instead of handling it earlier in the function? Maybe a nit but it seems a bit weird to set it _then_ clear it.

renaynay, line 32:
```suggestion
		prefix := fmt.Sprintf("/celestia/%s", nparams.GetNetwork())
		mode := dht.ModeAuto
		if !cfg.Bootstrapper {
				bpeers := nparams.BootstrappersInfos()
```

renaynay: there will be formatting issues w my above suggestion ^

renaynay (commented):
I think everything looks good other than nit comment

renaynay:
needs rebase

node/p2p/routing.go
Wondertan: True, Rene's suggestion should fix this

Wondertan: I had another look at it after some time and made it even cleaner

Wondertan:
Rebased

adlerjohn (approved):
utACK

node/p2p/routing.go
```diff
@@ -25,30 +25,27 @@ func ContentRouting() routing.ContentRouting {
...
 			dht.DisableProviders(),
-		)
+		}
+
+		if cfg.Bootstrapper {
+			// override options for bootstrapper
+			opts = append(opts,
+				dht.Mode(dht.ModeServer), // it must accept incoming connections
+				dht.BootstrapPeers(),     // no bootstrappers for a bootstrapper ¯\_(ツ)_/¯
+			)
```
liamsi, line 45: Is the comment still accurate? 
dht.BootstrapPeers() == no bootstrappers ?

Wondertan: Yes, I am basically overriding the options

liamsi (approved):

renaynay (approved):
also utACK but lgtm

PR #506, opened 2022-03-08
docs(adr): Public API

Wondertan:
image

### TODO
- [ ] Include API and Node construction
- [ ] Document each method
- [ ] Add method params
~~- [ ] Consquances~~


docs/adr/adr-007-public-api.md
```diff
@@ -0,0 +1,194 @@
...
+
+##### State
+```go
+type StateModule interface {
+    Balance
+    BalanceOf
+    Submit
+    SubmitFrom
+
+    // TODO: Feels like something is missing here
```
renaynay, line 154: `BuildTx` || `BuildMsg`
`SubmitSync` || `SubmitAsync` 

wdyt?

renaynay (commented):
I like the idea of modules vs services and everything LGTM at first glance. Made some comments.

docs/adr/adr-007-public-api.md
```diff
@@ -0,0 +1,194 @@
...
+## Authors
+@Wondertan
+
+## Changelog
+
+- 2022-03-08: initial version
+
+## Context
+
+Celestia Node is being built for almost a half of a year with the bottom-up approach to development. The core lower
```
liamsi, line 12:
```suggestion
Celestia Node using a bottom-up approach. The core lower level components are built first and public API around
them is getting organically shaped.
```

docs/adr/adr-007-public-api.md
```diff
@@ -0,0 +1,194 @@
...
+@Wondertan
+
+## Changelog
+
+- 2022-03-08: initial version
+
+## Context
+
+Celestia Node is being built for almost a half of a year with the bottom-up approach to development. The core lower 
+level components are built first and public API around them is getting organically shaped. The project has raised some
```
liamsi, line 13:
```suggestion
The project's lowlevel components are maturing more and more and now is the right time to start defining a first public API.
```

docs/adr/adr-007-public-api.md
```diff
@@ -0,0 +1,194 @@
...
+
+## Changelog
+
+- 2022-03-08: initial version
+
+## Context
+
+Celestia Node is being built for almost a half of a year with the bottom-up approach to development. The core lower 
+level components are built first and public API around them is getting organically shaped. The project has raised some 
+meat and the time to define the v0 API has come.
```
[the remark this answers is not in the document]
liamsi: I think Hlib wants to say that it is maturing more and more. See the two suggestions above.
```suggestion
```

docs/adr/adr-007-public-api.md
```diff
@@ -0,0 +1,194 @@
...
+
+## Alternative Approaches
+
+### Node Type centric Design
+Another discussed approach to defining an API could be an onion style, where each new layer is a feature set of bigger TODO Needs a better word
```
liamsi, line 19: TODO Needs a better word ?

docs/adr/adr-007-public-api.md
```diff
@@ -0,0 +1,194 @@
...
+## Design
+
+### Goals
+* Ergonomic. Simple, idiomatic and self-explanatory.
+* Module centric(modular). The API is not monolithic and segregated into different categorized and independent modules.
+* Unified. All the node types implement the same set of APIs. The difference is defined by different implementations of
+some modules to meet resource requirements of a type. Example: FullAvailability and LightAvailability
+* Embeddable. Simply constructable Node with library style API. Not an SDK/Framework which dictates users the way to build
+an app, but users are those who decide how they want to build the app using the API.
+* Language agnostic. It should be simple to implement similar module interfaces/traits in other languages over RPC clients
```
liamsi, line 32:
```suggestion
* Language agnostic. It should be simple to implement similar module interfaces/traits in other languages over RPC clients.
```

docs/adr/adr-007-public-api.md
```diff
@@ -0,0 +1,194 @@
...
+
+### Details
+#### Services Deprecation
+The initial step is to deprecate and remove our services in code and stop relying on services as the notion. Instead, 
+the Module should be used, e.g. HeaderService -> HeaderModule.
+
+* We're organically moving towards the direction of modularized libraries. That is, our `share`, `header` and `state`
+packages are getting shaped as independent sovereign modules which now lack their own API definition. 
+* Consistency. Semantically, Modules are closer to what high level Celestia project goals and slogans.
```
liamsi, line 41:
```suggestion
* Consistency. Semantically, Modules are closer to Celestia's overarching project goals and slogans.
```

liamsi: Though this might be a bit confusing as with modular, we specifically mean different blockchain layers and not modular software necessarily. But modularity is good anyways :)

docs/adr/adr-007-public-api.md
```diff
@@ -0,0 +1,194 @@
...
+##### Header
+```go
+type HeaderModule interface {
+    Head
+    Get
+    GetByHeight
+    GetRangeByHeight
+    Subscribe
+    SyncState
+    SyncWait
```
liamsi, line 56: Are the full method signatures omitted on purpose? Or are all of them additional interfaces again?

docs/adr/adr-007-public-api.md
```diff
@@ -0,0 +1,194 @@
...
+    BitSwapStats
+    BitSwapLedger
+
+    DHTFindPeer
+}
+```
+
+##### Wallet
+```go
+type WalletModule interface {
```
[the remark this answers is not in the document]
liamsi: Yes, please. We do not want to re-implement all of this for now.

docs/adr/adr-007-public-api.md
```diff
@@ -0,0 +1,194 @@
...
+    New
+    Del
+    Has
+    List
+    Sign
+    Export
+    Import
+    // TODO: Mnemonics stuff?
```
liamsi, line 119: Comes for free if we use concrete existing libs.

docs/adr/adr-007-public-api.md
```diff
@@ -0,0 +1,194 @@
...
+### Goals
+* Ergonomic. Simple, idiomatic and self-explanatory.
+* Module centric(modular). The API is not monolithic and segregated into different categorized and independent modules.
+* Unified. All the node types implement the same set of APIs. The difference is defined by different implementations of
+some modules to meet resource requirements of a type. Example: FullAvailability and LightAvailability
+* Embeddable. Simply constructable Node with library style API. Not an SDK/Framework which dictates users the way to build
+an app, but users are those who decide how they want to build the app using the API.
+* Language agnostic. It should be simple to implement similar module interfaces/traits in other languages over RPC clients
+
+### Details
```
renaynay, line 34:
```suggestion
### High level description

All node types in `celestia-node` will be referred to as `data availability nodes (DA nodes)` whose sole purpose is to interact with the `data availability network layer (DA layer)` such that the node contains all functionality necessary to post and retrieve messages from the DA layer.

This means that DA nodes will be able to query for / modify celestia state such that the DA nodes are able to pay for posting their messages on the network. The state-related API will be documented below in detail.

Furthermore, interaction between the celestia consensus network and the celestia data availability network will be the responsibility of the **bridge** DA node type. However, that interaction will not be exposed on a public level (meaning a **bridge** node will not expose the same RPC interface as the celestia-core node to which it is connected). A **bridge** node, for all intents and purposes, will provide the same API as that of a **full** node.

### Details
```

renaynay: cc @liamsi Please read through this and give your thoughts bc it's incredibly important we all align on what a DA node provides in terms of public API.

liamsi: The public API should also enable to retry DASing the data behind headers. Also querying the status of the celestia light node (height etc).

renaynay: > The public API should also enable to retry DASing the data behind headers. Also querying the status of the celestia light node (height etc).

This should be a suggestion on the ADR itself imo since it has more to do with method available

Wondertan: > The public API should also enable to retry DASing the data behind headers.
`ShareModule.IsAvailable`

docs/adr/adr-007-public-api.md
```diff
@@ -0,0 +1,194 @@
...
+    // TODO: Auth
+
+}
+```
+
+##### DAS
+// TODO: Should we rebrand DASer to ShareSync?
+```go
+    State
+    Wait
```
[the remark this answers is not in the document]
renaynay: The more I think about it, the more I am convinced that `IsAvailable()` method belongs inside of `DASModule`. I think what would be a nice under-the-hood implementation of it is that `IsAvailable()` would check first in the [`ShareCache`](#180) whether shares available and if not, trigger a sampling routine on the given height. I think the `SharesModule` should deal more with share-specific methods, like the ones that already exist (`GetShare____` ).

docs/adr/adr-007-public-api.md
```diff
@@ -0,0 +1,194 @@
...
+    SyncState
+    SyncWait
+}
+```
+
+##### Shares
+```go
+type SharesModule interface {
+	GetShare
+	Get
```
renaynay, line 64: What would `Get` do as opposed to `GetShare`?  Would `Get` return all shares from given `Root` ?

If so, i suggest naming scheme like

`shares.Get`
`shares.GetByIndex`
`shares.GetByNamespace`

docs/adr/adr-007-public-api.md
```diff
@@ -0,0 +1,194 @@
...
+    State
+    Wait
+```
+
+##### State
+```go
+type StateModule interface {
+    Balance
+    BalanceOf
+    Submit
```
renaynay, line 151: Since `Submit` already exists in `SharesModule`, maybe this should be renamed 

```suggestion
    SubmitTx
```

docs/adr/adr-007-public-api.md
```diff
@@ -0,0 +1,194 @@
...
+    SubmitFrom
```
renaynay, line 152: Same suggestion here as above^ 

```suggestion
    SubmitTxFrom
```

docs/adr/adr-007-public-api.md
```diff
@@ -0,0 +1,194 @@
...
+```
+
+##### Shares
+```go
+type SharesModule interface {
+	GetShare
+	Get
+	GetByNamespace
+	IsAvailable
+	Submit
```
renaynay, line 67:
```suggestion
	Submit
	BuildPayForData
```

Where would PFD constructor live? Would `Submit` implicitly take care of building the PFD?

renaynay (commented):
another round of comments.

docs/adr/adr-007-public-api.md
Wondertan: why do we need a constructor for a tx? Submit can build and send it on it's own

docs/adr/adr-007-public-api.md
Wondertan: They won't conflict as they are from different interfaces. I would stick to having a consistent naming module for data with Submit.

docs/adr/adr-007-public-api.md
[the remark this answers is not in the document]
Wondertan: ShareCache is a CacheAvailabilty implementation of the Availavility interface which has IsAvailable method that Daser relies on. So Daser depends on this, not provides.

Let's arrange sync discussion about this.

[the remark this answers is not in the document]
Wondertan: Wait is a method that allows users to block until Dasing is finished

docs/adr/adr-007-public-api.md
Wondertan: If this is something that geth does, let's discuss if this is suitable for our case

docs/adr/adr-007-public-api.md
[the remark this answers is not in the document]
renaynay: I would rename `Wait` then and have two methods

`IsAvailable()`
`WaitTilAvailable()` or something like that. 

`Wait` name makes me think that admin can pause `daser` routines.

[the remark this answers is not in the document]
renaynay: Update: 

@Wondertan and I had a call regarding where `IsAvailable` should live, and at the moment, let's just keep it inside `SharesModule` and if we find that people are using the `DASModule` more than expected, we can just provide the method there too as it will just use `Availability` implementation under the hood.

docs/adr/adr-007-public-api.md
renaynay: or `shares.GetByCoordinates` instead of `GetByIndex`

Wondertan:
SharesModule is going to be renamed to DataModule as per discussion with @renaynay

docs/adr/adr-007-public-api.md
renaynay: Update: spoke with @Wondertan and we decided that we will not expose `BuildPayForData` at the moment and just keep it hidden behind `Submit` but will add optionality to submit from different account.

Wondertan:
Converting to Draft as WIP

docs/adr/adr-009-public-api.md
```diff
@@ -0,0 +1,214 @@
...
+    New
+    Del
+    Has
+    List
+    Sign
+    Export
+    Import
+    // TODO: Mnemonics stuff?
+}
+```
```
renaynay, line 140: I want to move the wallet module to _nice to have_ because we have the `cel-key` utility which is good ~enough~ for now.

docs/adr/adr-009-public-api.md
```diff
@@ -0,0 +1,214 @@
...
+
+	Start(ctx context.Context) error
+	Stop(ctx context.Context) error
+
+    // TODO: Auth
+}
+```
+
+##### DAS
+// TODO: Should we rebrand DASer to ShareSync?
```
renaynay, line 159: IMO no. We need to retain the key word "sample" bc sharesync implies a full sync of all shares

renaynay:
Do we still want to do this @Wondertan ? 
https://github.com/celestiaorg/celestia-node/pull/506#issuecomment-1103839573

docs/adr/adr-009-public-api.md
```diff
@@ -0,0 +1,228 @@
...
+##### Shares
+```go
+type SharesModule interface {
+	// GetShare returns the Share from the given data Root at the given row/col coordinates.
+	GetShare(ctx context.Context, dah *Root, row, col int) (Share, error)
+	// GetSharesByNamespace returns all shares of the given nID from the given data Root.
+    GetSharesByNamespace(ctx context.Context, root *Root, nID namespace.ID) ([]Share, error)
+    // SharesAvailable subjectively validates if Shares committed to the given data Root are 
+	// available on the network. 
+    SharesAvailable(ctx context.Context, root *Root) error
```
rootulp, line 88: [optional][nit] prefixing this method with an `is` may make it clearer to readers that this method has boolean output.

Proposals: `isEveryShareAvailable`. The 99.9% is a placeholder

```suggestion
	// isEveryShareAvailable checks if the Shares committed to in the given data
	// Root are available on the network. Returns nil if higher than 99.9%
	// confidence the shares are available, otherwise returns an error.
	isEveryShareAvailable(ctx context.Context, root *Root) error
```

Ref: https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig

Update: on second thought the return value of this function is `error` so an `is` prefix may be misleading. If the `error` doesn't need to be propagated to callers, we could return `boolean` for clarity. If the error does need to be propagated, we could return `(boolean, error)`. I don't have strong opinions here.

docs/adr/adr-009-public-api.md
```diff
@@ -0,0 +1,228 @@
...
+    // ProbabilityOfAvailability calculates the probability of the data square
+    // being available based on the number of samples collected.
+    ProbabilityOfAvailability() float64
```
rootulp, line 91: [question] which data square is this method using if it doesn't accept any arguments?

docs/adr/adr-009-public-api.md
```diff
@@ -0,0 +1,228 @@
...
+    // and verifies it against the corresponding block's AppHash.
+    Balance(ctx context.Context) (*state.Balance, error)
+    // BalanceForAddress retrieves the Celestia coin balance for the given address and verifies
+    // the returned balance against the corresponding block's AppHash.
+    BalanceForAddress(ctx context.Context, addr state.Address) (*state.Balance, error)
+    // SubmitTx submits the given transaction/message to the Celestia network and blocks until 
+	// the tx is included in a block.
+    SubmitTx(ctx context.Context, tx state.Tx) (*state.TxResponse, error)
+    // SubmitPayForData builds, signs and submits a PayForData transaction.
+    SubmitPayForData(ctx context.Context, nID namespace.ID, data []byte, gasLim uint64) (*state.TxResponse, error)
```
rootulp, line 185: [nit] for consistency with the argument in the the `Transfer()` function below

```suggestion
    SubmitPayForData(ctx context.Context, nID namespace.ID, data []byte, gasLimit uint64) (*state.TxResponse, error)
```

rootulp (commented):
All of my feedback is optional. 

[nit] I think indentation for codeblocks has a mixture of tabs / spaces which causes it to render incorrectly in GH UI.

docs/adr/adr-009-public-api.md
```diff
@@ -0,0 +1,227 @@
...
+    // and verifies it against the corresponding block's AppHash.
+    Balance(ctx context.Context) (*state.Balance, error)
+    // BalanceForAddress retrieves the Celestia coin balance for the given address and verifies
+    // the returned balance against the corresponding block's AppHash.
+    BalanceForAddress(ctx context.Context, addr state.Address) (*state.Balance, error)
+    // SubmitTx submits the given transaction/message to the Celestia network and blocks until 
+	// the tx is included in a block.
+    SubmitTx(ctx context.Context, tx state.Tx) (*state.TxResponse, error)
+    // SubmitPayForData builds, signs and submits a PayForData transaction.
+    SubmitPayForData(ctx context.Context, nID namespace.ID, data []byte, gasLimit uint64) (*state.TxResponse, error)
```
tzdybal, line 184: I think it will be good to let users use different key management. 
This will require to expose a method to query account numbers or build entire PFD without signing.

docs/adr/adr-009-public-api.md
```diff
@@ -0,0 +1,227 @@
...
+    // Transfer sends the given amount of coins from default wallet of the node to the given account address.
+    Transfer(ctx context.Context, to types.Address, amount types.Int, gasLimit uint64) (*state.TxResponse, error)
```
tzdybal, line 186: `Transfer` is related to `State` but it's also feels like a `wallet` operation.

docs/adr/adr-009-public-api.md
tzdybal: This separation is good.

Also, "key utility" is IMHO way better name than "WalletModule" (because wallets typically shows your bags, let you see history, do transfers, etc).

tzdybal (commented):
CNWBAL (Celestia Node Will Be A Library)

docs/adr/adr-009-public-api.md
renaynay: I guess we can expose an additional `BuildPayForData` and they're responsible for signing it and submitting it via `SubmitTx`

docs/adr/adr-009-public-api.md
renaynay: We probably won't have a wallet module for mainnet.

renaynay: We will have the key utility instead 👍🏻

docs/adr/adr-009-public-api.md
```diff
@@ -0,0 +1,227 @@
...
+
+#### Modules Definition
+
+##### Header
+```go
+type HeaderModule interface {
+	// Head returns the node's local head (tip of the chain of the header store).
+	Head(ctx context.Context) (*header.ExtendedHeader, error)
+	// Get returns the header of the given hash from the node's header store.
+	Get(ctx context.Context, hash tmbytes.HexBytes) (*header.ExtendedHeader, error)
```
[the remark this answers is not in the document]
Wondertan: We could. If we want to go into the direction of general purpose Header lib, we could also consider a general purpose hash definition

docs/adr/adr-009-public-api.md
```diff
@@ -209,6 +209,15 @@ type FraudModule interface {
 }
 ```
 
+##### Metrics
+
+```go
+type MetricsModule interface {
+ // TODO	
+}
+```
```
Wondertan, line 222: Can you elaborate on this module pls? I don't really get its purpose and don't see how its API would look like

Wondertan:
@renaynay, let's better keep MetricsModule so that we have 
```go
type MetricsModule interface {
	// List shows all the registered meters.
	List(ctx) (string[], error)
	
	// Enable turns on the specific meter.
	Enable(string) error
	Disable(string) error
	
	// ExportTo  sets the endpoint the metrics should be exported to
	ExportTo(string)
}
```

This has better ergonomics that making each module have their own Enable/DisableMetrics and also feature like ExportTo can work only globally and not for a specific module

docs/adr/adr-009-public-api.md
Wondertan: So now it's much clear what it should be doing :)

docs/adr/adr-009-public-api.md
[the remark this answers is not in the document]
renaynay: We could use plain []byte but `HexBytes` provides some nice convenience methods (e.g. `String()`, `MarshalJSON()`) that could be useful. We can also alias it? Just something to consider.

Wondertan:
`AuthModule`, which could be part of `NodeModule`

Wondertan:
We need to ensure that all the subscriptions we provide return a channel to be compatible with RPC.
cc @renaynay

docs/adr/adr-009-public-api.md
renaynay: @tzdybal is this separation something you need before mainnet? Otherwise I'll stick it in "nice to have post mainnet"

tzdybal: It's definitely not critical before mainnet, not a blocker.

Wondertan: @renaynay, we should introduce like a `SmthOptions` structure and put it as the last param asap. Then we can add more optional fields without breaking the interface. 

The question I still have here is: should we make such optional fields structure per module or per method? Per module seems fine if modules are granular enough.

Wondertan: Another alternative is to add methods for each optional use case, but I don't really like this, as it bloats the interface.

docs/adr/adr-009-public-api.md
```diff
@@ -0,0 +1,376 @@
...
+    AuthModule
+  }
+```
+
+#### AuthModule
+
+```go
+type AuthModule interface {
+    SetModuleAuth(ctx context.Context, module string) error  // TODO
+    UnsetModuleAuth(ctx context.Context, module string) error // TODO
```
Wondertan, line 201: This API limits to only one user who has auth/dawg/privileges, meaning the API doesn't allow us to have multiple authenticated users: CLI, Optimint, Astria. Or they should have shared access token, which is a bad security practice. 

We should be able to generate multiple access tokens through the API and potentially with another permissions dimension. So that the node operator can grant access to various apps.

The minimalistic example of such an API is https://github.com/filecoin-project/lotus/blob/master/api/api_common.go. Find Auth there.

docs/adr/adr-009-public-api.md
```diff
@@ -0,0 +1,376 @@
...
+  }
+```
+
+#### Fraud
+
+```go
+  type FraudModule interface {
+    Subscribe(type)
+    Get(type)
+    List(type)
```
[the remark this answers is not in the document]
Wondertan: Fraud Proofs

docs/adr/adr-009-public-api.md
distractedm1nd: It is the theoretical probability for any given data square - for example, it returns `1` on full nodes, and `1 - math.Pow(0.75, float64(DefaultSampleAmount))` on light nodes

docs/adr/adr-009-public-api.md
```diff
@@ -0,0 +1,376 @@
...
+    // State reports the current state of the DASer.
+    State() das.SamplingState
+  }
+```
+
+#### State
+
+```go
+  type StateModule interface {
+    // Balance retrieves the Celestia coin balance for the node's account/signer
```
Bidon15, line 220:
```suggestion
    // AccountAddress retrieves the address of the node's account/signer
    AccountAddress(ctx context.Context) (state.Address, error)
    // Balance retrieves the Celestia coin balance for the node's account/signer
```

Bidon15 (commented):
Nit that eases dev ux comparing to using `cel-key` cli tooling

docs/adr/adr-009-public-api.md
```diff
@@ -0,0 +1,376 @@
...
+    QueryRedelegation
+    QueryUnbondingDelegation
+  }
+```
+
+#### Fraud
+
+```go
+  type FraudModule interface {
+    Subscribe(type)
```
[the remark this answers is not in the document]
Wondertan: We might allow Broadcasting of custom FPs eventually. This is a discussion for broader audience. For now let's just Subscribe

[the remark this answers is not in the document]
renaynay: Not broadcaster as that shouldn't be accessible IMO outside of the component that is in charge of constructing + broadcasting proof.

docs/adr/adr-009-public-api.md
```diff
@@ -0,0 +1,376 @@
...
+    // ProbabilityOfAvailability calculates the probability of the data square
+    // being available based on the number of samples collected.
+    ProbabilityOfAvailability() float64
+  }
+```
+
+##### P2P
+
+```go
+  type P2PModule interface {
```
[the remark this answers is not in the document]
renaynay: done

docs/adr/adr-009-public-api.md
```diff
@@ -0,0 +1,376 @@
...
+```go
+type AuthModule interface {
+    SetModuleAuth(ctx context.Context, module string) error  // TODO
+    UnsetModuleAuth(ctx context.Context, module string) error // TODO
+}
+```
+
+#### DAS
+
+// TODO: Should we rebrand DASer to ShareSync?
```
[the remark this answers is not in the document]
renaynay: done

docs/adr/adr-009-public-api.md
```diff
@@ -0,0 +1,376 @@
...
+
+```go
+  type DASModule interface {
+    // State reports the current state of the DASer.
+    State() das.SamplingState
```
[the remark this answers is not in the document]
renaynay: done

docs/adr/adr-009-public-api.md
```diff
@@ -0,0 +1,376 @@
...
+
+Ideally all the state modules below should be implemented on top of only
+StateModule, but there is no way we can have an abstract state requesting method,
+yet.
+
+##### Staking
+
+```go
+  type StakingModule interface {
+    Delegate
```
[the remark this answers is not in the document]
renaynay: done

docs/adr/adr-009-public-api.md
renaynay: done

docs/adr/adr-009-public-api.md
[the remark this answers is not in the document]
renaynay: check now

docs/adr/adr-009-public-api.md
```diff
@@ -0,0 +1,446 @@
...
+    // UnblockPeer removes a peer from the set of blocked peers.
+    UnblockPeer(p peer.ID) error
+    // ListBlockedPeers returns a list of blocked peers.
+    ListBlockedPeers() []peer.ID
+
+    // MutualAdd adds a peer to the list of peers who have a bidirectional
+    // peering agreement that they are protected from being trimmed, dropped
+    // or negatively scored.
+    MutualAdd(id peer.ID, tag string)
+    // MutualAdd removes a peer from the list of peers who have a bidirectional
```
renaynay, line 172: MutualRm

renaynay (changes requested):

docs/adr/adr-009-public-api.md
```diff
@@ -0,0 +1,446 @@
...
+```go
+type AuthModule interface {
+    // SetModuleAuth makes the given module endpoints accessible only 
+    // with authentication.
+    SetModuleAuth(ctx context.Context, module string) error 
+    // UnsetModuleAuth makes the given module endpoints accessible without
+    // authentication.
+    UnsetModuleAuth(ctx context.Context, module string) error
+}
+```
```
renaynay, line 243: remove this

docs/adr/adr-009-public-api.md
```diff
@@ -0,0 +1,446 @@
...
+    // LogLevelSet sets the given component log level to the given level.
+    LogLevelSet(ctx context.Context, name, level string) error
+   
+    // Start starts the node.
+    Start(ctx context.Context) error
+    // Stop stops the node.
+    Stop(ctx context.Context) error
+
+    // NodeModule provides AuthModule 
+    AuthModule
```
renaynay, line 228: delete

docs/adr/adr-009-public-api.md
```diff
@@ -0,0 +1,446 @@
...
+
+```go
+  type NodeModule interface {
+    // Type returns the node type.
+    Type() node.Type
+    // Version returns information about the current binary build.
+    Version() string
+   
+    // LogLevelList lists all the log levels per component.
+    LogLevelList(ctx context.Context) (map[string]string, error)
```
renaynay, line 218: we cannot list log level per logger.

docs/adr/adr-009-public-api.md
distractedm1nd: done

distractedm1nd (approved):
merge please

distractedm1nd (dismissed):

docs/adr/adr-009-public-api.md
```diff
@@ -0,0 +1,438 @@
...
+}
+```
+
+##### Shares
+
+```go
+  type SharesModule interface {
+    // GetShare returns the Share from the given data Root at the given row/col
+    // coordinates.
+    GetShare(ctx context.Context, dah *Root, row, col int) (Share, error)
```
MSevey, line 122: nit: since it is `root` in all the other ones. 
```suggestion
    GetShare(ctx context.Context, root *Root, row, col int) (Share, error)
```

docs/adr/adr-009-public-api.md
```diff
@@ -0,0 +1,438 @@
...
+    Info() p2p.Info
+    // Peers returns all peer IDs used across all inner stores.
+    Peers() peer.IDSlice 
+    // PeerInfo returns a small slice of information Peerstore has on the
+    // given peer.
+    PeerInfo(id peer.ID) peer.AddrInfo
+   
+    // Connect ensures there is a connection between this host and the peer with
+    // given peer.ID. 
+    Connect(ctx context.Context, pi peer.AddrInfo) error
```
MSevey, line 153: based on the comment
```suggestion
    Connect(ctx context.Context, pi peer.ID) error
```

docs/adr/adr-009-public-api.md
```diff
@@ -0,0 +1,438 @@
...
+    // SubmitPayForData builds, signs and submits a PayForData transaction.
+    SubmitPayForData(
+      ctx context.Context, 
+      nID namespace.ID, 
+      data []byte, 
+      gasLimit uint64,
+    ) (*state.TxResponse, error)
+    // Transfer sends the given amount of coins from default wallet of the node 
+    // to the given account address.
+    Transfer(
```
MSevey, line 251: [non-blocking]
Do we want a full balance transfer option?  Or have we discussed how someone would do that?

I assume the full cost of this method is amount + gasFee. So it's hard for users to completely empty a wallet. 
But if we can structure it to allow for amount being the full cost, and what is sent to the new address is amount - gasFee, then users can drain a wallet easily. 

Something a simple as a `subGass bool` argument or just a second method.

docs/adr/adr-009-public-api.md
```diff
@@ -0,0 +1,438 @@
...
+  type StakingModule interface {
+    // Delegate sends a user's liquid tokens to a validator for delegation.
+    Delegate(
+        ctx context.Context, 
+        delAddr state.ValAddress, 
+        amount state.Int, 
+        gasLim uint64,
+    ) (*state.TxResponse, error)
+    // BeginRedelegate sends a user's delegated tokens to a new validator for redelegation.
+    BeginRedelegate(
```
MSevey, line 279: nit: why no just `Redelegate`?

MSevey (commented):
LGTM, not sure I have sufficient historical context to fully approve, will leave that to others.

renaynay (approved):

docs/adr/adr-009-public-api.md
renaynay: We keep the naming scheme of celestia-core

renaynay (approved):

distractedm1nd (approved):

docs/adr/adr-009-public-api.md
```diff
@@ -0,0 +1,438 @@
...
+that this approach does not seem to match the reality we have, where each node
+type implements all the possible APIs but with the variable implementations
+matching resource constraints of a type.
+
+## Design
+
+### Goals
+
+- Ergonomic. Simple, idiomatic and self-explanatory.
+- Module centric(modular). The API is not monolithic and is segregated into
```
liamsi, line 35: nit/non-blocking:
```suggestion
- Module centric (modular). The API is not monolithic and is segregated into
```

I see this a lot in other documentation and comments fwiw.

docs/adr/adr-009-public-api.md
```diff
@@ -0,0 +1,438 @@
...
+### Details
+
+#### Services Deprecation
+
+The initial step is to deprecate services in favor of modules. Ex.
+`HeaderService` -> `HeaderModule`.
+
+- We're organically moving towards the direction of modularized libraries. That
+is, our `share`, `header` and `state` packages are getting shaped as independent
+modules which now lack their own API definition.
```
liamsi, line 81: Could be clearer: 
```suggestion
modules which currently lack their own API definition.
```
_now_ could also mean once the ADR is fully implemented.

docs/adr/adr-009-public-api.md
```diff
@@ -0,0 +1,438 @@
...
+// Multiple Subscriptions can be created.
+Subscribe() (Subscription, error)
+// SyncState returns the current state of the header Syncer. 
+SyncState() sync.State
+// SyncWait blocks until the header Syncer is synced to network head. 
+SyncWait(ctx context.Context) error
+// SyncHead provides the Syncer's view of the current network head.
+SyncHead(ctx context.Context) (*header.ExtendedHeader, error)
+}
+```
```
liamsi, line 114: This is not correctly/consistently formatted. Compare `SharesModule` code-block below for instance.

docs/adr/adr-009-public-api.md
```diff
@@ -0,0 +1,438 @@
...
+    // UnblockPeer removes a peer from the set of blocked peers.
+    UnblockPeer(p peer.ID) error
+    // ListBlockedPeers returns a list of blocked peers.
+    ListBlockedPeers() []peer.ID
+
+    // MutualAdd adds a peer to the list of peers who have a bidirectional
+    // peering agreement that they are protected from being trimmed, dropped
+    // or negatively scored.
+    MutualAdd(id peer.ID, tag string)
+    // MutualAdd removes a peer from the list of peers who have a bidirectional
```
liamsi, line 172:
```suggestion
    // MutualRm removes a peer from the list of peers who have a bidirectional
```

docs/adr/adr-009-public-api.md
```diff
@@ -0,0 +1,438 @@
...
+    // peering agreement that they are protected from being trimmed, dropped
+    // or negatively scored, returning a bool representing whether the given
+    // peer is protected or not.
+    MutualRm(id peer.ID, tag string) bool
```
liamsi, line 176: This is confusing: If I want to remove a peer and `MutualRm` returns `true`,  it means the peer was protected? Does `true` mean it does not get removed then?

docs/adr/adr-009-public-api.md
```diff
@@ -0,0 +1,438 @@
...
+```go
+  type MetricsModule interface {
+    // List shows all the registered meters.
+    List(ctx) (string[], error)
+    
+    // Enable turns on the specific meter.
+    Enable(string) error
+    Disable(string) error
+    
+    // ExportTo  sets the endpoint the metrics should be exported to
```
liamsi, line 353:
```suggestion
    // ExportTo sets the endpoint the metrics should be exported to.
```

liamsi (approved):
Left some minor comments. My understanding is that this will be merged and any outstanding comments will be addressed in a followup PR.

PR #507, opened 2022-03-09
service/header/core_listener: Fix bug where old block headers are spammed to headersub by bridge nodes

### LINKED ISSUE #499: service/header/core_listener: Ensure core node connection is not syncing before publishing headers to `headersub`
## Problem
Old block headers (those older than network head) are getting spammed to `headersub`.

image

## Solution

In order to prevent the spamming of old block headers to `headersub`, **Bridge** nodes should ensure that their running celestia-core node connection is **NOT** syncing before publishing new `ExtendedHeaders` to `headersub`. 

This can be done by only launching the `listen` loop after the `core.BlockFetcher` is not syncing. A new method can be implemented for `BlockFetcher` to return whether core connection is syncing or not and the listen loop can wait to launch until it is no longer syncing.

renaynay:
Resolves #499.

### DIFF
--- a/core/fetcher.go
+++ b/core/fetcher.go
@@ -3,6 +3,7 @@ package core
 import (
 	"context"
 	"fmt"
+	"time"
 
 	tmbytes "github.com/tendermint/tendermint/libs/bytes"
 	"github.com/tendermint/tendermint/types"
@@ -190,7 +191,9 @@ func (f *BlockFetcher) IsSyncing(ctx context.Context) (bool, error) {
 }
 
 // WaitFinishSync will block until the Core connection has finished syncing,
-// returning a nil error.
+// returning a nil error. A non-nil error will be returned if the context is
+// canceled or if there is an error fetching syncing status from the Core
+// connection.
 func (f *BlockFetcher) WaitFinishSync(ctx context.Context) error {
 	for {
 		select {
@@ -204,7 +207,8 @@ func (f *BlockFetcher) WaitFinishSync(ctx context.Context) error {
 			if !syncing {
 				return nil
 			}
-			// TODO @renaynay: should I wait a second or two to request again? Doesn't make sense to keep requesting.
+			// wait a second before requesting sync status again
+			time.Sleep(time.Second)
 		}
 	}
 }
--- a/core/fetcher.go
+++ b/core/fetcher.go
@@ -3,7 +3,6 @@ package core
 import (
 	"context"
 	"fmt"
-	"time"
 
 	tmbytes "github.com/tendermint/tendermint/libs/bytes"
 	"github.com/tendermint/tendermint/types"
@@ -188,27 +187,4 @@ func (f *BlockFetcher) IsSyncing(ctx context.Context) (bool, error) {
 		return false, err
 	}
 	return resp.SyncInfo.CatchingUp, nil
-}
-
-// WaitFinishSync will block until the Core connection has finished syncing,
-// returning a nil error. A non-nil error will be returned if the context is
-// canceled or if there is an error fetching syncing status from the Core
-// connection.
-func (f *BlockFetcher) WaitFinishSync(ctx context.Context) error {
-	for {
-		select {
-		case <-ctx.Done():
-			return ctx.Err()
-		default:
-			syncing, err := f.IsSyncing(ctx)
-			if err != nil {
-				return err
-			}
-			if !syncing {
-				return nil
-			}
-			// wait a second before requesting sync status again
-			time.Sleep(time.Second)
-		}
-	}
 }
--- a/core/fetcher.go
+++ b/core/fetcher.go
@@ -177,3 +177,34 @@ func (f *BlockFetcher) UnsubscribeNewBlockEvent(ctx context.Context) error {
 
 	return f.client.Unsubscribe(ctx, newBlockSubscriber, newBlockEventQuery)
 }
+
+// IsSyncing returns the sync status of the Core connection: true for
+// syncing, and false for already caught up. It can also return an error
+// in the case of a failed status request.
+func (f *BlockFetcher) IsSyncing(ctx context.Context) (bool, error) {
+	resp, err := f.client.Status(ctx)
+	if err != nil {
+		return false, err
+	}
+	return resp.SyncInfo.CatchingUp, nil
+}
+
+// WaitFinishSync will block until the Core connection has finished syncing,
+// returning a nil error.
+func (f *BlockFetcher) WaitFinishSync(ctx context.Context) error {
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+			syncing, err := f.IsSyncing(ctx)
+			if err != nil {
+				return err
+			}
+			if !syncing {
+				return nil
+			}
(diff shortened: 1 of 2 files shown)


adlerjohn:
If there are still TODOs, can we convert this to draft?

renaynay:
How do you turn it back to draft mode @adlerjohn ?

adlerjohn:
draft

renaynay:
`TestSyncStartStopLightWithBridge` is flakey on main branch as well.

core/fetcher.go
```diff
@@ -177,3 +177,34 @@ func (f *BlockFetcher) UnsubscribeNewBlockEvent(ctx context.Context) error {
...
+func (f *BlockFetcher) IsSyncing(ctx context.Context) (bool, error) {
+	resp, err := f.client.Status(ctx)
+	if err != nil {
+		return false, err
+	}
+	return resp.SyncInfo.CatchingUp, nil
+}
+
+// WaitFinishSync will block until the Core connection has finished syncing,
+// returning a nil error.
```
adlerjohn, line 193: Can this document in which cases an error will be returned? Just reading the comment it looks like it will never return an error!

core/fetcher.go
```diff
@@ -177,3 +177,34 @@ func (f *BlockFetcher) UnsubscribeNewBlockEvent(ctx context.Context) error {
...
+			return ctx.Err()
+		default:
+			syncing, err := f.IsSyncing(ctx)
+			if err != nil {
+				return err
+			}
+			if !syncing {
+				return nil
+			}
+			// TODO @renaynay: should I wait a second or two to request again? Doesn't make sense to keep requesting.
```
adlerjohn, line 207: Maybe quadratic falloff? :shrug:

core/fetcher.go
```diff
@@ -177,3 +177,34 @@ func (f *BlockFetcher) UnsubscribeNewBlockEvent(ctx context.Context) error {
...
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+			syncing, err := f.IsSyncing(ctx)
+			if err != nil {
+				return err
+			}
+			if !syncing {
+				return nil
+			}
```
adlerjohn, line 206: Wait, what about the case where `err` is `nil` and `syncing` is `true`?

renaynay: If it's syncing and there's no error, we wait a second and request sync state again. This method blocks until the core client is no longer syncing.

core/fetcher.go
renaynay: resolved?

adlerjohn (approved):

service/header/core_listener.go
```diff
@@ -36,34 +36,63 @@ func NewCoreListener(bcast Broadcaster, fetcher *core.BlockFetcher, dag format.D
...
+	case <-ctx.Done():
+		log.Errorw("core-listener: listener loop failed to start", "err", ctx.Err())
+		return
+	default:
+		if err := cl.fetcher.WaitFinishSync(ctx); err != nil {
+			// TODO @renaynay: should this be a fatal error? if the core listener cannot start, then the bridge node is
+			// 	useless as it won't be broadcasting new headers to the network.
+			log.Errorw("core-listener: listener loop failed to start", "err", err)
+		}
+	}
```
Wondertan, line 88: This is the case where you don't need to select over the context as this is being done inside WaitFinishSync

service/header/core_listener.go
```diff
@@ -36,34 +36,63 @@ func NewCoreListener(bcast Broadcaster, fetcher *core.BlockFetcher, dag format.D
...
+		// only attempt to unsubscribe from new block events if currently
+		// subscribed
+		if sub != nil {
+			err := cl.fetcher.UnsubscribeNewBlockEvent(ctx)
+			if err != nil {
+				log.Errorw("core-listener: failed to unsubscribe from new block events", "err", err)
+			}
+		}
+		log.Info("core-listener: listening stopped")
+	}()
```
Wondertan, line 74: Why did you take it out of Stop?

renaynay: Putting it back.

service/header/core_listener.go
```diff
@@ -36,34 +36,63 @@ func NewCoreListener(bcast Broadcaster, fetcher *core.BlockFetcher, dag format.D
...
+			log.Errorw("core-listener: listener loop failed to start", "err", err)
+		}
+	}
+	// all caught up, start broadcasting new headers
+	sub, err := cl.fetcher.SubscribeNewBlockEvent(ctx)
+	if err != nil {
+		log.Errorw("core-listener: failed to subscribe to new block events", "err", err)
+		return
+	}
+
```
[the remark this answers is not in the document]
renaynay: Pls check it now: https://github.com/celestiaorg/celestia-node/pull/507/commits/d56b23295452c69dd15d5b13cf563814f7543735

I think it's fine the way it is

service/header/core_listener.go
renaynay: Resolved? https://github.com/celestiaorg/celestia-node/pull/507/commits/d56b23295452c69dd15d5b13cf563814f7543735

service/header/core_listener.go
renaynay: Resolved?

renaynay:
Core_listener should still listen for new blocks while bridge node syncing and erasure code them but should not broadcast them yet.

service/header/core_listener.go
```diff
@@ -56,14 +51,26 @@ func (cl *CoreListener) Start(ctx context.Context) error {
...
-func (cl *CoreListener) listen(ctx context.Context, sub <-chan *types.Block) {
+func (cl *CoreListener) listen(ctx context.Context) {
 	defer log.Info("core-listener: listening stopped")
+
+	sub, err := cl.fetcher.SubscribeNewBlockEvent(ctx)
+	if err != nil {
+		log.Errorw("core-listener: failed to subscribe to new block events", "err", err)
+		return
+	}
+
```
Wondertan, line 73: why do we need to take it from the Start? I believe it is ok if the Start fails when a subscription cannot be created. The failed Start prevents the node to be started and the bridge node without the working subscription is useless, thus no need to be started.

renaynay: relic from previous design where sub only started after synced. Re-adding to start.

https://github.com/celestiaorg/celestia-node/pull/507/commits/dfbfdf68e8d84bc68a7502704a21bb28c654d64d

resolved?

liamsi (approved):
This is a straightforward change and PR. Thanks!

liamsi (commented):
Only thing that could be improved is related to:

> Core_listener should still listen for new blocks while bridge node syncing and erasure code them but should not broadcast them yet.

Ideally this was captured by a test somehow (unit/swamp).

Wondertan:
We found an issue with the approach there and after discussing plenty of solutions decided to change the solution here. Specifically, we decided to contribute to the PubSub for a specific feature that would allow us to solve the problem: https://github.com/libp2p/go-libp2p-pubsub/pull/481
We won't wait for it to be accepted as we maintain our own PubSub fork due to #306, so I am cherry-picking the solution now and making an alternative PR. If something is change on main pubsub branch, I will do changes here acrrodingly.

Wondertan:
Closed in favour of #549

PR #513, opened 2022-03-10
docs/adr: Update state access ADR to include balance verification

renaynay:
See file change.

### DIFF
--- a/docs/adr/adr-004-state-interaction.md
+++ b/docs/adr/adr-004-state-interaction.md
@@ -63,6 +63,13 @@ Merkle proofs from the celestia-app and verify them against the latest head's `A
 to do this, it would need access to the `header.Store`'s `Head()` method in order to get the latest known header of the 
 node and check its `AppHash`.
 
+### Availability of `StateService` during sync
+The `Syncer` in the `header`  package provides one public method, `Finished()`, that indicates whether the syncer has 
+finished syncing. Introducing the availability of `StateService` would require extending the public API for `Syncer` 
+with an additional method, `NetworkHead()`, in order to be able to fetch *current* state from the network. The `Syncer`
+would then have to be passed to any implementation of `StateService` upon construction and relied on in order to access 
+the network head even if the syncer is still syncing, as the network head is still verified even during sync.
+
 ### 1. Core Implementation of `StateAccessor`: `CoreAccess`
 
 `CoreAccess` will be the direct RPC connection implementation of `StateAccessor`. It will use the [lens implementation of ChainClient](https://github.com/strangelove-ventures/lens/blob/main/client/chain_client.go#L23)
--- a/docs/adr/adr-004-state-interaction.md
+++ b/docs/adr/adr-004-state-interaction.md
@@ -57,6 +57,12 @@ type StateAccessor interface {
 3. **LOCAL**: eventually, **full** nodes will be able to provide state to the network, and can therefore execute and 
 respond to the queries without relaying queries to celestia-core (*to be scoped out and implemented in later iterations*)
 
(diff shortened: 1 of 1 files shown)


docs/adr/adr-004-state-interaction.md
```diff
@@ -57,6 +57,12 @@ type StateAccessor interface {
 3. **LOCAL**: eventually, **full** nodes will be able to provide state to the network, and can therefore execute and 
 respond to the queries without relaying queries to celestia-core (*to be scoped out and implemented in later iterations*)
 
+### Verification of Balances
+In order to check that the balances returned via the `AccountBalance` query are correct, it is necessary to also request
+Merkle proofs from the celestia-app and verify them against the latest head's `AppHash`. In order for the `StateAccessor`
+to do this, it would need access to the `header.Store`'s `Head()` method in order to get the latest known header of the 
+node and check its `AppHash`.
```
Wondertan, line 64: @renaynay, this is doesn't take into account the syncing aspect. During the sync, we should either (1)block until syncing is done or (2)we should rely on the latest header in the pending hash. Note that headers coming to the pending hash are verified non-adjacently following Tendermint consensus rules with code adapted from Tendermint software. Thus, we can trust headers in the pending cache.

That's being said, using only `store.Head` is not reliable for cases when a node is syncing, e.g. you just opened the lid of the laptop and want to send a tx.

renaynay: `StateService` should not be started / made available on the node while node is syncing, along with many other methods in the Public API you've listed so far. The question is, do we pass sync state to every module that relies on being synced to be able to either serve requests / operate? Or should this be done via some sort of service manager on the node that waits for header service to sync and then kicks off the other dependent services? 

WDYT?

renaynay: Decision from @Wondertan: 

Let's have services rely on syncer.NetworkHead() and services can be made available and work off network head instead of waiting to be synced completely

renaynay: Please review [dc0a0ad](https://github.com/celestiaorg/celestia-node/pull/513/commits/dc0a0ad3ea9728fff0473078de57d2b1b0b74bb0) @Wondertan

Wondertan (approved):
🚢

docs/adr/adr-004-state-interaction.md
```diff
@@ -57,6 +57,19 @@ type StateAccessor interface {
 3. **LOCAL**: eventually, **full** nodes will be able to provide state to the network, and can therefore execute and 
 respond to the queries without relaying queries to celestia-core (*to be scoped out and implemented in later iterations*)
 
+### Verification of Balances
+In order to check that the balances returned via the `AccountBalance` query are correct, it is necessary to also request
+Merkle proofs from the celestia-app and verify them against the latest head's `AppHash`. In order for the `StateAccessor`
```
adlerjohn, line 62: What if a new block arrives in between requesting a proof and receiving it? Shouldn't the node verify the proof against _a given_ block? Should clarify this case.

Wondertan: Good point. I think we need to look deeper into the API which requests proof. Maybe it provides a block height/hash so we can verify proofs specifically against the identified header. 

cc @renaynay

Wondertan: @adlerjohn, after looking into this again I think your case is an implementation detail that can be resolved once the above is implemented. I think we can merge without this clarification

adlerjohn (approved):

liamsi (approved):

PR #517, opened 2022-03-17
ipld: implement retry mechanism for dataSqure repairing if unrepairab…

vgonkivs:
…le error is received

### DIFF
--- a/ipld/read.go
+++ b/ipld/read.go
@@ -2,6 +2,7 @@ package ipld
 
 import (
 	"context"
+	"math"
 	"math/rand"
 
 	"github.com/ipfs/go-cid"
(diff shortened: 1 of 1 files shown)


ipld/read.go
```diff
@@ -50,17 +86,9 @@ func RetrieveData(
...
 
-type quadrant struct {
-	isLeftSubtree bool
-	from          int
-	to            int
-	rootCids      []cid.Cid
-}
-
-func pickRandomQuadrant(roots [][]byte) (*quadrant, error) {
+func pickQuadrant(qNumber int, roots [][]byte) (*quadrant, error) {
```
adlerjohn, line 89: 1. Please do this refactoring in a different PR, as it's orthogonal.
2. As for the refactoring, the original `pickRandomQuadrant` is _good_ because it abstracts away specifics about the random range. Note that your refactoring doesn't actually check that the random number generated is between 1 and 4. It happens to work due to some implementation details (e.g. checking specifically odd then >2), but in the future those implementation details may change and cause spooky action at a distance.
3. If you want to be able to pick the specific quadrant, then you should have _both_ `pickQuadrant` and `pickRandomQuadrant`, the latter simply calling the former with a bounded random number.

ipld/read.go
```diff
@@ -69,20 +97,16 @@ func pickRandomQuadrant(roots [][]byte) (*quadrant, error) {
...
 		quadrant.isLeftSubtree = true
 	}
-
 	// define range of shares for sampling
-	if q > 2 {
-		quadrant.from = edsWidth / 2
-		quadrant.to = edsWidth
-	} else {
-		quadrant.to = edsWidth / 2
+	if qNumber > 2 {
```
adlerjohn, line 104: Comment here seems incorrect now with the refactor, or at least imprecise.

ipld/read.go
```diff
@@ -26,21 +26,57 @@ func RetrieveData(
...
+	// Trying one more time to repair dataSquare, using dah.Columns
+	quadrant, err = pickQuadrant(q, dah.ColumnRoots)
+	if err != nil {
+		return nil, err
+	}
+
+	return repairDataSquare(parentCtx, quadrant, dah, dag, dataSquare, codec, false)
+}
+
+type quadrant struct {
```
adlerjohn, line 51: I think idiomatic Go says this should be upper-case? Not sure.
```suggestion
type Quadrant struct {
```

ipld/read.go
```diff
@@ -26,21 +26,57 @@ func RetrieveData(
 	codec rsmt2d.Codec,
 ) (*rsmt2d.ExtendedDataSquare, error) {
 	edsWidth := len(dah.RowsRoots)
-	rowRoots := dah.RowsRoots
 	dataSquare := make([][]byte, edsWidth*edsWidth)
 
-	quadrant, err := pickRandomQuadrant(rowRoots)
+	q := rand.Intn(4) + 1 //nolint:gosec
+	quadrant, err := pickQuadrant(q, dah.RowsRoots)
```
adlerjohn, line 32: Maybe unrelated to this PR, but I see that it looks like you're only sampling from row roots here? This should be changed to sample from both rows and columns to be secure. Otherwise rows will be sampled more than columns.

ipld/read.go
vgonkivs: But in this case current struct will be exported. As it is related only for this package I'm not sure that making in Upper case is a good idea.

ipld/read.go
vgonkivs: can make it within this PR. Randomising between rows/cols will not take much time I guess.

ipld/read.go
vgonkivs: Ok, will revert changes with refactoring

vgonkivs:
moved to draft due to more priority tasks.

Wondertan:
Closing in favour of https://github.com/celestiaorg/celestia-node/pull/623

PR #518, opened 2022-03-19
chore: Remove issue templates.

adlerjohn:
Issue templates don't seem to be used by the active contributors, so removing.


Wondertan (changes requested):
Even if we don't use them, outside contributors use them and have to use them for bug reports, etc.

adlerjohn:
In that case could we instead add an "Other" issue template that's minimal?

Wondertan:
@adlerjohn, what should it look like, wdyt? 

cc @Bidon15

adlerjohn:
It should be set up such that you guys will use it and not have to delete anything from it. Maybe blank if that's what it takes.

Bidon15:
@adlerjohn are you ok with the idea that I can take care of editing existing templates to a better state?

adlerjohn:
> are you ok with the idea that I can take care of editing existing templates to a better state?

Yes

adlerjohn:
Closing in favor of #534

PR #526, opened 2022-03-21
Update docker-build GH action to push sha tagged images

jbowen93:
This PR updates the docker-build GH Action. Notable changes
- Runs on pushes to all branches and tags the image with the short sha of the git commit
- Builds multi platform (amd64 and arm64) images using cross compilation

You can see an example of a successful run [here](https://github.com/jbowen93/celestia-node/actions/runs/2018410000) with the resulting images pushed [here](https://github.com/jbowen93/celestia-node/pkgs/container/celestia-node)

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -45,7 +45,7 @@ jobs:
       - name: Build and push
         uses: docker/build-push-action@v2
         with:
-          platform: linux/amd64, linux/arm64
+          platforms: linux/amd64, linux/arm64
           push: true
-          tags: ghcr.io/celestiaorg/celestia-node:latest
+          tags: ${{ steps.meta.outputs.tags }}
           file: docker/Dockerfile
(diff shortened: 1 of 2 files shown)


renaynay (approved):
uTACK

.github/workflows/docker-build.yml
```diff
@@ -3,9 +3,13 @@ name: "docker-build"
 on:
   push:
     branches:
-      - "main"
+      - "**"
```
adlerjohn, line 6: Instead you could just say 
```yml
on:
  push:
```
and avoid `branches`?

adlerjohn (approved):

Wondertan (approved):
Don't have good enough qualifications here to deeply understand the change, so uTACK

.github/workflows/docker-build.yml
jbowen93: I'm not sure you can based on my reading of (https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-branches-and-tags) and I'd rather not find out and reset the reviews just for the nit.

PR #538, opened 2022-03-22
swamp: remove redunancy in node creation

### LINKED ISSUE #454: swamp: rework node creation agnostic to types
Currently, we have redundancy code sections in node creation by types.
Better to incorporate all redundancy in 1 place and be compliant with DRY

Bidon15:
Removed all `node<T>WithStore` creation functions for each type of nodes. 
In the meantime, I've seen that execution on macOS is taking much more time comparing to linuxOS, so I've bumped the context timeouts 

Resolves #454

### DIFF
--- a/node/tests/swamp/swamp.go
+++ b/node/tests/swamp/swamp.go
@@ -178,76 +178,54 @@ func (s *Swamp) getTrustedHash(ctx context.Context) (string, error) {
 }
 
 // NewBridgeNode creates a new instance of a BridgeNode providing a default config
-// and a mockstore to the NewBridgeNodeWithStore method
+// and a mockstore to the NewNodeWithStore method
 func (s *Swamp) NewBridgeNode(options ...node.Option) *node.Node {
 	cfg := node.DefaultConfig(node.Bridge)
 	store := node.MockStore(s.t, cfg)
 
-	return s.NewBridgeNodeWithStore(store, options...)
+	return s.NewNodeWithStore(node.Bridge, store, options...)
 }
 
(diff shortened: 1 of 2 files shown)


vgonkivs (approved):

node/tests/swamp/swamp.go
```diff
@@ -178,76 +178,54 @@ func (s *Swamp) getTrustedHash(ctx context.Context) (string, error) {
...
-		node.WithHost(s.createPeer(ks)),
-		node.WithTrustedHash(s.trustedHash),
-	)
-
-	node, err := node.New(node.Bridge, store, options...)
-	require.NoError(s.t, err)
-	s.BridgeNodes = append(s.BridgeNodes, node)
-	return node
-}
+func (s *Swamp) NewNodeWithStore(t node.Type, store node.Store, options ...node.Option) *node.Node {
```
liamsi, line 207: Maybe add a godoc comment for `NewNodeWithStore`.

liamsi (approved):
changes make sense and LGTM. Would recommend adding docs for all public methods.

renaynay (commented):
I would make `40*time.Second` a global var and just use that everywhere instead of passing that val directly

renaynay:
Otherwise LGTM

Wondertan (approved):
Bellissimo! 👌🏻

node/tests/swamp/swamp.go
Bidon15: I'll add the godoc in a separate PR with some nits changes to not ask for another round of reviews 🙏

PR #541, opened 2022-03-23
fix(cmd): return back sliced trusted peers flags and fix usage string

Wondertan:
This was introduced in https://github.com/celestiaorg/celestia-node/pull/455 and was broken in https://github.com/celestiaorg/celestia-node/pull/390. Thanks @Bidon15 for finding this and credits to @vgonkivs for original implementation.


Bidon15 (approved):
🚢 :shipit: 🎬

liamsi (approved):
Can stuff like this be found by Tests as well 🤔

Issue #530, opened 2022-03-22
fraud: Implement proto messages


renaynay:
Implement fraud proof proto messages based on the design described in the [BEFP ADR](#492).

Issue #263, opened 2021-12-06
[Feature Request]: implement stubs for BadEncodingFraudProofs


adlerjohn:
### Summary

Tracking issue: #4

Implement stubs to produce and verify BadEncodingFraudProofs.

### Impact on users

BadEncodingFraudProofs are necessary to have secure partial nodes.

### Cost of not doing

No partial nodes ☹️

### Implementation ideas

As the first step towards having BadEncodingFraudProofs, add stubs to the node for functions to produce and verify BadEncodingFraudProofs.

#### Producing BEFPs

**Full node** starts performing `FullAvailability` check.

https://github.com/celestiaorg/celestia-node/blob/40c4c62b7fb0ba8ef26c644c99a741295efd13d8/node/services/service.go#L137

During `FullAvailability`, repairing the square is attempted.

https://github.com/celestiaorg/celestia-node/blob/40c4c62b7fb0ba8ef26c644c99a741295efd13d8/ipld/read.go#L60-L62

The `error` is returned immediately and not processed. `RepairExtendedDataSquare` needs to be changed to check that the error is not `ErrUnrepairableDataSquare` (which indicates a need for more samples, orthogonally), and if not, instead create a BEFP using

1. The returned `ErrByzantineRow` or `ErrByzantineCol`, and,
2. A Merkle proof for shares in the above error that are missing from the error struct.

The second point will require `retrieveShare` to return a Merkle proof along with the share value.

https://github.com/celestiaorg/celestia-node/blob/40c4c62b7fb0ba8ef26c644c99a741295efd13d8/ipld/read.go#L49

A produced BEFP must then be disseminated across the network with some new p2p message (a separate issue).

#### Verifying BEFPs

When receiving a BEFP from the network (a separate issue), it is verified.

### Urgency

Medium

renaynay:
Closing in favour of #528

PR #545, opened 2022-03-23
fraud: Define Bad Encoding Fraud Poof struct. Provide functionality for befp creation/validation.

### LINKED ISSUE #529: fraud: Implement BEFP generation and validation
Implement BEFPs as described in the [BEFP ADR](#492).

- [ ] Implement BEFP struct
- [ ] Construction/generation of a BEFP 
- [ ] Validation of BEFP

vgonkivs:
Fixes #529 and #530


Wondertan:
pls add the kind label

liamsi:
>  pls add the kind label

pls do this: https://github.com/celestiaorg/quantum-gravity-bridge/issues/108

.github/workflows/labels.yml
```diff
@@ -0,0 +1,14 @@
...
+    types: opened
+jobs:
+  label:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: mheap/github-action-required-labels@v1
+        with:
+          mode: minimum
+          count: 1
+          labels: "kind:bug-fix, kind:miscellaneous, kind:breaking, kind:improvement, kind:feature"
```
Bidon15, line 14: new line
otherwise, LGTM

fraud/pb/proof.proto
```diff
@@ -0,0 +1,23 @@
...
+    bytes       Raw   = 2;
+    MerkleProof Proof = 3;
+}
+
+message BadEnconding {
+    uint64         Height   = 1;
+    repeated Share Shares   = 2;
+    uint32         Position = 3;
+    bool           isRow    = 4;
+ }
```
Wondertan, line 23: Add line

fraud/share.go
```diff
@@ -0,0 +1,33 @@
...
+		NamespaceID: s.NamespaceID,
+		Raw:         s.Raw,
+		Proof: &pb.MerkleProof{
+			Start:    int64(s.Proof.Start()),
+			End:      int64(s.Proof.End()),
+			Nodes:    s.Proof.Nodes(),
+			LeafHash: s.Proof.LeafHash(),
+		},
+	}
+	return &share
```
Wondertan, line 32: Any reason not to make it a one-liner?

fraud/bad_encoding_test.go
```diff
@@ -0,0 +1,180 @@
...
+		dataSquare,
+		rsmt2d.NewRSGF8Codec(),
+		tree.Constructor,
+	)
+	require.Error(t, err)
+
+	var errCol *rsmt2d.ErrByzantineCol
+	require.True(t, errors.As(err, &errCol))
+
+	errCol = err.(*rsmt2d.ErrByzantineCol)
```
Wondertan, line 114: You don't to type cast here. The type is already defined at loc 111

fraud/bad_encoding_test.go
```diff
@@ -0,0 +1,180 @@
...
+	fmt.Println(errCol.Shares)
```
Wondertan, line 115: Delete?

fraud/bad_encoding_test.go
Wondertan: Same for other tests.

fraud/bad_encoding_test.go
```diff
@@ -0,0 +1,180 @@
...
+
+	require.Error(t, err)
+	// [2] and [3] will be empty
+	var errRow *rsmt2d.ErrByzantineRow
+	require.True(t, errors.As(err, &errRow))
+
+	errRow = err.(*rsmt2d.ErrByzantineRow)
+	fmt.Println(errRow.Shares)
+
+	p, err := CreateBadEncodingFraudProof(1, uint8(errRow.RowNumber), true, eds1, errRow.Shares)
```
Wondertan, line 73: Shouldn't we use here the eds returned from repair on loc 57?

fraud/pb/proof.proto
Wondertan, line 23: New line?

fraud/bad_encoding.go
```diff
@@ -0,0 +1,179 @@
...
+	return nil
+}
+
+func buildTreeFromDataRoot(tree *ErasuredNamespacedMerkleTree, root [][]byte, index int) error {
+	for idx, data := range root {
+		if data == nil {
+			return errors.New("empty share")
+		}
+		// error handling is not supposed here because ErasuredNamespacedMerkleTree panics in case of any error
+		// TODO: should we rework this mechanism?
```
Wondertan, line 171: There shouldn't be a case where it panics if everything is implemented properly, so I would remove this todo

Wondertan (commented):
Amazing job, @vgonkivs. It's a lot of effort to implement this in this time frame having zero knowledge on `rsmt2d/nmt/ipld/FPs`, etc. 

This is an initial review with mostly nits. Will get to this tomorrow again

.github/workflows/labels.yml
```diff
@@ -0,0 +1,14 @@
+name: Pull Request Required Labels
```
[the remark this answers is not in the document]
liamsi: Not sure, where is this title linter? I wasn't aware we had this in celestia-node tbh.

fraud/bad_encoding.go
```diff
@@ -0,0 +1,179 @@
...
+	shares := make([]*Share, len(errShares))
+
+	for index, share := range errShares {
+		if share != nil {
+			tree = NewErasuredNamespacedMerkleTree(uint64(len(errShares) / 2))
+			data := eds.Row(uint(index))
+			if isRow {
+				data = eds.Col(uint(index))
+			}
+			if err := buildTreeFromDataRoot(&tree, data, index); err != nil {
```
liamsi, line 44: Q: Why is this called `*FromDataRoot`? From what I understand it takes the row or column data (and not a root)? Or am I missing sth obvious?

fraud/bad_encoding.go
liamsi: Does this todo refer to the `if data == nil` block above? If so, then yes you can assume that the EDS will never be just an empty square and hence data won't be `nil`.

fraud/bad_encoding.go
```diff
@@ -0,0 +1,179 @@
...
+	// compare Merkle Roots
+	if bytes.Equal(tree.Root(), merkleRoot) {
+		return errors.New("invalid fraud proof: merkle root matches")
+	}
+
+	return nil
+}
+
+func buildTreeFromDataRoot(tree *ErasuredNamespacedMerkleTree, root [][]byte, index int) error {
+	for idx, data := range root {
```
liamsi, line 166: Similar to [above](https://github.com/celestiaorg/celestia-node/pull/545/files#r843229315): why is the row/Col leaf data called root here?

fraud/nmt.go
```diff
@@ -0,0 +1,79 @@
+// TEMPORARY
+package fraud
```
liamsi, line 2: Remind me why we have to duplicate this code? Can't we just use the implementation in core? Because that `Tree()` method that was added but there was no tagged release yet?

fraud/nmt.go
```diff
@@ -0,0 +1,79 @@
+// TEMPORARY
```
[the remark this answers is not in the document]
liamsi: Remind me why we have to duplicate this code? Can't we just use the implementation in core? Because that Tree() method that was added but there was no tagged release yet?

fraud/bad_encoding.go
```diff
@@ -0,0 +1,179 @@
...
+	// Shares contains all shares from row/col
+	// Shares that did not pass verification in rmst2d will be nil
+	// For non-nil shares MerkleProofs are computed
+	Shares []*Share
+	// Position represents the number of row/col where ErrByzantineRow/ErrByzantineColl occurred
+	Position uint8
+	isRow    bool
+}
+
+func CreateBadEncodingFraudProof(
```
liamsi, line 27: Should have doc comments. Maybe even a runnable [example test](https://go.dev/blog/examples) to explain the usage/flow here better.

fraud/bad_encoding.go
```diff
@@ -0,0 +1,179 @@
...
+			if err := buildTreeFromDataRoot(&tree, data, index); err != nil {
+				continue
+			}
+
+			proof, err := tree.InclusionProof(position)
+			if err != nil {
+				return nil, err
+			}
+			/*
+				NamespaceID sholud be replaced with ParitySharesNamespaceID for
```
liamsi, line 53:
```suggestion
				NamespaceID should be replaced with ParitySharesNamespaceID for
```

fraud/bad_encoding.go
```diff
@@ -0,0 +1,179 @@
...
+				For EDS 4x4 Q0 is [[0;0],[0;1],[1:0],[1;1]].
+				x	x	x	x
+				x	x	x	x
+				x	x	x	x
+				x	x	x	x
+			*/
+			namespaceID := share[:consts.NamespaceSize]
+			if index >= len(errShares)/2 || int(position) >= len(errShares)/2 {
+				namespaceID = consts.ParitySharesNamespaceID
+			}
```
liamsi, line 64: Q: While the `ErasuredNamespacedMerkleTree` does handle that internally, we manually have to do this again here because we need the exact shares that were pushed as leaves into the tree, right?

fraud/bad_encoding.go
```diff
@@ -0,0 +1,179 @@
...
+	badEncodingFraudProof := pb.BadEnconding{
+		Height:   p.BlockHeight,
+		Shares:   shares,
+		Position: uint32(p.Position),
+		IsRow:    p.isRow,
+	}
+	return badEncodingFraudProof.Marshal()
+}
+
+func (p *BadEncodingProof) Validate(header *header.ExtendedHeader) error {
```
liamsi, line 104: Should have doc comment too.

fraud/bad_encoding.go
```diff
@@ -0,0 +1,179 @@
...
+			continue
+		}
+		shares[index] = share.Raw
+		if ok := share.Validate(roots[index]); !ok {
+			return fmt.Errorf("invalid fraud proof: incorrect share received at position %d", index)
+		}
+	}
+
+	// rebuild a row/col
+	codec := rsmt2d.NewRSGF8Codec()
```
liamsi, line 136: If we later switch to leopard, I hope we won't forget to change all the codec instantiations here and in core.

Actually, it makes more sense to use this constructor instead: https://github.com/celestiaorg/celestia-core/blob/cb696bbd70577c08cd8340ca8cecdff54c723388/pkg/consts/consts.go#L75

fraud/bad_encoding_test.go
liamsi: I also think that was the conclusion of a call with @adlerjohn one or two weeks ago 🤔

fraud/nmt.go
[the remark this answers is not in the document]
Wondertan: We can't depend on the latest core rn due to app not being updated with PreProcessProposal or something not implemented

fraud/bad_encoding.go
liamsi: Also, I think it might be better if the buildTree method constructs the tree entirely, i.e. put the `tree = NewErasuredNamespacedMerkleTree(uint64(len(errShares) / 2))` into the buildTree method and return a pointer instead.

fraud/share.go
vgonkivs: Thought it looks better this way

fraud/bad_encoding_test.go
vgonkivs: No, because repairing process can return new eds which is not the same as `eds1`. As Markle proof validation should be done against our bad eds, we should you it for creation also.

.github/workflows/labels.yml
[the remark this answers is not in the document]
SweeXordious: True, my bad.

vgonkivs:
Merkle roots of `bad` eds and `repaired` eds will differ. It means, we cannot prove that Merkle Proof created from `repaired` eds(from corresponded Merkle Root) will be valid for the bad one.

fraud/nmt.go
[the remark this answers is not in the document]
liamsi: I also think we should quickly do a release and depend on that before we merge this PR. Should we set it back to draft and mention the blocker in the opening comment? 
Alternative: continue reviewing and merge this with the code duplication but then we should open an issue to remove this before we merge the PR.

fraud/bad_encoding.go
vgonkivs: Yes

fraud/proof.go
```diff
@@ -0,0 +1,26 @@
...
+type Proof interface {
+	// Type returns the exact type of fraud proof
+	Type() ProofType
+	// Height returns header's height
+	Height() uint64
+	// Validate check the validity of fraud proof.
+	// Validate throws an error if some conditions don't pass and thus fraud proof is not valid
+	Validate(*header.ExtendedHeader) error
+
+	encoding.BinaryMarshaler
```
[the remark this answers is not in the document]
vgonkivs: Added

fraud/pb/proof.proto
vgonkivs: Done

fraud/bad_encoding_test.go
vgonkivs: Done

fraud/bad_encoding_test.go
vgonkivs: Removed

fraud/bad_encoding.go
vgonkivs: Renamed to `buildTreeFromLeaves`

fraud/bad_encoding.go
vgonkivs: Reworked.

fraud/bad_encoding.go
vgonkivs: Removed codec creation at all. Now it is passing as a function argument

fraud/bad_encoding.go
vgonkivs: Removed `todo` comment

ipld/test_helpers.go
```diff
@@ -84,6 +84,23 @@ func RandNamespacedShares(t *testing.T, total int) NamespacedShares {
 	return shares
 }
 
+func GenerateRandEDS(t *testing.T, originalSquareWidth int) *rsmt2d.ExtendedDataSquare {
```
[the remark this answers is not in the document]
vgonkivs: returned back

fraud/bad_encoding_test.go
```diff
@@ -0,0 +1,161 @@
...
+	copy(dataSquare, shares)
+	dataSquare[2] = nil
+	dataSquare[3] = nil
+	dataSquare[8] = nil
+	dataSquare[12] = nil
+	brokenEDS, err := rsmt2d.RepairExtendedDataSquare(
+		attackerEDS.RowRoots(),
+		attackerEDS.ColRoots(),
+		dataSquare,
+		rsmt2d.NewRSGF8Codec(),
```
liamsi, line 65: IMO, we should later test this also with the leopard codec (or we should use the constructor from core everywhere; so the tests will work with whatever library we end up using for launch).

fraud/bad_encoding.go
```diff
@@ -0,0 +1,219 @@
...
+	if err := in.Unmarshal(data); err != nil {
+		return err
+	}
+	befp := UnmarshalBefp(&in)
+	*p = *befp
+
+	return nil
+}
+
+func UnmarshalBefp(befp *pb.BadEnconding) *BadEncodingProof {
```
renaynay, line 121:
```suggestion
func UnmarshalBEFP(befp *pb.BadEnconding) *BadEncodingProof {
```

fraud/bad_encoding.go
```diff
@@ -0,0 +1,219 @@
...
+	for idx, data := range leaves {
+		if bytes.Equal(data, emptyData) {
+			// if share was not repaired, then we should get it through sampling
+			share, err := getShare(ctx, root, idx, len(leaves), dag)
+			if err != nil {
+				return nil, err
+			}
+			data = share[consts.NamespaceSize:]
+		}
+		// Axis is an exteranl shifting(e.g between Rows); Cell - is an internal shifting(e.g inside one col)
```
renaynay, line 206:
```suggestion
		// Axis is an external shifting (e.g between Rows); Cell - is an internal shifting(e.g inside one col)
```

fraud/bad_encoding.go
```diff
@@ -0,0 +1,219 @@
...
+	"github.com/celestiaorg/celestia-node/ipld"
+	"github.com/celestiaorg/celestia-node/ipld/plugin"
+	"github.com/celestiaorg/celestia-node/service/header"
+)
+
+type BadEncodingProof struct {
+	BlockHeight uint64
+	// Shares contains all shares from row/col
+	// Shares that did not pass verification in rmst2d will be nil
+	// For non-nil shares MerkleProofs are computed
```
renaynay, line 25:
```suggestion
	// Shares contains all shares from rows and columns.
	// Shares that did not pass verification in rmst2d will be nil.
	// For non-nil Shares, MerkleProofs are computed.
```

fraud/proof.go
```diff
@@ -0,0 +1,34 @@
...
+
+// Proof is a generic interface that will be used for all types of fraud proofs in the network.
+type Proof interface {
+	// Type returns the exact type of fraud proof
+	Type() ProofType
+	// Height returns header's height
+	Height() uint64
+	// Validate check the validity of fraud proof.
+	// Validate throws an error if some conditions don't pass and thus fraud proof is not valid
+	Validate(*header.ExtendedHeader, rsmt2d.Codec) error
```
Wondertan, line 26: Validate must not depend on rsmt2d.Codec. rsmt2d is only part of BEFP. The interface is supposed to be implemented by other frauds

fraud/bad_encoding.go
```diff
@@ -0,0 +1,219 @@
...
+	Shares []*Share
+	// Position represents the number of row/col where ErrByzantineRow/ErrByzantineColl occurred
+	Position uint8
+	isRow    bool
```
renaynay, line 29: Can you document this field too?

fraud/bad_encoding.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+func UnmarshalBefp(befp *pb.BadEnconding) *BadEncodingProof {
+	return &BadEncodingProof{
+		BlockHeight: befp.Height,
+		Shares:      ProtoToShare(befp.Shares),
+	}
+}
+
+func (p *BadEncodingProof) Validate(header *header.ExtendedHeader, codec rsmt2d.Codec) error {
```
Wondertan, line 128: Pls use the codec constructor defined in [core/consts](https://github.com/celestiaorg/celestia-core/blob/master/pkg/consts/consts.go#L75)

Wondertan: And not just here, but every where you create the codec.

Wondertan (changes requested):

renaynay (commented):
Initial review

docs/adr/adr-006-fraud-service.md
```diff
@@ -75,10 +75,10 @@ type BadEncodingProof struct {
...
 
 message MerkleProof {
-  int64          total     = 1;
-  int64          index     = 2;
-  bytes          leaf_hash = 3;
-  repeated bytes aunts     = 4;
+   int64          start     = 1;
+   int64          end       = 2;
+   repeated bytes nodes     = 3;
+   bytes          leaf_hash = 4;
```
adlerjohn, line 81: Fmt. There shouldn't be 3 spaces for indentation here. 2, maybe. 4, also maybe.

fraud/bad_encoding.go
```diff
@@ -0,0 +1,219 @@
...
+	roots [][]byte,
+	errShares [][]byte,
+	dag format.NodeGetter,
+) (Proof, error) {
+	shares := make([]*Share, len(errShares))
+	for index, share := range errShares {
+		if share != nil {
+			data := eds.Row(uint(index))
+			if isRow {
+				data = eds.Col(uint(index))
```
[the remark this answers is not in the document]
adlerjohn: If the row is what's bad, then you look through the completely repaired columns to generate Merkle trees for. This needs to be documented though.

fraud/bad_encoding.go
```diff
@@ -0,0 +1,219 @@
...
+			}
+
+			tree, err := buildTreeFromLeaves(ctx, data, roots[index], index, dag)
+			if err != nil {
+				return nil, err
+			}
```
adlerjohn, line 53: This doesn't look right. It will terminate the loop as soon as the first non-complete col/row is hit. You need to compute trees for _all_ complete col/rows. This should be a `continue`.

fraud/bad_encoding.go
```diff
@@ -0,0 +1,219 @@
...
+			/*
+				NamespaceID should be replaced with ParitySharesNamespaceID for
+				all shares except Q0(all erasure coded shares).
+				For EDS 4x4 Q0 is [[0;0],[0;1],[1:0],[1;1]].
+				x	x	x	x
+				x	x	x	x
+				x	x	x	x
+				x	x	x	x
+			*/
+			namespaceID := share[:consts.NamespaceSize]
```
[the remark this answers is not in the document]
adlerjohn: +1. IIRC you're not supposed to manipulate the NID of the parity shares _within the share data_.

(Just as an aside, but making use of the type system instead of remembering that the NID is here vs there would be better.)

fraud/bad_encoding.go
```diff
@@ -0,0 +1,219 @@
...
+
+			shares[index] = &Share{namespaceID, share, &proof}
+		}
+	}
+
+	return &BadEncodingProof{
+		BlockHeight: height,
+		Shares:      shares,
+		isRow:       isRow,
+		Position:    position,
```
adlerjohn, line 80: Shouldn't this be `index` for consistency.

fraud/bad_encoding.go
```diff
@@ -0,0 +1,219 @@
...
+		return errors.New("invalid fraud proof: invalid extended header")
+	}
+	if len(merkleRowRoots) != len(p.Shares) || len(merkleColRoots) != len(p.Shares) {
+		return errors.New("invalid fraud proof: invalid shares")
+	}
+
+	roots := merkleRowRoots
+	if p.isRow {
+		roots = merkleColRoots
+	}
```
adlerjohn, line 144: Document why

fraud/bad_encoding.go
```diff
@@ -0,0 +1,219 @@
...
+	// rebuild a row/col
+	rebuiltShares, err := codec.Decode(shares)
+	if err != nil {
+		return err
+	}
+	rebuiltExtendedShares, err := codec.Encode(rebuiltShares)
+	if err != nil {
+		return err
+	}
+	rebuiltShares = append(rebuiltShares, rebuiltExtendedShares...)
```
adlerjohn, line 168: Double-check that this works the same with Leopard, IIRC Leopard behaves differently with that it returns, but this might be handled by the wrapper.

fraud/bad_encoding.go
```diff
@@ -0,0 +1,219 @@
...
+
+	tree := wrapper.NewErasuredNamespacedMerkleTree(uint64(len(shares) / 2))
```
adlerjohn, line 170: For consistency, use one of `len(shares)` or `len(roots)` in this function.

fraud/bad_encoding.go
```diff
@@ -0,0 +1,219 @@
...
+	position uint8,
+	isRow bool,
+	eds *rsmt2d.ExtendedDataSquare,
+	roots [][]byte,
+	errShares [][]byte,
+	dag format.NodeGetter,
+) (Proof, error) {
+	shares := make([]*Share, len(errShares))
+	for index, share := range errShares {
+		if share != nil {
```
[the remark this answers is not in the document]
vgonkivs: Yes, but `share == nil` condition looks better. Will change

fraud/pb/proof.proto
```diff
@@ -0,0 +1,23 @@
...
+
+message MerkleProof {
+   int64          start     = 1;
+   int64          end       = 2;
+   repeated bytes nodes     = 3;
+   bytes          leaf_hash = 4;
+}
+
+message Share {
+    bytes NamespaceID = 1;
```
adlerjohn, line 13: fmt

fraud/proof.go
```diff
@@ -0,0 +1,34 @@
...
+	pb "github.com/celestiaorg/celestia-node/fraud/pb"
+	"github.com/celestiaorg/celestia-node/service/header"
+	"github.com/celestiaorg/nmt"
+	"github.com/celestiaorg/rsmt2d"
+)
+
+type ProofType string
+
+const (
+	BadEncoding ProofType = "BadEncoding"
```
adlerjohn, line 15: This doesn't guarantee that all variants are unique. Prefer `iota`?

fraud/bad_encoding.go
```diff
@@ -0,0 +1,219 @@
...
+	root []byte,
+	axis int,
+	dag format.NodeGetter,
+) (*wrapper.ErasuredNamespacedMerkleTree, error) {
+	tree := wrapper.NewErasuredNamespacedMerkleTree(uint64(len(leaves) / 2))
+	emptyData := bytes.Repeat([]byte{0}, len(leaves[0]))
+	for idx, data := range leaves {
+		if bytes.Equal(data, emptyData) {
+			// if share was not repaired, then we should get it through sampling
+			share, err := getShare(ctx, root, idx, len(leaves), dag)
```
adlerjohn, line 200: You shouldn't sample the share here. Rather, by construction of the algorithm, this share must have been gotten through sampling before you even tried to repair the square. You just need to save the Merkle proof when you sampled the share originally.

See https://github.com/celestiaorg/celestia-node/issues/263

> The second point will require `retrieveShare` to return a Merkle proof along with the share value.

https://github.com/celestiaorg/celestia-node/blob/40c4c62b7fb0ba8ef26c644c99a741295efd13d8/ipld/read.go#L49

fraud/bad_encoding.go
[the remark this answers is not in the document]
vgonkivs: No, It's not prepended twice at this point. NID is copying to be able to validate the share correctly, as VerifyInclusion is required correct NID.

fraud/nmt.go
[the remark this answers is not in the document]
liamsi: see: #607

vgonkivs:
should be reviewed after [this PR](https://github.com/celestiaorg/celestia-node/pull/623) will be merged

fraud/nmt.go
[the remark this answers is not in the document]
vgonkivs: File is removed

fraud/bad_encoding.go
vgonkivs: Added comment

fraud/bad_encoding.go
liamsi: IIRC, I've updated leopard/the go-wrapper to behave the same with this regard but would be good to confirm this behaviour.

fraud/proof.go
```diff
@@ -0,0 +1,37 @@
...
+	"encoding"
+	"fmt"
+
+	"github.com/celestiaorg/celestia-node/service/header"
+)
+
+type ProofType int
+
+const (
+	BadEncoding ProofType = iota
```
liamsi, line 13: Is this used anywhere?

vgonkivs: Will be used in FraudService

ipld/read.go
```diff
@@ -2,15 +2,27 @@ package ipld
...
 	"github.com/celestiaorg/nmt"
+
 	"github.com/celestiaorg/nmt/namespace"
 )
 
+type ErrByzantine struct {
+	Index  uint8
+	Shares []*NamespacedShareWithProof
+	IsRow  bool
+}
```
liamsi, line 20: Shouldn't we re-use this struct: https://github.com/celestiaorg/rsmt2d/blob/5dc63b1ad762ffe6d65c2a920e9ea49f1fdff327/extendeddatacrossword.go#L31-L36

?

ipld/share.go
```diff
@@ -47,3 +57,63 @@ func (ns NamespacedShares) Raw() [][]byte {
...
+			shares,
+			&NamespacedShareWithProof{share.NamespaceID, share.Data, &proof},
+		)
+	}
+	return shares
+}
+
+func ProtoToProof(protoProof *pb.MerkleProof) nmt.Proof {
+
+	// p
```
liamsi, line 117: ??

vgonkivs: removed

go.mod
```diff
@@ -255,5 +256,5 @@ replace (
 	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v0.44.2-celestia-rc-1
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
 	github.com/libp2p/go-libp2p-pubsub v0.5.7-0.20211029175501-5c90105738cf => github.com/celestiaorg/go-libp2p-pubsub v0.5.6-celestia
-	github.com/tendermint/tendermint v0.34.14 => github.com/celestiaorg/celestia-core v0.34.15-celestia
+	github.com/tendermint/tendermint v0.34.14 => github.com/celestiaorg/celestia-core v0.34.16-celestia-rc-1
```
liamsi, line 259: Is there a particular reason we aren't using the most recent release here? https://github.com/celestiaorg/celestia-core/releases/tag/v1.1.0-tm-v0.34.16

fraud/bad_encoding.go
```diff
@@ -0,0 +1,155 @@
...
+// rebuilds bad row or col from received shares, computes Merkle Root
+// and compares it with block's Merkle Root
+func (p *BadEncodingProof) Validate(header *header.ExtendedHeader) error {
+	merkleRowRoots := header.DAH.RowsRoots
+	merkleColRoots := header.DAH.ColumnRoots
+	if int(p.Index) >= len(merkleRowRoots) || int(p.Index) >= len(merkleColRoots) {
+		return errors.New("invalid fraud proof: incorrect Index of bad row/col")
+	}
+	if len(merkleRowRoots) != len(merkleColRoots) {
+		return errors.New("invalid fraud proof: invalid extended header")
```
liamsi, line 103: nit: in this case it isn't really the fraud proof that is invalid but the user passed in an invalid header only.

fraud/bad_encoding.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	// comparing rebuilt Merkle Root of bad row/col with respective Merkle Root of row/col from block
+	merkleRoot := merkleColRoots[p.Index]
+	if p.isRow {
+		merkleRoot = merkleRowRoots[p.Index]
+	}
+
+	if bytes.Equal(tree.Root(), merkleRoot) {
+		return errors.New("invalid fraud proof: merkle root matches")
```
liamsi, line 151:
```suggestion
		return errors.New("invalid fraud proof: recomputed Merkle root matches the header's row/column root")
```

fraud/bad_encoding.go
liamsi: resolved?

ipld/read.go
vgonkivs: Well, we can do this. But we still need custom error to save []nmt.Proof of shares. but instead of one slice.` Shares []*NamespacedShareWithProof` we will have 2:
` Shares []byte`
` Proofs []*nmt.Proof`

Not sure, that this solution is better.

go.mod
vgonkivs: This release was added after merging additional nmt method. Will update to [v1.1.0-tm-v0.34.16](https://github.com/celestiaorg/celestia-core/releases/tag/v1.1.0-tm-v0.34.16)

fraud/bad_encoding.go
liamsi: We can do this in followup work. I think we should simply switch to leopard by default in rsmt2d and core. Then, this code will automatically test with leopard once we updated these deps.

liamsi: see: https://github.com/celestiaorg/rsmt2d/issues/74 and https://github.com/celestiaorg/rsmt2d/pull/75

fraud/share.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"crypto/sha256"
+
+	"github.com/celestiaorg/nmt"
+
+	pb "github.com/celestiaorg/celestia-node/fraud/pb"
+)
+
+type Share struct {
```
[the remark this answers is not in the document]
liamsi: now called ShareWithProof

fraud/bad_encoding.go
liamsi: It now says: `// isRow shows that verification failed on row`

fraud/bad_encoding.go
liamsi: resolved?

adlerjohn (commented):
Merge conflict

fraud/bad_encoding.go
liamsi: resolved?

go.sum
```diff
@@ -1377,8 +1377,6 @@ github.com/pascaldekloe/goe v0.1.0 h1:cBOtyMzM9HTpWjXfbbunk26uA6nG3a8n06Wieeh0Mw
 github.com/pascaldekloe/goe v0.1.0/go.mod h1:lzWF7FIEvWOWxwDKqyGYQf6ZUaNfKdP144TG7ZOy1lc=
 github.com/pborman/uuid v0.0.0-20170112150404-1b00554d8222/go.mod h1:VyrYX9gd7irzKovcSS6BIIEwPRkP2Wm2m9ufcdFSJ34=
 github.com/pborman/uuid v1.2.0/go.mod h1:X/NO0urCmaxf9VXbdlT7C2Yzkj2IKimNn4k+gtPdI/k=
-github.com/pborman/uuid v1.2.0/go.mod h1:X/NO0urCmaxf9VXbdlT7C2Yzkj2IKimNn4k+gtPdI/k=
-github.com/pelletier/go-toml v1.2.0/go.mod h1:5z9KED0ma1S8pY6P1sdut58dfprrGBbd/94hg7ilaic=
```
liamsi, line 1381: Only the go.sum seems modified 🤔 but not go.mod? Can you run go mod tidy to make sure this isn't by accident?

go.mod
liamsi: Thanks! I'm surprised that go.sum is different though go.mod stayed the same: https://github.com/celestiaorg/celestia-node/pull/545/files#r861699632

ipld/read.go
```diff
@@ -2,15 +2,27 @@ package ipld
...
 	"github.com/ipfs/go-cid"
 	ipld "github.com/ipfs/go-ipld-format"
 
 	"github.com/celestiaorg/celestia-node/ipld/plugin"
 	"github.com/celestiaorg/nmt"
+
 	"github.com/celestiaorg/nmt/namespace"
 )
 
+type ErrByzantine struct {
```
renaynay, line 16: Can we add doc?

ipld/read.go
```diff
@@ -88,6 +100,74 @@ func GetLeaf(ctx context.Context, dag ipld.NodeGetter, root cid.Cid, leaf, total
 	return GetLeaf(ctx, dag, root, leaf, total)
 }
 
+// GetProvesForShares fetches Merkle Proof for shares range and converts it to NamespacedShareWithProof
+func GetProvesForShares(
```
renaynay, line 104: Can we rename to `GetProofsForShares`

ipld/read.go
```diff
@@ -88,6 +100,74 @@ func GetLeaf(ctx context.Context, dag ipld.NodeGetter, root cid.Cid, leaf, total
 	return GetLeaf(ctx, dag, root, leaf, total)
 }
 
+// GetProvesForShares fetches Merkle Proof for shares range and converts it to NamespacedShareWithProof
```
renaynay, line 103:
```suggestion
// GetProvesForShares fetches Merkle proofs for the given shares and returns the result as an array of NamespacedShareWithProof.
```

ipld/read_test.go
```diff
@@ -441,3 +444,107 @@ func generateRandNamespacedRawData(total, nidSize, leafSize uint32) [][]byte {
...
+	eds := generateRandEDS(t, 2)
+	width := eds.Width()
+	shares := ExtractODSShares(eds)
+
+	in, err := PutData(ctx, shares, dag)
+	require.NoError(t, err)
+
+	// limit with deadline, specifically retrieval
+	_ctx, cancel := context.WithTimeout(ctx, time.Second*2)
+	defer cancel()
```
renaynay, line 507: we usually use `t.Cleanup(cancel)`

ipld/read_test.go
```diff
@@ -441,3 +444,107 @@ func generateRandNamespacedRawData(total, nidSize, leafSize uint32) [][]byte {
...
+	tree := wrapper.NewErasuredNamespacedMerkleTree(uint64(size/2), nmt.NodeVisitor(batchAdder.Visit))
+	attackerEDS, _ := rsmt2d.ImportExtendedDataSquare(shares, consts.DefaultCodec(), tree.Constructor)
+	err := batchAdder.Commit()
+	require.NoError(t, err)
+
+	da := da.NewDataAvailabilityHeader(attackerEDS)
+	r := NewRetriever(dag, consts.DefaultCodec())
+	_, err = r.Retrieve(context.Background(), &da)
+	var errByz *ErrByzantine
+	require.True(t, errors.As(err, &errByz))
```
renaynay, line 549: nit but you can use `require.ErrorAs`

ipld/retriever.go
```diff
@@ -56,13 +58,55 @@ func (r *Retriever) Retrieve(ctx context.Context, dah *da.DataAvailabilityHeader
...
+				return nil, byzErr
+			}
 		}
 		// retry quadrants until we can reconstruct the EDS or error out
 	}
 
 	return nil, format.ErrNotFound
 }
 
+func (r *Retriever) handleByzantineError(
```
renaynay, line 71: I would prefer if this were called `checkForByzantineError`

ipld/retriever.go
```diff
@@ -56,13 +58,55 @@ func (r *Retriever) Retrieve(ctx context.Context, dah *da.DataAvailabilityHeader
...
+	ctx context.Context,
+	dah *da.DataAvailabilityHeader,
+	byzErr error,
+) error {
+	var errRow *rsmt2d.ErrByzantineRow
+	var errCol *rsmt2d.ErrByzantineCol
```
renaynay, line 77:
```suggestion
        var (
           errRow *rsmt2d.ErrByzantineRow
           errCol *rsmt2d.ErrByzantineCol
        )
```

pls lint after applying suggestion

ipld/retriever.go
```diff
@@ -56,13 +58,55 @@ func (r *Retriever) Retrieve(ctx context.Context, dah *da.DataAvailabilityHeader
...
+	if !errors.As(byzErr, &errRow) && !errors.As(byzErr, &errCol) {
+		return nil
+	}
+	var errShares [][]byte
+	var root []byte
+	isRow := false
+	var index uint8
```
renaynay, line 84:
```suggestion
	var (
		errShares [][]byte
		root      []byte
		index     uint8
	)
	isRow := false
```

ipld/retriever.go
renaynay: Also can you add documentation to this method? And a couple comments explaining what you are doing inside so it's easier to follow.

ipld/share.go
```diff
@@ -47,3 +57,59 @@ func (ns NamespacedShares) Raw() [][]byte {
 	}
 	return res
 }
+
+type NamespacedShareWithProof struct {
```
renaynay, line 61: doc pls

ipld/share.go
```diff
@@ -47,3 +57,59 @@ func (ns NamespacedShares) Raw() [][]byte {
...
+	ID namespace.ID
+	Share
+	Proof *nmt.Proof
+}
+
+// NewShareWithProof takes leaf and it's path, starting from the tree root,
+// and computes nmt.Proof for this leaf
```
renaynay, line 68:
```suggestion
// NewShareWithProof takes the given leaf and its path, starting from the tree root,
// and computes the nmt.Proof for it.
```

ipld/share.go
```diff
@@ -47,3 +57,59 @@ func (ns NamespacedShares) Raw() [][]byte {
...
+func NewShareWithProof(index int, leaf []byte, pathToLeaf []cid.Cid) *NamespacedShareWithProof {
+	rangeProofs := make([][]byte, 0)
+	for idx := len(pathToLeaf) - 1; idx >= 0; idx-- {
```
renaynay, line 71: Nit but it's better to use something like `i` as it's conventional for index inside loops.

fraud/bad_encoding.go
liamsi: When I saw this for the first time, I thought that we could even panic in this case as this means sth is seriously wrong here and there is likely no way to recover from this (we should never test a befp against a malformed extended header). For this PR I think we can simply return a better error message and a comment though:
```suggestion
		// NOTE: This should never happen as callers of this method should not feed it with a 
		// malformed extended header. 
		return errors.New("invalid extended header: length of row and column roots do not match")
```

Alternative to the above: add a comment to the method signature that the passed-in extended header had to already pass basic validation and panic here.

go.sum
vgonkivs: re-run go mod tidy but go sum is not modified

ipld/read_test.go
```diff
@@ -441,3 +443,107 @@ func generateRandNamespacedRawData(total, nidSize, leafSize uint32) [][]byte {
...
+		{dah.RowsRoots},
+		{dah.ColumnRoots},
+	}
+
+	for i, tt := range tests {
+		t.Run(strconv.Itoa(i), func(t *testing.T) {
+			for _, root := range tt.roots {
+				rootCid := plugin.MustCidFromNamespacedSha256(root)
+				for index := 0; uint(index) < width; index++ {
+					fmt.Println(index)
```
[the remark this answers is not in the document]
vgonkivs: It was for debug purposes. Will delete it. Thanks!

fraud/bad_encoding.go
```diff
@@ -0,0 +1,160 @@
...
+	if int(p.Index) >= len(merkleRowRoots) || int(p.Index) >= len(merkleColRoots) {
+		return errors.New("invalid fraud proof: incorrect Index of bad row/col")
+	}
+	if len(merkleRowRoots) != len(merkleColRoots) {
+		// NOTE: This should never happen as callers of this method should not feed it with a
+		// malformed extended header.
+		panic("invalid extended header: length of row and column roots do not match")
+	}
+	if len(merkleRowRoots) != len(p.Shares) || len(merkleColRoots) != len(p.Shares) {
+		return errors.New("invalid fraud proof: invalid shares")
```
adlerjohn, line 111:
```suggestion
		return fmt.Errorf("invalid fraud proof: incorrect number of shares %d != %d", len(p.Shares), len(merkleRowRoots))
```

fraud/bad_encoding.go
```diff
@@ -0,0 +1,160 @@
...
+	tree := wrapper.NewErasuredNamespacedMerkleTree(uint64(len(shares) / 2))
+	for i, share := range rebuiltShares {
+		tree.Push(share, rsmt2d.SquareIndex{Axis: uint(p.Index), Cell: uint(i)})
+	}
+
+	// comparing rebuilt Merkle Root of bad row/col with respective Merkle Root of row/col from block
+	merkleRoot := merkleRowRoots[p.Index]
+	if !p.isRow {
+		merkleRoot = merkleColRoots[p.Index]
+	}
```
adlerjohn, line 153: Isn't this the same value as `root`?

adlerjohn (commented):
The verification logic seems sound. Just some style nits mostly.

docs/adr/adr-006-fraud-service.md
```diff
@@ -31,15 +31,15 @@ The result of `RepairExtendedDataSquare` will be an error [`ErrByzantineRow`](ht
...
 ```go
-type ErrBadEncoding struct {
+type ErrByzantine struct {
    // Shares contains all shares from row/col.
    // For non-nil shares MerkleProof is computed
-   Shares []*Share
-   // Position represents the number of row/col where ErrByzantineRow/ErrByzantineColl occurred.
-   Position uint8
+   Shares []*NamespacedShareWithProof
+   // Index represents the number of row/col where ErrByzantineRow/ErrByzantineColl occurred.
```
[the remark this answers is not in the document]
Wondertan: did we updated to latest version yet?

fraud/bad_encoding.go
```diff
@@ -0,0 +1,160 @@
...
+		Shares:      shares.Shares,
+		isRow:       shares.IsRow,
+		Index:       shares.Index,
+	}
+}
+
+// Type returns type of fraud proof
+func (p *BadEncodingProof) Type() ProofType {
+	return BadEncoding
+}
```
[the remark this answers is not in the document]
adlerjohn: I'm okay with leaving this as-is.

ipld/read.go
```diff
@@ -88,6 +102,75 @@ func GetLeaf(ctx context.Context, dag ipld.NodeGetter, root cid.Cid, leaf, total
...
+		if share != nil {
+			proof := make([]cid.Cid, 0)
+			s, err := GetLeafData(ctx, root, uint32(index), uint32(len(shares)), dag)
+			if err != nil {
+				return nil, err
+			}
+			proof, err = GetProof(ctx, dag, root, proof, index, len(shares))
+			if err != nil {
+				return nil, err
+			}
```
Wondertan, line 125: Pls make a TODO and issue, mentioning that `GetLeafData` and `GetProof` can be combined into one function, as they traverse the same tree until the same share. The diff is only the data they collect, which can be controlled with options in the proposed unified function.

docs/adr/adr-006-fraud-service.md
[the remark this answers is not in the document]
vgonkivs: No, Celestia-node uses previous version.

ipld/read.go
```diff
@@ -2,15 +2,29 @@ package ipld
...
+
 	"github.com/celestiaorg/nmt/namespace"
 )
 
+// ErrByzantine is an error converted from rsmt2d.ByzantineRow/Col +
+// Merkle Proof of each share.
+type ErrByzantine struct {
+	Index  uint8
+	Shares []*NamespacedShareWithProof
+	IsRow  bool
```
Wondertan, line 21: Would be nice to rely on the enum defined in the new version of rsmt2d. The one is called `Axis`. IsRow feels kinda ugly, but the new version of rsmt2d is breaking the existing code.

fraud/bad_encoding.go
```diff
@@ -0,0 +1,160 @@
...
+// Validate ensures that fraud proof is correct.
+// Validate checks that provided Merkle Proofs correspond to the shares,
+// rebuilds bad row or col from received shares, computes Merkle Root
+// and compares it with block's Merkle Root.
+func (p *BadEncodingProof) Validate(header *header.ExtendedHeader) error {
+	merkleRowRoots := header.DAH.RowsRoots
+	merkleColRoots := header.DAH.ColumnRoots
+	if header.Height != int64(p.BlockHeight) {
+		return errors.New("invalid fraud proof: incorrect block height")
+	}
```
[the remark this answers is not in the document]
vgonkivs: Yes, but I think we have to handle this case.

ipld/retriever.go
```diff
@@ -50,19 +52,66 @@ func NewRetriever(dag format.DAGService, codec rsmt2d.Codec) *Retriever {
...
 	for _, qs := range newQuadrants(dah) {
-		// TODO: ErrByzantineRow/Col
 		for _, q := range qs {
 			eds, err := ses.retrieve(ctx, q)
 			if err == nil {
 				return eds, nil
 			}
+			if byzErr := r.checkForByzantineError(ctx, dah, err); byzErr != nil {
+				return nil, byzErr
+			}
```
Wondertan, line 62: After we update to the new rsmt2d version, we can make just one if here for `ErrByzantineData` and extract the logic into `handleByzantineError`

ipld/retriever.go
```diff
@@ -50,19 +52,66 @@ func NewRetriever(dag format.DAGService, codec rsmt2d.Codec) *Retriever {
...
+		r.dag,
+		plugin.MustCidFromNamespacedSha256(root),
+		errShares,
+	)
+	if err != nil {
+		return err
+	}
+
+	byz := &ErrByzantine{Index: index, Shares: sharesWithProof, IsRow: isRow}
+	return byz
```
Wondertan, line 112: nit: why not just one line return?

Wondertan, line 112:
```suggestion
	return &ErrByzantine{Index: index, Shares: sharesWithProof, IsRow: isRow}
```

fraud/bad_encoding.go
vgonkivs: Yes. Removed it at all. Thank you.

fraud/bad_encoding.go
```diff
@@ -0,0 +1,160 @@
...
+	return nil
+}
+
+// UnmarshalBEFP converts given data to BadEncodingProof
+func UnmarshalBEFP(befp *pb.BadEncoding) *BadEncodingProof {
+	return &BadEncodingProof{
+		BlockHeight: befp.Height,
+		Shares:      ipld.ProtoToShare(befp.Shares),
+	}
+}
```
Wondertan, line 90: nit: Is this really Unmarshall, if doesn't covert actual bytes to the struct? Do we need this to be exported?

docs/adr/adr-006-fraud-service.md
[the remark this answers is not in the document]
Wondertan: Ok. The new update is breaking, as it defines the new error and makes repair a method, we should probably update this in a subsequent PR.

fraud/bad_encoding_test.go
```diff
@@ -0,0 +1,62 @@
...
+	out := make([][]byte, flattenedEDSSize)
+	count := 0
+	for i := uint(0); i < eds.Width(); i++ {
+		for _, share := range eds.Row(i) {
+			out[count] = share
+			count++
+		}
+	}
+	return out
+}
```
Wondertan, line 62: nit:There is a helper in ipld called ExtractEDS, which does exactly this

fraud/bad_encoding.go
vgonkivs: `ProtoToBEFP` ?

fraud/bad_encoding_test.go
Wondertan: It is called ExtractODS lol, which is a typo you can also fix

fraud/bad_encoding.go
Wondertan: Ok

fraud/bad_encoding_test.go
Wondertan: Actually, it is called ODS because it extracts the Original square, so this helper is unusable in this case. What we can do thought, is to define another ExtractEDS which does the equivalent of the above functions which we will then reuse everywhere.

vgonkivs: Not a typo, because it returns original data square. And flatted - full data square

ipld/read.go
vgonkivs: Will add a todo and create an issue because in our version rsmt2d `row` and `col` are lowercase.

Wondertan: Yes pls

ipld/read.go
vgonkivs: Added

Wondertan (approved):
Great job!!! 
:shipit:

adlerjohn (approved):
:shipit:

renaynay (approved):
Great work @vgonkivs

PR #551, opened 2022-03-25
swamp/tests: light node syncs with multiple trusted peers

### LINKED ISSUE #418: swamp/tests/sync: Light node can be initialised with multiple trusted peers
ATM, the light node can be initialised with only 1 trusted peers. Soon this will be changed due to https://github.com/celestiaorg/celestia-node/issues/364

We need a sync test that checks that this is possible and a negative case where one of the trusted peer is down, but still the light node continues with syncing process

Bidon15:
- Edit WithTrustedPeers to accept a slice of strings/multiaddresses
- Adapt flags
- Minor prettifying of how we treat addr[0] as a separate var called `trustedPeers`
- Increased timeout context for start/stop for stable local execution on macOS, when user executes `make test` in cli

Resolves #418

### DIFF
--- a/node/config_opts.go
+++ b/node/config_opts.go
@@ -19,7 +19,7 @@ func WithTrustedHash(hash string) Option {
 }
 
 // WithTrustedPeers appends a slice of TrustedPeers to the Config.
-func WithTrustedPeers(addr []string) Option {
+func WithTrustedPeers(addr ...string) Option {
 	return func(cfg *Config, _ *settings) (_ error) {
 		cfg.Services.TrustedPeers = append(cfg.Services.TrustedPeers, addr...)
 		return
--- a/node/config_opts.go
+++ b/node/config_opts.go
@@ -18,7 +18,7 @@ func WithTrustedHash(hash string) Option {
 	}
 }
 
-// WithTrustedPeers appends a slice of TrustedPeers to the Config.
+// WithTrustedPeers appends new "trusted peers" to the Config.
 func WithTrustedPeers(addr ...string) Option {
 	return func(cfg *Config, _ *settings) (_ error) {
 		cfg.Services.TrustedPeers = append(cfg.Services.TrustedPeers, addr...)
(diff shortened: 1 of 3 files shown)


node/config_opts.go
```diff
@@ -18,10 +18,10 @@ func WithTrustedHash(hash string) Option {
 	}
 }
 
-// WithTrustedPeer appends TrustedPeer to the Config.
-func WithTrustedPeer(addr string) Option {
+// WithTrustedPeers appends a slice of TrustedPeers to the Config.
+func WithTrustedPeers(addr []string) Option {
```
Wondertan, line 22:
```suggestion
// WithTrustedPeers appends new "trusted peers" to the Config.
func WithTrustedPeers(addr ...string) Option {
```

Wondertan: This way you won't need to wrap values with an array and simply pass them in options as many as you want

node/tests/sync_test.go
```diff
@@ -217,3 +223,66 @@ func TestSyncLightWithFull(t *testing.T) {
...
+2. Start a BN
+3. Check BN is synced to height 20
+4. Create a Full Node(FN) with a connection to BN as a trusted peer
+5. Start a FN
+6. Check FN is synced to height 30
+7. Create a Light Node(LN) with a connection to BN, FN as trusted peers
+8. Start LN
+9. Check LN is synced to height 50
+*/
+func TestSyncLightWithMultiplePeers(t *testing.T) {
```
Wondertan, line 243:
```suggestion
func TestSyncLightWithTrustedPeers(t *testing.T) {
```

Wondertan:
Pls add the kind label

node/tests/sync_test.go
Bidon15: Done

node/config_opts.go
Bidon15: Done

Wondertan (approved):

vgonkivs (approved):

Issue #550, opened 2022-03-25
params: Node can be set for a private/other network type


Bidon15:
ATM, when the node is initialised,  it is automatically trying to connect to bootstrap peers that are serving for `devnet` network. 
We need more flexibility in setting up such parameter as `swamp` tests contains their own network and this is causing warnings on execution with messages like `can not connect to bootstrap peers` which are intended for another use-case. 

This behaviour is also effecting our ephemeral-cluster. 
Some refs can be found in `params/networks.go`

Wondertan:
I thinking about ENV var that allows setting desired network preference

liamsi:
What was the reason to not make this part of the config instead?

Wondertan:
Configs, according to my experience, are something that users often play around with a lot, trying different things, etc. The network switch in the config is something I expected users to touch and break subsequently breaking the synchronization. But I might be overcautious.

liamsi:
> . The network switch in the config is something I expected users to touch and break subsequently breaking the synchronization.

I think proper documentation makes this a non-issue. Can also happen with env vars.

PR #552, opened 2022-03-25
feat(params): network configurations via environment vars

Wondertan:
Mainly this PR adds:
* New Private network type for private or testing setups
* `CELESTIA_PRIVATE_GENESIS` env var which intstructs _private network_ on a node via setting genesis hash
* `CELESTIA_BOOTSTRAPPERS` env var which sets custom bootstrappers for _any_ network.
* `CELESTIA_NETWORK` env var which sets the network for the node.
Closes: https://github.com/celestiaorg/celestia-node/issues/550

### DIFF
--- a/node/tests/swamp/swamp.go
+++ b/node/tests/swamp/swamp.go
@@ -17,11 +17,19 @@ import (
 	rpctest "github.com/tendermint/tendermint/rpc/test"
 	"github.com/tendermint/tendermint/types"
 
+	"github.com/celestiaorg/celestia-node/params"
+
 	"github.com/celestiaorg/celestia-node/core"
 	"github.com/celestiaorg/celestia-node/libs/keystore"
 	"github.com/celestiaorg/celestia-node/node"
 	"github.com/celestiaorg/celestia-node/node/p2p"
 )
+
+func init() {
+	// NOTE: This way any pkg that imports swamp also gets it network to be private.
+	// This is ok as Swamp must not be imported from non-testing environment.
+	params.DefaultNetwork = params.Private
+}
 
 var blackholeIP6 = net.ParseIP("100::")
 
--- a/params/bootstrap.go
+++ b/params/bootstrap.go
@@ -1,6 +1,9 @@
 package params
 
 import (
+	"os"
+	"strings"
+
 	"github.com/libp2p/go-libp2p-core/peer"
 	ma "github.com/multiformats/go-multiaddr"
 )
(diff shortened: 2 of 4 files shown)


Wondertan:
@jbowen93, tagging you here as you might be interested in this. I recommend migrating the ephemeral cluster to use Private Network. Also, the trusted-hash flag will soon be deprecated for Bridge Node, so migration will be required at some point.

Wondertan:
Converting into the draft as #556 changes this a little.

liamsi (commented):
In general: I'd be really in favour if what can be configured via a config, would be configurable via a config. I do not really see why we'd want to introduce ENV vars here. 
IMO even the bootsrappers should be part of a config file instead of being hardcoded here. Similarly the different networks.

Wondertan:
@liamsi, the code _must_ commit to such network-wide stuff.

liamsi:
why? Also default configs can also be considered part of the code or not?

Wondertan:
And more: we define bootstrappers, not the user. We define genesis, not the user. Allowing users to change network-wide params we define is error prone.

liamsi:
I need to understand the rationale here better.

Wondertan:
@liamsi, the rationale behind hardcoding stuff like we do already:
* Genesis per network should be hardcoded because this is the only way you can have trustless sync for a node from scratch and without the need to manually get somewhere else. If you have a valid binary with genesis info inside, you have almost everything to sync reliably, the missing part is a long-range attack. (I believe we discussed this already and mentioned that Ethereum, Bitcoin and etc are doing this. Tendermint uses manual genesis doc distribution)
* The long-range attack is addressed by hardcoding the set of bootstrappers. Bootstrappers in this case are not only providing us entrance to the p2p network but also peers that we(Celestia) guarantee to be honest. This is what prevents any node to be eclipsed by bots and etc, as each node has at least one honest peer to whom it maintains a connection(with celestia-node we don't have a guarantee for such connection). The important part is hardcoding the address and pubkey(peerid), so this honest connection can't be MITMed.

The rationale for env vars:
* There should be a way to run private networks. Previously, when you run celestia-node you are automatically connect to some default public network. This PR preserves this behavior, but allows running the node with custom genesis for a custom private network with a custom set of bootstrappers through env. The reasons why it is through env:
	* For setups like Ephemeral cluster or any CI it is easier to control things from the env and a good practice. Accessing values on the config is PITA in such cases.
	* Config fields are available to users. It might be ok to expose them in config, but I don't see a reason why regular user would won't to run their own private network. In case it is a dev, his dev environment will use the env.
* Similarly there should be a way to control the network bootstrappers for private network, thus the env var to do so.

Wondertan:
Closed in favour of https://github.com/celestiaorg/celestia-node/pull/561

PR #555, opened 2022-03-26
service/block: Delete all of `BlockService` as it is unused.

renaynay:
As we refactored the codebase to focus around `header.Service`, we no longer need `block.Service` to store/retrieve or deal with block data in its entirety in any way.

We now only need the `share` and `header` packages to deal with the concept of blocks in celestia-node.


service/block/store.go
```diff
@@ -1,23 +0,0 @@
...
-func (s *Service) StoreBlockData(ctx context.Context, data *ExtendedBlockData) error {
-	shares := ipld.ExtractODSShares(data)
-	// TODO @renaynay: it's inefficient that we generate the EDS twice: once in event loop
-	// once in the IPLD plugin.
-	_, err := ipld.PutData(ctx, shares, s.store)
-	return err
-}
-
-func (s *Service) GetBlockData(ctx context.Context, dah *da.DataAvailabilityHeader) (*ExtendedBlockData, error) {
-	return ipld.RetrieveData(ctx, dah, s.store, rsmt2d.NewRSGF8Codec())
```
liamsi, line 22: Where does the equivalent of this happen now?

Wondertan: ShareService

Wondertan (approved):
👍🏻 from me.
I would still provide a way our full blocks through an API, but this case be done elsewhere

liamsi (approved):

service/block/store.go
renaynay: for bridge nodes, [here](https://github.com/celestiaorg/celestia-node/blob/539ac21a965c5680d72daefa797a762b02e63325/service/header/header.go#L39).

for full nodes, [here](https://github.com/celestiaorg/celestia-node/blob/539ac21a965c5680d72daefa797a762b02e63325/service/share/full_availability.go#L29).

PR #557, opened 2022-03-27
refactor|chore(node): new TestNode testing utility and node tests cleanup

Wondertan:
Just took my broom and put on my cleaning hat.

PR #561, opened 2022-03-28
feat(params): add new Private network for private/testing setups

### LINKED ISSUE #550: params: Node can be set for a private/other network type

Wondertan:
Substitutes #552.
Based on #556.
Closes #550 

Also, makes Node and Swamp tests run in the private setup. This is also why I did a cleanup in https://github.com/celestiaorg/celestia-node/pull/557

### DIFF
--- a/params/default.go
+++ b/params/default.go
@@ -1,9 +1,21 @@
 package params
 
+import (
+	"os"
+	"strings"
+)
+
 // defaultNetwork defines a default network for the Celestia Node.
 var defaultNetwork = DevNet
 
 // DefaultNetwork returns the network of the current build.
 func DefaultNetwork() Network {
 	return defaultNetwork
+}
+
+func init() {
+	if genesis, ok := os.LookupEnv("CELESTIA_PRIVATE_GENESIS"); ok {
+		defaultNetwork = Private
+		genesisList[Private] = strings.ToUpper(genesis)
+	}
 }
--- a/node/node_bridge_test.go
+++ b/node/node_bridge_test.go
@@ -7,6 +7,8 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	"github.com/celestiaorg/celestia-node/params"
+
 	"github.com/celestiaorg/celestia-node/core"
 )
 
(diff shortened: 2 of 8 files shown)


renaynay (approved):

adlerjohn (approved):
lgtm, but merge conflict

Wondertan:
Will be fixed by rebase on main

params/default.go
```diff
@@ -28,3 +33,10 @@ func DefaultBootstrappersInfos() []peer.AddrInfo {
 func DefaultBootstrappers() []string {
 	return bootstrapList[defaultNetwork] // network is guaranteed to be valid
 }
+
+func init() {
+	if genesis, ok := os.LookupEnv("CELESTIA_PRIVATE_GENESIS"); ok {
```
liamsi, line 38: Can we also document this as an example somewhere? Or do you rather expect anyone who wants to run a private network to read the godoc comment of `GenesisFor`?

liamsi (approved):
Unsure if documentation for the new env var is visible enough but otherwise looks good to me.

params/default.go
Wondertan: @liamsi, WDYM? Like a `something_example.go`?
> as an example somewhere

I expect this but am happy to document this somewhere as well. Ideally, our docs website under the special section about running your own celestia-node networks. cc @YazzyYaz
> Or do you rather expect anyone who wants to run a private network to read the godoc comment of GenesisFor?

liamsi: Maybe simply add an example command-line usage somewhere.

adlerjohn (approved):

params/default.go
Wondertan: https://github.com/celestiaorg/celestia-node/issues/574

Wondertan: Have no idea where to put such clip example right now. Making a special document for it right now is overkill.

Wondertan:
Will rebase and merge after https://github.com/celestiaorg/celestia-node/pull/583 is in main

PR #567, opened 2022-03-30
core: Remove embedded core node option

renaynay:
This PR removes the embedded core node functionality from the repository as users opt to run bridge nodes with remote celestia-core connections rather than embedded at the moment. 

We can reintroduce this functionality if there is a compelling use case later in the future, but for now, it is unnecessary.


liamsi (approved):
I was always against introducing this embedded functionality before we actually had a clear plan how to merge core/app and node (which is likely a post-launch prio only anyways).

node/core/core.go
```diff
@@ -11,44 +13,31 @@ import (
...
 			}
-
-			cfg, err := store.Config()
+			client, err := RemoteClient(cfg)
 			if err != nil {
 				return nil, err
 			}
-
-			return core.NewEmbedded(cfg)
+			err = client.Start()
```
Wondertan, line 39: Wrap it with hook

Wondertan, line 39: We also have to stop it, with hook

core/fetcher.go
```diff
@@ -4,13 +4,17 @@ import (
...
+	logging "github.com/ipfs/go-log/v2"
 	tmbytes "github.com/tendermint/tendermint/libs/bytes"
 	"github.com/tendermint/tendermint/types"
 )
 
 const newBlockSubscriber = "NewBlock/Events"
 
-var newBlockEventQuery = types.QueryForEvent(types.EventNewBlock).String()
+var (
+	log                = logging.Logger("fetcher")
```
Wondertan, line 15: `core` or `core/fetcher`?

Wondertan (commented):
I ok with this, just a few bits of findings.

As we discussed with @renaynay, there is still an unexplored use case for the embedded node and I believe it could speed up things setting up Core+Bridge setups for folks, however, we haven't ever tested it, and thus advertise it for them to try. Still bullish about it but together with App(Core)+Node molding at some far point as per @liamsi.

Oh yeah, almost forget about IOTWAL

Wondertan:
@renaynay, seems like this PR introduces a new flakey test, unless the test is failing constantly locally for you

node/core/core.go
renaynay: `client.Start` and `client.Stop` do not take context :(

renaynay: jk, resolved.

renaynay: resolved?

core/fetcher.go
renaynay: resolved?

node/core/core.go
Wondertan: And? You should start and stop it anyway, just wrap it with anon func

renaynay:
@Wondertan test works for me locally but other tests break with race locally for me (RetrieveBlockData)

Wondertan:
@renaynay, there was core related failed test before you didn't fixed my comments

renaynay:
test w/ race or just unit test? Was it ctx deadline exceeded?

Wondertan:
You can look at the history of CI. There was a strange issue with core saying the height 2 is the initial height, not 1

Wondertan (approved):

PR #573, opened 2022-03-30
node|cmd: Make `StateService` required on node for all node types

renaynay:
Make `StateService` required on `node`. As we only have the `CoreAccess` implementation of `StateAccessor` for now, all nodes will be required to pass a `--core.remote` endpoint on init/start or else construction will fail. Eventually when we implement `P2PAccess`, a core endpoint will be optional, but `StateService` will still be required.


Wondertan (changes requested):
Let's extract the .girignore change into a separate PR which also includes change in makefile which make it build not in the root but into ‘build’ folder

adlerjohn:
Do you mean temporary build files, or the final binary? The final binary should probably be in `bin` rather than `build`.

Wondertan:
> Do you mean temporary build files, or the final binary? The final binary should probably be in `bin` rather than `build`.

@adlerjohn I assume that ‘make build’ is for temporary builds. We don't store final builds in the repo.

liamsi (approved):

cmd/celestia/cmd.go
```diff
@@ -0,0 +1,41 @@
+package main
+
+import (
+	"github.com/spf13/cobra"
+
+	cmdnode "github.com/celestiaorg/celestia-node/cmd"
+)
+
+func parseFlags(cmd *cobra.Command, args []string) error {
```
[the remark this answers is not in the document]
renaynay: I mean the same way we have `baseComponents`, we can have `baseFlags` func parsing for base flags and then have the remainder custom to node type flags additionally?

cmd/celestia/full.go
```diff
@@ -48,7 +55,12 @@ var fullCmd = &cobra.Command{
 			return err
 		}
 
-		err = cmdnode.ParseHeadersFlags(cmd, env)
+		err = cmdnode.ParseCoreFlags(cmd, env)
+		if err != nil {
+			return err
+		}
+
+		err = cmdnode.ParseTrustedHashFlags(cmd, env)
```
Wondertan, line 63: Full node also needs trusted peer thus we should allow it to be modified

cmd/celestia/light.go
```diff
@@ -54,17 +54,17 @@ var lightCmd = &cobra.Command{
...
 		}
 
-		err = cmdnode.ParseMiscFlags(cmd)
+		err = cmdnode.ParseTrustedHashFlags(cmd, env)
 		if err != nil {
 			return err
 		}
 
-		err = cmdnode.ParseCoreFlags(cmd, env)
+		err = cmdnode.ParseMiscFlags(cmd)
```
Wondertan, line 67: * Why order change?
* Same here about trusted peer. Why did you remove it?

Wondertan (changes requested):

Wondertan (approved):

PR #575, opened 2022-03-30
.gitignore | Makefile: Dump bin in `build/` directory and `gitignore` the directory  

renaynay:


Wondertan:
Pls test it manually:
* The lack of build directory may break go build -o
    * If so, we should add the build folder to git and probably add a shadow file inside, so git doesn't ignore an empty dir
    * Maybe there is a better way
* The may -o may also need the name of the output file as well

renaynay:
I have tested this manually. It works.

Bidon15 (approved):

vgonkivs (approved):

Wondertan (approved):

adlerjohn (approved):
utACK

PR #576, opened 2022-03-30
Makefile: Add `make clean` cmd

renaynay:
Cleans up celestia bin from `build/` dir.

Based on #575.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,11 @@ build:
 	@go build -ldflags ${LDFLAGS} ./cmd/celestia
 .PHONY: build
 
+## clean: Clean up celestia-node binary.
+clean:
+	@echo "--> Cleaning up celestia binary (located in ./build/celestia)"
+	@rm build/celestia
+
 ## install: Build and install the celestia-node binary into the GOBIN directory.
 install:
 	@echo "--> Installing Celestia"


Makefile
```diff
@@ -14,6 +14,11 @@ build:
 	@go build -ldflags ${LDFLAGS} ./cmd/celestia
 .PHONY: build
 
+## clean: Clean up celestia-node binary.
+clean:
+	@echo "--> Cleaning up celestia binary (located in ./build/celestia)"
+	@rm build/celestia
```
Wondertan, line 20:
```suggestion
	@echo "--> Cleaning up builds (located in ./build)"
	@rm -rf build/*
```

adlerjohn (approved):
utACK

FWIW, these makefile should probably be made to work using the path of the Makefile itself, rather than the current working directory, to avoid affecting things in unexpected places.

renaynay:
What do you mean @adlerjohn? You can only run the make commands from the same directory as the Makefile

adlerjohn:
https://linux.die.net/man/1/make

> +-f file, --file=file, --makefile=FILE
    Use file as a makefile.

Bidon15:
> https://linux.die.net/man/1/make
> 
> > +-f file, --file=file, --makefile=FILE
> > Use file as a makefile.

wdym @adlerjohn ? 

- using the makefile 's flag `-f --file` is just to specify which file you want to use as a makefile. Such as `make -f /swamp/ogre/shrek` and the `shrek` is the name of the makefile
- using the `-C` flag to only point to a directory where the makefile is used (considering that the name of the makefile is `Makefile`) 

I don't think we are going to move the existing makefile somewhere else from the repo's root. In addition that I highly doubt that any future inner makefiles will make their own make clean cmd(still a `.phony` will be a nice touch, which will be added)

adlerjohn:
We may not move the makefile, but it's possible the makefile is called from a different directory. I remember I had to do that a lot with C++ because C++.

Wondertan (approved):
LGTM, don't think the above is necesary

PR #578, opened 2022-03-31
ci: exclude flaky test from CI test race runs

Bidon15:
Proposal: 
- Adding a new method to exclude test race detector runs in CI but still maintain full execution locally. 
- From now and onwards, I propose we create a an issue and attach it in comment section of the skipped test to remember what happened

Ref: #491

### DIFF
--- a/service/header/header_test.go
+++ b/service/header/header_test.go
@@ -1,4 +1,4 @@
-//go:build test_unit
+//go:build test_unit_core
 
 package header
 


renaynay:
We only need to exclude them from `-race` btw not from regular go test

Bidon15:
> We only need to exclude them from `-race` btw not from regular go test

Correct. it is only for -race flag. All other cases are left as was before 💯 

Ref: #579 #582

.github/workflows/test.yml
```diff
@@ -39,7 +42,7 @@ jobs:
         with:
           go-version: ${{ matrix.go }}
       - name: execute test run
-        run: make test-unit-race
+        run: make SKIP_CI=${{ env.SKIP_CI }} test-unit-race
```
liamsi, line 45: my only concern with this approach is that we'll likely never run these tests other than in CI 🤔

Bidon15: There is risk. But it's manageable for now if we don't plan to pump more test cases with core 🤞

Wondertan (approved):
I hope that we won't forget to include them once Core upgrades to the latest tendermint

Wondertan:
I think this is a good idea. We may run the failed tests several times or make them run in a separate job that won't fail the merge. Have no idea if that is possible and if there is any playbook to just ctrl+c/v it. Otherwise, we would need some CI engineering involved.
> Instead of just skipping, can the way these tests are ran be refactored, so that the sometimes-failing tests run in a different job?

renaynay:
This approach would be optimal: https://github.com/celestiaorg/celestia-node/pull/578#pullrequestreview-927862365

Bidon15:
Now, we are using build tags `test_unit` `test_unit_core` `test_swamp` to actually point what package run in respective job runs

I don't know why, but all the CI bugs are not reproducible after core tests migration to a separate job 😓

renaynay (approved):
utACK, thank you for this 🙏🏻

.github/workflows/test.yml
```diff
@@ -6,6 +6,9 @@ on:
       - main
   pull_request:
 
+env:
+  SKIP_CI: TRUE
```
adlerjohn, line 10: Is this still used? I see it used later in the CI yaml, but not in the Go code.

Bidon15: Great question! it is used for the `Makefile`. This can be convenient as if we want to add more tags to the test run in the future but not change the .yml as the downstream dep along the way

https://github.com/celestiaorg/celestia-node/pull/578/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R45

adlerjohn (approved):

Wondertan (commented):
Lemme _wondert _ about this for some time. It is _really_ hard for me to embrace right now that we will need to add go build to each test file we make. I haven't seen anything like this and feel a bit defensive, though I think this solution is clever.
 
Also, should we make our devs think about CI details, i.e. what action to put my test to by tagging the file? Or should the CI process be isolated and decide on its own where and how things should be run.

cc @renaynay @adlerjohn @liamsi

liamsi:
> I hope that we won't forget to include them once Core upgrades to the latest tendermint

Let's open an issue before we merge this then!

> Also, should we make our devs think about CI details, i.e. what action to put my test to by tagging the file? 

Ideally no. But I think for now this is simpler than making the CI more "clever".

adlerjohn (commented):
Merge conflict

PR #581, opened 2022-03-31
service/state: Add unit test for `CoreAccess`

renaynay:
BLOCKED ON https://github.com/celestiaorg/celestia-app/pull/267

### DIFF
--- a/service/state/core_access_test.go
+++ b/service/state/core_access_test.go
@@ -1,0 +1,59 @@
+package state
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


Wondertan:
@renaynay, what is the status here. https://github.com/celestiaorg/celestia-app/pull/267 is closed now

renaynay:
We still don't have ability to mock out the application so right now, i've been testing this manually. I think it's fine if our endpoints are tested via something like swamp eventually instead of in unit tests (until we have a good mock celestia-app)

Wondertan:
@Bidon15 finally found a way how to get Core working in Testground with code only. Which means we can then adapt it in Swamp and unblock this

PR #589, opened 2022-04-05
refactor(node): update FX DI

### LINKED ISSUE #503: deps: update FX
This might also include full removal of temporary `fxutil` constructions

Wondertan:
This PR:
* Updates FX to the newest version, which has a feature requested by us: https://github.com/uber-go/fx/issues/825
* Stops relying on the custom implementation of the feature in `fxutil` and uses FX's directly
* Refactor/simplifies Node's options 

The follow-up PR will
* Remove unused code in `fxutil`
* Move `fxutil` into `libs`

Closes #503


node/services/service.go
```diff
@@ -36,13 +36,13 @@ func HeaderSyncer(
 }
 
 // P2PSubscriber creates a new header.P2PSubscriber.
-func P2PSubscriber(lc fx.Lifecycle, sub *pubsub.PubSub) *header.P2PSubscriber {
+func P2PSubscriber(lc fx.Lifecycle, sub *pubsub.PubSub) (*header.P2PSubscriber, *header.P2PSubscriber) {
```
renaynay, line 39: Why returning it twice?

Wondertan: Good Q. According to docs: https://github.com/uber-go/fx/blob/master/annotated.go#L228-L234.

So if you return the same thing as two different types, you have to return it twice. We didn't have to do so with `fxutil` before, as I made this a special case there. There is no special case in FX, and I also feel why they do so, so we have what we have.

renaynay: ack

renaynay (approved):

PR #594, opened 2022-04-06
chore(das): prevent logging of success on failed sample

Wondertan:
Successful logging can confuse when it is not really successful


renaynay (approved):
thank

PR #598, opened 2022-04-06
test: write full node reconstruction test on the share service level

Wondertan:
This PR:
* Add test to assert that full node can reconstruct block from light clients only
* Updates test utilities accordingly
* Uses bitswap sessions for ShareAvailability
	* Provides share retrieval speed improvements.

### DIFF
--- a/ipld/read.go
+++ b/ipld/read.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/ipfs/go-cid"
 	ipld "github.com/ipfs/go-ipld-format"
+	"github.com/ipfs/go-merkledag"
 	"golang.org/x/sync/errgroup"
 
 	"github.com/tendermint/tendermint/pkg/da"
@@ -35,7 +36,7 @@ func RetrieveData(
 	}
 	errGroup, ctx := errgroup.WithContext(parentCtx)
 	errGroup.Go(func() error {
-		return fillQuadrant(ctx, quadrant, dag, dataSquare)
+		return fillQuadrant(ctx, quadrant, merkledag.NewSession(ctx, dag), dataSquare)
 	})
 	if err := errGroup.Wait(); err != nil {
 		return nil, err
--- a/service/share/full_availability.go
+++ b/service/share/full_availability.go
@@ -2,6 +2,7 @@ package share
 
 import (
 	"context"
+	"errors"
 
 	format "github.com/ipfs/go-ipld-format"
 
@@ -26,6 +27,17 @@ func NewFullAvailability(service format.DAGService) Availability {
 // SharesAvailable reconstructs the data committed to the given Root by requesting
 // enough Shares from the network.
 func (fa *fullAvailability) SharesAvailable(ctx context.Context, root *Root) error {
+	ctx, cancel := context.WithTimeout(ctx, AvailabilityTimeout)
+	defer cancel()
+
 	_, err := ipld.RetrieveData(ctx, root, fa.service, rsmt2d.NewRSGF8Codec())
+	if err != nil {
+		log.Errorw("availability validation failed", "root", root.Hash(), "err", err)
+		if errors.Is(err, format.ErrNotFound) || errors.Is(err, context.DeadlineExceeded) {
+			return ErrNotAvailable
+		}
+
+		return err
+	}
 	return err
 }
--- a/service/share/full_availability_test.go
+++ b/service/share/full_availability_test.go
@@ -3,8 +3,10 @@ package share
 import (
 	"context"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"github.com/celestiaorg/celestia-node/service/header"
 )
@@ -35,10 +37,40 @@ func TestShareAvailableOverMocknet_Full(t *testing.T) {
 	defer cancel()
 
 	net := NewDAGNet(ctx, t)
-	_, root := net.RandFullService(16)
-	serv := net.CleanService()
+	_, root := net.RandFullService(32)
+	serv := net.CleanFullService()
 	net.ConnectAll()
 
 	err := serv.SharesAvailable(ctx, root)
+	assert.NoError(t, err)
+}
+
+func TestShareAvailable_FullOverLights(t *testing.T) {
+	const (
+		origSquareSize = 8
+		lightNodes     = 192
+	)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
+	defer cancel()
+
+	net := NewDAGNet(ctx, t)
+	_, root := net.RandFullService(origSquareSize) // make a source node, a.k.a bridge
+	full := net.CleanFullService()                 // make a full which reconstructs data
+
+	lights := make([]Service, lightNodes)
+	for i := 0; i < len(lights); i++ {
+		lights[i] = net.CleanLightService()
+	}
+	net.ConnectAll()
+	for i := 0; i < len(lights); i++ {
+		err := lights[i].SharesAvailable(ctx, root)
+		require.NoError(t, err)
+	}
+
+	// ensure there is no connection between source and full nodes
+	// so that full reconstructs from the light nodes only
+	net.Disconnect(0, 1)
+
+	err := full.SharesAvailable(ctx, root)
 	assert.NoError(t, err)
 }
--- a/service/share/light_availability.go
+++ b/service/share/light_availability.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 
 	format "github.com/ipfs/go-ipld-format"
+	"github.com/ipfs/go-merkledag"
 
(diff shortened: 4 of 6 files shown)


ipld/read.go
```diff
@@ -35,7 +36,7 @@ func RetrieveData(
 	}
 	errGroup, ctx := errgroup.WithContext(parentCtx)
 	errGroup.Go(func() error {
-		return fillQuadrant(ctx, quadrant, dag, dataSquare)
+		return fillQuadrant(ctx, quadrant, merkledag.NewSession(ctx, dag), dataSquare)
```
liamsi, line 39: Huh, didn't you already introduce this a few months ago?

service/share/full_availability_test.go
```diff
@@ -35,10 +37,40 @@ func TestShareAvailableOverMocknet_Full(t *testing.T) {
...
+	const (
+		origSquareSize = 8
+		lightNodes     = 192
+	)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
+	defer cancel()
+
+	net := NewDAGNet(ctx, t)
+	_, root := net.RandFullService(origSquareSize) // make a source node, a.k.a bridge
+	full := net.CleanFullService()                 // make a full which reconstructs data
```
liamsi, line 58: make a full ?

service/share/full_availability_test.go
```diff
@@ -35,10 +37,40 @@ func TestShareAvailableOverMocknet_Full(t *testing.T) {
...
+	}
+	net.ConnectAll()
+	for i := 0; i < len(lights); i++ {
+		err := lights[i].SharesAvailable(ctx, root)
+		require.NoError(t, err)
+	}
+
+	// ensure there is no connection between source and full nodes
+	// so that full reconstructs from the light nodes only
+	net.Disconnect(0, 1)
```
liamsi, line 72: Q: This means that the fullnode is connected to _all_ light nodes though, right?

ipld/read.go
Wondertan: In core? Yes. This is not the core.

service/share/full_availability_test.go
Wondertan: Huh. It is full.

service/share/full_availability_test.go
Wondertan: There is a connect all above

liamsi: I'm wondering if we should not also test the case where that full node is not directly connected to all light nodes directly. Also, we should have a test where there are a few full nodes that are each connected to a (potenially overlapping) subset of light nodes.
But this test is already super cool!

service/share/full_availability_test.go
liamsi:
```suggestion
	full := net.CleanFullService()                 // make a full availability service which reconstructs data
```

service/share/full_availability.go
```diff
@@ -26,6 +27,17 @@ func NewFullAvailability(service format.DAGService) Availability {
 // SharesAvailable reconstructs the data committed to the given Root by requesting
 // enough Shares from the network.
 func (fa *fullAvailability) SharesAvailable(ctx context.Context, root *Root) error {
+	ctx, cancel := context.WithTimeout(ctx, AvailabilityTimeout)
+	defer cancel()
+
 	_, err := ipld.RetrieveData(ctx, root, fa.service, rsmt2d.NewRSGF8Codec())
```
liamsi, line 33: If we switch to leopard there will be a lot of places like this that will need changing. I think we should simply use the constructor in the consts package in core everywhere instead. Orthogonal to this PR though.

service/share/full_availability_test.go
Wondertan: So you for referring to comment)

service/share/full_availability_test.go
Wondertan: We can do more such tests, sure.  

About the case with full not connected to all light clients, how is it supposed to find others?

service/share/full_availability_test.go
liamsi: yes :)

renaynay (approved):
utACK, but LGTM 

Only thing I would say is that the method name `Clean_____` are confusing bc you're returning a fresh new instance, but I usually assume `Clean` to mean you're clearing an already - used instance.

service/share/full_availability.go
adlerjohn: Can someone file an issue to track this refactoring?

Wondertan: https://github.com/celestiaorg/celestia-node/issues/604

renaynay (approved):

service/share/full_availability_test.go
```diff
@@ -34,11 +36,97 @@ func TestShareAvailableOverMocknet_Full(t *testing.T) {
...
+	net.Disconnect(source.ID(), full.ID())
+
+	err := full.SharesAvailable(ctx, root)
+	assert.NoError(t, err)
+}
+
+func TestShareAvailable_MultipleFullOverLights(t *testing.T) {
+	const (
+		origSquareSize = 8
+		lightNodes     = 192
```
liamsi, line 85: I think these numbers should documented.

service/share/full_availability_test.go
```diff
@@ -34,11 +36,97 @@ func TestShareAvailableOverMocknet_Full(t *testing.T) {
...
+
+	for i := 0; i < len(lights1); i++ {
+		err := lights1[i].SharesAvailable(ctx, root)
+		require.NoError(t, err)
+	}
+
+	for i := 0; i < len(lights2); i++ {
+		err := lights2[i].SharesAvailable(ctx, root)
+		require.NoError(t, err)
+	}
```
liamsi, line 120: How is it guaranteed that a fullnode can't reconstruct from `lights1` or `lights2` exclusively?

liamsi (commented):
I think because these tests are critical they should be documented.

service/share/full_availability_test.go
Wondertan: By the connection we make once the light node is constructed above in the code

Wondertan: `lights2` are connected to `full2` and `lights` respectively

liamsi: Yes, but from looking at the code it could also be that `full2` could already reconstruct the block via `lights2` exclusively. Then this is basically the same test `TestShareAvailable_FullOverLights` with two fullnodes instead of one.

service/share/full_availability_test.go
```diff
@@ -34,11 +36,129 @@ func TestShareAvailableOverMocknet_Full(t *testing.T) {
...
+
+func TestShareAvailable_MultipleFullOverLights(t *testing.T) {
+	// S - Source
+	// L - Light Node
+	// F - Full Node
+	// ── - connection
+	//
+	// Topology:
+	// NOTE: There are more Light Nodes in practice
+	// ┌─┬─┬─B─┬─┬─┐
```
liamsi, line 90:
```suggestion
	// ┌─┬─┬─S─┬─┬─┐
```

service/share/full_availability_test.go
```diff
@@ -34,11 +36,129 @@ func TestShareAvailableOverMocknet_Full(t *testing.T) {
...
+	defer cancel2()
+
+	// check that full2 cannot verify on its own
+	err = full2.SharesAvailable(ctx2, root)
+	require.ErrorIs(t, err, ErrNotAvailable)
+
+	// but after they connect
+	net.Connect(full1.ID(), full2.ID())
+
+	// they both should be able to reconstruct the block
```
liamsi, line 159: ❤️

Wondertan:
Will be rebased and merged once everything to make tests happy is done. Specifically #637 and #623

service/share/full_availability_test.go
liamsi: Now there is also `TestShareAvailable_MultipleFullOverLights` which is what I meant. 

> About the case with full not connected to all light clients, how is it supposed to find others?

That is a good question. I think we should capture this in an issue if it does not exist already. IIRC, we are now assuming that the light nodes find the full nodes and not vice versa, right?

adlerjohn:
Looks like blocking PRs have been merged.

Wondertan:
@adlerjohn, thanks for notifying. It turns out that I referenced the wrong issue. It should be #650 instead of #637, and the former is in progress, so we cannot proceed with the PR yet.

service/share/full_availability_test.go
Wondertan: Yes, so you can resolve this comment now until we implement discovery.(https://github.com/celestiaorg/celestia-node/issues/599)
> IIRC, we are now assuming that the light nodes find the full nodes and not vice versa, right?

adlerjohn:
Converting to draft until unblocked.

Wondertan:
Closing in favour of https://github.com/celestiaorg/celestia-node/pull/702

Issue #606, opened 2022-04-11
Use tagged releases for Celestia dependencies


adlerjohn:

liamsi:
currently:
https://github.com/celestiaorg/celestia-node/blob/16dba9747d46b8b26cffc9385054703b3a8d8b9c/go.mod#L9

liamsi:
Also, it would be good if a linter could automatically check for this. So we won't miss these again. The line above was introduce in #420 and approved by 3 reviewers...

Wondertan:
True, we should've caught this

PR #607, opened 2022-04-11
fix go.mod to use tagged releases

liamsi:
quick fix for #606


go.mod
```diff
@@ -255,5 +255,5 @@ replace (
 	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v0.44.1-celestia
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
 	github.com/libp2p/go-libp2p-pubsub v0.5.7-0.20211029175501-5c90105738cf => github.com/celestiaorg/go-libp2p-pubsub v0.5.7-0.20220325112944-d33a3e5e13d5
```
liamsi, line 257: @Wondertan should I also update this dependency?

go.mod
```diff
@@ -255,5 +255,5 @@ replace (
...
-	github.com/tendermint/tendermint v0.34.14 => github.com/celestiaorg/celestia-core v0.34.14-celestia.0.20220228062426-721562a3544d
+	github.com/tendermint/tendermint v0.34.14 => github.com/celestiaorg/celestia-core v0.34.15-celestia
```
liamsi, line 258: This should also make the code duplication in #529 unnecessary.

go.mod
Wondertan: We can tag things on our pubsub fork just to be consistent with tagging. Could you do it?

liamsi: Yes

Wondertan: The branch: https://github.com/celestiaorg/go-libp2p-pubsub/tree/old-datastore

liamsi: https://github.com/celestiaorg/go-libp2p-pubsub/releases/tag/v0.5.6-celestia

Wondertan (approved):
Thanks!

liamsi:
unfortunately this is blocked on `TestSyncLightWithBridge`, `TestSyncFullWithBridge`, `TestSyncLightWithFull`, `TestSyncLightWithTrustedPeers` consistently failing.

liamsi:
OK, actually. Same tests above also fail here: https://github.com/celestiaorg/celestia-node/pull/612
So these failures are orthogonal to this PR. Though I think we must fix these tests asap instead of ignoring them.

PR #609, opened 2022-04-11
Remove title from bug issue template

adlerjohn:
The issue template already automatically adds the `bug` label, so the `[Bug]: ` in the title is redundant.


Wondertan (approved):
Agree

PR #617, opened 2022-04-12
(cmd | node/rpc): Make RPC port configurable and ensure port 0 used for swamp

renaynay:
Make RPC port configurable and ensure that `TestSwamp` uses port 0 option to prevent trying to use same port for every node.


liamsi (approved):
👏🏼 ❤️ 🥳 ✅

PR #619, opened 2022-04-13
service/state: Add `SubmitPayForData` endpoint

renaynay:
This PR introduces a `SubmitPayForData` endpoint.

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -52,7 +52,8 @@ type Node struct {
 	DataExchange exchange.Interface
 	DAG          format.DAGService
 	// p2p protocols
-	PubSub     *pubsub.PubSub
+	PubSub *pubsub.PubSub
+	// services
 	ShareServ  share.Service   // not optional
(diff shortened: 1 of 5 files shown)


renaynay:
Initially @liamsi suggested that I add a parameter to the `SubmitPFD` endpoint called `maxSize` so that the user could specify the `(max share size || max num shares)-- clarify this for me @liamsi` they're willing to pay for and if the message size exceeds that, then the request would fail. 

After speaking w/ @evan-forbes, we decided that we should just leave this parameter out for now until they've explored non-interactive defaults further in celestia-app. Currently, it will only add something like max 240 bytes to the message, but with non interactive defaults, that could balloon to something much larger in which case it would be useful.

Another point is that we want to remove this code entirely from celestia-node and have the canonical PFD constructor / submitter methods to exist inside celestia-app instead.

WDYT @liamsi (and maybe @adlerjohn)

PR #622, opened 2022-04-14
Fix `GetSharesByNamespace` ordering

### LINKED ISSUE #621: `GetSharesByNamespace` returns shares in non-deteministic order
[`GetSharessByNamespace`](https://github.com/celestiaorg/celestia-node/blob/28020da54d46b2dec29763a51c6c434b157e84dd/service/share/share.go#L153) spawns a goroutine for each `roodCID`, and results are returned via channel, and appended to an array:
https://github.com/celestiaorg/celestia-node/blob/28020da54d46b2dec29763a51c6c434b157e84dd/service/share/share.go#L188

Because of this, the order of returned shares is non-deterministic.

tzdybal:
Resolves #621.

### DIFF
--- a/service/share/share.go
+++ b/service/share/share.go
@@ -161,34 +161,34 @@ func (s *service) GetSharesByNamespace(ctx context.Context, root *Root, nID name
 		return nil, ipld.ErrNotFoundInRange
 	}
 
-	type res struct {
-		nodes []format.Node
-		err   error
-	}
-	resultCh := make(chan *res)
-
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
-	for _, rootCID := range rowRootCIDs {
-		go func(rootCID cid.Cid) {
-			nodes, err := ipld.GetLeavesByNamespace(ctx, s.dag, rootCID, nID)
-			resultCh <- &res{nodes: nodes, err: err}
-		}(rootCID)
+	errCh := make(chan error)
+	nodes := make([][]format.Node, len(rowRootCIDs))
+	for i, rootCID := range rowRootCIDs {
+		go func(i int, rootCID cid.Cid) {
+			var err error
+			nodes[i], err = ipld.GetLeavesByNamespace(ctx, s.dag, rootCID, nID)
+			errCh <- err
+		}(i, rootCID)
+	}
+
+	for i := 0; i < len(rowRootCIDs); i++ {
+		select {
+		case err := <-errCh:
+			if err != nil {
+				return nil, err
+			}
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		}
 	}
 
 	namespacedShares := make([]Share, 0)
 	for i := 0; i < len(rowRootCIDs); i++ {
-		select {
-		case result := <-resultCh:
-			if result.err != nil {
-				return nil, result.err
-			}
-			for _, node := range result.nodes {
-				namespacedShares = append(namespacedShares, node.RawData()[1:])
-			}
-		case <-ctx.Done():
-			return nil, ctx.Err()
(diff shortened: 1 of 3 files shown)


service/share/share.go
```diff
@@ -161,37 +161,37 @@ func (s *service) GetSharesByNamespace(ctx context.Context, root *Root, nID name
...
+		case err := <-errCh:
+			if err != nil {
+				return nil, err
 			}
 		case <-ctx.Done():
 			return nil, ctx.Err()
 		}
 	}
 
+	namespacedShares := make([]Share, 0)
```
Wondertan, line 188: Preallocate capacity pls

service/share/share.go
```diff
@@ -161,37 +161,37 @@ func (s *service) GetSharesByNamespace(ctx context.Context, root *Root, nID name
...
-			resultCh <- &res{nodes: nodes, err: err}
-		}(rootCID)
+	errCh := make(chan error)
+	nodes := make([][]format.Node, len(rowRootCIDs))
+	for i, rootCID := range rowRootCIDs {
+		go func(i int, rootCID cid.Cid) {
+			var err error
+			nodes[i], err = ipld.GetLeavesByNamespace(ctx, s.dag, rootCID, nID)
+			errCh <- err
+		}(i, rootCID)
```
Wondertan, line 174: Nice!

service/share/share.go
```diff
@@ -161,37 +161,37 @@ func (s *service) GetSharesByNamespace(ctx context.Context, root *Root, nID name
...
 		select {
-		case result := <-resultCh:
-			if result.err != nil {
-				return nil, result.err
-			}
-			for _, node := range result.nodes {
-				namespacedShares = append(namespacedShares, node.RawData()[1:])
+		case err := <-errCh:
+			if err != nil {
+				return nil, err
```
Wondertan, line 181: nit: In this case now, it would be nicer/cleaner to use [errgoup](https://pkg.go.dev/golang.org/x/sync/errgroup) instead, but not necessary.

Wondertan (commented):
Thank you for the quick fix. And a special thank you for propper commit format!

Left two nits, otherwise LGTM

service/share/share.go
tzdybal: Each node can have arbitrary number of shares inside. So without iterating, can't per-allocate _precisely_. Another option is to push number of shares along with the error using a channel. Both adds complexity.
I can also pre-allocate for `len(nodes)` aka `len(rowRootCIDs)`.

service/share/share.go
tzdybal: Oh, I didn't know you use errgroup in node :) Will fix.

service/share/share.go
Wondertan: Oh, true, we never know precisely how many shares we submit unless we encode the number of shares in the first metadata like share when PFD is sent. 

It is fine for now to keep it as it is then.

Wondertan (approved):
👍

renaynay (approved):
Thanks for finding + fixing :)

liamsi (approved):

service/share/share.go
```diff
@@ -161,34 +163,25 @@ func (s *service) GetSharesByNamespace(ctx context.Context, root *Root, nID name
...
-		select {
-		case result := <-resultCh:
-			if result.err != nil {
-				return nil, result.err
-			}
-			for _, node := range result.nodes {
-				namespacedShares = append(namespacedShares, node.RawData()[1:])
-			}
-		case <-ctx.Done():
-			return nil, ctx.Err()
```
liamsi, line 191: Q: Wouldn't this have been fixable by sorting the returned shares before returning them?

Issue #608, opened 2022-04-11
flaky TestBridge_WithMockedCoreClient


liamsi:
### Steps to reproduce it

CI

### Expected result

not flaky

### Actual result

flaky

### Relevant log output

```shell
--- SKIP: TestBridge_WithMockedCoreClient (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xb21c80]

goroutine 1282 [running]:
github.com/tendermint/tendermint/libs/pubsub.(*Subscription).Out(0x2)
	/home/runner/go/pkg/mod/github.com/celestiaorg/celestia-core@v0.34.14-celestia.0.20220228062426-721562a3544d/libs/pubsub/subscription.go:43
github.com/tendermint/tendermint/rpc/core.Subscribe.func1()
	/home/runner/go/pkg/mod/github.com/celestiaorg/celestia-core@v0.34.14-celestia.0.20220228062426-721562a3544d/rpc/core/events.go:50 +0x9e
created by github.com/tendermint/tendermint/rpc/core.Subscribe
	/home/runner/go/pkg/mod/github.com/celestiaorg/celestia-core@v0.34.14-celestia.0.20220228062426-721562a3544d/rpc/core/events.go:47 +0x485
FAIL	github.com/celestiaorg/celestia-node/node	0.193s
FAIL
exit status 1
```
https://github.com/celestiaorg/celestia-node/runs/5958778931?check_suite_focus=true#step:4:406

Wondertan:
@liamsi, @adlerjohn

Consult https://github.com/celestiaorg/celestia-core/issues/667 for why this happens

liamsi:
I know, just keeping track of the flaky tests makes still sense. Otherwise someone seeing a test failure won't know if the team is aware or not.

PR #623, opened 2022-04-14
shares|ipld: Rework data retrieval 

### LINKED ISSUE #508: ipld: implement retry mechanism for block repairing if it fails with ErrUnrepairableDataSquare.
Supplement for #442 by adding a retry mechanism in case when `RetrieveData` by rows returns `ErrUnrepairableDataSquare`. 

Solution:
Try to repair a block using columns for the same quadrant.

### LINKED ISSUE #509: ipld: implement retry mechanism if dag request was timed out
Supplement for #442. 
Currently, we are using requests to ipld without any timeout. So we can run into a situation when `dag.Get` will be blocked forever. To avoid this we have to provide a timeout for each `dag.Get` request.
In case if timeout will be reached, we have to:
1) save successfully fetched shares;
2) start fetching shares from the next quadrant;
3) try to repair block with fetched shares.

NOTE:
We should repeat these steps if timeout will be reached for the next quadrants.

### LINKED ISSUE #516: ipld: convert roots from [][]bytes to cids in RetrieveData
Supplement for #442.
We need to convert` rowRoots/colRoots` from [][]byte to cids in the beginning of `RetrevieData`  to be consistent.

### LINKED ISSUE #604: share: ensure constant rsmt2d contructor from the core is used throughout codebase

Wondertan:
### What and Why
This PR comes from the requirement to make tests in #598 pass deterministically by reimplementing the original goal of #517. At first, I implemented the goal by simply updating the current code without reworking it. Unfortunately, that didn't work well, and there was a bug that made me stuck for a few days. After extensive multiday debugging, I cleaned up and reworked the existing implementation of RetrieveData in studying the code for the bug. The rework landed into a new `ipld.Retriever` component, which now not just requests the data via sampling quadrants but also retries for other quadrants. 

Eventually, the bug was about rsmt2d not being suitable for recovering over the same data square multiple times, as it sets nil shares to zeroed shares to the given flattened square slice(it actually updates the given slice in place after all!). ~~I will report and link here the issue soon with the suggested fix.~~ The issue https://github.com/celestiaorg/rsmt2d/issues/72. Unless it is fixed, this PR has a workaround with the TODO on it. 

### TODO
- [x] Rework IPLD batch committing logic
- [x] Fix rsmt2d's zeroing of nil shares (https://github.com/celestiaorg/rsmt2d/pull/76)
- [x] Use [rsmt2d v0.4.0](https://github.com/celestiaorg/rsmt2d/releases/tag/v0.4.0) with the above fix
- [x] Parallelize downloading from different roots
- [x] Add Debug logging
- [x] Improve comments
- [x] Add randomization with Col roots and retries(will fix in #508)
- [ ] Write unit test for quadrants retries(Currently tested only with unmerged tests in #598)
- [ ] Manually test that a Full Node is able to sync devnet

Closes #509
Closes #508
Closes #516 , 
Partially closes #604

Note: @vgonkivs was informed about the new implementation of RetrieveData

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/celestiaorg/celestia-app v0.2.0-rc-1
 	github.com/celestiaorg/go-libp2p-messenger v0.1.0
 	github.com/celestiaorg/nmt v0.8.0
-	github.com/celestiaorg/rsmt2d v0.3.1
+	github.com/celestiaorg/rsmt2d v0.4.0
 	github.com/cosmos/cosmos-sdk v0.44.3
 	github.com/dgraph-io/badger/v2 v2.2007.4
 	github.com/gogo/protobuf v1.3.3
--- a/ipld/read.go
+++ b/ipld/read.go
@@ -2,162 +2,17 @@ package ipld
 
 import (
 	"context"
-	"math/rand"
 
 	"github.com/ipfs/go-cid"
 	ipld "github.com/ipfs/go-ipld-format"
-	"golang.org/x/sync/errgroup"
-
-	"github.com/tendermint/tendermint/pkg/da"
-	"github.com/tendermint/tendermint/pkg/wrapper"
 
 	"github.com/celestiaorg/celestia-node/ipld/plugin"
 	"github.com/celestiaorg/nmt"
 	"github.com/celestiaorg/nmt/namespace"
-	"github.com/celestiaorg/rsmt2d"
 )
 
-// RetrieveData asynchronously fetches block data using the minimum number
-// of requests to IPFS. It fails if one of the random samples sampled is not available.
-func RetrieveData(
-	parentCtx context.Context,
-	dah *da.DataAvailabilityHeader,
-	dag ipld.DAGService,
-	codec rsmt2d.Codec,
-) (*rsmt2d.ExtendedDataSquare, error) {
-	edsWidth := len(dah.RowsRoots)
-	rowRoots := dah.RowsRoots
-	dataSquare := make([][]byte, edsWidth*edsWidth)
-
-	quadrant, err := pickRandomQuadrant(rowRoots)
-	if err != nil {
-		return nil, err
-	}
-	errGroup, ctx := errgroup.WithContext(parentCtx)
-	errGroup.Go(func() error {
-		return fillQuadrant(ctx, quadrant, dag, dataSquare)
-	})
-	if err := errGroup.Wait(); err != nil {
-		return nil, err
-	}
-	batchAdder := NewNmtNodeAdder(parentCtx, ipld.NewBatch(parentCtx, dag, ipld.MaxSizeBatchOption(batchSize(edsWidth))))
-	tree := wrapper.NewErasuredNamespacedMerkleTree(uint64(edsWidth)/2, nmt.NodeVisitor(batchAdder.Visit))
-	extended, err := rsmt2d.RepairExtendedDataSquare(dah.RowsRoots, dah.ColumnRoots, dataSquare, codec, tree.Constructor)
-	if err != nil {
-		return nil, err
-	}
-
-	return extended, batchAdder.Commit()
-}
-
-type quadrant struct {
-	isLeftSubtree bool
-	from          int
-	to            int
-	rootCids      []cid.Cid
-}
-
-func pickRandomQuadrant(roots [][]byte) (*quadrant, error) {
-	edsWidth := len(roots)
-	// get the random number from 1 to 4.
-	q := rand.Intn(4) + 1 //nolint:gosec
-	quadrant := &quadrant{rootCids: make([]cid.Cid, edsWidth/2)}
-	// quadrants 1 and 3 corresponds to left subtree,
-	// 2 and 4 to the right subtree
-	/*
-		| 1 | 2 |
-		| 3 | 4 |
-	*/
-	// choose subtree
-	if q%2 == 1 {
-		quadrant.isLeftSubtree = true
-	}
-
-	// define range of shares for sampling
-	if q > 2 {
-		quadrant.from = edsWidth / 2
-		quadrant.to = edsWidth
-	} else {
-		quadrant.to = edsWidth / 2
-	}
-
-	var err error
-	for index, counter := quadrant.from, 0; index < quadrant.to; index++ {
-		quadrant.rootCids[counter], err = plugin.CidFromNamespacedSha256(roots[index])
-		if err != nil {
-			return nil, err
-		}
-		counter++
-	}
-	return quadrant, nil
-}
-
-// fillQuadrant fetches 1/4 of shares for the given root
-func fillQuadrant(
-	ctx context.Context,
-	quadrant *quadrant,
-	dag ipld.NodeGetter,
-	dataSquare [][]byte,
-) error {
-	errGroup, ctx := errgroup.WithContext(ctx)
-	for i := 0; i < len(quadrant.rootCids); i++ {
-		i := i
-		errGroup.Go(func() error {
-			leaves, err := getSubtreeLeaves(
-				ctx,
-				quadrant.rootCids[i],
-				dag,
-				quadrant.isLeftSubtree,
-				uint32(len(quadrant.rootCids)/2),
-			)
-			if err != nil {
-				return err
-			}
-			quadrantWidth := len(quadrant.rootCids)
-			for leafIdx, leaf := range leaves {
-				// shares from the right subtree should be
-				// inserted after all shares from the left subtree
-				if !quadrant.isLeftSubtree {
-					leafIdx += len(leaves)
-				}
-				shareData := leaf.RawData()[1:]
-				// dataSquare represents a single dimensional slice
-				// of 4 quadrants. The representation of quadrants in dataSquare will be
-				// | 1 | | 2 | | 3 | | 4 |
-				// position is calculated by offsetting the index to respective quadrant
-				position := ((i + quadrant.from) * quadrantWidth * 2) + leafIdx
-				dataSquare[position] = shareData[NamespaceSize:]
-
-			}
-			return err
-		})
-	}
-	return errGroup.Wait()
-}
-
-// getSubtreeLeaves returns only one subtree - left or right
-func getSubtreeLeaves(
-	ctx context.Context,
-	root cid.Cid,
-	dag ipld.NodeGetter,
-	isLeftSubtree bool,
-	treeSize uint32,
-) ([]ipld.Node, error) {
-	nd, err := dag.Get(ctx, root)
-	if err != nil {
-		return nil, err
-	}
-	links := nd.Links()
-	subtreeRootHash := links[0].Cid
-	if !isLeftSubtree && len(links) > 1 {
-		subtreeRootHash = links[1].Cid
-	}
-
-	return getLeaves(ctx, dag, subtreeRootHash, make([]ipld.Node, 0, treeSize))
-}
-
-// getLeaves recursively starts going down to find all leafs from the given root
-func getLeaves(ctx context.Context, dag ipld.NodeGetter, root cid.Cid, leaves []ipld.Node) ([]ipld.Node, error) {
+// GetLeaves recursively starts going down to find all leafs from the given root
+func GetLeaves(ctx context.Context, dag ipld.NodeGetter, root cid.Cid, leaves []ipld.Node) ([]ipld.Node, error) {
 	// request the node
 	nd, err := dag.Get(ctx, root)
 	if err != nil {
@@ -177,8 +32,8 @@ func getLeaves(ctx context.Context, dag ipld.NodeGetter, root cid.Cid, leaves []
 	}
 
 	for _, node := range lnks {
-		// recursively walk down through selected children to get the leaves
-		leaves, err = getLeaves(ctx, dag, node.Cid, leaves)
+		// recursively walk down through selected children to request the leaves
+		leaves, err = GetLeaves(ctx, dag, node.Cid, leaves)
 		if err != nil {
 			return nil, err
 		}
--- a/ipld/read.go
+++ b/ipld/read.go
@@ -11,7 +11,8 @@ import (
 	"github.com/celestiaorg/nmt/namespace"
 )
 
-// GetLeaves recursively starts going down to find all leafs from the given root
+// GetLeaves gets all the leaves under the given root. It recursively starts traversing the DAG to find all the leafs.
+// It also takes a pre-allocated slice 'leaves'.
 func GetLeaves(ctx context.Context, dag ipld.NodeGetter, root cid.Cid, leaves []ipld.Node) ([]ipld.Node, error) {
 	// request the node
 	nd, err := dag.Get(ctx, root)
--- a/ipld/retriever.go
+++ b/ipld/retriever.go
@@ -120,7 +120,7 @@ func (rs *retrieverSession) retrieve(ctx context.Context, q *quadrant) (*rsmt2d.
 	rs.request(ctx, q)
 
 	// try repair
-	eds, err := rsmt2d.RepairExtendedDataSquare(rs.dah.RowsRoots, rs.dah.ColumnRoots, rs.square, rs.codec, rs.treeFn)
+	err := rsmt2d.RepairExtendedDataSquare(rs.dah.RowsRoots, rs.dah.ColumnRoots, rs.square, rs.codec, rs.treeFn)
 	if err != nil {
 
 		// TODO(@Wondertan): This is a hack around limitation of rsmt2d. It sets empty zeroed shares back to nil
@@ -136,7 +136,12 @@ func (rs *retrieverSession) retrieve(ctx context.Context, q *quadrant) (*rsmt2d.
 		return nil, format.ErrNotFound
 	}
 
-	return eds, rs.adder.Commit()
+	err = rs.adder.Commit()
+	if err != nil {
+		return nil, err
+	}
+
+	return rsmt2d.ImportExtendedDataSquare(rs.square, rs.codec, rs.treeFn)
 }
 
 type quadrant struct {
--- a/ipld/retriever.go
+++ b/ipld/retriever.go
@@ -114,6 +114,14 @@ func (r *Retriever) newSession(ctx context.Context, dah *da.DataAvailabilityHead
 }
 
 func (rs *retrieverSession) retrieve(ctx context.Context, q *quadrant) (*rsmt2d.ExtendedDataSquare, error) {
+	defer func() {
+		// all shares which were requested or repaired are written to disk with the commit
+		// we store *all*, so they are served to the network, including incorrectly committed data(BEFP case),
+		// so that network can check BEFP
+		if err := rs.adder.Commit(); err != nil {
+			log.Errorw("committing DAG", "err", err)
+		}
+	}()
 	// request and fill quadrant's into
 	// we don't care about the errors here, just need to request as much data as we can to be able to reconstruct below
 	rs.request(ctx, q)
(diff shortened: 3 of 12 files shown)


Wondertan:
The test failure is #608

adlerjohn:
https://github.com/celestiaorg/rsmt2d/releases/tag/v0.4.0 is now released, which should be able to handle multiple repairs on the same square by representing missing shares as `nil` internally.

ipld/retriever.go
```diff
@@ -0,0 +1,167 @@
...
+func (rs *retrieverSession) retrieve(ctx context.Context, q *quadrant) (*rsmt2d.ExtendedDataSquare, error) {
+	// request and fill quadrant's into
+	// we don't care about the errors here, just need to request as much data as we can to be able to reconstruct below
+	rs.request(ctx, q)
+
+	// try repair
+	err := rsmt2d.RepairExtendedDataSquare(rs.dah.RowsRoots, rs.dah.ColumnRoots, rs.square, rs.codec, rs.treeFn)
+	if err != nil {
+		log.Errorw("not enough shares sampled to recover, retrying...", "err", err)
+		return nil, format.ErrNotFound
```
vgonkivs, line 125: We have to return err in this case, not ErrNotFound. Because all data from ByzantineRow/Col will be missed

ipld/retriever.go
```diff
@@ -0,0 +1,167 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return rsmt2d.ImportExtendedDataSquare(rs.square, rs.codec, rs.treeFn)
+}
+
+type quadrant struct {
+	roots []cid.Cid
+	x, y  int
```
vgonkivs, line 138: Could you please leave a comment what do these coordinates mean? And how to detect that quadrant corresponded to erasured data

Wondertan: Added comment

ipld/retriever.go
```diff
@@ -0,0 +1,178 @@
...
+	"github.com/celestiaorg/celestia-node/ipld/plugin"
+	"github.com/celestiaorg/nmt"
+	"github.com/celestiaorg/rsmt2d"
+)
+
+var log = logging.Logger("ipld")
+
+// RetrieveQuadrantTimeout limits the time for retrieval of a quadrant
+// so that Retriever can retry another quadrant.
+var RetrieveQuadrantTimeout = time.Minute * 5
```
liamsi, line 24: What is this timeout based on?

ipld/retriever.go
```diff
@@ -0,0 +1,178 @@
...
+// Instead of requesting data 'share by share' it requests data by quadrants
+// minimizing bandwidth usage in the happy cases.
+//
+//  ---- ----
+// | 0  | 1  |
+//  ---- ----
+// | 2  | 3  |
+//  ---- ----
+// Retriever randomly picks one of the data square quadrants and tries to request them one by one until it is able
+// the whole square.
```
liamsi, line 36: There is a word missing here:
> until it is able
> // the whole square.

service/share/interface.go
```diff
@@ -12,7 +12,7 @@ var ErrNotAvailable = errors.New("da: data not available")
 // AvailabilityTimeout specifies timeout for DA validation during which data have to be found on the network,
 // otherwise ErrNotAvailable is fired.
 // TODO: https://github.com/celestiaorg/celestia-node/issues/10
-const AvailabilityTimeout = 10 * time.Minute
+const AvailabilityTimeout = 20 * time.Minute
```
liamsi, line 15: why did you chose to double this?

liamsi (commented):
Left some smaller comments. will try to properly review soon.

ipld/retriever.go
```diff
@@ -0,0 +1,178 @@
...
+	// ------  ------
+	x, y int
+}
+
+func (rs *retrieverSession) request(ctx context.Context, q *quadrant) {
+	ctx, cancel := context.WithTimeout(ctx, RetrieveQuadrantTimeout)
+	defer cancel()
+
+	size := len(q.roots)
+	for i, root := range q.roots {
```
vgonkivs, line 157: May be we can request Leaves for each root in a separate go routine? to increase fetching speed.

ipld/retriever.go
```diff
@@ -0,0 +1,178 @@
...
+	roots []cid.Cid
+	// Coordinates of each quadrant(x;y)
+	// ------  -------
+	// |  Q0 | | Q1  |
+	// |(0:0)| |(1:0)|
+	// ------  -------
+	// |  Q2 | | Q3  |
+	// |(0;1)| |(1;1)|
+	// ------  ------
+	x, y int
```
liamsi, line 149: Q: Is there a particular reason, we did not just number then 0, 1, 2, 3 instead of (0,0), (1,0), (0,1), (1, 1)

ipld/retriever.go
```diff
@@ -0,0 +1,178 @@
...
+			log.Errorw("getting root", "err", err)
+			continue // so that we still try for other roots
+		}
+
+		// TODO(@Wondertan): GetLeaves should return everything it was able to request even on error,
+		// 	so that we fill as much data as possible
+		// get leaves of left or right subtree
+		nds, err := GetLeaves(ctx, rs.dag, nd.Links()[q.x].Cid, make([]format.Node, 0, size))
+		if err != nil {
+			log.Errorw("getting all the leaves", "err", err)
```
liamsi, line 169:
```suggestion
			log.Errorw("failed retrieving all leaves", "err", err)
```

ipld/retriever.go
```diff
@@ -0,0 +1,178 @@
...
+func NewRetriever(dag format.DAGService, codec rsmt2d.Codec) *Retriever {
+	return &Retriever{dag: dag, codec: codec}
+}
+
+// Retrieve retrieves all the data committed to DataAvailabilityHeader.
+// It aims to request from the network only 1/4 of the data and reconstructs other 3/4 parts using the rsmt2d.Codec.
+// It steadily tries to request other 3/4 data if 1/4 is not found within RetrieveQuadrantTimeout or unrecoverable.
+// TODO(@Wondertan): Randomize requesting from Col Roots with retrying.
+func (r *Retriever) Retrieve(ctx context.Context, dah *da.DataAvailabilityHeader) (*rsmt2d.ExtendedDataSquare, error) {
+	daroots := dah.RowsRoots
```
liamsi, line 52: Q: Can you remind me why we only sample from rows?

liamsi (commented):
Looks really good. Just a few questions and minor things. Hope we can merge this soon.

service/share/share.go
```diff
@@ -9,18 +9,20 @@ import (
...
 	"github.com/celestiaorg/celestia-node/ipld/plugin"
 	"github.com/celestiaorg/nmt"
 	"github.com/celestiaorg/nmt/namespace"
-	"github.com/celestiaorg/rsmt2d"
 )
 
 var log = logging.Logger("share")
 
+// DefaultCodec sets the default rsmt2d.Codec for shares.
+var DefaultCodec = consts.DefaultCodec
```
renaynay, line 24: Do we need to export this too?

renaynay (commented):
Can we maybe add some `debug` or `trace` level logs in retriever.go?

service/share/share.go
Wondertan: Yes

ipld/retriever.go
```diff
@@ -0,0 +1,178 @@
...
+			y = 1
+		}
+
+		qs[i] = &quadrant{
+			roots: roots[origSize*y : origSize*(y+1)],
+			x:     x,
+			y:     y,
+		}
+	}
+	// shuffling helps data to be equally distributed around the network
```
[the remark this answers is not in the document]
Wondertan: This way Full Nodes sample random quadrant instead of the first one.

ipld/retriever.go
```diff
@@ -0,0 +1,178 @@
...
+func (rs *retrieverSession) retrieve(ctx context.Context, q *quadrant) (*rsmt2d.ExtendedDataSquare, error) {
+	defer func() {
+		// all shares which were requested or repaired are written to disk with the commit
+		// we store *all*, so they are served to the network, including incorrectly committed data(BEFP case),
+		// so that network can check BEFP
+		if err := rs.adder.Commit(); err != nil {
+			log.Errorw("committing DAG", "err", err)
+		}
+	}()
+	// request and fill quadrant's into
```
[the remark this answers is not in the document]
Wondertan: Into rs.sqaure. Thanks for catching

ipld/retriever.go
```diff
@@ -0,0 +1,178 @@
...
+			x:     x,
+			y:     y,
+		}
+	}
+	// shuffling helps data to be equally distributed around the network
+	rand.Shuffle(len(qs), func(i, j int) { qs[i], qs[j] = qs[j], qs[i] })
+
+	ses := r.newSession(ctx, dah)
+	for _, q := range qs {
+		// TODO: ErrByzantineRow/Col
```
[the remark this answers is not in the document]
renaynay: ack

ipld/retriever.go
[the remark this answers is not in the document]
renaynay: comment is confusing then

ipld/retriever.go
Wondertan: I’ll add a todo in this pr. Later on, while fixing all the left todos, I all triage the time it may take to implement this, and if it takes more than 30mins, then I will make an issue for this.

ipld/retriever.go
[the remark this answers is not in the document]
Wondertan: How would you describe this without mentioning the full node? The full node is a concept outside of this pkg.

ipld/retriever.go
Wondertan: The reason why the log string is “getting all the leaves” is because the full log line can be read like ERROR getting all the leaves. 

I saw how PL folks were doing this and followed this pattern. We have it already throughout the code.

ipld/retriever.go
Wondertan: There is no reason. We should be ideally fetch randomly from either cols or rows. There is a todo mentioning this, and there is even an issue for retries from col. I just made it consistent with the original implementation.

ipld/retriever.go
Wondertan: I didn't put too much thought into this but rn my thoughts are:
* This was just the first number that came into my mind after Mustafa calculated the possible amount of proofs that need to be downloaded with 4mb blocks, thus roundtrips. 
* This number should cover some discovery waiting time. 
* Setting this number with some leg room ensures we dont cancel the operation too early if the operation tends to be long
* To have a better timeout we should make some real-world measurements.

Wondertan: An issue: https://github.com/celestiaorg/celestia-node/issues/639

service/share/interface.go
Wondertan: It can be the same if I make RetrieveQuadrantTimeout = 2.5 min

ipld/retriever.go
Wondertan: Fixed

ipld/retriever.go
Wondertan: Fixed

ipld/read.go
```diff
@@ -2,162 +2,17 @@ package ipld
...
-	if !isLeftSubtree && len(links) > 1 {
-		subtreeRootHash = links[1].Cid
-	}
-
-	return getLeaves(ctx, dag, subtreeRootHash, make([]ipld.Node, 0, treeSize))
-}
-
-// getLeaves recursively starts going down to find all leafs from the given root
-func getLeaves(ctx context.Context, dag ipld.NodeGetter, root cid.Cid, leaves []ipld.Node) ([]ipld.Node, error) {
+// GetLeaves recursively starts going down to find all leafs from the given root
```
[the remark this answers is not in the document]
Wondertan: Done

ipld/retriever.go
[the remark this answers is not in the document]
Wondertan: fixed

ipld/retriever.go
Wondertan: I updated the PR to include randomization and retrying for cols

ipld/retriever.go
Wondertan: The fix was easy, and it matters on block sizes we will work with. The problem with _correctness_ should be resolved by _correct_ testing, which includes checking for the ordering of things in case of concurrency involved.

Wondertan:
I updated the branch with randomization and retries over columns. Better be done within this PR.

ipld/retriever.go
renaynay: @Wondertan I disagree, especially when the code is as complicated as it is here -- it's already difficult to review and check for correctness as is without adding in the parallelisation. IMO it makes it easier to review when you take the problem one step at a time. I understand it's the "right thing to do" and I do not disagree with the optimisation in general, I just think things like this would belong in a separate, subsequent PR (the way that you've done with others before). I don't think that the argument that go is designed to be used with goroutines justifies including the optimisation in this PR.

ipld/retriever.go
```diff
@@ -0,0 +1,153 @@
...
+	ses := r.newSession(ctx, dah)
+	for _, qs := range newQuadrants(dah) {
+		// TODO: ErrByzantineRow/Col
+		for _, q := range qs {
+			eds, err := ses.retrieve(ctx, q)
+			if err == nil {
+				return eds, nil
+			}
+		}
+		// retry quadrants until we have something or error
```
renaynay, line 60:
```suggestion
		// retry quadrants until we can reconstruct the EDS or error out
```

ipld/retriever.go
```diff
@@ -0,0 +1,153 @@
...
+			nds, err := GetLeaves(ctx, rs.dag, nd.Links()[q.x].Cid, make([]format.Node, 0, size))
+			if err != nil {
+				log.Errorw("getting all the leaves", "root", root.String(), "err", err)
+				return
+			}
+			// fill leaves into the square
+			for j, nd := range nds {
+				rs.square[q.index(i, j)] = nd.RawData()[1+NamespaceSize:]
+			}
+			log.Debugw("received root", "root", root.String())
```
renaynay, line 147: what's the point of this log?

ipld/retriever.go
```diff
@@ -0,0 +1,153 @@
...
+		}
+	}()
+	// request quadrant and fill it into rs.square slice
+	// we don't care about the errors here, just need to request as much data as we can to be able to reconstruct below
+	rs.request(ctx, q)
+
+	// try repair
+	err := rsmt2d.RepairExtendedDataSquare(rs.dah.RowsRoots, rs.dah.ColumnRoots, rs.square, rs.codec, rs.treeFn)
+	if err != nil {
+		log.Errorw("not enough shares to reconstruct data square, retrying...", "err", err)
```
renaynay, line 110:
```suggestion
		log.Errorw("not enough shares to reconstruct data square, requesting more...", "err", err)
```

renaynay: maybe make this a debug level log? How often will something like this pop up? In this case it's not really an "error" if we can still possibly reconstruct from other quadrants. We should only log at error level if reconstruction fails after *all* quadrants tried.

ipld/retriever_quadrant.go
```diff
@@ -0,0 +1,102 @@
...
+	// ------  -------
+	x, y int
+	// source defines the axis for quadrant
+	// it can be either 1 or 0 similar to x and y
+	// where 0 is Row source and 1 is Col respectively
+	source int
+}
+
+// newQuadrants constructs a slice of quadrants from DAHeader.
+// Constantly, there are 4 quadrants for each source: row and col (8 in total).
```
renaynay, line 35:
```suggestion
// There are always 4 quadrants per each source (row and col), so 8 in total.
```

ipld/retriever_quadrant.go
```diff
@@ -0,0 +1,102 @@
...
+	for source, quadrants := range sources {
+		size, qsize := len(daRoots[source]), len(daRoots[source])/2
+		roots := make([]cid.Cid, size)
+		for i, root := range daRoots[source] {
+			roots[i] = plugin.MustCidFromNamespacedSha256(root)
+		}
+
+		for i := range quadrants {
+			// convert quadrant index into coordinates
+			x, y := i%2, i>>1
```
renaynay, line 57: can't you just do `x, y := i%2, i/2` ?

ipld/retriever_quadrant.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// index calculates index for a share in a data square slice flattened by rows.
+//
+// NOTE: The complexity of the formula below comes from:
+//  * Goal to avoid share copying
+//  * Goal to make formula generic for both rows and cols
+//  	* While data square is flattened by rows only
+// TODO(@Wondertan): This can be simplified by making rsmt2d working over 3D byte slice(not flattened)
+func (q *quadrant) index(rootIdx, cellIdx int) int {
```
renaynay, line 84: can you add a unit test for this please

ipld/retriever_quadrant.go
```diff
@@ -0,0 +1,102 @@
...
+	// source defines the axis for quadrant
+	// it can be either 1 or 0 similar to x and y
+	// where 0 is Row source and 1 is Col respectively
+	source int
+}
+
+// newQuadrants constructs a slice of quadrants from DAHeader.
+// Constantly, there are 4 quadrants for each source: row and col (8 in total).
+// The ordering of quadrants is random.
+func newQuadrants(dah *da.DataAvailabilityHeader) [][]*quadrant {
```
renaynay, line 37: can you add a unit test for this?

service/share/share.go
renaynay: I don't know why but ok

Wondertan: Well actually we can make itpublic. The case I was thinking can be done separately. Will do it privately.

ipld/retriever.go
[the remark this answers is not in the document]
renaynay: ack

ipld/retriever.go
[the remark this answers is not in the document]
renaynay: ack

ipld/read.go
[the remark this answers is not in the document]
renaynay: ack

ipld/retriever.go
Wondertan: Agree that this is not an error. I would do a warning instead. The reason is that retry is not a normal case. Usually, everything should be fine and without retries.

ipld/retriever_quadrant.go
Wondertan: I can, let's simplify this a bit

ipld/retriever.go
Wondertan: Changed

ipld/retriever_quadrant.go
Wondertan: Done

ipld/retriever_quadrant.go
Wondertan: Will add in a subsequent PR as we discussed.

ipld/retriever.go
Wondertan: The tradeoff here is a time we don't have. It is faster to solve multiple related problems simultaneously, rather than spreading those over multiple rounds of PRs and reviews. We are not in the state to follow all the software development best practices (which I agree it is), and we have to sacrifice something to speed up development. Btw, from the git history PoV those are different commits and done iteratively. Advise: always review my PRs commit by commit
> IMO it makes it easier to review when you take the problem one step at a time

go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/celestiaorg/celestia-app v0.2.0-rc-1
 	github.com/celestiaorg/go-libp2p-messenger v0.1.0
 	github.com/celestiaorg/nmt v0.8.0
-	github.com/celestiaorg/rsmt2d v0.3.1
+	github.com/celestiaorg/rsmt2d v0.4.0
```
liamsi, line 12: Should we bump this to 0.5.0 directly?

liamsi (approved):
👏🏼

go.mod
Wondertan: I tried, and there are some issues with it. Specifically, it makes the code even bulkier. We discussed this with John and planned to create a few more issues in the rsmt2d repo to improve this.

PR #625, opened 2022-04-15
ci: add job to check for required labels

### LINKED ISSUE #624: ci: Add job to check for required labels in pull request
We need to add a job the will check minimum 1 of required label in pull request:
* kind:bug-fix
* kind:miscellaneous
* kind:breaking
* kind:improvement
* kind:feature

vgonkivs:
Resolves #624

### DIFF
--- a/.github/workflows/labels.yml
+++ b/.github/workflows/labels.yml
@@ -1,0 +1,14 @@
+name: Required Labels
+
+on:
+  pull_request:
+    types: opened
+jobs:
+  label:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: mheap/github-action-required-labels@v1
+        with:
+          mode: minimum
+          count: 1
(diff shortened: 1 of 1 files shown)


.github/workflows/labels.yml
```diff
@@ -0,0 +1,14 @@
+name: Required Labels
+
+on:
+  pull_request:
+    types: opened
```
adlerjohn, line 5: `opened` only is probably not good, because it will fail if a label isn't applied, and then never succeed even if labels are added later. Should make it run when opened and when labels are modified maybe?

vgonkivs: Extend types. Now, it will re-run job on adding/removing labels

liamsi (approved):

adlerjohn (approved):

Wondertan (approved):
LGTM

PR #626, opened 2022-04-15
ipld: Test `GetSharesByNamespace` to check ordering

tzdybal:
testing(service/share): message->shares->IPLD roundtrip test

Resolves https://github.com/celestiaorg/celestia-node/issues/621


Wondertan:
@tzdybal, Why is the block called `invalid`? Is it invalid or valid?

Wondertan:
@tzdybal, so I run the test locally and seems like the data we get is ok, the problem with the ordering again

liamsi:
Is this sth we could integrate with the ephemeral cluster? So instead of hard coding a test vector only, we would have a test that asserts this round trip works with every released versions of the stack? Cc @Bidon15 @jbowen93

Wondertan:
Overkill

jbowen93:
> Member

The

jbowen93:
> Is this sth we could integrate with the ephemeral cluster? So instead of hard coding a test vector only, we would have a test that asserts these kind of round trips works with every released versions of the stack? Cc @Bidon15 @jbowen93

The Ephemeral cluster stands up a celestia-app node (currently multiple, see https://github.com/celestiaorg/ephemeral-cluster/issues/15), a celestia-node bridge node and a celestia-node light node. If there is an RPC that exposes the state of things then this could be tested, but I don't believe this is supported right now.

Wondertan:
@tzdybal, Kindly ask you to co-author @vgonkivs in the commit with the fix.

tzdybal:
> @tzdybal, Kindly ask you to co-author @vgonkivs in the commit with the fix.
Done :)

liamsi:
We can move the discussion around testing into a separate issue but IMO:
 - the ephemeral cluster already spins up node as well and with testing ethermint it could/should also test retrieving data per namespace (as that ethermint instance will have its own namespace already)
 - integration aka swamp tests are good but orthogonal to the above; and so is the mock app (which will simplify integration tests a lot)

Regarding the cluster: we should only clarify when this is run/which events trigger this (only tagging new releases of particular repos or on each push to master, or on a regular basis etc). I defer to @Bidon15's and @jbowen93's judgement here.

service/share/share_test.go
```diff
@@ -135,3 +142,822 @@ func BenchmarkService_GetSharesByNamespace(b *testing.B) {
...
+        "last_commit_hash": "00AAFEB5461158D198967E750E108367EE6D0739E69C0AC2F4D102E0F87D2595",
+        "data_hash": "2AE354BAA0A6C75EDAAB7AA9D6FAAF873FD6FF3E0884DD479306EC6C1AAFD796",
+        "validators_hash": "A76865C38C998C200C6056F8B1D08CB4009683B1110ACE85D95B9889A47650ED",
+        "next_validators_hash": "A76865C38C998C200C6056F8B1D08CB4009683B1110ACE85D95B9889A47650ED",
+        "consensus_hash": "048091BC7DDC283F77BFBF91D73C44DA58C3DF8A9CBC867405D8B7F3DAADA22F",
+        "app_hash": "E2DD3E32D66B330B16BCAB389DA17C2B6BC1B082B2B58ED0A90E8197E36A59A6",
+        "last_results_hash": "E3B0C44298FC1C149AFBF4C8996FB92427AE41E4649B934CA495991B7852B855",
+        "evidence_hash": "E3B0C44298FC1C149AFBF4C8996FB92427AE41E4649B934CA495991B7852B855",
+        "proposer_address": "2CB2349B8516D69839FE49892B1AF6909D7361A7"
+      },
```
liamsi, line 328: As all this contains specifics/internal of the tendermint block + commits, can we add a comment for future maintainers how to handle this test-vector in case there was a breaking change in tendermint (e.g. what if we switched to BLS, or, remove the partSetHeader etc).

liamsi (commented):
Left one comment and I'll update the title. Otherwise looks really good!

service/share/share_test.go
liamsi: Both swamp, as well as e2e tests in the ephemeral cluster would make it unnecessary to hardcode test-vectors that contain and spill over implementation details of structs defined in other repos.

Wondertan: I don't like the fact that we need to copy the JSON structure here for the test, as we should be ideally making a real block with an integration test, but we are blocked on the mocked app, so this way turns out to be the simplest one atm.

Wondertan (approved):
LGTM!

vgonkivs:
May be we can keep data from `sampleBlock` in json file and read from it in test?

renaynay:
Maybe keep it in a directory `test_files` or something like that.

vgonkivs:
`test_files` or `resources`

renaynay:
`testfiles` is fine i think @vgonkivs

Wondertan:
Wow, that is what we need. There is also a new feature in Go for embedding files: https://pkg.go.dev/embed. Pls use that

Wondertan:
Plan let's not make a centralized dir for all the rest files. Ipld can have it's own dir, which the test should embed and rely upon.

vgonkivs:
> Wow, that is what we need. There is also a new feature in Go for embedding files: https://pkg.go.dev/embed. Pls use that

The we should move on 1.18, no?

Wondertan:
@vgonkiv, well, I already did that on my machine ;) But yeah, we should do that for the repo ideally. There are tons of things in 1.18 that we can gain from

tzdybal:
Embed was introduced in [go 1.16](https://tip.golang.org/doc/go1.16#library-embed).

vgonkivs (approved):

PR #633, opened 2022-04-20
github/admin: fix the empty title for bug template

Bidon15:
After merging #609 PR, bug template is not available anymore in our repo due to a template issue containing an explicit empty title. `title: ""`

This PR should fix this proprely and return back the bug template.


liamsi (approved):
utACK

renaynay (approved):

PR #658, opened 2022-04-27
service/state: Add `SubmitPayForData` endpoint

renaynay:
This PR adds a `SubmitPayForData` endpoint that builds, signs and submits a `PayForData` msg using celestia-app's new `PayForData` API. 

So far, I've only tested the endpoint manually. Unit tests for checking state endpoints are likely not feasible (until a lightweight mock celestia-app is created), but we will look into creating a swamp test to stand up a single instance of celestia-app to test the state endpoints as an integration test instead.

### DIFF
--- a/service/state/core_access.go
+++ b/service/state/core_access.go
@@ -10,7 +10,9 @@ import (
 	"google.golang.org/grpc/credentials/insecure"
 
 	"github.com/celestiaorg/celestia-app/app"
+	"github.com/celestiaorg/celestia-app/x/payment"
 	apptypes "github.com/celestiaorg/celestia-app/x/payment/types"
+	"github.com/celestiaorg/nmt/namespace"
 )
 
 // CoreAccessor implements Accessor over an RPC connection
@@ -64,6 +66,15 @@ func (ca *CoreAccessor) Stop(context.Context) error {
 	ca.coreConn = nil
 	ca.queryCli = nil
 	return nil
+}
+
+func (ca *CoreAccessor) SubmitPayForData(
+	ctx context.Context,
+	nID namespace.ID,
+	data []byte,
+	gasLim uint64,
+) (*TxResponse, error) {
+	return payment.SubmitPayForData(ctx, ca.signer, ca.coreConn, nID, data, gasLim)
 }
 
 func (ca *CoreAccessor) Balance(ctx context.Context) (*Balance, error) {
--- a/service/state/rpc.go
+++ b/service/state/rpc.go
@@ -47,7 +47,10 @@ func (s *Service) handleBalanceRequest(w http.ResponseWriter, r *http.Request) {
 	bal, err := s.accessor.Balance(r.Context())
 	if err != nil {
 		w.WriteHeader(http.StatusInternalServerError)
-		w.Write([]byte(err.Error())) //nolint:errcheck
+		_, werr := w.Write([]byte(err.Error()))
+		if werr != nil {
+			log.Errorw("writing response", "endpoint", balanceEndpoint, "err", werr)
+		}
 		log.Errorw("serving request", "endpoint", balanceEndpoint, "err", err)
 		return
 	}
@@ -71,14 +74,20 @@ func (s *Service) handleBalanceForAddrRequest(w http.ResponseWriter, r *http.Req
 	addr, err := types.AccAddressFromBech32(addrStr)
 	if err != nil {
 		w.WriteHeader(http.StatusBadRequest)
-		w.Write([]byte(err.Error())) //nolint:errcheck
+		_, werr := w.Write([]byte(err.Error()))
+		if werr != nil {
+			log.Errorw("writing response", "endpoint", balanceEndpoint, "err", werr)
+		}
 		log.Errorw("serving request", "endpoint", balanceEndpoint, "err", err)
 		return
 	}
 	bal, err := s.accessor.BalanceForAddress(r.Context(), addr)
 	if err != nil {
 		w.WriteHeader(http.StatusInternalServerError)
-		w.Write([]byte(err.Error())) //nolint:errcheck
+		_, werr := w.Write([]byte(err.Error()))
+		if werr != nil {
+			log.Errorw("writing response", "endpoint", balanceEndpoint, "err", werr)
+		}
 		log.Errorw("serving request", "endpoint", balanceEndpoint, "err", err)
 		return
 	}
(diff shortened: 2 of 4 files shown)


service/state/core_access.go
```diff
@@ -66,6 +68,15 @@ func (ca *CoreAccessor) Stop(context.Context) error {
 	return nil
 }
 
+func (ca *CoreAccessor) SubmitPayForData(
+	ctx context.Context,
+	nID namespace.ID,
+	data []byte,
+	gasLim uint64,
```
adlerjohn, line 75: Q: why do you need a gas limit as a parameter her? The gas cost of PFD txs is known from the size of the tx in bytes.
cc @evan-forbes

service/state/rpc.go
```diff
@@ -13,17 +14,26 @@ import (
...
+	NamespaceID string `json:"namespace_id"`
+	Data        string `json:"data"`
+	GasLimit    uint64 `json:"gas_limit"`
+}
+
 func (s *Service) RegisterEndpoints(rpc *rpc.Server) {
 	rpc.RegisterHandlerFunc("/balance", s.handleBalanceRequest, http.MethodGet)
 	rpc.RegisterHandlerFunc(fmt.Sprintf("/balance/{%s}", addrKey), s.handleBalanceForAddrRequest, http.MethodGet)
 	rpc.RegisterHandlerFunc(fmt.Sprintf("/submit_tx/{%s}", txKey), s.handleSubmitTx, http.MethodPost)
+	rpc.RegisterHandlerFunc("/submit_pfd", s.handleSubmitPFD, http.MethodPost)
```
adlerjohn, line 36: Can you abstract `"/submit_pfd"` into a constant since this string is used elsewhere?

service/state/core_access.go
evan-forbes: True! Using the current gas pricing, we could hardcore this. Gas price however we might want to expose.

renaynay: can we hardcode gas lim in the app?

service/state/rpc.go
renaynay: done.

service/state/core_access.go
evan-forbes: yes, we could, and we might want to swap `gasLim` with `gasPrice` here, or alternatively leave it blank and rely on [passing options](https://github.com/celestiaorg/celestia-app/blob/d7bb3934e0c13259bf3d666378ea67fda852b6e5/x/payment/payfordata.go#L30)

evan-forbes: ahhh wait, I think I remember trying this ages ago, and unless we want to hardcode an overly large limit, when submitting the wirePFD, CheckTx will sometimes return an error even though it should and the tx doesn't make it to the mempool :disappointed: 

are we okay with hardcoding a large value, at least for now?

evan-forbes: https://github.com/celestiaorg/celestia-app/issues/378

adlerjohn (approved):
lgtm, but will defer to @evan-forbes for app check

evan-forbes (approved):
looks good from an app perspective

liamsi (approved):

service/state/rpc.go
```diff
@@ -13,35 +14,52 @@ import (
...
+	rpc.RegisterHandlerFunc(fmt.Sprintf("%s/{%s}", submitTxEndpoint, txKey), s.handleSubmitTx, http.MethodPost)
+	rpc.RegisterHandlerFunc(submitPFDEndpoint, s.handleSubmitPFD, http.MethodPost)
 }
 
 func (s *Service) handleBalanceRequest(w http.ResponseWriter, r *http.Request) {
 	bal, err := s.accessor.Balance(r.Context())
 	if err != nil {
 		w.WriteHeader(http.StatusInternalServerError)
-		log.Errorw("serving /balance request", "err", err)
+		w.Write([]byte(err.Error())) //nolint:errcheck
```
Wondertan, line 50: So this way, we send the raw error to the user? Good.

service/state/rpc.go
```diff
@@ -80,24 +100,77 @@ func (s *Service) handleSubmitTx(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	// perform request
+	txResp, err := s.accessor.SubmitPayForData(r.Context(), nID, data, req.GasLimit)
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		_, werr := w.Write([]byte(err.Error()))
+		if werr != nil {
+			log.Errorw("writing response", "endpoint", submitPFDEndpoint, "err", werr)
+		}
```
Wondertan, line 162: For the sake of consistency - //nolint:errcheck or do actual err check in the error writes. Ideally add errchecks everywhere, so that if something very wrong happens, we could debug it from logs

service/state/core_access.go
Wondertan: @evan-forbes, can you pls clarify how decision in https://github.com/celestiaorg/celestia-app/issues/378 affects the `gasLim` param here?

Wondertan (commented):
LGTM

service/state/rpc.go
Wondertan: We might make error handling a bit fancier eventually, but this will probably be part of PublicAPI work

evan-forbes:
> @evan-forbes, can you pls clarify how decision in https://github.com/celestiaorg/celestia-app/issues/378 affects the gasLim param here?

yeah, the gasLim needs to be here, as the gas limit will change depending on the size of the data being posted. Previously we discussed that the gas limit should be identical for all `PFD`s, but that's not the case, as the `WirePFD` (which includes the data) is checked before being admitted to the mempool, and we will deduct the appropriate amount of gas in the state machine soon.

service/state/core_access.go
evan-forbes: yeah, the gasLim needs to be here, as the gas limit will change depending on the size of the data being posted. Previously we discussed that the gas limit should be identical for all PFDs, but that's not the case, as the WirePFD (which includes the data) is checked before being admitted to the mempool, and we will deduct the appropriate amount of gas in the state machine soon.

service/state/rpc.go
renaynay: Yes, for now raw error as the user needs to see what's happening in the case of failure.

service/state/rpc.go
renaynay: done

service/state/core_access.go
renaynay: can we resolve this? @evan-forbes @Wondertan ?

PR #661, opened 2022-04-28
chore(ipld|shares): cleanup Share types, global default rsmt2d codec, various cleanups

Wondertan:
So when onsite ended on Saturday, I had an urge to cleanup `ipld` and `share` pkgs. Mainly, I was confusing myself with multiple Share types and shooting into my own shipooors, so I took my cleaning hat again and asked Mr proper to join me. Mr Proper say hi!
proper-hi

So finally, we managed to finish cleanup which should improve the coherence of the pkgs and remove some tech debts:
* Remove redundant Share types and keep only one type
* Make all the constant dependent on the consts in the Core
* Define default rsmt2d codec and use it everywhere
* Simplify tests
* Resolve couple todos
* Removed ipld.ErrNotFoundInRange
* Improved docs

### DIFF
--- a/ipld/read_test.go
+++ b/ipld/read_test.go
@@ -406,7 +406,7 @@ func TestRetrieveDataFailedWithByzantineError(t *testing.T) {
 	copy(shares[14][8:], shares[15][8:])
 
 	// import corrupted eds
-	batchAdder := NewNmtNodeAdder(ctx, format.NewBatch(ctx, dag))
+	batchAdder := NewNmtNodeAdder(ctx, format.NewBatch(ctx, dag, format.MaxSizeBatchOption(batchSize(width*2))))
 	tree := wrapper.NewErasuredNamespacedMerkleTree(uint64(width), nmt.NodeVisitor(batchAdder.Visit))
 	attackerEDS, err := rsmt2d.ImportExtendedDataSquare(shares, DefaultRSMT2DCodec(), tree.Constructor)
 	require.NoError(t, err)


Wondertan:
@tzdybal, if you have time, take a look at this PR. It is related to what I did recently(even touches yours) and maybe you have some feedback

ipld/read.go
```diff
@@ -125,3 +148,10 @@ func GetLeavesByNamespace(
 	}
 	return out, err
 }
+
+// leafToShare convert an NMT leaf represented in IPLD into a Share
+func leafToShare(nd ipld.Node) Share {
+	// * First byte represents the type fo the node, and it's unrelated to the actual share data
```
adlerjohn, line 154:
```suggestion
	// * First byte represents the type of the node, and is unrelated to the actual share data
```

ipld/read.go
```diff
@@ -125,3 +148,10 @@ func GetLeavesByNamespace(
...
+	// * Additional namespace is prepended so tha parity data can be identified with a parity namespace which we cut off
```
adlerjohn, line 155:
```suggestion
	// * Additional namespace is prepended so that parity data can be identified with a parity namespace, which we cut off
```

ipld/read.go
```diff
@@ -125,3 +148,10 @@ func GetLeavesByNamespace(
...
+	return nd.RawData()[1+NamespaceSize:] // TODO(@Wondertan): Rework NMT/IPLD plugin to avoid the type byte
```
adlerjohn, line 156:
```suggestion
	return nd.RawData()[1+consts.NamespaceSize:] // TODO(@Wondertan): Rework NMT/IPLD plugin to avoid the type byte
```

ipld/write.go
```diff
@@ -38,7 +38,7 @@ func PutData(ctx context.Context, shares [][]byte, adder ipld.NodeAdder) (*rsmt2
 // ExtractODSShares returns the original shares of the given ExtendedDataSquare. This
 // is a helper function for circumstances where PutData must be used after the EDS has already
 // been generated.
-func ExtractODSShares(eds *rsmt2d.ExtendedDataSquare) [][]byte {
+func ExtractODSShares(eds *rsmt2d.ExtendedDataSquare) []Share {
```
adlerjohn, line 41: This is a departure from how we've done things so far, which is just use `[]byte` to represent share. It's something that's disappointed me since we've had countless instances of confusion with what exactly a "share" is (does it contain NIDs, etc.). While I approve of this change in general, it should be done everywhere rather than only in a single case.

Wondertan: Can you point me to the place where similar change is not done, so I missed it?

adlerjohn: Yeah I was referring to the NMT and rsmt2d, which use flat byte slices. rsmt2d has an excuse that it doesn't know about namespaces, but the NMT does and should therefore use `Share` rather than `[]byte`. Well, it's a change for another day.

ipld/read.go
adlerjohn: Shouldn't the same constant be used in all places?

Wondertan: It is the same, because ‘NamespaceSize = consts.NamespaceSize’. I just redefined some of ipld pkg specific consts.

adlerjohn: Is there a reason to redefine the constant? Why not just use `consts.NamespaceSize` directly?

ipld/read.go
```diff
@@ -88,7 +87,31 @@ func GetLeaf(ctx context.Context, dag ipld.NodeGetter, root cid.Cid, leaf, total
...
+	shares := make([]Share, len(leaves))
+	for i, leaf := range leaves {
+		shares[i] = leafToShare(leaf)
+	}
+
+	return shares, nil
+}
+
+// ErrNotFoundInRange is returned when namespace.ID is not found under the root.
+var ErrNotFoundInRange = fmt.Errorf("ipld: namespace.ID not found in range")
```
tzdybal, line 112: I'm a bit confused here TBH. 
Personally, I would prefer to just return `nil` slice instead of error. Not finding namespace is normal (not exceptional/erroneous) situation IMHO.

Internally this error is used only here: https://github.com/celestiaorg/celestia-node/pull/661/files#diff-213b77ca67ac98ceba2a0a80fafb1cd3b312906cb991a5003c808e673160f5f0R140-R147

But if we return `nil, nil` this will streamline the code to something like: 
```go
nds, err := GetLeavesByNamespace(ctx, dag, lnk.Cid, nID)
if err != nil {
	return nil, err
}
out = append(out, nds...)
```
Extra notes:
 * currently error handling is broken anyways (in case of actual error (from dag, if `err != nil` and `err != ErrNotFoundInRange`))
 * `errors.Is` could be used

ipld/read_test.go
```diff
@@ -290,47 +241,50 @@ func TestGetLeavesByNamespace_AbsentNamespaceId(t *testing.T) {
...
+			assertNoRowContainsNID(t, dag, eds, tt.missingNid)
 		})
 	}
 }
 
 func TestGetLeavesByNamespace_MultipleRowsContainingSameNamespaceId(t *testing.T) {
-	t.Run("Same namespace id across multiple rows", func(t *testing.T) {
-		rawData := RandNamespacedShares(t, 16).Raw()
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
renaynay, line 253: use `t.Cleanup(cancel)` pls

ipld/share.go
```diff
@@ -1,49 +1,34 @@
...
 // Share contains the raw share data without the corresponding namespace.
-type Share []byte
-
-// TODO(Wondertan): Consider using alias to namespace.PrefixedData instead
-// NamespacedShare extends a Share with the corresponding namespace.
-type NamespacedShare struct {
-	Share
-	ID namespace.ID
-}
+// NOTE: Alias for the byte is chosen to keep maximal compatibility, especially with rsmt2d. Ideally, we should define
```
renaynay, line 22: Yes ➕

service/share/share_test.go
```diff
@@ -46,13 +45,13 @@ func TestService_GetSharesByNamespace(t *testing.T) {
...
+		{squareSize: 4, expectedShareCount: 2},
 		{squareSize: 16, expectedShareCount: 2},
-		{squareSize: 128, expectedShareCount: 1},
+		{squareSize: 128, expectedShareCount: 2},
 	}
 
-	for i, tt := range tests {
-		t.Run(strconv.Itoa(i), func(t *testing.T) {
+	for _, tt := range tests {
+		t.Run("size"+strconv.Itoa(tt.squareSize), func(t *testing.T) {
```
renaynay, line 54:
```suggestion
		t.Run("size: "+strconv.Itoa(tt.squareSize), func(t *testing.T) {
```

ipld/read.go
Wondertan: @renaynay, well, you originally introduced this error, so I am not sure of the original intention here. 

According to the code, the answer is yes.
>corresponds to it not being found inside a specific row right

ipld/read.go
Wondertan: @adlerjohn, @renaynay. The reason being is to make the `ipld` pkg self-sufficient, meaning that the user(we mostly) don't have to know about yet another dependency(app) to work with the API of this pkg/lib. This is just a general rule of thumb when you design go pkgs.

ipld/read.go
```diff
@@ -42,22 +43,20 @@ func GetLeaves(ctx context.Context, dag ipld.NodeGetter, root cid.Cid, leaves []
 	return leaves, nil
 }
 
-// GetLeafData fetches and returns the data for leaf leafIndex of root rootCid.
-// It stops and returns an error if the provided context is canceled before
-// finishing
-func GetLeafData(
+// GetShare fetches and returns the data for leaf leafIndex of root rootCid.
+func GetShare(
```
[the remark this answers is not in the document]
Wondertan: I got your point and agreed with it. However, the concept of Share [will leave outside](https://github.com/celestiaorg/celestia-node/pull/661#discussion_r861766526) of the node repo. For example, the definition of a share can leave in its repo, and all(rsmt2d, nmt, node) can rely on it. The `share` pkg in `node` will just rely on the `share` defined elsewhere and will only provide the logic around the share structures. 

The IPFS team did that to their low-level definition of data [Block](https://github.com/ipfs/go-block-format), and after years of this being a status quo, it turns out to be not that bad of an architecture decision. In our case, this can be like a globally defined Share format, which all other libs will depend on. Right now, we still have yet another Share definition in the Core/App and the nmt that we should eventually remove. Luckily we don't have multiple Share definitions anymore.

ipld/read.go
Wondertan: @tzdybal, I am fine with removing this error and making it a nil case. Although, you can think of this case from the `io.EOF` perspective, something like `EOData`, where there is no more data to read or no data under the root.

ipld/read.go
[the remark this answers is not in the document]
Wondertan: I still like your idea of decoupling `ipld` from the notion of share. We can make this possible by extracting rsmt2d logic out of `ipld` pkg. The reason is that `rsmt2d` works with shares, not with IPLD leaves(ipld.Node) and this is the main reason for keeping Share within `ipld`. If we move rsmt2d-related  logic up to the `share`, then we can do what you suggest. We should make an issue about this.

ipld/read.go
renaynay:
```suggestion
var ErrNotFoundInRoot = fmt.Errorf("ipld: namespace.ID not found in root")
```

can we do this then? The word `range` is confusing now. (I know i intro'd this err)

ipld/read.go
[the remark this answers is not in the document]
renaynay: ack. This is really just a cleanliness/optimisation thing. I'll quickly make an issue.

ipld/read_test.go
renaynay: We use `t.Cleanup` in tests instead of defer usually, so that's why i mentioned it. Our code is kind of all over the place with using defer / t.Cleanup so whatever.

ipld/read.go
adlerjohn: Oh I see, that makes sense. Though I am concerned that if the value of the two constants drifts, then it's not immediately obvious. Would it be better to have some shared constants package?

ipld/read_test.go
Wondertan: Yeah, my code uses defer, yours uses t.Cleanup, so yeah ,we need to find a consensus on what we are using to be consistent.
> Our code is kind of all over the place with using defer / t.Cleanup so whatever

renaynay: I started using `t.Cleanup` in tests when you suggested it 😂  but I'm fine with either. Same thing, but we need to start being consistent.

ipld/read.go
Wondertan: @adlerjohn,

The aliases are defined like this:
```
const (
	MaxSquareSize = consts.MaxSquareSize
	NamespaceSize = consts.NamespaceSize
	ShareSize = consts.ShareSize
)
```
and not like this:
```
const (
	MaxSquareSize = 128
	NamespaceSize = 8
	ShareSize = 8
)
```

This means that we still rely on the consts pkg by proxying values, and if something is drifting, we don't need to change things twice.

ipld/read_test.go
Wondertan: True, but not for context, iirc
> I started using t.Cleanup in tests when you suggested it

Wondertan: Issued an issue https://github.com/celestiaorg/celestia-node/issues/668

ipld/read.go
liamsi: > Personally, I would prefer to just return nil slice instead of error. Not finding namespace is normal (not exceptional/erroneous) situation IMHO.
If this is is the only case, where the method would return `nil` and no error, I think it makes sense to remove the error completely. IIRC the original intention was to signal that the namespace holds no data in an unambiguous and handleable way.

liamsi: > Personally, I would prefer to just return nil slice instead of error. Not finding namespace is normal (not exceptional/erroneous) situation IMHO.
If this is is the only case, where the method would return `nil` and no error, I think it makes sense to remove the error completely. IIRC the original intention was to signal that the namespace holds no data in an unambiguous and handleable way.

service/share/share.go
```diff
@@ -22,21 +21,15 @@ import (
...
 // Share is a fixed-size data chunk associated with a namespace ID, whose data will be erasure-coded and committed
 // to in Namespace Merkle trees.
-type Share = namespace.PrefixedData8
+type Share = ipld.Share
+
+// GetID extracts namespace ID out of a Share.
+var GetID = ipld.ShareID
+
+// GetData extracts data out of a Share.
+var GetData = ipld.ShareData
```
liamsi, line 32: I took me a while to notice that these are functions 🙈 
It seems like GetData is not (yet?) used anywhere.

liamsi, line 32: I took me a while to notice that these are functions 🙈 
It seems like GetData is not (yet?) used anywhere.

liamsi (approved):
Great cleanup 👏🏼

vgonkivs:
Conflicts

ipld/read.go
Wondertan: Agree with removal. Lets do it

service/share/share.go
Wondertan: Yes. It is not used yet. Unfortunately, I had to define those as funcs and not as methods, because its not possible to define methods for byte aliases in Go.

Wondertan:
Rebase on master with the latest FPs related changes

ipld/get.go
```diff
@@ -0,0 +1,16 @@
...
+// 	"github.com/ipfs/go-cid"
+// )
+//
+// type ShareGetter struct {
+// 	getter blockservice.BlockGetter
+// }
+//
+// func (sg *ShareGetter) GetShares(ctx context.Context, root cid.Cid, len int) ([]Share, error) {
+// 	sg.getter.GetBlocks()
+// }
```
vgonkivs, line 16: Why commented?

Wondertan: That's a rebase artifact. Pushing without this

ipld/read.go
Wondertan: Removed. Looks good without it. Thanks for the suggestion @tzdybal and pls resolve the comment

service/share/share.go
Wondertan: Now its used after rebase on the main

vgonkivs (approved):

vgonkivs:
Looks great.

tzdybal (approved):
Looks good :)

renaynay (approved):
ship

PR #673, opened 2022-05-04
service/share: Rename root to dah in JSON body of `/namespaced_shares` endpoint

renaynay:
## Description

This PR renames the field in the JSON request for `/namespaced_shares` from `root` to `dah` as `root` is a variable name we use inside celestia-node that not many people know about whereas `DAH` is more intuitive.

### GET `/namespaced_shares`

Request

```
curl -X GET -d '{"namespace_id": "fffffffffffffffe", "dah": "7b22726f775f726f6f7473223a5b22414141414141414141414541414141414141414141622b30524d7377796534365a47482b4b66347a784a48364d55336457564d726f33794b4f6a4f3448644a31222c2241414141414141414141482f2f2f2f2f2f2f2f2f2f692f584d58737158512f665377634d633643486a746c7a3954644a6b3546713676362b356f614f66563171222c222f2f2f2f2f2f2f2f2f2f2f2f2f2f2f2f2f2f2f2f2f7a556e4b41357a426661756d79674d696646436e4738616537534d6c46724a416d3947346e75584858484d222c222f2f2f2f2f2f2f2f2f2f2f2f2f2f2f2f2f2f2f2f2f2b3662644d455478527a75774f6873744c416c2f78303248396a5a4671325349524747795041623250756b225d2c22636f6c756d6e5f726f6f7473223a5b22414141414141414141414541414141414141414141576a344361724468634c58356667596c6d586d5a525231336a4e485a67442f7a6379656e6b6544472f6235222c2241414141414141414141482f2f2f2f2f2f2f2f2f2f686b4e37664e67757436416c4b736f4a4f567a4d343532486965376c7265375378347044747433537a3257222c222f2f2f2f2f2f2f2f2f2f2f2f2f2f2f2f2f2f2f2f2f312b794e563651376377685178313637766b705665787a576a744c7767416b696b326636344b4655503661222c222f2f2f2f2f2f2f2f2f2f2f2f2f2f2f2f2f2f2f2f2f7a5a685a736e75684749597071474d513566594f594d44446c2f5444486364762f6b6976647235356d7758225d7d"}' http://127.0.0.1:26658/namespaced_shares
```

Response

```
[]
```

### DIFF
--- a/service/share/rpc.go
+++ b/service/share/rpc.go
@@ -14,7 +14,7 @@ const namespacedSharesEndpoint = "/namespaced_shares"
 // request payload
 type sharesByNamespaceRequest struct {
 	NamespaceID string `json:"namespace_id"`
-	Root        string `json:"root"`
+	DAH         string `json:"dah"`
 }
 
 func (s *service) RegisterEndpoints(rpc *rpc.Server) {
@@ -38,7 +38,7 @@ func (s *service) handleSharesByNamespaceRequest(w http.ResponseWriter, r *http.
 		return
 	}
 	// decode and unmarshal root
-	rawRoot, err := hex.DecodeString(req.Root)
+	rawRoot, err := hex.DecodeString(req.DAH)
 	if err != nil {
 		w.WriteHeader(http.StatusBadRequest)
 		log.Errorw("serving request", "endpoint", namespacedSharesEndpoint, "err", err)
--- a/service/share/rpc.go
+++ b/service/share/rpc.go
@@ -37,22 +37,22 @@ func (s *service) handleSharesByNamespaceRequest(w http.ResponseWriter, r *http.
 		log.Errorw("serving request", "endpoint", namespacedSharesEndpoint, "err", err)
 		return
 	}
-	// decode and unmarshal root
-	rawRoot, err := hex.DecodeString(req.DAH)
+	// decode and unmarshal DAH
+	rawDAH, err := hex.DecodeString(req.DAH)
 	if err != nil {
 		w.WriteHeader(http.StatusBadRequest)
 		log.Errorw("serving request", "endpoint", namespacedSharesEndpoint, "err", err)
 		return
 	}
-	var root Root
-	err = json.Unmarshal(rawRoot, &root)
+	var dah Root
+	err = json.Unmarshal(rawDAH, &dah)
 	if err != nil {
 		w.WriteHeader(http.StatusBadRequest)
 		log.Errorw("serving request", "endpoint", namespacedSharesEndpoint, "err", err)
 		return
 	}
 	// perform request
-	shares, err := s.GetSharesByNamespace(r.Context(), &root, nID)
+	shares, err := s.GetSharesByNamespace(r.Context(), &dah, nID)
 	if err != nil {
 		w.WriteHeader(http.StatusInternalServerError)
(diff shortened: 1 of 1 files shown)


adlerjohn (commented):
Why not use block hash? Or at least data root?

renaynay:
Eventually we need a `GetSharesbyNamespace(height)` but that would require having some sort of communication between shares module and header module.

service/share/rpc.go
```diff
@@ -38,7 +38,7 @@ func (s *service) handleSharesByNamespaceRequest(w http.ResponseWriter, r *http.
 		return
 	}
 	// decode and unmarshal root
-	rawRoot, err := hex.DecodeString(req.Root)
+	rawRoot, err := hex.DecodeString(req.DAH)
```
adlerjohn, line 41: Rename variable too?

adlerjohn (commented):
suspicious lint failure

Wondertan (changes requested):
IDK TBH about this being more intuitive:
* 'dah' on its own doesn't mean anything for the outsider and is far less intuitive then the word that has a meaning(root)
* even if you change 'dah' to `data_availabiltiy_header`, it doesn't really add anything here for a user. A header can mean anything and does not give any insights into the content of the header. It says 'data availability' but do users have to care about it if they assume Celestia provides it for them automatically? 
* 'root' is more intuitive for people familiar with merklized data structures

'data_root' should be the most intuitive yet succint.

tzdybal:
I'm definitely biased, because I used term data availability header a lot, but for me `root` is very confusing. I had to take a look into the code to check what it was.

`data_root` seems fine for me.

tzdybal:
On the other hand, `dah` is a good name, because there will either be:
1. `dah` endpoint returning DAH at given height, or
2. `header` endpoint will returning header that contains `dah` field
... so for consistency, it would be great to use same name everywhere.

liamsi:
I'm not so sure what the best name would be but we use "data availability header" throughout the codebase a lot. I think it would be a bit confusing if we call the DAH root as it is not a single root but many row and column roots.

This also applies to this alias btw:
https://github.com/celestiaorg/celestia-node/blob/38a6a967d8ab4ad35841e2646858b8d9b5421053/service/share/share.go#L35-L37

adlerjohn:
what happened to "data root"?

Wondertan:
It was called root because
* we were thinking about referencing data not by dah, but by data root by putting merkle tree over roots in ipld
* the user of dah doesn't really care if it consists of multiple roots or one. The user doesn't really access those underlying roots and uses the whole dah as an atomic value to get everything needed. Really, there are only cases where dah is used as a whole. Essentially, calling dah as a root is like hiding away implementation details of it. Eventually, share.Root will become data.Root which still makes the most sense to me, unless I am missing something.

YazzyYaz:
LGTM on name change to `dah`

liamsi:
> what happened to "data root"?

It still there. Here: https://github.com/celestiaorg/celestia-core/blob/7e29d6560b7e2c96c60c2c5b2ae5d9e371e382d8/types/block.go#L361

But that is the DAH merkelized (i.e. a simple binary merkle tree with all roots in the dah as leaves).

renaynay:
Closing in favour of #693

Issue #533, opened 2022-03-22
fraud: Implement the stopping of all dependent services upon fraud proof receipt


vgonkivs:
Implement stopping mechanism for all corresponded services when valid FraudProof will be received.

Issue #532, opened 2022-03-22
fraud: Integrate FraudSub with the node


vgonkivs:
Integrate FraudSub service with the node as described in [BEFP ADR](#492)

PR #674, opened 2022-05-04
Implement fraud service

### LINKED ISSUE #531: fraud: Implement FraudSub and integrate it into corresponding services
Implement FraudSub based on the design described in [BEFP ADR](#492) and integrate with corresponded services:

- [x] Implement FraudSub service;
- [x] Integration of FraudSub  with Header verification(for light nodes);
- [x] Integration of FraudSub wit DASer(for full nodes);

vgonkivs:
Resolves #531, #532, #533


node/tests/sync_test.go
```diff
@@ -281,3 +282,40 @@ func TestSyncLightWithTrustedPeers(t *testing.T) {
...
+- CoreClient is started by swamp
+- CoreClient has generated 15 blocks
+Steps:
+1. Create a Bridge Node(BN) with broken block an height 20
+2. Start a BN
+3. Create a Full Node(FN) with a connection to BN as a trusted peer
+4. Start a FN
+5. Check FN is not synced to 21
+*/
+func TestFraudProofBroadcasting(t *testing.T) {
```
Bidon15, line 298: Let's create a separate test file for fraud proofs as this test is not directly testing sync but rather relies on sync as a pre-req

node/tests/sync_test.go
```diff
@@ -281,3 +282,40 @@ func TestSyncLightWithTrustedPeers(t *testing.T) {
...
+Test-Case: Broadcast Fraud Proof to different nodes
+Pre-Requisites:
+- CoreClient is started by swamp
+- CoreClient has generated 15 blocks
+Steps:
+1. Create a Bridge Node(BN) with broken block an height 20
+2. Start a BN
+3. Create a Full Node(FN) with a connection to BN as a trusted peer
+4. Start a FN
+5. Check FN is not synced to 21
```
Bidon15, line 296: More description why not synced is the expected state

Bidon15 (changes requested):

node/tests/sync_test.go
vgonkivs: moved to `fraud_test.go`

fraud/interface.go
```diff
@@ -0,0 +1,45 @@
...
+	// Subscribe allows to subscribe on pub sub topic by its type.
+	// Subscribe should register pub-sub validator on topic.
+	Subscribe(ProofType) (Subscription, error)
+	// RegisterUnmarshaler registers unmarshaler for the given ProofType.
+	// If there is no unmarshaler for `ProofType`, then `Subscribe` returns an error.
+	RegisterUnmarshaler(ProofType, proofUnmarshaler) error
+	// UnregisterUnmarshaler removes unmarshaler for the given ProofType.
+	// If there is no unmarshaler for `ProofType`, then it returns an error.
+	UnregisterUnmarshaler(ProofType) error
+
```
Wondertan, line 35:
```suggestion
```

fraud/service.go
```diff
@@ -0,0 +1,171 @@
...
+
+	logging "github.com/ipfs/go-log/v2"
+	"github.com/libp2p/go-libp2p-core/peer"
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
+)
+
+var log = logging.Logger("fraud")
+
+// proofUnmarshaler aliases a function that parses data to `Proof`
+type proofUnmarshaler func([]byte) (Proof, error)
```
Wondertan, line 20: It is part of the interfaces, so let's move it to interfaces.go

fraud/interface.go
```diff
@@ -0,0 +1,45 @@
...
+// Subscriber encompasses the behavior necessary to
+// subscribe/unsubscribe from new FraudProofs events from the
+// network.
+type Subscriber interface {
+	// Subscribe allows to subscribe on pub sub topic by its type.
+	// Subscribe should register pub-sub validator on topic.
+	Subscribe(ProofType) (Subscription, error)
+	// RegisterUnmarshaler registers unmarshaler for the given ProofType.
+	// If there is no unmarshaler for `ProofType`, then `Subscribe` returns an error.
+	RegisterUnmarshaler(ProofType, proofUnmarshaler) error
```
Wondertan, line 31: If proofUnmarshaler supposed to be passed by users outside of the pkg(and it is), it makes sense to make it public.

fraud/service.go
```diff
@@ -0,0 +1,171 @@
...
+
+// headerFetcher aliases a function that is used to fetch ExtendedHeader from store
+type headerFetcher func(context.Context, uint64) (*header.ExtendedHeader, error)
+
+// Validator aliases a function that validate pubsub incoming messages
+type Validator func(ctx context.Context, proofType ProofType, data []byte) pubsub.ValidationResult
```
Wondertan, line 26: Same here. Lets' move it to the interface.go

fraud/service.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+func NewService(p *pubsub.PubSub, hstore header.Store) Service {
+	return &service{
+		pubsub: p, topics: make(map[ProofType]*pubsub.Topic),
+		unmarshalers: make(map[ProofType]proofUnmarshaler),
+		getter:       hstore.GetByHeight,
+		mu:           &sync.Mutex{},
+	}
+}
```
Wondertan, line 45: You can simply pass(even private) headerFetcher to the the ctor, so that you request only one specific method of the Store instead of the whole Store. Might simplify testing and live if the Service is going to be used elsewhere. Duck typing!

fraud/service.go
```diff
@@ -0,0 +1,171 @@
...
+	"sync"
+
+	"github.com/celestiaorg/celestia-node/header"
+
+	logging "github.com/ipfs/go-log/v2"
+	"github.com/libp2p/go-libp2p-core/peer"
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
+)
+
+var log = logging.Logger("fraud")
```
Wondertan, line 17: Also, let's mpve to interfaces.go, to the top

fraud/service.go
```diff
@@ -0,0 +1,171 @@
...
+
+func (f *service) Stop(context.Context) error {
+	log.Info("fraud service is stopping...")
+	return nil
+}
+
+func (f *service) Subscribe(proofType ProofType) (Subscription, error) {
+	// TODO: @vgonkivs check if fraud proof is in fraud store, then return with error
+	f.mu.Lock()
+	defer f.mu.Unlock()
```
Wondertan, line 70: The function call here may take some time because of `pubsub.Join`. Let's not unlock on defer but once we are done reading unmarshallers.

Wondertan: This read here is also a reason to use RWMutex instead

Wondertan: topics map should have its own mutex.

fraud/service.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+func (f *service) Broadcast(ctx context.Context, p Proof) error {
+	bin, err := p.MarshalBinary()
+	if err != nil {
+		return err
+	}
+	if topic, ok := f.topics[p.Type()]; ok {
+		return topic.Publish(ctx, bin)
+	}
```
Wondertan, line 119: Read from the map without locking!
topics map should have its own RWMutex!

fraud/service.go
```diff
@@ -0,0 +1,171 @@
...
+type Validator func(ctx context.Context, proofType ProofType, data []byte) pubsub.ValidationResult
+
+// service is propagating and validating Fraud Proofs
+// service implements Service interface.
+type service struct {
+	pubsub       *pubsub.PubSub
+	topics       map[ProofType]*pubsub.Topic
+	unmarshalers map[ProofType]proofUnmarshaler
+	getter       headerFetcher
+	mu           *sync.Mutex
```
Wondertan, line 35: Idiomatically, mutexes are defined without a pointer but as a value.

fraud/service.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+func NewService(p *pubsub.PubSub, hstore header.Store) Service {
+	return &service{
+		pubsub: p, topics: make(map[ProofType]*pubsub.Topic),
+		unmarshalers: make(map[ProofType]proofUnmarshaler),
+		getter:       hstore.GetByHeight,
+		mu:           &sync.Mutex{},
```
Wondertan, line 43: If you make Mutex as value, this line won't be needed.

Wondertan (changes requested):
Didn't review everything yet. Needs changes

service/state/service.go
```diff
@@ -39,3 +48,23 @@ func (s *Service) BalanceForAddress(ctx context.Context, addr Address) (*Balance
 func (s *Service) SubmitTx(ctx context.Context, tx Tx) (*TxResponse, error) {
 	return s.accessor.SubmitTx(ctx, tx)
 }
+
+func (s *Service) Start(context.Context) error {
+	ctx, cancel := context.WithCancel(context.Background())
+	s.cancel = cancel
+	go fraud.SubscribeToBefp(ctx, s.fsub, func(context.Context) error {
+		atomic.StoreUint64(&s.isBefpReceived, 1)
```
renaynay, line 56: shouldn't we store this at 0, not 1? 0 == false in Go

Also pls rename to `s.befpReceived

renaynay (changes requested):
General comments. I want to give this a 2nd, in-depth review as a group later once first round of comments addressed

header/p2p/subscriber.go
```diff
@@ -60,6 +60,11 @@ func (p *Subscriber) AddValidator(val header.Validator) error {
 	return p.pubsub.RegisterTopicValidator(PubSubTopic, pval)
 }
 
+// RemoveValidator removes pubsub message validation
+func (p *Subscriber) RemoveValidator() error {
```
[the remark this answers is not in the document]
vgonkivs: Uh, I think I had to add `RemoveValidator` in a separate PR but absence of it became actual only after adding the FraudService:
Validator in `header/p2p/subscriber`  is registering in `Syncer.Start`.  This validator is adding incoming Extheader to headerstore (please see `processIncoming`) under the hood. So, once BEFP is received we are removing validation and thus adding new Extheader to headerstore

fraud/interface.go
vgonkivs: done

fraud/interface.go
vgonkivs: Moved to `interface` and made public

fraud/service.go
vgonkivs: Moved to interface.go

fraud/service.go
vgonkivs: done

fraud/service.go
vgonkivs: done

fraud/service.go
vgonkivs: done

das/daser.go
```diff
@@ -31,6 +36,9 @@ type DASer struct {
 
 	sampleDn  chan struct{} // done signal for sample loop
 	catchUpDn chan struct{} // done signal for catchUp loop
+
+	// isServiceRunnig shows service state
+	isServiceRunnig uint64
```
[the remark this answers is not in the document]
vgonkivs: done

header/sync/sync.go
```diff
@@ -45,14 +48,18 @@ type Syncer struct {
 	pending ranges
 	// cancel cancels syncLoop's context
 	cancel context.CancelFunc
+
+	// isServiceRunnig shows service state
+	isServiceRunnig uint64
```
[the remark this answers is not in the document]
vgonkivs: done

service/rpc/middleware.go
```diff
@@ -0,0 +1,35 @@
...
+		w.Header().Add("Content-Type", "application/json")
+		next.ServeHTTP(w, r)
+	})
+}
+
+func checkAllowance(state *state.Service) mux.MiddlewareFunc {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// check if state service was halted and deny the transaction
+			if r.Method == http.MethodPost && state.IsBefpReceived() {
```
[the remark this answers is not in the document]
vgonkivs: renamed

header/p2p/subscriber.go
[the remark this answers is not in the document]
Wondertan: Just discussed with @vgonkivs how we can remove AddValidator completely.

fraud/service.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+func getSubTopic(p ProofType) string {
+	switch p {
+	case BadEncoding:
+		return "befp-sub"
+	default:
+		panic(fmt.Sprintf("invalid proof type %d", p))
+	}
+}
```
Wondertan, line 192: This requires us to add a case for each new added type. Instead, you can get a topic name by p.Type() or p.Type() + "Sub"

fraud/service.go
```diff
@@ -0,0 +1,192 @@
...
+	if !wasjoined {
+		// add internal validation to topic inside libp2p for provided ProofType
+		if err = f.topics.pubsub.RegisterTopicValidator(
+			getSubTopic(proofType),
+			func(ctx context.Context, _ peer.ID, msg *pubsub.Message) pubsub.ValidationResult {
+				return f.processIncoming(ctx, proofType, msg.Data)
+			},
+		); err != nil {
+			return nil, err
+		}
```
Wondertan, line 114: This could be a method on `topics`

fraud/service.go
```diff
@@ -0,0 +1,192 @@
...
+	return nil
+}
+
+func (f *service) Subscribe(proofType ProofType) (Subscription, error) {
+	// TODO: @vgonkivs check if fraud proof is in fraud store, then return with error
+	f.mu.RLock()
+	u, ok := f.unmarshalers[proofType]
+	f.mu.RUnlock()
+	if !ok {
+		return nil, errors.New("unmarshaler is not registered")
```
Wondertan, line 97:
```suggestion
		return nil, errors.New("fraud: unmarshaler is not registered")
```
Prepending pkg name simplfies debugging errors

fraud/service.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+func (f *service) RegisterUnmarshaler(proofType ProofType, u ProofUnmarshaler) error {
+	f.mu.RLock()
+	defer f.mu.RUnlock()
+
+	if _, ok := f.unmarshalers[proofType]; ok {
+		return errors.New("unmarshaler is registered")
+	}
+	f.unmarshalers[proofType] = u
```
Wondertan, line 127: This is writing, while we take an RLock

fraud/service.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+func (f *service) UnregisterUnmarshaler(proofType ProofType) error {
+	f.mu.RLock()
+	defer f.mu.RUnlock()
+
+	if _, ok := f.unmarshalers[proofType]; !ok {
+		return errors.New("unmarshaler is not registered")
+	}
+	delete(f.unmarshalers, proofType)
```
Wondertan, line 139: This is writing, while we take an RLock

fraud/service.go
```diff
@@ -0,0 +1,192 @@
...
+
+	return newSubscription(t, u)
+}
+
+func (f *service) RegisterUnmarshaler(proofType ProofType, u ProofUnmarshaler) error {
+	f.mu.RLock()
+	defer f.mu.RUnlock()
+
+	if _, ok := f.unmarshalers[proofType]; ok {
+		return errors.New("unmarshaler is registered")
```
Wondertan, line 125:
```suggestion
		return errors.New("fraud: unmarshaler is already registered")
```

fraud/service.go
```diff
@@ -0,0 +1,192 @@
...
+
+	return nil
+}
+
+func (f *service) UnregisterUnmarshaler(proofType ProofType) error {
+	f.mu.RLock()
+	defer f.mu.RUnlock()
+
+	if _, ok := f.unmarshalers[proofType]; !ok {
+		return errors.New("unmarshaler is not registered")
```
Wondertan, line 137:
```suggestion
		return errors.New("fraud: unmarshaler is not registered")
```

fraud/service.go
```diff
@@ -0,0 +1,192 @@
...
+	if err != nil {
+		return err
+	}
+
+	return f.topics.publish(ctx, bin, p.Type())
+}
+
+func (f *service) processIncoming(ctx context.Context, proofType ProofType, data []byte) pubsub.ValidationResult {
+	f.mu.RLock()
+	defer f.mu.RUnlock()
```
Wondertan, line 156: The same issue with defer here. `proof.Validate` can take some time, same for the `f.getter`. Taking the lock all that time is not really a problem for the case of fraud proofs, but this is more like pointing out that this issue can happen in other hotter code paths you write. This can easily become a contention problem if this was e.g. a msg processing.

fraud/service.go
Wondertan: Honestly, if `topics` meant to manage topics, it also makes sense to make a registration for the topics there. I mean `getTopic` could do this on its own, however, it does have access to the processIncoming...

fraud/subscription.go
```diff
@@ -0,0 +1,42 @@
...
+	return &subscription{sub, u}, nil
+}
+
+func (s *subscription) Proof(ctx context.Context) (Proof, error) {
+	data, err := s.subscription.Next(ctx)
+	if err != nil {
+		if err == context.Canceled {
+			return nil, err
+		}
+		return nil, fmt.Errorf("error during listening to the next proof: %s ", err.Error())
```
Wondertan, line 31: Next does not return any other errors besides context ones, so this is unnecessary.

header/sync/sync.go
```diff
@@ -45,14 +48,18 @@ type Syncer struct {
 	pending ranges
 	// cancel cancels syncLoop's context
 	cancel context.CancelFunc
+
+	// isRunning shows service state
+	isRunning uint64
```
Wondertan, line 53: For the Syncer this isn't actually needed if the goal is only to make Stop():
* Thread-safe
* Safe to be called multiple times. E.g. FP happened and then FX calls Stop

header/p2p/subscriber.go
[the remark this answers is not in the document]
Wondertan: `P2PSubscriber` already has a Start and Stop methods. Stop for it already implements unregistering, complete stop of the topic and this what we need. We can expose those on the interface instead of new `RemoveValidator` and call it on Syncer, as Syncer already keeps the subscriber.

header/sync/sync.go
Wondertan: Instead, we can remove setting `s.cancel` to nil in Stop + removing the check for nil `s.cancel` in Start.
It is also _safe_ to ignore the error coming from Subscriber.RemoveValidator or Subsciber.Stop(as suggested in another comment) in this case.

header/sync/sync.go
```diff
@@ -77,8 +85,13 @@ func (s *Syncer) Start(context.Context) error {
 
 // Stop stops Syncer.
 func (s *Syncer) Stop(context.Context) error {
-	s.cancel()
-	s.cancel = nil
+	if atomic.LoadUint64(&s.isRunning) == 1 {
+		// switch state to avoid getting in this statement twice
+		atomic.StoreUint64(&s.isRunning, 0)
```
Wondertan, line 90: Btw, this could be on CompareAndSwap, but it is better to implement the suggestion above.

Wondertan (changes requested):
Another round.

fraud/service.go
vgonkivs: package was added in interface.go : `var log = logging.Logger("fraud")`

fraud/service.go
vgonkivs: done

fraud/service.go
vgonkivs: Added `registerValidator` for topics

fraud/service.go
vgonkivs: yes, thanks.

fraud/service.go
vgonkivs: changed to `Lock()/Unlock()`

fraud/service.go
vgonkivs: unlocking a mutex after picking an unmarshaler from map

fraud/subscription.go
vgonkivs: removed

header/sync/sync.go
vgonkivs: Yes, good idea. Thanks.

vgonkivs: I removed flag as you proposed. So, now after receiving the BEFP syncer will call `Stop()` in P2PSubscriber

header/p2p/subscriber.go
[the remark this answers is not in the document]
vgonkivs: Removed this functionality at all. Instead, Syncer will call exposed Stop method from P2PSubscriber

service/state/service.go
renaynay: Aha! ack

fraud/service.go
```diff
@@ -0,0 +1,189 @@
...
+
+	"github.com/libp2p/go-libp2p-core/peer"
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
+)
+
+// headerFetcher aliases a function that is used to fetch ExtendedHeader from store
+type headerFetcher func(context.Context, uint64) (*header.ExtendedHeader, error)
+
+// topics allows to operate with pubsub connection and pubsub topics
+type topics struct {
```
Wondertan, line 19: nit: move to another file?

Wondertan (changes requested):
Getting closer

das/daser.go
```diff
@@ -143,6 +158,14 @@ func (d *DASer) sample(ctx context.Context, sub header.Subscription, checkpoint
 			if err == context.Canceled {
 				return
 			}
+			var byzantineErr *ipld.ErrByzantine
+			if errors.As(err, &byzantineErr) {
+				err := d.fService.Broadcast(ctx, fraud.CreateBadEncodingProof(uint64(h.Height), byzantineErr))
+				if err != nil {
+					log.Errorw("fraud proof propagating failed with error ", err)
+				}
```
adlerjohn, line 166: Maybe to leave for a future PR, but shouldn't this be asynchronous?

fraud/service.go
```diff
@@ -0,0 +1,189 @@
...
+			"err", err)
+		return pubsub.ValidationReject
+	}
+	log.Debugw("received Bad Encoding Fraud Proof from block ", "hash", hex.EncodeToString(extHeader.DAH.Hash()))
+	return pubsub.ValidationAccept
+}
+
+func getSubTopic(p ProofType) string {
+	return p.String() + "-sub"
+}
```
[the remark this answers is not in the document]
vgonkivs: We are using it multiple times, so I think it's better to keep it in one place to avoid typo issues

fraud/service.go
vgonkivs: Yes, I thought about that. Will extract into `topics.go`

das/daser.go
```diff
@@ -31,6 +36,9 @@ type DASer struct {
 
 	sampleDn  chan struct{} // done signal for sample loop
 	catchUpDn chan struct{} // done signal for catchUp loop
+
+	// isRunning shows service state
+	isRunning uint64
```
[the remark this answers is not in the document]
vgonkivs: Used uint to be able to change it atomically.

das/daser.go
vgonkivs: it could be async but I think it does not matted in this case, because we will stop sampling in case of befp.

fraud/bad_encoding.go
```diff
@@ -71,6 +71,12 @@ func (p *BadEncodingProof) MarshalBinary() ([]byte, error) {
 	return badEncodingFraudProof.Marshal()
 }
 
+// UnmarshalBEFP converts given data to BadEncodingProof
+func UnmarshalBEFP(data []byte) (Proof, error) {
+	befp := &BadEncodingProof{}
+	return befp, befp.UnmarshalBinary(data)
```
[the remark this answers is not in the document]
vgonkivs: Yes, but does not it modify in-place befp every time we call `befp.UnmarshalBinary(data)`?

fraud/bad_encoding.go
```diff
@@ -80,6 +86,8 @@ func (p *BadEncodingProof) UnmarshalBinary(data []byte) error {
 	befp := &BadEncodingProof{
 		BlockHeight: in.Height,
 		Shares:      ipld.ProtoToShare(in.Shares),
+		Index:       uint8(in.Index),
```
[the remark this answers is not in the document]
vgonkivs: Yes, it's quite limiting. will change the type.

fraud/service.go
```diff
@@ -0,0 +1,189 @@
...
+		return nil, errors.New("unmarshaler is not registered")
+	}
+
+	t, wasjoined, err := f.topics.getTopic(proofType)
+	if err != nil {
+		return nil, err
+	}
+	// if topic was joined for the first time then we should register a validator for it
+	if !wasjoined {
+		if err = f.topics.registerValidator(proofType, f.processIncoming); err != nil {
```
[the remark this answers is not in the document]
vgonkivs: Why? One liner `if else `statements are an idiomatic way in go to return errors, no?

fraud/bad_encoding.go
```diff
@@ -154,3 +162,26 @@ func (p *BadEncodingProof) Validate(header *header.ExtendedHeader) error {
...
+		return
+	}
+	defer subscription.Cancel()
+
+	log.Info("Start listening to bad encoding fraud proof")
+	// At this point we receive already verified fraud proof,
+	// so there are no needs to call Validate.
+	_, err = subscription.Proof(ctx)
+	if err != nil {
+		if err == context.Canceled {
```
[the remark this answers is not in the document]
vgonkivs: will remove this statement

fraud/service.go
vgonkivs: Yes, sorry. Missed that

fraud/service.go
```diff
@@ -0,0 +1,189 @@
...
+		return pubsub.ValidationReject
+	}
+
+	err = proof.Validate(extHeader)
+	if err != nil {
+		log.Errorw("validation err: ",
+			"err", err)
+		return pubsub.ValidationReject
+	}
+	log.Debugw("received Bad Encoding Fraud Proof from block ", "hash", hex.EncodeToString(extHeader.DAH.Hash()))
```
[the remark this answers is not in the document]
vgonkivs: Yeah, good point, thank you

das/daser.go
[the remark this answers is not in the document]
Wondertan: Yeah, there is no atomic bool in std yet. This will change pretty soon. 

@vgonkivs, you can use uint32 though

fraud/service.go
```diff
@@ -0,0 +1,189 @@
...
+func (f *service) Subscribe(proofType ProofType) (Subscription, error) {
+	// TODO: @vgonkivs check if fraud proof is in fraud store, then return with error
+	f.mu.RLock()
+	u, ok := f.unmarshalers[proofType]
+	f.mu.RUnlock()
+	if !ok {
+		return nil, errors.New("unmarshaler is not registered")
+	}
+
+	t, wasjoined, err := f.topics.getTopic(proofType)
```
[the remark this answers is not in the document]
vgonkivs: Then `getTopic` will have reverse logic for this flag and the code will be harder to read

fraud/bad_encoding.go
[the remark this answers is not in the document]
Wondertan: It does modify it.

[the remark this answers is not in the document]
Wondertan: You can use append like this in some cases, if you know the append rules.

header/sync/sync.go
```diff
@@ -78,7 +76,7 @@ func (s *Syncer) Start(context.Context) error {
 // Stop stops Syncer.
 func (s *Syncer) Stop(context.Context) error {
 	s.cancel()
-	s.cancel = nil
+	s.sub.Stop(context.Background()) //nolint:errcheck
```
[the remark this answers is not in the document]
vgonkivs: done

das/daser.go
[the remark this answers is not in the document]
vgonkivs: changed to uint32

fraud/bad_encoding.go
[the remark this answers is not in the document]
vgonkivs: done

fraud/bad_encoding.go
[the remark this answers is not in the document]
adlerjohn: It does modify it in place, it's just really not-obvious by looking at it. Knowing that is extra cognitive overhead that's not needed because you can just split it up into another line.

fraud/service.go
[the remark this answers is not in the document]
adlerjohn: image

das/daser.go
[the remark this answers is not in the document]
adlerjohn: Can you add a comment for why this is a `uint32` then?

fraud/bad_encoding.go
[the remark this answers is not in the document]
Wondertan: > Note that we've used append incorrectly in our own code in the past. That alone should disqualify return befp, befp.UnmarshalBinary(data).

I disagree here. The fact that we used _tabs_ incorrectly does not disqualify the usage of _spaces_.  Though I am not feeling strong about an additional line or not, I just think append is a different case, though it has common _subjective_ cognitive overhead property.

When I see something that requires me to put in additional cognitive efforts, my response to this is usually: "wow, that's cool or smart". This is why I meant subjective.

fraud/service.go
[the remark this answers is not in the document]
Wondertan: From what you can read in this code:
* `getTopic` gives you a flag `wasjoined`, which says if getting topic also joined to it or just got it
* Then, `!wasjoined,` or if _we did not join, we register the validator_

fraud/proof.go
```diff
@@ -16,7 +16,7 @@ const (
 func (p ProofType) String() string {
 	switch p {
 	case BadEncoding:
-		return "BadEncoding"
```
Wondertan, line 19: The "E" is still capitalized btw

fraud/bad_encoding.go
```diff
@@ -144,13 +155,38 @@ func (p *BadEncodingProof) Validate(header *header.ExtendedHeader) error {
...
+		return errors.New(
+			"fraud: invalid fraud proof. recomputed Merkle root matches the header's row/column root",
+		)
 	}
 
 	return nil
 }
+
+// SubscribeToBEFP listens to Bad Encoding Fraud Proof and stops services once it was received
+func SubscribeToBEFP(ctx context.Context, s Subscriber, stop func(context.Context) error) {
```
Wondertan, line 172: nit: What about calling this `OnBEFP` or `StopOnBEFP`? This is not really just `subscribe`ing here. But subscription until first proof + calling a stop function.

Wondertan (commented):
Also, for visibility posting here other are two things we are discussing privately:
* Spawn `fraud.SubscribeToBefp` routine in DI, Instead of passing fraud.Service to each service that needs to be stopped, and calling `fraud.SubscribeToBefp` on Start. Something like this:
```go 
das := das.NewDASer(...)
lc.Append(fx.Hook{
	  OnStart: das.Start,
	  OnStop:  das.Stop,
})
go fraud.SubscribeToBEFP(ctx, fraudSub, das.Stop)
```

This way we keep lifecycle code together and manage it from one place. Also, we reduce coupling between services that needs to be stopped and fraud service.

* Move `RegisterUnmarshaler` our of FraudService.Start to DI.
FraudService should not do any registering on its own. It provides an API for users to (un)register any Frauds they want, but it shouldn't do it itself. Going further, this Register and Unregister API can be entirely dropped out by generics, so no more runtime logic that can be done at compile time 🦀 
* Remove Start and Stop from FraudService interface as redundant.

fraud/bad_encoding.go
vgonkivs: `OnBEFP` sounds good to me.

fraud/proof.go
Wondertan: Oh, this comment landed on the deleted side of the diff. If you look the latest change there, you will see "badEncoding" with capitalized E in the middle.

das/daser.go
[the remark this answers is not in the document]
vgonkivs: added

fraud/bad_encoding.go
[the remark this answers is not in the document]
liamsi: > When I see something that requires me to put in additional cognitive efforts, my response to this is usually: "wow, that's cool or smart". This is why I meant subjective.

Without judging the above sample at hand: in general this might be cool or smart from an intellectual point of view, but we should aim for the most readable code without cognitive overhead (even if that is often boring).

fraud/service.go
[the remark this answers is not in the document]
liamsi: The second example of an if-statement in effective go is:
```go 
if err := file.Chmod(0664); err != nil {
    log.Print(err)
    return err
}
```
https://go.dev/doc/effective_go#if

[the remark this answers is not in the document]
Wondertan: ☝🏻

node/services/service.go
```diff
@@ -132,20 +136,30 @@ func ShareService(lc fx.Lifecycle, bServ blockservice.BlockService, avail share.
...
 		OnStop:  das.Stop,
 	})
+	go fraud.OnBEFP(fxutil.WithLifecycle(ctx, lc), fService, das.Stop)
 	return das
 }
 
+// FraudService constructs fraud proof service
+func FraudService(sub *pubsub.PubSub, hstore header.Store) (fraud.Service, fraud.Service) {
+	f := fraud.NewService(sub, hstore.GetByHeight)
+	f.RegisterUnmarshaler(fraud.BadEncoding, fraud.UnmarshalBEFP) //nolint:errcheck
```
Wondertan, line 159: Why no error check?

service/state/service.go
```diff
@@ -2,20 +2,29 @@ package state
...
 
 // Service can access state-related information via the given
 // Accessor.
 type Service struct {
 	accessor Accessor
+
+	fsub   fraud.Subscriber
+	cancel context.CancelFunc
+
+	befpReceived uint64
```
Wondertan, line 20: You don't really need it here and we can simplify this by relying on the context. as we already have cancel there. Also, I don't see this being used anywhere. There is `BEFPReceived` but is unused.

Instead of having the uint64 here(which should have been uint32 btw), we can cancel the context and check if it was canceled before doing anything with
```go
if err := ctx.Err(); err != nil {
	return err
}
```

So if you want to prevent users from Submitting new txs - add the if above to `SubmitTx`.
Also:
* Remove s.cancel = nil from Stop, so there are no race conditions
* Move out `go fraud.OnBEFP` to DI

node/services/service.go
```diff
@@ -24,16 +25,19 @@ import (
...
 	sub header.Subscriber,
+	fsub fraud.Subscriber,
 ) (*sync.Syncer, error) {
-	syncer := sync.NewSyncer(ex, store, sub)
+	syncer := sync.NewSyncer(ex, store, sub, fsub)
 	lc.Append(fx.Hook{
 		OnStart: syncer.Start,
 		OnStop:  syncer.Stop,
 	})
+	go fraud.OnBEFP(fxutil.WithLifecycle(ctx, lc), fsub, syncer.Stop)
```
Wondertan, line 40: ❤️ 

I didn't thing of `WithLifecycle` here

node/services/service.go
vgonkivs: Though it's not needed if we are registering on system start

service/state/service.go
vgonkivs: Done

node/services/service.go
vgonkivs: Added

service/rpc/middleware.go
```diff
@@ -0,0 +1,35 @@
...
+	rpc.RegisterMiddleware(setContentType)
+	rpc.RegisterMiddleware(checkAllowance(h.state))
+}
+
+func setContentType(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Add("Content-Type", "application/json")
+		next.ServeHTTP(w, r)
+	})
+}
```
Wondertan, line 22: 👍🏻

das/daser.go
```diff
@@ -229,6 +251,14 @@ func (d *DASer) catchUp(ctx context.Context, job *catchUpJob) (int64, error) {
...
 				return height - 1, nil
 			}
+			var byzantineErr *ipld.ErrByzantine
+			if errors.As(err, &byzantineErr) {
+				err := d.fService.Broadcast(ctx, fraud.CreateBadEncodingProof(uint64(h.Height), byzantineErr))
+				if err != nil {
+					log.Errorw("fraud proof propagating failed with error ", err)
+				}
+				return height, err
+			}
```
Wondertan, line 261: Now it seems even more that `sample` and `catchUp` can be DRYied. 
cc @renaynay. 

Orthogonal to the PR

das/daser.go
```diff
@@ -82,17 +94,19 @@ func (d *DASer) Start(context.Context) error {
...
+		// wait for both sampling routines to exit
+		for i := 0; i < 2; i++ {
+			select {
+			case <-d.catchUpDn:
+			case <-d.sampleDn:
+			case <-ctx.Done():
+				return ctx.Err()
+			}
 		}
+		d.cancel = nil
```
Wondertan, line 108: What if we rely on the context.Err() here is as well?

Wondertan:
```go
if d.ctx.Err() != nil {
	retrun nil
}
```

das/daser.go
```diff
@@ -143,6 +157,14 @@ func (d *DASer) sample(ctx context.Context, sub header.Subscription, checkpoint
 			if err == context.Canceled {
 				return
 			}
+			var byzantineErr *ipld.ErrByzantine
+			if errors.As(err, &byzantineErr) {
```
liamsi, line 161: I'm wondering if we shouldn't fix https://github.com/celestiaorg/celestia-node/issues/787 before proceeding with merging this PR. Because it's likely we will see these fake byzantineErrs popping up in the network and being gossiped to each other as well. 
At the very least we should hold off with cutting a release that contains fraud service without fixing #787 first.

fraud/topics.go
```diff
@@ -0,0 +1,61 @@
...
+
+// registerValidator adds an internal validation to topic inside libp2p for provided ProofType
+func (t *topics) registerValidator(
+	proofType ProofType,
+	val func(context.Context, ProofType, []byte) pubsub.ValidationResult,
+) error {
+	return t.pubsub.RegisterTopicValidator(
+		getSubTopic(proofType),
+		func(ctx context.Context, _ peer.ID, msg *pubsub.Message) pubsub.ValidationResult {
+			return val(ctx, proofType, msg.Data)
```
liamsi, line 58: Won't _-ing the peer.ID prevent us from e.g. rejecting proofs from a peer that sent us e.g. invalid proofs (and we blacklisted them). Or will that be handled by other parts of the code?

node/tests/fraud_test.go
```diff
@@ -0,0 +1,59 @@
...
+	require.NoError(t, err)
+	full := sw.NewFullNode(node.WithTrustedPeers(addrs[0].String()))
+
+	err = full.Start(ctx)
+	require.NoError(t, err)
+
+	subscr, err := full.FraudServ.Subscribe(fraud.BadEncoding)
+	require.NoError(t, err)
+	_, err = subscr.Proof(ctx)
+	require.NoError(t, err)
```
liamsi, line 51: Is it guaranteed that this returns a fraud proof if error == nil?

fraud/service.go
```diff
@@ -0,0 +1,122 @@
...
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
+)
+
+// service is propagating and validating Fraud Proofs
+// service implements Service interface.
+type service struct {
+	topics       *topics
+	unmarshalers map[ProofType]ProofUnmarshaler
+	getter       headerFetcher
+	mu           sync.RWMutex
```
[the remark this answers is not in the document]
Wondertan: I missed this somehow. Thanks

fraud/service.go
```diff
@@ -0,0 +1,122 @@
...
+}
```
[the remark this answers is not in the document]
Wondertan: +1 for the first

fraud/service.go
```diff
@@ -0,0 +1,122 @@
...
+		return pubsub.ValidationReject
+	}
+
+	err = proof.Validate(extHeader)
+	if err != nil {
+		log.Errorw("validation err: ",
+			"err", err)
+		return pubsub.ValidationReject
+	}
+	log.Warnw("received Bad Encoding Fraud Proof from block ", "hash", hex.EncodeToString(extHeader.DAH.Hash()))
```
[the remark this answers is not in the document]
Wondertan: It is possible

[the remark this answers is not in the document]
Wondertan: Let's do

fraud/testing.go
```diff
@@ -0,0 +1,51 @@
+package fraud
+
+import (
+	"context"
+
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
+)
+
+type DummyService struct {
+}
```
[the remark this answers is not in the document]
Wondertan: No need for now imho. Unless we want to inject custom logic

fraud/topics.go
Wondertan: This is handled by pubsub. It decreases the scroll of the peer sending it if validation result is error. However, it does not blacklists it completely and this is what we should probably do.

node/tests/fraud_test.go
vgonkivs: Yes. verification logic is done in `processIncoming`

fraud/bad_encoding.go
```diff
@@ -80,6 +89,8 @@ func (p *BadEncodingProof) UnmarshalBinary(data []byte) error {
 	befp := &BadEncodingProof{
 		BlockHeight: in.Height,
 		Shares:      ipld.ProtoToShare(in.Shares),
+		Index:       uint(in.Index),
```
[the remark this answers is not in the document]
vgonkivs: changed to `uint32`

fraud/service.go
```diff
@@ -0,0 +1,132 @@
...
+	"context"
+	"encoding/hex"
+	"errors"
+	"sync"
+
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
+)
+
+// service is propagating and validating Fraud Proofs
+// service implements Service interface.
```
renaynay, line 13:
```suggestion
// service is responsible for validating and propagating Fraud Proofs.
// It implements the Service interface.
```

fraud/service.go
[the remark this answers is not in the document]
renaynay: I would actually put it on top

```
mu
unmarshalers
```

fraud/service.go
```diff
@@ -0,0 +1,132 @@
...
+func (f *service) Subscribe(proofType ProofType) (Subscription, error) {
+	// TODO: @vgonkivs check if fraud proof is in fraud store, then return with error
+	f.mu.RLock()
+	u, ok := f.unmarshalers[proofType]
+	f.mu.RUnlock()
+	if !ok {
+		return nil, errors.New("fraud: unmarshaler is not registered")
+	}
+
+	t, wasNotjoined, err := f.topics.getTopic(proofType)
```
renaynay, line 43: i'd prefer idiomatic `ok` here

fraud/service.go
```diff
@@ -0,0 +1,132 @@
...
+	if err != nil {
+		return nil, err
+	}
+	// if topic was joined for the first time then we should register a validator for it.
+	if wasNotjoined {
```
renaynay, line 48: This concept is very confusing. If a topic gets added to `topics`, it must also be required that a validator is mounted onto the topic before it is registered on topics. The current design is quite messy.

fraud/service.go
```diff
@@ -0,0 +1,132 @@
...
+
+	if _, ok := f.unmarshalers[proofType]; ok {
+		return errors.New("fraud: unmarshaler is already registered")
+	}
+	f.unmarshalers[proofType] = u
+
+	return nil
+}
+
+func (f *service) UnregisterUnmarshaler(proofType ProofType) error {
```
renaynay, line 69: Again I do not know where this will be used and why it's even necessary. IMO, the things that go together should be grouped together and not allowed to be "registered" unless the entire package is there. That means

A topic needs a validator and needs an unmarshaler. Those things should be all required at the same time in order for the blob (topic+unmarshaler+validator) to be registered on the fraud service.

I think this part can be done better. Maybe lets have a call on it tomorrow.

header/testing.go
```diff
@@ -198,6 +200,45 @@ func RandBlockID(t *testing.T) types.BlockID {
 	return bid
 }
 
+// FraudMaker creates a custom ConstructFn that breaks the block at the given height
```
renaynay, line 203:
```suggestion
// FraudMaker creates a custom ConstructFn that breaks the block at the given height.
```

node/services/service.go
```diff
@@ -132,20 +136,32 @@ func ShareService(lc fx.Lifecycle, bServ blockservice.BlockService, avail share.
...
+	das := das.NewDASer(avail, sub, hstore, ds, fService)
 	lc.Append(fx.Hook{
 		OnStart: das.Start,
 		OnStop:  das.Stop,
 	})
+	go fraud.OnBEFP(fxutil.WithLifecycle(ctx, lc), fService, das.Stop)
 	return das
 }
 
+// FraudService constructs fraud proof service
```
renaynay, line 156:
```suggestion
// FraudService constructs fraud proof service.
```

node/components.go
```diff
@@ -69,6 +70,7 @@ func baseComponents(cfg *Config, store Store) fx.Option {
 		fx.Provide(services.HeaderService),
 		fx.Provide(services.HeaderStore),
 		fx.Invoke(services.HeaderStoreInit(&cfg.Services)),
+		fxutil.ProvideAs(services.FraudService, new(fraud.Service), new(fraud.Subscriber)),
```
renaynay, line 73: do you still need the two return values then from the constructor if you're doing this too?

renaynay (changes requested):
First go, need to revisit the proof abstraction portion later (preferably let's discuss on a call tomorrow)

das/daser.go
```diff
@@ -70,18 +78,21 @@ func (d *DASer) Start(context.Context) error {
...
 	// kick off sampling routine for recently received headers
-	go d.sample(dasCtx, sub, checkpoint)
+	go d.sample(d.ctx, sub, checkpoint)
 	return nil
 }
 
 // Stop stops sampling.
 func (d *DASer) Stop(ctx context.Context) error {
+	if d.ctx.Err() != nil {
+		return nil
```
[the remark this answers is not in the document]
vgonkivs: no. `ctx.Err()` in this case shows that context has been already cancelled and all routines has been already stopped thus we do not need to go further.

fraud/bad_encoding.go
```diff
@@ -147,10 +161,35 @@ func (p *BadEncodingProof) Validate(header *header.ExtendedHeader) error {
...
+	if err != nil {
+		log.Errorw("failed to subscribe on bad encoding fraud proof ", err)
+		return
+	}
+	defer subscription.Cancel()
+
+	log.Info("Start listening to bad encoding fraud proof")
+	// At this point we receive already verified fraud proof,
+	// so there are no needs to call Validate.
+	_, err = subscription.Proof(ctx)
```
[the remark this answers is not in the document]
vgonkivs: as far as I see only context error

fraud/interface.go
```diff
@@ -0,0 +1,51 @@
...
+	// interface and broadcasts it to all subscribed peers.
+	Broadcast(context.Context, Proof) error
+}
+
+// Subscriber encompasses the behavior necessary to
+// subscribe/unsubscribe from new FraudProofs events from the
+// network.
+type Subscriber interface {
+	// Subscribe allows to subscribe on pub sub topic by its type.
+	// Subscribe should register pub-sub validator on topic.
```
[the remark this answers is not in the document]
vgonkivs: Left from the previous version. Will remove this comment as we agreed with @Wondertan that we will not allow registering custom validator.

node/components.go
vgonkivs: Yes, Daser requires `fraud.Service` and Syncer + State/Service require` fraud.Subscriber`

fraud/service.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	err = proof.Validate(extHeader)
+	if err != nil {
+		log.Errorw("validation err: ",
+			"err", err)
+		f.topics.addPeerToBlacklist(msg.ReceivedFrom)
+		return pubsub.ValidationReject
+	}
+	log.Warnw("received Bad Encoding Fraud Proof from block ",
```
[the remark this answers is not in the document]
vgonkivs: nice catch

node/rpc/component.go
```diff
@@ -39,4 +39,5 @@ func Handler(
 ) {
 	handler := rpc.NewHandler(state, share, header)
 	handler.RegisterEndpoints(serv)
+	handler.RegisterMiddlewares(serv)
```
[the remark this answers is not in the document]
vgonkivs: Even if we register several of them?

das/daser.go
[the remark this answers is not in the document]
renaynay: But it should explicitly be `context.Canceled`?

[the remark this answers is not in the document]
Wondertan: No

das/daser.go
```diff
@@ -143,6 +154,14 @@ func (d *DASer) sample(ctx context.Context, sub header.Subscription, checkpoint
 			if err == context.Canceled {
 				return
 			}
+			var byzantineErr *ipld.ErrByzantine
```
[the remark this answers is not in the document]
Wondertan: This function can also touch parts of the code not introduced in this PR, so let's make it after we merge this. I think you will be in the best position to see how this can DRYied out when you start to do DASState.

fraud/proof.go
```diff
@@ -16,21 +16,21 @@ const (
...
+		return "badencoding"
```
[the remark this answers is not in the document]
Wondertan: or `BadEncoding`.

header/sync/sync.go
```diff
@@ -76,9 +73,9 @@ func (s *Syncer) Start(context.Context) error {
 }
 
 // Stop stops Syncer.
-func (s *Syncer) Stop(context.Context) error {
+func (s *Syncer) Stop(ctx context.Context) error {
 	s.cancel()
-	s.cancel = nil
+	s.sub.Stop(ctx) //nolint:errcheck
```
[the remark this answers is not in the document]
Wondertan: It is problematic here. @vgonkivs can you leave a comment there explaining why?

node/rpc/component.go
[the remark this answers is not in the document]
Wondertan: Middleware in someway implies multiple plural

node/components.go
Wondertan: Daser could now require a broadcaster only.

Wondertan: btw

fraud/service.go
Wondertan: I tend to agree here. Not sure how to make it better here besides integrating topics into the service.

fraud/proof.go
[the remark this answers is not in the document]
vgonkivs: or `badencoding` to be consistent with `header-sub` as I use it in pub-sub topic -> `badencoding-sub`

fraud/service.go
vgonkivs: I can suggest :

```
func (f *service) Subscribe(proofType ProofType) (Subscription, error) {
	// TODO: @vgonkivs check if fraud proof is in fraud store, then return with error
	f.mu.Lock()
	u, ok := f.unmarshalers[proofType]
	
	if !ok {
		return nil, errors.New("fraud: unmarshaler is not registered")
	}

       // returns only cached topic
	t, ok := f.topics.getTopic(proofType)
	if !ok {
		if err = f.topics.registerValidator(proofType, f.processIncoming); err != nil {
                        f.mu.Unlock()
			return nil, err
		}
                // register topic
	}
       f.mu.Unlock()
	return newSubscription(t, u)
}
```

header/sync/sync.go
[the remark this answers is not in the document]
vgonkivs: it seems to be ok. fx will stop all services and then will throw a warning/error in log

node/tests/fraud_test.go
```diff
@@ -0,0 +1,59 @@
...
+Pre-Requisites:
+- CoreClient is started by swamp
+Steps:
+1. Create a Bridge Node(BN) with broken extended header at height 10
+2. Start a BN
+3. Create a Full Node(FN) with a connection to BN as a trusted peer
+4. Start a FN
+5. Subscribe to BEFP and wait when it will be received
+6. Check FN is not synced to 15.
+15 is not available because DASer will be stopped after receiving BEFP.
```
Bidon15, line 29:
```suggestion
Note: 15 is not available because DASer will be stopped before reaching this height due to receiving BEFP.
```

node/tests/fraud_test.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/celestiaorg/celestia-node/fraud"
+	"github.com/celestiaorg/celestia-node/header"
+	"github.com/celestiaorg/celestia-node/node"
+	"github.com/celestiaorg/celestia-node/node/tests/swamp"
+)
+
+/*
+Test-Case: Broadcast Fraud Proof to different nodes
```
Bidon15, line 19: I don't quite understand who is broadcasting BEFP for this case. Could the naming be more detailed?

node/tests/fraud_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+	subscr, err := full.FraudServ.Subscribe(fraud.BadEncoding)
+	require.NoError(t, err)
+	_, err = subscr.Proof(ctx)
+	require.NoError(t, err)
+
+	// Since GetByHeight is a blocking operation for headers that is not received, we
+	// should set a timeout because all daser/syncer are stopped at this point
+	newCtx, cancel := context.WithTimeout(context.Background(), time.Millisecond*500)
+	t.Cleanup(cancel)
```
Bidon15, line 56: does cancel assignment here omit/dereferences the cancel above for the cleanup stage? 🧐 
Maybe a newCancel or smt similar

Bidon15 (commented):
👏

fraud/topics.go
Wondertan: The simplest thing we can do for now is [this](https://github.com/libp2p/go-libp2p-pubsub/blob/68cdae031b494ea8b48c4dc7fd9dac506f0ce70e/pubsub.go#L1335).

node/tests/fraud_test.go
vgonkivs: I think it's not a big problem, since Cleanup will be invoked after test finishes.

fraud/service.go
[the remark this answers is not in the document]
vgonkivs: done

fraud/service.go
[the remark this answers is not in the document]
vgonkivs: done

fraud/topics.go
vgonkivs: added

das/daser.go
```diff
@@ -2,35 +2,41 @@ package das
...
 
 var log = logging.Logger("das")
 
 // DASer continuously validates availability of data committed to headers.
 type DASer struct {
-	da   share.Availability
-	hsub header.Subscriber
+	da       share.Availability
+	hsub     header.Subscriber
+	fService fraud.Service
```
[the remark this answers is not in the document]
vgonkivs: As @Wondertan mentioned, now Daser can take only `fraud.Broadcaster`

das/daser.go
```diff
@@ -143,6 +154,14 @@ func (d *DASer) sample(ctx context.Context, sub header.Subscription, checkpoint
 			if err == context.Canceled {
 				return
 			}
+			var byzantineErr *ipld.ErrByzantine
+			if errors.As(err, &byzantineErr) {
+				err := d.fService.Broadcast(ctx, fraud.CreateBadEncodingProof(uint64(h.Height), byzantineErr))
+				if err != nil {
+					log.Errorw("fraud proof propagating failed with error ", err)
```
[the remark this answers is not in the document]
vgonkivs: move to a separate function

node/tests/fraud_test.go
vgonkivs: Yes, thanks. Updated test case name.

node/components.go
renaynay: My question is, then why do you do the `ProvideAs` inside baseComponents?

fraud/service.go
```diff
@@ -33,24 +32,29 @@ func NewService(p *pubsub.PubSub, getter headerFetcher) Service {
...
+			return nil, err
+		}
+		t, err = f.topics.join(proofType)
+		if err != nil {
+			f.mu.Unlock()
 			return nil, err
 		}
 	}
-
+	f.mu.Unlock()
```
renaynay, line 57: Since this is at the end anyway, you can just defer `f.mu.Unlock` after line 35 and remove the unlocks at line 48 and line 53

fraud/topics.go
```diff
@@ -11,29 +11,34 @@ import (
...
+
+	pubsub *pubsub.PubSub
 }
 
-// getTopic joins a pubsub.Topic if it was not joined before and returns it.
-func (t *topics) getTopic(proofType ProofType) (*pubsub.Topic, bool, error) {
+// join allows to subscribe on pubsub topic
+func (t *topics) join(
+	proofType ProofType,
+) (*pubsub.Topic, error) {
```
renaynay, line 23:
```suggestion
func (t *topics) join(proofType ProofType) (*pubsub.Topic, error) {
```

node/components.go
vgonkivs: Because different services require different interfaces that are based on `FraudService` and we should explicitly provide them to fx.

fraud/topics.go
```diff
@@ -11,29 +11,34 @@ import (
...
-			return nil, ok, err
-		}
-		log.Debugf("successfully subscibed to topic: %s", getSubTopic(proofType))
-		t.pubSubTopics[proofType] = topic
+
+	topic, err := t.pubsub.Join(getSubTopic(proofType))
+	if err != nil {
+		return nil, err
 	}
+	log.Debugf("successfully subscibed to topic: %s", getSubTopic(proofType))
```
renaynay, line 31:
```suggestion
log.Debugf("successfully subscribed to topic: %s", getSubTopic(proofType))
```

node/components.go
renaynay:
```diff
diff --git a/node/components.go b/node/components.go
index b7f185d..484febf 100644
--- a/node/components.go
+++ b/node/components.go
@@ -71,7 +71,8 @@ func baseComponents(cfg *Config, store Store) fx.Option {
 		fx.Provide(services.HeaderStore),
 		fx.Invoke(services.HeaderStoreInit(&cfg.Services)),
 		fx.Provide(services.HeaderSyncer),
-		fxutil.ProvideAs(services.P2PSubscriber, new(header.Broadcaster), new(header.Subscriber)),
+		fxutil.ProvideAs(services.P2PSubscriber, new(header.Broadcaster)),
+		fxutil.ProvideAs(services.P2PSubscriber, new(header.Subscriber)),
 		fx.Provide(services.HeaderP2PExchangeServer),
 		// p2p components
 		fx.Invoke(invokeWatchdog(store.Path())),
diff --git a/node/services/service.go b/node/services/service.go
index a65159f..3b0ac2c 100644
--- a/node/services/service.go
+++ b/node/services/service.go
@@ -38,13 +38,13 @@ func HeaderSyncer(
 }
 
 // P2PSubscriber creates a new p2p.Subscriber.
-func P2PSubscriber(lc fx.Lifecycle, sub *pubsub.PubSub) (*p2p.Subscriber, *p2p.Subscriber) {
+func P2PSubscriber(lc fx.Lifecycle, sub *pubsub.PubSub) *p2p.Subscriber {
 	p2pSub := p2p.NewSubscriber(sub)
 	lc.Append(fx.Hook{
 		OnStart: p2pSub.Start,
 		OnStop:  p2pSub.Stop,
 	})
-	return p2pSub, p2pSub
+	return p2pSub
 }
 
 // HeaderService creates a new header.Service.
```

fraud/service.go
```diff
@@ -0,0 +1,136 @@
...
+	if !ok {
+		return nil, errors.New("fraud: unmarshaler is not registered")
+	}
+
+	t, ok := f.topics.getTopic(proofType)
+
+	// if topic was not stored in cache, then we should register a validator and
+	// join pubsub topic.
+	if !ok {
+		err := f.topics.registerValidator(proofType, f.processIncoming)
```
renaynay, line 46: Better to register validator after joining

fraud/service.go
vgonkivs: In this case unlock will be called after `newSubscription`. But `newSubscription` subscribes to the given topic and it can take sometime. I think it's better to manually unlock mutex before calling `newSubscription`

node/components.go
vgonkivs: May be. @Wondertan wdyt?

Wondertan: @renaynay, the problem with the above approach is that it will create multiple instances of the same structure that is supposed to be a singleton. In other words, there will be two FraudServices and two P2PSubscribers hidden via different types.

I agree thought that this is not the prettiest approach, but lets not bikeshed into it very deep.

fraud/service.go
vgonkivs: done

vgonkivs:
As we discussed with @Wondertan and @renaynay , this PR will be splited into several as it became huge.
Moving it to draft until all parts will be merged

* [Fraud service interface](https://github.com/celestiaorg/celestia-node/pull/808)

Wondertan:
@vgonkivs, can we close this one already?

vgonkivs:
Closing this PR bc main parts of it are already merged

Issue #322, opened 2022-01-03
header: Consider extracting some of the compoenents into sub-packages


Wondertan:
`header` package became a little bulky with multiple implementations over the same interfaces. We should consider extracting some components intro separate sub-packages under `header`. Originally, this is brought up because of the [need](https://github.com/celestiaorg/celestia-node/pull/317#discussion_r777512503) to prepend log statements with something like "impl-iface: statement", which is ugly and easy to fail with consistency.

We can either create sub-package with a name after the interface with multiple implementations to include a separate file for each. Where structure would look like:
`header/exchange/core|p2p.go`.

Also we can have `header/core/exchange|subscriber.go`, where we create a sub-package with the name after implementation relation and with files named after interfaces.

PR #677, opened 2022-05-06
Move extheader into different package

### LINKED ISSUE #676: header: move ext header into different package
The purpose of moving Extended header into different package is to avoid cyclic dependency between fraud `Service` and `Syncer`:
* fraud service(which is from package fraud) is required to get ExtHeader(which is in package header) to verify fraud proof;
* `Syncer`(which is from package header) is required to have Subscription on BEFP;

After discussion with @Wondertan we think that it's the fastest and the easiest solution now.

vgonkivs:
Resolves #676


Wondertan:
We discussed this again with @vgonkivs, and it turns out that this change is pretty invasive, which we are not satisfied with. Instead, we found a better solution:
* move `services/header/...` to `header/`
* keep `services/header/service.go` that depends on `header/`
This way, we keep all the services still defined under `services` but with actual components/data strictures being defined into their own pkg in the root. While resolving the dep cycle, its is also improves coherence in out architecture. This is also part of https://github.com/celestiaorg/celestia-node/issues/322 effort.

PR #681, opened 2022-05-06
chore(ipld): minor cleanups, renaming and docs fixes

Wondertan:
Сontinuation of chores in `ipld` pkg
* Ensure retrival tests are random on each run
* Fix logs on Retriever
* Rename PutData -> AddShares
	* We work with the notion of shares, not data
	* We add them using adder thus Add and not put
* write.go -> add.go
	* Same reason, we add nodes using adder, not writing them
* read.go -> get.go
	* All the funcs in the file start with Get and renaming makes this consistent with the file name


renaynay (approved):
lgtm

liamsi (approved):

PR #682, opened 2022-05-06
feat(ipld): concurrent GetShares

### LINKED ISSUE #650: ipld: GetLeaves should return fetched shares even on error
We use GetLeaves to retrieve shares of any row or column. Currently, it returns either all the shares in the success case or nothing, even if some shares were retrieved. To increase the success rate of reconstruction, we have to change this to (1) request as many shares as possible(configurable amount of blocking goroutines) at the same time and (2) return fetched ones.

Wondertan:
The main goal here is to make share requesting non-sequential s.t. if one share is not available, it does not block others to be downloaded, as it was with the previous recursive GetLeaves approach. 

For more details, check out extensive comments on the implementation.

Closes #650 

P.S. Credits to simply amazing [workerpool](https://github.com/gammazero/workerpool) lib by @gammazero. It is written so well to prevent me from reinventing the wheel and writing my own worker pool 😅. The lib fits our use case ideally.


Wondertan:
Something is broken for 1x1 square size, investigating

ipld/get_shares.go
```diff
@@ -0,0 +1,101 @@
...
+
+// NumWorkersLimit sets global limit for workers spawned by GetShares.
+// GetShares could be called MaxSquareSize(128) times per data square each spawning up to 128/2 goroutines and
+// altogether this is 8192. Considering there can be N blocks fetched at the same time, e.g. during catching up data
+// from the past, we multiply this number by the amount of allowed concurrent data square fetches(NumConcurrentSquares).
+// NOTE: This value only limits amount of simultaneously running workers that are spawned as the load increases and
+// are killed, once the load declines. Thanks to simple yet powerful workerpool library by @gammazero.
+// TODO(@Wondertan): This assumes we have parallelized DASer implemented. Sync the values once it is shipped.
+// TODO(@Wondertan): Allow configuration of values.
+var NumWorkersLimit = MaxSquareSize * MaxSquareSize / 2 * NumConcurrentSquares
```
adlerjohn, line 20: Is there a reason not to declare these as constants?

ipld/get_shares.go
```diff
@@ -0,0 +1,101 @@
...
+	"sync"
+
+	"github.com/gammazero/workerpool"
+	"github.com/ipfs/go-cid"
+	format "github.com/ipfs/go-ipld-format"
+)
+
+// NumWorkersLimit sets global limit for workers spawned by GetShares.
+// GetShares could be called MaxSquareSize(128) times per data square each spawning up to 128/2 goroutines and
+// altogether this is 8192. Considering there can be N blocks fetched at the same time, e.g. during catching up data
```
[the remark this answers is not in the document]
Wondertan: True, will change this.

ipld/get_shares.go
Wondertan: So we at least have some control over those. Specifically, I played around with those in reconstruction tests.
`ipld.NumWorkersLimit=<num>`

Wondertan:
@tzdybal, it is slightly related to [GetSharesByNamespace](https://github.com/celestiaorg/celestia-node/issues/697), but not directly. Also, I would like someone to do a good review of concurrency logic, and I trust you on this

ipld/get_shares.go
adlerjohn: Seems like a leaky abstraction. Is there a better way to do that while guaranteeing that the values won't be changed accidentally by the code?

liamsi: I also do not think it is great to have this exported and mutable.

ipld/get_shares.go
```diff
@@ -0,0 +1,101 @@
...
+// from the past, we multiply this number by the amount of allowed concurrent data square fetches(NumConcurrentSquares).
+// NOTE: This value only limits amount of simultaneously running workers that are spawned as the load increases and
+// are killed, once the load declines. Thanks to simple yet powerful workerpool library by @gammazero.
+// TODO(@Wondertan): This assumes we have parallelized DASer implemented. Sync the values once it is shipped.
+// TODO(@Wondertan): Allow configuration of values.
+var NumWorkersLimit = MaxSquareSize * MaxSquareSize / 2 * NumConcurrentSquares
+
+// NumConcurrentSquares limits the amount of squares that are fetched concurrently/simultaneously.
+var NumConcurrentSquares = 8
```
liamsi, line 23: I think the same feedback applies here: https://github.com/celestiaorg/celestia-node/pull/682/files#r867424676

ipld/get_shares.go
```diff
@@ -0,0 +1,101 @@
...
+// (got all shares) or on context cancellation.
+//
+// It works concurrently by spawning workers in the pool which do one basic thing - block until data is fetched, s. t.
+// share processing is never sequential, and thus we request *all* the shares available without waiting for others to
+// finish. It is the required property to maximize data availability. As a side effect, we get concurrent tree traversal
+// reducing time to data time.
+//
+// GetShares heavily abuses the fact that the underlying data structure is a binary tree (bin-tree-feat),
+// so it's not suitable for anything else besides that.
+func GetShares(ctx context.Context, dag format.NodeGetter, root cid.Cid, shares int, put func(int, Share)) {
```
liamsi, line 39: Where is this tested?

ipld/get_shares.go
```diff
@@ -0,0 +1,101 @@
...
+// GetShares gets shares from either local storage, or, if not found, requests them from immediate/connected peers.
+// It puts them into the given 'put' func, does not return any error, and returns/unblocks only on success
+// (got all shares) or on context cancellation.
+//
+// It works concurrently by spawning workers in the pool which do one basic thing - block until data is fetched, s. t.
+// share processing is never sequential, and thus we request *all* the shares available without waiting for others to
+// finish. It is the required property to maximize data availability. As a side effect, we get concurrent tree traversal
+// reducing time to data time.
+//
+// GetShares heavily abuses the fact that the underlying data structure is a binary tree (bin-tree-feat),
```
liamsi, line 37: bin-tree-feat?

Wondertan: Look at the implementation comments. I've pointed out parts of the code that relies on binary tree properties with this tag

ipld/retriever.go
```diff
@@ -172,27 +176,26 @@ func (rs *retrieverSession) request(ctx context.Context, q *quadrant) {
...
-				rs.square[q.index(i, j)] = nd.RawData()[1+NamespaceSize:]
-			}
+			// and go get shares of left or the right side of the whole col/row axis
+			// the left or the right side of the tree represent some portion of the quadrant
+			// which we put into the rs.square share-by-share by calculating shares' indexes using q.index
+			GetShares(ctx, rs.dag, nd.Links()[q.x].Cid, size, func(j int, share Share) {
+				// the R lock here is *not* to protect rs.square from multiple concurrent shares writes
+				// but to avoid races between share writes and repairing attempts
+				// shares are written atomically in their own slice slot
+				rs.squareLk.RLock()
```
tzdybal, line 195: Super counter-intuitive, but great (ab)use of RWMutex :imp:

ipld/retriever.go
```diff
@@ -154,9 +156,11 @@ func (rs *retrieverSession) retrieve(ctx context.Context, q *quadrant) (*rsmt2d.
 	rs.request(ctx, q)
 
 	// try repair
+	// TODO: Avoid reimporting of the square which can potentially remove the requirement for the lock
+	rs.squareLk.Lock()
```
tzdybal, line 160: I'm a bit confused.
1. `rs.request` call in line 156 is synchronous (there is a `WaitGroup` inside). 
2. `put` func passed to `GetShares` is invoked by workerpool, but `GetShares` is also synchronous (there's `WaitGroup` inside).
3. therefore, before reaching this line, all "puts" should already be finished - or I'm missing something?
4. `retrieve` function is also synchronous, and it's only called [in a loop](https://github.com/celestiaorg/celestia-node/pull/682/files#diff-32233bd82783aec17000ced7c82d67d3626a9f344eccde8f6b1c67f7273bba7bR58) - so only one instance of `retrieve` is running at any given point of time.

All above means, that lock is not needed at all. Correct me if I'm wrong.

Wondertan: Believe me or not, I thought the same, but the race detector told me I was wrong and this locking fixed the thing. 

The race condition happens specifically when we reimport the square after the second repair. I also placed a todo that after rsmt2d update the race might disappear. To know exactly, we need to understand race detector better. The rsmt2d update is about importing a square once and repair with a method on the square. I can give more details about importing or else in rsmt2d if you are interested.

ipld/retriever.go
Wondertan: Yeah, it's smart :) and avoids contention. I hope comment explains it well enough as you get it. The reason why it's needed is below.

ipld/retriever.go
Wondertan: Happy that someone is finally reviewing the concurrency code😌

ipld/get_shares.go
```diff
@@ -0,0 +1,101 @@
...
+// so it's not suitable for anything else besides that.
```
liamsi, line 38:
```suggestion
// GetShares heavily (ab)uses the fact that the underlying data structure is a binary tree,
// so it's not suitable for anything else besides that. Parts on the implementation that rely 
// on this property are explicitly tagged with (bin-tree-feat).
```

ipld/get_shares.go
liamsi: Ah I see. Then we should change the comment slightly: https://github.com/celestiaorg/celestia-node/pull/682/files#r869729453

ipld/retriever.go
```diff
@@ -61,6 +62,7 @@ func (r *Retriever) Retrieve(ctx context.Context, dah *da.DataAvailabilityHeader
 			if byzErr := r.checkForByzantineError(ctx, dah, err); byzErr != nil {
 				return nil, byzErr
 			}
+			log.Warnw("not enough shares to reconstruct data square, requesting more...", "err", err)
```
renaynay, line 65: Why warn level?

ipld/get_shares.go
Wondertan: This comment is a nit, fixing it is time sink. It is value is super small comparting to what this PR tries to achieve. Pls better look at the implementaion and find issues with it. I also agree that the should not be a global var thing and this is why I made a TODO there before this converstation is started. I can change this now, but it would take more time than you might thing. The possible damage of this var being misused is negligible.

ipld/retriever.go
Wondertan: Because it not an error and because you need to warn node operator about unhappy case/code path. Usually there should not be any retries

ipld/get_shares.go
```diff
@@ -0,0 +1,101 @@
...
+				lnks := nd.Links()
+				if len(lnks) == 1 { // so we are almost there
+					// the reason why the comment on 'total' is lying, as each leaf has its own additional leaf(hack)
+					// so get it
+					nd, err := dag.Get(ctx, lnks[0].Cid)
+					if err != nil {
+						// again, we don't care
+						return
+					}
+					// ladies and gentlemen, we got em!
```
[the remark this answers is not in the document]
Wondertan: Nah, I prefer comments that are at least intersting to read

ipld/get_shares.go
Wondertan: Why move?

ipld/get_shares.go
```diff
@@ -0,0 +1,101 @@
+package ipld
```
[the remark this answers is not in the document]
Wondertan: Yes, I want to split those Gets into separate files eventually. I put GetShares into get originally, but I appeared to very bloated.

ipld/get_shares.go
Wondertan: The same TODO I made applies here as well

ipld/get_shares.go
```diff
@@ -0,0 +1,101 @@
...
+				for i, lnk := range lnks {
+					// send those to be processed
+					select {
+					case jobs <- &job{
+						id: lnk.Cid,
+						// calc position for children nodes (bin-tree-feat),
+						// s.t. 'if' above knows where to put a share
+						pos: j.pos*2 + i,
+					}:
+					case <-ctx.Done():
```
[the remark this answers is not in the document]
Wondertan: Order in `select` cases does not matter at all. `select` reads from one of the channels. The one that has some value first. First to come, first to be processed. If they are two or more channels ready to be read from, select chooses them in pseudo-random order.

ipld/get_shares.go
Wondertan: Ok

ipld/get_shares.go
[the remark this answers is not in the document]
Wondertan: Done

ipld/get_shares.go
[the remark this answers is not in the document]
Wondertan: Keeping both

Wondertan:
I have zero guesses why docker started to fail. Anyone?

ipld/retriever.go
renaynay: Okay if it's not a frequent issue, then fine with me. Ack

Wondertan:
Maybe this will fix it https://github.com/celestiaorg/celestia-node/pull/689

ipld/retriever.go
```diff
@@ -172,27 +176,26 @@ func (rs *retrieverSession) request(ctx context.Context, q *quadrant) {
...
-			nds, err := GetLeaves(ctx, rs.dag, nd.Links()[q.x].Cid, make([]format.Node, 0, size))
-			if err != nil {
-				log.Errorw("getting all the leaves", "root", root.String(), "err", err)
-				return
-			}
-			// fill leaves into the square
-			for j, nd := range nds {
-				rs.square[q.index(i, j)] = nd.RawData()[1+NamespaceSize:]
-			}
+			// and go get shares of left or the right side of the whole col/row axis
```
renaynay, line 188: You are really only getting shares of left or right side of the **row** axis, since you are using `nd.Links()[q.x].CID` correct?

renaynay (commented):
Not much else to say, lgtm since it's also passing retriever / reconstruction tests, 👍🏻

ipld/retriever.go
Wondertan: Yes and this was part of the previous PR. The left or right sides are identified by a quadrant we are downloading.

renaynay: Okay so can you change then the comment to say `go get shares of left or right side of the row axis` bc you're only using the row coordinate to determine this.

Wondertan: Well, it is also about columns actually

Wondertan: Check out code that creates quadrants

Wondertan: newQuadrants and check for coordinates swapping in there

ipld/get_shares.go
liamsi: It's not only moving but explicitly stating this: https://github.com/celestiaorg/celestia-node/pull/682#discussion_r869422592

ipld/get_shares.go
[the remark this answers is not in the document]
Wondertan: https://github.com/celestiaorg/celestia-node/issues/708

ipld/get_shares.go
```diff
@@ -0,0 +1,113 @@
...
+// there can be N blocks fetched at the same time, e.g. during catching up data
+// from the past, we multiply this number by the amount of allowed concurrent
+// data square fetches(NumConcurrentSquares).
+//
+// NOTE: This value only limits amount of simultaneously running workers that
+// are spawned as the load increases and are killed, once the load declines.
+// TODO(@Wondertan): This assumes we have parallelized DASer implemented.
+//  Sync the values once it is shipped.
+// TODO(@Wondertan): Allow configuration of values without global state.
+var NumWorkersLimit = MaxSquareSize * MaxSquareSize / 2 * NumConcurrentSquares
```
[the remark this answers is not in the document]
Wondertan: Good point. MaxSquareSize is just copied from the core, and NumSomethingLimit is just how I used to do naming for such things.

ipld/get_shares.go
Wondertan: Ah, ok

Wondertan:
Rebase on main

ipld/get_shares.go
Wondertan: I updated the comment with you suggestion manually, as its outdated by GH, also tagged you on the commit

ipld/retriever.go
Wondertan: @tzdybal was able to reproduce the race locally to prove that lock is required, so I am resolving the issue

ipld/get_shares.go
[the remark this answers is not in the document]
renaynay: ack

ipld/retriever.go
renaynay: ack

renaynay (approved):

liamsi (approved):
I did not go through the subtleties of the locks and the concurrent code but it seems @tzdybal has you covered here 🙏🏼

PR #685, opened 2022-05-06
header: sub services to root

vgonkivs:


header/headerexchange/core_exchange.go
```diff
@@ -1,17 +1,21 @@
...
 	format "github.com/ipfs/go-ipld-format"
+	logging "github.com/ipfs/go-log/v2"
 
 	tmbytes "github.com/tendermint/tendermint/libs/bytes"
 
 	"github.com/celestiaorg/celestia-node/core"
+	"github.com/celestiaorg/celestia-node/header"
 )
 
+var log = logging.Logger("header_exchanger")
```
renaynay, line 17:
```suggestion
var log = logging.Logger("header/exchange")
```

header/headerstore/store.go
```diff
@@ -1,16 +1,23 @@
...
 	lru "github.com/hashicorp/golang-lru"
 	"github.com/ipfs/go-datastore"
 	"github.com/ipfs/go-datastore/namespace"
 
 	tmbytes "github.com/tendermint/tendermint/libs/bytes"
+
+	"github.com/celestiaorg/celestia-node/header"
 )
 
+var log = logging.Logger("header_store")
```
renaynay, line 19:
```suggestion
var log = logging.Logger("header/store")
```

header/headerstore/store.go
```diff
@@ -22,8 +29,12 @@ var (
 	DefaultWriteBatchSize = 2048
 )
 
-// errStoppedStore is returned for attempted operations on a stopped store
-var errStoppedStore = errors.New("header: stopped store")
+var (
+	// errStoppedStore is returned for attempted operations on a stopped store
+	errStoppedStore = errors.New("header_store: stopped store")
```
renaynay, line 34:
```suggestion
	errStoppedStore = errors.New("header/store: stopped store")
```

header/headerstore/store.go
```diff
@@ -22,8 +29,12 @@ var (
...
+	// errNoHead is returned when Store is empty (does not contain any known header).
+	errNoHead = fmt.Errorf("header_store: no chain head")
```
renaynay, line 36:
```suggestion
	errNoHead = fmt.Errorf("header/store: no chain head")
```

header/headersync/sync.go
```diff
@@ -7,10 +7,16 @@ import (
...
 
+	logging "github.com/ipfs/go-log/v2"
+
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	tmbytes "github.com/tendermint/tendermint/libs/bytes"
+
+	"github.com/celestiaorg/celestia-node/header"
 )
 
+var log = logging.Logger("header_sync")
```
renaynay, line 18:
```suggestion
var log = logging.Logger("header/sync")
```

header/interface.go
```diff
@@ -63,9 +63,6 @@ var (
 	// ErrNotFound is returned when there is no requested header.
 	ErrNotFound = errors.New("header: not found")
 
-	// ErrNoHead is returned when Store is empty (does not contain any known header).
-	ErrNoHead = fmt.Errorf("header/store: no chain head")
-
 	// ErrNonAdjacent is returned when Store is appended with a header not adjacent to the stored head.
 	ErrNonAdjacent = fmt.Errorf("header/store: non-adjacent")
```
renaynay, line 67: this should also go in `header/store` pkg errors, no?

node/node.go
```diff
@@ -20,7 +20,7 @@ import (
 	"github.com/celestiaorg/celestia-node/das"
 	"github.com/celestiaorg/celestia-node/node/rpc"
 	"github.com/celestiaorg/celestia-node/params"
-	"github.com/celestiaorg/celestia-node/service/header"
+	headerservice "github.com/celestiaorg/celestia-node/service/header"
```
renaynay, line 23: why do you need to rename import and can't leave it at just `header` ?

service/header/service.go
```diff
@@ -4,27 +4,31 @@ import (
...
 
 	logging "github.com/ipfs/go-log/v2"
+
+	"github.com/celestiaorg/celestia-node/header"
+	"github.com/celestiaorg/celestia-node/header/headerexchange"
+	"github.com/celestiaorg/celestia-node/header/headersync"
 )
 
-var log = logging.Logger("header")
+var log = logging.Logger("header_service")
```
renaynay, line 13:
```suggestion
var log = logging.Logger("header/service")
```

service/header/service.go
```diff
@@ -47,8 +51,8 @@ func (s *Service) Stop(context.Context) error {
 
 // GetByHeight returns the ExtendedHeader at the given height, blocking
 // until header has been processed by the store or context deadline is exceeded.
-func (s *Service) GetByHeight(ctx context.Context, height uint64) (*ExtendedHeader, error) {
-	return s.syncer.store.GetByHeight(ctx, height)
+func (s *Service) GetByHeight(ctx context.Context, height uint64) (*header.ExtendedHeader, error) {
+	return s.syncer.GetByHeight(ctx, height)
```
renaynay, line 55: general nit: I find it weird that this method is really on the store but we call it via `syncer`.

renaynay (changes requested):
Few comments: 

* I am against just moving `service/header/doc.go` to pkg `header` because it contains docs for how flow of header service works. Can you please split the docs for the actual header struct itself from the rest of the docs and have a `doc.go` in each pkg that contains only info about whats inside the pkg? 
* pls rename `header/headerstore` to `header/store`
* pls rename `header/headerexchange` to `header/exchange`
* pls rename `header/headersync` to `header/sync`
* pls split up `header/interface.go` into separate per-pkg interface.go files that contain only interface relevant to that pkg (so `Exchange` interface belongs in file `interface.go` inside of `exchange` pkg)  -- unless this is not possible due to cyclical dependency

header/interface.go
vgonkivs: no , because it is will lead to a cyclic dependence errors.

service/header/service.go
vgonkivs: It is done because service does not have `store` instance. We can change the implementation and request it from network using `Exchange` . @Wondertan wdyt?

renaynay: Why don't we have store inside `header.Service` again @Wondertan ?

header/interface.go
renaynay: ack

renaynay (approved):
Awesome, this was a long overdue refactoring. Thank you @vgonkivs !

service/header/service.go
Wondertan: Well, we should get it from the store directly without syncer.

Wondertan: @vgonkivs, the idea about Exchange is good, but not for now. We can make simple logic, if not found in store, requests via Exchange, but we anyway need to check the store for the asked header.

liamsi (approved):
Only one nit: I think now that we have a sub-packages we could shorten some names. e.g. `header.ExtendedHeader` could be just `header.Extended` imo. The package name provides enough context.
I would not change this in this PR as it can be discussed in a separate PR.

Something I would look at before merging is how the go docs look locally after the refactoring (run `godoc -http=:6060` and look at the rendered docs in your browser ideally).

Wondertan (changes requested):
Yay! Finally, we have already wanted this for a long time and are happy that we found a good hard reason to justify this refactor. 

A few more things: 
* lets split `exchange` pkg into two `p2p` and `core`
* rename P2PExchange to Exchange(so it's p2p.Exchange)
* rename CoreExchange to Exchange(so it's core.Exchange)
* Double check log statements so they don't have excessive info, i.e remove "core:"
* Rename files to `p2p/exchange.go`, etc

The main reason to do so, besides cleaner separation of concerns, is to avoid importing redundant pkgs for binaries. So that light client importing p2p.Exchange has nothing to do with core.Exchange and does not import it.

node/services/service.go
```diff
@@ -67,13 +72,13 @@ func HeaderExchangeP2P(cfg Config) func(params.Network, host.Host) (header.Excha
...
 		}
-		return header.NewP2PExchange(host, ids), nil
+		return p2p.NewExchange(host, ids), nil
 	}
 }
 
-// HeaderP2PExchangeServer creates a new header.P2PExchangeServer.
-func HeaderP2PExchangeServer(lc fx.Lifecycle, host host.Host, store header.Store) *header.P2PExchangeServer {
-	p2pServ := header.NewP2PExchangeServer(host, store)
+// HeaderP2PExchangeServer creates a new header.ExchangeServer.
```
Wondertan, line 79:
```suggestion
// HeaderP2PExchangeServer creates a new header/p2p.ExchangeServer.
```

node/core/core.go
```diff
@@ -10,7 +10,8 @@ import (
 	"github.com/celestiaorg/celestia-node/libs/fxutil"
 
 	"github.com/celestiaorg/celestia-node/core"
-	"github.com/celestiaorg/celestia-node/service/header"
+	"github.com/celestiaorg/celestia-node/header"
+	coreexchange "github.com/celestiaorg/celestia-node/header/core"
```
Wondertan, line 14:
```suggestion
	headercore "github.com/celestiaorg/celestia-node/header/core"
```

node/core/core.go
```diff
@@ -29,8 +30,8 @@ func DefaultConfig() Config {
 func Components(cfg Config) fx.Option {
 	return fx.Options(
 		fx.Provide(core.NewBlockFetcher),
-		fxutil.ProvideAs(header.NewCoreExchange, new(header.Exchange)),
-		fx.Invoke(HeaderCoreListener),
+		fxutil.ProvideAs(coreexchange.NewExchange, new(header.Exchange)),
```
Wondertan, line 33:
```suggestion
		fxutil.ProvideAs(headercore.NewExchange, new(header.Exchange)),
```

node/core/core.go
Wondertan: Why? Because header/core pkg can have other core related stuff and not necessarily exchange.

header/header.go
```diff
@@ -5,16 +5,24 @@ import (
...
-	header_pb "github.com/celestiaorg/celestia-node/service/header/pb"
+	header_pb "github.com/celestiaorg/celestia-node/header/pb"
 )
 
+var log = logging.Logger("header")
+
+// PubSubTopic hardcodes the name of the ExtendedHeader
+// gossipsub topic.
+const PubSubTopic = "header-sub"
+
```
Wondertan, line 25: I believe this should be part of p2p pkg

renaynay:
@vgonkivs if you look at the files changed, linter is recommending some nice suggestions for type renaming

renaynay (approved):
lfg

Wondertan (changes requested):
A few more things:
* ExtendedHeaderRequest is p2p Exchange only thing, so we should move it to p2p. Similarly, we have to put its pb form into p2p/pb
* I don't like LocalExchange being in core. We should either make it in a separate pkg or move it into header
*

node/core/core.go
vgonkivs: done

header/p2p/pb/extended_header_request.proto
```diff
@@ -0,0 +1,9 @@
+syntax = "proto3";
+
+package p2p.pb;
+
+message ExtendedHeaderRequest {
+  uint64 origin = 1;
+  bytes hash = 2;
+  uint64 amount = 3;
+}
```
Wondertan, line 9:
```suggestion
}

```

Wondertan (commented):
Almost there

header/p2p/pb/extended_header_request.proto
Wondertan: New line at the end of file

header/local/exchange.go
```diff
@@ -0,0 +1,48 @@
+package local
```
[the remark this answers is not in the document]
vgonkivs: don't know. used `git mv` as always...

Wondertan (approved):
Lgtm

PR #693, opened 2022-05-09
service/rpc: Add a separate `rpc` service to handle all exposed endpoints

renaynay:
# Description
This PR refactors the current design of how endpoints are made available on celestia-node. Now, instead of having per-package endpoints, this PR introduces a `service/rpc` package that contains a new `rpc.Handler` type designed to register all available endpoints on the `rpc.Server`. This makes it possible to make it so that shares can be retrieved just by a `namespaceID` and a block `height`. 

# Endpoints

### GET `/header/{height}`
Returns the header of the given height.

Request

```
curl -X GET http://<ip>:26658/header/300
```

Response 

```json
[...]

PR #695, opened 2022-05-09
service/share: Remove unused `Service` interface and replace it with struct instead

renaynay:
This PR removes the unused `Service` interface and replaces it with the struct instead.


Wondertan:
@renaynay, pls base it on top of #693, so that this PR does not show changes from it

node/rpc/component.go
```diff
@@ -2,18 +2,34 @@ package rpc
...
 			OnStart: serv.Start,
 			OnStop:  serv.Stop,
 		})
 		return serv
 	}
 }
+
+func HandlerComponents(
+	state *state.Service,
+	share share.Service,
```
vgonkivs, line 28: should not it also be a pointer?

Wondertan:
Because of the work, iirc
> Github prevents it

renaynay:
Closing in favour of #707 due to hairy merge conflicts.

PR #696, opened 2022-05-09
cmd: Set rpc flag defaults

renaynay:
This PR adds flag defaults for rpc addr and port.

### DIFF
--- a/cmd/flags_rpc.go
+++ b/cmd/flags_rpc.go
@@ -5,6 +5,7 @@ import (
 	flag "github.com/spf13/pflag"
 
 	"github.com/celestiaorg/celestia-node/node"
+	"github.com/celestiaorg/celestia-node/node/rpc"
 )
 
 var (
@@ -18,12 +19,12 @@ func RPCFlags() *flag.FlagSet {
 
 	flags.String(
 		addrFlag,
-		"0.0.0.0",
+		rpc.DefaultConfig().Address,
 		"Set a custom RPC listen address (default: localhost)",
 	)
 	flags.String(
 		portFlag,
-		"26658",
+		rpc.DefaultConfig().Port,
 		"Set a custom RPC port (default: 26658)",
 	)
 
--- a/cmd/flags_rpc.go
+++ b/cmd/flags_rpc.go
@@ -5,7 +5,6 @@ import (
 	flag "github.com/spf13/pflag"
 
 	"github.com/celestiaorg/celestia-node/node"
-	"github.com/celestiaorg/celestia-node/node/rpc"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


cmd/flags_rpc.go
```diff
@@ -18,12 +18,12 @@ func RPCFlags() *flag.FlagSet {
...
 	flags.String(
 		addrFlag,
-		"",
+		"0.0.0.0",
 		"Set a custom RPC listen address (default: localhost)",
 	)
 	flags.String(
 		portFlag,
-		"",
+		"26658",
```
tzdybal, line 26: Value is hardcoded in 2 places - here and in https://github.com/celestiaorg/celestia-node/blob/62697fd56ec2b732b3a43afaf9bcad61d139263f/node/rpc/config.go#L12

tzdybal: `DefaultConfig()` can be reused here, as a source of "defaults".

renaynay: https://github.com/celestiaorg/celestia-node/pull/696/commits/05163a55c4196b1db89e6f5ea39298691e179fd5

cmd/flags_rpc.go
```diff
@@ -18,12 +19,12 @@ func RPCFlags() *flag.FlagSet {
 
 	flags.String(
 		addrFlag,
-		"",
+		rpc.DefaultConfig().Address,
 		"Set a custom RPC listen address (default: localhost)",
```
adlerjohn, line 23: Can `rpc.DefaultConfig().Address` not be used to replace `localhost` in this string?

Wondertan: +1
I don't think importing config from flags is a good idea.

Wondertan (changes requested):

cmd/flags_rpc.go
adlerjohn: Oh I meant the string `"Set a custom RPC listen address (default: localhost)"`, not `""`. Are we talking about the same thing?

adlerjohn (commented):
Not sure about using writing two literals 3 times each now!

Wondertan:
@renaynay, what was the problem that this PR fixes? Why default config is not enough?

Wondertan:
@renaynay, what was the problem that this PR fixes? Why default config is not enough?

The problem with this approach is that changing the config in the folder won't affect the node as flags have a higher priority than config. The default value for the flag will always make it as it has some value and produce options overriding the loaded config on the node. So after this PR is merged, we will have another issue.

Wondertan:
@renaynay, closing this one until we find a better solution. We need an issue, though. Would be good if you or @tzdybal can explain it

renaynay:
@Wondertan https://github.com/celestiaorg/celestia-node/issues/718

PR #700, opened 2022-05-11
fix(cmd): use ~ instead of $HOME in paths

### LINKED ISSUE #699: cmd: `$HOME` used in keyring paths
Environment variables are not automatically resolved in paths, so using `$HOME` is an error (directory called `$HOME` is created in working directory).
All the keyring methods works, because all of them use the wrong directory.

Simple fix is to use `~` instead of `$HOME`.

I found 3 locations:
https://github.com/celestiaorg/celestia-node/blob/7bb20a3cab9423b9b4869500712e730538a63723/cmd/celestia/light.go#L16
https://github.com/celestiaorg/celestia-node/blob/7bb20a3cab9423b9b4869500712e730538a63723/cmd/celestia/full.go#L16
https://github.com/celestiaorg/celestia-node/blob/7bb20a3cab9423b9b4869500712e730538a63723/cmd/celestia/bridge.go#L16

tzdybal:
Resolves #699.

### DIFF
--- a/cmd/celestia/bridge.go
+++ b/cmd/celestia/bridge.go
@@ -13,7 +13,7 @@ import (
 // parent command.
 
 func init() {
-	bridgeKeyCmd := keys.Commands("$HOME/.celestia-bridge/keys")
+	bridgeKeyCmd := keys.Commands("~/.celestia-bridge/keys")
 	bridgeKeyCmd.Short = "Manage your bridge node account keys"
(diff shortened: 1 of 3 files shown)


renaynay (approved):

vgonkivs (approved):

Wondertan (approved):
I actually pointed to this in the PR where this were introduced, but Rene understood me differently

PR #705, opened 2022-05-12
header: introduce ConstructFn type and integrate it into core.Listener and core.Exchange

### LINKED ISSUE #704: header: allow setting function that will create ExtendedHeader in header services
Currently our header creation is done by `MakeExtendedHeader` once new block is received(or block is requested by its hash). This restricts us from writing different tests(for example where  the header can be broken to verify the Fraud Service). 

As we discussed with @Wondertan , we can pass the exact function that will create our header and thus gives us flexibility to write different test cases.

vgonkivs:
Resolves #704

### DIFF
--- a/header/core/exchange.go
+++ b/header/core/exchange.go
@@ -19,12 +19,14 @@ var log = logging.Logger("header/core")
 type Exchange struct {
 	fetcher    *core.BlockFetcher
 	shareStore format.DAGService
+	maker      header.Maker
 }
 
-func NewExchange(fetcher *core.BlockFetcher, dag format.DAGService) *Exchange {
+func NewExchange(fetcher *core.BlockFetcher, dag format.DAGService, maker header.Maker) *Exchange {
 	return &Exchange{
 		fetcher:    fetcher,
 		shareStore: dag,
+		maker:      maker,
 	}
 }
 
@@ -65,7 +67,7 @@ func (ce *Exchange) RequestByHash(ctx context.Context, hash tmbytes.HexBytes) (*
 		return nil, err
 	}
 
-	eh, err := header.MakeExtendedHeader(ctx, block, comm, vals, ce.shareStore)
+	eh, err := ce.maker(ctx, block, comm, vals, ce.shareStore)
 	if err != nil {
 		return nil, err
 	}
@@ -94,5 +96,5 @@ func (ce *Exchange) getExtendedHeaderByHeight(ctx context.Context, height *int64
 		return nil, err
 	}
 
-	return header.MakeExtendedHeader(ctx, b, comm, vals, ce.shareStore)
+	return ce.maker(ctx, b, comm, vals, ce.shareStore)
 }
--- a/header/core/exchange.go
+++ b/header/core/exchange.go
@@ -19,14 +19,14 @@ var log = logging.Logger("header/core")
 type Exchange struct {
 	fetcher    *core.BlockFetcher
 	shareStore format.DAGService
-	generator  header.Generator
+	construct  header.ConstructFn
 }
 
-func NewExchange(fetcher *core.BlockFetcher, dag format.DAGService, generator header.Generator) *Exchange {
+func NewExchange(fetcher *core.BlockFetcher, dag format.DAGService, construct header.ConstructFn) *Exchange {
 	return &Exchange{
 		fetcher:    fetcher,
 		shareStore: dag,
-		generator:  generator,
+		construct:  construct,
 	}
 }
 
@@ -67,7 +67,7 @@ func (ce *Exchange) RequestByHash(ctx context.Context, hash tmbytes.HexBytes) (*
 		return nil, err
 	}
 
-	eh, err := ce.generator(ctx, block, comm, vals, ce.shareStore)
+	eh, err := ce.construct(ctx, block, comm, vals, ce.shareStore)
 	if err != nil {
 		return nil, err
 	}
@@ -96,5 +96,5 @@ func (ce *Exchange) getExtendedHeaderByHeight(ctx context.Context, height *int64
 		return nil, err
 	}
 
-	return ce.generator(ctx, b, comm, vals, ce.shareStore)
+	return ce.construct(ctx, b, comm, vals, ce.shareStore)
 }
--- a/header/core/exchange.go
+++ b/header/core/exchange.go
@@ -19,14 +19,14 @@ var log = logging.Logger("header/core")
 type Exchange struct {
 	fetcher    *core.BlockFetcher
 	shareStore format.DAGService
-	maker      header.Maker
+	generator  header.Generator
 }
 
-func NewExchange(fetcher *core.BlockFetcher, dag format.DAGService, maker header.Maker) *Exchange {
+func NewExchange(fetcher *core.BlockFetcher, dag format.DAGService, generator header.Generator) *Exchange {
 	return &Exchange{
 		fetcher:    fetcher,
 		shareStore: dag,
-		maker:      maker,
+		generator:  generator,
 	}
 }
 
@@ -67,7 +67,7 @@ func (ce *Exchange) RequestByHash(ctx context.Context, hash tmbytes.HexBytes) (*
 		return nil, err
 	}
 
-	eh, err := ce.maker(ctx, block, comm, vals, ce.shareStore)
+	eh, err := ce.generator(ctx, block, comm, vals, ce.shareStore)
 	if err != nil {
 		return nil, err
 	}
@@ -96,5 +96,5 @@ func (ce *Exchange) getExtendedHeaderByHeight(ctx context.Context, height *int64
 		return nil, err
 	}
 
-	return ce.maker(ctx, b, comm, vals, ce.shareStore)
+	return ce.generator(ctx, b, comm, vals, ce.shareStore)
 }
--- a/header/interface.go
+++ b/header/interface.go
@@ -11,7 +11,7 @@ import (
 	core "github.com/tendermint/tendermint/types"
 )
 
-// Maker aliases a func that creates ExtendedHeader.
+// Maker aliases a function that creates an ExtendedHeader.
 type Maker = func(
 	context.Context,
 	*core.Block,
--- a/header/interface.go
+++ b/header/interface.go
@@ -11,8 +11,8 @@ import (
 	core "github.com/tendermint/tendermint/types"
 )
 
-// Maker aliases a function that creates an ExtendedHeader.
-type Maker = func(
+// Generator aliases a function that creates an ExtendedHeader.
+type Generator = func(
 	context.Context,
 	*core.Block,
 	*core.Commit,
--- a/header/interface.go
+++ b/header/interface.go
@@ -11,6 +11,7 @@ import (
 	core "github.com/tendermint/tendermint/types"
 )
 
+// Maker aliases a func that creates ExtendedHeader.
 type Maker = func(
 	context.Context,
 	*core.Block,
--- a/header/interface.go
+++ b/header/interface.go
@@ -11,8 +11,8 @@ import (
 	core "github.com/tendermint/tendermint/types"
 )
 
-// Generator aliases a function that creates an ExtendedHeader.
-type Generator = func(
+// ConstructFn aliases a function that creates an ExtendedHeader.
+type ConstructFn = func(
 	context.Context,
 	*core.Block,
 	*core.Commit,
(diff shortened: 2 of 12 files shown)


node/testing.go
```diff
@@ -27,3 +34,43 @@ func TestNode(t *testing.T, tp Type, opts ...Option) *Node {
 	require.NoError(t, err)
 	return nd
 }
+
+// WithFaultMaker is used to enable test cases with invalid blocks
```
Bidon15, line 38: we also need to explicitly provide info on how the invalid blocks are created. Either it is from block N onwards. Or if it is the 1 of the M blocks produced.

node/testing.go
```diff
@@ -27,3 +34,43 @@ func TestNode(t *testing.T, tp Type, opts ...Option) *Node {
...
+		sets.opts = append(sets.opts, fx.Replace(faultMaker))
+	}
+}
+
+func faultMaker(ctx context.Context,
+	b *types.Block,
+	comm *types.Commit,
+	vals *types.ValidatorSet,
+	dag format.NodeAdder) (*header.ExtendedHeader, error) {
+	log.Info("using fault header maker...")
```
Bidon15, line 50: in the dynamic set of fault headers, we need more explicit info log (can be included in the issue creation described below 👇 in the next comment review line)

Bidon15 (commented):
Awesome idea with the fault maker 💪

node/testing.go
```diff
@@ -27,3 +34,43 @@ func TestNode(t *testing.T, tp Type, opts ...Option) *Node {
...
+	var dah da.DataAvailabilityHeader
+
+	namespacedShares, _ := b.Data.ComputeShares()
+
+	extended, err := ipld.AddShares(ctx, namespacedShares.RawShares(), dag)
+	if err != nil {
+		return nil, err
+	}
+
+	if b.Height == int64(1) {
```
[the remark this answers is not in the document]
vgonkivs: done

header/interface.go
```diff
@@ -5,10 +5,20 @@ import (
 	"errors"
 	"fmt"
 
+	format "github.com/ipfs/go-ipld-format"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	tmbytes "github.com/tendermint/tendermint/libs/bytes"
+	core "github.com/tendermint/tendermint/types"
 )
 
+type Maker = func(
```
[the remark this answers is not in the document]
vgonkivs: added

node/testing.go
vgonkivs: added

Wondertan:
PR Naming suggestion/improvement: `header: introduce Maker type and integrate into core.Listener and core.Exchange`. There is also `FaultMaker` involved here and it seems now it could be part of a separate PR with its own PR, but it's fine.

ipld/test_helpers.go
```diff
@@ -66,3 +71,16 @@ func RandShares(t *testing.T, total int) []Share {
...
+	batchAdder := NewNmtNodeAdder(
+		ctx,
+		format.NewBatch(ctx, na, format.MaxSizeBatchOption(int(size)*2)),
+	)
+	tree := wrapper.NewErasuredNamespacedMerkleTree(size/2, nmt.NodeVisitor(batchAdder.Visit))
+	attackerEDS, _ := rsmt2d.ImportExtendedDataSquare(shares, consts.DefaultCodec(), tree.Constructor)
+	_ = batchAdder.Commit()
+
+	return da.NewDataAvailabilityHeader(attackerEDS)
+}
```
Wondertan, line 86: * We must not ignore the error from the Commit and Import!!!
* If you make it as a testhelper, then pass the `t` and check for errors with t. Pls never ignore errors
* There is no need to pass EDS size as we have it in the `shares` already = `len(shares)^2`
* Use ipld.DefaultCodec.
* There is already a constructor for DAH in core/da pkg, which makes one from the EDS.
  * Lets better make this func similar to the `AddShares` as they are basically the same besides that one does rsmt2d.Compute and this one rsmt2d.Import and probably not make it a testhelper but a regular func that returns an error.  So we should call it ImportShares, place it below AddShares and make it return EDS.

node/testing.go
```diff
@@ -27,3 +34,41 @@ func TestNode(t *testing.T, tp Type, opts ...Option) *Node {
 	require.NoError(t, err)
 	return nd
 }
+
+// WithFaultMaker is used to enable test cases with invalid blocks
+func WithFaultMaker(faultHeight int64) Option {
```
Wondertan, line 39: `node.WithFaultMaker` name does not explain what exactly is getting faulted. Add Header to the option name

node/testing.go
```diff
@@ -27,3 +34,41 @@ func TestNode(t *testing.T, tp Type, opts ...Option) *Node {
...
+			eh := &header.ExtendedHeader{
+				RawHeader:    b.Header,
+				DAH:          &dah,
+				Commit:       comm,
+				ValidatorSet: vals,
+			}
+			return eh, nil
+		}))
+	}
+}
```
Wondertan, line 74: When I proposed an idea to make an option to set _custom_ Maker I meant setting a _custom_ Maker, not specifically a Faulty Maker.  To be honest, It feels strange to me to make an option specifically for one test case in the Swamp, even though it can be later enabled in the Testgound as well(cc @Bidon15).

Let's:
* Make an option `WithHeaderMaker(header.Maker)` which allows setting a custom header.Maker via fx.Replace. It should not be part of the testing.go in the node pkg, but a regular option.
* Extract the faultMaking logic into `header/testing.go` as `FraudMaker`, so that (1) you can import it in the Swamp test, @Bidon15 can import it in testgound tests and (2) its clear by the name that this maker causes FraudProof, not some arbitrary fault.

Wondertan (changes requested):

ipld/test_helpers.go
vgonkivs: done

ipld/add.go
```diff
@@ -35,6 +35,28 @@ func AddShares(ctx context.Context, shares []Share, adder ipld.NodeAdder) (*rsmt
 	return eds, batchAdder.Commit()
 }
 
+// ImportShares imports flattend chunks of data into Extended Data square and saves it in IPLD DAG
+func ImportShares(ctx context.Context, shares [][]byte, na ipld.NodeAdder) (*rsmt2d.ExtendedDataSquare, error) {
+	if len(shares) == 0 {
+		return nil, fmt.Errorf("empty data") // empty block is not an empty Data
```
renaynay, line 41: what does this comment mean?

ipld/add.go
```diff
@@ -35,6 +35,28 @@ func AddShares(ctx context.Context, shares []Share, adder ipld.NodeAdder) (*rsmt
...
+	// create the nmt wrapper to generate row and col commitments
+	tree := wrapper.NewErasuredNamespacedMerkleTree(uint64(squareSize/2), nmt.NodeVisitor(batchAdder.Visit))
+	// recompute the eds
+	eds, err := rsmt2d.ImportExtendedDataSquare(shares, DefaultRSMT2DCodec(), tree.Constructor)
+	if err != nil {
+		return nil, fmt.Errorf("failure to recompute the extended data square: %w", err)
+	}
+	// compute roots
+	eds.RowRoots()
+	// commit the batch to ipfs
```
renaynay, line 56:
```suggestion
	// commit the batch to DAG
```

renaynay (commented):
minor nits but looks awesome!

header/core/exchange.go
```diff
@@ -19,12 +19,14 @@ var log = logging.Logger("header/core")
 type Exchange struct {
 	fetcher    *core.BlockFetcher
 	shareStore format.DAGService
+	maker      header.Maker
```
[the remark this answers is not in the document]
vgonkivs: I'm ok with this.

node/testing.go
vgonkivs: FaultMaker will be added in a different PR.

ipld/add.go
```diff
@@ -35,6 +35,28 @@ func AddShares(ctx context.Context, shares []Share, adder ipld.NodeAdder) (*rsmt
 	return eds, batchAdder.Commit()
 }
 
+// ImportShares imports flattend chunks of data into Extended Data square and saves it in IPLD DAG
+func ImportShares(ctx context.Context, shares [][]byte, na ipld.NodeAdder) (*rsmt2d.ExtendedDataSquare, error) {
+	if len(shares) == 0 {
+		return nil, fmt.Errorf("empty data")
```
Wondertan, line 41:
```suggestion
		return nil, fmt.Errorf("ipld: importing empty data")
```

Wondertan (commented):
LGTM.

@renaynay, I don't really get the Extender thing and we should discuss this. It is not clear what are we extending to call it Extender. There is one process of extending within our project and this is an extending of square. There is also an ExtendedHeader which means it has some additional value to the base Core header. To what Extending we are referring here specifically and if both at the same time, isn't this confusing?  

Previous Maker made sense to me because Maker is a synonym for the constructor. And it was constructing new Header. Now, it is not really clear.

ipld/add.go
vgonkivs: Alternatively, I can suggest `Converter` as it converts shares from` core.Block` to `ExtendedHeader`

Wondertan: Converting is about transforming something from one form into another. Here we convert multiple inputs into one, which is not really converting, and the resulting Header is not a cumulative of all the inputs, as it does not store has shares which are passed to the func in the form of a block. I would propose `header.Commiter` instead. It is not that broad as `Maker` and semantically close to blockchain/crypto terminology.

Wondertan: cc @renaynay

renaynay: I am in favour of either 

* `Generator`
* `Constructor` 

Either is fine with me bc encompasses the general thing that's happening within the function.

Wondertan: The function does much more stuff to be called `Constructor`, but `Generator` is fine. I would still be more precise with `Commiter` as it also involves what `Generator` means in some way and also makes it more connected to what we are doing. Basically, Bridge Node that executes this function commits to the correctness of the Header and distributes it on the network so others can verify the commitment/header. 

cc @liamsi, WDYT?

vgonkivs: `Commiter` and `Generator` are both ok for me

renaynay: `Generator`, lets just ship. I am not in favour of `Committer`. If we both like `Generator`, lets just compromise and use.

vgonkivs: Changed to generator

renaynay (approved):
👍🏻

ipld/add.go
liamsi: I'd even call it sth like `header.ConstructExtendedFn` or `header.CreateExtendedFn`(even the comment above the type definition says "a function that creates an ExtendedHeader" so the latte would call the type accordingly). You can skip the `Fn` suffix.


Just regarding the nuances of the verbs used here: 
generate/create is to bring something into being
construct is to build or form (something) by assembling parts

liamsi: Hlib mentioned to me that this is planned to become an interface. then the name should follow golang naming conventions for interfaces: 
> By convention, one-method interfaces are named by the method name plus an -er

IMO, we can do this once this will be an interface (and not necessarily in this PR).

renaynay: I personally like `Fn` suffix. It's descriptive and good 👍🏻 

I'm in favour of `header.ConstructExtendedFn`.

renaynay: `WithHeaderConstructorFn` but let's get @Wondertan approval first

Wondertan: Ok, let's do `header.ConstructFn`

vgonkivs: renamed

renaynay (approved):

node/settings.go
```diff
@@ -70,3 +71,10 @@ func WithCoreClient(client core.Client) Option {
 		sets.opts = append(sets.opts, fxutil.ReplaceAs(client, new(core.Client)))
 	}
 }
+
+// WithHeaderConstructorFn sets custom func that creates extended header
+func WithHeaderConstructorFn(construct header.ConstructFn) Option {
```
Wondertan, line 76:
```suggestion
// WithHeaderConstructFn sets custom func that creates extended header
func WithHeaderConstructFn(construct header.ConstructFn) Option {
```

Wondertan (approved):
Apply the naming nit suggestion above, and we are ready to go

Bidon15 (approved):
Agree with Hlib's comment above 👆

PR #707, opened 2022-05-13
service/share: Remove unused `share.Service` interface

renaynay:
This PR removes the `share.Service` interface as it doesn't have any use and exports the previously private `service` struct instead, using it throughout the codebase.

Overrides #695

### DIFF
--- a/service/share/share.go
+++ b/service/share/share.go
@@ -35,9 +35,18 @@ var GetData = ipld.ShareData
 // In practice, it is a commitment to all the Data in a square.
 type Root = da.DataAvailabilityHeader
 
-// Service provides a simple interface to access any data square or block share on the network.
+// NewService creates new basic share.Service.
+func NewService(dag format.DAGService, avail Availability) *Service {
+	return &Service{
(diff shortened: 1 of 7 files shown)


service/share/share.go
```diff
@@ -35,9 +35,18 @@ var GetData = ipld.ShareData
...
-// Service provides a simple interface to access any data square or block share on the network.
+// NewService creates new basic share.Service.
+func NewService(dag format.DAGService, avail Availability) *Service {
+	return &Service{
+		rtrv:         ipld.NewRetriever(dag, ipld.DefaultRSMT2DCodec()),
+		Availability: avail,
+		dag:          dag,
+	}
+}
+
```
Wondertan, line 46: nit: Move it under the Service struct?

Wondertan (approved):

liamsi (approved):

service/share/share.go
renaynay: done

adlerjohn (approved):
utACK

Issue #415, opened 2022-02-08
[Feature Request]: Keys Management


Bidon15:
### Summary

Celestia Node can CRUD keys

### Impact on users

It will be beneficial to produce a feature that allows Celestia Node to do CRUD on keys as this can allow signing Txs. 

Currently, we have a merged ADR https://github.com/celestiaorg/celestia-node/pull/369 that is proposing a state interaction for Celestia Node and a respective PR https://github.com/celestiaorg/celestia-node/pull/411

### Implementation ideas

As we are already using RPC approach lately, we can use it for this feature

### Urgency

Medium

liamsi:
The cosmos-sdk uses this for key storage: https://github.com/cosmos/cosmos-sdk/blob/master/crypto/keyring/doc.go
And this is also what lens uses under the hood I think.

liamsi:
I think independent of lens we should be consistent and use SDK's keyring.

renaynay:
We already do with #420

Wondertan:
While reviewing #420 I went deeper into SDK's keyring. My observations are:
* They have some custom migration logic for amino and stuff that we don't need
* They introduce a keyring interface that is overloaded, not ergonomic and can be simplified
	* They treat Ledger as something that needs additional methods rather than an implementation detail
	* `SupportedAlgorithms() (SigningAlgoList, SigningAlgoList)` two lists instead of one?
	* `Exporter` interface whose responsibility is only ASCII armoring shouldn't really be an interface.
		* There can be only `GetPubKey` and `GetPrivKey` methods with armoring as separate utility functions and not an interface.
		* These getters will also remove the need for an `Unsafe` interface.
		* Importers could also use generic setters/writers with armoring treated elsewhere. Basically, armoring is a (de)serialization format that should not be a part of the interface.
	* Methods for migration 
	* Multisig special methods can be implementation details
	* Why wallets are indexed by bother uid and address? Why only address or uid is not enough?
		* There might be a good reason for this, but it seems like a legacy
* They use github.com/99designs/keyring which looks sane. 
	* The SDK's Keyring interface does not look sane and long term we should write our own generic wallet management which may wrap 99designs/keyring
	* We should look deeper into extending 99designs/keyring backend implementations with Ledger as a backend and possibly other hardware wallets. If its Keyring interface allows us to do so then we can remove our own Keystore interface, If not we should look for other generic wallets and/or improve on our Keystore
	
cc @liamsi @renaynay

[...]

PR #712, opened 2022-05-16
node/state: Generate key for node if no key found in keystore path

renaynay:
This PR adds logic to check if a key was found in the given keystore path, and if not, a new key is generated on spot for the node to use for state interaction, allowing for better UX upon node start up.

New logs: 

```
2022-05-16T11:04:36.200+0100	INFO	state-access-constructor	state/core.go:42	NO KEY FOUND IN PATH, GENERATING NEW KEY...	{"path": "/Users/<>/.celestia-full/keys"}
2022-05-16T11:04:36.208+0100	INFO	state-access-constructor	state/core.go:48	NEW KEY GENERATED...

NAME: celes
ADDRESS: celes1gflh3mcttn02w7ahhjuucw8ajt69zu2g43wl9y
MNEMONIC (save this somewhere safe!!!):
present disorder solve luxury junk army lamp toward motor hello butter rate oppose parade absurd turtle person arrive tomato junk noise script mass rocket

```

### DIFF
--- a/node/state/core.go
+++ b/node/state/core.go
@@ -1,6 +1,7 @@
 package state
 
 import (
+	"github.com/cosmos/cosmos-sdk/crypto/hd"
 	"os"
 
 	"github.com/cosmos/cosmos-sdk/crypto/keyring"
@@ -31,9 +32,23 @@ func CoreAccessor(endpoint string) func(fx.Lifecycle, keystore.Keystore, params.
 			return nil, err
 		}
 		signer := apptypes.NewKeyringSigner(ring, keyringAccName, string(net))
+		keys, err := signer.List()
+		if err != nil {
+			return nil, err
+		}
+		// if no key was found in keystore path, generate new key for node
+		if len(keys) == 0 {
+			log.Infow("NO KEY FOUND IN PATH, GENERATING NEW KEY...", "path", ks.Path())
+			info, mn, err := signer.NewMnemonic(keyringAccName, keyring.English, "", "",
+				hd.Secp256k1)
+			if err != nil {
+				return nil, err
+			}
+			log.Infow("NEW KEY GENERATED...", "name", info.GetName(), "mnemonic", mn)
+		}
 
 		log.Infow("constructed keyring signer", "backend", keyring.BackendTest, "path", ks.Path(),
-			"keyring account name", keyringAccName)
+			"keyring account name", keyringAccName, "chain-id", string(net))
 
 		ca := state.NewCoreAccessor(signer, endpoint)
 		lc.Append(fx.Hook{
--- a/node/state/core.go
+++ b/node/state/core.go
@@ -16,13 +16,12 @@ import (
 	"github.com/celestiaorg/celestia-node/service/state"
 )
 
-var (
-	log = logging.Logger("state-access-constructor")
+var log = logging.Logger("state-access-constructor")
 
-	keyringAccName = "celes"
-)
-
-func CoreAccessor(endpoint string) func(fx.Lifecycle, keystore.Keystore, params.Network) (state.Accessor, error) {
+func CoreAccessor(
+	endpoint,
+	keyringAccName string,
+) func(fx.Lifecycle, keystore.Keystore, params.Network) (state.Accessor, error) {
 	return func(lc fx.Lifecycle, ks keystore.Keystore, net params.Network) (state.Accessor, error) {
 		// TODO @renaynay: Include option for setting custom `userInput` parameter with
 		//  implementation of https://github.com/celestiaorg/celestia-node/issues/415.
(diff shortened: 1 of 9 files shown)


node/state/core.go
```diff
@@ -31,9 +33,25 @@ func CoreAccessor(endpoint string) func(fx.Lifecycle, keystore.Keystore, params.
 			return nil, err
 		}
 		signer := apptypes.NewKeyringSigner(ring, keyringAccName, string(net))
+		keys, err := signer.List()
+		if err != nil {
+			return nil, err
+		}
+		// if no key was found in keystore path, generate new key for node
+		if len(keys) == 0 {
```
Wondertan, line 41: That's a working way to check if there are no keys, but is it possible to directly check if `keyringAccName` is present or no? Technically, there could be a case where the user copied some keys into the folder, and the keyring may see it here and does not go to the generation code path.

node/state/core.go
```diff
@@ -31,9 +33,25 @@ func CoreAccessor(endpoint string) func(fx.Lifecycle, keystore.Keystore, params.
...
+			log.Infow("NO KEY FOUND IN PATH, GENERATING NEW KEY...", "path", ks.Path())
```
Wondertan, line 42:
```suggestion
			log.Infow("NO KEY FOUND IN STORE, GENERATING NEW KEY...", "path", ks.Path())
```
I know that we hardcode path right now for the key storage, but we shouldn't do so and tell users that we store key in the store of the node which can be anywhere users wants. 
(Now it makes sense to differentiate between store and repository btw)

node/state/core.go
```diff
@@ -31,9 +33,25 @@ func CoreAccessor(endpoint string) func(fx.Lifecycle, keystore.Keystore, params.
...
+			info, mn, err := signer.NewMnemonic(keyringAccName, keyring.English, "", "",
+				hd.Secp256k1)
+			if err != nil {
+				return nil, err
+			}
+			log.Info("NEW KEY GENERATED...")
+			fmt.Printf("\nNAME: %s\nADDRESS: %s\nMNEMONIC (save this somewhere safe!!!): \n%s\n\n",
```
Wondertan, line 49: Should we print mnemonic? It is basically printing a private key to stdout or even worse, as its multiple private keys.

renaynay: Key utility prints it anyway

renaynay: It's a fresh-unfunded account.

renaynay: What's the alternative here?

node/state/core.go
renaynay: done

node/state/core.go
Wondertan: Not printing it? Writing to some file? TBH not sure, maybe it fine for now

node/state/core.go
Wondertan: ?

Wondertan: Why not just signer.Key(keyringAccName) != nil(simplified) ?

evan-forbes (commented):
nice! I think this should work just fine for now. 

We've discussed this in the past, but in the long run we probably just need a different keyring that is specifically for signing things programmatically. If we combine that with the authz module, then I think we can reach a really good UX and still be very secure.

this gets my approval, but will defer to final approval to other reviewers

Wondertan (approved):
Ok then, a bit more of ugliness, but its all over the place in cosmos :)

node/state/core.go
adlerjohn: Writing it to an unencrypted file is probably less secure that printing it out on screen unencrypted. Writing to an encrypted file isn't a bad idea. Either way, can be left to later.

liamsi: I think in the SDK the user can chose if the mnemonic is printed or not.

node/state/core.go
```diff
@@ -31,9 +33,39 @@ func CoreAccessor(endpoint string) func(fx.Lifecycle, keystore.Keystore, params.
...
+		if err != nil {
+			return nil, err
+		}
+		// check if key exists with prefix `celes`
+		var (
+			exists = false
+			name   = ""
+		)
+		for _, key := range keys {
+			if key.GetName()[:5] == keyringAccName {
```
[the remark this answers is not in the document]
renaynay: Okay how about this flow @liamsi @adlerjohn: 

* expose `--keyring.AccName` flag
* If user sets this flag during `init` or `start`, keyring constructor will try and find key prefixed by the given name, and uses it as the node's default key. If the keyring constructor cannot find a key prefixed by the given name, it will **fail out**.
* If the flag is **not set**, the default keyringAccName will be `""` such that the keyring constructor will look for keys in the keystore path. 
  * If **no keys are found**, constructor will construct a fresh new key and log that it has done so. 
  * If **several keys are found**, constructor will use the first key from the key list `keys[0]` and will use that key as the node's default account. (It will always be logged which key name the node is using as default key for node's account).
  * If one key is found, constructor will use it.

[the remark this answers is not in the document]
adlerjohn: How can I create a key with the prefix? If the only two options are fail if no key found with the given prefix or create a key with an empty prefix, then I can't ever create a key with the given prefix.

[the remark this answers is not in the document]
renaynay: you can do `celestia <node_type> keys add <name_of_key> --keyring-backend test`

and then `celestia <node_type> <init || start> --keyring.AccName <name_of_key>`

If the key is created for user, the name will be `my_celes_key`.

adlerjohn (approved):
utACK

node/state/core.go
renaynay: For now, we will log the mnemonic. If people want higher security for testnet, they can load their own keys into the path.

renaynay: Can I resolve @liamsi @adlerjohn

adlerjohn: Fine for now

renaynay: we have an issue that tracks keyring issues https://github.com/celestiaorg/celestia-node/issues/415

PR #717, opened 2022-05-16
params: Introduce `CUSTOM` network type

renaynay:
This PR introduces the ability to set a custom network type, genesis hash, and bootstrappers via setting environment variables. If a custom network is specified, the following warning will be printed for the user: 

```
WARNING: Celestia custom network specified. Only use this option if the node is freshly created and initialized.
**DO NOT** run a custom network over an already-existing node store!
```

*Note*: the terms `chain-id` and `network name/ID` are used interchangeably.

### ENV VAR DOCS

* `CELESTIA_CUSTOM_GENESIS="<chain-id>=<genesis_hash>"` adds the new network to the registry with the given genesis hash
* `CELESTIA_CUSTOM_NETWORK="<chain-id>"` picks the given network from the registry (it must already exist inside the registry)
* `CELESTIA_CUSTOM_BOOTSTRAPPERS="<chain-id>=<comma-separated list of boostrappers>"` adds or overrides bootstrappers for the given network in the registry

### Example start command

This command will start a celestia full node on the custom network `MyCustomNetwork`, with the genesis hash `D001B097E894F94B37F8FB0A5B8185812F5FBF8A07E6B0BCB560D647CAFB709B` and the bootstrapper `/dns4/andromeda.celestia-devops.dev/tcp/2121/p2p/12D3KooWKvPXtV1yaQ6e3BRNUHa5Phh8daBwBi3KkGaSSkUPys6D`.

```
CELESTIA_CUSTOM_GENESIS="MyCustomNetwork=D001B097E894F94B37F8FB0A5B8185812F5FBF8A07E6B0BCB560D647CAFB709B"  CELESTIA_CUSTOM_BOOTSTRAPPERS="MyCustomNetwork=/dns4/andromeda.celestia-devops.dev/tcp/2121/p2p/12D3KooWKvPXtV1yaQ6e3BRNUHa5Phh8daBwBi3KkGaSSkUPys6D" CELESTIA_CUSTOM_NETWORK="MyCustomNetwork" ./build/celestia full start
```

Based on some logic from https://github.com/celestiaorg/celestia-node/pull/552

### DIFF
--- a/node/init.go
+++ b/node/init.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/celestiaorg/celestia-node/libs/fslock"
 	"github.com/celestiaorg/celestia-node/libs/utils"
+	"github.com/celestiaorg/celestia-node/params"
 )
 
 // Init initializes the Node FileSystem Store for the given Node Type 'tp' in the directory under 'path' with
@@ -118,7 +119,7 @@ func initDir(path string) error {
 	if utils.Exists(path) {
 		// if the dir already exists and `CELESTIA_CUSTOM` env var is set,
 		// fail out to prevent store corruption
-		if _, ok := os.LookupEnv("CELESTIA_CUSTOM"); ok {
+		if _, ok := os.LookupEnv(params.EnvCustomNetwork); ok {
 			return fmt.Errorf("cannot run a custom network over an already-existing node store")
 		}
 		return nil
--- a/node/init.go
+++ b/node/init.go
@@ -2,6 +2,7 @@ package node
 
 import (
 	"errors"
+	"fmt"
 	"os"
 	"path/filepath"
 
@@ -115,6 +116,11 @@ func initRoot(path string) error {
 // initDir creates a dir if not exist
 func initDir(path string) error {
 	if utils.Exists(path) {
+		// if the dir already exists and `CELESTIA_CUSTOM` env var is set,
+		// fail out to prevent store corruption
+		if _, ok := os.LookupEnv("CELESTIA_CUSTOM"); ok {
+			return fmt.Errorf("cannot run a custom network over an already-existing node store")
+		}
 		return nil
 	}
 	return os.Mkdir(path, perms)
--- a/params/default.go
+++ b/params/default.go
@@ -16,48 +16,30 @@ func DefaultNetwork() Network {
 
 func init() {
 	// check if custom network option set
-	if customNet, ok := os.LookupEnv("CELESTIA_CUSTOM_GENESIS"); ok {
+	if custom, ok := os.LookupEnv("CELESTIA_CUSTOM"); ok {
 		fmt.Print("\n\nWARNING: Celestia custom network specified. Only use this option if the node is " +
 			"freshly created and initialized.\n**DO NOT** run a custom network over an already-existing node " +
 			"store!\n\n")
-
-		params := strings.Split(customNet, "=")
-		// ensure both params are present
-		if len(params) != 2 {
-			panic("must provide CELESTIA_CUSTOM_GENESIS in this format: <network_ID>=<genesis hash>")
+		// ensure all three params are present
+		params := strings.Split(custom, ":")
+		if len(params) != 3 {
+			panic("must provide CELESTIA_CUSTOM in this format: " +
+				"<network_ID>:<genesis hash>:<comma-separated list of bootstrappers>")
 		}
-		netID, genHash := params[0], params[1]
-
-		// register new network and set as default for node to use
-		networksList[Network(netID)] = struct{}{}
-		genesisList[defaultNetwork] = strings.ToUpper(genHash)
-	}
-	// check if a custom network was specified
-	if network, ok := os.LookupEnv("CELESTIA_CUSTOM_NETWORK"); ok {
-		if err := Network(network).Validate(); err != nil {
-			panic("unknown network specified")
-		}
-		defaultNetwork = Network(network)
-	}
-	// check if custom bootstrappers were provided for a network
-	if bootstrappers, ok := os.LookupEnv("CELESTIA_CUSTOM_BOOTSTRAPPERS"); ok {
-		params := strings.Split(bootstrappers, "=")
-		// ensure both params are present
-		if len(params) != 2 {
-			panic("must provide CELESTIA_CUSTOM_BOOTSTRAPPERS in this format: " +
-				"<network_ID>=<boostrappers comma separated list>")
-		}
-
-		netID, list := params[0], params[1]
-		// validate correctness
-		bs := strings.Split(list, ",")
+		netID, genHash, bootstrappers := params[0], params[1], params[2]
+		// validate bootstrappers
+		bs := strings.Split(bootstrappers, ",")
 		_, err := parseAddrInfos(bs)
 		if err != nil {
-			println("env CELESTIA_CUSTOM_BOOTSTRAPPERS: contains invalid multiaddress")
+			println("env CELESTIA_CUSTOM: contains invalid multiaddress")
 			panic(err)
 		}
-
 		bootstrapList[Network(netID)] = bs
+		// set custom network as default network for node to use
+		defaultNetwork = Network(netID)
+		// register network and genesis hash
+		networksList[defaultNetwork] = struct{}{}
+		genesisList[defaultNetwork] = strings.ToUpper(genHash)
 	}
 	// check if private network option set
 	if genesis, ok := os.LookupEnv("CELESTIA_PRIVATE_GENESIS"); ok {
--- a/params/default.go
+++ b/params/default.go
@@ -15,21 +15,8 @@ func DefaultNetwork() Network {
 }
 
 func init() {
-	// check if a different network from the registry was specified
-	if network, ok := os.LookupEnv("CELESTIA_NETWORK"); ok {
-		if _, exists := networksList[Network(network)]; !exists {
-			panic("unknown network specified")
-		}
-		defaultNetwork = Network(network)
-		return
-	}
-	// check if private network option set
-	if genesis, ok := os.LookupEnv("CELESTIA_PRIVATE_GENESIS"); ok {
-		defaultNetwork = Private
-		genesisList[Private] = strings.ToUpper(genesis)
-	}
 	// check if custom network option set
-	if customNet, ok := os.LookupEnv("CELESTIA_CUSTOM"); ok {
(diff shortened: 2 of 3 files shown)


Wondertan:
After sync discussion with @renaynay, we agreed on:
> CELESTIA_CUSTOM=chain-id=genesis_hash adds the new network to the registry
> CELESTIA_NETWORK=chain-id picks the network from the registry
> CELESTIA_BOOTSTAPPERS=chain-id>=adrr,addr.. adds or overrides bootstrappers for the particular network in a registry

params/default.go
```diff
@@ -14,8 +15,47 @@ func DefaultNetwork() Network {
 }
 
 func init() {
+	// check if a different network from the registry was specified
+	if network, ok := os.LookupEnv("CELESTIA_NETWORK"); ok {
+		if _, exists := networksList[Network(network)]; !exists {
```
Wondertan, line 20: There is a validate func for this,

params/default.go
```diff
@@ -14,8 +15,47 @@ func DefaultNetwork() Network {
...
+		if len(params) != 2 {
+			panic("must provide CELESTIA_CUSTOM in this format: <network_ID>=<genesis hash>")
+		}
+		netID, genHash := params[0], params[1]
+
+		defaultNetwork = Network(netID)
+		// register new network and set as default for node to use
+		networksList[defaultNetwork] = struct{}{}
+		genesisList[defaultNetwork] = strings.ToUpper(genHash)
+	}
```
Wondertan, line 48: CELESTIA_CUSTOM should before CELESTIA_NETWORK so that it adds to the registry and the forme reads. Otherwise it will fail

params/default.go
```diff
@@ -14,8 +15,47 @@ func DefaultNetwork() Network {
...
+			"store!\n\n")
+
+		params := strings.Split(customNet, "=")
+		// ensure both params are present
+		if len(params) != 2 {
+			panic("must provide CELESTIA_CUSTOM in this format: <network_ID>=<genesis hash>")
+		}
+		netID, genHash := params[0], params[1]
+
+		defaultNetwork = Network(netID)
```
Wondertan, line 44: You don't need to set the default here if you do the comment above

params/default.go
```diff
@@ -14,8 +15,47 @@ func DefaultNetwork() Network {
...
+	if bootstrappers, ok := os.LookupEnv("CELESTIA_BOOTSTRAPPERS"); ok {
+		params := strings.Split(bootstrappers, "=")
+		// ensure both params are present
+		if len(params) != 2 {
+			panic("must provide CELESTIA_BOOTSTRAPPERS in this format: " +
+				"<network_ID>=<boostrappers comma separated list>")
+		}
+
+		netID, list := params[0], params[1]
+		bootstrapList[Network(netID)] = strings.Split(list, ",")
```
Wondertan, line 59: It would be nice to do validation of address correctness. As I did in https://github.com/celestiaorg/celestia-node/pull/552

Wondertan (changes requested):

params/default.go
renaynay: resolved?

params/default.go
renaynay: resolved?

params/default.go
renaynay: resolveD?

params/default.go
```diff
@@ -14,6 +15,51 @@ func DefaultNetwork() Network {
 }
 
 func init() {
+	// check if custom network option set
+	if customNet, ok := os.LookupEnv("CELESTIA_CUSTOM_GENESIS"); ok {
+		fmt.Print("\n\nWARNING: Celestia custom network specified. Only use this option if the node is " +
+			"freshly created and initialized.\n**DO NOT** run a custom network over an already-existing node " +
+			"store!\n\n")
```
adlerjohn, line 22: Is this printed out after the node is already running?

params/default.go
```diff
@@ -14,6 +15,51 @@ func DefaultNetwork() Network {
...
+
+		params := strings.Split(customNet, "=")
+		// ensure both params are present
+		if len(params) != 2 {
+			panic("must provide CELESTIA_CUSTOM_GENESIS in this format: <network_ID>=<genesis hash>")
```
adlerjohn, line 27: Nit, but would `:` be a better separator?

params/default.go
```diff
@@ -14,6 +15,51 @@ func DefaultNetwork() Network {
...
+		// register new network and set as default for node to use
+		networksList[Network(netID)] = struct{}{}
+		genesisList[defaultNetwork] = strings.ToUpper(genHash)
+	}
+	// check if a custom network was specified
+	if network, ok := os.LookupEnv("CELESTIA_CUSTOM_NETWORK"); ok {
+		if err := Network(network).Validate(); err != nil {
+			panic("unknown network specified")
+		}
+		defaultNetwork = Network(network)
```
adlerjohn, line 40: Hmmm. This is a bit weird because you construct `Network(network)` twice. Maybe refactor?
```suggestion
		n = Network(network)
		if err := n.Validate(); err != nil {
			panic("unknown network specified")
		}
		defaultNetwork = n
```

params/default.go
```diff
@@ -14,6 +15,51 @@ func DefaultNetwork() Network {
...
+	}
+	// check if custom bootstrappers were provided for a network
+	if bootstrappers, ok := os.LookupEnv("CELESTIA_CUSTOM_BOOTSTRAPPERS"); ok {
+		params := strings.Split(bootstrappers, "=")
+		// ensure both params are present
+		if len(params) != 2 {
+			panic("must provide CELESTIA_CUSTOM_BOOTSTRAPPERS in this format: " +
+				"<network_ID>=<boostrappers comma separated list>")
```
adlerjohn, line 48: Why do these custom network parameters require network ID over and over?

params/default.go
```diff
@@ -14,6 +15,51 @@ func DefaultNetwork() Network {
...
+		}
+
+		netID, list := params[0], params[1]
+		// validate correctness
+		bs := strings.Split(list, ",")
+		_, err := parseAddrInfos(bs)
+		if err != nil {
+			println("env CELESTIA_CUSTOM_BOOTSTRAPPERS: contains invalid multiaddress")
```
adlerjohn, line 56: Print out the address in question

params/default.go
renaynay: Yes.

params/default.go
renaynay: We could use that too, I don't mind.

params/default.go
adlerjohn: Seems like a useless message then, since it could have bad effects on your already-existing setup. Is there some way to detect that then abort with this message instead?

params/default.go
adlerjohn: It's mostly that `CELESTIA_CUSTOM_GENESIS=networkname=hash` looks weird, while `CELESTIA_CUSTOM_GENESIS=networkname:hash` is better IMO.

params/default.go
adlerjohn: If there's no need for the network ID to actually be specified each time and can instead be specified once, then that seems like better UX.

renaynay: Or we could do `CELESTIA_CUSTOM`=`networkID:genHash:bootstrapperList`

params/default.go
renaynay: kk

renaynay: resolved https://github.com/celestiaorg/celestia-node/pull/717/commits/c2ac8d229a329b9d6ef592cadc7e4315f55dd059 ?

params/default.go
renaynay: check this: https://github.com/celestiaorg/celestia-node/pull/717/commits/c2ac8d229a329b9d6ef592cadc7e4315f55dd059

params/default.go
renaynay: w/ new changes (https://github.com/celestiaorg/celestia-node/pull/717/commits/c2ac8d229a329b9d6ef592cadc7e4315f55dd059), validation on network seems redundant. `Validate` gets called elsewhere on start-up anyway. All it does is check that network exists in map and we add it to map anyway.

params/default.go
renaynay: https://github.com/celestiaorg/celestia-node/pull/717/commits/c2ac8d229a329b9d6ef592cadc7e4315f55dd059 ?

params/default.go
renaynay: I can only really do this check on `Init` not on `Start` because `Start` will only ever run over a store that already exists (so we can't really fail out in `Start` if the store exists bc it *has* to exist in order for node to start). 

I will implement the check for `Init` but I think since this isn't a flag provided to users and instead is an environment variable, it should be fine.

node/init.go
```diff
@@ -115,6 +116,11 @@ func initRoot(path string) error {
 // initDir creates a dir if not exist
 func initDir(path string) error {
 	if utils.Exists(path) {
+		// if the dir already exists and `CELESTIA_CUSTOM` env var is set,
+		// fail out to prevent store corruption
+		if _, ok := os.LookupEnv("CELESTIA_CUSTOM"); ok {
```
adlerjohn, line 121: Refactor `"CELESTIA_CUSTOM"` into a constant.

renaynay: https://github.com/celestiaorg/celestia-node/pull/717/commits/d91c77f8e54f86520d023cea22e593d7320f4c56

params/default.go
Wondertan: We need them separated.

renaynay: I don't understand why the flags can't be used for that?

```
      --headers.trusted-hash string     Hex encoded header hash. Used to subjectively initialize header synchronization
      --headers.trusted-peers strings   Multiaddresses of a reliable peers to fetch headers from. (Format: multiformats.io/multiaddr)
```

Wondertan: Trusted hash and trusted peers are different concepts from genesis hash and bootstrap peers. It just happens so that right now we set those from genesis and bootstrap and this will be reworked a bit with future headerex changes

renaynay: If you declare a custom network: 

`CELESTIA_CUSTOM` (must provide net name, trusted hash, and trusted peers)

if we have network already in registry, that should mean we already have bootstrappers for that network (meaning we've hard-coded it)

Lets say we have several networks available at once: 

```
testnet
testnet-2
mainnet
```

we should expose the ability to choose network via a flag and then that would take the bootstrappers for those nets from the registry (unless trusted peers flag is passed)

Wondertan (approved):
To unblock @renaynay.

PR #721, opened 2022-05-17
fix(ipld): fix race in rare case when Retriver fetches a share from both Row and Col

### LINKED ISSUE #720: ipld: race between written shares during retrieval with Retriever
This is happening because shares can be obtained via Row and Col, and with some chance, during the reconstruction tests a race appears where LN sample the same share from different source and Retriever tries to fetch both

Wondertan:
Closes #720


Bidon15 (approved):

renaynay (approved):
LGTM

PR #722, opened 2022-05-17
fix(ipld): simply rely on global rsmt2d.Codec constructor for now to avoid races

### LINKED ISSUE #719: ipld: Retriever shouldn't reuse the same rsmt2d codec for multiple retrievals
Otherwise, we observe race conditions as the codec is not supported to be sync. Instead, the same codec constructor should be used.

Wondertan:
Closes #719


vgonkivs (approved):

renaynay (approved):
g

PR #725, opened 2022-05-17
introduce entity that will keep info about services current state and integrate it into DASer and Syncer

### LINKED ISSUE #724: service: Add flag for stopped services
We should add a flag that will prevent the service from being stopped twice.

#Why
Currently, all our Stop calls depends on fx.Lifecycle which guarantees it will be called once. Receiving of BEFP also cause service stopping. The second Stop call can either crash or freeze the node.

vgonkivs:
Resolves #724

### DIFF
--- a/das/daser.go
+++ b/das/daser.go
@@ -9,7 +9,7 @@ import (
 	logging "github.com/ipfs/go-log/v2"
 
 	"github.com/celestiaorg/celestia-node/header"
-	"github.com/celestiaorg/celestia-node/libs/utils"
+	"github.com/celestiaorg/celestia-node/libs/service"
 	"github.com/celestiaorg/celestia-node/service/share"
 )
 
@@ -17,6 +17,8 @@ var log = logging.Logger("das")
 
 // DASer continuously validates availability of data committed to headers.
 type DASer struct {
+	service.Service
+
 	da   share.Availability
 	hsub header.Subscriber
 
@@ -32,8 +34,6 @@ type DASer struct {
 
 	sampleDn  chan struct{} // done signal for sample loop
 	catchUpDn chan struct{} // done signal for catchUp loop
-
-	state *utils.StateWrapper
 }
 
 // NewDASer creates a new DASer.
@@ -52,7 +52,6 @@ func NewDASer(
 		jobsCh:    make(chan *catchUpJob, 16),
 		sampleDn:  make(chan struct{}),
 		catchUpDn: make(chan struct{}),
-		state:     utils.NewStateWrapper(),
 	}
 }
 
@@ -77,7 +76,7 @@ func (d *DASer) Start(context.Context) error {
 	dasCtx, cancel := context.WithCancel(context.Background())
 	d.cancel = cancel
 
-	d.state.SetState(utils.Running)
+	d.SetState(service.Running)
 	// kick off catch-up routine manager
 	go d.catchUpManager(dasCtx, checkpoint)
 	// kick off sampling routine for recently received headers
@@ -87,11 +86,11 @@ func (d *DASer) Start(context.Context) error {
 
 // Stop stops sampling.
 func (d *DASer) Stop(ctx context.Context) error {
-	if d.state.State() == utils.Stopped {
+	if d.State() == service.Stopped {
 		log.Debug("Daser is stopped")
 		return nil
 	}
-	defer d.state.SetState(utils.Stopped)
+	defer d.SetState(service.Stopped)
 	d.cancel()
 	// wait for both sampling routines to exit
 	for i := 0; i < 2; i++ {
--- a/das/daser.go
+++ b/das/daser.go
@@ -9,6 +9,7 @@ import (
 	logging "github.com/ipfs/go-log/v2"
 
 	"github.com/celestiaorg/celestia-node/header"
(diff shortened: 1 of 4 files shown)


libs/utils/state_wrapper.go
```diff
@@ -0,0 +1,34 @@
+package utils
+
+import "sync/atomic"
+
+type state uint32
+
+const (
+	Stopped state = iota
+	Runnnig
```
renaynay, line 9:
```suggestion
	Running
```

libs/utils/state_wrapper.go
```diff
@@ -0,0 +1,34 @@
+package utils
```
Wondertan, line 1: * `libs/utils` does not make sense(even though I made it, but I don't like this) 
* If this is needed for service management, we should make a `service` pkg.

libs/utils/state_wrapper.go
```diff
@@ -0,0 +1,34 @@
...
+
+type state uint32
+
+const (
+	Stopped state = iota
+	Runnnig
+)
+
+// StateWrapper provides an interface to change an check service state
+type StateWrapper interface {
```
Wondertan, line 13: StateWrapper is a wrong name for the interface, IMO. Also, do we really need this interface? Do you expect multiple implementations of it?  What about a simple Service struct that all our services embed through composition?

das/daser.go
```diff
@@ -82,6 +87,11 @@ func (d *DASer) Start(context.Context) error {
 
 // Stop stops sampling.
 func (d *DASer) Stop(ctx context.Context) error {
+	if d.state.GetState() == utils.Stopped {
+		log.Debug("Daser is stopped")
+		return nil
+	}
+	d.state.SetState(utils.Stopped)
```
Wondertan, line 94: Should be in the end of the method

libs/utils/state_wrapper.go
```diff
@@ -0,0 +1,34 @@
...
+	Stopped state = iota
+	Running
+)
+
+// StateWrapper provides an interface to change an check service state
+type StateWrapper interface {
+	// SetState allows to change service state
+	SetState(state)
+	// GetState returns service current state
+	GetState() state
```
Wondertan, line 17: You can omit Get and make it just State

Wondertan (commented):
Making Service like this `libs/service.Service` and usage like :

```go
type Syncer struct{
	service.Service
	...
}

func (s *Syncer) Stop(context.Context) error {
	if s.state.State() == service.Stopped {
		return nil
	}
}
```
WDYT?

Wondertan:
Other thoughts:
* As we discussed, this should be part of FX, and FX can track on its own the state of each Service that was started or not. No matter how we try to solve this problem, it will be better solved by the uber-go/fx and their team. We need deeper integration with the dependency graph to not just stop a Service but also stop everything that depends on the stopped service. We should make a proposal as I did in https://github.com/uber-go/fx/issues/825 for this.
* @vgonkivs, can you tell me if you have other plans for this structure? It feels like it is a bit of overengineering for a simple atomic switch. If there are other long term plans/ideas on how we can use this, then I am fine to keep it, but if not, I see no reason to keep it, considering the proper solution should be done on the FX side.

vgonkivs:
@Wondertan, no other plans except provided. I do not think that it's overengineering. It allows to re-use this structure in different services and provides simple aliases for On/Off state

libs/utils/state_wrapper.go
vgonkivs: removed

libs/utils/state_wrapper.go
vgonkivs: done

Wondertan:
Can you also rename the PR to something more appropriate? We already have a state service that accesses the State of the blockchain

das/daser.go
```diff
@@ -82,6 +86,11 @@ func (d *DASer) Start(context.Context) error {
 
 // Stop stops sampling.
 func (d *DASer) Stop(ctx context.Context) error {
+	if d.State() == service.Stopped {
+		log.Debug("Daser is stopped")
```
adlerjohn, line 90:
```suggestion
		log.Debug("DASer is stopped")
```

libs/service/service.go
```diff
@@ -0,0 +1,25 @@
...
+type State uint32
+
+const (
+	Stopped State = iota
+	Running
+)
+
+// Service provides an interface to change an check service state
+type Service struct {
+	state uint32
```
adlerjohn, line 14: ?
```suggestion
	state State
```

vgonkivs: In this case storing and loading state will be more complex:
```
* atomic.StoreUint32((*uint32)(&s.state), uint32(state))
* return State(atomic.LoadUint32((*uint32)(&s.state)))
```

IMO, better to leave `uint32`

adlerjohn: Then maybe the naming of `state` here is inconsistent? `s.state` seems to be a key for the state, not a state? If so, it it's 1) not clearly named or documented, and 2) should be `uint64`, unless there's a very specific reason it must be a `uint32`.

vgonkivs:
Closing this PR in favour of more priority tasks.

PR #729, opened 2022-05-19
node: Include expected err msg in`/namespaced_shares` test

renaynay:
This PR fixes a failing test for the `/namespaced_shares` endpoint by adding an expected err message check.


vgonkivs (approved):

Wondertan:
@vgonkivs, that is true, but for the first failure. Look at the second failure with race detector, it is still failing for namespace size related issue

Wondertan (approved):

PR #734, opened 2022-05-20
chore(ipld): extract byzantine error and constructor for it into a separate file

Wondertan:
* It does make sense to keep ErrByzantine in ipld/get.go
* Not to bloat Retriever

### DIFF
--- a/ipld/retriever_byzantine.go
+++ b/ipld/retriever_byzantine.go
@@ -1,0 +1,73 @@
+package ipld
+
+import (
+	"context"
+	"fmt"
+
+	format "github.com/ipfs/go-ipld-format"
+	"github.com/tendermint/tendermint/pkg/da"
+
+	"github.com/celestiaorg/celestia-node/ipld/plugin"
+	"github.com/celestiaorg/rsmt2d"
+)
+
+// ErrByzantine is a thrown when recovered data square is not correct
+// (merkle proofs do not match parity erasure-coding data).
+//
+// It is converted from rsmt2d.ByzantineRow/Col +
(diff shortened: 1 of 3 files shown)


ipld/retriever_byzantine.go
```diff
@@ -0,0 +1,73 @@
...
+		plugin.MustCidFromNamespacedSha256(root),
+		errShares,
+	)
+	if err != nil {
+		// Fatal as rsmt2d proved that error is byzantine,
+		// but we cannot properly collect the proof,
+		// so verification will fail and thus services won't be stopped
+		// while we still have to stop them.
+		// TODO(@Wondertan): Find a better way to handle
+		log.Fatalw("getting proof for ErrByzantine", "err", err)
```
vgonkivs, line 69: may be we can panic here?

Wondertan: It is basically the same

Wondertan: Where?

Wondertan: Panic is dev oriented way of exiting the app. Log fatal is user-oriented and does not spit with stack traces into the user

vgonkivs: Actually, we are using `log.Fatal` only once.

vgonkivs: But it's up to you

renaynay (approved):

renaynay:
@Wondertan maybe rebase and re-run ci

vgonkivs (approved):

PR #736, opened 2022-05-20
chore: bump versions of app/sdk/core dependecies to latest

Bidon15:
- app  0.5.0-rc-2 
- sdk 0.46 
- core 0.35.4

needed for successful testnet pairing with app


go.mod
```diff
@@ -6,17 +6,17 @@ replace github.com/ipfs/go-verifcid => github.com/celestiaorg/go-verifcid v0.0.1
...
-	github.com/cosmos/cosmos-sdk v0.44.3
+	github.com/celestiaorg/rsmt2d v0.5.0
+	github.com/cosmos/cosmos-sdk v0.46.0-beta2.0.20220418184507-c53157dd63f6
 	github.com/dgraph-io/badger/v2 v2.2007.4
 	github.com/gammazero/workerpool v1.1.2
 	github.com/gogo/protobuf v1.3.3
 	github.com/gorilla/mux v1.8.0
 	github.com/hashicorp/go-retryablehttp v0.7.1-0.20211018174820-ff6d014e72d9
-	github.com/hashicorp/golang-lru v0.5.4
+	github.com/hashicorp/golang-lru v0.5.5-0.20210104140557-80c98217689d
```
adlerjohn, line 19: If this isn't an indirect dependency, it shouldn't use a non-tag.

Bidon15: True. Still, this is automatically done after we execute 
`go mod tidy -compat=1.17`
otherwise, using `go mod tidy` will cause such messages:
```
To upgrade to the versions selected by go 1.16, leaving some packages unresolved:
	go mod tidy -e -go=1.16 && go mod tidy -e -go=1.17
If reproducibility with go 1.16 is not needed:
	go mod tidy -compat=1.17
```

I still don't understand how go is deciding on versioning names as the diff between 0.5.4 and this(0.5.5-xxxx) is just 2 commits without any go.mod changes
1. eb529947af531eb529020ba979a7a887338904d1
2. 80c98217689d6df152309d574ccc682b21dc802c

adlerjohn: Where is `golang-lru` actually used/imported?

node/state/keyring.go
```diff
@@ -8,25 +8,30 @@ import (
...
 	"github.com/celestiaorg/celestia-node/node/key"
 	"github.com/celestiaorg/celestia-node/params"
 )
 
+// func(fx.Lifecycle, keystore.Keystore, params.Network) (state.Accessor, error) {
+// 	return func(lc fx.Lifecycle, ks keystore.Keystore, net params.Network) (state.Accessor, error) {
+// TODO @renaynay: Include option for setting custom `userInput` parameter with
+//  implementation of https://github.com/celestiaorg/celestia-node/issues/415.
+// TODO @renaynay @Wondertan: ensure that keyring backend from config is passed
+//  here instead of hardcoded `BackendTest`: https://github.com/celestiaorg/celestia-node/issues/603.
```
Bidon15, line 23: Needs to be removed 😓

Bidon15 (commented):
some housekeeping will be pushed after the CI run

go.mod
```diff
@@ -6,17 +6,17 @@ replace github.com/ipfs/go-verifcid => github.com/celestiaorg/go-verifcid v0.0.1
 
 require (
 	github.com/BurntSushi/toml v1.1.0
-	github.com/celestiaorg/celestia-app v0.4.0
+	github.com/celestiaorg/celestia-app v0.5.0-rc-2.0.20220522173832-fcc1bd0cbc22
```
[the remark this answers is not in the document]
Bidon15: Thanks for the hint - changed

Wondertan:
Everything works now.

cmd/celestia/main.go
```diff
@@ -5,16 +5,19 @@ import (
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
 func init() {
 	// This is necessary to ensure that the account addresses are correctly prefixed
 	// as in the celestia application.
-	cosmoscmd.SetPrefixes(app.AccountAddressPrefix)
+	cfg := sdk.GetConfig()
+	cfg.SetBech32PrefixForAccount(app.Bech32PrefixAccAddr, app.Bech32PrefixAccPub)
+	cfg.Seal()
```
liamsi, line 20: Isn't the app config already sealed at this point? cc @evan-forbes

Bidon15:
The only unit test failing in race is `TestRemoteClient_StartBlockSubscription_And_GetBlock`
This will be fixed on a separate PR to not overflow this one

go.mod
Bidon15: > Where is `golang-lru` actually used/imported?

Somewhere in core 🫤

cmd/celestia/main.go
liamsi: @Bidon15 did you try this against a recent core/app combo?

header/header.go
```diff
@@ -47,7 +47,11 @@ func MakeExtendedHeader(
 ) (*ExtendedHeader, error) {
 	var dah DataAvailabilityHeader
 	if len(b.Txs) > 0 {
-		namespacedShares, _ := b.Data.ComputeShares()
+		// TODO(@Bidon15): Do we need curLen(int)?
+		namespacedShares, _, err := b.Data.ComputeShares(uint64(0))
```
liamsi, line 51: >  If a square size
> of 0 is passed, then it is determined based on how many shares are needed to
> fill the square for the underlying block data.

Hmm, can't this simply re-use the square size as well?

go.mod
Wondertan: No, we use it in header.Store

header/header.go
evan-forbes: we actually need to use the OriginalSquare size here, otherwise the data root could be calculated differently.

in the future, we won't pass the square to the method, but for now we are

go.mod
adlerjohn: Still unclear to me why `go mod tidy -compat=1.17` bumps this to an exact commit instead of just using the latest tag.

cmd/celestia/main.go
evan-forbes: we only seal them when we are init our cli. unless we are importing and calling that cli, then I don't think the config is sealed

go.mod
Wondertan: This is not true anymore. Try to checkout locally. This is something new to me and I didn't have enough time to dive and understand what it the reasoning behind this thing
> It is not golang that decides here but the developer that imports this.

Wondertan: Really, if you don't add the commit at the end. The go does not compile and throws an error above

Wondertan:
The PR that fixes flakeness is prepared to be made after this one is merged

cmd/celestia/main.go
liamsi: But what is the purpose of sealing then? I'd assume you can't modify the config afterwards.

go.mod
liamsi: That is probably due to some indirect dep conflict then. Internally sth uses that newer version and there is no way to resolve this other than using the same version. Resolving this thread for now!

liamsi (approved):
Awesome 👍🏼

node/tests/swamp/config.go
```diff
@@ -18,9 +19,13 @@ type Components struct {
 // DefaultComponents creates a KvStore with a block retention of 200
 // In addition, the empty block interval is set to 200ms
 func DefaultComponents() *Components {
-	app := core.CreateKVStore(200)
-	tnCfg := tn.TestConfig()
-	tnCfg.Consensus.CreateEmptyBlocksInterval = 200 * time.Millisecond
+	app := core.CreateKVStore(2000)
+	tnCfg, err := rpctest.CreateConfig("swamp_tm")
+	if err != nil {
+		// todo remove panic
```
Wondertan, line 25:
```suggestion
```

Wondertan (approved):

PR #742, opened 2022-05-23
node: Use in-mem keyring for tests and clean up test redundancy

renaynay:
This PR:
* creates a `WithCustomKeyring` option to override the default keyring constructed by the node and uses that option to override the keyring with an in-memory keyring for testing purposes inside all unit tests using a `TestNode`; and
* cleans up a lot of redundancy in tests inside the `node` pkg


vgonkivs (approved):

vgonkivs:
LGTM 🚀

node/settings.go
```diff
@@ -78,3 +79,11 @@ func WithHeaderConstructFn(construct header.ConstructFn) Option {
 		sets.opts = append(sets.opts, fx.Replace(construct))
 	}
 }
+
+// WithCustomKeyringSigner overrides the default keyring signer constructed
+// by the node.
+func WithCustomKeyringSigner(signer *apptypes.KeyringSigner) Option {
```
Wondertan, line 85: nit: With...Option already assumes customizability, so `Custom` is redundant, or we should add Custom to each existing options for consistency

node/testing.go
```diff
@@ -21,9 +23,23 @@ func MockStore(t *testing.T, cfg *Config) Store {
...
+		WithNetwork(params.Private),
+		WithCustomKeyringSigner(InMemKeyringSigner(t)),
+	)
 	store := MockStore(t, DefaultConfig(tp))
 	nd, err := New(tp, store, opts...)
 	require.NoError(t, err)
 	return nd
 }
+
+func InMemKeyringSigner(t *testing.T) *apptypes.KeyringSigner {
```
Wondertan, line 38: Let's call it `TestKeyringSigner`

Wondertan (approved):

node/settings.go
renaynay: kk will remove Custom, np

node/testing.go
renaynay: yup

PR #745, opened 2022-05-23
feat(service/rpc): add /namespaced_data endpoint

### LINKED ISSUE #740: [Feature Request]: `/namespaced_data` RPC endpoint
### Implementation ideas

Reasons why `/namespaced_data` RPC endpoint is important:
* user submits data using `/submit_pfd` - each data blob ends up as a separate message in Celestia block
* data returned by `/namespaced_shares` consists of shares, which is internal format of data used by Celestia for DA; users need to rebuild messages (originally submitted blobs) from the shares
* there is only Go implementation for rebuilding messages
  * other languages are not supported; implementation is not trivial and error prone
  * optimint can't use celestia-app as dependency - code needs to be copied (therefore any changes/updates/bugfixes are problematic)

My proposition is to create new RPC endpoint: `/namespaced_data/{nid}/height/{height}`. It will return reconstructed "data" - the same messages/blobs that was submitted using `/submit_pfd`.

tzdybal:
Resolves #740.

### GET `/namespaced_data/{nID}`
Returns data (messages) of the given namespace ID `nID` from the latest block (chain head).

Request

```
curl -X GET http://<ip>:26658/namespaced_data/0000000000000001
```

Response

```
{"data":["gAIDCpcCCp4BCjcvY29zbW9zLmRpc3RyaWJ1dGlvbi52MWJldGExLk1zZ1dpdGhkcmF3RGVsZWdhdG9yUmV3YXJkEmMKLGNlbGVzMTJwZ3Z5ZGw4dDRzejg2N3NhY3A5ZGc4cmE2eDJ2ZTBrZ3k2Nm4yEjNjZWxlc3ZhbG9wZXIxMnBndnlkbDh0NHN6ODY3c2FjcDlkZzhyYTZ4MnZlMGtkZjRjeXoKdAo7L2Nvc21vcy5kaXN0cmlidXRpb24udjFiZXRhMS5Nc2dXaXRoZHJhd1ZhbGlkYXRvckNvbW1pc3Npb24SNQozY2VsZXN2YWxvcGVygAKAAmd2eWRsOA=="],"height":1056372}
```

### GET `/namespaced_data/{nID}/height/{height}`
Returns data (messages) of the given namespace ID `nID` from the block of the given height `height`.

Request

```
curl -X GET http://<ip>:26658/namespaced_data/0000000000000001/height/357889
```

Response

```
{"data":["gAICCpEBCowBChwvY29zbW9zLmJhbmsudjFiZXRhMS5Nc2dTZW5kEmwKLGNlbGVzMXk0cXAwZm41OTk5bm4zejV6NjJ6OG1meHo3cHN4anp2dGE2bDhzEixjZWxlczF0MzZzcXN0cWYyM2dnMHBmdmNyOXJoOW1rN2RscnhmcHY3czJlNRoOCgVjZWxlcxIFMTAwMDASABJmClAKRgofL2Nvc21vcy5jcnlwdG8uc2VjcDI1NmsxLlB1YktleRIjCiECxHg4PwK2kHl3YHITfqZfoaHnYt9yVKqnwm+PwllHBwMSBAoCCH8YABISCgwKBWNlbGVzgAKAAjAwEIDxBA==","gAIuY3J5cHRvLnNlY3AyNTZrMS5QdWJLZXkSIwohAloFm7w2T2insKy/0MWY0kFX2bS5P1G9LYsg85HGWOPTEgQKAggBGOAWEgQQwJoMGkBxcP4sAopLkpfTzK4+K+VYR6eRUg9L+vQqCyjShLEn/kTh/q2DxEsoU4wUfQck4ItYWa0j6T4zYNV4Li4CDyBnugIKmgEKlwEKIy9jb3Ntb3Muc3Rha2luZy52MWJldGExLk1zZ0RlbGVnYXRlEnAKLGNlbGVzMTl6OGpndGxwNXl6cnlsOXg5eGZ2cjlsemtlZjBkNWFubHBrc25rEjNjZWxlc3ZhgAKAAmVyMTl6OA=="],"height":357889}
```

### DIFF
--- a/node/rpc_test.go
+++ b/node/rpc_test.go
@@ -4,12 +4,11 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"net/http"
 	"strconv"
-	"strings"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/fx"
 
@@ -25,11 +24,34 @@ import (
 // celestia-node. They will be removed upon refactoring of the RPC
 // architecture and Public API. @renaynay @Wondertan.
 
+const testHeight = uint64(2)
+
 // TestNamespacedSharesRequest tests the `/namespaced_shares` endpoint.
 func TestNamespacedSharesRequest(t *testing.T) {
+	testGetNamespacedRequest(t, "namespaced_shares", func(t *testing.T, resp *http.Response) {
+		t.Helper()
+		namespacedShares := new(rpc.NamespacedSharesResponse)
+		err := json.NewDecoder(resp.Body).Decode(namespacedShares)
+		assert.NoError(t, err)
+		assert.Equal(t, testHeight, namespacedShares.Height)
+	})
+}
+
+// TestNamespacedDataRequest tests the `/namespaced_shares` endpoint.
+func TestNamespacedDataRequest(t *testing.T) {
+	testGetNamespacedRequest(t, "namespaced_data", func(t *testing.T, resp *http.Response) {
+		t.Helper()
+		namespacedData := new(rpc.NamespacedDataResponse)
+		err := json.NewDecoder(resp.Body).Decode(namespacedData)
+		assert.NoError(t, err)
+		assert.Equal(t, testHeight, namespacedData.Height)
+	})
+}
+
+func testGetNamespacedRequest(t *testing.T, endpointName string, assertResponseOK func(*testing.T, *http.Response)) {
+	t.Helper()
 	nd := setupNodeWithModifiedRPC(t)
 	// create several requests for header at height 2
-	height := uint64(2)
 	var tests = []struct {
 		nID         string
 		expectedErr bool
@@ -53,8 +75,8 @@ func TestNamespacedSharesRequest(t *testing.T) {
 
 	for i, tt := range tests {
 		t.Run(strconv.Itoa(i), func(t *testing.T) {
-			endpoint := fmt.Sprintf("http://127.0.0.1:%s/namespaced_shares/%s/height/%d",
-				nd.RPCServer.ListenAddr()[5:], tt.nID, height)
+			endpoint := fmt.Sprintf("http://127.0.0.1:%s/%s/%s/height/%d",
+				nd.RPCServer.ListenAddr()[5:], endpointName, tt.nID, testHeight)
 			resp, err := http.Get(endpoint)
 			defer func() {
 				err = resp.Body.Close()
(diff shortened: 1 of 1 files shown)


service/rpc/share.go
```diff
@@ -76,3 +62,104 @@ func (h *Handler) handleSharesByNamespaceRequest(w http.ResponseWriter, r *http.
...
+	length := uint64(len(share))
+	n := binary.PutUvarint(lenBuf, length)
+
+	var withDelim []byte
+	if len(share)+n <= cap(share) {
+		withDelim = share[:cap(share)]
+	} else {
+		withDelim = make([]byte, len(share)+n)
+		copy(withDelim[:ipld.NamespaceSize], share[:ipld.NamespaceSize])
+	}
```
tzdybal, line 142: LMAO, it's `realloc`, but in Go :D

While debugging, I noticed, that usually `cap(share)` is significantly bigger than  `len(share)` (`share` is a result of multiple `append`s, so it's reallocated few times, and has some space on the end).

tzdybal:
CC: @YazzyYaz

renaynay (commented):
Looks good - can you add some unit tests in `node/rpc_test.go` for this by chance? I can add them as well and push to the branch if you want @tzdybal

tzdybal:
Yes, sure, I'm adding them now.

service/rpc/share.go
```diff
@@ -76,3 +62,104 @@ func (h *Handler) handleSharesByNamespaceRequest(w http.ResponseWriter, r *http.
...
+		return nil, 0, err
+	}
+	// perform request
+	shares, err := h.share.GetSharesByNamespace(ctx, header.DAH, nID)
+	return shares, header.Height, err
+}
+
+func dataFromShares(shares []share.Share) ([][]byte, error) {
+	for i := range shares {
+		insertDelimiter(shares[i])
```
tzdybal, line 118: @evan-forbes I was testing this mostly on blocks from `devnet-2`, generated with older node/core/app.
Do I understand correctly, that this step will no longer be needed with up-to-date PFD processing?

evan-forbes: I don't think you should have had to add that at all, even for old devnet!

it should just work like this https://github.com/celestiaorg/dalc/blob/6589e5dc61037646c663f86bd97e5036c56ec77e/server/server.go#L130-L138

https://github.com/celestiaorg/dalc/blob/6589e5dc61037646c663f86bd97e5036c56ec77e/server/server_test.go#L14-L55

tzdybal: It was always hidden behind other issues (like random share ordering).

YazzyYaz:
Is the plan to merge this today or tomorrow? If so, I can also update the docs by making a PR now

node/rpc_test.go
```diff
@@ -25,11 +24,34 @@ import (
 // celestia-node. They will be removed upon refactoring of the RPC
 // architecture and Public API. @renaynay @Wondertan.
 
+const testHeight = uint64(2)
```
liamsi, line 27: I would rather make this part of the test-case struct:
```go
	[]struct {
		nID         string
		height      uint64
		expectedErr bool
		errMsg      string
	}
```
 But can happen in a follow-up PR. We should not only test with height 2.

liamsi (approved):
Looks good!

node/rpc_test.go
tzdybal: We can't query for actual data, as we're not able to `/submit_pfd` - results are empty, so it doesn't matter what block we check.
This is more like a "smoke test" it only checks that nothing terrible happens when RPC is called.

Wondertan (approved):
Nice!

PR #746, opened 2022-05-23
cmd/bridge: Remove `--headers.trusted-hash` flag for bridge node

renaynay:
This PR removes the `headers.trusted-hash` flag from bridge node options.

### DIFF
--- a/cmd/celestia/bridge.go
+++ b/cmd/celestia/bridge.go
@@ -1,4 +1,3 @@
-//nolint:dupl
 package main
 
 import (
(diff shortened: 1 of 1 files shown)


cmd/celestia/bridge.go
```diff
@@ -1,4 +1,3 @@
-//nolint:dupl
```
adlerjohn, line 1: Are you sure this should be removed?

Wondertan: Yeah. Linter was bugging us with duplicate code while it was not really so we have to put nolint there. And this change finally allows us to remove nolint

Wondertan:
However, this will close a manual pruning hack for bridge node runners. It is arguable we if we should allow it or not for Bridges and Fulls cc @renaynay @adlerjohn

adlerjohn (approved):

Issue #748, opened 2022-05-24
node/services: Check if an empty block has already been stored


renaynay:
Inside `StoreEmptyBlock` inside the `node/services` pkg, implement a check to see if an empty block has already been stored in the DAG. If not, store empty block. If yes, return.

Wondertan:
@renaynay, IIRC this was implemented, pls confirm

renaynay:
Confirm

PR #749, opened 2022-05-24
node/services: Invoke function to ensure that an empty block is stored at least once

renaynay:
This fixes an issue with an optimisation inside of `header.MakeExtendedHeader` where if a block has empty data, we do not store it to prevent unnecessary redundancy. The problem is, we forgot to ensure that an empty block was stored **at least once**. We found this bug while trying to run a light node against a bridge node on `dryrun-2` and found that the `DAS` functionality in light node was hanging -- this was due to the bridge node not being able to serve blocks to the light node as a result of never having stored that empty block.

This PR ensures that. 

There is an additional problem with this PR however, which is that an empty block will be stored upon construction *every time*. To fix that, we need to implement a solution that checks to see if an empty block has already been stored in the DAG, and if so, skip storing. Issue: https://github.com/celestiaorg/celestia-node/issues/748. 

Thanks to @Wondertan for remembering this optimization :)

### DIFF
--- a/node/components.go
+++ b/node/components.go
@@ -17,6 +17,7 @@ import (
 	"github.com/celestiaorg/celestia-node/node/services"
 	statecomponents "github.com/celestiaorg/celestia-node/node/state"
 	"github.com/celestiaorg/celestia-node/params"
+	"github.com/celestiaorg/celestia-node/service/share"
 )
 
 // lightComponents keeps all the components as DI options required to build a Light Node.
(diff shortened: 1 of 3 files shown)


Wondertan (commented):
One comment regarding code placement. Also, we should make an issue about making the check work.

Wondertan (approved):
LFG Mamaki

liamsi (approved):

PR #757, opened 2022-05-24
cmd/cel-key: Extract key utility into separate binary to prevent tight coupling between sdk and celestia-node binary

renaynay:
This PR extracts the cosmos sdk key utility from the main celestia-node binary and into a separate `cel-key` binary. 

Use the following flag `--node.type <bridge || full || light>` in order to set the keyring-dir to bridge, light, or full store directory. 

This prevents us having to use cosmos sdk client context initialisation (which is quite messy) inside celestia-node binary.

Thank you @evan-forbes and @Bidon15 for figuring out how to make the key utility actually work after it was broken during the upgrade :) 

### Makefile

`make cel-key` --> builds cel-key binary

`make install-key` --> installs cel-key in GOBIN

### DIFF
--- a/cmd/cel-key/node_types.go
+++ b/cmd/cel-key/node_types.go
@@ -1,0 +1,52 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+	flag "github.com/spf13/pflag"
+
+	sdkflags "github.com/cosmos/cosmos-sdk/client/flags"
+)
+
+var (
+	nodeDirKey = "node.type"
+
+	bridgeDir = "~/.celestia-bridge/keys"
+	fullDir   = "~/.celestia-full/keys"
+	lightDir  = "~/.celestia-light/keys"
+)
+
+func DirectoryFlags() *flag.FlagSet {
+	flags := &flag.FlagSet{}
+	flags.String(nodeDirKey, "", "Sets key utility to use the node type's directory (e.g. "+
+		"~/.celestia-light if --node.type light is passed.")
+
+	return flags
+}
+
+func ParseDirectoryFlags(cmd *cobra.Command) error {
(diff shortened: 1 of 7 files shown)


liamsi (commented):
This needs @YazzyYaz' attention for docs I assume

cmd/celestia/main.go
```diff
@@ -36,12 +29,17 @@ func main() {
 }
 
 func run() error {
+	cfg := sdk.GetConfig()
+	cfg.SetBech32PrefixForAccount(app.Bech32PrefixAccAddr, app.Bech32PrefixAccPub)
+	cfg.Seal()
+
```
Wondertan, line 35: Why not init? 
Also, It seems weird that this is required for `celestia` binary to function. Can we also extract it into the separate bin or it is required for stateaccess to function?

cmd/cel-key/node_types.go
```diff
@@ -0,0 +1,52 @@
...
+
+func DirectoryFlags() *flag.FlagSet {
+	flags := &flag.FlagSet{}
+	flags.String(nodeDirKey, "", "Sets key utility to use the node type's directory (e.g. "+
+		"~/.celestia-light if --node.type light is passed.")
+
+	return flags
+}
+
+func ParseDirectoryFlags(cmd *cobra.Command) error {
```
Wondertan, line 28: Any chance we can make dir customizable as we extracted this into a separate bin?

Wondertan (approved):
LGTM and a few questions

cmd/cel-key/node_types.go
renaynay: User can do `--keyring-dir`

cmd/celestia/main.go
renaynay: good catch

Bidon15 (approved):

PR #765, opened 2022-05-25
Celestia Node TR

ruesandora:
Celestia node kurulumu Türkçe


Bidon15 (commented):
Thanks for the effort! Left a small nit 💪

PR #768, opened 2022-05-28
cmd: Trim `--core.grpc` flag value

### LINKED ISSUE #766: Passing IP Address with a https://, tcp:// or http:// prefix results in "Too Many Colon" Error
### Celestia Node version

latest

### OS

Ubuntu

### Install tools

_No response_

### Others

_No response_

### Steps to reproduce it

```sh
celestia bridge start --core.grpc http://localhost:9090
```

will result in an error of too many colons but:

```sh
celestia bridge start --core.grpc localhost:9090
```

won't.

It would be good to have both options supported from a UX point of view.

### Expected result

```sh
celestia bridge start --core.grpc http://localhost:9090
```
Should work

### Actual result

Only

```sh
celestia bridge start --core.grpc localhost:9090
```
works

### Relevant log output

_No response_

### Notes

_No response_

renaynay:
Resolves #766.

### DIFF
--- a/cmd/flags_core.go
+++ b/cmd/flags_core.go
@@ -52,7 +52,9 @@ func ParseCoreFlags(cmd *cobra.Command, env *Env) error {
 	grpcEndpoint := cmd.Flag(coreGRPCFlag).Value.String()
 	if grpcEndpoint != "" {
 		// trim ip:port from the endpoint
-		grpcEndpoint = strings.Trim(grpcEndpoint, "http://") //nolint:staticcheck
+		grpcEndpoint = strings.Trim(grpcEndpoint, "http://")  //nolint:staticcheck
+		grpcEndpoint = strings.Trim(grpcEndpoint, "https://") //nolint:staticcheck
+		grpcEndpoint = strings.Trim(grpcEndpoint, "tcp://")   //nolint:staticcheck
 		env.AddOptions(node.WithGRPCEndpoint(grpcEndpoint))
 	}
 
--- a/cmd/flags_core.go
+++ b/cmd/flags_core.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net"
 	"net/url"
+	"strings"
 
 	"github.com/spf13/cobra"
 	flag "github.com/spf13/pflag"
@@ -50,6 +51,8 @@ func ParseCoreFlags(cmd *cobra.Command, env *Env) error {
 
 	grpcEndpoint := cmd.Flag(coreGRPCFlag).Value.String()
 	if grpcEndpoint != "" {
+		// trim ip:port from the endpoint
+		grpcEndpoint = strings.Trim(grpcEndpoint, "http://") //nolint:staticcheck
 		env.AddOptions(node.WithGRPCEndpoint(grpcEndpoint))
 	}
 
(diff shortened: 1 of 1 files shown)


cmd/flags_core.go
```diff
@@ -50,6 +51,10 @@ func ParseCoreFlags(cmd *cobra.Command, env *Env) error {
 
 	grpcEndpoint := cmd.Flag(coreGRPCFlag).Value.String()
 	if grpcEndpoint != "" {
+		// trim ip:port from the endpoint
+		grpcEndpoint = strings.Trim(grpcEndpoint, "http://")  //nolint:staticcheck
+		grpcEndpoint = strings.Trim(grpcEndpoint, "https://") //nolint:staticcheck
+		grpcEndpoint = strings.Trim(grpcEndpoint, "tcp://")   //nolint:staticcheck
```
adlerjohn, line 57: This seems rather...hacky. Isn't there some URL library that can be used to trim these properly?

Wondertan: There is. https://pkg.go.dev/net/url. @renaynay

renaynay:
I've just included your suggested change @Wondertan . This however means that the endpoint MUST be passed with a preceding `http://` `https://` or `tcp://` in order to be parsed properly by the `net/url` pkg. 

This means all docs with `--core.grpc` flag must be updated from `--core.grpc <ip>:<port>` to `--core.grpc http://<ip>:<port>` @YazzyYaz

cmd/flags_core.go
```diff
@@ -51,11 +50,12 @@ func ParseCoreFlags(cmd *cobra.Command, env *Env) error {
...
-		grpcEndpoint = strings.Trim(grpcEndpoint, "http://")  //nolint:staticcheck
-		grpcEndpoint = strings.Trim(grpcEndpoint, "https://") //nolint:staticcheck
-		grpcEndpoint = strings.Trim(grpcEndpoint, "tcp://")   //nolint:staticcheck
-		env.AddOptions(node.WithGRPCEndpoint(grpcEndpoint))
+		// parse ip:port from the endpoint
+		grpcURL, err := url.Parse(grpcEndpoint)
+		if err != nil {
+			return err
+		}
+		env.AddOptions(node.WithGRPCEndpoint(grpcURL.Host))
```
liamsi, line 58: maybe dumb question but what's the point of passing in the protocol (tcp/http(s)) if it is not used anyways?

YazzyYaz:
Hi @renaynay does this apply also to `--core.remote`?

renaynay:
> Hi @renaynay does this apply also to `--core.remote`?

@YazzyYaz core remote takes the protocol already so yes

renaynay:
Let me push though to ensure parsing there too

cmd/flags_core.go
liamsi: I guess I do not really understand why we are telling users that they must add the protocol only to remove it internally.
I'm wondering if users were only doing it in the first place bc of the docs?

YazzyYaz: Users are already adding http:// and it's resulting in a lot of them not knowing why they can't connect to the RPC.

YazzyYaz: I can make the docs say "don't add http://" but feel users will run into this issue anyways and forget to remove itand from a UX, it shouldn't be a blocker for developers looking to connect to an RPC imo.

liamsi: OK, resolving.

liamsi (approved):

adlerjohn (approved):

PR #769, opened 2022-05-28
Add install target in Makefile, move old behavior to go-install

### LINKED ISSUE #332: `make install` should not use `go install`
Even though it is simple to use `go install`, it requires the person building the Node from scratch to set the GOBIN for this to work reliably, thus to build the Node locally, one has to be aware of Go's semantics. 

I propose to avoid this by making `make install` not to rely on `go install`, but to `cp` the results from `make build` to `/usr/local/bin` which may require the user's `sudo` permissions. Furthermore, we should use built-in GNU `install` instead of `cp`.

iofq:
Resolves #332 

Although I agree with [marbar3778](https://github.com/marbar3778)'s assertion that this is an issue maybe better fixed by documentation and we could continue to use the golang native tooling, here is a PR with the Makefile target implemented as proposed.

 I moved the existing `go install ...` behavior to `make go-install`. I also elected to always build before install and prompt for sudo within the `install` target, to keep the UX as `make install` rather than `make build; sudo make install`.

`$PREFIX` can be overridden like `make PREFIX=/my/dir` if needed, but I'm not sure this is worth mentioning in the README

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,9 @@
 SHELL=/usr/bin/env bash
 PROJECTNAME=$(shell basename "$(PWD)")
 LDFLAGS="-X 'main.buildTime=$(shell date)' -X 'main.lastCommit=$(shell git rev-parse HEAD)' -X 'main.semanticVersion=$(shell git describe --tags --dirty=-dev)'"
+ifeq (${PREFIX},)
+	PREFIX := /usr/local
+endif
 
 ## help: Get more info on make commands.
 help: Makefile
@@ -19,11 +22,17 @@ clean:
 	@echo "--> Cleaning up ./build"
 	@rm -rf build/*
 
-## install: Build and install the celestia-node binary into the GOBIN directory.
-install:
+## install: Build and install the celestia-node binary into the $PREFIX (/usr/local/ by default) directory.
+install: build
+	@echo "--> Installing Celestia"
+	@install -v ./build/* -t ${PREFIX}/bin/
+.PHONY: install
+
+## go-install: Build and install the celestia-node binary into the GOBIN directory.
+go-install:
 	@echo "--> Installing Celestia"
 	@go install -ldflags ${LDFLAGS}  ./cmd/celestia
-.PHONY: install
+.PHONY: go-install
 
 ## shed: Build cel-shed binary.
 cel-shed:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -19,12 +22,18 @@ clean:
 	@echo "--> Cleaning up ./build"
 	@rm -rf build/*
 
-## install: Build and install the celestia-node binary into the GOBIN directory.
-install:
+## install: Build and install the celestia-node binary into the $PREFIX (/usr/local/ by default) directory.
+install: build
 	@echo "--> Installing Celestia"
-	@go install -ldflags ${LDFLAGS}  ./cmd/celestia
+	@sudo install -v ./build/* -t ${PREFIX}/bin/
```
adlerjohn, line 28: This makes the calling shell save the sudo credentials. I.e. the next time I do anything that requires `sudo` it will pass silently. Not sure this is better behavior than explicitly require `make install` to be called with `sudo`.

adlerjohn (approved):
looks fine to me, but will defer to one of the other reviewers for final approval

Makefile
```diff
@@ -1,6 +1,9 @@
 SHELL=/usr/bin/env bash
 PROJECTNAME=$(shell basename "$(PWD)")
 LDFLAGS="-X 'main.buildTime=$(shell date)' -X 'main.lastCommit=$(shell git rev-parse HEAD)' -X 'main.semanticVersion=$(shell git describe --tags --dirty=-dev)'"
+ifeq (${PREFIX},)
+	PREFIX := /usr/local
```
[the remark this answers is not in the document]
iofq: I think `PREFIX` is the Makefile convention [(source)](https://www.gnu.org/prep/standards/html_node/Directory-Variables.html#Directory-Variables). `EXEC_PREFIX` maybe, or I guess `PATH_PREFIX` is fine.

README.md
```diff
@@ -33,7 +33,7 @@ See the official docs page for system requirements per node type:
 ```sh
 git clone https://github.com/celestiaorg/celestia-node.git 
 cd celestia-node
-make install
+make build; sudo make install
```
adlerjohn, line 36:
```suggestion
make build
sudo make install
```

iofq: updated.

adlerjohn (approved):

PR #779, opened 2022-05-30
params: Fix issue in `CELESTIA_CUSTOM` to be able to parse genesis even if bootstrappers not provided

### LINKED ISSUE #776: Genesis Hash in CELESTIA_CUSTOM is parsed only when bootstrappers are not provided
It's not possible to use custom genesis hash and bootstrappers at the same time.

The condition checks for exact value: https://github.com/celestiaorg/celestia-node/blob/f22aae25867c0734ad29da61218d63f477f99683/params/default.go#L38

This should be changed to `>= 2`.

renaynay:
Resolves #776

### DIFF
--- a/params/default.go
+++ b/params/default.go
@@ -35,7 +35,7 @@ func init() {
 		defaultNetwork = Network(netID)
 		networksList[defaultNetwork] = struct{}{}
 		// check if genesis hash provided and register it if exists
-		if len(params) == 2 {
+		if len(params) >= 2 {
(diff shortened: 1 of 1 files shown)


Wondertan (approved):
😄

Wondertan:
wtf?
image

tzdybal (approved):
:)

adlerjohn (approved):
utACK

PR #783, opened 2022-05-31
feat(cmd): add new flag to allow granular control over level of each logging module

Wondertan:
No issue here, just as simple flag and lowering verbosity of some annoying logs in our deps

### DIFF
--- a/cmd/flags_misc.go
+++ b/cmd/flags_misc.go
@@ -64,6 +64,10 @@ func ParseMiscFlags(cmd *cobra.Command) error {
 	}
 	for _, ll := range logModules {
 		params := strings.Split(ll, ":")
+		if len(params) != 2 {
+			return fmt.Errorf("cmd: %s arg must be in form <module>:<level>, e.g. pubsub:debug", logLevelModuleFlag)
+		}
+
 		err := logging.SetLogLevel(params[0], params[1])
 		if err != nil {
 			return err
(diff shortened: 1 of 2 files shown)


vgonkivs (approved):

cmd/flags_misc.go
```diff
@@ -14,8 +15,9 @@ import (
 )
 
 var (
-	logLevelFlag = "log.level"
-	pprofFlag    = "pprof"
+	logLevelFlag       = "log.level"
+	logLevelModuleFlag = "log.level.module"
```
renaynay, line 19: can't this just be `log.module`

cmd/flags_misc.go
```diff
@@ -50,6 +58,18 @@ func ParseMiscFlags(cmd *cobra.Command) error {
 		logs.SetAllLoggers(level)
 	}
 
+	logModules, err := cmd.Flags().GetStringSlice(logLevelModuleFlag)
+	if err != nil {
+		return err
+	}
+	for _, ll := range logModules {
+		params := strings.Split(ll, ":")
+		err := logging.SetLogLevel(params[0], params[1])
```
renaynay, line 67: This could panic if user inputs log module the wrong way (doesn't use `:` or something). Could check that len(params) == 2.

Wondertan: Yeah, true.

Wondertan: Fixed

cmd/flags_misc.go
renaynay: lets just leave as is. I never seen a flag with two `.` in it 😂  

If we have complaints, we can simplify.

PR #786, opened 2022-06-01
share: Implement `cacheAvailability`

### LINKED ISSUE #180: share: Cache Availability

renaynay:
Resolves #180


renaynay:
should we keep a cache as well in the impl tho? but also persist to dis

Wondertan:
So to proceed, you need to import Datastore and store there the fact that DASing happened. This might be not the most efficient way to do this, but I have no other ideas right now(cc @liamsi or @adlerjohn). Further, if we make our own protocol for ShareExchange(ShrEx 😄 ) instead of Bitswap along with custom storage for shares this probably won't be needed. 

Besides, I would suggest identifying headers by their hashes and not heights.

Wondertan:
> should we keep a cache as well in the impl tho? but also persist to disk?

No need to. The caching in RAM is helpful when the value is accessed multiple times. In this case, we just need to know if the header was dased or no and DASer will do this once only.

adlerjohn:
What's the overhead of keeping track of this info? Can't be too much, can it? Bloom filters are...uh, not encouraged lol.

service/share/share.go
```diff
@@ -82,7 +82,7 @@ func (s *Service) Start(context.Context) error {
 	return nil
 }
 
-func (s *Service) Stop(context.Context) error {
```
Wondertan, line 85: ?

Wondertan: It is not used

service/share/local_availability.go
```diff
@@ -0,0 +1,55 @@
...
+func NewLocalAvailability(avail Availability, ds datastore.Datastore) (Availability, error) {
+	ds = namespace.Wrap(ds, localAvailabilityPrefix)
+	return &localAvailability{
+		avail: avail,
+		ds:    ds,
+	}, nil
+}
+
+// SharesAvailable will store, upon success, the hash of the given Root to disk.
+func (la localAvailability) SharesAvailable(ctx context.Context, root *Root) error {
```
Wondertan, line 36:
```suggestion
func (la *localAvailability) SharesAvailable(ctx context.Context, root *Root) error {
```

service/share/local_availability.go
```diff
@@ -0,0 +1,55 @@
...
+// to disk.
+type localAvailability struct {
+	avail Availability
+
+	ds datastore.Datastore
+}
+
+// NewLocalAvailability wraps the given Availability with an additional datastore
+// for sampling result caching.
+func NewLocalAvailability(avail Availability, ds datastore.Datastore) (Availability, error) {
```
Wondertan, line 27: Remove error? It is not used

service/share/local_availability.go
```diff
@@ -0,0 +1,55 @@
...
+	ds = namespace.Wrap(ds, localAvailabilityPrefix)
+	return &localAvailability{
+		avail: avail,
+		ds:    ds,
```
Wondertan, line 31: One IO optimization suggestion is to use https://github.com/ipfs/go-datastore/blob/master/autobatch/autobatch.go. It collects multiple IO ops and flushes them only after a certain threshold is reached.

Wondertan (commented):
Noice job!

Any better names for localAvailability? Local seems undescriptive to me in this case and localExchange tbh.

Caching is still possible on disk. And this is what we do here in some way.

Wondertan:
Another way to think about this is that we are storing the result of the availability, maybe the name can be something result.

service/share/local_availability.go
Wondertan: It also requests datastore.Batching which we do provide in DI

service/share/cache_availability.go
```diff
@@ -0,0 +1,56 @@
...
+	avail Availability
+
+	ds *autobatch.Datastore
+}
+
+// NewCacheAvailability wraps the given Availability with an additional datastore
+// for sampling result caching.
+func NewCacheAvailability(avail Availability, ds datastore.Batching) Availability {
+	ds = namespace.Wrap(ds, cacheAvailabilityPrefix)
+	autoDS := autobatch.NewAutoBatching(ds, 50)
```
Wondertan, line 30: Make it a global var so that we don't miss this, once we start making all our global vars configurable.

renaynay: https://github.com/celestiaorg/celestia-node/pull/786/commits/0c1f194bb4df5ec64e79341b37715c5fa18f1dba

service/share/cache_availability.go
```diff
@@ -0,0 +1,58 @@
...
+	avail Availability
+
+	ds *autobatch.Datastore
+}
+
+// NewCacheAvailability wraps the given Availability with an additional datastore
+// for sampling result caching.
+func NewCacheAvailability(avail Availability, ds datastore.Batching) Availability {
+	ds = namespace.Wrap(ds, cacheAvailabilityPrefix)
+	autoDS := autobatch.NewAutoBatching(ds, store.DefaultWriteBatchSize)
```
Wondertan, line 32: We should not make share pkg depend on the header store pkg. Also, it is totally fine to make those values independent and configured separately.

Wondertan (commented):
Last implementation comment, LGTM. I didn't look at the tests yet

service/share/cache_availability.go
renaynay: https://github.com/celestiaorg/celestia-node/pull/786/commits/671c56d1a13ed1fb737c3c55ee9f695946b14193

wdyt?

Wondertan: We should not share those values. They will be eventually defined independently in the config.

Wondertan (approved):

vgonkivs (approved):

service/share/cache_availability.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/ipfs/go-datastore/autobatch"
+	"github.com/ipfs/go-datastore/namespace"
+)
+
+var (
+	// DefaultWriteBatchSize defines the size of the batched header write.
+	// Headers are written in batches not to thrash the underlying Datastore with writes.
+	// TODO(@Wondertan, @renaynay): Those values must be configurable and proper defaults should be set for specific node
+	//  type. (#709)
+	DefaultWriteBatchSize = 2048
```
liamsi, line 16: So what happens if in between 2048 sample requests, the node goes down in between? Worst case scenario the root does not get stored and we resample, right? Not too bad I guess. But 2048 headers might be a bit too much. Also it is not really headers that are batched her but actually just root hashes, right?

service/share/cache_availability.go
```diff
@@ -0,0 +1,64 @@
...
+	return &cacheAvailability{
+		avail: avail,
+		ds:    autoDS,
+	}
+}
+
+// SharesAvailable will store, upon success, the hash of the given Root to disk.
+func (ca *cacheAvailability) SharesAvailable(ctx context.Context, root *Root) error {
+	// do not sample over Root that has already been sampled
+	exists, err := ca.ds.Has(rootKey(root))
```
liamsi, line 48: Extract a var for `rootKey(root)` and use it here and below?

service/share/cache_availability.go
```diff
@@ -0,0 +1,64 @@
...
+	if exists {
+		return nil
+	}
+	err = ca.avail.SharesAvailable(ctx, root)
+	if err != nil {
+		return err
+	}
+	err = ca.ds.Put(rootKey(root), []byte{})
+	if err != nil {
+		log.Errorw("storing result of successful SharesAvailable request to disk", "err", err)
```
liamsi, line 61:
```suggestion
		log.Errorw("storing root of successful SharesAvailable request to disk", "err", err)
```

service/share/cache_availability.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+func rootKey(root *Root) datastore.Key {
+	return datastore.NewKey(root.String())
+}
+
+// cacheAvailability wraps a given Availability (whether it's light or full)
+// and stores the results of a successful sampling routine over a given Root's hash
+// to disk.
+type cacheAvailability struct {
```
liamsi, line 28: I wonder if sth like `PersistedAvailability` could have been a better name?

renaynay: Yeah could be.

service/share/cache_availability.go
Wondertan: Yep, we are batching here hashes of dahs that were dased. The number is ok.

renaynay: I believe the underlying datastore that gets passed to `cacheAvailability`, upon stop, gets closed which flushes the queued up writes to disk. This is true, right @Wondertan ? Otherwise we should probably add the datastore's "lifecycle" to the node? Or i can make a hook for flushing writes on stop inside the constructor.

Wondertan: Yep, it is getting closed. The problem is with auto batch which is not getting closed and it wraps our datastore on the node. So we should close the auto batch first, so it can write all the outstanding writes to the datastore and then close it as well. This can be simply done by the Stop on cacheAvailability which stops the autobatch. The ordering is handled by fx

service/share/cache_availability.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+	err = ca.ds.Put(key, []byte{})
+	if err != nil {
+		log.Errorw("storing root of successful SharesAvailable request to disk", "err", err)
+	}
+	return err
+}
+
+// Stop flushes all queued writes to disk.
+func (ca *CacheAvailability) Stop(context.Context) error {
```
Wondertan, line 68: nit: if there is no `Start` symmetrical to the `Stop`, then it is better to call it `Close` and without ctx passed in it

service/share/cache_availability.go
```diff
@@ -0,0 +1,70 @@
...
+	autoDS := autobatch.NewAutoBatching(ds, DefaultWriteBatchSize)
+	return &CacheAvailability{
+		avail: avail,
+		ds:    autoDS,
+	}
+}
+
+// SharesAvailable will store, upon success, the hash of the given Root to disk.
+func (ca *CacheAvailability) SharesAvailable(ctx context.Context, root *Root) error {
+	// do not sample over Root that has already been sampled
```
Wondertan, line 47: This would be a nice place to make a special check for an empty DAH's Root. Currently, even if we store it locally, we still do a disk operation to check if it is there. Instead, we could short circuit here and simply return success if the given `root` equals the empty DAH. If you are going to do that, pls ensure empty DAH is allocated once, e.g, by private global var.

renaynay: Empty DAH is an error case though no? Do you mean min DAH?

Wondertan: Oh yeah, Min Dah

service/share/cache_availability.go
```diff
@@ -44,6 +44,10 @@ func NewCacheAvailability(avail Availability, ds datastore.Batching) *CacheAvail
 
 // SharesAvailable will store, upon success, the hash of the given Root to disk.
 func (ca *CacheAvailability) SharesAvailable(ctx context.Context, root *Root) error {
+	// short-circuit if given root is minimum DAH
```
Wondertan, line 47:
```suggestion
// short-circuit if the given root is minimum DAH of an empty data square
```

Wondertan (approved):

Wondertan (approved):

Wondertan (approved):

Wondertan (approved):

liamsi (approved):

Wondertan:
Oops, there was some bug on the phone.

PR #788, opened 2022-06-01
params: add WithBootstrappers option

### LINKED ISSUE #785: params: BootstrapPeers type and WithBootstrapPeers option for the node
Currently, there is a `params.Network` type that is registered on the Node that defaults to `params.Network` and has an option/modifier to alter it. Similar to this, we should have `params.Bootstrappers` type with defaults taken from `params.BootstrappersFor`(as there are right now) and with the option to modify those on the Node.

vgonkivs:
Resolves #785

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -36,10 +36,10 @@ var log = logging.Logger("node")
 // * Light
 // * Full
 type Node struct {
-	Type           Type
-	Network        params.Network
-	BootstrapPeers params.Bootstrappers
-	Config         *Config
+	Type          Type
+	Network       params.Network
+	Bootstrappers params.Bootstrappers
+	Config        *Config
 
 	// CoreClient provides access to a Core node process.
 	CoreClient core.Client `optional:"true"`
--- a/node/node.go
+++ b/node/node.go
@@ -36,9 +36,10 @@ var log = logging.Logger("node")
 // * Light
 // * Full
 type Node struct {
-	Type    Type
-	Network params.Network
-	Config  *Config
+	Type           Type
+	Network        params.Network
+	BootstrapPeers params.BootstrapPeers
+	Config         *Config
 
 	// CoreClient provides access to a Core node process.
 	CoreClient core.Client `optional:"true"`
--- a/node/node.go
+++ b/node/node.go
@@ -38,7 +38,7 @@ var log = logging.Logger("node")
 type Node struct {
 	Type           Type
 	Network        params.Network
-	BootstrapPeers params.BootstrapPeers
+	BootstrapPeers params.Bootstrappers
 	Config         *Config
 
 	// CoreClient provides access to a Core node process.
--- a/node/services/config.go
+++ b/node/services/config.go
@@ -30,14 +30,10 @@ func DefaultConfig() Config {
 	}
 }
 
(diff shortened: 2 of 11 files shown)


params/network.go
```diff
@@ -16,6 +18,9 @@ const (
 // Network is a type definition for DA network run by Celestia Node.
 type Network string
 
+// BootstrapPeers is a type definition for nodes that will be used as bootstrappers
+type BootstrapPeers []*peer.AddrInfo
```
Wondertan, line 22: We need to be consistent with namings. Our func in the pkg is called BootstrappersFor while the type is BootstrapPeer.

Wondertan: Also, now I think we should keep peer.AddrInfo as a value(not pointer) to be consistent with DHT and they libp2p uses them

node/tests/sync_test.go
```diff
@@ -281,3 +282,26 @@ func TestSyncLightWithTrustedPeers(t *testing.T) {
 
 	assert.EqualValues(t, h.Commit.BlockID.Hash, sw.GetCoreBlockHashByHeight(ctx, 50))
 }
+
+func TestUseBridgeNodeAsBootstraper(t *testing.T) {
```
Wondertan, line 286: Please make another file `p2p.go` and move it there.

Wondertan (commented):
That was quick, awesome! Thanks.

node/node.go
```diff
@@ -36,9 +36,10 @@ var log = logging.Logger("node")
 // * Light
 // * Full
 type Node struct {
-	Type    Type
-	Network params.Network
-	Config  *Config
+	Type           Type
+	Network        params.Network
+	BootstrapPeers params.Bootstrappers
```
Wondertan, line 41: Bootstrappers

params/bootstrap.go
```diff
@@ -9,17 +9,17 @@ import (
...
 		return nil, err
 	}
 
 	return parseAddrInfos(bs)
 }
 
-// BootstrappersFor reports multiaddresses of bootstrap peers for a given network.
-func BootstrappersFor(net Network) ([]string, error) {
+// bootstrappersFor reports multiaddresses of bootstrap peers for a given network.
+func bootstrappersFor(net Network) ([]string, error) {
```
Wondertan, line 22: If me make this private, we can omit Infos from BootstrappersInfoFor to make shorter. Also, BootstrappersInfoFor returns the type Bootstrappers and not the info for the type bootstrappers

Wondertan (approved):
Last bits and lfg

Wondertan (approved):

node/services/config.go
```diff
@@ -30,26 +30,23 @@ func DefaultConfig() Config {
...
-		cfg.TrustedPeers, err = params.BootstrappersFor(net)
-		if err != nil {
-			return
-		}
+		log.Infof("No trusted peers in config, initializing with default bootstrappers as trusted peers")
+		return bpeers, nil
 	}
 
-	infos = make([]*peer.AddrInfo, len(cfg.TrustedPeers))
+	infos = make([]peer.AddrInfo, len(cfg.TrustedPeers))
```
renaynay, line 39: Why this change?

vgonkivs: to be consistent wit libp2p. It requires `peer.AddrInfo` in `dht.BootstrapPeers()`

PR #799, opened 2022-06-06
Implement full node discovery

vgonkivs:


liamsi:
Should we throw above flow diagram into an ADR and summarize full node discovery there as well?

vgonkivs:
Yeah. I think it's a good point. Thanks.

node/tests/p2p_test.go
```diff
@@ -2,19 +2,31 @@ package tests
...
 	"github.com/libp2p/go-libp2p-core/peer"
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
 
 	"github.com/celestiaorg/celestia-node/node"
 	"github.com/celestiaorg/celestia-node/node/tests/swamp"
 )
 
+/*
+Test-Case: Sync a Light Node with a Bridge Node
```
Bidon15, line 22: Same TC naming as at line 54

node/tests/p2p_test.go
```diff
@@ -37,3 +49,59 @@ func TestUseBridgeNodeAsBootstraper(t *testing.T) {
...
+		node.WithBootstrappers([]peer.AddrInfo{*addr}),
+		node.WithRefreshRoutingTablePeriod(time.Second*30),
+	)
+	nodes := []*node.Node{full, light}
+	for index := range nodes {
+		require.NoError(t, nodes[index].Start(ctx))
+		assert.Equal(t, *addr, nodes[index].Bootstrappers[0])
+		assert.True(t, nodes[index].Host.Network().Connectedness(addr.ID) == network.Connected)
+	}
+	sub0, err := light.Host.EventBus().Subscribe(&event.EvtPeerConnectednessChanged{})
```
Bidon15, line 90: why sub0 🧐 ? 

sz_comp

node/tests/p2p_test.go
```diff
@@ -37,3 +49,59 @@ func TestUseBridgeNodeAsBootstraper(t *testing.T) {
...
+		select {
+		case <-ctx.Done():
+			t.Fatal("peer was not found")
+		case p := <-sub0.Out():
+			ev := p.(event.EvtPeerConnectednessChanged)
+			if ev.Peer == full.Host.ID() && ev.Connectedness == network.Connected {
+				break LOOP
+			}
+		}
+	}
```
Bidon15, line 104: Proposal: It can be better to just move this into swamp. wdyt?

Bidon15 (changes requested):

service/share/testing.go
```diff
@@ -138,3 +177,7 @@ func NewBrokenAvailability() Availability {
 func (b *brokenAvailability) SharesAvailable(context.Context, *Root) error {
 	return ErrNotAvailable
 }
+
+func (b *brokenAvailability) StartPeerDiscovery() {}
+
+func (b *brokenAvailability) StopPeerDiscovery() {}
```
Bidon15, line 183: Can we add more comments/functionality or remove these?

vgonkivs: We need these method so brokenAvailability can implement `Availability` interface.

node/tests/p2p_test.go
vgonkivs: There were his brothers - sub1 and sub2. But they were destroyed 😄

node/tests/p2p_test.go
vgonkivs: nice catch

node/tests/p2p_test.go
vgonkivs: reworked this logic. now it looks better but I do not think that we should extract it into separate method in swamp.

Bidon15: Cool. Then no need. 🦜

service/share/testing.go
Bidon15: Then please leave a comment for them. Otherwise, this can be removed by accident later 🙏

Bidon15 (commented):
Some nits and a lgtm from me 💯

service/share/discovery/peer_cache.go
```diff
@@ -0,0 +1,48 @@
+package discovery
+
+import (
+	"sync"
+
+	"github.com/libp2p/go-libp2p-core/peer"
+)
+
+// PeerCache holds all discovered and available peers
+type PeerCache struct {
```
Wondertan, line 10: I just realized there is a `peer.Set` that has exactly this.

service/share/discovery/discovery.go
```diff
@@ -0,0 +1,66 @@
...
+
+			wait := 7 * ttl / 8
+			select {
+			case <-time.After(wait):
+			case <-ctx.Done():
+				return
+			}
+		}
+	}()
+}
```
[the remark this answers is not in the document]
vgonkivs: Yes, the function is available but the time interval is 2 minutes between announcings and it's could not be configurable.

service/share/interface.go
```diff
@@ -18,4 +18,9 @@ const AvailabilityTimeout = 20 * time.Minute
 type Availability interface {
 	// SharesAvailable subjectively validates if Shares committed to the given Root are available on the Network.
 	SharesAvailable(context.Context, *Root) error
+
+	// StartPeerDiscovery starts finding new peers in network.
+	StartPeerDiscovery()
+	// StopPeerDiscovery stops finding new peers.
+	StopPeerDiscovery()
```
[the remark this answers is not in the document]
vgonkivs: Yes, we did. But what actually can `Availability` start? It's not obvious that it is starting discovery IMO.
The Start and Stop also do not have to be part of the interface and can be only on the implementation.

> The Start and Stop also do not have to be part of the interface and can be only on the implementation.

Then StartDiscovery should be called in constructor.

service/share/discovery/notifee.go
```diff
@@ -0,0 +1,86 @@
...
+		if peer.ID == n.host.ID() || len(peer.Addrs) == 0 || n.cache.Has(peer.ID) {
+			continue
+		}
+		if n.host.Network().Connectedness(peer.ID) != network.Connected {
+			ctx, cancel = context.WithTimeout(context.Background(), time.Minute*1)
+			err := n.host.Connect(ctx, peer)
+			if err != nil {
+				log.Warn(err)
+				continue
+			}
```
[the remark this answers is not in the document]
vgonkivs: Are you sure the we should go this way? If connection failed, then we can just wait for another peer.

service/share/interface.go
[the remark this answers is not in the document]
vgonkivs: reworked this functionality. please take a look at https://github.com/celestiaorg/celestia-node/pull/799/commits/8f3e93c6f6e85071fe077b037e166dc8e650c2d5

service/share/discovery/peer_cache.go
vgonkivs: done

service/share/discovery/notifee.go
```diff
@@ -0,0 +1,86 @@
...
+func (n *Notifee) HandlePeersFound(topic string, peers []peer.AddrInfo) error {
+	var (
+		ctx    context.Context
+		cancel context.CancelFunc
+	)
+
+	for _, peer := range peers {
+		if cancel != nil {
+			cancel()
+		}
```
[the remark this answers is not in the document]
vgonkivs: moved inside the loop

service/share/discovery/notifee.go
```diff
@@ -0,0 +1,86 @@
...
+	emitter, err := n.host.EventBus().Emitter(&event.EvtPeerConnectednessChanged{})
+	if err != nil {
+		log.Warn(err)
+		return
+	}
+	err = emitter.Emit(event.EvtPeerConnectednessChanged{Peer: id, Connectedness: state})
+	if err != nil {
+		log.Warn(err)
+	}
+}
```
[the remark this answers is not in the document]
vgonkivs: Used `network.Bundle` in test to 'subscribe' on connection.

service/share/testing.go
vgonkivs: removed at all

node/tests/p2p_test.go
vgonkivs: reworked the rework 😄

service/share/discovery/discovery.go
```diff
@@ -0,0 +1,66 @@
...
+
+				select {
+				case <-time.After(interval):
+					continue
+				case <-ctx.Done():
+					return
+				}
+			}
+
+			wait := 7 * ttl / 8
```
Bidon15, line 35: Will it be better to declare `7` and `8` as const var at the top?

node/tests/p2p_test.go
Bidon15: Proposal: 
```suggestion
Test-Case: Full/Light Nodes connection to Bridge as a Bootstapper
```

service/share/discovery/discovery.go
```diff
@@ -0,0 +1,66 @@
...
+			}
+		}
+	}()
+}
+
+// FindPeers starts peer discovery every 2 minutes until peer cache will not reach peersLimit.
+// Can be simplified when https://github.com/libp2p/go-libp2p/pull/1379 will be merged.
+func FindPeers(ctx context.Context, d core.Discoverer, n *Notifee) {
+	go func() {
+		t := time.NewTicker(interval * 3)
```
Bidon15, line 49: In the comment section, we are referring to 2 minutes. However we have a ticker of 30 seconds. 
Can we elaborate the message better?

Bidon15 (commented):
👏 🎬

service/share/discovery/discovery.go
vgonkivs: added a calculation as a var in the top

service/share/discovery/discovery.go
vgonkivs: left from previous version. changed to 30 seconds.

service/share/discovery/peer_cache.go
vgonkivs: I just noticed that `peer.Set` does not provide functionality to remove peer from set. So, I will add a custom set.

service/share/cache_availability.go
```diff
@@ -27,18 +29,22 @@ var (
 type CacheAvailability struct {
 	avail Availability
 
-	ds *autobatch.Datastore
+	cancel context.CancelFunc
+
+	ds         *autobatch.Datastore
+	discoverer *discovery.Discoverer
```
renaynay, line 35:
```suggestion
	disc *discovery.Discoverer
```

renaynay: nicer to shorten

service/share/discovery/discoverer.go
```diff
@@ -0,0 +1,75 @@
...
+
+	"github.com/libp2p/go-libp2p-core/discovery"
+	"github.com/libp2p/go-libp2p-core/host"
+	"github.com/libp2p/go-libp2p-core/peer"
+)
+
+const (
+	// PeersLimit is max amount of peers that will be discovered
+	PeersLimit = 5
+	// peerWeight total weight of discovered peers
```
renaynay, line 18:
```suggestion

	// peerWeight total weight of discovered peers
```

we should separate exported vars from unexported

service/share/discovery/discoverer.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+var log = logging.Logger("discovery")
+
+// Discoverer finds and stores discovered full nodes.
+type Discoverer struct {
+	set  *LimitedSet
+	host host.Host
+
+	discoverer discovery.Discoverer
```
renaynay, line 31:
```suggestion
	disc discovery.Discoverer
```

service/share/discovery/discoverer.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+func (d *Discoverer) Start(ctx context.Context) {
```
renaynay, line 34: we don't need this imo since it's not conforming to `Start` hook signature (i.e. `Start(ctx) error`)

just call `FindPeers` instead.

service/share/discovery/discovery.go
```diff
@@ -0,0 +1,72 @@
...
+
+var (
+	// waitF calculates time to restart announcing.
+	waitF = func(ttl time.Duration) time.Duration {
+		return 7 * ttl / 8
+	}
+)
+
+// Advertise is a utility function that persistently advertises a service through an Advertiser.
+func Advertise(ctx context.Context, a core.Advertiser) {
```
renaynay, line 24: Nit: for both `Advertise` and `FindPeers`, IMO it's more idiomatic to do: 

```
func main() {
  go DoSomethingAsync()
}

func DoSomethingAsync() {
  // some worker
}
```

service/share/discovery/discovery.go
```diff
@@ -0,0 +1,72 @@
...
+			case <-time.After(waitF(ttl)):
+			case <-ctx.Done():
+				return
+			}
+		}
+	}()
+}
+
+// FindPeers starts peer discovery every 30 seconds until peer cache will not reach peersLimit.
+// Can be simplified when https://github.com/libp2p/go-libp2p/pull/1379 will be merged.
```
renaynay, line 52: Pls add TODO here referencing yourself `TODO @vgonkivs: Simplify when ....`

service/share/discovery/discovery.go
```diff
@@ -0,0 +1,72 @@
...
+		t := time.NewTicker(interval * 3)
+		defer t.Stop()
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-t.C:
+				peers, err := discovery.FindPeers(ctx, d.discoverer, namespace)
+				if err != nil {
+					continue
```
renaynay, line 64: pls debug log err

service/share/discovery/discovery.go
```diff
@@ -0,0 +1,72 @@
...
+				}
+				if err = d.handlePeersFound(namespace, peers); err != nil {
+					return
```
renaynay, line 67: also pls warn log here bc you're exiting this loop

service/share/discovery/set.go
```diff
@@ -0,0 +1,69 @@
+package discovery
```
renaynay, line 1: Pls add note with why this is copy/pasted and link to issue in libp2p

renaynay: Please also add test files / unit tests for this structure

service/share/discovery/set.go
```diff
@@ -0,0 +1,69 @@
...
+	return ok
+}
+
+func (ps *LimitedSet) Size() int {
+	ps.lk.RLock()
+	defer ps.lk.RUnlock()
+	return len(ps.ps)
+}
+
+// TryAdd Attempts to add the given peer into the set.
```
renaynay, line 40:
```suggestion
// TryAdd attempts to add the given peer into the set.
```

service/share/discovery/set.go
```diff
@@ -0,0 +1,69 @@
...
+// This operation will fail if the number of peers in the set is equal to size
```
renaynay, line 41:
```suggestion
// This operation will fail if the number of peers in the set is equal to size.
```

service/share/light_availability.go
```diff
@@ -13,24 +13,25 @@ import (
...
+	la := &LightAvailability{
 		bserv: bserv,
 	}
+	return la
 }
 
 // SharesAvailable randomly samples DefaultSamples amount of Shares committed to the given Root.
 // This way SharesAvailable subjectively verifies that Shares are available.
-func (la *lightAvailability) SharesAvailable(ctx context.Context, dah *Root) error {
+func (la *LightAvailability) SharesAvailable(ctx context.Context, dah *Root) error {
```
renaynay, line 34: why export? Same question for full avail?

renaynay (changes requested):

service/share/discovery/discovery.go
vgonkivs: only `peers limit reached` is expected here. All other errors are printed inside `handlePeersFound`

service/share/discovery/set.go
vgonkivs: I think NOTE is not relevant now bc `peer.Set` was removed at all.

service/share/light_availability.go
vgonkivs: forgot to push commit with reverting these changes.

service/share/discovery/discoverer.go
vgonkivs: Moved FindPeers to share.Service as we discussed.

service/share/discovery/discovery.go
renaynay: Lets maybe Info log that then?

service/share/discovery/set.go
renaynay: Did they remove it already? Damn.

renaynay (approved):
LGTM!

service/share/discovery/set.go
vgonkivs: yeah

Bidon15 (approved):

Wondertan (commented):
Let's try updating libp2p next week. We are past 5 versions already and rebase + reimplement some parts over newest versions. My bad that i didn't push this earlier, but we really overdue for updating the p2p stack.

adlerjohn (commented):
Merge conflicts

liamsi (approved):

Bidon15 (approved):

adlerjohn (approved):
utACK

service/share/testing.go
```diff
@@ -242,3 +250,7 @@ func NewBrokenAvailability() Availability {
 func (b *brokenAvailability) SharesAvailable(context.Context, *Root) error {
 	return ErrNotAvailable
 }
+
+func Null() *routing.RoutingDiscovery {
```
Wondertan, line 254: `share.Nill()` returning RoutingDiscoverry looks off. WDYT about making the below instead
```go
func TestLightAvailability(t *testing.T)
``` 

which would create mock blockservice and nil routing inside, so that Daser tests could use it directly.

service/share/set.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"errors"
+	"sync"
+
+	"github.com/libp2p/go-libp2p-core/peer"
+)
+
+// LimitedSet is a thread safe set of peers with given limit.
+// Inspired by libp2p peer.Set but extended with Remove method.
+type LimitedSet struct {
```
Wondertan, line 12: Also private like the constructor?

service/share/discovery.go
```diff
@@ -0,0 +1,46 @@
...
+			}
+		}
+
+		select {
+		case <-time.After(waitF(ttl)):
+		case <-ctx.Done():
+			return
+		}
+	}
+}
```
Wondertan, line 46: Should we be consistent with discover, by making it singe purpose and hidden in  `share` as well?

service/share/discovery.go
```diff
@@ -0,0 +1,46 @@
...
+				continue
+			case <-ctx.Done():
+				return
+			}
+		}
+
+		select {
+		case <-time.After(waitF(ttl)):
+		case <-ctx.Done():
+			return
```
Wondertan, line 43: Pls, also rework it with time.Timer instead of time.After, due to known and documented inefficiencies on it

node/services/service.go
```diff
@@ -167,21 +175,38 @@ func LightAvailability(
...
+	h host.Host,
 ) share.Availability {
-	la := share.NewLightAvailability(bServ)
+	la := share.NewLightAvailability(bServ, routingdisc.NewRoutingDiscovery(r), h)
 	ca := share.NewCacheAvailability(la, ds)
 	lc.Append(fx.Hook{
-		OnStop: ca.Close,
+		OnStart: la.Start,
+		OnStop: func(ctx context.Context) error {
+			_ = la.Stop(ctx)
```
Wondertan, line 186: I know that we don't return an error there, but we might at some point and when we do this, we can forget the error here, so let's add a check here. Same with Full

service/share/discovery.go
vgonkivs: Yes 👍

node/services/service.go
Wondertan: Ideally, we should do this instead https://github.com/celestiaorg/celestia-node/issues/879.
Maybe this fix could be part of that issue.

service/share/light_availability.go
```diff
@@ -6,31 +6,55 @@ import (
...
 // DefaultSampleAmount sets the default amount of samples to be sampled from the network by lightAvailability.
 var DefaultSampleAmount = 16
 
-// lightAvailability implements Availability using Data Availability Sampling technique.
+// LightAvailability implements Availability using Data Availability Sampling technique.
 // It is light because it does not require the downloading of all the data to verify
 // its availability. It is assumed that there are a lot of lightAvailability instances
 // on the network doing sampling over the same Root to collectively verify its availability.
-type lightAvailability struct {
+type LightAvailability struct {
```
renaynay, line 22: also why exported?

service/share/discovery.go
vgonkivs: done

node/services/service.go
vgonkivs: added multierr for now

service/share/full_availability.go
```diff
@@ -6,27 +6,49 @@ import (
...
 	"github.com/celestiaorg/celestia-node/ipld"
 )
 
-// fullAvailability implements Availability using the full data square
+// FullAvailability implements Availability using the full data square
 // recovery technique. It is considered "full" because it is required
 // to download enough shares to fully reconstruct the data square.
-type fullAvailability struct {
-	rtrv *ipld.Retriever
+type FullAvailability struct {
```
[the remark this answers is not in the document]
vgonkivs: To be able to call Start/Stop without extending `Availability` interface. Returning unexported struct will cause linter warning.

service/share/discovery.go
```diff
@@ -0,0 +1,134 @@
...
+		d,
+	}
+}
+
+// handlePeersFound receives peers and tries to establish a connection with them.
+// Peer will be added to PeerCache if connection succeeds.
+func (d *discovery) handlePeersFound(topic string, peers []peer.AddrInfo) error {
+	for _, peer := range peers {
+		if d.set.Size() == PeersLimit {
+			return errors.New("amount of peers reaches the limit")
```
Wondertan, line 59:
```suggestion
	return fmt.Errof("discovered enough(%d) peers", PeersLimit)
```

service/share/discovery.go
```diff
@@ -0,0 +1,134 @@
...
+		h,
+		d,
+	}
+}
+
+// handlePeersFound receives peers and tries to establish a connection with them.
+// Peer will be added to PeerCache if connection succeeds.
+func (d *discovery) handlePeersFound(topic string, peers []peer.AddrInfo) error {
+	for _, peer := range peers {
+		if d.set.Size() == PeersLimit {
```
Wondertan, line 58: What's the point of this check if TryAdd already does it?

service/share/discovery.go
```diff
@@ -0,0 +1,134 @@
...
+	t := time.NewTicker(interval * 3)
+	defer t.Stop()
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-t.C:
+			peers, err := util.FindPeers(ctx, d.disc, topic)
+			if err != nil {
+				log.Debug(err)
```
Wondertan, line 96: Error?

service/share/discovery.go
```diff
@@ -0,0 +1,134 @@
...
+		newLimitedSet(PeersLimit),
+		h,
+		d,
+	}
+}
+
+// handlePeersFound receives peers and tries to establish a connection with them.
+// Peer will be added to PeerCache if connection succeeds.
+func (d *discovery) handlePeersFound(topic string, peers []peer.AddrInfo) error {
+	for _, peer := range peers {
```
Wondertan, line 57: Let's handle each peer in a separate routine. The Connect call inside is blocking and may take up to 1 minute.

service/share/discovery.go
```diff
@@ -0,0 +1,134 @@
...
+
+		if peer.ID == d.host.ID() || len(peer.Addrs) == 0 || d.set.Contains(peer.ID) {
+			continue
+		}
+		err := d.set.TryAdd(peer.ID)
+		if err != nil {
+			return err
+		}
+
+		ctx, cancel := context.WithTimeout(context.Background(), connectionTimeout)
```
Wondertan, line 70: Ideally, this should be context on the discoverer. There is no need for a ctx per connection. Also, it is better to use this https://github.com/libp2p/go-libp2p-core/blob/master/network/context.go#L93 for the purpose

service/share/discovery.go
```diff
@@ -0,0 +1,134 @@
...
+	// waitF calculates time to restart announcing.
+	waitF = func(ttl time.Duration) time.Duration {
+		return 7 * ttl / 8
+	}
+)
+
+// discovery implements discoverer interface to avoid using it in light availability.
+type discoverer interface {
+	findPeers(ctx context.Context)
+}
```
Wondertan, line 36: If it used only once in `light`, then why?

Wondertan (commented):
Almost there

service/share/discoverer.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/libp2p/go-libp2p-core/host"
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/libp2p/go-libp2p/p2p/discovery/util"
+)
+
+const (
+	// PeersLimit is max amount of peers that will be discovered
+	PeersLimit = 5
+
+	// peerWeight total weight of discovered peers
```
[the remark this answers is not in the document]
Wondertan: I  believe 1000 is arbitrary and just high enough to ensure ConnectionManager won't eve trim such connections

service/share/discovery.go
vgonkivs: to protect calling advertise from light node

service/share/full_availability.go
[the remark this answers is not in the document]
renaynay: I'd rather add `Start/Stop` to the availability interface.

Then we can unexport CacheAvailability and call `Close` -> `Stop`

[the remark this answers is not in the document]
Wondertan: Lets' not add Start/Stop to the interfaces.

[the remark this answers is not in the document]
Wondertan: Or generally, let's not add to interfaces things that can be omitted and implementation-specific.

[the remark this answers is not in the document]
vgonkivs: Agree with @Wondertan.

service/share/discoverer.go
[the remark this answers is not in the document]
vgonkivs: Yes.

service/share/discovery.go
vgonkivs: used context that is passed to handlePeer

service/share/discovery.go
vgonkivs: handlePeerFound is called in a separate routine for each peer.

vgonkivs:
[p2p discovery.pdf](https://github.com/celestiaorg/celestia-node/files/9040638/p2p.discovery.pdf)

vgonkivs:
discovery

service/share/discovery.go
```diff
@@ -0,0 +1,130 @@
...
+	topic      = "full"
+	interval   = time.Second * 10
+)
+
+// waitF calculates time to restart announcing.
+var waitF = func(ttl time.Duration) time.Duration {
+	return 7 * ttl / 8
+}
+
+// discoverer used to protect light nodes of being advertised as they support only peer discovery.
```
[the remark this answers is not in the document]
vgonkivs: Sure

docs/adr/adr-008-p2p-discovery.md
```diff
@@ -0,0 +1,55 @@
...
+1. Node starts finding full nodes to be able to establish a connection with them using `discoverer` interface.
+2. As soon as a new peer is found, node will try to establish a connection with it. In case if the connection is successful
+   node will Tag Peer and add to the internal storage.
+
+Tagging allows to protect peer from being killed by ConnManager.
+
+![discovery](https://user-images.githubusercontent.com/40579846/177183260-92d1c390-928b-4c06-9516-24afea94d1f1.png)
+
+## Status
+Proposed
```
Wondertan, line 55: Just a small reminder to change to merged once we all approve this.

docs/adr/adr-008-p2p-discovery.md
```diff
@@ -0,0 +1,55 @@
...
+	// libp2p.Discovery allows to advertise and discover peers.
+	disc core.Discovery
+}
+
+// discoverer used to protect light nodes of being advertised as they support only peer discovery.
+type discoverer interface {
+    findPeers(ctx context.Context)
+}
+```
+## Full Nodes behavior:
```
Wondertan, line 38:
```suggestion
### Full Nodes behavior:
```

docs/adr/adr-008-p2p-discovery.md
```diff
@@ -0,0 +1,55 @@
...
+1. Node starts advertising itself at `full` topic after system boots up in order to be found.
+2. Node starts finding other full nodes to be able to establish a connection with them.
+3. As soon as a new peer is found, node will try to establish a connection with it. In case if the connection is successful
+node will Tag Peer and add to the internal storage.
+
+
+## Light Nodes behavior:
```
Wondertan, line 45:
```suggestion
### Light Nodes behavior:
```

docs/adr/adr-008-p2p-discovery.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Changelog
+
+- 2022.07.04 - discovery data structure
+
+## Authors
+
+@vgonkivs
+
+## P2P Full Node Discovery
```
Wondertan, line 11:
```suggestion
```

docs/adr/adr-008-p2p-discovery.md
```diff
@@ -0,0 +1,55 @@
...
+	disc core.Discovery
+}
+
+// discoverer used to protect light nodes of being advertised as they support only peer discovery.
+type discoverer interface {
+    findPeers(ctx context.Context)
+}
+```
+## Full Nodes behavior:
+1. Node starts advertising itself at `full` topic after system boots up in order to be found.
```
Wondertan, line 39:
```suggestion
1. Node starts advertising itself over DHT at `full` namespace after the system boots up in order to be found.
```

docs/adr/adr-008-p2p-discovery.md
```diff
@@ -0,0 +1,55 @@
...
+2. Node starts finding other full nodes to be able to establish a connection with them.
```
Wondertan, line 40:
```suggestion
2. Node starts finding other full nodes to be able to join the Full Node network.
```

docs/adr/adr-008-p2p-discovery.md
```diff
@@ -0,0 +1,55 @@
...
+3. As soon as a new peer is found, node will try to establish a connection with it. In case if the connection is successful
+node will Tag Peer and add to the internal storage.
```
Wondertan, line 42:
```suggestion
3. As soon as a new peer is found, the node will try to establish a connection with it. In case the connection is successful
the node will Tag Peer and add to the internal storage.
```

Wondertan: WDYM by internal storage btw? We don't store connections on disk

docs/adr/adr-008-p2p-discovery.md
```diff
@@ -0,0 +1,55 @@
...
+
+
+## Light Nodes behavior:
+1. Node starts finding full nodes to be able to establish a connection with them using `discoverer` interface.
```
Wondertan, line 46:
```suggestion
1. Node starts finding full nodes over DHT at "full" namespace using the `discoverer` interface.
```

docs/adr/adr-008-p2p-discovery.md
```diff
@@ -0,0 +1,55 @@
...
+2. As soon as a new peer is found, node will try to establish a connection with it. In case if the connection is successful
+   node will Tag Peer and add to the internal storage.
```
Wondertan, line 48:
```suggestion
2. As soon as a new peer is found, the node will try to establish a connection with it. In case the connection is successful
   the node will Tag Peer and add to the internal storage.
```

Wondertan: Same question, also worth explaining what we mean by Tag

Wondertan (commented):
ADR review

service/share/discovery.go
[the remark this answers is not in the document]
Wondertan: We discussed this privately as well and its ok

docs/adr/adr-008-p2p-discovery.md
vgonkivs: sure. thanks for pointing out.

service/share/discovery.go
```diff
@@ -0,0 +1,126 @@
...
+			}
+			for peer := range peers {
+				go d.handlePeerFound(ctx, topic, peer)
+			}
+		}
+	}
+}
+
+// advertise is a utility function that persistently advertises a service through an Advertiser.
+func (d *discovery) advertise(ctx context.Context) {
```
adlerjohn, line 100: For consistency, could we use American English, i.e. adverti**z**e? (this coming from a Canadian)

service/share/set.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/libp2p/go-libp2p-core/peer"
+)
+
+// limitedSet is a thread safe set of peers with given limit.
+// Inspired by libp2p peer.Set but extended with Remove method.
+type limitedSet struct {
+	lk sync.RWMutex
+	ps map[peer.ID]struct{}
+
+	limit int
```
adlerjohn, line 16: Will this ever be negative?

service/share/discovery.go
```diff
@@ -0,0 +1,130 @@
...
+	core "github.com/libp2p/go-libp2p-core/discovery"
+	"github.com/libp2p/go-libp2p-core/host"
+	"github.com/libp2p/go-libp2p-core/peer"
+)
+
+const (
+	// PeersLimit is max amount of peers that will be discovered.
+	PeersLimit = 5
+
+	// peerWeight total weight of discovered peers.
```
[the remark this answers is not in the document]
vgonkivs: added more information.

service/share/set.go
vgonkivs: https://github.com/celestiaorg/celestia-node/pull/799/commits/814b3b677daf7cbffa5b987f1a97611aeeb1cfa3

docs/adr/adr-008-p2p-discovery.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Authors
+
+@vgonkivs
+
+## Context
+
+This ADR is intended to describe p2p full node discovery in celestia node. 
+P2P discovery helps for light and full nodes to find other full nodes on the network at specified topic(`full`).
+As soon as full node is found and connection is established with it, then it will be added to a set of peers(limitedSet).
```
Bidon15, line 15:
```suggestion
P2P discovery helps light and full nodes to find other full nodes on the network at the specified topic(`full`).
As soon as a full node is found and connection is established with it, then it(full node) will be added to a set of peers(limitedSet).
```
```suggestion
P2P discovery helps for light and full nodes to find other full nodes on the network at specified topic(`full`).
As soon as full node is found and connection is established with it, then it will be added to a set of peers(limitedSet).
```
```suggestion
P2P discovery helps for light and full nodes to find other full nodes on the network at specified topic(`full`).
As soon as full node is found and connection is established with it, then it will be added to a set of peers(limitedSet).
```

Bidon15: Please elaborate more on what is the `limitedSet`. Is it `peersLimit`? 🧐

docs/adr/adr-008-p2p-discovery.md
```diff
@@ -0,0 +1,74 @@
...
+type limitedSet struct {
+    lk sync.RWMutex
+    ps map[peer.ID]struct{}
+
+    limit int
+}
+```
+### Full Nodes behavior:
+1. Node starts advertising itself over DHT at `full` namespace after the system boots up in order to be found.
+2. Node starts finding other full nodes to be able to join the Full Node network.
```
Bidon15, line 51:
```suggestion
1. A Node starts advertising itself over DHT at `full` namespace after the system boots up in order to be found.
2. A Node starts finding other full nodes, so it can be able to join the Full Node network.
```
```suggestion
1. Node starts advertising itself over DHT at `full` namespace after the system boots up in order to be found.
2. Node starts finding other full nodes to be able to join the Full Node network.
```
```suggestion
1. Node starts advertising itself over DHT at `full` namespace after the system boots up in order to be found.
2. Node starts finding other full nodes to be able to join the Full Node network.
```

docs/adr/adr-008-p2p-discovery.md
```diff
@@ -0,0 +1,74 @@
...
+3. As soon as a new peer is found, the node will try to establish a connection with it. In case the connection is successful
```
Bidon15, line 52: If we add `in case` of A and describing the outcome of scenario A, then it's good to add a `case` B if the opposite happens (tbe if the connection is not successful)

docs/adr/adr-008-p2p-discovery.md
```diff
@@ -0,0 +1,74 @@
...
+the node will call [Tag Peer](https://github.com/libp2p/go-libp2p-core/blob/525a0b13017263bde889a3295fa2e4212d7af8c5/connmgr/manager.go#L35) and add peer to the peer set.
+
+
+### Light Nodes behavior:
+1. Node starts finding full nodes over DHT at `full` namespace using the `discoverer` interface.
```
Bidon15, line 57: Some as for full - A Node

Bidon15 (changes requested):
Some nits and we are good to go 💪

service/share/discovery.go
```diff
@@ -0,0 +1,126 @@
...
+			if d.set.Size() >= peersLimit {
+				return
+			}
+			peers, err := d.disc.FindPeers(ctx, topic)
+			if err != nil {
+				log.Error(err)
+				continue
+			}
+			for peer := range peers {
+				go d.handlePeerFound(ctx, topic, peer)
```
Wondertan, line 93: We need to make an issue(and likely implement the solution in #829) about limiting the number of possible routines. Some adversaries could create bots advertising them as "full" which would eventually over spawn the number of routines. The problem would also be visible on big networks in normal operation, but it is still there.  

* The easiest solution: just do a buffered throttling channel. Write to it once handling started and read on the finish. This way the buffer of the chan would be a limit of concurrent routines/handlers/connectors.
* A bit more work, while more efficient is to preallocate a limited number of workers who rcv handling jobs from findPeers.

Bidon15:
I have no idea why github thinks duplicating suggestions(that I intended to create only once per line of code) is a good idea 😐 

Sorry for that @vgonkivs 🙏

docs/adr/adr-008-p2p-discovery.md
```diff
@@ -0,0 +1,74 @@
...
+2. As soon as a new peer is found, the node will try to establish a connection with it. In case the connection is successful
+   the node will call [Tag Peer](https://github.com/libp2p/go-libp2p-core/blob/525a0b13017263bde889a3295fa2e4212d7af8c5/connmgr/manager.go#L35) and add peer to the peer set.
+
+*NOTE* Bridge node behaves the same as Full nodes.
+
+Tagging allows to protect peer from being killed by ConnManager.
```
Wondertan, line 63:
```suggestion
Tagging protects connections from ConnManager trimming/GCing. 
```

Wondertan (approved):
Aprooove! Good job!
A list of things for us yet needs to be done that will likely land in a subsequent PRs:
* Handling disconnected peers which we originally discovered
	* Removing Tag from them
	* Backing off them, so we don't reconnect them within a certain period
	* Find more peers if we are below the peerLimit(#829)
* Throttle handlePeerFound goroutines(a comment below)
* We need to update the reconstruction tests to include the discovery bit in them

service/share/discovery.go
renaynay: I also use advertise

service/share/full_availability.go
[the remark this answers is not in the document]
renaynay: I think lifecycle is one of those cases that is applicable to almost everything, so I'm fine having Availability implement Start/Stop. Yes generally I also oppose implementations having empty methods, but the only empty method for Start/Stop now would be CacheAvailability Start which is ok IMO bc it would just call underlying avail `Start` so that we only need to add the cache availability lifecycle to fx.

CacheAvailability `Close` -> `Stop` and can wrap underlying avail stop

service/share/full_availability.go
```diff
@@ -6,27 +6,47 @@ import (
...
 	}
 }
 
+func (fa *FullAvailability) Start(context.Context) error {
+	ctx, cancel := context.WithCancel(context.Background())
+	fa.cancel = cancel
+	go fa.disc.advertise(ctx)
+	go fa.disc.findPeers(ctx)
+
+	return nil
```
renaynay, line 39: nit
```suggestion

	go fa.disc.advertise(ctx)
	go fa.disc.findPeers(ctx)
	return nil
```

service/share/light_availability.go
```diff
@@ -7,31 +7,56 @@ import (
...
 	bserv blockservice.BlockService
+	// disc is using to discovered new nodes in network.
+	// it's not allowed to call advertise for light nodes(Full nodes only).
```
renaynay, line 26:
```suggestion
	// disc discovers new full nodes in the network.
	// it is not allowed to call advertise for light nodes (Full nodes only).
```

service/share/discovery.go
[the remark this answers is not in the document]
renaynay: ack

service/share/testing.go
```diff
@@ -246,3 +254,11 @@ func (b *brokenAvailability) SharesAvailable(context.Context, *Root) error {
 func (b *brokenAvailability) ProbabilityOfAvailability() float64 {
 	return 0
 }
+
+func TestLigthAvailability(bServ blockservice.BlockService) *LightAvailability {
```
renaynay, line 258:
```suggestion
func TestLightAvailability(bServ blockservice.BlockService) *LightAvailability {
```

renaynay (commented):
few small nit

docs/adr/adr-008-p2p-discovery.md
```diff
@@ -0,0 +1,69 @@
...
+3. As soon as a new peer is found, the node will try to establish a connection with it. In case the connection is successful
+the node will call [Tag Peer](https://github.com/libp2p/go-libp2p-core/blob/525a0b13017263bde889a3295fa2e4212d7af8c5/connmgr/manager.go#L35) and add peer to the peer set.
+
+
+### Light Nodes behavior:
+1. Node starts finding full nodes over DHT at `full` namespace using the `discoverer` interface.
+2. As soon as a new peer is found, the node will try to establish a connection with it. In case the connection is successful
+   the node will call [Tag Peer](https://github.com/libp2p/go-libp2p-core/blob/525a0b13017263bde889a3295fa2e4212d7af8c5/connmgr/manager.go#L35) and add peer to the peer set.
+
+*NOTE* Bridge node behaves the same as Full nodes.
```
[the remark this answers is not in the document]
Wondertan: Full Node behavior includes both advertising and discovering and bridge behaves the same

service/share/full_availability.go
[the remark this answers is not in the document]
Wondertan: There are a few other points here that we should also consider. Let's extract this discussion to some issue and not block the PR on it.

service/share/discovery.go
Wondertan: advertiSe seems to be "Briiish" as well

Bidon15 (approved):
🚢

PR #802, opened 2022-06-07
service/rpc: Add `/data_available` endpoint

renaynay:
This PR adds another endpoint: `/data_available`.

## GET `/data_available/{height}`

Request

```
curl -X GET http://<ip>:26658/data_available/25000
```

Response

```
{"available":true,"probability_of_availability":"0.9899774042423815"}
```

____________________________________________________________________________________________

Based on #786 , will rebase once merged.


service/rpc/availability.go
```diff
@@ -0,0 +1,44 @@
...
+		return
+	}
+
+	header, err := h.header.GetByHeight(r.Context(), uint64(height))
+	if err != nil {
+		writeError(w, http.StatusInternalServerError, heightAvailabilityEndpoint, err)
+		return
+	}
+
+	err = h.share.Availability.SharesAvailable(r.Context(), header.DAH)
```
liamsi, line 29: Could this spawn too many sample requests via rpc?

YazzyYaz:
Hi @renaynay is the response from this endpoint in JSON format or just a text response?

renaynay:
it should be json, lemme fix.

renaynay:
@YazzyYaz i've updated the description

renaynay:
Implement #806

service/rpc/availability.go
```diff
@@ -0,0 +1,66 @@
...
+	switch err {
+	case nil:
+		resp, err := json.Marshal(&availabilityResponse{
+			Height:    uint64(height),
```
adlerjohn, line 39: Why is height returned? Isn't it known as part of the request?

renaynay: we return it for namespaced_data and namespaced_shares, so I thought to add it here too. We can remove it though since the availability request always takes a height.

renaynay: done

service/rpc/availability.go
renaynay: resolved?

service/share/util.go
```diff
@@ -0,0 +1,12 @@
...
+import "math"
+
+// ProbabilityOfAvailability calculates the probability that the
+// data square is available based on the given amount of
+// samples collected.
+//
+// Formula: 1 - (0.75 ** amount of samples)
+func ProbabilityOfAvailability(sampleAmount int) float64 {
+	return 1 - math.Pow(0.75, float64(sampleAmount))
+}
```
Wondertan, line 12: I would honestly add the probability to be a return value of the availability interface, but that is not that important right now.

service/rpc/availability.go
```diff
@@ -0,0 +1,68 @@
...
+
+	header, err := h.header.GetByHeight(r.Context(), uint64(height))
+	if err != nil {
+		writeError(w, http.StatusInternalServerError, heightAvailabilityEndpoint, err)
+		return
+	}
+
+	availResp := &AvailabilityResponse{
+		Probability: strconv.FormatFloat(
+			share.ProbabilityOfAvailability(share.DefaultSampleAmount), 'g', -1, 64),
```
Wondertan, line 38: This is incorrect for the FullAvailability case. Meaning it always takes value from the number of samples being sampled which only applies to LightAvailability. It seems like the problem can be solved by extending the Availability interface explained in the comment above so that each impl makes the probability calculation inside. For the full availability, we would need to calculate the number of shares by doing EDSSize^2 and putting that into the formula.

Wondertan (changes requested):
Requesting changes because FullAvailability case responds with like it is a LightAvailability.

service/rpc/availability.go
renaynay: Don't we just return 1 for full availability given if the square can be reconstructed, the probability that it is available (to that full node) is 100%

Wondertan: 1 is fine

renaynay: Hm, we could do this w/ metrics though.

Wondertan: Metrics are a good addition to this, yes, but mainly the probability needs to be exposed to a user who wants to know it. Thinking further, we might need to add a runtime feature to change the number of samples a light node does sample, and then the probability n will be stored in the cache availability.

renaynay: Is there anything you'd like to add in this current PR or can we make an issue with all the ideas documented above? @Wondertan

Wondertan: I would make 2 issues out of what we have here:
* Report probability even for failed availability calls. E.g light client was able to find only N samples our of M or full node wasn't able to reconstruct, but was able to find N shares. From the gotten shares, we can get probability.
* Ability to modify the amount of sampled shares in the runtime for the Light Node.

Wondertan: https://github.com/celestiaorg/celestia-node/issues/876

Wondertan: and https://github.com/celestiaorg/celestia-node/issues/877

renaynay: Cool thanks, can i close this thread?

Wondertan: Yep

Wondertan: Resolving

service/share/interface.go
```diff
@@ -18,4 +18,7 @@ const AvailabilityTimeout = 20 * time.Minute
 type Availability interface {
 	// SharesAvailable subjectively validates if Shares committed to the given Root are available on the Network.
 	SharesAvailable(context.Context, *Root) error
+	// ProbabilityOfAvailability calculates the probability of the data square
+	// being available based on the number of samples collected.
+	ProbabilityOfAvailability() float64
```
Wondertan, line 23:
```suggestion
	// being available based on the number of samples collected.
	// TODO(@Wondertan): Merge with SharesAvailable method, eventually
	ProbabilityOfAvailability() float64
```

Wondertan (approved):

service/rpc/availability.go
adlerjohn: Poking @liamsi

adlerjohn (approved):

service/rpc/availability.go
renaynay: Closing as this is a non-issue.

PR #737, opened 2022-05-20
deps: bump rsmt2d to 0.5.0

Wondertan:
I tried updating to this version before but didn't like that we needed to keep both flattened and imported squares, so that code became even bulkier for the next currently unpushed iteration of Retriever, so I decided to wait until the requested rsmt2d API changes to land before upgrading(cc @adlerjohn), but now we have to update as core updated already.

PR #702, opened 2022-05-11
test(share): add Full Node reconstruction tests

Wondertan:
Mainly:
* Improves testnet utility in share pkg
* Introduces tests that assert that our reconstruction works in semi-ideal scenarios with values taken from the original paper
	* 1x One Full and multiple Lights
	* 2x Two Fulls and two subnets with multiple Lights that cannot reconstruct data on their own
	
___For an in-depth description of each test consult a complementary comment to each test.___

Substitutes https://github.com/celestiaorg/celestia-node/pull/598
Based on #682 
Closes https://github.com/celestiaorg/celestia-node/issues/600

My favourite part about these tests is that they won't require any changes if we e.g., move to a completely new protocol for Share transportation

Issue #709, opened 2022-05-13
Cleanup global variables


Wondertan:
In multiple places in the code, we rely on global variables, which is usually not that bad but has its downsides. Eventually, we should allocate some time to clean them up, which includes:
* "Deglobalization" of vars. It should be straightforward to configure those via direct optionable configuration of the structs/logic that those vars affect. Currently, it's not obvious to newcomers that those do exist and that they can be configurable.
* Pass those options/direct configurations into the Node configs, so that node operators can control them and fine-tune 
according to their setup
* Consistent naming

Bidon15:
Putting into Q3 Roadmap as we have a refactoring planning ahead

derrandz:
I think we can close this @Wondertan

PR #804, opened 2022-06-08
fix(ipld): fix reconstruction by re-enabling re-importing of rsmt2d.ExtendedDataSquare

### LINKED ISSUE #787: flaky TestSharesAvailable_Full
image

Wondertan:
An update of rsmt2d happened in https://github.com/celestiaorg/celestia-node/pull/737 during Mamaki launch preparation. This update changed our code not to reimport the square on each reconstruction attempt which silently sneaked in the problem, which is not reproducible with current tests. Reconstruction tests(#702) reproduce it, but they are not yet merged.

I was able to change the code to reproduce it reliably. Mainly I had to fetch only from columns and force the code to download only [2:4) quadrants. Unfortunately, it is hard to extract the reproducible case in a separate test, so we can either rely on reconstruction tests that are soon to be merged, or I can still try to make a test.

The reason for the bug is explained in #787, as well as a cleaner way to fix it. 
Closes #787

P.S. Was pretty hard to debug this case. I went on the wrong paths like 3 times. Though collected some minor findings that I am going to submit further.

### DIFF
--- a/ipld/retriever.go
+++ b/ipld/retriever.go
@@ -151,6 +151,19 @@ func (rs *retrievalSession) Reconstruct() (*rsmt2d.ExtendedDataSquare, error) {
 	// prevent further writes to the square
 	rs.squareLk.Lock()
 	defer rs.squareLk.Unlock()
+
+	// TODO(@Wondertan): This is bad!
+	//  * We should not reimport the square multiple times
+	//  * We should set shares into imported square via SetShare(https://github.com/celestiaorg/rsmt2d/issues/83)
+	//  to accomplish the above point.
+	{
+		squareImported, err := rsmt2d.ImportExtendedDataSquare(rs.square, rs.codec, rs.treeFn)
+		if err != nil {
+			return nil, err
+		}
+		rs.squareImported = squareImported
+	}
+
 	// and try to repair with what we have
 	err := rs.squareImported.Repair(rs.dah.RowsRoots, rs.dah.ColumnRoots, rs.codec, rs.treeFn)
 	if err != nil {
--- a/ipld/retriever_test.go
+++ b/ipld/retriever_test.go
@@ -94,3 +94,33 @@ func TestRetriever_ByzantineError(t *testing.T) {
 	var errByz *ErrByzantine
 	require.ErrorAs(t, err, &errByz)
 }
+
+// TestRetriever_MultipleRandQuadrants asserts that reconstruction succeeds
+// when any three random quadrants requested.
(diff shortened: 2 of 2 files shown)


liamsi (approved):
I would really, really prefer a testcase for this but I trust your hours of debugging this.

vgonkivs (approved):

Wondertan:
Force push with a cleaner commit

ipld/retriever.go
```diff
@@ -151,6 +151,19 @@ func (rs *retrievalSession) Reconstruct() (*rsmt2d.ExtendedDataSquare, error) {
 	// prevent further writes to the square
 	rs.squareLk.Lock()
 	defer rs.squareLk.Unlock()
+
+	// TODO(@Wondertan): This is bad!
+	//  * We should not reimport the square multiple times
+	//  * We should set shares into imported square via SetShare(https://github.com/celestiaorg/rsmt2d/issues/83)
```
adlerjohn, line 157: It's not clear to me how specifically this fixes a bug. This is indicative of a bug in the rsmt2d, not a missing feature.

Wondertan:
@liamsi, I added a test anyway. You can try removing the fix commit and reproduce it yourself. Sometimes it fails with other rsmt2d errors besides byzantine. The test ensures that at least 2 more quadrants are fetching besides the initial one, and all of those written into the shared square, and somewhere in between, the reconstruction fails.

Wondertan:
As always on CI something goes not like locally, aaaaa

Wondertan:
Force push with an updated test that should fail on CI, lets see...

Wondertan:
Ok, test is fixed

adlerjohn (approved):

ipld/retriever_test.go
```diff
@@ -94,3 +94,33 @@ func TestRetriever_ByzantineError(t *testing.T) {
 	var errByz *ErrByzantine
 	require.ErrorAs(t, err, &errByz)
 }
+
+// TestRetriever_MultipleRandQuadrants asserts that reconstruction succeeds
+// when any three random quadrants requested.
+func TestRetriever_MultipleRandQuadrants(t *testing.T) {
+	RetrieveQuadrantTimeout = time.Millisecond * 500
```
liamsi, line 101: If this global var is used in multiple tests, this assignment here will have side-effects.

Wondertan: Yeah, but I am controlling that. We have https://github.com/celestiaorg/celestia-node/issues/709 to fix this.

PR #808, opened 2022-06-09
fraud: introduce fraud service interface

vgonkivs:
* added Fraud service interface

### DIFF
--- a/docs/adr/adr-006-fraud-service.md
+++ b/docs/adr/adr-006-fraud-service.md
@@ -5,6 +5,11 @@
 - 2022.03.03 - init commit
 - 2022.03.08 - added pub-sub
 - 2022.03.15 - added BEFP verification
+- 2022.06.08 - 
+  * updated rsmtd error naming(as it was changed in implementation);
+  * changed from NamespaceShareWithProof to ShareWithProof;
+  * made ProofUnmarshaler public and extended return params;
+  * fixed typo issues;
 
 ## Authors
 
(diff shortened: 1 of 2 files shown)


liamsi (commented):
PR looks good. I was under the impression that we wanted to remove the interfaces (or at least some of them)? If not we can morge this.

docs/adr/adr-006-fraud-service.md
```diff
@@ -5,6 +5,11 @@
 - 2022.03.03 - init commit
 - 2022.03.08 - added pub-sub
 - 2022.03.15 - added BEFP verification
+- 2022.06.08 - 
+  * updated rsmtd error naming(as it was changed in implementation);
```
liamsi, line 9:
```suggestion
  * updated rsmt2d error naming(as it was changed in implementation);
```

Wondertan:
@liamsi, we decided to keep interfaces and split PR into smaller logical pieces

fraud/interface.go
```diff
@@ -0,0 +1,43 @@
+package fraud
+
+import (
+	"context"
+)
+
+// ProofUnmarshaler aliases a function that parses data to `Proof`.
+type ProofUnmarshaler func([]byte) (Proof, error)
```
liamsi, line 8: What is the rationale of not making this an interface as well then?

Wondertan: @liamsi, Why? if we just need a stateless function, not a type. Otherwise, there would be an empty struct like
```go
type unmarshaller struct {}
```

Feels like you are missing the point.

Wondertan (approved):
Thanks. Let's move on to implementation

vgonkivs:
Sir, yes sir

PR #809, opened 2022-06-09
chore: update nmt version

vgonkivs:


liamsi (commented):
Not sure we should update nmt and core to different versions from app.

go.mod
```diff
@@ -278,5 +278,5 @@ replace (
 	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.1.0-sdk-v0.46.0
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
 	github.com/libp2p/go-libp2p-pubsub v0.5.7-0.20211029175501-5c90105738cf => github.com/celestiaorg/go-libp2p-pubsub v0.5.6-celestia
-	github.com/tendermint/tendermint => github.com/celestiaorg/celestia-core v1.2.2-tm-v0.35.4
+	github.com/tendermint/tendermint => github.com/celestiaorg/celestia-core v1.2.3-tm-v0.35.6
```
liamsi, line 281: same comment as for nmt version bump

go.mod
```diff
@@ -8,7 +8,7 @@ require (
 	github.com/BurntSushi/toml v1.1.0
 	github.com/celestiaorg/celestia-app v0.5.2
 	github.com/celestiaorg/go-libp2p-messenger v0.1.0
-	github.com/celestiaorg/nmt v0.8.0
+	github.com/celestiaorg/nmt v0.10.0
```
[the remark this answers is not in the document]
liamsi: cc @evan-forbes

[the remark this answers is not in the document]
liamsi: That is also sth the node team should consider. cc @Wondertan @renaynay

As you already depend on app, you could use the re-exported types, which would make nmt and maybe rsmt2d indirect deps instead of direct deps. Dependabot wouldn't bother you about nmt, rsmt2d independently. Just sth to consider.

go.sum
```diff
@@ -1,25 +1,19 @@
-4d63.com/gochecknoglobals v0.1.0/go.mod h1:wfdC5ZjKSPr7CybKEcgJhUOgeAQW1+7WcyK8OvUilfo=
```
Wondertan, line 1: Did you update to the 1.18 locally? I have the same go.sum tidy changes, but I wasn't uploading them unless the whole team updates. And if the whole team updates, there is no sense in keeping our go.mod file in 1.17 and not 1.18. 

E.g, if someone git checks out with go 1.17 this updated go.sum, then he would need to do go tidy to be able to compile.

liamsi: do you even get these changes with 1.18 when doing `go mod tidy -compat=1.17`?

vgonkivs: yes. I actually executed `go mod tidy -compat=1.17`

vgonkivs: I will downgrade go and update pr then.

Wondertan:
I am taking over this PR as we discussed with @vgonkivs

go.sum
Wondertan: Independent update fixed that

go.mod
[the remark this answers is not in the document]
Wondertan: It might be a good idea for the app to re-export this. Not sure yet. Resolving this as it fine to depend on 0.10.0 fo now

liamsi (approved):

PR #810, opened 2022-06-09
[WIP] Telemetry poc

sysrex:


das/daser.go
```diff
@@ -138,8 +141,11 @@ func (d *DASer) sample(ctx context.Context, sub header.Subscription, checkpoint
 
 		startTime := time.Now()
 
-		err = d.da.SharesAvailable(ctx, h.DAH)
+		ctxWithSpan, span := otel.Tracer(tracerName).Start(ctx, "sample")
+
+		err = d.da.SharesAvailable(ctxWithSpan, h.DAH)
 		if err != nil {
+			span.End()
```
Wondertan, line 148: nit: do defer `span.End()`. This would give the same result, and the span will end on any success or error return.

das/daser.go
```diff
@@ -138,8 +141,11 @@ func (d *DASer) sample(ctx context.Context, sub header.Subscription, checkpoint
 
 		startTime := time.Now()
 
-		err = d.da.SharesAvailable(ctx, h.DAH)
+		ctxWithSpan, span := otel.Tracer(tracerName).Start(ctx, "sample")
```
Wondertan, line 144: AFAIK, the `otel.Tracer` should be defined once globally and then accessed multiple times from the code.

Wondertan: The guide also mentions this https://opentelemetry.uptrace.dev/guide/go-tracing.html#quickstart

das/daser.go
Wondertan: We should record errors so that the operation does not look like a success. https://opentelemetry.uptrace.dev/guide/go-tracing.html#recording-errors

das/daser.go
```diff
@@ -222,8 +229,11 @@ func (d *DASer) catchUp(ctx context.Context, job *catchUpJob) (int64, error) {
 
 		startTime := time.Now()
 
-		err = d.da.SharesAvailable(ctx, h.DAH)
+		ctxWithSpan, span := otel.Tracer(tracerName).Start(ctx, "catchUp")
```
Wondertan, line 232: Same here. We create multiple tracers with the same name, which conflict with each other. Instead, there should be one global.

Wondertan (commented):
I really appreciate how quickly you were able to implement this and understood some of the patterns and libs we are using her. A few things that need to be fixed before proceeding:
* Rebase on main. The PR has many LoCs removed from go.mod. This is due to go1.18 changes, and you are likely using it. We weren't updated to 1.18 when you started the POC, but now were are, so if you rebase, all the orthogonal to the PR go.mod changes will disappear.
* Keep only metrics(histogram) for ShareAvailable. Traces are good and valuable, but they are out of the scope of the task. I commented below on some problems with current integration), but let's not fix them now and only keep metrics for now in this PR. Also, the histogram is currently integrated for FullAvailability only, and we need it for the Light as well.

I will look at the `telemetry` pkg deeper in the second review.

Wondertan:
We can close this as #907 was merged

PR #811, opened 2022-06-09
fraud: implement fraud service

vgonkivs:
Implemented fraud service on top of fraud interface.


fraud/service.go
```diff
@@ -0,0 +1,147 @@
...
+			pubSubTopics: make(map[ProofType]*pubsub.Topic),
+		},
+		unmarshalers: make(map[ProofType]ProofUnmarshaler),
+		getter:       getter,
+	}
+}
+
+func (f *service) Subscribe(proofType ProofType) (Subscription, error) {
+	// TODO: @vgonkivs check if fraud proof is in fraud store, then return with error
+	f.mu.Lock()
```
renaynay, line 36: pls defer f.mu.Unlock here bc if !ok and it returns, map is still locked.

renaynay: Also if you unlock at the end of the func, might as well be a defer instead of 3 diff unlock statements

fraud/service.go
```diff
@@ -0,0 +1,147 @@
...
+	u, ok := f.unmarshalers[proofType]
```
renaynay, line 37: Also I think that serialisation, validator registration/(deregistration?), topic should have a tighter coupling than this as you cannot have a topic w/o serialisation and should not have topic without validator. Then such checks can be avoided.

Something like: 

```
// topic constructor should also take validator so that there is no need for a separate register validator step, this should happen in the construction of Topic
type Topic struct {
 topic *topic
 unmarshaler ProofUnmarshaler
}
```

```
	pubSubTopics map[ProofType]*Topic
```

renaynay: bc `topics` inside topics.go is essentially a topics manager, why can't it also manage the topic's serialisation?

renaynay: This will simplify a lot of the code as well IMO.

renaynay (changes requested):

fraud/service.go
```diff
@@ -0,0 +1,130 @@
...
+func (f *service) UnregisterUnmarshaler(proofType ProofType) error {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	t, ok := f.topics[proofType]
+	if !ok {
+		return errors.New("fraud: topic was not created")
+	}
+	delete(f.topics, proofType)
+	return t.close()
+
```
renaynay, line 63:
```suggestion
```

fraud/topic.go
```diff
@@ -0,0 +1,57 @@
...
+	"context"
+
+	"github.com/libp2p/go-libp2p-core/peer"
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
+)
+
+// topic holds pubsub topic and unmarshaler of corresponded Fraud Proof
+type topic struct {
+	topic *pubsub.Topic
+	codec ProofUnmarshaler
```
renaynay, line 13: I would just call this `unmarshal`

renaynay (approved):
So good 🚀

fraud/service.go
```diff
@@ -0,0 +1,130 @@
...
+	t, ok := f.topics[proofType]
+	f.mu.RUnlock()
+	if !ok {
+		log.Error("topic was not created")
+		return pubsub.ValidationReject
+	}
+	proof, err := t.unmarshal(msg.Data)
+	if err != nil {
+		log.Errorw("unmarshaling header error", err)
+		return pubsub.ValidationReject
```
Wondertan, line 95: We should also blacklist a peer here.

fraud/service.go
```diff
@@ -0,0 +1,130 @@
...
+			log.Fatal("could not get block. Timeout reached. Please restart your node.")
+		}
+		log.Errorw("failed to fetch block: ",
+			"err", err)
+		return pubsub.ValidationReject
+	}
+	defer cancel()
+	err = proof.Validate(extHeader)
+	if err != nil {
+		log.Errorw("validation err: ",
```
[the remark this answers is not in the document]
Wondertan: Also, would be nice to log other detail of FP here. Like height and type

fraud/service.go
```diff
@@ -0,0 +1,130 @@
...
+		// Timeout means there is a problem with the network.
+		// As we cannot prove or discard Fraud Proof, user must restart the node.
+		if errors.Is(err, context.DeadlineExceeded) {
+			log.Fatal("could not get block. Timeout reached. Please restart your node.")
+		}
+		log.Errorw("failed to fetch block: ",
+			"err", err)
+		return pubsub.ValidationReject
+	}
+	defer cancel()
```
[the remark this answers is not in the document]
vgonkivs: Yes, correct. But inside `err != nil ` I'm calling `log.Fatal` to stop the node timeout is reached. In this case defer cancel() won't be invoked. That's the reason of adding an extra cancel() call.
But now I think `log.Error` will be enough

fraud/service.go
```diff
@@ -0,0 +1,130 @@
...
+	ctx context.Context,
+	proofType ProofType,
+	msg *pubsub.Message,
+) pubsub.ValidationResult {
+	f.mu.RLock()
+	t, ok := f.topics[proofType]
+	f.mu.RUnlock()
+	if !ok {
+		log.Error("topic was not created")
+		return pubsub.ValidationReject
```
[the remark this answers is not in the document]
vgonkivs: We still have Unregister. So it's possible. panic will be good. thanks.

fraud/service.go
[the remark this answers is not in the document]
Wondertan: Ok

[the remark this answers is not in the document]
Wondertan: If we make Error here instead of Fatal, we might need to consider how to handle the error here.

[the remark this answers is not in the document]
Wondertan: As the proof might be valid, while we could just get a header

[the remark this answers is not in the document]
vgonkivs: Will create an issue for this and post it here

fraud/service.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+	proof, err := t.unmarshal(msg.Data)
+	if err != nil {
+		log.Errorw("unmarshaling header error", err)
+		return pubsub.ValidationReject
+	}
+
+	// create a timeout for block fetching since our validator will be called synchronously
+	// and getter is a blocking function.
+	newCtx, cancel := context.WithTimeout(ctx, time.Minute*2)
```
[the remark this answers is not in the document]
vgonkivs: done

fraud/service.go
```diff
@@ -0,0 +1,130 @@
...
+	if err != nil {
+		cancel()
+		// Timeout means there is a problem with the network.
+		// As we cannot prove or discard Fraud Proof, user must restart the node.
+		if errors.Is(err, context.DeadlineExceeded) {
+			log.Fatal("could not get block. Timeout reached. Please restart your node.")
+		}
+		log.Errorw("failed to fetch block: ",
+			"err", err)
+		return pubsub.ValidationReject
```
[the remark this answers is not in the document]
vgonkivs: done

fraud/service.go
vgonkivs: done

fraud/service.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+	defer cancel()
+	err = proof.Validate(extHeader)
+	if err != nil {
+		log.Errorw("validation err: ",
+			"err", err)
+		f.pubsub.BlacklistPeer(msg.ReceivedFrom)
+		return pubsub.ValidationReject
+	}
+	log.Warnw("received an inbound proof", "kind", proof.Type(),
```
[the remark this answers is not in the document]
vgonkivs: removed

fraud/service.go
```diff
@@ -0,0 +1,147 @@
...
+		return pubsub.ValidationIgnore
+	}
+	err = proof.Validate(extHeader)
+	if err != nil {
+		log.Errorw("proof validation err: ",
+			"err", err, "proofType", proof.Type(), "height", proof.Height())
+		f.pubsub.BlacklistPeer(msg.ReceivedFrom)
+		return pubsub.ValidationReject
+	}
+	log.Warnw("received fraud proof", "proofType", proof.Type(),
```
Wondertan, line 138: Why no height here?

fraud/service.go
```diff
@@ -0,0 +1,147 @@
...
+	f.topicsLk.Unlock()
+	if !ok {
+		return nil, errors.New("fraud: unmarshaler for the given proof type is not registered")
+	}
+	return newSubscription(t)
+}
+
+func (f *service) RegisterUnmarshaler(proofType ProofType, u ProofUnmarshaler) error {
+	f.topicsLk.Lock()
+	defer f.topicsLk.Unlock()
```
Wondertan, line 49: nit: lock specifically the write to the map

fraud/service.go
```diff
@@ -0,0 +1,147 @@
...
+			return f.processIncoming(ctx, proofType, msg)
+		},
+		// make validation synchronous.
+		pubsub.WithValidatorInline(true),
+	)
+	if err != nil {
+		return err
+	}
+
+	f.topics[proofType] = &topic{topic: t, unmarshal: u}
```
Wondertan, line 67: Should we check if the unmarshaller is already registered?

Wondertan (commented):
Nice. We should add a few simple units here and we are fully read to go

fraud/service.go
```diff
@@ -0,0 +1,147 @@
...
+	"errors"
+	"sync"
+	"time"
+
+	"github.com/libp2p/go-libp2p-core/peer"
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
+)
+
+const (
+	fetchHeaderTimeout = time.Minute * 2
```
liamsi, line 15: I know this is private but I still think this deserves a comment. Ideally, what the rationale behind the 2 mins is as well.

fraud/service.go
Wondertan: I mean other logs shows the height of the proof, while this one is not

fraud/service.go
Wondertan: Or allow overwriting

Wondertan: We don't call it on init and its runtime or build time for the Node. We can also log this out. Fine with this, as long as it gives something to debug if someone additionally register it twice

fraud/service.go
vgonkivs: comment added.

liamsi (approved):

Wondertan (approved):

fraud/service_test.go
```diff
@@ -0,0 +1,70 @@
...
+	_, err := s.Subscribe(BadEncoding)
+	require.Error(t, err)
+}
+
+func TestService_Subscribe(t *testing.T) {
+	s := createService(t)
+	require.NoError(t, s.RegisterUnmarshaler(BadEncoding, UnmarshalBEFP))
+
+	_, err := s.Subscribe(BadEncoding)
+	require.NoError(t, err)
```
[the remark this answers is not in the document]
vgonkivs: Thought about that. But I will add Swamp test which will do the same in my next PR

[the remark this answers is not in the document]
vgonkivs: Thought about that. But I will add Swamp test which will do the same in my next PR

fraud/subscription.go
```diff
@@ -0,0 +1,38 @@
...
+func (s *subscription) Proof(ctx context.Context) (Proof, error) {
+	data, err := s.subscription.Next(ctx)
+	if err != nil {
+		return nil, err
+	}
+	proof, err := s.unmarshaler(data.Data)
+	if err != nil {
+		return nil, err
+	}
+	return proof, nil
```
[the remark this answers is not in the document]
vgonkivs: yes. Thanks

vgonkivs:
Снимок экрана 2022-06-13 в 17 46 03

Issue #814, opened 2022-06-11
ipld: race in Retriever


Wondertan:
The race is found in #702 branch, but locally.
```
=== RUN   TestShareAvailable_OneFullNode
2022-06-11T00:07:38.964+0200    ERROR   engine  decision/engine.go:691  wantlist index doesn't match peer's wantlist    {"peer": "1WRsHVXq899cA7"}
2022-06-11T00:07:38.989+0200    ERROR   engine  decision/engine.go:691  wantlist index doesn't match peer's wantlist    {"peer": "1WRsHVXbUUKN1C"}
2022-06-11T00:07:39.207+0200    ERROR   engine  decision/engine.go:691  wantlist index doesn't match peer's wantlist    {"peer": "1WRsHVXbUUKN1C"}
2022-06-11T00:07:39.503+0200    ERROR   engine  decision/engine.go:691  wantlist index doesn't match peer's wantlist    {"peer": "1WRsHVXecAgMfs"}
==================
WARNING: DATA RACE
Read at 0x00c0005ecc30 by goroutine 82302:
  github.com/celestiaorg/celestia-node/ipld.(*retrievalSession).doRequest.func1.1()
      /home/Wondertan/src/go/celestia-node/ipld/retriever.go:241 +0xe4
  github.com/celestiaorg/celestia-node/ipld.GetShares.func1()
      /home/Wondertan/src/go/celestia-node/ipld/get_shares.go:95 +0x242
  github.com/gammazero/workerpool.startWorker()
      /home/Wondertan/go/pkg/mod/github.com/gammazero/workerpool@v1.1.2/workerpool.go:233 +0x35
  github.com/gammazero/workerpool.(*WorkerPool).dispatch.func2()
      /home/Wondertan/go/pkg/mod/github.com/gammazero/workerpool@v1.1.2/workerpool.go:195 +0x47

Previous write at 0x00c0005ecc30 by goroutine 29115:
  github.com/celestiaorg/celestia-node/ipld.(*retrievalSession).doRequest.func1.1()
      /home/Wondertan/src/go/celestia-node/ipld/retriever.go:242 +0x144
  github.com/celestiaorg/celestia-node/ipld.GetShares.func1()
      /home/Wondertan/src/go/celestia-node/ipld/get_shares.go:95 +0x242
  github.com/gammazero/workerpool.worker()
[...]

PR #815, opened 2022-06-12
fix(ipld): fix races in the Retriever's doRequest method

Wondertan:
Mainly this PR fixes two races:
1. One race happens when in one slice slot, two routines are trying to write the same share. Two because we have two sources for each share: Row and Col. Also, this does not happen on a happy path, and only in case, Retriever tries to fetch as many shares as possible to reconstruct the block, explaining why the race was uncovered with #702.

The mentioned race is a known one, and we had an incorrect attempt previously to fix it https://github.com/celestiaorg/celestia-node/commit/cca832acb6dbb03ac734fa5090d50bc319b9d0ba. 

This fix is correct and achieved by keeping a mutex per share. If a share comes from any of two sources first, it locks the slot and prevents a share from another source from being written.

2. The second race happened after the square was reconstructed. During the reconstruction, some new shares could arrive and wait on the lock. Once reconstructed, they were unlocked and continued to write in an already filled-up square slice causing multiple races with the same root. Basically, the reconstructed square was returned to the user, while Retriever internal was continue to writing. This PR fixes that and ensure that after unlocking writings routines check if reconstruction happened and if so aborts them.

Closes https://github.com/celestiaorg/celestia-node/issues/814

P.S. The complexity of the code in Retriever is starting to grow, and it's hard to test it. We will likely need to refactor the code(I have a brief picture in my head). However, though, this should happen after all the API improvements land in rsmt2d.

### DIFF
--- a/ipld/retriever.go
+++ b/ipld/retriever.go
@@ -104,9 +104,10 @@ type retrievalSession struct {
 	sharesLks   []sync.Mutex
 	sharesCount uint32
 
-	squareLk sync.RWMutex
-	square   [][]byte
-	squareDn chan struct{}
+	squareLk  sync.RWMutex
+	square    [][]byte
+	squareSig chan struct{}
+	squareDn  chan struct{}
 }
 
 // newSession creates a new retrieval session and kicks off requesting process.
@@ -129,7 +130,8 @@ func (r *Retriever) newSession(ctx context.Context, dah *da.DataAvailabilityHead
 		quadrants: newQuadrants(dah),
 		sharesLks: make([]sync.Mutex, size*size),
 		square:    make([][]byte, size*size),
-		squareDn:  make(chan struct{}, 1),
+		squareSig: make(chan struct{}, 1),
+		squareDn:  make(chan struct{}),
 	}
 
 	square, err := rsmt2d.ImportExtendedDataSquare(ses.square, ses.codec, ses.treeFn)
@@ -146,11 +148,14 @@ func (r *Retriever) newSession(ctx context.Context, dah *da.DataAvailabilityHead
 // square reconstruction. "Attempt" because there is no way currently to
 // guarantee that reconstruction can be performed with the shares provided.
 func (rs *retrievalSession) Done() <-chan struct{} {
-	return rs.squareDn
+	return rs.squareSig
 }
 
 // Reconstruct tries to reconstruct the data square and returns it on success.
 func (rs *retrievalSession) Reconstruct() (*rsmt2d.ExtendedDataSquare, error) {
+	if rs.isReconstructed() {
+		return rs.squareImported, nil
+	}
 	// prevent further writes to the square
 	rs.squareLk.Lock()
 	defer rs.squareLk.Unlock()
@@ -173,7 +178,20 @@ func (rs *retrievalSession) Reconstruct() (*rsmt2d.ExtendedDataSquare, error) {
 		return nil, err
 	}
 	log.Infow("data square reconstructed", "data_hash", hex.EncodeToString(rs.dah.Hash()), "size", len(rs.dah.RowsRoots))
+	close(rs.squareDn)
 	return rs.squareImported, nil
+}
+
+// isReconstructed report true whether the square attached to the session
+// is already reconstructed.
+func (rs *retrievalSession) isReconstructed() bool {
+	select {
+	case <-rs.squareDn:
+		// return early if square is already reconstructed
+		return true
+	default:
+		return false
+	}
 }
 
 func (rs *retrievalSession) Close() error {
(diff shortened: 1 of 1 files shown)


Wondertan:
Oh shoot, I just realized I am using a feature from 1.18 here, ahhh.

Wondertan:
Ok, so this race fixing PR uncovered another race on CI. Nice

Wondertan:
Ok, so now this PR fixes two independent races. Will update the description

ipld/retriever.go
```diff
@@ -229,30 +250,55 @@ func (rs *retrievalSession) doRequest(ctx context.Context, q *quadrant) {
...
 				}
+				rs.square[idx] = share
 				rs.squareLk.RUnlock()
+				// if we have >= 1/4 of the square we can start trying to Reconstruct
+				// TODO(@Wondertan): This is not an ideal way to know when to start
+				//  reconstruction and can cause idle reconstruction tries in some cases,
+				//  but it is totally fine for the happy case and for now.
+				//  The earlier we correctly know that we have the full square - the earlier
+				//  we cancel ongoing requests - the less data is being wastedly transferred.
+				if atomic.AddUint32(&rs.sharesCount, 1) >= uint32(size*size) {
```
vgonkivs, line 287: Shouldn't it be `math.Sqrt(size*size)` ?

vgonkivs, line 287: Shouldn't it be `math.Sqrt(size*size)`. Once we reach 1/4 we need to start reconstruction.

Wondertan:
Rebased on main with go1.18. Removed temporal semaphore as we have the needed feature from 1.18 now

ipld/retriever.go
vgonkivs: Yes, sorry. Did not notice that it's related to session.

ipld/retriever.go
```diff
@@ -101,10 +101,13 @@ type retrievalSession struct {
...
-	squareLk    sync.RWMutex
-	square      [][]byte
+	sharesLks   []sync.Mutex
 	sharesCount uint32
-	done        chan struct{}
+
+	squareLk  sync.RWMutex
+	square    [][]byte
+	squareSig chan struct{}
+	squareDn  chan struct{}
```
liamsi, line 110: DN?!

liamsi (commented):
> P.S. The complexity of the code in Retriever is starting to grow, and it's hard to test it. We will likely need to refactor the code(I have a brief picture in my head). However, though, this should happen after all the API improvements land in rsmt2d.

Can you please write that down (also which changes in r2mt2d are necessary and how they related to simplifying the retriever code here).

This is really hard to reason about and I can't really tell if this PR fixes things while introducing a bunch of others.

ipld/retriever.go
Wondertan: Gotem!

ipld/retriever.go
```diff
@@ -229,29 +250,45 @@ func (rs *retrievalSession) doRequest(ctx context.Context, q *quadrant) {
...
 			// which we put into the rs.square share-by-share by calculating shares' indexes using q.index
 			GetShares(ctx, rs.bget, nd.Links()[q.x].Cid, size, func(j int, share Share) {
-				// the R lock here is *not* to protect rs.square from multiple concurrent shares writes
-				// but to avoid races between share writes and repairing attempts
-				// shares are written atomically in their own slice slot
+				// NOTE: Each share can be fetched twice, from two sources(Row or Col).
+				// These shares are always equal, and we allow only the first one to be written
+				// in the square.
+				// NOTE-2: We never actually fetch shares from the network *twice*,
+				// and they are cached on IPLD(blockservice) level.
```
liamsi, line 257: Doesn't Note-2 contradict the first note?
> Each share can be fetched twice, from two sources(Row or Col).
> We never actually fetch shares from the network *twice*

this is confusing.

Wondertan:
@liamsi, to reason about the code is better to check out and see the whole code, rather than changes only on GH.

Wondertan:
I also put a lot of effort into comments. If they don't make sense, pls point that out. Otherwise, I am pretty confident in this code now.

ipld/retriever.go
Wondertan: There is a chance that the same share could appear in this func twice, without being fetched twice from the network. Even though we traverse over the same share twice, it's never fetched over the network twice because of the caching on BlockService layer. I really don't know ho to explain this better.

liamsi: I think I understand. And I think you explained it well enough. It just sounds contradictory at first sight but it is actually quite clear.

ipld/retriever.go
```diff
@@ -229,29 +250,45 @@ func (rs *retrievalSession) doRequest(ctx context.Context, q *quadrant) {
 			// the left or the right side of the tree represent some portion of the quadrant
 			// which we put into the rs.square share-by-share by calculating shares' indexes using q.index
 			GetShares(ctx, rs.bget, nd.Links()[q.x].Cid, size, func(j int, share Share) {
-				// the R lock here is *not* to protect rs.square from multiple concurrent shares writes
-				// but to avoid races between share writes and repairing attempts
-				// shares are written atomically in their own slice slot
+				// NOTE: Each share can be fetched twice, from two sources(Row or Col).
+				// These shares are always equal, and we allow only the first one to be written
+				// in the square.
```
liamsi, line 255:
```suggestion
				// NOTE: Each share can appear twice here, for a Row and Col, respectively..
				// These shares are always equal, and we allow only the first one to be written
				// in the square.
```

ipld/retriever.go
liamsi, line 257:
```suggestion
				// NOTE-2: We never actually fetch shares from the network *twice*.
				// Once a share is downloaded from the network it is cached on the 
				// IPLD(blockservice) level.
```

Wondertan:
Thanks you @liamsi!

vgonkivs (approved):

adlerjohn (approved):

PR #828, opened 2022-06-15
fraud: extend proof stucture

### LINKED ISSUE #827: fraud: extend proof structure with block hash
Add block hash as an additional  field in proof struct, so we can store fraud proofs by block hash as a key in fraud store.

vgonkivs:
Resolves #827


docs/adr/adr-006-fraud-service.md
```diff
@@ -117,6 +118,7 @@ type ProofType int
 // Proof is a generic interface that will be used for all types of fraud proofs in the network.
 type Proof interface {
    Type() ProofType
+   Hash() string
```
liamsi, line 121: In the interface it might be a bit unclear what this hash refers to. Is it the Hash() (summary) of the Proof? That is what you'd usually expect from a Hash() method.

fraud/proof.go
```diff
@@ -26,6 +26,8 @@ func (p ProofType) String() string {
 type Proof interface {
 	// Type returns the exact type of fraud proof
 	Type() ProofType
+	// Hash returns the block hash.
+	Hash() string
```
liamsi, line 30: I'm wondering if we should call this what it is: `BlockHash()`? Or maybe `BlockID()` even simply `ID()`? As it identifies the Proof to map it uniquely to a block and hashes are usually of type `[]byte`. 🤔

fraud/bad_encoding.go
```diff
@@ -50,6 +53,11 @@ func (p *BadEncodingProof) Type() ProofType {
 	return BadEncoding
 }
 
+// Hash returns block height.
```
adlerjohn, line 56: ?
```suggestion
// Hash returns block hash.
```

fraud/proof.go
adlerjohn: +1 on just calling it `BlockHash()`.

fraud/bad_encoding.go
vgonkivs: nice catch 😄

liamsi (approved):

fraud/proof.go
```diff
@@ -26,6 +26,8 @@ func (p ProofType) String() string {
 type Proof interface {
 	// Type returns the exact type of fraud proof
 	Type() ProofType
+	// BlockHash returns the block hash.
+	BlockHash() []byte
```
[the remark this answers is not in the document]
vgonkivs: Block hash will be used to store proof in fraud store and now we do not need block height anymore as we can fetch the block by its hash to get roots and verify Fraud Proof. Will create a task for it.

vgonkivs:
In a private discussion with @Wondertan we agreed that better to name it `HeaderHash`. It will look more natural and will not confuse users. I agree with him. We are not working with blocks but with headers.

adlerjohn (approved):

PR #832, opened 2022-06-16
Update Docker image tag creation

sysrex:
Update Docker image tag creation
- Generate sha tag for main branch
- generate branch name tag for every other branch

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,0 +1,38 @@
+#!/bin/sh
+
+set -eu -o pipefail
+
+STAGED_GO_FILES=$(git diff --cached --name-only -- '*.go')
+STAGED_MD_FILES=$(git diff --cached --name-only -- '*.md')
+
+if [[ $STAGED_GO_FILES == "" ]] && [[ $STAGED_MD_FILES == "" ]]; then
+  echo "--> Found no go or markdown files, skipping linting"
+elif [[ $STAGED_GO_FILES == "" ]]; then
+  echo "--> Found markdown files, linting"
+  if ! command -v markdownlint &> /dev/null ; then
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-build.yml
```diff
@@ -29,7 +29,8 @@ jobs:
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
           tags: |
-            type=sha
+            type=sha,format=long,enable=${{ github.ref == format('refs/heads/{0}', 'main') }}
+            type=ref,enable=${{ github.ref != format('refs/heads/{0}', 'main') }},event=branch
```
liamsi, line 33: How will the resulting tag look like?

sysrex: if the branch is main the docker tag will be sha-xxxxxxx, for any other branch the docker tag is the branch name (that way we can create a matrix of testing regardless of where the docker image is coming from and we're also able to differentiate between them)

sysrex: yes, that is correct

Wondertan: I am fine with this, but don't understand the usecase behind this, can you elaborate?

Wondertan:
@sysrex, how should we proceed here?

Wondertan:
Do we still need this?

Bidon15:
bumping this @sysrex

SweeXordious (commented):
Awesome stuff. I guess it makes sense also to do the same for celestia-app

sysrex:
We still require this to have the docker image tag match the commit sha, this is imperative to be able to continue unobstructed working on the devnet.

PR #835, opened 2022-06-18
p2p: add black list option

### LINKED ISSUE #819: node/p2p: blacklist set of peers
In order to complete our reconstruction tests effort (described here: https://github.com/celestiaorg/test-infra/issues/21), we need to provide an option for the user to blacklist the set of peers.

In addition, this can be a handy way for future cases outside the scope of testground

vgonkivs:
Resolves #819

### DIFF
--- a/node/p2p/misc.go
+++ b/node/p2p/misc.go
@@ -50,8 +50,8 @@ func ConnectionManager(cfg Config) func(params.Bootstrappers) (coreconnmgr.ConnM
 }
 
 // ConnectionGater constructs a ConnectionGater.
-func ConnectionGater(cfg Config) func(datastore.Batching) (coreconnmgr.ConnectionGater, error) {
-	return func(ds datastore.Batching) (coreconnmgr.ConnectionGater, error) {
+func ConnectionGater(cfg Config) func(datastore.Batching) (*conngater.BasicConnectionGater, error) {
+	return func(ds datastore.Batching) (*conngater.BasicConnectionGater, error) {
 		m, err := conngater.NewBasicConnectionGater(ds)
 		if err != nil {
 			return nil, err
--- a/node/p2p/misc.go
+++ b/node/p2p/misc.go
@@ -50,8 +50,17 @@ func ConnectionManager(cfg Config) func(params.Bootstrappers) (coreconnmgr.ConnM
 }
 
 // ConnectionGater constructs a ConnectionGater.
-func ConnectionGater(ds datastore.Batching) (coreconnmgr.ConnectionGater, error) {
-	return conngater.NewBasicConnectionGater(ds)
+func ConnectionGater(cfg Config) func(datastore.Batching) (coreconnmgr.ConnectionGater, error) {
+	return func(ds datastore.Batching) (coreconnmgr.ConnectionGater, error) {
+		m, err := conngater.NewBasicConnectionGater(ds)
+		if err != nil {
+			return nil, err
+		}
+		for _, peer := range cfg.BlacklistPeers {
+			_ = m.BlockPeer(peer)
+		}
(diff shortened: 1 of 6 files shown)


node/tests/p2p_test.go
```diff
@@ -37,3 +37,28 @@ func TestUseBridgeNodeAsBootstraper(t *testing.T) {
...
+
+/*
+Test-Case: Add peer to blacklist
+Steps:
+1. Create a Full Node(BN)
+2. Start a FN
+3. Create a Light Node(LN) and add id of FN to black list
+5. Start a LN
+6. Check FN is allowed to connect to LN
+7. Check LN is not allowed to connect to FN
```
Bidon15, line 49: Can we elaborate better what does these "allowed/not allowed connect" means? I would rather use terminology that you defined for BlackListPeers `dial` instead of connect. 

In addition, what will happen to the light node, if it has added the full node as trusted peer initially and then opted to blacklist it later? 🧐

Bidon15 (changes requested):
Some nits. Otherwise, lgtm

node/p2p/misc.go
```diff
@@ -50,8 +50,17 @@ func ConnectionManager(cfg Config) func(params.Bootstrappers) (coreconnmgr.ConnM
 }
 
 // ConnectionGater constructs a ConnectionGater.
-func ConnectionGater(ds datastore.Batching) (coreconnmgr.ConnectionGater, error) {
-	return conngater.NewBasicConnectionGater(ds)
+func ConnectionGater(cfg Config) func(datastore.Batching) (*conngater.BasicConnectionGater, error) {
+	return func(ds datastore.Batching) (*conngater.BasicConnectionGater, error) {
+		m, err := conngater.NewBasicConnectionGater(ds)
```
renaynay, line 55: random question, but why `m` ?

renaynay (approved):

node/p2p/misc.go
```diff
@@ -50,8 +50,17 @@ func ConnectionManager(cfg Config) func(params.Bootstrappers) (coreconnmgr.ConnM
...
+		if err != nil {
+			return nil, err
+		}
+		for _, peer := range cfg.BlacklistPeers {
+			_ = m.BlockPeer(peer)
```
Wondertan, line 60: The ConnGater takes a Datastore and all the blacklisted peers are getting persisted inside it. This means that each time you run a node, the peers from the config are getting stored, again and again.

Wondertan: So if we store them in Datastore - there is no need to store them in the Config. 
The persisted blocklist can be accessed in the runtime, via accessing the ConnGater directly. No need for it to be a node build time option.

Wondertan (changes requested):
The blacklisting use case should be solved in the _runtime_ of the node, not a _build_ time. So instead of making the option(build time), the user should access the running node via API to tell it what peers should be blacklisted. In our case, where we don't have the API yet, the user(Testground as per issue) can access the ConnectionGater field directly from the Node and add a peer to the blacklist.

node/p2p/misc.go
Wondertan: So basically, this PR should only ensure the ConnGater is accessible from the node and that's it

vgonkivs: So, your suggestion is to only keep changes with moving from interface type to direct implementation type, right?

Wondertan: Right

vgonkivs: Then I suggest to close this Pr in favour of another with suggested changes.

vgonkivs:
closing this PR in favour of https://github.com/celestiaorg/celestia-node/pull/863

PR #842, opened 2022-06-20
header: `Exchange` interface implements `Getter`

renaynay:
This PR changes `Exchange` interface to just implement `Getter` as they are performing the same functions just with different names.

### DIFF
--- a/header/interface.go
+++ b/header/interface.go
@@ -55,19 +55,7 @@ type Broadcaster interface {
 // Exchange encompasses the behavior necessary to request ExtendedHeaders
 // from the network.
 type Exchange interface {
-	// RequestHead requests the latest ExtendedHeader. Note that the ExtendedHeader
-	// must be verified thereafter.
-	RequestHead(ctx context.Context) (*ExtendedHeader, error)
-	// RequestHeader performs a request for the ExtendedHeader at the given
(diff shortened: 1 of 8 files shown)


header/interface.go
```diff
@@ -55,19 +55,7 @@ type Broadcaster interface {
...
-	// RequestHeader performs a request for the ExtendedHeader at the given
-	// height to the network. Note that the ExtendedHeader must be verified
-	// thereafter.
-	RequestHeader(ctx context.Context, height uint64) (*ExtendedHeader, error)
-	// RequestHeaders performs a request for the given range of ExtendedHeaders
-	// to the network. Note that the ExtendedHeaders must be verified thereafter.
-	RequestHeaders(ctx context.Context, origin, amount uint64) ([]*ExtendedHeader, error)
-	// RequestByHash performs a request for the ExtendedHeader by the given hash corresponding
-	// to the RawHeader. Note that the ExtendedHeader must be verified thereafter.
-	RequestByHash(ctx context.Context, hash tmbytes.HexBytes) (*ExtendedHeader, error)
```
Wondertan, line 70: Can you pls move the comments from the interface to the implementations of the P2P exchange where relevant?

Wondertan (approved):
❤️ this change, even though it has a low priority.

PR #845, opened 2022-06-20
Remove hardcoding of init path based on node type

jbowen93:
Fix dockerfile so that it doesn't hardcode the init path.

The init path is still variable but now dependent on the logic in the `celestia-node` code itself and not the `entrypoint.sh`.


docker/Dockerfile
```diff
@@ -1,6 +1,5 @@
 FROM --platform=$BUILDPLATFORM golang:1.18 as builder
-RUN apt-get update && \
-    apt-get install make bash
```
Wondertan, line 3: Why do we remove 'update' and bash installation?

Wondertan (approved):
LGTM

docker/Dockerfile
jbowen93: Because this is the builder image and these aren't needed to perform the build successfully. Just hygiene.

adlerjohn (approved):
utACK

renaynay (approved):

PR #848, opened 2022-06-22
das: Implement `DASState`

### LINKED ISSUE #427: das: Implement DASer public API

renaynay:
Resolves #427 and https://github.com/celestiaorg/celestia-node/issues/869

TODO: 

- [x] Implement square width tracking of last sampled height
- [x] IsRunning method

---------------------------------------------------------------------------------------------------------------------------

### GET `/daser`

Request

```
curl -X GET http://<ip>:26658/daser/state
```

Response

```
{"sample_routine":{"error":null,"is_running":true,"latest_sampled_height":104143,"latest_sampled_square_width":16},"catch_up_routine":{"start":"2022-07-07T15:01:25.747118+02:00","end":"0001-01-01T00:00:00Z","error":null,"id":1,"height":96247,"from":96162,"to":104142}}
```

### DIFF
--- a/das/daser.go
+++ b/das/daser.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sync/atomic"
 	"time"
 
 	"github.com/ipfs/go-datastore"
@@ -19,18 +20,16 @@ var log = logging.Logger("das")
 
 // DASer continuously validates availability of data committed to headers.
 type DASer struct {
-	da    share.Availability
-	bcast fraud.Broadcaster
-	hsub  header.Subscriber
-
-	// getter allows the DASer to fetch an ExtendedHeader (EH) at a certain height
-	// and blocks until the EH has been processed by the header store.
-	getter header.Getter
-	// checkpoint store
-	cstore datastore.Datastore
-
 	ctx    context.Context
 	cancel context.CancelFunc
+
+	da     share.Availability
+	bcast  fraud.Broadcaster
+	hsub   header.Subscriber // listens for new headers in the network
+	getter header.Getter     // retrieves past headers
+
+	cstore datastore.Datastore // checkpoint store
+	state  state
 
 	jobsCh chan *catchUpJob
 
@@ -113,8 +112,12 @@ func (d *DASer) Stop(ctx context.Context) error {
 
 // sample validates availability for each Header received from header subscription.
 func (d *DASer) sample(ctx context.Context, sub header.Subscription, checkpoint int64) {
+	// indicate sampling routine is running
+	atomic.StoreUint64(&d.state.sample.IsRunning, 1)
 	defer func() {
 		sub.Cancel()
+		// indicate sampling routine is stopped
+		atomic.StoreUint64(&d.state.sample.IsRunning, 0)
 		// send done signal
 		d.sampleDn <- struct{}{}
 	}()
@@ -153,9 +156,14 @@ func (d *DASer) sample(ctx context.Context, sub header.Subscription, checkpoint 
 
 		err = d.sampleHeader(ctx, h)
 		if err != nil {
+			// record error
+			d.updateSampleState(h, err)
+			log.Warn("DASer SAMPLING ROUTINE WILL BE STOPPED. IN ORDER TO CONTINUE SAMPLING, " +
+				"RE-START THE NODE")
 			return
 		}
 
+		d.updateSampleState(h, nil)
 		sampleHeight = h.Height
 	}
 }
@@ -187,13 +195,22 @@ func (d *DASer) catchUpManager(ctx context.Context, checkpoint int64) {
 		case <-ctx.Done():
 			return
 		case job := <-d.jobsCh:
+			// record details of incoming job
+			d.recordJobDetails(job)
 			// perform catchUp routine
 			height, err := d.catchUp(ctx, job)
 			// store the height of the last successfully sampled header
 			checkpoint = height
 			// exit routine if a catch-up job was unsuccessful
 			if err != nil {
+				// record error
+				d.state.catchUpLk.Lock()
+				d.state.catchUp.Error = err
+				d.state.catchUpLk.Unlock()
+
 				log.Errorw("catch-up routine failed", "attempted range: from", job.from, "to", job.to)
+				log.Warn("DASer CATCH-UP SAMPLING ROUTINE WILL BE STOPPED. IN ORDER TO CONTINUE SAMPLING, " +
+					"RE-START THE NODE")
 				return
 			}
 		}
@@ -203,7 +220,6 @@ func (d *DASer) catchUpManager(ctx context.Context, checkpoint int64) {
 // catchUp starts a sampling routine for headers starting at the next header
 // after the `from` height and exits the loop once `to` is reached. (from:to]
 func (d *DASer) catchUp(ctx context.Context, job *catchUpJob) (int64, error) {
-	routineStartTime := time.Now()
 	log.Infow("sampling past headers", "from", job.from, "to", job.to)
 
 	// start sampling from height at checkpoint+1 since the
@@ -226,10 +242,17 @@ func (d *DASer) catchUp(ctx context.Context, job *catchUpJob) (int64, error) {
 		if err != nil {
 			return h.Height - 1, err
 		}
+		d.state.catchUpLk.Lock()
+		d.state.catchUp.Height = uint64(h.Height)
+		d.state.catchUpLk.Unlock()
 	}
+	d.state.catchUpLk.Lock()
+	d.state.catchUp.End = time.Now()
+	d.state.catchUpLk.Unlock()
 
+	jobDetails := d.CatchUpRoutineState()
 	log.Infow("successfully sampled past headers", "from", job.from,
-		"to", job.to, "finished (s)", time.Since(routineStartTime))
+		"to", job.to, "finished (s)", jobDetails.Duration())
 	// report successful result
 	return job.to, nil
 }
@@ -252,7 +275,6 @@ func (d *DASer) sampleHeader(ctx context.Context, h *header.ExtendedHeader) erro
 		}
 		log.Errorw("sampling failed", "height", h.Height, "hash", h.Hash(),
 			"square width", len(h.DAH.RowsRoots), "data root", h.DAH.Hash(), "err", err)
-		log.Warn("DASer WILL BE STOPPED. IN ORDER TO CONTINUE SAMPLING, RE-START THE NODE")
 		// report previous height as the last successfully sampled height
 		return err
 	}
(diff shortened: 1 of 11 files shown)


cmd/flags_core.go
```diff
@@ -73,7 +73,7 @@ func validateAddress(address string) (string, string, error) {
 		return "", "", fmt.Errorf("both protocol and host must present in the address")
 	}
 
-	if _, port, err := net.SplitHostPort(u.Host); err != nil || port == "" {
+	if _, _, err := net.SplitHostPort(u.Host); err != nil {
```
renaynay, line 76: This check to ensure port is not empty is unnecessary, it will fail if port is empty.

das/state.go
```diff
@@ -0,0 +1,54 @@
...
+	LatestSampledHeight() uint64
+}
+
+// state collects information about the DASer process.
+type state struct {
+	// tracks whether sampling routine is running
+	isRunning uint64
+	// indicates whether the DASer catchUp routine
+	// has successfully caught up to network head.
+	caughtUp uint64
```
adlerjohn, line 18: This is used as a Boolean, so should be prefixed with `is` to make is clear from the name.
```suggestion
	isCaughtUp uint64
```

das/state.go
```diff
@@ -0,0 +1,54 @@
+package das
+
+import "sync/atomic"
+
+// State is an interface that allows other components to access the
+// current state of the DASer safely.
+type State interface {
```
adlerjohn, line 7: Should `IsRunning()` be added to this interface?

das/daser.go
```diff
@@ -201,6 +214,9 @@ func (d *DASer) catchUpManager(ctx context.Context, checkpoint int64) {
 // catchUp starts a sampling routine for headers starting at the next header
 // after the `from` height and exits the loop once `to` is reached. (from:to]
 func (d *DASer) catchUp(ctx context.Context, job *catchUpJob) (int64, error) {
+	// indicate that DASer is catching up to local chain head
+	atomic.StoreUint64(&d.state.caughtUp, 0)
```
adlerjohn, line 218: I don't really feel comfortable with dealing with literals here. How to improve:

1. Use constants instead of literals.
2. Even better, hide the setting/getting behind an interface that uses the appropriate types (i.e. `bool` in this case).

das/state.go
renaynay: The interface was actually unused. I removed it.

das/state.go
renaynay: done

das/state.go
```diff
@@ -0,0 +1,71 @@
...
+
+import "sync/atomic"
+
+// state collects information about the DASer process.
+type state struct {
+	// tracks whether sampling routine is running
+	isRunning uint64
+	// indicates whether the DASer catchUp routine
+	// has successfully caught up to network head.
+	isCaughtUp uint64
```
Wondertan, line 11: While you can represent each state with booleans, adding more booleans per each new state is not sustainable + it's getting harder to reason about the code when you have multiple booleans and the relations between them.

I would suggest from the beginning starting with a simple state machine which can be extended going further:
> Stopped -> Running -> CatchingUp
With these 3 states, you can explain any state of a DASer.

das/daser.go
```diff
@@ -153,6 +160,11 @@ func (d *DASer) sample(ctx context.Context, sub header.Subscription, checkpoint
 		log.Infow("sampling successful", "height", h.Height, "hash", h.Hash(),
 			"square width", len(h.DAH.RowsRoots), "finished (s)", sampleTime.Seconds())
 
+		d.stateLk.Lock()
+		d.state.latestSampledHeight = uint64(h.Height)
+		d.state.latestSampledSquareWidth = uint64(len(h.DAH.RowsRoots))
+		d.stateLk.Unlock()
+
```
Wondertan, line 167: Hmm, you made a method for every access of the state, besides here.

das/state.go
Wondertan, line 11: While you can represent each state with booleans, adding more booleans per each new state is not sustainable + it's getting harder to reason about the code when you have multiple booleans and the relations between them.

I would suggest from the beginning starting with a simple state machine which can be extended going further:
>Stopped -> Running -> CatchingUp -> Error

With these 4 states, you can explain any state of a DASer.

das/daser_test.go
```diff
@@ -251,6 +251,58 @@ func TestDASer_catchUp_fails(t *testing.T) {
...
+	err := daser.Start(ctx)
+	require.NoError(t, err)
+	defer func() {
+		// wait for all "future" headers to be sampled
+		for {
+			select {
+			case <-ctx.Done():
+				t.Fatal(ctx.Err())
+			default:
+				if daser.LatestSampledHeight() == expectedFinalSampleHeight {
```
Wondertan, line 276: While it is still more or less fine for a test to do a busy loop(wastes CPU constantly), the real implementation should not be doing this and there should be a way to _wait_ for the desired height to be dased.

cmd/flags_core.go
```diff
@@ -73,7 +73,7 @@ func validateAddress(address string) (string, string, error) {
 		return "", "", fmt.Errorf("both protocol and host must present in the address")
 	}
 
-	if _, port, err := net.SplitHostPort(u.Host); err != nil || port == "" {
```
Wondertan, line 76: Honestly, this is not part of this PR, especially if you are going to squash it into one commit.

das/state.go
Wondertan: I can give some examples of how to implement it, if you want

Wondertan (changes requested):
We also need to keep the error in the state if any

das/state.go
renaynay: @Wondertan Yes i agree that we shouldn't be adding more booleans, but there are two clear routines running here: 
* sampling over current network headers
* sampling from last sampled checkpoint -> network head. 

It would be good to distinguish the two - from a user perspective and from a debugging perspective.

i agree though that this needs to be improved somehow bc right now `running` refers to main sample loop over current headers, and catchup refers to catchup loop.

das/daser_test.go
renaynay: This would require a das sub similar to how there is a heightsub.

The problem is, from a user standpoint, we already have the ability to check if a height has been dased via `CacheAvailability` and the endpoint that I've been working on here #802 

If you think this is really necessary beyond testing, I'm happy to make an issue, but for now, I wouldn't see why we need to wait for DASer to get to a certain height. For example, if a user/rollup wants to see if block at height 5000 is available and daser is still catching up at height 3400 or something, they wouldn't want to sit and wait for daser to catch up to 5000, they would just call `SharesAvailable` under the hood of the endpoint provided in #802, which will check if shares available locally, and if not, will trigger a share sampling process on the header 5000.

renaynay: I'll make an issue for this as a potential enhancement but it's not high priority bc it's only testing utility for now.

adlerjohn (commented):
merge conflict

service/share/testing.go
```diff
@@ -156,3 +156,16 @@ func NewBrokenAvailability() Availability {
 func (b *brokenAvailability) SharesAvailable(context.Context, *Root) error {
 	return ErrNotAvailable
 }
+
+// TestAvailability returns an Availability that always
+// returns successfully when SharesAvailable is called.
```
adlerjohn, line 161: Should this comment be on `NewTestAvailability`?

das/state.go
```diff
@@ -0,0 +1,114 @@
...
+
+	catchUpStateLk sync.RWMutex
+	catchUpState   RoutineState // tracks information related to the `catchUp` routine
+}
+
+// RoutineState contains important information about the state of a
+// current sampling routine.
+type RoutineState struct {
+	// tracks whether routine is running
+	IsRunning uint64 `json:"is_running"`
```
vgonkivs, line 26: You can use https://github.com/uber-go/atomic/blob/master/bool.go instead of uint64

service/share/testing.go
```diff
@@ -242,3 +242,16 @@ func NewBrokenAvailability() Availability {
 func (b *brokenAvailability) SharesAvailable(context.Context, *Root) error {
 	return ErrNotAvailable
 }
+
+type TestAvailability struct {
```
vgonkivs, line 246: We already have `brokenAvailability` that always fails on `SharesAvailable`. May be we should give better naming to this one to be consistent ?

service/rpc/das.go
```diff
@@ -0,0 +1,52 @@
...
+	LatestSampledHeight uint64 `json:"latest_sampled_height"`
+	LatestSampledWidth  uint64 `json:"latest_sampled_square_width"`
+
+	Err map[string]das.SampleError `json:"errors"`
+}
+
+func (h *Handler) handleDASStateRequest(w http.ResponseWriter, r *http.Request) {
+	dasState := new(DasStateResponse)
+	sampleState := h.das.SampleRoutineState()
+	catchUpState := h.das.CatchUpRoutineState()
```
vgonkivs, line 29: What if das state will change between fetching the state and returning the response?

service/rpc/das.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"encoding/json"
+	"net/http"
+
+	"github.com/celestiaorg/celestia-node/das"
+)
+
+const (
+	dasStateEndpoint = "/daser"
```
vgonkivs, line 11: `/daser/state` ?

renaynay: It would actually be nice to consolidate all of the current endpoints we have  / and also consolidate endpoints once we merge #802 . 

I will make an issue for this.

renaynay: Actually, the problem is, we have other endpoints like `/data_available`, `/namespaced_shares`, so I think it should be `/daser_state` maybe.

service/share/testing.go
Wondertan: We seem to be aligned with Dummy wording

service/rpc/das.go
vgonkivs: Do we expect to have endpoints that will return other services state? Then we can have `/state/daser`, `/state/syncer`.

Wondertan:
@renaynay, there is a race specifically in the test code

service/rpc/das.go
Wondertan: +1 for daser_stare or daser/state

das/daser.go
```diff
@@ -263,3 +284,52 @@ func (d *DASer) sampleHeader(ctx context.Context, h *header.ExtendedHeader) erro
...
+	state := d.state.sample
+	d.state.sampleLk.RUnlock()
+	return state
+}
+
+func (d *DASer) updateSampleState(h *header.ExtendedHeader, err error) {
+	height := uint64(h.Height)
+
+	d.state.sampleLk.Lock()
+	defer d.state.sampleLk.Unlock()
```
adlerjohn, line 301: Is there a reason to use `defer` here but not in the other functions?

das/state.go
```diff
@@ -0,0 +1,53 @@
...
+	Height uint64 `json:"height"`
+	From   uint64 `json:"from"`
+	To     uint64 `json:"to"`
+}
+
+func (ji JobInfo) Finished() bool {
+	return ji.To == ji.Height
+}
+
+func (ji JobInfo) Duration() time.Duration {
```
Wondertan, line 51: Good find. I would add this to the Syncer as well. GFI?

renaynay:
I've extracted some modifiers (indicateRunning, indicateStopped) into methods to not crowd the otherwise long function. I can undo that for consistency, or I can keep it to maintain a cleaner looking function (`sample`).

Re one public State, I think i'll just keep it separate for now, and if there is a reason for it later, we can provide a method that returns both at once.

WDYT @Wondertan ?

Also pushing the duration method to syncer rn

vgonkivs (approved):

Wondertan:
>  can undo that for consistency, or I can keep it to maintain a cleaner looking function (sample).

I am fine with the way it is right now. Consistency is important, but this is not the case where it is really important, e.g. by the following consistency in one place, you can brake it for another one.

> Re one public State, I think i'll just keep it separate for now, and if there is a reason for it later, we can provide a method that returns both at once.

It's the other way around, though. There is no reason to keep em separated, but it's fine

Wondertan:
However, I would also fix the defer usage as per Johns's comment. Either to use defer or not to use it consistently

das/daser.go
renaynay: https://github.com/celestiaorg/celestia-node/pull/848/commits/821d5108f178fa8e84d98cf348514cbdc01b76fa done

adlerjohn (approved):

PR #850, opened 2022-06-22
cmd/celestia: changing randomness seed

### LINKED ISSUE #678: cmd/celestia: change randomness seed
Changes randomness seed in the main so that it is not deterministic via `rand.Seed`. Specifically, this way, Full Nodes will sample quadrants randomly. The whole change is as simple as the following:
```go
import rand
...
func main() {
	rand.Seed(time.Now().UnixNano())
	...
}
```

distractedm1nd:
Wasn't sure what the conclusion of the last discussion was, so hoping this resparks a convo about seeding the randomness to close #678 .


cmd/celestia/main.go
```diff
@@ -35,6 +38,13 @@ func main() {
 }
 
 func run() error {
+	seed := make([]byte, 32)
+	_, err := crypto_rand.Read(seed)
+	if err != nil {
+		return err
+	}
+	math_rand.Seed(int64(binary.LittleEndian.Uint64(seed)))
```
liamsi, line 46: This definitely works to make picking the quadrant less deterministic. That said, I feel the code should be closer to where the randomness is actually used. Maybe in an init. Also, is there a way to use your own instantiation of the a prng instead of using the global one from math/rand? Then it could be instantiated as the above without having any side-effects.

distractedm1nd: We could `init` it somewhere and then pass around a `*rand.Rand` maybe?

liamsi: > We could init it somewhere and then pass around a *rand.Rand maybe?

Yeah, or even construct that rand.Rand where needed once as an unexported field. Both works.

Wondertan: * There is global randomness and there could be other places that use the global randomness, not only quadrant requesting logic. It is ok to set up this global randomness on the application startup. It is normal in Go to use global randmoness for cases where cryptographic randmoness is not required.
* Whether quadrant fetching logic should use or not use global randomness is orthogonal.

Wondertan: Regarding the change itself. Seeding with `time.Now().Unix()` is enough and feeding math_rand with crypto_rand's number does not give cryptographic secureness. 
> Funnily, I had similar impression, but than Go's crypto lead explained me this, somewhere in celestia-core repo. @liamsi should remember that PR. The lead even made a change to a GO's doc pointing this out and linked the lazyledger-core back then into the commit history.

Wondertan: Wanna make an issue with some justification about using crypto.Rand instead? 
> Every time we discuss this, I wonder if we should not simply use crypto/rand everywhere instead :D

Wondertan, line 46: As per https://github.com/celestiaorg/celestia-node/pull/850/files#r906162596, lets simplify the code here by using time.Unix and we are good to go with this PR. Thanks

liamsi: Yeah, can do: https://github.com/celestiaorg/celestia-node/issues/865

Wondertan (approved):

adlerjohn (approved):

PR #863, opened 2022-06-27
p2p: change conn gater type

### LINKED ISSUE #819: node/p2p: blacklist set of peers

vgonkivs:
Resolves #819


renaynay (approved):
So `BlockPeer` only means that the node cannot dial the peer? But it can still accept incoming connections from the peer?

Wondertan:
@renaynay, consult the documentation on the ConnGater interface https://github.com/libp2p/go-libp2p-core/blob/master/connmgr/gater.go#L54

PR #592, opened 2022-04-05
tests(node/swamp): block reconstruction

Bidon15:
1. Full node block reconstruction from bridge node only 
2. Full node block reconstruction from ligts (s=20;k=16;c=69)

Issue #483, opened 2022-02-28
das: Parallelise `catchUp` method


renaynay:
Supplement https://github.com/celestiaorg/celestia-node/pull/473 by additionally parallelising the DASing of past headers. 

TODOs mentioned [here](https://github.com/celestiaorg/celestia-node/pull/473#discussion_r828138649).

renaynay:
<img width="795" alt="Screenshot 2022-03-21 at 15 08 58" src="https://user-images.githubusercontent.com/41963722/159278716-9e18c92a-c645-42e0-b438-1cad483b0cf8.png">

renaynay:
Now that [DASState](#848) is almost implemented, we should consider the storing of errors for parallel `catchUp` workers whenever this issue is tackled.

Bidon15:
We should revisit https://github.com/celestiaorg/celestia-node/issues/504 and close it if after implementation of parallelisation, the issue is not reproducible again

PR #870, opened 2022-06-29
fix(share): protect datastore autobatch in CacheAvailability with a lock

Wondertan:
During the review of #786, I was assuming the autobatch datastore has an internal lock, but it turns out to be not the case.

Was found while making #592

### DIFF
--- a/service/share/cache_availability.go
+++ b/service/share/cache_availability.go
@@ -3,6 +3,7 @@ package share
 import (
 	"bytes"
 	"context"
+	"sync"
 
 	"github.com/ipfs/go-datastore"
 	"github.com/ipfs/go-datastore/autobatch"
@@ -27,7 +28,10 @@ var (
 type CacheAvailability struct {
 	avail Availability
 
-	ds *autobatch.Datastore
+	// TODO(@Wondertan): Once we come to parallelized DASer, this lock becomes a contention point
+	//  Related to #483
+	dsLk sync.RWMutex
+	ds   *autobatch.Datastore
 }
 
 // NewCacheAvailability wraps the given Availability with an additional datastore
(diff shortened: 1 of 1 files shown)


vgonkivs (approved):

service/share/cache_availability.go
```diff
@@ -50,18 +53,22 @@ func (ca *CacheAvailability) SharesAvailable(ctx context.Context, root *Root) er
...
-	}
+
 	err = ca.avail.SharesAvailable(ctx, root)
 	if err != nil {
 		return err
 	}
+
+	ca.dsLk.Lock()
 	err = ca.ds.Put(ctx, key, []byte{})
+	ca.dsLk.Unlock()
```
Wondertan, line 71: By contention point TODO I meant specifically this place. Basically, the routine hitting the autobatching threshold will lock all the other writers in future parallelized DASer until the batch is synced on disk. This fundamentally degrades the parallelization of a DASer. The solution I know is to decouple reads and write in header.Store

service/share/cache_availability.go
```diff
@@ -27,7 +28,9 @@ var (
 type CacheAvailability struct {
 	avail Availability
 
-	ds *autobatch.Datastore
+	// TODO(@Wondertan): Once we come to parallelized DASer, this lock becomes a contention point
```
renaynay, line 31: Can you link to issue https://github.com/celestiaorg/celestia-node/issues/483

service/share/cache_availability.go
```diff
@@ -50,18 +53,22 @@ func (ca *CacheAvailability) SharesAvailable(ctx context.Context, root *Root) er
 	}
 	// do not sample over Root that has already been sampled
 	key := rootKey(root)
+
+	ca.dsLk.Lock()
```
renaynay, line 57: RLock?

service/share/cache_availability.go
```diff
@@ -50,18 +53,22 @@ func (ca *CacheAvailability) SharesAvailable(ctx context.Context, root *Root) er
...
 	exists, err := ca.ds.Has(ctx, key)
-	if err != nil {
+	ca.dsLk.Unlock()
```
renaynay, line 59: RUnlock?

service/share/cache_availability.go
renaynay: Can you document this pls on #483

service/share/cache_availability.go
Wondertan: Oh yeah lol. I thought I made an Rlock there. I remember looking at source of has to check if it doesn't do any writes in the background. Thanks for catching

service/share/cache_availability.go
renaynay: whoops accidentally resolved.. I don't know if u want to resolve or not

renaynay (approved):

PR #820, opened 2022-06-13
node: integrate fraud service

vgonkivs:
The last part of #674

PR #872, opened 2022-06-29
swamp: fix TestFraudProofBroadcasting call of BlockTime

Bidon15:
On latest main run, CI jobs have detected an issue : 
https://github.com/celestiaorg/celestia-node/runs/7115039774?check_suite_focus=true

### DIFF
--- a/node/tests/fraud_test.go
+++ b/node/tests/fraud_test.go
@@ -29,7 +29,7 @@ import (
  Note: 15 is not available because DASer will be stopped before reaching this height due to receiving BEFP.
 */
 func TestFraudProofBroadcasting(t *testing.T) {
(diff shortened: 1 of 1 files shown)


vgonkivs (approved):

Wondertan:
A bit of context. #592 changed the option, while #820 uses it. The #820 was merged right before #592 and there were no conflicts for  #592, so I merged it without rebasing, which turned out to be breaking for the main branch

Bidon15:
> A bit of context. #592 changed the option, while #820 uses it. The #820 was merged right before #592 and there were no conflicts for #592, so I merged it without rebasing, which turned out to be breaking for the main branch

Yeah. It's just unfortunate timing tbh

PR #886, opened 2022-07-05
Add CI check for go mod tidy.

### LINKED ISSUE #884: Add `go mod tidy` check to CI

adlerjohn:
Fixes #884

### DIFF
--- a/.github/workflows/go-mod-tidy.yml
+++ b/.github/workflows/go-mod-tidy.yml
@@ -1,0 +1,28 @@
+name: Go Mod Tidy Check
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+  release:
+    types: [published]
+
+env:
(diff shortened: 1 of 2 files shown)


.github/workflows/go-mod-tidy.yml
```diff
@@ -0,0 +1,28 @@
...
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+  release:
+    types: [published]
+
+env:
+  GO_VERSION: 1.18
```
Wondertan, line 12: We really need to have some global var for this. Last time I made an update, I had to update this field in like ~7 places

adlerjohn: Ideally we should combine all our Go CI into a single file. Filing an issue

vgonkivs (approved):

PR #891, opened 2022-07-06
node: providing separated CacheAvailability to DASer and ShareService

### LINKED ISSUE #879: node: Separate out construction of `CacheAvailability` into separate function that accepts an `Availability`
Separate out construction of `CacheAvailability`, have it accept a general availability. 

[Here](https://github.com/celestiaorg/celestia-node/blob/main/node/services/service.go#L171-L176) and [here](https://github.com/celestiaorg/celestia-node/blob/main/node/services/service.go#L181-L186)

distractedm1nd:
Attempts to close #879. This PR is more of a question, so leaving it as a draft at first. 

Is it fine for the `DASer` and `ShareService` to take `CacheAvailability`? New to the `fx` package, my goal in doing this was that the `Availability` generated by `service.LightAvailability` and `service.FullAvailability` is only used to generate `CacheAvailability`.

I'm still trying to understand the big picture a bit, will this cache always be used for all node types, or should it be optional?

### DIFF
--- a/node/components.go
+++ b/node/components.go
@@ -32,6 +32,7 @@ func lightComponents(cfg *Config, store Store) fx.Option {
 		fx.Provide(services.DASer),
 		fx.Provide(services.HeaderExchangeP2P(cfg.Services)),
 		fx.Provide(services.LightAvailability),
+		fx.Provide(services.CacheAvailability[*share.LightAvailability]),
 		fx.Invoke(rpc.Handler),
 	)
 }
@@ -44,6 +45,7 @@ func bridgeComponents(cfg *Config, store Store) fx.Option {
 		nodecore.Components(cfg.Core),
 		fx.Supply(header.MakeExtendedHeader),
 		fx.Provide(services.FullAvailability),
+		fx.Provide(services.CacheAvailability[*share.FullAvailability]),
 		fx.Invoke(func(
 			state *state.Service,
 			share *share.Service,
@@ -63,6 +65,7 @@ func fullComponents(cfg *Config, store Store) fx.Option {
 		fx.Provide(services.DASer),
 		fx.Provide(services.HeaderExchangeP2P(cfg.Services)),
 		fx.Provide(services.FullAvailability),
+		fx.Provide(services.CacheAvailability[*share.FullAvailability]),
 		fx.Invoke(rpc.Handler),
 	)
 }
--- a/service/share/cache_availability.go
+++ b/service/share/cache_availability.go
@@ -46,16 +46,6 @@ func NewCacheAvailability(avail Availability, ds datastore.Batching) *CacheAvail
 	}
 }
 
-// Start starts the underlying Availability.
-func (ca *CacheAvailability) Start(ctx context.Context) error {
-	return ca.avail.Start(ctx)
-}
-
-// Stop is an alias for Close to conform to the Availability interface.
-func (ca *CacheAvailability) Stop(ctx context.Context) error {
-	return ca.Close(ctx)
-}
-
 // SharesAvailable will store, upon success, the hash of the given Root to disk.
 func (ca *CacheAvailability) SharesAvailable(ctx context.Context, root *Root) error {
 	// short-circuit if the given root is minimum DAH of an empty data square
--- a/service/share/cache_availability.go
+++ b/service/share/cache_availability.go
@@ -9,7 +9,6 @@ import (
 	"github.com/ipfs/go-datastore/autobatch"
 	"github.com/ipfs/go-datastore/namespace"
 	"github.com/tendermint/tendermint/pkg/da"
-	"go.uber.org/multierr"
 )
 
 var (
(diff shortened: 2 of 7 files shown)


renaynay:
Hey @distractedm1nd , this is a good start! I see you're starting to figure out the nuances of the fx lib we use :)

There are two challenges here: 

1. All services that receive `share.Availability` must received an instance that is *already wrapped* by `share.CacheAvailability`. E.g., the `DASer` must receive a `share.CacheAvailability` that wraps either Full/Light Availability. The reason being is that `CacheAvailability` stores the result of a successful sampling process so that there aren't any redundant/duplicate sampling processes -- a header that has already been sampled for will not be sampled for again if it was successful the first time 😄 
2. `share.CacheAvailability` should control the lifecycle of its underlying Availability. E.g. Start on `CacheAvailability` would also start it's underlying Availability. Same goes for Stop (in CacheAvailability, this is called `Close` right now).

distractedm1nd:
Alright, so the diff solves 1 already right? By passing in the `share.CacheAvailability` to `service.DASer` and `service.ShareService` it uses the wrapped availability correctly (well, I crudely tested it with logging). Is there anywhere else where a `share.Availability` would be injected by fx? 

As for number two, is there a method needed to start the Availability lifecycle? Would it be correct to say that there is no setup needed, because the initialized `share.CacheAvailability` is passed into the services that need them directly through fx? I **think** the `Close` is still being invoked correctly, I'll have to look into writing tests using `fx` though to make sure

distractedm1nd:
Ok, will rebase - thanks! If you pull my diff and add a log inside the `CacheAvailability`, you will see that its constructor is called - since the DASer and ShareService use it, fx injects it. `CacheAvailability` is only not constructed if we don't use it in those services (because then it just uses the share.Availability directly)

distractedm1nd:
Oh wow, yeah, a lot has changed! Guess I forgot to pull before branching 🥴 see what you mean now

Wondertan:
> if you pull my diff and add a log inside the CacheAvailability, you will see that its constructor is called - since the DASer and ShareService use it, fx injects it

Yeah, and the reason is that you changed other constructors to directly rely on `*CacheAvailability` type and we should not do that.

distractedm1nd:
Not sure if this is the direction you were aiming at, looking forward to feedback

Wondertan:
Yep, we discuss this with @renaynay in sync. Patterns:
* Lifecycles are only managed by FX. 
* Components do not manage lifecycles of their dependencies

service/share/cache_availability.go
```diff
@@ -52,11 +51,9 @@ func (ca *CacheAvailability) Start(ctx context.Context) error {
 	return ca.avail.Start(ctx)
 }
 
-// Stop stops the underlying Availability and closes the CacheAvailability.
+// Stop is an alias for Close to conform to the Availability interface.
```
distractedm1nd, line 54: I really don't like this, does "Close" need to exist for some other reason or can we just replace it entirely with "Stop"? Or is this kind of alias okay? 😬

Wondertan (changes requested):
I guess there was some misunderstanding in the discussion above. 
* We don't extend the Availability interface with Start/Stop
* We don't allow controlling lifecycles of inner components, e.g CacheAvailability does not call Start and Stop for wrapped Availability
* We don't merge the main into the working branch and rebase on the main instead. Otherwise, the git history is getting pretty dirty

service/share/cache_availability.go
Wondertan: I don't like it as well. It doesn't need to exist. See my review below

Wondertan: Also, see https://github.com/celestiaorg/celestia-node/pull/891#issuecomment-1179186540

distractedm1nd:
Ohhhhh! Yeah, sorry for the misunderstanding - as for the git history, what is the cleanest way for me to fix it? Just revert the merge commit and then rebase?

Wondertan:
@distractedm1nd, revert still keeps the existing history of added commit and the revert commit, so this won't work. Instead, pls reset your branch on the commit before the merge commit, do rebase, and force push the result. Alternatively, start from a fresh branch and apply your changes in a single commit on top.

node/components.go
```diff
@@ -44,6 +45,7 @@ func bridgeComponents(cfg *Config, store Store) fx.Option {
 		nodecore.Components(cfg.Core),
 		fx.Supply(header.MakeExtendedHeader),
 		fx.Provide(services.LightAvailability), // TODO(@Wondertan): Remove strict requirements to have Availability
```
Wondertan, line 47: You can remove this TODO now as it is irrelevant and make it Full, as discussed in sync.

Wondertan (approved):
LGTM

distractedm1nd:
Rebased on 924

renaynay (approved):

PR #892, opened 2022-07-06
share: handle disconnects of discovered peers

### LINKED ISSUE #829: p2p: implement `heartbeat` for discovered peers
In order to keep only active connections with discovered peers, we should implement a mechanism that will detect once discovered node is disconnected and re-start peer discovery.

Depends on https://github.com/celestiaorg/celestia-node/pull/799
Supplement for #599

vgonkivs:
resolves #829 

- [x] Removing ConnManager Tag for disconnected peer.
- [x] Backing off a peer, so we don't reconnect to it within a certain period.
- [x] Find more peers if we are below the peerLimit.
- [x] update the reconstruction tests to include the discovery bit in them.

### DIFF
--- a/logs/logs.go
+++ b/logs/logs.go
@@ -14,6 +14,7 @@ func SetAllLoggers(level logging.LogLevel) {
 	_ = logging.SetLogLevel("connmgr", "WARN")
 	_ = logging.SetLogLevel("nat", "INFO")
 	_ = logging.SetLogLevel("dht/RtRefreshManager", "FATAL")
+	_ = logging.SetLogLevel("bitswap_network", "ERROR")
 }
 
 func SetDebugLogging() {
--- a/node/node.go
+++ b/node/node.go
@@ -133,16 +133,23 @@ func (n *Node) Run(ctx context.Context) error {
 // Stop shuts down the Node, all its running Components/Services and returns.
 // Canceling the given context earlier 'ctx' unblocks the Stop and aborts graceful shutdown forcing remaining
 // Components/Services to close immediately.
-func (n *Node) Stop(ctx context.Context) error {
+func (n *Node) Stop(ctx context.Context) (err error) {
 	ctx, cancel := context.WithTimeout(ctx, Timeout)
 	defer cancel()
+	defer func() {
+		if err != nil {
+			log.Errorf("Stopping %s Node: %s", n.Type, err)
 
-	err := n.stop(ctx)
+		}
+	}()
+	err = n.stop(ctx)
 	if err != nil {
-		log.Errorf("Stopping %s Node: %s", n.Type, err)
 		return err
 	}
-
+	err = n.Host.Close()
+	if err == nil {
+		return err
+	}
 	log.Infof("stopped %s Node", n.Type)
 	return nil
 }
--- a/node/node.go
+++ b/node/node.go
@@ -133,23 +133,16 @@ func (n *Node) Run(ctx context.Context) error {
 // Stop shuts down the Node, all its running Components/Services and returns.
 // Canceling the given context earlier 'ctx' unblocks the Stop and aborts graceful shutdown forcing remaining
 // Components/Services to close immediately.
-func (n *Node) Stop(ctx context.Context) (err error) {
+func (n *Node) Stop(ctx context.Context) error {
 	ctx, cancel := context.WithTimeout(ctx, Timeout)
 	defer cancel()
-	defer func() {
-		if err != nil {
-			log.Errorf("Stopping %s Node: %s", n.Type, err)
 
-		}
-	}()
-	err = n.stop(ctx)
+	err := n.stop(ctx)
 	if err != nil {
+		log.Errorf("Stopping %s Node: %s", n.Type, err)
 		return err
 	}
-	err = n.Host.Close()
-	if err == nil {
-		return err
-	}
+
 	log.Infof("stopped %s Node", n.Type)
 	return nil
 }
--- a/node/settings.go
+++ b/node/settings.go
@@ -97,6 +97,7 @@ func WithBootstrappers(peers params.Bootstrappers) Option {
 }
 
 // WithRefreshRoutingTablePeriod sets custom refresh period for dht.
+// Currently, it is used to speed up tests.
 func WithRefreshRoutingTablePeriod(interval time.Duration) Option {
 	return func(sets *settings) {
 		sets.cfg.P2P.RoutingTableRefreshPeriod = interval
--- a/node/tests/p2p_test.go
+++ b/node/tests/p2p_test.go
@@ -138,6 +138,7 @@ func TestBootstrapNodesFromBridgeNode(t *testing.T) {
 		assert.True(t, light.Host.Network().Connectedness(addrFull.ID) == network.Connected)
 	}
 	require.NoError(t, full.Stop(ctx))
+	require.NoError(t, full.Host.Network().Close())
 	select {
 	case <-ctx.Done():
 		t.Fatal("peer was not disconnected")
--- a/node/tests/p2p_test.go
+++ b/node/tests/p2p_test.go
@@ -85,6 +85,7 @@ Steps:
 5. Ensure that nodes are connected to bridge
 6. Wait until light will find full node
 7. Check that full and light nodes are connected to each other
+8. Stop FN and ensure that it's not connected to LN
 */
 func TestBootstrapNodesFromBridgeNode(t *testing.T) {
 	sw := swamp.NewSwamp(t)
@@ -118,19 +119,29 @@ func TestBootstrapNodesFromBridgeNode(t *testing.T) {
 			ch <- struct{}{}
 		}
 	}
+	bundle.DisconnectedF = func(_ network.Network, conn network.Conn) {
+		if conn.RemotePeer() == full.Host.ID() {
+			ch <- struct{}{}
+		}
+	}
 	light.Host.Network().Notify(bundle)
 	for index := range nodes {
 		require.NoError(t, nodes[index].Start(ctx))
 		assert.Equal(t, *addr, nodes[index].Bootstrappers[0])
 		assert.True(t, nodes[index].Host.Network().Connectedness(addr.ID) == network.Connected)
 	}
-
+	addrFull := host.InfoFromHost(full.Host)
 	select {
 	case <-ctx.Done():
 		t.Fatal("peer was not found")
 	case <-ch:
-		break
+		assert.True(t, light.Host.Network().Connectedness(addrFull.ID) == network.Connected)
 	}
-	addrFull := host.InfoFromHost(full.Host)
-	assert.True(t, light.Host.Network().Connectedness(addrFull.ID) == network.Connected)
+	require.NoError(t, full.Stop(ctx))
+	select {
+	case <-ctx.Done():
+		t.Fatal("peer was not disconnected")
+	case <-ch:
+		assert.True(t, light.Host.Network().Connectedness(addrFull.ID) == network.NotConnected)
+	}
 }
--- a/node/tests/p2p_test.go
+++ b/node/tests/p2p_test.go
@@ -8,6 +8,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	"github.com/libp2p/go-libp2p-core/event"
 	"github.com/libp2p/go-libp2p-core/host"
 	"github.com/libp2p/go-libp2p-core/network"
 	"github.com/libp2p/go-libp2p-core/peer"
@@ -113,30 +114,32 @@ func TestBootstrapNodesFromBridgeNode(t *testing.T) {
 	light := sw.NewLightNode(nodesConfig...)
 	nodes := []*node.Node{full, light}
 	ch := make(chan struct{})
-	bundle := &network.NotifyBundle{}
-	bundle.ConnectedF = func(_ network.Network, conn network.Conn) {
-		if conn.RemotePeer() == full.Host.ID() {
-			ch <- struct{}{}
-		}
-	}
-	bundle.DisconnectedF = func(_ network.Network, conn network.Conn) {
-		if conn.RemotePeer() == full.Host.ID() {
-			ch <- struct{}{}
-		}
-	}
-	light.Host.Network().Notify(bundle)
+	sub, err := light.Host.EventBus().Subscribe(&event.EvtPeerConnectednessChanged{})
+	require.NoError(t, err)
+	defer sub.Close()
 	for index := range nodes {
 		require.NoError(t, nodes[index].Start(ctx))
 		assert.Equal(t, *addr, nodes[index].Bootstrappers[0])
 		assert.True(t, nodes[index].Host.Network().Connectedness(addr.ID) == network.Connected)
 	}
 	addrFull := host.InfoFromHost(full.Host)
+	go func() {
+		for e := range sub.Out() {
+			connStatus := e.(event.EvtPeerConnectednessChanged)
+			if (connStatus.Connectedness == network.Connected ||
+				connStatus.Connectedness == network.NotConnected) && connStatus.Peer == full.Host.ID() {
+				ch <- struct{}{}
+			}
+		}
+	}()
+
 	select {
 	case <-ctx.Done():
 		t.Fatal("peer was not found")
 	case <-ch:
 		assert.True(t, light.Host.Network().Connectedness(addrFull.ID) == network.Connected)
 	}
+
 	require.NoError(t, full.Stop(ctx))
 	require.NoError(t, full.Host.Network().Close())
 	select {
--- a/node/tests/p2p_test.go
+++ b/node/tests/p2p_test.go
@@ -155,20 +155,26 @@ Test-Case: Restart full node discovery after one node is disconnected
 Steps:
 1. Create a Bridge Node(BN)
 2. Start a BN
-3. Create 4 full nodes with bridge node as bootstrapper peer
-4. Start 3 of full nodes
-5. Check that nodes are connected to each other
-6. Start the last FN
-7. Stop one of full node in order to restart discovery
-8. Check that disconnected node throws EvPeerDisconnected and connection status to it is NotConnected
-9. Check that the last FN is connected to one of the nodes
+3. Create 2 full nodes with bridge node as bootstrapper peer and start them
+4. Check that nodes are connected to each other
+5. Create one more node with disabled discovery
+6. Stop one of full node in order to restart discovery
+7. Check that the last FN is connected to one of the nodes
 *NOTE*: this test will take some time because it relies on several cycles of peer discovery
 */
 func TestRestartNodeDiscovery(t *testing.T) {
 	sw := swamp.NewSwamp(t)
 	cfg := node.DefaultConfig(node.Bridge)
 	cfg.P2P.Bootstrapper = true
-	bridge := sw.NewBridgeNode(node.WithConfig(cfg), node.WithRefreshRoutingTablePeriod(time.Second*30))
+	const defaultTimeInterval = time.Second * 10
+	var defaultOptions = []node.Option{
+		node.WithPeersLimit(2),
+		node.WithRefreshRoutingTablePeriod(defaultTimeInterval),
+		node.WithDiscoveryInterval(defaultTimeInterval),
+		node.WithAdvertiseInterval(defaultTimeInterval),
+	}
+	bridgeConfig := append([]node.Option{node.WithConfig(cfg)}, defaultOptions...)
+	bridge := sw.NewBridgeNode(bridgeConfig...)
 
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
 	t.Cleanup(cancel)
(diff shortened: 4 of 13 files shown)


node/node.go
```diff
@@ -133,16 +133,23 @@ func (n *Node) Run(ctx context.Context) error {
...
-	err := n.stop(ctx)
+		}
+	}()
+	err = n.stop(ctx)
 	if err != nil {
-		log.Errorf("Stopping %s Node: %s", n.Type, err)
 		return err
 	}
-
+	err = n.Host.Close()
```
Wondertan, line 149: We have lifecycles for components closage

vgonkivs: I think it does not work correctly in `host.go`, because `host.Close` is never called.

vgonkivs:
Test failed because of https://github.com/celestiaorg/celestia-node/issues/899

```
2022-07-11T14:26:37.2739578Z panic: send on closed channel
2022-07-11T14:26:37.2739749Z 	panic: close of closed channel
2022-07-11T14:26:37.2739776Z 
2022-07-11T14:26:37.2739920Z goroutine 84172 [running]:
2022-07-11T14:26:37.2740284Z github.com/libp2p/go-libp2p-pubsub.(*PubSub).processLoop.func1()
2022-07-11T14:26:37.2740907Z 	/home/runner/go/pkg/mod/github.com/celestiaorg/go-libp2p-pubsub@v0.6.2-0.20220628100036-657948473f1f/pubsub.go:535 +0x74
2022-07-11T14:26:37.2741114Z panic({0x1e3c060, 0x2daf760})
2022-07-11T14:26:37.2741384Z 	/opt/hostedtoolcache/go/1.18.3/x64/src/runtime/panic.go:838 +0x207
2022-07-11T14:26:37.2741785Z github.com/libp2p/go-libp2p-pubsub.(*PubSub).announce(0xc003747e00, {0x217239d, 0xa}, 0x0)
2022-07-11T14:26:37.2742410Z 	/home/runner/go/pkg/mod/github.com/celestiaorg/go-libp2p-pubsub@v0.6.2-0.20220628100036-657948473f1f/pubsub.go:889 +0x21f
2022-07-11T14:26:37.2742941Z github.com/libp2p/go-libp2p-pubsub.(*PubSub).handleRemoveSubscription(0xc003747e00, 0xc0133254f0)
2022-07-11T14:26:37.2743556Z 	/home/runner/go/pkg/mod/github.com/celestiaorg/go-libp2p-pubsub@v0.6.2-0.20220628100036-657948473f1f/pubsub.go:787 +0x12a
2022-07-11T14:26:37.2744068Z github.com/libp2p/go-libp2p-pubsub.(*PubSub).processLoop(0xc003747e00, {0x2dcf3d8, 0xc005e349c0})
2022-07-11T14:26:37.2744686Z 	/home/runner/go/pkg/mod/github.com/celestiaorg/go-libp2p-pubsub@v0.6.2-0.20220628100036-657948473f1f/pubsub.go:583 +0xa25
2022-07-11T14:26:37.2745087Z created by github.com/libp2p/go-libp2p-pubsub.NewPubSub
2022-07-11T14:26:37.2745698Z 	/home/runner/go/pkg/mod/github.com/celestiaorg/go-libp2p-pubsub@v0.6.2-0.20220628100036-657948473f1f/pubsub.go:323 +0xdea
2022-07-11T14:26:37.2746100Z FAIL	github.com/celestiaorg/celestia-node/node/tests	65.620s
2022-07-11T14:26:37.2746276Z FAIL
```

node/tests/reconstruct_test.go
```diff
@@ -102,30 +101,47 @@ func TestFullReconstructFromLights(t *testing.T) {
...
+		go func(i int) {
+			light := sw.NewLightNode(
+				node.WithBootstrappers([]peer.AddrInfo{*addr}),
+				node.WithRefreshRoutingTablePeriod(time.Second*30),
+			)
+			err = light.Start(ctx)
+			light.Host.Network().Notify(bundle)
+			require.NoError(t, err)
+			lights[i] = light
+		}(i)
```
Wondertan, line 133: nit: this is a good case for errgroup

node/tests/reconstruct_test.go
```diff
@@ -102,30 +101,47 @@ func TestFullReconstructFromLights(t *testing.T) {
...
-		trusted[i] = getMultiAddr(t, light.Host)
+		go func(i int) {
+			light := sw.NewLightNode(
+				node.WithBootstrappers([]peer.AddrInfo{*addr}),
+				node.WithRefreshRoutingTablePeriod(time.Second*30),
+			)
+			err = light.Start(ctx)
+			light.Host.Network().Notify(bundle)
+			require.NoError(t, err)
+			lights[i] = light
```
Wondertan, line 132: nit: why Notify is between Start and error handling for it

service/share/discovery.go
```diff
@@ -65,19 +70,32 @@ func (d *discovery) handlePeerFound(ctx context.Context, topic string, peer peer
 	d.host.ConnManager().TagPeer(peer.ID, topic, peerWeight)
 }
 
-// findPeers starts peer discovery every 30 seconds until peer cache will not reach peersLimit.
+// ensurePeers subscribes on Event Bus and starts peer discovery every 30 seconds
+// until peer cache will not reach peersLimit.
+// Discovery will be stopped once peers limit will be reached and will be restarted
+// when one of stored peer will be disconnected.
 // TODO (@vgonkivs): simplify when https://github.com/libp2p/go-libp2p/pull/1379 will be merged.
```
Wondertan, line 77: I believe it is not more relevant.

service/share/discovery.go
```diff
@@ -65,19 +70,32 @@ func (d *discovery) handlePeerFound(ctx context.Context, topic string, peer peer
 	d.host.ConnManager().TagPeer(peer.ID, topic, peerWeight)
 }
 
-// findPeers starts peer discovery every 30 seconds until peer cache will not reach peersLimit.
+// ensurePeers subscribes on Event Bus and starts peer discovery every 30 seconds
+// until peer cache will not reach peersLimit.
+// Discovery will be stopped once peers limit will be reached and will be restarted
+// when one of stored peer will be disconnected.
```
Wondertan, line 76:
```suggestion
// ensurePeers ensures we always have 'peerLimit' connected peers.
// It starts peer discovery every 30 seconds until peer cache reaches peersLimit. Discovery is restarted if any previously discovered peers disconnect. 
```

service/share/discovery.go
```diff
@@ -65,19 +70,32 @@ func (d *discovery) handlePeerFound(ctx context.Context, topic string, peer peer
...
+		return
+	}
 	t := time.NewTicker(interval * 3)
 	defer t.Stop()
 	for {
 		select {
 		case <-ctx.Done():
+			if err = sub.Close(); err != nil {
+				log.Warn(err)
+			}
```
Wondertan, line 92: Mmm, defer ?

service/share/discovery.go
```diff
@@ -87,13 +105,26 @@ func (d *discovery) findPeers(ctx context.Context) {
...
+			// reset timer in order to restart the discovery, once stored peer is disconnected
+			connStatus := e.(event.EvtPeerConnectednessChanged)
+			if connStatus.Connectedness == network.NotConnected {
+				if d.set.Contains(connStatus.Peer) {
+					d.connector.Reset(connStatus.Peer)
+					d.set.Remove(connStatus.Peer)
+					d.host.ConnManager().UntagPeer(connStatus.Peer, topic)
+					t.Reset(interval * 3)
+				}
+			}
```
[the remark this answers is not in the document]
Wondertan: Nested ifs is a bad practice is GO, especially when you can combine them into one

https://github.com/uber-go/guide/blob/master/style.md#reduce-nesting

service/share/discovery.go
```diff
@@ -87,13 +105,26 @@ func (d *discovery) findPeers(ctx context.Context) {
...
+		case e := <-sub.Out():
+			// listen to disconnect event to remove peer from set and reset backoff time
+			// reset timer in order to restart the discovery, once stored peer is disconnected
+			connStatus := e.(event.EvtPeerConnectednessChanged)
+			if connStatus.Connectedness == network.NotConnected {
+				if d.set.Contains(connStatus.Peer) {
+					d.connector.Reset(connStatus.Peer)
+					d.set.Remove(connStatus.Peer)
+					d.host.ConnManager().UntagPeer(connStatus.Peer, topic)
+					t.Reset(interval * 3)
```
Wondertan, line 117: Do we test that discovery is actually restarted?

service/share/backoff.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/libp2p/go-libp2p-core/host"
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/libp2p/go-libp2p/p2p/discovery/backoff"
+)
+
+// backoffConnector wraps a libp2p.Host to establish a connection with peers
+// with adding a delay for the next connection attempt.
+type backoffConnector struct {
+	h       host.Host
+	backoff backoff.BackoffStrategy
```
Wondertan, line 18: Each peer should have its own backoff strategy, currently, it is global for everyone.

service/share/backoff.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+	b.backoffCache[p.ID] = connectionCacheData{
+		time.Now().Add(b.backoff.Delay()),
+	}
+	b.cacheLK.Unlock()
+	return b.h.Connect(ctx, p)
+}
+
+// Reset removes a delay for the next connection attempt.
+func (b *backoffConnector) Reset(p peer.ID) {
```
Wondertan, line 56: Comment and func do not correspond to what is happening inside the method. Inside we add, not remove or reset

Wondertan (changes requested):
The above was mostly nit and stylistic changes. This is review is about required changes

service/share/backoff.go
vgonkivs: reworked

service/share/discovery.go
vgonkivs: Will try. But the problem is that test could take much time to finish.

node/settings.go
```diff
@@ -97,6 +97,7 @@ func WithBootstrappers(peers params.Bootstrappers) Option {
 }
 
 // WithRefreshRoutingTablePeriod sets custom refresh period for dht.
+// Currently, it is used to speed up tests.
```
Wondertan, line 100: I meant mentioning on the test itself, so that reader of the test can understand why that option is part of the test, but this is also fine

Wondertan: No action really needed, just FYI

service/share/backoff.go
```diff
@@ -0,0 +1,70 @@
...
+	b.cacheLK.Unlock()
+	return b.h.Connect(ctx, p)
+}
+
+// RestartBackOff resets delay time between attempts and add a delay for the next connection attempt to remote peer.
+// Mainly it will be call when host will receive a notification that remote peer was disconnected.
+func (b *backoffConnector) RestartBackOff(p peer.ID) {
+	b.cacheLK.Lock()
+	defer b.cacheLK.Unlock()
+	strategy := b.backoffCache[p].backoff
```
Wondertan, line 64: This will panic if peer does not present in the map

service/share/backoff.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// Connect puts peer to the backoffCache and tries to establish a connection with it.
+func (b *backoffConnector) Connect(ctx context.Context, p peer.AddrInfo) error {
+	b.cacheLK.Lock()
+	cacheData, ok := b.backoffCache[p.ID]
+	strategy := b.backoff()
+	if ok {
+		now := time.Now()
+		if now.Before(cacheData.nexttry) {
```
Wondertan, line 45: consistency nit: in other places time.Now() is inlined with Add, here it is extracted in a separate var and used only once. You can also consider still extracting it, while reusing in two places in the method

service/share/backoff.go
```diff
@@ -0,0 +1,70 @@
...
+		backoff:      factory,
+		backoffCache: make(map[peer.ID]connectionCacheData),
+	}
+}
+
+// Connect puts peer to the backoffCache and tries to establish a connection with it.
+func (b *backoffConnector) Connect(ctx context.Context, p peer.AddrInfo) error {
+	b.cacheLK.Lock()
+	cacheData, ok := b.backoffCache[p.ID]
+	strategy := b.backoff()
```
Wondertan, line 42: nit: don't really understand the grouping here and why ok check is not close to its definition/read from map

service/share/backoff.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/libp2p/go-libp2p/p2p/discovery/backoff"
+)
+
+// backoffConnector wraps a libp2p.Host to establish a connection with peers
+// with adding a delay for the next connection attempt.
+type backoffConnector struct {
+	h       host.Host
+	backoff backoff.BackoffFactory
+
+	cacheLK      sync.Mutex
```
Wondertan, line 20: nit: Lk?

Wondertan (dismissed):
The logic is now fully correct from what I see, besides one thing. Also left some nits

Wondertan:
Another thing we will need to have here is connection throttling, but we can keep this for later or implement something simple over the buffered channel

service/share/backoff.go
vgonkivs: Yes, I made wrong assumption. Will fix it.

vgonkivs: I was added this way to be able to assign it in one place after `if ok` condition for both cases.

service/share/discovery.go
```diff
@@ -87,13 +106,25 @@ func (d *discovery) findPeers(ctx context.Context) {
 			for peer := range peers {
 				go d.handlePeerFound(ctx, topic, peer)
 			}
+		case e := <-sub.Out():
+			// listen to disconnect event to remove peer from set and reset backoff time
+			// reset timer in order to restart the discovery, once stored peer is disconnected
+			connStatus := e.(event.EvtPeerConnectednessChanged)
```
renaynay, line 112: could *technically* cause panic

```suggestion
			connStatus, ok = e.(event.EvtPeerConnectednessChanged)
			if !ok {
			 // err and return
			}
```

service/share/discovery.go
```diff
@@ -87,13 +106,25 @@ func (d *discovery) findPeers(ctx context.Context) {
...
+			if connStatus.Connectedness != network.NotConnected || !d.set.Contains(connStatus.Peer) {
+				continue
+			}
+			d.connector.RestartBackOff(connStatus.Peer)
+			d.set.Remove(connStatus.Peer)
+			d.host.ConnManager().UntagPeer(connStatus.Peer, topic)
+			t.Reset(interval * 3)
```
renaynay, line 119: maybe extract this in something called `defaultBackoffInterval` or something ? since you use this val twice

renaynay (approved):
small nits, otherwise good :)

service/share/discovery.go
Wondertan: It might look like it does, but it fact that channel always returns a type to which it was subscribed

logs/logs.go
```diff
@@ -14,6 +14,7 @@ func SetAllLoggers(level logging.LogLevel) {
 	_ = logging.SetLogLevel("connmgr", "WARN")
 	_ = logging.SetLogLevel("nat", "INFO")
 	_ = logging.SetLogLevel("dht/RtRefreshManager", "FATAL")
+	_ = logging.SetLogLevel("bitswap_network", "ERROR")
```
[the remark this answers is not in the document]
vgonkivs: it throws a lot of unneeded logs during p2p testing. So I think it can be here.

service/share/discovery.go
vgonkivs: if we subscribes on` event.EvtPeerConnectednessChanged` then only this event is expected

service/share/backoff.go
```diff
@@ -0,0 +1,72 @@
...
+		h:            h,
+		backoff:      factory,
+		backoffCache: make(map[peer.ID]connectionCacheData),
+	}
+}
+
+// Connect puts peer to the backoffCache and tries to establish a connection with it.
+func (b *backoffConnector) Connect(ctx context.Context, p peer.AddrInfo) error {
+	strategy := b.backoff()
+	b.cacheLk.Lock()
```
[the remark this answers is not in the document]
vgonkivs: Because then we will store data in map. So, in this case I think it's better to make one common `Lock()` for the whole method. it does not take much time because it does not have any blocking operation

node/tests/p2p_test.go
```diff
@@ -109,25 +111,133 @@ func TestBootstrapNodesFromBridgeNode(t *testing.T) {
...
+		select {
+		case <-ctx.Done():
+			t.Fatal("peer was not found")
+		case p := <-ch:
+			assert.True(t, full1.Host.Network().Connectedness(p) == network.Connected)
+
+		}
+	}
+	require.NoError(t, full4.Start(ctx))
+	time.Sleep(time.Second * 30)
```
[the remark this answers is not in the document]
vgonkivs: good point. Will think about it

service/share/backoff.go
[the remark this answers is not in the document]
Wondertan: A general rule of thumb would be: introduce rwmutex only and only if there are multiple readers

service/share/backoff.go
```diff
@@ -0,0 +1,72 @@
...
+	backoff backoff.BackoffFactory
+
+	cacheLk   sync.Mutex
+	cacheData map[peer.ID]connectionCacheData
+}
+
+// connectionCacheData stores time when next connection attempt with the remote peer.
+type connectionCacheData struct {
+	nexttry time.Time
+	backoff backoff.BackoffStrategy
```
Wondertan, line 27: So it turns out this map is never really cleaned up, which is technically a leak.

service/share/backoff.go
```diff
@@ -0,0 +1,72 @@
...
+func (b *backoffConnector) RestartBackoff(p peer.ID) {
+	b.cacheLk.Lock()
+	defer b.cacheLk.Unlock()
+	cache, ok := b.cacheData[p]
+	if !ok {
+		return
+	}
+	cache.backoff.Reset()
+	b.cacheData[p] = connectionCacheData{
+		time.Now().Add(cache.backoff.Delay()),
```
Wondertan, line 69: Instead of allocating a new `connectionCacheData`, we can change the `nexttry` field to be directly modified on `cache` var, like you with `Reset`(loc:67) We might also use ptr instead of value in the map to fix subsequent issues

service/share/discovery.go
```diff
@@ -39,13 +44,14 @@ func newDiscovery(h host.Host, d core.Discovery) *discovery {
 		newLimitedSet(peersLimit),
 		h,
 		d,
+		newBackoffConnector(h, backoff.NewFixedBackoff(time.Hour)),
```
Wondertan, line 47: let's move to the consts above as well

service/share/backoff.go
Wondertan: I honestly don't see how we can fix this without periodic GCing

vgonkivs: Yes, we can resolve it only by adding a periodic GCing.

service/share/discovery.go
renaynay: Yes I understand this, which is why I said "technically". 
However, we have other parts of the codebase that also follow the same pattern, for example https://github.com/celestiaorg/celestia-node/blob/main/core/fetcher.go#L147

In this case too, it is not *technically* possible to receive an event of a diff type via that subscription but we still perform the check

We should be consistent here too.

service/share/backoff.go
```diff
@@ -11,27 +11,33 @@ import (
 	"github.com/libp2p/go-libp2p/p2p/discovery/backoff"
 )
 
+// defaultTimeToLive is a default period after which disconnected peers will be removed from cache
+const defaultTimeToLive = time.Hour
```
renaynay, line 15: can just do `defaultTTL`

renaynay (commented):
Nice solution, what's the reason for changing to a pointer though?

service/share/backoff.go
vgonkivs: sure

vgonkivs:
> Nice solution, what's the reason for changing to a pointer though?

To be able to modify data in place without allocating memory for a new struct.

service/share/discovery.go
Wondertan: I would prefer to be consistent with things that make sense, like here. We can make GFI to remove the if in the fetcher.

service/share/backoff.go
Wondertan: Well, calling this TTL or TimeToLive is not correct. TTLs are attached to a specific entity/object with some living time, while here, we have a periodic duration for GC.

service/share/backoff.go
```diff
@@ -0,0 +1,99 @@
...
+			b.cacheLk.Unlock()
+			return fmt.Errorf("share/discovery: backoff period is not ended for peer=%s", p.ID.String())
+		}
+		strategy = b.cacheData[p.ID].backoff
+	}
+	cacheData = &connectionCacheData{
+		nexttry: time.Now().Add(strategy.Delay()),
+		backoff: strategy,
+	}
+	b.cacheData[p.ID] = cacheData
```
Wondertan, line 59: Avoid reallocating the cachedata here for the case where it is already on the map.

Wondertan: Not fixed yet

service/share/backoff.go
```diff
@@ -0,0 +1,99 @@
...
+			b.cacheLk.Unlock()
+			return fmt.Errorf("share/discovery: backoff period is not ended for peer=%s", p.ID.String())
+		}
+		strategy = b.cacheData[p.ID].backoff
+	} else {
+		cache = &connectionCacheData{}
+		b.cacheData[p.ID] = cache
+	}
+	cache.nexttry = time.Now().Add(strategy.Delay())
+	cache.backoff = strategy
```
Wondertan, line 59: Previously, in the ok case, we were taking current `b.cacheData[p.ID].backoff` and set it to the `strategy `, to then set it back again to the cache on the loc:59, so the code wasn't doing anything useful, and the suggestion removes it. Also, I moved the !ok case close to the map read.
```suggestion
	b.cacheLk.Lock()
	cache, ok := b.cacheData[p.ID]
	if !ok {
		cache = &connectionCacheData{}
		cache.backoff = b.backoff()
		b.cacheData[p.ID] = cache
	} else {
		if time.Now().Before(cache.nexttry) {
			b.cacheLk.Unlock()
		        return fmt.Errorf("share/discovery: backoff period is not ended for peer=%s", p.ID.String())
		}
	}
	
	cache.nexttry = time.Now().Add(cache.backoff.Delay())
```

service/share/backoff.go
```diff
@@ -0,0 +1,99 @@
...
+		backoff := b.backoff()
+		cache = &connectionCacheData{
+			nexttry: time.Now().Add(backoff.Delay()),
+			backoff: backoff,
+		}
+		b.cacheData[p] = cache
+		return
+	}
+	cache.backoff.Reset()
+	cache.nexttry = time.Now().Add(cache.backoff.Delay())
```
Wondertan, line 80: Now you can see how similar the code from the above and here are and we might do something about it
```suggestion
	cache, ok := b.cacheData[p]
	if !ok {
		cache = &connectionCacheData{}
		cache.backoff = b.backoff()
		b.cacheData[p] = cache
		return
	}
	cache.backoff.Reset()
	cache.nexttry = time.Now().Add(cache.backoff.Delay())
```

service/share/discovery.go
```diff
@@ -101,13 +121,30 @@ func (d *discovery) findPeers(ctx context.Context) {
...
+					d.connector.RestartBackoff(connStatus.Peer)
+					d.set.Remove(connStatus.Peer)
+					d.host.ConnManager().UntagPeer(connStatus.Peer, topic)
+					t.Reset(d.discoveryInterval)
+				}
+			}
+			d.connector.RestartBackoff(connStatus.Peer)
+			d.set.Remove(connStatus.Peer)
+			d.host.ConnManager().UntagPeer(connStatus.Peer, topic)
+			t.Reset(d.discoveryInterval)
```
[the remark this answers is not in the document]
Wondertan: Have the same concern

node/tests/p2p_test.go
```diff
@@ -112,25 +114,114 @@ func TestBootstrapNodesFromBridgeNode(t *testing.T) {
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
+	t.Cleanup(cancel)
+
+	err := bridge.Start(ctx)
+	require.NoError(t, err)
+	addr := host.InfoFromHost(bridge.Host)
+	nodes := make([]*node.Node, 2)
+	nodesConfig := append([]node.Option{node.WithBootstrappers([]peer.AddrInfo{*addr})}, defaultOptions...)
+	for index := 0; index < 2; index++ {
```
Wondertan, line 187: Extracting this 2 together with peerlimit opt 2 into one var would help readability

service/share/discovery.go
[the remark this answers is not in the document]
vgonkivs: Artifacts after resolving merge conflicts. Will remove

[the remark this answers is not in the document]
vgonkivs: So, the flow here must be: if peer is disconnected, then check that this peer was stored in our cache and perform all necessary actions to it:
1) restart backoff
2) remove from cache;
3) untag;
4) restart discovery timer;

service/share/backoff.go
vgonkivs: reworked

service/share/backoff.go
vgonkivs: reworked

node/tests/p2p_test.go
```diff
@@ -112,25 +114,114 @@ func TestBootstrapNodesFromBridgeNode(t *testing.T) {
...
 		require.NoError(t, nodes[index].Start(ctx))
 		assert.Equal(t, *addr, nodes[index].Bootstrappers[0])
 		assert.True(t, nodes[index].Host.Network().Connectedness(addr.ID) == network.Connected)
 	}
+	addrFull := host.InfoFromHost(full.Host)
+	go func() {
+		for e := range sub.Out() {
+			connStatus := e.(event.EvtPeerConnectednessChanged)
+			if (connStatus.Connectedness == network.Connected ||
+				connStatus.Connectedness == network.NotConnected) && connStatus.Peer == full.Host.ID() {
```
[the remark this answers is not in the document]
vgonkivs: removed

Wondertan (approved):
Now, the logic and the code are excellent! Left only one small naming nit. Also, leaving a review of comments to @renaynay

service/share/backoff.go
```diff
@@ -0,0 +1,96 @@
...
+type backoffConnector struct {
+	h       host.Host
+	backoff backoff.BackoffFactory
+
+	cacheLk   sync.Mutex
+	cacheData map[peer.ID]*connectionCacheData
+}
+
+// connectionCacheData stores time when next connection attempt with the remote peer.
+type connectionCacheData struct {
```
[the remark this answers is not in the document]
vgonkivs: Let's do it

renaynay (approved):
Nice! Everything looking good to me.

Issue #885, opened 2022-07-05
Auto generate binaries for different platforms on releases


Wondertan:

iofq:
What platforms would we target? amd64, arm64, windows, mac?

Currently we don't tag any docker builds besides sha commit tags, would `latest` and semver tags be appropriate on release?

What would the preferred release workflow be? Create a release and then upload assets during the workflow? Or create a semver tag, have the workflow create a release draft with assets attached, and then finalize that release manually?

renaynay:
> What platforms would we target?

amd64, arm64 and mac are higher prio -- windows would also be cool :) 

> Currently we don't tag any docker builds besides sha commit tags, would latest and semver tags be appropriate on release?

Yes ( cc @sysrex ) 

> What would the preferred release workflow be? Create a release and then upload assets during the workflow? Or create a semver tag, have the workflow create a release draft with assets attached, and then finalize that release manually?

@sysrex @Wondertan do you guys have any thoughts? I think we should go with the latter?

iofq:
Following up on this:

For windows, the [fslock package](https://github.com/celestiaorg/celestia-node/tree/main/libs/fslock) doesn't appear to support windows yet, which would need to happen first.

For mac/darwin, a dependancy of libp2p, [elastic/gosigar](url), doesn't want to build under `GOOS=darwin` for me. But I may be missing something here.

Wondertan:
@iofq, true, I just tried me to cross-compile for darwin and saw the same issue with gosigar. This is weird because I know some of our folks are using mac with no issues

Wondertan:
Ok so this is definitely related: https://github.com/elastic/go-sysinfo/issues/127

renaynay:
I've run into issues with gosigar before and basically had to fork and maintain a fork of gosigar in order to prevent this issue. It was a long time ago though.

renaynay:
https://github.com/gballet/gosigar

go-ethereum used to indirectly depend on it I think bc of this dep: https://github.com/olebedev/go-duktape which has since been deprecated.

ramin:
closing this issue as it was taken care of by https://github.com/celestiaorg/celestia-node/pull/2661. once we have ci/release flakey tests passing and the entire release process continues we'll have the binaries

PR #894, opened 2022-07-06
Add `concurrency` block to cancel previous CI runs

### LINKED ISSUE #889: Use concurrency to cancel previous CI runs
Ref: https://github.com/FuelLabs/fuels-ts/pull/361

iofq:
Resolves #889

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -11,6 +11,10 @@ env:
   IMAGE_NAME: ${{ github.repository }}
   REGISTRY: ghcr.io
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
   docker-build:
     runs-on: "ubuntu-latest"
--- a/.github/workflows/go-ci.yml
+++ b/.github/workflows/go-ci.yml
@@ -11,6 +11,10 @@ on:
 env:
   GO_VERSION: 1.18
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
   lint:
     name: Lint
--- a/.github/workflows/labels.yml
+++ b/.github/workflows/labels.yml
@@ -3,6 +3,11 @@ name: Required Labels
 on:
   pull_request:
     types: [opened, labeled, unlabeled, synchronize]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
   label:
     runs-on: ubuntu-latest


.github/workflows/docker-build.yml
```diff
@@ -11,6 +11,10 @@ env:
   IMAGE_NAME: ${{ github.repository }}
   REGISTRY: ghcr.io
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
adlerjohn, line 15: So, one issue with this setup is that it cancels CI on the default branch if a new PR is merged before CI completes. Cancelling CI in a PR is fine because it's in active development. But for `main` we actually want CI to run to completion on every merge, especially for CI with artifacts like this one.

Is there some way to make this cancel CI only for PRs?

iofq: Good points. With that in mind, this is what I propose:

**Go CI & labels:** Based on the [docs](https://docs.github.com/en/actions/using-jobs/using-concurrency), the suggested way to run on PRs only is `${{ github.head_ref || github.run_id }}`; `head_ref` will only exist on PRs, and if it doesn't exist, concurrency will collate on `run_id`, which is unique for each workflow run, meaning workflows won't be cancelled for non-PR related triggers.

**docker-build:** Barring any billing concerns, I see no harm in letting docker-builds complete and push sha tags to the container store. `${{ github.workflow }}-${{ github.sha }}` means that `docker-build` won't be cancelled unless two runs are triggered on the same git commit. 

I could see this theoretically occurring if a branch is pushed to and then quickly merged into `main`, but we're only tagging by sha anyway, so in any case the tagged artifact will still get created. This could potentially cause problems with #885, if multiple docker-builds trigger on release, targeting different tags, and race or something. An alternative would be to just skip concurrency on `docker-build` entirely and let all builds complete.

jbowen93: No billing concerns on docker-build. It's all open source so GH makes it free. I'm in favor of this and the node team can report back and tell us if the new workflow is more convenient.

iofq: @Wondertan @adlerjohn I believe this thread is resolved. Let me explain:

Github lets us define a concurrency tag format for each workflow. When a workflow gets queued, if it has the same tag as another running workflow, the older workflow gets cancelled (if we have `cancel-in-progress` set). For our purposes, the docs recommend the following tag:

---

`group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}`
* Example tags: `go-ci-feature_branch` if the workflow is from a pull request, or `go-ci-<unique number>` if not.
* Because the tags for a feature branch PR are always the same, new `go-ci`, `docker-build` and `labels` workflows in a branch will cancel any currently running workflows in that branch. If we're NOT in a feature branch (`main`), `head_ref` is not set, and the unique `run_id` will instead ensure that all workflows complete. 

This is essentially the same as what I described in my previous post, but I made `docker-build` use the same cancelling behavior as the other workflows.

Wondertan: Ok, thanks for explaining.

Wondertan (approved):

adlerjohn (approved):

.github/workflows/docker-build.yml
adlerjohn: lgtm

adlerjohn:
Thanks for working through this @iofq! Good work!

PR #904, opened 2022-07-12
fix: add condition to restore routing table refresh period

### LINKED ISSUE #897: p2p: add condition to check RoutingtableRefreshPeriod
Currently, default value for [RoutingTableRefreshPeriod](https://github.com/celestiaorg/celestia-node/blob/3f2d74340eb7891dd859a493ce636cb989d825c3/node/p2p/p2p.go#L53) is 1 minute. But there is an [option](https://github.com/celestiaorg/celestia-node/blob/3f2d74340eb7891dd859a493ce636cb989d825c3/node/settings.go#L100) that could override default value. We should add a condition that will check overriding value. It have to be more the 0 or we will get a panic.

vgonkivs:
Resolves #897

### DIFF
--- a/node/p2p/p2p.go
+++ b/node/p2p/p2p.go
@@ -9,6 +9,8 @@ import (
 	"go.uber.org/fx"
 )
 
+const defaultRoutingRefreshPeriod = time.Minute
+
 // Config combines all configuration fields for P2P subsystem.
 type Config struct {
 	// ListenAddresses - Addresses to listen to on local NIC.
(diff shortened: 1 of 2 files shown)


renaynay (approved):
Tiny nits but LGTM !

renaynay:
Tested with an "old" config and works as expected ✅

liamsi (approved):

Wondertan (approved):

PR #907, opened 2022-07-13
feat(cmd): initial integration of OpenTelemetry with OTLP exporter

### LINKED ISSUE #934: cmd: OTLP TLS connection
Currently, Otel data is transferred with OTLP over a plain HTTP connection, and instead, we should use TLS for data protection reasons.

To do this, we need to stop using [WithInsecure](https://github.com/open-telemetry/opentelemetry-go/blob/main/exporters/otlp/otlpmetric/otlpmetrichttp/options.go#L161) option for both metrics and tracing. Next, we should consider if using the default TLS configuration works for us and if not use [WithTLSClientConfig](https://github.com/open-telemetry/opentelemetry-go/blob/main/exporters/otlp/otlpmetric/otlpmetrichttp/options.go#L155).

Wondertan:
* Initial integration of OpenTelemetry with ~~Jaeger~~OTLP exporter
* Adds new flags switches for both tracing and metrics with endpoint customization
	* `--tracing` and `--tracing.endpoint`
	* `--metrics` and `--metrics.endpoint`
* Simple tracing coverage for ipld.Retriever
* Simple metric for current store height

P.S. OTLP integration could be separated from introduced metrics and tracing coverage, but it is the bare minimum needed to manually verify that the integration works as expected by looking at actual data reported by the local nodes.

TODO:
- [x] Ensure different node types advertise as different services
- [x] Use OTLP instead of Jaeger
- [x] Integrate OTLP for metrics
- [x] Store height metric

As always, recommend looking at the PR commit by commit(CBC)

Substitutes #810
Closes #934


cmd/env.go
```diff
@@ -12,6 +12,9 @@ import (
 // 1. Propagate values from parent to child commands.
 // 2. To group common logic that multiple commands rely on.
 // Usage can be extended.
+// TODO(@Wondertan): We should move to using context only instead.
```
renaynay, line 15: issue?

cmd/flags_misc.go
```diff
@@ -87,5 +134,72 @@ func ParseMiscFlags(cmd *cobra.Command) error {
...
+	ok, err = cmd.Flags().GetBool(tracingFlag)
+	if err != nil {
+		return err
+	}
+
+	if ok {
+		exp, err := otlptracehttp.New(cmd.Context(),
+			otlptracehttp.WithEndpoint(cmd.Flag(tracingEndpointFlag).Value.String()),
+			otlptracehttp.WithCompression(otlptracehttp.GzipCompression),
+			otlptracehttp.WithInsecure(),
```
renaynay, line 147: Is this required?

ipld/retriever.go
```diff
@@ -239,6 +274,10 @@ func (rs *retrievalSession) doRequest(ctx context.Context, q *quadrant) {
 			// get the root node
 			nd, err := plugin.GetNode(ctx, rs.bget, root)
 			if err != nil {
+				rs.span.RecordError(err, trace.WithAttributes(
+					attribute.String("requesting-root", root.String()),
+					attribute.Int("root-id", i),
```
renaynay, line 279:
```suggestion
					attribute.Int("root-index", i),
```
maybe a bit more descriptive?

Wondertan:
@rootulp, you can try running collector locally with default endpoints/port while passing to Light or Full node new flags described on the PR desc. And you should see the height of the node being reported, etc. To see traces you would need to run Full node

renaynay (approved):
Looks good! It would be nice to get a demo for this some time soon. One question, tracing happens regardless of whether tracing is enabled, correct? `--tracing` being enabled just adds the default (or custom) exporter endpoint. So if `--tracing` isn't passed, but tracing still occurs, does it impact performance?

cmd/flags_misc.go
Wondertan: Insecure or all options you point to?

ipld/retriever.go
Wondertan: Not sure

Wondertan:
> One question, tracing happens regardless of whether tracing is enabled, correct? --tracing being enabled just adds the default (or custom) exporter endpoint. So if --tracing isn't passed, but tracing still occurs, does it impact performance?

That's a good question, and I am glad you asked it. I had the same concern when I even looked at the implementation. My conclusion is that it won't noticeably affect performance. The default tracing exporter does nothing besides keeping information about spans to be delegated if tracing is turned on in runtime(see #937).

The same is true for metrics. The default exporter does nothing besides logic to delegate to some custom exporter that can be enabled in runtime. I could potentially make `MonitorHead` always called without adding any options for the node to enable it, similar to tracing, but I am still unsure about this.

cmd/flags_misc.go
Wondertan: Do you ask about the Insecure option or all the pointed options?

rootulp:
Screen Shot 2022-07-22 at 10 50 14 PM

Screenshot of a Grafana dashboard featuring the head metric (implemented in this PR). This is for a light node I'm running on AWS which is emitting metrics to an OTEL Collector deployed on the same instance. The OTEL Collector is exporting metrics to Prometheus that is hosted by Grafana Cloud. Setup instructions are in https://github.com/celestiaorg/celestia-node/pull/922

tagging @renaynay because I'm happy to hop on a call and demo this or set it up for you

cmd/flags_misc.go
renaynay: Insecure option

ipld/retriever.go
renaynay: `id` i feel like corresponds to hash, and `i` is index

renaynay:
@rootulp would love to do a demo on how to set it up, thanks!

header/metrics.go
```diff
@@ -0,0 +1,46 @@
...
+				headC.Observe(ctx, 0, attribute.String("err", err.Error()))
+				return
+			}
+
+			headC.Observe(
+				ctx,
+				head.Height,
+				attribute.String("hash", head.Hash().String()),
+				attribute.String("app_hash", head.AppHash.String()),
+				attribute.String("data_root", head.DAH.String()),
```
rootulp, line 38: Apologies for raising this concern late but I'm concerned these labels will spike our [metric label cardinality](https://grafana.com/blog/2022/02/15/what-are-cardinality-spikes-and-why-do-they-matter/). These three fields will be unique per block so we will be generating many series. 

Proposal to remove labels that would be unique per block (i.e. these three). `square_size` seems safe to preserve as a label because the # of possible values is constrained.

Screen Shot 2022-07-26 at 4 32 27 PM

cmd/flags_misc.go
rootulp: Linking https://github.com/celestiaorg/celestia-node/issues/934

Wondertan: Yes, it is necessary for a simple setup like we have.

Wondertan: If we don't provide that, nothing works until we don't provide custom TLSConfig

cmd/env.go
Wondertan: #965

header/metrics.go
Wondertan: Made a deep look into the question
* The cardinality issue seems related to the Prometheus inefficiencies and lack of scalability, and the solution that Prometheus provides is to use fewer series:
> When you instrument your code and add a new metric, sometimes you attach more context than you need. 
*
	* It's like if Ethereum would say: "you transfer more value than you need" instead of putting efforts into scaling. And how do they know that we attach more than we need? I believe this is a common response from huge well-established status quo overmanaged enterprise companies on performance issues with their tech. They don't wanna change their legacy code, as they are too lazy and clumsy.

* In our case, this data is sent per block every block time(~30 secs in Mamaki). Data spikes could happen if there is an influx of users/node operators. 
	* In case of the further incentivized testnet the number of participants fixed? So no spikes?
		*  We still don't know if we will continue using this setup for the mainnet.
	* Also, our spikes would be slightly different than the case described in the article, AFAIU. 
		* In their case, there is _one_ microservice sending series but has a spike of new users, while in our, _n_ nodes send the same data with the same sending data size and time patterns.

* The data we send here _is useful for local monitoring setup_. but is not useful and just duplicates for global network view.
	* Should we do early optimizations to split the code into two types of monitoring(one configuration for local and another for global view)? 
		* We still don't know if such spikes will be a real issue according to our unique patterns, so I don't think we have to spend time splitting these and forcing devs to think about what should be sent to global view or not. 
		* If performance becomes an issue, we can consider using a more scalable drop-in replacement for Prometheus, like VictoriaMetrics(should we do it from the beginning?)
		* We can also investigate if Solana had any performance issues with Prometheus, as their metrics patterns might be similar to ours.

rootulp: I expect other monitoring databases may have better performance with regards to high cardinality data but they still advocate avoiding it:

> CAUTION: Remember that every unique combination of key-value label pairs represents a new time series, which can dramatically increase the amount of data stored. Do not use labels to store dimensions with high cardinality (many different label values), such as user IDs, email addresses, or other unbounded sets of values.

- https://prometheus.io/docs/practices/naming/#labels

> High cardinality usually means a high number of [active time series](https://docs.victoriametrics.com/FAQ.html#what-is-an-active-time-series). High cardinality may lead to high memory usage and/or to a high percentage of [slow inserts](https://docs.victoriametrics.com/FAQ.html#what-is-a-slow-insert). The source of high cardinality is usually a label with a large number of unique values, which presents a big share of the ingested time series. The solution is to identify and remove the source of high cardinality with the help of [cardinality explorer](https://docs.victoriametrics.com/#cardinality-explorer).
- https://docs.victoriametrics.com/FAQ.html#what-is-high-cardinality

Are you proposing using an alternative to Prometheus or VictoriaMetrics? If so, I think this should be evaluated in one of our ADRs.

rootulp: For clarification: my proposal is to avoid high cardinality data by removing these labels.

If you feel strongly that we should preserve these labels then we can continue down the path of exploring alternative DBs that better support high cardinality data. I can explore other databases in ADR-010 because it already considers Prometheus and InfluxDB

Wondertan: My main point above is that we need these values for the local setup. The primary use cases I assumed for that are local debugging and the ability for a node operator to look up some data related to the height quickly. Later on, I realized that the former should be done with tracing instead, and the latter with some form of network explorer/indexer specialized in storing all the data about blockchain. Putting all that pressure on metrics is unnecessary, so let's remove those. 

We can still consider using other solutions to Prometheus, but I don't feel strongly about that. If you see the need and potential overload of the network, then let's add this to ADR; if not, let's not come back to the question until there is a performance issue.

Wondertan: Removed

rootulp: > The primary use cases I assumed for that are local debugging and the ability for a node operator to look up some data related to the height quickly.

One thought is that celestia-node still emits logs that contain block height and hash, ex.
```
2022-07-28T23:28:54.109Z	INFO	das	das/daser.go:282	sampled header	{"height": 153029, "hash": "7757A6368C9BE0833B644A1873DAD68FD93DA98BFFFE1C86FD1BC821CF3842AF", "square width": 16, "finished (s)": 0.482185124}
```

so for debugging locally, an operator or celestia-node developer may be able to get comparable information from logs rather than metrics. Defer to you on how best to capture traces or logs with this data.

Metric labels with a large # of possible values seems like a non-obvious foot gun that we should try to avoid. I'm open to suggestions on how to enforce this in the codebase. We can monitor cardinality (see [docs](https://grafana.com/docs/grafana-cloud/data-configuration/metrics/metrics-control-usage/control-prometheus-metrics-usage/cardinality-management/?pg=blog&plcmt=body-txt)) however this seems less helpful than catching and preventing implementation of metrics with high cardinality.

rootulp (approved):

header/metrics.go
Wondertan: As long I am working on celestia-node, I will take care of this and other devs who joined and learned this will follow along.

cmd/flags_misc.go
rootulp: Given it is possible to configure celestia-node to send metrics to an OTEL Collector over HTTPS and that this will be the convention during the incentivized testnet, I think we should remove this line

```suggestion

```

In a subsequent issue we can add a CLI flag for `--insecure-metrics` and `--insecure-tracing` that enables the `WithInsecure()` option for node operators who want to host their own OTEL Collector and don't want to configure TLS certificates on their OTEL Collector. What do you think?

Wondertan: I've implemented tls option for both metrics and tracing and removed Insecure option being the default, so resolving

rootulp (approved):

cmd/flags_misc.go
```diff
@@ -169,11 +186,15 @@ func ParseMiscFlags(cmd *cobra.Command, env *Env) error {
 	}
 
 	if ok {
-		exp, err := otlpmetrichttp.New(cmd.Context(),
-			otlpmetrichttp.WithEndpoint(cmd.Flag(metricsEndpointFlag).Value.String()),
+		opts := []otlpmetrichttp.Option{
 			otlpmetrichttp.WithCompression(otlpmetrichttp.GzipCompression),
-			otlpmetrichttp.WithInsecure(),
-		)
+			otlpmetrichttp.WithEndpoint(cmd.Flag(tracingEndpointFlag).Value.String()),
```
renaynay, line 193: metricsEndpointFlag

cmd/flags_misc.go
```diff
@@ -169,11 +186,15 @@ func ParseMiscFlags(cmd *cobra.Command, env *Env) error {
...
+		}
+		if ok, _ := cmd.Flags().GetBool(tracingTlS); !ok {
```
renaynay: metricsTls

cmd/flags_misc.go
```diff
@@ -72,6 +73,12 @@ and their lower-case forms`,
 		"Sets HTTP endpoint for OTLP traces to be exported to. Depends on '--tracing'",
 	)
 
+	flags.Bool(
+		tracingTlS,
```
renaynay, line 77: `tracingTls` or `tracingTLS`

cmd/flags_misc.go
```diff
@@ -84,6 +91,12 @@ and their lower-case forms`,
 		"Sets HTTP endpoint for OTLP metrics to be exported to. Depends on '--metrics'",
 	)
 
+	flags.Bool(
+		metricsTlS,
```
renaynay, line 95: same here, `metricsTls` or `metricsTLS`

renaynay (changes requested):

cmd/flags_misc.go
```diff
@@ -182,15 +184,17 @@ func ParseMiscFlags(cmd *cobra.Command, env *Env) error {
...
+		panic(err)
 	}
 
 	if ok {
 		opts := []otlpmetrichttp.Option{
 			otlpmetrichttp.WithCompression(otlpmetrichttp.GzipCompression),
 			otlpmetrichttp.WithEndpoint(cmd.Flag(tracingEndpointFlag).Value.String()),
 		}
-		if ok, _ := cmd.Flags().GetBool(tracingTlS); !ok {
+		if ok, err := cmd.Flags().GetBool(tracingTlS); err != nil {
```
renaynay, line 195: `metricsTls`

Wondertan: Oh, good catch. Thank you!!!

cmd/flags_misc.go
Wondertan: This line is fine

cmd/flags_misc.go
Wondertan: Copy of https://github.com/celestiaorg/celestia-node/pull/907#discussion_r935371791

renaynay (approved):
👍🏻

PR #910, opened 2022-07-15
cmd | node: Refactor core flags

renaynay:
## Description
This PR consolidates flags related to adding a core endpoint. The reasoning behind this is to prevent users from having to pass too many flags once #571 is resolved (since the core accessor will now need both a `gRPC` and `RPC` endpoint for the celestia-core connection).

## New flags

**In the case that the default ports are exposed (grpc:9090, rpc:26657) on the celestia-core endpoint**:
```
celestia <node_type> start --core <ip addr of core endpoint>
```
**In the case that custom ports are exposed on the celestia-core endpoint**:
```
celestia <node_type> start --core <ip addr of core endpoint> --core.grpc 9090 --core.rpc 26657
```

If `--core.grpc` || `--core.rpc` are not provided, we assume the defaults and attempt to connect.

Issue #570, opened 2022-03-30
Add unit test for `CoreAccess`


renaynay:
**TODO** link to tracking issue for mock celestia-app blocker

Wondertan:
We now have App test suit, so we can cover all state endpoints now. Also, they will be covered by #2337

PR #911, opened 2022-07-15
service/state: Implement balance verification against `AppHash`

### LINKED ISSUE #571: Implement balance verification
Implement balance verification as described [here](https://github.com/celestiaorg/celestia-node/pull/513).

TODO: 

- [ ] #911 
- [ ]  #

renaynay:
Tl;dr, instead of requesting balance from core endpoint, the node now requests proofs as well and, using the proofs, verifies the returned balance against the `AppHash` of the corresponding block.

TODO: 
- [x] Implementation
- [x] ADR to document use of ABCI query and proofs

Resolves #571.

Based on #910 , will rebase and mark as ready for review once merged.

### DIFF
--- a/docs/adr/adr-004-state-interaction.md
+++ b/docs/adr/adr-004-state-interaction.md
@@ -69,29 +69,29 @@ type StateAccessor interface {
 
 ### Verification of Balances
 In order to check that the balances returned via the `AccountBalance` query are correct, it is necessary to also request
-Merkle proofs from celestia-app and verify them against the latest head's `AppHash`. 
+Merkle proofs from celestia-app and verify them against the latest head's `AppHash`.
 
 In order for the `StateAccessor` to do this, it would need access to the `header.Store`'s `Head()` method in order to get the latest known header of the node and check its `AppHash`.
 Then, instead of performing a regular `gRPC` query against the celestia-app's bank module, it would perform an ABCI request query via RPC as such: 
 
 ```go
-     prefixedAccountKey := append(bank_types.CreateAccountBalancesPrefix(addr.Bytes()), []byte(app.BondDenom)...)
-     abciReq := abci.RequestQuery{
-         Path:   fmt.Sprintf("store/%s/key", bank_types.StoreKey),
-    	 // we request the balance at the block *previous* to the current head as
-    	 // the AppHash committed to the head is calculated by applying the previous
-    	 // block's transaction list, not the current block, meaning the head's AppHash
-    	 // is actually a result of the previous block's transactions.
-         Height: head.Height - 1,
-         Data:   prefixedAccountKey,
-    	 // we set this value to `true` in order to get proofs in the response
-    	 // that we can use to verify against the head's AppHash.
-         Prove:  true,
-     }
-     result, err := ca.rpcCli.ABCIQueryWithOptions(ctx, abciReq.Path, abciReq.Data,  rpc_client.ABCIQueryOptions{})
-     if err != nil {
-         return nil, err
-     }
+  prefixedAccountKey := append(bank_types.CreateAccountBalancesPrefix(addr.Bytes()), []byte(app.BondDenom)...)
+  abciReq := abci.RequestQuery{
+      Path:   fmt.Sprintf("store/%s/key", bank_types.StoreKey),
+      // we request the balance at the block *previous* to the current head as
+      // the AppHash committed to the head is calculated by applying the previous
+      // block's transaction list, not the current block, meaning the head's AppHash
+      // is actually a result of the previous block's transactions.
+      Height: head.Height - 1,
+      Data:   prefixedAccountKey,
+      // we set this value to `true` in order to get proofs in the response
+      // that we can use to verify against the head's AppHash.
+      Prove:  true,
+  }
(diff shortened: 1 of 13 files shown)


docs/adr/adr-004-state-interaction.md
```diff
@@ -69,9 +69,54 @@ type StateAccessor interface {
...
+	}
+	root := proof_utils.NewMerkleRoot(head.AppHash)
+	// VerifyMembership expects the path as:
+	// []string{<store key of module>, <actual key corresponding to requested value>}
+	path := proof_utils.NewMerklePath(bank_types.StoreKey, string(prefixedAccountKey))
+	err = merkleproof.VerifyMembership(proof_utils.GetSDKSpecs(), root, path, result.Response.Value)
+	if err != nil {
+		return nil, err
+	}
+```
```
liamsi, line 119: Thanks for updating the ADR.

liamsi (approved):
I only reviewed this from a high-level perspective and not line by line. I'm on-board with the decisions made (reflected in the ADR) and the high-level code seems fine.

Wondertan (commented):
Comments from sync review:
* Consolidate `Balance` methods with `Verified` methods(TBD in this PR)
* Wait until syncing is done before requesting balances(TBD in this PR) and responding to the users.
	* Otherwise, users will get outdated balances during sync
* Take the latest pending Head from the Syncer and verify balances against it(Keeping it for later, needs an issue)
	* Pending headers are verified with known valset(within unbonding period) and safe to rely on
	* This way, users will be able to get balances(and other state) instantly, without waiting for sync to finish

service/state/interface.go
```diff
@@ -17,20 +17,24 @@ type Accessor interface {
...
-	SubmitPayForData(ctx context.Context, nID namespace.ID, data []byte, gasLim uint64) (*TxResponse, error)
-
-	// Balance retrieves the Celestia coin balance
-	// for the node's account/signer.
+	// Balance retrieves the Celestia coin balance for the node's account/signer
+	// and verifies it against the corresponding block's AppHash.
 	Balance(ctx context.Context) (*Balance, error)
-	// BalanceForAddress retrieves the Celestia coin balance
-	// for the given types.AccAddress.
+	// BalanceForAddress retrieves the Celestia coin balance for the given types.AccAddress.
```
adlerjohn, line 23: Duplicate?

adlerjohn (commented):
Where is `verified_balance` implemented? I don't actually see any code for that endpoint.

renaynay:
@adlerjohn originally, I created a separate endpoint for verified balance, but after speaking with hlib, we decided to just replace the current balance endpoint with this verified one -- so a user requesting `/balance` will now get their balance returned (but it's now verified against the AppHash instead of being blindly returned)

adlerjohn:
Oh got it, I got confused by the old PR description

adlerjohn (approved):

service/state/core_access.go
```diff
@@ -157,10 +220,23 @@ func (ca *CoreAccessor) Transfer(
...
-	msg := banktypes.NewMsgSend(from, to, coins)
+	msg := bank_types.NewMsgSend(from, to, coins)
 	signedTx, err := ca.constructSignedTx(ctx, msg, apptypes.SetGasLimit(gasLim))
 	if err != nil {
 		return nil, err
 	}
 	return ca.SubmitTx(ctx, signedTx)
 }
+
+func sdkErrorToGRPCError(resp sdk_abci.ResponseQuery) error {
```
vgonkivs, line 231: may be it is worth to put such functions in a separate file? `service/state/helpers.go ` or `service/state/utils `

Wondertan (commented):
LGMT, but we also need a test for the Balance with verification. I know it is not straightforward to do that for App, but there should be something to test ABCI with the banking module. At least knowing how to do that should be part of this PR, so we can make an issue to write a test.

renaynay:
@Wondertan regarding unit test for ABCI query -- the problem here is again bottlenecked by simulating tendermint+app. It will be easy to tell by using the endpoint itself whether it is working or not. I know it's not ideal, but until we can simulate this pairing easily, it will be way more work than it's worth to do this test. The underlying methods used to prove membership *are* covered by tests inside `ibc-go` [here](https://github.com/cosmos/ibc-go/blob/main/modules/core/23-commitment/types/merkle_test.go).

Ref: https://github.com/celestiaorg/celestia-node/issues/570

service/state/core_access.go
```diff
@@ -109,21 +129,60 @@ func (ca *CoreAccessor) Balance(ctx context.Context) (*Balance, error) {
...
+		Path:   fmt.Sprintf("store/%s/key", banktypes.StoreKey),
+		Height: head.Height - 1,
+		Data:   prefixedAccountKey,
+		Prove:  true,
+	}
+	opts := rpcclient.ABCIQueryOptions{
+		Height: abciReq.Height,
+		Prove:  abciReq.Prove,
+	}
+	result, err := ca.rpcCli.ABCIQueryWithOptions(ctx, abciReq.Path, abciReq.Data, opts)
```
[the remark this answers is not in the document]
renaynay: I know 😢  the reason being that the other option `ABCIQuery` doesn't take a height, it only takes the path and data so in order to specify a specific height, I have to use this option 😢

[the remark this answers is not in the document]
Wondertan: Ok

Wondertan (approved):

PR #913, opened 2022-07-15
feat(node/p2p): enable QUIC by default with the config

### LINKED ISSUE #686: p2p: use quic as default transport
Currently, we default to tcp, which is fine, but we can do better and follow libp2p defaults about using quic by default. There is literally no downside to doing so already and we can even make for incoming testnet

Wondertan:
> Modular software plague warning ⚠️ ☢️ 

This makes our nodes use QUIC by default while still supporting TCP and falling back to it if there is a case. 
As simple as it is. 
 
Fixes #686

### DIFF
--- a/node/p2p/p2p.go
+++ b/node/p2p/p2p.go
@@ -39,11 +39,16 @@ type Config struct {
 func DefaultConfig() Config {
 	return Config{
 		ListenAddresses: []string{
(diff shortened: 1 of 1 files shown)


adlerjohn (approved):
utACK

liamsi (approved):
👍🏼

renaynay (approved):
utACK

PR #924, opened 2022-07-20
fix: change applied availability for bridge node

vgonkivs:
Changing applied Availability type is done in order the bridge node can advertise itself in p2p network.


Wondertan (approved):
Please remove the todo and smorge

renaynay (approved):

PR #927, opened 2022-07-21
deps: replacing blake2b implementation with native /x/crypto

### LINKED ISSUE #908: deps: drop usage of minio/blake2b in favour of native /x/crypto implementation
We can follow the [multihash example](https://github.com/multiformats/go-multihash/pull/157) here. which states the Go's implementation of blake is no faster and drop the dependency.

The task would be to find all the usages(I can think of one only right now in `header/p2p/subscriber`) and swap with native Go's one

distractedm1nd:
Closes #908


Wondertan (approved):
Noice!

renaynay (approved):

adlerjohn (approved):

PR #928, opened 2022-07-21
feat: `/heath` rpc endpoint

### LINKED ISSUE #739: [Feature Request]: `/health` RPC endpoint
### Implementation ideas

Right now there is no good way to check that Node RPC is functional/accessible:
 - `/` gives HTTP 404
 - other endpoints can return app-specific errors

Simple `/health` endpoint enables easy check of RPC availability.
Requirements:
 - no cache
 - return valid JSON (KISS: `{"status": "ok"}` should be enough for now)
 - HTTP status code 200 if everything is fine

In the future, more information may be added to endpoint.

References/reading:
https://testfully.io/blog/api-health-check-monitoring/
https://docs.microsoft.com/en-us/azure/architecture/patterns/health-endpoint-monitoring

distractedm1nd:
Would close #739 ?

If I've understood correctly, the `/health` endpoint is just for checking if the rpc is running - not sure if this is still wanted/necessary though, feel free to delete.

Should more health endpoints be created for any other services? For example, with the new `DASState`, the /daser endpoint seems like a great "health check" to me

### DIFF
--- a/service/rpc/health.go
+++ b/service/rpc/health.go
@@ -1,0 +1,30 @@
+package rpc
+
+import (
+	"encoding/json"
+	"net/http"
+)
+
+const healthEndpoint = "/health"
+
+// HealthResponse represents the response to a
+// `/health` request.
+type HealthResponse struct {
+	Status string `json:"status"`
+}
+
+func (h *Handler) handleHealthRequest(w http.ResponseWriter, r *http.Request) {
+	availResp := &HealthResponse{
+		Status: "ok",
+	}
+
+	resp, err := json.Marshal(availResp)
+	if err != nil {
+		writeError(w, http.StatusInternalServerError, healthEndpoint, err)
+		return
+	}
+	_, werr := w.Write(resp)
+	if werr != nil {
+		log.Errorw("serving request", "endpoint", healthEndpoint, "err", err)
+	}
+}
--- a/service/rpc/health.go
+++ b/service/rpc/health.go
@@ -10,12 +10,24 @@ const healthEndpoint = "/health"
 // HealthResponse represents the response to a
 // `/health` request.
 type HealthResponse struct {
-	Status string `json:"status"`
+	StateService  bool             `json:"state_service"`
+	ShareService  bool             `json:"share_service"`
+	DASStatus     DasStateResponse `json:"das_status"`
+	HeaderSyncing bool             `json:"header_syncing"`
 }
 
 func (h *Handler) handleHealthRequest(w http.ResponseWriter, r *http.Request) {
 	availResp := &HealthResponse{
-		Status: "ok",
+		StateService:  !h.state.IsStopped(),
+		ShareService:  !h.share.IsStopped(),
+		HeaderSyncing: h.header.IsSyncing(),
+	}
+
+	if h.das != nil {
+		dasState := new(DasStateResponse)
+		dasState.SampleRoutine = h.das.SampleRoutineState()
+		dasState.CatchUpRoutine = h.das.CatchUpRoutineState()
+		availResp.DASStatus = *dasState
 	}
 
 	resp, err := json.Marshal(availResp)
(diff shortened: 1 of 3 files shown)


renaynay (commented):
@distractedm1nd looks great so far - maybe we can extend the `/health` response to also have a map of services -> running/not running

distractedm1nd:
```go
type HealthResponse struct {
	StateService  bool             `json:"state_service"`
	ShareService  bool             `json:"share_service"`
	DASStatus     DasStateResponse `json:"das_status"`
	HeaderSyncing bool             `json:"header_syncing"`
}
```

@renaynay were you thinking something like this?

renaynay:
> ```go
> type HealthResponse struct {
> 	StateService  bool             `json:"state_service"`
> 	ShareService  bool             `json:"share_service"`
> 	DASStatus     DasStateResponse `json:"das_status"`
> 	HeaderSyncing bool             `json:"header_syncing"`
> }
> ```
> 
> @renaynay were you thinking something like this?
> 
> Thoughts:
> 
> * should it really contain the das state? Or should i also just add a `IsStopped` like the other services?
> * took out the `"status": "ok"`, not sure if that is good or bad yet
> * used header syncing, since the header service looks to be "always on"?

@distractedm1nd lets do something like (excuse formatting): 

```
type HealthResponse struct {
         RPC string `json:"rpc_service"`
	StateService  string             `json:"state_service"`
	DASService    string `json:"das_service"`
	HeaderSyncing string             `json:"sync_service"`
}
```
So basically call all of these things services (we also don't need share service bc das covers that, and share service is kind of an under the hood thing)

Bool would be fine but I think it's better to have `running` || `stopped` strings bc asking for /health and getting `true || false` is not descriptive.

Wdyt about this?

renaynay:
It would also be nice to have a list of available endpoints on the node the way that tendermint does: https://rpc-mamaki.pops.one/

distractedm1nd:
Yes, makes sense - two things:

1. Should we maybe put this in a `/status` endpoint, and keep `/health` as a `{"status": "ok"}`? I noticed that the rpc at your link does this
2. `RPC` would always return `running` for now, correct? Since it has to be running to respond to the endpoint

renaynay:
@distractedm1nd yes to both 1 and 2

service/rpc/health.go
```diff
@@ -0,0 +1,42 @@
...
+		StateService:  !h.state.IsStopped(),
+		ShareService:  !h.share.IsStopped(),
+		HeaderSyncing: h.header.IsSyncing(),
+	}
+
+	if h.das != nil {
+		dasState := new(DasStateResponse)
+		dasState.SampleRoutine = h.das.SampleRoutineState()
+		dasState.CatchUpRoutine = h.das.CatchUpRoutineState()
+		availResp.DASStatus = *dasState
```
Wondertan, line 30: We don't need to allocate response and then copy it again
```suggestion
		availResp.DASStatus = &DasStateResponse{
			SampleRoutine: h.das.SampleRoutineState(),
			CatchUpRoutine: h.das.CatchUpRoutineState(),
		}
```

Wondertan (changes requested):
I am good with this, however:
* Share and State services running or not give node operator _zero_ info. Starting and stopping happen only at the start and stop of the node. Besides State service which can be stopped in runtime with fraud proofs case, but more on this later. 
	* Also, why only those services, and why not Fraud?
* Call this `status` endpoint rather than `health` (extending on the point above)
	* `status` is more about data being returned and implies that the node operator himself understands if the node is running healthy based on the returned data,
	* A health check implies some additional "smart" logic that evaluates node health itself and only returns a simple true and false. In our case, this logic could be a combination of:
		* Checking that header are constantly synced
		* Checking that header are constantly sampled
		* Checking if any FraudProof exist/received 
	* `health` might be extracted as a separate issue

Note that this will need to be reworked with newer PublicAPI over RPC, so I am fine with not making this super clean and changes are only requested for the comments below.

Thank you!

distractedm1nd:
Not worth pursuing until the refactorings, so removing from the roadmap for now

Issue #947, opened 2022-07-26
ci: Lint Markdown files


rootulp:
### Implementation ideas

celestia-node contains ADRs in [docs/adr](https://github.com/celestiaorg/celestia-node/tree/main/docs/adr) that could benefit from consistent styling in Markdown files. One idea is to use [Markdownlint](https://github.com/markdownlint/markdownlint).

Originally motivated by: https://github.com/celestiaorg/celestia-node/pull/922#discussion_r930044696

adlerjohn:
https://github.com/FuelLabs/fuel-indexer/blob/a96f8ec9e9b3edd35d564f4d079a2739a3bb0a7c/.github/workflows/ci.yml#L217-L227

PR #948, opened 2022-07-26
Add GitHub CI check for Markdownlint

rootulp:
Resolves https://github.com/celestiaorg/celestia-node/issues/947


.markdownlintrc
```diff
@@ -0,0 +1,5 @@
+{
+    "default": true,
+    "MD013": false,
```
rootulp, line 3: This line disables the rule for max line length: https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md013---line-length= because there were many errors and fixing all would bloat this diff. If we want this enabled, we can enable in a separate follow-up PR.

.markdownlintrc
```diff
@@ -0,0 +1,5 @@
...
+    "MD033": false
```
rootulp, line 4: This line disables the rule banning inline HTML: https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md033---inline-html= but our ADRs use inline HTML

docs/adr/adr-006-fraud-service.md
```diff
@@ -62,157 +64,169 @@ In addition, `das.Daser`:
 
 1. Creates a BEFP:
 
-```go
```
rootulp, line 65: The following sections were indented so that they display under the ordered list item they are relevant to.

extensions.json
```diff
@@ -0,0 +1,5 @@
+{
+    "recommendations": [
+        "DavidAnson.vscode-markdownlint"
```
rootulp, line 3: This file can be removed but I think it may aid contributors if they install the most popular markdownlint extension

Screen Shot 2022-07-26 at 2 30 41 PM
 
so that contributors can catch markdownlint violations at development time rather than at CI time.

adlerjohn: Can you split this up into a separate PR just so it doesn't block? Some devs don't use vscode so this could result in bikeshedding.

.markdownlintrc
adlerjohn: Is it possible to add comments to these lines to explain what they do?

rootulp: great idea! Something felt off about explaining this in a PR comment rather than inline

Screen Shot 2022-07-26 at 5 07 21 PM

Comments are possible in YAML so I'll switch this JSON file to YAML

extensions.json
rootulp: Extracted to https://github.com/celestiaorg/celestia-node/issues/949

adlerjohn (approved):

Wondertan (approved):
Awesome. Thanks for this

renaynay (approved):
one rebase pls @rootulp 🙏🏻

rootulp:
@Wondertan @renaynay this PR is ready to be merged. 

May I request write access to this repo so that I can merge it / future approved PRs?

Screen Shot 2022-07-27 at 11 41 09 AM

PR #938, opened 2022-07-23
fraud: rework unmarshaler registration

vgonkivs:
Resolves https://github.com/celestiaorg/celestia-node/issues/941

Partly resolves #823

Issue #823, opened 2022-06-14
header|fraud: a simple trick to lower allocations with PubSub


Wondertan:
Semantically, we have FraudSub and HeaderSub, which rely on libp2p's PubSub. Every msgs in pubsub goes through validation and then to subscription. To validate a msg, we deserialize it, and per each subscription, we deserialize it. In the HeaderSub we even deserialize additionally for msgID gen. With the trick we can do one only and here is it:
PubSub's msg has an internal field [`ValidatorData`](https://github.com/libp2p/go-libp2p-pubsub/blob/master/pubsub.go#L223) which we can set during validation. This field is propagated to each subscription that can access it, convert to the correct type, and pass them to the user.

PR #953, opened 2022-07-28
header/p2p: set ValidatorData field for header message

vgonkivs:
Second part for #823

Note: ValidatorData for fraud package is implemented in #938

### DIFF
--- a/header/p2p/subscriber.go
+++ b/header/p2p/subscriber.go
@@ -54,7 +54,7 @@ func (p *Subscriber) AddValidator(val header.Validator) error {
 				"err", err)
 			return pubsub.ValidationReject
 		}
-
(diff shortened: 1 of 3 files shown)


Wondertan (approved):
Nice. Thank you!

renaynay (approved):
I'm fine with this -- just so I understand correctly, the header gets stored both in `msg.Data` and `msg.ValidatorData` but only inside topic validator locally, correct?

vgonkivs:
Header is stored in msg.Data as raw bytes and in msg.ValidatorData as unmarshaled struct. So, we can unmarshal in only once and avoid extra memory allocation.

>  inside topic validator locally

Technically, yes. We set it on our side.

Issue #954, opened 2022-07-28
fraud: blacklist republisher of the incorrect message, not the original publisher 


Wondertan:
Hmm, in terms of blocklisting, we might be doing it wrong, because we should probably blocklist this peer we ignore he in the first place. Can be wrong, but iirc, this peer is the one from which we received the message, while the peer on the msg is originator. The difference here is that someone could regossip a msg to us from the originator and we should blaklist regossiper in the first place, which we probably ignore here.

_Originally posted by @Wondertan in https://github.com/celestiaorg/celestia-node/pull/938#discussion_r931965005_

We confirmed that we are blacklisting the wrong peer. Blacklisting the publisher is pointless because an adversarial peer can spam the network with msgs signed/published by randomly generating peer

PR #966, opened 2022-07-29
fraud: add peer that sent an invalid fraud proof to black list

vgonkivs:
Resolves https://github.com/celestiaorg/celestia-node/issues/954

### DIFF
--- a/fraud/interface.go
+++ b/fraud/interface.go
@@ -5,6 +5,8 @@ import (
 
 	logging "github.com/ipfs/go-log/v2"
 
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
+
 	"github.com/celestiaorg/celestia-node/header"
 )
 
@@ -28,7 +30,7 @@ type Service interface {
 type Broadcaster interface {
 	// Broadcast takes a fraud `Proof` data structure that implements standard BinaryMarshal
 	// interface and broadcasts it to all subscribed peers.
-	Broadcast(context.Context, Proof) error
+	Broadcast(context.Context, Proof, ...pubsub.PubOpt) error
 }
 
 // Subscriber encompasses the behavior necessary to
--- a/fraud/interface.go
+++ b/fraud/interface.go
@@ -5,8 +5,6 @@ import (
 
 	logging "github.com/ipfs/go-log/v2"
 
-	pubsub "github.com/libp2p/go-libp2p-pubsub"
-
 	"github.com/celestiaorg/celestia-node/header"
 )
 
@@ -30,7 +28,7 @@ type Service interface {
 type Broadcaster interface {
 	// Broadcast takes a fraud `Proof` data structure that implements standard BinaryMarshal
 	// interface and broadcasts it to all subscribed peers.
-	Broadcast(context.Context, Proof, ...pubsub.PubOpt) error
+	Broadcast(context.Context, Proof) error
 }
 
 // Subscriber encompasses the behavior necessary to
--- a/fraud/service.go
+++ b/fraud/service.go
@@ -60,7 +60,7 @@ func (f *service) Subscribe(proofType ProofType) (_ Subscription, err error) {
 	return &subscription{subs}, nil
 }
 
-func (f *service) Broadcast(ctx context.Context, p Proof, opts ...pubsub.PubOpt) error {
+func (f *service) Broadcast(ctx context.Context, p Proof) error {
 	bin, err := p.MarshalBinary()
 	if err != nil {
 		return err
@@ -71,7 +71,7 @@ func (f *service) Broadcast(ctx context.Context, p Proof, opts ...pubsub.PubOpt)
 	if !ok {
 		return fmt.Errorf("fraud: unmarshaler for %s proof is not registered", p.Type())
 	}
-	return t.Publish(ctx, bin, opts...)
+	return t.Publish(ctx, bin)
 }
 
 func (f *service) processIncoming(
--- a/fraud/service.go
+++ b/fraud/service.go
@@ -9,7 +9,6 @@ import (
 	"time"
 
 	"github.com/ipfs/go-datastore"
-	"github.com/ipfs/go-datastore/namespace"
 	"github.com/libp2p/go-libp2p-core/peer"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 )
@@ -119,14 +118,14 @@ func (f *service) processIncoming(
 	msg.ValidatorData = proof
 	f.storesLk.RLock()
 	store, ok := f.stores[proofType]
+	if !ok {
+		store = initStore(proofType, f.ds)
+		f.stores[proofType] = store
+	}
 	f.storesLk.RUnlock()
-	if ok {
-		err = put(ctx, store, string(proof.HeaderHash()), msg.Data)
-		if err != nil {
-			log.Error(err)
-		}
-	} else {
-		log.Warnf("no store for incoming proofs type %s", proof.Type())
+	err = put(ctx, store, string(proof.HeaderHash()), msg.Data)
+	if err != nil {
+		log.Error(err)
 	}
 	log.Warn("Shutting down services...")
 	return pubsub.ValidationAccept
@@ -136,7 +135,7 @@ func (f *service) Get(ctx context.Context, proofType ProofType) ([]Proof, error)
 	f.storesLk.Lock()
 	store, ok := f.stores[proofType]
 	if !ok {
-		store = namespace.Wrap(f.ds, makeKey(proofType))
+		store = initStore(proofType, f.ds)
 		f.stores[proofType] = store
 	}
 	f.storesLk.Unlock()
--- a/fraud/service.go
+++ b/fraud/service.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/ipfs/go-datastore"
 	"github.com/ipfs/go-datastore/namespace"
+	"github.com/libp2p/go-libp2p-core/peer"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 )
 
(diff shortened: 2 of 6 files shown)


fraud/service_test.go
```diff
@@ -56,18 +60,216 @@ func TestService_Broadcast(t *testing.T) {
...
+	_, err = serviceA.Subscribe(BadEncoding)
+	require.NoError(t, err)
+
+	subsB, err := serviceB.Subscribe(BadEncoding)
+	require.NoError(t, err)
+	defer subsB.Cancel()
+
+	subsC, err := serviceC.Subscribe(BadEncoding)
+	require.NoError(t, err)
+	defer subsC.Cancel()
```
Wondertan, line 230: Cancel for A?

Wondertan: Same for another test

fraud/service_test.go
```diff
@@ -56,18 +60,216 @@ func TestService_Broadcast(t *testing.T) {
...
+	require.NoError(t, err)
+
+	_, err = subsB.Proof(ctx)
+	require.NoError(t, err)
+
+	newCtx, cancel := context.WithTimeout(ctx, time.Millisecond*500)
+	t.Cleanup(cancel)
+	_, err = subsC.Proof(newCtx)
+	require.Error(t, err)
+	require.True(t, bl.Contains(net.Hosts()[1].ID()))
```
Wondertan, line 154: Also, Add here a check that publisher A is not blacklisted

fraud/interface.go
```diff
@@ -28,7 +30,7 @@ type Service interface {
 type Broadcaster interface {
 	// Broadcast takes a fraud `Proof` data structure that implements standard BinaryMarshal
 	// interface and broadcasts it to all subscribed peers.
-	Broadcast(context.Context, Proof) error
+	Broadcast(context.Context, Proof, ...pubsub.PubOpt) error
```
Wondertan, line 33: Feel very conflicted about relying on PubSub structures in our interface. Is there a way to avoid it by publishing directly via pubsub accessed through service? We anyway convert the Service interface to a specific type, so that should be doable.

Wondertan (commented):
Nice!

One last thing we should to care about is testing/mocking implementations of Proofs so that we can test network behaviours, like in this tests, without relying on a specific Proof type.

fraud/service_test.go
vgonkivs: `subsA` is not created: `_, err = serviceA.Subscribe(BadEncoding)`

fraud/interface.go
vgonkivs: yes, sure.

fraud/service_test.go
Wondertan: Well, it is created because you call the func, you just ignore the value

vgonkivs: technically, yes. but should we still cancel it in this case?

Wondertan: and technically we should cancel. We should be nice to our code ;)

vgonkivs:
> One last thing we should to care about is testing/mocking implementations of Proofs so that we can test network behaviours, like in this tests, without relying on a specific Proof type.

Yes, I agree. But let's make it in scope of separate PR. IMO, it's not related to this one. Will create and post here an issue.

vgonkivs:
https://github.com/celestiaorg/celestia-node/issues/971

fraud/service_test.go
```diff
@@ -56,18 +60,334 @@ func TestService_Broadcast(t *testing.T) {
...
+	defer t.Cleanup(cancel)
+	// create mock network
+	net, err := mocknet.FullMeshLinked(3)
+	require.NoError(t, err)
+	bServ := mdutils.Bserv()
+
+	// create first fraud service that will broadcast incorrect Fraud Proof
+	serviceA, store1 := createServiceWithHost(ctx, t, net.Hosts()[0])
+
+	h, err := store1.GetByHeight(context.TODO(), 1)
```
Wondertan, line 279: 🚓

fraud/service_test.go
```diff
@@ -56,18 +60,334 @@ func TestService_Broadcast(t *testing.T) {
...
```
Wondertan, line 179: 🚓

Wondertan (approved):
Nice!

The further #971 should prettify things a bit more

fraud/service.go
```diff
@@ -106,25 +107,25 @@ func (f *service) processIncoming(
...
-		"from", msg.ReceivedFrom.String(),
+		"from", from.String(),
 	)
 	msg.ValidatorData = proof
 	f.storesLk.RLock()
 	store, ok := f.stores[proofType]
+	if !ok {
+		store = initStore(proofType, f.ds)
+		f.stores[proofType] = store
+	}
```
renaynay, line 124: if you're going to do this, then you need Lock, not RLock/RUnlock

fraud/service_test.go
renaynay: Lol +1

renaynay (changes requested):
Everything looks generally okay but I don't understand the reasoning behind adding the `from peer.ID` param to the validator func instead of just getting the peer from the `msg.ReceivedFrom` field -- what does this change?

Also need to go through tests - they are quite intense so i'm gonna do another pass through just for the test

fraud/service_test.go
```diff
@@ -56,18 +60,334 @@ func TestService_Broadcast(t *testing.T) {
...
+	f := func(ctx context.Context, from peer.ID, msg *pubsub.Message) pubsub.ValidationResult {
+		msg.ValidatorData = befp
+		return pubsub.ValidationAccept
+	}
+	fserviceB.pubsub.RegisterTopicValidator(getSubTopic(BadEncoding), f) //nolint:errcheck
+	bin, err := befp.MarshalBinary()
+	require.NoError(t, err)
+	topic, ok := fserviceA.topics[BadEncoding]
+	require.True(t, ok)
+	// we cannot avoid sleep because it helps to avoid flakiness
```
renaynay, line 150: What is this sleep for though? What are you waiting on?

fraud/service_test.go
```diff
@@ -56,18 +60,334 @@ func TestService_Broadcast(t *testing.T) {
...
+	time.Sleep(time.Millisecond * 100)
+	err = topic.Publish(ctx, bin, pubsub.WithReadiness(pubsub.MinTopicSize(1)))
+	require.NoError(t, err)
+
+	_, err = subsB.Proof(ctx)
+	require.NoError(t, err)
+
+	newCtx, cancel := context.WithTimeout(ctx, time.Millisecond*500)
+	defer cancel()
+	_, err = subsC.Proof(newCtx)
```
renaynay, line 160:
```suggestion

	_, err = subsC.Proof(newCtx)
```

fraud/service_test.go
```diff
@@ -56,18 +60,334 @@ func TestService_Broadcast(t *testing.T) {
...
+	fserviceA := serviceA.(*service)
+	require.NotNil(t, fserviceA)
+
+	// create second service that will receive and validate Fraud Proof
+	serviceB, _ := createServiceWithHost(ctx, t, net.Hosts()[1])
+
+	fserviceB := serviceB.(*service)
+	require.NotNil(t, fserviceB)
+
+	bl, err := pubsub.NewTimeCachedBlacklist(time.Hour * 1)
```
renaynay, line 93: In a longer test like this, especially where the use of the var is further below the definition, can we use full names like `blocklist` here?

fraud/service_test.go
```diff
@@ -56,18 +60,334 @@ func TestService_Broadcast(t *testing.T) {
...
+	require.NoError(t, err)
+
+	newCtx, cancel := context.WithTimeout(ctx, time.Millisecond*500)
+	defer cancel()
+	_, err = subsC.Proof(newCtx)
+	require.Error(t, err)
+	require.False(t, bl.Contains(net.Hosts()[0].ID()))
+	require.True(t, bl.Contains(net.Hosts()[1].ID()))
+	// we cannot avoid sleep because it helps to avoid flakiness
+	time.Sleep(time.Millisecond * 100)
```
renaynay, line 165: Why sleep here too?

fraud/service_test.go
```diff
@@ -56,18 +60,334 @@ func TestService_Broadcast(t *testing.T) {
...
+}
+
+func TestService_GossipingOfFaultBEFP(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
+	defer t.Cleanup(cancel)
```
renaynay, line 170: you don't need to `defer` for `t.Cleanup` - it's redundant

renaynay (commented):
So the tests are quite long and hard to follow what's going on and why - it would be nice to clean them up a bit, reduce LOC as much as possible (for example, no need to define `bServ` if it's only used in one place, just pass `mdutils.Bserv()` directly as a param). The tests inside `sync_test.go` are really nice to follow, for example.

vgonkivs:
msg.ReceivedFrom field is the address of node that have created a Fraud Proof. But it could be a case when other node receives a Fraud Proof and regossips it further

fraud/service_test.go
vgonkivs: libp2p specific sleeps 😞 
Test will fail without it

fraud/service_test.go
vgonkivs: to avoid flakiness

fraud/service_test.go
```diff
@@ -352,7 +348,8 @@ func TestService_GossipingOfBEFP(t *testing.T) {
 	require.NoError(t, err)
 
 	newCtx, cancel := context.WithTimeout(ctx, time.Millisecond*100)
-	t.Cleanup(cancel)
+	defer cancel()
```
renaynay, line 351: Nit but let's be consistent with way we cancel context. I see some places use t.Cleanup and some defer

renaynay (approved):

Issue #967, opened 2022-07-29
[Feature Request]: add Makefile task to run markdownlint


rootulp:
### Implementation ideas

Proposed by @Wondertan

adlerjohn:
Related: #940

PR #969, opened 2022-07-29
feat(Makefile): formatting and linter for markdown

Wondertan:
A simple change; the only problem is that it requires installed `markdownlint` under the PATH. If we don't want this, we can either: ignore the error if not installed or extract into a separate Makefile command

Closes https://github.com/celestiaorg/celestia-node/issues/967

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,7 +70,7 @@ fmt:
 lint:
 	@echo "--> Running linter"
 	@golangci-lint run
-	@markdownlint --config .markdownlint.yaml .
+	@markdownlint --config .markdownlint.yaml '**/*.md'
 .PHONY: lint
 
 ## test-unit: Running unit tests
--- a/Makefile
+++ b/Makefile
@@ -63,12 +63,14 @@ fmt:
 	@find . -name '*.go' -type f -not -path "*.git*" -not -name '*.pb.go' -not -name '*pb_test.go' | xargs gofmt -w -s
 	@find . -name '*.go' -type f -not -path "*.git*"  -not -name '*.pb.go' -not -name '*pb_test.go' | xargs goimports -w -local github.com/celestiaorg
 	@go mod tidy -compat=1.17
+	@markdownlint -fq -c .markdownlint.yaml .
 .PHONY: fmt
 
 ## lint: Linting *.go files using golangci-lint. Look for .golangci.yml for the list of linters.
 lint:
 	@echo "--> Running linter"
 	@golangci-lint run
+	@markdownlint -c .markdownlint.yaml .
 .PHONY: lint
 
 ## test-unit: Running unit tests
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -63,12 +63,14 @@ fmt:
...
 	@find . -name '*.go' -type f -not -path "*.git*"  -not -name '*.pb.go' -not -name '*pb_test.go' | xargs goimports -w -local github.com/celestiaorg
 	@go mod tidy -compat=1.17
+	@markdownlint -fq -c .markdownlint.yaml .
 .PHONY: fmt
 
 ## lint: Linting *.go files using golangci-lint. Look for .golangci.yml for the list of linters.
 lint:
 	@echo "--> Running linter"
 	@golangci-lint run
+	@markdownlint -c .markdownlint.yaml .
```
rootulp, line 73:
```suggestion
	@markdownlint --config .markdownlint.yaml .
```

Makefile
```diff
@@ -63,12 +63,14 @@ fmt:
 	@find . -name '*.go' -type f -not -path "*.git*" -not -name '*.pb.go' -not -name '*pb_test.go' | xargs gofmt -w -s
 	@find . -name '*.go' -type f -not -path "*.git*"  -not -name '*.pb.go' -not -name '*pb_test.go' | xargs goimports -w -local github.com/celestiaorg
 	@go mod tidy -compat=1.17
+	@markdownlint -fq -c .markdownlint.yaml .
```
rootulp, line 66:
```suggestion
	@markdownlint --fix --quiet --config .markdownlint.yaml .
```

the long command option is clearer in a file that other developers will read. The short command option is faster when typing command in your shell. 

https://changelog.com/posts/use-long-flags-when-scripting

rootulp (approved):
> the only problem is that it requires installed markdownlint under the PATH.

We can inform developers if they haven't installed markdownlint via something like: https://github.com/open-telemetry/opentelemetry-specification/blob/main/Makefile#L56-L63

Additionally we may want to include a note in the README for contributors to install markdownlint.

Makefile
Wondertan: Good point

Makefile
```diff
@@ -63,12 +63,14 @@ fmt:
...
 	@find . -name '*.go' -type f -not -path "*.git*"  -not -name '*.pb.go' -not -name '*pb_test.go' | xargs goimports -w -local github.com/celestiaorg
 	@go mod tidy -compat=1.17
+	@markdownlint --fix --quiet --config .markdownlint.yaml .
 .PHONY: fmt
 
 ## lint: Linting *.go files using golangci-lint. Look for .golangci.yml for the list of linters.
 lint:
 	@echo "--> Running linter"
 	@golangci-lint run
+	@markdownlint --config .markdownlint.yaml .
```
adlerjohn, line 73: Should these be

```suggestion
	@markdownlint --config .markdownlint.yaml '**/*.md'
```

(ref: https://github.com/igorshubovych/markdownlint-cli/tree/8652eb2bd9f7339f0569cee254fc695b952c7c70#globbing)

Wondertan: I've tested it with a single dot multiple times, and it works for all the subdirectories, but I can change this to match specification

renaynay (approved):

PR #970, opened 2022-08-01
feat(ipld): concurrent GetSharesByNamespace

### LINKED ISSUE #697: ipld: concurrent GetSharesByNamespace
Stricly an optimization that would speed the time it takes to fetch shares for a namespace. Currently, we do it sequentially. The way to approach this should be taken from https://github.com/celestiaorg/celestia-node/pull/682

distractedm1nd:
Closes #697. Still very much a WIP, so a draft, have some open questions:

- [X] I feel like there should be a more elegant solution to keeping the leaves in order during the walk, so that we can avoid sorting afterwards. Haven't really been able to figure it out yet though `Answer: Overallocation is possible`
- [X] Should we be using the same pool from `get_shares.go`? `Answer: No`
- [X] Since we cannot preallocate the number of shares in the namespace, I buffered the channel with `NumWorkersLimit`. Does this make sense? `Answer: No`
- [X] Since the `job` struct is now in two locations, should we export it somewhere? I have a feeling that it may be too early to abstract it `Answer: Exported`
- [X]  `GetLeavesByNamespace` doesn't return an error anymore, felt right but still not 100% sure about it `Answer: now returns error again`

### DIFF
--- a/ipld/get.go
+++ b/ipld/get.go
@@ -2,6 +2,8 @@ package ipld
 
 import (
 	"context"
+	"sort"
+	"sync"
 
 	"github.com/ipfs/go-blockservice"
 	"github.com/ipfs/go-cid"
@@ -134,10 +136,10 @@ func GetSharesByNamespace(
 	root cid.Cid,
 	nID namespace.ID,
 ) ([]Share, error) {
-	leaves, err := GetLeavesByNamespace(ctx, bGetter, root, nID)
-	if err != nil {
-		return nil, err
-	}
+	leaves := GetLeavesByNamespace(ctx, bGetter, root, nID)
+	// if err != nil {
+	// 	return nil, err
+	// }
 
 	shares := make([]Share, len(leaves))
 	for i, leaf := range leaves {
@@ -148,42 +150,98 @@ func GetSharesByNamespace(
 }
 
 // GetLeavesByNamespace returns all the leaves from the given root with the given namespace.ID.
-// If nothing is found it returns both data and err as nil.
+// If nothing is found it returns data as nil.
 func GetLeavesByNamespace(
 	ctx context.Context,
 	bGetter blockservice.BlockGetter,
 	root cid.Cid,
 	nID namespace.ID,
-) ([]ipld.Node, error) {
-	err := SanityCheckNID(nID)
-	if err != nil {
-		return nil, err
+) []ipld.Node {
+	type job struct {
+		id  cid.Cid
+		pos int
 	}
-	rootH := plugin.NamespacedSha256FromCID(root)
-	if nID.Less(nmt.MinNamespace(rootH, nID.Size())) || !nID.LessOrEqual(nmt.MaxNamespace(rootH, nID.Size())) {
-		return nil, nil
+
+	// TODO: There has to be a more elegant solution
+	// bookkeeping is needed to be able to sort the leaves after the walk
+	type result struct {
+		node ipld.Node
+		pos  int
 	}
-	// request the node
-	nd, err := plugin.GetNode(ctx, bGetter, root)
-	if err != nil {
-		return nil, err
+
+	// TODO: Should this be NumWorkersLimit?
+	// we don't know the amount of shares in the namespace, so we cannot preallocate properly
+	jobs := make(chan *job, NumWorkersLimit)
+	jobs <- &job{id: root}
+
+	// the wg counter cannot be preallocated either, it is incremented with each job
+	wg := sync.WaitGroup{}
+	wg.Add(1)
+
+	var leaves []result
+	mu := &sync.Mutex{}
+
+	// once the wait group is done, we can close the jobs channel to break the following loop
+	go func() {
+		wg.Wait()
+		close(jobs)
+	}()
+
+	for j := range jobs {
+		j := j
+		// TODO: This is using the pool from get_shares.go, is this okay?
+		pool.Submit(func() {
+			defer wg.Done()
+			err := SanityCheckNID(nID)
+			if err != nil {
+				return
+			}
+
+			rootH := plugin.NamespacedSha256FromCID(j.id)
+			if nID.Less(nmt.MinNamespace(rootH, nID.Size())) || !nID.LessOrEqual(nmt.MaxNamespace(rootH, nID.Size())) {
+				return
+			}
+
+			nd, err := plugin.GetNode(ctx, bGetter, j.id)
+			if err != nil {
+				return
+			}
+
+			lnks := nd.Links()
+			if len(lnks) == 1 {
+				mu.Lock()
+				leaves = append(leaves, result{nd, j.pos})
+				mu.Unlock()
+				return
+			}
+
+			for i, lnk := range lnks {
+				select {
+				case jobs <- &job{
+					id:  lnk.Cid,
+					pos: j.pos*2 + i,
+				}:
+					wg.Add(1)
+				case <-ctx.Done():
+					return
+				}
+			}
+		})
 	}
-	// check links
-	lnks := nd.Links()
-	if len(lnks) == 1 {
-		// if there is one link, then this is a leaf node, so just return it
-		return []ipld.Node{nd}, nil
-	}
-	// if there are some links, then traverse them
-	var out []ipld.Node
-	for _, lnk := range nd.Links() {
-		nds, err := GetLeavesByNamespace(ctx, bGetter, lnk.Cid, nID)
-		if err != nil {
-			return out, err
+
+	if len(leaves) > 0 {
+		sort.SliceStable(leaves, func(i, j int) bool {
+			return leaves[i].pos < leaves[j].pos
+		})
+
+		output := make([]ipld.Node, len(leaves))
+		for i, leaf := range leaves {
+			output[i] = leaf.node
 		}
-		out = append(out, nds...)
+		return output
 	}
-	return out, nil
+
+	return nil
 }
 
 // leafToShare converts an NMT leaf into a Share.
--- a/ipld/get.go
+++ b/ipld/get.go
@@ -2,17 +2,12 @@ package ipld
 
 import (
 	"context"
-	"sort"
-	"sync"
 
 	"github.com/ipfs/go-blockservice"
 	"github.com/ipfs/go-cid"
 	ipld "github.com/ipfs/go-ipld-format"
 
 	"github.com/celestiaorg/celestia-node/ipld/plugin"
-	"github.com/celestiaorg/nmt"
-
-	"github.com/celestiaorg/nmt/namespace"
 )
 
 // GetShare fetches and returns the data for leaf `leafIndex` of root `rootCid`.
@@ -126,126 +121,6 @@ func GetProof(
 
 	// recursively walk down through selected children
 	return GetProof(ctx, bGetter, root, proof, leaf, total)
-}
-
-// GetSharesByNamespace returns all the shares from the given root
-// with the given namespace.ID.
-func GetSharesByNamespace(
-	ctx context.Context,
-	bGetter blockservice.BlockGetter,
-	root cid.Cid,
-	nID namespace.ID,
-) ([]Share, error) {
-	leaves := GetLeavesByNamespace(ctx, bGetter, root, nID)
-	// if err != nil {
-	// 	return nil, err
-	// }
-
-	shares := make([]Share, len(leaves))
-	for i, leaf := range leaves {
-		shares[i] = leafToShare(leaf)
-	}
-
-	return shares, nil
-}
-
-// GetLeavesByNamespace returns all the leaves from the given root with the given namespace.ID.
-// If nothing is found it returns data as nil.
-func GetLeavesByNamespace(
-	ctx context.Context,
-	bGetter blockservice.BlockGetter,
-	root cid.Cid,
-	nID namespace.ID,
-) []ipld.Node {
-	type job struct {
-		id  cid.Cid
-		pos int
-	}
-
-	// TODO: There has to be a more elegant solution
-	// bookkeeping is needed to be able to sort the leaves after the walk
-	type result struct {
-		node ipld.Node
-		pos  int
-	}
-
-	// TODO: Should this be NumWorkersLimit?
-	// we don't know the amount of shares in the namespace, so we cannot preallocate properly
-	jobs := make(chan *job, NumWorkersLimit)
-	jobs <- &job{id: root}
-
-	// the wg counter cannot be preallocated either, it is incremented with each job
-	wg := sync.WaitGroup{}
-	wg.Add(1)
-
-	var leaves []result
-	mu := &sync.Mutex{}
-
-	// once the wait group is done, we can close the jobs channel to break the following loop
-	go func() {
-		wg.Wait()
-		close(jobs)
-	}()
-
-	for j := range jobs {
-		j := j
-		// TODO: This is using the pool from get_shares.go, is this okay?
-		pool.Submit(func() {
-			defer wg.Done()
-			err := SanityCheckNID(nID)
-			if err != nil {
-				return
-			}
-
-			rootH := plugin.NamespacedSha256FromCID(j.id)
-			if nID.Less(nmt.MinNamespace(rootH, nID.Size())) || !nID.LessOrEqual(nmt.MaxNamespace(rootH, nID.Size())) {
-				return
-			}
-
-			nd, err := plugin.GetNode(ctx, bGetter, j.id)
-			if err != nil {
-				return
-			}
-
-			lnks := nd.Links()
-			// wg counter should be incremented before adding new jobs
-			if len(lnks) > 1 {
-				wg.Add(len(lnks))
-			}
-
-			if len(lnks) == 1 {
-				mu.Lock()
-				leaves = append(leaves, result{nd, j.pos})
-				mu.Unlock()
-				return
-			}
-
-			for i, lnk := range lnks {
-				select {
-				case jobs <- &job{
-					id:  lnk.Cid,
-					pos: j.pos*2 + i,
-				}:
-				case <-ctx.Done():
-					return
-				}
-			}
-		})
-	}
-
-	if len(leaves) > 0 {
-		sort.SliceStable(leaves, func(i, j int) bool {
-			return leaves[i].pos < leaves[j].pos
-		})
-
-		output := make([]ipld.Node, len(leaves))
-		for i, leaf := range leaves {
-			output[i] = leaf.node
-		}
-		return output
-	}
-
-	return nil
 }
 
 // leafToShare converts an NMT leaf into a Share.
--- a/ipld/get.go
+++ b/ipld/get.go
@@ -208,6 +208,11 @@ func GetLeavesByNamespace(
 			}
 
 			lnks := nd.Links()
+			// wg counter should be incremented before adding new jobs
+			if len(lnks) > 1 {
+				wg.Add(len(lnks))
+			}
+
 			if len(lnks) == 1 {
 				mu.Lock()
 				leaves = append(leaves, result{nd, j.pos})
@@ -221,7 +226,6 @@ func GetLeavesByNamespace(
 					id:  lnk.Cid,
 					pos: j.pos*2 + i,
 				}:
-					wg.Add(1)
 				case <-ctx.Done():
 					return
 				}
--- a/ipld/get_namespaced_shares.go
+++ b/ipld/get_namespaced_shares.go
@@ -17,7 +17,7 @@ import (
 	"github.com/celestiaorg/nmt/namespace"
 )
 
-// TODO(@distractedm1nd) Find a better figure than NumWorkersLimit for this pool.
+// TODO(@distractedm1nd) Find a better figure than NumWorkersLimit for this pool. Issue #970
 // namespacePool is a worker pool responsible for the goroutines spawned by getLeavesByNamespace
 var namespacePool = workerpool.New(NumWorkersLimit)
 
--- a/ipld/get_namespaced_shares.go
+++ b/ipld/get_namespaced_shares.go
@@ -4,13 +4,14 @@ import (
 	"context"
 	"sync"
 
-	"github.com/celestiaorg/celestia-node/ipld/plugin"
-	"github.com/celestiaorg/nmt"
-	"github.com/celestiaorg/nmt/namespace"
 	"github.com/gammazero/workerpool"
 	"github.com/ipfs/go-blockservice"
 	"github.com/ipfs/go-cid"
 	ipld "github.com/ipfs/go-ipld-format"
+
+	"github.com/celestiaorg/celestia-node/ipld/plugin"
+	"github.com/celestiaorg/nmt"
+	"github.com/celestiaorg/nmt/namespace"
 )
 
 // TODO(@distractedm1nd) Should the namespace pool use NumWorkersLimit? This should probably be configurable.
--- a/ipld/get_namespaced_shares.go
+++ b/ipld/get_namespaced_shares.go
@@ -2,6 +2,8 @@ package ipld
 
 import (
 	"context"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
 	"sync"
 	"sync/atomic"
 
@@ -30,6 +32,9 @@ func GetSharesByNamespace(
 	nID namespace.ID,
 	maxShares int,
 ) ([]Share, error) {
+	ctx, span := tracer.Start(ctx, "get-shares-by-namespace")
+	defer span.End()
+
 	leaves, err := getLeavesByNamespace(ctx, bGetter, root, nID, maxShares)
 	if err != nil && leaves == nil {
 		return nil, err
@@ -99,6 +104,14 @@ func getLeavesByNamespace(
 	nID namespace.ID,
 	maxShares int,
 ) ([]ipld.Node, error) {
+	ctx, span := tracer.Start(ctx, "get-leaves-by-namespace")
+	defer span.End()
+
+	span.SetAttributes(
+		attribute.String("namespace", nID.String()),
+		attribute.String("root", root.String()),
+	)
+
 	err := SanityCheckNID(nID)
 	if err != nil {
 		return nil, err
@@ -132,12 +145,14 @@ func getLeavesByNamespace(
 			if !ok {
 				// we didn't find any leaves, return nil
 				if bounds.lowest == int64(maxShares) {
-					return nil, nil
+					return nil, retrievalErr.Wait()
 				}
 
 				return leaves[bounds.lowest : bounds.highest+1], retrievalErr.Wait()
 			}
 			namespacePool.Submit(func() {
+				ctx, span := tracer.Start(ctx, "process-job")
+				defer span.End()
 				defer wg.Done()
 
 				rootH := plugin.NamespacedSha256FromCID(j.id)
@@ -151,6 +166,9 @@ func getLeavesByNamespace(
 						log.Errorw("getSharesByNamespace: could not retrieve node", "nID", nID, "pos", j.pos, "err", err)
 						return err
 					})
+					span.RecordError(err, trace.WithAttributes(
+						attribute.Int("pos", j.pos),
+					))
 					// we need to explicitly write nil at the index,
 					// for the case that the final share cannot be fetched
 					leaves[j.pos] = nil
@@ -164,6 +182,7 @@ func getLeavesByNamespace(
 					wg.Add(int64(linkCount))
 				} else if linkCount == 1 {
 					// successfully fetched a leaf belonging to the namespace
+					span.AddEvent("found-leaf")
 					leaves[j.pos] = nd
 					// we found a leaf, so we update the bounds
 					// the update routine is repeated until the atomic swap is successful
@@ -180,6 +199,10 @@ func getLeavesByNamespace(
 						// so we can return a slice of leaves in order
 						pos: j.pos*2 + i,
 					}:
+						span.AddEvent("added-job", trace.WithAttributes(
+							attribute.String("cid", lnk.Cid.String()),
+							attribute.Int("pos", j.pos*2+i),
+						))
 					case <-ctx.Done():
 						return
 					}
--- a/ipld/get_namespaced_shares.go
+++ b/ipld/get_namespaced_shares.go
@@ -2,7 +2,6 @@ package ipld
 
 import (
 	"context"
-	"golang.org/x/sync/errgroup"
 	"sync"
 	"sync/atomic"
 
@@ -14,6 +13,8 @@ import (
 	"github.com/celestiaorg/celestia-node/ipld/plugin"
 	"github.com/celestiaorg/nmt"
 	"github.com/celestiaorg/nmt/namespace"
+
+	"golang.org/x/sync/errgroup"
 )
 
 // TODO(@distractedm1nd) Find a better figure than NumWorkersLimit for this pool.
--- a/ipld/get_namespaced_shares.go
+++ b/ipld/get_namespaced_shares.go
@@ -22,11 +22,11 @@ import (
 )
 
 // TODO(@distractedm1nd) Find a better figure than NumWorkersLimit for this pool.
-// Worker pool responsible for the goroutines spawned by getLeavesByNamespace
+// namespacePool is a worker pool responsible for the goroutines spawned by getLeavesByNamespace
 var namespacePool = workerpool.New(NumWorkersLimit)
 
 // GetSharesByNamespace walks the tree of a given root and returns its shares within the given namespace.ID.
-// If a share could not be retrieved, err is not nil, and the returned array contains nil shares.
+// If a share could not be retrieved, err is not nil, and the returned array contains nil shares in place of the shares it was unable to retrieve.
 func GetSharesByNamespace(
 	ctx context.Context,
 	bGetter blockservice.BlockGetter,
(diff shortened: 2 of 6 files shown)


distractedm1nd:
Data race tests passed on my machine, so the current code is flaky: The goroutine that closes the channel when the waitgroup is finished closed the channel too early

ipld/get.go
```diff
@@ -148,42 +150,102 @@ func GetSharesByNamespace(
...
+	var leaves []result
+	mu := &sync.Mutex{}
+
+	// once the wait group is done, we can close the jobs channel to break the following loop
+	go func() {
+		wg.Wait()
+		close(jobs)
+	}()
+
+	for j := range jobs {
```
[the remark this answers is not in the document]
Wondertan: Considering we don't know how many loops we have, we can do a `do/while` style for loop which is stopped from the inside. Chan canceling routine might not be needed

ipld/get.go
```diff
@@ -148,42 +150,102 @@ func GetSharesByNamespace(
...
-		return nil, nil
-	}
-	// request the node
-	nd, err := plugin.GetNode(ctx, bGetter, root)
-	if err != nil {
-		return nil, err
+
+	// TODO: There has to be a more elegant solution
+	// bookkeeping is needed to be able to sort the leaves after the walk
+	type result struct {
```
Wondertan, line 167: Let's combine those by extending the job with the Node field. No need to allocate another structure per each node.

ipld/get.go
```diff
@@ -148,42 +150,102 @@ func GetSharesByNamespace(
...
+
+			nd, err := plugin.GetNode(ctx, bGetter, j.id)
+			if err != nil {
+				return
+			}
+
+			lnks := nd.Links()
+			// wg counter should be incremented before adding new jobs
+			if len(lnks) > 1 {
+				wg.Add(len(lnks))
```
Wondertan, line 213: Put lnLnk into a reusable var? We use it like 3 times here

ipld/get.go
```diff
@@ -148,42 +150,102 @@ func GetSharesByNamespace(
...
+			}
+
+			if len(lnks) == 1 {
```
Wondertan, line 216: Can be combined into one if/switch statement.

ipld/get.go
```diff
@@ -148,42 +150,102 @@ func GetSharesByNamespace(
...
 	}
-	// if there are some links, then traverse them
-	var out []ipld.Node
-	for _, lnk := range nd.Links() {
-		nds, err := GetLeavesByNamespace(ctx, bGetter, lnk.Cid, nID)
-		if err != nil {
-			return out, err
+
+	if len(leaves) > 0 {
+		sort.SliceStable(leaves, func(i, j int) bool {
```
Wondertan, line 237: Did you look at the difference between `sort.Slice` and `sort.SliceStable`

ipld/get.go
```diff
@@ -148,42 +150,102 @@ func GetSharesByNamespace(
...
+			lnks := nd.Links()
+			// wg counter should be incremented before adding new jobs
+			if len(lnks) > 1 {
+				wg.Add(len(lnks))
+			}
+
+			if len(lnks) == 1 {
+				mu.Lock()
+				leaves = append(leaves, result{nd, j.pos})
+				mu.Unlock()
```
Wondertan, line 219: We should ideally avoid contention between routines in the worker pool. It can become an issue when the pool reaches the limit of routine. We can avoid it if we write the resulting nodes into some preallocated slice. However, we don't know the exact amount of nodes we will fetch, so we can preallocate more __pointers__ to nodes than is needed. The more here is the len of either Row or Col, similar to what we do in `GetShares`, so we might extend the func with one more param. On return, we can just slice away the resulting slice `(results[:fetched])`

Wondertan (commented):
Great result, it's working and I haven't found any `substantial` bugs! Also, glad that our existing testing covers things well. Leaving comments on how we can improve the implementation further.

ipld/get.go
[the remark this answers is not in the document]
distractedm1nd: Yeah, will wrap with context :)
Regarding the do/while, that was my first approach but the issue is that the logic deciding if the loop should break depends on the `WaitGroup` being finished. The waitgroup's counter isn't exported, so what it would probably look like is wrapping the wg in a new struct where we keep track of the counter ourselves. Should I try that approach?

ipld/get.go
distractedm1nd: Can't the number of nodes in the namespace be larger than Row/Col?

Wondertan: It can, but then the namespace is spread over multiple rows. And downloading from multiple rows is tackled on the layer above. Within one row, like in our case, it's not possible to have more shares that in the row.

ipld/get.go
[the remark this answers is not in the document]
Wondertan: Yes, the described approach should work, and its cheaper than spawning another goroutine

ipld/get.go
distractedm1nd: I totally missed that, and it explains a lot - thanks

ipld/get.go
distractedm1nd: Yeah, I guess we never have any with the same position anyways, switching to `sort.Slice`

renaynay (changes requested):
Some initial feedback from our sync discussion -- 

_Nice to have_: doc ipld/GetSharesByNamespace and shares/GetSharesByNamespace to document what the diff is

`shares/GetSharesByNamespace` --> iterating over the square's row roots
`ipld/GetSharesByNamespace` --> walking the tree of the given root

ipld/get_namespaced_shares.go
```diff
@@ -0,0 +1,188 @@
...
+		id  cid.Cid
+		pos int
+	}
+
+	// we don't know where in the tree the leaves in the namespace are,
+	// so we keep track of the bounds to return the correct slice
+
+	// maxShares acts as a sentinel to know if we find any leaves
+	bounds := fetchedBounds{maxShares, 0}
+	mu := sync.Mutex{}
```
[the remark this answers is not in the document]
distractedm1nd: moved into bounds struct

Wondertan (changes requested):
The logic is again correct, however, the current approach with locks will have a harmful effect on performance when there is a lot of data in practice and this should be fixed

ipld/get_namespaced_shares.go
```diff
@@ -0,0 +1,186 @@
...
+	jobs <- &job{id: root}
+
+	// the wg counter cannot be pre-allocated either, it is incremented with each job
+	wg := wrappedWaitGroup{sync.WaitGroup{}, sync.Mutex{}, 0}
+	wg.Add(1)
+
+	// we overallocate space for leaves since we do not know how many we will find
+	// on the level above, the length of the Row is passed in as maxShares
+	leaves := make([]ipld.Node, maxShares)
+	leafLk := sync.Mutex{}
```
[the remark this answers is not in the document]
distractedm1nd: What do you mean by this? Like I can just declare with var?

ipld/get_namespaced_shares.go
```diff
@@ -0,0 +1,186 @@
...
+					}:
+					case <-ctx.Done():
+						return
+					}
+				}
+			})
+		case <-ctx.Done():
+			return nil, nil
+		// default case is needed to prevent blocking
+		default:
```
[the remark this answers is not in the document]
distractedm1nd: Ok yup, i'll try again - couldn't figure out a way to get it working yet without the default case. But that was before I was able to check IsWorking

distractedm1nd:
Won't pass the race test because there is a race condition with how I close the jobs channel still

ipld/get_namespaced_shares.go
```diff
@@ -0,0 +1,186 @@
...
+	w.wg.Wait()
+}
+
+func (w *wrappedWaitGroup) IsWorking() bool {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+	return w.counter > 0
+}
+
+type fetchedBounds struct {
```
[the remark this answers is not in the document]
distractedm1nd: I prefer the separation for readability

ipld/get_namespaced_shares.go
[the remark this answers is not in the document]
distractedm1nd: @Wondertan Ok, I definitely made some progress with it if you check the latest commit. We now always block on jobs, and close the channel when the wg counter reaches 0. This means we only check if we are done once per job.

One issue remains to avoid a potential race condition: What is the best pattern to close a channel atomically? Is a mutex okay here?

ipld/get_namespaced_shares.go
```diff
@@ -0,0 +1,185 @@
...
+				}
+			})
+		case <-ctx.Done():
+			return nil, nil
+		}
+	}
+
+	// we didn't find any leaves, return nil
+	if bounds.lowest == int64(maxShares) {
+		return nil, nil
```
vgonkivs, line 181: I think we should have an error in this case. Something similar to https://github.com/celestiaorg/celestia-node/blob/2cc807743a16f4a955f6d9c819d829f98e6fbbed/header/interface.go#L65

ipld/get_namespaced_shares.go
```diff
@@ -0,0 +1,185 @@
...
+				defer wg.Done(jobs)
+
+				rootH := plugin.NamespacedSha256FromCID(j.id)
+				if nID.Less(nmt.MinNamespace(rootH, nID.Size())) || !nID.LessOrEqual(nmt.MaxNamespace(rootH, nID.Size())) {
+					return
+				}
+
+				nd, err := plugin.GetNode(ctx, bGetter, j.id)
+				if err != nil {
+					return
```
vgonkivs, line 141: should we just skip this error?

ipld/get_namespaced_shares.go
```diff
@@ -0,0 +1,185 @@
...
+	wg      sync.WaitGroup
+	counter int64
+}
+
+func (w *wrappedWaitGroup) Add(count int64) {
+	w.wg.Add(int(count))
+	atomic.AddInt64(&w.counter, count)
+}
+
+func (w *wrappedWaitGroup) Done(jobs chan *job) {
```
Wondertan, line 55: nit: the chan can be a field on the`wrappedWaitGroup`, as (1) the same chan is always passed and (2) the chan is constructed before the wg in `getLeavesByNamespace`

ipld/get_namespaced_shares.go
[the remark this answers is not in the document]
Wondertan: I think the current logic in wrappedWaitGroup closes the channel well enough and won't cause any issues with how it is ___currently___ used. If wrappedWaitGroup is incorrectly used, then yes, it may cause issues, which can be prevented by additional safety measures, but as long wrappedWaitGroup is not abstracted and private, I think it is fine to keep things the way they are currently.

ipld/get_namespaced_shares.go
```diff
@@ -0,0 +1,185 @@
...
+	if lowest := atomic.LoadInt64(&b.lowest); index < lowest {
+		ok = ok && atomic.CompareAndSwapInt64(&b.lowest, lowest, index)
+	}
+	// not an `else if` because an element can be both the lower and higher bound
+	// for example, if there is only one share in the namespace
+	if highest := atomic.LoadInt64(&b.highest); index > highest {
+		ok = ok && atomic.CompareAndSwapInt64(&b.highest, highest, index)
+	}
+	return ok
+}
```
Wondertan, line 87: nit: I totally see the code cleanness idea here, but it would be even better from a struct API standpoint to not have to return a bool and make the use of it to loop it until it's true. 
* Two `for`s inside and making `Update` atomically guarantee the success of the update operation > making the user worry about calling for over it.
* Also we avoid idle `lowest` loadings, when `highest` failed to be swapped in the first attempt

ipld/get_namespaced_shares.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+}
+
+func (w *wrappedWaitGroup) Wait() {
+	w.wg.Wait()
+}
+
+func (w *wrappedWaitGroup) IsWorking() bool {
+	return atomic.LoadInt64(&w.counter) > 0
+}
```
Wondertan, line 69: Unused code which can be removed

ipld/get_namespaced_shares.go
Wondertan: History note: we had something similar before but decided to get rid of it in favour of making `len(0)` returned slice a success case when @tzdybal chimed in during development of DA API

Wondertan: I don't feel strongly about any of the options, but we might reconsider this. @vgonkivs check out higher-level GetShareByNamespace to see how introduction of such error may affect the code

distractedm1nd: I don't think we should return an error. It sounds like a normal case that querying a namespace will have no shares

ipld/get_namespaced_shares.go
Wondertan: Hmm, good point. @distractedm1nd likely followed the approach taken in `GetShares`. In `GetShares` we can ignore it as we don't need all shares, but only 1/4 at minimum. Here, we have a different situation as we need to get everything without any exceptions, so ignoring errors might have undesirable effects, which we should think through. At least, we should log that out.

distractedm1nd: Totally missed that! I would also be leaning towards logging, but I don't know what effects leaving a share empty in the output will have downstream. Regardless, stopping the whole routine because a single share cannot be retrieved feels wrong - but is it even helpful for applications to retrieve a response which does not contain all shares in the namespace?

ipld/get_namespaced_shares.go
```diff
@@ -0,0 +1,188 @@
...
+		}
+	}
+	// not an `else if` because an element can be both the lower and higher bound
+	// for example, if there is only one share in the namespace
+	for ok := false; !ok; {
+		ok = true
+		if highest := atomic.LoadInt64(&b.highest); index > highest {
+			ok = atomic.CompareAndSwapInt64(&b.highest, highest, index)
+		}
+	}
```
Wondertan, line 97:
```suggestion
	ok := false
	for !ok {
		if lowest := atomic.LoadInt64(&b.lowest); index < lowest {
			ok = atomic.CompareAndSwapInt64(&b.lowest, lowest, index)
		}
	}

	ok = false
	for !ok {
		if highest := atomic.LoadInt64(&b.highest); index > highest {
			ok = atomic.CompareAndSwapInt64(&b.highest, highest, index)
		}
	}
```
Same logic, while takes less cod and `ok` is allocated once.

ipld/get_namespaced_shares.go
```diff
@@ -0,0 +1,188 @@
...
+
+// wrappedWaitGroup is needed because waitgroups do not expose their internal counter,
+// and we don't know in advance how many jobs we will have to wait for.
+type wrappedWaitGroup struct {
+	wg      sync.WaitGroup
+	jobs    chan *job
+	counter int64
+}
+
+func createWrappedWaitGroup(jobs chan *job) *wrappedWaitGroup {
```
Wondertan, line 51:
```suggestion
func newWrappedWaitGroup(jobs chan *job) *wrappedWaitGroup {
```

ipld/get_namespaced_shares.go
```diff
@@ -0,0 +1,188 @@
...
+	wg := &wrappedWaitGroup{
+		wg:      sync.WaitGroup{},
+		jobs:    jobs,
+		counter: 0,
```
Wondertan, line 55: optional nit: redundant

ipld/get_namespaced_shares.go
```diff
@@ -0,0 +1,188 @@
...
+// and we don't know in advance how many jobs we will have to wait for.
+type wrappedWaitGroup struct {
+	wg      sync.WaitGroup
+	jobs    chan *job
+	counter int64
+}
+
+func createWrappedWaitGroup(jobs chan *job) *wrappedWaitGroup {
+	wg := &wrappedWaitGroup{
+		wg:      sync.WaitGroup{},
```
Wondertan, line 53: optional nit: redundant

ipld/get_namespaced_shares.go
Wondertan: This can be generified and extracted into `libs` as a utility 😃 . I see this being reusable. Up to you to decide whether to extract now, later, or not at all.

ipld/get_namespaced_shares.go
Wondertan: It is the right question to ask. The downstream blocks forever until it either finds a share/block or the ctx is canceled. There can be thought errors related to decoding, i.e, bad data was requested. The hash/cid matches the retrieved content. However, the content itself is broken. This can only happen due to some vulnerability we don't know about or a validator set committing incorrectly modified user data in the block, which honest majority issue that should be fundamentally fixed by `MessageInclusionFraudProof`, afaiu(cc @adlerjohn). 

From all of this, I would not stop the downloading process in case there is an error, but I would definitely:
* Log the error
* Emit error event in a trace(will discuss this later)
* Save the first rcvd error(with `sync.Once`) until the downloading finishes and return it to the user:
	* Either by having both return values being non-nil
	* Or by making a special error type with two fields: inner error that happened during share decoding/downloading and incomplete downloaded data
	* Both ways are fine IMO and both needs to be properly documented in the method

ipld/get_namespaced_shares.go
distractedm1nd: This would not work, because the if statement is not always entered

distractedm1nd: Was my first attempt as well!

Wondertan: Oh right!

ipld/get_namespaced_shares.go
distractedm1nd: Hmmm. I feel like the inclusion of the jobs channel makes this not very suitable for re-use. I would wait to extract this until we want to use it somewhere else, since it smells like premature abstraction

distractedm1nd: I am just going to get rid of this method, I imagined it better than it actually is 😅

ipld/get_namespaced_shares.go
Wondertan: Ok, so the current code still feels overcomplicated, so I updated the suggestion which obsoleted the ok completely.

ipld/get_namespaced_shares.go
Wondertan: > I feel like the inclusion of the jobs channel makes this not very suitable for re-use.
That's why generify. 

>  I would wait to extract this until we want to use it somewhere else, since it smells like premature abstraction
Yeah. I've written similar code before, which closes the chan once the atomic counter is zero, and this reminds me of it and makes me wanna extract it, but as I said, completely up to you.

ipld/get_namespaced_shares.go
Wondertan: @distractedm1nd, you might resolved it before seeing the update

Wondertan: Updated again with the most succinct form of it. Now there is no other way to compress it even more

ipld/get_namespaced_shares.go
vgonkivs: One more suggestion: Worker pool could be changed to error group with limiting the amount of routines. https://cs.opensource.google/go/x/sync/+/886fb937:errgroup/errgroup.go;l=123

distractedm1nd: Latest change uses an errgroup, it seemed to be the most readable option. Still will have to add the trace after syncing with @Wondertan next week. What would be the best way to test this? (What is the easiest hack to influence `GetNode` to return an error)?

distractedm1nd: One question I think should be thought through carefully is how we return the data when it's only partial. Should the leaves be nil in the array, or should they be excluded completely? This is also relevant to the caller, `GetSharesByNamespace`. By including nils, we are directly forcing apps to handle partial data (which is good), BUT the question is how helpful is this data? If you know data at an index is missing, an app could write logic to retrieve it itself with a little tree traversal, but will this really be done? I assume apps will just retry when something is missing.

distractedm1nd: Just pushed a commit to show how it would look with nil spots, kinda ugly, but just POC

ipld/get_namespaced_shares.go
```diff
@@ -0,0 +1,192 @@
...
+
+				nd, err := plugin.GetNode(ctx, bGetter, j.id)
+				if err != nil {
+					retrievalErr.Go(func() error {
+						log.Errorw("getSharesByNamespace: could not retrieve node", "nID", nID, "pos", j.pos, "err", err)
+						return err
+					})
+					// we need to explicitly write nil at the index,
+					// for the case that the final share cannot be fetched
+					leaves[j.pos] = nil
```
vgonkivs, line 156: Should not be it `nil` by default?

ipld/get_namespaced_shares.go
distractedm1nd: We decided to switch to `sync.Once`, and I will leave the return of the partial data because it doesn't really change much (not as big of a deal as I thought it was, since the blockservice caches blocks)

ipld/get_namespaced_shares.go
```diff
@@ -0,0 +1,226 @@
...
+	w.wg.Add(int(count))
+	atomic.AddInt64(&w.counter, count)
+}
+
+func (w *wrappedWaitGroup) Done() {
+	w.wg.Done()
+	numRemaining := atomic.AddInt64(&w.counter, -1)
+
+	// Close channel if this job was the last one
+	if numRemaining == 0 {
```
Wondertan, line 73:
```suggestion
	// Close channel if this job was the last one
	if atomic.AddInt64(&w.counter, -1) == 0 {
```

Wondertan: Ah, readability? The comment is good enough.

ipld/get_namespaced_shares.go
```diff
@@ -0,0 +1,226 @@
...
+							attribute.String("cid", newJob.id.String()),
+							attribute.Int("pos", newJob.pos),
+						))
+					case <-ctx.Done():
+						return
+					}
+				}
+			})
+		case <-ctx.Done():
+			return nil, retrievalErr
```
Wondertan, line 223:
```suggestion
			return nil, ctx.Err()
```

ipld/get_namespaced_shares.go
```diff
@@ -0,0 +1,226 @@
...
+	leaves := make([]ipld.Node, maxShares)
+
+	// stillWorking will be set to false when the waitgroup counter reaches 0, closing the jobs channel
+	for {
+		select {
+		case j, ok := <-jobs:
+			if !ok {
+				// we didn't find any leaves, return nil
+				if bounds.lowest == int64(maxShares) {
+					return nil, retrievalErr
```
Wondertan, line 152: We should document the case where both nils are returned here. When there is no namespace under the given root

ipld/get_namespaced_shares.go
```diff
@@ -0,0 +1,226 @@
...
+					leaves[j.pos] = nd
+					// we found a leaf, so we update the bounds
+					// the update routine is repeated until the atomic swap is successful
+					bounds.Update(int64(j.pos))
+					return
+				}
+
+				// this node has links in the namespace, so keep walking
+				for i, lnk := range links {
+
```
Wondertan, line 193:
```suggestion
```

ipld/get_namespaced_shares.go
```diff
@@ -0,0 +1,226 @@
...
+					// if the link's nID isn't in range we don't need to create a new job for it
+					jobNid := plugin.NamespacedSha256FromCID(newJob.id)
+					if nID.Less(nmt.MinNamespace(jobNid, nID.Size())) || !nID.LessOrEqual(nmt.MaxNamespace(jobNid, nID.Size())) {
+						continue
+					}
+
+					// by passing the previous check, we know we will have one more node to process
+					// note: it is important to increase the counter before sending to the channel
+					wg.Add(1)
+
```
Wondertan, line 210:
```suggestion
```

ipld/get_namespaced_shares.go
```diff
@@ -0,0 +1,226 @@
...
+	if numRemaining == 0 {
+		close(w.jobs)
+	}
+}
+
+func (w *wrappedWaitGroup) Wait() {
+	w.wg.Wait()
+}
+
+type fetchedBounds struct {
```
Wondertan, line 82: Optional refactoring: Extending this struct with the actual nodes/leaves seems reasonable. E.g. when you do `Update`, you can also pass the node as well. And then, It will also have a method `Result`(or else) that encapsulates slicing.

ipld/get_namespaced_shares.go
```diff
@@ -0,0 +1,226 @@
...
+				if err != nil {
+					singleErr.Do(func() {
+						log.Errorw("getSharesByNamespace: could not retrieve node", "nID", nID, "pos", j.pos, "err", err)
+						retrievalErr = err
+					})
+					span.RecordError(err, trace.WithAttributes(
+						attribute.Int("pos", j.pos),
+					))
+					// we still need to update the bounds
+					bounds.Update(int64(j.pos))
```
Wondertan, line 174: Hmm, can this corrupt the resulting boundary if this fails for any inner node?

Wondertan (approved):
Great job, only doc stuff, and optional nits

ipld/get_namespaced_shares.go
distractedm1nd: The compiler takes care of this optimization, no? I find the former more readable, @renaynay thoughts?

ipld/get_namespaced_shares.go
distractedm1nd: I feel like keeping the "separation of concerns" principle is better here, so I'll keep it as it is, but I'm on the fence about this suggestion - both solutions are nice in their own ways

ipld/get_namespaced_shares.go
distractedm1nd: Can you give some more clarification as to what you mean by corruption? If it fails for an inner node, the array may possibly not return as many children as it would (it is not extended outwards), but it wouldn't corrupt anything no?

The reason the update is even needed here is for an edge case I found during testing: If the **last** node is nil, then the bounds are not updated and it is not kept in the result array.

ipld/get_namespaced_shares.go
Wondertan: Are you sure? If it does then I am fine, but IDK
> The compiler takes care of this optimization, no?

Wondertan: Ok, Ryan proved me this in sync so resolving

renaynay (commented):
I am currently in progress testing you branch against mamaki with share submission + retrieval.. standby

ipld/get_namespaced_shares.go
```diff
@@ -0,0 +1,226 @@
...
+	wg.Add(1)
+
+	var singleErr sync.Once
+	var retrievalErr error
+
+	// we overallocate space for leaves since we do not know how many we will find
+	// on the level above, the length of the Row is passed in as maxShares
+	leaves := make([]ipld.Node, maxShares)
+
+	// stillWorking will be set to false when the waitgroup counter reaches 0, closing the jobs channel
```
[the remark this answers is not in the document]
distractedm1nd: old comment, removing

ipld/get_namespaced_shares.go
```diff
@@ -0,0 +1,226 @@
...
+				defer span.End()
+				defer wg.Done()
+
+				// if an error is likely to be returned or not depends on
+				// the underlying impl of the blockservice, currently it is not a realistic probability
+				nd, err := plugin.GetNode(ctx, bGetter, j.id)
+				if err != nil {
+					singleErr.Do(func() {
+						log.Errorw("getSharesByNamespace: could not retrieve node", "nID", nID, "pos", j.pos, "err", err)
+						retrievalErr = err
```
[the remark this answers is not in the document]
distractedm1nd: Hmmm. It was only important that the error is only set once, so I am moving the log outside of the `sync.Once`

ipld/get_namespaced_shares.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	return shares, err
+}
+
+// wrappedWaitGroup is needed because waitgroups do not expose their internal counter,
+// and we don't know in advance how many jobs we will have to wait for.
+type wrappedWaitGroup struct {
+	wg      sync.WaitGroup
+	jobs    chan *job
```
walldiss, line 59: Maybe we can get away by sharing over chan _copy_ of the struct instead of _pointer_? It could be dangerous to have access by pointer to the same struct from different routines

distractedm1nd: Each job is only accessed by a single goroutine, so it is safe here

ipld/get_namespaced_shares.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+type fetchedBounds struct {
+	lowest  int64
+	highest int64
+}
+
+func (b *fetchedBounds) Update(index int64) {
+	lowest := atomic.LoadInt64(&b.lowest)
+	for index < lowest && !atomic.CompareAndSwapInt64(&b.lowest, lowest, index) {
```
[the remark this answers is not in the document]
distractedm1nd: Can you check if the new doc is okay? @renaynay

ipld/get_namespaced_shares.go
distractedm1nd: @Wondertan does the new comment doc this as intended?

ipld/get_namespaced_shares.go
```diff
@@ -0,0 +1,226 @@
...
+	"github.com/ipfs/go-blockservice"
+	"github.com/ipfs/go-cid"
+	ipld "github.com/ipfs/go-ipld-format"
+
+	"github.com/celestiaorg/celestia-node/ipld/plugin"
+
+	"github.com/celestiaorg/nmt/namespace"
+)
+
+// TODO(@distractedm1nd) Find a better figure than NumWorkersLimit for this pool.
```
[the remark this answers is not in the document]
distractedm1nd: Will do once merged

ipld/get_namespaced_shares.go
```diff
@@ -0,0 +1,226 @@
...
+// GetSharesByNamespace walks the tree of a given root and returns its shares within the given namespace.ID.
+// If a share could not be retrieved, err is not nil, and the returned array contains nil shares.
+func GetSharesByNamespace(
+	ctx context.Context,
+	bGetter blockservice.BlockGetter,
+	root cid.Cid,
+	nID namespace.ID,
+	maxShares int,
+) ([]Share, error) {
+	ctx, span := tracer.Start(ctx, "get-shares-by-namespace")
```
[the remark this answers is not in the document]
distractedm1nd: It modifies it (adding the current span to the context)

ipld/get_namespaced_shares.go
walldiss: Extra point: As a value it can be stack allocated, thus reducing load on GC and heap size.
It will increase performance by small margin. So feel free to resolve

distractedm1nd: @Wondertan  thoughts?

Wondertan: Check out with `
go build -gcflags "-m" file.go
`

Wondertan: It suggests if a value is allocated on the heap and other potential small optimisations

ipld/get_namespaced_shares.go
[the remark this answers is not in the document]
renaynay: Ack

ipld/get_namespaced_shares.go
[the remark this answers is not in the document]
renaynay: Okay just don't forget to update comment with a issue #

ipld/get_namespaced_shares.go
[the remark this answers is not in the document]
renaynay: awesome

ipld/get_namespaced_shares.go
```diff
@@ -0,0 +1,226 @@
...
+					return
+				}
+
+				links := nd.Links()
+				linkCount := uint64(len(links))
+				// wg counter needs to be incremented **before** adding new jobs
+				if linkCount == 1 {
+					// successfully fetched a leaf belonging to the namespace
+					span.AddEvent("found-leaf")
+					leaves[j.pos] = nd
```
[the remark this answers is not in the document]
renaynay: ack

ipld/helpers.go
```diff
@@ -1,6 +1,16 @@
 package ipld
 
-import "fmt"
+import (
+	"github.com/ipfs/go-cid"
+
+	"fmt"
+)
+
+// job represents an encountered node to investigate during the `GetShares` and `GetSharesByNamespace` routines
```
renaynay, line 9:
```suggestion
// job represents an encountered node to investigate during the `GetShares` and `GetSharesByNamespace` routines.
```

ipld/helpers.go
```diff
@@ -1,6 +1,16 @@
 package ipld
 
-import "fmt"
+import (
+	"github.com/ipfs/go-cid"
+
+	"fmt"
```
renaynay, line 6: format pls

renaynay (approved):
minor nits, awesome work

ipld/get_namespaced_shares.go
distractedm1nd: Thanks! Didnt catch it, resolving

Issue #790, opened 2022-06-02
params: `BlockTimeFor` func reporting block time of a network


Wondertan:
There should be a way for the Node to know what block time of a network the node operates on. Options:
* The node can calc an avrg time between headers it rcvs from the network
* Or we can hardcode another registry in `params` of block times per each network we support.

The second option simplifies logic for the code that works with block time, as it's constant and does not change in the runtime. And there is no clear reason to complicate things with the first option yet, so we should go with the second.

It can be implemented similarly to existing registry for network bootstrappers.

PR #978, opened 2022-08-01
header: Extract `Head` method into separate `Head` interface, make `Syncer` implement it 

renaynay:
This PR extracts `Head()` method into a separate interface and makes `Syncer` implement it.

Prerequisite for #933 to restrict the State service's access to headers to the only methods it really needs (Head).

### DIFF
--- a/header/interface.go
+++ b/header/interface.go
@@ -104,8 +104,7 @@ type Store interface {
 // Getter contains the behavior necessary for a component to retrieve
 // headers that have been processed during header sync.
 type Getter interface {
-	// Head returns the ExtendedHeader of the chain head.
-	Head(context.Context) (*ExtendedHeader, error)
+	Head
 
 	// Get returns the ExtendedHeader corresponding to the given hash.
 	Get(context.Context, tmbytes.HexBytes) (*ExtendedHeader, error)
@@ -115,4 +114,12 @@ type Getter interface {
 
 	// GetRangeByHeight returns the given range [from:to) of ExtendedHeaders.
 	GetRangeByHeight(ctx context.Context, from, to uint64) ([]*ExtendedHeader, error)
+}
+
+// Head contains the behavior necessary for a component to retrieve
+// the chain head. Note that "chain head" is subjective to the component
+// reporting it.
+type Head interface {
+	// Head returns the ExtendedHeader of the chain head.
+	Head(context.Context) (*ExtendedHeader, error)
 }
--- a/header/sync/sync.go
+++ b/header/sync/sync.go
@@ -96,7 +96,7 @@ func (s *Syncer) WaitSync(ctx context.Context) error {
 // Head tries to return the Syncer's view of the objective head of the
 // network.
 func (s *Syncer) Head(ctx context.Context) (*header.ExtendedHeader, error) {
-	return s.trustedHead(ctx)
+	return s.objectiveHead(ctx)
 }
 
 // State collects all the information about a sync.
@@ -130,16 +130,18 @@ func (s *Syncer) State() State {
 	return state
 }
 
-// trustedHead returns the latest known trusted header that is within the trusting period.
-func (s *Syncer) trustedHead(ctx context.Context) (*header.ExtendedHeader, error) {
-	// check pending for trusted header and return it if applicable
-	// NOTE: Pending cannot be expired, guaranteed
+func (s *Syncer) subjectiveHead(ctx context.Context) (*header.ExtendedHeader, error) {
 	pendHead := s.pending.Head()
 	if pendHead != nil {
 		return pendHead, nil
 	}
 
-	sbjHead, err := s.store.Head(ctx)
+	return s.store.Head(ctx)
+}
+
+// objectiveHead returns the latest known trusted header that is within the trusting period.
+func (s *Syncer) objectiveHead(ctx context.Context) (*header.ExtendedHeader, error) {
+	sbjHead, err := s.subjectiveHead(ctx)
 	if err != nil {
 		return nil, err
 	}
@@ -150,39 +152,29 @@ func (s *Syncer) trustedHead(ctx context.Context) (*header.ExtendedHeader, error
 		return sbjHead, nil
 	}
 
-	// otherwise, attempt to request head from a trustedPeer or, in other words, do
-	// automatic subjective initialization
+	// otherwise, attempt to request head from a trustedPeer
 	objHead, err := s.exchange.Head(ctx)
 	if err != nil {
 		return nil, err
 	}
 
-	// basic sanity check to ensure that head from trustedPeer is more current than
-	// subjective head
-	if !sbjHead.IsBefore(objHead) {
-		log.Warnw("received known header",
-			"height", objHead.Height,
-			"hash", objHead.Hash())
-		// return most current known header to the Syncer
-		return sbjHead, nil
+	if sbjHead.IsExpired() {
+		// trust objective head from trusted peer as new sync target
+		// as our subjective head is expired
+		// or, in other words, do automatic subjective initialization
+		s.pending.Add(objHead)
+		return objHead, nil
 	}
 
-	// if the subjective head is not yet expired, attempt to verify the objective head
-	// against the subjective head
-	if !sbjHead.IsExpired() {
-		// verify head returned from trustedPeer against subjective head
-		res := s.newHead(sbjHead, objHead)
-		if res == pubsub.ValidationAccept {
-			// return new head
-			return objHead, nil
-		}
-		// objHead was either ignored or rejected, return most current known
-		// header to the Syncer
-		return sbjHead, nil
+	// verify head returned from trustedPeer against subjective head
+	res := s.newHead(sbjHead, objHead)
+	if res == pubsub.ValidationAccept {
+		// return new head
+		return objHead, nil
 	}
-	// otherwise, trust objective head from trusted peer as new sync target
-	s.pending.Add(objHead)
-	return objHead, nil
+	// objHead was either ignored or rejected, return most current known
+	// header to the Syncer
+	return sbjHead, nil
 }
 
 // wantSync will trigger the syncing loop (non-blocking).
@@ -207,13 +199,13 @@ func (s *Syncer) syncLoop(ctx context.Context) {
 
 // sync ensures we are synced up to any trusted header.
 func (s *Syncer) sync(ctx context.Context) {
-	trstHead, err := s.trustedHead(ctx)
+	objHead, err := s.objectiveHead(ctx)
 	if err != nil {
-		log.Errorw("getting trusted head", "err", err)
+		log.Errorw("getting objective head", "err", err)
 		return
 	}
 
-	s.syncTo(ctx, trstHead)
+	s.syncTo(ctx, objHead)
 }
 
 // processIncoming processes new processIncoming Headers, validates them and stores/caches if applicable.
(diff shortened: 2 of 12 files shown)


header/sync/sync.go
```diff
@@ -88,6 +88,23 @@ func (s *Syncer) WaitSync(ctx context.Context) error {
...
+		return pendHead, nil
+	}
+	// if no pending head, eagerly fetch head from network
+	objHead, err := s.exchange.Head(ctx)
+	if err != nil {
+		return nil, err
+	}
+	s.pending.Add(objHead)
+	return objHead, nil
+}
```
Wondertan, line 106: This should work; however: 
* we should think about not making Head requests via Exchange in a happy case when pending is empty and the node is fully synced.
	* When the node is fully synced, new gossiped headers are written directly into the store, avoiding the pending cache.
* We might need to trigger sync if we added a new header to the pending cache so that node does not wait for gossiped headers, as it has a sync target already

Wondertan: Let's sync on those

node/services/config.go
```diff
@@ -22,6 +22,8 @@ type Config struct {
 	// Note: The trusted does *not* imply Headers are not verified, but trusted as reliable to fetch headers
 	// at any moment.
 	TrustedPeers []string
+	// BlockTime is the block time of the network.
+	BlockTime time.Duration
```
renaynay, line 26: I am still kind of iffy on this being a config value. While it doesn't really impact anything other than the frequency at which the syncer requests a new objective head from its trustedPeer, if we do end up using this config param elsewhere in the code, we would have to be very cautious on how it is used. Wdyt @Wondertan ?

params/network.go
```diff
@@ -13,6 +14,8 @@ const (
 	// Private can be used to set up any private network, including local testing setups.
 	// Use CELESTIA_PRIVATE_GENESIS env var to enable Private by specifying its genesis block hash.
 	Private Network = "private"
+	// TODO @renaynay @Wondertan: Set this once it's agreed upon, ideally this could point at a core const
```
renaynay, line 17: https://github.com/celestiaorg/celestia-node/issues/790

header/sync/sync.go
```diff
@@ -133,16 +145,23 @@ func (s *Syncer) trustedHead(ctx context.Context) (*header.ExtendedHeader, error
...
+	// otherwise, attempt to request head from a trustedPeer or, in other words, do
+	// automatic subjective initialization
 	objHead, err := s.exchange.Head(ctx)
 	if err != nil {
 		return nil, err
 	}
+	// ensure that head returned from trustedPeer is more recent than subjective head
+	// as it is possible that the trustedPeer's reported head is not current.
+	if objHead.Height < sbj.Height {
+		return nil, fmt.Errorf("trusted peer failed to return current network head")
```
renaynay, line 163: return sbj head here bc it's highest we know of at that point

header/sync/sync.go
```diff
@@ -133,16 +145,23 @@ func (s *Syncer) trustedHead(ctx context.Context) (*header.ExtendedHeader, error
...
-	// otherwise, request head from a trustedPeer or, in other words, do automatic subjective initialization
+	// otherwise, attempt to request head from a trustedPeer or, in other words, do
+	// automatic subjective initialization
 	objHead, err := s.exchange.Head(ctx)
 	if err != nil {
 		return nil, err
 	}
+	// ensure that head returned from trustedPeer is more recent than subjective head
+	// as it is possible that the trustedPeer's reported head is not current.
+	if objHead.Height < sbj.Height {
```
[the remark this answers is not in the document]
renaynay: make a test for this.

node/services/config.go
Wondertan: We should not make it part of the config value but a part of `params` pkg(#790). Taken from the params,  the value then gets pass to the Syncer

Wondertan: Similarly, I don't think we should make it an option on the node, as by now.

renaynay: pls check: https://github.com/celestiaorg/celestia-node/pull/978/commits/9d897749c65fc3fb5e2e0fc8b80cecfd7bcb6bae

node/testing.go
```diff
@@ -42,6 +42,9 @@ func TestNode(t *testing.T, tp Type, opts ...Option) *Node {
 		WithNetwork(params.Private),
 		WithRPCPort("0"),
 		WithKeyringSigner(TestKeyringSigner(t)),
+		func(sets *settings) { // override the block time
+			sets.cfg.Services.BlockTime = time.Microsecond
+		},
```
distractedm1nd, line 47: Couldn't this be pulled out as a `WithBlockTime` in `config_opts.go`?

header/interface.go
```diff
@@ -116,3 +115,11 @@ type Getter interface {
 	// GetRangeByHeight returns the given range [from:to) of ExtendedHeaders.
 	GetRangeByHeight(ctx context.Context, from, to uint64) ([]*ExtendedHeader, error)
 }
+
+// Head contains the behavior necessary for a component to retrieve
+// the chain head. Note that "chain head" is subjective to the component
+// reporting it.
+type Head interface {
```
distractedm1nd, line 122: super unimportant nit, especially since I'm lacking context, but could there be a better name for the interface itself? For example, could something along the lines of `HeadRetriever` be more clear? But if we do this pattern in other places as well I would stick with it for consistency

header/verify.go
```diff
@@ -26,6 +26,11 @@ func (eh *ExtendedHeader) IsExpired() bool {
 	return !expirationTime.After(time.Now())
 }
 
+// IsRecent checks if header is recent against the given blockTime.
+func (eh *ExtendedHeader) IsRecent(blockTime time.Duration) bool {
+	return time.Now().Sub(eh.Time) <= blockTime // TODO @renaynay: should we allow for a 5-10 block drift here?
```
distractedm1nd, line 31: What would speak against allowing for a drift?

node/testing.go
renaynay: Originally I had this option available in `config_opts.go` but we shouldn't really expose this as an option configurable to users. We mainly need to override the default network block time for testing purposes. Having it as a config value in general is a bit iffy because if we do end up using this value somewhere else in the codebase, we have to be very careful how it's passed in / make sure there are no discrepancies, etc.

header/interface.go
renaynay: We have an interface called `Getter`

We could call it `HeadGetter` but I like the simplicity of it. I'm not too biased tho.

header/verify.go
renaynay: So the argument **for allowing a certain amount of drift** is that this method determines how often the `trustedHead` method would eagerly request an objective head from a trusted peer (in the case that pending is empty for some reason, not really likely though). It could essentially render our subjective head "useless" (depending on the block time of the network, which hasn't yet been definitively decided) if we have a very tight constraint that sbjHead cannot even be 1 block behind -- by "useless", I mean it wouldn't really even make sense to fetch sbjHead from store if we're very likely going to toss it and ask our trusted peers for their Head instead.

The argument **against drift** is that it could (likely not though because we still have pending cache that is always checked first) return a slightly outdated head for a service that is relying on a head that is as current as possible (the state service for example). If you take a look at #933, when we request a balance from our core node connection, it will use the Syncer's latest known tip of chain to determine at which height it will request the balance from. This means if the Syncer returns a header that is a few blocks in the past, the balance could be outdated. However, all of this is still subjective to the Syncer as the Syncer is the one service inside of node that is most up-to-date with what the current network head is.

renaynay: I hope this makes sense, and maybe @Wondertan can weigh in

header/interface.go
distractedm1nd: yeah after thinking, the simplicity is nice, agreed

header/sync/sync.go
```diff
@@ -201,43 +220,48 @@ func (s *Syncer) processIncoming(ctx context.Context, maybeHead *header.Extended
...
 
-	// 3. Filter out maybeHead if behind trusted
-	if maybeHead.Height <= trstHead.Height {
+	// 3. Attempt to verify maybeHead
+	return s.verifyMaybeHead(trstHead, maybeHead)
+}
+
+// verifyMaybeHead verifies the potential new head against the trusted head, and if
+// the potential new head is valid, sets it as the sync target.
+func (s *Syncer) verifyMaybeHead(trusted, maybe *header.ExtendedHeader) pubsub.ValidationResult {
```
Wondertan, line 236: Let's call it - `maybeNewHead` or just `newHead`

header/sync/sync.go
```diff
@@ -201,43 +220,48 @@ func (s *Syncer) processIncoming(ctx context.Context, maybeHead *header.Extended
...
 		return pubsub.ValidationReject
 	}
-
-	// 5. Save verified header to pending cache
+	// 3. Save verified header to pending cache
 	// NOTE: Pending cache can't be DOSed as we verify above each header against a trusted one.
-	s.pending.Add(maybeHead)
-	// and trigger sync to catch-up
+	s.pending.Add(maybe)
+	// 4. And trigger sync to catch up to new sync target
```
Wondertan, line 260: Hmm, this can cause an endless syncing loop. TrustedHead is triggered by wantSync, and now trustedHead triggers the wantSync throught `verifyMaybeHead`, so there is definitely a cycel.

Wondertan (commented):
There are some a few issues here that would be faster to fix in sync on the call

header/sync/sync.go
```diff
@@ -128,24 +139,32 @@ func (s *Syncer) trustedHead(ctx context.Context) (*header.ExtendedHeader, error
...
+	// automatic subjective initialization
 	objHead, err := s.exchange.Head(ctx)
 	if err != nil {
 		return nil, err
 	}
 
-	s.pending.Add(objHead)
-	return objHead, nil
+	// verify head returned from trustedPeer against subjective head
+	res := s.verifyMaybeHead(sbjHead, objHead)
```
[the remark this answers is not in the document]
renaynay: Only make this check with header !expired && !recent

header/sync/sync.go
```diff
@@ -128,24 +139,32 @@ func (s *Syncer) trustedHead(ctx context.Context) (*header.ExtendedHeader, error
...
 		return nil, err
 	}
 
-	// check if our subjective header is not expired and use it
-	if !sbj.IsExpired() {
-		return sbj, nil
+	// check if our subjective header is not expired and not too outdated (relative to the
+	// network's block time) and use it
+	if !sbjHead.IsExpired() && sbjHead.IsRecent(s.blockTime) {
+		return sbjHead, nil
```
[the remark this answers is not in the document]
renaynay: if IsRecent() {
 -- return sbj
}
if sbj !expired {
 verifyObjHead
} else {
  trustObjHead and return
  
 }

header/sync/sync.go
Wondertan: Because it also applies the newHead besides verification only

header/verify.go
distractedm1nd: Thanks for the explanation, it helped me understand a bit better now. Sounds like a tradeoff that I can't judge properly yet - for example, not sure how negative it is to be asking the peers for their head more often. Feel free to close this!

Wondertan:
Implementing the `Head` method for `Syncer` is the original goal of this PR. However, some internal logic has to be changed to allow this, and these logic changes are now extracted into https://github.com/celestiaorg/celestia-node/pull/990. This PR should rebase on top of it(after or before merge) and simply implement the `Head` method

renaynay:
Already implemented in main

PR #982, opened 2022-08-02
swamp/discovery: fix flakiness in restart node discovery test

### LINKED ISSUE #963: test: flakiness TestRestartNodeDiscovery

vgonkivs:
Resolves #963

Root cause: pubsub was restoring the connection between the nodes, so we didn't get the disconnect event as expected.

https://github.com/libp2p/go-libp2p-pubsub/blob/60cf38003244a277084c6f3eec0e584ab6cc07bd/pubsub.go#L710

### DIFF
--- a/node/tests/p2p_test.go
+++ b/node/tests/p2p_test.go
@@ -157,7 +157,7 @@ Steps:
 3. Create 2 full nodes with bridge node as bootstrapper peer and start them
 4. Check that nodes are connected to each other
 5. Create one more node with disabled discovery
-6. Stop one of full node in order to restart discovery
+6. Disconnect FNs from each other
 7. Check that the last FN is connected to one of the nodes
 *NOTE*: this test will take some time because it relies on several cycles of peer discovery
 */
@@ -165,7 +165,7 @@ func TestRestartNodeDiscovery(t *testing.T) {
 	sw := swamp.NewSwamp(t)
 	cfg := node.DefaultConfig(node.Bridge)
 	cfg.P2P.Bootstrapper = true
-	const defaultTimeInterval = time.Second * 10
+	const defaultTimeInterval = time.Second * 2
 	const fullNodes = 2
 	var defaultOptions = []node.Option{
 		node.WithPeersLimit(fullNodes),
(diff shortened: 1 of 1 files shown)


node/tests/p2p_test.go
```diff
@@ -188,42 +188,51 @@ func TestRestartNodeDiscovery(t *testing.T) {
...
 	// create one more node with disabled discovery
 	nodesConfig[1] = node.WithPeersLimit(0)
 	node := sw.NewFullNode(nodesConfig...)
+	connectSub, err := nodes[0].Host.EventBus().Subscribe(&event.EvtPeerConnectednessChanged{})
+	require.NoError(t, err)
+	defer connectSub.Close()
+
+	// disconnect and unlink peers in order to receive disconnected event
+	// NOTE: Consistence is very important here. We have to unlink peers and only after that call disconnect
+	// otherwise pubsub could re-establish the connection.
```
Wondertan, line 218:
```suggestion
	// NOTE: Order is very important here. We have to unlink peers, and only after that call disconnect,
	// otherwise pubsub could re-establish the connection.
```

node/tests/p2p_test.go
```diff
@@ -188,42 +188,51 @@ func TestRestartNodeDiscovery(t *testing.T) {
 		nodes[index] = sw.NewFullNode(nodesConfig...)
 	}
 
-	sub, err := nodes[0].Host.EventBus().Subscribe(&event.EvtPeerConnectednessChanged{})
+	identitySub, err := nodes[0].Host.EventBus().Subscribe(&event.EvtPeerIdentificationCompleted{})
```
Wondertan, line 191: Don't really get the reasoning behind moving to the Identification event. I know we do those header/p2p tests, but still, I am curious if this change really does anything

Wondertan (approved):

renaynay (approved):
utACK and +1 to hlib's comments

node/tests/p2p_test.go
vgonkivs: Good question. As per implementation `event.EvtPeerIdentificationCompleted` is emitting when identification of a remote peer is done and also its' done *after* `event.EvtPeerConnectednessChanged`. So, by changing EvtPeerConnectednessChanged to EvtPeerIdentificationCompleted I want:

1) to be 100% sure that connection between peers works smoothly;
2) to give more time for schedular to process waiting goroutines(I hope it helps to avoid adding time.Sleep in test);
3) Avoid using for loop because EvtPeerConnectednessChanged could be received more the once(with different peer.IDs).

PR #990, opened 2022-08-07
feat|refactor(header/sync): network head determination

Wondertan:
### Context
In https://github.com/celestiaorg/celestia-node/pull/978 we started adding the `Head` method to Syncer; however, it's not that straightforward, and the deeper we looked, the more issues we found in Syncer related to the new `Head` method, so it was decided to extract a separate preparation PR that ensures the new `Head` is safe to use by multiple routines and returns as recent header as possible.

The previous version of Syncer struggled with two issues:
* On Syncer's start, synchronization didn't start and waited for a gossiped header trigger sync and set a sync target (only when the subjective head was not expired)
  * This is why Node could wait up to block time second to start syncing
* There was no way to request the most recent objective header of the network
  * I.e. if the user wanted to request the latest possible state, it wasn't able to do that besides waiting for full sync to finish

### Changes
The new reimplementation fixes these two problems and improves code readability and docs. Mainly, it splits the existing `trustedHead` into two methods `subjectiveHead` and `networkHead`. The `networkHead` is supposed to be used by the future `Head`.
Where the latter now relies on the latest known header timestamp and block time to determine its recency. If the header is not recent, we request it from the trusted peer(s), assuming it's always synced.
> In fact two of our bootstrappers falling out of sync, creating syncing issues.

Besides that, three more issues under `fix` commits related to multithreading were fixed with supporting tests.

### Other
Tested sync manually(thought with some struggles due to broken bootstrappers).
_As always, review CBC and checkout locally to see the full picture._

### TODO
- [x] Double check that the node correctly determines recency, as their DA network might be lagging one block behind
- [ ] Do we need time drift?
- [ ] Telemetry

### DIFF
--- a/header/header.go
+++ b/header/header.go
@@ -87,7 +87,7 @@ func (eh *ExtendedHeader) IsBefore(h *ExtendedHeader) bool {
 	return eh.Height < h.Height
 }
 
-// Equals returns whether the hash of the given header matches.
+// Equals returns whether the hash and height of the given header match.
 func (eh *ExtendedHeader) Equals(header *ExtendedHeader) bool {
 	return eh.Height == header.Height && bytes.Equal(eh.Hash(), header.Hash())
 }
--- a/header/header.go
+++ b/header/header.go
@@ -82,6 +82,16 @@ func (eh *ExtendedHeader) LastHeader() bts.HexBytes {
 	return eh.RawHeader.LastBlockID.Hash
 }
 
+// IsBefore returns whether the given header is of a higher height.
+func (eh *ExtendedHeader) IsBefore(h *ExtendedHeader) bool {
+	return eh.Height < h.Height
+}
+
+// Equals checks equality of tho headers by comparing hashes
+func (eh *ExtendedHeader) Equals(header *ExtendedHeader) bool {
+	return eh.Height == header.Height && bytes.Equal(eh.Hash(), header.Hash())
+}
+
 // ValidateBasic performs *basic* validation to check for missed/incorrect fields.
 func (eh *ExtendedHeader) ValidateBasic() error {
 	err := eh.RawHeader.ValidateBasic()
--- a/header/header.go
+++ b/header/header.go
@@ -87,7 +87,7 @@ func (eh *ExtendedHeader) IsBefore(h *ExtendedHeader) bool {
 	return eh.Height < h.Height
 }
 
-// Equals checks equality of tho headers by comparing hashes
+// Equals returns whether the hash of the given header matches.
 func (eh *ExtendedHeader) Equals(header *ExtendedHeader) bool {
 	return eh.Height == header.Height && bytes.Equal(eh.Hash(), header.Hash())
 }
--- a/header/sync/sync.go
+++ b/header/sync/sync.go
@@ -16,19 +16,20 @@ var log = logging.Logger("header/sync")
 
 // Syncer implements efficient synchronization for headers.
 //
-// Subjective header - the latest known local header that is not expired(within trusting period)
-// Network header - the latest network header
+// Subjective header - the latest known header that is not expired (within trusting period)
+// Network header - the latest header received from the network
+
 //
 // There are two main processes running in Syncer:
 // 1. Main syncing loop(s.syncLoop)
-//    * Performs syncing from the known subjective header up to the network header
+//    * Performs syncing from the subjective header up to the network head
 //    * Syncs by requesting missing headers from Exchange or
 //    * By accessing cache of pending network headers received from PubSub
 // 2. Receives new headers from PubSub subnetwork (s.incomingNetHead)
 //    * Once received, tries to append it to the store
 //    * Or, if not adjacent to head of the store,
 //      * verifies against the latest known subjective header
-//    	* adds the header to pending cache(making it the latest known subjective header)
+//    	* adds the header to pending cache, thereby making it the latest known subjective header
 //      * and triggers syncing loop to catch up to that point.
 type Syncer struct {
 	sub      header.Subscriber
@@ -153,7 +154,7 @@ func (s *Syncer) syncLoop() {
 	}
 }
 
-// sync ensures we are synced from the local header Store's Head up to the new subjective head
+// sync ensures we are synced from the Store's head up to the new subjective head
 func (s *Syncer) sync(ctx context.Context) {
 	newHead := s.pending.Head()
 	if newHead == nil {
--- a/header/sync/sync.go
+++ b/header/sync/sync.go
@@ -7,8 +7,6 @@ import (
 	"time"
 
 	logging "github.com/ipfs/go-log/v2"
-
-	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	tmbytes "github.com/tendermint/tendermint/libs/bytes"
 
 	"github.com/celestiaorg/celestia-node/header"
@@ -18,12 +16,15 @@ var log = logging.Logger("header/sync")
 
 // Syncer implements efficient synchronization for headers.
 //
+// Subjective header - the latest known local header that is not expired(within trusting period)
+// Network header - the latest network header
+//
 // There are two main processes running in Syncer:
 // 1. Main syncing loop(s.syncLoop)
-//    * Performs syncing from the known subjective header up to the objective header
+//    * Performs syncing from the known subjective header up to the network header
 //    * Syncs by requesting missing headers from Exchange or
-//    * By accessing cache of pending headers received from PubSub
-// 2. Receives new headers from PubSub subnetwork (s.incomingHead)
+//    * By accessing cache of pending network headers received from PubSub
+// 2. Receives new headers from PubSub subnetwork (s.incomingNetHead)
 //    * Once received, tries to append it to the store
 //    * Or, if not adjacent to head of the store,
 //      * verifies against the latest known subjective header
@@ -43,10 +44,10 @@ type Syncer struct {
 	state   State
 	// signals to start syncing
 	triggerSync chan struct{}
-	// pending keeps ranges of valid headers received from the network awaiting to be appended to store
+	// pending keeps ranges of valid new network headers awaiting to be appended to store
 	pending ranges
-	// objLk ensures only one objective head is requested at any moment
-	objLk sync.RWMutex
+	// netReqLk ensures only one network head is requested at any moment
+	netReqLk sync.RWMutex
 
 	// controls lifecycle for syncLoop
 	ctx    context.Context
@@ -69,12 +70,12 @@ func (s *Syncer) Start(ctx context.Context) error {
 	s.ctx, s.cancel = context.WithCancel(context.Background())
 	// register validator for header subscriptions
 	// syncer does not subscribe itself and syncs headers together with validation
-	err := s.sub.AddValidator(s.incomingHead)
+	err := s.sub.AddValidator(s.incomingNetHead)
 	if err != nil {
 		return err
 	}
 	// get the latest head and set it as syncing target
-	_, err = s.objectiveHead(ctx)
+	_, err = s.networkHead(ctx)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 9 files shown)


Wondertan:
Tests fail because`Syncer.Start` now requires initialized `Store`. Will fix those later

header/sync/sync.go
```diff
@@ -119,138 +132,186 @@ func (s *Syncer) State() State {
...
+
 	head, err := s.store.Head(ctx)
 	if err != nil {
 		log.Errorw("getting head during sync", "err", err)
 		return
 	}
 
 	if head.Height == newHead.Height {
-		return
+		return // should never happen, but just in case
```
renaynay, line 314: If we are trying to catch "should never happen" case down here, maybe we can also check the case

`if head.Height >= newHead.Height`, log out a WARN and return?

renaynay (commented):
Initial thoughts here -- will do a deeper review after our call.

Want to discuss `objectiveHead` method.

Also we should consider splitting out process/validation-related functions into a separate file (maybe `head.go`) and just leaving the sync loop logic in `sync.go`

header/sync/sync.go
```diff
@@ -119,138 +132,186 @@ func (s *Syncer) State() State {
...
 
-// wantSync will trigger the syncing loop (non-blocking).
-func (s *Syncer) wantSync() {
-	select {
-	case s.triggerSync <- struct{}{}:
-	default:
+// objectiveHead returns the latest objective network header.
+// Known subjective header is considered objective if it is recent enough(now-timestamp<=blocktime).
+// Otherwise, objective header is requested from the network and set as a new subjective head.
+func (s *Syncer) objectiveHead(ctx context.Context) (*header.ExtendedHeader, error) {
```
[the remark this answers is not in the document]
Wondertan: Yes

header/sync/sync.go
```diff
@@ -20,53 +20,66 @@ var log = logging.Logger("header/sync")
 //
 // There are two main processes running in Syncer:
 // 1. Main syncing loop(s.syncLoop)
-//    * Performs syncing from the subjective(local chain view) header up to the latest known trusted header
+//    * Performs syncing from the known subjective header up to the objective header
```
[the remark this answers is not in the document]
Wondertan: Good idea. I will do that

header/sync/sync.go
```diff
@@ -119,138 +132,186 @@ func (s *Syncer) State() State {
...
 }
 
-// wantSync will trigger the syncing loop (non-blocking).
-func (s *Syncer) wantSync() {
-	select {
-	case s.triggerSync <- struct{}{}:
-	default:
+// objectiveHead returns the latest objective network header.
+// Known subjective header is considered objective if it is recent enough(now-timestamp<=blocktime).
+// Otherwise, objective header is requested from the network and set as a new subjective head.
```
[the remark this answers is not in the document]
Wondertan: Later on we would need to append s 'the trusted peer'

header/header.go
```diff
@@ -82,6 +82,16 @@ func (eh *ExtendedHeader) LastHeader() bts.HexBytes {
 	return eh.RawHeader.LastBlockID.Hash
 }
 
+// IsBefore returns whether the given header is of a higher height.
+func (eh *ExtendedHeader) IsBefore(h *ExtendedHeader) bool {
+	return eh.Height < h.Height
+}
+
+// Equals checks equality of tho headers by comparing hashes
```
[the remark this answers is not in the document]
Wondertan: Changed

header/sync/sync.go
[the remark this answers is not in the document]
Wondertan: Changed

header/sync/sync.go
```diff
@@ -119,138 +132,186 @@ func (s *Syncer) State() State {
...
-	// 4. Verify maybeHead against trusted
-	err = trstHead.VerifyNonAdjacent(maybeHead)
+// validate checks validity of given the header against the subjective head.
+func (s *Syncer) validate(ctx context.Context, new *header.ExtendedHeader) pubsub.ValidationResult {
+	sbjHead, err := s.subjectiveHead(ctx)
+	if err != nil {
+		log.Errorw("getting subjective head", "err", err)
+		return pubsub.ValidationIgnore // local error, so ignore
+	}
+	// if header is from the past, we don't know, so ignore
```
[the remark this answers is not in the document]
Wondertan: Changed

header/sync/sync.go
Wondertan: Done

header/header.go
```diff
@@ -82,6 +82,16 @@ func (eh *ExtendedHeader) LastHeader() bts.HexBytes {
 	return eh.RawHeader.LastBlockID.Hash
 }
 
+// IsBefore returns whether the given header is of a higher height.
+func (eh *ExtendedHeader) IsBefore(h *ExtendedHeader) bool {
```
renaynay, line 86: Shouldn't this check include `==` as well? Or at least, on line 258, do 

`if !sbjHead.IsBefore && !sbjHead.Equals=`

header/sync/sync.go
```diff
@@ -119,76 +132,94 @@ func (s *Syncer) State() State {
 	return state
 }
 
-// trustedHead returns the latest known trusted header that is within the trusting period.
-func (s *Syncer) trustedHead(ctx context.Context) (*header.ExtendedHeader, error) {
-	// check pending for trusted header and return it if applicable
-	// NOTE: Pending cannot be expired, guaranteed
+// subjectiveHead returns the latest known subjective header that is not expired(within trusting period).
+// Lazily performs 'automatic subjective initialization' in case the header is expired.
```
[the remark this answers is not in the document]
Wondertan: Mmm, there is a reason I put 'automatic subjective initialization', and yes, the method doc assumes the reader is familiar with this assumption. I will keep the existing doc, but I will remove the familiarity assumption by elaborating on what the term means

header/verify.go
```diff
@@ -26,6 +26,11 @@ func (eh *ExtendedHeader) IsExpired() bool {
 	return !expirationTime.After(time.Now())
 }
 
+// IsRecent checks if header is recent against the given blockTime.
+func (eh *ExtendedHeader) IsRecent(blockTime time.Duration) bool {
+	return time.Since(eh.Time) <= blockTime // TODO @renaynay: should we allow for a 5-10 block drift here?
```
renaynay, line 31: Let's just put this in an issue

renaynay: For now, let's not allow any drift unless it becomes an issue.

header/sync/sync.go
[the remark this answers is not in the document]
Wondertan: A bit outdated IMO + not specifically about Tendermint, but still, [the good read](https://blog.ethereum.org/2014/11/25/proof-stake-learned-love-weak-subjectivity/) to follow.

header/sync/sync.go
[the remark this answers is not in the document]
Wondertan: I've changed "objective" to "network".

header/sync/sync.go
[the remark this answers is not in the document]
Wondertan: Comment updated, try giving it another read

header/verify.go
Wondertan: Drift, yes, but we still need to figure out if we lag on block or not

header/header.go
Wondertan: Nope

Wondertan:
@renaynay, @distractedm1nd, pls give another read to the docs. I think it's a different level now.

Wondertan:
We should not merge this one until we downgrade our network to 0.34. Testing recency on Mamaki is impossible due to bigger block times than they should be. Also, this change does not bring ant value without stable block times
image

header/sync/sync.go
```diff
@@ -119,35 +133,6 @@ func (s *Syncer) State() State {
...
-	}
-
-	// otherwise, request head from a trustedPeer or, in other words, do automatic subjective initialization
-	objHead, err := s.exchange.Head(ctx)
-	if err != nil {
-		return nil, err
-	}
-
-	s.pending.Add(objHead)
-	return objHead, nil
```
distractedm1nd, line 148: If we change from objective header to network header, we should probably also change the local var naming

distractedm1nd (approved):
docs are great! found one last thing, but its a nit :)

header/sync/sync.go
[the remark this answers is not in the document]
distractedm1nd: looks great now!

header/sync/sync.go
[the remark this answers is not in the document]
distractedm1nd: I think I like network head better than objective head as well, great idea :)

header/sync/sync.go
Wondertan: You pointed to removed code. In new code everything is renamed

distractedm1nd: Oh, ok sorry! Was shown the new file, so assumed the other files were also synced.. weird

renaynay (commented):
testing this against arabica and everything looks fine so far -- the problem is not enough blocks have been produced to be able to observe any larger range syncs. 

I think we should each (@Wondertan and I) review this PR again with fresh eyes and then we can likely merge.

header/header.go
renaynay: Ignore the above ^ 

the diff has changed since

Wondertan:
> testing this against arabica and everything looks fine so far -- the problem is not enough blocks have been produced to be able to observe any larger range syncs.

@renaynay, How do you test? Do you compare that the node's instant head is equal to the core's one? Do you request core's one directly from the validator?

header/sync/sync.go
```diff
@@ -157,100 +142,37 @@ func (s *Syncer) wantSync() {
...
 
-	if head.Height == newHead.Height {
-		return
+	if head.Height >= newHead.Height {
+		log.Warnw("sync attempt to an already synced header",
+			"synced_height", head.Height,
+			"attempted_height", newHead.Height,
+		)
+		log.Warn("PLEASE REPORT THIS AS A BUG")
+		return // should never happen, but just in case
```
liamsi, line 175: If this should really never happen, shouldn't this be logged out as an error instead?

header/sync/sync_head.go
```diff
@@ -0,0 +1,167 @@
...
+	"errors"
+
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
+
+	"github.com/celestiaorg/celestia-node/header"
+)
+
+// subjectiveHead returns the latest known local header that is not expired(within trusting period).
+// If the header is expired, it is retrieved from a trusted peer without validation;
+// in other words, an automatic subjective initialization is performed.
```
liamsi, line 14: I think this must be clearly communicated in the usage instruction that setting a "trusted peer", translates into trusting any header that is received from that peer.

liamsi (commented):
The PR looks generally good. I think it is critical that we document the assumptions around subjective head, network head, trusted peers clearly and make any implicit assumptions very explicit to the user.

header/sync/sync.go
```diff
@@ -18,55 +16,71 @@ var log = logging.Logger("header/sync")
 
 // Syncer implements efficient synchronization for headers.
 //
+// Subjective header - the latest known local header that is not expired(within trusting period)
```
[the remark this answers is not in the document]
Wondertan: It is defined in Syncer's godoc. We can copy it wherever we want, like in node docs.

header/sync/sync.go
Wondertan: Don't think so. This case does not break syncer

header/sync/sync_head.go
Wondertan: > translates into trusting _any_ header that is received from that peer.
@liamsi, a small correction here. Not _any_ but only `network head` during automatic subjective init. Other than that, `network head` and past headers are verified as per logic in the PR. 

Besides the code itself, this should be clearly documented on node docs and in the config. This is though orthogonal to the PR.

renaynay (dismissed):
Tested on all node types -- works very well. Thank you!

header/sync/sync.go
```diff
@@ -18,55 +16,71 @@ var log = logging.Logger("header/sync")
...
+// Network header - the latest network header
+//
 // There are two main processes running in Syncer:
 // 1. Main syncing loop(s.syncLoop)
-//    * Performs syncing from the subjective(local chain view) header up to the latest known trusted header
+//    * Performs syncing from the known subjective header up to the network header
```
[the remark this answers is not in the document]
Wondertan: Why not 'known'?

renaynay (approved):
morge

Bidon15 (approved):

PR #996, opened 2022-08-08
feat (service/state): If account doesn't exist at given height, fail gracefully

renaynay:
This PR implements a check to ensure that the `value` returned by the ABCI balance request is not empty. If the balance is empty, it will return 0 and log out an err.

### DIFF
--- a/service/state/core_access.go
+++ b/service/state/core_access.go
@@ -9,6 +9,7 @@ import (
 	sdktx "github.com/cosmos/cosmos-sdk/types/tx"
 	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
 	proofutils "github.com/cosmos/ibc-go/v4/modules/core/23-commitment/types"
+	logging "github.com/ipfs/go-log/v2"
 	rpcclient "github.com/tendermint/tendermint/rpc/client"
 	"github.com/tendermint/tendermint/rpc/client/http"
 	"google.golang.org/grpc"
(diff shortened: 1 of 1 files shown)


distractedm1nd (approved):
ship it

distractedm1nd:
actually..... wait. Why not return 0? And then keep the err message as a log

renaynay:
Good point @distractedm1nd

service/state/core_access.go
```diff
@@ -157,6 +160,15 @@ func (ca *CoreAccessor) BalanceForAddress(ctx context.Context, addr Address) (*B
 	}
 	// unmarshal balance information
 	value := result.Response.Value
+	// if the value returned is empty, the account balance does not yet exist
+	if len(value) == 0 {
+		log.Errorf("balance for account %s does not exist at block height %d, "+
+			"wait for node to sync", addr.String(), head.Height)
```
Wondertan, line 166: Should we not keep this suggestion? It is not going to be actual once we merge Syncer

Wondertan (commented):
LGTM besides one q

service/state/core_access.go
renaynay: Yes, how likely is it going to be that a user is requesting balance of acct that doesn't exist in trie?

Wondertan: I mean specifically about the `wait for node to sync"`

renaynay: The assumption that I'm referring to here is that a user would *likely* not request the balance of an account that does not yet exist in acct trie, meaning the node should sync up to network head where some acct data exists, does that make sense?

Wondertan (approved):

PR #1001, opened 2022-08-10
tracing: Adding traces to `get_shares.go`

distractedm1nd:
This PR adds traces to `get_shares.go`, similar to the namespaced shares tracing

### DIFF
--- a/ipld/get_shares.go
+++ b/ipld/get_shares.go
@@ -60,7 +60,7 @@ func GetShares(ctx context.Context, bGetter blockservice.BlockGetter, root cid.C
 
 	// this buffer ensures writes to 'jobs' are never blocking (bin-tree-feat)
 	jobs := make(chan *job, (shares+1)/2) // +1 for the case where 'shares' is 1
-	jobs <- &job{id: root}
+	jobs <- &job{id: root, ctx: ctx}
 	// total is an amount of routines spawned and total amount of nodes we process (bin-tree-feat)
 	// so we can specify exact amount of loops we do, and wait for this amount
 	// of routines to finish processing
@@ -74,7 +74,7 @@ func GetShares(ctx context.Context, bGetter blockservice.BlockGetter, root cid.C
 			// work over each job concurrently, s.t. shares do not block
 			// processing of each other
 			pool.Submit(func() {
-				ctx, span := tracer.Start(ctx, "process-job")
+				ctx, span := tracer.Start(j.ctx, "process-job")
 				defer wg.Done()
 				defer span.End()
 
@@ -113,6 +113,9 @@ func GetShares(ctx context.Context, bGetter blockservice.BlockGetter, root cid.C
 						// calc position for children nodes (bin-tree-feat),
 						// s.t. 'if' above knows where to put a share
 						pos: j.pos*2 + i,
+						// we pass the context to job so that spans are tracked in a tree
+						// structure
+						ctx: ctx,
 					}:
 						span.AddEvent("added-job", trace.WithAttributes(
 							attribute.String("cid", lnk.Cid.String()),
--- a/ipld/get_shares.go
+++ b/ipld/get_shares.go
@@ -1,16 +1,15 @@
 package ipld
 
 import (
-	"sync"
-
-	"go.opentelemetry.io/otel/attribute"
-	"go.opentelemetry.io/otel/trace"
-
 	"context"
+	"go.opentelemetry.io/otel/codes"
+	"sync"
 
 	"github.com/gammazero/workerpool"
 	"github.com/ipfs/go-blockservice"
 	"github.com/ipfs/go-cid"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
 
 	"github.com/celestiaorg/celestia-node/ipld/plugin"
 )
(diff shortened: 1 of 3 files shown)


ipld/get_shares.go
```diff
@@ -68,7 +75,10 @@ func GetShares(ctx context.Context, bGetter blockservice.BlockGetter, root cid.C
 			// work over each job concurrently, s.t. shares do not block
 			// processing of each other
 			pool.Submit(func() {
+				ctx, span := tracer.Start(ctx, "process-job")
+				defer span.End()
 				defer wg.Done()
```
renaynay, line 80: LIFO

renaynay (commented):
Any way we can add an attribute to track time?

ipld/get_shares.go
```diff
@@ -1,9 +1,13 @@
...
 
 import (
-	"context"
 	"sync"
 
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+
+	"context"
+
```
Wondertan, line 10: Pls sort imports

ipld/get_shares.go
```diff
@@ -82,11 +92,13 @@ func GetShares(ctx context.Context, bGetter blockservice.BlockGetter, root cid.C
 					// leaf has its own additional leaf(hack) so get it
 					nd, err := plugin.GetNode(ctx, bGetter, lnks[0].Cid)
 					if err != nil {
-						// again, we don't care
+						// again, we don't really care much, just fetch as much as possible
+						span.RecordError(err)
```
Wondertan, line 96: Let's set status on error and on success, as per https://opentelemetry.io/docs/reference/specification/trace/api/#set-status

An example: https://uptrace.dev/opentelemetry/go-tracing.html#setting-status-code

ipld/get_shares.go
distractedm1nd: done

ipld/get_shares.go
distractedm1nd: done

ipld/get_shares.go
distractedm1nd: Alright, tried this, is it what you had in mind?

Wondertan (commented):
Just realize that we should do nesting of spans. We create a new span for each job here, but those spans are not interconnected in the form of a tree like they are in practice. Debuggability characteristics via the current flat structure is questionable, while tree structure will be much more helpful. 

To do this, we need to extend the job struct to take a `ctx` field, based on which child spans will be created.

Also, we don't need to create an event for each added job as each is a span(loc 115), so you can just add attributes(cid and pos) to the span directly by using `SetAttributes(kv ...attribute.KeyValue)` on the span.

ipld/get_shares.go
```diff
@@ -76,13 +75,15 @@ func GetShares(ctx context.Context, bGetter blockservice.BlockGetter, root cid.C
 			// processing of each other
 			pool.Submit(func() {
 				ctx, span := tracer.Start(ctx, "process-job")
-				defer span.End()
 				defer wg.Done()
+				defer span.End()
```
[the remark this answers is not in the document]
distractedm1nd: @renaynay investigated this and found that multiple defers are LIFO.

ipld/get_shares.go
```diff
@@ -99,7 +113,14 @@ func GetShares(ctx context.Context, bGetter blockservice.BlockGetter, root cid.C
...
 						// s.t. 'if' above knows where to put a share
 						pos: j.pos*2 + i,
+						// we pass the context to job so that spans are tracked in a tree
+						// structure
+						ctx: ctx,
 					}:
+						span.AddEvent("added-job", trace.WithAttributes(
+							attribute.String("cid", lnk.Cid.String()),
+							attribute.Int("pos", j.pos*2+i),
+						))
```
Wondertan, line 123: Pls, add those attributes to the `process-job` span instead of event, as mentioned in the last review

Wondertan: Similarly for the namespace pls

Wondertan (commented):
Thanks for covering `GetShares` as well 🙏🏻

ipld/get_shares.go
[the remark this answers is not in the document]
Wondertan: yes and why this order is needed here, don't really get it?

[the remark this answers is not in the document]
Wondertan: Then you won't need to record error with attributes on loc 170 as well

ipld/get_shares.go
Wondertan: Then you won't need to record error with attributes on loc 170 as well

Wondertan: because the error is attached to the span which would have those attributes attached there already

distractedm1nd: Ohh now I get it. Makes sense, also removing the added-job event in total because the new correct tree structure makes it obselete

ipld/get_shares.go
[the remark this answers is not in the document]
distractedm1nd: Should the span always be the last thing to be closed? Felt/sounded right at first to end the span, then "close" the job, but I can see how that probably doesn't make sense, since the span should encompass ("span") over absolutely everything

ipld/get_shares.go
Wondertan: Yes

ipld/get_shares.go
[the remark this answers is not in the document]
Wondertan: Yes

Wondertan (commented):
Did we move span.End() back? @distractedm1nd

distractedm1nd:
yes

distractedm1nd:
oh wait didn't push

Wondertan (approved):

vgonkivs (approved):

PR #1004, opened 2022-08-11
cmd: Drop Env in favour of vanilla context.WithValue wrappers

### LINKED ISSUE #965: cmd: Drop `Env` in favour of vanilla context.WithValue wrappers
Env, in fact, only keeps some additional and often optional fields taken from flags which we can keep in the context using `WithValue`

distractedm1nd:
Closes #965 .

### DIFF
--- a/cmd/celestia/bridge.go
+++ b/cmd/celestia/bridge.go
@@ -40,7 +40,6 @@ var bridgeCmd = &cobra.Command{
 			ctx = cmd.Context()
 			err error
 		)
-		defer cmd.SetContext(ctx)
 
 		ctx = cmdnode.WithNodeType(ctx, node.Bridge)
 
@@ -71,6 +70,7 @@ var bridgeCmd = &cobra.Command{
 
 		ctx = cmdnode.ParseKeyFlags(ctx, cmd)
 
+		cmd.SetContext(ctx)
 		return nil
 	},
 }
--- a/cmd/celestia/bridge.go
+++ b/cmd/celestia/bridge.go
@@ -42,7 +42,7 @@ var bridgeCmd = &cobra.Command{
 		)
 		defer cmd.SetContext(ctx)
 
-		ctx = cmdnode.SetNodeType(ctx, node.Bridge)
+		ctx = cmdnode.WithNodeType(ctx, node.Bridge)
 
 		ctx, err = cmdnode.ParseNodeFlags(ctx, cmd)
 		if err != nil {
--- a/cmd/celestia/bridge.go
+++ b/cmd/celestia/bridge.go
@@ -54,11 +54,6 @@ var bridgeCmd = &cobra.Command{
 			return err
 		}
 
-		ctx, err = cmdnode.ParseHeadersFlags(ctx, cmd)
-		if err != nil {
-			return err
-		}
-
 		ctx, err = cmdnode.ParseMiscFlags(ctx, cmd)
 		if err != nil {
 			return err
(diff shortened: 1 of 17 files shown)


cmd/env.go
```diff
@@ -2,60 +2,40 @@ package cmd
...
+func AddOptions(ctx context.Context, opts ...node.Option) context.Context {
+	options := Options(ctx)
+	return context.WithValue(ctx, optionsKey{}, append(options, opts...))
 }
 
-// envCtxKey is a key used to identify Env on a ctx.Context.
-type envCtxKey struct{}
+type optionsKey struct{}
+type storePathKey struct{}
+type nodeTypeKey struct{}
```
renaynay, line 41:
```suggestion
type (
	optionsKey   struct{}
	storePathKey struct{}
	nodeTypeKey  struct{}
)
```

renaynay (commented):
Minor comments and also pls format imports

cmd/env.go
```diff
@@ -1,61 +1,48 @@
...
-	env.NodeType = tp
+// SetStorePath sets Store Path to the Env.
+func SetStorePath(ctx context.Context, storePath string) context.Context {
+	return context.WithValue(ctx, storePathKey{}, storePath)
 }
 
 // Options returns Node Options parsed from Environment(Flags, ENV vars, etc)
-func (env *Env) Options() []node.Option {
-	return env.opts
+func Options(ctx context.Context) []node.Option {
```
Wondertan, line 30: NodeOptions?

Wondertan, line 30: NodeOptions?

cmd/env.go
```diff
@@ -1,61 +1,48 @@
...
-//  Env, in fact, only keeps some additional fields which should be
-//  kept in the context directly using WithValue (#965)
-type Env struct {
-	NodeType  node.Type
-	StorePath string
-
-	opts []node.Option
+// NodeType reads the node.Type from the context.
+func NodeType(ctx context.Context) node.Type {
+	return ctx.Value(nodeTypeKey{}).(node.Type)
```
Wondertan, line 11: This could panic if the values are not given, which is fine(?). Yeah, it's fine because if the programmer forgot to pass the Node type, that's his fault, and panic will show during testing. Although some would consider erroring a better thing

Wondertan (commented):
❤️ . One quick suggestion, let's follow  ctx naming pattern by naming setters prefixed with `With` for the constituency with std. Other libs around also do that.

Also, one thing to consider: should we allow overriding values, or should we check if values are already present on the ctx.

cmd/env.go
distractedm1nd: Renamed to NodeOptions, but one question: Should AddNodeOptions also be refactored to WithNodeOptions? I don't want it to be unclear that the previous ones are not being overwritten, but just appended

cmd/celestia/bridge.go
```diff
@@ -36,40 +36,37 @@ var bridgeCmd = &cobra.Command{
...
-		err = cmdnode.ParseRPCFlags(cmd, env)
+		ctx, err = cmdnode.ParseRPCFlags(ctx, cmd)
 		if err != nil {
 			return err
 		}
 
-		cmdnode.ParseKeyFlags(cmd, env)
+		ctx = cmdnode.ParseKeyFlags(ctx, cmd)
 
+		cmd.SetContext(ctx)
```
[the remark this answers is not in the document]
distractedm1nd: resolved in sync, this breaks because ctx isnt a pointer and we keep assigning to it after calling defer

cmd/env.go
distractedm1nd: Agreed, resolving

renaynay:
> ❤️ . One quick suggestion, let's follow ctx naming pattern by naming setters prefixed with `With` for the constituency with std. Other libs around also do that.
> 
> Also, one thing to consider: should we allow overriding values, or should we check if values are already present on the ctx.

I think allowing overriding for now is fine.

Wondertan (approved):
In the end, this looks even better than I saw in my head. Great

renaynay (approved):

PR #1013, opened 2022-08-17
feat: adding pre-commit hook and makefile installation script

distractedm1nd:
Adds a pre-commit git hook that runs the linter and `go mod tidy`

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -8,7 +8,7 @@ if ! command -v golangci-lint &> /dev/null ; then
     exit 1
 fi
 
-golangci-lint run
+make lint 
 
 if go mod tidy -v 2>&1 | grep -q 'updates to go.mod needed'; then
     exit 1
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,0 +1,23 @@
+#!/bin/sh
+
+set -eu -o pipefail
+
+if ! command -v golangci-lint &> /dev/null ; then
+    echo "golangci-lint not installed or available in the PATH" >&2
+    echo "please check https://github.com/golangci/golangci-lint" >&2
+    exit 1
+fi
+
+golangci-lint run
+
+if go mod tidy -v 2>&1 | grep -q 'updates to go.mod needed'; then
+    exit 1
+fi
+
+git diff --exit-code go.* &> /dev/null
+
+if [ $? -eq 1 ]; then
+    echo "go.mod or go.sum differs, please re-add it to your commit"
+
+    exit 1
+fi
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -2,13 +2,28 @@
 
 set -eu -o pipefail
 
-if ! command -v golangci-lint &> /dev/null ; then
-    echo "golangci-lint not installed or available in the PATH" >&2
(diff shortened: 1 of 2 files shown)


adlerjohn (commented):
Pre-commit hooks should generally be avoided IMO.

1. They're not portable.
2. They can lead to issues such as https://github.com/trufflesuite/ganache/issues/562
3. They're unexpected, as they mutate code opaquely.

It's better to _verify_ rather than _execute_, i.e. run `go mod tidy` in CI and fail if there's a diff.

distractedm1nd:
@adlerjohn Right, but these are not forced, each contributor has to opt-in to it - prevents `chore: lint` commits and yucky `go mod tidy` commits

adlerjohn:
> each contributor has to opt-in

Do you mean opt-out? The documentation I'm reading says you can opt out with `--no-verify`.

distractedm1nd:
No, I mean opt-in. You only have to install the hooks if you want. They are not installed automatically when downloading the repo (as with husky), thus adding the command in the makefile. Once you have installed them, you can ignore them for single commits with `--no-verify` if you want (for example, checkpointing during refactorings)

distractedm1nd:
@adlerjohn https://github.com/celestiaorg/celestia-app/tree/main/contrib/githooks
Celestia app also has them!

adlerjohn:
Ah got it, thanks.

Makefile
```diff
@@ -11,6 +11,12 @@ help: Makefile
 	@sed -n 's/^##//p' $< | column -t -s ':' |  sed -e 's/^/ /'
 .PHONY: help
 
+## init: Install git-hooks from .githooks directory.
+init-hooks:
```
Wondertan, line 15:
```suggestion
## install: Install git-hooks from .githooks directory.
install-hooks:
```

.githooks/pre-commit
```diff
@@ -0,0 +1,23 @@
...
+
+set -eu -o pipefail
+
+if ! command -v golangci-lint &> /dev/null ; then
+    echo "golangci-lint not installed or available in the PATH" >&2
+    echo "please check https://github.com/golangci/golangci-lint" >&2
+    exit 1
+fi
+
+golangci-lint run
```
Wondertan, line 11:
```suggestion
make lint 
```

Wondertan (commented):
Can we also add `make test` in the precommit hook?

distractedm1nd:
@Wondertan Good question, renaynay and I talked about this and decided to leave testing out of the commit hook, because checkpointing/committing shouldn't necessarily mean all tests pass (and, waiting a long time for tests to pass before the commit goes through gets annoying very quickly)

Wondertan (approved):
Makes sense

vgonkivs:
I'd suggest:
1) add condition that skips pre-commit hook if go files were not modified(pr with ADRs only)
2) run markdown check in case of 1)

vgonkivs (changes requested):
I'd suggest:

add condition that skips pre-commit hook if go files were not modified(pr with ADRs only)
run markdown check in case of 1)

Wondertan (approved):

adlerjohn:
Relevant: https://github.com/celestiaorg/celestia-app/pull/672

vgonkivs (approved):

PR #1030, opened 2022-08-22
refactor: migration of cmd flags/opts to a new `node/header` subpackage

distractedm1nd:
Inspired and forked from #1017 , just a draft PR to check if I'm doing this right first :)

Moves header flags and opts from cmd to `node/header`. Still looking for other things here that I should move into this subpackage.

Will have to rebase on 1017 when updated.

Related to #958


distractedm1nd:
Closing in favor of https://github.com/renaynay/celestia-node/pull/90

PR #1033, opened 2022-08-23
fraud: add proof mock

vgonkivs:


vgonkivs:
```
=== RUN   TestBlockFetcherHeaderValues
==================
WARNING: DATA RACE
Write at 0x0000020a5cf0 by goroutine 123:
  github.com/tendermint/tendermint/rpc/core.SetEnvironment()
      /home/runner/go/pkg/mod/github.com/celestiaorg/celestia-core@v1.3.3-tm-v0.34.20/rpc/core/env.go:44 +0xa18
  github.com/tendermint/tendermint/node.(*Node).ConfigureRPC()
      /home/runner/go/pkg/mod/github.com/celestiaorg/celestia-core@v1.3.3-tm-v0.34.20/node/node.go:1063 +0x289
  github.com/tendermint/tendermint/node.(*Node).startRPC()
      /home/runner/go/pkg/mod/github.com/celestiaorg/celestia-core@v1.3.3-tm-v0.34.20/node/node.go:1091 +0x4a
  github.com/tendermint/tendermint/node.(*Node).OnStart()
      /home/runner/go/pkg/mod/github.com/celestiaorg/celestia-core@v1.3.3-tm-v0.34.20/node/node.go:949 +0x3c8
  github.com/tendermint/tendermint/libs/service.(*BaseService).Start()
      /home/runner/go/pkg/mod/github.com/celestiaorg/celestia-core@v1.3.3-tm-v0.34.20/libs/service/service.go:144 +0x582
  github.com/tendermint/tendermint/rpc/test.StartTendermint()
      /home/runner/go/pkg/mod/github.com/celestiaorg/celestia-core@v1.3.3-tm-v0.34.20/rpc/test/helpers.go:127 +0x149
  github.com/celestiaorg/celestia-node/core.StartTestNode()
      /home/runner/work/celestia-node/celestia-node/core/testing.go:28 +0xa4
  github.com/celestiaorg/celestia-node/core.StartTestKVApp()
      /home/runner/work/celestia-node/celestia-node/core/testing.go:41 +0x107
  github.com/celestiaorg/celestia-node/core.StartTestClient()
      /home/runner/work/celestia-node/celestia-node/core/testing.go:55 +0x67
  github.com/celestiaorg/celestia-node/core.TestBlockFetcherHeaderValues()
      /home/runner/work/celestia-node/celestia-node/core/fetcher_test.go:47 +0xa4
  testing.tRunner()
      /opt/hostedtoolcache/go/1.18.5/x64/src/testing/testing.go:1439 +0x213
  testing.(*T).Run.func1()
      /opt/hostedtoolcache/go/1.18.5/x64/src/testing/testing.go:1486 +0x47

Previous read at 0x0000020a5cf0 by goroutine 51:
  github.com/tendermint/tendermint/rpc/core.Subscribe.func1()
      /home/runner/go/pkg/mod/github.com/celestiaorg/celestia-core@v1.3.3-tm-v0.34.20/rpc/core/events.go:95 +0x1010
```

Wondertan:
Let's rebase it on main

fraud/registry.go
```diff
@@ -1,17 +1,45 @@
...
 	unmarshalersLk      = sync.RWMutex{}
 	defaultUnmarshalers = map[ProofType]ProofUnmarshaler{}
+
+	proofsLk            = sync.RWMutex{}
+	supportedProofTypes = map[ProofType]string{}
 )
 
-// Register sets unmarshaler in map by provided ProofType.
-func Register(proofType ProofType, unmarshaler ProofUnmarshaler) {
+// Register sets supported proofs with it string representation and unmarshalers in maps by provided ProofType.
```
distractedm1nd, line 17:
```suggestion
// Register adds support for the provided ProofType, adding a string representation and unmarshaller
```

distractedm1nd (approved):
Nice!

fraud/registry.go
```diff
@@ -1,17 +1,45 @@
...
+	defaultUnmarshalers[p.Type()] = func(data []byte) (Proof, error) {
+		err := p.UnmarshalBinary(data)
+		return p, err
+	}
+	unmarshalersLk.Unlock()
+	return nil
+}
+
+// GetTopic returns string representation of topic by provided ProofType.
+func GetTopic(proofType ProofType) (string, error) {
```
Wondertan, line 37: Let's make it private. Unless there is a outside usecase for it, which I don't see.

fraud/registry.go
```diff
@@ -1,17 +1,45 @@
...
+		return errors.New("fraud: proofType is registered")
+	}
+	supportedProofTypes[p.Type()] = p.Name()
+	proofsLk.Unlock()
+
 	unmarshalersLk.Lock()
-	defer unmarshalersLk.Unlock()
-	defaultUnmarshalers[proofType] = unmarshaler
+	defaultUnmarshalers[p.Type()] = func(data []byte) (Proof, error) {
+		err := p.UnmarshalBinary(data)
```
Wondertan, line 29: Hmm, there is an issue here. P is not copied and the same value might be reused causing races. Can we make a copy of an empty proof somehow before unmarshalling it?

Wondertan (commented):
Quick review

fraud/bad_encoding.go
```diff
@@ -17,8 +17,15 @@ import (
...
+	badEncodingTopic = "badencoding"
+)
+
 func init() {
-	Register(BadEncoding, UnmarshalBEFP)
+	err := Register(BadEncoding, badEncodingTopic, UnmarshalBEFP)
+	if err != nil {
+		panic(err)
+	}
 }
```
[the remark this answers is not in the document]
vgonkivs: why not

fraud/registry.go
vgonkivs: ok

fraud/registry.go
vgonkivs: created a copy of proof

fraud/registry.go
```diff
@@ -9,9 +10,16 @@ var (
...
 )
 
-// Register sets unmarshaler in map by provided ProofType.
-func Register(proofType ProofType, unmarshaler ProofUnmarshaler) {
+// Register adds a string representation and unmarshaller for the provided ProofType.
+func Register(p Proof) {
 	unmarshalersLk.Lock()
 	defer unmarshalersLk.Unlock()
-	defaultUnmarshalers[proofType] = unmarshaler
+	defaultUnmarshalers[p.Type()] = func(data []byte) (Proof, error) {
```
Wondertan, line 17: We should still panic if the Unmarshaler is already registered. We must not allow overriding

fraud/registry.go
```diff
@@ -9,9 +11,19 @@ var (
...
 	defer unmarshalersLk.Unlock()
-	defaultUnmarshalers[proofType] = unmarshaler
+	if _, ok := defaultUnmarshalers[p.Type()]; ok {
+		panic(fmt.Sprintf("fraud: unmarshaler for %s proof is registered", p.Type()))
+	}
+	defaultUnmarshalers[p.Type()] = func(data []byte) (Proof, error) {
+		// the underlying type of `p` is a pointer to a struct and assigning `p` to a new variable is not the
+		// case, because it could lead to data races.
+		// So, there is no easier way to create a hard copy of Proof other than using a reflection.
+		proof := reflect.New(reflect.ValueOf(p).Elem().Type()).Interface().(Proof)
```
renaynay, line 25: doing proof := *p would lead to a data race?

fraud/testing.go
```diff
@@ -66,3 +70,54 @@ func generateByzantineError(
...
+	return newMockProof(true)
+}
+
+func newInvalidProof() *mockProof {
+	return newMockProof(false)
+}
+
+func newMockProof(valid bool) *mockProof {
+	p := &mockProof{valid}
+	if _, ok := defaultUnmarshalers[p.Type()]; !ok {
```
renaynay, line 92: test should probably also use the lock to prevent races/flakeyness, no?

renaynay (commented):
Can you add a description in the PR **and to the adr** adding rationale why we use string instead of int now to represent type of proof and why we re-added the unmarshaller?

Same for

fraud/testing.go
vgonkivs: we run test synchronously.

fraud/registry.go
vgonkivs: you are not able to do  `proof := *p`, bc `p` is an interface. So, the only way is `proof := p`, but in this proof address will be copied, so getting unmarshaller from different threads will cause a data race, because you will be accessing the same address inside func.

Wondertan: We are talking about different races here btw. 
@renaynay to answer yours, no, because copying is reading and reading only is race free

renaynay: I see Proof is an interface yes. Okay.

Wondertan: There is a tradeoff between code readability and API simplicity. We can either pass one param as the Proof itself to Register it, or we pass two params(type and Unmarshaller) to register it. 

We consider readability being solvable if properly documented and simplifies the life of users(like you) who don't want to care or understand why the registration for unmarshalling is needed. Here, we just registered a proof and everything else is done for you

renaynay (approved):
👍🏻

Issue #11, opened 2021-06-06
DHT usage and future work


Wondertan:
## Background
The Data Availability model we use requires data discovery. We rely on [IPFS's Kademlia DHT](https://github.com/libp2p/go-libp2p-kad-dht), which basically allows any network participant to find a host for a certain piece of data by its hash. 

## Usage Description
To describe the way we use it, let's introduce a simple pseudo-code interface for it:
```
interface DHT {
	// Find the nearest peer to the hash and ask him to keep a record of us hosting the data.
	// By default, records are stored for 24h.
	Provide(hash) 
	// Find peers hosting the data by its hash. 
	FindProviders(hash) []peer 
	// Periodically execute Provide for a given hash to keep record around.
	Reprovide(hash) 
}
```

When a block producer creates a block, it saves it and calls `Provide` for every Data Availability root of the block, making it discoverable and afterward available. After, any other node that wants to get the block's data or validate its availability can call `FindProviders`, detect the block producer, and finally access the block data through Bitswap. The block producer and block requester also call `Reprovide`. Overall, with the described flow, we aim for maximum confidence that data of any particular block is always discoverable from peers storing it.

## What's Left
The current state of the implementation does not conform to the flow above, and these things are left to be done:
- [ ] Periodic reproviding for roots(https://github.com/lazyledger/lazyledger-core/issues/393)
- [ ] Providing for non-producer nodes that retrieved a full block(https://github.com/lazyledger/lazyledger-core/issues/394)

### Node churn
Records of someone hosting data are stored on peers selected not by their qualities but by the simple XOR metric. Unfortunately, this eventually makes different light clients store those records unreliably, as they are not meant to be full-featured daemons. Therefore, some data may become undiscoverable for some period of time.

#### Solutions
* Basically, reproviding helps here. However, we never know when a light client leaves and data may be undiscoverable for many hours until the next reprovide happens, which would store records on another node.
* Full routing table DHT client can keep records besides ones chosen by XOR metric, and the nodes running it are expected to be reliable. Thus they can fill a gap of undiscoverable hours.

### Providing Time
We need to ensure providing takes less time than the time between two subsequent block proposals by a node. Otherwise, DHT providing wouldn't keep up with block production, creating an evergrowing providing queue. Unfortunately, for the standard DHT client, providing can take up to 3 mins on a large-scale network.

[...]

PR #1037, opened 2022-08-23
docs(adr): BlockSync Overhaul ADR #011 - Part 1

Wondertan:
[Rendered](https://github.com/celestiaorg/celestia-node/blob/hlib/docs/adr-11/docs/adr/adr-11-blocksync-overhaul.md)

### DIFF
--- a/docs/adr/adr-11-blocksync-overhaul.md
+++ b/docs/adr/adr-11-blocksync-overhaul.md
@@ -19,6 +19,7 @@
 - FN - Full Node
 - BN - Bridge Node
 - EDS(Extended Data Square) - plain Block data omitting headers and other metadata.
+- NMT - Namespaced Merkle Tree
 
 ## Context
 
@@ -86,6 +87,8 @@ or proposing custom Bitswap message extention.
 - __New libp2p based Exchange protocol is introduced for data/share exchange purposes.__ FNs/BNs servers LNs clients.
 
 ### Detailed Design
+
+> All the comments on the API definitions should be preserved and potentially improved by implementations.
 
 #### Block/EDS Storage
 
@@ -176,8 +179,8 @@ so this is out of scope of the document.
 // Put stores the given data square with DataRoot as key.
 //
 // The square is verified on the Exchange level and Put only stores the square trusting it.
-// Put serializes the full EDS into a CARv1 file internally and mounts it onto the DAGStore.
 // The resulting file stores all the shares and NMT Merkle Proofs of the EDS.
+// Additionally, the file gets indexed s.t. Store.Blockstore can access them. 
 func (s *Store) Put(context.Context, DataRoot, *rsmt2d.ExtendedDataSquare) error
 ```
 
@@ -247,10 +250,13 @@ To remove stored EDS `Remove` methods is introduced. Internally it:
   - `DAGStore` cleans up indices itself
 - Removes FS/File Mount of CARv1 file from disk
 
-NOTE: It's not necessary, but an API ergonomics/symmetry nice-to-have
+NOTES:
+
+- It's not necessary, but an API ergonomics/symmetry nice-to-have
+- GC logic on the DAGStore has to be investigated so that Removing is correctly implemented
 
 ```go
-// Remove removes EDS from Store by given DataRoot.
+// Remove removes EDS from Store by given DataRoot and cleans up all the indexing.
 func (s *Store) Remove(context.Context, DataRoot) error
 
 ```
--- a/docs/adr/adr-11-blocksync-overhaul.md
+++ b/docs/adr/adr-11-blocksync-overhaul.md
@@ -56,9 +56,9 @@ in 2 days to match the following requirements:
 
 ### Decision
 
-This ADR is intended to outline design decisions for block data storage. In a nutshell, the decision is to use 
-___[CAR format](https://ipld.io/specs/transport/car/carv2/)___ and ___[Dagstore](https://github.com/filecoin-project/dagstore)___ 
-for ___extended block storage___ and ___custom p2p Req/Resp protocol for block data syncing___(whole block and data by 
+This ADR is intended to outline design decisions for block data storage. In a nutshell, the decision is to use
+___[CAR format](https://ipld.io/specs/transport/car/carv2/)___ and ___[Dagstore](https://github.com/filecoin-project/dagstore)___
+for ___extended block storage___ and ___custom p2p Req/Resp protocol for block data syncing___(whole block and data by
 namespace id) in the happy path. The p2p portion of the document will come in the subsequent Part II document.
 
 #### Key Design Decisions
@@ -215,8 +215,7 @@ func (s *Store) Put(context.Context, DataRoot, *rsmt2d.ExtendedDataSquare) error
 
 To read an EDS as a byte stream `GetCAR` method is introduced. Internally it
 
-- Converts `DataRoot`'s hash into the [`shard.Key`](<https://github.com/filecoin-project/dagstore/blob/master/shard/key.go#L12>
-- Gets Mount by `shard.Key` from [`mount.Registry`](https://github.com/filecoin-project/dagstore/blob/master/mount/registry.go#L22)
+- Converts `DataRoot`'s hash into the [`shard.Key`](https://github.com/filecoin-project/dagstore/blob/master/shard/key.go#L12)
 - Acquires `ShardAccessor` and returns `io.ReadCloser` from it
 
 _NOTES:_
@@ -277,8 +276,9 @@ func (s *Store) Get(context.Context, DataRoot) (*rsmt2d.ExtendedDataSquare, erro
 
 To check if EDSStore keeps an EDS `Has` method is introduced. Internally it:
 
-- Converts `DataRoot`'s hash into the [`shard.Key`](<https://github.com/filecoin-project/dagstore/blob/master/shard/key.go#L12>
-- Tries to acquire `ShardAccessor` and on success closes it returning postive.
+- Converts `DataRoot`'s hash into the [`shard.Key`](https://github.com/filecoin-project/dagstore/blob/master/shard/key.go#L12)
+- Checks if [`GetShardInfo`](https://github.com/filecoin-project/dagstore/blob/master/dagstore.go#L483) does not return
+[ErrShardUnknown](https://github.com/filecoin-project/dagstore/blob/eac7733212fdd7c80be5078659f7450b3956d2a6/dagstore.go#L55)
 
 _NOTE: It's not necessary, but an API ergonomics/symmetry nice-to-have._
 
@@ -290,11 +290,12 @@ func (s *Store) Has(context.Context, DataRoot) (bool, error)
 
 ##### `share.EDSStore.Remove`
 
-To remove stored EDS `Remove` methods is introduced. Internally it:
+To remove stored EDS `Remove` method is introduced. Internally it:
 
-- Converts `DataRoot`'s hash into the [`shard.Key`](<https://github.com/filecoin-project/dagstore/blob/master/shard/key.go#L12>
-- Removes respective `Mount` and `Shard` from `DAGStore`
-- Removes `FileMount` file of CARv1 file from disk under `storepath/DataRoot.Hash` path
+- Converts `DataRoot`'s hash into the [`shard.Key`](https://github.com/filecoin-project/dagstore/blob/master/shard/key.go#L12)
+- Destroys `Shard` via `DAGStore`
+  - Internally removes it's `Mount` as well
+- Removes CARv1 file from disk under `storepath/DataRoot.Hash` path
 
 _NOTES:_
 
--- a/docs/adr/adr-11-blocksync-overhaul.md
+++ b/docs/adr/adr-11-blocksync-overhaul.md
@@ -69,14 +69,14 @@ sequentially in one read and transferred over the wire. Additionally, [CARv2](ht
 introduces pluggable indexes over the blob allowing efficient random access to shares and NMT Proofs in one read
 (if the index is cached in memory).
 
-- __FNs/BNs manage a top-level index for _hash_ to _CAR block file_ mapping.__ Current DASing for LNs requires FNs/BNs 
-to serve simple hash to data requests. The top-level index maps any hash to any block CARv1file so that FNs/BNs can 
+- __FNs/BNs manage a top-level index for _hash_ to _CAR block file_ mapping.__ Current DASing for LNs requires FNs/BNs
+to serve simple hash to data requests. The top-level index maps any hash to any block CARv1file so that FNs/BNs can
 quickly serve requests.
 
-- __FNs/BNs run a single instance of [`DAGStore`](https://github.com/filecoin-project/dagstore) to manage CAR block 
+- __FNs/BNs run a single instance of [`DAGStore`](https://github.com/filecoin-project/dagstore) to manage CAR block
 files.__ DAGStore provides both the top-level indexing and CARv2 based indexing per each CAR file. In essence, it's an
 engine for managing any CAR files with indexing, convenient abstractions, tools, recovery mechanisms, etc.
-  - __EDSes as _CARv1_ files over _CARv2_.__ CARv2 encodes indexes into the file, while DAGStore maintains CARv2 based 
+  - __EDSes as _CARv1_ files over _CARv2_.__ CARv2 encodes indexes into the file, while DAGStore maintains CARv2 based
 indexing as well. Usage of CARv1 keeps only one copy of index, stores/transfers less metadata per EDS and simplifies
 reading EDS from file.
 
@@ -109,14 +109,14 @@ chosen.
 To write EDS into a stream/file `WriteEDS` is introduced.
 // TODO Describe logic
 
-NOTE: CAR provides [a utility](https://github.com/ipld/go-car/blob/master/car.go#L47) to serialize any DAG into the file
+_NOTE: CAR provides [a utility](https://github.com/ipld/go-car/blob/master/car.go#L47) to serialize any DAG into the file
 and there is a way to serialize EDS into DAG(`share/ipld.ImportShares`). This approach is the simplest and traverses
 shares and Merkle Proofs in depth-first manner packing them in a CAR file. However, this is incompatible with the
 requirement of being able to truncate the CAR file to read out __only__ the first quadrant out of it without NMT proofs,
-so serialization must be different from the utility to support that.
+so serialization must be different from the utility to support that._
 
-NOTE2: Alternatively to `WriteEDS`, and `EDSReader` could be introduced to make EDS-to-stream handling more idiomatic
-and efficient in some cases, with the cost of more complex implementation.
+_NOTE2: Alternatively to `WriteEDS`, and `EDSReader` could be introduced to make EDS-to-stream handling more idiomatic
+and efficient in some cases, with the cost of more complex implementation._
 
 ```go
 // WriteEDS writes whole EDS into given io.Writer as CARv1 file.
@@ -152,8 +152,6 @@ to every share and/or Merkle proof over every registered CARv1 file. The `EDSSto
 implementation to achieve the access. The main use-case is randomized sampling over the whole chain of EDS block data
 and getting data by namespace.
 
-EDSStore constructor should instantiate
-
 ```go
 type EDSStore struct {
  basepath string 
@@ -170,16 +168,23 @@ func NewEDSStore(basepath string, ds datastore.Batching) *EDSStore {
  topIdx := index.NewInverted(datastore)
  carIdx := index.NewFSRepo(basepath + "/index")
  mounts := mount.NewRegistry()
+ r := mount.NewRegistry()
+ err = r.Register("fs", &mount.FSMount{FS: os.DirFS(basePath + "/eds/")}) // registration is a must 
+ if err != nil {
+  panic(err)
+ }
+ 
  return &EDSStore{
   basepath: basepath,
   dgst: dagstore.New(dagstore.Config{...}),
   topIdx: index.NewInverted(datastore),
   carIdx: index.NewFSRepo(basepath + "/index")
   mounts: mounts,
-    }   
+ }   
 }
-
 ```
+
+_NOTE: EDSStore should have lifecycle methods(Start/Stop)._
 
 ##### `share.EDSStore.Put`
 
@@ -190,11 +195,9 @@ To write an entire EDS `Put` method is introduced. Internally it
 - Wraps it with `DAGStore`'s [FileMount](https://github.com/filecoin-project/dagstore/blob/master/mount/file.go#L10)
 - Converts `DataRoot` into the [`shard.Key`](https://github.com/filecoin-project/dagstore/blob/master/shard/key.go#L12)
 - Registers the `Mount` as a `Shard` on the `DAGStore`
-  - This register the `Mount` in [`mount.Registry`](https://github.com/filecoin-project/dagstore/blob/master/mount/registry.go#L22)
-passed to `DAGStore`.
 
-NOTE: Registering on the DAGStore automatically populates top-level index with shares/proofs accessible from stored EDS
-so this is out of scope of the document.
+_NOTE: Registering on the DAGStore automatically populates top-level index with shares/proofs accessible from stored EDS
+so this is out of scope of the document._
 
 ```go
 // Put stores the given data square with DataRoot as key.
@@ -213,8 +216,8 @@ To read an EDS as a byte stream `GetCAR` method is introduced. Internally it
 - Gets Mount by `shard.Key` from [`mount.Registry`](https://github.com/filecoin-project/dagstore/blob/master/mount/registry.go#L22)
 - Returns `io.ReadeCloser` from [`Mount.Fetch`](https://github.com/filecoin-project/dagstore/blob/master/mount/mount.go#L71)
 
-NOTE: It might be necessary to acquire EDS mount via `DAGStore` keeping `ShardAccessor`
-and closing it when operation is done. This has to be confirmed.
+_NOTE: It might be necessary to acquire EDS mount via `DAGStore` keeping `ShardAccessor`
+and closing it when operation is done. This has to be confirmed._
 
 ```go
 // GetCAR takes a DataRoot and returns a buffered reader to the respective EDS serialized as CARv1 file.
@@ -235,7 +238,7 @@ by namespace. // TODO Link to the section
 There is a [frozen/un-merged implementation](https://github.com/filecoin-project/dagstore/pull/116) of `Blockstore`
 over `DAGStore` and CARv2 indexes.
 
-NOTE: We can either use it(and finish if something is missing for our case) or implement custom optimized for our needs.
+_NOTE: We can either use it(and finish if something is missing for our case) or implement custom optimized for our needs._
 
 ```go
 // Blockstore returns an IPFS Blockstore providing access to individual shares/nodes of all EDS 
@@ -252,7 +255,7 @@ To read an entire EDS `Get` method is introduced. Internally it:
 - Gets a serialized EDS `io.Reader` via `Store.GetCAR`
 - Deserializes the EDS and validates it via `share.ReadEDS`
 
-NOTE: It's not necessary, but an API ergonomics/symmetry nice-to-have
+_NOTE: It's not necessary, but an API ergonomics/symmetry nice-to-have._
 
 ```go
 // Get reads EDS out of Store by given DataRoot.
@@ -268,13 +271,12 @@ To remove stored EDS `Remove` methods is introduced. Internally it:
 
 - Converts `DataRoot` into the [`shard.Key`](<https://github.com/filecoin-project/dagstore/blob/master/shard/key.go#L12>
 - Removes respective `Mount` and `Shard` from `DAGStore`
-  - `DAGStore` cleans up indices and `mount.Registery` itself
 - Removes `FileMount` file of CARv1 file from disk under `storepath/DataRoot` path
 
-NOTES:
+_NOTES:_
 
-- It's not necessary, but an API ergonomics/symmetry nice-to-have
-- GC logic on the DAGStore has to be investigated so that Removing is correctly implemented
+- _It's not necessary, but an API ergonomics/symmetry nice-to-have_
+- _GC logic on the DAGStore has to be investigated so that Removing is correctly implemented_
 
 ```go
 // Remove removes EDS from Store by given DataRoot and cleans up all the indexing.
--- a/docs/adr/adr-11-blocksync-overhaul.md
+++ b/docs/adr/adr-11-blocksync-overhaul.md
@@ -20,6 +20,8 @@
 - BN - Bridge Node
 - EDS(Extended Data Square) - plain Block data omitting headers and other metadata.
 - NMT - Namespaced Merkle Tree
+- [DataRoot](https://github.com/celestiaorg/celestia-node/blob/main/service/share/share.go#L35) - alias for 
+[DAHeader](https://github.com/celestiaorg/celestia-core/blob/v0.34.x-celestia/pkg/da/data_availability_header.go#L29)
 
 ## Context
 
@@ -70,8 +72,8 @@ introduces pluggable indexes over the blob allowing efficient random access to s
 (if the index is cached in memory).
 
 - __FNs/BNs manage a top-level index for _hash_ to _CAR block file_ mapping.__ Current DASing for LNs requires FNs/BNs
-to serve simple hash to data requests. The top-level index maps any hash to any block CARv1file so that FNs/BNs can
-quickly serve requests.
+to serve simple hash to data requests. The top-level index maps any hash of Share/NMT Proof to any block CARv1 file so 
+that FNs/BNs can quickly serve DA requests.
 
 - __FNs/BNs run a single instance of [`DAGStore`](https://github.com/filecoin-project/dagstore) to manage CAR block
 files.__ DAGStore provides both the top-level indexing and CARv2 based indexing per each CAR file. In essence, it's an
@@ -128,13 +130,13 @@ func WriteEDS(context.Context, *rsmt2d.ExtendedDataSquare, io.Writer) error
 
 To read an EDS out of a stream/file, `ReadEDS` is introduced. Internally, it
 
-- Imports EDS with an empty pre-allocated slice. NOTE: Size can be taken from DataRoot
+- Imports EDS with an empty pre-allocated slice. _NOTE: Size can be taken from DataRoot Row/Col len_
 - Wraps given io.Reader with [`BlockReader`](https://github.com/ipld/go-car/blob/master/v2/block_reader.go#L16)
 - Reads out blocks one by one and fills up the EDS quadrant via `EDS.SetCell`
 - Recomputes and validates via `EDS.Repair`
 
 ```go
-// ReadEDS reads an EDS quadrant(1/4) from an io.Reader CAR file.//
+// ReadEDS reads an EDS quadrant(1/4) from an io.Reader CAR file.
 // It expects strictly first EDS quadrant(top left).
 // The returned EDS is guaranteed to be full and valid against the DataRoot, otherwise ReadEDS errors.
 func ReadEDS(context.Context, io.Reader, DataRoot) (*rsmt2d.ExtendedDataSquare, error)
@@ -190,17 +192,17 @@ _NOTE: EDSStore should have lifecycle methods(Start/Stop)._
 
 To write an entire EDS `Put` method is introduced. Internally it
 
-- Opens a file under `storepath/DataRoot` path
+- Opens a file under `storepath/DataRoot.Hash` path
 - Serializes the EDS into the file via `share.WriteEDS`
 - Wraps it with `DAGStore`'s [FileMount](https://github.com/filecoin-project/dagstore/blob/master/mount/file.go#L10)
-- Converts `DataRoot` into the [`shard.Key`](https://github.com/filecoin-project/dagstore/blob/master/shard/key.go#L12)
+- Converts `DataRoot`'s hash into the [`shard.Key`](https://github.com/filecoin-project/dagstore/blob/master/shard/key.go#L12)
 - Registers the `Mount` as a `Shard` on the `DAGStore`
 
 _NOTE: Registering on the DAGStore automatically populates top-level index with shares/proofs accessible from stored EDS
 so this is out of scope of the document._
 
 ```go
-// Put stores the given data square with DataRoot as key.
+// Put stores the given data square with DataRoot's hash as key.
 //
 // The square is verified on the Exchange level and Put only stores the square trusting it.
 // The resulting file stores all the shares and NMT Merkle Proofs of the EDS.
@@ -212,7 +214,7 @@ func (s *Store) Put(context.Context, DataRoot, *rsmt2d.ExtendedDataSquare) error
 
 To read an EDS as a byte stream `GetCAR` method is introduced. Internally it
 
-- Converts `DataRoot` into the [`shard.Key`](<https://github.com/filecoin-project/dagstore/blob/master/shard/key.go#L12>
+- Converts `DataRoot`'s hash into the [`shard.Key`](<https://github.com/filecoin-project/dagstore/blob/master/shard/key.go#L12>
 - Gets Mount by `shard.Key` from [`mount.Registry`](https://github.com/filecoin-project/dagstore/blob/master/mount/registry.go#L22)
 - Returns `io.ReadeCloser` from [`Mount.Fetch`](https://github.com/filecoin-project/dagstore/blob/master/mount/mount.go#L71)
 
@@ -269,9 +271,9 @@ func (s *Store) Get(context.Context, DataRoot) (*rsmt2d.ExtendedDataSquare, erro
 
 To remove stored EDS `Remove` methods is introduced. Internally it:
 
-- Converts `DataRoot` into the [`shard.Key`](<https://github.com/filecoin-project/dagstore/blob/master/shard/key.go#L12>
+- Converts `DataRoot`'s hash into the [`shard.Key`](<https://github.com/filecoin-project/dagstore/blob/master/shard/key.go#L12>
 - Removes respective `Mount` and `Shard` from `DAGStore`
-- Removes `FileMount` file of CARv1 file from disk under `storepath/DataRoot` path
+- Removes `FileMount` file of CARv1 file from disk under `storepath/DataRoot.Hash` path
 
 _NOTES:_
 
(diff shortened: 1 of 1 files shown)


docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,329 @@
...
+// WriteEDS writes whole EDS into given io.Writer as CARv1 file.
+// All its shares and recomputed NMT proofs.
+func WriteEDS(context.Context, *rsmt2d.ExtendedDataSquare, io.Writer) error
+```
+
+###### `share.ReadEDS`
+
+To read EDS out of stream/file `ReadEDS` is introduced. Internally, it
+
+- Imports EDS with an empty pre-allocated slice. NOTE: Size can be taken from DataRoot
```
distractedm1nd, line 132: Is the dataroot not just a merkle hash? I know that if we have the length of rowroots/colroots we know how much data there is, but can you help me understand what we can do to go from root -> length here?

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,329 @@
...
+- Wraps given io.Reader with [`BlockReader`](https://github.com/ipld/go-car/blob/master/v2/block_reader.go#L16)
+- Reads out blocks one by one and fills up the EDS quadrant via `EDS.SetCell`
+- Recomputes and validates via `EDS.Repair`
+
+```go
+// ReadEDS reads EDS quadrant(1/4) from io.Reader CAR file.
```
distractedm1nd, line 138:
```suggestion
// ReadEDS reads an EDS quadrant(1/4) from an io.Reader CAR file.
```

docs/adr/adr-11-blocksync-overhaul.md
distractedm1nd: Do you mean by first opening the CAR? (since the dataroot is in the top level index)

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,329 @@
...
+//
+// It expects strictly first EDS quadrant(top left).
```
distractedm1nd, line 140: I do not think that ReadEDS should be passed an io.Reader that is already responsible for only reading out the part of the file that ReadEDS needs - ReadEDS sounds like it should handle this logic itself

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,329 @@
...
+// Returned EDS is guaranteed to be full and valid against DataRoot, otherwise ReadEDS errors.
```
distractedm1nd, line 141:
```suggestion
// The returned EDS is guaranteed to be full and valid against the DataRoot, otherwise ReadEDS errors.
```

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,329 @@
...
+func ReadEDS(context.Context, io.Reader, DataRoot) (*rsmt2d.ExtendedDataSquare, error)
+```
+
+##### `share.EDSStore`
+
+To manage every EDS on the disk FNs/BNs keep `EDSStore` type is introduced in `share` pkg. Each EDS together with its
```
distractedm1nd, line 147:
```suggestion
To manage every EDS on the disk, FNs/BNs keep an `EDSStore`. The `EDSStore` type is introduced in the `share` pkg. Each EDS together with its
```

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,329 @@
...
+Merkle Proofs serializes into CARv1 file. All the serialized CARv1 file blobs are stored as OS FS files as DAGStore
```
distractedm1nd, line 148: small nuance is just that the mounts only exist during registration
```suggestion
Merkle Proofs serializes into CARv1 file. All the serialized CARv1 file blobs are stored as OS FS files, and registered on the DAGStore using
```

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,329 @@
...
+[Mounts](https://github.com/filecoin-project/dagstore/blob/master/mount/mount.go).
+
+The introduced `EDSStore` also maintains (via DAGStore) a top-level index enabling granular and efficient random access
+to every share and/or Merkle proof over every registered CARv1 file. The `EDSStore` provides a custom `Blockstore` interface
+implementation to achieve the access. The main use-case is randomized sampling over the whole chain of EDS block data
```
distractedm1nd, line 153: We should mention that the Blockstore interface implementation is only partial

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,329 @@
...
+and getting data by namespace.
+
+EDSStore constructor should instantiate
+
+```go
+type EDSStore struct {
+ basepath string 
+ dgstr dagstore.DAGStore
+ topIdx index.Inverted
```
distractedm1nd, line 162: The inverted index is already exposed through dgstr

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,329 @@
...
+
+EDSStore constructor should instantiate
+
+```go
+type EDSStore struct {
+ basepath string 
+ dgstr dagstore.DAGStore
+ topIdx index.Inverted 
+ carIdx index.FullIndexRepo
+ mounts  *mount.Registry
```
distractedm1nd, line 164: I don't think we need to store the registry here (we don't need to access it outside of instantiation). It is only there for registering what kind of mounts are valid (you still need to make a new mount for every car anyways).

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,329 @@
...
+ ...
+}
+
+// NewEDSStore constructs EDStore over OS directory to store CARv1 files of EDSes and indices for them.
+// Datastore is used to keep inverted/top-level index.
+func NewEDSStore(basepath string, ds datastore.Batching) *EDSStore {
```
distractedm1nd, line 170: Can I make this instantiation a little more close to reality or is that out of scope of the ADR? It would only add a few lines and is helpful for understanding I think (adding the FS mount to the registry, passing the relevant things to the EDS store config, and starting the dagstore before returning)

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,329 @@
...
+func ReadEDS(context.Context, io.Reader, DataRoot) (*rsmt2d.ExtendedDataSquare, error)
+```
+
+##### `share.EDSStore`
+
+To manage every EDS on the disk FNs/BNs keep `EDSStore` type is introduced in `share` pkg. Each EDS together with its
+Merkle Proofs serializes into CARv1 file. All the serialized CARv1 file blobs are stored as OS FS files as DAGStore
+[Mounts](https://github.com/filecoin-project/dagstore/blob/master/mount/mount.go).
+
+The introduced `EDSStore` also maintains (via DAGStore) a top-level index enabling granular and efficient random access
```
distractedm1nd, line 151: We need to add a manual LRU cache to the EDSStore (can't wrap this blockstore implementation with a cache like we are currently doing). It caches the ShardAccessors and CAR-level blockstores. A hacky version of this is in my PR if you wanna look. Would this be something mentioned in the ADR?

docs/adr/adr-11-blocksync-overhaul.md
distractedm1nd: Also by the way. I think we can't use dataroot as a primary index, since it is not unique (I thought it was, but it isn't, in the case of empty data, right?). I am currently using it as a filename but it isn't working for this reason

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,329 @@
...
+    }   
+}
+
+```
+
+##### `share.EDSStore.Put`
+
+To write an entire EDS `Put` method is introduced. Internally it
+
+- Opens a file under `storepath/DataRoot` path
```
distractedm1nd, line 189:
```suggestion
- Opens a file under `storepath/DataRoot` path using `OpenReadWrite` creating a CAR-level blockstore
```

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,329 @@
...
+- Serializes the EDS into the file via `share.WriteEDS`
+- Wraps it with `DAGStore`'s [FileMount](https://github.com/filecoin-project/dagstore/blob/master/mount/file.go#L10)
+- Converts `DataRoot` into the [`shard.Key`](https://github.com/filecoin-project/dagstore/blob/master/shard/key.go#L12)
+- Registers the `Mount` as a `Shard` on the `DAGStore`
+  - This register the `Mount` in [`mount.Registry`](https://github.com/filecoin-project/dagstore/blob/master/mount/registry.go#L22)
```
distractedm1nd, line 194: This is incorrect

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,329 @@
...
+NOTE: Registering on the DAGStore automatically populates top-level index with shares/proofs accessible from stored EDS
+so this is out of scope of the document.
+
+```go
+// Put stores the given data square with DataRoot as key.
+//
+// The square is verified on the Exchange level and Put only stores the square trusting it.
+// The resulting file stores all the shares and NMT Merkle Proofs of the EDS.
+// Additionally, the file gets indexed s.t. Store.Blockstore can access them. 
+func (s *Store) Put(context.Context, DataRoot, *rsmt2d.ExtendedDataSquare) error
```
distractedm1nd, line 206: Where is this called from?

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,329 @@
...
+```
+
+##### `share.EDSStore.GetCAR`
+
+To read an EDS as a byte stream `GetCAR` method is introduced. Internally it
+
+- Converts `DataRoot` into the [`shard.Key`](<https://github.com/filecoin-project/dagstore/blob/master/shard/key.go#L12>
+- Gets Mount by `shard.Key` from [`mount.Registry`](https://github.com/filecoin-project/dagstore/blob/master/mount/registry.go#L22)
+- Returns `io.ReadeCloser` from [`Mount.Fetch`](https://github.com/filecoin-project/dagstore/blob/master/mount/mount.go#L71)
```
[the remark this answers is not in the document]
distractedm1nd: Also, we only get an io.Reader and just close the shard accessor

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,329 @@
...
+
+NOTE: It might be necessary to acquire EDS mount via `DAGStore` keeping `ShardAccessor`
+and closing it when operation is done. This has to be confirmed.
+
+```go
+// GetCAR takes DataRoot and returns a buffered reader to respective EDS serialized as CARv1 file.
```
distractedm1nd, line 221:
```suggestion
// GetCAR takes a DataRoot and returns a buffered reader to the respective EDS serialized as CARv1 file.
```

docs/adr/adr-11-blocksync-overhaul.md
distractedm1nd: Ok I read further down that `GetCar` should be responsible for this - I still think GetCAR should return a reader over the entire CAR then

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,329 @@
...
+##### `share.EDSStore.Blockstore`
+
+`Blockstore` method return a `Blockstore` interface implementation instance, providing random access over share and NMT
+Merkle proof in every stored EDS. It is required for FNs/BNs to serve DAS requests over the Bitswap and for reading data
+by namespace. // TODO Link to the section
+
+There is a [frozen/un-merged implementation](https://github.com/filecoin-project/dagstore/pull/116) of `Blockstore`
+over `DAGStore` and CARv2 indexes.
+
+NOTE: We can either use it(and finish if something is missing for our case) or implement custom optimized for our needs.
```
distractedm1nd, line 239: It does not really fit our case exactly, I think we should rather implement a custom one (and it will probably look pretty similar to what i have in the draft PR already anyways, so there is only very little effort needed for this)

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,329 @@
...
+
+```go
+// Blockstore returns an IPFS Blockstore providing access to individual shares/nodes of all EDS 
+// registered on the Store. NOTE: The Blockstore does not store full Celestia Blocks, but IPFS blocks.
```
distractedm1nd, line 243: I feel like this NOTE should be mentioned in the above markdown. It is actually a very important point and one I didn't understand for a while myself.

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,329 @@
...
+// We represent `shares` and NMT Merkle proofs as IPFS blocks and IPLD nodes, so that Bitswap can access those.
+func (s *Store) Blockstore() blockstore.Blockstore
+
+```
+
+##### `share.EDSStore.Get`
+
+To read an entire EDS `Get` method is introduced. Internally it:
+
+- Gets serialized EDS `io.Reader` via `Store.GetCAR`
```
distractedm1nd, line 253:
```suggestion
- Gets a serialized EDS `io.Reader` via `Store.GetCAR`
```

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,329 @@
...
+- Deserializes EDS and validates it via `share.ReadEDS`
```
distractedm1nd, line 254:
```suggestion
- Deserializes the EDS and validates it via `share.ReadEDS`
```

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,329 @@
...
+
+NOTE: It's not necessary, but an API ergonomics/symmetry nice-to-have
```
distractedm1nd, line 256: I do think this is necessary! Not just nice-to-have :) Something would very much be missing without it

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,329 @@
...
+
+```
+
+##### `share.EDSStore.Remove`
+
+To remove stored EDS `Remove` methods is introduced. Internally it:
+
+- Converts `DataRoot` into the [`shard.Key`](<https://github.com/filecoin-project/dagstore/blob/master/shard/key.go#L12>
+- Removes respective `Mount` and `Shard` from `DAGStore`
+  - `DAGStore` cleans up indices and `mount.Registery` itself
```
distractedm1nd, line 272: doesn't need to do anything on the registry, or remove any mount

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,329 @@
...
+
+#### Key Design Decisions
+
+- __FNs/BNs store EDSes serialized as [CAR files](https://ipld.io/specs/transport/car).__ CAR format provides an
+efficient way to store Merkle DAG data, like EDS with NMT. It packs such DAG data into a single blob which can be read
+sequentially in one read and transferred over the wire. Additionally, [CARv2](https://ipld.io/specs/transport/car/carv2/)
+introduces pluggable indexes over the blob allowing efficient random access to shares and NMT Proofs in one read
+(if the index is cached in memory).
+
+  - __EDSes as _CARv1_ files over _CARv2_.__ CARv2 encodes indexes into the file, however they should not be transferred
```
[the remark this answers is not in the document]
distractedm1nd: Also, why keep a FullIndexRepo when you don't want to use CAR level indexes?

distractedm1nd (changes requested):
Absolutely amazing! Looking great. Excited for the diagrams as well

I mainly have questions, but also had some comments from what I learned from my experiments. And some opinions I want to debate on

Still some open questions on my mind but I will think about them and extend my review later

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,329 @@
...
+in case of EDS, so keeping them separately is a better strategy. Also CARv2 takes more space for metadata which is not
+needed in our case.
+
+- __FNs/BNs manage a top-level index for _hash_ to _CARv1 block file_ mapping.__ Current DASing for LNs requires FNs/BNs to serve
```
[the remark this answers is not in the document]
distractedm1nd: i mentioned it somewhere else further down but the dataroot is not unique, correct?

docs/adr/adr-11-blocksync-overhaul.md
distractedm1nd: oops, not starting the dagstore, since I forgot this will be done by the lifecycle

docs/adr/adr-11-blocksync-overhaul.md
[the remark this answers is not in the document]
Wondertan: FullIndexRepo is the reason we don't need to store EDS as CARv2. FullIndexRepo maintains a [_separate_](https://github.com/filecoin-project/dagstore/blob/master/index/repo_fs.go#L30) and just uses CARv2 code. Basically, storing EDS as CARv2 means having two CAR level indices: in the FullIndexRepo and in the CARv2 file itself, which we don't want.

docs/adr/adr-11-blocksync-overhaul.md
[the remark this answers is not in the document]
Wondertan: If the data root happens to be the same for two different blocks, there is no reason to store the data twice,  e.g. empty block.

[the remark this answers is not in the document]
Wondertan: > Can we clarify which hash/cid we mean here?

Of each share and NMT Proof. That's mentioned multiple times

docs/adr/adr-11-blocksync-overhaul.md
Wondertan: It's not unique, but we don't need to store each empty block as its the same data.

docs/adr/adr-11-blocksync-overhaul.md
Wondertan: This Reader is the data that is served over the wire. If we make Reader read the whole EDS, then it means the whole EDS is transferred over the wire, which is what we want to avoid. We can, though, engineer something smart that Reads only quadrant out of full Reader, but that's tricky, adds complexity(not justified), and anyway, this logic has to live somewhere, like in Store, or in protocol

docs/adr/adr-11-blocksync-overhaul.md
Wondertan: If we use the one from DAGStore, then it is not.

docs/adr/adr-11-blocksync-overhaul.md
Wondertan: Verbosity does not hurt here. Especially when EDSStore owns the index and passes is to the DAGStore

docs/adr/adr-11-blocksync-overhaul.md
Wondertan: We need to access mounts to read CAR files from them. Or do you know another way?

docs/adr/adr-11-blocksync-overhaul.md
Wondertan: Yes 
> if you wanna look

We should do it we go with a custom implementation of Blockstore, what we have to decide on btw
> Would this be something mentioned in the ADR?

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,329 @@
...
+}
+
+```
+
+##### `share.EDSStore.Put`
+
+To write an entire EDS `Put` method is introduced. Internally it
+
+- Opens a file under `storepath/DataRoot` path
+- Serializes the EDS into the file via `share.WriteEDS`
```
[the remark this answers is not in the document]
Wondertan: DataRoot is passed to the Put

docs/adr/adr-11-blocksync-overhaul.md
Wondertan: Let's add this. Happy to review your suggestion

docs/adr/adr-11-blocksync-overhaul.md
Wondertan: From the Networking layer

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,329 @@
...
+// Additionally, the file gets indexed s.t. Store.Blockstore can access them. 
+func (s *Store) Put(context.Context, DataRoot, *rsmt2d.ExtendedDataSquare) error
+```
+
+##### `share.EDSStore.GetCAR`
+
+To read an EDS as a byte stream `GetCAR` method is introduced. Internally it
+
+- Converts `DataRoot` into the [`shard.Key`](<https://github.com/filecoin-project/dagstore/blob/master/shard/key.go#L12>
+- Gets Mount by `shard.Key` from [`mount.Registry`](https://github.com/filecoin-project/dagstore/blob/master/mount/registry.go#L22)
```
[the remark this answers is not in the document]
Wondertan: Hmm, where does AcquireShard returns a [Reader](https://github.com/filecoin-project/dagstore/blob/master/dagstore.go#L435). There is a result channel, but it only returns ShardAccessor which is not the Reader but a Store basically.

docs/adr/adr-11-blocksync-overhaul.md
Wondertan: Checking...

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,329 @@
...
+- Returns `io.ReadeCloser` from [`Mount.Fetch`](https://github.com/filecoin-project/dagstore/blob/master/mount/mount.go#L71)
+
+NOTE: It might be necessary to acquire EDS mount via `DAGStore` keeping `ShardAccessor`
```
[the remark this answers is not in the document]
Wondertan: It's starting to feel uncertain about this when you realize that Acquire does not return you a Reader

docs/adr/adr-11-blocksync-overhaul.md
Wondertan: I mean for the networking protocol and its not required and it would use just `GetCAR`

docs/adr/adr-11-blocksync-overhaul.md
Wondertan: Related to the discussions re getting car Reader above

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,329 @@
...
+func (s *Store) Remove(context.Context, DataRoot) error
+
+```
+
+#### Reading Data By Namespace
+
+Generally stays unchanged with minor edits:
+
+- `share/ipld.GetByNamespace` is kept to load data from disk only and not from network anymore
+  - Using `Blockservice` with offline exchange // TODO Links
```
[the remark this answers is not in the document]
Wondertan: We can enable this, but the case where it is missing something locally should never happen, otherwise, it does not store the block. 

The protocol above will respond as not found in this case, and the requesting client will go ask someone else. This is a better strategy than waiting for the Server to find somewhere else data it does not have.

docs/adr/adr-11-blocksync-overhaul.md
distractedm1nd: It is tho 

Not all methods are supported operations (put doesn't make any sense for example)

docs/adr/adr-11-blocksync-overhaul.md
distractedm1nd: you still need to create a new mount for every single car file, but not every single one of these mounts is registered in the mount registry (just the type of mount)

docs/adr/adr-11-blocksync-overhaul.md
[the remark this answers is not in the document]
distractedm1nd: Thinking if you really need the reader here we should just extend the interface to return it

docs/adr/adr-11-blocksync-overhaul.md
distractedm1nd: ok that was the secondary question, can you answer the one about how we get to the size?

docs/adr/adr-11-blocksync-overhaul.md
distractedm1nd: hmmm I would keep it only if we actually access it somewhere, it is unnecessary otherwise. of course the second we need it, it will be trivial to add to the struct

docs/adr/adr-11-blocksync-overhaul.md
distractedm1nd: the rationale is in the go doc of my experimentations i'll post it when i have PC access again

And: we should have absolutely 0 bias towards keeping what I've already done, if we throw it all away I still gained very very much by doing it! The hope is of course that it could uncover some details that would help the ADR or downstream impl, but that's really it

docs/adr/adr-11-blocksync-overhaul.md
distractedm1nd: I looked into it again, yes the intended way to create mounts is with `r.Instantiate`, so we should indeed keep it

BUT we definitely only have to call register once, unless we add different schemes (but this is in a different review comment)

docs/adr/adr-11-blocksync-overhaul.md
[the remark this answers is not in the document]
distractedm1nd: my POV here is that caching shardaccessors is really nice because it allows you to access the underlying blockstore, which is at least good for bitswap

docs/adr/adr-11-blocksync-overhaul.md
distractedm1nd: Mostly, the custom implementation is simpler (doesn't take things we don't need) and doesn't throw on Put/PutMany which would break the exchange otherwise (blockservices call put on the blockstore)

```
// EDSStore implements the blockstore interface on a DAGStore.
// The lru cache approach is heavily inspired by the open PR filecoin-project/dagstore/116.
// The main differences to the implementation here are that we do not support multiple shards per key,
// call GetSize directly on the underlying RO blockstore, and do not throw errors on Put/PutMany.
```

docs/adr/adr-11-blocksync-overhaul.md
distractedm1nd: still nothing to clean on the registry

docs/adr/adr-11-blocksync-overhaul.md
distractedm1nd: yeah that complexity would not be justified true

just hearing "This Reader is the data that is served over the wire." changed my mind on this

docs/adr/adr-11-blocksync-overhaul.md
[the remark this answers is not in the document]
distractedm1nd: Ok that is very cool and I am CARv1 pilled now

docs/adr/adr-11-blocksync-overhaul.md
Wondertan: DataRoot equals to DAHeader in this case(The DataRoot comes from [share.Root](https://github.com/celestiaorg/celestia-node/blob/main/service/share/share.go#L35) and I should clarify that). And from DAHeader we can get the height.

docs/adr/adr-11-blocksync-overhaul.md
Wondertan: > > implementation is only partial
> If we use the one from DAGStore, then it is not.
This is about being partial or not. DAGStore implements everything. The fact that we don't need Put from it seems orthogonal. We _could_ mention, though that we don't need it there, I agree

docs/adr/adr-11-blocksync-overhaul.md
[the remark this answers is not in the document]
Wondertan: > which is at least good for bitswap 

Yes, and irrelevant for the case above(Get). Get just needs to get a Reader and serve it above. There might be multiple readers accessing the same data. One of those readers might be a DAGStore shardaccessor.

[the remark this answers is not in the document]
Wondertan: Ok, so it's not locking it to one only accessor(it has a refcount), which is good, and we should acquire it then.
(I assumed that acquire locks to one accessor because of it's the word "acquire", which assumes one owner)

docs/adr/adr-11-blocksync-overhaul.md
Wondertan: True!

docs/adr/adr-11-blocksync-overhaul.md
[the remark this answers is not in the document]
Wondertan: > Thinking if you really need the reader here we should just extend the interface to return it

The whole purpose of `GetCAR` is to get a Reader 😁. Extending the interface is the way to go if it's allowed to acquire shard for multiple readers.
---
Mounts are just DAGStore native abstraction to get such a Reader. I've decided to stick to those instead of keeping os.File.

docs/adr/adr-11-blocksync-overhaul.md
Wondertan: > we definitely only have to call register once, unless we add different schemes

True

Wondertan: Resolved then

docs/adr/adr-11-blocksync-overhaul.md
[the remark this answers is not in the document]
Wondertan: No, we get this [Reader](https://github.com/filecoin-project/dagstore/blob/master/mount/mount.go#L87-L95), which has closing functionality. The user does not need to have others, so I limited it to 

It's also `io.ReadCloser`, not a `io.ReaderCloser`

docs/adr/adr-11-blocksync-overhaul.md
Wondertan: Resolved then

docs/adr/adr-11-blocksync-overhaul.md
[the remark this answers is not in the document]
Wondertan: Resolved then

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,329 @@
...
+
+To facilitate and speedup the resolution of the problem we decided to make a team gathering in Berlin for 4 days. With
+the help of preliminary preparations by @Wondertan and invited guest @willscott, we were able to find a solution
+in 2 days to match the following requirements:
+
+- Sync time less than block time(ideally sub-second)
+- Data by namespace less than block time(ideally sub-second)
+- Pragmatic timeframe
+  - We need this done before incentivized testnet
+  - So we don't have time to redesign protocol from scratch
```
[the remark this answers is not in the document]
Wondertan: But it's fine actually to remove So

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,329 @@
...
+- Keep Bitswap as it suffices DAS and solves data withholding attack
```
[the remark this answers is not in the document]
Wondertan: Applied but without attacks as its a one specific attack. Also added "the" before. I still don't fully understand the concept of articles...

docs/adr/adr-11-blocksync-overhaul.md
Wondertan: Rephrased to make to make it more correct

docs/adr/adr-11-blocksync-overhaul.md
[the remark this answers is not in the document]
Wondertan: Also clarified this double storing point on the PR

docs/adr/adr-11-blocksync-overhaul.md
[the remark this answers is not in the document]
Wondertan: but not specifically here and it does not hurt to add this, so I extended the point

docs/adr/adr-11-blocksync-overhaul.md
Wondertan: Clarified

docs/adr/adr-11-blocksync-overhaul.md
Wondertan: Additionally added a note on this

docs/adr/adr-11-blocksync-overhaul.md
Wondertan: This is super minor, and I would keep this up to the implementor. For the document, I keep this verbose so resolved

docs/adr/adr-11-blocksync-overhaul.md
Wondertan: I extended the constructor a bit, but it still not as full as in your experiment

docs/adr/adr-11-blocksync-overhaul.md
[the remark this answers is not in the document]
Wondertan: Updating this section

docs/adr/adr-11-blocksync-overhaul.md
Wondertan: Corrected

docs/adr/adr-11-blocksync-overhaul.md
Wondertan: (this is coming from Ismail always wanting me in the past to be pragmatic and implement what is absolutely required 😁)

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,329 @@
...
+Storing EDS as a whole requires EDS (de)serialization. For this the [CAR format](https://ipld.io/specs/transport/car) is
+chosen.
+
+###### `share.WriteEDS`
+
+To write EDS into a stream/file `WriteEDS` is introduced.
+// TODO Describe logic
+
+NOTE: CAR provides [a utility](https://github.com/ipld/go-car/blob/master/car.go#L47) to serialize any DAG into the file and
+there is a way to serialize EDS into DAG(`share/ipld.ImportShares`). This approach is the simplest and traverses shares
```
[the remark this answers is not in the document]
distractedm1nd: Ok excellent, so it just needs to be passed down

docs/adr/adr-11-blocksync-overhaul.md
distractedm1nd: My reasoning for this was just that if we used their implementation (at least in my experiments), it wouldn't work, because they panic on Put, which we never call explicitly but the exchange on the BlockService calls it on Add

But this is besides the point here and I think we can resolve this discussion

docs/adr/adr-11-blocksync-overhaul.md
distractedm1nd: There is still no mount to remove from the DAGStore, only the Shard. The mount is not stored in the DAGStore/Registry (but the template, which is used to create ALL mounts, is)

docs/adr/adr-11-blocksync-overhaul.md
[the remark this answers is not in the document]
distractedm1nd: Ok, typo is still here though!

docs/adr/adr-11-blocksync-overhaul.md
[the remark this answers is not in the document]
distractedm1nd: ping me when updated

docs/adr/adr-11-blocksync-overhaul.md
Wondertan: The mount is [part](https://github.com/filecoin-project/dagstore/blob/12408ab3874cd3be952e207d00627253d9bb4878/shard.go#L40) of the Shard, though. I will clarify the removal still

Wondertan: The mount is [part](https://github.com/filecoin-project/dagstore/blob/12408ab3874cd3be952e207d00627253d9bb4878/shard.go#L40) of the Shard. The mount's path also persisted. We still have to mention Mounts because while being different from Shard, we still create those and hence should remove those as well. 
I will mention that the Mount is cleared up on Shard destruction and put more emphasis on the Shard.

Wondertan: Clarified

docs/adr/adr-11-blocksync-overhaul.md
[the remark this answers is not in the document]
Wondertan: Fixed

docs/adr/adr-11-blocksync-overhaul.md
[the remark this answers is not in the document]
Wondertan: Ping

docs/adr/adr-11-blocksync-overhaul.md
Wondertan: I will double-check the necessity for the Put. Good point

docs/adr/adr-11-blocksync-overhaul.md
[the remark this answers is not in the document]
distractedm1nd: Looks great now

docs/adr/adr-11-blocksync-overhaul.md
distractedm1nd: Not sure I understand your comment here - so when do we decide?

I am strongly leaning toward a custom implementation because it is pretty trivial and can be simpler than their's (we don't need to support one cid spread over multiple shards)

docs/adr/adr-11-blocksync-overhaul.md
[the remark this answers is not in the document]
distractedm1nd: True, that naming is not that nice

docs/adr/adr-11-blocksync-overhaul.md
distractedm1nd: - It doesnt require us to use CARv2 (there is an option on it called `WriteAsCarV1`)
- It looks like it does lay down things in order, but I cannot tell if it will be fully compatible
- It isn't clear to me yet why either indexing timepoint would be better, so I trust your instincts there

But yes I think we can leave it out for now, not sure its relevant for the ADR

docs/adr/adr-11-blocksync-overhaul.md
Wondertan: Btw, check some new updates there https://github.com/filecoin-project/dagstore/pull/116


From what I can see they do support multiple shards for a CID and more shards are lazily loaded in case one of them fails, which is fine.

docs/adr/adr-11-blocksync-overhaul.md
Wondertan: It's not relevant, so resolving. WriteEDS now describes almost custom way to write the CARv1 while which `blockstore.OpenReadWrite` won't support, so resolved

docs/adr/adr-11-blocksync-overhaul.md
Wondertan: @distractedm1nd. The Put is not suitable for our Blockstore, because of the WriteEDS specifics. We could make Retriever(in the fallback case) write into the specific CAR file and its blockstore, but that breaks the quadrant ordering of writes, meaning that `ReadEDS` would read out a corrupted quadrant. Hence we should stick to the DAGStore' Blockstore, unless the caching issues you've mentioned is harmful to us

docs/adr/adr-11-blocksync-overhaul.md
Wondertan: Should be resolved

docs/adr/adr-11-blocksync-overhaul.md
Wondertan: https://github.com/celestiaorg/celestia-node/pull/1037#discussion_r971023278

distractedm1nd (approved):
Best ADR yet!

walldiss (approved):
Awesome work!

renaynay (approved):
This ADR is awesome and should be used as an example for writing ADRs org-wide :)

One nit/question: have you considered extracting EDSStore into a subpkg of `share` module? `eds.Store` ?

Wondertan:
@renaynay
> One nit/question: have you considered extracting EDSStore into a subpkg of share module? eds.Store ?

I thought that the share pkg would likely become too bloated and made a point about restructuring it in the [EPIC](https://github.com/celestiaorg/celestia-node/issues/1099). I like your idea to have a subpackage `share/eds`. Wil add it to the issue.

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,372 @@
...
+
+### Status Quo
+
+Current block data synchronization is done over Bitswap, traversing NMT trees of rows and columns of data square quadrants.
+We know from empirical evidence that it takes more than 200 seconds(~65000 network requests) to download a 4MB block of
+256kb shares, which is unacceptable and must be much less than the block time(15/30sec).
+
+The DASing, on the other hand, shows acceptable metrics for the block sizes we are aiming for initially. In the case of
+the same block, a DAS operation takes 50ms * 8(technically 9) blocking requests, which is ~400ms in an ideal scenario
+(excluding disk IO).
```
liamsi, line 36: This may sounds as if we won't be touching/improving DASing at all. Though my understandig is that why it is acceptable initially, we still will want to optimize DASing as well (I think that is sth @derrandz will look into among other things AFIU).
Nothing to change here but just double-checking that we are aligned.

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,372 @@
...
+  - We don't have time to redesign the protocol from scratch
+- Keep Bitswap, as it suffices for DAS and solves the data withholding attack
+  - Existing Bitswap logic kept as a fallback mechanism for the case of reconstruction from light nodes
+- Keeping random hash-addressed access to shares for Bitswap to work
+
+### Decision
+
+This ADR intends to outline design decisions for block data storage. In a nutshell, the decision is to use
+___[CAR format](https://ipld.io/specs/transport/car/carv2/)___ and ___[Dagstore](https://github.com/filecoin-project/dagstore)___
+for ___extended block storage___ and ___custom p2p Req/Resp protocol for block data syncing___(whole block and data by
```
liamsi, line 60: nit: whitespaces between words and `(` in general. No need to change anything but something to watch out in the future.

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,372 @@
...
+
+The new block storage design is solely additive. All the existing storage-related components and functionality
+are kept with additional components introduced. Altogether, existing and new components will be recomposed to serve as the
+foundation of our improved block storage subsystem.
+
+The central data structure representing Celestia block data is EDS(`rsmt2d.ExtendedDataSquare`), and the new storage design
+is focused around storing entire EDSes as a whole rather than a set of individual chunks, s.t. storage subsystem
+can handle storing and streaming/serving blocks of 4MB and more.
+
+#### EDS Serde
```
liamsi, line 100: What does serde stand for? Shouldn't the title here be:
#### EDS (De-)Serialization

docs/adr/adr-11-blocksync-overhaul.md
```diff
@@ -0,0 +1,372 @@
...
+  - With [`NodeVisitor`](https://github.com/celestiaorg/celestia-node/blob/main/ipld/add.go#L56), which saves to the
+`Blockstore` only NMT Merkle proofs(no shares) _NOTE: `len(node.Links()) == 2`_
+    - Actual shares are written further in a particular way explained further
+- Creates and [writes](https://github.com/ipld/go-car/blob/master/car.go#L86) header [`CARv1Header`](https://github.com/ipld/go-car/blob/master/car.go#L30)
+  - Fills up `Roots` field with `EDS.RowRoots/EDS.ColRoots` roots converted into CIDs
+- Iterates over shares in quadrant-by-quadrant order via `EDS.GetCell`
+  - [Writes](https://github.com/ipld/go-car/blob/master/car.go#L118) the shares in row-by-row order
+- Iterates over in-memory Blockstore and [writes]((https://github.com/ipld/go-car/blob/master/car.go#L118)) NMT Merkle
+proofs stored in it
+
```
liamsi, line 122: All above links should be perma-links. Otherwise they will be out of date shortly.

liamsi (commented):
Awesome work!

Issue #635, opened 2022-04-20
header: RW deadlines on stream


renaynay:
Implement read/write deadlines on libp2p streams in `HeaderEx`

Issue #497, opened 2022-03-06
[EPIC] HeaderEx protocol extensions and hardening


Wondertan:
Mainly we should remove hard dependency on trusted peers, so any node is able to sync headers from any immediate peers and not only trusted and/or bootstrappers. The trusted peers would still remain for automatic subjective initialization. Also, there are multiple networking optimization can be applied to speed up significantly speedup synchronization time.

### TODO

- [x] #635
	* Fixes potential vulnerability
- [x] #636
- [x] #906
	* ErrNotFound on the wire should be implemented, so client header ex once requested could go further and ask other peers for headers. This is needed to avoid the general request retrying logic and simply go further to another peer.
- [x] https://github.com/celestiaorg/celestia-node/issues/191
	* Currently, we only limit the requestor, but an attacker can currently request millions of headers in a range and OOM the responder
	* An error code response saying the request can't be processed as it's bigger than the limit
- [x] #1019
    - Request trusted header from all and compared, choosing the one that the majority chooses
- [x] #1020
	* Measure response latency and prefer peers with the shortest
- [ ] Fallback to peer discovery for peers serving headers in case all our immediate(connected) peers don't have the requested header(s)
	* The simplest working approach would be searching for more peers in `header-sub` HeaderSub topic(or custom header discovery topic as they are pros and cons of a unified topic that needs to be justified) with _transient connections_ and asking them for a header until we find the required one.
	* exchange.Head should only rely on trustedPeer as subjective initialization
- [x] https://github.com/celestiaorg/celestia-node/issues/197
- [x] https://github.com/celestiaorg/celestia-node/issues/366
- [x] #980
- [x] https://github.com/celestiaorg/celestia-node/issues/1140
- [x] https://github.com/celestiaorg/celestia-node/issues/1201 // should be done in scope on one PR
- [x] https://github.com/celestiaorg/celestia-node/issues/1078 // should be done in scope on one PR
- [x] https://github.com/celestiaorg/celestia-node/issues/1309
- [x] #1282
- [x] #1316
- [x] https://github.com/celestiaorg/celestia-node/issues/1406
- [x] https://github.com/celestiaorg/celestia-node/issues/1411
- [x] https://github.com/celestiaorg/celestia-node/issues/1529
- [x] https://github.com/celestiaorg/celestia-node/issues/1531

Nice-to-haves:
- [ ] #905
	* Optimization to get heads as early as possible without requested msgs
[...]

Issue #366, opened 2022-01-19
header: P2PExchange should respect context on requests.


Wondertan:
Currently, P2PExchange in `performRequest` method can be blocked for an arbitrary amount of time on `Write` and `Read` io operations. In such a case our Node may be stuck and can only be forcefully killed. This has to present along with basic 3 min WR deadline.

vgonkivs:
Will be fix by adding deadlines on read/write (#635)

Issue #197, opened 2021-11-15
service/header: Use `oneof` for `ExtendedHeaderRequest` so we can define the request by hashOrNumber


renaynay:
Use`oneof` in the proto file for `ExtendedHeaderRequest` so we can define the request by hashOrNumber instead of manually checking that one field is non-nil.

PR #1038, opened 2022-08-25
feat!(header/p2p): add deadlines on stream and rework proto

vgonkivs:
Let's the party begins 😅 .

This is a first part of changes requested in https://github.com/celestiaorg/celestia-node/issues/497. 

Current PR resolves:
* #635 
* #197
* #366 (n/a)

### DIFF
--- a/header/p2p/exchange.go
+++ b/header/p2p/exchange.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"math/rand"
+	"time"
 
 	logging "github.com/ipfs/go-log/v2"
 
@@ -22,6 +23,13 @@ import (
 )
 
 var log = logging.Logger("header/p2p")
+
+const (
+	// writeDeadline sets timeout for sending messages to the stream
+	writeDeadline = time.Second * 5
+	// readDeadline sets timeout for reading messages from the stream
+	readDeadline = time.Minute
+)
 
 // PubSubTopic hardcodes the name of the ExtendedHeader
 // gossipsub topic.
@@ -131,14 +139,24 @@ func (ex *Exchange) performRequest(
 	if err != nil {
 		return nil, err
 	}
+	if err = stream.SetWriteDeadline(time.Now().Add(writeDeadline)); err != nil {
+		log.Warn(err)
+	}
 	// send request
 	_, err = serde.Write(stream, req)
 	if err != nil {
 		stream.Reset() //nolint:errcheck
 		return nil, err
 	}
+	err = stream.CloseWrite()
+	if err != nil {
+		log.Warn(err)
+	}
 	// read responses
 	headers := make([]*header.ExtendedHeader, req.Amount)
+	if err = stream.SetReadDeadline(time.Now().Add(readDeadline * time.Duration(req.Amount))); err != nil {
+		log.Warn(err)
+	}
 	for i := 0; i < int(req.Amount); i++ {
 		resp := new(header_pb.ExtendedHeader)
 		_, err := serde.Read(stream, resp)
--- a/header/p2p/exchange.go
+++ b/header/p2p/exchange.go
@@ -140,7 +140,7 @@ func (ex *Exchange) performRequest(
 		return nil, err
 	}
 	if err = stream.SetWriteDeadline(time.Now().Add(writeDeadline)); err != nil {
-		log.Warn(err)
+		log.Warnf("error setting deadline: %s", err)
 	}
 	// send request
 	_, err = serde.Write(stream, req)
@@ -154,17 +154,19 @@ func (ex *Exchange) performRequest(
 	}
 	// read responses
 	headers := make([]*header.ExtendedHeader, req.Amount)
-	if err = stream.SetReadDeadline(time.Now().Add(readDeadline * time.Duration(req.Amount))); err != nil {
-		log.Warn(err)
-	}
 	for i := 0; i < int(req.Amount); i++ {
 		resp := new(header_pb.ExtendedHeader)
+		if err = stream.SetReadDeadline(time.Now().Add(readDeadline)); err != nil {
+			log.Warnf("error setting deadline: %s", err)
+		}
 		_, err := serde.Read(stream, resp)
 		if err != nil {
 			stream.Reset() //nolint:errcheck
 			return nil, err
 		}
-
+		if err = stream.SetReadDeadline(time.Time{}); err != nil {
+			log.Warnf("error resetting deadline: %s", err)
+		}
 		header, err := header.ProtoToExtendedHeader(resp)
 		if err != nil {
 			stream.Reset() //nolint:errcheck
--- a/header/p2p/exchange.go
+++ b/header/p2p/exchange.go
@@ -35,7 +35,7 @@ const (
 // gossipsub topic.
 const PubSubTopic = "header-sub"
 
-var exchangeProtocolID = protocol.ID(fmt.Sprintf("/header-ex/v0.0.2/%s", params.DefaultNetwork()))
+var exchangeProtocolID = protocol.ID(fmt.Sprintf("/header-ex/v0.0.3/%s", params.DefaultNetwork()))
 
 // Exchange enables sending outbound ExtendedHeaderRequests to the network as well as
 // handling inbound ExtendedHeaderRequests from the network.
--- a/header/p2p/exchange.go
+++ b/header/p2p/exchange.go
@@ -150,7 +150,7 @@ func (ex *Exchange) performRequest(
 	}
 	err = stream.CloseWrite()
 	if err != nil {
-		log.Warn(err)
+		log.Error(err)
 	}
 	// read responses
 	headers := make([]*header.ExtendedHeader, req.Amount)
--- a/header/p2p/exchange.go
+++ b/header/p2p/exchange.go
@@ -140,7 +140,7 @@ func (ex *Exchange) performRequest(
 		return nil, err
 	}
 	if err = stream.SetWriteDeadline(time.Now().Add(writeDeadline)); err != nil {
-		log.Warnf("error setting deadline: %s", err)
+		log.Debugf("error setting deadline: %s", err)
 	}
 	// send request
 	_, err = serde.Write(stream, req)
@@ -157,7 +157,7 @@ func (ex *Exchange) performRequest(
 	for i := 0; i < int(req.Amount); i++ {
 		resp := new(header_pb.ExtendedHeader)
 		if err = stream.SetReadDeadline(time.Now().Add(readDeadline)); err != nil {
-			log.Warnf("error setting deadline: %s", err)
+			log.Debugf("error setting deadline: %s", err)
 		}
 		_, err := serde.Read(stream, resp)
 		if err != nil {
@@ -165,7 +165,7 @@ func (ex *Exchange) performRequest(
 			return nil, err
 		}
 		if err = stream.SetReadDeadline(time.Time{}); err != nil {
-			log.Warnf("error resetting deadline: %s", err)
+			log.Debugf("error resetting deadline: %s", err)
 		}
 		header, err := header.ProtoToExtendedHeader(resp)
 		if err != nil {
--- a/header/p2p/exchange.go
+++ b/header/p2p/exchange.go
@@ -50,7 +50,7 @@ func (ex *Exchange) Head(ctx context.Context) (*header.ExtendedHeader, error) {
 	log.Debug("requesting head")
 	// create request
 	req := &p2p_pb.ExtendedHeaderRequest{
-		Origin: uint64(0),
+		Data:   &p2p_pb.ExtendedHeaderRequest_Origin{Origin: uint64(0)},
 		Amount: 1,
 	}
 	headers, err := ex.performRequest(ctx, req)
@@ -71,7 +71,7 @@ func (ex *Exchange) GetByHeight(ctx context.Context, height uint64) (*header.Ext
 	}
 	// create request
 	req := &p2p_pb.ExtendedHeaderRequest{
-		Origin: height,
+		Data:   &p2p_pb.ExtendedHeaderRequest_Origin{Origin: height},
 		Amount: 1,
 	}
 	headers, err := ex.performRequest(ctx, req)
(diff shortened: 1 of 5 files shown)


header/p2p/server.go
```diff
@@ -52,19 +54,27 @@ func (serv *ExchangeServer) Stop(context.Context) error {
 
 // requestHandler handles inbound ExtendedHeaderRequests.
 func (serv *ExchangeServer) requestHandler(stream network.Stream) {
+	err := stream.SetReadDeadline(time.Now().Add(readDeadline))
```
renaynay, line 57: This only set's a read deadline on the host's end right?

header/p2p/server.go
```diff
@@ -52,19 +54,27 @@ func (serv *ExchangeServer) Stop(context.Context) error {
...
 	}
+	if err = stream.CloseRead(); err != nil {
+		log.Warn(err)
+	}
 	// retrieve and write ExtendedHeaders
-	if pbreq.Hash != nil {
-		serv.handleRequestByHash(pbreq.Hash, stream)
-	} else {
-		serv.handleRequest(pbreq.Origin, pbreq.Origin+pbreq.Amount, stream)
+	switch pbreq.Data.(type) {
```
renaynay, line 73: default case should log out error and return

header/p2p/server.go
```diff
@@ -124,6 +137,10 @@ func (serv *ExchangeServer) handleRequest(from, to uint64, stream network.Stream
 		}
 		headers = headersByRange
 	}
+
+	if err := stream.SetWriteDeadline(time.Now().Add(writeDeadline * time.Duration(len(headers)))); err != nil {
```
renaynay, line 141: Does the write deadline persist per-write? If not, we should just set it before writing, but we shouldn't make the write deadline so long.
```suggestion
	if err := stream.SetWriteDeadline(time.Now().Add(writeDeadline)); err != nil {
```

vgonkivs: No, it can't be set per-write. But the case that you suggested assumes that timeouts for writing 1 block or 200 blocks are equal. Is it correct?

header/p2p/server.go
vgonkivs: or panic 😄  as we are expecting only these two cases

header/p2p/server.go
vgonkivs: yes

header/p2p/server.go
renaynay: The thing is, you are performing 1 write per header that is being written bc it happens in a loop. The write deadline shouldn't be too long on that. So if you set the write deadline on the given stream, does it persist on the stream for that host or does it get "cleared" after one write has been made?

header/p2p/server.go
renaynay: No, not panic. It's possible there will be garbage requests from the network, we should definitely not panic in that instance as that's a DoS vector (similar to the panic in the proof unmarshaller thing in fraudsync 😅 )

header/p2p/server.go
renaynay: ack

header/p2p/server.go
Wondertan: It not cleared. To clear it and empty time should be passed to it. 

@vgonkivs in fraudsync we didn't care about resetting the deadline because we closed it straight after the one only write. If we reuse the stream for multiple writes, then deadlines must be reset for both reads and writes

vgonkivs: ok

header/p2p/server.go
vgonkivs: ok

Wondertan:
Note that this is ___protocol breaking___

Wondertan (changes requested):
Note that this is ___protocol breaking___, because of PB changes breaking backward compatibility. Thus the protocol ID/version has to be changed

header/p2p/server.go
```diff
@@ -52,19 +54,29 @@ func (serv *ExchangeServer) Stop(context.Context) error {
...
-		serv.handleRequestByHash(pbreq.Hash, stream)
-	} else {
-		serv.handleRequest(pbreq.Origin, pbreq.Origin+pbreq.Amount, stream)
+	switch pbreq.Data.(type) {
+	case *p2p_pb.ExtendedHeaderRequest_Hash:
+		serv.handleRequestByHash(pbreq.GetHash(), stream)
+	case *p2p_pb.ExtendedHeaderRequest_Origin:
+		serv.handleRequest(pbreq.GetOrigin(), pbreq.GetOrigin()+pbreq.Amount, stream)
+	default:
+		log.Error("server: invalid data type received")
```
Wondertan, line 79: Reset stream and return

Wondertan (commented):
We also should not merge this in `main` before the upcoming release. Let's keep multiple breaking changes for HeaderEx protocol together and not spread those over multiple releases to avoid joint headache

header/p2p/exchange.go
```diff
@@ -130,22 +139,34 @@ func (ex *Exchange) performRequest(
 	if err != nil {
 		return nil, err
 	}
+	if err = stream.SetWriteDeadline(time.Now().Add(writeDeadline)); err != nil {
+		log.Warnf("error setting deadline: %s", err)
```
renaynay, line 143: I would demote all warn logs to debug level here since we don't actually care *too* much about this

header/p2p/exchange.go
```diff
@@ -130,22 +139,34 @@ func (ex *Exchange) performRequest(
...
+		if err = stream.SetReadDeadline(time.Now().Add(readDeadline)); err != nil {
+			log.Warnf("error setting deadline: %s", err)
+		}
 		_, err := serde.Read(stream, resp)
 		if err != nil {
 			stream.Reset() //nolint:errcheck
 			return nil, err
 		}
-
+		if err = stream.SetReadDeadline(time.Time{}); err != nil {
```
renaynay, line 167: Do we need this?

renaynay: Shouldn't `SetReadDeadline` override the previous read deadline?

header/p2p/server.go
```diff
@@ -139,5 +160,9 @@ func (serv *ExchangeServer) handleRequest(from, to uint64, stream network.Stream
 			stream.Reset() //nolint:errcheck
 			return
 		}
+
+		if err = stream.SetWriteDeadline(time.Time{}); err != nil {
```
renaynay, line 164: Same here, whether we actually need this.

renaynay (commented):
Is the underlying impl of stream here yamux? If it so, then 
```
// SetReadDeadline sets the deadline for future Read calls.
```

header/p2p/exchange.go
vgonkivs: I think no. Did not find such mechanism: 
here we set a deadline https://github.com/libp2p/go-yamux/blob/781eed82a74eb7b3868c390fce8091c3c40c708d/deadline.go#L31 , that will close done channel at #L55 when timeout is reached. So, I assumed that we will get here https://github.com/libp2p/go-yamux/blob/781eed82a74eb7b3868c390fce8091c3c40c708d/stream.go#L115 every time once deadline will be reached

header/p2p/exchange.go
vgonkivs: ok

renaynay:
> Is the underlying impl of stream here yamux? If it so, then
> 
> ```
> // SetReadDeadline sets the deadline for future Read calls.
> ```

header/p2p/exchange.go
vgonkivs: done.

header/p2p/server.go
vgonkivs: the same as above. libp2p does not reset timer per write. So, we have to do it manually

header/p2p/exchange.go
Wondertan: Note that libp2p folks use Warn in this case.

header/p2p/server.go
Wondertan: Yes, this resetting is required

header/p2p/exchange.go
renaynay: Ack, I see that's how libp2p does it too.

header/p2p/server.go
renaynay: ack.

renaynay (approved):

header/p2p/exchange.go
renaynay: you can resolve this @vgonkivs

header/p2p/server.go
```diff
@@ -52,19 +54,31 @@ func (serv *ExchangeServer) Stop(context.Context) error {
...
-		serv.handleRequestByHash(pbreq.Hash, stream)
-	} else {
-		serv.handleRequest(pbreq.Origin, pbreq.Origin+pbreq.Amount, stream)
+	switch pbreq.Data.(type) {
+	case *p2p_pb.ExtendedHeaderRequest_Hash:
+		serv.handleRequestByHash(pbreq.GetHash(), stream)
+	case *p2p_pb.ExtendedHeaderRequest_Origin:
+		serv.handleRequest(pbreq.GetOrigin(), pbreq.GetOrigin()+pbreq.Amount, stream)
+	default:
+		log.Error("server: invalid data type received")
```
renaynay, line 79: We should eventually have some sort of blocking behaviour for peers that send us multiple bad requests. cc @Wondertan 

This should be p2p_exchange - wide IMO where we track peer IDs who send us bad requests server-side and blocklist them if they hit 2+ bad reqs. WDYT? We can stick this in the epic as an issue. I thought we had some notes about that in the `Head` sanity checking issue but we should have it on the server side for any inbound request that is bad.

Wondertan: There should be some single peer score tracking component that each protocol we have can report to. Sending bad data is something every protocol faces. What we could do for now, is just to blacklist a peer completely.

Wondertan (dismissed):
The commit and the PR name should have an exclamation mark pointing to this PR as breaking. Other then that, lGTM

header/p2p/exchange.go
walldiss: Sorry for writing in closed thread, but I had the exact same question. I've digged into libp2p and found it is quite interesting piece of code. There are several places where this situation is handled very smart.

1. If we set new deadline before previous one was reached [#L55](https://github.com/libp2p/go-yamux/blob/781eed82a74eb7b3868c390fce8091c3c40c708d/deadline.go#L55) could be handled by line [#L35](https://github.com/libp2p/go-yamux/blob/781eed82a74eb7b3868c390fce8091c3c40c708d/deadline.go#L35). `!d.timer.Stop()` call will stop `time.AfterFunc` timer and will prevent it from executing callback. So new deadline will override previous one.

2. Another case is if deadline has happen after `Read()` was called , meaning `time.AfterFunc` has fired and closed the `d.closed` channel. If we try to set new deadline by `stream.SetReadDeadline()` the `d.closed` needs to be recreated. Which will happen in line  [#L52](https://github.com/libp2p/go-yamux/blob/781eed82a74eb7b3868c390fce8091c3c40c708d/deadline.go#L52) 

3. The more interesting part is the fact that potentially `time.AfterFunc` could fired, but didn't finished closing `d.closed` channel yet (imagine there is `time.Sleep` inside callback before `close(d.close)`) . In this case `d.timer.Stop()` wouldn't stop internal timer because it is already stoped and will return `false`. So we wait `d.close` to be closed  inside `time.AfterFunc` callback in [#L36](https://github.com/libp2p/go-yamux/blob/781eed82a74eb7b3868c390fce8091c3c40c708d/deadline.go#L36). Then proceed to [#L52](https://github.com/libp2p/go-yamux/blob/781eed82a74eb7b3868c390fce8091c3c40c708d/deadline.go#L52)  path.

So seems like the line could be removed but won't hurt having it either.

vgonkivs:
```
2022-09-19T10:53:02.2788980Z     reconstruct_test.go:180: 
2022-09-19T10:53:02.2789936Z         	Error Trace:	/home/runner/work/celestia-node/celestia-node/node/tests/reconstruct_test.go:180
2022-09-19T10:53:02.2790496Z         	Error:      	Received unexpected error:
2022-09-19T10:53:02.2791066Z         	            	context deadline exceeded
2022-09-19T10:53:02.2791697Z         	Test:       	TestFullReconstructFromLights
```

Wondertan:
That's the one that fails on main as well

Wondertan (approved):

header/p2p/exchange.go
```diff
@@ -130,22 +139,31 @@ func (ex *Exchange) performRequest(
...
+	}
 	// send request
 	_, err = serde.Write(stream, req)
 	if err != nil {
 		stream.Reset() //nolint:errcheck
 		return nil, err
 	}
+	err = stream.CloseWrite()
+	if err != nil {
+		log.Warn(err)
```
renaynay, line 153: log.Err please

header/p2p/server.go
```diff
@@ -52,19 +54,31 @@ func (serv *ExchangeServer) Stop(context.Context) error {
...
 	pbreq := new(p2p_pb.ExtendedHeaderRequest)
-	_, err := serde.Read(stream, pbreq)
+	_, err = serde.Read(stream, pbreq)
 	if err != nil {
 		log.Errorw("server: reading header request from stream", "err", err)
 		stream.Reset() //nolint:errcheck
 		return
 	}
+	if err = stream.CloseRead(); err != nil {
+		log.Warn(err)
```
renaynay, line 70: log.Err please

header/p2p/server.go
renaynay: Yeah, let's flesh this out post-mainnet. Maybe create a stub issue for this in the moment? Doesn't have to be fleshed out fully.

renaynay (dismissed):
Please hold off on merging this until we batch breaking changes.

Wondertan (dismissed):

walldiss (dismissed):

Wondertan (dismissed):
If we started to break things in main already via #997. We can start merging this

renaynay (approved):

renaynay (changes requested):
Please increment exchange protocol version

renaynay (approved):

distractedm1nd (dismissed):

walldiss (approved):

Wondertan (approved):

renaynay (approved):
All tests passing besides integration (which is expected).

Wondertan (approved):

Issue #1040, opened 2022-08-25
[Feature Request]: metrics for PayForData and last sample requested


rootulp:
| Metric Name                              | Light node | Full Node | Bridge node |
| ---------------------------------------- | ---------- | ----------------- | ----------- |
| Total PayForData messages sent (count)   | x          | x                 | x           |
| Last PayForData message sent (timestamp) | x          | x                 | x           |
| Last sample requested (timestamp)        | x          | x                 |             |
| Last block served (timestamp)            |            |                   | x           |
| Last full sampling request (timestamp)   |            | x                 |             |

Feature request to emit the above metrics with a label that uniquely identifies a node (i.e. peer id). Note that a peer id is bounded on incentivized testnet but unbounded on mainnet so we likely want to remove the label prior to mainnet to avoid issues with high cardinality (see [Prometheus docs](https://prometheus.io/docs/practices/naming/#labels)).

rootulp:
@YazzyYaz and @Wondertan brainstormed additional metrics that may be necessary so I updated the table above

distractedm1nd:
What is meant by "Last full sampling request"? Since bridge nodes don't sample

I'm assuming on every reconstruction or something? Since bridge+full both have a `FullAvailability`

rootulp:
> What is meant by "Last full sampling request"? Since bridge nodes don't sample

Good question! Maybe it pertains to which data root was last requested / retrieved?https://github.com/celestiaorg/celestia-node/blob/9e45ada26417fc4895e33326584b896754cff1c6/service/share/full_availability.go#L58

If so, we can likely update the name to not include `sample` because that seems confusing. @Wondertan likely has input here

distractedm1nd:
Yes that seems the most likely as well, will add there

Wondertan:
I'm not sure we need to track the last requested(sampled) block by Bridge, and I don't remember discussing this, to be honest.

We also need to track the last ___served___ Block from Full nodes besides only Bridges(currently the table sell for Fulls is empty). But we cannot do this right now, as there is no way to differentiate between whole block requests and sampling requests. We can do, though, wrap IPFS blockstore with some metering that sends metrics on each sample/share read, but that's way too granular and actually hits high cardinality issue. Still, this approach does not answer whether this is a block request or a sampling. cc @distractedm1nd 

Another point is metering block serving/requesting right now can be too early considering the new blocksync work. Doing it now will be required redoing it later.

[...]

PR #1047, opened 2022-08-30
metrics: add metrics to fraud package

vgonkivs:
Added metrics and traces

```
Sep 29 11:20:10 rootulp-otel-collector otelcol[110017]:      -> Name: badencoding
Sep 29 11:20:10 rootulp-otel-collector otelcol[110017]:      -> Description: Stored fraud proof
Sep 29 11:20:10 rootulp-otel-collector otelcol[110017]:      -> Unit: 1
Sep 29 11:20:10 rootulp-otel-collector otelcol[110017]:      -> DataType: Sum
Sep 29 11:20:10 rootulp-otel-collector otelcol[110017]:      -> IsMonotonic: true
Sep 29 11:20:10 rootulp-otel-collector otelcol[110017]:      -> AggregationTemporality: AGGREGATION_TEMPORALITY_>
Sep 29 11:20:10 rootulp-otel-collector otelcol[110017]: NumberDataPoints #0
Sep 29 11:20:10 rootulp-otel-collector otelcol[110017]: StartTimestamp: 2022-09-29 11:14:56.06662 +0000 UTC
Sep 29 11:20:10 rootulp-otel-collector otelcol[110017]: Timestamp: 2022-09-29 11:20:10.073583 +0000 UTC
Sep 29 11:20:10 rootulp-otel-collector otelcol[110017]: Value: 0
```

Resolves https://github.com/celestiaorg/celestia-node/issues/999


fraud/service.go
```diff
@@ -129,6 +146,11 @@ func (f *ProofService) processIncoming(
 	from peer.ID,
 	msg *pubsub.Message,
 ) pubsub.ValidationResult {
+	ctx, span := tracer.Start(ctx, "fraud proof validating", trace.WithAttributes(
```
Wondertan, line 149:
```suggestion
	ctx, span := tracer.Start(ctx, "process-proof", trace.WithAttributes(
```

fraud/service.go
```diff
@@ -160,15 +183,25 @@ func (f *ProofService) processIncoming(
 		log.Errorw("proof validation err: ",
 			"err", err, "proofType", proof.Type(), "height", proof.Height())
 		f.pubsub.BlacklistPeer(from)
+		span.RecordError(err)
 		return pubsub.ValidationReject
 	}
 
+	span.AddEvent("received-fraud-proof", trace.WithAttributes(
```
Wondertan, line 190:
```suggestion
	span.AddEvent("received-valid-proof", trace.WithAttributes(
```

fraud/service.go
```diff
@@ -103,8 +116,12 @@ func (f *ProofService) Subscribe(proofType ProofType) (_ Subscription, err error
 	}
 	subs, err := t.Subscribe()
 	if err != nil {
+		span.RecordError(err)
+		span.SetStatus(codes.Error, "error during subscribing to a pubsub")
 		return nil, err
 	}
+	go observeFraudProofs(proofType, f.ds)
+	span.SetStatus(codes.Ok, "subscribed to fraud proof")
```
Wondertan, line 124: We don't need tracing for this func

fraud/service.go
```diff
@@ -103,8 +116,12 @@ func (f *ProofService) Subscribe(proofType ProofType) (_ Subscription, err error
 	}
 	subs, err := t.Subscribe()
 	if err != nil {
+		span.RecordError(err)
+		span.SetStatus(codes.Error, "error during subscribing to a pubsub")
 		return nil, err
 	}
+	go observeFraudProofs(proofType, f.ds)
```
Wondertan, line 123: This will make every subscription run observer for the same fraud type...

Also, let's make it similar to the header pkg, where it has to be:
* A public func
* Working over public API of fraud service
	* Extend API if necessary
* Optionable on the node

fraud/service.go
```diff
@@ -160,15 +183,25 @@ func (f *ProofService) processIncoming(
...
+	))
+
 	// add the fraud proof to storage.
 	err = f.put(ctx, proof.Type(), hex.EncodeToString(proof.HeaderHash()), msg.Data)
 	if err != nil {
 		log.Errorw("failed to store fraud proof", "err", err)
+		span.RecordError(err)
 	}
-
+	log.Warn("Shutting down services...")
```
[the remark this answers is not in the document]
vgonkivs: appeared during resolving merge conflicts...

fraud/service.go
```diff
@@ -160,15 +183,25 @@ func (f *ProofService) processIncoming(
...
+	span.SetStatus(codes.Ok, "fraud proof verified")
```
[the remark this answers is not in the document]
vgonkivs: yeap. thanks

fraud/proof.go
```diff
@@ -80,3 +93,40 @@ func Unmarshal(proofType ProofType, msg []byte) (Proof, error) {
 	}
 	return unmarshaler(msg)
 }
+
+// MonitorProofs enables metrics to monitor fraud proofs.
+func MonitorProofs(proofType ProofType) func(datastore.Datastore) {
```
Wondertan, line 98: Let's design this around pkg as well. Similarly how we now join topics for any registered proof type on the pkg. So that any registered proof is automatically reported.

fraud/proof.go
```diff
@@ -80,3 +93,40 @@ func Unmarshal(proofType ProofType, msg []byte) (Proof, error) {
...
+		counter, _ := meter.AsyncInt64().Counter(string(proofType),
+			instrument.WithUnit(unit.Dimensionless),
+			instrument.WithDescription("Stored fraud proof"),
+		)
+		err := meter.RegisterCallback(
+			[]instrument.Asynchronous{
+				counter,
+			},
+			func(ctx context.Context) {
+				proofs, err := getAll(ctx, store, proofType)
```
Wondertan, line 110: Can we make it over Service API am not over private methods?

fraud/proof.go
vgonkivs: Why? We join topics for any register proofs as it requires pubsub instance that is a part of our ProofService and also we do not want to shift responsibility of creating a topic to a user as it's a core part of our subscription. From the other side, metrics is `nice to have` but not required. In terms of our node, why do we need to observe some other fraud proofs?

fraud/proof.go
distractedm1nd: Sounds like it can actually be combined with #1040 then, no? Since that will add metering to Service

distractedm1nd: Or do you mean the getAll specifically?

fraud/registry.go
```diff
@@ -28,8 +28,8 @@ func Register(p Proof) {
 	}
 }
 
-// getRegisteredProofTypes returns all available proofTypes.
-func getRegisteredProofTypes() []ProofType {
+// GetRegisteredProofTypes returns all available proofTypes.
+func GetRegisteredProofTypes() []ProofType {
```
Wondertan, line 32: Let's shorten it to `RegisteredProofs`

Wondertan (commented):
Nice

nodebuilder/fraud/opts.go
```diff
@@ -0,0 +1,15 @@
...
+
+	"github.com/celestiaorg/celestia-node/fraud"
+)
+
+// WithMetrics enables metrics exporting for the node.
+func WithMetrics(enable bool) fx.Option {
+	if !enable {
+		return fx.Options()
+	}
+	return fx.Options(fx.Invoke(fraud.MonitorProofs(fraud.RegisteredProofTypes()...)))
```
Wondertan, line 14: nit: You can remove fx.Options wrapping.

Wondertan (approved):

nodebuilder/fraud/module.go
```diff
@@ -15,11 +16,17 @@ func Module(tp node.Type) fx.Option {
 		return fx.Module(
 			"fraud",
 			fx.Provide(ServiceWithSyncer),
+			fx.Provide(func(service fraud.Service) fraud.Getter {
```
renaynay, line 19: make `baseComponents` for this since it's on all node types?

fraud/sync.go
```diff
@@ -57,14 +59,24 @@ func (f *ProofService) syncFraudProofs(ctx context.Context) {
 		peerCache[connStatus.Peer] = struct{}{}
 		// valid peer found, so go send proof requests
 		go func(pid peer.ID) {
+			ctx, span := tracer.Start(ctx, "sync-proofs")
+			defer span.End()
+
+			span.SetAttributes(
+				attribute.String("pid", pid.String()),
```
renaynay, line 66:
```suggestion
				attribute.String("peerID", pid.String()),
```

renaynay: pid usually means process identifier, we should be more explicit

renaynay (commented):
nits

distractedm1nd (approved):

nodebuilder/fraud/module.go
vgonkivs: done

nodebuilder/fraud/opts.go
vgonkivs: done

distractedm1nd (dismissed):

Wondertan (approved):

fraud/proof.go
```diff
@@ -80,3 +93,39 @@ func Unmarshal(proofType ProofType, msg []byte) (Proof, error) {
...
+						counter.Observe(ctx,
+							int64(len(proofs)),
+							attribute.String("proofType", string(proofType)),
+						)
+					case datastore.ErrNotFound:
+						counter.Observe(ctx, 0)
+						return
+					default:
+						counter.Observe(ctx, 0, attribute.String("err", err.Error()))
+					}
```
walldiss, line 122: I think Counter observes increasing increment, so there should be no reason to observe 0 increment since it won't affect the metric.

Also setting arbitrary error as vector value could lead to problems in metric storage indexes. I would suggest mapping error into error codes or simply storing as bool(err != nil). This way amount of possible values in vector is limited.

walldiss (dismissed):

Wondertan (dismissed):

Wondertan (dismissed):

fraud/proof.go
```diff
@@ -80,3 +93,39 @@ func Unmarshal(proofType ProofType, msg []byte) (Proof, error) {
...
+					case nil:
+						counter.Observe(ctx,
+							int64(len(proofs)),
+							attribute.String("proof_type", string(proofType)),
+						)
+					case datastore.ErrNotFound:
+						counter.Observe(ctx, 1, attribute.String("err", "not_found"))
+						return
+					default:
+						counter.Observe(ctx, 1, attribute.String("err", "unknown"))
```
renaynay, line 121: This is a bit weird to me but it's fine I guess.

renaynay (approved):

Wondertan:
Oops, conflict

renaynay (approved):

fraud/proof.go
```diff
@@ -80,3 +93,39 @@ func Unmarshal(proofType ProofType, msg []byte) (Proof, error) {
...
 }
+
+// MonitorProofs enables metrics to monitor fraud proofs.
+func MonitorProofs(proofTypes ...ProofType) func(Getter) {
+	return func(store Getter) {
+		for _, proofType := range proofTypes {
+			counter, _ := meter.AsyncInt64().Counter(string(proofType),
+				instrument.WithUnit(unit.Dimensionless),
+				instrument.WithDescription("Stored fraud proof"),
+			)
```
walldiss, line 104: Counter has to be always incremented to record values, if we use Gauge here it will be possible to update metric without increasing its value. It will be easier with gauge to observe cases when no proofs or an error returned from store.

fraud/proof.go
walldiss, line 121: If Gauge is used, we can set it to 0 with error vectorization.

fraud/sync.go
```diff
@@ -57,14 +59,24 @@ func (f *ProofService) syncFraudProofs(ctx context.Context) {
 		peerCache[connStatus.Peer] = struct{}{}
 		// valid peer found, so go send proof requests
 		go func(pid peer.ID) {
+			ctx, span := tracer.Start(ctx, "sync_proofs")
+			defer span.End()
+
+			span.SetAttributes(
+				attribute.String("peer-id", pid.String()),
```
walldiss, line 66:
```suggestion
				attribute.String("peer_id", pid.String()),
```

walldiss (commented):
Added a small suggestion how to make metric a bit easier to observe.

walldiss (approved):

Wondertan (approved):
Thanks for helping with polishing this @walldiss!

nodebuilder/fraud/opts.go
```diff
@@ -0,0 +1,15 @@
...
+	"go.uber.org/fx"
+
+	"github.com/celestiaorg/celestia-node/fraud"
+)
+
+// WithMetrics enables metrics exporting for the node.
+func WithMetrics(enable bool) fx.Option {
+	if !enable {
+		return fx.Options()
+	}
```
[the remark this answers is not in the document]
Wondertan: This pattern is common and allows users to avoid the if. An example would passing value of the bool flag directly into the option. In our usage case though it's not that important so I we can remove it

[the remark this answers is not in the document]
vgonkivs: removed

Wondertan (approved):

renaynay (approved):

PR #1051, opened 2022-09-01
feat(blocksync): WIP experiments with DAGStore as a Blockstore

distractedm1nd:
ahhhh we're CARinggggg

This PR in its current state is purely just for experimentation, and is hack upon hack upon hack just to see what obstacles we may run into during implementation of the new blocksync. I expect to throw this all away once the ADR #1037 is finished (this PR could also possibly influence the ADR, though), and take what we learned from this into it.

If you try to run this pls mkdir `/tmp/carexample/` and `/tmp/transients/`

Current features:
- DAGStore that implements a BlockStore with a persisted top level index and LRU cache
- CARBlockStore that acts as a NodeAdder and handles the CAR-level blockstores
- Every block reconstructed on a full node creates a CARv2
- Each contained CID is registered on the top level index for retrieval from the blockstore

Biggest issues:
- [X] Uhhhh im pretty sure light nodes and bridge nodes wouldnt work rn if you tried to run them (it uses a DAGBlockstore for everything) (Artifacts from playing around are everywhere, DI is plain wrong in places)
- [X] I don't think getting over the top level index is currently functional (I think it fails and goes straight to the exchange)
- [x] There is no cache on the DAGBlockStore implementation
- [ ] The persisted datastore for the top level index does NOT use go-indexer-core as a backend
- [ ] The naming of the files uses the DAH header hash (which is not unique) instead of the extended header hash
- [X] I commented out testing.go
- [ ] Tests are not passing
- [ ] Config/Management over file directory (will wait for ADR on this one)
- [ ] Bridge nodes should also use the blockstore, so: make AddShares compatible with it (difficult, because we need root up front)


distractedm1nd:
Closing in favor of #1232

PR #1052, opened 2022-09-05
params: Switch default network to `arabica`

renaynay:
In preparation for new devnet launch.

### DIFF
--- a/params/bootstrap.go
+++ b/params/bootstrap.go
@@ -29,6 +29,7 @@ func bootstrappersFor(net Network) ([]string, error) {
 
 // NOTE: Every time we add a new long-running network, its bootstrap peers have to be added here.
 var bootstrapList = map[Network][]string{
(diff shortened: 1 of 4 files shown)


params/bootstrap.go
```diff
@@ -29,6 +29,7 @@ func bootstrappersFor(net Network) ([]string, error) {
 
 // NOTE: Every time we add a new long-running network, its bootstrap peers have to be added here.
 var bootstrapList = map[Network][]string{
+	Arabica: {}, // @TODO fill out once boostrappers are live.
```
Bidon15, line 32: Now bootstrappers are live, please add them here as ip4 🙏

renaynay: dns?

Wondertan (approved):
Thank you!

vgonkivs (approved):

PR #1056, opened 2022-09-07
refactor(nodebuilder): moving service/<service> services to respective node sub-packages

### LINKED ISSUE #960: Migrate `service/<service>` services to respective node sub-packages
E.g. service/share --> node/share only the high level accessor methods and all the substance (implementation of those methods) get migrated to share pkg at root level

distractedm1nd:
Moves services into `nodebuilder` and makes the internal structs private, offering a public interface instead.

Closes #960 

Packages:
- [x] ~~RPC~~
- [x] Share
- [X] Header
- [X] State

I took the branch from @renaynay's remote and pushed it to our remote here so that the diffs will be clear. #997 can then rebase off of that branch when things get merged into it. 

Leaving RPC out of things since it will be ripped out anyways. Should we at least move it to the root though?


ipld/retriever.go
```diff
@@ -44,7 +44,7 @@ type Retriever struct {
 	bServ blockservice.BlockService
 }
 
-// NewRetriever creates a new instance of the Retriever over IPLD Service and rmst2d.Codec
+// NewRetriever creates a new instance of the Retriever over IPLD service and rmst2d.Codec
```
Wondertan, line 47:
```suggestion
// NewRetriever creates a new instance of the Retriever over IPLD BlockService and rmst2d.Codec
```

header/p2p/subscriber.go
```diff
@@ -12,14 +12,14 @@ import (
 	"github.com/celestiaorg/celestia-node/header"
 )
 
-// Subscriber manages the lifecycle and relationship of header Service
```
Wondertan, line 15: There are bunch of decapitalization in changes and as thosenl international or relic from some global renaming?

das/daser_test.go
```diff
@@ -17,7 +17,8 @@ import (
 
 	"github.com/celestiaorg/celestia-node/fraud"
 	"github.com/celestiaorg/celestia-node/header"
-	"github.com/celestiaorg/celestia-node/service/share"
+	"github.com/celestiaorg/celestia-node/nodebuilder/share"
+	service "github.com/celestiaorg/celestia-node/share"
```
Wondertan, line 21: This should be called share not service.

das/daser_test.go
```diff
@@ -17,7 +17,8 @@ import (
 
 	"github.com/celestiaorg/celestia-node/fraud"
 	"github.com/celestiaorg/celestia-node/header"
-	"github.com/celestiaorg/celestia-node/service/share"
+	"github.com/celestiaorg/celestia-node/nodebuilder/share"
```
Wondertan, line 20: Daser should not import bodybuilder.

nodebuilder/header/service.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/celestiaorg/celestia-node/header"
+	"github.com/celestiaorg/celestia-node/header/p2p"
+	"github.com/celestiaorg/celestia-node/header/sync"
+)
+
+type Service interface {
+	// Start starts the header service.
+	Start(context.Context) error
+	// Stop stops the header service.
+	Stop(context.Context) error
```
Wondertan, line 15: Do we need to keep those Start and Stop in each service by the way? I would remove those and make a separate API or Service in the node module that can granularly start and stop a service/module by it's name.

Wondertan: Right now implementations are just empty functions. They were there before, but it's time to remove those.

das/daser_test.go
distractedm1nd: What? Why? That is where the service definition is that we need.

header/p2p/subscriber.go
distractedm1nd: Yes, must have been from refactoring in GoLand

das/daser_test.go
distractedm1nd: Ah, will try to cast to share.Availability then

nodebuilder/header/service.go
distractedm1nd: done

das/daser_test.go
distractedm1nd: done

header/p2p/subscriber.go
distractedm1nd: Replaced now where it makes sense, there are still some usages of lowercase service inside the package that are correct

ipld/retriever.go
distractedm1nd: done

header/p2p/subscriber.go
Wondertan: 👌🏻

das/daser_test.go
Wondertan: Yes. Thanks

nodebuilder/share/service.go
```diff
@@ -0,0 +1,35 @@
...
+//     * Return
+//  3. If not
+//     * Find provider on the network
+//     * Fetch the Share from the provider
+//     * Store the Share
+//     * Return
+type Service interface {
+	share.Availability
+	Start(ctx context.Context) error
+	Stop(ctx context.Context) error
```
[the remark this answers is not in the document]
distractedm1nd: Oops, do you mean in all services? I thought you only meant in header service, my bad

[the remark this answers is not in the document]
distractedm1nd: Sorry, but rereading you original comment:
> Do we need to keep those Start and Stop in each service by the way? I would remove those and make a separate API or Service in the node module that can granularly start and stop a service/module by it's name.

How would we be able to start and stop the services if start and stop are not in the interface, and the node works over the interface? The share modules start and stop are important, so I am leaving them there until this is answered. What would justify the added complexity of abstracting away Start and Stop, and how exactly would that work?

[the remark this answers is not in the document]
distractedm1nd: The Node works over the interfaces, of course - where do you want to give it a handle to the raw, unexported service?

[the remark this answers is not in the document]
Wondertan: > where do you want to give it a handle to the raw, unexported service?

During construction. You can provide Services raw and as interface implementation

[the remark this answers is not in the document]
distractedm1nd: Then I don't understand the point of having the interface. If it was only for RPC usage, this is no longer relevant. Should I close this PR then?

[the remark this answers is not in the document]
Wondertan: It's not only for RPC usage. These services/APIs will be used by directly importing the node and its services/APIs.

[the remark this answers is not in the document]
distractedm1nd: Ok - so why should we restrict users from starting and stopping the services at their will?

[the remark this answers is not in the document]
Wondertan: Each module provides some functionality and provides API that lives in the module, so this PR that does such moving is what we want independently of RPC/REST question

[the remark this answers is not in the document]
Wondertan: Yes

nodebuilder/fraud/service.go
```diff
@@ -0,0 +1,11 @@
+package fraud
+
+import "github.com/celestiaorg/celestia-node/fraud"
+
+// Module encompasses the behavior necessary to subscribe and broadcast
+// fraud proofs within the network.
+type Module interface {
```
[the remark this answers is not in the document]
vgonkivs: And retrning it back will also allow to rename back it to Service 🥲

nodebuilder/p2p/module.go
```diff
@@ -6,8 +6,8 @@ import (
 	"github.com/celestiaorg/celestia-node/nodebuilder/node"
 )
 
-// Module collects all the components and services related to p2p.
-func Module(tp node.Type, cfg *Config) fx.Option {
+// ConstructModule collects all the components and services related to p2p.
+func ConstructModule(tp node.Type, cfg *Config) fx.Option {
```
Wondertan, line 10: The previous Module thing was fine and according to the #958. I understand that the reasoning to change this is because API is now Module. IMO, we should still name those as API. WDYT?

service/rpc/state.go
```diff
@@ -6,10 +6,10 @@ import (
 	"errors"
 	"net/http"
 
+	"github.com/celestiaorg/celestia-node/state"
+
 	"github.com/cosmos/cosmos-sdk/types"
 	"github.com/gorilla/mux"
-
-	"github.com/celestiaorg/celestia-node/service/state"
 )
```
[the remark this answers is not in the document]
distractedm1nd: yes

nodebuilder/fraud/service.go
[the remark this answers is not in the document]
distractedm1nd: done

share/testing.go
```diff
@@ -26,30 +26,30 @@ import (
 	"github.com/celestiaorg/celestia-node/ipld"
 )
 
-// RandLightServiceWithSquare provides a share.Service filled with 'n' NMT
+// randLightServiceWithSquare provides a share.Service filled with 'n' NMT
 // trees of 'n' random shares, essentially storing a whole square.
-func RandLightServiceWithSquare(t *testing.T, n int) (*Service, *Root) {
+func randLightServiceWithSquare(t *testing.T, n int) (*Service, *Root) {
```
Wondertan, line 31: What's the reasoning behind privatizing all these? It's normal to provide testing utilities to outsiders. Thought we could make em public once the use case comes, and hide the for now. Not strictly opposed to this

distractedm1nd: They were just privatized because `Service` was `service` for a while, and it was a better solution to unexport than to tag each method with a nolint revive. But We can also keep them unexported if its not a big deal? I like the thinking of "we could make them public once the use case comes"

Wondertan: Yeah, let's keep it unexported. It is fine

Wondertan (approved):

Wondertan:
Hmm too many CI issues @distractedm1nd

distractedm1nd:
All tests pass for me locally, let's see if something flakes

Wondertan (approved):

PR #1060, opened 2022-09-08
feat(rpc): implementing staking reads and replace sdk.Address

### LINKED ISSUE #986: Implement staking module reads
https://github.com/cosmos/cosmos-sdk/blob/71035879e4af4551b9bcf16f79bc52785793602e/x/staking/types/querier.go#L17-L18

and

https://github.com/cosmos/cosmos-sdk/blob/71035879e4af4551b9bcf16f79bc52785793602e/x/staking/types/querier.go#L13

distractedm1nd:
Closes #986.

Also removes the abstraction of `sdk.Address` to make it explicit that our queries only support their corresponding address type, `AccAddress` or `ValAddress`

### DIFF
--- a/service/rpc/state.go
+++ b/service/rpc/state.go
@@ -13,17 +13,17 @@ import (
 )
 
 const (
-	balanceEndpoint           = "/balance"
-	submitTxEndpoint          = "/submit_tx"
-	submitPFDEndpoint         = "/submit_pfd"
-	transferEndpoint          = "/transfer"
-	delegationEndpoint        = "/delegate"
-	undelegationEndpoint      = "/begin_unbonding"
-	cancelUnbondingEndpoint   = "/cancel_unbond"
-	beginRedelegationEndpoint = "/begin_redelegate"
-	queryDelegationEndpoint   = "/query_delegation"
-	queryUnbondingEndpoint    = "/query_unbonding"
-	queryRedelegationEndpoint = "/query_redelegations"
+	balanceEndpoint            = "/balance"
+	submitTxEndpoint           = "/submit_tx"
+	submitPFDEndpoint          = "/submit_pfd"
+	transferEndpoint           = "/transfer"
+	delegationEndpoint         = "/delegate"
+	undelegationEndpoint       = "/begin_unbonding"
+	cancelUnbondingEndpoint    = "/cancel_unbond"
+	beginRedelegationEndpoint  = "/begin_redelegate"
+	queryDelegationEndpoint    = "/query_delegation"
+	queryUnbondingEndpoint     = "/query_unbonding"
+	queryRedelegationsEndpoint = "/query_redelegations"
 )
 
 var addrKey = "address"
@@ -78,7 +78,7 @@ type cancelUnbondRequest struct {
 	GasLimit uint64 `json:"gas_limit"`
 }
 
-type queryRedelegationRequest struct {
+type queryRedelegationsRequest struct {
 	From string `json:"from"`
 	To   string `json:"to"`
 }
@@ -367,9 +367,9 @@ func (h *Handler) handleQueryDelegation(w http.ResponseWriter, r *http.Request) 
 	}
 
 	// convert address to Address type
-	addr, addrerr := types.ValAddressFromBech32(addrStr)
-	if addrerr != nil {
-		writeError(w, http.StatusBadRequest, queryDelegationEndpoint, addrerr)
+	addr, err := types.ValAddressFromBech32(addrStr)
+	if err != nil {
+		writeError(w, http.StatusBadRequest, queryDelegationEndpoint, err)
 		return
 	}
 	delegation, err := h.state.QueryDelegation(r.Context(), addr)
@@ -398,9 +398,9 @@ func (h *Handler) handleQueryUnbonding(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// convert address to Address type
-	addr, addrerr := types.ValAddressFromBech32(addrStr)
-	if addrerr != nil {
-		writeError(w, http.StatusBadRequest, queryUnbondingEndpoint, addrerr)
+	addr, err := types.ValAddressFromBech32(addrStr)
+	if err != nil {
+		writeError(w, http.StatusBadRequest, queryUnbondingEndpoint, err)
 		return
 	}
 	unbonding, err := h.state.QueryUnbonding(r.Context(), addr)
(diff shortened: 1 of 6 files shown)


service/rpc/state.go
```diff
@@ -348,3 +356,98 @@ func (h *Handler) handleRedelegation(w http.ResponseWriter, r *http.Request) {
...
+	vars := mux.Vars(r)
+	addrStr, exists := vars[addrKey]
+	if !exists {
+		writeError(w, http.StatusBadRequest, queryUnbondingEndpoint, errors.New("address not specified"))
+		return
+	}
+
+	// convert address to Address type
+	addr, addrerr := types.ValAddressFromBech32(addrStr)
+	if addrerr != nil {
```
renaynay, line 402: same here

service/state/core_access.go
```diff
@@ -332,3 +336,58 @@ func (ca *CoreAccessor) Delegate(
...
+		DelegatorAddr: delAddr.String(),
+		ValidatorAddr: valAddr.String(),
+	})
+}
+func (ca *CoreAccessor) QueryRedelegations(
+	ctx context.Context,
+	srcValAddr,
+	dstValAddr Address,
+) (*stakingtypes.QueryRedelegationsResponse, error) {
+	_, ok := srcValAddr.(sdktypes.ValAddress)
```
renaynay, line 380: same here

service/rpc/state.go
```diff
@@ -75,6 +78,11 @@ type cancelUnbondRequest struct {
 	GasLimit uint64 `json:"gas_limit"`
 }
 
+type queryRedelegationRequest struct {
```
renaynay, line 81: `queryRedelegationsRequest`
```suggestion
type queryRedelegationsRequest struct {
```

walldiss (dismissed):

service/state/core_access.go
```diff
@@ -332,3 +336,58 @@ func (ca *CoreAccessor) Delegate(
 	}
 	return ca.SubmitTx(ctx, signedTx)
 }
+
+func (ca *CoreAccessor) QueryDelegation(
+	ctx context.Context,
+	valAddr Address,
+) (*stakingtypes.QueryDelegationResponse, error) {
+	_, ok := valAddr.(sdktypes.ValAddress)
```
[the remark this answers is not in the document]
distractedm1nd: Well, we still cast the `sdk.Address` to `AccAddress` for things like balances, this is just an extra sanity check. I would say if we change the parameter to ValAddress, we should change them in the other methods to AccAddress as well, no?

service/rpc/state.go
distractedm1nd: Also renamed `queryRedelegationEndpoint` to `queryRedelegationsEndpoint`

service/state/core_access.go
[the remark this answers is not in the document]
renaynay: Could do for consistency, sure.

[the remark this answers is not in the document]
distractedm1nd: done

service/state/state.go
```diff
@@ -15,8 +15,11 @@ type Tx = coretypes.Tx
 // TxResponse is an alias to the TxResponse type from Cosmos-SDK.
 type TxResponse = sdk.TxResponse
 
-// Address is an alias to the Address type from Cosmos-SDK.
-type Address = sdk.Address
+// ValAddress is an alias to the ValAddress type from Cosmos-SDK.
+type ValAddress = sdk.ValAddress
+
+// AccAddress is an alias to the AccAddress type from Cosmos-SDK.
+type AccAddress = sdk.AccAddress
```
Wondertan, line 22: What's the reason for this change?

Wondertan (commented):
The RPC part is LGTM to me. I am a bit concerned about types change here

service/state/state.go
distractedm1nd: CC: @renaynay 

To make it explicit what account type is accepted for each request, instead of being hidden in the implementation

Wondertan: Feel conflicted about this refactor slipping through. I like explicitness in the case where we work with ValAddress in delegation related methods. However, the Address abstraction should still be preserved for Transfer and Balance methods to allows operations for any kind of addresses!

distractedm1nd: it doesn't though. transactions and balances were already only using AccAddresses before

distractedm1nd: (they do AccAddressFromBech32 which only accepts user accounts)

Wondertan: We doesn't, but it does not meant that we shouldn't or won't. Current abstraction allows this, this _uninvited_ refactor prevents this. Once there is a user requesting feature to get balance of validator, we will readd this. 

If we really want to make this PR complete all features:
Extending support for ValAddress via Address abstraction > refactoring the code to remove abstraction at all.

distractedm1nd: Sorry but this was invited, it was part of rene's review. Let's discuss in Sync today.

renaynay: @Wondertan let me clarify: 

We have 4 things going on under the hood in state access: 
1. bank module writes (transfer, submitTx, submitPFD which is just fancy submitTx)
2. bank module reads (balance, balanceForAddress same thing)
3. staking module writes (Delegate, Redelegate, etc.)
4. staking module reads (queryDelegate, queryUnbonding, etc.)

1 and 2 access same account state tree via the bank store key (albeit in separate ways since we do proof verification for bank module reads)

3 and 4 access same staking state tree via staking store key (this happens via grpc query client)

There are different representations of addresses that are used as keys to read from the respective trees. IMO it's good to make it explicit for bank module reads vs staking module reads since the error that is returned when you try to use something other than AccAddress for bank module read from the app is quite ugly and difficult for the user to parse through (even though it's very low probability they run into it if they're using our key management utility).

renaynay: What we can and *should* do is clarify this in the PR title and description pls 🙏🏻 @distractedm1nd

renaynay (approved):
Implementation looks clean, thank you @distractedm1nd  

Let's please test every endpoint against `arabica` before we merge. 🙏🏻

service/state/state.go
distractedm1nd: #1085

Wondertan (approved):

distractedm1nd:
All endpoints including queries have been tested on arabica, BUT:
- Redelegation and Redelegation Query are not 100% tested because I can't do a redelegation with only one ValAddr. But the errors returned when redelegating to the same ValAddr are reassuring enough.

PR #1061, opened 2022-09-08
chore: resolving linter errors on main

distractedm1nd:
On main rn there are two linter errors, making any branches need to fix the linting errors when a PR is opened. This fixes that

So `fx.Extract` got deprecated (actually it was soft-deprecated a while ago but they recently forced the deprecation)
The solution is to throw the `Node` struct directly into `fx.Populate`, meaning we need to add `fx.In` to the struct, and ignore the unexported fields (start, stop)

Other one is probably from fraudsync, as its in `fraud/testing.go`

### DIFF
--- a/fraud/testing.go
+++ b/fraud/testing.go
@@ -108,7 +108,7 @@ func (m *mockProof) Height() uint64 {
 }
 
 func (m *mockProof) Validate(*header.ExtendedHeader) error {
-	if m.Valid != true {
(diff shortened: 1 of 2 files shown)


renaynay (approved):
thanks for this, sad that lint errs got into main

can you make a quick note in pr description re fx extract deprecation pls?

Wondertan (approved):

Issue #873, opened 2022-06-29
[Feature Request]: OpenRPC integration into Celestia Node


YazzyYaz:
### Implementation ideas

With the Node API having more and more features and endpoints added, it might be worth to use OpenRPC to spin up a self-documenting API endpoint with a playground included whenever there's a new Node release.

https://playground.open-rpc.org/

This can allows us to brand our API more with it's own website and playground we can deploy and link developers to from our documentation page.

Wondertan:
Linking #169 here

Wondertan:
The way https://github.com/filecoin-project/lotus does OpenRPC things:
* In-house docsgen: https://github.com/filecoin-project/lotus/tree/master/api/docgen-openrpc
* Example of autogenerated docs: https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-v0-methods.md
* The way Go clients are "generated" https://github.com/filecoin-project/lotus/blob/master/api/client/client.go.
	* This is not code generation. github.com/filecoin-project/go-jsonrpc takes any interface and provides back an implementation of it that can talk to a remote instance also set up with go-jsonrpc.
* The server side is similar https://github.com/filecoin-project/lotus/blob/master/node/rpc.go#L73. You give an implementation of an interface to the jsonrpc and that's all. Super simple, yet powerful. 
	* In comparison, in geth, you need to do [this](https://github.com/ethereum/go-ethereum/blob/master/consensus/consensus.go#L118).
* The example interface can be found in https://github.com/filecoin-project/lotus/blob/master/api/api_full.go#L62.
	* Also, check out the comment on how to update it.
* The autogeneration makefile rules they have https://github.com/filecoin-project/lotus/blob/master/Makefile#L334-L360
* @meowsbits's PR which adds OpenRPC support https://github.com/filecoin-project/lotus/pull/4711(finished in https://github.com/filecoin-project/lotus/pull/5843)
	* Interesting point form MagiK - (go-ethereum being LGPL is the reason why we've implemented go-jsonrpc in the first place)
* JS Client https://github.com/filecoin-shipyard/js-lotus-client. Also, check a link there. Don't really know how it works.

cc @tzdybal, @renaynay, @YazzyYaz

distractedm1nd:
It seems like the tooling around OpenAPI/Swagger is much more mature than OpenRPC, can you please give your perspective on why you prefer it @Wondertan?

distractedm1nd:
[...]

PR #822, opened 2022-06-14
fix: fix linter

vgonkivs:

PR #1062, opened 2022-09-08
feat(rpc): writing error out as json

distractedm1nd:
Our endpoints should write the error out in a JSON body to be API friendly, so instead of just

`rpc error: code = NotFound desc = account celestia1ljj5eduqftemu3vpkyd4z80knr2rprg8mqd5sl not found`

we write

```json
{
    "error": "rpc error: code = NotFound desc = account celestia1ljj5eduqftemu3vpkyd4z80knr2rprg8mqd5sl not found"
}
```

### DIFF
--- a/node/rpc_test.go
+++ b/node/rpc_test.go
@@ -92,11 +92,11 @@ func testGetNamespacedRequest(t *testing.T, endpointName string, assertResponseO
 				require.False(t, resp.StatusCode == http.StatusOK)
 				require.Equal(t, "application/json", resp.Header.Get("Content-Type"))
 
-				var errorMessage string
+				errorMessage := new(rpc.JSONError)
 				err := json.NewDecoder(resp.Body).Decode(&errorMessage)
 
 				require.NoError(t, err)
-				require.Equal(t, tt.errMsg, errorMessage)
+				require.Equal(t, tt.errMsg, errorMessage.Err)
 
 				return
 			}
--- a/node/rpc_test.go
+++ b/node/rpc_test.go
@@ -93,7 +93,7 @@ func testGetNamespacedRequest(t *testing.T, endpointName string, assertResponseO
 				require.Equal(t, "application/json", resp.Header.Get("Content-Type"))
 
 				errorMessage := new(rpc.JSONError)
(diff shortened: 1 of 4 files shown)


distractedm1nd:
ugh, contains the lint fixes for main as well, in #822. doesnt matter anyways i guess

Wondertan:
https://github.com/celestiaorg/go-cnc/blob/main/client.go

distractedm1nd:
Thats fine, but yes we will have to fix go-cnc then - but this is definitely the expected api behavior

node/rpc_test.go
```diff
@@ -92,11 +92,11 @@ func testGetNamespacedRequest(t *testing.T, endpointName string, assertResponseO
 				require.False(t, resp.StatusCode == http.StatusOK)
 				require.Equal(t, "application/json", resp.Header.Get("Content-Type"))
 
-				var errorMessage string
+				errorMessage := new(rpc.JSONError)
 				err := json.NewDecoder(resp.Body).Decode(&errorMessage)
```
walldiss, line 96: nit: errorMessage could be passed as is, because it is a pointer type

```suggestion
				errorMessage := new(rpc.JSONError)
				err := json.NewDecoder(resp.Body).Decode(errorMessage)
```

distractedm1nd: Thanks! nice catch

distractedm1nd:
It is valid JSON, but it is not intuitive or standard. I've never seen a API just return a raw string on an error, and would also hate that as a user - but open for discussion here then

tzdybal:
This API was designed as temporary solution, just to expose critical functionalities in simplest way. Originally, we planned to replace this API with new one instead of making incremental changes (if I recall correctly). The reason for this is to avoid cosmetic changes, that will need fixes across repositories, while we still plan to completely change everything (for example: https://github.com/celestiaorg/celestia-node/issues/873).

Wondertan:
@distractedm1nd, let's keep this a draft then

renaynay:
Should we close this and open an issue on the epic for better errs @distractedm1nd ?

distractedm1nd:
Yes we can close because whatever standard we use also standardizes error outputs.

PR #1073, opened 2022-09-12
feat(cmd): Provide ability to specify network to user

### LINKED ISSUE #1066: params | cmd: Provide ability to specify network to user
Remove panic when using `CELESTIA_CUSTOM` during re-init
Currently, node panics when we try to re-initialize an existing store. 
Maybe an error log with confirmation flag `--yes` to proceed further can help eliminate this over-protective behaviour

distractedm1nd:
Closes #1066 .

- node store, extend home directory with network name
- Keep CELESTIA_CUSTOM, deprecate CELESTIA_PRIVATE_GENESIS
    - panic will be taken care of by the node store naming extension
- Add `--node.network < arabica || mamaki || ... >`
- Fixes cel-shed and cel-key packages

### DIFF
--- a/cmd/cel-key/node_types.go
+++ b/cmd/cel-key/node_types.go
@@ -2,25 +2,35 @@ package main
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/spf13/cobra"
 	flag "github.com/spf13/pflag"
+
+	"github.com/celestiaorg/celestia-node/params"
 
 	sdkflags "github.com/cosmos/cosmos-sdk/client/flags"
 )
 
 var (
-	nodeDirKey = "node.type"
-
-	bridgeDir = "~/.celestia-bridge/keys"
-	fullDir   = "~/.celestia-full/keys"
-	lightDir  = "~/.celestia-light/keys"
+	nodeDirKey     = "node.type"
+	nodeNetworkKey = "node.network"
 )
 
 func DirectoryFlags() *flag.FlagSet {
 	flags := &flag.FlagSet{}
-	flags.String(nodeDirKey, "", "Sets key utility to use the node type's directory (e.g. "+
-		"~/.celestia-light if --node.type light is passed).")
+	defaultNetwork := string(params.DefaultNetwork())
+
+	flags.String(
+		nodeDirKey,
+		"",
+		"Sets key utility to use the node type's directory (e.g. "+
+			"~/.celestia-light-"+strings.ToLower(defaultNetwork)+" if --node.type light is passed).")
+	flags.String(
+		nodeNetworkKey,
+		defaultNetwork,
+		"Sets key utility to use the node network's directory (e.g. "+
+			"~/.celestia-light-mynetwork if --node.network MyNetwork is passed).")
 
 	return flags
 }
@@ -31,17 +41,11 @@ func ParseDirectoryFlags(cmd *cobra.Command) error {
 		return nil
 	}
 
+	network := cmd.Flag(nodeNetworkKey).Value.String()
 	switch nodeType {
-	case "bridge":
-		if err := cmd.Flags().Set(sdkflags.FlagKeyringDir, bridgeDir); err != nil {
-			return err
-		}
-	case "full":
-		if err := cmd.Flags().Set(sdkflags.FlagKeyringDir, fullDir); err != nil {
-			return err
-		}
-	case "light":
-		if err := cmd.Flags().Set(sdkflags.FlagKeyringDir, lightDir); err != nil {
+	case "bridge", "full", "light":
+		keyPath := fmt.Sprintf("~/.celestia-%s-%s", nodeType, strings.ToLower(network))
+		if err := cmd.Flags().Set(sdkflags.FlagKeyringDir, keyPath); err != nil {
 			return err
 		}
 	default:
--- a/cmd/flags_node.go
+++ b/cmd/flags_node.go
@@ -9,20 +9,22 @@ import (
 	flag "github.com/spf13/pflag"
 
 	"github.com/celestiaorg/celestia-node/node"
+	"github.com/celestiaorg/celestia-node/params"
 )
 
 var (
-	nodeStoreFlag  = "node.store"
-	nodeConfigFlag = "node.config"
+	nodeStoreFlag   = "node.store"
+	nodeConfigFlag  = "node.config"
+	nodeNetworkFlag = "node.network"
 )
 
 // NodeFlags gives a set of hardcoded Node package flags.
-func NodeFlags(tp node.Type) *flag.FlagSet {
+func NodeFlags() *flag.FlagSet {
 	flags := &flag.FlagSet{}
 
 	flags.String(
 		nodeStoreFlag,
-		fmt.Sprintf("~/.celestia-%s", strings.ToLower(tp.String())),
+		"",
 		"The path to root/home directory of your Celestia Node Store",
 	)
 	flags.String(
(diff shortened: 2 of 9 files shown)


cmd/cel-key/node_types.go
```diff
@@ -31,17 +41,11 @@ func ParseDirectoryFlags(cmd *cobra.Command) error {
...
-		}
-	case "full":
-		if err := cmd.Flags().Set(sdkflags.FlagKeyringDir, fullDir); err != nil {
-			return err
-		}
-	case "light":
-		if err := cmd.Flags().Set(sdkflags.FlagKeyringDir, lightDir); err != nil {
+	case "bridge", "full", "light":
+		keyPath := fmt.Sprintf("~/.celestia-%s-%s", nodeType, strings.ToLower(network))
+		if err := cmd.Flags().Set(sdkflags.FlagKeyringDir, keyPath); err != nil {
```
renaynay, line 48: beautiful.

cmd/flags_node.go
```diff
@@ -9,34 +9,53 @@ import (
...
 
 	return flags
 }
 
 // ParseNodeFlags parses Node flags from the given cmd and applies values to Env.
 func ParseNodeFlags(ctx context.Context, cmd *cobra.Command) (context.Context, error) {
-	ctx = WithStorePath(ctx, cmd.Flag(nodeStoreFlag).Value.String())
+	network := cmd.Flag(nodeNetworkFlag).Value.String()
+	if network != "" {
+		params.SetDefaultNetwork(params.Network(network))
```
renaynay, line 48: we need to sanity check the network, no? You want to allow setting custom via the `--network` flag? The issue is, we also need genesis hash and bootstrappers for a custom network / private network.

params/default.go
```diff
@@ -19,6 +18,10 @@ func DefaultNetwork() Network {
 	return defaultNetwork
 }
 
+func SetDefaultNetwork(net Network) {
```
renaynay, line 21: I'm not sure about this approach.

cmd/flags_node.go
distractedm1nd: validation is now done in SetDefaultNetwork and an error is passed back up

params/default.go
distractedm1nd: me either

renaynay (approved):
Tested, with and without CELESTIA_CUSTOM opts too

CELESTIA_CUSTOM will only panic if it sees a store of the same network name that has been initialised already :)

cmd/flags_node.go
```diff
@@ -9,34 +9,56 @@ import (
...
 		"The path to root/home directory of your Celestia Node Store",
 	)
 	flags.String(
 		nodeConfigFlag,
 		"",
 		"Path to a customized node config TOML file",
 	)
+	flags.String(
+		nodeNetworkFlag,
+		"",
```
Wondertan, line 37: nit: We can set `params.DefaultNetwork()` here so that when we `ParseNodeFlags`, the `network` is never empty, and this reduces code branchiness.

cmd/flags_node.go
```diff
@@ -9,34 +9,56 @@ import (
...
+		"The name of the network to connect to",
```
Wondertan, line 38: nit: we can list all the networks from params pkg here.

cmd/flags_node.go
```diff
@@ -9,34 +9,56 @@ import (
...
 
 // NodeFlags gives a set of hardcoded Node package flags.
-func NodeFlags(tp node.Type) *flag.FlagSet {
+func NodeFlags() *flag.FlagSet {
 	flags := &flag.FlagSet{}
 
 	flags.String(
 		nodeStoreFlag,
-		fmt.Sprintf("~/.celestia-%s", strings.ToLower(tp.String())),
+		"",
```
Wondertan, line 27: Why remove the default? It is convenient, especially for users printing out help for the CLI

params/default.go
```diff
@@ -19,6 +18,14 @@ func DefaultNetwork() Network {
...
 }
 
+func SetDefaultNetwork(net Network) error {
+	if err := net.Validate(); err != nil {
+		return err
+	}
+	defaultNetwork = net
+	return nil
+}
+
```
Wondertan, line 28: We should make this a Node Option. Writing to global will create issues going further and brakes further celestia as a library approach.

Wondertan (changes requested):
Requesting changes to keep the global state of params pkg being read-only and keeping network selection a Node instance specific. We should also keep defaults on the flags, so they explicit on the flag helps

cmd/flags_node.go
distractedm1nd: The main reason for doing this is so that it is clear when a user overrides it: This is helpful because when it is done explicitly, we definitely need to call SetDefaultNetwork, because the params will not hear about it otherwise.

params/default.go
renaynay: Can we make an issue for this? It will change via refactoring, would rather not introduce new options.

cmd/flags_node.go
distractedm1nd: We do this for a similar reason as above: When a user overrides the default network, it needs to be placed in that string. If we create a default, we have to override it anyways when parsing.

params/default.go
distractedm1nd: This sounds like something that should be put into the refactoring tasks then, I do not feel like it belongs here.

Wondertan: I am fine with postponing this as long as you keep this part of the refactoring efforts you take.

cmd/flags_node.go
distractedm1nd: For this reason it will not reduce code branchiness: We will have to test if it is custom anyways, and then have some weird/complex logic to determine if it was the default value or not

params/default.go
Wondertan: It belongs here because it breaks the encapsulation of params pkg that it was designed with, but again I am fine with postponing this if you promise to fix this later on.

cmd/flags_node.go
Wondertan: It's not an issue when you don't override params pkg

params/default.go
distractedm1nd: Yes I mean I definitely agree, just think it fits much better with the refactorings than here. Whatever I do here will be superseeded by the refactorings anyways

cmd/flags_node.go
distractedm1nd: Nothing is an issue when you don't do the thing that causes the issue

renaynay: What do you mean?

cmd/flags_node.go
renaynay: This would be nice to include `e.g. arabica or mamaki`

params/default.go
renaynay: I also dislike this so will make sure to include it in refactorings -- can even rebase current refactoring PR and include as a commit.

renaynay: rebase current refactoring PR on main once this is merged, is what i meant

Wondertan (dismissed):
Thanks, and looking forward to implementation of #1076

params/network.go
```diff
@@ -39,3 +39,16 @@ var networksList = map[Network]struct{}{
...
+func ListProvidedNetworks() string {
+	var networks string
+	for net := range networksList {
+		// "private" network isn't really a choosable option, so skip
+		if net != Private {
+			networks += string(net) + ", "
+		}
+	}
+	// chop off trailing ", "
+	return networks[:len(networks)-2]
```
Wondertan, line 53: `strings.Builder` is the true way, but the gains are negligible, so approve

Wondertan (approved):
Thanks!

renaynay (approved):
APPROOOOOVE

PR #1067, opened 2022-09-12
chore(deps): Upgrading to go 1.19

distractedm1nd:
The biggest change with 1.19 in our existing codebase is godocs.

Some things to note:
- It kinda breaks multiline TODO comments in Goland
- Nested lists are not supported, so try to mix list markers instead

PR #1079, opened 2022-09-12
header/store: use an atomic.Pointer instead of Lock

### LINKED ISSUE #864: header/store: Use `RLock` instead of `Lock` inside `Append` during adjacency verification
Ref https://github.com/celestiaorg/celestia-node/issues/861#issuecomment-1167378119

Martichou:
The verification of the headers can be done with only a RLock. By doing so, we provide a "very minor optimization" that can avoid the Lock to wait for processIncoming to finish (as he can take the lock).

Moving the verification to a separate function so that the deref for the RLock is scoped only to the needed part.
An alternative to this would be to get rid of the `defer s.writeLk.RUnlock()` and add a `s.writeLk.RUnlock()` at each return/exit of the function manually (but this wouldn't handle panic!() cases), so might be a no-go.

Close #864

### DIFF
--- a/header/store/store.go
+++ b/header/store/store.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"sync"
+	"sync/atomic"
 
 	logging "github.com/ipfs/go-log/v2"
 
@@ -53,13 +53,12 @@ type store struct {
 
 	// writing to datastore
 	//
-	writeLk sync.Mutex
 	// queue of headers to be written
 	writes chan []*header.ExtendedHeader
 	// signals when writes are finished
 	writesDn chan struct{}
 	// writeHead maintains the current write head
-	writeHead *header.ExtendedHeader
+	writeHead atomic.Pointer[header.ExtendedHeader]
 	// pending keeps headers pending to be written in one batch
 	pending *batch
 }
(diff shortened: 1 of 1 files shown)


Wondertan (changes requested):
Thanks for the contribution. 

The code LGTM, but it _really_ does not give us anything in the way it is used currently by Syncer. The (negligible, if existing) contention between reads and writes is still there even with RWMutex. The change would give us something if there were multiple processIncoming routines trying to Append.

We still can proceed with this though by keeping the Header atomically instead of locking it with new [atomic.Pointer](https://pkg.go.dev/sync/atomic#Pointer.Swap) in go1.19.

Let's do this after our update in #1067

Martichou:
Hi, thanks for the feedback ! 

I know it's not a "very important" nor really useful change, it was mainly for me to take my hands on the project's code :).
I'll keep an eyes on #1067 and update this PR accordingly.

Wondertan:
Pls rebase on main, rather than merge main in

Martichou:
@Wondertan oups, my bad ! (done)

Wondertan:
@Martichou, so the #1067 was finally merged, so we can rebase this one and merge it as well 🎉

Martichou:
@Wondertan done ! :tada:

Martichou:
@Wondertan from the CI trace it seems like `TestFullReconstructFromLights` failed with:
```
2022-09-21T15:17:00.2118990Z     reconstruct_test.go:180: 
2022-09-21T15:17:00.2119894Z         	Error Trace:	/home/runner/work/celestia-node/celestia-node/node/tests/reconstruct_test.go:180
2022-09-21T15:17:00.2120464Z         	Error:      	Received unexpected error:
2022-09-21T15:17:00.2121307Z         	            	da: data not available
2022-09-21T15:17:00.2121804Z         	Test:       	TestFullReconstructFromLights
```

I've ran the tests locally (with and without race detection) and they passed everytimes. Do you have any idea on what can cause such issue on the CI ?

renaynay (approved):
Ran this against arabica, looks good to me.

renaynay:
@Martichou likely flakiness in our tests - your code doesn't impact this.

Wondertan (approved):

PR #1080, opened 2022-09-13
Update node_types.go

### LINKED ISSUE #1074: fix(cmd/cel-key): Add error when no node type is provided to cel-key
In our cel-key utility, we should error out with "no node type provided" during the following:

https://github.com/celestiaorg/celestia-node/blob/8657edc37d3911d42de447d9d0c93c4676a148c9/cmd/cel-key/node_types.go#L29-L32

0xJepsen:
Closes #1074 

Let me know if you have something else in mind, and I will happy to revise it.


distractedm1nd:
@0xJepsen Thank you for your contribution! Since the method returns an error, can we return an `errors.New` instead of panicking?

renaynay:
Thank you for your contribution @0xJepsen but we already merged. #1081

We have many other `good first issue`s though!

PR #1083, opened 2022-09-13
feat(metrics): metrics for PFD transactions, sampling, and broadcasts

### LINKED ISSUE #1040: [Feature Request]: metrics for PayForData and last sample requested

distractedm1nd:
Closes #1040 

First time playing with metrics, so unsure if this is the right approach

| Metric Name                              | Light node | Full Node | Bridge node | Implemented |
| ---------------------------------------- | ---------- | ----------------- | ----------- | ----- |
| Total PayForData messages sent (count)   | x          | x                 | x           | ✅ |
| Last PayForData message sent (timestamp) | x          | x                 | x           | ✅ |
| Last sample requested (timestamp)        | x          | x                 |             | ✅ |
| Last block served (timestamp)            |            |                   | x           | ❌ |
| Last full sampling request (timestamp)   |            |                  | x           | ❌

- `Last full sampling req` is not implemented because it is the same as `Last sample requested`
- `Last block served` is not possible until we have implemented the networking of the new BlockSync
- All metrics reporters are now identifiable through their `peer.ID` 
- Tested on grafana cloud

I am marking as out of draft for reviews but this PR should not be merged until #997 is merged, because rebasing on this would take more time than this being rebased.


distractedm1nd:
Started rebasing, but realized that it will be easier once #1125 is merged first - let's just make sure that the "Last sampled timestamp" makes it in there

Wondertan:
Wanted to review this, but will wait for the #1125 to land first

nodebuilder/settings.go
```diff
@@ -21,10 +35,71 @@ func WithBootstrappers(peers params.Bootstrappers) fx.Option {
...
+	baseComponents := fx.Options(
+		fx.Supply(metricOpts),
+		fx.Invoke(InitializeMetrics),
+		fx.Invoke(header.MonitorHead),
+		fx.Invoke(state.MonitorPFDs),
+	)
+
+	var opts fx.Option
+	switch nodeType {
+	case node.Full:
```
renaynay, line 52: is there a reason to separate case node.Full and case node.Light? Unless it is expected to add diff metrics for full and light in the near future.

header/core/listener.go
```diff
@@ -25,6 +26,9 @@ type Listener struct {
 	bServ     blockservice.BlockService
 	construct header.ConstructFn
 	cancel    context.CancelFunc
+
+	// lastHeaderBroadcasted is the timestamp of the last block served by the broadcaster, used for metrics.
+	lastHeaderBroadcasted int64
```
renaynay, line 31: i would call this `lastHeaderBroadcast` bc that implies you're measuring the broadcast time rather than the header height

nodebuilder/settings.go
```diff
@@ -21,10 +35,71 @@ func WithBootstrappers(peers params.Bootstrappers) fx.Option {
...
+		)),
 	)
+
+	err = pusher.Start(ctx)
+	if err != nil {
+		return err
+	}
+	global.SetMeterProvider(pusher)
+
+	return nil
```
renaynay, line 104:
```suggestion
	}
	
	global.SetMeterProvider(pusher)
	return nil
```

renaynay (commented):
teeny nits

header/core/listener.go
distractedm1nd: done

nodebuilder/settings.go
distractedm1nd: done

vgonkivs (approved):

service/state/metrics.go
```diff
@@ -0,0 +1,35 @@
...
+	"context"
+
+	"go.opentelemetry.io/otel/metric/global"
+	"go.opentelemetry.io/otel/metric/instrument"
+	"go.opentelemetry.io/otel/metric/unit"
+)
+
+var meter = global.MeterProvider().Meter("state")
+
+func MonitorPFDs(s *Service) {
```
Wondertan, line 13: What about `WithMetrics` so that all the modules on `nodebuilder` follow the same pattern? Even though it does PFD monitoring only, for now.

The main reason for `WithMetrics` because we are not planning to granular select which piece of the stack you want to monitor, so it's either `all or nothing` or `all for a module or nothing`.

header/core/listener.go
```diff
@@ -97,6 +101,7 @@ func (cl *Listener) listen(ctx context.Context, sub <-chan *types.Block) {
 
 			// broadcast new ExtendedHeader, but if core is still syncing, notify only local subscribers
 			err = cl.bcast.Broadcast(ctx, eh, pubsub.WithLocalPublication(syncing))
+			cl.lastHeaderBroadcast = time.Now().UnixMilli()
```
Wondertan, line 104: Hmm, I don't really get why we would need this metric. It's also not part of the table.

service/state/metrics.go
distractedm1nd: Yes, agree

nodebuilder/settings.go
```diff
@@ -21,10 +36,65 @@ func WithBootstrappers(peers params.Bootstrappers) fx.Option {
...
+			semconv.ServiceNameKey.String(fmt.Sprintf("Celestia-%s", nodeType.String())),
+			// TODO(@Wondertan): Versioning: semconv.ServiceVersionKey
+			semconv.ServiceInstanceIDKey.String(peerID.String()),
+		)),
+	)
+
+	err = pusher.Start(ctx)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
distractedm1nd: agreed

header/core/listener.go
distractedm1nd: I can remove it if you want - source: you commented in #1040 that it wasnt in the table but would be useful to keep

Just respond here and i will either remove/resolve

Wondertan: Quote from that issue:
 > The issue is about the last block served, not the header.
 
 Here we broadcast the header. Which IMO is an entirely useless metric.

distractedm1nd: image

distractedm1nd: will remove though, understood

nodebuilder/settings.go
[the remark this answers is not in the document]
distractedm1nd: done

service/state/metrics.go
distractedm1nd: done

header/core/listener.go
distractedm1nd: done

renaynay (dismissed):

nodebuilder/settings.go
```diff
@@ -21,10 +34,74 @@ func WithBootstrappers(peers params.Bootstrappers) fx.Option {
...
+			// add more monitoring here
+		)
+	default:
+		panic("invalid node type")
+	}
+	return opts
+}
+
+// InitializeMetrics initializes the global meter provider.
+func InitializeMetrics(
```
Wondertan, line 69: Let's make it private and lfg

Wondertan (approved):

nodebuilder/settings.go
```diff
@@ -21,10 +34,74 @@ func WithBootstrappers(peers params.Bootstrappers) fx.Option {
 }
 
 // WithMetrics enables metrics exporting for the node.
-func WithMetrics(enable bool) fx.Option {
-	return fx.Options(
-		// TODO: Add metrics to more modules and enable here.
-		headermodule.WithMetrics(enable),
-		daser.WithMetrics(enable),
+func WithMetrics(enable bool, metricOpts []otlpmetrichttp.Option, nodeType node.Type) fx.Option {
```
renaynay, line 37: Where is `WithMetrics` called that needs to pass in `false`? Let's remove this param?

renaynay: it just returns fx.Option so no need to short-circuit with the `enable` param

distractedm1nd: done

Wondertan (approved):

renaynay (approved):

PR #1089, opened 2022-09-14
refactor(service/state): Fix log to return correct height of balance request

renaynay:
Self explanatory.


Wondertan (approved):
Minus one because of the deffered execution. Had to recall these and lgtm

distractedm1nd (approved):

Issue #1077, opened 2022-09-12
header/store: More detailed log for non-adjacent header application attempt


renaynay:
Current: 
```
2022-09-12T19:10:04.630+0200	ERROR	header/sync	sync/sync.go:188	syncing headers	{"from": 17409, "to": 17945, "err": "header/store: non-adjacent"}
```

We should log out the height that was attempted to be applied and the current head of the store as well.

PR #1123, opened 2022-09-19
log(header): Add detailed error for non-adjacent header attempt

renaynay:
resolves https://github.com/celestiaorg/celestia-node/issues/1077

### DIFF
--- a/header/sync/sync_head.go
+++ b/header/sync/sync_head.go
@@ -98,13 +98,14 @@ func (s *Syncer) networkHead(ctx context.Context) (*header.ExtendedHeader, error
 func (s *Syncer) incomingNetHead(ctx context.Context, netHead *header.ExtendedHeader) pubsub.ValidationResult {
 	// Try to short-circuit netHead with append. If not adjacent/from future - try it as new network header
 	_, err := s.store.Append(ctx, netHead)
-	switch err {
-	case nil:
+	if err == nil {
 		// a happy case where we appended maybe head directly, so accept
 		return pubsub.ValidationAccept
-	case header.ErrNonAdjacent:
+	}
+	var nonAdj *header.ErrNonAdjacent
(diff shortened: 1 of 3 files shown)


header/verify.go
```diff
@@ -49,6 +49,8 @@ func (eh *ExtendedHeader) VerifyNonAdjacent(untrst *ExtendedHeader) error {
 // VerifyAdjacent validates adjacent untrusted header against trusted 'eh'.
 func (eh *ExtendedHeader) VerifyAdjacent(untrst *ExtendedHeader) error {
 	if untrst.Height != eh.Height+1 {
+		log.Errorw("non-adjacent header attempted", "trusted head", eh.Height,
+			"attempted header", untrst.Height)
 		return ErrNonAdjacent
```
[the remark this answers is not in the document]
renaynay: done

header/sync/sync_head.go
```diff
@@ -98,13 +98,14 @@ func (s *Syncer) networkHead(ctx context.Context) (*header.ExtendedHeader, error
...
+	if err == nil {
 		// a happy case where we appended maybe head directly, so accept
 		return pubsub.ValidationAccept
-	case header.ErrNonAdjacent:
+	}
+	var nonAdj *header.ErrNonAdjacent
+	if errors.As(err, &nonAdj) {
 		// not adjacent, maybe we've missed a few headers or its from the past
-	default:
+	} else {
```
Wondertan, line 108: Wonder if there is a way to shorten/avoid `nonAdj` in case we don't need it's body. 

We can add a debug log here as well

renaynay: Let's just leave it as is -- added debug log.

walldiss (approved):

Wondertan (approved):

PR #1125, opened 2022-09-19
das: add metrics to DASer

walldiss:
~~Creating PR based on DASer PR. Most of the changes are in `metrics.go` and there are few additional lines in `das.go`, `coordinator.go`, `worker.go`.~~

Metrics preview:
image

Resolves 
https://github.com/celestiaorg/celestia-node/issues/1064

### DIFF
--- a/das/daser.go
+++ b/das/daser.go
@@ -90,7 +90,7 @@ func (d *DASer) Start(ctx context.Context) error {
 	}
 
 	if err = d.sampler.initMetrics(); err != nil {
-		return fmt.Errorf("init sampling coordinator metrics: %w", err)
+		return fmt.Errorf("das: init sampling coordinator metrics: %w", err)
 	}
 
 	// load latest DASed checkpoint
--- a/das/daser.go
+++ b/das/daser.go
@@ -89,10 +89,6 @@ func (d *DASer) Start(ctx context.Context) error {
 		return err
 	}
 
-	if err = d.sampler.initMetrics(); err != nil {
-		return fmt.Errorf("das: init sampling coordinator metrics: %w", err)
-	}
-
 	// load latest DASed checkpoint
 	cp, err := d.store.load(ctx)
 	if err != nil {
--- a/das/daser.go
+++ b/das/daser.go
@@ -89,6 +89,10 @@ func (d *DASer) Start(ctx context.Context) error {
 		return err
 	}
 
+	if err = d.sampler.initMetrics(); err != nil {
+		return fmt.Errorf("init sampling coordinator metrics: %w", err)
+	}
+
 	// load latest DASed checkpoint
 	cp, err := d.store.load(ctx)
 	if err != nil {
--- a/das/metrics.go
+++ b/das/metrics.go
@@ -3,12 +3,12 @@ package das
 import (
 	"context"
 	"fmt"
+	"sync/atomic"
 	"time"
 
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/metric/global"
 	"go.opentelemetry.io/otel/metric/instrument"
-	"go.opentelemetry.io/otel/metric/instrument/asyncint64"
 	"go.opentelemetry.io/otel/metric/instrument/syncfloat64"
 	"go.opentelemetry.io/otel/metric/instrument/syncint64"
 
@@ -20,63 +20,72 @@ var (
 )
 
 type metrics struct {
-	sampled    syncint64.Counter
-	sampleTime syncfloat64.Histogram
-
-	newHead syncint64.Counter
-
-	busyWorkers        asyncint64.Gauge
-	isRunning          asyncint64.Gauge
-	networkHead        asyncint64.Gauge
-	sampledChainHead   asyncint64.Gauge
+	sampled       syncint64.Counter
+	sampleTime    syncfloat64.Histogram
+	getHeaderTime syncfloat64.Histogram
+	newHead       syncint64.Counter
+	lastSampledTS int64
 }
 
 func (sc *samplingCoordinator) initMetrics() error {
-	sampled, err := meter.SyncInt64().Counter("das_sampled_headers_count",
+	sampled, err := meter.SyncInt64().Counter("das_sampled_headers_counter",
 		instrument.WithDescription("sampled headers counter"))
 	if err != nil {
 		return err
 	}
 
-	sampleTimeHist, err := meter.SyncFloat64().Histogram("das_sample_time_hist",
-		instrument.WithDescription("header sampling time"))
+	sampleTime, err := meter.SyncFloat64().Histogram("das_sample_time_hist",
+		instrument.WithDescription("duration of sampling a single header"))
 	if err != nil {
 		return err
 	}
 
-	newHead, err := meter.SyncInt64().Counter("das_head_updated",
+	getHeaderTime, err := meter.SyncFloat64().Histogram("das_get_header_time_hist",
+		instrument.WithDescription("duration of getting header from header store"))
+	if err != nil {
+		return err
+	}
+
+	newHead, err := meter.SyncInt64().Counter("das_head_updated_counter",
 		instrument.WithDescription("amount of times DAS'er advanced network head"))
 	if err != nil {
 		return err
 	}
 
-	busyWorkers, err := meter.AsyncInt64().Gauge("das_busy_workers_count",
+	lastSampledTS, err := meter.AsyncInt64().Gauge("das_latest_sampled_ts",
+		instrument.WithDescription("latest sampled timestamp"))
+	if err != nil {
+		return err
+	}
+
+	busyWorkers, err := meter.AsyncInt64().Gauge("das_busy_workers_amount",
 		instrument.WithDescription("number of active parallel workers in DAS'er"))
 	if err != nil {
 		return err
 	}
 
-	isRunning, err := meter.AsyncInt64().Gauge("das_is_running",
-		instrument.WithDescription("indicates whether das is running"))
-	if err != nil {
-		return err
-	}
-
-	headHeight, err := meter.AsyncInt64().Gauge("das_network_head",
+	networkHead, err := meter.AsyncInt64().Gauge("das_network_head",
 		instrument.WithDescription("most recent network head"))
 	if err != nil {
 		return err
 	}
 
-	headOfSampledChain, err := meter.AsyncInt64().Gauge("das_sampled_chain_head",
+	sampledChainHead, err := meter.AsyncInt64().Gauge("das_sampled_chain_head",
 		instrument.WithDescription("height of the sampled chain - all previous headers have been successfully sampled"))
 	if err != nil {
 		return err
 	}
 
+	sc.metrics = &metrics{
+		sampled:       sampled,
+		sampleTime:    sampleTime,
+		getHeaderTime: getHeaderTime,
+		newHead:       newHead,
+	}
+
 	err = meter.RegisterCallback(
 		[]instrument.Asynchronous{
-			busyWorkers, isRunning, headHeight, headOfSampledChain,
+			lastSampledTS, busyWorkers, networkHead, sampledChainHead,
 		},
 		func(ctx context.Context) {
 			stats, err := sc.stats(ctx)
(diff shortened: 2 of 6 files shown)


das/metrics.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	return nil
+}
+
+func (m *metrics) observeSample(ctx context.Context, h *header.ExtendedHeader, sampleTime time.Duration, err error) {
+	if m == nil {
+		return
+	}
+	m.sampleTime.Record(ctx, sampleTime.Seconds(),
+		attribute.Bool("failed", err != nil),
```
Wondertan, line 120: Why not report the error itself?

das/metrics.go
```diff
@@ -0,0 +1,132 @@
...
+			if stats.IsRunning {
+				isRunning.Observe(ctx, 1)
+			} else {
+				isRunning.Observe(ctx, 0)
+			}
+		},
+	)
+
+	if err != nil {
+		return fmt.Errorf("regestering metrics callback: %w", err)
```
Wondertan, line 100:
```suggestion
		return fmt.Errorf("das: regestering metrics callback: %w", err)
```

das/daser.go
```diff
@@ -89,6 +89,10 @@ func (d *DASer) Start(ctx context.Context) error {
 		return err
 	}
 
+	if err = d.sampler.initMetrics(); err != nil {
+		return fmt.Errorf("init sampling coordinator metrics: %w", err)
```
Wondertan, line 93:
```suggestion
		return fmt.Errorf("das: init sampling coordinator metrics: %w", err)
```

das/daser.go
```diff
@@ -89,6 +89,10 @@ func (d *DASer) Start(ctx context.Context) error {
...
+	}
```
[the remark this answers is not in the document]
Wondertan: The `m == nil` check is also complimentary to this

das/metrics.go
```diff
@@ -0,0 +1,132 @@
...
+	"go.opentelemetry.io/otel/metric/instrument/syncint64"
+
+	"github.com/celestiaorg/celestia-node/header"
+)
+
+var (
+	meter = global.MeterProvider().Meter("das")
+)
+
+type metrics struct {
```
distractedm1nd, line 22: Can we add a "last sampled timestamp" for #1040?

das/metrics.go
```diff
@@ -0,0 +1,132 @@
...
+	sampled    syncint64.Counter
+	sampleTime syncfloat64.Histogram
+
+	newHead syncint64.Counter
+
+	busyWorkers        asyncint64.Gauge
+	isRunning          asyncint64.Gauge
```
distractedm1nd, line 29: I'm wondering if we should keep `isRunning`- would it be clear from the other stats?

das/metrics.go
```diff
@@ -0,0 +1,132 @@
...
+
+func (sc *samplingCoordinator) initMetrics() error {
+	sampled, err := meter.SyncInt64().Counter("das_sampled_headers_count",
+		instrument.WithDescription("sampled headers counter"))
+	if err != nil {
+		return err
+	}
+
+	sampleTimeHist, err := meter.SyncFloat64().Histogram("das_sample_time_hist",
+		instrument.WithDescription("header sampling time"))
```
distractedm1nd, line 42: I think I would take `hist` out of the name of the meter, and then maybe have the description as "duration of sampling a single header"

das/metrics.go
```diff
@@ -0,0 +1,132 @@
...
+
+	busyWorkers        asyncint64.Gauge
+	isRunning          asyncint64.Gauge
+	headHeight         asyncint64.Gauge
+	headOfSampledChain asyncint64.Gauge
+}
+
+func (sc *samplingCoordinator) initMetrics() error {
+	sampled, err := meter.SyncInt64().Counter("das_sampled_headers_count",
+		instrument.WithDescription("sampled headers counter"))
```
distractedm1nd, line 36:
```suggestion
		instrument.WithDescription("total count of sampled headers"))
```

das/metrics.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	headHeight, err := meter.AsyncInt64().Gauge("das_network_head_height",
+		instrument.WithDescription("most recent network head height"))
+	if err != nil {
+		return err
+	}
+
+	headOfSampledChain, err := meter.AsyncInt64().Gauge("das_head_of_sampled_chain",
+		instrument.WithDescription("all headers before das_head_of_sampled_chain were successfully sampled"))
```
distractedm1nd, line 72:
```suggestion
	headOfSampledChain, err := meter.AsyncInt64().Gauge("das_sampled_chain_head",
		instrument.WithDescription("height of the sampled chain - all previous headers have been successfully sampled"))
```

das/metrics.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	isRunning, err := meter.AsyncInt64().Gauge("das_is_running",
+		instrument.WithDescription("indicates whether das is running"))
+	if err != nil {
+		return err
+	}
+
+	headHeight, err := meter.AsyncInt64().Gauge("das_network_head_height",
+		instrument.WithDescription("most recent network head height"))
```
distractedm1nd, line 66:
```suggestion
	headHeight, err := meter.AsyncInt64().Gauge("das_network_head",
		instrument.WithDescription("most recent network head"))
```

das/metrics.go
```diff
@@ -0,0 +1,132 @@
...
+type metrics struct {
+	sampled    syncint64.Counter
+	sampleTime syncfloat64.Histogram
+
+	newHead syncint64.Counter
+
+	busyWorkers        asyncint64.Gauge
+	isRunning          asyncint64.Gauge
+	headHeight         asyncint64.Gauge
+	headOfSampledChain asyncint64.Gauge
```
distractedm1nd, line 31: I like this naming better (and this is linked to my comments below) but up to you
```suggestion
	networkHead        asyncint64.Gauge
	sampledChainHead   asyncint64.Gauge
```

das/worker.go
```diff
@@ -54,6 +56,7 @@ func (w *worker) run(
 			break
 		}
 		w.setResult(curr, err)
+		metrics.observeSample(ctx, h, time.Since(start), err)
```
distractedm1nd, line 59: Hmmm... this looks a lot like synchronous metrics

das/metrics.go
distractedm1nd: registering*

das/daser.go
```diff
@@ -89,6 +89,10 @@ func (d *DASer) Start(ctx context.Context) error {
 		return err
 	}
 
+	if err = d.sampler.initMetrics(); err != nil {
```
distractedm1nd, line 92: Initializing the metrics shouldn't be done in `Start` if possible - we want to be able to pass a function to the DI when the metrics flag is enabled, so something like `MontitorDASer` in `metrics.go`  would work. Although I see that the challenge here is that getting the stats from the coordinator is private

distractedm1nd (commented):
I mostly just have naming suggestions, looks good! I really like the pattern of having a metrics struct, to not contaminate the main service struct. Going to try to incorporate this elsewhere

das/daser.go
walldiss: I'm not sure how it is possible to do, considering there are `sync` metrics that are tighten to DASer code and can't be easily disabled. To disable metrics we can just use  dummy metrics collector.
And I really like metrics being `internal` rather than `external` part of DASer. It is so much easier to observe this way.

das/metrics.go
walldiss: Thank you!.  This one meant to be **not** a total count, but I struggled to come up with better naming. 
Explanation: This one is tricky, because technically counter is not a total sum, it could do counts over duplicates and will increase upon retries. So counter value could be higher than actual amount of sampled headers.

das/metrics.go
walldiss: It will be used as a key for metrics index. Every error string would become a value in index and create a mess and slow down prometheus. I think it could be done by incorporating custom error types that has method to be encoded into prometheus readable values. Like http codes, but not full error.

das/worker.go
```diff
@@ -38,11 +38,13 @@ func (w *worker) run(
 	ctx context.Context,
 	getter header.Getter,
 	sample sampleFn,
+	metrics *metrics,
 	resultCh chan<- result) {
 	jobStart := time.Now()
 	log.Debugw("start sampling worker", "from", w.state.From, "to", w.state.To)
 
 	for curr := w.state.From; curr <= w.state.To; curr++ {
+		start := time.Now()
```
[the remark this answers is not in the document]
walldiss: Good point, thank you.

das/worker.go
walldiss: It is.

das/daser.go
walldiss: I think @Wondertan suggestion below to do `m == nil` check is the cleanest one and will implement it

distractedm1nd: @Wondertan how will this look in DI?

das/metrics.go
Wondertan: Can this be like an additional string attribute?

das/daser.go
Wondertan: This is the downside of synchronous metrics, and we cannot decouple those from the code. In the DASer case they are kinda justified and the simplest way to solve this would be through the DASer config/options that, besides configuring concurrency params and etc., would also tell DASer whether metrics should be enabled or not.

das/metrics.go
walldiss: I think if variety of errors will be high (it is not restricted right now) the prometheus will suffer. I've experience with mapping errors into error_codes and observing later in metrics. What could be the usecase if it is plain string in attribute?

Wondertan: I wonder if seeing errors is helpful for the metrics system, but I've never actually played with those deep enough. I trust your reasoning here about the lack of real use case and if an error happens, one can just query the rpc(and later cli) to check if there are any errors.

das/metrics.go
walldiss: Thank you, for pointing for issue! I've added new metric for it.

das/metrics.go
walldiss: you are right, any async metric submission is essentially "Hey, I'm running!" msg. Removed redundant metric, thank you!

das/metrics.go
walldiss: Maybe it is just me, but when there are 100+ metrics in prometheus with all sorts of types having little `_hist` or `_counter` suffix helps to navigate easier. So it is mostly me hating prometheus query interface. I think we can discuss  how to name metrics so it is consistent across different pkgs.

For this particular one if I saw `das_sample_time`, I would expect a `Gauge`. This could be a personal habbit.

das/metrics.go
distractedm1nd: Ahhh, I see, makes sense. I would just leave it then, I also cannot think of a better name

das/daser.go
distractedm1nd: can't we export an `InitMetrics` that just calls this? And then in the DI we can do fx.Invoke that receives the daser? Like

```go
func (d *DASer) InitMetrics() error {
    if err = d.sampler.initMetrics(); err != nil {
        return fmt.Errorf("das: init sampling coordinator metrics: %w", err)
    }
    return nil
}
```

and then


```go
fx.Invoke(func(d *DASer) err {
    return d.InitMetrics()
})
```

distractedm1nd (dismissed):
Looks great, excited for this

das/daser.go
Wondertan: @distractedm1nd, thanks. This approach is good and is also aligned with our `WithMetrics` option for the node. Not sure rn what is better here

Wondertan (approved):

Wondertan (changes requested):
Actually, wait, we need a way to make metrics optionable before we merge

das/daser.go
walldiss: @distractedm1nd Thank you! I really like this idea and incorporated this approach in new commit.

distractedm1nd: Looks good now

nodebuilder/daser/opts.go
```diff
@@ -0,0 +1,15 @@
+package daser
+
+import (
+	"github.com/celestiaorg/celestia-node/das"
+	"go.uber.org/fx"
+)
+
+func withMetrics(enable bool) fx.Option {
```
distractedm1nd, line 9: Can you export this function please?

nodebuilder/daser/module.go
```diff
@@ -28,6 +28,7 @@ func Module(tp node.Type) fx.Option {
 					return das.Stop(ctx)
 				}),
 			)),
+			withMetrics(true), // TODO: make configurable when daser gets its own cfg
```
distractedm1nd, line 31: Remove this line, should be disabled by default, and I will add to `nodebuilder/settings.go` once merged. The only reason this is difficult in this PR is because the node type isn't passed to `settings.go:WithMetrics` in main yet

distractedm1nd: #1083 is an example of how it will look in the DI (I will be removing my daser things from there once this is merged)

distractedm1nd (changes requested):

nodebuilder/daser/opts.go
distractedm1nd: and you can remove the `enable` param and empty return, since it won't be in `daser/module.go` - this `WithMetrics` will only be called if metrics are enabled

walldiss: I wanted to pass bool param from config in this function. What is the benefit of making it exportable?

nodebuilder/daser/module.go
walldiss: We can derive that the fact that metrics is on by checking if metrics endpoint is passed. Why would you prefer to move it outside of DASer pkg if we can pass in metricsEnabled flag with daser config instead?

walldiss: I got an idea. Sorry, thinking slow today. Moved option to `nodebuilder/settings.go` so it is consistent with other pkgs.

Wondertan (approved):

distractedm1nd (approved):

distractedm1nd:
For future reference I think both @Wondertan  and @renaynay need to review before you can merge, right guys?

renaynay:
ideally when they're not one-line fixes. I didn't get to review this PR.

PR #1131, opened 2022-09-19
refactor(service/state): Use DefaultProofRuntime instead of ibc-go `VerifyMembership` endpoint

renaynay:
Removes ibc-go as a dependency of node and uses cosmos-sdk DefaultProofRuntime instead.


Wondertan (dismissed):
Perfect! Trust that you've tested it

Linter fails though

Wondertan (approved):

liamsi (approved):
🏎️

PR #1135, opened 2022-09-20
feat!(header/p2p): extend header response with status code

### LINKED ISSUE #906: header: Error codes to respond with meaningful information if something wrong happens on the responding side, so clients can act accordingly

vgonkivs:
Resolves #906

Based on #1038 .


header/p2p/pb/extended_header_request.proto
```diff
@@ -9,3 +9,8 @@ message ExtendedHeaderRequest {
   }
   uint64 amount = 3;
 }
+
+message ExtendedHeaderResponse {
+  bytes body = 1;
+  uint32 statusCode = 2;
```
Wondertan, line 15: We should make statusCode as enum defined in proto. This way, the protocol definition would be self-sufficient. The convertStatusCodeToError can just rely on the PB definition

header/p2p/exchange_test.go
```diff
@@ -85,13 +84,12 @@ func TestExchange_RequestByHash(t *testing.T) {
...
-	resp := new(header_pb.ExtendedHeader)
+	resp := new(p2p_pb.ExtendedHeaderResponse)
 	_, err = serde.Read(stream, resp)
 	require.NoError(t, err)
-	// compare
-	eh, err := header.ProtoToExtendedHeader(resp)
+	eh, err := header.UnmarshalExtendedHeader(resp.Body)
 	require.NoError(t, err)
-
+	// compare
```
Wondertan, line 92: ?

header/p2p/response.go
```diff
@@ -0,0 +1,33 @@
...
+	}
+
+	out, err := UnmarshalExtendedHeaderResponse(data)
+	if err != nil {
+		return err
+	}
+
+	*ehr = *out
+	return nil
+}
```
Wondertan, line 33: I feel like we don't need this. Nor imported publicly nor privately. I know that we do that for the Request, but we should not. It doesn't give us anything besides additional allocations, so can you pls make it work directly with PB structs and stop using Request definition in the subsequent issue/PR

vgonkivs: Yes, you are right. It was my first thought, but then I decided to make it similar to `request`.

header/p2p/pb/extended_header_request.proto
```diff
@@ -3,7 +3,19 @@ syntax = "proto3";
...
+  oneof data {
+    uint64 origin = 1;
+    bytes hash = 2;
+  }
   uint64 amount = 3;
 }
+
+enum StatusCode {
+  StatusOK = 0;
+  StatusNotFound = 1;
```
Wondertan, line 15:
```suggestion
  OK = 0;
  NOT_FOUND = 1;
```

header/p2p/response.go
Wondertan: Pls, let's add to the epic the removal of Request structure

vgonkivs: https://github.com/celestiaorg/celestia-node/pull/1141

header/p2p/server.go
```diff
@@ -52,19 +54,66 @@ func (serv *ExchangeServer) Stop(context.Context) error {
...
+	if err != nil {
+		switch err {
+		case header.ErrNotFound:
+			headers = make([]*header.ExtendedHeader, 1)
+			code = p2p_pb.StatusCode_NOT_FOUND
+		default:
+			stream.Reset() //nolint:errcheck
+			return
+		}
+	}
```
walldiss, line 95: Looks like this part is handling error from `stream.CloseRead()` happening at line 69? Maybe it is possible to combine it with same if err != nil{} check? 

Also optionally `switch` with 1 case could be reduced by 
```
if !errorrs.Is(header.ErrNotFound){ return fast}
```

vgonkivs: > Looks like this part is handling error from stream.CloseRead()
Not exactly. It handles error from `handleRequestByHash/handleRequest`

> Also optionally switch with 1 case could be reduced by
there will be one more case in subsuquent pr

header/p2p/server.go
```diff
@@ -2,15 +2,17 @@ package p2p
 
 import (
 	"context"
+	"time"
 
 	"github.com/libp2p/go-libp2p-core/host"
 	"github.com/libp2p/go-libp2p-core/network"
```
distractedm1nd, line 9:
```suggestion
```

header/p2p/server.go
```diff
@@ -75,69 +124,37 @@ func (serv *ExchangeServer) requestHandler(stream network.Stream) {
 
 // handleRequestByHash returns the ExtendedHeader at the given hash
 // if it exists.
-func (serv *ExchangeServer) handleRequestByHash(hash []byte, stream network.Stream) {
+func (serv *ExchangeServer) handleRequestByHash(hash []byte) ([]*header.ExtendedHeader, error) {
```
distractedm1nd, line 127: This subroutine should return a single extended header, and it can be put into an array easily above

header/p2p/exchange.go
```diff
@@ -70,7 +78,7 @@ func (ex *Exchange) GetByHeight(ctx context.Context, height uint64) (*header.Ext
 	}
 	// create request
 	req := &p2p_pb.ExtendedHeaderRequest{
-		Origin: height,
+		Data:   &p2p_pb.ExtendedHeaderRequest_Origin{Origin: height},
```
distractedm1nd, line 81: idk if its possible the way I'm imagining it, and its up to you, but I think it would be really nice if these types could be aliased at the top of the file to something short and sweet

distractedm1nd (commented):
Looks good

header/p2p/exchange.go
vgonkivs: Well, it's possible. but I'm not sure that this is a right pattern to make aliases for generated structures.

header/p2p/server.go
vgonkivs: True. But I wanted to avoid additional logic inside switch-case statement beside assigning. IMO, it looks clearer this way.

walldiss (dismissed):
Looks good and clean.

distractedm1nd (approved):

header/p2p/server.go
```diff
@@ -52,19 +53,66 @@ func (serv *ExchangeServer) Stop(context.Context) error {
...
+	default:
+		log.Error("server: invalid data type received")
+		stream.Reset() //nolint:errcheck
+		return
+	}
+	code := p2p_pb.StatusCode_OK
+	if err != nil {
+		switch err {
+		case header.ErrNotFound:
+			headers = make([]*header.ExtendedHeader, 1)
```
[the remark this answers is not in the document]
renaynay: So actually in the case of `GetRangeByHeight` which is what is used for such a range, we actually fail out if we *do not* have the last header immediately, so it's not possible to even return the range we *do* have.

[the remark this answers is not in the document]
renaynay: So technically then, this code is correct, but my question is should this be the desired behaviour? cc @Wondertan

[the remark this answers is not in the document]
renaynay: parked in an issue: https://github.com/celestiaorg/celestia-node/issues/1163

renaynay (approved):

renaynay (approved):

distractedm1nd (approved):

header/p2p/pb/extended_header_request.proto
```diff
@@ -3,7 +3,20 @@ syntax = "proto3";
...
+    bytes hash = 2;
+  }
   uint64 amount = 3;
 }
+
+enum StatusCode {
+  OK = 0;
+  NOT_FOUND = 1;
+  LIMIT_EXCEEDED = 2;
+};
```
walldiss, line 17: It is a good practice for proto error enum type is keeping zero value invalid. So if caller forget to set value and it becomes zero-value, it will not collide with actual status code. Example:
```
enum StatusCode {
    INVALID = 0;
    OK = 1;
    NOT_FOUND = 2;
    LIMIT_EXCEEDED = 3;
}
```

Wondertan (approved):
LGTM so far. Vlads recommendation is good tofollow

header/p2p/pb/extended_header_request.proto
vgonkivs: Thank you. Done!

header/p2p/exchange_test.go
```diff
@@ -170,6 +211,9 @@ func (m *mockStore) GetByHeight(ctx context.Context, height uint64) (*header.Ext
 
 func (m *mockStore) GetRangeByHeight(ctx context.Context, from, to uint64) ([]*header.ExtendedHeader, error) {
 	headers := make([]*header.ExtendedHeader, to-from)
+	if to-1 > m.Height() {
```
walldiss, line 214: Could you please add a comment why compare to `to-1`?

Wondertan (approved):
LGTM

header/p2p/server.go
```diff
@@ -69,17 +68,52 @@ func (serv *ExchangeServer) requestHandler(stream network.Stream) {
...
+		return
+	}
+
+	// write all headers to stream
+	for _, h := range headers {
+		if err := stream.SetWriteDeadline(time.Now().Add(writeDeadline)); err != nil {
+			log.Debugf("error setting deadline: %s", err)
+		}
+		var bin []byte
+		if h != nil {
```
walldiss, line 102: I'm not familiar with this code, so probably a silly question: What is the case when we send empty response with `statusOk` , e.g. `h==nil` situation? How is it different from `header.ErrNotFound`? Would be nice to have a comment for it.

walldiss: Another question I have whether we should use return an error in situation where is partial data is returned. Maybe add another `statusCodePartial` to keep status explicit?

vgonkivs: Good question. Could you please add it [here](https://github.com/celestiaorg/celestia-node/issues/1163) ?

renaynay (changes requested):

renaynay (approved):

header/p2p/server.go
```diff
@@ -69,17 +68,52 @@ func (serv *ExchangeServer) requestHandler(stream network.Stream) {
...
+	switch err {
+	case nil:
+		code = p2p_pb.StatusCode_OK
+	case header.ErrNotFound:
+		headers = make([]*header.ExtendedHeader, 1)
+		code = p2p_pb.StatusCode_NOT_FOUND
+	default:
+		stream.Reset() //nolint:errcheck
+		return
+	}
```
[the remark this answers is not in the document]
Wondertan: It's useless.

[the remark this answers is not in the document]
vgonkivs: Agree, checking error on `stream.Reset` is redundant in our case.

header/p2p/exchange.go
```diff
@@ -164,7 +163,12 @@ func (ex *Exchange) performRequest(
 			stream.Reset() //nolint:errcheck
 			return nil, err
 		}
-		header, err := header.ProtoToExtendedHeader(resp)
+
+		if err = convertStatusCodeToError(resp.StatusCode); err != nil {
+			stream.Reset() //nolint:errcheck
```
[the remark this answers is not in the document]
vgonkivs: It's redundant. It wll be just +1 line in logs.

renaynay (approved):

walldiss (approved):

PR #1146, opened 2022-09-21
Migrated build logic to `cmd/info.go`

### LINKED ISSUE #347: build: Move in build info currently stored next verion cmd
Let's move all the build-related stuff to a related `build` package. Some of the build related info(filled with ldflags) is stored in `cmd/celestia/version.go` and should be moved to `build/info.go`

0xJepsen:
closes #347 

Would it make sense to turn it into an info flag?
Or would it make more sense to add as a command in`init.go`


renaynay (commented):
Hi @0xJepsen thank you for your contribution! 

Two things:
* While the issue is a bit unclear about this (and it's a bit old, so sorry for that!), there is no compelling reason to have 2 different commands for outputting version info + build info, so let's keep them together under `versionCmd` and remove the `buildInfo` command. 
* Let's still move the version command out of the `cmd/celestia` into the `cmd` package

renaynay:
Also, the version command should still be available on the root command, not as a subcommand of `light | full | bridge` please 🙏🏻

PR #1153, opened 2022-09-22
feat(header/p2p): implement ErrHeadersLimitExceeded error

### LINKED ISSUE #191: service/header: Implement throttling on how many `ExtendedHeader`s can be requested/sent at once.
We need to figure out the max packet size we are willing to send across the network for *any* given message, not just `ExtendedHeader` - related packets. If we are to send batched headers eventually, then we need to measure the packet size so that it does not increase whatever max packet size limit we send. 

We could do this either on the request side (so only able to request max 100, for example, headers at a time), or we can just break up the `[]*ExtendedHeader` packet so that it sends maximum 50 headers in a packet at a time. 

These are obviously arbitrary numbers and will require testing to see what kind of max packet size is most efficient for our network, but we need to define these metrics post Devnet.

vgonkivs:
Resolves #191

### DIFF
--- a/header/interface.go
+++ b/header/interface.go
@@ -60,12 +60,18 @@ type Exchange interface {
 	Getter
 }
 
+// RequestSize defines the max amount of headers that can be requested/handled at once.
+var RequestSize uint64 = 512
+
 var (
 	// ErrNotFound is returned when there is no requested header.
 	ErrNotFound = errors.New("header: not found")
 
 	// ErrNoHead is returned when Store is empty (does not contain any known header).
 	ErrNoHead = fmt.Errorf("header/store: no chain head")
+
+	// ErrHeadersLimitExceeded is returned when ExchangeServer receives header request for more then RequestSize headers.
+	ErrHeadersLimitExceeded = errors.New("header/p2p: header limit per 1 request exceeded")
 )
 
 // ErrNonAdjacent is returned when Store is appended with a header not adjacent to the stored head.
--- a/header/interface.go
+++ b/header/interface.go
@@ -60,9 +60,6 @@ type Exchange interface {
 	Getter
 }
 
-// RequestSize defines the max amount of headers that can be requested/handled at once.
-var RequestSize uint64 = 512
-
 var (
 	// ErrNotFound is returned when there is no requested header.
 	ErrNotFound = errors.New("header: not found")
@@ -70,7 +67,8 @@ var (
 	// ErrNoHead is returned when Store is empty (does not contain any known header).
 	ErrNoHead = fmt.Errorf("header/store: no chain head")
 
-	// ErrHeadersLimitExceeded is returned when ExchangeServer receives header request for more then RequestSize headers.
+	// ErrHeadersLimitExceeded is returned when ExchangeServer receives header request for more
+	// than maxRequestSize headers.
 	ErrHeadersLimitExceeded = errors.New("header/p2p: header limit per 1 request exceeded")
 )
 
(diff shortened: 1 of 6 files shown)


header/interface.go
```diff
@@ -60,12 +60,18 @@ type Exchange interface {
 	Getter
 }
 
+// RequestSize defines tha max amount of headers that can be requested/handled at once.
```
distractedm1nd, line 63:
```suggestion
// RequestSize defines the max amount of headers that can be requested/handled at once.
```

header/interface.go
```diff
@@ -60,12 +60,18 @@ type Exchange interface {
...
+
 var (
 	// ErrNotFound is returned when there is no requested header.
 	ErrNotFound = errors.New("header: not found")
 
 	// ErrNoHead is returned when Store is empty (does not contain any known header).
 	ErrNoHead = fmt.Errorf("header/store: no chain head")
+
+	// ErrHeadersLimitExceeded is returned when ExchangeServer receives header request for more then RequestSize headers.
+	ErrHeadersLimitExceeded = fmt.Errorf("header/p2p: header limit per 1 request exceeded")
```
distractedm1nd, line 74: Why are these errors `fmt.Errorf` if nothing is formatted? I think `errors.New` is what we use elsewhere

header/p2p/exchange_test.go
```diff
@@ -77,24 +110,57 @@ func TestExchange_RequestByHash(t *testing.T) {
...
 	// compare
-	eh, err := header.ProtoToExtendedHeader(resp)
+	eh, err := header.UnmarshalExtendedHeader(resp.Body)
 	require.NoError(t, err)
 
 	assert.Equal(t, store.headers[reqHeight].Height, eh.Height)
 	assert.Equal(t, store.headers[reqHeight].Hash(), eh.Hash())
 }
 
+// TestExchange_RequestByHash tests that the Exchange instance can
```
distractedm1nd, line 131:
```suggestion
// TestExchange_RequestByHashFails tests that the Exchange instance can
```

header/sync/sync_test.go
```diff
@@ -20,7 +20,7 @@ var blockTime = 30 * time.Second
 func TestSyncSimpleRequestingHead(t *testing.T) {
 	// this way we force local head of Syncer to expire, so it requests a new one from trusted peer
 	header.TrustingPeriod = time.Microsecond
-	requestSize = 13 // just some random number
+	header.RequestSize = 13 // just some random number
```
distractedm1nd, line 23: we could fuzz this maybe, just an idea

header/p2p/exchange.go
```diff
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"math/rand"
+	"time"
 
 	logging "github.com/ipfs/go-log/v2"
```
distractedm1nd, line 11:
```suggestion
```

distractedm1nd (changes requested):

header/sync/sync_test.go
vgonkivs: not in scope of this PR I thing

renaynay: But in general would be nice to add to testing ideas

renaynay (dismissed):

header/interface.go
```diff
@@ -60,12 +60,18 @@ type Exchange interface {
 	Getter
 }
 
+// RequestSize defines the max amount of headers that can be requested/handled at once.
+var RequestSize uint64 = 512
```
Wondertan, line 64: I see the motivation here. To make it reusable from the HeaderEx and Syncer but we should not pull protocol-level specifics to the interfaces. Instead, it should be `maxRequestSize` const in the HeaderEx and the `requestSize` global var in the Syncer the way it was. That is, these parameters are two separate things. One is the limit, the other is the size.

renaynay: agree

Wondertan (approved):
LGTM

renaynay (approved):
👍🏻

Issue #1155, opened 2022-09-22
Missing dased past header log


Bidon15:
### Celestia Node version

8bce8d023f9d0a1929e56885e439655717aea4e4

### OS

alpine-docker

### Install tools

Tested using testground's PR https://github.com/celestiaorg/test-infra/pull/68

### Steps to reproduce it

1. Install testground
2. Executed command
```bash
testground run composition -f compositions/local-docker/003-full-sync-past-12.toml --wait > log2.log
```

### Expected result

logs of past das appear 

### Actual result

no logs

### Relevant log output

```shell
Sep 21 11:15:13.967092	INFO	825.1331s      OTHER << fulls[002] (ee8e53) >> 2022-09-21T11:15:13.966Z	INFO	das	das/daser.go:282	sampled header	{"height": 4, "hash": "2A48FE6DEB2B64ECF143E6A0245DE59C8BD6C92F459C17ECDD85F9712BC9F543", "square width": 128, "finished (s)": 7.842592032}
```

### Notes

Worked on version 74de636b3f4757dd149fe00bf57db898ecacaa3d

PR #1157, opened 2022-09-22
das: add extra debug logs for sampled headers

walldiss:
Adds some logs for easier debugging
Resolves https://github.com/celestiaorg/celestia-node/issues/1155


Bidon15 (dismissed):
Smooth loggerator 
Thanks for the work 🙏

das/daser.go
```diff
@@ -168,6 +169,8 @@ func (d *DASer) sample(ctx context.Context, h *header.ExtendedHeader) error {
 			"square width", len(h.DAH.RowsRoots), "data root", h.DAH.Hash(), "err", err)
 		return err
 	}
+	log.Debugw("sampled header", "height", h.Height, "hash", h.Hash(),
+		"square width", len(h.DAH.RowsRoots), "data root", h.DAH.Hash(), "finished (s)", time.Since(start))
```
Wondertan, line 173: We might want to move it to the worker loop because metrics introduce the same time tracking, and doing it twice is wasteful.

walldiss: Good point. updated

das/worker.go
```diff
@@ -43,10 +43,18 @@ func (w *worker) run(
...
 		h, err := getter.GetByHeight(ctx, curr)
 		if err == nil {
-			err = sample(ctx, h)
+			log.Debugw("got header from header store", "height", h.Height, "hash", h.Hash(),
+				"square width", len(h.DAH.RowsRoots), "data root", h.DAH.Hash(), "finished (s)", time.Since(startGet))
+
+			startSample := time.Now()
+			if err = sample(ctx, h); err == nil {
+				log.Debugw("sampled header", "height", h.Height, "hash", h.Hash(),
+					"square width", len(h.DAH.RowsRoots), "data root", h.DAH.Hash(), "finished (s)", time.Since(startSample))
```
Wondertan, line 56: One last thing. Let's make it more readable by following Uber's Go guidelines and [reduce nesting](https://github.com/uber-go/guide/blob/master/style.md#reduce-nesting).

das/daser_test.go
```diff
@@ -297,6 +299,7 @@ func (m getterStub) Head(context.Context) (*header.ExtendedHeader, error) {
 
 func (m getterStub) GetByHeight(_ context.Context, height uint64) (*header.ExtendedHeader, error) {
 	return &header.ExtendedHeader{
+		Commit:    &types.Commit{},
```
renaynay, line 302: What was the issue that this solves?

das/worker.go
renaynay: +1

das/worker.go
```diff
@@ -43,17 +43,30 @@ func (w *worker) run(
...
+			"square width", len(h.DAH.RowsRoots), "data root", h.DAH.Hash(), "finished (s)", time.Since(startGet))
 
+		startSample := time.Now()
+		err = sample(ctx, h)
 		if errors.Is(err, context.Canceled) {
 			// sampling worker will resume upon restart
 			break
 		}
 		w.setResult(curr, err)
+
```
Wondertan, line 67: I assume one `setResult` with nil is always required, or should we also check `if err != nil` in here?

Wondertan (approved):

Bidon15 (dismissed):

das/worker.go
walldiss: No need, error should be propagated into setResults.

renaynay (dismissed):

Bidon15 (approved):
🚢 🚢 🚢 🚢 🚢

PR #1147, opened 2022-09-21
chore: upgrade to celestia-app v0.7.0

rootulp:
Closes https://github.com/celestiaorg/celestia-node/issues/1142
Opens https://github.com/celestiaorg/celestia-node/issues/1231

PR #1160, opened 2022-09-23
refactor(swamp): use celestia-app instead of the kvstore

### LINKED ISSUE #628: swamp: Integration tests with celestia-app instead of kvstore
### Implementation ideas

Use the app in swamp tests: https://github.com/celestiaorg/celestia-node/pull/626#issuecomment-1100609951

evan-forbes:
refactors the swamp test to use celestia-app and a normal tendermint node instead of the kvstore and the rpc test tendermint node.

close #628


evan-forbes:
the current refactor in this PR is using `testutil/network` from the cosmos-sdk, which could be removed or modified sometime soon https://github.com/cosmos/cosmos-sdk/issues/12527.

fwiw, I could not find yet find the reason why the cosmos-sdk's `testutil/network` is incompatible with sub 1 second block times. For whatever reason, it continues to increase proposal time by one second per block if the timeout commit is set below 1s. If the timeout commit is set greater than 1s, then the times are as expected. This does not occur with a normal one node network using celestia-app, only when using the cosmos-sdk's `testutil/network`

there are a few weird things about this bug:
- the block times decrease irl despite the times increasing by 1 second each time
- when the bug occurs, the precision of the block times are identical.
```
bug not occuring
timeout-commit = 1000ms
2022-09-26 23:06:00.547264861 +0000 UTC
2022-09-26 23:06:01.580626296 +0000 UTC
2022-09-26 23:06:02.614770553 +0000 UTC
2022-09-26 23:06:03.648607489 +0000 UTC
2022-09-26 23:06:04.684149147 +0000 UTC
```
```
bug occuring
timeout-commit = 950ms
2022-09-26 23:06:39.503282545 +0000 UTC
2022-09-26 23:06:40.503282545 +0000 UTC
2022-09-26 23:06:41.503282545 +0000 UTC
2022-09-26 23:06:42.503282545 +0000 UTC
```
```
bug occuring
timeout-commit = 100ms
2022-09-26 23:10:10.135204764 +0000 UTC
2022-09-26 23:10:11.135204764 +0000 UTC
2022-09-26 23:10:12.135204764 +0000 UTC
2022-09-26 23:10:13.135204764 +0000 UTC
```
Still not sure if we should fix this, or go with our own proprietary option that we have already started on.

evan-forbes:
had some time to work on this, and the last bug ended up being that we needed to change a the consensus parameter `TimeIotaMs`, to configure the network to not have a minimum amount of time required to produce a block. By default this is 1000ms, and if the block time is shorter than this, then tendermint will just continually add the minimum time to each commit signature, which is used to determine the block time. 

There was another issue where the cosmos-sdk will override this value while collecting the gentxs if using the this [genutil func](https://github.com/celestiaorg/cosmos-sdk/blob/482d8986c7842b93f40239a210fac8b3975db39a/x/genutil/utils.go#L31-L49).

both of those are fixed and now we have block using 100ms. This did fix a few swamp tests, but I'm still seeing some errors the last test that it would be useful to get a second pair of eyes on. 

```
TestFullReconstructFromLights
2022-09-29T23:40:46.430-0500    ERROR   header/sync     sync/sync_head.go:122   appending network header        {"height": 680, "hash": "0C9556A3C326B1D2C92E4B6A260608AF79A8D68745E91539D8032ABD44BCBDD1", "err": "header/store: no chain head"}
2022-09-29T23:40:46.430-0500    ERROR   header/sync     sync/sync_head.go:151   getting subjective head during validation       {"err": "header/store: no chain head"}
...
2022-09-29T23:40:46.473-0500    ERROR   header/sync     sync/sync.go:190        syncing headers {"from": 630, "to": 680, "err": "header/store: non-adjacent: head 671, attempted 631"}
2022-09-29T23:40:46.474-0500    ERROR   header/sync     sync/sync.go:190        syncing headers {"from": 475, "to": 680, "err": "header/store: non-adjacent: head 647, attempted 476"}
panic: runtime error: slice bounds out of range [:2] with capacity 1

goroutine 25551 [running]:
github.com/celestiaorg/celestia-node/header/sync.(*headerRange).Before(0xc028069c00, 0x29d)
        /home/evan/go/src/github.com/celestiaorg/celestia-node/header/sync/ranges.go:147 +0x165
github.com/celestiaorg/celestia-node/header/sync.(*Syncer).findHeaders(0xc01f246000, {0x3315538, 0xc00ea56140}, 0x227, 0x29d)
        /home/evan/go/src/github.com/celestiaorg/celestia-node/header/sync/sync.go:272 +0x26a
```

Wondertan:
@evan-forbes, I will checkout the branch and try to reproduce this

evan-forbes:
thanks! this PR is using this [branch](https://github.com/celestiaorg/celestia-app/tree/evan/v0.7.0-rc-1-test-node/testutil/testnode) on the application to have a working version of the new testnode on v0.7.0-rc-1

evan-forbes:
update: I'll try and work on this tmrw/later this week. The branch of the app that it uses needs to be slightly updated to something closer to https://github.com/celestiaorg/celestia-app/pull/824.

Wondertan:
@evan-forbes, why do we need 10k of funded accounts in the `testnode` setup? It makes test execution for me unbelievably slow

Wondertan:
@evan-forbes, so the issue with `header: not found` happens because the App generates blocks with zero timeouts and is thus ultra-quick. This is something unobvious, but it is what it is.

Wondertan:
So I pushed fixes for the synchronization logic. Now the reconstruction Swamp test causes issues. It succeeds, but it takes much longer than it should. The reason is somewhere in PFDs and FillBlocks implementation.

Wondertan:
Ok, now the reconstruction tests work correctly, and the last one to fix are BEFP related

Wondertan:
Cool, so this migration uncovered a bug related to services shutdown on FraudProof arrival

Wondertan:
I've included one commit that fixes CI on the main to see that all the tests here reliably pass. It will remove it before merging.

Wondertan:
I also rebased it on the main and desperately trying to make CI green. Locally everything works fine!

evan-forbes:
I should have time today or tmrw to continue debugging this. Will likely end up trying quite a few temporary commits to fix the CI. I can create another branch if we need to try different things in parallel.

Wondertan:
Hey @evan-forbes. Thanks. I will also continue debugging. I feel like I am very close

Wondertan:
@evan-forbes, so the race I tried to fix in https://github.com/celestiaorg/celestia-app/pull/845 is now hitting this PR. Another reason to merge it ASAP 
image

Wondertan:
There is another one deep in the core. It was fixed by 0.35, but returned back due to a downgrade. It is rare but still happens from time to time. I actually saw it dozens of times since the first test in the swamp. It would be nice if the new test suite in app/core could fix it. 
image

Wondertan:
Making it ready for review for the team

evan-forbes:
CI took 1 re-run, but its passing

Wondertan:
> CI took 1 re-run, but its passing
@evan-forbes, what was it failing with on the first run?

vgonkivs (dismissed):

Wondertan:
I tried to rerun the tests and got this, even with the right tag for core included https://github.com/celestiaorg/celestia-core/pull/869#issuecomment-1273881766
image

walldiss (dismissed):

evan-forbes:
I believe the previous failure was just that the rpc port was already taken.

Damn, then using the mutex in core isn't as simple as I thought.

Wondertan:
@evan-forbes, are you ok with merging this with temporary tags?

evan-forbes:
yeah, definitely

Wondertan:
Then I would like to have a last look from @renaynay and @Bidon15 before merging

Wondertan:
Would like to have a last look from @renaynay and @Bidon15 before merging

Bidon15 (dismissed):

go.mod
```diff
@@ -63,6 +64,8 @@ require (
 	google.golang.org/grpc v1.49.0
 )
 
+require github.com/celestiaorg/celestia-app v0.0.0-00010101000000-000000000000
```
Bidon15, line 67: I hope this will be changed in https://github.com/celestiaorg/celestia-node/pull/1147

go.mod
```diff
@@ -271,18 +280,20 @@ require (
...
 	gopkg.in/tomb.v1 v1.0.0-20141024135613-dd632973f1e7 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	lukechampine.com/blake3 v1.1.7 // indirect
+	nhooyr.io/websocket v1.8.6 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
 
 replace (
+	github.com/celestiaorg/celestia-app => github.com/celestiaorg/celestia-app v0.7.0-rc-1.0.20220930201625-f8540b2aec19
```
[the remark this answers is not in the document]
Wondertan: I assume it will be removed later with the real update. But I don't really see a difference from just directly requiring the dep

go.mod
Wondertan: This actually can be removed already so I did it

Wondertan:
The last change here, I hope. I updated timeouts and tide up the go.mod

vgonkivs (dismissed):

Bidon15 (dismissed):

vgonkivs (approved):

Bidon15 (approved):

PR #1161, opened 2022-09-26
refactor(nodebuilder): Move parse methods and flag definitions to nodebuilder subpackages

### LINKED ISSUE #1126: refactor(nodebuilder): Move parse functions + flag defs to the respective modules
https://github.com/celestiaorg/celestia-node/pull/997#discussion_r970083829

Blocked by #997

distractedm1nd:
Closes #1126


vgonkivs (approved):

Wondertan (changes requested):
Actually, we should make one nice thing here. Every flag name is prepended with the name of a module, so let's extract module name strings we pass to fx.Module into constants and reuse them as categories for flag names.

distractedm1nd:
```golang
var (
	ipFlag   = moduleName + ".ip"
	rpcFlag  = moduleName + ".rpc.port"
	grpcFlag = moduleName + ".grpc.port"
)
```
Ok so I tried it out @Wondertan:

I don't think this is more readable, and then there is for example the "keyring.accname" in the state package. I'm not convinced extracting this into a const is better yet

nodebuilder/header/flags.go
```diff
@@ -27,16 +24,16 @@ func HeadersFlags() *flag.FlagSet {
 	return flags
 }
 
-// ParseHeadersFlags parses Header package flags from the given cmd and applies values to Env.
-func ParseHeadersFlags(ctx context.Context, cmd *cobra.Command, cfg *nodebuilder.Config) (context.Context, error) {
-	if ctx, err := ParseTrustedHashFlags(ctx, cmd, cfg); err != nil {
-		return ctx, err
+// ParseFlags parses Header package flags from the given cmd and saves them to the passed config.
```
renaynay, line 27:
```suggestion
// ParseFlags parses Header package flags from the given cmd and applies them to the passed config.
```

we don't necessarily *save* the modified conf, only on Init

nodebuilder/header/flags.go
```diff
@@ -50,28 +47,24 @@ func TrustedPeersFlags() *flag.FlagSet {
 	return flags
 }
 
-// ParseTrustedPeerFlags parses Header package flags from the given cmd and applies values to Env.
+// ParseTrustedPeerFlags parses Header package flags from the given cmd and saves them to the passed config.
```
renaynay, line 50: same thing here, just leave it as `applies`

renaynay (dismissed):

nodebuilder/header/flags.go
distractedm1nd: done

renaynay:
also generally, if we have ideas like this that get spawned from a PR, unless they're like one-line fixes, we should make a habit of sticking them in an issue and linking the PR inspiring it rather than increasing scope of PR

walldiss (approved):

Wondertan (approved):
> I don't think we should do this now as it is yet another breaking change

It is fine to group multiple breaking changes together.

> We can make an issue for it to do it down the line but i wouldn't hold this PR up for it and i think our current naming is okay.

Agreed

renaynay (approved):

PR #1168, opened 2022-09-26
refactor(params): Moving params to nodebuilder and adding config

### LINKED ISSUE #1076: fix(cmd/params): Repair encapsulation for network params
PR #1073 will temporarily break the encapsulation of the `params` package by exporting a setter of the default network.

A node option should be made for this, and the default values that were removed in 1073 should be restored once possible without convoluted logic. This means `NodeOptions` will need to take the node type and network, to be able to set the default values correctly.

This is important and will be completed as part of the refactoring efforts.

distractedm1nd:
Closes #1076 

Reviewers: Please test for yourself to verify it still works.

@Wondertan what do you think of this approach? 

In the style of #1161


nodebuilder/network/config.go
```diff
@@ -0,0 +1,15 @@
+package network
+
+type Config struct {
+	Network Network
```
Wondertan, line 4: We already store the network as part of the Store path. What will happen if someone changes the config to mamaki under `celestia-light-arabica`? Don't think we should allow changing the network in the config.

Wondertan (commented):
Generally, I like the idea of making `params` as a module or a part of a module What if we go one step further and make it part of the `node` module? `node.Network`, `node.Bootstrapper`, etc.  I don't think `network` deserves its own module, just for 3 types there, but could be a coherent addition to the node module.

nodebuilder/network/config.go
distractedm1nd: The network flags are parsed before creating the store path, see `flags_node.go`. So if they change it in the config, it affects the store path (as long as they don't pass a custom store, ofc)

renaynay (commented):
I'm not sure that i like params living inside `nodebuilder/node`

I think of `node` as administrative -- would you be opposed to a `nodebuilder/netparams` package instead?

renaynay:
These params could honestly even live inside of p2p module bc they are related to p2p.

distractedm1nd:
I wouldnt be opposed to that, I originally had them in their own package - @Wondertan what do you think?

nodebuilder/node/flags.go
```diff
@@ -0,0 +1,79 @@
...
+	return Network(parsedNetwork), nil
+}
+
+// ParseNetworkFromEnv tries to parse the network from the environment,
+// and returns either the parsed network or the build's default network
+func ParseNetworkFromEnv() (Network, error) {
+	var network Network
+	// check if custom network option set
+	// format: CELESTIA_CUSTOM=<netID>:<genesisHash>:<bootstrapPeerList>
+	if custom, ok := os.LookupEnv(EnvCustomNetwork); ok {
```
walldiss, line 49: Looks like it returns empty network if there is no env variable, but comment suggests there should be default one. Maybe there is some guarantee that env variable will persist?

Wondertan:
I agree that `p2p` is fine also, so let's do this.

nodebuilder/node/flags.go
```diff
@@ -0,0 +1,79 @@
...
+// ParseFlags tries to parse the network from the flags and environment,
+// and returns either the parsed network or the build's default network
+func ParseFlags(cmd *cobra.Command) (Network, error) {
+	parsedNetwork := cmd.Flag(nodeNetworkFlag).Value.String()
+	// no network set through the flags, so check if there is an override in the env
+	if parsedNetwork == "" {
+		return ParseNetworkFromEnv()
+	}
+	return Network(parsedNetwork), nil
+}
```
Wondertan, line 41: `ParseFlags` does not override the `DefaultNetwork`, meaning that flag won't affect our protocol IDs and breaks compatibility with Mamaki. We could still override as a short-term fix. The correct long-term fix here is to make protocol IDs not rely on a constant. Instead, each protocol like HeaderEx should have an option or config through which it can receive and prepend a chain ID in the constructor and not during global initialization.,

distractedm1nd: Ok, another reason why network might be helpful to be in the config no?

distractedm1nd: Or I can pass the network to the modules

Wondertan: Network to Modules 👍

distractedm1nd:
Some issues came up during rebasing, fixing

nodebuilder/node/flags.go
distractedm1nd: done btw

Wondertan (changes requested):
Requesting changes because our libs/pkgs should not import node modules. Instead, we should pass the network name as a protocol prefix, which should be an option and not the required param. 

Let's revert the change and solve the #1189 issue in the subsequent PR, as it is out of the scope. It's fine that the main will be broken for some time for Mamaki.

Wondertan (approved):
Awesome

nodebuilder/header/header.go
```diff
@@ -10,12 +10,12 @@ import (
 	"github.com/celestiaorg/celestia-node/header"
 	"github.com/celestiaorg/celestia-node/header/p2p"
 	"github.com/celestiaorg/celestia-node/header/store"
-	"github.com/celestiaorg/celestia-node/params"
+	network "github.com/celestiaorg/celestia-node/nodebuilder/p2p"
```
Wondertan, line 13: The overlapping of module pkgs and real libs will be a common thing. Should we agree on the pattern naming module imports something like `modp2p`? 

cc @distractedm1nd @renaynay

renaynay: Yes I think that's a good idea cc @distractedm1nd

Wondertan:
@distractedm1nd, conflicts

distractedm1nd:
i literally rebased this morning 🥲

Wondertan (commented):
Just a note that this PR breaks support connection to the non-default networks. Perhaps, we should not merge it for the soonish upcoming release, so we have enough time to fix the issue correctly right after merging it

cmd/flags_node.go
```diff
@@ -34,31 +32,17 @@ func NodeFlags() *flag.FlagSet {
...
-	if network != "" {
-		err := params.SetDefaultNetwork(params.Network(network))
-		if err != nil {
-			return ctx, err
-		}
-	} else {
-		network = string(params.DefaultNetwork())
-	}
-
+	network := Network(ctx)
```
renaynay, line 41: This is more of a maintenance nit [ non-blocking ]: 
the context of why `ParseNodeFlags` must come after `ParseNetwork` (even though it makes sense here) might get forgotten later on, so because getting the network is a requirement for constructing the node store name, maybe `ParseNodeFlags` should take network explicitly as a param?

renaynay:
> Requesting changes because the solution for #1189 in this PR does not fix the core problem of #1189: our libs/pkgs should not import node and its modules, as it is a logical cyclic dependency(node depends on header and wise-versa) and prevents usage of header pkg without importing the node.
> 
> The alternative is passing the network name as a protocol prefix string which is better be optional. Our protocols will be used in other projects. Specifically, the header pkg and there is already ongoing work in https://github.com/celestiaorg/go-header
> 
> Let's revert the change and solve the #1189 issue in the subsequent PR, as it is out of scope. It's fine that the main will be broken for some time for Mamaki.

@Wondertan as long as celestia-node requires header-ex and fraud-ex to contain appended chainIDs, I'm fine with this approach.

nodebuilder/p2p/flags.go
```diff
@@ -2,14 +2,20 @@ package p2p
...
 	flag "github.com/spf13/pflag"
 )
 
-var (
-	p2pMutualFlag = "p2p.mutual"
+const (
+	// DefaultNetwork is the default network of the current build.
+	DefaultNetwork   = Arabica
+	EnvCustomNetwork = "CELESTIA_CUSTOM"
+	nodeNetworkFlag  = "node.network"
```
renaynay, line 17:
```suggestion
	nodeNetworkFlag  = "p2p.network"
```

is the goal of keeping it `node.network` for it to not be breaking? I think i'd rather break CLI now and make it consistent.

nodebuilder/p2p/flags.go
```diff
@@ -24,6 +30,11 @@ Such connection is immune to peer scoring slashing and connection manager trimmi
 Peers must bidirectionally point to each other. (Format: multiformats.io/multiaddr)
 `,
 	)
+	flags.String(
+		nodeNetworkFlag,
+		"",
+		"The name of the network to connect to, e.g. "+listProvidedNetworks(),
```
renaynay, line 36: We should specify here that `p2p.network` should only be passed on `start` and will not have a persisted effect if only passed on `init`

nodebuilder/p2p/flags.go
renaynay, line 17: would call it `networkFlag`

renaynay (changes requested):
Two things to note for clarity (which have already been noted but good to repeat for doc reasons): 

1. **This PR is breaking for all networks besides `arabica`** -- if you try to run with `--p2p.network mamaki`, it will not run against mamaki unless mamaki bootstrappers are upgraded, in which case, they can run but they'll be running on the same `header-ex` protocol ID as `arabica` which is potentially wonky (in the case that a mamaki node accidentally sets an arabica bootstrapper or something, it could try to sync blocks from arabica node).

We need to get a follow-up PR addressing https://github.com/celestiaorg/celestia-node/issues/1189 preferably soon after this is merged (or at least get a PR based on this one out already so we can merge one after the other).

2. This PR will break CLI as well: `node.network` --> `p2p.network`.

nodebuilder/p2p/flags.go
Wondertan: We don't need to specify that as this is default behavior for all the flags

nodebuilder/p2p/flags.go
```diff
@@ -50,3 +61,52 @@ func ParseFlags(
...
+	if parsedNetwork == "" {
+		return ParseNetworkFromEnv()
+	}
+	return Network(parsedNetwork), nil
+}
+
+// ParseNetworkFromEnv tries to parse the network from the environment,
+// and returns either the parsed network or the build's default network
+func ParseNetworkFromEnv() (Network, error) {
+	network := DefaultNetwork
```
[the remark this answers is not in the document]
distractedm1nd: Hmm, not so sure about this one. If the env doesn't find a network in its environment then, what should it return, nil?

Then we need to go from 
```go
func ParseNetwork(cmd *cobra.Command) (Network, error) {
	parsedNetwork := cmd.Flag(nodeNetworkFlag).Value.String()
	// no network set through the flags, so check if there is an override in the env
	if parsedNetwork == "" {
		return parseNetworkFromEnv()
	}
	return Network(parsedNetwork), nil
}
```

to something like 
```go
func ParseNetwork(cmd *cobra.Command) (Network, error) {
	parsedNetwork := cmd.Flag(nodeNetworkFlag).Value.String()
	// no network set through the flags, so check if there is an override in the env
	if parsedNetwork == "" {
		network, err := parseNetworkFromEnv()
		if network != nil {
		        return DefaultNetwork, err
		}
		return network, err
	}
	return Network(parsedNetwork), nil
}
```

which is more awkward imo. Does this snippet change your mind? otherwise ill just implement as you suggest

nodebuilder/p2p/flags.go
distractedm1nd: should p2pMutualFlag be called mutualFlag too? Or rather p2pNetworkFlag and p2pMutual flag together?

nodebuilder/p2p/flags.go
renaynay: Wait a minute - passed flags on `init` have a persisted effect, but not on `start` @Wondertan - that's the default behaviour, not the vice versa

nodebuilder/p2p/flags.go
renaynay: i think the p2p is redundant - lets just make all flag names 

`networkFlag` `mutualFlag`

renaynay: since the flags are defined per-subpkg

nodebuilder/p2p/flags.go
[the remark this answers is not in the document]
renaynay: I don't think it's more awkward -- it's more explicit then what `parseNetworkFromEnv` is doing

`parseNetworkFromEnv` can then return (Network, bool) and you can check envNetwork, ok := ....; if ok {

nodebuilder/p2p/flags.go
Wondertan: I was worried about flags' conceptual dependences on init/start semantics, but I def misread your comment. Adding the note here makes sense to me now

nodebuilder/p2p/flags.go
```diff
@@ -2,28 +2,41 @@ package p2p
...
 
-var (
-	p2pMutualFlag = "p2p.mutual"
+const (
+	// DefaultNetwork is the default network of the current build.
+	DefaultNetwork   = Arabica
+	EnvCustomNetwork = "CELESTIA_CUSTOM"
+	networkFlag      = "p2p.network"
+	mutualFlag       = "p2p.mutual"
 )
```
Wondertan, line 19: optional nit: all the const does not have to be under one const keyword group. In this case, we could clearly separate flag names from Network related costs. Further, we could move DefaultNetwork const closer to where the Network is defined

nodebuilder/header/config.go
```diff
@@ -28,7 +28,7 @@ func DefaultConfig() Config {
 	}
 }
 
-func (cfg *Config) trustedPeers(bpeers params.Bootstrappers) (infos []peer.AddrInfo, err error) {
+func (cfg *Config) trustedPeers(bpeers p2p.Bootstrappers) (infos []peer.AddrInfo, err error) {
```
Wondertan, line 31: idea: Similarly to `modp2p.Bootstrappers`, we may want to make `modheader.TrustedPeers` an application- and module-specific type.

Wondertan (dismissed):
Looking through this again. I love the idea so far. Making `Bootstrappers/Network/...` a module- and application-specific types(independent from our pkgs/libs) fits very well with the modularity vision. Thanks, @distractedm1nd!

Q: Should we keep the main broken for network switching, or should we implement a fix on top of this PR first and then merge everything together? The second approach seems better in case some other bug appears or so.

renaynay:
branch no longer builds

```
--> Building Celestia
# github.com/celestiaorg/celestia-node/cmd/celestia
cmd/celestia/helpers.go:23:41: not enough arguments in call to cmdnode.ParseNodeFlags
	have (context.Context, *cobra.Command)
	want (context.Context, *cobra.Command, "github.com/celestiaorg/celestia-node/nodebuilder/p2p".Network)
make: *** [build] Error 2
```

nodebuilder/p2p/flags.go
```diff
@@ -2,28 +2,41 @@ package p2p
...
 Such connection is immune to peer scoring slashing and connection manager trimming.
 Peers must bidirectionally point to each other. (Format: multiformats.io/multiaddr)
 `,
 	)
+	flags.String(
+		networkFlag,
+		"",
+		"The name of the network to connect to, e.g. "+
+			listProvidedNetworks()+
+			". Is only effective for the start command, and has no effect on init.",
```
renaynay, line 38:
```suggestion
			". Must be passed on both init and start to take effect.",
```

This is because it's needed during `init` for store initialisation purposes but **must also** be passed on `start`

renaynay (changes requested):

nodebuilder/p2p/flags.go
distractedm1nd: done

nodebuilder/p2p/flags.go
distractedm1nd: done

nodebuilder/p2p/flags.go
distractedm1nd: done

nodebuilder/p2p/flags.go
distractedm1nd: done

cmd/flags_node.go
distractedm1nd: done

nodebuilder/p2p/flags.go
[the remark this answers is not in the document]
distractedm1nd: done, but not with (Network, bool)

Wondertan (approved):

PR #1169, opened 2022-09-26
bugfix(nodebuilder/fraud): remove fraud.Subscriber from module

vgonkivs:
It's not nessesary to invoke `new(fraud.Subscriber)` in the Module creation because this instances of fraud.Subscriber is not used anywhere. According to our logic it should be passed to [OnProof](https://github.com/celestiaorg/celestia-node/blob/b95d48647a3a1da2de7a69c4a071ae85ef9159e3/fraud/proof.go#L52) and OnProof is invoked from [Lifecycle](https://github.com/celestiaorg/celestia-node/blob/b95d48647a3a1da2de7a69c4a071ae85ef9159e3/nodebuilder/fraud/fraud.go#L74) that receives fraud.Service as an input parameter.

### DIFF
--- a/nodebuilder/fraud/module.go
+++ b/nodebuilder/fraud/module.go
@@ -4,8 +4,6 @@ import (
 	logging "github.com/ipfs/go-log/v2"
 	"go.uber.org/fx"
 
-	"github.com/celestiaorg/celestia-node/fraud"
-	"github.com/celestiaorg/celestia-node/libs/fxutil"
 	"github.com/celestiaorg/celestia-node/nodebuilder/node"
 )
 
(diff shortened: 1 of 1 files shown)


nodebuilder/fraud/module.go
```diff
@@ -16,12 +16,12 @@ func Module(tp node.Type) fx.Option {
 	case node.Light:
 		return fx.Module(
 			"fraud",
-			fxutil.ProvideAs(ServiceWithSyncer, new(fraud.Service), new(fraud.Subscriber)),
+			fxutil.ProvideAs(ServiceWithSyncer, new(fraud.Service)),
```
renaynay, line 19: you can just do fxutil.Provide(ServiceWithSyncer) bc it already returns a fraud.Service

same with the case for full and bridge

renaynay (changes requested):

nodebuilder/fraud/module.go
renaynay: fx.Provide *

vgonkivs: Yes, thank you.

Wondertan (approved):

distractedm1nd (approved):

PR #1173, opened 2022-09-28
refactor(ipld): use Set/GetCell API from rstm2d

### LINKED ISSUE #824: deps: bump rsmt2d to latest
Note that the task is not to just update the rsmt2d version in go.mod but also leverage new API that simplifies the code using it. Essentially, this involves a bit of refactoring

Wondertan:
The latest rsmt2d version allows us to set cells/share directly into the imported square. Hence, we can avoid using flattened slices and the complexity it introduces when calculating position or index to set a share. Also, we now avoid reimporting the square on each repair retry and overall optimizing the whole process.

Closes #824 
Blocked on https://github.com/celestiaorg/rsmt2d/issues/127 and https://github.com/celestiaorg/rsmt2d/pull/128

### DIFF
--- a/share/eds/retriever.go
+++ b/share/eds/retriever.go
@@ -104,23 +104,18 @@ func (r *Retriever) Retrieve(ctx context.Context, dah *da.DataAvailabilityHeader
 // quadrant request retries. Also, provides an API
 // to reconstruct the block once enough shares are fetched.
 type retrievalSession struct {
+	dah   *da.DataAvailabilityHeader
 	bget  blockservice.BlockGetter
 	adder *ipld.NmtNodeAdder
 
-	treeFn rsmt2d.TreeConstructorFn
-	codec  rsmt2d.Codec
-
-	dah            *da.DataAvailabilityHeader
-	squareImported *rsmt2d.ExtendedDataSquare
-
-	quadrants   []*quadrant
-	sharesLks   []sync.Mutex
-	sharesCount uint32
-
-	squareLk  sync.RWMutex
-	square    [][]byte
-	squareSig chan struct{}
-	squareDn  chan struct{}
+	// TODO(@Wondertan): Extract into a separate data structure https://github.com/celestiaorg/rsmt2d/issues/135
+	squareQuadrants  []*quadrant
+	squareCellsLks   [][]sync.Mutex
+	squareCellsCount uint32
+	squareSig        chan struct{}
+	squareDn         chan struct{}
+	squareLk         sync.RWMutex
+	square           *rsmt2d.ExtendedDataSquare
 
 	span trace.Span
 }
@@ -133,29 +128,31 @@ func (r *Retriever) newSession(ctx context.Context, dah *da.DataAvailabilityHead
 		r.bServ,
 		ipld.MaxSizeBatchOption(size),
 	)
-	ses := &retrievalSession{
-		bget:  blockservice.NewSession(ctx, r.bServ),
-		adder: adder,
-		treeFn: func(_ rsmt2d.Axis, index uint) rsmt2d.Tree {
-			tree := wrapper.NewErasuredNamespacedMerkleTree(uint64(size)/2, index, nmt.NodeVisitor(adder.Visit))
-			return &tree
-		},
-		codec:     share.DefaultRSMT2DCodec(),
-		dah:       dah,
-		quadrants: newQuadrants(dah),
-		sharesLks: make([]sync.Mutex, size*size),
-		square:    make([][]byte, size*size),
-		squareSig: make(chan struct{}, 1),
-		squareDn:  make(chan struct{}),
-		span:      trace.SpanFromContext(ctx),
+
+	treeFn := func(_ rsmt2d.Axis, index uint) rsmt2d.Tree {
+		tree := wrapper.NewErasuredNamespacedMerkleTree(uint64(size)/2, index, nmt.NodeVisitor(adder.Visit))
+		return &tree
 	}
 
-	square, err := rsmt2d.ImportExtendedDataSquare(ses.square, ses.codec, ses.treeFn)
+	square, err := rsmt2d.ImportExtendedDataSquare(make([][]byte, size*size), share.DefaultRSMT2DCodec(), treeFn)
 	if err != nil {
 		return nil, err
 	}
 
-	ses.squareImported = square
+	ses := &retrievalSession{
+		dah:             dah,
+		bget:            blockservice.NewSession(ctx, r.bServ),
+		adder:           adder,
+		squareQuadrants: newQuadrants(dah),
+		squareCellsLks:  make([][]sync.Mutex, size),
+		squareSig:       make(chan struct{}, 1),
+		squareDn:        make(chan struct{}),
+		square:          square,
+		span:            trace.SpanFromContext(ctx),
+	}
+	for i := range ses.squareCellsLks {
+		ses.squareCellsLks[i] = make([]sync.Mutex, size)
+	}
 	go ses.request(ctx)
 	return ses, nil
 }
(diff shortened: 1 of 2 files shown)


Wondertan:
The race tests will fail here until https://github.com/celestiaorg/rsmt2d/pull/128 is merged and released

vgonkivs (dismissed):
LGTM 🚀

ipld/get_namespaced_shares.go
```diff
@@ -77,11 +77,11 @@ type fetchedBounds struct {
 	highest int64
 }
 
-// update checks if the passed index is outside the current bounds,
+// update checks if the passed pos is outside the current bounds,
```
distractedm1nd, line 80: the renames here are incorrect, because they are referring to the argument of update

distractedm1nd (changes requested):
Beautiful, only one nit

ipld/retriever.go
```diff
@@ -131,29 +126,28 @@ func (r *Retriever) newSession(ctx context.Context, dah *da.DataAvailabilityHead
...
-		quadrants: newQuadrants(dah),
-		sharesLks: make([]sync.Mutex, size*size),
-		square:    make([][]byte, size*size),
-		squareSig: make(chan struct{}, 1),
-		squareDn:  make(chan struct{}),
-		span:      trace.SpanFromContext(ctx),
-	}
-
-	square, err := rsmt2d.ImportExtendedDataSquare(ses.square, ses.codec, ses.treeFn)
+	square, err := rsmt2d.ImportExtendedDataSquare(make([][]byte, size*size), DefaultRSMT2DCodec(), func() rsmt2d.Tree {
```
renaynay, line 129: Nit but I'd prefer the following for method calls with verbose args: 

```go
	emptySquare := make([][]byte, size*size)
	treeConstructorFn := func() rsmt2d.Tree {
		tree := wrapper.NewErasuredNamespacedMerkleTree(uint64(size)/2, nmt.NodeVisitor(adder.Visit))
		return &tree
	}	
	square, err := rsmt2d.ImportExtendedDataSquare(emptySquare, DefaultRSMT2DCodec(), treeConstructorFn)
	if err != nil {
		return nil, err
	}

```

ipld/retriever.go
```diff
@@ -310,14 +292,17 @@ func (rs *retrievalSession) doRequest(ctx context.Context, q *quadrant) {
 				if rs.isReconstructed() {
 					return
 				}
-				rs.square[idx] = share
+				if rs.square.GetCell(uint(x), uint(y)) != nil {
```
renaynay, line 295: Is it possible that the caller can call `GetCell` with an index that does not exist inside the square, causing the short circuit and then subsequently try to set the cell of the non-existent index, causing a panic?

ipld/retriever_quadrant.go
```diff
@@ -92,30 +89,16 @@ func newQuadrants(dah *da.DataAvailabilityHeader) []*quadrant {
...
-
-	rootIdx *= halfSquareOffsetRow
-	cellIdx *= halfSquareOffsetCol
-	return rootIdx + cellIdx + offsetX + offsetY
-}
-
-func pow(x, y int) int {
-	return int(math.Pow(float64(x), float64(y)))
+// pos calculates position of a share in a data square.
+func (q *quadrant) pos(rootIdx, cellIdx int) (int, int) {
```
renaynay, line 93: we should probs have a unit test for this method since an inaccuracy in it could cause panic via nil `GetCell` -> `SetCell`

renaynay (commented):
Generally looks good to me but some questions, 1 nit.

ipld/retriever.go
Wondertan: No, it's not possible. The boundaries of x and y are defined by sqaureSize and quadrants

ipld/retriever_quadrant.go
Wondertan: It won't panic, but even if the unit test won't fix this either. The solution is to do checks for the size. However, these rootIdx and cellIdx are bounded by squareSize gotten from DAH and it won't ever get a wrong value out of it in the privatized code, so those ifs are unnecessary

renaynay: > It won't panic, but even if the unit test won't fix this either. 

a unit test wouldn't hurt.

Wondertan: > Why is GetShares exported if it's only used internally to ipld pkg?

Unrelated to this PR

Wondertan: > a unit test wouldn't hurt.

Yeah, butit's useless in this case. The existing tests cover each quadrant fetching, and it works

ipld/retriever.go
renaynay: This should be fuzzed eventually IMO.

ipld/retriever.go
Wondertan: Updated

ipld/retriever.go
Wondertan: AFAIK, fuzzing helps with the testing of logic with arbitrary bytes input. In this case, its only about two numbers

Wondertan:
Rebased on main

walldiss (approved):
Looks good!

share/eds/retriever.go
```diff
@@ -104,23 +104,18 @@ func (r *Retriever) Retrieve(ctx context.Context, dah *da.DataAvailabilityHeader
...
-
-	quadrants   []*quadrant
-	sharesLks   []sync.Mutex
-	sharesCount uint32
-
-	squareLk  sync.RWMutex
-	square    [][]byte
-	squareSig chan struct{}
-	squareDn  chan struct{}
+	// TODO(@Wondertan): Extract into a separate data structure
```
distractedm1nd, line 111: [non-blocking] why? Also we should probably refrain from having TODOs like this unless there is also a relevant issue open (but i am also guilty of this)

share/eds/retriever_quadrant.go
```diff
@@ -93,31 +89,16 @@ func newQuadrants(dah *da.DataAvailabilityHeader) []*quadrant {
...
+	cellIdx += len(q.roots) * q.x
+	rootIdx += len(q.roots) * q.y
+	switch q.source {
+	case rsmt2d.Row:
+		return rootIdx, cellIdx
+	case rsmt2d.Col:
+		return cellIdx, rootIdx
+	default:
+		panic("unknown axis")
+	}
```
distractedm1nd, line 103: nice, much cleaner

share/eds/retriever_quadrant.go
```diff
@@ -45,7 +45,7 @@ type quadrant struct {
 	// source defines the axis for quadrant
 	// it can be either 1 or 0 similar to x and y
 	// where 0 is Row source and 1 is Col respectively
```
distractedm1nd, line 47: [non-blocking] not sure this comment is still necessary, also maybe we could change the name of the field to axis as well?

distractedm1nd (approved):
looks great, thank you, just a few non-blocking questions

share/eds/retriever.go
Wondertan: It's about [this](https://github.com/celestiaorg/rsmt2d/issues/135) issue. The original idea was to just extract a thread-safe EDS within the node repo, but rsmt2d authors actually want to have it there.

share/eds/retriever_quadrant.go
Wondertan: This comment is still correct if we look at the implementation detail of Axis, which is either 1 or 0.

To clarify, I could extend the opening comment. 
> source defines the axis to fetch the quadrant from

The reason it's called source is that all the shares are committed twice, in rows and cols(which you know), and we can get it from either source.

share/eds/retriever.go
Wondertan: I can link it there btw

share/eds/retriever_quadrant.go
Wondertan: This was honestly my main motivation to make this PR 🙂

share/eds/retriever_quadrant.go
distractedm1nd: its correct but the 1/0 is an impl detail that is not relevant to the understanding of its usage anymore
it can probably be simplified, but your choice

Wondertan: Agree

Wondertan: Updated

vgonkivs (approved):

PR #1175, opened 2022-09-29
feat(`api/rpc`): OpenRPC scaffolding, migrating OpenAPI gateway to `api/gateway`

distractedm1nd:
Based on #1056

- Moves the gateway to `api/gateway`. This is no longer bundled with the node as of right now.
- Removes `/service` directory
- Creates new JSON-RPC server in `api/rpc`
- Moves rpc Config definition to `nodebuilder/rpc` (no more cyclical dependency problem)

I would suggest reintroducing the gateway, with an additional config and a default "off" mode, should be done in a different PR.


api/rpc/server.go
```diff
@@ -0,0 +1,60 @@
...
+	"net"
+	"net/http"
+	"time"
+
+	"github.com/filecoin-project/go-jsonrpc"
+)
+
+type Server struct {
+	srv      *http.Server
+	handler  *jsonrpc.RPCServer
```
renaynay, line 14: this naming is confusing, maybe lets do 

`http` and `rpc` ?

api/rpc/server.go
```diff
@@ -0,0 +1,60 @@
...
+		srv: &http.Server{
+			Addr:    address + ":" + port,
+			Handler: handler,
+			// the amount of time allowed to read request headers. set to the default 2 seconds
+			ReadHeaderTimeout: 2 * time.Second,
+		},
+	}
+}
+
+func (s *Server) Start(context.Context) error {
```
renaynay, line 31: doc

nodebuilder/config.go
```diff
@@ -4,6 +4,8 @@ import (
 	"io"
 	"os"
 
+	"github.com/celestiaorg/celestia-node/nodebuilder/rpc"
```
renaynay, line 7: format

nodebuilder/node.go
```diff
@@ -6,6 +6,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/celestiaorg/celestia-node/api/rpc"
```
renaynay, line 9: format

nodebuilder/rpc/module.go
```diff
@@ -3,28 +3,29 @@ package rpc
 import (
 	"context"
 
+	"github.com/celestiaorg/celestia-node/api/rpc"
```
renaynay, line 6: format

api/rpc/handler.go
```diff
@@ -0,0 +1,40 @@
...
+	return &Handler{
+		state:  state,
+		share:  share,
+		header: header,
+		das:    das,
+	}
+}
+
+func (h *Handler) RegisterEndpoints(rpc *Server) {
+	rpc.handler.Register("State", h.state)
```
renaynay, line 36: Does namespace have to be capitalised?

renaynay (commented):
needs some godoc and format but otherwise looks good already.

api/rpc/handler.go
distractedm1nd: no, but I did find out that they need to all be the same in the next PR - its only one connection per namespace

nodebuilder/rpc/module.go
distractedm1nd: done

nodebuilder/node.go
distractedm1nd: done

nodebuilder/config.go
distractedm1nd: done

api/rpc/server.go
distractedm1nd: done

api/rpc/server.go
distractedm1nd: done

api/rpc/server.go
```diff
@@ -0,0 +1,61 @@
...
+// Start starts the RPC Server.
+func (s *Server) Start(context.Context) error {
+	listener, err := net.Listen("tcp", s.http.Addr)
+	if err != nil {
+		return err
+	}
+	s.listener = listener
+	log.Infow("RPC server started", "listening on", listener.Addr().String())
+	//nolint:errcheck
+	go s.http.Serve(listener)
```
walldiss, line 40: Looks like we are listening for just `tcp` here anyway. Maybe we can replace explicit listener here with  more convinient `ListenAndServe` and also don't store it in  the `Server` struct?

Wondertan (changes requested):
RPC is not a module and should not have `nodebuilder`. It's like commands, config, or flags of each module. Each module has an API and each module can be accessed through RPC.

If the RPC server is provided to DI, each module should register its handler there, not some high-level module or API importing it and registering handlers manually. This is something that should be done in this or other RPC related PR and not deferred to some later refactorings.

Gateway though could be a module.

distractedm1nd:
I disagree here. The RPC is a service that is running for the lifecycle of the node, and accesses the other modules - just like the other services. I do not understand the comparison with commands, config, or flags.

Generally, by letting each module be in charge of registering itself on the RPC server, it will no longer be possible to see at a glance which APIs the RPC is exposing. More importantly, this clearly leads to higher coupling (of the individual modules to the RPC) and lower cohesion (inside the individual nodebuilder modules, since they now have to register on an RPC). I can explain this and the consequences more in depth if you disagree.

renaynay:
> RPC is not a module and should not have `nodebuilder`. It's like commands, config, or flags of each module. Each module has an API and each module can be accessed through RPC.
> 
> If the RPC server is provided to DI, each module should register its handler there, not some high-level module or API importing it and registering handlers manually. This is something that should be done in this or other RPC related PR and not deferred to some later refactorings.
> 
> Gateway though could be a module.

@Wondertan I think rpc warrants its own fx.Module -- it is not one of our "Modules" offered by node per-se but I would consider it a module in our software. 

I am, however, okay with each module registering its own endpoint on RPC, but I don't really see the added benefit rn.

api/rpc/server.go
```diff
@@ -0,0 +1,61 @@
...
+type Server struct {
+	http     *http.Server
+	rpc      *jsonrpc.RPCServer
+	listener net.Listener
+}
+
+func NewServer(address string, port string) *Server {
+	handler := jsonrpc.NewServer()
+	return &Server{
+		rpc: handler,
```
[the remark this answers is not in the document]
distractedm1nd: I wasn't sure what this was referencing, so I renamed that handler variable to rpc. Or did you mean the Server struct -> RPC?

api/rpc/server.go
```diff
@@ -0,0 +1,61 @@
...
+		http: &http.Server{
+			Addr:    address + ":" + port,
+			Handler: handler,
+			// the amount of time allowed to read request headers. set to the default 2 seconds
+			ReadHeaderTimeout: 2 * time.Second,
```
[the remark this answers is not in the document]
distractedm1nd: #1276

Not sure what they should be configured to, so I also think its better to be in a different issue

renaynay (dismissed):
* 	// rpc components
	RPCServer *rpc.Server // not optional
*

api/rpc/server.go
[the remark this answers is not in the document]
walldiss: You got it right

walldiss:
I think each module should expose / return `ServeHTTP` or other `handler`- type interface so they can be registered manually in nodebuilder or other higher level caller. It will only expose required handler logic without need to pass transport or middleware level of abstractions inside the components that they don't have any use of. So we can construct server as an independent module and assign all required transport settings and middleware according to our vision of node, not of the pkg itself.

Wondertan (approved):
After plenty of discussions, we(@distractedm1nd, @renaynay, @vgonkivs, and @Wondertan) agreed on this approach.

vgonkivs (approved):

renaynay (approved):
lfg

PR #1190, opened 2022-09-30
nodebuilder: Remove current borked RPC tests in favour of implementing better ones with #962

renaynay:
New tests will replace these outdated ones in https://github.com/celestiaorg/celestia-node/issues/962.

The goal was to remove these tests from the start.


Wondertan:
Two fails, hmm

MSevey:
We want some sort of test like this in the future right? Does it make more sense just to `t.Skip("pending improvements")` so that we don't need to search for this PR to revert when we are ready?  We also then have the constant reminder in the CI about this test being skipped.

renaynay:
Good point @MSevey . We need tests for our rpc but they definitely wouldn't and shouldn't be structured as the ones deleted in this PR. IMO, the new unit tests should be implemented along with the PR that impl this issue: https://github.com/celestiaorg/celestia-node/issues/962.

distractedm1nd (approved):

walldiss (approved):

Wondertan (approved):
👍🏻

PR #1191, opened 2022-09-30
.github: add new ci_release workflow to unify CI

MSevey:
Edit: ~~Testing some things.~~

# Overview
The topic of releases came up in a recent weekly retro. This PR has some suggested updates of how to improve the release process and DRY up the GitHub actions.  Below are some key takeaways from the changes.

- new `make cover` command
- use of composite github actions
- use of reusable github workflows
- common CI file for PRs and release

## `make cover`
This is a simple change to move the coverage report logic into the local Makefile so that coverage reports can be made locally as well as on the online CI.

## Composite GitHub Actions
Composite GitHub actions are great 😄 
They allow for a nice way to refactor common actions. These can be used internal to a repo and external to a repo. In this change you can see the `yamllint` stage of the CI is a composite action that is stored in the `celestiaorg/.github` repo. This means all Celestia repos can use the same `yamllint` action to ensure consistent yaml file linting organization wide. 

The `markdown-lint` stage is another good candidate to be refactored into the `celestiaorg/.github` repo so that all markdown files are consistent. 

Read more here: https://docs.github.com/en/actions/creating-actions/creating-a-composite-action

## Reusable GitHub Workflows
Reusable workflows are similar to composite actions. Both can be used and have different trade offs. In practice it seems more dealers choice as to which you use.

The `hadolint` job is an example of a reusable workflow for linkting of dockerfiles. This PR also moved the go-ci and the docker-build workflows to be reusable to not change any core logic. 

Read more here https://docs.github.com/en/actions/using-workflows/reusing-workflows

## Common CI
The benefit of having a singular common CI for PRs and releases is to ensure that we are always linting, testing, and building our code the same way and not skipping steps for different cases, which can lead to bugs getting released.

**NOTE:** This change will also change the team's workflow in how they are doing releases. Instead of doing a new tag locally, the process will be to manually trigger this workflow from the actions tab on this repo. You will see a drop down for what type of version you are releasing (patch, minor, major). 

## Notes
@renaynay can you give me write access? I don't have write access so had to fork and push. 

@Wondertan the updates to the github actions that this PR makes is what I was referring to on the retro call today.


m-kus:
Hey! 
It would be cool to have semver docker tags for the celestia-node image, I noticed that this PR touches the docker-build workflow so decided to first write here :)
It will require triggering the docker-build workflow on git tags and adding `type=semver,pattern={{version}}` to the Docker meta step.
Pls let me know if it's ok and whether it should be done in a separate PR.

MSevey:
> Hey! It would be cool to have semver docker tags for the celestia-node image, I noticed that this PR touches the docker-build workflow so decided to first write here :) It will require triggering the docker-build workflow on git tags and adding `type=semver,pattern={{version}}` to the Docker meta step. Pls let me know if it's ok and whether it should be done in a separate PR.

We can easily do that, in fact I already have some composite actions here that do the the semver releases https://github.com/celestiaorg/.github/tree/main/.github/actions/major-minor-version

I wanted to talk with the team first about the impacts of that change before making them. Didn't want to change the flow too much :-)

Wondertan:
@MSevey, wen ready for review?

MSevey:
> @MSevey, wen ready for review?

@renaynay I mostly created this PR as a test to play around with some things and show an example of what I was talking about during the retro around release.  I'll start a convo in slack with the Team to see if this is something we would like to close out. If not it is totally fine to push out and close this for now.

MSevey:
Reverted `docker-build` changes as there were permission issues. Ultimately the docker builds can be common in the `celestiaorg/.github` repo so it isn't worth fixing the permission issues here.

MSevey:
Ready for review 👍

.github/workflows/ci_release.yml
```diff
@@ -0,0 +1,69 @@
...
+    name: Markdown Lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18
+      - run: |
+          npm install -g markdownlint-cli@0.32.1
+          markdownlint --config .markdownlint.yaml '**/*.md'
```
Wondertan, line 44: Should we use Makefile here?

.github/workflows/ci_release.yml
```diff
@@ -0,0 +1,69 @@
...
+      - uses: actions/checkout@v3
+      - name: Version Release
+        uses: celestiaorg/.github/.github/actions/version-release@main
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          version-bump: ${{inputs.version}}
+
+  # TODO: permission issue, but not worth fixing as this should be refactored
+  # into the celestiaorg/.github repo, at which point any permission issues will
+  # be resolved.
```
Wondertan, line 65: I saw a lot of progress on `.github` repo. When do we migrate?

docker/Dockerfile
```diff
@@ -1,13 +1,15 @@
...
 COPY . .
+
 ARG TARGETOS TARGETARCH
-RUN env GOOS=$TARGETOS GOARCH=$TARGETARCH make build
+ENV GOOS=$TARGETOS
+ENV GOARCH=$TARGETARCH
+RUN make build
 
-FROM ubuntu
+FROM ubuntu:20.04
```
Wondertan, line 12: I assume no version means the latest. Why not the latest as everywhere else?

Makefile
```diff
@@ -28,6 +28,19 @@ clean:
...
+	@echo "--> Generating Code Coverage"
+	@go install github.com/ory/go-acc@v0.2.6
+	@go-acc -o coverage.txt `go list ./... | grep -v nodebuilder/tests` -- -v
+.PHONY: cover
+
+## deps: install dependencies.
+deps:
+	@echo "--> Installing Dependencies"
+	@go mod download
+.PHONY: deps
```
Wondertan, line 42: Nice

Wondertan (commented):
Mostly questions. Thanks for tidying this up!

.github/workflows/ci_release.yml
MSevey: The reason I didn't move this to the Makefile was because markdown linting is helpful in non go repos, so I would split this out to a composite action in the `celestiaorg/.github` repo instead.

.github/workflows/ci_release.yml
MSevey: As soon as we want. It would be good to make that a follow up after this PR is merged to begin refactoring more. 

Some can be refactored immediate, others it makes sense to coordinate with the other teams first.

docker/Dockerfile
MSevey: From a source control standpoint, it is better to be explicit.  Despite the best intentions, bugs get released, and we don't want to waste time debugging some bug in a dependency.  Better to explicitly upgrade so we can more easily determine when bugs are introduced.

Wondertan: My question mostly comes from the usage of latest for the CI jobs, but ok

Wondertan (approved):

renaynay (approved):

PR #1139, opened 2022-09-21
feat(blocksync): `share.WriteEDS`

distractedm1nd:
Closes #1105

Still a few todos to be solved and comments to be added, but its ready for the first round of reviews now

Will change the base branch of the PR to the feature branch when it is set up (I will wait until there is preliminary work on it)

PR #1196, opened 2022-10-04
fix(ipld/plugin): don't truncate a type byte when it's not in the data

Wondertan:
The code has a special case for the share data with type byte and without. However, it truncates the type byte like it is always there.
The #1139 is the first user of share data without a type byte, which uncovered the bug. The type byte causes issues for us again.

#### refactor(ipld/plugin): extract namespaceHasher into a separate file with some additional cleanups

In preparation for the upcoming test for the namespaceHasher, we extract it into a separate file. Also, some additional cosmetics were done
reducing the API surfuce of the pkg, hiding things that are not supposed to be used outside.

#### perf(ipld/plugin): use sha256-simd to gain 40% speedup for hash computations


renaynay:
nice find

renaynay (dismissed):

renaynay:
Should we have a test for this as well? we don't have coverage for Write in the current tests.

distractedm1nd (changes requested):
I agree that we should add some test coverage - not only for having tests, but also as a sort of "secondary documentation" showing usage

Wondertan:
I'll add a simple test. Re "secondary documentation": this hasher is not supposed to be used outside.

distractedm1nd (approved):

vgonkivs (approved):

PR #1158, opened 2022-09-23
feat(blocksync): `share.ReadEDS`

distractedm1nd:
Closes #1104 
Includes implementation and tests

Based on #1139

The base of this PR will change along with 1139 when the feature branch is set up and ready

PR #1197, opened 2022-10-04
feat(blocksync): Benchmarks for `ReadEDS` and `WriteEDS`

### LINKED ISSUE #1106: share: benchmark `Read/WriteEDS`
Introduced #1104 and #1105 are allocations and CPU-intensive serialization functions. We should know how they perform and always have a starting point on where to improve further.

distractedm1nd:
Based on #1158 

Closes #1106

### DIFF
--- a/share/eds/eds_test.go
+++ b/share/eds/eds_test.go
@@ -176,6 +176,8 @@ func TestReadEDSContentIntegrityMismatch(t *testing.T) {
 	require.ErrorContains(t, err, "share: content integrity mismatch: imported root")
 }
 
+// BenchmarkReadWriteEDS benchmarks the time it takes to write and read an EDS from disk.
+// The benchmark is run with a 4x4 ODS to a 64x64 ODS - a higher value can be used but it will run for much longer.
 func BenchmarkReadWriteEDS(b *testing.B) {
 	ctx, cancel := context.WithCancel(context.Background())
 	b.Cleanup(cancel)
--- a/share/eds/eds_test.go
+++ b/share/eds/eds_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"embed"
 	"encoding/json"
+	"fmt"
 	"os"
 	"testing"
 
@@ -173,6 +174,34 @@ func TestReadEDSContentIntegrityMismatch(t *testing.T) {
 
 	_, err := ReadEDS(context.Background(), f, dah)
 	require.ErrorContains(t, err, "share: content integrity mismatch: imported root")
+}
+
+func BenchmarkReadWriteEDS(b *testing.B) {
+	ctx, cancel := context.WithCancel(context.Background())
+	for originalDataWidth := 4; originalDataWidth <= 64; originalDataWidth *= 2 {
+		tmpDir := b.TempDir()
+		err := os.Chdir(tmpDir)
+		require.NoError(b, err)
+		eds := share.RandBenchmarkEDS(b, originalDataWidth)
+		dah := da.NewDataAvailabilityHeader(eds)
+		b.Run(fmt.Sprintf("Writing %dx%d", originalDataWidth, originalDataWidth), func(b *testing.B) {
+			for i := 0; i < b.N; i++ {
+				f, _ := os.OpenFile(fmt.Sprintf("%d.car", originalDataWidth), os.O_WRONLY|os.O_CREATE, 0600)
+				err := WriteEDS(ctx, eds, f)
+				require.NoError(b, err)
+				f.Close()
+			}
+		})
+		b.Run(fmt.Sprintf("Reading %dx%d", originalDataWidth, originalDataWidth), func(b *testing.B) {
+			for i := 0; i < b.N; i++ {
+				f, _ := os.OpenFile(fmt.Sprintf("%d.car", originalDataWidth), os.O_RDONLY, 0600)
+				_, err = ReadEDS(ctx, f, dah)
+				require.NoError(b, err)
+				f.Close()
+			}
+		})
+	}
+	b.Cleanup(cancel)
 }
 
 func writeRandomEDS(t *testing.T) *rsmt2d.ExtendedDataSquare {
--- a/share/eds/eds_test.go
+++ b/share/eds/eds_test.go
@@ -183,9 +183,6 @@ func BenchmarkReadWriteEDS(b *testing.B) {
 	ctx, cancel := context.WithCancel(context.Background())
 	b.Cleanup(cancel)
 	for originalDataWidth := 4; originalDataWidth <= 64; originalDataWidth *= 2 {
-		tmpDir := b.TempDir()
-		err := os.Chdir(tmpDir)
-		require.NoError(b, err)
 		eds := share.RandEDS(b, originalDataWidth)
 		dah := da.NewDataAvailabilityHeader(eds)
 		b.Run(fmt.Sprintf("Writing %dx%d", originalDataWidth, originalDataWidth), func(b *testing.B) {
@@ -203,7 +200,7 @@ func BenchmarkReadWriteEDS(b *testing.B) {
 				f := new(bytes.Buffer)
 				_ = WriteEDS(ctx, eds, f)
 				b.StartTimer()
-				_, err = ReadEDS(ctx, f, dah)
+				_, err := ReadEDS(ctx, f, dah)
 				require.NoError(b, err)
 			}
 		})
--- a/share/eds/eds_test.go
+++ b/share/eds/eds_test.go
@@ -178,6 +178,7 @@ func TestReadEDSContentIntegrityMismatch(t *testing.T) {
 
 func BenchmarkReadWriteEDS(b *testing.B) {
 	ctx, cancel := context.WithCancel(context.Background())
+	b.Cleanup(cancel)
 	for originalDataWidth := 4; originalDataWidth <= 64; originalDataWidth *= 2 {
 		tmpDir := b.TempDir()
 		err := os.Chdir(tmpDir)
@@ -201,7 +202,6 @@ func BenchmarkReadWriteEDS(b *testing.B) {
 			}
 		})
 	}
-	b.Cleanup(cancel)
 }
 
 func writeRandomEDS(t *testing.T) *rsmt2d.ExtendedDataSquare {
--- a/share/eds/eds_test.go
+++ b/share/eds/eds_test.go
@@ -177,8 +177,8 @@ func TestReadEDSContentIntegrityMismatch(t *testing.T) {
 	require.ErrorContains(t, err, "share: content integrity mismatch: imported root")
 }
 
-// BenchmarkReadWriteEDS benchmarks the time it takes to write and read an EDS from disk.
-// The benchmark is run with a 4x4 ODS to a 64x64 ODS - a higher value can be used but it will run for much longer.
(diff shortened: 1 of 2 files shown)


share/eds/eds_test.go
```diff
@@ -175,6 +176,34 @@ func TestReadEDSContentIntegrityMismatch(t *testing.T) {
...
+		b.Run(fmt.Sprintf("Reading %dx%d", originalDataWidth, originalDataWidth), func(b *testing.B) {
+			for i := 0; i < b.N; i++ {
+				f, _ := os.OpenFile(fmt.Sprintf("%d.car", originalDataWidth), os.O_RDONLY, 0600)
+				_, err = ReadEDS(ctx, f, dah)
+				require.NoError(b, err)
+				f.Close()
+			}
+		})
+	}
+	b.Cleanup(cancel)
```
renaynay, line 204: this should go L181

renaynay (dismissed):

share/eds/eds_test.go
```diff
@@ -175,6 +176,34 @@ func TestReadEDSContentIntegrityMismatch(t *testing.T) {
 	require.ErrorContains(t, err, "share: content integrity mismatch: imported root")
 }
 
+func BenchmarkReadWriteEDS(b *testing.B) {
+	ctx, cancel := context.WithCancel(context.Background())
+	b.Cleanup(cancel)
+	for originalDataWidth := 4; originalDataWidth <= 64; originalDataWidth *= 2 {
```
MSevey, line 182: I'd add either a comment here, or preferably a docstring on the benchmark what these criteria are and why we are using these specific values.

share/test_helpers.go
```diff
@@ -65,3 +65,24 @@ func RandShares(t *testing.T, total int) []Share {
 
 	return shares
 }
+
+func RandBenchmarkEDS(b *testing.B, size int) *rsmt2d.ExtendedDataSquare {
```
MSevey, line 69: All exported functions should have a docstring. Adding detail to confirm what the function is doing.

MSevey (changes requested):

renaynay (approved):
all good once @MSevey 's comments are addressed.

share/eds/eds_test.go
distractedm1nd: done

share/test_helpers.go
distractedm1nd: done

renaynay (approved):

share/test_helpers.go
renaynay: you can resolve @distractedm1nd

share/eds/eds_test.go
```diff
@@ -175,6 +176,36 @@ func TestReadEDSContentIntegrityMismatch(t *testing.T) {
...
+		dah := da.NewDataAvailabilityHeader(eds)
+		b.Run(fmt.Sprintf("Writing %dx%d", originalDataWidth, originalDataWidth), func(b *testing.B) {
+			for i := 0; i < b.N; i++ {
+				f, _ := os.OpenFile(fmt.Sprintf("%d.car", originalDataWidth), os.O_WRONLY|os.O_CREATE, 0600)
+				err := WriteEDS(ctx, eds, f)
+				require.NoError(b, err)
+				f.Close()
+			}
+		})
+		b.Run(fmt.Sprintf("Reading %dx%d", originalDataWidth, originalDataWidth), func(b *testing.B) {
```
Wondertan, line 198: We should `b.ReportAlloc` for these benchmarks. I would say the main point of further optimizations is to reduce allocations and we should see the baseline.

share/eds/eds_test.go
```diff
@@ -175,6 +176,36 @@ func TestReadEDSContentIntegrityMismatch(t *testing.T) {
...
+	for originalDataWidth := 4; originalDataWidth <= 64; originalDataWidth *= 2 {
+		tmpDir := b.TempDir()
+		err := os.Chdir(tmpDir)
+		require.NoError(b, err)
+		eds := share.RandBenchmarkEDS(b, originalDataWidth)
+		dah := da.NewDataAvailabilityHeader(eds)
+		b.Run(fmt.Sprintf("Writing %dx%d", originalDataWidth, originalDataWidth), func(b *testing.B) {
+			for i := 0; i < b.N; i++ {
+				f, _ := os.OpenFile(fmt.Sprintf("%d.car", originalDataWidth), os.O_WRONLY|os.O_CREATE, 0600)
+				err := WriteEDS(ctx, eds, f)
```
Wondertan, line 193: For more accurate results, the timer has to be reset(b.ResetTimer) after opening the file, as opening a file is a syscall and takes a noticeable amount of time. Or you could just open one file and write over and over in it. This does not have to be different files.

share/eds/eds_test.go
```diff
@@ -175,6 +176,36 @@ func TestReadEDSContentIntegrityMismatch(t *testing.T) {
...
+	b.Cleanup(cancel)
+	for originalDataWidth := 4; originalDataWidth <= 64; originalDataWidth *= 2 {
+		tmpDir := b.TempDir()
+		err := os.Chdir(tmpDir)
+		require.NoError(b, err)
+		eds := share.RandBenchmarkEDS(b, originalDataWidth)
+		dah := da.NewDataAvailabilityHeader(eds)
+		b.Run(fmt.Sprintf("Writing %dx%d", originalDataWidth, originalDataWidth), func(b *testing.B) {
+			for i := 0; i < b.N; i++ {
+				f, _ := os.OpenFile(fmt.Sprintf("%d.car", originalDataWidth), os.O_WRONLY|os.O_CREATE, 0600)
```
Wondertan, line 192: I am not actually convinced we should be benchmarking this over OS files. This pollutes the results of each iteration, as there are OS buffers involved, and a Write to OS file can take longer/allocate more than another Write. Let's do these over mem `bytes.Buffer` for cleaner benchmarks of our code by minimizing external factors.

Wondertan (changes requested):

share/eds/eds_test.go
MSevey: this isn't an issue if we switch to an in memory implementation right?

share/eds/eds_test.go
```diff
@@ -175,6 +176,36 @@ func TestReadEDSContentIntegrityMismatch(t *testing.T) {
...
+				err := WriteEDS(ctx, eds, f)
+				require.NoError(b, err)
+				f.Close()
+			}
+		})
+		b.Run(fmt.Sprintf("Reading %dx%d", originalDataWidth, originalDataWidth), func(b *testing.B) {
+			for i := 0; i < b.N; i++ {
+				f, _ := os.OpenFile(fmt.Sprintf("%d.car", originalDataWidth), os.O_RDONLY, 0600)
```
MSevey, line 200: same thing here about Hlib's comment on using an in memory buffer instead of a file to have a cleaner benchmark that isn't touching disk i/o

share/eds/eds_test.go
Wondertan: Right

share/test_helpers.go
```diff
@@ -65,3 +65,26 @@ func RandShares(t *testing.T, total int) []Share {
 
 	return shares
 }
+
+// RandBenchmarkEDS generates an EDS filled with random data using size as the ODS width.
+// It is used for benchmarking purposes, and does not check any errors.
+func RandBenchmarkEDS(b *testing.B, size int) *rsmt2d.ExtendedDataSquare {
```
Wondertan, line 71: We don't need this, as it duplicates the `RandEDS`. However, the `RandEDS` strictly accepts `T` and does not work with `B`. The solution without copying code is simple - use `require.TestingT` as a param in the `RandEDS` and where else is needed. Both `B` and `T` implement the interface and we already depend on the `require` in testhelpers.

share/eds/eds_test.go
distractedm1nd: done

share/eds/eds_test.go
distractedm1nd: done

share/eds/eds_test.go
distractedm1nd: done

share/eds/eds_test.go
```diff
@@ -175,6 +177,39 @@ func TestReadEDSContentIntegrityMismatch(t *testing.T) {
 	require.ErrorContains(t, err, "share: content integrity mismatch: imported root")
 }
 
+// BenchmarkReadWriteEDS benchmarks the time it takes to write and read an EDS from disk.
+// The benchmark is run with a 4x4 ODS to a 64x64 ODS - a higher value can be used but it will run for much longer.
```
MSevey, line 181: line wrap :-)

share/test_helpers.go
```diff
@@ -44,9 +43,10 @@ func RandEDS(t *testing.T, size int) *rsmt2d.ExtendedDataSquare {
 }
 
 // RandShares generate 'total' amount of shares filled with random data.
-func RandShares(t *testing.T, total int) []Share {
+func RandShares(t require.TestingT, total int) []Share {
 	if total&(total-1) != 0 {
-		t.Fatal("Namespace total must be power of 2")
+		t.Errorf("Namespace total must be power of 2")
+		t.FailNow()
```
MSevey, line 49: Why change from `Fatal` to `Error` and `FailNow`? Seems like an unnecessary line add. 

Did you mean this? 
```suggestion
		t.Fatal("Namespace total must be power of 2:", total)
```

share/test_helpers.go
```diff
@@ -44,9 +43,10 @@ func RandEDS(t *testing.T, size int) *rsmt2d.ExtendedDataSquare {
 }
 
 // RandShares generate 'total' amount of shares filled with random data.
-func RandShares(t *testing.T, total int) []Share {
+func RandShares(t require.TestingT, total int) []Share {
 	if total&(total-1) != 0 {
```
MSevey, line 47: nit: since we are touching this function, I'd add a little comment here explaining this little trick.

MSevey (commented):
Mostly nits

MSevey (dismissed):

share/test_helpers.go
distractedm1nd: Yes, it is necessary because of testing.T

MSevey: [non-blocking]
Personal preference, I would structure helper functions as stand alone functions. I don't think they need to have the tester passed through. I think it is cleaner to just return an error and have the calling test handle the testing failure.  

This then also allows us to test these functions to ensure that our tests are using sound functions and we aren't building tests off of faulty helpers.

distractedm1nd: #1275

share/eds/eds_test.go
```diff
@@ -175,6 +177,39 @@ func TestReadEDSContentIntegrityMismatch(t *testing.T) {
...
 
+// BenchmarkReadWriteEDS benchmarks the time it takes to write and read an EDS from disk. The benchmark is run with a
+// 4x4 ODS to a 64x64 ODS - a higher value can be used, but it will run for much longer.
+func BenchmarkReadWriteEDS(b *testing.B) {
+	ctx, cancel := context.WithCancel(context.Background())
+	b.Cleanup(cancel)
+	for originalDataWidth := 4; originalDataWidth <= 64; originalDataWidth *= 2 {
+		tmpDir := b.TempDir()
+		err := os.Chdir(tmpDir)
+		require.NoError(b, err)
```
renaynay, line 188:
```suggestion
```

Wondertan (approved):

PR #1200, opened 2022-10-05
fix(state): returning CoreAccessor from nodebuilder for state.WithMetrics

distractedm1nd:
#1083 introduced `state.WithMetrics` that takes a `CoreAccessor` instead of the `Module`, so the DI currently fails to build.  This fixes the DI for nodes running metrics

### DIFF
--- a/nodebuilder/state/core.go
+++ b/nodebuilder/state/core.go
@@ -23,7 +23,7 @@ func CoreAccessor(
 	corecfg core.Config,
 	signer *apptypes.KeyringSigner,
 	sync *sync.Syncer,
-) Module {
+) (Module, *state.CoreAccessor) {
 	service := state.NewCoreAccessor(signer, sync, corecfg.IP, corecfg.RPCPort, corecfg.GRPCPort)
 	lc.Append(fx.Hook{
 		OnStart: func(ctx context.Context) error {
@@ -34,5 +34,5 @@ func CoreAccessor(
 			return service.Stop(ctx)
 		},
 	})
-	return service
+	return service, service
 }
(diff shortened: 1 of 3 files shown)


vgonkivs (changes requested):
to avoid retunring double `service`, you can make 
```
	baseComponents := fx.Provide(func(module Module) *state.CoreAccessor {
		return module
	})
```

nodebuilder/state/core.go
```diff
@@ -18,21 +10,9 @@ import (
...
-			lifecycleCtx := fxutil.WithLifecycle(ctx, lc)
-			return fraudServ.Lifecycle(ctx, lifecycleCtx, fraud.BadEncoding, fService, service.Start, service.Stop)
-		},
-		OnStop: func(ctx context.Context) error {
-			return service.Stop(ctx)
-		},
-	})
-	return service
+) *state.CoreAccessor {
+	return state.NewCoreAccessor(signer, sync, corecfg.IP, corecfg.RPCPort, corecfg.GRPCPort)
```
renaynay, line 17: Can you use the constructor directly as an fx.Provide?

distractedm1nd: No, because we need to unpack the config values into the constructor - there was a discussion about this with hlib, and we decided to reimplement this "proxy constructor" for this reason (the alternative is changing the constructor params of state.NewCoreAccessor)

vgonkivs (approved):

nodebuilder/state/core.go
renaynay: Ah okay makes sense.

renaynay (approved):

PR #1204, opened 2022-10-06
fix(nodebuilder): fix ordering of modules according to new fx upgrade to 1.18.2

renaynay:
Ordering of invokes inside of modules matters now due to fx 1.18.2 upgrade which was impacting integration tests - now fixed

TODO: 
- [ ] unit tests for ensuring all components are built and running properly **per node type**
- [ ] if can't figure out by EOD (10.10.22), downgrade fx and include upgrade in this PR

### DIFF
--- a/nodebuilder/module.go
+++ b/nodebuilder/module.go
@@ -30,10 +30,10 @@ func ConstructModule(tp node.Type, cfg *Config, store Store) fx.Option {
 		// modules provided by the node
 		p2p.ConstructModule(tp, &cfg.P2P),
 		state.ConstructModule(tp, &cfg.State),
-		header.ConstructModule(tp, &cfg.Header),
 		share.ConstructModule(tp, &cfg.Share),
 		rpc.ConstructModule(tp, &cfg.RPC),
 		core.ConstructModule(tp, &cfg.Core),
+		header.ConstructModule(tp, &cfg.Header),
 		daser.ConstructModule(tp),
 		fraud.ConstructModule(tp),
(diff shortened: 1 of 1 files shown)


nodebuilder/module.go
```diff
@@ -30,10 +30,10 @@ func ConstructModule(tp node.Type, cfg *Config, store Store) fx.Option {
 		// modules provided by the node
 		p2p.ConstructModule(tp, &cfg.P2P),
 		state.ConstructModule(tp, &cfg.State),
-		header.ConstructModule(tp, &cfg.Header),
 		share.ConstructModule(tp, &cfg.Share),
 		rpc.ConstructModule(tp, &cfg.RPC),
 		core.ConstructModule(tp, &cfg.Core),
+		header.ConstructModule(tp, &cfg.Header),
```
MSevey, line 36: REQ: for things like this, where ordering is important, we definitely need some sort of comment to make that explicit so that we don't change this in the future and spend hours/days debugging the same thing. 

Do we know if it is that `header` needs to be after `share`, `rpc`, and `core`? Or just `core`?

MSevey (changes requested):

nodebuilder/module.go
Wondertan: > Do we know if it is that header needs to be after share, rpc, and core? Or just core?

It needs to be after `rpc` as per our investigations.

renaynay:
Superseded by #1218

PR #1210, opened 2022-10-07
chore: disable debug log in Makefile

Wondertan:
The motivation here is to solve debug logs overload on the CI. In practice, they are not useful in any way for debugging issues on CI, which are usually flaky. Debug logs only clog the output requiring devs to download the log archive and analyze what test did fail locally, which is a lot of unnecessary friction. 

Removing the debug log keeps logging failed tests and error logs, allowing devs to identify and resolve the problem locally. Furthermore, it reduces the action run execution time by ___2 minutes___.

Also, we keep existing Makefile commands untouched and introduce new ones that CI can depend on.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,25 +82,25 @@ lint:
 ## test-unit: Running unit tests
 test-unit:
 	@echo "--> Running unit tests"
-	@go test -v `go list ./... | grep -v nodebuilder/tests` -covermode=atomic -coverprofile=coverage.out
+	@go test `go list ./... | grep -v nodebuilder/tests` -covermode=atomic -coverprofile=coverage.out
 .PHONY: test-unit
 
 ## test-unit-race: Running unit tests with data race detector
 test-unit-race:
 	@echo "--> Running unit tests with data race detector"
-	@go test -v -race `go list ./... | grep -v nodebuilder/tests`
+	@go test -race `go list ./... | grep -v nodebuilder/tests`
 .PHONY: test-unit-race
 
 ## test-swamp: Running swamp tests located in nodebuilder/tests
 test-swamp:
 	@echo "--> Running swamp tests"
-	@go test -v ./nodebuilder/tests
+	@go test ./nodebuilder/tests
 .PHONY: test-swamp
 
 ## test-swamp: Running swamp tests with data race detector located in node/tests
 test-swamp-race:
 	@echo "--> Running swamp tests with data race detector"
-	@go test -v -race ./nodebuilder/tests
+	@go test -race ./nodebuilder/tests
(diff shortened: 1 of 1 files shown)


MSevey:
Has the team discussed moving the logs to log files?  This is only really a problem because we are logging to stdout right? 

>  In practice, they are not useful in any way for debugging issues on CI
I would disagree with this. They might not be useful now in the current form, but debug logs in general are very useful. Especially if we are seeing differences between local and online CI runs. 

It seems that the current handling of debug logs is not helpful, which is different than saying debug logs in general are not useful. 

If there isn't a plan already, I would recommend putting together a plan to have logfiles for the node so that we can have the debug logs for CI and in production while at the same time, improving the online CI experience. 

If that means merging this for the short term and bring back the `-v` flag in the future, that is OK with me.

Wondertan:
> Has the team discussed moving the logs to log files? This is only really a problem because we are logging to stdout right?

@MSevey, Do you mean writing logs into the file during CI runs? How we can access those through GitHub? Not sure it provides such features, but I can see it sending logs to some endpoint or so. This is essentially the same as downloading them from GH, but still allows to ignore them if they are not needed(which is the usual case), so I am in favour.

> It seems that the current handling of debug logs is not helpful, which is different than saying debug logs in general are not useful

Yeah, my English is not perfect, but i didn't mean in general, but in particular.

I am happy to improve the strategy for debug logs on CI without cutting them off completely. Previously, we just focused on reducing verbosity of dependencies, but this also reduce verbosity locally, which potentially complicates debugging.

MSevey:
> @MSevey, Do you mean writing logs into the file during CI runs? How we can access those through GitHub? Not sure it provides such features, but I can see it sending logs to some endpoint or so. This is essentially the same as downloading them from GH, but still allows to ignore them if they are not needed(which is the usual case), so I am in favour.


I believe what we want is to use artifacts. https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts

Then ideally, the workflow logs are just showing the tests running/failing and all log output is stored in logfiles that can be downloaded in the event of a failure for debug purposes.

renaynay:
@MSevey I've created an issue to track the above suggestion https://github.com/celestiaorg/celestia-node/issues/1211

renaynay (commented):
Fine with this. 

Ref https://github.com/celestiaorg/celestia-node/issues/1211

renaynay (approved):

distractedm1nd (approved):
Having debug logs the way they currently are prevent us in most cases from seeing what is actually failing, so this is necessary at least until we find a better solution

PR #1213, opened 2022-10-10
Move sanity check logic from rpc handlers to implementations.

### LINKED ISSUE #998: rpc: Move sanity check logic from rpc handlers to implementations
There is some sanity check logic in our rpc handlers as discussed here: https://github.com/celestiaorg/celestia-node/pull/984#discussion_r938706320. This should be moved down and handled by the implementation beneath it, so that the checks don't only occur for RPC users, but also for those using the public API directly.

HoytRen:
I see only 1 implementation. Tell me if here are more.

Close #998


state/core_access.go
```diff
@@ -25,7 +26,10 @@ import (
 	"github.com/celestiaorg/celestia-node/header"
 )
 
-var log = logging.Logger("state")
+var (
+	log              = logging.Logger("state")
+	ErrInvalidAmount = errors.New("amount must be greater than zero")
```
Wondertan, line 31:
```suggestion
	ErrInvalidAmount = errors.New("state: amount must be greater than zero")
```

Wondertan (approved):
LGTM, but defer the proper review to @renaynay and @distractedm1nd

distractedm1nd (approved):
Looks great 🫡 thank you for the contribution!

HoytRen:
Rebased twice, and it still failed...

Wondertan:
Yeah, we have some flakiness

HoytRen:
happy to see it passed :)

distractedm1nd (approved):

renaynay (approved):
Thank you @HoytRen!

PR #1216, opened 2022-10-10
deps: Downgrade uber fx to 1.18.1

renaynay:
Due to a seemingly small change in https://github.com/uber-go/fx/releases/tag/v1.18.2, the ordering of our modules/invocations/provides now impacts the integration tests and node such that certain components do not work as intended. 

An attempt to resolve it was made here: #1204 which **did** fix the issue, but it is still unclear why. Rather than merge that PR without understanding **why** the change makes a difference, we would rather downgrade the fx dependency and keep that PR open as an investigation. 

This PR should fix the integration tests on CI.

cc @Wondertan


renaynay:
Superseded by #1218

PR #1218, opened 2022-10-10
fix(nodebuider/header): initialize Store during start lifecycle instead of fx.Invoke

Wondertan:
Here we solve two issues:
* Tests and red CI without downgrading FX while keeping node modules construction order independent
	*  The recent FX update(to 1.18.2) changed the order of fx.Invoke executions and uncovered an issue with order between connection to a trusted peer and joining the header-sub topic. Connection to the trusted peer happens implicitly during store initialization, and initialization happens during Invoke, hence it's ordering was changed after the update.
* Node constructor free from network calls
	* store.Init does a network call, and it was invoked during node construction. Now, we do this only during the Start, where network calls are acceptable.

Store initialization during the Start lifecycle goes after the Syncer is started, causing Syncer to error, as it requires initialized Store. To overcome this a private `initStore` proxy type is introduced, which explicitly chains the order of fx provides and, subsequently, Starts.

### DIFF
--- a/nodebuilder/header/header.go
+++ b/nodebuilder/header/header.go
@@ -2,19 +2,22 @@ package header
 
 import (
 	"context"
+	"time"
 
 	"github.com/libp2p/go-libp2p-core/host"
 	"github.com/libp2p/go-libp2p-core/peer"
 	"github.com/libp2p/go-libp2p-core/peerstore"
+	"go.uber.org/fx"
 
 	"github.com/celestiaorg/celestia-node/header"
 	"github.com/celestiaorg/celestia-node/header/p2p"
 	"github.com/celestiaorg/celestia-node/header/store"
+	"github.com/celestiaorg/celestia-node/header/sync"
 	"github.com/celestiaorg/celestia-node/params"
 )
 
-// P2PExchange constructs new Exchange for headers.
-func P2PExchange(cfg Config) func(params.Bootstrappers, host.Host) (header.Exchange, error) {
+// newP2PExchange constructs new Exchange for headers.
+func newP2PExchange(cfg Config) func(params.Bootstrappers, host.Host) (header.Exchange, error) {
 	return func(bpeers params.Bootstrappers, host host.Host) (header.Exchange, error) {
 		peers, err := cfg.trustedPeers(bpeers)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


nodebuilder/header/header.go
```diff
@@ -2,19 +2,22 @@ package header
...
 	"github.com/celestiaorg/celestia-node/header/p2p"
 	"github.com/celestiaorg/celestia-node/header/store"
+	"github.com/celestiaorg/celestia-node/header/sync"
 	"github.com/celestiaorg/celestia-node/params"
 )
 
-// P2PExchange constructs new Exchange for headers.
-func P2PExchange(cfg Config) func(params.Bootstrappers, host.Host) (header.Exchange, error) {
+// newP2PExchange constructs new Exchange for headers.
+func newP2PExchange(cfg Config) func(params.Bootstrappers, host.Host) (header.Exchange, error) {
```
Wondertan, line 20: I privatized these intentionally. Such proxy constructors should not be exported. I don't see a use case and we can export those if there is a request or so

cc @renaynay and @distractedm1nd

vgonkivs (approved):

distractedm1nd (approved):

walldiss (approved):

Wondertan:
Substitutes https://github.com/celestiaorg/celestia-node/pull/1216 and https://github.com/celestiaorg/celestia-node/pull/1204

nodebuilder/header/header.go
```diff
@@ -29,22 +32,43 @@ func P2PExchange(cfg Config) func(params.Bootstrappers, host.Host) (header.Excha
...
 
-// InitStore initializes the store.
-func InitStore(ctx context.Context, cfg Config, net params.Network, s header.Store, ex header.Exchange) error {
+// newSyncer constructs new Syncer for headers.
+func newSyncer(ex header.Exchange, store initStore, sub header.Subscriber, duration time.Duration) *sync.Syncer {
+	return sync.NewSyncer(ex, store, sub, duration)
+}
+
+// initStore is a type representing initialized header store.
+// NOTE: It is needed to ensure that Store is always initialized before Syncer is started.
```
renaynay, line 41:
```suggestion
// initStore is a type representing initialized header store.
// NOTE:  It is needed to ensure a deterministic order for integration testing such that 
// Store is always initialized before the Syncer is started.
```

renaynay (approved):

nodebuilder/header/header.go
Wondertan: This is not only for integration testing, btw. The ordering problem the initStore solves and the ordering problem of invokes are two different problems

Issue #1063, opened 2022-09-08
das: make DAS'er params to be configurable 


walldiss:
### Implementation ideas

There are several parameters that are hardcoded into DAS'er right now based on rough observation. Depending on node type and user machine and network configuration those parameters could have different optimal values and should be accessible by config

```
samplingRange
concurrencyLimit
backgroundStoreInterval
genesisHeight
```

Issue #912, opened 2022-07-15
node/config: All time related config fields are in nanoseconds


Wondertan:
The default values for `GracePeriod` and `RoutingTableRefreshPeriod` are set as `time.Minute`, however, in the config it is written in nanoseconds which is not very human-friendly.

Instead, it should be a string that is parsed via `time.ParseDuration`

Wondertan:
Tentatively assigning @vgonkivs

HoytRen:
It seems this is done? in the config.toml, I see 'GracePeriod = "1m0s"', and I believe it's spf13/viper handling the duration string.

derrandz:
Yeah I suppose this is done cc @Wondertan

PR #1177, opened 2022-09-30
refactor(share): Consolidate share pkg

walldiss:
Resolves https://github.com/celestiaorg/celestia-node/issues/878

PR #1221, opened 2022-10-11
Refactor: deprecated global const variables in favor of configurable functional options

derrandz:
# Overview

The codebase is currently littered with default constant global variables acting as configuration variables.
This makes changing configuration values for the Celestia node non-trivial, as it requires code changes and binary re-building.

In this PR, we refactor the global configuration variables logic to use the functional options pattern, which will trivialize the process of configuring a Celestia node.

# References

#709 
#1063 
#912 

## Impact

- [x] the `share` module
- [ ] the `nodebuilder` module
- [x] the `das` module
- [ ] the `fraud` module
- [ ] the `header` module
- [ ] the `ipld` module
- [ ] the `params` module
- [ ] the `rpc` module

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,11 @@ clean:
 	@echo "--> Cleaning up ./build"
 	@rm -rf build/*
 
+## clean-cache: removes all go cache, useful for when your tests aren't passing due to cached old code
+clean-cache:
+	@go clean -r -testcache -cache -modcache .
+.PHONY: clean-cache
+
 ## install: Build and install the celestia-node binary into the $PREFIX (/usr/local/ by default) directory.
 install: build
 	@echo "--> Installing Celestia"
--- a/Makefile
+++ b/Makefile
@@ -28,11 +28,6 @@ clean:
 	@echo "--> Cleaning up ./build"
 	@rm -rf build/*
 
-## clean-cache: removes all go cache, useful for when your tests aren't passing due to cached old code
-clean-cache:
-	@go clean -r -testcache -cache -modcache .
-.PHONY: clean-cache
-
 ## install: Build and install the celestia-node binary into the $PREFIX (/usr/local/ by default) directory.
 install: build
 	@echo "--> Installing Celestia"
--- a/das/coordinator.go
+++ b/das/coordinator.go
@@ -9,9 +9,10 @@ import (
 
 // samplingCoordinator runs and coordinates sampling workers and updates current sampling state
 type samplingCoordinator struct {
-	concurrencyLimit int
-	getter           header.Getter
-	sampleFn         sampleFn
+	daserCfg Config
+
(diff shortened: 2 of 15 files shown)


Wondertan (commented):
@derrandz, I thought you wanted to start from DASer 😅 . There is ongoing refactoring in share pkg #1177, and it will be easier to work on share pkg afterward. 

Also, you made the Impact todo list of all the pkgs within this PR, but we should go about this with 1pkg:1PR, so I would rebrand this pr to be specifically for the `share` pkg and leave it until the #1177 is merged. Meanwhile, you can work on other pkgs. 

It will be useful to transfer the Impact list on the issue and track progress of each pkg there

Makefile
```diff
@@ -28,6 +28,11 @@ clean:
 	@echo "--> Cleaning up ./build"
 	@rm -rf build/*
 
+## clean-cache: removes all go cache, useful for when your tests aren't passing due to cached old code
```
Wondertan, line 31: Not sure I get the use case. How tests cannot pass because of old cached code? I've never encountered this. Maybe there is another solution to the problem?

Makefile
```diff
@@ -28,6 +28,11 @@ clean:
...
+clean-cache:
+	@go clean -r -testcache -cache -modcache .
+.PHONY: clean-cache
```
Wondertan, line 34: This should not be part of the PR. Pls extract it into a separate one

Makefile
derrandz: Prior to go1.18, I've used to run into issues such as not seeing a debug line I've put in a test or in a module being tested, and pruning the cache made those lines appear, and while I was making my way with this PR, I had to run this line to remove any doubts.

I agree, the use-case is not well-defined. (I don't feel strongly about this addition)

derrandz: Actually, I've just ran into this with the `das` package. Maybe we can discuss this and ship it in another PR.

derrandz:
> @derrandz, I thought you wanted to start from DASer 😅 . There is ongoing refactoring in share pkg #1177, and it will be easier to work on share pkg afterward.
> 
> Also, you made the Impact todo list of all the pkgs within this PR, but we should go about this with 1pkg:1PR, so I would rebrand this pr to be specifically for the `share` pkg and leave it until the #1177 is merged. Meanwhile, you can work on other pkgs.
> 
> It will be useful to transfer the Impact list on the issue and track progress of each pkg there

@Wondertan Thank you for eyeballing this WIP, and thanks for raising my attention to the on-going refactors (_I wasn't aware_). I agree with the rebranding suggestion, I will close this PR in favor of PRs of other modules (_the 1pkg:1pr_)

So far, this PR has only impacted:

- `share` package
- `das` package

derrandz:
Closing in favor of #1224 #1225 as per the conversation above.

Issue #183, opened 2021-11-09
IPLD/NMT improvements


Wondertan:
Right now, we store/exchange on the network redundant data that brings no benefits, and we should fix this together during BlockSync Overhaul:
- [ ] Remove prepended one byte from each NMT IPLD node body([here](https://github.com/celestiaorg/celestia-node/blob/main/ipld/plugin/nmt.go#L145) and [here](https://github.com/celestiaorg/celestia-node/blob/main/ipld/plugin/nmt.go#L247))
	*  It is possible to indirectly get the NMT Node type without serializing it. We can look at the number of links and the data size to determine its type(leaf or non-leaf). Look [here](https://github.com/celestiaorg/celestia-node/blob/main/ipld/plugin/nmt.go#L56-L67) for an example.
- [ ] Remove additional NMT IPLD node per each leaf
	* Each `nmtLeafNode` [links](https://github.com/celestiaorg/celestia-node/blob/main/ipld/plugin/nmt.go#L292) to an additional IPLD Node, while it should keep that body inside itself
	* This is needed to reduce one roundtrip for DASing operation and, in general, reduce the size of each block

evan-forbes:
> It is possible to get Namespace for each share encoded in inner non-leaf nodes of the NMT tree.

I think I get it, but I'm not 100% sure. Mind elaborating on this a bit more? What if some non-leaf nodes, along with their respective leaf nodes, can't be found?

liamsi:
Referencing this here as it is highly related: https://github.com/celestiaorg/celestia-node/pull/244#discussion_r761576295

Wondertan:
@liamsi, I made a separate issue out of this. Even though they are related, your comments deserve a separate issue.

Wondertan:
https://github.com/celestiaorg/celestia-node/pull/464 relies on the fact that our NMT tree looks like:
```
---X
-X---X
X-X-X-X
X-X-X-X
```

While one of the points of this issue is to make the tree look like this: 
```
---X
-X---X
X-X-X-X
```

So once we fix this issue, we should also fix https://github.com/celestiaorg/celestia-node/pull/464

adlerjohn:
> 8 bytes additional Namespace for each share via NMT wrapper 
> Can be avoided if we enable once

Why can't it be avoided today? Storage format is an implementation detail. So long as you can losslessly convert between formats, it doesn't matter what it's stored as.

liamsi:
[...]

PR #1223, opened 2022-10-11
refactor(share): IPLD/NMT improvements

walldiss:
Closes https://github.com/celestiaorg/celestia-node/issues/183

### DIFF
--- a/share/get.go
+++ b/share/get.go
@@ -90,7 +90,7 @@ func GetProofsForShares(
 			if err != nil {
 				return nil, err
 			}
-			proofs[index] = NewShareWithProof(index, s.RawData()[1:], proof)
+			proofs[index] = NewShareWithProof(index, s.RawData(), proof)
 		}
 	}
 
@@ -99,7 +99,6 @@ func GetProofsForShares(
 
 // leafToShare converts an NMT leaf into a Share.
 func leafToShare(nd format.Node) Share {
-	// * First byte represents the type of the node, and is unrelated to the actual share data
 	// * Additional namespace is prepended so that parity data can be identified with a parity namespace, which we cut off
-	return nd.RawData()[1+NamespaceSize:] // TODO(@Wondertan): Rework NMT/IPLD plugin to avoid the type byte
+	return nd.RawData()[NamespaceSize:]
 }
--- a/share/ipld/get.go
+++ b/share/ipld/get.go
@@ -58,9 +58,9 @@ func GetLeaf(
 
 	// look for links
 	lnks := nd.Links()
-	if len(lnks) == 1 {
-		// in case there is only one we reached tree's bottom, so finally request the leaf.
-		return GetNode(ctx, bGetter, lnks[0].Cid)
+	if len(lnks) == 0 {
+		// in case there is none, we reached tree's bottom, so finally return the leaf.
+		return nd, err
 	}
 
 	// route walk to appropriate children
@@ -137,16 +137,7 @@ func GetLeaves(ctx context.Context,
 				}
 				// check links to know what we should do with the node
 				lnks := nd.Links()
-				if len(lnks) == 1 { // so we are almost there
-					// the reason why the comment on 'total' is lying, as each
-					// leaf has its own additional leaf(hack) so get it
-					nd, err := GetNode(ctx, bGetter, lnks[0].Cid)
-					if err != nil {
-						// again, we don't really care much, just fetch as much as possible
-						span.RecordError(err)
-						span.SetStatus(codes.Error, err.Error())
-						return
-					}
+				if len(lnks) == 0 {
 					// successfully fetched a share/leaf
 					// ladies and gentlemen, we got em!
 					span.SetStatus(codes.Ok, "")
@@ -262,8 +253,7 @@ func GetLeavesByNamespace(
 				}
 
 				links := nd.Links()
-				linkCount := uint64(len(links))
-				if linkCount == 1 {
+				if len(links) == 0 {
 					// successfully fetched a leaf belonging to the namespace
 					span.SetStatus(codes.Ok, "")
 					leaves[j.pos] = nd
@@ -323,7 +313,7 @@ func GetProof(
 	}
 	// look for links
 	lnks := nd.Links()
-	if len(lnks) == 1 {
+	if len(lnks) == 0 {
 		p := make([]cid.Cid, len(proof))
 		copy(p, proof)
 		return p, nil
--- a/share/ipld/namespace_hasher.go
+++ b/share/ipld/namespace_hasher.go
@@ -45,17 +45,11 @@ func (n *namespaceHasher) Write(data []byte) (int, error) {
 	innerNodeSize, leafNodeSize := (nln*2+hln)*2, nln+appconsts.ShareSize
 	switch ln {
 	default:
-		return 0, fmt.Errorf("ipld: wrong sized data written to the hasher")
+		return 0, fmt.Errorf("ipld: wrong sized data written to the hasher, len: %v", ln)
 	case innerNodeSize:
 		n.tp = nmt.NodePrefix
 	case leafNodeSize:
 		n.tp = nmt.LeafPrefix
-	case innerNodeSize + typeSize: // w/ additional type byte
-		n.tp = nmt.NodePrefix
-		data = data[typeSize:]
-	case leafNodeSize + typeSize: // w/ additional type byte
-		n.tp = nmt.LeafPrefix
-		data = data[typeSize:]
 	}
 
 	n.data = data
--- a/share/ipld/nmt.go
+++ b/share/ipld/nmt.go
@@ -17,7 +17,6 @@ import (
 
 	"github.com/celestiaorg/celestia-app/pkg/appconsts"
 	"github.com/celestiaorg/celestia-app/pkg/da"
-	"github.com/celestiaorg/nmt"
 )
 
 var (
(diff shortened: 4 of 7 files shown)


share/ipld/nmt.go
```diff
@@ -95,7 +92,7 @@ func NewNMTNode(id cid.Cid, l, r []byte) ipld.Node {
 }
 
 func (n nmtNode) RawData() []byte {
-	return append([]byte{nmt.NodePrefix}, append(n.l, n.r...)...)
+	return append(n.l, n.r...)
```
Wondertan, line 95: What about that optimization?😉 

I feel like you had another one on your mind, but don't remember which one exactly.

share/ipld/nmt.go
```diff
@@ -188,32 +185,15 @@ func (n nmtNode) Size() (uint64, error) {
...
-	hash: 		%x,
-	len(Data): 	%v
-}`, l.cid.Hash(), len(l.Data))
-}
-
-func (l nmtLeafNode) Loggable() map[string]interface{} {
-	return nil
+	b, err := blocks.NewBlockWithCid(data, id)
+	if err != nil {
+		log.Warn("wrong hash for block", "cid", id)
```
Wondertan, line 194: Maybe we should panic here. If this happens, its definitely a broken code

share/ipld/get.go
```diff
@@ -0,0 +1,391 @@
...
+	// request the node
+	nd, err := GetNode(ctx, bGetter, root)
+	if err != nil {
+		return nil, err
+	}
+
+	// look for links
+	lnks := nd.Links()
+	if len(lnks) == 0 {
+		// in case there is only one we reached tree's bottom, so finally request the leaf.
```
Wondertan, line 62: Update comment

share/ipld/namespace_hasher.go
```diff
@@ -50,12 +50,6 @@ func (n *namespaceHasher) Write(data []byte) (int, error) {
 		n.tp = nmt.NodePrefix
 	case leafNodeSize:
 		n.tp = nmt.LeafPrefix
-	case innerNodeSize + typeSize: // w/ additional type byte
-		n.tp = nmt.NodePrefix
-		data = data[typeSize:]
-	case leafNodeSize + typeSize: // w/ additional type byte
-		n.tp = nmt.LeafPrefix
-		data = data[typeSize:]
```
Wondertan, line 58: Question to the whole team. Do we want to keep compatibility with the past version?

share/ipld/nmt_adder.go
```diff
@@ -60,3 +56,26 @@ func (n *NmtNodeAdder) Commit() error {
...
+
+// MaxSizeBatchOption sets the maximum amount of buffered data before writing
+// blocks.
+func MaxSizeBatchOption(size int) ipld.BatchOption {
+	return ipld.MaxSizeBatchOption(BatchSize(size))
+}
+
+// BatchSize calculates the amount of nodes that are generated from block of 'squareSizes'
+// to be batched in one write.
+func BatchSize(squareSize int) int {
```
Wondertan, line 68: This has to be updated because we remov one additional layer. Hmm, I think there was a test verifying its correctness. Is it still there?

share/ipld/nmt.go
walldiss: Added another commit with allocation optimisation.

share/ipld/get.go
walldiss: done

share/ipld/namespace_hasher.go
Wondertan: As per our sync discussions, we don't want to have backwards compatibility

distractedm1nd (changes requested):
Looks great, excited for this

One more addition that I couldn't add as a review comment: The comment in `BatchSize` must also be updated

share/ipld/nmt_adder.go
Wondertan: As per the sync discussion with Vlad, the calculations are correct here. The comments still needs to be updated though

share/ipld/nmt_adder.go
```diff
@@ -39,10 +39,10 @@ func (n *NmtNodeAdder) Visit(hash []byte, children ...[]byte) {
 	switch len(children) {
 	case 1:
```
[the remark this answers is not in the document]
walldiss: Yes, the structure is the same, the extra reference was only the matter of `GetNode` function doing extra call.

share/ipld/nmt.go
```diff
@@ -65,212 +65,62 @@ func GetNode(ctx context.Context, bGetter blockservice.BlockGetter, root cid.Cid
...
 }
 
 var _ ipld.Node = (*nmtNode)(nil)
-var _ ipld.Node = (*nmtLeafNode)(nil)
 
 type nmtNode struct {
-	cid  cid.Cid
-	l, r []byte
+	blocks.Block
+	ipld.Resolver
```
Wondertan, line 75: Resolver, hmm. Do we even need this if we stick panics anyway?

share/get.go
```diff
@@ -101,5 +101,5 @@ func GetProofsForShares(
 func leafToShare(nd format.Node) Share {
 	// * First byte represents the type of the node, and is unrelated to the actual share data
 	// * Additional namespace is prepended so that parity data can be identified with a parity namespace, which we cut off
-	return nd.RawData()[1+NamespaceSize:] // TODO(@Wondertan): Rework NMT/IPLD plugin to avoid the type byte
+	return nd.RawData()[NamespaceSize:] // TODO(@Wondertan): Rework NMT/IPLD plugin to avoid the type byte
```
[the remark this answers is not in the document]
walldiss: Thank you, updated comment here and in `BatchSize` description

share/ipld/nmt.go
```diff
@@ -65,212 +65,62 @@ func GetNode(ctx context.Context, bGetter blockservice.BlockGetter, root cid.Cid
...
-	return nil, fmt.Errorf(
-		"expected first byte of block to be either the leaf or inner node prefix: (%x, %x), got: %x)",
-		leafPrefix,
-		innerPrefix,
-		domainSeparator,
-	)
+	return nmtNode{Block: block}, nil
 }
 
 var _ ipld.Node = (*nmtNode)(nil)
```
Wondertan, line 71: This is a good thing to do in general for non-ide users, but we are not doing this, so I think we can just remove this as well to clean up things completely.

Wondertan:
@walldiss, please also double check all the comments next change code, so we didn't miss anything and rebase

share/ipld/nmt.go
walldiss: Agree, only one stubbing approach is nicer. Removed Resolver

share/ipld/nmt.go
walldiss: Removed compile time interface check.

share/ipld/nmt.go
```diff
@@ -65,212 +66,67 @@ func GetNode(ctx context.Context, bGetter blockservice.BlockGetter, root cid.Cid
...
-func (l nmtLeafNode) Stat() (*ipld.NodeStat, error) {
-	return &ipld.NodeStat{}, nil
+func (n nmtNode) left() []byte {
+	return n.RawData()[:nmtHashSize]
 }
 
-func (l nmtLeafNode) Size() (uint64, error) {
-	return 0, nil
+func (n nmtNode) right() []byte {
+	return n.RawData()[nmtHashSize:]
```
renaynay, line 129: general Q: I don't see this as an issue bc shareSize > nmtHashSize+nmtHashSize but should these methods return nil if nmtNode is leaf? 

I know they're only called on innerNode but wondering if we should add in an additional check.

share/ipld/get.go
```diff
@@ -58,9 +58,9 @@ func GetLeaf(
 
 	// look for links
 	lnks := nd.Links()
-	if len(lnks) == 1 {
-		// in case there is only one we reached tree's bottom, so finally request the leaf.
-		return GetNode(ctx, bGetter, lnks[0].Cid)
+	if len(lnks) == 0 {
+		// in case there is none, we reached tree's bottom, so finally request the leaf.
```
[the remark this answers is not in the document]
walldiss: updated

share/ipld/nmt.go
```diff
@@ -69,82 +63,42 @@ func GetNode(ctx context.Context, bGetter blockservice.BlockGetter, root cid.Cid
...
-	if bytes.Equal(domainSeparator, innerPrefix) {
-		return &nmtNode{
-			cid: block.Cid(),
-			l:   data[typeSize : typeSize+nmtHashSize],
-			r:   data[typeSize+nmtHashSize:],
-		}, nil
+	innerNodeSize, leafNodeSize := (nmtHashSize)*2, NamespaceSize+consts.ShareSize
+	switch len(block.RawData()) {
+	default:
+		return nil, fmt.Errorf("ipld: wrong sized data carried in block")
```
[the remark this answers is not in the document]
walldiss: added here and in another switch related to data length

share/ipld/nmt.go
```diff
@@ -40,6 +37,12 @@ const (
 	// nmtHashSize is the size of a digest created by an NMT in bytes.
 	nmtHashSize = 2*consts.NamespaceSize + sha256.Size
 
+	// innerNodeSize is the size of data in inner nodes.
+	innerNodeSize = nmtHashSize * 2
+
+	// innerNodeSize is the size of data in leaf nodes.
```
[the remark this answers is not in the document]
walldiss: good catch, fixed it

share/ipld/nmt.go
```diff
@@ -65,212 +66,67 @@ func GetNode(ctx context.Context, bGetter blockservice.BlockGetter, root cid.Cid
...
-	return nmtNode{id, l, r}
+	blocks.Block
 }
 
-func (n nmtNode) RawData() []byte {
-	return append([]byte{nmt.NodePrefix}, append(n.l, n.r...)...)
+func newNMTNode(id cid.Cid, data []byte) nmtNode {
+	b, err := blocks.NewBlockWithCid(data, id)
+	if err != nil {
+		panic(fmt.Sprintf("wrong hash for block, cid: %s", id))
```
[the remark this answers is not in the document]
walldiss: added suggested

share/ipld/nmt.go
walldiss: removed those funcs in favour of subslicing directly, they had single call each.

renaynay (approved):

Wondertan (approved):

distractedm1nd (approved):

PR #1225, opened 2022-10-12
refactor(daser): use functional options pattern to configure daser

derrandz:
# Overview
The DAS module is currently reliant on default constant global variables acting as configuration variables.

This makes changing configuration values for the DAS module non-trivial, as it requires code changes and binary re-building.

In this PR, we refactor the global configuration variables used in the DAS module to use the functional options pattern, which will trivialize the process of configuring a Celestia node.

# Reference Issue
#1063 
#709 

# Impact / Change

The `das` package is now configurable using a set of functional options, namely:
```
WithSamplingRange(uint64)
WithConcurrencyLimit(int)
WithBackgroundStoreInterval(time.Duration)
WithPrioritySizeQueue(int)
WithGenesisHeight)
```

This allows the `das` package to be configured with values other than the default parameters defined in `das/options.go:29`

An example of the `das` package being configured with these options and instantiated with them is: `nodebuilder/das/module.go:22`

# Owners

@derrandz

### DIFF
--- a/das/coordinator.go
+++ b/das/coordinator.go
@@ -10,8 +10,9 @@ import (
 // samplingCoordinator runs and coordinates sampling workers and updates current sampling state
 type samplingCoordinator struct {
 	concurrencyLimit int
-	getter           header.Getter
-	sampleFn         sampleFn
+
+	getter   header.Getter
+	sampleFn sampleFn
 
 	state coordinatorState
 
@@ -35,15 +36,15 @@ type result struct {
 }
 
 func newSamplingCoordinator(
-	concurrency int,
-	samplingRange uint64,
+	params Parameters,
 	getter header.Getter,
-	sample sampleFn) *samplingCoordinator {
+	sample sampleFn,
+) *samplingCoordinator {
 	return &samplingCoordinator{
-		concurrencyLimit: concurrency,
+		concurrencyLimit: params.ConcurrencyLimit,
 		getter:           getter,
 		sampleFn:         sample,
-		state:            newCoordinatorState(samplingRange),
+		state:            newCoordinatorState(params),
 		resultCh:         make(chan result),
 		updHeadCh:        make(chan uint64),
 		waitCh:           make(chan *sync.WaitGroup),
--- a/das/coordinator_test.go
+++ b/das/coordinator_test.go
@@ -14,18 +14,13 @@ import (
 )
 
 func TestCoordinator(t *testing.T) {
-	concurrency := 10
-	samplingRange := uint64(10)
-	networkHead := uint64(500)
-	sampleFrom := uint64(genesisHeight)
-	timeoutDelay := 125 * time.Second
-
 	t.Run("test run", func(t *testing.T) {
-		ctx, cancel := context.WithTimeout(context.Background(), timeoutDelay)
+		testParams := defaultTestParams()
 
-		sampler := newMockSampler(sampleFrom, networkHead)
+		ctx, cancel := context.WithTimeout(context.Background(), testParams.timeoutDelay)
+		sampler := newMockSampler(testParams.sampleFrom, testParams.networkHead)
+		coordinator := newSamplingCoordinator(testParams.dasParams, getterStub{}, onceMiddleWare(sampler.sample))
 
-		coordinator := newSamplingCoordinator(concurrency, samplingRange, getterStub{}, onceMiddleWare(sampler.sample))
 		go coordinator.run(ctx, sampler.checkpoint)
 
 		// check if all jobs were sampled successfully
@@ -36,25 +31,27 @@ func TestCoordinator(t *testing.T) {
 		assert.Emptyf(t, coordinator.state.failed, "failed list should be empty")
 
 		cancel()
-		stopCtx, cancel := context.WithTimeout(context.Background(), timeoutDelay)
+		stopCtx, cancel := context.WithTimeout(context.Background(), testParams.timeoutDelay)
 		defer cancel()
 		assert.NoError(t, coordinator.wait(stopCtx))
 		assert.Equal(t, sampler.finalState(), newCheckpoint(coordinator.state.unsafeStats()))
 	})
 
 	t.Run("discovered new headers", func(t *testing.T) {
-		ctx, cancel := context.WithTimeout(context.Background(), timeoutDelay)
+		testParams := defaultTestParams()
 
-		sampler := newMockSampler(sampleFrom, networkHead)
+		ctx, cancel := context.WithTimeout(context.Background(), testParams.timeoutDelay)
 
-		coordinator := newSamplingCoordinator(concurrency, samplingRange, getterStub{}, sampler.sample)
+		sampler := newMockSampler(testParams.sampleFrom, testParams.networkHead)
+
+		coordinator := newSamplingCoordinator(testParams.dasParams, getterStub{}, sampler.sample)
 		go coordinator.run(ctx, sampler.checkpoint)
 
 		time.Sleep(50 * time.Millisecond)
 		// discover new height
 		for i := 0; i < 200; i++ {
 			// mess the order by running in go-routine
-			sampler.discover(ctx, networkHead+uint64(i), coordinator.listen)
+			sampler.discover(ctx, testParams.networkHead+uint64(i), coordinator.listen)
 		}
 
 		// check if all jobs were sampled successfully
@@ -65,36 +62,44 @@ func TestCoordinator(t *testing.T) {
 		assert.Emptyf(t, coordinator.state.failed, "failed list should be empty")
 
 		cancel()
-		stopCtx, cancel := context.WithTimeout(context.Background(), timeoutDelay)
+		stopCtx, cancel := context.WithTimeout(context.Background(), testParams.timeoutDelay)
 		defer cancel()
 		assert.NoError(t, coordinator.wait(stopCtx))
 		assert.Equal(t, sampler.finalState(), newCheckpoint(coordinator.state.unsafeStats()))
 	})
 
 	t.Run("prioritize newly discovered over known", func(t *testing.T) {
-		sampleFrom := uint64(1)
-		networkHead := uint64(10)
+
+		testParams := defaultTestParams()
+
+		testParams.dasParams.ConcurrencyLimit = 1
+		testParams.dasParams.SamplingRange = 4
+
+		testParams.sampleFrom = 1
+		testParams.networkHead = 10
 		toBeDiscovered := uint64(20)
-		samplingRange := uint64(4)
-		concurrency := 1
 
-		sampler := newMockSampler(sampleFrom, networkHead)
+		sampler := newMockSampler(testParams.sampleFrom, testParams.networkHead)
 
-		ctx, cancel := context.WithTimeout(context.Background(), timeoutDelay)
+		ctx, cancel := context.WithTimeout(context.Background(), testParams.timeoutDelay)
 
 		// lock worker before start, to not let it indicateDone before discover
-		lk := newLock(sampleFrom, sampleFrom)
+		lk := newLock(testParams.sampleFrom, testParams.sampleFrom)
 
 		// expect worker to prioritize newly discovered  (20 -> 10) and then old (0 -> 10)
-		order := newCheckOrder().addInterval(sampleFrom, samplingRange) // worker will pick up first job before discovery
-		order.addStacks(networkHead+1, toBeDiscovered, samplingRange)
-		order.addInterval(samplingRange+1, toBeDiscovered)
+		order := newCheckOrder().addInterval(
+			testParams.sampleFrom,
+			testParams.dasParams.SamplingRange,
+		) // worker will pick up first job before discovery
+
+		order.addStacks(testParams.networkHead+1, toBeDiscovered, testParams.dasParams.SamplingRange)
+		order.addInterval(testParams.dasParams.SamplingRange+1, toBeDiscovered)
 
 		// start coordinator
-		coordinator := newSamplingCoordinator(concurrency, samplingRange, getterStub{},
+		coordinator := newSamplingCoordinator(testParams.dasParams, getterStub{},
 			lk.middleWare(
-				order.middleWare(
-					sampler.sample)),
+				order.middleWare(sampler.sample),
+			),
 		)
 		go coordinator.run(ctx, sampler.checkpoint)
 
@@ -108,7 +113,7 @@ func TestCoordinator(t *testing.T) {
 		assert.Equal(t, 0, sampler.sampledAmount())
 
 		// unblock worker
-		lk.release(sampleFrom)
+		lk.release(testParams.sampleFrom)
 
 		// check if all jobs were sampled successfully
 		assert.NoError(t, sampler.finished(ctx), "not all headers were sampled")
@@ -118,26 +123,28 @@ func TestCoordinator(t *testing.T) {
 		assert.Emptyf(t, coordinator.state.failed, "failed list should be empty")
 
 		cancel()
-		stopCtx, cancel := context.WithTimeout(context.Background(), timeoutDelay)
+		stopCtx, cancel := context.WithTimeout(context.Background(), testParams.timeoutDelay)
 		defer cancel()
 		assert.NoError(t, coordinator.wait(stopCtx))
 		assert.Equal(t, sampler.finalState(), newCheckpoint(coordinator.state.unsafeStats()))
 	})
 
 	t.Run("priority routine should not lock other workers", func(t *testing.T) {
-		networkHead := uint64(20)
-		ctx, cancel := context.WithTimeout(context.Background(), timeoutDelay)
+		testParams := defaultTestParams()
 
-		sampler := newMockSampler(sampleFrom, networkHead)
+		testParams.networkHead = uint64(20)
+		ctx, cancel := context.WithTimeout(context.Background(), testParams.timeoutDelay)
 
-		lk := newLock(sampleFrom, networkHead) // lock all workers before start
-		coordinator := newSamplingCoordinator(concurrency, samplingRange, getterStub{},
+		sampler := newMockSampler(testParams.sampleFrom, testParams.networkHead)
+
+		lk := newLock(testParams.sampleFrom, testParams.networkHead) // lock all workers before start
+		coordinator := newSamplingCoordinator(testParams.dasParams, getterStub{},
 			lk.middleWare(sampler.sample))
 		go coordinator.run(ctx, sampler.checkpoint)
 
 		time.Sleep(50 * time.Millisecond)
 		// discover new height and lock it
-		discovered := networkHead + 1
+		discovered := testParams.networkHead + 1
 		lk.add(discovered)
 		sampler.discover(ctx, discovered, coordinator.listen)
 
@@ -151,7 +158,7 @@ func TestCoordinator(t *testing.T) {
 		time.Sleep(100 * time.Millisecond)
 
 		// check that only last header is pending
-		assert.EqualValues(t, int(discovered-sampleFrom), sampler.doneAmount())
+		assert.EqualValues(t, int(discovered-testParams.sampleFrom), sampler.doneAmount())
 		assert.False(t, sampler.heightIsDone(discovered))
 
 		// release all headers for coordinator
@@ -165,27 +172,28 @@ func TestCoordinator(t *testing.T) {
 		assert.Emptyf(t, coordinator.state.failed, "failed list is not empty")
 
 		cancel()
-		stopCtx, cancel := context.WithTimeout(context.Background(), timeoutDelay)
+		stopCtx, cancel := context.WithTimeout(context.Background(), testParams.timeoutDelay)
 		defer cancel()
 		assert.NoError(t, coordinator.wait(stopCtx))
 		assert.Equal(t, sampler.finalState(), newCheckpoint(coordinator.state.unsafeStats()))
 	})
 
 	t.Run("failed should be stored", func(t *testing.T) {
-		sampleFrom := uint64(1)
-		ctx, cancel := context.WithTimeout(context.Background(), timeoutDelay)
+		testParams := defaultTestParams()
+		testParams.sampleFrom = 1
+		ctx, cancel := context.WithTimeout(context.Background(), testParams.timeoutDelay)
 
 		bornToFail := []uint64{4, 8, 15, 16, 23, 42}
-		sampler := newMockSampler(sampleFrom, networkHead, bornToFail...)
+		sampler := newMockSampler(testParams.sampleFrom, testParams.networkHead, bornToFail...)
 
-		coordinator := newSamplingCoordinator(concurrency, samplingRange, getterStub{}, onceMiddleWare(sampler.sample))
+		coordinator := newSamplingCoordinator(testParams.dasParams, getterStub{}, onceMiddleWare(sampler.sample))
 		go coordinator.run(ctx, sampler.checkpoint)
 
 		// wait for coordinator to indicateDone catchup
 		assert.NoError(t, coordinator.state.waitCatchUp(ctx))
 
 		cancel()
-		stopCtx, cancel := context.WithTimeout(context.Background(), timeoutDelay)
+		stopCtx, cancel := context.WithTimeout(context.Background(), testParams.timeoutDelay)
 		defer cancel()
 		assert.NoError(t, coordinator.wait(stopCtx))
 
@@ -198,16 +206,18 @@ func TestCoordinator(t *testing.T) {
 	})
 
 	t.Run("failed should retry on restart", func(t *testing.T) {
-		sampleFrom := uint64(50)
-		ctx, cancel := context.WithTimeout(context.Background(), timeoutDelay)
+		testParams := defaultTestParams()
 
-		failedLastRun := map[uint64]int{4: 1, 8: 2, 15: 1, 16: 1, 23: 1, 42: 1, sampleFrom - 1: 1}
+		testParams.sampleFrom = uint64(50)
+		ctx, cancel := context.WithTimeout(context.Background(), testParams.timeoutDelay)
+
+		failedLastRun := map[uint64]int{4: 1, 8: 2, 15: 1, 16: 1, 23: 1, 42: 1, testParams.sampleFrom - 1: 1}
 		failedAgain := []uint64{16}
 
-		sampler := newMockSampler(sampleFrom, networkHead, failedAgain...)
+		sampler := newMockSampler(testParams.sampleFrom, testParams.networkHead, failedAgain...)
 		sampler.checkpoint.Failed = failedLastRun
 
-		coordinator := newSamplingCoordinator(concurrency, samplingRange, getterStub{}, onceMiddleWare(sampler.sample))
+		coordinator := newSamplingCoordinator(testParams.dasParams, getterStub{}, onceMiddleWare(sampler.sample))
 		go coordinator.run(ctx, sampler.checkpoint)
 
 		// check if all jobs were sampled successfully
@@ -217,7 +227,7 @@ func TestCoordinator(t *testing.T) {
 		assert.NoError(t, coordinator.state.waitCatchUp(ctx))
 
 		cancel()
-		stopCtx, cancel := context.WithTimeout(context.Background(), timeoutDelay)
+		stopCtx, cancel := context.WithTimeout(context.Background(), testParams.timeoutDelay)
 		defer cancel()
 		assert.NoError(t, coordinator.wait(stopCtx))
 
@@ -231,13 +241,15 @@ func TestCoordinator(t *testing.T) {
 }
 
 func BenchmarkCoordinator(b *testing.B) {
-	concurrency := 100
-	samplingRange := uint64(10)
 	timeoutDelay := 5 * time.Second
+
+	params := DefaultParameters()
+	params.SamplingRange = 10
+	params.ConcurrencyLimit = 100
 
 	b.Run("bench run", func(b *testing.B) {
 		ctx, cancel := context.WithTimeout(context.Background(), timeoutDelay)
-		coordinator := newSamplingCoordinator(concurrency, samplingRange, newBenchGetter(),
+		coordinator := newSamplingCoordinator(params, newBenchGetter(),
 			func(ctx context.Context, h *header.ExtendedHeader) error { return nil })
 		go coordinator.run(ctx, checkpoint{
 			SampleFrom:  1,
@@ -520,5 +532,22 @@ func onceMiddleWare(out sampleFn) sampleFn {
 		}
 		m.Unlock()
 		return out(ctx, h)
+	}
+}
+
+type testParams struct {
+	networkHead  uint64
+	sampleFrom   uint64
+	timeoutDelay time.Duration
+	dasParams    Parameters
+}
+
+func defaultTestParams() testParams {
+	dasParamsDefault := DefaultParameters()
+	return testParams{
+		networkHead:  uint64(500),
+		sampleFrom:   dasParamsDefault.SampleFrom,
+		timeoutDelay: 125 * time.Second,
+		dasParams:    dasParamsDefault,
 	}
 }
--- a/das/daser.go
+++ b/das/daser.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"sync/atomic"
-	"time"
 
 	"github.com/ipfs/go-datastore"
 	logging "github.com/ipfs/go-log/v2"
@@ -17,26 +16,10 @@ import (
 
 var log = logging.Logger("das")
 
-// TODO: parameters needs performance testing on real network to define optimal values
-const (
-	//  samplingRange is the maximum amount of headers processed in one job.
-	samplingRange = 100
-
-	// concurrencyLimit defines the maximum amount of sampling workers running in parallel.
-	concurrencyLimit = 16
-
-	// backgroundStoreInterval is the period of time for background checkpointStore to perform a checkpoint backup.
-	backgroundStoreInterval = 10 * time.Minute
-
-	// priorityQueueSize defines the size limit of the priority queue
-	priorityQueueSize = concurrencyLimit * 4
-
-	// genesisHeight is the height sampling will start from
-	genesisHeight = 1
-)
-
 // DASer continuously validates availability of data committed to headers.
 type DASer struct {
+	params Parameters
+
 	da     share.Availability
 	bcast  fraud.Broadcaster
 	hsub   header.Subscriber // listens for new headers in the network
@@ -61,8 +44,10 @@ func NewDASer(
 	getter header.Getter,
 	dstore datastore.Datastore,
 	bcast fraud.Broadcaster,
-) *DASer {
+	options ...Option,
+) (*DASer, error) {
 	d := &DASer{
+		params:         DefaultParameters(),
 		da:             da,
 		bcast:          bcast,
 		hsub:           hsub,
@@ -71,9 +56,18 @@ func NewDASer(
 		subscriber:     newSubscriber(),
 		subscriberDone: make(chan struct{}),
 	}
-	d.sampler = newSamplingCoordinator(concurrencyLimit, samplingRange, getter, d.sample)
 
-	return d
+	for _, applyOpt := range options {
+		applyOpt(d)
+	}
+
+	err := d.params.Validate()
+	if err != nil {
+		return nil, err
+	}
+
+	d.sampler = newSamplingCoordinator(d.params, getter, d.sample)
+	return d, nil
 }
 
 // Start initiates subscription for new ExtendedHeaders and spawns a sampling routine.
@@ -93,8 +87,8 @@ func (d *DASer) Start(ctx context.Context) error {
 		log.Warnw("checkpoint not found, initializing with height 1")
 
 		cp = checkpoint{
-			SampleFrom:  genesisHeight,
-			NetworkHead: genesisHeight,
+			SampleFrom:  d.params.SampleFrom,
+			NetworkHead: d.params.SampleFrom,
 		}
 
 		// attempt to get head info. No need to handle error, later DASer
@@ -110,7 +104,7 @@ func (d *DASer) Start(ctx context.Context) error {
 
 	go d.sampler.run(runCtx, cp)
 	go d.subscriber.run(runCtx, sub, d.sampler.listen)
-	go d.store.runBackgroundStore(runCtx, backgroundStoreInterval, d.sampler.getCheckpoint)
+	go d.store.runBackgroundStore(runCtx, d.params.BackgroundStoreInterval, d.sampler.getCheckpoint)
 
 	return nil
 }
--- a/das/daser_test.go
+++ b/das/daser_test.go
@@ -39,9 +39,10 @@ func TestDASerLifecycle(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	t.Cleanup(cancel)
 
-	daser := NewDASer(avail, sub, mockGet, ds, mockService)
+	daser, err := NewDASer(avail, sub, mockGet, ds, mockService)
+	require.NoError(t, err)
 
-	err := daser.Start(ctx)
+	err = daser.Start(ctx)
 	require.NoError(t, err)
 	defer func() {
 		err = daser.Stop(ctx)
@@ -73,9 +74,10 @@ func TestDASer_Restart(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	t.Cleanup(cancel)
 
-	daser := NewDASer(avail, sub, mockGet, ds, mockService)
+	daser, err := NewDASer(avail, sub, mockGet, ds, mockService)
+	require.NoError(t, err)
 
-	err := daser.Start(ctx)
+	err = daser.Start(ctx)
 	require.NoError(t, err)
 
 	// wait for dasing catch-up routine to indicateDone
@@ -100,7 +102,9 @@ func TestDASer_Restart(t *testing.T) {
 	restartCtx, restartCancel := context.WithTimeout(context.Background(), timeout)
 	t.Cleanup(restartCancel)
 
-	daser = NewDASer(avail, sub, mockGet, ds, mockService)
+	daser, err = NewDASer(avail, sub, mockGet, ds, mockService)
+	require.NoError(t, err)
+
 	err = daser.Start(restartCtx)
 	require.NoError(t, err)
 
@@ -146,7 +150,9 @@ func TestDASer_stopsAfter_BEFP(t *testing.T) {
 	newCtx := context.Background()
 
 	// create and start DASer
-	daser := NewDASer(avail, sub, mockGet, ds, f)
+	daser, dErr := NewDASer(avail, sub, mockGet, ds, f)
+	require.NoError(t, dErr)
+
 	resultCh := make(chan error)
 	go fraud.OnProof(newCtx, f, fraud.BadEncoding,
 		func(fraud.Proof) {
--- a/das/daser_test.go
+++ b/das/daser_test.go
@@ -150,8 +150,8 @@ func TestDASer_stopsAfter_BEFP(t *testing.T) {
 	newCtx := context.Background()
 
 	// create and start DASer
-	daser, dErr := NewDASer(avail, sub, mockGet, ds, f)
-	require.NoError(t, dErr)
+	daser, err := NewDASer(avail, sub, mockGet, ds, f)
+	require.NoError(t, err)
 
 	resultCh := make(chan error)
 	go fraud.OnProof(newCtx, f, fraud.BadEncoding,
--- a/das/options.go
+++ b/das/options.go
@@ -5,18 +5,22 @@ import (
 	"time"
 )
 
-var (
-	ErrInvalidOption = fmt.Errorf("das: invalid option")
-)
+// ErrInvalidOption is an error that is returned by Parameters.Validate
+// when supplied with invalid values.
+// This error will also be returned by NewDASer if supplied with an invalid option
+var ErrInvalidOption = fmt.Errorf("das: invalid option")
 
+// errInvalidOptionValue is a utility function to dedup code for error-returning
+// when dealing with invalid parameter values
 func errInvalidOptionValue(optionName string, value string) error {
 	return fmt.Errorf("%w: value %s cannot be %s", ErrInvalidOption, optionName, value)
 }
 
-// Type Option is the functional option that is applied to the daser instance for parameters configuration.
+// Option is the functional option that is applied to the daser instance
+// to configure DASing parameters (the Parameters struct)
 type Option func(*DASer)
 
-// Type parameters is the set of parameters that must be configured for the daser
+// Parameters is the set of parameters that must be configured for the daser
 type Parameters struct {
 	//  SamplingRange is the maximum amount of headers processed in one job.
 	SamplingRange uint64
@@ -36,7 +40,7 @@ type Parameters struct {
 
 // DefaultParameters returns the default configuration values for the daser parameters
 func DefaultParameters() Parameters {
-	// TODO(@derrandz): parameters needs performance testing on real network to define optimal values
+	// TODO(@derrandz): parameters needs performance testing on real network to define optimal values (#1261)
 	return Parameters{
 		SamplingRange:           100,
 		ConcurrencyLimit:        16,
@@ -48,10 +52,9 @@ func DefaultParameters() Parameters {
 
 // Validate validates the values in Parameters
 //
-// All parameters must be positive and non-zero, except:
-//
-// BackgroundStoreInterval = 0 disables background storer,
-// PriorityQueueSize = 0 disables prioritization of recently produced blocks for sampling
+//	All parameters must be positive and non-zero, except:
+//		BackgroundStoreInterval = 0 disables background storer,
+//		PriorityQueueSize = 0 disables prioritization of recently produced blocks for sampling
 func (p *Parameters) Validate() error {
 	// SamplingRange = 0 will cause the jobs' queue to be empty
 	// Therefore no sampling jobs will be reserved and more importantly the DASer will break
@@ -84,18 +87,18 @@ func (p *Parameters) Validate() error {
 }
 
 // WithSamplingRange is a functional option to configure the daser's `SamplingRange` parameter
-// ```
 //
-//	WithSamplingRange(10)(daser)
-//
-// ```
+//	Usage:
+//	```
+//		WithSamplingRange(10)(daser)
+//	```
 //
 // or
-// ```
 //
-//	option := WithSamplingRange(10)
-//	// shenanigans to create daser
-//	option(daser)
+//	```
+//		option := WithSamplingRange(10)
+//		// shenanigans to create daser
+//		option(daser)
 //
 // ```
 func WithSamplingRange(samplingRange uint64) Option {
--- a/das/options.go
+++ b/das/options.go
@@ -1,0 +1,137 @@
+package das
+
+import (
+	"fmt"
+	"time"
+)
+
+var (
+	ErrInvalidOption = fmt.Errorf("das: invalid option")
+)
+
+func errInvalidOptionValue(optionName string, value string) error {
+	return fmt.Errorf("%w: value %s cannot be %s", ErrInvalidOption, optionName, value)
+}
+
+// Type Option is the functional option that is applied to the daser instance for parameters configuration.
+type Option func(*DASer)
+
+// Type parameters is the set of parameters that must be configured for the daser
+type Parameters struct {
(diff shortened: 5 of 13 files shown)


das/daser.go
```diff
@@ -17,28 +17,73 @@ import (
 	"github.com/celestiaorg/celestia-node/share"
 )
 
+// TODO(@team): Should we DI this?
```
Wondertan, line 20: We use pkg level configs, so no

nodebuilder/config.go
```diff
@@ -27,6 +28,7 @@ type Config struct {
 	RPC    rpc.Config
 	Share  share.Config
 	Header header.Config
+	DASer  das.Config
```
Wondertan, line 31: Thi should be a module level config as per our discussion

das/coordinator_test.go
```diff
@@ -125,13 +134,14 @@ func TestCoordinator(t *testing.T) {
 	})
 
 	t.Run("priority routine should not lock other workers", func(t *testing.T) {
+		params.samplingRange = 20
```
derrandz, line 137: @walldiss tests wouldn't pass until I made this. Can you scrutinize these changes?

walldiss: This line will make test pointless. The point is to have priority routine discover new headers and sample recent headers sooner than catchup workers reach it. I'll point to root of the issue in different comment

das/coordinator_test.go
```diff
@@ -72,11 +78,14 @@ func TestCoordinator(t *testing.T) {
 	})
 
 	t.Run("prioritize newly discovered over known", func(t *testing.T) {
+
+		// params := dasingParams{}
+		params.concurrencyLimit = 1
+		params.samplingRange = 4
+
```
walldiss, line 85: The issue here that previously in tests variables were shadowed inside each test if they had to be changed, meaning no change of the original variable. But now you modify value of `Params` struct  inside test so it affects all tests that depend on it. In case you need to modify config, easy fix could be to make a copy of config first and then modify the copy. Looks like you had an idea of it by introducing shadowing with copy, but commented it out.
Example of solution:
```
t.Run("priority routine should not lock other workers", func(t *testing.T) {
  tParams := params
  tParams. concurrencyLimit = 100
  tParams.samplingRange = 4
  
  .. use tParams 
}
```
```

File: nodebuilder/daser/config.go [hunk 5]
```diff
@@ -0,0 +1,46 @@
...
+	}
+}
+
+// Validate performs basic validation of the config.
+func (cfg *Config) Validate() error {
+	if cfg.BackgroundStoreInterval <= 0 {
+		return ErrNegativeInterval
+	}
+	return nil
+}
```
walldiss, line 46: `BackgroundStoreInterval` == 0 is a valid option that suppose to disable background store I think it would be nice if it is documented in config struct. 
It would be nice to add checks for `SamplingRange`,  `ConcurrencyLimit`, `GenesisHeight` to be > 0.

nodebuilder/daser/config.go
```diff
@@ -0,0 +1,46 @@
...
+	"time"
+)
+
+var (
+	ErrNegativeInterval = errors.New("nodebuilder/daser: interval must be positive")
+)
+
+type Config struct {
+	//  samplingRange is the maximum amount of headers processed in one job.
+	SamplingRange uint64 // TODO(@derrandz): Question to @vlad, why is this uint64?
```
walldiss, line 14: it was convenient to have it as `uint64` constant, because it was compared with other `uitn54` values in code later on. For config it is better to have all values of same type (`uint`)) and convert it internally in daser where appropriate.

das/daser.go
```diff
@@ -63,8 +88,10 @@ func NewDASer(
 	getter header.Getter,
 	dstore datastore.Datastore,
 	bcast fraud.Broadcaster,
+	options ...Option,
 ) *DASer {
 	d := &DASer{
+		params:         dasingParams{},
```
walldiss, line 94: Could we please start from some default config instead of zero-value here so if there some options are missing (they are optional, right?) we don't end up with 0 in some config fields. It could make DASer broken otherwise.

das/state.go
```diff
@@ -6,11 +6,13 @@ import (
 
 // coordinatorState represents the current state of sampling
 type coordinatorState struct {
-	rangeSize uint64
+	genesisHeight uint
+	samplingRange uint // is the maximum amount of headers processed in one job.
```
walldiss, line 10: lets use `uint64` here to avoid multiple conversions on some operations.

das/state.go
```diff
@@ -6,11 +6,13 @@ import (
...
 
-	priority   []job                      // list of headers heights that will be sampled with higher priority
-	inProgress map[int]func() workerState // keeps track of running workers
-	failed     map[uint64]int             // stores heights of failed headers with amount of attempt as value
+	priorityQueueSize uint                       // the size of the priority queue
```
walldiss, line 12: store as `int` -> less convertions

das/daser.go
```diff
@@ -73,7 +100,16 @@ func NewDASer(
...
 		subscriberDone: make(chan struct{}),
 	}
-	d.sampler = newSamplingCoordinator(concurrencyLimit, samplingRange, getter, d.sample)
+
+	for _, applyOpt := range options {
+		applyOpt(d)
+	}
+
+	if len(options) == 0 {
+		log.Warn("das/daser: no options applied, daser parameters will be instantiated with zero-values, make sure to configure the daser")
```
walldiss, line 109: I wonder if user could be scared by this. Like we are pushing him to do some proper set up here, but does he need to? I think it should work fine with just default config.

das/coordinator_test.go
```diff
@@ -14,18 +14,24 @@ import (
...
 func TestCoordinator(t *testing.T) {
-	concurrency := 10
-	samplingRange := uint64(10)
+
+	params := dasingParams{}
+	params.samplingRange = 10
+	params.concurrencyLimit = 10
+	params.bgStoreInterval = 10 * time.Minute
+	params.priorityQueueSize = 16 * 4
+	params.genesisHeight = 1
```
walldiss: nit: we already have struct literals here, so we can use it to set fields.

das/coordinator.go
```diff
@@ -9,9 +9,10 @@ import (
 
 // samplingCoordinator runs and coordinates sampling workers and updates current sampling state
 type samplingCoordinator struct {
-	concurrencyLimit int
-	getter           header.Getter
-	sampleFn         sampleFn
+	concurrencyLimit uint
```
walldiss, line 12: `int` for convenience

walldiss (changes requested):
Looks good to me. Like your functional approach with options to remove dependancy on `nodebuilder`. There are 2 things that I think could be improved:
 - types in structs could be better aligned with their usage to avoid extra conversions. 
 - we need to set up default values inside DASer in case missing options scenario.

 Besides that looks legit.

nodebuilder/daser/config.go
derrandz: These values are already `uint`s and will always be `> 0`.

nodebuilder/daser/config.go
derrandz: I agree about the point of having all types be `uint`. We should also scrutinize in the other variables that you mentioned were `uint64` and revise if they should remain so.

Tagging @Wondertan

das/daser.go
derrandz: Exactly, I hear you. I thought about this.
Noted.

das/state.go
derrandz: I will for the time being, and will also add a `TODO` to revise whether we should keep `uint64` for the variables it's being compared with.

derrandz:
@walldiss I've addressed your comments about typing to do away with conversions in operations, however, I would still like to understand [why are the compared-with variables `uint64`s to begin with](https://github.com/celestiaorg/celestia-node/pull/1225/commits/9ac8dbdcd77811190db3c467bcf9415b386937e1#diff-b1ae42f22e1981440a95dc39c783b02b94e0f13021dc3ee21aeee926db5220c1R84)? variables like `networkHead` and so on.

das/daser.go
```diff
@@ -15,28 +15,63 @@ import (
...
 )
 
+// TODO(@team): Should we DI this?
 var log = logging.Logger("das")
 
-// TODO: parameters needs performance testing on real network to define optimal values
-const (
-	//  samplingRange is the maximum amount of headers processed in one job.
-	samplingRange = 100
+type dasingParams struct {
```
renaynay, line 21:
```suggestion
type params struct {
```

das/daser.go
```diff
@@ -15,28 +15,63 @@ import (
...
+	samplingRange     uint
+	concurrencyLimit  uint
+	bgStoreInterval   time.Duration
+	priorityQueueSize uint
+	genesisHeight     uint
+}
 
-	// concurrencyLimit defines the maximum amount of sampling workers running in parallel.
-	concurrencyLimit = 16
+func defaultDasingParams() dasingParams {
```
renaynay, line 29:
```suggestion
func defaultParams() dasingParams {
```

das/daser.go
```diff
@@ -15,28 +15,63 @@ import (
...
+	}
+}
+
+func WithParamPriorityQueueSize(priorityQueueSize uint) Option {
+	return func(d *DASer) {
+		d.params.priorityQueueSize = priorityQueueSize
+	}
+}
+
+func WithParamGenesisHeight(genesisHeight uint) Option {
```
renaynay, line 65: I don't think these options, nor `Option` belong in the `das/` pkg

This should all live at `nodebuilder/daser` since they're configuration options.

Did you already discuss the placement with @Wondertan ?

nodebuilder/daser/config.go
```diff
@@ -0,0 +1,44 @@
+package daser
+
+import (
+	"errors"
+	"time"
+)
+
+var (
+	ErrNegativeInterval = errors.New("nodebuilder/daser: interval must be positive")
```
renaynay, line 9: unused.

nodebuilder/daser/config.go
```diff
@@ -0,0 +1,44 @@
...
+		ConcurrencyLimit:        16,
+		BackgroundStoreInterval: 10 * time.Minute,
+		PriorityQueueSize:       16 * 4,
+		GenesisHeight:           1,
+	}
+}
+
+// Validate performs basic validation of the config.
+func (cfg *Config) Validate() error {
+	// TODO(team): what should validate for in here? seems like all configuration fields can accept zero-values.
```
renaynay, line 42: you can ensure all fields are > 0

das/daser.go
renaynay, line 29: also we duplicate the defaults between `nodebuilder/daser` and `das`

das/coordinator_test.go
```diff
@@ -13,19 +13,43 @@ import (
...
-	sampleFrom := uint64(genesisHeight)
-	timeoutDelay := 125 * time.Second
+type testParams struct {
+	networkHead  uint64
+	sampleFrom   uint64
+	timeoutDelay time.Duration
+	dasParams    dasingParams
+}
+
+func defaultDasParams() dasingParams {
```
renaynay, line 23: why is this defined again?

das/coordinator_test.go
```diff
@@ -231,13 +266,18 @@ func TestCoordinator(t *testing.T) {
...
-	concurrency := 100
-	samplingRange := uint64(10)
 	timeoutDelay := 5 * time.Second
 
+	params := dasingParams{}
+	params.samplingRange = 10
+	params.concurrencyLimit = 100
+	params.bgStoreInterval = 10 * time.Minute
+	params.priorityQueueSize = 16 * 4
+	params.genesisHeight = 1
```
renaynay, line 276: You can just take the default and only change the fields you want different from the default params

renaynay (changes requested):

das/daser.go
Wondertan: They should be here. These are options for the DASer, not the module.

nodebuilder/daser/config.go
derrandz: they are `unit` so the conditional will always be truthy even if assigned `-1`

das/coordinator_test.go
derrandz: good call, will remove. The original intention of this was `defaultTestDasParams` but we can just use the existing defaults

das/daser.go
Wondertan: The daser lib/pkg has its own defaults. The defaults in the module are specific to our node types.Ffull and light will have different defaults. 
We can make the pkg level defaults be tuned for e.g. lights, and module defaults would just use the defaults from it.

das/daser.go
derrandz: To add to @Wondertan's comment, it is with these `Option`s that the `nodebuilder/daser` will be able to "configure" the `das/daser`.

nodebuilder/daser/config.go
walldiss: They could be 0 and it is not a valid case

nodebuilder/daser/config.go
walldiss: Either of `SamplingRange` == 0 (empty jobs) OR `ConcurrencyLimit` == 0 (no workers) OR `GenesisHeight ` == 0 (can't sample block height 0) will break DASer.
`BackgroundStoreInterval` == 0 disables background storer,
`PriorityQueueSize` == 0 disables prioritisation of recently produced blocks for sampling 

Lets protect user from breaking the DASer by passing 0 to config fields that could not support it.

derrandz:
@renaynay @walldiss @Wondertan 

Following on the thread about configuration from above, specifically this [commit](https://github.com/celestiaorg/celestia-node/pull/1225/commits/3eb8465a4881d626e70382a931f6ee3e497fd657)

I think we should think about the case **when we take the `das` pkg out of the celestia-node repository**. For the moment it makes sense to validate only at the level of the nodebuilder, but since zero values for `SamplingRange` `ConcurrencyLimit` and `GenesisHeight` do break the DASer, I was thinking about extending the functional options pattern to include errors as well:

```
type Option func (*DASer) error

func WithParamX(x uint) Option {
  return func (d *DASer) error {
      if x == 0 {
         return err
      }
      //...
  }
}
```
Such that when the constructor applies these options, it can also panic with these errors or return them and halt the DASer creation.

This is relevant because, although we start off of default parameter values in the `das` DASer, the possibility of changing these values through options opens the `das` DASer to the possibility of becoming broken due to misconfiguration.

Wondertan:
@derrandz, this pattern you've shown is _more_ correct independently of `das` pkg being out and we should stick to it in the daser and other pkg level options

derrandz:
@Wondertan Great then! I've gone ahead and did an iteration of how we can achieve innate functional options validation like discussed above. Waiting on your review.

das/daser.go
```diff
@@ -15,28 +15,152 @@ import (
...
+type Option func(*DASer) error
+type OptionValidator func(value any, optionSetter optionSetter) error
+
+type parameters struct {
+	samplingRange     uint
+	concurrencyLimit  uint
+	bgStoreInterval   time.Duration
+	priorityQueueSize uint
+	genesisHeight     uint
+}
```
Wondertan, line 47: There seems to be a lot of new code for options. Let's move it into the separate file

nodebuilder/daser/config.go
```diff
@@ -0,0 +1,76 @@
...
+import (
+	"errors"
+	"time"
+)
+
+var (
+	ErrNoSamplingRange = errors.New("nodebuilder/daser: misconfiguration error, sampling range cannot be 0")
+	ErrNoConcLimit     = errors.New("nodebuikder/daser: misconfiguration error, concurrency limit cannot be 0")
+	ErrNoGenesisHeight = errors.New("nodebuilder/daser: misconfiguration error, genesis height cannot be 0")
+)
```
Wondertan, line 12: This error does not have to be exported and extracted in var, or do you see them being somehow? Because I don't

nodebuilder/daser/config.go
```diff
@@ -0,0 +1,76 @@
+package daser
+
+import (
+	"errors"
+	"time"
+)
+
+var (
+	ErrNoSamplingRange = errors.New("nodebuilder/daser: misconfiguration error, sampling range cannot be 0")
+	ErrNoConcLimit     = errors.New("nodebuikder/daser: misconfiguration error, concurrency limit cannot be 0")
```
Wondertan, line 10: Typo in the `builder`. But anyway, in some other PR we agreed to call our module pkg as `modp2p`, `modheader`, and `moddaser` instead of `nodebuilder/daser`.

nodebuilder/daser/config.go
```diff
@@ -0,0 +1,76 @@
+package daser
```
Wondertan, line 1: Relic from the past. The pkg is called `daser` while the pkgs defining Daser is called `das`. Let's make them consistent as with any other modules we have.

das/daser.go
```diff
@@ -15,28 +15,152 @@ import (
...
+		bgStoreInterval:   10 * time.Minute,
+		priorityQueueSize: 16 * 4,
+		genesisHeight:     1,
+	}
+}
 
-	// backgroundStoreInterval is the period of time for background checkpointStore to perform a checkpoint backup.
-	backgroundStoreInterval = 10 * time.Minute
+// Option for: samplingRange
+func WithParamSamplingRange(samplingRange uint) Option {
```
Wondertan, line 60: Why do we need to have `Param` prefixed to each option?

das/daser.go
```diff
@@ -15,28 +15,152 @@ import (
...
+//	the value to be set.
+//	example: both d and a are in the "global scope" relatively to the closure, thus no parameter passing is required.
+//	   func() {
+//
+// .       d.params.option = a
+//
+//	}
+type optionSetter func()
+type Option func(*DASer) error
+type OptionValidator func(value any, optionSetter optionSetter) error
```
Wondertan, line 39: TBH, this seems over-engineered though I totally understand where you are coming from. The approach from your [original comment](https://github.com/celestiaorg/celestia-node/pull/1225#issuecomment-1285258589) is much simpler and easier to understand. Basically, it feels to much for simple option setters

derrandz: 🤔 I thought about this a while before commiting to this road. It does feel over-engineered, however the reason I went with `OptionValidator` is because I did not want the `WithOption` setter to become an "un-patterned" function full of senseless if elses... But quit frankly at second thought, it is a bit too much. Reverting to the initial proposition.

das/daser.go
derrandz: Actually, we don't

das/options.go
```diff
@@ -0,0 +1,94 @@
...
+import (
+	"errors"
+	"fmt"
+	"time"
+)
+
+var (
+	ErrInvalidOptionValue = func(optionName string, value string) error {
+		errorMsg := fmt.Sprintf("das/daser: invalid option value: %s, cannot be %s", optionName, value)
+		return errors.New(errorMsg)
```
Wondertan, line 12: Can be simplified to one-liner `fmt.Errorf`

das/options.go
```diff
@@ -0,0 +1,94 @@
+package das
+
+import (
+	"errors"
+	"fmt"
+	"time"
+)
+
+var (
+	ErrInvalidOptionValue = func(optionName string, value string) error {
```
Wondertan, line 10: Don't think this needs to be exported. It's _not_ an Err you can use to compare errors against. (The main goal of exported errors is to be comparable so that any user from the outside can check for it).

We can though defined a base error: `ErrInvalidOption = "das: invalid option"`. 
Then, we can wrap it with `fmt.Errorf("%w: value '%s' cannot be %s")`
This way, any outside user would be able to check for the base error via `errors.Is(err, das.ErrInvalidOption)` if he needs to.

das/options.go
```diff
@@ -0,0 +1,94 @@
...
+		errorMsg := fmt.Sprintf("das/daser: invalid option value: %s, cannot be %s", optionName, value)
```
Wondertan, line 11:
```suggestion
		errorMsg := fmt.Sprintf("das: invalid option value: %s, cannot be %s", optionName, value)
```

nodebuilder/das/config.go
```diff
@@ -0,0 +1,70 @@
...
+	if cfg.SamplingRange == 0 {
+		return errors.New("moddaser: misconfiguration error, sampling range cannot be 0")
+	}
+
+	if cfg.ConcurrencyLimit == 0 {
+		return errors.New("moddaser: misconfiguration error, concurrency limit cannot be 0")
+	}
+
+	if cfg.GenesisHeight == 0 {
+		return errors.New("nodebuilder/daser: misconfiguration error, genesis height cannot be 0")
```
Wondertan, line 66: Missed renaming to `moddaser`

nodebuilder/das/config.go
```diff
@@ -0,0 +1,70 @@
...
+
+	// priorityQueueSize defines the size limit of the priority queue
+	PriorityQueueSize uint
+
+	// genesisHeight is the height sampling will start from
+	GenesisHeight uint
+}
+
+// TODO(@derrandz): parameters needs performance testing on real network to define optimal values
+func DefaultConfig() Config {
```
Wondertan, line 26: note: @derrandz, this is where `node.Type` will be passed eventually to have different Configs for each type

das/daser.go
```diff
@@ -71,7 +56,15 @@ func NewDASer(
 		subscriber:     newSubscriber(),
 		subscriberDone: make(chan struct{}),
 	}
-	d.sampler = newSamplingCoordinator(concurrencyLimit, samplingRange, getter, d.sample)
+
+	for _, applyOpt := range options {
+		err := applyOpt(d)
+		if err != nil {
+			panic(err)
```
Wondertan, line 63: Let's return an error here.

das/options.go
```diff
@@ -0,0 +1,94 @@
...
+		samplingRange:     100,
+		concurrencyLimit:  16,
+		bgStoreInterval:   10 * time.Minute,
+		priorityQueueSize: 16 * 4,
+		genesisHeight:     1,
+	}
+}
+
+// Option for: samplingRange
+func WithSamplingRange(samplingRange uint) Option {
```
Wondertan, line 37: Godoc requires the name of func/var/etc in the beginning of the comment

das/options.go
```diff
@@ -0,0 +1,94 @@
...
+)
+
+var (
+	ErrInvalidOptionValue = func(optionName string, value string) error {
+		errorMsg := fmt.Sprintf("das/daser: invalid option value: %s, cannot be %s", optionName, value)
+		return errors.New(errorMsg)
+	}
+)
+
+type Option func(*DASer) error
```
Wondertan, line 16: godoc

nodebuilder/das/config.go
```diff
@@ -0,0 +1,70 @@
...
+	ConcurrencyLimit uint
+
+	// backgroundStoreInterval is the period of time for background checkpointStore to perform a checkpoint backup.
+	BackgroundStoreInterval time.Duration
+
+	// priorityQueueSize defines the size limit of the priority queue
+	PriorityQueueSize uint
+
+	// genesisHeight is the height sampling will start from
+	GenesisHeight uint
```
Wondertan, line 22: Capitalize comments as the names

nodebuilder/das/config.go
```diff
@@ -0,0 +1,70 @@
...
+//
+// Both of which won't break the DASer
+func (cfg *Config) Validate() error {
+	// TODO(team): what should validate for in here? seems like all cfg.nfiguration fields can accept zero-values.
+	if cfg.SamplingRange == 0 {
+		return errors.New("moddaser: misconfiguration error, sampling range cannot be 0")
+	}
+
+	if cfg.ConcurrencyLimit == 0 {
+		return errors.New("moddaser: misconfiguration error, concurrency limit cannot be 0")
```
Wondertan, line 62:
```suggestion
		return errors.New("moddas: misconfiguration: concurrency limit cannot be 0")
```

nodebuilder/das/config.go
```diff
@@ -0,0 +1,70 @@
...
+// OR GenesisHeight == 0 there is no block height 0 to be sampled thus will break the DASer.
+//
+// # On the other hand
+//
+// BackgroundStoreInterval == 0 disables background storer,
+// PriorityQueueSize == 0 disables prioritization of recently produced blocks for sampling
+//
+// Both of which won't break the DASer
+func (cfg *Config) Validate() error {
+	// TODO(team): what should validate for in here? seems like all cfg.nfiguration fields can accept zero-values.
```
Wondertan, line 56: @walldiss

nodebuilder/das/config.go
```diff
@@ -0,0 +1,70 @@
+package das
+
+import (
+	"errors"
+	"time"
+)
+
+type Config struct {
```
Wondertan, line 8: godoc

nodebuilder/das/config.go
derrandz: I think `return errors.New("moddas misconfiguration: concurrency limit cannot be 0")` is better, wdyt?

nodebuilder/das/config.go
derrandz: Noted.

nodebuilder/das/config.go
derrandz: He answered previously, so resolving.

nodebuilder/das/config.go
derrandz: Keeping track of this in #1261

nodebuilder/das/module.go
```diff
@@ -11,11 +11,31 @@ import (
...
+		),
+	)
+
 	switch tp {
 	case node.Light, node.Full:
 		return fx.Module(
 			"daser",
+			baseComponents,
 			fx.Provide(fx.Annotate(
 				NewDASer,
```
derrandz, line 40: @renaynay I want to double-check, we changed the signature for this constructor to return an error, if this errors out, will fx halt the instantion and `app.Err()` should return the error causing the halt, is this correct?

Wondertan (commented):
Last nits. Also, we should stick to some integer types everywhere to avoid  type conversion in the params and Daser itself

nodebuilder/das/module.go
Wondertan: I am not Rene, but def yes

das/options.go
```diff
@@ -0,0 +1,113 @@
+package das
+
+import (
+	"fmt"
+	"time"
+)
+
+var (
+	ErrInvalidOption      = fmt.Errorf("das: invalid option")
+	ErrInvalidOptionValue = func(optionName string, value string) error {
```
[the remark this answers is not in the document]
Wondertan: Specifically, the func. It also doesn't have to be a var and can be declared as a simple func

nodebuilder/das/config.go
```diff
@@ -0,0 +1,70 @@
...
+//
+// # On the other hand
+//
+// BackgroundStoreInterval == 0 disables background storer,
+// PriorityQueueSize == 0 disables prioritization of recently produced blocks for sampling
+//
+// Both of which won't break the DASer
+func (cfg *Config) Validate() error {
+	if cfg.SamplingRange == 0 {
+		return errors.New("moddas misconfiguration: sampling range cannot be 0")
```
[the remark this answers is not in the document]
derrandz: https://github.com/celestiaorg/celestia-node/pull/1225#discussion_r1001975784

nodebuilder/das/config.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	if cfg.ConcurrencyLimit == 0 {
+		return errors.New("moddas misconfiguration: concurrency limit cannot be 0")
```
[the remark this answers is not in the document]
derrandz: Same here, I think the format we agreed to is more sensible (refer to https://github.com/celestiaorg/celestia-node/pull/1225#discussion_r1001975784)

das/options.go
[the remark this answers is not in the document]
derrandz: We should keep the `ErrInvalidOption` exported to allow for error comparison like mentioned in our previous conversation in https://github.com/celestiaorg/celestia-node/pull/1225#discussion_r1001825643

das/coordinator_test.go
```diff
@@ -13,19 +13,31 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestCoordinator(t *testing.T) {
-	concurrency := 10
-	samplingRange := uint64(10)
-	networkHead := uint64(500)
-	sampleFrom := uint64(genesisHeight)
-	timeoutDelay := 125 * time.Second
+type testParams struct {
```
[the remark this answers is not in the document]
derrandz: Cool 👍🏻 I like it

nodebuilder/das/config.go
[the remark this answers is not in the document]
Wondertan: This is more about the error styling being inconsistent. The GenesisHeight error one

das/options.go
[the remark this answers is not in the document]
Wondertan: Yes, but I specifically mention a func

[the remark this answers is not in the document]
derrandz: Agreed

nodebuilder/das/config.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	if cfg.GenesisHeight == 0 {
+		return errors.New("moddas: misconfiguration: genesis height cannot be 0")
```
Wondertan, line 66: Are we fixing this or not?

das/state.go
```diff
@@ -77,17 +81,18 @@ func (s *coordinatorState) updateHead(last uint64) bool {
 		return false
 	}
 
-	if s.networkHead == genesisHeight {
+	// TODO(@derrandz): question to @walldiss, why is networkHead uint64 to begin with?
```
Wondertan, line 84: TODO to be resolved

das/options.go
```diff
@@ -0,0 +1,114 @@
...
+
+		return nil
+	}
+}
+
+// WithConcurrencyLimit is a functional option to configure the daser's `concurrencyLimit` parameter
+// Refer to WithSamplingRange documentation to see an example of how to use this
+func WithConcurrencyLimit(concurrencyLimit int) Option {
+	return func(d *DASer) error {
+		if concurrencyLimit == 0 {
```
Wondertan, line 72: This check does not check "-" values, while `int` allows them. Also true for other options.

das/options.go
```diff
@@ -0,0 +1,114 @@
...
+			return errInvalidOptionValue(
+				"concurrencyLimit",
+				"0",
```
Wondertan, line 75: This should take an actual `concurrencyLimit` that could be a -

nodebuilder/das/config.go
derrandz: 🤣 we are!

das/state.go
walldiss: It is `uint64` so it is possible to compare it with other variables of same type without conversion.

Wondertan (dismissed):

das/state.go
derrandz: @walldiss I am trying to understand which variable is the one that has to be `unit64` and that caused the rest to become `uint64` to reduce conversion.

Previously in this PR, you mentioned https://github.com/celestiaorg/celestia-node/pull/1225#discussion_r998376041 and https://github.com/celestiaorg/celestia-node/pull/1225#discussion_r998406090 which have lead me to believe that `networkHead` was the original variable that had to be `uint64` (_for some reason_) and thus caused the rest to become so to avoid conversion.

If there is no particular reason to using `uint64` specifically, let's revert to `uint`.

walldiss: Thare are many of them, `uint64` is used for any block heights variables inside DASer.  `uint64` is used over `uint` to avoid usage of uint32 on 32bit platforms.

derrandz: As per our conversation, I will revert everything in daser to uint with the argument that:

No concrete argument to raise precision to uint64 or other.
We only know so far that we need positive values for the concerned variables.
:+1:

nodebuilder/das/config.go
```diff
@@ -0,0 +1,70 @@
...
+	SamplingRange uint
+
+	// ConcurrencyLimit defines the maximum amount of sampling workers running in parallel.
+	ConcurrencyLimit int
+
+	// BackgroundStoreInterval is the period of time for background checkpointStore to perform a checkpoint backup.
+	BackgroundStoreInterval time.Duration
+
+	// PriorityQueueSize defines the size limit of the priority queue
+	PriorityQueueSize int
```
Wondertan, line 20: What about making these a units?

derrandz: For these, since they are used mostly in comparisons against slices lens (_`len(slice) > concurrencyLimit`, similarly for priorityQueueSize_) it makes more sense to keep them as ints to avoid conversions.

das/checkpoint.go
```diff
@@ -5,18 +5,18 @@ import (
 )
 
 type checkpoint struct {
-	SampleFrom  uint64 `json:"sample_from"`
-	NetworkHead uint64 `json:"network_head"`
+	SampleFrom  uint `json:"sample_from"`
```
renaynay, line 8: We don't see the added benefit of this change - it really only bloats the diff for no improvement. Also @walldiss made a point that if we decide to use atomics later, we'd need to revert anyway. Can you please revert this and use `uint64` in the config

das/options.go
```diff
@@ -0,0 +1,114 @@
...
+				"concurrencyLimit",
+				"negative or 0",
+			)
+		}
+		d.params.concurrencyLimit = concurrencyLimit
+		return nil
+	}
+}
+
+// WithBackgroundStoreInterval is a functional option to configure the daser's `backgroundStoreINterval` parameter
```
renaynay, line 83:
```suggestion
// WithBackgroundStoreInterval is a functional option to configure the daser's `backgroundStoreInterval` parameter
```

nodebuilder/das/module.go
```diff
@@ -11,11 +11,31 @@ import (
 	"github.com/celestiaorg/celestia-node/nodebuilder/node"
 )
 
-func ConstructModule(tp node.Type) fx.Option {
+func ConstructModule(tp node.Type, cfg *Config) fx.Option {
+
```
renaynay, line 15:
```suggestion
```

nodebuilder/das/config.go
```diff
@@ -0,0 +1,70 @@
+package das
+
+import (
+	"errors"
+	"time"
+)
+
+// Config contains configuration parameters for the DASer (or DASing process)
+type Config struct {
```
renaynay, line 9: @derrandz can you please export `params` and its fields from `das` package and make a 
`type Config das.Params`  ? Also, we don't technically need this `DefaultConfig()` since the `daser` is taking options rather than a config / config values.

nodebuilder/das/config.go
```diff
@@ -0,0 +1,70 @@
...
+//
+// OR GenesisHeight == 0 there is no block height 0 to be sampled thus will break the DASer.
+//
+// # On the other hand
+//
+// BackgroundStoreInterval == 0 disables background storer,
+// PriorityQueueSize == 0 disables prioritization of recently produced blocks for sampling
+//
+// Both of which won't break the DASer
+func (cfg *Config) Validate() error {
```
renaynay, line 56: We also don't need validation since we won't provide a Config here, it would just point to das.Params

nodebuilder/das/config.go
renaynay: This is with the goal of deduplication of the code
For clarification: we need to keep the optional parameters approach.

das/checkpoint.go
renaynay: For clarity - we're referring to all `uint64` -> `uint` changes

das/options.go
```diff
@@ -0,0 +1,114 @@
...
+// Type Option is the functional option that is applied to the daser instance for parameters configuration.
+type Option func(*DASer) error
+
+// Type parameters is the set of parameters that must be configured for the daser
+type parameters struct {
+	samplingRange     uint
+	concurrencyLimit  int
+	bgStoreInterval   time.Duration
+	priorityQueueSize int
+	genesisHeight     uint
```
renaynay, line 25: let's rename this to `sampleFrom` since `genesis` typically always refers to 1

das/state.go
```diff
@@ -6,32 +6,36 @@ import (
 
 // coordinatorState represents the current state of sampling
 type coordinatorState struct {
-	rangeSize uint64
+	genesisHeight uint
```
renaynay, line 9: can we rename this to `sampleFrom`

das/state.go
```diff
@@ -110,7 +114,8 @@ func (s *coordinatorState) nextJob() (next job, found bool) {
 
 	j := s.newJob(s.next, s.networkHead)
 
-	s.next += s.rangeSize
+	// TODO(@derrandz): same here @walldiss why is next a uint?
```
renaynay, line 117: If resolved, pls remove comment.

das/daser_test.go
```diff
@@ -39,7 +39,8 @@ func TestDASerLifecycle(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	t.Cleanup(cancel)
 
-	daser := NewDASer(avail, sub, mockGet, ds, mockService)
+	daser, dErr := NewDASer(avail, sub, mockGet, ds, mockService)
```
renaynay, line 42: nit but please just use `err`

das/daser.go
```diff
@@ -156,6 +149,7 @@ func (d *DASer) sample(ctx context.Context, h *header.ExtendedHeader) error {
 		var byzantineErr *share.ErrByzantine
 		if errors.As(err, &byzantineErr) {
 			log.Warn("Propagating proof...")
+			// TODO(team): remove type conversion uint64(h.Height) in dedicated PR that deals with the fraud service
```
renaynay, line 152: We cannot do this, so it can be removed

das/daser.go
```diff
@@ -71,9 +56,17 @@ func NewDASer(
...
-	return d
+	for _, applyOpt := range options {
+		err := applyOpt(d)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	d.sampler = newSamplingCoordinator(d.params, getter, d.sample)
+
```
renaynay, line 68:
```suggestion
```

renaynay (changes requested):
Reviewed by @Wondertan @distractedm1nd @walldiss @vgonkivs

das/checkpoint.go
derrandz: Claro 👍🏻

nodebuilder/das/config.go
derrandz: 🤔 hmmm, I see what you are saying, and I agree, but for the record, if we wanted to be very strict with our patterns, here are two arguments of why you would keep this redundant layer of validation:

1. Validation in this level concerns `nodebuilder`, meaning that this is validation for values coming from a file or CLI, irrespective of where it will go (_i.e: the underlying implementation of `das` that will receive this configuration_), so we must keep it to respect the pattern. To drive this point home, we could use a different daser down the line (_we won't but you see my point_)

2. The validation of `das.paramaeters` in the functional options inside of the `das` package was made in an effort to further isolate the package from `nodebuilder` and make it more of a standalone (_per my understand, that's the purpose we are doing this, correct me if I am wrong_) thus, that validation is useful for any code that will invoke those functional parameters. For the case of nodebuilder, it just so happens that it already validates its own configuration, but for other users (_when we separate das into its own go package_) it might not be the case.

nodebuilder/das/config.go
derrandz: Yeah I see what you are saying. This was an attempt to be compliant with the pattern used with the other modules but it's very redundant.

nodebuilder/das/config.go
derrandz: After some thought, we need to keep validation at this level, but change it slightly.

The reason is, we want `das.Config` to reflect in `das.Parameters` of the daser when `das.Config` has non-zero values retrieved from CLI or file or API, this is due to [this piece of code](https://github.com/celestiaorg/celestia-node/pull/1225/files#diff-9eef3a4c09b5618e9b60c229170908fc389f8a28ce84e29e47c82d33b74e32b2R23-R28) that retrieves configuration values and "passes them as options".

However this will cause us to change the [functional options validation ](https://github.com/celestiaorg/celestia-node/pull/1225/files#diff-fa4e1a1d6adb3bac7ebcb4524081bfce3ae7764b2c0a249c8a5655cdcac5b3bdR55-R64) to not return errors and cause the constructor to fail but rather to use default options when validation fails, and log a warning line.

I will hold until we reach a decision on this.

@renaynay @Wondertan

derrandz: I've settled on:

1. Deduping the code by using `type Config das.Parameters`
2. Keeping the `(*Config) Validate` to avoid the impact of removing it that is described in [the second comment](https://github.com/celestiaorg/celestia-node/pull/1225#discussion_r1004365160) based on argumentation from [the first comment](https://github.com/celestiaorg/celestia-node/pull/1225#discussion_r1004333336)

Wondertan: @derrandz, sorry for the confusion. We didn't want to avoid the second validation, but only dedup the code and keep it grouped where it logically should be in `das`. So the point where you settled is exactly what we discussed.

das/coordinator.go
```diff
@@ -9,9 +9,10 @@ import (
 
 // samplingCoordinator runs and coordinates sampling workers and updates current sampling state
 type samplingCoordinator struct {
-	concurrencyLimit int
-	getter           header.Getter
-	sampleFn         sampleFn
+	ConcurrencyLimit int
```
Wondertan, line 12: Why exported?

das/coordinator.go
```diff
@@ -128,7 +129,7 @@ func (sc *samplingCoordinator) getCheckpoint(ctx context.Context) (checkpoint, e
 	return newCheckpoint(stats), nil
 }
 
-// concurrencyLimitReached indicates whether concurrencyLimit has been reached
-func (sc *samplingCoordinator) concurrencyLimitReached() bool {
-	return len(sc.state.inProgress) >= sc.concurrencyLimit
+// ConcurrencyLimitReached indicates whether ConcurrencyLimit has been reached
+func (sc *samplingCoordinator) ConcurrencyLimitReached() bool {
+	return len(sc.state.inProgress) >= sc.ConcurrencyLimit
```
Wondertan, line 134: Why exported? It seems like a relic from IDE find-and-replace

das/state.go
```diff
@@ -6,11 +6,13 @@ import (
 
 // coordinatorState represents the current state of sampling
 type coordinatorState struct {
-	rangeSize uint64
+	SampleFrom    uint64
```
Wondertan, line 9: Again

nodebuilder/das/config.go
```diff
@@ -0,0 +1,50 @@
...
+		return errors.New("moddas misconfiguration: sampling range cannot be negative or 0")
+	}
+
+	if !(cfg.ConcurrencyLimit > 0) {
+		return errors.New("moddas misconfiguration: concurrency limit cannot be negative or 0")
+	}
+
+	if !(cfg.SampleFrom > 0) {
+		return errors.New("moddas misconfiguration: genesis height cannot be negative or 0")
+	}
```
Wondertan, line 47: ___optional___ nit: `cfg.SamplingRange <= 0` is a bit more readable and as per @distractedm1nd using negation should be avoided where possible.

Wondertan, line 47: Move to Params as a method. See https://github.com/celestiaorg/celestia-node/pull/1225/files#r1005327363

nodebuilder/das/config.go
derrandz: Clear, I will do one more iteration on this.

nodebuilder/das/config.go
derrandz: +1 for the nit @distractedm1nd

das/coordinator.go
derrandz: Indeed

nodebuilder/das/config.go
derrandz: @renaynay @Wondertan I think this is brilliant, thanks for the comments

das/coordinator.go
```diff
@@ -128,7 +129,7 @@ func (sc *samplingCoordinator) getCheckpoint(ctx context.Context) (checkpoint, e
 	return newCheckpoint(stats), nil
 }
 
-// concurrencyLimitReached indicates whether concurrencyLimit has been reached
+// ConcurrencyLimitReached indicates whether ConcurrencyLimit has been reached
```
Wondertan, line 132: pls

derrandz: thanks 👍🏻

Wondertan (approved):

das/options.go
```diff
@@ -0,0 +1,142 @@
...
+		return errors.New("sampling range cannot be negative or 0")
+	}
+
+	if p.ConcurrencyLimit <= 0 {
+		return errors.New("concurrency limit cannot be negative or 0")
+	}
+
+	if p.SampleFrom <= 0 {
+		return errors.New("genesis height cannot be negative or 0")
+	}
```
Wondertan, line 62: Let's use the errInvalidOption here

das/daser.go
```diff
@@ -71,9 +56,16 @@ func NewDASer(
 		subscriber:     newSubscriber(),
 		subscriberDone: make(chan struct{}),
 	}
-	d.sampler = newSamplingCoordinator(concurrencyLimit, samplingRange, getter, d.sample)
 
-	return d
+	for _, applyOpt := range options {
+		err := applyOpt(d)
```
Wondertan, line 61: To avoid duplication of errors in options and Validate on Params, let's remove errors from Options and call here `Validate` after applying all the options. This way we use Validate twice on the `moddaser` and inside the Daser constructor here, but we avoid duplication which good. Suggested by @walldiss and I think this is good idea and hope that's the last fix before we can say our configuration pattern is complete

Wondertan (changes requested):

das/daser_test.go
```diff
@@ -73,7 +74,8 @@ func TestDASer_Restart(t *testing.T) {
...
```
renaynay, line 77: nits but everywhere should not be `dErr` but `err`.

das/options.go
```diff
@@ -0,0 +1,142 @@
...
+	"time"
+)
+
+var (
+	ErrInvalidOption = fmt.Errorf("das: invalid option")
+)
+
+func errInvalidOptionValue(optionName string, value string) error {
+	return fmt.Errorf("%w: value %s cannot be %s", ErrInvalidOption, optionName, value)
+}
```
renaynay, line 15: We can remove this once we remove err checking in options

nodebuilder/das/config.go
```diff
@@ -0,0 +1,45 @@
...
+//
+// BackgroundStoreInterval == 0 disables background storer,
+// PriorityQueueSize == 0 disables prioritization of recently produced blocks for sampling
+//
+// Both of which won't break the DASer
+func (cfg *Config) Validate() error {
+	params := das.Parameters(*cfg)
+	err := params.Validate()
+
+	if err != nil {
```
Wondertan, line 40:
```suggestion
	err := (*das.Parameters)(cfg).Validate()
	if err != nil {
```

Wondertan: A small nit/trick/optimization

nodebuilder/das/config.go
```diff
@@ -0,0 +1,45 @@
...
+//	no threads will be assigned to the waiting jobs therefore breaking the DASer
+//
+// OR GenesisHeight == 0 there is no block height 0 to be sampled thus will break the DASer.
+//
+// # On the other hand
+//
+// BackgroundStoreInterval == 0 disables background storer,
+// PriorityQueueSize == 0 disables prioritization of recently produced blocks for sampling
+//
+// Both of which won't break the DASer
```
renaynay, line 35: Can we move this comment to the actual implementation of `Validate` on params? 
Also all of the explanations for the checks can be comments in the code itself.

das/options.go
```diff
@@ -0,0 +1,142 @@
...
+	return Parameters{
+		SamplingRange:           100,
+		ConcurrencyLimit:        16,
+		BackgroundStoreInterval: 10 * time.Minute,
+		PriorityQueueSize:       16 * 4,
+		SampleFrom:              1,
+	}
+}
+
+// Validate validates the valuea in Parameters
```
renaynay, line 50:
```suggestion
// Validate validates the values in Parameters
```

also refer to my comment on das.Config pls 🙏🏻

renaynay:
Also @derrandz can you check all your godocs on the methods you've implemented and make it according to godoc style? 

for example: 

```
// Type Parameters .....
type Parameters struct {
```

should be 

```
// Parameters .....
type Parameters struct {
```

das/options.go
derrandz: It will still be used in `das.Parameters.Validate()`

nodebuilder/das/daser.go
```diff
@@ -15,6 +15,7 @@ func NewDASer(
 	store header.Store,
 	batching datastore.Batching,
 	fraudService fraud.Module,
-) *das.DASer {
-	return das.NewDASer(da, hsub, store, batching, fraudService)
+	options ...das.Option,
+) (*das.DASer, error) {
+	return das.NewDASer(da, hsub, store, batching, fraudService, options...)
 }
```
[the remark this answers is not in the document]
derrandz: If this file is removed, then using `das.DASer` in the [module construction](https://github.com/celestiaorg/celestia-node/pull/1225/files#diff-9eef3a4c09b5618e9b60c229170908fc389f8a28ce84e29e47c82d33b74e32b2L20) breaks DI.

The reason whyis: _Quoting @distractedm1nd_

> its because of the casting

> ```
> store header.Store,
>	batching datastore.Batching,
>	fraudService fraud.Module,
> ```
>```
>  getter header.Getter,
>	dstore datastore.Datastore,
>	bcast fraud.Broadcaster,
>```
> we had a discussion about this before (whether to keep such proxy constructors for casting, or to change the original constructor to use what the DI provides) and we decided that its better to keep this ugliness on the DI so that the constructor can take exactly what is needed

Thus we are keeping this file.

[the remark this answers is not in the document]
Wondertan: Ok, thanks for noting that!

nodebuilder/das/config.go
```diff
@@ -0,0 +1,36 @@
...
+// TODO(@derrandz): Address #1261
+func DefaultConfig() Config {
+	return Config(das.DefaultParameters())
+}
+
+// Validate performs basic validation of the config.
+// Upon encountering an invalid value, Validate returns an error of type: ErrMisConfig
+func (cfg *Config) Validate() error {
+	err := (*das.Parameters)(cfg).Validate()
+
```
Wondertan, line 30:
```suggestion
```

Wondertan (approved):
Thanks for bringing this over the finish line. This PR was a great way for us to develop the configuration pattern for our codebase and get everyone onboarded with it.

das/daser_test.go
```diff
@@ -146,7 +150,9 @@ func TestDASer_stopsAfter_BEFP(t *testing.T) {
 	newCtx := context.Background()
 
 	// create and start DASer
-	daser := NewDASer(avail, sub, mockGet, ds, f)
+	daser, dErr := NewDASer(avail, sub, mockGet, ds, f)
```
renaynay, line 153: dErr 😭

nodebuilder/das/config.go
renaynay: pls @derrandz

nodebuilder/das/config.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"fmt"
+
+	"github.com/celestiaorg/celestia-node/das"
+)
+
+// ErrMisConfig is an error that is returned by (*Config).Validate
+// when supplied with invalid values.
+var ErrMisConfig = fmt.Errorf("moddas misconfiguration")
```
renaynay, line 11: unnecessary error definition as it's returned in one place.

das/daser_test.go
derrandz: ouch, sorry! thought I finished with those

nodebuilder/das/config.go
Wondertan: I think it's a good idea though

Wondertan (dismissed):

renaynay (approved):

walldiss (approved):
Looks amazing, Great work!

nodebuilder/das/config.go
derrandz: yup went with rene's suggestion. At second thought, I should have stuck to this style, because it would allow `nodebuilder` to check for `das.ErrMisConfig` (_`nodebuilder/das` that is_) which is more contextualized and clean than checking for `das.ErrInvalidOption`.

I think we can address this in a separate issue when we make sure we truly want that all proxy modules under `nodebuilder` become full proxies and leak nothing about their external origins.

derrandz: Tracking this in https://github.com/celestiaorg/celestia-node/issues/1289 
Can be closed if irrelevant

Wondertan (approved):

renaynay (approved):

PR #1232, opened 2022-10-14
feat(share): EDSStore scaffolding

### LINKED ISSUE #1007: test: Flakiness TestService_BlackListPeer
```
=== RUN   TestService_BlackListPeer
    service_test.go:162: 
        	Error Trace:	/home/runner/work/celestia-node/celestia-node/fraud/service_test.go:162
        	Error:      	Should be true
        	Test:       	TestService_BlackListPeer
--- FAIL: TestService_BlackListPeer (0.65s)
```

### LINKED ISSUE #1108: share: `EDSStore.Put`
## Context

To write an entire EDS `Put` method is introduced on the `EDSStore`

## Implementation Details

Internally it:
- Opens a file under `Store.Path/DataRoot.Hash` path
- Serializes the EDS into the file via `share.WriteEDS`
- Wraps it with `DAGStore`'s [FileMount](https://github.com/filecoin-project/dagstore/blob/master/mount/file.go#L10)
- Converts `DataRoot`'s hash into the [`shard.Key`](https://github.com/filecoin-project/dagstore/blob/master/shard/key.go#L12)
- Registers the `Mount` as a `Shard` on the `DAGStore`

___NOTE:___ _Registering on the DAGStore populates the top-level index with shares/proofs accessible from stored EDS, which is
out of the scope of the document._

```go
// Put stores the given data square with DataRoot's hash as a key.
//
// The square is verified on the Exchange level, and Put only stores the square trusting it.
// The resulting file stores all the shares and NMT Merkle Proofs of the EDS.
// Additionally, the file gets indexed s.t. Store.Blockstore can access them. 
func (s *EDSStore) Put(context.Context, DataRoot, *rsmt2d.ExtendedDataSquare) error
```

## Refs

Depends on #1105 
Depends on #1107

### LINKED ISSUE #1110: share: `EDSStore.Remove`
## Context

To remove stored EDS `Remove` method is introduced.

## Implementation Details

Internally it:

- Converts `DataRoot`'s hash into the [`shard.Key`](https://github.com/filecoin-project/dagstore/blob/master/shard/key.go#L12)
- Destroys `Shard` via `DAGStore`
  - Internally removes its `Mount` as well
- Removes CARv1 file from disk under `Store.Path/DataRoot.Hash` path

___NOTES:___

- _GC logic on the DAGStore has to be investigated so that Removing is correctly implemented_

```go
// Remove removes EDS from Store by the given DataRoot and cleans up all the indexing.
func (s *EDSStore) Remove(context.Context, DataRoot) error

```

### LINKED ISSUE #1111: share: `EDSStore.Get`
## Context

To read an entire EDS `Get` method is introduced. 

## Implementation Details

Internally it:

- Gets a serialized EDS `io.Reader` via [`EDSStore.GetCAR`](https://github.com/celestiaorg/celestia-node/issues/1109)
- Deserializes the EDS and validates it via `share.ReadEDS`

```go
// Get reads EDS out of Store by given DataRoot.
// 
// It reads only one quadrant(1/4) of the EDS and verifies the integrity of the stored data by recomputing it.
func (s *EDSStore) Get(context.Context, DataRoot) (*rsmt2d.ExtendedDataSquare, error)

```

## Refs

Depends on https://github.com/celestiaorg/celestia-node/issues/1109
Depends on #1107

### LINKED ISSUE #1112: share: `EDSStore.Has`
## Context

To check if EDSStore keeps an EDS `Has` method is introduced.

## Implemenation Details

 Internally it:

- Converts `DataRoot`'s hash into the [`shard.Key`](https://github.com/filecoin-project/dagstore/blob/master/shard/key.go#L12)
- Checks if [`GetShardInfo`](https://github.com/filecoin-project/dagstore/blob/master/dagstore.go#L483) does not return
[ErrShardUnknown](https://github.com/filecoin-project/dagstore/blob/eac7733212fdd7c80be5078659f7450b3956d2a6/dagstore.go#L55)

```go
// Has checks if EDS exists by the given DataRoot.
func (s *EDSStore) Has(context.Context, DataRoot) (bool, error)

```

### LINKED ISSUE #1115: share: `EDSStore.Blockstore`
## Context 

`Blockstore` method returns a [`Blockstore`](https://github.com/ipfs/go-ipfs-blockstore/blob/master/blockstore.go#L33) interface implementation instance, providing random access over share and NMT Merkle proof in every stored EDS. It is required for FNs/BNs to serve DAS requests over the Bitswap and for reading data by namespace.

## Implementation Details

There is a [frozen/un-merged implementation](https://github.com/filecoin-project/dagstore/pull/116) of `Blockstore` over `DAGStore` and CARv2 indexes. _EDIT: Already unfrozen and almost ready to be merged_

___NOTES:___

- _We can either use it(and finish if something is missing for our case) or implement custom optimized for our needs_(#1051)
- _The Blockstore does not store whole Celestia Blocks, but IPFS blocks. We represent Merkle proofs and shares in IPFS
blocks._

```go
// Blockstore returns an IPFS Blockstore providing access to individual shares/nodes of all EDS 
// registered on the Store. NOTE: The Blockstore does not store whole Celestia Blocks but IPFS blocks. 
// We represent `shares` and NMT Merkle proofs as IPFS blocks and IPLD nodes so Bitswap can access those.
func (s *Store) Blockstore() blockstore.Blockstore

```

distractedm1nd:
Based on top of #1158, replaces #1051  

Closes #1007
Closes #1108 
Closes #1115 
Closes #1110 
Closes #1111 
Closes #1112

Would recommend that I open a feature branch for this, and merge into there. This is the scaffolding, but work is very much still required:

[ ] Write tests
[ ] Check if EDSStore.Blockstore has parity with the new default impl's Blockstore
[ ] Add logging


share/eds/blockstore.go
```diff
@@ -0,0 +1,177 @@
...
+	keys, err := bs.store.dgstr.ShardsContainingMultihash(ctx, cid.Hash())
+	if err != nil {
+		return nil, err
+	}
+	if len(keys) > 1 {
+		return nil, ErrMultipleShardsFound
+	}
+
+	// try to fetch from cache
+	shardKey := keys[0]
```
renaynay, line 118: ensure len(keys) is at least 1

share/eds/blockstore.go
```diff
@@ -0,0 +1,177 @@
...
+
+	// wasn't found in cache, so acquire it and add to cache
+	ch := make(chan dagstore.ShardResult, 1)
+	err = bs.store.dgstr.AcquireShard(ctx, shardKey, ch, dagstore.AcquireOpts{})
+	if err != nil {
+		return nil, err
+	}
+	result := <-ch
+
+	blockStore, err := bs.addToBSCache(shardKey, result.Accessor)
```
renaynay, line 132: shouldn't you check result.Error?

renaynay: and then you can just return bs.addToBSCache

share/eds/blockstore.go
```diff
@@ -0,0 +1,177 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return blockStore, err
+}
+
+func (bs *Blockstore) readFromBSCache(shardContainingCid shard.Key) (dagstore.ReadBlockstore, error) {
+	lk := &bs.bsStripedLocks[shardKeyToStriped(shardContainingCid)]
```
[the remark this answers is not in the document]
distractedm1nd: no because shardKeyToStriped returns a byte

share/eds/blockstore.go
```diff
@@ -0,0 +1,177 @@
...
+	lk.Lock()
+	defer lk.Unlock()
+
+	// We've already ensured that the given shard has the cid/multihash we are looking for.
+	val, ok := bs.blockstoreCache.Get(shardContainingCid)
+	if !ok {
+		return nil, errors.New("not found in cache")
+	}
+
+	rbs := val.(*accessorWithBlockstore).bs
```
renaynay, line 151: also check ok?

share/eds/blockstore.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	bs.blockstoreCache.Add(shardContainingCid, &accessorWithBlockstore{
+		bs: blockStore,
+		sa: accessor,
+	})
+	return blockStore, nil
+}
+
+func shardKeyToStriped(sk shard.Key) byte {
```
renaynay, line 175: can you doc this

share/eds/blockstore.go
```diff
@@ -0,0 +1,177 @@
...
+
+// Blockstore implements the blockstore.Blockstore interface on an EDSStore.
+// The lru cache approach is heavily inspired by the open PR filecoin-project/dagstore/116.
+// The main differences to the implementation here are that we do not support multiple shards per key,
+// call GetSize directly on the underlying RO blockstore, and do not throw errors on Put/PutMany.
+// Also, we do not abstract away the blockstore operations.
+type Blockstore struct {
+	store *EDSStore
+
+	bsStripedLocks [256]sync.Mutex
```
renaynay, line 38: can you doc what these are guarding and how they're indexed?

share/eds/blockstore.go
```diff
@@ -0,0 +1,177 @@
...
+func (bs *Blockstore) PutMany(ctx context.Context, blocks []blocks.Block) error {
+	return nil
+}
+
+func (bs *Blockstore) AllKeysChan(ctx context.Context) (<-chan cid.Cid, error) {
+	return nil, ErrUnsupportedOperation
+}
+
+func (bs *Blockstore) HashOnRead(enabled bool) {
+	panic(ErrUnsupportedOperation)
```
renaynay, line 105: why panic here?

share/eds/blockstore.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+	return blockstr.GetSize(ctx, cid)
+}
+
+// Put needs to not return an error because it is called by the exchange
+func (bs *Blockstore) Put(ctx context.Context, block blocks.Block) error {
+	return nil
+}
+
+// PutMany needs to not return an error because it is called by the exchange
```
renaynay, line 95: It's called by the exchange?

renaynay (changes requested):
* can you doc all noops in blockstore and why they're noops (even if it's just a quick sentence and say check ADR for more info)

share/eds/blockstore.go
distractedm1nd: ya, here I mean bitswap, will clarify

share/eds/blockstore.go
distractedm1nd: thanks for catching this, it shouldn't

share/eds/blockstore.go
distractedm1nd: == 1

share/eds/store_test.go
```diff
@@ -0,0 +1,172 @@
...
+func TestEDSStore_GetCARRestrictsToODS(t *testing.T) {
+	t.Skip()
+}
+
+func TestEDSStore_ContainsEmptyRoot(t *testing.T) {
+	t.Skip()
+}
+
+func TestEDSStore_Remove(t *testing.T) {
+	// Should also not be in blockstore? There is a cache there
```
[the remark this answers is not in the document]
distractedm1nd: removing comment cuz not relevant

share/eds/store_test.go
```diff
@@ -0,0 +1,172 @@
...
+	err = edsStore.Put(ctx, dah, eds)
+	require.NoError(t, err)
+
+	stat, err = edsStore.carIdx.StatFullIndex(shard.KeyFromString(dah.String()))
+	require.True(t, stat.Exists)
+	require.NoError(t, err)
+}
+
+// TestEDSStore_GetCAR ensures that the reader returned from GetCAR is capable of reading the CAR header and ODS.
+// NOTE: I haven't been able to come up with a reasonable way to restrict the reader to only read the ODS.
```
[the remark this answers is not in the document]
distractedm1nd: removing comment, but issue linked to #1352

share/eds/store.go
```diff
@@ -0,0 +1,211 @@
...
+
+// Has checks if EDS exists by the given share.Root.
+func (s *EDSStore) Has(ctx context.Context, root share.Root) (bool, error) {
+	key := root.String()
+	info, err := s.dgstr.GetShardInfo(shard.KeyFromString(key))
+	if err == dagstore.ErrShardUnknown {
+		return false, err
+	}
+
+	return true, info.Error
```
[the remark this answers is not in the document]
distractedm1nd: i think it is a bit more nuanced than this because shards have different states than just exists and not exists. for example the dagstore can know that the dagstore exists but is not retrievable (but i think it doesn't apply to FS mount). will look more into this tomorrow

share/eds/store.go
```diff
@@ -0,0 +1,211 @@
...
+	transientsPath = "/transients/"
+)
+
+type EDSStore struct { //nolint:revive
+	dgstr  *dagstore.DAGStore
+	bs     blockstore.Blockstore
+	mounts *mount.Registry
+
+	topIdx index.Inverted
+	carIdx index.FullIndexRepo
```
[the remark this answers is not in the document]
distractedm1nd: they are not different for our purposes, they are only backreferences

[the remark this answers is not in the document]
distractedm1nd: for reference: there was a discussion around whether these should exist on our EDSStore struct, since we can also access them through the `dgstr`, and they are not needed but should be kept for convenience for future use cases (IIRC @Wondertan 's argument)

share/eds/blockstore.go
```diff
@@ -0,0 +1,188 @@
...
+	blocks "github.com/ipfs/go-block-format"
+	"github.com/ipfs/go-cid"
+	blockstore "github.com/ipfs/go-ipfs-blockstore"
+	ipld "github.com/ipfs/go-ipld-format"
+)
+
+var _ blockstore.Blockstore = (*Blockstore)(nil)
+
+var (
+	maxCacheSize            = 100
```
distractedm1nd, line 20: @Wondertan @renaynay what do you think I should do about this?

share/eds/blockstore.go
```diff
@@ -0,0 +1,188 @@
...
+		return blockstr, nil
+	}
+
+	// wasn't found in cache, so acquire it and add to cache
+	ch := make(chan dagstore.ShardResult, 1)
+	err = bs.store.dgstr.AcquireShard(ctx, shardKey, ch, dagstore.AcquireOpts{})
+	if err != nil {
+		return nil, err
+	}
+	result := <-ch
```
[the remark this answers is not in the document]
distractedm1nd: Yes. If we shouldn't, what would be the pattern to fix it?

share/eds/store.go
```diff
@@ -0,0 +1,207 @@
...
+	if err != nil {
+		return err
+	}
+
+	result := <-ch
+	if result.Error != nil {
+		return result.Error
+	}
+
+	dropped, err := s.carIdx.DropFullIndex(shard.KeyFromString(key))
```
Wondertan, line 167: I checked the DAGStores's logic and they really don't drop the full index on Destroy. Good Catch. We need to add this to the ADR.

share/eds/store.go
```diff
@@ -0,0 +1,207 @@
...
+
+// Blockstore returns an IPFS Blockstore providing access to individual shares/nodes of all EDS
+// registered on the Store. NOTE: The Blockstore does not store whole Celestia Blocks but IPFS blocks.
+// We represent `shares` and NMT Merkle proofs as IPFS blocks and IPLD nodes so Bitswap can access those.
+func (s *EDSStore) Blockstore() blockstore.Blockstore {
+	return s.bs
+}
+
+// Remove removes EDS from Store by the given share.Root and cleans up all the indexing.
+func (s *EDSStore) Remove(ctx context.Context, root share.Root) error {
```
Wondertan, line 153: We also must remove the CAR file here.

Wondertan (changes requested):
I deep review of EDStore only. I witsh Blockstore was a separate PR.

General comments for the EDSStore:
* We should wrap errors in each method with meaningful messages via `fmt.Errorf(%w)`
	* We should start a habit of doing this more often. Our codebase lacks this, and usually, it is hard to grasp what happened from the error message and such wrappings should improve this.
* Each result chan read must obey Context. Check one of the comments for more details on that
* `share.Root` is actually a full blown DAHeader. The networking layer we build on top, won't have access to it, as we don't want to send it over the wire. Thus, we have to change it to the `DataHash` or `DAHeader.Hash` result. Initially, I wanted to ask to do the change in this PR, but then decided to do in a subsequent PR after we clarify this in the Storage ADR first. We also need to come up with cool alias name for the DataHash in `share` pkg 😂

share/eds/store.go
```diff
@@ -0,0 +1,207 @@
...
+	if !dropped {
+		log.Warnf("failed to drop index for %s", key)
+	}
+	return err
+}
+
+// Get reads EDS out of Store by given DataRoot.
+//
+// It reads only one quadrant(1/4) of the EDS and verifies the integrity of the stored data by recomputing it.
+func (s *EDSStore) Get(ctx context.Context, root share.Root) (*rsmt2d.ExtendedDataSquare, error) {
```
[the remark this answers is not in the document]
distractedm1nd: How is it implemented differently than in the spec?

share/eds/store.go
distractedm1nd: 👍🏼 lets update the storage ADR with this + EDSStore name change + share.Root stuff once networking ADR is completed

share/eds/store.go
```diff
@@ -0,0 +1,207 @@
...
+	"github.com/celestiaorg/rsmt2d"
+)
+
+const (
+	blocksPath     = "/blocks/"
+	indexPath      = "/index/"
+	transientsPath = "/transients/"
+)
+
+type EDSStore struct { //nolint:revive
```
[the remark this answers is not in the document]
Wondertan: eds.Store

share/eds/blockstore.go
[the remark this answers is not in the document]
renaynay: If we shouldn't, we should wrap the read with a select and wait for ctx.Done otherwise.

share/eds/blockstore.go
renaynay: Given a 30 second blocktime, this would keep almost an hour's worth of blockdata in cache. I guess that's reasonable. I wouldn't want to extend it / make it larger as I'm not sure how necessary that would be. We'd have to have good justification for it. Maybe @Wondertan has a diff idea.

share/eds/store.go
```diff
@@ -0,0 +1,207 @@
...
+
+	fsRepo, err := index.NewFSRepo(basepath + indexPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create index repository: %w", err)
+	}
+
+	invertedRepo := index.NewInverted(ds)
+	dagStore, err := dagstore.NewDAGStore(
+		dagstore.Config{
+			TransientsDir: basepath + transientsPath,
```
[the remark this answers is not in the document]
distractedm1nd: I have to look into it deeper. It would make sense that we don't use them according to the description of the transients directory, but we definitely create transients, and unused transients are actually what get cleaned up during GC.

share/eds/store.go
```diff
@@ -0,0 +1,207 @@
...
+	ch := make(chan dagstore.ShardResult, 1)
+	err = s.dgstr.RegisterShard(ctx, shard.KeyFromString(key), &mount.FSMount{
+		FS:   os.DirFS(s.basepath + blocksPath),
+		Path: key,
+	}, ch, dagstore.RegisterOpts{})
+	if err != nil {
+		return err
+	}
+
+	result := <-ch
```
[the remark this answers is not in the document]

share/eds/store.go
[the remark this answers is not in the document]

share/eds/store.go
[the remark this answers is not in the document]
renaynay: ok so we're keeping topIdx even though we don't really touch it?

share/eds/blockstore.go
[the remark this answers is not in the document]
distractedm1nd: done

share/eds/blockstore.go
renaynay: @distractedm1nd why 256 again? cache size is 100 ? shouldn't this be equivalent to cache size? If so, extract into var and use for both instantiating cache and fixing the size here.

share/eds/blockstore.go
```diff
@@ -0,0 +1,195 @@
...
+
+	// We've already ensured that the given shard has the cid/multihash we are looking for.
+	val, ok := bs.blockstoreCache.Get(shardContainingCid)
+	if !ok {
+		return nil, errors.New("not found in cache")
+	}
+
+	rbs, ok := val.(*accessorWithBlockstore)
+	if !ok {
+		return nil, errors.New("couldn't cast value from cache to accessor with blockstore")
```
renaynay, line 166: can you also return a reflect.TypeOf here?

share/eds/store.go
Wondertan: Do you say that the file we Open during Put gets removed when we Drop the index?

Wondertan: So I double-checked the test, and it does not check if the file was removed.

share/eds/store.go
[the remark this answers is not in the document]
Wondertan: Get calls GetCar internally, which acquires the shard

share/eds/store.go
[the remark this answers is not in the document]
Wondertan: AFAIU, transients are created during fetch of remote mounts

share/eds/blockstore.go
```diff
@@ -0,0 +1,195 @@
...
+
+	blockStore, err := accessor.Blockstore()
+	if err != nil {
+		return nil, err
+	}
+
+	bs.blockstoreCache.Add(shardContainingCid, &accessorWithBlockstore{
+		bs: blockStore,
+		sa: accessor,
+	})
```
walldiss, line 187: Small optimisation here could be to obtain lock after `blockStore` is returned, if lock doesn't need to protect `accessor`.

walldiss (commented):
Looks good, awesome work!

I've added one small optimisation suggestion.

share/eds/store.go
distractedm1nd: Done now, can you check if im doing it correctly?

share/eds/store.go
distractedm1nd: #1382

share/eds/store.go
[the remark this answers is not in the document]
distractedm1nd:
```
func TestTransientExists(t *testing.T) {
	t.Cleanup(cancel)

	edsStore, err := newStore(t)
	require.NoError(t, err)
	err = edsStore.Start(ctx)
	require.NoError(t, err)

	eds, dah := randomEDS(t)
	err = edsStore.Put(ctx, dah, eds)
	assert.NoError(t, err)

	files, err := os.ReadDir(edsStore.basepath + transientsPath)
	fmt.Println(files[0].Name())
}
```

Prints:
```
=== RUN   TestTransientExists
transient-13F9FB2771E0545486363633AE523764EEDE9BE96102AA9BE3733F643D814AC6.complete
--- PASS: TestTransientExists (0.01s)
PASS
```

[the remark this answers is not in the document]
Wondertan: I see

share/eds/store.go
Wondertan: `os.Remove` does the thing 👍🏻

share/eds/store.go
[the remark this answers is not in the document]

share/eds/store.go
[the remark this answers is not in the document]
distractedm1nd: We do use it for testing at least, which counts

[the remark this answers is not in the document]
distractedm1nd: @Wondertan im fine with either way, depends on what you both think. I don't think we need to store it on the struct

distractedm1nd:
Removed blockstore so it can go into a new PR and can be reviewed more thoroughly, this PR is too big rn and the main part of it can get merged

share/eds/store_test.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+	}
+}
+
+func TestEDSStore_GetCARRestrictsToODS(t *testing.T) {
+	t.Skip()
+}
+
+func TestEDSStore_ContainsEmptyRoot(t *testing.T) {
+	t.Skip()
```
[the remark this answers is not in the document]
distractedm1nd: i can remove them, i was originally planning to do these issues in this PR but its an artifact now

share/eds/store.go
```diff
@@ -0,0 +1,231 @@
...
+}
+
+// Start starts the underlying DAGStore.
+func (s *Store) Start(ctx context.Context) error {
+	return s.dgstr.Start(ctx)
+}
+
+// Stop stops the underlying DAGStore.
+func (s *Store) Stop() error {
+	return s.dgstr.Close()
```
Wondertan, line 90: Indexes shouldn't be stopped or stopped by the dagstore, right?

share/eds/store.go
```diff
@@ -0,0 +1,231 @@
...
+	carIdx index.FullIndexRepo
+
+	basepath string
+}
+
+// NewStore creates a new EDS Store under the given basepath and datastore.
+func NewStore(basepath string, ds datastore.Batching) (*Store, error) {
+	err := setupPath(basepath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to setup EDSStore directories: %w", err)
```
Wondertan, line 44:
```suggestion
		return nil, fmt.Errorf("failed to setup eds.Store directories: %w", err)
```

Wondertan (approved):
Awesome. Thanks man

share/eds/store.go
distractedm1nd: The index interfaces do not have Start/Stop, and a dagstore.Close just finishes ongoing writes and cancels the ctx

Wondertan (approved):

share/eds/store.go
```diff
@@ -0,0 +1,231 @@
...
+func setupPath(basepath string) error {
+	err := os.Mkdir(basepath+blocksPath, 0755)
+	if err != nil {
+		return fmt.Errorf("failed to create blocks directory: %w", err)
+	}
+	err = os.Mkdir(basepath+transientsPath, 0755)
+	if err != nil {
+		return fmt.Errorf("failed to create transients directory: %w", err)
+	}
+	err = os.Mkdir(basepath+indexPath, 0755)
```
Wondertan, line 226: optional nit: perms into a var?

share/eds/store.go
```diff
@@ -0,0 +1,231 @@
...
+
+	s := &Store{
+		basepath: basepath,
+		dgstr:    dagStore,
+		topIdx:   invertedRepo,
+		carIdx:   fsRepo,
+		mounts:   r,
+	}
+
+	return s, nil
```
Wondertan, line 80: optional
```suggestion
	rerturn &Store{
		basepath: basepath,
		dgstr:    dagStore,
		topIdx:   invertedRepo,
		carIdx:   fsRepo,
		mounts:   r,
	}, nil
```

Wondertan (approved):

renaynay (approved):
Looks good, i've already reviewed this and i guess blockstore has been removed so no real comments other than re topIdx

Issue #1100, opened 2022-09-15
fraud: Move BEFP definitions to `share` pkg


Wondertan:
The `fraud` pkg is designed as a generic library that encompasses p2p logic for any Fraud Proof. It allows any pkg depending on it to define the custom implementation of a Fraud Proof and register it on the `fraud` pkg. The pkg itself should not contain any specific implementation of the Proofs(besides testing), while actual users should keep them. 

In our case, the BEFP Fraud Proof currently lays in the `fraud` pkg and has to be grouped logically with `share`. BEFP is proof of bad encoding of shares. 

Depends on #878

PR #1241, opened 2022-10-18
refactor(share): move bad encoding fraud proof to share pkg

walldiss:
Resolves https://github.com/celestiaorg/celestia-node/issues/1100


share/eds/byzantine/share_proof_test.go
```diff
@@ -0,0 +1,81 @@
+package byzantine
```
Wondertan, line 1: Can you pls use `git mv` or similar to make the diff shorter so it's shown as a file move and not as a new file

P.S. I just realized that we should have done this for all previous refactorings because there it overwrites all the history.

Wondertan: Pls also check this for other files as well

Wondertan (changes requested):

share/eds/byzantine/share_proof_test.go
walldiss: Initially I got super excited about your suggestion, but seems it won't do the magic. According to [this discussion](https://stackoverflow.com/questions/1094269/whats-the-purpose-of-git-mv), the `git mv` command is just a shorthand for deleting and re-adding. Still, history is accessible  by `git blame`  as long as the add+delete happens in the same commit.

It would be nice to make git track changes as moved files instead of new lines.

das/daser.go
```diff
@@ -7,6 +7,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/celestiaorg/celestia-node/share/eds/byzantine"
```
Wondertan, line 10: Sort imports.

share/eds/byzantine/pb/share.proto
```diff
@@ -0,0 +1,30 @@
...
+  COL = 1;
+}
+
+message BadEncoding {
+  bytes HeaderHash = 1;
+  uint64 Height   = 2;
+  repeated Share Shares = 3;
+  uint32 Index    = 4;
+  axis Axis    = 5;
+}
```
Wondertan, line 30: New line

Wondertan, line 30: Would love to use [buf](https://github.com/bufbuild/buf) at some point to lint

share/eds/byzantine/share_proof.go
```diff
@@ -0,0 +1,105 @@
...
+	logging "github.com/ipfs/go-log/v2"
+	"github.com/minio/sha256-simd"
+
+	"github.com/celestiaorg/celestia-node/share"
+	"github.com/celestiaorg/celestia-node/share/eds/byzantine/pb"
+	"github.com/celestiaorg/celestia-node/share/ipld"
+	"github.com/celestiaorg/nmt"
+)
+
+var log = logging.Logger("share/validation")
```
Wondertan, line 17: share/byzantine?

share/eds/byzantine/pb/share.proto
```diff
@@ -0,0 +1,30 @@
+syntax = "proto3";
+
+package share.eds.byzantine.pb;
+
+option go_package="github.com/celestiaorg/celestia-node/share/eds/byzantine/pb";
```
Wondertan, line 5: Do we need it?

share/eds/byzantine/pb/share.proto
walldiss: Yeah, I love buf and I think we can switch to it for lint and generation, but in another PR

das/daser.go
walldiss: done

share/eds/byzantine/share_proof.go
walldiss: i wish this could be automated :)

share/eds/byzantine/pb/share.proto
walldiss: No really, since no more imports if this proto. Removed it

Wondertan (approved):

vgonkivs (approved):

renaynay (approved):

PR #1242, opened 2022-10-18
cmd/celestia: Extract common parse funcs into single function

renaynay:
Self explanatory.


renaynay:
Closing as some files missing, extracting into issue instead.

PR #1244, opened 2022-10-19
refactor(share/ipld): Use consts for node type determination

renaynay:
Self-explanatory

### DIFF
--- a/share/ipld/namespace_hasher.go
+++ b/share/ipld/namespace_hasher.go
@@ -7,7 +7,6 @@ import (
 	"github.com/minio/sha256-simd"
 	mhcore "github.com/multiformats/go-multihash/core"
 
-	"github.com/celestiaorg/celestia-app/pkg/appconsts"
 	"github.com/celestiaorg/nmt"
 )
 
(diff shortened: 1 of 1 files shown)


share/ipld/namespace_hasher.go
```diff
@@ -41,8 +40,7 @@ func (n *namespaceHasher) Write(data []byte) (int, error) {
 		return 0, fmt.Errorf("ipld: only one write to hasher is allowed")
 	}
 
-	ln, nln, hln := len(data), int(n.NamespaceLen), n.Hash.Size()
-	innerNodeSize, leafNodeSize := (nln*2+hln)*2, nln+appconsts.ShareSize
+	ln := len(data)
 	switch ln {
```
Wondertan, line 44:
```suggestion
	switch len(data) {
```

renaynay: we return ln in error so no

Wondertan: Then we might also want to use it on line 56

renaynay: done

Wondertan (approved):

Bidon15 (approved):
🚢

walldiss (approved):

MSevey (approved):

Issue #1246, opened 2022-10-19
Can't submit pfd


Bidon15:
### Celestia Node version

b8e25216983f1790e7c22e00218fb21cfecfd64d

### OS

macOS 12.0.1

### Steps to reproduce it

1. Request balances 
2. Start the light node
3. Submit pfd

### Expected result

code 0 -> data is included in the block

### Actual result

not possible to submit data

### Relevant log output

```Shell
curl -X POST -d '{"namespace_id": "0c204d39600fddd3",                                                                                                  
  "data": "f1f20ca8007e910a3bf8b2e61da0f26bca07ef78717a6ea54165f5",
  "gas_limit": 70000}' http://localhost:26658/submit_pfd
"not possible to submit data"%
```

PR #1247, opened 2022-10-19
fix(state): `CoreAccessor` manages its own context

renaynay:
Resolves https://github.com/celestiaorg/celestia-node/issues/1248 and https://github.com/celestiaorg/celestia-node/issues/1246

### DIFF
--- a/state/core_access.go
+++ b/state/core_access.go
@@ -109,10 +109,7 @@ func (ca *CoreAccessor) Stop(context.Context) error {
 		log.Warn("no connection found to close")
 		return nil
 	}
-	defer func() {
-		ca.cancel()
-		ca.cancel = nil
-	}()
+	defer ca.cancelCtx()
 
 	// close out core connection
 	err := ca.coreConn.Close()
@@ -123,6 +120,11 @@ func (ca *CoreAccessor) Stop(context.Context) error {
 	ca.coreConn = nil
 	ca.queryCli = nil
 	return nil
+}
+
+func (ca *CoreAccessor) cancelCtx() {
+	ca.cancel()
+	ca.cancel = nil
 }
 
 func (ca *CoreAccessor) constructSignedTx(
--- a/state/core_access.go
+++ b/state/core_access.go
@@ -73,10 +73,11 @@ func NewCoreAccessor(
 }
 
 func (ca *CoreAccessor) Start(ctx context.Context) error {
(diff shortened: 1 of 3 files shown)


Wondertan (dismissed):
For this to be tested. We need either a Swamp test that sends PFDs or a per-module API testing. The latter would construct the API/Module, and test each method on the API

state/core_access.go
```diff
@@ -100,16 +101,25 @@ func (ca *CoreAccessor) Start(ctx context.Context) error {
...
+		log.Warn("core accessor already stopped")
+		return nil
+	}
 	if ca.coreConn == nil {
 		log.Warn("no connection found to close")
 		return nil
 	}
+	defer func() {
+		ca.cancel()
+		ca.cancel = nil
```
MSevey, line 114: [non-blocking][recommendation]
When there are multiple steps to something like this, I think it is nice to have a helper function to ensure that, if we need to call cancel in other places, we don't miss a step.
```go
func (ca *CoreAccessor) cancelCTX() {
    ca.cancel()
    ca.cancel = nil
}
```

distractedm1nd (approved):

state/core_access.go
renaynay: @MSevey done.

Wondertan (dismissed):

renaynay:
> For this to be tested. We need either a Swamp test that sends PFDs or a per-module API testing. The latter would construct the API/Module, and test each method on the API

we have testground tests that incorporate PFD submission cc @Bidon15

Wondertan:
> we have testground tests that incorporate PFD submission cc @Bidon15

True, the problem is that it didn't catch it

renaynay:
@Wondertan I think they *did* catch it because that's how @Bidon15 found the error

renaynay:
Actually it was found via testing directly on arabica, I stand corrected

state/core_access.go
```diff
@@ -100,21 +101,32 @@ func (ca *CoreAccessor) Start(ctx context.Context) error {
 }
 
 func (ca *CoreAccessor) Stop(context.Context) error {
-	defer ca.cancel()
+	if ca.cancel == nil {
```
MSevey, line 104: a new condition should result in a new/updated test :-)

Bidon15 (approved):

Wondertan (approved):

PR #1253, opened 2022-10-19
chore: gen less acounts to reduce test execution times

Wondertan:
Before: PASS: TestFraudProofSyncing (7.57s)

After: PASS: TestFraudProofSyncing (5.93s)

@evan-forbes, any clue why it takes so long to generate accounts?


evan-forbes:
> any clue why it takes so long to generate accounts?

idk, nothing sticks out as super obvious. Its doing a fair bit of serialization to and from json, then into the state db.

reducing the account generation from 100 to 10 saves about ~100ms each time I spin up a new testnode when testing locally.

distractedm1nd (approved):

evan-forbes (approved):

renaynay (approved):

PR #1256, opened 2022-10-20
feat(ipld): integration test for namespace hasher

### LINKED ISSUE #1249: testing: Integration testing for nmtHasher
Related: #1245 

There are a few panics possible from the nmtHasher. We need to mock Bitswap in order to test giving it garbage data without things panicking.

From @Wondertan :
>The test should include two Bitswap nodes. One requestor and one responder. The responder should mock away Blockstore with a custom bogus implementation returning data that does not match CIDs. Those CIDs should be generated from RandEDS, and the responder should know them in advance to know how to return corrupted data.
[The DagNet](https://github.com/celestiaorg/celestia-node/blob/main/share/availability/test/testing.go#L58) test suite can be helpful here. You can either extend it for the use case or make a new single-purpose one inspired by DagNet.

distractedm1nd:
Closes #1249

### DIFF
--- a/share/availability/full/reconstruction_test.go
+++ b/share/availability/full/reconstruction_test.go
@@ -47,7 +47,7 @@ func TestShareAvailable_OneFullNode(t *testing.T) {
 		return full.SharesAvailable(errCtx, root)
 	})
 
-	lights := make([]*availability_test.Node, lightNodes)
+	lights := make([]*availability_test.TestNode, lightNodes)
 	for i := 0; i < len(lights); i++ {
 		lights[i] = light.Node(net)
 		go func(i int) {
@@ -112,7 +112,9 @@ func TestShareAvailable_ConnectedFullNodes(t *testing.T) {
 	})
 
 	// create light nodes and start sampling for them immediately
-	lights1, lights2 := make([]*availability_test.Node, lightNodes/2), make([]*availability_test.Node, lightNodes/2)
+	lights1, lights2 := make(
+		[]*availability_test.TestNode, lightNodes/2),
+		make([]*availability_test.TestNode, lightNodes/2)
 	for i := 0; i < len(lights1); i++ {
 		lights1[i] = light.Node(net)
 		go func(i int) {
@@ -208,7 +210,9 @@ func TestShareAvailable_DisconnectedFullNodes(t *testing.T) {
 	})
 
 	// create light nodes and start sampling for them immediately
-	lights1, lights2 := make([]*availability_test.Node, lightNodes/2), make([]*availability_test.Node, lightNodes/2)
+	lights1, lights2 := make(
+		[]*availability_test.TestNode, lightNodes/2),
+		make([]*availability_test.TestNode, lightNodes/2)
 	for i := 0; i < len(lights1); i++ {
 		lights1[i] = light.Node(net)
 		go func(i int) {
--- a/share/availability/test/corrupt_data.go
+++ b/share/availability/test/corrupt_data.go
@@ -13,7 +13,7 @@ import (
 	blockstore "github.com/ipfs/go-ipfs-blockstore"
 )
 
-var _ blockstore.Blockstore = (*MockBlockstore)(nil)
+var _ blockstore.Blockstore = (*FraudulentBlockstore)(nil)
 
 // CorruptBlock is a block where the cid doesn't match the data. It fulfills the blocks.Block interface.
 type CorruptBlock struct {
@@ -46,41 +46,41 @@ func NewCorruptBlock(data []byte, fakeCID cid.Cid) *CorruptBlock {
 	}
 }
 
-// MockBlockstore is a mock blockstore.Blockstore that saves both corrupted and original data for every block it
-// receives. If MockBlockstore.Attacking is true, it will serve the corrupted data on requests.
-type MockBlockstore struct {
+// FraudulentBlockstore is a mock blockstore.Blockstore that saves both corrupted and original data for every block it
+// receives. If FraudulentBlockstore.Attacking is true, it will serve the corrupted data on requests.
+type FraudulentBlockstore struct {
 	ds.Datastore
 	Attacking bool
 }
 
-func (m MockBlockstore) Has(ctx context.Context, cid cid.Cid) (bool, error) {
+func (fb FraudulentBlockstore) Has(ctx context.Context, cid cid.Cid) (bool, error) {
 	return false, nil
 }
 
-func (m MockBlockstore) Get(ctx context.Context, cid cid.Cid) (blocks.Block, error) {
+func (fb FraudulentBlockstore) Get(ctx context.Context, cid cid.Cid) (blocks.Block, error) {
 	key := cid.String()
-	if m.Attacking {
+	if fb.Attacking {
 		key = "corrupt" + key
 	}
 
-	data, err := m.Datastore.Get(ctx, ds.NewKey(key))
+	data, err := fb.Datastore.Get(ctx, ds.NewKey(key))
 	if err != nil {
 		return nil, err
 	}
 	return NewCorruptBlock(data, cid), nil
 }
 
-func (m MockBlockstore) GetSize(ctx context.Context, cid cid.Cid) (int, error) {
+func (fb FraudulentBlockstore) GetSize(ctx context.Context, cid cid.Cid) (int, error) {
 	key := cid.String()
-	if m.Attacking {
+	if fb.Attacking {
 		key = "corrupt" + key
 	}
 
-	return m.Datastore.GetSize(ctx, ds.NewKey(key))
+	return fb.Datastore.GetSize(ctx, ds.NewKey(key))
 }
 
-func (m MockBlockstore) Put(ctx context.Context, block blocks.Block) error {
-	err := m.Datastore.Put(ctx, ds.NewKey(block.Cid().String()), block.RawData())
+func (fb FraudulentBlockstore) Put(ctx context.Context, block blocks.Block) error {
+	err := fb.Datastore.Put(ctx, ds.NewKey(block.Cid().String()), block.RawData())
 	if err != nil {
 		return err
 	}
@@ -88,12 +88,12 @@ func (m MockBlockstore) Put(ctx context.Context, block blocks.Block) error {
 	// create data that doesn't match the CID with arbitrary lengths between 0 and len(block.RawData())*2
 	corrupted := make([]byte, mrand.Int()%(len(block.RawData())*2))
 	mrand.Read(corrupted)
-	return m.Datastore.Put(ctx, ds.NewKey("corrupt"+block.Cid().String()), corrupted)
+	return fb.Datastore.Put(ctx, ds.NewKey("corrupt"+block.Cid().String()), corrupted)
 }
 
-func (m MockBlockstore) PutMany(ctx context.Context, blocks []blocks.Block) error {
+func (fb FraudulentBlockstore) PutMany(ctx context.Context, blocks []blocks.Block) error {
 	for _, b := range blocks {
-		err := m.Put(ctx, b)
+		err := fb.Put(ctx, b)
 		if err != nil {
 			return err
 		}
@@ -101,26 +101,23 @@ func (m MockBlockstore) PutMany(ctx context.Context, blocks []blocks.Block) erro
 	return nil
 }
 
-func (m MockBlockstore) DeleteBlock(ctx context.Context, cid cid.Cid) error {
-	// TODO implement me
+func (fb FraudulentBlockstore) DeleteBlock(ctx context.Context, cid cid.Cid) error {
 	panic("implement me")
 }
 
-func (m MockBlockstore) AllKeysChan(ctx context.Context) (<-chan cid.Cid, error) {
-	// TODO implement me
+func (fb FraudulentBlockstore) AllKeysChan(ctx context.Context) (<-chan cid.Cid, error) {
 	panic("implement me")
 }
 
-func (m MockBlockstore) HashOnRead(enabled bool) {
-	// TODO implement me
+func (fb FraudulentBlockstore) HashOnRead(enabled bool) {
 	panic("implement me")
 }
 
-// MockNode creates a TestNode that uses a MockBlockstore to simulate serving corrupted data.
-func MockNode(t *testing.T, net *TestDagNet) (*TestNode, *MockBlockstore) {
+// MockNode creates a TestNode that uses a FraudulentBlockstore to simulate serving corrupted data.
+func MockNode(t *testing.T, net *TestDagNet) (*TestNode, *FraudulentBlockstore) {
 	t.Helper()
 	dstore := dssync.MutexWrap(ds.NewMapDatastore())
-	mockBS := &MockBlockstore{
+	mockBS := &FraudulentBlockstore{
 		Datastore: dstore,
 		Attacking: false,
 	}
--- a/share/availability/test/corrupt_data.go
+++ b/share/availability/test/corrupt_data.go
@@ -1,28 +1,22 @@
-package ipld_test
+package availability_test
 
 import (
 	"context"
 	"fmt"
-	"testing"
-	"time"
-
 	mrand "math/rand"
+	"testing"
 
 	blocks "github.com/ipfs/go-block-format"
 	"github.com/ipfs/go-cid"
 	ds "github.com/ipfs/go-datastore"
 	dssync "github.com/ipfs/go-datastore/sync"
 	blockstore "github.com/ipfs/go-ipfs-blockstore"
-	"github.com/stretchr/testify/require"
-
-	"github.com/celestiaorg/celestia-node/share/availability/full"
-	availability_test "github.com/celestiaorg/celestia-node/share/availability/test"
-	"github.com/celestiaorg/celestia-node/share/service"
 )
 
 var _ blockstore.Blockstore = (*MockBlockstore)(nil)
 
 // CorruptBlock is a block where the cid doesn't match the data.
+// It fulfills the blocks.Block interface.
 type CorruptBlock struct {
 	cid  cid.Cid
 	data []byte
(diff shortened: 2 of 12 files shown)


share/ipld/nmt_hasher_integration_test.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+func NewCorruptBlock(data []byte, fakeCID cid.Cid) *CorruptBlock {
+	return &CorruptBlock{
+		fakeCID,
+		data,
+	}
+}
+
+type MockBlockstore struct {
```
MSevey, line 56: [non-blocking] Same here, a simple comment about the interface it is mocking and any changes

share/ipld/nmt_hasher_integration_test.go
```diff
@@ -0,0 +1,162 @@
...
+	dssync "github.com/ipfs/go-datastore/sync"
+	blockstore "github.com/ipfs/go-ipfs-blockstore"
+	"github.com/stretchr/testify/require"
+
+	"github.com/celestiaorg/celestia-node/share/availability/full"
+	availability_test "github.com/celestiaorg/celestia-node/share/availability/test"
+	"github.com/celestiaorg/celestia-node/share/service"
+)
+
+var _ blockstore.Blockstore = (*MockBlockstore)(nil)
```
MSevey, line 23: [non-blocking][suggestion]

There is a lot of template code for the mocks, might be cleaner to have that in a separate file since I expect more tests to be added to this file over time.

share/ipld/nmt_hasher_integration_test.go
```diff
@@ -0,0 +1,162 @@
...
+	mockBS := &MockBlockstore{
+		Datastore: dstore,
+		attacking: false,
+	}
+	provider := net.NodeWithBlockStore(dstore, mockBS)
+	provider.ShareService = service.NewShareService(provider.BlockService, full.TestAvailability(provider.BlockService))
+	return provider, mockBS
+}
+
+// TestNamespaceHasher_CorruptedData is an integration test that
```
MSevey, line 136: nit: your line wraps seem non-unified, you should be able to either use a code editor setting to auto line wrap or if you are a vim fan, `gq`

share/ipld/nmt_hasher_integration_test.go
```diff
@@ -0,0 +1,162 @@
...
+	// TODO implement me
+	panic("implement me")
+}
+
+func (m MockBlockstore) HashOnRead(enabled bool) {
+	// TODO implement me
+	panic("implement me")
+}
+
+func MockNode(t *testing.T, net *availability_test.DagNet) (*availability_test.Node, *MockBlockstore) {
```
MSevey, line 124: does this need to be exported? If not, I'd recommend not exporting test methods by default.

share/availability/test/testing.go
```diff
@@ -99,6 +96,13 @@ func (dn *DagNet) Node() *Node {
 	return nd
 }
 
+// Node create a plain network node that can serve and request data.
+func (dn *DagNet) Node() *Node {
```
MSevey, line 100: [non-blocking][suggestion]

To be explicit and avoid any potential confusion, I like prepending `Test` to methods like this to ensure we don't accidentally call it from non Test purposes.

share/availability/test/testing.go
```diff
@@ -72,12 +72,9 @@ func NewTestDAGNet(ctx context.Context, t *testing.T) *DagNet {
 	}
 }
 
-// Node create a plain network node that can serve and request data.
-func (dn *DagNet) Node() *Node {
+func (dn *DagNet) NodeWithBlockStore(dstore ds.Datastore, bstore blockstore.Blockstore) *Node {
```
MSevey, line 75: Does this need to be exported? If not, I'd recommend non-exported test methods by default.

MSevey (changes requested):

share/ipld/nmt_hasher_integration_test.go
distractedm1nd: It doesn't need to be, I will unexport it. @Wondertan How does exporting from _test.go work? Does it actually get exported + end up in the binary?

share/availability/test/testing.go
distractedm1nd: We cannot import these methods from other packages, since it is a _test.go and has its own package - right @Wondertan ?

share/availability/test/testing.go
distractedm1nd: It does need to be exported because it comes from a different package (and not a _test.go file, so exporting works)

share/ipld/nmt_hasher_integration_test.go
distractedm1nd: https://intellij-support.jetbrains.com/hc/en-us/community/posts/115000766224-gq-not-working-in-IdeaVim
down so bad

thanks for bringing this up, I was confused about how its done everywhere else. The comment line length seems to differ from the code line length, so I always just broke it up arbitrarily 😬

share/availability/test/testing.go
renaynay: @distractedm1nd I believe this is only true for exported methods in `_test.go`, but not true for methods exported inside a `test` pkg. I think go ignores `testdata` 

Not 100% sure re `test` pkg.

share/ipld/nmt_hasher_integration_test.go
```diff
@@ -0,0 +1,162 @@
...
+	err := requestor.SharesAvailable(getCtx, root)
+	require.NoError(t, err)
+
+	// clear the storage of the requestor so that it must retrieve again, then start attacking
+	requestor.ClearStorage()
+	mockBS.attacking = true
+	getCtx, cancelGet = context.WithTimeout(ctx, 2*time.Second)
+	t.Cleanup(cancelGet)
+	err = requestor.SharesAvailable(getCtx, root)
+	require.Error(t, err, "availability vailidation failed")
```
renaynay, line 161: Are you intending to check for ErrorIs here? 

```suggestion
	require.Error(t, err, "availability validation failed")
```

renaynay (commented):
* can we call `nmt_hasher_test.go` and we can just, in the docs for `TestNamespaceHasher_CorruptedData`, explain that it's an integration test, etc. This will get run as part of unit testing though (unless we put it in swamp tests)

* another nit, I prefer to have test functions first and all mocks used in those test functions defined below

* additionally noticed `SubNetNode` is dangling

share/ipld/nmt_hasher_integration_test.go
Wondertan: `_test.go` can only export from each other. You cannot import anything defined in a test file from a non-test file. Same as you cannot import anything from the `main` pkg.

I do agree, though, that in this case is better to keep the test utility lowercased for it to be perceived as unexported

share/availability/test/testing.go
Wondertan: > since it is a _test.go and has its own package - right 

It's not a `_test.go` file, but a `test` pkg and go does not make any assumption about `test` pkgs.

share/ipld/nmt_hasher_integration_test.go
MSevey: > thanks for bringing this up, I was confused about how its done everywhere else. The comment line length seems to differ from the code line length, so I always just broke it up arbitrarily 😬

yea I usually manually trigger the line wrap on comments with `gq` since I use vim bindings in vscode.

share/ipld/nmt_hasher_integration_test.go
MSevey: > _test.go can only export from each other. You cannot import anything defined in a test file from a non-test file.

tbh TIL haha has this always been the case? Or is that a recent version update?

Maybe I've just been confusing exported methods in a testing pkg and exported methods in `_test.go` files.  But good to know

Wondertan: > tbh TIL haha has this always been the case? Or is that a recent version update?

I remember it has always been like this, but I am in Golang for only 3-4 years in this change could land before.

share/availability/test/testing.go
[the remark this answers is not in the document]
distractedm1nd: delete

share/ipld/nmt_hasher_integration_test.go
distractedm1nd: done

share/ipld/nmt_hasher_integration_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+var _ blockstore.Blockstore = (*MockBlockstore)(nil)
+
+// CorruptBlock is a block where the cid doesn't match the data.
+type CorruptBlock struct {
+	cid  cid.Cid
+	data []byte
+}
+
+func (b *CorruptBlock) RawData() []byte {
```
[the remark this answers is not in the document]
distractedm1nd: done, added to CorruptBlock docstring

share/ipld/nmt_hasher_integration_test.go
distractedm1nd: done

share/ipld/nmt_hasher_integration_test.go
distractedm1nd: I moved the mocks and new `TestNode` type to the testing package where their "base types" are defined

share/ipld/nmt_hasher_integration_test.go
distractedm1nd: reformatted with gq

share/ipld/nmt_hasher_integration_test.go
distractedm1nd: These are now located in the correct spot, but are exported so that the actual `_test.go` can access them. I renamed the struct types so that it is clear that everything is a `TestNode` on a `TestDagNet`, and not some actual type. Regardless, the constructor takes a `testing.T`, so it can only be used by tests.

share/availability/test/testing.go
distractedm1nd: I prefixed the structs and constructors with Test

share/availability/test/corrupt_data.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+func NewCorruptBlock(data []byte, fakeCID cid.Cid) *CorruptBlock {
+	return &CorruptBlock{
+		fakeCID,
+		data,
+	}
+}
+
+// MockBlockstore is a mock blockstore.Blockstore that saves both corrupted and original data for every block it
```
MSevey, line 50: line wrap

share/availability/test/testing.go
```diff
@@ -55,29 +54,27 @@ func (n *Node) ClearStorage() {
...
+	return &TestDagNet{
 		ctx: ctx,
 		T:   t,
 		net: mocknet.New(),
 	}
 }
 
-// Node create a plain network node that can serve and request data.
-func (dn *DagNet) Node() *Node {
+// NewTestNodeWithBlockstore creates a new plain TestNode with the given blockstore that can serve and request data.
```
MSevey, line 74: line wrap here and a few other places in this file.

share/ipld/corrupted_data_test.go
```diff
@@ -0,0 +1,41 @@
...
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/celestiaorg/celestia-node/share/availability/full"
+	availability_test "github.com/celestiaorg/celestia-node/share/availability/test"
+	"github.com/celestiaorg/celestia-node/share/service"
+)
+
+// TestNamespaceHasher_CorruptedData is an integration test that verifies that the NamespaceHasher of a recipient of
```
MSevey, line 15: line wrap comments in this file :-)

share/ipld/corrupted_data_test.go
```diff
@@ -0,0 +1,41 @@
...
+	net := availability_test.NewTestDAGNet(ctx, t)
+
+	requestor := full.Node(net)
+	provider, mockBS := availability_test.MockNode(t, net)
+	provider.ShareService = service.NewShareService(provider.BlockService, full.TestAvailability(provider.BlockService))
+	net.ConnectAll()
+
+	// before the provider starts attacking, we should be able to retrieve successfully
+	root := availability_test.RandFillBS(t, 16, provider.BlockService)
+	getCtx, cancelGet := context.WithTimeout(ctx, 2*time.Second)
```
MSevey, line 29: Where does the 2s come from? Any timing constants in  testing should have a comment explaining where they came from so that when/if tests become flaky due to timing issues we know which timing constants can be bumped vs debugged.

MSevey (changes requested):

share/availability/test/corrupt_data.go
distractedm1nd: This one is `gq`'ed, but a different one was not

share/ipld/corrupted_data_test.go
distractedm1nd: is `gq`ed already

share/ipld/corrupted_data_test.go
distractedm1nd: I extracted this into a constant with an explanation that it is arbitrary. But we need to fix this problem node wide, because our tests pass in random time intervals very often.

distractedm1nd: #1273

share/availability/test/testing.go
distractedm1nd: was already `gq`'ed

share/ipld/nmt_hasher_integration_test.go
renaynay: https://github.com/celestiaorg/celestia-node/pull/1256/files#diff-c374312b3ea40d59f26c88cd7b963b7db87bbeeb9e169d4419a7e960cd2f7503R45

MSevey (dismissed):

share/ipld/nmt_hasher_integration_test.go
distractedm1nd: I fixed the typo, but no its not supposed to be `ErrorIs`, because it contains a longer string and components that I don't have access to from the outside

share/ipld/corrupted_data_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/celestiaorg/celestia-node/share"
+	"github.com/celestiaorg/celestia-node/share/availability/full"
+	availability_test "github.com/celestiaorg/celestia-node/share/availability/test"
+	"github.com/celestiaorg/celestia-node/share/service"
+)
+
+// timeout is an arbitrarily picked interval of time in which a TestNode is expected to be able to complete a
```
renaynay, line 16:
```suggestion
// sharesAvailableTimeout is an arbitrarily picked interval of time in which a TestNode is expected to be able to complete a
```

share/availability/test/corrupt_data.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func (m MockBlockstore) AllKeysChan(ctx context.Context) (<-chan cid.Cid, error) {
+	// TODO implement me
+	panic("implement me")
+}
+
+func (m MockBlockstore) HashOnRead(enabled bool) {
+	// TODO implement me
+	panic("implement me")
```
Wondertan, line 116: Remove TODOs? We don't need to implement those as they are not used.

share/availability/test/corrupt_data.go
```diff
@@ -0,0 +1,129 @@
...
+func NewCorruptBlock(data []byte, fakeCID cid.Cid) *CorruptBlock {
+	return &CorruptBlock{
+		fakeCID,
+		data,
+	}
+}
+
+// MockBlockstore is a mock blockstore.Blockstore that saves both corrupted and original data for every block it
+// receives. If MockBlockstore.Attacking is true, it will serve the corrupted data on requests.
+type MockBlockstore struct {
```
Wondertan, line 51: optional nit: FraudBlockstore as MockBlockstore is too general.

share/availability/test/corrupt_data.go
```diff
@@ -0,0 +1,129 @@
...
+	ds.Datastore
+	Attacking bool
+}
+
+func (m MockBlockstore) Has(ctx context.Context, cid cid.Cid) (bool, error) {
+	return false, nil
```
Wondertan, line 57: This should also cause panic, as it has not been implemented properly

Wondertan (approved):
The test logic is legit. Only leaving some nits here

share/availability/test/corrupt_data.go
distractedm1nd: nope, you need Has

share/availability/test/corrupt_data.go
distractedm1nd: done

Wondertan (approved):

renaynay (approved):

share/availability/full/reconstruction_test.go
```diff
@@ -112,7 +112,9 @@ func TestShareAvailable_ConnectedFullNodes(t *testing.T) {
 	})
 
 	// create light nodes and start sampling for them immediately
-	lights1, lights2 := make([]*availability_test.Node, lightNodes/2), make([]*availability_test.Node, lightNodes/2)
+	lights1, lights2 := make(
+		[]*availability_test.TestNode, lightNodes/2),
+		make([]*availability_test.TestNode, lightNodes/2)
```
walldiss, line 117: Nit: Personally for me it is much harder to read code as 3 line like that than if it was 2 lines:

```suggestion
	lights1 := make([]*availability_test.TestNode, lightNodes/2)
	lights2 := make([]*availability_test.TestNode, lightNodes/2)
```

share/availability/test/corrupt_data.go
```diff
@@ -0,0 +1,126 @@
...
+func (b *CorruptBlock) RawData() []byte {
+	return b.data
+}
+
+func (b *CorruptBlock) Cid() cid.Cid {
+	return b.cid
+}
+
+func (b *CorruptBlock) String() string {
+	return fmt.Sprintf("[Block %s]", b.Cid())
```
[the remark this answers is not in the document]
walldiss: Will resolve those, since it is merged and comment is nit

PR #1258, opened 2022-10-20
refactor(state): `CoreAccessor` restricts writes on fraud

### LINKED ISSUE #1235: nodebuilder/state: `CoreAccessor` should only restrict state writes, not reads, on fraud
Right now, we blanket stop the entire `CoreAccessor` upon receipt of a fraud proof, but it should only restrict **writes**, not **reads**, meaning the connection to core client should not be stopped.

renaynay:
This PR refactors `CoreAccessor` such that it contains a field that determines whether `CoreAccessor` is read-only: 

* implements `RestrictWrites()` method on core accessor (and state module) that gets called from fraud lifecycle in the case of fraud instead of calling stop 
* middleware now calls `ReadOnly()` on state module to check if state is read-only, and if so, restricts POST endpoints
* `CoreAccessor` `Stop` method now returns an error if stop is called twice (as it is now no longer expected).

Resolves #1235 

Tagging as a bug-fix as this is technically fixing a bug in specified behaviour.

### DIFF
--- a/nodebuilder/state/service.go
+++ b/nodebuilder/state/service.go
@@ -17,6 +17,8 @@ import (
 type Module interface {
 	// IsStopped checks if the Module's context has been stopped
 	IsStopped() bool
+	// ReadOnly checks if the Module has disabled write access
+	ReadOnly() bool
 
 	// AccountAddress retrieves the address of the node's account/signer
 	AccountAddress(ctx context.Context) (state.Address, error)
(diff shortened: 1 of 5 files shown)


nodebuilder/state/service.go
```diff
@@ -17,6 +17,8 @@ import (
 type Module interface {
 	// IsStopped checks if the Module's context has been stopped
 	IsStopped() bool
+	// ReadOnly checks if the Module has disabled write access
+	ReadOnly() bool
```
Wondertan, line 21: Our module interfaces are soon to be used by the RPC. This means that all the methods are exposed over RPC. This type of _hack_ is not aligned with that. Also, note that this `ReadOnly` is only used by the middleware, but not in the case we use the module interface directly as a lib, which we defined as the main use case. 

I recommend closing this PR and revisiting this issue later once the RPC work is done.

Also note that current "RPC/API" endpoints will become a gateway soon, which means it will be read-only by default, and sending tx transactions from the node's key through the key must be forbidden.

Wondertan (changes requested):

renaynay:
New solution is needed for new RPC design.

PR #1263, opened 2022-10-24
refactor(core/fetcher): Improve error messages for fetcher methods

renaynay:
Self-explanatory (found by @jbowen93)


Wondertan (approved):
It would be nice to make an error type in here, but it definitely not worth doing it now

distractedm1nd (approved):

PR #1274, opened 2022-10-24
refactor: linewrapping comments to 80 width

distractedm1nd:
I reformatted all comments in the repository to wrap to 80 line width using https://github.com/alexkohler/cfmt.


MSevey (dismissed):
Awesome 👏

Also +1 for adding this to `make fmt`

walldiss:
Love this being automated! Why 80 is chosen to be the value for line length? Diff seems to have many 1 word newlines so it seems there was some formatting before that used a higher value.

MSevey:
> Love this being automated! Why 80 is chosen to be the value for line length? Diff seems to have many 1 word newlines so it seems there was some formatting before that used a higher value.

It actually dates back to punch cards I think haha but nowadays it it more about consistency and readability. Some languages  recommend a maximum line length in their standards, see this short little article https://www.codereadability.com/maximum-line-length/

The comments bring up a number of nice points, like readability on laptops with split screens, accessibility concerns, etc.

Wondertan:
@MSevey, if the argument is about the split screen, should we then also stick to the code itself being no more than 80, not just comments like this PR and tool do? The article you sent argues about the code length AFAIS.

If we agree that we should shorten our code length, then we should also do it in this PR, imo

MSevey:
> @MSevey, if the argument is about the split screen, should we then also stick to the code itself being no more than 80, not just comments like this PR and tool do? The article you sent argues about the code length AFAIS.
> 
> If we agree that we should shorten our code length, then we should also do it in this PR, imo

@Wondertan up to the team on line wrapping the code to 80. For me the difference is that the code has `go fmt` which ensures all code is formatted nicely and uniformly, where as comments were all over the place until this PR.  

So for me it is less the number and more that there is something that is keeping things consistent. It just makes the code nicer to work with and you don't get PRs with huge formatting diffs because people have different formatting settings locally.

Wondertan:
> because people have different formatting settings locally.

Good point. 

The question on changing the code length number remains open for the team. Personally, I feel like the length of the comments should be the same as the code or just a little shorter. Like if the code is 120, the comments should be 100/120 and not 80. So the gap of 20 is acceptable, but not more than that. Otherwise it feels a bit odd to me.

MSevey:
> The question on changing the code length number remains open for the team. Personally, I feel like the length of the comments should be the same as the code or just a little shorter. Like if the code is 120, the comments should be 100/120 and not 80. So the gap of 20 is acceptable, but not more than that. Otherwise it feels a bit odd to me.

yea, I'm not tied to 80, I'm just tied to consistency :-) 

```
when some
comments are like this and have no consistent line length
but at least are wrapped


and others just yolo'd and never heard of line wrapping and were like, this is totally fine, I love scrolling right for days to finish reading a comment about how this method works but you know what I can't see the method any more and have to start scrolling back and forth now to see what this comment is describing.

it leaves me :triggered:
```

Wondertan:
Ok, so let's:
* add this to `make fmt` in this PR 
* Try to add the same rule for the code itself
* Rebase

And we are good to go

distractedm1nd:
Set it to linelength 100 (code is at 120), and added it to `make fmt`.

One thing to note: I do not agree (and assume others are with me here) that the /minimum/ line length before a break should also be 100. It makes sense in a lot of places in our docs and explanation comments to line break earlier.

renaynay (approved):
ok thank u

renaynay:
@distractedm1nd minimum shouldn't be 100 tho

Makefile
```diff
@@ -82,6 +82,7 @@ fmt:
 	@find . -name '*.go' -type f -not -path "*.git*" -not -name '*.pb.go' -not -name '*pb_test.go' | xargs gofmt -w -s
 	@find . -name '*.go' -type f -not -path "*.git*"  -not -name '*.pb.go' -not -name '*pb_test.go' | xargs goimports -w -local github.com/celestiaorg
 	@go mod tidy -compat=1.17
+	@cfmt -w -m=100 ./...
```
MSevey, line 85: we need to add this to `deps` right? so people can install it, otherwise they'll get errors

MSevey (changes requested):

Makefile
Wondertan: We don't have `make deps`. Also note that we also have other tools like `markdownlint` and `golangci-lint` that would need to be downloaded via `deps`, but these all can be part of another issue and not necessarily this one.

Wondertan: A bit skeptical though about choosing the way things are installed for users. E.g. `golangci-lint` can be installed through `apt/pacman`(`brew` for Linux) and installing it strictly by `go install` may be annoying.

Wondertan (approved):

MSevey (dismissed):

MSevey (approved):

renaynay (approved):

Issue #63, opened 2021-09-15
build: Bootstrapper


Wondertan:
## Goal
We should support a simple flag during compilation to tell this binary should be used as a network bootstrapper.

## Current Solution
In upcoming #57 there is the flag in Config that configures a node to run as a Bootstrapper. However, this should not be done there, as users won't run personal bootstrappers, while the Config is a part of UX, so we should allow users to mistakenly configure their nodes as bootstrappers.

## Desired solution
The bootstrap should be a complied time parameter.

Bidon15:
Grooming 12/07/2022: 

- Changing the issue from Post-Mainnet to Mainnet
- it's still a low priority one
  - we can defer it again to Post-Mainnet if we don't have the capacity

ramin:
no longer relevant

PR #1280, opened 2022-10-25
env: create new package and introduce build variables. 

MSevey:
## Proposal

I would like to introduce the concept of build time variables. These build time variables allow for switching between values based on the build, i.e. testing vs production. 

## Why?

One major benefit of build specific variables is in testing. Timeouts are a common variable to define, and the expectations of timeouts in testing and in production are usually quite different. For example on production, we might measure response time in seconds while in a virtualized test environment we measure in milliseconds.  

Ultimately this adds a more granular level of control over our variables. 

## PR Context

It would have been nice for the new package to be called `build` since these are build time variables, but we currently use the `build` directory as the output for `make build`.  I would be open to refactoring that in a follow up if others agree with the naming. 

To properly using the build variables, the `-tags` flag needs to be used in testing (see updates to the `Makefile`). 

I updated one variable in the `fraud` package to illustrate how this new variable interface can be used.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ clean:
 cover:
 	@echo "--> Generating Code Coverage"
 	@go install github.com/ory/go-acc@latest
-	@go-acc -o coverage.txt -tags='testing' `go list ./... | grep -v nodebuilder/tests` -- -v
+	@go-acc -o coverage.txt `go list ./... | grep -v nodebuilder/tests` -- -v -tags='testing'
 .PHONY: cover
 
 ## deps: install dependencies.
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ clean:
 cover:
 	@echo "--> Generating Code Coverage"
 	@go install github.com/ory/go-acc@latest
-	@go-acc -o coverage.txt `go list ./... | grep -v nodebuilder/tests` -- -v
+	@go-acc -o coverage.txt -tags='testing' `go list ./... | grep -v nodebuilder/tests` -- -v
 .PHONY: cover
 
 ## deps: install dependencies.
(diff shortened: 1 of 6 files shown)


evan-forbes (commented):
what's the status of this? is this meant to be used org wide? if so, should we move it somewhere else?

Wondertan (changes requested):
I am very much in favor of such functionality. I believe that some binary configurations should not be exposed via user-facing configs. Users tend to play around with things, break them, report and then waste their time debugging weird edge cases that wouldn't happen if we just hide them. Such things should be hidden from them via compile-time flags, like in this PR.

Requesting changes to remove fraud `writeDeadline` example for a few reasons:
* Chozen `writeDeadline` is something we already planned to tackle via #709 and wanted to follow the #1063 example. So we essentially want to remove it from consts.
	* I wouldn't say that the `writeDeadline` needs production/testing configurations, but I understand that this is just an example you wanted to show us, which I appreciate.
* The introduced `env` package brings `celestia-node` application build semantics to it. We usually aim to avoid this and keep our root-level pkgs as independent as possible. The main reason is the reusability of pkgs by other teams or even other projects. This is now happening to `header` pkg, and rollmint team will use it for all things headers instead of reinventing the wheel. Making `header` dependent on `env` pkg would require them to use it as well when they might not need it. The same thing will happen to the `fraud` pkg.

Happy to merge it the way it is without any examples, but if we still want one, let's try it for #63, which could be a good example of a build time var. The integration should work with `p2p` module which is currently located in `nodebuilder/p2p`

MSevey:
@Wondertan I can remove the example, no problem.

As for the package level concern, maybe we can address that with @evan-forbes question. Is there maybe a better org level repo to put this in?

We could even create a `utils` repo that has things like this and other common helpers that many different repos could use.

PR #1298, opened 2022-10-28
feat(header/p2p): implement gcing for peerTracker

vgonkivs:
Resolves https://github.com/celestiaorg/celestia-node/issues/1411, https://github.com/celestiaorg/celestia-node/issues/1317

### DIFF
--- a/header/p2p/exchange.go
+++ b/header/p2p/exchange.go
@@ -53,23 +53,35 @@ func NewExchange(
 	if err != nil {
 		return nil, err
 	}
+
+	ctx, cancel := context.WithCancel(context.Background())
 	return &Exchange{
+		ctx:          ctx,
+		cancel:       cancel,
 		host:         host,
 		protocolID:   protocolID(protocolSuffix),
 		trustedPeers: peers,
-		peerTracker:  newPeerTracker(host, params.GC, params.MaxAwaitingTime, params.DefaultScore),
-		Params:       params,
+		peerTracker: newPeerTracker(
+			host,
+			params.GC,
+			params.MaxAwaitingTime,
+			params.DefaultScore,
+			params.MaxPeerTrackerSize,
+		),
+		Params: params,
 	}, nil
 }
 
 func (ex *Exchange) Start(context.Context) error {
-	ex.ctx, ex.cancel = context.WithCancel(context.Background())
-	go ex.peerTracker.track(ex.ctx)
+	go ex.peerTracker.track()
 	return nil
 }
 
 func (ex *Exchange) Stop(context.Context) error {
+	// cancel the session if it exists
 	ex.cancel()
+	// stop the peerTracker
+	ex.peerTracker.stop()
 	return nil
 }
 
--- a/header/p2p/exchange.go
+++ b/header/p2p/exchange.go
@@ -60,7 +60,6 @@ func NewExchange(
 		trustedPeers: peers,
 		peerTracker: newPeerTracker(
 			host,
-			params.GC,
 			params.MaxAwaitingTime,
 			params.DefaultScore,
 			params.MaxPeerTrackerSize,
--- a/header/p2p/exchange.go
+++ b/header/p2p/exchange.go
@@ -57,7 +57,7 @@ func NewExchange(
 		host:         host,
 		protocolID:   protocolID(protocolSuffix),
 		trustedPeers: peers,
-		peerTracker:  newPeerTracker(host),
+		peerTracker:  newPeerTracker(host, params.GC, params.MaxAwaitingTime, params.DefaultScore),
 		Params:       params,
 	}, nil
 }
--- a/header/p2p/exchange.go
+++ b/header/p2p/exchange.go
@@ -73,6 +73,7 @@ func NewExchange(
 }
 
 func (ex *Exchange) Start(context.Context) error {
+	go ex.peerTracker.gc()
 	go ex.peerTracker.track()
 	return nil
 }
--- a/header/p2p/exchange.go
+++ b/header/p2p/exchange.go
@@ -54,10 +54,7 @@ func NewExchange(
 		return nil, err
 	}
 
-	ctx, cancel := context.WithCancel(context.Background())
 	return &Exchange{
-		ctx:          ctx,
-		cancel:       cancel,
 		host:         host,
 		protocolID:   protocolID(protocolSuffix),
 		trustedPeers: peers,
@@ -73,6 +70,8 @@ func NewExchange(
 }
 
 func (ex *Exchange) Start(context.Context) error {
+	ex.ctx, ex.cancel = context.WithCancel(context.Background())
+
 	go ex.peerTracker.gc()
 	go ex.peerTracker.track()
 	return nil
--- a/header/p2p/options.go
+++ b/header/p2p/options.go
@@ -90,8 +90,6 @@ type ClientParameters struct {
 	MaxRequestSize uint64
 	// MaxHeadersPerRequest defines the max amount of headers that can be requested per 1 request.
 	MaxHeadersPerRequest uint64
-	// GC defines the duration after which the peerTracker starts removing peers.
-	GC time.Duration
 	// MaxAwaitingTime specifies the duration that gives to the disconnected peer to be back online,
 	// otherwise it will be removed on the next GC cycle.
 	MaxAwaitingTime time.Duration
@@ -108,7 +106,6 @@ func DefaultClientParameters() *ClientParameters {
 		MinResponses:         2,
 		MaxRequestSize:       512,
 		MaxHeadersPerRequest: 64,
-		GC:                   time.Minute * 30,
 		MaxAwaitingTime:      time.Hour,
 		DefaultScore:         1,
 		MaxPeerTrackerSize:   100,
@@ -134,9 +131,6 @@ func (p *ClientParameters) Validate() error {
 	if p.MaxHeadersPerRequest > p.MaxRequestSize {
 		return fmt.Errorf("MaxHeadersPerRequest should not be more than MaxRequestSize."+
 			"MaxHeadersPerRequest: %v, MaxRequestSize: %v", p.MaxHeadersPerRequest, p.MaxRequestSize)
-	}
-	if p.GC == 0 {
-		return fmt.Errorf("invalid gc period for peerTracker: %s. %s: %v", greaterThenZero, providedSuffix, p.GC)
 	}
 	if p.MaxAwaitingTime == 0 {
 		return fmt.Errorf("invalid MaxAwaitingTime for peerTracker: "+
@@ -171,17 +165,6 @@ func WithMaxHeadersPerRequest[T ClientParameters](amount uint64) Option[T] {
 			t.MaxHeadersPerRequest = amount
 		}
 
-	}
-}
-
-// WithGCCycle is a functional option that configures the
-// `GC` parameter.
-func WithGCCycle[T ClientParameters](cycle time.Duration) Option[T] {
-	return func(p *T) {
-		switch t := any(p).(type) { //nolint:gocritic
-		case *ClientParameters:
-			t.GC = cycle
-		}
 	}
 }
 
--- a/header/p2p/options.go
+++ b/header/p2p/options.go
@@ -1,6 +1,7 @@
 package p2p
 
 import (
+	"errors"
 	"fmt"
 	"time"
 )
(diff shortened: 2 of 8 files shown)


header/p2p/options.go
```diff
@@ -100,18 +106,30 @@ func DefaultClientParameters() *ClientParameters {
 		MinResponses:         2,
 		MaxRequestSize:       512,
 		MaxHeadersPerRequest: 64,
+		GC:                   time.Minute * 30,
+		MaxAwaitingTime:      time.Hour,
```
walldiss, line 110: Could you please share why those value are chosen to be default?

Those value seems to be very high compared to one that are usually used for keep-alive in persisted connections. For example nginx has default keep-alive timeout of 30 sec and max 3600sec(hour). Are those values so high because peertracker tries to not lose accumulated score of peers?

vgonkivs: yes, the main reason why these values are high is keep peers with peers as much time as we can, but from the other side we should not cache hundreds of dead peers. Anyway, we can discuss and choose any other values.

header/p2p/options.go
```diff
@@ -100,18 +106,30 @@ func DefaultClientParameters() *ClientParameters {
...
-		return fmt.Errorf("invalid minimum amount of responses: %s", errSuffix)
+		return errors.New("invalid minimum amount of responses: value should be positive and non-zero")
 	}
 	if p.MaxRequestSize == 0 {
-		return fmt.Errorf("invalid max request size: %s", errSuffix)
+		return fmt.Errorf("invalid max request size: %d", p.MaxRequestSize)
 	}
 	if p.MaxHeadersPerRequest == 0 || p.MaxHeadersPerRequest > p.MaxRequestSize {
-		return fmt.Errorf("invalid max headers per request: %s", errSuffix)
+		return errors.New("invalid max headers per request")
```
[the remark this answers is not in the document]
vgonkivs: Yeah, I agree. Will return more meaningful errors.

header/p2p/peer_tracker.go
```diff
@@ -12,23 +13,48 @@ import (
...
+		maxAwaitingTime:    maxAwaitingTime,
+		defaultScore:       defaultScore,
+		maxPeerTrackerSize: maxPeerTrackerSize,
+		done:               make(chan struct{}, 2),
 	}
 }
 
-func (p *peerTracker) track(ctx context.Context) {
+func (p *peerTracker) track() {
+	go p.gc()
```
renaynay, line 57: can you please spawn garbage collection from `ex.Start` ? makes it more clear what routines are spawned from start of exchange. I don't like spawning long-lived routines from other long-lived routines.

header/p2p/peer_tracker.go
```diff
@@ -102,3 +140,50 @@ func (p *peerTracker) peers() []*peerStat {
...
+			go func() {
+				defer wg.Done()
+				for id, peer := range p.connectedPeers {
+					if peer.peerScore <= p.defaultScore {
+						delete(p.connectedPeers, id)
+					}
+				}
+			}()
+			wg.Wait()
+			p.Unlock()
```
renaynay, line 177: ? why does this have to be async?

header/p2p/peer_tracker.go
```diff
@@ -37,15 +63,18 @@ func (p *peerTracker) track(ctx context.Context) {
 	subs, err := p.host.EventBus().Subscribe(&event.EvtPeerConnectednessChanged{})
 	if err != nil {
 		log.Errorw("subscribing to EvtPeerConnectednessChanged", "err", err)
+		p.done <- struct{}{}
```
renaynay, line 66: defer this at the top of the func so you don't have to call at every return.

renaynay (changes requested):

header/p2p/peer_tracker.go
vgonkivs: I made this change before I realised the we should have peerTracker limit size. And probably we could have 100000 of peers in cache, so it was added to speed gc(as we are locking the mutex). But we do not need them to be async.

header/p2p/peer_tracker.go
vgonkivs: yes, thanks

header/p2p/peer_tracker_test.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestPeerTracker_GC(t *testing.T) {
+	h := createMocknet(t, 1)
+	p := newPeerTracker(h[0], time.Millisecond*200, time.Millisecond*1, 1, 5)
+	pid1 := peer.ID("peer1")
+	pid2 := peer.ID("peer2")
+	p.connectedPeers[pid1] = &peerStat{peerID: pid1, peerScore: 0.5}
+	p.disconnectedPeers[pid2] = &peerStat{peerID: pid2, removedAt: time.Now()}
```
distractedm1nd, line 18: please also add 2 examples of peers that shouldn't be removed, and assert that the correct one has been removed

header/p2p/peer_tracker.go
```diff
@@ -63,19 +92,27 @@ func (p *peerTracker) connected(pID peer.ID) {
 	if p.host.ID() == pID {
 		return
 	}
+	p.Lock()
+	defer p.Unlock()
+	// skip adding the peer to avoid overfilling of the peerTracker with unused peers if:
+	// peerTracker reaches the maxTrackerSize and the size of the connected peers
+	// is more than the size of the disconnected peers.
```
distractedm1nd, line 100:
```suggestion
	// peerTracker reaches the maxTrackerSize and there are more connected peers
	// than disconnected peers.
```

header/p2p/peer_tracker.go
```diff
@@ -12,23 +13,48 @@ import (
...
 	// we cache the peer once they disconnect,
 	// so we can guarantee that peerQueue will only contain active peers
 	disconnectedPeers map[peer.ID]*peerStat
 
 	host host.Host
+
+	gcPeriod           time.Duration
+	maxAwaitingTime    time.Duration
+	defaultScore       float32
+	maxPeerTrackerSize int
```
distractedm1nd, line 30: short doc for these fields

distractedm1nd (changes requested):

header/p2p/peer_tracker.go
```diff
@@ -102,3 +142,43 @@ func (p *peerTracker) peers() []*peerStat {
...
+func (p *peerTracker) gc() {
+	ticker := time.NewTicker(p.gcPeriod)
+	for {
+		select {
+		case <-p.ctx.Done():
+			p.done <- struct{}{}
+			return
+		case <-ticker.C:
+			p.Lock()
+			// disconnectedPeers will be removed it removedAt < time.Now
```
distractedm1nd, line 159:
```suggestion
```

header/p2p/peer_tracker.go
```diff
@@ -102,3 +142,43 @@ func (p *peerTracker) peers() []*peerStat {
...
+			for id, peer := range p.disconnectedPeers {
+				if peer.removedAt.Before(time.Now()) {
+					delete(p.disconnectedPeers, id)
+				}
+			}
+
+			// connectedPeers will be removed if their score is less of equal to defaultScore
```
distractedm1nd, line 166:
```suggestion
```

distractedm1nd (commented):
sync looks a lot better, just recommending removing two comments

header/p2p/peer_tracker.go
vgonkivs: https://github.com/celestiaorg/celestia-node/pull/1298/commits/74283254d13527d615d6c57b70e6e62846413ad1

renaynay (commented):
last nits

header/p2p/peer_tracker.go
```diff
@@ -102,3 +148,41 @@ func (p *peerTracker) peers() []*peerStat {
...
+				if peer.pruneDeadline.Before(time.Now()) {
+					delete(p.disconnectedPeers, id)
+				}
+			}
+
+			for id, peer := range p.connectedPeers {
+				if peer.peerScore <= p.defaultScore {
+					delete(p.connectedPeers, id)
+				}
+			}
```
Wondertan, line 175: We should rename connectedPeers to trackedPeers or something. Previously, we could call them so as we had 1:1 mapping of connected peers, but now we can connect to more than what we track. Because of this check and the global limit introduced in this PR

Wondertan (commented):
LGTM. Have minor suggestion and questions

header/p2p/options.go
```diff
@@ -92,6 +90,16 @@ type ClientParameters struct {
 	MaxRequestSize uint64
 	// MaxHeadersPerRequest defines the max amount of headers that can be requested per 1 request.
 	MaxHeadersPerRequest uint64
+	// GC defines the duration after which the peerTracker starts removing peers.
+	GC time.Duration
```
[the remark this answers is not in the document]
vgonkivs: As I mentioned some time ago, it would be good to extract gc and rely on it from peerTracker or @distractedm1nd PR or  backoff connector.  And in this case we will have to specify how much time do we need for our gc cycle.

header/p2p/exchange.go
```diff
@@ -53,23 +53,36 @@ func NewExchange(
 	if err != nil {
 		return nil, err
 	}
+
+	ctx, cancel := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
vgonkivs: 🤔 . I think it's an artifact from rebasing(had some issues with it). Will move it back

header/p2p/options.go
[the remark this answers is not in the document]
Wondertan: It's unrelated to GC times though.

header/p2p/peer_tracker.go
```diff
@@ -12,23 +13,57 @@ import (
...
+	gcPeriod time.Duration
+	// maxAwaitingTime specifies the duration that gives to the disconnected peer to be back online,
+	// otherwise it will be removed on the next GC cycle.
+	maxAwaitingTime time.Duration
+	// defaultScore specifies the score for newly connected peers.
+	defaultScore float32
+	// maxPeerTrackerSize specifies the max amount of peers that can be added to the peerTracker.
+	maxPeerTrackerSize int
+
+	// done is used to stop background routines in peerTracker.
```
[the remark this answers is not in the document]
vgonkivs: added proper comments

renaynay (approved):

header/p2p/options.go
```diff
@@ -100,18 +108,45 @@ func DefaultClientParameters() *ClientParameters {
...
+	if p.GC == 0 {
+		return fmt.Errorf("invalid gc period for peerTracker: %s. %s: %v", greaterThenZero, providedSuffix, p.GC)
+	}
+	if p.MaxAwaitingTime == 0 {
+		return fmt.Errorf("invalid MaxAwaitingTime for peerTracker: "+
+			"%s. %s: %v", greaterThenZero, providedSuffix, p.MaxAwaitingTime)
+	}
+	if p.DefaultScore < 0 {
+		return fmt.Errorf("invalid DefaultScore: %s. %s: %f", greaterThenZero, providedSuffix, p.DefaultScore)
+	}
```
walldiss, line 147: Small typo: Evaluation logic and error message mismatch if `DefaultScore` == 0.

walldiss (commented):
Found small type. Besides that, looks legit!

header/p2p/options.go
vgonkivs: Nice catch. But lets make it `p.DefaultScore =< 0`

header/p2p/options.go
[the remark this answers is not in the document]
Wondertan: I mean, the abstraction is not related to the question: should we make GC time configurable or not? Is this something that brings value, or just bloats the config? 

I feel it doesn't bring anything and can be a constant.

header/p2p/peer_tracker.go
```diff
@@ -12,23 +13,58 @@ import (
 
 type peerTracker struct {
```
Wondertan, line 14: The last suggestion is to logically sort all the fields and document what trackedPeers are

header/p2p/peer_tracker.go
```diff
@@ -63,42 +100,91 @@ func (p *peerTracker) connected(pID peer.ID) {
...
+		len(p.trackedPeers) > len(p.disconnectedPeers) {
+		return
+	}
+
 	for _, c := range p.host.Network().ConnsToPeer(pID) {
 		// check if connection is short-termed and skip this peer
 		if c.Stat().Transient {
 			return
 		}
 	}
```
Wondertan, line 118: Just realized that here we have a bug, rare but still. There can be two connections between two peers, one of which is transient, this logic aborts fully if one of them is transient.

Wondertan: Another suggestion is to check connections first, and only then maxPeerTrackerSize

vgonkivs: > Just realized that here we have a bug, rare but still. There can be two connections between two peers, one of which is transient, this logic aborts fully if one of them is transient.

Actually I discussed it with you. And we agreed to skip the peer.

vgonkivs: > Another suggestion is to check connections first, and only then maxPeerTrackerSize

Why? if our peerTracker is full, then we do not need to go further.

renaynay (dismissed):

Wondertan (approved):

PR #1300, opened 2022-10-31
chore: bump celestia app v0.9.0

evan-forbes:
bumps to v0.9.0 of celestia-app

this also requires a bump to v0.11.0 of nmt, v0.7.0 of rsmt2d, and v1.4.0 of our fork of the sdk

most of the diff is updating to the new api introduced in rsmt2d and the wrapper

pls feel free to push or edit this branch to fix the remaining bugs


walldiss:
@evan-forbes Could you please grant me write access to this branch? I've found solution to one of the problems and want to commit it.

walldiss:
I've made all tests pass except few that was flaky because of unrelated to PR changes. CI test still could fail due to tests issues that will be fixed in https://github.com/celestiaorg/celestia-node/pull/1314

The problems I've found through tests debugging were:
 - Before update`ErrByzantineData`  returned from `rsmt2d` was bugged. It should had `Shares` field populated with pre-build shares, but had used rebuild ones instead. It got fixed in new version, which is nice! That introduced possibility to contain nil shares. We didn't handled nil shares case before, so I've added related checks.
 - Turned out`nmt.Batch` is not concurrency safe and there is even an open [issue](https://github.com/ipfs/go-ipld-format/issues/47) for it. The solution was to add external mutex to protect it from parallel operations.

share/eds/byzantine/share_proof.go
```diff
@@ -51,6 +51,10 @@ func (s *ShareWithProof) Validate(root cid.Cid) bool {
 }
 
 func (s *ShareWithProof) ShareWithProofToProto() *pb.Share {
+	if s == nil {
+		return &pb.Share{}
+	}
```
Wondertan, line 56: Is this a real case?

share/ipld/nmt_adder.go
```diff
@@ -12,6 +14,7 @@ import (
 // NmtNodeAdder adds ipld.Nodes to the underlying ipld.Batch if it is inserted
 // into a nmt tree.
 type NmtNodeAdder struct {
+	lock   sync.Mutex
```
Wondertan, line 17: Let's document that the lock is used for both the Batch and cid.Set

share/ipld/nmt_adder.go
```diff
@@ -67,11 +76,18 @@ func (n *NmtNodeAdder) VisitInnerNodes(hash []byte, children ...[]byte) {
 
 // Commit checks for errors happened during Visit and if absent commits data to inner Batch.
 func (n *NmtNodeAdder) Commit() error {
+	n.lock.Lock()
+	defer n.lock.Unlock()
+
 	if n.err != nil {
-		return n.err
+		return fmt.Errorf("before commit: %w", n.err)
```
Wondertan, line 83: Let's mention what kind of a commit this is. Same for after commit message

share/eds/byzantine/share_proof.go
walldiss: It could be triggered when encoding `ErrByzantineData` that can contain nil shares now.

share/ipld/nmt_adder.go
walldiss: It is used for whole `NmtNodeAdder`. Do you think its usage could be confusing?

share/eds/byzantine/share_proof.go
walldiss: I've have added that later this case is also handled on unmarshaller side

share/ipld/nmt_adder.go
Wondertan: Just a quick mention that cover all the fields wouldn't hurt

Wondertan (dismissed):
LGTM

distractedm1nd (approved):
great finds with the nmt node adder lock and proof proto!

share/eds/byzantine/bad_encoding.go
```diff
@@ -155,9 +155,9 @@ func (p *BadEncodingProof) Validate(header *header.ExtendedHeader) error {
 	}
 	rebuiltShares = append(rebuiltShares, rebuiltExtendedShares...)
 
-	tree := wrapper.NewErasuredNamespacedMerkleTree(uint64(len(shares) / 2))
-	for i, share := range rebuiltShares {
-		tree.Push(share, rsmt2d.SquareIndex{Axis: uint(p.Index), Cell: uint(i)})
+	tree := wrapper.NewErasuredNamespacedMerkleTree(uint64(len(shares)/2), uint(p.Index))
```
renaynay, line 158: why p.Index instead of p.Axis ?

renaynay (changes requested):
hopefully quick ones

share/eds/byzantine/bad_encoding.go
evan-forbes: good question, p.Axis only determines if its a row or column. When creating wrapped NMTs, we need to know which row or column in order to determine the correct namespace, so we use the index.

nodebuilder/tests/fraud_test.go
```diff
@@ -31,7 +31,8 @@ Steps:
 Note: 15 is not available because DASer will be stopped before reaching this height due to receiving a fraud proof.
 */
 func TestFraudProofBroadcasting(t *testing.T) {
-	sw := swamp.NewSwamp(t, swamp.WithBlockTime(time.Millisecond*100))
+	testTimeout := time.Millisecond * 200
```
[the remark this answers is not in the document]
evan-forbes: just to decrease flakiness of the CI. Passes everytime locally with 100ms. added docs in https://github.com/celestiaorg/celestia-node/pull/1300/commits/5919a563646755b090c383211afe968894090ba2

share/eds/byzantine/bad_encoding.go
renaynay: Ah okay, I was confused bc the param is called `axisIndex` for `NewErasuredNamespacedMerkleTree`

renaynay (approved):

Wondertan (approved):

PR #1304, opened 2022-10-31
feat(header)!: introduce `header.Header` interface

tzdybal:
Most important changes in this PR:
* `tmbytes.HexBytes` replaced by `header.Hash`
* `header.Header` used in interfaces in `header` package
* `IsBefore` inlined, so we don't need to include this method in interface (it's a oneliner used once)
* `VerifyNonAdjacent` added to `Header` interface (at least for now)
* Generics introduced to avoid casting

Resolves https://github.com/celestiaorg/rollmint/issues/575.


Wondertan:
Yay, you came back!

tzdybal:
Instead of rebasing, I re-introduced all the changes/refactorings. Now the commit history is pretty :joy:
Some work still needed - for example `RequestExtendedHeader` should be renamed to `RequestHeader`, but I should be done before the end of weekend.

header/interface.go
```diff
@@ -2,139 +2,30 @@ package header
...
-// ConstructFn aliases a function that creates an ExtendedHeader.
+type Subscriber = header.Subscriber[*ExtendedHeader]
+type Subscription = header.Subscription[*ExtendedHeader]
+type Broadcaster = header.Broadcaster[*ExtendedHeader]
+type Exchange = header.Exchange[*ExtendedHeader]
+type Store = header.Store[*ExtendedHeader]
+type Getter = header.Getter[*ExtendedHeader]
+type Head = header.Head[*ExtendedHeader]
+type Syncer = sync.Syncer[*ExtendedHeader]
+type ExchangeServer = p2p.ExchangeServer[*ExtendedHeader]
```
Wondertan, line 22: Question for the team. Do we want to keep those aliases or should we use the types directly?

Mainly, these aliases used in the node modules and are meant to prettify/hide generics(type parameters) from the code there. However, if you check it, the constructors(which don't alias her) for aliased components still require type parameters. So it seems like we are constructing one type but providing to the FX a completely different one, which is ambigious. So I would say we can safely remove these aliases. 

cc @renaynay and @distractedm1nd, who usuallly have opinions on such things

libs/header/hash.go
```diff
@@ -0,0 +1,34 @@
+package header
+
+import (
+	"encoding/hex"
+	"fmt"
+	"strings"
+)
+
+type Hash []byte
```
Wondertan, line 9: godoc

Wondertan (changes requested):
Awesome! 

First deep review. Found a few things that has to be cleaned up so requesting changes.

share/eds/store.go
```diff
@@ -88,13 +88,13 @@ func NewStore(basepath string, ds datastore.Batching) (*Store, error) {
 	}
 
 	store := &Store{
-		basepath: basepath,
-		dgstr:    dagStore,
-		topIdx:   invertedRepo,
-		carIdx:   fsRepo,
+		basepath:   basepath,
```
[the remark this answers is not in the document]
tzdybal: After rebasing linter complained about wrong indentation.

vgonkivs (commented):
Agree with @Wondertan that type aliases look redundant. Other than that, fantastic job.

libs/header/header.go
```diff
@@ -0,0 +1,35 @@
+package header
+
+import (
+	"encoding"
+	"time"
+)
+
+// Header abstracts all methods required to perform header sync.
+type Header interface {
```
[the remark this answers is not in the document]
tzdybal: Done!

header/verify.go
```diff
@@ -48,11 +55,15 @@ func (eh *ExtendedHeader) VerifyNonAdjacent(untrst *ExtendedHeader) error {
...
-		return &ErrNonAdjacent{
-			Head:      eh.Height,
-			Attempted: untrst.Height,
+func (eh *ExtendedHeader) VerifyAdjacent(untrusted headerpkg.Header) error {
+	untrst, ok := untrusted.(*ExtendedHeader)
+	if !ok {
+		return &VerifyError{errors.New("invalid header type: expected *ExtendedHeader")}
+	}
+	if untrst.Height() != eh.Height()+1 {
+		return &headerpkg.ErrNonAdjacent{
```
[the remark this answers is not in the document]
tzdybal: Done!

header/metrics.go
```diff
@@ -32,7 +32,7 @@ func WithMetrics(store Store) {
 
 			headC.Observe(
 				ctx,
-				head.Height,
+				head.Height(),
```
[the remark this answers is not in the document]
tzdybal: We'll definitely use opentelemetry, but it's not a priority RN.

header/interface.go
tzdybal: Removed!

libs/header/hash.go
tzdybal: Done!

nodebuilder/header/module.go
```diff
@@ -77,25 +79,28 @@ func ConstructModule(tp node.Type, cfg *Config) fx.Option {
...
+			p2p.NewSubscriber[*header.ExtendedHeader],
+			fx.OnStart(func(ctx context.Context, sub *p2p.Subscriber[*header.ExtendedHeader]) error {
 				return sub.Start(ctx)
 			}),
-			fx.OnStop(func(ctx context.Context, sub *p2p.Subscriber) error {
+			fx.OnStop(func(ctx context.Context, sub *p2p.Subscriber[*header.ExtendedHeader]) error {
 				return sub.Stop(ctx)
 			}),
 		)),
+		fx.Provide(func() pubsub.MsgIdFunction {
```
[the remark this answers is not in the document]
tzdybal: Done! But please take a look to ensure that it's the best way of dealing with it (https://github.com/celestiaorg/celestia-node/pull/1304/files#diff-0baca247ddd481193a9b3946dd77b83c9f4e803f9bdb8a296bbf2e89749118a4R88)

header/verify.go
```diff
@@ -2,10 +2,13 @@ package header
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"time"
 
 	"github.com/tendermint/tendermint/light"
+
+	headerpkg "github.com/celestiaorg/celestia-node/libs/header"
```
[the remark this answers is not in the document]
tzdybal: All usages changed to `libhead`

libs/header/header.go
```diff
@@ -0,0 +1,35 @@
...
+	// IsRecent checks if header is recent against the given blockTime.
+	IsRecent(duration time.Duration) bool
+	// IsExpired checks if header is expired against trusting period.
+	IsExpired() bool
+	// VerifyAdjacent validates adjacent untrusted header against trusted header.
+	VerifyAdjacent(Header) error
+	// VerifyNonAdjacent validates non-adjacent untrusted header against trusted header.
+	VerifyNonAdjacent(Header) error
+	// Verify performs basic verification of untrusted header.
+	Verify(Header) error
```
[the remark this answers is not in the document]
tzdybal: Combining this and next comment  - I'll get rid of `VerifyAdjacent` and `VerifyNonAdjacent`, and `Verify` will replace them both.

libs/header/header.go
```diff
@@ -0,0 +1,35 @@
...
+	// Time returns time when header was created.
+	Time() time.Time
+	// IsRecent checks if header is recent against the given blockTime.
+	IsRecent(duration time.Duration) bool
+	// IsExpired checks if header is expired against trusting period.
+	IsExpired() bool
+	// VerifyAdjacent validates adjacent untrusted header against trusted header.
+	VerifyAdjacent(Header) error
+	// VerifyNonAdjacent validates non-adjacent untrusted header against trusted header.
+	VerifyNonAdjacent(Header) error
```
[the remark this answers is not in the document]
tzdybal: There are some cases where we depend on "adjacency" of the headers. In such places we would need extra condition in the calling code. I'm not hypothesizing -  I did draft implementation and tests are failing.

libs/header/header.go
```diff
@@ -0,0 +1,37 @@
...
+	// Verify performs basic verification of untrusted header.
+	Verify(Header) error
```
Wondertan, line 31: Should we still remove the Verify from here and from the implementation for now?  Assuming we return it later when we unify non and adjacent verification

Wondertan (commented):
Needs a rebase again 😭 
Happy to help with it btw

libs/header/store/heightsub.go
```diff
@@ -6,44 +6,44 @@ import (
...
 }
 
 // Sub subscribes for a header of a given height.
 // It can return errElapsedHeight, which means a requested header was already provided
 // and caller should get it elsewhere.
-func (hs *heightSub) Sub(ctx context.Context, height uint64) (*header.ExtendedHeader, error) {
+func (hs *heightSub[H]) Sub(ctx context.Context, height uint64) (H, error) {
 	if hs.Height() >= height {
-		return nil, errElapsedHeight
+		return *new(H), errElapsedHeight //nolint:gocritic
```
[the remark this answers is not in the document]
Wondertan: Also, `new` allocates to heap by  Go's specs, which is undesirable in here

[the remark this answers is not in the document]
Wondertan: https://stackoverflow.com/a/70589302 🤮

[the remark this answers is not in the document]
Wondertan: So either this or named return param...

libs/header/header.go
tzdybal: Yes, definitely, I just forgot to remove it after rolling by my other changes.

header/verify.go
```diff
@@ -23,47 +26,55 @@ var TrustingPeriod = 168 * time.Hour
...
+	return time.Since(eh.Time()) <= blockTime // TODO @renaynay: should we allow for a 5-10 block drift here?
 }
 
 // VerifyNonAdjacent validates non-adjacent untrusted header against trusted 'eh'.
-func (eh *ExtendedHeader) VerifyNonAdjacent(untrst *ExtendedHeader) error {
+func (eh *ExtendedHeader) VerifyNonAdjacent(untrusted libhead.Header) error {
+	untrst, ok := untrusted.(*ExtendedHeader)
+	if !ok {
+		return &libhead.VerifyError{Reason: errors.New("invalid header type: expected *ExtendedHeader")}
+	}
```
vgonkivs, line 43: May be extract in a func? It is the same as in `VerifyAdjacent`

libs/header/store/heightsub.go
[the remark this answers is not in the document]
tzdybal: Yes, this  is definitely not pretty idiom.

header/testing.go
```diff
@@ -25,8 +24,14 @@ import (
 	"github.com/celestiaorg/celestia-app/pkg/da"
 
 	"github.com/celestiaorg/celestia-node/core"
+	libhead "github.com/celestiaorg/celestia-node/libs/header"
+	"github.com/celestiaorg/celestia-node/libs/header/test"
+	"github.com/celestiaorg/celestia-node/share"
 )
 
+// Validator aliases a func that validates Header.
+type Validator = func(context.Context, *ExtendedHeader) pubsub.ValidationResult
```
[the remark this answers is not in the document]
tzdybal: Removed!

header/verify.go
vgonkivs: Yes, I think extracting error is enough.

libs/header/test/testing.go
```diff
@@ -0,0 +1,294 @@
...
+// TrustingPeriod is period through which we can trust a header's validators set.
+// This is copied from ceelstia-node
+//
+// Should be significantly less than the unbonding period (e.g. unbonding
+// period = 3 weeks, trusting period = 2 weeks).
+//
+// More specifically, trusting period + Time needed to check headers + Time
+// needed to report and punish misbehavior should be less than the unbonding
+// period.
+var TrustingPeriod = 168 * time.Hour
```
Wondertan, line 87: It can be removed now, no? We are now passing it through the config

header/header.go
```diff
@@ -42,6 +43,29 @@ type ExtendedHeader struct {
...
+	return eh.RawHeader.Height
+}
+
+func (eh *ExtendedHeader) Time() time.Time {
+	return eh.RawHeader.Time
+}
+
+func (eh *ExtendedHeader) Validate() error {
+	// TODO(tzdybal): rename ValidateBasic to Validate
+	return eh.ValidateBasic()
```
Wondertan, line 64: Does anything prevent us from doing this already?

tzdybal: No. Renamed.

libs/header/p2p/subscription.go
```diff
@@ -0,0 +1,54 @@
...
+	msg, err := s.subscription.Next(ctx)
+	if err != nil {
+		var zero H
+		return zero, err
+	}
+	log.Debugw("received message", "topic", msg.Message.GetTopic(), "sender", msg.ReceivedFrom)
+
+	header, ok := msg.ValidatorData.(H)
+	if !ok {
+		panic(fmt.Sprintf("invalid type received %s", reflect.TypeOf(msg.ValidatorData)))
```
renaynay, line 44: probably shouldn't be a panic? - wouldn't it be possible to send a header of a diff type if you're connected to the specific protocolID of the sub?

Wondertan (commented):
I just realized that we don't actually need IsExpired and IsRecent to be defined on the interface. Just having Time is enough, the syncing logic itself can figure out isExpired or isRecent by only accessing Time.

So let's remove them from the interface and instead make two private methods (isExpired and isRecent) on Syncer that takes Header and returns bool

libs/header/hash.go
```diff
@@ -0,0 +1,38 @@
...
+	jbz := make([]byte, len(s)+2)
+	jbz[0] = '"'
+	copy(jbz[1:], s)
+	jbz[len(jbz)-1] = '"'
+	return jbz, nil
+}
+
+// UnmarshalJSON deserializes JSON representation of a Hash into object.
+func (h *Hash) UnmarshalJSON(data []byte) error {
+	if len(data) < 2 || data[0] != '"' || data[len(data)-1] != '"' {
```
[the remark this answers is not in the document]
tzdybal: This is copied from https://github.com/tendermint/tendermint/blob/main/libs/bytes/bytes.go to maintain compatibility. Apparently hex bytes needs to be inside a string.

libs/header/header.go
```diff
@@ -0,0 +1,37 @@
...
+	// IsZero reports whether Header is a zero value of it's concrete type.
+	IsZero() bool
+	// ChainID returns identifier of the chain (ChainID).
+	ChainID() string
+	// Hash returns hash of a header.
+	Hash() Hash
+	// Height returns the height of a header.
+	Height() int64
+	// LastHeader returns the hash of last header before this header (aka. previous header hash).
+	LastHeader() Hash
```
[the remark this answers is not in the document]
tzdybal: This will be `nil` (as this is the default value for `ExtendedHeader.RawHeader.LastBlockId.Hash`)

libs/header/p2p/subscription.go
tzdybal: This code is moved from header/p2p/subscription (again, Github has difficulties to show this properly).

I don't feel competent to answer your question.

CC: @Wondertan @vgonkivs

Wondertan: It should be a panic. Otherwise, there is a critical bug in the code. We set H type to Validator Data explicitly in AddValidator

libs/header/test/testing.go
```diff
@@ -0,0 +1,287 @@
...
+
+	Height int64
+	Time   time.Time
+}
+
+type DummyHeader struct {
+	Raw
+
+	hash header.Hash
+}
```
Wondertan, line 33: Q: why are two separate structs and not a single combination?

tzdybal: To expose `ChainID()`, `Time()`, `Height()`, `PreviousHash()` methods, while also being able to easily assign values for those fields outside of `test` package.

renaynay (approved):

Wondertan (approved):

PR #1313, opened 2022-11-02
feat(nodebuilder/node): Implement `Module`

### LINKED ISSUE #977: Implement additional modules that don't yet exist
- [ ] Node API
- [ ] P2P API
- [x] Fraud API

renaynay:
Resolves #977


nodebuilder/node/node.go
```diff
@@ -0,0 +1,110 @@
...
+
+func (api *API) AuthVerify(ctx context.Context, token string) ([]auth.Permission, error) {
+	return api.Internal.AuthVerify(ctx, token)
+}
+
+func (api *API) AuthNew(ctx context.Context, perms []auth.Permission) ([]byte, error) {
+	return api.Internal.AuthNew(ctx, perms)
+}
+
+type admin struct {
```
distractedm1nd, line 67: can you move this struct and its methods to a new file

nodebuilder/node/node.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/filecoin-project/go-jsonrpc/auth"
+	logging "github.com/ipfs/go-log/v2"
+)
+
+var (
+	buildTime       string
+	lastCommit      string
+	semanticVersion string
+)
+
```
distractedm1nd, line 17:
```suggestion

var _ Module = (*API)(nil)

```

renaynay: done

nodebuilder/node/node.go
renaynay: done

distractedm1nd (approved):

Makefile
```diff
@@ -1,6 +1,6 @@
 SHELL=/usr/bin/env bash
 PROJECTNAME=$(shell basename "$(PWD)")
-LDFLAGS="-X 'main.buildTime=$(shell date)' -X 'main.lastCommit=$(shell git rev-parse HEAD)' -X 'main.semanticVersion=$(shell git describe --tags --dirty=-dev)'"
+LDFLAGS="-X 'main.buildTime=$(shell date)' -X 'main.lastCommit=$(shell git rev-parse HEAD)' -X 'main.semanticVersion=$(shell git describe --tags --dirty=-dev)' -X 'github.com/celestiaorg/celestia-node/nodebuilder/node.buildTime=$(shell date)' -X 'github.com/celestiaorg/celestia-node/nodebuilder/node.lastCommit=$(shell git rev-parse HEAD)' -X 'github.com/celestiaorg/celestia-node/nodebuilder/node.semanticVersion=$(shell git describe --tags --dirty=-dev)'"
```
Wondertan, line 3: Hmm, yeah, ok, sure, well, oh, nah, nooo. At least, we could extract the common cmds into the var within the makefile itself, but really, we should have one place to inject this data to.

docs/adr/adr-009-public-api.md
```diff
@@ -204,12 +204,17 @@ SyncHead(ctx context.Context) (*header.ExtendedHeader, error)
 
   type NodeModule interface {
     // Type returns the node type.
-    Type() node.Type
+    Type(context.Context) node.Type
     // Version returns information about the current binary build.
-    Version() string
+    Version(context.Context) string
```
Wondertan, line 209: * Do we really need to return the build version? Usually the version is accessed through the CLI
* We want to introduce API versioning, so if we want to show both the build and API versions, we should introduce a struct containing those already so that we break our API less.

Wondertan: Ah, I got confused. You return a structure, but this is an ADR, so we should update the ADR.

nodebuilder/node/node.go
```diff
@@ -0,0 +1,64 @@
...
+// Module defines the API related to interacting with the "administrative"
+// node.
+//
+//go:generate mockgen -destination=mocks/api.go -package=mocks . Module
+type Module interface {
+	// Type returns the node type.
+	Type(context.Context) Type
```
Wondertan, line 21: Do we need to implement JSONMarshal and Unmarshal for the type? It seems like it will work just by marshaling the enum number, but we may want to have human-readable

nodebuilder/node/admin.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"context"
+	"fmt"
+	"runtime"
+
+	"github.com/filecoin-project/go-jsonrpc/auth"
+	logging "github.com/ipfs/go-log/v2"
+)
+
+type admin struct {
```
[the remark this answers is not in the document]
renaynay: file name is `admin.go` bc i split it out from node module definition - can we leave as is or is it blocking?

docs/adr/adr-009-public-api.md
```diff
@@ -204,12 +204,17 @@ SyncHead(ctx context.Context) (*header.ExtendedHeader, error)
...
    
     // LogLevelSet sets the given component log level to the given level.
     LogLevelSet(ctx context.Context, name, level string) error
```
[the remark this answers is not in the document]
renaynay: why? it spits out error if it doesn't work.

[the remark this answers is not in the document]
renaynay: LogLevelSet in logger takes strings anyway and checks for validity

docs/adr/adr-009-public-api.md
renaynay: I'm actually okay to 
1. omit the build version from the Version endpoint
2. combine Type and the api info into one single struct called `Info` that contains fields `Type` and `APIVersion`

docs/adr/adr-009-public-api.md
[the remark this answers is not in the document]
Wondertan: It internally [converts](https://github.com/ipfs/go-log/blob/master/setup.go#L200) string to a specific type

nodebuilder/node/admin.go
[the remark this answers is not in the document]
Wondertan: we call out Module implementations as `moduler`. I don't see why we would violate this in the Node module

distractedm1nd (approved):

nodebuilder/node/admin.go
```diff
@@ -19,26 +18,17 @@ func newAdmin(tp Type) Module {
...
-}
-
-func (a *admin) Version(context.Context) Version {
-	return Version{
-		SemanticVersion: semanticVersion,
-		LastCommit:      lastCommit,
-		BuildTime:       buildTime,
-		SystemVersion:   fmt.Sprintf("%s/%s", runtime.GOARCH, runtime.GOOS),
-		GoVersion:       runtime.Version(),
+func (a *admin) AdminInfo(context.Context) Info {
```
Wondertan, line 28: Just Info, like the struct? Admin prefix feels like there is another less sensitive Info. Also, all methods on the Node struct are sort of an admin. Wait, actually, Info can be read or public. There is nothing wrong with API users knowing what API version they depend on and what node type they are connected to.

renaynay: Can't do `Info`. I wish we could, but prevented by `client.API`.

docs/adr/adr-009-public-api.md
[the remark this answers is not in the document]
renaynay: No need to add additional logic to duplicate the check that occurs inside the lib already.

Makefile
renaynay: removed

docs/adr/adr-009-public-api.md
renaynay: Implemented, but API version implementation will be later PR

docs/adr/adr-009-public-api.md
```diff
@@ -203,13 +203,16 @@ SyncHead(ctx context.Context) (*header.ExtendedHeader, error)
 ```go
 
   type NodeModule interface {
-    // Type returns the node type.
-    Type() node.Type
-    // Version returns information about the current binary build.
-    Version() string
-   
+    // AdminInfo returns administrative information about the node.
+    AdminInfo(context.Context) Info
```
distractedm1nd, line 207: info, error

distractedm1nd (changes requested):

docs/adr/adr-009-public-api.md
Wondertan, line 207: Update

nodebuilder/node/admin.go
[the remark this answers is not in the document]
Wondertan: @renaynay

Wondertan (commented):
LGTM modulo two comments

Wondertan (approved):
:ship:

distractedm1nd (approved):

liamsi (approved):

Issue #1306, opened 2022-11-01
Tests: Flaky Unit Test Cases


derrandz:
# Overview

While `make test-unit` sometimes the following test cases fail:

1. `TestRPCCallsUnderlyingNode`: _is being caused by the fact that we are launching at once both the client query and server listen function, and that causes a race condition such that the client queries the server before it starts_

2. `TestShareAvailable_DisconnectedFullNodes`: _is caused by [the context deadline being set to 3 seconds](https://github.com/celestiaorg/celestia-node/blob/main/share/availability/full/reconstruction_test.go#L186)_

```
--> Running unit tests
2022-11-01T12:18:18.209+0100    ERROR   header-module   header/header.go:72     initializing header store failed: no trusted peers
The p2p host is listening on:
*  /ip4/192.168.11.103/udp/2121/quic/p2p/12D3KooWLX9ANy1T4e8TsoJmKzT4vKpswW1mfDVxoZm3pRzXc7y5
*  /ip4/127.0.0.1/udp/2121/quic/p2p/12D3KooWLX9ANy1T4e8TsoJmKzT4vKpswW1mfDVxoZm3pRzXc7y5
*  /ip6/::1/udp/2121/quic/p2p/12D3KooWLX9ANy1T4e8TsoJmKzT4vKpswW1mfDVxoZm3pRzXc7y5
*  /ip4/192.168.11.103/tcp/2121/p2p/12D3KooWLX9ANy1T4e8TsoJmKzT4vKpswW1mfDVxoZm3pRzXc7y5
*  /ip6/::1/tcp/2121/p2p/12D3KooWLX9ANy1T4e8TsoJmKzT4vKpswW1mfDVxoZm3pRzXc7y5

--- FAIL: TestRPCCallsUnderlyingNode (0.29s)
    rpc_test.go:43: 
                Error Trace:    /Users/derrandz/Lab/celestia/celestia-node/api/rpc_test.go:43
                Error:          Received unexpected error:
                                RPC client error: sendRequest failed: Post "http://0.0.0.0:26655": dial tcp 0.0.0.0:26655: connect: connection refused
                Test:           TestRPCCallsUnderlyingNode
    controller.go:137: missing call(s) to *mocks.MockModule.Balance(is anything) /Users/derrandz/Lab/celestia/celestia-node/api/rpc_test.go:40
    controller.go:137: aborting test due to missing call(s)
FAIL
coverage: [no statements]
[...]

PR #1314, opened 2022-11-02
fix(gateway): switch to separate Listen/Serve

### LINKED ISSUE #1286: feat(rpc/gateway): move to ListenAndServe
We can use `ListenAndServe`, and get rid of the `listener` on the `Server` structs in gateway+rpc ~~by creating a new constructor that overrides the http server. Using ListenAndServe alone does not allow us to use port 0 in tests without losing the port that gets assigned. @walldiss pointed out we can create a new constructor and pass in a server from the `httptest` package.~~

        Update the server to not use Listener and make lifecycling over HTTP Server directly

_Originally posted by @Wondertan in https://github.com/celestiaorg/celestia-node/pull/1199#discussion_r1006542803_

distractedm1nd:
Closes #1286

This is mainly a fix for test flakiness.

### DIFF
--- a/api/gateway/server.go
+++ b/api/gateway/server.go
@@ -2,6 +2,7 @@ package gateway
 
 import (
 	"context"
+	"net"
 	"net/http"
 	"sync/atomic"
 	"time"
@@ -11,8 +12,10 @@ import (
 
 // Server represents a gateway server on the Node.
 type Server struct {
-	srv     *http.Server
-	srvMux  *mux.Router // http request multiplexer
+	srv      *http.Server
+	srvMux   *mux.Router // http request multiplexer
+	listener net.Listener
+
 	started atomic.Bool
 }
 
@@ -40,9 +43,14 @@ func (s *Server) Start(context.Context) error {
 		log.Warn("cannot start server: already started")
 		return nil
 	}
-	//nolint:errcheck
-	go s.srv.ListenAndServe()
+	listener, err := net.Listen("tcp", s.srv.Addr)
+	if err != nil {
+		return err
+	}
+	s.listener = listener
 	log.Infow("server started", "listening on", s.srv.Addr)
+	//nolint:errcheck
+	go s.srv.Serve(listener)
 	return nil
 }
 
@@ -57,6 +65,7 @@ func (s *Server) Stop(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+	s.listener = nil
 	log.Info("server stopped")
 	return nil
 }
@@ -80,5 +89,8 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 // ListenAddr returns the listen address of the server.
 func (s *Server) ListenAddr() string {
-	return s.srv.Addr
+	if s.listener == nil {
+		return ""
+	}
+	return s.listener.Addr().String()
 }
--- a/api/gateway/server.go
+++ b/api/gateway/server.go
@@ -2,7 +2,6 @@ package gateway
 
 import (
 	"context"
-	"net"
 	"net/http"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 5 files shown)


api/gateway/server.go
```diff
@@ -88,5 +80,5 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 // ListenAddr returns the listen address of the server.
 func (s *Server) ListenAddr() string {
-	return s.listener.Addr().String()
+	return s.srv.Addr
```
renaynay, line 83: does this give you the full addr + port

renaynay (approved):

distractedm1nd:
Also is a part of #1306

api/rpc_test.go
```diff
@@ -118,14 +131,15 @@ func setupNodeWithModifiedRPC(t *testing.T) (*nodebuilder.Node, *mockAPI) {
 		dasMock.NewMockModule(ctrl),
 	}
 
-	overrideRPCHandler := fx.Invoke(func(srv *rpc.Server) {
+	// given the behavior of fx.Invoke, this invoke will be called last as it is outside a module
```
renaynay, line 134:
```suggestion
	// given the behavior of fx.Invoke, this invoke will be called last as it is added at the root level module. For further information, check the documentation on fx.Invoke.
```

renaynay (approved):

api/rpc_test.go
renaynay: [blocking] ❤️

walldiss:
Do we still consider moving or rpc test infra to `httptest.Server` in future?

api/rpc_test.go
```diff
@@ -28,9 +29,21 @@ func TestRPCCallsUnderlyingNode(t *testing.T) {
...
 	nd, server := setupNodeWithModifiedRPC(t)
 	url := nd.RPCServer.ListenAddr()
-	client, err := client.NewClient(context.Background(), "http://"+url)
-	t.Cleanup(client.Close)
+	// we need to run this a few times to prevent the race where the server is not yet started
+	var (
+		rpcClient *client.Client
+		err       error
+	)
+	for i := 0; i < 3; i++ {
```
MSevey, line 37: [non blocking]
This type of retry is really useful in testing. Since it seems like we don't have a helper function for this, we should create one like so, maybe in a `test` package:

```go

// Retry will call 'fn' 'tries' times, waiting 'durationBetweenAttempts'
// between each attempt, returning 'nil' the first time that 'fn' returns nil.
// If 'nil' is never returned, then the final error returned by 'fn' is
// returned.
func Retry(tries int, durationBetweenAttempts time.Duration, fn func() error) (err error) {
	for i := 1; i < tries; i++ {
		err = fn()
		if err == nil {
			return nil
		}
		time.Sleep(durationBetweenAttempts)
	}
	return fn()
}
```

Then this example would turn into:
```go
	err = test.Retry(3, time.Second, func() error {
		return client.NewClient(ctx, "http://"+url)
	})
	require.NoError(t, err)
```

nodebuilder/testing.go
```diff
@@ -46,7 +48,9 @@ func TestNodeWithConfig(t *testing.T, tp node.Type, cfg *Config, opts ...fx.Opti
 	require.NoError(t, err)
 	cfg.Core.IP = ip
 	cfg.Core.RPCPort = port
-	cfg.RPC.Port = "26655"
+	freePort, err := tennet.GetFreePort()
```
Wondertan, line 51: As discussed, we go with 0 rather than this. #1322

Wondertan (changes requested):

api/rpc_test.go
distractedm1nd: #1332

distractedm1nd:
Now closes #1322 . Deprecates #1286 .

nodebuilder/testing.go
distractedm1nd: done

Wondertan (approved):
LGTM

renaynay (approved):

Wondertan (approved):

PR #1315, opened 2022-11-02
feat(.github): add labels on issues workflow

### LINKED ISSUE #1259: .github: label management workflow
To ensure all issues and PRs are labelled in a way that helps us manage the work in the node repo, I recommend using the github action listed in the documentation here: https://docs.github.com/en/actions/managing-issues-and-pull-requests/adding-labels-to-issues#creating-the-workflow

It can be found in the marketplace here: https://github.com/marketplace/actions/simple-issue-labeler

The pre-requisite is aligning on the labelling schema.

renaynay:
Resolves #1259

### DIFF
--- a/.github/workflows/issue-labels.yml
+++ b/.github/workflows/issue-labels.yml
@@ -2,10 +2,6 @@
 # They are provided by a third-party and are governed by
 # separate terms of service, privacy policy, and support
 # documentation.
-
-# GitHub recommends pinning actions to a commit SHA.
-# To get a newer version, you will need to update the SHA.
-# You can also reference a tag or branch, but the action may change without warning.
 name: Label issues
 on:
   issues:
@@ -22,3 +18,4 @@ jobs:
         with:
           add-labels: "needs-triage"
           repo-token: ${{ secrets.GITHUB_TOKEN }}
+          
--- a/.github/workflows/issue-labels.yml
+++ b/.github/workflows/issue-labels.yml
@@ -17,4 +17,5 @@ jobs:
         uses: mheap/github-action-required-labels@v2
         with:
           add-labels: "needs-triage"
+          ignore-if-labeled: true
           repo-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


Wondertan (commented):
So, any new issue will require setting a label, or will it be set automatically?

"Very nice" both ways

renaynay:
@Wondertan I've just pushed a new commit where I changed it so that it will only add `needs-triage` if the issue isn't already labeled once it's opened.

.github/workflows/issue-labels.yml
```diff
@@ -0,0 +1,21 @@
...
+name: Label issues
+on:
+  issues:
+    types: [opened]
+
+jobs:
+  label_issues:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
```
walldiss, line 14: Hmm, why does it need write permission?

.github/workflows/issue-labels.yml
```diff
@@ -0,0 +1,21 @@
...
+    steps:
+      - name: Label issues
+        uses: mheap/github-action-required-labels@v2
+        with:
+          add-labels: "needs-triage"
+          ignore-if-labeled: true
```
Wondertan, line 20: If I label an issue with `area:share` during issue creation, will the `ignore-if-labeled` ignore the `needs:triage`?

MSevey (approved):

.github/workflows/issue-labels.yml
```diff
@@ -0,0 +1,21 @@
...
+jobs:
+  label_issues:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - name: Label issues
+        uses: mheap/github-action-required-labels@v2
+        with:
+          add-labels: "needs-triage"
```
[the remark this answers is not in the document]
renaynay: good point, i will delete old label and make new

.github/workflows/issue-labels.yml
renaynay: it will not add the `needs:triage` label if a label is already added. Only if people open issues on node repo without adding any labels AFAIK

walldiss (approved):

Wondertan (approved):

Issue #1309, opened 2022-11-01
improvement(header/p2p): calculate best head on available peers in case of timeout.


walldiss:
### Implementation ideas

Currently bestHead could not be determined  if at least one of trusted peers didn't respond on get head request, [resulting in returned error](https://github.com/celestiaorg/celestia-node/blob/f5f5adb1f6e892e2e0cb8b97ab6349c80e01a611/header/p2p/exchange.go#L95). It is possible that at that point we have some info from other peers so it is better to use `bestHead` on available data instead of returning an error.

PR #1319, opened 2022-11-03
improvement(header/p2p): get bestHead if timeout is reached

vgonkivs:
Closes https://github.com/celestiaorg/celestia-node/issues/1309

### DIFF
--- a/header/p2p/exchange.go
+++ b/header/p2p/exchange.go
@@ -85,6 +85,7 @@ func (ex *Exchange) Head(ctx context.Context) (*header.ExtendedHeader, error) {
 	}
 
 	result := make([]*header.ExtendedHeader, 0, len(ex.trustedPeers))
+LOOP:
 	for range ex.trustedPeers {
 		select {
 		case h := <-headerCh:
@@ -92,7 +93,7 @@ func (ex *Exchange) Head(ctx context.Context) (*header.ExtendedHeader, error) {
 				result = append(result, h)
 			}
 		case <-ctx.Done():
-			return nil, ctx.Err()
+			break LOOP
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


distractedm1nd (dismissed):
nice!

header/p2p/exchange.go
```diff
@@ -85,14 +85,15 @@ func (ex *Exchange) Head(ctx context.Context) (*header.ExtendedHeader, error) {
...
+LOOP:
 	for range ex.trustedPeers {
 		select {
 		case h := <-headerCh:
 			if h != nil {
 				result = append(result, h)
 			}
 		case <-ctx.Done():
-			return nil, ctx.Err()
+			break LOOP
```
Wondertan, line 96: We might also want to limit the Head call itself with a deadline. We don't do it where call `Head` in Syncer. The question is: should the user of the API do this himself, or should we do the limit for each Head call internally like we do it in the Server?

Wondertan: Other than this Q, LGTM, clean solution

vgonkivs: I would suggest to let users decide themselves. We have done it for Server side as this is a dos vector - to request 1000 range request that will block the whole service. But I don't think it's relevant for the client side.

renaynay: Let's put the above into an issue bc it applies to every client-side call @Wondertan 

At the moment, let's leave it for user/caller to decide until we see it causes issues / can be problematic.

Wondertan: @vgonkivs, pls make an issue

vgonkivs: https://github.com/celestiaorg/celestia-node/issues/1349

Wondertan (changes requested):
Another thing I just realized about this PR. Instead of returning the expected context error on its cancellation, we may actually return a successful result. Such unexpected behaviors have to be explicitly documented, so pls, let's do this.

Wondertan:
@vgonkivs, the comment should mainly be on the `Head`, because the behaviors of the method are different from what users usually expect. Note that this package will soon be used by other teams and such details will be useful for them

renaynay (approved):

Wondertan (approved):

Issue #1323, opened 2022-11-03
nodebuilder: Rename sub-pkgs to `<subpkg>_constructor` to differentiate from root-level pkgs


renaynay:
So instead of it being 

`nodebuilder/state`, it should be `nodebuilder/state_constructor`

Wondertan:
[For the same reason](https://github.com/celestiaorg/celestia-node/pull/1324#discussion_r1014683336), we should change the namings. These are modules, not the constructors.

Jpserrat:
@renaynay @Wondertan  So now it will be <subpkg>_module?

Wondertan:
We haven't found a consensus within a team on this yet, so I would punt this for a while. 

Alternatively, we can quickly go with my following proposal if @renaynay and @distractedm1nd agree in this thread. I propose prepending each subpkg name in nodebuilder pkg with a `mod` prefix, so `fraud` would be `modfraud`.

saudch007:
@Wondertan @ramin @renaynay  Can I work on this one? I am really looking forward to consistent contributions.

renaynay:
I am find with `mod` prefix for the subpkg names.

Sorry for the delay on responding @saudch007 , of course feel free to work on this issue. Thank you

PR #1324, opened 2022-11-03
refactor(nodebuilder): Logger names in subpackages to `constructor/<subpkg>`

### LINKED ISSUE #1288: chore(nodebuilder): change `log` loggers to `$MODULENAME/constructor` instead of `$MODULENAME-module`
Self explanatory. Probably should be done after #1176

renaynay:
Resolves #1288

### DIFF
--- a/nodebuilder/fraud/module.go
+++ b/nodebuilder/fraud/module.go
@@ -8,7 +8,7 @@ import (
 	"github.com/celestiaorg/celestia-node/nodebuilder/node"
 )
 
-var log = logging.Logger("fraud-module")
+var log = logging.Logger("constructor/fraud")
 
 func ConstructModule(tp node.Type) fx.Option {
 	baseComponent := fx.Provide(func(module Module) fraud.Getter {
--- a/nodebuilder/fraud/module.go
+++ b/nodebuilder/fraud/module.go
@@ -8,7 +8,7 @@ import (
 	"github.com/celestiaorg/celestia-node/nodebuilder/node"
 )
 
-var log = logging.Logger("constructor/fraud")
+var log = logging.Logger("module/fraud")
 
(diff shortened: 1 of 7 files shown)


nodebuilder/fraud/module.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/celestiaorg/celestia-node/nodebuilder/node"
 )
 
-var log = logging.Logger("fraud-module")
+var log = logging.Logger("constructor/fraud")
```
Wondertan, line 11: The constructor is confusing. 
1. Constructor of the fraud references the fraud pkg more
2. This pkg could log out things beyond the module construction.

Let's do `modfraud` or `fraudmod` pls

distractedm1nd: So, the reasoning behind this is that the user won't know what a "nodebuilder" is nor will understand what the difference between "fraud" and "modfraud" packages would be. I like the constructor pattern but am open for more ideas. Not `modX` pilled yet, do you have any other ideas?

distractedm1nd (dismissed):

nodebuilder/fraud/module.go
Wondertan: The point is that the logger name should reference the fact that it is a module. The `constructor` does not do that.

Wondertan: Why not just `modfraud` or `mod/fraud`?

Wondertan (changes requested):
The logger is for the module, not only the constructor.
`modfraud` or `mod/fraud` or `module/fraud`

Related #1323

Wondertan (approved):
Thanks for changing this to module!

distractedm1nd (approved):
module/x much better than modX, thanks

vgonkivs (approved):

PR #1283, opened 2022-10-26
feat(rpc): tooling for openrpc spec autogeneration

distractedm1nd:
Closes #1188 

First prototype:
- [X] Example parameter types
- [X] Relying on explicitly defining default implementations for method discovery
- [x] Refined metadata
- [x] Scaffolding for example parameter types (`NOTE: finishing this will be its own PR`)

Long term(not this PR, issues to be created):
- [ ] Allow for embedded interfaces in nodebuilder Module interfaces
- [x] Autogenerated docs for methods that return a channel
- [x] Create Make targets for the code autogeneration
- [ ] Have CI generate the spec and host it for use with playground

PR #1325, opened 2022-11-04
feat(rpc): creating scaffolding for JWT authentication

distractedm1nd:
Based on #1283

Will have to be rebased on the next RPC PRs to make cleaner

TODO (in stacked PRs):
- [ ] token generation <- cli tool?
- [ ] token verification
- [ ] create docs


nodebuilder/header/service.go
```diff
@@ -22,16 +22,8 @@ type Module interface {
...
-// TODO(@distractedm1nd): These structs need to be autogenerated.
-type API struct {
-	GetByHeight func(context.Context, uint64) (*header.ExtendedHeader, error)
-	Head        func(context.Context) (*header.ExtendedHeader, error)
-	IsSyncing   func() bool
-}
-
 // Service represents the header service that can be started / stopped on a node.
-// Service's main function is to manage its sub-services. Service can contain several
+// Sjrvice's main function is to manage its sub-services. service can contain several
```
renaynay, line 26:
```suggestion
// Service manages its sub-services. Service can contain several
```

nodebuilder/p2p/p2p.go
```diff
@@ -188,22 +191,96 @@ func (m *module) PubSubPeers(topic string) []peer.ID {
...
+	return api.Internal.NATStatus()
+}
+
+func (api *API) BlockPeer(p peer.ID) error {
+	return api.Internal.BlockPeer(p)
+}
+
+func (api *API) UnblockPeer(p peer.ID) error {
+	return api.Internal.UnblockPeer(p)
+}
```
vgonkivs, line 250: Do we really need to be able to block/unblock peers from through API?

nodebuilder/p2p/p2p.go
```diff
@@ -188,22 +191,96 @@ func (m *module) PubSubPeers(topic string) []peer.ID {
...
+func (api *API) NATStatus() (network.Reachability, error) {
+	return api.Internal.NATStatus()
+}
+
+func (api *API) BlockPeer(p peer.ID) error {
+	return api.Internal.BlockPeer(p)
+}
+
+func (api *API) UnblockPeer(p peer.ID) error {
+	return api.Internal.UnblockPeer(p)
```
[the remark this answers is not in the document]
Wondertan: Yes. E.g for testground

nodebuilder/das/das.go
```diff
@@ -17,6 +19,16 @@ type Module interface {
 // API is a wrapper around Module for the RPC.
 // TODO(@distractedm1nd): These structs need to be autogenerated.
 type API struct {
-	SamplingStats func(ctx context.Context) (das.SamplingStats, error)
-	WaitCatchUp   func(ctx context.Context) error
+	Internal struct {
+		SamplingStats func(ctx context.Context) (das.SamplingStats, error) `perm:"read"`
+		WaitCatchUp   func(ctx context.Context) error                      `perm:"write"`
```
Wondertan, line 24: Tend to agree that this should be `write`, but still curios on your rationale

nodebuilder/state/state.go
```diff
@@ -79,47 +81,154 @@ type Module interface {
...
-	)
-	QueryDelegation    func(ctx context.Context, valAddr state.ValAddress) (*types.QueryDelegationResponse, error)
-	QueryUnbonding     func(ctx context.Context, valAddr state.ValAddress) (*types.QueryUnbondingDelegationResponse, error)
-	QueryRedelegations func(
-		ctx context.Context,
-		srcValAddr,
-		dstValAddr state.ValAddress,
-	) (*types.QueryRedelegationsResponse, error)
+	Internal struct {
+		AccountAddress    func(ctx context.Context) (state.Address, error)                      `perm:"read"`
```
Wondertan, line 85: Hmm, I don't think the address of the node should be public

nodebuilder/state/state.go
```diff
@@ -79,47 +81,154 @@ type Module interface {
...
+		IsStopped         func(ctx context.Context) bool                                        `perm:"read"`
+		Balance           func(ctx context.Context) (*state.Balance, error)                     `perm:"read"`
```
Wondertan, line 87: Same here

Wondertan (changes requested):
Very nice description! LGTM. 
Requesting changes due to privacy concerns of two API methods.

nodebuilder/state/state.go
```diff
@@ -79,47 +81,154 @@ type Module interface {
 // API is a wrapper around Module for the RPC.
 // TODO(@distractedm1nd): These structs need to be autogenerated.
 type API struct {
```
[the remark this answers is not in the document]
distractedm1nd: I would actually keep them in the same file as long as Module is at the top but long term they will end up in their own files once I create autogeneration for these structs - should I bite the bullet and do it now, or can it stay until then? Not very opinionated here

nodebuilder/das/das.go
distractedm1nd: Hmm. Yeah `WaitCatchUp` definitely seems like a write.
On the fence about `SamplingStats`, I can see the case for both read and write. Would love to hear from @walldiss

nodebuilder/state/state.go
distractedm1nd: good catch

nodebuilder/das/das.go
walldiss: `WaitCatchUp` will block client until catchup happens. Both methods don't mutate any data, so there is nothing to write, unless I misunderstand concept of `write` permission.

api/rpc/server.go
```diff
@@ -23,11 +30,19 @@ type Server struct {
 
 func NewServer(address, port string) *Server {
 	rpc := jsonrpc.NewServer()
+	authHandler := &auth.Handler{
+		Verify: func(ctx context.Context, token string) ([]auth.Permission, error) {
+			// TODO(distractedm1nd/renaynay): implement auth
+			log.Warn("auth not implemented, token: ", token)
+			return DefaultPerms, nil
```
walldiss, line 37: This probably has been discussed in API planning somewhere, so might be a silly question.
Looks like `read` == `public` right now. Is it possible that we might need to have  private `read` that allows reads only for a set of users?

walldiss (commented):
Oh it was hard to understand the problem with reflection. Nice and clever workaround for the given problem! Make it generated and noone will ever read it :)

nodebuilder/das/das.go
Wondertan: I saw this as a long-blocking operation that might have. Nothing to write but also nothing to read. Such blocking operations have DOS implications, so that I would hide them away(but that might be overcautious, as every single read op have similar issues) Although, read is fine if apply suggestion with separate `public` perm which SGTM

api/rpc/server.go
Wondertan: It's worth considering this idea @distractedm1nd @renaynay

distractedm1nd: this is what I meant by something between read and admin, i am not opposed. seems like the best option.

admin is too high, read too low, write is semantically incorrect

nodebuilder/das/das.go
walldiss: Oh, it is suppose to be a subscription, but with rpc it turned into long-living rpc call.  I'm not even sure we should expose it this way. At very least we should be careful a timeout on such call.

Wondertan: We can expose it so that our CLI can block and wait until catching up is done, but that does not require us to have full public perms and can be something private.

api/rpc/server.go
renaynay: do you propose the following levels then?: 

- public
- read (read + public)
- write (read, write, public)
- admin (all perms)

walldiss: I would purpose to keep read/write as access **type** and role as access **level**. So private write should not grant private read. I think the permission that does both is `readwrite`.

For our reduced case it will be:
 -  `public`
 -  private  level
    - `read`
    - `write`
    - `readwrite`
 -  admin level

distractedm1nd: public is the new default, and protected reads are behind `read`

@walldiss users can sign tokens with arbitrary permission sets, there is no actual hierarchy in this PR. If you sign a token with only the `write` perm you cannot use `read` methods. Discussion of whether requesting `write` perms should add `read` by default are in a PR on top of this

nodebuilder/state/state.go
distractedm1nd: done

nodebuilder/state/state.go
distractedm1nd: done

nodebuilder/das/das.go
distractedm1nd: both methods are behind read now

nodebuilder/state/state.go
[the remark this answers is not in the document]
distractedm1nd: @renaynay and I decided in sync to delay this until autogen is done, primarily because doing so now would make rebasing the stacked PRs a nightmare for little direct benefit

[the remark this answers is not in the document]
Wondertan: 👌🏻

Wondertan (approved):
LGTM. Some of the public could go into read instead, but we can revisit those once we revisit what each module should expose and ADR API compliance

vgonkivs (approved):
LGTM 🚀

renaynay (approved):

PR #1337, opened 2022-11-09
Update Docker image build pipelines

jbowen93:
# PULL REQUEST

## Overview

Updates the Docker image build Github Action Pipelines.

The `docker-build` job is run on all pushes to all branches. If pushing to the `main` or when any new tags are pushed It creates both a `linux/arm64` image and a `linux/amd64` image. When run against `main` the image is tagged with the short commit sha. When run against a tag the image is tagged with the git tag. 

If pushing to a non `main` branch it creates a  `linux/amd64` docker image that is tagged with the short commit sha.  

## Checklist


.github/workflows/main-docker-build.yml
```diff
@@ -0,0 +1,50 @@
+name: "main-docker-build"
+
+on:
+  push:
```
MSevey, line 4: Let's add Pull requests so we can combine the workflows
```suggestion
  pull_request:
  push:
```

.github/workflows/main-docker-build.yml
```diff
@@ -0,0 +1,50 @@
...
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build and push
+        uses: docker/build-push-action@v3
+        with:
+          platforms: linux/arm64, linux/amd64
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          file: docker/Dockerfile
```
MSevey, line 50: We can use an `if` condition to handle the different triggers. This allows us to only have the one workflow file that handles all cases. 

```suggestion
      - name: Build and Push amd Image
        if: ${{ github.event_name == 'pull_request' }}
        uses: docker/build-push-action@v3
        with:
          platforms: linux/amd64
          push: true
          tags: ${{ steps.meta.outputs.tags }}
          file: docker/Dockerfile
      - name: Build and Push All Images
        if: ${{ github.event_name != 'pull_request' }}
        uses: docker/build-push-action@v3
        with:
          platforms: linux/arm64, linux/amd64
          push: true
          tags: ${{ steps.meta.outputs.tags }}
          file: docker/Dockerfile
```

.github/workflows/amd64-docker-build.yml
```diff
@@ -1,4 +1,4 @@
-name: "docker-build"
```
MSevey, line 1: This file isn't needed

.github/workflows/labels.yml
```diff
@@ -13,6 +13,8 @@ jobs:
     runs-on: ubuntu-latest
```
MSevey, line 13: I think your fork is out of date. @renaynay pushed an update to fix the label action and I think changed the name of the file as well.

MSevey (changes requested):
We can combine the docker builds into one workflow with the use of the `if` condition.  I think you also need to update your fork to pull in the updates to the label action that Rene made that should also remove the `label.yml` file as it was renamed I believe.


Then the workflow action can also just be `docker-build.yml` again.

sysrex (approved):
Looks good. But I do agree with @MSevey that they could be merge into a single workflow.

jbowen93:
I updated it to remain using only the `push` action but set conditionals on the individual steps. 
- build `amd64` images on every push *except* pushes to main and tags 
- build `amd64 + arm64` images on pushes to main and tags

I presume it isn't running the github action because there's a disconnect between the github action that's configured on the main branch and the one in the PR.

.github/workflows/docker-build.yml
```diff
@@ -4,10 +4,11 @@ on:
   push:
```
MSevey, line 4: > I presume it isn't running the github action because there's a disconnect between the github action that's configured on the main branch and the one in the PR.

It isn't running because you have the trigger to only be push, tags, and workflow dispatch. You need to add `pull_request` for it to trigger here.

MSevey, line 4:
```suggestion
  pull_request:
  push:
```

.github/workflows/docker-build.yml
```diff
@@ -4,10 +4,11 @@ on:
...
     branches:
       - "**"
```
MSevey, line 6: This should be main, then you don't need to have the `github.ref` check down below. 
```suggestion
      - main
```

.github/workflows/docker-build.yml
```diff
@@ -21,40 +22,41 @@ jobs:
...
-
-      - name: Build and push
+      - name: Build and Push amd64 Image
+        if: |
+          github.ref != 'refs/heads/main' &&
+          !startsWith(github.ref, 'refs/tags/v')
+        uses: docker/build-push-action@v3
+        with:
+          platforms: linux/amd64
+          push: true
```
MSevey, line 50: Do we need to push on all Pull Request commits? That seems like like a lot. Can't we just verify that the image builds? 

Or is the reason for pushing on all PR commits to run the smoke test? If so, maybe we can build the run the tests without pushing the image and cluttering up the image repo

.github/workflows/docker-build.yml
```diff
@@ -21,40 +22,41 @@ jobs:
...
         with:
           registry: ghcr.io
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
-
-      - name: Build and push
+      - name: Build and Push amd64 Image
+        if: |
+          github.ref != 'refs/heads/main' &&
+          !startsWith(github.ref, 'refs/tags/v')
```
MSevey, line 46: We should build the images on pull requests to verify they build. We can then just not push them. 
```suggestion
        if: ${{ github.event_name == 'pull_request' }}
```

.github/workflows/docker-build.yml
```diff
@@ -21,40 +22,41 @@ jobs:
...
+        uses: docker/build-push-action@v3
+        with:
+          platforms: linux/amd64
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          file: docker/Dockerfile
+      - name: Build and Push amd64+arm64 Images
+        if: |
+          github.ref == 'refs/heads/main' ||
+          startsWith(github.ref, 'refs/tags/v')
```
MSevey, line 56: With the changes to the triggers, we can just check for the push event type because we are already limiting to push events on `main` and `v*` tags
```suggestion
        if: ${{ github.event_name == 'push' }}
```

MSevey (changes requested):

.github/workflows/docker-build.yml
jbowen93: See a test example here: https://github.com/jbowen93/celestia-node/pull/8 

Once the GH action configuration is on main pushes to other branches will also trigger a build during PRs.

.github/workflows/docker-build.yml
MSevey: testing just `push` event due to the `if` condition filtering

MSevey (approved):

Wondertan (approved):
Thanks, @MSevey, for the deep review. I made a glimpse and I don't see nothing criminal in here, so LGTMerged

PR #1338, opened 2022-11-10
chore!: bump celestia-app to v0.10.0-rc1

evan-forbes:
## Overview

bumps to the latest release candidate of celestia-app. This bump is very breaking, as we are updating the default codec introduced in v0.7.0 of rsmt2d. The integration tests are passing for me locally, and I don't have permissions to try them again here.

## Checklist

- [x] Required CI checks are passing <- passing locally, but I don't have permission to retry tests


evan-forbes:
The ReadEDS test will fail cause we need to regenerate the car file. I saw the WriteEDS function, but I'm unsure of the data that we need to write. 

cc @Wondertan

evan-forbes:
pls feel free to push to this branch to regenerate the car test file if I'm asleep

share/eds/byzantine/bad_encoding.go
```diff
@@ -150,13 +150,16 @@ func (p *BadEncodingProof) Validate(header *header.ExtendedHeader) error {
...
-	rebuiltExtendedShares, err := codec.Encode(rebuiltShares[0 : len(shares)/2])
+
+	squareWidth := uint64(len(header.DAH.RowsRoots) / 2)
+
+	rebuiltExtendedShares, err := codec.Encode(rebuiltShares[0:squareWidth])
 	if err != nil {
 		return err
 	}
-	rebuiltShares = append(rebuiltShares, rebuiltExtendedShares...)
+	copy(rebuiltShares[squareWidth:], rebuiltExtendedShares)
```
evan-forbes, line 160: I'm not sure how this was working before, or why switching the codec broke this. this seems like what we're trying to do, but I could be wrong. If we don't do this, we end up with a slice that has a len == 6, which will break the nmt wrapper.

share/ipld/nmt_test.go
```diff
@@ -21,10 +21,8 @@ func TestNamespaceFromCID(t *testing.T) {
 	var tests = []struct {
 		randData [][]byte
 	}{
-		{randData: generateRandNamespacedRawData(4, appconsts.NamespaceSize, appconsts.ShareSize)},
-		{randData: generateRandNamespacedRawData(16, 16, appconsts.ShareSize)},
-		{randData: generateRandNamespacedRawData(4, 4, appconsts.ShareSize)},
```
evan-forbes, line 26: changed these tests to only use the standard size. we can switch them back if we want. The new codec has to use slices that exactly 512 bytes, so if we do want to switch them back we just have to adjust for that

Wondertan:
> > The ReadEDS test will fail cause we need to regenerate the car file. I saw the WriteEDS function, but I'm unsure of the data that we need to write.

> cc @Wondertan

cc @distractedm1nd

walldiss:
I've fixed testdate issue. To generate new testdata there is `createTestData` func.

walldiss (approved):
Looks good! :rocket:

share/ipld/nmt_test.go
Wondertan: The exact size is fine, imo. We don't really care about non-canonical sizes

share/eds/byzantine/bad_encoding.go
Wondertan: Seems like the new implementation uses slices slightly differently.

@vgonkivs, pls review this change

share/eds/byzantine/bad_encoding.go
```diff
@@ -138,7 +138,7 @@ func (p *BadEncodingProof) Validate(header *header.ExtendedHeader) error {
 		if share == nil {
 			continue
 		}
-		shares[index] = share.Share
```
Wondertan, line 141: This seems suspicious. @evan-forbes, do you know why we now need to cut the namespace?

evan-forbes: yes what @walldiss said. @Wondertan, perhaps it would be better to implement this change elsewhere? https://github.com/celestiaorg/celestia-node/blob/a95b114cf0353a4760314bcd198e8cd9c831f83b/share/eds/byzantine/share_proof.go#L91

unlike the old codec, the new one requires chunks of data be length 256 or 512. I'm not the most familiar with this codebase, so I'm far from certain, but it might have been a bug if we're BEFPs using the extra namespace added by the nmt wrapper to the erasure code.

vgonkivs: Wait, so now there are twice prepended namespace instead of 1(as we received before), right? Or I missed something?

share/eds/byzantine/bad_encoding.go
vgonkivs: looks correct

Wondertan: @evan-forbes, does this mean that the klauspost leopard was introduced in v0.10.0? I thought we migrated before, but we didn't have these issues.

share/eds/byzantine/bad_encoding.go
Wondertan: There must be a bug somewhere I feel

walldiss: Well, i was wrong with 2 namespaces assumption.  The thing is the share that is passed to validator is original share (1 namespace + 512 bytes). Later, inside the validation function namespace will be cut .
 But the `shares` slice that is filled in line 141 contains only Share data (512 bytes), later elements of this slice passed to Decoder. 
 It used to work before because Decodor didn't check `Share` length and consumed it with prepended namespace without any sign of error.

evan-forbes: found the bug if my logic is correct, pls see https://github.com/celestiaorg/celestia-node/issues/1355 and https://github.com/celestiaorg/celestia-node/pull/1354

evan-forbes: the test in #1354 passes on this branch but fails on main

share/eds/byzantine/bad_encoding.go
evan-forbes: I forgot to switch the default code :sweat_smile: in v0.9.0, although we were using rsmt2d v0.7.0

share/eds/byzantine/bad_encoding.go
```diff
@@ -150,13 +150,16 @@ func (p *BadEncodingProof) Validate(header *header.ExtendedHeader) error {
 	if err != nil {
 		return err
 	}
-	rebuiltExtendedShares, err := codec.Encode(rebuiltShares[0 : len(shares)/2])
+
+	squareWidth := uint64(len(header.DAH.RowsRoots) / 2)
```
renaynay, line 154: can we still leave this as taking the square width from `merkleRowRoots` ? (since that's len(shares)) -- it's not a diff length and adds a bit of cognitive overhead.

renaynay (changes requested):

share/eds/byzantine/bad_encoding.go
evan-forbes: sorry about that, forgot I switched while debugging. sure thing! https://github.com/celestiaorg/celestia-node/pull/1338/commits/6986d738d3de5418319f53894b11380e9956dece

Wondertan:
@evan-forbes, I will extract the BEFP fix from here into #1354, so this PR is only the update, and #1354 has the fix and tests proving the problem.

share/ipld/nmt_test.go
```diff
@@ -21,10 +21,8 @@ func TestNamespaceFromCID(t *testing.T) {
...
-		{randData: generateRandNamespacedRawData(4, appconsts.NamespaceSize, appconsts.ShareSize/2)},
+		{randData: generateRandNamespacedRawData(4, appconsts.NamespaceSize, appconsts.ShareSize-appconsts.NamespaceSize)},
+		{randData: generateRandNamespacedRawData(16, appconsts.NamespaceSize, appconsts.ShareSize-appconsts.NamespaceSize)},
```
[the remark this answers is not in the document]
Wondertan: Not sure, seems like a relic from the past. These tests should be rewritten.

Issue #1117, opened 2022-09-16
share: Extend `GetSharesByNamespace` with an option to collect Merkle proofs


Wondertan:
## Context

We want FN/BNs to serve data by namespace as a whole. For that, we need to read Merkle proofs while getting data by namespace so that requestors can verify its inclusion in the header.

## Implementation Details

- `share.GetSharesByNamespace` is used to load data from disk only and not from the network anymore
  - Using [`Blockservice`](https://github.com/ipfs/go-blockservice/blob/master/blockservice.go#L46) with [offline exchange](https://github.com/ipfs/go-ipfs-exchange-offline/blob/master/offline.go#L16)
  - Using [`Blockstore`](https://github.com/ipfs/go-ipfs-blockstore/blob/master/blockstore.go#L33) provided by `EDSStore`
- `share.GetSharesByNamespace` is extended to return NMT Merkle proofs
  - Similar to `share/ipld.GetProofsForShares`
  - Ensure Merkle proofs are not duplicated!
  
 ## Refs
 
 Depends on #878 
 Depends on #183

PR #1339, opened 2022-11-10
feat(share): add get proof by namespace

walldiss:
Resolves https://github.com/celestiaorg/celestia-node/issues/1117


vgonkivs:
It seems that `GetSharesWithProofByNameSpace` and `GetSharesByNamespace` are pretty similar. I'm wondering, can leave only one implementation but pass additional param(e.g. `withProof` flag)

wdyt @walldiss , @Wondertan ?

Wondertan:
Yeah, we should do this. Leaving this to @walldiss to decide whether we should do in this PR or make a tech debt/refactoring issues

share/get_proof_test.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/minio/sha256-simd"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/celestiaorg/celestia-node/share/ipld"
+)
+
+func TestGetSharesWithProofsByNamespace(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
distractedm1nd, line 20:
```suggestion
	t.Cleanup(cancel)
```

share/get_proof_test.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	for i, tt := range tests {
+		t.Run(strconv.Itoa(i), func(t *testing.T) {
+			rand.Seed(time.Now().UnixNano())
+			// choose random range in shares f
+			from := rand.Intn(len(tt.rawData) - 1)
+			to := rand.Intn(len(tt.rawData) - 1)
+
+			//
```
distractedm1nd, line 38: comment here or remove

share/get_proof_test.go
```diff
@@ -0,0 +1,88 @@
...
+			nID := expected[:NamespaceSize]
+
+			if to < from {
+				tmp := from
+				from, to = to, tmp
+			}
+
+			// change rawData to contain several shares with same nID
+			for i := from; i <= to; i++ {
+				tt.rawData[i] = expected
```
distractedm1nd, line 49: could you maybe only write the nID instead of using the same share?

share/get_proof_test.go
```diff
@@ -0,0 +1,88 @@
...
+			to := rand.Intn(len(tt.rawData) - 1)
+
+			//
+			expected := tt.rawData[from]
+			nID := expected[:NamespaceSize]
+
+			if to < from {
+				tmp := from
+				from, to = to, tmp
+			}
```
distractedm1nd, line 45:
```suggestion
			if to < from {
				tmp := from
				from, to = to, tmp
			}
			
			expected := tt.rawData[from]
			nID := expected[:NamespaceSize]
```

share/get_proof_test.go
distractedm1nd: i see that it would make the share validation check more complicated though :/ but whatever you think is a better test, ill trust

share/get_proof_test.go
```diff
@@ -0,0 +1,88 @@
...
+			for _, row := range eds.RowRoots() {
+				rcid := ipld.MustCidFromNamespacedSha256(row)
+				rowShares, err := GetSharesWithProofsByNamespace(ctx, bServ, rcid, nID, len(eds.RowRoots()))
+				require.NoError(t, err)
+				if rowShares != nil {
+					// append shares to check integrity later
+					shares = append(shares, rowShares.Shares...)
+
+					// construct nodes from shares by prepending namespace
+					var leafs [][]byte
```
distractedm1nd, line 66:
```suggestion
					var leaves [][]byte
```

share/get_test.go
```diff
@@ -216,10 +216,10 @@ func TestGetLeavesByNamespace_IncompleteData(t *testing.T) {
 	err = bServ.DeleteBlock(ctx, r.Cid())
 	require.NoError(t, err)
 
-	nodes, err := ipld.GetLeavesByNamespace(ctx, bServ, rcid, nid, len(shares))
-	assert.Equal(t, nil, nodes[1])
+	nodes, err := ipld.GetLeavesByNamespace(ctx, bServ, rcid, nid, len(shares), false)
+	assert.Equal(t, nil, nodes.Leaves[1])
 	// TODO(distractedm1nd): Decide if we should return an array containing nil
```
distractedm1nd, line 221: this TODO can be removed

share/ipld/get.go
```diff
@@ -383,7 +429,8 @@ func (b *fetchedBounds) update(index int64) {
 // job represents an encountered node to investigate during the `GetLeaves`
 // and `GetLeavesByNamespace` routines.
 type job struct {
-	id  cid.Cid
-	pos int
-	ctx context.Context
+	id       cid.Cid
+	sharePos int
```
distractedm1nd, line 433: @Wondertan this is exactly why I'm glad we didn't abstract job :)

share/ipld/get.go
```diff
@@ -367,8 +414,7 @@ type fetchedBounds struct {
 func (b *fetchedBounds) update(index int64) {
 	lowest := atomic.LoadInt64(&b.lowest)
 	// try to write index to the lower bound if appropriate, and retry until the atomic op is successful
-	// CAS ensures that we don't overwrite if the bound has been updated in another goroutine after the
-	// comparison here
+	// CAS ensures that we don't overwrite if the bound has been updated in another goroutine after the comparison here
```
distractedm1nd, line 417: does this pass line wrap?

share/ipld/get.go
```diff
@@ -42,6 +42,13 @@ var NumConcurrentSquares = 8
 //		so that workers spawned between each reconstruction for every new block are reused.
 var pool = workerpool.New(NumWorkersLimit)
 
+// LeavesWithProofs contains data with corresponding Merkle Proof
+type LeavesWithProofs struct {
```
distractedm1nd, line 46: can you doc that the proof fields are empty if the user didn't request proofs?

share/ipld/get.go
```diff
@@ -216,17 +225,44 @@ func GetLeavesByNamespace(
...
+				}
+
+				var proofs [][]byte
+				// left side traversed in bottom-up order
+				for i := range leftProofs {
+					if leftProofs[i] != nil {
+						proofs = append(proofs, leftProofs[i])
+					}
+				}
+				// right side of the tree will be traversed from up to bottom,
```
distractedm1nd, line 252:
```suggestion
				// right side of the tree will be traversed from top to bottom,
```

share/ipld/get.go
```diff
@@ -216,17 +225,44 @@ func GetLeavesByNamespace(
...
+				// so append in reversed order
+				for i := range rightProofs {
+					if rightProofs[len(rightProofs)-i-1] != nil {
+						proofs = append(proofs, rightProofs[len(rightProofs)-i-1])
+					}
```
distractedm1nd, line 257: Can you check if this loop would be more readable if you use a normal for loop that just counts down instead of up? but if you don't like it we can keep it this way

share/ipld/get.go
```diff
@@ -245,39 +281,50 @@ func GetLeavesByNamespace(
...
 				// this node has links in the namespace, so keep walking
 				for i, lnk := range links {
 					newJob := &job{
 						id: lnk.Cid,
+						// sharePos represents potential share position in share slice
+						sharePos: j.sharePos*2 + i,
 						// position represents the index in a flattened binary tree,
 						// so we can return a slice of leaves in order
-						pos: j.pos*2 + i,
+						pos: j.pos*2 + i + 1,
```
distractedm1nd, line 310: I don't understand why we are adding 1 now?

distractedm1nd (commented):
Looks great!

share/ipld/get.go
walldiss: Code here was working, but comment was incorrect. Old `j.pos*2 + i` is tracking share coordinates in shares array. `j.pos*2 + i + 1` operation maps tree into 1d struct without overwrites. We need second one here to collect proofs in tree, because second one will override values for node links (node children).

Example for old algo:
root ids: 0
root links: 2*0 (overwrite), 2*0+1 (1)
next links: 0*2 (overwrite), 2*0+1 (overwrite), 2*1+1 (3), 2*1+1 (4)

share/get_proof_test.go
walldiss: Yeah, it gets more complicated.

share/get_test.go
walldiss: Removed

share/ipld/get.go
walldiss: Looks cleaner with counting down, thank you.

share/get_proof.go
```diff
@@ -0,0 +1,51 @@
...
+	shares := make([]Share, 0, nodes.ProofEnd-nodes.ProofStart)
+	for _, leaf := range nodes.Leaves {
+		if leaf != nil {
+			shares = append(shares, leafToShare(leaf))
+		}
+	}
+
+	return &SharesWithProofs{
+		Shares: shares,
+		Proof:  nmt.NewInclusionProof(nodes.ProofStart, nodes.ProofEnd, nodes.Proofs, true),
```
renaynay, line 49: `ignoreMaxNamespace` should be some const we abide by throughout the codebase, no?

share/get_proof_test.go
```diff
@@ -0,0 +1,86 @@
...
+	}{
+		{rawData: RandShares(t, 4)},
+		{rawData: RandShares(t, 16)},
+		{rawData: RandShares(t, 64)},
+	}
+
+	for i, tt := range tests {
+		t.Run(strconv.Itoa(i), func(t *testing.T) {
+			rand.Seed(time.Now().UnixNano())
+			// choose random range in shares f
```
renaynay, line 34: f?

renaynay (commented):
intermediate comments, haven't reviewed fully.

share/ipld/get.go
distractedm1nd: np

share/ipld/get.go
distractedm1nd: Ohhhhhhhhhhh! Ok thanks for the explanation

share/get_proof.go
walldiss: It is and option from NMT that is now always set to `true` without using consts. I've created const in our codebase to have single point of reference.

share/get_proof_test.go
walldiss: To pay respect...

share/ipld/get.go
walldiss: doced

share/ipld/get.go
```diff
@@ -170,22 +179,25 @@ func GetLeaves(ctx context.Context,
...
+// implementation is based on `GetShares`.
 func GetLeavesByNamespace(
 	ctx context.Context,
 	bGetter blockservice.BlockGetter,
 	root cid.Cid,
 	nID namespace.ID,
 	maxShares int,
-) ([]ipld.Node, error) {
+	collectProofs bool,
+) (*LeavesWithProofs, error) {
```
Wondertan, line 195: api suggestion:
```go
proofs := Proofs{} // create proofs struct
GetLeavesByNamespace(..., WithProofs(proofs)) // supply it as an option
// use the populated proofs struct
```

Wondertan: I would really love to simplify the API to something like this. Do we think it is worth the time?

Wondertan: The fact that we return the `LeavesWithProofs` struct always, even when proofs are not requested makes me feel bad

share/get_proof_test.go
Wondertan: Pressed

share/ipld/get.go
Wondertan: Ok, we decided to avoid options for now and just to pass Proofs struct as the last params. If you don't need proofs, pass nil

share/get_proof.go
```diff
@@ -0,0 +1,52 @@
...
+	// Shares are the full data including namespace
+	Shares []Share
+	// Proof is a Merkle Proof for the corresponding Shares
+	Proof nmt.Proof
+}
+
+// GetSharesWithProofsByNamespace walks the tree of a given root and returns its shares within the
+// given namespace.ID. If a share could not be retrieved, err is not nil, and the returned array
+// contains nil shares in place of the shares it was unable to retrieve.
+func GetSharesWithProofsByNamespace(
```
Wondertan, line 25: Why not follow the same approach by passing nmt.Proof optionally?

walldiss: Yeah, it is better to update this func too. But I'd opt for using `*ipld.Proof` instead, since it could be encoded for wire transfer and `nmt.Proof` doesn't have such option.

share/ipld/get.go
```diff
@@ -170,22 +179,25 @@ func GetLeaves(ctx context.Context,
...
 	if len(nID) != NamespaceSize {
 		return nil, fmt.Errorf("expected namespace ID of size %d, got %d", NamespaceSize, len(nID))
 	}
 
-	ctx, span := tracer.Start(ctx, "get-leaves-by-namespace")
+	ctx, span := tracer.Start(ctx, "get-leaves-with-proofs-by-namespace")
```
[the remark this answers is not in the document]
walldiss: I think you might have reviewed previous version, because I've update this line yesterday.

share/get_proof.go
```diff
@@ -0,0 +1,51 @@
...
+type SharesWithProofs struct {
+	// Shares are the full data including namespace
+	Shares []Share
+	// Proof is a Merkle Proof for the corresponding Shares
+	Proof nmt.Proof
+}
+
+// GetSharesWithProofsByNamespace walks the tree of a given root and returns its shares within the
+// given namespace.ID. If a share could not be retrieved, err is not nil, and the returned array
+// contains nil shares in place of the shares it was unable to retrieve.
```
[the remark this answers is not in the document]
walldiss: Thank you, good find here and in other comments! I've fixed this comment block.

share/get_test.go
```diff
@@ -219,10 +219,9 @@ func TestGetLeavesByNamespace_IncompleteData(t *testing.T) {
 	err = bServ.DeleteBlock(ctx, r.Cid())
 	require.NoError(t, err)
 
-	nodes, err := ipld.GetLeavesByNamespace(ctx, bServ, rcid, nid, len(shares))
-	assert.Equal(t, nil, nodes[1])
-	// TODO(distractedm1nd): Decide if we should return an array containing nil
-	assert.Equal(t, 4, len(nodes))
+	nodes, err := ipld.GetLeavesByNamespace(ctx, bServ, rcid, nid, len(shares), false)
+	assert.Equal(t, nil, nodes.Leaves[1])
```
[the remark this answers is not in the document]
walldiss: Updated, thank you for suggestion!

share/ipld/get.go
```diff
@@ -170,16 +176,21 @@ func GetLeaves(ctx context.Context,
...
-// A non-nil error means that only partial data is returned, because at least one share retrieval
-// failed The following implementation is based on `GetShares`.
+// GetLeavesByNamespace returns leaves and corresponding proof that could be used to verify leaves
+// inclusion. It as many leaves from the given root with the given namespace.ID as it can retrieve.
+// If no shares are found, it returns both data and error as nil. If non-nil proofToFill param
+// passed, it will be filled with data required for inclusion verification. A non-nil error means
+// that only partial data is returned, because at least one share retrieval failed The following
+// implementation is based on
+
+// `GetShares`.
```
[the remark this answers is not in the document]
walldiss: Wow ,what a weird formatting, fixed

share/ipld/get.go
```diff
@@ -170,16 +176,19 @@ func GetLeaves(ctx context.Context,
...
+// param passed, it will be filled with data required for inclusion verification. A non-nil error
+// means that only partial data is returned, because at least one share retrieval failed. The
+// following implementation is based on `GetShares`.
 func GetLeavesByNamespace(
 	ctx context.Context,
 	bGetter blockservice.BlockGetter,
 	root cid.Cid,
 	nID namespace.ID,
 	maxShares int,
+	proofToFill *Proof,
```
distractedm1nd, line 191: [non-blocking] [optional] I don't like this name 100% but also cannot think of a better one. `proofContainer` maybe? But idk, more of just a comment in case someone else reading this has a better idea

distractedm1nd (dismissed):
I like this a lot, great work

share/get_proof_test.go
distractedm1nd: F

renaynay: F

share/ipld/get.go
walldiss: How about `optionalProof`?

share/ipld/get.go
Wondertan: Not sure what the reference is here. I don't remember anyone proposing job abstraction.

share/ipld/get.go
```diff
@@ -235,7 +273,7 @@ func GetLeavesByNamespace(
```
MSevey, line 273: commenting here because Github doesn't let you comment on code outside of the changes...

[non-blocking]
Unless I'm missing something, we are calling `wg.add()`, significantly more that `wg.done()`
We have one `defer wg.done()` call here but we are calling `wg.add()` in a loop here https://github.com/celestiaorg/celestia-node/pull/1339/files#diff-2ff88a4977c65c215324c40905aa73d0f8a9e09c6715437e058dc0fa95d2c478R335

additionally, we start off with a `wg.add` call here https://github.com/celestiaorg/celestia-node/pull/1339/files#diff-2ff88a4977c65c215324c40905aa73d0f8a9e09c6715437e058dc0fa95d2c478R217 with no corresponding `wg.done` call. 

Where are the rest of the `wg.done` calls?

I know this is out of scope of this PR, I'd just like to understand the flow here better.

MSevey: this comment needs more context imo https://github.com/celestiaorg/celestia-node/pull/1339/files#diff-2ff88a4977c65c215324c40905aa73d0f8a9e09c6715437e058dc0fa95d2c478R334

Wondertan: Unfortunately, links you've shared are not working for me

Wondertan: I got the question, though.

Wondertan: All `wg.add`s happen for each CID/hash in the tree. They are added right before they are queued and `wg,done`ed in defer as the end of the  processing.

The root `wg.add` is called for the first root CID before processing starts. While being processed, we extract two child links/cids/hashes, `wg.add` and queue them, and lastly the defer`wg.done` for the root is invoked. After the newle extracted links are processed in the same way, with the respective defers being called. Altogether, the amount of `adds` and `dones` is always the same.

share/get_test.go
```diff
@@ -301,17 +302,95 @@ func TestGetLeavesByNamespace_MultipleRowsContainingSameNamespaceId(t *testing.T
...
+		{rawData: RandShares(t, 4)},
+		{rawData: RandShares(t, 16)},
+		{rawData: RandShares(t, 64)},
+	}
+
+	for i, tt := range tests {
+		t.Run(strconv.Itoa(i), func(t *testing.T) {
+			rand.Seed(time.Now().UnixNano())
+			// choose random range in shares slice
+			from := rand.Intn(len(tt.rawData) - 1)
```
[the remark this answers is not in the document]
walldiss: Thank you, TIL `Intn` is half-open interval [0,n)

share/ipld/get.go
walldiss: @Wondertan explained it well. I want to add small detail that WaitGroup is a counter for amount of goruotine running in parallel. For every node being processed by workerpool a `Job` is created. When node is processed and goroutine exits it calls `Done`.

share/ipld/get.go
walldiss: Updated for `proofContainer`

share/ipld/get.go
```diff
@@ -245,39 +283,50 @@ func GetLeavesByNamespace(
...
 						// structure
 						ctx: ctx,
 					}
-
-					// if the link's nID isn't in range we don't need to create a new job for it
+					// if the link's nID isn't in range we don't need to create a new job for it,
+					// but need to collect a proof
 					jobNid := NamespacedSha256FromCID(newJob.id)
-					if nID.Less(nmt.MinNamespace(jobNid, nID.Size())) || !nID.LessOrEqual(nmt.MaxNamespace(jobNid, nID.Size())) {
+					if nID.Less(nmt.MinNamespace(jobNid, nID.Size())) {
```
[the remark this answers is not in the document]
walldiss: Yeah this is tricky part, added suggested comments

share/ipld/get.go
```diff
@@ -216,84 +225,131 @@ func GetLeavesByNamespace(
...
 	leaves := make([]ipld.Node, maxShares)
 
+	// if non-nil proof container provided, fill it while traversing the tree
+	var collectProofs = proofContainer != nil
+	var leftProofs, rightProofs [][]byte
+	if collectProofs {
+		// TODO: (@walldiss) this is massively overallocating and should be optimized later with clever
+		// append
+		leftProofs = make([][]byte, BatchSize(maxShares))
+		rightProofs = make([][]byte, BatchSize(maxShares))
```
Wondertan, line 235: It is a massive overallocation. `BatchSize` gives the number of nodes in the whole EDS while we work here within one row. You can use `((n*2)-1)-n`, where `(n*2)-1)` is the number of nodes in the tree and `-n` to cut the shares. We can also try to cut these in half more and simplify formula to get left and right proofs exactly.

walldiss: Half the size won't work here, because both arrays represent full tree with exactly `BatchSize` amount of nodes.

Ideally size of each array could be reduced to `treeHeight`. It will require smart append, because nodes needs to be ordered  according to indexes. I wanted to make it as separate PR but can add here as extra commit.

Wondertan: Extra commit is fine

walldiss: Done. I also like how GetProofsByNamespace became a bit smaller and simpler now, because I've offload proof collection logic into another simple component.

share/ipld/proof.go
```diff
@@ -0,0 +1,76 @@
...
+	defer c.leftLk.Unlock()
+	c.leftProofs = append(c.leftProofs, nodeWithIdx{
+		idx:  idx,
+		node: node,
+	})
+}
+
+func (c *proofCollector) addRight(idx int, node []byte) {
+	c.rightLk.Lock()
+	defer c.rightLk.Unlock()
```
Wondertan, line 47: Can we try to keep `GetShareByNamespace` lock-free? These are now the only locks there. What if we treat `idx` as a slice index?

Wondertan (commented):
Very nice refactor! One comment

share/ipld/proof.go
walldiss: Good suggestion, thank you. I've realised there is no need for locks, because there is actually no races. Updated.

Wondertan (dismissed):
Very nice! 🚢

renaynay (approved):

share/ipld/proof.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+// Proof contains information required for Leaves inclusion validation.
+type Proof struct {
+	Nodes      []cid.Cid
+	Start, End int
+}
+
+// proofCollector collects proof nodes CIDs provides CIDs array that
+// could be used to construct nmt.Proof for shares inclusion validation.
```
distractedm1nd, line 16:
```suggestion
// proofCollector collects proof nodes' CIDs for the construction of a shares inclusion validation nmt.Proof.
```

distractedm1nd: does this have the same meaning? the original comment was a bit difficult to understand
probably would also still need to be line wrapped

distractedm1nd (approved):
great stuff!

Wondertan (approved):

renaynay (approved):

vgonkivs (approved):

distractedm1nd (approved):

PR #1341, opened 2022-11-10
tests(nodebuilder): sync and das over non-empty blocks

Wondertan:
Assert with swamp integration tests that Light and Full Node can sync and das over non-empty blocks

### DIFF
--- a/nodebuilder/tests/sync_test.go
+++ b/nodebuilder/tests/sync_test.go
@@ -3,6 +3,7 @@ package tests
 import (
 	"context"
 	"testing"
+	"time"
 
 	"github.com/libp2p/go-libp2p-core/host"
 	"github.com/libp2p/go-libp2p-core/peer"
@@ -15,7 +16,7 @@ import (
 )
 
 /*
-Test-Case: Sync a Light Node with a Bridge Node
+Test-Case: Sync a Light Node with a Bridge Node(includes DASing of non-empty blocks)
 Steps:
 1. Create a Bridge Node(BN)
 2. Start a BN
@@ -25,12 +26,19 @@ Steps:
 6. Check LN is synced to height 30
 */
 func TestSyncLightWithBridge(t *testing.T) {
-	sw := swamp.NewSwamp(t)
-
-	bridge := sw.NewBridgeNode()
-
 	ctx, cancel := context.WithTimeout(context.Background(), swamp.DefaultTestTimeout)
 	t.Cleanup(cancel)
+
+	const (
+		blocks = 20
+		bsize  = 16
+		btime  = time.Millisecond * 300
+	)
+
+	sw := swamp.NewSwamp(t, swamp.WithBlockTime(btime))
+	fillDn := sw.FillBlocks(ctx, bsize, blocks)
+
+	bridge := sw.NewBridgeNode()
 
 	sw.WaitTillHeight(ctx, 20)
 
@@ -55,7 +63,22 @@ func TestSyncLightWithBridge(t *testing.T) {
 	h, err = light.HeaderServ.GetByHeight(ctx, 30)
 	require.NoError(t, err)
 
+	err = light.ShareServ.SharesAvailable(ctx, h.DAH)
+	assert.NoError(t, err)
+
+	for {
+		s, err := light.DASer.SamplingStats(ctx)
+		assert.NoError(t, err)
+
+		// TODO(@Wondertan): should be subscription instead
+		time.Sleep(time.Millisecond * 10)
+		if s.CatchUpDone {
+			break
+		}
+	}
+
 	assert.EqualValues(t, h.Commit.BlockID.Hash, sw.GetCoreBlockHashByHeight(ctx, 30))
+	require.NoError(t, <-fillDn)
 }
 
 /*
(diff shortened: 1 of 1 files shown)


nodebuilder/tests/sync_test.go
```diff
@@ -55,7 +63,22 @@ func TestSyncLightWithBridge(t *testing.T) {
...
+
+	for {
+		s, err := light.DASer.SamplingStats(ctx)
+		assert.NoError(t, err)
+
+		// TODO(@Wondertan): should be subscription instead
+		time.Sleep(time.Millisecond * 10)
+		if s.CatchUpDone {
+			break
+		}
```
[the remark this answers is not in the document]
Wondertan: >  possibility for timeout

Do you mean to check for context? True.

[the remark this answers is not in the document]
Wondertan: > Side note: Implementation could be done by exposing catchUpDoneCh from DASer.

I would love to have it sooner than earlier, so I can rebase this test on top of that change. Do you mind quickly implementing this API on DASer?

nodebuilder/tests/sync_test.go
```diff
@@ -55,7 +63,22 @@ func TestSyncLightWithBridge(t *testing.T) {
 	h, err = light.HeaderServ.GetByHeight(ctx, 30)
 	require.NoError(t, err)
 
+	err = light.ShareServ.SharesAvailable(ctx, h.DAH)
+	assert.NoError(t, err)
+
+	for {
+		s, err := light.DASer.SamplingStats(ctx)
+		assert.NoError(t, err)
```
renaynay, line 71: require.NoError here

nodebuilder/tests/sync_test.go
```diff
@@ -159,6 +189,23 @@ func TestSyncFullWithBridge(t *testing.T) {
 	require.NoError(t, err)
 
 	assert.EqualValues(t, h.Commit.BlockID.Hash, sw.GetCoreBlockHashByHeight(ctx, 30))
+
+	err = full.ShareServ.SharesAvailable(ctx, h.DAH)
+	assert.NoError(t, err)
+
+	for {
+		s, err := full.DASer.SamplingStats(ctx)
+		assert.NoError(t, err)
```
renaynay, line 198: also require

renaynay (commented):
+1 to subscription to catchUpDone from daser

nodebuilder/tests/sync_test.go
Wondertan: Why?

nodebuilder/tests/sync_test.go
```diff
@@ -55,7 +63,22 @@ func TestSyncLightWithBridge(t *testing.T) {
 	h, err = light.HeaderServ.GetByHeight(ctx, 30)
 	require.NoError(t, err)
 
+	err = light.ShareServ.SharesAvailable(ctx, h.DAH)
+	assert.NoError(t, err)
+
+	for {
```
MSevey, line 69: As an alternative, this could use the new utils Retry method instead of a potentially infinite loop.

```go
err = test.Retry(100, 100*time.Millisecond, func() error {
   // Checking this first allows the duration between attempts to be the sleep
   if s.CatchUpDone {
      return nil
   }
   return fmt.Errorf("catch up not done; SharesAvailable error %v",light.ShareServ.SharesAvailable(ctx, h.DAH))
}
```

nodebuilder/tests/sync_test.go
```diff
@@ -130,13 +153,20 @@ Steps:
...
-	bridge := sw.NewBridgeNode()
-
 	ctx, cancel := context.WithTimeout(context.Background(), swamp.DefaultTestTimeout)
 	t.Cleanup(cancel)
 
+	const (
+		blocks = 20
+		bsize  = 16
+		btime  = time.Millisecond * 300
+	)
```
MSevey, line 163: nit: DRY, seems like it is worthwhile defining these in a common place if both tests are using them.

nodebuilder/tests/sync_test.go
```diff
@@ -159,6 +189,23 @@ func TestSyncFullWithBridge(t *testing.T) {
 	require.NoError(t, err)
 
 	assert.EqualValues(t, h.Commit.BlockID.Hash, sw.GetCoreBlockHashByHeight(ctx, 30))
+
+	err = full.ShareServ.SharesAvailable(ctx, h.DAH)
+	assert.NoError(t, err)
+
+	for {
```
MSevey, line 196: same suggestion here about the use of `Retry` instead of a potential infinite loop

nodebuilder/tests/sync_test.go
Wondertan: Thanks. It would be nice, for sure. Although, we should go an even cleaner way - subscribing for the done event.

Wondertan: As per [this](https://github.com/celestiaorg/celestia-node/pull/1341#discussion_r1019375142) tread

nodebuilder/tests/sync_test.go
Wondertan: Fixed

renaynay (approved):

vgonkivs (approved):

PR #1354, opened 2022-11-11
fix(eds/byzantine): trim `NMTWrapper`'s namespace during BEFP validation

evan-forbes:
## Overview
Fixes https://github.com/celestiaorg/celestia-node/issues/1355
Adds a test for a false positive for BEFPs. 

## Checklist

### DIFF
--- a/share/eds/byzantine/bad_encoding_test.go
+++ b/share/eds/byzantine/bad_encoding_test.go
@@ -7,14 +7,15 @@ import (
 	"sort"
 	"testing"
 
+	mdutils "github.com/ipfs/go-merkledag/test"
+	"github.com/stretchr/testify/require"
+	core "github.com/tendermint/tendermint/types"
+
 	"github.com/celestiaorg/celestia-app/pkg/da"
 	"github.com/celestiaorg/celestia-node/header"
 	"github.com/celestiaorg/celestia-node/share"
 	"github.com/celestiaorg/celestia-node/share/ipld"
 	"github.com/celestiaorg/rsmt2d"
-	mdutils "github.com/ipfs/go-merkledag/test"
-	"github.com/stretchr/testify/require"
-	core "github.com/tendermint/tendermint/types"
 )
 
 func TestFalsePositiveBadEncodingFraudProof(t *testing.T) {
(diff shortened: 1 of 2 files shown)


share/eds/byzantine/bad_encoding_test.go
```diff
@@ -0,0 +1,92 @@
...
+	sortByteArrays(data)
+	for i := uint32(0); i < total; i++ {
+		d := make([]byte, leafSize)
+
+		_, _ = rand.Read(d)
+		data[i] = append(data[i], d...)
+	}
+
+	return data
+}
```
evan-forbes, line 88: ideally we shouldn't copy the code here, but I wasn't sure what else todo. pls advise

share/eds/byzantine/share_proof.go
```diff
@@ -75,21 +75,23 @@ func GetProofsForShares(
...
-			proofs[index] = NewShareWithProof(index, s.RawData(), proof)
+	for index, sh := range shares {
+		if sh == nil {
+			continue
+		}
+
+		proof := make([]cid.Cid, 0)
+		// TODO(@vgonkivs): Combine GetLeafData and GetProof in one function as the are traversing the same
+		// tree. Add options that will control what data will be fetched.
+		s, err := share.GetShare(ctx, bGetter, root, index, len(shares))
```
Wondertan, line 86: This is where the main fix is placed. Mainly instead of trimming the namespace on the validator side, we should not send it at all and trim it on the sender(FN) side. This is what `GetShare` does. It calls the substituted `ipld.GetLeaf` and trims the namespace. `share` pkg has a bunch of proxy getters that hide leaky namespace by subslicing it. We should have done this from the beginning but somehow missed it.

@evan-forbes, @renaynay, @walldiss

share/eds/byzantine/bad_encoding_test.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+func TestFalsePositiveBadEncodingFraudProof(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	bServ := mdutils.Bserv()
+
+	squareSize := 8
+	shares := share.RandShares(t, squareSize*squareSize)
```
Wondertan, line 25: @evan-forbes, we have a utiltiy for this, so I just used it. Not sure what was the reasoning behind making share size 504 btw

evan-forbes: nice! and 504 = 512 - namespaceIDSize

share/eds/byzantine/share_proof.go
evan-forbes: makes sense!

Wondertan: @evan-forbes, I completely forgot that we need that namespace during share inclusion verification, and just using `share.GetShare` fixes the FalsePostive test, but breaks others. I tried to dynamically prepend a NID, similarly to what NMTWrapper does, but it increased hairiness, looked complex and required more params for `share.Validate`, so I reverted it to your original solution with the goal of simplifying code that it is already hard to understand

distractedm1nd (approved):

walldiss (dismissed):

Wondertan (approved):

renaynay (approved):

PR #1359, opened 2022-11-14
feat(share): Periodic GC over EDSStore

### LINKED ISSUE #1116: share: Periodic GC over `DAGStore` inside `EDSStore`

distractedm1nd:
# PULL REQUEST

## Overview
This PR implements a goroutine that runs during the lifetime of the `EDSStore`, which calls `dgstr.GC` periodically. This garbage collection cleans up shards from transient storage that are available but inactive or errored.

The PR is based on #1232 and closes #1116.
## Checklist

### DIFF
--- a/nodebuilder/p2p/p2p.go
+++ b/nodebuilder/p2p/p2p.go
@@ -67,7 +67,8 @@ type Module interface {
 	// BandwidthForPeer returns a Stats struct with bandwidth metrics associated with the given peer.ID.
 	// The metrics returned include all traffic sent / received for the peer, regardless of protocol.
 	BandwidthForPeer(id peer.ID) metrics.Stats
-	// BandwidthForProtocol returns a Stats struct with bandwidth metrics associated with the given protocol.ID.
+	// BandwidthForProtocol returns a Stats struct with bandwidth metrics associated with the given
+	// protocol.ID.
 	BandwidthForProtocol(proto protocol.ID) metrics.Stats
 
 	// ResourceState returns the state of the resource manager.
--- a/share/eds/store.go
+++ b/share/eds/store.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"time"
 
 	"github.com/filecoin-project/dagstore"
 	"github.com/filecoin-project/dagstore/index"
@@ -17,6 +18,10 @@ import (
 	"github.com/celestiaorg/rsmt2d"
 )
 
+// TODO(@distractedm1nd): this probably shouldn't be a var. it could be configured next to the blockstore's path and bs
+// cache size.
+var gcInterval = time.Hour
+
 const (
 	blocksPath     = "/blocks/"
 	indexPath      = "/index/"
@@ -28,6 +33,8 @@ const (
 // Blockstore interface implementation to achieve access. The main use-case is randomized sampling
 // over the whole chain of EDS block data and getting data by namespace.
 type Store struct {
+	cancel context.CancelFunc
+
 	dgstr  *dagstore.DAGStore
 	mounts *mount.Registry
 
@@ -78,15 +85,37 @@ func NewStore(basepath string, ds datastore.Batching) (*Store, error) {
 	}, nil
 }
 
-// Start starts the underlying DAGStore.
+// Start starts the underlying DAGStore and gc routine.
 func (s *Store) Start(context.Context) error {
-	ctx := context.Background()
+	ctx, cancel := context.WithCancel(context.Background())
+	s.cancel = cancel
+
+	go s.gc(ctx)
 	return s.dgstr.Start(ctx)
 }
 
 // Stop stops the underlying DAGStore.
 func (s *Store) Stop(context.Context) error {
+	s.cancel()
 	return s.dgstr.Close()
+}
+
+// gc periodically removes all inactive or errored shards.
+func (s *Store) gc(ctx context.Context) {
+	ticker := time.NewTicker(gcInterval)
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			_, err := s.dgstr.GC(ctx)
+			if err != nil {
+				log.Errorf("garbage collecting dagstore: %v", err)
+				return
+			}
+		}
+
+	}
 }
 
 // Put stores the given data square with DataRoot's hash as a key.
--- a/share/eds/store.go
+++ b/share/eds/store.go
@@ -41,7 +41,8 @@ type Store struct {
 	topIdx index.Inverted
 	carIdx index.FullIndexRepo
 
-	basepath     string
+	basepath string
+	// lastGCResult is only stored on the store for testing purposes.
 	lastGCResult *dagstore.GCResult
 }
 
@@ -90,8 +91,7 @@ func NewStore(basepath string, ds datastore.Batching) (*Store, error) {
 	}, nil
 }
 
-// Start starts the underlying DAGStore and gc routine.
-func (s *Store) Start(context.Context) error {
+func (s *Store) Start(ctx context.Context) error {
 	ctx, cancel := context.WithCancel(context.Background())
 	s.cancel = cancel
 
--- a/share/eds/store.go
+++ b/share/eds/store.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"sync/atomic"
 	"time"
 
 	"github.com/filecoin-project/dagstore"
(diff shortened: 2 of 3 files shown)


share/eds/store_test.go
```diff
@@ -0,0 +1,197 @@
...
+func TestEDSStore_PutRegistersShard(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	t.Cleanup(cancel)
+
+	edsStore, err := newEDSStore(t)
+	require.NoError(t, err)
+	err = edsStore.Start(ctx)
+	require.NoError(t, err)
+
+	eds, dah := randomEDS(t)
```
MSevey, line 32: Copying comment here about DRY'ing up this test code.

share/eds/store.go
```diff
@@ -0,0 +1,245 @@
...
+// cache size.
+var gcInterval = time.Hour
+
+const (
+	blocksPath     = "/blocks/"
+	indexPath      = "/index/"
+	transientsPath = "/transients/"
+)
+
+type EDSStore struct { //nolint:revive
```
MSevey, line 32: why no lint?

Exports should have comments.

share/eds/store.go
```diff
@@ -71,23 +80,50 @@ func NewStore(basepath string, ds datastore.Batching) (*Store, error) {
...
+
+	go s.gc(runCtx)
 	return s.dgstr.Start(ctx)
 }
 
-// Stop stops the underlying DAGStore.
 func (s *Store) Stop() error {
+	s.cancel()
 	return s.dgstr.Close()
 }
```
walldiss, line 106: Please wait for GC to exit before returning

share/eds/store_test.go
```diff
@@ -156,6 +156,33 @@ func TestEDSStore_Has(t *testing.T) {
...
+
+	err = edsStore.Put(ctx, dah, eds)
+	assert.NoError(t, err)
+
+	// doesn't exist yet
+	assert.NotContains(t, edsStore.lastGCResult.Shards, shard.KeyFromString(dah.String()))
+
+	time.Sleep(time.Second)
+	assert.Contains(t, edsStore.lastGCResult.Shards, shard.KeyFromString(dah.String()))
```
walldiss, line 180: Could this potentially be flaky in CI? If there is no signals available from Maybe we can run for loop with retry for limited amount time

nodebuilder/p2p/p2p.go
```diff
@@ -64,7 +64,8 @@ type Module interface {
 	// BandwidthForPeer returns a Stats struct with bandwidth metrics associated with the given peer.ID.
 	// The metrics returned include all traffic sent / received for the peer, regardless of protocol.
 	BandwidthForPeer(id peer.ID) metrics.Stats
-	// BandwidthForProtocol returns a Stats struct with bandwidth metrics associated with the given protocol.ID.
+	// BandwidthForProtocol returns a Stats struct with bandwidth metrics associated with the given
+	// protocol.ID.
```
Wondertan, line 68: @renaynay, do you have cfmt installed? Please enable it locally

share/eds/store.go
```diff
@@ -17,6 +18,10 @@ import (
 	"github.com/celestiaorg/rsmt2d"
 )
 
+// TODO(@distractedm1nd): this probably shouldn't be a var. it could be configured next to the
+// blockstore's path and bs cache size.
+var gcInterval = time.Hour
+
```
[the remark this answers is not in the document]
Wondertan: obviously not in this PR

share/eds/store.go
```diff
@@ -71,23 +80,50 @@ func NewStore(basepath string, ds datastore.Batching) (*Store, error) {
...
 	}, nil
 }
 
-// Start starts the underlying DAGStore.
 func (s *Store) Start(ctx context.Context) error {
+	runCtx, cancel := context.WithCancel(context.Background())
+	s.ctx = runCtx
+	s.cancel = cancel
+
+	go s.gc(runCtx)
```
Wondertan, line 99:
```suggestion
	s.ctx , s.cancel = context.WithCancel(context.Background())
	go s.gc(s.ctx)
```

share/eds/store.go
Wondertan: Why do you think we need this in this case?

share/eds/store.go
```diff
@@ -28,13 +33,17 @@ const (
...
+
 	dgstr  *dagstore.DAGStore
 	mounts *mount.Registry
 
 	topIdx index.Inverted
 	carIdx index.FullIndexRepo
 
-	basepath string
+	basepath     string
+	lastGCResult *dagstore.GCResult
```
Wondertan, line 46: Ok, now I see the what was you question about testing with logs

Wondertan: Add a comment that is used in tests only

Wondertan (dismissed):

share/eds/store_test.go
renaynay: Agree, use retry please

distractedm1nd: done

share/eds/store.go

share/eds/store.go
distractedm1nd: done

distractedm1nd:
The GC test is racy! will try to fix today

share/eds/store.go
Wondertan: Yeah, it's fine because it also survives restarts

share/eds/store.go
```diff
@@ -71,23 +81,48 @@ func NewStore(basepath string, ds datastore.Batching) (*Store, error) {
...
+			res, err := s.dgstr.GC(ctx)
+			if err != nil {
+				log.Errorf("garbage collecting dagstore: %v", err)
+				return
+			}
+			s.lastGCResult = res
+		}
+
+	}
+}
```
vgonkivs, line 124: I wondering, should we create a GC package to be able to re-use it in different places, as the main waiting logic is the same?

Wondertan: We could extract a pkg like `periodically.Do` or similar into libs or utils. Care for a refactoring issue?

vgonkivs: The code here is similar to backOffConnetor.GC and also peerTracker will have the same logic. So, yeah, I think we should extract it somewhere

share/eds/store.go
```diff
@@ -71,23 +81,48 @@ func NewStore(basepath string, ds datastore.Batching) (*Store, error) {
...
-// Start starts the underlying DAGStore.
 func (s *Store) Start(ctx context.Context) error {
+	s.ctx, s.cancel = context.WithCancel(context.Background())
+	go s.gc(s.ctx)
+
 	return s.dgstr.Start(ctx)
 }
 
-// Stop stops the underlying DAGStore.
 func (s *Store) Stop() error {
```
walldiss, line 102: We usually pass context to  `Stop()`, let's do it here too.

```suggestion
func (s *Store) Stop(context.Context) error {
```

share/eds/store.go
```diff
@@ -71,23 +81,48 @@ func NewStore(basepath string, ds datastore.Batching) (*Store, error) {
...
+		dgstr:  dagStore,
+		topIdx: invertedRepo,
+		carIdx: fsRepo,
+		mounts: r,
 	}, nil
 }
 
-// Start starts the underlying DAGStore.
 func (s *Store) Start(ctx context.Context) error {
+	s.ctx, s.cancel = context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
distractedm1nd: on Stop the context is cancelled, which stops the gc routine

share/eds/store.go
```diff
@@ -70,24 +79,50 @@ func NewStore(basepath string, ds datastore.Batching) (*Store, error) {
...
+		mounts:     r,
 	}, nil
 }
 
-// Start starts the underlying DAGStore.
 func (s *Store) Start(ctx context.Context) error {
+	s.ctx, s.cancel = context.WithCancel(context.Background())
+	go s.gc(s.ctx)
+
 	return s.dgstr.Start(ctx)
```
renaynay, line 95: I know you're taking care of this in another PR but just make sure you pass the long-lived ctx

Wondertan (dismissed):
LGTM

share/eds/store.go
Wondertan: We usually do it because FX requires us, but with its recent features, it no more does this, so we can stop passing context where  unnecessary

share/eds/store.go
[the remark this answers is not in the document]
walldiss: You can keep cancel only then. No need to store ctx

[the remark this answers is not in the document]
renaynay: tru [non-blocking]

renaynay (approved):

share/eds/store.go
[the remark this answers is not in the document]
Wondertan: Doesn't hurt to keep it for once we need actually need it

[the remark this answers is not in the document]
Wondertan: But it can also be seen as don't introduce what is not immediately required, still nice to follow our pattern

share/eds/store.go
distractedm1nd: I don't like abstracting this sort of thing, it usually convolutes. especially now that im using an atomic pointer

share/eds/store.go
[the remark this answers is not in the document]
distractedm1nd: ohhh now I see what you meant, thanks for the clarification, done

renaynay (approved):

Wondertan (approved):

walldiss (approved):

Wondertan (approved):

PR #1364, opened 2022-11-14
fix(header): Extend `ValidateBasic` on `ExtendedHeader` to include data root check

### LINKED ISSUE #1362: header:  `ExtendedHeader`'s `ValidateBasic` must check if the `DAH.Hash()` matches `DataHash` in `RawHeader`

renaynay:
Resolves #1362

### DIFF
--- a/header/header_test.go
+++ b/header/header_test.go
@@ -8,14 +8,17 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	"github.com/celestiaorg/celestia-app/pkg/da"
+
 	"github.com/celestiaorg/celestia-node/core"
+	"github.com/celestiaorg/celestia-node/share"
 )
 
 func TestMakeExtendedHeaderForEmptyBlock(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
 	t.Cleanup(cancel)
 
-	_, client := core.StartTestClient(ctx, t)
+	_, client := core.StartTestCoreWithApp(t)
 	fetcher := core.NewBlockFetcher(client)
 
 	store := mdutils.Bserv()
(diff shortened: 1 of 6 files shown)


vgonkivs (dismissed):

header/testing.go
```diff
@@ -171,15 +177,15 @@ func RandExtendedHeader(t *testing.T) *ExtendedHeader {
 }
 
 // RandRawHeader provides a RawHeader fixture.
-func RandRawHeader(t *testing.T) *RawHeader {
+func RandRawHeader(t *testing.T, dataHash []byte) *RawHeader {
```
Wondertan, line 180: nit: `RandRawHeader` is meant to really be rand, so it can be used in tests that don't care about data hashes or other fields. This change makes `RandRawHeader` more oriented toward data testing, thus making it not credibly neutral :joy:, so I would just set DataHash field on the returned RawHeader instead.

Wondertan: In this PR in `TestSuites`'s `GenRawHeader` we pass the dataHash and then assign its field another time btw

header/header_test.go
```diff
@@ -36,3 +39,33 @@ func TestMakeExtendedHeaderForEmptyBlock(t *testing.T) {
...
+	sub, err := fetcher.SubscribeNewBlockEvent(ctx)
+	require.NoError(t, err)
+	<-sub
+
+	height := int64(1)
+	b, err := fetcher.GetBlock(ctx, &height)
+	require.NoError(t, err)
+
+	comm, val, err := fetcher.GetBlockInfo(ctx, &height)
+	require.NoError(t, err)
```
Wondertan, line 61: simplification: If the goal is to unit test the new check, then all this setup is not necessary. We can use TestSute which is simpler and generates a valid header and we can similarly augment it with an incorrect hash to see that error happens.

renaynay: done.

header/testing.go
renaynay: ok

Wondertan (approved):

PR #1366, opened 2022-11-14
fix(das): Correct `WithMetrics` fx Invocation

derrandz:
## Overview

This PR fixes the `das.WithMetrics` fx invokation which enables metrics collection in the DASer module.

## Impact

Negligeable

## Owners

@derrandz

### DIFF
--- a/nodebuilder/das/opts.go
+++ b/nodebuilder/das/opts.go
@@ -1,13 +1,11 @@
 package das
 
 import (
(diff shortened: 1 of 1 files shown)


Wondertan (dismissed):
Good catch

walldiss (dismissed):
Good catch

Wondertan (approved):

PR #1368, opened 2022-11-14
fix(coreaccessor): pfd submission nil check

distractedm1nd:
# PULL REQUEST

## Overview
PFD submissions collect metrics. I swapped the arguments to an && condition that ensures a struct is only accessed if the err != nil.

## Checklist

### DIFF
--- a/state/core_access.go
+++ b/state/core_access.go
@@ -153,7 +153,7 @@ func (ca *CoreAccessor) SubmitPayForData(
 ) (*TxResponse, error) {
 	response, err := payment.SubmitPayForData(ctx, ca.signer, ca.coreConn, nID, data, gasLim)
 	// metrics should only be counted on a successful PFD tx
-	if response.Code == 0 && err == nil {
+	if response != nil && response.Code == 0 && err == nil {
 		ca.lastPayForData = time.Now().UnixMilli()
 		ca.payForDataCount++
 	}
--- a/state/core_access.go
+++ b/state/core_access.go
@@ -153,7 +153,7 @@ func (ca *CoreAccessor) SubmitPayForData(
 ) (*TxResponse, error) {
 	response, err := payment.SubmitPayForData(ctx, ca.signer, ca.coreConn, nID, data, gasLim)
 	// metrics should only be counted on a successful PFD tx
-	if response != nil && response.Code == 0 && err == nil {
+	if err == nil && response.Code == 0 {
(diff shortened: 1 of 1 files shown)


state/core_access.go
```diff
@@ -153,7 +153,7 @@ func (ca *CoreAccessor) SubmitPayForData(
 ) (*TxResponse, error) {
 	response, err := payment.SubmitPayForData(ctx, ca.signer, ca.coreConn, nID, data, gasLim)
 	// metrics should only be counted on a successful PFD tx
-	if response.Code == 0 && err == nil {
+	if response != nil && response.Code == 0 && err == nil {
```
Wondertan, line 156: Bruh
This is the best DevX I've ever seen 😂

Wondertan (approved):

state/core_access.go
distractedm1nd: its so bad... We should probably hunt down where they do this and open a PR, no?

distractedm1nd: Wait.. we're just stupid. Let me fix

derrandz: 😂

state/core_access.go
```diff
@@ -153,7 +153,7 @@ func (ca *CoreAccessor) SubmitPayForData(
 ) (*TxResponse, error) {
 	response, err := payment.SubmitPayForData(ctx, ca.signer, ca.coreConn, nID, data, gasLim)
 	// metrics should only be counted on a successful PFD tx
-	if response.Code == 0 && err == nil {
+	if err == nil && response.Code == 0 {
```
derrandz, line 156:
```suggestion
	if err == nil && response != nil && response.Code == 0 {
```

derrandz: This is more semantic as it reads:

1. if there is no error
2. and we have an actual response
3. and the response Code is 0

derrandz (changes requested):

state/core_access.go
Wondertan: If I would design the API it would be just an error from which you can get an error code if needed. 3 conditions after doing a call is 😬

renaynay (approved):

Bidon15 (approved):
no panics on testground tests 💪

derrandz (approved):
Approving since the original assumption turned out to be false and this correction as is is sufficient.

Issue #1372, opened 2022-11-15
Bridge node's config contains DASer section


Bidon15:
### Celestia Node version

7787c6bcb1b32a9f2016bb504215bc18cd59a873

### OS

macOS 13.0.1

### Install tools

`make go-install`

### Others

_No response_

### Steps to reproduce it

1. Install v0.5.0-rc4 version
2. Init the bridge node 
3. Check the config.toml

### Expected result

No DASer section for bridge node

### Actual result

Daser section is on for the bridge node

### Relevant log output

```Shell
Error: moddas misconfiguration: das: invalid option: value SamplingRange cannot be negative or 

cat /Users/bidon4/.celestia-bridge-arabica-2/config.toml                                                                                                                                                                                                                                                 bidon4@Bidon15
[Core]
  IP = "0.0.0.0"
  RPCPort = "0"
  GRPCPort = "0"
​
[State]
  KeyringAccName = ""
​
[P2P]
  ListenAddresses = ["/ip4/0.0.0.0/udp/2121/quic", "/ip6/::/udp/2121/quic", "/ip4/0.0.0.0/tcp/2121", "/ip6/::/tcp/2121"]
  AnnounceAddresses = []
  NoAnnounceAddresses = ["/ip4/0.0.0.0/udp/2121/quic", "/ip4/127.0.0.1/tcp/2121/quic", "/ip6/::/udp/2121/quic", "/ip4/0.0.0.0/tcp/2121", "/ip4/127.0.0.1/tcp/2121", "/ip6/::/tcp/2121"]
  Bootstrapper = false
  MutualPeers = []
  PeerExchange = false
  RoutingTableRefreshPeriod = "1m0s"
  [P2P.ConnManager]
    Low = 50
    High = 100
    GracePeriod = "1m0s"
​
[RPC]
  Address = "0.0.0.0"
  Port = "26658"
​
[Gateway]
  Address = "0.0.0.0"
  Port = "26659"
  Enabled = false
​
[Share]
  PeersLimit = 3
  DiscoveryInterval = "30s"
  AdvertiseInterval = "30s"
​
[Header]
  TrustedHash = ""
  TrustedPeers = []
​
[DASer]
  SamplingRange = 100
  ConcurrencyLimit = 16
  BackgroundStoreInterval = "10m0s"
  PriorityQueueSize = 64
  SampleFrom = 1
```


### Notes

_No response_

renaynay:
[...]

PR #1373, opened 2022-11-15
fix(nodebuilder/das): Skip config validation for bridge node

renaynay:
Related to #1372

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ replace github.com/ipfs/go-verifcid => github.com/celestiaorg/go-verifcid v0.0.1
 
 require (
 	cosmossdk.io/math v1.0.0-beta.3
-	github.com/BurntSushi/toml v1.2.0
+	github.com/BurntSushi/toml v1.2.1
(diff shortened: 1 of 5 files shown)


distractedm1nd (approved):

Wondertan:
Hmm, it does not seem to fix the issue attached

renaynay:
@Wondertan yes it will still emit the DASer config but it won't validate it since it does not need it.

Wondertan:
It does not fix the attached issue, that's the point.

distractedm1nd:
I updated the burntsushi toml dep which now has a working omitempty, so an empty daser config is now not included in the TOML file. I also refactored the readwrite config test to run all node types.

Wondertan (approved):
Did you do a sanity check by initializing a new Bridge node?

distractedm1nd:
@Wondertan yes, it works - that part of the config is not included in the toml then. And it has no problems using it on start

distractedm1nd (approved):
Excellent hack w/ the toml tag, great work

PR #1378, opened 2022-11-17
fix: allowing hostnames in gateway/rpc addr field

### LINKED ISSUE #1374: Error is thrown when providing `--gateway.addr` flag
### Celestia Node version

039facd94104c741f8fa995e80f1321776ae4ad3

### OS

macOS 13.0.1

### Install tools

make go-install

### Others

_No response_

### Steps to reproduce it

1. Init the light node
2. Execute start with gateway flags

```shell
celestia light start --core.ip https://limani.celestia-devops.dev --core.grpc.port 9090 --gateway --gateway.addr localhost --gateway.port 26659           
```

### Expected result

Light Node starts with no error 

### Actual result

Light Node fails with error invalid address format

Note: if tried also with `http://localhost` and other variants -> same outcome with errors

### Relevant log output

```Shell
Error: gateway: invalid listen address format: localhost
```

### Notes

As a mitigation, we can just start with this style to get back the REST API
```shell
celestia light start --core.ip https://limani.celestia-devops.dev --core.grpc.port 9090 --gateway
```

distractedm1nd:
## Overview
Closes #1374 

IP/Hostname sanitization logic has been moved to `utils`. We now accept hostnames as valid addresses for the rpc and gateway, and the core ip parsing logic works the same way as the others now.

## Checklist

### DIFF
--- a/libs/utils/address_test.go
+++ b/libs/utils/address_test.go
@@ -1,0 +1,49 @@
+package utils
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestSanitizeAddr(t *testing.T) {
+	var tests = []struct {
+		addr string
+		want string
+	}{
+		// Testcase: trims protocol prefix
+		{addr: "http://celestia.org", want: "celestia.org"},
+		// Testcase: trims protocol prefix, and trims port and trailing slash suffix
+		{addr: "tcp://192.168.42.42:5050/", want: "192.168.42.42"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.addr, func(t *testing.T) {
+			got, err := SanitizeAddr(tt.addr)
+			require.NoError(t, err)
+			require.Equal(t, tt.want, got)
+		})
+	}
+}
+
+func TestValidateAddr(t *testing.T) {
+	var tests = []struct {
+		addr string
+		want string
+	}{
+		// Testcase: ip is valid
+		{addr: "192.168.42.42:5050", want: "192.168.42.42"},
+		// Testcase: hostname is valid
+		{addr: "https://celestia.org", want: "celestia.org"},
+		// Testcase: resolves localhost
+		{addr: "http://localhost:8080/", want: "localhost"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.addr, func(t *testing.T) {
+			got, err := ValidateAddr(tt.addr)
+			require.NoError(t, err)
+			require.Equal(t, tt.want, got)
+		})
+	}
+}
--- a/nodebuilder/gateway/config.go
+++ b/nodebuilder/gateway/config.go
@@ -23,7 +23,11 @@ func DefaultConfig() Config {
 
 func (cfg *Config) Validate() error {
 	if ip := net.ParseIP(cfg.Address); ip == nil {
-		return fmt.Errorf("gateway: invalid listen address format: %s", cfg.Address)
+		// ip was not a valid IP, so try to see if given string is a valid hostname
+		_, err := net.LookupHost(cfg.Address)
+		if err != nil {
+			return fmt.Errorf("gateway: invalid listen address format: %s", cfg.Address)
+		}
 	}
 	_, err := strconv.Atoi(cfg.Port)
 	if err != nil {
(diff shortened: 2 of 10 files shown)


Wondertan (dismissed):
Great that you extracted those in utils. Bad that we touched on unrelated comments. Can you remove those and extract into a separate PR, so that the squashed commit does not touch unrelated comments?

distractedm1nd:
@Wondertan ya i can remove them. I already opened an issue that will fix the problem of make fmt fixing something the linter doesn't catch

renaynay (dismissed):
beautiful thank u

nodebuilder/gateway/config.go
```diff
@@ -22,10 +23,13 @@ func DefaultConfig() Config {
...
-	if ip := net.ParseIP(cfg.Address); ip == nil {
-		return fmt.Errorf("gateway: invalid listen address format: %s", cfg.Address)
+	sanitizedAddress, err := utils.ValidateAddr(cfg.Address)
+	if err != nil {
+		return fmt.Errorf("gateway: invalid address: %w", err)
 	}
-	_, err := strconv.Atoi(cfg.Port)
+	cfg.Address = sanitizedAddress
+
+	_, err = strconv.Atoi(cfg.Port)
```
MSevey, line 32: nit: is there a reason this wouldn't be part of `ValidateAddr`?

nodebuilder/rpc/config.go
```diff
@@ -20,10 +21,13 @@ func DefaultConfig() Config {
...
-	if ip := net.ParseIP(cfg.Address); ip == nil {
-		return fmt.Errorf("service/rpc: invalid listen address format: %s", cfg.Address)
+	sanitizedAddress, err := utils.ValidateAddr(cfg.Address)
+	if err != nil {
+		return fmt.Errorf("service/rpc: invalid address: %w", err)
 	}
-	_, err := strconv.Atoi(cfg.Port)
+	cfg.Address = sanitizedAddress
+
+	_, err = strconv.Atoi(cfg.Port)
```
MSevey, line 30: feels like it should be, so we don't have to remember to do the extra validation step in multiple places.

libs/utils/address_test.go
```diff
@@ -0,0 +1,49 @@
...
+		})
+	}
+}
+
+func TestValidateAddr(t *testing.T) {
+	var tests = []struct {
+		addr string
+		want string
+	}{
+		// Testcase: ip is valid
```
MSevey, line 34: [non-blocking]
Do we want to support ipv6? Or is that something we've talked about?

libs/utils/address_test.go
```diff
@@ -0,0 +1,49 @@
+package utils
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestSanitizeAddr(t *testing.T) {
```
MSevey, line 9: Additional Test Cases needed

- https
- no leading protocol
- localhost

libs/utils/address_test.go
```diff
@@ -0,0 +1,49 @@
...
+	for _, tt := range tests {
+		t.Run(tt.addr, func(t *testing.T) {
+			got, err := SanitizeAddr(tt.addr)
+			require.NoError(t, err)
+			require.Equal(t, tt.want, got)
+		})
+	}
+}
+
+func TestValidateAddr(t *testing.T) {
```
MSevey, line 29: We should add tcp test case.

also could refactor the `tests` since they should be the same for both.

MSevey (commented):
Some minor comments, fine as follow ups.

renaynay (approved):

nodebuilder/gateway/config.go
distractedm1nd: Ref #1385

libs/utils/address_test.go
distractedm1nd: yeah we can add a testcase for it

libs/utils/address_test.go
distractedm1nd: Ref #1385

libs/utils/address_test.go
distractedm1nd: Ref #1385

PR #1380, opened 2022-11-17
feat(das): make WaitCatchUp public

walldiss:
## Overview

Exposes WaitCatchUp from DASer to subscribe to catchup event from external caller. Requested for https://github.com/celestiaorg/celestia-node/pull/1341

## Checklist
- [x] New and updated code has appropriate documentation
- [x] New and updated code has new and/or updated testing
- [x] Required CI checks are passing

PR #1383, opened 2022-11-18
feat(module/das): add WaitCatchUp method to the module

Wondertan:
#1380 made a blocking helper to wait until catching up is done. To use it for tests and further, we have to expose it on the Module either.


walldiss (dismissed):

vgonkivs (approved):

Wondertan:
Had to fix one test and move it closer to the DAS module. Fixed now

walldiss (approved):

vgonkivs (approved):

PR #1388, opened 2022-11-21
improvement(header/store): make all global variables configurable

vgonkivs:
Related to [#709](https://github.com/celestiaorg/celestia-node/issues/709)


nodebuilder/header/store/config.go
```diff
@@ -0,0 +1,25 @@
+package store
+
+import (
+	"fmt"
+
+	"github.com/celestiaorg/celestia-node/header/store"
+)
+
+// Config contains configuration parameters for the store
+type Config store.Parameters
```
Wondertan, line 10: It should be part of the global header config, not separate. Similarly, we should pass only one config to the ConstructModuler

nodebuilder/header/store/config.go
```diff
@@ -0,0 +1,25 @@
...
+func DefaultConfig() Config {
+	return Config(store.DefaultParameters())
+}
+
+// Validate performs basic validation of the config.
+// Upon encountering an invalid value, Validate returns an error of type: ErrMisConfig
+func (cfg *Config) Validate() error {
+	err := (*store.Parameters)(cfg).Validate()
+	if err != nil {
+		return fmt.Errorf("moddas misconfiguration: %w", err)
```
Wondertan, line 21:
```suggestion
		return fmt.Errorf("module/header: misconfiguration: %w", err)
```

header/store/options.go
```diff
@@ -0,0 +1,67 @@
+package store
+
+import "errors"
+
+// Option is the functional option that is applied to the store instance
+// to configure store parameters.
+type Option func(*store)
```
Wondertan, line 7: Pls make the option apply to the Parameters on not the store

Wondertan (changes requested):

Wondertan:
Think of the module/header.Config{ store.Parameters, syncer.Parameters, exchange.Parameters}

header/store/options.go
vgonkivs: ooops. 👌

header/store/options.go
```diff
@@ -0,0 +1,67 @@
...
+
+import "errors"
+
+// Option is the functional option that is applied to the store instance
+// to configure store parameters.
+type Option func(*Parameters)
+
+// Parameters is the set of parameters that must be configured for the store.
+type Parameters struct {
+	// StoreCacheSize defines the amount of max entries allowed in the Header Store cache.
```
distractedm1nd, line 11:
```suggestion
	// StoreCacheSize defines the maximum amount of entries in the Header Store cache.
```

header/store/options.go
```diff
@@ -0,0 +1,67 @@
...
+	StoreCacheSize int
+
+	// IndexCacheSize defines the amount of max entries allowed in the Height to Hash index
+	// cache.
```
distractedm1nd, line 15:
```suggestion
	// IndexCacheSize defines the maximum amount of entries in the Height to Hash index cache.
```

header/store/options.go
```diff
@@ -0,0 +1,67 @@
...
+	IndexCacheSize int
+
+	// WriteBatchSize defines the size of the batched header write.
+	// Headers are written in batches not to thrash the underlying Datastore with writes.
+	WriteBatchSize int
+}
+
+// DefaultParameters returns default params to configure store.
```
distractedm1nd, line 23:
```suggestion
// DefaultParameters returns the default params to configure the store.
```

header/store/options.go
```diff
@@ -0,0 +1,67 @@
...
+		return errors.New("invalid store cache size")
+	}
+	if p.IndexCacheSize <= 0 {
+		return errors.New("invalid indexer cache size")
+	}
+	if p.WriteBatchSize <= 0 {
+		return errors.New("invalid batch size")
+	}
+	return nil
+}
```
distractedm1nd, line 43: Note to myself: Create an issue to unit test all parameter and config validations

header/store/options.go
```diff
@@ -0,0 +1,67 @@
...
+
+// WithDefaultStoreCacheSize is a functional option that allows to configure
+// `StoreCacheSize` parameter.
```
distractedm1nd, line 46:
```suggestion
// WithDefaultStoreCacheSize is a functional option that configures the
// `StoreCacheSize` parameter.
```

distractedm1nd: and same with the other options

header/store/store.go
```diff
@@ -63,46 +50,59 @@ type store struct {
...
-		ds:          ds,
-		cache:       cache,
-		heightIndex: index,
-		heightSub:   newHeightSub(),
-		writes:      make(chan []*header.ExtendedHeader, 16),
-		writesDn:    make(chan struct{}),
-		pending:     newBatch(DefaultWriteBatchSize),
-	}, nil
+	return store, nil
 }
```
distractedm1nd, line 106: couldnt leave as suggestion so pasting here:

```golang
func newStore(ds datastore.Batching, opts ...Option) (*store, error) {
	params := DefaultParameters()
	for _, opt := range opts {
		opt(params)
	}

	cache, err := lru.NewARC(params.StoreCacheSize)
	if err != nil {
		return nil, fmt.Errorf("failed to create index cache: %w", err)
	}

	wrappedStore := namespace.Wrap(ds, storePrefix)
	index, err := newHeightIndexer(wrappedStore, params.IndexCacheSize)
	if err != nil {
		return nil, fmt.Errorf("failed to create height indexer: %w", err)
	}

	return &store{
		params:      params,
		ds:          wrappedStore,
		heightSub:   newHeightSub(),
		writes:      make(chan []*header.ExtendedHeader, 16),
		writesDn:    make(chan struct{}),
		cache:       cache,
		heightIndex: index,
		pending:     newBatch(params.WriteBatchSize),
	}, nil
}
```

distractedm1nd (changes requested):
love this pattern, I mostly only have doc nits

header/store/store.go
vgonkivs: yes, looks cleaner. thanks 🙏

header/store/options.go
distractedm1nd: #1391

distractedm1nd (approved):
tremendous

Wondertan (approved):
Ставлю лайкосик. Дальше другие компоненты

nodebuilder/header/module.go
```diff
@@ -26,6 +26,15 @@ func ConstructModule(tp node.Type, cfg *Config) fx.Option {
...
 		fx.Supply(modp2p.BlockTime),
+		fx.Provide(
+			func(cfg *Config) []store.Option {
+				return []store.Option{
+					store.WithDefaultStoreCacheSize(cfg.Store.StoreCacheSize),
+					store.WithDefaultIndexCacheSize(cfg.Store.IndexCacheSize),
+					store.WithDefaultWriteBatchSize(cfg.Store.WriteBatchSize),
+				}
+			},
+		),
```
Wondertan, line 37: One more thing. We really need to test that the provided `[]store.Option` is automatic and gets passed to the constructor's variadic `...store.Option` param. Right now, everything works, but it can false positive.

vgonkivs: @distractedm1nd has already created a task for unit testing.

vgonkivs: ok

Wondertan: Also. On the line 26 the value of the config is provided, while the provide is expecting a pointer. This makes me wonder if the Provide for the options is called at all

header/store/store.go
```diff
@@ -63,45 +50,58 @@ type store struct {
...
 
-func newStore(ds datastore.Batching) (*store, error) {
-	ds = namespace.Wrap(ds, storePrefix)
-	cache, err := lru.NewARC(DefaultStoreCacheSize)
+func newStore(ds datastore.Batching, opts ...Option) (*store, error) {
+	params := DefaultParameters()
+	for _, opt := range opts {
+		opt(params)
+	}
+
```
Wondertan, line 84: Just realized that we also should do validation pkg level validation here.

Wondertan (changes requested):

nodebuilder/header/module.go
Wondertan: No. The validate func is called before we provide anything

header/store/store.go
Wondertan, line 84: After we applied params I mean. We do the same for the DASer

header/store/options.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	if p.WriteBatchSize <= 0 {
+		return errors.New("invalid batch size")
+	}
+	return nil
+}
+
+// WithDefaultStoreCacheSize is a functional option that configures the
+// `StoreCacheSize` parameter.
+func WithDefaultStoreCacheSize(size int) Option {
```
renaynay, line 46: Why is it called `WithDefault` ? it is not the default, it is custom.

renaynay: this goes for every opt.

renaynay (approved):

header/store/options.go
Wondertan: Good catch

header/store/options.go
```diff
@@ -0,0 +1,66 @@
...
+	return &Parameters{
+		StoreCacheSize: 4096,
+		IndexCacheSize: 16384,
+		WriteBatchSize: 2048,
+	}
+}
+
+func (p *Parameters) Validate() error {
+	if p.StoreCacheSize <= 0 {
+		return errors.New("invalid store cache size")
```
derrandz, line 33: I recommend that the error be more information. We can add to this message: `value should be positive and non-zero`

header/store/options.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	if p.IndexCacheSize <= 0 {
+		return errors.New("invalid indexer cache size")
+	}
+	if p.WriteBatchSize <= 0 {
+		return errors.New("invalid batch size")
+	}
```
derrandz, line 40:
```suggestion
		return ErrInvalidSizeOpt("store cache")
	}
	if p.IndexCacheSize <= 0 {
		return ErrInvalidSizeOpt("indexer cache")
	}
	if p.WriteBatchSize <= 0 {
		return ErrInvalidSizeOpt("batch")
	}
```

With `ErrInvalidSizeOpt` defined somewhere at the top.

header/store/store.go
derrandz: 👍🏻

nodebuilder/header/module.go
derrandz: Double checked from the DASer PR, we are providing a slice, and I do remember reading in fx's documentation that variadic signatures are treated like slices, so we should be fine.

But nonetheless, @vgonkivs please test this. And if we have a unit test for it then that's gold!

derrandz (commented):
A bunch of nits really, keeping this as a comment.

header/store/options.go
distractedm1nd: i dont like this. too much indirection, not necessary

vgonkivs: Agree with Ryan. `value should be positive and non-zero` as a suffix sounds good for me

nodebuilder/header/module_test.go
```diff
@@ -0,0 +1,40 @@
...
+			func(s header.Store) {
+				ss := s.(*store.Store)
+				headerStore = ss
+			}),
+	)
+	require.NoError(t, app.Err())
+	require.Equal(t, headerStore.Params.StoreCacheSize, cfg.Store.StoreCacheSize)
+	require.Equal(t, headerStore.Params.IndexCacheSize, cfg.Store.IndexCacheSize)
+	require.Equal(t, headerStore.Params.WriteBatchSize, cfg.Store.WriteBatchSize)
+}
```
Wondertan, line 40: Perfetto
image

header/store/store.go
```diff
@@ -18,26 +18,13 @@ import (
...
-)
-
 var (
 	// errStoppedStore is returned for attempted operations on a stopped store
 	errStoppedStore = errors.New("stopped store")
 )
 
 // store implements the Store interface for ExtendedHeaders over Datastore.
-type store struct {
+type Store struct {
```
Wondertan, line 27: For the team. As per our sync with @vgonkivs, we need to have it public so that we can access Params and test that they are set correctly. It's not a big deal, but in return, we always know that the config is set correctly by the unit test, with FX indirections.

header/store/store.go
```diff
@@ -63,49 +50,66 @@ type store struct {
...
+
+	Params *Parameters
 }
 
 // NewStore constructs a Store over datastore.
 // The datastore must have a head there otherwise Start will error.
 // For first initialization of Store use NewStoreWithHead.
-func NewStore(ds datastore.Batching) (header.Store, error) {
-	return newStore(ds)
+func NewStore(ds datastore.Batching, opts ...Option) (header.Store, error) {
```
Wondertan, line 60: We should return the concrete type now and have a proxy constructor in the FX that returns the Store instead.

nodebuilder/header/module.go
Wondertan: We now have it! And it turned out in our case that variadic signature is not treated like slices, we potentially missed something last time, @derrandz and should double check, and even better, write the same test like @vgonkivs did

distractedm1nd (approved):
Looks great

Wondertan (approved):

renaynay (approved):

walldiss (approved):
Looks good!

Added nit suggestion

header/store/options.go
```diff
@@ -0,0 +1,70 @@
...
+func (p *Parameters) Validate() error {
+	if p.StoreCacheSize <= 0 {
+		return fmt.Errorf("invalid store cache size:%s", errSuffix)
+	}
+	if p.IndexCacheSize <= 0 {
+		return fmt.Errorf("invalid indexer cache size:%s", errSuffix)
+	}
+	if p.WriteBatchSize <= 0 {
+		return fmt.Errorf("invalid batch size:%s", errSuffix)
+	}
```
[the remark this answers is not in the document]
vgonkivs: idk what values can we count as a small. This should be discussed additionally I think.

> Other suggestion is if user provided invalid option values, launch with default value and log.Warn user mistake. Up to you, whatever you find a better user experience.

For now I think it's not necessary, as invalid options are `<=0`.

@Wondertan , @renaynay wdyt?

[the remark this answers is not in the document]
vgonkivs: Discussed with @renaynay internally and we think that it's not relevant right now.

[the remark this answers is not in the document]
Wondertan: > Other suggestion is if user provided invalid option values, launch with default value and log.Warn user mistake. Up to you, whatever you find a better user experience.

We should hardly fail and not allow running the node if configured incorrectly. Users will likely ignore the Warn log and will think that the configuration was actually correct.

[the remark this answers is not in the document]
Wondertan: The small value protection is valid

PR #1395, opened 2022-11-24
feat(share): eds.Blockstore

### LINKED ISSUE #1115: share: `EDSStore.Blockstore`

distractedm1nd:
## Overview
This PR adds a blockstore to `eds.Store`. It was extracted from the eds.Store scaffolding PR. Tests are still needed, so still in draft for now.

Closes #1115, #1427

## Checklist

- ~~Visual proof for any user facing features like CLI or documentation updates~~ NA


share/eds/blockstore.go
```diff
@@ -0,0 +1,201 @@
...
+	if len(keys) != 1 {
+		return nil, ErrInvalidShardCount
+	}
+
+	// try to fetch from cache
+	shardKey := keys[0]
+	blockstr, err := bs.readFromBSCache(shardKey)
+	if err == nil && blockstr != nil {
+		return blockstr, nil
+	}
```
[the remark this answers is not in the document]
distractedm1nd: We can really only debug log this. Because MOST calls will have err != nil, since most calls will miss the cache. But adding a debug log

share/eds/blockstore.go
```diff
@@ -0,0 +1,201 @@
...
+		return false, err
+	}
+	return len(keys) > 0, nil
+}
+
+func (bs *Blockstore) Get(ctx context.Context, cid cid.Cid) (blocks.Block, error) {
+	blockstr, err := bs.getReadOnlyBlockstore(ctx, cid)
+	if err != nil {
+		return nil, ipld.ErrNotFound{Cid: cid}
+	}
```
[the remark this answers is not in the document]
distractedm1nd: good catch, thanks, added, can you check?

share/eds/store.go
```diff
@@ -69,13 +71,20 @@ func NewStore(basepath string, ds datastore.Batching) (*Store, error) {
...
 		topIdx:   invertedRepo,
 		carIdx:   fsRepo,
 		mounts:   r,
-	}, nil
+	}
+
+	store.bs, err = NewEDSBlockstore(store)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create EDSBlockstore: %w", err)
+	}
```
Wondertan, line 85: optional nit: creat the blockstore first and then the store

share/eds/blockstore.go
```diff
@@ -0,0 +1,210 @@
...
+type Blockstore struct {
+	store *Store
+
+	// bsStripedLocks prevents simultaneous RW access to the blockstore cache for a shard. Instead
+	// of using only one lock or one lock per key, we stripe the shard keys across 256 locks. 256 is
+	// chosen because it 0-255 is the range of values we get looking at the last byte of the key.
+	bsStripedLocks [256]sync.Mutex
+	// caches the blockstore for a given shard for shard read affinity i.e.
+	// further reads will likely be from the same shard. Maps (shard key -> blockstore).
+	blockstoreCache *lru.Cache
```
Wondertan, line 49: It would be nice to update to the latest version of the lru lib, but non-blocking

share/eds/blockstore.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+	return len(keys) > 0, nil
+}
+
+func (bs *Blockstore) Get(ctx context.Context, cid cid.Cid) (blocks.Block, error) {
+	blockstr, err := bs.getReadOnlyBlockstore(ctx, cid)
+	if err != nil {
+		log.Debugf("failed to get blockstore for cid %s: %s", cid, err)
+		// nmt's GetNode expects an ipld.ErrNotFound when a cid is not found.
+		return nil, ipld.ErrNotFound{Cid: cid}
```
Wondertan, line 84: This could become a debugging foot gun, but at least we are logging the real error. It might be the case, though, that the blockstore from the accessor already returns it.

share/eds/blockstore.go
```diff
@@ -0,0 +1,210 @@
...
+		blockstoreCache: bslru,
+	}, nil
+}
+
+func (bs *Blockstore) Has(ctx context.Context, cid cid.Cid) (bool, error) {
+	keys, err := bs.store.dgstr.ShardsContainingMultihash(ctx, cid.Hash())
+	if err != nil {
+		return false, fmt.Errorf("failed to find shards containing multihash: %w", err)
+	}
+	return len(keys) > 0, nil
```
Wondertan, line 76: What if we have an inconsistency between the index and the car file for some reason? Should we still call Has on the real blockstore?

Wondertan (commented):
Mostly nits and questions. Nothing criminal was found. The locking design is legit

share/eds/blockstore.go
```diff
@@ -0,0 +1,210 @@
...
+	blockstr, err := bs.getReadOnlyBlockstore(ctx, cid)
+	if err != nil {
+		return 0, err
+	}
+	return blockstr.GetSize(ctx, cid)
+}
+
+// DeleteBlock is a noop on the EDS Blockstore that returns an ErrUnsupportedOperation when called.
+func (bs *Blockstore) DeleteBlock(context.Context, cid.Cid) error {
+	return ErrUnsupportedOperation
```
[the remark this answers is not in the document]
distractedm1nd: Unfortunately no, because some of the operations (Put and PutMany) are called and aren't allowed to error. I need to find the part of code that causes this again, IIRC it's in a part of bitswap where they also comment that it's a weird behavior.

Also I am against panicking where we don't have to - the cast down below is probably okay to panic on

share/eds/blockstore.go
```diff
@@ -0,0 +1,210 @@
...
+	"github.com/ipfs/go-cid"
+	blockstore "github.com/ipfs/go-ipfs-blockstore"
+	ipld "github.com/ipfs/go-ipld-format"
+)
+
+var _ blockstore.Blockstore = (*Blockstore)(nil)
+
+var (
+	maxCacheSize            = 100
+	ErrUnsupportedOperation = errors.New("unsupported operation")
```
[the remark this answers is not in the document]
distractedm1nd: good catch

share/eds/blockstore.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+// getReadOnlyBlockstore finds the underlying blockstore of the shard that contains the given CID.
+func (bs *Blockstore) getReadOnlyBlockstore(ctx context.Context, cid cid.Cid) (dagstore.ReadBlockstore, error) {
+	keys, err := bs.store.dgstr.ShardsContainingMultihash(ctx, cid.Hash())
+	if err != nil {
+		return nil, fmt.Errorf("failed to find shards containing multihash: %w", err)
+	}
+	if len(keys) != 1 {
+		return nil, ErrInvalidShardCount
```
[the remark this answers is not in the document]
distractedm1nd: I see, very good catch. I was convinced the CID will always only exist once, but yes a share can be found in multiple shards 

Upstream dagstore abstracts this into a "shardSelectorFn" but we definitely dont need this. We will just take the first one as you suggest

share/eds/blockstore.go
```diff
@@ -0,0 +1,210 @@
...
+		return nil, errors.New("not found in cache")
+	}
+
+	rbs, ok := val.(*accessorWithBlockstore)
+	if !ok {
+		return nil, fmt.Errorf(
+			"casting value from cache to accessorWithBlockstore: %s",
+			reflect.TypeOf(val),
+		)
+	}
```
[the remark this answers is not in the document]
distractedm1nd: @renaynay debate

share/eds/blockstore.go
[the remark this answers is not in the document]
walldiss: You can check if error is not a `cache miss` one and log. So logs happens only on unexpected errors.

share/eds/blockstore.go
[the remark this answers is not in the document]
walldiss: Should we panic on non-critical execution pass? We can `log.Error` it so node keeps being operational if it happens. I would suggest creating an variable for error:
```
var errNotFound = errors.New("not found in cache")
```
Return it and if returned error is not `errNotFound` log an error.

share/eds/blockstore.go
```diff
@@ -0,0 +1,210 @@
...
+	blocks "github.com/ipfs/go-block-format"
+	"github.com/ipfs/go-cid"
+	blockstore "github.com/ipfs/go-ipfs-blockstore"
+	ipld "github.com/ipfs/go-ipld-format"
+)
+
+var _ blockstore.Blockstore = (*Blockstore)(nil)
+
+var (
+	maxCacheSize            = 100
```
walldiss, line 22: Just curious, how this value is chosen? 
Also u
sually preallocated buffer sizes tends to be power of 2, because it fits memory layout better.

share/eds/blockstore.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+	if len(keys) != 1 {
+		return nil, ErrInvalidShardCount
+	}
+
+	// try to fetch from cache
+	shardKey := keys[0]
+	blockstr, err := bs.readFromBSCache(shardKey)
+	if err != nil {
+		log.Debugf("blockstore cache miss for key %s: %s", shardKey, err)
```
renaynay, line 141:
```suggestion
		log.Debugf("blockstore cache missing for key %s: %s", shardKey, err)
```

share/eds/blockstore.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+	if err == nil && blockstr != nil {
```
renaynay, line 143: err will == nil here always, just check if blockstr != nil

share/eds/blockstore.go
```diff
@@ -0,0 +1,210 @@
...
+
+func NewEDSBlockstore(s *Store) (*Blockstore, error) {
+	// instantiate the blockstore cache
+	bslru, err := lru.NewWithEvict(maxCacheSize, func(_ interface{}, val interface{}) {
+		// ensure we close the blockstore for a shard when it's evicted from the cache so dagstore can gc
+		// it.
+		abs, ok := val.(*accessorWithBlockstore)
+		if ok {
+			abs.sa.Close()
+		}
```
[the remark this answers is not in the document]
renaynay: ideally like you did for the other

```
	if !ok {
		return nil, fmt.Errorf(
			"casting value from cache to accessorWithBlockstore: %s",
			reflect.TypeOf(val),
		)
	}
```

share/eds/blockstore.go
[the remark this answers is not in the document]
renaynay: There is nothing to really debate since both are technically fine but, I don't really see a reason to panic here since it isn't critical like @walldiss mentioned.

share/eds/blockstore.go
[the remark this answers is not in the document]
distractedm1nd: can't return an error here but I can log

share/eds/blockstore.go
distractedm1nd: Inconsistency as in, the CAR file changes somehow after indexing?

share/eds/store.go
distractedm1nd: But the blockstore requires the store for the backreference, or am I missing something?

share/eds/blockstore.go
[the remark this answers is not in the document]
distractedm1nd: can you check?

share/eds/blockstore.go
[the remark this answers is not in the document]
distractedm1nd: took @walldiss 's suggestion

share/eds/blockstore.go
distractedm1nd: Yeah, don't know why I picked 100, changed to 128, thanks!

share/eds/blockstore.go
distractedm1nd: changed logic a bit here, check again

share/eds/blockstore.go
distractedm1nd: done

share/eds/blockstore.go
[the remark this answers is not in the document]
distractedm1nd: done

share/eds/blockstore.go
```diff
@@ -0,0 +1,210 @@
...
+type accessorWithBlockstore struct {
+	sa *dagstore.ShardAccessor
+	bs dagstore.ReadBlockstore
+}
+
+// Blockstore implements the blockstore.Blockstore interface on an EDSStore.
+// The lru cache approach is heavily inspired by the open PR filecoin-project/dagstore/116.
+// The main differences to the implementation here are that we do not support multiple shards per
+// key, call GetSize directly on the underlying RO blockstore, and do not throw errors on
+// Put/PutMany. Also, we do not abstract away the blockstore operations.
```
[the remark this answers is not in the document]
distractedm1nd: done

share/eds/blockstore.go
```diff
@@ -0,0 +1,210 @@
...
+	ErrInvalidShardCount    = errors.New("the provided cid does not map to exactly one shard")
+)
+
+type accessorWithBlockstore struct {
+	sa *dagstore.ShardAccessor
+	bs dagstore.ReadBlockstore
+}
+
+// Blockstore implements the blockstore.Blockstore interface on an EDSStore.
+// The lru cache approach is heavily inspired by the open PR filecoin-project/dagstore/116.
```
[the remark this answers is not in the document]
distractedm1nd: done

share/eds/blockstore.go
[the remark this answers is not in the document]
distractedm1nd: done

share/eds/blockstore.go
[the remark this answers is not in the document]
Wondertan: > Should we panic on non-critical execution pass?

Guys, it's critical. If we get the value from the map, which we are the only writers - it's something really wrong.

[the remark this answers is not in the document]
Wondertan: Unfortunately, we are not in the rust world to catch such things during compile time, but we are in the Golang world where we have panics that are used when something impossible happens. This is the case where something impossible happens, and we should panic here, so once it becomes possible due to some new dev coming here and changing, it breaks early locally, in the test network, or the integration test.

share/eds/blockstore.go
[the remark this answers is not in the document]
walldiss: Why don't you return wrapped error from `getReadOnlyBlockstore` here?

share/eds/blockstore.go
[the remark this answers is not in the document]
renaynay: fair point @Wondertan

share/eds/accessor_cache.go
```diff
@@ -0,0 +1,106 @@
...
+	"reflect"
+	"sync"
+
+	"github.com/filecoin-project/dagstore"
+	"github.com/filecoin-project/dagstore/shard"
+	lru "github.com/hashicorp/golang-lru"
+)
+
+var (
+	maxCacheSize = 128
```
Wondertan, line 15: We need to have this configurable. Beefy nodes, like our bootstrappers, may want to have it much bigger

share/eds/accessor_cache.go
```diff
@@ -0,0 +1,106 @@
...
+			log.Errorf("couldn't close accessor after cache eviction: %s", err)
+		}
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to instantiate blockstore cache: %w", err)
+	}
+	return &blockstoreCache{cache: bslru}, nil
+}
+
+func (bc *blockstoreCache) Read(shardContainingCid shard.Key) (*accessorWithBlockstore, error) {
```
Wondertan, line 60: Add and Get? Read/Write seems like a misterm here

share/eds/accessor_cache.go
```diff
@@ -0,0 +1,106 @@
...
+		return nil, errCacheMiss
+	}
+
+	accessor, ok := val.(*accessorWithBlockstore)
+	if !ok {
+		return nil, fmt.Errorf(
+			"casting value from cache to accessorWithBlockstore: %s",
+			reflect.TypeOf(val),
+		)
+	}
```
Wondertan, line 77: pls

share/eds/blockstore.go
Wondertan: Ok, cool

share/eds/accessor_cache.go
```diff
@@ -0,0 +1,106 @@
...
+
+// accessorWithBlockstore is the value that we store in the blockstore cache
+type accessorWithBlockstore struct {
+	sa *dagstore.ShardAccessor
+	// blockstore is stored separately because each access to the blockstore over the shard accessor
+	// reopens the underlying CAR.
+	bs dagstore.ReadBlockstore
+}
+
+type blockstoreCache struct {
```
Wondertan, line 27: Matt in me: "where are units for this struct?"

share/eds/blockstore.go
```diff
@@ -0,0 +1,115 @@
...
+// the blockstore operations.
+//
+// The intuition here is that each CAR file is its own blockstore, so we need this top level
+// implementation to allow for the blockstore operations to be routed to the underlying stores.
+type Blockstore struct {
+	store *Store
+	cache *blockstoreCache
+}
+
+func NewEDSBlockstore(store *Store, cache *blockstoreCache) *Blockstore {
```
Wondertan, line 35: No one can create a blockstoreCache so this is effectively private. Also, I don't see a use case for this outside of the pkg and suggest privatizing. The struct and its constructor

share/eds/accessor_cache.go
distractedm1nd: okay fine

share/eds/blockstore.go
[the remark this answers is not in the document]
distractedm1nd: If I wrap the error will the equality check I commented about of err == ipld.ErrNotFound still work?

share/eds/accessor_cache.go
distractedm1nd: done

share/eds/accessor_cache.go
distractedm1nd: Is now passed to the constructor, any more configurability (either through config or functional params) needs to be a later PR, after it is being supplied to FX (which doesn't happen here)

share/eds/blockstore.go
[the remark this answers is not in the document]
distractedm1nd: we now panic

share/eds/accessor_cache.go
Wondertan: 👌

Wondertan (commented):
One last comment and LFG

share/eds/blockstore.go
```diff
@@ -0,0 +1,115 @@
...
+// it is unclear which CAR file the block should be written to.
+func (bs *blockstore) Put(context.Context, blocks.Block) error {
+	return nil
+}
+
+// PutMany is a noop on the EDS blockstore, but it does not return an error because it is called by
+// bitswap. For clarification, an implementation of PutMany does not make sense in this context
+// because it is unclear which CAR file the blocks should be written to.
+func (bs *blockstore) PutMany(context.Context, []blocks.Block) error {
+	return nil
```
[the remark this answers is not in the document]
distractedm1nd: It must not return an error or bitswap will error, read comment

share/eds/cache_test.go
```diff
@@ -0,0 +1,12 @@
...
+import (
+	"testing"
+)
+
+func TestBlockstoreCache_Add(t *testing.T) {
+}
+
+func TestBlockstoreCache_Get(t *testing.T) {
+	t.Skip("not implemented")
+}
```
Wondertan, line 12: ?

distractedm1nd: oh shit that file should have been deleted

share/eds/store.go
Wondertan: It's a private structure so that you can set up the Store afterward, but that's a very minor thing. Not worth spending any more cycles on it

Wondertan: The way it is right now looks very good, so resolved

Wondertan (approved):
🚢 it

share/eds/blockstore.go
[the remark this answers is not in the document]
distractedm1nd: #1440

walldiss (dismissed):

share/eds/cache_test.go
```diff
@@ -0,0 +1,12 @@
+package eds
+
+import (
+	"testing"
+)
+
+func TestBlockstoreCache_Add(t *testing.T) {
```
renaynay: doesn't this also need a t.Skip?

renaynay (approved):

share/eds/cache_test.go
distractedm1nd: this test is gone

share/eds/blockstore.go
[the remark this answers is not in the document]
distractedm1nd: resolving this because the error check is in a different repo and its not worth the propagation here, logging should be enough for now

PR #1400, opened 2022-11-27
core: Deprecate/remove StartTestClient in favour of StartTestCoreWithApp

### LINKED ISSUE #1365: core: Deprecate/remove `StartTestClient` in favour of `StartTestCoreWithApp`
Replace all uses of `StartTestClient` with `StartTestCoreWithApp` in the codebase.

Based on merging of #1364

dougEfresh:
Fixes #1365

## Overview

Replace StartTestClient with StartTestCoreWithApp within tests

### DIFF
--- a/core/client_test.go
+++ b/core/client_test.go
@@ -13,7 +13,7 @@ func TestRemoteClient_Status(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
 	t.Cleanup(cancel)
 
-	_, client := StartTestClient(ctx, t)
+	_, client := StartTestCoreWithApp(t)
 	status, err := client.Status(ctx)
 	require.NoError(t, err)
 	require.NotNil(t, status)
@@ -23,7 +23,7 @@ func TestRemoteClient_StartBlockSubscription_And_GetBlock(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
 	t.Cleanup(cancel)
 
-	_, client := StartTestClient(ctx, t)
+	_, client := StartTestCoreWithApp(t)
 	eventChan, err := client.Subscribe(ctx, newBlockSubscriber, newBlockEventQuery)
 	require.NoError(t, err)
 
--- a/core/client_test.go
+++ b/core/client_test.go
@@ -10,7 +10,7 @@ import (
 )
 
 func TestRemoteClient_Status(t *testing.T) {
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*3)
 	t.Cleanup(cancel)
 
 	_, client := StartTestCoreWithApp(t)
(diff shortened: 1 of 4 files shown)


core/testing.go
```diff
@@ -52,26 +52,16 @@ func CreateKVStore(retainBlocks int64) *kvstore.Application {
...
+func getFreePort() (port int, err error) {
+	var a *net.TCPAddr
+	if a, err = net.ResolveTCPAddr("tcp", "localhost:0"); err == nil {
+		var l *net.TCPListener
+		if l, err = net.ListenTCP("tcp", a); err == nil {
+			defer l.Close()
+			return l.Addr().(*net.TCPAddr).Port, nil
+		}
+	}
+	return
```
Wondertan, line 64: Doesn't the zero port allocate the free one on the kernel? Why do we need to do it manually?

dougEfresh: Later in the code there is a [GetEndpoint](https://github.com/celestiaorg/celestia-node/blob/main/core/testing.go#L116-L128) which is used to build the test client.

```
func GetEndpoint(cfg *config.Config) (string, error) {
	url, err := url.Parse(cfg.RPC.ListenAddress)
	if err != nil {
		return "", err
	}
	host, _, err := net.SplitHostPort(url.Host)
	if err != nil {
		return "", err
	}
	return fmt.Sprintf("%s:%s", host, url.Port()), nil
}
```

So I need to know which port the kernel gave the test in order to build the  client.  
I do this [here](https://github.com/celestiaorg/celestia-node/pull/1400/files#diff-9750405595d72572f9a78b11e628a3cf4331798bb84dcd88746e61a8ba215657L91)
```
	// get a random port for running test in parallel
	freePort, err := getFreePort()
	require.NoError(t, err)
	tmNode.Config().RPC.ListenAddress = fmt.Sprintf("tcp://127.0.0.1:%d", freePort)
	freePort, err = getFreePort()
	require.NoError(t, err)
	tmNode.Config().P2P.ListenAddress = fmt.Sprintf("tcp://0.0.0.0:%d", freePort)

```

core/client_test.go
```diff
@@ -10,20 +10,20 @@ import (
 )
 
 func TestRemoteClient_Status(t *testing.T) {
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*3)
```
dougEfresh, line 13: I increased the timeout because I saw  random timeouts when running tests locally

core/testing.go
Wondertan: Ok, I see. So we need RPC to get a port and create a client. Why do we need to get a free port for a P2P listen-to address, though? @dougEfresh

Wondertan: Yeah, so let's just set it to 0

Wondertan: and for RPC we can use a freePort

Wondertan: But actually, this is a very minor thing to worry about, and using getFreePort is not that bad.
I just don't really like doing what kernel already provides, but its fine in this case

Wondertan (approved):

core/testing.go
dougEfresh: Ah, I see what you mean. Minor thing yes, but I changed p2p port to use port 0 and tests pass.

vgonkivs (approved):
LGTM

renaynay (approved):
Thanks for the contribution @dougEfresh :)

distractedm1nd (approved):
🫡

Issue #1187, opened 2022-09-30
add authorization


renaynay:
- [ ] scaffolding for perms
- [ ] jwt token gen (in node module)
- [ ] add header and token in client wrapper 
- [ ] add auth methods on node module

PR #1402, opened 2022-11-28
feat(api/rpc): Implement auth middleware on `Server`

renaynay:
Related to #1187 

**This PR is based on #1313** 

**It implements:** 
* providing a new JWT secret from the `nodebuilder/node` module to both the node module and the rpc server.
* auth middleware for the `rpc.Server` that verifies the given token with the server's secret

In a follow-up PR will come the ability to create an rpc client with elevated permissions (this branch restricts client access to read-only API methods).

### DIFF
--- a/api/rpc/permissions/permissions.go
+++ b/api/rpc/permissions/permissions.go
@@ -3,9 +3,9 @@ package permissions
 import "github.com/filecoin-project/go-jsonrpc/auth"
 
 var (
-	DefaultPerms   = []auth.Permission{"read"}
-	ReadWritePerms = []auth.Permission{"read", "write"}
-	AllPerms       = []auth.Permission{"read", "write", "admin"}
+	DefaultPerms   = []auth.Permission{"public"}
+	ReadWritePerms = []auth.Permission{"public", "read", "write"}
+	AllPerms       = []auth.Permission{"public", "read", "write", "admin"}
 )
 
 // JWTPayload is a utility struct for marshaling/unmarshalling
--- a/api/rpc/permissions/permissions.go
+++ b/api/rpc/permissions/permissions.go
@@ -1,0 +1,15 @@
+package permissions
+
+import "github.com/filecoin-project/go-jsonrpc/auth"
+
+var (
+	DefaultPerms   = []auth.Permission{"read"}
+	ReadWritePerms = []auth.Permission{"read", "write"}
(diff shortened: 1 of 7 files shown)


nodebuilder/node/module.go
```diff
@@ -1,14 +1,16 @@
...
 import (
+	"github.com/cristalhq/jwt"
 	"go.uber.org/fx"
 )
 
 func ConstructModule(tp Type) fx.Option {
 	return fx.Module(
 		"node",
-		fx.Provide(func() Module {
+		fx.Provide(func(secret jwt.Signer) Module {
```
renaynay, line 11: This will be actually used in a later PR (where node module will be able to generate new tokens and verify perms of tokens w the given secret

api/rpc/permissions/permissions.go
```diff
@@ -0,0 +1,15 @@
+package permissions
+
+import "github.com/filecoin-project/go-jsonrpc/auth"
+
+var (
+	DefaultPerms   = []auth.Permission{"read"}
```
distractedm1nd, line 6: this is now incorrect, dont forget public

distractedm1nd (changes requested):

api/rpc/permissions/permissions.go
distractedm1nd: this is why unit tests are failing btw

vgonkivs (approved):

distractedm1nd (approved):

api/rpc/server.go
```diff
@@ -2,51 +2,66 @@ package rpc
...
 	"reflect"
 	"sync/atomic"
 	"time"
 
+	"github.com/cristalhq/jwt"
 	"github.com/filecoin-project/go-jsonrpc"
 	"github.com/filecoin-project/go-jsonrpc/auth"
 	logging "github.com/ipfs/go-log/v2"
+
+	"github.com/celestiaorg/celestia-node/api/rpc/permissions"
```
Wondertan, line 17: nit: Name pkg as common`perms`?

Wondertan (approved):
Glad that jsonrpc does not actually enforce a specific auth dependency. Would also be nice to have some tests here, but I assume they will come in later PRs

walldiss (approved):

api/rpc/server.go
renaynay: I will do this in next PR 🙏🏻

PR #1405, opened 2022-11-28
improvement(header/p2p): unify sendMessage method to request headers

### LINKED ISSUE #1404: header/p2p: use ctx.Deadline to set stream deadline instead of constants.
Self-explanatory.

vgonkivs:
Resolves #1404

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -16,7 +16,7 @@ linters:
     # - gocyclo
     # - godox
     - gofmt
-    - goimports
+#    - goimports
     # - golint - deprecated since v1.41. revive will be used instead
     - revive
     - gosec
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -16,7 +16,7 @@ linters:
     # - gocyclo
     # - godox
     - gofmt
-#    - goimports
+    - goimports
     # - golint - deprecated since v1.41. revive will be used instead
     - revive
     - gosec
--- a/header/p2p/exchange.go
+++ b/header/p2p/exchange.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"io"
 	"math/rand"
 	"sort"
 	"time"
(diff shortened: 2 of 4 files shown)


header/p2p/exchange.go
```diff
@@ -288,3 +259,57 @@ func convertStatusCodeToError(code p2p_pb.StatusCode) error {
 		return fmt.Errorf("unknown status code %d", code)
 	}
 }
+
+// sendMessage opens the stream to the given peers and sends ExtendedHeaderRequest to fetch ExtendedHeaders.
+func sendMessage(
```
Wondertan, line 264: nit: can you group code blocks nicer?

header/p2p/exchange.go
```diff
@@ -288,3 +259,57 @@ func convertStatusCodeToError(code p2p_pb.StatusCode) error {
...
+	}
+	// send request
+	_, err = serde.Write(stream, req)
+	if err != nil {
+		stream.Reset() //nolint:errcheck
+		return nil, 0, 0, err
+	}
+	err = stream.CloseWrite()
+	if err != nil {
+		log.Error(err)
```
Wondertan, line 293: We cannot ignore this error, iirc. If it errors, the stream is unusable

header/p2p/exchange.go
```diff
@@ -288,3 +259,57 @@ func convertStatusCodeToError(code p2p_pb.StatusCode) error {
...
+			stream.Reset() //nolint:errcheck
+			return nil, 0, 0, err
+		}
+		totalRequestSize += uint64(msgSize)
+		headers = append(headers, resp)
+	}
+	duration := time.Since(startTime).Milliseconds()
+	if err = stream.Close(); err != nil {
+		log.Errorw("closing stream", "err", err)
+	}
```
Wondertan, line 313: Same here, I would return this error

Wondertan (commented):
Thanks for doing this!

header/p2p/exchange.go
vgonkivs: But we do not need stream anymore. Error can appear on closing the stream

Wondertan: No, we need to read the data out of it

vgonkivs: So, if we are getting error here, then we should immediately return it?

.golangci.yml
```diff
@@ -16,7 +16,7 @@ linters:
     # - gocyclo
     # - godox
     - gofmt
-    - goimports
+#    - goimports
```
distractedm1nd, line 19: why

header/p2p/helpers.go
```diff
@@ -0,0 +1,89 @@
...
+
+	"github.com/celestiaorg/celestia-node/header"
+	p2p_pb "github.com/celestiaorg/celestia-node/header/p2p/pb"
+)
+
+func protocolID(protocolSuffix string) protocol.ID {
+	return protocol.ID(fmt.Sprintf("/header-ex/v0.0.3/%s", protocolSuffix))
+}
+
+// sendMessage opens the stream to the given peers and sends ExtendedHeaderRequest to fetch ExtendedHeaders.
```
distractedm1nd, line 23: I know it didnt exist before but can you add to the godoc what the return values represent? mainly the two uints

distractedm1nd (commented):
I'm generally against helper files where we can avoid them since it means we don't know where else to put them. I can imagine sendMessage being in its own file, but that leaves the other two helpers alone. So I think it's justified here

.golangci.yml
vgonkivs: forgot to uncomment

header/p2p/helpers.go
vgonkivs: Yes, good point. Thank you

header/p2p/exchange.go
Wondertan: Yes

header/p2p/exchange.go
Wondertan: If it's not closed properly, it likely means that the written data to the stream was not sent. Think of Close as the final Flush on the buffer as well.

vgonkivs: Discussed internally: we shouldn't return the error in this case, otherwise we fail the entire request (which was probably succeed)

Wondertan: Yeah, I mixed up the read and write sides in the comment.

distractedm1nd (dismissed):

header/p2p/helpers.go
```diff
@@ -0,0 +1,91 @@
...
+// As a result sendMessage returns ExtendedHeaderResponse, the size of fetched data,
+// the duration of the request and an error.
+func sendMessage(
```
Wondertan, line 26: The code in here is still not grouped, but that's nit

Wondertan (dismissed):

header/p2p/helpers.go
vgonkivs: done

renaynay (approved):

Wondertan (approved):

PR #1409, opened 2022-11-30
feat(share): Ensure empty CAR exists in eds.Store

### LINKED ISSUE #1113: node: Ensure `EDStore` has EDS of an empty block
## Context

The empty block is valid and small EDS. It can happen in the early stage of the network. Its body is constant, and to avoid
transferring it over the wire, the `EDSStore` should be pre-initialized with an empty EDS value.

## Refs

Depends on #1114

distractedm1nd:
## Overview
Closes #1113 
This PR adds an empty CAR to the eds.Store during startup of the store. 
## Checklist

### DIFF
--- a/nodebuilder/share/constructors.go
+++ b/nodebuilder/share/constructors.go
@@ -57,7 +57,7 @@ func newModule(lc fx.Lifecycle, bServ blockservice.BlockService, avail share.Ava
 	return &module{serv}
 }
 
-// ensureEmptyCARExists adds the empty EDS from share.EnsureEmptySquareExists to the provided EDS store.
+// ensureEmptyCARExists adds an empty EDS to the provided EDS store.
 func ensureEmptyCARExists(ctx context.Context, store *eds.Store) error {
 	emptyEDS := share.EmptyExtendedDataSquare()
 	emptyDAH := da.NewDataAvailabilityHeader(emptyEDS)
--- a/nodebuilder/share/constructors.go
+++ b/nodebuilder/share/constructors.go
@@ -59,14 +59,10 @@ func newModule(lc fx.Lifecycle, bServ blockservice.BlockService, avail share.Ava
 
 // ensureEmptyCARExists adds the empty EDS from share.EnsureEmptySquareExists to the provided EDS store.
 func ensureEmptyCARExists(ctx context.Context, store *eds.Store) error {
-	bServ := blockservice.New(store.Blockstore(), nil)
-	// we ignore the error because we know that the batchAdder will not be able to commit to our
-	// Blockstore, which is not meant for writes.
-	eds, _ := share.EnsureEmptySquareExists(ctx, bServ)
+	emptyEDS := share.EmptyExtendedDataSquare()
+	emptyDAH := da.NewDataAvailabilityHeader(emptyEDS)
 
-	dah := da.NewDataAvailabilityHeader(eds)
-
-	err := store.Put(ctx, dah.Hash(), eds)
+	err := store.Put(ctx, emptyDAH.Hash(), emptyEDS)
 	if errors.Is(err, dagstore.ErrShardExists) {
 		return nil
 	}
--- a/nodebuilder/share/constructors.go
+++ b/nodebuilder/share/constructors.go
@@ -2,7 +2,9 @@ package share
 
 import (
 	"context"
+	"errors"
 
+	"github.com/filecoin-project/dagstore"
 	"github.com/ipfs/go-blockservice"
 	"github.com/ipfs/go-datastore"
 	"github.com/libp2p/go-libp2p-core/host"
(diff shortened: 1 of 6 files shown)


share/eds/store.go
```diff
@@ -298,6 +304,17 @@ func (s *Store) Has(ctx context.Context, root share.Root) (bool, error) {
...
+// ensureEmptyCARExists adds the empty EDS from share.EnsureEmptySquareExists
+func (s *Store) ensureEmptyCARExists(ctx context.Context) error {
+	bServ := blockservice.New(s.Blockstore(), nil)
+	// we ignore the error because we know that the batchAdder will not be able to commit to our
+	// Blockstore, which is not meant for writes.
+	eds, _ := share.EnsureEmptySquareExists(ctx, bServ)
+
+	dah := da.NewDataAvailabilityHeader(eds)
+	return s.Put(ctx, dah, eds)
+}
```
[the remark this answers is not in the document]
distractedm1nd: agreed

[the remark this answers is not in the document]

nodebuilder/share/module.go
```diff
@@ -31,6 +30,7 @@ func ConstructModule(tp node.Type, cfg *Config, options ...fx.Option) fx.Option
 		return fx.Module(
 			"share",
 			baseComponents,
+			fx.Invoke(share.EnsureEmptySquareExists),
```
Wondertan, line 33: Should we still keep this in the base, until we fully migrate to the Store

nodebuilder/share/module.go
```diff
@@ -52,13 +52,21 @@ func ConstructModule(tp node.Type, cfg *Config, options ...fx.Option) fx.Option
...
 				},
-				fx.OnStart(func(ctx context.Context, eds *eds.Store) error {
-					return eds.Start(ctx)
+				fx.OnStart(func(ctx context.Context, store *eds.Store) error {
+					err := store.Start(ctx)
+					if err != nil {
+						return err
+					}
+
+					return ensureEmptyCARExists(ctx, store)
```
Wondertan, line 61: Let's do it in Invoke as we do for blockservice? We won't need a forceful Invoke this way, either.

nodebuilder/share/constructors.go
```diff
@@ -52,3 +56,19 @@ func newModule(lc fx.Lifecycle, bServ blockservice.BlockService, avail share.Ava
 	})
 	return serv
 }
+
+// ensureEmptyCARExists adds the empty EDS from share.EnsureEmptySquareExists to the provided EDS store.
+func ensureEmptyCARExists(ctx context.Context, store *eds.Store) error {
+	bServ := blockservice.New(store.Blockstore(), nil)
+	// we ignore the error because we know that the batchAdder will not be able to commit to our
+	// Blockstore, which is not meant for writes.
+	eds, _ := share.EnsureEmptySquareExists(ctx, bServ)
```
Wondertan, line 65: Feels a bit ugly. What if we make another helper in share that returns us an empty block EDS?

nodebuilder/share/module.go
distractedm1nd: Can't, because invokes happen before start is successful. If we invoke this, it will hang forever because the EDS store is not started

nodebuilder/share/constructors.go
distractedm1nd: done, using da.ExtendShares

nodebuilder/share/module.go
distractedm1nd: done

share/empty.go
```diff
@@ -3,24 +3,33 @@ package share
...
-	_, err := AddShares(ctx, shares, bServ)
-	return err
+// EmptyExtendedDataSquare returns the EDS of the empty block data square.
+func EmptyExtendedDataSquare() *rsmt2d.ExtendedDataSquare {
+	shares := emptyDataSquare()
+	squareSize := uint64(math.Sqrt(float64(appconsts.MinSquareSize)))
+	eds, err := da.ExtendShares(squareSize, shares)
+	if err != nil {
+		panic(fmt.Errorf("failed to create empty EDS: %w", err))
+	}
```
vgonkivs, line 31: why panic and not just return an error here?

Wondertan (dismissed):

share/empty.go
distractedm1nd: Only ever gets called on startup, and if we cant create an empty EDS something is VERY wrong

Wondertan (approved):

share/empty.go
vgonkivs: But if we return an error, then the node will not start. Or it is mostly done to be able to identify this case?

renaynay (dismissed):

vgonkivs (approved):

renaynay (approved):

PR #1410, opened 2022-11-30
fix(header): add extra arg to fix build

evan-forbes:
## Overview

main isn't building due to a change in the api of a dependency and a recent PR. this PR adds an arbitrary number to fix that. 

NOTE: I don't actually know what number we should put here. 3 seemed like a good number, so I used it.

## Checklist

### DIFF
--- a/header/p2p/exchange.go
+++ b/header/p2p/exchange.go
@@ -160,7 +160,7 @@ func (ex *Exchange) GetVerifiedRange(
 	session := newSession(ex.ctx, ex.host, ex.peerTracker.peers(), ex.protocolID)
 	defer session.close()
 
-	headers, err := session.getRangeByHeight(ctx, uint64(from.Height)+1, amount)
+	headers, err := session.getRangeByHeight(ctx, uint64(from.Height)+1, amount, 3)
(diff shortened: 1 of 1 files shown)


header/p2p/exchange.go
```diff
@@ -160,7 +160,7 @@ func (ex *Exchange) GetVerifiedRange(
 	session := newSession(ex.ctx, ex.host, ex.peerTracker.peers(), ex.protocolID)
 	defer session.close()
 
-	headers, err := session.getRangeByHeight(ctx, uint64(from.Height)+1, amount)
+	headers, err := session.getRangeByHeight(ctx, uint64(from.Height)+1, amount, 3)
```
vgonkivs, line 163:
```suggestion
	headers, err := session.getRangeByHeight(ctx, uint64(from.Height)+1, amount, ex.Params.MaxHeadersPerRequest)
```

vgonkivs (changes requested):
sorry, it's my fault 😢

vgonkivs (approved):

renaynay (approved):

PR #1426, opened 2022-12-02
feat(api/client): Implement RPC client authentication

### LINKED ISSUE #1187: add authorization

renaynay:
Resolves #1187

* Adds `NewClientWithPerms` which takes a token and adds it to the request header for rpc calls
* Adds tests for authenticated methods

### DIFF
--- a/api/rpc/perms/permissions.go
+++ b/api/rpc/perms/permissions.go
@@ -1,4 +1,4 @@
-package permissions
+package perms
 
 import (
 	"encoding/json"
--- a/api/rpc/perms/permissions.go
+++ b/api/rpc/perms/permissions.go
@@ -9,6 +9,7 @@ import (
 
 var (
 	DefaultPerms   = []auth.Permission{"public"}
+	ReadPerms      = []auth.Permission{"public", "read"}
 	ReadWritePerms = []auth.Permission{"public", "read", "write"}
 	AllPerms       = []auth.Permission{"public", "read", "write", "admin"}
 )
--- a/api/rpc_test.go
+++ b/api/rpc_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"reflect"
+	"strconv"
 	"testing"
 	"time"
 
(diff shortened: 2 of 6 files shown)


api/rpc_test.go
```diff
@@ -110,6 +115,188 @@ func TestModulesImplementFullAPI(t *testing.T) {
 	}
 }
 
+// TestAuthedRPC tests whether a client with admin permissions can access all
+// methods at all permission levels via RPC.
+func TestAuthedRPC(t *testing.T) {
```
distractedm1nd, line 120: Can you rewrite the test cases using multiple `t.Run()`s? Think it would be more readable

distractedm1nd (changes requested):

api/rpc_test.go
```diff
@@ -110,6 +115,188 @@ func TestModulesImplementFullAPI(t *testing.T) {
...
+
+	nd, server := setupNodeWithAuthedRPC(t, signer)
+	url := nd.RPCServer.ListenAddr()
+
+	// we need to run this a few times to prevent the race where the server is not yet started
+	var rpcClient *client.Client
+	require.NoError(t, err)
+	for i := 0; i < 3; i++ {
+		time.Sleep(time.Second * 1)
+		rpcClient, err = client.NewClient(ctx, "http://"+url, string(token))
```
walldiss, line 138: I know it is just a test, but do we plan to use `tls` to protect token from MITM for rpc?

api/rpc/perms/permissions.go
```diff
@@ -0,0 +1,36 @@
+package perms
```
Wondertan, line 1: Weird again that it's not moved. Were there any changes to it?

api/rpc_test.go
Wondertan: This seems to be a test where TLS is not needed, but the concern is totally valid and we `must` use TLS, so that we don't transfer access token over plaintext

Wondertan (approved):
Changes LGTM
utACK for tests

api/rpc/perms/permissions.go
renaynay: I didn't just rename the file, i also added in `ReadPerms`

walldiss (approved):

distractedm1nd (approved):

PR #1429, opened 2022-12-04
feat(cmd/celestia): Add `auth` subcommand to bridge, full and light

### LINKED ISSUE #1187: add authorization

renaynay:
Resolves #1187

This PR adds an `auth` subcommand to `bridge`, `full` and `light` nodes. 

Intended use: after a node is already initialized and started, a user can use the CLI to generate a signed token with some given perms via the `auth` cmd.

Example:

```
celestia bridge auth admin 
```

The above command outputs a signed JWT token with admin permissions that a user can use with `client.NewClient()` to access admin-level methods on the node.

Reference [this commit](https://github.com/celestiaorg/celestia-node/pull/1429/commits/2c833bca89790ff5503822875751f3330207b87f) for the actual diff.

### DIFF
--- a/cmd/celestia/bridge.go
+++ b/cmd/celestia/bridge.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 
 	cmdnode "github.com/celestiaorg/celestia-node/cmd"
 	"github.com/celestiaorg/celestia-node/nodebuilder/core"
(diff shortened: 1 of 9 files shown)


distractedm1nd (dismissed):

cmd/celestia/bridge.go
```diff
@@ -35,6 +35,15 @@ func init() {
...
 		),
+		authCmd(
+			cmdnode.NodeFlags(),
+			p2p.Flags(),
+			core.Flags(),
+			cmdnode.MiscFlags(),
+			rpc.Flags(),
+			gateway.Flags(),
+			state.Flags(),
+		),
```
Wondertan, line 46: * Why authCmd needs all these flags? It seems like it doesn't
* AuthCmd should be located in `cmd` not `cmd/celestia`

Wondertan (changes requested):
There is a bunch of places where we say permissions in plural while in fact, there is only once permission string. See cmd description, variables, and error messages.

cmd/celestia/bridge.go
renaynay: It needs to be able to parse node store flags, p2p flags, etc. Also persistent pre run on bridge, light and full cmds.

distractedm1nd (approved):
excellent

Wondertan (approved):

liamsi (approved):
Only reviewed the tiny change in the ADR which LGTM

PR #1431, opened 2022-12-05
feat(shrex/eds): `Client`/`Server`

### LINKED ISSUE #1416: eds: `Client`
## Context

The EDS Client implements the client side of the [`ShrEx/EDS`](https://github.com/celestiaorg/celestia-node/blob/hlib/adr-blocksyncadr-networking/docs/adr/adr-013-blocksync-overhaul-part-2.md#shrexeds-protocol) protocol.

### `Client.RequestEDS`

```go
// RequestEDS requests the full EDS from one of the given peers.
//
// The peers are requested in a round-robin manner with retries until one of them gives a valid response.
// Blocks forever until the context is canceled or a valid response is given.
func (c *Client) RequestEDS(context.Context, share.Root, peer.IDSlice) (rsmt2d.EDS, error)
```

### LINKED ISSUE #1417: eds: `Server`
## Context

The EDS `Server` implements server side of [`ShrEx/EDS`](https://github.com/celestiaorg/celestia-node/blob/hlib/adr-blocksyncadr-networking/docs/adr/adr-013-blocksync-overhaul-part-2.md#shrexeds-protocol) protocol. It serves `Client`s' `EDSRequest`s, responses with
`EDSResponse` and streams ODS data coming from `eds.Store.GetCAR` wrapped in `ODSReader`.

`Server` may not provide any API, besides constructor and lifecycle methods.

### LINKED ISSUE #1418: modshare: Provide `eds.Client` and `eds.Server` to FX

distractedm1nd:
## Overview
Closes #1416, Closes #1417, Closes #1418

Open questions from me:
- [x] Should this live somewhere else? What is the envisoned directory structure for the three protocols?
- [x] When should the server mark a request as invalid?
- [x] Should a client ever exclude a peer during round robin (for example, after receiving an invalid status or err)
- [x] What other test cases are missing?

## Checklist


share/eds/p2p/client.go
```diff
@@ -0,0 +1,143 @@
...
+				eds, err := c.doRequest(ctx, req, root, to)
+				if err != nil {
+					// TODO: should we exclude the peer from retries if we get an error?
+					log.Errorw("client: eds request to peer failed", "peer", to, "hash", root.String())
+				}
+
+				edsCh <- eds
+			}
+		}
+	}()
```
distractedm1nd, line 69: wtf is going on with indents here? does new goland not use tabs?

share/eds/p2p/client.go
```diff
@@ -0,0 +1,143 @@
...
+			for _, to := range peers {
+				eds, err := c.doRequest(ctx, req, root, to)
+				if err != nil {
+					// TODO: should we exclude the peer from retries if we get an error?
+					log.Errorw("client: eds request to peer failed", "peer", to, "hash", root.String())
+				}
+
+				edsCh <- eds
+			}
+		}
```
vgonkivs, line 68: You should exit from this routine once the correct eds is received and cancel all other requests.

share/eds/p2p/client.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	if err = stream.SetWriteDeadline(time.Now().Add(writeDeadline)); err != nil {
+		log.Warn(err)
+	}
+	_, err = serde.Write(stream, req)
+	if err != nil {
+		stream.Reset() //nolint:errcheck
+		return nil, fmt.Errorf("failed to write request to stream: %w", err)
+	}
```
vgonkivs, line 103: nit: can you please make it consistent: `if err = ...` or `err = ...`

share/eds/p2p/server.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/celestiaorg/celestia-node/share/eds"
+	p2p_pb "github.com/celestiaorg/celestia-node/share/eds/p2p/pb"
+	"github.com/celestiaorg/go-libp2p-messenger/serde"
+)
+
+const (
+	// writeDeadline sets timeout for sending messages to the stream
+	writeDeadline = time.Second * 5
+	// readDeadline sets timeout for reading messages from the stream
+	readDeadline = time.Minute
```
vgonkivs, line 22: should make them configurable.

share/eds/p2p/server.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	// inform the client of our status
+	if err = s.writeStatus(status, stream); err != nil {
+		log.Errorw("server: writing status to stream", "err", err)
+		return
+	}
+	// if we cannot serve the EDS, we are done
+	if status != p2p_pb.Status_OK {
+		stream.Close()
```
vgonkivs, line 85:
```suggestion
		stream.Reset()
```

vgonkivs (changes requested):

share/eds/p2p/server.go
```diff
@@ -0,0 +1,154 @@
...
+		return
+	}
+
+	// start streaming the ODS to the client
+	if err = s.writeODS(edsReader, stream); err != nil {
+		log.Errorw("server: writing ods to stream", "err", err)
+		return
+	}
```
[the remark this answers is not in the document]
distractedm1nd: the stream gets reset in writeODS if an err occurred, should i remove the stream reset from the subroutine and then do it here?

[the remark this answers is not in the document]
vgonkivs: oh, sorry. Did not notice. I'd keep all resets in one method.

share/eds/p2p/server.go
```diff
@@ -0,0 +1,154 @@
...
+		status = p2p_pb.Status_NOT_FOUND
+	} else {
+		status = p2p_pb.Status_OK
+	}
+
+	// inform the client of our status
+	if err = s.writeStatus(status, stream); err != nil {
+		log.Errorw("server: writing status to stream", "err", err)
+		return
+	}
```
vgonkivs, line 82: The same here.

share/eds/p2p/server.go
distractedm1nd: discussed in sync, should be close

share/eds/p2p/server.go
distractedm1nd: done

share/eds/p2p/server.go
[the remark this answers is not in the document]
distractedm1nd: done

share/eds/p2p/client.go
distractedm1nd: done

share/eds/p2p/client.go
distractedm1nd: done, but can you check if I did it correctly?

share/eds/p2p/server.go
```diff
@@ -0,0 +1,163 @@
...
+	resp := &p2p_pb.EDSResponse{Status: status}
+	_, err = serde.Write(stream, resp)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (s *Server) writeODS(edsReader io.ReadCloser, stream network.Stream) error {
```
Wondertan, line 144: We should stream the data instead of sending it all as one message. Streaming implies sending the data in portions..The simplest way of doing so is to use `io.Copy`, which would read data out of ODSReader and write it into the stream.

Wondertan (changes requested):

share/eds/p2p/server.go
distractedm1nd: done. thanks for pointing me to io.Copy, that's really cool

share/eds/p2p/pb/extended_data_square.proto
```diff
@@ -0,0 +1,16 @@
...
+enum Status {
+  INVALID = 0;
+  OK = 100; // data found
+  NOT_FOUND = 200; // data not found
+  REFUSED = 201; // request refused
+}
+
+message EDSResponse {
+  Status status = 1;
+}
```
Wondertan, line 16: new line

share/eds/p2p/client.go
```diff
@@ -0,0 +1,161 @@
...
+	p2p_pb "github.com/celestiaorg/celestia-node/share/eds/p2p/pb"
+
+	"github.com/celestiaorg/rsmt2d"
+)
+
+var log = logging.Logger("shrex/eds")
+
+var errNoMorePeers = errors.New("all peers returned invalid responses")
+
+// TODO(@distractedm1nd): make version suffix configurable
```
Wondertan, line 29: We should add the suffix in this PR to minimize tech debt. Also, If we start implementing new patterns on existing code, the new code should integrate those while it developed, not later.

share/eds/p2p/client.go
```diff
@@ -0,0 +1,161 @@
...
+	if err != nil {
+		log.Warn(err)
+	}
+	_, err = serde.Write(stream, req)
+	if err != nil {
+		stream.Reset() //nolint:errcheck
+		return nil, fmt.Errorf("failed to write request to stream: %w", err)
+	}
+	err = stream.CloseWrite()
+	if err != nil {
```
Wondertan, line 122: Reset as well.

share/eds/p2p/client.go
```diff
@@ -0,0 +1,161 @@
...
+
+	switch resp.Status {
+	case p2p_pb.Status_OK:
+		// read all CARv1 header and ODS bytes from stream
+		carBytes, err := io.ReadAll(stream)
+		if err != nil && err != io.EOF {
+			return nil, fmt.Errorf("unexpected error while reading ods from stream: %w", err)
+		}
+		// use header and ODS bytes to construct EDS and verify it against dataHash
+		eds, err := eds.ReadEDS(ctx, bytes.NewReader(carBytes), dataHash)
```
Wondertan, line 146: We can give the stream directly to the ReadEDS.

share/eds/p2p/client.go
```diff
@@ -0,0 +1,161 @@
...
+	to peer.ID,
+) (*rsmt2d.ExtendedDataSquare, error) {
+	log.Debugf("client: requesting eds %s from peer %s", dataHash.String(), to)
+	stream, err := c.host.NewStream(ctx, to, c.protocolID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open stream: %w", err)
+	}
+
+	// request status from peer to provide the requested EDS
+	err = stream.SetWriteDeadline(time.Now().Add(writeDeadline))
```
Wondertan, line 112: To follow what we've discussed in HeaderEx, the deadlines should be set from the context. Otherwise context cancellation does not take any effect and blocks forever

Wondertan (commented):
First look, now going for lunch

share/eds/p2p/client.go
distractedm1nd: sick, dont know why i didnt think about trying that

share/eds/p2p/server.go
Wondertan: In this PR. The reasoning is the same for including protocolSuffix earlier.

share/eds/p2p/client.go
```diff
@@ -0,0 +1,143 @@
...
+				continue
+			}
+
+			eds, err := c.doRequest(ctx, req, dataHash, to)
+			if eds != nil {
+				return eds, nil
+			}
+			if errors.Is(err, context.DeadlineExceeded) || errors.Is(err, context.Canceled) {
+				return nil, ctx.Err()
+			}
```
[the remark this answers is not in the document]
Wondertan: Also cc @walldiss, for the stream helper you are working on.

share/eds/p2p/server.go
```diff
@@ -0,0 +1,163 @@
...
+		log.Errorw("server: reading request from stream", "err", err)
+		stream.Reset() //nolint:errcheck
+		return
+	}
+
+	var edsReader io.ReadCloser
+	status := p2p_pb.Status_OK
+	err = share.DataHash(req.Hash).Validate()
+	if err != nil {
+		status = p2p_pb.Status_INVALID
```
Wondertan, line 74: Status Invalid does not mean to be used in practice. It's a 0-value code and is a PB convention to prevent uninitialized enums from being sent over the wire. 

If the data hash is invalid(it's size is less than 32) it means we have a malicious client, and we should just reset the stream without caring about responding to anything. 

Will add this to the document

share/eds/p2p/server.go
```diff
@@ -0,0 +1,163 @@
...
+	} else {
+		// determine whether the EDS is available in our store
+		edsReader, err = s.store.GetCAR(s.ctx, req.Hash)
+		if err != nil {
+			status = p2p_pb.Status_NOT_FOUND
+		} else {
+			defer edsReader.Close()
+		}
+	}
```
Wondertan, line 83: nit: too much nesting.

share/eds/p2p/server.go
```diff
@@ -0,0 +1,163 @@
...
+		log.Warn(err)
+	}
+
+	resp := &p2p_pb.EDSResponse{Status: status}
+	_, err = serde.Write(stream, resp)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
Wondertan, line 144:
```suggestion
	return err
```

share/eds/p2p/client.go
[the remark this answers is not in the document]
distractedm1nd: This error does happen, and is covered in two tests

share/eds/p2p/client.go
```diff
@@ -0,0 +1,143 @@
...
+	err = stream.CloseWrite()
+	if err != nil {
+		stream.Reset() //nolint:errcheck
+		return nil, fmt.Errorf("failed to close write on stream: %w", err)
+	}
+
+	// read and parse status from peer
+	resp := new(p2p_pb.EDSResponse)
+	_, err = serde.Read(stream, resp)
+	if err != nil && err != io.EOF {
```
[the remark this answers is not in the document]
distractedm1nd: Thanks for the context

share/eds/p2p/server.go
distractedm1nd: done

share/eds/p2p/server.go
distractedm1nd: done

share/eds/p2p/server.go
distractedm1nd: done (resets stream now)

share/eds/p2p/client.go
```diff
@@ -0,0 +1,143 @@
...
+	case p2p_pb.Status_INVALID:
+		return nil, fmt.Errorf("request for root %s marked as invalid by peer", dataHash.String())
+	default:
+		return nil, fmt.Errorf("peer responded with unknown status %s", resp.GetStatus())
+	}
+}
+
+func protocolID(protocolSuffix string) protocol.ID {
+	return protocol.ID(fmt.Sprintf("/shrex/eds/v0.0.1/%s", protocolSuffix))
+}
```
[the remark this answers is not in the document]
distractedm1nd: put at top of file

share/eds/p2p/client.go
```diff
@@ -0,0 +1,143 @@
...
+
+		return eds, nil
+	case p2p_pb.Status_NOT_FOUND, p2p_pb.Status_REFUSED:
+		log.Debugf("client: peer %s couldn't serve eds %s with status %s", to.String(), dataHash.String(), resp.GetStatus())
+		// no eds was returned, but the request was valid and should be retried
+		return nil, nil
+	case p2p_pb.Status_INVALID:
+		return nil, fmt.Errorf("request for root %s marked as invalid by peer", dataHash.String())
+	default:
+		return nil, fmt.Errorf("peer responded with unknown status %s", resp.GetStatus())
```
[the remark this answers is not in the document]
distractedm1nd: done, we fallthrough now

share/eds/p2p/client.go
```diff
@@ -0,0 +1,143 @@
...
+
+			// no eds was found, continue
+		}
+	}
+}
+
+func (c *Client) doRequest(
+	ctx context.Context,
+	req *p2p_pb.EDSRequest,
+	dataHash share.DataHash,
```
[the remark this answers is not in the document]
distractedm1nd: good idea, done

share/eds/p2p/client.go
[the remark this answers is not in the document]
distractedm1nd: done

share/eds/p2p/client.go
distractedm1nd: done, but server side still uses durations yeah? Since there is no context on handleStream

share/eds/p2p/client.go
```diff
@@ -0,0 +1,161 @@
...
+) (*rsmt2d.ExtendedDataSquare, error) {
+	req := &p2p_pb.EDSRequest{Hash: dataHash}
+
+	// requests are retried for every peer until a valid response is received
+	excludedPeers := make(map[peer.ID]struct{})
+	for {
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		default:
```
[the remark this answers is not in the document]
renaynay: this is probs outdat4ed @distractedm1nd

share/eds/p2p/client.go
```diff
@@ -132,12 +135,8 @@ func (c *Client) doRequest(
 		// no eds was returned, but the request was valid and should be retried
 		return nil, nil
 	case p2p_pb.Status_INVALID:
-		return nil, fmt.Errorf("request for root %s marked as invalid by peer", dataHash.String())
+		fallthrough
 	default:
-		return nil, fmt.Errorf("peer responded with unknown status %s", resp.GetStatus())
+		return nil, fmt.Errorf("request status for root %s marked as invalid", dataHash.String())
```
renaynay, line 140: can you still add details re the resp.GetStatus as well?

share/eds/p2p/exchange_test.go
```diff
@@ -0,0 +1,110 @@
...
+			err = store.Put(ctx, dah.Hash(), eds)
+			// require.NoError(t, err)
+		}()
+
+		now := time.Now()
+		requestedEDS, err := client.RequestEDS(ctx, dah.Hash(), []peer.ID{server.host.ID()})
+		finished := time.Now()
+
+		assert.Greater(t, finished.Sub(now), storageDelay)
+		assert.Nil(t, err)
```
renaynay, line 61: assert.NoError assuming you're testing to ensure that there is no error.

share/eds/p2p/server.go
Wondertan: It's not a bloat. It's a tech debt that we should cover sooner rather than later. 

If we ever stick to some pattern, every new component should follow it. If we introduce a new feature and we know how we want to configure it, we should introduce it with configuration.

Originally, we didn't do it and have to spend some time to catch up everywhere. We really don't want to pile up more things.

I am fine though to extract this into separate PR, which should be part of the epic and come right after.

share/eds/p2p/client.go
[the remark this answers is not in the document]
Wondertan: I see. When I wrote this I thought specifically about Read and Write which do not return context errors. The error is returned by either NewStream or ReadEDS inside doRequest. 

The problem with current code is that it does not cover for errors coming once deadline is reached on blocking Read or Write. The error returned there is net.Error and we should swap it out with context error and return. Otherwise, we go on the next try even when the operation is cancelled.

share/eds/p2p/client.go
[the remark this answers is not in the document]
distractedm1nd: this is outdated

share/eds/p2p/client.go
```diff
@@ -0,0 +1,161 @@
...
+
+var log = logging.Logger("shrex/eds")
+
+var errNoMorePeers = errors.New("all peers returned invalid responses")
+
+// TODO(@distractedm1nd): make version suffix configurable
+var protocolID = protocol.ID("/shrex/eds/0.0.1")
+
+// Client is responsible for requesting EDSs for blocksync over the ShrEx/EDS protocol.
+// This client is run by light nodes and full nodes. For more information, see ADR #13
```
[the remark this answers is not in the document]
distractedm1nd: According to the ADR, yes
> ShrEx/EDS is a pull-based protocol with a client-server model, where LNs are clients, BNs are servers, and FNs are both. 
The protocol has only one request-response interaction followed by the ODS stream, which then reconstructs into EDS.

But its more of an optional thing afaik. This godoc is probably misleading for readers trying to understand the network

share/eds/p2p/server.go
distractedm1nd: Let's open a discussion about making read and write deadlines configurable. There is absolutely no reason for it, and if we only want to do it for consistency then we need to discuss if it even makes sense in general. Won't happen in this PR, but looking forward to figuring out what's best for the whole codebase, and going from there

share/eds/p2p/exchange_test.go
distractedm1nd: done

share/eds/p2p/client.go
distractedm1nd: done

share/eds/p2p/server.go
```diff
@@ -0,0 +1,160 @@
...
+	// ensure the requested dataHash is a valid root
+	err = share.DataHash(req.Hash).Validate()
+	if err != nil {
+		stream.Reset() //nolint:errcheck
+		return
+	}
+
+	status := p2p_pb.Status_OK
+	// determine whether the EDS is available in our store
+	edsReader, err := s.store.GetCAR(s.ctx, req.Hash)
```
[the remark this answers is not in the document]
distractedm1nd: done

share/eds/p2p/server.go
```diff
@@ -0,0 +1,160 @@
...
+
+	store *eds.Store
+}
+
+// NewServer creates a new ShrEx/EDS server.
+func NewServer(host host.Host, store *eds.Store, protocolSuffix string) *Server {
+	return &Server{
+		host:       host,
+		protocolID: protocolID(protocolSuffix),
+		store:      store,
```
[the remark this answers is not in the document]
distractedm1nd: done, nice catch

share/eds/p2p/server.go
Wondertan: It's not just about read/write and also about protocolPrefix. We might also have rate-limiting further with configurations, which can simply add a new config param instead of implementing the whole thing. The question is not about read or write, but "does server need configuration or not?" If so, the feature should include it

share/eds/p2p/client.go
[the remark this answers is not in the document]
renaynay: Yes, light nodes wouldn't really run this by "default" - it's not the default behaviour for LNs to request EDSs -- AFAIU it's on a case-by-case basis ? cc @Wondertan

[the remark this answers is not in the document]
Wondertan: Yes. All our node types follow the same API and LN will be able to request the EDS to comply with it. Don't see why we should mention this in godoc. It's highly irrelevant imo

share/eds/p2p/client.go
Wondertan: Yes, server uses durations.

share/eds/p2p/client.go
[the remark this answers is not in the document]
vgonkivs: @Wondertan, if you meant [this one ](https://github.com/celestiaorg/celestia-node/blob/3a58679ed84da966d01173f32780134c7b830594/header/p2p/session.go#L113), then it's ok and we should not return err there.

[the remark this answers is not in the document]
Wondertan: No, not this one @vgonkivs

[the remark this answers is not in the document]
Wondertan: Yeah, check how @distractedm1nd made it work https://github.com/celestiaorg/celestia-node/pull/1431/files#diff-677f2124709be3e4dbc7f0df73e8e4914f4d08cc80869fec9c55a74071fcd7d6R77-R80

Wondertan (changes requested):

nodebuilder/share/module.go
```diff
@@ -48,6 +50,23 @@ func ConstructModule(tp node.Type, cfg *Config, options ...fx.Option) fx.Option
 		return fx.Module(
 			"share",
 			baseComponents,
+			fx.Provide(fx.Annotate(
+				func(host host.Host, store *eds.Store, path node.StorePath) *p2p.Server {
```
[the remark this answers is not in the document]
distractedm1nd: OOF dont know how that happened, great catch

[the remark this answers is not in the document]
distractedm1nd: resolved

share/eds/p2p/client.go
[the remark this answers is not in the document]
vgonkivs: so, [this approach](https://github.com/celestiaorg/celestia-node/pull/1431/files#diff-677f2124709be3e4dbc7f0df73e8e4914f4d08cc80869fec9c55a74071fcd7d6R77-R80) is basically to handle mocknet errors?

share/eds/p2p/server.go
```diff
@@ -0,0 +1,161 @@
...
+		log.Error(err)
+	}
+
+	return req, nil
+}
+
+func (s *Server) writeStatus(status p2p_pb.Status, stream network.Stream) error {
+	err := stream.SetWriteDeadline(time.Now().Add(writeDeadline))
+	if err != nil {
+		log.Warn(err)
```
vgonkivs, line 137:
```suggestion
		log.Debug(err)
```

share/eds/p2p/server.go
```diff
@@ -0,0 +1,161 @@
...
+
+	resp := &p2p_pb.EDSResponse{Status: status}
+	_, err = serde.Write(stream, resp)
+	return err
+}
+
+func (s *Server) writeODS(edsReader io.ReadCloser, stream network.Stream) error {
+	err := stream.SetWriteDeadline(time.Now().Add(writeDeadline))
+	if err != nil {
+		log.Warn(err)
```
vgonkivs, line 148:
```suggestion
		log.Debug(err)
```

share/eds/p2p/server.go
renaynay: `protocolPrefix` is a very different granularity than read/write deadlines...

share/eds/p2p/server.go
```diff
@@ -0,0 +1,161 @@
...
+	// read request from stream to get the dataHash for store lookup
+	req, err := s.readRequest(stream)
+	if err != nil {
+		log.Errorw("server: reading request from stream", "err", err)
+		stream.Reset() //nolint:errcheck
+		return
+	}
+
+	// ensure the requested dataHash is a valid root
+	err = share.DataHash(req.Hash).Validate()
```
renaynay, line 70: i'd cast into a datahash, call validate, and then pass as a datahash to the getcar method

like: 
hash := share.DataHash(req.Hash)
err := hash.Validate()
if err blah blah

GetCar(..., hash, ..)

share/eds/p2p/client.go
```diff
@@ -0,0 +1,147 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to open stream: %w", err)
+	}
+	if dl, ok := ctx.Deadline(); ok {
+		if err = stream.SetDeadline(dl); err != nil {
+			log.Debugw("error setting deadline: %s", err)
+		}
+	}
+
+	// request status from peer to provide the requested EDS
```
renaynay, line 109:
```suggestion
	// request ODS
```

share/eds/p2p/server.go
```diff
@@ -0,0 +1,161 @@
...
+	if err != nil {
+		log.Errorw("server: writing ods to stream", "err", err)
+		stream.Reset() //nolint:errcheck
+		return
+	}
+
+	err = stream.Close()
+	if err != nil {
+		log.Errorw("server: closing stream", "err", err)
+		return
```
renaynay, line 111:
```suggestion
```

renaynay (commented):
really nice :)

share/eds/p2p/client.go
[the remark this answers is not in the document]
Wondertan: Not only, also for the yamux and quic errors in production when context is canceled by a deadline

share/eds/p2p/server.go
Wondertan: Explain

Wondertan: @renaynay, explain

share/eds/p2p/server.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+}
+
+func (s *Server) readRequest(stream network.Stream) (*p2p_pb.EDSRequest, error) {
+	err := stream.SetReadDeadline(time.Now().Add(readDeadline))
+	if err != nil {
+		log.Warn(err)
```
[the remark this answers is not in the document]
Wondertan: +1

nodebuilder/share/module.go
```diff
@@ -48,6 +51,23 @@ func ConstructModule(tp node.Type, cfg *Config, options ...fx.Option) fx.Option
...
+					return server.Start(ctx)
+				}),
+				fx.OnStop(func(ctx context.Context, server *p2p.Server) error {
+					return server.Stop(ctx)
+				}),
+			)),
+			// Bridge Nodes need a client as well, for requests over FullAvailability
+			fx.Provide(
+				func(host host.Host, network modp2p.Network) *p2p.Client {
+					return p2p.NewClient(host, string(network))
```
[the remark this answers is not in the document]
Wondertan: That's not the case. The client is not an interface and won't be provided for the BN at all.

share/eds/p2p/client.go
```diff
@@ -0,0 +1,147 @@
...
+		for _, to := range peers {
+			// skip over excluded peers
+			if _, ok := excludedPeers[to]; ok {
+				continue
+			}
+			eds, err := c.doRequest(ctx, req, to)
+			if eds != nil {
+				return eds, nil
+			}
+			if errors.Is(err, context.DeadlineExceeded) || errors.Is(err, context.Canceled) {
```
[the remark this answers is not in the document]
distractedm1nd: discussed in sync, decided against

share/eds/p2p/server.go
distractedm1nd: Let's add the complexity that comes with more fine-grained configuration first once we have things that make sense to configure. The configurability of the protocolID matches the rest of our codebase right now.

renaynay: @Wondertan  - protcolPrefix configures who the node can talk to / who talks to the node on that specific protocol (not granular, it is about distinguishing this specific protocol from others) whereas read/write deadlines is a very granular configuration option on the stream level.

We should allow for customisability to the extent that it matters (protocolIDs) but (until we have a VERY good reason for it) not for params as granular as reading/writing to the stream. The defaults are good enough and if someone really wants to toy with the granularity, they can change it in the code. I am against bloat as much as possible unless there's a properly compelling reason for it and "does server need configuration or not?" argument is too generic bc configuration is a spectrum. 

Anyway no need to waste more time discussing this, I hope we can all agree on a general philosophy about this and not have to discuss it again.

share/eds/p2p/client.go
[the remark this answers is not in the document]
renaynay: LN can request EDS on a case by case basis though so yeah lets remove from comment and only mention in a doc.go eventually or point at ADR.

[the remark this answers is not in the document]
distractedm1nd: removed from comment already

share/eds/p2p/client.go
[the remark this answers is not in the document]
vgonkivs: Ok, will add the same there. Thanks

vgonkivs (approved):

share/eds/p2p/server.go
```diff
@@ -0,0 +1,161 @@
...
+	err := stream.SetWriteDeadline(time.Now().Add(writeDeadline))
+	if err != nil {
+		log.Debug(err)
+	}
+
+	odsReader, err := eds.ODSReader(edsReader)
+	if err != nil {
+		return fmt.Errorf("creating ODS reader: %w", err)
+	}
+	_, err = io.Copy(stream, odsReader)
```
Wondertan, line 155: Copy internally creates a buffer with default size. Just realized that we should have more control over it and use CopyBuffer instead, which allows setting a custom buffer and size. We can go with 32kb as default, which we should allow to configure

Wondertan (changes requested):
I also request to move files according to the ADR, just in `eds` pkg. I promise we will restructure them later. For now, it's not really clear how. Because we can either have the sub` p2p` pkg in eds or `share` or we can have pkg names after protocol e.g. `shrex` We will discuss this altogether

share/eds/p2p/server.go
Wondertan: We should discuss this to the point when all of us agree. At this point, I don't see a configuration being a bloat. Configuration is something that each new core feature should have when it's implemented. However, I don't want to block this PR anymore and force anyone. 

Another performance-related configuration that we _have to_ expose eventually: https://github.com/celestiaorg/celestia-node/pull/1431#discussion_r1053143556

share/eds/p2p/server.go
distractedm1nd: Agree that we should have a const, adding now

Disagree it should be configurable for the same reason the timeouts should not be configurable.

share/eds/p2p/server.go
renaynay: > We should discuss this to the point when all of us agree. 

My argument is that configurability exists on a spectrum and movement towards extreme customisation should be justified as it adds additional LOCs. I see justification as 1. several users ask for this, or 2. the param could be a **critical** testing variable.

> At this point, I don't see a configuration being a bloat. 

I disagree -- each additional customisable parameter introduces code bloat ( **objectively** ) in LOCs.

> Configuration is something that each new core feature should have when it's implemented

Also disagree - a feature could be implemented with defaults and tested for correctness, and *later* should come optimisations via greater configurability. The idea that every feature should be implemented from the start with configurability slows down dev iteration process and is just unnecessary.

distractedm1nd: KISS, YAGNI is all I have to say here. Implement it when its needed.

share/eds/p2p/server.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	// ensure the requested dataHash is a valid root
+	err = share.DataHash(req.Hash).Validate()
+	if err != nil {
+		stream.Reset() //nolint:errcheck
+		return
+	}
+
+	ctx, cancel := context.WithTimeout(s.ctx, readDeadline)
```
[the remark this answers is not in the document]
Wondertan: Just saw how @walldiss called this timeout https://github.com/celestiaorg/celestia-node/pull/1497/files#diff-80f2895dde00e80d0edb947fefa69adb94d87556eca76e348c78fc335e239570R32, like the name

share/eds/p2p/server.go
Wondertan: All of this will be configurable eventually 😉

distractedm1nd: Hope not.

share/eds/p2p/server.go
distractedm1nd: implemented params

share/eds/p2p/server.go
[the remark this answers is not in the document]
distractedm1nd: Like that name too, but doesn't make the same sense here - went for `ReadCARDeadline` but open to another name

Wondertan (commented):
Last nits

Wondertan (approved):
Very nice. In Russian, we say "reference work", which means perfect, and everyone should take it as a reference.

share/eds/p2p/server.go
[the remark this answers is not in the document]
Wondertan: Ok

share/p2p/shrexeds/options.go
```diff
@@ -0,0 +1,74 @@
...
+
+// WithProtocolSuffix is a functional option that configures the `protocolSuffix` parameter
+func WithProtocolSuffix(protocolSuffix string) Option {
+	return func(parameters *Parameters) {
+		parameters.protocolSuffix = protocolSuffix
+	}
+}
+
+func protocolID(protocolSuffix string) protocol.ID {
+	return protocol.ID(fmt.Sprintf("%s%s", protocolPrefix, protocolSuffix))
```
Wondertan, line 73: hmm, our network ID does not have '/', so this will be concatenated

distractedm1nd: protocolPrefix does have the /

renaynay (approved):

PR #1434, opened 2022-12-05
chore!(nodebuilder/p2p): Upgrade `mamaki` to `mocha`

renaynay:
Upgrades `mamaki` to `mocha`

TODO 
- [x] genesis hash

### DIFF
--- a/nodebuilder/p2p/network.go
+++ b/nodebuilder/p2p/network.go
@@ -13,8 +13,8 @@ const (
 	DefaultNetwork = Arabica
 	// Arabica testnet. See: celestiaorg/networks.
 	Arabica Network = "arabica-2"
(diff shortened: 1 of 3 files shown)


nodebuilder/p2p/network.go
```diff
@@ -13,8 +13,8 @@ const (
 	DefaultNetwork = Arabica
```
Wondertan, line 13: We might want to make Mocha the default one.

Wondertan (approved):

nodebuilder/p2p/network.go
Wondertan: Let's change this here @renaynay

distractedm1nd (approved):

Issue #1419, opened 2022-12-01
eds: `NDClient`


Wondertan:
## Context

The EDS Client implements the client side of the [`ShrEx/NS`](https://github.com/celestiaorg/celestia-node/blob/hlib/adr-blocksyncadr-networking/docs/adr/adr-013-blocksync-overhaul-part-2.md#shrexnd-protocol) protocol.

### `Client.RequestND`

```go
// RequestND requests namespaced data from one of the given peers.
//
// The peers are requested in a round-robin manner with retries until one of them gives a valid response.
// Blocks forever until the context is canceled and/or a valid response is given.
//
// Returns only valid data with verified inclusion against share.Root.
func (c *Client) RequestND(context.Context, share.Root, peer.IDSlice, namespace.ID) ([][]byte, error)
```

PR #1437, opened 2022-12-06
feat(ShrEx/ND):  Create client/server for namespaced data communication

walldiss:
## Overview

The purposed changes abstracts p2p server / client level logic from handler / requestor implementation. 
New approach allows to reuse common code components across multiple libp2p protocols. Such components could be:
 - Interceptors:
    - logging (implemented in PR as example)
    - metrics
    - tracing injection (imagining being able to visualise requests graph!)
    - hello checks
    - recovery
...

 - single place for connection management logic such as connection reset and timeouts.
 
## TODO
- [ ] fix lint
- [ ] add tests for middleware
- [ ] add tests for error resulting calls
## Checklist


Wondertan (changes requested):
Hmm, we don't need to track peers like HeaderEx is doing. The main reason is that not every peer we connect with can serve us data, e.g., connManager can return you an LN for an LN, and LNs do not store and serve the data we request.

Instead, we rely on the discovery component that currently lives in `share/discovery`. It will give us peers to request data from. The client should not track them and just request that were given to the client as a param. Pls see https://github.com/celestiaorg/celestia-node/issues/1419#issue-1471534910

walldiss:
Thank you for pointing this out. The connManager could be split into two components: `discovery` and `balancer`. First one should establish and maintain connections. `balancer` should take ready connection from `discovery` and picks one(or many) for communication based on arbitrary logic.

Wondertan (changes requested):
I've been wonderting about the p2p pkg and components in them since yesterday. Additionally, we thoroughly discussed it with @renaynay and concluded that this pattern ___produces more code and complexity than it attempts to duplicate and simplify at this point___. 

We totally agree that there is a lot of room for improvements and deduplication of the standard code we also want a cleanup. However, with tight deadlines at this stage, it is clearly one of the last priorities. 

I am requesting changes to this PR to follow our current approach in other protocols where we work with streams directly without Session/Client and Server, with the promise of allocating some time for refactoring for our protocols after incentivised testnet. It would be much cleaner to design standard components after looking at all the protocols and covering all the cases we've had so far at one. Then steadily, migrate all the protocols.

vgonkivs:
I'd suggest to create a different branch with the implementation of Client/Server and to keep this branch as it is. So we can return to it in future.

PR #1439, opened 2022-12-06
refactor(share/eds): Store methods take `DataHash` instead of `share.Root`

### LINKED ISSUE #1428: eds: Use `DataHash` instead of `share.Root` in `Store` methods

renaynay:
Resolves #1428 

Technically breaking API but it's not really used so it doesn't matter :)


share/eds/store.go
```diff
@@ -93,8 +91,13 @@ func (s *Store) Stop() error {
...
 // Additionally, the file gets indexed s.t. store.Blockstore can access them.
-func (s *Store) Put(ctx context.Context, root share.Root, square *rsmt2d.ExtendedDataSquare) error {
-	key := root.String()
+func (s *Store) Put(ctx context.Context, root DataHash, square *rsmt2d.ExtendedDataSquare) error {
+	// sanity check the root
+	if err := root.Validate(); err != nil {
+		return err
+	}
+
+	key := fmt.Sprintf("%X", root)
```
distractedm1nd, line 100: make DataHash implement Stringer w the "%X"

distractedm1nd (changes requested):

share/eds/store.go
renaynay: tru

share/eds/store.go
```diff
@@ -189,7 +193,12 @@ func (s *Store) Remove(ctx context.Context, root share.Root) error {
 //
 // It reads only one quadrant(1/4) of the EDS and verifies the integrity of the stored data by
 // recomputing it.
-func (s *Store) Get(ctx context.Context, root share.Root) (*rsmt2d.ExtendedDataSquare, error) {
+func (s *Store) Get(ctx context.Context, root DataHash) (*rsmt2d.ExtendedDataSquare, error) {
+	// sanity check the root
+	if err := root.Validate(); err != nil {
+		return nil, err
+	}
```
Wondertan, line 200: Why do some of the methods validate it and some do not?

Another team-wide point. We usually don't do sanity checks for params within internal components and only do those from the user or network inputs. So the Q: should we do it here, or should we trust our networking components that it will do the check?

share/eds/hash.go
```diff
@@ -0,0 +1,12 @@
+package eds
+
+import "fmt"
+
+type DataHash []byte
```
Wondertan, line 5: We usually keep such types in the root `share` pkg, and `eds` right now does not hold any core type definitions but only relevant components, so suggesting to move it to `share`. However, the `eds` pkg grows so fast that it might become a root one on its own, but that's a different discussion after ITN refactorings. 

Also go doc.

share/eds/store.go
renaynay: we should have sanity checks at network entrypoints.

share/eds/hash.go
renaynay: https://github.com/celestiaorg/celestia-node/pull/1439/commits/a3fa93bc0a92fc01141c6fe800e5c39958a34451

share/eds/store.go
renaynay: https://github.com/celestiaorg/celestia-node/pull/1439/commits/1d035d79e0b27f7964ee6971e45301d9fdb638b9

share/share.go
```diff
@@ -39,3 +41,18 @@ func ID(s Share) namespace.ID {
...
+
+func (dh DataHash) Validate() error {
+	if len(dh) != 32 {
+		return fmt.Errorf("invalid hash size, expected 32, got %d", len(dh))
+	}
+	return nil
+}
+
+func (dh DataHash) String() string {
+	c := dh
```
Wondertan, line 56: ?

share/eds/store.go
Wondertan: @renaynay

share/share.go
```diff
@@ -39,3 +41,17 @@ func ID(s Share) namespace.ID {
...
+
+func (dh DataHash) Validate() error {
+	if len(dh) != 32 {
+		return fmt.Errorf("invalid hash size, expected 32, got %d", len(dh))
+	}
+	return nil
+}
+
+func (dh DataHash) String() string {
+	return fmt.Sprintf("%X", string(dh))
```
Wondertan, line 56:
```suggestion
	return fmt.Sprintf("%X", []byte(dh))
```

renaynay: same thing but yes

Wondertan (approved):

PR #1446, opened 2022-12-07
github/CI: Fix over kill of '.gitignore'.

### LINKED ISSUE #1442: github/CI: Make .gitignore file more exact so that no new files get blocked.
Some statements in the file are over kill and need to be fixed. I will create PR very soon.

HoytRen:
Closes #1442 

Limit some pattern to root folder.
Remove duplicated pattern.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-build/
+/build/
 *.bak
 *.iml
 *.log
@@ -10,14 +10,13 @@ build/
 *.coverprofile
 *.test
 *.orig
-*/vendor
 vendor
 .DS_Store
 .bak
 .idea/
 .vscode/
-celestia
-cel-shed
-cel-key
+/celestia
+/cel-shed
+/cel-key
 coverage.txt
 go.work


Wondertan:
@HoytRen, don't fully understand the problem we are solving here. We had a similar PR before https://github.com/celestiaorg/celestia-node/pull/566 btw. By mistake, we added `.` before paths and that didn't work, but it seems like it can work with just a slash. What's the diff between slash and no-slash in this case?

HoytRen:
If there is a separator at the beginning or middle (or both) of the pattern, then the pattern is relative to the directory level of the particular .gitignore file itself. Otherwise the pattern may also match at any level below the .gitignore level.

Wondertan:
Ok. I see. Just for ref, the msg was copied from [here](https://git-scm.com/docs/gitignore#_pattern_format).

.gitignore
```diff
@@ -10,14 +10,13 @@ build/
...
 .DS_Store
 .bak
 .idea/
 .vscode/
-celestia
-cel-shed
-cel-key
+/celestia
+/cel-shed
+/cel-key
```
Wondertan, line 20: Why do we need to track the root dir only(where .gitignore is located)? If someone compiles a binary, not in the root, but somewhere deeper in the path, it should not be included in git history either, imo.

HoytRen: Because the binary file has no extension name, git can't distinguish file and folder. When I add new files to cmd/celestia, git can't track it. If we limit celestia binary in the root folder, I feel it's better to apply to other binaries too, easier for understanding. Each coder may have their own configuration so we can only focus on the default config. They could have their local version when they need it. According to the current default config, the celestia binary isn't in the root folder, but I don't know if there are coders who still prefer it, we don't need to bother them, just provide backward compatibility. For the 'build' folder, I see a lot of projects put scripts about the building processes in this folder, and use some like 'target' as output, and I don't know if we will import code from other projects later, then limiting our build folder in root may get better compatibility.

HoytRen: @Wondertan What do you prefer, change the one caused problem and keep others untouched?

Wondertan: Sounds reasonable.

Wondertan (approved):

renaynay (approved):

renaynay (approved):

HoytRen:
Thank all of you. I'm caught by covid-19 these days, recovering, will get back soon.

PR #1451, opened 2022-12-07
feat(share/discovery): add Peers method

vgonkivs:
## Checklist


share/availability/discovery/discovery.go
```diff
@@ -169,3 +169,18 @@ func (d *Discovery) Advertise(ctx context.Context) {
...
+// If Discovery hasn't found any peers, it blocks until at least one peer is found.
+func (d *Discovery) Peers(ctx context.Context) ([]peer.ID, error) {
+	if d.set.Size() > 0 {
+		return d.set.Peers(), nil
+	}
+
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	case p := <-d.set.waitPeer:
```
Wondertan, line 183: nit: this break encapsulation of the set and putting this logic to the set itself would be cleaner.

Wondertan (approved):
Note for everyone. This design has one problem, it does not allow us to get newly discovered peers in real-time, so protocols relying on it won't get them without retries. We are going this way for simplicity first and can improve with further iterations of the protocol.

share/availability/discovery/discovery.go
vgonkivs: yeah, I know, thanks.

Wondertan (dismissed):

Wondertan (approved):

share/availability/discovery/set.go
```diff
@@ -47,6 +50,11 @@ func (ps *limitedSet) TryAdd(p peer.ID) error {
 	}
 	if len(ps.ps) < int(ps.limit) {
 		ps.ps[p] = struct{}{}
+
+		select {
+		case ps.waitPeer <- p:
+		default:
+		}
```
renaynay, line 57: pls doc here why we don't block on write

share/availability/discovery/set_test.go
```diff
@@ -50,7 +52,32 @@ func TestSet_Peers(t *testing.T) {
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*1)
+	t.Cleanup(cancel)
+
+	peers, err := set.Peers(ctx)
+	require.NoError(t, err)
+	require.True(t, len(peers) == 2)
+}
+
+func TestSet_WaitPeers(t *testing.T) {
```
renaynay, line 64: doc on test pls

renaynay (commented):
just some doc requests, otherwise gtg

renaynay (approved):

distractedm1nd (approved):

Wondertan (approved):

share/availability/discovery/set.go
```diff
@@ -47,6 +50,13 @@ func (ps *limitedSet) TryAdd(p peer.ID) error {
 	}
 	if len(ps.ps) < int(ps.limit) {
 		ps.ps[p] = struct{}{}
+
+		// peer will be pushed to the channel only when somebody is reading from it.
+		// this is done to handle case when Peers() was called on empty set.
+		select {
+		case ps.waitPeer <- p:
+		default:
+		}
```
walldiss, line 59: Looks like if there are multiple waiters we will only notify one. I suggest for loop with return in default case of select.

Nit: Could refactor it to return error earlier and then.

Issue #1467, opened 2022-12-12
core: Broadcast `ShrEx/Sub` notification in `Listener`


Wondertan:
## Overview

The `core.Listener` is the component that exists only in BN. It listens for new blocks coming from the Core Network,
computes EDS with `ExtendedHeader`, and broadcasts it to the network. With the introduction of `ShrEx/Sub` BNs sends
notifications once it is ready to serve the new block; thus, we make `core.Listener` additionally broadcast the 
notification via `eds.PubSub`.

The eds notification must be sent ___after___ Header.

PR #1473, opened 2022-12-12
feat: implement broadcasting a data hash through eds-sub

vgonkivs:
## Overview
Resolves https://github.com/celestiaorg/celestia-node/issues/1467

## Checklist

### DIFF
--- a/header/core/listener.go
+++ b/header/core/listener.go
@@ -11,6 +11,7 @@ import (
 	"github.com/celestiaorg/celestia-node/core"
 	"github.com/celestiaorg/celestia-node/header"
 	libhead "github.com/celestiaorg/celestia-node/libs/header"
+	"github.com/celestiaorg/celestia-node/share/eds"
 )
 
 // Listener is responsible for listening to Core for
@@ -23,6 +24,7 @@ import (
 type Listener struct {
 	bcast     libhead.Broadcaster[*header.ExtendedHeader]
 	fetcher   *core.BlockFetcher
+	pubsub    *eds.PubSub
 	bServ     blockservice.BlockService
 	construct header.ConstructFn
 	cancel    context.CancelFunc
@@ -31,12 +33,14 @@ type Listener struct {
 func NewListener(
 	bcast libhead.Broadcaster[*header.ExtendedHeader],
 	fetcher *core.BlockFetcher,
+	pubsub *eds.PubSub,
 	bServ blockservice.BlockService,
 	construct header.ConstructFn,
 ) *Listener {
 	return &Listener{
 		bcast:     bcast,
 		fetcher:   fetcher,
+		pubsub:    pubsub,
 		bServ:     bServ,
 		construct: construct,
 	}
@@ -100,6 +104,12 @@ func (cl *Listener) listen(ctx context.Context, sub <-chan *types.Block) {
 			err = cl.bcast.Broadcast(ctx, eh, pubsub.WithLocalPublication(syncing))
 			if err != nil {
 				log.Errorw("listener: broadcasting next header", "height", eh.Height(),
+					"err", err)
+			}
+
+			err = cl.pubsub.Broadcast(ctx, eh.DataHash.Bytes())
+			if err != nil {
+				log.Errorw("listener: broadcasting data hash", "height", eh.Height,
 					"err", err)
 			}
 		case <-ctx.Done():
--- a/header/core/listener.go
+++ b/header/core/listener.go
@@ -11,7 +11,7 @@ import (
 	"github.com/celestiaorg/celestia-node/core"
 	"github.com/celestiaorg/celestia-node/header"
 	libhead "github.com/celestiaorg/celestia-node/libs/header"
-	"github.com/celestiaorg/celestia-node/share/eds"
+	"github.com/celestiaorg/celestia-node/share/p2p/shrexpush"
 )
 
 // Listener is responsible for listening to Core for
@@ -28,7 +28,7 @@ type Listener struct {
 	construct header.ConstructFn
 
 	broadcastExtendedHeader libhead.Broadcaster[*header.ExtendedHeader]
-	broadcastEDSHash        eds.BroadcastEDSHash
+	broadcastEDSHash        shrexpush.BroadcastEDSHash
 
 	cancel context.CancelFunc
 }
@@ -36,7 +36,7 @@ type Listener struct {
 func NewListener(
 	bcast libhead.Broadcaster[*header.ExtendedHeader],
 	fetcher *core.BlockFetcher,
-	broadcastEDSHash eds.BroadcastEDSHash,
+	broadcastEDSHash shrexpush.BroadcastEDSHash,
 	bServ blockservice.BlockService,
 	construct header.ConstructFn,
 ) *Listener {
--- a/header/core/listener.go
+++ b/header/core/listener.go
@@ -33,16 +33,16 @@ type Listener struct {
 func NewListener(
 	bcast libhead.Broadcaster[*header.ExtendedHeader],
 	fetcher *core.BlockFetcher,
-	broadcastFn eds.BroadcastFn,
+	broadcastFn eds.BroadcastEDSHash,
 	bServ blockservice.BlockService,
 	construct header.ConstructFn,
 ) *Listener {
 	return &Listener{
-		bcast:     bcast,
-		fetcher:   fetcher,
-		broadcast: broadcastFn,
-		bServ:     bServ,
-		construct: construct,
+		bcast:                   bcast,
+		fetcher:                 fetcher,
+		broadcastExtendedHeader: broadcastFn,
+		bServ:                   bServ,
+		construct:               construct,
 	}
 }
 
@@ -108,7 +108,7 @@ func (cl *Listener) listen(ctx context.Context, sub <-chan *types.Block) {
 			}
 
 			// notify network of new EDS hash
-			err = cl.broadcast(ctx, eh.DataHash.Bytes())
+			err = cl.broadcastExtendedHeader(ctx, eh.DataHash.Bytes())
 			if err != nil {
 				log.Errorw("listener: broadcasting data hash", "height", eh.Height,
 					"err", err)
--- a/header/core/listener.go
+++ b/header/core/listener.go
@@ -28,7 +28,7 @@ type Listener struct {
 	construct header.ConstructFn
 
 	broadcastExtendedHeader libhead.Broadcaster[*header.ExtendedHeader]
-	broadcastEDSHash        shrexpush.BroadcastEDSHash
+	broadcastEDSHash        shrexpush.BroadcastFn
 
 	cancel context.CancelFunc
 }
@@ -36,7 +36,7 @@ type Listener struct {
 func NewListener(
 	bcast libhead.Broadcaster[*header.ExtendedHeader],
 	fetcher *core.BlockFetcher,
-	broadcastEDSHash shrexpush.BroadcastEDSHash,
+	broadcastEDSHash shrexpush.BroadcastFn,
 	bServ blockservice.BlockService,
 	construct header.ConstructFn,
 ) *Listener {
--- a/header/core/listener.go
+++ b/header/core/listener.go
@@ -24,7 +24,7 @@ import (
 type Listener struct {
 	bcast     libhead.Broadcaster[*header.ExtendedHeader]
 	fetcher   *core.BlockFetcher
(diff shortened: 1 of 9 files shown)


nodebuilder/share/constructors.go
```diff
@@ -52,3 +55,24 @@ func newModule(lc fx.Lifecycle, bServ blockservice.BlockService, avail share.Ava
...
+	pubsub, err := eds.NewPubSub(
+		fxutil.WithLifecycle(context.Background(), lc),
+		h,
+		string(network),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	lc.Append(fx.Hook{
```
distractedm1nd, line 69: [optional] Can you try to do this without the hooks? So no separate constructor. But just see if you like it better

share/eds/pubsub.go
```diff
@@ -57,23 +57,10 @@ func (s *PubSub) Start(context.Context) error {
 // * Unregisters all the added Validators
 // * Closes the `ShrEx/Sub` topic
 func (s *PubSub) Stop(context.Context) error {
-	err := s.pubSub.UnregisterTopicValidator(s.pubSubTopic)
-	if err != nil {
-		return err
-	}
-
+	// TODO(@vgonkivs): unregister the topic validator
```
distractedm1nd, line 60: Why does it get removed here?

distractedm1nd (dismissed):

share/eds/pubsub.go
vgonkivs: We've not had a subscriber yet. So, no one is registering a validator rn. Call `UnregisterValidator` fails if there is no validator.

Register/Unregister validator should be added in a PR with the subscribing to the "eds-sub".

nodebuilder/share/constructors.go
vgonkivs: How can we stop pubsub without hooks?

share/eds/pubsub.go
renaynay: 👍🏻

share/eds/pubsub.go
```diff
@@ -57,23 +57,10 @@ func (s *PubSub) Start(context.Context) error {
...
 	return s.topic.Close()
 }
 
-// AddValidator registers given Validator for EDS notifications (DataHash).
-// Any amount of Validators can be registered.
-func (s *PubSub) AddValidator(validate Validator) error {
```
renaynay, line 70: why? are you adding it back in another pr?

header/core/listener.go
```diff
@@ -30,12 +32,14 @@ type Listener struct {
 func NewListener(
 	bcast header.Broadcaster,
 	fetcher *core.BlockFetcher,
+	pubsub *eds.PubSub,
```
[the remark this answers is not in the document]
vgonkivs: Yes. It makes sense. I did not want to create additional `Subscriber/Broadcaster ` interfaces for this particular case and decided to leave a pubSub instance. But `BroadcastFn` looks good to me.

share/eds/pubsub.go
vgonkivs: yes. I do not want to keep this logic until the subscriber will be implemented.

nodebuilder/share/constructors.go
distractedm1nd: i will checkout and try it, then show you the diff in sync

renaynay (commented):
in general looks good, just small nits

just to note, it's annoying that we have a `Broadcaster` interface for headersub but not for edssub. I would like to go with 1 approach for both subs - either we alias the broadcast fn and pass it to the listener, or we create an interface for broadcasters so 
`header.Broadcaster` and `eds.Broadcaster` or something like that. Probably aliasing the broadcast fn is better than abstracting w a one-method interface.

distractedm1nd (approved):

renaynay (dismissed):

share/p2p/shrexpush/pubsub.go
```diff
@@ -61,23 +64,10 @@ func (s *PubSub) Start(context.Context) error {
...
 
-// AddValidator registers given Validator for EDS notifications (DataHash).
-// Any amount of Validators can be registered.
-func (s *PubSub) AddValidator(validate Validator) error {
-	return s.pubSub.RegisterTopicValidator(s.pubSubTopic,
-		func(ctx context.Context, p peer.ID, msg *pubsub.Message) pubsub.ValidationResult {
-			return validate(ctx, p, msg.Data)
-		})
-}
-
```
Wondertan, line 80: Why????

nodebuilder/share/module.go
```diff
@@ -93,6 +95,21 @@ func ConstructModule(tp node.Type, cfg *Config, options ...fx.Option) fx.Option
 					return avail.Stop(ctx)
 				}),
 			)),
+			fx.Provide(fx.Annotate(
+				func(lc fx.Lifecycle, h host.Host, network modp2p.Network) (*shrexpush.PubSub, error) {
+					return shrexpush.NewPubSub(
+						fxutil.WithLifecycle(context.Background(), lc),
```
Wondertan, line 101: We have a global long-lived context provided, so pls use it

header/core/listener.go
```diff
@@ -20,24 +21,30 @@ import (
...
-	fetcher   *core.BlockFetcher
-	bServ     blockservice.BlockService
+	fetcher *core.BlockFetcher
+	bServ   blockservice.BlockService
+
 	construct header.ConstructFn
-	cancel    context.CancelFunc
+
+	broadcastExtendedHeader header.Broadcaster
+	broadcastEDSHash        shrexpush.BroadcastEDSHash
```
Wondertan, line 30: It's OK to depend on shrexsub here directly. This will make a sub pkg of the `header` depend on `eds`, but it is okay for now unless it does not produce any cycles. After the header interface PR is merged, we will move Listener to the `core` pkg.

Wondertan: So pls just make it directly depend

Wondertan (changes requested):

header/core/listener.go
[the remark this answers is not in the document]
Wondertan: We don't need additional interfaces and it's fine to depend on shrex directly there: https://github.com/celestiaorg/celestia-node/pull/1473#discussion_r1054694005

share/eds/pubsub.go
Wondertan: Ahhh, then why did you keep it in the original PR??? Pls return it back

header/core/listener.go
vgonkivs: Does not it break the encapsulation rule? We do not need the whole service but only one method. So, here I agree with @renaynay about having only one method that is expected to use.

share/p2p/shrexpush/pubsub.go
vgonkivs: Just did not want to keep the dead code until subscriber will be implemented

vgonkivs: returned back

vgonkivs:
```
	err := s.pubSub.UnregisterTopicValidator(s.pubSubTopic)
	if err != nil {
		return err
	}
```

It was removed bc pubsub will return an error if there is no validator registered

Wondertan: @walldiss, already started implementing things in DASer relying on AddValidator

header/core/listener.go
Wondertan: It's about abstracting it first of all. I am fine to abstract it though, just do not see a big reason, but let's keep it, ok. The only nit is to call it `shrexpush.BroadcastFn` or `shrexpush.Broadcaster`. The EDSHash suffix is implied by shrexpush as it's only pushes data hashes

renaynay: I agree with @Wondertan then re naming but the reason for why it's abstracted is bc the header.Broadcaster is also "abstracted" - ideally they'd be abstracted in the same way at some point but 🤷🏻

vgonkivs: renamed

header/core/listener.go
```diff
@@ -20,24 +21,30 @@ import (
...
-	bcast     header.Broadcaster
-	fetcher   *core.BlockFetcher
-	bServ     blockservice.BlockService
+	fetcher *core.BlockFetcher
+	bServ   blockservice.BlockService
+
 	construct header.ConstructFn
-	cancel    context.CancelFunc
+
+	broadcastExtendedHeader header.Broadcaster
```
renaynay, line 29: last optional nit (and I know i suggested this naming) but i realised it's bit ugly to do `broadcastExtendedHeader.Broadcast` so let's do something like `extHeaderBroadcaster.Broadcast`

header/core/listener.go
```diff
@@ -96,11 +103,18 @@ func (cl *Listener) listen(ctx context.Context, sub <-chan *types.Block) {
...
 			if err != nil {
 				log.Errorw("listener: broadcasting next header", "height", eh.Height,
 					"err", err)
 			}
+
+			// notify network of new EDS hash
+			err = cl.broadcastEDSHash(ctx, eh.DataHash.Bytes())
+			if err != nil {
+				log.Errorw("listener: broadcasting data hash", "height", eh.Height,
+					"err", err)
```
renaynay, line 116: we should also log hash

share/p2p/shrexpush/pubsub.go
```diff
@@ -61,11 +64,7 @@ func (s *PubSub) Start(context.Context) error {
 // * Unregisters all the added Validators
 // * Closes the `ShrEx/Sub` topic
 func (s *PubSub) Stop(context.Context) error {
-	err := s.pubSub.UnregisterTopicValidator(s.pubSubTopic)
-	if err != nil {
-		return err
-	}
-
+	// TODO(@vgonkivs): unregister the topic validator
```
renaynay, line 67: I would leave the unregistration and just log the error if it happens ( we can make an issue for handling the error later if it's really necessary ).

header/core/listener.go
```diff
@@ -20,24 +21,30 @@ import (
...
+	extHeaderBroadcaster header.Broadcaster
+	edsHashBroadcaster   shrexpush.BroadcastFn
+
+	cancel context.CancelFunc
 }
 
 func NewListener(
 	bcast header.Broadcaster,
 	fetcher *core.BlockFetcher,
+	broadcastEDSHash shrexpush.BroadcastFn,
```
distractedm1nd, line 38: rename to edsHashBroadcaster

header/core/listener_test.go
```diff
@@ -32,20 +33,30 @@ func TestListener(t *testing.T) {
 
 	// create one block to store as Head in local store and then unsubscribe from block events
 	fetcher := createCoreFetcher(t)
-
+	eds := createEdsPubSub(ctx, t)
 	// create Listener and start listening
-	cl := createListener(ctx, t, fetcher, ps0)
+	cl := createListener(ctx, t, fetcher, ps0, eds)
+	require.NoError(t, err)
```
distractedm1nd, line 39:
```suggestion
```

distractedm1nd (changes requested):

renaynay (dismissed):

distractedm1nd (dismissed):

renaynay (approved):
lfg, lets rename shrexpush package

distractedm1nd (dismissed):

renaynay (dismissed):

header/core/listener.go
```diff
@@ -21,24 +22,30 @@ import (
...
-	fetcher   *core.BlockFetcher
-	bServ     blockservice.BlockService
+	fetcher *core.BlockFetcher
+	bServ   blockservice.BlockService
+
 	construct header.ConstructFn
-	cancel    context.CancelFunc
+
+	extHeaderBroadcaster libhead.Broadcaster[*header.ExtendedHeader]
+	edsHashBroadcaster   shrexsub.BroadcastFn
```
Wondertan, line 31:
```suggestion
	headerBroadcaster libhead.Broadcaster[*header.ExtendedHeader]
	hashBroadcaster   shrexsub.BroadcastFn
```

Wondertan: We already have Ext mentioning in the type param

Wondertan (commented):
Only one naming nit

header/core/listener.go
vgonkivs: done

Wondertan (approved):

renaynay (approved):
oof

Issue #1406, opened 2022-11-28
improvement(header/p2p): implement retry mechanism for p2p.session


vgonkivs:
Currently, retry mechanism is available only in case of network error in `GetRangeByHeight`, but in fact there are some other use cases(header.ErrHeaderNotFound, unmarshaling errors) where we should also start requesting ranges from another peers.

PR #1477, opened 2022-12-12
feat(header/p2p)!: implement retry mechanism for header/p2p session

vgonkivs:
## Overview
Resolves https://github.com/celestiaorg/celestia-node/issues/1406

- [x] implement a retry mechanism in case of failed requests;
- [x] punish peer in case of invalid data;
- [x] fix issue with peer scoring;

## Checklist

### DIFF
--- a/header/p2p/peer_stats.go
+++ b/header/p2p/peer_stats.go
@@ -4,6 +4,7 @@ import (
 	"container/heap"
 	"context"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/libp2p/go-libp2p-core/peer"
@@ -91,6 +92,7 @@ type peerQueue struct {
 
 	statsLk sync.RWMutex
 	stats   peerStats
+	maxSize uint32
 
 	havePeer chan struct{}
 }
@@ -105,6 +107,7 @@ func newPeerQueue(ctx context.Context, stats []*peerStat) *peerQueue {
 	for _, stat := range stats {
 		pq.push(stat)
 	}
+	atomic.AddUint32(&pq.maxSize, uint32(len(stats)))
 	return pq
 }
 
@@ -112,6 +115,10 @@ func newPeerQueue(ctx context.Context, stats []*peerStat) *peerQueue {
 // in case if there are no peer available in current session, it blocks until
 // the peer will be pushed in.
 func (p *peerQueue) waitPop(ctx context.Context) *peerStat {
+	if atomic.LoadUint32(&p.maxSize) == 0 {
+		return &peerStat{}
+	}
+
 	select {
 	case <-ctx.Done():
 		return &peerStat{}
@@ -131,4 +138,8 @@ func (p *peerQueue) push(stat *peerStat) {
 	p.statsLk.Unlock()
 	// notify that the peer is available in the queue, so it can be popped out
 	p.havePeer <- struct{}{}
+}
+
+func (p *peerQueue) reduceSize() {
+	atomic.AddUint32(&p.maxSize, ^uint32(0)) // decrease counter
 }
--- a/header/p2p/peer_stats.go
+++ b/header/p2p/peer_stats.go
@@ -29,9 +29,9 @@ type peerStat struct {
 func (p *peerStat) updateStats(amount uint64, time uint64) {
 	p.Lock()
 	defer p.Unlock()
-	var averageSpeed float32
+	averageSpeed := float32(amount)
 	if time != 0 {
-		averageSpeed = float32(amount / time)
+		averageSpeed /= float32(time)
 	}
 	if p.peerScore == 0.0 {
 		p.peerScore = averageSpeed
--- a/header/p2p/peer_stats.go
+++ b/header/p2p/peer_stats.go
@@ -4,7 +4,6 @@ import (
 	"container/heap"
 	"context"
 	"sync"
-	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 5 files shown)


distractedm1nd (changes requested):

header/p2p/session.go
```diff
@@ -125,7 +148,17 @@ func (s *session) doRequest(
...
+		switch err {
+		case header.ErrNotFound:
+		default:
+			s.peerTracker.blockPeer(stat.peerID)
+			log.Errorw("header/p2p: unexpected error received", "err", err, "pID", stat.peerID)
+		}
+		select {
+		case <-ctx.Done():
+		case <-s.ctx.Done():
+		case s.reqCh <- req:
```
distractedm1nd, line 160:
```suggestion
		// retry request
		case s.reqCh <- req:
```

distractedm1nd (approved):
optional suggestion

header/p2p/peer_tracker.go
```diff
@@ -190,3 +194,28 @@ func (p *peerTracker) stop() {
...
+	log.Debugw("header/p2p: blocked peer", "pID", pID)
+
+	p.peerLk.Lock()
+	defer p.peerLk.Unlock()
+	// remove peer from cache.
+	if _, ok := p.trackedPeers[pID]; ok {
+		delete(p.trackedPeers, pID)
+	} else if _, ok = p.disconnectedPeers[pID]; ok {
+		delete(p.disconnectedPeers, pID)
+	}
```
Wondertan, line 220: In Go, you don't need to check if the entry is on the map or not, because it does so internally. If there is nothing remove, the delete is just noop

header/p2p/session.go
```diff
@@ -125,7 +148,17 @@ func (s *session) doRequest(
 
 	h, err := s.processResponse(r)
 	if err != nil {
-		s.errCh <- err
+		switch err {
+		case header.ErrNotFound:
+		default:
+			s.peerTracker.blockPeer(stat.peerID)
```
Wondertan, line 154: Just in general. We need to be really careful with that and not block peers for the wrong reasons. E.g., our protocol allows `header.ErrMessageLimitExceeded` error, but we don't check for it in the switch.

Wondertan (changes requested):
LGTM, modulo 3 comments

header/p2p/session.go
vgonkivs: Our client checks the size of a message before sending it. So, It's not possible to get such an error in a normal case.

Wondertan: Why do we need this error than to be a status code?

vgonkivs: Our server can handle requests from external clients.

Wondertan: WDYM?

vgonkivs: I mean that anyone can write a custom client and open the stream with our server and start requesting headers. I think that is the reason why did we add this status code. On the other side, our client checks if the range is less than the limit and immediately returns an error to avoid any networking staff if we already know that this range request won't be handled.

distractedm1nd (approved):

header/p2p/session.go
Wondertan: > I mean that anyone can write a custom client and open the stream with our server and start requesting headers

This means this custom client is misbehaving according to the protocol, and in other shrex protocols we recently went with the approach where in case of misbehavior, we just reset a stream without telling why. Mainly because misbehaviors are "speced out". Additionally, there is no need to be nice about it and send a status code, because there is no actual usage for this error code, really our client does nothing about the status code.  Discuss!

N.B. I understand why we need to have the Go error for the client side and I was my suggestion, the point is about the error code on the protocol level. It seems like an additional bloat to the protocol without real usage, which we may want cleanup, not in this PR ofc.

vgonkivs: Why can't we respond to this client about limitations? So it will be able to send a request with less range.

vgonkivs: Stream reset in this case is not the best solution IMO

Wondertan: > Why can't we respond to this client about limitations?

But it does not do so, right? And how is it supposed to ___guess___ the right size if the server does not give it?

header/p2p/peer_stats.go
```diff
@@ -112,6 +112,10 @@ func newPeerQueue(ctx context.Context, stats []*peerStat) *peerQueue {
 // in case if there are no peer available in current session, it blocks until
 // the peer will be pushed in.
 func (p *peerQueue) waitPop(ctx context.Context) *peerStat {
+	// TODO(vgonkivs): implement fallback solution for cases when peer queue is empty.
+	// As we discussed with @Wondertan there could be 2 possible solutions:
+	// * use libp2p.Discovery to find new peers outside peerTracker to request headers;
+	// * implement IWANT/IHAVE messaging system and start requesting ranges from the Peerstore;
```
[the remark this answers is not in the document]
vgonkivs: Will create

Wondertan (approved):
LGTM, +1 to Rene's comments

header/p2p/peer_stats.go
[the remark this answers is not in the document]
vgonkivs: https://github.com/celestiaorg/celestia-node/issues/1536

renaynay (approved):

Wondertan (approved):

PR #1482, opened 2022-12-13
improvement(cel-key): Network alias check for `--p2p.network` flag

renaynay:
Allows users to pass the network alias instead of the full network name to the `cel-key` utility as is done with the celestia binary. 

We allow networks other than the "valid" networks to be specified with the cel-key utility since it can be used for testing, but it will print a warning.

### DIFF
--- a/nodebuilder/p2p/network.go
+++ b/nodebuilder/p2p/network.go
@@ -46,10 +46,10 @@ var networksList = map[Network]struct{}{
 	Private: {},
 }
 
-// networkAliases is a strict list of all known long-standing networks
+// NetworkAliases is a strict list of all known long-standing networks
 // mapped from the string representation of their *alias* (rather than
 // their actual value) to the Network.
-var networkAliases = map[string]Network{
+var NetworkAliases = map[string]Network{
 	"arabica": Arabica,
 	"mocha":   Mocha,
 	"private": Private,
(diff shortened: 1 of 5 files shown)


nodebuilder/p2p/network.go
```diff
@@ -46,10 +46,10 @@ var networksList = map[Network]struct{}{
 	Private: {},
 }
 
-// networkAliases is a strict list of all known long-standing networks
+// NetworkAliases is a strict list of all known long-standing networks
 // mapped from the string representation of their *alias* (rather than
 // their actual value) to the Network.
-var networkAliases = map[string]Network{
+var NetworkAliases = map[string]Network{
```
Wondertan, line 52: networkList is private, but aliases are not 🤔 

My suggestion is to make `Validate` check for aliases as well to keep aliases encapsulated

Wondertan: Dunk, I missed that we actually need the right Network to be swapped so that we determine the path correctly, and this is why you make Validate return it, which is also a bit weird. Another thought I had is to make the path to take aliases instead of real network names. That's nice because there will never be two Arabicas simultaneously, but it is slightly breaking. This, though, would allow us to make Validate not return anything, as we use aliases everywhere besides where real chain-id is actually needed.

vgonkivs (approved):

distractedm1nd (approved):

nodebuilder/p2p/network.go
renaynay: I think that adds more work than is worth.

Wondertan: Agreed, just don't like that Validate returns new network

Wondertan (approved):
Long term the 🅱️ alidate should not return a network, but its fine for now

PR #1483, opened 2022-12-14
fix(state): Verify the state inclusion proof without splitting keys

### LINKED ISSUE #1461: bug: key mismatch on operation
### Celestia Node version

v0.5.0-rc5

### OS

MacOS and Ubuntu

### Install tools

_No response_

### Others

_No response_

### Steps to reproduce it

address: `celestia13uwfur9qylte248utjl7m2yq97mjdkcr3g857c`
The curl and error:
```
jcs @ ~/celestia-node % curl -X GET http://localhost:26659/balance
"key mismatch on operation #0: expected \ufffd\u0026\ufffd\u0003utia but got \u0002\u0014\ufffd\u001c\ufffd\u000c\ufffd'וT\ufffd\\\ufffd\ufffd\ufffd\ufffd/\ufffd\u0026\ufffd\u0003utia"%
```
on the Light Node, the error shows:
```
2022-12-09T09:00:29.446-0500	ERROR	gateway	gateway/util.go:9	serving request	{"endpoint": "/balance", "err": "key mismatch on operation #0: expected \ufffd&\ufffd\u0003utia but got \u0002\u0014\ufffd\u001c\ufffd\u000c\ufffd'וT\ufffd\\\ufffd\ufffd\ufffd\ufffd/\ufffd&\ufffd\u0003utia"}
```
I can confirm that I have topped up with some tokens again this morning with 1 TIA.
—
i can also confirm that this persists even if I query the balance from another account running my node, which is odd. also, if i re-import the mnemonic, it persists
```
jcs @ ~/celestia-node % curl -X GET http://localhost:26659/balance/celestia13uwfur9qylte248utjl7m2yq97mjdkcr3g857c
"key mismatch on operation #0: expected \ufffd\u0026\ufffd\u0003utia but got \u0002\u0014\ufffd\u001c\ufffd\u000c\ufffd'וT\ufffd\\\ufffd\ufffd\ufffd\ufffd/\ufffd\u0026\ufffd\u0003utia"%
```

### Expected result

```
jcs @ ~/celestia-node % curl -X GET http://localhost:26659/balance
{"denom":"utia","amount":"1000000"}%
```

### Actual result

```
jcs @ ~/celestia-node % curl -X GET http://localhost:26659/balance/celestia13uwfur9qylte248utjl7m2yq97mjdkcr3g857c
"key mismatch on operation #0: expected \ufffd\u0026\ufffd\u0003utia but got \u0002\u0014\ufffd\u001c\ufffd\u000c\ufffd'וT\ufffd\\\ufffd\ufffd\ufffd\ufffd/\ufffd\u0026\ufffd\u0003utia"%
```
and
```
jcs @ ~/celestia-node % curl -X GET http://localhost:26659/balance
"key mismatch on operation #0: expected \ufffd\u0026\ufffd\u0003utia but got \u0002\u0014\ufffd\u001c\ufffd\u000c\ufffd'וT\ufffd\\\ufffd\ufffd\ufffd\ufffd/\ufffd\u0026\ufffd\u0003utia"%
```

### Relevant log output

_No response_

### Notes

_No response_

evan-forbes:
## Overview

There has to be better way of fixing this, but I'm unsure of how we want to handle it, so submitting this for now.

The bug is that, when verifying state inclusion proofs, tendermint will perform [some processing](https://github.com/celestiaorg/celestia-core/blob/a0ccbd068927b638705f63acdf6303d3091da985/crypto/merkle/proof_key_path.go#L85-L110) on the keypath that you give it that the sdk simply doesn't have to do. The iavl doesn't do this, as we just pass it the key it wants, not the path. It just sticks the key that you give in the [request right back](https://github.com/cosmos/cosmos-sdk/blob/b46d83850270b82496e4cd61421d9ebc8bac87f2/store/iavl/store.go#L324-L326), and because the processing in tendermint sometimes results in a different key, the proof verification [fails](https://github.com/celestiaorg/celestia-core/blob/a0ccbd068927b638705f63acdf6303d3091da985/crypto/merkle/proof_op.go#L52).

Specifically, the string encoding of the bytes we pass it sometimes have the string `/` and `%` in them from the address, the processing tendermint does thinks its supposed to do something when it actually isn't, and then the correct key returned by the sdk no longer matches. This is why we're only seeing some addresses break!

[bad processing point 1](https://github.com/celestiaorg/celestia-core/blob/a0ccbd068927b638705f63acdf6303d3091da985/crypto/merkle/proof_key_path.go#L91)
[bad processing point 2](https://github.com/celestiaorg/celestia-core/blob/a0ccbd068927b638705f63acdf6303d3091da985/crypto/merkle/proof_key_path.go#L102) (idek why this is there)

My quick and dirty solution was to copy paste fork ~~the file from tendermint~~ a few methods that verify proofs, then change the api so we don't have to split the string :laughing: . We should fix it properly soon. This might be a bug with tendermint, but I was admittedly too tired to look into it or write up an issue. We might just need to pass a different formatted string(?)

as a bonus, I added the foundation for an integration test for the core accessor, that should make it waaaaaaaayyyy easier to round trip test things! :slightly_smiling_face: :christmas_tree: 

pls feel to edit this PR and add whatever else is needed to get this across the finish line

closes #1461 

## Checklist


state/integration_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+func (s *IntegrationTestSuite) TestGetBalance() {
+	require := s.Require()
+	expectedBal := sdk.NewCoin(app.BondDenom, sdk.NewInt(int64(99999999999999999)))
+	for _, acc := range s.accounts {
+		bal, err := s.accessor.BalanceForAddress(context.Background(), s.getAddress(acc))
+		require.NoError(err)
+		require.Equal(&expectedBal, bal)
+	}
+}
```
evan-forbes, line 113: fails on main

evan-forbes:
what version of `goimports` are we supposed to use? I tried a few different version using the `-local` and `-w` flag and none of them change anything to the file

share/eds/store_test.go
```diff
@@ -85,7 +84,6 @@ func TestEDSStore_GetCAR(t *testing.T) {
 	assert.NoError(t, err)
 	carReader, err := car.NewCarReader(r)
 
-	fmt.Println(car.HeaderSize(carReader.Header))
```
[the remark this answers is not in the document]
evan-forbes: I put this back, cause the tests failed in CI? is that a thing, or just a coincidence and flakey tests?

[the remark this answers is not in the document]
renaynay: @distractedm1nd

state/core_access.go
```diff
@@ -220,9 +233,14 @@ func (ca *CoreAccessor) BalanceForAddress(ctx context.Context, addr Address) (*B
 		return nil, fmt.Errorf("cannot convert %s into sdktypes.Int", string(value))
 	}
 	// verify balance
-	path := fmt.Sprintf("/%s/%s", banktypes.StoreKey, string(prefixedAccountKey))
-	prt := rootmulti.DefaultProofRuntime()
-	err = prt.VerifyValue(result.Response.GetProofOps(), head.AppHash, path, value)
+	prt := prover.DefaultProofRuntime()
+	rootmulti.DefaultProofRuntime()
```
Wondertan, line 237: Redundant call?

state/prover/proof_op.go
```diff
@@ -0,0 +1,136 @@
+package prover
```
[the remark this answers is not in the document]
Wondertan: Please document what diverged from Tendermint in the code

Wondertan, line 1: Additionally, we can consider switching back to IBC prover, but we also have to verify it for the correctness

cc @renaynay

state/integration_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+	"github.com/celestiaorg/celestia-app/app"
+	"github.com/celestiaorg/celestia-app/testutil/testnode"
+	blobtypes "github.com/celestiaorg/celestia-app/x/payment/types"
+	"github.com/celestiaorg/celestia-node/header"
+)
+
+func TestIntegrationTestSuite(t *testing.T) {
+	suite.Run(t, new(IntegrationTestSuite))
+}
```
Wondertan, line 20: Thank you! ❤️ 
Finally, we have tests for the state interactions. We should also allocate time after incentivized testnet to cover all other APIs with roundtrip tests(potential in the Swamp) Added this to our post-incentivized TODO list @renaynay

state/core_access.go
```diff
@@ -220,9 +221,15 @@ func (ca *CoreAccessor) BalanceForAddress(ctx context.Context, addr Address) (*B
 		return nil, fmt.Errorf("cannot convert %s into sdktypes.Int", string(value))
 	}
 	// verify balance
-	path := fmt.Sprintf("/%s/%s", banktypes.StoreKey, string(prefixedAccountKey))
-	prt := rootmulti.DefaultProofRuntime()
-	err = prt.VerifyValue(result.Response.GetProofOps(), head.AppHash, path, value)
+	prt := merkle.DefaultProofRuntime()
+	prt.RegisterOpDecoder(storetypes.ProofOpIAVLCommitment, storetypes.CommitmentOpDecoder)
+	prt.RegisterOpDecoder(storetypes.ProofOpSimpleMerkleCommitment, storetypes.CommitmentOpDecoder)
```
evan-forbes, line 226: we might want to abstract this out so its easier to use by others, but I'll leave that for some other PR

go.mod
```diff
@@ -327,5 +327,5 @@ replace (
 	github.com/filecoin-project/dagstore => github.com/celestiaorg/dagstore v0.0.0-20221014072825-395797efb659
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
 	github.com/libp2p/go-libp2p-pubsub v0.7.0 => github.com/celestiaorg/go-libp2p-pubsub v0.6.2-0.20220812132010-46b2a019f2f2
-	github.com/tendermint/tendermint => github.com/celestiaorg/celestia-core v1.5.0-tm-v0.34.20
+	github.com/tendermint/tendermint => github.com/celestiaorg/celestia-core v1.5.0-tm-v0.34.20.0.20221214133108-8692e50bffa7
```
evan-forbes, line 330: do we want a special tag for this? it should be fixed in the new release, but we already released v1.6, so we'd have to have a weird release to continue semi following semver. I'm also ok with completely not following semver for core, and releasing v1.5.1, but would prefer not to release 1.12 that doesn't have the many breaking changes in v1.11

renaynay: I'm okay with not having a special tag in favour if semver for core.

state/core_access.go
renaynay: Not sure what you mean by this, can you explain?

renaynay (dismissed):

state/prover/proof_op.go
renaynay: resolved?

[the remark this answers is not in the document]
renaynay: outdated

state/core_access.go
evan-forbes: creating a runtime to verify that has the iavl proof op decoder registered

so just a func to return a runtime that has that already registered, just like what rootmulti used to do

state/core_access.go
```diff
@@ -24,6 +24,7 @@ import (
 	"github.com/celestiaorg/nmt/namespace"
 
 	"github.com/celestiaorg/celestia-node/header"
+	storetypes "github.com/cosmos/cosmos-sdk/store/types"
```
distractedm1nd, line 27: [nit, optional] put this import in the block with the other cosmos imports

distractedm1nd (dismissed):

state/core_access.go
distractedm1nd: could have been what goimports was complaining about?

distractedm1nd (dismissed):

state/integration_test.go
```diff
@@ -0,0 +1,135 @@
...
+func (s *IntegrationTestSuite) SetupSuite() {
+	if testing.Short() {
+		s.T().Skip("skipping test in unit-tests or race-detector mode.")
+	}
+
+	s.T().Log("setting up integration test suite")
+	require := s.Require()
+
+	// we create an arbitrary number of funded accounts
+	for i := 0; i < 300; i++ {
```
Wondertan, line 45: nit: we create only ten in Swamp because creating those accounts tends to be very slow. You test it out yourself, but I don't wanna block release with it.

state/integration_test.go
```diff
@@ -0,0 +1,135 @@
...
+func (s *IntegrationTestSuite) getAddress(acc string) sdk.Address {
+	rec, err := s.cctx.Keyring.Key(acc)
+	if err != nil {
+		panic(err)
+	}
+
+	addr, err := rec.GetAddress()
+	if err != nil {
+		panic(err)
+	}
```
Wondertan, line 107: optional nit: use require

go.mod
Wondertan: Ok, from me as well. Next bump should have a real version please

Wondertan (approved):

state/integration_test.go
evan-forbes: yeah, this was just a remnant of debugging, where we didn't know which addresses where failing. Sometimes we would get 10 addresses that work fine, but never 300. I should have used specifc priv keys that have addresses that fail

distractedm1nd (approved):

Wondertan (approved):

renaynay (approved):

PR #1484, opened 2022-12-14
feat(state)!: Adding fee parameter to `CoreAccessor` operations

distractedm1nd:
## Overview
Adding fee parameter to `CoreAccessor` operations.

All transaction endpoints now have a "fee" key needed in the JSON request. An example value for this can be 2000. This includes the following endpoints:

- SubmitPayForData
- Transfer
- Delegate
- BeginRedelegate
- Undelegate
- CancelUnbondingDelegation

After this change, a new field is added to the json requests in the gateway:
`{oldRequest..., "fee": 2000}`

### DIFF
--- a/api/gateway/state.go
+++ b/api/gateway/state.go
@@ -43,12 +43,14 @@ type submitTxRequest struct {
 type submitPFDRequest struct {
 	NamespaceID string `json:"namespace_id"`
 	Data        string `json:"data"`
+	Fee         int64  `json:"fee"`
 	GasLimit    uint64 `json:"gas_limit"`
 }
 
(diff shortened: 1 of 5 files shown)


renaynay (commented):
Looks good - can u update public API ADR ?

Also we need to update docs repo with the following req changes

Wondertan (approved):
LGTM, the only concern I have is bloat of params, which we can fix later by passing some of the optional params as a structure. Very low prio rn to even care, but still

vgonkivs (approved):

liamsi (approved):

Issue #1488, opened 2022-12-15
share: `Getter` interface and `IPLDGetter`


Wondertan:
## Overview

In celestia-node, we have various sources to get the data from, and in order to compose them, we introduce
the `Getter` interface.

```go
// Getter interface provides a set of accessors for shares by the Root.
// Automatically verifies integrity of shares(exceptions possible depending on the implementation)
type Getter interface {
  // GetShare gets Share by coordinates in EDS.
  GetShare(context.Context, *Root, row, col int) (Share, error)
  
  // GetEDS gets all the shares in the form of EDS square.
  GetEDS(context.Context, *Root) (*rsmt2d.EDS, error)
  
  // GetSharesByNamespace gets all the shares of the given namespace.
  // Shares are returned in a row-by-row order if the namespace spans multiple rows.
  GetSharesByNamespace(context.Context, *Root, namespace.ID) (NamespaceShares, error)
}

// NamespaceShares represents all the shares with proofs of a specific namespace of an EDS.
type NamespaceShares []RowNamespaceShares

// RowNamespaceShares represents all the shares with proofs of a specific namespace within a Row of an EDS.
type RowNamespaceShares struct {
	Shares []Share
	Proof  *ipld.Proof
}

// Verify checks for the inclusion of the NamespacedShares in the Root and verifies their belonging 
// to the given namespace.
func (NamespaceShares) Verify(*Root, namespace.ID) error
```

### `IPLDGetter`

`IPLDGetter` stems from the existing implementation of `ShareService`, but omits to embed of `Availability`.
`IPLDGetter` verifies data integrity as it's a default IPLD behavior.

distractedm1nd:
`IPLDGetter` shouldn't know about `eds.Store` - and because the results from the queries are not full EDSes it cannot store/cache them there anyways - but this can happen on the blockservice level (which would need to be wrapped like [this](https://github.com/walldiss/celestia-node/blob/shrex_nd/share/eds/adapters.go)).

Wondertan:
The caching is about `GetShares` case, which requests the full EDS underneath. The `FullAvailability` will be changed to use `GetShares` instead and this caching is exactly the place where FN will store retrieved data.

PR #1497, opened 2022-12-15
feat(ShrEx/ND): Create client/server for namespaced data communication

walldiss:
## Overview
New implementation is based on bootstrapped `Getter` interface that will be removed and replaced with proper one once https://github.com/celestiaorg/celestia-node/issues/1488 is merged.
 ```
 type Getter interface {
	GetSharesWithProofsByNamespace(...)
}
```	

 Resolves:
 - implement shrex/nd server https://github.com/celestiaorg/celestia-node/issues/1420
 - implement shrex/nd client https://github.com/celestiaorg/celestia-node/issues/1419

## Checklist

### DIFF
--- a/share/p2p/shrexnd/client.go
+++ b/share/p2p/shrexnd/client.go
@@ -124,7 +124,7 @@ func (c *Client) doRequest(
 		return nil, fmt.Errorf("client-nd: response code is not OK: %w", err)
 	}
 
-	shares, err := responseToNamespacedShares(resp.Rows)
+	shares, err := convertToNamespacedShares(resp.Rows)
 	if err != nil {
 		return nil, fmt.Errorf("client-nd: converting response to shares: %w", err)
 	}
@@ -137,8 +137,8 @@ func (c *Client) doRequest(
 	return shares, nil
 }
 
-// responseToNamespacedShares converts proto Rows to share.NamespacedShares
-func responseToNamespacedShares(rows []*pb.Row) (share.NamespacedShares, error) {
+// convertToNamespacedShares converts proto Rows to share.NamespacedShares
+func convertToNamespacedShares(rows []*pb.Row) (share.NamespacedShares, error) {
 	shares := make([]share.NamespacedRow, 0, len(rows))
 	for _, row := range rows {
 		var proof *ipld.Proof
@@ -178,15 +178,15 @@ func (c *Client) setStreamDeadlines(ctx context.Context, stream network.Stream) 
 		return
 	}
 
-	if c.params.ReadTimeout != 0 {
-		err := stream.SetReadDeadline(time.Now().Add(c.params.ReadTimeout))
+	if c.params.readTimeout != 0 {
+		err := stream.SetReadDeadline(time.Now().Add(c.params.readTimeout))
 		if err != nil {
 			log.Debugf("client-nd: set read deadline: %s", err)
 		}
 	}
 
-	if c.params.WriteTimeout != 0 {
-		err := stream.SetWriteDeadline(time.Now().Add(c.params.ReadTimeout))
+	if c.params.writeTimeout != 0 {
+		err := stream.SetWriteDeadline(time.Now().Add(c.params.readTimeout))
 		if err != nil {
 			log.Debugf("client-nd: set write deadline: %s", err)
 		}
--- a/share/p2p/shrexnd/options.go
+++ b/share/p2p/shrexnd/options.go
@@ -18,14 +18,14 @@ type Option func(*Parameters)
 
 // Parameters is the set of parameters that must be configured for the shrex/eds protocol.
 type Parameters struct {
-	// ReadTimeout sets the timeout for reading messages from the stream.
-	ReadTimeout time.Duration
+	// readTimeout sets the timeout for reading messages from the stream.
+	readTimeout time.Duration
 
-	// WriteTimeout sets the timeout for writing messages to the stream.
-	WriteTimeout time.Duration
+	// writeTimeout sets the timeout for writing messages to the stream.
+	writeTimeout time.Duration
 
-	// ServeTimeout defines the deadline for serving request.
-	ServeTimeout time.Duration
+	// serveTimeout defines the deadline for serving request.
+	serveTimeout time.Duration
 
 	// protocolSuffix is appended to the protocolID and represents the network the protocol is
 	// running on.
@@ -34,23 +34,23 @@ type Parameters struct {
 
 func DefaultParameters() *Parameters {
 	return &Parameters{
-		ReadTimeout:  time.Second * 5,
-		WriteTimeout: time.Second * 10,
-		ServeTimeout: time.Second * 10,
+		readTimeout:  time.Second * 5,
+		writeTimeout: time.Second * 10,
+		serveTimeout: time.Second * 10,
 	}
 }
 
 const errSuffix = "value should be positive and non-zero"
 
 func (p *Parameters) Validate() error {
-	if p.ReadTimeout <= 0 {
-		return fmt.Errorf("invalid stream read timeout: %v, %s", p.ReadTimeout, errSuffix)
+	if p.readTimeout <= 0 {
+		return fmt.Errorf("invalid stream read timeout: %v, %s", p.readTimeout, errSuffix)
 	}
-	if p.WriteTimeout <= 0 {
-		return fmt.Errorf("invalid write timeout: %v, %s", p.WriteTimeout, errSuffix)
+	if p.writeTimeout <= 0 {
+		return fmt.Errorf("invalid write timeout: %v, %s", p.writeTimeout, errSuffix)
 	}
-	if p.ServeTimeout <= 0 {
-		return fmt.Errorf("invalid serve timeout: %v, %s", p.ServeTimeout, errSuffix)
+	if p.serveTimeout <= 0 {
+		return fmt.Errorf("invalid serve timeout: %v, %s", p.serveTimeout, errSuffix)
 	}
 	return nil
 }
@@ -59,6 +59,27 @@ func (p *Parameters) Validate() error {
 func WithProtocolSuffix(protocolSuffix string) Option {
 	return func(parameters *Parameters) {
 		parameters.protocolSuffix = protocolSuffix
+	}
+}
+
+// WithReadTimeout is a functional option that configures the `readTimeout` parameter
+func WithReadTimeout(readTimeout time.Duration) Option {
+	return func(parameters *Parameters) {
+		parameters.readTimeout = readTimeout
+	}
+}
+
+// WithWriteTimeout is a functional option that configures the `writeTimeout` parameter
+func WithWriteTimeout(writeTimeout time.Duration) Option {
+	return func(parameters *Parameters) {
+		parameters.writeTimeout = writeTimeout
+	}
+}
+
+// WithServeTimeout is a functional option that configures the `serveTimeout` parameter
+func WithServeTimeout(serveTimeout time.Duration) Option {
+	return func(parameters *Parameters) {
+		parameters.serveTimeout = serveTimeout
 	}
 }
 
--- a/share/p2p/shrexnd/server.go
+++ b/share/p2p/shrexnd/server.go
@@ -70,7 +70,7 @@ func (srv *Server) Stop() {
 }
 
 func (srv *Server) handleNamespacedData(ctx context.Context, stream network.Stream) {
-	err := stream.SetReadDeadline(time.Now().Add(srv.params.ReadTimeout))
+	err := stream.SetReadDeadline(time.Now().Add(srv.params.readTimeout))
 	if err != nil {
 		log.Debugf("server: setting read deadline: %s", err)
 	}
@@ -82,6 +82,7 @@ func (srv *Server) handleNamespacedData(ctx context.Context, stream network.Stre
 		stream.Reset() //nolint:errcheck
 		return
 	}
+	log.Debugw("server: new request", "namespaceId", string(req.NamespaceId), "roothash", string(req.RootHash))
 
 	err = stream.CloseRead()
 	if err != nil {
@@ -95,7 +96,7 @@ func (srv *Server) handleNamespacedData(ctx context.Context, stream network.Stre
 		return
 	}
 
-	ctx, cancel := context.WithTimeout(ctx, srv.params.ServeTimeout)
+	ctx, cancel := context.WithTimeout(ctx, srv.params.serveTimeout)
 	defer cancel()
 
 	dah, err := srv.store.GetDAH(ctx, req.RootHash)
@@ -114,6 +115,7 @@ func (srv *Server) handleNamespacedData(ctx context.Context, stream network.Stre
 
 	resp := namespacedSharesToResponse(shares)
 	srv.respond(stream, resp)
+	log.Debugw("server: handled request", "namespaceId", string(req.NamespaceId), "roothash", string(req.RootHash))
 }
 
 // validateRequest checks correctness of the request
@@ -167,7 +169,7 @@ func namespacedSharesToResponse(shares share.NamespacedShares) *pb.GetSharesByNa
 }
 
 func (srv *Server) respond(stream network.Stream, resp *pb.GetSharesByNamespaceResponse) {
-	err := stream.SetWriteDeadline(time.Now().Add(srv.params.WriteTimeout))
+	err := stream.SetWriteDeadline(time.Now().Add(srv.params.writeTimeout))
 	if err != nil {
 		log.Debugf("server: seting write deadline: %s", err)
 	}

### CODE REVIEW

**Review by Wondertan — CHANGES_REQUESTED** [review]

**Comment by vgonkivs** [comment]
Are we supposed to have `v2/pb, v3/pb ...` in future?

File: share/p2p/shrex/handler.go [hunk 1]
```diff
@@ -0,0 +1,158 @@
...
+	h.respond(stream, resp)
+}
+
+// respondInternal sends internal error response to client
+func (h *Handler) respondInternal(stream network.Stream) {
+	resp := &share_p2p_v1.GetSharesByNamespaceResponse{
+		Status: share_p2p_v1.StatusCode_INTERNAL,
+	}
+	h.respond(stream, resp)
+}
```
distractedm1nd, line 109: this could be one helper that just takes the status as a param

distractedm1nd (changes requested):
first readthrough

share/p2p/v1/pb/share.proto
```diff
@@ -0,0 +1,35 @@
+syntax = "proto3";
+
+package share.p2p.v1;
+
+message GetSharesByNamespaceRequest{
+  bytes root_hash = 1;
+  repeated bytes row_roots = 2;
+  bytes namespace_id = 3;
+  int64 max_shares = 4;
+  bool collect_proofs = 5;
```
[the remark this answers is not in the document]
walldiss: I agree, do you think it would be worth to place verification login on the level of new Getter implementation, rather than client/server itself? So for handler side data coming from getter should be already verified, and for requester it will return what it got, and the caller will need to verify.

share/p2p/v1/pb/share.proto
```diff
@@ -0,0 +1,35 @@
...
+  StatusCode status = 1;
+  repeated Row rows = 2;
+}
+
+enum StatusCode {
+  INVALID = 0;
+  OK = 1;
+  NOT_FOUND = 2;
+  INVALID_ARGUMENT = 3;
+  INTERNAL = 4;
```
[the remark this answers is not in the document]
walldiss: I've added it so if client got this code in response, it can identify, that it is broken internally, rather than it is networking / server issue.

share/p2p/v1/pb/share.proto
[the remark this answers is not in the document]
walldiss: I've added verification on client side. As for server, I think it is up to Getter interface implementation to verify returned data or not.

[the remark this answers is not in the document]
walldiss: I've added verification on client side. As for server, I think it is up to Getter interface implementation to verify returned data or not.

share/p2p/shrex/handler.go
```diff
@@ -0,0 +1,158 @@
...
+
+var log = logging.Logger("share/server")
+
+const (
+	writeTimeout = time.Second * 10
+	readTimeout  = time.Second * 10
+	serveTimeout = time.Second * 10
+)
+
+type Handler struct {
```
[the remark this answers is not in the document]
walldiss: Done, now it is Server that sets handler internally.

share/p2p/shrex/handler.go
```diff
@@ -0,0 +1,158 @@
...
+	logging "github.com/ipfs/go-log/v2"
+	"github.com/libp2p/go-libp2p-core/network"
+
+	"github.com/celestiaorg/celestia-node/share"
+	"github.com/celestiaorg/celestia-node/share/ipld"
+	share_p2p_v1 "github.com/celestiaorg/celestia-node/share/p2p/v1/pb"
+	"github.com/celestiaorg/go-libp2p-messenger/serde"
+)
+
+const ndProtocolID = "shrEx/nd"
```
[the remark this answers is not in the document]
walldiss: Fixed

share/p2p/shrex/requestor.go
```diff
@@ -0,0 +1,141 @@
...
+		CollectProofs: collectProofs,
+	}
+
+	// if context doesn't have deadline use default one
+	deadline, ok := ctx.Deadline()
+	if !ok {
+		deadline = time.Now().Add(c.defaultTimeout)
+	}
+
+	err = stream.SetWriteDeadline(deadline)
```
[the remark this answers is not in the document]
walldiss: NIce one. Updated to set combined deadline.

share/p2p/shrex/requestor.go
```diff
@@ -0,0 +1,141 @@
...
+
+// GetSharesWithProofs gets shares with option to collect Merkle proofs from remote host
+func (c *Client) GetSharesWithProofs(
+	ctx context.Context,
+	rootHash []byte,
+	rowRoots [][]byte,
+	maxShares int,
+	nID namespace.ID,
+	collectProofs bool,
+	peerID peer.ID,
```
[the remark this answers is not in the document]
walldiss: Oh yeah. Client layer logic was removed after PR split. I've recreated Client and added retry logic.

share/p2p/shrex/handler.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+
+		rows = append(rows, row)
+	}
+
+	resp := &share_p2p_v1.GetSharesByNamespaceResponse{
+		Status: share_p2p_v1.StatusCode_OK,
+		Rows:   rows,
+	}
+	h.respond(stream, resp)
```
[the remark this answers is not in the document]
walldiss: Agree, looks nicer with your suggestion

share/p2p/shrex/handler.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+func (h *Handler) respond(stream network.Stream, resp *share_p2p_v1.GetSharesByNamespaceResponse) {
+	err := stream.SetWriteDeadline(time.Now().Add(h.writeTimeout))
```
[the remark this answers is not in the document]
walldiss: It gets reset one every call, because it is based on relative timeout, rather than fixed deadline

share/p2p/shrex/requestor.go
```diff
@@ -0,0 +1,141 @@
...
+		log.Debugf("set read deadline: %s", err)
+	}
+
+	var resp share_p2p_v1.GetSharesByNamespaceResponse
+	_, err = serde.Read(stream, &resp)
+	if err != nil {
+		return nil, fmt.Errorf("read resp: %w", err)
+	}
+
+	if resp.Status != share_p2p_v1.StatusCode_OK {
```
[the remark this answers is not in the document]
walldiss: Cleaned this one in client implementation

share/p2p/shrex/handler.go
```diff
@@ -0,0 +1,158 @@
...
+
+	"github.com/celestiaorg/celestia-node/share"
+	"github.com/celestiaorg/celestia-node/share/ipld"
+	share_p2p_v1 "github.com/celestiaorg/celestia-node/share/p2p/v1/pb"
+	"github.com/celestiaorg/go-libp2p-messenger/serde"
+)
+
+const ndProtocolID = "shrEx/nd"
+
+var log = logging.Logger("share/server")
```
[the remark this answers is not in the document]
walldiss: Done

share/p2p/shrex/handler.go
walldiss: True, I just like how it looks in main handler:
```
respondInternal(stream)
```
rather than
```
respondWithCode(stream, share_p2p_v1.StatusCode_INTERNAL)
```

walldiss:
@vgonkivs The version will grow if breaking changes are introduced to protocol and we want to keep backwards compatibility.

share/p2p/shrex/server.go
```diff
@@ -0,0 +1,216 @@
...
+	host         host.Host
+	writeTimeout time.Duration
+	readTimeout  time.Duration
+	serveTimeout time.Duration
+
+	// TODO: will be removed after CARBlockstore is ready to return DAH
+	testDAH *share.Root
+}
+
+func StartServer(host host.Host, store *eds.Store) *Server {
```
distractedm1nd, line 46: Why not do a normal constructor + `Start` as a method?

share/p2p/shrex/server.go
```diff
@@ -0,0 +1,216 @@
...
+)
+
+type Server struct {
+	store        *eds.Store
+	host         host.Host
+	writeTimeout time.Duration
+	readTimeout  time.Duration
+	serveTimeout time.Duration
+
+	// TODO: will be removed after CARBlockstore is ready to return DAH
```
distractedm1nd, line 42: can you do this yet?

share/p2p/shrex/server.go
```diff
@@ -0,0 +1,216 @@
...
+	readTimeout  = time.Second * 10
+	serveTimeout = time.Second * 10
+)
+
+type Server struct {
+	store        *eds.Store
+	host         host.Host
+	writeTimeout time.Duration
+	readTimeout  time.Duration
+	serveTimeout time.Duration
```
distractedm1nd, line 40: We can use the params pattern like in #1431 ? But non-blocking, can be done in a followup

share/p2p/shrex/client.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+// GetBlobByNamespace request shares with option to collect proofs from remote peers using shrex protocol
+func (c *Client) GetBlobByNamespace(ctx context.Context, root *share.Root, nID namespace.ID) (*share.Blob, error) {
+	peers := c.testPeers
+	if len(peers) == 0 { //nolint:staticcheck
+		//TODO: collect peers from discovery here
+	}
+
+	for _, peer := range peers {
```
distractedm1nd, line 53: I THINK it should be retried across all peers until all of them are excluded or the data can be returned, like in shrex/eds, but this isn't clear in the ADR. @Wondertan can you clarify?

share/p2p/shrex/client.go
```diff
@@ -0,0 +1,172 @@
...
+	share_p2p_v1 "github.com/celestiaorg/celestia-node/share/p2p/v1/pb"
+	"github.com/celestiaorg/go-libp2p-messenger/serde"
+	"github.com/celestiaorg/nmt/namespace"
+)
+
+var (
+	errInvalidStatusCode = errors.New("INVALID")
+	errNotFound          = errors.New("NOT_FOUND")
+	errServerInternal    = errors.New("SERVER_INTERNAL")
+	errConnRefused       = errors.New("CONN_REFUSED")
```
distractedm1nd, line 25: I dont think we need to create an error variable for each of these, can we not just do an errors.New(status.String()) in the switch and have the cases `fallthrough`?

share/p2p/shrex/client.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	err = stream.SetDeadline(deadline)
+	if err != nil {
+		log.Debugf("set write deadline: %s", err)
+	}
+
+	_, err = serde.Write(stream, req)
+	if err != nil {
+		return nil, fmt.Errorf("write req: %w", err)
```
distractedm1nd, line 97:
```suggestion
		return nil, fmt.Errorf("client: writing request: %w", err)
```

share/p2p/shrex/client.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	err = stream.CloseWrite()
+	if err != nil {
+		log.Debugf("close write side of the stream: %s", err)
```
distractedm1nd, line 102:
```suggestion
		log.Debugf("client: closing stream for writes: %s", err)
```

share/p2p/shrex/client.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	var resp share_p2p_v1.GetSharesByNamespaceResponse
+	_, err = serde.Read(stream, &resp)
+	if err != nil {
+		return nil, fmt.Errorf("read resp: %w", err)
```
distractedm1nd, line 108:
```suggestion
		return nil, fmt.Errorf("client: reading response: %w", err)
```

share/p2p/shrex/client.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	if err := statusToErr(resp.Status); err != nil {
+		return nil, fmt.Errorf("response code is not OK: %w", err)
```
distractedm1nd, line 112:
```suggestion
		return nil, fmt.Errorf("client: response code is not OK: %w", err)
```

share/p2p/shrex/client.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	blob, err := responseToBlob(resp.Rows)
+	if err != nil {
+		return nil, fmt.Errorf("convert response to blob: %w", err)
```
distractedm1nd, line 117:
```suggestion
		return nil, fmt.Errorf("client: converting response to blob: %w", err)
```

share/p2p/shrex/client.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	err = blob.Verify(root, nID)
+	if err != nil {
+		return nil, err
```
distractedm1nd, line 122: fmt.Errorf

share/p2p/shrex/client.go
```diff
@@ -0,0 +1,172 @@
...
+	case share_p2p_v1.StatusCode_OK:
+		return nil
+	case share_p2p_v1.StatusCode_NOT_FOUND:
+		return errNotFound
+	case share_p2p_v1.StatusCode_INTERNAL:
+		return errServerInternal
+	case share_p2p_v1.StatusCode_REFUSED:
+		return errConnRefused
+	}
+	return errInvalidStatusCode
```
distractedm1nd, line 171:
```suggestion
	switch code {
	case share_p2p_v1.StatusCode_OK:
		return nil
	case share_p2p_v1.StatusCode_INVALID:
		fallthrough
	case share_p2p_v1.StatusCode_NOT_FOUND:
		fallthrough
	case share_p2p_v1.StatusCode_INTERNAL:
		fallthrough
	case share_p2p_v1.StatusCode_REFUSED:
		return errors.New(code.String())
	default:
		return errInvalidStatusCode
	}
```

share/p2p/shrex/server.go
```diff
@@ -0,0 +1,216 @@
...
+func (srv *Server) handleNamespacedData(stream network.Stream) {
+	err := stream.SetReadDeadline(time.Now().Add(srv.readTimeout))
+	if err != nil {
+		log.Debugf("set read deadline: %s", err)
+	}
+
+	var req share_p2p_v1.GetSharesByNamespaceRequest
+	_, err = serde.Read(stream, &req)
+	if err != nil {
+		log.Errorw("read msg", "err", err)
```
distractedm1nd, line 72:
```suggestion
		log.Errorw("server: reading request", "err", err)
```

share/p2p/shrex/server.go
```diff
@@ -0,0 +1,216 @@
...
+		stream.Reset() //nolint:errcheck
+		return
+	}
+
+	err = stream.CloseRead()
+	if err != nil {
+		log.Debugf("close read side of the stream: %s", err)
```
distractedm1nd, line 79:
```suggestion
		log.Debugf("server: closing stream for reads: %s", err)
```

share/p2p/shrex/server.go
```diff
@@ -0,0 +1,216 @@
...
+		stream.Reset() //nolint:errcheck
+		return
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), srv.serveTimeout)
+	defer cancel()
+
+	bs, dah, err := srv.store.CARBlockstore(ctx, req.RootHash)
+	if err != nil {
+		log.Errorw("get header for datahash", "err", err)
```
distractedm1nd, line 94:
```suggestion
		log.Errorw("server: retrieving DAH for datahash", "err", err)
```

share/p2p/shrex/server.go
```diff
@@ -0,0 +1,216 @@
...
+
+	// replace dah with testDAH within test scope
+	if srv.testDAH != nil {
+		dah = srv.testDAH
+	}
+
+	getter := eds.NewBlockGetter(bs)
+	blob, err := getBlobByNamespace(ctx, getter, dah, req.NamespaceId)
+	if err != nil {
+		log.Errorw("get shares", "err", err)
```
distractedm1nd, line 107:
```suggestion
		log.Errorw("server: retrieving shares", "err", err)
```

share/p2p/shrex/server.go
```diff
@@ -0,0 +1,216 @@
...
+	getter blockservice.BlockGetter,
+	root *share.Root,
+	nID namespace.ID,
+) (*share.Blob, error) {
+	// select rows that contain shares with desired namespace.ID
+	selectedRows := make([]cid.Cid, 0)
+	for _, row := range root.RowsRoots {
+		if !nID.Less(nmt.MinNamespace(row, nID.Size())) && nID.LessOrEqual(nmt.MaxNamespace(row, nID.Size())) {
+			selectedRows = append(selectedRows, ipld.MustCidFromNamespacedSha256(row))
+		}
```
distractedm1nd, line 127: I believe this already exists as a helper function, and if it doesn't yet it should

share/p2p/shrex/server.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+	if len(req.RootHash) != sha256.Size {
+		return fmt.Errorf("incorrect root hash length: %v", len(req.RootHash))
+	}
+
+	return nil
+}
+
+// respondInternal sends internal error response to client
+func (srv *Server) respondInternal(stream network.Stream) {
```
distractedm1nd, line 166: rename to something that indicates that Internal means an error, "respondInternal" seems like it could just be a private response helper

Maybe `respondInternalErr` or `respondInternalError`

distractedm1nd (changes requested):

distractedm1nd:
@walldiss I think we can rename it to v1 once we have a v2 and want such backwards compatibility

share/p2p/shrex/client.go
```diff
@@ -0,0 +1,172 @@
...
+	if len(peers) == 0 { //nolint:staticcheck
+		//TODO: collect peers from discovery here
+	}
+
+	for _, peer := range peers {
+		blob, err := c.getBlobByNamespace(
+			ctx, root, nID, peer)
+		if err != nil {
+			log.Debugw("peer returned err", "peer_id", peer.String(), "err", err)
+			continue
```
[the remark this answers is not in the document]
walldiss: I've not implemented retry here, nor punishment. I think peer management should be a matter of another component or at least a caller. Keeping in mind that discovery could found new peers while try cycle happening, should we go for retry cycle or try new peers instead? And if we go try peers, should we try peers for first those that misbehaved in another call? The separate component can address such case.

share/p2p/shrex/server.go
walldiss: Server always needs to be started right after creation. Why split one operation into 2 funcs?

share/p2p/shrex/server.go
walldiss: now it's time. :headstone:

share/p2p/shrex/server.go
walldiss: Will do a followup.

share/p2p/shrex/client.go
walldiss: Errors has been planned to be handled to punish broken peers (errInvalidStatusCode, errServerInternal). Removed for now, since no peer management is implemented in this RP

share/p2p/shrex/client.go
walldiss: I've updated all logs msgs

share/p2p/shrex/client.go
walldiss: reworked

share/p2p/shrexnd/client.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	err = stream.SetDeadline(deadline)
+	if err != nil {
+		log.Debugf("set write deadline: %s", err)
+	}
+
+	_, err = serde.Write(stream, req)
+	if err != nil {
+		return nil, fmt.Errorf("client: writing request: %w", err)
```
vgonkivs, line 99: stream.Reset? in all error cases

share/p2p/shrexnd/client.go
```diff
@@ -0,0 +1,173 @@
...
+	switch code {
+	case share_p2p_v1.StatusCode_OK:
+		return nil
+	case share_p2p_v1.StatusCode_INVALID,
+		share_p2p_v1.StatusCode_NOT_FOUND,
+		share_p2p_v1.StatusCode_INTERNAL,
+		share_p2p_v1.StatusCode_REFUSED:
+		fallthrough
+	default:
+		code = share_p2p_v1.StatusCode_INVALID
```
vgonkivs, line 170: Can you leave only the default case?

share/p2p/shrexnd/server.go
```diff
@@ -0,0 +1,168 @@
...
+type Server struct {
+	getter       share.Getter
+	store        *eds.Store
+	host         host.Host
+	writeTimeout time.Duration
+	readTimeout  time.Duration
+	serveTimeout time.Duration
+}
+
+func StartServer(host host.Host, store *eds.Store, getter *getters.IPLDGetter) *Server {
```
vgonkivs, line 40: Actually, It does not start anything. it creates an instance of a server. So, I'd suggest to rename to a `NewServer`

share/p2p/shrexnd/client.go
```diff
@@ -0,0 +1,173 @@
...
+	peerIDs peer.IDSlice,
+) (share.NamespacedShares, error) {
+	// overwrite peers in test
+	if len(c.testPeers) != 0 {
+		peerIDs = c.testPeers
+	}
+
+	for _, peer := range peerIDs {
+		shares, err := c.getSharesByNamespace(
+			ctx, root, nID, peer)
```
distractedm1nd, line 57: why this formatting?

share/p2p/shrexnd/client.go
```diff
@@ -0,0 +1,173 @@
...
+)
+
+var (
+	errNotAvailable = errors.New("none of the peers has requested data")
+)
+
+// Client implements Getter interface, requesting data from remote peers
+type Client struct {
+	host           host.Host
+	defaultTimeout time.Duration
```
distractedm1nd, line 28: I wouldn't name any struct fields default, only consts

share/p2p/shrexnd/server.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	err = validateRequest(req)
+	if err != nil {
+		log.Errorw("server: invalid request", "err", err)
+		stream.Reset() //nolint:errcheck
+		return
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), srv.serveTimeout)
```
distractedm1nd, line 84: maybe we should store a ctx, cancel on the struct during `Start` (and `cancel`s on `Stop`) and use it here instead of Background

share/p2p/shrexnd/server.go
```diff
@@ -0,0 +1,168 @@
...
+
+	resp := namespacedSharesToResponse(shares)
+	srv.respond(stream, resp)
+}
+
+// validateRequest checks correctness of the request
+func validateRequest(req share_p2p_v1.GetSharesByNamespaceRequest) error {
+	if len(req.NamespaceId) != ipld.NamespaceSize {
+		return fmt.Errorf("incorrect namespace id length: %v", len(req.NamespaceId))
+	}
```
distractedm1nd, line 109: we could move the helper function from getter/utils to share and then call it here (but just an idea)

share/p2p/shrexnd/client.go
```diff
@@ -0,0 +1,173 @@
...
+	root *share.Root,
+	nID namespace.ID,
+	peerIDs peer.IDSlice,
+) (share.NamespacedShares, error) {
+	// overwrite peers in test
+	if len(c.testPeers) != 0 {
+		peerIDs = c.testPeers
+	}
+
+	for _, peer := range peerIDs {
```
distractedm1nd, line 55: Should it retry until it receives a valid response? It only requests once per peer here. I think this isn't clear in the ADR

share/p2p/shrexnd/client.go
```diff
@@ -0,0 +1,173 @@
...
+// Client implements Getter interface, requesting data from remote peers
+type Client struct {
+	host           host.Host
+	defaultTimeout time.Duration
+
+	// fake peers for testing purposes
+	testPeers []peer.ID
+}
+
+// NewClient creates a new shrEx client
```
distractedm1nd, line 34:
```suggestion
// NewClient creates a new shrex/nd client
```

share/p2p/shrexnd/client.go
```diff
@@ -0,0 +1,173 @@
...
+	share_p2p_v1 "github.com/celestiaorg/celestia-node/share/p2p/shrexnd/v1/pb"
+	"github.com/celestiaorg/go-libp2p-messenger/serde"
+	"github.com/celestiaorg/nmt/namespace"
+)
+
+var (
+	errNotAvailable = errors.New("none of the peers has requested data")
+)
+
+// Client implements Getter interface, requesting data from remote peers
```
distractedm1nd, line 25: https://github.com/uber-go/guide/blob/master/style.md#verify-interface-compliance
```suggestion
var _ share.Getter = (*Client)(nil)

// Client is responsible for requesting namespaced data over the ShrEx/ND protocol.
```

share/p2p/shrexnd/server.go
```diff
@@ -0,0 +1,168 @@
...
+
+var log = logging.Logger("shrex/nd")
+
+const (
+	writeTimeout = time.Second * 10
+	readTimeout  = time.Second * 10
+	serveTimeout = time.Second * 10
+)
+
+type Server struct {
```
distractedm1nd, line 31: godoc

share/p2p/shrexnd/server.go
distractedm1nd: yes, + add godoc

distractedm1nd (changes requested):

share/p2p/shrexnd/server.go
vgonkivs: Also, I'd suggest extracting SetStreamHandler in `Start()`. I know, it will look redundant but we have Start/Stop pattern in all of our services, so can we follow it in shrexnd also?

share/p2p/shrexnd/client.go
```diff
@@ -0,0 +1,173 @@
...
+type Client struct {
+	host           host.Host
+	defaultTimeout time.Duration
+
+	// fake peers for testing purposes
+	testPeers []peer.ID
```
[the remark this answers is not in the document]
walldiss: Gone. `GetSharesByNamespace` reworked to use only one peer.ID

share/p2p/shrexnd/client.go
walldiss: ctrl-v ```stream.Reset() //nolint:errcheck```

share/p2p/shrexnd/client.go
walldiss: Fixed. `fallthrough` was a mistake, removed it. I want to have an actual error string being returned, but not "INVALID" all the time.

share/p2p/shrexnd/client.go
walldiss: one line now

share/p2p/shrexnd/client.go
walldiss: renamed to just `timeout`, it is not default tho

share/p2p/shrexnd/server.go
walldiss: I've passed contexted directly, and saved `cancel` for `Stop` call

share/p2p/shrexnd/server.go
walldiss: It is just a single compare. Feels small enough to keep as is.

share/p2p/shrexnd/client.go
walldiss: Disscussed with @distractedm1nd  and @Wondertan  to move retry logic out of client and let the caller decide on retry

share/p2p/shrexnd/client.go
walldiss: done.

share/p2p/shrexnd/client.go
walldiss: TY. Comment was misleading, it is not an interface implementation. Updated

share/p2p/shrexnd/server.go
walldiss: done

share/p2p/shrexnd/server.go
```diff
@@ -0,0 +1,184 @@
...
+		return fmt.Errorf("incorrect namespace id length: %v", len(req.NamespaceId))
+	}
+	if len(req.RootHash) != sha256.Size {
+		return fmt.Errorf("incorrect root hash length: %v", len(req.RootHash))
+	}
+
+	return nil
+}
+
+// respondInternal sends internal error response to client
```
distractedm1nd, line 133: fix name in godoc

distractedm1nd (changes requested):
last nit i promise

share/p2p/shrexnd/server.go
walldiss: TY. good find

share/p2p/shrexnd/client.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+}
+
+// GetSharesByNamespace request shares with option to collect proofs from remote peers using shrex
+// protocol
+func (c *Client) GetSharesByNamespace(
+	ctx context.Context,
+	root *share.Root,
+	nID namespace.ID,
+	peerID peer.ID,
```
Wondertan, line 42: Can we make it variadic? It still allows only one peer to be requested, but after mapping out the implementation of ShrexGetter today, I see the value of having multiple peers requested.

share/p2p/shrexnd/server.go
```diff
@@ -0,0 +1,184 @@
...
+
+const ndProtocolID = "/shrex/nd/0.0.1"
+
+var log = logging.Logger("shrex/nd")
+
+const (
+	writeTimeout = time.Second * 10
+	readTimeout  = time.Second * 10
+	serveTimeout = time.Second * 10
+)
```
Wondertan, line 29: Can we make this to be params like we did for shrex/eds? It does not have to be registered in our nodebuilder config, but it still should be configurable.

Wondertan (commented):
Two requests

share/p2p/shrexnd/shrex_test.go
```diff
@@ -0,0 +1,93 @@
...
+	"time"
+
+	"github.com/celestiaorg/celestia-node/share/getters"
+
+	bsrv "github.com/ipfs/go-blockservice"
+	"github.com/ipfs/go-datastore"
+	ds_sync "github.com/ipfs/go-datastore/sync"
+	mdutils "github.com/ipfs/go-merkledag/test"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/net/context"
```
Wondertan, line 15: Pls use std context pkg

Wondertan: Will also remove the change in imports

share/p2p/shrexnd/server.go
```diff
@@ -0,0 +1,184 @@
...
+// Stop stops the server
+func (srv *Server) Stop() {
+	srv.cancel()
+	srv.host.RemoveStreamHandler(ndProtocolID)
+}
+
+func (srv *Server) handleNamespacedData(ctx context.Context, stream network.Stream) {
+	err := stream.SetReadDeadline(time.Now().Add(srv.readTimeout))
+	if err != nil {
+		log.Debugf("server: seting read deadline: %s", err)
```
Wondertan, line 77:
```suggestion
		log.Debugf("server: setting read deadline: %s", err)
```

share/p2p/shrexnd/server.go
```diff
@@ -0,0 +1,184 @@
...
+		log.Debugf("server: seting write deadline: %s", err)
+	}
+
+	_, err = serde.Write(stream, resp)
+	if err != nil {
+		log.Errorf("server: writing response: %s", err.Error())
+		stream.Reset() //nolint:errcheck
+		return
+	}
+	stream.Close()
```
Wondertan, line 183: Log out error?

share/p2p/shrexnd/server.go
```diff
@@ -0,0 +1,184 @@
...
+	srv := &Server{
+		getter:       getter,
+		store:        store,
+		host:         host,
+		writeTimeout: writeTimeout,
+		readTimeout:  readTimeout,
+		serveTimeout: serveTimeout,
+	}
+
+	return srv
```
Wondertan, line 55: nit:
```suggestion
	return &Server{
		getter:       getter,
		store:        store,
		host:         host,
		writeTimeout: writeTimeout,
		readTimeout:  readTimeout,
		serveTimeout: serveTimeout,
	}
```

share/p2p/shrexnd/server.go
Wondertan: This will also eliminate the requirement to store them on the server and the serve can just hold params structure

share/p2p/shrexnd/client.go
```diff
@@ -0,0 +1,143 @@
...
+	req := &pb.GetSharesByNamespaceRequest{
+		RootHash:    root.Hash(),
+		NamespaceId: nID,
+	}
+
+	// if context doesn't have deadline use default one
+	deadline, ok := ctx.Deadline()
+	if !ok {
+		deadline = time.Now().Add(c.timeout)
+	}
```
Wondertan, line 59: We use the pattern where if the client does not pass the timeout through context, it means the client does not want any timeouts. This model makes it obsolete passion the timeout to the client. Opninons?

share/p2p/shrexnd/client.go
distractedm1nd: this is just the single request, the exported func is variadic, maybe we should rename to `doRequest`?

share/p2p/shrexnd/client.go
walldiss: This approach gives:
  - ability for client to set timeout once and and worry about it anymore
  - it is safety feature to prevent unintended unset deadline for client call. There is default value for timeout option, if it is unset. 
  
  I would suggest that If we want to let user to have no timeout, it should be explicit, by setting c.timeout to 0:
  ```
  deadline, ok := ctx.Deadline()
if !ok && c.timeout == 0{
     deadline = time.Now().Add(c.timeout)
}
```

File: go.mod [hunk 51]
```diff
@@ -65,6 +65,7 @@ require (
 	go.uber.org/fx v1.18.2
 	go.uber.org/multierr v1.9.0
 	golang.org/x/crypto v0.5.0
+	golang.org/x/net v0.5.0
```
Wondertan: go mod tidy

share/p2p/shrexnd/client.go
```diff
@@ -0,0 +1,193 @@
...
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/libp2p/go-libp2p/core/network"
+
+	"github.com/ipfs/go-cid"
+	"github.com/libp2p/go-libp2p/core/host"
+	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/libp2p/go-libp2p/core/protocol"
```
Wondertan, line 14: sort

share/p2p/shrexnd/server.go
```diff
@@ -0,0 +1,185 @@
...
+)
+
+// Server implements server side of shrex/nd protocol to obtain namespaced shares data from remote
+// peers.
+type Server struct {
+	params     *Parameters
+	protocolID protocol.ID
+	getter     share.Getter
+	store      *eds.Store
+	host       host.Host
```
Wondertan, line 29: nit: it might be worth grouping comps with comps and params with pid

share/p2p/shrexnd/server.go
```diff
@@ -0,0 +1,185 @@
...
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/libp2p/go-libp2p/core/protocol"
+
+	"github.com/libp2p/go-libp2p/core/host"
+	"github.com/libp2p/go-libp2p/core/network"
+	"github.com/minio/sha256-simd"
```
Wondertan, line 12: sort

share/p2p/shrexnd/client.go
```diff
@@ -0,0 +1,193 @@
...
+		return nil, fmt.Errorf("client: reading response: %w", err)
+	}
+
+	if err = statusToErr(resp.Status); err != nil {
+		return nil, fmt.Errorf("client: response code is not OK: %w", err)
+	}
+
+	shares, err := responseToNamespacedShares(resp.Rows)
+	if err != nil {
+		return nil, fmt.Errorf("client: converting response to shares: %w", err)
```
Wondertan, line 114: ndclient or client-nd everywhere?

share/p2p/shrexnd/client.go
walldiss: renamed

share/p2p/shrexnd/server.go
```diff
@@ -0,0 +1,185 @@
...
+		protocolID: protocolID(params.protocolSuffix),
+	}
+
+	return srv, nil
+}
+
+// Start starts the server
+func (srv *Server) Start(ctx context.Context) {
+	ctx, cancel := context.WithCancel(ctx)
+	srv.cancel = cancel
```
Wondertan, line 59: This should be a new context because the context given at the start is short-lived. It's a hug because after ten second(default start timeout) no more headers will be served as it's context deadline error

I like though the way you pass it to the stream handler, though cancel still has to be stored(

Wondertan (changes requested):

share/p2p/shrexnd/options.go
```diff
@@ -0,0 +1,67 @@
...
+
+// WithProtocolSuffix is a functional option that configures the `protocolSuffix` parameter
+func WithProtocolSuffix(protocolSuffix string) Option {
+	return func(parameters *Parameters) {
+		parameters.protocolSuffix = protocolSuffix
+	}
+}
+
+func protocolID(protocolSuffix string) protocol.ID {
+	return protocol.ID(fmt.Sprintf("%s%s", protocolPrefix, protocolSuffix))
```
distractedm1nd, line 66: protocolPrefix is missing a / so rn it will look like: "/shrex/nd/0.0.1mocha"

Wondertan (dismissed):
LGTM
Also fine with using one Param struct for Client and Server

share/p2p/shrexnd/options.go
Wondertan: Good find

walldiss: ty, updated the prefix

Wondertan (approved):

share/p2p/shrexnd/client.go
```diff
@@ -0,0 +1,193 @@
...
+	if err != nil {
+		stream.Reset() //nolint:errcheck
+		return nil, fmt.Errorf("client-nd: reading response: %w", err)
+	}
+
+	if err = statusToErr(resp.Status); err != nil {
+		return nil, fmt.Errorf("client-nd: response code is not OK: %w", err)
+	}
+
+	shares, err := responseToNamespacedShares(resp.Rows)
```
renaynay, line 127: can we name this func `convertToNamespacedShares`

share/p2p/shrexnd/options.go
renaynay: looks like you didn't

share/p2p/shrexnd/pb/share.proto
```diff
@@ -0,0 +1,32 @@
...
+  StatusCode status = 1;
+  repeated Row rows = 2;
+}
+
+enum StatusCode {
+  INVALID = 0;
+  OK = 1;
+  NOT_FOUND = 2;
+  INTERNAL = 3;
+  REFUSED = 4;
```
renaynay, line 20: In what case would we return a `REFUSED`?

renaynay (commented):
Please also add debug logs throughout server even for success paths 🙏🏻

share/p2p/shrexnd/client.go
```diff
@@ -0,0 +1,193 @@
...
+	"github.com/celestiaorg/go-libp2p-messenger/serde"
+	"github.com/celestiaorg/nmt/namespace"
+)
+
+var errNoMorePeers = errors.New("shrex-nd: all peers returned invalid responses")
+
+// Client implements client side of shrex/nd protocol to obtain namespaced shares data from remote
+// peers.
+type Client struct {
+	params     *Parameters
```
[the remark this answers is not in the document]
Wondertan: Yes, i.e Config of shrex/eds Config was not integrated either

share/p2p/shrexnd/options.go
Wondertan: He updated on the const

share/p2p/shrexnd/pb/share.proto
Wondertan: I am not convinced we need it, but it happens when there is an error getting shares by namespace on the server

share/p2p/shrexnd/client.go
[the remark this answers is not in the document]
renaynay: kk

share/p2p/shrexnd/pb/share.proto
renaynay: then that would be INTERNAL

Wondertan: That's what REFUSED is for

Wondertan: Ahhh, I mixed up those by looking at the diff and see INTERNAL there first

renaynay: got it - lets resolve

distractedm1nd (approved):

Wondertan (approved):

renaynay (approved):

PR #1498, opened 2022-12-15
fix(eds): WriteEDS thread safety for concurrent writingSessions

distractedm1nd:
## Overview
This PR removes usage of nmt's defaultHasher to allow for concurrent writingSessions.

## Checklist

- ~~New and updated code has new and/or updated testing~~
- ~~Visual proof for any user facing features like CLI or documentation updates~~
- ~~Linked issues closed with keywords~~

### DIFF
--- a/share/eds/eds.go
+++ b/share/eds/eds.go
@@ -29,10 +29,6 @@ import (
 	"github.com/celestiaorg/celestia-node/share/ipld"
 )
 
-// childHashLen is the length of each half of an inner node's data. It consists of the respective
-// min and max namespaces, followed by a sha256 hash.
-const childHashLen = (appconsts.NamespaceSize * 2) + sha256.Size
-
 var ErrEmptySquare = errors.New("share: importing empty data")
 
 // writingSession contains the components needed to write an EDS to a CARv1 file with our custom
@@ -112,7 +108,7 @@ func initializeWriter(ctx context.Context, eds *rsmt2d.ExtendedDataSquare, w io.
 	return &writingSession{
 		eds:    eds,
 		store:  store,
-		hasher: nmt.NewNmtHasher(sha256.New(), appconsts.NamespaceSize, true),
+		hasher: nmt.NewNmtHasher(sha256.New(), ipld.NamespaceSize, ipld.NMTIgnoreMaxNamespace),
 		w:      w,
 	}, nil
 }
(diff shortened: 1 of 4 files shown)


share/eds/eds.go
```diff
@@ -3,6 +3,7 @@ package eds
 import (
 	"bytes"
 	"context"
+	"crypto/sha256"
```
Wondertan, line 6: pls use minio dep instead

Wondertan: check it out namespace hasher

vgonkivs (approved):

Wondertan (changes requested):
LGTM, just one thing needs to be changed

share/eds/eds.go
distractedm1nd: https://github.com/celestiaorg/nmt/blob/95e8313eb788b46b30f3f85b20fd9a40bf68d432/hasher.go#L5

Wondertan: I know, but we use a different, more efficient one. The NMT lib does not import it directly not to bloat go.mod, while it still allows setting custom hasher exactly for this case.

distractedm1nd: done

Wondertan (approved):
🚢 

One last thing is to double check other usages of `nmt.Sha256Namespace8FlaggedLeaf`

distractedm1nd:
yup, checked, no other usages

share/eds/eds.go
```diff
@@ -28,16 +29,19 @@ import (
 	"github.com/celestiaorg/celestia-node/share/ipld"
 )
 
+// childHashLen is the length of each half of an inner node's data. It consists of the respective
+// min and max namespaces, followed by a sha256 hash.
+const childHashLen = (appconsts.NamespaceSize * 2) + sha256.Size
```
walldiss, line 34: I wonder if we should exposes `ipld.nmtHashSize` and use it here, it is the same.

walldiss: + for exposing is that I use it in `shrex/nd`. So that would be 2 usages already

distractedm1nd: good find, done

Wondertan (approved):

vgonkivs (approved):
LGTM

renaynay (approved):

walldiss (approved):

PR #1500, opened 2022-12-16
fix(node/p2p)!: remove resource manager and it's API

Wondertan:
This is a temporary fix for bootstrappers. Infinite Limits seem to be buggy. The proper fix is to update to the latest version of libp2p


vgonkivs (approved):

renaynay (approved):

distractedm1nd:
@Wondertan No we don't need to run openrpc-gen, the error is coming from a missing example type and is unrelated

distractedm1nd (changes requested):
please update ADR

distractedm1nd (approved):
Didn't realize not removing from ADR was intentional

PR #1507, opened 2022-12-18
fix: return a struct instead of a pointer in params creation

vgonkivs:
## Overview
Return a struct instead of a pointer to a struct in params creation 
## Checklist

### DIFF
--- a/header/p2p/options.go
+++ b/header/p2p/options.go
@@ -26,8 +26,8 @@ type ServerParameters struct {
 }
 
 // DefaultServerParameters returns the default params to configure the store.
-func DefaultServerParameters() *ServerParameters {
-	return &ServerParameters{
+func DefaultServerParameters() ServerParameters {
+	return ServerParameters{
 		WriteDeadline:  time.Second * 5,
 		ReadDeadline:   time.Minute,
 		MaxRequestSize: 512,
@@ -101,8 +101,8 @@ type ClientParameters struct {
 }
 
 // DefaultClientParameters returns the default params to configure the store.
-func DefaultClientParameters() *ClientParameters {
-	return &ClientParameters{
+func DefaultClientParameters() ClientParameters {
+	return ClientParameters{
 		MinResponses:         2,
 		MaxRequestSize:       512,
 		MaxHeadersPerRequest: 64,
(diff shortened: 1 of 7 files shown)


nodebuilder/header/config.go
```diff
@@ -93,7 +93,7 @@ func (cfg *Config) Validate() error {
 		return fmt.Errorf("module/header: misconfiguration of p2p exchange server: %w", err)
 	}
 
-	if cfg.Client != nil {
+	if !cfg.Client.Empty() {
```
Wondertan, line 96: This check is not necessary, subsequently the Empty method as well

Wondertan (commented):
Can you also check this in other modules besides the `header`?

header/p2p/options.go
```diff
@@ -112,6 +112,13 @@ func DefaultClientParameters() *ClientParameters {
 	}
 }
 
+// Empty checks if config was initialized.
+func (p *ClientParameters) Empty() bool {
+	return p.MinResponses == 0 && p.MaxRequestSize == 0 &&
+		p.MaxHeadersPerRequest == 0 && p.MaxAwaitingTime == 0 &&
+		p.DefaultScore == 0 && p.MaxPeerTrackerSize == 0
+}
```
renaynay, line 120: since bridge nodes are the only nodes that don't run with client, let's just check the type of node and if not bridge, then call validate on client (also make note of that pls in code). Bc this ^ will be annoying to maintain

nodebuilder/header/config.go
renaynay: It is necessary bc bridge nodes will have "invalid" client configs

renaynay (changes requested):

nodebuilder/header/config.go
Wondertan: Ahhh, we should have passed node type to Validate, so that we visible check for the Bridge, or at least left a comment

vgonkivs: checked for Bridge. Left a comment

Wondertan (approved):

renaynay (approved):

PR #1511, opened 2022-12-19
feat(eds/store): `store.GetDAH`

### LINKED ISSUE #1509: share/eds: `Store.GetRoot`
We found out with @walldiss, that we can greatly simplify the `ShrEx/ND` protocol if the NDServer could know the full DAH. One of the approaches is to get from the `header` pkg, which we would like to avoid. Another idea is to get from the CAR file header which is actually stored and addressable by the DataHash that `Shrex/ND` relies on,

distractedm1nd:
## Overview
Closes #1509

## Checklist

### DIFF
--- a/share/eds/store.go
+++ b/share/eds/store.go
@@ -1,6 +1,7 @@
 package eds
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"fmt"
@@ -226,12 +227,12 @@ func (s *Store) GetDAH(ctx context.Context, root share.DataHash) (*share.Root, e
 	}
 	defer accessor.Close()
 
-	reader, err := carv1.NewCarReader(accessor)
+	carHeader, err := carv1.ReadHeader(bufio.NewReader(accessor))
 	if err != nil {
-		return nil, fmt.Errorf("eds/store: failed to read DAH from car reader: %w", err)
+		return nil, fmt.Errorf("eds/store: failed to read car header: %w", err)
 	}
 
-	dah := dahFromCARHeader(reader.Header)
+	dah := dahFromCARHeader(carHeader)
 	if !bytes.Equal(dah.Hash(), root) {
 		return nil, fmt.Errorf("eds/store: content integrity mismatch from CAR for root %x", root)
 	}
@@ -241,9 +242,9 @@ func (s *Store) GetDAH(ctx context.Context, root share.DataHash) (*share.Root, e
 // dahFromCARHeader returns the DataAvailabilityHeader stored in the CIDs of a CARv1 header.
 func dahFromCARHeader(carHeader *carv1.CarHeader) *header.DataAvailabilityHeader {
 	rootCount := len(carHeader.Roots)
-	rootBytes := make([][]byte, rootCount)
-	for i, root := range carHeader.Roots {
-		rootBytes[i] = ipld.NamespacedSha256FromCID(root)
+	rootBytes := make([][]byte, 0)
+	for _, root := range carHeader.Roots {
+		rootBytes = append(rootBytes, ipld.NamespacedSha256FromCID(root))
 	}
 	return &header.DataAvailabilityHeader{
 		RowsRoots:   rootBytes[:rootCount/2],
--- a/share/eds/store.go
+++ b/share/eds/store.go
@@ -14,10 +14,13 @@ import (
 	"github.com/filecoin-project/dagstore/shard"
 	"github.com/ipfs/go-datastore"
 	bstore "github.com/ipfs/go-ipfs-blockstore"
+	carv1 "github.com/ipld/go-car"
 
 	"github.com/celestiaorg/rsmt2d"
 
+	"github.com/celestiaorg/celestia-node/header"
 	"github.com/celestiaorg/celestia-node/share"
+	"github.com/celestiaorg/celestia-node/share/ipld"
 )
 
 const (
@@ -197,16 +200,42 @@ func (s *Store) Blockstore() bstore.Blockstore {
 	return s.bs
 }
 
-// CARBlockstore returns the IPFS blockstore that provides access to the IPLD blocks stored in an
-// individual CAR file.
-func (s *Store) CARBlockstore(ctx context.Context, dataHash []byte) (dagstore.ReadBlockstore, error) {
+// CARBlockstore returns the IPFS blockstore and DAH that provides access to the IPLD blocks stored
+// in an individual CAR file.
+func (s *Store) CARBlockstore(
+	ctx context.Context,
+	dataHash []byte,
+) (dagstore.ReadBlockstore, *header.DataAvailabilityHeader, error) {
 	key := shard.KeyFromString(fmt.Sprintf("%X", dataHash))
 	accessor, err := s.getAccessor(ctx, key)
 	if err != nil {
-		return nil, err
+		return nil, nil, fmt.Errorf("failed to get accessor: %w", err)
 	}
 
-	return accessor.bs, nil
+	reader, err := carv1.NewCarReader(accessor.sa)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to read DAH from car reader: %w", err)
+	}
+	dah := dahFromCARHeader(reader.Header)
+	err = accessor.sa.Close()
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to close car reader after DAH retrieval: %w", err)
+	}
+
+	return accessor.bs, dah, nil
+}
+
+// dahFromCARHeader returns the DataAvailabilityHeader stored in the CIDs of a CARv1 header.
+func dahFromCARHeader(carHeader *carv1.CarHeader) *header.DataAvailabilityHeader {
+	rootCount := len(carHeader.Roots)
+	rootBytes := make([][]byte, rootCount)
+	for i, root := range carHeader.Roots {
+		rootBytes[i] = ipld.NamespacedSha256FromCID(root)
+	}
+	return &header.DataAvailabilityHeader{
+		RowsRoots:   rootBytes[:rootCount/2],
+		ColumnRoots: rootBytes[rootCount/2:],
+	}
 }
 
 func (s *Store) getAccessor(ctx context.Context, key shard.Key) (*accessorWithBlockstore, error) {
--- a/share/eds/store.go
+++ b/share/eds/store.go
@@ -211,7 +211,7 @@ func (s *Store) CARBlockstore(
 	key := shard.KeyFromString(root.String())
 	shardAccessor, err := s.getAccessor(ctx, key)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get accessor: %w", err)
+		return nil, fmt.Errorf("eds/store: failed to get accessor: %w", err)
 	}
 	return shardAccessor.bs, nil
 }
@@ -221,13 +221,13 @@ func (s *Store) GetDAH(ctx context.Context, root share.DataHash) (*share.Root, e
 	key := shard.KeyFromString(root.String())
 	shardAccessor, err := s.acquireShard(ctx, key)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get accessor: %w", err)
+		return nil, fmt.Errorf("eds/store: failed to get accessor: %w", err)
 	}
 	defer shardAccessor.Close()
 
 	reader, err := carv1.NewCarReader(shardAccessor)
 	if err != nil {
-		return nil, fmt.Errorf("failed to read DAH from car reader: %w", err)
+		return nil, fmt.Errorf("eds/store: failed to read DAH from car reader: %w", err)
 	}
 	return dahFromCARHeader(reader.Header), nil
 }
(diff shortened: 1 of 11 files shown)


Wondertan:
@distractedm1nd, It seems like we would have to fully access the shard-avoiding cache in GetCar to solve the issue. The Seeker is fine, but it does not work when multiple requests for the same car happen at the same time. Readers must be exclusive. Pretty embarrassing that I didn't catch it earlier. 

The optimization using the same mmap file works only for the multiple blockstore accessors because it has internal mutex protection. We can try to make similar synchronization and make the same mmap file work over simultaneous GetCar, but in fact, this would be sequential instead of parallel, though it has only one CAR copy in mem, rather than the amount of requests amount.

The simplest solution right now is just to avoid cache in GetCar and further, we can optimize memory usage by making multiple Readers over the same mmap, rather than multiple Reders with personal mmap

share/eds/store.go
```diff
@@ -197,16 +200,42 @@ func (s *Store) Blockstore() bstore.Blockstore {
...
 
-// CARBlockstore returns the IPFS blockstore that provides access to the IPLD blocks stored in an
-// individual CAR file.
-func (s *Store) CARBlockstore(ctx context.Context, dataHash []byte) (dagstore.ReadBlockstore, error) {
+// CARBlockstore returns the IPFS blockstore and DAH that provides access to the IPLD blocks stored
+// in an individual CAR file.
+func (s *Store) CARBlockstore(
+	ctx context.Context,
+	dataHash []byte,
+) (dagstore.ReadBlockstore, *header.DataAvailabilityHeader, error) {
```
Wondertan, line 208: Pls, use `share.Root` instead. Also, it should be `share.DataHash` ,but for some reason it wasn't

Wondertan (changes requested):

share/eds/store.go
distractedm1nd: ah yeah, probably was merged before share.DataHash, good catch

distractedm1nd: fixed

walldiss (dismissed):
Nice!

share/eds/store.go
```diff
@@ -197,16 +200,42 @@ func (s *Store) Blockstore() bstore.Blockstore {
...
+// in an individual CAR file.
+func (s *Store) CARBlockstore(
+	ctx context.Context,
+	dataHash share.DataHash,
+) (dagstore.ReadBlockstore, *share.Root, error) {
+	key := shard.KeyFromString(dataHash.String())
 	accessor, err := s.getAccessor(ctx, key)
 	if err != nil {
-		return nil, err
+		return nil, nil, fmt.Errorf("failed to get accessor: %w", err)
```
vgonkivs, line 212:
```suggestion
		return nil, nil, fmt.Errorf("eds/store: failed to get accessor: %w", err)
```

share/eds/store.go
```diff
@@ -197,16 +200,42 @@ func (s *Store) Blockstore() bstore.Blockstore {
...
 	}
 
-	return accessor.bs, nil
+	reader, err := carv1.NewCarReader(accessor.sa)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to read DAH from car reader: %w", err)
```
vgonkivs, line 217:
```suggestion
		return nil, nil, fmt.Errorf("eds/store: failed to read DAH from car reader: %w", err)
```

share/eds/store.go
```diff
@@ -197,16 +200,42 @@ func (s *Store) Blockstore() bstore.Blockstore {
...
+	}
+	dah := dahFromCARHeader(reader.Header)
+	err = accessor.sa.Close()
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to close car reader after DAH retrieval: %w", err)
```
vgonkivs, line 222:
```suggestion
		return nil, nil, fmt.Errorf("eds/store: failed to close car reader after DAH retrieval: %w", err)
```

vgonkivs (changes requested):

vgonkivs:
Unit test failed :
```
=== RUN   TestBlockstore_Operations
[1266](https://github.com/celestiaorg/celestia-node/actions/runs/3731466266/jobs/6329873704#step:4:1267)
    blockstore_test.go:36: 
[1267](https://github.com/celestiaorg/celestia-node/actions/runs/3731466266/jobs/6329873704#step:4:1268)
        	Error Trace:	/home/runner/work/celestia-node/celestia-node/share/eds/blockstore_test.go:36
[1268](https://github.com/celestiaorg/celestia-node/actions/runs/3731466266/jobs/6329873704#step:4:1269)
[1269](https://github.com/celestiaorg/celestia-node/actions/runs/3731466266/jobs/6329873704#step:4:1270)
        	            	failed to read DAH from car reader: malformed car; header is bigger than util.MaxAllowedSectionSize
[1270](https://github.com/celestiaorg/celestia-node/actions/runs/3731466266/jobs/6329873704#step:4:1271)
        	Test:       	TestBlockstore_Operations
```

share/eds/store.go
Wondertan, line 222: We cannot close the accessor and use the bs afterward, as it closes the mmap internally.

Wondertan (changes requested):

Wondertan:
Note about updating the ADR

vgonkivs (dismissed):
LGTM 🚀

share/eds/store.go
```diff
@@ -197,31 +201,59 @@ func (s *Store) Blockstore() bstore.Blockstore {
...
+	key := shard.KeyFromString(root.String())
+	accessor, err := s.acquireShard(ctx, key)
+	if err != nil {
+		return nil, fmt.Errorf("eds/store: failed to get accessor: %w", err)
 	}
-	if accessor != nil {
-		return accessor, nil
+	defer accessor.Close()
+
+	reader, err := carv1.NewCarReader(accessor)
```
walldiss, line 229: `carv1.ReadHeader` could be also an option here.

share/eds/store.go
```diff
@@ -197,31 +201,59 @@ func (s *Store) Blockstore() bstore.Blockstore {
...
+	if err != nil {
+		return nil, fmt.Errorf("eds/store: failed to read DAH from car reader: %w", err)
 	}
 
-	// wasn't found in cache, so acquire it and add to cache
+	dah := dahFromCARHeader(reader.Header)
+	if !bytes.Equal(dah.Hash(), root) {
```
walldiss, line 235: I wonder if we should handle this one a bit more seriously, then just returning an error to caller. It is strong sign of eds store integrity corruption.

share/eds/store.go
```diff
@@ -197,31 +201,59 @@ func (s *Store) Blockstore() bstore.Blockstore {
...
+	return dah, nil
+}
+
+// dahFromCARHeader returns the DataAvailabilityHeader stored in the CIDs of a CARv1 header.
+func dahFromCARHeader(carHeader *carv1.CarHeader) *header.DataAvailabilityHeader {
+	rootCount := len(carHeader.Roots)
+	rootBytes := make([][]byte, rootCount)
+	for i, root := range carHeader.Roots {
+		rootBytes[i] = ipld.NamespacedSha256FromCID(root)
+	}
```
walldiss, line 247: (Optional) could do append

share/eds/store.go
distractedm1nd: forgot that existed, great catch

share/eds/store.go
Wondertan: We can utilize DagStores recovery mechanisms, etc. Though we can keep it for later. At this point, we just trust SSDs a node is setup on :)

distractedm1nd: Content integrity mismatch is handed up in a similar way from the `dagstore`, so I think I'll keep it like this. But I'm on the fence about this.. panic also is justified here

share/eds/store.go

share/eds/store.go

share/eds/store.go
walldiss: This is not necessarily Disk corruption, but could be data being overwritten by third party, or dagstore bug, or data was corrupted before it has been stored.
The bad thing about this one is that it will be thrown again with same request. It definitely should notify node operator if no fallback recovery logic is triggered.
Anyway, I think this should be discussed/solved outside of the scope of this PR, so resolving for now.

share/eds/store.go
```diff
@@ -231,12 +264,30 @@ func (s *Store) getAccessor(ctx context.Context, key shard.Key) (*accessorWithBl
...
 	case <-ctx.Done():
 		return nil, ctx.Err()
 	}
 }
 
+func (s *Store) getAccessor(ctx context.Context, key shard.Key) (*accessorWithBlockstore, error) {
+	// try to fetch from cache
+	accessor, err := s.cache.Get(key)
+	if err != nil && err != errCacheMiss {
+		log.Errorw("unexpected error while reading key from bs cache %s: %s", key, err)
```
walldiss, line 277:
```suggestion
		log.Errorf("unexpected error while reading key from bs cache %s: %s", key, err)
```

share/eds/store.go
```diff
@@ -231,12 +264,30 @@ func (s *Store) getAccessor(ctx context.Context, key shard.Key) (*accessorWithBl
...
+	if accessor != nil {
+		return accessor, nil
+	}
+
+	// wasn't found in cache, so acquire it and add to cache
+	shardAccessor, err := s.acquireShard(ctx, key)
+	if err != nil {
+		return nil, err
+	}
+	return s.cache.Add(key, shardAccessor)
```
walldiss, line 288: Does the acquireShard has internal synchronisation inside dagstore? If 2 routines simultaneously do not find same key in the cache and then try to acquire it and save it, could that lead to any problems?

share/eds/store.go
```diff
@@ -197,31 +202,59 @@ func (s *Store) Blockstore() bstore.Blockstore {
...
+	for _, root := range carHeader.Roots {
+		rootBytes = append(rootBytes, ipld.NamespacedSha256FromCID(root))
+	}
+	return &header.DataAvailabilityHeader{
+		RowsRoots:   rootBytes[:rootCount/2],
+		ColumnRoots: rootBytes[rootCount/2:],
+	}
+}
+
+func (s *Store) acquireShard(ctx context.Context, key shard.Key) (*dagstore.ShardAccessor, error) {
```
Wondertan, line 255: style nit: in both cases, we return some form of accessor, but we call the method as acquireShard, what about calling this acquireAcessor

share/eds/store.go
```diff
@@ -197,31 +202,59 @@ func (s *Store) Blockstore() bstore.Blockstore {
...
+	key := shard.KeyFromString(root.String())
+	accessor, err := s.acquireShard(ctx, key)
+	if err != nil {
+		return nil, fmt.Errorf("eds/store: failed to get accessor: %w", err)
 	}
-	if accessor != nil {
-		return accessor, nil
+	defer accessor.Close()
+
+	carHeader, err := carv1.ReadHeader(bufio.NewReader(accessor))
```
Wondertan, line 230: Why do we need a buffer?

share/eds/store.go
Wondertan: No, no, there shouldn't be any problems; the DagStore is synchronized and designed around thread safety

Wondertan (approved):
LGTM, a q and a nit

share/eds/store.go
distractedm1nd: `ReadHeader` requires it unfortunately

share/eds/store.go
distractedm1nd: Discussed in sync, going for `getAccessor` and `getCachedAccessor`

vgonkivs (approved):

Wondertan (dismissed):

walldiss (approved):

share/eds/store.go
```diff
@@ -197,31 +202,59 @@ func (s *Store) Blockstore() bstore.Blockstore {
...
-	accessor, err := s.cache.Get(key)
-	if err != nil && err != errCacheMiss {
-		log.Errorw("unexpected error while reading key from bs cache %s: %s", key, err)
+// dahFromCARHeader returns the DataAvailabilityHeader stored in the CIDs of a CARv1 header.
+func dahFromCARHeader(carHeader *carv1.CarHeader) *header.DataAvailabilityHeader {
+	rootCount := len(carHeader.Roots)
+	rootBytes := make([][]byte, 0)
+	for _, root := range carHeader.Roots {
+		rootBytes = append(rootBytes, ipld.NamespacedSha256FromCID(root))
 	}
```
renaynay, line 248: you know how large rootBytes can be so allocate and copy into it

Wondertan: The append style is nice, but you can set the capacity in advance after zero to preallocate

distractedm1nd: oops yeah, forgot about 3 argument make. done

Wondertan (approved):

renaynay (approved):

adlerjohn (approved):
utACK

Issue #1458, opened 2022-12-09
nodebuilder/header.newInitStore failed: context deadline exceeded


rootulp:
### Celestia Node version

v0.4.2

### OS

MacOS

### Steps to reproduce it

I'm attempting to follow [these steps](https://docs.celestia.org/nodes/light-node#start-the-light-node) for Arabica. 

```shell
$ celestia light init
2022-12-09T01:52:57.491-0500	INFO	node	nodebuilder/init.go:19	Initializing Light Node Store over '/Users/rootulp/.celestia-light-arabica-2'
2022-12-09T01:52:57.492-0500	INFO	node	nodebuilder/init.go:50	Saving config	{"path": "/Users/rootulp/.celestia-light-arabica-2/config.toml"}
2022-12-09T01:52:57.492-0500	INFO	node	nodebuilder/init.go:51	Node Store initialized

$ celestia light start --core.ip https://limani.celestia-devops.dev --core.grpc.port 9090 --gateway --gateway.addr localhost --gateway.port 26659
2022-12-09T01:52:58.562-0500	INFO	badger	v2@v2.2007.4/levels.go:183	All 0 tables opened in 0s

2022-12-09T01:52:58.572-0500	INFO	module/state	state/keyring.go:46	NO KEY FOUND IN STORE, GENERATING NEW KEY...	{"path": "/Users/rootulp/.celestia-light-arabica-2/keys"}
2022-12-09T01:52:58.578-0500	INFO	module/state	state/keyring.go:52	NEW KEY GENERATED...

NAME: my_celes_key
ADDRESS: celestia1w6cm9l5320vkknvvvj74g0935e99y9ycf3kgea
MNEMONIC (save this somewhere safe!!!):
purity flat during glory page balcony venue force powder fan swarm nature join link warm harbor patient citizen bunker pilot twelve tonight weird focus

2022-12-09T01:52:58.580-0500	INFO	module/state	state/keyring.go:69	constructed keyring signer	{"backend": "test", "path": "/Users/rootulp/.celestia-light-arabica-2/keys", "key name": "my_celes_key", "chain-id": "arabica-2"}
2022-12-09T01:52:58.580-0500	INFO	module/header	header/config.go:54	No trusted peers in config, initializing with default bootstrappers as trusted peers
2022-12-09T01:53:13.583-0500	ERROR	module/header	header/constructors.go:101	initializing header store failed: context deadline exceeded
[...]

PR #1515, opened 2022-12-20
refactor: exporting nodebuilder/header.InitStore, moving mock to nodebuilder/header

distractedm1nd:
## Overview
Semi-old branch I forgot to open a PR for, that is related to #1458. Although we cannot fix the ugly error message without https://github.com/uber-go/fx/issues/999, we can get rid of an ugly todo in a start hook.

This PR is purely a suggestion, and if the team doesn't like it we can just close it. 

## Checklist


header/p2p/exchange_test.go
```diff
@@ -12,12 +11,11 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
-	tmbytes "github.com/tendermint/tendermint/libs/bytes"
-
 	"github.com/celestiaorg/go-libp2p-messenger/serde"
 
 	"github.com/celestiaorg/celestia-node/header"
 	p2p_pb "github.com/celestiaorg/celestia-node/header/p2p/pb"
+	headerMock "github.com/celestiaorg/celestia-node/nodebuilder/header/mocks"
```
Wondertan, line 18: 😠

Wondertan: It's a custom mock, so we can keep it somewhere in to the header pkg

Wondertan: Nodebuilder can be used to store autogen mocks

Wondertan (changes requested):
Very Nice! Suprised this idea didn't come when we introduced `initStore`. A bit concerned about it being public, but definitely worth it.

Requesting change to mock location

header/p2p/exchange_test.go
distractedm1nd: thanks, fixed

renaynay (approved):

Wondertan (approved):

PR #1518, opened 2022-12-20
feat!: `share.Getter` and `getters.IPLDGetter`

### LINKED ISSUE #1488: share: `Getter` interface and `IPLDGetter`

distractedm1nd:
## Overview
Closes #1488
- removes `service.ShareService`
- introduces `share.Getter`
- implements `getters.IPLDGetter`
- Full/Light `share.Availability` implementations now take a `share.Getter`
- Updates ADR with new interface
- Updates share Module/API defs and mocks

TODO:
- [x] Different directory structure?
- [ ] Test changes, looking forward to suggestions + improvements

## Checklist


Wondertan:
https://github.com/celestiaorg/celestia-node/issues/1488 was updated with the final form of the Getter interface.

We don't have to implement `NamespaceShares.Verify` here and can live it to #1497/

share/getters/ipld.go
```diff
@@ -0,0 +1,126 @@
...
+		return nil, nil
+	}
+
+	errGroup, ctx := errgroup.WithContext(ctx)
+	shares := make([]share.RowNamespaceShares, len(rowRootCIDs))
+	for i, rootCID := range rowRootCIDs {
+		// shadow loop variables, to ensure correct values are captured
+		i, rootCID := i, rootCID
+		errGroup.Go(func() (err error) {
+			var proof *ipld.Proof
```
Wondertan, line 111: This should not be nil so that we actually collect proofs. 
Should we do some basic tests to ensure we don't miss anything like this?

Wondertan (changes requested):
Breaking and needs some adiditonal reviewes on comments and error messages, but overall a great work + refactor. 

Mostly thoughts and notes, but also a few things needs to be changed, like proof collection seems to be broken.

share/availability/full/availability.go
```diff
@@ -61,7 +59,7 @@ func (fa *ShareAvailability) SharesAvailable(ctx context.Context, root *share.Ro
 		panic(err)
 	}
 
-	_, err := fa.rtrv.Retrieve(ctx, root)
+	_, err := fa.getter.GetShares(ctx, root)
```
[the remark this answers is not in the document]
distractedm1nd: ya, sucks

share/availability/light/availability.go
```diff
@@ -76,16 +73,13 @@ func (la *ShareAvailability) SharesAvailable(ctx context.Context, dah *share.Roo
...
 
 	log.Debugw("starting sampling session", "root", dah.Hash())
-	ses := blockservice.NewSession(ctx, la.bserv)
 	errs := make(chan error, len(samples))
 	for _, s := range samples {
 		go func(s Sample) {
-			root, leaf := ipld.Translate(dah, s.Row, s.Col)
-
-			log.Debugw("fetching share", "root", dah.Hash(), "leaf CID", leaf)
-			_, err := share.GetShare(ctx, ses, root, leaf, len(dah.RowsRoots))
```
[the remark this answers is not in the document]
distractedm1nd: Ok, I removed the blockservice session because I wasn't able to understand how it was being used and incorrectly assumed it wasn't being used at all. Very interesting, would love to talk about this in sync

share/getters/ipld.go
distractedm1nd: Why is it nil? We collect proofs in GetSharesByNamespace! Yes we should add tests in this PR for it

distractedm1nd: Vlad helped me understand here, disregard last comment

Wondertan: Let's not disregard the testing part of your message))

share/getters/ipld.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func (ig *IPLDGetter) GetShare(ctx context.Context, dah *share.Root, row, col int) (share.Share, error) {
+	root, leaf := ipld.Translate(dah, row, col)
+	nd, err := share.GetShare(ctx, ig.bServ, root, leaf, len(dah.RowsRoots))
+	if err != nil {
+		return nil, err
+	}
+
+	return nd, nil
```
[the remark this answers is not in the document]
vgonkivs: session manager is at your service 🤣

Wondertan (commented):
In sync with @distractedm1nd, we decided to:
* Remove caching out of `IPLDGetter` and future `ShrexGetter` and make this caching work via a new `TeeGetter`
* Change `GetShares` to `Getter` interface to `GetEDS` returning full rsmt2d EDS structure
* EDS to implement `JSON(Un)Marhaller` interfaces to make it compatible with RPC

share/getters/session.go
```diff
@@ -0,0 +1,43 @@
+package getters
```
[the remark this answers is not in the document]
distractedm1nd: done, resolving

api/rpc_test.go
```diff
@@ -129,6 +129,10 @@ func TestAllReturnValuesAreMarshalable(t *testing.T) {
 }
 
 func implementsMarshaler(t *testing.T, typ reflect.Type) {
+	// TODO(@distractedm1nd): Write marshaller for ExtendedDataSquare
+	if typ.Name() == "ExtendedDataSquare" {
```
Wondertan, line 133: I can make an issue tmrw

share/availability/light/availability_test.go
```diff
@@ -173,20 +154,20 @@ func BenchmarkService_GetSharesByNamespace(b *testing.B) {
 	for _, tt := range tests {
 		b.Run(strconv.Itoa(tt.amountShares), func(b *testing.B) {
 			t := &testing.T{}
-			serv, root := RandServiceWithSquare(t, tt.amountShares)
+			getter, _, root := RandServiceWithSquare(t, tt.amountShares)
```
Wondertan, line 157: Do we have to return the second param? I see it's not being used in multiple places.

Wondertan (approved):

api/rpc_test.go
Wondertan: https://github.com/celestiaorg/rsmt2d/issues/139

Wondertan: https://github.com/celestiaorg/celestia-node/issues/1541

distractedm1nd: https://github.com/celestiaorg/rsmt2d/pull/140

share/availability/light/availability_test.go
distractedm1nd: I fixed this and renamed a few test methods to better match what they really return

vgonkivs (approved):

share/getters/ipld.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	s.Lock()
+	defer s.Unlock()
+	val = s.Load()
+	if val == nil {
+		val = blockservice.NewSession(ctx, service)
+		s.Store(val)
+	}
+	return val
```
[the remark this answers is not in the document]
distractedm1nd: `GetShare` is called multiple times from the same context in light availability when sampling

share/getters/ipld.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+// WithSession stores an empty ses in the context, indicating that a blockservice session should be
+// created.
+func WithSession(ctx context.Context) context.Context {
+	return context.WithValue(ctx, ses{}, &ses{})
+}
+
+func getGetter(ctx context.Context, service blockservice.BlockService) blockservice.BlockGetter {
+	s, ok := ctx.Value(ses{}).(*ses)
```
Wondertan, line 121: I just realized that there is actually a bug here. The key should be the same value. The go doc on Value method recommends making private global var used as a key.

Wondertan (changes requested):

share/getters/ipld.go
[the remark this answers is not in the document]
walldiss: Ah I see, pretty cool approach with optimistic atomic operations :sunglasses:

share/getter.go
```diff
@@ -0,0 +1,98 @@
...
+
+	if len(originalRoots) != len(ns) {
+		return fmt.Errorf("amount of rows differs between root and namespace shares: %v vs %v",
+			len(originalRoots), len(ns))
+	}
+
+	for i, row := range ns {
+		// verify row data against row hash from original root
+		if !row.verify(originalRoots[i], nID) {
+			return fmt.Errorf("row verification failed: %vth row doesn't match original root: %s", i, root.Hash())
```
renaynay, line 66:
```suggestion
			return fmt.Errorf("row verification failed: row %d doesn't match original root: %s", i, root.Hash())
```

nit but it's not `3th`

share/getter.go
```diff
@@ -0,0 +1,98 @@
...
+	originalRoots := make([][]byte, 0)
+	for _, row := range root.RowsRoots {
+		if !nID.Less(nmt.MinNamespace(row, nID.Size())) && nID.LessOrEqual(nmt.MaxNamespace(row, nID.Size())) {
+			originalRoots = append(originalRoots, row)
+		}
+	}
+
+	if len(originalRoots) != len(ns) {
+		return fmt.Errorf("amount of rows differs between root and namespace shares: %v vs %v",
+			len(originalRoots), len(ns))
```
renaynay, line 60:
```suggestion
		return fmt.Errorf("amount of rows differs between root and namespace shares: expected %d, got %d",
			len(originalRoots), len(ns))
```

share/getter.go
```diff
@@ -0,0 +1,98 @@
...
+func (ns NamespaceShares) Flatten() []Share {
+	shares := make([]Share, 0)
+	for _, row := range ns {
+		shares = append(shares, row.Shares...)
+	}
+	return shares
+}
+
+// Verify validates NamespaceShares by checking every row with nmt inclusion proof.
+func (ns NamespaceShares) Verify(root *Root, nID namespace.ID) error {
```
renaynay, line 50: why did this have to be part of this PR cc @walldiss @distractedm1nd 

can we extract into a follow-up?

nodebuilder/share/share.go
```diff
@@ -34,11 +34,13 @@ type Module interface {
...
 	GetShare(ctx context.Context, dah *share.Root, row, col int) (share.Share, error)
-	GetShares(ctx context.Context, root *share.Root) ([][]share.Share, error)
-	// GetSharesByNamespace iterates over a square's row roots and accumulates the found shares in the
-	// given namespace.ID.
-	GetSharesByNamespace(ctx context.Context, root *share.Root, namespace namespace.ID) ([]share.Share, error)
+	// GetEDS gets the full EDS identified by the given root.
+	GetEDS(ctx context.Context, root *share.Root) (*rsmt2d.ExtendedDataSquare, error)
+	// GetSharesByNamespace gets all shares from an EDS within the given namespace.
+	// Shares are returned in a row-by-row order if the namespace spans multiple rows.
+	GetSharesByNamespace(ctx context.Context, root *share.Root, namespace namespace.ID) (share.NamespaceShares, error)
```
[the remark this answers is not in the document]
distractedm1nd: Yes, this sounds better. @Wondertan this has to be updated in the networking ADR as well

share/getters/ipld.go
distractedm1nd: use an empty struct now for the key

docs/adr/adr-009-public-api.md
```diff
@@ -120,13 +120,16 @@ SyncHead(ctx context.Context) (*header.ExtendedHeader, error)
     // GetShare returns the Share from the given data Root at the given row/col
     // coordinates.
     GetShare(ctx context.Context, root *Root, row, col int) (Share, error)
-    // GetSharesByNamespace returns all shares of the given nID from the given data
-    // Root.
+    // GetShares gets all shares in an EDS.
+    // Shares are returned in a row-by-row order.
```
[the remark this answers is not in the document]
distractedm1nd: yes, good catch

nodebuilder/share/share.go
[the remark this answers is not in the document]
distractedm1nd: also renamed `RowNamespaceShares` to `NamespacedRow`, open to better names

share/getter.go
```diff
@@ -0,0 +1,98 @@
...
+	// Shares are returned in a row-by-row order if the namespace spans multiple rows.
+	GetSharesByNamespace(context.Context, *Root, namespace.ID) (NamespaceShares, error)
+}
+
+// NamespaceShares represents all the shares with proofs of a specific namespace of an EDS.
+type NamespaceShares []RowNamespaceShares
+
+// RowNamespaceShares represents all the shares with proofs of a specific namespace within a Row of
+// an EDS.
+type RowNamespaceShares struct {
```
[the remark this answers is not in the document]
distractedm1nd: done

share/getter.go
distractedm1nd: Its just part of the new type, and is tested in this PR - we could extract it but it would add another PR to the stack needed for shrex/nd.

share/getters/ipld.go
walldiss: I think the suggestion by godoc is to use same global `var` (e.g. value of a type) rather than creating new value every time. Also it could be empty struct of `ses` since you can use any comparable type as a key. Struct is comparable and empty struct == another empty struct.

Wondertan: Yeah, we don't need a new type there, but a same value which can be the same session type

distractedm1nd: Found some interesting things inspecting bytecode that I thought I would share:

 1. creating a new type of `type sessionKey struct{}` and then using `sessionKey{}` has the same bytecode as
 ```golang
 type key struct{}
 var sessionKey = key{}
 ``` 
 and using `sessionKey`. I initially picked this new type because I compared it to `var sessionKey = session{}` with `sessionKey` as the key, which produces a lot more bytecode (it copies the value, and linter complains), BUT i had forgotten to check:
 
 2. using `var sessionKey = session{}` and then using `&sessionKey` as the key has the same bytecode as above (except with different addresses ofc)
 
going for #2 as you've suggested

walldiss: First approach has more bytecode because to compare structs golang compares struct fields. An if struct filed happens to be a struct it does the same recursively (for mutex fields inside session). Comparing pointer on the other hand is just int64 compare.

Wondertan: > Comparing pointer on the other hand is just int64 compare

Wait, but if we take two different ptrs from the same value, those pointers would be different

Wondertan: cc @distractedm1nd

vgonkivs (approved):

walldiss (approved):

renaynay (approved):

liamsi (approved):

share/availability/light/availability.go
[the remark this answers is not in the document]
Jorropo: This will be released in the `v0.18.0` release of boxo.
It has been implemented in ipfs/boxo@22fa8b16a3e530b86542286bebba77c08e17487e

PR #1520, opened 2022-12-20
chore(header/p2p): improve logging

vgonkivs:
## Overview
Change the logging level in the case when headers are not found in the store.
## Checklist

### DIFF
--- a/header/p2p/server.go
+++ b/header/p2p/server.go
@@ -2,6 +2,7 @@ package p2p
 
 import (
 	"context"
+	"errors"
 	"time"
 
 	"github.com/libp2p/go-libp2p-core/host"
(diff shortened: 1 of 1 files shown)


renaynay (approved):

distractedm1nd (dismissed):

header/p2p/server.go
```diff
@@ -183,6 +184,10 @@ func (serv *ExchangeServer) handleRequest(from, to uint64) ([]*header.ExtendedHe
 	defer cancel()
 	headersByRange, err := serv.store.GetRangeByHeight(ctx, from, to)
 	if err != nil {
+		if errors.Is(err, context.DeadlineExceeded) {
+			log.Warnw("server: requested headers not found", "from", from, "to", to)
+			return nil, err
```
Wondertan, line 189: We should return `header.ErrNotFound` here, actually, so that the logic above sends the correct status code to the remote side, otherwise, we just do a stream reset

Wondertan (changes requested):

header/p2p/server.go
Wondertan: This is actually a bug and it very nice that we found it

renaynay: +1

vgonkivs: Nice catch. Thank you

distractedm1nd (approved):

renaynay (approved):

Wondertan (approved):

Issue #1448, opened 2022-12-07
Discrepancy between share data of `namespaced_shares` endpoint and `namespaced_data` endpoint


rootulp:
Originally reported by @preston-evans98 
Celestia node version: v0.5.0-rc5

```bash
~/Developer$ curl -X GET 'http://localhost:26659/namespaced_data/0000000000000001/height/45963'
{"data":["AAAPAMMCCiCZ9+JcsjNA3DSTJ8e+krN9ultzZ6tMrU0vVDuPMu1BXRKcAgp8CnoKFi9wYXltZW50Lk1zZ1BheUZvckRhdGESYAovY2VsZXN0aWExaDZtZHpudHU1YXhjNThoOXZ2Nm5yYTNjZjQ2eWZuM3cyMm5lNWMSCNuEGxw2TrEZGIkDIiDymBZpIYaCo4NxzP3Oc9coOtobH/VMEUhOX13/bIYkIxJaClEKRgofL2Nvc21vcy5jcnlwdG8uc2VjcDI1NmsxLlB1YktleRIjCiEDP4HInbKE6pwV+gIRgAUcIlznG4oPxuMQdVIw7aQsLQ8SBAoCCAEYwgISBRCAm+4CGkBJkJPpPidYN5Ob6STnf8qSNI3KeN/hhGjnKkBBg+Yh32ErCfqqeWBSAX8IIoGYS3EttMhfj9/A2mJlAMam2SA=="],"height":45963}%

~/Developer$ curl -X GET 'http://localhost:26659/namespaced_shares/0000000000000001/height/45963'
[...]

PR #1526, opened 2022-12-21
fix: `namespaced_data` endpoint

rootulp:
Closes https://github.com/celestiaorg/celestia-node/issues/1448

## Tasks
- [x] merge https://github.com/celestiaorg/celestia-app/pull/1146
- [x] cut a release of celestia-app (see https://github.com/celestiaorg/celestia-app/releases/tag/v0.12.0-rc2)
- [x] extract a separate PR that upgrades to this release ^ and likely includes https://github.com/celestiaorg/celestia-node/pull/1526/commits/671ca88aa0a2e199722830e652dfc05b4d4831aa (see https://github.com/celestiaorg/celestia-node/pull/1617)
- [x] rebase this PR so that it only includes the relevant change in `api/gateway/share.go`
- [x] merge `main` into this PR after upgrading to celestia-app v0.12.0-rc-4

## Follow ups
It would be nice to include a swamp test that verifies the fix because manually verifying (e.g. via [these steps](https://gist.github.com/rootulp/3e966bba69ef3d748cd9cb0c1c2f21eb)) is time consuming and doesn't run in CI.


api/gateway/share_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+func Test_dataFromShares(t *testing.T) {
+	type testCase struct {
+		name    string
+		input   [][]byte
+		want    [][]byte
+		wantErr bool
+	}
+
+	smallTxInput := padShare([]uint8{
```
rootulp, line 19: [for reviewers] this unit test is tightly coupled to the share format defined in celestia-app so I can remove these unit tests if maintainers think celestia-node should treat the shares returned from celestia-app as opaque

rootulp:
Findings from manual test in [Go playground](https://go.dev/play/p/ELZwK9R9MjY). The namespaced data is 4 bytes off because it thinks the PFB namespace is not a compact share. This was fixed in https://github.com/celestiaorg/celestia-app/pull/1281 but not included in 0.12.0-rc2 so next steps here are to get that PR included in the next release candidate and retry the manual test.

rootulp:
🎉 fixed with rc-3 https://go.dev/play/p/V0oEGLem3c4

api/gateway/share.go
```diff
@@ -111,14 +111,17 @@ func (h *Handler) getShares(ctx context.Context, height uint64, nID namespace.ID
 	return shares.Flatten(), header.Height(), err
 }
 
-func dataFromShares(shares []share.Share) ([][]byte, error) {
-	blobs, err := appshares.ParseBlobs(shares)
```
rootulp, line 115: [for reviewers] this was the main issue. celestia-node was previously treating all types of shares as blob shares but now there are different types of shares (blobs occupy sparse shares and normal transactions / PFB transactions occupy compact shares). `ParseShares` is able to handle both types of shares.

api/gateway/share.go
```diff
@@ -111,14 +111,17 @@ func (h *Handler) getShares(ctx context.Context, height uint64, nID namespace.ID
...
+func dataFromShares(input []share.Share) (data [][]byte, err error) {
+	sequences, err := shares.ParseShares(input)
```
Wondertan, line 115: I would really love to dive deeper into what this does, and we might want actually to return it from our API(integrate into [Getter](https://github.com/celestiaorg/celestia-node/blob/main/share/getter.go#L29) somehow) or substitute current `Flatten` method. There might be a case where unparsed shares are not useful at all, and we might always want to return them parsed. WDYT? 

At this point, I don't have enough knowledge to understand how to move forward with this, as we want to merge this ASAP. If you don't have any ideas about this, we can make an issue and move forward with this and clean this up later.

Wondertan (approved):

api/gateway/share.go
rootulp: I think celestia-node API should continue to offer an endpoint that returns unparsed shares so that power users can interpret the raw share data however they choose to.

I think typical users will use endpoints that parse the share data back into share sequences and then the blob data that they published to a namespace.

[ParseShares](https://github.com/celestiaorg/celestia-app/blob/222bb1e0a6013efe44e910d7328c6d141fa99d1a/pkg/shares/share_merging.go#L103) uses the sequence start indicator that is written to every share to determine which chunks of shares are part of the same share sequence.

The [RawData](https://github.com/celestiaorg/celestia-app/blob/222bb1e0a6013efe44e910d7328c6d141fa99d1a/pkg/shares/share_merging.go#L145-L147) method on a share sequence extracts the actual raw data. 

Happy to dive deeper over a call. I don't think a new API endpoint is needed at the moment but will create a follow-up issue for us to discuss.

rootulp: Created https://github.com/celestiaorg/celestia-node/issues/1670

rootulp: Discussed synchronously and in this context it's appropriate to think of a share sequence as a blob. Since a namespace can contain multiple blobs, it contains multiple share sequences.

Issue #1531, opened 2022-12-22
feat(header|p2p): add blacklisting to peerTracker


vgonkivs:
PeerTracker should have the ability to block malicious peers.

PR #1532, opened 2022-12-22
feat(header|p2p): add blacklisting to peerTracker

vgonkivs:
## Overview
Resolves https://github.com/celestiaorg/celestia-node/issues/1531

## Checklist

### DIFF
--- a/header/p2p/peer_tracker.go
+++ b/header/p2p/peer_tracker.go
@@ -9,13 +9,15 @@ import (
 	"github.com/libp2p/go-libp2p-core/host"
 	"github.com/libp2p/go-libp2p-core/network"
 	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/libp2p/go-libp2p/p2p/net/conngater"
 )
 
 // gcCycle defines the duration after which the peerTracker starts removing peers.
 var gcCycle = time.Minute * 30
 
 type peerTracker struct {
-	host host.Host
+	host      host.Host
+	connGater *conngater.BasicConnectionGater
 
 	peerLk sync.RWMutex
 	// trackedPeers contains active peers that we can request to.
@@ -43,6 +45,7 @@ type peerTracker struct {
 
 func newPeerTracker(
 	h host.Host,
+	connGater *conngater.BasicConnectionGater,
 	maxAwaitingTime time.Duration,
 	defaultScore float32,
 	maxPeerTrackerSize int,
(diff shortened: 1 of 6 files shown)


Wondertan (approved):

header/p2p/peer_tracker.go
```diff
@@ -190,3 +194,25 @@ func (p *peerTracker) stop() {
...
+	// add peer to the blacklist, so we can't connect to it in the future.
+	err := p.connGater.BlockPeer(pID)
+	if err != nil {
+		log.Errorw("header/p2p: blocking peer failed", "err", err, "pID", pID)
+	}
+	// close connections to peer.
+	err = p.host.Network().ClosePeer(pID)
+	if err != nil {
+		log.Errorw("header/p2p: closing connection with peer failed", "err", err, "pID", pID)
+	}
```
renaynay, line 209: small nit but i prefer to log err last, so format should be

some msg, pid, err

header/p2p/peer_tracker.go
```diff
@@ -190,3 +194,25 @@ func (p *peerTracker) stop() {
...
+
+	log.Warnw("header/p2p: blocked peer", "pID", pID)
```
renaynay, line 211: would be good to log reason too but we can do in separate PR

nodebuilder/header/module_test.go
```diff
@@ -60,6 +61,9 @@ func TestConstructModule_ExchangeParams(t *testing.T) {
 			return datastore.NewMapDatastore()
 		}),
 		ConstructModule(node.Light, &cfg),
+		fx.Provide(func(b datastore.Batching) (*conngater.BasicConnectionGater, error) {
+			return conngater.NewBasicConnectionGater(b)
+		}),
```
renaynay, line 66: we already provide a connection gater to fx in the p2p pkg, we don't need this.

renaynay (changes requested):

renaynay (approved):

header/p2p/peer_tracker.go
renaynay: we can make an issue for it since it would break method, lets just keep retry to retry

vgonkivs: ok, sure

header/p2p/peer_tracker.go
vgonkivs: Will do in a separate PR

PR #1539, opened 2022-12-22
refactor(nodebuilder/p2p): Add errors for all methods

renaynay:
Based on #1429 

Adds an error as a return value for all methods as it is needed for elevated permission methods.


Wondertan:
Seems to include auth commands

renaynay:
yes @Wondertan it's based on authcmd forgot to mention

Wondertan:
Should be in draft then

Wondertan (approved):

distractedm1nd (approved):

liamsi (approved):

PR #1545, opened 2022-12-23
fix(share): order proof nodes by depth before returning

walldiss:
## Overview

Fix the order of inclusion proof nodes returned by `GetSharesByNamespace` to be possible incorrect.

## Checklist

### DIFF
--- a/share/ipld/proof.go
+++ b/share/ipld/proof.go
@@ -1,11 +1,9 @@
 package ipld
 
 import (
+	"github.com/ipfs/go-cid"
 	"math"
 	"sort"
-	"sync"
-
-	"github.com/ipfs/go-cid"
 )
 
 // Proof contains information required for Leaves inclusion validation.
@@ -17,8 +15,7 @@ type Proof struct {
 // proofCollector collects proof nodes' CIDs for the construction of a shares inclusion validation
 // nmt.Proof.
 type proofCollector struct {
-	leftMu, rightMu sync.Mutex
-	left, right     []node
+	left, right []node
 }
 
 type node struct {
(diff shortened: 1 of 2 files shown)


share/ipld/proof.go
```diff
@@ -15,37 +15,60 @@ type Proof struct {
...
+		cid:   cid,
+		depth: depth,
+	}
 }
 
 // Nodes returns nodes collected by proofCollector in the order that nmt.Proof validator will use
 // to traverse the tree.
 func (c *proofCollector) Nodes() []cid.Cid {
-	nodes := make([]cid.Cid, 0, len(c.left)+len(c.left))
+	cids := make([]cid.Cid, 0, len(c.left)+len(c.right))
```
renaynay, line 52: oof how did we miss this

share/ipld/proof.go
```diff
@@ -15,37 +15,60 @@ type Proof struct {
...
-		right: make([]cid.Cid, 0, height),
+		left:  make([]node, height),
+		right: make([]node, height),
 	}
 }
 
-func (c *proofCollector) addLeft(node cid.Cid) {
-	c.left = append(c.left, node)
+func (c *proofCollector) addLeft(cid cid.Cid, depth int) {
+	c.left[depth] = node{
```
renaynay, line 36: why do you need to put in depth index if you're gonna sort anyway?

share/ipld/proof.go
Wondertan: left and right are always the same size, check constructor

share/ipld/proof.go
walldiss: Depth is needed to make inserts lock free. Not sure if that what you meant, but I've removed sorts, since slices already sorted.

share/ipld/proof.go
```diff
@@ -20,32 +20,39 @@ type proofCollector struct {
...
+			cids = append(cids, cid)
+		}
+	}
 
 	// right side of the tree will be traversed from top to bottom,
 	// so sort in reversed order
 	for i := len(c.right) - 1; i >= 0; i-- {
-		nodes = append(nodes, c.right[i])
+		cid := c.right[i]
+		if cid.ByteLen() != 0 {
```
Wondertan, line 53: last nit: use Defined

Wondertan (approved):
A very nice solution in the end

distractedm1nd (approved):

renaynay (approved):

Issue #1399, opened 2022-11-26
[BUG] header/sync: Repeat sync on same header, running light node on Arabica.


zklim:
### Celestia Node version

Semantic version: v0.5.0-rc5 Commit: 039facd94104c741f8fa995e80f1321776ae4ad3 Build Date: Sat Nov 26 11:37:54 +08 2022 System version: amd64/linux Golang version: go1.19.1

### OS

Windows Subsystem for Linux Distributions: Ubuntu-22.04

### Install tools

Installation follow official docs https://docs.celestia.org/nodes/light-node/

### Steps to reproduce it

Just start running light node and the warning pops up sometime.

### Expected result

No warning

### Actual result

Warning of sync attempt to an already synced header.
![Capture](https://user-images.githubusercontent.com/53001910/204080730-2f6ba11d-93f6-4c05-9d47-942979c4cdc1.PNG)

### Relevant log output

```Shell
2022-11-26T14:16:22.032+0800    WARN    header/sync     sync/sync.go:173        sync attempt to an already synced header        {"synced_height": 36471, "attempted_height": 36471}
2022-11-26T14:16:22.032+0800    WARN    header/sync     sync/sync.go:177        PLEASE REPORT THIS AS A BUG
```

renaynay:
Thank you for reporting @zklim

zklim:
Recently did not see warning of repeat sync but getting new warning now.

![Capture](https://user-images.githubusercontent.com/53001910/204508574-87460a91-bd27-4e81-94e1-02fe397589c0.PNG)

renaynay:
Likely to have been resolved with `syncedHead` implementation -- will reopen if we see it again.

PR #1550, opened 2022-12-26
bugfix(header|sync): clear range after syncing

vgonkivs:
## Overview
Resolves https://github.com/celestiaorg/celestia-node/issues/1399
I think the root cause of this issue is in keeping synced ranges in pending cache. Unfortunately, the log is not full, so this is just an assumption. 
Storing the new head failed after receiving "new" network head with Height ... 
// incomingNetHead processes new gossiped network headers.
func (s *Syncer) incomingNetHead(ctx context.Context, netHead *header.ExtendedHeader) pubsub.ValidationResult {
```
	// Try to short-circuit netHead with append. If not adjacent/from future - try it as new network
	// header
	_, err := s.store.Append(ctx, netHead)
	if err == nil {
		// a happy case where we appended maybe head directly, so accept
	}
```
So, we fall into this condition(log level is debug so we don't see any logs in the output):
```
if errors.As(err, &nonAdj) {
		// not adjacent, maybe we've missed a few headers or its from the past
		log.Debugw("attempted to append non-adjacent header", "store head",
			nonAdj.Head, "attempted", nonAdj.Attempted)
}
```
We are calling `subjectiveHead` inside the `validate` function, which returns `head`:
// validate checks validity of the given header against the subjective head.
```
func (s *Syncer) validate(ctx context.Context, new *header.ExtendedHeader) pubsub.ValidationResult {
	sbjHead, err := s.subjectiveHead(ctx)
	if err != nil {
		log.Errorw("getting subjective head during validation", "err", err)
		return pubsub.ValidationIgnore // local error, so ignore
	}
```
As we haven't cleared any ranges before, it seems that we received `s.pending.Head` As this header is from the past - all other conditions might be finished successfully.
```
	// pending head is the latest known subjective head Syncer syncs to, so try to get it
	// NOTES:
	// * Empty when no sync is in progress
	// * Pending cannot be expired, guaranteed
	pendHead := s.pending.Head()
	if pendHead != nil {
		return pendHead, nil
	}
```
so we add our old-new head to pending and started explicit syncing:

```
// and if valid, set it as new subjective head
	s.pending.Add(netHead)
	s.wantSync()
	log.Infow("new network head", "height", netHead.Height, "hash", netHead.Hash())

```
My suggestion is to remove pending range once we have synced it.
## Checklist

### DIFF
--- a/header/sync/ranges.go
+++ b/header/sync/ranges.go
@@ -25,6 +25,9 @@ func (rs *ranges) Head() *header.ExtendedHeader {
 	}
 
 	head := rs.ranges[ln-1]
+	if head == nil {
+		return nil
+	}
 	return head.Head()
 }
 
@@ -89,11 +92,20 @@ func (rs *ranges) First() (*headerRange, bool) {
 		}
 
 		out := rs.ranges[0]
-		if !out.Empty() {
+		if out != nil && !out.Empty() {
 			return out, true
 		}
 
 		rs.ranges = rs.ranges[1:]
+	}
+}
+
+// Clear removes the first range of headers.
+func (rs *ranges) Clear() {
+	rs.lk.Lock()
+	defer rs.lk.Unlock()
+	if len(rs.ranges) > 0 {
+		rs.ranges[0] = nil
 	}
 }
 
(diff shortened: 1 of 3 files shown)


vgonkivs:
Unit tests fail due to a new logic. Will fix them

vgonkivs:
After more deeper look into Syncer, I have found how do we clear pending ranges. So, my previous analysis is no longer valid. Will continue investigating this issue. But, for now, this PR will be closed.

PR #1551, opened 2022-12-27
test(core|state|swamp): unify and refactor core/app testing utilities

Wondertan:
This PR is one of the cleaning PRs that I am doing. It removes the usage of KVStore. Besides simplifying and unifying tests, it also fixes all the existing and potential "bind already in used" errors from the app and tendermint tandem. I tested this by running all the tests with `-parallel 12`, which runs in parallel per pkg. With this flag, we can potentially speed up the time it takes to run all the tests on Ci and locally. Going further, we can also consider running all the tests in parallel with only one instance of the app running, rather than an instance per test

Would love this diff to be reddish, but unfortunately, I had to copy a bunch of code from Cosmos SDK. The rationale is in the comments. Also, see [this](https://github.com/cosmos/cosmos-sdk/issues/14429).


distractedm1nd (approved):
Much cleaner, looks great

renaynay (changes requested):
Pls rename `grpc` file to `grpc_testing` or something to indicate that this functionality is for testing only. Otherwise lgtm and thank you for this :)

renaynay (approved):

PR #1552, opened 2022-12-27
fix(header/p2p): preconnect trusted peers on header-ex start

Wondertan:
We need to connect to trusted peers and kick off various protocols to start talking to them. In the past, a node was always connected to trusted peers to get headers from them. As from recent improvements to header-ex, we now only rely on trusted peers for initialization. Hence, if a node is already initialized, it won't contact go to connect to trusted peers and this commit fixes this.

This was found during fixing `no trusted peers` issue in swamp tests in https://github.com/celestiaorg/celestia-node/pull/1556

### DIFF
--- a/header/p2p/peer_tracker.go
+++ b/header/p2p/peer_tracker.go
@@ -71,8 +71,8 @@ func (p *peerTracker) track() {
 	}()
 
 	// store peers that have been already connected
-	for _, peer := range p.host.Peerstore().Peers() {
-		p.connected(peer)
+	for _, c := range p.host.Network().Conns() {
+		p.connected(c.RemotePeer())
 	}
 
 	subs, err := p.host.EventBus().Subscribe(&event.EvtPeerConnectednessChanged{})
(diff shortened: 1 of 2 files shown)


header/p2p/peer_tracker.go
```diff
@@ -71,8 +71,8 @@ func (p *peerTracker) track() {
 	}()
 
 	// store peers that have been already connected
-	for _, peer := range p.host.Peerstore().Peers() {
-		p.connected(peer)
```
Wondertan, line 75: * We have in memory peerstore, which is reset on which restart. This was only useful to add trusted peers, preset there during header module construction. Otherwise, it's useless for us.
* Also, we need this change because the logic of  `p.connected` assumes it's called only once for a peer. If we have trusted peers in the peer store and then connect, it will execute connected twice on the same peer.

vgonkivs (approved):

distractedm1nd (approved):

renaynay (approved):

PR #1555, opened 2022-12-27
tests(swamp): correct fraud syncing test

Wondertan:
Previously it succeeded, but the FP was actually transmitted through PubSub and not fraud syncing.
It also makes the test discovery independent.

PR #1556, opened 2022-12-27
feat(swamp): preinit all swamp nodes with genesis header

Wondertan:
Fixes `no trusted peers` failure on main in Swamp tests

HeaderEx recently has started to sync from any peers, and the only place where trusted peers are required now is initialization. This change does initialization for nodes during instantiation in swap completely avoiding the requirement for trusted peers in the node's config.

As a side effect, we won't cover initialization anymore in swamp tests, so we should write some single test to specifically this.

Keeping it in draft until PRs below are merged, as they fix them: 
* https://github.com/celestiaorg/celestia-node/pull/1552
* https://github.com/celestiaorg/celestia-node/pull/1555
* https://github.com/celestiaorg/celestia-node/pull/1554

### DIFF
--- a/nodebuilder/tests/swamp/swamp.go
+++ b/nodebuilder/tests/swamp/swamp.go
@@ -8,6 +8,7 @@ import (
 	"testing"
 	"time"
 
+	mdutils "github.com/ipfs/go-merkledag/test"
 	"github.com/libp2p/go-libp2p/core/host"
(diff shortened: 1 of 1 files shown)


Wondertan:
This PR makes all the integration tests pass locally.

renaynay (approved):

Bidon15 (approved):

distractedm1nd (approved):

Issue #1466, opened 2022-12-12
das: Integrate `ShrEx/Sub` using `eds.PubSub`


Wondertan:
## Context

The `das.DASer` engine keeps the chain sampled and availability checked and provided. Currently, it relies on
`header.Subscribtion` to get every newly produced header from the network. Once a header is received, it can reliably 
check its availability with `share.Availability` interface. However, as we are changing the protocol, the usage of 
`share.Availability` has to be synchronized with EDS notifications from our peers via `eds.PubSub`.

## New Flow

To synchronize them we:
* Cache headers coming from `header.Subscribtion`, instead of calling `SharesAvailable` with `DataHash` as the key.
* Add validator to `eds.PubSub` which
  * Gets the header out of the cache by `DataHash` message body
    * If not found, waits for it within the configurable timeout
    * If nothing, rejects the message
  * Calls `SharesAvailable` with the DAH from the header and the peer the message received from
> I feel the code in the end might become clumsy, but we will fine the way to improve it 

## Open Problem

If a valid header arrives from a subscription, but the corresponding `DataHash` is never received. The above flow does not handle that, but DASer must be able to track this case.  One of the solutions is to inspect the header's cache periodically.

PR #1578, opened 2023-01-09
feat(share/shrex): Peer management for shrex.Getter

walldiss:
## Overview
Closes https://github.com/celestiaorg/celestia-node/issues/1466

TODO:
   - add metrics, tracing

## Checklist


share/p2p/peers/manager.go
```diff
@@ -0,0 +1,225 @@
...
+	pools     map[hashStr]syncPool
+	fullNodes pool
+
+	cancel context.CancelFunc
+	done   chan struct{}
+}
+
+type hashStr string
+
+type syncPool struct {
```
distractedm1nd, line 45: I would move syncPool to pool.go, and create a `newSyncPool` constructor to clean up `getOrCreateUnsyncedPool`

share/p2p/peers/manager.go
```diff
@@ -0,0 +1,225 @@
...
+
+	m         *sync.Mutex
+	pools     map[hashStr]syncPool
+	fullNodes pool
+
+	cancel context.CancelFunc
+	done   chan struct{}
+}
+
+type hashStr string
```
distractedm1nd, line 43: I think we should just use string, and call it dataHash/hash when used as a parameter

share/p2p/peers/manager.go
```diff
@@ -117,11 +118,12 @@ func (s *Manager) Get(ctx context.Context, datahash string) (peer.ID, error) {
...
 }
 
-func (s *Manager) Remove(datahash string, ids ...peer.ID) {
-	p := s.markSynced(hashStr(datahash))
+func (s *Manager) RemovePeers(datahash string, ids ...peer.ID) {
+	p := s.getOrCreateValidatedPool(hashStr(datahash))
 	p.pool.remove(ids...)
 }
 
+// subscribeHeader subscribes to headerSub and
```
distractedm1nd, line 126: and what

share/p2p/peers/pool.go
```diff
@@ -0,0 +1,137 @@
...
+
+func (p *pool) checkHasPeers() {
+	if p.aliveCount > 0 && !p.hasPeer {
+		p.hasPeer = true
+		close(p.hasPeerCh)
+		return
+	}
+
+	if p.aliveCount == 0 && p.hasPeer {
+		p.hasPeerCh = make(chan struct{})
```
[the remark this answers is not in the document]
walldiss: No. `checkHasPeers() ` call is always mutex protected

share/p2p/peers/manager.go
walldiss: I like the second suggestion more. If caller always has access to share.DataHash directly, that the way to go.

share/p2p/peers/manager.go
walldiss: syncPool is also contructed in 2 places with completly opposite preset:
 - getOrCreateValidatedPool
 - getOrCreateUnvalidatedPool
 I'll be happy if you can suggest how to make a common constructor for both cases

walldiss:
I've came up with solution that would wait for sampling before message Validator unlock. The code needs some cleanup and tests update. The idea could summarised as:
 - There is going to be only one validator blocked(first one). All subsequent message validation will short-circuit with validation.Ignore. This will allow only one message to be retransmitted. 
 - Validator will block and wait for 2 things: 
    - header Sync. There is an internal timeout for it.
    - sampling to be done. No internal timeout as for now (we are sure this is valid header).
 - After 2 of those done, validator will be unlocked and retransmit the message.  
 - Successful sampling should be indicated by Getter. When Getter requests a peer, it also  gets `DoneFunc` returned. This func should be called after call to the peer is finished. It will allow Peer manager to know whether sampling is done or not. Peer manager on its own doesn’t do any Sampling and relies on Getter to provide sufficient info about sampling results.

renaynay (commented):
* `header/subscription.go` (there is a typo)

quick pass

share/p2p/peers/manager.go
```diff
@@ -0,0 +1,239 @@
...
+				return
+			}
+			s.fullNodes.remove(peerID)
+		})
+
+	err := shrexSub.AddValidator(s.validate)
+	return s, err
+}
+
+func (s *Manager) Start() {
```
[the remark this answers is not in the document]
renaynay: unless shrex getter will manage manager lifecycle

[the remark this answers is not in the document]
distractedm1nd: shrex getter **should** manage manager lifecycle imo

share/p2p/peers/pool.go
```diff
@@ -0,0 +1,150 @@
...
+	defer p.m.Unlock()
+
+	if p.activeCount == 0 {
+		return "", false
+	}
+
+	start := p.next
+	for {
+		peerID := p.peersList[p.next]
+		if p.next++; p.next == len(p.peersList) {
```
distractedm1nd, line 45: can you split this into two lines please

share/p2p/peers/manager.go
```diff
@@ -0,0 +1,239 @@
...
+// the message via shrex.Sub
+func (s *Manager) sampled(datahash share.DataHash) {
+	p := s.getOrCreateValidatedPool(datahash.String())
+	if p.isSampled.CompareAndSwap(false, true) {
+		close(p.waitSamplingCh)
+	}
+}
+
+// RemovePeers removes peers for given datahash from store
+func (s *Manager) RemovePeers(datahash share.DataHash, ids ...peer.ID) {
```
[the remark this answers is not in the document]
distractedm1nd: no

share/p2p/peers/manager.go
```diff
@@ -15,6 +15,12 @@ import (
 	"github.com/celestiaorg/celestia-node/share/availability/discovery"
 )
 
+const (
+	ResultSuccess sampleResult = iota
+	ResultFail
+	ResultPeerMMisbehaved
```
renaynay, line 21: do we have to expose these consts?

share/p2p/peers/manager.go
```diff
@@ -75,7 +81,9 @@ func (s *Manager) Stop(ctx context.Context) error {
 	}
 }
 
-type DoneFunc func(success bool)
+type DoneFunc func(result sampleResult)
```
renaynay, line 84: does this have to be exposed?

share/p2p/peers/manager.go
renaynay, line 21: There is type, should be `ResultPeerMisbehaved`

share/p2p/peers/manager.go
```diff
@@ -103,17 +111,21 @@ func (s *Manager) GetPeer(
...
 func (s *Manager) doneFunc(datahash share.DataHash, peerID peer.ID) DoneFunc {
-	return func(success bool) {
-		if success {
+	return func(result sampleResult) {
+		switch result {
+		case ResultSuccess:
 			s.markSampled(datahash)
-			return
+		case ResultFail:
+		case ResultPeerMMisbehaved:
```
renaynay, line 124: We shouldn't reject from Validator if failed to serve, we should just ignore rn (but we should figure out how this would impact connectivity / retrieval before we arrive at a decision)

share/p2p/peers/manager.go
distractedm1nd: yes unless we make a shrexgetter package where both this and shrexgetter live

share/p2p/peers/manager.go
distractedm1nd: thats the point of ResultFail being a noop

share/p2p/peers/manager.go
distractedm1nd: yes unless we make a shrexgetter package where both this and shrexgetter live

share/p2p/peers/manager.go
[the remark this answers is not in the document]
distractedm1nd: yes, but now with donefunc it can be private. also we may end up putting this in one package together with shrexgetter

share/p2p/peers/manager.go
```diff
@@ -0,0 +1,234 @@
...
+			s.deletePool(hash.String())
+			return pubsub.ValidationReject
+		}
+	}
+
+	// headerSub found corresponding ExtendedHeader for dataHash,
+	// wait for it to be sampled, before retransmission
+	if sampled := p.waitSampling(ctx); sampled {
+		return pubsub.ValidationAccept
+	}
```
Wondertan, line 208: doc nit: The use of sampling terminology here is not correct. Sampling is a concern of dasing; here, we only ensure the data is loaded/synced/fetched in it's entirety, even though we all understand what we mean here by sampling.

Wondertan (commented):
First in-depth review. I wish we could decouple Discovery and fullNodes pool from Manager, s.t:
* fullNodes pool does not  duplicate the `limitedSet` in Discovery
* we avoid sampling results and return a done function in Peers

There is a way we can do it, but all this improvements and simplification refactorings are time sinks that we cannot afford at this point. The logic here is correct and should work for the goal defined in Blocksync Overhaul PR.

share/p2p/peers/pool.go
```diff
@@ -0,0 +1,152 @@
...
+		if p.next == start {
+			return "", false
+		}
+	}
+}
+
+// waitNext waits for any peer to become available in the pool and sends it to the channel.
+func (p *pool) waitNext(ctx context.Context) <-chan peer.ID {
+	peerCh := make(chan peer.ID, 1)
+	go func() {
```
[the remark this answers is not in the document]
walldiss: We want to wait in async manner here. Channel is returned as a promise, so next available peer wiil be sent to it.

share/p2p/peers/manager.go
```diff
@@ -0,0 +1,244 @@
...
+			}
+			log.Errorw("get next header from sub", "err", err)
+			continue
+		}
+
+		s.getOrCreateValidatedPool(h.DataHash.String())
+	}
+}
+
+func (s *Manager) getOrCreateValidatedPool(datahash string) *syncPool {
```
distractedm1nd, line 175:
```suggestion
// getOrCreateValidatedPool gets/creates a validated pool from the given datahash.
// A pool is considered validated when its datahash has been received over headersub.
func (s *Manager) getOrCreateValidatedPool(datahash string) *syncPool {
```

share/p2p/peers/manager.go
```diff
@@ -0,0 +1,244 @@
...
+	// wait for it to be synced, before retransmission
+	if synced := p.WaitSync(ctx); synced {
+		// block with given datahash was synced, allow the pubsub to retransmit the message by
+		// returning Accept
+		return pubsub.ValidationAccept
+	}
+	return pubsub.ValidationIgnore
+}
+
+func (s *Manager) getOrCreateUnvalidatedPool(datahash string) *syncPool {
```
distractedm1nd, line 222:
```suggestion
// getOrCreateUnvalidatedPool gets/creates an unvalidated pool from the given datahash.
// A pool is considered unvalidated when its datahash has been received over shrexsub,
// but has not yet been seen on headersub.
func (s *Manager) getOrCreateUnvalidatedPool(datahash string) *syncPool {
```

share/p2p/peers/pool.go
```diff
@@ -0,0 +1,153 @@
...
+	m           sync.Mutex
+	peersList   []peer.ID
+	active      map[peer.ID]bool
+	activeCount int
+	next        int
+
+	hasPeer   bool
+	hasPeerCh chan struct{}
+
+	cleanupDisabled bool
```
distractedm1nd, line 21: let's remove this, it should always be enabled

share/p2p/peers/pool.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	// do cleanup if too much garbage
+	if len(p.peersList) > p.activeCount*2 && !p.cleanupDisabled {
+		p.cleanup()
+	}
+	p.checkHasPeers()
+}
+
+// cleanup will reduce memory footprint of pool.
```
distractedm1nd, line 118:
```suggestion
// cleanup will reduce the memory footprint of pool.
```

share/p2p/peers/pool.go
```diff
@@ -0,0 +1,153 @@
...
+// newPool creates new pool
+func newPool() *pool {
+	return &pool{
+		peersList: make([]peer.ID, 0),
+		active:    make(map[peer.ID]bool),
+		hasPeerCh: make(chan struct{}),
+	}
+}
+
+// tryGet returns peer along with bool flag indicating success of operation.
```
distractedm1nd, line 33:
```suggestion
// tryGet returns a peer along with bool flag indicating success of operation.
```

share/p2p/peers/sync_pool.go
```diff
@@ -0,0 +1,60 @@
...
+func (p *syncPool) waitValidation(ctx context.Context) (valid bool) {
+	select {
+	case <-p.validatorWaitCh:
+		return p.isValidDataHash.Load()
+	case <-ctx.Done():
+		return false
+	}
+}
+
+func (p *syncPool) WaitSync(ctx context.Context) (synced bool) {
```
distractedm1nd, line 43: this can be private

share/p2p/peers/pool.go
```diff
@@ -0,0 +1,153 @@
...
+				continue
+			}
+			p.next = len(newList)
+		}
+	}
+	p.peersList = newList
+}
+
+// checkHasPeers will check and indicate if there are peers in the pool.
+func (p *pool) checkHasPeers() {
```
distractedm1nd, line 142: Just a random idea i had, and is completely optional:
I think I like the name of `peerAvailable` more than `hasPeer`, and `checkHasPeer` could be `setPeerAvailable`

distractedm1nd (approved):
The logic is ready and we should merge this asap, my changes are optional so I will approve

share/p2p/peers/manager.go
```diff
@@ -0,0 +1,239 @@
...
+		})
+
+	err := shrexSub.AddValidator(s.validate)
+	return s, err
+}
+
+func (s *Manager) Start() {
+	ctx, cancel := context.WithCancel(context.Background())
+	s.cancel = cancel
+	go s.disc.EnsurePeers(ctx)
```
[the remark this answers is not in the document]
Wondertan: Where bridge nodes will run advertisements then?

share/p2p/peers/manager.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+func (s *Manager) doneFunc(datahash share.DataHash, peerID peer.ID) DoneFunc {
+	return func(result syncResult) {
+		switch result {
+		case ResultSuccess:
+			s.markSynced(datahash)
+		case ResultFail:
+		case ResultPeerMisbehaved:
+			// TODO: signal to peers Validator to return Reject
```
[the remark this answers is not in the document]
walldiss: This is not implemented. Currently Reject resolution (peer punishment) is given only on messages that has invalid datahash. Ideally Reject should also be given to messages from peers that misbehaved. But as for now resolution is the same for all messages for given datahash. Not a critical issue, so left as TODO here

share/p2p/peers/pool.go
walldiss: Need it for tests

distractedm1nd:
closing in favor of #1707

Issue #1465, opened 2022-12-12
node/share: Data sync modes for FN


Wondertan:
## Context

We need a configuration to run an FN in reconstruction-only mode, so we can continue benchmarking reconstruction performance and apply various obvious optimizations coming from __happy path__ integration.

This should be done by configuring the node to run without `ShrExGetter`

PR #1581, opened 2023-01-09
feat(share/getter): CascadeGetter

### LINKED ISSUE #1491: share/get: `CascadeGetter`
## Overview

`CascadeGetter` implements `Getter` that composes multiple `Getters`. The composition is ordered with priority.
For every `Getter` operation, the first registered `Getter` is used. If it fails with an error or timeout from a configured
deadline controlled by context, the following `Getter` is used, and so forth, until one of them succeeds.

`CascadeGetter` must not verify date integrity and leave it up to the composing implementations.

Wondertan:
This PR implements `CascadeGetter`. The logic is described on the issue and documented on the cascade method. The auxiliary `SingleEDSGetter` is also implemented for testing purposes but can be used elsewhere. The `cascade` function is implemented nicely using generics and can be extracted into the utils repo for later reusal. 

Closes #1491 
Based on #1518 

###  TODO 
* [x] Finish godocing
* [x] Unify tests for cascade func
* [x] ~Logging~ Tracing
* [x] ~Move cascade out to uitls repo?~


Wondertan:
The imports were changed by `goimports`

distractedm1nd:
If you rebase on main there is a getters utils file you can move to

share/getters/cascade.go
```diff
@@ -0,0 +1,151 @@
...
+			return get(ctx, getter)
+		})
+	}
+
+	return cascade[V](ctx, fns, interval)
+}
+
+// cascade implements a cascading retry algorithm for getting a value from multiple sources
+// sorted by priority. Cascading implies trying the sources one-by-one in the given order with the
+// given interval until either:
```
distractedm1nd, line 80: Can you please add that the given timeout interval does not cancel the last source, but starts the next one and results can still come from the previous source?

share/getters/cascade.go
```diff
@@ -0,0 +1,151 @@
...
+			case results <- struct {
+				val V
+				err error
+			}{val: val, err: err}:
+			case <-ctx.Done():
+			}
+		}(src)
+	}
+
+	for i := 0; i < len(srcs)-len(errs); i++ {
```
distractedm1nd, line 137: please add comment explaining why this is necessary

share/getters/cascade.go
```diff
@@ -0,0 +1,151 @@
...
+
+		t.Reset(interval)
+		go func(src func(context.Context) (V, error)) {
+			val, err := src(ctx)
+			select {
+			case results <- struct {
+				val V
+				err error
+			}{val: val, err: err}:
+			case <-ctx.Done():
```
distractedm1nd, line 132: add comment explaining why ctx error isnt passed up

share/getters/single_eds.go
```diff
@@ -0,0 +1,51 @@
...
+	"context"
+	"fmt"
+
+	"github.com/celestiaorg/celestia-app/pkg/da"
+	"github.com/celestiaorg/celestia-node/share"
+	"github.com/celestiaorg/nmt/namespace"
+	"github.com/celestiaorg/rsmt2d"
+)
+
+// SingleEDSGetter contains single EDS and allows getting data only out of it.
```
distractedm1nd, line 13:
```suggestion
// SingleEDSGetter contains a single EDS where data is retrieved from.
// Its primary use is testing, and GetSharesByNamespace is not supported.
```

distractedm1nd (changes requested):
just comment/doc nits right now

share/getters/cascade.go
Wondertan: I don' think there is a need to comment on such things. It's just a part of the logic

share/getters/cascade.go
```diff
@@ -0,0 +1,153 @@
...
+		err error
+	})
+
+	for _, src := range srcs {
+		select {
+		case res := <-results:
+			if res.err == nil {
+				return res.val, nil
+			}
+			errs = append(errs, res.err)
```
walldiss, line 120:
```
// If a program has already received a value from t.C, the timer is known
// to have expired and the channel drained, so t.Reset can be used directly.
// If a program has not yet received a value from t.C, however,
// the timer must be stopped and—if Stop reports that the timer expired
// before being stopped—the channel explicitly drained:
```
```suggestion
			errs = append(errs, res.err)
			if !t.Stop() {
		            <-t.C
		        }
```

share/getters/cascade.go
walldiss: (Optional). If you like it, same could be done by:
```for i := len(errs); i < len(srcs); i++```

File: share/getters/cascade.go [hunk 6]
```diff
@@ -0,0 +1,153 @@
...
+
+	for i := 0; i < len(srcs)-len(errs); i++ {
+		select {
+		case res := <-results:
+			if res.err == nil {
+				return res.val, nil
+			}
+			errs = append(errs, res.err)
+		case <-ctx.Done():
+			return zero, ctx.Err()
```
walldiss, line 147: This should rather return combined errors with an addition to ctx.Err. Or we will lose errors from failed getter.

share/getters/single_eds.go
```diff
@@ -0,0 +1,51 @@
...
+type SingleEDSGetter struct {
```
walldiss, line 14: Is it going to be used outside of testing purposes? If no, I would vote for creating subpkg `getterstest` to reduce import surface and API of main getters pkg

Wondertan: Very likely not to be used anywhere else but made reusable not only in testing by purpose. Agree thought that we should move it elsewhere and we should keep this post-implementation chores.

share/getters/cascade.go
Wondertan: When user cancels the context it does not expect other errors there

share/getters/cascade.go
Wondertan: I like it, seems more readable

share/getters/cascade.go
Wondertan: Actually, this is implied by `All of the sources errors`,does not hurt to clarify that though

share/getters/cascade.go
Wondertan: But I can mansplain for you :)
Because the loop will anyway catch the context error in the main cascade routine

share/getters/cascade.go
Wondertan: added the comment

share/getters/cascade.go
Wondertan: Thanks!. Good catch.

share/getters/cascade.go
walldiss: It could be a context deadline error. BTW same should also apply for another ctx.Done() case.

renaynay (commented):
* why not put single eds getter in testing.go? lets do that or rename file to `testgetter` or something?
* let's also come to decision together re: async spawning getters.

share/getters/cascade.go
```diff
@@ -0,0 +1,160 @@
...
+// cascade implements a cascading retry algorithm for getting a value from multiple sources.
+// Cascading implies trying the sources one-by-one in the given order with the
+// given interval until either:
+//   - One of the sources returns the value
+//   - All of the sources errors
+//   - Context is canceled
+//
+// NOTE: New source attempts after interval do not suspend running sources in progress.
+// TODO(@Wondertan): Move to utils
+func cascade[V any](ctx context.Context, srcs []func(context.Context) (V, error), interval time.Duration) (V, error) {
```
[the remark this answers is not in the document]
Wondertan: LGTM

Wondertan:
> let's also come to decision together re: async spawning getters.

I am still in favor of increasing the likelihood of data being retrieved and async getters giving this to us. FWIW, it's the main purpose of cel-node. The only point against that is that it _may_ degrade debuggability, but the tracing I've just added should mitigate this.

renaynay:
> I am still in favor of increasing the likelihood of data being retrieved and async getters giving this to us. FWIW, it's the main purpose of cel-node. The only point against that is that it _may_ degrade debuggability, but the tracing I've just added should mitigate this.

I am not against increasing likelihood of data being retrieved **eventually** -- these are a couple lines of code we can add in the future once we have a more stable understanding of how our new architecture works. Please, let's take it one step at a time and not hold this PR up over this.

renaynay:
Ok we (@Wondertan) agreed to leave the code as is and to disable the IPLD fallback case

Wondertan:
> disable the IPLD fallback case

When we want to test `ShrexGetter` or happy sync path in isolation 

https://github.com/celestiaorg/celestia-node/issues/1465

renaynay:
> > disable the IPLD fallback case
> 
> When we want to test `ShrexGetter` or happy sync path in isolation
> 
> #1465

which we agreed would be the **default** case for now

share/getters/cascade.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+}
+
+// GetShare gets a share from any of registered share.Getters in cascading order.
+func (cg *CascadeGetter) GetShare(ctx context.Context, root *share.Root, row, col int) (share.Share, error) {
+	ctx, span := cascadeTracer.Start(ctx, "get-share", trace.WithAttributes(
+		attribute.String("root", root.String()),
+		attribute.Int("row", row),
+		attribute.Int("col", col),
+	))
```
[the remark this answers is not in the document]
Wondertan: Good q and I think so, so this data is going to be duplicated

Wondertan:
@renaynay, moved

distractedm1nd:
I agree with Rene here. While I do think the added complexity may very well be worth it in the future, the stack of getters has to be as simple as possible in its first iteration in order to both debug properly and iterate easily, because there **will** be bugs, and there **will** be things we learn, such as unexpected bottlenecks. 

Yes, the main purpose of cel-node is to maximize the likelihood of data being retrieved, but maximizing performance incrementally becomes more and more difficult with stacked complexity - if we can't fit the whole stack in our head (and it is tending in this direction!) it will become extremely difficult to optimize and reason over.

I'm not saying this approach isn't objectively better, just that it is too complex right now and I am uncomfortable adding it this early, especially since it is adding onto many layers of abstract wrapping of these getters, all of which is both new code and a new architecture. I of course also don't want to block this for too long, we should discuss this synchronously in tomorrow's standup and then decide based on there :)

Wondertan:
Keeping it draft until we return to this

share/getters/cascade.go
walldiss: If user set timeout and it happen, it will be impossible to know how many getters were tried and if there getter were failed by a time deadline happen. Multierr would still satisfy context.Deadline error, while providing more info to the caller why particular call failed in multiple getters.

Wondertan:
@walldiss, we put this PR on a long pause and went with a simpler version https://github.com/celestiaorg/celestia-node/pull/1628. It's reviewed and merged at this point. Still pls take a look at it.

share/getters/cascade.go
```diff
@@ -0,0 +1,207 @@
...
+	if len(srcs) == 1 {
+		return srcs[0](ctx)
+	}
+	// once we got value from one of the fns
+	// this cancels all others on return
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	// start from zero so the first iteration is instantaneous
+	// later on in the loop below we reset to real interval
+	t := time.NewTimer(0)
```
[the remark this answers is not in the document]
Wondertan: That might work. Will take a deeper look once we return to this PR

share/getters/cascade.go
```diff
@@ -0,0 +1,207 @@
...
+		getters:  getters,
+	}
+}
+
+// GetShare gets a share from any of registered share.Getters in cascading order.
+func (cg *CascadeGetter) GetShare(ctx context.Context, root *share.Root, row, col int) (share.Share, error) {
+	ctx, span := cascadeTracer.Start(ctx, "get-share", trace.WithAttributes(
+		attribute.String("root", root.String()),
+		attribute.Int("row", row),
+		attribute.Int("col", col),
```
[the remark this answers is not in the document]
Wondertan: Hmm:
* Each Getter would have its span anyway, with a custom name describing the implementation, like [here](https://github.com/celestiaorg/celestia-node/pull/1628/files#diff-830b9c493d28c2431bda3668469f2d55db8e9112f3211c1f12f264570baf4780R41-R44). In this PR we have a personal tracer for cascade getter which would mark it as so, so wouldn't be hard to debug.
* I am not a fan of extending interfaces with statically returned values. We could make it through anonymous interface(check if a Getter implement Name() - use it), but I still don't see why according to the point above

share/getters/cascade.go
```diff
@@ -0,0 +1,207 @@
...
+	))
+	defer func() {
+		defer span.End()
+		if len(errs) == len(srcs) {
+			// we do not set the actual errors to the description, as they were already recorded
+			span.SetStatus(codes.Error, "all sources failed")
+			return
+		}
+
+		span.SetStatus(codes.Ok, "")
```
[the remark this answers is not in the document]
Wondertan: Will take a look once we get back to this

share/getters/cascade.go
Wondertan: IIRC, we argued in sync when we decided to return one only error, though I agree it does not hurt to return, even though I believe it's pointless

share/getters/cascade.go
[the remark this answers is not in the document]
walldiss: Getter spans don't have info of getter name. In the provided pr, there is arbitary getter_idx attribute, that could be confusing to identify what getter implementation was actually in use. Wrapping span or log with getter name could see great benefit for debugging.

It could be also to use extended interface like.
```
type NamedGetter interface{
    Name() string
    Getter
}

Or even log/trace wrap getter... 
But thats unnecessary overcomplicating things, when adding extra static method is super easy.

Wondertan:
Closing for, we will come to it once we there is a strong reason for async Cascading

PR #1583, opened 2023-01-10
chore(swamp): replace cosmos prefix with celestia in swamp tests

vgonkivs:
## Overview

Use "celestia" prefix for accounts instead of "cosmos" in swamp tests

## Checklist


Wondertan:
Does it fixes tests on mac? Also the question why didn't it work before

nodebuilder/tests/swamp/swamp.go
```diff
@@ -31,6 +34,13 @@ import (
 	"github.com/celestiaorg/celestia-node/nodebuilder/state"
 )
 
+func init() {
+	cfg := sdk.GetConfig()
+	cfg.SetBech32PrefixForAccount(app.Bech32PrefixAccAddr, app.Bech32PrefixAccPub)
+	cfg.SetBech32PrefixForValidator(app.Bech32PrefixValAddr, app.Bech32PrefixValPub)
+	cfg.Seal()
+}
```
Wondertan, line 42: Let's move this init to the core pkg. This way we always use the right prefix everywhere including swamp. Also if the node is going to be used as a library, then users won't need to do this as well.

Wondertan (changes requested):

nodebuilder/tests/swamp/swamp.go
vgonkivs: we are already using it here https://github.com/celestiaorg/celestia-node/blob/311d9dc757f716289978da23913a72fe01c99ede/cmd/celestia/main.go#L18-L21. So, we have to add these lines only for swamp tests

distractedm1nd (dismissed):

renaynay (approved):

core/fetcher.go
```diff
@@ -17,6 +20,15 @@ var (
...
 )
 
+func init() {
+	// This is necessary to ensure that the account addresses are correctly prefixed
+	// as in the celestia application.
+	cfg := sdk.GetConfig()
+	cfg.SetBech32PrefixForAccount(app.Bech32PrefixAccAddr, app.Bech32PrefixAccPub)
+	cfg.SetBech32PrefixForValidator(app.Bech32PrefixValAddr, app.Bech32PrefixValPub)
+	cfg.Seal()
+}
```
Wondertan, line 30: Can we make a separate file for this? Fetcher is not a good place.

distractedm1nd (approved):

Wondertan (approved):

renaynay (approved):

Issue #1489, opened 2022-12-15
share/get: `ShrExGetter`


Wondertan:
## Overview

`ShrExGetter` implements `Getter` over `ShrEx/ND` and `ShrEx/EDS` and their respective clients. It relies on `Discovery`
to get data providers/peers. `ShrExGetter` must verify data integrity itself or by relying on underlying protocol clients.

```go
// GetShare is no-op.
func (shrg *ShrExGetter) GetShare(context.Context, *Root, row, col int) (share.Share, error)

// GetEDS fetches all the shares in the form of EDS committed to the given root over `ShrEx/EDS`.
func (shrg *ShrExGetter) GetEDS(context.Context, *Root) (*rsmt2d.EDS, error)

// GetShareByNamespace fetches all the namespaced shares committed to the given root over `ShrEx/ND`.
func (shrg *ShrExGetter) GetShareByNamespace(context.Context, *Root, namespace.ID) (share.NamespaceShares, error)

// WithProvider hints ShrExGetter to fetch data from the given peers instead of Discovery.
func WithProvider(context.Context, ...peer.ID) ctx
```

distractedm1nd:
Closed by #1624

PR #1591, opened 2023-01-13
refactor(shrexeds): removing retry logic from client

distractedm1nd:
## Overview
This PR removes the retry logic from the shrex/eds client and adds test skipping to test cases that can either be refactored or removed when implementing #1489. This logic is being removed so that ShrexGetter can handle retry/round-robin logic for both the shrex/eds and shrex/nd clients.

## Checklist

### DIFF
--- a/share/p2p/shrexeds/client.go
+++ b/share/p2p/shrexeds/client.go
@@ -42,52 +42,34 @@ func NewClient(host host.Host, opts ...Option) (*Client, error) {
 	}, nil
 }
 
-// RequestEDS requests the full ODS from one of the given peers and returns the EDS.
-//
-// The peers are requested in a round-robin manner with retries until one of them gives a valid
-// response, blocking until the context is canceled or a valid response is given.
+// RequestEDS requests the ODS from the given peers and returns the EDS upon success.
 func (c *Client) RequestEDS(
 	ctx context.Context,
 	dataHash share.DataHash,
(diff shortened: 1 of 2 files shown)


share/p2p/shrexeds/client.go
```diff
@@ -42,52 +42,34 @@ func NewClient(host host.Host, opts ...Option) (*Client, error) {
...
 
-	// requests are retried for every peer until a valid response is received
-	excludedPeers := make(map[peer.ID]struct{})
-	for {
-		// if no peers are left, return
-		if len(peers) == len(excludedPeers) {
-			return nil, errNoMorePeers
+	for _, to := range peers {
+		eds, err := c.doRequest(ctx, req, to)
+		if eds != nil {
```
walldiss, line 55: why not `err != nil`? Is it possible that non-nil error will be returned with valid eds?

distractedm1nd: because the error cases are covered below

walldiss: Why not `err == nil`? I know it is usually either data or non-nil error returned, but if it is not a case, logic will be wrong.

distractedm1nd: Discussed in sync, we return `nil, nil` sometimes so `if eds != nil`  is necessary

renaynay (approved):

PR #1592, opened 2023-01-13
fix(header/p2p): set deadline on a request to peer

Wondertan:


header/p2p/session.go
```diff
@@ -130,14 +131,13 @@ func (s *session) doRequest(
...
+	defer cancel()
+
 	r, size, duration, err := sendMessage(ctx, s.host, stat.peerID, s.protocolID, req)
 	if err != nil {
-		if err == context.Canceled || err == context.DeadlineExceeded {
-			return
-		}
 		log.Errorw("requesting headers from peer failed.", "failed peer", stat.peerID, "err", err)
 		select {
-		case <-ctx.Done():
```
vgonkivs, line 140: why? we will request headers from another peer. Previous one, will not be added

Wondertan: because then request won't be retried!

Wondertan: ctx.Done succeds and the `case s.reqCh <- req` does not do anything

vgonkivs: yes, correctly. Sorry, I was blind on Friday

Wondertan:
Closing in favour of https://github.com/celestiaorg/celestia-node/pull/1603

PR #1589, opened 2023-01-12
feat(header/p2p): add traces to server

vgonkivs:
## Overview
Added traces to the headerEx.Server

## Checklist

- [x] New and updated code has appropriate documentation
- [x] New and updated code has new and/or updated testing
- [x] Required CI checks are passing
- [x] Visual proof for any user facing features like CLI or documentation updates
- [x] Linked issues closed with keywords

PR #1594, opened 2023-01-16
feat(share/eds): adding traces to edsstore, write/read eds, and getters

### LINKED ISSUE #1361: share: metrics and traces for `EDSStore`

distractedm1nd:
## Overview
Closes #1361

### DIFF
--- a/share/eds/eds.go
+++ b/share/eds/eds.go
@@ -17,6 +17,7 @@ import (
 	"github.com/ipld/go-car"
 	"github.com/ipld/go-car/util"
 	"github.com/minio/sha256-simd"
+	"go.opentelemetry.io/otel/codes"
 
 	"github.com/celestiaorg/celestia-app/pkg/appconsts"
 	"github.com/celestiaorg/celestia-app/pkg/da"
@@ -45,9 +46,13 @@ type writingSession struct {
 // Order: [ Carv1Header | Q1 |  Q2 | Q3 | Q4 | inner nodes ]
 // For more information about the header: https://ipld.io/specs/transport/car/carv1/#header
 func WriteEDS(ctx context.Context, eds *rsmt2d.ExtendedDataSquare, w io.Writer) error {
+	ctx, span := tracer.Start(ctx, "write-eds")
+	defer span.End()
+
 	// 1. Reimport EDS. This is needed to traverse the NMT tree and cache the inner nodes (proofs)
 	writer, err := initializeWriter(ctx, eds, w)
 	if err != nil {
+		span.SetStatus(codes.Error, err.Error())
 		return fmt.Errorf("share: creating eds writer: %w", err)
 	}
 
@@ -55,18 +60,21 @@ func WriteEDS(ctx context.Context, eds *rsmt2d.ExtendedDataSquare, w io.Writer) 
 	//    - Roots are the eds Row + Col roots
 	err = writer.writeHeader()
 	if err != nil {
+		span.SetStatus(codes.Error, err.Error())
 		return fmt.Errorf("share: writing carv1 header: %w", err)
 	}
 
 	// 3. Iterates over shares in quadrant order via eds.GetCell
 	err = writer.writeQuadrants()
 	if err != nil {
+		span.SetStatus(codes.Error, err.Error())
 		return fmt.Errorf("share: writing shares: %w", err)
 	}
 
 	// 4. Iterates over in-memory blockstore and writes proofs to the CAR
 	err = writer.writeProofs(ctx)
 	if err != nil {
+		span.SetStatus(codes.Error, err.Error())
 		return fmt.Errorf("share: writing proofs: %w", err)
 	}
 	return nil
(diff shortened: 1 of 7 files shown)


share/eds/eds.go
```diff
@@ -45,28 +46,35 @@ type writingSession struct {
...
 // For more information about the header: https://ipld.io/specs/transport/car/carv1/#header
 func WriteEDS(ctx context.Context, eds *rsmt2d.ExtendedDataSquare, w io.Writer) error {
+	ctx, span := tracer.Start(ctx, "write-eds")
+	defer span.End()
+
 	// 1. Reimport EDS. This is needed to traverse the NMT tree and cache the inner nodes (proofs)
 	writer, err := initializeWriter(ctx, eds, w)
 	if err != nil {
+		span.SetStatus(codes.Error, err.Error())
 		return fmt.Errorf("share: creating eds writer: %w", err)
```
Wondertan, line 56: * Writing the error we compose here should be more readable.
* Try defining errors at the top of the methods and then do one defer that sets the status if the error is not nil. 
* Also, there is a `RecordError` API that specifically takes errors, and let's use it instead of passing the error string to the status. Note that `RecordError` still requires setting the status.

share/eds/store.go
```diff
@@ -148,14 +151,19 @@ func (s *Store) gc(ctx context.Context) {
 // The resulting file stores all the shares and NMT Merkle Proofs of the EDS.
 // Additionally, the file gets indexed s.t. store.Blockstore can access them.
 func (s *Store) Put(ctx context.Context, root share.DataHash, square *rsmt2d.ExtendedDataSquare) error {
+	ctx, span := tracer.Start(ctx, "store/put", trace.WithAttributes(attribute.String("root", root.String())))
```
Wondertan, line 154: Square size would also be useful to track here

share/eds/eds.go
```diff
@@ -45,28 +46,35 @@ type writingSession struct {
 // Order: [ Carv1Header | Q1 |  Q2 | Q3 | Q4 | inner nodes ]
 // For more information about the header: https://ipld.io/specs/transport/car/carv1/#header
 func WriteEDS(ctx context.Context, eds *rsmt2d.ExtendedDataSquare, w io.Writer) error {
+	ctx, span := tracer.Start(ctx, "write-eds")
```
derrandz, line 49: i suggest that we put span names in constants

share/eds/eds.go
```diff
@@ -45,28 +46,35 @@ type writingSession struct {
...
+	defer span.End()
+
 	// 1. Reimport EDS. This is needed to traverse the NMT tree and cache the inner nodes (proofs)
 	writer, err := initializeWriter(ctx, eds, w)
 	if err != nil {
+		span.SetStatus(codes.Error, err.Error())
```
derrandz, line 55: is `span.RecordError` not required to be called in here? Check #1589

share/eds/eds.go
derrandz: Applies to the rest of the spans

derrandz (changes requested):
A few nits and we are good to go

share/eds/eds.go
distractedm1nd: This would hurt readability and is optimized by the compiler to do this anyways

share/eds/eds.go
Wondertan: It's not

renaynay (approved):

share/eds/store.go
```diff
@@ -150,20 +150,21 @@ func (s *Store) gc(ctx context.Context) {
...
 // Additionally, the file gets indexed s.t. store.Blockstore can access them.
-func (s *Store) Put(ctx context.Context, root share.DataHash, square *rsmt2d.ExtendedDataSquare) error {
-	ctx, span := tracer.Start(ctx, "store/put", trace.WithAttributes(attribute.String("root", root.String())))
-	defer span.End()
+func (s *Store) Put(ctx context.Context, root share.DataHash, square *rsmt2d.ExtendedDataSquare) (err error) {
+	ctx, span := tracer.Start(ctx, "store/put", trace.WithAttributes(
+		attribute.String("root", root.String()),
+		attribute.Int("width", int(square.Width())),
+	))
+	defer utils.SetStatusAndEnd(span, err)
```
Wondertan, line 158: I really love the idea here, and my eyes truly opened when I saw this. 
However, this will always be nil because it passes the value at the moment of the defer call.

Here is a snippet to play around https://go.dev/play/p/i2yVr-bkIjy. 

The fix is simple though https://go.dev/play/p/jSA8Kora7xS

Wondertan (changes requested):

share/eds/store.go
distractedm1nd: Ah yes makes sense! Thanks for catching this

renaynay (approved):

Wondertan (approved):

renaynay (approved):

vgonkivs (approved):

derrandz (approved):

PR #1609, opened 2023-01-17
feat(share/discovery): Add callback on peers set updates.

walldiss:
## Overview

Notify Discovery clients about peers set changes in async manner by calling provided callback.

`(ps *limitedSet)Peers` method splitted into two funcs: `ListPeers` and `WaitPeer`, so callback is not locked in case there are no peers.

Closes https://github.com/celestiaorg/celestia-node/issues/1610

## Checklist

### DIFF
--- a/share/availability/discovery/discovery.go
+++ b/share/availability/discovery/discovery.go
@@ -71,6 +71,7 @@ func NewDiscovery(
 	}
 }
 
+// WithOnPeersUpdate adds OnPeersUpdate callback call on every update of discovered peers list.
 func (d *Discovery) WithOnPeersUpdate(f OnPeersUpdate) {
 	d.onPeersUpdate = func(info updateInfo) {
 		d.onPeersUpdate(info)
@@ -200,15 +201,5 @@ func (d *Discovery) Advertise(ctx context.Context) {
 // Peers provides a list of discovered peers in the "full" topic.
 // If Discovery hasn't found any peers, it blocks until at least one peer is found.
 func (d *Discovery) Peers(ctx context.Context) ([]peer.ID, error) {
-	peers := d.set.ListPeers()
-	if len(peers) > 0 {
-		return peers, nil
-	}
-
-	// block until a new peer will be discovered
-	p, err := d.set.WaitPeer(ctx)
-	if err != nil {
-		return nil, err
-	}
-	return []peer.ID{p}, nil
+	return d.set.Peers(ctx)
 }
--- a/share/availability/discovery/discovery.go
+++ b/share/availability/discovery/discovery.go
@@ -39,11 +39,11 @@ type Discovery struct {
 	discoveryInterval time.Duration
 	// advertiseInterval is an interval between advertising sessions.
 	advertiseInterval time.Duration
-	// onPeersUpdate will be called on peer set changes
-	onPeersUpdate OnPeersUpdate
+	// onUpdatedPeers will be called on peer set changes
+	onUpdatedPeers OnUpdatedPeers
 }
 
-type OnPeersUpdate func(peerID peer.ID, isAdded bool)
+type OnUpdatedPeers func(peerID peer.ID, isAdded bool)
 
 // NewDiscovery constructs a new discovery.
 func NewDiscovery(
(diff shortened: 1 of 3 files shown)


share/availability/discovery/discovery.go
```diff
@@ -173,5 +193,15 @@ func (d *Discovery) Advertise(ctx context.Context) {
...
+	peers := d.set.ListPeers()
+	if len(peers) > 0 {
+		return peers, nil
+	}
+
+	// block until a new peer will be discovered
+	p, err := d.set.WaitPeer(ctx)
+	if err != nil {
+		return nil, err
+	}
```
Wondertan, line 205: Makes sense to seperate it for `d.onPeersUpdate(d.set.ListPeers())`

share/availability/discovery/discovery.go
```diff
@@ -57,6 +61,19 @@ func NewDiscovery(
 		peersLimit,
 		discInterval,
 		advertiseInterval,
+		nil,
+	}
+}
+
+func (d *Discovery) WithOnPeersUpdate(f OnPeersUpdate) {
```
Wondertan, line 68: I would make it via a channel to protect `OnPeersUpdate` from blocking `handlePeerFound` (API hygiene), but this is also fine

Wondertan (dismissed):

share/availability/discovery/discovery.go
Wondertan: Some basic test would hurt

share/availability/discovery/discovery.go
```diff
@@ -39,8 +39,12 @@ type Discovery struct {
 	discoveryInterval time.Duration
 	// advertiseInterval is an interval between advertising sessions.
 	advertiseInterval time.Duration
+	// onPeersUpdate will be called on peers set changes
```
distractedm1nd, line 42:
```suggestion
	// onPeersUpdate will be called on peer set changes
```

share/availability/discovery/set.go
```diff
@@ -74,24 +74,23 @@ func (ps *limitedSet) Remove(id peer.ID) {
...
-		for p := range ps.ps {
-			out = append(out, p)
-		}
-		ps.lk.Unlock()
-		return out, nil
-	}
-	ps.lk.Unlock()
-
-	// block until a new peer will be discovered
+// WaitPeer blocks until first peer discovered.
```
distractedm1nd, line 77:
```suggestion
// WaitPeer blocks until a peer is discovered.
```

distractedm1nd (changes requested):
One question: why send all peers through `onPeersUpdate` since it's called in `handlePeerFound`? Can't we just name it `onPeerFound`/`OnPeerFound` and only send in the new peer? And I agree with hlib that it might be better with a channel, if you agree (but really optional)

Also needs a test

share/availability/discovery/discovery.go
```diff
@@ -57,6 +61,19 @@ func NewDiscovery(
...
+}
+
+func (d *Discovery) WithOnPeersUpdate(f OnPeersUpdate) {
+	if d.onPeersUpdate == nil {
+		d.onPeersUpdate = f
+		return
+	}
+
+	d.onPeersUpdate = func(peers []peer.ID) {
+		d.onPeersUpdate(peers)
```
vgonkivs, line 75: why do we need to call `d.onPeersUpdate(peers)` inside assigning to `d.onPeersUpdate`?

share/availability/discovery/discovery.go
vgonkivs, line 68: could you please add a comment, that the current option is extending(and not replacing) the default callback with the custom one?

walldiss: done

walldiss:
Updated with @distractedm1nd suggestion to only emit diff, not the full peers list. This removes some changes in set.go and in tests. As a note: callback provider will needs to be careful with keeping state up to date so it don't diverge from discovery one.

distractedm1nd (changes requested):
I __think__ I would prefer the callback taking two args instead of a struct, but that is entirely up to you

distractedm1nd (approved):
beautiful.

vgonkivs (approved):

PR #1611, opened 2023-01-17
refactor(share/p2p): ironing out differences between shrex protocols

distractedm1nd:
## Overview
This PR mainly changes the error handling in both protocols to be more consistent with each other, and prepare them for handing errors up to `ShrexGetter`. It also removes variadic peer input to client requests in favor of a single peer, and makes the shrexeds protobuf enum identifiers follow the same pattern as shrexnd.

## Checklist


share/p2p/shrexeds/pb/extended_data_square.proto
```diff
@@ -6,9 +6,9 @@ message EDSRequest {
 
 enum Status {
   INVALID = 0;
-  OK = 100; // data found
-  NOT_FOUND = 200; // data not found
-  REFUSED = 201; // request refused
+  OK = 1; // data found
+  NOT_FOUND = 2; // data not found
+  REFUSED = 3; // request refused
```
Wondertan, line 11: Ok. I'll update the ADR afterwards

share/p2p/shrexnd/server.go
```diff
@@ -32,7 +32,7 @@ type Server struct {
 }
 
 // NewServer creates new Server
-func NewServer(host host.Host, store *eds.Store, getter *getters.IPLDGetter, opts ...Option) (*Server, error) {
+func NewServer(host host.Host, store *eds.Store, getter share.Getter, opts ...Option) (*Server, error) {
```
Wondertan, line 35: Just a note that we should make a proxy constructor for this during integration that takes an actual implementation rather than an interface, so we don't accidentally end up with a Server that requests data instead of returning not found. Honestly, keeping the specific implementation there feels safer

share/p2p/errors.go
```diff
@@ -0,0 +1,34 @@
...
+		return ctx.Err()
+	}
+	var ne net.Error
+	if errors.As(err, &ne) && ne.Timeout() {
+		if deadline, _ := ctx.Deadline(); deadline.Before(time.Now()) {
+			return context.DeadlineExceeded
+		}
+	}
+	return nil
+}
```
Wondertan, line 34: This is a cute deduplication function, but it's a bit early for it. We could duplicate more code in a func like `SendMessage` and `SendRequest`, which would cover such error handling, stream resets, and further. We wanted to do this deduplication after ITN and let's remove this for now.

Wondertan (commented):
Two notes and one request. Looks good otherwise

share/p2p/shrexnd/server.go
distractedm1nd: causes a cyclical dependency when shrexgetter is introduced, so proxy is better here for now

renaynay (approved):

share/p2p/shrexeds/client.go
```diff
@@ -102,28 +91,28 @@ func (c *Client) doRequest(
...
 		log.Debugf("client: peer %s couldn't serve eds %s with status %s", to.String(), dataHash.String(), resp.GetStatus())
-		// no eds was returned, but the request was valid and should be retried
-		return nil, nil
-	case p2p_pb.Status_INVALID:
+		return nil, p2p.ErrUnavailable
+	case pb.Status_INVALID:
 		fallthrough
 	default:
-		return nil, fmt.Errorf("request status %s returned for root %s", resp.GetStatus(), dataHash.String())
+		log.Errorf("unknown request status %s returned for root %s", resp.Status.String(), dataHash.String())
```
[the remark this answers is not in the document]
distractedm1nd: good catch

share/p2p/shrexnd/client.go
```diff
@@ -55,32 +53,19 @@ func (c *Client) GetSharesByNamespace(
...
+	if err == nil {
+		return shares, err
 	}
-
-	return nil, errNoMorePeers
+	if p2p.ExtractContextError(ctx, err) != nil {
+		return nil, err
+	}
+	if err != p2p.ErrUnavailable {
+		log.Errorw("client-nd: peer returned err", "peer_id", peer.String(), "err", err)
```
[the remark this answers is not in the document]
distractedm1nd: done, but made it the same as shex/eds

share/p2p/shrexnd/client.go
```diff
@@ -197,12 +182,12 @@ func statusToErr(code pb.StatusCode) error {
...
-		pb.StatusCode_NOT_FOUND,
-		pb.StatusCode_INTERNAL,
-		pb.StatusCode_REFUSED:
+	case pb.StatusCode_NOT_FOUND, pb.StatusCode_REFUSED:
+		return p2p.ErrUnavailable
+	case pb.StatusCode_INTERNAL, pb.StatusCode_INVALID:
+		fallthrough
 	default:
-		code = pb.StatusCode_INVALID
+		log.Errorf("unknown request status %s returned", code.String())
```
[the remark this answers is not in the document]
distractedm1nd: done

Wondertan (approved):

renaynay (dismissed):

renaynay (approved):

Wondertan (approved):

Issue #1438, opened 2022-12-06
fix(share): removing eds.Store reliance on `FS`


distractedm1nd:
We should use[FS](https://pkg.go.dev/io/fs#FS) and pass it to the EDS store instead of relying directly on the `os.Open`. This way, we can easily pass testing FS impl and enable EDSStore to be tested in memory.

ef3n9r98:
Hi @distractedm1nd, it's great to see you have active repos with many good first issues! I know some devs who'd be interested in working on this. Is there a budget for fixing any of your issues, and if so, how much? 

I'd put this issue up on [OpenQ](http://www.openq.dev/).

Thanks!

Jpserrat:
Hey guys, can I work on this?

distractedm1nd:
@Jpserrat Yes! Feel free to reach out to me for guidance/any questions

Jpserrat:
@distractedm1nd Thanks for your quick response on this and sorry for the delay on my side. I already did the PR but I have some questions. 
Running make test I got some failing tests with error message:  NotFound desc = account cosmos1hh846ae9gqs9p00hdkpp4gv37ma9y7q7ysuljt not found (Not related to this issue)

Do you know what I can do?

Jpserrat:
@distractedm1nd I noticed that I got it all wrong, I now understand that it is only the FS interface, but I'm not sure where exactly in the eds store it's supposed to be replaced

distractedm1nd:
@Jpserrat Hey thanks for working on this! Basically we don't want to pass a path, but an FS interface to the EDS Store constructor - I took a look at what you did and its very close to what it should be I think, just without the dependency and using the FS interface instead. I think this should be possible but tell me if you run into any blockers

Jpserrat:
@distractedm1nd Did the changes, thanks a lot for your patience and help

distractedm1nd:
@Wondertan What is the in-memory `fs.FS` implementation you had in mind to use in testing?

distractedm1nd:
I see no reason to need to test the eds Store in memory. We have a good test suite and it works perfectly fine with tempdir in the filesystem, which we need anyways in swamp etc. Abstracting this makes it more complicated for no reason, so closing.

PR #1622, opened 2023-01-19
share/eds: replace OS use with fs

Jpserrat:
## Overview

This pr relates to #1438.
I've replaced the use of OS from eds/store to use fs. I've also changed the tests to use fs impl.

@distractedm1nd is this what you were thinking? I wasn't able to remove the use of tmpDir completely because the dagstore use os call, let me know if there is anything that I can do better 

## Checklist

### DIFF
--- a/share/eds/store.go
+++ b/share/eds/store.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"sync/atomic"
 	"time"
@@ -30,6 +31,7 @@ import (
 
 const (
 	blocksPath     = "/blocks/"
+	fsBlocksPath   = "blocks"
 	indexPath      = "/index/"
 	transientsPath = "/transients/"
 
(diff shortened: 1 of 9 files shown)


share/eds/store.go
```diff
@@ -53,17 +54,19 @@ type Store struct {
 	gcInterval time.Duration
 	// lastGCResult is only stored on the store for testing purposes.
 	lastGCResult atomic.Pointer[dagstore.GCResult]
+	fs           afero.Fs
 }
 
 // NewStore creates a new EDS Store under the given basepath and datastore.
-func NewStore(basepath string, ds datastore.Batching) (*Store, error) {
-	err := setupPath(basepath)
+func NewStore(basepath string, ds datastore.Batching, aFs afero.Fs) (*Store, error) {
```
Wondertan, line 61: We shouldn’t use the Fs defined in the external dependency. The issue was about usage of the recently introduced FS interface as linked in the case.

Wondertan (changes requested):
That's for the start. Have one major comment

share/eds/store.go
Jpserrat: Now using the FS interface

distractedm1nd:
@Jpserrat Thanks for working on this, but I am closing the original issue and this PR. Testing is fine using the filesystem and tempdir, I don't see any reason to remove the reliance on the filesystem here anymore, this issue was probably opened without enough discussion

PR #1631, opened 2023-01-23
fix(daser) replace catchUpDoneCh channel gracefuly 

walldiss:
DASer `waitCatchUp` method has a data race for catchUpDoneCh pointer. Pointer needs to be swapped gracefully. 
Would occur only if user was fortunate enough to call `waitCatchUp` in the span of time the pointer was being swapped. 
## Checklist

### DIFF
--- a/das/state.go
+++ b/das/state.go
@@ -2,6 +2,7 @@ package das
 
 import (
 	"context"
+	"sync/atomic"
 )
 
 // coordinatorState represents the current state of sampling
@@ -18,7 +19,7 @@ type coordinatorState struct {
 	next        uint64 // all headers before next were sent to workers
 	networkHead uint64
 
-	catchUpDone   bool          // indicates if all headers are sampled
+	catchUpDone   *atomic.Bool  // indicates if all headers are sampled
 	catchUpDoneCh chan struct{} // blocks until all headers are sampled
 }
 
@@ -34,7 +35,7 @@ func newCoordinatorState(params Parameters) coordinatorState {
 		nextJobID:         0,
 		next:              params.SampleFrom,
 		networkHead:       params.SampleFrom,
-		catchUpDone:       false,
+		catchUpDone:       new(atomic.Bool),
 		catchUpDoneCh:     make(chan struct{}),
 	}
 }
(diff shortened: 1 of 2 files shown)


vgonkivs (approved):

das/state.go
```diff
@@ -201,28 +202,31 @@ func (s *coordinatorState) unsafeStats() SamplingStats {
...
 	}
 }
 
 // waitCatchUp waits for sampling process to indicate catchup is done
 func (s *coordinatorState) waitCatchUp(ctx context.Context) error {
+	if s.catchUpDone.Load() {
+		return nil
+	}
 	select {
 	case <-s.catchUpDoneCh:
```
Wondertan, line 231: There is still can be a race condition for `catchUpDoneCh`. Because we rewrite the actual chan.

Wondertan (changes requested):

das/state.go
```diff
@@ -18,7 +19,7 @@ type coordinatorState struct {
 	next        uint64 // all headers before next were sent to workers
 	networkHead uint64
 
-	catchUpDone   bool          // indicates if all headers are sampled
+	catchUpDone   *atomic.Bool  // indicates if all headers are sampled
```
Wondertan, line 22: It's totally ok to make it a value instead.

Wondertan: One less ptr is always good, also we would avoid initing in the constructor

das/state.go
walldiss: It is not possible, because overwrite is by atomic.Bool var. read could be done only if catchUpDone == false. Write could be done only if catchUpDone == true, when read is short-circuit. So to overwrite it, we make sure it is catchUpDone == true, then do overwrite and only then allow reads, by setting catchUpDone.Store(false).

walldiss: On the other hand, closing the channel is concurrently same operation, so we can use CAS instead of 2 operations.

Wondertan: I see, missed the short-circuit part.

Wondertan: Should we store the channel in the atomic value instead?

Wondertan: Then we can avoid a bool fully.

Wondertan: Just a suggestion and this full up to you

Wondertan (approved):

renaynay (approved):

PR #1632, opened 2023-01-23
Core listener put eds

renaynay:
Replaces #1562


nodebuilder/core/module.go
```diff
@@ -35,13 +35,14 @@ func ConstructModule(tp node.Type, cfg *Config, options ...fx.Option) fx.Option
 			fx.Provide(core.NewBlockFetcher),
 			fxutil.ProvideAs(core.NewExchange, new(libhead.Exchange[*header.ExtendedHeader])),
 			fx.Invoke(fx.Annotate(
-				func(bcast libhead.Broadcaster[*header.ExtendedHeader],
+				func(
```
distractedm1nd, line 38: Can you add a comment saying that this is a ConstructFn? This is a place that I have gotten confused before

headertest/testing.go
```diff
@@ -222,13 +294,14 @@ func RandBlockID(t *testing.T) types.BlockID {
 }
 
 // FraudMaker creates a custom ConstructFn that breaks the block at the given height.
-func FraudMaker(t *testing.T, faultHeight int64) header.ConstructFn {
+func FraudMaker(t *testing.T, faultHeight int64, bServ blockservice.BlockService) header.ConstructFn {
 	log.Warn("Corrupting block...", "height", faultHeight)
```
distractedm1nd, line 298: should this warn not actually be inside the returned function? otherwise it will only ever be called once no? (when providing it)

core/listener.go
```diff
@@ -96,11 +97,26 @@ func (cl *Listener) listen(ctx context.Context, sub <-chan *types.Block) {
...
+			}
+			// generate extended header
+			eh, err := cl.construct(ctx, b, comm, vals, eds)
 			if err != nil {
 				log.Errorw("listener: making extended header", "err", err)
 				return
 			}
+			// store block data if not empty
+			if eds != nil {
+				err = cl.store.Put(ctx, eh.DAH.Hash(), eds)
```
distractedm1nd, line 114: do the edses of empty blocks end up here? I'm asking because Put will throw a `dagstore.ShardAlreadyExists` when the same DAH is put into the store twice

core/listener.go
```diff
@@ -22,10 +22,11 @@ import (
 // network.
 type Listener struct {
 	fetcher *BlockFetcher
-	bServ   blockservice.BlockService
 
 	construct header.ConstructFn
 
+	store *eds.Store
```
distractedm1nd, line 28: why is this grouped alone

core/listener.go
renaynay: good q

nodebuilder/core/module.go
renaynay: actually why do we `fx.Invoke` this?

distractedm1nd (approved):
works well in the integration testing branch, looks ready to merge into feature branch

walldiss (approved):

vgonkivs (approved):

PR #1537, opened 2022-12-22
bump(metrics): upgrade otel to the latest version

vgonkivs:
## Overview
Updating otel collector to the latest version
## Checklist

- [x] New and updated code has appropriate documentation
- [x] New and updated code has new and/or updated testing
- [x] Required CI checks are passing
- [x] Visual proof for any user facing features like CLI or documentation updates
- [x] Linked issues closed with keywords

PR #1376, opened 2022-11-16
feat(telemetry): add blackbox instrumentation to the header module + share module + p2p bandwidth metrics

derrandz:
## Overview

To enable blackbox instrumentation for our benchmarking tests, we need to enrich the celestia-node codebase with blackbox metrics, measurements that the benchmark will rely on to count the final results.

In this PR, we ship blackbox instrumentation for the HeaderService + ShareService in a well defined pattern that relies on:

1. Proxying the existing service
2. Decorating the injected service with the proxied service (`fx`)

This pattern will allow blackbox instrumentation to be:

1. Optional
2. Enable-able by using `WithBlackBoxMetrics`

## Changes

- [x] Implement `blackboxInstrument` for the `HeaderService`
- [x] Implement `blackboxInstrument` for the `ShareService`
- [x] libp2p bandwidth metrics using open telemetry

## Checklist

- [x] New and updated code has appropriate documentation
- [x] New and updated code has new and/or updated testing
- [x] Required CI checks are passing
- [x] Visual proof for any user facing features like CLI or documentation updates
- [x] Linked issues closed with keywords

PR #1638, opened 2023-01-25
feat(node | das | libs/header/sync): add total uptime node metrics + totalSampled das metrics + totalSynced sync metrics

derrandz:
## Overview

This PR introduces node uptime metrics + das total sampled headers metrics to support calculating the uptime index proposed by mustafa on the monitoring side.

This PR introduces a new module named `Telemetry` to support node related telemetry. This module can also host all general telemetry and observability that does not interest specific modules.

## Changes

- [x] Introduced uptime metrics for node under `nodebuilder/node/uptime.go`
- [x] Introduced persistent uptime metrics using datastore to persist node start time
- [x] Testing for uptime metrics persistence using the store
- [x] Unit testing for uptime metrics
- [x] Integration testing for uptime metrics
- [ ] e2e testing for uptime metrics

## Checklist

## Blocked By

PR: #1537


nodebuilder/node/uptime.go
```diff
@@ -0,0 +1,83 @@
+// This file defines UptimeMetrics relative to the nodebuilder package.
```
renaynay, line 1:
```suggestion
```

nodebuilder/node/uptime.go
```diff
@@ -0,0 +1,83 @@
...
+	"go.opentelemetry.io/otel/metric/instrument/asyncfloat64"
+	"go.opentelemetry.io/otel/metric/instrument/syncfloat64"
+)
+
+// UptimeMetrics is a struct that records
+// 1. node start time: the timestamp when the node was started
+// 2. node up time: total time the node has been running
+//
+// the node start time is recorded imperatively when RecordNodeStartTime is called
+// whereas the node up time is recorded periodically upon callback recalling (re-mettering from optl)
```
renaynay, line 19:
```suggestion
// whereas the node uptime is recorded periodically upon callback recalling (re-metering from optl)
```

nodebuilder/node/uptime.go
```diff
@@ -0,0 +1,83 @@
...
+
+	"go.opentelemetry.io/otel/metric/global"
+	"go.opentelemetry.io/otel/metric/instrument"
+	"go.opentelemetry.io/otel/metric/instrument/asyncfloat64"
+	"go.opentelemetry.io/otel/metric/instrument/syncfloat64"
+)
+
+// UptimeMetrics is a struct that records
+// 1. node start time: the timestamp when the node was started
+// 2. node up time: total time the node has been running
```
renaynay, line 16:
```suggestion
// 2. node uptime: total time the node has been running
```

nodebuilder/node/uptime.go
```diff
@@ -0,0 +1,83 @@
...
+	// lastNodeUptimeTS is the last timestamp when the node uptime was recorded.
+	lastNodeUptimeTS float64
+}
+
+var meter = global.MeterProvider().Meter("node")
+
+// NewUptimeMetrics creates a new UptimeMetrics
+// and registers a callback to re-meter the totalNodeUptime metric.
+func NewUptimeMetrics() (*UptimeMetrics, error) {
+	nodeStartTS, err := meter.
```
mojtaba-esk, line 36: Does this store the node start value in some sort of db and retrieve it if available? because if it doesn't, once the node is restarted, we get a new start time for the node which yields in a wrong uptime calculation.

mojtaba-esk (changes requested):

das/metrics.go
```diff
@@ -76,11 +78,22 @@ func (d *DASer) InitMetrics() error {
 		return err
 	}
 
+	totalSampled, err := meter.
+		SyncInt64().
+		UpDownCounter(
```
walldiss, line 83: Consider using gauge. Counter will start from previous value upon restart, that will break observations

das/metrics.go
```diff
@@ -110,29 +123,43 @@ func (d *DASer) InitMetrics() error {
...
 
+// observeSample records the time it took to sample a header +
+// the amount of sampled contiguous headers
 func (m *metrics) observeSample(ctx context.Context, h *header.ExtendedHeader, sampleTime time.Duration, err error) {
 	if m == nil {
 		return
 	}
 	m.sampleTime.Record(ctx, sampleTime.Seconds(),
 		attribute.Bool("failed", err != nil),
-		attribute.Int("header_width", len(h.DAH.RowsRoots)))
```
walldiss, line 119: Why did you remove attribute?

das/metrics.go
```diff
@@ -110,29 +123,43 @@ func (d *DASer) InitMetrics() error {
...
 	}
 	m.newHead.Add(ctx, 1)
 }
+
+// recordTotalSampled records the total amount of sampled headers.
+func (m *metrics) recordTotalSampled(ctx context.Context, n int64) {
+	if m == nil {
+		return
+	}
+	m.totalSampled.Add(ctx, n)
```
walldiss, line 164: Oh man, the counter could not be used like that, you need gauge. I won't be long until you observe sampled percentage to be > 100%

nodebuilder/node/uptime.go
```diff
@@ -0,0 +1,83 @@
...
+// 2. node up time: total time the node has been running
+//
+// the node start time is recorded imperatively when RecordNodeStartTime is called
+// whereas the node up time is recorded periodically upon callback recalling (re-mettering from optl)
+type UptimeMetrics struct {
+	// nodeStartTS is the timestamp when the node was started.
+	nodeStartTS syncfloat64.UpDownCounter
+
+	// totalNodeUptime is the total time the node has been running.
+	totalNodeUptime asyncfloat64.Gauge
```
walldiss, line 25: This could benefit from using a single counter. Add 1 every second node is running. It will give you run indicater + sum could be used as uptime in seconds later. Gauge is bad here, the value could be calculated by aggregate. 

Start time on the other hand should be a Gauge. Do you even need start time? You can do uptime calculation with it.

das/coordinator.go
```diff
@@ -55,6 +55,12 @@ func newSamplingCoordinator(
 
 func (sc *samplingCoordinator) run(ctx context.Context, cp checkpoint) {
 	sc.state.resumeFromCheckpoint(cp)
+
+	// the amount of sampled headers from the checkpoint
+	// is the point from which we should start sampling - 1
+	totalSampledFromCheckpoint := int64(cp.SampleFrom - 1)
```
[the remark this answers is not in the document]
derrandz: Got it

das/metrics.go
derrandz: a relic

nodebuilder/node/uptime.go
derrandz: Turning this into a counter adding 1 every other second is a great idea! Thank you
On the other hand, I think we need to record the node start time timestamp to achieve the uptime index formula from Mustafa.

nodebuilder/node/uptime.go
derrandz: Good point, now implemented a mechanism to store it in the datastore in the [following commit](https://github.com/celestiaorg/celestia-node/pull/1638/commits/f01f9281292bc8926691f98e330177c33b8a4bda)

derrandz:
We're successfully retrieving metrics at the moment, however we need to think about calculating the uptime index (from mustafa's proposal) somewhere either in node or in mojtaba's API, since Grafana's min function returns a min over a vector and not a minimum of two values.

Here's a screen short of the two values, green being `totalSampled/network_head`, and yellow being `(time-nodeStartTs)/total_uptime_in_s`

Screen Shot 2023-01-27 at 14 56 12

das/coordinator.go
```diff
@@ -76,6 +81,12 @@ func (sc *samplingCoordinator) run(ctx context.Context, cp checkpoint) {
 			}
 		case res := <-sc.resultCh:
 			sc.state.handleResult(res)
+
+			// totalSampledFromWorker is the amount of successfully
+			// sampled headers from the worker
+			totalSampledFromWorker := int64(res.To-res.From) - int64(len(res.failed))
+			sc.metrics.recordTotalSampled(ctx, totalSampledFromWorker)
```
walldiss, line 88: Async metrics could only be observed inside metrics Callback. Consider storing a int64 value inside metrics struct and observe it on callback.

das/checkpoint.go
```diff
@@ -48,3 +48,17 @@ func (c checkpoint) String() string {
...
 }
+
+// WorkersTotalSampled returns the total amount of sampled headers
+func (c checkpoint) WorkersTotalSampled() int {
+	var total int
+	for _, w := range c.Workers {
+		total += int(w.To - w.From)
+	}
+	return total
+}
```
walldiss, line 59: Name of the function doesn't match the logic. You calculating amount of not sampled items, that workers will pick up upon resume. Those headers are not yet sampled.

Also if w.To == w.From there is 1 header to be sampled. So total should be like this:
```
total += int(w.To - w.From) + 1
```

das/checkpoint.go
```diff
@@ -48,3 +48,17 @@ func (c checkpoint) String() string {
...
+
+// TotalSampled returns the total amount of sampled headers
+func (c checkpoint) TotalSampled() int {
+	return int(c.SampleFrom) - 1 + c.WorkersTotalSampled() - len(c.Failed)
+}
```
walldiss, line 64: TotalSampled and WorkersTotalSampled could be non-exported. Also consider making 1 function onstead of 2, there is no need to have them both in API surface.

nodebuilder/node/uptime.go
```diff
@@ -0,0 +1,139 @@
...
+
+	err = meter.RegisterCallback(
+		[]instrument.Asynchronous{
+			totalNodeUptime,
+		},
+		func(ctx context.Context) {
+			m.totalNodeUpTimeTicks++
+			totalNodeUptime.Observe(ctx, float64(m.totalNodeUpTimeTicks))
+		},
+	)
```
walldiss, line 83: Looks like you are observing amount of callback calls here, but not the amount of seconds from node start.

renaynay (commented):
just define the const in `uptime.go`

nodebuilder/node/uptime.go
derrandz: Correct, noticed this while visually testing this. Fixed it, thank you.

das/checkpoint.go
derrandz: I'm slightly confused, what you are saying regarding the "to be sampled" ranges for workers, makes sense, however in your [previous comment](https://github.com/celestiaorg/celestia-node/pull/1638#discussion_r1087757155) you mentioned that `w.To - w.From` would be the all done from workers.

So which is correct, `c.SampleFrom - 1 - allFailed` or should add to these the workers' to and from? I'm thinking the first formula.

walldiss: The comment you mentioned doesn't say w.To - w.From would be the all done. It is items that were not finished, and will be continued after restart by resumed workers.  Thus sum of those needs to be subtracted from current allSampled value.

das/coordinator.go
derrandz: 👍🏻

das/checkpoint.go
derrandz: Thank you for the elaboration @walldis, as you've helped point out through discussion, the correct formula is:

```
totalSampled = c.SampleFrom - TotalInProgress - TotalFailed
```

derrandz:
@mojtaba-esk and I discussed the following [comment](https://github.com/celestiaorg/celestia-node/pull/1638#discussion_r1087714225) and concluded to not store any metrics on store at the node side, and defer this requirement to the parties that require it (_i.e: NodeLogger and Leaderboard backend_)

das/coordinator.go
```diff
@@ -55,6 +55,11 @@ func newSamplingCoordinator(
 
 func (sc *samplingCoordinator) run(ctx context.Context, cp checkpoint) {
 	sc.state.resumeFromCheckpoint(cp)
+
+	// the amount of sampled headers from the last checkpoint
+	totalSampledFromCheckpoint := int64(cp.totalSampled())
+	sc.metrics.recordTotalSampled(ctx, totalSampledFromCheckpoint)
```
walldiss, line 61: Nit: no need for extra variable

das/coordinator.go
walldiss, line 88: resultsCh is working with sampling ranges. There is`observeSample` func, that is called every time block is sampled.  You could get much more granular results by updating totalSampled inside `observeSample`.  Would also cleanup coordinator a bit.

das/metrics.go
```diff
@@ -110,29 +132,46 @@ func (d *DASer) InitMetrics() error {
...
 	m.newHead.Add(ctx, 1)
 }
+
+// recordTotalSampled records the total amount of sampled headers.
+func (m *metrics) recordTotalSampled(ctx context.Context, n int64) {
+	if m == nil {
+		return
+	}
+	totalSampledInt := atomic.LoadInt64(&m.totalSampledInt)
+	atomic.StoreInt64(&m.totalSampledInt, totalSampledInt+n)
```
walldiss, line 176: Two subsequent atomic operation combined does not have transaction property. Please use `atomic.AddInt64(&m.totalSampledInt,1)`

nodebuilder/settings.go
```diff
@@ -83,7 +92,7 @@ func initializeMetrics(
 			exp,
 		),
 		controller.WithExporter(exp),
-		controller.WithCollectPeriod(2*time.Second),
+		controller.WithCollectPeriod(node.OptlCollectPeriodInSeconds*time.Second),
```
walldiss, line 95: collect period affects load on collector and a lot of other things, it should not be set based on one logic of one particular metric. Please remove

nodebuilder/node/uptime.go
```diff
@@ -0,0 +1,90 @@
...
+		totalNodeUpTimeTicks: 1,
+	}
+
+	err = meter.RegisterCallback(
+		[]instrument.Asynchronous{
+			totalNodeRunTime,
+		},
+		func(ctx context.Context) {
+			m.totalNodeUpTimeTicks += OptlCollectPeriodInSeconds
+			totalNodeRunTime.Observe(ctx, float64(m.totalNodeUpTimeTicks))
```
walldiss, line 76: consider storing last observation time and collecting seconds by using something like `time.Since(lastObservationTime).Seconds()`

nodebuilder/settings.go
derrandz: It's the other way around, the metric is depending on this value. So we decided to put this hard-coded value into a constant so that it's accessible elsewhere.

nodebuilder/node/uptime.go
walldiss: You need to increment the counter by amount of seconds passed since last observation (time.Since().Seconds()). You don't need to have any constants for that, nor modifying internals of collector.

nodebuilder/settings.go
walldiss: Lets discuss in another thread, since they are related

derrandz:
Here's a screenshot of the last changes that shows:

1. In green: the time factor of uptime: `(total node runtime in seconds) / (time() - node start time ts)`
2. In yellow: the work factor of uptime `(total headers sampled) / network head`

Screen Shot 2023-01-31 at 01 10 36

nodebuilder/node/metrics.go
```diff
@@ -0,0 +1,59 @@
...
+		)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Observe node start timestamp
+	nodeStartTS.Observe(context.Background(), float64(time.Now().UTC().Unix()))
+
+	// Observe total node run time
+	var totalNodeUpTimeInSeconds int64
```
walldiss, line 46: Looks like this value needs to be init with `time.Now().Unix()` value instead of 0

walldiss (commented):
Just one more small issue and it looks good. Could you please add test for DASer sampled amount metric?

das/metrics.go
```diff
@@ -76,16 +80,31 @@ func (d *DASer) InitMetrics() error {
 		return err
 	}
 
+	totalSampled, err := meter.
+		AsyncInt64().
+		Gauge(
+			"das_total_sampled_headers_gauge",
```
renaynay, line 86:
```suggestion
			total_sampled_headers",
```

das/metrics.go
```diff
@@ -24,7 +25,10 @@ type metrics struct {
 	sampleTime    syncfloat64.Histogram
 	getHeaderTime syncfloat64.Histogram
 	newHead       syncint64.Counter
-	lastSampledTS int64
+	totalSampled  asyncint64.Gauge
```
renaynay, line 28: this looks like it's unused on the struct - why are you storing it? you only need it for the callback func

das/metrics.go
```diff
@@ -110,29 +132,54 @@ func (d *DASer) InitMetrics() error {
...
+// observeNewHead records the network head.
 func (m *metrics) observeNewHead(ctx context.Context) {
 	if m == nil {
 		return
 	}
 	m.newHead.Add(ctx, 1)
 }
+
+// recordTotalSampled records the total sampled headers.
+func (m *metrics) recordTotalSampled(ctx context.Context, totalSampled int) {
```
renaynay, line 180: you don't use the context here, do you need to pass it?

das/options.go
```diff
@@ -38,7 +38,8 @@ type Parameters struct {
 	// SampleFrom is the height sampling will start from if no previous checkpoint was saved
 	SampleFrom uint64
 
-	// SampleTimeout is a maximum amount time sampling of single block may take until it will be canceled
+	// SampleTimeout is a maximum amount time sampling of
+	// single block may take until it will be canceled
```
renaynay, line 42: looks weird formatting

das/worker.go
```diff
@@ -60,8 +60,16 @@ func (w *worker) run(
 		}
 
 		metrics.observeGetHeader(ctx, time.Since(startGet))
-		log.Debugw("got header from header store", "height", h.Height(), "hash", h.Hash(),
-			"square width", len(h.DAH.RowsRoots), "data root", h.DAH.Hash(), "finished (s)", time.Since(startGet))
+
+		// TODO(@team): find linter rule to enforce this
```
renaynay, line 64: there is a linter rule to enforce that a LOC doesn't exceed certain char amount, so you can remove this todo

nodebuilder/node/metrics.go
```diff
@@ -0,0 +1,62 @@
...
+	"go.opentelemetry.io/otel/metric/instrument"
+)
+
+var (
+	meter = global.MeterProvider().Meter("node")
+	log   = logging.Logger("node")
+)
+
+// WithMetrics registers node metrics.
+func WithMetrics() {
```
renaynay, line 20: please return error here instead of doing log.Fatal

nodebuilder/node/metrics.go
```diff
@@ -0,0 +1,62 @@
...
+			"node_runtime_counter_in_seconds",
+			instrument.WithDescription("total time the node has been running"),
+			instrument.WithUnit("seconds"),
+		)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	var started bool = false
+	var totalNodeUpTimeInSeconds int64 // Observe total node run time
```
renaynay, line 43:
```suggestion
	var (
	    started bool
	    totalNodeUptimeInSeconds int64
	 ) 
```

renaynay: and obviously format pls

nodebuilder/node/metrics.go
```diff
@@ -0,0 +1,62 @@
...
+			if !started {
+				// Observe node start timestamp
+				nodeStartTS.Observe(context.Background(), float64(time.Now().UTC().Unix()))
+				started = true
+			}
+
+			now := time.Now().Unix()
+			last := atomic.SwapInt64(&totalNodeUpTimeInSeconds, now)
+
+			totalNodeRunTime.Observe(ctx, float64(now-last))
```
renaynay, line 56: use time.Since

nodebuilder/node/metrics.go
```diff
@@ -0,0 +1,62 @@
...
+		},
+	)
+	if err != nil {
+		log.Fatal(err)
```
renaynay, line 60: return err

renaynay (changes requested):

das/metrics.go
derrandz: 👍🏻

das/worker.go
derrandz: I meant more like, the style of having `"attribute", var` per line than the LOC length. Seems that such rule doesn't exist.

nodebuilder/node/metrics.go
derrandz: We agreed me and @walldiss to rely on simple subtraction for simplicity since both values are just "distances" in time in seconds.

das/metrics.go
derrandz: will keep ftm to make use of dependency injection during test to inject a mock.

derrandz:
Merge is blocked by #1537

mojtaba-esk:
Hello,
I have to observations to report.

The total runtime shows a floating point value:
`curl -s  http://localhost:9093/api/v1/query?query=node_runtime_counter_in_seconds | jq`
```json
{
  "status": "success",
  "data": {
    "resultType": "vector",
    "result": [
      {
        "metric": {
          "__name__": "node_runtime_counter_in_seconds",
          "exported_instance": "12D3KooWMuCkRJfEaae4mQ4HPL5FnNe1zc6q8ASCk9MdvEkK5DUj",
          "exported_job": "Celestia-Bridge",
          "instance": "otel-collector:8889",
          "job": "da_nodes"
        },
        "value": [
          1675688581.167,
          "0.00000006"
        ]
      }
    ]
  }
}
```
The `total_synced` metric is not appearing in my Bridge node instance metrics.
`curl -s  http://localhost:9093/api/v1/label/__name__/values | jq`
```json
{
  "status": "success",
  "data": [
    "badencoding",
    "das_busy_workers_amount",
    "das_get_header_time_hist_bucket",
    "das_get_header_time_hist_count",
    "das_get_header_time_hist_sum",
    "das_head_updated_counter",
    "das_latest_sampled_ts",
    "das_network_head",
    "das_sample_time_hist_bucket",
    "das_sample_time_hist_count",
    "das_sample_time_hist_sum",
    "das_sampled_chain_head",
    "das_sampled_headers_counter",
    "das_total_sampled_headers",
    "das_total_sampled_headers_gauge",
    "head",
    "last_pfb_timestamp",
    "node_runtime_counter_in_seconds",
    "node_start_ts",
    "pfb_count",
    "scrape_duration_seconds",
    "scrape_samples_post_metric_relabeling",
    "scrape_samples_scraped",
    "scrape_series_added",
    "target_info",
    "up"
  ]
}
```

share/p2p/shrexsub/pubsub.go
```diff
@@ -39,7 +39,8 @@ type PubSub struct {
 
 // NewPubSub creates a libp2p.PubSub wrapper.
 func NewPubSub(ctx context.Context, h host.Host, suffix string) (*PubSub, error) {
-	// WithSeenMessagesTTL without duration allows to process all incoming messages(even with the same msgId)
+	// WithSeenMessagesTTL without duration allows to
+	// process all incoming messages(even with the same msgId)
```
renaynay, line 43: why ? pls revert unrelated changes

nodebuilder/tests/swamp/swamp_tx.go
```diff
@@ -2,13 +2,16 @@ package swamp
...
 	"context"
+	"time"
 
 	"github.com/cosmos/cosmos-sdk/client/flags"
 )
 
 // FillBlocks produces the given amount of contiguous blocks with customizable size.
 // The returned channel reports when the process is finished.
 func (s *Swamp) FillBlocks(ctx context.Context, bsize, blocks int) chan error {
+	// TODO @renaynay: figure out why sleep is necessary to prevent flakeyness for macOS
```
renaynay, line 13: Please drop commit associated w these changes.

.gitignore
```diff
@@ -20,3 +20,4 @@ vendor
 /cel-key
 coverage.txt
 go.work
+./metals
```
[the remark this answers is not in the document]
derrandz: was caused by a weird vscode extension from Scala stuff, removed.

derrandz:
Back to you @mojtaba-esk , @renaynay has been able to provide a few fixes to the issues you've raised.

das/checkpoint.go
```diff
@@ -48,3 +48,12 @@ func (c checkpoint) String() string {
 
 	return str
 }
+
+// totalSampled returns the total amount of sampled headers
+func (c checkpoint) totalSampled() int {
```
vgonkivs, line 53: What if `totalSampled < 0`?
Can we make the return value `uint64`? It will allow us to avoid converting `uint64` to `int` in this func.  We can convert it into int in`recordTotalSampled`?

das/metrics.go
```diff
@@ -24,7 +25,10 @@ type metrics struct {
 	sampleTime    syncfloat64.Histogram
 	getHeaderTime syncfloat64.Histogram
 	newHead       syncint64.Counter
-	lastSampledTS int64
+	totalSampled  asyncint64.Gauge
+
+	lastSampledTS   int64
+	totalSampledInt int64
```
vgonkivs, line 31: Is it necessary to make it `int64`? Or we can also make it `uint64`

das/metrics_test.go
```diff
@@ -0,0 +1,111 @@
...
+	// require.NoError(t, err)
+
+	// totalSampled = mr.ScopeMetrics[0].Metrics[8]
+	// totalSampledGauge, _ = (totalSampled.Data).(metricdata.Gauge[float64])
+
+	// // assert that the collected metrics were the correct onces
+	// assert.Equal(t, totalSampled.Name, "das_total_sampled_headers")
+
+	// // assert for data correctness
+	// assert.Equal(t, len(totalSampledGauge.DataPoints), 2)
```
vgonkivs, line 66: commented code?

das/metrics_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+	err = daser.Stop(ctx)
+	require.NoError(t, err)
+}
+
+// test recordTotalSampled
+func TestRecordTotalSampled(t *testing.T) {
+	t.Skip()
+}
```
vgonkivs, line 75: ???

libs/header/sync/metrics.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"context"
+	"sync/atomic"
+
+	"go.opentelemetry.io/otel/metric/global"
+	"go.opentelemetry.io/otel/metric/instrument"
+)
+
+var meter = global.MeterProvider().Meter("header/sync")
```
vgonkivs, line 11: since we now have two `header` folders, I'd suggest renaming it to `libs/syncer`.
 @renaynay , wdyt?

nodebuilder/node/metrics_test.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	reader := metric.NewPeriodicReader(exp, metric.WithTimeout(1*time.Second))
+
+	provider := metric.NewMeterProvider(
+		metric.WithReader(reader),
+	)
+
+	return provider, reader, nil
+}
```
vgonkivs, line 82: We are using the same function in node/metrics_test.go and in das/metrics_test.go. Is it expected to be used somewhere else? If so, can we move it `libs/utils`?

libs/header/sync/metrics.go
renaynay: might be hard to differentiate for metrics-sake

nodebuilder/node/metrics_test.go
derrandz: Deprecating for now

das/metrics.go
derrandz: I think `uint64` is more fitting

das/checkpoint.go
derrandz: sounds good

das/metrics.go
derrandz: although this will cause conversion when calling `.Observe` since the metric type is `AsyncInt64.Gauge`

das/metrics.go
derrandz: thought I'd need it for testing, turns out not needed, and in either case now tests are removed so...

libs/header/sync/metrics.go
derrandz: keeping it as is ftm unless you guys think otherwise

das/metrics.go
```diff
@@ -110,29 +129,54 @@ func (d *DASer) InitMetrics() error {
...
 		return
 	}
 	m.newHead.Add(ctx, 1)
 }
+
+// recordTotalSampled records the total sampled headers.
+func (m *metrics) recordTotalSampled(totalSampled uint64) {
+	if m == nil {
+		return
+	}
```
renaynay, line 180:
```suggestion
```

please instead ensure that syncer never calls any method on a nil metrics.

das/metrics.go
```diff
@@ -24,7 +24,9 @@ type metrics struct {
 	sampleTime    syncfloat64.Histogram
 	getHeaderTime syncfloat64.Histogram
 	newHead       syncint64.Counter
-	lastSampledTS int64
+
+	lastSampledTS   int64
```
renaynay, line 28: fwiw this could also be uint64 for consistency

libs/header/sync/sync.go
```diff
@@ -212,7 +214,9 @@ func (s *Syncer[H]) sync(ctx context.Context) {
 	log.Infow("finished syncing",
 		"from", s.syncedHead.Height(),
 		"to", newHead.Height(),
-		"elapsed time", s.state.End.Sub(s.state.Start))
+		"elapsed time",
+		s.state.End.Sub(s.state.Start),
```
renaynay, line 218:
```suggestion
		"elapsed time", s.state.End.Sub(s.state.Start),
```

clearer

nodebuilder/tests/swamp/swamp_tx.go
```diff
@@ -9,6 +9,8 @@ import (
 // FillBlocks produces the given amount of contiguous blocks with customizable size.
 // The returned channel reports when the process is finished.
 func (s *Swamp) FillBlocks(ctx context.Context, bsize, blocks int) chan error {
+	// wait for first block to begin filling blocks
+	s.WaitTillHeight(ctx, 1)
```
renaynay, line 13: please remove these lines bc i'll add it in eventually with a diff pr

renaynay (changes requested):

das/metrics.go
derrandz: Do you mean `DASer`? We had a conversation me and @Wondertan about this pattern, and I mentioned that we should remove it.

For me to change this `if m == nil` would mean a major refactor to how we are doing metrics atm for DASer. The "better" pattern that we should have, that makes sense that metrics collection methods are not called in the first place when `Metrics` are disabled, is present in the blackbox metrics PR #1376.

To change this, we need a non-trivial refactor

derrandz: Referencing this in https://github.com/celestiaorg/celestia-node/issues/1705

das/metrics.go
derrandz: noted, (_although not related to this PR_)

das/metrics.go
renaynay: Since you already implemented the check here: https://github.com/celestiaorg/celestia-node/pull/1638/files#diff-a79f038a558af1c9917378c69183fb2563c3a0b6f49ae9464b1bfae68d2f1339R238-R240

we no longer need this check anyway

renaynay (dismissed):

nodebuilder/node_test.go
```diff
@@ -47,3 +55,83 @@ func TestLifecycle(t *testing.T) {
...
+		rawResponse, _ := proto.Marshal(&response)
+		contentType := "application/x-protobuf"
+		status := http.StatusOK
+
+		log.Debug("Responding to optl POST request")
+		w.Header().Set("Content-Type", contentType)
+		w.WriteHeader(status)
+		_, _ = w.Write(rawResponse)
+
+		log.Debug("Responded to optl POST request")
```
distractedm1nd, line 132:
```suggestion
		log.Debug("Responding to otlp POST request")
		w.Header().Set("Content-Type", contentType)
		w.WriteHeader(status)
		_, _ = w.Write(rawResponse)

		log.Debug("Responded to otlp POST request")
```

distractedm1nd (approved):

distractedm1nd (approved):

das/checkpoint.go
```diff
@@ -48,3 +48,12 @@ func (c checkpoint) String() string {
...
 	return str
 }
+
+// totalSampled returns the total amount of sampled headers
+func (c checkpoint) totalSampled() uint64 {
+	totalInProgress := 0
+	for _, w := range c.Workers {
+		totalInProgress += int(w.To-w.From) + 1
+	}
+	return uint64(int(c.SampleFrom) - totalInProgress - len(c.Failed))
```
vgonkivs, line 58:
```suggestion
	var totalInProgress uint64
	for _, w := range c.Workers {
		totalInProgress += w.To-w.From + 1
	}
	return c.SampleFrom - totalInProgress - uint64(len(c.Failed)))
```

vgonkivs (commented):
Small nits regarding type conversion but other than that lgtm

distractedm1nd (approved):

renaynay (approved):

vgonkivs (approved):

PR #1639, opened 2023-01-25
fix(share/discovery): decouple peer discovery from event processing

vgonkivs:
## Overview
Starting to listen to subscriptions async will help us to avoid any blocking in the case when we will not have the needed amount of FNs and will be blocked in `FindPeers`.
## Checklist


renaynay (commented):
let's test this on arabica bootstrappers

share/availability/discovery/discovery.go
```diff
@@ -107,7 +108,7 @@ func (d *Discovery) EnsurePeers(ctx context.Context) {
 		return
 	}
 	// subscribe on Event Bus in order to catch disconnected peers and restart the discovery
-	sub, err := d.host.EventBus().Subscribe(&event.EvtPeerConnectednessChanged{})
+	sub, err := d.host.EventBus().Subscribe(&event.EvtPeerConnectednessChanged{}, eventbus.BufSize(512))
```
renaynay, line 111: let's extract into a var in this file - Can we also document that the default buf size is 16 and why we increased it to 512

share/availability/discovery/discovery.go
```diff
@@ -121,6 +122,32 @@ func (d *Discovery) EnsurePeers(ctx context.Context) {
 			log.Error(err)
 		}
 	}()
+
+	// starting to listen to subscriptions async will help us to avoid any blocking
+	// in the case when we will not have the needed amount of FNs and will be blocked in `FindPeers`.
+	go func() {
+		select {
+		case <-ctx.Done():
+			return
```
renaynay, line 131: let's add some debug logs throughout this routine pls

share/availability/discovery/discovery.go
Wondertan: 512 is overkill btw, and 256 is more than enough, or even 128. There is literally nothing that could block event reading routine anymore

share/availability/discovery/discovery.go
```diff
@@ -121,9 +129,40 @@ func (d *Discovery) EnsurePeers(ctx context.Context) {
 			log.Error(err)
 		}
 	}()
+
+	// starting to listen to subscriptions async will help us to avoid any blocking
+	// in the case when we will not have the needed amount of FNs and will be blocked in `FindPeers`.
+	go func() {
+		select {
+		case <-ctx.Done():
+			log.Debug("Context canceled. Finish listening for connectedness events.")
```
Wondertan, line 138: Imo, it's pointless, but ok

share/availability/discovery/discovery.go
```diff
@@ -121,9 +129,40 @@ func (d *Discovery) EnsurePeers(ctx context.Context) {
...
+			return
+		case e, ok := <-sub.Out():
+			if !ok {
+				log.Debug("Subscription for connectedness events is closed.")
```
Wondertan, line 142: Same here.

Wondertan (approved):
Just two opinion farts, but LGTM

renaynay (approved):

distractedm1nd (approved):

Issue #1623, opened 2023-01-20
Unstable connectivity and stream negotiation on test networks


Wondertan:
## Context
We have two testing networks, Arabica and Mocha. Mocha has shown to be stable, while Arabica is still not. We have been fighting with issues on our bootstrappers of the Arabica network for about a month, and it's finally time to document what step we approached for future us.

## Main issues
Mainly, unstableness manifested via 
- [ ] (1) `failed to find any peer in table` errors on node startup 
- [ ] (2) `context.Deadline` errors during Header Store initialization on node startup
- [x] (3) inability to sync headers when the previous two errors didn't appear.

### Resource Manager

In the beginning, all the problems started with enabling `ResourceManager` in the libp2p host, which we presumed to be buggy(because of the libp2p discussions about broken auto-scaling in resource manager) on the version we were running(v0.20.1), so we [disabled](#1500) it, [temporarily](#1566). Our logs were cluttered with unclear logs that streams were not established due to limits, and it wasn't clear where they were coming from until we updated our libp2p version, which included [the fix to make logs less confusing](https://github.com/libp2p/go-libp2p/issues/1928).

### Infra

Unfortunately, removing the resource manager and updating the libp2p version with less confusing resource manager errors didn't help us resolve unstableness, so we blamed the infrastructure due to Mocha working fine. Still, we had doubts over that as the Mocha and Arabica were running different node versions(The diff was not too big, but still). 

We presumed that Arabica is a developer-focused network that should have a higher load. @sysrex tried vertically scaling the hardware of 3 bootstrappers to _a massive_ size, which didn't help. Additionally, @sysrex analyzed resource usage on the machines, which was negligible. Finally, we compared the number of peers connected to bootstrappers on Mocha and Arabica. The results showed that Mocha has 10x more peers(yeah, we should have seen this earlier, and our network observability is in a poor state at this moment) with humble hardware for bootstrappers, so the conclusion is that this is def not a hardware issue. 

### Software

Next, we brought both networks to full parity in the setup and software version to see if something in the diff could explain it. However, Arabica was still unstable. Meanwhile, @Bidon15 and @Wondertan  went full debug mode and found out that the reason behind (3) is a bug inside the new version of header ex, which was initially fixed by #1592 as an experiment and superseded with #1603.

[...]

PR #1647, opened 2023-01-26
fix(header/p2p): add trivial retrying for trustedPeer requested

Wondertan:
Anyway, we should have a retry mechanism for requests here(but long term, it should be a better than this implementation), and also, this seems to be helping with #1623 during the first initialization of a node.

### DIFF
--- a/libs/header/p2p/exchange.go
+++ b/libs/header/p2p/exchange.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"math/rand"
 	"sort"
+	"time"
 
 	logging "github.com/ipfs/go-log/v2"
 	"github.com/libp2p/go-libp2p/core/host"
@@ -224,9 +225,21 @@ func (ex *Exchange[H]) performRequest(
 		return nil, fmt.Errorf("no trusted peers")
 	}
 
-	//nolint:gosec // G404: Use of weak random number generator
-	index := rand.Intn(len(ex.trustedPeers))
-	return ex.request(ctx, ex.trustedPeers[index], req)
+	for ctx.Err() == nil {
+		//nolint:gosec // G404: Use of weak random number generator
+		index := rand.Intn(len(ex.trustedPeers))
+		cctx, cancel := context.WithTimeout(context.Background(), time.Second)
+		h, err := ex.request(cctx, ex.trustedPeers[index], req)
+		cancel()
+		if err != nil {
+			log.Error(err)
+			continue
+		}
+		return h, nil
+
+	}
+
+	return nil, ctx.Err()
 }
 
 // request sends the HeaderRequest to a remote peer.
--- a/libs/header/p2p/exchange.go
+++ b/libs/header/p2p/exchange.go
@@ -82,7 +82,7 @@ func (ex *Exchange[H]) Start(context.Context) error {
 		// We don't really care if we succeed at this point
 		// and just need any peers in the peerTracker asap
 		go func(p peer.ID) {
-			err := ex.host.Connect(ex.ctx, peer.AddrInfo{ID: p}) //nolint:errcheck
+			err := ex.host.Connect(ex.ctx, peer.AddrInfo{ID: p})
 			if err != nil {
 				log.Debugw("err connecting to a bootstrap peer", "err", err, "peer", p)
 			}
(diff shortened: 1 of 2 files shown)


vgonkivs (dismissed):

libs/header/p2p/exchange.go
```diff
@@ -224,9 +225,21 @@ func (ex *Exchange[H]) performRequest(
 		return nil, fmt.Errorf("no trusted peers")
 	}
 
-	//nolint:gosec // G404: Use of weak random number generator
-	index := rand.Intn(len(ex.trustedPeers))
-	return ex.request(ctx, ex.trustedPeers[index], req)
+	for ctx.Err() == nil {
+		//nolint:gosec // G404: Use of weak random number generator
+		index := rand.Intn(len(ex.trustedPeers))
+		cctx, cancel := context.WithTimeout(context.Background(), time.Second)
```
renaynay, line 231: why context.Background here?

Wondertan: Does not have to be but it better be

renaynay: Why not just use `ex.ctx` here like we do everywhere else throughout?

Wondertan: I mean that it's better to be a passed context, not a background. Changing it to `ctx` though, not `ex.ctx`

renaynay (dismissed):

libs/header/p2p/exchange.go
```diff
@@ -220,13 +229,21 @@ func (ex *Exchange[H]) performRequest(
...
 	}
 
-	if len(ex.trustedPeers) == 0 {
-		return nil, fmt.Errorf("no trusted peers")
+	for {
+		//nolint:gosec // G404: Use of weak random number generator
+		idx := rand.Intn(len(ex.trustedPeers))
+		ctx, cancel := context.WithTimeout(ctx, ex.Params.TrustedPeersRequestTimeout)
+
+		h, err := ex.request(ctx, ex.trustedPeers[idx], req)
```
renaynay, line 237:
```suggestion
                 tpeer := ex.trustedPeers[idx]
		h, err := ex.request(ctx, tpeer, req)
```

libs/header/p2p/exchange.go
```diff
@@ -220,13 +229,21 @@ func (ex *Exchange[H]) performRequest(
...
+		cancel()
+		switch err {
+		default:
+			log.Debug(err)
```
renaynay, line 241:
```suggestion
			log.Debugw("request failed", "peer", tpeer.String(), "err", err)
```

renaynay (commented):
needs lint

libs/header/p2p/exchange.go
Wondertan: All the errors are wrapped in `request`, so doing this duplicate info mostly make the whole log uglier. I tried it originally

renaynay (dismissed):

libs/header/p2p/exchange.go
```diff
@@ -220,13 +225,21 @@ func (ex *Exchange[H]) performRequest(
 		return make([]H, 0), nil
 	}
 
-	if len(ex.trustedPeers) == 0 {
-		return nil, fmt.Errorf("no trusted peers")
+	for {
+		//nolint:gosec // G404: Use of weak random number generator
+		idx := rand.Intn(len(ex.trustedPeers))
```
walldiss, line 230: Is it intended that errored peer could be retried before all non-tried peers has been attempted? For example it could be the same random result subsequently.

Wondertan: Yeah, but it's fine. The trustedPeers are assumed to be highly available, so retrying the same peer is fine. Also, this is not the final solution for retrying and peer selection logic: #1665

libs/header/p2p/exchange.go
```diff
@@ -220,13 +225,21 @@ func (ex *Exchange[H]) performRequest(
 		return make([]H, 0), nil
 	}
 
-	if len(ex.trustedPeers) == 0 {
-		return nil, fmt.Errorf("no trusted peers")
```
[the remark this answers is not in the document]
Wondertan: Returned it back and made an issue to move it constructor

walldiss (approved):

walldiss (approved):

renaynay (approved):

PR #1649, opened 2023-01-27
Tzdybal/fix ranges

tzdybal:
## Overview

`ranges.Add` method is not thread safe. if you execute it in parallel you can end up with ranges out-of-order.
In `syncer.newNetHead` nothing prohibits parallel execution of  `pending.Add` - syncing point is in `wantSync` using buffered channel.
`newNetHead` is called ass libp2p gossip validator, which is asynchronous - can be executed in parallel.

This change ensures that `Add` method is thread-safe.
Added test fails in first commit (to proof the point), and second commit fixes the issue.

## Checklist


Wondertan:
Can you try to run the reconstruction swamp teat locally with -count 100 to see if it's really fixes it? 

Good catch anyway!

renaynay (dismissed):
Looks fine to me - am running it against arabica where a lot of these issues were observed and looks good so far.

Thanks for the fix @tzdybal

renaynay:
https://github.com/celestiaorg/celestia-node/actions/runs/4023389491/jobs/6914202398#step:4:320

renaynay (approved):

Wondertan (approved):

PR #1678, opened 2023-02-01
fix: always calculate the data root even for blocks with no txs

### LINKED ISSUE #1677: Incorrect assumption for empty blocks on mocha
Currently, when creating the extended header, we're making the assumption that the square is empty if there are no transactions. Most of the time this is the case, but when there is evidence, this is not the case as evidence (in v0.11.1 of celestia-app and below) is also included in the square.

https://github.com/celestiaorg/celestia-node/blob/534ec0df88fd885d69c04daf41c0ba10e4e1512f/header/header.go#L45-L77

one simple solution to this is to simply always split the block into shares and create the extended data square, instead of conditionally using the minimum data availability header function (which should likely only be used for unit testing purposes).

evan-forbes:
## Overview

closes #1677 

## Checklist

### DIFF
--- a/header/header.go
+++ b/header/header.go
@@ -82,22 +82,15 @@ func MakeExtendedHeader(
 	vals *core.ValidatorSet,
 	bServ blockservice.BlockService,
 ) (*ExtendedHeader, error) {
-	var dah DataAvailabilityHeader
-	if len(b.Txs) > 0 {
-		shares, err := appshares.Split(b.Data, true)
-		if err != nil {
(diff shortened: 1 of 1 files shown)


evan-forbes:
converted to a draft because of https://github.com/celestiaorg/celestia-node/issues/1677#issuecomment-1412744049

evan-forbes:
closing because of https://github.com/celestiaorg/celestia-node/issues/1677#issuecomment-1413658851

PR #1685, opened 2023-02-02
feat: adding example types and auth comments to OpenRPC doc generation

### LINKED ISSUE #1358: docs(rpc+gateway): Finish adding example types
We should have example types for every possible input and output type. The scaffolding for doing this is introduced in #1283.

distractedm1nd:
## Overview
This PR adds auth comments and example types to our OpenRPC API docgen. It replaces #1408 and closes #1358.

### DIFF
--- a/api/rpc/client/client.go
+++ b/api/rpc/client/client.go
@@ -17,6 +17,18 @@ import (
 	"github.com/celestiaorg/celestia-node/nodebuilder/state"
 )
 
+// TODO: this duplication of strings many times across the codebase can be avoided with issue #1176
+var client Client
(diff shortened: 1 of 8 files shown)


api/rpc/client/client.go
```diff
@@ -17,6 +17,18 @@ import (
 	"github.com/celestiaorg/celestia-node/nodebuilder/state"
 )
 
+// TODO: this duplication of strings many times across the codebase can be avoided with issue #1176
+var client Client
```
renaynay, line 21:
```suggestion
var client Client

```

renaynay (approved):

PR #1686, opened 2023-02-02
fix: flake in TestNamespaceHasher_CorruptedData

distractedm1nd:
## Overview
Fixes a flake in TestNamespaceHasher_CorruptedData that occurs when we try to hash an empty byte array.


Wondertan (approved):
:gun:

vgonkivs (approved):

Issue #1697, opened 2023-02-07
improvement(libs/header): check max header height before fetching the header range from the store in header-ex Server


vgonkivs:
Currently, we are handling range requests by fetching them directly from the store. In fact, when the node is syncing and does not have these ranges, we are subscribing to the `to` and waiting for it. But request timeout is reached in most of the cases. As we request ranges using peerTracker, we can avoid this 5-second timeout by checking max stored height in `header.Store`. In case if requested height > `header.Store.Height()`, the server will return `ErrHeaderNotFound` and the client will request headers from a different peer.

PR #1706, opened 2023-02-08
improvement(libs/header): Check for requested height before fetching it from the store

vgonkivs:
## Overview

Resolves https://github.com/celestiaorg/celestia-node/issues/1697

## Checklist

### DIFF
--- a/libs/header/p2p/server.go
+++ b/libs/header/p2p/server.go
@@ -28,8 +28,8 @@ var (
 type ExchangeServer[H header.Header] struct {
 	protocolID protocol.ID
 
-	host   host.Host
-	getter header.Getter[H]
+	host  host.Host
+	store header.Store[H]
 
 	ctx    context.Context
 	cancel context.CancelFunc
@@ -41,7 +41,7 @@ type ExchangeServer[H header.Header] struct {
 // header-related requests.
 func NewExchangeServer[H header.Header](
 	host host.Host,
-	getter header.Getter[H],
+	store header.Store[H],
 	opts ...Option[ServerParameters],
 ) (*ExchangeServer[H], error) {
 	params := DefaultServerParameters()
@@ -55,7 +55,7 @@ func NewExchangeServer[H header.Header](
 	return &ExchangeServer[H]{
 		protocolID: protocolID(params.protocolSuffix),
 		host:       host,
-		getter:     getter,
+		store:      store,
 		Params:     params,
 	}, nil
 }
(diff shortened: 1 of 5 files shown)


libs/header/store/store.go
```diff
@@ -283,6 +283,10 @@ func (s *Store[H]) Has(ctx context.Context, hash header.Hash) (bool, error) {
 	return s.ds.Has(ctx, datastore.NewKey(hash.String()))
 }
 
+func (s *Store[H]) HasAt(height uint64) bool {
+	return height != uint64(0) && s.Height() >= height
+}
```
Wondertan, line 288: Let's make ctx as a first param. 

Right now, we assume that the Store stores everything below `Height`. However, we will break this assumption once we will introduce pruning. Then we will do a disk look up which requires context and breaking the interface. Let's avoid breaking in it in future by introducing future proof form of it

Wondertan (commented):
One request, and I think this is a very good idea to have

libs/header/store/store.go
renaynay: fine w me

libs/header/p2p/server.go
```diff
@@ -201,7 +201,13 @@ func (serv *ExchangeServer[H]) handleRequest(from, to uint64) ([]H, error) {
 	}
 
 	log.Debugw("server: handling headers request", "from", from, "to", to)
-	headersByRange, err := serv.getter.GetRangeByHeight(ctx, from, to)
+	if !serv.store.HasAt(to - 1) {
+		span.SetStatus(codes.Error, header.ErrNotFound.Error())
+		log.Warnw("server: requested headers not stored")
```
renaynay, line 206: I'd do a debug log as this can be more frequent - also log the `to` height

renaynay (approved):

libs/header/store/store.go
vgonkivs: true. Good idea, thank you.

libs/header/p2p/server.go
vgonkivs: Changed to debug. But I do not think we really need to print `to` height one more time. We are logging 
`log.Debugw("server: handling headers request", "from", from, "to", to)`  before `HasAt` check.

Wondertan (dismissed):

libs/header/p2p/server.go
renaynay: those can always be lost though, meaning logs get cloudy so it's nice to know which request this is attributed to

renaynay (approved):

libs/header/p2p/server.go
vgonkivs: ok, np.

renaynay (approved):

distractedm1nd (approved):

PR #1719, opened 2023-02-09
feat(nodebuilder/p2p): migrate to EDSStore Blockstore for Bitswap for FNs/BNs

### LINKED ISSUE #1422: node: migrate `Blockstore` over `eds.Store.Blockstore` for FN/BNs
## Context

Right now, we have a global `Blockstore` over `Datastore`. As per #1099, we need to migrate to another global `Blockstore` provided by `eds.Store`, which is compatible with the new block storage.

Mainly, we should stop providing the existing `Blockstore` and a new one instead. We should also double-check that we aren't writing anywhere in the new `Blockstore` and that it is provided to `Bitswap`

distractedm1nd:
## Overview
Closes #1422

## Checklist

### DIFF
--- a/fraud/service_test.go
+++ b/fraud/service_test.go
@@ -16,7 +16,6 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/celestiaorg/celestia-node/header"
-	"github.com/celestiaorg/celestia-node/nodebuilder/p2p"
 )
 
 func TestService_Subscribe(t *testing.T) {
@@ -133,7 +132,7 @@ func TestService_ReGossiping(t *testing.T) {
 		},
 		sync.MutexWrap(datastore.NewMapDatastore()),
 		false,
-		string(p2p.Private),
+		"private",
 	)
 	addrB := host.InfoFromHost(net.Hosts()[1]) // -> B
 
@@ -149,7 +148,7 @@ func TestService_ReGossiping(t *testing.T) {
 		},
 		sync.MutexWrap(datastore.NewMapDatastore()),
 		false,
-		string(p2p.Private),
+		"private",
 	)
 	// establish connections
 	// connect peers: A -> B -> C, so A and C are not connected to each other
--- a/nodebuilder/p2p/bitswap.go
+++ b/nodebuilder/p2p/bitswap.go
@@ -42,7 +42,7 @@ func dataExchange(params bitSwapParams) exchange.Interface {
 }
 
 func blockstoreFromDatastore(ctx context.Context, ds datastore.Batching) (blockstore.Blockstore, error) {
-	bs, err := blockstore.CachedBlockstore(
+	return blockstore.CachedBlockstore(
 		ctx,
 		blockstore.NewBlockstore(ds),
 		blockstore.CacheOpts{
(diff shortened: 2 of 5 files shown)


distractedm1nd:
Will fail until rebased on #1632 because of an import cycle that is introduced here but resolved there

fraud/service_test.go
```diff
@@ -133,7 +132,7 @@ func TestService_ReGossiping(t *testing.T) {
 		},
 		sync.MutexWrap(datastore.NewMapDatastore()),
 		false,
-		string(p2p.Private),
```
renaynay, line 136: y

nodebuilder/p2p/bitswap.go
```diff
@@ -25,30 +27,48 @@ const (
...
 		// NOTE: These below ar required for our protocol to work reliably.
 		// See https://github.com/celestiaorg/celestia-node/issues/732
 		bitswap.SetSendDontHaves(false),
 		bitswap.SetSimulateDontHavesOnTimeout(false),
-	), bs, nil
+	)
+}
+
+func blockstoreFromDatastore(ctx context.Context, ds datastore.Batching) (blockstore.Blockstore, error) {
+	bs, err := blockstore.CachedBlockstore(
```
renaynay, line 45: return blockstore.CachedBlockstore

nodebuilder/p2p/bitswap.go
```diff
@@ -25,30 +27,48 @@ const (
...
+		},
+	)
+	if err != nil {
+		return nil, err
+	}
+	return bs, err
+}
+
+func blockstoreFromEDSStore(ctx context.Context, store *eds.Store) (blockstore.Blockstore, error) {
+	bs, err := blockstore.CachedBlockstore(
```
renaynay, line 61: same here

nodebuilder/share/module.go
```diff
@@ -38,6 +37,7 @@ func ConstructModule(tp node.Type, cfg *Config, options ...fx.Option) fx.Option
 		return fx.Module(
 			"share",
 			baseComponents,
+			fx.Invoke(share.EnsureEmptySquareExists),
```
renaynay, line 40: does this happen somewhere implicitly for nodes running eds store?

renaynay:
also lint broke

fraud/service_test.go
distractedm1nd: Ideologically, because things outside nodebuilder should not depend on nodebuilder (@Wondertan proud of me?)
More practically, because there is an import cycle if you do this. nodebuilder/p2p now depends on share/eds for the edsstore.Blockstore, but eds depends on fraud, so fraud cannot depend on nodebuilder/p2p.

nodebuilder/share/module.go
distractedm1nd: Yes, inside the start function for edsStore (ensureEmptyCarExists)

renaynay: kk

renaynay (approved):

vgonkivs (approved):

walldiss (approved):

PR #1734, opened 2023-02-10
fix(share/shrex/peer_manager): peer manager needs to do soft delete on synced pool instead of hard one

walldiss:
## Overview

This fix resolves an issue where peers who would broadcast hashes via shrex-sub that were already retrieved by the node would get blacklisted. This happened because after retrieval of a given hash, the `syncPool` corresponding to the given hash gets deleted, so when a delayed peer would broadcast the same hash after retrieval, the manager would create a new pool that goes unvalidated since the header was already received via header-sub, causing the delayed peers in the recreated pool to get blacklisted. 

The solution is to do a soft-delete of the pool such that delayed peers are GC'd rather than blacklisted.

## Checklist

### DIFF
--- a/share/p2p/peers/manager.go
+++ b/share/p2p/peers/manager.go
@@ -67,8 +67,8 @@ type syncPool struct {
 
 	// isValidatedDataHash indicates if datahash was validated by receiving corresponding extended
 	// header from headerSub
-	isSynced            atomic.Bool
 	isValidatedDataHash atomic.Bool
+	isSynced            atomic.Bool
 	createdAt           time.Time
 }
 
--- a/share/p2p/peers/manager.go
+++ b/share/p2p/peers/manager.go
@@ -7,6 +7,7 @@ import (
 	"sync"
 	"sync/atomic"
 	"time"
+	"unsafe"
 
 	"github.com/libp2p/go-libp2p/p2p/net/conngater"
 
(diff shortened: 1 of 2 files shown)


share/p2p/peers/manager.go
```diff
@@ -66,6 +67,7 @@ type syncPool struct {
 
 	// isValidatedDataHash indicates if datahash was validated by receiving corresponding extended
 	// header from headerSub
+	isSynced            atomic.Bool
 	isValidatedDataHash atomic.Bool
```
distractedm1nd, line 71:
```suggestion
	// isValidatedDataHash indicates if datahash was validated by receiving corresponding extended
	// header from headerSub
	isValidatedDataHash atomic.Bool
	isSynced            atomic.Bool
```
match comment with field

distractedm1nd (changes requested):

distractedm1nd (approved):

share/p2p/peers/manager.go
```diff
@@ -278,18 +274,16 @@ func (s *Manager) hashIsBlacklisted(hash share.DataHash) bool {
...
 	ticker := time.NewTicker(gcInterval)
 	defer ticker.Stop()
 
 	var blacklist []peer.ID
 	for {
 		blacklist = s.cleanUp()
-		s.blacklistPeers(blacklist...)
+		if len(blacklist) > 0 {
+			s.blacklistPeers(blacklist...)
+		}
```
renaynay, line 286: do you need this check? It'll just short circuit as there's nothing to loop over

renaynay (approved):

share/p2p/peers/manager.go
walldiss: It was producing misleading and scary logs of "blacklisting peers, peers:[]" would rather just skip :)

PR #1740, opened 2023-02-13
fix(getters/shrex): removing implicit context assumption

distractedm1nd:
## Overview
Prevents infinite processing during availability calls

### DIFF
--- a/share/getters/shrex.go
+++ b/share/getters/shrex.go
@@ -33,7 +33,12 @@ func (sg *ShrexGetter) GetShare(ctx context.Context, root *share.Root, row, col 
 }
 
 func (sg *ShrexGetter) GetEDS(ctx context.Context, root *share.Root) (*rsmt2d.ExtendedDataSquare, error) {
-	for ctx.Err() != nil {
+	for {
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		default:
+		}
 		peer, setStatus, err := sg.peerManager.Peer(ctx, root.Hash())
 		if err != nil {
 			log.Debugw("couldn't find peer", "datahash", root.String(), "err", err)
@@ -55,7 +60,6 @@ func (sg *ShrexGetter) GetEDS(ctx context.Context, root *share.Root) (*rsmt2d.Ex
 			setStatus(peers.ResultFail)
 		}
 	}
-	return nil, ctx.Err()
 }
 
 func (sg *ShrexGetter) GetSharesByNamespace(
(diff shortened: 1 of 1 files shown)


share/getters/shrex.go
```diff
@@ -34,6 +34,11 @@ func (sg *ShrexGetter) GetShare(ctx context.Context, root *share.Root, row, col
 
 func (sg *ShrexGetter) GetEDS(ctx context.Context, root *share.Root) (*rsmt2d.ExtendedDataSquare, error) {
 	for {
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		default:
+		}
 		peer, setStatus, err := sg.peerManager.Peer(ctx, root.Hash())
```
Wondertan, line 42: AFAICS, Peer method handles context correctly, and usually, such ctx check means the call down the stack is blocking. Why do we need to do it?

share/getters/shrex.go
```diff
@@ -34,6 +34,11 @@ func (sg *ShrexGetter) GetShare(ctx context.Context, root *share.Root, row, col
 
 func (sg *ShrexGetter) GetEDS(ctx context.Context, root *share.Root) (*rsmt2d.ExtendedDataSquare, error) {
 	for {
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		default:
+		}
```
Wondertan, line 41: alternative and shorter form
```suggestion
	for ctx.Err == nil {
```

share/getters/shrex.go
distractedm1nd: We need it in the case that there is a peer in the pool. `Peer` returns a peer before it checks the context in this case.

share/getters/shrex.go
distractedm1nd: based

walldiss (approved):

share/getters/shrex.go
```diff
@@ -33,7 +33,7 @@ func (sg *ShrexGetter) GetShare(ctx context.Context, root *share.Root, row, col
 }
 
 func (sg *ShrexGetter) GetEDS(ctx context.Context, root *share.Root) (*rsmt2d.ExtendedDataSquare, error) {
-	for {
+	for ctx.Err() != nil {
```
renaynay, line 36: huh? shouldn't it be the other way around?  and even so, can we stick w wrapping w select pattern pls?

renaynay (changes requested):

share/getters/shrex.go
distractedm1nd: yes

share/getters/shrex.go
renaynay: we use select pattern everywhere else so lets stick to it but im fine w shorter form if we unify later on

renaynay (approved):

PR #1729, opened 2023-02-10
feat(share): Integrate rearchitectured blocksync into fx

distractedm1nd:
## Overview
This PR puts the implemented puzzle pieces of the rearchitectured blocksync together, and integrates them into our app.

Closes #1492
Closes #1433 
Closes #1421
Closes #1464

PR #1743, opened 2023-02-14
feat!(share): adding config variables for blocksync

### LINKED ISSUE #1465: node/share: Data sync modes for FN

distractedm1nd:
## Overview
Closes #1465 
Based on #1729

### DIFF
--- a/nodebuilder/share/config.go
+++ b/nodebuilder/share/config.go
@@ -18,9 +18,6 @@ type Config struct {
 	// AdvertiseInterval is a interval between advertising sessions.
 	// NOTE: only full and bridge can advertise themselves.
 	AdvertiseInterval time.Duration
-	// GetterTimeout is the duration that each share retrieval getter will be run before moving to
-	// the next getter.
-	GetterTimeout time.Duration
 	// UseShareExchange is a flag toggling the usage of shrex protocols for blocksync.
 	// NOTE: This config variable only has an effect on full and bridge nodes.
 	UseShareExchange bool
@@ -31,14 +28,13 @@ func DefaultConfig() Config {
 		PeersLimit:        3,
 		DiscoveryInterval: time.Second * 30,
 		AdvertiseInterval: time.Second * 30,
-		GetterTimeout:     time.Second * 30,
 		UseShareExchange:  true,
 	}
 }
 
 // Validate performs basic validation of the config.
 func (cfg *Config) Validate() error {
-	if cfg.DiscoveryInterval <= 0 || cfg.AdvertiseInterval <= 0 || cfg.GetterTimeout <= 0 {
+	if cfg.DiscoveryInterval <= 0 || cfg.AdvertiseInterval <= 0 {
 		return fmt.Errorf("nodebuilder/share: %s", ErrNegativeInterval)
 	}
 	return nil
--- a/nodebuilder/share/config.go
+++ b/nodebuilder/share/config.go
@@ -18,6 +18,12 @@ type Config struct {
 	// AdvertiseInterval is a interval between advertising sessions.
 	// NOTE: only full and bridge can advertise themselves.
 	AdvertiseInterval time.Duration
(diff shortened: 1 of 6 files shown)


share/getters/shrex.go
```diff
@@ -28,6 +28,23 @@ type ShrexGetter struct {
...
+	}
+}
+
+func (sg *ShrexGetter) Start(ctx context.Context) error {
+	return sg.peerManager.Start(ctx)
+}
+
+func (sg *ShrexGetter) Stop(ctx context.Context) error {
+	return sg.peerManager.Stop(ctx)
+}
```
Wondertan, line 46: If ShrexGetter maintains the lifecycle of `peer.Manager`, it should also construct it, i.e. `peer.Manager` is a child of ShrexGetter but not a dependency. This also means that peer manager won't be part of FX dep graph, which is totally ok as the peer manager is a single purpose.

I don't wanna push this, and we can do it during later refactorings, so can make an issue

Wondertan (commented):
Have only one requests

share/getters/shrex.go
Wondertan: Consider this is a feedback for https://github.com/celestiaorg/celestia-node/pull/1729

nodebuilder/share/constructors.go
```diff
@@ -57,3 +65,30 @@ func ensureEmptyCARExists(ctx context.Context, store *eds.Store) error {
...
+	shrexSub *shrexsub.PubSub,
+	discovery *disc.Discovery,
+	host host.Host,
+	connGater *conngater.BasicConnectionGater,
+) *peers.Manager {
+	// TODO: find better syncTimeout duration?
+	return peers.NewManager(headerSub, shrexSub, discovery, host, connGater, modp2p.BlockTime*5)
+}
+
+func fullGetter(
```
renaynay, line 80:
```suggestion
func cascadeGetter(
```

renaynay (commented):
looks like this contains commits from other pr

nodebuilder/share/config.go
```diff
@@ -18,19 +18,27 @@ type Config struct {
 	// AdvertiseInterval is a interval between advertising sessions.
 	// NOTE: only full and bridge can advertise themselves.
 	AdvertiseInterval time.Duration
+	// GetterTimeout is the duration that each share retrieval getter will be run before moving to
+	// the next getter.
+	GetterTimeout time.Duration
+	// UseShareExchange is a flag toggling the usage of shrex protocols for blocksync.
+	// NOTE: This config variable only has an effect on full and bridge nodes.
+	UseShareExchange bool
```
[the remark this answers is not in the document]
distractedm1nd: No, it is a tracked issue, the one this pr is trying to close #1465

nodebuilder/share/constructors.go
distractedm1nd: light nodes will also use a cascadeGetter

nodebuilder/share/config.go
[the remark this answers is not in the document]
Wondertan: For testing, yes

nodebuilder/share/constructors.go
Wondertan: light nodes will also use a cascadeGetter

Wondertan (approved):

nodebuilder/share/constructors.go
distractedm1nd: light nodes will also use a cascadeGetter

renaynay: light nodes will also use a cascadeGetter

renaynay (approved):
have u tested these default params yet?

PR #1745, opened 2023-02-14
feat(share/shrex): add Rate limit middleware for shrex servers

walldiss:
## Overview
Closes https://github.com/celestiaorg/celestia-node/issues/1747

Add concurrency rate limiting on server side of shrex. 

## Checklist


distractedm1nd:
Hey, I will merge this into the integration branch and run some tests with it before I approve, but looks really good

share/p2p/middleware.go
```diff
@@ -0,0 +1,30 @@
...
+		defer atomic.AddInt64(&parallelRequests, -1)
+
+		if current > limit {
+			log.Debug("concurrency limit reached")
+			err := stream.Close()
+			if err != nil {
+				log.Errorw("server: closing stream", "err", err)
+			}
+			return
+		}
```
Wondertan, line 27: I think we had a special error that was supposed to be sent in this case and not just stream closage

distractedm1nd: We did, but we want to rethink this design - this would still require the overloaded nodes to read the request from the client, and then send a response - it is much better if we can just close the stream gracefully, and have the client interpret it as a rate limit response

distractedm1nd:
Works beautifully on TG

share/p2p/middleware.go
distractedm1nd: @Wondertan Can we merge this for now? We won't merge it until we have your okay, but we do want it in the feature branch. It significantly improves metrics

walldiss: As far as I remember initially @Wondertan purposed to Close the stream without response, but we revised it later with idea of sending response with status instead. Motivation was that it would allow us to know if stream was unintentionally closed or it was expected rate limiting. 

With closer look it turned out we can identify graceful stream close with error handling check, so there is no need to send extra message.

Pros of getting rid of extra message:
With explicit message approach we would have an extra round trip. Server that is already exhausted, would need to read request first, which will be thrown away. And then it would need to send response containing "Status_REFUSED". We could achieve the same, without extra round trip.

Also we can have rate limiting protocol agnostic, as it would not need to read out protocol specific request.

Wondertan: The reasoning makes sense to me. Thanks for explaining. As long as there is explicit handling for the client side of this case I am ok with this approach. We will also need to update the specs in the Adr right after.

Wondertan (approved):

share/p2p/shrexeds/server.go
```diff
@@ -63,6 +64,7 @@ func (s *Server) Stop(context.Context) error {
 func (s *Server) handleStream(stream network.Stream) {
 	log.Debug("server: handling eds request")
 
+	// TODO: add read timeout!
```
distractedm1nd, line 67: we can remove this right

distractedm1nd (changes requested):

share/p2p/shrexeds/server.go
walldiss: yep

distractedm1nd (approved):

share/p2p/shrexnd/options.go
```diff
@@ -52,6 +56,9 @@ func (p *Parameters) Validate() error {
 	if p.serveTimeout <= 0 {
 		return fmt.Errorf("invalid serve timeout: %v, %s", p.serveTimeout, errSuffix)
 	}
+	if p.concurrencyLimit < 1 {
+		return fmt.Errorf("invalid concurrency limit: value should be greater than 0")
```
renaynay, line 60: why not errSuffix

renaynay (approved):

Issue #1329, opened 2022-11-09
nodebuilder/header: Sanity check `RawHeader.ChainID` against p2p.Network during `InitStore`


renaynay:
This is to provide further assurance that node is not syncing against incorrect network.

renaynay:
compare `RawHeader.ChainID` against the node's current network.

Wondertan:
Note that InitStore is not allowed to import `p2p.Network`

renaynay:
@ninocomputer

PR #1746, opened 2023-02-14
misc(libs/header): verify chainID in incoming headers

vgonkivs:
## Overview

Resolves https://github.com/celestiaorg/celestia-node/issues/1329

Special thanks to @ninocomputer for helping me with this PR.

## Checklist


libs/header/p2p/exchange.go
```diff
@@ -270,8 +273,16 @@ func (ex *Exchange[H]) request(
 		if err != nil {
 			return nil, err
 		}
+		err = validateChainID(ex.Params.protocolSuffix, header.(H).ChainID())
+		if err != nil {
+			return nil, err
+		}
```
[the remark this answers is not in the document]
vgonkivs: 1) protocolSuffix is hardcoded by default and could be changed by someone else;
2) told you @renaynay 😂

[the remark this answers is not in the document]
vgonkivs: The idea, that I had is to store chainID of the header, that will be received after `Get(trustedHash)`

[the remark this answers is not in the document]
Wondertan: 1. Its optional: https://github.com/celestiaorg/celestia-node/blob/main/libs/header/p2p/exchange.go#L38-L42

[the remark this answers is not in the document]
vgonkivs: huh, yes. You are right, sorry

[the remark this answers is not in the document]
vgonkivs: We can add one more field in the config...

libs/header/test/testing.go
```diff
@@ -41,7 +41,7 @@ func (d *DummyHeader) IsZero() bool {
 }
 
 func (d *DummyHeader) ChainID() string {
-	return d.Raw.ChainID
+	return "private"
```
renaynay, line 44: p2p.Private

libs/header/p2p/exchange.go
```diff
@@ -236,12 +236,17 @@ func (ex *Exchange[H]) performRequest(
 		cancel()
 		switch err {
 		default:
+			reqErr = err
 			log.Debug(err)
 			continue
 		case context.Canceled, context.DeadlineExceeded, nil:
+			if err != nil && reqErr != nil {
+				err = reqErr
+			}
```
renaynay, line 245: can you document why you're doing this? isn't the whole point of this loop with trusted peers is so that if a request fails to one peer, you can move on to next peer and ask them and if it's successful, just return successfully? This makes is so if previous peer had error, you will still return that err even if the next peer was successful, no?

Also can you make the debug log on #L240 more descriptive ?

libs/header/p2p/exchange.go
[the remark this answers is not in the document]
renaynay: I would just store it on the struct as `chainID`

ninocomputer (changes requested):
wen credit

libs/header/p2p/exchange.go
vgonkivs: Actually here we are handling context.Cancel error. If reqError is not empty, then it means that we received errors some errors from our trusted peers, and IMO it's better to return reqError(we can reqError.Join() after moving to Go1.20)

libs/header/test/testing.go
vgonkivs: this is lib. It should not know anything about p2p package I think.

renaynay: oh fair point

libs/header/p2p/exchange.go
renaynay: Okay but still - look at what the caller if this is doing. If one request succeeds and some other one failed, we will still error out bc one failed.

libs/header/p2p/exchange.go
```diff
@@ -230,24 +230,23 @@ func (ex *Exchange[H]) performRequest(
...
 		idx := rand.Intn(len(ex.trustedPeers))
-		ctx, cancel := context.WithTimeout(ctx, ex.Params.TrustedPeersRequestTimeout)
 
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		default:
+		}
+
+		ctx, cancel := context.WithTimeout(ctx, ex.Params.TrustedPeersRequestTimeout)
```
renaynay, line 240: can we call this reqCtx or something?

vgonkivs: Yes, meme

libs/header/p2p/exchange.go
```diff
@@ -270,8 +276,16 @@ func (ex *Exchange[H]) request(
 		if err != nil {
 			return nil, err
 		}
+		err = validateChainID(ex.Params.chainID, header.(H).ChainID())
+		if err != nil {
+			return nil, err
+		}
```
Wondertan, line 282: As chain ID is optional, the validation should also be optional. Do validate only if chainId is not ""

Wondertan (commented):
I really wish we make chainID detection automatic. E.g., when we call Head, we trust headers coming from trustedPeers and from their headers, we can detect what chainId we are no and then validate all the further headers against it, but that doesn't seem to be the best idea and can cause confusion.

If there are no other ideas for automatic detection, let's just move forward with this approach and also fix one last nit.

Wondertan:
No we can just use newest Network option here @vgonkivs

libs/header/p2p/exchange.go
```diff
@@ -226,22 +226,30 @@ func (ex *Exchange[H]) performRequest(
...
 		//nolint:gosec // G404: Use of weak random number generator
 		idx := rand.Intn(len(ex.trustedPeers))
-		ctx, cancel := context.WithTimeout(ctx, ex.Params.TrustedPeersRequestTimeout)
 
-		h, err := ex.request(ctx, ex.trustedPeers[idx], req)
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		default:
+		}
```
Wondertan, line 238: Why?

Wondertan: request will respect it anyway

libs/header/p2p/exchange.go
```diff
@@ -226,22 +226,30 @@ func (ex *Exchange[H]) performRequest(
...
+		reqCtx, cancel := context.WithTimeout(ctx, ex.Params.TrustedPeersRequestTimeout)
+		h, err := ex.request(reqCtx, ex.trustedPeers[idx], req)
 		cancel()
 		switch err {
 		default:
+			reqErr = err
 			log.Debug(err)
 			continue
-		case context.Canceled, context.DeadlineExceeded, nil:
-			return h, err
```
Wondertan, line 242: Why?

vgonkivs: in case we perform `len(ex.trustedPeers)` that will fail , we will return the last stored error. If `context.DeadlineExceeded`, it's fine IMO and we can try to get header from another peer.

libs/header/p2p/options.go
```diff
@@ -254,3 +256,14 @@ func WithMaxTrackerSize[T ClientParameters](size int) Option[T] {
...
+// WithChainID is a functional option that configures the
+// `chainID` parameter.
+func WithChainID[T ClientParameters](chainID string) Option[T] {
+	return func(p *T) {
+		switch t := any(p).(type) { //nolint:gocritic
+		case *ClientParameters:
+			t.chainID = chainID
+		}
+	}
+}
```
Wondertan, line 269: Can remove and use network id

vgonkivs:
Simplified @Wondertan

libs/header/p2p/exchange.go
vgonkivs: n/a

libs/header/p2p/exchange.go
vgonkivs: removed

libs/header/p2p/exchange.go
vgonkivs: reworked

Wondertan (approved):

renaynay (approved):

Issue #1474, opened 2022-12-12
RPC/API Improvements Tracking


distractedm1nd:
# RPC/Autogen

## First stable version
- [x] https://github.com/celestiaorg/celestia-node/issues/1345
- [x] https://github.com/celestiaorg/celestia-node/issues/1358
- [x] #425
- [ ] Add reflect tests to ensure compliance with RPC requirements
	- [ ] Context always exists as first parameter
	- [ ] Two return types, both JSON serializable, second one is always err
- [ ] API module definitions should match specification in ADR
- [ ] update ADR with overview of architecture

### Nice to haves
- [x] blob module 
- [x] improve UX around generating auth token

## Hardening
- [x] https://github.com/celestiaorg/celestia-node/issues/1401
- [x] OpenRPC Schema generation improvements for autogenerated clients
- [x] CLI over RPC
- [ ] https://github.com/celestiaorg/celestia-node/issues/1276
- [ ] https://github.com/celestiaorg/celestia-node/issues/739
- [ ] #2420

# Gateway:
- [ ] https://github.com/celestiaorg/celestia-node/issues/739
- [ ] Add JWT auth to writes
- [x] #1184

# Refactorings:
- [x] https://github.com/celestiaorg/celestia-node/issues/1375
- [ ] nodebuilder: extract module names into consts and reuse those consts where possible(https://github.com/celestiaorg/celestia-node/issues/1176)
- [x] Researching alternatives to `Internal` fields in API structs for `go-jsonrpc`
- [x] https://github.com/celestiaorg/celestia-node/issues/1401

# General API endpoint improvements / feature requests
- [ ] `header.GetAt` endpoint
- [x] consider deprecating `header.GetRangeByHeight` in favour of `header.GetVerifiedRangeByHeight`
- [x] #1942

## Future considerations
- [ ] #2811 
- [x] #1143
- [ ] #1302
- [ ] #1144

Wondertan:
We must find a way to add options to our APIs for optional things like fee prices, versioning, etc.
The canonical way of using functional patterns would not be compatible with our RPC autogen. Though there might be a brilliant way to make the functional pattern work with reflection, but it's not known nor implemented in the RPC lib.

As a simple step, I propose adding an optional structure to each method, requiring an option. That structure would contain all the optional fields of the method. This way is compatible with our RPC lib and should be easy to replicate on different languages.

ramin:
[...]

PR #1748, opened 2023-02-14
feat(nodebuilder/header)!: Implement remainder of `header.Module`

renaynay:
This PR implements several additional endpoints for the `header.Module` to bring it closer to its specification (defined in the public API ADR).

This PR is breaking as it includes renaming for an endpoint: `header.Head` is now `header.LocalHead`

Two notes: 
* this PR does **not** include the `header.Subscribe` endpoint as gomock does not yet support generics (so we have to wait for that in order to be able to test the module properly)
* Some additional decisions need to be made around the header API (for example, deprecating `GetRangeByHeight` and putting `GetVerifiedRangeByHeight` in its place) that will follow in later PRs


nodebuilder/header/header.go
```diff
@@ -4,32 +4,78 @@ import (
...
+		to uint64,
+	) ([]*header.ExtendedHeader, error)
 	// GetByHeight returns the ExtendedHeader at the given height, blocking
 	// until header has been processed by the store or context deadline is exceeded.
 	GetByHeight(context.Context, uint64) (*header.ExtendedHeader, error)
-	// Head returns the ExtendedHeader of the chain head.
-	Head(context.Context) (*header.ExtendedHeader, error)
+
 	// IsSyncing returns the status of sync
 	IsSyncing(context.Context) bool
```
Wondertan, line 33: We should return the state of syncer instead, by which we can determine if its syncing or not

nodebuilder/header/header.go
```diff
@@ -4,32 +4,78 @@ import (
...
+	// Get returns the header of the given hash from the node's header store.
+	Get(ctx context.Context, hash libhead.Hash) (*header.ExtendedHeader, error)
+	// GetVerifiedRangeByHeight returns the given range [from:to) of ExtendedHeaders
+	// from the node's header store and verifies that the returned headers are
+	// adjacent to each other.
+	GetVerifiedRangeByHeight(
+		ctx context.Context,
+		from *header.ExtendedHeader,
+		to uint64,
+	) ([]*header.ExtendedHeader, error)
```
Wondertan, line 27: I would omit Verified because our API should always return them verified. Our Getter interface will do this as well at some point

nodebuilder/header/header.go
```diff
@@ -4,32 +4,78 @@ import (
...
 // Module exposes the functionality needed for querying headers from the network.
 // Any method signature changed here needs to also be changed in the API struct.
 //
 //go:generate mockgen -destination=mocks/api.go -package=mocks . Module
 type Module interface {
+	// Head returns the ExtendedHeader of the chain head.
+	Head(context.Context) (*header.ExtendedHeader, error)
+
+	// Get returns the header of the given hash from the node's header store.
+	Get(ctx context.Context, hash libhead.Hash) (*header.ExtendedHeader, error)
```
Wondertan, line 19: We might want to wait for GetAt and introduce it here as well

nodebuilder/header/header.go
```diff
@@ -4,32 +4,78 @@ import (
...
 	GetByHeight(context.Context, uint64) (*header.ExtendedHeader, error)
-	// Head returns the ExtendedHeader of the chain head.
-	Head(context.Context) (*header.ExtendedHeader, error)
+
 	// IsSyncing returns the status of sync
 	IsSyncing(context.Context) bool
+	// WaitSync blocks until the header Syncer is synced to network head.
+	WaitSync(ctx context.Context) error
+	// SyncHead provides the Syncer's view of the current network head.
+	SyncHead(ctx context.Context) (*header.ExtendedHeader, error)
```
Wondertan, line 37: What about NetworkHead instead? Together with LocalHead

nodebuilder/header/header.go
```diff
@@ -4,32 +4,78 @@ import (
...
+
+	// Subscribe creates long-living Subscription for validated ExtendedHeaders.
+	// Multiple Subscriptions can be created.
```
Wondertan, line 40:
```suggestion
	// Subscribe creates long-living Subscription for newly validated ExtendedHeaders.
	// Multiple Subscriptions can be created.
```

nodebuilder/header/header.go
```diff
@@ -4,32 +4,78 @@ import (
...
+			context.Context,
+			*header.ExtendedHeader,
+			uint64,
+		) ([]*header.ExtendedHeader, error) `perm:"public"`
 		GetByHeight func(context.Context, uint64) (*header.ExtendedHeader, error) `perm:"public"`
-		Head        func(context.Context) (*header.ExtendedHeader, error)         `perm:"public"`
-		IsSyncing   func(context.Context) bool                                    `perm:"public"`
+		IsSyncing   func(context.Context) bool                                    `perm:"read"`
+		WaitSync    func(ctx context.Context) error                               `perm:"read"`
+		SyncHead    func(ctx context.Context) (*header.ExtendedHeader, error)     `perm:"read"`
```
Wondertan, line 61: SyncHead(NetworkHead) should be public IMO, while Head(LocalHead) should be read instead

Wondertan (changes requested):
Trully excited about this one

nodebuilder/header/header.go
renaynay: We need to then discuss removing the method entirely from the interface (the non-verified range) 

cc @vgonkivs

renaynay: I'm happy to do this but in a subsequent PR pls @Wondertan so that it's an atomic change

nodebuilder/header/header.go
renaynay: That can come later, also as atomic change (e.g. `Add heightsub/WaitFor` endpoint)

nodebuilder/header/header.go
renaynay: kk

nodebuilder/header/header.go
renaynay: removed this endpoint for now as mocks don't support generics yet.

nodebuilder/header/header.go
renaynay: done

nodebuilder/header/header.go
renaynay: tracked in #1474

nodebuilder/header/header.go
renaynay: tracked in #1474

nodebuilder/header/header.go
renaynay: In follow-up PR - I think we should have both. `IsSyncing` is just a bool and doesn't require user to import sync.State

Wondertan: There is no reason to have one more method if there is another one from which we can simply derive the same information

Wondertan: I don't get why would we do this in another PR if the goal of this PR is to extend the header module

renaynay: > There is no reason to have one more method if there is another one from which we can simply derive the same information

for convenience - the same way we will have blob module. I think people would care more about whether node is Syncing than to have the entire sync state.

> I don't get why would we do this in another PR if the goal of this PR is to extend the header module

Because we already have this method `IsSyncing` and i want to keep it as is, and then have another method to return all of syncstate which is an additional endpoint, so i'd rather do it later.

docs/adr/adr-009-public-api.md
```diff
@@ -91,25 +91,27 @@ microservice architectures.
...
+Subscribe(context.Context) (Subscription, error)
 // SyncState returns the current state of the header Syncer. 
-SyncState() sync.State
-// SyncWait blocks until the header Syncer is synced to network head. 
-SyncWait(ctx context.Context) error
-// SyncHead provides the Syncer's view of the current network head.
-SyncHead(ctx context.Context) (*header.ExtendedHeader, error)
+SyncState(context.Context) sync.State
+// WaitSync blocks until the header Syncer is synced to network head. 
+WaitSync(ctx context.Context) error
```
Wondertan, line 112: I like SyncWait more because it's nicely grouped with SyncState then

Wondertan: We might also want to call `SyncHead` instead of `NetworkHead`

nodebuilder/header/header.go
Wondertan: In private, we agreed to remove it

docs/adr/adr-009-public-api.md
renaynay: Let's call it network head, that's easier to understand.

libs/header/interface.go
```diff
@@ -113,7 +113,7 @@ type Getter[H Header] interface {
 	GetRangeByHeight(ctx context.Context, from, amount uint64) ([]H, error)
 
 	// GetVerifiedRange requests the header range from the provided Header and
-	// verifies that the returned headers are adjacent to each other.
+	// .
```
Wondertan, line 116: and above is redundant

Wondertan (approved):

Wondertan (approved):

liamsi (approved):
Record the ADR only. Changes make sense to me.

vgonkivs (approved):

Issue #1567, opened 2023-01-05
feat(header|syncer): Request verified range of headers from syncer


vgonkivs:
Self-explanatory

PR #1749, opened 2023-02-14
refactoring(header/syncer): request verified range of headers

vgonkivs:
## Overview
Resolves https://github.com/celestiaorg/celestia-node/issues/1567

## Checklist

### DIFF
--- a/libs/header/p2p/session.go
+++ b/libs/header/p2p/session.go
@@ -74,7 +74,7 @@ func (s *session[H]) getRangeByHeight(
 	ctx context.Context,
 	from, amount, headersPerPeer uint64,
 ) ([]H, error) {
-	log.Debugw("requesting headers", "from", from, "amount", amount)
+	log.Debugw("requesting headers", "from", from, "to", from+amount-1) // -1 need to exclude to+1 height
 
 	requests := prepareRequests(from, amount, headersPerPeer)
 	result := make(chan []H, len(requests))
--- a/libs/header/p2p/session.go
+++ b/libs/header/p2p/session.go
@@ -74,7 +74,7 @@ func (s *session[H]) getRangeByHeight(
 	ctx context.Context,
 	from, amount, headersPerPeer uint64,
 ) ([]H, error) {
-	log.Debugw("requesting headers", "from", from, "to", from+amount)
+	log.Debugw("requesting headers", "from", from, "amount", amount)
 
 	requests := prepareRequests(from, amount, headersPerPeer)
 	result := make(chan []H, len(requests))
@@ -104,6 +104,11 @@ LOOP:
 	sort.Slice(headers, func(i, j int) bool {
 		return headers[i].Height() < headers[j].Height()
 	})
+
+	log.Debugw("received headers range",
+		"from", headers[0].Height(),
+		"to", headers[len(headers)-1].Height(),
+	)
 	return headers, nil
 }
 
--- a/libs/header/sync/ranges.go
+++ b/libs/header/sync/ranges.go
@@ -33,6 +33,21 @@ func (rs *ranges[H]) head() H {
 	return head.Head()
 }
 
+// cached returns all headers sub-ranges of headers that could be found in between the requested range.
+func (rs *ranges[H]) cached(fromHeader H, to uint64) [][]H {
+	r := rs.FindAllRangesWithin(uint64(fromHeader.Height()+1), to) // start looking for headers for the next height
+	if len(r) == 0 {
+		return nil
+	}
+
+	out := make([][]H, len(r))
+	for i, headersRange := range r {
+		cached, _ := headersRange.Before(to)
+		out[i] = append(out[i], cached...)
+	}
+	return out
+}
+
 // Add appends the new Header to existing range or starts a new one.
 // It starts a new one if the new Header is not adjacent to any of existing ranges.
 func (rs *ranges[H]) Add(h H) {
@@ -81,6 +96,26 @@ func (rs *ranges[H]) FirstRangeWithin(start, end uint64) (*headerRange[H], bool)
 	}
 
 	return nil, false
+}
+
+// FindAllRangesWithin returns all headerRanges that apply the requested range.
+// Returned ranges contain `start` within their bounds.
+// NOTE: ranges do not need to contain `end`.
+// There could be multiple headerRanges:
+// start = 2 , end 40
+// headerRange0 = [3;20];
+// headerRange1 = [30; 35];
+func (rs *ranges[H]) FindAllRangesWithin(start, end uint64) []*headerRange[H] {
+	ranges := make([]*headerRange[H], 0)
+	for _, r := range rs.ranges {
+		if r.Empty() {
+			continue
+		}
+		if r.start >= start && r.start <= end {
+			ranges = append(ranges, r)
+		}
+	}
+	return ranges
 }
 
 // First provides a first non-empty range, while cleaning up empty ones.
--- a/libs/header/sync/ranges.go
+++ b/libs/header/sync/ranges.go
@@ -33,21 +33,6 @@ func (rs *ranges[H]) head() H {
 	return head.Head()
 }
 
-// cached returns all headers sub-ranges of headers that could be found in between the requested range.
-func (rs *ranges[H]) cached(fromHeader H, to uint64) [][]H {
-	r := rs.FindAllRangesWithin(uint64(fromHeader.Height()+1), to) // start looking for headers for the next height
-	if len(r) == 0 {
-		return nil
-	}
-
-	out := make([][]H, len(r))
-	for i, headersRange := range r {
-		cached, _ := headersRange.Before(to)
-		out[i] = append(out[i], cached...)
-	}
-	return out
-}
-
 // Add appends the new Header to existing range or starts a new one.
 // It starts a new one if the new Header is not adjacent to any of existing ranges.
 func (rs *ranges[H]) Add(h H) {
@@ -81,41 +66,6 @@ func (rs *ranges[H]) Add(h H) {
 		// once we start rcving them again we save those in new range
 		// so 'Syncer.findHeaders' can fetch what was missed
 	}
-}
-
-// FirstRangeWithin checks if the first range is within a given height span [start:end]
-// and returns it.
-func (rs *ranges[H]) FirstRangeWithin(start, end uint64) (*headerRange[H], bool) {
-	r, ok := rs.First()
-	if !ok {
-		return nil, false
-	}
-
-	if r.start >= start && r.start <= end {
-		return r, true
-	}
-
-	return nil, false
-}
-
-// FindAllRangesWithin returns all headerRanges that apply the requested range.
-// Returned ranges contain `start` within their bounds.
-// NOTE: ranges do not need to contain `end`.
-// There could be multiple headerRanges:
-// start = 2 , end 40
-// headerRange0 = [3;20];
-// headerRange1 = [30; 35];
-func (rs *ranges[H]) FindAllRangesWithin(start, end uint64) []*headerRange[H] {
-	ranges := make([]*headerRange[H], 0)
-	for _, r := range rs.ranges {
-		if r.Empty() {
-			continue
-		}
-		if r.start >= start && r.start <= end {
-			ranges = append(ranges, r)
-		}
-	}
-	return ranges
 }
 
 // First provides a first non-empty range, while cleaning up empty ones.
--- a/libs/header/sync/sync.go
+++ b/libs/header/sync/sync.go
@@ -254,21 +254,31 @@ func (s *Syncer[H]) processHeaders(
 	fromHead H,
 	to uint64,
 ) (err error) {
-	cached := s.pending.cached(fromHead, to)
-	for _, cachedRange := range cached {
-		if fromHead.Height()+1 != cachedRange[0].Height() {
+	for {
+		headersRange, ok := s.pending.First()
+		if !ok {
+			break
+		}
+
+		headers, amount := headersRange.Before(to)
+		if amount == 0 {
+			break
+		}
+
+		if fromHead.Height()+1 != headers[0].Height() {
 			// make an external request
-			err = s.requestHeaders(ctx, fromHead, uint64(cachedRange[0].Height())-1)
+			err = s.requestHeaders(ctx, fromHead, uint64(headers[0].Height()-1))
 			if err != nil {
 				return err
 			}
 		}
+
 		// apply cached headers
-		if err = s.storeHeaders(ctx, cachedRange); err != nil {
+		if err = s.storeHeaders(ctx, headers); err != nil {
 			return err
 		}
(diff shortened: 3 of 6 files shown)


renaynay (approved):
* FirstRangeWithin is dangling - should we remove it? 
* So is `First` in ranges -- should we remove?

libs/header/sync/ranges.go
```diff
@@ -83,6 +98,26 @@ func (rs *ranges[H]) FirstRangeWithin(start, end uint64) (*headerRange[H], bool)
...
 }
 
+// FindAllRangesWithin returns all headerRanges that apply the requested range.
+// Returned ranges contain `start` within their bounds.
+// NOTE: ranges do not need to contain `end`.
+// There could be multiple headerRanges:
+// start = 2 , end 40
+// headerRange0 = [3;20];
+// headerRange1 = [30; 35];
+func (rs *ranges[H]) FindAllRangesWithin(start, end uint64) []*headerRange[H] {
```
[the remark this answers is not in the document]
vgonkivs: just for consistency. All ranges methods are exported...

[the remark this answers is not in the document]
renaynay: ahhh i wish they were not. lol but whatever

Wondertan, line 108: * We allocate here an additional slice, unnecessary
* We leak empty ranges here. The canonical First method below was responsible for cleaning them

libs/header/sync/ranges.go
```diff
@@ -33,6 +33,21 @@ func (rs *ranges[H]) head() H {
...
 }
 
+// cached returns all headers sub-ranges of headers that could be found in between the requested range.
+func (rs *ranges[H]) cached(fromHeader H, to uint64) [][]H {
+	r := rs.FindAllRangesWithin(uint64(fromHeader.Height()+1), to) // start looking for headers for the next height
+	if len(r) == 0 {
+		return nil
+	}
+
+	out := make([][]H, len(r))
```
Wondertan, line 43: Another allocation for very an unclear reason

libs/header/sync/ranges.go
```diff
@@ -33,6 +33,21 @@ func (rs *ranges[H]) head() H {
 	return head.Head()
 }
 
+// cached returns all headers sub-ranges of headers that could be found in between the requested range.
```
derrandz, line 36:
```suggestion
// cached returns all headers contiguous sub-ranges of headers that could be found within the requested range.
```
is this what you meant?

Wondertan (changes requested):

libs/header/sync/ranges.go
```diff
@@ -83,6 +98,26 @@ func (rs *ranges[H]) FirstRangeWithin(start, end uint64) (*headerRange[H], bool)
 	return nil, false
 }
 
+// FindAllRangesWithin returns all headerRanges that apply the requested range.
+// Returned ranges contain `start` within their bounds.
+// NOTE: ranges do not need to contain `end`.
```
derrandz, line 103: Can you give an example in this function's documentation of a case without an end?

libs/header/sync/sync.go
```diff
@@ -256,49 +246,72 @@ func (s *Syncer[H]) doSync(ctx context.Context, fromHead, toHead H) (err error)
...
+// If some headers are missing, it starts requesting them from the network.
+// All headers that will be received are verified to be contiguous.
+func (s *Syncer[H]) processHeaders(
+	ctx context.Context,
+	fromHead H,
+	to uint64,
+) (err error) {
+	cached := s.pending.cached(fromHead, to)
+	for _, cachedRange := range cached {
+		if fromHead.Height()+1 != cachedRange[0].Height() {
```
derrandz, line 259: are the `cached` ranges sorted? if not, we may have for instance:

- range `[10, 20]`
- range `[7, 25]`

and have `fromHead.Height() + 1` be 7, and thus on the first iteration end up doing an "unnecessary" external request to retrieve an already existing/cached height.

derrandz (changes requested):
Dismiss my review if my assumptions are not correct.

renaynay:
@derrandz if you only have questions related to a PR, then don't request changes - but just comment. Otherwise your review becomes blocking.

libs/header/sync/ranges.go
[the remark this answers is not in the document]
vgonkivs: removed at all

vgonkivs: no more leaks. No functions - no problems 😄

libs/header/sync/ranges.go
vgonkivs: yes. but I removed it

libs/header/sync/ranges.go
vgonkivs: n/a

libs/header/sync/sync.go
vgonkivs: ranges are sorted

libs/header/sync/sync_test.go
```diff
@@ -281,16 +281,13 @@ func TestSyncer_FindHeadersReturnsCorrectRange(t *testing.T) {
...
-	require.Equal(t, h[0].Height(), int64(2))
-	require.Equal(t, h[len(h)-1].Height(), int64(21))
-	header := h[0]
-	for i := 1; i < len(h); i++ {
-		require.NoError(t, header.Verify(h[i]))
-		header = h[i]
-	}
+	require.NoError(t, err)
+	err = syncer.processHeaders(ctx, head, 21)
+	time.Sleep(time.Millisecond * 100) // give time to store all fetched headers
```
renaynay, line 286: can you use syncedHead to check this

renaynay (approved):

walldiss (dismissed):
Nice simplification!

libs/header/sync/sync_test.go
vgonkivs: yes b64d22fdd473e379102bffbfb899a98bd670a91d

renaynay: No i mean, can we remove the sleep and just use syncedHead?

vgonkivs: okay

renaynay (approved):

derrandz (dismissed):

libs/header/sync/sync.go
```diff
@@ -256,50 +246,83 @@ func (s *Syncer[H]) doSync(ctx context.Context, fromHead, toHead H) (err error)
...
+	to uint64,
+) error {
+	amount := to - uint64(fromHead.Height())
+	// start requesting headers until amount will be 0
+	for amount > 0 {
+		size := s.Params.MaxRequestSize
+		if amount < size {
+			size = amount
+		}
+		headers, err := s.exchange.GetVerifiedRange(ctx, fromHead, size)
```
Wondertan, line 300: Waiiit. I just realized. So GetVerifiedRange for exchange take an amount and store's one takes actual heigts :joy: We should fix this ASAP tbh

libs/header/sync/sync.go
```diff
@@ -256,50 +246,83 @@ func (s *Syncer[H]) doSync(ctx context.Context, fromHead, toHead H) (err error)
...
-		amount = s.Params.MaxRequestSize
-	}
+		// check that returned range is adjacent to `fromHead`
+		if fromHead.Height()+1 != headers[0].Height() {
+			// make an external request
+			err = s.requestHeaders(ctx, fromHead, uint64(headers[0].Height()-1))
+			if err != nil {
+				return err
+			}
+		}
```
Wondertan, line 275: nit: Also do a call + if check in one line like you do for storeHeaders

libs/header/sync/sync.go
Wondertan: Not necessary in this PR, but it's super cofooosing

Wondertan (changes requested):
Love it! 

Besides comments below, let's also rename and update the description of the PR to reflect what actually happens here. Like this is clearly a refactoring not the feature and besides just intergating GetVerifiedRange this also nicely cleans up the logic

libs/header/sync/sync.go
vgonkivs: Yeap, to make conistent

libs/header/sync/sync.go
```diff
@@ -256,50 +246,83 @@ func (s *Syncer[H]) doSync(ctx context.Context, fromHead, toHead H) (err error)
...
-func (s *Syncer[H]) findHeaders(ctx context.Context, from, to uint64) ([]H, error) {
-	amount := to - from + 1 // + 1 to include 'to' height as well
-	if amount > s.Params.MaxRequestSize {
-		to = from + s.Params.MaxRequestSize - 1 // `from` is already included in range
-		amount = s.Params.MaxRequestSize
-	}
+		// check that returned range is adjacent to `fromHead`
+		if fromHead.Height()+1 != headers[0].Height() {
+			// make an external request
+			err = s.requestHeaders(ctx, fromHead, uint64(headers[0].Height()-1))
```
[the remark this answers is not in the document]
vgonkivs: `fromHead.Height()+1 != headers[0].Height()` -> this condition means that the cached range is not adjacent to `fromHead` and we need an extra request to fill missed headers.

libs/header/p2p/session.go
```diff
@@ -236,25 +241,38 @@ func (s *session[H]) processResponse(responses []*p2p_pb.HeaderResponse) ([]H, e
...
-	// if `from` is empty, then additional validation for the header`s range is not needed.
+	// if `s.from` is empty, then additional validation for the header`s range is not needed.
 	if s.from.IsZero() {
 		return nil
 	}
 
+	if len(headers) == 0 {
+		return nil
+	}
+
```
Wondertan, line 254: Unnecessary

libs/header/p2p/session.go
```diff
@@ -236,25 +241,38 @@ func (s *session[H]) processResponse(responses []*p2p_pb.HeaderResponse) ([]H, e
...
 	trusted := s.from
-	// verify that the whole range is valid.
+	// verify that the whole range is valid and adjacent.
 	for i := 0; i < len(headers); i++ {
```
Wondertan, line 257: Let's use range here instead

libs/header/p2p/session.go
```diff
@@ -236,25 +241,38 @@ func (s *session[H]) processResponse(responses []*p2p_pb.HeaderResponse) ([]H, e
...
 			return err
 		}
-		if trusted.Height()+1 != headers[i].Height() {
-			// Exchange requires requested ranges to always consist of adjacent headers
-			return fmt.Errorf("peer sent valid but non-adjacent header")
-		}
 
+		// extra check for the adjacency should be performed only for the received range,
+		// as `s.from` can't be adjacent to this range
+		if trusted.Height() != s.from.Height() {
```
Wondertan, line 265: It's worth mentioning why. Because headers are received out of order

Wondertan (commented):
It's not misc either 😭 It's a refactoring!

Also, I kindly ask you to extract the bug fix in the separate PR.

vgonkivs:
https://github.com/celestiaorg/celestia-node/pull/1871

@Wondertan

libs/header/p2p/session.go
```diff
@@ -74,7 +74,7 @@ func (s *session[H]) getRangeByHeight(
 	ctx context.Context,
 	from, amount, headersPerPeer uint64,
 ) ([]H, error) {
-	log.Debugw("requesting headers", "from", from, "to", from+amount)
+	log.Debugw("requesting headers", "from", from, "amount", amount)
```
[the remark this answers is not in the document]
vgonkivs: from - to will be printed in the end of the session

libs/header/sync/sync.go
```diff
@@ -256,50 +246,83 @@ func (s *Syncer[H]) doSync(ctx context.Context, fromHead, toHead H) (err error)
...
 
-		// apply cached range if any
-		cached, ln := r.Before(to)
-		out = append(out, cached...)
-		from += ln
+// storeHeaders updates store with new headers and updates current syncedHead.
+func (s *Syncer[H]) storeHeaders(ctx context.Context, headers []H) error {
+	stored, err := s.store.Append(ctx, headers...)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
Wondertan: It's not possible anymore because verification now happens on the Exchange side

libs/header/sync/sync.go
```diff
@@ -256,50 +246,83 @@ func (s *Syncer[H]) doSync(ctx context.Context, fromHead, toHead H) (err error)
...
 // processHeaders gets and stores headers starting at the given 'from' height up to 'to' height -
 // [from:to]
-func (s *Syncer[H]) processHeaders(ctx context.Context, from, to uint64) (int, error) {
-	headers, err := s.findHeaders(ctx, from, to)
-	if err != nil {
-		return 0, err
-	}
+// processHeaders checks headers in pending cache that apply to the requested range.
+// If some headers are missing, it starts requesting them from the network.
+// All headers that will be received are verified to be contiguous.
```
Wondertan, line 251: This info seems redundant. Also Exchange verifies this now, not the Syncer
```suggestion
```

Wondertan (approved):
LGTM. Still worried about that bug you showed me. It is good that it does not happen anymore, but IIRC we didn't chang anything, so it may still appear

libs/header/p2p/session.go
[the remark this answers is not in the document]
renaynay: Yes but the logs cloud, so let's just keep it as a range instead of amount.

libs/header/sync/sync.go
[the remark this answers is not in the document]
renaynay: Okay then we should document that in doc for `storeHeaders` IMO since it's now implicit to the syncer and won't be obvious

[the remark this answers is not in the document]
Wondertan: I am not opposed to the document but wanna mention that it is explicit! Even overly explicit with GetVerifiedRange. Instead, we should just have GetRange that always guarantees correctness of the data.

vgonkivs:
> Still worried about that bug you showed me. It is good that it does not happen anymore, but IIRC we didn't chang anything, so it may still appear

Resolved in https://github.com/celestiaorg/celestia-node/pull/1878

libs/header/sync/sync.go
```diff
@@ -202,7 +202,7 @@ func (s *Syncer[H]) sync(ctx context.Context) {
 	}
 
 	log.Infow("syncing headers",
-		"from", header.Height(),
+		"from", header.Height()+1,
```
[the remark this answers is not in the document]
vgonkivs: Think it will be confusing (from; to]. Let's keep it consistent [from;to]

libs/header/sync/sync.go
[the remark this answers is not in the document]
renaynay: It is explicit in the general scope of the syncer but not to that function itself, but sure.

renaynay (approved):

Wondertan (approved):

Issue #1722, opened 2023-02-09
swamp: reduce logs in swamp tests


vgonkivs:
Currently, our swamp tests generate a lot of unnecessary logs during execution.
For example, `TestFullReconstructionFromLight` generates 500 mb of logs during execution. 

[Proposal]: let's add an option to swamp that will turn off unneeded libp2p/pubsub logs.

PR #1754, opened 2023-02-15
feat(swamp): add option that allows to override log level

vgonkivs:
## Overview
Resolves https://github.com/celestiaorg/celestia-node/issues/1722
## Checklist


nodebuilder/tests/reconstruct_test.go
```diff
@@ -103,7 +103,17 @@ func TestFullReconstructFromLights(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), swamp.DefaultTestTimeout)
 
 	t.Cleanup(cancel)
-	sw := swamp.NewSwamp(t, swamp.WithBlockTime(btime))
+	sw := swamp.NewSwamp(t, swamp.WithBlockTime(btime), swamp.WithLogLevel(
+		[]string{
+			"bitswap_network",
```
distractedm1nd, line 108: can you add "dagstore" and "dagstore/upgrader"? They also spam a lot of logs

vgonkivs: added as an example. Now *you* are the artist of your own logs 😄 

But, sure, will add dagstore

nodebuilder/tests/swamp/config.go
```diff
@@ -35,3 +37,20 @@ func WithBlockTime(t time.Duration) Option {
 		c.Tendermint.Consensus.SkipTimeoutCommit = false
 	}
 }
+
+// WithLogLevel overrides basic log levels for test purposes
+func WithLogLevel(components []string, levels []string) Option {
+	return func(_ *Components) {
+		if len(components) != len(levels) && len(levels) != 1 {
```
derrandz, line 44:
```suggestion
		if len(levels) == 0 || len(components) == 0 || len(levels) != len(components) {
```

nodebuilder/tests/swamp/config.go
```diff
@@ -35,3 +37,20 @@ func WithBlockTime(t time.Duration) Option {
...
+			panic("inconsistent amount of components and levels")
+		}
+
+		level := levels[0]
+		for i, c := range components {
+			if len(levels) == len(components) {
+				level = levels[i]
```
derrandz, line 51:
```suggestion
```

derrandz (changes requested):
I think there was a fault in logic, lmk if I understood what you are trying to do correctly

nodebuilder/tests/swamp/config.go
vgonkivs: instead of passing the log level for each component, you can pass only one that will be applied for all.

nodebuilder/tests/swamp/config.go
vgonkivs: Responded in the thread above

nodebuilder/tests/swamp/config.go
vgonkivs: So this condition means that you can have log level for each component or one for all components. Otherwise, panic

nodebuilder/tests/swamp/config.go
```diff
@@ -35,3 +37,20 @@ func WithBlockTime(t time.Duration) Option {
 		c.Tendermint.Consensus.SkipTimeoutCommit = false
 	}
 }
+
+// WithLogLevel overrides basic log levels for test purposes
+func WithLogLevel(components []string, levels []string) Option {
+	return func(_ *Components) {
+		if len(components) != len(levels) && len(levels) != 1 {
+			panic("inconsistent amount of components and levels")
```
Bidon15, line 45: can we log before we panic where was the mismatch by printing all levels + components? 🙏 

This gives much better assurance for future debugging sessions 😅

nodebuilder/tests/swamp/config.go
```diff
@@ -35,3 +37,20 @@ func WithBlockTime(t time.Duration) Option {
...
+		}
+
+		level := levels[0]
```
Bidon15, line 48: [nit][proposl][non-blocking]: a one-liner comment might help

Bidon15 (commented):
Legend

nodebuilder/tests/swamp/config.go
```diff
@@ -35,3 +37,20 @@ func WithBlockTime(t time.Duration) Option {
 		c.Tendermint.Consensus.SkipTimeoutCommit = false
 	}
 }
+
+// WithLogLevel overrides basic log levels for test purposes
+func WithLogLevel(components []string, levels []string) Option {
```
walldiss, line 42: How about make it 
```
func WithLogLevel(level string, components ...string) Option
```
So log level could be applied to multiple components. 
Multiple levels could be applied with several WithLogLevel calls with corresponding levels. 
This could simplify logic and usage.

vgonkivs: Yeap, that makes a lot of thanks. Thank you. I'll rework it

derrandz (approved):
OK I see, so this is a forcing function. LGTM

Bidon15 (approved):

walldiss (approved):

nodebuilder/tests/reconstruct_test.go
```diff
@@ -103,9 +103,19 @@ func TestFullReconstructFromLights(t *testing.T) {
...
+		"bitswap_network",
+		"bitswap-server",
+		"bitswap-client",
+		"net/identify",
+		"mocknet",
+		"pubsub",
+		"basichost",
+		"bs:sprmgr",
+		"dagstore/upgrader",
+	))
```
Wondertan, line 116: Don't think it has to be an option as it applies the levels globally rather than to the specific swamp. 
When we do `make test` all the tests are run in one process with shared global space, meaning that once this test is executed it changes the logging for all subsequent tests and swamps, rather than the particular Swamp. 

Not a strong opinion here, it just doesn't make sense to me

vgonkivs: It makes a lot of sense when you are checking 9mb of logs instead of 500mb

Wondertan: I am not saying to remove changing log level lol, but to make it directly with log.SetLogLevel rather than with the option

vgonkivs: It for different test you may need different logs. And I assume it will turn into a mess with manual log level editing. But I've got your point. Will try to find another pragmatic way.

PR #1766, opened 2023-02-17
feat(libhead): introduce peer scoring params for HeaderSub topic

Wondertan:
Extracted out of #1756


libs/header/p2p/helpers.go
```diff
@@ -20,7 +20,7 @@ func protocolID(protocolSuffix string) protocol.ID {
 	return protocol.ID(fmt.Sprintf("/header-ex/v0.0.3/%s", protocolSuffix))
 }
 
-func pubsubTopicID(protocolSuffix string) string {
+func PubsubTopicID(protocolSuffix string) string {
```
vgonkivs, line 23: Why exported though?

Wondertan: It will be required for the setup

As you can see here, we don't set/use scores anywhere in this PR. Using setting those will be done in another PR.
It can only be done as an option to PubSub which takes a map of topic name -> score and to get the topic name we need this exported.

vgonkivs: ok. I see. Thanks for the explanation. APPPPROOOOOVED

vgonkivs (approved):

libs/header/p2p/gossip_score.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"time"
+
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
+)
+
+// GossibSubScore provides a set of recommended parameters for header GossipSub topic, a.k.a
+// HeaderSub.
+var GossibSubScore = &pubsub.TopicScoreParams{
```
renaynay, line 11: why define these here instead of in the nodebuilder?

Wondertan: So that rollkit can reuse them. It's improbable that they would need a different setup

renaynay: they can copy our code then - if our libs are really meant to be used generally then we shouldn't include our network-specific configuration options in there.

Wondertan: The thing is, it's not that network specific. I don't see a problem with having a general recommendation here. Users may opt-in or opt-out. But this anyway would simplify their lives

renaynay (approved):
🏳️

PR #1773, opened 2023-02-17
chore(modp2p): update blocktime to match core

distractedm1nd:
## Overview
Updates blocktime to match core.

I am not yet updating the default daser sync timeout because of the downstream issues it may cause for large (128 sq size) blocks in the near term (before async cascade getter is merged).

If IPLD timeout may take up to 28 seconds for sq size 128, then the fallback needs to run for a minimum of 30 seconds. With sync cascade getter the daser SamplingTimeout is split in half, meaning it should stay a minute right now.


Wondertan (approved):
Very nice

nodebuilder/header/module.go
```diff
@@ -28,7 +28,6 @@ func ConstructModule(tp node.Type, cfg *Config) fx.Option {
 	baseComponents := fx.Options(
 		fx.Supply(*cfg),
 		fx.Error(cfgErr),
-		fx.Supply(modp2p.BlockTime),
```
renaynay, line 31: ? where do u supply it then

distractedm1nd: it doesn't need to be supplied

walldiss (approved):

PR #1777, opened 2023-02-18
refactor(libhead): unify VerifyAdjacent and VerifyNonAdjacent into Verify

Wondertan:
Last beautification for read-only header interface. We didn't do it previously with @tzdybal, as the syncing logic depended on adjacency through ErrNotNotAdjacent, which VerifyAdjacent returned. The solution is to keep returning this error but from Append.                                                                                                                                                                                                                                                                   

Additionally, this solves an edge case found by @gupadhyaya, where when Syncer starts and instantly requests and fails to apply a new network header due to adjacency.


Wondertan:
I also moved /headertest from root to /header. It shouldn't be in the root

vgonkivs:
conflicts

Wondertan:
Lint is borken

Wondertan:
Merge over GH didn't go well, so rebasing locally

libs/header/p2p/session.go
```diff
@@ -201,23 +203,18 @@ func (s *session[H]) validate(headers []H) error {
...
-	for i := 1; i < len(headers); i++ {
-		err = trusted.VerifyAdjacent(headers[i])
+	for i := 0; i < len(headers); i++ {
+		err := trusted.Verify(headers[i])
 		if err != nil {
 			return err
 		}
+		if trusted.Height()+1 != headers[i].Height() {
+			// Exchange requires requested ranges to always consist of adjacent headers
+			return fmt.Errorf("peer sent valid but undjacent header")
```
renaynay, line 215:
```suggestion
			return fmt.Errorf("peer sent valid but non-adjacent header")
```

libs/header/p2p/session.go
```diff
@@ -201,23 +203,18 @@ func (s *session[H]) validate(headers []H) error {
...
-		return nil
-	}
-
-	trusted := headers[0]
+	trusted := s.from
 	// verify that the whole range is valid.
-	for i := 1; i < len(headers); i++ {
-		err = trusted.VerifyAdjacent(headers[i])
+	for i := 0; i < len(headers); i++ {
+		err := trusted.Verify(headers[i])
```
renaynay, line 209: if last header passes verification, shouldn't trusted = header[i-1] then or no?

libs/header/sync/sync_head.go
```diff
@@ -183,6 +183,9 @@ func (s *Syncer[H]) validate(ctx context.Context, new H) pubsub.ValidationResult
 	return pubsub.ValidationAccept
 }
 
+// TODO(@Wondertan): We should request TrustingPeriod from the network's state params or
```
renaynay, line 186: can you put in issue?

renaynay (dismissed):

libs/header/sync/sync_head.go
Wondertan: Will revalidate this first

vgonkivs (approved):

libs/header/sync/sync_head.go
Wondertan: https://github.com/celestiaorg/celestia-node/issues/1857

renaynay (approved):

PR #1789, opened 2023-02-21
refactor(share/p2p): Discovery controls `ensurePeers` lifecycle

renaynay:
Extracts `ensurePeers` lifecycle management from `peer.Manager`, and privatizes `ensurePeers`.

As All the node types call it, so we can do it by default.


Wondertan (approved):
❤️

walldiss (approved):

Issue #1790, opened 2023-02-21
improvement(das / share/getters): cleanup sampling timeout


walldiss:
### Implementation ideas

Sampling timeouts are being set on multiple layers without relation to each other. Cleanup unnecessary timeouts.

Issue #10, opened 2021-04-13
share: Define reasonable timeout for `SharesAvailable` for `full` and `light`


Wondertan:
### Background 
To validate a block's availability, we need to randomly sample some of its parts by requesting them from the network. Network requests must have a timeout not to wait for a response infinitely. Also, the nature of the IPFS software does not specify any timeout for data requests for its users, thus not to wait endlessly, we have to specify an adequate limit for sample response time.

### Current state
I defined a random timeout for the request in 1 min that is not backed up by any rationale. 

### Proposed solution
I am not entirely about the proper way for finding ideal timings here, but I think we need to bench medium, numbers over a real-world environment and add on top slightly more additional time for inaccuracies.

Bidon15:
Relies on share-exchange protocol
Reasonable timeout should be defined properly after we get the telemetry data to analyse and decide

Still relevant post-mainnet status of 12/07/2022

PR #1791, opened 2023-02-21
refactor(das|share): Cleanup timeouts through sampling stack

walldiss:
## Overview

Resolves https://github.com/celestiaorg/celestia-node/issues/1790 and closes https://github.com/celestiaorg/celestia-node/issues/10

### DIFF
--- a/das/options.go
+++ b/das/options.go
@@ -46,12 +46,14 @@ type Parameters struct {
 func DefaultParameters() Parameters {
 	// TODO(@derrandz): parameters needs performance testing on real network to define optimal values
 	// (#1261)
+	concurrencyLimit := 16
 	return Parameters{
 		SamplingRange:           100,
-		ConcurrencyLimit:        16,
+		ConcurrencyLimit:        concurrencyLimit,
 		BackgroundStoreInterval: 10 * time.Minute,
 		SampleFrom:              1,
-		SampleTimeout:           15 * time.Second * 16, // block time * max amount of catchup workers
+		// SampleTimeout = block time * max amount of catchup workers
+		SampleTimeout: 15 * time.Second * time.Duration(concurrencyLimit),
 	}
 }
 
--- a/das/options.go
+++ b/das/options.go
@@ -36,7 +36,9 @@ type Parameters struct {
 	SampleFrom uint64
 
 	// SampleTimeout is a maximum amount time sampling of single block may take until it will be
-	// canceled
+	// canceled. High ConcurrencyLimit value may increase sampling time due to node resources being
+	// divided between parallel workers. SampleTimeout should be adjusted proportional to
+	// ConcurrencyLimit.
 	SampleTimeout time.Duration
 }
 
@@ -49,7 +51,7 @@ func DefaultParameters() Parameters {
 		ConcurrencyLimit:        16,
 		BackgroundStoreInterval: 10 * time.Minute,
 		SampleFrom:              1,
-		SampleTimeout:           time.Minute,
+		SampleTimeout:           15 * time.Second * 16, // block time * max amount of catchup workers
 	}
 }
 
--- a/das/options.go
+++ b/das/options.go
@@ -37,7 +37,7 @@ type Parameters struct {
 
 	// SampleTimeout is a maximum amount time sampling of single block may take until it will be
 	// canceled. High ConcurrencyLimit value may increase sampling time due to node resources being
-	// divided between parallel workers. SampleTimeout should be adjusted proportional to
+	// divided between parallel workers. SampleTimeout should be adjusted proportionally to
 	// ConcurrencyLimit.
 	SampleTimeout time.Duration
 }
--- a/nodebuilder/das/config.go
+++ b/nodebuilder/das/config.go
@@ -19,7 +19,7 @@ type Config das.Parameters
 // TODO(@derrandz): Address #1261
 func DefaultConfig() Config {
 	cfg := das.DefaultParameters()
-	cfg.SampleTimeout = modp2p.BlockTime * time.Duration(cfg.SampleTimeout)
+	cfg.SampleTimeout = modp2p.BlockTime * time.Duration(cfg.ConcurrencyLimit)
 	return Config(cfg)
 }
 
--- a/nodebuilder/das/config.go
+++ b/nodebuilder/das/config.go
@@ -2,8 +2,10 @@ package das
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/celestiaorg/celestia-node/das"
+	modp2p "github.com/celestiaorg/celestia-node/nodebuilder/p2p"
 )
 
 // Config contains configuration parameters for the DASer (or DASing process)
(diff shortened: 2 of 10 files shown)


vgonkivs (approved):

share/getters/utils.go
```diff
@@ -91,3 +92,17 @@ func verifyNIDSize(nID namespace.ID) error {
 	}
 	return nil
 }
+
+// splitCtxTimeout will split timeout stored in context by splitFactor and return max of the
+// result and minTimeout
```
renaynay, line 97:
```suggestion
// splitCtxTimeout will split timeout stored in context by splitFactor and return the result if 
// it is greater than minTimeout.
```

share/getters/shrex.go
```diff
@@ -18,23 +18,28 @@ import (
...
 // ShrexGetter is a share.Getter that uses the shrex/eds and shrex/nd protocol to retrieve shares.
 type ShrexGetter struct {
 	edsClient *shrexeds.Client
 	ndClient  *shrexnd.Client
 
-	peerManager        *peers.Manager
-	maxRequestDuration time.Duration
+	peerManager       *peers.Manager
+	minRequestTimeout time.Duration
+	minAttemptsCount  int
```
renaynay, line 33: pls comment

share/getters/shrex.go
```diff
@@ -18,23 +18,28 @@ import (
 
 var _ share.Getter = (*ShrexGetter)(nil)
 
-const defaultMaxRequestDuration = time.Second * 10
+const (
+	defaultMinRequestTimeout = time.Second * 10
```
renaynay, line 22: shouldn't this be based in min attempts count ?

walldiss: No, it is absolute minimal timeout to try sample from peer. We need to limit lowest edge, because it is reasonable time to sample large block from healthy peer.  Timeout could not be less than this value. Real timeout could be higher and affected by sampling timeout configuration (workers amount*blocktime) / minAttempts.

das/options.go
```diff
@@ -49,7 +51,7 @@ func DefaultParameters() Parameters {
 		ConcurrencyLimit:        16,
 		BackgroundStoreInterval: 10 * time.Minute,
 		SampleFrom:              1,
-		SampleTimeout:           time.Minute,
+		SampleTimeout:           15 * time.Second * 16, // block time * max amount of catchup workers
```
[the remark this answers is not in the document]
Wondertan: Didn't we want to put multiplier 2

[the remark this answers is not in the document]
walldiss: updated to reuse same var for SampleTimeout and ConcurrencyLimit. 

It is already reasonably high value without extra 2x multiplication. I don't see any issue having it without extra multiplication

vgonkivs (dismissed):

das/options.go
```diff
@@ -36,20 +36,24 @@ type Parameters struct {
...
+	concurrencyLimit := 16
 	return Parameters{
 		SamplingRange:           100,
-		ConcurrencyLimit:        16,
+		ConcurrencyLimit:        concurrencyLimit,
 		BackgroundStoreInterval: 10 * time.Minute,
 		SampleFrom:              1,
-		SampleTimeout:           time.Minute,
+		// SampleTimeout = block time * max amount of catchup workers
+		SampleTimeout: 15 * time.Second * time.Duration(concurrencyLimit),
```
Wondertan, line 56: We should pass this block time from as a param tbh. There are multiple places where we rely on it, and changing it in one place is convenient. modp2p.Blocktime

das/options.go
```diff
@@ -36,20 +36,24 @@ type Parameters struct {
 	SampleFrom uint64
 
 	// SampleTimeout is a maximum amount time sampling of single block may take until it will be
-	// canceled
+	// canceled. High ConcurrencyLimit value may increase sampling time due to node resources being
+	// divided between parallel workers. SampleTimeout should be adjusted proportional to
+	// ConcurrencyLimit.
```
Wondertan, line 41:
```suggestion
	// divided between parallel workers. SampleTimeout should be adjusted proportionally to
	// ConcurrencyLimit.
```

Wondertan (approved):
Very nice. Changed to cascade also makes sense to me.

das/options.go
Wondertan: We would also need to merge https://github.com/celestiaorg/celestia-node/pull/1773

share/getters/cascade.go
```diff
@@ -3,17 +3,15 @@ package getters
...
 	"go.opentelemetry.io/otel/trace"
 	"go.uber.org/multierr"
 
-	"github.com/celestiaorg/nmt/namespace"
-	"github.com/celestiaorg/rsmt2d"
-
 	"github.com/celestiaorg/celestia-node/libs/utils"
 	"github.com/celestiaorg/celestia-node/share"
+	"github.com/celestiaorg/nmt/namespace"
+	"github.com/celestiaorg/rsmt2d"
```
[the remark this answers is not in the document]
walldiss: ARGHHHH. This happen enough times to me so here is ultimate fix: https://github.com/celestiaorg/celestia-node/pull/1806

das/options.go
Wondertan: No, it can just expect the caller to pass it in DefaultParams

Wondertan: But they are not really default then

Wondertan: We should then just reassign those on the nodebuilder lever

walldiss: We can do it too. But we need to set default and it should not depend on node pkg. Maybe magic number is not that bad, considering we don't really care about being exact here, just using the assumption to make some pre-defined default.

nodebuilder/das/config.go
```diff
@@ -16,7 +18,9 @@ type Config das.Parameters
 //
 // TODO(@derrandz): Address #1261
 func DefaultConfig() Config {
-	return Config(das.DefaultParameters())
+	cfg := das.DefaultParameters()
+	cfg.SampleTimeout = modp2p.BlockTime * time.Duration(cfg.SampleTimeout)
```
Wondertan, line 22: This seem to make the timeout huge because it take the default and then additionally multiples it

share/availability.go
```diff
@@ -3,19 +3,13 @@ package share
...
 
 	"github.com/celestiaorg/celestia-app/pkg/da"
 )
 
 // ErrNotAvailable is returned whenever DA sampling fails.
 var ErrNotAvailable = errors.New("share: data not available")
 
-// AvailabilityTimeout specifies timeout for DA validation during which data have to be found on
-// the network, otherwise ErrNotAvailable is fired.
-// TODO: https://github.com/celestiaorg/celestia-node/issues/10
```
Wondertan, line 16: We can actually close this issue by this PR, imo

renaynay (commented):
lint err and address this comment https://github.com/celestiaorg/celestia-node/pull/1791/files#r1114931113
and good to go

Wondertan (approved):

renaynay (approved):

PR #1794, opened 2023-02-21
feat: add libheader exchange metrics

derrandz:
## Overview

This PR introduces `header_p2p_headers_response_size` and `header_p2p_headers_request_duration` metrics to celestia-node.

These metrics will be used primarily for testground benchmarks, but also are informative for testnet and potentially mainnet.

## Checklist


libs/header/p2p/metrics.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	m := &metrics{
+		responseSize:     responseSize,
+		responseDuration: responseDuration,
+	}
+
+	ex.metrics = m
+
+	return nil
```
Wondertan, line 48:
```suggestion
	ex.metrics = &metrics{
		responseSize:     responseSize,
		responseDuration: responseDuration,
	}
	return nil
```

libs/header/p2p/metrics.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	responseDuration, err := meter.
+		SyncFloat64().
+		Histogram(
+			"header_p2p_headers_request_duration",
+			instrument.WithDescription("Duration of get headers request in seconds"),
+		)
+	if err != nil {
+		return err
```
Wondertan, line 38: panic as well then, pls be consistent

libs/header/p2p/metrics.go
```diff
@@ -0,0 +1,57 @@
...
+
+var (
+	meter = global.MeterProvider().Meter("header/p2p")
+)
+
+func (ex *Exchange[H]) RegisterMetrics() error {
+	responseSize, err := meter.
+		SyncFloat64().
+		Histogram(
+			"header_p2p_headers_response_size",
```
Wondertan, line 24: no need to copy header_p2p, as it's already in the meter name

Wondertan: Same for others

nodebuilder/settings.go
```diff
@@ -15,6 +15,7 @@ import (
 
 	fraudPkg "github.com/celestiaorg/celestia-node/fraud"
 	headerPkg "github.com/celestiaorg/celestia-node/header"
+	header "github.com/celestiaorg/celestia-node/nodebuilder/header"
```
Wondertan, line 18: modhead. We aim to call all the modules with mod prefixed

libs/header/p2p/metrics.go
```diff
@@ -0,0 +1,57 @@
...
+type metrics struct {
+	responseSize     syncfloat64.Histogram
+	responseDuration syncfloat64.Histogram
+}
+
+var (
+	meter = global.MeterProvider().Meter("header/p2p")
+)
+
+func (ex *Exchange[H]) RegisterMetrics() error {
```
Wondertan, line 20: Instead, there should be a private field on Params that can be enabled by an option for Exchange.

Wondertan: WithMetrics on nodebuilder will then supply this option to the FX and to Exchange

Wondertan (changes requested):

libs/header/p2p/metrics.go
derrandz: When the metrics are displayed, the meter name is not considered.

nodebuilder/settings.go
derrandz: +1

libs/header/p2p/metrics.go
derrandz: As I mentioned in our discussion, I encountered FX issues with this approach, so we can afford to stay consistent to DASer ftm.

nodebuilder/header/opts.go
```diff
@@ -0,0 +1,20 @@
...
+	header "github.com/celestiaorg/celestia-node/header"
+	libhead "github.com/celestiaorg/celestia-node/libs/header"
+	p2p "github.com/celestiaorg/celestia-node/libs/header/p2p"
+)
+
+// WithMetrics provides sets `MetricsEnabled` to true on ClientParameters for the header exchange
+func WithMetrics(ex libhead.Exchange[*header.ExtendedHeader]) error {
+	exchange, ok := (ex).(*p2p.Exchange[*header.ExtendedHeader])
+	if !ok {
+		// not all implementations of libhead.Exchange[*header.ExtendedHeader]
```
Wondertan, line 13: It's fine to panic in this case because when WithMetrics is provided, we always expect it to be the type we want. Yeah, there may be other implementations, but if it's given here, it means something went wrong with the program, so panicking earlier is helpful.

libs/header/p2p/metrics.go
Wondertan: It consireded in traces 100%, are you sure it's not there for metrics? I really don't like redundant/duplicate info, but that’s nit. The real concern is to keep this consistent in the node which is not at this point.

Wondertan (commented):
Almost there, we just need to align on the reasoning for panic or not panic.

nodebuilder/header/opts.go
derrandz: The thing is, we received multiple implementations of `libhead.Exchange` interface, right, so we want to only impact a specific `implementer. I mean just from reading what I just wrote, I really dislike how I am doing it, but the other solution to avoid this, would be to add `RegisterMetrics() error` to the interface and have all implementers define it. That's somewhat impactful in terms of code.

This quick hack will absolve us from that. So in summary, we got multiple implementers, and we only want a specific implementer that defines `RegisterMetrics() error` thus this ugliness.

Wondertan:
@derrandz, I am ok to keep such naming, but I am also making you responsible for keeping them consistent around the codebase 👍🏻

nodebuilder/header/opts.go
Wondertan: I understand, but that does not explain why we shouldn't panic there.

vgonkivs:
Hello @derrandz. Could you please add an additional field inside p2p.Exchange `averageRequestDuration` and store the average duration there? It will be very helpful for cases when our node connects to peers that are far from us and the latency could be bigger than provided requestTimeout. In case, when `averageRequestDuration > requestTimeout,` we will use it to sync the range of headers.

nodebuilder/header/opts.go
```diff
@@ -0,0 +1,26 @@
...
+		exchange, ok := (ex).(*p2p.Exchange[*header.ExtendedHeader])
+		if !ok {
+			return fmt.Errorf("header.WithMetrics: type is to *p2p.Exchange but cast failed")
+		}
+		return exchange.RegisterMetrics()
+	case *core.Exchange:
+		return nil
+	default:
+		return fmt.Errorf("header.WithMetrics: unknown/unsupported (provided) exchange type")
+	}
```
Wondertan, line 25: I think we agreed to avoid hard WithMetrics to have context of what other Exchanges exist and instead just give em check for one he knows about and do nothing otherwise.

go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/celestiaorg/go-libp2p-messenger v0.1.0
 	github.com/celestiaorg/nmt v0.14.0
 	github.com/celestiaorg/rsmt2d v0.8.0
+	github.com/celestiaorg/utils v0.1.0
```
Wondertan, line 16: Just saw it. Doesn't seem to be used

nodebuilder/header/opts.go
derrandz: Aaah

derrandz:
@vgonkivs I think we are clear from out discussion over Slack 👍🏻

derrandz:
@Wondertan Regarding naming, yes sir :salut:

nodebuilder/header/opts.go
```diff
@@ -0,0 +1,23 @@
...
+	case *p2p.Exchange[*header.ExtendedHeader]:
+		exchange, ok := (ex).(*p2p.Exchange[*header.ExtendedHeader])
+		if !ok {
+			return fmt.Errorf("header.WithMetrics: type is to *p2p.Exchange but cast failed")
+		}
+		return exchange.RegisterMetrics()
+	default:
+		return nil
+	}
+}
```
Wondertan, line 23:
```suggestion

// WithMetrics provides sets `MetricsEnabled` to true on ClientParameters for the header exchange
func WithMetrics(ex libhead.Exchange[*header.ExtendedHeader]) error {
	switch p2pex := ex.(type) {
	case *p2p.Exchange[*header.ExtendedHeader]:
		return p2pex.RegisterMetrics()
	default:
		return nil
	}
}
```

Wondertan (changes requested):
One last simplification comment

Wondertan (approved):

nodebuilder/header/opts.go
```diff
@@ -0,0 +1,17 @@
...
+	p2p "github.com/celestiaorg/celestia-node/libs/header/p2p"
+)
+
+// WithMetrics provides sets `MetricsEnabled` to true on ClientParameters for the header exchange
+func WithMetrics(ex libhead.Exchange[*header.ExtendedHeader]) error {
+	switch p2pex := any(ex).(type) {
+	case *p2p.Exchange[*header.ExtendedHeader]:
+		return p2pex.RegisterMetrics()
+	default:
+		return nil
```
renaynay, line 15: if this is called and there's no p2p exchange - shouldn't it return err?

libs/header/p2p/metrics.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	ex.metrics = &metrics{
+		responseSize:     responseSize,
+		responseDuration: responseDuration,
+	}
+	return nil
+}
+
+func (m *metrics) ObserveRequest(ctx context.Context, size uint64, duration uint64) {
```
renaynay, line 48: does this have to be exported?

renaynay: also we should call this `observeResponse`

renaynay: and IMO it should also record the err if there is one

nodebuilder/header/opts.go
Wondertan: No, because of Core.Exchange

libs/header/p2p/metrics.go
Wondertan: Agree with Response and error tracking, don't really care about publicity if it's anyway inaccessible

renaynay: still weird pattern, lets privatise.

derrandz: Good call, will implemented

core/testing_grpc.go
```diff
@@ -109,6 +109,5 @@ func startGRPCServer(
 			log.Error("serving GRPC: ", err)
 		}
 	}()
-
```
renaynay, line 112: ?

renaynay (commented):
what about adding err observation?

derrandz:
@renaynay Added minimal tracking of err observation until we add more elaborate schemes through peerTracker observability (#1812)

Wondertan (dismissed):

Wondertan:
It took a while, but it's worth it!

renaynay (approved):

Wondertan (approved):

vgonkivs (approved):

PR #1798, opened 2023-02-22
feat(share)!: add functional options to share pkg

derrandz:
## Overview

This PR introduces functional options to the share package and deprecates usage of default values directly in code, in favor of using parameterized values.

## Breaking

This PR breaks the configuration. The on-disk configuration becomes (_note that `Share.Availability` is only available for light nodes_):
```
[Share]
  [Share.Availability]
    SampleAmount = 16
  [Share.Discovery]
    PeersLimit = 5
    AdvertiseInterval = "8h0m0s"
```
## Checklist


share/availability/options.go
```diff
@@ -0,0 +1,344 @@
...
+	// and ShareParameters.Validate when supplied with invalid values.
+	// This error will also be returned by NewDiscovery if supplied with an invalid option
+	ErrInvalidDiscOption = fmt.Errorf("discovery: invalid option")
+)
+
+// ErrInvalidOption is a utility function to dedup code for error-returning
+// when dealing with invalid parameter values.
+func ErrInvalidOption(err error, optionName string, value string) error {
+	return fmt.Errorf("%w: value %s cannot be %s", err, optionName, value)
+}
```
walldiss, line 26: Lets write what was the provided value, so user can relate to existing config and be sure it was parsed correctly. Also nit, but usually preferred direct wording how to fix the issue instead of reversed one. So instead of "value cannot be", better to use "value should be".

share/availability/options.go
```diff
@@ -0,0 +1,344 @@
...
+// Option is a function that configures a ShareParameters
+type Option[T Parameters] func(*T)
+
+// FullAvailabilityParameters is the set of parameters that must be configured for the full availability implementation
+type FullAvailabilityParameters struct {
+	// AvailabilityTimeout specifies timeout for DA validation during which data have to be found on the network,
+	// otherwise ErrNotAvailable is fired.
+	// TODO: https://github.com/celestiaorg/celestia-node/issues/10)
+	AvailabilityTimeout time.Duration
+}
```
walldiss, line 42: Lets put specific params into corresponding subpackages.

nodebuilder/share/constructors.go
```diff
@@ -29,9 +30,15 @@ func discovery(cfg Config) func(routing.ContentRouting, host.Host) *disc.Discove
...
-			cfg.AdvertiseInterval,
+			availability.WithPeersLimit[availability.DiscoveryParameters](
+				cfg.PeersLimit,
+			),
+			availability.WithDiscoveryInterval[availability.DiscoveryParameters](
+				cfg.DiscoveryInterval,
+			),
+			availability.WithAdvertiseInterval[availability.DiscoveryParameters](
+				cfg.AdvertiseInterval,
+			),
```
walldiss, line 41: Those settings are absolutely required by constructor. I would suggest making mandatory settings to be provided by config struct(that is always in there) and  options to set only optional values. I know we done "only options, no config" in some pkgs previously, but we can reconsider this approach. 

WDYT @renaynay @Wondertan @vgonkivs

share/availability.go
```diff
@@ -3,23 +3,42 @@ package share
...
 
+const (
+	// AvailabilityTimeout specifies timeout for DA validation during which data have to be found on the network,
+	// otherwise ErrNotAvailable is fired.
+	// TODO: https://github.com/celestiaorg/celestia-node/issues/10
+	DefaultAvailabilityTimeout = 20 * time.Minute
+
+	// SampleAmount specifies the minimum required amount of samples a light node must perform
+	// before declaring that a block is available
+	DefaultSampleAmount = 16
```
walldiss, line 22: This needs to be in light subpkg, since it only applies to Light availability.

share/availability.go
```diff
@@ -3,23 +3,42 @@ package share
...
+
+	// DefaultWriteBatchSize defines the size of the batched header write.
+	// Headers are written in batches not to thrash the underlying Datastore with writes.
+	// TODO(@Wondertan, @renaynay): Those values must be configurable
+	// and proper defaults should be set for specific node type. (#709)
+	DefaultWriteBatchSize = 2048
```
walldiss, line 28: same here, should go to `cache` pkg

share/availability.go
```diff
@@ -3,23 +3,42 @@ package share
...
+
+	// DefaultCacheAvailabilityPrefix defines the string key to be used with in
+	// the datastore as the cache availability data store key
+	DefaultCacheAvailabilityPrefix = "sampling_result"
```
walldiss, line 32: and this one. General rule is to keep only abstract top level and put everything implementation specific into subpkg

walldiss (changes requested):
Just a top level overview. Lets reorganise it into subpkgs and rebase on latest main.

nodebuilder/share/constructors.go
Wondertan: Sounds gut

derrandz: I suggest we do this in a separate PR, logging into an issue for now https://github.com/celestiaorg/celestia-node/issues/1905

share/availability/options.go
derrandz: Putting them in their corresponding subpackages and then importing them in `share/availability/options.go` to aggregate all parameters for configuration's sake would cause import cycles. I suggest we keep it as is.

walldiss: You can make options typed and pkg specific (not generic I know). Subpkg should be fine importing higher level pkg. Otherwise I don't see where is import cycle

derrandz:
I resolved locally but ended up facing the import cycle issue and not pushing @walldiss so they are implemented but not pushed yet. Will unresolve til I push.

share/availability/options.go
derrandz: Okay we can sync and I can show you.

share/availability/cache/options.go
```diff
@@ -0,0 +1,87 @@
...
+
+// Validate validates the values in Parameters
+func (ca *Parameters) Validate() error {
+	if ca.WriteBatchSize <= 0 {
+		return fmt.Errorf(
+			"cache availability: invalid option: value %s was %s, where it should be %s",
+			"DefaultWriteBatchSize",
+			"<= 0", // current value
+			">= 0", // what the valueshould be
+		)
```
renaynay, line 45: can this please conform to the other error messages we have ? `value must be greater than 0`

share/availability/cache/options.go
```diff
@@ -0,0 +1,87 @@
...
+			"CacheAvailabilityPrefix",
+			"is empty",  // current value
+			"non empty", // what the should be
+		)
+	}
+
+	return nil
+}
+
+// WithTimeout is a functional option that the Availability interface
```
renaynay, line 60: godoc.

share/availability/light/options.go
```diff
@@ -0,0 +1,52 @@
+package light
+
+import (
+	"fmt"
+)
+
+// SampleAmount specifies the minimum required amount of samples a light node must perform
+// before declaring that a block is available
+var (
+	DefaultSampleAmount uint = 16
```
renaynay, line 10: do we even still need to extract this to a var if it's just the default?

renaynay (changes requested):
Please lint and refer to comments re errors.

nodebuilder/share/config.go
```diff
@@ -13,41 +16,61 @@ var (
 	ErrNegativeInterval = errors.New("interval must be positive")
 )
 
+// // ShareParameters is the aggergation of all configuration parameters of the the share module
+type ShareParameters struct {
+	// FullAvailabilityParameters
```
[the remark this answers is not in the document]
derrandz: Left over from a refactor, this will no longer be needed

nodebuilder/share/config.go
```diff
@@ -13,41 +16,61 @@ var (
...
+
+	err = cfg.ShareParameters.Cache.Validate()
+	if err != nil {
+		return fmt.Errorf("nodebuilder/share: %w", err)
+	}
+
+	err = cfg.ShareParameters.Discovery.Validate()
+	if err != nil {
+		return fmt.Errorf("nodebuilder/share: %w", err)
+	}
```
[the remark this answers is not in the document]
derrandz: +1

share/availability/cache/options.go
derrandz: +1

share/availability/discovery/options.go
```diff
@@ -0,0 +1,102 @@
...
+	return Parameters{
+		PeersLimit:        3,
+		DiscoveryInterval: time.Second * 30,
+		AdvertiseInterval: time.Second * 30,
+	}
+}
+
+// Validate validates the values in Parameters
+func (p *Parameters) Validate() error {
+	if p.PeersLimit <= 0 {
```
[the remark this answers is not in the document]
derrandz: +1

share/availability/light/options.go
derrandz: Good question, I am keeping it just to avoid refactoring tests that rely on this value for the time being. Such refactor will include usage of the new `WithSampleAmount` option. I am referring to `./nodebuilder/tests/reconstruct_test.go:94`

nodebuilder/share/constructors.go
```diff
@@ -21,17 +21,17 @@ import (
...
 	) *disc.Discovery {
 		return disc.NewDiscovery(
 			h,
 			routingdisc.NewRoutingDiscovery(r),
-			cfg.PeersLimit,
-			cfg.DiscoveryInterval,
-			cfg.AdvertiseInterval,
+			disc.WithPeersLimit(cfg.Discovery.PeersLimit),
+			disc.WithDiscoveryInterval(cfg.Discovery.DiscoveryInterval),
+			disc.WithAdvertiseInterval(cfg.Discovery.AdvertiseInterval),
```
Wondertan, line 34: Don't see logic that uses CacheAvailability options anywehre

Wondertan (changes requested):

share/availability/cache/options.go
```diff
@@ -0,0 +1,85 @@
...
+// To be used with the construction
+// example:
+//
+// NewShareAvailability(
+//
+//	bServ,
+//	disc,
+//	WithWriteBatchSize(uint),
+//
+// )
```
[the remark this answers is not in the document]
Wondertan: More like an opinion, don't feel strong about removing it

share/availability/discovery/options.go
```diff
@@ -0,0 +1,99 @@
...
+// implementers use to set the DiscoveryInterval configuration param
+//
+// To be used with the construction, see example in WithPeersLimit documentation
+func WithDiscoveryInterval(discInterval time.Duration) Option {
+	return func(p *Parameters) {
+		p.DiscoveryInterval = discInterval
+	}
+}
+
+// WithAdvertiseInterval is a functional option that the Discoverable interface
```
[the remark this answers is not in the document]
Wondertan: Also, DIscovery is not an interface without implementers, so this configures the Discovery process/component

share/availability/cache/options.go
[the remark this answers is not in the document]
renaynay: Only comment if it is critical to understanding, otherwise it's just more stuff to maintain

nodebuilder/share/config.go
```diff
@@ -14,16 +16,15 @@ var (
 	ErrNegativeInterval = errors.New("interval must be positive")
 )
 
+// Parameters is the aggregation of all configuration parameters of the the share module
+type Parameters struct {
+	Light     light.Parameters
+	Cache     cache.Parameters
+	Discovery discovery.Parameters
+}
```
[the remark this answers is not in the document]
derrandz: To what are you referring specifically? Cache? To be honest, This was a robotic extraction, as I went about it in a no brainer manner to achieve the functional options without much thought as to what can be kept private. I am open for suggestions.

share/availability/cache/options.go
```diff
@@ -0,0 +1,85 @@
...
+func WithWriteBatchSize(writeBatchSize uint) Option {
+	return func(p *Parameters) {
+		p.WriteBatchSize = writeBatchSize
+	}
+}
+
+// WithCacheAvailabilityPrefix is a functional option that the Availability interface
+// implementers use to set the CacheAvailabilityPrefix configuration param
```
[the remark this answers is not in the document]
derrandz: removed this as it does not make sense to open it for configuration as per your comment above

nodebuilder/share/constructors.go
derrandz: Agreed, in this PR we open it for configuration (_specifically and only, `WriteBatchSize`_) just in case somewhere down the line we might need to play with this in tests or something.

nodebuilder/share/config.go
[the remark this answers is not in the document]
derrandz: I believe this was a NIT from rene that I implemented to reduce the amount of `validate` calls in `config.Validate`. It's syntactic sugar basically.

nodebuilder/share/constructors.go
distractedm1nd: Hmmm. not so sure about this. YAGNI

nodebuilder/share/config.go
```diff
@@ -1,42 +1,67 @@
...
 	"github.com/celestiaorg/celestia-node/share/p2p/shrexnd"
 )
 
+var (
+	ErrNegativeInterval = errors.New("interval must be positive")
+)
+
 type Config struct {
-	// UseShareExchange is a flag toggling the usage of shrex protocols for blocksync.
+	Light     light.Parameters
```
Wondertan, line 19: Let's call this Availability, as otherwise, the Light does not mean anything.

We should also make it LN only

cc @distractedm1nd, as I remember you found a hacky way to do this

share/availability/discovery/options.go
```diff
@@ -0,0 +1,70 @@
...
+	// PeersLimit defines the soft limit of FNs to connect to via discovery.
+	// Set 0 to disable.
+	PeersLimit uint
+	// AdvertiseInterval is a interval between advertising sessions.
+	// Set -1 to disable.
+	// NOTE: only full and bridge can advertise themselves.
+	AdvertiseInterval time.Duration
+	// discoveryRetryTimeout is an interval between discovery attempts
+	// when we discovered lower than PeersLimit peers.
+	// Set -1 to disable.
```
Wondertan, line 19: Relic from past PR, this line is no longer true

nodebuilder/share/config.go
```diff
@@ -1,42 +1,67 @@
...
+	"github.com/celestiaorg/celestia-node/share/availability/discovery"
+	"github.com/celestiaorg/celestia-node/share/availability/light"
 	"github.com/celestiaorg/celestia-node/share/p2p/peers"
 	"github.com/celestiaorg/celestia-node/share/p2p/shrexeds"
 	"github.com/celestiaorg/celestia-node/share/p2p/shrexnd"
 )
 
+var (
+	ErrNegativeInterval = errors.New("interval must be positive")
+)
```
Wondertan, line 16: This does not seem to be used?

Wondertan:
Also PR description needs to be updated

Wondertan (dismissed):

share/availability/cache/availability.go
```diff
@@ -42,9 +37,12 @@ type ShareAvailability struct {
...
 // NewShareAvailability wraps the given share.Availability with an additional datastore
 // for sampling result caching.
-func NewShareAvailability(avail share.Availability, ds datastore.Batching) *ShareAvailability {
+func NewShareAvailability(
+	avail share.Availability,
+	ds datastore.Batching,
+) *ShareAvailability {
 	ds = namespace.Wrap(ds, cacheAvailabilityPrefix)
-	autoDS := autobatch.NewAutoBatching(ds, DefaultWriteBatchSize)
+	autoDS := autobatch.NewAutoBatching(ds, int(writeBatchSize))
```
renaynay, line 45: unnecessary type conversion

distractedm1nd (dismissed):

nodebuilder/share/config.go
```diff
@@ -1,42 +1,68 @@
...
 	"github.com/celestiaorg/celestia-node/share/p2p/shrexeds"
 	"github.com/celestiaorg/celestia-node/share/p2p/shrexnd"
 )
 
 type Config struct {
-	// UseShareExchange is a flag toggling the usage of shrex protocols for blocksync.
+	Availability light.Parameters `toml:",omitempty"`
+	Discovery    discovery.Parameters
+
 	UseShareExchange bool
```
renaynay, line 18: nit but i'd put this at the bottom

Also for consistency - some params are pointers and others not. Let's fix this.

nodebuilder/share/config.go
```diff
@@ -1,42 +1,68 @@
...
+	"github.com/celestiaorg/celestia-node/share/availability/discovery"
+	"github.com/celestiaorg/celestia-node/share/availability/light"
 	"github.com/celestiaorg/celestia-node/share/p2p/peers"
 	"github.com/celestiaorg/celestia-node/share/p2p/shrexeds"
 	"github.com/celestiaorg/celestia-node/share/p2p/shrexnd"
 )
 
 type Config struct {
-	// UseShareExchange is a flag toggling the usage of shrex protocols for blocksync.
+	Availability light.Parameters `toml:",omitempty"`
```
renaynay, line 15: Let's call this LightAvailability

share/availability/discovery/discovery.go
```diff
@@ -94,15 +63,21 @@ type OnUpdatedPeers func(peerID peer.ID, isAdded bool)
...
+	params := DefaultParameters()
+
+	for _, opt := range opts {
+		opt(&params)
+	}
+
 	return &Discovery{
-		params:         params.withDefaults(),
-		set:            newLimitedSet(params.PeersLimit),
+		set:            newLimitedSet(uint(params.PeersLimit)),
```
renaynay, line 75: unnecessary type conversion.

share/availability/discovery/options.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+}
+
+// Validate validates the values in Parameters
+func (p *Parameters) Validate() error {
+	if p.PeersLimit < 0 {
+		return fmt.Errorf(
+			"discovery: invalid option: value PeersLimit %s, %s",
+			"is negative.",
+			"value must be 0 or positive",
```
renaynay, line 40: nit on formatting (same as other comment i made) and same for check below

share/availability/discovery/options.go
```diff
@@ -0,0 +1,69 @@
...
+	discoveryRetryTimeout time.Duration
+}
+
+// Option is a function that configures Discovery Parameters
+type Option func(*Parameters)
+
+// DefaultParameters returns the default Parameters' configuration values
+// for the Discovery module
+func DefaultParameters() Parameters {
+	return Parameters{
```
renaynay, line 28: why isn't the default for discoveryRetryTimeout included here?

This will panic the ticker.

renaynay (changes requested):
* please check to ensure units + units+race are passing, looks like there's an error associated with types

nodebuilder/share/config.go
```diff
@@ -1,42 +1,68 @@
...
 		ShrExEDSParams:    shrexeds.DefaultParameters(),
 		ShrExNDParams:     shrexnd.DefaultParameters(),
+		UseShareExchange:  true,
 		PeerManagerParams: peers.DefaultParameters(),
-		Discovery:         disc.DefaultParameters(),
 	}
+
+	if tp == node.Light {
+		cfg.Availability = light.DefaultParameters()
+	}
```
walldiss, line 39: If the value filled only for light node, would it use zero values for all other type of nodes?

nodebuilder/share/config.go
walldiss, line 15: is `Config` used only for `light` or for both? If for both would it be possible cleaner to extract common part into generic Config and for light use inheritance by embedding generic Config and adding extra field?

Wondertan: Ok

share/availability/discovery/options.go
walldiss: It will panic for sure.  discoveryRetryTimeout  should be overwritten by defaults in constructor the same way it was before. Currently we read defaults from config file, not from code and private values could not be populated this way. So the simplest way to make local params work, without major config defaults refactoring is to overwrite.

share/availability/discovery/options.go
```diff
@@ -0,0 +1,69 @@
...
+	// Set -1 to disable.
+	// NOTE: only full and bridge can advertise themselves.
+	AdvertiseInterval time.Duration
+	// discoveryRetryTimeout is an interval between discovery attempts
+	// when we discovered lower than PeersLimit peers.
+	discoveryRetryTimeout time.Duration
+}
+
+// Option is a function that configures Discovery Parameters
+type Option func(*Parameters)
```
walldiss, line 23: What we really need functional options for config values that are exported and could be set directly?

nodebuilder/share/config.go
derrandz: Yes, and will be omitted in the on-disk config

nodebuilder/share/config.go
derrandz: Adding this as a TODO

share/availability/discovery/options.go
derrandz: Unclear question

nodebuilder/share/config.go
derrandz: I think `omitempty` is sufficient for now

share/availability/discovery/options.go
Wondertan: defaultRetryTimeout

nodebuilder/share/config.go
Wondertan: +1, we don't need to overcomplicate this right now. This will be and should be a part of a larger refactoring

nodebuilder/share/config.go
```diff
@@ -1,42 +1,69 @@
...
+
+	"github.com/celestiaorg/celestia-node/nodebuilder/node"
+	"github.com/celestiaorg/celestia-node/share/availability/discovery"
+	"github.com/celestiaorg/celestia-node/share/availability/light"
 	"github.com/celestiaorg/celestia-node/share/p2p/peers"
 	"github.com/celestiaorg/celestia-node/share/p2p/shrexeds"
 	"github.com/celestiaorg/celestia-node/share/p2p/shrexnd"
 )
 
+// TODO: some params are pointers and other are not, Let's fix this.
```
Wondertan, line 14: Can be a good first issue

Wondertan (approved):

share/availability/discovery/discovery.go
```diff
@@ -90,15 +58,25 @@ type OnUpdatedPeers func(peerID peer.ID, isAdded bool)
...
 	h host.Host,
 	d discovery.Discovery,
-	params Parameters,
+	opts ...Option,
 ) *Discovery {
+	params := DefaultParameters()
+	// TODO: DefaultParameters was changed to not include `discoveryRetryTimeout`
+	// since it's a private field and it breaks nodebuilder/config_test.go
+	// so this is gonna be where we set the default value for it for now.
+	params.discoveryRetryTimeout = defaultRetryTimeout
```
Wondertan, line 67: Ah, that's annoying but makes sense

share/availability/light/options.go
```diff
@@ -0,0 +1,52 @@
...
+		)
+	}
+
+	return nil
+}
+
+// WithSampleAmount is a functional option that the Availability interface
+// implementers use to set the SampleAmount configuration param
+//
+// To be used with the construction, see example in WithAvailabilityTimeout documentation
```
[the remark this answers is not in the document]
renaynay: remove this doc if not, confusing.

nodebuilder/share/config.go
```diff
@@ -1,42 +1,68 @@
...
 	// PeerManagerParams sets peer-manager configuration parameters
 	PeerManagerParams peers.Parameters
-	// Discovery sets peer discovery configuration parameters.
-	Discovery disc.Parameters
 }
 
-func DefaultConfig() Config {
-	return Config{
-		UseShareExchange:  true,
+// TODO: Remove share/availability/options.go and reorg configs here
```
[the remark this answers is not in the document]
renaynay: okay so remove pls.

share/availability/discovery/discovery.go
Wondertan: I agree

derrandz: This made its way to discovery through the revamp PR, so dunno what's the rational honestly, but it makes sense to turn it into just a var

share/availability/discovery/discovery_test.go
```diff
@@ -42,7 +42,7 @@ func TestDiscovery(t *testing.T) {
 	discs := make([]*Discovery, nodes)
 	for i := range discs {
 		discs[i] = tn.discovery(WithPeersLimit(0), WithAdvertiseInterval(time.Millisecond*100))
-		discs[i].params.discoveryRetryTimeout = -1
+		discoveryRetryTimeout = -1 // defined in discovery.go
```
renaynay, line 45: comments unnecessary but its ok

share/availability/discovery/discovery.go
```diff
@@ -35,53 +35,23 @@ const (
...
-}
-
-func DefaultParameters() Parameters {
-	return Parameters{
-		PeersLimit: 5,
-		// based on https://github.com/libp2p/go-libp2p-kad-dht/pull/793
-		AdvertiseInterval: time.Hour * 22,
-	}
-}
+var discoveryRetryTimeout = defaultRetryTimeout
```
Wondertan, line 38: Comment left the building

Wondertan (approved):
It took us a while but we finally updated the config

vgonkivs (approved):

walldiss (approved):

Issue #1799, opened 2023-02-22
das: throttle routines spawned for recent header sampling


Wondertan:
[Here](https://github.com/celestiaorg/celestia-node/blob/main/das/coordinator.go#L106-L110) we spawn sampling workers. There might be a case where processing happens slower than the production of new routines. Thus, there should be a limiter for the number of routines(16-32 )

Issue #1779, opened 2023-02-20
ShrEx bugs


renaynay:
Running list of known issues:

- [ ] #1776 
- [ ] Lights and fulls cannot retrieve/sample historical 256 blocks
- [ ] `shard not found` error for historical block

PR #1802, opened 2023-02-22
improvement(das): use header from header.sub instead of store for recent jobs

walldiss:
- use header from sub for recent headers, instead of header store

## Overview
Part of fixing bugs https://github.com/celestiaorg/celestia-node/issues/1779
 - For recent jobs, header will be taken from header.Sub instead of header store.
 - Add timeout for header store call. Will target https://github.com/celestiaorg/celestia-node/issues/1799

## Checklist


Wondertan (dismissed):
So this all means that previously, our single routine for fetching recent headers was just stuck forever, then we started to launch a routine per header which produced a leak, which now fixes it!

LGTM, found only one log issue

das/worker.go
```diff
@@ -18,6 +18,11 @@ import (
 type worker struct {
 	lock  sync.Mutex
 	state workerState
+
+	getter    libhead.Getter[*header.ExtendedHeader]
+	sampleFn  sampleFn
+	broadcast shrexsub.BroadcastFn
+	metrics   *metrics
```
renaynay, line 25: I don't understand need for this change in this PR

das/worker.go
```diff
@@ -138,14 +101,80 @@ func (w *worker) run(
...
+}
+
+func (w *worker) getHeader(ctx context.Context, height uint64) (*header.ExtendedHeader, error) {
+	if w.state.header != nil {
+		return w.state.header, nil
+	}
+
+	// TODO: get headers in batches
+	start := time.Now()
+	h, err := w.getter.GetByHeight(ctx, height)
```
renaynay, line 158: shouldn't we still implement some reasonable timeout here as well? Or is this no longer necessary with the other timeouts PR?

das/worker.go
walldiss: The function is splitted and dependancies needs to be accessed from multiple methods. Passing the whole pack every time looks ugly

das/worker.go
Wondertan: Sometimes DASing goes much faster than syncing headers, e.g, blocks are empty. In such a case, there is nothing else DASer can do besides waiting for the header. If we do timeout, they will be hit sometimes; we would crap out with logs and start to wait again. Vlad put a timeout before and had failure logs when bootstrappers didn't serve new headers and after this, we decided to avoid timeout here. Again, because there is nothing we can do and should just wait.

walldiss: I would also like to add that if syncing is stuck for any reason, only catchup workers will be locked. With changes from this PR recent jobs will keep sampling recent headers

renaynay: ok

renaynay (approved):

Wondertan (approved):

PR #1803, opened 2023-02-22
misc(libs/header/p2p): store already fetched headers in case of timeout

vgonkivs:
## Overview
In mocha we already faced an issue when we do not have enough time to fetch all headers, mainly because of their size, and in case of any error, we discard all fetched headers and start from the beginning. This PR aims to fix this issue. So, now if we get a timeout/deadline error, we will try to store all headers that we've already fetched, rebuild the request with the new range and send it to the processLoop.
*NOTE*: in case the timeout is reached and we've fetched even 1 header, then the peer will not be punished and will be added to the queue `as ready to be used`.

## Checklist


vgonkivs:
Just realized that this improvement allows the server side to return only available headers and do not respond w `ErrNotFound`

Wondertan:
> NOTE: in case the timeout is reached and we've fetched even 1 header, then the peer will not be punished and will be added to the queue as ready to be used.

I am slightly concerned about this statement, as it means that even slow peers will get good scores, which defeats the purpose of scoring in the first place

vgonkivs:
> I am slightly concerned about this statement, as it means that even slow peers will get good scores, which defeats the purpose of scoring in the first place

It's ok. If you'll take a look at the peerScore formula, you'll notice that we are calculating it by dividing amount of bytes on the request time. After this, we are calculating new as `peerScore = (prevPeerScore + newScore) / 2; `
In case when we are punishing the peer, we are decreasing its score by 20%. I'm pretty sure that the case that you've concerned about will punish peers more than 20%

Wondertan:
I see, then no more concerns

libs/header/p2p/session.go
```diff
@@ -139,30 +145,47 @@ func (s *session[H]) doRequest(
...
-			log.Debug("retrying the request from different peer")
-		}
-		return
+		// we should not punish peer at this point and should try to parse responses, despite that error was received.
+		log.Debugw("requesting headers from peer failed", "failed peer", stat.peerID, "err", err)
 	}
 
 	h, err := s.processResponse(r)
 	if err != nil {
 		switch err {
```
renaynay, line 154: just switch on err here

libs/header/p2p/session.go
```diff
@@ -139,30 +145,47 @@ func (s *session[H]) doRequest(
...
+		"receivedAmount", len(h),
+		"requestedAmount", req.Amount,
+	)
+
+	// ensure that we received the correct amount of headers.
+	if uint64(len(h)) != req.Amount {
+		from := uint64(h[len(h)-1].Height())
+		amount := req.Amount - from
+		select {
+		case <-s.ctx.Done():
```
renaynay, line 181: return here, no?

libs/header/p2p/session.go
```diff
@@ -139,30 +145,47 @@ func (s *session[H]) doRequest(
...
 	h, err := s.processResponse(r)
 	if err != nil {
 		switch err {
 		case header.ErrNotFound:
+			// errNoResponses means that we did not fetch any headers from the peer, so we can decrease its score.
+		case errEmptyResponse:
+			stat.decreaseScore()
 		default:
 			s.peerTracker.blockPeer(stat.peerID, err)
+			log.Errorw("parsing headers failed", "failed peer", stat.peerID, "err", err)
```
renaynay, line 161:
```suggestion
			log.Errorw("processing headers response failed", "failed peer", stat.peerID, "err", err)
```

renaynay (changes requested):
small comments

Also on L212 in session.go, can you please return unmarshalling error? 

Can you also create a test for this (partial response from peer) 🙏🏻

libs/header/p2p/helpers.go
```diff
@@ -62,28 +62,37 @@ func sendMessage(
...
 		}
 
 		totalRequestSize += uint64(msgSize)
 		headers = append(headers, resp)
 	}
 
 	duration := time.Since(startTime).Milliseconds()
-	if err = stream.Close(); err != nil {
-		log.Errorw("closing stream", "err", err)
+	if err == nil {
```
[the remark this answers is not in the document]
vgonkivs: the idea is to measure the duration just after calls happened.

libs/header/p2p/session.go
vgonkivs: it does. Next line is another case

libs/header/p2p/session.go
vgonkivs: No. It will not work as expected then. In case we are receiving an error, we have to check the error and then return the request to the process loop, to re-request from another peer.

vgonkivs:
> Can you also create a test for this (partial response from peer) 
Will do in scope of https://github.com/celestiaorg/celestia-node/pull/1807, bc the current implementation is client-side. But I also need the server-side

libs/header/p2p/session.go
vgonkivs: `updateStats` + return bc we are receiving ctx.Done after external calls happened.

libs/header/p2p/helpers.go
```diff
@@ -62,28 +62,37 @@ func sendMessage(
...
+		// we do not need to close the stream as it was already reset.
```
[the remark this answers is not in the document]
vgonkivs: reworked

derrandz (changes requested):
Lil nits here and there 👍🏻

libs/header/p2p/session.go
```diff
@@ -172,6 +198,10 @@ func (s *session[H]) doRequest(
 
 // processResponse converts HeaderResponse to Header.
 func (s *session[H]) processResponse(responses []*p2p_pb.HeaderResponse) ([]H, error) {
+	if len(responses) == 0 {
+		return nil, errEmptyResponse
+	}
+
```
Wondertan, line 204: Because of this, ErrNotFound will never be the case, and peers who told us they don't have anything, will have their score decreased. 

It's actually might be a good thing to do, but then we should unify errEmptyResponse to be ErrNotFound case. Instead of having them to be separate

libs/header/p2p/session.go
```diff
@@ -139,30 +145,50 @@ func (s *session[H]) doRequest(
...
+	)
+
+	// ensure that we received the correct amount of headers.
+	if uint64(len(h)) != req.Amount {
+		from := uint64(h[len(h)-1].Height())
+		amount := req.Amount - from
+		select {
+		case <-s.ctx.Done():
+			// we should update stats anyway in this case
+			stat.updateStats(size, duration)
```
Wondertan, line 183: What if do this on defer then?

libs/header/p2p/session.go
```diff
@@ -139,30 +145,50 @@ func (s *session[H]) doRequest(
...
+			return
+		// create a new request with the remaining headers.
+		// prepareRequests will return a slice with 1 element at this point
+		case s.reqCh <- prepareRequests(from+1, amount, req.Amount)[0]:
```
Wondertan, line 187: IIRC, we discussed splitting the requeest up into more smaller pieces or did we cancel that?

Wondertan (changes requested):
We should also rename this to be fix and not misc.

libs/header/p2p/session.go
vgonkivs: We did. But then we invented this case where  the amount will be dynamically change during requests, so now it's not necessary to split it into smaller pieces

libs/header/p2p/helpers.go
```diff
@@ -72,28 +72,38 @@ func sendMessage(
...
+		msgSize, err = serde.Read(stream, resp)
 		if err != nil {
+			// this could be a valid case when server closes the connection on its side.
 			if err == io.EOF {
-				break
+				err = nil
 			}
-			stream.Reset() //nolint:errcheck
-			return nil, 0, 0, fmt.Errorf("header/p2p: failed to read a response: %w", err)
+			break
```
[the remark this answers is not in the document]
vgonkivs: My main idea was to explicitly enable the case when the server can return back everything that it has, instead of sending us `ErrNotFound`

libs/header/p2p/session.go
```diff
@@ -139,30 +145,50 @@ func (s *session[H]) doRequest(
...
+	log.Debugw("request headers from peer succeeded",
+		"peer", stat.peerID,
+		"receivedAmount", len(h),
+		"requestedAmount", req.Amount,
+	)
+
+	// ensure that we received the correct amount of headers.
+	if uint64(len(h)) != req.Amount {
+		from := uint64(h[len(h)-1].Height())
+		amount := req.Amount - from
```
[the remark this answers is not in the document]
vgonkivs: The same `amount` is used everywhere in the code and will be confusing, imo

libs/header/p2p/helpers.go
```diff
@@ -72,28 +72,38 @@ func sendMessage(
...
-	if err = stream.Close(); err != nil {
-		log.Errorw("closing stream", "err", err)
-	}
 
-	return headers, totalRequestSize, uint64(duration), nil
+	if err != nil {
+		// reset stream in case of an error
+		stream.Reset() //nolint:errcheck
+
+	} else {
```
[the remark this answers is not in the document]
vgonkivs: then it should err == io.EOF || err == nil for stream.Close() and stream.Reset for all other.

libs/header/p2p/session.go
```diff
@@ -139,30 +145,50 @@ func (s *session[H]) doRequest(
...
 	h, err := s.processResponse(r)
 	if err != nil {
 		switch err {
 		case header.ErrNotFound:
+			// errNoResponses means that we did not fetch any headers from the peer, so we can decrease its score.
+		case errEmptyResponse:
+			stat.decreaseScore()
 		default:
 			s.peerTracker.blockPeer(stat.peerID, err)
+			log.Errorw("processing headers response failed", "failed peer", stat.peerID, "err", err)
```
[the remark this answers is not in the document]
vgonkivs: Well, I did not want. But I can add

libs/header/p2p/session.go
vgonkivs: You are not right. We will receive a single message with ErrNotFound statusCode. This case helps to prevent case when the malicious server will close the connection without giving us any header

libs/header/p2p/helpers.go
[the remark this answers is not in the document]
renaynay: Hmm i don't think it matters *that* much but up to you.

libs/header/p2p/session.go
```diff
@@ -82,14 +84,18 @@ func (s *session[H]) getRangeByHeight(
...
+LOOP:
+	for {
 		select {
 		case <-s.ctx.Done():
 			return nil, errors.New("header/p2p: exchange is closed")
 		case <-ctx.Done():
 			return nil, ctx.Err()
 		case res := <-result:
 			headers = append(headers, res...)
+			if uint64(len(headers)) == amount {
```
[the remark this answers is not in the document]
renaynay: https://github.com/celestiaorg/celestia-node/issues/1853 made an issue - can resolve.

libs/header/p2p/session.go
renaynay: can you at least move the `errNoResponses` comment to the proper case statement and fix its name to match the err name?

libs/header/p2p/session.go
renaynay: Okay i could swear there was no `return` statement here when i reviewed.

renaynay (approved):

libs/header/p2p/helpers.go
```diff
@@ -72,28 +72,31 @@ func sendMessage(
...
 		headers = append(headers, resp)
 	}
 
 	duration := time.Since(startTime).Milliseconds()
-	if err = stream.Close(); err != nil {
-		log.Errorw("closing stream", "err", err)
-	}
 
-	return headers, totalRequestSize, uint64(duration), nil
+	if err == nil || err == io.EOF {
```
Wondertan, line 91: let's mention why we allow EOF

libs/header/p2p/session.go
Wondertan: Got it. We still may want to decrease the score of a peer, because it does have nothing and we don't really want to request it again.

libs/header/p2p/session.go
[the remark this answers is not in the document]
Wondertan: I really doubt we need the context for syncer. It's the same problem as for DASer and GetByHeight. There is really nothing Syncer can do besides waiting until Exchange gives the requested header, and it's the responsibility of the exchange to retry internally

libs/header/p2p/exchange.go
```diff
@@ -264,7 +265,7 @@ func (ex *Exchange[H]) request(
 	log.Debugw("requesting peer", "peer", to)
 	responses, size, duration, err := sendMessage(ctx, ex.host, to, ex.protocolID, req)
 	ex.metrics.observeResponse(ctx, size, duration, err)
-	if err != nil {
+	if err != nil && err != io.EOF {
```
Wondertan, line 268: Maybe really swap out io.EOF with nil to avoid changing such ifs all over the place?

Wondertan: and in metrics as well

libs/header/p2p/session.go
Wondertan: Don't have to happen in this PR, but can be parked on the issue

libs/header/p2p/helpers.go
Wondertan: It's an uncommon thing for request-response protocols

libs/header/p2p/session.go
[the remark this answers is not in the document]
Wondertan: I think we should log out if peer gave us nothing

libs/header/p2p/exchange.go
vgonkivs: > Maybe really swap out io.EOF with nil to avoid changing such ifs all over the place?

You mean here? Yesterday, I thought that it's not correct that sendMessage decides what error is allowed... So I decided to return any kind of error from the sendMessage and the caller will decide what and how he will handle it

Wondertan: > that sendMessage decides what error is allowed.

I haven't said anything like this, or why do you see sendMessage should not decide this?

vgonkivs:
> Got it. We still may want to decrease the score of a peer, because it does have nothing and we don't really want to request it again.

But there can be a peer that is syncing rn, so it would be unfair to decrease its score just because we get ErrNotFound. But I'm open to any suggestions and we can discuss it

Wondertan:
> But there can be a peer that is syncing rn, so it would be unfair to decrease its score just because we get ErrNotFound. 

We don't block him, but we just decrease his score to avoid wasting time on it. As we can recursively as that peer over and over again, while it is still syncing. I think the solution is to make the score to be affected by time.

libs/header/p2p/exchange.go
vgonkivs: There can be a case where EOF is not a valid error, e.g. Responsibilities of sendMessage are:
* send request;
* return response/error;

And caller should decide how to handle responses/errors, imo

vgonkivs:
> We don't block him, but we just decrease his score to avoid wasting time on it. As we can recursively as that peer over and over again, while it is still syncing. I think the solution is to make the score to be affected by time.

Ok, agreed

libs/header/p2p/exchange.go
Wondertan: Then Maybe it should not do the EOF check for logs there?

vgonkivs: for these particular cases, we can mark eof as a valid response.

vgonkivs: regarding the check, I can reset it after sendMessage, but it will look weird. So, I think the best place to keep is really inside sendMessage.

Wondertan: 👌🏻

Wondertan (approved):

libs/header/p2p/session.go
```diff
@@ -139,53 +147,85 @@ func (s *session[H]) doRequest(
...
 	if err != nil {
 		switch err {
-		case header.ErrNotFound:
+		case header.ErrNotFound, errEmptyResponse:
+			stat.decreaseScore()
 		default:
 			s.peerTracker.blockPeer(stat.peerID, err)
 		}
+
+		// exclude header.ErrNotFound from being logged as it is a `valid` error
```
renaynay, line 163: a lot of this could be avoided if we eagerly exchanged HELLO messages with peers as we connect to them that contain their Head :)

renaynay (approved):

libs/header/p2p/session.go
Wondertan: This doesn't solve potential issues when we enable pruning

PR #1806, opened 2023-02-22
lint(everything): sort imports

walldiss:
## Overview

Sort imports by using amazing consol tool.  Add a command to use it with MakeFile and as part of `fmt`. IDE users could set filewatcher rule with same tool, to forget about unsorted imports problem forever!


walldiss:
The tool allows us to do ordering you described. On the other hand `golangci-lint` only allows to do checks for 3 groups:
1. standard lib
2. local pkgs
3. 3rd party pkgs

It doesn't provide rule to split local pkgs from same org pkgs, so we woun't be able to enforce such order by CI.

walldiss:
Sorted in 4 groups now.

renaynay (approved):

Makefile
```diff
@@ -155,3 +161,5 @@ openrpc-gen:
 	@go run ./cmd/docgen fraud header state share das p2p node
 .PHONY: openrpc-gen
 
+sort-imports: $(GOIMPORTSREVISER)
+	@$(GOIMPORTSREVISER) -company-prefixes "github.com/celestiaorg"  -project-name "github.com/celestiaorg/celestia-node" -output stdout ./...
```
Wondertan, line 165:
```suggestion
sort-imports: $(GOIMPORTSREVISER)
	@$(GOIMPORTSREVISER) -company-prefixes "github.com/celestiaorg"  -project-name "github.com/celestiaorg/celestia-node" -output stdout ./...
.PHONY sort-imports
```

Makefile
```diff
@@ -5,6 +5,12 @@ ifeq (${PREFIX},)
 	PREFIX := /usr/local
 endif
 
+GOBIN ?= $(shell go env GOPATH)/bin
```
Wondertan, line 8:
```suggestion
GOBIN ?= $(shell go env GOBIN)
```

Wondertan (changes requested):

Makefile
walldiss: GOBIN could be not set. It is the case for me, for example.

walldiss: The GOBIN environment variable refers to the bin directory inside the go installation directory such as $GOROOT/bin. This variable refers to the location where you can find the binaries. Setting up this environment variable is particularly useful if we frequently switch between different GOPATH directories.

Wondertan: > The directory where go install and go get will place binaries after building main packages. Generally this is set to somewhere on the system PATH so that installed binaries can be run and discovered easily.

Wondertan: > The GOBIN environment variable refers to the bin directory inside the go installation directory such as $GOROOT/bin.

That is not true. GOBIN is where go install binaries, not where go installation is

Makefile
```diff
@@ -5,6 +5,12 @@ ifeq (${PREFIX},)
...
+
+GOIMPORTSREVISER = $(GOBIN)/goimports-reviser
+$(GOIMPORTSREVISER):
+	go install -v github.com/incu6us/goimports-reviser/v3@latest
```
Wondertan, line 12: As discussed, let's remove auto installation for now and make an issue to cover all util bins we depend on at once

Makefile
Wondertan: Pls apply

Wondertan (changes requested):

walldiss:
I've removed auto-install part and created an issue to make it available for all tools we use for fmt / lint.
https://github.com/celestiaorg/celestia-node/issues/1854

renaynay (dismissed):

vgonkivs (approved):

Wondertan (approved):

PR #1807, opened 2023-02-22
misc(libs/header/p2p): allow server to return all stored headers that correspond to the requested range

vgonkivs:
## Overview

Based on https://github.com/celestiaorg/celestia-node/pull/1803

Will be rebased.

## Checklist


libs/header/p2p/server.go
```diff
@@ -201,10 +201,33 @@ func (serv *ExchangeServer[H]) handleRequest(from, to uint64) ([]H, error) {
...
+
+		log.Debugw("server: change requested range",
+			"prevMaxHeight", to,
+			"newMaxHeight", uint64(head.Height())+1,
+		)
+		// change `to` height to return a partial range
+		to = uint64(head.Height()) + 1
 	}
 
 	headersByRange, err := serv.store.GetRangeByHeight(ctx, from, to)
```
Wondertan, line 233: Why not make GetRangeByHeight return what it has from the given range?

Wondertan: This way, we can just do `if store.Has(from) { GetRangeByHeight } else NotFound`

vgonkivs: It's not the store's responsibility to change the bounds. It should only handle the request.

Wondertan: By covering another case, we could make the store API of getRangeByHeaght slightly more universal. In the end, we would else have less code with the same functionality. Don't think we have to do it in this way right now and keep the way you implemented it. Need to park this in an issue though

libs/header/p2p/server.go
```diff
@@ -201,10 +201,33 @@ func (serv *ExchangeServer[H]) handleRequest(from, to uint64) ([]H, error) {
...
+			log.Debugw("server: requested headers not stored",
+				"from", from,
+				"to", to,
+				"currentHead",
+				head.Height(),
+			)
+			return nil, header.ErrNotFound
+		}
+
+		log.Debugw("server: change requested range",
```
Wondertan, line 225:
```suggestion
		log.Debugw("server: serving partial range",
```

libs/header/p2p/exchange_test.go
```diff
@@ -382,6 +382,38 @@ func TestExchange_RequestHeadersFromAnotherPeerWhenTimeout(t *testing.T) {
...
+	exchg.peerTracker.trackedPeers[hosts[2].ID()] = &peerStat{peerID: hosts[2].ID(), peerScore: 50}
+	exchg.peerTracker.peerLk.Unlock()
+	h, err := exchg.GetRangeByHeight(context.Background(), 1, 8)
+	require.NotNil(t, h)
+	require.NoError(t, err)
+
+	exchg.peerTracker.peerLk.Lock()
+	prevScoreAfter1 := exchg.peerTracker.trackedPeers[hosts[1].ID()].peerScore
+	prevScoreAfter2 := exchg.peerTracker.trackedPeers[hosts[2].ID()].peerScore
+	exchg.peerTracker.peerLk.Unlock()
```
Wondertan, line 411: The fact that we need to modify peertracker tin each test hints us that something is wrong with the API. Not a concern of this PR, but it is good to keep that in mind and try to avoid that

libs/header/p2p/exchange_test.go
```diff
@@ -382,6 +382,38 @@ func TestExchange_RequestHeadersFromAnotherPeerWhenTimeout(t *testing.T) {
...
+	)
+	require.NoError(t, err)
+	require.NoError(t, serverSideEx.Start(context.TODO()))
+	exchg.peerTracker.peerLk.Lock()
+	prevScoreBefore1 := exchg.peerTracker.trackedPeers[hosts[1].ID()].peerScore
+	prevScoreBefore2 := 50
+	// reducing peerScore of the second server, so our exchange will request host[1] first.
+	exchg.peerTracker.trackedPeers[hosts[2].ID()] = &peerStat{peerID: hosts[2].ID(), peerScore: 50}
+	exchg.peerTracker.peerLk.Unlock()
+	h, err := exchg.GetRangeByHeight(context.Background(), 1, 8)
```
Wondertan, line 404: different ctxs also makes this look inconsistent

Wondertan (commented):
Looks good

libs/header/p2p/exchange_test.go
vgonkivs: Why? peerScore is an internal logic of the peerTracker and has to be hidden(encapsulated) without any chance to get the write access to it. All peerScore updates should be done in a single way(updateStats). From the other side, we are trying to cover as edges many cases as we can using UT, so for these purposes, I do not see anything bad in manual modifying.

libs/header/p2p/server.go
vgonkivs: > By covering another case, we could make the store API of getRangeByHeaght slightly more universal.

Still can't agree with that. It has to have a set of methods that cover basic needs. The store should not make any decisions, imo. We can have a proxy layer - to construct requests in a proper way(universal) - yes

libs/header/p2p/exchange_test.go
vgonkivs: oops, yes, will change. Thanks

libs/header/p2p/server.go
Wondertan: Giving a sub-range of requested range is not a decision and can be part of the interface contract and docs

Wondertan:
This PR is not misc

libs/header/p2p/server.go
vgonkivs: No, it's actually a decision, that store is making on its own. I did not request a sub-range, but the full range.

libs/header/p2p/exchange_test.go
Wondertan: It's still ugly though. One ctx defiend at the top of the test, instead of multiple Background calls

libs/header/p2p/exchange_test.go
vgonkivs: No, it's fine. Just did not see anything bad in this, but only in cases when logic is hidden and only for unit tests. Also, I'm not sure that we can unify these API to be able to cover all our needs in UT. There still could be tricky cases...

libs/header/p2p/session.go
```diff
@@ -181,14 +181,13 @@ func (s *session[H]) doRequest(
...
-		stat.updateStats(size, duration)
-	}()
+	// update peer stats
+	stat.updateStats(size, duration)
 
 	// ensure that we received the correct amount of headers.
 	if uint64(len(h)) < req.Amount {
 		from := uint64(h[len(h)-1].Height())
-		amount := req.Amount - from
+		amount := req.Amount - uint64(len(h))
```
renaynay, line 190: nit would put `uint64(len(h))` in a var

renaynay (approved):

libs/header/p2p/server.go
Wondertan: Again, the fact that GetRange returns a sub-range can be documented on the interface.

libs/header/p2p/session.go
Wondertan: +1

libs/header/p2p/server.go
Wondertan: > brings complexity

Opposite, it reduces complexity for users. Checking `if err != nil && len(h) != 0` is not complex if appropriately documented. It's also not required to do such a check; if the user does not care about partial ranges, it will just ignore returned headers on ErrNotFound. The power users will check for that when needed, like Exchange. Alternatively, there could be ErrPartialRange that contains partial range, and power users may check for it if needed. 

The complexity comes, exactly, in the current implementation, because of the need to call Has and Head, or "городить огород"

vgonkivs: > "городить огород"
The same can say about adding additional logic inside the store. That's why I suggested proxy

vgonkivs: Let's finish a dispute on this point. We have two opposite thoughts and now we start going nowhere

Wondertan: I will finish on saying that this огород should live somewhere anyway and it's better be an implementation detail

Wondertan (approved):

libs/header/p2p/server.go
vgonkivs: Actually, we should keep our arch clean without огород and костыль

Wondertan: Огород in this case is a feature 😂

vgonkivs: I thought you'd meant "mess" lol

PR #1810, opened 2023-02-23
fix: fix protocolID representation

vgonkivs:
self-explanatory


Wondertan (approved):
It's still a bit of a mess between where we set slashes and where we not, but not critical

walldiss (approved):

PR #1811, opened 2023-02-23
fix(modp2p): pass only resolved addresses to resource manager

Wondertan:


vgonkivs (approved):

renaynay (approved):

derrandz (approved):
should we warn when we end up with an empty `allowlist` i.e: nothing got resolved at all! gonna be a weird case!

Wondertan:
> should we warn when we end up with an empty allowlist i.e: nothing got resolved at all! gonna be a weird case!

Then the node wouldn't start, as there will no one to connect to

PR #1813, opened 2023-02-23
feat: Add `make` targets to run celestia docker nodes

derrandz:
## Overview

This PR adds make targets to use docker (_as a process manager_) to never the nodes you run. Example of running nodes in containers:

Screen Shot 2023-02-23 at 17 41 43

Use the following command to run your node from your codebase:
```
$ make docker-build
```

```
$ make docker-run NODE=light NETWORK=arabica [...]
```

Check `README.md` for more documentation (_documentation includes how to setup local telemetry infrastructure and hook it to your running nodes_)

## Checklist


docker/entrypoint.sh
```diff
@@ -17,3 +17,4 @@ echo "$@"
 echo ""
 
 exec "$@"
+fi
```
renaynay, line 20: ?

renaynay (commented):
I'm not sure we should have this as part of the repo - this means we have to maintain it and it produces bloat. Not many people will run nodes in docker long term - it's mainly for testing IMO.

docker/entrypoint.sh
derrandz: mistake

PR #1814, opened 2023-02-23
perf(share): don't broadcast datahash of empty EDS

### LINKED ISSUE #1775: bug(share/p2p/shrexsub | das): don't broadcast empty blocks' data hashes
When dasing over empty blocks, the `BroadcastFn` is still called, broadcasting empty block hashes over shrexsub

Wondertan:
Closes #1775

### DIFF
--- a/share/empty.go
+++ b/share/empty.go
@@ -19,6 +19,7 @@ var (
 )
 
 func init() {
+	// compute empty block EDS and DAH for it
 	shares := emptyDataSquare()
 	squareSize := uint64(math.Sqrt(float64(appconsts.DefaultMinSquareSize)))
 	eds, err := da.ExtendShares(squareSize, shares)
@@ -29,6 +30,9 @@ func init() {
 	emptyEDS = eds
 	dah := da.NewDataAvailabilityHeader(eds)
 	emptyRoot = &dah
+
+	// precompute Hash, so it's cached internally to avoid potential races
+	emptyRoot.Hash()
 }
 
 // EmptyRoot returns Root of an empty EDS.
--- a/share/empty.go
+++ b/share/empty.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"math"
+	"sync/atomic"
 
 	"github.com/ipfs/go-blockservice"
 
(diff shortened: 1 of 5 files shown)


vgonkivs (approved):

share/empty.go
```diff
@@ -13,6 +14,29 @@ import (
 	"github.com/celestiaorg/rsmt2d"
 )
 
+var (
+	emptyRoot atomic.Pointer[Root]
+	emptyEDS  atomic.Pointer[rsmt2d.ExtendedDataSquare]
```
walldiss, line 19: Why do you need atomic if you only write during init?

share/p2p/shrexsub/pubsub.go
```diff
@@ -100,5 +101,10 @@ func (s *PubSub) Subscribe() (*Subscription, error) {
 
 // Broadcast sends the EDS notification (DataHash) to every connected peer.
 func (s *PubSub) Broadcast(ctx context.Context, data share.DataHash) error {
+	if bytes.Equal(share.EmptyRoot().Hash(), data) {
+		// no need to broadcast datahash of an empty block
+		return nil
+	}
+
```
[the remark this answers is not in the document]
Wondertan: Good idea

share/empty.go
Wondertan: True, originally, I did that lazily with atomict, but now it makes sense to avoid atomic at all

walldiss (dismissed):

renaynay (approved):

vgonkivs (approved):

share/empty.go
```diff
@@ -14,26 +14,27 @@ import (
 )
 
 var (
-	emptyRoot *Root
-	emptyEDS  *rsmt2d.ExtendedDataSquare
+	// EmptyRoot returns Root of an empty EDS.
+	EmptyRoot *Root
+	// EmptyEDS returns the EDS of the empty block data square.
+	EmptyEDS *rsmt2d.ExtendedDataSquare
```
walldiss, line 20: Previous version with funcs as getters didn't allow to modify values. Now it is possible to write to EmptyRoot, EmptyEDS vars

Wondertan:
Ok, so I simplified some things and fixed potential panic

share/empty.go
Wondertan: Well, yeah

Wondertan: Better keep it encapsulated

walldiss (dismissed):

renaynay (approved):

walldiss (approved):

PR #1815, opened 2023-02-23
feat(das): Don't sample recent jobs twice

walldiss:
## Overview

 - Prevent DASer from doing same job twice in case of recent sampling is run on synced node.
 - Log recent jobs with info level

### DIFF
--- a/das/worker.go
+++ b/das/worker.go
@@ -81,7 +81,7 @@ func (w *worker) run(ctx context.Context, timeout time.Duration, resultCh chan<-
 	}
 
 	log.Infow(
-		"finished sampling job",
+		"finished sampling headers",
 		"from", w.state.From,
 		"to", w.state.Curr,
 		"errors", len(w.state.failed),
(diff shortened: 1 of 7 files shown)


renaynay (approved):

das/worker.go
```diff
@@ -80,16 +80,13 @@ func (w *worker) run(ctx context.Context, timeout time.Duration, resultCh chan<-
...
-		log.Infow(
-			"sampled headers",
-			"from", w.state.From,
-			"to", w.state.Curr,
-			"finished (s)",
-			jobTime.Seconds(),
-		)
-	}
+	log.Infow(
+		"finished sampling job",
```
Wondertan, line 84: We should keep the word headers there.

walldiss: ok

Wondertan (approved):

renaynay (dismissed):

renaynay (approved):

Wondertan (approved):

PR #1816, opened 2023-02-24
refactor(share/p2p/shrexeds): Increase ODS write deadline to block time and add hash to log

renaynay:
Increases ODS write deadline to block time and adds the hash to failure log

### DIFF
--- a/share/p2p/shrexeds/options.go
+++ b/share/p2p/shrexeds/options.go
@@ -41,7 +41,7 @@ type Parameters struct {
 func DefaultParameters() *Parameters {
 	return &Parameters{
 		ReadDeadline:     time.Minute,
-		WriteDeadline:    time.Second * 5,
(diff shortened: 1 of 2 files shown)


vgonkivs (approved):

share/p2p/shrexeds/options.go
```diff
@@ -41,7 +41,7 @@ type Parameters struct {
 func DefaultParameters() *Parameters {
 	return &Parameters{
 		ReadDeadline:     time.Minute,
-		WriteDeadline:    time.Second * 5,
+		WriteDeadline:    time.Second * 15, // based on block time
```
Wondertan, line 44: We observed even longer times at this point. Let's make it 30

Wondertan (approved):

walldiss (approved):

PR #1818, opened 2023-02-24
bugfix(libs/header/exchange): do not request an empty range

vgonkivs:
## Overview
This case is possible after the initial sync, as during sync process we continue receiving headers and store them in the pending cache.

## Checklist


Wondertan (approved):
Not sure if this case possible, but it's anyway a nice cure little sanity check

libs/header/sync/sync.go
```diff
@@ -276,25 +276,25 @@ func (s *Syncer[H]) findHeaders(ctx context.Context, from, to uint64) ([]H, erro
...
 	}
-
 	out := make([]H, 0, amount)
-	for from < to {
+	for from <= to {
 		// if we have some range cached - use it
 		r, ok := s.pending.FirstRangeWithin(from, to)
 		if !ok {
-			hs, err := s.exchange.GetRangeByHeight(ctx, from, amount)
+			hs, err := s.exchange.GetRangeByHeight(ctx, from, to-from+1)
```
Wondertan, line 284: Ok, so using the `amount` is a real bug here. Because it does not include changes to `from` from previous iterations

libs/header/sync/sync.go
```diff
@@ -276,25 +276,25 @@ func (s *Syncer[H]) findHeaders(ctx context.Context, from, to uint64) ([]H, erro
 	if amount > s.Params.MaxRequestSize {
 		to, amount = from+s.Params.MaxRequestSize, s.Params.MaxRequestSize
 	}
-
```
Wondertan, line 279: Pls, don't. This empty line separates var definition from the actual logic and enhances readability

libs/header/sync/sync.go
```diff
@@ -276,25 +276,25 @@ func (s *Syncer[H]) findHeaders(ctx context.Context, from, to uint64) ([]H, erro
...
-			hs, err := s.exchange.GetRangeByHeight(ctx, from, amount)
+			hs, err := s.exchange.GetRangeByHeight(ctx, from, to-from+1)
 			return append(out, hs...), err
 		}
 
-		// first, request everything between from and start of the found range
-		hs, err := s.exchange.GetRangeByHeight(ctx, from, r.start-from)
-		if err != nil {
-			return nil, err
+		if r.start > from {
```
Wondertan, line 288: Now that I have come back to this code again. I recalled that I made this specifically to have a case with 0, as I expected it to be handled by the implementation. Basically hide the if inside GetRangeByHeight.

I think all implementations must have a sanity check for when zero GetRangeByHeight is as you initially did, independently of what we do in Syncer. The sanity check should either return an error when the 0 range is requested or return an empty slice as a successful result for requested range 0.

If we go with the successful results for request range 0, there is no big reason to have the `if r.start > from` as anyway, we will get zero len slice which will be noop. But if we go with returning the error when 0 range is requested, we would have to keep the if. 

Pls choice the option you are most comfortable with regarding the if, but we still need to have sanity checks in Getter impls

renaynay (commented):
I propose to add trace-level logs for `findHeaders`

libs/header/sync/sync.go
vgonkivs: I'm ok with returning an empty slice(actually I've done this already, but then decided to rework it).
So, do you want me to add sanity checks in all getters in scope of this PR? Or should I do it only for p2p.Exchange(all others in subsequent pr)?

Wondertan: I am fine with doing such a check in this PR

libs/header/sync/sync.go
Wondertan: Pls

Wondertan (approved):

renaynay (approved):
Test doesn't really reproduce the case that a range of 0 would be requested but it's a fair check anyway.

PR #1822, opened 2023-02-24
misc(core): fix tendermint config folder delete

SweeXordious:
## Overview

This attempts to fix the CI flakiness due to the tendermint config folder delete.

This fix will be merged in the app with this PR https://github.com/celestiaorg/celestia-app/pull/1426 and then this change can be reverted.

## Checklist

### DIFF
--- a/core/testing.go
+++ b/core/testing.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"net"
 	"net/url"
+	"os"
+	"path"
 	"testing"
 
 	appconfig "github.com/cosmos/cosmos-sdk/server/config"
@@ -86,6 +88,7 @@ func StartTestNodeWithConfig(t *testing.T, cfg *TestConfig) testnode.Context {
 	t.Cleanup(func() {
 		err := cleanupCoreNode()
 		require.NoError(t, err)
+		require.NoError(t, removeDir(path.Join([]string{cfg.Tendermint.RootDir, "config"}...)))
 	})
 
 	cctx, cleanupGRPCServer, err := StartGRPCServer(app, cfg.App, cctx)
@@ -137,4 +140,28 @@ func getEndpoint(cfg *tmconfig.Config) (string, string, error) {
 		return "", "", err
 	}
 	return host, url.Port(), nil
+}
+
+// removeDir removes the directory `rootDir`.
+// The main use of this is to reduce the flakiness of the CI when it's unable to delete
+// the config folder of the tendermint node.
+// This will manually go over the files contained inside the provided `rootDir`
+// and delete them one by one.
+func removeDir(rootDir string) error {
+	dir, err := os.ReadDir(rootDir)
+	if err != nil {
+		return err
+	}
+	for _, d := range dir {
+		p := path.Join([]string{rootDir, d.Name()}...)
+		err := os.RemoveAll(p)
+		if err != nil {
+			return err
+		}
+	}
+	err = os.RemoveAll(rootDir)
+	if err != nil {
+		return err
+	}
+	return nil
 }
--- a/core/testing.go
+++ b/core/testing.go
@@ -159,9 +159,5 @@ func removeDir(rootDir string) error {
 			return err
 		}
 	}
-	err = os.RemoveAll(rootDir)
-	if err != nil {
-		return err
-	}
-	return nil
+	return os.RemoveAll(rootDir)
 }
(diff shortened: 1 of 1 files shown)


core/testing.go
```diff
@@ -138,3 +141,27 @@ func getEndpoint(cfg *tmconfig.Config) (string, string, error) {
...
+		err := os.RemoveAll(p)
+		if err != nil {
+			return err
+		}
+	}
+	err = os.RemoveAll(rootDir)
+	if err != nil {
+		return err
+	}
+	return nil
```
renaynay, line 166:
```suggestion
	return os.RemoveAll(rootDir)
```

renaynay (approved):

renaynay:
Thank you @SweeXordious 🙏🏻

core/testing.go
```diff
@@ -86,6 +88,7 @@ func StartTestNodeWithConfig(t *testing.T, cfg *TestConfig) testnode.Context {
 	t.Cleanup(func() {
 		err := cleanupCoreNode()
 		require.NoError(t, err)
+		require.NoError(t, removeDir(path.Join([]string{cfg.Tendermint.RootDir, "config"}...)))
```
Wondertan, line 91: I think this should be part of the cleanupCoreNode func returned from `StartNode`, as it's clearly buggy. Think this way:
* You provide and API to setup something
* Additionally you provide API to clean up this something, as this something writes to disk
* Users calls this cleanup API, but it turns out to be not enough and users had to manually cleanup additional things.
* Do we want every user to stumble upon this themselves or do we want to do proper cleanup in one place without making users even noticing such issues?

Wondertan (changes requested):
Thanks for this, but this should live in app not in the node

core/testing.go
SweeXordious: It is on app side:
https://github.com/celestiaorg/celestia-app/blob/b56c1bb292d8f4e16c4e55d47032551d23e2cf8e/testutil/testnode/full_node.go#L218

Except that it's for the default network, which, I guess, we expect the users to use if they don't wanna do any extra step.
It's a good idea to add it to `StartNode`. However, I added it here since you seem to depend on releases and not sure when next Celestia-app release will, and your CI will stay flaky until you update.

I will open a PR for this:
> I think this should be part of the cleanupCoreNode func returned from StartNode

And if you prefer to wait for this to come from next release, we can close this one :D

SweeXordious: PR on app side: https://github.com/celestiaorg/celestia-app/pull/1426

Wondertan: Ok, it would have been nice to mention in the PR desc that this fix was merged into the app. The idea to have it temporarily here makes sense to me. Pls add the TODO here to remove once we update the app and we can merge this one

core/testing.go
```diff
@@ -138,3 +141,23 @@ func getEndpoint(cfg *tmconfig.Config) (string, string, error) {
 	}
 	return host, url.Port(), nil
 }
+
+// removeDir removes the directory `rootDir`.
+// The main use of this is to reduce the flakiness of the CI when it's unable to delete
+// the config folder of the tendermint node.
+// This will manually go over the files contained inside the provided `rootDir`
+// and delete them one by one.
```
SweeXordious, line 149:
```suggestion
// and delete them one by one.
// TODO: remove after updating the app to a version that contains this PR https://github.com/celestiaorg/celestia-app/pull/1426
```

core/testing.go
SweeXordious: [4faafcf](https://github.com/celestiaorg/celestia-node/pull/1822/commits/4faafcf19fcf837bc6945b2df1bb815f34b3b460)

Wondertan (approved):
Thanks @SweeXordious !

Wondertan:
Also, do we have any confirmationyet  that this actually fixes the problem? @SweeXordious

SweeXordious:
I tried it on celestia-app and orchestrator-relayer and I don't see the flakiness that often.

However, if the flakiness occurs, with this PR, we will know exactly which file is causing it and we will be able to further debug it. Because, currently, all we know is that the `config` folder is what's causing it. And this folder contains many files and is used by my routines. So we can't know for sure which resource is leaking or not closing properly.

I was hoping to see the flakiness on those repos to further debug. But it just disappeared :D which is a good thing and made me propose this change to you so it reduces, if not fix, the flakiness.

SweeXordious:
This change will be included into v0.12 release: https://github.com/celestiaorg/celestia-app/pull/1429

Wondertan:
Maybe we can just wait for it then, instead of emerging, as it will be very soon(in a few days)

renaynay:
Closing in favour of #1843 which contains this change

PR #1824, opened 2023-02-24
ADR 013: Tail Padding Optimizations

nashqueue:
## Overview
This PR is to add some tail padding-related optimizations. 

[rendered](https://github.com/celestiaorg/celestia-node/blob/f3bee3cbb602041ec41c052a03a0aaf110bd80a6/docs/adr/adr-013-tail-padding-optimizations.md)

## Checklist 

- [ ] Add reference to tail-padding specs
- [ ] Add reference to https://github.com/celestiaorg/celestia-app/issues/577
- [ ] Add Graphic of @rootulp explaining that erasure coded tailpadding is tailpadding as well 

IMPORTANT:
PRs should be left in Draft until the below checklist is completed.
-->

### DIFF
--- a/docs/adr/adr-013-tail-padding-optimizations.md
+++ b/docs/adr/adr-013-tail-padding-optimizations.md
@@ -1,0 +1,55 @@
+# ADR 13: Tail Padding Optimizations
+
+## Changelog
+
+- 24.02.2023: First Draft
+
+## Author
+
+@nashqueue
+
+## Context
+
+If the Celestia block is not full it will create tail padding. Tail padded shares look always the same. Tail padded shares create tail padded rows. tail padded rows erasure to erasure-coded tail padded rows. Tail padded rows combined with erasure-coded tail padded rows result in a full tail padded row. A full tail padded row of the same length will always have the same Namespace Merkle Trie on top of it. This will result in a tail padded row root.
+All of those things are deterministic and can be computed beforehand.
+
+Given a tail padded row root inside the DA-Header you can deduce that the underlying row is a full tail padded row. This information can be beneficial in many ways.
+
+### DAS
+
+If a Light Node receives a tail padded row root in the DAH it does not have to DAS this root. It knows the underlying shares and could generate the shares and path itself.
+
+Hypothesis: The more tail padded rows exist the fewer samples a Light Node has to do for the same DA guarantee.
+
+### Block reconstruction
+
+Given a tail padded rowroot you can use the predetermined shares to fill up the store. You will not need to request those shares from light nodes. This will save bandwidth and reconstruction time.
+
+### Storage
+
+You don`t need to save copies of the same shares multiple times. Given that 3/4 of the ODS can be tail padding it can save up to 3/16 + 3/16 = 6/16 = 3/8 of storage of the EDS.
+
+### Bandwidth
+
+You don`t need to share tail-padded rows over the wire. Given the root, the recipient can deduce the whole row.
+
+## Decision for each Optimization
+
+- DAS:
+- Block reconstruction:
+- Storage:
+- Bandwidth:
+
+## Status
+
+Proposed
+
+## Consequences
+
+### Positive
+
+Less Bandwith, Storage, and  Computation required.
+
+### Negative
+
+### Neutral


Wondertan:
@nashqueue, Thanks for writing this. 

Noting here that we have most of the optimizations implemented. We use IPLD hash addressable primitives for DASing, Reconstruction, and Storage, which provide caching by default. If you have a share locally, you won't fetch it from the network again. So, currently, all nodes will download the tail padding share once and will avoid doing further. What we can do is to generate and store in cache this tail padding share, as we do for [the empty block](https://github.com/celestiaorg/celestia-node/blob/main/share/empty.go#L38-L71).

While writing this, I realized that an empty block already contains tail padding share, so it means nodes never request it from the network nor store it, besides the BEFP case, as it does not use IPLD.

docs/adr/adr-013-tail-padding-optimizations.md
```diff
@@ -0,0 +1,55 @@
...
+
+Hypothesis: The more tail padded rows exist the fewer samples a Light Node has to do for the same DA guarantee.
+
+### Block reconstruction
+
+Given a tail padded rowroot you can use the predetermined shares to fill up the store. You will not need to request those shares from light nodes. This will save bandwidth and reconstruction time.
+
+### Storage
+
+You don`t need to save copies of the same shares multiple times. Given that 3/4 of the ODS can be tail padding it can save up to 3/16 + 3/16 = 6/16 = 3/8 of storage of the EDS.
```
rootulp, line 30:
```suggestion
You don't need to save copies of the same shares multiple times. Given that 3/4 of the ODS can be tail padding it can save up to 3/16 + 3/16 = 6/16 = 3/8 of storage of the EDS.
```

docs/adr/adr-013-tail-padding-optimizations.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Bandwidth
+
+You don`t need to share tail-padded rows over the wire. Given the root, the recipient can deduce the whole row.
```
rootulp, line 34:
```suggestion
You don't need to share tail-padded rows over the wire. Given the root, the recipient can deduce the whole row.
```

docs/adr/adr-013-tail-padding-optimizations.md
```diff
@@ -0,0 +1,55 @@
...
+- 24.02.2023: First Draft
+
+## Author
+
+@nashqueue
+
+## Context
+
+If the Celestia block is not full it will create tail padding. Tail padded shares look always the same. Tail padded shares create tail padded rows. tail padded rows erasure to erasure-coded tail padded rows. Tail padded rows combined with erasure-coded tail padded rows result in a full tail padded row. A full tail padded row of the same length will always have the same Namespace Merkle Trie on top of it. This will result in a tail padded row root.
+All of those things are deterministic and can be computed beforehand.
```
rootulp, line 14: [no change needed][informational] FYI to celestia-node implementooors, the tail padding share schema has changed slightly since initial implementation but the [specs](https://github.com/celestiaorg/celestia-app/blob/specs-staging/specs/src/specs/data_structures.md#tail-padding-share) match the current implementation. Additionally celestia-app exposes [`TailPaddingShares`](https://pkg.go.dev/github.com/celestiaorg/celestia-app@v0.11.1/pkg/shares#TailPaddingShares) which can be used to generate a tail padding share. 

celestia-app can consider exporting a constant for the row root for an extended data square tail padding row if celestia-node doesn't want to compute this itself

docs/adr/adr-013-tail-padding-optimizations.md
rootulp: To clarify: this is `3/16` instead of `12/16` of the EDS because the shares in quadrants 1, 2, 3 won't have any share that is exactly a tail padding share (due to the process of erasure coding) - right?

rootulp (commented):
LGTM

docs/adr/adr-013-tail-padding-optimizations.md
nashqueue: So in quadrant one you have the ODS . In ODS you can have max 3/4 tail padding which ich 3/16 of all quadrants. Erasurecoding the same thing gets you the same. So you will have 3/4 of Quadrant 2 (the one right to Q1) be the same. That's how you get the second 3/16

docs/adr/adr-013-tail-padding-optimizations.md
nashqueue: I should add this to reference.

docs/adr/adr-013-tail-padding-optimizations.md
rootulp: > Erasurecoding the same thing gets you the same.

Wow that's a very interesting property that I did not expect. Thanks for sharing. I verified by debugging a process proposal test. For a square size of 4 where the bottom two rows of the square are tail padding, in the extended data square the bottom two rows of Q2 are also tail padding.

## Screenshot
Screenshot 2023-02-24 at 7 46 46 PM

nashqueue: Wait, maybe I misunderstand my explanation. Can you explain the screenshot? What do you mean by also tail padding?

rootulp: **Legend**
B = PFB tx compact share
R = Reserved namespace padding share
1 = Blob 1 share
2 = Blob 2 share
3 = Blob 3 share
T = Tail padding share
P = Parity share


In the test we have an original data square of:
```ascii
BBR1
2233
TTTT
TTTT
```

and the extended data square's Q0 and Q1 looks like

```ascii
BBR1 | PPPP
2233 | PPPP
TTTT | TTTT
TTTT | TTTT
```

TBH I'm surprised by the observation that the Q1 shares in the top two rows look like parity shares (i.e. they aren't identical to the corresponding share from Q0) but the shares in the bottom two rows of Q1 look identical to tail padding shares.

nashqueue: wow nice observation. That is actually really interesting. I was expecting the erasure-coded tailpadding different from the tailpadding. Every share the same but different from T. But this looks so intuitive rn :D Good catch.

renaynay:
Done, we don't need ADR

PR #1830, opened 2023-02-27
refactoring(libs/header): cleanup config params

### LINKED ISSUE #1667: libheader: Refine configuration parameters names and descriptions
E.g. we have `MinResponses` and a comment
> // the target minimum amount of responses with the same chain head

Personally, even with being in context, I don't understand what's the purpose of this config field

Also, we have `DefaultScore`. It's clear what it does, but it's very arguable if it needs to be configured at all by the user in the config.

So in essence, this issue is about removing unnecessary configuration params and refining docs and names for it is necessary

vgonkivs:
## Overview
Resolves #1667 

## Checklist

### DIFF
--- a/libs/header/interface.go
+++ b/libs/header/interface.go
@@ -8,6 +8,11 @@ import (
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 )
 
+const (
+	// MaxRangeRequestSize defines the max amount of headers that can be handled/requested at once.
+	MaxRangeRequestSize uint64 = 512
+)
+
 // Subscriber encompasses the behavior necessary to
 // subscribe/unsubscribe from new Header events from the
 // network.
--- a/libs/header/p2p/exchange.go
+++ b/libs/header/p2p/exchange.go
@@ -20,6 +20,9 @@ import (
 
 var log = logging.Logger("header/p2p")
 
+// minResponses the target minimum amount of responses with the same chain head
+const minResponses = 2
+
 // Exchange enables sending outbound HeaderRequests to the network as well as
 // handling inbound HeaderRequests from the network.
 type Exchange[H header.Header] struct {
@@ -60,9 +63,6 @@ func NewExchange[H header.Header](
 		peerTracker: newPeerTracker(
 			host,
 			connGater,
-			params.MaxAwaitingTime,
-			params.DefaultScore,
-			params.MaxPeerTrackerSize,
 		),
 		Params: params,
 	}, nil
@@ -144,7 +144,7 @@ LOOP:
 		}
 	}
 
-	return bestHead[H](result, ex.Params.MinResponses)
+	return bestHead[H](result)
 }
 
 // GetByHeight performs a request for the Header at the given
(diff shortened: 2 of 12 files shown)


renaynay (approved):
yey

renaynay (approved):

libs/header/p2p/options.go
```diff
@@ -120,26 +101,14 @@ func WithNetworkID[T parameters](networkID string) Option[T] {
...
+	// RangeRequestTimeout defines a timeout after which the session will try to re-request headers
 	// from another peer.
-	RequestTimeout time.Duration // TODO: Rename to RangeRequestTimeout
-	// TrustedPeersRequestTimeout a timeout for any request to a trusted peer.
-	TrustedPeersRequestTimeout time.Duration
-	// MaxTrackerSize specifies the max amount of peers that can be added to the peerTracker.
-	MaxPeerTrackerSize int
+	RangeRequestTimeout time.Duration
+	// TrustedPeersRangeRequestTimeout a timeout for any request to a trusted peer.
+	TrustedPeersRangeRequestTimeout time.Duration
```
Wondertan, line 111: Why is this RangeRequest? We only have special single header requests to trusted peers, correct?

libs/header/interface.go
```diff
@@ -8,6 +8,11 @@ import (
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 )
 
+const (
+	// MaxRangeRequestSize defines the max amount of headers that can be handled/requested at once.
+	MaxRangeRequestSize uint64 = 512
+)
+
```
Wondertan, line 15: I am not opposed to this one being public and want to understand why only this is part of the public interface and all other changed params are private.

libs/header/p2p/exchange.go
```diff
@@ -20,6 +20,9 @@ import (
 
 var log = logging.Logger("header/p2p")
 
+// minResponses the target minimum amount of responses with the same chain head
+const minResponses = 2
```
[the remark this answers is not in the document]
Wondertan: minTrustedHeadResponses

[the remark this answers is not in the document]
Wondertan: Also, let's make a comment explaining what it does. Current comment is very ambiguous

libs/header/interface.go
vgonkivs: Which one exactly? Other options are related to the peerTracker and should be hidden from the user.

```
const (
 	// defaultScore specifies the score for newly connected peers.
 	defaultScore float32 = 1
 	// maxTrackerSize specifies the max amount of peers that can be added to the peerTracker.
 	maxPeerTrackerSize = 100
 )

 var (
 	// maxAwaitingTime specifies the duration that gives to the disconnected peer to be back online,
 	// otherwise it will be removed on the next GC cycle.
 	maxAwaitingTime = time.Hour
 	// gcCycle defines the duration after which the peerTracker starts removing peers.
 	gcCycle = time.Minute * 30
 )
```

vgonkivs: The second argument is that all these options are used in one place(closely to the peerTracker). But we are using `MaxRangeRequestSize` in two different packages.

libs/header/p2p/options.go
vgonkivs: oops. It seems that `variable refactoring` in Goland changed this one too. Thanks

libs/header/interface.go
Wondertan: > But we are using MaxRangeRequestSize in two different packages.
ah, right ok

Wondertan (approved):

renaynay (approved):

PR #1831, opened 2023-02-27
refactoring!(libs/header): simplify return params in store.Append

vgonkivs:
## Overview
Amount of added headers to the store is redundant information, IMO. The only thing that we should know is whether `Append` was successful or not
## Checklist

### DIFF
--- a/libs/header/interface.go
+++ b/libs/header/interface.go
@@ -95,7 +95,7 @@ type Store[H Header] interface {
 	// as it applies them contiguously on top of the current head height.
 	// It returns the amount of successfully applied headers,
 	// so caller can understand what given header was invalid, if any.
-	Append(context.Context, ...H) (int, error)
+	Append(context.Context, ...H) error
 }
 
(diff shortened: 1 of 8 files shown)


Wondertan (commented):
Originally, when verification of headers happend only on Store level, it made sense to have such protection, as networking layer could give us incorrect header somewhere in between the range. Now, as we moving towards VerifiedRangeByHeight this becomes unnecessary and we can safely remove it.

Wondertan:
Also not that this is breaking and refactor type of PR, not misc

Wondertan (approved):

Wondertan:
Units are still seems to be failing on CI

Wondertan (approved):
Not a feature but refactoring

Wondertan:
@vgonkivs, why dic you remove breaking?

renaynay (approved):

renaynay:
PR is breaking, pls mark

PR #1855, opened 2023-03-02
fix(share/ipld): use parent context for ipld session instead of request context

walldiss:
## Overview
Introduction of cascade getter created a case where ctx is created just before call to `ipld.GetShare` and this local context is canceled right after the call. 

It turns out ipld.Session is created with ctx from first call that created it, instead of parent context from Availability. Session context should common parent context from ShareAvailable call, instead of local single request call.

### DIFF
--- a/share/availability/full/reconstruction_test.go
+++ b/share/availability/full/reconstruction_test.go
@@ -4,7 +4,6 @@ package full
 
 import (
 	"context"
-	"sync"
 	"testing"
 	"time"
 
@@ -214,12 +213,9 @@ func TestShareAvailable_DisconnectedFullNodes(t *testing.T) {
 	lights1, lights2 := make(
 		[]*availability_test.TestNode, lightNodes/2),
 		make([]*availability_test.TestNode, lightNodes/2)
-	var wg sync.WaitGroup
-	wg.Add(lightNodes)
 	for i := 0; i < len(lights1); i++ {
 		lights1[i] = light.Node(net)
 		go func(i int) {
-			defer wg.Done()
 			err := lights1[i].SharesAvailable(ctx, root)
 			if err != nil {
 				t.Log("light1 errors:", err)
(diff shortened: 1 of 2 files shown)


share/availability/full/reconstruction_test.go
```diff
@@ -260,4 +265,5 @@ func TestShareAvailable_DisconnectedFullNodes(t *testing.T) {
 	require.NoError(t, err, share.ErrNotAvailable)
 	err = full2.SharesAvailable(ctx, root)
 	require.NoError(t, err, share.ErrNotAvailable)
+	wg.Wait()
```
Wondertan, line 268: Why? We don't really care if the shares available end for lights. Or was there a panic about test fail in the routine after the test ended?

Wondertan: Also, it would be nice to extract this into separate PR

share/getters/ipld.go
```diff
@@ -109,14 +109,15 @@ var sessionKey = &session{}
 // session is a struct that can optionally be passed by context to the share.Getter methods using
 // WithSession to indicate that a blockservice session should be created.
 type session struct {
+	ctx context.Context
```
Wondertan, line 112: nit: put as the last field

share/availability/full/reconstruction_test.go
walldiss: This test panicing now. The panic is caused by error log happening after test has exit. I just don't like panicing tests in main.

Wondertan: Is it constantly panicing? We can also merge the extracted PR first before this one

walldiss: it panics in CI only, but did it 3 out of 3 times

walldiss:
I've extracted broken test fix into separate PR https://github.com/celestiaorg/celestia-node/pull/1858

renaynay (approved):

Wondertan (approved):

PR #1859, opened 2023-03-03
fix(share/peer-manager): round robin pointer out of range

walldiss:
## Overview

Fixes possible situations when pointer gets out of range

## Checklist

### DIFF
--- a/share/p2p/peers/pool.go
+++ b/share/p2p/peers/pool.go
@@ -54,6 +54,11 @@ func (p *pool) tryGet() (peer.ID, bool) {
 		return "", false
 	}
 
+	// if pointer is out of range, point to first element
+	if p.nextIdx > len(p.peersList)-1 {
+		p.nextIdx = 0
+	}
+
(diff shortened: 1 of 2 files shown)


share/p2p/peers/pool.go
```diff
@@ -54,6 +54,11 @@ func (p *pool) tryGet() (peer.ID, bool) {
 		return "", false
 	}
 
+	// if pointer is out of range, point to first element
+	if p.nextIdx > len(p.peersList)-1 {
```
renaynay, line 58: this always assumes there's at least 1 peer in peerslist yeah? is it possible that there are none -- meaning activeCount > 0 but peerslist ==0

Wondertan (approved):
The change makes total sense

renaynay (approved):

share/p2p/peers/pool.go
walldiss: It should be impossible

Issue #1782, opened 2023-02-20
das: add retry failed headers during runtime


walldiss:
### Implementation ideas

DASer needs to do retry for failed headers. As for now, it does retry failed headers only upon restart. Need to add mechanics to let DASer retry without restart. 

Initial idea is to implement limited size worker pool that will be run only for failed headers. Considering we don't expect many failures to happen, worker pool size could be just 1. The failed worker need to run with higher timeout than normal one, to allow blocks with low availability in network to be sampled.

PR #1860, opened 2023-03-03
feat(daser): introduce retry logic for daser

walldiss:
## Overview
DASer jobs now have `job_type` field that identifies purpose of the job. There are 3 possible types of jobs:

 - catchup
 - recent
 - retry

Introduce new type of jobs for DASer - `retry`. `retry` jobs will be launched only if there is enough concurrency slots after running  `recent` jobs and `catchup` jobs. Meaning DASer will start retry with lowest priority, only when catchup is done and recent header are synced without falling behind. 

Also includes minor simplifications for affected code.

Resolves https://github.com/celestiaorg/celestia-node/issues/1782, https://github.com/celestiaorg/celestia-node/issues/2027

## Checklist


das/state.go
```diff
@@ -101,59 +106,51 @@ func (s *coordinatorState) newRecentJob(header *header.ExtendedHeader) job {
...
-		To:             height,
+		id:     s.nextJobID,
+		kind:   recentJob,
+		header: header,
+		From:   height,
+		To:     height,
 	}
 }
 
 // nextJob will return header height to be processed and done flag if there is none
```
renaynay, line 117: can you fix the doc?

das/checkpoint.go
```diff
@@ -25,6 +26,7 @@ func newCheckpoint(stats SamplingStats) checkpoint {
 		workers = append(workers, workerCheckpoint{
 			From: w.Curr,
 			To:   w.To,
+			Kind: w.Kind,
```
Bidon15, line 29: [proposal][non-blocking]: can we have a dedicated issue to keep this awesome [ADR-012](https://github.com/celestiaorg/celestia-node/blob/main/docs/adr/adr-012-daser-parallelization.md) up to date? 🙏

das/stats.go
```diff
@@ -22,6 +22,7 @@ type SamplingStats struct {
 }
 
 type WorkerStats struct {
+	Kind string `json:"kind"`
```
Bidon15, line 25: Awesome!

das/checkpoint.go
walldiss: Good suggestion, thank you! I've opened a tracking issue https://github.com/celestiaorg/celestia-node/issues/1886

das/coordinator.go
```diff
@@ -36,7 +36,7 @@ type samplingCoordinator struct {
 // result will carry errors to coordinator after worker finishes the job
 type result struct {
 	job
-	failed []uint64
+	failed map[uint64]int
```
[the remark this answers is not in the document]
walldiss: If any headers within retry job will fail again, the try count will carried over and header will be retried by another job. Only 1 attempt per header within worker / job.

das/state.go
```diff
@@ -12,9 +13,9 @@ type coordinatorState struct {
 	sampleFrom    uint64 // is the height from which the DASer will start sampling
 	samplingRange uint64 // is the maximum amount of headers processed in one job.
 
-	retry      []job                      // list of headers heights that will be retried after last run
 	inProgress map[int]func() workerState // keeps track of running workers
 	failed     map[uint64]int             // stores heights of failed headers with amount of attempt as value
+	retry      map[int]int                // height that previously failed and currently are being process by workers
```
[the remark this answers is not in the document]
walldiss: Yes, it uses jobID as a key. Will document

das/state.go
```diff
@@ -42,34 +43,38 @@ func newCoordinatorState(params Parameters) coordinatorState {
...
 
 func (s *coordinatorState) handleResult(res result) {
 	delete(s.inProgress, res.id)
 
-	failedFromWorker := make(map[uint64]bool)
-	for _, h := range res.failed {
-		failedFromWorker[h] = true
-	}
-
 	// check if the worker retried any of the previously failed heights
```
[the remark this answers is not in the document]
walldiss: Not necessary. It could be normal `catchup` job going over headers that has failed by `recent` job.

das/state.go
```diff
@@ -42,34 +43,38 @@ func newCoordinatorState(params Parameters) coordinatorState {
...
+			attempts += s.retry[res.job.id]
+		}
+		delete(s.retry, res.job.id)
+	}
+
 	// add newly failed heights
-	for h := range failedFromWorker {
-		s.failed[h]++
+	for h := range res.failed {
+		s.failed[h] = +attempts
```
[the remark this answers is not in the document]
walldiss: This is type. For some reason it is valid syntax, but with unexpected results:
```
var i int
fmt.Println("i", i)
i += 1
fmt.Println("+=", i)
i = +1
fmt.Println("= +", i)
i += +1
fmt.Println("+= +", i)
```
	
Compiles, results:
```
i 0
+= 1
=+ 1
= + 1
+= + 2
```

Wondertan (commented):
Do we have backoff logic for retries? So we retry but each try the timeout increases

Wondertan:
Consider a case where we synced up to the tip, but there are few unavailable heights. The DASer will retry them indefinitely without any time gap between. There is no good reason to make a busy loop for network requests the results of which are unlikely to change on milliseconds scale.

walldiss:
thank you @Wondertan, thats a good point. I'll create an issue for retry backoff in DASer

das/state.go
```diff
@@ -42,34 +52,38 @@ func newCoordinatorState(params Parameters) coordinatorState {
...
+	// update failed heights
+	for h := range res.failed {
+		failCount := 1
+		if res.job.Type == retryJob {
+			// if job was already in retry and failed again, persist attempt count
+			failCount += s.inRetry[h]
+			delete(s.inRetry, h)
+		}
+
+		s.failed[h] = failCount
```
renaynay, line 86: It's a bit confusing here - if the job was already retried once before, it should already be in the `s.failed` map so then just increment that value for the failed height by one, no? (this is in ref to my other comment regarding storing num retry attempts in result)

das/state.go
```diff
@@ -93,67 +107,76 @@ func (s *coordinatorState) updateHead(newHead int64) {
...
 }
 
-func (s *coordinatorState) nextFromRetry() (job, bool) {
-	if len(s.retry) == 0 {
-		return job{}, false
-	}
+// retryJob creates a job to retry previously failed header
+func (s *coordinatorState) retryJob() (next job, found bool) {
+	for h, count := range s.failed {
+		// TODO(@walldiss): limit max amount of retries until retry backoff is implemented
```
renaynay, line 156: pls make issue for this if you haven't already. I think that we should handle backoff a bit differently bc if you retry the height 3+ times and you still can't sample it, that means something's pretty wrong.

das/stats.go
```diff
@@ -37,3 +38,12 @@ func (s SamplingStats) totalSampled() uint64 {
 	}
 	return s.CatchupHead - inProgress - uint64(len(s.Failed))
 }
+
+// totalSampled returns the total amount of sampled headers
```
renaynay, line 42: godoc

das/state.go
walldiss: When `retryJob` is created, failed height moved from failed map to inRetry map, so next job won't pick same height. If the job fails again, retry count needs to be restored from `inRetry` map and also be incremented by failed count that happen inside job (currently always 1).

For some reason I cannot answer to another comment about num of retry, so will post answer here:
I've moved from slice to map for fast lookup in the map. the failed count could only be 1 at this point, but it is much so 
 much nicer to work with actual fail count then checking and filling empty structs in multiple places, so I've sacrificed 4 bytes for that.

das/state.go
walldiss: There is an issue for that https://github.com/celestiaorg/celestia-node/issues/2027

das/stats.go
```diff
@@ -22,9 +22,10 @@ type SamplingStats struct {
 }
 
 type WorkerStats struct {
-	Curr uint64 `json:"current"`
-	From uint64 `json:"from"`
-	To   uint64 `json:"to"`
+	JobType jobType `json:"job_type"`
```
distractedm1nd, line 25: Please update api/docgen/exampledata/samplingstats to include this - you can just run `celestia rpc das SamplingStats | jq` on a running node and paste the output into that file

das/worker.go
```diff
@@ -26,21 +32,22 @@ type worker struct {
...
-	isRecentHeader bool
-	header         *header.ExtendedHeader
+	id      int
+	jobType jobType
+	from    uint64
+	to      uint64
 
-	From uint64
-	To   uint64
+	// header will be filled for job with type recentJob to avoid unnecessary call to the header store
```
distractedm1nd, line 49:
```suggestion
	// header is set only for recentJobs, avoiding an unnecessary call to the header store
```
optional

das/state.go
```diff
@@ -93,69 +107,78 @@ func (s *coordinatorState) updateHead(newHead int64) {
...
+		if count > 3 {
```
distractedm1nd, line 157: This wont be here for long right? Backoff is coming next?

distractedm1nd (changes requested):

das/stats.go
distractedm1nd: this is my only blocking comment by the way, the rest are optional

walldiss: TIL. thank you!

das/metrics.go
```diff
@@ -15,6 +15,12 @@ import (
 	"github.com/celestiaorg/celestia-node/header"
 )
 
+const (
+	jobTypeLabel     = "job_type"
+	headerWidthLabel = "header_width"
+	failedLabel      = "failed"
+)
```
derrandz, line 22: why did you choose to go with this approach? I remember trying this before and there was feedback and preference to not use constants as the compiler will strip them away anyway. I personally am in favor of them generally, they make things easier to update

derrandz (dismissed):
👍🏻

das/state.go
walldiss: Added backoff, pls check latest commit.

das/metrics.go
walldiss: Why do you think compiler will strip them away? As far as I understand constants should be more efficient, because they are not allocated in the runtime.
But main motivation is the fact that labels are reused for multiple metrics and I wanted to be sure they are aligned nicely everywhere.

distractedm1nd: @derrandz IIRC that feedback was for single use strings that were extracted as consts (which are automatically set by compiler) - if there are multiple uses its fine

das/coordinator.go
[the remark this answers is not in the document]
distractedm1nd: I looked into this and also think an array of failed headers is enough, I am posting the diff here:

```diff
diff --git a/das/coordinator.go b/das/coordinator.go
index 2184dce2..4162104d 100644
--- a/das/coordinator.go
+++ b/das/coordinator.go
@@ -37,7 +37,7 @@ type samplingCoordinator struct {
 // result will carry errors to coordinator after worker finishes the job
 type result struct {
 	job
-	failed map[uint64]int
+	failed []uint64
 	err    error
 }
 
diff --git a/das/state.go b/das/state.go
index cbf7e83e..0823747e 100644
--- a/das/state.go
+++ b/das/state.go
@@ -75,7 +75,7 @@ func (s *coordinatorState) handleResult(res result) {
 	}
 
 	// update failed heights
-	for h := range res.failed {
+	for _, h := range res.failed {
 		failCount := 1
 		if res.job.jobType == retryJob {
 			// if job was already in retry and failed again, persist attempt count
@@ -203,7 +203,7 @@ func (s *coordinatorState) unsafeStats() SamplingStats {
 			ErrMsg:  errMsg,
 		})
 
-		for h := range wstats.failed {
+		for _, h := range wstats.failed {
 			failed[h]++
 			if h < lowestFailedOrInProgress {
 				lowestFailedOrInProgress = h
diff --git a/das/state_test.go b/das/state_test.go
index 1bef1ba2..7f224994 100644
--- a/das/state_test.go
+++ b/das/state_test.go
@@ -26,7 +26,7 @@ func Test_coordinatorStats(t *testing.T) {
 									from:    21,
 									to:      30,
 								},
-								failed: map[uint64]int{22: 1},
+								failed: []uint64{22},
 								err:    errors.New("22: failed"),
 							},
 							curr: 25,
@@ -40,7 +40,7 @@ func Test_coordinatorStats(t *testing.T) {
 									from:    11,
 									to:      20,
 								},
-								failed: map[uint64]int{12: 1, 13: 1},
+								failed: []uint64{12, 13},
 								err:    errors.Join(errors.New("12: failed"), errors.New("13: failed")),
 							},
 							curr: 15,
diff --git a/das/worker.go b/das/worker.go
index 4319928b..497d5643 100644
--- a/das/worker.go
+++ b/das/worker.go
@@ -50,7 +50,8 @@ type job struct {
 	header *header.ExtendedHeader
 }
 
-func newWorker(j job,
+func newWorker(
+	j job,
 	getter libhead.Getter[*header.ExtendedHeader],
 	sample sampleFn,
 	broadcast shrexsub.BroadcastFn,
@@ -64,8 +65,7 @@ func newWorker(j job,
 		state: workerState{
 			curr: j.from,
 			result: result{
-				job:    j,
-				failed: make(map[uint64]int),
+				job: j,
 			},
 		},
 	}
@@ -182,7 +182,7 @@ func (w *worker) setResult(curr uint64, err error) {
 	w.lock.Lock()
 	defer w.lock.Unlock()
 	if err != nil {
-		w.state.failed[curr]++
+		w.state.failed = append(w.state.failed, curr)
 		w.state.err = errors.Join(w.state.err, fmt.Errorf("height: %d, err: %w", curr, err))
 	}
 	w.state.curr = curr

```

distractedm1nd (changes requested):
Only thing I would change is using the array instead of a map, just because it is easier to understand (the map definitely confused me a bit when trying to grok the pr). But if you have a reason against it, its fine and I will approve PR

walldiss:
@distractedm1nd I agree that array is easier to understand representation. I've used it to make O(1) lookup [here](https://github.com/celestiaorg/celestia-node/blob/31594d743d3f3dd285f9d3a1ae71c53e1e35f39f/das/state.go#L72). I think your diff misses this usage.

das/state.go
```diff
@@ -93,69 +107,78 @@ func (s *coordinatorState) updateHead(newHead int64) {
...
+	s.next = to + 1
 	return j, true
 }
 
-func (s *coordinatorState) nextFromRetry() (job, bool) {
-	if len(s.retry) == 0 {
-		return job{}, false
-	}
+// retryJob creates a job to retry previously failed header
+func (s *coordinatorState) retryJob() (next job, found bool) {
```
Wondertan, line 154: Would be nice if this logic could find adjacent headers that are failed and put into one job

das/state.go
```diff
@@ -93,69 +107,78 @@ func (s *coordinatorState) updateHead(newHead int64) {
...
+		id:      s.nextJobID,
+		jobType: recentJob,
+		header:  header,
+		from:    height,
+		to:      height,
 	}
 }
 
-// nextJob will return header height to be processed and done flag if there is none
+// nextJob will return next job according to priority (catchup -> retry)
```
Wondertan, line 127:
```suggestion
// nextJob will return next catchup or retry job according to priority (catchup > retry) 
```

Wondertan (commented):
* Job-type refactoring makes total sense
* Again reassured me of the need for backoff
* Left only minor comments, LGTM

das/worker.go
```diff
@@ -75,20 +84,17 @@ func (w *worker) run(ctx context.Context, timeout time.Duration, resultCh chan<-
...
 		"finished (s)", time.Since(jobStart),
 	)
 
 	select {
-	case resultCh <- result{
-		job:    w.state.job,
-		failed: w.state.failed,
-		err:    w.state.Err,
-	}:
+	case resultCh <- w.state.result:
```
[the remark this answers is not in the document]
walldiss: not really, I've clean up structures a bit by removing duplicating fields. Saved allocation is also nice

das/state.go
```diff
@@ -93,69 +107,78 @@ func (s *coordinatorState) updateHead(newHead int64) {
...
+		// TODO(@walldiss): limit max amount of retries until retry backoff is implemented
+		if count > 3 {
+			continue
+		}
 
-	next := s.retry[len(s.retry)-1]
-	s.retry = s.retry[:len(s.retry)-1]
+		// move header from failed into retry
+		delete(s.failed, h)
+		s.inRetry[h] = count
```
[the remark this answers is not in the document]
walldiss: count is amount of failed attempts, it will be increased in result handnling

das/state.go
walldiss: Would land in another issue. This one is already hard to review

distractedm1nd (approved):

renaynay (approved):
+1 to ryan's suggestion

PR #1869, opened 2023-03-06
Extend label automation to add external label

MSevey:
## Overview

Rene was interested in external issue labels, was easier to just do it then explain it in slack :-) 

## Checklist

### DIFF
--- a/.github/workflows/issue-label-automation.yml
+++ b/.github/workflows/issue-label-automation.yml
@@ -8,9 +8,24 @@ jobs:
     permissions:
       issues: write
     steps:
+      - name: Check for External Contributor
+        uses: tspascoal/get-user-teams-membership@v2
+        id: teamCheck
+        with:
+          username: ${{ github.actor }}
+          team: "celestia-node"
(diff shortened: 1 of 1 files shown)


Wondertan (approved):

vgonkivs (approved):

Wondertan:
Seems like there is no `external` label created yet, will it automatically create one?

Wondertan:
Can it mark PRs or only issues?

MSevey:
> Seems like there is no `external` label created yet, will it automatically create one?

not sure, I went ahead and created a label just in case.

MSevey:
> Can it mark PRs or only issues?

I'll follow up on the PRs, there is definitely a way to do it, just not sure if the current action will target PRs or if it is specific to issues.

MSevey:
@Wondertan the `https://github.com/andymckay/labeler` action does support PR labels it looks like. 

If you'd like me to submit the change for that I'd recommend we merge this and then I would do a follow up as there are a few things I would need to test out.

MSevey:
Also confirmed for the future that if the label doesn't exist it will be created when it is first run

PR #1879, opened 2023-03-09
fix(cmd/cel-key): Ensure keyring-backend is actually `test` by default

### LINKED ISSUE #1876: Cel-key doesn't correctly set default keyring backend
### Celestia Node version

b7213bfa9e84098834bc869a465cfc4cbb85b7ba

### OS

Ubuntu 22.04.2 LTS

### Install tools

_No response_

### Others

_No response_

### Steps to reproduce it

Create any key with explicit `test` keyring backend with empty initial keyring.
```bash
cel-key add fookey --keyring-backend test --node.type light --p2p.network arabica
```
List all keys for the node type + network pair, omitting the backend parameter
```bash
cel-key list --node.type light --p2p.network arabica
```

### Expected result

List of keys containing the newly created key.

### Actual result

`No records were found in keyring`

### Relevant log output

_No response_

### Notes

If one runs the `cel-key list --help` he could see that `test` backend is assumed to be default.
```
--keyring-backend string   Select keyring's backend (os|file|test) (default "test")
```

renaynay:
`KeysCommand` from the sdk is quite wonky in that it really only sets a keyring backend if the flag was `changed`, so this hack ensures we can properly set a default and that the default is `test`.

Resolves #1876

### DIFF
--- a/cmd/cel-key/main.go
+++ b/cmd/cel-key/main.go
@@ -2,13 +2,13 @@ package main
 
 import (
 	"context"
-	"github.com/cosmos/cosmos-sdk/crypto/keyring"
 	"os"
 
 	"github.com/cosmos/cosmos-sdk/client"
 	"github.com/cosmos/cosmos-sdk/client/config"
 	"github.com/cosmos/cosmos-sdk/client/flags"
 	"github.com/cosmos/cosmos-sdk/client/keys"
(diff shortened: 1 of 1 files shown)


Wondertan (approved):
👍🏻

vgonkivs (approved):

PR #1723, opened 2023-02-09
chore: bump celestia-app to v0.12.0-rc7

evan-forbes:
## Overview

very minor bump, with the only celestia-node relevant change being that we can remove the secondary `config.Seal` in this repo now that we're doing it in a single spot in the application.

## Checklist

- [x] New and updated code has appropriate documentation
- [x] New and updated code has new and/or updated testing
- [x] Required CI checks are passing
- [x] Visual proof for any user facing features like CLI or documentation updates
- [x] Linked issues closed with keywords

PR #1881, opened 2023-03-10
refactor(nodebuilder/p2p)!: Configure bootstrapper from env instead of config

smuu:
## Overview

This PR introduces the environment variable BOOTSTRAPPER, which starts the node in bootstrapper mode.
It also removes the bootstrapper setting from the config.

## Checklist

closes celestiaorg/devops#54


jrmanes (dismissed):
LGTM 👌

sysrex (commented):
Since boots trappers are only ran by Celestia, I don't think we should put it in the node public documentation. There is no scenario in which an end user will care about it or want to run it.

renaynay:
> Since boots trappers are only ran by Celestia, I don't think we should put it in the node public documentation. There is no scenario in which an end user will care about it or want to run it.

There may be other entities that would want to run nodes in bootstrapper mode - whether they want to run a bootstrapper for the testnets or for their own private networks.

README.md
```diff
@@ -68,13 +70,19 @@ More information can be found [here](https://github.com/celestiaorg/celestia-nod
...
 
 ```sh
 celestia <node_type> start
 ```
 
+### Environment variables
+
+| Variable       | Explanation                         | Default value | Required |
+| -------------- | ----------------------------------- | ------------- | -------- |
+| `BOOTSTRAPPER` | Start the node in bootstrapper mode | `false`       | Optional |
```
Bidon15, line 84: [proposal][non-blocking]: Please create a separate issue in node repo explaining that there are other variables need to be documented 🙏 , like
```
CELESTIA_CUSTOM=<chain-id>:<gen-hash>
```

nodebuilder/util/env.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"os"
+)
+
+var IsBootstrapper bool
+
+func init() {
+	IsBootstrapper = getEnv("BOOTSTRAPPER", "false") == "true"
+}
```
Bidon15, line 11: [blocking]: this can hit us hard in testground tests. 
Can we do it without `func init()` 🙏

Bidon15 (commented):
Awesome job @smuu 🥳 
1 small request from me to think about 🙏

nodebuilder/util/env.go
Bidon15: Imagine we repeat the same scenario like with sealing config for account prefix balances
This https://github.com/celestiaorg/celestia-node/pull/1583

Caused tg test to fail on sealing issues due to init from both sides (app/node) when used a go.mod for testground tests 

Which caused this PR to appear and later removal of init in node/core/config.go
https://github.com/celestiaorg/celestia-app/pull/1348/

It took a whole month for the proper RC release-fix so the testground test can use RCs without forks 

Proof: https://github.com/celestiaorg/celestia-node/pull/1723

Thus, I propose we stop such practices in node repo that is flooded in sdk/app and rely on different mechanism on how to check/set any Env var in the future.

smuu: Is it okay to use blank identifiers like [this](https://github.com/celestiaorg/celestia-node/pull/1881/commits/85cb8a533acae96ccc9bd4d7d73c16243c3639d9)? Or do we have the same issue then?

README.md
smuu: https://github.com/celestiaorg/celestia-node/issues/1896

nodebuilder/util/env.go
Wondertan: These are different stories. The issue you've mentioned can only happen in this case can happen if the key value for env var is the same across different projects. This shouldn't happen between the app and node because the app does not have a notion of a bootstrapper, but a seed node.

Anyway, the node operators may run other software that has such a notion, so we should rename `BOOTSTRAPPER` to `CELESTIA_BOOTSTRAPPER`

Wondertan: JFYI, having it as @smuu did it [here](https://github.com/celestiaorg/celestia-node/pull/1881/commits/85cb8a533acae96ccc9bd4d7d73c16243c3639d9) is effectively the same as using init func.

nodebuilder/util/env.go
```diff
@@ -0,0 +1,17 @@
+package util
```
Wondertan, line 1: Please don't. The previous version where you put it into p2p was ok. The notion of Bootstapper is logically a p2p thing, so it should be there rather than some utility. Ther is bootstrapp.go file where you can put this to.

Wondertan (changes requested):
Thanks for doing this. Also check another comment on renaming env var

nodebuilder/tests/p2p_test.go
```diff
@@ -91,7 +92,7 @@ Steps:
 func TestBootstrapNodesFromBridgeNode(t *testing.T) {
 	sw := swamp.NewSwamp(t)
 	cfg := nodebuilder.DefaultConfig(node.Bridge)
-	cfg.P2P.Bootstrapper = true
+	p2p.IsBootstrapper = true
```
Wondertan, line 95: Let's remove ___all___ the `p2p.IsBootstrapper = true`. Previously when we set it through the config, it was affecting only the specific node, but with global var it makes everyone a bootstrapper. It's fine to remove it; even we had `cfg.P2P.Bootstrapper = true` before. Our tests should not rely on assumption that someone is a bootstrapper and someone is not. Even if it breaks them now, will will fix them

Wondertan (changes requested):
Another request

nodebuilder/p2p/bootstrap.go
```diff
@@ -1,10 +1,26 @@
...
+
 	"github.com/libp2p/go-libp2p/core/peer"
 	ma "github.com/multiformats/go-multiaddr"
 )
 
+var IsBootstrapper bool
+
+func init() {
+	IsBootstrapper = getEnv("CELESTIA_BOOTSTRAPPER", "false") == "true"
+}
```
Wondertan, line 14: I think making this oneliner is fine btw. It's less code with the same effect

nodebuilder/p2p/bootstrap.go
```diff
@@ -1,10 +1,24 @@
...
 import (
+	"os"
+
 	"github.com/libp2p/go-libp2p/core/peer"
 	ma "github.com/multiformats/go-multiaddr"
 )
 
+var (
+	IsBootstrapper = getEnv("CELESTIA_BOOTSTRAPPER", "false") == "true"
+)
```
Wondertan, line 12: Not
```suggestion
var IsBootstrapper = getEnv("CELESTIA_BOOTSTRAPPER", "false") == "true"
```

Wondertan (approved):
LGTM and thanks

Wondertan (dismissed):

Bidon15 (approved):
Thanks for getting it done 👍

nodebuilder/p2p/bootstrap.go
```diff
@@ -1,10 +1,22 @@
 package p2p
 
 import (
+	"os"
+
 	"github.com/libp2p/go-libp2p/core/peer"
 	ma "github.com/multiformats/go-multiaddr"
 )
 
+var IsBootstrapper = getEnv("CELESTIA_BOOTSTRAPPER", "false") == "true"
```
walldiss, line 10: I think it would be more go-way if `IsBootstrapper` was a function and it would populate `Config` struct field during struct creation. Global var are usually avoided due to ability to be overwrited from elsewhere.

Wondertan: We want to read from env var can be overwritten anytime

nodebuilder/util/env.go
Wondertan: I still don't see why removing init changes anything for your case. 
Considering the approve, I am resolving this

nodebuilder/p2p/bootstrap.go
walldiss: But we need to read env only in nodebuilder, meaning it will be called only on launch. There is no need to store it in var after node has started.

Wondertan: That's fine I you feel strongly about it.

Wondertan: @smuu, pls apply the suggestion

walldiss (approved):

vgonkivs (approved):

PR #1883, opened 2023-03-10
fix(header/p2p): return on context to avoid ugly errors on stop

Wondertan:
Self-explanatory

### DIFF
--- a/libs/header/p2p/session.go
+++ b/libs/header/p2p/session.go
@@ -14,7 +14,8 @@ import (
 	p2p_pb "github.com/celestiaorg/celestia-node/libs/header/p2p/pb"
 )
 
-// errEmptyResponse means that server side closes the connection without sending at least 1 response.
+// errEmptyResponse means that server side closes the connection without sending at least 1
+// response.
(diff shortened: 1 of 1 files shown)


libs/header/p2p/session.go
```diff
@@ -174,6 +176,7 @@ func (s *session[H]) doRequest(
 
 		select {
 		case <-s.ctx.Done():
+			return
```
vgonkivs, line 179: Nice catch

vgonkivs (approved):

libs/header/p2p/session.go
renaynay: Wait didn't I mention this in one of the PRs?

renaynay (approved):

libs/header/p2p/session.go
vgonkivs: It seems that it was missed during resolving merge conflicts 😞

vgonkivs:
I think we can stop using `area:header` label in `lib/header` related PRs in favour of `libs:header`.

PR #1884, opened 2023-03-10
feat(share/shrex): collect errors from shrex getter

walldiss:
## Overview

Shrex getter client needs to collect and propagate request errors to the caller. Also introduce per error logging.


share/getters/shrex.go
```diff
@@ -62,32 +64,45 @@ func (sg *ShrexGetter) GetShare(ctx context.Context, root *share.Root, row, col
...
 			return eds, nil
 		case context.DeadlineExceeded:
-			log.Debugw("request exceeded deadline, trying with new peer", "datahash", root.String())
 		case p2p.ErrInvalidResponse:
 			setStatus(peers.ResultBlacklistPeer)
 		default:
 			setStatus(peers.ResultCooldownPeer)
 		}
+
+		err = multierr.Append(err, getErr)
```
Wondertan, line 99: Join 🤤

Wondertan (approved):
Still disagree that we should return an error for each attempt, especially if the errors are the same on each try, but it does not hurt so approve

share/getters/shrex.go
walldiss: Found 1.20 enthusiast.

BTW good point about collecting duplicate errors. I"ll add extra filter

Wondertan (dismissed):
Very nice

vgonkivs (approved):

Wondertan (approved):

share/getters/shrex.go
```diff
@@ -97,33 +113,46 @@ func (sg *ShrexGetter) GetSharesByNamespace(
...
-			log.Debugw("request exceeded deadline, trying with new peer", "datahash", root.String())
 		case p2p.ErrInvalidResponse:
 			setStatus(peers.ResultBlacklistPeer)
 		default:
 			setStatus(peers.ResultCooldownPeer)
 		}
+
+		if !ErrorContains(err, getErr) {
+			err = multierr.Append(err, getErr)
+		}
```
renaynay, line 150: I'm confused -- so if we get two peers who return to us an `ErrInvalidResponse` for example, we would only collect that err once?

renaynay (approved):

PR #1885, opened 2023-03-10
feat(share/shrex): cleanup shrex server side timeouts

walldiss:
## Overview
Depends on https://github.com/celestiaorg/celestia-node/pull/1884

Recent tests on arabica shows that 256 block sampling time could be up to 50s. Increase shrex server side timeouts accordingly.


share/p2p/shrexnd/client.go
```diff
@@ -178,15 +178,17 @@ func (c *Client) setStreamDeadlines(ctx context.Context, stream network.Stream)
 		return
 	}
 
-	if c.params.readTimeout != 0 {
-		err := stream.SetReadDeadline(time.Now().Add(c.params.readTimeout))
+	// client read deadline is the same as server write deadline
+	if c.params.serverWriteTimeout != 0 {
+		err := stream.SetReadDeadline(time.Now().Add(c.params.serverWriteTimeout))
```
Wondertan, line 183: Why do we set write deadline on reading on the client side? 

Also, if we want to use write timeout as global timeout for both read and write ops we can use SetDeadline, but before would be nice to understand why we want a global one

share/getters/shrex.go
```diff
@@ -71,10 +73,9 @@ func (sg *ShrexGetter) GetEDS(ctx context.Context, root *share.Root) (*rsmt2d.Ex
 		peer, setStatus, err := sg.peerManager.Peer(ctx, root.Hash())
 		if err != nil {
 			log.Debugw("couldn't find peer", "datahash", root.String(), "err", err)
-			return nil, fmt.Errorf("getter/shrex: %w", err)
 		}
 
-		reqCtx, cancel := ctxWithSplitTimeout(ctx, sg.minAttemptsCount, sg.minRequestTimeout)
+		reqCtx, cancel := ctxWithSplitTimeout(ctx, sg.minAttemptsCount-attempt+1, sg.minRequestTimeout)
```
Wondertan, line 78: So on each new attempt, we decrease the split factor, meaning timeout grows on each attempt until the split factor is zero, correct?

share/p2p/shrexnd/client.go
walldiss: We expect it reasonable to write request in 5sec and if it didn't happen we could try another peer faster that whole request timeout happens. Full request timeout in general is 1 min (there are cases when it could be different).

Wondertan: > We expect it reasonable to write request in 5sec and if it didn't happen we could try another peer faster that whole request timeout happens. 

Yeah, so why it's now `serverWriteTimeout` which is 1 min, instead of `serverReadTimeout` which is 5 secs?

walldiss: Because it client side timeouts (client.go). Client read timeout will either timeout itself or reach server write timeout. So for simplicity for now client read timeout == server write timeout.

nodebuilder/share/config.go
```diff
@@ -37,5 +46,11 @@ func (cfg *Config) Validate() error {
 	if cfg.DiscoveryInterval <= 0 || cfg.AdvertiseInterval <= 0 {
 		return fmt.Errorf("nodebuilder/share: %s", ErrNegativeInterval)
 	}
+	if err := cfg.ShrExNDParams.Validate(); err != nil {
+		return err
+	}
+	if err := cfg.ShrExEDSParams.Validate(); err != nil {
+		return err
+	}
 	return nil
```
renaynay, line 55: just return cfg.Shrexedsparams.Validate

share/p2p/params.go
```diff
@@ -0,0 +1,69 @@
...
+// Parameters is the set of parameters that must be configured for the shrex/eds protocol.
+type Parameters struct {
+	// ServerReadTimeout sets the timeout for reading messages from the stream.
+	ServerReadTimeout time.Duration
+
+	// ServerWriteTimeout sets the timeout for writing messages to the stream.
+	ServerWriteTimeout time.Duration
+
+	// HandleTimeout defines the deadline for handling request.
+	HandleTimeout time.Duration
```
renaynay, line 19:
```suggestion
	// HandleRequestTimeout defines the deadline for handling request.
	HandleRequestTimeout time.Duration
```

share/p2p/shrexnd/params.go
```diff
@@ -0,0 +1,18 @@
...
+
+	"github.com/celestiaorg/celestia-node/share/p2p"
+)
+
+const protocolString = "/shrex/nd/0.0.1"
+
+var log = logging.Logger("shrex/nd")
+
+// Parameters is the set of parameters that must be configured for the shrex/eds protocol.
+type Parameters = p2p.Parameters
```
renaynay, line 14: why do we have to alias?

walldiss: Relating to higher level p2p pkg inside shrex and in nodebuilder feels confusing, so I done it to make imports and usages to look nicer.

renaynay: okay

renaynay (approved):

renaynay:
Added breaking tag bc this breaks conf and requires users to re-init

Wondertan (approved):

PR #1682, opened 2023-02-02
bugfix(libs/syncer): fix data race in libs/syncer

vgonkivs:
## Overview
Resolves #1681 

## Checklist
- [ ] New and updated code has appropriate documentation
- [ ] New and updated code has new and/or updated testing
- [ ] Required CI checks are passing
- [ ] Visual proof for any user facing features like CLI or documentation updates
- [ ] Linked issues closed with keywords

PR #1889, opened 2023-03-10
refactor(header/sync): cleanup sync head detection logic

Wondertan:
Since #1682, we have kept the "synchronized head" in the Syncer because of synchronization oversight. However, the fix does not fit nicely into the model of subjective and network head and introduced more confusion and hairiness to the code.

This PR fixes the original issue and integrates the solution coherently into the code responsible for keeping and detecting the current head.

Additionally, sync_head.go is no more dependent on the adjacency of the Store and is prepared for future Store refactoring and storing of pending headers, instead of caching. Also, this PR does minor code cleanups, including comments and var name fixes.

Synced up to the tip 2 times on Arabica.

> Anybody needs their syncussy cleaned?


libs/header/sync/sync_test.go
```diff
@@ -214,40 +209,6 @@ func TestSyncPendingRangesWithMisses(t *testing.T) {
 	assert.Empty(t, syncer.pending.Head()) // assert all cache from pending is used
 }
 
-// Test that only one objective header is requested at a time
-func TestSyncer_OnlyOneRecentRequest(t *testing.T) {
```
Wondertan, line 218: This test test is not need anymore, as there is exactly the same and simpler one in sync_getter_test.go

libs/header/sync/sync_test.go
```diff
@@ -84,17 +83,15 @@ func TestDoSyncFullRangeFromExternalPeer(t *testing.T) {
 	// give store time to update heightSub index
 	time.Sleep(time.Millisecond * 100)
 
-	localHead, err := localStore.Head(ctx)
+	// trigger sync by calling Head
+	_, err = syncer.Head(ctx)
```
Wondertan, line 87: Yeah, Head triggers sync and we can avoid calling private method here

libs/header/sync/sync_head.go
```diff
@@ -10,160 +10,131 @@ import (
...
+// Otherwise, head is requested from a trusted peer and
+// set as the new subjective head, assuming that trusted peer is always fully synced.
 func (s *Syncer[H]) Head(ctx context.Context) (H, error) {
-	return s.networkHead(ctx)
-}
-
-// subjectiveHead returns the latest known local header that is not expired(within trusting period).
-// If the header is expired, it is retrieved from a trusted peer without validation;
-// in other words, an automatic subjective initialization is performed.
-func (s *Syncer[H]) subjectiveHead(ctx context.Context) (H, error) {
```
Wondertan, line 22: Don't be afraid of this red diff, it wasn't deleted. It was shortened and rearanged

libs/header/sync/sync_head.go
```diff
@@ -10,160 +10,131 @@ import (
...
 
-// Head returns the Syncer's latest known header. It calls 'networkHead' in order to
-// either return or eagerly fetch the most recent header.
+// Head returns the Network Head.
+//
+// Known subjective head is considered network head if it is recent enough(now-timestamp<=blocktime)
+// Otherwise, head is requested from a trusted peer and
+// set as the new subjective head, assuming that trusted peer is always fully synced.
 func (s *Syncer[H]) Head(ctx context.Context) (H, error) {
-	return s.networkHead(ctx)
```
Wondertan, line 16: No reason to have a private method for this

libs/header/sync/sync_head.go
```diff
@@ -10,160 +10,131 @@ import (
...
 			"hash", netHead.Hash().String(),
 			"err", err)
 	}
-	// try as new head
-	return s.newNetHead(ctx, netHead, false)
-}
 
-// newNetHead sets the network header as the new subjective head with preceding validation(per
-// request).
-func (s *Syncer[H]) newNetHead(ctx context.Context, netHead H, trust bool) pubsub.ValidationResult {
```
Wondertan, line 150: Found a way to avoid passing this trust bool

libs/header/sync/sync_head.go
```diff
@@ -10,160 +10,131 @@ import (
...
+	return
+}
+
+// incomingNetworkHead processes new gossiped network headers.
+func (s *Syncer[H]) incomingNetworkHead(ctx context.Context, netHead H) pubsub.ValidationResult {
+	// first of all, check the validity of the netHead
+	res := s.validateHead(ctx, netHead)
+	if res != pubsub.ValidationAccept {
+		return res
+	}
```
Wondertan, line 130: Now since validating every incoming header, instead of Store. So we always have one verification codepath that covers all the cases.

libs/header/sync/sync_head.go
```diff
@@ -10,160 +10,131 @@ import (
...
+	// and set it
+	s.setSubjectiveHead(ctx, netHead)
+	return res
 }
 
-// validate checks validity of the given header against the subjective head.
-func (s *Syncer[H]) validate(ctx context.Context, new H) pubsub.ValidationResult {
+// validateHead checks validity of the given header against the subjective head.
+func (s *Syncer[H]) validateHead(ctx context.Context, new H) pubsub.ValidationResult {
```
Wondertan, line 137: Just to be consistent with other methods that end on Head

libs/header/sync/sync_head.go
```diff
@@ -10,160 +10,131 @@ import (
...
-	if !trust {
-		if res := s.validate(ctx, netHead); res != pubsub.ValidationAccept {
-			// netHead was either ignored or rejected
-			return res
-		}
+	storeHead, err := s.store.Head(ctx)
+	if err == nil && storeHead.Height() >= netHead.Height() {
+		// we already synced it up - do nothing
+		return
 	}
```
Wondertan, line 116: Syncer now does an adjacency check itself instead of relying on store and does not trigger sync if so.

libs/header/sync/sync.go
```diff
@@ -147,7 +140,15 @@ func (s *Syncer[H]) State() State {
 	s.stateLk.RLock()
 	state := s.state
 	s.stateLk.RUnlock()
-	state.Height = s.store.Height()
+
+	head, err := s.store.Head(s.ctx)
```
Wondertan, line 144: Cannot use height for the same reason we need a syncStore

libs/header/sync/sync.go
```diff
@@ -171,43 +172,35 @@ func (s *Syncer[H]) syncLoop() {
 	}
 }
 
-// sync ensures we are synced from the Store's head up to the new subjective head
+// sync ensures we are synced from the Store's head up to the new subjective head.
 func (s *Syncer[H]) sync(ctx context.Context) {
-	newHead := s.pending.Head()
-	if newHead.IsZero() {
+	subjHead, err := s.subjectiveHead(ctx)
+	if err != nil {
```
Wondertan, line 178: Ensures we always have a valid header to sync to

libs/header/p2p/session.go
```diff
@@ -179,7 +179,7 @@ func (s *session[H]) doRequest(
 			return
 		case s.reqCh <- req:
 		}
-		log.Errorw("processing response", "err", err)
+		log.Errorw("processing response", "from", req.GetOrigin(), "to", req.Amount+req.GetOrigin()-1, "err", err)
```
Wondertan, line 182: Unrelated to this PR and can extract if reeeealy necessary

Wondertan (commented):
Self-review + clarifications for other reviewers. IDK why but I love this code so much that I can stare at it over and over

libs/header/p2p/session.go
vgonkivs: I'm fine w this

libs/header/sync/sync.go
```diff
@@ -247,9 +240,8 @@ func (s *Syncer[H]) doSync(ctx context.Context, fromHead, toHead H) (err error)
 	return err
 }
 
-// processHeaders gets and stores headers starting at the given 'from' height up to 'to' height -
-// [from:to]
-// processHeaders checks headers in pending cache that apply to the requested range.
+// processHeaders fetches and stores asked headers [from:to].
```
vgonkivs, line 243: for now, the range is (from; to]

libs/header/sync/sync_getter.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"context"
+	"sync"
+
+	"github.com/celestiaorg/celestia-node/libs/header"
+)
+
+// syncGetter is a Getter wrapper that ensure only one Head call happens at the time
+type syncGetter[H header.Header] struct {
```
vgonkivs, line 11: I would call it `syncHeadGetter` if you are expecting a single `Head` method. But I do not insist. Just a nit.

libs/header/sync/sync.go
```diff
@@ -16,51 +15,45 @@ var log = logging.Logger("header/sync")
...
+//   - Sets as the new Subjective Head, which
+//   - if there is a gap between the previous and the new Subjective Head
+//   - Triggers s.syncLoop and saves the Subjective Head in the pending so s.syncLoop can access it
 type Syncer[H header.Header] struct {
-	sub      header.Subscriber[H]
-	exchange header.Exchange[H]
-	store    header.Store[H]
+	sub     header.Subscriber[H]
+	store   syncStore[H]
+	getter  syncGetter[H]
```
vgonkivs, line 35: could you please mention that this getter is fetching the head from the network?

vgonkivs (changes requested):

libs/header/sync/sync.go
Wondertan: Hmm, we fetch everything from it, not only head

libs/header/sync/sync_getter.go
Wondertan: I didn't call it so because we still use it for other purposes besides Head. syncGetter just overrides one method

libs/header/sync/sync.go
Wondertan: Ok, I just kept it like it was before, but good catch

Wondertan: Fixed

libs/header/sync/sync.go
vgonkivs: Yes, right. I meant, that it can be confusing now to understand what is getter, what, and where it gets headers.

We used `header.Exchange`, that has a comment:

```
Exchange encompasses the behavior necessary to request Headers
from the network.
```

Wondertan: Ok, added

vgonkivs (approved):
Thanks for helping me with syncer ❤️

vgonkivs (dismissed):

libs/header/sync/sync.go
```diff
@@ -171,43 +172,35 @@ func (s *Syncer[H]) syncLoop() {
...
 }
 
-// sync ensures we are synced from the Store's head up to the new subjective head
+// sync ensures we are synced from the Store's head up to the new subjective head.
 func (s *Syncer[H]) sync(ctx context.Context) {
-	newHead := s.pending.Head()
-	if newHead.IsZero() {
+	subjHead, err := s.subjectiveHead(ctx)
+	if err != nil {
+		log.Errorw("getting new subjective head", "err", err)
```
renaynay, line 179: `getting subjective head`

libs/header/sync/sync.go
```diff
@@ -299,32 +292,32 @@ func (s *Syncer[H]) requestHeaders(
...
 		}
-		headers, err := s.exchange.GetVerifiedRange(ctx, fromHead, size)
+
+		headers, err := s.getter.GetVerifiedRange(ctx, fromHead, size)
 		if err != nil {
 			return err
 		}
-		amount -= size
+		// TODO(@Wondertan): This fixed syncing for me, but the len should always be = to size
+		//  Something fishy in here.
```
renaynay, line 301: 😱

renaynay: can you open issue for this at least

libs/header/sync/sync_getter.go
```diff
@@ -0,0 +1,31 @@
...
+
+	headLk  sync.RWMutex
+	headErr error
+	head    H
+}
+
+func (se *syncGetter[H]) Head(ctx context.Context) (H, error) {
+	// the lock construction here ensures only one routine calling Head at a time
+	// while others wait via Rlock
+	if !se.headLk.TryLock() {
```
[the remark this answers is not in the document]
Wondertan: What's the first one?

libs/header/sync/sync.go
Wondertan: I'll actually retry again if it's really the case, as things changed since then. Thanks for tagging this, as I completely forgot to double check.

libs/header/sync/sync_getter.go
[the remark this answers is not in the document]
renaynay: retriever.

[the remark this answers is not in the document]
Wondertan: Ah, right. In both cases it's a totally valid usage though. Even it's not recommended, there are use cases for which it was added in 1.18. 

If you now how to solve the problems in both cases without TryLock, I am also happy to listen. Also note that we had this code before and I just moved it.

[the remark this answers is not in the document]
renaynay: I mean the proper solution is to ensure from a codepath standpoint that two calls to exchange could not happen at once. But I'm okay with trylock for the moment bc that would require a larger refactoring.

libs/header/sync/sync.go
vgonkivs: Have you observed it on the network or in tests also?

libs/header/sync/sync.go
```diff
@@ -297,14 +297,23 @@ func (s *Syncer[H]) requestHeaders(
...
+				"got", headersLn,
+			)
+			log.Warn("PLEASE REPORT THIS AS A BUG")
+		}
 
 		if err := s.storeHeaders(ctx, headers...); err != nil {
 			return err
 		}
+
+		amount -= headersLn
```
renaynay, line 314: let's keep this at size then?

renaynay: unless it's expected that this occurs in which case, shouldn't we just return err?

libs/header/sync/sync_getter.go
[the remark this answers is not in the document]
Wondertan: I can't entirely agree here. We have one single code path that solves all the problems to determine the current Head and stuff, as the logic is exactly the same if we do it locally in syncer or if someone else triggers syncer.Head that triggers syncGetter.Head. Furthermore, independently of codepaths or not, the event when Syncer needs to request a new header and someone called syncer.Head may happen causing to request at once.  

What I can agree on though is that there is a better solution that should live in the exchange itself that guarantees that only the head is requested at the time, but that solution doesn't include refactoring syncing logic and just removing sync getter

libs/header/sync/sync.go
Wondertan: I retried, and on the second attempt syncing broke. It went requesting things beyond network head
image

vgonkivs: Yes. The same for me. And the root cause is inside `ranges`. For some reason, it stores already handled header. I run node on your branch with extra logs and it shows that:
```
// finished requesting the range
2023-03-14T15:36:40.006+0200	DEBUG	header/p2p	p2p/session.go:109	received headers range	{"from": 212994, "to": 213336}
```

```
// applied cached head
2023-03-14T15:36:40.050+0200	INFO	header/store	store/store.go:349	new head	{"height": 213337, "hash": "566CE3183AD172A408FEA9FA7FE60648C8BEF6E153266EFB8913CF69FE7BFB2D"}
```

```
// and then got the same header from the cache after 
		headers := headersRange.Truncate(to) // amount=1 // 3066
		if len(headers) == 0 {
			break
		}
```
 
`{"header":{"version":{"block":"11"},"chain_id":"arabica-6","height":"213337","time":"2023-03-14T13:34:12.03287407Z"`

libs/header/sync/sync.go
```diff
@@ -297,14 +297,23 @@ func (s *Syncer[H]) requestHeaders(
 		if err != nil {
 			return err
 		}
-		// TODO(@Wondertan): This fixed syncing for me, but the len should always be = to size
-		//  Something fishy in here.
-		amount -= uint64(len(headers))
+
+		headersLn := uint64(len(headers))
+		if headersLn != size {
+			log.Warnw("verified range != requested range",
```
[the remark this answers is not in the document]
vgonkivs: tested multiple times(syncing node from 2 to head). `headersLn == size`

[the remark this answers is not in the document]
vgonkivs: Exchange can't return the partial range.

libs/header/sync/sync.go
Wondertan: We found the root cause and its now fixed. It could happen in case we receive a duplicate from pubsub

libs/header/sync/sync.go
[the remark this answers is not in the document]
Wondertan: I removed this

libs/header/sync/sync.go
Wondertan: Returned to size

libs/header/sync/sync.go
Wondertan: Even though I had this case in mind while making this PR, I totally forgot that we truncated the pending just right when we started syncing, and @vgonkivs proposed to split Truncate into Get and Remove, so we clean up pending only after processing a range finished and not before. This way, pending will keep the sync target long enough, and duplicates will be detected/ignored correctly.

vgonkivs (dismissed):

renaynay (dismissed):

libs/header/sync/sync_getter.go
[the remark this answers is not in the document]
Wondertan: The protocol itself should be responsible for not doing the same requests, not the user of the protocol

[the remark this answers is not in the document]
renaynay: can you open an issue so we can track please I don't quite agree yet.

renaynay (approved):

vgonkivs (approved):

libs/header/sync/sync_getter.go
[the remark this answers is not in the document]
Wondertan: https://github.com/celestiaorg/celestia-node/issues/1914

PR #1895, opened 2023-03-13
test(nodebuilder): TestLifecycle also checks that empty block is stored

renaynay:
Small additional check in lifecycle.

### DIFF
--- a/nodebuilder/node_test.go
+++ b/nodebuilder/node_test.go
@@ -14,6 +14,7 @@ import (
 	"google.golang.org/protobuf/proto"
 
 	"github.com/celestiaorg/celestia-node/nodebuilder/node"
+	"github.com/celestiaorg/celestia-node/share"
 )
 
 func TestLifecycle(t *testing.T) {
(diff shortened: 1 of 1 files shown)


nodebuilder/node_test.go
```diff
@@ -45,6 +46,10 @@ func TestLifecycle(t *testing.T) {
 			// ensure the state service is running
 			require.False(t, node.StateServ.IsStopped(ctx))
 
+			// ensure an empty block exists in store
+			err = node.ShareServ.SharesAvailable(ctx, share.EmptyRoot())
+			require.NoError(t, err)
+
```
Wondertan, line 52: This check is unrelated to the test. There should be a different one

Wondertan (changes requested):

nodebuilder/node_test.go
renaynay: A different test for this check you mean? That's fair

Wondertan (approved):
Rename the PR and lfg

walldiss (approved):

distractedm1nd (approved):

PR #1897, opened 2023-03-13
fix(share): Fix empty EDS / empty root initialisation

renaynay:
We did not generate the empty EDS correctly - moving forward, it's best that celestia-app `pkg/da` contains a convenience function for generating empty shares that we can rely on instead of duplicating logic here. I will do it now, but for now we need this solution b/c it's a bug.


Wondertan (dismissed):

rootulp:
>  it's best that celestia-app pkg/da contains a convenience function for generating empty shares

Is [TailPaddingShares](https://github.com/celestiaorg/celestia-app/blob/60fde95de97b214cddf44e75b0205099213fcfc2/pkg/shares/padding.go#L63-L66) sufficient? I see it's used in this PR but I wonder if you want anything else exposed from celestia-app?

walldiss (dismissed):

walldiss (approved):

Wondertan:
@rootulp, I think Rene wants the utility that returns empty eds.

Wondertan (approved):

PR #1900, opened 2023-03-13
WIP -> fix: Go ldflags inside Dockerfile

jrmanes:
## Overview

Hello team!

Small change to the `Dockerfiles` and the way that we build the images.

Until now, we are building the `go` service using the `Makefile`, but seems like for any reason, the `ldflags` are not being generated properly and we are not able to identify what version are we running once we deploy the container...

Screenshot 2023-03-13 at 18 05 24

I've check some version (until 0.6.0), and they have the same issue, so it's not new.

I have made these changes and test them, after build the container in this way, everything seems to be ok:

The build pipeline works ok:

You can check the logs: [here](https://github.com/jrmanes/celestia-node/actions/runs/4407452652)

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,14 @@ COPY . .
 ARG TARGETOS TARGETARCH
 ENV GOOS=$TARGETOS
 ENV GOARCH=$TARGETARCH
-RUN make build && make cel-key
+
+RUN echo "--> Building celestia" &&\
+	go build -o build/ \
+	-ldflags="-X 'main.buildTime=$(date)' -X 'main.lastCommit=$(git rev-parse HEAD)' -X 'main.semanticVersion=$(git describe --tags)'" \
+	./cmd/celestia
+
+RUN echo "--> Building cel-key" &&\
+	go build -o build/ ./cmd/cel-key
 
 FROM ubuntu:20.04
 # Default node type can be overwritten in deployment manifest
@@ -18,7 +25,7 @@ COPY docker/entrypoint.sh /
 
 # Copy in the binary
 COPY --from=builder /src/build/celestia /
-COPY --from=builder /src/./cel-key /
+COPY --from=builder /src/build/cel-key /
 
 EXPOSE 2121
 


smuu (approved):
LGTM

Makefile
```diff
@@ -1,5 +1,6 @@
 SHELL=/usr/bin/env bash
 PROJECTNAME=$(shell basename "$(PWD)")
+# Remember: if you update the LDFLAGS, please do it also in the Dockerfile
```
Wondertan, line 3: ❤️

Wondertan: Or actually can we make a single build script or debug why makefile does not work?

Wondertan (commented):
Linter complains

Makefile
jrmanes: I have tried to make it work in a different ways, but looks like the `make` command is not reporting the flags correctly inside the container, but it works when we execute outside (in our localhost)

I don't rule out the option to have an script (we can create a `scripts` folder) and try to use it from the `Dockerfile` as well, instead of using the `Makefile`, WDYT? 🤔

MSevey: I think the problem is with the `build` command and other places we are using the ldflags variable. 

I think it needs to be `-ldflags=${LDFLAGS}` for how we are defining things.

Ref: https://stackoverflow.com/questions/68100849/ldflags-is-ignored-when-building-go-project

Based on the stack overflow, we might even consider updating the `LDFLAGS` variable definition too so that all commands that need the `ldflags` can just reference the variable and don't need to even include the flag and `=`.

MSevey (changes requested):
The bug is in the Makefile, please revert dockerfile changes and update the makefile and test.

Makefile
MSevey: Looks like we are doing something similar with app, so we probably need to test the ldflags there as well and verify if we need to update this line https://github.com/celestiaorg/celestia-app/blob/main/Makefile#L19

Assuming my suggestion fixes the issue can you follow up on app @jrmanes ?

cc @evan-forbes

jrmanes: hello @MSevey 

good catch, I tested it and seems like this was the issue!

I kept the whole flag inside the variable for using only the variable `LDFLAGS`

Bidon15 (approved):

Makefile
Wondertan: Nice one. Glad that we avoiding copying.

Wondertan (approved):
Thank you

vgonkivs (approved):

PR #1901, opened 2023-03-13
feat(nodebuilder/node | api): Version the API

### LINKED ISSUE #1345: rpc: decide on versioning
We need to make a decision on how we want to version our RPC. In the future we could provide multiple versions at multiple endpoints. At the very least, it will be a failsafe for someone accidentally trying to use an old version of our api

We can then provide the first RPC at `/v1` for example

Also: return it as part of `node.Info`

renaynay:
Introduce a version string to the API

Resolves #1345

### DIFF
--- a/api/docgen/openrpc.go
+++ b/api/docgen/openrpc.go
@@ -18,7 +18,7 @@ import (
 )
 
 const (
-	APIVersion     = "v0.0.1"
+	APIVersion     = "v0.1.0"
 	APIDescription = "The Celestia Node API is the collection of RPC methods that " +
 		"can be used to interact with the services provided by Celestia Data Availability Nodes."
 	APIName  = "Celestia Node API"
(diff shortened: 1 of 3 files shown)


renaynay:
Kinda hesitant to merge this til we have `header.Subscription` endpoint - but who knows how long gomock will take to impl support for generics.

distractedm1nd (changes requested):
Please check `api/docgen/openrpc.go` line 23 - there is a const there that declares the api version, so we should combine these

renaynay:
Marking ready for review as API is not yet stable so let's just make it first minor version.

distractedm1nd (approved):

distractedm1nd (approved):

Wondertan (approved):

vgonkivs (approved):

PR #1903, opened 2023-03-14
bugfix(libs/header/p2p): shuffle trusted peers before sending a request

vgonkivs:
## Overview
We were using a weak random number generator that does not guarantee that we will get a really random peer within multiple attempts and we will request the same peer each time.

## Checklist

### DIFF
--- a/libs/header/p2p/exchange.go
+++ b/libs/header/p2p/exchange.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"math/rand"
 	"sort"
+	"time"
 
 	logging "github.com/ipfs/go-log/v2"
 	"github.com/libp2p/go-libp2p/core/host"
@@ -235,18 +236,21 @@ func (ex *Exchange[H]) performRequest(
 		return nil, fmt.Errorf("no trusted peers")
 	}
 	var reqErr error
-	for i := 0; i < len(ex.trustedPeers); i++ {
-		//nolint:gosec // G404: Use of weak random number generator
-		idx := rand.Intn(len(ex.trustedPeers))
+	peers := ex.trustedPeers
 
+	// we need this each time before calling shuffle
+	rand.Seed(time.Now().UnixNano())
+	// change order to get a random peer
+	rand.Shuffle(len(peers), func(i, j int) { peers[i], peers[j] = peers[j], peers[i] })
+	for _, peer := range peers {
 		ctx, cancel := context.WithTimeout(ctx, ex.Params.TrustedPeersRequestTimeout)
-		h, err := ex.request(ctx, ex.trustedPeers[idx], req)
+		h, err := ex.request(ctx, peer, req)
 		cancel()
 		switch err {
 		default:
 			reqErr = err
 			log.Debugw("requesting header from trustedPeer failed",
-				"trustedPeer", ex.trustedPeers[idx], "err", err)
+				"trustedPeer", peer, "err", err)
 			continue
 		case context.Canceled, context.DeadlineExceeded, nil:
 			return h, err


libs/header/p2p/exchange.go
```diff
@@ -235,18 +236,21 @@ func (ex *Exchange[H]) performRequest(
 		return nil, fmt.Errorf("no trusted peers")
 	}
 	var reqErr error
-	for i := 0; i < len(ex.trustedPeers); i++ {
-		//nolint:gosec // G404: Use of weak random number generator
-		idx := rand.Intn(len(ex.trustedPeers))
+	peers := ex.trustedPeers
 
+	// we need this each time before calling shuffle
+	rand.Seed(time.Now().UnixNano())
```
Wondertan, line 242: Why do we need it each time? The seed should be populated once at the start of the app. The results of the shuffle will be different

Wondertan (commented):
It's not worth it. We gain nothing from cryptographic randomness here while complementing code

vgonkivs:
It's actually worth it. We are getting the same index each time in case of failed response, so we don't have a chance to request another peer one of them goes offline rn

libs/header/p2p/exchange.go
walldiss, line 242: According to `Seed`. godoc, this will Seed global rand var. It is also deprecated since 1.20 (I know we are not there yet, but soon).  For this case we can use local random sourcer by calling:
```
rand.New(rand.NewSource(time.Now().UnixNano())).Shuffle( ... )
```

vgonkivs:
```
// Seed uses the provided seed value to initialize the default Source to a
// deterministic state. If Seed is not called, the generator behaves as
// if seeded by Seed(1). Seed values that have the same remainder when
// divided by 2³¹-1 generate the same pseudo-random sequence.
// Seed, unlike the Rand.Seed method, is safe for concurrent use.
```

Wondertan: >  If Seed is not called, the generator behaves as if seeded by Seed(1)

I didn't say it should not be called, but I said it should be called only once when the node starts.

Wondertan: But this is the comment in case we go with global randomness. The @walldiss's suggestion to create an instance each time as it soon be deprecated is also valid

distractedm1nd (approved):

Wondertan (commented):
Lint

walldiss (approved):

libs/header/p2p/exchange.go
```diff
@@ -235,18 +236,24 @@ func (ex *Exchange[H]) performRequest(
 		return nil, fmt.Errorf("no trusted peers")
 	}
 	var reqErr error
-	for i := 0; i < len(ex.trustedPeers); i++ {
-		//nolint:gosec // G404: Use of weak random number generator
-		idx := rand.Intn(len(ex.trustedPeers))
+	peers := ex.trustedPeers
```
Wondertan, line 239: This is racy because slices are copied by pointer not by value. Shuffling peers shuffles them everywhere, while there is no synchronisation around Head request. We actually should allow only one Head request at a time.

libs/header/p2p/exchange.go
```diff
@@ -66,14 +69,23 @@ func NewExchange[H header.Header](
...
 		),
 		Params: params,
-	}, nil
+	}
+
+	ex.trustedPeers = func() peer.IDSlice {
+		tpeers := make(peer.IDSlice, len(peers))
+		copy(tpeers, peers)
+		return shuffledPeers(tpeers)
+	}
```
Wondertan, line 78: We don't usually do such things in our code, but it's fine in this case

Wondertan (approved):

Wondertan (approved):

renaynay (approved):
LGTM

PR #1912, opened 2023-03-16
refactor(nodebuilder/das): update default DASer setting 

walldiss:
## Overview

DASer default settings are based on IPLD only getters measurements. With introduction of blocksync to full nodes as main sampling protocol, value should to be revised. Shrex has much less roundtrips to obtain data, thus it is much better at utilising network bandwidth with fewer workers. Default DASer setting are now updated with relation to node type.

### DIFF
--- a/nodebuilder/config.go
+++ b/nodebuilder/config.go
@@ -50,7 +50,7 @@ func DefaultConfig(tp node.Type) *Config {
 	case node.Bridge:
 		return commonConfig
 	case node.Light, node.Full:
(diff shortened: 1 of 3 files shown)


Wondertan (approved):
Great as discussed.

renaynay (approved):
needs lint

distractedm1nd (approved):

distractedm1nd (approved):

renaynay (approved):

vgonkivs (approved):

PR #1919, opened 2023-03-16
handle error when failing to load config

cmwaters:
I'm guessing this is ignored because if this fails, the node will just use the default config but perhaps you want to distinguish between the config not existing and it existing but failing to decode


cmd/flags_node.go
```diff
@@ -68,9 +69,12 @@ func ParseNodeFlags(ctx context.Context, cmd *cobra.Command, network p2p.Network
...
 		}
 		cfg, err := nodebuilder.LoadConfig(filepath.Join(expanded, "config.toml"))
-		if err == nil {
-			ctx = WithNodeConfig(ctx, cfg)
+		if err != nil {
+			if !errors.Is(err, os.ErrNotExist) {
+				return ctx, err
+			}
 		}
+		ctx = WithNodeConfig(ctx, cfg)
```
renaynay, line 77: this will panic if os.ErrNotExist

Let's instead do this: 

if os.ErrNotExist, we should `cfg = nodebuilder.DefaultConfig()` and also `log.Warn` that no config was found in given path(log out the path), initialising with default config

renaynay (changes requested):
Where did you run into this issue? I'm assuming it's in the case that a user has removed their config but not their node store and they're trying to run their node over a node store that doesn't contain a config. 

Regardless, it's a weird edge case and I'm fine with adding this err check, just refer to comment I left there.

cmd/flags_node.go
cmwaters: Yup makes sense

renaynay (commented):
@cmwaters did you run into an issue where the error was not handled? This doesn't really resolve the core issue which I'm assuming is - if you don't have a config present on start, that it initialises a default config for you and uses that. For that to work, you'd need to also write the config to the node store.

renaynay:
Closing after sync discussion w @cmwaters

PR #1921, opened 2023-03-17
(api/docgen): Example types

### LINKED ISSUE #1358: docs(rpc+gateway): Finish adding example types

renaynay:
Replaces #1408, closes #1358


distractedm1nd (dismissed):

distractedm1nd (approved):

distractedm1nd (approved):

walldiss (approved):
LGTM.

If developer changes the returned object from API, that has an example, what would be the easiest way to generate json?

distractedm1nd:
@walldiss The easiest way to generate the JSON is to just query the endpoint again using the RPC. For smaller types, creating type literals is the best option, because it will be immediately clear when you edit the type

PR #1922, opened 2023-03-17
fix(header/sync): add drift to recency detection

Wondertan:
Pretty often, when you start a node for the first time, there is a warning complaining about an unsynced trusted peer, while we know that the peer is 100% synced. This warning also means we make an additional Head request for a bad reason. The fix is introducing a half-block time drift. I've tested it like 20 times and wasn't able to reproduce the warning. We also don't need to be super precise about head to worry about this additional half. The main goal here is to get the head quick enough after being a day or more out of sync.


renaynay (approved):

renaynay:
Nice thanks for this

distractedm1nd (approved):

walldiss (approved):

PR #1924, opened 2023-03-17
fear(shrex/peer-manager): add metrics to peer manager

walldiss:
## Overview
Introduce metrics to peer manager.

### DIFF
--- a/share/p2p/peers/manager.go
+++ b/share/p2p/peers/manager.go
@@ -23,16 +23,16 @@ import (
 )
 
 const (
-	// ResultSuccess indicates operation was successful and no extra action is required
-	ResultSuccess result = iota
+	// ResultNoop indicates operation was successful and no extra action is required
+	ResultNoop result = "result_noop"
 	// ResultSynced will save the status of pool as "synced" and will remove peers from it
-	ResultSynced
+	ResultSynced = "result_synced"
 	// ResultCooldownPeer will put returned peer on cooldown, meaning it won't be available by Peer
 	// method for some time
-	ResultCooldownPeer
+	ResultCooldownPeer = "result_cooldown_peer"
 	// ResultBlacklistPeer will blacklist peer. Blacklisted peers will be disconnected and blocked from
 	// any p2p communication in future by libp2p Gater
-	ResultBlacklistPeer
+	ResultBlacklistPeer = "result_blacklist_peer"
 )
 
 var log = logging.Logger("shrex/peer-manager")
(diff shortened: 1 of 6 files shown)


walldiss:
I've used some features from go1.20 need to rollback those or wait for upgrade of golang in node.
Putting in draft until resolved

share/p2p/peers/pool.go
```diff
@@ -194,3 +194,9 @@ func (p *pool) checkHasPeers() {
 		p.hasPeer = false
 	}
 }
+
+func (p *pool) len() int {
+	p.m.Lock()
+	defer p.m.Unlock()
```
renaynay, line 200: rlock

share/p2p/peers/metrics.go
```diff
@@ -0,0 +1,275 @@
...
+		return res
+	}
+}
+
+// observeBlacklistPeers stores amount of blacklisted peers by reason
+func (m *metrics) observeBlacklistPeers(reason blacklistPeerReason, amount int) {
+	if m == nil {
+		return
+	}
+	for {
```
renaynay, line 239: is there some sort of cancellation for this?

share/p2p/peers/manager.go
```diff
@@ -24,16 +24,16 @@ import (
 )
 
 const (
-	// ResultSuccess indicates operation was successful and no extra action is required
-	ResultSuccess result = iota
+	// ResultNoop indicates operation was successful and no extra action is required
+	ResultNoop result = "result_noop"
```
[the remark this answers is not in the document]
walldiss: It was exported before and exported now. Naming changed because it is unclear from caller perspective  how ResultSuccess is different from ResultSynced.

share/p2p/peers/metrics.go
```diff
@@ -0,0 +1,275 @@
...
+	sourceFullNodes peerSource = "full_nodes"
+
+	blacklistPeerReasonKey                     = "blacklist_reason"
+	reasonInvalidHash      blacklistPeerReason = "invalid_hash"
+	reasonMisbehave        blacklistPeerReason = "misbehave"
+
+	validationResultKey = "validation_result"
+	validationAccept    = "accept"
+	validationReject    = "reject"
+	validationIgnore    = "ignore"
```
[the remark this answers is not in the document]
walldiss: wdym? if you are talking about pubsub consts - they are iota and are not suited for metrics.

share/p2p/peers/metrics.go
walldiss: it will return on successful update

distractedm1nd (approved):

share/p2p/peers/metrics.go
renaynay: why does it have to be in a loop?

renaynay (approved):

share/p2p/peers/metrics.go
walldiss: If the value is updated after read and before CAS, it will not get updated. This is concurrency CAS protection

Wondertan (approved):
Did not go over each metrics but it looks good and correct

share/p2p/peers/metrics.go
renaynay: Alright.

renaynay (approved):

Issue #1917, opened 2023-03-16
(shrex/peer-manager): hash validation on subjective node start


walldiss:
### Implementation ideas

When full node started on already existing network, it has a chance to receive `shrex.Sub` message about hashes that were recent, but not the latest network head (networkhead - 1, networkhead -2 ...). In this case peer-manager will create invalidated pools for those header hashes and will wait for validation. If node is lagging behind network head a lot and won't be able to reach those headers, peer-manager might make a false positive blacklist decision. It would block those hashes with all peers for the remaining node run time.

Wondertan:
Or we can get the header by datahash somehow

renaynay:
Would rather not index, would rather extend message by one field.

Wondertan:
We also agreed that indexing wont work without backward sync

PR #1945, opened 2023-03-20
fix(shrex/peer-manager)!: hash validation on subjective node start

walldiss:
## Overview

Resolves 
https://github.com/celestiaorg/celestia-node/issues/1917


share/p2p/peers/manager.go
```diff
@@ -348,11 +365,14 @@ func (m *Manager) cleanUp() []peer.ID {
 	addToBlackList := make(map[peer.ID]struct{})
 	for h, p := range m.pools {
 		if time.Since(p.createdAt) > m.poolValidationTimeout && !p.isValidatedDataHash.Load() {
+			delete(m.pools, h)
+			if p.headerHeight < m.firstHeader {
+				continue
+			}
```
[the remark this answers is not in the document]
walldiss: Yes. This code handles the corner case when hash is received from shrex.Sub before any headers from headerSub. Peer manager will create pool for such hash and if it will not get validated, we will just remove it. Note that after first header is received from headerSub such pool will no longer be created, messages will be ignored on Height check inside validator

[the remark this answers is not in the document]
walldiss: There are 2 test cases added to target different order of Hashes and headerSub.

share/p2p/shrexsub/pubsub.go
```diff
@@ -94,9 +94,9 @@ func (s *PubSub) AddValidator(v ValidatorFn) error {
 }
 
 func (v ValidatorFn) validate(ctx context.Context, p peer.ID, msg *pubsub.Message) pubsub.ValidationResult {
-	var pbmsg pb.EDSAvailableMessage
+	var pbmsg pb.RecentEDSNotification
 	if err := pbmsg.Unmarshal(msg.Data); err != nil {
-		log.Debugw("shrex-push: unmarshal error", "err", err)
+		log.Debugw("validator: unmarshal error", "err", err)
```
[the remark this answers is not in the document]
walldiss: It will be prefixed by logger name "shrex-sub" anyway. I have added "validator" prefix to distinguish this error from another unmarshal error inside subscription

renaynay (commented):
might be a tiny bit outdated but posting anyway

share/p2p/peers/manager.go
```diff
@@ -286,8 +302,9 @@ func (m *Manager) getOrCreatePool(datahash string) *syncPool {
 	p, ok := m.pools[datahash]
 	if !ok {
 		p = &syncPool{
-			pool:      newPool(m.peerCooldownTime),
-			createdAt: time.Now(),
+			pool:         newPool(m.peerCooldownTime),
+			createdAt:    time.Now(),
+			headerHeight: 1, // headerHeight needs to be > 0 for pool to be not Validated
```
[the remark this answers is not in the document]
walldiss: Ok ♻️

share/p2p/shrexsub/subscription.go
```diff
@@ -25,15 +26,23 @@ func newSubscription(t *pubsub.Topic) (*Subscription, error) {
 
 // Next blocks the caller until any new EDS DataHash notification arrives.
 // Returns only notifications which successfully pass validation.
-func (subs *Subscription) Next(ctx context.Context) (share.DataHash, error) {
+func (subs *Subscription) Next(ctx context.Context) (Notification, error) {
```
vgonkivs, line 29: why not `*Notification`?

share/p2p/peers/manager.go
```diff
@@ -322,37 +342,59 @@ func (m *Manager) hashIsBlacklisted(hash share.DataHash) bool {
...
 
 	addToBlackList := make(map[peer.ID]struct{})
 	for h, p := range m.pools {
-		if time.Since(p.createdAt) > m.poolValidationTimeout && !p.isValidatedDataHash.Load() {
+		fmt.Println("RUN", !p.isValidatedDataHash.Load(), time.Since(p.createdAt) > m.poolValidationTimeout)
+		if !p.isValidatedDataHash.Load() && time.Since(p.createdAt) > m.poolValidationTimeout {
+			delete(m.pools, h)
+			if p.headerHeight < m.initialHeight {
+				// outdated pools could still be valid even if not validated, no need to blacklist
+				fmt.Println("SKIP")
```
vgonkivs, line 390: ?? 😄

share/p2p/peers/manager.go
```diff
@@ -322,37 +342,59 @@ func (m *Manager) hashIsBlacklisted(hash share.DataHash) bool {
...
+				continue
+			}
+			fmt.Println("BLACKLIST")
```
vgonkivs, line 393: 😄

share/p2p/peers/manager.go
```diff
@@ -243,39 +248,54 @@ func (m *Manager) subscribeHeader(ctx context.Context, headerSub libhead.Subscri
...
+	if m.isBlacklistedPeer(peerID) {
+		log.Debugw("received message from blacklisted peer, reject validation",
+			"peer", peerID,
+			"datahash", msg.DataHash.String())
 		return pubsub.ValidationReject
 	}
 
-	m.getOrCreatePool(hash.String()).add(peerID)
-	log.Debugw("got hash from shrex-sub", "peer", peerID, "datahash", hash.String())
+	if msg.Height <= 0 {
```
vgonkivs, line 281: you can avoid checking `<0`, bc the Height type, is uint64

share/p2p/peers/manager.go
```diff
@@ -322,37 +342,59 @@ func (m *Manager) hashIsBlacklisted(hash share.DataHash) bool {
 	return m.blacklistedHashes[hash.String()]
 }
 
+func (m *Manager) validatedPool(hashStr string) *syncPool {
+	p := m.getOrCreatePool(hashStr)
+	if p.isValidatedDataHash.CompareAndSwap(false, true) {
+		fmt.Println("validate")
```
vgonkivs, line 348: should be removed

share/p2p/peers/manager.go
```diff
@@ -322,37 +342,59 @@ func (m *Manager) hashIsBlacklisted(hash share.DataHash) bool {
...
+		return nil
+	}
+
 	m.lock.Lock()
 	defer m.lock.Unlock()
 
 	addToBlackList := make(map[peer.ID]struct{})
 	for h, p := range m.pools {
-		if time.Since(p.createdAt) > m.poolValidationTimeout && !p.isValidatedDataHash.Load() {
+		fmt.Println("RUN", !p.isValidatedDataHash.Load(), time.Since(p.createdAt) > m.poolValidationTimeout)
```
vgonkivs, line 385: and this one

vgonkivs (changes requested):

share/p2p/peers/manager.go
walldiss: woops 😂

distractedm1nd (dismissed):

share/p2p/peers/manager.go
```diff
@@ -322,37 +342,54 @@ func (m *Manager) hashIsBlacklisted(hash share.DataHash) bool {
 	return m.blacklistedHashes[hash.String()]
 }
 
+func (m *Manager) validatedPool(hashStr string) *syncPool {
```
renaynay, line 345: can we call this `validatePool` ?

share/p2p/peers/manager.go
```diff
@@ -243,39 +248,54 @@ func (m *Manager) subscribeHeader(ctx context.Context, headerSub libhead.Subscri
...
 	if peerID == m.host.ID() {
-		log.Debugw("received datahash from self", "datahash", hash.String())
+		log.Debugw("received datahash from self", "datahash", msg.DataHash.String())
 		return pubsub.ValidationAccept
 	}
 
 	// punish peer for sending invalid hash if it has misbehaved in the past
-	if m.hashIsBlacklisted(hash) {
-		log.Debugw("received blacklisted hash, reject validation", "peer", peerID, "datahash", hash.String())
+	if m.hashIsBlacklisted(msg.DataHash) {
```
renaynay, line 269: also let's rename `isBlacklistedHash` ?

share/p2p/shrexsub/pubsub.go
```diff
@@ -80,29 +87,52 @@ func (s *PubSub) Stop(context.Context) error {
...
-func (s *PubSub) AddValidator(validate Validator) error {
-	return s.pubSub.RegisterTopicValidator(s.pubsubTopic,
-		func(ctx context.Context, p peer.ID, msg *pubsub.Message) pubsub.ValidationResult {
-			return validate(ctx, p, msg.Data)
-		})
+func (s *PubSub) AddValidator(v ValidatorFn) error {
+	return s.pubSub.RegisterTopicValidator(s.pubsubTopic, v.validate)
+}
+
+func (v ValidatorFn) validate(ctx context.Context, p peer.ID, msg *pubsub.Message) pubsub.ValidationResult {
```
renaynay, line 96: I don't understand this - why can't we have this as a standalone function? why is it a method on `ValidatorFn` type?

renaynay: Can we unify all this logic into one validation func?

share/p2p/peers/manager.go
```diff
@@ -48,7 +48,10 @@ type Manager struct {
 	connGater *conngater.BasicConnectionGater
 
 	// pools collecting peers from shrexSub
-	pools                 map[string]*syncPool
+	pools map[string]*syncPool
+	// messages from shrex.Sub with height below initialHeight will be ignored, since we don't need to
+	// track peers for those headers
+	initialHeight         uint64
```
Wondertan, line 54: atomic.Uint64

share/p2p/peers/manager.go
```diff
@@ -243,39 +248,54 @@ func (m *Manager) subscribeHeader(ctx context.Context, headerSub libhead.Subscri
...
+		return pubsub.ValidationReject
+	}
+
+	if msg.Height < atomic.LoadUint64(&m.initialHeight) {
+		// we can use peers from discovery for headers before the first one from headerSub
+		// if we allow pool creation for those headers, there is chance the pool will not be validated in
+		// time and will be false-positively trigger blacklisting of hash and all peers that sent msgs for
+		// that hash
+		log.Debugw("received message for past header", "peer", peerID, "datahash", msg.DataHash.String())
+		return pubsub.ValidationIgnore
```
Wondertan, line 292: We should do hard reject for most heights unless it's recent. At this point, it does not matter as for FloodSub the Ignore and the Reject are effectively the same, but it's likely that we will move to a different router at some point. It's also just nice to codify the exact rules, what is hard reject, and what is acceptable via ignore

share/p2p/peers/manager.go
Wondertan: The main purpose is to get validated pool here and a side affect is making it validated if not

share/p2p/shrexsub/pubsub.go
walldiss: This function handles wire msg decoding and validation and then passes `Notification` to ValidatorFn (line 111) to get validation result. it could be a free function, but still would need ValidatorFn as an input. We can't make it one validation, because Validator is passed from external pkg.  Maybe you have ideas for better naming?

share/p2p/peers/manager.go
```diff
@@ -130,14 +138,14 @@ func (m *Manager) Start(startCtx context.Context) error {
...
 
-	err := m.shrexSub.Start(startCtx)
+	err := m.shrexSub.AddValidator(m.validate)
 	if err != nil {
-		return fmt.Errorf("starting shrexsub: %w", err)
+		return fmt.Errorf("registering validator: %w", err)
 	}
 
-	err = m.shrexSub.AddValidator(m.validate)
+	err = m.shrexSub.Start(startCtx)
```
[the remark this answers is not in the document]
walldiss: I've not seen issues in this part, but the order of operations could lead some messages skipping the peer-manager validation. It is nice to have correct order here.

share/p2p/peers/manager.go
walldiss: In case of messages handled by this rule, we we don't target any peer scoring or filtering mechanics, just want to address corner case of subjective start and doing as little logic as possible. The reason why `pubsub.ValidationIgnore` is used here is because there is no strong evidence of any misbehave from peer and thus no punishment needed.

share/p2p/peers/manager.go
```diff
@@ -376,12 +413,12 @@ func (p *syncPool) markSynced() {
...
-
 func (p *syncPool) add(peers ...peer.ID) {
 	if !p.isSynced.Load() {
 		p.pool.add(peers...)
 	}
 }
+
+func (p *syncPool) storeHeight(h uint64) {
+	p.headerHeight.Store(h)
+}
```
Wondertan, line 424: optional nit: there is no point imo to have this method and can be inlined

Wondertan (approved):
LGTM

share/p2p/shrexsub/pb/notification.proto
```diff
@@ -0,0 +1,9 @@
+syntax = "proto3";
+
+package share.p2p.shrex.sub;
+
+message RecentEDSNotification {
```
Wondertan, line 5: optional: Inconsistent naming with Notification go type

Wondertan (approved):

Wondertan (approved):

distractedm1nd (approved):

vgonkivs (approved):

Issue #1640, opened 2023-01-26
improvement(core/fetcher): Listen for `EventDataSignedBlock` to eliminate requests for redundant block info


renaynay:
Switch to using this endpoint to reduce redundant RPC requests to core.

Ref https://github.com/celestiaorg/celestia-core/pull/940

This change will likely land in v.0.12.0 app release

Wondertan:
Pls see the discussion #1 before continuing

PR #1948, opened 2023-03-21
feat(core/listener): listen for EventDataSignedBlock

walldiss:
## Overview

Resolves https://github.com/celestiaorg/celestia-node/issues/1640

### DIFF
--- a/core/fetcher.go
+++ b/core/fetcher.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	logging "github.com/ipfs/go-log/v2"
+	coretypes "github.com/tendermint/tendermint/rpc/core/types"
 	"github.com/tendermint/tendermint/types"
 
 	libhead "github.com/celestiaorg/celestia-node/libs/header"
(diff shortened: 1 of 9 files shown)


core/fetcher.go
```diff
@@ -78,6 +79,11 @@ func (f *BlockFetcher) GetBlockByHash(ctx context.Context, hash libhead.Hash) (*
 	return res.Block, nil
 }
 
+// GetSignedBlock queries Core for a `Block` at the given height.
+func (f *BlockFetcher) GetSignedBlock(ctx context.Context, height *int64) (*coretypes.ResultSignedBlock, error) {
```
renaynay, line 83: why make this change in this PR? don't you only need it for core exchange PR?

renaynay (commented):
otherwise lgtm

core/fetcher.go
walldiss: TestBlockFetcher_GetBlock_and_SubscribeNewBlockEvent is designed around checking results from sub against rpc call. I would either comment it out in this PR or combine both in one

renaynay: ack

Wondertan (approved):

vgonkivs (approved):

PR #1950, opened 2023-03-22
fix(share): enabling discovery on light nodes

distractedm1nd:
Since light nodes no longer call a constructor that takes discovery, fx was never starting it.
This bug was caught by `TestBootstrapNodesFromBridgeNode`


renaynay (approved):
oof.

Wondertan (approved):
oof.

vgonkivs (approved):

vgonkivs:
oof

PR #1951, opened 2023-03-22
fix(swamp): fixing p2p_test and reconstruct_test

distractedm1nd:
Some hotfixes that fix swamp tests. Would like to understand why these fixes work before taking out of draft

### DIFF
--- a/nodebuilder/p2p/bootstrap.go
+++ b/nodebuilder/p2p/bootstrap.go
@@ -8,10 +8,10 @@ import (
 	ma "github.com/multiformats/go-multiaddr"
 )
 
-const envKeyCelestiaBootstrapper = "CELESTIA_BOOTSTRAPPER"
+const EnvKeyCelestiaBootstrapper = "CELESTIA_BOOTSTRAPPER"
 
 func isBootstrapper() bool {
-	return os.Getenv(envKeyCelestiaBootstrapper) == strconv.FormatBool(true)
+	return os.Getenv(EnvKeyCelestiaBootstrapper) == strconv.FormatBool(true)
 }
 
 // BootstrappersFor returns address information of bootstrap peers for a given network.
--- a/nodebuilder/p2p/pubsub.go
+++ b/nodebuilder/p2p/pubsub.go
@@ -67,7 +67,7 @@ func pubSub(cfg Config, params pubSubParams) (*pubsub.PubSub, error) {
 	//  * lotus
 	//  * prysm
 	topicScores := topicScoreParams(params.Network)
-	peerScores, err := peerScoreParams(isBootstrapper, params.Bootstrappers, cfg)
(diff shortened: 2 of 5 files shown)


go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/BurntSushi/toml v1.2.1
 	github.com/alecthomas/jsonschema v0.0.0-20200530073317-71f438968921
 	github.com/benbjohnson/clock v1.3.0
-	github.com/celestiaorg/celestia-app v0.12.0
+	github.com/celestiaorg/celestia-app v0.13.0-rc0
```
Wondertan, line 12: hmmm?

core/testing.go
```diff
@@ -78,15 +82,12 @@ func StartTestNodeWithConfig(t *testing.T, cfg *TestConfig) testnode.Context {
...
+		"private",
 	)
 	require.NoError(t, err)
 
 	cctx, cleanupCoreNode, err := testnode.StartNode(tmNode, cctx)
 	require.NoError(t, err)
-	t.Cleanup(func() {
-		err := cleanupCoreNode()
-		require.NoError(t, err)
-	})
```
Wondertan, line 89: Can we keep the cleanup hidden away from the caller as it was?

nodebuilder/tests/swamp/swamp.go
```diff
@@ -226,6 +227,10 @@ func (s *Swamp) NewNodeWithStore(
 
 	switch t {
 	case node.Bridge:
+		os.Setenv(p2p.EnvKeyCelestiaBootstrapper, "true")
+		defer func() {
+			os.Unsetenv(p2p.EnvKeyCelestiaBootstrapper)
+		}()
```
Wondertan, line 233: A discussion we had not that far ago(you were probably out) is that our tests should not rely on the presence of a bootstrapper. Also, I'd like to point out that setting this effectively makes all the nodes as bootstrappers because all the nodes share the same ENV space within one test run. 

We set bootstrappers before to enable peer exchange, which is no longer necessary as every FN and BN has peer exchange by default, so in theory, removing this should still make the test work unless I miss anything.

distractedm1nd: Its not only peer exchange, its also the DHT server mode (which is the thing that breaks the tests)

Many swamp tests rely on the presence of the bridge node as a bootstrapper or assume the started bridge node to act as one

Wondertan: As discussed in sync, we will go with enabling ServerMode for FN and BN

distractedm1nd: done in #1954, will rebase this pr once merged

Wondertan (dismissed):

renaynay (dismissed):
needs rebase but i approve, let's merge pls to clean up our CI

renaynay (dismissed):

PR #1952, opened 2023-03-22
fix(core): Only assume empty square if reported square size is 1 *and* there are no txs

renaynay:
A bug was caught on blockspacerace network that caused a data root mismatch between bridge nodes and core nodes because celestia-node incorrectly assumes an empty square if there are no transactions which is not the case as squares can contain only tail-padding even if no transactions land in the block itself. 

We also will now: 
* rely on using `data.SquareSize` from the block data for extending shares instead of calculating square size from `len(shares)`, and
* panic on data root mismatch

**Future considerations:**

* we need a fuzzing suite against our block extension mechanism that can generate all kinds of blocks (our integration testing doesn't test for edge cases like this.
* we need to audit the rest of the codebase to ensure we perform the same calculations over square size or assume empty square where we shouldn't be.

### DIFF
--- a/core/eds.go
+++ b/core/eds.go
@@ -20,7 +20,7 @@ import (
 // ExtendedDataSquare (EDS). If there are no transactions in the block,
 // nil is returned in place of the eds.
 func extendBlock(data types.Data) (*rsmt2d.ExtendedDataSquare, error) {
-	if len(data.Txs) == 0 {
+	if len(data.Txs) == 0 && data.SquareSize == uint64(1) {
 		return nil, nil
 	}
 	shares, err := appshares.Split(data, true)
--- a/core/eds.go
+++ b/core/eds.go
@@ -11,7 +11,6 @@ import (
 	appshares "github.com/celestiaorg/celestia-app/pkg/shares"
 	"github.com/celestiaorg/rsmt2d"
 
-	"github.com/celestiaorg/celestia-node/libs/utils"
 	"github.com/celestiaorg/celestia-node/share"
 	"github.com/celestiaorg/celestia-node/share/eds"
 )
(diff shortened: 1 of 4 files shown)


distractedm1nd (dismissed):

vgonkivs (approved):

distractedm1nd (approved):

core/eds.go
```diff
@@ -20,15 +19,15 @@ import (
...
 		return nil, nil
 	}
 	shares, err := appshares.Split(data, true)
 	if err != nil {
 		return nil, err
 	}
-	size := utils.SquareSize(len(shares))
-	return da.ExtendShares(size, appshares.ToBytes(shares))
+
+	return da.ExtendShares(data.SquareSize, appshares.ToBytes(shares))
```
evan-forbes, line 30: I think I was wrong to say that this is necessary, but this is a good change anyway imo

Wondertan (approved):

header/header.go
```diff
@@ -141,8 +141,8 @@ func (eh *ExtendedHeader) Validate() error {
 
 	// ensure data root from raw header matches computed root
 	if !bytes.Equal(eh.DAH.Hash(), eh.DataHash) {
-		return fmt.Errorf("mismatch between data hash commitment from core header and computed data root: "+
-			"data hash: %X, computed root: %X", eh.DataHash, eh.DAH.Hash())
+		panic(fmt.Sprintf("mismatch between data hash commitment from core header and computed data root "+
+			"at height %d: data hash: %X, computed root: %X", eh.Height(), eh.DataHash, eh.DAH.Hash()))
```
evan-forbes, line 145: [very optional]

stopping the node is the correct thing to do here imo, but perhaps not showing the user a stack trace would be good

evan-forbes (approved):

walldiss (approved):

core/eds.go
cmwaters: Yeah I think these should always be the same number. But better probably to use `data.SquareSize` from a defensive perspective

PR #1954, opened 2023-03-22
fix(nodebuilder/p2p): all full and bridge nodes use `dht.ModeServer`

distractedm1nd:
## Overview
Previously, only bootstrappers would use `dht.ModeServer`. Now it is enabled for all full and bridge nodes.


Wondertan (approved):
LGTM. @vgonkivs had a concern that the order of those options matters, but I haven't checked it out

walldiss (approved):

distractedm1nd:
@vgonkivs Should I swap places of the if statements? Will wait for your approval before merging

renaynay (approved):

vgonkivs (approved):

PR #1955, opened 2023-03-22
refactor(core/fetcher): cleanup core fetcher new block subscriber lifecycle

walldiss:
## Overview

Cleanup core fetcher new block subscriber lifecycle:
 - Make Signed block subscriber inside core respect passed ctx timeout and cancel. 
 - Log error if underlying subscription is closed unexpectedly
 - Wait for listener to stop before unsubscribing
 - General cleanup


distractedm1nd (approved):
excellent

PR #1961, opened 2023-03-23
chore(core): Add logging

renaynay:
Add more logs

### DIFF
--- a/core/exchange.go
+++ b/core/exchange.go
@@ -67,8 +67,8 @@ func (ce *Exchange) GetVerifiedRange(
 	for _, h := range headers {
 		err := from.Verify(h)
 		if err != nil {
-			log.Errorw("verifying next header", "last verified height", from.Height(), "err", err)
-			return nil, err
+			return nil, fmt.Errorf("verifying next header against last verified height: %d: %w",
+				from.Height(), err)
 		}
 		from = h
 	}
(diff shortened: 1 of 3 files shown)


Wondertan:
We should wrap error instead and make the caller log the error out

core/exchange.go
```diff
@@ -78,31 +79,36 @@ func (ce *Exchange) Get(ctx context.Context, hash libhead.Hash) (*header.Extende
 	log.Debugw("requesting header", "hash", hash.String())
 	block, err := ce.fetcher.GetBlockByHash(ctx, hash)
 	if err != nil {
+		log.Errorw("fetching block by hash", "hash", hash, "err", err)
 		return nil, err
 	}
 
 	comm, vals, err := ce.fetcher.GetBlockInfo(ctx, &block.Height)
 	if err != nil {
+		log.Errorw("fetching block info", "height", &block.Height, "err", err)
```
walldiss, line 88: Does the caller log errors? I wonder if we actually could wrap error here and log on caller side.

Wondertan: https://github.com/celestiaorg/celestia-node/pull/1961#issuecomment-1481863370

core/exchange.go
```diff
@@ -116,17 +117,21 @@ func (ce *Exchange) Head(ctx context.Context) (*header.ExtendedHeader, error) {
...
 	// extend block data
 	eds, err := extendBlock(b.Data)
 	if err != nil {
+		log.Errorw("extending block data", "height", b.Header.Height, "err", err)
 		return nil, err
 	}
 	// create extended header
 	eh, err := ce.construct(ctx, &b.Header, &b.Commit, &b.ValidatorSet, eds)
 	if err != nil {
+		log.Errorw("constructing extended header", "height", b.Header.Height, "err", err)
```
Wondertan, line 134: wrap?

Wondertan: Every one in the method

vgonkivs (approved):

Issue #1840, opened 2023-02-28
state: `SubmitPayForBlob` should support multiple blobs


Wondertan:
## Overview

The current version of `SubmitPayForBlob` only supports sending only one data blob. At the same time, v0.12.0 of the app introduced the PFB version capable of sending multiple blobs per transaction, which includes them into the chain atomically. The goal is to extend `SubmitPayForBlob` to leverage the new PFB.

## Motivation 

This is required for the `Submit` method of https://github.com/celestiaorg/celestia-node/issues/1835, which allows sending multiple Blobs per call. We could implement each blob as a separate transaction, but then we lose atomicity between their transaction inclusion.

## Details

The recommended path is to extend the method to achieve the goal

renaynay:
@evan-forbes are there any plans to provide an endpoint in app for this ^ ?

evan-forbes:
we don't have any plans atm the use a different endpoint other than the standard one for tendermint that accepts arbitrary txs.

Wondertan:
We don't need any custom endpoints to implement this

renaynay:
so basically `data` field will be `[][]byte` ?

Wondertan:
We should also allow sending two blobs with different namespaces

PR #1963, opened 2023-03-24
refactoring!(state): extend SubmitPayForBlob method

vgonkivs:
## Overview
Resolves https://github.com/celestiaorg/celestia-node/issues/1840

## Checklist

### DIFF
--- a/state/core_access.go
+++ b/state/core_access.go
@@ -20,11 +20,9 @@ import (
 	"google.golang.org/grpc/credentials/insecure"
 
 	"github.com/celestiaorg/celestia-app/app"
-	"github.com/celestiaorg/celestia-app/pkg/appconsts"
 	"github.com/celestiaorg/celestia-app/x/blob"
 	apptypes "github.com/celestiaorg/celestia-app/x/blob/types"
 	libhead "github.com/celestiaorg/go-header"
-	"github.com/celestiaorg/nmt/namespace"
 
 	"github.com/celestiaorg/celestia-node/header"
 )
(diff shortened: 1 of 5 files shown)


state/core_access.go
```diff
@@ -158,17 +158,29 @@ func (ca *CoreAccessor) constructSignedTx(
...
+		return nil, fmt.Errorf("state: amounts of nids=%d and data=%d are not equal", len(nIDs), len(data))
+	}
+
+	blobs := make([]*apptypes.Blob, len(data))
+	for i, share := range data {
+		blobs[i] = &apptypes.Blob{
+			NamespaceId:  nIDs[i],
+			Data:         share,
+			ShareVersion: uint32(appconsts.DefaultShareVersion),
+		}
```
Wondertan, line 176: I think we should just pass the Blob type in the method. Mainly because we have to allow our users to specify share versions and extending one with one more param is a no go

Wondertan (approved):

Wondertan:
* Those who use us must import the app whether we rely on the type or not.
* We cannot make `state` dependant on `blob` as `blob` is supposed to depend on it.
	* However, I assumed the `modblob`, and if we can make the root level `blob` with blob definition, we might avoid cyclical dependency
	* Yeah, `state` would depend on `blob `, and `modblob` would depend on it. No cyclic dep between `modblob` and `state`, so I think we should do it.
	* We can merge this one, then introduce `blob` with the same blob and update this one or punt this PR until we make `blob`, up to @vgonkivs

renaynay (changes requested):
I would actually prefer we hold off on breaking API right now (at least before next release).

vgonkivs:
Created a feature branch for the Blob Module. So, I can merge everything there and then we can test it

vgonkivs:
> I'm worried about adding app types to the API

I agree here with you. We've discussed with Evan extracting the Blob type to a separate repo, so, we can re-use it. But it's not a high priority rn. I'd suggest leaving app types for now, keeping in mind that they will be changed in the future.

@distractedm1nd

distractedm1nd (approved):

renaynay (approved):

state/core_access.go
```diff
@@ -158,17 +156,22 @@ func (ca *CoreAccessor) constructSignedTx(
...
 	gasLim uint64,
+	blobs ...*apptypes.Blob,
 ) (*TxResponse, error) {
-	b := &apptypes.Blob{NamespaceId: nID, Data: data, ShareVersion: uint32(appconsts.DefaultShareVersion)}
+	if len(blobs) == 0 {
+		return nil, errors.New("state: no blobs provided")
+	}
+
+	b := make([]*apptypes.Blob, len(blobs))
+	copy(b, blobs)
```
walldiss, line 168: Did you mean to use b later to protect `blobs` from modifications? There is something wring, since b is not used after copy.

walldiss (approved):

Wondertan (approved):

PR #1968, opened 2023-03-27
perf(modshare): increase the PeersLimit default

Wondertan:
We expect 750 LN for 200 FN(including BN) and 750/200 ~= 4
And we add plus one because the balancing is randomized and not even

### DIFF
--- a/nodebuilder/share/config.go
+++ b/nodebuilder/share/config.go
@@ -32,7 +32,7 @@ type Config struct {
 
 func DefaultConfig() Config {
 	return Config{
-		PeersLimit:        3,
+		PeersLimit:        5,
 		DiscoveryInterval: time.Second * 30,
 		AdvertiseInterval: time.Second * 30,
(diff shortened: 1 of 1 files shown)


renaynay (approved):

vgonkivs (approved):

walldiss:
Maybe a silly question. How proportion of LN/FN could affect `PeersLimit` values? As far as I understand `PeersLimit` is the amount of connections to FN and LN/FN is amount of LN per FN. Why increased amount of LN would need higher values of `PeersLimit`?

Wondertan:
It's more like an increased amount of FNs that can handle more LNs. The goal here is to give LNs more FNs for ITN and increase resilience in multiple dimensions(trust, data, and header availability.) The proportion is just a heuristic I used to find a good value, please suggest a better one if you know.

walldiss (approved):

PR #1969, opened 2023-03-27
ci(.github): Fix kind label for dependabot

renaynay:
The only issue I can see with this is if the dep upgrade leads to a breaking change for some reason. But in that case, we can manually change the label.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -18,7 +18,7 @@ updates:
       - Wondertan
       - renaynay
     labels:
-      - kind:dependencies
+      - kind:deps
   - package-ecosystem: docker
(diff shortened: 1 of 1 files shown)


distractedm1nd (approved):

.github/dependabot.yml
```diff
@@ -19,3 +19,4 @@ updates:
       - renaynay
     labels:
       - kind:dependencies
+      - kind:misc
```
Wondertan, line 22:
```suggestion
      - kind:deps
```

distractedm1nd (approved):

walldiss (approved):

PR #1974, opened 2023-03-27
feat(cmd): RPC over CLI

distractedm1nd:
Based on #3047. Beginning of the CLI.

Adds CLI over RPC. Format:
`celestia rpc [module] [method] [...args]`
`--auth TOKEN` flag sets the auth header (otherwise it reads it from the env, if not found, doesn't set auth)
`--url URL` sets request url, default localhost:26658

Most methods are easy to call. The args for `SubmitPayForBlob` are parsed specially, to improve UX. Other methods can be given special argument parsing rules for improved UX.

These special rules:
- The namespace ID can be encoded as either hex or base64
- The blob can be hex (`0x...`), base64 (`"..."`), or a normal string which will be encoded to base64 (`'"Hello There!"'`)
- `fee` doesn't need to be passed as a string as it normally would need to be, it is passed the same was as gasLim

examples of valid commands:
 ```
 celestia rpc state SubmitPayForBlob 0x1874e642f5dde589 '"Hello there!!"' 2000 100000

 celestia rpc node Info

celestia rpc share GetShare '{"row_roots":["//////////7//////////ql+/VFmJ8PWE9BcjrTDLrY/hzVeGdzFCpfEhiXDXZmt","/////////////////////zHeGnUtPJn8QyPpePSYl4qRVrcUvG2fwptyoA85Myik"],"column_roots":["//////////7//////////ql+/VFmJ8PWE9BcjrTDLrY/hzVeGdzFCpfEhiXDXZmt","/////////////////////zHeGnUtPJn8QyPpePSYl4qRVrcUvG2fwptyoA85Myik"]}' 1 1

celestia rpc header GetByHeight 5
```

### DIFF
--- a/cmd/celestia/rpc.go
+++ b/cmd/celestia/rpc.go
@@ -99,12 +99,22 @@ func parseParams(method string, params []string) []interface{} {
 	}
 
 	for i, param := range params {
-		// try to parse arguments as numbers before adding them as strings
-		num, err := strconv.ParseInt(param, 10, 64)
-		if err != nil {
+		if param[0] == '{' || param[0] == '[' {
+			var raw json.RawMessage
+			if err := json.Unmarshal([]byte(param), &raw); err == nil {
+				parsedParams[i] = raw
+			} else {
+				parsedParams[i] = param
+			}
+		} else {
+			// try to parse arguments as numbers before adding them as strings
+			num, err := strconv.ParseInt(param, 10, 64)
+			if err == nil {
+				parsedParams[i] = num
+				continue
+			}
 			parsedParams[i] = param
 		}
-		parsedParams[i] = num
 	}
 
 	return parsedParams
--- a/cmd/celestia/rpc.go
+++ b/cmd/celestia/rpc.go
@@ -99,7 +99,12 @@ func parseParams(method string, params []string) []interface{} {
 	}
 
 	for i, param := range params {
-		parsedParams[i] = param
+		// try to parse arguments as numbers before adding them as strings
+		num, err := strconv.ParseInt(param, 10, 64)
+		if err != nil {
+			parsedParams[i] = param
+		}
+		parsedParams[i] = num
 	}
 
 	return parsedParams
--- a/cmd/celestia/rpc.go
+++ b/cmd/celestia/rpc.go
@@ -16,6 +16,10 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const (
+	authEnvKey = "CELESTIA_NODE_AUTH_TOKEN"
+)
+
 var requestURL string
 var authTokenFlag string
 
@@ -37,7 +41,7 @@ func init() {
 		&authTokenFlag,
 		"auth",
 		"",
-		"Authorization token (if not provided, the AUTH_TOKEN environment variable will be used)",
+		"Authorization token (if not provided, the "+authEnvKey+" environment variable will be used)",
 	)
 	rootCmd.AddCommand(rpcCmd)
 }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: cmd/celestia/rpc.go [hunk 1]
```diff
@@ -0,0 +1,165 @@
...
+		&requestURL,
+		"url",
+		"http://localhost:26658",
+		"Request URL",
+	)
+	rpcCmd.PersistentFlags().StringVar(
+		&authTokenFlag,
+		"auth",
+		"",
+		"Authorization token (if not provided, the AUTH_TOKEN environment variable will be used)",
```
MSevey, line 40: [non-blocking] I'd update this to `CELESTIA_NODE_AUTH_TOKEN` as `AUTH_TOKEN` is pretty generic for a environment variable name.

distractedm1nd: agreed

cmd/celestia/rpc.go
```diff
@@ -0,0 +1,165 @@
...
+	req, err := http.NewRequest("POST", url, bytes.NewBuffer(requestBody))
+	if err != nil {
+		log.Fatalf("Error creating JSON-RPC request: %v", err)
+	}
+
+	req.Header.Set("Content-Type", "application/json")
+
+	authToken := authTokenFlag
+	if authToken == "" {
+		authToken = os.Getenv("AUTH_TOKEN")
```
MSevey, line 146: [non-blocking] the environment variable string should be a constant

MSevey (commented):
Saw it in action today, awesome!

cmd/celestia/rpc.go
```diff
@@ -0,0 +1,165 @@
...
+			parsedParams[1] = base64.StdEncoding.EncodeToString([]byte(params[1]))
+		default:
+			// otherwise, we assume the user has already encoded their input to base64
+			parsedParams[1] = params[1]
+		}
+		// 3. Fee (state.Int is a string)
+		parsedParams[2] = params[2]
+		// 4. GasLimit (uint64)
+		num, err := strconv.ParseUint(params[3], 10, 64)
+		if err == nil {
```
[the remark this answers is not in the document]
distractedm1nd: Good catch, finished

cmd/celestia/rpc.go
distractedm1nd: finished

renaynay (dismissed):

cmd/celestia/rpc.go
```diff
@@ -0,0 +1,284 @@
...
+	Run: func(cmd *cobra.Command, args []string) {
+		namespace := args[0]
+		method := args[1]
+		params := parseParams(method, args[2:])
+
+		sendJSONRPCRequest(namespace, method, params)
+	},
+}
+
+func parseParams(method string, params []string) []interface{} {
```
Wondertan, line 89: The fact that we still need to switch over most of the commands kinda justifies having a separate command for each. Although, I am ok with current approach.

Wondertan: The possible issue I see is code scalability. Like whether the API changes or we add options to the API, changing this parse func would be painful

cmd/celestia/rpc.go
```diff
@@ -0,0 +1,284 @@
...
+		parsedParams[0], err = parseAddressFromString(params[0])
+		if err != nil {
+			panic(err)
+		}
+		return parsedParams
+	case "QueryRedelegations":
+		var err error
+		parsedParams[0], err = parseAddressFromString(params[0])
+		if err != nil {
+			panic(err)
```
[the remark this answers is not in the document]
distractedm1nd: The error that is passed is a readable error, but I will wrap it

renaynay (approved):

Wondertan (approved):

cmd/celestia/rpc.go
Wondertan: I see. The scalability point is still there

PR #1981, opened 2023-03-28
fix(shrex/peer-manager): put on peer cooldown for peers from discovery

walldiss:
## Overview
In case for given datahash there is no shrex peers, peer is be obtained from discovery pool. Before the fix, If such peer was put on cooldown, it would only be put on cooldown for corresponding to datahash pool. It would mean, that same peer could be obtained for same datahash from discovery pool without cooldown. It could resulting in unrestricted retries to the same peer from discovery and would lead to increased network strain for that peer .

This fix is incomplete and ugly. But it will target and fix the problem, just a bit harder, than it needs to be. Proper fix will take some time to implement and will me done following PR. 
Biggest issue with given solution is that if peer from discovery has ben put on cooldown for one hash, it will be not available for `Peer` method for another hash until cooldown is over.

### DIFF
--- a/share/p2p/peers/manager.go
+++ b/share/p2p/peers/manager.go
@@ -196,7 +196,7 @@ func (m *Manager) Peer(
 		}
 		log.Debugw("returning shrex-sub peer", "hash", datahash.String(),
 			"peer", peerID.String())
-		return peerID, m.doneFunc(datahash, peerID), nil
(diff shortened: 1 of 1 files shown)


Wondertan (approved):
LGTM

distractedm1nd (approved):

renaynay (approved):

PR #1985, opened 2023-03-29
Ugly fix with retries for ITN peering issue

Wondertan:
Should not be merged. The ugly fix for peering issues we observe on the ITN blockspace-race network. 
If a sufficient number of people report that they cannot reproduce it - we should release a patch basing on this branch.

Might be related: https://github.com/libp2p/go-libp2p/issues/2220


renaynay:
Should we close @Wondertan ?

Wondertan:
Solved

PR #1994, opened 2023-03-31
fix(edsstore): close files after Put and Get

distractedm1nd:
Self explanatory


Wondertan (approved):
I feel embarrassed, but glad we found it

PR #2000, opened 2023-04-01
fix(share/eds): use cached accessors in `GetCAR` and `GetDAH`

### LINKED ISSUE #1514: eds/store: Fix `ShardAccessor` caching for use in `GetCAR` and `CARBlockstore`
In #1511 we stopped using the cached `ShardAccessor` on every `GetCAR` and `CARBlockstore` request, as we realized it shares an underlying reader with other callers. 

This caching of the reader was made possible with https://github.com/celestiaorg/dagstore/pull/1, but new modifications need to be made in order to protect the reader from both exhaustion and concurrent reads.

One proposed solution from @walldiss is to use `io.TeeReader` with concurrency protection, making sure a copy is saved into a buffer for subsequent reads.

distractedm1nd:
Closes #1514 . This PR enables accessor cache usage for GetCAR and GetDAH. This will allow shrexeds and shrexnd servers to only require opening the underlying file once (until removed from cache). If many peers request the EDS at once, the server previously needed to open the file each time. 

Will still require a dependency bump of our dagstore fork for the new test to not fail - so draft until then. Related: https://github.com/celestiaorg/dagstore/pull/2


share/eds/store.go
```diff
@@ -298,8 +297,11 @@ func (s *Store) getAccessor(ctx context.Context, key shard.Key) (*dagstore.Shard
 }
 
 func (s *Store) getCachedAccessor(ctx context.Context, key shard.Key) (*accessorWithBlockstore, error) {
-	// try to fetch from cache
-	accessor, err := s.cache.Get(key)
+	lk := &s.cache.stripedLocks[shardKeyToStriped(key)]
+	lk.Lock()
+	defer lk.Unlock()
```
Wondertan, line 302: Is there a way to prevent the Store from accessing the cache's internal fields?

Wondertan (commented):
Another idea that we have to investigate here is the ability to cache EDSes on write. We know that it will be soon accessed and instead of re-reading it back in the LRU cache, we should just keep it in there.

share/eds/store.go
distractedm1nd: Not that I could think of, I thought about it for quite a long time. Was hoping you would have a better idea that would still be clean

distractedm1nd: Let's park that caching EDSes on write/ new mount type as a new issue?

Wondertan: SGTM

Wondertan:
Was trying to find a quick way to solve the access of internal fields but nothing obvious there

share/eds/store.go
```diff
@@ -247,13 +248,12 @@ func (s *Store) GetDAH(ctx context.Context, root share.DataHash) (*share.Root, e
...
 	key := shard.KeyFromString(root.String())
-	accessor, err := s.getAccessor(ctx, key)
+	accessor, err := s.getCachedAccessor(ctx, key)
 	if err != nil {
 		return nil, fmt.Errorf("eds/store: failed to get accessor: %w", err)
 	}
-	defer accessor.Close()
 
-	carHeader, err := carv1.ReadHeader(bufio.NewReader(accessor))
+	carHeader, err := carv1.ReadHeader(bufio.NewReader(accessor.sa.Reader()))
```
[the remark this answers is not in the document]
distractedm1nd: yes, it takes a buffered reader

Wondertan (approved):
It would be nice to have a tagged release, but probably after we upstream

go.mod
```diff
@@ -324,7 +324,7 @@ require (
 
 replace (
 	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.8.0-sdk-v0.46.7
-	github.com/filecoin-project/dagstore => github.com/celestiaorg/dagstore v0.0.0-20221014072825-395797efb659
+	github.com/filecoin-project/dagstore => github.com/celestiaorg/dagstore v0.0.0-20230404123415-177451f83136
```
walldiss, line 330: Nit: we need to tag dagstore impl if it is not going into upstream

share/eds/store.go
```diff
@@ -367,7 +370,6 @@ func (s *Store) Get(ctx context.Context, root share.DataHash) (eds *rsmt2d.Exten
 	if err != nil {
 		return nil, fmt.Errorf("failed to get CAR file: %w", err)
 	}
-	defer f.Close()
```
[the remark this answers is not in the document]
distractedm1nd: It was closing both the mount and the blockstore before this change, which is actually double bad: It closed the file that light nodes were using for sampling, and also closed the file that shrex was using to serve. But now it shares the same location for both blockstore and all shrex requests

share/eds/store.go
```diff
@@ -297,8 +297,11 @@ func (s *Store) getAccessor(ctx context.Context, key shard.Key) (*dagstore.Shard
 }
 
 func (s *Store) getCachedAccessor(ctx context.Context, key shard.Key) (*accessorWithBlockstore, error) {
-	// try to fetch from cache
-	accessor, err := s.cache.Get(key)
+	lk := &s.cache.stripedLocks[shardKeyToStriped(key)]
+	lk.Lock()
+	defer lk.Unlock()
+
+	accessor, err := s.cache.unsafeGet(key)
```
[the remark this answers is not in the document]
distractedm1nd: its just get without a lock, yes

go.mod
distractedm1nd: It is going into upstream, there is a PR right now to add it :)

distractedm1nd:
@renaynay , it closes both the underlying reader (the mount), as well as the mmap reader

share/eds/store.go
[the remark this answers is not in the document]
distractedm1nd: yes but I need to then add after get. the bug comes from someone acquiring the lock between get and add operation, and seeing nothing is there, causing the accessor to be loaded from disk an additional time

Issue #2006, opened 2023-04-03
Error: node: failed to start: error getting latest head during Start: header: not found


jcstein:
### Celestia Node version

v0.8.1

### OS

macOS ARM

### Steps to reproduce it

stop a node and try to restart it

### Expected result

node starts

### Actual result

```bash
celestia light start --core.ip https://grpc-blockspacerace.pops.one/ --core.grpc.port 9090 --gateway --gateway.addr localhost --gateway.port 26659 --p2p.network blockspacerace --metrics.tls=false --metrics --metrics.endpoint otel.celestia.tools:4318

2023-04-03T12:31:56.933-0400	INFO	node	nodebuilder/module.go:22	ccessing keyring...
2023-04-03T12:31:56.937-0400	INFO	module/state	state/keyring.go:52	onstructed keyring signer	{"backend": "test", "path": "/Users/joshstein/.celestia-light-blockspacerace-0/keys", "key name": "my_celes_key", "chain-id": "blockspacerace-0"}
2023-04-03T12:31:56.939-0400	WARN	module/gateway	gateway/module.go:25Gateway is enabled, however gateway endpoints are not authenticated. Use with caution!
2023-04-03T12:31:56.948-0400	INFO	badger	v2@v2.2007.4/levels.go:183	ll 3 tables opened in 4ms

2023-04-03T12:31:56.958-0400	INFO	badger	v2@v2.2007.4/value.go:1158	eplaying file id: 2 at offset: 50239062

2023-04-03T12:31:56.958-0400	INFO	badger	v2@v2.2007.4/value.go:1178	eplay took: 1.75µs

2023-04-03T12:31:56.964-0400	INFO	module/header	header/config.go:58	o trusted peers in config, initializing with default bootstrappers as trusted peers
2023-04-03T12:31:56.966-0400	INFO	header/p2p	p2p/exchange.go:80	lient: starting client	{"protocol ID": "/blockspacerace-0/header-ex/v0.0.3"}
2023-04-03T12:31:56.970-0400	INFO	header/store	store/store.go:169	oaded head	{"height": 136193, "hash": "4668996B4E3EC769F2F711C0554B8E478DDD66502ED8FB232DB6767F137712DD"}
2023-04-03T12:31:56.970-0400	INFO	header/p2p	p2p/subscriber.go:40joining topic	{"topic ID": "/blockspacerace-0/header-sub/v0.0.1"}
[...]

PR #2010, opened 2023-04-04
bypass backoff on dial

Wondertan:
So far we know that connecting to bootstrappers in issues like #2006 fails with `dial backoff`. This context option always forces a dial. An experiment to know if it happens or not


renaynay (approved):
lets try

vgonkivs (approved):

PR #2002, opened 2023-04-03
chore(log): Suppress pubsub logs to WARN and net/identify to ERROR

HoytRen:
Partially fix #1992.

I'm not sure this is a final fix, but at least let's disable the spam message first, or we could not see any real problem.

I let it run on blockspacerace for about 2 hours after modify, don't see any problem besides having to resync bridge again.

My bridge established no more than 20 connections, seems very low.

PR #2011, opened 2023-04-04
feat(logs): Suppress pubsub to WARN level and supress net/identify to ERROR

renaynay:
After some discussions with @wpank and @distractedm1nd , I think it's better if we, by default, suppress pubsub + net/identify logs as they cloud log output without much value to the node runners.


renaynay:
Closing in favour of #2002 ❤️

Wondertan:
The only concern I have is that we also rely on these logs as devs. If we had these logs off before the ShrexSub feature we wouldn't even catch this, which is likely to be an actual problem underneath.

PR #2012, opened 2023-04-04
feat(blob): introduce structure of blob

vgonkivs:
## Overview
Based on https://github.com/celestiaorg/celestia-node/pull/1963.

## Checklist

### DIFF
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -15,7 +15,7 @@ type Proof []nmt.Proof
 
 // Blob represents any application-specific binary data that anyone can submit to Celestia.
 type Blob struct {
-	*types.Blob
+	types.Blob
 
 	commitment Commitment
 }
@@ -31,7 +31,7 @@ func NewBlob(nID namespace.ID, data []byte) (*Blob, error) {
 	if err != nil {
 		return nil, err
 	}
-	return &Blob{Blob: blob, commitment: com}, nil
+	return &Blob{Blob: *blob, commitment: com}, nil
 }
 
 // NamespaceID returns blobs namespaceId.
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -1,0 +1,55 @@
+package blob
+
+import (
+	"github.com/celestiaorg/celestia-app/x/blob/types"
+	"github.com/celestiaorg/nmt"
+	"github.com/celestiaorg/nmt/namespace"
+)
+
+// Commitment is a Merkle Root of the subtree built from shares of the Blob.
+// It is computed by splitting the blob into shares and building the Merkle subtree to be included after Submit.
+type Commitment []byte
+
+// Proof is a collection of nmt.Proofs that verifies the inclusion of the data.
+type Proof []nmt.Proof
+
+// Blob represents any application-specific binary data that anyone can submit to Celestia.
+type Blob struct {
+	*types.Blob
+
+	commitment Commitment
+}
+
+// NewBlob constructs a new blob from the provided namespace.ID and data.
+func NewBlob(nID namespace.ID, data []byte) (*Blob, error) {
+	blob, err := types.NewBlob(nID, data)
+	if err != nil {
+		return nil, err
+	}
+
+	com, err := types.CreateCommitment(blob)
+	if err != nil {
+		return nil, err
+	}
+	return &Blob{Blob: blob, commitment: com}, nil
+}
+
+// NamespaceID returns blobs namespaceId.
+func (b *Blob) NamespaceID() []byte {
+	return b.Blob.NamespaceId
+}
+
+// Data returns blobs raw data.
+func (b *Blob) Data() []byte {
+	return b.Blob.Data
+}
+
+// Commitment returns the commitment of current blob.
+func (b *Blob) Commitment() Commitment {
+	return b.commitment
+}
+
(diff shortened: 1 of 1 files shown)


blob/blob.go
```diff
@@ -0,0 +1,55 @@
...
+type Commitment []byte
+
+// Proof is a collection of nmt.Proofs that verifies the inclusion of the data.
+type Proof []nmt.Proof
+
+// Blob represents any application-specific binary data that anyone can submit to Celestia.
+type Blob struct {
+	*types.Blob
+
+	commitment Commitment
```
distractedm1nd, line 20: Later when users have to send blobs over the API (so don't have access to the constructor), they shouldn't need to include the commitment in the request - so `Commitment() Commitment` should probably check if nil, and call `CreateCommitment` if it doesn't exist

vgonkivs: I would avoid using Mutex just for one case.

distractedm1nd: hmmm. my main wish is about it being optional in the json request, which would leave the commitment field empty when sending it over the wire. This means the node that receives it, will receive it without the commitment. In this case, it should be computed on the fly. I think for this reason its fine if we have a nil check in there - in 99% of cases, users will have used the constructor anyways so it will already be filled.

Your point about the mutex is good - I would leave out the mutex, and leave a comment in the godoc maybe? WDYT?

vgonkivs: So are you convincing me to commit a crime? 😆
Mutex department will arrest me, lol

distractedm1nd: cc @Wondertan thoughts?

distractedm1nd (approved):
My comment can be addressed later on, when we are making sure everything checks out with the RPC :)

blob/blob.go
Wondertan: I don't get why you think introducing mutex here is problematic. I am all for lazy computing of computing. It simplifies API with the cost of one mutex. We don't have to introduce it, though and can comment that it's not concurrency friendly, but it's nice to have it by default

blob/blob.go
```diff
@@ -0,0 +1,55 @@
...
+
+// Commitment returns the commitment of current blob.
+func (b *Blob) Commitment() Commitment {
+	return b.commitment
+}
+
+// Version is the format that was used to encode Blob data and namespaceID into shares.
+func (b *Blob) Version() uint32 {
+	return b.Blob.ShareVersion
+}
```
Wondertan, line 55: Users should be able to set the version and constructor of the Blob does not reflect that

blob/blob.go
Wondertan: By simplifying, I mean it's unnecessary to use a constructor, and the user can simply use struct literal

blob/blob.go
```diff
@@ -0,0 +1,55 @@
...
+// Commitment is a Merkle Root of the subtree built from shares of the Blob.
+// It is computed by splitting the blob into shares and building the Merkle subtree to be included after Submit.
+type Commitment []byte
+
+// Proof is a collection of nmt.Proofs that verifies the inclusion of the data.
+type Proof []nmt.Proof
+
+// Blob represents any application-specific binary data that anyone can submit to Celestia.
+type Blob struct {
+	*types.Blob
```
Wondertan, line 18: Does it have to be a pointer here?

blob/blob.go
vgonkivs: Then it requires additional changes on the app side, bc now they are using `DefaultShareVersion` by default

blob/blob.go
vgonkivs: > I don't get why you think introducing mutex here is problematic. I am all for lazy computing of computing. It simplifies API with the cost of one mutex. We don't have to introduce it, though and can comment that it's not concurrency friendly, but it's nice to have it by default

I did not say it is problematic. Just wanted to avoid unnecessary mutex(if construct commitment in the c-tor)

vgonkivs: This mutex will be needed only once - when commitment is empty and we have to construct it. For other cases, it will be read-only operations.

vgonkivs: > By simplifying, I mean it's unnecessary to use a constructor, and the user can simply use struct literal

wdym? using struct literal to pass there what? Currently, we have to call `types.NewBlob(nID, data)` as it also checks the validity of the namespace inside + the length of the blob. Moving towards the structure literal, we have to ensure the blob is correct each time we will interact with it(meaning receive it from the user) and this is a huge overhead in comparison with a c-tor that will check the validity of data before creating a blob.

blob/blob.go
vgonkivs: no

blob/blob.go
Wondertan: I think we should communicate this to the app team then. They requested the ability to configure the version in the Blob discussion

Wondertan (commented):
My last concern is share versioning. We can merge it as is if the app’s blob doesn't allow us to modify it, but if not we should allow changing this in this PR

vgonkivs:
app already had an opened [PR](https://github.com/celestiaorg/celestia-app/pull/1603) that allows passing share version. Will extend our c-tor with one more param.

PR #2020, opened 2023-04-04
fix: enabling badger gc

distractedm1nd:
While profiling with pyroscope, I noticed that while `badger/v2.(*DB).doWrites` stays constant for days, `badger/v2/y.(*WaterMark)` grows consistently over days.

`Watermark` helps in managing the memory usage and ensuring that certain thresholds are not exceeded, allowing the system to perform optimizations, such as compaction. Badger's garbage collection (GC) primarily targets the value log to free up space on disk, not in memory. The value log GC reclaims space by removing stale and obsolete data from the value log files, which are stored on disk. BUT: invoking the RunValueLogGC() method can have an _indirect_ impact on memory usage: value log GC occurs helps stabilize the growth of the WaterMark by preventing an excessive number of small value log files from accumulating. This can, in turn, lead to more efficient memory usage, as the LSM tree components (MemTable and SSTables) do not need to track as many small value log files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ For more information on setting up a node and the hardware requirements needed, 
 
 ## API docs
 
-Celestia-node public API is documented [here](https://docs.celestia.org/developers/node-api/).
+Celestia-node public API is documented [here](https://docs.celestia.org/category/node-api/).
 
 ## Node types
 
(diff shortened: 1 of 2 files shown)


distractedm1nd:
image

At the red line, I cleared the datastore and started the node with GC enabled. You can see this reflected in the flamegraph as well: that the watermark value is no longer increasing. GC occurs every 15 minutes and takes about ~1 second of CPU time.

renaynay (approved):

Wondertan (approved):

PR #2023, opened 2023-04-04
refactoring(share/p2p): remove TTL in pubsub

vgonkivs:
## Overview
Return TTL back to the default value to avoid message duplication with the same msgID.

## Checklist

### DIFF
--- a/share/p2p/shrexsub/pubsub.go
+++ b/share/p2p/shrexsub/pubsub.go
@@ -46,9 +46,7 @@ type PubSub struct {
 
 // NewPubSub creates a libp2p.PubSub wrapper.
 func NewPubSub(ctx context.Context, h host.Host, networkID string) (*PubSub, error) {
-	// WithSeenMessagesTTL without duration allows to process all incoming messages(even with the same
-	// msgId)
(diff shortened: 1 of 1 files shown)


renaynay:
Can you convert to draft @vgonkivs ? Let's test this first with testground. I'm not sure this resolves the issue.

Wondertan:
This won't help. Because we never rebroadcast messages in the current code via PubSub means, as we always return ValidarionIgnore. We always rebroadcast manually Broadcast.


After thinking about this for a while, I think the actual problem is that we send notifications to everyone, including to those from who we already received the notification.

vgonkivs:
Yeap, I agree that it will not help, but we don't need TTL=0 anyway

Wondertan:
True, I agree that we can merge it as is. It does not fix the root problem, but it doesn't make it worse and removes a confusion configuration.

renaynay:
@vgonkivs you can mark ready for review then, just change description

renaynay (approved):
okya just update the description

Wondertan:
@vgonkivs, pls reference an issue this pr aimed to solve originally

PR #2024, opened 2023-04-05
ADR-14: bootstrap from previously seen peers 

### LINKED ISSUE #1851: feat: Persist peers to disk such that nodes can bootstrap from previously seen peers
All our infrastructure is based around bootstrappers, and each time a node starts; it connects to them and only to them. We do not save any information about other discovered peers during runtime on disk. 

This is a choke point for our networks. In case network bootstrappers go offline, a node connected to the network won't be able to rejoin it after the restart. 

Saving known peers on disk and using them for bootstrapping will solve the problem, decrease the load on bootstrappers, and enhance network decentralization.

To ultimately achieve the goal, we should:
* Use persistent PeerStore instead of in memory
* Periodically saving some fraction of connected peers and using them for bootstrapping instead of hardcoded bootstrappers
	* For this, we should introduce a Bootstrapper p2p component that handles/contains the saving and reconnecting logic
	* We must implement this first and then migrate to libp2p native solutions once https://github.com/libp2p/go-libp2p-kad-dht/issues/254 is implemented.
* The storing of discovered full nodes and the reconnection to them on start should be in the Discovery component. We should additionally make periodic rotations of discovered full nodes.
	* This point needs more thinking, though, and saving might be unnecessary if we do rotations anyway

derrandz:
## Overview

Self-explanatory, closes #1851 

## Checklist


renaynay (commented):
Submitting partial review, ended right before Peer Selection section

docs/adr/adr-013-bootstrap-from-previous-peers.md
```diff
@@ -0,0 +1,210 @@
...
+### Database
+
+We will use a badgerDB datastore to store the addresses of previously seen peers. The database will be stored in the `data` directory, and will be named `peers.db`. The database will have a single key-value pair, where the key is the peer ID, and the value is the peer multiaddress.
+
+The peer store will implement the following interface:
+
+```go
+type PeerStore interface {
+    // Put adds a peer to the store.
+    Put(peer.ID, ma.Multiaddr) error
```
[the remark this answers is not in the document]
derrandz: Totally inline here, since this was an initial draft, it didn't include this but I left a TODO (_if you've seen it_) on the ADR to basically describe the logic for evicting/keeping only nodes since the last (_say_) 24 hours

docs/adr/adr-013-bootstrap-from-previous-peers.md
```diff
@@ -0,0 +1,210 @@
...
+    // Get returns a peer from the store.
+    Get(peer.ID) (ma.Multiaddr, error)
```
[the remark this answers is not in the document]
derrandz: Indeed

docs/adr/adr-013-bootstrap-from-previous-peers.md
```diff
@@ -0,0 +1,210 @@
...
+    // Delete removes a peer from the store.
+    Delete(peer.ID) error
```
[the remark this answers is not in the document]
derrandz: I expect it to be used in the clean up routine that maintains the freshest nodes in the store, but this part is yet to be added to the ADR.

docs/adr/adr-013-bootstrap-from-previous-peers.md
```diff
@@ -0,0 +1,370 @@
...
+type peerStore struct {
+    db datastore.Datastore
+}
+```
+
+### Peer Selection & Persistence
+
+To periodically select the _"good peers"_ that we're connected to and store them in the peer store, we will rely on the internals of `go-header`, specifically the `peerTracker` and `libhead.Exchange` structs.
+
+`peerTracker` has internal logic that continuiously tracks and garbage collects peers, so that if new peers connect to the node, peer tracker keeps track of them in an internal list, as well as it marks disconnected peers for removal. `peerTracker` also blocks peers that fail to answer*.
```
renaynay, line 64:
```suggestion
`peerTracker` has internal logic that continuiously tracks and garbage collects peers, so that if new peers connect to the node, peer tracker keeps track of them in an internal list, as well as it marks disconnected peers for removal. `peerTracker` also blocks peers that behave maliciously*.
```

docs/adr/adr-013-bootstrap-from-previous-peers.md
```diff
@@ -0,0 +1,370 @@
...
+ 
+        p.peerLk.Lock()
+        defer p.peerLk.Unlock()
+        // remove peer from cache.
+        delete(p.trackedPeers, pID)
+        delete(p.disconnectedPeers, pID)
++       p.onBlockedPeer(PIDToAddrInfo(pID))
+ }
+```
+We are assuming a function named `PIDToAddrInfo` that converts a `peer.ID` to a `peer.AddrInfo` struct for this example's purpose.
```
renaynay, line 141: host.Peerstore().PeerInfo(id)

docs/adr/adr-013-bootstrap-from-previous-peers.md
```diff
@@ -0,0 +1,370 @@
...
+        RangeRequestTimeout time.Duration
+        // TrustedPeersRequestTimeout a timeout for any request to a trusted peer.
+        TrustedPeersRequestTimeout time.Duration
+        // networkID is a network that will be used to create a protocol.ID
+        networkID string
+        // chainID is an identifier of the chain.
+        chainID string
++
++       onUpdatedPeers func([]peer.AddrInfo)
++       onBlockedPeer func(peer.AddrInfo)
```
renaynay, line 173: if we want to put it in params, it'd have to be accessible externally so needs to be exported.

docs/adr/adr-013-bootstrap-from-previous-peers.md
```diff
@@ -0,0 +1,370 @@
...
+       }
+}
+```
+
+The event handlers to be supplied are callbacks that either:
+
+1. Put the new peer list into the peer store
+2. Remove a blocked peer from the peer store
+
+Such callbacks are easily passable as options sat header module construction time, example:
```
renaynay, line 240:
```suggestion
Such callbacks are easily passable as options at header module construction time, example:
```

docs/adr/adr-014-bootstrap-from-previous-peers.md
```diff
@@ -0,0 +1,557 @@
...
+
+`peerTracker` has internal logic that continuiously tracks and garbage collects peers based on whether they connected to the node, disconnected from the node, or responded to the node while taking into consideration response speed. All connected peers are kept track of in a list, and disconnected peers are marked for removal, all as a part of a garbage collection routine. `peerTracker` also blocks peers that behave maliciously, but not in a routine, instead, it does so on `blockPeer` call.
+
+We intend to use this internal logic to periodically select peers that are "good" and store them in the peer store mentioned in the storage section. To do this, we will "hook" into the internals of `peerTracker` and `libhead.Exchange` by defining new "event handlers" intended to handle two main events from `peerTracker`:
+
+1. `OnUpdatedPeers`: _This event is triggered on every garbage collection cycle from `peerTracker`. Before the triggering of this event, the list would have underwent changes from other routines as well and thus the most final version will be available at the time of triggering this event ._
```
distractedm1nd, line 82: This hook should not be called OnUpdatedPeers if it is triggered on every gc cycle. Probably should be called OnGCPeers or OnPeerGC

distractedm1nd, line 82:
```suggestion
1. `OnUpdatedPeers`: _This event is triggered on every garbage collection cycle from `peerTracker`. The list will have undergone changes from the other routines before this event is triggered._
```

docs/adr/adr-014-bootstrap-from-previous-peers.md
```diff
@@ -0,0 +1,557 @@
...
+</br>
+
+### What systems will be affected?
+
+* `peerTracker` and `libhead.Exchange` from `go-header`
+* `newInitStore` in `nodebuilder/header.go`
+
+### Storage
+
+We will use a badgerDB datastore to store the addresses of previously seen good peers. The database will be stored in the `data` directory, and will have a store prefix `good_peers`. The database will have a single key-value pair, where the key is `peers` and the value is a `[]PeerInfo`
```
distractedm1nd, line 49: This will be the only key in the datastore that receives updates, filling the value log. Because it's not updated very often and the value is very large, it won't cause bloat, but we should still note this. It probably means we should keep GC enabled.

docs/adr/adr-014-bootstrap-from-previous-peers.md
```diff
@@ -0,0 +1,557 @@
...
+
+The event handlers to be supplied are callbacks that either:
+
+1. Put the new peer list into the peer store
+2. Remove a blocked peer from the peer store
+
+Such callbacks are easily passable as options at header module construction time, example:
+
+```go
+// newP2PExchange constructs a new Exchange for headers.
```
distractedm1nd, line 269: godoc is incorrect

docs/adr/adr-014-bootstrap-from-previous-peers.md
```diff
@@ -0,0 +1,557 @@
...
+func newPeerStore(cfg Config) func(ds datastore.Datastore) (PeerStore, error) {
+       return PeerStore(namespace.Wrap(ds, datastore.NewKey("peer_list")))
```
distractedm1nd, line 271: this key is not consistent with the one named above

docs/adr/adr-014-bootstrap-from-previous-peers.md
```diff
@@ -0,0 +1,557 @@
...
+            ...
+            p2p.WithChainID(network.String()),
++           p2p.WithOnUpdatedPeers(func(peers []peer.ID) {
++               var topTen []peer.ID
++               if len(peers) >= 10 {
++                  topTen = peers[:9]
++               } else {
++                   topTen = peers
++               }
++               peerStore.Put(topTen)
```
distractedm1nd, line 295: Here the top 10 are stored but above you say the sorting logic is optional - please add it above for consistency

docs/adr/adr-014-bootstrap-from-previous-peers.md
```diff
@@ -0,0 +1,557 @@
...
+In this approach, we will periodically sample peers from the host's `Peerstore` and select the "good peers", with the "good peers" being the ones that have been connected to the longest possible period of time.
+
+Taking inspiration from [this proposition](https://github.com/libp2p/go-libp2p-kad-dht/issues/254#issuecomment-633806123) we suggest to look for peers that have been continuously online, such that, for the given buckets of connection durations:
+
+* < 1 hour ago
+* < 1 day ago
+* < 1 week ago
+* < 1 month ago.
+
+the "good enough peers" to select should be present in most recent buckets, and should be peers we've seen often. For example, if a peer has been connected to the node for 1 day, then it should figure up in the buckets: "< 1 hour" and , "< 1 day", hence making it a "good enough peer" to select for persistence.
```
distractedm1nd, line 376: figure up?

docs/adr/adr-014-bootstrap-from-previous-peers.md
```diff
@@ -0,0 +1,557 @@
...
+
+The bucketing logic that triages peers into the mentioned buckets is executed periodically, with a new list of "good enough peers" being selected and persisted at that same time, and also at the moment of node shutdown.
+
+## Implementation
+
+### Storage
+
+We suggest to use the same storage mechanism proposed in approach A.
```
distractedm1nd, line 384: Please note how often this key will be written to in this approach, because it is relevant for our storage backend

docs/adr/adr-014-bootstrap-from-previous-peers.md
```diff
@@ -0,0 +1,557 @@
...
+A peer that has a list of timestamps whose length is equal to 10, and the oldest timestamp was a week ago and the newest from an hour ago. This will make the peer fall into "< 1 week ago" bucket but also all the other recent buckets.
+
+We suggest to also turn the "amount of times seen" into buckets, such that we sort peers into buckets of "how many times" they were seen, modifying the selection criteria for good peers to become:
+
+* Peers that fall in most recent buckets + Peers that fall in highest seen count buckets
+
+**The interface**: We will go with the name `Bucketer` for the lack of a better one for the moment
+
+```go
+type Bucketer interface {
```
distractedm1nd, line 407: Why does this need an interface?

docs/adr/adr-014-bootstrap-from-previous-peers.md
```diff
@@ -0,0 +1,557 @@
...
+```
+
+The `AddPeer` or `AddPeers`, and the `Sort` methods are intended to be called from within a procedure `fetchSortPersistPeers` that is called within a routine that periodically fetches peers from the peer store of the host (_`peerstore.Peerstore`_) which contains all the peers we are connected to.
+
+`fetchSortPersistPeers` will also be called on node shutdown.
+
+`fetchSortPersistPeers` will be as follows:
+
+```go
+func fetchSortPersistPeers(bcktr Bucketer, store PeerStore, host host.Host) {
```
distractedm1nd, line 445: This function does too much. The routine that periodically fetches peers could maybe add to the bucket, even if its in another subroutine - and pass the peers to this function as a slice. This function can be called `persistPeers` and be responsible for sorting and persisting the peers to the store. Why is the storage not a method on the peerstore?

Also, maybe, add a method to the bucketer that returns the sorted best peers?

docs/adr/adr-014-bootstrap-from-previous-peers.md
```diff
@@ -0,0 +1,557 @@
...
++     ticker := time.NewTicker(cfg.filterBootstrappersInterval * time.Millisecond)
++     go func() {
++       select {
++         case <-ctx.Done():
++         case <-ch:
++         case <-ticker.C:
++            fetchSortPersistPeers(bcktr, pstore, host) 
++       }
++
++     lc.Append(fx.Hook{
```
distractedm1nd, line 477: Please use fx.Annotate for start and stop instead of hooks

distractedm1nd (commented):
first scan

renaynay (commented):
My only suggestion is to decrease the amount of code snippets you include for approach A 

I think the most important thing is to outline where the hooks will be passed, why they occur inside peer tracker, and what the actual hooks will look like themselves. The other stuff is unnecessary and distracting - it'll also probably look different in the implementation. The idea here is just to outline how it'd be integrated. 

ideally this ADR could also exist inside go-header somehow if we decide to go with approach A :)

docs/adr/adr-014-bootstrap-from-previous-peers.md
derrandz: Although `OnUpdatedPeers` seems to be a misnomer because it's an event that happens within GC, it's still descriptive as to what actually happens to the peer list. `OnGCPeers` sounds a bit vague to me tbh

docs/adr/adr-014-bootstrap-from-previous-peers.md
derrandz: show up

docs/adr/adr-014-bootstrap-from-previous-peers.md
derrandz: That's the best way I could use to describe a behavior in this ADR

docs/adr/adr-014-bootstrap-from-previous-peers.md
derrandz: In terms of storage, we have to distinguish between `Peerstore` (_libp2p's in-mem peerstore_) and the `PeerStore`, the badger store we define for this ADR's purposes.

If you read through the proposed code, the bucketer has a `BestPeers` method that returns the sorted best peers.

I am not entirely opposed to turning this method into `persistPeers([]peer.ID, host.Host)` such that it only receives best peers as a subroutine. This was intended to be an illustrative example and not necessarily a spec for how to write the function.

docs/adr/adr-014-bootstrap-from-previous-peers.md
derrandz: Since this was used in the context of an `Invoke` I am not sure what to annotate, since we are dealing with an invocation. Will look into this.

docs/adr/adr-014-bootstrap-from-previous-peers.md
distractedm1nd: Never name things for how they are used elsewhere, name them for what they are

docs/adr/adr-014-bootstrap-from-previous-peers.md
distractedm1nd: example from existing code:

```
	fx.Invoke(fx.Annotate(
				func(
					bcast libhead.Broadcaster[*header.ExtendedHeader],
					fetcher *core.BlockFetcher,
					pubsub *shrexsub.PubSub,
					construct header.ConstructFn,
					store *eds.Store,
				) *core.Listener {
					return core.NewListener(bcast, fetcher, pubsub.Broadcast, construct, store, p2p.BlockTime)
				},
				fx.OnStart(func(ctx context.Context, listener *core.Listener) error {
					return listener.Start(ctx)
				}),
				fx.OnStop(func(ctx context.Context, listener *core.Listener) error {
					return listener.Stop(ctx)
				}),
			)),
```

derrandz: @distractedm1nd I am familiar with that example, except that we'd have to have something to annotate. In this example, I would create a new object type called:
```go
type TimeManager struct {
   ch chan struct{}
   ticker time.Ticker
}
```
And have a function to provide this type, as the thing to be annotated, so that we can convert the `Invoke` style into the `Annotate` style
```go
	fx.Invoke(fx.Annotate(func(ctx context.Context) TimeManager {
  	               return TimeManager{ ticker: time.NewTicker(...) }
                               },
				fx.OnStart(func(ctx context.Context) error {
// launch go routine to periodically call fetchSortPersist...
				}),
				fx.OnStop(func(ctx context.Context) error {
				// trigger routine stop by signaling on channel
				}),
			)),

```

distractedm1nd (approved):

docs/adr/adr-014-bootstrap-from-previous-peers.md
```diff
@@ -0,0 +1,474 @@
...
+
+Currently, when a node joins/comes up on the network, it relies, by default, on hardcoded bootstrappers for 3 things:
+
+1. initializing its store with the first (genesis) block
+2. requesting the network head to which it sets its initial sync target, and
+3. bootstrapping itself with some other peers on the network which it can utilise for its operations.
+
+This is a centralization bottleneck as it happens both during initial start-up and for any re-starts.
+
+In this ADR, we wish to aleviate this problem by allowing nodes to bootstrap from previously seen peers. This is a simple solution that allows the network alleviate the start-up centralisation bottleneck such that nodes can join the network without the need for hardcoded bootstrappers up re-start
```
liamsi, line 23: maybe:
```suggestion
In this ADR, we wish to aleviate this problem by allowing nodes to bootstrap from previously seen peers. This is a simple solution that allows the network alleviate the start-up centralisation bottleneck such that nodes can join the network without the need for hardcoded bootstrappers being up on re-start.
```

docs/adr/adr-014-bootstrap-from-previous-peers.md
```diff
@@ -0,0 +1,474 @@
...
+            p2p.WithChainID(network.String()),
++           p2p.WithOnPeersGC(func(peers []peer.ID) {
++               var topTen []peer.ID
++               if len(peers) >= 10 {
++                  topTen = peers[:9]
++               } else {
++                   topTen = peers
++               }
++               peerStore.Put(topTen)
++           }),
```
Wondertan, line 225: This code snippet demonstrates where the main logic of decided Approach A is supposed to land. And it looks like the idea is to land it inside headermodule inside nodebuilder. 

This approach has two important downsides:
* go-header does not have access to the peers it stored.
	* It needs access to those peers because of [this issue](https://github.com/celestiaorg/go-header/issues/25). We want `p2p.Exchange` to be able to request on `Head`(and on other similar methods) non-bootstrap peers we persist. This issue is ___required___ to fully complete the story of coming to the network without bootstrap peers.
	* I would argue that this ADR should consider the above issue and even provide a solution. (cc @renaynay) 
* It makes the whole solution for the bootstrapping problems we have only usable by the node.
	* We can achieve greater and more impactful reusability(comparing to e.g. checkpointStore reuse I mentioned above) if we keep this logic inside of the go-header, together with any storing solution we settle on(with interface or not).
	* As of now, we have only one more consumer of go-header besides node, which is Rollkit. At a much later point, they will face the same problem as we do here. The valuable thing we can do now is to package go-header with our complete solution.

My suggestion to fix the above:
* Implement option `WithPeerPersitence(datastore.Datastore)` instead of `WithOnPeersGC` for `p2p.Exchange`
* If set, persists all the peers locally with or without GC cycles.
  * If with, we should decrease the GC cycle time to like 1 min, as 30mins are rare, and LNs running Exchange with this opt won't save anything. GCing also is not that expensive to run it that seldom.
* On start, it will auto-connect to these peers in a non-blocking manner(like we do for trusted peers in the 'p2p.Exchange')
	* Note that we cannot call these peer we persisted as trusted(will explain in further comment)

docs/adr/adr-014-bootstrap-from-previous-peers.md
```diff
@@ -0,0 +1,474 @@
...
++               if err != nil {
++                       return nil, err
++               }
++               peers := append(peers, list...)
+                ids := make([]peer.ID, len(peers))
+                for index, peer := range peers {
+                        ids[index] = peer.ID
+                        host.Peerstore().AddAddrs(peer.ID, peer.Addrs, peerstore.PermanentAddrTTL)
+                }
+                exchange, err := p2p.NewExchange[*header.ExtendedHeader](host, ids, conngater, opts...)
```
Wondertan, line 269: Here we take all the persistent peers and use them as trusted. However, trusted peers have a special trusted status for ___security___ reasons, and we can't take any random peers, even with high scores, and put them as trusted.

Trusted peers are exceptional because they are used for 'automatic subjective initialization'. Subjective initialization is part of any ProofOfStake protocol, and, in simple terms, it is bootstrapping of the current network validator set or header. Any user joining the network has to get the header somehow. Usually, in other protocols, users must manually get these headers/valset from trusted parties and manually provide them to their software. We went on a different, more user-friendly path where we request the initial headers from these trusted peers instead of forcing users to find them manually. Users are free to select custom trusted peers if they don't want to trust default bootstrappers and our infrastructure. 

If any random peer becomes trusted, it opens up a new vulnerability for go-header consumers. The malicious trusted peer can fool a user into following his fake fork by performing the Long Range Attack. This is why we cannot make persisted peers trusted and have to manage them separately.

Wondertan (changes requested):
Requesting subtle changes for Approach A. Overall, I like the way the doc is structured and written.

docs/adr/adr-014-bootstrap-from-previous-peers.md
```diff
@@ -0,0 +1,474 @@
...
+
+(_Original issue available in references_)
+
+## Decision
+
+Periodically store the addresses of previously seen peers in a key-value database, and allow the node to bootstrap from these peers on start-up, by following the methodology of:
+
+* Approach A.
+
+Caveats: In regards to the storage section of approach A, we decided to factorize the store implementation under `das/store.go` to be used by both approach A and the checkpoint store in `daser`.
```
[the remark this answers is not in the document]
derrandz: I think this point needs clarification, I should add to the ADR's doc.

What's meant in here is basically factorizing the `Put` and `Load` behaviors that are implemented against the `ds.Datastore` which should cause no coupling whatsoever.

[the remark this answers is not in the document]
derrandz: On second read, the main benefit of extraction is less code repetition, but such factorization will be impactful for daser for sure, so to avoid impacting DASer, I will mention that we will follow the same pattern only and no factorization will take place. 👍🏻

docs/adr/adr-014-bootstrap-from-previous-peers.md
```diff
@@ -214,67 +210,67 @@ Such callbacks are easily passable as options at header module construction time
...
-                exchange, err := p2p.NewExchange[*header.ExtendedHeader](host, ids, conngater, opts...)
-                ...
-                return exchange, nil
-        }
- }
+        exchange, err := p2p.NewExchange(...,
+                p2p.WithParams(cfg.Client),
+                p2p.WithNetworkID[p2p.ClientParameters](network.String()),
+                p2p.WithChainID(network.String()),
++               p2p.WithSubjectiveInitialization(storeChainHead.IsExpired),
```
renaynay, line 240: could either be a method on the header or just an exported utility function, doesn't matter.

docs/adr/adr-014-bootstrap-from-previous-peers.md
```diff
@@ -214,67 +210,67 @@ Such callbacks are easily passable as options at header module construction time
...
-(_Code Snippet 4.a: Example of required changes to: celestia-node/nodebuilder/header/constructors.go_)
+(_Code Snippet 3.c: Example of required changes to celestia-node/nodebuilder/header/constructors.go_)
+
+This piece information will allow the `Exchange` to distinguish between the two cases and use the correct peer set for each case:
+
+```diff
+ func (ex *Exchange[H]) Head(ctx context.Context) (H, error) {
+        log.Debug("requesting head")
++       // pool of peers to request from is determined by whether the head of the store is expired or not
++       if ex.Params.subjectiveInitialisation {
```
renaynay, line 254: one thing we need to consider additionally that i forgot to mention on our call yesterday is that the Head method on exchange can get called during runtime (not just on start for setting the initial sync target). There should be a way to also determine during runtime whether Head should use all good peers or only trusted. It might have to become an argument for the method. Let's see if we can work around it as I'd prefer to not break interface.

docs/adr/adr-014-bootstrap-from-previous-peers.md
```diff
@@ -214,67 +210,67 @@ Such callbacks are easily passable as options at header module construction time
...
+                // allocate 90% of caller's set deadline for requests
+    ...
+-   for _, from := range trustedPeers {
++         for _, from := range peerset {
+                go func(from peer.ID) {
+                        headers, err := ex.request(reqCtx, from, headerReq)
+                        if err != nil {
+```
 
+Other side-effect changes will have to be done to methods that use `trustedPeers` restrictively for requests (_such as `performRequest`_) to allow requesting to previously seen good peers as well (_Reference: [Issue #25](https://github.com/celestiaorg/go-header/issues/25)_)
```
renaynay, line 273: to use previously seen good peers as well.

docs/adr/adr-014-bootstrap-from-previous-peers.md
Wondertan: I discussed with Rene that we need runtime detection for subjective initialization rather than build time, like in this case. The clean way to keep it without breaking existing abstractions of go-header is to introduce _interface-level options_ for the Head interface. That is, so Syncer can then do `exchange.Head(ctx, WithSubjectiveHead(head))`. This option would allow any `Head` implementation to access the subjective header, like `p2p.Exchange` which will decide which way to request basing on the header

docs/adr/adr-014-bootstrap-from-previous-peers.md
```diff
@@ -0,0 +1,470 @@
...
+```diff
+ func (ex *Exchange[H]) Head(ctx context.Context) (H, error) {
+        log.Debug("requesting head")
++       // pool of peers to request from is determined by whether the head of the store is expired or not
++       if ex.Params.subjectiveInitialisation {
++              // if the local chain head is outside the unbonding period, only use TRUSTED PEERS for determining the head
++              peerset = ex.trustedPeers
++        } else {
++              // otherwise, node has a chain head that is NOT outdated so we can actually request random peers in addition to trusted
++              peerset = append(ex.trustedPeers, ex.params.peerstore.Get())
```
walldiss, line 259: Would it make sense on `Put` operation to store additional time.Now value and check the time passed since last store operation on `Load`, to identify if stored information is likely to be outdate and should be ignored?

docs/adr/adr-014-bootstrap-from-previous-peers.md
```diff
@@ -0,0 +1,514 @@
...
+        delete(p.disconnectedPeers, pID)
++       p.onBlockedPeer(PIDToAddrInfo(pID))
+}
+```
+
+(_Code Snippet 1.c: Example of required changes to: go-header/p2p/peer_tracker.go_)
+
+The `PIDToAddrInfo` converts a `peer.ID` to a `peer.AddrInfo` by retrieving such info from the host's peer store (_`host.Peerstore().PeerInfo(peerId)`_).
+
+The `peerTracker`'s constructor should be updated to accept these new event handlers:
```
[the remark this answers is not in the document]
walldiss: Ignore it, this one is from looong time ago

docs/adr/adr-014-bootstrap-from-previous-peers.md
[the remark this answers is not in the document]
derrandz: I ended up not following this path in the implementation, and the ADR no longer mentions this "factorizing".

For the record, @renaynay I meant by factorizing, DRYing up the code such that we have one base store implementation to be used by both das and peerstore, but that's a forsaken path now.

docs/adr/adr-014-bootstrap-from-previous-peers.md
derrandz: Suggestions implemeneted 👍🏻

docs/adr/adr-014-bootstrap-from-previous-peers.md
derrandz: This concern has been addressed 👍🏻

docs/adr/adr-014-bootstrap-from-previous-peers.md
derrandz: Implemented in PRs and in this ADR

docs/adr/adr-014-bootstrap-from-previous-peers.md
derrandz: Yes indeed

docs/adr/adr-014-bootstrap-from-previous-peers.md
```diff
@@ -0,0 +1,470 @@
...
+
+The peer store will implement the following interface:
+
+```go
+type PeerStore interface {
+   // Put adds a list of peers to the store.
+    Put([]peer.AddrInfo) error
+    // Get returns a peer from the store.
+    Load() ([]peer.AddrInfo, error)
+}
```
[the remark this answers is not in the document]
derrandz: In implementation yes, but in ADR I think an interface is eloquent and transmits the message clearly

docs/adr/adr-014-bootstrap-from-previous-peers.md
```diff
@@ -0,0 +1,470 @@
...
+        // from another peer.
+        RangeRequestTimeout time.Duration
+        // TrustedPeersRequestTimeout a timeout for any request to a trusted peer.
+        TrustedPeersRequestTimeout time.Duration
+        // networkID is a network that will be used to create a protocol.ID
+        networkID string
+        // chainID is an identifier of the chain.
+        chainID string
++
++       peerstore Peerstore
```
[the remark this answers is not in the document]
derrandz: It is odd indeed, but we want this dependency to be optional and given the current signature, we got to do it this way. The implementation is in https://github.com/celestiaorg/go-header/pull/36 please give it a review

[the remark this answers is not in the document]
derrandz: On second thought, I think it shouldn't be a param.
cc @renaynay @Wondertan

renaynay:
Closing this in favour of a doc on go-header instead as the solution is almost fully contained to that repo

PR #2038, opened 2023-04-07
refactor(WIP): only storing ODS on disk

distractedm1nd:
This PR is a very rough, badly coded POC of only storing the ODS on disk. Seems functional on blockspace-race.

The way this works is that 
1. We store only the CAR Header and first quadrant on disk. "ODS CAR"
2. The file gets registered on the dagstore using the`RSMount`. The `RSMount` mount reads the ODS CAR from disk and generates the EDS on the fly, creating the "extended CAR" which is cached in memory
3. On shard registration, the dagstore populates both the top level index of all IPLD blocks (multihash -> shard) as well as the individual CAR index (shard key x CID -> byte offset in extended CAR)
4. On shard acquire, the `RSMount` is fetched so that the ODS CAR is read from disk but the shard accessor returns a mount Reader over the extended CAR.
    - This means that the Blockstore returned will be over the extended CAR, used as usual, and
    - The accessor also works over shrex in the same way, so ODSes can be served as usual

Because recent shard accessors stay in the `eds.Store`'s LRU cache, the extended data square is only calculated once and left in the cache until it is no longer being requested. This can be further optimized by already caching on `Put`, since we know the shard will be accessed almost immediately after (unless the node is doing its initial sync, but this is fine).

There are other efficiency problems with this and a lot of things to be optimized but I think its at the very least a promising PoC

### DIFF
--- a/share/eds/eds.go
+++ b/share/eds/eds.go
@@ -40,6 +40,35 @@ type writingSession struct {
 	w      io.Writer
 }
 
+func WriteODS(ctx context.Context, eds *rsmt2d.ExtendedDataSquare, w io.Writer) (err error) {
+	ctx, span := tracer.Start(ctx, "write-ods")
+	defer func() {
+		utils.SetStatusAndEnd(span, err)
+	}()
+
+	// 1. Reimport EDS. This is needed to traverse the NMT tree and cache the inner nodes (proofs)
+	// TODO: optim, we don't need to cache proofs for writing ODS
+	writer, err := initializeWriter(ctx, eds, w)
+	if err != nil {
+		return fmt.Errorf("share: creating eds writer: %w", err)
+	}
+
+	// 2. Creates and writes Carv1Header
+	//    - Roots are the eds Row + Col roots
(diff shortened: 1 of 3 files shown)


musalbas:
In terms of generating the EDS on the fly, we should ideally not regenerate the entire EDS on the fly if not needed, but only specific chunks or rows that are requested (whatever the API method used for that is) using the `ReconstructSome` method in the reedsolomon library (we should probably implement this in `rsmt2d`). The `ReconstructSome` method in the klauspost library right now is not efficiently implemented, because just recomputes the entire code, but recomputing the specific row/column needed is still way better than recomputing the entire square, and we can probably make that method more efficient in the future by only computing the polynomial at a specific point if we only need a specific chunk (or ie using [barycentric evaluation](https://hackmd.io/@vbuterin/barycentric_evaluation)).

distractedm1nd:
I am closing, as currently during cleanup, but leaving the branch. This was just a POC and would have to be completely redone anyways, and not in this PR. Is tracked by #2059

PR #2039, opened 2023-04-07
refactor(libs/fraud): fraud tests overhaul

Wondertan:
The main goal here is to prepare everything for extraction. There was a cyclic dependency and dependence on ExtendedHeader inside of `fraud`, which blocked the extraction. This patch fixes this in conjunction with cleanings for testing utilities and actual tests.

List of other notable changes:
* Updates go-header and 
	* migrates to the new API for testing in `header`
* Restructures fraud pkg and extract `fraud.Service` and related logic into sub pkg `fraudserv`
	* Keeping only interface definitions inside of the root `fraud`
* Extracts testing utilities into `fraudtest`
* Adds sprinkles of documentation


vgonkivs (approved):
LGTM

header/headertest/testing.go
```diff
@@ -44,6 +43,10 @@ type TestSuite struct {
 	head *header.ExtendedHeader
 }
 
+func NewStore(t *testing.T) libhead.Store[*header.ExtendedHeader] {
+	return headertest.NewStore[*header.ExtendedHeader](t, NewTestSuite(t, 3), 10)
```
renaynay, line 47: do we want to allow these vals to be passed as params here?

libs/fraud/proof.go
```diff
@@ -63,7 +39,6 @@ type Proof interface {
 func OnProof(ctx context.Context, subscriber Subscriber, p ProofType, handle func(proof Proof)) {
 	subscription, err := subscriber.Subscribe(p)
 	if err != nil {
-		log.Error(err)
```
renaynay, line 66: why do you remove logs?

libs/fraud/fraudserv/service_test.go
```diff
@@ -0,0 +1,196 @@
...
+		},
+		sync.MutexWrap(datastore.NewMapDatastore()),
+		enabledSyncer,
+		"private",
+	)
+
+	t.Cleanup(func() {
+		err := serv.Stop(ctx)
+		if err != nil {
+			t.Fatal(err)
```
renaynay, line 192: require?

renaynay (approved):

header/headertest/testing.go
Wondertan: In this case, not. All our tests do not care how many validators are there, as well as how many headers we pregenerate

Wondertan: If such granularity is needed, it is still possible

libs/fraud/proof.go
Wondertan: Subscription does not return any other type of error besides context.Deadline etc, and having a global logger introduced just for this does not look like a idea worth playing with

libs/fraud/fraudserv/service_test.go
Wondertan: Yeah, can do it, but does not really matter and I would just merge it as is already

renaynay: consistency 🤌

PR #2048, opened 2023-04-11
refactor(share/eds): add ErrNotFound to eds store and shrex server

walldiss:
## Overview
If item doesn't exist in store, return explicit error `ErrNotFound`.

### DIFF
--- a/share/eds/store.go
+++ b/share/eds/store.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -35,6 +36,8 @@ const (
 
 	defaultGCInterval = time.Hour
 )
+
+var ErrItemNotExist = errors.New("item not exist in store")
 
 // Store maintains (via DAGStore) a top-level index enabling granular and efficient random access to
 // every share and/or Merkle proof over every registered CARv1 file. The EDSStore provides a custom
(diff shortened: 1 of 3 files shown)


share/eds/store.go
```diff
@@ -36,6 +37,8 @@ const (
 	defaultGCInterval = time.Hour
 )
 
+var ErrItemNotExist = errors.New("item not exist in store")
```
renaynay, line 40:
```suggestion
var ErrItemNotExist = errors.New("item does not exist in store")
```

renaynay (commented):
I like checking for the shard not exist part and responding with NOT FOUND, but I don't think the error wrapper is necessary - can we just check `if errors.Is(err, dagstore.ErrShardUnknown) ` instead?

walldiss:
error wrapper is made to clean up dependancy graph. So shrex server doesn't rely on internal eds store implementation dependancies.

share/eds/store.go
```diff
@@ -36,6 +37,8 @@ const (
 	defaultGCInterval = time.Hour
 )
 
+var ErrItemNotExist = errors.New("item does not exist in store")
```
Wondertan, line 40:
```suggestion
var ErrItemNotExist = errors.New("eds: item does not exist in store")
```

Wondertan (commented):
Why do you add item? It's only about eds, so we can omit item. I would say item is even confusing as its not immediately clear that item means eds.

We also use NotFound everywhere wording rather then NotExist so let's make it consistent

share/eds/store.go
renaynay: fine w me, let's just say 
`eds: not found in store`

Wondertan (approved):

Wondertan:
Also we need to update pR title and description

distractedm1nd (approved):

PR #1481, opened 2022-12-13
feat(cmd/celestia): add config subcommand to bridge

dixitaniket:
Closes: #1414 

## Checklist

- [ ] New and updated code has appropriate documentation
- [ ] New and updated code has new and/or updated testing
- [ ] Required CI checks are passing
- [ ] Visual proof for any user facing features like CLI or documentation updates
- [ ] Linked issues closed with keywords

PR #2068, opened 2023-04-13
feat(cmd): Add `config-remove` and `config-update` commands to CLI

### LINKED ISSUE #1414: cmd/celestia: Create `config` command with subcommands to manage config
`celestia bridge config remove` --> deletes config
`celestia bridge config init` --> initialises with the passed in config or default conf if nothing passed
`celestia bridge config reinit` --> takes config if it exists, applies the diff (to add new sections)

these commands would also have to take a `p2p.network` flag to know which repository to target or a `--node.store` flag if custom.

renaynay:
This PR supersedes #1481 as it was outdated (my fault for not addressing earlier) and also simplifies the `config-update` command. 

It introduces two new commands: 
* `config-remove` which will delete a node's config
* `config-update` which will update an outdated config by loading it, applying any newly added fields, and saving it back in the config path.

There is one unfortunate caveat which is that if a newly added field is merged into the old config, some old custom values from that field may be overwritten which is why I included a prompt to the user to ensure their old custom values were preserved. Custom values from other fields that were unchanged in the config are preserved however.

If there's time, I will investigate how to preserve custom values of a specific field that has been upgraded.

Credit to @dixitaniket for implementing this feature.

Closes #1414


renaynay:
@dixitaniket can you confirm the email I used to co-author the commit is correct?

dixitaniket:
> @dixitaniket can you confirm the email I used to co-author the commit is correct?

Yes it's correct :)

nodebuilder/config_test.go
```diff
@@ -33,3 +34,104 @@ func TestConfigWriteRead(t *testing.T) {
...
+	// ensure this config field is now set after updating the config
+	require.Equal(t, newCfg.Share.PeerManagerParams, cfg.Share.PeerManagerParams)
+	// ensure old custom values were not changed
+	require.Equal(t, "thisshouldnthavechanged", cfg.State.KeyringAccName)
+	require.Equal(t, "7979", cfg.RPC.Port)
+	require.True(t, cfg.Gateway.Enabled)
+}
+
+// outdatedConfig is an outdated config from a light node
+var outdatedConfig = `
```
Wondertan, line 61: Woow, do we have to change this string everytime we change the config?

nodebuilder/config.go
```diff
@@ -80,6 +81,48 @@ func LoadConfig(path string) (*Config, error) {
...
+
+// removeConfig removes Config from the given 'path'.
+func removeConfig(path string) error {
+	return os.Remove(path)
+}
+
+// UpdateConfig loads the node's config and applies new values
+// from the default config of the given node type, saving the
+// newly updated config into the node's config path.
+func UpdateConfig(tp node.Type, storePath string) error {
```
Wondertan, line 97: For both Remove and Update we should lock the repo, so that we are not doing dirty reads/writes

nodebuilder/config_test.go
renaynay: No, it's an `outdatedConfig` :) it's purposefully old.

Wondertan: Ok, cool

nodebuilder/config.go
renaynay: done

Wondertan (approved):
LFG, but let's make an issue to investigate buggy merging. COpy pasting from the description should just work

distractedm1nd (approved):

PR #2069, opened 2023-04-13
feat: if NODE_STORE is set use it in custom init command

smuu:
This change in the `entrypoint.sh` is for supporting docker compose setup in [rollkit/gm](https://github.com/rollkit/gm). See [here](https://github.com/rollkit/gm/pull/2).

When the `NODE_STORE` env variable is set, the init command specifies the `node.store` parameter to this environment variable.

This is non-breaking; if the variable is not set, nothing changes.

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -4,11 +4,16 @@ set -e
 
 if [ "$1" = 'celestia' ]; then
     echo "Initializing Celestia Node with command:"
-    echo "celestia "${NODE_TYPE}" init --p2p.network "${P2P_NETWORK}""
+
+    if [[ -n "$NODE_STORE" ]]; then
+        echo "celestia "${NODE_TYPE}" init --p2p.network "${P2P_NETWORK}" --node.store "${NODE_STORE}""
+        celestia "${NODE_TYPE}" init --p2p.network "${P2P_NETWORK}" --node.store "${NODE_STORE}"
+    else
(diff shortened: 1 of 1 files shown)


distractedm1nd (approved):

docker/entrypoint.sh
```diff
@@ -4,12 +4,17 @@ set -e
...
+    if [[ -n "$NODE_STORE" ]]; then
+        echo "celestia "${NODE_TYPE}" init --p2p.network "${P2P_NETWORK}" --node.store "${NODE_STORE}""
+        celestia "${NODE_TYPE}" init --p2p.network "${P2P_NETWORK}" --node.store "${NODE_STORE}"
+    else
+        echo "celestia "${NODE_TYPE}" init --p2p.network "${P2P_NETWORK}""
+        celestia "${NODE_TYPE}" init --p2p.network "${P2P_NETWORK}"
+    fi
 
     echo ""
+    echo ""
```
renaynay, line 17: why?

renaynay (approved):

docker/entrypoint.sh
smuu: To separate the logs between `init` and `start` when starting the container.

```
...
NAME: my_celes_key
ADDRESS: celestia1gn5ed7fatxpn2ate7vfsd0wwz6vfvc8jk4hjah
MNEMONIC (save this somewhere safe!!!): 
*** *** *** *** *** *** ***


Starting Celestia Node with command:
celestia light start
...
```

I had issues when debugging something, so it's helpful to have some separation between them.

PR #2070, opened 2023-04-13
feat(cli): adding unsafe-reset-store command

### LINKED ISSUE #109: cmd/celestia: Add command to delete celestia repository (datastore)
We need something like `unsafe-reset-all` as a subcommand of all node types. The reason for `unsafe-reset-all` being a subcommand of a node type is because when you init a `full` node (for example), it creates a repository called `celestia-full` which corresponds to the full nodes configuration / data. The same is true for other node types.

distractedm1nd:
Closes #109. Doesn't include any flags. Tested for all node types locally, works well.

### DIFF
--- a/cmd/reset_store.go
+++ b/cmd/reset_store.go
@@ -1,0 +1,26 @@
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+	flag "github.com/spf13/pflag"
+
+	"github.com/celestiaorg/celestia-node/nodebuilder"
+)
+
+// ResetStore constructs a CLI command to reset the store of Celestia Node.
+func ResetStore(fsets ...*flag.FlagSet) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "unsafe-reset-store",
+		Short: "Resets the node's store to a new state. Leaves the keystore and config intact.",
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := cmd.Context()
+
+			return nodebuilder.Reset(StorePath(ctx), NodeType(ctx))
+		},
+	}
+	for _, set := range fsets {
+		cmd.Flags().AddFlagSet(set)
+	}
+	return cmd
+}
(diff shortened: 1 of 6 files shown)


nodebuilder/store.go
```diff
@@ -186,6 +186,19 @@ func keysPath(base string) string {
 	return filepath.Join(base, "keys")
 }
 
+func blocksPath(base string) string {
+	return filepath.Join(base, "blocks")
+}
+
+func transientsPath(base string) string {
+	// we don't actually use the transients directory anymore, but it could be populated from previous versions.
```
renaynay, line 194: should you make an issue to remove this later?

nodebuilder/init.go
```diff
@@ -71,6 +71,53 @@ func Init(cfg Config, path string, tp node.Type) error {
 	return nil
 }
 
+// Reset removes all data from the datastore and dagstore directories. It leaves the keystore and config intact.
+func Reset(path string, tp node.Type) error {
+	path, err := storePath(path)
+	if err != nil {
+		return err
+	}
```
renaynay, line 79: doesn't the storePath already get passed directly to this?

renaynay (approved):

nodebuilder/store.go
distractedm1nd: yes

cmd/reset_store.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/spf13/cobra"
+	flag "github.com/spf13/pflag"
+
+	"github.com/celestiaorg/celestia-node/nodebuilder"
+)
+
+// ResetStore constructs a CLI command to reset the store of Celestia Node.
+func ResetStore(fsets ...*flag.FlagSet) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "unsafe-reset-store",
```
Wondertan, line 13: I liked nuke😁

Wondertan (approved):

walldiss (approved):

Issue #2007, opened 2023-04-03
Ensure we are running only one HTTP server


Wondertan:
We run an HTTP server for every auxiliary service(gateway, RPC, pprof, libp2p metrics) Instead, we should have only one and register new endpoints for each service

tuxcanfly:
Feature request: it would be nice to have default metrics exported similar to libp2p metrics.

Wondertan:
@tuxcanfly, they are exported via opentelemetry and libp2p are soon gonna be exposed only through opentelemetry as well

PR #2078, opened 2023-04-14
fix(modp2p): fix libp2p metrics

Wondertan:
* Introduces unit tests for p2p module builds to avoid such issues in future
* Reworks and cleans up resource manager registering with an actual fix
* Removes libp2p metrics config
	* There is no need for users to configure Prometheus port because of future #2007 
* Moves `WithP2PMetrics` `nodebuilder` -> `modp2p`, because it needs access to private funcs

### DIFF
--- a/nodebuilder/p2p/metrics.go
+++ b/nodebuilder/p2p/metrics.go
@@ -4,60 +4,41 @@ import (
 	"context"
 	"fmt"
 	"net/http"
-	"strconv"
 	"time"
 
-	"github.com/libp2p/go-libp2p"
-	"github.com/libp2p/go-libp2p/core/network"
-	rcmgr "github.com/libp2p/go-libp2p/p2p/host/resource-manager"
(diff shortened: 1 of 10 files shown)


vgonkivs (approved):

distractedm1nd (approved):

walldiss (approved):

nodebuilder/p2p/metrics.go
```diff
@@ -4,60 +4,41 @@ import (
...
-}
+const (
+	promAgentEndpoint = "/metrics"
+	promAgentPort     = "8890"
+)
 
-// WithMetrics option sets up native libp2p metrics up
-func WithMetrics(lifecycle fx.Lifecycle, cfg Config) error {
-	rcmgrObs.MustRegisterWith(prometheus.DefaultRegisterer)
+// prometheusMetrics option sets up native libp2p metrics up
```
renaynay, line 29:
```suggestion
// prometheusMetrics option sets up native libp2p metrics
```

renaynay (approved):

PR #2079, opened 2023-04-14
refactor(share/shrex): cleanup shrex logs

walldiss:
## Overview

Sort out logging level for shrex client / server logs.

DEBUG: unimportant errors, that doesn't affect logic
WARN: errors that could been induced by either party (client or server)
ERROR: node malfunction that needs operator attention

Since shrex client / server implementation propagates errors to the caller, default logging level should be set to ERROR, to reduce cognitive load on operator. Currently it is set in nodebuilder, but I would rather move default log level setting inside the package at some point.

### DIFF
--- a/logs/logs.go
+++ b/logs/logs.go
@@ -19,8 +19,8 @@ func SetAllLoggers(level logging.LogLevel) {
 	_ = logging.SetLogLevel("basichost", "INFO")
 	_ = logging.SetLogLevel("pubsub", "WARN")
 	_ = logging.SetLogLevel("net/identify", "ERROR")
-	_ = logging.SetLogLevel("shrex/nd", "ERROR")
-	_ = logging.SetLogLevel("shrex/eds", "ERROR")
+	_ = logging.SetLogLevel("shrex/nd", "WARN")
+	_ = logging.SetLogLevel("shrex/eds", "WARN")
 }
 
 func SetDebugLogging() {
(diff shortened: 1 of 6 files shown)


logs/logs.go
```diff
@@ -19,6 +19,8 @@ func SetAllLoggers(level logging.LogLevel) {
 	_ = logging.SetLogLevel("basichost", "INFO")
 	_ = logging.SetLogLevel("pubsub", "WARN")
 	_ = logging.SetLogLevel("net/identify", "ERROR")
+	_ = logging.SetLogLevel("shrex/nd", "ERROR")
+	_ = logging.SetLogLevel("shrex/eds", "ERROR")
```
Wondertan, line 23: Why?

Wondertan (dismissed):
Nice. То что доктор прописал

logs/logs.go
walldiss: Explained in PR description. I want to hide warn logs from shrex client/server by default.

Wondertan: Ok, the logic of which level means what makes total sense, but I still don't get why we want to hide warning logs from operators(and from us because they report what they see).  Also operators seems to follow the logic you described and won't panic on Warns

renaynay (approved):
do we consider name change for the logger breaking?

renaynay (approved):

logs/logs.go
walldiss: Yeah, I agree. Lets make default WARN

Wondertan (approved):

PR #2082, opened 2023-04-14
fix(modp2p): disable metrics on libp2p by default

Wondertan:
So metrics were always enabled, even when we didn't ask libp2p. This PR changes that, so they are permanently disabled unless --p2p.metrics are given.

There might be a case that this is a reason behind the famous `header: not found` and our peering issues. Having metrics enabled but never exported can cause blocking in the connection pipeline, but we need further observations to prove that. So far, bootstrappers with metrics enabled and proper exporting are working flawlessly.


Wondertan:
No, there is no relation between metrics and the mentioned peering issue, as it's solely a client-side timeout problem

renaynay (approved):
still makes sense as a change.

walldiss (approved):

PR #2084, opened 2023-04-16
refactor(nodebuilder): Use `my_celes_key` as default if no `--keyring.accname` is specified

renaynay:
Resolves some confusion for users as previous behaviour would pick the first key in the keyring if no keyring.accname was specified and several keys were present in the keyring.


Wondertan (approved):
Gm, gm. I wonder if this will break anything for existing users who embraced the confusion

Issue #2041, opened 2023-04-07
Rework circuit breaking for components on fraud


Wondertan:
The current implementation of circuit breaking is fragile and obscure. Currently, It follows a functional approach, and we just [wrap](https://github.com/celestiaorg/celestia-node/blob/9c0a5fb0626ada6e6cdb8bcd816d01a3aa5043ad/nodebuilder/fraud/lifecycle.go#L12-L15) the Start method of a component to subscribe for a specific fraud type and also give its Stop method, so that once fraud happens, the component is automatically stopped. 

* The [Lifecycle](https://github.com/celestiaorg/celestia-node/blob/9c0a5fb0626ada6e6cdb8bcd816d01a3aa5043ad/nodebuilder/fraud/lifecycle.go#L12-L15) logic is hardly readable
* Requires two contexts in params which is an antipattern
* Is not responsive.
	* The given stop is called asynchronously
	* The subscription closing happens somewhere in the background after the context is canceled, and there is no way to wait until it's done.
		* This prevents us from fixing #2040 
		
Instead, we should introduce a new component, something like `FraudLifecycler`, with an API that allows registering callbacks for fraud events of a specific type. This will: 
* Detangle the logic
* Separate the lifecycle context from the start context
* Enable responsiveness for its internal events

PR #2091, opened 2023-04-17
refactor(nodebuiler): introduce ServiceBreaker and update FX

Wondertan:
This PR solves two problems:
* Updates FX and fixes issues we observed in [the PR](https://github.com/celestiaorg/celestia-node/pull/1801)
* Fixes https://github.com/celestiaorg/celestia-node/issues/2041

Surprisingly, those two problems were related, so I decided to fix them once and for all. The issue with FX happens after [this change](https://github.com/uber-go/fx/pull/983). The outcome of this change is summarized:
> In other words, lifecycle hook annotations can no longer pull in extra dependencies outside of things on which
the annotated constructor is dependent, results that the annotated constructor provides, context.Context
object which is injected by Lifecycle, and the Lifecycle object itself.

Our current code does pull extra dependencies in the service having `modfraud.Lifecycle` on them, like [DASer](https://github.com/celestiaorg/celestia-node/blob/main/nodebuilder/das/module.go#L47). Specifically, it pulls FraudService, and this is no longer allowed. This forces us to rewrite the fraud lifecycling and here is the solution, which additionally satisfies #2041.

This also unblocks https://github.com/celestiaorg/celestia-node/issues/2040, which is now implemented in https://github.com/celestiaorg/go-fraud/pull/1

I tried to split FX update and the refactor into two diff PRs. However, the solution does not work with the old FX version, so we have to couple those.

The chain here is that I needed a new version of FX that fixes `OnStart/OnStop` hooks, and updating created the whole story. The PR that does clean-ups basing on new version of FX will com right after.


distractedm1nd (approved):
Wow, I like this pattern a lot more. The type itself with the generics can look a bit verbose, but all in all its definitely clearer what is happening with the fraud lifecycle now

vgonkivs (approved):
LGTM

Wondertan:
@distractedm1nd, I also don't like how long generic types can be(hello 🦀) and I had to do it so all the ServiceBreaker are considered different types in the DI. There is no other reason for making ServiceBreaker generic. The alternative was naming types via tags/naming types via FX means, but that looks even more cumbersome.

PR #2102, opened 2023-04-19
refactor(modheader): use WithParams options

Wondertan:
Almost got hit by the scope creep and wanted to refactor much more than this change, but stopped and just submitted this particular change I wanted

### DIFF
--- a/nodebuilder/node.go
+++ b/nodebuilder/node.go
@@ -149,6 +149,7 @@ func (n *Node) Stop(ctx context.Context) error {
 func newNode(opts ...fx.Option) (*Node, error) {
 	node := new(Node)
 	app := fx.New(
+		fx.NopLogger,
 		fx.Populate(node),
 		fx.Options(opts...),
 	)
--- a/nodebuilder/node_test.go
+++ b/nodebuilder/node_test.go
@@ -21,7 +21,7 @@ func TestLifecycle(t *testing.T) {
 	var test = []struct {
 		tp node.Type
 	}{
-		{tp: node.Bridge},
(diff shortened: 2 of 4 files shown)


nodebuilder/node_test.go
```diff
@@ -21,7 +21,7 @@ func TestLifecycle(t *testing.T) {
 	var test = []struct {
 		tp node.Type
 	}{
-		{tp: node.Bridge},
+		// {tp: node.Bridge},
```
vgonkivs, line 24: commented code

nodebuilder/node.go
```diff
@@ -149,6 +149,7 @@ func (n *Node) Stop(ctx context.Context) error {
 func newNode(opts ...fx.Option) (*Node, error) {
 	node := new(Node)
 	app := fx.New(
+		fx.NopLogger,
```
vgonkivs, line 152: I'm a bit confused. My local main has `fx.NopLogger`, but our remote main has not. I understand why we need id, but why was it removed?

Wondertan: It was remove by accident

Wondertan: Can extract into a separate PR if necessary

vgonkivs (approved):
lgtm

Wondertan:
Forgot to write that the main motivation for this PR is to minimise cases where we add a new config field but forget to add and set the option for it. This has happened with TrustedPeersRequestTimeout and made new waste ton of time on a debugging into the wrong direction as I assumed the configuration value was set correctly.

Issue #2027, opened 2023-04-05
das: add backoff to retry jobs


walldiss:
### Implementation ideas

If some height are unavailable, DASer will keep retrying without any delay or backoff which could cause network strain. Add backoff strategy inside DASer for headers that has failed multiple times.

PR #2103, opened 2023-04-19
feat(das): add backoff to retry jobs

walldiss:
## Overview
Resolves  https://github.com/celestiaorg/celestia-node/issues/2027


das/backoff.go
```diff
@@ -0,0 +1,53 @@
...
+	backoffStrategy []time.Duration
+}
+
+// retry represents a retry attempt with a backoff delay.
+type retry struct {
+	// count specifies the number of retry attempts made so far.
+	count int
+	// after specifies the time for the next retry attempt.
+	after time.Time
+}
```
Wondertan, line 16: It feels like it should be defined closer in das/state.go

das/state.go
```diff
@@ -41,8 +45,9 @@ func newCoordinatorState(params Parameters) coordinatorState {
 		sampleFrom:    params.SampleFrom,
 		samplingRange: params.SamplingRange,
 		inProgress:    make(map[int]func() workerState),
-		failed:        make(map[uint64]int),
-		inRetry:       make(map[uint64]int),
+		retryStrategy: newRetryBackOff(exponentialBackoff(time.Minute, 4, 4)),
```
Wondertan, line 48: Let's pls move these magic nums to constant with some basic godoc

das/state.go
```diff
@@ -16,11 +17,14 @@ type coordinatorState struct {
 
 	// keeps track of running workers
 	inProgress map[int]func() workerState
+
+	// retryStrategy implements retry backoff backoff
```
Wondertan, line 21:
```suggestion
	// retryStrategy implements retry backoff
```

das/state.go
```diff
@@ -77,14 +82,19 @@ func (s *coordinatorState) handleResult(res result) {
...
 			delete(s.inRetry, h)
 		}
 
-		s.failed[h] = failCount
+		nextRetry, retryExceeded := s.retryStrategy.nextRetry(lastRetry, time.Now())
+		if retryExceeded {
+			log.Warnw("header exceeded maximum amount of sampling attempts",
+				"height", h,
+				"attempts", nextRetry.count)
+		}
```
Wondertan, line 96: Why do we need to limit max amount?

das/backoff.go
```diff
@@ -0,0 +1,53 @@
...
+// retry instance and a boolean value indicating whether the retries amount have exceeded.
+func (s retryBackoff) nextRetry(lastRetry retry, lastAttempt time.Time) (retry retry, retriesExceeded bool) {
+	lastRetry.count++
+
+	if len(s.backoffStrategy) == 0 {
+		return lastRetry, false
+	}
+
+	if lastRetry.count > len(s.backoffStrategy) {
+		// try count exceeded backoff backoff try limit
```
Wondertan, line 34:
```suggestion
		// try count exceeded backoff try limit
```

das/backoff.go
```diff
@@ -0,0 +1,53 @@
...
+		retry.after = lastAttempt.Add(s.backoffStrategy[len(s.backoffStrategy)-1])
+		return lastRetry, true
+	}
+
+	lastRetry.after = lastAttempt.Add(s.backoffStrategy[lastRetry.count-1])
+	return lastRetry, false
+}
+
+// exponentialBackoff generates an array of time.Duration values using an exponential backoff
+// backoff.
```
Wondertan, line 44:
```suggestion
// exponentialBackoff generates an array of time.Duration values using exponential backoff.
```

das/state.go
Wondertan: With exponential factor 4 backoff, I don't see why we should limit anything.

walldiss: Probably logic seems unclear. It is exponential backoff, but after retry reaches max backoff it will keep retrying with max backoff interval, but also log warn.

das/backoff.go
```diff
@@ -0,0 +1,53 @@
+package das
+
+import "time"
+
+// retryBackoff defines a backoff backoff for retries.
+type retryBackoff struct {
+	backoffStrategy []time.Duration
```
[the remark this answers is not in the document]
walldiss: I've added comment section and renamed the accordingly. Hope is is easier to understand now

Wondertan (approved):
:shipit:

das/state.go
```diff
@@ -3,10 +3,21 @@ package das
...
 
+var (
+	// first retry should happen after defaultBackoffInitialInterval
+	defaultBackoffInitialInterval = time.Minute
+	// next retry will happen with delay defaultBackoffMultiplier to previous one
+	defaultBackoffMultiplier = 4
+	// after defaultBackoffMaxRetryCount amount of attempts retry backoff interval will stop growing
+	// and each retry attempt will produce WARN log
+	defaultBackoffMaxRetryCount = 4
+)
```
renaynay, line 19: can we put these defaults in the backoff file?

das/state.go
```diff
@@ -35,14 +49,26 @@ type coordinatorState struct {
 	catchUpDoneCh chan struct{}
 }
 
+// retry represents a retry attempt with a backoff delay.
+type retry struct {
+	// count specifies the number of retry attempts made so far.
+	count int
+	// after specifies the time for the next retry attempt.
+	after time.Time
```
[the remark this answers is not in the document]
renaynay: Can you also make a method on retry called `After()` that will do this inside: 

```
return retry.after.After(time.Now())
```
so that you only have to call if retry.After() ? looks a bit cleaner

You can even rename method to make it more readable.

das/state.go
```diff
@@ -125,15 +156,15 @@ func (s *coordinatorState) recentJob(header *header.ExtendedHeader) job {
...
-// nextJob will return next catchup or retry job according to priority (catchup > retry)
+// nextJob will return next catchup or retry job according to priority (retry -> catchup)
 func (s *coordinatorState) nextJob() (next job, found bool) {
 	// check for catchup job
-	if job, found := s.catchupJob(); found {
+	if job, found := s.retryJob(); found {
 		return job, found
 	}
 
 	// if caught up already, make a retry job
```
renaynay, line 166: change comments in code pls

das/state.go
```diff
@@ -77,14 +103,19 @@ func (s *coordinatorState) handleResult(res result) {
...
+		// if job was already in retry and failed again, persist attempt count
+		lastRetry, ok := s.inRetry[h]
+		if ok {
 			delete(s.inRetry, h)
 		}
 
-		s.failed[h] = failCount
+		nextRetry, retryExceeded := s.retryStrategy.nextRetry(lastRetry, time.Now())
+		if retryExceeded {
+			log.Warnw("header exceeded maximum amount of sampling attempts",
```
renaynay, line 114: so basically all that happens here is that we log a warning if max retries were exceeded but we still keep retrying the height at the last backoff interval rate forever? 

IMO we should have some sort of special categorisation inside daser state that will indicate if height is truly unavailable that would be very apparent when querying das SamplingStats but that can be left for another PR.

walldiss: Thats a good idea and it could be done on representation layer of daser stats. I'll create and issue for it

Wondertan (dismissed):

renaynay (dismissed):

renaynay (approved):

Wondertan (approved):

distractedm1nd (approved):
Do we want to include in today's release?

PR #2108, opened 2023-04-20
refactor(api/gateway) return error from gateway if header is not synced yet

walldiss:
## Overview

If header was not in header store, gateway would hang until client context was canceled. PR allows gateway to return an error immediately in this case.

### DIFF
--- a/api/gateway/availability.go
+++ b/api/gateway/availability.go
@@ -28,6 +28,7 @@ func (h *Handler) handleHeightAvailabilityRequest(w http.ResponseWriter, r *http
 		return
 	}
 
+	//TODO: change this to NetworkHead once the adjacency in the store is fixed.
 	head, err := h.header.LocalHead(r.Context())
 	if err != nil {
 		writeError(w, http.StatusInternalServerError, heightAvailabilityEndpoint, err)
@@ -35,7 +36,8 @@ func (h *Handler) handleHeightAvailabilityRequest(w http.ResponseWriter, r *http
 	}
 	if headHeight := int(head.Height()); headHeight < height {
 		err = fmt.Errorf(
-			"current header store head height: %v is lower then requested height: %v", headHeight, height)
+			"current head store head height: %v is lower then requested height: %v"+
+				" give header sync some time and retry later", headHeight, height)
 		writeError(w, http.StatusServiceUnavailable, heightAvailabilityEndpoint, err)
 		return
 	}
--- a/api/gateway/availability.go
+++ b/api/gateway/availability.go
@@ -36,7 +36,7 @@ func (h *Handler) handleHeightAvailabilityRequest(w http.ResponseWriter, r *http
 	}
 	if headHeight := int(head.Height()); headHeight < height {
 		err = fmt.Errorf(
-			"current head store head height: %v is lower then requested height: %v"+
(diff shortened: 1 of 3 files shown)


Wondertan:
Perfect, let's add a TODO that we will change this to NetworkHead from LocalHead once the adjacency in the store is fixed

api/gateway/availability.go
```diff
@@ -27,6 +28,18 @@ func (h *Handler) handleHeightAvailabilityRequest(w http.ResponseWriter, r *http
...
 	}
 
+	head, err := h.header.LocalHead(r.Context())
+	if err != nil {
+		writeError(w, http.StatusInternalServerError, heightAvailabilityEndpoint, err)
+		return
+	}
+	if headHeight := int(head.Height()); headHeight < height {
+		err = fmt.Errorf(
+			"current header store head height: %v is lower then requested height: %v", headHeight, height)
```
Wondertan, line 38: Let's give users a hint that they have to wait for the sync to finish

Wondertan (approved):
One comment only

api/gateway/availability.go
```diff
@@ -27,6 +28,20 @@ func (h *Handler) handleHeightAvailabilityRequest(w http.ResponseWriter, r *http
...
 
+	//TODO: change this to NetworkHead once the adjacency in the store is fixed.
+	head, err := h.header.LocalHead(r.Context())
+	if err != nil {
+		writeError(w, http.StatusInternalServerError, heightAvailabilityEndpoint, err)
+		return
+	}
+	if headHeight := int(head.Height()); headHeight < height {
+		err = fmt.Errorf(
+			"current head store head height: %v is lower then requested height: %v"+
```
renaynay, line 39:
```suggestion
			"local chain head: %d is lower than requested height: %d"+
```

api/gateway/header.go
```diff
@@ -69,6 +70,19 @@ func (h *Handler) performGetHeaderRequest(
...
 	}
+	//TODO: change this to NetworkHead once the adjacency in the store is fixed.
+	head, err := h.header.LocalHead(r.Context())
+	if err != nil {
+		writeError(w, http.StatusInternalServerError, heightAvailabilityEndpoint, err)
+		return nil, err
+	}
+	if headHeight := int(head.Height()); headHeight < height {
+		err = fmt.Errorf(
+			"current head store head height: %v is lower then requested height: %v"+
```
renaynay, line 81: same thing here

api/gateway/share.go
```diff
@@ -98,10 +99,19 @@ func (h *Handler) getShares(ctx context.Context, height uint64, nID namespace.ID
...
+	//TODO: change this to NetworkHead once the adjacency in the store is fixed.
+	header, err = h.header.LocalHead(ctx)
+	if err != nil {
+		return nil, 0, err
+	}
+
+	if height > 0 {
+		if storeHeight := uint64(header.Height()); storeHeight < height {
+			return nil, 0, fmt.Errorf(
+				"current head store head height: %v is lower then requested height: %v"+
```
renaynay, line 112: same

renaynay (approved):
i would just say "local chain head"

Wondertan (approved):

renaynay:
Fixes #2118

Issue #2109, opened 2023-04-20
shrex-eds/client: doRequest does not respect context


distractedm1nd:
We spent 45 minutes inside of doRequest, needs to be fixed
```
2023-04-19T11:58:06.710-0700    DEBUG   share/getters   getters/shrex.go:137    request failed  {"datahash": "8457031D5912BBB059317F665D2E6463EAE19643DF69EDA4701DD9892783DA96", "peer": "12D3KooWQpubiFH2MFPYj2YvfqeV2r2wvcMHYC3W4BUpDRTetgnh", "attempt": 2, "err": "failed to read status from stream: Application error 0x0 (remote)", "finished (s)": 2681.052818153}
```

walldiss:
Should be fixed after merge of https://github.com/celestiaorg/celestia-node/pull/2115

PR #2112, opened 2023-04-20
fix(shrexeds): hotfixes for respecting context

### LINKED ISSUE #2110: getters/shrex: GetEDS does not respect parent context deadline in all cases
There is an assumption inside of GetEDS that the call to `Peer` will check context for a deadline. This is not always the case, and we should be explicit about checking contexts anyways.

distractedm1nd:
Closes #2110, related to #2109.
Only adds a hotfix for 2109

### DIFF
--- a/share/getters/shrex.go
+++ b/share/getters/shrex.go
@@ -67,6 +67,11 @@ func (sg *ShrexGetter) GetEDS(ctx context.Context, root *share.Root) (*rsmt2d.Ex
 		err     error
 	)
 	for {
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		default:
+		}
 		attempt++
 		start := time.Now()
 		peer, setStatus, getErr := sg.peerManager.Peer(ctx, root.Hash())
--- a/share/getters/shrex.go
+++ b/share/getters/shrex.go
@@ -67,10 +67,8 @@ func (sg *ShrexGetter) GetEDS(ctx context.Context, root *share.Root) (*rsmt2d.Ex
 		err     error
 	)
 	for {
-		select {
-		case <-ctx.Done():
+		if ctx.Err() != nil {
 			return nil, ctx.Err()
-		default:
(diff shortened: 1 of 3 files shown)


share/p2p/shrexeds/client.go
```diff
@@ -45,29 +45,46 @@ func (c *Client) RequestEDS(
...
+			log.Warnw("client: eds request to peer failed",
+				"peer", peer,
+				"hash", dataHash.String(),
+				"err", err)
+		}
+
+		return nil, err
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	}
```
walldiss, line 87: Ideally this context should have higher deadline than original one and log and error if it happen to timeout before `doRequest` call returns. Also would be nice to reset the stream here to enforce network call cancelation

share/getters/shrex.go
```diff
@@ -67,6 +67,11 @@ func (sg *ShrexGetter) GetEDS(ctx context.Context, root *share.Root) (*rsmt2d.Ex
 		err     error
 	)
 	for {
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		default:
+		}
```
walldiss, line 74: why use select on single channel when you could check `ctx.Err != nil`?

share/p2p/shrexeds/client.go
```diff
@@ -45,29 +45,46 @@ func (c *Client) RequestEDS(
...
+	// this is a hotfix for something inside of doRequest not respecting context.
+	// should be reverted when closing issue #2109
+	go func() {
+		eds, err := c.doRequest(ctx, dataHash, peer)
+		if err != nil {
+			errChan <- err
+			return
+		}
+		resultChan <- eds
+	}()
```
renaynay, line 60: see my dm

distractedm1nd:
closing and reopening because gh broke diffs

PR #2113, opened 2023-04-20
fix(cmd/rpc): parsing JSON for all parameters, and adding handler to GetSharesByNamespace

distractedm1nd:
@jcstein Noticed that the json parsing stopped working for GetSharesByNamespace, this PR fixes that and also fixes a minor bug where only the first parameter was checked for JSON contents

### DIFF
--- a/cmd/celestia/rpc.go
+++ b/cmd/celestia/rpc.go
@@ -92,7 +92,11 @@ func parseParams(method string, params []string) []interface{} {
 	switch method {
 	case "GetSharesByNamespace":
 		// 1. Share Root
-		parsedParams[0] = params[0]
+		root, err := parseJson(params[0])
+		if err != nil {
+			panic(fmt.Errorf("couldn't parse share root as json: %v", err))
+		}
+		parsedParams[0] = root
 		// 2. NamespaceID
 		if strings.HasPrefix(params[1], "0x") {
 			decoded, err := hex.DecodeString(params[1][2:])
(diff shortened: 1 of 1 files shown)


jcstein:
Logs for reference:

```
celestia rpc share GetSharesByNamespace "$(celestia rpc header GetByHeight $HEIGHT | jq '.result.dah' -r)" GHTmQvXd5Yk=
{"jsonrpc":"2.0","id":1,"error":{"code":-32700,"message":"unmarshaling params for 'share.GetSharesByNamespace' (param: **da.DataAvailabilityHeader): json: cannot unmarshal string into Go value of type da.DataAvailabilityHeader"}}
```

cmd/celestia/rpc.go
```diff
@@ -219,11 +223,11 @@ func parseParams(method string, params []string) []interface{} {
...
 		if param[0] == '{' || param[0] == '[' {
-			var raw json.RawMessage
-			if err := json.Unmarshal([]byte(param), &raw); err == nil {
-				parsedParams[i] = raw
-			} else {
+			rawJSON, err := parseJson(param)
+			if err != nil {
 				parsedParams[i] = param
+			} else {
+				parsedParams[i] = rawJSON
```
renaynay, line 230: this would be nicer

```suggestion
                         parsedParams[i] = param
			rawJSON, err := parseJson(param)
			if err == nil {
				parsedParams[i] = rawJSON
			}
```

renaynay (dismissed):

Wondertan (approved):

PR #2114, opened 2023-04-20
refactor(share/p2p/shrex): use default timeout for shrex-eds client 

walldiss:
## Overview
Use default timeout for shrex-eds client if no deadline specified by provided context.


share/p2p/shrexeds/client.go
```diff
@@ -131,3 +131,28 @@ func (c *Client) doRequest(
...
+	// if deadline not set, client read deadline defaults to server write deadline
+	if c.params.ServerWriteTimeout != 0 {
+		err := stream.SetReadDeadline(time.Now().Add(c.params.ServerWriteTimeout))
+		if err != nil {
+			log.Debugw("client: setting read deadline", "err", err)
+		}
+	}
+
+	// if deadline not set, client write deadline defaults to server read deadline
+	if c.params.ServerReadTimeout != 0 {
```
Wondertan, line 152: Server timeouts on the client?

distractedm1nd: This is how its done in shrexnd. It makes sense, maybe we could add some constant for the actual networking though

share/p2p/shrexeds/client.go
```diff
@@ -131,3 +131,28 @@ func (c *Client) doRequest(
 		return nil, p2p.ErrInvalidResponse
 	}
 }
+
+func (c *Client) setStreamDeadlines(ctx context.Context, stream network.Stream) {
+	// set read/write deadline to use context deadline if it exists
+	if dl, ok := ctx.Deadline(); ok {
+		if err := stream.SetDeadline(dl); err != nil {
+			log.Debugw("client: setting deadline: %s", "err", err)
+		}
```
renaynay, line 140:
```suggestion
		if err := stream.SetDeadline(dl); err == nil {
		      return
		}
		log.Debugw("client: setting deadline: %s", "err", err)
```

share/p2p/shrexeds/client.go
walldiss: This is setting same timeout for the other side of pipe. We could do separate client side timeouts later if we would need, but as for default settings those seems reasonable to mirror server side.

share/p2p/shrexeds/client.go
walldiss: It should be more like:
```
if err == nil{
   return
}
log.Error
```
Nice catch!

Wondertan (dismissed):

share/p2p/shrexnd/client.go
```diff
@@ -160,10 +160,10 @@ func (c *Client) setStreamDeadlines(ctx context.Context, stream network.Stream)
 	deadline, ok := ctx.Deadline()
 	if ok {
 		err := stream.SetDeadline(deadline)
-		if err != nil {
-			log.Debugw("client-nd: set write deadline", "err", err)
+		if err == nil {
+			return
 		}
-		return
+		log.Debugw("client-nd: set write deadline", "err", err)
```
renaynay, line 166:
```suggestion
		log.Debugw("client-nd: set stream deadline", "err", err)
```

renaynay (approved):

Wondertan (approved):

renaynay (approved):

PR #2117, opened 2023-04-21
fix(share/discovery)!: revamp Discovery

### LINKED ISSUE #2107: bug(share/discovery): amount of peers in discovery shrinks over time
### Problem

It was observed in multiple runs that amount of peers connection established by discovery is shrinking over time and could even reach only 1 peer. Currently shrex uses only peers maintained by discovery for historical syncing. Low amount of peers available from discovery means all blocksync requests are hitting few (or 1) peers and resulting in very high load on those peers. Need to investigate the root cause.

distractedm1nd:
Closes #2107 

There was a discrepancy between the amount of peers inside of the peer manager full node pool, and the limitedSet inside of discovery. It was possible for the node to get in a state where it had no more full nodes to sample from, because the peers inside the limitedSet were blocking on the connection, and never being handed off to the OnPeersUpdate callback. 

To fix this, we (@Wondertan, @walldiss  and I):
* Set a context deadline for the call to Connect with a peer. Previously, it would never timeout. This is because RoutedHost will block until the context is canceled, even though there is a peer dial timeout.
* Only add peers to the limitedSet after we have successfully connected. This turns the previous peerLimit into a "soft" limit. This prevents in-progress connections from clogging spots in the limitedSet, but also allows for more peers to end up in the limited set than the set limit (we don't throw away already connected peers).
* Removed the timer reset upon a peer disconnecting. This peer disconnection happens often, causing a significant delay in the next call to FindPeers.
* Added logs, cleaned up various comments
* Backoff
	* Adds a RemoveBackoff method to the backoffConnector, to remove cancelled connections from the cache
	* Adds Backoff method
* Cleans up limited set
* Added tests for Discovery
* Handle the case where discovered peer is already connected
* Introduces importable Discovery params
* Revisits constans

I have tested this locally, and it works well with various peer limits. The first few calls to findPeers cause the highest peer count in the limited set (roughly about double), but bad connections are trimmed relatively quickly, and within a few cycles, we land with a pretty stable peer set where there are only 1-2 peer changes every interval, if any.

___NOTE: Config breaking!___

### DIFF
--- a/nodebuilder/share/config.go
+++ b/nodebuilder/share/config.go
@@ -17,8 +17,8 @@ var (
 type Config struct {
 	// PeersLimit defines the soft limit of FNs to connect to via discovery.
 	PeersLimit uint
-	// DialTimeout is the timeout for connecting to peers found via discovery.
-	DialTimeout time.Duration
+	// ConnectTimeout is the timeout for connecting to peers found via discovery.
+	ConnectTimeout time.Duration
 	// DiscoveryInterval is an interval between discovery sessions.
 	DiscoveryInterval time.Duration
 	// AdvertiseInterval is a interval between advertising sessions.
@@ -38,7 +38,7 @@ type Config struct {
 func DefaultConfig() Config {
 	return Config{
 		PeersLimit:        5,
-		DialTimeout:       time.Second * 60,
+		ConnectTimeout:    time.Minute * 2,
 		DiscoveryInterval: time.Second * 30,
 		AdvertiseInterval: time.Second * 30,
 		UseShareExchange:  true,
@@ -50,7 +50,7 @@ func DefaultConfig() Config {
 
 // Validate performs basic validation of the config.
 func (cfg *Config) Validate() error {
-	if cfg.DiscoveryInterval <= 0 || cfg.AdvertiseInterval <= 0 || cfg.DialTimeout <= 0 {
+	if cfg.DiscoveryInterval <= 0 || cfg.AdvertiseInterval <= 0 || cfg.ConnectTimeout <= 0 {
 		return fmt.Errorf("nodebuilder/share: %s", ErrNegativeInterval)
 	}
 	if err := cfg.ShrExNDParams.Validate(); err != nil {
--- a/nodebuilder/share/config.go
+++ b/nodebuilder/share/config.go
@@ -17,8 +17,6 @@ var (
 type Config struct {
 	// PeersLimit defines the soft limit of FNs to connect to via discovery.
 	PeersLimit uint
-	// ConnectTimeout is the timeout for connecting to peers found via discovery.
-	ConnectTimeout time.Duration
 	// DiscoveryInterval is an interval between discovery sessions.
 	DiscoveryInterval time.Duration
 	// AdvertiseInterval is a interval between advertising sessions.
@@ -38,7 +36,6 @@ type Config struct {
 func DefaultConfig() Config {
 	return Config{
 		PeersLimit:        5,
-		ConnectTimeout:    time.Minute * 2,
 		DiscoveryInterval: time.Second * 30,
 		AdvertiseInterval: time.Second * 30,
 		UseShareExchange:  true,
@@ -50,7 +47,7 @@ func DefaultConfig() Config {
 
 // Validate performs basic validation of the config.
 func (cfg *Config) Validate() error {
-	if cfg.DiscoveryInterval <= 0 || cfg.AdvertiseInterval <= 0 || cfg.ConnectTimeout <= 0 {
+	if cfg.DiscoveryInterval <= 0 || cfg.AdvertiseInterval <= 0 {
 		return fmt.Errorf("nodebuilder/share: %s", ErrNegativeInterval)
 	}
 	if err := cfg.ShrExNDParams.Validate(); err != nil {
--- a/nodebuilder/share/config.go
+++ b/nodebuilder/share/config.go
@@ -1,27 +1,13 @@
 package share
 
 import (
-	"errors"
-	"fmt"
-	"time"
-
+	disc "github.com/celestiaorg/celestia-node/share/availability/discovery"
 	"github.com/celestiaorg/celestia-node/share/p2p/peers"
 	"github.com/celestiaorg/celestia-node/share/p2p/shrexeds"
 	"github.com/celestiaorg/celestia-node/share/p2p/shrexnd"
 )
 
-var (
-	ErrNegativeInterval = errors.New("interval must be positive")
-)
-
 type Config struct {
-	// PeersLimit defines the soft limit of FNs to connect to via discovery.
-	PeersLimit uint
-	// DiscoveryInterval is an interval between discovery sessions.
-	DiscoveryInterval time.Duration
-	// AdvertiseInterval is a interval between advertising sessions.
-	// NOTE: only full and bridge can advertise themselves.
-	AdvertiseInterval time.Duration
 	// UseShareExchange is a flag toggling the usage of shrex protocols for blocksync.
 	// NOTE: This config variable only has an effect on full and bridge nodes.
 	UseShareExchange bool
@@ -31,25 +17,22 @@ type Config struct {
 	ShrExNDParams *shrexnd.Parameters
 	// PeerManagerParams sets peer-manager configuration parameters
 	PeerManagerParams peers.Parameters
+	// Discovery sets peer discovery configuration parameters.
+	Discovery disc.Parameters
 }
 
 func DefaultConfig() Config {
 	return Config{
-		PeersLimit:        5,
-		DiscoveryInterval: time.Second * 30,
-		AdvertiseInterval: time.Second * 30,
 		UseShareExchange:  true,
 		ShrExEDSParams:    shrexeds.DefaultParameters(),
 		ShrExNDParams:     shrexnd.DefaultParameters(),
 		PeerManagerParams: peers.DefaultParameters(),
+		Discovery:         disc.DefaultParameters(),
 	}
 }
 
 // Validate performs basic validation of the config.
 func (cfg *Config) Validate() error {
-	if cfg.DiscoveryInterval <= 0 || cfg.AdvertiseInterval <= 0 {
-		return fmt.Errorf("nodebuilder/share: %s", ErrNegativeInterval)
-	}
 	if err := cfg.ShrExNDParams.Validate(); err != nil {
 		return err
 	}
--- a/nodebuilder/share/config.go
+++ b/nodebuilder/share/config.go
@@ -17,6 +17,8 @@ var (
 type Config struct {
 	// PeersLimit defines how many peers will be added during discovery.
 	PeersLimit uint
+	// DialTimeout is the timeout for connecting to peers found via discovery.
+	DialTimeout time.Duration
 	// DiscoveryInterval is an interval between discovery sessions.
 	DiscoveryInterval time.Duration
 	// AdvertiseInterval is a interval between advertising sessions.
@@ -36,6 +38,7 @@ type Config struct {
 func DefaultConfig() Config {
 	return Config{
 		PeersLimit:        5,
+		DialTimeout:       time.Second * 15,
 		DiscoveryInterval: time.Second * 30,
 		AdvertiseInterval: time.Second * 30,
 		UseShareExchange:  true,
--- a/nodebuilder/share/config.go
+++ b/nodebuilder/share/config.go
@@ -15,7 +15,7 @@ var (
 )
 
 type Config struct {
-	// PeersLimit defines how many peers will be added during discovery.
+	// PeersLimit defines the soft limit of FNs to connect to via discovery.
 	PeersLimit uint
 	// DialTimeout is the timeout for connecting to peers found via discovery.
 	DialTimeout time.Duration
@@ -38,7 +38,7 @@ type Config struct {
 func DefaultConfig() Config {
 	return Config{
 		PeersLimit:        5,
-		DialTimeout:       time.Second * 15,
+		DialTimeout:       time.Second * 60,
 		DiscoveryInterval: time.Second * 30,
 		AdvertiseInterval: time.Second * 30,
 		UseShareExchange:  true,
@@ -50,7 +50,7 @@ func DefaultConfig() Config {
 
 // Validate performs basic validation of the config.
 func (cfg *Config) Validate() error {
-	if cfg.DiscoveryInterval <= 0 || cfg.AdvertiseInterval <= 0 {
+	if cfg.DiscoveryInterval <= 0 || cfg.AdvertiseInterval <= 0 || cfg.DialTimeout <= 0 {
 		return fmt.Errorf("nodebuilder/share: %s", ErrNegativeInterval)
 	}
 	if err := cfg.ShrExNDParams.Validate(); err != nil {
--- a/share/availability/discovery/backoff.go
+++ b/share/availability/discovery/backoff.go
@@ -12,7 +12,11 @@ import (
 )
 
 // gcInterval is a default period after which disconnected peers will be removed from cache
-const gcInterval = time.Hour
+const (
+	gcInterval = time.Hour
+	// connectTimeout is the timeout used for dialing peers and discovering peer addresses.
+	connectTimeout = time.Minute * 2
+)
 
 var (
 	defaultBackoffFactory = backoff.NewFixedBackoff(time.Hour)
@@ -55,6 +59,10 @@ func (b *backoffConnector) Connect(ctx context.Context, p peer.AddrInfo) error {
 	}
 	cache.nexttry = time.Now().Add(cache.backoff.Delay())
 	b.cacheLk.Unlock()
+
+	ctx, cancel := context.WithTimeout(ctx, connectTimeout)
+	defer cancel()
+
 	return b.h.Connect(ctx, p)
 }
 
--- a/share/availability/discovery/backoff.go
+++ b/share/availability/discovery/backoff.go
@@ -2,7 +2,7 @@ package discovery
 
 import (
 	"context"
-	"fmt"
+	"errors"
 	"sync"
 	"time"
 
@@ -14,7 +14,10 @@ import (
 // gcInterval is a default period after which disconnected peers will be removed from cache
 const gcInterval = time.Hour
 
-var defaultBackoffFactory = backoff.NewFixedBackoff(time.Hour)
+var (
+	defaultBackoffFactory = backoff.NewFixedBackoff(time.Hour)
+	errBackoffNotEnded    = errors.New("share/discovery: backoff period has not ended")
+)
 
 // backoffConnector wraps a libp2p.Host to establish a connection with peers
 // with adding a delay for the next connection attempt.
@@ -48,7 +51,7 @@ func (b *backoffConnector) Connect(ctx context.Context, p peer.AddrInfo) error {
 	cache := b.connectionData(p.ID)
 	if time.Now().Before(cache.nexttry) {
 		b.cacheLk.Unlock()
-		return fmt.Errorf("share/discovery: backoff period has not ended for peer=%s", p.ID.String())
+		return errBackoffNotEnded
 	}
 	cache.nexttry = time.Now().Add(cache.backoff.Delay())
 	b.cacheLk.Unlock()
@@ -65,6 +68,14 @@ func (b *backoffConnector) connectionData(p peer.ID) *backoffData {
 		b.cacheData[p] = cache
 	}
 	return cache
+}
+
+// RemoveBackoff removes peer from the backoffCache. It is called when we cancel the attempt to
+// connect to a peer after calling Connect.
+func (b *backoffConnector) RemoveBackoff(p peer.ID) {
+	b.cacheLk.Lock()
+	defer b.cacheLk.Unlock()
+	delete(b.cacheData, p)
 }
 
 // RestartBackoff resets delay time between attempts and adds a delay for the next connection
--- a/share/availability/discovery/backoff.go
+++ b/share/availability/discovery/backoff.go
@@ -49,10 +49,7 @@ func newBackoffConnector(h host.Host, factory backoff.BackoffFactory) *backoffCo
 
 // Connect puts peer to the backoffCache and tries to establish a connection with it.
 func (b *backoffConnector) Connect(ctx context.Context, p peer.AddrInfo) error {
-	// we should lock the mutex before calling backoffData and not inside because otherwise it could
-	// be modified from another goroutine as it returns a pointer
-	cache := b.backoffData(p.ID)
-	if time.Now().Before(cache.nexttry) {
+	if b.HasBackoff(p.ID) {
 		return errBackoffNotEnded
 	}
 
@@ -74,6 +71,16 @@ func (b *backoffConnector) Backoff(p peer.ID) {
 	b.cacheLk.Lock()
 	b.cacheData[p] = data
 	b.cacheLk.Unlock()
+}
+
+func (b *backoffConnector) HasBackoff(p peer.ID) bool {
+	b.cacheLk.Lock()
+	cache, ok := b.cacheData[p]
+	b.cacheLk.Unlock()
+	if ok && time.Now().Before(cache.nexttry) {
+		return true
+	}
+	return false
 }
 
 // backoffData returns backoffData from the map if it was stored, otherwise it will instantiate
--- a/share/availability/discovery/backoff.go
+++ b/share/availability/discovery/backoff.go
@@ -13,7 +13,7 @@ import (
 
 // gcInterval is a default period after which disconnected peers will be removed from cache
 const (
-	gcInterval = time.Hour
+	gcInterval = time.Minute
 	// connectTimeout is the timeout used for dialing peers and discovering peer addresses.
 	connectTimeout = time.Minute * 2
 )
@@ -30,7 +30,7 @@ type backoffConnector struct {
 	backoff backoff.BackoffFactory
 
 	cacheLk   sync.Mutex
-	cacheData map[peer.ID]*backoffData
+	cacheData map[peer.ID]backoffData
 }
 
 // backoffData stores time when next connection attempt with the remote peer.
@@ -43,35 +43,48 @@ func newBackoffConnector(h host.Host, factory backoff.BackoffFactory) *backoffCo
 	return &backoffConnector{
 		h:         h,
 		backoff:   factory,
-		cacheData: make(map[peer.ID]*backoffData),
+		cacheData: make(map[peer.ID]backoffData),
 	}
 }
 
 // Connect puts peer to the backoffCache and tries to establish a connection with it.
 func (b *backoffConnector) Connect(ctx context.Context, p peer.AddrInfo) error {
-	// we should lock the mutex before calling connectionData and not inside because otherwise it could
+	// we should lock the mutex before calling backoffData and not inside because otherwise it could
 	// be modified from another goroutine as it returns a pointer
-	b.cacheLk.Lock()
-	cache := b.connectionData(p.ID)
+	cache := b.backoffData(p.ID)
 	if time.Now().Before(cache.nexttry) {
-		b.cacheLk.Unlock()
 		return errBackoffNotEnded
 	}
-	cache.nexttry = time.Now().Add(cache.backoff.Delay())
-	b.cacheLk.Unlock()
 
 	ctx, cancel := context.WithTimeout(ctx, connectTimeout)
 	defer cancel()
 
-	return b.h.Connect(ctx, p)
+	err := b.h.Connect(ctx, p)
+	// we don't want to add backoff when the context is canceled.
+	if !errors.Is(err, context.Canceled) {
+		b.Backoff(p.ID)
+	}
+	return err
 }
 
-// connectionData returns backoffData from the map if it was stored, otherwise it will instantiate
+func (b *backoffConnector) Backoff(p peer.ID) {
+	data := b.backoffData(p)
+	data.nexttry = time.Now().Add(data.backoff.Delay())
+
+	b.cacheLk.Lock()
+	b.cacheData[p] = data
+	b.cacheLk.Unlock()
+}
+
+// backoffData returns backoffData from the map if it was stored, otherwise it will instantiate
 // a new one.
-func (b *backoffConnector) connectionData(p peer.ID) *backoffData {
+func (b *backoffConnector) backoffData(p peer.ID) backoffData {
+	b.cacheLk.Lock()
+	defer b.cacheLk.Unlock()
+
 	cache, ok := b.cacheData[p]
 	if !ok {
-		cache = &backoffData{}
+		cache = backoffData{}
 		cache.backoff = b.backoff()
 		b.cacheData[p] = cache
 	}
@@ -86,15 +99,13 @@ func (b *backoffConnector) RemoveBackoff(p peer.ID) {
 	delete(b.cacheData, p)
 }
 
-// RestartBackoff resets delay time between attempts and adds a delay for the next connection
+// ResetBackoff resets delay time between attempts and adds a delay for the next connection
 // attempt to remote peer. It will mostly be called when host receives a notification that remote
 // peer was disconnected.
-func (b *backoffConnector) RestartBackoff(p peer.ID) {
-	b.cacheLk.Lock()
-	defer b.cacheLk.Unlock()
-	cache := b.connectionData(p)
+func (b *backoffConnector) ResetBackoff(p peer.ID) {
+	cache := b.backoffData(p)
 	cache.backoff.Reset()
-	cache.nexttry = time.Now().Add(cache.backoff.Delay())
+	b.Backoff(p)
 }
 
 func (b *backoffConnector) GC(ctx context.Context) {
--- a/share/availability/discovery/backoff.go
+++ b/share/availability/discovery/backoff.go
@@ -91,6 +91,7 @@ func (b *backoffConnector) HasBackoff(p peer.ID) bool {
 	return false
 }
 
+// GC is a perpetual GCing loop.
 func (b *backoffConnector) GC(ctx context.Context) {
 	ticker := time.NewTicker(gcInterval)
 	for {
--- a/share/availability/discovery/backoff.go
+++ b/share/availability/discovery/backoff.go
@@ -85,10 +85,7 @@ func (b *backoffConnector) HasBackoff(p peer.ID) bool {
 	b.cacheLk.Lock()
 	cache, ok := b.cacheData[p]
 	b.cacheLk.Unlock()
-	if ok && time.Now().Before(cache.nexttry) {
-		return true
-	}
-	return false
+	return ok && time.Now().Before(cache.nexttry)
 }
 
 // GC is a perpetual GCing loop.
--- a/share/availability/discovery/backoff.go
+++ b/share/availability/discovery/backoff.go
@@ -64,15 +64,23 @@ func (b *backoffConnector) Connect(ctx context.Context, p peer.AddrInfo) error {
 	return err
 }
 
+// Backoff adds or extends backoff delay for the peer.
 func (b *backoffConnector) Backoff(p peer.ID) {
-	data := b.backoffData(p)
-	data.nexttry = time.Now().Add(data.backoff.Delay())
-
 	b.cacheLk.Lock()
+	defer b.cacheLk.Unlock()
+
+	data, ok := b.cacheData[p]
+	if !ok {
+		data = backoffData{}
+		data.backoff = b.backoff()
+		b.cacheData[p] = data
+	}
+
+	data.nexttry = time.Now().Add(data.backoff.Delay())
 	b.cacheData[p] = data
-	b.cacheLk.Unlock()
 }
 
+// HasBackoff checks if peer is in backoff.
 func (b *backoffConnector) HasBackoff(p peer.ID) bool {
 	b.cacheLk.Lock()
 	cache, ok := b.cacheData[p]
@@ -81,38 +89,6 @@ func (b *backoffConnector) HasBackoff(p peer.ID) bool {
 		return true
 	}
 	return false
-}
-
-// backoffData returns backoffData from the map if it was stored, otherwise it will instantiate
-// a new one.
-func (b *backoffConnector) backoffData(p peer.ID) backoffData {
-	b.cacheLk.Lock()
-	defer b.cacheLk.Unlock()
-
-	cache, ok := b.cacheData[p]
-	if !ok {
-		cache = backoffData{}
-		cache.backoff = b.backoff()
-		b.cacheData[p] = cache
-	}
-	return cache
-}
-
-// RemoveBackoff removes peer from the backoffCache. It is called when we cancel the attempt to
-// connect to a peer after calling Connect.
-func (b *backoffConnector) RemoveBackoff(p peer.ID) {
-	b.cacheLk.Lock()
-	defer b.cacheLk.Unlock()
-	delete(b.cacheData, p)
-}
-
-// ResetBackoff resets delay time between attempts and adds a delay for the next connection
-// attempt to remote peer. It will mostly be called when host receives a notification that remote
-// peer was disconnected.
-func (b *backoffConnector) ResetBackoff(p peer.ID) {
-	cache := b.backoffData(p)
-	cache.backoff.Reset()
-	b.Backoff(p)
 }
 
 func (b *backoffConnector) GC(ctx context.Context) {
--- a/share/availability/discovery/backoff.go
+++ b/share/availability/discovery/backoff.go
@@ -11,8 +11,8 @@ import (
 	"github.com/libp2p/go-libp2p/p2p/discovery/backoff"
 )
 
-// gcInterval is a default period after which disconnected peers will be removed from cache
 const (
+	// gcInterval is a default period after which disconnected peers will be removed from cache
 	gcInterval = time.Minute
 	// connectTimeout is the timeout used for dialing peers and discovering peer addresses.
 	connectTimeout = time.Minute * 2
--- a/share/availability/discovery/backoff.go
+++ b/share/availability/discovery/backoff.go
@@ -19,7 +19,7 @@ const (
 )
 
 var (
-	defaultBackoffFactory = backoff.NewFixedBackoff(time.Hour)
+	defaultBackoffFactory = backoff.NewFixedBackoff(time.Minute * 10)
 	errBackoffNotEnded    = errors.New("share/discovery: backoff period has not ended")
 )
 
--- a/share/availability/discovery/discovery.go
+++ b/share/availability/discovery/discovery.go
@@ -44,10 +44,8 @@ type Discovery struct {
 	connector *backoffConnector
 	// peersLimit is the soft limit of peers to add to the set.
 	peersLimit uint
-	// dialTimeout is the timeout used for dialing peers.
-	// network.WithDialPeerTimeout is not sufficient here,
-	// as RoutedHost only uses it for the underlying dial.
-	dialTimeout time.Duration
+	// connectTimeout is the timeout used for dialing peers and discovering peer addresses.
+	connectTimeout time.Duration
 	// discInterval is an interval between discovery sessions.
 	discoveryInterval time.Duration
 	// advertiseInterval is an interval between advertising sessions.
@@ -69,7 +67,7 @@ func NewDiscovery(
 	h host.Host,
 	d discovery.Discovery,
 	peersLimit uint,
-	dialTimeout,
+	connectTimeout,
 	discInterval,
 	advertiseInterval time.Duration,
 ) *Discovery {
@@ -79,7 +77,7 @@ func NewDiscovery(
 		disc:              d,
 		connector:         newBackoffConnector(h, defaultBackoffFactory),
 		peersLimit:        peersLimit,
-		dialTimeout:       dialTimeout,
+		connectTimeout:    connectTimeout,
 		discoveryInterval: discInterval,
 		advertiseInterval: advertiseInterval,
 		onUpdatedPeers:    func(peer.ID, bool) {},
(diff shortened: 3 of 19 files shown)


share/availability/discovery/discovery.go
```diff
@@ -183,23 +199,42 @@ func (d *Discovery) ensurePeers(ctx context.Context) {
...
+			d.handlePeerFound(ctx, peer, cancel)
+		}(p)
+	}
+
+	// Wait for all goroutines to finish, then cancel the context
+	go func() {
+		wg.Wait()
+		cancel()
+	}()
+}
```
Wondertan, line 236: The peers channel is returned immediately and closed when either query ends, or the context to FindPeers is canceled. The query can take minutes, especially when the network is significant, and many peers are undialable( e.g., in DHT server mode but behind a NAT). 

So if we obtain all the necessary connections, let's cancel FindPeers as we cancel all outgoing connections once we no longer need them. In other words, current logic works well when FindPeers ends earlier than connections are obtained, but not vice-versa.

share/availability/discovery/discovery.go
```diff
@@ -183,23 +199,42 @@ func (d *Discovery) ensurePeers(ctx context.Context) {
...
+			defer wg.Done()
+			d.handlePeerFound(ctx, peer, cancel)
+		}(p)
+	}
+
+	// Wait for all goroutines to finish, then cancel the context
+	go func() {
+		wg.Wait()
+		cancel()
+	}()
```
Wondertan, line 235: I wondert why this would be necessary. This logic makes cancel only after the FindPeers query is finished(peers chan closed) and all the peers are handled with any result(wg.Wait). Nothing is technically left to cancel in this case, as everything is finished.

I would keep the wg and make the `findPeers` block on it so that we don't start new `findPeers` before the previous one hasn't finished yet. Also, I would remove the cancel after wg, as it effectively noop. The only issue is that wg.Wait does not respect the context, but this can be fixed with errgroup

distractedm1nd: The only reason the waitgroup exists is to avoid a context leak. We leak context in case the amount of peers returned from FindPeers doesn't result in a number of connections >= peerLimit

share/availability/discovery/discovery.go
```diff
@@ -96,26 +104,35 @@ func (d *Discovery) WithOnPeersUpdate(f OnUpdatedPeers) {
...
+		}
 		return
 	}
 
-	err = d.connector.Connect(ctx, peer)
+	if uint(d.set.Size()) >= d.peersLimit {
+		log.Debugw("peer limit reached", "count", d.set.Size(), "peer", peer.ID)
+		cancelOngoing()
+	}
+	err = d.set.TryAdd(peer.ID)
```
walldiss, line 127: I think previously add was before connect to mitigate potential shadow write issue. Example of such issue could this example order of operations:
1. (line114) Peer connected
2. (line182) disconnect event happen
3. (line183) Peer is not in the limited set, so noop. 
4. (line127) Peer added to the pool

If this order happens, peer will land in limited set even it is already disconnected. With previous approach (add->try connect-> remove if fail) this would not be a problem.

There is clearly a benefit of doing add after connect to, but could you please resolve shadow write?

share/availability/discovery/discovery.go
```diff
@@ -96,26 +104,35 @@ func (d *Discovery) WithOnPeersUpdate(f OnUpdatedPeers) {
...
+		if errors.Is(err, context.Canceled) || errors.Is(err, routing.ErrNotFound) {
+			d.connector.RemoveBackoff(peer.ID)
+		}
 		return
 	}
 
-	err = d.connector.Connect(ctx, peer)
+	if uint(d.set.Size()) >= d.peersLimit {
+		log.Debugw("peer limit reached", "count", d.set.Size(), "peer", peer.ID)
+		cancelOngoing()
```
walldiss, line 125: Do you think it is reasonable to cancel establishing other connection if we almost done all work for it by this point? I feel it could be very wasteful approach. Could you consider to calculate amount of peers you need before establish ing connection and do not connect to peers you would not require later?

share/availability/discovery/discovery.go
walldiss: I don't get the leak either. By the time wg.Wait() returns, there are no any other `d.handlePeerFound` call running in goroutines.

share/availability/discovery/discovery.go
Wondertan: > Do you think it is reasonable to cancel establishing other connection if we almost done all work for it by this point? 

Yes. The connect inside does multiple things, and it's not just Dial. Internally, it also does DHT queries through RoutedHost.

> Could you consider to calculate amount of peers you need before establish ing connection and do not connect to peers you would not require later?

This could work, but it's less responsive. Another idea we explored is batching. You know how many connections you need, and you discover this amount of peers through FindPeers, however, getting N peers through FindPeers may take a while, and peers are discovered one by one with intervals in between. Therefore, with the batching approach, we waste time reaching N discovered peers, when we could already start connecting to discover peers immediately.

share/availability/discovery/discovery.go
Wondertan: I will show an alternative soon in this branch

share/availability/discovery/discovery.go
Wondertan: That is an interesting point and we will think about it. Maybe you have some ideas as well

share/availability/discovery/discovery.go
walldiss: If we already half way to establish the extra connection, maybe just let it settle and have few extra connection in discovery, since we already done so much work to make it?

walldiss: > However, getting N peers through FindPeers may take a while, and peers are discovered individually with intervals in between. Therefore, with the batching approach, we waste time reaching N discovered peers, when we could have already connected to discover peers immediately. The other problem with batching is that you would need to restart FindPeers if some of the N peers are unreachable

I'm suggesting to initiate FindPeers only when amount_connected + amount_connecting < target_amount. One of the possible solutions to implement it would be to trigger FindPeers in event based manner in another routine.

Wondertan: > If we already half way to establish the extra connection, maybe just let it settle and have few extra connection in discovery, since we already done so much work to make it?

Actually, I tend to agree with this. So we would cancel only the FindPeers once we got N connections and leave the in-proc connection untouched 👍🏻. It does not have to have a few more connections.

Wondertan: > I'm suggesting to initiate FindPeers only when amount_connected + amount_connecting < target_amount. One of the possible solutions to implement it would be to trigger FindPeers in event based manner in another routine

If we allow only one FindPeers at a time, we wouldn't need to track the amount of connecting.

vgonkivs: What about introducing a buffer of peers? So, we can connect N(where N=available size of peerSet) amount of peers at a time. All other discovered peers will be added to the buffer and in case if we failed at some point, we will get a peer from the buffer.

Wondertan: Does the buffer contain connected/connecting peers?

share/availability/discovery/discovery.go
Wondertan: Well, I think the only correct solution is to keep logic that adds and removes peers synchronized and, ideally, in the same thread. There always will be such races between Add and Remove depending on where event happens.

Wondertan: This example is actually one of the reasons behind Rob Pike's quote "Don't communicate by sharing memory; share memory by communicating

share/availability/discovery/discovery.go
vgonkivs: Discovered recently.

walldiss: > If we allow only one FindPeers at a time, we wouldn't need to track the amount of connecting.
Thats is part of my suggestion. Have single routine that is responsible to find enough peers. It could be trigger `FindPeers` by external notification(by channel) when amount_connected + amount_connecting drops below target_amount (could be done from disconnect event subscriber). Essentially communicate the state.

share/availability/discovery/discovery.go
walldiss: i didn't think long, so my suggestion could be not optimal. One solution to this corner case that comes to my mind is to track peers that are currently trying to connect in another set.

nodebuilder/share/config.go
```diff
@@ -36,6 +38,7 @@ type Config struct {
 func DefaultConfig() Config {
 	return Config{
 		PeersLimit:        5,
+		DialTimeout:       time.Second * 60,
```
walldiss, line 41: Is there any possible problems of dialing takes longer than discovery interval? Like triggering extra discovery for example 

Do we have any idea behind timeout having this value? If so would be nice to add comment section explaining why this value is default.

share/availability/discovery/discovery.go
vgonkivs: > We want to connect immediately so each peer we discover instantly goes into 'connecting' state

I got this. But I'm proposing to connect only peerSet.Size() at a time. All other peers will wait. As we now have 1 minute dealing timeout will not take much time.

share/availability/discovery/discovery.go
```diff
@@ -183,24 +209,42 @@ func (d *Discovery) ensurePeers(ctx context.Context) {
...
+	for p := range peers {
+		peer := p
+		wg.Go(func() error {
+			// pass the cancel so that we cancel FindPeers when we connected to enough peers
+			d.handlePeerFound(findCtx, peer, findCancel)
+			return nil
+		})
+	}
+	// we expect no errors
+	_ = wg.Wait()
```
walldiss, line 255: Casual `sync.Waitgroup` seems to be sufficient here, since `errgroup` never gets an error.

share/availability/discovery/discovery.go
Wondertan: I uploaded the solution

share/availability/discovery/discovery.go
Wondertan: I explained the reason for the errorgroup in the comment above

nodebuilder/share/config.go
Wondertan: > Is there any possible problems of dialing takes longer than discovery interval? Like triggering extra discovery for example

No, as long as we keep waitgroup, we won't start a new discovery until we stop the previous one

This value is exactly the same as it is in libp2p for the actual low-level dial. Have no reasoning for why it is 1min, could be two as well. It should be just big enough for the Connect so that It can perform internal address discoveries.

share/availability/discovery/discovery.go
walldiss: I don't get it. There is no error inside error group callback. If you only need to wait for all routines to finish, why not use `sync.Waitgroup`?

Wondertan: > 	// we use errgroup as it obeys the context

The regular WaitGroup does not do this

share/availability/discovery/discovery.go
Wondertan: Ok, so this idea might bet useful in case we want to minimize going beyond the limit with possible tradeoff where we defer connecting.

nodebuilder/share/config.go
Wondertan: > Does the waitgroup sync means that if one peer fails to connect, discovery will need to wait until all connection attempts finish before looking for replacement peer?

No, because FindPeer returned channel still gives us new values.

share/availability/discovery/discovery.go
Wondertan: This buffering can be implemented over errgroup as it provides rate limiting API

Wondertan: Ok, made it

share/availability/discovery/discovery.go
```diff
@@ -183,24 +244,44 @@ func (d *Discovery) ensurePeers(ctx context.Context) {
...
+		log.Debugw("at peer limit, skipping FindPeers", "size", d.set.Size())
+		return
+	}
+
+	findCtx, findCancel := context.WithCancel(ctx)
+	defer findCancel()
+
+	peers, err := d.disc.FindPeers(findCtx, topic)
+	if err != nil {
+		log.Warn(err)
```
renaynay, line 263: add some extra detail

renaynay (dismissed):
`findCancel` and `cancelFind` - lets pick one

share/availability/discovery/discovery.go
```diff
@@ -96,26 +109,42 @@ func (d *Discovery) WithOnPeersUpdate(f OnUpdatedPeers) {
 
 // handlePeersFound receives peers and tries to establish a connection with them.
 // Peer will be added to PeerCache if connection succeeds.
-func (d *Discovery) handlePeerFound(ctx context.Context, topic string, peer peer.AddrInfo) {
-	if peer.ID == d.host.ID() || len(peer.Addrs) == 0 || d.set.Contains(peer.ID) {
+func (d *Discovery) handlePeerFound(ctx context.Context, peer peer.AddrInfo, cancelFind context.CancelFunc) {
```
Wondertan, line 112: Ok, so I verified and `peer.AddrInfo` here, and it should always contain empty addresses and only the ID filled is set. 
All the references to [this](https://github.com/libp2p/go-libp2p-kad-dht/blob/master/providers/providers_manager.go#L242) do only put the ID, but not the actual addresses. Therefore, the RoutedHost must find the address behind the peers here, and we should have a higher Connect timeout.

share/availability/discovery/discovery.go
```diff
@@ -183,24 +221,42 @@ func (d *Discovery) ensurePeers(ctx context.Context) {
...
+		return
+	}
+
+	// we use errgroup as it obeys the context
+	wg, findCtx := errgroup.WithContext(ctx)
+	for p := range peers {
+		peer := p
+		wg.Go(func() error {
+			// pass the cancel so that we cancel FindPeers when we connected to enough peers
+			d.handlePeerFound(findCtx, peer, findCancel)
```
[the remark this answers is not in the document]
Wondertan: Yeah, feels like it wrong

share/availability/discovery/discovery.go
```diff
@@ -96,27 +104,46 @@ func (d *Discovery) WithOnPeersUpdate(f OnUpdatedPeers) {
...
+
+	ctx, cancel := context.WithTimeout(ctx, d.dialTimeout)
+	defer cancel()
+
+	err := d.connector.Connect(ctx, peer)
 	if err != nil {
-		log.Debug(err)
+		// we don't want to add backoff when the context is canceled.
+		if errors.Is(err, context.Canceled) || errors.Is(err, routing.ErrNotFound) {
+			d.connector.RemoveBackoff(peer.ID)
```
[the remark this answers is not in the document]
distractedm1nd: I had this at first, it's very spammy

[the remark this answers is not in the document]
distractedm1nd: There are a lot more rejected peers than you'd think!

Wondertan:
I will also write tests over the weekends and maybe refactor if enough inspiration

share/availability/discovery/set.go
```diff
@@ -40,30 +40,25 @@ func (ps *limitedSet) Size() int {
...
 
-// TryAdd attempts to add the given peer into the set.
+// Add attempts to add the given peer into the set.
 // This operation will fail if the number of peers in the set is equal to size.
-func (ps *limitedSet) TryAdd(p peer.ID) error {
+func (ps *limitedSet) Add(p peer.ID) error {
 	ps.lk.Lock()
-	defer ps.lk.Unlock()
 	if _, ok := ps.ps[p]; ok {
 		return errors.New("share: discovery: peer already added")
```
vgonkivs, line 48: hello, I'm a deadlock

share/availability/discovery/set.go
```diff
@@ -40,30 +40,25 @@ func (ps *limitedSet) Size() int {
 	return len(ps.ps)
 }
 
-// TryAdd attempts to add the given peer into the set.
+// Add attempts to add the given peer into the set.
 // This operation will fail if the number of peers in the set is equal to size.
-func (ps *limitedSet) TryAdd(p peer.ID) error {
+func (ps *limitedSet) Add(p peer.ID) error {
 	ps.lk.Lock()
-	defer ps.lk.Unlock()
```
vgonkivs, line 47: hello, I'm a deadlock

share/availability/discovery/discovery.go
```diff
@@ -57,6 +69,7 @@ func NewDiscovery(
 	h host.Host,
 	d discovery.Discovery,
 	peersLimit uint,
+	dialTimeout,
```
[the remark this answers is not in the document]
Wondertan: I agree that it is bloat

share/availability/discovery/set.go
Wondertan: Good catch

share/availability/discovery/discovery.go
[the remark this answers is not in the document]
Wondertan: Made it a constant

share/availability/discovery/discovery.go
```diff
@@ -161,17 +181,49 @@ func (d *Discovery) ensurePeers(ctx context.Context) {
...
+					log.Debugw("removed peer from the peer set",
+						"peer", evnt.Peer, "status", evnt.Connectedness.String())
+				case network.Connected:
+					peerID := evnt.Peer
+					d.connectingLk.Lock()
+					cancelFind, ok := d.connecting[peerID]
+					d.connectingLk.Unlock()
+					if !ok {
+						continue
 					}
```
walldiss, line 204: Do we skip because it could be light client connected?

Wondertan: Yes.

share/availability/discovery/discovery.go
```diff
@@ -56,27 +84,22 @@ type OnUpdatedPeers func(peerID peer.ID, isAdded bool)
...
+		connecting:     make(map[peer.ID]context.CancelFunc),
 	}
 }
 
 func (d *Discovery) Start(context.Context) error {
-	ctx, cancel := context.WithCancel(context.Background())
-	d.cancel = cancel
-
-	go d.ensurePeers(ctx)
+	d.ctx, d.cancel = context.WithCancel(context.Background())
```
walldiss, line 101: Nit: there is no need to store ctx

share/availability/discovery/discovery.go
```diff
@@ -183,26 +251,50 @@ func (d *Discovery) ensurePeers(ctx context.Context) {
...
+func (d *Discovery) findPeers(ctx context.Context) {
+	if d.set.Size() >= d.set.Limit() {
+		log.Debugw("at peer limit, skipping FindPeers", "size", d.set.Size())
+		return
+	}
+
+	findCtx, findCancel := context.WithCancel(ctx)
+	defer findCancel()
+
+	peers, err := d.disc.FindPeers(findCtx, topic)
```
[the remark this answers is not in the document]
Wondertan: Reasonable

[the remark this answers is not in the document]
distractedm1nd: I tried this a few ways, a simple recursive call with a goroutine looked the cleanest but didn't actually work well. Please review the latest commit to see an approach that worked as expected (although, it doesnt restart until actually trying to connect... but I couldn't find a good way to do this that didn't spam while waiting for connections)

```
2023-04-27T09:22:25.632+0200	DEBUG	share/discovery	discovery/discovery.go:270	finding peers	{"remaining": 50}
2023-04-27T09:22:25.768+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWB2jFi9eAPhhpVRHNpHe535LVwxgqDtpkM5NDorGA5Y1A"}
2023-04-27T09:22:25.771+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWBASepmTQtATM196L1wWH6kVd4TBgpUN57koyfjT2vheX"}
2023-04-27T09:22:25.772+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWAkMsu2Um5YbEEUhXKNFx277ER5aDEtVrjb3RsdPa5SJv"}
2023-04-27T09:22:25.775+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWAM4QADEHeZGttANbzKs2wDAWPa8LrGhcc1HjoE4my8uT"}
2023-04-27T09:22:25.775+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWBc9nKfmKJZKqkto9UHvQce1ZZZy3MoNdZzDQpgi2SFkD"}
2023-04-27T09:22:25.776+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWAMKhsAMRy8XX74BC1hwL5hS7u6Zq62v9bZ5sYnRQUrao"}
2023-04-27T09:22:25.776+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWCqB6gS7oyYYYNbxoK2FQYmdW7EGcmCVWPwk5sraH19W1"}
2023-04-27T09:22:25.780+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWAUUt2C2i2owYgk7tox45tnJF6urEDBPgiH7ZmxgojbmW"}
2023-04-27T09:22:25.780+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWAuemDk5eDa2wjY4LukoNkhH1ZN2DrVdLPb3SLCMVGRd9"}
2023-04-27T09:22:25.803+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWCshF2rNvcMJREScaP69JtHjSGY4aySpXcpbtuSgAGWC8"}
2023-04-27T09:22:25.803+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWDAQ3pHz1tdsbR7cAMZW87QLZfptHNCLK3VFQ5TV7mrES"}
2023-04-27T09:22:25.803+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWBTFMz9doATz6gxb9VUZ5d2iJ1udBfJCkho6ixgeu6nhE"}
2023-04-27T09:22:25.803+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWD1oyJtzc3dhfUK2DZ1SJzPwLson92fnGBckiXpGixxBg"}
2023-04-27T09:22:25.804+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWDWJUWz3zRfB2bfXJxK7WPSM9QsDDPeME1BYJqq2h9oTh"}
2023-04-27T09:22:25.804+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWB7GvRuTWm6DUd4pt5VTRfYfciWKLkUpCKrCtJnGNKEv7"}
2023-04-27T09:22:25.804+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWBct27GArLQoZZLwnShT7DnaXmNZpEiK8oQPvNuZrptp8"}
2023-04-27T09:22:25.805+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWCT9aCgwDVt8D5UGiiDhsJHNhbuif1utdxRBgLfTuWwvC"}
2023-04-27T09:22:25.805+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWARYrY8eZFZx3BwsUNbXDiddXDNN3a2PgHPcLh2cPwusc"}
2023-04-27T09:22:25.805+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWAdXB5LHZmWd3wNjJgLPBkVNF7WcgFvLKpzsK9f8sPB7N"}
2023-04-27T09:22:25.805+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWAktyreXmqQN6iLr1ausLgSJFU7rSdbT9vMXemku8W9GW"}
2023-04-27T09:22:25.805+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWD13hWGqHXobDggdrFk8UPSBDRGHfheDzwECLsiK8HNFD"}
2023-04-27T09:22:25.844+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWC2NnJKTnx3fd1d1rbVL5Kaa3U66L2XeKMxYjmamPTXut"}
2023-04-27T09:22:25.853+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWCmkFRsUipHAbdnEfTnB5CKAxdKGDMghriPv5MJsboRp8"}
2023-04-27T09:22:25.854+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWDQjXzsvXzrLBZXbJ3jxey58LLeMXEzjC34UDKmyZqkBr"}
2023-04-27T09:22:25.854+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWDHmxrf5AJYVq28YfzAaq1yuMedE9avrGAbSp5JbKeWW9"}
2023-04-27T09:22:25.861+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWCAMGzWrcnbPradgkJEe2yTEWNYi5LPM4YqohDxDdE9Ng"}
2023-04-27T09:22:25.906+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWC1M3zN1eJ5Sh7XWqgvK8M5LcmNQC5wkLJSDNsGceJj8o"}
2023-04-27T09:22:25.906+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWC6Vg18RW56jjhmCNQEtWg6iyAmPyiy3jcC2WPKKiCyKZ"}
2023-04-27T09:22:25.912+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWBbc9ixMAH4SzF3EH6SwSfo8gRW1bVrDUX19oxC4DaRbz"}
2023-04-27T09:22:25.978+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWBCfAdTYhtpuzbKqnc5WM3qh7Zkp3DmjrbAwgTsgiKQPW"}
2023-04-27T09:22:26.121+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWC5gfx3LmvRS96jC7QpXRg5nF2pextta854LVzVRtoRnc"}
2023-04-27T09:22:26.187+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWEF1cTm8eCevybhxicYWDuYWHsa2YgpV3VeS667XhFcK9"}
2023-04-27T09:22:26.187+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWEGgFtcvfmn749UDYFavU2n5DPrNnnc3PUog9KVf4kppr"}
2023-04-27T09:22:26.237+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWEkFUj5rrWoeQL9stJR6AkC2F1MBT4MUeiqaoi8ZUAADA"}
2023-04-27T09:22:26.240+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWFC3v38kFFJYWVNjR92iX962YpSVpAbGsDwYcgVqvTWk8"}
2023-04-27T09:22:26.241+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWJQwLTJJzX7N88jkvMQm49Bajpaey87AQE9jNQuY1Js7U"}
2023-04-27T09:22:26.284+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWEXj2bfecMMnWnbfGPQJjPozGjH3JZdaNSMmgwQWDGUWd"}
2023-04-27T09:22:26.285+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWEvGQbiiBWcZv71FS5kwkh5g5ZtYJXQYoDgNM23w3jmam"}
2023-04-27T09:22:26.285+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWEvvjKpd1Lckiujpt11udm8gtowMV8fmaf6vdFPZ7dHS4"}
2023-04-27T09:22:26.312+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWFFNrfN6JBbvy3ub4Nsy1hP9DJk8dVDieaE4s4EBTPdkf"}
2023-04-27T09:22:26.320+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWFKiR9qk2P9f2bkHUKCjdNUvexSUKU6eacQPzG3aCfJf2"}
2023-04-27T09:22:26.322+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWFu69RanHPBzSb8JWE8vNNvM1doiDoost7zQAtjCwEhV7"}
2023-04-27T09:22:26.322+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWFvZxuvPibmMKEW2WHGCzZBmPfa3AefKDvUtjWvsAcadz"}
2023-04-27T09:22:26.333+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWFNPkdnk2mHoZeecJ41L6KUqrqsQZdKgc7NyHtBMLfTvd"}
2023-04-27T09:22:26.333+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWFd3bjJycHnid6f3o2939U47fMtoxw9ReqJTDDUKLr9bf"}
2023-04-27T09:22:26.357+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWFKNTDwLrbbZcdouc4qY32oCbw6BfXMJXaT3fUJDyD9N4"}
2023-04-27T09:22:26.357+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWGDN4XBmcfbeQumByWPFKTVGeG76RVtrTL5a8VXhXcbf9"}
2023-04-27T09:22:26.358+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWGbKAMueVfwiY2p43Qb3gquaXFKkezfSFoiuNmmnVixXq"}
2023-04-27T09:22:26.358+0200	DEBUG	share/discovery	discovery/discovery.go:270	finding peers	{"remaining": 2}
2023-04-27T09:22:26.360+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWGtDBVQJycdaWtSbeMV9VBMLA2uDeei9VMUxHWypy12kF"}
2023-04-27T09:22:26.361+0200	DEBUG	share/discovery	discovery/discovery.go:227	added peer to set	{"id": "12D3KooWAY38q5safauiJ7XqwnUUi34nJUCKtJ4iXMpW8ueM6hVc"}
2023-04-27T09:22:26.361+0200	INFO	share/discovery	discovery/discovery.go:235	soft peer limit reached	{"count": 50}
```

share/availability/discovery/set.go
```diff
@@ -44,8 +44,11 @@ func (ps *limitedSet) Size() int {
 }
 
 // Add attempts to add the given peer into the set.
-func (ps *limitedSet) Add(p peer.ID) {
+func (ps *limitedSet) Add(p peer.ID) (added bool) {
 	ps.lk.Lock()
+	if _, ok := ps.ps[p]; ok {
+		return false
+	}
```
Wondertan: Everywhere we do Add, we also do Contains before, so returning bool is unnecessary

share/availability/discovery/discovery.go
```diff
@@ -242,6 +242,10 @@ func (d *Discovery) ensurePeers(ctx context.Context) {
 	defer t.Stop()
 	for {
 		d.findPeers(ctx)
+		if d.set.Size() < d.set.Limit() {
+			// rerun discovery is amount of peers didn't reach the limit
+			continue
+		}
```
Wondertan, line 265: From discussion with Ryan I get the impression that you decided against that

share/availability/discovery/discovery.go
```diff
@@ -273,12 +275,18 @@ func (d *Discovery) findPeers(ctx context.Context) {
...
 	var amount int
 	for {
+		ticker.Reset(findPeersStuckWarnDelay)
 		select {
 		case <-findCtx.Done():
 			log.Debugw("found enough peers", "amount", d.set.Size())
 			return
+		case <-ticker.C:
+			log.Error("wasn't able to find new peers for long time")
+			continue
```
Wondertan: Does this happen in practise?

share/availability/discovery/discovery.go
```diff
@@ -256,6 +264,7 @@ func (d *Discovery) ensurePeers(ctx context.Context) {
 			continue
 		}
 
+		//TODO: drain the channel, because it could be filled already or use timer with proper drain
```
Wondertan, line 267: Ticked does not have to be drained, as per docs. It accounts for slow reader

share/availability/discovery/set.go
walldiss: Contains is happening outside of mutex lock scope, so it is not a safe to assume peer was not added inbetween operations

share/availability/discovery/discovery.go
walldiss: Previous implementation had a lot of issues. This one is doing good, because of other changes to errorgroup. Also in reality, when discovery is launched it is only able to find local host for few seconds, and this check allows it to find legit peers as soon as possible.

share/availability/discovery/discovery.go
walldiss: Very unlikely, as you probably expected. Bit It could in certain conditions:
- if there are not enough good peers in network
- if discovery reached internal discovery limit and we still didn't found enough peers
- if target limit set high (like >100) it will happen often (not our real case)

Ideally we should not see it, but if we do, it means discovery is unable to give new peers to connect. I want this case to be explicitly logged, so if it happen we get reports and investigate.

share/availability/discovery/discovery.go
walldiss: Yeah, but check the ticker code. If channel was filled it won't be drained on Reset. Doc is not what you expect, because Reset operation is not meant to make a fresh new ticker that will yield first time after delay. It just updates the tick time, while persisting current ticker state. For our usecase we need to drain (could be my next commit)

walldiss: Example code for the ticker that is not properly reset:
```
ticker := time.NewTicker(2 * time.Second)
for {
time.Sleep(2* time.Second )
ticker.Reset(2 * time.Second)

select {
case <-ticker.C:
	fmt.Println("ticker fired too early")
case <-time.After(time.Second):
	fmt.Println("ticker didn't fire too early")
}
}
```

Wondertan:
Making the test non-flakey is so annoying

share/availability/discovery/set.go
Wondertan: Yeah, so let remove the contains as unnecessary check

Wondertan:
There is still one flakey case left; please don't merge until it's fixed.

For some reason, we may not receive disconnect PeerUpdate

share/availability/discovery/discovery.go
```diff
@@ -348,6 +352,9 @@ func (d *Discovery) Advertise(ctx context.Context) {
 
 			select {
 			case <-timer.C:
+				if !timer.Stop() {
+					<-timer.C
+				}
```
Wondertan: @walldiss, this is not needed before Reset in this case, as we just drained C

Additionally, this change introduces more flakes which I am in the process of debugging. Removing this exact locs helped.

walldiss: You are right. Timer is already stoped and channel drain at that point

share/availability/discovery/discovery.go
```diff
@@ -96,35 +117,74 @@ func (d *Discovery) WithOnPeersUpdate(f OnUpdatedPeers) {
...
+		d.connector.Backoff(peer.ID)
+	case network.NotConnected:
+		d.connectingLk.Lock()
+		if _, ok := d.connecting[peer.ID]; ok {
+			d.connectingLk.Unlock()
+			log.Debug("skip handle: connecting to the peer in another routine")
+			return false
+		}
+		d.connecting[peer.ID] = cancelFind
+		d.connectingLk.Unlock()
```
walldiss, line 161: @Wondertan  Looks like there is a shadowed read issue that could happen here. Here is example of events sequence that would lead to bug:

1. check connectivity state in `handlePeer` -> NotConnected 
2. connect event in sub handler -> skipped, since peer not in `connecting` pool
3. added to connecting pool in highlighted code.
4.??? (no connecting event, since it is already skipped)

To fix it we need to add to connecting pool before connectivity check. Do you mind if I make a quick fix?

Wondertan: I don't mind, please do

share/availability/discovery/backoff.go
```diff
@@ -12,9 +12,16 @@ import (
 )
 
 // gcInterval is a default period after which disconnected peers will be removed from cache
```
vgonkivs, line 14:
```suggestion
```

share/availability/discovery/backoff.go
```diff
@@ -12,9 +12,16 @@ import (
...
-const gcInterval = time.Hour
+const (
+	gcInterval = time.Minute
```
vgonkivs, line 16:
```suggestion
       // gcInterval is a default period after which disconnected peers will be removed from cache
	gcInterval = time.Minute
```

share/availability/discovery/set.go
```diff
@@ -14,12 +13,12 @@ type limitedSet struct {
...
 
-	limit    uint
+	limit    int
 	waitPeer chan peer.ID
 }
 
 // newLimitedSet constructs a set with the maximum peers amount.
-func newLimitedSet(limit uint) *limitedSet {
+func newLimitedSet(limit int) *limitedSet {
 	ps := new(limitedSet)
```
vgonkivs, line 22: if you've changed from uint to int, then I'd propose adding a sanity check for` limit > 0`

vgonkivs (changes requested):

share/availability/discovery/set.go
Wondertan: We allow -1 as a way to disable discovery

vgonkivs: May be we should also add comment here, that explains 0 limit case?

Wondertan: Making it uint again

share/availability/discovery/backoff.go
```diff
@@ -36,48 +43,55 @@ func newBackoffConnector(h host.Host, factory backoff.BackoffFactory) *backoffCo
...
-	defer b.cacheLk.Unlock()
-	cache := b.connectionData(p)
-	cache.backoff.Reset()
-	cache.nexttry = time.Now().Add(cache.backoff.Delay())
+	cache, ok := b.cacheData[p]
+	b.cacheLk.Unlock()
+	if ok && time.Now().Before(cache.nexttry) {
+		return true
+	}
+	return false
```
renaynay, line 91: can just return the statement itself

return ok && time.Now etc etc

share/availability/discovery/set.go
renaynay: ahhh i dislike but its ok.

share/availability/discovery/discovery.go
```diff
@@ -11,42 +11,65 @@ import (
...
-	// onUpdatedPeers will be called on peer set changes
+	params Parameters
+
+	set            *limitedSet
+	host           host.Host
+	disc           discovery.Discovery
+	connector      *backoffConnector
 	onUpdatedPeers OnUpdatedPeers
 
+	triggerDisq chan struct{}
```
renaynay, line 71: can we just name triggerDisc

vgonkivs (changes requested):

share/availability/discovery/discovery.go
```diff
@@ -94,36 +113,65 @@ func (d *Discovery) WithOnPeersUpdate(f OnUpdatedPeers) {
...
-		log.Debug(err)
-		d.set.Remove(peer.ID)
-		return
+	switch d.host.Network().Connectedness(peer.ID) {
+	case network.Connected:
+		d.connector.Backoff(peer.ID) // we still have to backoff the connected peer
+	case network.NotConnected:
+		err := d.connector.Connect(ctx, peer)
+		if err != nil {
+			log.Debugw("unable to connect", "err", err)
```
renaynay, line 148: log peer as well

share/availability/discovery/discovery.go
```diff
@@ -11,42 +11,65 @@ import (
...
 }
 
+type Parameters struct {
+	// PeersLimit defines the soft limit of FNs to connect to via discovery.
+	// Set 0 to disable.
+	PeersLimit uint
+	// DiscoveryRetryTimeout is an interval between discovery attempts
+	// when we discovered lower than PeersLimit peers.
+	// Set -1 to disable.
+	DiscoveryRetryTimeout time.Duration
```
walldiss, line 41: Nit: This setting looks doesn't seem to be very usefull for node runners and could be private to allow to be set only onside code(for tests).

share/availability/discovery/discovery.go
```diff
@@ -11,42 +11,65 @@ import (
...
+	// AdvertiseInterval is a interval between advertising sessions.
+	// Set -1 to disable.
+	// NOTE: only full and bridge can advertise themselves.
+	AdvertiseInterval time.Duration
+}
+
+func DefaultParameters() Parameters {
+	return Parameters{
+		PeersLimit:            5,
+		DiscoveryRetryTimeout: time.Second * 1,
```
walldiss, line 51: Nit: `something * 1`?

walldiss (approved):

renaynay (approved):

share/availability/discovery/discovery.go
walldiss: the logger is enhanced by peer field by that point.

share/availability/discovery/discovery.go
```diff
@@ -235,3 +356,27 @@ func (d *Discovery) Advertise(ctx context.Context) {
 func (d *Discovery) Peers(ctx context.Context) ([]peer.ID, error) {
 	return d.set.Peers(ctx)
 }
+
+func (p Parameters) withDefaults() Parameters {
+	def := DefaultParameters()
+	if p.PeersLimit == 0 {
+		p.PeersLimit = def.PeersLimit
+	}
```
Wondertan, line 364: We should not override it. 0 is an allowed value and needed to disable the limit

share/availability/discovery/discovery.go
```diff
@@ -153,11 +155,11 @@ func (d *Discovery) handlePeerFound(ctx context.Context, peer peer.AddrInfo) boo
 	}
 
 	if !d.set.Add(peer.ID) {
-		log.Debugw("peer is already in discovery set", "peer", peer.ID)
+		log.Debug("peer is already in discovery set")
 		return false
 	}
 	d.onUpdatedPeers(peer.ID, true)
-	log.Debugw("added peer to set", "peer", peer.ID)
+	log.Debug("added peer to set")
```
renaynay, line 163: why did you remove logging the peer IDs?

renaynay (dismissed):

share/availability/discovery/discovery.go
walldiss: log has been shadowed by field log, that has peer field included, in the beginning of the func.
Reminds me one of the review comments to one of my PRs, when I did introduced field log, but didn't shadowed it, instead named it `logger` instead of common `log`

walldiss: Pls take a look at renamed `log` to `logger` version. I thin it should be less confusing

renaynay (approved):

share/availability/discovery/discovery.go
```diff
@@ -317,46 +308,51 @@ func (d *Discovery) findPeers(ctx context.Context) bool {
...
-				timer.Reset(d.params.AdvertiseInterval)
-				continue
-			case <-ctx.Done():
-				return
-			}
+			logger.Debugw("unable to connect", "err", err)
+			return false
 		}
+	default:
+		panic("unknown connectedness")
```
renaynay, line 340: aren't there two more kinds of connectedness ? feels weird to panic here.

renaynay (approved):
only one comment re panic

share/availability/discovery/discovery.go
Wondertan: We had an argument with Vlad

share/availability/discovery/discovery.go
```diff
@@ -247,7 +189,56 @@ func (d *Discovery) ensurePeers(ctx context.Context) {
...
+	// subscribe on EventBus in order to catch disconnected peers and restart
+	// the discovery. We specify a larger buffer size for the channel where
+	// EvtPeerConnectednessChanged events are sent (by default it is 16, we
+	// specify 32) to avoid any blocks on writing to the full channel.
+	sub, err := d.host.EventBus().Subscribe(&event.EvtPeerConnectednessChanged{}, eventbus.BufSize(eventbusBufSize))
+	if err != nil {
+		log.Error(err)
+		return
+	}
+	defer sub.Close()
```
walldiss: Would be nice if you init subscribtion in Discovery constractor and provide it to runner routine. So if it error, we would catch it in FX. In current implementation there would be only error log line, but node would start normally and it is not good since this would break discovery.

share/availability/discovery/discovery.go
Wondertan: There are no more types of connectedness here, but if there are introduced and we missed that, it's a correctness bug and our testing pipeline should catch the panic so we can resolve

Wondertan (approved):

walldiss (approved):
Wow 58 commits! Discovery is so much cleaner now!

renaynay (approved):

distractedm1nd:
Blocked because @vgonkivs has previously requested changes

vgonkivs (approved):

PR #2125, opened 2023-04-24
chore(nodebuilder): timeouts, errors and logs

Wondertan:
Was originally for debugging purposes but actually solved the issue. The bootstrapper with a corrupted DAGstore state was not able to start quickly enough. 

Ideally, we should make this value configurable(https://github.com/celestiaorg/celestia-node/issues/2126)

### DIFF
--- a/nodebuilder/node.go
+++ b/nodebuilder/node.go
@@ -27,7 +27,7 @@ import (
 	"github.com/celestiaorg/celestia-node/nodebuilder/state"
 )
 
-const Timeout = time.Second * 15
+const Timeout = time.Minute
 
 var log = logging.Logger("node")
 
--- a/nodebuilder/node.go
+++ b/nodebuilder/node.go
@@ -2,6 +2,7 @@ package nodebuilder
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 	"time"
@@ -27,7 +28,7 @@ import (
 	"github.com/celestiaorg/celestia-node/nodebuilder/state"
 )
 
-const Timeout = time.Minute
+const Timeout = time.Second * 30
 
 var log = logging.Logger("node")
 
@@ -92,7 +93,10 @@ func (n *Node) Start(ctx context.Context) error {
 
 	err := n.start(ctx)
 	if err != nil {
-		log.Errorf("starting %s Node: %s", n.Type, err)
+		log.Debugf("error starting %s Node: %s", n.Type, err)
+		if errors.Is(err, context.DeadlineExceeded) {
+			return fmt.Errorf("node: failed to start within timeout(%s): %w", Timeout, errors.Unwrap(err))
+		}
 		return fmt.Errorf("node: failed to start: %w", err)
 	}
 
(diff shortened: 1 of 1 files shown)


Wondertan:
So this actually fixed bootstrapper, as it didn't have enough time to Start.

renaynay (dismissed):

nodebuilder/node.go
```diff
@@ -149,7 +161,11 @@ func (n *Node) Stop(ctx context.Context) error {
 func newNode(opts ...fx.Option) (*Node, error) {
 	node := new(Node)
 	app := fx.New(
-		fx.NopLogger,
+		fx.WithLogger(func() fxevent.Logger {
+			zl := &fxevent.ZapLogger{Logger: fxLog.Desugar()}
+			zl.UseLogLevel(zapcore.DebugLevel)
+			return zl
+		}),
```
walldiss, line 168: It is unclear from the PR, what component would take  `fxevent.Logger` as dependancy?  Also would you consider fx log level configurable?

nodebuilder/node.go
```diff
@@ -134,11 +143,14 @@ func (n *Node) Stop(ctx context.Context) error {
 
 	err := n.stop(ctx)
 	if err != nil {
-		log.Errorf("Stopping %s Node: %s", n.Type, err)
-		return err
+		log.Debugf("error stopping %s Node: %s", n.Type, err)
+		if errors.Is(err, context.DeadlineExceeded) {
+			return fmt.Errorf("node: failed to stop within timeout(%s): %w", Timeout, errors.Unwrap(err))
+		}
```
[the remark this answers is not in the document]
Wondertan: Sure, but I don't see any issues with accessing globals in this case. We anyway do that in this method already

nodebuilder/node.go
Wondertan: This is a special WithLogger option and it's not a Provide. The given logger is only used by fx itself and not by us. 

We can now control the log verbosity of fx with this change by `fx`.

nodebuilder/node.go
```diff
@@ -27,9 +30,12 @@ import (
 	"github.com/celestiaorg/celestia-node/nodebuilder/state"
 )
 
-const Timeout = time.Second * 15
+var Timeout = time.Second * 30
```
distractedm1nd, line 33: The issue is that this number will actually grow with the number of blocks in the DAGstore. It takes quite a while to start already. Maybe we can parallelize this process upstream, or find some other way to let the dagstore initialize before its used by the rest of node

nodebuilder/node.go
distractedm1nd: Does this change mean that fx logs will now show up when DEBUG logs are turned on?

Wondertan: Yes

nodebuilder/node.go
Wondertan: Thus the issue to make it configurable. Also, keep in mind that it took long on the the node that got kernel panic and had a corrupted DAGStore state

distractedm1nd (approved):

renaynay (approved):

PR #2136, opened 2023-04-27
Revert "deps: bump libp2p (#2127)"

Wondertan:
This reverts commit 82fbba833efbaca4dc3bbb9b73827d32223274db.

The update breaks the event system. It likely related to


distractedm1nd (approved):

Wondertan:
We would catch this if we put more attention into the Swamp tests

vgonkivs (approved):

PR #2139, opened 2023-04-27
fix(modp2p): increase ConnMgr defaults for Full and Bridge

Wondertan:
We have ~1000 peers on BSR. Full and Bridge nodes can handle such load as libp2p connections, especially QUIC are cheap. 
Low defaults cause aggressive connection trimming and trimmed peers have to rediscover new FN/BNs over and over. This eventually leaves peers with no FN/BNs to get data from, as we don't reconnect to previously discovered peers due to backoff.


renaynay (approved):
vibes here LGTM

PR #2153, opened 2023-05-02
fix(share/eds): recover Shards immediately

Wondertan:
Discovered after looking at CarStore sources(https://github.com/filecoin-saturn/L2-node/blob/f5c5ca0cd8cab27d83424e7a77943f900a4d4105/carstore/carstore.go#L177) I think it should aid long startup times by recovering shards immediately, rather than on restart/startup only.

### DIFF
--- a/share/eds/store.go
+++ b/share/eds/store.go
@@ -34,7 +34,8 @@ const (
 	indexPath      = "/index/"
 	transientsPath = "/transients/"
 
-	defaultGCInterval = time.Hour
+	defaultGCInterval  = time.Hour
+	maxRecoverAttempts = 1
 )
 
 var ErrNotFound = errors.New("eds not found in store")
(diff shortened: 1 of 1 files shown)


distractedm1nd:
I wasn't fully in the context with the concurrent map read/write panic, would this maybe cause a regression to this panic if the shards are being recovered in the bg?

Wondertan:
The panic only happens on startup when Bitswap receives requests for data in some shard, but Dagstore hasn't finished the Start yet. If this reduces the time it takes for DagStore to Start(I am not fully convinced), then this should minimize the chances for panic. 

The proper fix for the panic anyway is to ensure to introduce the Start method on Bitswap, so it never starts before EdsStore.

Wondertan:
This might not be necessary and needs further investigation, so I'm just drafting this for now. It can be the case that this PR targets the wrong recovery or recovery that does not take long time

distractedm1nd:
Closing this because it does not affect startup time as intended, shard failure channel is handled differently now, and also shard recovery itself makes no sense for us the way it is currently implemented.

We will implement our own shard recovery eventually, but dagstore recovery does nothing helpful and it is best to just remove the shard and reget from network for now, as we currently have planned.

PR #2154, opened 2023-05-02
fix(nodebuilder/tests): Add dummy trusted peers to satisfy fx

renaynay:
Fx requires trusted peers now so this fix is needed in the meantime to satisfy that.

Don't worry, these tests will be THOROUGHLY cleanussied in follow up PRs.


Wondertan (approved):

Wondertan:
My bad I didn't fix them on the go-header update PR

distractedm1nd (approved):

PR #2155, opened 2023-05-03
feat(share/discovery): add discovery metrics

walldiss:
## Overview

Adds metrics for discovery component 

Draft before discovery rewanp is merged


share/availability/discovery/discovery.go
```diff
@@ -283,14 +286,17 @@ func (d *Discovery) discover(ctx context.Context) bool {
 		drainChannel(ticker.C)
 		select {
 		case <-findCtx.Done():
+			d.metrics.observeFindPeers(true, d.set.Size() >= d.set.Limit())
 			return true
 		case <-ticker.C:
```
distractedm1nd, line 291: maybe we should add an event for this?

distractedm1nd (approved):
very clean

share/availability/discovery/metrics.go
```diff
@@ -0,0 +1,166 @@
...
+	if m == nil {
+		return
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), observeTimeout)
+	defer cancel()
+
+	m.discoveryResult.Add(ctx, 1,
+		attribute.Bool(discoveryFindCancledKey, canceled),
+		attribute.Bool(discoveryEnougPeersKey, isEnoughPeers))
+}
```
Wondertan, line 130: I saw we pass custom context every time we emot metric, but I don't understand why do this.
* We always have context in the places where we emit metric and it's legit context, which if canceled also cancels metric operations
* The metrics emissions are non-blocking according to the implementation, as it just buffers them and then asynchronously sends them, but if the underlying metrics emitor is blocking this won't respect the caller's context.
* It's just heavy to launch a new context for each internal call everywhere. New context means a new routine at least with timers inside which are not free as well.

So I would like to understand why we follow this pattern, it does not seem right to me.

share/availability/discovery/discovery.go
```diff
@@ -283,14 +286,17 @@ func (d *Discovery) discover(ctx context.Context) bool {
 		drainChannel(ticker.C)
 		select {
 		case <-findCtx.Done():
+			d.metrics.observeFindPeers(true, d.set.Size() >= d.set.Limit())
 			return true
```
Wondertan, line 290: A bit weird that we return true, but for the metric, we still do the comparison. Either we should emit true for both, or we do the comparison for both. I vote the former as we know at this point that it's true 100%

walldiss: Yeah, just rebasing artifact. Should be (true,true)

share/availability/discovery/metrics.go
Wondertan: From a sync discussion with @walldiss, we agreed to move without spawning new context just for metrics. The agreed approach is to use the user-given context. In case it is canceled, we will pass context.Background to still track canceled operations in metrics. 

@walldiss, also proposed to make a PR that changes this throughout codebase

share/availability/discovery/discovery.go
walldiss: Good idea, added this one.

Wondertan (approved):

Wondertan:
But needs to be rebases

derrandz (approved):

Wondertan (approved):

renaynay (approved):

Wondertan (approved):

distractedm1nd (approved):

PR #2156, opened 2023-05-03
fix(share/p2p)!: `share.ErrNamespaceNotFound` and integration into shrex-nd

### LINKED ISSUE #2145: bug(shares/GetSharesByNamespace): missing response type `ErrNamespaceNotFound`
In the current implementation, the response from a shrex-nd server for "Hey, my EDS store doesn't have that EDS, so I can't give you the ND" and "Hey, that EDS is in my store, and the data you are looking for doesn't exist in it" are the same. This causes many issues in the stack, leading to a timeout for `GetSharesByNamespace` calls, even when the correct response has been found. This will clearly require a protocol breaking change.

This must also be fixed in CascadeGetter, which should recognize the difference between `ErrNotFound`, which moves onto the next getter, and `ErrNamespaceNotFound`, which should return the data and error.

This means IPLDGetter should also return a `share.ErrNamespaceNotFound` for the same reason.

distractedm1nd:
This PR resolves #2145 . A followup PR may come that changes the response layer between the RPC and the internal GetSharesByNamespace call.


walldiss:
Please add support in ipld.Getter too

share/getters/store.go
```diff
@@ -123,10 +123,11 @@ func (sg *StoreGetter) GetSharesByNamespace(
 	shares, err = collectSharesByNamespace(ctx, blockGetter, root, nID)
 	if errors.Is(err, ipld.ErrNodeNotFound) {
 		// convert error to satisfy getter interface contract
-		err = share.ErrNotFound
+		err = share.ErrNamespaceNotFound
```
Wondertan, line 126: Q: It's possible to infer from just DAH whether there is namespaces data or not. Why aren't we doing this on client side?

distractedm1nd:
@walldiss There is support for IPLDGetter - its just through the collectSharesByNamespace util. You can see getter_test to verify it works correctly

share/getter.go
```diff
@@ -12,8 +12,13 @@ import (
 	"github.com/celestiaorg/rsmt2d"
 )
 
-// ErrNotFound is used to indicated that requested data could not be found.
-var ErrNotFound = errors.New("data not found")
+var (
+	// ErrNotFound is used to indicate that requested data could not be found.
+	ErrNotFound = errors.New("data not found")
+	// ErrNamespaceNotFound is used to indicate that the data does not include any shares from the
+	// requested namespace
```
walldiss, line 19:
```suggestion
	// ErrNamespaceNotFound is returned by GetSharesByNamespace when data for requested root does not include
	// any shares from the given namespace
```

walldiss (commented):
Need support in shrex-getter to stop retries after client returns `ErrNamespaceNotFound`

share/getters/store.go
distractedm1nd: This is a great idea, but not complete: The client side check can skip network requests if the nID isn't even found in the range of nIDs that the root covers. However, it can still be the case that an nID exists in the root that does not actually have any data (e.g. if a root contains the nids nID-1...nid+1)

I propose the following diff:
```diff
index c93a138e..805dae20 100644
--- a/share/getters/shrex.go
+++ b/share/getters/shrex.go
@@ -122,6 +122,13 @@ func (sg *ShrexGetter) GetSharesByNamespace(
                attempt int
                err     error
        )
+
+       // verify that the namespace could exist inside the roots before starting network requests
+       roots := filterRootsByNamespace(root, id)
+       if len(roots) == 0 {
+               return nil, share.ErrNamespaceNotFound
+       }
+
        for {
                if ctx.Err() != nil {
                        return nil, ctx.Err()
```

I don't want to include it in shrexnd client code, because the client should not care about such things, and also the logic already exists in the other getters as a quick-fail (through collectSharesByNamespace)

distractedm1nd:
Thank you @walldiss, somehow I forgot the most important part of the PR 😂

share/getter.go
distractedm1nd: done, but had to reformat

share/getters/utils.go
```diff
@@ -84,9 +84,9 @@ func collectSharesByNamespace(
 		return nil, err
 	}
 
-	// return ErrNotFound if no shares are found for namespaceID
+	// return ErrNamesapceNotFound if no shares are found for namespaceID
```
Wondertan, line 87:
```suggestion
	// return ErrNamespaceNotFound if no shares are found for the namespace.ID
```

share/getters/store.go
Wondertan: Ok. Makse sense!

share/getter.go
```diff
@@ -12,8 +12,13 @@ import (
...
 )
 
-// ErrNotFound is used to indicated that requested data could not be found.
-var ErrNotFound = errors.New("data not found")
+var (
+	// ErrNotFound is used to indicate that requested data could not be found.
+	ErrNotFound = errors.New("data not found")
+	// ErrNamespaceNotFound is returned by GetSharesByNamespace when data for requested root does
+	// not include any shares from the given namespace
+	ErrNamespaceNotFound = errors.New("namespace not found in data")
```
Wondertan, line 20:
```suggestion
	ErrNotFound = errors.New("share: data not found")
	// ErrNamespaceNotFound is returned by GetSharesByNamespace when data for requested root does
	// not include any shares from the given namespace
	ErrNamespaceNotFound = errors.New("share: namespace not found in data")
```

share/p2p/shrexnd/params.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/celestiaorg/celestia-node/share/p2p"
 )
 
-const protocolString = "/shrex/nd/0.0.1"
+const protocolString = "/shrex/nd/0.0.2"
```
derrandz, line 11: I had a question about this while I was writing docs for peer manager and exploring getters, why is it that versions in other protocols we have are prefixed with `v` and `shrex/nd` is not. (e.g: `/shrex/nd/v0.0.2`

Maybe since this is a breaking PR we can fix this and make this protocol uniform with the rest.

renaynay: you still didnt fix

renaynay (approved):

Wondertan:
Conflicts

share/p2p/shrexnd/params.go
distractedm1nd: it is fixed, just not showing in the diff

share/getters/store.go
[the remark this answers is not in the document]
distractedm1nd: This is actually incorrect AFAIU. ErrNodeNotFound should in fact be converted to an ErrNotFound

share/getters/shrex.go
```diff
@@ -185,6 +185,13 @@ func (sg *ShrexGetter) GetSharesByNamespace(
 		attempt int
 		err     error
 	)
+
+	// verify that the namespace could exist inside the roots before starting network requests
+	roots := filterRootsByNamespace(root, id)
+	if len(roots) == 0 {
+		return nil, share.ErrNamespaceNotFound
+	}
```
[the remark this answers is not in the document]
distractedm1nd: done but not showing in diff, will reopen PR

PR #2160, opened 2023-05-03
fix(nodebuilder): fix error log and shut the fx

Wondertan:
Granular change descriptions are in comments

### DIFF
--- a/logs/logs.go
+++ b/logs/logs.go
@@ -21,6 +21,7 @@ func SetAllLoggers(level logging.LogLevel) {
 	_ = logging.SetLogLevel("net/identify", "ERROR")
 	_ = logging.SetLogLevel("shrex/nd", "WARN")
 	_ = logging.SetLogLevel("shrex/eds", "WARN")
+	_ = logging.SetLogLevel("fx", "FATAL")
 }
 
 func SetDebugLogging() {
--- a/nodebuilder/node.go
+++ b/nodebuilder/node.go
@@ -30,7 +30,7 @@ import (
 	"github.com/celestiaorg/celestia-node/nodebuilder/state"
 )
 
-var Timeout = time.Second * 30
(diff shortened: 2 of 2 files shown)


nodebuilder/node.go
```diff
@@ -100,7 +100,7 @@ func (n *Node) Start(ctx context.Context) error {
 	if err != nil {
 		log.Debugf("error starting %s Node: %s", n.Type, err)
 		if errors.Is(err, context.DeadlineExceeded) {
-			return fmt.Errorf("node: failed to start within timeout(%s): %w", Timeout, errors.Unwrap(err))
```
Wondertan, line 103: I forgot to remove Unwrap after I was experimenting with error handling...

nodebuilder/node.go
```diff
@@ -30,7 +30,7 @@ import (
 	"github.com/celestiaorg/celestia-node/nodebuilder/state"
 )
 
-var Timeout = time.Second * 30
+var Timeout = time.Minute * 2
```
Wondertan, line 33: It's not sustainable to increase all the time, so we should consider removing timeout for BN/LNs or have diff timeouts for node types. Light node should definitely start quick and within 15 secs

logs/logs.go
```diff
@@ -21,6 +21,7 @@ func SetAllLoggers(level logging.LogLevel) {
 	_ = logging.SetLogLevel("net/identify", "ERROR")
 	_ = logging.SetLogLevel("shrex/nd", "WARN")
 	_ = logging.SetLogLevel("shrex/eds", "WARN")
+	_ = logging.SetLogLevel("fx", "FATAL")
```
Wondertan, line 24: Shut the FX by default. We only need it when we need it

renaynay (approved):

PR #2163, opened 2023-05-04
fix(share/discovery): fixes from Advertise logic audit

Wondertan:
* Use only TTL provided by us.
* Don't wait 8hs upon error
* Increase the advertise timeout

### DIFF
--- a/share/availability/discovery/discovery.go
+++ b/share/availability/discovery/discovery.go
@@ -35,11 +35,6 @@ const (
 	defaultRetryTimeout = time.Second
 )
 
-// waitF calculates time to restart announcing.
-var waitF = func(ttl time.Duration) time.Duration {
-	return 7 * ttl / 8
-}
-
 type Parameters struct {
 	// PeersLimit defines the soft limit of FNs to connect to via discovery.
 	// Set 0 to disable.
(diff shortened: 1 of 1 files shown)


share/availability/discovery/discovery.go
```diff
@@ -151,32 +147,38 @@ func (d *Discovery) Peers(ctx context.Context) ([]peer.ID, error) {
...
 	defer timer.Stop()
 	for {
-		ttl, err := d.disc.Advertise(ctx, rendezvousPoint)
+		_, err := d.disc.Advertise(ctx, rendezvousPoint)
 		if err != nil {
-			log.Debugf("Error advertising %s: %s", rendezvousPoint, err.Error())
 			if ctx.Err() != nil {
 				return
 			}
+			log.Debugf("error advertising %s: %s", rendezvousPoint, err.Error())
```
walldiss, line 162: Maybe Warn?

share/availability/discovery/discovery.go
```diff
@@ -151,32 +147,38 @@ func (d *Discovery) Peers(ctx context.Context) ([]peer.ID, error) {
...
 
+			errTimer := time.NewTimer(time.Minute)
 			select {
-			case <-timer.C:
+			case <-errTimer.C:
+				errTimer.Stop()
+				if !timer.Stop() {
+					<-timer.C
+				}
 				timer.Reset(d.params.AdvertiseInterval)
```
walldiss, line 171: Does it make sense to reset advertiseTimer only after successful advertise?

Wondertan: Yes, makes sense

share/availability/discovery/discovery.go
Wondertan: 👌🏻

renaynay (approved):

PR #2162, opened 2023-05-04
chore(nodebuilder/tests): Clean up sync_test tests

renaynay:
This PR is a part of the swamp cleanussy effort to clean up our integration tests and make them more deterministic + readable. 

This PR focuses on `sync_test.go` file.

It also contains a change to fraud proof broadcasting test but I can happily split out into separate PR.

PR #2166, opened 2023-05-04
chore(nodebuilder/tests): Clean up integration tests in `p2p_test`

renaynay:
Cleans up swamp tests in `p2p_test.go`.

Based on #2162

### DIFF
--- a/nodebuilder/tests/p2p_test.go
+++ b/nodebuilder/tests/p2p_test.go
@@ -121,8 +121,8 @@ Test-Case: Full node discovery of disconnected full nodes
 Steps:
 1. Create a Bridge Node(BN)
 2. Start a BN
-3. Create 2 full nodes with bridge node as bootstrapper peer and start them
-4. Check that nodes are connected to each other
+3. Create 2 FNs with bridge node as bootstrapper peer and start them
+4. Check that the FNs discover each other
 5. Disconnect the FNs
 6. Create one more node with discovery process disabled (however advertisement is still enabled)
 7. Check that the FN with discovery disabled is still found by the other two FNs
--- a/nodebuilder/tests/p2p_test.go
+++ b/nodebuilder/tests/p2p_test.go
@@ -5,7 +5,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/libp2p/go-libp2p/core/event"
 	"github.com/libp2p/go-libp2p/core/host"
 	"github.com/libp2p/go-libp2p/core/network"
 	"github.com/libp2p/go-libp2p/core/peer"
@@ -18,22 +17,22 @@ import (
 )
 
 /*
-Test-Case: Full/Light Nodes connection to Bridge as a Bootstapper
+Test-Case: Full/Light Nodes connection to Bridge as a Bootstrapper
 Steps:
 1. Create a Bridge Node(BN)
 2. Start a BN
-3. Create full/light nodes with bridge node as bootsrapped peer
+3. Create full/light nodes with bridge node as bootstrap peer
 4. Start full/light nodes
 5. Check that nodes are connected to bridge
 */
-func TestUseBridgeNodeAsBootstraper(t *testing.T) {
-	sw := swamp.NewSwamp(t)
-
-	bridge := sw.NewBridgeNode()
-
+func TestBridgeNodeAsBootstrapper(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), swamp.DefaultTestTimeout)
 	t.Cleanup(cancel)
 
+	sw := swamp.NewSwamp(t)
+
+	// create and start BN
+	bridge := sw.NewBridgeNode()
 	err := bridge.Start(ctx)
 	require.NoError(t, err)
 
@@ -41,39 +40,14 @@ func TestUseBridgeNodeAsBootstraper(t *testing.T) {
 
 	full := sw.NewFullNode(nodebuilder.WithBootstrappers([]peer.AddrInfo{*addr}))
 	light := sw.NewLightNode(nodebuilder.WithBootstrappers([]peer.AddrInfo{*addr}))
-	nodes := []*nodebuilder.Node{full, light}
-	for index := range nodes {
-		require.NoError(t, nodes[index].Start(ctx))
-		assert.Equal(t, *addr, nodes[index].Bootstrappers[0])
-		assert.True(t, nodes[index].Host.Network().Connectedness(addr.ID) == network.Connected)
+
+	for _, nd := range []*nodebuilder.Node{full, light} {
+		// start node and ensure that BN is correctly set as bootstrapper
+		require.NoError(t, nd.Start(ctx))
+		assert.Equal(t, *addr, nd.Bootstrappers[0])
+		// ensure that node is actually connected to BN
+		assert.True(t, nd.Host.Network().Connectedness(addr.ID) == network.Connected)
 	}
-}
-
-/*
-Test-Case: Add peer to blacklist
-Steps:
-1. Create a Full Node(BN)
-2. Start a FN
-3. Create a Light Node(LN)
-5. Start a LN
-6. Explicitly block FN id by LN
-7. Check FN is allowed to dial with LN
-8. Check LN is not allowed to dial with FN
-*/
-func TestAddPeerToBlackList(t *testing.T) {
-	sw := swamp.NewSwamp(t)
-	full := sw.NewFullNode()
-	ctx, cancel := context.WithTimeout(context.Background(), swamp.DefaultTestTimeout)
-	t.Cleanup(cancel)
-	require.NoError(t, full.Start(ctx))
-
-	addr := host.InfoFromHost(full.Host)
-	light := sw.NewLightNode()
-	require.NoError(t, light.Start(ctx))
-	require.NoError(t, light.ConnGater.BlockPeer(addr.ID))
-
-	require.True(t, full.ConnGater.InterceptPeerDial(host.InfoFromHost(light.Host).ID))
-	require.False(t, light.ConnGater.InterceptPeerDial(addr.ID))
 }
 
 /*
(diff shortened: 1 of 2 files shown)


nodebuilder/tests/sync_test.go
```diff
@@ -265,54 +256,52 @@ Steps:
...
 	err = light.Start(ctx)
 	require.NoError(t, err)
-
-	h, err = light.HeaderServ.GetByHeight(ctx, 50)
+	err = light.HeaderServ.SyncWait(ctx)
 	require.NoError(t, err)
 
-	assert.EqualValues(t, h.Commit.BlockID.Hash, sw.GetCoreBlockHashByHeight(ctx, 50))
+	// wait for the core block filling process to exit
+	require.NoError(t, <-fillDn)
```
Wondertan, line 306: we don't trust that chans will get closed

nodebuilder/tests/sync_test.go
```diff
@@ -265,54 +256,52 @@ Steps:
...
-	addrs, err = peer.AddrInfoToP2pAddrs(host.InfoFromHost(full.Host))
+	err = full.HeaderServ.SyncWait(ctx)
 	require.NoError(t, err)
 
+	// create a LN with both FN and BN as trusted peers
 	cfg = nodebuilder.DefaultConfig(node.Light)
-	cfg.Header.TrustedPeers = append(cfg.Header.TrustedPeers, addrs[0].String())
+	fullInfo, err := peer.AddrInfoToP2pAddrs(host.InfoFromHost(full.Host))
+	require.NoError(t, err)
+	cfg.Header.TrustedPeers = append(cfg.Header.TrustedPeers, bridgeInfo[0].String(), fullInfo[0].String())
```
Wondertan, line 296: I think we should hide this as part of the node setup. IIRC, I did that.

nodebuilder/tests/sync_test.go
```diff
@@ -265,54 +256,52 @@ Steps:
...
 5. Start a FN
-6. Check FN is synced to height 30
-7. Create a Light Node(LN) with a connection to BN, FN as trusted peers
+6. Check FN is synced to network head
+7. Create a Light Node(LN) with a connection to BN and FN as trusted peers
 8. Start LN
-9. Check LN is synced to height 50
+9. Check LN is synced to network head.
 */
 func TestSyncLightWithTrustedPeers(t *testing.T) {
```
Wondertan, line 264: It's now the same as TestSyncLightWithTrustedPeers from what it tests perspective or do I miss something?

nodebuilder/tests/sync_test.go
Wondertan: Relates to all places

nodebuilder/tests/sync_test.go
```diff
@@ -17,60 +17,99 @@ import (
...
-	assert.NoError(t, err)
-
-	err = light.DASer.WaitCatchUp(ctx)
+	h, err := bridge.HeaderServ.GetByHeight(ctx, numBlocks)
 	require.NoError(t, err)
-
-	assert.EqualValues(t, h.Commit.BlockID.Hash, sw.GetCoreBlockHashByHeight(ctx, 30))
+	require.EqualValues(t, h.Commit.BlockID.Hash, sw.GetCoreBlockHashByHeight(ctx, numBlocks))
+
+	t.Run("light sync against bridge", func(t *testing.T) {
```
Wondertan, line 64: Nice

nodebuilder/tests/sync_test.go
```diff
@@ -17,60 +17,99 @@ import (
...
+		// create a light node that is connected to the bridge node as
+		// a bootstrapper
+		cfg := nodebuilder.DefaultConfig(node.Light)
+		addrs, err := peer.AddrInfoToP2pAddrs(host.InfoFromHost(bridge.Host))
+		require.NoError(t, err)
+		cfg.Header.TrustedPeers = append(cfg.Header.TrustedPeers, addrs[0].String())
+		light := sw.NewNodeWithConfig(node.Light, cfg)
```
Wondertan, line 71: This really asks to be hidden away

nodebuilder/tests/sync_test.go
```diff
@@ -17,60 +17,99 @@ import (
...
-	bsize  = 16
-	btime  = time.Millisecond * 300
+	numBlocks = 20
+	bsize     = 16
+	btime     = time.Millisecond * 300
 )
 
 /*
-Test-Case: Sync a Light Node with a Bridge Node(includes DASing of non-empty blocks)
+Test-Case: Header and block/sample sync against a Bridge Node of non-empty blocks.
```
Wondertan, line 26: I would keep some test with empty blocks as well

Wondertan (commented):
Can we remove unrelated tests from the diff?

nodebuilder/tests/sync_test.go
renaynay: will wrap w select on ctx

renaynay: https://github.com/celestiaorg/celestia-node/pull/2162/commits/0c19eb09514b56d1dad13d4ac251270cf9c2f2a1

nodebuilder/tests/sync_test.go
renaynay: https://github.com/celestiaorg/celestia-node/pull/2162/commits/0c19eb09514b56d1dad13d4ac251270cf9c2f2a1

nodebuilder/tests/sync_test.go
renaynay: https://github.com/celestiaorg/celestia-node/pull/2162/commits/0c19eb09514b56d1dad13d4ac251270cf9c2f2a1

nodebuilder/tests/sync_test.go
renaynay: https://github.com/celestiaorg/celestia-node/pull/2162/commits/09cb4987429a5d49c9807ac9eda8a4f226945b17

distractedm1nd (approved):

nodebuilder/tests/p2p_test.go
```diff
@@ -61,19 +62,27 @@ Steps:
 8. Check LN is not allowed to dial with FN
 */
 func TestAddPeerToBlackList(t *testing.T) {
```
walldiss, line 64: Weird test, do you know what is the reason to have it in swamp? Looks like a unit test for connGater, which is external dependancy, so maybe it should be removed.

nodebuilder/tests/p2p_test.go
```diff
@@ -86,131 +95,123 @@ Steps:
...
-			}
+
+	for {
+		if ctx.Err() != nil {
+			t.Fatal(ctx.Err())
+		}
+		if light.Host.Network().Connectedness(host.InfoFromHost(full.Host).ID) == network.Connected {
+			// LN discovered FN successfully and is now connected
+			break
 		}
```
walldiss, line 150: Polling loop without delay would eat cpu in tests and slow down CI. Why not stick with event bus as it used to be?

nodebuilder/tests/p2p_test.go
```diff
@@ -86,131 +95,123 @@ Steps:
...
 1. Create a Bridge Node(BN)
 2. Start a BN
 3. Create 2 full nodes with bridge node as bootstrapper peer and start them
 4. Check that nodes are connected to each other
-5. Create one more node with disabled discovery
-6. Disconnect FNs from each other
-7. Check that the last FN is connected to one of the nodes
+5. Disconnect the FNs
+6. Create one more node with disabled discovery
+7. Check that the FN with discovery disabled is still found by the other two FNs
```
walldiss, line 163: I wonder, if test can skip steps 4,5: connecting FN and then disconnecting them right after. Unless theres is the reason to have those?

nodebuilder/tests/p2p_test.go
```diff
@@ -86,131 +95,123 @@ Steps:
...
 Steps:
 1. Create a Bridge Node(BN)
 2. Start a BN
 3. Create 2 full nodes with bridge node as bootstrapper peer and start them
 4. Check that nodes are connected to each other
-5. Create one more node with disabled discovery
-6. Disconnect FNs from each other
-7. Check that the last FN is connected to one of the nodes
+5. Disconnect the FNs
+6. Create one more node with disabled discovery
```
walldiss, line 162: After reading description I was a bit confused how node could be discovered, when it has discovery disabled. Turned out it description is a bit misleading. According to code last node is not really have disabled discovery, but rather it is not searching for other nodes, but still advertises itself to allow it to be discovered by others.

nodebuilder/tests/sync_test.go
```diff
@@ -17,61 +15,184 @@ import (
...
-
 	err := bridge.Start(ctx)
 	require.NoError(t, err)
 
-	h, err := bridge.HeaderServ.WaitForHeight(ctx, 20)
+	h, err := bridge.HeaderServ.WaitForHeight(ctx, numBlocks)
 	require.NoError(t, err)
+	require.EqualValues(t, h.Commit.BlockID.Hash, sw.GetCoreBlockHashByHeight(ctx, numBlocks))
+
+	t.Run("light sync against bridge", func(t *testing.T) {
```
walldiss, line 63: Suggestion: Looks like at this stage, blocks and headers are already produced by BN and there the Light and Full node only need to sync. Maybe we can use lower context, that is local for each `t.Run` to allow faster end if tests fails.

walldiss (commented):
Thank you for cleanup! Swamp looks so much better!

Wondertan:
@walldiss, @distractedm1nd, as this PR is based on https://github.com/celestiaorg/celestia-node/pull/2162
let's review it first, unless I miss something @renaynay

nodebuilder/tests/p2p_test.go
renaynay: Yeah I'm also unsure as to why that test is there. I'm happy to remove it - cc @Wondertan

Wondertan: I am also happy to remove it. IIRC, it was added by @Bidon15 to ensure blacklisting works correctly once we exposed he feature for testground tests

nodebuilder/tests/sync_test.go
renaynay: Hmm why lower the context? We still need to wait for DASing too.

nodebuilder/tests/p2p_test.go
renaynay: listening for a connected event does not work for some reason even though it definitely should. Will continue investigating.

walldiss: Probably because you need to spin out listener before even happens. With for loop you check after the event

renaynay: I started an event listening loop before the event happens and it still doesn't work. Will continue investigating but if this is blocking, we can move on other cleanussy PRs.

distractedm1nd (approved):

nodebuilder/tests/p2p_test.go
renaynay: Investigate why subscription isn't working in this test.

renaynay: https://github.com/celestiaorg/celestia-node/issues/2649

nodebuilder/tests/p2p_test.go
renaynay: added description there

nodebuilder/tests/p2p_test.go
renaynay: added description there - no reason, just additional check that they're able to discover each other

walldiss (approved):

distractedm1nd (approved):

PR #2167, opened 2023-05-05
feat(Makefile): add fetch tags action

vgonkivs:
## Overview

Added fetching tags action in order to get the correct semantic version.

## Checklist

### DIFF
--- a/.github/workflows/docker-build-publish.yml
+++ b/.github/workflows/docker-build-publish.yml
@@ -17,6 +17,6 @@ jobs:
     permissions:
       contents: write
       packages: write
-    uses: celestiaorg/.github/.github/workflows/reusable_dockerfile_pipeline.yml@v0.1.1 # yamllint disable-line rule:line-length
+    uses: celestiaorg/.github/.github/workflows/reusable_dockerfile_pipeline.yml@test_context # yamllint disable-line rule:line-length
     with:
       dockerfile: Dockerfile
--- a/Makefile
+++ b/Makefile
@@ -17,8 +17,9 @@ install-hooks:
 .PHONY: init-hooks
 
 ## build: Build celestia-node binary.
-build: fetch-tags
+build:
 	@echo "--> Building Celestia"
+	@git fetch --tags
 	@go build -o build/ ${LDFLAGS} ./cmd/celestia
 .PHONY: build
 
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -93,6 +93,12 @@ lint: lint-imports
 	@cfmt -m=100 ./...
 .PHONY: lint
 
+## fetch-tags: Fetching all tags
+fetch-tags:
+	@echo "--> Fetching all tags"
+	@git fetch --all --tags
```
Wondertan, line 99: From `man git-fetch`
`--all                 fetch from all remotes`

Why do we need tags from _all_ remotes?

Wondertan (commented):
The fetch-tags is effectively an alias for `git fetch --tags`. Is it worth adding a make command for it? We basically write `fetch-tags` instead of `git fetch --all` and this feels like an overkill and Makefile bloat

renaynay (changes requested):
I wouldn't add fetch-tags as part of the other make targets - lts just keep it separate

Also +1 to hlib comment re all remotes

renaynay (approved):

derrandz (approved):
Although it might seem as Makefile bloat, having this as a target allows referencing it in other targets easily.

Makefile
```diff
@@ -17,7 +17,7 @@ install-hooks:
 .PHONY: init-hooks
 
 ## build: Build celestia-node binary.
-build:
+build: fetch-tags
```
renaynay, line 20: can you remove fetch-tags from every other target?

renaynay (changes requested):

Makefile
vgonkivs: should the build only contain it?

Then it is no sense to have target

.github/workflows/docker-build-publish.yml
```diff
@@ -17,6 +17,6 @@ jobs:
     permissions:
       contents: write
       packages: write
-    uses: celestiaorg/.github/.github/workflows/reusable_dockerfile_pipeline.yml@v0.1.1 # yamllint disable-line rule:line-length
+    uses: celestiaorg/.github/.github/workflows/reusable_dockerfile_pipeline.yml@test_context # yamllint disable-line rule:line-length
```
renaynay, line 20: ? @vgonkivs

vgonkivs: test branch w modified `reusable_dockerfile_pipeline`

vgonkivs:
N/A

PR #2173, opened 2023-05-08
refactor(header/service): ensure that the requested height is less or equal than the network head

vgonkivs:
## Overview
Related to https://github.com/celestiaorg/celestia-node/pull/2052#discussion_r1187252215

## Checklist

### DIFF
--- a/nodebuilder/header/service.go
+++ b/nodebuilder/header/service.go
@@ -72,7 +72,8 @@ func (s *Service) GetByHeight(ctx context.Context, height uint64) (*header.Exten
 		return nil, err
 	}
 
-	if uint64(head.Height()) < height {
+	// `+1` prevents of getting an error `syncing in progress` for header that we have synced recently.
+	if uint64(head.Height()+1) < height {
 		return nil, fmt.Errorf("header: syncing in progress: "+
 			"localHeadHeight: %d, requestedHeight: %d", head.Height(), height)
 	}
--- a/nodebuilder/header/service.go
+++ b/nodebuilder/header/service.go
@@ -54,7 +54,20 @@ func (s *Service) GetVerifiedRangeByHeight(
 func (s *Service) GetByHeight(ctx context.Context, height uint64) (*header.ExtendedHeader, error) {
 	// TODO(vgonkivs): remove after https://github.com/celestiaorg/go-header/issues/32 will be
 	// implemented
-	head, err := s.syncer.Head(ctx)
+	head, err := s.store.Head(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if uint64(head.Height()) < height {
+		return nil, fmt.Errorf("header: given height exceeds local head."+
+			"localHeadHeight:%d, requestedHeight:%d", head.Height(), height)
+	}
+	if uint64(head.Height()) == height {
+		return head, nil
+	}
+
+	head, err = s.syncer.Head(ctx)
 	if err != nil {
 		return nil, err
 	}
@@ -65,10 +78,6 @@ func (s *Service) GetByHeight(ctx context.Context, height uint64) (*header.Exten
 	}
 	if uint64(head.Height()) == height {
 		return head, nil
-	}
-
-	if !s.store.HasAt(ctx, height) {
-		return nil, fmt.Errorf("header: given height exceeds local head. requestedHeight:%d", height)
 	}
 	return s.store.GetByHeight(ctx, height)
 }
--- a/nodebuilder/header/service.go
+++ b/nodebuilder/header/service.go
@@ -53,10 +53,6 @@ func (s *Service) GetVerifiedRangeByHeight(
 
 func (s *Service) GetByHeight(ctx context.Context, height uint64) (*header.ExtendedHeader, error) {
 	head, err := s.syncer.Head(ctx)
-	if err != nil {
-		return nil, err
-	}
-
 	switch {
 	case uint64(head.Height()) == height, err != nil:
 		return head, err
@@ -68,10 +64,6 @@ func (s *Service) GetByHeight(ctx context.Context, height uint64) (*header.Exten
 	// TODO(vgonkivs): remove after https://github.com/celestiaorg/go-header/issues/32 will be
 	// implemented
 	head, err = s.store.Head(ctx)
-	if err != nil {
-		return nil, err
-	}
-
 	switch {
 	case uint64(head.Height()) == height, err != nil:
 		return head, err
--- a/nodebuilder/header/service.go
+++ b/nodebuilder/header/service.go
@@ -2,6 +2,7 @@ package header
 
 import (
 	"context"
+	"fmt"
 
 	libhead "github.com/celestiaorg/go-header"
 	"github.com/celestiaorg/go-header/p2p"
@@ -51,6 +52,22 @@ func (s *Service) GetVerifiedRangeByHeight(
 }
 
 func (s *Service) GetByHeight(ctx context.Context, height uint64) (*header.ExtendedHeader, error) {
+	head, err := s.syncer.Head(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if uint64(head.Height()) < height {
+		return nil, fmt.Errorf("header/service: given height exceeds network head."+
+			"networkHeight:%d, requestedHeight:%d", head.Height(), height)
+	}
+	if uint64(head.Height()) == height {
+		return head, nil
+	}
+
+	if !s.store.HasAt(ctx, height) {
+		return nil, fmt.Errorf("header/service: given height exceeds local head. requestedHeight:%d", height)
+	}
 	return s.store.GetByHeight(ctx, height)
 }
 
--- a/nodebuilder/header/service.go
+++ b/nodebuilder/header/service.go
@@ -52,29 +52,29 @@ func (s *Service) GetVerifiedRangeByHeight(
 }
 
 func (s *Service) GetByHeight(ctx context.Context, height uint64) (*header.ExtendedHeader, error) {
-	// TODO(vgonkivs): remove after https://github.com/celestiaorg/go-header/issues/32 will be
-	// implemented
-	head, err := s.store.Head(ctx)
+	head, err := s.syncer.Head(ctx)
 	if err != nil {
 		return nil, err
 	}
 
-	if uint64(head.Height()) < height {
-		return nil, fmt.Errorf("header: given height exceeds local head."+
-			"localHeadHeight:%d, requestedHeight:%d", head.Height(), height)
+	if uint64(head.Height()+1) < height {
+		return nil, fmt.Errorf("header: given height is from the future: "+
+			"networkHeight: %d, requestedHeight: %d", head.Height(), height)
 	}
 	if uint64(head.Height()) == height {
 		return head, nil
 	}
 
-	head, err = s.syncer.Head(ctx)
+	// TODO(vgonkivs): remove after https://github.com/celestiaorg/go-header/issues/32 will be
+	// implemented
+	head, err = s.store.Head(ctx)
 	if err != nil {
 		return nil, err
 	}
 
 	if uint64(head.Height()) < height {
-		return nil, fmt.Errorf("header: given height exceeds network head."+
-			"networkHeight:%d, requestedHeight:%d", head.Height(), height)
+		return nil, fmt.Errorf("header: syncing in progress: "+
+			"localHeadHeight: %d, requestedHeight: %d", head.Height(), height)
 	}
 	if uint64(head.Height()) == height {
 		return head, nil
--- a/nodebuilder/header/service.go
+++ b/nodebuilder/header/service.go
@@ -57,12 +57,12 @@ func (s *Service) GetByHeight(ctx context.Context, height uint64) (*header.Exten
 		return nil, err
 	}
 
-	if uint64(head.Height()) < height {
+	switch {
+	case uint64(head.Height()) == height, err != nil:
+		return head, err
+	case uint64(head.Height()) < height:
 		return nil, fmt.Errorf("header: given height is from the future: "+
 			"networkHeight: %d, requestedHeight: %d", head.Height(), height)
-	}
-	if uint64(head.Height()) == height {
-		return head, nil
 	}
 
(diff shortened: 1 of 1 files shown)


nodebuilder/header/service.go
```diff
@@ -51,6 +52,22 @@ func (s *Service) GetVerifiedRangeByHeight(
...
+	if uint64(head.Height()) < height {
+		return nil, fmt.Errorf("header/service: given height exceeds network head."+
+			"networkHeight:%d, requestedHeight:%d", head.Height(), height)
+	}
+	if uint64(head.Height()) == height {
+		return head, nil
+	}
+
+	if !s.store.HasAt(ctx, height) {
+		return nil, fmt.Errorf("header/service: given height exceeds local head. requestedHeight:%d", height)
```
Wondertan, line 69: no need to mention `service` in the error, only namespaces is enough

nodebuilder/header/service.go
```diff
@@ -51,6 +52,22 @@ func (s *Service) GetVerifiedRangeByHeight(
 }
 
 func (s *Service) GetByHeight(ctx context.Context, height uint64) (*header.ExtendedHeader, error) {
+	head, err := s.syncer.Head(ctx)
+	if err != nil {
+		return nil, err
+	}
```
Wondertan, line 58: For now, we should also check whether we synced up to the requested height through LocalHead/StoreHead and if not, return an error

Wondertan: We should also have this logic with TODO to remove this after https://github.com/celestiaorg/go-header/issues/32 is fixed

vgonkivs: > LocalHead/StoreHead

I'm checking it at line 68

Wondertan: No. Syncer head is network head, while localhead can be different

Wondertan: Ah, you mean the HasAt. Lines shifted after recent changes...

nodebuilder/header/service.go
```diff
@@ -51,6 +52,24 @@ func (s *Service) GetVerifiedRangeByHeight(
...
+	if uint64(head.Height()) < height {
+		return nil, fmt.Errorf("header: given height exceeds network head."+
+			"networkHeight:%d, requestedHeight:%d", head.Height(), height)
+	}
+	if uint64(head.Height()) == height {
+		return head, nil
+	}
+
+	if !s.store.HasAt(ctx, height) {
+		return nil, fmt.Errorf("header: given height exceeds local head. requestedHeight:%d", height)
```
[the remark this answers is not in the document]
vgonkivs: The only way is to use `Head` instead

nodebuilder/header/service.go
vgonkivs: We can't get `head.Height`, without getting the Head. So it should be reworked to `Head` request. I wanted to avoid this...., but ok

nodebuilder/header/service.go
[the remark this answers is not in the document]
vgonkivs: 604757d90fad4106753fabd917c9fcae169bcfd2

nodebuilder/header/service.go
vgonkivs: 604757d90fad4106753fabd917c9fcae169bcfd2

distractedm1nd (dismissed):

nodebuilder/header/service.go
```diff
@@ -51,6 +52,33 @@ func (s *Service) GetVerifiedRangeByHeight(
...
+	// TODO(vgonkivs): remove after https://github.com/celestiaorg/go-header/issues/32 will be
+	// implemented
+	head, err := s.store.Head(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if uint64(head.Height()) < height {
+		return nil, fmt.Errorf("header: given height exceeds local head."+
+			"localHeadHeight:%d, requestedHeight:%d", head.Height(), height)
```
renaynay, line 64:
```suggestion
		return nil, fmt.Errorf("header: given height exceeds local head: "+
			"localHeadHeight: %d, requestedHeight: %d", head.Height(), height)
```

nodebuilder/header/service.go
```diff
@@ -51,6 +52,33 @@ func (s *Service) GetVerifiedRangeByHeight(
...
 
 func (s *Service) GetByHeight(ctx context.Context, height uint64) (*header.ExtendedHeader, error) {
+	// TODO(vgonkivs): remove after https://github.com/celestiaorg/go-header/issues/32 will be
+	// implemented
+	head, err := s.store.Head(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if uint64(head.Height()) < height {
```
Wondertan, line 62: head.Height should have always return uint64. Will do an issue in go-header

nodebuilder/header/service.go
```diff
@@ -51,6 +52,33 @@ func (s *Service) GetVerifiedRangeByHeight(
...
+		return head, nil
+	}
+
+	head, err = s.syncer.Head(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if uint64(head.Height()) < height {
+		return nil, fmt.Errorf("header: given height exceeds network head."+
```
Wondertan, line 76:
```suggestion
		return nil, fmt.Errorf("header: given height is from the future"+
```

nodebuilder/header/service.go
renaynay, line 62: the problem here is that sync head can (and likely will) be > store head in many cases -- therefore we should check if height == syncHead, if not, check if it exceeds localHead and short circuit.

nodebuilder/header/service.go
```diff
@@ -51,6 +52,33 @@ func (s *Service) GetVerifiedRangeByHeight(
...
+	if uint64(head.Height()) == height {
+		return head, nil
+	}
+
+	head, err = s.syncer.Head(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if uint64(head.Height()) < height {
```
Wondertan, line 75: We should allow syncerHead+1 as well to accolade for a lag

nodebuilder/header/service.go
```diff
@@ -51,6 +52,33 @@ func (s *Service) GetVerifiedRangeByHeight(
...
 func (s *Service) GetByHeight(ctx context.Context, height uint64) (*header.ExtendedHeader, error) {
+	// TODO(vgonkivs): remove after https://github.com/celestiaorg/go-header/issues/32 will be
+	// implemented
+	head, err := s.store.Head(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if uint64(head.Height()) < height {
+		return nil, fmt.Errorf("header: given height exceeds local head."+
```
Wondertan, line 63:
```suggestion
		return nil, fmt.Errorf("header: node is syncing"+
```

Wondertan: Let's make it clear for the user what the error means. As before it seems like the input was wrong

nodebuilder/header/service.go
Wondertan: Let's make it clear to the user what the error means. In the current version, it tells that the user gave wrong input, while it's not.

nodebuilder/header/service.go
Wondertan: > We can't get head.Height, without getting the Head. So it should be reworked to Head request. I wanted to avoid this...., but ok

WDYM

nodebuilder/header/service.go
Wondertan: We check here against the local head, not the sync head. I don't get why you think we should do syncHead here

nodebuilder/header/service.go
```diff
@@ -51,6 +52,33 @@ func (s *Service) GetVerifiedRangeByHeight(
 }
 
 func (s *Service) GetByHeight(ctx context.Context, height uint64) (*header.ExtendedHeader, error) {
+	// TODO(vgonkivs): remove after https://github.com/celestiaorg/go-header/issues/32 will be
+	// implemented
+	head, err := s.store.Head(ctx)
+	if err != nil {
+		return nil, err
+	}
```
Wondertan, line 60: I was wrong about checking local first. We should check against the network Head first to catch the header from future errors. If we do local first, even when the head is from the future, we will get syncing still in progress, while it's not.

nodebuilder/header/service.go
renaynay: Not for now.

nodebuilder/header/service.go
```diff
@@ -51,6 +52,33 @@ func (s *Service) GetVerifiedRangeByHeight(
 }
 
 func (s *Service) GetByHeight(ctx context.Context, height uint64) (*header.ExtendedHeader, error) {
+	head, err := s.syncer.Head(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if uint64(head.Height()+1) < height {
```
renaynay, line 60: should just be `head.Height()` here, not +1 because even if we allow for some lag, it still might be inaccessible for some time due to sync process.

nodebuilder/header/service.go
vgonkivs: meant that we could not get height using HasAt, so I reworked my approach and use Head() directly

nodebuilder/header/service.go
renaynay: @Wondertan - so we should not allow sync head to be used in this component? What differentiates this from CoreAccessor for example?

nodebuilder/header/service.go
vgonkivs: I think here is the answer https://github.com/celestiaorg/celestia-node/pull/2173#discussion_r1189612682

nodebuilder/header/service.go
Wondertan: The sync time for one header is negligible. The reason I want this because I believe this is great UX enhancer that prevents our user to see failures for events that are millisecond away.

nodebuilder/header/service.go
Wondertan: We use syncHead here

nodebuilder/header/service.go
renaynay: Ok.

nodebuilder/header/service.go
renaynay: Synchead is not milliseconds - it could be 10 seconds and then a new synchead.Head request would have to be made to catch that header. So while we can do this with localHead, we cannot do this with synchead.

renaynay:
In conclusion, i am okay with localHead + 1 but not syncHead + 1 cc @Wondertan

vgonkivs:
I don't get why we need this +1 at all. GetByHeight will block anyway(and it will not take much time)

nodebuilder/header/service.go
```diff
@@ -51,6 +52,34 @@ func (s *Service) GetVerifiedRangeByHeight(
...
+	}
+
+	// `+1` prevents of getting an error `syncing in progress` for header that we have synced recently.
+	if uint64(head.Height()+1) < height {
+		return nil, fmt.Errorf("header: syncing in progress: "+
+			"localHeadHeight: %d, requestedHeight: %d", head.Height(), height)
+	}
+	if uint64(head.Height()) == height {
+		return head, nil
+	}
```
Wondertan, line 82: This is an excellent place to do a switch, will make the code cleaner 👍🏻

Wondertan: and same for syncer check

nodebuilder/header/service.go
```diff
@@ -72,7 +72,8 @@ func (s *Service) GetByHeight(ctx context.Context, height uint64) (*header.Exten
 		return nil, err
 	}
 
-	if uint64(head.Height()) < height {
+	// `+1` prevents of getting an error in GetByHeight.
```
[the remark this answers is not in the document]
vgonkivs: done

nodebuilder/header/service.go
```diff
@@ -51,6 +52,34 @@ func (s *Service) GetVerifiedRangeByHeight(
...
+
+	// `+1` prevents of getting an error `syncing in progress` for header that we have synced recently.
+	if uint64(head.Height()+1) < height {
+		return nil, fmt.Errorf("header: syncing in progress: "+
+			"localHeadHeight: %d, requestedHeight: %d", head.Height(), height)
+	}
+	if uint64(head.Height()) == height {
+		return head, nil
+	}
 	return s.store.GetByHeight(ctx, height)
```
Wondertan, line 83:
```suggestion
	switch {
	case uint64(head.Height()) == height, err != nil:
		return head, err
	// `+1` allows for one header network lag, e.g. user request header that is milliseconds away
	case uint64(head.Height())+1 < height: 
		return nil, fmt.Errorf("header: syncing in progress: "+
			"localHeadHeight: %d, requestedHeight: %d", head.Height(), height)
	default:
	  return s.store.GetByHeight(ctx, height)
	}
```

renaynay (approved):

derrandz (dismissed):

nodebuilder/header/service.go
```diff
@@ -51,7 +52,36 @@ func (s *Service) GetVerifiedRangeByHeight(
 }
 
 func (s *Service) GetByHeight(ctx context.Context, height uint64) (*header.ExtendedHeader, error) {
-	return s.store.GetByHeight(ctx, height)
+	head, err := s.syncer.Head(ctx)
+	if err != nil {
+		return nil, err
+	}
```
Wondertan, line 58: We check in the switch so we can remove the if

nodebuilder/header/service.go
```diff
@@ -51,7 +52,36 @@ func (s *Service) GetVerifiedRangeByHeight(
...
+		return nil, fmt.Errorf("header: given height is from the future: "+
+			"networkHeight: %d, requestedHeight: %d", head.Height(), height)
+	}
+
+	// TODO(vgonkivs): remove after https://github.com/celestiaorg/go-header/issues/32 will be
+	// implemented
+	head, err = s.store.Head(ctx)
+	if err != nil {
+		return nil, err
+	}
```
Wondertan, line 73: same here

Wondertan (approved):

renaynay (approved):

PR #2175, opened 2023-05-09
refactor(metrics): use user context for metrics observation

walldiss:
## Overview

Use same approach as discussed in https://github.com/celestiaorg/celestia-node/pull/2155#discussion_r1188313387

### DIFF
--- a/das/metrics.go
+++ b/das/metrics.go
@@ -149,6 +149,9 @@ func (m *metrics) observeSample(
 	if m == nil {
 		return
 	}
+	if ctx.Err() != nil {
+		ctx = context.Background()
+	}
 	m.sampleTime.Record(ctx, sampleTime.Seconds(),
 		attribute.Bool(failedLabel, err != nil),
 		attribute.Int(headerWidthLabel, len(h.DAH.RowsRoots)),
@@ -169,6 +172,9 @@ func (m *metrics) observeGetHeader(ctx context.Context, d time.Duration) {
 	if m == nil {
 		return
 	}
+	if ctx.Err() != nil {
+		ctx = context.Background()
+	}
 	m.getHeaderTime.Record(ctx, d.Seconds())
 }
 
@@ -176,6 +182,9 @@ func (m *metrics) observeGetHeader(ctx context.Context, d time.Duration) {
 func (m *metrics) observeNewHead(ctx context.Context) {
 	if m == nil {
 		return
+	}
+	if ctx.Err() != nil {
+		ctx = context.Background()
 	}
 	m.newHead.Add(ctx, 1)
 }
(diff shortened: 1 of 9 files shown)


Wondertan (approved):

distractedm1nd (dismissed):

distractedm1nd:
conflict

Wondertan (commented):
Should we report an error in metrics that the operation was cancelled by client? It is  incorrect to submit cancelled operation as succeeded

walldiss:
@Wondertan wdym by canceled operation? If you mean that ctx could gets canceled while inside observe function, there is no way for us to catch it, since observe otel api doesn't return errors. And there is not much to worry about anyway, since metrics are not meant to be precise, it is statistical tool and it should not be a problem if we lose few.

das/metrics.go
```diff
@@ -169,6 +172,9 @@ func (m *metrics) observeGetHeader(ctx context.Context, d time.Duration) {
 	if m == nil {
 		return
 	}
+	if ctx.Err() != nil {
+		ctx = context.Background()
+	}
```
Wondertan, line 177: @walldiss, I mean for this case. The operation was canceled, and you still wanted to track it, but we should not track it as a successful one as we do right now or not track it at all due to your argument:
> And there is not much to worry about anyway, since metrics are not meant to be precise, it is statistical tool and it should not be a problem if we lose few.

walldiss: it is canceled for original operation, not for metric observation. Why would we respect original cancel for observing canceled call?

Wondertan: > Why would we respect original cancel for observing canceled call?

I am not saying that we need to __respect__ the original cancel. The code here submits metrics for both canceled and normal operations. My point is to separate canceled operations from successful ones.

share/getters/shrex.go
```diff
@@ -39,22 +38,23 @@ type metrics struct {
...
 	if m == nil {
 		return
 	}
-
-	ctx, cancel := context.WithTimeout(context.Background(), metricObservationTimeout)
-	defer cancel()
+	if ctx.Err() != nil {
+		ctx = context.Background()
+	}
 	m.ndAttempts.Record(ctx, int64(attemptCount), attribute.Bool("success", success))
```
Wondertan, line 58: @walldiss, an in code explanation
```suggestion
	if ctx.Err() != nil { // cancelled 
		ctx = context.Background() // we ignore that it is cancelled 
	}
	m.ndAttempts.Record(ctx, int64(attemptCount), attribute.Bool("success", success)) // and record success anyway, when it's not
```

walldiss: In your example `success` is a var and it is passed from caller, allowing him to control submitted value as he want.

Wondertan: Resolved then

Wondertan (approved):

PR #2178, opened 2023-05-09
fix(daser): do not spawn retry job twice

walldiss:
## Overview
In some rare cases when header height was not available for sampling and also picked up by both `recent` and `catchup` workers it could result in two parallel `retry` workers spawned for same header height. This PR fixes given case.

### DIFF
--- a/das/state.go
+++ b/das/state.go
@@ -96,6 +96,10 @@ func (s *coordinatorState) handleResult(res result) {
 		// if job was already in retry and failed again, carry over attempt count
 		lastRetry, ok := s.inRetry[h]
 		if ok {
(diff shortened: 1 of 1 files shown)


Wondertan (approved):
What would be the proper fix?

distractedm1nd (approved):

walldiss:
@Wondertan This fixes the case when jobs with type not `retry` would trigger logic that should happen only by `retry`. Other jobs should not move retry counter from InRetry to Failed, so the PR is fixing exactly this situation. Do you think comment sections is not clear or you mean some other changes that would fix this?

Wondertan:
I am asking because you mentioned this is as hotfix

PR #2180, opened 2023-05-09
fix(discovery): finxCtx is canceled too early by defer

walldiss:
## Overview
If underlying discovery.FindPeers closes the channel, `discover` call would return calling the defer stack and canceling `fincCtx` too early. This would result in all routines trying to establish connection being canceled before having time to succeed. Fixed by changing defer stack order of `wg.Wait` and `context.Cancel` calls.

The change also fixes flaky `get_peer_from_discovery` test, since flakiness was actually a bug.

### DIFF
--- a/share/availability/discovery/discovery.go
+++ b/share/availability/discovery/discovery.go
@@ -241,15 +241,15 @@ func (d *Discovery) discover(ctx context.Context) bool {
 	}
 	log.Infow("discovering peers", "want", want)
 
+	// stop discovery when we are done
+	findCtx, findCancel := context.WithCancel(ctx)
+	defer findCancel()
+
 	// we use errgroup as it provide limits
 	var wg errgroup.Group
 	// limit to minimize chances of overreaching the limit
 	wg.SetLimit(int(d.set.Limit()))
 	defer wg.Wait() //nolint:errcheck
-
-	// stop discovery when we are done
-	findCtx, findCancel := context.WithCancel(ctx)
-	defer findCancel()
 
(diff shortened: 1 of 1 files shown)


Wondertan (dismissed):
Let's comment that the order matters.

renaynay (dismissed):
I would just group the defer statements into a

```
defer func() { 
 wg.Wait()
 findCancel()
}()
```

so the ordering of exec is explicitly clear. 

alternatively make comment but i think if the ordering matters that much, it's better to make explicitly obvious

share/availability/discovery/discovery.go
```diff
@@ -245,11 +245,14 @@ func (d *Discovery) discover(ctx context.Context) bool {
...
 	wg.SetLimit(int(d.set.Limit()))
-	defer wg.Wait() //nolint:errcheck
 
 	// stop discovery when we are done
 	findCtx, findCancel := context.WithCancel(ctx)
-	defer findCancel()
+	defer func() {
+		// some workers could still be running, wait them to finish before canceling findCtx
+		wg.Wait() //nolint:errcheck
+		findCancel()
```
distractedm1nd, line 254: I thought one of the parts of the solution was cancelling connections that were taking too long - this happens in the real world mostly because of NAT we assume, and wont really show in tests

But since we are no longer "filling pending spots" with ongoing connection attempts, I assume this is no longer a concern?

walldiss: connections take different context, not the one being canceled. It is made this way to allow ongoing handshake to finish, since it could be already half done.

PR #2185, opened 2023-05-10
fix(daser):  persist retry count after restart

walldiss:
## Overview
If job was being handled by retry worker on node shutdown the retry backoff counter would be lost after restart.

### DIFF
--- a/das/coordinator_test.go
+++ b/das/coordinator_test.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"github.com/tendermint/tendermint/types"
 
 	"github.com/celestiaorg/celestia-node/header"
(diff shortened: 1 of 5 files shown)


das/worker.go
```diff
@@ -77,14 +77,13 @@ func (w *worker) run(ctx context.Context, timeout time.Duration, resultCh chan<-
 
 	for curr := w.state.from; curr <= w.state.to; curr++ {
 		err := w.sample(ctx, timeout, curr)
-		w.setResult(curr, err)
 		if errors.Is(err, context.Canceled) {
 			// sampling worker will resume upon restart
-			break
+			return
 		}
+		w.setResult(curr, err)
```
Wondertan, line 84: Why not to set err result?

Wondertan (commented):
Just one Q

das/worker.go
walldiss: if node is shut down, last header sampling will end with context.Canceled, that will put it into failed map after restart and would be picked by a retry job. I would still sample just fine after retry tho, thats why it wasn't noticed before. I changed it to clean up stored checkpoint.

Wondertan: If it is canceled, it means it didn't finish, so restart is needed or do I miss anything

renaynay: Yeah I am also confused.

walldiss: Sorry for confusion, I will try to explain by example:
`catchup` job(1-100) was canceled in the middle (lets say 50)
 
Previously: it result for 50 to be stored as failed(context.Canceled). So would be stored into checkpoint as:
```
workersToResume: [51:100]
failed: 50
 ```
 After restart it would be handled by 2 workers: `retry`(50:50) and `catchup`(51:100), so all header would be processed.
 
 Now:
 If context.Cancel happens, the worker will not be count as processed and job will be storred as :
 ```
workersToResume: [50:100]
```
After restart it will resume with one `catchup`(50:100) worker, starting from canceled height

renaynay: Okay -- super hairy tho 😭

das/state.go
```diff
@@ -111,7 +124,11 @@ func (s *coordinatorState) handleResult(res result) {
 		}
 		s.failed[h] = nextRetry
 	}
-	s.checkDone()
+
+	// cleanup retry from processed heights
```
derrandz, line 128:
```suggestion
	// cleanup processed heights from retry
```

derrandz (changes requested):
smoll improvement

das/coordinator_test.go
```diff
@@ -252,6 +253,52 @@ func TestCoordinator(t *testing.T) {
...
+		cancel()
+		wg.Wait()
+		close(waitCh)
+
+		stopCtx, cancel := context.WithTimeout(context.Background(), testParams.timeoutDelay)
+		defer cancel()
+		assert.NoError(t, coordinator.wait(stopCtx))
+
+		st := coordinator.state.unsafeStats()
+		fmt.Println(st)
```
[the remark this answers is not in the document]
walldiss: oops

das/state.go
walldiss: It is all way around. We need to clean up global retry map from heights that has been processed by job. I've added small explanation to this comment section

derrandz (approved):

distractedm1nd (approved):

Wondertan (approved):

PR #2190, opened 2023-05-12
fix(share/discovery): deadlock in limitedSet

Wondertan:
Self-explanatory.

The deadlock eventually blocks the disconnect event processing in Discovery, causing libp2p connection processing to stall. Was discovered via libp2p metrics

### DIFF
--- a/share/availability/discovery/set.go
+++ b/share/availability/discovery/set.go
@@ -47,6 +47,7 @@ func (ps *limitedSet) Size() uint {
 func (ps *limitedSet) Add(p peer.ID) (added bool) {
 	ps.lk.Lock()
 	if _, ok := ps.ps[p]; ok {
+		ps.lk.Unlock()
 		return false
 	}
 	ps.ps[p] = struct{}{}
@@ -65,10 +66,8 @@ func (ps *limitedSet) Add(p peer.ID) (added bool) {
 
 func (ps *limitedSet) Remove(id peer.ID) {
 	ps.lk.Lock()
-	defer ps.lk.Unlock()
-	if ps.limit > 0 {
-		delete(ps.ps, id)
-	}
+	delete(ps.ps, id)
+	ps.lk.Unlock()
 }
 
 // Peers returns all discovered peers from the set.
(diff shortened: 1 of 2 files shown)


share/availability/discovery/set.go
```diff
@@ -47,6 +47,7 @@ func (ps *limitedSet) Size() uint {
 func (ps *limitedSet) Add(p peer.ID) (added bool) {
 	ps.lk.Lock()
 	if _, ok := ps.ps[p]; ok {
+		ps.lk.Unlock()
```
Wondertan, line 50: This

renaynay: Nice find

renaynay (approved):
What was the reason for the `ps.limit > 0` check inside of Remove anyway?

vgonkivs (approved):

Wondertan:
Because there is no reason to grow set indefinitely when limits is zero

share/availability/discovery/set.go
walldiss: Could be one unlock:
```
ps.lk.Lock()
_, ok := ps.ps[p]
ps.lk.Unlock()
if ok{...
```

walldiss (approved):
we need to release asap

Wondertan:
The surprising fact here is that I remember fixing this, and I even found a commit in revamp discovery PR https://github.com/celestiaorg/celestia-node/pull/2117/commits/23b590e933546034bf68b8ab2fc543a2f899aa52, and @vgonkivs pointed tp the deadlock. But somehow, it was removed afterward.

Wondertan:
> Eh weird @Wondertan - if discovery short-circuits bc peerslimit is 0, then Remove wouldn't even get called.

Yeah and there is no need to have that if, if that case is impossible

Wondertan:
Found the commit which introduced the bug https://github.com/celestiaorg/celestia-node/pull/2117/commits/82eb331fec85b05d0548521b47ef153b50c45215.

share/availability/discovery/set.go
Wondertan: No, because we write to the map afterwards

distractedm1nd (approved):

PR #2192, opened 2023-05-13
fix(share/p2p/shrexeds): ugly hotfix shrex contexts for bsr

distractedm1nd:
Currently we are timing out in shrex for the full context length (2.5 minutes) because we fail to
1. Open streams
2. Read status from streams

This PR limits stream open to 5s, and reading the status from the stream to 5s (via read Deadline config param). We also put peers on timeout that return a context deadline. It is a bad fix and the correct way to do this can be discussed once we can look deeper into the problems - but for now, we need to fix the test network before it ends next week. 

This fix does not appear to be enough to fully fix syncing stall/syncing speed (the only solution that has done that so far is a new `ResultDropPeer` in peerman), but it is simple and helps a bit.

Related: https://github.com/celestiaorg/celestia-node/discussions/2191


share/p2p/shrexeds/client.go
```diff
@@ -143,11 +147,13 @@ func (c *Client) doRequest(
...
-			return
-		}
-		log.Debugw("client: setting deadline: %s", "err", err)
+		stream.SetDeadline(dl) //nolint:errcheck
+		// TODO(@distractedm1nd): This is commented out as part of a hotfix until we find good defaults for deadlines.
+		// particularly, we can get stuck reading the status from the stream for 60s if we don't comment this out
+		// if err == nil {
+		//	return
+		// }
+		// log.Debugw("client: setting deadline: %s", "err", err)
```
walldiss, line 156: I don't get this change, it looks like you commented out logging an error of deadline is not set.

distractedm1nd: The commenting out is to avoid the early return, and also avoid the log after (which should only be there on err).

This means we set the ReadDeadline to 5s (or whatever is in the config), which fixes the second context deadline (reading status from stream)

share/p2p/shrexeds/client.go
```diff
@@ -106,11 +108,15 @@ func (c *Client) doRequest(
 
 	// read and parse status from peer
 	resp := new(pb.EDSResponse)
+	err = stream.SetReadDeadline(time.Now().Add(c.params.ServerReadTimeout))
+	if err != nil {
+		return nil, fmt.Errorf("failed to set read deadline for reading status: %w", err)
+	}
```
walldiss, line 114: No need to return, we still can do request

share/p2p/shrexeds/client.go
```diff
@@ -119,6 +125,8 @@ func (c *Client) doRequest(
 
 	switch resp.Status {
```
walldiss, line 126: Make sure to update read deadline after you read status. You need bigger timeout for ods read

walldiss (changes requested):

share/p2p/shrexeds/client.go
walldiss: my bad, found the line

share/p2p/shrexeds/client.go
distractedm1nd: dont know how i didnt catch this, thanks

share/getters/shrex.go
```diff
@@ -129,6 +129,7 @@ func (sg *ShrexGetter) GetEDS(ctx context.Context, root *share.Root) (*rsmt2d.Ex
 	)
 	for {
 		if ctx.Err() != nil {
+			sg.metrics.recordEDSAttempt(ctx, attempt, false)
```
walldiss, line 132: Could you please also add same fix for shrex-nd

share/p2p/shrexeds/client.go
```diff
@@ -106,11 +108,15 @@ func (c *Client) doRequest(
...
+	if err != nil {
+		log.Debugw("client: failed to set read deadline for reading status", "err", err)
+	}
 	_, err = serde.Read(stream, resp)
 	if err != nil {
 		// server is overloaded and closed the stream
 		if errors.Is(err, io.EOF) {
-			c.metrics.ObserveRequests(ctx, 1, p2p.StatusRateLimited)
+			c.metrics.ObserveRequests(ctx, 1, p2p.StatusNotFound)
 			return nil, p2p.ErrNotFound
```
walldiss, line 120: Is it the returned value that was set incorrectly?

distractedm1nd: Yes

renaynay: yeah, I'm confused here - It's not actually "NotFound" no? 

also make sure to be consistent with nd err.

walldiss: I guess the comment section is off:
```
// server is overloaded and closed the stream
```

Wondertan (dismissed):
It's a bit confusing about why we choose specific timeouts for certain ops, but we can clean this up later

share/p2p/shrexeds/client.go
distractedm1nd: will change the comment

Wondertan (approved):

walldiss (approved):

PR #2207, opened 2023-05-15
feat(share/discovery): Discard method

### LINKED ISSUE #2204: share/discovery: DiscardPeer method
We need a way for the peer manager to report back to discovery if some peer is not useful anymore and to discover someone else instead.

Wondertan:
Fixes #2204 

Regarding the case where we exhausted all the peers on the network and backed them all off. It can happen on the smaller network, and in the worst case, the node will wait 10 minutes(the default backoff time), which is acceptable considering peers from shrexsub #2105

### DIFF
--- a/share/availability/discovery/discovery.go
+++ b/share/availability/discovery/discovery.go
@@ -122,6 +122,28 @@ func (d *Discovery) Peers(ctx context.Context) ([]peer.ID, error) {
 	return d.set.Peers(ctx)
 }
 
+// Discard removes the peer from the peer set and rediscovers more if soft peer limit is not
+// reached.
+func (d *Discovery) Discard(id peer.ID) {
+	if !d.set.Contains(id) {
+		return
+	}
+
+	d.host.ConnManager().Unprotect(id, rendezvousPoint)
+	d.connector.Backoff(id)
+	d.set.Remove(id)
+	d.onUpdatedPeers(id, false)
+	log.Debugw("removed peer from the peer set", "peer", id)
+
+	if d.set.Size() < d.set.Limit() {
+		// trigger discovery
+		select {
+		case d.triggerDisc <- struct{}{}:
+		default:
+		}
+	}
+}
+
 // Advertise is a utility function that persistently advertises a service through an Advertiser.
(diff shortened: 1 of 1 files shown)


walldiss:
It seems like with current solution peer could be rediscovered right away. Could we move discarded peer to some set that will prevent it from be in discovery set?

walldiss:
Nvm, answer was in description. I'm a bit sleepy and skipped reading it

Wondertan:
...
```
=== RUN   TestIntegration/get_peer_from_discovery
    manager_test.go:446: 
        	Error Trace:	/home/runner/work/celestia-node/celestia-node/share/p2p/peers/manager_test.go:446
        	            	context deadline exceeded
        	Test:       	TestIntegration/get_peer_from_discovery
```

walldiss (approved):

distractedm1nd:
For the integration it may be easier if the Discard API returns a boolean, if it was in discovery or not. This is because:

For removing a peer from the pool, we rely on discovery to call our callback, which calls pool.remove
If we call disc.Discard(peer), it will call pool.remove through the callback
But not all peers are from discovery, so if Discard didnt get rid of the peer we need to call pool.remove ourselves

Not ideal but its a solution

share/availability/discovery/discovery.go
```diff
@@ -122,6 +122,30 @@ func (d *Discovery) Peers(ctx context.Context) ([]peer.ID, error) {
...
+	d.onUpdatedPeers(id, false)
+	log.Debugw("removed peer from the peer set", "peer", id)
+
+	if d.set.Size() < d.set.Limit() {
+		// trigger discovery
+		select {
+		case d.triggerDisc <- struct{}{}:
+		default:
+		}
+	}
```
renaynay, line 144: could make a method to DRY this but nit

renaynay (approved):

share/availability/discovery/discovery.go
Wondertan: I would, but it is only used in one place.

distractedm1nd (approved):

PR #2210, opened 2023-05-15
feat(share/discovery): persist discovered peers

### LINKED ISSUE #2208: share/discovery: persist discovered peers
Discovered peers, if not discarded are proven to be good, there is no need to rediscover peers on restart.

Wondertan:
Closes #2208 

- [ ] Integrate option
- [ ] Load peers into peer manager

### DIFF
--- a/share/availability/discovery/discovery_test.go
+++ b/share/availability/discovery/discovery_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/ipfs/go-datastore"
 	dht "github.com/libp2p/go-libp2p-kad-dht"
 	"github.com/libp2p/go-libp2p/core/discovery"
 	"github.com/libp2p/go-libp2p/core/host"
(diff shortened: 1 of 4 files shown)


distractedm1nd:
I'm not convinced this is necessary outside of the context of saving peers for bootstrapper usage

Wondertan:
> I'm not convinced this is necessary outside of the context of saving peers for bootstrapper usage, as I have never seen issues finding peers very quickly once we join the network. In fact, until other errors are fixed, this will probably give the disadvantage that we save bad peers (which is why restarting helps so often)

Think of this in conjunction with Discard, which filters unhealthy peers

share/availability/discovery/discovery_test.go
```diff
@@ -109,6 +162,18 @@ func (t *testnet) discovery(opts ...Option) *Discovery {
 	return disc
 }
 
+func (t *testnet) borkenDiscovery(opts ...Option) *Discovery {
```
vgonkivs, line 165: typo `brokenDiscovery`

Wondertan: It was intentional 😈

PR #2215, opened 2023-05-16
fix(header): check that Commit corresponds to Header

Wondertan:
During the BEFP discussion on TG with @vgonkivs, we figured out that with have ___CRITICAL___ vulnerability in the header validation process. The funniest part is that we abuse this vulnerability to make our BEFP test in Swamp work, and the test represents the exact scenario of such vulnerability: a malicious BN/FN can substitute any value in RawHeader it wants like DAH and the verification on the client side still passes. This is because we don't check that BlockID(Hash) in Commit is aligned with recomputed RawHeader hash.

Unfortunately, this breaks our BEFP testing suite, and the only way to reliably produce FPs now is by modifying the Core/App node.

image

### DIFF
--- a/header/header.go
+++ b/header/header.go
@@ -118,7 +118,6 @@ func (eh *ExtendedHeader) Validate() error {
 	err = eh.Commit.ValidateBasic()
 	if err != nil {
 		return fmt.Errorf("ValidateBasic error on Commit at height %d: %w", eh.Height(), err)
-
 	}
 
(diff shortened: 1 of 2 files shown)


vgonkivs (dismissed):

vgonkivs:
Next step: rework all swamp tests to accept the already broken block

vgonkivs (approved):

renaynay (approved):
Holy hell.

PR #2218, opened 2023-05-16
bugfix(blob/service): fix issues found during reviewing  the blob service

vgonkivs:
## Overview
Fixed all findings in https://github.com/celestiaorg/celestia-node/pull/2052:
* rework Proof creation(removed unneeded proofs from previous rows);
* Added a condition that verifies passed proof inside `Included`
* extended tests

## Checklist

### DIFF
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -42,6 +42,15 @@ func (p Proof) equal(input Proof) error {
 				return ErrInvalidProof
 			}
 		}
+
+		if proof.Start() != input[i].Start() || proof.End() != input[i].End() {
+			return ErrInvalidProof
+		}
+
+		if !bytes.Equal(proof.LeafHash(), input[i].LeafHash()) {
+			return ErrInvalidProof
+		}
+
 	}
 	return nil
 }
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -1,6 +1,8 @@
 package blob
 
 import (
+	"bytes"
+
 	"github.com/celestiaorg/celestia-app/x/blob/types"
 	"github.com/celestiaorg/nmt"
 	"github.com/celestiaorg/nmt/namespace"
@@ -13,6 +15,11 @@ type Commitment []byte
 
 func (com Commitment) String() string {
 	return string(com)
+}
+
+// Verify ensures that commitments are the same
+func (com Commitment) Verify(c Commitment) bool {
+	return bytes.Equal(com, c)
 }
 
 // Proof is a collection of nmt.Proofs that verifies the inclusion of the data.
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -17,8 +17,8 @@ func (com Commitment) String() string {
 	return string(com)
 }
 
-// Verify ensures that commitments are the same
-func (com Commitment) Verify(c Commitment) bool {
+// Equal ensures that commitments are the same
+func (com Commitment) Equal(c Commitment) bool {
 	return bytes.Equal(com, c)
 }
 
@@ -26,6 +26,25 @@ func (com Commitment) Verify(c Commitment) bool {
 type Proof []*nmt.Proof
 
 func (p Proof) Len() int { return len(p) }
+
+// equal is a temporary method that compares two proofs.
+// should be removed in BlobService V1.
+func (p Proof) equal(input Proof) error {
+	if p.Len() != input.Len() {
+		return ErrInvalidProof
+	}
+
+	for i, proof := range p {
+		pNodes := proof.Nodes()
+		inputNodes := input[i].Nodes()
+		for i, node := range pNodes {
+			if !bytes.Equal(node, inputNodes[i]) {
+				return ErrInvalidProof
+			}
+		}
+	}
+	return nil
+}
 
 // Blob represents any application-specific binary data that anyone can submit to Celestia.
 type Blob struct {
--- a/blob/blob_test.go
+++ b/blob/blob_test.go
@@ -1,14 +1,11 @@
 package blob
 
 import (
-	"bytes"
-	"sort"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	tmrand "github.com/tendermint/tendermint/libs/rand"
-	"github.com/tendermint/tendermint/types"
 
 	"github.com/celestiaorg/celestia-app/pkg/appconsts"
 	"github.com/celestiaorg/celestia-app/pkg/shares"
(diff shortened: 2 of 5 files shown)


blob/service.go
```diff
@@ -258,6 +276,13 @@ func (s *Service) getByCommitment(
 
 		// drop info of the checked blob
 		rawShares = rawShares[amount:]
+		if len(rawShares) > 0 {
+			// save proof for the last row in case if we have rawShares
```
distractedm1nd, line 280:
```suggestion
			// save proof for the last row in case we have rawShares
```

blob/service.go
```diff
@@ -198,8 +212,12 @@ func (s *Service) getByCommitment(
 
 	namespacedShares, err := s.shareGetter.GetSharesByNamespace(ctx, header.DAH, nID)
 	if err != nil {
-		return nil, nil, err
+		if strings.Contains(err.Error(), share.ErrNotFound.Error()) {
```
distractedm1nd, line 215:
```suggestion
		if errors.Is(err, share.ErrNotFound) {
```

distractedm1nd: Commenting this as a reminder to update to include share.ErrNamespaceNotFound once merged into main

In sync we said both can be converted to a BlobNotFound, but after a little thinking the error handling should probably be to return empty data for NamespaceNotFound with a nil error

distractedm1nd (approved):

blob/service.go
renaynay, line 215: use errors.Is?

blob/service.go
```diff
@@ -198,8 +212,12 @@ func (s *Service) getByCommitment(
...
+			err = ErrBlobNotFound
+		}
+		return nil, nil, ErrBlobNotFound
```
renaynay, line 218: also if you're returning `ErrBlobNotFound` anyway, no need for this additional check

vgonkivs: Oooops, it should be an err

blob/blob_test.go
```diff
@@ -49,7 +46,9 @@ func TestBlob(t *testing.T) {
 		{
 			name: "shares to blobs",
 			expectedRes: func(t *testing.T) {
-				b, err := sharesToBlobs(splitBlob(t, blob...))
+				sh, err := BlobsToShares(blob...)
+				require.NoError(t, err)
+				b, err := SharesToBlobs(sh)
```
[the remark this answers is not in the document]
vgonkivs: `assert.Equal(t, blob[0].Commitment(), b[0].Commitment())`

This is enough I guess, because commitments are built from the data

blob/service.go
vgonkivs: yes

walldiss (approved):

blob/service_test.go
```diff
@@ -26,10 +26,15 @@ import (
...
+		blobSize2 = 20
+		blobSize3 = 12
+	)
+	blobs0 := generateBlob(t, []int{blobSize0, blobSize1}, false)
+	blobs1 := generateBlob(t, []int{blobSize2, blobSize3}, true)
 	service := createService(ctx, t, append(blobs0, blobs1...))
-
 	var test = []struct {
 		name           string
 		doFn           func() (interface{}, error)
```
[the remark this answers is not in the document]
vgonkivs: lol. I did not have any reason 😄 . Designed it at some point like this and continue working with this approach

renaynay (commented):
just this one tiny comment

blob/blob.go
```diff
@@ -15,6 +17,11 @@ func (com Commitment) String() string {
 	return string(com)
 }
 
+// Verify ensures that commitments are the same
+func (com Commitment) Verify(c Commitment) bool {
```
Wondertan, line 21: It should be named just Equals.

blob/service.go
```diff
@@ -165,11 +169,19 @@ func (s *Service) Included(
...
+	switch err {
+	case nil:
+	case ErrBlobNotFound:
+		return false, nil
+	default:
 		return false, err
 	}
-	return true, nil
+
+	if !reflect.DeepEqual(proof, resProof) {
```
Wondertan, line 181: There is a cleaner way to do it. Proof has getters for every private field, and we can easily make Equals method on the proof that accesses all of them and compares.

Wondertan (changes requested):

blob/service.go
vgonkivs: But. it will be the same. This approach gives us less code

blob/blob.go
vgonkivs: `Equals` sounds good

blob/service.go
Wondertan: It's effectively the same - yes, but reflect is a dirty hack for when we don't have access to the internal fields. Ideally, we would contribute to NMT with the Equals method. 

Also, note that this is going to be a hot path for a while(until v1), and having reflect on a hot path has its performance implications

blob/service.go
```diff
@@ -245,21 +248,38 @@ func (s *Service) getByCommitment(
...
+			return blob, &proofs, nil
 		}
 
 		// drop info of the checked blob
 		rawShares = rawShares[amount:]
+		if len(rawShares) > 0 {
+			// save proof for the last row in case we have rawShares
+			proofs = proofs[len(proofs)-1:]
+		} else {
+			// otherwise slash proofs
```
Wondertan, line 265: Slash always gives me different vibes in blockchain context :D

vgonkivs: okie :D

blob/helper.go
```diff
@@ -35,6 +46,29 @@ func sharesToBlobs(rawShares [][]byte) ([]*Blob, error) {
...
+			ShareVersion: uint8(blob.Version()),
+		})
+	}
+
+	sort.Slice(b, func(i, j int) bool {
+		val := bytes.Compare(b[i].NamespaceID, b[j].NamespaceID)
+		return val <= 0
+	})
+
+	rawShares, err := shares.SplitBlobs(0, nil, b, false)
```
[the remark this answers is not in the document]
vgonkivs: our blobs should go in increasing order of nID otherwise we will get an error if will try to build an nmt tree from `rawShares`. Sorting protects from this error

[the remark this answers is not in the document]
Wondertan: OK

blob/helper.go
```diff
@@ -1,15 +1,26 @@
...
-func sharesToBlobs(rawShares [][]byte) ([]*Blob, error) {
+// SharesToBlobs takes raw shares and converts them to the blobs.
+func SharesToBlobs(rawShares []share.Share) ([]*Blob, error) {
 	if len(rawShares) == 0 {
 		return nil, ErrBlobNotFound
 	}
+	rawShares, err := removePadding(rawShares)
+	if err != nil {
+		return nil, err
+	}
```
Wondertan, line 23: Is there a reason why parse shares does not removed padding automatically? It feels kinda ugly that users have to do it themselves + it's unnecessary allocations. 

I know it's more of a question to the app team, but maybe you know.

Wondertan (approved):
Approve, but we should still update the name of the PR

bugfix: improve - doesn't compile semantically

PR #2219, opened 2023-05-16
fix(header): fix testing utilities and tests

Wondertan:
#2215 was approved and merged too quickly, making some tests fail. This PR fixes that. Expect further a new PR to clean up Validate func and increase coverage for it.

### DIFF
--- a/header/header.go
+++ b/header/header.go
@@ -132,6 +132,12 @@ func (eh *ExtendedHeader) Validate() error {
 		)
 	}
 
+	// ensure data root from raw header matches computed root
+	if !bytes.Equal(eh.DAH.Hash(), eh.DataHash) {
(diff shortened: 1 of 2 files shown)


renaynay (approved):

header/header.go
```diff
@@ -132,6 +132,12 @@ func (eh *ExtendedHeader) Validate() error {
 		)
 	}
 
+	// ensure data root from raw header matches computed root
```
renaynay, line 135: why the change in order

renaynay (approved):

header/header.go
Wondertan: I will reorder those things in the next PR anyway. The order of verification matters here as well and ours is slightly off

vgonkivs (approved):

PR #2223, opened 2023-05-17
feat(core): remote client logging

Wondertan:
This way we can enable logging of remote core client on demand

Untested


Wondertan:
Closing until it becomes needed. Right now there is not much value to add it.

PR #2226, opened 2023-05-18
improvement(test/swamp): set core IP and Port to nodes

vgonkivs:
## Overview
configure test nodes to be able to access the core node and set keyring from an already funded account in order to submit transactions.

## Checklist


nodebuilder/tests/swamp/swamp.go
```diff
@@ -224,11 +237,9 @@ func (s *Swamp) NewNodeWithStore(
 ) *nodebuilder.Node {
 	var n *nodebuilder.Node
 
-	ks, err := store.Keystore()
-	require.NoError(s.t, err)
-
+	signer := apptypes.NewKeyringSigner(s.ClientContext.Keyring, s.Accounts[0], s.ClientContext.ChainID)
```
Wondertan, line 240: Would be nice to rotate accounts for diff nodes, but also good

Wondertan (approved):
Nice

nodebuilder/tests/swamp/swamp.go
vgonkivs: I think it's not necessary to have such functionality because for test purposes we should only be able to submit txs and it does not matter how many accounts will be used

walldiss (approved):

renaynay (approved):

PR #2228, opened 2023-05-18
fix(utils): splitCtx return original timeout when it is less than min timeout

walldiss:
## Overview

Handle case when timeout left in original context is less than min timeout. Also refactored one more if because it is nicer this way.


Wondertan (approved):
Makes sense

distractedm1nd (approved):

PR #2229, opened 2023-05-19
refactor(nodebuilder/share): don't wrap store getter with tee getter 

walldiss:
## Overview

The change will save extra dagstore lookup operation on every call to store getter resulting with success. For free.


distractedm1nd (approved):

Wondertan (approved):
Nice find

PR #2230, opened 2023-05-19
fix(share/p2p)!: share.ErrNamespaceNotFound and integration into shrex-nd

### LINKED ISSUE #2145: bug(shares/GetSharesByNamespace): missing response type `ErrNamespaceNotFound`

distractedm1nd:
This PR replaces #2156 , as gh was not showing the diff properly. It closes #2145


share/p2p/shrexnd/server.go
```diff
@@ -130,12 +130,15 @@ func (srv *Server) handleNamespacedData(ctx context.Context, stream network.Stre
 	}
 
 	shares, err := srv.getter.GetSharesByNamespace(ctx, dah, req.NamespaceId)
-	if errors.Is(err, share.ErrNotFound) {
+	switch {
+	case errors.Is(err, share.ErrNotFound):
 		logger.Warn("server: nd not found")
```
renaynay, line 135: this log should be more relevant to the not found err

share/p2p/shrexnd/server.go
```diff
@@ -130,12 +130,15 @@ func (srv *Server) handleNamespacedData(ctx context.Context, stream network.Stre
...
 		srv.respondNotFoundError(ctx, logger, stream)
 		return
-	}
-	if err != nil {
+	case errors.Is(err, share.ErrNamespaceNotFound):
```
renaynay, line 138: debug log here?

renaynay (approved):

share/p2p/shrexnd/server.go
distractedm1nd: No need for a debug log. Its a successful case

walldiss (approved):

PR #2239, opened 2023-05-23
share/discovery: moving the package closer to shrex

### LINKED ISSUE #2206: share/discovery: move the package closer to shrex
A minor issue is to move the discovery to `share/p2p/`. Currently, it's in `availability/discovery` but it does not make sense to keep there anymore

Chirag018:
Closes #2206


Wondertan (approved):
Thank you, @Chirag018!

PR #2243, opened 2023-05-23
fix(share/shrex): return aggregated error if context Deadline reached

walldiss:
## Overview

Errors were lost on context deadline

### DIFF
--- a/share/getters/shrex.go
+++ b/share/getters/shrex.go
@@ -130,6 +130,10 @@ func (sg *ShrexGetter) GetEDS(ctx context.Context, root *share.Root) (*rsmt2d.Ex
 	for {
 		if ctx.Err() != nil {
 			sg.metrics.recordEDSAttempt(ctx, attempt, false)
+			if err != nil {
+				// return aggregated error
+				return nil, err
+			}
 			return nil, ctx.Err()
 		}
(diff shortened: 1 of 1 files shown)


share/getters/shrex.go
```diff
@@ -130,6 +130,10 @@ func (sg *ShrexGetter) GetEDS(ctx context.Context, root *share.Root) (*rsmt2d.Ex
 	for {
 		if ctx.Err() != nil {
 			sg.metrics.recordEDSAttempt(ctx, attempt, false)
+			if err != nil {
+				// return aggregated error
+				return nil, err
+			}
 			return nil, ctx.Err()
```
renaynay, line 137: do you want to just append the context err as well instead of the check?

Wondertan (commented):
Hmm, it's correct to return a context error when the caller canceled it, unless the context cancelation error comes from the internal or sub-context. Is this the latter case?

share/getters/shrex.go
walldiss: Nice idea! Thank you!

Wondertan (approved):
Ok, the Join kinda fixes my concern

renaynay (approved):

PR #2245, opened 2023-05-24
perf(share/p2p/peers): Re-enable blacklisting in peer manager

renaynay:
This PR should be kept in draft mode until confirmed on Robusta that blacklisting is working as intended (no peers are getting blacklisted in an honest environment).

We should also create swamp tests with fraudulent actors to test that blacklisting is actually working as well


renaynay:
Closing this PR as https://github.com/celestiaorg/celestia-node/issues/3011 should handle the consideration of blacklisting in peer manager.

PR #2246, opened 2023-05-24
fix(modheader): check err first in switch

Wondertan:
Otherwise, we call Height on the nil header

### DIFF
--- a/nodebuilder/header/service.go
+++ b/nodebuilder/header/service.go
@@ -54,7 +54,7 @@ func (s *Service) GetVerifiedRangeByHeight(
 func (s *Service) GetByHeight(ctx context.Context, height uint64) (*header.ExtendedHeader, error) {
 	head, err := s.syncer.Head(ctx)
 	switch {
-	case uint64(head.Height()) == height, err != nil:
+	case err != nil, uint64(head.Height()) == height:
 		return head, err
 	case uint64(head.Height()) < height:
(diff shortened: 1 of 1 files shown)


distractedm1nd (dismissed):

renaynay (approved):
<3

walldiss:
Maybe split error case and non-error case? I think it will be easier to read

Wondertan:
Optimizing for LOCs 😁 
But I agree that's more readable and we should do that

vgonkivs (approved):

nodebuilder/header/service.go
```diff
@@ -17,12 +17,20 @@ import (
...
 
-	syncer    *sync.Syncer[*header.ExtendedHeader]
+	syncer    syncer
 	sub       libhead.Subscriber[*header.ExtendedHeader]
 	p2pServer *p2p.ExchangeServer[*header.ExtendedHeader]
 	store     libhead.Store[*header.ExtendedHeader]
 }
 
+// syncer bare minimum Syncer interface for testing
+type syncer interface {
```
renaynay, line 27: can we define this elsewhere then?

renaynay: Actually, this is fine.

renaynay (approved):

nodebuilder/header/service.go
Wondertan: It's meant to be a localized private interface with no external usage.

PR #2255, opened 2023-05-25
refactor(misc): pretty print namespace.iD and peer.ID

walldiss:
## Overview

Log `namespace.ID` and `peer.ID` in pretty format. Also opened an issue for nmt pkg to add `StringHex` method and if it is **OK**, will update PR to use it. 

https://github.com/celestiaorg/nmt/issues/200


distractedm1nd (approved):
Thanks for this

share/getters/shrex.go
```diff
@@ -204,6 +205,7 @@ func (sg *ShrexGetter) GetSharesByNamespace(
 		if getErr != nil {
 			log.Debugw("nd: couldn't find peer",
 				"hash", root.String(),
+				"nid", hex.EncodeToString(id),
```
renaynay, line 208: we should eventually have method on namespace.ID to pretty print the hex right?

share/p2p/peers/manager.go
```diff
@@ -390,15 +390,14 @@ func (m *Manager) getOrCreatePool(datahash string) *syncPool {
...
 	m.metrics.observeBlacklistPeers(reason, len(peerIDs))
 
-	if !m.params.EnableBlackListing {
-		return
-	}
 	for _, peerID := range peerIDs {
+		log.Debugw("blacklisting peer", "peer", peerID.String(), "reason", reason)
+		if !m.params.EnableBlackListing {
+			continue
+		}
```
renaynay, line 399: can you put debug log after this then?

renaynay (approved):

share/getters/shrex.go
walldiss: It is already merged and will be there in next nmt release.

share/p2p/peers/manager.go
walldiss: I made it before, to make it possible to debug blacklisting while it is disabled.

renaynay: can you make a note of that then 😬  hopefully we'll remove this EnableBlacklisting altogether soon

walldiss: I've added log line, to clarify why it is always logged

renaynay (approved):

Wondertan (approved):

PR #2256, opened 2023-05-25
feat(share/getter) Add support for Non-inclusion proofs 

walldiss:
## Overview

This PR implements support for non-inclusion proof for getters and shrex protocol.
Requires https://github.com/celestiaorg/celestia-node/pull/2242 to be merged first


share/getters/shrex_test.go
```diff
@@ -101,9 +101,11 @@ func TestShrexGetter(t *testing.T) {
 		})
 
 		// corrupt NID
-		nID[4]++
+		newNID := make([]byte, 8)
+		copy(newNID, nID)
+		newNID[4]++
```
walldiss, line 106: This one took me around 4 hours.

distractedm1nd: omg, hate those days

share/p2p/shrexnd/client.go
```diff
@@ -55,7 +55,7 @@ func (c *Client) RequestND(
 	peer peer.ID,
 ) (share.NamespacedShares, error) {
 	shares, err := c.doRequest(ctx, root, nID, peer)
-	if err == nil {
+	if err == nil || errors.Is(err, share.ErrNamespaceNotFound) {
```
distractedm1nd, line 58: When will it hit this instead of line 122?

share/getters/shrex_test.go
Wondertan: Why was it failing? It's not apparent why copy is needed

share/getters/utils.go
```diff
@@ -84,9 +84,10 @@ func collectSharesByNamespace(
 		return nil, err
 	}
 
-	// return ErrNamespaceNotFound if no shares are found for the namespace.ID
+	// return ErrNamespaceNotFound along with collected proofs if no shares are found for the
+	// namespace.ID
```
Wondertan, line 88: We should add this special case to the documentation of the interface.

share/p2p/shrexnd/pb/share.proto
```diff
@@ -28,5 +28,6 @@ message Row {
 message Proof {
   int64 start = 1;
   int64 end = 2;
-  repeated bytes Nodes = 3;
+  repeated bytes nodes = 3;
+  bytes hashleaf = 4;
```
Wondertan, line 32: Optional?

share/p2p/shrexnd/client.go
walldiss: If 122 is not hit, it will hit no error case. Both needs to return data + error

share/getters/utils.go
walldiss: The error is defined right next to interface and has same description.

share/p2p/shrexnd/pb/share.proto
walldiss: I would say it is optional to make it optional 😆 . There is no check for hashleaf being nil or zero-value on client side, since proof construction is based on status code model

share/getters/utils.go
Wondertan: // ErrNamespaceNotFound is returned by GetSharesByNamespace when data for requested root does
// not include any shares from the given namespace

From main

I don't see the comment being changed in this PR

Wondertan: Also, mentioning the special case just on the error is not enough

walldiss: I've updated the description of ErrNamespaceNotFound to indicate how it is supposed to be handled along with non-inclusion proofs by user.

share/getter.go
```diff
@@ -35,6 +39,8 @@ type Getter interface {
 
 	// GetSharesByNamespace gets all shares from an EDS within the given namespace.
 	// Shares are returned in a row-by-row order if the namespace spans multiple rows.
+	// If data for the requested root does not include any shares from the given namespace
+	// ErrNamespaceNotFound will be returned along with non-inclusion proofs if there are any.
```
Wondertan, line 43: We should still assess the choice of returning value and the error on the error case:
* It's not idiomatic. Even io.EOF is returned after we read out everything.
* The go-jsonrpc may enforce idiomatic behavior and return either rather than both.

Wondertan (commented):
I have only one general API concern. The impl LGTM turned out to be much easier than I originally thought.

Wondertan (changes requested):
Ok, so we decided in sync to remove the error and make the data not found the case a success case, so requesting changes until this is done

share/getters/shrex_test.go
```diff
@@ -88,23 +89,56 @@ func TestShrexGetter(t *testing.T) {
...
-		_, err := getter.GetSharesByNamespace(ctx, &dah, nID)
-		require.ErrorIs(t, err, share.ErrNamespaceNotFound)
+		emptyShares, err := getter.GetSharesByNamespace(ctx, &dah, nID)
+		require.NoError(t, err)
+		// no shares should be returned
+		require.Empty(t, emptyShares.Flatten())
+		require.Nil(t, emptyShares.Verify(&dah, nID))
+	})
+
+	t.Run("ND_namespace_not_in_root", func(t *testing.T) {
```
Wondertan, line 118: Means not in at all in the DAH?

blob/service.go
```diff
@@ -193,8 +193,7 @@ func (s *Service) getByCommitment(
 
 	namespacedShares, err := s.shareGetter.GetSharesByNamespace(ctx, header.DAH, nID)
 	if err != nil {
-		if errors.Is(err, share.ErrNamespaceNotFound) ||
-			errors.Is(err, share.ErrNotFound) {
+		if errors.Is(err, share.ErrNotFound) {
 			err = ErrBlobNotFound
```
Wondertan, line 197: Should we also short-circuit here for when len of namespacedShares == 0?

walldiss: It depends on BlobModule `ErrBlobNotFound` definition. I assumed that it is indicating that node was not able to find data, which means user should retry. I think it would be nice to document `ErrBlobNotFound` so it is clear what it actually means. 

More detailed answer to your question. When namespacedShares == 0 it could be 2 cases:
1. `share.ErrNotFound` (node wasn't able to find data, retry needed)
2. namespace not included (no retry needed)

share/get_test.go
```diff
@@ -237,7 +237,7 @@ func TestCollectLeavesByNamespace_AbsentNamespaceId(t *testing.T) {
 	t.Cleanup(cancel)
 	bServ := mdutils.Bserv()
 
-	shares := RandShares(t, 16)
+	shares := RandShares(t, 1024)
```
renaynay, line 240: can we just extract into var

share/getters/cascade.go
```diff
@@ -122,7 +122,7 @@ func cascadeGetters[V any](
 		getCtx, cancel := ctxWithSplitTimeout(ctx, len(getters)-i, 0)
 		val, getErr := get(getCtx, getter)
 		cancel()
-		if getErr == nil || errors.Is(getErr, share.ErrNamespaceNotFound) {
+		if getErr == nil {
 			return val, getErr
```
renaynay, line 126: let's return nil explicitly then

share/getters/shrex_test.go
```diff
@@ -88,23 +89,56 @@ func TestShrexGetter(t *testing.T) {
...
 		peerManager.Validate(ctx, srvHost.ID(), shrexsub.Notification{
 			DataHash: dah.Hash(),
 			Height:   1,
 		})
 
 		// corrupt NID
-		nID[4]++
+		nID := make([]byte, ipld.NamespaceSize)
+		copy(nID, maxNID)
+		nID[ipld.NamespaceSize-1]-- // pray for last byte to not be 0x00
```
renaynay, line 107: LOL lets just do a check?

share/getters/shrex_test.go
```diff
@@ -166,8 +200,8 @@ func newStore(t *testing.T) (*eds.Store, error) {
 func generateTestEDS(t *testing.T) (*rsmt2d.ExtendedDataSquare, da.DataAvailabilityHeader, nmtnamespace.ID) {
 	eds := share.RandEDS(t, 4)
 	dah := da.NewDataAvailabilityHeader(eds)
-	randNID := dah.RowRoots[(len(dah.RowRoots)-1)/2][:namespace.NamespaceSize]
-	return eds, dah, randNID
+	max := nmt.MaxNamespace(dah.RowRoots[(len(dah.RowRoots))/2-1], ipld.NamespaceSize)
```
renaynay, line 203: did you mean to do `dah.RowRoots[(len(dah.RowRoots))/2-1` ?

share/get_test.go
Wondertan: ubernit: but might be a const at the top of the test, but still inside

share/getters/shrex_test.go
walldiss: yep, sneaky fix to always select last row in ods. Previous code relied on int division rounding, that I think could be confusing.

share/get_test.go
walldiss: It is single value and already in top of the test. Why would you want var for to store it?

share/getters/shrex_test.go
walldiss: I would rather wait for utility func, since it is not easy to sub 1 from multibyte int

blob/service.go
walldiss: `share.ErrNotFound` will be gone soon: https://github.com/celestiaorg/celestia-node/issues/2386

renaynay (approved):

Wondertan (approved):

Issue #2258, opened 2023-05-26
bug: discovery cannot find peers after network disconnect 


walldiss:
### Problem

I had experienced network going offline for few seconds while running node. It caused 2 problems related to discovery:
 - Network disconnect caused discovery to lose all peers and never recover the peer amount until node restart restart. 
 - After restart discovery was able to find enough peer, but lost them due to peer enforced disconnects shortly after. Even tho discovered peers has been disconnected, lowering size of discovery set, the new discovery search was not triggered.
 
<img width="861" alt="Screenshot 2023-05-26 at 16 25 47" src="https://github.com/celestiaorg/celestia-node/assets/13818348/d34f0ae9-110f-403e-9b2d-83649e2671f6">
<img width="861" alt="Screenshot 2023-05-26 at 16 27 59" src="https://github.com/celestiaorg/celestia-node/assets/13818348/87aee829-2f11-4236-b687-3b9adc9029ea">

[log.txt](https://github.com/celestiaorg/celestia-node/files/11575685/log1.txt)

 The problem in discovery is related to new search trigger not working correctly. Solutions options:
 -  Quick fix would be to trigger discovery once in period of time.
 - Debug the triggering condition logic and find the root cause

Wondertan:
> Even tho discovered peers has been disconnected, lowering size of discovery set, the new discovery search was not triggered.

How do you know the discovery was not triggered?

Wondertan:
> peer enforced disconnects shortly after

What do you mean by "peer enforced"? Remote peers disconnected from your node?

walldiss:
>I remember @distractedm1nd had a similar issue. We concluded that we were out of peers as they all got disconnected and in the backoff. Could this explain this issue?

I think the answer is no. I've made a metric for it.
<img width="861" alt="image" src="https://github.com/celestiaorg/celestia-node/assets/13818348/b64e0921-cd0a-4fad-9438-8ec7d111c55a">

>How do you know the discovery was not triggered?

There is no discovery events in logs. It is attached to the issue

>What do you mean by "peer enforced"? Remote peers disconnected from your node?

Yes, peer disconnected from my node. Check `removed peer` metric

Wondertan:
Ok. Makes sense. I am inclined to investigate the code and if we find no answer until Monday, we can do a quick fix.

ramin:
no longer happening

PR #2263, opened 2023-05-26
fix(share/discovery): timeout FindPeers operation

Wondertan:
Closes https://github.com/celestiaorg/celestia-node/issues/2258

FindPeers can get stuck sometimes for an indefinite time and so forth, stopping the whole discovery. We should stop and restart it.


Wondertan:
@walldiss, pls test to reconfirm the fix works

walldiss (dismissed):

share/p2p/discovery/discovery.go
```diff
@@ -79,7 +78,7 @@ func NewDiscovery(
 		connector:      newBackoffConnector(h, defaultBackoffFactory),
 		onUpdatedPeers: func(peer.ID, bool) {},
 		params:         params,
-		triggerDisc:    make(chan struct{}),
+		triggerDisc:    make(chan struct{}, 1), // buffer to memoize discovery request
```
renaynay, line 81: so do we think the problem here is that signal is sent when it's not ready to read? Is this the fix here or is the findPeersTimeout the fix? From what i understand from our discussion, it seems like the latter

share/p2p/discovery/metrics.go
```diff
@@ -13,8 +13,7 @@ import (
 )
 
 const (
-	discoveryEnougPeersKey  = "enough_peers"
-	discoveryFindCancledKey = "is_canceled"
+	discoveryEnougPeersKey = "enough_peers"
```
renaynay, line 16:
```suggestion
	discoveryEnoughPeersKey = "enough_peers"
```

share/p2p/discovery/metrics.go
```diff
@@ -130,7 +121,7 @@ func initMetrics(d *Discovery) (*metrics, error) {
 	return metrics, nil
 }
 
-func (m *metrics) observeFindPeers(ctx context.Context, canceled, isEnoughPeers bool) {
+func (m *metrics) observeFindPeers(ctx context.Context, isEnoughPeers bool) {
```
renaynay, line 124: so we can just assume now that if not enough peers, there was some sort of issue with disc

share/p2p/discovery/discovery.go
```diff
@@ -28,11 +28,10 @@ const (
...
 
-	// findPeersStuckWarnDelay is the duration after which discover will log an error message to
-	// notify that it is stuck.
-	findPeersStuckWarnDelay = time.Minute
+	// findPeersTimeout limits the FindPeers operation in time
+	findPeersTimeout = time.Minute
 
-	// defaultRetryTimeout defines time interval between discovery attempts.
+	// defaultRetryTimeout defines time interval between discovery and advertise attempts.
 	defaultRetryTimeout = time.Second
```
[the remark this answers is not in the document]
Wondertan: Good point, I would just remove the default

share/p2p/discovery/discovery.go
Wondertan: This buffer is an opportunistic fix. Honestly, there is no understanding of why it would help, but there is some evidence from @walldiss that it helps. We need to reconfirm this I think

share/p2p/discovery/metrics.go
Wondertan: We know that FindPeers can end earlier than the timeout in success case so no

walldiss:
I have run node with this PR and never seen stuck discovery since then

walldiss (dismissed):

share/p2p/discovery/discovery.go
walldiss: Based on my understanding of logs attached to original issue, is that at some moment discovery lost peers and wasn't triggered, but it should have been. It wasn't stuck at that point either, because there were no specific log lines for that. There was another period of time when it was stuck and there were periodic logs for that. 
So PR is addressing those 2 plus another one, that Advertise could error and retry time was low.

walldiss: Just realised that I didn't answered your Q 😆 . I have no evidence why discovery wasn't triggered, but my only guess is that it could be because the signal was lost due to no reader on other side of the channel at the time of signal.

renaynay: So I won't block on this because it's not _really_ a big deal to buffer the channel but it does suck bc I don't fully understand why this exact change is necessary yet. The best thing would've been to test the two solutions separately and then together so we could see which fix precisely is the actual fix but 🤷🏻

Wondertan: @walldiss, it would be nice if you could do that, specifically test the node without the buffer

walldiss: FindPeers was resulting in periodic logs, we had this case covered and now it should be never the case with your fix. There was another issue that has not resulted in such logs, so it wasn't stuck in FindPeers, but rather wasn't starting.

share/p2p/discovery/discovery.go
```diff
@@ -28,16 +28,15 @@ const (
...
 
-	// defaultRetryTimeout defines time interval between discovery attempts.
-	defaultRetryTimeout = time.Second
+	// retryTimeout defines time interval between discovery and advertise attempts.
+	retryTimeout = time.Second
 )
 
-// defaultRetryTimeout defines time interval between discovery attempts.
-var discoveryRetryTimeout = defaultRetryTimeout
+// retryTimeout defines time interval between discovery attempts, needed for tests
```
vgonkivs, line 38:
```suggestion
// discoveryRetryTimeout defines time interval between discovery attempts, needed for tests
```

Wondertan:
We discussed in sync that we don't want to merge the fix, we are not sure it is fixing anything, so we removed the buffer. We can continue testing to see if the discovery still gets stuck

vgonkivs (approved):

PR #2269, opened 2023-05-29
chore(blob): adapt changes

vgonkivs:
## Checklist

### DIFF
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -10,9 +10,6 @@ import (
 	"github.com/celestiaorg/nmt/namespace"
 )
 
-// versionIndex defines the position of the namespace version in the namespace
-const versionIndex = 0
-
 // Commitment is a Merkle Root of the subtree built from shares of the Blob.
 // It is computed by splitting the blob into shares and building the Merkle subtree to be included
 // after Submit.
@@ -64,7 +61,7 @@ func (p Proof) equal(input Proof) error {
 type Blob struct {
 	types.Blob
 
-	commitment Commitment
+	Commitment Commitment
 }
 
 // NewBlob constructs a new blob from the provided namespace.ID and data.
@@ -73,7 +70,7 @@ func NewBlob(shareVersion uint8, namespace namespace.ID, data []byte) (*Blob, er
 		return nil, fmt.Errorf("invalid size of the namespace id. got:%d, want:%d", len(namespace), appns.NamespaceSize)
 	}
 
-	ns, err := appns.New(namespace[versionIndex], namespace[versionIndex+1:])
+	ns, err := appns.New(namespace[appns.NamespaceVersionSize-1], namespace[appns.NamespaceVersionSize:])
 	if err != nil {
 		return nil, err
 	}
@@ -87,15 +84,10 @@ func NewBlob(shareVersion uint8, namespace namespace.ID, data []byte) (*Blob, er
 	if err != nil {
 		return nil, err
 	}
-	return &Blob{Blob: *blob, commitment: com}, nil
+	return &Blob{Blob: *blob, Commitment: com}, nil
 }
 
 // Namespace returns blob's namespace.
 func (b *Blob) Namespace() namespace.ID {
 	return append([]byte{uint8(b.NamespaceVersion)}, b.NamespaceId...)
-}
-
-// Commitment returns the commitment of current blob.
-func (b *Blob) Commitment() Commitment {
-	return b.commitment
 }
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -65,12 +65,12 @@ type Blob struct {
 }
 
 // NewBlob constructs a new blob from the provided namespace.ID and data.
-func NewBlob(shareVersion uint8, nID namespace.ID, data []byte) (*Blob, error) {
-	if len(nID) != appns.NamespaceSize {
+func NewBlob(shareVersion uint8, namespace namespace.ID, data []byte) (*Blob, error) {
+	if len(namespace) != appns.NamespaceSize {
 		return nil, fmt.Errorf("invalid size of the namespace id. got:%d, want:%d", len(nID), appns.NamespaceSize)
 	}
 
-	ns, err := appns.New(nID[0], nID[1:])
+	ns, err := appns.New(namespace[0], namespace[1:])
 	if err != nil {
 		return nil, err
 	}
@@ -87,7 +87,7 @@ func NewBlob(shareVersion uint8, nID namespace.ID, data []byte) (*Blob, error) {
 	return &Blob{Blob: *blob, commitment: com}, nil
 }
 
-// NamespaceID returns blobs namespaceId.
+// Namespace returns blob's namespace.
 func (b *Blob) Namespace() appns.Namespace {
 	return appns.Namespace{
 		ID:      b.Blob.GetNamespaceId(),
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -10,6 +10,9 @@ import (
 	"github.com/celestiaorg/nmt/namespace"
 )
 
+// versionIndex defines the position of the namespace version in the namespace
+const versionIndex = 0
+
 // Commitment is a Merkle Root of the subtree built from shares of the Blob.
 // It is computed by splitting the blob into shares and building the Merkle subtree to be included
 // after Submit.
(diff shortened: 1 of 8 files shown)


blob/blob.go
```diff
@@ -77,8 +88,11 @@ func NewBlob(_ uint8, nID namespace.ID, data []byte) (*Blob, error) {
 }
 
 // NamespaceID returns blobs namespaceId.
```
rootulp, line 90:
```suggestion
// Namespace returns the blob's namespace.
```

blob/blob.go
```diff
@@ -63,8 +65,17 @@ type Blob struct {
 }
 
 // NewBlob constructs a new blob from the provided namespace.ID and data.
-func NewBlob(_ uint8, nID namespace.ID, data []byte) (*Blob, error) {
-	blob, err := types.NewBlob(nID, data)
+func NewBlob(shareVersion uint8, nID namespace.ID, data []byte) (*Blob, error) {
```
rootulp, line 68: [optional] I've been using the following terminology to avoid confusion when referring to namespace variables:

version + ID = namespace

or longer form:

namespace version + namespace ID = namespace

so in this case, the second parameter is the entire namespace because it contains the version

```suggestion
func NewBlob(shareVersion uint8, namespace namespace.ID, data []byte) (*Blob, error) {
```

blob/blob_test.go
```diff
@@ -42,7 +42,7 @@ func TestBlob(t *testing.T) {
 		{
 			name: "verify nID",
 			expectedRes: func(t *testing.T) {
-				require.NoError(t, apptypes.ValidateBlobNamespaceID(blob[0].NamespaceID()))
+				require.NoError(t, apptypes.ValidateBlobNamespaceID(blob[0].Namespace()))
```
rootulp, line 45: [no change needed] oof `ValidateBlobNamespaceID` should be renamed so created https://github.com/celestiaorg/celestia-app/issues/1849

rootulp (approved):
LGTM

share/nid.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"fmt"
+
+	appns "github.com/celestiaorg/celestia-app/pkg/namespace"
+	"github.com/celestiaorg/nmt/namespace"
+)
+
+// NewNamespaceV0 takes variable size byte slice anc creates version 0 Namespace ID.
+// The bytes slice must be <= 10 bytes.
+func NewNamespaceV0(subNId []byte) (namespace.ID, error) {
```
Wondertan, line 12: @rootulp, we introduced this helper so that users don't have to struggle with defining the exact correct namespace size nor importing the app to get the size. It would be nice if the namespace in app had this user-friendly helper

blob/blob.go
Wondertan: That's helpful. Thank you!

blob/blob.go
```diff
@@ -2,12 +2,17 @@ package blob
...
 	"bytes"
+	"fmt"
 
+	appns "github.com/celestiaorg/celestia-app/pkg/namespace"
 	"github.com/celestiaorg/celestia-app/x/blob/types"
 	"github.com/celestiaorg/nmt"
 	"github.com/celestiaorg/nmt/namespace"
 )
 
+// versionIndex defines the position of the namespace version in the namespace
```
renaynay, line 13:
```suggestion
// versionIndex defines the position of the namespace version in the namespaceID
```

renaynay: lets call it `namespaceVersionIndex`

walldiss (approved):

blob/blob.go
```diff
@@ -67,10 +70,10 @@ type Blob struct {
 // NewBlob constructs a new blob from the provided namespace.ID and data.
 func NewBlob(shareVersion uint8, namespace namespace.ID, data []byte) (*Blob, error) {
 	if len(namespace) != appns.NamespaceSize {
-		return nil, fmt.Errorf("invalid size of the namespace id. got:%d, want:%d", len(nID), appns.NamespaceSize)
+		return nil, fmt.Errorf("invalid size of the namespace id. got:%d, want:%d", len(namespace), appns.NamespaceSize)
 	}
 
-	ns, err := appns.New(namespace[0], namespace[1:])
+	ns, err := appns.New(namespace[versionIndex], namespace[versionIndex+1:])
```
Wondertan, line 76: It can also be `NamespaceVersionSize-1` and `NamespaceVersionSize`, so that we always up to date with size changes(unlikely but its nice to all the things in sync)

blob/blob.go
```diff
@@ -76,22 +90,12 @@ func NewBlob(_ uint8, nID namespace.ID, data []byte) (*Blob, error) {
...
-	return b.Blob.Data
+// Namespace returns blob's namespace.
+func (b *Blob) Namespace() namespace.ID {
+	return append([]byte{uint8(b.NamespaceVersion)}, b.NamespaceId...)
 }
 
 // Commitment returns the commitment of current blob.
 func (b *Blob) Commitment() Commitment {
 	return b.commitment
 }
```
Wondertan, line 101: On second thought, we can just make the commitment Public. There is no need to encapsulate it as we always compute it with the constructor. We don't compute it when we interact with the Blob

renaynay (approved):

share/nid.go
rootulp: > with defining the exact correct namespace size

the namespace that gets used will still have a fixed length so even if NewNamespaceV0 is called with a subNId length of less than 10, the caller must know that the usable portion of the namespace ID is 10 bytes.

See https://go.dev/play/p/jNv2FkEG-ew

celestia-app uses a fixed namespace in tests like [here](https://github.com/celestiaorg/celestia-app/blob/ddaa261b9b0dec5f5ea10dbfab4c8a03c95d7e72/app/test/check_tx_test.go#L23) which I think may be more readable / understandable than the approach with `subNId`.

blob/blob.go
rootulp: preference to call it `namespace` here and not `namespaceID` because this `namespace` contains the version and ID. See https://github.com/celestiaorg/celestia-node/pull/2269#discussion_r1209406533

share/nid.go
Wondertan: > the caller must know that the usable portion of the namespace ID is 10 bytes.

That's a documentation problem(the func is not perfectly documented, like NMT). 

What I'm afraid I have to disagree with is that we should force the user to give _the exact size_ of the namespace ID. DevX-wise, this is horrible.
 
Let's count how many things a user with zero context should do to create a namespace:
 * import consts
 * find(there is insane number of consts now) and understand which const is correct 
 * preallocate the slice with the constant
 * fill in his wanted NamespaceID
 * import bytes
 * fill in the slice with zeroes.
 
In this PR's approach, you should know where to get the constructor(documentation issue) and give the wanted NamespaceID. That's all. Implementation details of how many bytes there are actually are not necessary for simple users

blob/blob.go
Wondertan: The only thing is that this got removed

Wondertan (approved):

share/nid.go
rootulp: celestia-app already contains https://github.com/celestiaorg/celestia-app/blob/ddaa261b9b0dec5f5ea10dbfab4c8a03c95d7e72/pkg/namespace/namespace.go#L41-L43 so celestia-node can do:

```diff
		if !sameNID {
+			nid := appns.MustNewV0(tmrand.Bytes(appns.NamespaceVersionZeroIDSize)).Bytes()
-			nid, err := share.NewNamespaceV0(tmrand.Bytes(7))
-			if err != nil {
-				return nil, err
-			}
			appBlob.NamespaceVersion = nid[0]
			appBlob.NamespaceID = nid[1:]
		}
```

however, I see your point that this approach relies on consumers importing a const from celestia-app. I'm open to adding the following function to celestia-app's namespace package given it seems slightly easier to use from DevX perspective:

```go
// NewV0 returns a new namespace with version 0 and id based on subID. subID
// must be <= 10 bytes.
//
// Note: the namespace ID will be right-padded with 0s if subID is < 10 bytes.
func NewV0(subID []byte) (Namespace, error) {
	if lenSubID := len(subID); lenSubID > NamespaceVersionZeroIDSize {
		return Namespace{}, fmt.Errorf("subID must be <= %v, but it was %v bytes", NamespaceVersionZeroIDSize, lenSubID)
	}

	id := make([]byte, NamespaceIDSize)
	copy(id[NamespaceVersionZeroPrefixSize:], subID)

	ns, err := New(NamespaceVersionZero, id)
	if err != nil {
		return Namespace{}, err
	}

	return ns, nil
}
```

Note: it has been slightly modified from the `NewNamespaceV0` in this PR because it returns the `Namespace` type rather than `namespace.ID`. Is that okay for your use case? The motivation for the change is that this function will be defined in the celestia-app `namespace` package which ideally has functions that deal with [Namespace](https://github.com/celestiaorg/celestia-app/blob/ddaa261b9b0dec5f5ea10dbfab4c8a03c95d7e72/pkg/namespace/namespace.go#L8-L11)

PR #2286, opened 2023-05-31
feat(blob): add marshaling for blob and proof

vgonkivs:
## Overview
Add marshaler fro blob and proof

## Checklist


blob/blob.go
```diff
@@ -57,11 +62,47 @@ func (p Proof) equal(input Proof) error {
 	return nil
 }
 
+type jsonProof struct {
+	Start int      `json:"start"`
+	End   int      `json:"end"`
+	Nodes [][]byte `json:"nodes"`
+}
```
walldiss, line 69: I curious, why ignoreMaxNamespace param is not included? I think if it is added, it will remove dependancies on node setting (ipld.NMTIgnoreMaxNamespace) on unmashal caller side.

Would support for non-inclusion proof be used with marshaller in future? If so maybe add `leafhash` []byte now, so it till support non-inclusion roofs right away.

distractedm1nd (dismissed):

blob/blob.go
Wondertan: We don't need to support Blob Non-Inclusion proofs, but only for the whole namespace. I will send you a link with the discussion about that

walldiss: Thank you for info! So please, dismiss my second question. I still don't know the usecase for marshal / unmarshal in blob, so not sure if first question is relevant.

Wondertan: ignoreMaxNamespace is a protocol constant, so there is no need to pass it

blob/blob.go
```diff
@@ -91,3 +132,47 @@ func NewBlob(shareVersion uint8, namespace namespace.ID, data []byte) (*Blob, er
...
+		return nil, err
+	}
+
+	return json.Marshal(&struct {
+		Blob       json.RawMessage `json:"blob"`
+		Commitment []byte          `json:"commitment"`
+	}{
+		Blob:       data,
+		Commitment: b.Commitment,
+	},
```
Wondertan, line 154: Dedup as private type?

Wondertan: I would also recommend having Blob JSON marshaling without the inner Blob.
So it's not Blob{blob: {}}. 
We will eventually make Blob one single struct without embedding the app's blob. Once we come to that point, it would be super weird to have Blob{blob: {}}, and changing that would be breaking, so let's make it correct from the beginning.

We could also avoid importing tmjson this way (?)

Wondertan (dismissed):
Nice

renaynay (dismissed):

renaynay (approved):

blob/blob.go
```diff
@@ -91,3 +130,35 @@ func NewBlob(shareVersion uint8, namespace namespace.ID, data []byte) (*Blob, er
 func (b *Blob) Namespace() namespace.ID {
 	return append([]byte{uint8(b.NamespaceVersion)}, b.NamespaceId...)
 }
+
+type jsonBlob struct {
+	Namespace    namespace.ID `json:"namespace"`
+	Data         []byte       `json:"data"`
+	ShareVersion uint32       `json:"shareVersion"`
```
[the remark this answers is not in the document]
Wondertan: There is no conventions on this btw, but it looks nicer , I agree

Wondertan (approved):

blob/blob.go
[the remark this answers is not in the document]
distractedm1nd: In golang yes, in rust client it gets very annoying when you aren't - so we should 100% stick to snake case

PR #2290, opened 2023-05-31
chore(nodebuilder): decrease lifecycle timeout for LN

Wondertan:
LNs don't need that much time to run

### DIFF
--- a/nodebuilder/node_test.go
+++ b/nodebuilder/node_test.go
@@ -136,6 +136,9 @@ func StartMockOtelCollectorHTTPServer(t *testing.T) (string, func()) {
 }
 
 func TestEmptyBlockExists(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
 	var test = []struct {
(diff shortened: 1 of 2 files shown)


vgonkivs (approved):

nodebuilder/node_test.go
```diff
@@ -136,6 +136,9 @@ func StartMockOtelCollectorHTTPServer(t *testing.T) (string, func()) {
 }
 
 func TestEmptyBlockExists(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
```
renaynay, line 139: no timeout for test?

distractedm1nd (approved):

renaynay (approved):

nodebuilder/node_test.go
Wondertan: This test does not involve any blocking operations

Wondertan: This test does not involve any blocking operations

PR #2291, opened 2023-05-31
fix(gateway): allowing header requests for localhead and localhead+1

distractedm1nd:
Ref: https://github.com/rollkit/rollkit/pull/953

This PR allows for requesting the localhead and localhead+1 over the gateway.

### DIFF
--- a/api/gateway/availability.go
+++ b/api/gateway/availability.go
@@ -34,7 +34,7 @@ func (h *Handler) handleHeightAvailabilityRequest(w http.ResponseWriter, r *http
 		writeError(w, http.StatusInternalServerError, heightAvailabilityEndpoint, err)
 		return
 	}
-	if headHeight := int(head.Height()); headHeight < height {
+	if headHeight := int(head.Height()); height >= headHeight+1 {
 		err = fmt.Errorf(
 			"current head local chain head: %d is lower than requested height: %d"+
 				" give header sync some time and retry later", headHeight, height)
--- a/api/gateway/header.go
+++ b/api/gateway/header.go
@@ -76,10 +76,10 @@ func (h *Handler) performGetHeaderRequest(
 		writeError(w, http.StatusInternalServerError, heightAvailabilityEndpoint, err)
 		return nil, err
 	}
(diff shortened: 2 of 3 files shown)


Wondertan (approved):

renaynay (dismissed):

Wondertan (changes requested):
There is also `handleHeightAvailabilityRequest` and `performGetHeaderRequest` where we should change this

distractedm1nd (commented):
@Wondertan performGetHeaderRequest was already covered, but ive added 1handleHeightAvailabilityRequest`

renaynay (dismissed):

Wondertan:
@distractedm1nd, I mixed up `performGetHeaderRequest` with `getShares` which we still need to cover

api/gateway/availability.go
```diff
@@ -34,7 +34,7 @@ func (h *Handler) handleHeightAvailabilityRequest(w http.ResponseWriter, r *http
 		writeError(w, http.StatusInternalServerError, heightAvailabilityEndpoint, err)
 		return
 	}
-	if headHeight := int(head.Height()); headHeight < height {
+	if headHeight := int(head.Height()); height >= headHeight+1 {
```
Wondertan, line 37: Looking into the error that happened for the [Rollkit](https://github.com/rollkit/rollkit/pull/953#issuecomment-1562215384). They requested a height of 27, and the node head was 26

27 >= 26+1

So it will still error. 

Along with fixing this, I propose to reformat this to IMO more readable format:
`localHeight +1` < `height` (If `localHeight+1` is less than the requested `height` - error out.)

26+1 < 27 - does not hits the error case

Wondertan (changes requested):
Optionally, I would move this into a helper func

api/gateway/availability.go
renaynay: yes

renaynay: dream team ⭐

Wondertan:
No more needed with #2297

PR #2297, opened 2023-06-01
fix(gateway): remove the GetByHeight hussle and rely on Module's logic

Wondertan:
Supersedes #2291

### DIFF
--- a/api/gateway/header.go
+++ b/api/gateway/header.go
@@ -1,6 +1,7 @@
 package gateway
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"net/http"
@@ -9,6 +10,7 @@ import (
 	"github.com/gorilla/mux"
 
 	"github.com/celestiaorg/celestia-node/header"
+	modheader "github.com/celestiaorg/celestia-node/nodebuilder/header"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


api/gateway/header.go
```diff
@@ -70,24 +72,34 @@ func (h *Handler) performGetHeaderRequest(
...
+
+	header, err := mod.GetByHeight(ctx, uint64(height))
 	if err != nil {
-		writeError(w, http.StatusInternalServerError, endpoint, err)
-		return nil, err
+		return nil, http.StatusInternalServerError, err
 	}
-	return header, nil
+
+	return header, 0, nil
```
vgonkivs, line 104: Even if the second param won't be used in case of successful handlings, I would propose to return `http.StatusOK`, just to keep the consistency, because `0` confuses me a lot

api/gateway/share.go
```diff
@@ -93,33 +91,14 @@ func (h *Handler) handleDataByNamespaceRequest(w http.ResponseWriter, r *http.Re
...
-		}
-		header, err = h.header.GetByHeight(ctx, height)
-	}
-	if err != nil {
-		return nil, 0, err
-	}
-	// perform request
 	shares, err := h.share.GetSharesByNamespace(ctx, header.DAH, nID)
-	return shares.Flatten(), header.Height(), err
+	return shares.Flatten(), http.StatusInternalServerError, err
```
vgonkivs, line 101: if `GetSharesByNamespace` fails, you can get `nil` instead of shares and it will panic

vgonkivs (approved):

api/gateway/share.go
Wondertan: True, so the back was there before

walldiss (approved):

distractedm1nd (approved):

Wondertan:
Zombie PR, closing

PR #2307, opened 2023-06-02
chore(state): extend error handling in core accessor

vgonkivs:
## Overview

Extend error handling in core accessor

## Checklist

### DIFF
--- a/blob/service.go
+++ b/blob/service.go
@@ -67,7 +67,7 @@ func (s *Service) Submit(ctx context.Context, blobs []*Blob) (uint64, error) {
 	if err != nil {
 		return 0, err
 	}
-	return uint64(resp.Height), nil
+	return uint64(resp.Height), err
 }
 
 // Get retrieves all the blobs for given namespaces at the given height by commitment.
--- a/blob/service.go
+++ b/blob/service.go
@@ -67,7 +67,7 @@ func (s *Service) Submit(ctx context.Context, blobs []*Blob) (uint64, error) {
 	if err != nil {
 		return 0, err
 	}
-	return uint64(resp.Height), err
+	return uint64(resp.Height), nil
 }
 
 // Get retrieves all the blobs for given namespaces at the given height by commitment.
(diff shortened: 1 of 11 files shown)


blob/service.go
```diff
@@ -70,7 +70,7 @@ func (s *Service) Submit(ctx context.Context, blobs []*Blob) (uint64, error) {
 	if err != nil {
 		return 0, err
 	}
-	return uint64(resp.Height), nil
+	return uint64(resp.Height), err
```
Wondertan: The error is always nil here

state/helpers.go
```diff
@@ -1,21 +1,64 @@
...
 		return status.Error(codes.Unauthenticated, resp.Log)
-	case sdk_errors.ErrKeyNotFound.ABCICode():
+	case sdk_errors.ErrKeyNotFound:
 		return status.Error(codes.NotFound, resp.Log)
 	default:
 		return status.Error(codes.Unknown, resp.Log)
 	}
 }
+
+func codeToError(code uint32) error {
```
Wondertan, line 25: Use https://github.com/cosmos/cosmos-sdk/blob/main/errors/errors.go#L75 instead

Wondertan (changes requested):

state/core_access.go
```diff
@@ -177,6 +178,10 @@ func (ca *CoreAccessor) SubmitPayForBlob(
 		ca.lastPayForBlob = time.Now().UnixMilli()
 		ca.payForBlobCount++
 	}
+
+	if response.Code != 0 {
```
Wondertan, line 182: very optional nit: can be coupled with the if above

Wondertan (approved):

state/core_access.go
vgonkivs: I would suggest keeping it separately: the previous statement contains metrics handling, and the new one is error handling.

walldiss (approved):

PR #2308, opened 2023-06-02
feat(cmd/rpc): adding blob module to rpc cli and fixing namespace parsing

distractedm1nd:
This PR does the following:

### Adds blob module support
`blob.Submit` has been added with custom parsing for posting a single blob.
`blob.GetAll` also only takes a single namespace id for now. This will make devx easier

### Adds --print-request flag
Now the request can also be displayed along with the json response. This will both help for debugging and building things that need to build the requests themselves

### Fixes namespace parsing
Namespaces are now encoded correctly again. In addition, if you pass a namespace of 8 bytes, it will zero-pad and use share/nID version 0.

### Fixes `state.SubmitPayForBlob`
Another issue was the change to an array. We now submit an array of a single blob, and changed the order of the arguments.
The order is now: fee, gasLimit, namespaceID, blob data

### Adds formatted output
JQ is no longer needed to have an indented output. But colors are missing

### DIFF
--- a/cmd/celestia/rpc.go
+++ b/cmd/celestia/rpc.go
@@ -18,7 +18,11 @@ import (
 	"github.com/cosmos/cosmos-sdk/types"
 	"github.com/spf13/cobra"
 
+	apptypes "github.com/celestiaorg/celestia-app/x/blob/types"
+	"github.com/celestiaorg/nmt/namespace"
+
 	"github.com/celestiaorg/celestia-node/api/rpc/client"
+	"github.com/celestiaorg/celestia-node/blob"
 	"github.com/celestiaorg/celestia-node/state"
 )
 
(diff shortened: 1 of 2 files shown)


distractedm1nd:
image
@jcstein

Wondertan (dismissed):
yoLOLack

vgonkivs (approved):

cmd/celestia/rpc.go
```diff
@@ -322,6 +423,29 @@ func parseSignatureForHelpstring(methodSig reflect.StructField) string {
...
+		// otherwise, it's just a base64 string
+		nID, err = base64.StdEncoding.DecodeString(param)
+		if err != nil {
+			return nil, fmt.Errorf("error decoding namespace ID: %w", err)
+		}
+	}
+	// if the namespace ID is 8 bytes, add v0 share + namespace prefix and zero pad
+	if len(nID) == 8 {
+		nID = append(make([]byte, 21), nID...)
+	}
```
Wondertan, line 445: share.NewV0Namespace

renaynay (approved):

Wondertan (approved):

Issue #2174, opened 2023-05-08
`getLeavesWithProofsByNamespace:could not retrieve node`


jcstein:
### Celestia Node version

v0.9.2

### OS

mac/linux

### Steps to reproduce it

This is from running a rollup and the full logs can be found [here](https://drive.google.com/file/d/1bWFBiqsH9dY3-3IjL0uJ7p_j0Zl6vxe3/view?usp=share_link)

### Expected result

shares are retrieved, no error

### Actual result

`ERROR   ipld    ipld/namespace_data.go:216  getLeavesWithProofsByNamespace:could not retrieve node  {"nID": "\ufffd\ufffd\ufffdy\ufffdq\u0016\ufffd", "pos": 0, "err": "context canceled"}`

jcstein:
Seeing this occur on v0.11.0-rc1 of node:

```Shell
ERROR	ipld	ipld/namespace_data.go:216	getLeavesWithProofsByNamespace:could not retrieve node	{"nID": "\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0008\ufffd\ufffdy\ufffdq\u0016\ufffd", "pos": 0, "err": "context canceled"}
```

Wondertan:
@walldiss, @distractedm1nd, seems like we cancel ShrexND and move to IPLD and then it also gets canceled

jcstein:
cc @tuxcanfly

PR #2310, opened 2023-06-02
fix(getter/cascade): cascade getter should return fast on context cancel

walldiss:
## Overview

Also fixes incorrect errors.Is for `share.ErrNamespaceNotFound`

Resolves https://github.com/celestiaorg/celestia-node/issues/2174

### DIFF
--- a/share/getters/cascade.go
+++ b/share/getters/cascade.go
@@ -121,16 +121,18 @@ func cascadeGetters[V any](
 		getCtx, cancel := ctxWithSplitTimeout(ctx, len(getters)-i, 0)
 		val, getErr := get(getCtx, getter)
 		cancel()
-		if getErr == nil {
-			return val, nil
-		}
-		if errors.Is(getErr, share.ErrNamespaceNotFound) || errors.Is(getErr, context.Canceled) {
-			return zero, getErr
+		if getErr == nil || errors.Is(getErr, share.ErrNamespaceNotFound) {
+			return val, getErr
 		}
 
-		if !errors.Is(getErr, errOperationNotSupported) {
-			err = errors.Join(err, getErr)
-			span.RecordError(getErr, trace.WithAttributes(attribute.Int("getter_idx", i)))
+		if errors.Is(getErr, errOperationNotSupported) {
+			continue
+		}
+
+		err = errors.Join(err, getErr)
+		span.RecordError(getErr, trace.WithAttributes(attribute.Int("getter_idx", i)))
+		if errors.Is(getErr, context.Canceled) {
+			return zero, err
 		}
 	}
 	return zero, err
--- a/share/getters/cascade.go
+++ b/share/getters/cascade.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/hex"
 	"errors"
-
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
 
@@ -125,7 +124,7 @@ func cascadeGetters[V any](
 		if getErr == nil {
 			return val, nil
 		}
-		if errors.Is(share.ErrNamespaceNotFound, getErr) {
+		if errors.Is(getErr, share.ErrNamespaceNotFound) || errors.Is(getErr, context.Canceled) {
 			return zero, getErr
 		}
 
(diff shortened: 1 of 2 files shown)


share/getters/cascade.go
```diff
@@ -125,7 +124,7 @@ func cascadeGetters[V any](
 		if getErr == nil {
 			return val, nil
 		}
-		if errors.Is(share.ErrNamespaceNotFound, getErr) {
+		if errors.Is(getErr, share.ErrNamespaceNotFound) || errors.Is(getErr, context.Canceled) {
```
Wondertan, line 127: Do we want to track user deadline here?

walldiss: Good find! Yeah, we also want to keep previous errors. I've reordered err handlers

renaynay (approved):

share/getters/cascade.go
```diff
@@ -122,16 +121,18 @@ func cascadeGetters[V any](
...
+		if errors.Is(getErr, errOperationNotSupported) {
+			continue
 		}
 
-		if !errors.Is(getErr, errOperationNotSupported) {
-			err = errors.Join(err, getErr)
-			span.RecordError(getErr, trace.WithAttributes(attribute.Int("getter_idx", i)))
+		err = errors.Join(err, getErr)
+		span.RecordError(getErr, trace.WithAttributes(attribute.Int("getter_idx", i)))
+		if errors.Is(getErr, context.Canceled) {
```
Wondertan, line 134: So where is deadline check?

renaynay: ah you also want context deadline

share/getters/cascade.go
```diff
@@ -122,16 +121,18 @@ func cascadeGetters[V any](
...
+			continue
 		}
 
-		if !errors.Is(getErr, errOperationNotSupported) {
-			err = errors.Join(err, getErr)
-			span.RecordError(getErr, trace.WithAttributes(attribute.Int("getter_idx", i)))
+		err = errors.Join(err, getErr)
+		span.RecordError(getErr, trace.WithAttributes(attribute.Int("getter_idx", i)))
+		if ctx.Err() != nil {
+			return zero, err
```
Wondertan, line 135:
```suggestion
			return zero, ctx.Err()
```

walldiss: We return aggregated error, because there could be legit errors from few getters before Ctx got cancled/timedout

share/getters/cascade.go
```diff
@@ -122,16 +121,18 @@ func cascadeGetters[V any](
...
+		if errors.Is(getErr, errOperationNotSupported) {
+			continue
 		}
 
-		if !errors.Is(getErr, errOperationNotSupported) {
-			err = errors.Join(err, getErr)
-			span.RecordError(getErr, trace.WithAttributes(attribute.Int("getter_idx", i)))
+		err = errors.Join(err, getErr)
+		span.RecordError(getErr, trace.WithAttributes(attribute.Int("getter_idx", i)))
+		if ctx.Err() != nil {
```
Wondertan, line 134: Let's do this check right after the call. If user cancelled, there is no point to go further.

Wondertan: Ok, if you want to return aggregated error, then it's ok to leave it like this

share/getters/cascade.go
Wondertan: Ok, let's return aggregated. I am still not sure if there is any value from returning it when op is cancelled, but I want to see this in practice more.

renaynay: append the ctx.Err 👍🏻

Wondertan: I assume the last getErr will be the context one so it will be already appended.

walldiss: We aggregate to see what errors happen before timeout happen. Aggregating context error also helps to identify what getter was in use when error happen. It will help us debug errors in cascade

walldiss: Also we collect error span in case context Error, which is also nice to have.

Wondertan (approved):

renaynay (dismissed):

Wondertan (approved):

renaynay (approved):

Issue #2321, opened 2023-06-05
panic: Error creating blob: invalid size of the namespace id. got:10, want:29


jcstein:
### Celestia Node version

v0.11.0-rc2

### OS

macOS

### Install tools

https://docs.celestia.org/nodes/celestia-node/ for Mac (Apple) x Arabica

### Steps to reproduce it

fund account:
```bash
celestia rpc state Balance
{
  "jsonrpc": "2.0",
  "result": {
    "denom": "utia",
    "amount": "100000000"
  },
  "id": 1
}
```

submit PFB with 10 byte namespace:
```bash
celestia rpc blob Submit 0x42690c204d39600fddd3 0xf1f20ca8007e910a3bf8b2e61da0f26bca07ef78717a6ea54165f5
```

### Expected result

```bash
{
 “jsonrpc”: “2.0”,
 “result”: BLOCK_HEIGHT_OF_PFB,
 “id”: 1
}
```

### Actual result

panic: Error creating blob: invalid size of the namespace id. got:10, want:29

### Relevant log output

```Shell
celestia rpc blob Submit 0x42690c204d39600fddd3 0xf1f20ca8007e910a3bf8b2e61da0f26bca07ef78717a6ea54165f5

panic: Error creating blob: invalid size of the namespace id. got:10, want:29

goroutine 1 [running]:
main.parseParams({0x16aea7822, 0x6}, {0x1400059c820, 0x2, 0x0?})
	/Users/joshstein/celestia-node/cmd/celestia/rpc.go:151 +0x1920
main.glob..func5(0x109348320?, {0x1400059c800?, 0x4?, 0x4?})
	/Users/joshstein/celestia-node/cmd/celestia/rpc.go:100 +0x70
github.com/spf13/cobra.(*Command).execute(0x109348320, {0x1400059c780, 0x4, 0x4})
	/Users/joshstein/go/pkg/mod/github.com/spf13/cobra@v1.6.1/command.go:920 +0x5ac
github.com/spf13/cobra.(*Command).ExecuteC(0x109348040)
	/Users/joshstein/go/pkg/mod/github.com/spf13/cobra@v1.6.1/command.go:1044 +0x340
github.com/spf13/cobra.(*Command).Execute(...)
	/Users/joshstein/go/pkg/mod/github.com/spf13/cobra@v1.6.1/command.go:968
github.com/spf13/cobra.(*Command).ExecuteContext(...)
[...]

Issue #2301, opened 2023-06-01
Rename `nID` to `namespace` where applicable


rootulp:
## Context

https://github.com/celestiaorg/celestia-node/blob/c77f6db35fe57ccef4110d0c7815e53a24921948/share/get.go#L52

## Problem
In a few locations in this repo, `nID: namespace.ID` is used when I think a namespace is being passed (i.e. version + ID).

## Proposal

1. Rename `nID` to `namespace` or `ns` where applicable. 
2. Replace the usage of the NMT type [`namespace.ID`](https://github.com/celestiaorg/nmt/blob/5466a618180e5127dc2beef2e9641849953cc1c2/namespace/id.go#L8) with celestia-app type [Namespace](https://github.com/celestiaorg/celestia-app/blob/348f795d8635324db4cca0367b599b5edc5c2fdb/pkg/namespace/namespace.go#L8)

Wondertan:
Re 2. We want to move on Namespace from the app once it becomes a single slice. IIRC, with aligned on that in sync. When and how can we do it? Node team can help there

Wondertan:
> I don't think celestia-app can make this change in the short-term b/c it's breaking and we've already cut celestia-app v1.0.0-rcs. 

That's fair. The good thing is the change is not consensus-breaking, so we can still do it. Adding it to my list of things to take during share pkg extraction.

Meanwhile, cleaning up and defining the type in celestia-node makes total sense. Assigning this on me.

Wondertan:
@rootulp, thinking about this more, the namespace in NMT should be called just the namespace then as well because NMT commits to the version as well thus, it operates over namespaces rather than IDs. I am not sure anymore it is worth renaming in celestia-node only without nmt

PR #2325, opened 2023-06-05
fix: parseNamespace for <= 10 byte namespace IDs

rootulp:
Closes https://github.com/celestiaorg/celestia-node/issues/2321


cmd/celestia/rpc_test.go
```diff
@@ -0,0 +1,52 @@
...
+			want:    namespace.ID{0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x42, 0x69, 0xc, 0x20, 0x4d, 0x39, 0x60, 0xf, 0xdd, 0xd3},
+			wantErr: false,
+		},
+		// HACKHACK: This test case is disabled because it fails.
+		// {
+		// 	name:    "11 byte hex encoded namespace returns error",
+		// 	param:   "0x42690c204d39600fddd3a3",
+		// 	want:    namespace.ID{},
+		// 	wantErr: true,
+		// },
```
rootulp, line 37: I think there are other bugs lurking in `parseNamespace` but I wanted to help unblock @jcstein

cmd/celestia/rpc.go
```diff
@@ -440,8 +441,8 @@ func parseNamespace(param string) (namespace.ID, error) {
 			return nil, fmt.Errorf("error decoding namespace ID: %w", err)
 		}
 	}
-	// if the namespace ID is 8 bytes, add v0 share + namespace prefix and zero pad
-	if len(nID) == 8 {
+	// if the namespace ID is <= 10 bytes, add v0 share + namespace prefix and zero pad
```
rootulp, line 444: [question] does the RPC CLI let user's provide a namespace version? If not, is support for that being tracked somewhere? It's not urgent b/c there is only one namespace version (`0`) but I think we may want to expose it as an optional CLI flag b/c it impacts how namespaces are constructed here

Wondertan: @distractedm1nd

cmd/celestia/rpc.go
```diff
@@ -440,8 +441,8 @@ func parseNamespace(param string) (namespace.ID, error) {
...
+	if len(nID) <= appns.NamespaceVersionZeroIDSize {
```
Wondertan, line 445: This check happens in the constructor so it's redundant

Wondertan:
@rootulp, note that I am waiting for @distractedm1nd to return from sick leave and look deeper into this PR, as he is an author. I am not deep in the context of rpc cli

cmd/celestia/rpc.go
Wondertan: I assumed returning an error here was correct. Do we want to avoid throwing because we want to support full namespaces besides sub-namespaces(<=10)?

If so, I suggest adding logic that detects if a namespace is full or sub and returns instantiated namespace. 
Something like `share.DecodeNamespace([]byte) Namespace`, would enable anyone to take a namespace of any form and simply use it.

cmd/celestia/rpc.go
distractedm1nd: It's not tracked and not urgent until the next namespaceID versions are out. The CLI is here for providing the easiest way to interact/play with the RPC so I wouldn't add it until it is absolutely necessary

distractedm1nd (commented):
Sent @rootulp a fix for the missing test case and added two additional ones, now the coverage is pretty good. Once applied i will approve

cmd/celestia/rpc.go
Wondertan: Ideally, we would be able to accept all these 3 cases, and something like Decode/Parse would help. Ideally, we would also move the Parse/Decode helper next to app's namespace. How does that sound to you?

rootulp: > Ideally, we would be able to accept all these 3 cases

Are these the cases you're considering:

version | ID | namespace | example
--- | --- | --- | ---
✅ | ✅ | computed based on other params | `celestia rpc blob Submit 0x42690c204d39600fddd3 --namespace-version 0`
❌ (defaults to `0`) | ✅  | computed based on other params | `celestia rpc blob Submit 0x42690c204d39600fddd3`
❌ | ❌ | ✅ | `celestia rpc blob Submit 0x0000000000000000000000000000000000000001010101010101010101`

If so, I think it's confusing that the param in most cases accepts a namespace ID but in the last case accepts a namespace.

> Ideally, we would also move the Parse/Decode helper next to app's namespace

I'm onboard with extracting a helper from celestia-node to celestia-app if that's what is suggested. And/or moving it to a share or namespace package inside this repo. Defer to ya'll on where `parseNamespace` should live.

distractedm1nd (approved):

cmd/celestia/rpc.go
```diff
@@ -440,12 +441,14 @@ func parseNamespace(param string) (namespace.ID, error) {
...
 		nID, err = share.NewNamespaceV0(nID)
 		if err != nil {
 			return nil, err
 		}
+	} else if len(nID) < appns.NamespaceSize {
+		return nil, fmt.Errorf("passed namespace is too large")
```
walldiss, line 451: Looks like you meany `>` instead of `<`
```suggestion
	} else if len(nID) > appns.NamespaceSize {
		return nil, fmt.Errorf("passed namespace is too large")
```

rootulp:
Synced w/ @distractedm1nd offline and `parseNamespace` is intended to only be used for the **namespace ID** portion of a namespace so a test case in this PR is incorrect.

I'll push a few commits to this PR that renames things like namespace to **namespace ID** where applicable but we should really do https://github.com/celestiaorg/celestia-node/issues/2301 ASAP.

rootulp:
The scope of this PR has increased b/c it wasn't clear to me what the original intended behavior was. Would appreciate extra attention when reviewing b/c this couples a fix, refactor, and behavior change (left pad instead of right pad)

cmd/celestia/rpc.go
rootulp: Also we really need to clarify to users if they should be providing a 10 byte namespace ID, a partial namespace ID (e.g. 8 bytes), or a full namespace.

Wondertan:
> Synced w/ @distractedm1nd offline and parseNamespace is intended to only be used for the namespace ID portion of a namespace so a test case in this PR is incorrect.

@distractedm1nd, why do we want to disable versioning here?

cmd/celestia/rpc.go
Wondertan: passing this to @distractedm1nd as you will have a call today

Wondertan: From our sync discussion today:
> `namespace.Parse/Decode([]byte) Namespace` Which accepts:
    * Ver + Padded ID
    * Padded ID
    * ID
    * Maybe Ver + ID

> The text encoding is out of the scope of this function, so parseNamespace will stay, but just use it.

Wondertan (dismissed):
LGTM

cmd/celestia/rpc.go
```diff
@@ -112,17 +113,17 @@ func parseParams(method string, params []string) []interface{} {
 		}
 		parsedParams[0] = root
 		// 2. NamespaceID
-		nID, err := parseNamespace(params[1])
+		nID, err := parseV0NamespaceID(params[1])
```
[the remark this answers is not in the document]
distractedm1nd: The idea here is that it is actually only parsing the ID portion of the namespace: hence NamespaceID instead of Namespace.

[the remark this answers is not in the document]
Wondertan: Ah, right, forgot about that, then ok

Wondertan (commented):
I still don't get whats the point of panicking on user input though

cmd/celestia/rpc_test.go
```diff
@@ -0,0 +1,82 @@
...
+func Test_parseNamespaceID(t *testing.T) {
+	type testCase struct {
+		name    string
+		param   string
+		want    namespace.ID
+		wantErr bool
+	}
+	testCases := []testCase{
+		{
+			name:  "8 byte hex encoded namespace ID gets right padded",
```
rootulp, line 20: [blocking] this claims to right pad but the implementation claims to left pad. I think left padding is desired behavior

cmd/celestia/rpc.go
```diff
@@ -418,32 +419,41 @@ func parseSignatureForHelpstring(methodSig reflect.StructField) string {
...
+		return nil, err
+	}
+
+	if len(userBytes) > appns.NamespaceVersionZeroIDSize {
+		return nil, fmt.Errorf(
+			"namespace ID %v is too large to be a v0 namespace, want <= %v bytes",
+			userBytes, appns.NamespaceVersionZeroIDSize,
+		)
+	}
+	// if the namespace ID is <= 10 bytes, left pad it with 0s
```
rootulp, line 437: [blocking] does this actually left pad? I think it right pads b/c the first test case

cmd/celestia/rpc_test.go
distractedm1nd: It actually left pads, but the unit tests are failing! Didn't catch until now

cmd/celestia/rpc.go
distractedm1nd: thanks

cmd/celestia/rpc_test.go
distractedm1nd: fixed in [4a098ab](https://github.com/celestiaorg/celestia-node/pull/2325/commits/4a098abbd87ef34d5dbb7315a0d4cf52d55b379f)

Wondertan (approved):
Good catch @rootulp

renaynay (approved):

Issue #2328, opened 2023-06-06
gateway: submitPFB endpoint should return tx hash


Wondertan:
In case we [fail to submit PFB](https://github.com/celestiaorg/celestia-node/blob/main/api/gateway/state.go#L151), we still get the non-nil response and can access the TxHash field from it, which we can return together with error.

Note that this breaks the API and go-cnc will have to be updated as well to support it

PR #2329, opened 2023-06-06
!feat(api/gateway): return txdata from submitPFB if it is present

walldiss:
## Overview

Closes https://github.com/celestiaorg/celestia-node/issues/2328

### DIFF
--- a/api/gateway/state.go
+++ b/api/gateway/state.go
@@ -45,6 +45,13 @@ type submitPFBRequest struct {
 	GasLimit    uint64 `json:"gas_limit"`
 }
 
+// submitPFBResponse represents a response to submit a PayForBlob
+// transaction.
+type submitPFBResponse struct {
+	Tx    *state.TxResponse `json:"tx,omitempty"`
+	Error string            `json:"err,omitempty"`
+}
+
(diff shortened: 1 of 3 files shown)


api/gateway/state.go
```diff
@@ -149,16 +156,27 @@ func (h *Handler) handleSubmitPFB(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil && txResp == nil {
+		// no tx data to return
 		writeError(w, http.StatusInternalServerError, submitPFBEndpoint, err)
-		return
 	}
-	resp, err := json.Marshal(txResp)
+
+	resp := submitPFBResponse{
+		Tx:    txResp,
+		Error: err.Error(),
```
Wondertan, line 166: We actually don't need to return an error so we can just return txResp. In SubmitPayForBlob we derive the error from the txResponse, so by returning the error we just send duplicate data.

walldiss: Ah, ok! TIL

Wondertan: > Kinda weird API
💯

walldiss: Remved redundant error

Wondertan (approved):

PR #2332, opened 2023-06-06
fix(nodebuilder): Fix dns lookup

### LINKED ISSUE #2333: schema not working as intended in 0.10.4
### Celestia Node version

0.10.4

### OS

ubuntu 20.04

### Install tools

_No response_

### Others

_No response_

### Steps to reproduce it

so it seems some forms of --core.ip flag dont work, like  `localhost` and for some `http://localhost` and `http://localhost:26657/` and adding 127.0.0.1 seems to work

while the last 2 work for some, i would think it has to do with their hosts file, eg if they have:
 /etc/hosts  127.0.0.1 localhost

### Expected result

the --core.ip flag should work with just localhost, or with http/https normally

### Actual result

Error: failed to resolve DNS record: lookup : no such host

and 

Error: invalid IP address or hostname given: ::1 

### Relevant log output

_No response_

### Notes

_No response_

walldiss:
## Overview

Cleanup core ip parsing logic

Closes #2333


libs/utils/address.go
```diff
@@ -30,11 +31,14 @@ func ValidateAddr(addr string) (string, error) {
 	}
 
 	if ip := net.ParseIP(addr); ip == nil {
-		_, err = net.LookupHost(addr)
+		addrs, err := net.LookupHost(addr)
 		if err != nil {
 			return addr, fmt.Errorf("could not resolve hostname or ip: %w", err)
 		}
+		if len(addrs) == 0 {
+			return addr, errors.New("no IP addresses found for DNS record")
```
Wondertan, line 39: optional nit: would be nice to include actual addresses in the error

Wondertan (approved):
We will get it on the 3rd attempt!

libs/utils/address.go
walldiss: Added addr, thanks for suggestion!

libs/utils/address.go
```diff
@@ -30,11 +30,14 @@ func ValidateAddr(addr string) (string, error) {
 	}
 
 	if ip := net.ParseIP(addr); ip == nil {
-		_, err = net.LookupHost(addr)
+		addrs, err := net.LookupHost(addr)
 		if err != nil {
-			return addr, fmt.Errorf("could not resolve hostname or ip: %w", err)
+			return addr, fmt.Errorf("could not resolve %v, err: %w", addr, err)
```
Wondertan, line 35: optional, err is not needed, the column separates it well
```suggestion
			return addr, fmt.Errorf("could not resolve %v: %w", addr, err)
```

Wondertan:
@distractedm1nd, need you review here

Wondertan:
@walldiss, the unit test here is failing for some reason

distractedm1nd (approved):
utACK

Wondertan (approved):

PR #2341, opened 2023-06-07
fix(share): left pad nid in NewNamespaceV0 instead of right padding

### LINKED ISSUE #2320: share: NewNamespaceV0 should be left padded
Passing in `0x1111111111111111` gives `0000000000000000000000000000000000000011111111111111110000` it should be  `0000000000000000000000000000000000000000001111111111111111`

distractedm1nd:
Closes #2320


walldiss (approved):

Wondertan:
Once we are on it and there is no rush, let's add a unit test that confirms that we correctly left pad it. The example in the issue would be the best.

distractedm1nd:
Added unit tests and improved docs

Wondertan (approved):

walldiss (approved):

PR #2344, opened 2023-06-08
chore: update openrpc spec file to api version v0.2.0

distractedm1nd:
Updates openrpc spec for api version v0.2.0


Wondertan (approved):

distractedm1nd:
We should only merge this once we release

PR #2347, opened 2023-06-08
feat!(nodbuilder): making lifecycle timeouts configurable

### LINKED ISSUE #2315: nodebuilder: make startup timeout confgurable

distractedm1nd:
Closes #2315 

Doing this proactively even though it is a good-first issue in case other dagstore startup issues are not fixed in time. Breaks the config.

### DIFF
--- a/nodebuilder/node/config.go
+++ b/nodebuilder/node/config.go
@@ -5,7 +5,7 @@ import (
 	"time"
 )
 
-var DefaultLifecycleTimeout = time.Minute * 2
+var defaultLifecycleTimeout = time.Minute * 2
 
 type Config struct {
 	StartupTimeout  time.Duration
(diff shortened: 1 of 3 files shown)


Wondertan (approved):

walldiss (dismissed):

nodebuilder/node/config.go
```diff
@@ -0,0 +1,38 @@
+package node
+
+import (
+	"fmt"
+	"time"
+)
+
+var DefaultLifecycleTimeout = time.Minute * 2
```
renaynay, line 8: why do you need to define/export this?

distractedm1nd: good catch, it doesnt have to be anymore

walldiss (approved):

renaynay (approved):

Wondertan (approved):

Issue #2337, opened 2023-06-07
API testing


Wondertan:
Currently, we test high-level APIs via Swamp through Go APIs. However:
* We don't test APIs of all the modules, like State
* We don't test APIs over RPC(only BlobModule is tested in atm)
	* So we see issues like #2322
	
We should investigate which modules lack API coverage and ensure we test them directly over Go interfaces and over RPC.

Wondertan:
Related https://github.com/celestiaorg/celestia-node/issues/2240 as we need a way to nicely abstract nodes running over RPC and over direct Go

PR #2356, opened 2023-06-12
refactor(swamp): use RPC instead of service pointers

distractedm1nd:
~~This PR refactors the existing swamp tests to use the RPC. Still in draft because it broke a single test (`TestRestartNodeDiscovery`) and I cannot figure out why.~~

Based on #2361 
First step of #2337

### DIFF
--- a/nodebuilder/node.go
+++ b/nodebuilder/node.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/cristalhq/jwt"
 	"github.com/ipfs/go-blockservice"
 	exchange "github.com/ipfs/go-ipfs-exchange-interface"
 	logging "github.com/ipfs/go-log/v2"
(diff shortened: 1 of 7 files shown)


api/rpc/client/client.go
```diff
@@ -90,3 +84,17 @@ func newClient(ctx context.Context, addr string, authHeader http.Header) (*Clien
...
+		"share":  &client.Share.Internal,
+		"state":  &client.State.Internal,
+		"header": &client.Header.Internal,
+		"fraud":  &client.Fraud.Internal,
+		"das":    &client.DAS.Internal,
+		"p2p":    &client.P2P.Internal,
+		"node":   &client.Node.Internal,
+		"blob":   &client.Blob.Internal,
+	}
+}
```
Wondertan, line 100: Isn't this part of a different PR?

Wondertan:
I think we should merge cleanussy first

api/rpc/client/client.go
distractedm1nd: Yes, just forgot to rebase

distractedm1nd:
Will redo/rebase once cleanussy is fully merged

Issue #825, opened 2022-06-14
tests(share): enable race detector for reconstruction tests once we migrate to go1.19 


Wondertan:
The tests were introduced in https://github.com/celestiaorg/celestia-node/pull/702

Wondertan:
As https://github.com/celestiaorg/celestia-node/pull/1067 landed, we can now enable a race detector for those files.

ramin:
integration tests have been reworked with race enabled/disabled where flakeyness allows

PR #2358, opened 2023-06-12
fix(swamp): use synchronized map as container for nodes

Wondertan:
While fixing https://github.com/celestiaorg/celestia-node/issues/825, I discovered a race. We were not synchronizing writes to containers on the node. Additionally, I simplified storing of registered nodes by using map instead of per type slices. We don't use this per type separation anyhow

### DIFF
--- a/nodebuilder/tests/swamp/swamp.go
+++ b/nodebuilder/tests/swamp/swamp.go
@@ -48,7 +48,7 @@ const DefaultTestTimeout = time.Minute * 5
 // - trustedHash taken from the CoreClient and shared between nodes
 type Swamp struct {
 	t   *testing.T
-	cgf *Config
+	cfg *Config
 
 	Network       mocknet.Mocknet
 	ClientContext testnode.Context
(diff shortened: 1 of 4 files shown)


distractedm1nd (dismissed):
utACK, looks good and makes sense

nodebuilder/tests/swamp/swamp.go
```diff
@@ -45,16 +47,16 @@ const DefaultTestTimeout = time.Minute * 5
...
 // - trustedHash taken from the CoreClient and shared between nodes
 type Swamp struct {
-	t           *testing.T
-	Network     mocknet.Mocknet
-	BridgeNodes []*nodebuilder.Node
-	FullNodes   []*nodebuilder.Node
-	LightNodes  []*nodebuilder.Node
-	comps       *Config
+	t   *testing.T
+	cgf *Config
```
walldiss, line 51: typo

distractedm1nd (approved):

walldiss (approved):

PR #2359, opened 2023-06-12
deps: bump libp2p

### LINKED ISSUE #2211: deps: Update libp2p to v0.28
We already tried to update #2127 and then had to revert https://github.com/celestiaorg/celestia-node/pull/2136. The reason is hidden in https://github.com/libp2p/go-libp2p/pull/1574, where responsibility for sending events is now moved to the Network/Swarm from Host. However, mocknet was not updated to send notifications, causing our tests to fail.

I submitted the fix https://github.com/libp2p/go-libp2p/pull/2287 and waiting for the reviews there. Once it lands, we should be able to upgrade. Alternatively, https://github.com/libp2p/go-libp2p/issues/2293 might be introduced, which will likely require some migration

Wondertan:
Closes #2211

This release brings a whole set of important fixes.

I wish we could also update DHT, but [we can't](https://github.com/libp2p/go-libp2p-kad-dht/issues/846) atm :(


Wondertan:
The Discovery test is failing for some reason, going to investigate.

distractedm1nd (dismissed):

renaynay (approved):

distractedm1nd (approved):

PR #2360, opened 2023-06-13
chore!(gateway): add deprecation notices

distractedm1nd:
- Adds a warn log on startup if --gateway is passed
- Deprecated endpoints are now disabled by default. This includes:
    - /balance
    - /submit_pfd
    - /query_delegation
    - /query_unbonding
    - /query_redelegations
- Deprecated endpoints can be activated by passing `--gateway.deprecated-endpoints`
- When using deprecated endpoints, the node will log a warning, giving the alternative RPC method to call in the future. 

The ugly config change/flag will be removed when fully deprecating the endpoints

### DIFF
--- a/api/gateway/das.go
+++ b/api/gateway/das.go
@@ -10,6 +10,7 @@ const (
 )
 
 func (h *Handler) handleDASStateRequest(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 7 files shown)


Wondertan (commented):
Lovely way of deprecating things! J
ust one suggestion

PR #2361, opened 2023-06-13
fix(rpc/client): stop overwriting existing clients

distractedm1nd:
While working on #2356 I uncovered a bug:

When using the `client.NewClient` constructor, it uses a global static struct to set the modules. This led to clients overwriting each other when multiple are initialized.

### DIFF
--- a/api/rpc/client/client.go
+++ b/api/rpc/client/client.go
@@ -18,18 +18,11 @@ import (
 	"github.com/celestiaorg/celestia-node/nodebuilder/state"
 )
 
-// TODO: this duplication of strings many times across the codebase can be avoided with issue #1176
-var client Client
-var Modules = map[string]interface{}{
-	"share":  &client.Share.Internal,
-	"state":  &client.State.Internal,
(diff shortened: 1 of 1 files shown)


Wondertan (approved):

api/rpc/client/client.go
```diff
@@ -18,18 +18,11 @@ import (
...
-	"fraud":  &client.Fraud.Internal,
-	"das":    &client.DAS.Internal,
-	"p2p":    &client.P2P.Internal,
-	"node":   &client.Node.Internal,
-	"blob":   &client.Blob.Internal,
-}
+var (
+	// staticClient is used for generating the OpenRPC spec.
+	staticClient Client
+	Modules      = moduleMap(&staticClient)
```
renaynay, line 24: nit: exported should be listed first, white space between

renaynay (approved):

api/rpc/client/client.go
distractedm1nd: not blocking so merging

PR #2367, opened 2023-06-14
share: introduce Namespace type

Wondertan:
Mostly copied from the app's Namespace type with multiple modifications. In preparation for https://github.com/celestiaorg/celestia-node/issues/2301.

We want a more straightforward Namespace type from what the app provides:
* Repsents 1:1 mapping of the serialized form of Namespace [version byte + namespace id]
* So that it does not allocate on the hot path when getting data by namespace
* and had simpler json serialization

### DIFF
--- a/share/namespace.go
+++ b/share/namespace.go
@@ -39,16 +39,16 @@ func (n Namespace) ID() namespace.ID {
 	return namespace.ID(n[appns.NamespaceVersionSize:])
 }
 
-// AsNMT converts the whole Namespace(both Version and ID parts) into NMT's namespace.ID
+// ToNMT converts the whole Namespace(both Version and ID parts) into NMT's namespace.ID
 // NOTE: Once https://github.com/celestiaorg/nmt/issues/206 is closed Namespace should become NNT's
 // type.
-func (n Namespace) AsNMT() namespace.ID {
+func (n Namespace) ToNMT() namespace.ID {
 	return namespace.ID(n)
 }
 
-// AsAppNamespace converts the Namespace to App's definition of Namespace.
+// ToAppNamespace converts the Namespace to App's definition of Namespace.
 // TODO: Unify types between node and app
-func (n Namespace) AsAppNamespace() appns.Namespace {
+func (n Namespace) ToAppNamespace() appns.Namespace {
 	return appns.Namespace{Version: n.Version(), ID: n.ID()}
 }
 
@@ -90,7 +90,7 @@ func (n Namespace) ValidateDataNamespace() error {
 		return err
 	}
 	if n.Equals(ParitySharesNamespace) || n.Equals(TailPaddingNamespace) {
-		return fmt.Errorf("invalid data namespace(%s): parity and tail padding namespace are fobidden", n)
+		return fmt.Errorf("invalid data namespace(%s): parity and tail padding namespace are forbidden", n)
 	}
 	return nil
 }
--- a/share/namespace.go
+++ b/share/namespace.go
@@ -1,0 +1,134 @@
+package share
+
+import (
+	"bytes"
+	"encoding/hex"
+	"fmt"
+
+	appns "github.com/celestiaorg/celestia-app/pkg/namespace"
+	"github.com/celestiaorg/nmt/namespace"
+)
+
+// Various reserved namespaces.
+var (
+	MaxReservedNamespace     = appns.MaxReservedNamespace.Bytes()
+	ParitySharesNamespace    = appns.ParitySharesNamespace.Bytes()
+	TailPaddingNamespace     = appns.TailPaddingNamespace.Bytes()
+	ReservedPaddingNamespace = appns.ReservedPaddingNamespace.Bytes()
+	TxNamespace              = appns.TxNamespace.Bytes()
+	PayForBlobNamespace      = appns.PayForBlobNamespace.Bytes()
+)
+
+// Namespace represents namespace of a Share.
+// Consists of version byte and namespace ID.
+type Namespace []byte
+
+// NamespaceFromBytes converts bytes into Namespace and validates it.
+func NamespaceFromBytes(b []byte) (Namespace, error) {
+	n := Namespace(b)
+	return n, n.Validate()
+}
+
+// Version reports version of the Namespace.
+func (n Namespace) Version() byte {
+	return n[appns.NamespaceVersionSize-1]
+}
+
+// ID reports version of the Namespace.
+func (n Namespace) ID() namespace.ID {
+	return namespace.ID(n[appns.NamespaceVersionSize:])
+}
+
+// String stringifies the Namespace.
+func (n Namespace) String() string {
+	return hex.EncodeToString(n)
+}
+
+// Validate checks if the namespace is correct.
+func (n Namespace) Validate() error {
+	if len(n) != NamespaceSize {
+		return fmt.Errorf("invalid namespace length: %v must be %v", len(n), NamespaceSize)
+	}
+	if n.Version() != appns.NamespaceVersionZero && n.Version() != appns.NamespaceVersionMax {
+		return fmt.Errorf("unsupported namespace version %v", n.Version())
+	}
+	if len(n.ID()) != appns.NamespaceIDSize {
+		return fmt.Errorf("unsupported namespace id length: id %v must be %v bytes but it was %v bytes",
+			n.ID(), appns.NamespaceIDSize, len(n.ID()))
+	}
+	if n.Version() == appns.NamespaceVersionZero && !bytes.HasPrefix(n.ID(), appns.NamespaceVersionZeroPrefix) {
+		return fmt.Errorf("unsupported namespace id with version %v. GetNamespace %v must start with %v leading zeros",
+			n.Version(), n.ID(), len(appns.NamespaceVersionZeroPrefix))
+	}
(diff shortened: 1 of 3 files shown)


distractedm1nd (approved):
Looks good

share/namespace.go
```diff
@@ -0,0 +1,134 @@
...
+// NamespaceFromBytes converts bytes into Namespace and validates it.
+func NamespaceFromBytes(b []byte) (Namespace, error) {
+	n := Namespace(b)
+	return n, n.Validate()
+}
+
+// Version reports version of the Namespace.
+func (n Namespace) Version() byte {
+	return n[appns.NamespaceVersionSize-1]
+}
```
walldiss, line 35: This API returns one byte, so it won't support any values of `appns.NamespaceVersionSize` other then `1`. So maybe just return n[0] instead? Other options could be to return []byte or add type for version (my personal preference).

share/namespace.go
```diff
@@ -0,0 +1,134 @@
...
+
+// ID reports version of the Namespace.
```
renaynay, line 37:
```suggestion
// ID reports the NamespaceID.
```

share/namespace.go
```diff
@@ -0,0 +1,134 @@
...
+func (n Namespace) String() string {
+	return hex.EncodeToString(n)
+}
+
+// Validate checks if the namespace is correct.
+func (n Namespace) Validate() error {
+	if len(n) != NamespaceSize {
+		return fmt.Errorf("invalid namespace length: %v must be %v", len(n), NamespaceSize)
+	}
+	if n.Version() != appns.NamespaceVersionZero && n.Version() != appns.NamespaceVersionMax {
```
renaynay, line 52: so node will accept either namespace version 0 or the max version?

share/namespace.go
```diff
@@ -0,0 +1,134 @@
...
+		return fmt.Errorf("unsupported namespace version %v", n.Version())
+	}
+	if len(n.ID()) != appns.NamespaceIDSize {
+		return fmt.Errorf("unsupported namespace id length: id %v must be %v bytes but it was %v bytes",
```
renaynay, line 56: nit: let's make a uniform print for failures like this

expected %d, got %d

and not reverse the order

share/namespace.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	if n.Version() == appns.NamespaceVersionZero && !bytes.HasPrefix(n.ID(), appns.NamespaceVersionZeroPrefix) {
+		return fmt.Errorf("unsupported namespace id with version %v. GetNamespace %v must start with %v leading zeros",
+			n.Version(), n.ID(), len(appns.NamespaceVersionZeroPrefix))
+	}
+	return nil
+}
+
+// ValidateBlobNamespace returns an error if this namespace is not a valid blob namespace.
+func (n Namespace) ValidateBlobNamespace() error {
```
renaynay, line 67: also check for Tx Namespace and PFB namespace?

rootulp: Blob namespaces are the namespaces that can be specified by users when publishing a blob. The Tx namespace and PFB namespace are reserved namespaces so they can't be specified by users when publishing a blob.

If the tx namespace or PFB namespace is specified, it would be caught by https://github.com/celestiaorg/celestia-node/pull/2367/files/83977c48c94c4197dda01fdadefefc7e56660ed9#diff-46c85ee9c611a9b9c56eb104a6dca5bcbc6b3961683cdf0c941704b18496084dR71-R74

share/namespace.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	if n.Version() != appns.NamespaceVersionZero && n.Version() != appns.NamespaceVersionMax {
+		return fmt.Errorf("unsupported namespace version %v", n.Version())
+	}
+	if len(n.ID()) != appns.NamespaceIDSize {
+		return fmt.Errorf("unsupported namespace id length: id %v must be %v bytes but it was %v bytes",
+			n.ID(), appns.NamespaceIDSize, len(n.ID()))
+	}
+	if n.Version() == appns.NamespaceVersionZero && !bytes.HasPrefix(n.ID(), appns.NamespaceVersionZeroPrefix) {
+		return fmt.Errorf("unsupported namespace id with version %v. GetNamespace %v must start with %v leading zeros",
```
rootulp, line 60: [nit] this error message may be misleading if user's specify the namespace as a hexadecimal encoded value because one byte is two leading zeros.

rootulp (approved):

share/namespace.go
Wondertan: Ah, right, as reserved means PFB and TX

Wondertan: Ok, so I will do additional Validate like func that also restrict access for padding/erasured data

share/namespace.go
Wondertan: I like the idea and see your point. The reason I like the approach with the constant is that if it changes, the unit tests will fail and force us to do something.

share/namespace.go
```diff
@@ -0,0 +1,134 @@
...
+		return fmt.Errorf("invalid blob namespace: %v cannot use parity shares namespace", n)
+	}
+	if n.IsTailPadding() {
+		return fmt.Errorf("invalid blob namespace: %v cannot use tail padding namespace", n)
+	}
+	return nil
+}
+
+func (n Namespace) IsReserved() bool {
+	return bytes.Compare(n, MaxReservedNamespace) < 1
```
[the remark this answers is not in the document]
Wondertan: It works over hashes, not a namespace.

share/namespace.go
Wondertan: Fixed

share/namespace.go
Wondertan: Added

share/namespace.go
Wondertan: simplified error

share/namespace.go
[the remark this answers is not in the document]
Wondertan: Also, I moved these helpers as methods. Much much nicer now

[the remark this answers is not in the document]
walldiss: i was referring to namespace methods:
```
func (n Namespace) IsLessThan(n2 Namespace)
```

Also you would need to make MaxReservedNamespace, ParitySharesNamespace, etc a namespace type. I think it would be nicer for them to be Namespace type too, so user will be able to use it like:
```
mynaspace. Equals(ParitySharesNamespace)
```
And there will be no need for explicit method for every namespace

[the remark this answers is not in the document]
walldiss: Yes, but it is a type of `[]byte` not `Namespace`. I suggest to make it latter, so it is possible to use all new introduced tooling associated with type.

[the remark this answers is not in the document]
Wondertan: > func (n Namespace) IsLessThan(n2 Namespace)

I tried this in the integration and didn't really like it. It's much cleaner on the user side to take the whole hash. You will see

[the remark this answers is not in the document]
Wondertan: > And there will be no need for explicit method for every namespace

I like this idea though.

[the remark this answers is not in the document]
Wondertan: > Yes, but it is a type of []byte not Namespace. I suggest to make it latter, so it is possible to use all new introduced tooling associated with type.

Golang takes care of this automatically. You don't need to define. I think this is the case since generic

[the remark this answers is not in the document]
Wondertan: I am using Equals now on these vars that are []byte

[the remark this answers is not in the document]
Wondertan: I am not sure we need this, but It does not hurt to have it.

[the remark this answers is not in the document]
Wondertan: I think we should do it just to be consistent with types

walldiss (dismissed):
Beautiful

share/namespace.go
```diff
@@ -0,0 +1,151 @@
...
+
+// ID reports ID of the Namespace.
+func (n Namespace) ID() namespace.ID {
+	return namespace.ID(n[appns.NamespaceVersionSize:])
+}
+
+// AsNMT converts the whole Namespace(both Version and ID parts) into NMT's namespace.ID
+// NOTE: Once https://github.com/celestiaorg/nmt/issues/206 is closed Namespace should become NNT's
+// type.
+func (n Namespace) AsNMT() namespace.ID {
```
renaynay, line 45: nit but can we do `ToNMT` and `ToAppNamespace` instead of As?

share/namespace.go
```diff
@@ -0,0 +1,151 @@
...
+	return nil
+}
+
+// ValidateDataNamespace checks if the Namespace contains real/useful data.
+func (n Namespace) ValidateDataNamespace() error {
+	if err := n.Validate(); err != nil {
+		return err
+	}
+	if n.Equals(ParitySharesNamespace) || n.Equals(TailPaddingNamespace) {
+		return fmt.Errorf("invalid data namespace(%s): parity and tail padding namespace are fobidden", n)
```
renaynay, line 93:
```suggestion
		return fmt.Errorf("invalid data namespace(%s): parity and tail padding namespace are forbidden", n)
```

rootulp (approved):

distractedm1nd (approved):

walldiss (approved):
Bless

PR #2373, opened 2023-06-16
feat(gateway): enabling CORS * on gateway

distractedm1nd:
Closes #2263


Wondertan:
What about RPC? Did we check if it allows all the origins? The link we had was just for WS

renaynay (approved):

distractedm1nd:
@Wondertan I have not checked http over rpc, but the default connection for RPC should be ws://localhost anyways - since its the only transport that supports all of our endpoints. Adding in CORS for the (default) http client is pretty trivial but adds a lot of code (we are then no longer using the default http client so have to copy the rest of the default configuration to our client code)

Wondertan:
My worry here comes from @smuu stating that RPC still didn't work because of CORS. Maybe because they used http and not ws?

Wondertan (approved):

PR #2376, opened 2023-06-16
refactor!(share): integrate new Namespace type

### LINKED ISSUE #2301: Rename `nID` to `namespace` where applicable

Wondertan:
TODO: 
- [ ] Document changes on the PR for reviewers and for commit message
- [x] Use of ValidateDataNamespace 
- [ ] Document in issues further actionables

Based on #2367 
Closes #2301
Blocked on #2256 (because new rules on namespaces in existing tests cause to trigger not found within namespace range path logic, which we don't handle yet)


api/gateway/share.go
```diff
@@ -20,7 +19,7 @@ const (
 	namespacedDataEndpoint   = "/namespaced_data"
 )
 
-var nIDKey = "nid"
```
Wondertan, line 23: Forgot to change the actual string and it's breaking....

Don't have processing power rn to assess whether this worth a breakage here or not

Wondertan: We are not deprecating this, so it might be worth it. Does it need to changes somewhere outside of go-cnc?

Wondertan:
Currently contains all the changes in https://github.com/celestiaorg/celestia-node/pull/2256 to be sure all the tests are passing with it.

Issue #2377, opened 2023-06-16
Update BEFP creation and validation to accomodate more types of validator fraud


evan-forbes:
Per this [comment](https://github.com/celestiaorg/nmt/issues/97#issuecomment-1430327821) in nmt defining the attack scenario:
> Malicious consensus nodes could produce and commit a block, denoted as B, in which the shares are not ordered lexicographically based on their namespace IDs. A DA (data availability) full node obtains the block header, including the data root denoted as dataRoot, and the underlying shares, and will attempt to construct the block. However, if the node detects that the shares used to construct dataRoot are unordered, it must provide a proof of bad encoding to the light clients. Specifically, it must indicate that there are out-of-order leaves in the tree represented by dataRoot.

per the [LL white paper](https://arxiv.org/pdf/1905.09274.pdf) section 5.2
> An adversarial consensus node may attempt to produce a block that contains a Merkle tree with children
that are not ordered correctly. To prevent this, we can set a condition in nsHash such that there is no valid hash when leftMaxNs ≥ rightMinNs, and thus there would be no valid Merkle root for incorrectly ordered children. Therefore blockValid(hi) would return false in the simplistic and probabilistic validity rules as there is no possible Mi where root(Mi) = mRooti

addionally, 2/3s of the voting power could commit to shares that not sized properly and would result in an inability to decode the row or col.

We need to have a fraud proof that can handle out of order namespaces and arbitary things that would cause the erasure decoding to fail. BEFPs (perhaps need to be renamed after this), can and should handle these types of fraud.

During the reconstruction process, if there are two shares that are out of order, then we should still be able to create an inclusion proof to half of the shares in that row or col using _either_ the row or column roots. If the square is available, then this will always be the case.

For the second new type of fraud, if there is something wrong with the shares that would cause the row or col to not be decodable, then if the shares are committed to (step 1 below), then that means that the BEFP is valid.

To accomodate this, the verification logic must be altered to encode the following logic:

1. Check inclusion: if inclusion fails, proof is invalid
2. Reconstruction of row of col: if fails, **proof is valid**
4. Create merkel root: if creation of root fails, **proof is valid**
5. Compare existing root: if different, the proof is valid

currently, we are not following steps two and three of the validation process:
https://github.com/celestiaorg/celestia-node/blob/6654cdf4994dbd381efd0d6a29688c731177c855/share/eds/byzantine/bad_encoding.go#L162-L184

[...]

PR #2383, opened 2023-06-21
feat(share/byzantine): handle additional cases in befp validation

vgonkivs:
## Overview
Extend a BEFP validation.
Related to https://github.com/celestiaorg/celestia-node/issues/2377

## Checklist

### DIFF
--- a/share/eds/byzantine/bad_encoding.go
+++ b/share/eds/byzantine/bad_encoding.go
@@ -185,8 +185,8 @@ func (p *BadEncodingProof) Validate(hdr libhead.Header) error {
 	odsWidth := uint64(len(merkleRoots) / 2)
 	codec := share.DefaultRSMT2DCodec()
 
-	// We assume that the proof is valid in case if we proved the inclusion of `Shares` but
-	// a row/col can't be reconstructed or building a NMT tree fails.
+	// We assume that the proof is valid in case we proved the inclusion of `Shares` but
+	// the row/col can't be reconstructed, or the building of NMTree fails.
 
 	// rebuild a row or col.
 	rebuiltShares, err := codec.Decode(shares)
--- a/share/eds/byzantine/bad_encoding.go
+++ b/share/eds/byzantine/bad_encoding.go
@@ -2,6 +2,7 @@ package byzantine
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 
 	"github.com/celestiaorg/celestia-app/pkg/wrapper"
@@ -152,6 +153,20 @@ func (p *BadEncodingProof) Validate(hdr libhead.Header) error {
 		)
 	}
 
+	// find at least 1 non-empty share
+	nonEmpty := false
+	for _, share := range p.Shares {
+		if share == nil {
+			continue
+		}
+		nonEmpty = true
+		break
+	}
+
+	if !nonEmpty {
+		return errors.New("fraud: invalid proof: no shares provided to reconstruct row/col")
+	}
+
 	// verify that Merkle proofs correspond to particular shares.
 	shares := make([][]byte, len(merkleRoots))
 	for index, shr := range p.Shares {
@@ -170,14 +185,20 @@ func (p *BadEncodingProof) Validate(hdr libhead.Header) error {
 	odsWidth := uint64(len(merkleRoots) / 2)
 	codec := share.DefaultRSMT2DCodec()
 
+	// We assume that the proof is valid in case if we proved the inclusion of `Shares` but
+	// a row/col can't be reconstructed or building a NMT tree fails.
+
 	// rebuild a row or col.
 	rebuiltShares, err := codec.Decode(shares)
 	if err != nil {
-		return err
+		log.Errorw("failed to decode shares", "err", err)
+		return nil
 	}
+
 	rebuiltExtendedShares, err := codec.Encode(rebuiltShares[0:odsWidth])
 	if err != nil {
-		return err
+		log.Errorw("failed to encode shares", "err", err)
+		return nil
 	}
 	copy(rebuiltShares[odsWidth:], rebuiltExtendedShares)
 
@@ -185,13 +206,15 @@ func (p *BadEncodingProof) Validate(hdr libhead.Header) error {
 	for _, share := range rebuiltShares {
 		err = tree.Push(share)
 		if err != nil {
-			return err
+			log.Errorw("failed to build a tree from the reconstructed shares", "err", err)
+			return nil
 		}
 	}
 
 	expectedRoot, err := tree.Root()
 	if err != nil {
-		return err
+		log.Errorw("failed to build a tree root", "err", err)
+		return nil
 	}
 
 	// root is a merkle root of the row/col where ErrByzantine occurred
--- a/share/eds/byzantine/bad_encoding.go
+++ b/share/eds/byzantine/bad_encoding.go
@@ -191,14 +191,18 @@ func (p *BadEncodingProof) Validate(hdr libhead.Header) error {
 	// the row/col can't be reconstructed, or the building of NMTree fails.
 	rebuiltShares, err := codec.Decode(shares)
 	if err != nil {
-		log.Infow("failed to decode shares", "err", err)
+		log.Infow("failed to decode shares at height",
+			"height", header.Height(), "err", err,
+		)
 		return nil
 	}
 
 	odsWidth := uint64(len(merkleRoots) / 2)
 	rebuiltExtendedShares, err := codec.Encode(rebuiltShares[0:odsWidth])
 	if err != nil {
-		log.Infow("failed to encode shares", "err", err)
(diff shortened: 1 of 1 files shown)


share/eds/byzantine/bad_encoding.go
```diff
@@ -141,6 +141,20 @@ func (p *BadEncodingProof) Validate(hdr libhead.Header) error {
...
+			continue
+		}
+		nonEmpty = true
+		break
+	}
+
+	if !nonEmpty {
+		return errors.New("fraud: invalid proof: no shares provided to reconstruct row/col")
+	}
+
```
Wondertan, line 157: We can simply check len of `shares` right after the loop below for the same effect.

share/eds/byzantine/bad_encoding.go
```diff
@@ -159,34 +173,41 @@ func (p *BadEncodingProof) Validate(hdr libhead.Header) error {
 	odsWidth := uint64(len(merkleRowRoots) / 2)
 	codec := share.DefaultRSMT2DCodec()
 
+	// We assume that the proof is valid in case if we proved the inclusion of `Shares` but
+	// a row/col can't be reconstructed or building a NMT tree fails.
```
Wondertan, line 177:
```suggestion
	// We assume that the proof is valid in case we proved the inclusion of `Shares` but
	// the row/col can't be reconstructed, or the building of NMTree fails.
```

share/eds/byzantine/bad_encoding.go
vgonkivs: not exactly, because we can receive a slice with all nil values inside.

vgonkivs: oh, I got you. Thanks

Wondertan: You can preallocate by cap then. Then it's much less code

renaynay: let's at least make it not double negation, so the check should be `empty := true` and then check if `empty`

share/eds/byzantine/bad_encoding.go
```diff
@@ -159,34 +173,41 @@ func (p *BadEncodingProof) Validate(hdr libhead.Header) error {
...
 
+	// We assume that the proof is valid in case we proved the inclusion of `Shares` but
+	// the row/col can't be reconstructed, or the building of NMTree fails.
+
 	// rebuild a row or col.
 	rebuiltShares, err := codec.Decode(shares)
 	if err != nil {
-		return err
+		log.Errorw("failed to decode shares", "err", err)
+		return nil
```
renaynay, line 183: wait why are we returning nil now?

Wondertan: That's the whole point of the PR

Wondertan (commented):
Actually, I just realized that this is only the verification part. We have to triple-check that the generation works for these cases.

vgonkivs:
These cases are possible with out-of-order shares. But we can't generate them for now

vgonkivs:
Moving this PR to draft until we can test these conditions

share/eds/byzantine/bad_encoding.go
```diff
@@ -159,34 +173,41 @@ func (p *BadEncodingProof) Validate(hdr libhead.Header) error {
...
 	}
+
 	rebuiltExtendedShares, err := codec.Encode(rebuiltShares[0:odsWidth])
 	if err != nil {
-		return err
+		log.Errorw("failed to encode shares", "err", err)
```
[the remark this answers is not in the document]
vgonkivs: yeap, good idea. Thanks

Wondertan (dismissed):

share/eds/byzantine/bad_encoding.go
```diff
@@ -167,31 +185,37 @@ func (p *BadEncodingProof) Validate(hdr libhead.Header) error {
...
-	odsWidth := uint64(len(merkleRoots) / 2)
 	codec := share.DefaultRSMT2DCodec()
 
-	// rebuild a row or col.
+	// We assume that the proof is valid in case we proved the inclusion of `Shares` but
+	// the row/col can't be reconstructed, or the building of NMTree fails.
 	rebuiltShares, err := codec.Decode(shares)
 	if err != nil {
-		return err
+		log.Infow("failed to decode shares", "err", err)
```
renaynay, line 194: can you add some additional info here - like height / root it was for, same for other logs

renaynay (approved):

Wondertan (dismissed):

share/eds/byzantine/bad_encoding.go
```diff
@@ -114,6 +115,7 @@ func (p *BadEncodingProof) Validate(hdr libhead.Header) error {
 	if !ok {
 		panic(fmt.Sprintf("invalid header type received during BEFP validation: expected %T, got %T", header, hdr))
 	}
+
```
adlerjohn, line 118: Aside and non-blocking, but why is fmt adding newlines here all of a sudden?

share/eds/byzantine/bad_encoding.go
```diff
@@ -152,6 +156,20 @@ func (p *BadEncodingProof) Validate(hdr libhead.Header) error {
...
+	for _, share := range p.Shares {
+		if share == nil {
+			continue
+		}
+		empty = false
+		break
+	}
+
+	if empty {
+		return errors.New("fraud: invalid proof: no shares provided to reconstruct row/col")
```
adlerjohn, line 170: Hmmm shouldn't you check that at least half the shares are present, instead of at least one?

share/eds/byzantine/bad_encoding.go
```diff
@@ -167,31 +185,45 @@ func (p *BadEncodingProof) Validate(hdr libhead.Header) error {
 		shares[index] = share.GetData(shr.Share)
 	}
 
-	odsWidth := uint64(len(merkleRoots) / 2)
 	codec := share.DefaultRSMT2DCodec()
 
-	// rebuild a row or col.
+	// We assume that the proof is valid in case we proved the inclusion of `Shares` but
+	// the row/col can't be reconstructed, or the building of NMTree fails.
```
adlerjohn, line 191: Non-blocking, can improve in a follow-up issue. Instead of "assume," IMO it's better to say we can conclude. I.e. it's the end instead of the beginning of the process.

share/eds/byzantine/bad_encoding.go
```diff
@@ -167,31 +185,45 @@ func (p *BadEncodingProof) Validate(hdr libhead.Header) error {
...
 	rebuiltShares, err := codec.Decode(shares)
 	if err != nil {
-		return err
+		log.Infow("failed to decode shares at height",
```
adlerjohn, line 194: Non-blocking, can improve in a follow-up issue. Can this and the below info logs clarify that the failure is a `good thing`? Otherwise a user might think something went wrong.

adlerjohn (commented):
only one non-non-blocking comment

share/eds/byzantine/bad_encoding.go
vgonkivs: That's why we make an info log level(and not an error). Or do you think it's not enough?

share/eds/byzantine/bad_encoding.go
adlerjohn: Fixed in [0aafbec](https://github.com/celestiaorg/celestia-node/pull/2383/commits/0aafbec72e7b7bfa2ea84d6375e936aeb1079e63)

share/eds/byzantine/bad_encoding.go
adlerjohn: Fixed in [0aafbec](https://github.com/celestiaorg/celestia-node/pull/2383/commits/0aafbec72e7b7bfa2ea84d6375e936aeb1079e63)

share/eds/byzantine/bad_encoding.go
adlerjohn: Ah or does gofmt not care about single newlines. Nvm

share/eds/byzantine/bad_encoding.go
adlerjohn: You're correct that fail with info instead of warn or error _should_ be sufficient, but IMO it's not obvious enough to users. Making it explicit that this failure implies the fraud proof is valid is clearer. But as above, you don't need to fix this in this PR if you're not convinced, you can file a `good first issue`.

Wondertan (approved):

share/eds/byzantine/bad_encoding.go
vgonkivs: done

share/eds/byzantine/bad_encoding.go
vgonkivs: Thanks for the suggestion. I will make an issue

renaynay (approved):

PR #2393, opened 2023-06-23
fix(api/gateway): Handle err from tx properly to avoid panic

renaynay:
Fixes a panic that can occur if txResp is nil and also returns actual error in the case of nil txResp.

Reported by @tuxcanfly


vgonkivs (approved):

Wondertan (approved):
Good bad request usage. 
LGTM, Thank you!

PR #2394, opened 2023-06-23
fix(libs/utils): Use valid ip4 address

renaynay:
Fixes an issue where `localhost` was not accepted as a valid `--gateway.addr`. Uses net.ResolveIPAddr instead.

Found by @tuxcanfly

### DIFF
--- a/libs/utils/address.go
+++ b/libs/utils/address.go
@@ -34,14 +34,7 @@ func ValidateAddr(addr string) (string, error) {
 		if err != nil {
 			return addr, err
 		}
-		addrs, err := net.LookupHost(resolved.String())
-		if err != nil {
-			return addr, fmt.Errorf("could not resolve %v: %w", addr, err)
-		}
-		if len(addrs) == 0 {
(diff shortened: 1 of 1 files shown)


Wondertan:
We have tests for the func and should cover this case

walldiss:
Why do you think the address is invalid?

renaynay:
@walldiss first addr returned for localhost is ipv6 shorthand loopback addr and it's not compatible with validation

walldiss:
To filter out ipv6 you can use 
```
net.ResolveIPAddr("ip4",addr)
```

Wondertan (dismissed):

libs/utils/address.go
```diff
@@ -30,14 +30,11 @@ func ValidateAddr(addr string) (string, error) {
 	}
 
 	if ip := net.ParseIP(addr); ip == nil {
```
walldiss, line 32: nice to have would be to return fast in case ip != nil. It would also help to return fast resolved addr

renaynay: ok nice suggestion

walldiss (approved):

PR #2408, opened 2023-06-28
share/befp: rework collecting and verifying the BEFP

vgonkivs:
## Overview
* Use an orthogonal axis to collect and verify BEFP instead of collecting and verifying it against a single Merkle Root(where ErrByzantine occurred)
* collect only 1/2 of the shares instead of all available shares from the row/col

## Checklist


share/eds/byzantine/bad_encoding.go
```diff
@@ -118,45 +118,42 @@ func (p *BadEncodingProof) Validate(hdr libhead.Header) error {
...
-		return fmt.Errorf("fraud: invalid proof: incorrect number of shares %d != %d", len(p.Shares), len(merkleRowRoots))
-	}
 
-	root := merkleRowRoots[p.Index]
-	if p.Axis == rsmt2d.Col {
-		root = merkleColRoots[p.Index]
+	// change the order of dah roots as we are now collecting nmt proofs against orthogonal axis
+	roots := [][][]byte{header.DAH.ColumnRoots, header.DAH.RowRoots}
+	if int(p.Index) >= len(roots[rsmt2d.Row]) {
+		return fmt.Errorf("fraud: invalid proof: index out of bounds (%d >= %d)", int(p.Index), len(roots[rsmt2d.Row]))
```
Wondertan, line 135: Let's mention that it's befp somehow in all the errrors.
Also, it's not fraud pkg anymore, I don't think we should prepend the byzantine here as well, just remove fraud

share/eds/byzantine/byzantine.go
```diff
@@ -35,25 +36,48 @@ func NewErrByzantine(
...
+
+	sharesWithProof := make([]*ShareWithProof, len(errByz.Shares))
+	sharesAmount := 0
+	errGr, ctx := errgroup.WithContext(ctx)
+
+	for index, share := range errByz.Shares {
+		if share == nil {
+			continue
+		}
+		// skip next shares if we already requested needed amount of them.
```
Wondertan, line 53:
```suggestion
		// skip further shares if we already requested half of them, which is enough to recompute the row or col
```

share/eds/byzantine/bad_encoding.go
vgonkivs: https://github.com/celestiaorg/celestia-node/pull/2408/commits/68b161403000aae979a2bfd120fab69899c47c5f

share/eds/byzantine/bad_encoding.go
```diff
@@ -113,50 +113,51 @@ func (p *BadEncodingProof) UnmarshalBinary(data []byte) error {
...
-	}
-	if len(merkleRowRoots) != len(p.Shares) {
-		return fmt.Errorf("fraud: invalid proof: incorrect number of shares %d != %d", len(p.Shares), len(merkleRowRoots))
-	}
 
-	root := merkleRowRoots[p.Index]
-	if p.Axis == rsmt2d.Col {
-		root = merkleColRoots[p.Index]
+	// change the order of dah roots as we are now collecting nmt proofs against orthogonal axis
+	roots := [][][]byte{header.DAH.ColumnRoots, header.DAH.RowRoots}
```
walldiss, line 133: Would be nice to add clarity what is the direct order(this one is reversed) and where it is defined.

share/eds/byzantine/bad_encoding.go
```diff
@@ -113,50 +113,51 @@ func (p *BadEncodingProof) UnmarshalBinary(data []byte) error {
...
+	}
+	if len(p.Shares) != len(roots[rsmt2d.Row]) {
+		return fmt.Errorf("invalid %s proof: incorrect number of shares %d != %d",
+			BadEncoding, len(p.Shares), len(roots[rsmt2d.Row]),
+		)
 	}
 
 	// verify that Merkle proofs correspond to particular shares.
-	shares := make([][]byte, len(merkleRowRoots))
+	shares := make([][]byte, len(roots[rsmt2d.Row]))
```
walldiss, line 146: Could it also be `len(roots[p.Axis])`? Looks like we need only single element from roots array everywhere. How about remove roots 2 element array?

share/eds/byzantine/byzantine.go
```diff
@@ -35,25 +36,49 @@ func NewErrByzantine(
...
+		}
+		// skip further shares if we already requested half of them, which is enough to recompute the row
+		// or col
+		if sharesAmount == len(dah.RowRoots)/2 {
+			break
+		}
+		sharesAmount++
+
+		index := index
+		errGr.Go(func() error {
```
walldiss, line 61: index is captured for go-routine, but what about `share`?

share/eds/byzantine/byzantine.go
```diff
@@ -35,25 +36,49 @@ func NewErrByzantine(
...
+			share, err := getProofsAt(
+				ctx, bGetter,
+				ipld.MustCidFromNamespacedSha256(roots[index]),
+				int(errByz.Index), len(errByz.Shares),
+			)
+			sharesWithProof[index] = share
```
walldiss, line 67: Would it be possible to make single proof for combined required shares? Could be a nice optimisation

share/eds/byzantine/share_proof.go
```diff
@@ -78,24 +78,38 @@ func GetProofsForShares(
...
+) (*ShareWithProof, error) {
+	proof := make([]cid.Cid, 0)
+	// TODO(@vgonkivs): Combine GetLeafData and GetProof in one function as the are traversing the same
+	// tree. Add options that will control what data will be fetched.
+	node, err := ipld.GetLeaf(ctx, bGetter, root, index, total)
+	if err != nil {
+		return nil, err
+	}
+
+	proof, err = ipld.GetProof(ctx, bGetter, root, proof, index, total)
```
walldiss, line 106: Unrelated to PR:
I wonder why `proof`needs to be preallocated, then passed to `GetProof` and then returned value assigned to the same var. API seems weird and looks like it could be changed to not take proof as an arg of `GetProof`. Looks like it is a recursion being exposed, that we can hide. Would be nice to clean it up while making single traversing function. Is there an issue for that?

Wondertan: There was def an issue

share/eds/byzantine/byzantine.go
vgonkivs: only in case they have the same namespace.

share/eds/byzantine/byzantine.go
```diff
@@ -35,25 +36,49 @@ func NewErrByzantine(
...
-		// TODO(@Wondertan): Find a better way to handle
-		log.Fatalw("getting proof for ErrByzantine", "err", err)
+		dah.RowRoots,
+	}[errByz.Axis]
+
+	sharesWithProof := make([]*ShareWithProof, len(errByz.Shares))
+	sharesAmount := 0
+	errGr, ctx := errgroup.WithContext(ctx)
+
+	for index, share := range errByz.Shares {
```
renaynay, line 49:
```suggestion
	sharesAmount := 0
	
	errGr, ctx := errgroup.WithContext(ctx)
	for index, share := range errByz.Shares {
```

share/eds/byzantine/byzantine.go
```diff
@@ -35,25 +36,49 @@ func NewErrByzantine(
...
+		if share == nil {
+			continue
+		}
+		// skip further shares if we already requested half of them, which is enough to recompute the row
+		// or col
+		if sharesAmount == len(dah.RowRoots)/2 {
+			break
+		}
```
renaynay, line 57: nit, check this first, then check if share == nil

share/eds/byzantine/byzantine.go
```diff
@@ -35,25 +36,49 @@ func NewErrByzantine(
...
+				ipld.MustCidFromNamespacedSha256(roots[index]),
+				int(errByz.Index), len(errByz.Shares),
+			)
+			sharesWithProof[index] = share
+			return err
+		})
 	}
 
+	if err := errGr.Wait(); err != nil {
+		if err != nil {
```
renaynay, line 73: additional err check?

share/eds/byzantine/bad_encoding.go
vgonkivs: yes

share/eds/byzantine/byzantine.go
vgonkivs: we don't need this global share inside go routine. it is overwritten by the local variable that is received from `getProofsAt`.

share/eds/byzantine/byzantine.go
vgonkivs: "Are you sure that error happened?" 😆

share/eds/byzantine/bad_encoding.go
```diff
@@ -183,10 +184,15 @@ func (p *BadEncodingProof) Validate(hdr libhead.Header) error {
 		return err
 	}
 
+	// root is a merkle root of the row/col where ErrByzantine occurred
+	root := header.DAH.RowRoots[p.Index]
+	if p.Axis == rsmt2d.Col {
+		root = header.DAH.ColumnRoots[p.Index]
+	}
```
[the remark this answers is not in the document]
vgonkivs: 🤔 not sure that it will be more readable

share/eds/byzantine/bad_encoding.go
```diff
@@ -113,50 +113,51 @@ func (p *BadEncodingProof) UnmarshalBinary(data []byte) error {
...
+		if ok := shr.Validate(ipld.MustCidFromNamespacedSha256(roots[p.Axis][index])); !ok {
+			return fmt.Errorf("invalid %s proof: incorrect share received at index %d", BadEncoding, index)
 		}
 		// NMTree commits the additional namespace while rsmt2d does not know about, so we trim it
 		// this is ugliness from NMTWrapper that we have to embrace ¯\_(ツ)_/¯
 		shares[index] = share.GetData(shr.Share)
 	}
 
-	odsWidth := uint64(len(merkleRowRoots) / 2)
+	odsWidth := uint64(len(roots[rsmt2d.Row]) / 2)
```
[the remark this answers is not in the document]
vgonkivs: removed a 3d slice of row/col root, so now it relies on a single merkleRoot variable that corresponds to the particular axis.

Wondertan (approved):
LGTM

renaynay (approved):

Wondertan (approved):

adlerjohn (approved):
looks ok.

Issue #2018, opened 2023-04-04
share/eds: Store should cache EDSes on Put


Wondertan:
One of the ideas to make it work is to create the custom dagstore Mount implementation that wraps FileMount. This Mount will also have a pointer to the serialized CAR. If the pointer is not nil, then dagstore will access the serialized CAR from memory. Otherwise, it will behave as usual FileMount. The serialized version of CAR can be retrieved via io.TeeWriter, while writing to the file on disk.

PR #2409, opened 2023-06-28
feat(share/store): store eds in memory for index builder

walldiss:
## Overview

Previously generated eds was first store to disk and then was read again while building car index for dagstore.
This PR contains 2 changes:
- writes eds into buffer to write it as 1 chunk instead of many small writes to disk to reduce disk i/o pressure. 
- introduces in-memory cache that will be used to read eds for building car index.

Combined changes brings 3x performance improvement for eds.Put operation

### DIFF
--- a/share/eds/store.go
+++ b/share/eds/store.go
@@ -182,15 +182,24 @@ func (s *Store) Put(ctx context.Context, root share.DataHash, square *rsmt2d.Ext
 	}
 	defer f.Close()
 
-	err = WriteEDS(ctx, square, f)
+	// save encoded eds into buffer
+	buf := bytes.NewBuffer(nil)
+	err = WriteEDS(ctx, square, buf)
 	if err != nil {
 		return fmt.Errorf("failed to write EDS to file: %w", err)
 	}
 
+	// write whole buffer in one go to optimise i/o
+	if _, err = f.Write(buf.Bytes()); err != nil {
+		return fmt.Errorf("failed to write EDS to file: %w", err)
+	}
+
+	mount := &inMemoryOnceMount{
+		buf:   buf.Bytes(),
+		Mount: &mount.FileMount{Path: s.basepath + blocksPath + key},
+	}
 	ch := make(chan dagstore.ShardResult, 1)
-	err = s.dgstr.RegisterShard(ctx, shard.KeyFromString(key), &mount.FileMount{
-		Path: s.basepath + blocksPath + key,
-	}, ch, dagstore.RegisterOpts{})
+	err = s.dgstr.RegisterShard(ctx, shard.KeyFromString(key), mount, ch, dagstore.RegisterOpts{})
 	if err != nil {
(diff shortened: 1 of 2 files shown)


distractedm1nd (changes requested):
Mount type must be registered on registry inside edsstore constructor

share/eds/store.go
```diff
@@ -419,3 +431,30 @@ func setupPath(basepath string) error {
...
+// inMemoryOnceMount is used to allow reading once from buffer before using main mount.Reader
+type inMemoryOnceMount struct {
+	buf []byte
+
+	readOnce atomic.Bool
+	mount.Mount
+}
+
+// inmemoryReader extends bytes.Reader to implement mount.Reader interface
+type inmemoryReader struct {
```
Wondertan, line 444: Capitalize memory like in the mount

share/eds/store.go
```diff
@@ -419,3 +431,30 @@ func setupPath(basepath string) error {
 	}
 	return nil
 }
+
+// inMemoryOnceMount is used to allow reading once from buffer before using main mount.Reader
+type inMemoryOnceMount struct {
+	buf []byte
```
Wondertan, line 437: Why not embed the bytes.Buffer? Then inMemoryReader won't be needed, nor the buffer on 189.

The code would look like:
WriteEDS(mount)

mount.WriteTo(file)

RegisterShard(mount)

Wondertan (commented):
LGTM, fixes an old #2018

share/eds/store.go
walldiss: Thank for suggested, I've made small refactoring to incorporate buffer into mount. 
inMemoryReader is needed to extend bufio.Reader to be compatible with mount.Mount interface.

share/eds/store.go
```diff
@@ -419,3 +431,39 @@ func setupPath(basepath string) error {
...
+type inMemoryOnceMount struct {
+	buf *bytes.Buffer
+
+	readOnce atomic.Bool
+	mount.Mount
+}
+
+func (m *inMemoryOnceMount) Fetch(ctx context.Context) (mount.Reader, error) {
+	if !m.readOnce.Swap(true) {
+		reader := &inMemoryReader{Reader: bytes.NewReader(m.buf.Bytes())}
```
Wondertan, line 445: Why not pass Buffer instead of Reader?

walldiss: `bytes.Reader` implements `io.ReaderAt` and `io.Seeker` interfaces that required to satisfy `mount.Reader` interface. `bytes.Buffer` doesn't implement those. I didn't want to implement those interfaces, since it could be a bit complex while unnecessary.

Wondertan (approved):
It would be nice to merge this one as it's not breaking, yet very valuable

distractedm1nd (approved):

Wondertan (approved):

PR #2414, opened 2023-06-29
feat: update Go base alpine - add branch in makefile

jrmanes:
hello team!

This PR contains a couple of things:
- Update the base version of Go Alpine to `golang:1.20-alpine3.18`
- In the Makefile, I've added the following line:

```Makefile
main.semanticVersion=$(shell git describe --tags --dirty=-dev 2>/dev/null || git rev-parse --abbrev-ref HEAD)'"
```
Basically what it does is, in case this commit doesn't contain any tag, it uses the branch name.

At the moment we are having some issues using the flag `version`, the output command is:

Screenshot 2023-06-29 at 17 24 22

With this change, we'll have something like:

Screenshot 2023-06-29 at 17 08 17

*Already tested*

Let me know what you think about it and if it's worth it 😊 

Thanks in advance! 🚀 

Jose Ramon Mañes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,4 +49,4 @@ USER ${USER_NAME}
 EXPOSE 2121
 
 ENTRYPOINT [ "/bin/bash", "/opt/entrypoint.sh" ]
-CMD [ "celestia"]
+CMD [ "celestia" ]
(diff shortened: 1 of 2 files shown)


Bidon15 (approved):

vgonkivs (approved):
🚀

renaynay (approved):

PR #2421, opened 2023-07-03
fix: use embedded files instead of ldflags to get the build info during runtime

jrmanes:
hello team,

I've added this small fix to use embed files rather than ldflags, we are missing the build info during the runtime (*in Docker containers*) and I think that's the only way to fix it...

Also, I added a couple of things to the Makefile.
- **gen-embeded**: Generate the embed files for celestia-node binary.
- **docker-build**: Builds the Docker container

The result of this looks like this:

Screenshot 2023-07-03 at 14 07 39

Please let me know what you think about it.

Best,

Jose Ramon Mañes


renaynay (changes requested):
I don't believe `version.go` is used in any unit tests, so can you add `//+build: !test` at top of file pls? Otherwise it'll throw errors

Wondertan:
Are we 100% this is the only way to fix this? I've seen various projects using ldflags and docker together without needing to do this setup.

jrmanes:
> Are we 100% this is the only way to fix this? I've seen various projects using ldflags and docker together without needing to do this setup.

I think so, at least the other changes haven't worked, the other thing we can try is to move the `ldflags` and the build command to the `Dockerfile` [here](https://github.com/celestiaorg/celestia-node/blob/main/Dockerfile#L16), but if we want to keep using only the `make build`...

cmd/celestia/version.go
```diff
@@ -1,15 +1,20 @@
...
+	_ "embed"
+
 	"github.com/spf13/cobra"
 )
 
 var (
-	buildTime       string
-	lastCommit      string
+	buildTime string
+	//go:embed lastCommit.txt
```
Bidon15, line 15: we need to document this lastCommit.txt here. 
And exclude it from the the go test executor

renaynay: see my commetn

renaynay: Hmm the no build directive for tests seems to not work @jrmanes do you mind looking into it?

jrmanes: I can take a look 👌

PR #2426, opened 2023-07-04
chore: bump to rc6

evan-forbes:
## Overview

WIP, as I'm still debugging why https://github.com/celestiaorg/celestia-app/pull/1991 broke the blob module test. When testing using a local branch that doesn't have that change, the test works.

## Checklist


core/eds.go
```diff
@@ -20,21 +19,10 @@ import (
...
-		return nil, err
-	}
-
-	shares := make([][]byte, len(sqr))
-	for i, s := range sqr {
-		shares[i] = s.ToBytes()
-	}
-
-	return da.ExtendShares(shares)
+	return app.ExtendBlock(data, appconsts.LatestVersion)
```
Wondertan, line 25: Should we pass here the version from the header instead?

core/client_test.go
```diff
@@ -10,23 +10,20 @@ import (
...
 	require.NotNil(t, status)
 }
 
 func TestRemoteClient_StartBlockSubscription_And_GetBlock(t *testing.T) {
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second*3)
-	t.Cleanup(cancel)
-
 	client := StartTestNode(t).Client
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*4)
+	t.Cleanup(cancel)
```
Wondertan, line 24: What's the reason for these ctx changes? We have the unspoken convention in the node that ctx is created at the top of the test.

core/testing.go
```diff
@@ -67,40 +42,13 @@ func StartTestNode(t *testing.T) testnode.Context {
...
-
-	cctx, cleanupGRPCServer, err := StartGRPCServer(app, cfg.App, cctx)
-	require.NoError(t, err)
-	t.Cleanup(func() {
-		err := cleanupGRPCServer()
-		require.NoError(t, err)
-	})
-
+func StartTestNodeWithConfig(t *testing.T, cfg *testnode.Config) testnode.Context {
+	cctx, _, _ := testnode.NewNetwork(t, cfg)
```
Wondertan, line 46: Do you cleanup the started network with t underneath?

evan-forbes: yeah

core/client_test.go
evan-forbes: we can create the ctx at the top, but creating after results in a more consistent timing since we don't have to wait for the network to boot up

core/testing.go
evan-forbes: we can remove more code, as there's a bit of duplicated code here that is handled by the app
https://github.com/celestiaorg/celestia-app/blob/965eafb4357376aec31f84f3628f7703c5587f9a/test/util/testnode/full_node.go#L149-L198

core/client_test.go
Wondertan: Usually we set the timeout for the whole test, but as long StartTestNode avoids ctx, we can start timeout after the network start

Wondertan:
Do we want to include here the rc7?

cmwaters:
We can close this since we've merged rc9

PR #2441, opened 2023-07-07
fet(share/eds): Cancel the retrieval when enough shares are collected

walldiss:
## Overview

Previously reconstruction will be started after 1/4 of total shares are collected, but ipld retrieval would continue collecting shares despite there is already enough shares. Cancel the retrieval of nodes early, when there is enough shares collected for eds reconstruction.


Wondertan:
We cancel the context inside of the Retrieve method

walldiss:
Closing in favour of better rework described in: https://github.com/celestiaorg/celestia-node/issues/2446

PR #2443, opened 2023-07-07
header: Remove unnecessary VerifyCommitLightTrusting, use light.VerifyAdjacent

renaynay:
This PR reworks header.Verify such that non-adjacent headers do not need to pass VerifyCommitLightTrusting in order to pass verification as the Syncer in go-header performs sequential verification anyway on all headers synced from genesis -> syncTarget. 

This PR also introduces light.VerifyAdjacent method to the verification logic for adjacent headers as it is audited.

### DIFF
--- a/header/headertest/verify_test.go
+++ b/header/headertest/verify_test.go
@@ -39,14 +39,6 @@ func TestVerify(t *testing.T) {
 		},
 		{
 			prepare: func() libhead.Header {
-				untrusted := *untrustedNonAdj
-				untrusted.Commit = NewTestSuite(t, 2).Commit(RandRawHeader(t))
-				return &untrusted
-			},
-			err: true,
-		},
-		{
-			prepare: func() libhead.Header {
 				untrusted := *untrustedAdj
 				untrusted.RawHeader.LastBlockID.Hash = tmrand.Bytes(32)
 				return &untrusted
--- a/header/verify.go
+++ b/header/verify.go
@@ -6,9 +6,24 @@ import (
 	"time"
 
 	"github.com/tendermint/tendermint/light"
+	"github.com/tendermint/tendermint/types"
 
 	libhead "github.com/celestiaorg/go-header"
 )
+
+// clockDrift defines how much new header's time can drift into
+// the future relative to the now time during verification.
+var clockDrift = 10 * time.Second
+
+// trustingPeriod is period through which we can trust a header's validators set.
+//
+// Should be significantly less than the unbonding period (e.g. unbonding
+// period = 3 weeks, trusting period = 2 weeks).
+//
+// More specifically, trusting period + time needed to check headers + time
+// needed to report and punish misbehavior should be less than the unbonding
+// period.
+var trustingPeriod = 168 * time.Hour
 
 // Verify validates given untrusted Header against trusted ExtendedHeader.
 func (eh *ExtendedHeader) Verify(untrusted libhead.Header) error {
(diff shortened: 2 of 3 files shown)


header/verify.go
```diff
@@ -6,10 +6,25 @@ import (
...
+
+// trustingPeriod is period through which we can trust a header's validators set.
+//
+// Should be significantly less than the unbonding period (e.g. unbonding
+// period = 3 weeks, trusting period = 2 weeks).
+//
+// More specifically, trusting period + time needed to check headers + time
+// needed to report and punish misbehavior should be less than the unbonding
+// period.
+var trustingPeriod = 168 * time.Hour
```
renaynay, line 26: **TODO** this will likely mess with something if there's a discrepancy in syncer.Params and this -- Need to figure out a better way to reconcile.

header/verify.go
```diff
@@ -23,43 +38,39 @@ func (eh *ExtendedHeader) Verify(untrusted libhead.Header) error {
 	}
 
 	isAdjacent := eh.Height()+1 == untrst.Height()
-	if isAdjacent {
-		// Optimized verification for adjacent headers
-		// Check the validator hashes are the same
-		if !bytes.Equal(untrst.ValidatorsHash, eh.NextValidatorsHash) {
```
renaynay, line 29: check is performed in light.VerifyAdjacent

nodebuilder/p2p/genesis.go
```diff
@@ -24,7 +24,7 @@ func GenesisFor(net Network) (string, error) {
 // NOTE: Every time we add a new long-running network, its genesis hash has to be added here.
 var genesisList = map[Network]string{
 	Arabica:        "7A5FABB19713D732D967B1DA84FA0DF5E87A7B62302D783F78743E216C1A3550",
-	Mocha:          "831B81ADDC5CE999EBB0C150B778F76DAAD9E09DF75FACF164B1F11DCE93E2E1",
+	Mocha:          "79A97034D569C4199A867439B1B7B77D4E1E1D9697212755E1CE6D920CDBB541",
```
renaynay, line 27: return mocha to actual genesis hash, syncing seems to be working completely fine with these changes.

header/headertest/verify_test.go
```diff
@@ -37,14 +37,6 @@ func TestVerify(t *testing.T) {
 			},
 			err: true,
 		},
-		{
```
renaynay, line 40: this case is now covered by header.Validate instead which does basic sanity check on Commit

renaynay: However it is NOT done inside of Verify so it is important that Validate + Verify are called on every header

renaynay:
This PR is blocked on a test in go-header that will confirm the Syncer is able to discard an invalid sync target and wait for a valid one to sync the canonical chain

header/verify.go
```diff
@@ -23,43 +38,39 @@ func (eh *ExtendedHeader) Verify(untrusted libhead.Header) error {
...
-				Reason: fmt.Errorf("expected new header to point to last header hash (%X), but got %X)",
-					eh.Hash(),
-					untrst.LastHeader(),
-				),
-			}
-		}
-
+	if !isAdjacent {
 		return nil
 	}
```
liamsi, line 43: Does that mean every header will be just accepted as fully valid now? Or will be verified with a adjacent one at a later point? If so, there should be a comment here explaining why simply returning here is fine.

renaynay: Only non-adjacent headers will skip adjacent header verification logic, this is okay for sync targets - but we will always verify the chain sequentially moving forward.

renaynay: done.

header/verify.go
musalbas: Are we sure the syncer will still work when sycning blocks outside of the weak subjectivity period (with a trusted hash as the sync target)?
cc @liamsi also

Wondertan, line 26: Let's make it artificially low (or high) with the comment that it doesn't have an effect here and is handled in Syncer.

Having it here with a full description will cause more confusion.

header/verify.go
```diff
@@ -23,43 +38,41 @@ func (eh *ExtendedHeader) Verify(untrusted libhead.Header) error {
...
-				Reason: fmt.Errorf("expected new header to point to last header hash (%X), but got %X)",
-					eh.Hash(),
-					untrst.LastHeader(),
-				),
-			}
-		}
-
+	if !isAdjacent {
+		// if the header is non-adjacent, it will be verified later via
+		// VerifyAdjacent once its adjacent (n-1) header is synced
```
Wondertan, line 43: We should make a TODO here and, ideally link to the issue. That's a temporary fix., as we want to have skipping verification long term

header/verify.go
Wondertan: At this point, it does not seem to be worth using light.VerifyAdjacent tbh, let's revert it back

PR #2445, opened 2023-07-07
fix(share/availability): update availability ErrNotAvailable handling logic

walldiss:
## Overview

`ipldFormat.IsNotFound` could no longer reach Availability layer. It will get converted on getter layer to `share.ErrNotFound`. Update outdated error handling.


Wondertan:
Nice catch!

Wondertan (approved):
Is there a way to test this? Like we forgot to updated the error check and ideally there would be a test that would catch it

walldiss:
I've added test for error conversion. From my experience tests like this gets outdated and deleted very quickly since they highly depended on implementattion

Wondertan (dismissed):

renaynay (approved):

renaynay (approved):

Wondertan (approved):

PR #2464, opened 2023-07-12
feat: new metrics - deployment dashboard

jrmanes:
hello team!

I've added these changes to the code to expose some new metrics via Prometheus client to create the deployment dashboard in Grafana, this new dashboard will include info related to the chains, such as branches/tags, number of nodes, URLs, etc... 
*I'm using the values generated by the `ldflags`*

The dashboard will look like this:

Screenshot 2023-07-12 at 13 07 12

*Pending adding the consensus info, it will contain more info*

Please, check the code and let me know if you find something to improve, I'll gladly do it 😊 

Thanks in advance! 🚀 

Jose Ramon Mañes

closes: https://github.com/celestiaorg/devops/issues/401


share/version.go
```diff
@@ -0,0 +1,139 @@
...
+		},
+	}
+
+	for _, metric := range metrics {
+		m := prometheus.NewGaugeVec(
+			prometheus.GaugeOpts{
+				Name: metric.name,
+				Help: metric.help,
+			},
+			[]string{metric.name},
```
walldiss, line 113: Lets create single gauge metric with value 1 instead of multiple. All build info could be put as key:value attributes on this metric. 

This is common approach used for build exposing build info:
https://www.robustperception.io/exposing-the-software-version-to-prometheus/

nodebuilder/node/metrics.go
```diff
@@ -42,9 +51,22 @@ func WithMetrics() error {
...
+		// Observe build info with labels
+		labels := metric.WithAttributes(
+			attribute.String("buildTime", buildTime),
+			attribute.String("lastCommit", lastCommit),
+			attribute.String("semanticVersion", semanticVersion),
+			attribute.String("systemVersion", systemVersion),
+			attribute.String("golangVersion", golangVersion),
+		)
+
+		observer.ObserveFloat64(buildInfoGauge, 1, labels)
```
Bidon15, line 64: why 1 is here? 😅

jrmanes: this is because we are using `meter.Float64ObservableGauge`, and we are using attributes to add all the values that we need (strings), we need to set the value to the metric

for example, the content of one metric looks like this:
```
build_info{
  buildTime="Mon Jul 17 12:41:54 CEST 2023", golangVersion="go1.20.4", 
  instance="12D3KooWLfRoZnjLd1NsyKgS175iEqe1Y4okfhr4sDbpd1MsC1GB",
  job="Celestia-Full/semver-v0.0.2-24-gcde05d0-dev", 
  lastCommit="cde05d0a586c762c3de81f6b35c989c051f603a1", 
  semanticVersion="v0.0.2-24-gcde05d0-dev", 
  systemVersion="amd64/linux"
}
```

but we have to set some value to the metric itself:

Screenshot 2023-07-17 at 13 25 32

nodebuilder/node/metrics.go
```diff
@@ -33,6 +34,14 @@ func WithMetrics() error {
 		return err
 	}
 
+	buildInfoGauge, err := meter.Float64ObservableGauge(
+		"build_info",
```
jrmanes, line 38: q. node-team,
`build_info` is okay or you prefer `celestia_node_build_info`

Wondertan: I prefer less verbosity

nodebuilder/node/metrics.go
Bidon15: Wait. Can't we use another another type of metric? I am generally confused as it means all nodes in the same panel will have an gauge equal to 1, right?

jrmanes: > Wait. Can't we use another another type of metric? I am generally confused as it means all nodes in the same panel will have an gauge equal to 1, right?

The value of `1` is used to indicate that the metric is **present/active**, it's like this [article](https://www.robustperception.io/exposing-the-software-version-to-prometheus/), bc the goal is not to provide a quantitative measurement but rather to expose the values using the attributes


cc: @walldiss

walldiss: Gauge is the correct choice here, you are right. It is there just to submit data with attributes, metric value is not important.

nodebuilder/node/metrics.go
walldiss: it will be prefixed with `meter` name declared on package level. In this case prefix will be `node_`. I think your naming is fine

nodebuilder/node/metrics.go
Bidon15: Sometimes, I feel like the amount of dirty-hacks is just pilling up 😸 
All good for me then. Resolving this convo. Thanks for clarifications

Bidon15: Ah. I don't have rights to resolve the convo. 
@walldiss please consider this as resolved 🙏

Bidon15:
@jrmanes please add `kind:misc` so the label checker CI passes 🙏

walldiss (approved):
LGTM!

Bidon15 (approved):
LGTM. 🪨

Wondertan (approved):

PR #2466, opened 2023-07-12
fix(nodebuilder): fix nil pointer exception and simplify synchronization

Wondertan:
We hit an infamous use case here where nil is not nil. Also, I simplified locking here while thinking that's a race condition.


nodebuilder/store.go
```diff
@@ -144,29 +138,31 @@ func (f *fsStore) Datastore() (_ datastore.Batching, err error) {
 	// TODO(@Wondertan): Make configurable with more conservative defaults for Light Node
 	opts.MaxTableSize = 64 << 20
 
-	f.data, err = dsbadger.NewDatastore(dataPath(f.path), &opts)
+	ds, err := dsbadger.NewDatastore(dataPath(f.path), &opts)
 	if err != nil {
 		return nil, fmt.Errorf("node: can't open Badger Datastore: %w", err)
 	}
 
-	return f.data, nil
```
Wondertan, line 152: f.data wasn't nil because it contains the type, and thus the nil check on Close didn't work.

vgonkivs (dismissed):

renaynay (dismissed):

renaynay (approved):

vgonkivs (approved):

walldiss (approved):

PR #2471, opened 2023-07-13
chore(share/p2p/discovery): Increase discovery retry ticker to minute

renaynay:
Reduce logs, and unnecessary to retry so frequently.

### DIFF
--- a/cmd/celestia/rpc.go
+++ b/cmd/celestia/rpc.go
@@ -195,7 +195,7 @@ func parseParams(method string, params []string) []interface{} {
 		// 1. Height
 		num, err := strconv.ParseUint(params[0], 10, 64)
 		if err != nil {
-			panic("Error parsing gas limit: uint64 could not be parsed.")
+			panic("Error parsing height: uint64 could not be parsed.")
 		}
 		parsedParams[0] = num
 		// 2. NamespaceID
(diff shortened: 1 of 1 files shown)


vgonkivs (approved):

Wondertan (approved):

share/p2p/peers/manager_test.go
```diff
@@ -363,6 +363,8 @@ func TestIntegration(t *testing.T) {
 	})
 
 	t.Run("get peer from discovery", func(t *testing.T) {
+		discovery.DiscoveryRetryTimeout = time.Second
```
Wondertan, line 366: Why not 100*mils?

share/p2p/discovery/discovery.go
```diff
@@ -35,8 +35,8 @@ const (
 	retryTimeout = time.Second
 )
 
-// discoveryRetryTimeout defines time interval between discovery attempts, needed for tests
-var discoveryRetryTimeout = retryTimeout
+// DiscoveryRetryTimeout defines time interval between discovery attempts, needed for tests
+var DiscoveryRetryTimeout = time.Minute
```
Wondertan, line 39: I start to feel a problem with that, like all the light nodes will have to wait for 1 min, after they start, which sucks. Maybe 10 secs?

PR #2477, opened 2023-07-14
feat(nodebuilder): Invoke traces from fx

vgonkivs:
## Overview
Invoke traces from the fx

## Checklist

### DIFF
--- a/nodebuilder/settings.go
+++ b/nodebuilder/settings.go
@@ -147,12 +147,9 @@ func initializeTraces(
 			semconv.ServiceInstanceIDKey.String(fmt.Sprintf("%s/%s", network.String(), peerID.String()))),
 		))
 
-	otel.SetTracerProvider(tp)
-	if len(pyroOpts) == 0 {
-		return nil
+	if len(pyroOpts) > 0 {
+		tp = otelpyroscope.NewTracerProvider(tp, pyroOpts...)
 	}
-
-	tp = otelpyroscope.NewTracerProvider(tp, pyroOpts...)
 	otel.SetTracerProvider(tp)
 	return nil
 }
--- a/nodebuilder/settings.go
+++ b/nodebuilder/settings.go
@@ -67,10 +67,9 @@ func WithPyroscope(endpoint string, nodeType node.Type) fx.Option {
 }
 
 // WithMetrics enables metrics exporting for the node.
-func WithMetrics(metricOpts []otlpmetrichttp.Option, nodeType node.Type, buildInfo *node.BuildInfo) fx.Option {
+func WithMetrics(metricOpts []otlpmetrichttp.Option, nodeType node.Type) fx.Option {
 	baseComponents := fx.Options(
 		fx.Supply(metricOpts),
-		fx.Supply(buildInfo),
 		fx.Invoke(initializeMetrics),
 		fx.Invoke(state.WithMetrics),
 		fx.Invoke(fraud.WithMetrics),
@@ -124,17 +123,16 @@ func initializeTraces(
 	nodeType node.Type,
 	peerID peer.ID,
 	network p2p.Network,
-	buildInfo *node.BuildInfo,
 	opts []otlptracehttp.Option,
 	pyroOpts []otelpyroscope.Option,
 ) error {
-	var tp trace.TracerProvider
 	client := otlptracehttp.NewClient(opts...)
 	exporter, err := otlptrace.New(ctx, client)
 	if err != nil {
 		return fmt.Errorf("creating OTLP trace exporter: %w", err)
 	}
 
+	var tp trace.TracerProvider
 	tp = tracesdk.NewTracerProvider(
 		tracesdk.WithSampler(tracesdk.AlwaysSample()),
 		// Always be sure to batch in production.
@@ -142,8 +140,7 @@ func initializeTraces(
 		// Record information about this application in a Resource.
 		tracesdk.WithResource(resource.NewWithAttributes(
 			semconv.SchemaURL,
-			semconv.ServiceNamespaceKey.String(fmt.Sprintf("Celestia-%s", nodeType)),
-			semconv.ServiceNameKey.String(fmt.Sprintf("semver-%s", buildInfo.SemanticVersion)),
+			semconv.ServiceNamespaceKey.String(nodeType.String()),
 			semconv.ServiceInstanceIDKey.String(fmt.Sprintf("%s/%s", network.String(), peerID.String()))),
 		))
 
@@ -160,7 +157,6 @@ func initializeMetrics(
 	lc fx.Lifecycle,
 	peerID peer.ID,
 	nodeType node.Type,
-	buildInfo *node.BuildInfo,
 	network p2p.Network,
 	opts []otlpmetrichttp.Option,
 ) error {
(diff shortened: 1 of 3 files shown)


nodebuilder/settings.go
```diff
@@ -105,13 +110,61 @@ func WithMetrics(metricOpts []otlpmetrichttp.Option, nodeType node.Type, buildIn
...
+		))
+
+	otel.SetTracerProvider(tp)
+	if len(pyroOpts) == 0 {
+		return nil
+	}
+
+	tp = otelpyroscope.NewTracerProvider(tp, pyroOpts...)
+	otel.SetTracerProvider(tp)
+	return nil
```
Wondertan, line 157:
```suggestion
	if len(pyroOpts) > 0 {
		tp = otelpyroscope.NewTracerProvider(tp, pyroOpts...)
	}
	otel.SetTracerProvider(tp)
	return nil
```

nodebuilder/settings.go
```diff
@@ -105,13 +110,61 @@ func WithMetrics(metricOpts []otlpmetrichttp.Option, nodeType node.Type, buildIn
...
+	tp = tracesdk.NewTracerProvider(
+		tracesdk.WithSampler(tracesdk.AlwaysSample()),
+		// Always be sure to batch in production.
+		tracesdk.WithBatcher(exporter),
+		// Record information about this application in a Resource.
+		tracesdk.WithResource(resource.NewWithAttributes(
+			semconv.SchemaURL,
+			semconv.ServiceNamespaceKey.String(fmt.Sprintf("Celestia-%s", nodeType)),
+			semconv.ServiceNameKey.String(fmt.Sprintf("semver-%s", buildInfo.SemanticVersion)),
+			semconv.ServiceInstanceIDKey.String(fmt.Sprintf("%s/%s", network.String(), peerID.String()))),
```
Wondertan, line 147: Need to double check with Devops if this works for them and can be filtered.

vgonkivs: cc @jrmanes , @Bidon15

jrmanes: I can check it, I'm gonna need to build this commit and deploy it to an env, but it's okay, I'll keep you posted guys! thank you!

vgonkivs:
One more optimisation that we can do is to pass []*resource.Resource to initMetrics/initTraces as this options are equal to both metrics and traces:
```
resource.NewWithAttributes(
			semconv.SchemaURL,
			semconv.ServiceNamespaceKey.String(fmt.Sprintf("Celestia-%s", nodeType)),
			semconv.ServiceNameKey.String(fmt.Sprintf("semver-%s", buildInfo.SemanticVersion)),
			semconv.ServiceInstanceIDKey.String(fmt.Sprintf("%s/%s", network.String(), peerID.String()))),
		)
```

nodebuilder/settings.go
```diff
@@ -125,8 +175,7 @@ func initializeMetrics(
 			semconv.SchemaURL,
 			semconv.ServiceNamespaceKey.String(fmt.Sprintf("Celestia-%s", nodeType.String())),
 			semconv.ServiceNameKey.String(fmt.Sprintf("semver-%s", buildInfo.SemanticVersion)),
-			semconv.ServiceInstanceIDKey.String(peerID.String()),
-		)))
+			semconv.ServiceInstanceIDKey.String(fmt.Sprintf("%s/%s", network.String(), peerID.String())))))
```
renaynay, line 178: let's make the separator a `:` -- doesn't peerID.String() output the full multiaddr or only the key ?

nodebuilder/settings.go
```diff
@@ -105,13 +110,58 @@ func WithMetrics(metricOpts []otlpmetrichttp.Option, nodeType node.Type, buildIn
...
+func initializeTraces(
+	ctx context.Context,
+	nodeType node.Type,
+	peerID peer.ID,
+	network p2p.Network,
+	buildInfo *node.BuildInfo,
+	opts []otlptracehttp.Option,
+	pyroOpts []otelpyroscope.Option,
+) error {
+	var tp trace.TracerProvider
```
renaynay, line 131: nit but since you declare this anyway, just make a switch on `len(pyroOpts)` instead of setting tp the way you're doing now ? Also can you declare tp closer to where you set it?

renaynay:
As I understand, it's still being tested by @jrmanes

jrmanes:
> As I understand, it's still being tested by @jrmanes

yes @renaynay , I'm in contact with Slava about this topic :)

nodebuilder/settings.go
walldiss, line 178: Lets keep only those attributes:
- network/node_type
- peer.ID

vgonkivs: what about "Celestia" prefix?

vgonkivs: @jrmanes , are u ok with @walldiss suggestion?

renaynay: I don't think we need celestia prefix either, just network name is enough and peer ID

vgonkivs: I think we should combine peer.ID and network, but keep nodeType separately

jrmanes: I'd say network/peer.ID makes more sense to me, both are info related to the environment, and the node type can be separated, for me, it's more like if we want to filter by node types rather than useful info of the chain, but no problem if we want to do it in the other way

vgonkivs: Agree. `network/peer.ID` or `network:peer.ID` looks nicer imo

vgonkivs: I'd leave `/` as it's more visible for eyes than `:`

jrmanes: `network:peer.ID` reminds me of the `CELESTIA_CUSTOM` env var format which is 👌 if we want to keep the same format

vgonkivs: left "/" , but I'm open to any suggestions

jrmanes:
hello team!

regarding the latest commit:

This is how it will look like:

We could find the `network/peer.ID` in the services:

Screenshot 2023-07-14 at 16 43 41


and the node type in the metrics:
Screenshot 2023-07-14 at 16 44 09


thank you so much @vgonkivs 😃

nodebuilder/settings.go
walldiss: The point of network/node type is to be able to filter data by specific combination. So I can see only robusta-64 bridge nodes lets say. 
I'm not sure what is the point of combining network and peer.ID. But would love to hear your filtering usecases that will use it.

renaynay (approved):

nodebuilder/settings.go
```diff
@@ -123,10 +169,8 @@ func initializeMetrics(
 		sdk.WithReader(sdk.NewPeriodicReader(exp, sdk.WithTimeout(2*time.Second))),
 		sdk.WithResource(resource.NewWithAttributes(
 			semconv.SchemaURL,
-			semconv.ServiceNamespaceKey.String(fmt.Sprintf("Celestia-%s", nodeType.String())),
-			semconv.ServiceNameKey.String(fmt.Sprintf("semver-%s", buildInfo.SemanticVersion)),
-			semconv.ServiceInstanceIDKey.String(peerID.String()),
-		)))
+			semconv.ServiceNamespaceKey.String(nodeType.String()),
+			semconv.ServiceNameKey.String(fmt.Sprintf("%s/%s", network.String(), peerID.String())))))
```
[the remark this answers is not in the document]
Wondertan: (This one is on metrics)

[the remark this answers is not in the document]
vgonkivs: Yes I know. Decided to keep consistency.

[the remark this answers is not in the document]
vgonkivs: Why do we need this prefix?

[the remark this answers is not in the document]
Wondertan: Ah, ok, look irrelevant on the first look.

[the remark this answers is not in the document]
Wondertan: But should be ok in general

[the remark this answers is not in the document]
vgonkivs: We discussed with @jrmanes to keep it without prefix.

Wondertan (approved):

PR #2481, opened 2023-07-15
chore: cleanup usage of `share.Root`(DAH)

Wondertan:
Somewhere we were using Root and somewhere DAH. Also, we mostly use DAH generator from the app. 
These PRs clean up all the usages and localize the Root/DAH type in `share` pkg as was originally intended.

Up to discussion is rename of `share.Root` to `share.DAH` to minimize discrepancies between app and node. Alternatively, if we wanna keep Root naming as it is, we should also rename DAH name on the ExtendedHeader(and son tags) and in other places, which is breaking.

Now, I think that calling it Root was not right and we should be calling the DataHash type Root, which is in fact the Merkle root of all the shares in the EDS.

### DIFF
--- a/header/header.go
+++ b/header/header.go
@@ -55,14 +55,10 @@ func MakeExtendedHeader(
 	case nil:
 		dah = share.EmptyRoot()
 	default:
-		root, err := share.NewRoot(eds)
+		dah, err = share.NewRoot(eds)
 		if err != nil {
 			return nil, err
 		}
-		dah = &root
-	}
-	if err != nil {
-		return nil, err
 	}
 
 	eh := &ExtendedHeader{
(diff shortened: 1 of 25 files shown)


header/header.go
```diff
@@ -76,22 +72,26 @@ func MakeExtendedHeader(
...
-		dah, err = da.NewDataAvailabilityHeader(eds)
+		root, err := share.NewRoot(eds)
+		if err != nil {
+			return nil, err
+		}
+		dah = &root
 	}
 	if err != nil {
 		return nil, err
 	}
```
renaynay, line 90: then we don't need this err check

renaynay (commented):
LGTM and thank you for this - are we completely sure it's not breaking? I think it still is bc of header field change despite being same underlying type.

walldiss (approved):
Great cleanup PR!

renaynay (approved):

PR #2482, opened 2023-07-15
feat(edssser): introduce EDS Store Stresser and cel-shed utility

Wondertan:
We need a way to stress test EDS Store now and long-term. This PR provides a tool that can be used to run such tests and measure, debug, explore, and fine-tune various aspects of the EDS storing sub-system in isolation.

The main goal is to investigate https://github.com/celestiaorg/celestia-node/discussions/2480 and ensure the future fix/solution is

### DIFF
--- a/cmd/cel-shed/eds_store_stress.go
+++ b/cmd/cel-shed/eds_store_stress.go
@@ -1,0 +1,126 @@
+package main
+
+import (
+	"context"
+	"errors"
+	_ "expvar"
+	"fmt"
+	"math"
+	"net/http"
+	"os"
+
+	"github.com/mitchellh/go-homedir"
+	"github.com/spf13/cobra"
+
+	"github.com/celestiaorg/celestia-node/libs/edssser"
+	"github.com/celestiaorg/celestia-node/nodebuilder"
+	"github.com/celestiaorg/celestia-node/nodebuilder/node"
+)
+
+const (
+	edsStorePathFlag   = "path"
+	edsWritesFlag      = "writes"
+	edsSizeFlag        = "size"
+	edsDisableLogFlag  = "disable-log"
+	edsLogStatFreqFlag = "log-stat-freq"
+	edsCleanupFlag     = "cleanup"
+	edsFreshStartFlag  = "fresh"
+)
+
+func init() {
+	edsStoreCmd.AddCommand(edsStoreStress)
+
+	defaultPath := "~/.edssser"
+	path, err := homedir.Expand(defaultPath)
+	if err != nil {
+		panic(err)
+	}
+
+	pathFlagUsage := fmt.Sprintf("Directory path to use for stress test. Uses %s by default.", defaultPath)
+	edsStoreStress.Flags().String(edsStorePathFlag, path, pathFlagUsage)
+	edsStoreStress.Flags().Int(edsWritesFlag, math.MaxInt, "Total EDS writes to make. MaxInt by default.")
+	edsStoreStress.Flags().Int(edsSizeFlag, 128, "Chooses EDS size. 128 by default.")
+	edsStoreStress.Flags().Bool(edsDisableLogFlag, false, "Disables logging. Enabled by default.")
+	edsStoreStress.Flags().Int(edsLogStatFreqFlag, 10, "Write statistic logging frequency. 10 by default.")
+	edsStoreStress.Flags().Bool(edsCleanupFlag, false, "Cleans up the store on stop. Disabled by default.")
+	edsStoreStress.Flags().Bool(edsFreshStartFlag, false, "Cleanup previous state on start. Disabled by default.")
+
+	// kill redundant print
+	nodebuilder.PrintKeyringInfo = false
+}
+
+var edsStoreCmd = &cobra.Command{
+	Use:   "eds-store [subcommand]",
+	Short: "Collection of eds-store related utilities",
+}
+
+var edsStoreStress = &cobra.Command{
+	Use:          "stress",
+	Short:        `Runs eds.Store stress test over default node.Store Datastore backend (e.g. Badger).`,
+	SilenceUsage: true,
+	RunE: func(cmd *cobra.Command, args []string) (err error) {
+		// expose expvar vars over http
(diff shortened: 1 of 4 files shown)


cmd/cel-shed/eds_store_stress.go
```diff
@@ -0,0 +1,99 @@
...
+	edsEnableLogFlag = "log"
+	edsStoreCleanup  = "cleanup"
+)
+
+func init() {
+	shareCmd.AddCommand(edsStoreStress)
+
+	edsStoreStress.Flags().String(edsStorePathFlag, "", "Directory path to use for stress test. os.TempDir by default.")
+	edsStoreStress.Flags().Int(edsWritesFlag, math.MaxInt, "Total EDS writes to make. MaxInt by default.")
+	edsStoreStress.Flags().Int(edsSizeFlag, 128, "Chooses EDS size. 128 by default.")
```
musalbas, line 29: This actually ends up being ODS size, if you look at `RandEDS`, size*size shares get generated, than the get extended: https://github.com/celestiaorg/celestia-node/blob/9c40a7f379ca4cefa738e265e7d318b2e5245ef2/share/eds/edstest/testing.go#L28

Wondertan: tru, fixed

cmd/cel-shed/eds_store_stress.go
```diff
@@ -0,0 +1,99 @@
...
+			return err
+		}
+
+		stresser, err := edssser.NewEDSsser(path, datastore, cfg)
+		if err != nil {
+			return err
+		}
+
+		stats, err := stresser.Run(cmd.Context())
+		fmt.Printf("%s", stats.Finalize())
```
walldiss, line 96: Lets print stats once in a while just in case app gets stuck or killed by OOM.

cmd/cel-shed/eds_store_stress.go
```diff
@@ -0,0 +1,99 @@
...
+	edsSizeFlag      = "eds-size"
+	edsEnableLogFlag = "log"
+	edsStoreCleanup  = "cleanup"
+)
+
+func init() {
+	shareCmd.AddCommand(edsStoreStress)
+
+	edsStoreStress.Flags().String(edsStorePathFlag, "", "Directory path to use for stress test. os.TempDir by default.")
+	edsStoreStress.Flags().Int(edsWritesFlag, math.MaxInt, "Total EDS writes to make. MaxInt by default.")
```
walldiss, line 28: Given that we want to test store performance on certain amount of blocks stored, would it be easier to use this tool if this setting was rather "max amount of eds stored". So instead of local counter staring from 0 every time this app restarted, it would look for amount of already stored eds and start iterating from there until limit is reached.

I think it might be important, since we want to run this tool to find limits of storage and where things breaks. So it should be ready for failure state and do not expect to always stop gracefully. Would be nice to be able to resume it and restore previous run state.

libs/edssser/edssser.go
```diff
@@ -0,0 +1,117 @@
...
+	err = ss.edsstore.Start(ctx)
+	if err != nil {
+		return stats, err
+	}
+	defer func() {
+		err = errors.Join(err, ss.edsstore.Stop(ctx))
+	}()
+
+	t := &testing.T{}
+	for ctx.Err() == nil && ss.config.EDSWrites != 0 {
```
walldiss, line 59: I might be missing it, but shouldn't ss.config.EDSWrites be backwards iterated somewhere so `ss.config.EDSWrites != 0 ` becomes a termination condition at some point?

cmd/cel-shed/eds_store_stress.go
Wondertan: I like the idea. I will also then disable cleanup by default

walldiss: Maybe you might like an idea to replace "cleanup" option with "clean-start" instead, that will clear previous run data.

libs/edssser/edssser.go
Wondertan: No, I missed the decrement somehow

Wondertan: [6944037](https://github.com/celestiaorg/celestia-node/pull/2482/commits/6944037041b207e67a2897041c6ea592c25bf5b5)

Wondertan: https://github.com/celestiaorg/celestia-node/pull/2482/commits/6944037041b207e67a2897041c6ea592c25bf5b5

cmd/cel-shed/eds_store_stress.go
Wondertan: https://github.com/celestiaorg/celestia-node/pull/2482/commits/6944037041b207e67a2897041c6ea592c25bf5b5

cmd/cel-shed/eds_store_stress.go
Wondertan: Added `fresh` for similar effect

libs/edssser/edssser.go
```diff
@@ -0,0 +1,128 @@
...
+
+type Stats struct {
+	TotalWritten                         int
+	TotalTime, MinTime, MaxTime, AvgTime time.Duration
+	// Deviation ?
+}
+
+func (stats Stats) Finalize() Stats {
+	if stats.TotalTime != 0 {
+		stats.AvgTime = stats.TotalTime / time.Duration(stats.TotalWritten)
```
walldiss, line 109: stats.AvgTime should be float, otherwise it will do int rounding.

distractedm1nd (approved):
This is really great, my only very optional nit would be naming (not so sure about `EDSsser`, don't see whats wrong with `EDSStoreStresser`)

distractedm1nd:
let's of course merge vlad's improvement PR into this branch first

Wondertan:
@distractedm1nd, I just like saying EDSsssssser. It's first of a all a name, not abriviation

Wondertan:
Rebased on main

distractedm1nd (dismissed):

walldiss (approved):

distractedm1nd (approved):

PR #2483, opened 2023-07-17
fix: remove root from traces

vgonkivs:
## Overview
Self-explanatory

## Checklist

### DIFF
--- a/share/eds/retriever.go
+++ b/share/eds/retriever.go
@@ -64,7 +64,6 @@ func (r *Retriever) Retrieve(ctx context.Context, dah *da.DataAvailabilityHeader
 	defer span.End()
 	span.SetAttributes(
 		attribute.Int("size", len(dah.RowRoots)),
-		attribute.String("data_hash", dah.String()),
 	)
 
 	log.Debugw("retrieving data square", "data_hash", dah.String(), "size", len(dah.RowRoots))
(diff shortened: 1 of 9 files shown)


Wondertan (commented):
The fact that we remove roots makes most of the traces meaningless(besides the time which can be done by metrics anyway). Maybe there is a way to disable searching by this attribute to keep roots while solving cardinality? 
cc @walldiss

Wondertan:
Good idea @renaynay. Saving root only on error would remove my concern.

walldiss:
Traces will still contain all info from trace parent, that is stored in context. 
The problem is that root hashes is a bottleneck for tracing collection performance. The fact that 5 instances jaeger instances can serve traces from only 2 nodes is quite bad. 
Those attributes were added because they looked free, but they are not. As for now since we don't know for sure the use cases for those attributes it is fine remove them altogether and add once we absolutely need to.

renaynay:
Okay so @vgonkivs is there a way to just track root as attribute specifically if an error is recorded?

Wondertan:
We understand the problem. 
The concern is why do we need traces at all here if in most cases it does not bring any additional info besides just execution time that is better be covered by metrics? @renaynay‘s idea to include such additional info brings value on demand during the error. 

I don't wanna block this PR and we can move this discussion to the issue, as I want to know why not to remove the traces from here altogether or make them useful.

vgonkivs:
> Okay so @vgonkivs is there a way to just track root as attribute specifically if a critical error is recorded?

Yes, we can leave `root` only for error cases. @walldiss wdyt?

renaynay (approved):

Wondertan (approved):

PR #2511, opened 2023-07-24
feat(blob): improve gas estimation and track min gas price

cmwaters:
## Overview

This PR incorporates the more recent gas estimation method in celestia-app. It also makes use of a new gRPC endpoint for consensus nodes that reveals the nodes min gas price. This enables the blob module to submit transactions with the correct gas price. If the node changes that price, the blob module is able to parse the error and retry using the new gas price.

Additionally, a stubbed CoreAccessor was introduced for when a core endpoint is not provided in order to return a better/more readable error, whereas previously, the CoreAccessor depended on the grpc dial to silently fail during start and any call to the state module would return a "cannot dial address" error.

It now looks like this: 

```
{
  "jsonrpc": "2.0",
  "id": 1,
  "error": {
    "code": 1,
    "message": "node is running without state access"
  }
}
```

## Checklist


blob/service.go
```diff
@@ -24,20 +28,28 @@ var (
...
 	// shareGetter retrieves the EDS to fetch all shares from the requested header.
 	shareGetter share.Getter
-	//  headerGetter fetches header by the provided height
+	// headerGetter fetches header by the provided height
 	headerGetter func(context.Context, uint64) (*header.ExtendedHeader, error)
+	// minGasPrice is the minimum gas price that the node will accept.
+	// NOTE: just because the first node accepts the transaction, does not mean it
+	// will find a proposer that does accept the transaction. Better would be
+	// to set a global min gas price that correct processes conform to.
+	minGasPrice float64
```
Wondertan, line 52: I believe this should be part of the state accessor so that other services and transactions can rely on gas price tweaks introduced here. This would also remove the need for the QueryMinimuGasPrice interface method usage of which is better be implementation detail of general tx submission code path.

Wondertan (changes requested):

blob/service.go
cmwaters: yeah I thought about that initially. I wasn't sure if `core.Accessor` was just supposed to be some "dumb" quasi stateless access point to the consensus node or whether it was supposed to have logic within it. I guess if it's already doing nonce counting then perhaps it's okay. 

I can shift it across but we should be careful about overloading a single component with too much functionality.

cmwaters:
Writing tests...

state/core_access.go
```diff
@@ -174,24 +192,74 @@ func (ca *CoreAccessor) SubmitPayForBlob(
...
-		ca.lastPayForBlob = time.Now().UnixMilli()
-		ca.payForBlobCount++
+	// if the min gas price is not set we query the node that we are connected for it
+	if ca.minGasPrice < 0 {
+		minGasPrice, err := ca.queryMinimumGasPrice(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("querying minimum gas price: %w", err)
+		}
+		ca.minGasPrice = minGasPrice
+	}
```
Wondertan, line 202: This should be synchronized with singleflight protection as SubmitPayForBlob can be called from multiple routines.
(does not the case because no nounce handling yet, but we want concurrent PFBs in the future)

state/core_access.go
```diff
@@ -156,13 +170,17 @@ func (ca *CoreAccessor) constructSignedTx(
...
+// and returns the TxReponse. If gasLim is set to 0, the method will automatically estimate the gas limit. If the fee is
+// negative, the method will use the nodes min gas price multipled by the gas limit.
 func (ca *CoreAccessor) SubmitPayForBlob(
 	ctx context.Context,
 	fee Int,
 	gasLim uint64,
 	blobs []*blob.Blob,
 ) (*TxResponse, error) {
 	if len(blobs) == 0 {
+		// TODO: publilcly facing errors should be able to be handled
```
Wondertan, line 183: Do you mean make the error public var?
I agree in general, but in this particular case, it's useless. It's a sanity check.

state/core_access.go
```diff
@@ -174,24 +192,74 @@ func (ca *CoreAccessor) SubmitPayForBlob(
...
+		estimatedFee = false
+	)
+	// set the fee for the user as the minimum gas price multiplied by the gas limit
+	if fee.IsNegative() {
+		estimatedFee = true
+		fee = sdktypes.NewInt(int64(math.Ceil(ca.minGasPrice * float64(gasLim))))
+	}
+
+	for attempt := 0; attempt < maxRetries; attempt++ {
+		response, err := appblob.SubmitPayForBlob(
```
Wondertan, line 229: Let's fix https://github.com/celestiaorg/celestia-app/issues/1885, while we are on it.
This also means the comment on SubmitPayForBlob method will need to be slightly modified.

Wondertan, line 229: Similarly, all the transactions would benefit from retries. I am not pushing to make it in this PR, but it's not hard to get all birds with one sot either.

Wondertan (commented):
How complicated it is to do gas estimations and retries for all transactions and not just for SubmitPFB?

state/core_access.go
cmwaters: More practically its good for testing

state/core_access.go
cmwaters: Yeah I can look into that after this PR

cmwaters:
I am not able to write all the tests I would like because of the functionality that celestia-app needs to provide in their test infra. I'm going to work on this (on celestia-app side). We can still merge this if we think logically it makes sense and I can make an issue and follow up on the missing tests in a different PR

state/core_access.go
Wondertan: Then they should be private.

Wondertan: Let's remove the TODO until there is a clear need for such error var

state/core_access.go
```diff
@@ -174,24 +194,70 @@ func (ca *CoreAccessor) SubmitPayForBlob(
...
-		err = errors.Join(err, sdkErrors.ABCIError(response.Codespace, response.Code, response.Logs.String()))
+	// we only estimate gas if the user wants us to (by setting the gasLim to 0). In the future we may want
+	// to make these arguments optional.
+	if gasLim == 0 {
+		blobSizes := make([]uint32, len(blobs))
+		for i, blob := range blobs {
+			blobSizes[i] = uint32(len(blob.Data))
+		}
+
+		// TODO: the default gas per byte and the default tx size cost per byte could be changed through governance
```
Wondertan, line 211: Pls add you handle to the TODO

cmwaters: Can I do it in a separate PR

state/core_access.go
```diff
@@ -523,6 +521,32 @@ func (ca *CoreAccessor) QueryRedelegations(
...
+	ca.mtx.Lock()
+	defer ca.mtx.Unlock()
+	ca.minGasPrice = minGasPrice
+}
+
+func (ca *CoreAccessor) markSuccessfulPFB() {
+	ca.mtx.Lock()
+	defer ca.mtx.Unlock()
+	ca.lastPayForBlob = time.Now().UnixMilli()
+	ca.payForBlobCount++
```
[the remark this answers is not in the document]
cmwaters: What do you mean by synced?

state/core_access.go
```diff
@@ -455,6 +521,48 @@ func (ca *CoreAccessor) QueryRedelegations(
...
+}
+
+// QueryMinimumGasPrice returns the minimum gas price required by the node.
+func (ca *CoreAccessor) queryMinimumGasPrice(
+	ctx context.Context,
+) (float64, error) {
+	rsp, err := nodeservice.NewServiceClient(ca.coreConn).Config(ctx, &nodeservice.ConfigRequest{})
+	if err != nil {
+		return 0, err
+	}
```
Wondertan, line 557: I didn't look into the impl of it, but what if we construct it only once on start?

state/core_access.go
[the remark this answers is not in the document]
Wondertan: synchronized

state/core_access.go
Wondertan: I am asking to make it `//TODO(@cmwaters)`. Why a separate PR?

cmwaters: ah yup sorry will do

state/core_access.go
cmwaters: We can if you want but I assumed this was pretty lightweight (i.e. just boilerplate code doing the encoding/decoding)

Wondertan: Ok then

state/core_access_test.go
```diff
@@ -31,3 +48,134 @@ func TestLifecycle(t *testing.T) {
...
+// configurations. Provided accounts will be funded and their keys can be
+// accessed in keyring returned client.Context. All rpc, p2p, and grpc addresses
+// in the provided configs are overwritten to use open ports. The node can be
+// accessed via the returned client.Context or via the returned rpc and grpc
+// addresses. Provided genesis options will be applied after all accounts have
+// been initialized.
+//
+// FIXME: this function is copied from celestia-app because it doesn't currently
+// support the MinGasPrice query
+func NewNetwork(
```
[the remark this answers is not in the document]
Wondertan: Blocked on #2527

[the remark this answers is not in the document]
cmwaters: Fixed [clean up test](https://github.com/celestiaorg/celestia-node/pull/2511/commits/3724392ad1dc75485f9774a7ebaa9eca2224e787)

state/core_access.go
```diff
@@ -54,8 +62,15 @@ type CoreAccessor struct {
 	rpcPort  string
 	grpcPort string
 
+	// these fields are mutatable and thus need to be protected by a mutex
+	mtx             sync.Mutex
```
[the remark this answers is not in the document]
cmwaters: what's wrong with `mtx`? I've always called it that

renaynay:
```
{
  "jsonrpc": "2.0",
  "id": 1,
  "error": {
    "code": 1,
    "message": "node is running without state access"
  }
}
```

nodebuilder/node_test.go
```diff
@@ -43,14 +43,8 @@ func TestLifecycle(t *testing.T) {
 			err := node.Start(ctx)
 			require.NoError(t, err)
 
-			// ensure the state service is running
```
renaynay, line 46: useless checks

state/core_access.go
```diff
@@ -90,7 +105,11 @@ func (ca *CoreAccessor) Start(ctx context.Context) error {
 
 	// dial given celestia-core endpoint
 	endpoint := fmt.Sprintf("%s:%s", ca.coreIP, ca.grpcPort)
-	client, err := grpc.DialContext(ctx, endpoint, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	client, err := grpc.DialContext(
+		ctx,
+		endpoint,
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+		grpc.WithBlock())
```
Wondertan, line 112: What do?

nodebuilder/state/core.go
```diff
@@ -20,6 +20,11 @@ func coreAccessor(
 	sync *sync.Syncer[*header.ExtendedHeader],
 	fraudServ libfraud.Service,
 ) (*state.CoreAccessor, *modfraud.ServiceBreaker[*state.CoreAccessor]) {
+	if corecfg == core.DefaultConfig() {
```
Wondertan, line 23: Could you check for the exact fields that are missing? This feels wrong and fragile.

nodebuilder/state/module.go
```diff
@@ -25,14 +25,24 @@ func ConstructModule(tp node.Type, cfg *Config) fx.Option {
...
 			fx.OnStart(func(ctx context.Context, breaker *modfraud.ServiceBreaker[*state.CoreAccessor]) error {
+				if breaker == nil {
+					return nil
+				}
 				return breaker.Start(ctx)
 			}),
 			fx.OnStop(func(ctx context.Context, breaker *modfraud.ServiceBreaker[*state.CoreAccessor]) error {
+				if breaker == nil {
+					return nil
+				}
```
Wondertan, line 36: Returning nil on start and stop would be cleaner imo

nodebuilder/state/core_accessor_stub.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/celestiaorg/celestia-node/blob"
+	"github.com/celestiaorg/celestia-node/state"
+)
+
+var ErrNoStateAccess = errors.New("node is running without state access")
+
+// stubbedCoreAccessor provides a stub for the state module to return
+// errors when state endpoints are accessed without a running connection
+// to a core endpoint.
+type stubbedCoreAccessor struct{}
```
Wondertan, line 18: Let's call it after module. It's first of all the stub for the module, so that users running the node without state received a nice error if access the moduler through RPC.

Wondertan, line 18: This should be a separate PR, tbh

state/metrics.go
```diff
@@ -10,6 +10,12 @@ import (
 var meter = otel.Meter("state")
 
 func WithMetrics(ca *CoreAccessor) {
+	if ca == nil {
+		// short-circuit if node is running without state access
+		log.Error("node is running without state access, cannot enable metrics for state module")
+		return
+	}
```
Wondertan, line 17: There is a way to make FX not call this func when CoreAccess does not exist. Also this logs will be annoying, imo

state/core_access.go
Wondertan: Answered in sync

state/core_access.go
```diff
@@ -54,8 +62,15 @@ type CoreAccessor struct {
...
 	grpcPort string
 
+	// these fields are mutatable and thus need to be protected by a mutex
+	lock            sync.Mutex
 	lastPayForBlob  int64
 	payForBlobCount int64
+	// minGasPrice is the minimum gas price that the node will accept.
+	// NOTE: just because the first node accepts the transaction, does not mean it
+	// will find a proposer that does accept the transaction. Better would be
+	// to set a global min gas price that correct processes conform to.
```
Wondertan, line 72: Link to the issue in the app

state/metrics.go
Wondertan, line 17: There is an if IIRC in other PR that does exactly this

nodebuilder/state/module.go
```diff
@@ -27,9 +27,15 @@ func ConstructModule(tp node.Type, cfg *Config, coreCfg *core.Config) fx.Option
 		fxutil.ProvideIf(coreCfg.IsEndpointConfigured(), fx.Annotate(
 			coreAccessor,
 			fx.OnStart(func(ctx context.Context, breaker *modfraud.ServiceBreaker[*state.CoreAccessor]) error {
+				if breaker == nil {
+					return nil
+				}
```
Wondertan, line 32: Solved in a diff PR

nodebuilder/state/module.go
```diff
@@ -27,9 +27,15 @@ func ConstructModule(tp node.Type, cfg *Config, coreCfg *core.Config) fx.Option
...
 				return breaker.Start(ctx)
 			}),
 			fx.OnStop(func(ctx context.Context, breaker *modfraud.ServiceBreaker[*state.CoreAccessor]) error {
+				if breaker == nil {
+					return nil
+				}
```
Wondertan, line 38: Same

state/core_access.go
```diff
@@ -459,6 +524,53 @@ func (ca *CoreAccessor) QueryRedelegations(
...
+	ca.lock.Lock()
+	defer ca.lock.Unlock()
+	ca.minGasPrice = minGasPrice
+}
+
+func (ca *CoreAccessor) getMinGasPrice() float64 {
+	ca.lock.Lock()
+	defer ca.lock.Unlock()
+	return ca.minGasPrice
+}
```
vgonkivs, line 556: Does not it deserve to be a separate data structure inside the coreAccesor? Imo would be much cleaner

state/core_access.go
```diff
@@ -459,6 +524,53 @@ func (ca *CoreAccessor) QueryRedelegations(
...
+
+// QueryMinimumGasPrice returns the minimum gas price required by the node.
+func (ca *CoreAccessor) queryMinimumGasPrice(
```
vgonkivs, line 559: This could be exported, so we can expose it through the RPC

state/core_access.go
Wondertan: We can park this in the issue

state/core_access.go
Wondertan: We can export min gas price for sure. Care for an issue?

state/core_access.go
vgonkivs: ok

state/core_access.go
vgonkivs: yes

vgonkivs (approved):

Issue #2503, opened 2023-07-19
blob.GetAll not returning correct base64 data, GetShare and GetSharesByNamespace also don't work


jcstein:
### Celestia Node version

v0.11.0-rc8

### OS

mac

### Install tools

docs

### Others

nope

### Steps to reproduce it

Q: if i post plain-text data `'"gm"'` and retrieve it, the response is in base64, correct?
i’m updating the based64.xyz tutorial, and during testing I used `'"gm"'` as a check, the base64 share that i retrieve is: `SW1kdElnPT0=` which converts from base64 to text: `ImdtIg==`, not `gm` (plus some metadata bytes). i’m on arabica-9 v0.11.0-rc8
i previously had `celestia rpc share GetSharesByNamespace "$(celestia rpc header GetByHeight $HEIGHT | jq '.result.dah' -r)" GHTmQvXd5Yk=` but now am using `celestia rpc blob GetAll $HEIGHT 0x42690c204d39600fddd3`

the commands that previously worked, also do not work anymore as expected, detailed in this issue: 
https://github.com/celestiaorg/docs/issues/914

updated tutorial is at https://based64.xyz/

i also went through the [node tutorial](https://docs.celestia.org/developers/node-tutorial/) page, and the examples on retrieval work with CLI when submitting hex.

### Expected result

returns blob in base64 which can be converted to base64 to text. this is how it worked originally and there seems to be a bug on the serialization or deserialization in rpc.go (based on what Hlib and I were looking at yesterday)

### Actual result

returns data in wrong format

### Relevant log output

```Shell
celestia rpc share GetSharesByNamespace "$(celestia rpc header GetByHeight 103692 | jq '.result.dah' -r)" 0x42690c204d39600fddd3
{
  "jsonrpc": "2.0",
  "id": 1,
  "error": {
    "code": -32700,
    "message": "unmarshaling params for 'share.GetSharesByNamespace' (param: *share.Namespace): illegal base64 data at input byte 20"
  }
}

celestia rpc share GetShare "$(celestia rpc header GetByHeight 101810 | jq '.result.dah' -r)" 101810 0
{
  "jsonrpc": "2.0",
  "id": 1,
  "error": {
    "code": 0,
    "message": "fatal error calling 'share.GetShare': panic in rpc method 'share.GetShare': runtime error: index out of range [101810] with length 8"
  }
}
```

### Notes

I know for a fact that I submitted data to the namespace and at the block heights above.

[...]

Issue #2500, opened 2023-07-18
bug:  illegal base64 data at input byte during unmarshalling blob.Commitment


vgonkivs:
Blob Module requests fail sometimes with `illegal base64 data at input byte` error through rpc.

renaynay:
Can you reproduce this @vgonkivs ?

vgonkivs:
Haven't tried it yet, but I'm pretty sure that it will.

vgonkivs:
fixed in #2515 and #2521

PR #2515, opened 2023-07-27
fix(rpc): fix commands handling

vgonkivs:
## Overview
Fixes all found rpc issues

Fixes https://github.com/celestiaorg/celestia-node/issues/2503 and https://github.com/celestiaorg/celestia-node/issues/2500

## Checklist

### DIFF
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -144,20 +144,16 @@ func (b *Blob) Namespace() share.Namespace {
 }
 
 type jsonBlob struct {
-	Namespace share.Namespace `json:"namespace"`
-	// According to the Go documentation:
-	// Array and slice values encode as JSON arrays,
-	// except that []byte encodes as a base64-encoded string.
-	// So, this led to the incorrect representation of the `Data`.
-	Data         string     `json:"data"`
-	ShareVersion uint32     `json:"share_version"`
-	Commitment   Commitment `json:"commitment"`
+	Namespace    share.Namespace `json:"namespace"`
+	Data         []byte          `json:"data"`
+	ShareVersion uint32          `json:"share_version"`
+	Commitment   Commitment      `json:"commitment"`
 }
 
 func (b *Blob) MarshalJSON() ([]byte, error) {
 	blob := &jsonBlob{
 		Namespace:    b.Namespace(),
-		Data:         string(b.Data),
+		Data:         b.Data,
 		ShareVersion: b.ShareVersion,
 		Commitment:   b.Commitment,
 	}
@@ -173,7 +169,7 @@ func (b *Blob) UnmarshalJSON(data []byte) error {
 
 	b.Blob.NamespaceVersion = uint32(blob.Namespace.Version())
 	b.Blob.NamespaceId = blob.Namespace.ID()
-	b.Blob.Data = []byte(blob.Data)
+	b.Blob.Data = blob.Data
 	b.Blob.ShareVersion = blob.ShareVersion
 	b.Commitment = blob.Commitment
 	b.namespace = blob.Namespace
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -144,16 +144,20 @@ func (b *Blob) Namespace() share.Namespace {
 }
 
 type jsonBlob struct {
-	Namespace    share.Namespace `json:"namespace"`
-	Data         []byte          `json:"data"`
-	ShareVersion uint32          `json:"share_version"`
-	Commitment   Commitment      `json:"commitment"`
+	Namespace share.Namespace `json:"namespace"`
+	// According to the Go documentation:
+	// Array and slice values encode as JSON arrays,
+	// except that []byte encodes as a base64-encoded string.
+	// So, this led to the incorrect representation of the `Data`.
+	Data         string     `json:"data"`
+	ShareVersion uint32     `json:"share_version"`
+	Commitment   Commitment `json:"commitment"`
 }
 
 func (b *Blob) MarshalJSON() ([]byte, error) {
 	blob := &jsonBlob{
 		Namespace:    b.Namespace(),
-		Data:         b.Data,
+		Data:         string(b.Data),
 		ShareVersion: b.ShareVersion,
 		Commitment:   b.Commitment,
 	}
@@ -169,7 +173,7 @@ func (b *Blob) UnmarshalJSON(data []byte) error {
 
 	b.Blob.NamespaceVersion = uint32(blob.Namespace.Version())
 	b.Blob.NamespaceId = blob.Namespace.ID()
-	b.Blob.Data = blob.Data
+	b.Blob.Data = []byte(blob.Data)
 	b.Blob.ShareVersion = blob.ShareVersion
 	b.Commitment = blob.Commitment
 	b.namespace = blob.Namespace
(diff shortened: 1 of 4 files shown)


jcstein:
testing out https://based64.xyz (so testing retrieval of submitted plain-text) on this branch and it is not, functioning as expected, so this doesn't resolve the issue of plain text data that is expected to be returned in base64, being returned as base64

Screenshot 2023-07-27 at 14 05 01

my version matches last commit on this PR:

```bash
Semantic version: v0.11.0-rc8-13-ga9ac16a2
Commit: a9ac16a24181094162b54b2ce5e3616037038fdc
Build Date: Thu Jul 27 14:01:39 CEST 2023
System version: arm64/darwin
Golang version: go1.20.2
```

as an example, i will try to submit "gm" and retrieve base64 that converts to string as "gm". alternatively, as explained in this comment: https://github.com/celestiaorg/celestia-node/issues/2503#issuecomment-1641782724 the only difference is i am using the version above
1. `celestia light start --core.ip consensus-full-arabica-9.celestia-arabica.com --p2p.network arabica`
2. `export CELESTIA_NODE_AUTH_TOKEN=$(celestia light auth admin --p2p.network arabica)`
3. `celestia rpc blob Submit 0x42690c204d39600fddd3 '"gm"'`
4. `celestia rpc blob GetAll [blockheightfrom#3] 0x42690c204d39600fddd3`
5. it should convert to "gm" with some metadata, it doesn't
Screenshot 2023-07-27 at 14 10 00
6. oddly, converting from base64 -> hex and then from hex -> text returns the same data as converting base64 -> text as shown above: 
Screenshot 2023-07-27 at 14 11 43

jcstein:
linking docs context https://github.com/celestiaorg/docs/pull/915#issuecomment-1653527378

jcstein:
oops 😅

vgonkivs:
I feel like #2503 has two potential issues inside.
1. Issue with converting(that @jcstein mentioned in Actual Result)
2. Issue with incorrect handling of rpc commands(see Relevant Output). 

This PR is now fixing only issue #2. I need to look deeper what is wrong with the serialisation.

blob/blob.go
```diff
@@ -144,16 +144,20 @@ func (b *Blob) Namespace() share.Namespace {
...
-	Namespace    share.Namespace `json:"namespace"`
-	Data         []byte          `json:"data"`
-	ShareVersion uint32          `json:"share_version"`
-	Commitment   Commitment      `json:"commitment"`
+	Namespace share.Namespace `json:"namespace"`
+	// According to the Go documentation:
+	// Array and slice values encode as JSON arrays,
+	// except that []byte encodes as a base64-encoded string.
+	// So, this led to the incorrect representation of the `Data`.
+	Data         string     `json:"data"`
```
Wondertan, line 152: I think it's better to keep []byte and make the client side aware of this instead. The reason it's base64 is data compression and we should not loose such property. As well as it will require changes for all the types returning []byte on the server.

Wondertan (dismissed):

share/getters/cascade.go
```diff
@@ -37,7 +38,12 @@ func (cg *CascadeGetter) GetShare(ctx context.Context, root *share.Root, row, co
 		attribute.Int("col", col),
 	))
 	defer span.End()
-
+	if row >= len(root.RowRoots) || col >= len(root.ColumnRoots) {
```
Wondertan, line 41: When can this happen? 🤔

vgonkivs: Thought the rpc

`celestia rpc share GetShare "$(celestia rpc header GetByHeight 101810 | jq '.result.dah' -r)" 101810 0`

101810 -> rowIndex
0 -> colIndex

blob/blob.go
vgonkivs: Client-side does not anything about any concrete type. It receives `json.RawMessage` as a result(where data is already marshalled incorrectly).

share/getters/cascade.go
Wondertan: Ah, so the rpc request is malformed

vgonkivs: Yes

blob/blob.go
Wondertan: It does not but the better solution to make it do. The RawMessage can be converted to Blob struct where base64 will converted automatically back.

vgonkivs: Ok

vgonkivs: After more time I realised, that it won't work as you described. the raw message is:
```
{
  "jsonrpc": "2.0",
  "result": [
    {
      "data": "some_data"
      "share_version": 0,
      "commitment": "x0Z/D9P56ijvcVZ+s4al/d1cDNzQ+RaLNZKPUxITJNQ="
    }
  ],
  "id": 1
}
```

Raw message is not a Blob itself or jsonBlob. This is a struct and the `result` field is an empty interface with the underlying type of []jsobBlob. 
Also, it is clear that the "issue" is not on the unmarshaling part but on marshaling, just because golang marshals []byte this way. 
IMO, fixing it on the client side is not a good solution because we are fixing it only in our code, but anyone else who will be using blob API will face this issue and will have to deal with it(when will find the root cause 😄 ). 
My impression that keeping jsonBlob(not the Blob struct itself).Data as a string is the best solution. And yes, if we have other places w []byte, we need to fix it this way.

vgonkivs: n/a

cmd/celestia/rpc.go
```diff
@@ -119,9 +119,17 @@ var rpcCmd = &cobra.Command{
 
 func parseParams(method string, params []string) []interface{} {
 	parsedParams := make([]interface{}, len(params))
-
+	validateParamsFn := func(has, want int) error {
+		if has < want {
+			return fmt.Errorf("rpc: invalid amount of params. has=%d, want=%d", has, want)
+		}
```
renaynay, line 125: just do has != want

share/getters/cascade.go
```diff
@@ -37,7 +38,12 @@ func (cg *CascadeGetter) GetShare(ctx context.Context, root *share.Root, row, co
...
+		err := fmt.Errorf("cascade/get-share: invalid indexes were provided:rowIndex=%d, colIndex=%d."+
+			"rowsAmount=%d, colsAmount=%d", row, col, len(root.RowRoots), len(root.ColumnRoots))
```
renaynay, line 43: i would just log the row and col index and the width of the square

renaynay: len col + len row always ==

vgonkivs:
FYI @jcstein , we changed the `plaintext` flag to `base64` in order to convert data to the human readable format by default. To get the data as base64 string, you have to pass `--base64=true`

share/getters/cascade.go
```diff
@@ -37,7 +38,12 @@ func (cg *CascadeGetter) GetShare(ctx context.Context, root *share.Root, row, co
 		attribute.Int("col", col),
 	))
 	defer span.End()
-
+	if row >= len(root.RowRoots) || col >= len(root.ColumnRoots) {
+		err := fmt.Errorf("cascade/get-share: invalid indexes were provided:rowIndex=%d, colIndex=%d."+
```
distractedm1nd, line 42: indicies* but i will still approve, its very optional

distractedm1nd (approved):

jcstein:
okay, so what happens if no `--plaintext` or `--base64` flag?

vgonkivs:
Data will be converted to the plain text by default

PR #2516, opened 2023-07-28
fix(share/eds): print error for stuck register shard

walldiss:
Sometimes register shard can take longer, that provided context timeout. This PR logs errors if register shard is stuck.


Wondertan (commented):
Is this still the case after all the hangups are removed?

walldiss:
Latest tests shown not hangups. But still, we don't want to lose event of hangup and info all related info of if it happens.

Wondertan:
Agreed, yet sounds like it better be a metric

renaynay (commented):
I agree with @Wondertan , this is better recorded as a metric.

walldiss:
Metric would need logic implemented in this PR. There are 3 cases for hangups, so I would rather have specified log lines.  Also metrics are optional, but it will be important to be able to find those errors from logs in case this problem happens again

Wondertan (approved):

distractedm1nd (approved):
Have we discussed making Put async?

walldiss:
Async Put would just kick the can down the road.

Issue #2479, opened 2023-07-14
ipld/bitswap: Prepend block height to CIDs to eliminate Badger inverted index


musalbas:
If we can prepend the block height to CIDs, when requesting chunks over Bitswap, we can eliminate the Badger inverted index for CID => CAR file.

This is a simpler version of what's proposed in https://github.com/celestiaorg/celestia-node/issues/2116, as it doesn't involve adding the coordinates of shares to CIDs - just the block height.

ramin:
resolved by shwap

PR #2517, opened 2023-07-28
fix(share/eds)!: use separate badger for inverted index

walldiss:
Having same badger instance for dagstore shards and inverted index has proven to have hangup on `Put` operation for 1-10min. During tests it has shown, that splitting badger into 2 instances solves the hangup issue. 

Also once https://github.com/celestiaorg/celestia-node/issues/2479 is implemented and inverted index will needs to be removed. Separate badger will be much easier for users migration.


Wondertan (commented):
> Having same badger instance for dagstore shards and inverted index has proven to have hangup on Put operation for 1-10min. 

According to @musalbas, no hangup happens with a single instance with 20 compactors. Is it worth verifying again to avoid having two badgers when we can have one only?

walldiss:
Considering inverted index is major bottleneck right now an we plan to remove it soon. I think it is worth to do separate badgers just for sake of easy migration.

If we keep inverted index in same badger as shards badger, we would need to manually remove each key. It would mean to find all keys by prefix and remove them one by one, which will take ages.

I've started test to try 20 compactors on single badger

Wondertan:
Agreed

Wondertan (dismissed):

Wondertan:
We still may wanna keep all the storage breaking PRs together

distractedm1nd (approved):

Wondertan (approved):

distractedm1nd (approved):

PR #2521, opened 2023-07-29
feat(cmd): use rpc client instead of `http.Request`

vgonkivs:
## Overview
Ideas behind such rework are pretty simple:
1) to simplify code.
Me and @Wondertan love generic code but IMO the rpc is too complicated and hard to understand and maintain. Also, it contained a lot of hidden bugs that were found by @jcstein(https://github.com/celestiaorg/celestia-node/issues/2500, https://github.com/celestiaorg/celestia-node/issues/2503). I believe that extracting each module(or even a method) to its own command will simplify this package a lot.
2) move from the raw http.Request.
We've already had a rpc client but continue using http.Request. Also, the generic approach did not give us control over the returned value.

TODO: extract errors
## Checklist


jcstein:
ack - LGTM

Screenshot 2023-07-31 at 11 50 40 (2)

jcstein (approved):

cmd/celestia/blob.go
```diff
@@ -0,0 +1,236 @@
...
+	Use:   "Get [params]",
+	Args:  cobra.ExactArgs(3),
+	Short: "Returns the blob for the given namespace by commitment at a particular height.",
+	Run: func(cmd *cobra.Command, args []string) {
+		num, err := strconv.ParseUint(args[0], 10, 64)
+		if err != nil {
+			fmt.Fprintln(os.Stderr,
+				fmt.Errorf("error parsing height: uint64 could not be parsed"),
+			)
+			os.Exit(1)
```
Wondertan, line 56: You can use RunE instead

cmd/celestia/blob.go
```diff
@@ -0,0 +1,236 @@
...
+	"github.com/celestiaorg/celestia-node/blob"
+	"github.com/celestiaorg/celestia-node/share"
+)
+
+type response struct {
+	Result interface{} `json:"result"`
+	Error  string      `json:"error,omitempty"`
+}
+
+var rpcClient *client.Client
```
Wondertan, line 26: Nah, let's hide inside of context. 

Persistent sets it and cmds get it

cmd/celestia/blob.go
```diff
@@ -0,0 +1,236 @@
...
+		rpcClient = client
+		return nil
+	},
+	PersistentPostRun: func(_ *cobra.Command, _ []string) {
+		rpcClient.Close()
+	},
+}
+
+var getCmd = &cobra.Command{
+	Use:   "Get [params]",
```
Wondertan, line 47: Decapitalize

cmd/celestia/blob.go
```diff
@@ -0,0 +1,236 @@
...
+
+	"github.com/celestiaorg/celestia-node/api/rpc/client"
+	"github.com/celestiaorg/celestia-node/blob"
+	"github.com/celestiaorg/celestia-node/share"
+)
+
+type response struct {
+	Result interface{} `json:"result"`
+	Error  string      `json:"error,omitempty"`
+}
```
Wondertan, line 24: Why do we need this format for output? Why not output the error as a regular error to fmt and only the result in JSON or whatever

cmd/celestia/blob.go
```diff
@@ -0,0 +1,236 @@
...
+			if err != nil {
+				fmt.Fprintln(os.Stderr, errors.New("error decoding blob: hex string could not be decoded"))
+				os.Exit(1)
+			}
+			blobData = decoded
+		case strings.HasPrefix(args[1], "\""):
+			// user input an utf string that needs to be encoded to base64
+			src := []byte(args[1])
+			blobData = make([]byte, base64.StdEncoding.EncodedLen(len(src)))
+			base64.StdEncoding.Encode(blobData, []byte(args[1]))
```
Wondertan, line 153: I am not sure we need to encode if it will be encoded by JSON. I think the original purpose was to encoded to make it compatible with the server side which auto-decoded, but now it's done automatically. We can also likely remove tryDecode

cmd/celestia/blob.go
```diff
@@ -0,0 +1,236 @@
...
+		default:
+			// otherwise, we assume the user has already encoded their input to base64
+			blobData, err = base64.StdEncoding.DecodeString(args[1])
+			if err != nil {
+				fmt.Fprintln(os.Stderr, errors.New("error decoding blob data: base64 string could not be decoded"))
+				os.Exit(1)
+			}
+		}
```
Wondertan, line 161: These encodings are a mess...
In previous case, we encode to base64 to send it and here we decode from base64 to send raw bytes.

cmd/celestia/blob.go
vgonkivs: No, without tryDecode we will still received unexpected format

cmd/celestia/blob.go
vgonkivs: done

cmd/celestia/blob.go
vgonkivs: moved to rpc command

cmd/celestia/blob.go
vgonkivs: done

cmd/celestia/rpc.go
```diff
@@ -61,13 +64,34 @@ func init() {
 		false,
 		"Print JSON-RPC request along with the response",
 	)
+	blobCmd.AddCommand(getCmd, getAllCmd, submitCmd, getProofCmd)
+	rpcCmd.AddCommand(blobCmd)
```
Wondertan, line 68: Why not make blob a separate cmd then? It's not an part of general purpose rpc command anymore, rather standalone cmd.

cmd/celestia/rpc.go
```diff
@@ -61,13 +64,34 @@ func init() {
 		false,
 		"Print JSON-RPC request along with the response",
 	)
+	blobCmd.AddCommand(getCmd, getAllCmd, submitCmd, getProofCmd)
```
Wondertan, line 67: Nit: do registering of blob sub commands in init defined next to the blob cmd 

This way it'c clearer to see and control the relation between those

cmd/celestia/blob.go
```diff
@@ -0,0 +1,200 @@
...
+
+		fmt.Fprintln(os.Stdout, string(output))
+		return nil
+	},
+}
+
+func prepareOutput(data interface{}) ([]byte, error) {
+	bytes, err := json.MarshalIndent(response{
+		Result: data,
+	}, "", "  ")
```
Wondertan, line 195: Why wrap in response?

cmd/celestia/blob.go
```diff
@@ -0,0 +1,200 @@
...
+			return fmt.Errorf("error creating a blob:%v", err)
+		}
+
+		var output []byte
+		height, err := client.Blob.Submit(cmd.Context(), []*blob.Blob{parsedBlob})
+		if err != nil {
+			output, err = prepareOutput(err)
+			if err != nil {
+				return err
+			}
```
Wondertan, line 135: Why do we need to give error back in json and not just a string?

cmd/celestia/blob.go
```diff
@@ -0,0 +1,200 @@
...
+		}
+
+		fmt.Fprintln(os.Stdout, string(output))
+		return nil
+	},
+}
+
+var getAllCmd = &cobra.Command{
+	Use:   "getAll [params]",
```
Wondertan, line 71: get-all ?

cmd/celestia/rpc.go
distractedm1nd: Hmmm... not sure this is good UX if only one of the modules isn't an 'rpc subcommand'. I would +1 sticking to the current way slava has implemented it

cmd/celestia/blob.go
vgonkivs: In this case we should such pattern in all commands in the rpc(get-by-height,get-share). It does not look good to me and I'd prefer camel case

cmd/celestia/rpc.go
vgonkivs: Agree. We have a 'rpc' command that that allows to interacts with different modules through the rpc.

cmd/celestia/blob.go
vgonkivs: Because in some cases IMO we have to emphasise that the error is a result of request, received from the service. Otherwise you will have printed error and recommendations for usage current command, that does not make sense.

cmd/celestia/blob.go
vgonkivs: We have such approach for all responses mainly because in case of  multiple returned values, you can use indexing over the output. Please check getAll cmd and  pipelining with jq

cmd/celestia/rpc.go
renaynay: I think it's actually fine to separate blob out as it's going to be the most-used and would decrease surface area of the command that needs to be doc'd and kept up to date.

cmd/celestia/rpc.go
```diff
@@ -115,116 +139,6 @@ func parseParams(method string, params []string) []interface{} {
 			panic(fmt.Sprintf("Error parsing namespace: %v", err))
 		}
 		parsedParams[1] = namespace
-	case "Submit":
```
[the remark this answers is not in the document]
renaynay: We shouldn't remove SubmitPayForBlob regardless from state

I think submit is fine bc the new submit command covers it?

renaynay (commented):
One general question I have that doesn't need to be resolved by this PR is whether we should either 1. allow `Submit` to also return the commitment that corresponds to the blob created in the submit process or 2. create a separate endpoint that returns the commitment as well.

The reason i'm asking is bc it's not possible to do a GetProof or Get using purely the cli as it requires a commitment as arg.

vgonkivs:
Both 1. and 2. require us to modify the Blob Service interface, which is a breaking change for it. Also, iirc, we discussed such ideas and agreed that `GetAll` will be enough.(requesting getAll will return you the whole blob)

Wondertan:
I think @renaynay brought a good point. Returning Commitment on Submit would be very useful.

@vgonkivs, option 1 does not need any modifications to API. The cmd side can get commitment from the blob itself and on success return it to the user.

cmd/celestia/rpc.go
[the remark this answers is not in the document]
Wondertan: I am fine with removing this earlier.

vgonkivs:
oops, yes, you are right. Thanks, @renaynay

cmd/celestia/rpc.go
Wondertan: * Consider the case where we fully migrate to cmds for each module. In such case `rpc` becomes redundant. * It makes sense to have `rpc` when the module namespace is generic.
* Blob is what the user wants, RPC is a way of accessing the API. Which can be HTTP or IPC. Would you put `http` there instead?

cmd/celestia/blob.go
Wondertan: Note that it will feel. wrong to all heavy terminal users, like myself

cmd/celestia/blob.go
Wondertan: There are standard unix pipes for that. I don't think we should enforce json here

cmd/celestia/rpc.go
vgonkivs: I don't have objections to doing this once we migrate to cmdc from each module. But for now it does not look good: we have rpc command with all modules and we have a blob command that lives separately somewhere in the space.
Regarding 2: why don't we consider the case when a user would want to get data through the http and not ipc? then we should have 2 global commands http/ipc.

cmd/celestia/blob.go
vgonkivs: So, is this about feelings and not about usage? E.g. I feel more comfortable using camel case. 
@renaynay, @distractedm1nd, @walldiss need your opinionы lady and gentlemen

Wondertan: Yeah, ergonomics and usage are about feelings. I can't think of any CLI with camel case in it tbh

vgonkivs: Ok, I can't agree with it tbh, but @renaynay also voted for your suggestion, so I will rework it.

Wondertan: > I can't agree with it tbh

On what specifically? Definition of ergonomics or that unix people don't use camelCase for CLIs?

cmd/celestia/rpc.go
vgonkivs: done

cmd/celestia/blob.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+var blobCmd = &cobra.Command{
+	Use:   "blob [command]",
+	Short: "Allows to interact with the Blob Service via JSON-RPC",
+	Args:  cobra.NoArgs,
+}
+
+var getCmd = &cobra.Command{
+	Use:   "get [params]",
```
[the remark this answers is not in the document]
vgonkivs: done

cmd/celestia/blob.go
```diff
@@ -0,0 +1,200 @@
...
+	Args:  cobra.ExactArgs(3),
+	Short: "Returns the blob for the given namespace by commitment at a particular height.",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		client, err := rpcClient(cmd.Context())
+		if err != nil {
+			return err
+		}
+
+		num, err := strconv.ParseUint(args[0], 10, 64)
```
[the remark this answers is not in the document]
vgonkivs: done

cmd/celestia/blob.go
```diff
@@ -0,0 +1,189 @@
...
+
+	"github.com/spf13/cobra"
+
+	"github.com/celestiaorg/celestia-node/blob"
+	"github.com/celestiaorg/celestia-node/share"
+)
+
+type response struct {
+	Result interface{} `json:"result"`
+}
```
Wondertan, line 18: very optional nit: move it to where its used - close to `prepareOutput`

cmd/celestia/blob.go
Wondertan: Keeping it JSON makes sense, as we all got used to it. In a perfect world, we would have a flag to specify which encoding to use with plaintext by default and option to choose json.

cmd/celestia/blob.go
```diff
@@ -0,0 +1,189 @@
...
+			Commitment: parsedBlob.Commitment,
+		}
+
+		output, err := prepareOutput(response, err)
+		if err != nil {
+			return err
+		}
+
+		fmt.Fprintln(os.Stdout, string(output))
+		return nil
```
Wondertan, line 136: nit: There is an opportunity to compress code in all cmds by making `printOutput`

cmd/celestia/blob.go
vgonkivs: can even insert it inside prepareOutput

cmd/celestia/blob.go
vgonkivs: I'd not want non-blob-related cmds in this PR. Can make it in subsequent. Here https://github.com/celestiaorg/celestia-node/pull/2515 for example

Wondertan: I meant all blob cmds

cmd/celestia/rpc.go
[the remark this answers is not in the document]
vgonkivs: so..... should I return all cmds back 😄 ?

renaynay (dismissed):
Looks like there's a lint issue, but otherwise I'm okay with it 🙏🏻 tested it myself (not the plaintext option yet, but that will be a really nice feature)

I'll make an issue for the commitment to also be returned for Submit.

vgonkivs:
> I'll make an issue for the commitment to also be returned for Submit.

I've already implemented it @renaynay

vgonkivs:
FYI @jcstein, changes in Blob's commands:
* they are now lower case(submit, get)
* using `-` between words instead of camel case(get-all, get-proof)
* submitting the data now returns not only the height but also the commitment of the Blob.
* Added flag `--plaintext` that converts the Blob's data from base64 string(`gm` instead of `Z20=`). To enable this command you should add `--plaintext=true` in the end of the command

vgonkivs:
Снимок экрана 2023-08-04 в 14 18 42
Снимок экрана 2023-08-04 в 14 21 38

distractedm1nd (approved):

renaynay (approved):

jcstein:
thanks for the context! a question on this before opening a PR to the docs:

are the rest of the methods still camelcase? this seems like it would get confusing as a user if we didn't change all methods.

i.e.

`celestia rpc blob get-all [args...]`
`celestia rpc blob get [args...]`
and then
`celestia rpc state BalanceForAddress [args...]`
`celestia rpc p2p Info [args...]`
`celestia rpc header GetByHeight [ args...]`
`celestia rpc node Info`
`celestia rpc das SamplingStats`

jcstein:
suggestion: allow `--plaintext` to set `--plaintext=TRUE` by default

jcstein:
i also think we need to set the API version to match the expected node release version cc @MSevey @adlerjohn 

```bash
celestia rpc node Info

{
  "jsonrpc": "2.0",
  "result": {
    "type": 2,
    "api_version": "v0.2.1"
  },
  "id": 1
}
```

vgonkivs:
> are the rest of the methods still camelcase? this seems like it would get confusing as a user if we didn't change all methods.

Yeah, they will be fixed in subsequent PRs

jcstein:
this plaintext flag is goated 👏

Issue #2116, opened 2023-04-20
[Feature Request]: Change CID to reduce round trips for Bitswap to 1


nashqueue:
### Implementation ideas

Currently, the CID is the hash of the nodes inside the NMT or the hash of the Share. 

Therefore a Light Node will ask for the child hashes of a row root as the CIDs. And then traverse the NMT to the Share. This means we have log(EDS square size) amount of roundtrips as this is the height of the NMT tree. 

The idea is to have the position in the share derive the CID. To uniquely identify the share, you would use the hash(height of celestia, index in the square). The index would be from [0, EDS squaresize ^2 -1]. 

After an IWANT[h1,h2,h3...], you will be getting a (share, path to share, celestis-height) as a response. The path is the merkle proof to the share. From this proof, you can derive the position inside the square. Now you can point to the shares directly without traversing the tree.

Bitswap is content addressable and expects the CID to match the data. Therefore the hash would need to match the data provided on a Bitswap level. Fortunately, it supports custom hash schema. 

custom_hash(share,path,height) = hash(height,extract_index(path))

The resulting effect would be to reduce the Network load on Sampling Light Nodes.

musalbas:
Why not just request the actual coordinates of the share, without sending proof?

nashqueue:
The LN is requesting the coordinates. (through the index) . 

You need to send the Merkle proof of the Root. Basically, instead of traversing the ODS and giving you one Node at a time, Bitswap will give you all relevant nodes at once.

ramin:
should be taken care of with shwap

nashqueue:
I think we can close this now with Shawp 🚀

PR #2532, opened 2023-08-02
fix(eds/blockstore): Puts on EDSStore Blockstore

### LINKED ISSUE #2424: eds: Blockstore provided by the Store should support Put
## Context
During the design and implementation of #1099, we missed the necessity of caching the inner and leaf nodes during reconstruction via the Blockservice, leading to Bitswap having no access to them. This is due to [all the Puts being discarded](https://github.com/celestiaorg/celestia-node/blob/main/share/eds/blockstore.go#L88) on the Blockstore.

## Problems
* FNs do not share proofs and data during reconstruction between each other.
* FNs cannot generate BEFPs for withheld blocks. 

## Solution
The proposed solution here is changing Blockstore to support Put operations:
* Blockstore over on-disk Datastore
	* Dagstore's Blockstore would write to disk and, on reads, check the index first and then the Datastore.
	* The default badger Datastore scales well for write-heavy loads, and considering sparse read patterns for reconstruction and BEFP, it should also suit read requirements.
	* The only complication is clean-up, which can be naively solved by iterating over each stored piece with the DeleteBlock method on Blockstore at the end of the reconstruction session(or even after we put an indexed eds to avoid interruption of serving data to other nodes over bitswap)
	
* Blockstore over in-mem Datastore
	* Similar cleanup complication.
	* Similar implementation time estimate as for on-disk
	* Less robust. Does not tolerates restarts.
	
	
## Testing
We have a (flaky) Swamp reconstruction test that asserts that a FN can reconstruct a block from light nodes only, but we also need to add a test similar to the test below, which validates that FNs can collectively rebuild the block.  
https://github.com/celestiaorg/celestia-node/blob/4d986940d662649eba4b0d3daa85268318204ff2/share/availability/full/reconstruction_test.go#L162

Once we know that this works fine in Swamp, we should also replicate the same scenario on TG.

distractedm1nd:
Closes #2424 

Alternative Design Discussions:
- Using in-memory instead of in on disk: During an unexpected shutdown or in the case of AsyncGetter, cleanup would not occur and blocks would be stuck in the store indefinitely. Using an in-memory blockstore would fix this issue upon restarts
-  Using a local Blockgetter to pass to NewErrByzantine: Instead of putting the blocks into the EDS blockstore, the retrieval session could make an in-memory blockstore on the fly to hand to NewErrByzantine when needed. This would look cleaner in the code/make sense architecturally, but it would mean full nodes cannot share these shares with each other during reconstruction.


Wondertan (commented):
> Using an in-memory blockstore would fix this issue upon restarts

Losing all the shares and proofs collected and necessary for reconstruction after an unexpected shutdown. I think we should go with better resilience and availability with a tradeoff of storing a bit more data.

I don't like the current state of cleaning up in Retriever, and I think this is where and why your alternative ideas are coming from. Besides, the current solution is not atomic, and there will be gaps in data availability between the moment of Retriever cleaning up and EDSStore preparing and making all the data on disk available for serves. This hints that cleaning up logic can and should live in the EDSStore itself and be executed right after the EDS is successfully stored. 

As discussed in our standup, I suggest cleaning up logic in a separate PR, greatly simplifying this one and giving us more time to think about various options while still solving the original problem with the consequence of storing copies of rarely reconstructable EDSes.

walldiss:
>Using a local Blockgetter to pass to NewErrByzantine: Instead of putting the blocks into the EDS blockstore, the retrieval session could make an in-memory blockstore on the fly to hand to NewErrByzantine when needed. This would look cleaner in the code/make sense architecturally, but it would mean full nodes cannot share these shares with each other during reconstruction.

This solution will not allow to serve intermediate nodes while block is being sampled, but it is rare case, that has little benefit. All of those needs to happen at the same time for it to give benefit:
- it is only possible, when full node is using ipld sampling, which is very rare.
- Request for intermediate node should happen while sampling is still in process, that is also another rare event.

With all those rarity, it will allow client to sample just a little bit faster, increasing the pool of peers with available blocks, with those that are in the process of sampling.

Downsides:
- On the other hand this change introduces major engineering complexity, related to maintaining and **synchronising** the global state (blockstore). This state needs to be cleared on events, that are not tightly linked and thus do not have ACID properties. It feels there are lot of potential problems with leaked objects, that are forgotten to removed.
-  if items are stored to the disk in the flight, it introduces more write load to the storage, that is already a bottleneck. Given it is very rare read, that is not good utilisation of disk i/o throughput.
- [This important improvement](https://github.com/celestiaorg/celestia-node/issues/2116) will be much harder to implement, unless global state is removed.

Making local blockservice when BEFP happens without implementing globally managed blockstore, will reduce this PR to like 30 lines of code, while losing not much.

Wondertan:
> Yes, this solution will not allow to serve intermediate nodes while block is being sampled, 

> it is only possible, when full node is using ipld sampling, which is very rare.

> Request for intermediate node should happen while sampling is still in process, that is also another rare event.


@walldiss, This is a _strong_ requirement to consider reconstruction finished. It's part of a worst case scenario and security foundation the whole system works on.

cc @musalbas

walldiss:
Also it is worth considering to decompose this issue into 2:
- provide proofs for befp constructor. This is relatively easy, considering there is proofAdder, that contains all proofs to be used to construct befp. 
- allow intermediate nodes cache. This is not easy, so I would suggest to decouple it.

Wondertan:
We agreed to:
* Keep blockstore over on-disk datastore
* Extract cleaning up logic as a separate issue/PR(@distractedm1nd, please do the issue)
	* Where we discuss the cleaning-up strategy
* Implement a benchmark for Retriever running over the production datastore.

share/eds/blockstore.go
```diff
@@ -30,29 +33,46 @@ var (
...
 func (bs *blockstore) Has(ctx context.Context, cid cid.Cid) (bool, error) {
 	keys, err := bs.store.dgstr.ShardsContainingMultihash(ctx, cid.Hash())
-	if errors.Is(err, ErrNotFound) {
-		return false, nil
+	k := dshelp.MultihashToDsKey(cid.Hash())
+	if errors.Is(err, ErrNotFound) || errors.Is(err, ErrNotFoundInIndex) {
+		// key wasn't found in top level blockstore, but could be in datastore while being reconstructed
+		dsHas, dsErr := bs.ds.Has(ctx, k)
+		if dsErr != nil {
+			return false, ErrNotFound
```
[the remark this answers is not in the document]
distractedm1nd: great catch

distractedm1nd:
After removing cleanup I realized the benchmark doesnt really make sense, I cannot benchmark without introducing an option that disables puts @walldiss , or am I understanding incorrectly?

go.mod
```diff
@@ -81,6 +81,8 @@ require (
 	google.golang.org/protobuf v1.31.0
 )
 
+require github.com/ipfs/go-ipfs-ds-help v1.1.0
```
Wondertan, line 84: Separated?

share/eds/retriever.go
```diff
@@ -30,6 +30,10 @@ var (
 	tracer = otel.Tracer("share/eds")
 )
 
+type RetrieverOption func(*Retriever)
+
+type CleanupFunc func(context.Context, cid.Cid) error
+
```
Wondertan, line 36: Could you remove it? Because of the atomicity argument, it's not practical to do cleanup in Retriever. Cleaning should happen _only_ after the EDS gets indexed and stored on disk.

share/eds/accessor_cache.go
```diff
@@ -16,7 +16,7 @@ import (
 
 var (
 	defaultCacheSize = 128
-	errCacheMiss     = errors.New("accessor not found in blockstore cache")
+	errCacheMiss     = errors.New("accessor not found in Blockstore cache")
```
Wondertan, line 19:
```suggestion
	errCacheMiss     = errors.New("accessor not found in blockstore cache")
```

share/eds/accessor_cache.go
```diff
@@ -114,7 +114,7 @@ func (bc *blockstoreCache) unsafeAdd(
 ) (*accessorWithBlockstore, error) {
 	blockStore, err := accessor.Blockstore()
 	if err != nil {
-		return nil, fmt.Errorf("failed to get blockstore from accessor: %w", err)
+		return nil, fmt.Errorf("failed to get Blockstore from accessor: %w", err)
```
Wondertan, line 117:
```suggestion
		return nil, fmt.Errorf("failed to get blockstore from accessor: %w", err)
```

Wondertan (commented):
Need to review the test as well

share/eds/accessor_cache.go
```diff
@@ -65,7 +65,7 @@ func (bc *blockstoreCache) evictFn() func(_ interface{}, val interface{}) {
 		if err != nil {
 			log.Errorf("couldn't close accessor after cache eviction: %s", err)
 		}
-		bc.metrics.observeEvicted(err != nil)
```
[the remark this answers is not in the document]
distractedm1nd: oops, probably from rebase

go.mod
distractedm1nd: dont know how this happened, only ran go mod tidy

share/eds/accessor_cache.go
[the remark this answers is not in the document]
distractedm1nd: wait... have no idea how this happened. probably typed it manually from diff and mixed it up

nodebuilder/tests/reconstruct_test.go
```diff
@@ -81,6 +82,164 @@ func TestFullReconstructFromBridge(t *testing.T) {
...
+Test-Case: Full Node reconstructs blocks from each other, after unsuccessfully syncing the complete
+block from LN subnetworks. Analog to TestShareAvailable_DisconnectedFullNodes.
+*/
+func TestFullReconstructFromFulls(t *testing.T) {
+	light.DefaultSampleAmount = 10 // s
+	const (
+		blocks = 10
+		btime  = time.Millisecond * 300
+		bsize  = 8  // k
+		lnodes = 12 // c - total number of nodes on two subnetworks
```
Wondertan, line 95: How much more time higher param take?

Wondertan (commented):
Test LGTM

Has conflicts

nodebuilder/tests/reconstruct_test.go
distractedm1nd: I couldn't get it to pass with the other parameters, so we decided we should test this scenario in TG and have the swamp be this mild version - since avail tests already show its possible with the other parameter set anyways

share/eds/retriever.go
```diff
@@ -120,8 +120,11 @@ type retrievalSession struct {
 }
 
 // newSession creates a new retrieval session and kicks off requesting process.
-func (r *Retriever) newSession(ctx context.Context, dah *da.DataAvailabilityHeader) (*retrievalSession, error) {
+func (r *Retriever) newSession(
+	ctx context.Context, dah *da.DataAvailabilityHeader,
+) (*retrievalSession, error) {
```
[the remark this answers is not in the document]
distractedm1nd: Ah. This was because I changed the signature and changed it back, will fix

nodebuilder/tests/reconstruct_test.go
distractedm1nd: For clarification, I am convinced the only reason the other swamp tests work with so many light nodes is because the parameters ensure a very high share redundancy. The shares are available from many sources

Wondertan: avail tests made me think of diff avail first : )

go.mod
```diff
@@ -80,6 +80,8 @@ require (
...
```
Wondertan, line 83: Again :sob:

Wondertan (dismissed):

go.mod
distractedm1nd: go mod tidy is doing this. will fix

Wondertan:
Snyk is complaining for some unknown reason

share/eds/blockstore.go
```diff
@@ -75,27 +97,35 @@ func (bs *blockstore) GetSize(ctx context.Context, cid cid.Cid) (int, error) {
...
 
-// PutMany is a noop on the EDS blockstore, but it does not return an error because it is called by
-// bitswap. For clarification, an implementation of PutMany does not make sense in this context
-// because it is unclear which CAR file the blocks should be written to.
-//
-// TODO: throw errUnsupportedOperation after issue #1440
-func (bs *blockstore) PutMany(context.Context, []blocks.Block) error {
-	return nil
+func (bs *blockstore) PutMany(ctx context.Context, blocks []blocks.Block) error {
+	if len(blocks) == 1 {
```
renaynay, line 112: could PutMany ever get called on len 0? If so, should we short circuit before?

share/eds/inverted_index.go
```diff
@@ -14,6 +15,9 @@ import (
 
 const invertedIndexPath = "/inverted_index/"
 
+// ErrNotFoundInIndex is returned instead of ErrNotFound if the multihash doesn't exist in the index
+var ErrNotFoundInIndex = fmt.Errorf("failed to lookup index")
```
renaynay, line 19: this error message is confusing - shouldn't it be "does not exist in index" then?

nodebuilder/tests/reconstruct_test.go
```diff
@@ -81,6 +82,164 @@ func TestFullReconstructFromBridge(t *testing.T) {
...
+	)
+
+	ctx, cancel := context.WithTimeout(context.Background(), swamp.DefaultTestTimeout)
+	t.Cleanup(cancel)
+
+	sw := swamp.NewSwamp(t, swamp.WithBlockTime(btime))
+	fillDn := swamp.FillBlocks(ctx, sw.ClientContext, sw.Accounts, bsize, blocks)
+
+	const defaultTimeInterval = time.Second * 5
+	cfg := nodebuilder.DefaultConfig(node.Bridge)
```
renaynay, line 105: you never actually use this config.

nodebuilder/tests/reconstruct_test.go
```diff
@@ -81,6 +82,164 @@ func TestFullReconstructFromBridge(t *testing.T) {
...
+	cfg.Share.Discovery.PeersLimit = 0
+	setTimeInterval(cfg, defaultTimeInterval)
+	bridge := sw.NewBridgeNode()
+	addrsBridge, err := peer.AddrInfoToP2pAddrs(host.InfoFromHost(bridge.Host))
```
renaynay, line 109: you don't need to extract this, check how `swamp.SetBootstrapper` works

nodebuilder/tests/reconstruct_test.go
```diff
@@ -81,6 +82,164 @@ func TestFullReconstructFromBridge(t *testing.T) {
...
+				return err
+			}
+			subs[i] = sub
+			lights1[i] = light
+			return light.Start(errCtx)
+		})
+		errg.Go(func() error {
+			lnConfig := nodebuilder.DefaultConfig(node.Light)
+			setTimeInterval(lnConfig, defaultTimeInterval)
+			lnConfig.Header.TrustedPeers = append(lnConfig.Header.TrustedPeers, addrsBridge[0].String())
```
renaynay, line 142: unnecessary, check sw.SetBootstrapper

nodebuilder/tests/reconstruct_test.go
```diff
@@ -81,6 +82,164 @@ func TestFullReconstructFromBridge(t *testing.T) {
...
+	})
+	require.Error(t, errg.Wait())
+	cancelErr()
+
+	// Reconnect FNs
+	sw.Connect(t, full1, full2)
+
+	errg, bctx := errgroup.WithContext(ctx)
+	for i := 10; i < blocks+11; i++ {
+		i := i
```
renaynay, line 228: do we still need to do this on 1.21? also i don't think you need to do this in general

share/eds/blockstore.go
distractedm1nd: Yes, good catch

nodebuilder/tests/reconstruct_test.go
distractedm1nd: I'll check

distractedm1nd: can't do it yet - we'd have to build with a GOEXPERIMENT flag

share/eds/blockstore.go
```diff
@@ -30,29 +33,43 @@ var (
...
 		cache: cache,
+		ds:    namespace.Wrap(ds, blockstoreCacheKey),
 	}
 }
 
 func (bs *blockstore) Has(ctx context.Context, cid cid.Cid) (bool, error) {
 	keys, err := bs.store.dgstr.ShardsContainingMultihash(ctx, cid.Hash())
-	if errors.Is(err, ErrNotFound) {
-		return false, nil
+	k := dshelp.MultihashToDsKey(cid.Hash())
```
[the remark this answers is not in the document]
distractedm1nd: replacing w inline here because it is separated, but not inlining when directly above Get call because it is easier to read (less nested) and the compiler makes this optimization anyways

nodebuilder/tests/reconstruct_test.go
renaynay: regardless you're not passing this var into a routine so why copy?

renaynay:
BTW @vgonkivs rebased BEFP tests on this PR and they pass 🙏🏻

vgonkivs (approved):
LGTM

PR #2536, opened 2023-08-04
feat(share/store): add eds store metrics

walldiss:
## Overview

I hate boilerplate, but I think it is reasonable here. If I generalise all operations right now, it will be huge pain later to modify specific ones later (like put metric is already different).

### DIFF
--- a/share/eds/metrics.go
+++ b/share/eds/metrics.go
@@ -43,6 +43,7 @@ type metrics struct {
 	getTime              metric.Float64Histogram
 	hasTime              metric.Float64Histogram
 	listTime             metric.Float64Histogram
+	getAccessorTime      metric.Float64Histogram
 
 	longOpTime metric.Float64Histogram
 	gcTime     metric.Float64Histogram
@@ -97,6 +98,12 @@ func (s *Store) WithMetrics() error {
 		return err
 	}
 
+	getAccessorTime, err := meter.Float64Histogram("eds_store_get_accessor_time_histogram",
+		metric.WithDescription("eds store get accessor time histogram(s)"))
+	if err != nil {
+		return err
+	}
+
 	longOpTime, err := meter.Float64Histogram("eds_store_long_operation_time_histogram",
 		metric.WithDescription("eds store long operation time histogram(s)"))
 	if err != nil {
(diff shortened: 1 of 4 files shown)


share/eds/metrics.go
```diff
@@ -0,0 +1,259 @@
...
+
+func (m *metrics) observeGetAccessor(ctx context.Context, dur time.Duration, cached, failed bool) {
+	if m == nil {
+		return
+	}
+	if ctx.Err() != nil {
+		ctx = context.Background()
+	}
+
+	m.listTime.Record(ctx, dur.Seconds(), metric.WithAttributes(
```
renaynay, line 256: Is there no separate histogram for this metric (get accessor)?

walldiss: Nice catch! Fixed

share/eds/store.go
```diff
@@ -164,22 +168,30 @@ func (s *Store) gc(ctx context.Context) {
...
+	err := s.put(ctx, root, square)
+	switch {
+	case errors.Is(err, dagstore.ErrShardExists):
+		s.metrics.observePut(ctx, time.Since(tnow), putExists, square.Width())
+	case err != nil:
+		s.metrics.observePut(ctx, time.Since(tnow), putFailed, square.Width())
+	default:
+		s.metrics.observePut(ctx, time.Since(tnow), putOK, square.Width())
+	}
+	utils.SetStatusAndEnd(span, err)
```
Wondertan, line 186: Readability nit
```suggestion

  result := putOk 
	tnow := time.Now()
	err := s.put(ctx, root, square)
	switch {
	case errors.Is(err, dagstore.ErrShardExists):
		result = putExists
	case err != nil:
		result = putFailed
	}
	utils.SetStatusAndEnd(span, err)
	s.metrics.observePut(ctx, time.Since(tnow), putOK, square.Width())
```

Wondertan (approved):

share/eds/store.go
walldiss: Applied. Had to do manually, since you last line of changes needs to use result

Wondertan: All good nw

Wondertan (approved):

renaynay (approved):

PR #2550, opened 2023-08-09
nodebuilder/p2p: Decouple `ChainID` from `Network`

### LINKED ISSUE #2548: p2p: Decouple ChainID from DA networkID
Make it possible for DA network to run a different ID than the core chain ID (but still use the core chain ID within the DA network ID)

Also another issue found - it's possible to run a bridge node against core with a different chain ID.

renaynay:
Reasoning for this feature can be found in the issue linked below.

Resolves #2548


walldiss:
What is the reason for format "ChainID-da-NetworkID`? If those are independent (while related),  maybe we can use 2 separate config params and provide them separately instead of parsing both values from single param?

nodebuilder/p2p/network.go
```diff
@@ -50,6 +64,28 @@ func (n Network) String() string {
...
+	// parse the ChainID from the custom Network name
+	networkName := strings.Split(net.String(), "-")
+	switch len(networkName) {
+	case 3:
+		// handles case where no suffix has been added to core consensus chainID
+		return ChainID(networkName[0])
+	case 4:
+		return ChainID(strings.Join(networkName[:2], "-"))
+	}
+	return ""
```
renaynay, line 81: This should never happen, but if it does, it will be caught as an error on Start inside of store.Init

nodebuilder/p2p/network.go
```diff
@@ -50,6 +62,28 @@ func (n Network) String() string {
 	return string(n)
 }
 
+// ChainIDFromNetwork parses the ChainID from the Network name.
+func ChainIDFromNetwork(net Network) ChainID {
+	if !strings.Contains(net.String(), "da") {
```
musalbas, line 67: Doesn't this break if there's "da" in chain id, such as darkroast-3-da-x?

musalbas:
I feel there's a lot of weird/undocumented behavior that can occur from the fact that the separator is "-da-" (for example, you can't have a chain id called "dark-roast-3-da-x"), and it would make more sense to use a unique character as a separator, such as mocha-3/foo.

nodebuilder/p2p/network.go
vgonkivs: if this should never happen, can we simply panic here in order to show that smth went wrong?

nodebuilder/p2p/network.go
renaynay: Yes this does and I realised that earlier, but we can't have a special separator be `/` as directory creation will break.

We can use `:` or `.` potentially.

nodebuilder/p2p/network.go
renaynay: not applicable anymore

nodebuilder/p2p/network_test.go
```diff
@@ -0,0 +1,26 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestChainIDFromNetwork(t *testing.T) {
+	var tests = []struct {
+		name            string
+		expectedChainID string
+	}{
+		{name: "mocha-3-da-custom", expectedChainID: "mocha-3"},
```
musalbas, line 15: How are these tests still passing if : is the separator now?

renaynay: I forgot to update tests, my bad. I was pushing code while on call with john lol.

distractedm1nd (approved):

PR #2551, opened 2023-08-09
fix(cmd): fix formatting during submit

vgonkivs:
## Overview
As we expect direct types every time, I would suggest skipping error handling in case of format data(and simply return data if casting fails). I think this is the best way to keep logic generic and avoid any additional bool flags, options etc.

## Checklist


jcstein (approved):
worked for me as expected

```bash
celestia rpc blob submit 0x42690c204d39600fddd3 'gm'
{
  "result": {
    "uint64": 232874,
    "commitment": "IXg+08HV5RsPF3Lle8PH+B2TUGsGUsBiseflxh6wB5E="
  }
}
```

distractedm1nd (approved):

renaynay (approved):

walldiss (approved):

PR #2558, opened 2023-08-10
fix(header): Only bridge node should panic on data root mismatch in e…

### LINKED ISSUE #2555: header: Do not panic on data root mismatch on unmarshal
As the fields are public, if a full node or light node is sent an ExtendedHeader with mismatched Row/Col roots that don't compute to the DataRoot, the node will panic.

We should do a severe WARN log and ban the peer instead.

renaynay:
This PR fixes a DoS first discovered by @Wondertan and then secondarily by @vgonkivs 🤠 

Only bridge nodes should panic on receiving a header where the computed data root does not match the DataHash in the RawHeader on ExtendedHeader validation.

Resolves #2555


core/listener.go
```diff
@@ -160,6 +160,10 @@ func (cl *Listener) handleNewSignedBlock(ctx context.Context, b types.EventDataS
 	// generate extended header
 	eh, err := cl.construct(ctx, &b.Header, &b.Commit, &b.ValidatorSet, eds)
 	if err != nil {
+		var errMismatch *header.ErrDataRootMismatch
+		if errors.As(err, &errMismatch) {
+			panic(err.Error())
+		}
```
Wondertan, line 166: Why only panic on mismatch? If Bridge gives us invalid header for any other reason panicking makes sense. This would aloes simplify code a bit by removing special node type.

header/header.go
```diff
@@ -16,6 +17,8 @@ import (
 	"github.com/celestiaorg/rsmt2d"
 )
 
+var log = logging.Logger("header")
```
Wondertan, line 20: May overlap with header pkg

header/header.go
```diff
@@ -146,8 +149,13 @@ func (eh *ExtendedHeader) Validate() error {
...
 	// ensure data root from raw header matches computed root
 	if !bytes.Equal(eh.DAH.Hash(), eh.DataHash) {
-		panic(fmt.Sprintf("mismatch between data hash commitment from core header and computed data root "+
-			"at height %d: data hash: %X, computed root: %X", eh.Height(), eh.DataHash, eh.DAH.Hash()))
+		err = &ErrDataRootMismatch{
+			Height:       uint64(eh.Height()),
+			DataRoot:     eh.DataHash,
+			ComputedRoot: eh.DAH.Hash(),
+		}
+		log.Warnw("HIGH SEVERITY!!!", "err", err.Error())
```
Wondertan, line 157: All other checks are similarly sever, no?

Wondertan (approved):

vgonkivs (approved):

PR #2559, opened 2023-08-11
fix(share/eds): dagstore shard restore reflection workaround

walldiss:
Turns out `dagstore` has bug in its restore shard code. For each shard it attempts to create deep copy of registered type, but copies only basic struct fields of the mount type. In out case it means, that it copies a pointer.  That causes all shards to point to the same mount (same pointer).
Before `dagstore` is fixed to take into account possibility of pointers/interface (and any other reference type) in mount type struct fields, we will use direct copy of `mount.FileMount` instead of `mount.Mount` interface.

### DIFF
--- a/share/eds/store.go
+++ b/share/eds/store.go
@@ -75,7 +75,7 @@ func NewStore(basepath string, ds datastore.Batching) (*Store, error) {
 	}
 
 	r := mount.NewRegistry()
-	err = r.Register("fs", &inMemoryOnceMount{Mount: &mount.FileMount{}})
+	err = r.Register("fs", &inMemoryOnceMount{})
 	if err != nil {
 		return nil, fmt.Errorf("failed to register memory mount on the registry: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


renaynay (approved):

Wondertan:
Let's link the issue on the dagstore here

share/eds/store.go
```diff
@@ -565,7 +565,7 @@ type inMemoryOnceMount struct {
 	buf *bytes.Buffer
 
 	readOnce atomic.Bool
-	mount.Mount
```
Wondertan, line 568: The issue makes sense. When dagstore instantiates the registered mount it should know the exact type. It could derive the type from given `&mount.FileMount{}` but it's a bit advanced, thus the issue on DAGStore.

Wondertan (approved):

distractedm1nd:
https://github.com/filecoin-project/dagstore/issues/161

Linking

PR #2478, opened 2023-07-14
chore(deps): bump ipfs networking deps

walldiss:
## Overview
Main change is update of github.com/ipfs/go-libipfs , that is now located in github.com/ipfs/boxo 

Updates libs and their dependancies:
- github.com/libp2p/go-libp2p-kad-dht
- github.com/libp2p/go-libp2p-routing-helpers
 - github.com/ipfs/go-block-format 
- 	github.com/ipfs/go-blockservice 
 - 	github.com/ipfs/go-ipld-format
 - 	github.com/ipfs/go-merkledag 
 - 	github.com/ipld/go-car 
 - 	github.com/multiformats/go-multihash 

Supersedes:
https://github.com/celestiaorg/celestia-node/pull/2268
https://github.com/celestiaorg/celestia-node/pull/2378
https://github.com/celestiaorg/celestia-node/pull/2379
https://github.com/celestiaorg/celestia-node/pull/1957

PR #2560, opened 2023-08-11
fix: TestShareAvailable_DisconnectedFullNodes

distractedm1nd:
TestShareAvailable_DisconnectedFullNodes had multiple bugs which caused it to be a false positive. The following issues were fixed:

1. The parameters used now don't allow for reconstruction from a single subnetwork, as they did before. This has been verified with a simulation - for `s = 20`, and `k = 16`, `c` set to 32 gives a 99.5% chance of reconstruction, and 16 (single subnetwork) gives a 0.5% chance. Before, both values of `c` gave a near 100% chance for reconstruction.
2. Increased the timeout for the first reconstruction. There was a check that verified that reconstruction was not possible from a single subnetwork, but it only failed to reconstruct because the timeout was too short. Increasing the timeout with the original parameters verified this. 
3. Shape the network topology before the first reconstruction (for example, separating the LNs into subnets)
4. Calling final reconstruction in an error group. The full nodes are codependent on each other for either one to reconstruct. Calling them sequentially fails, because the first full node needs the second full node to have sampled in order to get shares from it. This was another signal that the full nodes could reconstruct from a single subnetwork, and it has now been fixed by allowing both FNs to sample concurrently.

### DIFF
--- a/share/availability/full/reconstruction_test.go
+++ b/share/availability/full/reconstruction_test.go
@@ -181,34 +181,14 @@ func TestShareAvailable_DisconnectedFullNodes(t *testing.T) {
 	light.DefaultSampleAmount = 20 // s
 	const (
 		origSquareSize = 16 // k
-		lightNodes     = 60 // c - total number of nodes on two subnetworks
+		lightNodes     = 32 // c - total number of nodes on two subnetworks
 	)
 
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*60)
 	defer cancel()
 
 	net := availability_test.NewTestDAGNet(ctx, t)
 	source, root := RandNode(net, origSquareSize)
-
-	// create two full nodes and ensure they are disconnected
-	full1 := Node(net)
-	full2 := Node(net)
-	net.Disconnect(full1.ID(), full2.ID())
-
-	// ensure fulls and source are not connected
-	// so that fulls take data from light nodes only
-	net.Disconnect(full1.ID(), source.ID())
-	net.Disconnect(full2.ID(), source.ID())
-
-	// start reconstruction for fulls that should fail
-	ctxErr, cancelErr := context.WithTimeout(ctx, eds.RetrieveQuadrantTimeout*8)
-	errg, errCtx := errgroup.WithContext(ctxErr)
-	errg.Go(func() error {
-		return full1.SharesAvailable(errCtx, root)
-	})
-	errg.Go(func() error {
-		return full2.SharesAvailable(errCtx, root)
-	})
 
 	// create light nodes and start sampling for them immediately
 	lights1, lights2 := make(
@@ -237,6 +217,16 @@ func TestShareAvailable_DisconnectedFullNodes(t *testing.T) {
 		}(i)
 	}
 
+	// create two full nodes and ensure they are disconnected
+	full1 := Node(net)
+	full2 := Node(net)
+	net.Disconnect(full1.ID(), full2.ID())
+
+	// ensure fulls and source are not connected
+	// so that fulls take data from light nodes only
+	net.Disconnect(full1.ID(), source.ID())
+	net.Disconnect(full2.ID(), source.ID())
+
 	// shape topology
 	for i := 0; i < len(lights1); i++ {
 		// ensure lights1 are only connected to source and full1
@@ -248,6 +238,16 @@ func TestShareAvailable_DisconnectedFullNodes(t *testing.T) {
 		net.Connect(lights2[i].ID(), full2.ID())
 		net.Disconnect(lights2[i].ID(), full1.ID())
 	}
+
+	// start reconstruction for fulls that should fail
+	ctxErr, cancelErr := context.WithTimeout(ctx, time.Second*5)
+	errg, errCtx := errgroup.WithContext(ctxErr)
+	errg.Go(func() error {
+		return full1.SharesAvailable(errCtx, root)
+	})
+	errg.Go(func() error {
+		return full2.SharesAvailable(errCtx, root)
+	})
 
 	// check that any of the fulls cannot reconstruct on their own
 	err := errg.Wait()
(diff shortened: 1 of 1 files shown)


renaynay (approved):

share/availability/full/reconstruction_test.go
```diff
@@ -181,35 +181,15 @@ func TestShareAvailable_DisconnectedFullNodes(t *testing.T) {
 	light.DefaultSampleAmount = 20 // s
 	const (
 		origSquareSize = 16 // k
-		lightNodes     = 60 // c - total number of nodes on two subnetworks
```
Wondertan, line 184: Did you look at git blame? I bet it was a debugging value accidentally committed.

distractedm1nd: This was the original value committed

Wondertan:
Thanks for this, @distractedm1nd! Before we merge it, I would like to make a deeper review of changes. Can this wait until Monday?
Cc @renaynay

share/availability/full/reconstruction_test.go
Wondertan: Ok, could you also double values for other reconstruction tests?

Wondertan (commented):
A few questions and recalling reasons behind why some things were done in that way:

1. I am trying to understand why 32. I presume you run some simulations using python script from Mus, but this is different from values on the [paper](https://arxiv.org/pdf/1809.09044.pdf). Table 1 states that for k=16 and s=20, c is 69 with 99% success. Why are these different? 

2. The quadrant timeout multiplied by the number of quadrants(4*2 axis) should work as it's the total time Retriever spends anyway. I didn't want tests to run more than necessary. The problem is in the init func at the top of the file that sets Retriever timeout to a turned out to be a low value. I think I didn't catch that due to false positive success.
	* Actually, the test failure at https://github.com/celestiaorg/celestia-node/pull/2478 shows that the timeout is enough with the latest optimizations in our dependencies so we may keep the timeouts minimal as they are rn.

3. I remember doing such an order purposefully. In other tests, we start reconstruction first and form the topology after in order to test that FNs are capable to reconstruct with LNs joining after the reconstruction starts. In practice, a FN may not have enough LNs to reconstruct, but more LNs would discover it, connect, and serve missing shares. However, while this is required for other tests, it's not required for this test scenario.

4. Makes total sense and is a great find!

share/availability/full/reconstruction_test.go
Wondertan: I checked the git blame, and it shows that it was committed with this value because in the ec3793fc I moved files over, but this look like a fully new file when it's not 😒. I didn't look for the original one.

distractedm1nd:
1. I know, - the problem here is that \gamma is IIUC referring to withholding attack. In our case, we only need `(k + 1)^2 + 1` shares to ensure full recovery. We should probably write tests including malicious withholding to back this up.
2. Okay, if the timeouts are minimal enough and aren't flaky then I will revert
3. Ok
4. <3

cc @musalbas

musalbas:
So you're saying you only need 69 light nodes if the block producer does an explicit block withholding attack (see below), but in normal/random cases where they aren't explicitly withholding data in the shape of a square, you only need 32?

image

distractedm1nd:
@musalbas That is my current understanding - running the simulation to collect at least (k + 1)^2 shares gives me the "correct" parameters - `k = 16` always fails, `k=18` always passes, which is consistent with the sim.

I am assuming that gamma represents some case that either happens very rarely, or if data is withheld in a clever way? Experimentally using completely random shares, we only need (k + 1)^2

musalbas:
Yes it assumes that the adversary is withholding the shares exactly in the shape of a square (see diagram above). In that case you actually need (2k)^2 - (k+1)^2 + 1 shares.

The amount of shares that light nodes need ultimately depends on what shares exactly are missing. Technically the minimum they need is not even (k+1)^2, but k^2.

walldiss:
Sorry for jumping in,  but I got super curious to simulation too. Actually previous script could result in less then 20 shares per node, since some shares can overlap within the node, if loop is limited by 20. I've build small util to play with params. Looks like 69 is the amount of light nodes to prevent withholding attack. Here are results:

```
amount of nodes, probability
58 0
59 0.001
60 0.006
61 0.039
62 0.093
63 0.253
64 0.473
65 0.676
66 0.828
67 0.926
68 0.974
69 0.993
70 0.998
``` 

```
func main() {
	k := 16
	shares := 20
	iter := 1000
	enough := (2*k)*(2*k) - (k+1)*(k+1) + 1
	lights := 71

	for l := 50; l < lights; l++ {
		count := 0
		for i := 0; i < iter; i++ {
			uniq := make(map[int]int)
			// loop over nodes
			for n := 0; n < l; n++ {
				// collect nodes local to elems
				elems := make(map[int]int)
				//loop until 20 uniq shares are stored
				for {
					r := rand.Intn(k * k * 4)
					elems[r]++
					uniq[r]++
					if len(elems) >= shares {
						break
					}
				}
			}
			if len(uniq) >= enough {
				count++
			}
		}

		probability := float64(count) / float64(iter)
		fmt.Println(l, probability)
	}
}
```

Wondertan:
I see a few options to proceed here:
* Keep the values from @distractedm1nd's simulations properly documenting why are they different from the paper
* Implement a more granular sampling test suite to reduce randomness yet keeping figures as close as possible described in the paper
* Extend the test suite with withholding attack via source. That is, make the source node serve only shares outside of the square shown in the diagram above. 


Personally, I prefer the 3rd option as it tests the worst case scenario, should be deterministic and 1 to 1 to the paper.

distractedm1nd:
They are not necessarily different from the paper, they're just a different part of the paper. I'll just add another test that withholds so we can test the third case, but let's do that in another PR?

musalbas:
Agree that the 3rd option makes most sense

Wondertan:
> I'll just add another test that withholds so we can test the third case, but let's do that in another PR?

Thinking about this more. All other reconstruction tests use similar ~69 figures, meaning they are incorrect. Also, I don't think with and without withholding attacks are separate test cases. It's more like the current reconstruction tests are not fully there yet and I am very glad that you identified issues with them. I think that we should just extend the current testsuite so that the source node always withholds data so that all the tests are fixed. This can definitely be in a separate PR and this one is good as it is, so approving.

Wondertan (approved):

Issue #2561, opened 2023-08-12
Account for secondary reserved namespace range


rootulp:
## Context

celestia-app is considering making a breaking change to reserve the last 256 namespaces (see https://github.com/celestiaorg/celestia-app/issues/2253). Relevant PR https://github.com/celestiaorg/celestia-app/pull/2257

## Problem

If https://github.com/celestiaorg/celestia-app/pull/2257 merges, we need to update celestia-node's namespace to account for the change. 

## Proposal

1. Rename [`MaxReservedNamespace`](https://github.com/celestiaorg/celestia-node/blob/87e9500120c27450e264ab4ffd59c87e92186194/share/namespace.go#L17C2-L17C2) to `MaxPrimaryReservedNamespace`
1. Rename [`ReservedPaddingNamespace`](https://github.com/celestiaorg/celestia-node/blob/87e9500120c27450e264ab4ffd59c87e92186194/share/namespace.go#L20C8-L20C8) to `PrimaryReservedPaddingNamespace`
1. Add `MinSecondaryReservedNamespace` constant to [this list](https://github.com/celestiaorg/celestia-node/blob/87e9500120c27450e264ab4ffd59c87e92186194/share/namespace.go#L15-L24)
1. [`ValidateForBlob`](https://github.com/celestiaorg/celestia-node/blob/87e9500120c27450e264ab4ffd59c87e92186194/share/namespace.go#L122-L131) should account for the secondary reserved namespace range.
1. Add unit tests

_Originally posted by @rootulp in https://github.com/celestiaorg/celestia-app/issues/2257#issuecomment-1675981337_

Wondertan:
@renaynay, just needs https://github.com/celestiaorg/celestia-node/pull/2597 to be merged

PR #2562, opened 2023-08-12
feat: account for secondary reserved namespace range

rootulp:
Closes https://github.com/celestiaorg/celestia-node/issues/2561
Blocked on https://github.com/celestiaorg/celestia-app/pull/2257


share/namespace.go
```diff
@@ -106,6 +107,7 @@ func (n Namespace) Validate() error {
 }
 
 // ValidateForData checks if the Namespace is of real/useful data.
+// TODO: it's not clear what "real/useful" data means.
```
rootulp, line 110: @Wondertan @vgonkivs can you please help me understand what "real/useful" data means? Do we need `ValidateForData` or is `ValidateForBlob` sufficient?

rootulp: If we need to maintain this function, it may need to be updated in this PR.

Wondertan: Not parity data

PR #2574, opened 2023-08-16
fix(share/ipld): Remove proofs from leaked ctx

walldiss:
Inside tee-getter ProofAdder is added to ctx to collect proofs and allow eds store to read them, without extra recompute. Ctx provided to store.Put method, will be used for register shard, that in some rare cases takes long time and returns, before result returned from operation. The provided context is still leaked to async operation, and it prevents golang GC from collecting memory allocated for ProofAdder, because there is still possibility for code to reach it. 

Solution
Dagstore register shard operation does not need to access to proofs, so the associated memory could be reclaimed right after proofs are extracted from ctx. For convenience it will happen after put operation 

Solves first point of https://github.com/celestiaorg/celestia-node/issues/2582


share/ipld/nmt_adder.go
```diff
@@ -151,6 +159,18 @@ func (a *ProofsAdder) VisitFn() nmt.NodeVisitorFn {
...
+// Purge removed proofs from ProofsAdder allowing GC to collect the memory
+func (a *ProofsAdder) Purge() {
+	if a == nil {
+		return
+	}
+
+	a.lock.RLock()
+	defer a.lock.RUnlock()
+
+	a.proofs = nil
```
Wondertan, line 171: nit: might be worth using `map.Clear`(or an old way to clear map) to help GC a bit

Wondertan (dismissed):
I assume you've tested this and seen stable RAM usage on Graphana

share/ipld/nmt_adder.go
walldiss: Not sure what is map.Clear, it is something I'm not aware of. The best way to help GC is to remove links to object, so it does exactly that 🙂

Wondertan: By helping GC I mean freeing memory in-place rather than deferring this to GC

walldiss: Is there a way to forcefully free up space for reference type? Asking because I'm not aware of it. Deleting items from map for example, will not remove the map, it will just mark items as soft-deleted

Wondertan: Doesnt answer the question, but related https://stackoverflow.com/a/70808604

walldiss:
>The fact that proofs are leaking is a hint that the context is leaking. Because if the context is cleared up, the proofs would be cleared up as well. WDYT?

I agree, Thats why PR is named for issue of leaked ctx, but not just proofs "Remove proofs from leaked ctx". Provided solution is a hotfix for latest release to prevent leak. Proper investigation of leaked ctx will take some time, since it is passed into many complex components, like ipld, bitswap, dagstore, badger. I'll create an issue to investigate where the ctx is leaked.

share/ipld/nmt_adder.go
```diff
@@ -151,6 +159,18 @@ func (a *ProofsAdder) VisitFn() nmt.NodeVisitorFn {
...
 }
 
+// Purge removed proofs from ProofsAdder allowing GC to collect the memory
+func (a *ProofsAdder) Purge() {
+	if a == nil {
+		return
+	}
+
+	a.lock.RLock()
+	defer a.lock.RUnlock()
```
renaynay, line 169: this should not be a read lock

Wondertan: There is also not so much value in read lock here. We write it once and use it everywhere. No simultaneous writes are possible. Also, visitFn is called sequentially by NMT.

renaynay (approved):

Wondertan (approved):

PR #2577, opened 2023-08-17
feat(nodebuilder/state): Provide stubbed state module if a core endpoint not provided

renaynay:
This is a pre-requisite for #2511 

It makes a check in CoreAccessor constructor to see if a core endpoint was provided, and returns nil CoreAccessor if not. A stubbed state module will then be provided if no core endpoint was provided so that errors are more readable. 

Previously, node start logic relied on the fact that the grpc Dial inside CoreAccessor.Start was non-blocking so it could silently fail under the hood and any calls made on state Module would return errors from the inability to reach the address that is the default for the core config (which was confusing).

### DIFF
--- a/nodebuilder/core/config.go
+++ b/nodebuilder/core/config.go
@@ -18,14 +18,18 @@ type Config struct {
 // node's connection to a Celestia-Core endpoint.
 func DefaultConfig() Config {
 	return Config{
-		IP:       "0.0.0.0",
-		RPCPort:  "0",
-		GRPCPort: "0",
+		IP:       "",
+		RPCPort:  "",
+		GRPCPort: "",
 	}
 }
 
 // Validate performs basic validation of the config.
 func (cfg *Config) Validate() error {
+	if !cfg.EndpointConfigured() {
+		return nil
+	}
+
 	ip, err := utils.ValidateAddr(cfg.IP)
 	if err != nil {
 		return err
@@ -43,8 +47,7 @@ func (cfg *Config) Validate() error {
 }
 
 // EndpointConfigured returns whether a core endpoint has been set
-// on the config.
+// on the config (true if set).
 func (cfg *Config) EndpointConfigured() bool {
-	defaultCfg := DefaultConfig()
-	return !(cfg.IP == defaultCfg.IP && (cfg.GRPCPort == defaultCfg.GRPCPort || cfg.RPCPort == defaultCfg.RPCPort))
+	return cfg.IP != ""
 }
--- a/nodebuilder/core/config.go
+++ b/nodebuilder/core/config.go
@@ -41,3 +41,10 @@ func (cfg *Config) Validate() error {
 	}
 	return nil
 }
+
+// EndpointConfigured returns whether a core endpoint has been set
+// on the config.
+func (cfg *Config) EndpointConfigured() bool {
+	defaultCfg := DefaultConfig()
+	return !(cfg.IP == defaultCfg.IP && (cfg.GRPCPort == defaultCfg.GRPCPort || cfg.RPCPort == defaultCfg.RPCPort))
+}
--- a/nodebuilder/core/config.go
+++ b/nodebuilder/core/config.go
@@ -26,7 +26,7 @@ func DefaultConfig() Config {
 
 // Validate performs basic validation of the config.
 func (cfg *Config) Validate() error {
-	if !cfg.EndpointConfigured() {
+	if !cfg.IsEndpointConfigured() {
 		return nil
 	}
 
@@ -46,8 +46,8 @@ func (cfg *Config) Validate() error {
 	return nil
 }
 
-// EndpointConfigured returns whether a core endpoint has been set
+// IsEndpointConfigured returns whether a core endpoint has been set
 // on the config (true if set).
-func (cfg *Config) EndpointConfigured() bool {
+func (cfg *Config) IsEndpointConfigured() bool {
 	return cfg.IP != ""
 }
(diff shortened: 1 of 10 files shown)


nodebuilder/node_test.go
```diff
@@ -43,14 +43,8 @@ func TestLifecycle(t *testing.T) {
...
 			require.NoError(t, err)
 
-			// ensure the state service is running
-			require.False(t, node.StateServ.IsStopped(ctx))
-
 			err = node.Stop(ctx)
 			require.NoError(t, err)
-
-			// ensure the state service is stopped
-			require.True(t, node.StateServ.IsStopped(ctx))
```
renaynay, line 53: these checks were useless

nodebuilder/core/config.go
```diff
@@ -41,3 +41,10 @@ func (cfg *Config) Validate() error {
 	}
 	return nil
 }
+
+// EndpointConfigured returns whether a core endpoint has been set
+// on the config.
+func (cfg *Config) EndpointConfigured() bool {
```
renaynay, line 47: I welcome a better name if anyone has ideas

nodebuilder/core/config.go
```diff
@@ -41,3 +41,10 @@ func (cfg *Config) Validate() error {
...
+	defaultCfg := DefaultConfig()
+	return !(cfg.IP == defaultCfg.IP && (cfg.GRPCPort == defaultCfg.GRPCPort || cfg.RPCPort == defaultCfg.RPCPort))
+}
```
Wondertan, line 50: The defaults are zeros. Do you remember why we have zeroes isntead of empty strings? We should consider emptying them to avoid this monstrosity.

nodebuilder/state/core.go
```diff
@@ -20,6 +20,11 @@ func coreAccessor(
 	sync *sync.Syncer[*header.ExtendedHeader],
 	fraudServ libfraud.Service,
 ) (*state.CoreAccessor, *modfraud.ServiceBreaker[*state.CoreAccessor]) {
+	if !corecfg.EndpointConfigured() {
+		log.Info("No core endpoint provided, running node without state access")
+		return nil, nil
+	}
```
Wondertan, line 26: We have a nice `fxutil.ProvideIf`. Why not it?

nodebuilder/state/core.go
```diff
@@ -20,6 +20,11 @@ func coreAccessor(
 	sync *sync.Syncer[*header.ExtendedHeader],
 	fraudServ libfraud.Service,
 ) (*state.CoreAccessor, *modfraud.ServiceBreaker[*state.CoreAccessor]) {
+	if !corecfg.EndpointConfigured() {
+		log.Info("No core endpoint provided, running node without state access")
```
Wondertan, line 24: I think we should start hardcodin our Core node bootstrappers so that everyone has state access by default. Otherwise, we are not really usable with just a few CLIs.

Wondertan: https://github.com/celestiaorg/celestia-node/issues/2579

nodebuilder/state/core.go
renaynay: less LOCs now: https://github.com/celestiaorg/celestia-node/pull/2577/commits/1b2435685721246bc28d3380da8453d9b28fb745

nodebuilder/core/config.go
renaynay: I think the defaults were 0s precisely because we did not provide a stubbed accessor and still attempted to dial the empty addr that was passed inside core_accessor.

Changed to empty strings now. https://github.com/celestiaorg/celestia-node/pull/2577/commits/1b2435685721246bc28d3380da8453d9b28fb745

nodebuilder/settings.go
```diff
@@ -74,7 +74,12 @@ func WithMetrics(metricOpts []otlpmetrichttp.Option, nodeType node.Type) fx.Opti
 	baseComponents := fx.Options(
 		fx.Supply(metricOpts),
 		fx.Invoke(initializeMetrics),
-		fx.Invoke(state.WithMetrics),
+		fx.Invoke(func(ca *state.CoreAccessor) {
+			if ca == nil {
+				return
+			}
+			state.WithMetrics(ca)
+		}),
```
Wondertan, line 82: Cane be InvokeIf

nodebuilder/state/core.go
```diff
@@ -19,10 +19,15 @@ func coreAccessor(
 	signer *apptypes.KeyringSigner,
 	sync *sync.Syncer[*header.ExtendedHeader],
 	fraudServ libfraud.Service,
-) (*state.CoreAccessor, *modfraud.ServiceBreaker[*state.CoreAccessor]) {
+) (*state.CoreAccessor, Module, *modfraud.ServiceBreaker[*state.CoreAccessor]) {
+	if !corecfg.EndpointConfigured() {
+		log.Info("No core endpoint provided, running node without state access")
+		return nil, &stubbedStateModule{}, nil
+	}
```
Wondertan, line 26: Can be two ProvideIfs, providing necessary coreaccessor basing on EndpointConfigured

Wondertan (approved):
Optional simplification nits

nodebuilder/core/config.go
MSevey: (☞ﾟヮﾟ)☞   `isEndpointConfigured`

distractedm1nd (approved):

distractedm1nd (approved):

nodebuilder/core/config.go
walldiss: How about `hasEndpoint`?

walldiss (approved):

PR #2581, opened 2023-08-18
chore!: update celestia-app to v1.0.0-rc12

evan-forbes:
## Overview

bumps app to v1.0.0-rc12

## Checklist


nodebuilder/tests/nd_test.go
```diff
@@ -48,8 +49,12 @@ func TestShrexNDFromLights(t *testing.T) {
...
 	require.NoError(t, <-fillDn)
 
-	// first 2 blocks are not filled with data
-	for i := 3; i < blocks; i++ {
+	// first 15 blocks are not filled with data
+	//
+	// TODO: we need to stop guessing
+	// the block that actually has transactions. We can get this data from the
+	// response returned by FillBlock.
+	for i := 16; i < blocks; i++ {
```
evan-forbes, line 56: I think we should change the channel returned by `FillBlocks` to return a channel of `sdk.TxResult` instead of error. This way, we can actually see which blocks the txs get included in and are not making crude trial and error estimates on which blocks have data in them.

If needed, we can make it return some summary of what was submitted as well if that would be useful.

state/core_access.go
```diff
@@ -178,6 +178,7 @@ func (ca *CoreAccessor) SubmitPayForBlob(
 		ctx,
 		ca.signer,
 		ca.coreConn,
+		sdktx.BroadcastMode_BROADCAST_MODE_BLOCK,
```
evan-forbes, line 181: kept this the same for now as I'm not sure how you all wanted to handle this, but we can change this to sync now

Wondertan: Let's do Sync please as you suggested on that issue

evan-forbes: originally I didn't change that as it seemed like a relatively breaking change as users would now have to have some other mechanism to know when their tx got included. 

Do you think that makes sense do that in a different PR? Keep this one just for bumping app, and then another follow up to change how light clients figure out if their tx was accepted or rejected. Happy to help out with this as I'm pretty familar with what can occur and what each response means

Wondertan: > Do you think that makes sense do that in a different PR? 

Makes sense

share/namespace.go
```diff
@@ -124,7 +130,10 @@ func (n Namespace) ValidateForBlob() error {
 	if err := n.ValidateForData(); err != nil {
 		return err
 	}
-	if bytes.Compare(n, MaxReservedNamespace) < 1 {
+	if bytes.Compare(n, MaxPrimaryReservedNamespace) < 1 {
+		return fmt.Errorf("invalid blob namespace(%s): reserved namespaces are forbidden", n)
+	}
+	if bytes.Compare(n, MinSecondaryReservedNamespace) > -1 {
```
rootulp, line 136: [optional][no change needed] this could also be https://github.com/celestiaorg/celestia-node/pull/2562/files#diff-46c85ee9c611a9b9c56eb104a6dca5bcbc6b3961683cdf0c941704b18496084dR129 so that if celestia-app makes changes to namespace validation, celestia-node gets those changes.

state/core_access.go
renaynay: I think the default SubmitPFB method should be blocking and we can add an option to make it async at some point but not in this PR pls

renaynay (approved):

nodebuilder/tests/nd_test.go
renaynay: This change may collide with some changes @walldiss is working on to make tests with filled blocks more deterministic

state/core_access.go
Wondertan: Block is getting deprecated, so we should migrate from it

Wondertan (approved):

share/namespace.go
```diff
@@ -14,13 +14,19 @@ const NamespaceSize = appns.NamespaceSize
...
+	// MaxPrimaryReservedNamespace is the highest primary reserved namespace.
+	// Namespaces lower than this are reserved for protocol use.
+	MaxPrimaryReservedNamespace = Namespace(appns.MaxPrimaryReservedNamespace.Bytes())
+	// MinSecondaryReservedNamespace is the lowest secondary reserved namespace
+	// reserved for protocol use. Namespaces higher than this are reserved for
+	// protocol use.
+	MinSecondaryReservedNamespace = Namespace(appns.MinSecondaryReservedNamespace.Bytes())
+	ParitySharesNamespace         = Namespace(appns.ParitySharesNamespace.Bytes())
+	TailPaddingNamespace          = Namespace(appns.TailPaddingNamespace.Bytes())
+	ReservedPaddingNamespace      = Namespace(appns.PrimaryReservedPaddingNamespace.Bytes())
```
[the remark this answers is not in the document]
renaynay: let's do follow-up

PR #2585, opened 2023-08-19
feat(share/eds): add EdsMaker

vgonkivs:
## Overview
Created an EdsMaker that is a type alias of `extendBlock` that we had previously. Having such improvement will give us flexibility in testing different scenarios via swamp(we can now replace the default MakeExtendedDataSquare). Also, it makes code consistent IMO(we already have header maker)
## Checklist


nodebuilder/core/opts.go
```diff
@@ -17,3 +17,8 @@ func WithClient(client core.Client) fx.Option {
 func WithHeaderConstructFn(construct header.ConstructFn) fx.Option {
 	return fx.Replace(construct)
 }
+
+// WithEdsConstructFn sets custom func that creates extended data square
+func WithEdsConstructFn(construct header.ConstructFn) fx.Option {
```
Wondertan, line 22:
```suggestion
func WithEdsConstructFn(construct eds.ConstructFn) fx.Option {
```

share/eds/eds.go
```diff
@@ -273,3 +277,59 @@ func ReadEDS(ctx context.Context, r io.Reader, root share.DataHash) (eds *rsmt2d
...
+	if eds == nil {
+		return nil
+	}
+	err := store.Put(ctx, hash, eds)
+	if errors.Is(err, dagstore.ErrShardExists) {
+		// block with given root already exists, return nil
+		return nil
+	}
+	return err
+}
```
Wondertan, line 335: Why do you think this deserves a helper?

share/eds/eds.go
```diff
@@ -273,3 +277,59 @@ func ReadEDS(ctx context.Context, r io.Reader, root share.DataHash) (eds *rsmt2d
...
+// ExtendedDataSquare (EDS). If the data is empty,
+// nil is returned in place of the eds.
+func MakeExtendedDataSquare(
+	data [][]byte,
+	appVersion uint64,
+	squareSizeBounds int,
+	options ...nmt.Option,
+) (*rsmt2d.ExtendedDataSquare, error) {
+	if len(data) == 0 {
+		return nil, nil
```
Wondertan, line 299: Returning two nils is an anti-pattern. We did that previously in the private code which was okaish. Now that we extracted this as public API func, we should change that.

Wondertan: This case does not happen anymore also, because you always check for `IsEmptyBlock` outside of this func

core/listener.go
```diff
@@ -155,19 +162,32 @@ func (cl *Listener) handleNewSignedBlock(ctx context.Context, b types.EventDataS
...
+
+	if !app.IsEmptyBlock(b.Data, b.Header.Version.App) {
+		dataSq, err = cl.edsConstruct(
+			b.Data.Txs.ToSliceOfBytes(),
+			b.Header.Version.App,
+			appconsts.SquareSizeUpperBound(b.Header.Version.App),
+			nmt.NodeVisitor(adder.VisitFn()),
+		)
+		if err != nil {
+			return fmt.Errorf("extending block data: %w", err)
```
Wondertan, line 178: might be worth panicking here, as we do for header construct.

share/eds/eds.go
vgonkivs: Currently yes, but app code relies on the version. What if future version will not check data length. It seems that this is impossible but imo this is a good pattern to check length of the slice which you are going to use. 

Regarding double nil, I agree. Let's then creat an empty Eds?

share/eds/eds.go
vgonkivs: You mean store Eds? It seems to me a really good function that stores  Eds if it does not exist in the store. We can re use in other places if needed. I did not want to insert this functionality inside edsMaker because I believe MakeEDS should take care of storing it - this is an absolutely another type of logic, imo

nodebuilder/share/module.go
```diff
@@ -146,6 +146,9 @@ func ConstructModule(tp node.Type, cfg *Config, options ...fx.Option) fx.Option
 			"share",
 			baseComponents,
 			bridgeAndFullComponents,
+			fx.Provide(func() eds.ConstructFn {
```
distractedm1nd, line 149: does fxutil.ProvideAs work here?

core/exchange.go
```diff
@@ -19,20 +20,23 @@ import (
 const concurrencyLimit = 4
 
 type Exchange struct {
-	fetcher   *BlockFetcher
-	store     *eds.Store
-	construct header.ConstructFn
+	fetcher         *BlockFetcher
+	store           *eds.Store
+	headerConstruct header.ConstructFn
+	edsConstruct    eds.ConstructFn
```
distractedm1nd, line 26: constructHeader/constructEDS make more sense from an english perspective, but optional

nodebuilder/share/module.go
vgonkivs: I've tried but it did not work 😅

core/exchange.go
vgonkivs: 🚔

core/listener.go
```diff
@@ -32,8 +35,10 @@ var tracer = otel.Tracer("core/listener")
 type Listener struct {
 	fetcher *BlockFetcher
 
-	construct header.ConstructFn
-	store     *eds.Store
+	headerConstruct header.ConstructFn
+	edsConstruct    eds.ConstructFn
```
walldiss, line 39: Looks like those are configurable to be allow mocking for tests. I wonder if at this point It would be worth to consider making a single function Interface instead of exact types. 

Just thinking, if we can make it a bit more go-way later. lets not do it in this PR 🙂

share/eds/eds.go
walldiss: Has check is also happening inside store.Put() call now. So you might want to replace this func with `store.Put()` everywhere.

core/listener.go
```diff
@@ -155,19 +162,32 @@ func (cl *Listener) handleNewSignedBlock(ctx context.Context, b types.EventDataS
...
+
+	if !app.IsEmptyBlock(b.Data, b.Header.Version.App) {
+		dataSq, err = cl.edsConstruct(
+			b.Data.Txs.ToSliceOfBytes(),
+			b.Header.Version.App,
+			appconsts.SquareSizeUpperBound(b.Header.Version.App),
+			nmt.NodeVisitor(adder.VisitFn()),
+		)
+		if err != nil {
+			panic(fmt.Errorf("extending block data: %w", err))
```
walldiss, line 178: Is it really necessary to crash here? Why not just return an error as it used to be?

share/eds/eds.go
renaynay: let's please use the app function to check if the block data really is empty and then can return `EmptyRoot` ? 

```
	if app.IsEmptyBlock(data, appVersion) {
		return share.EmptyExtendedDataSquare(), nil
	}

	dataSquare, err := square.Construct(data.Txs.ToSliceOfBytes(), appVersion, squareSizeBounds)
	if err != nil {
		return nil, err
	}

```

That would require changing func sig to take tendermint `types.Data`

Might be import cycle though so see if we can move empty EDS def to eds pkg?

share/eds/eds.go
```diff
@@ -273,3 +277,59 @@ func ReadEDS(ctx context.Context, r io.Reader, root share.DataHash) (eds *rsmt2d
...
+	if !shares.IsPowerOfTwo(len(s)) {
+		return nil, fmt.Errorf("number of shares is not a power of 2: got %d", len(s))
+	}
+	// here we construct a tree
+	// Note: uses the nmt wrapper to construct the tree.
+	squareSize := square.Size(len(s))
+	return rsmt2d.ComputeExtendedDataSquare(s,
+		appconsts.DefaultCodec(),
+		wrapper.NewConstructor(uint64(squareSize),
+			options...))
```
distractedm1nd, line 321: this has weird formatting, im surprised go linter accepts it, could you clean it up a bit?

share/eds/eds.go
distractedm1nd: I believe another reason for this though was the eds nil check which was a bug at some point during the initial impl for bridge nodes, correct @renaynay ?

distractedm1nd (changes requested):

vgonkivs:
not relevant anymore

PR #2591, opened 2023-08-21
fix(share/eds): scale up inverted_index badger levels

walldiss:
Current instance of badger uses default value (7) for amount of used levels, that allows badger to grow up to 1.1TiB. This limit could be reaching within 1 week of 128 ods size blocks, so we need to bump the value to at least 8 levels (11 TiB).


renaynay:
Can this be re-configured or would it be breaking?

vgonkivs (approved):

walldiss:
9 levels is harder for compaction there is reasonable doubts we would reach 11TB inverted index size before getting rid of it

walldiss:
> Can this be re-configured or would it be breaking?

It is not breaking, but could cause node to do long compaction while preventing all store related operation. Compaction could take 1-2 hours on badger size of 1.1 TB. Should be less for smaller sizes, but I don't have exact numbers.

Wondertan (approved):

PR #2597, opened 2023-08-22
refactor: address namespace follow-ups from celestia-app v1.0.0-rc12 bump

rootulp:
Addresses two follow-ups from https://github.com/celestiaorg/celestia-node/pull/2581
Closes https://github.com/celestiaorg/celestia-node/issues/2561

### DIFF
--- a/share/namespace.go
+++ b/share/namespace.go
@@ -20,13 +20,13 @@ var (
 	// MinSecondaryReservedNamespace is the lowest secondary reserved namespace
 	// reserved for protocol use. Namespaces higher than this are reserved for
 	// protocol use.
-	MinSecondaryReservedNamespace = Namespace(appns.MinSecondaryReservedNamespace.Bytes())
-	ParitySharesNamespace         = Namespace(appns.ParitySharesNamespace.Bytes())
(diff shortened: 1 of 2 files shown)


share/namespace.go
```diff
@@ -130,10 +130,7 @@ func (n Namespace) ValidateForBlob() error {
 	if err := n.ValidateForData(); err != nil {
 		return err
 	}
-	if bytes.Compare(n, MaxPrimaryReservedNamespace) < 1 {
-		return fmt.Errorf("invalid blob namespace(%s): reserved namespaces are forbidden", n)
-	}
-	if bytes.Compare(n, MinSecondaryReservedNamespace) > -1 {
+	if n.ToAppNamespace().IsReserved() {
```
Wondertan, line 133: Let's avoid conversions and copy code for now.

Wondertan (commented):
Thank you!

share/namespace.go
rootulp: It seems like a burden to maintain the same validation logic in two separate repos but defer to maintainers.

Resolved via f0f02f1a47e67f0c669291a775730fdf8750ef05

Wondertan (approved):

PR #2607, opened 2023-08-24
feat(cmd/rpc): add submitPFB cmd

vgonkivs:
## Overview

added a submitPFB cmd support for the rpc client
## Checklist

### DIFF
--- a/cmd/celestia/state.go
+++ b/cmd/celestia/state.go
@@ -10,16 +10,6 @@ import (
 	"github.com/celestiaorg/celestia-node/blob"
 )
 
-func init() {
-	stateCmd.AddCommand(submitPFB)
-}
-
-var stateCmd = &cobra.Command{
-	Use:   "state [command]",
-	Short: "Allows to interact with the State Service via JSON-RPC",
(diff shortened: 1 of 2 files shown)


distractedm1nd (approved):
I would rather just fix the existing command until the whole module is implemented, but I understand how much better this is for docs

cmd/celestia/state.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"github.com/cosmos/cosmos-sdk/types"
+	"github.com/spf13/cobra"
+
+	"github.com/celestiaorg/celestia-node/blob"
+)
+
+var submitPFB = &cobra.Command{
+	Use:   "submitPFB [namespace, data, fee, gasLim]",
+	Short: "Allows to submit pfbs",
```
renaynay, line 15:
```suggestion
	Short: "Allows to submit PFBs",
```

renaynay (changes requested):
@vgonkivs let's make this part of feature branch please ?

also 

`submitPFB` --> `submit-pfb` if we're gonna follow kebab case convention

PR #2612, opened 2023-08-25
feat(share/eds): Rework accessor cache

walldiss:
**Motivation** 
Previously, all accessors went into cache. They were closed only when they were evicted from it and there was no way to manually close one. Now, when we need to cache accessors only in some cases and in some not to, this approach will not work. Enabling control of accessor close required major cache rework.

New cache features:
- Old cache didn't allow to close and release of an accessor. Now, Instead of returning just a reader/blockstore, cache returns object that composed of builder and `io.Closer` (Accessor interface). This allows caller to have an ability to close and release accessor once it is done with reading from it. 
- Items stored in cache also provide io.Closer`, but it is noop. Lifecycle of cached items is controlled by cache itself. For now objects are released right after they are evicted from cache, but it could be changed to smarter release based on current readers count. Will be implemented as separate issue.
- Lazy blockstore builder. Old cache created blockstore every time accessor was acquired, even when only reader was used and blockstore was not needed. Creating blockstore is costly operation that involves building index, so this was reworked to create blockstore, only when it is requested and store it for subsequent requests. 
- adds cache unit tests.
- adds metrics for cache methods to track cache hits/miss.
- general cleanup of unused methods/logic.

It will be easier for reviewer to start from `store.go` and `blockstore.go` and then go into new `cache` pkg


share/eds/accessor_cache.go
```diff
@@ -39,7 +102,7 @@ type blockstoreCache struct {
 	metrics *cacheMetrics
 }
 
-func newBlockstoreCache(cacheSize int) (*blockstoreCache, error) {
+func newBlockstoreCache(name string, cacheSize int) (*blockstoreCache, error) {
```
distractedm1nd, line 105: linter complains that name is unused

share/eds/accessor_cache.go
```diff
@@ -15,10 +15,71 @@ import (
...
+	// getOrLoad attempts to get an item from the cache and, if not found, invokes
+	// the provided loader function to load it into the cache.
+	getOrLoad(ctx context.Context, key shard.Key, loader func(context.Context, shard.Key) (*dagstore.ShardAccessor, error)) (*accessorWithBlockstore, error)
+
+	// enableMetrics enables metrics in cache
+	enableMetrics() error
+}
+
+// multiCache represents a cache that looks into multiple caches one by one.
+type multiCache struct {
```
distractedm1nd, line 35: Is the point of this only for metrics? I think its a bit overkill unless we are planning on splitting the cache even further - I would much rather simply have two metrics references saved to the store - its less abstract and cleaner

share/eds/store.go
```diff
@@ -326,11 +347,21 @@ func (s *Store) carBlockstore(
 	root share.DataHash,
 ) (dagstore.ReadBlockstore, error) {
 	key := shard.KeyFromString(root.String())
-	accessor, err := s.getCachedAccessor(ctx, key)
+	accessor, err := s.cache.get(key)
+	if err == nil {
+		return accessor.bs, nil
+	}
+	// if accessor not found in cache, create new one from dagstore
```
distractedm1nd, line 354: please add to comment that we don't save this accessor to the cache and why

share/eds/accessor_cache.go
distractedm1nd: Hmm... on second thought, this is nice because it hides the multiple lookups higher up. But I would still say we should not make it abstract, and just have two caches saved to this multicache instead of an array.

distractedm1nd (commented):
First lookthrough, looking good so far

share/eds/blockstore.go
```diff
@@ -148,11 +160,30 @@ func (bs *blockstore) getReadOnlyBlockstore(ctx context.Context, cid cid.Cid) (d
 		return nil, fmt.Errorf("failed to find shards containing multihash: %w", err)
 	}
 
+	// a share can exist in multiple EDSes, check cache to contain any of accessors containing shard
```
distractedm1nd, line 163: this behavior is incorrect - we only store a single value, since we do not use dagstore's implementation. So we can just use the first element returned. This is related to the badger bloat issue that we discussed in the pruning call.

share/eds/blockstore.go
```diff
@@ -32,14 +35,17 @@ var (
 // implementation to allow for the blockstore operations to be routed to the underlying stores.
 type blockstore struct {
 	store *Store
-	cache *blockstoreCache
```
distractedm1nd, line 35: thank you for removing this and accessing it via EDSStore directly

share/eds/blockstore.go
```diff
@@ -72,15 +83,17 @@ func (bs *blockstore) Get(ctx context.Context, cid cid.Cid) (blocks.Block, error
...
+	log.Debugf("failed to get blockstore for cid %s: %s", cid, err)
+	return nil, err
 }
 
 func (bs *blockstore) GetSize(ctx context.Context, cid cid.Cid) (int, error) {
 	blockstr, err := bs.getReadOnlyBlockstore(ctx, cid)
+	if err == nil {
+		defer closeAndLog("blockstore", blockstr)
+		return blockstr.GetSize(ctx, cid)
+	}
```
distractedm1nd, line 96:
```suggestion
	}

```

share/eds/store.go
```diff
@@ -255,17 +258,35 @@ func (s *Store) put(ctx context.Context, root share.DataHash, square *rsmt2d.Ext
...
+
+	// accessor returned in result will be nil, so shard needs to be acquired first, to become
+	// available in cache. It might take some time and result should not affect put operation, so do it
+	// in goroutine
+	//TODO: Ideally only recent blocks should be put in cache, but there is no way right now to check
+	// such condition.
+	go func() {
+		ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+		defer cancel()
+		_, err := s.cache.GetOrLoad(ctx, result.Key, s.getAccessor)
```
distractedm1nd, line 282: would it make sense to add a Load method to the cache? This way we can avoid an expected missed cache hit here

share/eds/store.go
```diff
@@ -255,17 +258,35 @@ func (s *Store) put(ctx context.Context, root share.DataHash, square *rsmt2d.Ext
...
 	}
+
+	if result.Error != nil {
+		return fmt.Errorf("failed to register shard: %w", result.Error)
+	}
+
+	// accessor returned in result will be nil, so shard needs to be acquired first, to become
+	// available in cache. It might take some time and result should not affect put operation, so do it
+	// in goroutine
+	//TODO: Ideally only recent blocks should be put in cache, but there is no way right now to check
```
distractedm1nd, line 277: after pruning, only blocks inside the recency window will be put anyways. So I agree that this is not a big issue. It only happens during initial sync anyways

share/eds/store.go
```diff
@@ -312,13 +333,21 @@ func (s *Store) GetCAR(ctx context.Context, root share.DataHash) (io.Reader, err
...
-	key := root.String()
-	accessor, err := s.getCachedAccessor(ctx, shard.KeyFromString(key))
+func (s *Store) getCAR(ctx context.Context, root share.DataHash) (io.ReadCloser, error) {
+	key := shard.KeyFromString(root.String())
+	accessor, err := s.cache.Get(key)
+	if err == nil {
+		return newReadCloser(accessor), nil
+	}
+	// If the accessor is not found in the cache, create a new one from dagstore. We don't put accessor
+	// to the cache here, because getCAR is used by shrex.eds. There is a lower probability, compared
```
distractedm1nd, line 343:
```suggestion
	// to the cache here, because getCAR is used by shrex-eds. There is a lower probability, compared
```

share/eds/store.go
```diff
@@ -312,13 +333,21 @@ func (s *Store) GetCAR(ctx context.Context, root share.DataHash) (io.Reader, err
...
+	// to other cache put triggers, that the same block to be requested again.
```
distractedm1nd, line 344:
```suggestion
	// to the cache put trigger, that the same block will be requested again soon.
```

share/eds/utils.go
```diff
@@ -0,0 +1,9 @@
+package eds
+
+import "io"
+
+func closeAndLog(name string, closer io.Closer) {
+	if err := closer.Close(); err != nil {
+		log.Warnw("closing "+name, "err", err)
```
distractedm1nd, line 7: will this not spam logs?

distractedm1nd (changes requested):
This is looking really good, and I noticed a ton of small cleanups that I really appreciate <3

distractedm1nd (changes requested):
This is looking really good, and I noticed a ton of small cleanups that I really appreciate <3

share/eds/cache/accessor_cache.go
```diff
@@ -0,0 +1,191 @@
...
+	provider, err := loader(ctx, key)
+	if err != nil {
+		return nil, fmt.Errorf("unable to get accessor: %w", err)
+	}
+
+	abs := &accessorWithBlockstore{
+		shardAccessor: provider,
+	}
+
+	// create new accessor first to inc ref count in it, so it could not get evicted from inner cache
```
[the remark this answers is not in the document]
walldiss: Yeah, it is an artefact of rebasing from ref counting PR. It explains why order of operation is important here. Inner cache is lru cache. I'm open for better phrasing if you have any suggestions :)

share/eds/store.go
```diff
@@ -460,14 +468,13 @@ func (s *Store) Remove(ctx context.Context, root share.DataHash) error {
...
 func (s *Store) remove(ctx context.Context, root share.DataHash) (err error) {
-	key := root.String()
-
-	// Remove from accessor cache, so that existing readers are closed and
-	// DestroyShard can be executed.
-	s.cache.Remove(shard.KeyFromString(key))
-
+	key := shard.KeyFromString(root.String())
+	// remove open links to accessor from cache
+	if err := s.cache.Remove(key); err != nil {
```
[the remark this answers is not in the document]
walldiss: No, there is a check for cache miss to return nil.

share/eds/utils.go
walldiss: If close failed on accessor, it means this accessor is leaked and will be never closed. We would want to know about such cases to hunt them down and handle properly. So this log should not spam, unless something is broken

share/eds/store.go
walldiss: Load/Put method, that doesn't check if items exists and stores regardless of duplicates could be confusing API. I think extra Has check in this single case is neglect-able and would opt to keep Cache interface simple.

share/eds/store.go
walldiss: There are could be many triggers, so I've applied it partially

share/eds/store.go
```diff
@@ -343,25 +371,31 @@ func (s *Store) Blockstore() bstore.Blockstore {
...
-) (dagstore.ReadBlockstore, error) {
+) (*BlockstoreCloser, error) {
 	key := shard.KeyFromString(root.String())
-	accessor, err := s.getCachedAccessor(ctx, key)
+	accessor, err := s.cache.Get(key)
+	if err == nil {
+		return blockstoreCloser(accessor)
+	}
+
+	// if the accessor is not found in the cache, create a new one from dagstore
```
derrandz, line 393: If the accessor is not found in the cache, this would mean that some other routine has used it and evicted it. In this case, I think it would be insightful to have a metric for how many accessors we get from cache (_and were used_) and how many new ones were created and not retrieved from cache.

I am having a suspicion that maybe more are newly created than retrieved from cache, in which case, caching (_as is here_) would have minimal performance improvement (_cuz once used, the accessor is evicted_), and would make your second upcoming feature (_`but it could be changed to smarter release based on current readers count. Will be implemented as separate issue.`_) crucial for cache to be effective, but that's if my understanding holds true (_I am not sure when `cache.Remove` is called and by whom_), lmk.

share/p2p/shrexeds/server.go
```diff
@@ -100,8 +100,16 @@ func (s *Server) handleStream(stream network.Stream) {
...
-	status := p2p_pb.Status_OK
+	var status p2p_pb.Status
 	switch {
+	case err == nil:
+		defer func() {
+			if err := edsReader.Close(); err != nil {
+				log.Warnw("closing car reader", "err", err)
+			}
+		}()
+		edsReader.Close()
```
derrandz, line 111: Are closing the reader twice for a specific purpose or is this a mistake?

derrandz (commented):
Glanced over this for learning purposes, got only one comment/question.

share/eds/store.go
walldiss: Accessor could be not in cache, because of 2 reasons:
- it was never put in cache since the startup
- it was evicted, because the cache was full and another item has replaced this one.

Thank you for metrics suggestion. Those you mentioned are crucial to monitor cache performance and implemented in this PR.
 - cache Get hit/miss. Based on miss rate, we can know how many times new accessor will be loaded (either in cache or on stack for single access).
 - Amount loaded into cache is same as eviction rate (because cache is lru). There is eviction rate metric for that

One of the goals of this rework is to prevent caching everything. Given that cache is very limited in its size, we want to prioritise caching items, that has higher chance to be reused. So others accessors with lower probability of being requested should be not cached, because they could take place of accessors with higher demand. It is achieved by  some accessor are never going through cache, like the once you are reffering to

share/p2p/shrexeds/server.go
walldiss: Mistake. Good find, thank you!

share/eds/cache/accessor_cache.go
```diff
@@ -0,0 +1,189 @@
...
+var _ Cache = (*AccessorCache)(nil)
+
+// AccessorCache implements the Cache interface using an LRU cache backend.
+type AccessorCache struct {
+	// The name is a prefix that will be used for cache metrics if they are enabled.
+	name string
+	// stripedLocks prevents simultaneous RW access to the blockstore cache for a shard. Instead
+	// of using only one lock or one lock per key, we stripe the shard keys across 256 locks. 256 is
+	// chosen because it 0-255 is the range of values we get looking at the last byte of the key.
+	stripedLocks [256]sync.Mutex
```
derrandz, line 24: Stripped locks are really effective, except that an array of mutexes as is will cause what's called false sharing.

Although multiple routines will get to work concurrently on different shards and not block each other (_except for same shard routines_), some "unrelated" routines will "block" (_hardware level_) each other due to CPU cacheline, basically because they are -_say_- trying to access two different mutexes each, which live on the same cacheline. Although they are not sharing any resources, the way how the cpu cacheline works will cause them to "falsely share" that cache line.

Since an array is contiguous space in memory, and a mutex is exactly 8 bytes, 8 of them will live in the same cacheline, so with this specific case, if only 64 shards are being accessed by 64 routines, 1 routine for each shard, and let's say half of the routines want to access the shards from 0-31 and are on cpu core 1 and the other half that is trying to access 31-63 and are on cpu core 2, although in theory each should get their respective lock fine, what will happen underneath is that whenever the first half of the routines will try to get a lock from the array, the second half is basically blocked waiting for core 1 to send back the newly modified cacheline, and so on back end forth between the cores for each lock access and modification.

If we crank it up to 255 and dunno how many routines, the performance will see an impact.

So to avoid this, each lock should be on its own cache line, which is possible in go by using `cpu.CacheLinePad`, but then it would require the mutex to be embedded in a new type and also embeds `cpu.CacheLinePad`:

```go
type paddedLock struct {
     sync.Mutex
     _ cpu.CacheLinePad
}

// ...

var stripedLocks [256]paddedLock
```

Correct me on this

derrandz (commented):
a little optimization

share/eds/cache/accessor_cache.go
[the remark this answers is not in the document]
distractedm1nd: no the phrasing is fine, was just unsure if it was referring to some other refcounting or the one that isnt in this pr

share/eds/utils.go
distractedm1nd: I don't know why I thought that it was logging on every close and not just on errors - I must have misread something. agree with you

share/eds/blockstore.go
distractedm1nd: Yes but we don't actually save them in the inverted index - check how our inverted_index.go differs from the default dagstore impl - we only save one shard key per CID, even if we find another shard that the CID points to

share/eds/cache/accessor_cache.go
Wondertan: @derrandz, I think you are correct with the false sharing argument. We use stripped locking in several places in the code, so it would be nice to make an issue about making all of them cache padded and then PR which I'll happily review.

share/eds/blockstore.go
walldiss: Ah ok I get it. Is it no really incorrect, it just pointless for loop, that loops over 1 item every time. I've removed for loop and updated comments here.

share/eds/cache/accessor_cache.go
```diff
@@ -0,0 +1,194 @@
...
+
+// accessorWithBlockstore is the value that we store in the blockstore Cache. It implements the
+// Accessor interface.
+type accessorWithBlockstore struct {
+	sync.RWMutex
+	shardAccessor Accessor
+	// The blockstore is stored separately because each access to the blockstore over the shard
+	// accessor reopens the underlying CAR.
+	bs dagstore.ReadBlockstore
+}
```
Wondertan, line 40: [nit]
move methods closer to the struct

share/eds/cache/cache.go
```diff
@@ -0,0 +1,49 @@
...
+	"go.opentelemetry.io/otel"
+)
+
+var (
+	log   = logging.Logger("share/eds/cache")
+	meter = otel.Meter("eds_store_cache")
+)
+
+var (
+	ErrCacheMiss = errors.New("accessor not found in blockstore cache")
```
Wondertan, line 20: [q]
Is it used anywhere besides tests? Should we make it private then?

Wondertan (commented):
Trying this review format that app team uses a lot

share/eds/cache/cache.go
```diff
@@ -0,0 +1,49 @@
...
+	EnableMetrics() error
+}
+
+// Accessor is a interface type returned by cache, that allows to read raw data by reader or create
+// readblockstore
+type Accessor interface {
+	Blockstore() (dagstore.ReadBlockstore, error)
+	Reader() io.Reader
+	io.Closer
+}
```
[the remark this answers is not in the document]
walldiss: Accessor interface should be defined in dagstore. It is part of its api, that is implemented by dagstore.Accessor type. For some reason they use exact type instead of interface, so it is impossible to have a custom implementation of it and it such ability is crucial for this PR to be possible.
Ideally I would make PR to our dagstore fork to use interface, but it feels like pointless work at this point, given an estimate, that we will get rid of it soon with ipldv2 work going on. And if it is not in dagstore, I don't know where it could be moved and not cause dependency graph issues. So it lives here for now.

share/eds/cache/cache.go
```diff
@@ -0,0 +1,49 @@
...
+		ctx context.Context,
+		key shard.Key,
+		loader func(context.Context, shard.Key) (Accessor, error),
+	) (Accessor, error)
+
+	// Remove removes an item from Cache.
+	Remove(shard.Key) error
+
+	// EnableMetrics enables metrics in Cache
+	EnableMetrics() error
```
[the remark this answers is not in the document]
walldiss: It is the same pattern we use that in every other component, that uses metric in node codebase. So I would rather discuss and apply changes in scope of separate issue/PR.

share/eds/cache/cache.go
```diff
@@ -0,0 +1,49 @@
+package cache
```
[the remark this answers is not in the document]
walldiss: Yeah, it doesn't need to be exportable. But I would also prefer it to live close to other storage related code to make code navigation easier

share/eds/store.go
```diff
@@ -39,6 +40,8 @@ const (
 	// shards that are currently available but inactive, or errored.
 	// We don't use transient files right now, so GC is turned off by default.
 	defaultGCInterval = 0
+
+	defaultCacheSize = 128
```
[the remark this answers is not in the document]
walldiss: Would do different PR to not complicate reviewing. This one is dedicated to cache rework.

share/eds/store.go
```diff
@@ -262,17 +264,35 @@ func (s *Store) put(ctx context.Context, root share.DataHash, square *rsmt2d.Ext
...
-		return nil
 	}
+
+	if result.Error != nil {
+		return fmt.Errorf("failed to register shard: %w", result.Error)
+	}
+
+	// the accessor returned in the result will be nil, so the shard needs to be acquired first to
+	// become available in the cache. It might take some time, and the result should not affect the put
+	// operation, so do it in a goroutine
```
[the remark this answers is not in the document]
walldiss: >So let's look into what does make this some time. We already indexed everything at this point and we have in memory mount, so acquiring should not take a lot or do I miss anything?

It is happening, because dagstore uses single event loop, that causes everything to get stuck, once some operation obtained the control and not releasing it for some time. 


The motivation for this change was:
This operation sometimes happens to be long sometimes based on observed data for long hanging operations. Usually it is fast, but could take 10-50s to finish in times when storage is doing a lot of Put operations (while historical sampling). The extra delay to acquire the shard is putting Put operation into risk of context timeout error. If error happens, it will cause retry of whole sampling + storing which is heavy. So to remove this risk, accessor for the cache is acquired in parallel, allowing Put to be released right after shard is registered successfully. 

image


On your downsides, I think they are neglect able compared to errors that would be caused by sync cache registration:
 -  Tests with no sleeps.
 -> Only tests related to cache needs a sleep. There are not that many
 - Guarantee that there are no errors in acquiring.
 -> I don't see how async cache registration bring any risk of acquiring error. Would be great if you can add examples of such errors introduced by async.
 - Less implementation complexity
 -> It not that complex, with just extra "fire and forget" go-routine semantic being 4 lines of code
            
>Orthogonal thought: Why aren't we calling RegisterShard in parallel with WriteEDS?

Because if register shard fails, we won't be able to acquire it later. So the registration have to be done sync to allow transactional property of writeeds+registerShard

share/eds/cache/accessor_cache.go
walldiss: This is core struct for this file, so I moved it to the top to ease code reading.

share/eds/cache/cache.go
walldiss: Not now, but it is part of Cache interface. Errors that are returned by cache are not always ErrCacheMiss.

share/eds/cache/accessor_cache.go
derrandz: 🫡

go.mod
```diff
@@ -76,6 +75,8 @@ require (
 	google.golang.org/protobuf v1.31.0
 )
 
+require github.com/hashicorp/golang-lru v1.0.2 // indirect
```
Wondertan, line 78: Idk why it happens for you(you aren't the first one I do a review where this artifact happens), I never saw it myself.

go.mod
```diff
@@ -186,7 +187,7 @@ require (
 	github.com/hashicorp/go-safetemp v1.0.0 // indirect
 	github.com/hashicorp/go-version v1.6.0 // indirect
 	github.com/hashicorp/golang-lru/arc/v2 v2.0.5 // indirect
-	github.com/hashicorp/golang-lru/v2 v2.0.5 // indirect
+	github.com/hashicorp/golang-lru/v2 v2.0.5
```
Wondertan, line 190: This need go mod tidy to get the direct out of indirect group

share/eds/cache/accessor_cache.go
```diff
@@ -180,11 +159,11 @@ type accessorCloser struct {
 	io.Closer
 }
 
-func newCloser(abs *accessorWithBlockstore) (*accessorCloser, error) {
+func newCloser(abs *accessorWithBlockstore) *accessorCloser {
```
Wondertan, line 162: [nit] another closer wants to be moved to utils

share/eds/cache/cache.go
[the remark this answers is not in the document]
Wondertan: Agree with reasoning to ignore this

share/eds/cache/cache.go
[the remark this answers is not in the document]
Wondertan: I don't think internal pkg harms navigation experience anyhow, especially if you use JetBrains IDE. You can search and open everything quickly as well use CTRL to dive into the imported code

share/eds/cache/cache.go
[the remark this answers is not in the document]
Wondertan: Ofc, as discussed in sync

share/eds/cache/cache.go
Wondertan: 👌🏻

share/eds/cache/accessor_cache.go
walldiss: I'm confused tbh, could you suggest the order you think would be better? Currently, methods declared in order of declared types:
1. AccessorCache type declaration
2. accessorWithBlockstore  type declaration
3. AccessorCache methods
4. accessorWithBlockstore  methods

share/eds/cache/cache.go
Wondertan: So do you mean it will be used in a later PR?

share/eds/cache/accessor_cache.go
walldiss: This one is is noop closer, that will be replaced in following PR with ref counter, that will live in this file.

share/eds/cache/cache.go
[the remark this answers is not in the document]
Wondertan: This is resolved though

go.mod
walldiss: I think it is fine:
```
❯ go mod why github.com/hashicorp/golang-lru
# github.com/hashicorp/golang-lru
github.com/celestiaorg/celestia-node/cmd/cel-shed
github.com/celestiaorg/go-header/store
github.com/hashicorp/golang-lru

```

share/eds/store.go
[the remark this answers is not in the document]
Wondertan: Is there a way to construct a home-backed accessor using the data we have in put to avoid interactions with DagStore event loop? Theoretically seems possible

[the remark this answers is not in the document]
Wondertan: Although, we may just leave with this as we will stop using dagstore in a few months from now

go.mod
Wondertan: I mean the import in its own `require` instead of being grouped with others

share/eds/cache/accessor_cache.go
Wondertan: I see, ok

share/eds/store.go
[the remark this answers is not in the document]
walldiss: It is possible, but it will have to be different accessor type, unrelated to dagstore. It would require to make and adapter for inMemoryMount to satisfy Accessor interface. Such accessor will not be tracked by dagstore status model and refs count tho. It is a bit tricky, so definitely not in the scope of this PR.
My take is to keep it as is, and spent more time working on dagstore removal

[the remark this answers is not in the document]
Wondertan: 👍🏻

share/eds/cache/accessor_cache.go
Wondertan: My nit suggestion is to swap 3 and 4, because accessorWithBlockstore is core and has only two small methods

share/eds/cache/cache.go
walldiss: Made it private, will make public once there is a handler

Wondertan (approved):

distractedm1nd (approved):

share/eds/cache/accessor_cache.go
derrandz: https://github.com/celestiaorg/celestia-node/issues/4803

Issue #2608, opened 2023-08-24
fix(eds/store): Synchronize WriteEDS calls


distractedm1nd:
Bridges sometimes call writeEDS concurrently, which can lead to data corruption as seen in #2335. This can be fixed by making sure the file writing is only happening by one goroutine at a time.

derrandz:
I wonder if using locks on the store is sufficient solution or too simplistic, any suggestions or direction if I want to take up this issue?

Wondertan:
Stripped locking would be efficient enough

derrandz:
Do we have a test that reproduces the issue to work against? also your suggestion to use stripped locking implies that we have multiple store instance, is that the case? my understanding is that a bridge node has single store instance shared between multiple parties including exchange and listener, whose calls seem to be causing the issue. Your clarification would be much appreciated.

walldiss:
Locking will make 2 parallel put synchronised. We actually don't need to sync them, but rather only leave only one process running. So we save valuable store throughput capacity (biggest bottleneck for syncing rn). 

https://github.com/celestiaorg/celestia-node/pull/2613

musalbas:
In what scenarios are we writing EDS of the same block multiple times?

Wondertan:
We do right because of a diff bug. Even when we fix it, we should still synchronise the EDSStore puts so that data corruption is impossible.

PR #2613, opened 2023-08-25
fix(share/eds): skip concurrent writes on edsstore Put

walldiss:
Short circuit parallel operations  

Closes https://github.com/celestiaorg/celestia-node/issues/2608


share/eds/store.go
```diff
@@ -64,6 +65,9 @@ type Store struct {
 	// lastGCResult is only stored on the store for testing purposes.
 	lastGCResult atomic.Pointer[dagstore.GCResult]
 
+	// skip is used to skip parallel operations
+	skip sync.Map
```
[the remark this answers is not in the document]
walldiss: I can argues, that 256 mutexes will take more. The amount of items in map should be equal to amount of parallel writes.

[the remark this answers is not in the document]
walldiss: But the point is to node do eds.Write, when we don't need to. It is much a bigger performance factor here

share/eds/store.go
```diff
@@ -197,6 +201,12 @@ func (s *Store) Put(ctx context.Context, root share.DataHash, square *rsmt2d.Ext
 }
 
 func (s *Store) put(ctx context.Context, root share.DataHash, square *rsmt2d.ExtendedDataSquare) (err error) {
+	if _, exists := s.skip.LoadOrStore(root.String(), 1); exists {
+		// no need to do any work, another routine is putting eds for the same root
+		return nil
```
Wondertan, line 206: Another potential issue is that Put will return immediately for other routines without making them wait until the operation is actually finished. This is problematic if Put fails for any reason keeping other routines in false state.

walldiss: Thats a point! Changed skip to lock. Also some routines, like rpc get call, don't really need to wait for Put to succeed and should resolve regardless.

share/eds/store.go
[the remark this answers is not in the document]
walldiss: Well yeah, and allocating in long run with many operations. So stripped prefix should be better.

distractedm1nd (dismissed):

share/eds/store.go
```diff
@@ -197,6 +201,10 @@ func (s *Store) Put(ctx context.Context, root share.DataHash, square *rsmt2d.Ext
 }
 
 func (s *Store) put(ctx context.Context, root share.DataHash, square *rsmt2d.ExtendedDataSquare) (err error) {
+	lk := &s.stripedLocks[root[len(root)-1]]
+	lk.Lock()
+	defer lk.Unlock()
+
```
Wondertan, line 207: I wonder if we should put the lock in Put, so that we we don't track duplicates in traces and metrics

walldiss: Good point. Then `Has` check should be moved outside too.

Wondertan: That's good, still not sure why would we even track them

walldiss: It is valuable from component centric approach, when you don't know who are callers of put and how many duplicates client are trying to store. Also Has check is not free, since it is badger lookup

Wondertan (approved):

distractedm1nd (approved):

distractedm1nd:
linter seems to be failing from the weird cache issue again?

Wondertan (approved):

distractedm1nd (approved):

Issue #2604, opened 2023-08-24
share/p2p/peers: Only add nodes to PeerManager that are compatible on desired protocols


renaynay:
For lights, that's `shrex-nd` and for fulls, that's `shrex-nd` and `shrex-eds`

**Implementation**

Peer manager should look into peer store and check of peer supports required protocol. If peer does not support any protocols peer should not be added to peers pools.

Wondertan:
Can be filtered via Peerstore which keep supported protocols information

derrandz:
Proposed an initial PR, would be great having your feedback guys, thanks. 👍🏻

walldiss:
@renaynay Could you please add more details on whats is the motivation for this feature? Asking, to clarify, if peer-manager should aggregate peers with ALL supported protocols or ANY? What if one of the peers implements only some of the protocols? Considering peer-manager is used to aggregate peers for multiple protocols, would it imply, that peer-manager interface should be changed, to allow getting peer by specific protocolID?
```
func (m *Manager) Peer(ctx context.Context, protocolID string, datahash share.DataHash)  (peer.ID, DoneFunc, error) 
```

derrandz:
@walldiss Good point. If the choice is made to aggregate peers with ANY supported protocol, then this would have an impact on how we store such peers in the peer manager, and we should index them by their supported protocols, and thus (_as you mentioned_) would change the interface to support fetching peers by their specific protocol.ID

But I wonder if this might open room for some weird behavior, because if a full node is on the wrong protocol for `shrex-eds` but is on the right protocol for `shrex-nd`, I wonder if this would somehow break `shrex-nd`, cuz maybe the format of the retrieved shares from `shrex-eds` is out-dated and thus would (_dunno_) end up with breaking `shrex-nd` as a result?

Maybe the other question then is: do `shrex-eds` versions require specific `shrex-nd` versions for full nodes..?

Explaining the motivation for this feature request would help answer these questions properly.

renaynay:
> @renaynay Could you please add more details on whats is the motivation for this feature? Asking, to clarify, if peer-manager should aggregate peers with ALL supported protocols or ANY? What if one of the peers implements only some of the protocols? Considering peer-manager is used to aggregate peers for multiple protocols, would it imply, that peer-manager interface should be changed, to allow getting peer by specific protocolID?

[...]

PR #2616, opened 2023-08-26
feat(shrex/peer-manager): only add peers with compatible protocols

derrandz:
## Overview

This PR addresses #2604 

## Checklist

### DIFF
--- a/share/p2p/peers/manager.go
+++ b/share/p2p/peers/manager.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -352,7 +351,7 @@ func (m *Manager) Validate(_ context.Context, peerID peer.ID, msg shrexsub.Notif
 		return pubsub.ValidationReject
 	}
 
-	if !m.arePeerProtocolsCompatible(peerID) {
+	if !m.isSupported(peerID) {
 		logger.Debug("received message from peer with incompatible protocol, reject validation")
 		return pubsub.ValidationReject
 	}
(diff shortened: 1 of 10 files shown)


derrandz:
I won't be able to add a label to the PR and make the CI happy given github perms.

share/p2p/peers/manager.go
```diff
@@ -418,6 +428,31 @@ func (m *Manager) isBlacklistedPeer(peerID peer.ID) bool {
 	return !m.connGater.InterceptPeerDial(peerID)
 }
 
+func (m *Manager) arePeerProtocolsCompatible(peerID peer.ID) bool {
```
vgonkivs, line 431: I'd call it `isSupported()` or `supportsProtocols()`

share/p2p/peers/manager.go
```diff
@@ -418,6 +428,31 @@ func (m *Manager) isBlacklistedPeer(peerID peer.ID) bool {
...
+	peerProtos, _ := m.host.Peerstore().GetProtocols(peerID)
```
vgonkivs, line 432: There is one more method in the peerStore that fits better in this case - `SupportsProtocols`

vgonkivs (changes requested):
Hey Hamza, glad to see you. Left some comments

share/p2p/peers/manager.go
derrandz: Thanks!

share/p2p/peers/manager.go
derrandz: Cleaner, changed 👍🏻

derrandz:
Thanks @vgonkivs 
Reiterated on this to include your suggestions, lmk! 👍🏻

renaynay (changes requested):
@derrandz I left a comment on the associated issue - what we need is a way to retrieve peers from the peer manager `Get` for the specific protocol that's requesting the peer (e.g., shrex-eds needs a peer running the same shrex-eds protocol)

So `Get` needs to take a procolID and return a peer that supports it

Later on, we can optimise discovery for peers that "match best". Not in this PR though.

vgonkivs:
N/A

Issue #2335, opened 2023-06-06
Celestia Node v0.10.4 could not retrieve IPLD node error


jim-counter:
### Celestia Node version

Semantic version: v0.10.4

### OS

Ubuntu 22.04.2 LTS

### Install tools

`make`

### Steps to reproduce it

Start the node.

### Expected result

See `new head` messages and no errors.

### Actual result

See this error message repeating every 30 seconds give or take:

```
2023-06-06T22:06:57.839Z        ERROR        ipld        ipld/namespace_data.go:217        could not retrieve IPLD node        {"nID": "91b00635f5a0d733", "pos": 0, "err": "nmt node not found"}
```

### Relevant log output

```Shell
Jun 06 22:06:11 celestia[257688]: 2023-06-06T22:06:11.071Z        INFO        header/store        store/store.go:353        new head        {"height": 669018, "hash": "08B228615B8896CA4FE235E7430956861203ADC41959E3E8690DDCB642D053B5"}
Jun 06 22:06:21 celestia[257688]: 2023-06-06T22:06:21.857Z        INFO        header/store        store/store.go:353        new head        {"height": 669019, "hash": "689C7C6FA9208FBA2E75500FF5785ACE04A3CA84B48626E59EC0B8A1C562BC02"}
Jun 06 22:06:25 celestia[257688]: 2023-06-06T22:06:25.457Z        ERROR        ipld        ipld/namespace_data.go:217        could not retrieve IPLD node        {"nID": "91b00635f5a0d733", "pos": 0, "err": "nmt node not found"}
Jun 06 22:06:25 celestia[257688]: 2023-06-06T22:06:25.457Z        WARN        shrex/nd        shrexnd/server.go:136        server: nd not found        {"peer": "12D3KooWCdo1Ebe1k3HN9PHEma8DN1xqZGAgtpzeKWGoZtKfMpUq", "namespaceId": "91b00635f5a0d733", "hash": "41CDD8234F00DC74BFE122EB6DC392B0CAF621E4BF25300CFB99EDF1A5BA05D5"}
Jun 06 22:06:27 celestia[257688]: 2023-06-06T22:06:27.455Z        INFO        share/discovery        discovery/discovery.go:254        discovering peers        {"want": 1}
Jun 06 22:06:27 celestia[257688]: 2023-06-06T22:06:27.609Z        INFO        share/discovery        discovery/discovery.go:300        discovered wanted peers        {"amount": 5}
[...]

PR #2625, opened 2023-08-28
feat(eds/store): remove corrupted blocks from store

distractedm1nd:
WIP. Related #2335 

Still need to write some tests then I'll take out of draft


share/eds/store_test.go
```diff
@@ -122,6 +122,37 @@ func TestEDSStore(t *testing.T) {
...
+		assert.NoError(t, err)
+
+		err = os.Remove(path)
+		assert.NoError(t, err)
+
+		_, err = edsStore.GetCAR(ctx, dah.Hash())
+		assert.Error(t, err)
+
+		// give some time for the store to remove
+		time.Sleep(300 * time.Millisecond)
```
ramin, line 148: it might be nice to wrap this in a for {} and actually check that the path you deleted in os.Remove above is actually clear before proceeding to the next check, vs relying on the time.Sleep which could be a intermittent flakey test later. Say a runner is slow one time, or for some reason THIS time it look 301ms, or something else.

maybe something like 

```go

_, err := os.Remove(path)

maxAttemps := 5
attempts := 0
wait := 300 * time.Millisecond

for {
  _, err := os.Stat(path)
  if err != nil {
    if !os.IsNotFound(err) {
      break
    }   
    t.Error(err)
  }
  
  attempts++
  if retries >= maxAttempts {
     break
  } 

  time.Sleep(wait)
}

```

we could potentially max the 300ms much shorter and proceed through the test quicker than the 300ms in most cases too

share/p2p/peers/manager.go
```diff
@@ -148,14 +148,16 @@ func (m *Manager) Start(startCtx context.Context) error {
...
 	validatorFn := m.metrics.validationObserver(m.Validate)
-	err := m.shrexSub.AddValidator(validatorFn)
-	if err != nil {
-		return fmt.Errorf("registering validator: %w", err)
-	}
 
-	err = m.shrexSub.Start(startCtx)
-	if err != nil {
-		return fmt.Errorf("starting shrexsub: %w", err)
+	if !m.params.DisableShrexSub {
```
ramin, line 152: minor style thought: this requires a little mental arithmetic, `if not false` double negative, which is more easily parsed as `true` and is a little oblique how it relies on . It might be nicer to read as `m.params.ShrexSubEnabled`, but i appreciate this would mean we'd have to be mindful of having a default of true for the parameter any time the params are set.

share/eds/store.go
```diff
@@ -172,6 +180,26 @@ func (s *Store) gc(ctx context.Context) {
 	}
 }
 
+func (s *Store) watchForFailures(ctx context.Context) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case res := <-s.shardFailures:
+			log.Errorw("removing shard after failure", "key", res.Key, "err", res.Error)
```
ramin, line 189: is there anywhere we log could capture metrics around this, something as basic as a "shardFailureCount"? I am imagining a scenario where a node runner complains that the something is bad and it'd be possible to say "ah, whats the shardFailureCount" or even to graph that somewhere and see when something like a disk was failing as the cause.

i see some metrics stuff in node, but not sure if any are appropriate? this would be a "nice to have"

ramin (commented):
made some minor comments from perspective of a new set of eyes on the project. I think logging shard failures and removals would be super useful for both a node runner, but also as it feels like this doesn't solve the issue, but is a step towards more work in this area.

share/eds/store_test.go
distractedm1nd: Yeah, this was a bit lazy from me. Will change to a `select` with a ticker that will just repeat the check until the test context deadlines

share/p2p/peers/manager.go
distractedm1nd: Definitely agree and nit about this often, didn't catch it here so thanks 🙏

share/eds/store.go
distractedm1nd: Hmmm, this is supposed to be once in a blue moon event, but if we add a metric people could add alerts for such an event. This would definitely be a good thing, because this removal process is only a bandaid that requires manual attention anyways. @walldiss do you have anything against us adding a metric here?

ramin (dismissed):

share/p2p/peers/options.go
```diff
@@ -19,8 +21,15 @@ type Parameters struct {
 
 	// EnableBlackListing turns on blacklisting for misbehaved peers
 	EnableBlackListing bool
+
+	// EnableShrexSub enables shrexsub for the peer manager. Setting it to false
+	// is used to enable shrex on bridge nodes without creating a second
+	// instance of shrexsub.
+	EnableShrexSub bool
```
renaynay, line 28: pls make it more clear that the way shrexsub is used in peerman is for listening, and that disabling it means node won't listen for recent block notifications

share/p2p/peers/manager.go
```diff
@@ -148,14 +148,16 @@ func (m *Manager) Start(startCtx context.Context) error {
...
+		if err != nil {
+			return fmt.Errorf("registering validator: %w", err)
+		}
+		err = m.shrexSub.Start(startCtx)
+		if err != nil {
+			return fmt.Errorf("starting shrexsub: %w", err)
+		}
 	}
 
 	headerSub, err := m.headerSub.Subscribe()
```
renaynay, line 163: let's also not subscribe here then and not listen? The whole point is that bridge node shouldn't care to listen on the sub, no?

share/p2p/peers/manager.go
```diff
@@ -148,14 +148,16 @@ func (m *Manager) Start(startCtx context.Context) error {
...
+	if m.params.EnableShrexSub {
+		err := m.shrexSub.AddValidator(validatorFn)
+		if err != nil {
+			return fmt.Errorf("registering validator: %w", err)
+		}
+		err = m.shrexSub.Start(startCtx)
+		if err != nil {
+			return fmt.Errorf("starting shrexsub: %w", err)
+		}
 	}
```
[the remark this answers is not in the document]
distractedm1nd: I like this better, thanks for the suggestion

share/eds/store.go
distractedm1nd: will add

nodebuilder/share/module.go
```diff
@@ -48,6 +47,34 @@ func ConstructModule(tp node.Type, cfg *Config, options ...fx.Option) fx.Option
 		),
 	)
 
+	shrexGetterComponents := fx.Options(
```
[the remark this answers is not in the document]
distractedm1nd: I can't because it is now part of the shared bridge/full components so must be defined before

share/p2p/peers/manager.go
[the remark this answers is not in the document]
distractedm1nd: On second thought, shrexsub is already inside the NewManager constructor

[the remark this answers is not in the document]
walldiss: Remove from contractor and pass in the option.

share/eds/store.go
```diff
@@ -64,6 +64,8 @@ type Store struct {
 	// lastGCResult is only stored on the store for testing purposes.
 	lastGCResult atomic.Pointer[dagstore.GCResult]
 
+	shardFailures chan dagstore.ShardResult
+
```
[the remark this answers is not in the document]
distractedm1nd: Not really, because it needs to exist on dagstore construction and still be readable on start

renaynay (approved):
Lint and pls check swamp tests locally

walldiss (commented):
Small one left

share/p2p/peers/manager.go
```diff
@@ -169,7 +178,6 @@ func (m *Manager) Start(startCtx context.Context) error {
...
 
 	go m.subscribeDisconnectedPeers(ctx, sub)
```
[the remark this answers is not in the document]
distractedm1nd: Doesn't it remove disconnected peers from the discovery pool?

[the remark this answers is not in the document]
walldiss: Discovery will remove on its own, it has remove event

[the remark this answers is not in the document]
distractedm1nd: done

walldiss (approved):

PR #2630, opened 2023-08-29
improvement!(blob/service): add options to submit

### LINKED ISSUE #2629: blob: add WithFee and WithGasLimit options to Submit
The user should have the possibility to specify fee params

vgonkivs:
## Overview
Allows to configure fee and gasPrice in blob.Submit. 
Closes #2629 

Will update blob's rpc in a follow-up PR.
## Checklist

FYI @jcstein , @nashqueue


blob/service.go
```diff
@@ -55,10 +57,18 @@ func NewService(
 // Allows sending multiple Blobs atomically synchronously.
 // Uses default wallet registered on the Node.
 // Handles gas estimation and fee calculation.
-func (s *Service) Submit(ctx context.Context, blobs []*Blob) (uint64, error) {
+func (s *Service) Submit(ctx context.Context, blobs []*Blob, options ...Option) (uint64, error) {
```
distractedm1nd, line 60: We cannot do variadic params on the RPC :( I would suggest adding a new endpoint, it is less breaking and keeps the Submit signature clean for most people's use case

ramin: i think there was a suggestion/decision to add `SubmitWithOptions` this morning?

blob/service.go
```diff
@@ -55,10 +57,18 @@ func NewService(
...
 	log.Debugw("submitting blobs", "amount", len(blobs))
 
-	resp, err := s.blobSumitter.SubmitPayForBlob(ctx, types.OneInt().Neg(), 0, blobs)
+	s.feeParams = &feeParams{
```
ramin, line 63: it'd be nice and idiomatic to add a `DefaultParams() &feeParams` or similar in the options.go, then we can also add a test around the defaults

blob/options.go
```diff
@@ -0,0 +1,27 @@
...
+
+import "cosmossdk.io/math"
+
+// Option is a function that allows to configure the Blob Service requests.
+type Option func(*Service)
+
+// feeParams contains the information about fee and gasLimit price.
+type feeParams struct {
+	fee    math.Int
+	gasLim uint64
```
ramin, line 11: any reason not to call this `gasLimit` vs `gasLim`?

vgonkivs:
After a discussion with @distractedm1nd, we figured out that the Option param won't work on the RPC because 'Option' is a typedef on a `func()` and marshalling functions/methods does not make much sense, so the request will fail. We agreed to have one more endpoint `SubmitWithFee` that will have fee+gaslimit as input params, and `Submit` will be used as blobs submitter with the default fee and gasLim

blob/service.go
vgonkivs: There could be lots of potential options, so I'd suggest to call it SubmitWithFee for this particular case

vgonkivs:
One more idea came to my mind: we can rework blob.Submit by getting new type blob.Batch.

```go
type Batch struct {
Blob []*Blob
Fee int64
GasLimit uint64
}

func NewBatch(blobs *[]Blob) *Batch {
  return &Batch {
     Blob: blobs,
    Fee: -1,
}

type func(*Batch) Option

func WithFee(int64) Option {}
func WithGasLimit(uint64) Option {}
}

func(s *service)Submit(ctx context.Contex, blobs *Batch){}
```

vgonkivs:
Anyway we are going to make breaking changes but the second approach looks better to me as we will still have a single endpoint for the Submit

renaynay (changes requested):
@vgonkivs please don't break the method yet - let's just extend it by adding the additional SubmitWithOptions and have a struct called `SubmissionOptions` or just `Options` that contains fee and gas limit as field 

i think the concept of batch is unnecessarily confusing

vgonkivs:
ok

blob/options.go
vgonkivs: renamed to GasLimit

ramin (dismissed):
i like this and changes are good. I'll defer to @renaynay and @Wondertan on if they approve/deny the breaking change to Submit RPC but the things i suggested == :shipit:

blob/service.go
```diff
@@ -51,14 +51,32 @@ func NewService(
...
+		GasLimit: 0,
+	}
+}
+
 // Submit sends PFB transaction and reports the height in which it was included.
 // Allows sending multiple Blobs atomically synchronously.
 // Uses default wallet registered on the Node.
 // Handles gas estimation and fee calculation.
-func (s *Service) Submit(ctx context.Context, blobs []*Blob) (uint64, error) {
+func (s *Service) Submit(ctx context.Context, blobs []*Blob, options *SubmitOptions) (uint64, error) {
```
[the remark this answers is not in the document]
vgonkivs: I'm not sure, we have enough time/capacity to fix variadics now. We can go with `SubmitOptions` and once the variadics are fixed, we can simply remove `nil`. Or users can use `DefaultOptions` and then nothing should be removed at all.

I agree that removing the whole endpoint will be harder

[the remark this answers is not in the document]
walldiss: Yet another suggestion:
Keep only version with options, but make options a private struct. It will reduce ux to only 2 cases: defaults, manually set by constructor:
```
type submitOptions struct {
	fee      int64
	gasLimit uint64
}

func DefaultSubmitOptions() submitOptions {
	return submitOptions{
		fee:      -1,
		gasLimit: 0,
	}
}

func SubmitOptions(fee, gaslimit int) submitOptions{
	return submitOptions{
		fee:      -1,
		gasLimit: 0,
	}
}
```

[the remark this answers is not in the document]
vgonkivs: Also sounds good to me.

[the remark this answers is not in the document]
vgonkivs: Just realised that private field will not work as we are marshaling/unmarshaling them during rpc calls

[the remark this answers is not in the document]
distractedm1nd: Relying on passing nil is something golang specific, and this will be ugly in the rust version where you will always need to instantiate the struct. I would much much rather have two methods, it is also less breaking. I absolutely love having a method that doesn't take any parameters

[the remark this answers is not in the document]
vgonkivs: Still don't like the approach where you have to specify a new endpoint once you've faced some edge cases or you need a possibility to configure the request. If we accept this pattern everywhere in our code, the API can significantly grow in the future imo

nodebuilder/node/admin.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/celestiaorg/celestia-node/libs/authtoken"
 )
 
-const APIVersion = "v0.2.1"
+const APIVersion = "v0.2.2"
```
renaynay, line 13: This needs to match our node version btw

cc @distractedm1nd @MSevey @adlerjohn @Wondertan

renaynay (approved):
Fine w me

walldiss (approved):

Wondertan (approved):
Please make an issue to fix variadics and gtg

distractedm1nd (approved):

nodebuilder/node/admin.go
distractedm1nd: what do you mean by this? API version is not the same as node version

PR #2595, opened 2023-08-22
chore(deps): migrate to boxo

Wondertan:
Thanks to https://github.com/ipfs/boxo/pull/407, we can now fully migrate to boxo and get rid of the oldest `replace` directive in our go.mod for `verifcid`.

TODO: Issue about go-car/v2 migration

PR #2640, opened 2023-08-31
chore(lint/sort-imports): fix absolute path for sort-imports  

walldiss:
If repo folder name is different from "celestia-node" `lint` and `sort-imprort` commands from Makefile will not work.


distractedm1nd (dismissed):

Makefile
```diff
@@ -161,14 +161,14 @@ openrpc-gen:
 lint-imports:
 	@echo "--> Running imports linter"
 	@for file in `find . -type f -name '*.go'`; \
-		do goimports-reviser -list-diff -set-exit-status -company-prefixes "github.com/celestiaorg"  -project-name "github.com/celestiaorg/"$(PROJECTNAME)"" -output stdout $$file \
+		do goimports-reviser -list-diff -set-exit-status -company-prefixes "github.com/celestiaorg" -project-name "github.com/celestiaorg/celestia-node" -output stdout $$file \
```
Wondertan, line 164: You don't need to change that

Makefile
```diff
@@ -161,14 +161,14 @@ openrpc-gen:
...
 		 || exit 1;  \
     done;
 .PHONY: lint-imports
 
 ## sort-imports: Sort Go imports.
 sort-imports:
-	@goimports-reviser -company-prefixes "github.com/celestiaorg"  -project-name "github.com/celestiaorg/"$(PROJECTNAME)"" -output stdout .
+	@goimports-reviser -company-prefixes "github.com/celestiaorg" -project-name "github.com/celestiaorg/celestia-node" -output stdout ./...
```
Wondertan, line 171: On the latest version of goimports-reviser the ./... does not work anymore, so please use the diff from https://github.com/celestiaorg/celestia-node/pull/2595/files, which does exactly the same but loops in bash

walldiss: Jeez, now we have to maintain vendored versions of ci tooling, because it is exactly how it works for me. I guess we can achieve it by providing commands to install all ci dependancies with specified versions. 
It was unclear from your PR that the new version of imports-reviser has changed api.

Makefile
walldiss: If you store code in folder named differently (me) it won't work with value taken from `pwd`

walldiss: I happen not to and didn't change it, because there were no reason to do it. Anyway imports-reviser prefix is static and should be attached to github repo path, instead of local folder path.

Makefile
walldiss: CI seems to be happy about new command

Wondertan: Let's make sure our local limiting is happy too

walldiss: Work perfectly for me on both old and new imports-reviser version.

vgonkivs (approved):

Makefile
Wondertan: The current solution doesn't work for me locally

Wondertan: I just ran the command. 
`Failed to fix directory: path is not a directory` - the error I see when `./...` is used

Wondertan: image

Wondertan: Fixed in https://github.com/incu6us/goimports-reviser/releases/tag/v3.4.2, but wasn't yet published to my pkg manager

Wondertan: Using `go install` now for the latest version, so this work now for me, so resolved

Makefile
```diff
@@ -162,17 +162,14 @@ openrpc-gen:
 lint-imports:
 	@echo "--> Running imports linter"
 	@for file in `find . -type f -name '*.go'`; \
-		do goimports-reviser -list-diff -set-exit-status -company-prefixes "github.com/celestiaorg"  -project-name "github.com/celestiaorg/"$(PROJECTNAME)"" -output stdout $$file \
+		do goimports-reviser -list-diff -set-exit-status -company-prefixes "github.com/celestiaorg" -project-name "github.com/celestiaorg/celestia-node" -output stdout $$file \
 		 || exit 1;  \
```
Wondertan, line 166: Change to `./..`?

walldiss: As far as I remember this one was changed to throw an error on first one, to help CI to fail on this check. Otherwise it was closing with exit code 0 even when some files were not formatted properly

Wondertan: I don't like that this is inconsistent and would prefer to unify. But ok to keep if there is no other solution to this

walldiss: This one is different only because in this case goimports-reviser is used as linter. We expect it to return status code 1 at any errors, and it should be supported by`-set-exit-status` flag. But the flag doesn't work as internded, so we use a workaround with `find`. 

When we use it for fixing imports it works just fine with `./...`

Wondertan (approved):

distractedm1nd (approved):

PR #2641, opened 2023-08-31
chore(deps): bump golanci-lint workflow

distractedm1nd:
Self explanatory. The reason for the fix in core_accessor was an implicit memory aliasing. It was taking the address of the loop variable, which doesn't change in between iterations ofc, so the every blob pointed to the last blob

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -4,7 +4,7 @@ run:
 linters:
   enable:
     - bodyclose
-     # - depguard as of v1.54.2, the default config throws errors on our repository
+     # - depguard as of v1.54.2, the default config throws errors on our repo
     - dogsled
     - dupl
     - errcheck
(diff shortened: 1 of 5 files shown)


renaynay (approved):

.golangci.yml
```diff
@@ -4,7 +4,7 @@ run:
 linters:
   enable:
     - bodyclose
-    - depguard
+     # - depguard as of v1.54.2, the default config throws errors on our repo
```
Wondertan, line 7: Is it worth fixing them?

distractedm1nd: Don't think so, we already have good import linting

distractedm1nd: And also, didn't use it before. It whitelisted everything

vgonkivs (approved):

PR #2643, opened 2023-08-31
deps!: bumping nmt to v0.19.0, breaking rpc response fields

distractedm1nd:
Related: #2633


Wondertan:
Changes of the name do not break protos. Changes of the type and fields reorder do.

Wondertan:
> Or does the way it work not care about naming and just uses data types and offsets.

Yes

renaynay (commented):
Can we bump when we bump core?

distractedm1nd:
closing in favor of core bump #2646

PR #2644, opened 2023-08-31
feat(headertest): Validate exact error types in header tests

Manav-Aggarwal:
## Overview
Closes: #2645 

## Checklist


distractedm1nd (approved):
Thank you for this <3

header/headertest/verify_test.go
```diff
@@ -15,51 +17,80 @@ func TestVerify(t *testing.T) {
...
 		},
 	}
 
 	for i, test := range tests {
 		t.Run(strconv.Itoa(i), func(t *testing.T) {
 			err := trusted.Verify(test.prepare())
-			if test.err {
-				assert.Error(t, err)
-			} else {
+			// Case 1 & 2: test.err == nil and err == nil or err != nil
```
renaynay, line 67: Let's remove the case comments throughout the test pls

renaynay (changes requested):
I don't agree with this approach - we'd have nested error types with nested reasons. I don't think this is necessary nor good UX.

header/header_errors.go
```diff
@@ -0,0 +1,21 @@
...
+	return fmt.Sprintf("validator hash mismatch error: %s", e.Reason.Error())
+}
+
+type ErrLastHeaderHashMismatch struct {
+	Reason error
+}
+
+func (e *ErrLastHeaderHashMismatch) Error() string {
+	return fmt.Sprintf("last header hash mismatch error: %s", e.Reason.Error())
+}
```
Wondertan, line 21: * Let's make them public only when we need to
* Pls see example here: https://github.com/celestiaorg/go-header/blob/main/verify.go. We don't need to have a reason within an error here. Instead, this can be separate error sentinels with values appended via fmt.Errof

Wondertan (changes requested):

header/header_errors.go
Wondertan: Dismiss the first point. I realized that we have test in a separate pkg, so it has to be public.
Please still apply the second comment

Manav-Aggarwal: - They need to be public here since the tests are in a different package `headertest` whereas the header methods are in package `header`.
- I agree we don't need to have a reason specifically but replacing it with an error sentinel could work only when the error string inside `fmt.Errorf` is fixed unless we want to lose the message with actual error hashes:
```
if !bytes.Equal(untrst.LastHeader(), eh.Hash()) {
	return &libhead.VerifyError{
		Reason: &ErrLastHeaderHashMismatch{fmt.Errorf("expected new header to point to last header hash (%X), but got %X)",
			eh.Hash(),
			untrst.LastHeader(),
		)},
	}
}
```

That's because detecting the error is of that certain type in using an error type like in:

```
switch reason.(type) {
	case *header.ErrValidatorHashMismatch:
		assert.Equal(t, &header.ErrValidatorHashMismatch{}, testReason)
}
```

would have to be switched to comparing the error message instead which in the above case would have to do pattern matching instead of simply being able to look at the error struct type.

cc @MSevey since he originally suggested opening an issue about this: https://github.com/rollkit/rollkit/pull/1026#discussion_r1248284884

Wondertan: >     but replacing it with an error sentinel could work only when the error string inside fmt.Errorf is fixed unless we want to lose the message with actual error hashes:

See https://pkg.go.dev/errors#Is. When you do `fmt.Errorf("%w", ErrValidatorHashMismatch)`, you can still compare against `ErrValidatorHashMismatch` by using `errors.Is` or `assert.ErrorIs`

Manav-Aggarwal: Yes, I understand. Would you not lose actual error hashes which are formatted in the error string then? 

Currently, it looks like:
```
&ErrLastHeaderHashMismatch{fmt.Errorf("expected new header to point to last header hash (%X), but got %X)",
	eh.Hash(),
	untrst.LastHeader(),
)}
```

I know this works well for constant errors like here: https://github.com/celestiaorg/go-header/blob/2fd3b1bb637db76d5bbecf0731ec100dba6a7c02/verify.go#L88 or https://github.com/rollkit/rollkit/pull/1165/files#diff-d178d1a180ac9d7016ec909f5f2c8c4f0a2fc1bde0c72cc60a28b2788ef4183eR10

Manav-Aggarwal: Used error sentinels and wrapped errors with them instead. Feel free to take a look again 😄

header/header.go
```diff
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/pkg/errors"
```
Wondertan, line 9: There is std `errors` pkg, we don't need to import anything.

Wondertan: Specifically for wrapping functionality `fmt.Errorf(%w, err)` will do a trick.

header/headertest/verify_test.go
```diff
@@ -15,51 +16,77 @@ func TestVerify(t *testing.T) {
...
+				case header.ErrLastHeaderHashMismatch:
+					assert.ErrorIs(t, reason, header.ErrLastHeaderHashMismatch)
+				case header.ErrVerifyCommitLightTrustingFailed:
+					assert.ErrorIs(t, reason, header.ErrVerifyCommitLightTrustingFailed)
+				default:
+					assert.Equal(t, testReason, reason)
+				}
+			default:
+				assert.Failf(t, "unexpected error: %s\n", err.Error())
 			}
```
Wondertan, line 90: This all is redundant. We only need to make sure err coming out of `Verify` `errors.Is` to the one in the test case.

Wondertan (changes requested):

header/header.go
Manav-Aggarwal: Made the change, note that now, the formatted error message with the validator set hashes etc. is not propagated.

header/header_errors.go
Wondertan: @Manav-Aggarwal, wait, %w is just an example, you do this:

`fmt.Errorf("%w: expected old header next validators (%X) to match those from new header (%X)", err, values...)`

header/header.go
Wondertan: @Manav-Aggarwal, wait, %w is just an example, you do this:

`fmt.Errorf("%w: expected old header next validators (%X) to match those from new header (%X)", err, values...)`

Manav-Aggarwal: Thanks for clarifying, made the change.

header/header.go
```diff
@@ -174,19 +181,23 @@ func (eh *ExtendedHeader) Verify(untrst *ExtendedHeader) error {
...
-				Reason: fmt.Errorf("expected old header next validators (%X) to match those from new header (%X)",
-					eh.NextValidatorsHash,
-					untrst.ValidatorsHash,
-				),
+				Reason: fmt.Errorf("%w: %w",
+					ErrValidatorHashMismatch,
+					fmt.Errorf("expected old header next validators (%X) to match those from new header (%X)",
+						eh.NextValidatorsHash,
+						untrst.ValidatorsHash,
+					)),
```
Wondertan, line 189:
```suggestion
				Reason: fmt.Errorf("%w: expected old header next validators (%X) to match those from new header (%X)",
					ErrValidatorHashMismatch,
					eh.NextValidatorsHash,
					untrst.ValidatorsHash,
					),
```

Wondertan: Let's simplify it a bit

Wondertan: I would also decrease the verbosity for the error message inside fmt.Errorf and move as much as possible to  ErrValidatorHashMismatch

header/headertest/verify_test.go
```diff
@@ -15,52 +16,64 @@ func TestVerify(t *testing.T) {
...
 				assert.NoError(t, err)
+				return
+			}
+			if err == nil {
+				t.Errorf("expected err: %v, got nil", test.err)
+				return
 			}
+			reason := err.(*libhead.VerifyError).Reason
+			testReason := test.err.(*libhead.VerifyError).Reason
+			assert.ErrorIs(t, reason, testReason)
```
Wondertan, line 76:
```suggestion
			assert.ErrorIs(t, errors.Unwrap(test.err), err)
```

This all can be simplified to

Wondertan: Even for success case it should check that nil == nil

Wondertan (changes requested):

header/headertest/verify_test.go
Wondertan: Ah, we can actually remove the `errors.Unwrap` if we don't wrap the expected test case errors with `VerifyError`. We test for inner error first of all, we don't care much if its wrapped or not

header/header.go
Manav-Aggarwal: Reduced.

header/headertest/verify_test.go
Manav-Aggarwal: The first test checks for the success case where nil == nil. This is the code that enters ```If `test.err` == nil```  and then asserts that `err` is not an error which basically checks if it's nil. 

I tried adding your suggestion but it breaks the test both with `errors.Unwrap` and without it.

Wondertan: > I tried adding your suggestion but it breaks the test both with errors.Unwrap and without it so added errors.Unwrap to both instead which works fine with nil arguments as well.

For this to work you need to apply the following suggestion
> if we don't wrap the expected test case errors with VerifyError

Instead of this
```
err: &libhead.VerifyError{
	Reason: header.ErrVerifyCommitLightTrustingFailed,
}
```

This
```
err: header.ErrVerifyCommitLightTrustingFailed,
```

header/header.go
```diff
@@ -183,10 +191,12 @@ func (eh *ExtendedHeader) Verify(untrst *ExtendedHeader) error {
...
-				Reason: fmt.Errorf("expected new header to point to last header hash (%X), but got %X)",
-					eh.Hash(),
-					untrst.LastHeader(),
-				),
+				Reason: fmt.Errorf("%w: %w",
+					ErrLastHeaderHashMismatch,
+					fmt.Errorf("expected (%X), but got (%X)",
+						eh.Hash(),
+						untrst.LastHeader(),
+					)),
```
Wondertan, line 199: Pls also do this in one fmt.Errorf for consistency

header/header.go
```diff
@@ -166,6 +167,12 @@ func (eh *ExtendedHeader) Validate() error {
 	return nil
 }
 
+var (
+	ErrValidatorHashMismatch           = errors.New("validator hash mismatch error")
+	ErrLastHeaderHashMismatch          = errors.New("last header hash mismatch error")
```
Wondertan, line 172: uber nit: `error` is redundant

Wondertan (approved):
It's a much neater diff now! Thanks

renaynay (approved):

PR #2656, opened 2023-09-05
feat(share/store/cache): Split accessor cache into recent and blockstore

walldiss:
Adds MultiCache. Replaces eds store cache with multicache containing 2 separate caches:

- recently added edses (10 accessors, added on put only)
- edses requested by ipld (128 accessors, added on request to blockstore from bitswap server)


share/eds/cache/multicache.go
```diff
@@ -0,0 +1,51 @@
+package cache
+
+import (
+	"errors"
+
+	"github.com/filecoin-project/dagstore/shard"
+)
+
+// MultiCache represents a Cache that looks into multiple caches one by one.
```
Wondertan, line 9: [suggestion]
```suggestion
// MultiCache represents a Cache that looks into two caches one by one.
```

I would also rename to something like DoubleCache as Multi implies more than two

share/eds/store.go
```diff
@@ -284,7 +290,7 @@ func (s *Store) put(ctx context.Context, root share.DataHash, square *rsmt2d.Ext
 	go func() {
 		ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
 		defer cancel()
-		_, err := s.cache.GetOrLoad(ctx, result.Key, s.getAccessor)
+		_, err := s.cache.First().GetOrLoad(ctx, result.Key, s.getAccessor)
```
renaynay, line 293: comment above this go func can be modified

renaynay (commented):
lets call the file doublecache too

share/eds/store.go
walldiss: Could you please elaborate in what way it needs to be modified? It refers to the fact, that not any block, that is put into the store is actually recent, because it could be historical sampling block

share/eds/blockstore.go
```diff
@@ -146,28 +152,17 @@ func (bs *blockstore) getReadOnlyBlockstore(ctx context.Context, cid cid.Cid) (*
 		return nil, fmt.Errorf("failed to find shards containing multihash: %w", err)
 	}
 
-	// check if cache contains any of accessors
+	// check if any of two caches contains an accessor
```
distractedm1nd, line 155:
```suggestion
	// check if either cache contains an accessor
```

share/eds/store.go
```diff
@@ -114,7 +115,12 @@ func NewStore(basepath string, ds datastore.Batching) (*Store, error) {
...
 
-	accessorCache, err := cache.NewAccessorCache("cache", defaultCacheSize)
+	recentBlocksCache, err := cache.NewAccessorCache("recent", defaultRecentBlocksCacheSize)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create recent blocks cache: %w", err)
+	}
+
+	blockstoreCache, err := cache.NewAccessorCache("blockstore", defaultBlockstoreCacheSize)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create recent blocks cache: %w", err)
```
distractedm1nd, line 125:
```suggestion
		return nil, fmt.Errorf("failed to create blockstore cache: %w", err)
```

distractedm1nd (dismissed):

distractedm1nd:
Should we robusta test this before we merge into main?

walldiss:
I would rather run tests once all cache relateed PRs are merged. There are 2 more left, but they are relatedly simple

renaynay:
@walldiss then we should cut an RC once you get the next 2 up

Wondertan (approved):

share/eds/store.go
renaynay: Nevermind, I read it wrong

renaynay (approved):

PR #2661, opened 2023-09-06
Goreleaser

MSevey:
## Overview

**UPDATE**

This PR adds building binaries to the release process by using [goreleaser](https://goreleaser.com/).

The CI release process is largely unchanged. Releases can be trigger by either manually triggering the `ci_release` workflow, or they can be triggered by pushing a version tag. 

If the team has not been using the CI to generate releases, so I added a `make goreleaser-release` command to build the binaries for the release locally so that they can be manually uploaded. 

Result of running `make goreleaser-release`:
```
build
└── goreleaser
    ├── artifacts.json
    ├── celestia-node_Darwin_arm64.tar.gz
    ├── celestia-node_Darwin_x86_64.tar.gz
    ├── celestia-node_Linux_arm64.tar.gz
    ├── celestia-node_Linux_x86_64.tar.gz
    ├── celestia-node_darwin_amd64_v1
    │   └── celestia
    ├── celestia-node_darwin_arm64
    │   └── celestia
    ├── celestia-node_linux_amd64_v1
    │   └── celestia
    ├── celestia-node_linux_arm64
    │   └── celestia
    ├── checksums.txt
    ├── config.yaml
    └── metadata.json
```

The `.goreleaser.yaml` file is 90% stock generated from `goreleaser init`. The celestia node specific items are:
```yaml
builds:
  - main: ./cmd/celestia
    binary: celestia
    env:
      # NOTE: goreleaser doesn't fully support CGO natively. If CGO is needed
      # for any node features, this should be removed and a workaround might
      # need to be created.
      # REF: https://goreleaser.com/limitations/cgo/
      - CGO_ENABLED=0
      - VersioningPath={{ "github.com/celestiaorg/celestia-node/nodebuilder/node" }}
    goos:
      - linux
      - darwin
    goarch:
      - amd64
      - arm64
    ldflags:
      # Ref: https://goreleaser.com/customization/templates/#common-fields
      #
      # .CommitDate is used to help with reproducible builds, ensuring that the
      # same date is always used
      #
      # .FullCommit is git commit hash goreleaser is using for the release
      #
      # .Version is the version being released
      - -X "{{ .Env.VersioningPath }}.buildTime={{ .CommitDate }}"
      - -X "{{ .Env.VersioningPath }}.lastCommit={{ .FullCommit }}"
      - -X "{{ .Env.VersioningPath }}.semanticVersion={{ .Version }}"
dist: ./build/goreleaser
```

For building locally, i added a `make goreleaser-build` command. The binaries are put into a `build/goreleaser` directory. The `make goreleaser` command lists the `goreleaser` commands and also checks the version as a way to verify you have `goreleaser` installed.

Result of running `make goreleaser-build`:
```         
build
└── goreleaser
    ├── artifacts.json
    ├── celestia-node_darwin_amd64_v1
    │   └── celestia
    ├── config.yaml
    └── metadata.json
```

Successful github action run: https://github.com/MSevey/celestia-node/actions/runs/6123729144/job/16622244813
Example release generated: https://github.com/MSevey/celestia-node/releases

Created #2445 as a follow up discussion how to add signing.


.goreleaser.yaml
```diff
@@ -0,0 +1,66 @@
...
+    # You may remove this if you don't use go modules.
+    - go mod tidy
+builds:
+  - main: ./cmd/celestia
+    binary: celestia
+    env:
+      - CGO_ENABLED=0
+    goos:
+      - linux
+      - windows
```
liamsi, line 16: Very similar comments as for https://github.com/celestiaorg/celestia-app/pull/2437#pullrequestreview-1612900245
```suggestion
```

.goreleaser.yaml
```diff
@@ -0,0 +1,66 @@
...
+      - darwin
```
liamsi, line 17:
```suggestion
      - darwin
    goarch:
      - amd64
      - arm64
```

.goreleaser.yaml
MSevey: Same response, cc @renaynay and @Wondertan 
Any preferences on the final OS and Arch builds? 
These were just the defaults added by `init` so happy to update as the team wants.

liamsi (commented):
linter still failing. otherwise, lgtm

.goreleaser.yaml
```diff
@@ -0,0 +1,55 @@
+# This is an example .goreleaser.yml file with some sensible defaults.
+# Make sure to check the documentation at https://goreleaser.com
+before:
+  hooks:
+    - go mod tidy
```
Wondertan, line 5: Do we need it if we run it in a diff CI action?

Wondertan: I assume its from the default `goreleaser init` setup, but might not be needed.

.github/workflows/ci_release.yml
```diff
@@ -50,19 +50,48 @@ jobs:
...
+    if: |
+      github.event_name == 'workflow_dispatch' ||
+      (github.event_name == 'push' && contains(github.ref, 'refs/tags/'))
+    permissions: "write-all"
+    steps:
+      - uses: actions/checkout@v3
+      - run: git fetch --force --tags
+      - uses: actions/setup-go@v4
+        with:
+          go-version: 1.21
```
Wondertan, line 86: Is it possible to make this var? We have it in like 3-5 places in CI and changing it manually all the time is a PITA

Wondertan (approved):

.goreleaser.yaml
```diff
@@ -0,0 +1,55 @@
...
+builds:
+  - main: ./cmd/celestia
+    binary: celestia
+    env:
+      # NOTE: goreleaser doesn't fully support CGO natively. If CGO is needed
+      # for any node features, this should be removed and a workaround might
+      # need to be created.
+      # REF: https://goreleaser.com/limitations/cgo/
+      - CGO_ENABLED=0
+      - VersioningPath={{ "github.com/celestiaorg/celestia-node/nodebuilder/node" }}
```
ramin, line 15: `# yamllint disable-line rule:line-length`

.goreleaser.yaml
MSevey: This is also for when running locally, not just a CI config.

.github/workflows/ci_release.yml
MSevey: yea there are a number of ways you could make it a var. Happy to help in a follow up issue to clean up other places in the CI 👍

.goreleaser.yaml
MSevey: question for the team, the default line length for yamllint is 80, but we can set it to 120 via a config file. I'm pretty sure your other linters look at line length of 120 as well.

Also we set it to 120 here, but looks like this file doesn't get picked up even though it is a common workflow file. https://github.com/celestiaorg/.github/blob/main/.yamllint.yml

So I can either fix this one line, or add the yamllint config file which will help with future PRs as well. 

lmk what you prefer 👍

Wondertan: yamllint file is more preferable 👍

Wondertan (dismissed):

Wondertan:
A wild conflict arises

ramin:
@MSevey resolved the conflict, can you give it a glance and verify it's still looking good? Then let's do it

ramin (approved):

Wondertan (approved):

MSevey:
> @MSevey resolved the conflict, can you give it a glance and verify it's still looking good? Then let's do it

Thanks @ramin looks good 👍 

Waiting for CI to pass

PR #2667, opened 2023-09-07
Stale issue github workflow

ramin:
## Overview

Adds a basic stale issue and PR workflow. Will run one a day. I've only moderately changed the defaults so it is less aggressive on closing old issues. I also added recognition of a `keep-open` label that we can add add to an issue to....keep it open if we really love one of these old issues and consider it part of the team now

PRs will
- get a stale tag after 60 days
- be autoclosed 14 days after no additional activity

Issues
- get a stale tag after 90 days
- be autoclosed 21 days after no additional activity
- can be kept alive with `keep-open`

## Checklist

- [n/a] New and updated code has appropriate documentation
- [n/a ] New and updated code has new and/or updated testing
- [n/a] Visual proof for any user facing features like CLI or documentation updates
- [n/a] Linked issues closed with keywords


Wondertan:
How does it work for existing issues and PRs? Will it count the inactivity from when the workflow starts, or will it immediately mark all the old issues and PRs as stale?

ramin:
> How does it work for existing issues and PRs? Will it count the inactivity from when the workflow starts, or will it immediately mark all the old issues and PRs as stale?

@Wondertan it'll just start looking at everything. There IS a config to set a start-date param we can set which will only look at PRs and Issues created after the date. Shall we set that from today and it'll just sweep going forward?

https://github.com/actions/stale#start-date

Wondertan:
> Shall we set that from today and it'll just sweep going forward?
SGTM

distractedm1nd (approved):

Issue #2618, opened 2023-08-27
`GetSharesByNamespace` should get shares by reading them in the CAR file sequentially, rather than walking the IPLD


musalbas:
This will allow us to remove the CAR index and inverted index for shrex-ND, and will result in much faster shrex-ND store performance.

derrandz:
I've explored this a bit, and it seems that `GetSharesByNamespace` includes proofs for the shares it provides, and that requires traversing the tree down to determine the proof nodes of the retrieved shares.

With that being said, the inner nodes are of course also included in the CAR file, what I can't determine yet, is how to filter out only the proof nodes required for the filtered shares by namespace that have been retrieved from the CAR file.

I am probably overlooking something, is there a way to do this? cc @Wondertan @vgonkivs

derrandz:
Will give this a go and propose a PR

walldiss:
Actually shrex-ND doesn't use inverted_index, because it knows the required datahash from shrex-nd request. Implementing this is required to get rid of CAR index, but now enough. There also needs to be support for for bitswap to read shares without using CAR index. 
This one should be less of a priority compared to https://github.com/celestiaorg/celestia-node/issues/2479 (will allow to remove inverted_index) and pruning.

musalbas:
As part of https://github.com/celestiaorg/celestia-node/issues/2615#issuecomment-1696371526, this will allow us to remove CAR index for blocks outside of the storage window, because sampling doesn't need to be supported for blocks that are pruned by the network.

PR #2675, opened 2023-09-07
Shwap

Wondertan:
Prototype of the new storage for edses - edsfile and optimized roundtrip less ipldv2 protocol

TODO:
- edsfile
	- [x] Blockstore over EDSFile
	- [x] Support for Col sample reads
	- [x] ODS only mode
		- [ ] ~Support for Axis, Share and ShareWithProof~
	- [x] Measure size diff
	- [x] Delete EDS mode
	- [x] GetSharesByNamespace (assigned to @walldiss) 
	- [ ] File streaming (Write / Read)
	- [ ] Concurrency safety for file  
- edsStore	
	- [ ] Base implementation
	- [ ] Caching
	- [ ] Migrate shrex server to new storage
	- [ ] Store getter 
- ipldv2
	- [x] Define Sample type and refactor the protocol and the Hasher
	- [x] Migrate to proto
	- [x] Row Sampling
	- [ ] ~Versioning and backwards compatibility with old protocol~
	- [x] Use heights instead of DataHash to identify block 
	- [x] Hash requested Axis roots to shorten requests
	- [ ] Retriever over v2 (IPLDv2 reconstruction getter) (assigned to @walldiss)
	- [ ] Metrics 
	- [x] IPLDv2 row based getter
		- [x] Getter should support requesting batch of shares by idx at once 
		- [x] Store only samples by default
		- [x] Bitswap Sessions 
	- [ ] ~IPLDv2 Server over LN storage (blockservice)~ Will work as is 
	- [ ] Migrate actual sampling logic over GetShares #2984 
	
Stretch Goals:
* Range of shares Multihash(by extending SampleID with additional uint16 that defines the end idx of range)
* Blob requests. BlobID{DataID, Comm} -> Blob{BlobShares, ProofsToSubroot}
* Requesting proofs only for every request
* Reconsider DHT discovery.


derrandz (commented):
I love the direction this is taking, this is so good to look at.

I think this would better answer to the needs specified in https://github.com/celestiaorg/celestia-node/issues/2618 and would deprecate the gymnastics of reading from the car file done in https://github.com/celestiaorg/celestia-node/issues/2618

(_commenting for reference_)

Wondertan:
@derrandz, very true

share/eds/file.go
```diff
@@ -0,0 +1,171 @@
...
+
+	offset := int64(idx)*shrLn + HeaderSize
+	shr := make(share.Share, shrLn)
+	if _, err := f.fl.ReadAt(shr, offset); err != nil {
+		return nil, err
+	}
+	return shr, nil
+}
+
+func (f *File) ShareWithProof(idx int, axis rsmt2d.Axis) (share.Share, nmt.Proof, error) {
```
derrandz, line 120: [feature request]: Having `AxisWithProof` will be super useful for search operations to retrieve all rows with their proofs and roots to be able to filter out for specific shares and their proofs.

derrandz (commented):
a small feature request to make #2618 easier to implement

share/eds/file.go
Wondertan: Yep, wanted to make it

Wondertan: When you get the whole axis, there is no proof.

share/shwap/pb/shwap_pb.proto
```diff
@@ -0,0 +1,26 @@
+syntax = "proto3";
+
```
fl0rek, line 2:
```suggestion

package share.p2p.shwap;
```
Or was package specifier ommited for some reason?

Wondertan: It's not necessary from my understanding

fl0rek:
I don't think it's necessary but  other protobuf files specify the
namespace,so we expected this one too

share/shwap/pb/shwap_pb.proto
fl0rek: I don't think it's necessary, but other proto files specify the namespace so I expected this one too

Wondertan: Ok. Also, note that the name of the pkg is more like an implementation detail. The formal spec won't contain the pkg name in its proto file.

Wondertan:
We merged #3675, so can close this one

PR #2678, opened 2023-09-08
refactor(nodebuilder/header): get rid of `InitStore`

renaynay:
Not needed anymore

### DIFF
--- a/nodebuilder/header/module_test.go
+++ b/nodebuilder/header/module_test.go
@@ -36,8 +36,13 @@ func TestConstructModule_StoreParams(t *testing.T) {
 
 	app := fxtest.New(t,
 		fx.Supply(modp2p.Private),
-		fx.Provide(func() datastore.Batching {
-			return datastore.NewMapDatastore()
+		fx.Supply(modp2p.Bootstrappers{}),
+		fx.Provide(context.Background),
+		fx.Provide(libp2p.New),
+		fx.Provide(conngater.NewBasicConnectionGater),
+		fx.Provide(func() (datastore.Batching, datastore.Datastore) {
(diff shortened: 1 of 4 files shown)


nodebuilder/header/module_test.go
```diff
@@ -36,8 +36,13 @@ func TestConstructModule_StoreParams(t *testing.T) {
...
-		fx.Provide(func() datastore.Batching {
-			return datastore.NewMapDatastore()
+		fx.Supply(modp2p.Bootstrappers{}),
+		fx.Provide(context.Background),
+		fx.Provide(libp2p.New),
+		fx.Provide(conngater.NewBasicConnectionGater),
+		fx.Provide(func() (datastore.Batching, datastore.Datastore) {
+			ds := datastore.NewMapDatastore()
+			return ds, ds
 		}),
```
renaynay, line 46: FYI - not sure how this test worked without this to begin with..

distractedm1nd (approved):
utACK. Have you tested locally?

nodebuilder/header/module_test.go
Wondertan: Does it work without the change?

Wondertan (commented):
My intuition says me like there might be something subtle we miss by doing so. I clearly remember we had to do it, but don't remember why. 

Does it works properly on non and initialised node node? Also, the test change is suspicious

renaynay:
@Wondertan Yes works for inited and non-inited nodes.

Wondertan (approved):
It seems like there is no blocker anymore for this simplification.
Thanks for this @renaynay!

PR #2681, opened 2023-09-11
WIP: POC Storage Pruning

distractedm1nd:
Super simple storage pruning POC. Has already been rewritten a few times. It works under the following assumptions:
1. Only heights inside recency window are DASed
2. Upon registering height h, height (h - RecencyWindow) gets queued to be pruned <- this causes some edge cases, but is a super simple, readable mechanism

There are some edge cases this does not cover, but these known edge cases only occur on startup and could be fixed by some sort of cleanup routine that runs once some period after startup. Alternatively we can make a more complicated tracker, that catches the edge cases "correctly", but it would be really nice to keep the pruning mechanism simple.

This PR needs to be split up because it contains a lot of parts, some of which are unfinished:
- Fulls/Lights DASing only inside recency window
- RecencyWindow config
- Pruning on Bridge/Full nodes

Unfinished:
- Bridges still sync all heights
- RN, pruning is mandatory on bridge nodes the way the DI is set up.
- Pruning flag
- Mechanism to clean up edge cases
- DASer tests

Things it does not/will not handle:
- Optimizations using bloom/cuckoo filter
- Pruning of inverted indec

This PR needs to be split into many and is only here to start discussions/save as a checkpoint on the repo.

### DIFF
--- a/core/exchange.go
+++ b/core/exchange.go
@@ -11,6 +11,7 @@ import (
 	libhead "github.com/celestiaorg/go-header"
 	"github.com/celestiaorg/nmt"
 
+	"github.com/celestiaorg/celestia-node/das/pruner"
 	"github.com/celestiaorg/celestia-node/header"
 	"github.com/celestiaorg/celestia-node/share/eds"
 	"github.com/celestiaorg/celestia-node/share/ipld"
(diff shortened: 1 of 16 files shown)


musalbas:
Is there any easy way to define the recency window in terms of seconds instead of blocks? As we've seen previously, defining it in blocks where block times can be somewhat variable can cause unexpected behaviours.

distractedm1nd:
@musalbas Not really, because then downstream components will need to store the timestamp in the header rather than block height and things will get really messy for little benefit. A rough block count will be fine, we will use the upper bound for block time and so be storing a bit extra.

musalbas:
Isn't the timestamp in the header though?

PR #2697, opened 2023-09-11
feat(full/availability): adding different withholding options to tests

### LINKED ISSUE #2592: tests(availability): Withholding attacks in availability test suite
As discussed in #2560, the availability test suite needs to have accurate data withholding scenarios. These withholding scenarios will allow us to test the worst case reconstruction parameters as described in the paper.

distractedm1nd:
Closes #2592 by adding three different modes to full availability test nodes: FullyRecoverable, BarelyRecoverable, and Unrecoverable.

BarelyRecoverable test case requires 230 nodes for 99% chance of recoverability:
image

BarelyRecoverable is not used in DisconnectedFullNodes, only in ConnectedFullNodes, because as you can see by the graph by splitting the 230 node count in two we are not ensured failed recoverability.


share/availability/test/withholding_adder.go
```diff
@@ -0,0 +1,79 @@
...
+			block, err := ipld.GetLeaf(ctx, adder, root, idx, len(dah.RowRoots))
+			if err != nil {
+				return nil, fmt.Errorf("failure to get leaf: %w", err)
+			}
+			err = adder.DeleteBlock(ctx, block.Cid())
+			if err != nil {
+				return nil, fmt.Errorf("failure to delete block: %w", err)
+			}
+		}
+	}
```
ramin, line 77: it'd be nice to eventually be able to pass in some argument which was a handler that executed a "removal" strategy for setting up missing data. 

I'm probably missing something, but i'm imagining in the case where we have two Unrecoverable lightnode subnets, it'd be great to ensure BOTH sides were independently Unrecoverable, but once connected could provide samples to reconstruct WITHOUT hitting the source again. I think as it stands, both sides will be unrecoverable with identical missing ipld nodes?

not essential for this, but may be something that'll be nice to have as a utility for setting up more varied cases?

ramin (dismissed):
i had a good "in person" review of this that was convincing. I'm curious about the effectiveness of the "2 lightnode subnet split brain" test based on how the removed blocks are provisioned from Unrecoverable, but maybe i'm imagining a test that doesn't exist yet (ie: a full node transitions from failing to reconstruct from 1 subnet, to being able to with 2 subnets) but this might just be my limited grasp

share/availability/full/reconstruction_test.go
```diff
@@ -176,100 +255,141 @@ func TestShareAvailable_DisconnectedFullNodes(t *testing.T) {
...
+			errg.Go(func() error {
+				return full1.SharesAvailable(errCtx, root)
+			})
+			errg.Go(func() error {
+				return full2.SharesAvailable(errCtx, root)
+			})
+
+			// check that any of the fulls cannot reconstruct on their own
+			err := errg.Wait()
+			require.ErrorIs(t, err, share.ErrNotAvailable)
```
walldiss, line 365: Shouldn't it be that BOTH cannot reconstruct at this stage?

share/availability/full/reconstruction_test.go
```diff
@@ -176,100 +255,141 @@ func TestShareAvailable_DisconnectedFullNodes(t *testing.T) {
...
+			cancelErr()
+
+			// but after they connect
+			net.Connect(full1.ID(), full2.ID())
+
+			// with clean caches from the previous try
+			full1.ClearStorage()
+			full2.ClearStorage()
```
walldiss, line 373: Could you please add comment, why storage needs to be cleaned for reconstruction to happen?

share/availability/full/reconstruction_test.go
```diff
@@ -79,79 +118,119 @@ func TestShareAvailable_OneFullNode(t *testing.T) {
 // source, full node must be able to finish the availability process started in
 // the beginning.
 func TestShareAvailable_ConnectedFullNodes(t *testing.T) {
```
walldiss, line 120: I wonder, of this test is a subset of TestShareAvailable_DisconnectedFullNodes? It looks like it is just omitting part, when disconnected Full nodes cannot reconstruct and tests when they gets connected. Am I missing something or we can keep only TestShareAvailable_DisconnectedFullNodes?

share/availability/full/reconstruction_test.go
```diff
@@ -176,100 +255,141 @@ func TestShareAvailable_DisconnectedFullNodes(t *testing.T) {
...
+		origSquareSize  int  // k
+		lightNodes      int  // c
+		sampleAmount    uint // s
+		recoverability  Recoverability
+		expectedFailure bool
+	}{
+		{
+			name:            "fully recoverable",
+			origSquareSize:  16,
+			lightNodes:      32,
```
walldiss, line 269: In TestShareAvailable_ConnectedFullNodes this value is set to 24. Whats is the math behind it being different here?

share/availability/full/reconstruction_test.go
```diff
@@ -79,79 +118,119 @@ func TestShareAvailable_OneFullNode(t *testing.T) {
...
+			origSquareSize:  16,
+			lightNodes:      24,
+			sampleAmount:    20,
+			recoverability:  FullyRecoverable,
+			expectedFailure: false,
+		},
+		{
+			name:            "barely recoverable",
+			origSquareSize:  16,
+			lightNodes:      69,
```
walldiss, line 140: If you divide `int` 69 by 2 you will get 34, so total light nodes will be 68. So I assume lower value here should make this test flaky and the value needs to be 70, or need to use uneven devision for light nodes groups.

share/availability/full/reconstruction_test.go
```diff
@@ -79,79 +118,119 @@ func TestShareAvailable_OneFullNode(t *testing.T) {
...
+				require.NoError(t, err)
+			}
+		})
 	}
-
-	err := errg.Wait()
-	require.NoError(t, err)
 }
 
 // TestShareAvailable_DisconnectedFullNodes asserts that two disconnected full
```
walldiss, line 236: Random thought: Just curious, if it would still work same for similar topology, but for more than 2 full nodes.

share/availability/test/withholding_adder.go
```diff
@@ -0,0 +1,79 @@
...
+		return nil, fmt.Errorf("failure to create data availability header: %w", err)
+	}
+
+	// remove blocks from blockservice
+	for i := 0; i < withheldSize; i++ {
+		for j := 0; j < withheldSize; j++ {
+			// leave the last block in the last row and column if data should be left recoverable
+			if recoverable && i == withheldSize-1 && j == withheldSize-1 {
+				continue
+			}
```
walldiss, line 66: Optional: you might want save share with coordinates [0][0] instead, to make code a bit easier and start loops from 1, when eds is recoverable.

share/availability/full/reconstruction_test.go
```diff
@@ -26,49 +26,88 @@ func init() {
...
+			expectedFailure: false,
+		},
+		{
+			name:            "barely recoverable",
+			origSquareSize:  16,
+			lightNodes:      69,
+			sampleAmount:    20,
+			recoverability:  BarelyRecoverable,
+			expectedFailure: false,
+		},
```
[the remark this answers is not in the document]
distractedm1nd: Yes it would be, thanks

share/availability/full/reconstruction_test.go
distractedm1nd: Technically yes. I can change the check to be more explicit about this

share/availability/full/reconstruction_test.go
distractedm1nd: Its not exactly a subset but pretty close. I would agree that its kinda not necessary, but defer to @Wondertan for a decision on whether to keep it or not. I personally don't think it needs to exist

share/availability/full/reconstruction_test.go
distractedm1nd: This is an artifact from the old math, which was 16*2, but this is actually incorrect as it is more than necessary. Thanks for the catch

share/availability/full/reconstruction_test.go
distractedm1nd: Oh true, I set it to 70 in another case for this reason but forgot to apply everywhere

share/availability/full/reconstruction_test.go
distractedm1nd: It should, yes

share/availability/test/withholding_adder.go
distractedm1nd: great idea

share/availability/full/reconstruction_test.go
```diff
@@ -26,49 +26,88 @@ func init() {
...
-		net.Connect(lights[i].ID(), full.ID())
-	}
+			errg, errCtx := errgroup.WithContext(ctx)
+			errg.Go(func() error {
+				return full.SharesAvailable(errCtx, root)
+			})
 
-	err := errg.Wait()
-	require.NoError(t, err)
+			err := errg.Wait()
```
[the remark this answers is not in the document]
distractedm1nd: True, this is just an artifact from refactoring the test, thanks

share/availability/full/reconstruction_test.go
distractedm1nd: I am keeping it because it is exactly what we need for the BarelyRecoverable condition - it doesn't test unrecoverability with only half of the nodes

share/availability/test/withholding_adder.go
distractedm1nd: This ends up not working because it ignores the whole row/column, you still need the equality check in the innermost loop, so leaving it how it is

share/availability/test/testing.go
```diff
@@ -22,6 +22,17 @@ import (
 	"github.com/celestiaorg/celestia-node/share/sharetest"
 )
 
+// FillWithheldBS makes k + 1 columns of k + 1 rows unavailable if recoverable =
+// false, otherwise it acts the same but keeps the node at index (k + 1, k + 1).
+func FillWithheldBS(t *testing.T, n int, bServ blockservice.BlockService, recoverable bool) *share.Root {
+	shares := sharetest.RandShares(t, n*n)
+	eds, err := AddAndWithholdShares(context.TODO(), shares, bServ, recoverable)
```
renaynay, line 29: should we make this a context.WithCancel and then t.Cleanup the cancel

renaynay (approved):

share/availability/full/testing.go
```diff
@@ -26,10 +26,39 @@ func GetterWithRandSquare(t *testing.T, n int) (share.Getter, *share.Root) {
 	return getter, availability_test.RandFillBS(t, n, bServ)
 }
 
+type Recoverability int64
+
+const (
+	// FullyRecoverable makes all EDS shares available when filling the
+	// blockservice.
+	FullyRecoverable Recoverability = iota
+	// BarelyRecoverable withholds the (k + 1)^2 subsquare of the 2k^2 EDS,
```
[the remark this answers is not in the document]
distractedm1nd: I do not mean quadrants, look to the image in desc to see what I am referring to

share/availability/full/reconstruction_test.go
```diff
@@ -26,49 +27,99 @@ func init() {
...
+			origSquareSize:  16,
+			lightNodes:      19, // ~0.7% chance of recoverability
+			sampleAmount:    20,
+			recoverability:  FullyRecoverable,
+			expectedFailure: true,
+		},
+		{
+			name:            "barely recoverable",
+			origSquareSize:  16,
+			lightNodes:      230, // 99% chance of recoverability
```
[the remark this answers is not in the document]
distractedm1nd: Again, this case is not covered in the paper. Refer to my description, and I can send you the script so you can verify it for yourself

share/availability/test/withholding_adder.go
```diff
@@ -0,0 +1,79 @@
...
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failure to recompute the extended data square: %w", err)
+	}
+	// compute roots
+	_, err = eds.RowRoots()
+	if err != nil {
+		return nil, err
+	}
+	// commit the batch to ipfs
```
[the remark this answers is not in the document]
distractedm1nd: It was only copied in the first case because the intention was to use some internal state, but the most recent version just iterates over the adder and ipld Translates the DAH. 

So I will see if I can just call AddShares and do the iteration after

share/availability/full/testing.go
```diff
@@ -26,10 +26,39 @@ func GetterWithRandSquare(t *testing.T, n int) (share.Getter, *share.Root) {
 	return getter, availability_test.RandFillBS(t, n, bServ)
 }
 
+type Recoverability int64
```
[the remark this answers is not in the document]
distractedm1nd: Agree with this

share/availability/full/reconstruction_test.go
```diff
@@ -79,79 +130,133 @@ func TestShareAvailable_OneFullNode(t *testing.T) {
...
+			sampleAmount:    20,
+			recoverability:  FullyRecoverable,
+			expectedFailure: true,
+		},
+		// NOTE: This test contains cases for barely recoverable but
+		// DisconnectedFullNodes does not.  The reasoning for this is that
+		// DisconnectedFullNodes has the additional contstraint that the data
+		// should not be reconstructable from a single subnetwork, while this
+		// test only tests that the data is reconstructable once the subnetworks
+		// are connected.
```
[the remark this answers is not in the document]
distractedm1nd: Please refer to my description again to see - we also discussed this internally in slack. In order to have 99% confidence with BarelyRecoverable in DisconnectedFullNodes, you cannot use 2 subnetworks, as each subnetwork has a too high probability of the data being reconstructible. This is fine, as it is exactly what ConnectedFullNodes intends to test - that the data is available once connecting the full nodes, WITHOUT the assumption that it is unrecoverable from the independent subnetworks.

share/availability/full/reconstruction_test.go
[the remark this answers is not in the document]
Wondertan: The description didn't describe where 230 is coming from and thus the question. 

> Again, this case is not covered in the paper
Okay, then I missed some discussion around it because I don't remember such a fact. In my memory, figures from the paper assume withholding attack as it is extensively described in theorems in the same sections as all the plots I am referring to. 

Please share the script, I am curious about how it's constructed and would like to play and verify with it. The number 230 feels too much for me when 23 is enough to reconstruct the whole square. Did you verify the script with @musalbas

share/availability/full/reconstruction_test.go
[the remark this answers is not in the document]
Wondertan: Sorry if this is being repeated. I still would like to get to the bottom of this.

> In order to have 99% confidence with BarelyRecoverable in DisconnectedFullNodes, you cannot use 2 subnetworks, as each subnetwork has a too high probability of the data being reconstructible

Isn't that a sign that 230 is too big of a number of the 115 subnetwork has a very high chance of reconstructing? In other words, is there a number that would give 99% for subnetwork, but ~0% for the whole network? Is there a way to modify the script to support that? Note that my intuition for probabilities is not good.

share/availability/full/reconstruction_test.go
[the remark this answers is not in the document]
Wondertan: I meant 23/24 when data is not withheld, update the comment.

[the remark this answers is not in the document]
distractedm1nd: https://gist.github.com/distractedm1nd/d2b306da13581b97bbbcefaa5b050738
I apologize for the shitty code, didnt expect it to be read by others :)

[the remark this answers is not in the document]
walldiss: Main loop flow looks familiar 🙂  
If `ds.sample(s)` is acts as noop (for example if share is withheld), it would mean that given share is not sampled, but you count it as sampled anyway here:
```
for len(uniqueSamples) < shares {
	s := randomSample(k)
	uniqueSamples[s]++
	ds.sample(s)
}
```
Could you please clarify, why you count it is not sampled? Looks like this code could lead to less than 20 samples being assigned to each light node and make required amount of light nodes higher that it needs to be. 
It seems like code should be as following:
```
for len(uniqueSamples) < shares {
	s := randomSample(k)
	if ds.sample(s) {
		uniqueSamples[s]++
	}
}
```
				
On adjusted calculation algo, probability graph shows, that there is decent chance to do reconstruction for withholding case on amount of light node lower than 230:
image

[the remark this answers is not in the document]
distractedm1nd: Yes - I tried to match it with the current implementation that does not change the coordinates to sample after failure but just retries. Or is this incorrect?

share/availability/full/reconstruction_test.go
```diff
@@ -26,49 +27,99 @@ func init() {
...
+		origSquareSize  int  // k
+		lightNodes      int  // c
+		sampleAmount    uint // s
+		recoverability  Recoverability
+		expectedFailure bool
+	}{
+		{
+			name:            "fully recoverable",
+			origSquareSize:  16,
+			lightNodes:      24, // ~99% chance of recoverability
```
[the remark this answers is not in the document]
distractedm1nd: Good point, and we can adjust the LN count to be less at the threshold to decrease flakiness further

share/availability/full/reconstruction_test.go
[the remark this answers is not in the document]
walldiss: I've checked light availability and seems like it is acting like neither of simulations. 😆  I'll drop my finding into new issue to discuss further.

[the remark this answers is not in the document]
walldiss: https://github.com/celestiaorg/celestia-node/issues/2780

walldiss:
Closing because reconstruction tests will be fully reimplemented

PR #2699, opened 2023-09-12
refactor(swamp): use RPC client instead of service pointers

distractedm1nd:
Replaces #2356 
First step of #2337


nodebuilder/tests/api_test.go
```diff
@@ -13,11 +13,27 @@ import (
...
+
+	jwt, err := authtoken.NewSignedJWT(signer, []auth.Permission{"public", "read", "write", "admin"})
+	require.NoError(t, err)
+
+	client, err := client.NewClient(ctx, listenAddr, jwt)
+	require.NoError(t, err)
+
+	return client
+}
+
```
Wondertan, line 36: Eventually, we will abstract that away. E.g. `swamp.NewBridge() API` will give an API which is programmatic API or over RPC, depending on the node setup

Wondertan: The current UX of the node as a library is not there yet.

Wondertan (approved):

vgonkivs (approved):

ramin (approved):

PR #2705, opened 2023-09-14
fix(shrexeds): close stream

Wondertan:
If we receive a valid StatusMessage, we only CoseWrite side of the stream, but the stream interface states:
```
	// CloseWrite closes the stream for writing but leaves it open for
	// reading.
	//
	// CloseWrite does not free the stream, users must still call Close or
	// Reset.
	CloseWrite() error
```


Wondertan:
I am surprised we missed this somehow as well as no clue why this didn't show up in profiles/metrics

vgonkivs (approved):

renaynay (approved):

walldiss (approved):

PR #2706, opened 2023-09-15
refactor(cmd): extract rpc cmds

vgonkivs:
## Overview
Tested cmds against arabica-10 and mocha-4

## Checklist


cmd/celestia/internal/admin/admin.go
```diff
@@ -1,69 +1,59 @@
...
 
 import (
-	"context"
 	"errors"
 	"strings"
 
 	"github.com/filecoin-project/go-jsonrpc/auth"
 	"github.com/spf13/cobra"
+
+	"github.com/celestiaorg/celestia-node/cmd/celestia/internal"
```
ramin, line 10: I'd be in favor of aliasing this to something like `commands`. Internal is useful for restriction in go, but i'm not a fan of it being the package name. It may even make sense to rename things to `cmd/celestia/commands` is using internal wasn't totally deliberate, but i think the rest of the implementation would read much nicer as like `commands.RPCClient` or `commands.PrintOutput` vs `internal.PrintOutput`. Much more descriptive and nicer to read

cmd/celestia/internal/admin/admin.go
```diff
@@ -1,69 +1,59 @@
...
-	Args:  cobra.NoArgs,
-	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
-		rpcClient, err := newRPCClient(cmd.Context())
-		if err != nil {
-			return err
-		}
+	NodeCmd.PersistentFlags().StringVar(
+		&internal.RequestURL,
+		"url",
+		"http://localhost:26658",
```
ramin, line 19: i'm noticing some drift around defaults/settings and their usage. I think at a minimum we should 

a) move this to a const at the top vs inline
b) even better, create either commands/config/defaults.go or even a repo level config somewhere with a defaults.go

cmd/celestia/internal/rpc/das.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/celestiaorg/celestia-node/cmd/celestia/internal"
+)
+
+func init() {
+	DASCmd.AddCommand(samplingStatsCmd)
+
+	DASCmd.PersistentFlags().StringVar(
+		&internal.RequestURL,
+		"url",
+		"http://localhost:26658",
```
ramin, line 15: repeated in lots of places, hence we should consider a config/defaults.go SOMEWHERE, where do you think would be best @vgonkivs?

ramin (commented):
made a few organization suggestions, mostly around seeing some repeated stuff and for a nicer looking package name. We'll revisit after discussing/changing those. 

I'd also suggest maybe adding a internal/formatting/formatting.go package, and replacing internal.PrintOutput with something like commands/formatting -> `formatting.PrintOutput`

anything that reduces the need for maintain mental context around what internal is and where something might be coming from is my preference and having to "think upwards" to the import

cmd/celestia/internal/admin/admin.go
vgonkivs: I'm also not a big fan of internal, but in this particular case, I want to hide all cmds from other packages. What if I move all util.go to the separate package and call it `util` -> `util.PrintOutput`? I could also move there a const for `http://localhost:26658` or even instantiate the whole flag for the RequestURL, wdyt @ramin ?

cmd/celestia/internal/admin/admin.go
vgonkivs: I'd suggest creating a flag in one place and assigning it to all cmds

cmd/celestia/internal/admin/admin.go
ramin: yeah i think just alias-ing it and keeping the restriction from internal would be my preference.

on the util idea, i like that same with the const

Wondertan: What's the purpose of hiding cmds here?

Wondertan (changes requested):
We don't use internal anywhere yet, while we should. There are tons of internals we have that are supposed to be hidden away, and I don't think cmds are one of them. I propose to move the internalization to a separate issue and look at it holistically, what should be hidden in the whole project and what's not. We would also discuss there why we would wanna hide away cmds for our modules. Right now, I am not convinced that that's something we should do atm

cmd/celestia/internal/admin/admin.go
vgonkivs: The purpose of hiding is to restrict usage 🙂 .

vgonkivs:
Yes, we don't use internal anywhere in our node's code. But why can't we do it now for cmds? Anyway, this is a global refactoring of cmd package. Why can't we make it as it should be now and not in the future? I don't like when smth that is not related to the codebase could be used wherever. If cmds are supposed to be used only inside the `cmd/` why shouldn't they be hidden from other packages?

nodebuilder/node/cmds/node.go
```diff
@@ -1,32 +1,33 @@
...
-		rpcClient, err := newRPCClient(cmd.Context())
-		if err != nil {
-			return err
-		}
-
-		ctx := context.WithValue(cmd.Context(), rpcClientKey{}, rpcClient)
-		cmd.SetContext(ctx)
-		return nil
+		var err error
+		rpcClient, err = client.NewClient(cmd.Context(), client.RequestURL, "")
```
ramin, line 26: super nitpicky, but i think it'd make the rest of the commands look a LOT nicer if the naming here with one of either:

`rpc := client.NewClient(`

or we alias the import to be rpc then use the naming

`client := rpc.NewClient`

i'm not sure what i prefer, but either over rpcClient as the client var.

nodebuilder/share/cmds/share.go
```diff
@@ -190,7 +177,13 @@ var getEDS = &cobra.Command{
...
 		}
 
-		shares, err := client.Share.GetEDS(cmd.Context(), &root)
-		return printOutput(shares, err, nil)
+		shares, err := rpcClient.Share.GetEDS(cmd.Context(), &root)
+		return util.PrintOutput(shares, err, nil)
 	},
 }
+
+func parseJSON(param string) (json.RawMessage, error) {
```
ramin, line 185: imo it's fine to just move this up vs create the parseJSON function, makes the above command more readable, making the function either should live in util (i'm sure it'll happen again) or we just be verbose above

nodebuilder/state/cmds/state.go
```diff
@@ -30,24 +35,27 @@ func init() {
 	)
 }
 
-var stateCmd = &cobra.Command{
+var Cmd = &cobra.Command{
 	Use:   "state [command]",
 	Short: "Allows interaction with the State Module via JSON-RPC",
 	Args:  cobra.NoArgs,
+	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
```
ramin, line 42: i'm not sure how scoping would work, but wondering if would we could this PersistenPreRunE func somewhere else and return it from a util package somewhere?

nodebuilder/state/cmds/state.go
```diff
@@ -414,11 +369,20 @@ var queryRedelegationCmd = &cobra.Command{
...
 			cmd.Context(),
 			srcAddr.Address.(state.ValAddress),
 			dstAddr.Address.(state.ValAddress),
 		)
-		return printOutput(response, err, nil)
+		return util.PrintOutput(response, err, nil)
 	},
 }
+
+func parseAddressFromString(addrStr string) (state.Address, error) {
```
ramin, line 381: i've seen and used this in a few places as i've been looking around, again, i think there might be a compelling argument to create an `address.FromString(addr string) (state.Address error)` in the address package as part of this PR

api/rpc/client/client.go
```diff
@@ -22,6 +23,14 @@ var (
 	// staticClient is used for generating the OpenRPC spec.
 	staticClient Client
 	Modules      = moduleMap(&staticClient)
+
+	RequestURL string
+)
+
+const (
+	authEnvKey = "CELESTIA_NODE_AUTH_TOKEN"
```
ramin, line 31: this might be making go lint blow up, we have label this for the linter to ignore elsewhere, give it a try and see if the go-lint check stops acting foolish

ramin (approved):
i've made a few basic style and naming comments i think it'd be nice to make but aren't essential. I wasn't expecting the nodebuilder/<mod>/cmd as the refactor but really like it. I'd be fine if you ignore my name and organization things and merge, but think my ideas are great obviously 🤣

nodebuilder/state/cmds/state.go
vgonkivs: I thought about it, but it will end up with one more constructor that creates rpc client. I'm not sure that it is worth it

nodebuilder/state/cmds/state.go
vgonkivs: Yes, this makes sense. We can have a helper function somewhere in state module

api/rpc/client/client.go
vgonkivs: Locally everything is working for me lol

distractedm1nd (approved):
utACK, but looks really good, no nits

api/rpc/client/client.go
renaynay: just stick a //nolint:gosec there

api/rpc/client/client.go
```diff
@@ -65,8 +74,12 @@ func NewPublicClient(ctx context.Context, addr string) (*Client, error) {
 }
 
 // NewClient creates a new Client with one connection per namespace with the
-// given token as the authorization token.
+// given token as the authorization token. In case if token will be empty, then
+// `CELESTIA_NODE_AUTH_TOKEN` will be used in order to get the token.
 func NewClient(ctx context.Context, addr string, token string) (*Client, error) {
+	if token == "" {
+		token = os.Getenv(authEnvKey)
+	}
```
renaynay, line 82: This is a security issue, we cannot do this.

cmd/celestia/cmd_test.go
```diff
@@ -128,3 +128,23 @@ func TestBridge(t *testing.T) {
 			})
 	*/
 }
+
+func parseSignatureForHelpstring(methodSig reflect.StructField) string {
```
renaynay, line 132: let's not put a helper that's used in actual code in a test file

nodebuilder/state/cmds/state.go
renaynay, line 42: how does this operate with the `PersistentPreRunE` func inside of `util.go` ?

renaynay (changes requested):
* let's call all subdirectories `cmd` instead of `cmds` - more idiomatic

cmd/celestia/cmd_test.go
vgonkivs: we don't use it anywhere except current test

nodebuilder/state/cmds/state.go
vgonkivs: It does not. `util.PersistentPreRunE` operates only with full/light/bridge node cmds.

vgonkivs:
> just stick a //nolint:gosec there

I'm getting 
```
api/rpc/client/client.go:31:42: directive `//nolint:gosec` is unused for linter "gosec" (nolintlint)
	authEnvKey = "CELESTIA_NODE_AUTH_TOKEN" //nolint:gosec
```

cc @renaynay

api/rpc/client/client.go
vgonkivs: > just stick a //nolint:gosec there

I'm getting 
```
api/rpc/client/client.go:31:42: directive `//nolint:gosec` is unused for linter "gosec" (nolintlint)
	authEnvKey = "CELESTIA_NODE_AUTH_TOKEN" //nolint:gosec
```

api/rpc/client/client.go
vgonkivs: why?

cmd/celestia/util/rpc.go
```diff
@@ -0,0 +1,58 @@
+package util
```
Wondertan, line 1: This may also live in just `cmd`:
* It can be used outside of `celestia` binary, by something like `cel-shed`
* `cmd` contains reusable cmds and may as well contain reusable utils across out binaries

nodebuilder/blob/cmd/blob.go
```diff
@@ -96,28 +101,31 @@ var getAllCmd = &cobra.Command{
...
+		namespace, err := util.ParseV0Namespace(args[1])
 		if err != nil {
 			return fmt.Errorf("error parsing a namespace:%v", err)
 		}
 
 		blobs, err := client.Blob.GetAll(cmd.Context(), height, []share.Namespace{namespace})
-
+		fmt.Println(hex.EncodeToString(blobs[0].Namespace().ID()))
+		fmt.Println(blobs[0].Namespace().ID())
+		fmt.Println(blobs[0].Namespace())
```
Wondertan, line 123: [q]
Debug leftover?

cmd/rpc.go
```diff
@@ -0,0 +1,58 @@
+package cmd
```
Wondertan, line 1: [discussion]
Another idea is to decouple `cmd/binary` from cli/cmd utils. Basically, move all the `cmd/...go` files to `cli` or `lib/cli`, then `cmd` would only be a place for our main apps/entrypoints. 
Alternatively, we can move them to `nodebuilder/node/cmd`. I just looked at we have there and all the stuff there are node/admin related cmds - start, auth, utils for them, etc. So it logically makes sense to keep it there.
Generally, I like the idea of not keeping the `cmd` pkg with 0 code and only subpkgs with binaries, which is also idiomatic.

nodebuilder/blob/cmd/blob.go
```diff
@@ -50,28 +53,30 @@ func init() {
 	)
 }
 
-var blobCmd = &cobra.Command{
-	Use:   "blob [command]",
-	Short: "Allows to interact with the Blob Service via JSON-RPC",
-	Args:  cobra.NoArgs,
+var Cmd = &cobra.Command{
```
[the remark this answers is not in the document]
vgonkivs: The reason why I named it `Cmd` is that inside cmd/celestia/rpc.go I use ALL of them, so I had to add an alias there and it looks like `blob.Cmd` , `das.Cmd`etc

nodebuilder/blob/cmd/blob.go
```diff
@@ -1,14 +1,16 @@
...
 	"encoding/base64"
+	"encoding/hex"
 	"fmt"
 	"reflect"
 	"strconv"
 
 	"github.com/spf13/cobra"
 
 	"github.com/celestiaorg/celestia-node/blob"
+	util "github.com/celestiaorg/celestia-node/cmd"
```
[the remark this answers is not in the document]
vgonkivs: okay, let me use cmdnode everywhere

cmd/rpc.go
```diff
@@ -0,0 +1,58 @@
...
+
+type rpcClientKey struct{}
+
+func ParseClientFromCtx(ctx context.Context) (*rpc.Client, error) {
+	client, ok := ctx.Value(rpcClientKey{}).(*rpc.Client)
+	if !ok {
+		return nil, errors.New("rpc client was not set")
+	}
+	return client, nil
+}
```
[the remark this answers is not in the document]
vgonkivs: We don't have an intermediate cmd rn. It means that all our cmds have to do the same. So, IMO, it is better to keep everything in the `InitiClient`

cmd/rpc.go
vgonkivs: Agree, but let's do it in a subsequent PR.

cmd/rpc.go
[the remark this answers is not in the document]
Wondertan: What do you mean by intermediate cmd?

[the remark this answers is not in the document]
vgonkivs: previously we had a rpc cmd and all module cmds were a sub-cmds of it. I've removed it and now module cmds are sub-cmd of the root cmd:
```
celestia blob 
celestia das
```

renaynay (approved):

ramin (approved):

cmd/rpc.go
vgonkivs: https://github.com/celestiaorg/celestia-node/issues/2754

Wondertan (approved):

Issue #2672, opened 2023-09-07
CI: check that openrpc file matches current


adlerjohn:
For each PR push, check that the openrpc file matches the current. Should target https://github.com/celestiaorg/celestia-node/blob/openrpc-spec unless that branch is merged beforehand.

https://github.com/celestiaorg/celestia-node/blob/openrpc-spec/openrpc.json

PR #2717, opened 2023-09-18
openrpc.json: Detect drift in openrpc spec

ramin:
## Overview

closes: #2672 

Adds a simple ci check to generate openrpc.json file, and fail if it believes there are updates that are not present in the current openrpc.json file.

comments:

- should we simply add to `go-ci.yml`? I decided not to as it felt like different work that deserved it's own workflow
- possible we could add the `git diff` verify commands to the Makefile, yet this all felt ci specific and not worth polluting Makefile for, leaving the yml ugly vs the Makefile bloated.
- the target branch is hella out of date (go 1.19) we should think about getting main rebased in pretty soon @vgonkivs @distractedm1nd 

## Checklist

- [n/a] New and updated code has appropriate documentation
- [n/a] New and updated code has new and/or updated testing

### DIFF
--- a/openrpc.json
+++ b/openrpc.json
@@ -1,8569 +1,7685 @@
+--> Generating OpenRPC spec
 {
-  "openrpc": "1.2.6",
-  "info": {
-    "title": "Celestia Node API",
-    "description": "The Celestia Node API is the collection of RPC methods that can be used to interact with the services provided by Celestia Data Availability Nodes.",
-    "version": "v0.2.2"
-  },
-  "externalDocs": {
-    "description": "Celestia Node GitHub",
-    "url": "https://github.com/celestiaorg/celestia-node"
-  },
-  "methods": [
-    {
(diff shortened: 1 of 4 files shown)


renaynay (approved):
Nice one @ramin !

renaynay:
I would defo add this to go-ci.yaml btw @ramin, I think we should run it on every PR - it's not a big check.

openrpc.json
```diff
@@ -1,8569 +1,7685 @@
+--> Generating OpenRPC spec
```
distractedm1nd, line 1: This will cause problems, can you update the Makefile to remove this echo?

distractedm1nd (changes requested):
Excellent, thank you for this

We can delete the old branch since we will have the new file in main

openrpc.json
ramin: removed

PR #2723, opened 2023-09-18
feat(share/eds/cache): Add accessor cache refs counter

walldiss:
Adds refs counting to accessor cache, to prevent situations when underlying accessor is closed while there are readers that uses it.


blob/service.go
```diff
@@ -51,7 +51,8 @@ func NewService(
 	}
 }
 
-// SubmitOptions contains the information about fee and gasLimit price in order to configure the Submit request.
+// SubmitOptions contains the information about fee and gasLimit price in order to configure the
+// Submit request.
```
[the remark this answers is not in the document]
walldiss: It is not formatted and changed by `make fmt`

share/eds/cache/accessor_cache.go
```diff
@@ -55,6 +64,47 @@ func (s *accessorWithBlockstore) Reader() io.Reader {
...
+	if s.refs.Add(-1) <= 0 {
+		close(s.done)
+	}
+}
+
+func (s *accessorWithBlockstore) close() error {
+	s.Lock()
+	s.isClosed = true
+	done := s.done
+	s.Unlock()
```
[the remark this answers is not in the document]
walldiss: Because it could case race condition, when `close()` will resolve by closed done channel, which could allow closure of underlying accessore. While in parallel new ref could be created by `addRef`, making underlying accessor being in use.
So  `isClosed` is needed to prevent new ref, after accessor closure has been initialised.

share/eds/cache/accessor_cache.go
```diff
@@ -134,13 +187,27 @@ func (bc *AccessorCache) GetOrLoad(
...
+	lk.Lock()
+	accessor, err := bc.get(key)
+	lk.Unlock()
+	if errors.Is(err, errCacheMiss) {
+		// item is not in cache
+		return nil
+	}
+	if err = accessor.close(); err != nil {
+		return err
+	}
```
renaynay, line 210: do you need to close accessor here? doesn't it happen in evictFn?

share/eds/cache/accessor_cache.go
```diff
@@ -55,6 +64,47 @@ func (s *accessorWithBlockstore) Reader() io.Reader {
...
+		s.done = make(chan struct{})
+	}
+	return nil
+}
+
+func (s *accessorWithBlockstore) removeRef() {
+	s.RLock()
+	defer s.RUnlock()
+	if s.refs.Add(-1) <= 0 {
+		close(s.done)
```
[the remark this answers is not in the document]
walldiss: No, it needs to be closed to unblock any current or future readers

share/eds/cache/accessor_cache.go
walldiss: it will be closed inside `close()` call

share/eds/cache/accessor_cache.go
```diff
@@ -55,6 +64,47 @@ func (s *accessorWithBlockstore) Reader() io.Reader {
...
+	}
+}
+
+func (s *accessorWithBlockstore) close() error {
+	s.Lock()
+	s.isClosed = true
+	done := s.done
+	s.Unlock()
+
+	// TODO: add closing metrics
```
distractedm1nd, line 95: can you create an issue for this

share/eds/cache/accessor_cache.go
distractedm1nd: I think she means why close it here when it will get close() called in evictFn inside of bc.cache.Remove

distractedm1nd (dismissed):
Beautiful, just have the one question that Rene asked

Let's get this merged so we can test it in Robusta

share/eds/cache/accessor_cache.go
renaynay: Yes

walldiss: This close is sync and will yield error to client if it happens. While evict will call close() async without any error returned

renaynay: But doesn't it then get called twice?

walldiss: It will be called second time in evict, but there is no way around it. I've allowed subsequent calls of close() to be noop.

Wondertan (approved):

distractedm1nd (approved):

PR #2724, opened 2023-09-18
feat(share/eds/store): expose eds store params

walldiss:
Adds exported params to allow control of eds store configuration


share/eds/store.go
```diff
@@ -75,8 +67,12 @@ type Store struct {
 }
 
 // NewStore creates a new EDS Store under the given basepath and datastore.
-func NewStore(basepath string, ds datastore.Batching) (*Store, error) {
-	err := setupPath(basepath)
+func NewStore(params *Parameters, ds datastore.Batching) (*Store, error) {
```
Wondertan, line 70: Uber not: it much nicer when required param is required by the constructor. Path is still required via Validate, but it's api make it look like an option, which is confusing

distractedm1nd (dismissed):

share/eds/store.go
walldiss: Params validation check, that happens inside NewStore will ensure that param is set.  API is not an option, it is field setter function. I choose it this way to allows to set path in one line. Could be exported field instead, but I feel it is nicer to have this helper setter instead.

distractedm1nd: Do you mean path shouldn't be part of Paramus?

walldiss: Don't you think It feels weird to have it as component in constructor, when there is params containing all config variables?

Wondertan: Thats totally normal to have required params as constructor params and config/params with optional/defaultable ones

walldiss: if you have multiple required params the constructor will look like monster. It just feels cleaner to have all params aggregated in one struct, even tho it defer values check from compile time to run time. Badger constructor could be an example.
I don't have strong opinion about it, so up to you.

Wondertan: You are confusing the required fields and those that have defaults. Fields that have defaults are not required. In our EdsStore case only one two are really required, which is path and datastore.

Wondertan (approved):

distractedm1nd (approved):

PR #2726, opened 2023-09-18
refactoring(full/availability): removing TeeGetter, storing via SharesAvailable

distractedm1nd:
We never needed TeeGetter, full availability already has access to the store. This is also needed for pruning


Wondertan (approved):

share/availability/full/availability.go
```diff
@@ -80,7 +87,15 @@ func (fa *ShareAvailability) SharesAvailable(ctx context.Context, root *share.Ro
...
 		}
 	}
-	return err
+
+	if fa.store != nil {
+		err = fa.store.Put(ctx, root.Hash(), eds)
+		if err != nil && !errors.Is(err, dagstore.ErrShardExists) {
+			return fmt.Errorf("full availability: failed to store eds: %w", err)
+		}
+	}
```
walldiss, line 96: What is the case when `fa.store == nil`? It looks like misconfigured module, but I might miss something

distractedm1nd: For testing etc

walldiss: There are no tests that utilises it :) If we don't use it, maybe remove the possibility for store operation to be skipped, because in this case we would not even notice it happened. 

Also nice to have would be to add unit test, covering storing logic.

distractedm1nd: All full availability tests pass a nil store to NewShareAvailability via TestAvailability

distractedm1nd: I've refactored the tests to include it and removed the nil checks inside of availability

Wondertan (approved):

walldiss (dismissed):

renaynay (approved):

walldiss (dismissed):

renaynay (dismissed):
nice 1

walldiss (approved):

renaynay (approved):
ok

PR #2727, opened 2023-09-18
feat(daser): only DAS recent heights 

distractedm1nd:
I have extracted the FN/LNs sampling only inside the recency window from the POC. 

This targets the new pruning feature branch, as it starts introducing pruning config values and is the base for the next steps. 

It is not too bad if a few headers outside of the window are sampled first, in case I am missing anything. 

Special call to @walldiss to make sure there isn't a better way without introducing too much complexity.


distractedm1nd:
put in draft while we play around with timestamp based pruning

distractedm1nd:
Closing as superseded by #2738

Issue #2578, opened 2023-08-17
share/discovery: Implement versioning for `full` advertising string


renaynay:
This would ensure discovery will be optimised to discover most compatible `fulls`.

Related to #2548

renaynay:
Done already

PR #2730, opened 2023-09-19
feat(discovery): discover peers by tag

walldiss:
Allow to run multiple discovery subcomponents, where each can find new peers based on preset `tag` and notify its subscribers. It could allow any sort of separate discovery subcomponents (versioned, pruned/archived, etc). Essentially it allows discovery to be abstracted away and act as single function component.

Involves some minor refactoring here and there just to clean up things.

Resolves https://github.com/celestiaorg/celestia-node/issues/2578


share/p2p/discovery/discovery.go
```diff
@@ -133,7 +133,7 @@ func (d *Discovery) Discard(id peer.ID) bool {
 		return false
 	}
 
-	d.host.ConnManager().Unprotect(id, rendezvousPoint)
```
ramin, line 136: i think we can now remove this `rendezvourPoint` var up top?

ramin (approved):
only basic understanding of how this all works so take my review with grain of salt, but beautiful ❤️

nodebuilder/share/constructors.go
```diff
@@ -17,18 +17,20 @@ import (
...
 	return func(
 		r routing.ContentRouting,
 		h host.Host,
-	) *disc.Discovery {
+		manager *peers.Manager,
+	) (*disc.Discovery, error) {
 		return disc.NewDiscovery(
+			cfg,
 			h,
 			routingdisc.NewRoutingDiscovery(r),
```
distractedm1nd, line 33: I believe you forgot to add the WithOnPeersUpdate? If not, I can't find where it is being added

share/p2p/discovery/options.go
```diff
@@ -14,45 +21,56 @@ type Parameters struct {
...
 		AdvertiseInterval: time.Hour,
+		//TODO: remove fullNodesTag default value once multiple tags are supported
+		Tag: fullNodesTag,
 	}
 }
 
 // Validate validates the values in Parameters
 func (p *Parameters) Validate() error {
-	if p.AdvertiseInterval <= 0 {
+	if p.Tag == "" {
```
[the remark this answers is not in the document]
walldiss: Because it was incorrect. `AdvertiseInterval` == -1 disables it. We didn't caught it because config validation was called outside of the constructor, which happen to be before options were applied. So it was a bug

nodebuilder/share/constructors.go
walldiss: good catch

share/p2p/discovery/discovery_test.go
```diff
@@ -27,23 +27,34 @@ func TestDiscovery(t *testing.T) {
...
+	params.Tag = "full"
+
+	peerA := tn.discovery(params, host, routingDisc,
+		WithOnPeersUpdate(submit),
+	)
 
+	params = &Parameters{
+		PeersLimit:        0,
+		AdvertiseInterval: time.Millisecond * 100,
+		Tag:               "full",
```
renaynay, line 52: could just use fullNodesTag

share/p2p/peers/manager.go
```diff
@@ -248,6 +227,22 @@ func (m *Manager) Peer(
 	}
 }
 
+// UpdatedFullNodes is called by discovery when new full node is discovered or removed
+func (m *Manager) UpdatedFullNodes(peerID peer.ID, isAdded bool) {
```
renaynay, line 231: can we call this `UpdateFullNodePool` or something?

renaynay (approved):

distractedm1nd (dismissed):

distractedm1nd (approved):

distractedm1nd (approved):

distractedm1nd:
i didnt know i could double approve like that lol

share/p2p/discovery/options.go
```diff
@@ -14,45 +21,56 @@ type Parameters struct {
...
 // DefaultParameters returns the default Parameters' configuration values
 // for the Discovery module
-func DefaultParameters() Parameters {
-	return Parameters{
+func DefaultParameters() *Parameters {
+	return &Parameters{
 		PeersLimit:        5,
 		AdvertiseInterval: time.Hour,
+		//TODO: remove fullNodesTag default value once multiple tags are supported
+		Tag: fullNodesTag,
```
Wondertan, line 44: [blocking]
This will break our configs and will be removed anyway. Let's simply set `fullNodesTag` if the Tag is empty with the same todo that this should change once we introduce support for multiple

Wondertan (commented):
Ok, so this is pretty simple. For multiple discoveries we will need something like DiscoveryManager for Discovery CRUD

share/p2p/discovery/options.go
walldiss: Sounds good. I've moved tag to constructor, so it is not optional anymore and have to be set to non-zero value. So there is no need for this TODO anymore.  
Also cleaned up a bit params validation, because it looked bad that some params values were used to only configure tests. For example PeerLimit == 0 used to disable discovery scanning, when it was possible to just skip call `Start`.  This will achieve same result without weird params usage. Same for AdvertiseInterval == -1, when it was possible to just skip `Advertise` call to disable advertising.

renaynay (approved):

PR #2738, opened 2023-09-20
feat(storage/pruning): Timestamp based StoragePruner

distractedm1nd:
This PR replaces #2681, implementing a POC for timestamp based pruning. 

# SCOPE:
## Config
Pruning module is introduced to nodebuilder. I made this decision because there are multiple modules that need to read from its config (DASer and Core), so this is the cleanest way.

Config values are:
- PruningEnabled
- RecencyWindow
- EpochDuration

## StoragePruner
The StoragePruner works by putting each height into a bucket ("epoch") based off of the unix timestamp. Every EpochDuration, the hashes stored in that bucket are pruned. If there are more buckets than there should be even after the oldest epoch is pruned, gc is run that prunes all buckets that are too old (this covers startup after a longer period of inactivity for example). This PR includes some basic unit tests and metrics.

# Next stacked PRs will include:

## Flags
`--archival-mode` is introduced for bridges and fulls to disable pruning.

## DASer
Light Nodes only das below the RecencyWindow, and Full Nodes only download blocks inside of the RecencyWindow, unless PruningEnabled is false.

## Core
Exchange and Listener can now take a pruner via an ExchangeOption/ListenerOption that is only set if PruningEnabled = true, similar to the DASer. Headers are registered on the pruner before their calls to storeEDS. Note that the Exchange does not yet not save heights that are too old - every block will be saved, but they will be cleaned up


nodebuilder/pruner/pruner.go
```diff
@@ -0,0 +1,239 @@
...
+)
+
+var log = logging.Logger("pruner")
+
+type StoragePruner struct {
+	cancel context.CancelFunc
+	cfg    Config
+
+	// TODO: Race?
+	oldestEpoch uint64
```
ramin, line 32: i initially thought this was a way we could ship new versions with a known "prune before" timestamp/height, which i realize it isn't, but i wonder if we can/should (ie: like we ship the genesis config, we can set a "pruneBefore" for blocks 30 days before release so new nodes don't have the recency window issue)

nodebuilder/pruner/flags.go
```diff
@@ -0,0 +1,36 @@
+package pruner
+
+import (
+	"github.com/spf13/cobra"
+	flag "github.com/spf13/pflag"
+)
+
+var (
+	archivalModeFlag = "archival-mode"
```
ramin, line 9: personally think explicitly calling it "pruning" or "prune-mode" is clearer, than archival. To me, archival suggests the opposite (everything is archived here, so NOT pruned and guaranteed)

nodebuilder/pruner/pruner.go
```diff
@@ -0,0 +1,239 @@
...
+
+	stripedLocks [256]sync.Mutex
+	activeEpochs map[uint64]struct{}
+
+	ds    datastore.Batching
+	store *eds.Store
+
+	done chan struct{}
+}
```
ramin, line 41: general and only implementation comment: seeing the act of pruning via running the gc should/can be "background", it might be nice to put some sort of throttling/max worker/rate limit on how many blocks it tries to prune at any one run of the gc. Imagining a scenario where, for whatever reason, we get a big list of epochs to catch up on (not sure why, but lets just assume for some reason we do - maybe some resource pauses). Then a gc catchup run cant possibly ever degrade resources it touches (thinking deletes, badger, disk) by going ham in a loop when running through a big range.

might be nice to make this configurable too and test running prunung with resources constrained and observe out what amount of backpressure it could takes and what happens.

renaynay: I can see this potetnially happening on all the synced blocks that are synced for `sampling_window` prior

nodebuilder/pruner/pruner.go
```diff
@@ -0,0 +1,308 @@
...
+	results, err := sp.ds.Query(ctx, query.Query{})
+	if err != nil {
+		return fmt.Errorf("failed to recover pruner state from datastore: %w", err)
+	}
+	for {
+		res, ok := results.NextSync()
+		if !ok {
+			break
+		}
+		epoch, err := strconv.ParseUint(res.Key[1:], 10, 64)
```
renaynay, line 93: sanity check on res / res.key len?

nodebuilder/pruner/pruner.go
```diff
@@ -0,0 +1,308 @@
...
+	lk := &sp.stripedLocks[epoch%256]
+	lk.Lock()
+	defer lk.Unlock()
+	log.Debugf("registering datahash %X to epoch %d", h.DAH.Hash(), epoch)
+	_, ok := sp.activeEpochs[epoch]
+	if ok { // epoch already registered, load existing datahashes from datastore
+		datahashes, err = sp.getDatahashesFromEpoch(ctx, epoch)
+		if err != nil {
+			return err
+		}
```
renaynay, line 129: can you format all of this a bit more nicely (code blocking etc)

and short-circuit statement can be first.

nodebuilder/pruner/pruner.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+
+	epoch := sp.calculateEpoch(h.Time())
+	lk := &sp.stripedLocks[epoch%256]
+	lk.Lock()
+	defer lk.Unlock()
+	log.Debugf("registering datahash %X to epoch %d", h.DAH.Hash(), epoch)
+	_, ok := sp.activeEpochs[epoch]
+	if ok { // epoch already registered, load existing datahashes from datastore
+		datahashes, err = sp.getDatahashesFromEpoch(ctx, epoch)
```
renaynay, line 126: @distractedm1nd 

when you fetch datahashes from epoch, technically when register is called, it should be called for roots on the one specific active epoch, which can be derived from the current time. Why can't you hold all of those in memory instead of doing read / overwrites from *potentially* disk on a per-hour basis. That's ~250 hashes per hour which isn't much (unless you're syncing in which case that's 30 days worth of roots). Just something to think about and maybe discuss - I know you're using ds.Batching here.

nodebuilder/pruner/pruner.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+
+	oldest := sp.oldestEpoch.Load()
+	if epoch < oldest {
+		sp.oldestEpoch.CompareAndSwap(oldest, epoch)
+	}
+
+	datahashes = append(datahashes, h.DAH.Hash())
+	sp.metrics.observeRegister(ctx)
+	return sp.saveDatahashesToEpoch(ctx, epoch, datahashes)
```
renaynay, line 143: shouldn't you call observeRegister after successful completion of saving ?

nodebuilder/pruner/pruner.go
```diff
@@ -0,0 +1,308 @@
...
+}
+
+func (sp *StoragePruner) gc(ctx context.Context) {
+	log.Infow("epoch count exceeded max; starting gc")
```
renaynay, line 147: make it more explicit that you're starting pruning job.

nodebuilder/pruner/pruner.go
```diff
@@ -0,0 +1,308 @@
...
+func (sp *StoragePruner) calculateEpoch(timestamp time.Time) uint64 {
+	return uint64(timestamp.Unix() / int64(sp.cfg.EpochDuration.Seconds()))
+}
+
+func (sp *StoragePruner) epochToTime(epoch uint64) time.Time {
+	return time.Unix(int64(epoch*uint64(sp.cfg.EpochDuration.Seconds())), 0)
+}
+
+func (sp *StoragePruner) epochIsRecent(epoch uint64) bool {
+	return epoch >= sp.calculateEpoch(time.Now().Add(-sp.cfg.RecencyWindow))
```
renaynay, line 307: maybe we can eventually do this by # of active epochs.

renaynay (commented):
bunch of comments, thanks for splitting up the PRs 

One thing I also want to mention is that we should reduce (where possible) our reliance on time for recency estimation + epoch determination. There's ways to get around this just based on block time (except for the edge case of an initial sync which is annoying to handle for). Maybe it's worthwhile to discuss that a bit further, up to you @distractedm1nd

nodebuilder/pruner/flags.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+func Flags() *flag.FlagSet {
+	flags := &flag.FlagSet{}
+
+	flags.Bool(
+		archivalModeFlag,
+		false,
```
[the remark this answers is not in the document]
distractedm1nd: This shouldnt even be in this PR so oops

nodebuilder/pruner/metrics.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+	"context"
+
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/metric"
+)
+
+var (
+	meter = otel.Meter("storage_pruner")
```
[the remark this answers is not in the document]
distractedm1nd: should I rename the component as well too then?

nodebuilder/settings.go
```diff
@@ -117,6 +119,7 @@ func WithMetrics(metricOpts []otlpmetrichttp.Option, nodeType node.Type) fx.Opti
 			baseComponents,
 			fx.Invoke(share.WithStoreMetrics),
 			fx.Invoke(share.WithShrexServerMetrics),
+			fx.Invoke(pruner.WithStoragePrunerMetrics),
```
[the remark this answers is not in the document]
distractedm1nd: Why question?

nodebuilder/pruner/pruner.go
distractedm1nd: This is to keep it persistent for unexpected restarts or shutdowns. With a flushing from memory to disk on shutdown approach it is a lot less robust and we will need more recovery logic. I find this way to be simplest and don't think there are really any bottlenecks we have to worry about unless we had solana level blocktimes.

nodebuilder/pruner/pruner.go
distractedm1nd: Number of epochs isnt going to tell you if an epoch is recent or old enough to be pruned. The only thing number of epochs helps with is determining whether GC should run or not

nodebuilder/pruner/metrics.go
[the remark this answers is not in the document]
renaynay: wdym ? 

Yes i think `block_pruning` is more accurate.

nodebuilder/pruner/pruner.go
renaynay: The number of the epoch could help indicate whether it's old or not (e.g. smallest number in epoch)

I'm fine for this being the implementation currently and then we can think of moving away from time-based as it's a bit more brittle. I don't think switching the method for epoch determination would be super breaking.

walldiss:
As I understand, all headers are grouped into epochs that are stored persistently on the disk. Then there is a background process that runs once per epoch time (1 hour), pruning the latest epoch. Additionally, there is a garbage collection (gc) process that kicks in when there is a gap for more than 1 epoch that needs to be pruned.

I have a few questions:
- What is the reason for batched pruning? Would it create lots of work for store at the end of hour interval?
- Would it be less of a strain on the system if blocks are pruned as they fall out of the recency window? So there is one block pruned in blockTime.
- Why do we need to store headers to epochs if they are already available in the header store and could be accessed directly?

It looks kind of complex to me, but I might be missing something. Here is my alternative algorithm suggestion:
1. On startup:
   - Load the last max pruned header height (tip of the pruned chain).
   - Find and store the height of the header that is the oldest one within the recency window. This header will expire next.
   - Prune headers between the latest pruned and the one that is the oldest within the recency window.
2. Runloop:
   - We have the time when this header was produced, so we can wait until it is out of the recency window.
   - Prune headers between the latest pruned and the one that just fell out of the recency window.
   - Store the max pruned.
   - Find the new oldest within the recency window.

Benefits of this solution:
- No epochs calculations.
- We would not need to store epochs to disk.
- Pruning could be done on an event-based approach as well as batching with any batch size. With epochs, the batch is strictly epoch size, and an event-based approach is not possible.
- No secondary gc needed.
- Information about pruned/not pruned height is explicitly stored and could be accessed and visible.

But I might be missing why epochs were introduced, so I will be happy to discuss.

renaynay:
# Notes from call 23.11.23

* Pursue Vlad's approach in POC bc it's simpler and actually will keep pruning logic to one component instead of having logic to estimate window outside of the package
* pruner --> interface (?) (regardless, light node needs stub bc will be easy to prune with ipldV2)
* create interface + method in prune package to do IsPruned to determine from EH whether the header is outside sampling window or not (will inform DASer and will inform shrex)

PR #2751, opened 2023-09-23
refactor(cmd): Extract command functions for Init and Start

### LINKED ISSUE #2442: cmd: Add unit tests for Init and Start
Add unit tests for cmd.Init and cmd.Start

rise2semi:
I extracted command functions to test the Init and Start functions (closes #2442)

To be able to test these command functions they should be rewritten to accept functions from other packages as dependencies. But it would lead to significant changes, so I propose to create dedicated issues for that.

What do you think about it?


distractedm1nd (approved):
Thank you!
Yes, do you mind creating the issues? Sounds good to me

rise2semi:
> Thank you! Yes, do you mind creating the issues? Sounds good to me

Created two issues #2774 #2775

ramin (changes requested):
@rise2semi i'm not sure the tradeoff here between extracting the command and making the code slightly less legible is worth it for the quality of the unit test scope. What do you think about us abandoning this strategy, and working on more of an "integration" test that runs separately, spins up a node and then executes commands to evaluate their success/fail rate? Sure its not "unit" testing, but my sense is this current approach wont really test what happens in the Run of the executed commands, vs the command runner itself? We could work through it together if you like

rise2semi:
> @rise2semi i'm not sure the tradeoff here between extracting the command and making the code slightly less legible is worth it for the quality of the unit test scope. What do you think about us abandoning this strategy, and working on more of an "integration" test that runs separately, spins up a node and then executes commands to evaluate their success/fail rate? Sure its not "unit" testing, but my sense is this current approach wont really test what happens in the Run of the executed commands, vs the command runner itself? We could work through it together if you like

Sure, I agree with your proposal.
I was trying to address the issue #2442.

But, I want to note, that I would like to see the command function itself as a simple wrapper rather than a big function with a lot of logic (what we have right now with cmd/Start).

ramin:
i'm going to close this and move discussion about the refactoring of start and init, plus our strategies around making them tested, into the respective issues that were made

PR #2752, opened 2023-09-25
fix: improve ErrNoStateAccess message

distractedm1nd:
After getting the following response from the RPC
```
➜ celestia rpc state AccountAddress
{
  "jsonrpc": "2.0",
  "id": 1,
  "error": {
    "code": 1,
    "message": "node is running without state access"
  }
}
```

I realized it would be more helpful to tell users what they can do to resolve the issue. Knowing that the node needs to be passed --core.ip when you see "node is running without state access" is not something obvious

### DIFF
--- a/nodebuilder/state/stub.go
+++ b/nodebuilder/state/stub.go
@@ -10,7 +10,7 @@ import (
 	"github.com/celestiaorg/celestia-node/state"
 )
 
-var ErrNoStateAccess = errors.New("node is running without state access")
+var ErrNoStateAccess = errors.New("node is running without state access. run with --core.ip <CORE NODE IP> to resolve")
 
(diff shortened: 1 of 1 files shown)


Wondertan:
We recently discussed a different solution to this problem with @renaynay where we hardcode core.ip address per network so that users can run a fully functioning node with just init and start. After that, this error might be even removed. IIRC, @renaynay planned this for before mainnet.

distractedm1nd:
That would 100% be better UX, as long as we can provide the infra for it :D

renaynay (approved):

ramin (approved):

ramin:
this one got me the other day, nice

PR #2755, opened 2023-09-25
refactor(rpc/blob) extend docs in blob

### LINKED ISSUE #2707: Improve CLI documentation for blob submit
The `celestia rpc blob submit` method has the following docs:

```
Flags:
      --fee int          specifies fee for blob submission (default -1)
      --gas.limit uint   specifies max gas for the blob submission
```

1. The denomination of the fee isn't documented.
2. The `-1` magic number isn't documented. If it's `-1` does it do something automatically? If yes, why not just do that when `--fee` isn't specified?
3. The documentation for gas limit says "max gas" but really it's just "gas" since all gas is consumed.

vgonkivs:
## Overview
Closes #2707
Added usage template for `blob.submit`, that allows to hide confusing -1 param for a fee and add an extra line that explains what will happen if `fee` and `gas.Limit` won't  be provided

```
Submit the blob at the given namespace.
Note:
* only one blob is allowed to submit through the RPC.
* fee and gas.limit params will be calculated automatically if they are not provided as arguments

Usage:
  celestia blob submit [namespace] [blobData] [flags]

Flags:
      --fee int          specifies fee (in TIA) for blob submission [optional]
      --gas.limit uint   sets the maximum amount of gas that is allowed to consume during blob submission [optional]
  -h, --help             help for submit

Global Flags:
      --token string   Authorization token (if not provided, the CELESTIA_NODE_AUTH_TOKEN environment variable will be used)
      --url string     Request URL (default "http://localhost:26658")
```
## Checklist

### DIFF
--- a/nodebuilder/blob/cmd/blob.go
+++ b/nodebuilder/blob/cmd/blob.go
@@ -41,29 +41,18 @@ func init() {
 		&fee,
 		"fee",
 		-1,
-		"specifies fee(in TIA) for blob submission",
+		"specifies fee (in TIA) for blob submission [optional]",
 	)
 
 	submitCmd.PersistentFlags().Uint64Var(
 		&gasLimit,
 		"gas.limit",
 		0,
-		"sets the maximum amount of gas that is allowed to consume during blob submission",
+		"sets the maximum amount of gas that is allowed to consume during blob submission [optional]",
 	)
 
-	submitCmd.SetUsageTemplate("" +
-		"celestia blob submit [namespace] [blobData] [flags]\n\n " +
-		"Flags:\n" +
-		"      --fee       int      specifies fee(in TIA) for blob submission(optional)\n" +
-		"      --gas.limit uint64   sets the maximum amount of that gas is allowed " +
-		"to consume during blob submission(optional)\n" +
-		"      -h, --help           help for submit\n" +
-		"NOTE: fee and gas.limit params will be calculated automatically if they are not provided as arguments.\n\n" +
-		"Global Flags:\n" +
-		"      --token string    Authorization token (if not provided, " +
-		"the CELESTIA_NODE_AUTH_TOKEN environment variable will be used)\n" +
-		"      --url   string    Request URL (default \"http://localhost:26658\")",
-	)
+	// unset the default value to avoid users confusion
+	submitCmd.PersistentFlags().Lookup("fee").DefValue = "0"
 }
 
 var Cmd = &cobra.Command{
@@ -140,9 +129,12 @@ var getAllCmd = &cobra.Command{
 }
 
 var submitCmd = &cobra.Command{
-	Use:   "submit [namespace] [blobData]",
-	Args:  cobra.ExactArgs(2),
-	Short: "Submit the blob at the given namespace. Note: only one blob is allowed to submit through the RPC.",
+	Use:  "submit [namespace] [blobData]",
+	Args: cobra.ExactArgs(2),
+	Short: "Submit the blob at the given namespace.\n" +
+		"Note:\n" +
+		"* only one blob is allowed to submit through the RPC.\n" +
+		"* fee and gas.limit params will be calculated automatically if they are not provided as arguments",
 	RunE: func(cmd *cobra.Command, args []string) error {
 		client, err := cmdnode.ParseClientFromCtx(cmd.Context())
 		if err != nil {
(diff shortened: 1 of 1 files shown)


nodebuilder/blob/cmd/blob.go
```diff
@@ -41,14 +41,26 @@ func init() {
...
+		"sets the maximum amount of gas that is allowed to consume during blob submission",
+	)
+
+	submitCmd.SetUsageTemplate("" +
+		"celestia blob submit [namespace] [blobData] [flags]\n\n " +
+		"Flags:\n" +
+		"      --fee       int      specifies fee(in TIA) for blob submission(optional)\n" +
+		"      --gas.limit uint64   sets the maximum amount of that gas is allowed to consume during blob submission(optional)\n" +
+		"      -h, --help           help for submit\n" +
+		"NOTE: fee and gas.limit params will be calculated automatically if they will not be provided.\n\n" +
```
ramin, line 60: "...if they are not provided as arguments" might read better

ramin (approved):

nodebuilder/blob/cmd/blob.go
vgonkivs: Thanks 🙏

distractedm1nd (approved):

nodebuilder/blob/cmd/blob.go
```diff
@@ -41,14 +41,28 @@ func init() {
...
+		"Flags:\n" +
+		"      --fee       int      specifies fee(in TIA) for blob submission(optional)\n" +
+		"      --gas.limit uint64   sets the maximum amount of that gas is allowed " +
+		"to consume during blob submission(optional)\n" +
+		"      -h, --help           help for submit\n" +
+		"NOTE: fee and gas.limit params will be calculated automatically if they are not provided as arguments.\n\n" +
+		"Global Flags:\n" +
+		"      --token string    Authorization token (if not provided, " +
+		"the CELESTIA_NODE_AUTH_TOKEN environment variable will be used)\n" +
+		"      --url   string    Request URL (default \"http://localhost:26658\")",
```
renaynay, line 65: This sucks to have hardcoded like this in string - can we not add on usage options to the flags themselves?

renaynay: when you add flags to a specific cmd, it should spit out the flags for you

vgonkivs: cmd.Flag("flagKey") returns the flag but anyway we will have to hardcode the string if we want to hide unneeded information

vgonkivs: reworked

nodebuilder/blob/cmd/blob.go
```diff
@@ -145,7 +151,9 @@ var submitCmd = &cobra.Command{
 		if err != nil {
 			return fmt.Errorf("error creating a blob:%v", err)
 		}
-
+		if fee != -1 {
+			panic(fee)
+		}
```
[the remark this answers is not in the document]
vgonkivs: ooooops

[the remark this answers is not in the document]
vgonkivs: removed

nodebuilder/blob/cmd/blob.go
```diff
@@ -41,15 +41,18 @@ func init() {
 		&fee,
 		"fee",
 		-1,
-		"specifies fee for blob submission",
+		"specifies fee (in TIA) for blob submission [optional]",
```
adlerjohn, line 44: should this be `utia` instead of `tia`?

nodebuilder/blob/cmd/blob.go
```diff
@@ -41,15 +41,18 @@ func init() {
...
 	)
 
 	submitCmd.PersistentFlags().Uint64Var(
 		&gasLimit,
 		"gas.limit",
 		0,
-		"specifies max gas for the blob submission",
+		"sets the maximum amount of gas that is allowed to consume during blob submission [optional]",
```
adlerjohn, line 51: Gas is always fully consumed with no refund.
```suggestion
		"sets the amount of gas that is consumed during blob submission [optional]",
```

nodebuilder/blob/cmd/blob.go
adlerjohn, line 44: Hmmmm. So one thing that's tricky about this is that an astute reader will notice the type is `int` instead of `uint`, and will naturally want to know what will happen if a negative value is set. IMO this should be documented.

vgonkivs: done

vgonkivs: done

renaynay (approved):

PR #2759, opened 2023-09-25
chore(nodebuilder)!: Elevate perms for certain endpoints

### LINKED ISSUE #2406: nodebuilder/header: Elevate permissions of subscription endpoint

renaynay:
TODO: 
- [x] should perms be elevated for `share` endpoints like `GetEDS` ? IMO yes.

Resolves #2406


nodebuilder/state/state.go
```diff
@@ -150,11 +150,11 @@ type API struct {
 		QueryDelegation func(
 			ctx context.Context,
 			valAddr state.ValAddress,
-		) (*types.QueryDelegationResponse, error) `perm:"public"`
+		) (*types.QueryDelegationResponse, error) `perm:"read"`
 		QueryUnbonding func(
 			ctx context.Context,
 			valAddr state.ValAddress,
-		) (*types.QueryUnbondingDelegationResponse, error) `perm:"public"`
+		) (*types.QueryUnbondingDelegationResponse, error) `perm:"read"`
```
Wondertan, line 157: Why not public?

renaynay:
@Wondertan actually we should at least elevate perms of SharesAvailable as there is a sanity check there that could cause node to panic in case of malformed root -- I also elevated perms of the other calls as i don't think calls that require node to do work should be public.

distractedm1nd:
Read permissions are sufficient, even for SharesAvailable with the panic. Public methods are endpoints you can give anyone access to, read perms are next level up and should only be given to people who are not using with ill intent anyways

nodebuilder/state/state.go
distractedm1nd: True. We should maybe think about making these methods take the AccountAddress as well

distractedm1nd (approved):

renaynay:
@distractedm1nd i made the comment re SharesAvailable before i elevated the perm to read

distractedm1nd:
Please mark as breaking

renaynay:
already implemented

PR #2760, opened 2023-09-25
fix(header): Implement nil check on Commit to prevent panic in `Validate` on EH

renaynay:
A malformed `ExtendedHeader` with a nil `Commit` could cause a panic.

### DIFF
--- a/header/header.go
+++ b/header/header.go
@@ -120,6 +120,9 @@ func (eh *ExtendedHeader) Validate() error {
 			eh.RawHeader.Version.App)
 	}
 
+	if eh.Commit == nil {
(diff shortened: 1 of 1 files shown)


distractedm1nd (approved):

header/header.go
```diff
@@ -120,6 +120,9 @@ func (eh *ExtendedHeader) Validate() error {
 			eh.RawHeader.Version.App)
 	}
 
+	if eh.Commit == nil {
+		return fmt.Errorf("nil commit at height %d", eh.Height())
+	}
```
Wondertan, line 125: Should be inside core part of ValidateBasic. Let submit fix there

renaynay: Yea actually

PR #2761, opened 2023-09-25
fix(share/p2p/peers): Prevent making pools for nil DataHash

renaynay:
Not critical, but if nil DataHash is broadcasted, node could create pool for empty datahash and just keep it around until garbage collected. Best to short-circuit instead and reject message.


renaynay:
Actually this doesn't pass tests so converting to draft til i figure it out, will do tmw

renaynay:
Actually we already do this inside of validate -

ramin (dismissed):
seems reasonable, only suggestion might be to move the msg.DataHash.Validate() to be the first check so as to skip all the other checks proceeding first?

walldiss (approved):

ramin (approved):
:shipit:

distractedm1nd (approved):

PR #2762, opened 2023-09-26
feat(cmd/rpc): add node store flag

vgonkivs:
## Overview
Added node.store flag to all rpc cmds that allows to get auth token from the config

`celestia blob submit 0x42690c204d39600fddd3 'gm' --node.store=~/.celestia-light-arabica-10`

## Checklist

### DIFF
--- a/cmd/rpc.go
+++ b/cmd/rpc.go
@@ -5,17 +5,12 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"path/filepath"
 
-	"github.com/cristalhq/jwt"
-	"github.com/mitchellh/go-homedir"
 	"github.com/spf13/cobra"
 	flag "github.com/spf13/pflag"
 
 	rpc "github.com/celestiaorg/celestia-node/api/rpc/client"
 	"github.com/celestiaorg/celestia-node/api/rpc/perms"
-	"github.com/celestiaorg/celestia-node/libs/authtoken"
-	"github.com/celestiaorg/celestia-node/libs/keystore"
 	nodemod "github.com/celestiaorg/celestia-node/nodebuilder/node"
 )
 
@@ -60,20 +55,19 @@ func RPCFlags() *flag.FlagSet {
 }
 
 func InitClient(cmd *cobra.Command, _ []string) error {
-	var (
-		err   error
-		token string
-	)
+	if authTokenFlag == "" {
+		authTokenFlag = os.Getenv(authEnvKey)
+	}
 
 	if authTokenFlag == "" {
-		token, err = getToken(storePath)
+		token, err := getToken(storePath)
+		if err != nil {
+			return fmt.Errorf("cant get the access to the auth token:%v", err)
+		}
+		authTokenFlag = token
 	}
 
-	if err != nil {
-		token = os.Getenv(authEnvKey)
-	}
-
-	client, err := rpc.NewClient(cmd.Context(), requestURL, token)
+	client, err := rpc.NewClient(cmd.Context(), requestURL, authTokenFlag)
 	if err != nil {
 		return err
 	}
@@ -85,41 +79,20 @@ func InitClient(cmd *cobra.Command, _ []string) error {
 
 func getToken(path string) (string, error) {
 	if path == "" {
-		return "", errors.New("empty path")
+		return "", errors.New("root directory was not specified")
 	}
 
-	expanded, err := homedir.Expand(filepath.Clean(path))
+	ks, err := newKeystore(path)
 	if err != nil {
-		fmt.Printf("error expanding the root dir path:%v", err)
 		return "", err
 	}
 
-	ks, err := keystore.NewFSKeystore(filepath.Join(expanded, "keys"), nil)
-	if err != nil {
-		fmt.Printf("error creating the keystore:%v", err)
-		return "", err
-	}
-
-	var key keystore.PrivKey
-	key, err = ks.Get(nodemod.SecretName)
+	key, err := ks.Get(nodemod.SecretName)
 	if err != nil {
 		fmt.Printf("error getting the private key:%v", err)
 		return "", err
 	}
-
-	signer, err := jwt.NewHS256(key.Body)
-	if err != nil {
-		fmt.Printf("error creating a signer:%v", err)
-		return "", err
-	}
-
-	token, err := authtoken.NewSignedJWT(signer, perms.AllPerms)
-	if err != nil {
-		fmt.Println(err)
-		return "", err
-	}
-	authTokenFlag = token
-	return token, nil
+	return buildJWTToken(key.Body, perms.AllPerms)
 }
 
 type rpcClientKey struct{}
--- a/cmd/rpc.go
+++ b/cmd/rpc.go
@@ -3,11 +3,20 @@ package cmd
 import (
 	"context"
 	"errors"
+	"fmt"
 	"os"
+	"path/filepath"
 
+	"github.com/cristalhq/jwt"
+	"github.com/mitchellh/go-homedir"
 	"github.com/spf13/cobra"
+	flag "github.com/spf13/pflag"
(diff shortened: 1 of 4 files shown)


cmd/rpc.go
```diff
@@ -19,25 +28,52 @@ const (
...
+	)
+
 	if authTokenFlag == "" {
-		authTokenFlag = os.Getenv(authEnvKey)
+		token, err = getToken(storePath)
+	}
+
+	if err != nil {
+		token = os.Getenv(authEnvKey)
 	}
```
renaynay, line 74: IMO env check should be first as that's what's expected / documented, and THEN the token can be grabbed from the store path in case no auth token is set in env var

renaynay: or we could just do away with env var entirely and have it work by grabbing token from store by default and if user needs to specify token, they can do with flag. cc @distractedm1nd

cmd/rpc.go
```diff
@@ -47,6 +83,45 @@ func InitClient(cmd *cobra.Command, _ []string) error {
 	return nil
 }
 
+func getToken(path string) (string, error) {
```
renaynay, line 86: i think we already have some code for this somewehre - can you reuse?

cmd/rpc.go
vgonkivs: I like the second approach but in this case user has to specify either token or the path to the root each time, that could be annoying. I would move the reading from env upper - before getting the token from the config

cmd/rpc.go
```diff
@@ -19,24 +23,50 @@ const (
...
+	)
+
+	storeFlag := NodeFlags().Lookup(nodeStoreFlag)
+
+	fset.StringVar(
+		&storePath,
+		nodeStoreFlag,
+		"",
+		storeFlag.Usage,
+	)
```
renaynay, line 53: is there not a better way to import the flags? this looks messy

cmd/rpc.go
```diff
@@ -47,6 +77,24 @@ func InitClient(cmd *cobra.Command, _ []string) error {
...
+	}
+
+	ks, err := newKeystore(path)
+	if err != nil {
+		return "", err
+	}
+
+	key, err := ks.Get(nodemod.SecretName)
+	if err != nil {
+		fmt.Printf("error getting the private key:%v", err)
```
renaynay, line 92:
```suggestion
		fmt.Printf("error getting the JWT secret: %v", err)
```

cmd/rpc.go
```diff
@@ -19,24 +23,50 @@ const (
...
 
 func InitClient(cmd *cobra.Command, _ []string) error {
 	if authTokenFlag == "" {
 		authTokenFlag = os.Getenv(authEnvKey)
 	}
 
+	if authTokenFlag == "" {
+		token, err := getToken(storePath)
+		if err != nil {
+			return fmt.Errorf("cant get the access to the auth token:%v", err)
```
renaynay, line 65:
```suggestion
			return fmt.Errorf("cant get the access to the auth token: %v", err)
```

cmd/rpc.go
vgonkivs: Why is it messy? there is one extra line `storeFlag := NodeFlags().Lookup(nodeStoreFlag)`. All other code has the same approach.

cmd/rpc.go
distractedm1nd: This is moving into devex territory and so would love input from @jcstein 
To sum it up we have two approaches we are considering:
1. We use the token in the CELESTIA_NODE_AUTH_TOKEN, but if it doesnt exist it grabs it from the node store
2. We get rid of the environment variable, it gets from node store automatically, and if user wants to specify token they do it with a flag

vgonkivs: I would personally go with the second approach. If the user wants to use the token from the store but already has CELESTIA_NODE_AUTH_TOKEN in the env we will anyway get it from the env(even with the passed store flag). So, I would leave CELESTIA_NODE_AUTH_TOKEN as a fallback solution for getting the token.

cmd/rpc.go
```diff
@@ -19,24 +23,50 @@ const (
...
-	return &authTokenFlag,
+	fset.StringVar(
+		&authTokenFlag,
 		"token",
 		"",
-		"Authorization token (if not provided, the " + authEnvKey + " environment variable will be used)"
+		"Authorization token (if not provided, the "+authEnvKey+" environment variable will be used)",
+	)
+
+	storeFlag := NodeFlags().Lookup(nodeStoreFlag)
```
renaynay, line 46:
```suggestion
	storeFlag := NodeFlags().Lookup(nodeStoreFlag)
         set.AddFlag(storeFlag)
```

this is what i meant ^ @vgonkivs

cmd/rpc.go
renaynay: @vgonkivs @jcstein @distractedm1nd 

I think the cleanest approach is adding the `--p2p.network` command to rpc (in case it hasn't already) as well as the `--node.store` flag in the case there's a custom store, **AND** also add the `--auth.token` flag that allows user to pass token that will override the need for passing a network or node store path as the token will already have been passed. 

So from user perspective, it would look like this: 

**mocha** (default network):
```
celestia blob submit [params] # generates token from default mocha store
```

**arabica**: 
```
celestia blob submit [params] --p2p.network arabica # generates token from default arabica store
```

**custom node store**: 
```
celestia blob submit [params] --node.store <path to node store> # generates token from the given node store
```
**with auth token**: 
```
celestia blob submit [params] --auth.token <token> # takes the given token, no need to look in store
```

_**There is no more need for the env var at all in this case.**_

renaynay: With the above approach ^ there is no need for env var at all, it will only add confusion to user + difficulty to parsing logic

vgonkivs: In case of multiple calls, the user will have to specify auth.token/path each time which is weird imo. And having env solves this problem

vgonkivs: @renaynay, besides the network flag, the user will also have to pass node type flag

vgonkivs:
```
celestia blob submit [params] --p2p.network arabica --node.type light # generates token from default arabica store
```

vgonkivs:
```
celestia share get-share "$(celestia header get-by-height 147105 | jq '.result.dah' -r)" 0 2 --p2p.network arabica --node.type light
```

vgonkivs: queries could be more difficult than just submit ^

cmd/rpc.go
vgonkivs: reworked

cmd/rpc.go
renaynay: ah yes node.type would be necessary too in case of just p2p.network. 

hm.

env var likely would be the best UX for repeated requests.

cmd/rpc.go
```diff
@@ -44,13 +43,7 @@ func RPCFlags() *flag.FlagSet {
...
 
 	storeFlag := NodeFlags().Lookup(nodeStoreFlag)
-
-	fset.StringVar(
-		&storePath,
-		nodeStoreFlag,
-		"",
-		storeFlag.Usage,
-	)
+	fset.AddFlag(storeFlag)
```
renaynay, line 46: @vgonkivs did you change the structure of the node store flag so that it doesn't store directly to that variable?

vgonkivs: `node.store ` flag uses the internal field of the` *flag.Flag`. It did not use an additional var. And for the RPC cmds, yes, I removed that internal variable and now using the value directly from the flag(as you suggested):

```
if cmd.Flag(nodeStoreFlag).Changed {
		storePath = cmd.Flag(nodeStoreFlag).Value.String()	
	}
```

Wondertan (approved):
I didn't think through flags changes, but looks good overwise

PR #2764, opened 2023-09-26
refactor(gateway): remove deprecated endpoints

vgonkivs:
## Overview
Removed all deprecated endpoints 

## Checklist


renaynay (changes requested):
we only need to remove writes to state, all other requests that are simply reading from the node (including state reads) can be left open

renaynay:
actually i think balance should also be deprecated only bc it access the signer cc @Wondertan

vgonkivs:
They can be accessed through the rpc

ramin (approved):

api/gateway/state.go
```diff
@@ -9,17 +9,12 @@ import (
...
 )
 
 const (
-	balanceEndpoint            = "/balance"
-	submitTxEndpoint           = "/submit_tx"
-	submitPFBEndpoint          = "/submit_pfb"
-	queryDelegationEndpoint    = "/query_delegation"
-	queryUnbondingEndpoint     = "/query_unbonding"
-	queryRedelegationsEndpoint = "/query_redelegations"
+	balanceEndpoint  = "/balance"
```
renaynay, line 16: I'm a bit concerned with allowing reads for balance without passing an addr (which would be just a `Balance` query and would access the signers account addr and check its balance)

nodebuilder/state/cmd/state.go
```diff
@@ -156,47 +153,6 @@ var submitTxCmd = &cobra.Command{
 	},
 }
 
-var submitPFBCmd = &cobra.Command{
```
renaynay, line 159: is this meant to be part of this PR or it just contains commits from the deprecating submitPFB?

renaynay (commented):
one comment on `balance` that want to discuss with @distractedm1nd and @Wondertan

nodebuilder/state/cmd/state.go
vgonkivs: I think we can leave it here. Removing all at once

distractedm1nd: I would ask @jcstein if this is okay first

api/gateway/state.go
distractedm1nd: This PR handles that correctly - balance w/o an address writes an error

api/gateway/state.go
```diff
@@ -57,24 +37,25 @@ func (h *Handler) handleBalanceRequest(w http.ResponseWriter, r *http.Request) {
 	// read and parse request
 	vars := mux.Vars(r)
 	addrStr, exists := vars[addrKey]
-	if exists {
-		// convert address to Address type
-		var addr state.AccAddress
-		addr, err = types.AccAddressFromBech32(addrStr)
+	if !exists {
+		writeError(w, http.StatusBadRequest, balanceEndpoint, errors.New("invalid endpoint"))
```
distractedm1nd, line 41:
```suggestion
		writeError(w, http.StatusBadRequest, balanceEndpoint, errors.New("balance endpoint requires address"))
```

nodebuilder/state/cmd/state.go
vgonkivs: The target branch is a feature branch, so I guess it's fine. But sure, let's wait @jcstein

jcstein: if I understand correctly and this PR is to deprecate submitPFB as a gateway endpoint, we should make sure this is tracked properly with other breaking RPC changes and let teams know accordingly. but I do think this is okay to deprecate

that being said, while i will be updating docs, i cannot speak for everyone using the submitPFB endpoint still. so i will cc @Ferret-san and @YazzyYaz in case there are any teams that you know of still using this. we will have to let them know that this is deprecated when it is in a release

it is also important that the release is made properly as a breaking release, and not just another RC that breaks everything that worked in the previous RC on same semantic version i.e. v0.11.0-rc1 to v0.11.0-rc2 -- unless I am misunderstanding how breaking changes should be handled between RCs 🫡

vgonkivs: Please note, that this PR will be merged into a feature branch and not main.

renaynay: >it is also important that the release is made properly as a breaking release, and not just another RC that breaks everything that worked in the previous RC on same semantic version i.e. v0.11.0-rc1 to v0.11.0-rc2 -- unless I am misunderstanding how breaking changes should be handled between RCs 🫡

@jcstein we've been working towards a stable release which will be v0.11.0 and expected in a few weeks. Every RC up to that point has been / will likely be breaking in some way or another but we will 1. absolutely document when it is, and 2. we are making feature branches for breaks so that we get **all** the known API breaks that we want to do all in one go (one RC)

renaynay (approved):

nodebuilder/state/cmd/state.go
jcstein: sounds good! thank you for clarifying @vgonkivs and @renaynay. i am realizing that since we put these callouts in months ago to warn people that gateway will be deprecated, it should be no surprise and we will communicate accordingly
scrot 2023-09-28 at 9 43 04 AM

jcstein: i've made an issue to track this as "unreleased node" https://github.com/celestiaorg/docs/issues/1094

api/gateway/state.go
jcstein, line 16: oh hmm - @distractedm1nd @vgonkivs can you please help me understand if the `/balance` and `/submit_tx` endpoints are still there or not?

vgonkivs: We still have both endpoints available through the cmd(rpc client), but for the gateway we've removed `/balance` for self-node in favour of `/balance/addrKey`. Submit_tx is still available for the gateway

Wondertan:
utACK

Issue #2703, opened 2023-09-14
share/discovery: overly aggressive retrying 


Wondertan:
I was looking into the mocha network today and found out a few things we need to change in discovery:
* We retry too often. 
	* If we cannot get the needed value of peers, we retry every second. Instant retries are usually useless; no new peers would join the network or advertise themselves in the second timeframe. But they are not always useless, so
	* Instead, we should have a capped linear backoff.
		* Factor 5 starting from 5 seconds and capped by 10 mins
* #2573 introduced a minor bug: `discoveryLoop` does not listen to `triggerDisc` channel anymore

nodersteam:
Will work on it, thx

ramin:
hey @nodersteam, curious if you were feeling comfortable enough to open a new PR with the stuff we've been discussing for https://github.com/celestiaorg/celestia-node/pull/2766?

nodersteam:
> hey @nodersteam, curious if you were feeling comfortable enough to open a new PR with the stuff we've been discussing for #2766?

Whatever you say, we can do it this way

Then I close the old PR, divide it into two new ones that we discussed?

nodersteam:
@ramin

nodersteam:
@Wondertan Hello
As a result, we come to the conclusion that we are closing the old PR and making 2 new ones that will separately correct backoff and asynchrony?

We do not do this without your confirmation so as not to do something unnecessary

cristaloleg:
Point 1 and 3 are still relevant.

PR #2766, opened 2023-09-26
feat(discovery): Implement capped linear backoff and asynchronous tri…

nodersteam:
Implemented a capped linear backoff mechanism to reduce the frequency of retries during the discovery process. The backoff starts at 5 seconds and increases by a factor of 5, capped at a maximum duration of 10 minutes.

Added asynchronous triggering of the discoveryLoop using the triggerDisc channel

#2703

### DIFF
--- a/share/p2p/discovery/discovery.go
+++ b/share/p2p/discovery/discovery.go
@@ -36,7 +36,9 @@ const (
 
 	// logInterval defines the time interval at which a warning message will be logged
 	// if the desired number of nodes is not detected.
-	logInterval = 5 * time.Minute
+	logInterval            = 5 * time.Minute
+	backoffInitialDuration = 5 * time.Second
+	maxBackoffDuration     = 10 * time.Minute
 )
 
 // discoveryRetryTimeout defines time interval between discovery attempts, needed for tests
@@ -202,21 +204,32 @@ func (d *Discovery) Advertise(ctx context.Context) {
 // It initiates peer discovery upon request and restarts the process until the soft limit is
 // reached.
 func (d *Discovery) discoveryLoop(ctx context.Context) {
-	t := time.NewTicker(discoveryRetryTimeout)
-	defer t.Stop()
+	backoffDuration := backoffInitialDuration
+	backoffTicker := time.NewTicker(backoffDuration)
+	defer backoffTicker.Stop()
 
 	warnTicker := time.NewTicker(logInterval)
 	defer warnTicker.Stop()
 
+	// Function to initiate discovery and restart the timer
+	runDiscovery := func() {
+		if !d.discover(ctx) {
+			// rerun discovery if the number of peers hasn't reached the limit
+			return
+		}
+
+		backoffTicker.Stop()
+		backoffDuration = backoffInitialDuration
+		backoffTicker = time.NewTicker(backoffDuration)
+	}
+
 	for {
-		// drain all previous ticks from the channel
-		drainChannel(t.C)
 		select {
-		case <-t.C:
-			if !d.discover(ctx) {
-				// rerun discovery if the number of peers hasn't reached the limit
-				continue
+		case <-backoffTicker.C:
+			if backoffDuration > maxBackoffDuration {
+				backoffDuration = maxBackoffDuration
 			}
+			runDiscovery()
 		case <-warnTicker.C:
 			if d.set.Size() < d.set.Limit() {
 				log.Warnf(
@@ -225,8 +238,8 @@ func (d *Discovery) discoveryLoop(ctx context.Context) {
 					logInterval, d.set.Size(), d.set.Limit(),
 				)
 			}
-			// Do not break the loop; just continue
-			continue
+		case <-d.triggerDisc: // Listen for triggerDisc channel
+			runDiscovery()
 		case <-ctx.Done():
 			return
 		}
@@ -377,14 +390,4 @@ func (d *Discovery) handleDiscoveredPeer(ctx context.Context, peer peer.AddrInfo
 	//  connection should be kept or not, similar to mesh link in GossipSub.
 	d.host.ConnManager().Protect(peer.ID, rendezvousPoint)
 	return true
-}
-
-func drainChannel(c <-chan time.Time) {
-	for {
-		select {
-		case <-c:
-		default:
-			return
-		}
-	}
 }


share/p2p/discovery/discovery.go
```diff
@@ -202,21 +204,32 @@ func (d *Discovery) Advertise(ctx context.Context) {
...
 
+	// Function to initiate discovery and restart the timer
+	runDiscovery := func() {
+		if !d.discover(ctx) {
+			// rerun discovery if the number of peers hasn't reached the limit
+			return
+		}
+
+		backoffTicker.Stop()
+		backoffDuration = backoffInitialDuration
```
walldiss, line 222: backoffDuration doesn't have linear grows

share/p2p/discovery/discovery.go
```diff
@@ -202,21 +204,32 @@ func (d *Discovery) Advertise(ctx context.Context) {
...
-		// drain all previous ticks from the channel
-		drainChannel(t.C)
 		select {
-		case <-t.C:
-			if !d.discover(ctx) {
-				// rerun discovery if the number of peers hasn't reached the limit
-				continue
+		case <-backoffTicker.C:
+			if backoffDuration > maxBackoffDuration {
+				backoffDuration = maxBackoffDuration
```
walldiss, line 230: would be better to have this logic near the place where `backoffDuration` is updated

share/p2p/discovery/discovery.go
```diff
@@ -225,8 +238,8 @@ func (d *Discovery) discoveryLoop(ctx context.Context) {
 					logInterval, d.set.Size(), d.set.Limit(),
 				)
 			}
-			// Do not break the loop; just continue
-			continue
+		case <-d.triggerDisc: // Listen for triggerDisc channel
+			runDiscovery()
```
walldiss, line 242: This would cause discovery to retry without respect of backoff. At this point we know that we need to trigger discovery, since the loop has started, there is no need to skip backoff in this case.

share/p2p/discovery/discovery.go
```diff
@@ -202,21 +204,32 @@ func (d *Discovery) Advertise(ctx context.Context) {
 // It initiates peer discovery upon request and restarts the process until the soft limit is
 // reached.
 func (d *Discovery) discoveryLoop(ctx context.Context) {
-	t := time.NewTicker(discoveryRetryTimeout)
-	defer t.Stop()
+	backoffDuration := backoffInitialDuration
+	backoffTicker := time.NewTicker(backoffDuration)
```
walldiss, line 208: If you read only once from `ticker`, it is better to use `time.Timer` that is designed specifically for that.

share/p2p/discovery/discovery.go
```diff
@@ -202,21 +204,32 @@ func (d *Discovery) Advertise(ctx context.Context) {
...
+	defer backoffTicker.Stop()
```
walldiss, line 209: Not an issue here, but after making changes, pls make sure `Stop()` will not be called twice.

share/p2p/discovery/discovery.go
```diff
@@ -202,21 +204,32 @@ func (d *Discovery) Advertise(ctx context.Context) {
...
 
 	warnTicker := time.NewTicker(logInterval)
 	defer warnTicker.Stop()
 
+	// Function to initiate discovery and restart the timer
+	runDiscovery := func() {
+		if !d.discover(ctx) {
+			// rerun discovery if the number of peers hasn't reached the limit
+			return
```
walldiss, line 218: This `return` stop execution only of anonymous func, when you would want to return from outer loop

share/p2p/discovery/discovery.go
```diff
@@ -202,21 +204,32 @@ func (d *Discovery) Advertise(ctx context.Context) {
...
 		select {
-		case <-t.C:
-			if !d.discover(ctx) {
-				// rerun discovery if the number of peers hasn't reached the limit
-				continue
+		case <-backoffTicker.C:
+			if backoffDuration > maxBackoffDuration {
+				backoffDuration = maxBackoffDuration
 			}
+			runDiscovery()
```
walldiss, line 232: It seems to keep iteration regardless if `d.discover(ctx)` succeded or not (see comment about anonymous func)

walldiss (changes requested):

share/p2p/discovery/discovery.go
nodersteam: Will fix and update all

ramin (changes requested):
@nodersteam i think baking the backoff into the ticker is the wrong approach. I'd also request splitting the two tasks (backoff, async trigger) into two PRs for ease of review if you can.

For the backoff, i'd suggest something like adding a "backoffDuration" time.Interval to the Discover struct (or even something that wraps that to apply a backoff) then leave the tick interval the same, but add/update an incrementing backup duration on d.discover failures that will be checked before executing. Will make things simpler. Also, i'd suggest adding a little jitter to the backoff interval so very slightly randomize the backoff times. Want to give it a shot?

nodersteam:
> @nodersteam i think baking the backoff into the ticker is the wrong approach. I'd also request splitting the two tasks (backoff, async trigger) into two PRs for ease of review if you can.
> 
> 
> 
> For the backoff, i'd suggest something like adding a "backoffDuration" time.Interval to the Discover struct (or even something that wraps that to apply a backoff) then leave the tick interval the same, but add/update an incrementing backup duration on d.discover failures that will be checked before executing. Will make things simpler. Also, i'd suggest adding a little jitter to the backoff interval so very slightly randomize the backoff times. Want to give it a shot?
> 
> 
> 
> 

yes, I will try to implement this
the comments that were left above have already been practically corrected, but I will also take into account your comment

nodersteam:
@ramin @walldiss 

Hello! What do you think about improving test logging?

I’m currently writing tests to check the correct implementation of everything I need, and I’m faced with the fact that logging tests is not informative. What do you say about this format? Or is this not necessary?
image

ramin:
@nodersteam i quite like the idea of being able to see some more detailed logging if the `-v` flag is provided when running tests. I'd say propose it with the updates to the PR and explain why it was necessary and helpful so we can understand while reviewing.

nodersteam:
@ramin 

Hi! Sorry for ping here
Check pls our chat in Doscord

nodersteam:
@walldiss Is the question of continuing to search for peers after finding a sufficient number still relevant?

PR #2771, opened 2023-09-27
chore(api | nodebuilder)!: Elevate all api permissions to `read` and disallow `public`, remove `client.NewPublicClient`

renaynay:
We made a decision synchronously with @adlerjohn and @ramin to require at least `read`-level permissions to be able to access *any* endpoint on the node's API. 

Therefore, `NewPublicClient` is no longer necessary.


Wondertan (approved):

ramin (approved):

distractedm1nd:
This is breaking API so needs to either go into feature branch or you bump api version and regenerate the doc

distractedm1nd (changes requested):
see last comment

distractedm1nd (approved):

PR #2773, opened 2023-09-27
refactor(share)!: replacing share.Root with header.ExtendedHeader in APIS

### LINKED ISSUE #2772: refactor(share/getter)!: Take header.ExtendedHeader instead of share.DataHash
Originally, we wanted to have an architectural separation between the header and share module. 

For pruning and the new IPLD protocol, information from the extended header is useful and any solutions that do not break the API are overly complicated (for example, by maintaining indexes of DataHash -> Timestamp). So we will change the getter interfaces' methods to take a *header.ExtendedHeader instead of a share.DataHash.

distractedm1nd:
WIP. Closes #2772 

Mostly because I don't like where I moved the utilities and needing to export them, so will try to think of a better solution - the moving of utilities was to keep DRY between store getter and shrexnd, while removing shrexnd server's dependence on a share.Getter.


Wondertan:
Why don't we remove Root in this pr? Timing constraints?

distractedm1nd:
Also, this specific part breaks RPC while removal of share.Root is just a pure internal refactoring. By timing constraints I mainly mean I want to get all rpc breaks out ASAP

ramin (approved):
looks clean. The testing.TB stuff is totally new to me! nice.

share/p2p/shrexnd/server.go
```diff
@@ -39,13 +38,12 @@ type Server struct {
 }
 
 // NewServer creates new Server
-func NewServer(params *Parameters, host host.Host, store *eds.Store, getter share.Getter) (*Server, error) {
+func NewServer(params *Parameters, host host.Host, store *eds.Store) (*Server, error) {
```
Wondertan, line 41: Because the server does not have access to the Header, correct?

Wondertan (approved):

share/eds/utils.go
```diff
@@ -46,3 +55,98 @@ func closeAndLog(name string, closer io.Closer) {
 		log.Warnw("closing "+name, "err", err)
 	}
 }
+
+// RetrieveNamespaceFromStore gets all EDS shares in the given namespace from
+// the EDS store through the corresponding CAR-level blockstore. It is extracted
+// from the store getter to make it available for reuse in the shrexnd client.
```
walldiss, line 61:
```suggestion
// from the store getter to make it available for reuse in the shrexnd server.
```

share/getters/store.go
```diff
@@ -110,44 +114,5 @@ func (sg *StoreGetter) GetSharesByNamespace(
...
-			log.Errorf("getter/store: failed to remove CAR after detected corruption: %w", err)
-		}
-		err = share.ErrNotFound
-	}
-	if err != nil {
-		return nil, fmt.Errorf("getter/store: failed to retrieve shares by namespace: %w", err)
-	}
-
-	return shares, nil
+	return eds.RetrieveNamespaceFromStore(ctx, sg.store, header.DAH, namespace)
```
walldiss, line 117: Would it be possible to identfy that error happened in store getter and not in shrexnd server? I would suggest to wrap it here just in case, so later we don't have to worry to double check, that caller wraps it with identifiable text

share/eds/utils.go
```diff
@@ -46,3 +55,98 @@ func closeAndLog(name string, closer io.Closer) {
...
+		// IPLD node not found after the index pointed to this shard and the CAR
+		// blockstore has been opened successfully is a strong indicator of
+		// corruption. We remove the block on bridges and fulls and return
+		// share.ErrNotFound to ensure the data is retrieved by the next getter.
+		// Note that this recovery is manual and will only be restored by an RPC
+		// call to SharesAvailable that fetches the same datahash that was
+		// removed.
+		err = store.Remove(ctx, dah.Hash())
+		if err != nil {
+			log.Errorf("eds/utils: failed to remove CAR from store after detected corruption: %w", err)
```
walldiss, line 99: Optional/nit: "eds/utils:" feels kinda too generic and useless. Error should be wrapped by the caller.

share/getters/getter_test.go
```diff
@@ -41,92 +43,109 @@ func TestStoreGetter(t *testing.T) {
...
+				t.Fatal("context ended before successful retrieval")
+			case <-ticker.C:
+				has, err := edsStore.Has(ctx, eh.DAH.Hash())
+				if err != nil {
+					t.Fatal(err)
+				}
+				if !has {
+					require.NoError(t, err)
+					return
+				}
```
walldiss, line 146: Just curious. Why do we need to wait now and didn't need to before?

distractedm1nd: we did, but it was just flaky. passes most of the time on first try

share/p2p/shrexnd/server.go
renaynay: wdym?

header/headertest/testing.go
```diff
@@ -202,7 +202,7 @@ func (s *TestSuite) nextProposer() *types.Validator {
 }
 
 // RandExtendedHeader provides an ExtendedHeader fixture.
-func RandExtendedHeader(t *testing.T) *header.ExtendedHeader {
+func RandExtendedHeader(t testing.TB) *header.ExtendedHeader {
```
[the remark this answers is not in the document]
distractedm1nd: Because there is a benchmark now that needs it

nodebuilder/share/share.go
```diff
@@ -27,65 +28,67 @@ var _ Module = (*API)(nil)
...
+		ctx context.Context, header *header.ExtendedHeader, namespace share.Namespace,
+	) (share.NamespacedShares, error)
 }
 
 // API is a wrapper around Module for the RPC.
 // TODO(@distractedm1nd): These structs need to be autogenerated.
 type API struct {
 	Internal struct {
-		SharesAvailable func(context.Context, *share.Root) error `perm:"read"`
+		SharesAvailable func(context.Context, *header.ExtendedHeader) error `perm:"public"`
```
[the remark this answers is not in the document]
distractedm1nd: Already did, just merged this one incorrectly, thankfully you noticed

share/p2p/shrexnd/server.go
distractedm1nd: This is because we need to use the store directly now instead of implicit StoreGetter

renaynay (approved):

walldiss (approved):

PR #2809, opened 2023-10-05
refactoring!(rpc/cmd): rework rpc commands

vgonkivs:
## Overview

* reworked p2p/das/header/share/admin/state cmds
* extracted cmds to the respective packages in `nodebuilder`
* added rpc flag that allows to specify a path for the node's config
* removed deprecated endpoints
* removed authEnvKey in favour of authToke/node.store path

## Checklist


cmd/rpc.go
```diff
@@ -0,0 +1,98 @@
...
+	if err != nil {
+		fmt.Printf("error getting the JWT secret: %v", err)
+		return "", err
+	}
+	return buildJWTToken(key.Body, perms.AllPerms)
+}
+
+type rpcClientKey struct{}
+
+func ParseClientFromCtx(ctx context.Context) (*rpc.Client, error) {
```
ramin, line 92: realizing now as i see this as a whole, but i think this would be better as NewClientFromContext vs Parse

nodebuilder/share/cmd/share.go
```diff
@@ -0,0 +1,209 @@
...
+		if err != nil {
+			return err
+		}
+
+		shares, err := client.Share.GetEDS(cmd.Context(), &root)
+		return cmdnode.PrintOutput(shares, err, nil)
+	},
+}
+
+func parseJSON(param string) (json.RawMessage, error) {
```
ramin, line 205: don't love this sitting here...thoughts about moving it to cmd/util if we must have it?

ramin (approved):
obviously we've been reviewing these pretty thoroughly as each change got worked into the branch, so i've given this a light skim and it all looks familiar and i remember most discussion, so pretty lightly reviewed with just a couple of comments now seeing it organized as a whole.

massive work, good job

cmd/rpc.go
vgonkivs: okie. Will change. Would you mind, if I will do it in a subsequent PR?

nodebuilder/share/cmd/share.go
vgonkivs: The main concern that I have is that `parseJSON` is used only in share/cmd

ramin: i know! i agree with you, it's convenient only here, but also feels like it doesn't belong, lets leave it there, i'm good

cmd/rpc.go
ramin: YES! even better, as soon as this merges, i will do it

ramin (approved):

distractedm1nd (dismissed):

distractedm1nd:
oh wait, @vgonkivs can you make this go into rpc-breaks?
we will try to merge rpc-breaks today

renaynay (dismissed):
lfg

liamsi (approved):

ramin (approved):

Issue #2807, opened 2023-10-04
share/p2p: data race in peer manager


vgonkivs:
Found a data race during execution our swamp tests with -race flag:
```
WARNING: DATA RACE
(https://github.com/celestiaorg/celestia-node/actions/runs/6405300053/job/17388103459?pr=2802#step:5:866)
Write at 0x00c09d1b8b80 by goroutine 35395:
(https://github.com/celestiaorg/celestia-node/actions/runs/6405300053/job/17388103459?pr=2802#step:5:867)
  sync/atomic.StoreInt64()
(https://github.com/celestiaorg/celestia-node/actions/runs/6405300053/job/17388103459?pr=2802#step:5:868)
      /opt/hostedtoolcache/go/1.21.1/x64/src/runtime/race_amd64.s:237 +0xb
(https://github.com/celestiaorg/celestia-node/actions/runs/6405300053/job/17388103459?pr=2802#step:5:869)
  sync/atomic.StorePointer()
(https://github.com/celestiaorg/celestia-node/actions/runs/6405300053/job/17388103459?pr=2802#step:5:870)
      /opt/hostedtoolcache/go/1.21.1/x64/src/runtime/atomic_pointer.go:82 +0x44
(https://github.com/celestiaorg/celestia-node/actions/runs/6405300053/job/17388103459?pr=2802#step:5:871)
  github.com/celestiaorg/celestia-node/share/p2p/peers.(*Manager).newPeer.(*Manager).doneFunc.func1()
(https://github.com/celestiaorg/celestia-node/actions/runs/6405300053/job/17388103459?pr=2802#step:5:872)
      /home/runner/work/celestia-node/celestia-node/share/p2p/peers/manager.go:280 +0x32a

Previous read at 0x00c09d1b8b80 by goroutine 36050:
(https://github.com/celestiaorg/celestia-node/actions/runs/6405300053/job/17388103459?pr=2802#step:5:895)
  github.com/celestiaorg/celestia-node/share/p2p/peers.(*syncPool).add()
(https://github.com/celestiaorg/celestia-node/actions/runs/6405300053/job/17388103459?pr=2802#step:5:896)
      /home/runner/work/celestia-node/celestia-node/share/p2p/peers/manager.go:529 +0x4d
```

[...]

PR #2817, opened 2023-10-06
fix(share/peer-manager): use mutex for peer-manager pool cleanup

walldiss:
Resolves https://github.com/celestiaorg/celestia-node/issues/2807

### DIFF
--- a/share/p2p/peers/pool.go
+++ b/share/p2p/peers/pool.go
@@ -224,8 +224,9 @@ func (p *pool) len() int {
 
 // reset will reset the pool to its initial state.
 func (p *pool) reset() {
-	p.m.Lock()
-	defer p.m.Unlock()
+	lock := &p.m
+	lock.Lock()
+	defer lock.Lock()
 	// swap the pool with an empty one
 	*p = *newPool(time.Second)
 }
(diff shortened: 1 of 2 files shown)


ramin (dismissed):

distractedm1nd (dismissed):

Wondertan (dismissed):

renaynay (dismissed):

renaynay (commented):
https://github.com/celestiaorg/celestia-node/actions/runs/6429244191/job/17457937102?pr=2817#step:4:72

@walldiss  ^

share/p2p/peers/pool.go
```diff
@@ -221,3 +221,12 @@ func (p *pool) len() int {
...
 }
+
+// reset will reset the pool to its initial state.
+func (p *pool) reset() {
+	lock := &p.m
+	lock.Lock()
+	defer lock.Lock()
+	// swap the pool with an empty one
+	*p = *newPool(time.Second)
+}
```
Wondertan, line 232: So resetting works here by substituting existing pool with new pool. What if avoid allocating new pool by resetting each field on existing pool?

walldiss: Would be almost the same, because it would still need to realloc struct fields objects and maps. Probably would save 1 or 2 less pointer allocation. This reset operation happens once in 15 sec, so it is nothing for runtime. So I've optimised it for code reader performance 🙂

Wondertan (approved):

distractedm1nd (approved):

PR #2818, opened 2023-10-06
chore: simple rename of cmd helper, ParseFromCtx to NewClientFromContext

ramin:
discussed we'd do as a follow up after rpc merge yesterday with @vgonkivs, happy to wholeheartedly rejected here, but my feeling is

`NewClientFromContext` is a more idiomatic name for this than `ParseClientFromCtx`


Wondertan:
My concern here is that we don't instantiate the client, while New makes it seems like so. Get instead of New would be more appropriate, imo. 

However, I do think that we should instantiate the Client in this func in future, rather than creating it elsewhere and passing through ctx, thus not opposed to keep the change.

ramin:
@Wondertan oh man, i missed that is what was happening, I'm going to modify this so it really does take some of the flag stuff, set in context, and then instantiate the client, will be a little more "what you'd expect" then.

renaynay (approved):

Issue #2768, opened 2023-09-26
API version does not match celestia-node version


jcstein:
When checking out and building v0.11.0-rc14, and then running `make openrpc-gen` -- the resulting JSON shows`"version": "v0.2.2"`, when it should match v0.11.0-rc14

jcstein:
confirming that the release of `v0.11.0-rc15` has `v0.11.0` as the API version

PR #2825, opened 2023-10-09
fix(rpc)!: various rpc breaks

distractedm1nd:
RPC Break feature branch. Needs to not be squashed


distractedm1nd:
idk why it added me to every commit, can I change that? It could be from the rebase?

vgonkivs (changes requested):
Did you rebase it on the latest main? I feel there should be issues inside `nodebuilder/das/cmd`. Also, could you please go through all cmds and ensure that nothing were broken?

nodebuilder/node/admin.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/celestiaorg/celestia-node/libs/authtoken"
 )
 
-const APIVersion = "v0.2.2"
+const APIVersion = "v0.11.0"
```
Wondertan, line 13: So now we have to PR a string change every time we update the version. I wonder if there is a way to automate that.

Wondertan (dismissed):

vgonkivs:
cc @jcstein for visibility
Recent changes in the node's API:
* ProbalitiyOfAvailability endpoint(and cmd) was completely removed;
* GetShare, GetEDS, GetSharesByNamespace, and SharesAvailable now take not the DAH, but the full ExtendedHeader
* fraud.Subscribe returns `<-*Proof `instead of `<-Proof`

nodebuilder/node/admin.go
distractedm1nd: Dont think so unfortunately :/ This is one unforeseen issue with #2768

renaynay (dismissed):

Wondertan (dismissed):

ramin (dismissed):

walldiss (approved):

vgonkivs (approved):

renaynay (dismissed):

Wondertan (approved):

ramin (dismissed):

vgonkivs (dismissed):

renaynay (approved):

Wondertan (approved):

Issue #2808, opened 2023-10-05
blob not found by blob.Get, however it is listed in blob.GetAll


vvuwei:
### Celestia Node version

v0.11.0-rc14

### OS

Alpine Linux v3.18

### Install tools

installed from docker image `ghcr.io/celestiaorg/celestia-node:v0.11.0-rc14`
by docker compose with command:

```dockerfile
command: bash -c "
        celestia full init --p2p.network mocha-4 &&
        celestia full start \
          --rpc.addr 0.0.0.0 \
          --rpc.port 26658 \
          --p2p.network mocha-4
          --p2p.mutual \"..."
      "
```

### Others

getting **read** token by running command:

`celestia full auth read --p2p.network mocha-4`

all requests send with that token as Bearer Token .

### Steps to reproduce it

1. sending request to node
```JSON
{
    "jsonrpc": "2.0",
    "id":      1,
    "method":  "blob.GetAll",
    "params":  [
        20345,
        ["AAAAAAAAAAAAAAAAAAAAAAAAAMAOXWfCdVOJre0="]
    ]
}
```

2. getting as result
```JSON
{
	"jsonrpc": "2.0",
	"result": [
		// ...
		{
			"namespace": "AAAAAAAAAAAAAAAAAAAAAAAAAMAOXWfCdVOJre0=",
			"data": "569J69FEKedAt2NVXhxMHT1iUiM4bdEHcckaBD0sIdImLjE52AbXxQeQyWBeYO1orDA7/ODEglI59wfgCym6E2Zyw8YLMbmp3Q6fGDtZNp6VFuZsPxG/3fHljwmgdA3ilCktXw==",
			"share_version": 0,
			"commitment": "R/UBOXrYtemjlAwGKfskbvA+LoMPYUU1ZZTsMhTZgg0="
		},
		// ...
	],
	"id": 1
}
```

3. sending request to node
```JSON
{
    "jsonrpc": "2.0",
    "id":      1,
    "method":  "blob.Get",
    "params":  [
        20345,
        "AAAAAAAAAAAAAAAAAAAAAAAAAMAOXWfCdVOJre0=",
        "R/UBOXrYtemjlAwGKfskbvA+LoMPYUU1ZZTsMhTZgg0="
    ]
}
```

4. getting as result
```JSON
{
  "jsonrpc":  "2.0",
  "id": 1,
  "error": {
    "code": 1,
    "message":  "blob: not found"
  }
}
```

### Expected result

[...]

PR #2828, opened 2023-10-09
fix(blob/service): fix getByCommitment

vgonkivs:
## Overview
Fixes https://github.com/celestiaorg/celestia-node/issues/2808

Root cause: I did not check the entire row once the first blob was identified(and moved to the next row). So, in the end, we had a lot of unchecked blobs. Reworked logic now goes through the whole rawShares, until it is possible, constructing blobs and verifying them.
Keeping it in draft until I find a way to simplify the code because now it looks complicated
## Checklist


blob/blob.go
```diff
@@ -138,3 +139,37 @@ func (b *Blob) UnmarshalJSON(data []byte) error {
...
+
+		length, err := appShares[0].SequenceLen()
+		if err != nil {
+			return nil, nil, err
+		}
+
+		amount := shares.SparseSharesNeeded(length)
+		if amount > len(appShares) {
+			return blobs, appShares, nil
+		}
```
renaynay, line 160: can you document this case in a bit more detail? When would it happen that the shares would not be able to form a complete blob for a namespace at this point? Does this mean the caller wasn't able to recover the full shares of the namespace?

blob/service.go
```diff
@@ -198,10 +198,10 @@ func (s *Service) getByCommitment(
 	}
 
 	var (
-		rawShares = make([]share.Share, 0)
+		rawShares = make([]shares.Share, 0)
 		proofs    = make(Proof, 0)
-		amount    int
-		blobShare *shares.Share
+		// expanded specifies whether blob is expanded into multiple rows
+		expanded bool
```
renaynay, line 204: let's call it something else like `spansMultipleRows bool`

renaynay: or `overflowsRow`

blob/service.go
```diff
@@ -211,83 +211,38 @@ func (s *Service) getByCommitment(
...
 			return nil, nil, err
 		}
-		if same {
-			return blob, &proofs, nil
+		for _, b := range blobs {
+			if b.Commitment.Equal(commitment) {
+				return b, &proofs, nil
+			}
+			if expanded {
+				// only the first blob in the range can be expanded
```
renaynay, line 233: what do you mean by this? It's not possible that another blob in the set of blobs by namespace can overflow the row?

blob/service.go
```diff
@@ -304,3 +259,25 @@ func (s *Service) getBlobs(
 	}
 	return SharesToBlobs(namespacedShares.Flatten())
 }
+
+// toAppShares converts node's raw shares to the app shares, skipping padding
+func toAppShares(shrs ...share.Share) ([]shares.Share, error) {
+	appShrs := make([]shares.Share, 0)
```
renaynay, line 265: do you want to allocate up to cap here too?

blob/helper.go
```diff
@@ -17,16 +17,16 @@ func SharesToBlobs(rawShares []share.Share) ([]*Blob, error) {
...
+	appShares, err := toAppShares(rawShares...)
+	if err != nil {
+		return nil, err
 	}
+	return parseShares(appShares)
+}
 
-	shareSequences, err := shares.ParseShares(appShares, true)
+// parseShares takes shares and converts them to the []*Blob.
+func parseShares(appShrs []shares.Share) ([]*Blob, error) {
```
renaynay, line 28: what's the difference between this and `createBlobs` ?

renaynay (commented):
I'm a bit confused around some naming in this PR

blob/service.go
vgonkivs: No, because we are going row by row appending row.Shares to the shares from the previous rows(if they have existed), and they could be existed **only** if the blob spans multiple rows

renaynay: Okay but what do you mean by "only the first blob in the range" ?

blob/service.go
Wondertan: It's okay to over-allocate slices usually. The reason is that you still do one allocation compared to N times depending on the len of `shrs`. As you know, each time you don't have enough space in the slice, the new 2x one is allocated.

blob/service.go
```diff
@@ -198,10 +198,10 @@ func (s *Service) getByCommitment(
 	}
 
 	var (
-		rawShares = make([]share.Share, 0)
+		rawShares = make([]shares.Share, 0)
```
Wondertan, line 201: is `raw` intentional or is this a typo of `row`?

Wondertan (commented):
The logic looked correct to me and I don't have any simplification in mind. 
We should add a regression test for the fixed case.

blob/blob.go
vgonkivs: changed the comment of the whole method

blob/service.go
vgonkivs: renamed

blob/service.go
vgonkivs: True.

blob/helper.go
vgonkivs: renamed `createBlobs`  -> `buildBlobsIfExist`

blob/service.go
vgonkivs: Intentional

blob/blob.go
```diff
@@ -138,3 +139,38 @@ func (b *Blob) UnmarshalJSON(data []byte) error {
...
+		if err != nil {
+			return nil, nil, err
+		}
+
+		// only 1 blob will be created bc we passed the exact amount of shares
+		blobs = append(blobs, b[0])
+
+		if amount == len(appShares) {
+			return blobs, nil, nil
+		}
```
walldiss, line 173: Looks like this if could be removed, since len(appShares) == 0  will handle same case.

blob/service_test.go
```diff
@@ -344,6 +344,25 @@ func TestService_GetSingleBlobWithoutPadding(t *testing.T) {
...
+	appBlobs, err := blobtest.GenerateV0Blobs(sizes, true)
+	require.NoError(t, err)
+	blobs, err := convertBlobs(appBlobs...)
+	require.NoError(t, err)
+
+	service := createService(ctx, t, blobs)
+	for _, blob := range blobs {
+		b, err := service.Get(ctx, 1, blob.Namespace(), blob.Commitment)
+		require.NoError(t, err)
+		assert.Equal(t, b.Commitment, blob.Commitment)
```
Wondertan, line 362: Previously it would only work for the first Blob and with the fix it is now able to serve any blob in the namespace

Wondertan (approved):
Much much simpler and more readable then the initial solution. Well done

blob/blob.go
vgonkivs: otherwise, it could panic here:
`appShares = appShares[amount:]`

blob/service.go
```diff
@@ -211,83 +211,42 @@ func (s *Service) getByCommitment(
...
-		}
-
-		// move to the next row if the blob is incomplete.
-		if amount > len(rawShares) {
-			continue
-		}
-
-		blob, same, err := constructAndVerifyBlob(rawShares[:amount], commitment)
+		var blobs []*Blob
+		blobs, rawShares, err = buildBlobsIfExist(rawShares)
```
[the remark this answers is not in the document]
vgonkivs: Could you please specify what remaining shares do you mean:
* if there will be not the last iteration of the for loop, then these remaining shares will be applied to the next rows until our blob is completed
* if you mean remaining shares on the last iteration:
Theoretically, it's not possible. All blobs must be completed, converted to the shares, and added to the data data square. So, this means that we have a bug somewhere in our code because we can't have incomplete blobs. By somewhere I mean either getter or blob service.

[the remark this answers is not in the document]
walldiss: If it is impossible for last iteration, let’s add check outside the loop, to make sure nothing left. So we can capture the bug if it exist. As of now if bug is there code will silently omit remaining shares.

blob/blob.go
walldiss: If amount == 0 it should not panic

blob/service.go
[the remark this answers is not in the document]
vgonkivs: In the private discussion, we discussed with @walldiss two ways to handle that:
* Vlad's suggestion is to create a custom error if we have remaining shares after the loop ends. It will help users to detect the bug and properly report it.
* My opinion is that we don't need to because this is an internal bug(with storage/getters/blob-service) and returning an error will not be informative to the user. Instead, we should add a `log.Error("THIS IS A BUG. INCOMPLETE BLOB/s DETECTED") `and return add ErrBlobNotFound(if user requests blob or proof by commitment then he expects to get it back and he is almost 100% sure that this blob will be available at the particular height under the particular namespace). 

As we could not come to a consensus, we need you @Wondertan , @renaynay , @ramin

blob/service_test.go
vgonkivs: not exactly for the first blob, but yeah some blobs might be missed

blob/service.go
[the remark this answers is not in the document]
Wondertan: Note that the result of `GetSharesByNamespace` gives you _verified_ and _complete_ data, meaning that this case where there are some remaining shares left is only possible if validators are malicious and made a malformed block.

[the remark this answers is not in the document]
Wondertan: However, it would still be nice to report that to users immediately as fraud proofs may take time to propagate, and there might be bugs on our side, so some form of logging would be helpful.

On error reporting, I think we should return an error only if the user didn't get the data he requested. If there are some redundant shares, but the user got what they requested, this will be caught by FP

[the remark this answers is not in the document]
vgonkivs: This case only if the user did not get what he wanted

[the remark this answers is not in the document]
Wondertan: If the question is to return an error or not, when the user can't get the data they want, we should definitely give him an error instead of half baked blob or blocking.

blob/service.go
```diff
@@ -211,84 +211,50 @@ func (s *Service) getByCommitment(
...
-		return nil, nil, ErrBlobNotFound
+		proofs = nil
 	}
 
-	blob, same, err := constructAndVerifyBlob(rawShares, commitment)
-	if err != nil {
-		return nil, nil, err
+	err = ErrBlobNotFound
+	if len(rawShares) > 0 {
+		reportErr := fmt.Errorf("THIS IS A BUG: INCOMPLETE BLOB DETECTED at: %d height", height)
```
Wondertan, line 253:
```suggestion
		err = fmt.Errorf("incomplete blob detected at %d: %w", height, err)
```

This might not be a bug necessary, but a malformed block. In this case you can wrap error without join

blob/service.go
```diff
@@ -198,10 +198,10 @@ func (s *Service) getByCommitment(
 	}
 
 	var (
-		rawShares = make([]share.Share, 0)
+		rawShares = make([]shares.Share, 0)
 		proofs    = make(Proof, 0)
-		amount    int
-		blobShare *shares.Share
+		// expanded specifies whether blob is expanded into multiple rows
```
renaynay, line 203: comment should match, but nit

blob/service.go
```diff
@@ -211,84 +211,49 @@ func (s *Service) getByCommitment(
...
-		return nil, nil, ErrBlobNotFound
+		proofs = nil
 	}
 
-	blob, same, err := constructAndVerifyBlob(rawShares, commitment)
-	if err != nil {
-		return nil, nil, err
+	err = ErrBlobNotFound
+	if len(rawShares) > 0 {
+		err = fmt.Errorf("incomplete blob detected at %d: %w", height, err)
```
renaynay, line 253: nit: add namespace too? It's valuable info

blob/service.go
renaynay: So do you want to alloc to cap here or no?

renaynay (approved):

blob/service.go
renaynay, line 203: another nit, but can you move the declaration of these vars closer to their use?

vgonkivs: done

vgonkivs: done

blob/service.go
vgonkivs: added

blob/service.go
vgonkivs: yes, updated. Thank you

renaynay (approved):

Wondertan (approved):

Issue #2823, opened 2023-10-06
nodebuilder | nodebuilder/p2p: Hardcode mainnet chainID, change defaults to point to mainnet 


renaynay:
```[tasklist]
### Tasks
- [x] Hardcode mainnet chainID + add bootstrappers
- [x] Change default to point to mainnet instead of mocha
- [x] Ensure node store path does not append the network name if it is mainnet (only for testnets / non-default nets), so node store should be `~/.celestia-light` for mainnet and `~/.celestia-light-mocha` for `mocha`
```

renaynay:
Done

PR #2840, opened 2023-10-12
chore: adding mainnet chain id and setting as default

distractedm1nd:
:)
Also removed blockspace race
Part of #2823


ramin (approved):

nodebuilder/p2p/network.go
```diff
@@ -52,20 +52,20 @@ func (n Network) String() string {
 
 // networksList is a strict list of all known long-standing networks.
 var networksList = map[Network]struct{}{
-	Arabica:        {},
-	Mocha:          {},
-	BlockspaceRace: {},
-	Private:        {},
+	Mainnet: {},
```
Wondertan, line 55: [discussion]
We may wanna call it just Celestia. We don't have `Testnet` and we have names for all other testnets.

ramin: hmmmm, yes i think i agree with this

renaynay: p2p.Celestia

I think I _prefer_ `mainnet` but I'm fine with whatever. Not a big deal to me

renaynay (approved):

nodebuilder/p2p/network.go
distractedm1nd: I like Mainnet more but am also not very opinionated. in the worst case we can change it later if we don't like it without breaking anything

ramin: agree, don't feel very strong either way. BUT, with naming i always err on the side of "what do we say when we talk about it" and we're much more likely to say "on Arabica" or "if we are on Mainnet then we should remove the suffix" vs "if we are on Celestia". So....maybe just rolling with Mainnet is correct

distractedm1nd: also as rene said, I feel better about p2p.Mainnet naming than p2p.Celestia, which might be confusing

Wondertan: To me, it's similarly confusing as p2p.Arabica

Wondertan: > "if we are on Celestia"

When we say on Celestia it should mean on mainnet by default IMO

PR #2842, opened 2023-10-12
feat(modshare)!: versioning for discovery

### LINKED ISSUE #2578: share/discovery: Implement versioning for `full` advertising string

Wondertan:
Closes #2578

### DIFF
--- a/nodebuilder/share/constructors.go
+++ b/nodebuilder/share/constructors.go
@@ -22,7 +22,7 @@ import (
 
 const (
 	// fullNodesTag is the tag used to identify full nodes in the discovery service.
-	fullNodesTag = "full"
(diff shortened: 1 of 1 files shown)


renaynay:
Note: once merging, main will be incompatible with `arabica` and `mocha`

walldiss:
Bump could be done in backwards compatible manner if there are 2 instances of discovery running: with old and new tag. The downside is that in worst case scenario, it will ensure 2x more connections to full nodes.

renaynay:
@walldiss I think we plan to upgrade arabica + mocha anyway soon

PR #2846, opened 2023-10-13
fix(share/getters): short-circuit on empty root in ShrexGetter

distractedm1nd:
Should fix issue where GetEDS hangs on empty data root

To test run:

`celestia share get-eds “$(celestia header get-by-height 500 --node.store ~/.celestia-light-mocha-4 | jq .result)” --node.store ~/.celestia-light-mocha-4`

with and without the PR

### DIFF
--- a/share/getters/shrex.go
+++ b/share/getters/shrex.go
@@ -138,7 +138,6 @@ func (sg *ShrexGetter) GetEDS(ctx context.Context, header *header.ExtendedHeader
 	dah := header.DAH
 	// short circuit if the data root is empty
 	if dah.Equals(share.EmptyRoot()) {
-		fmt.Println("hitting")
 		return share.EmptyExtendedDataSquare(), nil
 	}
 	for {
--- a/share/getters/shrex.go
+++ b/share/getters/shrex.go
@@ -136,6 +136,11 @@ func (sg *ShrexGetter) GetEDS(ctx context.Context, header *header.ExtendedHeader
 	}()
 
 	dah := header.DAH
+	// short circuit if the data root is empty
+	if dah.Equals(share.EmptyRoot()) {
+		fmt.Println("hitting")
+		return share.EmptyExtendedDataSquare(), nil
+	}
 	for {
 		if ctx.Err() != nil {
 			sg.metrics.recordEDSAttempt(ctx, attempt, false)


Wondertan:
So I tested this on LN, and I never triggered `hitting`. I found though a weird behaviour. It's guaranteed that LN gets an empty block on the first time with slight delay, but it always stuck on the further tries.

Wondertan:
Note, we need to add empty block test case to our Getter test suite

distractedm1nd:
Hmm, a swamp test of a LN calling twice resolves both times and logs hitting - I wonder what is different between swamp and the real call. Are you sure the node was running the right build?

Wondertan:
I'lol double check tmrw @distractedm1nd

distractedm1nd:
It is logging hit for me every time, and doesn't hang ever even on multiple runs 🤔

But on main, it retrieves after the first time, logging
```
2023-10-16T14:05:26.436+0200	ERROR	header/p2p	p2p/exchange.go:166	verifying head received from tracked peer	{"tracked peer": "12D3KooWS7CW7y1VRuHif4nJqLyUyUtHaCFmJDEfTd9wMpgFucNk", "height": 290816, "err": "header verification failed: known header: '290816' <= current '290816'"}
2023-10-16T14:05:26.437+0200	ERROR	header/p2p	p2p/exchange.go:166	verifying head received from tracked peer	{"tracked peer": "12D3KooWHHUVwvXgdxg4k2GtfSA6yB6iDJ1fBYnvBU9LhSMzoi2A", "height": 290816, "err": "header verification failed: known header: '290816' <= current '290816'"}
2023-10-16T14:05:26.842+0200	ERROR	header/p2p	p2p/exchange.go:166	verifying head received from tracked peer	{"tracked peer": "12D3KooWQNp5N12C19cDvhDSAVLnW6tYjthZs1abi3mpXpSqwkpJ", "height": 290816, "err": "header verification failed: known header: '290816' <= current '290816'"}
2023-10-16T14:05:26.853+0200	ERROR	header/p2p	p2p/exchange.go:166	verifying head received from tracked peer	{"tracked peer": "12D3KooWHEU9efXKk5PeCyWyNVz2YYwW8bzxv25vjtfiJN6fLpTE", "height": 290816, "err": "header verification failed: known header: '290816' <= current '290816'"}
```

And then fails the second time.

So confirmed that this is fixing the issue

walldiss (approved):

share/getters/shrex.go
```diff
@@ -136,6 +136,10 @@ func (sg *ShrexGetter) GetEDS(ctx context.Context, header *header.ExtendedHeader
 	}()
 
 	dah := header.DAH
+	// short circuit if the data root is empty
+	if dah.Equals(share.EmptyRoot()) {
+		return share.EmptyExtendedDataSquare(), nil
+	}
```
renaynay, line 142: should we just have this first? Does it even make sense to start the span for this?

distractedm1nd: yes

renaynay (dismissed):

renaynay (approved):

PR #2853, opened 2023-10-17
fix(blob): properly wrap error in GetAll method

tzdybal:
## Overview

Error returned by `GetAll` is not formatted correctly. It's especially important, because users of the method will most probably try to detect `ErrBlobNotFound` (to differentiate from other errors). 

## Checklist


Wondertan (approved):
Thanks!

vgonkivs (approved):

Issue #1584, opened 2023-01-10
ci: create jobs for different configurations


vgonkivs:
We have faced issues when tests pass on Linux and fail on MacOs(and vice versa). I think we should create jobs and run tests on both platforms. Not sure, that we have to run them on each push, but we should definitely do this before our changes will be merged in the main branch.

MSevey:
you can add macos as a os type in the github actions
```
jobs:
  build:
    runs-on: macos-latest
```
then we can run the CI on both like so

```
jobs:
  example_matrix:
    strategy:
      matrix:
        os: [ubuntu-latest, macos-latest]
    runs-on: ${{ matrix.os }}
```

PR #2855, opened 2023-10-18
(ci) matrix tests for different OS

ramin:
Refs https://github.com/celestiaorg/celestia-node/issues/1584

Lets run our test suite on multi configurations as they sometimes pass one not other (as reported by @vgonkivs)

- Task: run tests on > 1 OS (previously ubuntu-latest only) now `ubuntu-latest` and `macos-latest`
- currently ONLY for the unit tests (do we want to change this for all the tests?)
- also makes a small unrequested tweak to make the execution in go-ci run AFTER the mod and lint checks, so it'll pause if those two fail and fail the entire pipeline faster. Save some wasted compute time, which will be at least x2 worse once we matrix these

example run: https://github.com/celestiaorg/celestia-node/actions/runs/6559066381/job/17813974060

### DIFF
--- a/.github/workflows/go-ci.yml
+++ b/.github/workflows/go-ci.yml
@@ -63,6 +63,7 @@ jobs:
         run: make test-unit
 
       - name: upload coverage
+        if: ${{ matrix.os == 'macos-latest' }}
         uses: codecov/codecov-action@v3.1.4
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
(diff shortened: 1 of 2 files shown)


renaynay (dismissed):

vgonkivs (dismissed):

walldiss (dismissed):

Wondertan (dismissed):

renaynay (dismissed):

vgonkivs (dismissed):

Wondertan (approved):
uTack

Wondertan (approved):
utACK

distractedm1nd (approved):
utACK

Wondertan (approved):

PR #2863, opened 2023-10-19
feat(core): Enable metrics for core package

renaynay:
While we do plan to remove the `bridge` node type eventually, it's useful to have some visibility into the core package in the short term.


core/metrics.go
```diff
@@ -0,0 +1,78 @@
...
+		m.blockTime = now
+	})
+}
+
+func (m *metrics) subscriptionStuck(ctx context.Context) {
+	m.observe(ctx, func(ctx context.Context) {
+		m.subscriptionStuckInst.Add(
+			ctx,
+			1,
+			metric.WithAttributes(attribute.String(subscriptionStuckTimestampKey, time.Now().String())),
```
Wondertan, line 75: Cardinality

core/metrics.go
```diff
@@ -0,0 +1,78 @@
...
+
+	subscriptionStuckInst metric.Int64Counter
+}
+
+func newMetrics() (*metrics, error) {
+	m := new(metrics)
+
+	var err error
+	m.blockTimeInst, err = meter.Float64Histogram(
+		"core_listener_block_time",
```
Wondertan, line 28: add _hist

core/metrics.go
```diff
@@ -0,0 +1,78 @@
...
+	if m == nil {
+		return
+	}
+
+	if ctx.Err() != nil {
+		ctx = context.Background()
+	}
+
+	observeFn(ctx)
+}
```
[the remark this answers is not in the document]
renaynay: can make a good first issue

Wondertan (changes requested):
I realized that we don't need blockTime metric. The bridge node publishes headers through Subscriber, which goes through that validator, and time is tracked on it. So effectively we will get the same data twice

renaynay:
@Wondertan the reason i wanted to capture block time here is that it's the only way to know on the DA-side what the approximate block time is so that we can see if there's a difference between core network block time and DA header propagation time. 

I removed it in this commit https://github.com/celestiaorg/celestia-node/pull/2863/commits/403bc9a924410d5675c652e16017cca5b049efc3

Wondertan:
I understand and it turns out we designed things in a way that we don't need to duplicate anything 😺

Wondertan:
We should aim to include fetcher metrics here as well

Wondertan:
Linter btw

Wondertan (approved):
LGTM

distractedm1nd (approved):

vgonkivs (approved):

ramin (approved):

Issue #1211, opened 2022-10-10
ci: Consider using artifacts for CI workflows for our debug logs


renaynay:
Ref: 
* https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts

Related: 
* https://github.com/celestiaorg/celestia-node/pull/1210/files

PR #2865, opened 2023-10-20
(ci) enable optional verbose logging on unit tests and upload artifacts

ramin:
- Optionally enabled verbose logging for unit tests
- adds ability to verbose log to a file, while sending "normal" quiet output to stdout
- updates CI to upload the test.log artifact and store for 5 days for later debugging
- local workflow and test runs will stay the same
- only enabled for `make test-unit` for now as a proposal for how we achieve this, can expand to rest if we like it @renaynay and @vgonkivs 

If everyone likes, i can expand to all test cases

fixes https://github.com/celestiaorg/celestia-node/issues/1211


ramin:
ok, check it out:

https://github.com/celestiaorg/celestia-node/actions/runs/6586192504?pr=2865#artifacts

You'll see we can download the unit test log and coverage, but the stdout in the github action log remain succinct

distractedm1nd (approved):

Wondertan:
@ramin, do I understand correctly that this enable `Rerun with debug log feature` on GH?

ramin:
@Wondertan no this is specific to OUR `-v` for unit tests, so will generate the verbose output but pipe them to file vs making more verbose stdout in the github action runner and upload the file for looking at later

https://github.com/celestiaorg/celestia-node/issues/1211

Wondertan:
@ramin, that issue was made before the GH feature was introduced. It would be helpful to have it as well. Really, it's a rare case when we need full debug logs. Also, accessing simple CI failures like races and panics through GH is easier than downloading artifacts every time. Let's integrate [the feature](https://github.blog/changelog/2022-05-24-github-actions-re-run-jobs-with-debug-logging/) while we are on it, so we have the option to get debug logs, but only when it's necessary.

ramin:
@Wondertan  cool we have the re-run with debug logs available already so nothing to add there 

Screenshot 2023-10-23 at 10 42 08 AM

Wondertan:
Have you tried if it's working? Because it wasn't for me and I assumed it needs additional setup

renaynay (approved):
I'm fine with this, but not sure I'll use it in CI too much. Nice to have as an option regardless

Wondertan (approved):
🙏🏻

Issue #2590, opened 2023-08-21
Update golangci-lint to 1.54.1


Wondertan:

ramin:
ended up to 1.55 here https://github.com/celestiaorg/celestia-node/pull/2868 with some additional CI action cleanup to reduce noise

PR #2868, opened 2023-10-23
update golangci-lint

ramin:
- updates `golangci-lint` to 1.55 
- removes "double cache" when setting up golang, thus removing linter noise (`Cannot open file...file exists`) so easier to see actual lint errors

closes https://github.com/celestiaorg/celestia-node/issues/2590


vgonkivs (approved):

Wondertan (approved):

Wondertan:
@ramin, this is meant to close #2590

Issue #2679, opened 2023-09-08
[Feature Request]: Signed Binaries


MSevey:
### Summary

As a follow up from introducing [Goreleaser](https://goreleaser.com/) in #2661  consider signing binaries.

### Problem Definition

Without signatures, users can't verify the binary downloaded was the one built by the team.

### Proposal

REF: https://goreleaser.com/customization/sign/

The github workflow would need the following update:
```yml     
      # Import the GPG key from github secrets to sign the binaries
      - name: Import GPG key
        id: import_gpg
        uses: crazy-max/ghaction-import-gpg@v4
        with:
          gpg_private_key: ${{ secrets.GPG_SIGNING_KEY }}
          passphrase: ${{ secrets.GPG_PASSPHRASE }}
      # Generate the binaries, release, and sign the checksum
      - uses: goreleaser/goreleaser-action@v4
        with:
          distribution: goreleaser
          version: latest
          args: release --clean
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          BuildTags: ${{ needs.build_tags.outputs.BUILDTAGS }}
          BuildTagsCommaSep: ${{ needs.build_tags.outputs.BUILDTAGSCOMMASEP }}
          GPG_FINGERPRINT: ${{ steps.import_gpg.outputs.fingerprint }}
```
The `.goreleaser.yaml` file needs the following update:
```yml
signs:
  - artifacts: checksum
    args: ["--batch", "-u", "{{ .Env.GPG_FINGERPRINT }}", "--output", "${signature}", "--detach-sign", "${artifact}"]
```

A GPG secret key and passphrase will need to be created and added to the repo's github secrets. The secret's names shold be `GPG_SIGNING_KEY` and `GPG_PASSPHRASE`.

PizBernina:
kind reminder that this is still pending... node-app you already add the checksums, please do so here too
https://github.com/celestiaorg/celestia-app/issues/2573

PR #2869, opened 2023-10-23
signed binaries for goreleaser

ramin:
mimics the signed binaries by goreleaser as implemented in `celestia-app` so node can also generate signed binaries

@MSevey confirmed that we have an org level secret for `GPG_SIGNING_KEY` and `GPG_PASSPHRASE` so should require no repo setup

implementation for app: https://github.com/celestiaorg/celestia-app/commit/934fdeda062229f5c150e9dce74f05b0557ea123

Will follow up with some work to ensure this workflow actually gets triggered (currently the steps before do not pass and generate binaries)

closes: https://github.com/celestiaorg/celestia-node/issues/2679

### DIFF
--- a/.github/workflows/ci_release.yml
+++ b/.github/workflows/ci_release.yml
@@ -87,6 +87,12 @@ jobs:
       - uses: actions/setup-go@v4
         with:
           go-version: 1.21
+      - name: Import GPG key
+        id: import_gpg
+        uses: crazy-max/ghaction-import-gpg@v4
+        with:
+          gpg_private_key: ${{ secrets.GPG_SIGNING_KEY }}
+          passphrase: ${{ secrets.GPG_PASSPHRASE }}
       # Generate the binaries and release
       - uses: goreleaser/goreleaser-action@v5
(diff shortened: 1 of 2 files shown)


renaynay (approved):

MSevey:
Changes lgtm. Actions unfortunately still don't really have a great way to test beside live testing. 

For this one, I'd recommend making a new clean fork of the repo so that you can merge this change into that fork. Then update the CI workflow to bypass the tests so that the release is triggered so that you can then verify that the binaries and release are generated and then you can download and verify the binary as well.

Wondertan (approved):

MSevey (approved):

ramin:
@MSevey replicated and got this setup working and signing on one of our test tools https://github.com/ramin/waypoint/actions/runs/6645867234/job/18058097433#step:5:54, really only thing i can't test is that our GPG secrets are available and correct, so we'll have to do it live. Will get this entire pipeline running next so we can see it generate some artifacts

PR #2870, opened 2023-10-23
fix(docgen): fix das naming for the openrpc spec

vgonkivs:


Wondertan (approved):
Another why there should have been a constant per each module

renaynay (approved):

walldiss (approved):

PR #2874, opened 2023-10-26
(ci) mark flakey test suites as "continue-on-error: true"

ramin:
in the interest of having a full pipeline for release 0.12.0, have marked the swamp and unit w/ race tests to `continue-on-error: true` so though the job will still fail in ci, the pipeline and dependent steps after those tests (namely version_bump + goreleaser) will still proceed as though they were successful, and we can get signed binaries with the tagged release

### DIFF
--- a/.github/workflows/go-ci.yml
+++ b/.github/workflows/go-ci.yml
@@ -89,6 +89,7 @@ jobs:
 
       - name: execute test run
         run: make test-unit-race
+        continue-on-error: true
 
   integration_test:
     needs: [lint, go_mod_tidy_check]
@@ -105,6 +106,8 @@ jobs:
 
       - name: Swamp Tests
         run: make test-swamp
+        continue-on-error: true
 
       - name: Swamp Tests with Race Detector
         run: make test-swamp-race
+        continue-on-error: true


ramin:
@MSevey have a look at the changes i had to add to version_bump here, seems it hadn't run in a looooooooong time but was failing once i was skipping flakey tests, copied what we have in celestia-app

MSevey:
> @MSevey have a look at the changes i had to add to version_bump here, seems it hadn't run in a looooooooong time but was failing once i was skipping flakey tests, copied what we have in celestia-app

@ramin there is currently a bug in app's version bumping that I'm digging into. 
https://github.com/celestiaorg/celestia-app/issues/2688

Will circle back once I've fixed app to know if there are changes to reapply here.

MSevey:
@ramin I would add this update to the version_bump job https://github.com/celestiaorg/celestia-app/pull/2773/files

This ensures proper handling of any non master/main release branches.

MSevey (changes requested):
needs update to version bump.

ramin:
@MSevey ok this is up to date

MSevey (approved):

renaynay (approved):
utack

PR #2882, opened 2023-10-27
test(core): Fix core/listener tests

renaynay:
When we bumped go-header, it broke the listener tests and we didn't notice. I thought it was a deeper bug at first, but turns out we are now just returning an error that we were silently ignoring before (as it was only logged, not returned) in `subscriber.Stop`.


ramin (approved):

vgonkivs (approved):

core/listener_test.go
```diff
@@ -169,8 +169,13 @@ func createListener(
 ) *Listener {
 	p2pSub, err := p2p.NewSubscriber[*header.ExtendedHeader](ps, header.MsgID, p2p.WithSubscriberNetworkID(networkID))
 	require.NoError(t, err)
+
 	err = p2pSub.Start(ctx)
 	require.NoError(t, err)
+	err = p2pSub.SetVerifier(func(ctx context.Context, msg *header.ExtendedHeader) error {
+		return nil
+	})
+	require.NoError(t, err)
```
Wondertan, line 178: Actually, looking at the fix made me realize that this is an actual API problem on Subscriber. Setting Verifier is optional, while Unregister on Stop happens all the time, causing this error. I think we should either document that setting verifier is a must or ignore the error Unregister as it was before so that the call is idempotent

renaynay: I completely agree.

Wondertan:
Turns out the fix was accidentally merged in #2881

PR #2897, opened 2023-11-03
Update LICENSE

Karotoka:
Correct form of copyright based on advice from external counsel.


Karotoka:
On Mus' instructions.

PR #2898, opened 2023-11-03
Delete NOTICE

Karotoka:
This NOTICE can be deleted as updated License contains the required information.


Karotoka:
On Mus' instructions.

Issue #2860, opened 2023-10-18
blob: TestBlobService_Get is flaky


Wondertan:
```
--- FAIL: TestBlobService_Get (0.02s)
    service_test.go:221: 
        	Error Trace:	/home/runner/work/celestia-node/celestia-node/blob/service_test.go:221
        	            				/home/runner/work/celestia-node/celestia-node/blob/service_test.go:294
        	Error:      	Received unexpected error:
        	            	empty shares received
        	Test:       	TestBlobService_Get
    --- FAIL: TestBlobService_Get/not_included (0.00s)
        testing.go:1490: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test
 ```

ramin:
@vgonkivs temporarily added https://github.com/celestiaorg/celestia-node/blob/main/blob/service_test.go#L1 to skip the test on mac os to help with CI passing, just in case it suddenly looks fixed

vgonkivs:
Had a private discussion with @walldiss regarding this case and figured out that GetSharesByNamespace can behave differently:
1. Row1 = [`Namespace1`, `Namespace3`]
`Namespace1` < `Namespace2`
`Namespace2` < `Namespace3`
Requesting Namespace2 will return empty shares + non-inclusion proof.

2. Row1 = [Namespace1]
    Row2 = [Namespace3]
`Namespace1` < `Namespace2`
`Namespace2` < `Namespace3`
Requesting Namespace2 will return empty shares + empty proof.

Before that, we have never considered non-inclusion proofs for the Blob Module. cc @Wondertan 

@nashqueue do you need non-inclusion proofs as a result of `GetProof` if the blob does not exist? If yes, how we can handle case 2. where no shares and no proofs?

huutrong1994:
🖐Hello! Welcome to #CelestiaCommunity

We are happy to announce  ($TIA) is now available on DEX #PancakeSwap 💪

👉  Our ONLY Contract:  
 CA:      0xf2caebbb49a7733a7c5cff571af849a6b52d346c

🥞 BUY ON PANCAKESWAP:

 https://pancakeswap.finance/swap?outputCurrency=0xf2caebbb49a7733a7c5cff571af849a6b52d346c

▫️Name:   Celestia 
▫️Symbol : TIA 
▫️Decimals : 18

📆Launch on  #BINANCE at Oct 31, 16:00 (UTC)

PR #2911, opened 2023-11-07
fix(blob/service): handle absence proof

vgonkivs:
Fixes https://github.com/celestiaorg/celestia-node/issues/2860

Proof:

Снимок экрана 2023-11-07 в 16 49 57

### DIFF
--- a/blob/service.go
+++ b/blob/service.go
@@ -220,6 +220,9 @@ func (s *Service) getByCommitment(
 
 	for _, row := range namespacedShares {
 		if len(row.Shares) == 0 {
+			// the above condition means that we've faced with an Absence Proof.
+			// This Proof proves that the namespace was not found in the DAH, so
+			// we can return `ErrBlobNotFound`.
 			return nil, nil, ErrBlobNotFound
 		}
 
--- a/blob/service.go
+++ b/blob/service.go
@@ -211,6 +211,10 @@ func (s *Service) getByCommitment(
 		return nil, nil, err
 	}
 
+	if namespacedShares.IsAbsenceProof() {
+		return nil, nil, ErrBlobNotFound
+	}
+
 	var (
 		rawShares = make([]shares.Share, 0)
 		proofs    = make(Proof, 0)
--- a/blob/service.go
+++ b/blob/service.go
@@ -211,10 +211,6 @@ func (s *Service) getByCommitment(
 		return nil, nil, err
 	}
 
-	if namespacedShares.IsAbsenceProof() {
-		return nil, nil, ErrBlobNotFound
-	}
-
 	var (
 		rawShares = make([]shares.Share, 0)
 		proofs    = make(Proof, 0)
@@ -223,6 +219,10 @@ func (s *Service) getByCommitment(
 	)
 
 	for _, row := range namespacedShares {
+		if len(row.Shares) == 0 {
+			return nil, nil, ErrBlobNotFound
+		}
+
 		appShares, err := toAppShares(row.Shares...)
 		if err != nil {
 			return nil, nil, err
--- a/blob/service_test.go
+++ b/blob/service_test.go
@@ -1,5 +1,3 @@
-//go:build linux
-
 package blob
 
 import (
--- a/share/eds/utils.go
+++ b/share/eds/utils.go
@@ -123,7 +123,7 @@ func CollectSharesByNamespace(
 
 	rootCIDs := ipld.FilterRootByNamespace(root, namespace)
 	if len(rootCIDs) == 0 {
-		return nil, nil
+		return []share.NamespacedRow{}, nil
 	}
 
 	errGroup, ctx := errgroup.WithContext(ctx)
--- a/share/getter.go
+++ b/share/getter.go
@@ -96,11 +96,3 @@ func (row *NamespacedRow) verify(rowRoot []byte, namespace Namespace) bool {
 		rowRoot,
 	)
 }
-
-// IsAbsenceProof checks if `NamespacedShares.Proof` proves the absence of the namespace.ID.
-func (ns NamespacedShares) IsAbsenceProof() bool {
-	if len(ns) == 0 || (len(ns[0].Shares) == 0 && ns[0].Proof != nil) {
-		return true
-	}
-	return false
-}
--- a/share/getter.go
+++ b/share/getter.go
@@ -97,13 +97,14 @@ func (row *NamespacedRow) verify(rowRoot []byte, namespace Namespace) bool {
 	)
 }
 
+// IsAbsenceProof checks if `NamespacedShares.Proof` proves the absence of the namespace.ID.
 func (ns NamespacedShares) IsAbsenceProof() bool {
-	isAbsenceProof := false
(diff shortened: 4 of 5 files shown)


share/getter.go
```diff
@@ -96,3 +96,11 @@ func (row *NamespacedRow) verify(rowRoot []byte, namespace Namespace) bool {
 		rowRoot,
 	)
 }
+
+// IsAbsenceProof checks if `NamespacedShares.Proof` proves the absence of the namespace.ID.
+func (ns NamespacedShares) IsAbsenceProof() bool {
+	if len(ns) == 0 || (len(ns[0].Shares) == 0 && ns[0].Proof != nil) {
```
Wondertan, line 102: So it case of absence proof it gives you one empty share?

Wondertan: Feel like leaky abstraction on the GetByNamespace side. If we make it truly nil on the implementation side, then this PR would not be needed and the case would be handled by a general path.

vgonkivs: > So it case of absence proof it gives you one empty share?

It gives 0 shares and one proof

vgonkivs: > If we make it truly nil on the implementation side, then this PR would not be needed and the case would be handled by a general path.

I've discussed this with @walldiss. The whole point of the non-nil result is different cases:
* when requested namespace B is not found within the range of namespaces in a single row[A, C] -> in this case we can produce an absence proof.
* When the requested namespace B is not found within multiple rows: Row0[A], Row1[C] -> in this case we can't produce anything and return nil.

vgonkivs: But I agree that such behavior is unclear and we should do smth about it. But now I'd suggest creating a separate issue and moving the discussion there.

Wondertan: For both given cases, we can have a general path. For the user of `GetSharesByNamespace`, there is no difference if it's with absence proof or not, and even if there is, they can derive this information by looking into the DAH header or by looking if the Proof is present on the returned NamespaceShares.

I suggest only never returning an empty share for `GetSharesByNamespace`:
* It's confusing. There is no documentation on what that means, as well as it's not obvious.
* It requires users on top to keep track of this edge case when it can be avoided. (Leaky abstraction)

Wondertan: I don't think we should postpone this in the issue either:
* It's not complicated to fix this now
* If we postpone this now, it's very unlikely we will come back to this.

vgonkivs: Ok, the general path for this is to return nil for both cases. That means if we can't prove the inclusion of the data, then we don't have to prove its absence, correct?

Wondertan: For the path when there is an absence of proof, we must return it. We just need to avoid returning non empty shares row

Wondertan: With or without proofs, it should return NamespacedShares both of which should succeed when we call Verify on it.

vgonkivs: ok, so in case if data was not found, the results will be:
* `[]share.NamespacedRow{}`-> when it's not possible to build an Absence Proof;
*` []share.NamespacedRow{Proof: *nmt.Proof}` -> when the Absence Proof exists;

vgonkivs: But we will anyway need `IsAbsenceProof`

Wondertan: We would we need it?

vgonkivs: I meant not the method itself, but this condition inside the blob module. Because the logic iterates over each row collecting shares and proofs and in this case (`[]share.NamespacedRow{Proof: *nmt.Proof}`), we'll get an error

Wondertan: So we actually don't need to do anything else besides the if above.

vgonkivs: 👍 true, but I'd go with `break`

Wondertan: break works if we are sure that only one absence proof is possible

vgonkivs: Get accepts only one namespace, so technically we can't have more then 1 absence proof

Wondertan: I don't remember if you can have multiple proofs over multiple rows with one namespace

vgonkivs: if NamespaceB falls in between Rows -> Row0[NamespaceA], Row1[NamespaceC], where 
`NamespaceA < NamespaceB < NamespaceC`, then no Absence proof is expected.

Wondertan: When I was trying to make an example below, I realized that the case below is not possible. Because that would imply B is actually there
* Row1[NamespaceA], Row1[NamespaceB]
* Row2[NamespaceB], Row2[NamespaceC]

Wondertan: Break is fine

blob/service_test.go
```diff
@@ -1,5 +1,3 @@
-//go:build linux
```
Wondertan, line 1: It's kinda surprising that this was anywhere helpful. Do you have any clue?

blob/service.go
```diff
@@ -219,6 +219,10 @@ func (s *Service) getByCommitment(
 	)
 
 	for _, row := range namespacedShares {
+		if len(row.Shares) == 0 {
+			return nil, nil, ErrBlobNotFound
```
Wondertan, line 223: So, to avoid incomplete blob error?

share/eds/utils.go
```diff
@@ -123,7 +123,7 @@ func CollectSharesByNamespace(
 
 	rootCIDs := ipld.FilterRootByNamespace(root, namespace)
 	if len(rootCIDs) == 0 {
-		return nil, nil
```
Wondertan, line 126: We return non-nil everywhere else, correct?

blob/service.go
vgonkivs: Yes. It's also not possible, but I'd like to keep it as safe as possible.

blob/service_test.go
vgonkivs: Yes, it was added by @ramin because macOs builds were flaky. We agreed to remove it once I fix this issue.

blob/service.go
Wondertan: Let's add a little comment describing why this if is special and why does it mean that the whole operation is cancelled.

share/eds/utils.go
vgonkivs: Correct. We create an empty slice on line 130 and then fulfil it further or return an empty slice

Wondertan: I mean in everywhere in other getters

vgonkivs: found one more place. Thanks

Wondertan (approved):

blob/service.go
```diff
@@ -219,6 +219,13 @@ func (s *Service) getByCommitment(
 	)
 
 	for _, row := range namespacedShares {
+		if len(row.Shares) == 0 {
```
renaynay, line 222: shouldn't this only happen in the case that len(namespacedShares) == 1 ?

vgonkivs: Yes, we expect to get `len(namespacedShares) == 1` from getter

renaynay (approved):

PR #2912, opened 2023-11-07
chore(modstate): Remove IsStopped endpoint from StateModule

### LINKED ISSUE #2906: modstate: Remove `IsStopped` endpoint on StateModule
We were supposed to remove it before 0.12 as part of all breaking changes removal, but it seems to be forgotten

Chirag018:
@Wondertan is this what you expected?

Closes #2906


api/gateway/middleware.go
```diff
@@ -41,7 +41,7 @@ func checkPostDisabled(state state.Module) mux.MiddlewareFunc {
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			// check if state service was halted and deny the transaction
-			if r.Method == http.MethodPost && state.IsStopped(r.Context()) {
+			if r.Method == http.MethodPost {
```
vgonkivs, line 44: in this case, POST requests will be disabled.

vgonkivs (changes requested):
Hello [Chirag018](https://github.com/Chirag018), thank you for your contribution. Thank you for your contribution. Unfortunately, it will break the logic of handling POST requests in the gateway. The condition` if r.Method == http.MethodPost && state.IsStopped(r.Context())` disables all POST requests only if the state module was stopped when a Bad Encoding Fraud Proof is received, and the network halts.

api/gateway/middleware.go
renaynay: Yeah we only have one `POST` endpoint left and it's `SubmitTx` which is safe to leave. 

let's just remove this check altogether as it's no longer necessary - if there is no state service running, it will error out regardless (it's just that the error might not be so nice for now).

renaynay (commented):
Looks good minus the comment left by @vgonkivs 

Please also lint and let's run CI against it.

Wondertan (approved):

vgonkivs:
@Chirag018 , could you please fix lint errors?

Chirag018:
> @Chirag018 , could you please fix lint errors?

yeah, will have a look on it

Chirag018:
@vgonkivs,  While addressing lint issues, I [identified](https://github.com/celestiaorg/celestia-node/actions/runs/6798218910/job/18481950971?pr=2912#step:4:32) code duplication in the state file. To resolve this, should I refactor the API.Interface [struct](https://github.com/celestiaorg/celestia-node/blob/main/nodebuilder/state/state.go#L100) or make other related changes?

cc: @Wondertan @renaynay

Wondertan:
@Chirag018, refactoring that struct should do the thing.

cc @distractedm1nd for confirmation

Chirag018:
OK, will wait for @distractedm1nd 's confirmation

distractedm1nd:
Yes, it needs to be removed from the API struct as well as the Module interface - don't forget to remove the implementations as well :)

That should do the trick

renaynay (commented):
Weird, it still has the duplicate lint issue on the state file. 

Maybe it needs a //nolint:dupl

not sure why it's coming up now and wasn't caught before.

renaynay:
@Chirag018 do you still want to finish up this PR? Just add `//nolint:dupl` to the file that's failing dupl lint

ramin (approved):

renaynay (approved):

vgonkivs (approved):

ramin:
@Chirag018 can you rebase main onto here? Or at least adjust the PR so we can update the PR with base branch, then this'll merge. Thanks!

Issue #2909, opened 2023-11-07
test(blob/service): add test case with two equal blobs


vgonkivs:
Add a test case, where multiple identical blobs are submitted under the same namespace at the same height.

PR #2914, opened 2023-11-08
test(blob): add submitting equal blobs test case

vgonkivs:
Closes https://github.com/celestiaorg/celestia-node/issues/2909


nodebuilder/tests/blob_test.go
```diff
@@ -123,6 +123,37 @@ func TestBlobModule(t *testing.T) {
...
+				height, err := fullClient.Blob.Submit(ctx, []*blob.Blob{b, b}, nil)
+				require.NoError(t, err)
+				_, err = fullClient.Header.WaitForHeight(ctx, height)
+				require.NoError(t, err)
+				b0, err := fullClient.Blob.Get(ctx, height, b.Namespace(), b.Commitment)
+				require.NoError(t, err)
+				require.Equal(t, b, b0)
+
+				// give some time to store the data,
+				// otherwise the test will hang on the IPLD level.
```
Wondertan, line 146: Issue link

nodebuilder/tests/blob_test.go
```diff
@@ -123,6 +123,37 @@ func TestBlobModule(t *testing.T) {
...
+				b, err := blob.NewBlob(
+					appBlob[0].ShareVersion,
+					append([]byte{appBlob[0].NamespaceVersion}, appBlob[0].NamespaceID...),
+					appBlob[0].Data,
+				)
+				require.NoError(t, err)
+				height, err := fullClient.Blob.Submit(ctx, []*blob.Blob{b, b}, nil)
+				require.NoError(t, err)
+				_, err = fullClient.Header.WaitForHeight(ctx, height)
+				require.NoError(t, err)
```
Wondertan, line 140: Doesn't blob.Get do it automatically?

nodebuilder/tests/blob_test.go
```diff
@@ -123,6 +123,37 @@ func TestBlobModule(t *testing.T) {
...
+				b0, err := fullClient.Blob.Get(ctx, height, b.Namespace(), b.Commitment)
+				require.NoError(t, err)
+				require.Equal(t, b, b0)
```
Wondertan, line 143: Irrc, the Equal just compares the pointers, while EqualValues would do deep comparison. 

I might be wrong, but could you verify b and b0 are actually different struct in men?

Also, let's add a case where we see that the namespace does have two equal blobs. It might deduplicate the same blobs on in a single PFB, but definitely doesn't deduplicate blobs if they are sent in different PFBs. In the test case we send both blobs in a single PFB

nodebuilder/tests/blob_test.go
distractedm1nd: #2915 
Hmmm. The deeper issue here seems to be that IPLD is hanging on the second time (if we give store long enough to store, it gets from StoreGetter)

Our ideas here are:
1. Bitswap peers maybe have an anti-spam mechanism where the same block will not be served twice to the same peer, which is fine, but:
2. These blocks should be cached on the blockstore level and not actually rerequested from the network. This makes it seem like the caching mechanism from IPLD retrieval isn't actually working.

nodebuilder/tests/blob_test.go
vgonkivs: It does. Not the `blob.Get`, the `header.Store`.

nodebuilder/tests/blob_test.go
vgonkivs:
```
// EqualValues asserts that two objects are equal or convertable to the same types
// and equal.
```
You are right, it checks the raw types

nodebuilder/tests/blob_test.go
Wondertan: I mean why do we need to wait here if Get waits in the same way?

nodebuilder/tests/blob_test.go
Wondertan: Yes and I am concerned that simple Equals succeeds here

vgonkivs: Double-checked: `Equal` also checks the underlying values in case of pointers:

```
// Equal asserts that two objects are equal.
// Pointer variable equality is determined based on the equality of the
// referenced values (as opposed to the memory addresses)
```

Wondertan: I see, thanks.

Please also address this comment 
> Also, let's add a case where we see that the namespace does have two equal blobs. It might deduplicate the same blobs on in a single PFB, but definitely doesn't deduplicate blobs if they are sent in different PFBs. In the test case we send both blobs in a single PFB

nodebuilder/tests/blob_test.go
vgonkivs: because we can get this error:
`header: given height is from the future:`

Wondertan: Ahh, I see, thank you. I forgot it works through our service

nodebuilder/tests/blob_test.go
vgonkivs: I'm a bit confused: do we expect an error on the second submit call?

vgonkivs: added one more case

nodebuilder/tests/blob_test.go
```diff
@@ -123,12 +129,87 @@ func TestBlobModule(t *testing.T) {
...
+				proof, err := fullClient.Blob.GetProof(ctx, height, b.Namespace(), b.Commitment)
+				require.NoError(t, err)
+
+				included, err := fullClient.Blob.Included(ctx, height, b.Namespace(), proof, b.Commitment)
+				require.NoError(t, err)
+				require.True(t, included)
+			},
+		},
+		{
+			name: "Submit equal blobs in different PFBs",
```
renaynay, line 168: what's the point of this test case if they're not in the same block?

vgonkivs: > Also, let's add a case where we see that the namespace does have two equal blobs. It might deduplicate the same blobs on in a single PFB, but definitely doesn't deduplicate blobs if they are sent in different PFBs. In the test case we send both blobs in a single PFB

to ensure that the second blob will be accepted if I understand @Wondertan correctly.

Wondertan: It does have to be a new test case and could be an additional assert on the existing test. The purpose is to make sure that we really have two equal blobs, because existing asserts(on the moment of original comment) could succeed even if there is only one blob.

vgonkivs: reworked

ramin (approved):

nodebuilder/tests/blob_test.go
renaynay: I'm still very confused as to what this is testing in particular. Can you add some docs or further explanation for this test case @vgonkivs ?

renaynay (approved):

nodebuilder/tests/blob_test.go
```diff
@@ -123,12 +129,73 @@ func TestBlobModule(t *testing.T) {
...
+				proof, err := fullClient.Blob.GetProof(ctx, height, b.Namespace(), b.Commitment)
+				require.NoError(t, err)
+
+				included, err := fullClient.Blob.Included(ctx, height, b.Namespace(), proof, b.Commitment)
+				require.NoError(t, err)
+				require.True(t, included)
+			},
+		},
+		{
+			name: "Submit the same blob",
```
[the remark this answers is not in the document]
vgonkivs: added explanations

PR #2937, opened 2023-11-19
Create LISEZMOI.md

solowmow:
This is the translated version of the README.md file into French.


renaynay:
@solowmow thank you for the contribution 😄 Unfortunately, we will not be maintaining any kind of documentation in languages other than English in this repository.

solowmow:
I see the value in having a documentation folder. There are many individuals interested in contributing, but they might face language barriers (I'm from Benin). Aaron Swartz might have proposed an alternative approach.

solowmow:
Nevertheless, I don't want to push for it. If you believe it's the right choice, then I'll trust your judgment.

PR #2954, opened 2023-11-23
feat(header.go): Implement backwards verification inside of `Verify`

### LINKED ISSUE #2953: header.go: Implement backwards verification inside of `Verify`
Implement backwards verification inside `Verify` such that it allows verification for an `untrusted` header of `trusted.Height()-1`

Ref: https://github.com/celestiaorg/go-header/issues/112

renaynay:
Closes #2953


Wondertan (changes requested):
We should do backward verification inside of the go-header by using Hash methods on the Header

renaynay:
can actually be part of verify inside go-header rather than implementation-side

PR #2955, opened 2023-11-23
feat(rpc): default binding to localhost vs 0.0.0.0 open to world

### LINKED ISSUE #2824: (feat): Allow binding RPC to only specific addresses
### Implementation ideas

Currently the node will accept RPC requests from any IP (modulo authentication). Add the ability for the node to restrict which IPs it will listen to for requests, allowing `localhost` only by default.

From https://geth.ethereum.org/docs/fundamentals/command-line-options#commands:

```
    --authrpc.addr value                (default: "localhost")             ($GETH_AUTHRPC_ADDR)
          Listening address for authenticated APIs

    --authrpc.port value                (default: 8551)                    ($GETH_AUTHRPC_PORT)
          Listening port for authenticated APIs

    --authrpc.vhosts value              (default: "localhost")             ($GETH_AUTHRPC_VHOSTS)
          Comma separated list of virtual hostnames from which to accept requests (server
          enforced). Accepts '*' wildcard.
```

ramin:
fixes #2824 which requests we bind to localhost by default (previously was world accessible)

- sets defaults to bind to localhost properly (previously flags SAID the defaults were localhost but were actually 0.0.0.0)
- removes an unused `DefaultConfig` defined in `api/gateway/package`
- moves defaults (bind address and port) to vars so we can use them to set in flag usage messages so they don't drift out of sync with reality
- adds tests for the defaults
- adds test cases for flag definitions
- adds test cases for flag parsing
 
Other

skips suggestion of binding vhosts, we don't do any kind of middleware checking in rpc (i believe in geth this would be a way to bind to 0.0.0.0 then limit to allowed remote IPs making requests). I'd suggest this is a more advanced configuration use case and belongs outside of the node itself to be managed by the infra operator where it'd be handled by a ingress gateway or routing rule to the machine/instance/orchestrator where the node software is running


Wondertan (approved):
LGTM

vgonkivs (approved):

zvolin:
I think this was actually a breaking change. If you run node in docker, you now need to add `--xxx.addr` flags if you want to reach out to the node from container published ports

PR #2957, opened 2023-11-24
fix(share/byzantine): fix proof collection

vgonkivs:
This PR fixes an issue in the proof collection method. Previously, we were sending `len(dah.RowRoots)/2` requests in order to get proofs but there is a use case when the process can get stuck for a particular share(out-of-order case). The new approach sends n requests, where n is the number of non-empty shares received from rsmt2d library, and then waits until `len(dah.RowRoots)/2` will be received.
I've also improved error handling as previously we were panicking in case any error appeared. The new approach will rely only on `context.DeadlineExceeded ` which basically means that for some reason we can't get proofs, so the data is not available.


share/eds/byzantine/byzantine.go
```diff
@@ -45,42 +43,50 @@ func NewErrByzantine(
...
 	sharesAmount := 0
 
-	errGr, ctx := errgroup.WithContext(ctx)
-	for index, share := range errByz.Shares {
-		// skip further shares if we already requested half of them, which is enough to recompute the row
-		// or col
-		if sharesAmount == len(dah.RowRoots)/2 {
-			break
-		}
+	type tempStruct struct {
```
Wondertan, line 46: nit: `tempStruct` is not descriptive; what about `request` as you describe it PR description or just `shareWithIdx`

share/eds/byzantine/byzantine.go
```diff
@@ -45,42 +43,50 @@ func NewErrByzantine(
...
+		share *ShareWithProof
+		index int
+		err   error
+	}
+	tempStructCh := make(chan *tempStruct)
```
Wondertan, line 51: same here

Wondertan (dismissed):
Changes here make sense. I would also add the test for the issue:
> This PR fixes an issue in the proof collection method

share/eds/byzantine/byzantine.go
```diff
@@ -45,42 +43,50 @@ func NewErrByzantine(
...
+			}
+			sharesWithProof[t.index] = t.share
+			sharesAmount++
+			if sharesAmount == len(dah.RowRoots)/2 {
+				return &ErrByzantine{
+					Index:  uint32(errByz.Index),
+					Shares: sharesWithProof,
+					Axis:   errByz.Axis,
+				}
+			}
```
walldiss, line 87: Would be a bit easier to read this part if `sharesAmount` was removed in favour of variable inside for loop initialisation. Like:
for amount:= 0; amount < len; amount++ {}

share/eds/byzantine/byzantine.go
walldiss: my 2c: Could be just `result` and `resutlCh` as it collects results from parallel func calls

share/eds/byzantine/byzantine.go
```diff
@@ -45,42 +43,50 @@ func NewErrByzantine(
...
+		case <-ctx.Done():
+			return ipld.ErrNodeNotFound
```
walldiss, line 89: Why ipld.ErrNodeNotFound? Is there any handling targeting this error up the stack?
Would rather have something more explicit here(could be wrapped error too). Like "wan't able to collect enough shares. found %v, required: %v"

walldiss (commented):
Is it possible for `n`(number of non-empty shares received from rsmt2d library) to be less then len(dah.RowRoots)/2? I guess in this case proof collection won't be able to collect enough samples. So we should short circuit with error.

share/eds/byzantine/byzantine.go
vgonkivs: reworked

Wondertan (approved):
Good test

walldiss (approved):

PR #2962, opened 2023-11-27
refactor(modp2p): decouple Bitswap

Wondertan:
Decouples Bitswap into Client and Server. This is necessary to provide new custom options to the Client. In particular, this brings `client.WithoutDuplicatedBlockStats()` option to the table as nice little optimization for LNs. It disables one metrics we don't care much about atm, but avoids expensive Has call to DB as profiles show.

### DIFF
--- a/nodebuilder/p2p/bitswap.go
+++ b/nodebuilder/p2p/bitswap.go
@@ -30,15 +30,14 @@ const (
 
 // dataExchange provides a constructor for IPFS block's DataExchange over BitSwap.
 func dataExchange(params bitSwapParams) exchange.Interface {
-	prefix := protocol.ID(fmt.Sprintf("/celestia/%s", params.Net))
-
+	prefix := protocolID(params.Net)
 	net := network.NewFromIpfsHost(params.Host, &routinghelpers.Null{}, network.Prefix(prefix))
 	srvr := server.New(
 		params.Ctx,
 		net,
 		params.Bs,
 		server.ProvideEnabled(false), // we don't provide blocks over DHT
-		// NOTE: These below ar required for our protocol to work reliably.
+		// NOTE: These below are required for our protocol to work reliably.
 		// // See https://github.com/celestiaorg/celestia-node/issues/732
 		server.SetSendDontHaves(false),
 	)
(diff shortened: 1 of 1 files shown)


nodebuilder/p2p/bitswap.go
```diff
@@ -29,16 +31,38 @@ const (
 // dataExchange provides a constructor for IPFS block's DataExchange over BitSwap.
 func dataExchange(params bitSwapParams) exchange.Interface {
 	prefix := protocol.ID(fmt.Sprintf("/celestia/%s", params.Net))
```
ramin, line 33: it'd be nice to have this as a func on bitSwapParams so you could spit out `params.ProtococID()` and we could write a little unit test for it, maybe even add `params.NeworkPrefix()` and we can test both and then just pass that as 3rd arg to `network.NewFromIpfsHost`

nodebuilder/p2p/bitswap.go
```diff
@@ -29,16 +31,38 @@ const (
...
+
+	net := network.NewFromIpfsHost(params.Host, &routinghelpers.Null{}, network.Prefix(prefix))
+	srvr := server.New(
 		params.Ctx,
-		network.NewFromIpfsHost(params.Host, &routinghelpers.Null{}, network.Prefix(prefix)),
+		net,
 		params.Bs,
-		bitswap.ProvideEnabled(false),
+		server.ProvideEnabled(false), // we don't provide blocks over DHT
 		// NOTE: These below ar required for our protocol to work reliably.
```
ramin, line 41: nit: typo `ar` -> `are`

nodebuilder/p2p/bitswap.go
```diff
@@ -29,16 +31,38 @@ const (
...
-		// See https://github.com/celestiaorg/celestia-node/issues/732
```
ramin, line 38: great history lesson for me 😸

ramin (dismissed):

nodebuilder/p2p/bitswap.go
Wondertan: I see your point. Not sure why that should be a method on bitSwapParams, though we could also extract that in a func and similarly unit test it. I don't think that the params struct needs any methods

Wondertan: Extract in a func

ramin: doing too many years of OOP is why i have the impulse to put it ON something, but func works exactly the same 👍

walldiss (approved):

vgonkivs (approved):

PR #2837, opened 2023-10-11
fix(modp2p): temporary disable quic

Wondertan:
Disables QUIC by default yet allows to enable it with ENV var. 
This is temporary and until https://github.com/libp2p/go-libp2p/issues/2591 is investigated and fixed.

The current solution disables QUIC programmatically while we still keep writing in the config QUIC listen addresses by default. This removes the need for users to reinit their configs once we turn QUIC back by default.

This works perfectly fine on mocha.

PR #2963, opened 2023-11-27
Revert "fix(modp2p): temporary disable quic (#2837)"

Wondertan:
Reverts disabled quic. We weren't able to reproduce [the issue](https://github.com/libp2p/go-libp2p/issues/2591) since we first observed it, as well as quic released changes to address this. Even if the issue is there, I think it's more or less safe as we always have TCP as fallback


renaynay (approved):
ok

vgonkivs (approved):

Wondertan:
I thought this was merged already 😱

PR #2967, opened 2023-11-27
feat(discovery): add jitter and fix backoff

nodersteam:
- Replacing time.NewTicker with time.NewTimer
- Adding backoff logic
- Adding jitter to backoff
- Removed unused test variable


nodersteam:
Is it acceptable to use math/rand or add generation logic via crypto/rand to fix Lint?

nodersteam:
@Wondertan Could you please check this PR?
Or is it possible to point out more important issues at the moment if this PR is not of value at the moment?

nodersteam:
> Hey @nodersteam, I am deferring the review of this to @walldiss, as the most prominent expert on that code.

Got it! Thx

share/p2p/discovery/discovery.go
```diff
@@ -4,16 +4,18 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math/rand"
 	"time"
 
+	"golang.org/x/sync/errgroup"
```
cristaloleg, line 10: Better to keep as it was.

share/p2p/discovery/discovery_test.go
```diff
@@ -26,9 +26,7 @@ const (
 func TestDiscovery(t *testing.T) {
 	const nodes = 10 // higher number brings higher coverage
 
-	discoveryRetryTimeout = time.Millisecond * 100 // defined in discovery.go
-
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second*20)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute*10)
```
cristaloleg, line 29: Why so big? 30x more than before 👀

share/p2p/discovery/discovery.go
```diff
@@ -199,21 +203,25 @@ func (d *Discovery) Advertise(ctx context.Context) {
...
-				// rerun discovery if the number of peers hasn't reached the limit
+				d.backoffDuration += backoffInitialDuration
+				if d.backoffDuration > maxBackoffDuration {
+					d.backoffDuration = maxBackoffDuration
+				}
+				jitter := time.Duration(float64(d.backoffDuration) * jitterFactor * (rand.Float64()*2 - 1))
+				backoffTimer.Reset(d.backoffDuration + jitter)
 				continue
 			}
+			d.backoffDuration = backoffInitialDuration
```
cristaloleg, line 224: Should we also reset `backoffTimer` here?

PR #2968, opened 2023-11-27
fix(modp2p): increase pubsub seencache

### LINKED ISSUE #2934: modp2p: increase the SeenMessage interval in PubSub
We periodically receive old headers from the network that were already synced. Increasing the seen message interval should decrease those.  

The only downside is that it may eat slightly more RAM, but it will improve GossipSub peering behaviors. Instead of punishing outdated messages as invalid, it will only reduce communication with the peer. This is especially important when the new CHOKE msg is introduced in the GossipSub v1.2

Wondertan:
Attempts to fix #2934 

Needs testing.


Wondertan:
Closing this as stale. Also needs some proof that it has any impacted before being merged

PR #2972, opened 2023-11-28
pruner: Implement skeleton for pruner package

renaynay:
This PR implements a skeleton for the pruner package and integrates the pruner's AvailabilityWindow determination into the DASer sampling routine.

Namings are up for discussion.


libs/utils/fs.go
```diff
@@ -1,6 +1,8 @@
 package utils
 
-import "os"
+import (
```
renaynay, line 3: needed for update to goimports-reviser

pruner/pruner.go
```diff
@@ -0,0 +1,41 @@
...
+	factory Factory
+}
+
+func NewPruner(factory Factory) *Pruner {
+	return &Pruner{
+		factory: factory,
+	}
+}
+
+func (p *Pruner) Start(context.Context) error {
```
renaynay, line 29: would be good to expose some method on factory / or a config to enable/disable the pruning routine in case it's not needed (for archival, for example)

ramin (commented):
i preferred when you named this a carcass, but skeleton makes sense :-)

Really my only genuine comments are that i'm stuck on naming, i wonder if the package should be called storage or something? Thinking out loud on naming, but maybe it's:

- storage (ie: concerns related to storage of blocks)
- storage/pruner = where pruner implements the various 'workers' implementations that performs pruning for diff node types
- storage/availability (which then implements a range of Window functions or Sampling strategies for node types)

Thinking then it conceptually reads like 

Daser, is injected with an availabilityWindowSampler, which checks sample/storage conditions based on the header coming in, and then storage has pruner package "pruning" which implements various strategies for pruning data which is adjacent to windowSampler.

Thinking this is not dissimilar to OTEL and trace sampling etc

is there a better name than storage as top level if we went with that?

so then `pruning.AvailabilityWindow` -> `availability.WindowSampler`, and maybe has `sampler.ShouldSample(h Header)` true etc?

all suggestions. also, reading about pruning so much makes me think of baby food.

pruner/pruner.go
renaynay: Like it immediately quits out? I don't think it's necessary. I'd rather have some sort of way to short-circuit the start + stop if there's no "active prune factory" found. We also usually like to have start and stop conform to the lifecycle iface

nodebuilder/prune/module.go
```diff
@@ -0,0 +1,49 @@
...
+				return p.Stop(ctx)
+			}),
+		)),
+		fx.Provide(
+			func(factory pruner.Factory) pruner.AvailabilityWindow {
+				return factory
+			}),
+	)
+
+	switch tp {
```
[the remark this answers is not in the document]
renaynay: The problem is the node type doesn't always correlate to one specific factory. Like a full node could have either archival or prune factory, for example. 

We also don't really pass down the concept of "node types" any further than the nodebuilder package, so those modules should not really know about `node.Type` imo.

we could make a nice helper function optionally in nodebuilder that could do that `newFromNodeType` thing you want.

das/daser_test.go
```diff
@@ -87,7 +87,7 @@ func TestDASer_Restart(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	t.Cleanup(cancel)
 
-	daser, err := NewDASer(avail, sub, mockGet, ds, mockService, newBroadcastMock(1))
+	daser, err := NewDASer(avail, sub, mockGet, ds, mockService, newBroadcastMock(1), &mockPruner{})
```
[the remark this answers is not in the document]
renaynay: We'd have to adopt that pattern for most constructors then but I think that would be more code than necessary. 

if args get out of control, we can consider it I guess.

pruner/pruner.go
```diff
@@ -0,0 +1,41 @@
...
+	"time"
+
+	"github.com/celestiaorg/celestia-node/header"
+)
+
+// AvailabilityWindow defines an interface for pruning historical data.
+type AvailabilityWindow interface {
+	// IsWithinAvailabilityWindow returns true if the given header is within the
+	// "availability" (sampling) window of the network.
+	IsWithinAvailabilityWindow(eh *header.ExtendedHeader, window time.Duration) bool
```
walldiss, line 14: Looks like there is no need to pass window var since it is constant. It can be part of the implementation and accessed locally.

pruner/window.go
```diff
@@ -0,0 +1,7 @@
+package pruner
+
+import (
+	"time"
+)
+
+const Window = 30 * 86400 * time.Second // 30 days worth of seconds
```
walldiss, line 7: Should be local. It will be available through exported interface anyway

pruner/factory.go
```diff
@@ -0,0 +1,14 @@
...
+
+	"github.com/celestiaorg/celestia-node/header"
+)
+
+// Factory contains methods necessary to prune data
+// from the node's datastore.
+type Factory interface {
+	AvailabilityWindow
+	Prune(context.Context, ...*header.ExtendedHeader) error
+}
```
walldiss, line 14: Could you please provide some details why those needs to be combined? Is it because AvailabilityWindow will have different implementations for LN and FN? II thought they have same recency windowm, so there is no need to make it part of same interface and separate implementation to same thing.

pruner/pruner.go
renaynay: I think I did this before for testing purposes (if we wanted to reduce the window / change the window somehow) but i'm happy to remove it and just make it a var if needed.

pruner/factory.go
renaynay: The reason why is so we can have an Archival implementation that returns true for every height.

pruner/window.go
renaynay: My only concern is about where it should live as it would be used by the `full` and `light` implementations - so it needs to live somewhere as a default / param

pruner/factory.go
walldiss: I think we can split interfaces, since they are not related to each other:
```
AvailabilityWindow interface - archival / pruned implementations
Prune - light / full implementations
```

pruner/window.go
walldiss: If interface is splitted, it could become part of `archival` / `pruned` implementation only, while not affecting `full` / `light` store prunning implementations

pruner/light/prune.go
```diff
@@ -0,0 +1,22 @@
+package light
+
+import (
+	"context"
+
+	"github.com/celestiaorg/celestia-node/header"
+)
+
+type Pruner struct{}
```
renaynay, line 9: rename to ______?________ + archival too

pruner/pruner.go
```diff
@@ -0,0 +1,40 @@
...
+	IsWithinAvailabilityWindow(eh *header.ExtendedHeader) bool
+}
+
+// Pruner handles the pruning routine for the node using the
+// prune Factory.
+type Pruner struct {
+	factory Factory
+}
+
+func NewPruner(factory Factory) *Pruner {
```
renaynay, line 22: Manager / Service

pruner/pruner.go
```diff
@@ -0,0 +1,40 @@
...
+// AvailabilityWindow defines an interface for pruning historical data.
+type AvailabilityWindow interface {
+	// IsWithinAvailabilityWindow returns true if the given header is within the
+	// "availability" (sampling) window of the network.
+	IsWithinAvailabilityWindow(eh *header.ExtendedHeader) bool
+}
+
+// Pruner handles the pruning routine for the node using the
+// prune Factory.
+type Pruner struct {
```
renaynay, line 18: Manager / Service

pruner/factory.go
```diff
@@ -0,0 +1,14 @@
...
+import (
+	"context"
+
+	"github.com/celestiaorg/celestia-node/header"
+)
+
+// Factory contains methods necessary to prune data
+// from the node's datastore.
+type Factory interface {
+	AvailabilityWindow
```
renaynay, line 12: need to think about where to actually return this / whether to have this in Pruner iface

pruner/pruner.go
renaynay, line 18: add a Config in this pkg

ramin (approved):
kinda love this tbh

Wondertan (approved):

pruner/light/window.go
```diff
@@ -0,0 +1,9 @@
+package light
+
+import (
+	"time"
+
+	"github.com/celestiaorg/celestia-node/pruner"
+)
+
+const Window = pruner.AvailabilityWindow(time.Second * 86400 * 30)
```
vgonkivs, line 9: non-blocking, but it would be nice to add a comment about how much time in days

vgonkivs (approved):
Looks great 🚀

pruner/light/window.go
renaynay: will add in follow up pr

PR #2975, opened 2023-11-29
chore(modheader): lower store cache size for LN

Wondertan:
Part of an effort to decrease the memory footprint for LNs. Headers are pretty beefy.

Extracted from #2960

### DIFF
--- a/nodebuilder/header/config.go
+++ b/nodebuilder/header/config.go
@@ -48,7 +48,6 @@ func DefaultConfig(tp node.Type) Config {
 
 	switch tp {
 	case node.Bridge:
-		cfg.Client = p2p_exchange.ClientParameters{}
 		return cfg
 	case node.Full:
 		return cfg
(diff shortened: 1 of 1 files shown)


nodebuilder/header/config.go
```diff
@@ -43,13 +43,19 @@ func DefaultConfig(tp node.Type) Config {
 		Store:        store.DefaultParameters(),
 		Syncer:       sync.DefaultParameters(),
 		Server:       p2p_exchange.DefaultServerParameters(),
+		Client:       p2p_exchange.DefaultClientParameters(),
 	}
 
 	switch tp {
 	case node.Bridge:
+		cfg.Client = p2p_exchange.ClientParameters{}
```
renaynay, line 51: why is this necessary?

Wondertan: I just dumbly followed the behavior that was before, where Bridge had empty Client params. But now I think it is fine to remove it.

Wondertan: Ah wait. It may actually impact the `celestia init`. Please check on the BN if the client config is initialized there.

Wondertan: @renaynay

renaynay:
```
	// we do not create a client for bridge nodes
	if tp == node.Bridge {
		return nil
	}
```

so no

renaynay (approved):

vgonkivs (approved):

PR #2976, opened 2023-11-29
chore(moddas): speed up sampling for LN

Wondertan:
It felt like this sped up sampling for my resyncs, but I didn't record any of the actual time/data to prove that, so needs more testing and keeping it as a draft for now.


renaynay (commented):
should we consider this even with sampling window coming up soon ?

renaynay:
@walldiss 

Please check this default and provide results

Wondertan:
Will revisit this once shwap comes

PR #2981, opened 2023-11-30
feat(pruner/light): Implement light sampling window

renaynay:
Implements a sampling window for light nodes of 30 days' worth of seconds. 

Based on #2972

### DIFF
--- a/pruner/light/prune_test.go
+++ b/pruner/light/prune_test.go
@@ -1,0 +1,35 @@
+package light
+
+import (
+	"strconv"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/celestiaorg/celestia-node/header/headertest"
+)
+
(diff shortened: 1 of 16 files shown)


pruner/light/prune_test.go
```diff
@@ -0,0 +1,35 @@
...
+	}
+
+	p := NewPruner()
+
+	for i, tt := range tests {
+		t.Run(strconv.Itoa(i), func(t *testing.T) {
+			eh := headertest.RandExtendedHeader(t)
+			eh.RawHeader.Time = eh.RawHeader.Time.Add(-tt.toSubtract)
+
+			assert.Equal(t, tt.withinWindow, p.IsWithinAvailabilityWindow(eh, time.Hour*24*30))
```
renaynay, line 32: can't import the window const from top-level pruner pkg so maybe redefine the const here?

PR #2989, opened 2023-12-04
create store v2 file interface

walldiss:
This is the foundation of new ~era~ sampling protocol and storage.


share/store/file.go
```diff
@@ -0,0 +1,21 @@
...
+	"github.com/celestiaorg/nmt"
+	"github.com/celestiaorg/rsmt2d"
+
+	"github.com/celestiaorg/celestia-node/share"
+)
+
+type File interface {
+	io.Closer
+	Size() int
+	ShareWithProof(ctx context.Context, axisType rsmt2d.Axis, axisIdx, shrIdx int) (share.Share, nmt.Proof, error)
```
Wondertan, line 16: Suggestion:
Let's just call it `Share`. The WithProof part would be needed only if there is a version to get a share without a proof.

share/store/file.go
```diff
@@ -0,0 +1,21 @@
...
+	Axis(ctx context.Context, axisType rsmt2d.Axis, axisIdx int) ([]share.Share, error)
```
Wondertan, line 17: Do we need it? The sampling protocol does not.

share/store/file.go
```diff
@@ -0,0 +1,21 @@
...
+	"context"
+	"io"
+
+	"github.com/celestiaorg/nmt"
+	"github.com/celestiaorg/rsmt2d"
+
+	"github.com/celestiaorg/celestia-node/share"
+)
+
+type File interface {
```
Wondertan, line 13: Please add comments for the File. In general, treat everything we do in the protocol_v2 branch as we do it in main

share/store/file.go
walldiss: Removed for now. Could be added later if needed

PR #2992, opened 2023-12-06
feat(store/mem_file): add in-memory eds file implementation

walldiss:
Adds in-memory file implementation

### DIFF
--- a/share/getter.go
+++ b/share/getter.go
@@ -73,15 +73,15 @@ func (ns NamespacedShares) Verify(root *Root, namespace Namespace) error {
 
 	for i, row := range ns {
 		// verify row data against row hash from original root
-		if !row.verify(originalRoots[i], namespace) {
+		if !row.Verify(originalRoots[i], namespace) {
 			return fmt.Errorf("row verification failed: row %d doesn't match original root: %s", i, root.String())
 		}
 	}
 	return nil
 }
 
-// verify validates the row using nmt inclusion proof.
-func (row *NamespacedRow) verify(rowRoot []byte, namespace Namespace) bool {
+// Verify validates the row using nmt inclusion proof.
+func (row *NamespacedRow) Verify(rowRoot []byte, namespace Namespace) bool {
 	// construct nmt leaves from shares by prepending namespace
 	leaves := make([][]byte, 0, len(row.Shares))
 	for _, shr := range row.Shares {
--- a/share/getter.go
+++ b/share/getter.go
@@ -72,6 +72,9 @@ func (ns NamespacedShares) Verify(root *Root, namespace Namespace) error {
 	}
 
 	for i, row := range ns {
+		if row.Proof == nil && row.Shares == nil {
+			return fmt.Errorf("row verification failed: no proofs and shares")
+		}
 		// verify row data against row hash from original root
 		if !row.Verify(originalRoots[i], namespace) {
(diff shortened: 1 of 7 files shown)


share/getter.go
```diff
@@ -73,15 +73,15 @@ func (ns NamespacedShares) Verify(root *Root, namespace Namespace) error {
 
 	for i, row := range ns {
 		// verify row data against row hash from original root
-		if !row.verify(originalRoots[i], namespace) {
+		if !row.Verify(originalRoots[i], namespace) {
```
Wondertan, line 76: A minor change from local changes. It makes sure we don't segfault when proof is malformed.
```suggestion
		if row.Proof == nil && row.Shares == nil {
			return fmt.Errorf("row verification failed: no proofs and shares")
		}
		if !row.Verify(originalRoots[i], namespace) {
```

share/store/mem_file.go
```diff
@@ -0,0 +1,101 @@
...
+	err = batchAdder.Commit()
+	if err != nil {
+		return share.NamespacedRow{}, err
+	}
+
+	cid := ipld.MustCidFromNamespacedSha256(root)
+	row, proof, err := ipld.GetSharesByNamespace(context.TODO(), bserv, cid, namespace, len(shares))
+	if err != nil {
+		return share.NamespacedRow{}, err
+	}
```
Wondertan, line 81: We might want to dedup this later

share/store/mem_file.go
```diff
@@ -0,0 +1,101 @@
...
+func (f *MemFile) axis(axisType rsmt2d.Axis, axisIdx int) []share.Share {
+	switch axisType {
+	case rsmt2d.Row:
+		return f.Eds.Row(uint(axisIdx))
+	case rsmt2d.Col:
+		return f.Eds.Col(uint(axisIdx))
+	default:
+		panic("unknown axis")
+	}
+}
```
Wondertan, line 101: I want this to be an rsmt2d method so bad

Wondertan: (no actionable here)

share/store/mem_file_test.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/celestiaorg/celestia-app/pkg/wrapper"
+	"github.com/celestiaorg/rsmt2d"
+
+	"github.com/celestiaorg/celestia-node/share"
+	"github.com/celestiaorg/celestia-node/share/eds/edstest"
+	"github.com/celestiaorg/celestia-node/share/sharetest"
+)
+
+func TestMemFileShare(t *testing.T) {
+	eds := edstest.RandEDS(t, 32)
```
Wondertan, line 22: kind big for the test but doesn't really matter

share/store/mem_file_test.go
```diff
@@ -0,0 +1,90 @@
...
+	if to < from {
+		from, to = to, from
+	}
+	expected := shares[from]
+	namespace := share.GetNamespace(expected)
+
+	// change namespace for all shares in range
+	for i := from; i <= to; i++ {
+		shares[i] = expected
+	}
```
Wondertan, line 73: I would leave a todo to make this a helper. I have tests where I need something similar

share/store/mem_file.go
walldiss: Later

share/store/mem_file_test.go
walldiss: I've extracted it to test pkg in last commit for you 🙂

PR #2998, opened 2023-12-11
chore(utils): test sanitizing empty addr

estensen:
Add coverage to the non-happy path

### DIFF
--- a/libs/utils/address_test.go
+++ b/libs/utils/address_test.go
@@ -9,8 +9,9 @@ import (
 
 func TestSanitizeAddr(t *testing.T) {
 	var tests = []struct {
-		addr string
-		want string
+		addr   string
+		want   string
+		expErr bool
 	}{
 		// Testcase: trims protocol prefix
 		{addr: "http://celestia.org", want: "celestia.org"},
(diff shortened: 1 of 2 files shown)


libs/utils/address_test.go
```diff
@@ -9,8 +9,9 @@ import (
 
 func TestSanitizeAddr(t *testing.T) {
 	var tests = []struct {
-		addr string
-		want string
+		addr   string
+		want   string
+		expErr bool
```
ramin, line 14: lets just name this `err`

ramin (changes requested):
this is cool, lets go a step further, and create a named var in the address.go file (`ErrInvalidIP` maybe) then instead of the expErr being a bool, we can make it an error and give it nil or the expected error and test `require.EqualErorr` or similar as part of the test, then we can skip the !tt.expErr and just test that its nil or the `ErrInvalidIP`

libs/utils/address_test.go
```diff
@@ -20,13 +22,20 @@ func TestSanitizeAddr(t *testing.T) {
...
 		t.Run(tt.addr, func(t *testing.T) {
 			got, err := SanitizeAddr(tt.addr)
-			require.NoError(t, err)
 			require.Equal(t, tt.want, got)
+			if tt.err != nil {
+				require.Error(t, err)
+				require.True(t, errors.Is(err, tt.err))
+			} else {
+				require.NoError(t, err)
+			}
```
estensen, line 38: Wanted this
```go
require.Equal(t, tt.want, got)
require.ErrorIs(t, tt.err, err)
```

But couldn't get it to work
```go
Error:      	Target error should be in err chain:
            	            	expected: "invalid IP address or hostname given: "
            	            	in chain: "invalid IP address or hostname given"
            	Test:       	TestSanitizeAddr/#00
```

ramin (approved):
lgtm :shipit:

libs/utils/address_test.go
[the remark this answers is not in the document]
ramin: @renaynay created a replacement with the change here to speed us along https://github.com/celestiaorg/celestia-node/pull/3051

PR #2999, opened 2023-12-11
chore(utils): test square function

estensen:
Improve test cov

### DIFF
--- a/libs/utils/square_test.go
+++ b/libs/utils/square_test.go
@@ -1,0 +1,59 @@
+package utils
+
+import (
+	"testing"
+
+	"github.com/ethereum/go-ethereum/common/math"
+)
+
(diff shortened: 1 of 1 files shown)


ramin (approved):
seems good, @vgonkivs what do you think?

libs/utils/square_test.go
```diff
@@ -0,0 +1,59 @@
+package utils
+
+import (
+	"testing"
+
+	"github.com/ethereum/go-ethereum/common/math"
```
vgonkivs, line 6: what is the point here to use go-eth lib?

estensen: sorry, I think it go auto imported somehow. Fixed now

ramin (approved):

PR #3004, opened 2023-12-13
chore(p2p/flags): Handle empty p2p.network flag.

### LINKED ISSUE #3003: [cmd/flag]: validation is skipping empty flag value
### Celestia Node version

v0.12.1

### OS

linux

### Install tools

_No response_

### Others

_No response_

### Steps to reproduce it

```bash
celestia light start --p2p.network= --node.store=/home/celestia/.celestia-light-mocha-4 --core.ip=full.consensus.mocha-4.celestia-mocha.com --gateway --da.grpc.namespace=000008e5f679bf7116cb --da.grpc.listen=0.0.0.0:26650
```

### Expected result

start fails with error that `p2p.network` is failed due to empty value provided

### Actual result

celestia mainnet is started

### Relevant log output

_No response_

### Notes

_No response_

00x-dx:
Fixes #3003.


README.md
```diff
@@ -70,11 +70,11 @@ More information can be found [here](https://github.com/celestiaorg/celestia-nod
 `<node_type>` can be: `bridge`, `full` or `light`.
 
 ```sh
-celestia <node_type> init
+celestia <node_type> init --p2p.network=mainnet
```
Wondertan, line 73: Running without p2p.network should still work, as it has the default value

Wondertan (changes requested):

README.md
00x-dx: @Wondertan Update the code. We would be defaulting to DefaultNetwork; which is in-deed mainnet.

nodebuilder/p2p/flags.go
```diff
@@ -32,10 +32,11 @@ Peers must bidirectionally point to each other. (Format: multiformats.io/multiad
 	)
 	flags.String(
 		networkFlag,
-		"",
+		"default",
```
Wondertan, line 35: This should be `celestia` to match chain-id, which is DefaultNetwork

nodebuilder/p2p/flags.go
```diff
@@ -68,12 +69,18 @@ func ParseFlags(
 // and returns either the parsed network or the build's default network
 func ParseNetwork(cmd *cobra.Command) (Network, error) {
 	parsed := cmd.Flag(networkFlag).Value.String()
+	if parsed == "default" {
```
Wondertan, line 72: Compare against the string behind `DefaultNetwork`

00x-dx: You have rightly pointed out. Compare against would be better here.

nodebuilder/p2p/flags.go
```diff
@@ -32,10 +32,11 @@ Peers must bidirectionally point to each other. (Format: multiformats.io/multiad
 	)
 	flags.String(
 		networkFlag,
-		"",
+		"celestia",
```
Wondertan, line 35: One last suggestion. Let's also make it `DefaultNetwork.String()`

nodebuilder/p2p/flags.go
```diff
@@ -68,12 +69,18 @@ func ParseFlags(
...
 		// no network found in env, so use the default network
 		if envNetwork == "" {
-			return DefaultNetwork, err
+			if err != nil {
+				return "", err
+			}
+			return "", fmt.Errorf("no network provided, allowed values: %s", listProvidedNetworks())
 		}
 		return envNetwork, err
 	}
```
Wondertan, line 86: I just realized this expects an empty flag and only triggers the Env Var parsing. With this change, env var will be ignored, which is a bug. For this fix to work, I think we need to prioritize Env Var, and only if it's not defined to look into flags, which needs a slight change in the logic here. We will also need to add t the flag description that the flag is overridden by the env var

I think this is what I meant in a previous comm 
> We need to be careful, though, not to break behaviors

Wondertan (changes requested):

nodebuilder/p2p/flags.go
00x-dx: I think right priority could be 

- Command line arguments
- Environment variables
- Defaults

With this, we would fallback onto defaults; and the args would always override the env vars. WDYT ?

Wondertan: I think you are right about this priority. The tricky part is not to treat the default coming from the flag as the command line argument overriding the env var. But that's doable

Wondertan: If we make `if flag == default => return env_var.network`, then there will be a catch, and the following example won't follow the priority, as the flag won't override the env var.
`CELESTIA_CUSTOM=arabica celestia light start --p2p.network celestia`

00x-dx: @Wondertan I have re-structured the network parsing .. And it make more sense to override with custom network. Lets look at some of the examples:

> celestia full|light start|init
`celestia`

> celestia full|light start|init --p2p.network=""
`error`

> celestia full|light start|init --p2p.network="mocha"
`mocha`

> CELESTIA_CUSTOM="celestia" celestia full|light start|init --p2p.network="mocha"
`celestia`

> CELESTIA_CUSTOM="" celestia full|light start|init --p2p.network=""
`error`

Tests should cover these cases.

00x-dx:
> @00x-dx this is nearly there, do you have time to finish this today? i'll test and we'll get it merged if so

Sure, i have pushed the additional changes. Let me know how the testing goes through ..

nodebuilder/p2p/flags.go
Wondertan: In general, flags should be prioritised over end vars, but I agree that in our case it's fine to make envs a priority, as long as we document that properly on the command.

distractedm1nd (approved):
Thank you!

ramin (approved):

Issue #1781, opened 2023-02-20
feat(share/p2p/peer-manager): optimise peer manager memory footprint


walldiss:
### Implementation ideas

Peer manager collects peers for different headers. It will slowly grow in size over lifespan of a node, so proper resource management technics addressing its memory size needs to be implemented. 

 - One of the possible optimisations that could address this issue could be limiting amount of stored peers inside the peer pools. At some point having more peers stored in the pool very unlikely to give any benefit. 
  - Another one could be limit the amount of pools stored inside peer manager. This could be done by limiting amount of stored pools approach (lru), or limited pool lifespan approach. Pool validation logic could blacklist removed pool and then recreated pools. In order to eliminate such false-positive trigger, there needs to be introduced check if header with given datahash exist in header store, when pool validation times out.

PR #3005, opened 2023-12-13
feat(share/p2p/peer-manager): limit amount of stored pools in peer-manager

walldiss:
This PR introduces `amountOfStoredPools` param to Peer manager, that will limit amount of stored pools. Defautlt value is 10 pools. Peer manager will try to keep `amountOfStoredPools` amount of pools only for recent headers. Older blocks don't need to be routed by manual pools and could be routed to any full node, as those are expected to have the block by that time. 

This will lower memory footprint of peer manager as well as resolve memory leaking issues. 

Resolves https://github.com/celestiaorg/celestia-node/issues/1781

### DIFF
--- a/share/p2p/peers/manager.go
+++ b/share/p2p/peers/manager.go
@@ -37,6 +37,9 @@ const (
 	// eventbusBufSize is the size of the buffered channel to handle
 	// events in libp2p
 	eventbusBufSize = 32
+
+	// amountOfStoredPools is the amount of pools for recent headers that will be stored in the peer manager
+	amountOfStoredPools = 10
 )
 
 type result string
@@ -303,7 +306,7 @@ func (m *Manager) subscribeHeader(ctx context.Context, headerSub libhead.Subscri
 
 		// update storeFrom if header height is higher than current value
 		old := m.storeFrom.Load()
-		upd := int64(h.Height()) - m.params.StoreWindow
+		upd := int64(h.Height()) - amountOfStoredPools
 		if old < upd {
 			if m.storeFrom.Swap(upd) != old {
 				log.Debugw("updated lowest stored height", "height", upd)
--- a/share/p2p/peers/manager.go
+++ b/share/p2p/peers/manager.go
@@ -27,8 +27,6 @@ import (
 const (
 	// ResultNoop indicates operation was successful and no extra action is required
 	ResultNoop result = "result_noop"
-	// ResultSynced will save the status of pool as "synced" and will remove peers from it
-	ResultSynced = "result_synced"
 	// ResultCooldownPeer will put returned peer on cooldown, meaning it won't be available by Peer
 	// method for some time
 	ResultCooldownPeer = "result_cooldown_peer"
@@ -56,11 +54,12 @@ type Manager struct {
 	host      host.Host
 	connGater *conngater.BasicConnectionGater
 
-	// pools collecting peers from shrexSub
+	// pools collecting peers from shrexSub and stores them by datahash
 	pools map[string]*syncPool
-	// messages from shrex.Sub with height below initialHeight will be ignored, since we don't need to
+	// messages from shrex.Sub with height below storeFrom will be ignored, since we don't need to
 	// track peers for those headers
 	initialHeight atomic.Uint64
+	storeFrom     atomic.Uint64
 
 	// fullNodes collects full nodes peer.ID found via discovery
 	fullNodes *pool
@@ -82,14 +81,12 @@ type DoneFunc func(result)
 type syncPool struct {
 	*pool
 
+	dataHash string
 	// isValidatedDataHash indicates if datahash was validated by receiving corresponding extended
 	// header from headerSub
 	isValidatedDataHash atomic.Bool
 	// headerHeight is the height of header corresponding to syncpool
-	headerHeight atomic.Uint64
-	// isSynced will be true if DoneFunc was called with ResultSynced. It indicates that given datahash
-	// was synced and peer-manager no longer need to keep peers for it
-	isSynced atomic.Bool
+	height uint64
 	// createdAt is the syncPool creation time
 	createdAt time.Time
 }
@@ -190,16 +187,15 @@ func (m *Manager) Stop(ctx context.Context) error {
 // full nodes, it will wait until any peer appear in either source or timeout happen.
 // After fetching data using given peer, caller is required to call returned DoneFunc using
 // appropriate result value
-func (m *Manager) Peer(
-	ctx context.Context, datahash share.DataHash,
+func (m *Manager) Peer(ctx context.Context, datahash share.DataHash, height uint64,
 ) (peer.ID, DoneFunc, error) {
-	p := m.validatedPool(datahash.String())
+	p := m.validatedPool(datahash.String(), height)
 
 	// first, check if a peer is available for the given datahash
 	peerID, ok := p.tryGet()
 	if ok {
 		if m.removeIfUnreachable(p, peerID) {
-			return m.Peer(ctx, datahash)
+			return m.Peer(ctx, datahash, height)
 		}
 		return m.newPeer(ctx, datahash, peerID, sourceShrexSub, p.len(), 0)
 	}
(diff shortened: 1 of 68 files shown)


share/p2p/peers/options.go
```diff
@@ -56,6 +59,7 @@ func DefaultParameters() Parameters {
 		// sync time for large blocks. This value gives our (discovery) peers enough time to sync
 		// the new block before we ask them again.
 		PeerCooldown: 3 * time.Second,
+		StoreWindow:  50,
```
Wondertan, line 62: Why 50? My intuition tells me that 10 would be enough because 10 heights is enough time for the network to sync, and everyone beyond 10 can simply go to FNs.

Wondertan, line 62: Also, why do you think this needs to be configured by the node operator?

Wondertan (commented):
Initial glimpse. Did you run the node to see stable mem usage? I can help with that

share/p2p/peers/options.go
walldiss: Just a safety measure in case block time becomes lower like 1s. I didn't want to make pool manager pool window to be time based as it is more difficult to control, but it could also could be a solution

walldiss: I don't think node operator needs to, but I think it is nice to have it exposed for testing

walldiss: Ok. if 10 will work with fast blocks I'll lower it

walldiss: Also might worth removing it just so it doesn't break config. Sucks that adding new fields to any subcomponent configs or changes of any default values requires re-init of the node.

share/p2p/peers/manager.go
```diff
@@ -39,6 +37,9 @@ const (
 	// eventbusBufSize is the size of the buffered channel to handle
 	// events in libp2p
 	eventbusBufSize = 32
+
+	// amountOfStoredPools is the amount of pools for recent headers that will be stored in the peer manager
+	amountOfStoredPools = 10
```
walldiss, line 42: I'm open for better name for this var

share/p2p/peers/options.go
walldiss: removed from config, don't wanna break

share/p2p/peers/manager.go
```diff
@@ -82,14 +86,12 @@ type DoneFunc func(result)
...
+	dataHash string
 	// isValidatedDataHash indicates if datahash was validated by receiving corresponding extended
 	// header from headerSub
 	isValidatedDataHash atomic.Bool
 	// headerHeight is the height of header corresponding to syncpool
-	headerHeight atomic.Uint64
-	// isSynced will be true if DoneFunc was called with ResultSynced. It indicates that given datahash
-	// was synced and peer-manager no longer need to keep peers for it
-	isSynced atomic.Bool
+	height uint64
```
Wondertan, line 94: comment also needs to change

share/p2p/peers/manager.go
```diff
@@ -82,14 +86,12 @@ type DoneFunc func(result)
 type syncPool struct {
 	*pool
 
+	dataHash string
```
Wondertan, line 89: doc nit: comment wouldn't hurt as with others fields

share/p2p/peers/manager.go
```diff
@@ -27,8 +27,6 @@ import (
 const (
 	// ResultNoop indicates operation was successful and no extra action is required
 	ResultNoop result = "result_noop"
-	// ResultSynced will save the status of pool as "synced" and will remove peers from it
-	ResultSynced = "result_synced"
```
Wondertan, line 31: Do I understand correctly that we don't need it because we now clean up things quickly, and there is no point in keeping synced pools?

share/p2p/peers/manager.go
Wondertan: Not much to suggest. Maybe storedPoolsAmount

share/p2p/peers/manager.go
```diff
@@ -298,12 +297,21 @@ func (m *Manager) subscribeHeader(ctx context.Context, headerSub libhead.Subscri
...
 
 		// store first header for validation purposes
 		if m.initialHeight.CompareAndSwap(0, h.Height()) {
 			log.Debugw("stored initial height", "height", h.Height())
 		}
+
+		// update storeFrom if header height is higher than current value
+		old := m.storeFrom.Load()
+		upd := int64(h.Height()) - amountOfStoredPools
+		if old < upd {
```
Wondertan, line 310: I don't think this is necessary. Subscription guarantees that new headers are always higher than previous

Wondertan: As well that debug log

share/p2p/peers/manager.go
```diff
@@ -360,17 +368,14 @@ func (m *Manager) Validate(_ context.Context, peerID peer.ID, msg shrexsub.Notif
 		return pubsub.ValidationReject
 	}
 
-	if msg.Height < m.initialHeight.Load() {
-		// we can use peers from discovery for headers before the first one from headerSub
-		// if we allow pool creation for those headers, there is chance the pool will not be validated in
-		// time and will be false-positively trigger blacklisting of hash and all peers that sent msgs for
-		// that hash
+	fmt.Println("incoming", msg.Height, m.storeFrom.Load())
```
Wondertan, line 371: debug log to remove

share/p2p/peers/manager.go
```diff
@@ -432,8 +445,9 @@ func (m *Manager) isBlacklistedHash(hash share.DataHash) bool {
 	return m.blacklistedHashes[hash.String()]
 }
 
-func (m *Manager) validatedPool(hashStr string) *syncPool {
-	p := m.getOrCreatePool(hashStr)
+func (m *Manager) validatedPool(hashStr string, height uint64) *syncPool {
+	fmt.Println("validate", height)
```
Wondertan, line 449: another one

share/p2p/peers/manager.go
```diff
@@ -473,6 +487,7 @@ func (m *Manager) GC(ctx context.Context) {
 
 func (m *Manager) cleanUp() []peer.ID {
 	if m.initialHeight.Load() == 0 {
+		fmt.Println("1")
```
Wondertan, line 490: hmmm

share/p2p/peers/manager.go
```diff
@@ -482,12 +497,24 @@ func (m *Manager) cleanUp() []peer.ID {
...
+		}
+
+		// can't validate datahashes below initial height
+		if p.height < m.initialHeight.Load() {
+			delete(m.pools, h)
+			continue
+		}
+
+		// find pools that are not validated in time
+		if time.Since(p.createdAt) > m.params.PoolValidationTimeout {
```
Wondertan, line 515: I think the validation timeout should be defined in heights now:
* Removes the need for createdAt (time operations are costly)
* In general, it is easier to reason about if we make `storedPoolsAmount` with heights.

We should also decrease it for LNs. I don't see a point of keeping pools for 2mins for LNs

share/p2p/peers/manager.go
walldiss: Yeah, tracking synced / unsynced pools not worth the complexity anymore, as there won't be as many pools now

share/p2p/peers/manager.go
Wondertan: Decided in sync to do defer this, s.t. we can release the fix asap.

walldiss: As we discussed privately. Can't rely on incoming height, as it could be set by sender to any value higher than know network head. It creates situation, when it is diffucult to distinguish if msg recieved was for invalid datahash, or headerSub is lagging behind shrexSub.
There could be simplier approach, that would require some refactoring, so I'll land it into separate issue.

walldiss:
Yes, memory leak is solved.

Wondertan (approved):

vgonkivs (approved):

share/p2p/peers/manager.go
walldiss: https://github.com/celestiaorg/celestia-node/issues/3011

PR #3007, opened 2023-12-14
chore(utils): test exists function

estensen:


libs/utils/fs_test.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestExists(t *testing.T) {
+	// Create a temporary file for testing
+	tmpFile, err := os.CreateTemp("", "test")
```
ramin, line 12: i think i'd prefer if we use t.TempDir() here please https://pkg.go.dev/testing#T.TempDir

ramin (changes requested):

ramin (approved):

estensen:
@ramin why was it approved and then closed?

PR #3012, opened 2023-12-17
feat(share/store): Allow Store implementation to choose proof axis for Share

walldiss:
Reworks `Share` method of `Store` to return shareWithProof containing axis field.

### DIFF
--- a/share/share.go
+++ b/share/share.go
@@ -54,8 +54,8 @@ type ShareWithProof struct { //nolint: revive
 }
 
 // Validate validates inclusion of the share under the given root CID.
-func (s *ShareWithProof) Validate(rootHash []byte, shrIdx, axisIdx, edsSize int) bool {
-	isParity := shrIdx >= edsSize/2 || axisIdx >= edsSize/2
+func (s *ShareWithProof) Validate(rootHash []byte, x, y, edsSize int) bool {
+	isParity := x >= edsSize/2 || y >= edsSize/2
 	namespace := ParitySharesNamespace
 	if !isParity {
 		namespace = GetNamespace(s.Share)
--- a/share/share.go
+++ b/share/share.go
@@ -2,10 +2,13 @@ package share
 
 import (
 	"bytes"
+	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
 
 	"github.com/celestiaorg/celestia-app/pkg/appconsts"
+	"github.com/celestiaorg/nmt"
+	"github.com/celestiaorg/rsmt2d"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


share/share.go
```diff
@@ -40,6 +43,31 @@ func GetData(s Share) []byte {
 	return s[NamespaceSize:]
 }
 
+// ShareWithProof contains data with corresponding Merkle Proof
+type ShareWithProof struct { //nolint: revive
```
Wondertan, line 47: Let's name it sample

share/store/eds_file.go
```diff
@@ -15,11 +14,23 @@ type EdsFile interface {
 	// Size returns square size of the file.
 	Size() int
 	// Share returns share and corresponding proof for the given axis and share index in this axis.
-	Share(ctx context.Context, axisType rsmt2d.Axis, axisIdx, shrIdx int) (share.Share, nmt.Proof, error)
+	Share(ctx context.Context, x, y int, proofType ProofAxis) (*share.ShareWithProof, error)
```
Wondertan, line 17: As we discussed, the file/storage itself should be responsible for the decision of which proof to return, not the user.

Wondertan: Imagine if only Col is available but we request the Row here.

Wondertan (changes requested):

share/store/eds_file.go
walldiss: >Imagine if only Col is available but we request the Row here.
Exactly for case, when specific proof is required this option is created. To allow caller to request explicit axis proofs if needed (for example for befp building).`But user is not forced to, and if he wants to allow implementation to choose there is `proofAxisAny` option.

For sampling protocol you would need to use only `proofAxisAny` option.

share/share.go
walldiss: ShareWithProof:
- It is not always sample. The type is also used within BEFP pkg, where `sample` name will be confusing and might be used in other implementation in future. It should be named protocol agnostic. 
- it literally extends `Share` type with Proof, so named according to the content inside

share/store/eds_file.go
Wondertan: I think befps will have to be generated inside of the file and not outside. If not befp, there is no other usecase needing to request specific proof.

walldiss: befps are happening on sampling (protocol level), during eds reconstruction. So should be generated on protocol level too, using api provided by store. There is no need to put it inside File reading / writing logic

walldiss: I can remove no problem if you are sure if befp will be inside Store

Wondertan: I got to think about this more and I don't think befps will anyhow be part of the file/store, nor anyone will need to ask the store for the proofs in case of befps.

Wondertan: Befp will be formed in runtime when we fetch a square. That bad square might be stored afterwards but bepf will be already formed and stored separately. And when it's formed we would only use pieces coming from the network.

Wondertan: > So generation logic should happen on protocol layer too, utilising api provided by Store

So I don't see how the protocol would need to access the file for the proof when protocol already has all the proofs in memory or fetches they from the network.

walldiss: Removed the option to select axis of proof.  Befp is a bit too far, so will change later if we need to.

share/share.go
```diff
@@ -40,6 +43,31 @@ func GetData(s Share) []byte {
...
+
+// Validate validates inclusion of the share under the given root CID.
+func (s *ShareWithProof) Validate(rootHash []byte, shrIdx, axisIdx, edsSize int) bool {
+	isParity := shrIdx >= edsSize/2 || axisIdx >= edsSize/2
+	namespace := ParitySharesNamespace
+	if !isParity {
+		namespace = GetNamespace(s.Share)
+	}
+	return s.Proof.VerifyInclusion(
+		sha256.New(), // TODO(@Wondertan): This should be defined somewhere globally
```
distractedm1nd, line 64: Don't know how old this comment is but some hasher has caused us issues in the past when we tried this, because it couldn't be shared across goroutines. I would assume its the same case here since the hasher can be written to from any goroutine at any time, so I would be careful with this

share/share.go
```diff
@@ -40,6 +43,31 @@ func GetData(s Share) []byte {
...
+	// Share is a full data including namespace
+	Share
+	// Proof is a Merkle Proof of current share
+	Proof *nmt.Proof
+	// Axis is a type of axis against which the share proof is computed
+	Axis rsmt2d.Axis
+}
+
+// Validate validates inclusion of the share under the given root CID.
+func (s *ShareWithProof) Validate(rootHash []byte, shrIdx, axisIdx, edsSize int) bool {
```
distractedm1nd, line 57: Should these maybe be x and y too instead of shrIdx and axisIdx?

Wondertan (approved):

PR #3013, opened 2023-12-17
refactor(share): Move shareWithProof to share pkg

walldiss:
Extracts shareWithProof definition and methods from byzantine pkg to the share pkg.


share/eds/byzantine/bad_encoding.go
```diff
@@ -208,15 +212,15 @@ func (p *BadEncodingProof) Validate(hdr *header.ExtendedHeader) error {
 		log.Debugw("failed to decode shares at height",
 			"height", hdr.Height(), "err", err,
 		)
-		return nil
+		return fmt.Errorf("failed to decode shares: %w", err)
```
walldiss, line 215: Previously error was not returned in this case and following ones. Fixed in advance, but I might be missing the idea of how those are handled by function caller.

share/eds/byzantine/bad_encoding.go
```diff
@@ -190,13 +196,11 @@ func (p *BadEncodingProof) Validate(hdr *header.ExtendedHeader) error {
...
 		// validate inclusion of the share into one of the DAHeader roots
-		if ok := shr.Validate(ipld.MustCidFromNamespacedSha256(merkleRoots[index])); !ok {
+		if ok := shr.Validate(merkleRoots[index], index, int(p.Index), int(odsWidth)*2); !ok {
 			log.Debugf("%s: %s at index %d", invalidProofPrefix, errIncorrectShare, index)
 			return errIncorrectShare
 		}
-		// NMTree commits the additional namespace while rsmt2d does not know about, so we trim it
-		// this is ugliness from NMTWrapper that we have to embrace ¯\_(ツ)_/¯
-		shares[index] = share.GetData(shr.Share)
+		shares[index] = shr.Share
```
walldiss, line 203: I'm so happy about cleaning up this part

share/eds/byzantine/byzantine.go
```diff
@@ -60,12 +65,13 @@ func NewErrByzantine(
 
 		index := index
 		errGr.Go(func() error {
-			share, err := getProofsAt(
+			sh, err := ipld.GetShareWithProof(
 				ctx, bGetter,
-				ipld.MustCidFromNamespacedSha256(roots[index]),
```
walldiss, line 65: removed double conversion of []byte -> cid.Cid - > []byte

share/eds/byzantine/bad_encoding.go
Wondertan: So, the fact that error exist proves the correctness of  BEFP, so its indeed Valid and no error should be returned here. You may add a little comment explaining that here.

share/share.go
```diff
@@ -61,7 +61,7 @@ func (s *ShareWithProof) Validate(rootHash []byte, x, y, edsSize int) bool {
 		namespace = GetNamespace(s.Share)
 	}
 	return s.Proof.VerifyInclusion(
-		sha256.New(), // TODO(@Wondertan): This should be defined somewhere globally
+		sha256.New(),
```
Wondertan, line 64: Its still is though

share/eds/byzantine/byzantine.go
```diff
@@ -41,8 +42,12 @@ func NewErrByzantine(
 		dah.ColumnRoots,
 		dah.RowRoots,
 	}[errByz.Axis]
+	axisType := rsmt2d.Row
+	if errByz.Axis == rsmt2d.Row {
+		axisType = rsmt2d.Col
+	}
```
Wondertan, line 48: This seems wrong. Why? To make test work after breaking Validate error?

Wondertan (changes requested):
https://github.com/celestiaorg/celestia-node/pull/3013/files#r1514610441 is the blocking comment

share/eds/byzantine/byzantine.go
walldiss: There is a comment out of the scope of Github (2 lines above) specifying that we need to collect proofs against ortogonal axis. Anyway, refactored it a bit more to get rid of [][][]byte. Looks cleaner now

share/share.go
walldiss: This happens in many places, so worth to have an issue instead of TODO

Wondertan (approved):

Issue #2470, opened 2023-07-13
feat(monitoring): define custom buckets per histogram


walldiss:
### Implementation ideas

All histogram metrics currently are using same predefined buckets values, that makes most of them unusable due to buckets being disproportional to observed values. With current version of otel it is only possible to set buckets values on global meter, but given different nature of each histogram it makes it unusable. 

This is known issue of otel and tracked in:
https://github.com/open-telemetry/opentelemetry-go/issues/3826. 
API spec improvement has been suggested and merged in:
https://github.com/open-telemetry/opentelemetry-specification/pull/3216
Implimentation is still to be done. And tracked in:
https://github.com/open-telemetry/opentelemetry-go/issues/4094

Once support of bucket boundary metric advice support is implemented in otel we need to add custom bucket per each histogram.

walldiss:
Since otel bump in https://github.com/celestiaorg/celestia-node/pull/3068 custom buckets are now possible. We need to revise our histogram definitions and set up sensible buckets values.

PR #3068, opened 2024-01-03
chore(deps): update otel group 

ramin:
reduced myself to fixing dependabot's attempts to upgrade us at https://github.com/celestiaorg/celestia-node/pull/3055 was not working quite right


walldiss:
I think it didn't initially add 
```
github.com/pyroscope-io/otel-profiling-go
```
Because it is not in actual otel group, but rather an pyroscope intagration, that lives in pyroscope repo. Does otel group depend on pyroscope pkg? Do we need it in same group?

walldiss (approved):

walldiss:
This bump unblocks this one
https://github.com/celestiaorg/celestia-node/issues/2470
It brings update to otel, that allows long waited specified bundaries for histogram metrics.

vgonkivs (approved):

PR #3072, opened 2024-01-08
chore(nodebuilder/state): SubmitTx endpoint takes read-level permissions

### LINKED ISSUE #2958: nodebuilder/state: Make `SubmitTx` endpoint a `read`-level permission
It's not necessary to have `SubmitTx` behind `write`-level permissions as the transaction that would be submitted through that endpoint would already have to be signed. Thus, we can have it as `read` level only.

moldis:
refactor: changes SubmitTX permissions from write to read.

Following the issue requirement, this PR changes SubmitTX permissions from _write_ to _read_

Changes:

* Updated nodebuilder/state/state.go from **write** to **read** permissions
* Updated following tests (api/rpc_test.go) and added extra check for SubmitTX rpc call (read permissions check).

closes #2958


api/rpc_test.go
```diff
@@ -177,21 +177,35 @@ func TestAuthedRPC(t *testing.T) {
 				stats, err := rpcClient.DAS.SamplingStats(ctx)
 				require.NoError(t, err)
 				require.Equal(t, expected, stats)
+
+				expectedResp := &state.TxResponse{}
+				server.State.EXPECT().SubmitTx(gomock.Any(),
+					gomock.Any()).Return(expectedResp, nil)
+				txResp, err := rpcClient.State.SubmitTx(ctx, []byte{})
+				require.NoError(t, err)
+				require.Equal(t, expectedResp, txResp)
```
renaynay, line 186: Not necessary as we have an example of a read-level permissions endpoint :) But thank you

api/rpc_test.go
```diff
@@ -177,21 +177,35 @@ func TestAuthedRPC(t *testing.T) {
...
 			} else {
 				_, err := rpcClient.DAS.SamplingStats(ctx)
 				require.Error(t, err)
 				require.ErrorContains(t, err, "missing permission")
+
+				_, err = rpcClient.State.SubmitTx(ctx, []byte{})
+				require.Error(t, err)
+				require.ErrorContains(t, err, "missing permission")
```
renaynay, line 194: here too, see above :)

api/rpc_test.go
moldis: Resolved

api/rpc_test.go
moldis: Resolved!

vgonkivs (approved):

walldiss (approved):

Partykingdk:
ok

renaynay (approved):
Looks good, thank you!

PR #2619, opened 2023-08-27
chore: bump go-header and go-fraud

Wondertan:
This PR bumps go-header and go-fraud which are both breaking - lots of changes here so please review carefully.

Note that this also contains re-introduction of VerifyCommitLightTrusting which caused us issues in `mocha` due to rapid voting power changes. At the moment, there are no test utilities to test that specific case, so we need to wait to roll it out to `mocha`. 

Related: https://github.com/celestiaorg/celestia-app/issues/2382

Issue #3071, opened 2024-01-07
header/p2p: Mainnet nodes getting `mocha-4` headers


mindstyle85:
### Celestia Node version

12.1

### OS

ubuntu

### Steps to reproduce it

install and run light node like in the docs (defaults to mainnet without flags etc)

### Expected result

run on mainnet

### Actual result

2024-01-07T09:04:19.472-0800        ERROR        header/sync        sync/sync_head.go:175        invalid network header        {"height_of_invalid": 894619, "hash_of_invalid": "9E53A68FAC3C4051046A498E58C1C15E59CF306AAE8CC80D459A83EDF23E3BB4", "height_of_subjective": 503228, "hash_of_subjective": "8AD1E4F9C65A163312F344FDBDD17B71444D0EA42A12240F7401634AA45C6DD9", "reason": "wrong chain id: 'mocha-4' != 'celestia'"}

mindstyle85:
note from the person observing this (and he wasnt the only one reporting it, we just figured they misconfigured): it seems that the error above happens only on that specific header

mindstyle85:
some more logs from other reports:


`sometimes i get "2023-12-29T17:43:26.465Z    ERROR    header/sync    sync/sync_head.go:175    invalid network header    {"height_of_invalid": 829037, "hash_of_invalid": "6F0A2391EB372F91482085B4D73B5B0579F5811A2B06855D6839AE5D0DA539C5", "height_of_subjective": 436976, "hash_of_subjective": "2606F33CC027A92A242ABFD4E2957C978449E3777B2EDFF45E4ED97BE653540F", "reason": "wrong chain id: 'mocha-4' != 'celestia'"}"`
`
2024-01-05T22:49:07.051+0100    ERROR    header/sync    sync/sync_head.go:175    invalid network header    {"height_of_invalid": 881445, "hash_of_invalid": "729B0D560C699F39EA7F7D42E1662B0A955F1709859E59CEB4B688AD7B67D611", "height_of_subjective": 489952, "hash_of_subjective": "2F1779EC9384A2D370935AA6CBC8C720B5304AE24C17C49F56B0C391DF998285", "reason": "wrong chain id: 'mocha-4' != 'celestia'"}`

[...]

PR #3076, opened 2024-01-08
fix(header): Return error from `Validate()` after `MakeExtendedHeader`

renaynay:
This PR re-introduces checking the error from `eh.Validate()` after constructing the ExtendedHeader. I'm not sure why it was removed in https://github.com/celestiaorg/celestia-node/pull/2619/files#diff-c7b74c4c56b9e973cda0ea769f215f89232c4b3a66d3cfdc5c42867b364b2745L98 but I think it makes sense to leave it as a basic sanity check as there doesn't seem to be a clear reason for why it was removed.

This change was found when debugging an issue with headers being propagated from `mocha-4` into mainnet.

### DIFF
--- a/header/header.go
+++ b/header/header.go
@@ -67,7 +67,7 @@ func MakeExtendedHeader(
 		Commit:       comm,
 		ValidatorSet: vals,
 	}
-	return eh, nil
+	return eh, eh.Validate()
 }
 
 func (eh *ExtendedHeader) New() *ExtendedHeader {
--- a/libs/utils/resetctx.go
+++ b/libs/utils/resetctx.go
@@ -1,6 +1,8 @@
 package utils
 
-import "context"
+import (
+	"context"
+)
 
 // ResetContextOnError returns a fresh context if the given context has an error.
 func ResetContextOnError(ctx context.Context) context.Context {


Wondertan:
The reason was to avoid expensive signature validation. Does it fix the issue?

walldiss:
If we leave it without validtation, invalid header might be transmitted by broadcaster. It will make many other nodes to waste CPU resources to run validation on such headers to only dismiss them.

Even if we trust source of header, there still possibility for integration errors / software bugs. Sanity check will help us to find it early, before retransmitting junk messages.

renaynay:
@walldiss bugs with header formation will be caught at a different stage on the local node, just not at the Broadcast level. It will be caught with the same validation logic used to validate external messages. We shouldn't duplicate if it's not absolutely necessary.

I agree with @Wondertan that the sanity check isn't worth the additional compute since the validity check still occurs in the software.

walldiss:
Yeah I agree, that we cannot prevent some modified nodes to send anything and thus have receiver validation in place.

But in case of https://github.com/celestiaorg/celestia-node/issues/3071, it is our software that sends those messages and we  are in control to prevent it from doing it. Also I suspect node runner is unaware of the problem with his node and think it operates fine, while it is not. If we enable broadcaster validation, there will be several benefits:
- no junk in messages in network from this particular case
- Bridge Node runner will detect problem early by seeing lots of ERROR logs and will fix configuration

renaynay:
@walldiss adding a basic validation check would not have found this error case by the way - this is only caught via `Verify` inside of go-header. 

To point #2, BN runner will see problem regardless bc it will catch the issue in the syncer. I agree it wouldn't be intuitive to the node runner though that the errors are local.

walldiss:
Would it resolve the issue if we do some lightweight sanity check instead of validate (like `Verify`) before sending the message?

renaynay:
@walldiss `Verify` is meant to check an unknown/untrusted header against one that is known and accepted as part of the chain. Validate is just a basic sanity check (although expensive). We could potentially separate out `VerifyCommitLight` and make a smaller sanity check subset func of Validate if you really think it's necessary.

PR #3077, opened 2024-01-08
fix(core): Use `Verify` from go-header pkg in `CoreExchange`

renaynay:
We should have been using the high-level `Verify` check from go-header inside of the `CoreExchange` instead of the `ExtendedHeader`-based implementation of Verify.


distractedm1nd (approved):
😬

ramin (approved):

walldiss (approved):

PR #3085, opened 2024-01-10
fix(nodebuilder/blob): allow blob submit args to be pre-encoded

### LINKED ISSUE #3082: Unable to upload pre-encoded (base64/hex) blobs
### Celestia Node version

v0.12.1

### OS

N/A

### Install tools

_No response_

### Others

_No response_

### Steps to reproduce it

[The docs](https://docs.celestia.org/developers/node-tutorial#submitting-data) describe that:

> Data can be hex-encoded (`0x...`), base64-encoded (`"..."`), or a plaintext string which will be encoded to base64 (`'Hello There!'`). 

Using pre-encoded hex and base64 does not work in this manner.

### Expected result

I would expect hex strings and base64 strings are converted to bytes from their standard representation, instead of the literal ascii values of their already encoded strings.

### Actual result

In actuality, all data is treated as text, and the ascii representations of the arg are converted to the ascii byte values.

### Relevant log output

_No response_

### Notes

https://github.com/celestiaorg/celestia-node/blob/73336eda96510652881080c7b339d0eca7bc4ca1/nodebuilder/blob/cmd/blob.go#L151 here's where the argument is converted into a blob, it doesn't do any check to see if it is already encoded as hex or base64.

Drewsapple:
Closes #3082 

When using the blobData arg from the `celestia blob submit` command, it now checks if the argument is a valid hex string or base64 string. If so, the canonical byte representation of these encodings is used for the blob. Otherwise, the previous behavior (encoding the ascii characters as bytes) is used.

### DIFF
--- a/nodebuilder/blob/cmd/blob.go
+++ b/nodebuilder/blob/cmd/blob.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"encoding/base64"
+	"encoding/hex"
 	"fmt"
 	"reflect"
 	"strconv"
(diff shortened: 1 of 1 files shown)


Drewsapple:
Superseded by #3088

PR #3087, opened 2024-01-10
feat(cmd/cel-shed): support display private key discreetly

### LINKED ISSUE #321: cmd: `cel-shed new key` should result in a file by default rather on stdout for security reasons

Halimao:
Closes #321 

This pr added support for printing private key on an alternate screen when running `cel-shed p2p new-key`

----

## Screenshots

image

image


ramin:
@Halimao this is cool, i think the original ticket requests to NOT output to stdout at all, vs an alternate screen, with that it may be we do not merge this. I'll let @Wondertan clarify the requirement and see if he wants to proceed with this solution, vs the original specified

Halimao:
> @Halimao this is cool, i think the original ticket requests to NOT output to stdout at all, vs an alternate screen, with that it may be we do not merge this. I'll let @Wondertan clarify the requirement and see if he wants to proceed with this solution, vs the original specified

OK. And if we still need to write it to file, I think it's fine that we merge this pr first? As it's more safe to write it to an alternate screen than current terminal directly. In that way, I will draft another PR to support writing to file😄

Halimao:
@ramin @Wondertan Hi sir, any new progress here?

cristaloleg:
#321 mentions that we should not print key to stdout but to a file. This PR still prints it to the terminal and not to a file.

I think writing to a file as it's stated in issues is better and doesn't require additional dependency for literally 1 use case. My 2c.

PR #3094, opened 2024-01-11
refactor(blob)!: replace SubmitOptions with a GasPrice parameter

### LINKED ISSUE #3053: [Feature Request]: Allow clients to configure Gas Price in SubmitOptions
### Implementation ideas

Currently, clients are able to configure `Fee` and `Gas` as part of `SubmitOptions`. As `Gas` depends on blob size, it requires the client to calculate it for each blob and adjust `Fee` accordingly. The client only cares about the Gas Price so that the transaction can be prioritised.

Hence this is a feature request that the client should be able to configure a Gas Price _once_ that can be used to calculate and adjust fee for all subsequent submitted blobs.

distractedm1nd:
This PR modifies the blob interface to replace SubmitOptions with GasLimit. The reasoning here is that setting a gas price is a lot more intuitive/easier than calculating the fee and gas limit, which take cosmossdk and celestia-app imports to calculate without much effort. Related: https://github.com/rollkit/go-da/issues/30

Based on a suggestion from @vgonkivs we alias float64 to be able to provide a default value, but now that its there I think we can just have float64 and the same constructor, as it doesn't change that much.

Breaks the API, cc @jcstein, @tuxcanfly, and @Ferret-san .

Closes #3053.


vgonkivs (commented):
What has happened with the order of `gomock` import in `/mocks/`?

blob/service.go
```diff
@@ -66,15 +80,21 @@ func DefaultSubmitOptions() *SubmitOptions {
...
-		options = DefaultSubmitOptions()
+	options := DefaultSubmitOptions()
+	if gasPrice > 0 {
+		blobSizes := make([]uint32, len(blobs))
+		for i, blob := range blobs {
+			blobSizes[i] = uint32(len(blob.Data))
+		}
+		options.GasLimit = blobtypes.EstimateGas(blobSizes, appconsts.DefaultGasPerBlobByte, auth.DefaultTxSizeCostPerByte)
+		options.Fee = types.NewInt(int64(math.Ceil(float64(gasPrice) * float64(options.GasLimit)))).Int64()
 	}
```
walldiss, line 98: Should gasPrice equal to 0 also be handled here? Otherwise, 0 falls under the default (-1) case.

vgonkivs: Nice catch, yes. 0 is a valid gas price

distractedm1nd: Very good catch, thank you

blob/service.go
```diff
@@ -66,15 +80,21 @@ func DefaultSubmitOptions() *SubmitOptions {
...
-	if options == nil {
-		options = DefaultSubmitOptions()
+	options := DefaultSubmitOptions()
+	if gasPrice >= 0 {
+		blobSizes := make([]uint32, len(blobs))
+		for i, blob := range blobs {
+			blobSizes[i] = uint32(len(blob.Data))
+		}
+		options.GasLimit = blobtypes.EstimateGas(blobSizes, appconsts.DefaultGasPerBlobByte, auth.DefaultTxSizeCostPerByte)
+		options.Fee = types.NewInt(int64(math.Ceil(float64(gasPrice) * float64(options.GasLimit)))).Int64()
```
renaynay, line 97: Is there a reason this logic needs to live in blob module rather than deeper down in state? Like state module can have a concept of GasPrice

What I'm concerned about mainly is that `Submit` won't be used by many due to wanting to create and sign the tx remotely rather than via the node's account.

distractedm1nd: Submit should have the easiest ux and SubmitPayForBlob should have full optionality, which is why this "optionality" is calculated here from the easy ux option of gasPrice

renaynay: The point is, we should have GasPrice at StateModule level too so users can use it more comfortable (like even for basic transfers). We can make this into a separate issue.

renaynay (dismissed):

nodebuilder/blob/cmd/blob.go
```diff
@@ -37,23 +36,13 @@ func init() {
...
-		&fee,
-		"fee",
-		-1,
-		"specifies fee (in utia) for blob submission.\n"+
+	submitCmd.PersistentFlags().Float64Var(
+		&gasPrice,
+		"gas.price",
+		float64(blob.DefaultGasPrice()),
+		"specifies gas price (in utia) for blob submission.\n"+
 			"Fee will be automatically calculated if negative value is passed [optional]",
```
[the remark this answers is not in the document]
distractedm1nd: Yes, our default is -1 which defaults to the app const. I will document this and clarify in the docs

nodebuilder/blob/cmd/blob.go
```diff
@@ -135,7 +124,7 @@ var submitCmd = &cobra.Command{
 	Short: "Submit the blob at the given namespace.\n" +
 		"Note:\n" +
 		"* only one blob is allowed to submit through the RPC.\n" +
-		"* fee and gas.limit params will be calculated automatically if they are not provided as arguments",
+		"* fee and gas limit params will be calculated automatically based on the gas.price parameter.\n",
```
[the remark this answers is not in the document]
distractedm1nd: Yeah, this phrasing sucks. Thanks

blob/service.go
distractedm1nd: Error: error: No such logger

Wondertan (approved):
utACK

renaynay (approved):
also utack

musalbas:
If we're removing the ability to set gas limit, we need to be 100% the gas limit estimate is correct, as I have heard reports from teams that the estimated gas limit is lower than it actually should be. This is why wallets like Keplr multiply gas limit estimates 1.2x

jcstein:
has this been tested?

jcstein:
Screenshot 2024-02-08 at 10 57 11 AM
however, i'm not sure if this is accurate?

jcstein:
How does this apply to the remainder of methods? context in this issue https://github.com/celestiaorg/celestia-node/issues

Issue #2737, opened 2023-09-20
[Feature Request]: allow subscribing to new blocks


SuperFluffy:
### Implementation ideas

A nice feature would be to subscribe to new blocks under a target namespace, maybe something like `blob.Subscribe` that takes 1 or more namespaces. It would be great to receive the current height and a map of `namespace -> [blob]` (the RPC should return an empty map and/or list if no blocks were written).

At the moment we can use `header.Subscribe` and then extract the most recent `height`. We can then follow this up with `blob.GetAll` using `height` and a single `[MyNamespace]` as parameters.

This is fine and not too inconvient, but a direct subscription to blocks that avoids the extra RPC would be a good feature for a data availability layer.

Wondertan:
Thanks for the issue! We had something like this in mind for a while but never received any feature requests from any of teams. We should def do blob.Subsribe

PR #3098, opened 2024-01-12
feat: allowing subscribe to new blocks with namespace

moldis:
Closes https://github.com/celestiaorg/celestia-node/issues/2737

This PR allows to subscribe to blob by multiple namespaces.

- Added new function Subscribe, which contains list of Namespaces.
- Added unit-tests to test function above.
- Updated RPC methods with function above.


distractedm1nd:
Thank you for taking this on!

I don't think blob Subscribe should return a map of height -> map of namespace -> blob list. This would mean sending all the blobs for all previous heights every block. 

I think it would make more sense to return something like a 

```golang
type BlobsubResponse struct {
          height uint64
          blobsByNamespace BlobsByNamespace
}


Subscribe func(context.Context, []share.Namespace) (<-chan BlobsubResponse, error) `perm:"read"`
```

I don't like the blobsubresponse name yet but the structure should probably be the same. Thoughts @renaynay, @walldiss ?

moldis:
> Thank you for taking this on!
> 
> I don't think blob Subscribe should return a map of height -> map of namespace -> blob list. This would mean sending all the blobs for all previous heights every block.
> 
> I think it would make more sense to return something like a
> 
> ```go
> type BlobsubResponse struct {
>           height uint64
>           blobsByNamespace BlobsByNamespace
> }
> 
> 
> Subscribe func(context.Context, []share.Namespace) (<-chan BlobsubResponse, error) `perm:"read"`
> ```
> 
> I don't like the blobsubresponse name yet but the structure should probably be the same. Thoughts @renaynay, @walldiss ?

Thanks, good idea. Now looks too overloaded with map, will change it to this structure

blob/service.go
```diff
@@ -186,6 +205,49 @@ func (s *Service) Included(
...
+				wg := sync.WaitGroup{}
+				for i, namespace := range namespaces {
+					wg.Add(1)
+					go func(i int, namespace share.Namespace) {
+						defer wg.Done()
+						blobs, err := s.getBlobs(ctx, namespace, head)
+						if err != nil {
+							log.Debugw("error getting blobs", "namespace", namespace.String(), "height", head.Height())
+							return
+						}
```
walldiss, line 233: In case of an error, the returned value will lack some data, making it impossible for the subscriber to identify whether there were errors or data missed in this scenario.

One way to handle this situation could be to shut down the subscription and return the error on the first occurrence. While this is a common approach, it would mean that the subscriber, upon encountering the first error, would need to resubscribe and could potentially miss some critical blocks during the process. I would suggest handling these errors differently.

I agree with the point @distractedm1nd raised about returning a single struct that contains both the blob and height. Perhaps the returned struct should also include information about errors. This would allow the subscription to continue even in the presence of errors, shifting the error-handling logic to the subscriber. The subscriber would still have the ability to shut down the subscription if the handling logic deems it necessary.

moldis: @walldiss I added extra channel for errors, including height, namespace and error itself. Since subscriber will decide himself to unsubscribe channel or not.

distractedm1nd:
Closing in favor of #3539

PR #3101, opened 2024-01-12
fix(share/eds): remove bitswap cache wrap from eds store

walldiss:
Two layers of caching were implemented for bitswap requests in full nodes using EDS store:

A bitswap cache that caches individual shares.
An internal EDS store cache that caches accessors.
In the event of a cache miss, both the internal cache of the EDS store and the bitswap cache will be populated with the same share, introducing redundancy in caching. Moreover, having a per-share cache doesn't seem to make much sense for full nodes, given the astronomically small chances of requesting the same share from a block that is not cached in the EDS cache.

The pull request removes the per-share cache as it brings close to no value while consuming RAM. Your understanding and consideration of this change are greatly appreciated.


Wondertan:
Iirc, it doesnt cache the full blocks there but there presence for Has method via bloomfilters. It is an optimization for bitswap to avoid doing Hazes(reads). This might still be helpful, but needs more thoughts.

walldiss:
This need more thought, as the cache we use here for full node is not even bloomfilter, but twoqueue. It caches if block exists and also when it does not exists. So there might be a problem with false positive cached `dont have`:
1. Get request `#1` for non stored block
2. Cached `don't have` for cid
3. Block obtained
4. Get request `#2` for same block hits cache and results in `don't have` response 

Same problem applies to bloomfilter cache, that is used in light nodes.

walldiss:
This is not a problem for Light nodes, as blockstore cache will be updated on successful sampling.

vgonkivs:
I think we can close this @walldiss

PR #3102, opened 2024-01-12
chore(nodebuilder, share): few typo fix

vehorny:
Fixed few typos I came across while studying the code. Hope I can be of any (little) help.
I should have squashed it all to one commit.
Doesn't require testing or changelog entry.
Have a nice evening.


ramin:
thanks @vehorny, we generally aren't inclined to accept small typo and trivial fixes on their own

PR #3113, opened 2024-01-18
feat(blob/trace): cover blob service with traces

### LINKED ISSUE #3112: metrics(blob): Add trace paths through blob module to trace Read paths
There have been some loose reports of "reads being slow". Though this is vague, a first step is to add traces to the blob module read codepath. 

Then we can use a testing tool (such as waypoint) to perform many reads and see what traces are emitted

success:

- blob module read paths are instruments with traces
- potentially add some timing metrics too

vgonkivs:
Cover Blob service with Otel traces. Partly fixes #3112


ramin (dismissed):
LGTM

distractedm1nd (approved):

blob/service.go
```diff
@@ -193,16 +202,28 @@ func (s *Service) getByCommitment(
...
+	ctx, span := tracer.Start(ctx, "get-by-commitment")
+	defer func() {
+		utils.SetStatusAndEnd(span, err)
+	}()
+	span.SetAttributes(
+		attribute.Int64("height", int64(height)),
+		attribute.String("commitment", string(commitment)),
+	)
+
 	header, err := s.headerGetter(ctx, height)
```
Wondertan, line 219: We also need to measure the time it takes to get the header, so we should either wrap it with a new span within BlobService or add tracing in given header getter(preffered)

vgonkivs: IMO, it's better to add a trace for fetching headers inside the blob service to detect `GetByHeight` from the blob service directly. Also header.Service relies both on Syncer and Store which are in another repo. So the fastest way to add additional trace here and in case the result is confusing we can additionally cover both syncer and store with traces

vgonkivs: so, my suggestion to keep it :

```go
span.AddEvent("received eds", trace.WithAttributes(
		attribute.Int64("eds-size", int64(len(header.DAH.RowRoots))),
		attribute.Int64("duration", time.Since(now).Milliseconds())),
	)
```

Wondertan: Agree that keeping it here would speed up things.

Wondertan: Regarding keeping duration as an attribute. 
* This makes it harder to debug, and on Jaeger, you would need to look that attribute up and loose other span feature
* Attributes on traces should be treated the same way we do in metrics, where high cardinality is an issue. 

The best way here would be to wrap headerGetter with an additional span

blob/service.go
```diff
@@ -193,24 +204,53 @@ func (s *Service) getByCommitment(
...
 	if err != nil {
+		headerGetterSpan.SetStatus(codes.Error, err.Error())
 		return nil, nil, err
 	}
 
-	namespacedShares, err := s.shareGetter.GetSharesByNamespace(ctx, header, namespace)
+	headerGetterSpan.SetStatus(codes.Ok, "")
+	headerGetterSpan.AddEvent("received eds", trace.WithAttributes(
+		attribute.Int64("eds-size", int64(len(header.DAH.RowRoots))),
+		attribute.Int64("duration", time.Since(now).Milliseconds())),
```
Wondertan, line 233: We don't need to do duration manually if we have it as spans.

distractedm1nd (approved):

Wondertan (approved):

PR #3117, opened 2024-01-20
feat(rpc): rpc.disable-auth flag

distractedm1nd:
In many setups, nodes are running in trusted environments and setting the jwt token is not necessary and very cumbersome.

This PR enables a quick fix by providing a flag/config option to the RPC to disable auth.

Breaking config.

New flag: `--rpc.disable-auth`. 

New default config:
```toml
[RPC]
  Address = "localhost"
  Port = "26658"
  AuthDisabled = false
 ```

_______________
Comes from a discussion with @joroshiba at Astria.

### DIFF
--- a/nodebuilder/rpc/flags.go
+++ b/nodebuilder/rpc/flags.go
@@ -10,6 +10,7 @@ import (
 var (
 	addrFlag = "rpc.addr"
 	portFlag = "rpc.port"
+	authFlag = "rpc.disable-auth"
 )
 
 // Flags gives a set of hardcoded node/rpc package flags.
@@ -26,6 +27,11 @@ func Flags() *flag.FlagSet {
 		"",
 		fmt.Sprintf("Set a custom RPC port (default: %s)", defaultPort),
 	)
+	flags.Bool(
+		authFlag,
+		false,
+		"Disable authentication for RPC requests",
+	)
 
 	return flags
 }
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by ramin — COMMENTED** [review]
I wonder if skip-auth (SkipAuth) might be a cleaner term for it, nicer looking and I never like when something bool-ey has ambiguity or requires you to think. ie: disable / true | false / true | false. vs skip is clearer what is happening and being enabled or not

**Comment by joroshiba** [comment]
For context on the request here: We spin up celestia light nodes in an automated fashion in a fully trusted environment. 

We currently generate a token and spin up a seperate "token server" it via api without auth. So essentially bypassing anyways. It will be much cleaner to not need a second server and startup scripts to fetch the token.

**Review by ramin — DISMISSED** [review]

File: nodebuilder/rpc/flags.go [hunk 1]
```diff
@@ -40,4 +46,11 @@ func ParseFlags(cmd *cobra.Command, cfg *Config) {
 	if port != "" {
 		cfg.Port = port
 	}
+	ok, err := cmd.Flags().GetBool(authFlag)
+	if err != nil {
+		panic(err)
+	}
+	if ok {
+		cfg.SkipAuth = true
```
renaynay, line 54: Log out WARN to user here please

renaynay (approved):

ramin:
just noticed the branch is "godmode" 😆 "hurry up with my damn croissants"

Wondertan (approved):

renaynay (approved):
Fine with me, warning should be more obnoxious but wtv.

PR #3118, opened 2024-01-20
feat(rpc/node): readiness check

distractedm1nd:
Implements a readiness check on the node module.

Question: I have set the permission level to public, but remember that we removed all public methods in favor of `read`. Is this fine or should this method also be `read`?
____________
Comes from a conversation with @joroshiba from Astria


ramin:
arguably fixes https://github.com/celestiaorg/celestia-node/issues/739

ramin:
@distractedm1nd read perms IMO, no public

Wondertan (commented):
This needs a test imo

distractedm1nd:
will add

nodebuilder/tests/api_test.go
```diff
@@ -58,6 +58,10 @@ func TestNodeModule(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, info.APIVersion, node.APIVersion)
 
+	ready, err := client.Node.Ready(ctx)
+	require.NoError(t, err)
+	require.True(t, ready)
+
```
Wondertan, line 64: Does it block when it's not ready?

Wondertan (commented):
CI needs a rerun

nodebuilder/tests/api_test.go
distractedm1nd: For context, see the issue - rn nodes are using some arbitrary endpoint for this like network head

Wondertan: I understand the purpose. My question is what happens when its not ready

distractedm1nd: If its not ready the rpc doesnt connect, so a connection error is returned by whatever library you are using to talk to the RPC. I get that its a bit silly, but we cannot provide an endpoint before the node is ready anyways

Wondertan: So that could be any endpoint. I don't mind having a special endpoint, though. We could also think of any other checks we can add to be sure the node is ready. Like p2p networking is ready and we have some connections

renaynay: Preferably all services started would be nice.

distractedm1nd: Discussed with Rene yesterday - it is a bigger issue but one thing we can ensure is that RPC is only started after all other service's lifecycles have begun. This should be enough for now, and implementing something more robust would require implementing readiness into the services themselves to query. I will add it to the PR today

Wondertan: So RPC server has already started the last. Could you also check if we connect to bootstrappers before RPC?

distractedm1nd: I can confirm. Cannot connect to node until bootstrapper connection is established as it is part of successful startup, and RPC calls fail before startup is finished

renaynay (dismissed):
Fine w me, utack on the RPC being started last. Does moving the construction of the rpc mod last affect the appends to the start hooks?

renaynay:
Why is this breaking @distractedm1nd ?

distractedm1nd:
adds new field to config

nodebuilder/tests/api_test.go
Wondertan: Cool, as the last step let's comment out inside Ready that it's in fact awaiting for Services and connections

ramin: personally always like having a /health(z) that is a simple "is it started" for simple up/down health checks (which this /ready check satisfies and i think the real ask so things like container orchestrators can use just to start serving traffic to the instance) , then a separate /status or something which can validates each internal type of connection/required service, so /status returns a check of all, then /status/{p2p,storage...} might exist for checking each individually. Readiness, vs liveness separately.

Wondertan: In my practice, what you call /healthz is /readyz) then and what you call /status is /healthz. Also, it makes more sense to call simple "is it started" as is /readyz

ramin (approved):

Wondertan (approved):

joroshiba:
@ramin I'd argue doesn't solve need for readiness endpoint because it's a json-rpc. HTTP probes in k8s are exclusively on GET, and JSON-RPC is exclusively post: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#http-probes

ramin:
@joroshiba was the original convo / request for a non-rpc readiness endpoint and not this?

jcstein:
should this be documented? if so, how does one use it?

PR #3132, opened 2024-01-22
fix(libs/pidstore): Detect corruption on startup and reset pidstore

renaynay:
The recovery of an existing pidstore is not critical to node functionality and should not block the node's start-up if it was corrupted during an ungraceful shutdown. If corruption is detected, a fresh pidstore can be instantiated and used by the node.

Found by conduit.


Wondertan (commented):
I can't think of any other reason why It could get corrupted besides an ungrateful shutdown. In this case just doing that only for pidstore might not solve the issue. Lets ask conduit if this fixes the issue for them without resetting the node and only after merge it

renaynay:
@Wondertan I agree that an ungraceful shutdown is most likely the cause for this. IMO fix is still ok to implement as pidstore is non-essential.

ramin (approved):

distractedm1nd (approved):
There are other datastore corruptions that happen, I wonder if we can wrap badger startup similarly or something (not resetting it, but just by giving a more readable error output when detected)

PR #3137, opened 2024-01-24
misc(header/headertest): Rename param for test suite to be more clear

renaynay:
I'm sorry but this was bothering me.

### DIFF
--- a/api/gateway/health.go
+++ b/api/gateway/health.go
@@ -1,6 +1,8 @@
 package gateway
 
-import "net/http"
+import (
+	"net/http"
+)
 
 const (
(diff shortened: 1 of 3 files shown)


ramin (approved):

Wondertan:
wut

renaynay:
@Wondertan I always assumed the test suite would take `num` as number of headers to generate when really it was number of validators.

Wondertan:
There is a comment that clearly states that this is validators, I don't think this is needed

Wondertan:
Now part of #3139

PR #3146, opened 2024-01-29
feat(rpc): implementing go-da interface

distractedm1nd:
Integrates go-da directly to eliminate the need for [celestia-da](https://github.com/rollkit/celestia-da).


ramin:
kind of a lot going on here and i realize it's mostly just copied over but....main question from me before getting into weeds is: "is da and the `go-da` the right naming for this bit of work/module in the node project?". I don't think i have an answer and maybe @Wondertan has more of an opinion on naming.

Only other thing that jumps out at me and could use a comment before deeper review is the Marshall func on blob...i vaguely sense we maybe had a bug around this not that long ago....

distractedm1nd:
@Wondertan and I discussed naming before I started this PR because I had the same concerns. It is very confusing to have both "da" and "das" package. But we came to the conclusion it's the best name for now - but if you have any suggestions I am very open to it, because I hate "da" as well

As per the blob proof marshalling: Yes, there was a bug that nmt.Proof was not json serializable. We added json serialization in the nmt package to solve this. Is this what you mean?

distractedm1nd:
Reminder that the point of this PR is simply to move celestia-da code into node, so that we can debug the existing problems more easily - fixes will be expected, but it needs to get released as soon as possible

Wondertan:
I am OK with calling this da if we change `das` to `sampler` or smlr. Changing, this shouldn't be part of this PR, but we should at least have an issue to clean this ambiguity after this is merged

Wondertan (approved):
utACK

ramin (approved):
new test fails, can we fix or remove then otherwise 👍

distractedm1nd:
Ahhh it's flaky

Wondertan (approved):

ramin (approved):

PR #3147, opened 2024-01-29
ci(unit tests): YOINK unit tests with race detector temporarily

ramin:
suggesting we YEET the `unit test with race` temporarily as the job intermittently fails across a handful of tests pretty regularly and we don't require this job to pass for a PR merge anyway.

Lets remove, i'll get it working and also enable for `macos-latest` which fails ALWAYS on github then we can / replace the normal unit tests with unit tests with race after we get it all running and keep CI green vs just ignoring a red x most of the time

### DIFF
--- a/.github/workflows/go-ci.yml
+++ b/.github/workflows/go-ci.yml
@@ -105,21 +105,27 @@ jobs:
           file: ./coverage.txt
           name: coverage-${{ matrix.os }}
 
-  unit_test_race:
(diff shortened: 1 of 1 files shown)


Wondertan (approved):
Reasoning makes sense to me. Let's land an issue about turning race detector back. There were plenty of cases for me where races were found on CI but not locally.

vgonkivs (approved):

PR #3152, opened 2024-01-31
fix: improve rpc.skip-auth

distractedm1nd:
In the current skip-auth implementation, there was an issue: Authentication could be skipped, but it still required the `Authentication` header to be set (it just wouldn't check it).

This PR improves the devex by removing this requirement. I missed it the first time because I tested by using the CLI, which always sets the header regardless of what you pass. This came up while implementing https://github.com/eigerco/lumina/pull/210, since the wasm client cannot set headers.

I also think the solution is cleaner than the original.

### DIFF
--- a/api/rpc/server.go
+++ b/api/rpc/server.go
@@ -61,15 +61,14 @@ func (s *Server) verifyAuth(_ context.Context, token string) ([]auth.Permission,
 
 // RegisterService registers a service onto the RPC server. All methods on the service will then be
 // exposed over the RPC.
-func (s *Server) RegisterService(namespace string, service interface{}) {
-	s.rpc.Register(namespace, service)
-}
+func (s *Server) RegisterService(namespace string, service interface{}, out interface{}) {
+	if s.authDisabled {
+		s.rpc.Register(namespace, service)
+		return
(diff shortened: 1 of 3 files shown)


api/rpc/server.go
```diff
@@ -61,17 +61,15 @@ func (s *Server) verifyAuth(_ context.Context, token string) ([]auth.Permission,
 
 // RegisterService registers a service onto the RPC server. All methods on the service will then be
 // exposed over the RPC.
-func (s *Server) RegisterService(namespace string, service interface{}) {
+func (s *Server) RegisterService(namespace string, service interface{}, out interface{}, authEnabled bool) {
```
ramin, line 64: there is an `authDisabled` on s already, can we use that and skip passing in the authEnabled? Never love a true or false tagged on the end of a method sig

distractedm1nd: Great catch

Wondertan: Also, there is not much to gain from per service auth granularity

Wondertan (approved):

ramin (approved):

PR #3153, opened 2024-02-01
feat(core)!: separating core.ip into core.grpc.ip and core.rpc.ip

### LINKED ISSUE #2931: [Feature Request][RPC][CLI]: Make providing GRPC and RPC endpoints possible in separate URLs
### Introduction

ATM, we are providing `core.ip` with an assumption that this address has opened ports for both rpc and grpc. In production environment, this is not a norm to do. Moreover, this makes public infra providers serve another endpoint for such purposes, thus creating redundancy and opening more vectors for bad UX in understanding what does what in our docs
https://docs.celestia.org/nodes/mainnet#da-rpc-endpoints

H/t @kjessec @jcstein @marcmacmac for good discussions around this topic 

### What makes this endeavour done: 

- [ ] Node binary accepts 2 new flags, like `--core.grpc.endpoint=grpc.whatever.io` and `--core.rpc.endpoint=rpc.whatever.io`
- [ ] Make secured connection possible

Usually, for secured connections, 80 for rpc and 443 for grpc are used. Configuring it as a concatenation to aforementioned with a `:` sign at the end of the string is a good enough imho and assuming default ports

distractedm1nd:
Closes #2931, although I fail to understand the requirements for the second todo, so can add next.

Breaks config by changing `core.IP` -> (`core.RPCIP`, `core.GRPCIP`).

We keep the `--core.ip` flag, which lets the user set both IPs at once. But it cannot be passed together with either of the new flags.

Failure cases:
image


ramin (commented):
would be good to add a few basic tests around the DefaultConfig and the IP switching while we're here

distractedm1nd:
What do you mean by IP switching? ParseFlags?

ramin:
replaced by #3242

PR #3160, opened 2024-02-03
chore: simple select case when Stop a Manager

htiennv:


Wondertan (changes requested):
This makes the select statement "either/or" rather than "and" for these channels, which is by design

PR #3165, opened 2024-02-05
feat(blob): extend blob struct with index field

vgonkivs:
What was done:
* added index to the blob struct;
* introduced a new `parser` entity that collects shares and transforms them to blob;
* added tests to compare shares received from share service by coordinates that have gotten from blob index;


blob/service.go
```diff
@@ -251,6 +251,18 @@ func (s *Service) getByCommitment(
...
+	for i, row := range header.DAH.RowRoots {
+		if !namespace.IsOutsideRange(row, row) {
+			offset = i
+			break
+		}
+	}
+	if offset == -1 {
+		return nil, nil, ErrBlobNotFound
+	}
+
```
Wondertan, line 265: Why? GetSharesByNamespace does this already iirc

walldiss: I think the idea here is to find offset, while it also allows to short circuit early. As there is demand to return index of commitment, I assume it is a matter of time we would want to add row index to `NamespacedRow` returned by `GetSharesByNamespace`. 

Lets add Idx field to `NamespacedRow`, so it can accessed it here and in `GetSharesByNamespace` rpc as well.

Wondertan: IIRC, the proof from nmt contains the start idx in the row so we can add index method there already.

walldiss: I mean that we can add `rowIdx` to `NamespacedRow`. Using rowIdx and proof.Start index it will be trivial to calculate commitment start index

vgonkivs: The proof contains the start idx of the whole namespace row, which can contain multiple blobs + there might be padding shares.

Wondertan:
We should mark this API and proto breaking

blob/blob.go
```diff
@@ -165,39 +176,88 @@ func (b *Blob) UnmarshalJSON(data []byte) error {
...
+	}
+	blob.index = b.index
+	return blob, nil
+}
+
+func (b *transformer) empty() bool {
+	return b.index == 0 && b.length == 0 && len(b.shares) == 0
+}
+
+func (b *transformer) restore() {
```
Wondertan, line 259: nit: usually called `reset`

walldiss:
>We should mark this API and proto breaking
Looks like there are only new fields added, so it should be backwards compatible.

blob/blob.go
```diff
@@ -165,39 +176,88 @@ func (b *Blob) UnmarshalJSON(data []byte) error {
...
+
+	blob, err := NewBlob(shareVersion, sequence[0].Namespace.Bytes(), data)
+	if err != nil {
+		return nil, err
+	}
+	blob.index = b.index
+	return blob, nil
+}
+
+func (b *transformer) empty() bool {
```
Wondertan, line 255: nit: isEmpty

blob/helper.go
```diff
@@ -75,7 +75,7 @@ func BlobsToShares(blobs ...*Blob) ([]share.Share, error) {
 
 	sort.Slice(b, func(i, j int) bool {
 		val := bytes.Compare(b[i].NamespaceID, b[j].NamespaceID)
-		return val <= 0
+		return val < 0
```
Wondertan, line 78: Was this a bug?

blob/blob.go
```diff
@@ -99,6 +99,10 @@ type Blob struct {
 	// the celestia-node's namespace type
 	// this is to avoid converting to and from app's type
 	namespace share.Namespace
+
+	// index represents index of the first share in the eds.
+	// before data is being published, the index set by default to -1.
```
Wondertan, line 104:
```suggestion
	// index represents index of the blob's first share in the eds.
	// before blob is posted and included onchain, the index set by default to -1.
```

Wondertan: The second sentence might mean that the instance of the blob will have its index changed after posting, but that's not true, and only the retrieved blobs(not created by the user) will have it set.

Wondertan:
```suggestion
	// index represents index of the blob's first share in the eds.
	// -1 by default and only retrieved onchain blobs will have it set 
```

blob/blob.go
```diff
@@ -165,39 +176,88 @@ func (b *Blob) UnmarshalJSON(data []byte) error {
...
+	return shares[index+1:], true
+}
 
-		if amount == len(appShares) {
-			return blobs, nil, nil
-		}
-		appShares = appShares[amount:]
+func (b *transformer) transform() (*Blob, error) {
+	if b.length != len(b.shares) {
+		return nil, errors.New("blob: blob is incomplete")
```
Wondertan, line 222:
```suggestion
		return nil, errors.New("blob: incomplete blob")
```

blob/service.go
```diff
@@ -216,13 +217,14 @@ func (s *Service) Included(
 	return true, resProof.equal(*proof)
 }
 
-// getByCommitment retrieves the DAH row by row, fetching shares and constructing blobs in order to
+// retrieve retrieves the DAH row by row, fetching shares and constructing blobs in order to
```
Wondertan, line 220: This comment is confusing. Retrieve does not retrieve the DAH, but Blob and the Proof.

Wondertan (commented):
Initial review of semantics

Wondertan:
> Looks like there are only new fields added, so it is backwards compatible.

I know, but that's not the point. We need to mark it so that eiger team can react to this PR and update rust client respectively

blob/helper.go
vgonkivs: Yes. But works fine before indexing was introduced.

blob/service.go
vgonkivs: It retrieves the namespace though. Let's change DAH to namespace, wdyt?

Wondertan: No, it retrieves the blob and the proof. It does so by retrieving the whole namespace and thats an implementation detail that we can mention

vgonkivs: The whole method is an implementation detail as it's unexpected and is not part of the API 🤣 . But, ok. will change the description

Wondertan: It's part of the private API

blob/blob.go
```diff
@@ -165,39 +179,88 @@ func (b *Blob) UnmarshalJSON(data []byte) error {
...
-		if err != nil {
-			return nil, nil, err
-		}
+}
 
-		amount := shares.SparseSharesNeeded(length)
-		if amount > len(appShares) {
-			return blobs, appShares, nil
+// setShares sets shares until the blob is completed and returns extra shares back.
+func (b *transformer) setShares(shares []shares.Share) (shrs []shares.Share, isComplete bool) {
```
Wondertan, line 202: nit: set implies overwriting of the data everytime you call set, not adding more, so I suggest renaming to to addShares or putShares

blob/service.go
```diff
@@ -216,13 +217,14 @@ func (s *Service) Included(
 	return true, resProof.equal(*proof)
 }
 
-// getByCommitment retrieves the DAH row by row, fetching shares and constructing blobs in order to
-// compare Commitments. Retrieving is stopped once the requested blob/proof is found.
-func (s *Service) getByCommitment(
+// retrieve retrieves blobs and their proofs under requested namespace and
+// compares Commitments Retrieving is stopped once the requested blob/proof is found.
```
Wondertan, line 221:
```suggestion
// retrieve retrieves blobs and their proofs by requesting the whole namespace and
// comparing Commitments with each blob. 
// Retrieving is stopped once the requested blob/proof is found.
```

blob/service.go
```diff
@@ -335,11 +398,18 @@ func (s *Service) getBlobs(
...
-		return nil, err
+
+	blobs := make([]*Blob, 0)
+	verifyFn := func(blob *Blob, _ Commitment) bool {
+		blobs = append(blobs, blob)
+		return false
 	}
-	return SharesToBlobs(namespacedShares.Flatten())
+
+	_, _, err = s.retrieve(ctx, header.Height(), namespace, nil, verifyFn)
```
Wondertan, line 408: architectural nit: 
So verifyFn is generalized to allow hacking it and injecting custom behavior. The previous SharesToBlobs approach is much cleaner and does precisely what it should. Btw, SharesToBlob is not used, so we should either remove it or update it to support indexing so that we can keep using it her and avoid passing verifyFn. I recommend the latter.

blob/blob.go
```diff
@@ -165,39 +179,88 @@ func (b *Blob) UnmarshalJSON(data []byte) error {
...
+}
+
+func (b *transformer) isEmpty() bool {
+	return b.index == 0 && b.length == 0 && len(b.shares) == 0
+}
+
+func (b *transformer) reset() {
+	b.index = 0
+	b.length = 0
+	b.shares = []shares.Share{}
```
Wondertan, line 265: using nil avoids allocating, but has the same effect.

Wondertan (commented):
Another semantical review

blob/helper.go
renaynay: I don't understand how indexing impacted this comparison though

renaynay (commented):
another general comment, the code inside `retrieve` is getting long/hairy and would benefit from some comments throughout.

blob/service.go
vgonkivs: It was cleaner because we did not care about indexing. Indexing in making the logic more complicated even for getting all blobs

vgonkivs: SharesToBlobs function could be used from external packages in case someone needs it. I'd suggest to leave it.

blob/service_test.go
```diff
@@ -454,3 +534,9 @@ func createService(ctx context.Context, t testing.TB, blobs []*Blob) *Service {
 	}
 	return NewService(nil, getters.NewIPLDGetter(bs), fn)
 }
+
+func index(rowLength, blobIndex int) (row, col int) {
+	row = blobIndex / rowLength
+	col = blobIndex - (row * rowLength)
+	return
+}
```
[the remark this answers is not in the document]
vgonkivs: done

blob/service_test.go
```diff
@@ -85,7 +86,47 @@ func TestBlobService_Get(t *testing.T) {
 				assert.Len(t, blobs, 2)
 
 				for i := range blobs1 {
-					bytes.Equal(blobs1[i].Commitment, blobs[i].Commitment)
+					require.True(t, bytes.Equal(blobs1[i].Commitment, blobs[i].Commitment))
```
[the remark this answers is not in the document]
vgonkivs: done

blob/service.go
```diff
@@ -350,16 +420,11 @@ func toAppShares(shrs ...share.Share) ([]shares.Share, error) {
...
-		if ok {
-			continue
-		}
-
 		appShrs = append(appShrs, *bShare)
 	}
 	return appShrs, nil
 }
+
+func verify(blob *Blob, com Commitment) bool {
```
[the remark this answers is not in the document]
vgonkivs: done

blob/blob.go
```diff
@@ -165,39 +179,88 @@ func (b *Blob) UnmarshalJSON(data []byte) error {
...
 
-// buildBlobsIfExist takes shares and tries building the Blobs from them.
-// It will build blobs either until appShares will be empty or the first incomplete blob will
-// appear, so in this specific case it will return all built blobs + remaining shares.
-func buildBlobsIfExist(appShares []shares.Share) ([]*Blob, []shares.Share, error) {
-	if len(appShares) == 0 {
-		return nil, nil, errors.New("empty shares received")
+// index represents index of the blob's first share in the eds.
+// -1 by default and only retrieved onchain blobs will have it set.
+type transformer struct {
```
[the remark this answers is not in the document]
vgonkivs: renamed

blob/blob.go
vgonkivs: yeap, thanks

blob/helper.go
```diff
@@ -11,21 +11,8 @@ import (
...
-	appShares, err := toAppShares(rawShares...)
-	if err != nil {
-		return nil, err
-	}
-	return parseShares(appShares)
-}
-
 // parseShares takes shares and converts them to the []*Blob.
-func parseShares(appShrs []shares.Share) ([]*Blob, error) {
+func parseShares(appShrs []shares.Share, indexes []int) ([]*Blob, error) {
```
renaynay, line 15: why take `[]int` for indexes ? This feels complicated / unreadable

vgonkivs: `appShrs []shares.Share` can contain multiple blobs, so we need support multiple indexes

vgonkivs: But I don't have any use case where I need multiple indexes, so I think it is fine to assume that parseShares returns single blob instead []Blob

Wondertan (commented):
Last little comments/questions. Looking slightly better

blob/service_test.go
```diff
@@ -27,7 +29,7 @@ import (
 )
 
 func TestBlobService_Get(t *testing.T) {
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute*5)
```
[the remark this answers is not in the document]
vgonkivs: was changed during debugging. will change back to seconds

blob/helper.go
```diff
@@ -11,53 +12,40 @@ import (
...
 
-// SharesToBlobs takes raw shares and converts them to the blobs.
-func SharesToBlobs(rawShares []share.Share) ([]*Blob, error) {
-	if len(rawShares) == 0 {
-		return nil, ErrBlobNotFound
-	}
-
-	appShares, err := toAppShares(rawShares...)
+// parseShares takes shares and converts them to the *Blob.
+func parseShares(appShrs []shares.Share, index int) (*Blob, error) {
```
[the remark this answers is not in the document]
vgonkivs: will merge with the parser under the `transform` method.

blob/parser.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	if index+1 >= len(shares) {
+		return shrs, true
+	}
+	return shares[index+1:], true
+}
+
+// transform parses shares and creates the Blob.
+func (p *parser) transform() (*Blob, error) {
```
Wondertan, line 97: Last nit. If this is a parser and transform is its main operation, it makes sense to name the method as `parse`.

Also consider rearranging methods order for readabilitiy

Wondertan (approved):
Please improve the PR description and add other secondary changes in this PR, like tracing

blob/parser.go
vgonkivs: makes sense

Wondertan (dismissed):

Wondertan (approved):

Wondertan (approved):

renaynay (approved):

jcstein:
- [ ] update examples for rpc docs

PR #3170, opened 2024-02-08
chore(nodebuilder/share): Make module construction more digestable

renaynay:
I tried to clean up the share module construction a bit so that it's easier to parse. Hopefully this is more helpful than the mess that was there prior.

**Also please carefully review this PR!!** I tried my best to make sure that all components are added where they are supposed to be but it's possible I may have missed something (although all related units are passing).


Wondertan (commented):
Looks much cleaner! I guess this was coming out work from networking part of pruning. I feel like it may also help(or the opposite) for the shwap integration. Will carefully review once I am back

ramin (approved):

nodebuilder/share/module.go
```diff
@@ -45,29 +113,64 @@ func ConstructModule(tp node.Type, cfg *Config, options ...fx.Option) fx.Option
...
+func shrexSubComponents() fx.Option {
+	return fx.Provide(
+		func(ctx context.Context, h host.Host, network modp2p.Network) (*shrexsub.PubSub, error) {
+			return shrexsub.NewPubSub(ctx, h, network.String())
+		},
+	)
+}
+
+// shrexGetterComponents provides components for a shrex getter that
+// is capable of requesting
```
Wondertan, line 155: nit: either comment all the func or no.

nodebuilder/share/module.go
```diff
@@ -34,7 +34,75 @@ func ConstructModule(tp node.Type, cfg *Config, options ...fx.Option) fx.Option
 		fx.Supply(*cfg),
 		fx.Error(cfgErr),
 		fx.Options(options...),
-		fx.Provide(newModule),
+		fx.Provide(newShareModule),
+		shrexSubComponents(),
+	)
+
+	bridgeAndFullComponents := fx.Options(
+		fx.Provide(getters.NewStoreGetter),
```
Wondertan, line 42: nit: I would then extract all the provides into Components func

Wondertan (dismissed):
I reviewed it carefully and couldn't find anything suspicious. We had to run the node with these changes to be very sure.

There are still a bunch of things to do to make module construction _really clean_, but I believe this a good step towards that

nodebuilder/share/module.go
```diff
@@ -34,7 +34,75 @@ func ConstructModule(tp node.Type, cfg *Config, options ...fx.Option) fx.Option
...
+				return cfg.PeerManagerParams
+			}),
+			fx.Provide(peers.NewManager),
+			fx.Provide(bridgeGetter),
+			fx.Invoke(func(lc fx.Lifecycle, sub *shrexsub.PubSub) error {
+				lc.Append(fx.Hook{
+					OnStart: sub.Start,
+					OnStop:  sub.Stop,
+				})
+				return nil
```
walldiss, line 69: I was really confused why shrexsub.Pubsub is only started for BN here and not for FN, when it should work for both.  Looked for changes it used to be the same before this PR, so I went deeper. Turned out shrexsub is started implicitly inside peer manager. But you might think that peer manager is provided to BN and FN, so whats the difference? It is started inside peer manager only if it is provided and for BN we do not provide it into peer-manager 🎉 
Really confusing and worth refactoring in separate PR.

nodebuilder/share/module.go
```diff
@@ -34,7 +34,75 @@ func ConstructModule(tp node.Type, cfg *Config, options ...fx.Option) fx.Option
...
 		fx.Error(cfgErr),
 		fx.Options(options...),
-		fx.Provide(newModule),
+		fx.Provide(newShareModule),
+		shrexSubComponents(),
+	)
+
+	bridgeAndFullComponents := fx.Options(
+		fx.Provide(getters.NewStoreGetter),
+		discoveryComponents(cfg),
```
walldiss, line 43: Discovery is used for all types of node, so could be extracted to base components.

nodebuilder/share/module.go
```diff
@@ -34,7 +34,75 @@ func ConstructModule(tp node.Type, cfg *Config, options ...fx.Option) fx.Option
...
+	switch tp {
+	case node.Bridge:
+		return fx.Module(
+			"share",
+			baseComponents,
+			bridgeAndFullComponents,
+			fx.Provide(func() peers.Parameters {
+				return cfg.PeerManagerParams
+			}),
+			fx.Provide(peers.NewManager),
```
walldiss, line 62: What do you think to pass in bool var `useShrexSub` into  `peerManagerComponents` to allow to use same constructor function for BN and FN/LN but with `false` value?

walldiss (commented):
I looked carefully into it and wasn't able to spot any issues. It looks much cleaner now. Thank you for this!

nodebuilder/share/module.go
renaynay: done

Wondertan (approved):

vgonkivs (approved):

Issue #2611, opened 2023-08-24
[Feature Request]: store openrpc.json in releases


jcstein:
### Implementation ideas

This is so that it is historically accessible, and so that storing past versions does not cause any risk of manual user error.

v0.11.0-rc8 would be available at:
https://raw.githubusercontent.com/celestiaorg/celestia-node/v0.11.0-rc8/openrpc.json

v0.11.0-rc9 would be available at:
https://raw.githubusercontent.com/celestiaorg/celestia-node/v0.11.0-rc9/openrpc.json

etc.

ramin:
@jcstein can we do this now? we will have the rpc stuff in a branch right?

jcstein:
also, to catch discrepancies in versions, etc. for example, if I check out to v0.12.1 of celestia-node and then run `make openrpc-gen`, the version reflected is v0.11.0 instead of v0.12.1

jcstein:
bumping this again. also, the API version hasn't been updated in a long time and the result still shows:

```
--> Generating OpenRPC spec
{
    "openrpc": "1.2.6",
    "info": {
        "title": "Celestia Node API",
        "description": "The Celestia Node API is the collection of RPC methods that can be used to interact with the services provided by Celestia Data Availability Nodes.",
        "version": "v0.11.0"
```

Issue #2549, opened 2023-08-09
[Feature Request]: match upcoming node version with API


jcstein:
### Implementation ideas

https://github.com/celestiaorg/celestia-node/pull/2521#issuecomment-1667938153

celestia-node v0.12.0-rc0 = api v0.12.0-rc0

jcstein:
for example, if I check out to v0.12.1 of celestia-node and then run make openrpc-gen, the version reflected is v0.11.0 instead of v0.12.1

jcstein:
checking in on this 👀 or are we going to stick with different versioning?

jcstein:
api specs generated still show v0.11.0 as of v0.13.3 release

jcstein:
https://github.com/celestiaorg/celestia-node/blob/332739f9e68196333bd1dea76fbeaac0f12737f8/nodebuilder/node/admin.go#L13

vgonkivs:
To resolve this issue, we need to set [SemanticVersion](https://github.com/celestiaorg/celestia-node/blob/6382dc13b92d7b40077154aab823367500371cb8/nodebuilder/node/buildInfo.go#L21) instead of the hardcoded value. In this case, `docgen` cmd should be a part of the node binary to get the correct semantic version: 
`celestia docgen ...`

cc @distractedm1nd @Wondertan

distractedm1nd:
Good idea 👍🏼 I'm on board

PR #3171, opened 2024-02-08
feat: add example value for blob.GasPrice

jcstein:
# Summary

Open RPC spec generation errors, due to a missing sample value for blob.GasPrice.

# Problem

Missing sample value in `api/docgen/examples.go`.

Logs from running on v0.13.0:

```json
--> Generating OpenRPC spec
failed to retrieve example value for type: blob.GasPrice on parent 'blob.GasPrice')
{
    "openrpc": "1.2.6",
    "info": {
        "title": "Celestia Node API",
        "description": "The Celestia Node API is the collection of RPC methods that can be used to interact with the services provided by Celestia Data Availability Nodes.",
        "version": "v0.11.0"
    },
```

Output from logs running on v0.11.0, as another check to see that this worked before:

```json
--> Generating OpenRPC spec
{
    "openrpc": "1.2.6",
    "info": {
        "title": "Celestia Node API",
        "description": "The Celestia Node API is the collection of RPC methods that can be used to interact with the services provided by Celestia Data Availability Nodes.",
        "version": "v0.11.0"
    },
```

# Solution

Add sample value.

[Output after testing](https://gist.github.com/jcstein/02c459fb781889405d7ebc802484509a)

## Other issues found while testing

API version is still v0.11.0, which is the same as last few releases, i just ran make openrpc-gen on v0.11.0 and the API version is the same there. This is already, addressed in https://github.com/celestiaorg/celestia-node/issues/2549, and I am highlighting to bring it out from stale state.

### Proposed solutions for other issues found

To avoid this in the future, i would like to see if we can work this [feature request to add openrpc.json on main](https://github.com/celestiaorg/celestia-node/issues/2611) into celestia-node, so that there is validation before publishing specs, and a history of past versions of the openrpc.json, stored on main with releases. otherwise we JTMB the person updating node-rpc-docs [on almost 10000 lines of spec code](https://github.com/celestiaorg/node-rpc-docs/pull/35/files)

> Note, this output must not contain the output which precedes the json:
> `--> Generating OpenRPC spec`

### DIFF
--- a/api/docgen/examples.go
+++ b/api/docgen/examples.go
@@ -51,6 +51,7 @@ var ExampleValues = map[reflect.Type]interface{}{
 	reflect.TypeOf(42):                       42,
 	reflect.TypeOf(byte(7)):                  byte(7),
 	reflect.TypeOf(float64(42)):              float64(42),
+	reflect.TypeOf(blob.GasPrice(0)):         blob.GasPrice(0.002),
 	reflect.TypeOf(true):                     true,
 	reflect.TypeOf([]byte{}):                 []byte("byte array"),
 	reflect.TypeOf(node.Full):                node.Full,


ramin (approved):

vgonkivs (approved):
Thanks for taking care of it 🙏 🚀

jcstein:
my pleasure, thanks for the reviews!

Issue #2492, opened 2023-07-17
misc(share/befp): improve logging in case of `good` errors 


vgonkivs:
We should make more informative logs about `good issues` during BEFP verification so as not to confuse users. There are 4 cases where we can conclude the BEFP is valid but can't finish verification([#1](https://github.com/celestiaorg/celestia-node/blob/6a6c68c42ff940467476c0a6876e7d986b0e84aa/share/eds/byzantine/bad_encoding.go#L196), [#2](https://github.com/celestiaorg/celestia-node/blob/6a6c68c42ff940467476c0a6876e7d986b0e84aa/share/eds/byzantine/bad_encoding.go#L204), [#3](https://github.com/celestiaorg/celestia-node/blob/6a6c68c42ff940467476c0a6876e7d986b0e84aa/share/eds/byzantine/bad_encoding.go#L215), [#4](https://github.com/celestiaorg/celestia-node/blob/6a6c68c42ff940467476c0a6876e7d986b0e84aa/share/eds/byzantine/bad_encoding.go#L224)). In these case we've verified shares but we can't finish BEFP verification, so we can conclude that BEFP is valid.

ptdatta:
Hey @vgonkivs Can I work on it?

vgonkivs:
Yes, thank you 🙏

ptdatta:
@vgonkivs please review

PR #3176, opened 2024-02-10
misc(share/befp): improve logging in case of good errors

ptdatta:
Fixes: #2492


ramin:
@vgonkivs can you nudge this over the line with @ptdatta to try and get this to satisfy closing the issue?

PR #901, opened 2022-07-11
docs(adr): ADR #009: Telemetry

Wondertan:
This is an initial draft for the Telemtery ADR focused on Tracing, as it solves most of our needs(reasoning is in the document). There are also a few other things that I am going to add there soon, but the document is already ready for review and ready to kick off discussions.

Main TODOs:
* [ ] Tracing integration points in the code(almost there)
* [ ] Design section for metrics

PR #3187, opened 2024-02-14
chore: updated path

Frierened:
current path leads to 404 File Not Found as #901 was never merged and ADR-009-telemetry doesn't exist.

remplaced it with the only mention of telemetry in the docs, but that's just my guess


renaynay:
Hey @Frierened Thank you for the contribution :) 

I think rather than temporarily fixing the link, we should just merge #901

PR #3188, opened 2024-02-15
feat(p2p/discovery)!: Implement `archival` discovery + syncing of historic blocks and blobs

renaynay:
This PR introduces the discovery of nodes advertising on the `archival` topic for the purpose of being able to request blobs and blocks that are older than the sampling window. 

In order to do so, there are now 2 instances of discovery (one for `full` nodes -- _all nodes running full block sync within the sampling window_, and one for `archival` nodes -- _nodes that sync, store and serve all historic data_) and 2 instances of peer managers (one for `full` and one for `archival`).

**This PR is _breaking_ due to changes to metrics names.**

Please also note that this PR introduces the ability to perform archival sync of blobs + blocks for **LIGHT AND FULL** nodes only regardless of whether the full nodes have pruning enabled or not. **Bridge nodes that have previously had pruning enabled will NOT be able to request historic data from the consensus core node.** The bridge node must run in archival mode (without `--experimental-pruning` enabled) in order to access historic data.

**Bridge and Full nodes in archival mode (without `--experimental-pruning`)**
```
2024-05-10T15:21:44.490+0200	INFO	share/discovery	discovery/manager.go:30	starting discovery	{"topic": "full"}
2024-05-10T15:21:44.491+0200	INFO	share/discovery	discovery/manager.go:33	advertising to topic	{"topic": "full"}
2024-05-10T15:21:44.491+0200	INFO	share/discovery	discovery/manager.go:30	starting discovery	{"topic": "archival"}
2024-05-10T15:21:44.491+0200	INFO	share/discovery	discovery/manager.go:33	advertising to topic	{"topic": "archival"}
```

**Bridge and Full nodes running `--experimental-pruning`**
```
2024-05-10T15:19:20.698+0200	INFO	share/discovery	discovery/manager.go:30	starting discovery	{"topic": "full"}
2024-05-10T15:19:20.698+0200	INFO	share/discovery	discovery/manager.go:33	advertising to topic	{"topic": "full"}
2024-05-10T15:19:20.698+0200	INFO	share/discovery	discovery/manager.go:30	starting discovery	{"topic": "archival"}
```

**Light nodes**
```
2024-05-10T15:20:42.440+0200	INFO	share/discovery	discovery/manager.go:30	starting discovery	{"topic": "full"}
2024-05-10T15:20:42.440+0200	INFO	share/discovery	discovery/manager.go:30	starting discovery	{"topic": "archival"}
```

-----------------------------------------------------------

**TODO**
- [x] test archival request routing per node type
- [x] renaming some things inside discovery away from being `full`-node-centric
- [x] rebase
- [x] swamp test

For follow-up PR:
- [x] clarify / make disc metrics generic as there will now be 2 instances of discovery happening

Nice to have; 
- [ ] benchmarks for LN


nodebuilder/share/module.go
```diff
@@ -276,3 +291,70 @@ func lightAvailabilityComponents(cfg *Config) fx.Option {
...
+		// archival service discovery
+		fx.Provide(func(
+			lc fx.Lifecycle,
+			pruneCfg *pruner.Config,
+			d *disc.Discovery,
+			h host.Host,
+			r routing.ContentRouting,
+			opt disc.Option,
+		) ([]*disc.Discovery, error) {
+			// if pruner is enabled, no archival service is necessary
```
distractedm1nd, line 306: Don't pruned nodes at least need archivalDisc?

renaynay: here it is: https://github.com/celestiaorg/celestia-node/pull/3188/commits/9604b96b7c9aa6fcf7fbfca31db11f4906487584

nodebuilder/share/module.go
```diff
@@ -276,3 +290,105 @@ func lightAvailabilityComponents(cfg *Config) fx.Option {
...
+
+		opts := []getters.Option{getters.WithArchivalPeerManager(archivalPeerManager)}
+		return opts, disc.WithOnPeersUpdate(archivalPeerManager.UpdateNodePool), nil
+	})
+}
+
+// advertiseHooks provides the lifecycle hooks for both `full` and, if
+// applicable, `archival` discovery.
+func advertiseHooks() fx.Option {
+	return fx.Invoke(func(lc fx.Lifecycle, tp node.Type, cfg *modprune.Config, discs map[string]*disc.Discovery) error {
```
Wondertan, line 362: The fact that we need to hold the map in FX already suggests adding a higher-level component. Something like DiscoveryManager that has discovery CRUD (or at least CR portion for the start.) This will decrease the complexity of node builder setup and avoid hassle with FX

nodebuilder/node.go
```diff
@@ -56,6 +57,9 @@ type Node struct {
 	RPCServer     *rpc.Server     // not optional
 	GatewayServer *gateway.Server `optional:"true"`
 
+	// block store
+	EDSStore *eds.Store `optional:"true"`
```
renaynay, line 61: This is needed for swamp testing (ability to look into EDSStore the same way we can do with header store).

nodebuilder/share/module.go
renaynay: will consider it

share/p2p/discovery/manager.go
```diff
@@ -0,0 +1,49 @@
+package discovery
```
renaynay, line 1: I didn't know where else to leave this comment so I'll do it here: 

Please note that there is one discovery parameters that is shared between both archival + full discovery instances. Right now I don't want to manage several configs per discovery type so i'm just going to leave it as is, but it's just something to note that both instance of discovery are tied to the same configuration options (minus `EnableAdvertise` bc I configure that per instance in nodebuilder).

renaynay: Also the separation between Parameters + options in discovery is kind of ugly. I propose to clean this up in later PRs as tech debt otherwise this PR will continue to grow in scope. 

(e.g. I'd like to clean up the way onPeersUpdate works so that I don't have to do annoying magic with fx to get it set up for archival peer manager)

walldiss:
Great work on the PR. I had a thought after going through it—what if we tried a slightly different approach with the routing and construction? We could introduce something like a PrunedGetter that utilizes two getters under the hood: pruned and archival. It would automatically route requests based on availability, which could help keep the routing logic out of shrex.getter, especially since it’s getting a bit complex. This setup might also streamline construction—you’d just need to set up the archival and pruned shrex getters independently and then link them to the PrunedGetter. Plus, each getter would manage its own peers and discovery, simplifying the need for a complex discovery manager. I think this could make future changes and maintenance a lot easier. Just a thought to consider! 😊

renaynay:
@walldiss This is a good suggestion and I share your concern for dirtying the shrex getter but I think that future-proofing it for other kinds of peer routing is out of scope for this PR (especially given how long it's taking to get it out the door).

I've made an issue describing the problem here and i think it could be worked on as part of partial nodes workstream once that is approved / kicked off: 

https://github.com/celestiaorg/celestia-node/issues/3393

share/p2p/discovery/manager.go
distractedm1nd: I agree with Vlad on this point. I would be comfortable shipping like this given that you or someone else already commits to working on it next

nodebuilder/pruner/module.go
```diff
@@ -78,7 +67,42 @@ func ConstructModule(tp node.Type, cfg *Config) fx.Option {
...
+func disabledPrunerComponents(tp node.Type, baseComponents fx.Option) fx.Option {
+	switch tp {
+	case node.Light:
+		// light nodes are still subject to sampling within window
+		// even if pruning is not enabled.
+		return fx.Options(
+			baseComponents,
+			fx.Provide(func() pruner.AvailabilityWindow {
+				return light.Window
+			}),
```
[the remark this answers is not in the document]
renaynay: Where are you referring?

share/p2p/discovery/manager.go
```diff
@@ -0,0 +1,49 @@
...
+
+func (m *Manager) Start(context.Context) error {
+	ctx, cancel := context.WithCancel(context.Background())
+	m.cancel = cancel
+
+	var err error
+	for _, d := range m.discs {
+		err = errors.Join(err, d.Start(ctx))
+		if err != nil {
+			log.Errorw("failed to start discovery", "err", err)
```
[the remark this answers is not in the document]
renaynay: Changed it a bit and also cancel the ctx if there's an error to short-circuit out of advertise loop in case of err

nodebuilder/tests/prune_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+// 1 BN w/ pruning
+// 3 FN w/ pruning
+// 1 FN archival
+
+// turn on archival BN
+// archival FN syncs against BN
+// turn off archival BN
+// turn on pruning BN
+// spin up 3 pruning FNs, connect
```
walldiss, line 31: Based on the description, it looks like those cases are covered:
- Sync from archival_BN to archival_FN
- Sync from archival_BN+pruning_BN to pruning_FN
- LN sampling from archival_FN

What is the purpose of the second test? It doesn’t seem to be testing anything meaningful. Perhaps it should be broken into separate test cases to isolate routing conditions:

- Pruning_FN can sync from only pruning_FN/BN.
- Pruning_FN from only archival_FN.

Also missing:
- Archival_FN from archival_FN. There is a test syncing from archival_BN, which is similar but not the same.

In the current setup, LN requests from both archival and pruning run at the same time. I would also suggest breaking the test into two:
- LN can manually request archival and recent blobs from ONLY archival_FN.
- LN can manually request recent blobs from ONLY pruning_FN.

share/p2p/discovery/manager.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func (m *Manager) Start(context.Context) error {
+	ctx, cancel := context.WithCancel(context.Background())
+	m.cancel = cancel
+
+	for _, d := range m.discs {
+		err := d.Start(ctx)
+		if err != nil {
+			log.Errorw("failed to start discovery", "err", err)
```
walldiss, line 27: If there are multiple discoveries and last one failed the function will return. Perhaps started ones should be stopped before exit.

share/p2p/discovery/manager.go
```diff
@@ -0,0 +1,51 @@
...
+		discs: discs,
+	}
+}
+
+func (m *Manager) Start(context.Context) error {
+	ctx, cancel := context.WithCancel(context.Background())
+	m.cancel = cancel
+
+	for _, d := range m.discs {
+		err := d.Start(ctx)
```
walldiss, line 25: Please, use context from Start function call instead of the Advertisement context(WithCancel).

share/p2p/discovery/metrics.go
```diff
@@ -53,43 +53,43 @@ func (d *Discovery) WithMetrics() error {
 }
 
 func initMetrics(d *Discovery) (*metrics, error) {
-	peersAmount, err := meter.Int64ObservableGauge("discovery_amount_of_peers",
+	peersAmount, err := meter.Int64ObservableGauge(fmt.Sprintf("%s_discovery_amount_of_peers", d.tag),
```
walldiss, line 56: Nit: could be just:
```
d.tag + "_discovery_amount_of_peers"
```
I personally find it cognitively harder to parse template.

walldiss (commented):
Can't tell if nodebuilder is correct. Changes are quite complex and difficult to fully understand. So relying on tests for correctness. Not a good thing, so we should prioritise refactoring of share module construction.

nodebuilder/pruner/module.go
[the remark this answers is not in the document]
vgonkivs: https://github.com/celestiaorg/celestia-node/pull/3188/files#diff-0b8ebf8b9930a092b6ce9bc7d7c9a95c638c5772d452ba2241a5d344c610fdabR70-R72

nodebuilder/pruner/module.go
```diff
@@ -78,7 +67,42 @@ func ConstructModule(tp node.Type, cfg *Config) fx.Option {
...
+		)
+	case node.Bridge:
+		return fx.Options(
+			baseComponents,
+			fx.Provide(func() pruner.AvailabilityWindow {
+				return archival.Window
+			}),
+			fx.Provide(func() []core.Option {
+				return []core.Option{}
+			}),
```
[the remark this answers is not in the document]
renaynay: Yes unfortunately for the bridge node's core components.

share/p2p/discovery/manager.go
renaynay: this could be resolved if fx called Stop if Start partially fails on the node

share/p2p/discovery/manager.go
renaynay: context from start function is short lived

walldiss: yes, and the Start context is meant to be passed to all Start calls

nodebuilder/share/module.go
```diff
@@ -276,3 +277,85 @@ func lightAvailabilityComponents(cfg *Config) fx.Option {
...
 }
+
+func discoveryComponents(cfg *Config) fx.Option {
+	return fx.Options(
+		discoveryManager(),
+		fullDiscoveryComponents(cfg),
+		archivalDiscoveryComponents(cfg),
+		archivalPeerManager(),
+	)
+}
```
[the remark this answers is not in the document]
renaynay: yea also realised

nodebuilder/share/constructors.go
```diff
@@ -23,17 +24,26 @@ import (
...
 	// fullNodesTag is the tag used to identify full nodes in the discovery service.
 	fullNodesTag = "full"
+	// archivalNodesTag is the tag used to identify archival nodes in the
+	// discovery service.
+	archivalNodesTag = "archival"
 )
 
-func newDiscovery(cfg *disc.Parameters,
-) func(routing.ContentRouting, host.Host, *peers.Manager) (*disc.Discovery, error) {
+func newFullDiscovery(cfg *disc.Parameters,
```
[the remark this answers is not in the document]
renaynay: it's `FullDiscovery` bc it's discovering fulls

nodebuilder/pruner/module.go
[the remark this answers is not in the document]
renaynay: tried to clean it up

nodebuilder/tests/prune_test.go
renaynay: > What is the purpose of the second test? It doesn’t seem to be testing anything meaningful. 

which second test do you mean? There is only one test here that tests to ensure a LN can discover + sync historic blobs off of an archival node in a network topology where pruned FNs are dominant.

> LN can manually request recent blobs from ONLY pruning_FN

We can't control that the LN requests recent blobs only from pruned FNs as archival nodes can serve the exact same data and also advertise as fulls in the network.

renaynay: I agree we should add an additional test case for a pruned FN requesting historic data from archival FN/BN though, but I'd like to put that in an issue and address it after this PR.

share/p2p/discovery/manager.go
renaynay: Okay I will, but that context is ignored inside discovery.Start anyway in favour of an internal context. It just spawns a bunch of routines.

share/p2p/discovery/manager.go
renaynay: https://github.com/celestiaorg/celestia-node/issues/3408

share/p2p/discovery/manager.go
Wondertan: Which point tho?

share/p2p/discovery/manager.go
```diff
@@ -0,0 +1,51 @@
+package discovery
+
+import (
+	"context"
+	"errors"
+)
+
+type Manager struct {
```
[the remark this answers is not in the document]
Wondertan: I really don't want to block on that as this PR is long overdue, but the current state is pretty messy and feels even messier since the last look I made. One of the options I see is to roll back to the last pre-refactoring version, and park cleans up in the issue. If we you see we can have some more time, then better doing it now ofc.

Wondertan (commented):
There are two main comments. The introduced Manager is net negative and the need for EnableAdvertise in config is unclear.

share/p2p/discovery/manager.go
Wondertan: Likely unneeded with approach described in https://github.com/celestiaorg/celestia-node/pull/3188#discussion_r1604349569

share/getters/shrex.go
```diff
@@ -268,3 +290,14 @@ func (sg *ShrexGetter) GetSharesByNamespace(
...
+
+func (sg *ShrexGetter) getPeer(
+	ctx context.Context,
+	header *header.ExtendedHeader,
+) (libpeer.ID, peers.DoneFunc, error) {
+	if sg.archivalPeerManager != nil && !pruner.IsWithinAvailabilityWindow(header.Time(), sg.availabilityWindow) {
+		p, df, err := sg.archivalPeerManager.Peer(ctx, header.DAH.Hash(), header.Height())
+		return p, df, err
+	}
+	return sg.fullPeerManager.Peer(ctx, header.DAH.Hash(), header.Height())
```
walldiss, line 302: Related to optional archivalPeerMan: Lets remove nil check.
```
sg.archivalPeerManager != nil 
```

share/getters/options.go
```diff
@@ -0,0 +1,20 @@
...
+	return func(s *ShrexGetter) {
+		s.archivalPeerManager = manager
+	}
+}
+
+func WithAvailabilityWindow(window pruner.AvailabilityWindow) func(*ShrexGetter) {
+	return func(s *ShrexGetter) {
+		s.availabilityWindow = window
+	}
+}
```
walldiss, line 20: Looking at this codes makes me wonder what is the case when archivalPeerManager is nil? It seems it should always be provided, to allow shrex.Getter to access historcal data. Lets make it archivalPeerManager and availabilityWindow part of constructor and not an option

share/getters/shrex.go
walldiss, line 302: Also if sg.availabilityWindow == 0, it will always result in call of `pruner.IsWithinAvailabilityWindow` to return true, even for historical. This default works well for DASer, but is incorrect here. This default setting will result in all request to be routed to prunned nodes. Perhaps lets disable sg.availabilityWindow == 0 being possible on params validation at least, to ensure misconfiguration does not happen.

nodebuilder/tests/prune_test.go
walldiss: >which second test do you mean? There is only one test here that tests to ensure a LN can discover + sync historic blobs off of an archival node in a network topology where pruned FNs are dominant.

Sorry for misleading wording, I meant the second case I've listened in covered cases. I. referred to  it as second test, because single test covers multiples scenarios.

>We can't control that the LN requests recent blobs only from pruned FNs as archival nodes can serve the exact same data and also advertise as fulls in the network.

Imagine pruned are not available and it always falls back to archival serving. We need the test to ensure pruned can serve too. You can shutdown archival to ensure pruned are able to serve.

share/p2p/discovery/manager.go
walldiss: discovery.Start call does not use context indeed, but it expects short lived one. What is the reason to use incorrect context here?

nodebuilder/share/opts.go
```diff
@@ -11,14 +13,22 @@ import (
 
 // WithPeerManagerMetrics is a utility function to turn on peer manager metrics and that is
 // expected to be "invoked" by the fx lifecycle.
-func WithPeerManagerMetrics(m *peers.Manager) error {
-	return m.WithMetrics()
+func WithPeerManagerMetrics(managers []*peers.Manager) error {
+	var err error
+	for _, m := range managers {
+		err = errors.Join(err, m.WithMetrics())
```
cristaloleg, line 19: Nit: looks like this has `O(N**2)` complexity (see https://cs.opensource.google/go/go/+/refs/tags/go1.22.3:src/errors/join.go;l=19)

Not sure how important it's but collecting errors and than joining them looks simpler. Same in the `func` below and in `share/p2p/discovery/manager.go`

share/getters/options.go
renaynay: At the time, I was unsure about whether every node needs to have an archival service, but now we said that they need to.

share/getters/shrex.go
renaynay: @walldiss full peers are not necessarily pruned 

archival peers also advertise on `full` topic.

nodebuilder/share/module.go
```diff
@@ -276,3 +278,88 @@ func lightAvailabilityComponents(cfg *Config) fx.Option {
...
+			r routing.ContentRouting,
+			opt disc.Option,
+		) ([]*disc.Discovery, error) {
+			discConfig := *cfg.Discovery
+			if tp == node.Bridge || tp == node.Full && pruneCfg.EnableService {
+				// TODO @renaynay: EnableAdvertise is true by default for bridges and fulls and
+				//  as there is no separation for configs per discovery instance, we have to
+				//  do this ugly check here to disable advertisement on the archival topic if
+				//  pruning is enabled.
+				discConfig.EnableAdvertise = false
```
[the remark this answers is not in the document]
renaynay: Reworked in https://github.com/celestiaorg/celestia-node/pull/3188/commits/aa091af3741143fba0305ef5d4fecd7d4f566a28

share/p2p/discovery/options.go
```diff
@@ -12,8 +12,10 @@ type Parameters struct {
 	// PeersLimit defines the soft limit of FNs to connect to via discovery.
 	// Set 0 to disable.
 	PeersLimit uint
+	// EnableAdvertise enables advertisement on the Discovery instance's
+	// topic
+	EnableAdvertise bool
```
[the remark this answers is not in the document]
renaynay: https://github.com/celestiaorg/celestia-node/pull/3188/commits/aa091af3741143fba0305ef5d4fecd7d4f566a28

nodebuilder/share/module.go
```diff
@@ -185,6 +110,46 @@ func shrexGetterComponents(cfg *Config) fx.Option {
...
+			fx.Provide(func(shrexSub *shrexsub.PubSub) shrexsub.BroadcastFn {
+				return shrexSub.Broadcast
+			}),
+			fx.Invoke(func(lc fx.Lifecycle, sub *shrexsub.PubSub) error {
+				lc.Append(fx.Hook{
+					OnStart: sub.Start,
+					OnStop:  sub.Stop,
+				})
+				return nil
+			}),
```
walldiss, line 148: shrexSub subscriber should be part of Full and Light and not be part of Bridge. Somehow it is the opposite in this constructor

nodebuilder/share/module.go
```diff
@@ -241,38 +206,34 @@ func edsStoreComponents(cfg *Config) fx.Option {
...
-		fx.Provide(fx.Annotate(
-			light.NewShareAvailability,
-			fx.OnStop(func(ctx context.Context, la *light.ShareAvailability) error {
-				return la.Close(ctx)
+		)
+	case node.Bridge, node.Full:
+		return fx.Options(
+			fx.Provide(full.NewShareAvailability),
+			fx.Provide(func(avail *full.ShareAvailability) share.Availability {
+				return avail
```
walldiss, line 233: Bridge does not need availability

nodebuilder/share/module.go
```diff
@@ -241,38 +206,34 @@ func edsStoreComponents(cfg *Config) fx.Option {
...
-				light.WithSampleAmount(cfg.LightAvailability.SampleAmount),
-			}
-		}),
-		fx.Provide(func(avail *light.ShareAvailability) share.Availability {
-			return avail
-		}),
-	)
+		)
+	default:
+		panic("invalid node type")
```
walldiss, line 237: Consider returning error like "Node type: %s is not supported for availability"

nodebuilder/share/opts.go
```diff
@@ -11,14 +13,22 @@ import (
 
 // WithPeerManagerMetrics is a utility function to turn on peer manager metrics and that is
 // expected to be "invoked" by the fx lifecycle.
-func WithPeerManagerMetrics(m *peers.Manager) error {
-	return m.WithMetrics()
+func WithPeerManagerMetrics(managers map[string]*peers.Manager) error {
```
walldiss, line 16: What is the reason to use map[string]*peers.Manager? Is it just the way to pass multiple components of same type?

share/getters/options.go
```diff
@@ -0,0 +1,13 @@
...
+	"github.com/celestiaorg/celestia-node/pruner"
+)
+
+type Option func(*ShrexGetter)
+
+func WithAvailabilityWindow(window pruner.AvailabilityWindow) func(*ShrexGetter) {
+	return func(s *ShrexGetter) {
+		s.availabilityWindow = window
+	}
+}
```
walldiss, line 13: I think it should be not an options, but part of constructor

share/p2p/discovery/manager.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func (m *Manager) Start(ctx context.Context) error {
+	advertiseCtx, cancel := context.WithCancel(context.Background())
+	m.cancel = cancel
+
+	for _, d := range m.discs {
+		err := d.Start(ctx)
+		if err != nil {
+			log.Errorw("failed to start discovery", "err", err)
```
walldiss, line 27: Lets make an issue about stopping started discoveries if error happen

share/p2p/peers/options.go
```diff
@@ -82,3 +82,10 @@ func (m *Manager) WithMetrics() error {
 	m.metrics = metrics
 	return nil
 }
+
+func WithTag(tag string) func(manager *Manager) error {
+	return func(m *Manager) error {
+		m.tag = tag
+		return nil
+	}
+}
```
walldiss, line 91: Should be part of the constructor, because it is not optional - needs to be set explicitly.

share/getters/shrex.go
renaynay: I still don't understand why the logic here isn't applicable. `full` topic contains *all* FN + BN in the network, not just pruned, so why would routing peer requests to `full` peers be an issue? It's how the network worked before as well. If it's just semantics, then maybe we can make an issue for it?

nodebuilder/share/module.go
renaynay: @walldiss shrexsub subscriber lifecycle is unfortunately implicitly managed inside peer manager, but we do not pass shrexsub to peer manager for bridge nodes bc we do not care to listen to shrexsub pools. BN still needs to broadcast to shrexsub though so we need to manage shrexsub lifecycle here.

nodebuilder/share/module.go
renaynay: It needs it for the top level share module.

nodebuilder/share/module.go
renaynay: we do this everywhere else.

nodebuilder/share/opts.go
renaynay: we provide this as a type to fx bc we need to pass in the managers in the correct order to shrex getter, so unless I provide them as an []*peers.Manager as well, this will have to do :/

share/p2p/discovery/manager.go
renaynay: We should make it for all services but yes.

share/getters/options.go
renaynay: https://github.com/celestiaorg/celestia-node/pull/3188/commits/da20efc103bf02797140b067d0913396fc196089

nodebuilder/share/module.go
renaynay: https://github.com/celestiaorg/celestia-node/pull/3188/commits/da20efc103bf02797140b067d0913396fc196089

this unlocked new change though

share/p2p/discovery/manager.go
renaynay: #3408

share/p2p/peers/options.go
renaynay: https://github.com/celestiaorg/celestia-node/pull/3188/commits/b48d69673b41689be9a2ff4012c9e4a20c35649c

nodebuilder/share/module.go
walldiss: Ughhh. Good to know. Who might have though it is good idea to manage lifecycle in peer-manager 🥇 https://github.com/celestiaorg/celestia-node/issues/3433
I'm fine if it will be fixed later.

nodebuilder/share/module.go
walldiss: Oh, yeah, it is used for Availability rpc.

share/p2p/discovery/manager.go
walldiss: Well, good to know that we don't stop node on fail. Problem with partially started discoveries will still persist, even if mentioned issue is implemented. So needs to land in mentioned/new issue or to be resolved with modified Stop here.

nodebuilder/share/opts.go
walldiss: Thank you! It didn't clicked for me initially, but now I got the idea!

share/getters/shrex.go
walldiss: Because sg.availabilityWindow == 0 disables routing to prunned nodes. You never want to disable routing here, regardless of local availability window setup. 
Perhaps confusion comes from the fact that it is configurable in the first place, when it should be just a constant.

share/p2p/discovery/discovery.go
```diff
@@ -48,8 +48,11 @@ type Discovery struct {
 	host      host.Host
 	disc      discovery.Discovery
 	connector *backoffConnector
+
 	// onUpdatedPeers will be called on peer set changes
 	onUpdatedPeers OnUpdatedPeers
+	// indicates whether the
+	advertise bool
```
Wondertan, line 55: This is not used anyhow, also comments needs a fix

Wondertan: I guess the idea was to avoid advertising all discoveries at once, but left unfinished

renaynay: It is used, good catch on the comment. I added advertise to the private options (where the callback func for peer manager exists as well)

Wondertan (approved):
I want to ship pruning and unblock this PR, so approving, 
Under different circumstances, I would request a redesign.
This PR leaves too much dech debt behind it

share/p2p/discovery/discovery.go
Wondertan: I see, it was used in the manager accessing the private field :sigh:

walldiss (approved):
Great effort on the nodebuilder in this PR! However, it's still pretty tough to tell if everything's correct. My biggest concern is that the feature needs thorough testing to ensure all setups work as intended. Given its complexity, future changes might get tricky, so let's prioritize refactoring as soon as we have the resources.

Wondertan (approved):

Issue #3194, opened 2024-02-20
bug: logs are misleading and node appears to sample old headers


jcstein:
### Celestia Node version

v0.13.0

### OS

macOS

### Install tools

docs

### Steps to reproduce it

run a light node

### Expected result

it doesn't sample further than 30 days back, per release notes: "This release introduces a light node sampling window of 30 days, specified in [CIP-4](https://github.com/celestiaorg/CIPs/blob/main/cips/cip-4.md). Light nodes will now only sample blocks within a 30 day window instead of sampling all blocks from genesis."

### Actual result

logs are misleading because it looks like it is still sampling further than 30 days back?

```bash
2024-02-19T20:05:46.307+0100	INFO	das	das/worker.go:88	finished sampling headers	{"type": "catchup", "from": 815902, "to": 816001, "errors": 0, "finished (s)": 13.106756458}
```

jcstein:
nice :) 

`2024-06-24T12:46:36.524-0400	INFO	das	das/worker.go:94	finished sampling headers	{"type": "catchup", "from": 16402, "to": 16501, "errors": 0, "# of headers skipped as outside of sampling window": 100, "finished (s)": 5.959832375}`

PR #3196, opened 2024-02-20
chore(das/worker): rename log verb to be clear that not all headers are sampled

ramin:
Addresses #3194 by suggesting we change the log message from "sampling" to "receiving". Other option could be "evaluating"?  Somethign else?

### DIFF
--- a/das/worker.go
+++ b/das/worker.go
@@ -86,7 +86,7 @@ func (w *worker) run(ctx context.Context, timeout time.Duration, resultCh chan<-
 
 	if w.state.jobType != recentJob {
 		log.Infow(
-			"finished sampling headers",
+			"finished receiving headers",
 			"type", w.state.jobType,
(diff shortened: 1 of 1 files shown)


das/worker.go
```diff
@@ -86,7 +86,7 @@ func (w *worker) run(ctx context.Context, timeout time.Duration, resultCh chan<-
 
 	if w.state.jobType != recentJob {
 		log.Infow(
-			"finished sampling headers",
+			"finished receiving headers",
```
renaynay, line 89: I think this would cause more confusion - "recency" in the context of DASer is related to whether or not a header coming in is a new networkHead (a header being gossiped). Headers that are !recent can still be sampled as long as they are > the sampling window cutoff.

jcstein:
so is it not sampling headers older than 30 days, and the logs are misleading?

Issue #3169, opened 2024-02-07
Why two distinct commands to install binary?


rootulp:
## Context

Step 4 of https://docs.celestia.org/nodes/celestia-node#install-celestia-node shows that the install command differs for Mac and Ubuntu. 

Platform | Command | Notes
--- | --- | ---
Mac | `make go-install` | Uses [`go install`](https://pkg.go.dev/cmd/go#hdr-Compile_and_install_packages_and_dependencies) to put binary in $GOPATH/bin (if GOPATH is set) or $HOME/go/bin (if GOPATH is not set)
Ubuntu | `make install` | Uses [`install`](https://en.wikipedia.org/wiki/Install_(Unix)) to put binary in `/usr/local/bin`

## Question

I'm wondering why? A prerequisite to this page is the [development environment](https://docs.celestia.org/nodes/environment) steps which instruct users to Install Go and configure PATH to include:
1. `/usr/local/go/bin`
2. `$HOME/go/bin`

So it seems to me like we only need to support the Go install command. Concretely, on a Linux machine, I can do:

```
root@rootulp-mocha:~/celestia-node# make go-install
--> Installing Celestia

root@rootulp-mocha:~/celestia-node# which celestia
/root/go/bin/celestia

root@rootulp-mocha:~/celestia-node# celestia version
Semantic version: v0.12.4
Commit: 8e5a71746bc9bd9300ad3e4853797e64fdd1d018
Build Date: Wed Feb  7 15:02:52 UTC 2024
System version: amd64/linux
Golang version: go1.21.6
```

## Proposal

1. Remove the existing `make install` command.
2. Alias the existing `make install` command to do the same as `make go-install`.

Wondertan:
It exists to support two ways of installing the binary. The native go way and native linux way which doesn't rely on `GOBIN`. It might be not documented well enough, but we should keep two ways.

Personally, I never use go-install and only install.

Another option I see is to unify them into one `install` directive  and check for `GOBIN` presence which will define the path. Remember that `go install` is basically `go build; mv`

rootulp:
Personally, I always try to use `make install` and hit an error:

```
$ make install
--> Installing Celestia
install: ./build/celestia -> /usr/local/bin//celestia
install: -t: No such file or directory
make: *** [install] Error 71
```

[...]

PR #3197, opened 2024-02-20
chore(Makefile): detect and default to make go-install on darwin vs install everywhere else

ramin:
refs #3169

Suggestion we add this little detect and switch. One issue i see with attempting to detect GOBIN is that it doesn't HAVE to be set (ie: there can be a default). This i would think is a good quick compromise to un-confuse @rootulp and (potentially) other devs and make this behavior more delightful most of the time. IF not, we should close this and the other issue.


Makefile
```diff
@@ -65,9 +66,20 @@ deps:
 
 ## install: Install all build binaries into the $PREFIX (/usr/local/ by default) directory.
 install:
+	@$(MAKE) detect-install
+.PHONY: install
+
+detect-install:
```
Wondertan, line 72: Why not to merge them into a single directive?

walldiss (dismissed):

Makefile
```diff
@@ -65,9 +66,17 @@ deps:
 
 ## install: Install all build binaries into the $PREFIX (/usr/local/ by default) directory.
```
rootulp, line 67: [nit] this line is still correct but it could be made more general:

```suggestion
## install: Install the celestia-node binary.
```

and then a comment could be added above `install-global` to say:

```make
## install-global: Install the celestia-node binary into the $PREFIX (/usr/local/ by default) directory.
install-global:
```

rootulp (commented):
LGTM. The suggestion I made is regarding Makefile comments

jcstein:
for docs, should i just use default for ubuntu/mac as `make install`? after this PR is in the release on the networks?

Wondertan (approved):

PR #3203, opened 2024-02-22
feat: go 1.22.x and pebble v1.1.0

faddat:
- **bump deps**
- **move to local machine: updating pyroscope**
- **Add Grafana profiling libraries**

## Summary

This PR allows Celestia to use pebble, a kv store from cockroachdb that is like Facebook's RocksDB except written in native go.  It makes celestia-appd much more performant, and might make celestia-node faster too, though that is still yet to be seen and understood.  

It's supported by or related to the following pull requests:

* https://github.com/celestiaorg/celestia-app/pull/3124
* https://github.com/celestiaorg/celestia-core/pull/1237
* https://github.com/cometbft/cometbft-db/pull/135
* https://github.com/tendermint/tm-db/pull/359
* https://github.com/cometbft/cometbft/pull/2285

In total, the goal of this pull request is to ensure that Celestia doesn't "run out" of keys in the kv store -- goleveldb does this, and so does pebble, but it happens much, much later with pebble.  Operators of celestia-appd who want archival state will find this set of PR's particularly appealing.  

## celestia-core

I think that these changes are especially important to Celestia core. The reason for that is that in fact, the changes to the database library, allow for a much much simpler build process for celestial core, one that does not even need darker, and one that is many times faster.  I've been trying to contribute it to tendermint or comet for over two years, but it hasn't worked out for whatever reason. 

I'm going to migrate these changes over to celestia core, as the complex docker based build process seen in cometbft is simply unnecessary, and we can make CI vastly faster while speeding up developer onboarding.  

In a very real way, getting this series of PRs landed is a first step to that, as the version of cometbft-db seen here doesn't even have a Dockerfile in the repository.


Wondertan:
Btw, node uses badgerdb for the KVBackend, and if configured well, it can yield faster write performance than pebble. Badger's defaults are horrible though

faddat:
@Wondertan I haven't seen badger beat pebble (ever)

Wondertan:
I've seen

PR #3184, opened 2024-02-12
WIP: Shwap prototype

walldiss:
Prototype for shwap protocol and storage. Contain all changes required for shwap implementation. This PR is not meant for merge. All changes will be extracted from prototype into smaller and nicer PRs. 

Based on: https://github.com/celestiaorg/celestia-node/pull/2675
Tracking issue: https://github.com/celestiaorg/celestia-node/issues/2971

PR #3205, opened 2024-02-23
spec: Shwap

Wondertan:
The core of the spec is ready. All the initial identifiers and containers are specified. I will still be adding Rationale and Consideration related paragraphs.


specs/src/shwap/spec.md
```diff
@@ -0,0 +1,267 @@
...
+## Share Identifiers
+
+This section defines list of supported share identifiers. Share identifiers defined by Shwap can be used to uniquely
+identify any [share container](#share-containers) over a chain with arbitrary number of [DataSquares][square], like a range of 
+[shares][shares], a row or a [blob][blob]. Every share identifier relates to a respective share container and wise-versa.
+
+Identifiers are embeddable to narrow down to the needed content. (TODO: Describe better)
+
+Identifiers MUST have a fixed size for their fields. Subsequently, protobuf can't be used for CID serialization due to
+varint usage. Instead, identifiers use simple binary big endian serialization.
```
oblique, line 79: In your current Shwap PR [little endian is used](https://github.com/celestiaorg/celestia-node/pull/3184/files#diff-ce49ccd8933606535254a3ab0e47d88a0a322cb5b6b3187aa26b7a2deac19b1bR82-R83). We also implemented it with little endian. Which one should we use?

zvolin:
what would you say for enforcing the [`package`](https://protobuf.dev/programming-guides/proto3/#packages) fields in proto definitions in the spec?

specs/src/shwap/spec.md
Wondertan: We will use the big endian. I did a slight research on that and found out that its is used in QUIC with rationale that its just faster on most architectures, so decided to follow that here.

Wondertan:
@zvolin, I find it unnecessary, but might missing some context from other languages.

specs/src/shwap/spec.md
```diff
@@ -0,0 +1,267 @@
...
+This section is a placeholder for future protocol extensions like new new identifiers and containers.
+
+## Considerations
+
+### Bitswap CID integration
+
+The naive question would be: "Why not to make content verification after Bitswap provided it back over its API?"
+Intuitively, this would simplify a lot and wouldn't require "hacking" CID. However, this has an important downside - 
+the Bitswap in such case would consider the content as fetched and valid, sending DONT_WANT message to its peers, while
+the message might be invalid according to the verification rules.
```
oblique, line 257: Because the verification is done in the Bitswap level, the only way knowing that data sampling failed and block should be rejected, is to have a timeout. So, how many seconds the timeout should be? Should we define it here?

Wondertan: Interesting thought. I think this timeout is out of Shwap's scope. The actual sampling logic - how, how many samples are selected, and what the timeout for sampling deserves as separate spec/cip that depends on Shwap protocol.

Wondertan: Shwap is a transport for specialized DA messaging. The sampling logic/details are out of its scope

Wondertan:
@zvolin, my hesitation with `package share.p2p.shwap` is that it enforces a certain package structure for the client implementation. I aim for the spec to be as language-agnostic as possible and broadly free from any non-protocol constraints. 

Making all the proto follow the same package forces all the types to be generated in a single place, like what the core and lumina are doing. Node on the other hand is more modular and we keep proto-types together with the protocols, where each protocol is its own independent module/package. This is an example of an architecture decision that implementers are free to make and enforcing package structure why `package` directive breaks that. 

I nice to be consistent and I see why consistency is good to follow in general, but not in this case.

specs/src/shwap/spec.md
```diff
@@ -0,0 +1,267 @@
...
+varint usage. Instead, identifiers use simple binary big endian serialization.
+
+Table of supported identifiers with their respective multihash and codec codes. This table is supposed to be extended
+whenever any new identifier is added.
+
+| Name     | Multihash | Codec  |
+|----------|-----------|--------|
+| RowID    | 0x7811    | 0x7810 |
+| SampleID | 0x7801    | 0x7800 |
+| DataID   | 0x7821    | 0x7820 |
```
walldiss, line 88: Just for nice aligment, why not those not ordered by desc multihash/codec?

Wondertan:
Noting that spec fully migrated to https://github.com/celestiaorg/CIPs/pull/77. Check description for more info

specs/.gitignore
```diff
@@ -1 +1,2 @@
 book
+theme
```
Wondertan, line 2: @rootulp, the theme dir can be similarly .gitignored in the app's repo

rootulp: thanks! Created https://github.com/celestiaorg/celestia-app/issues/3151

specs/src/shwap/spec.md
```diff
@@ -0,0 +1,267 @@
...
+_**Node**_: The peer that namespacesis both the client and the server.
+
+_**Proof**_: Merkle inclusion proof of the data in the DataSquare.
+
+## Rationale
+
+### Multihashes and CID
+
+Shwap takes inspiration from content addressability, but breaks-free from hash-based only model to optimize message sizes
+and data request patterns. In some way, it hacks into multihash abstraction to make it contain data that isn't in fact a
```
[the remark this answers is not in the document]
Wondertan: Not sure how this change anything to message I tried to convery. Arbitrary message fields are also quite general and can mean anything. 


Maybe something like "identification data" may improve it

specs/src/shwap/spec.md
Wondertan: I agree. I wanted to keep the order of IDs as defined in the spec and I didn't want to break what was already implemented, but as we break other things, it may not hurt to break this one.

specs/src/shwap/spec.md
```diff
@@ -0,0 +1,267 @@
...
+
+This document specifies the Shwap p2p protocol. Shwap provides scalable and extensible framework for exchanging and 
+swapping of shared data for Celestia's Data Availability network and beyond. 
+
+## Motivation
+
+The current Data Availability Sampling (DAS) network protocol is inefficient. A _single_ sample operation takes log2(k) network
+round-trips(where k is the square size). This is not practical and does not scale for the theoretically unlimited data
+square that the Celestia network enables. The main motive here is a protocol with O(1) round-trip for _multiple_ samples, preserving
+the assumption of having 1/n honest peers connected.
```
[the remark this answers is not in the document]
Wondertan: > Does "1/n" means that there needs to be at least one peer out of N, or something else?

1 honest peer.  

> Also, clarification is needed on whether peers must be both honest and possess the requested data shares

1/n is a pretty common assumption in p2p networks. In this context honesty assumes possession. Nevertheless, clarifying it makes. Thanks for pointing this out.

specs/src/shwap/spec.md
```diff
@@ -0,0 +1,267 @@
...
+
+## Share Identifiers
+
+This section defines list of supported share identifiers. Share identifiers defined by Shwap can be used to uniquely
+identify any [share container](#share-containers) over a chain with arbitrary number of [DataSquares][square], like a range of 
+[shares][shares], a row or a [blob][blob]. Every share identifier relates to a respective share container and wise-versa.
+
+Identifiers are embeddable to narrow down to the needed content. (TODO: Describe better)
+
+Identifiers MUST have a fixed size for their fields. Subsequently, protobuf can't be used for CID serialization due to
```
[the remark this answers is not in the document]
Wondertan: That's a good point. I really would like not to have bitswap as a rationale for this, as this breaks Shwap protocol independence. Ideally, we would find a way to avoid this limitation until the CIP is finalized or find another rationale for why this is necessary.

PR #3208, opened 2024-02-25
chore: fix typo

enjoyooor:


ramin:
@enjoyooor we generally try to not accept small typo PRs, just to avoid incentivizing a barrage of them. But in this case, ADRs are a more or a static document, typos and all

ramin:
@enjoyooor would love to work with you on identifying smaller issues to ramp up on the code and contribute changes if welcome? our 'good first issue' label is a good filter if you want to self serve

PR #3223, opened 2024-03-01
misc(blob/service)!: return empty values in case blobs were not found

### LINKED ISSUE #3192: [Feature Request]: `blob_get_all` could return empty `Vec` on "blob not found"
### Implementation ideas

The `blob_get_all` RPC method yields a `Result<Vec<Blob>, Error>`. When no blobs are published under a specified namespace and height, I am getting the forllowing error: 

`error is: ErrorObject { code: ServerError(1), message: "getting blobs for namespace(00000000000000000000000000000000000000736f762d746573742d70): blob: not found\nblob: not found", data: None }`.

Missing blobs feels more like a standard scenario rather than an error.  Currently there is no clean way to distinguish between a missing blob or some other HTTP error, as we can't pattern match on `ErrorObject`. I think it would be more appropriate to return an empty `Vec` in this case.

version: `celestia-node:v0.12.0`

vgonkivs:
Resolves #3192 , https://github.com/celestiaorg/celestia-node/issues/3292

### DIFF
--- a/blob/service.go
+++ b/blob/service.go
@@ -164,6 +164,7 @@ func (s *Service) GetProof(
 
 // GetAll returns all blobs under the given namespaces at the given height.
 // GetAll can return blobs and an error in case if some requests failed.
+// GetAll can return both empty values in case if blobs were not found for the requested namespaces.
 func (s *Service) GetAll(ctx context.Context, height uint64, namespaces []share.Namespace) ([]*Blob, error) {
 	header, err := s.headerGetter(ctx, height)
 	if err != nil {
@@ -186,7 +187,13 @@ func (s *Service) GetAll(ctx context.Context, height uint64, namespaces []share.
 			defer wg.Done()
 			blobs, err := s.getBlobs(ctx, namespace, header)
 			if err != nil {
-				resultErr[i] = fmt.Errorf("getting blobs for namespace(%s): %w", namespace.String(), err)
+				resErr := fmt.Errorf("getting blobs for namespace(%s): %w", namespace.String(), err)
+				if errors.Is(err, ErrBlobNotFound) {
+					log.Debug(resErr)
+					return
+				}
+
+				resultErr[i] = resErr
 				return
 			}
 
@@ -201,10 +208,6 @@ func (s *Service) GetAll(ctx context.Context, height uint64, namespaces []share.
 		if len(resBlobs) > 0 {
 			blobs = append(blobs, resBlobs...)
 		}
-	}
-
-	if len(blobs) == 0 {
-		resultErr = append(resultErr, ErrBlobNotFound)
 	}
 	return blobs, errors.Join(resultErr...)
 }
--- a/blob/service.go
+++ b/blob/service.go
@@ -190,14 +190,13 @@ func (s *Service) GetAll(ctx context.Context, height uint64, namespaces []share.
 			defer wg.Done()
 
 			blobs, err := s.getBlobs(ctx, namespace, header)
-			switch {
-			case err == nil:
-				log.Infow("retrieved blobs", "height", height, "total", len(blobs))
-				resultBlobs[i] = blobs
-			case errors.Is(err, ErrBlobNotFound):
-			default:
+			if err != nil && !errors.Is(err, ErrBlobNotFound) {
 				log.Debugf("getting blobs for namespace(%s): %v", namespace.String(), err)
 				resultErr[i] = err
+			}
+			if len(blobs) > 0 {
+				log.Infow("retrieved blobs", "height", height, "total", len(blobs))
+				resultBlobs[i] = blobs
 			}
 		}(i, namespace)
 	}
@@ -412,8 +411,5 @@ func (s *Service) getBlobs(
 	sharesParser := &parser{verifyFn: verifyFn}
 
 	_, _, err = s.retrieve(ctx, header.Height(), namespace, sharesParser)
-	if len(blobs) == 0 {
-		return nil, ErrBlobNotFound
-	}
-	return blobs, nil
+	return blobs, err
 }
--- a/blob/service.go
+++ b/blob/service.go
@@ -5,6 +5,8 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"slices"
+	"sync"
 
 	sdkmath "cosmossdk.io/math"
 	"github.com/cosmos/cosmos-sdk/types"
@@ -170,46 +172,32 @@ func (s *Service) GetAll(ctx context.Context, height uint64, namespaces []share.
 	}
 
 	var (
-		blobs = make([]*Blob, 0)
-
-		blobsCh = make(chan []*Blob, 1)
-		errCh   = make(chan error, 1)
+		resultBlobs = make([][]*Blob, len(namespaces))
+		resultErr   = make([]error, len(namespaces))
+		wg          = sync.WaitGroup{}
 	)
-
 	for i, namespace := range namespaces {
+		wg.Add(1)
 		go func(i int, namespace share.Namespace) {
 			log.Debugw("retrieving all blobs from", "namespace", namespace.String(), "height", height)
-			blobs, err := s.getBlobs(ctx, namespace, header)
-			if err != nil {
-				err = fmt.Errorf("getting blobs for namespace(%s): %w", namespace.String(), err)
-				errCh <- err
-				log.Debug(err)
-				return
-			}
+			defer wg.Done()
 
-			log.Debugw("receiving blobs", "height", height, "total", len(blobs))
-			blobsCh <- blobs
+			blobs, err := s.getBlobs(ctx, namespace, header)
+			switch {
+			case err == nil:
+				log.Debugw("receiving blobs", "height", height, "total", len(blobs))
+				resultBlobs[i] = blobs
+			case errors.Is(err, ErrBlobNotFound):
+			default:
+				log.Debug("getting blobs for namespace(%s): %w", namespace.String(), err)
+				resultErr[i] = err
+			}
 		}(i, namespace)
 	}
+	wg.Wait()
 
-	for range namespaces {
-		select {
-		case <-ctx.Done():
-			return blobs, ctx.Err()
-		case b := <-blobsCh:
-			blobs = append(blobs, b...)
-		case resultErr := <-errCh:
-			if errors.Is(resultErr, ErrBlobNotFound) {
-				continue
-
-			}
-			err = errors.Join(err, resultErr)
-		}
-	}
-
-	if len(blobs) == 0 && err == nil {
-		err = ErrBlobNotFound
-	}
+	blobs := slices.Concat(resultBlobs...)
+	err = errors.Join(resultErr...)
 	return blobs, err
 }
 
--- a/blob/service.go
+++ b/blob/service.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"math"
-	"sync"
 
 	sdkmath "cosmossdk.io/math"
 	"github.com/cosmos/cosmos-sdk/types"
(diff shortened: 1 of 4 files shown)


ramin (approved):

Wondertan (dismissed):

Wondertan:
This is technically API breaking as someone could depend on the error, so we should mark it as so

walldiss (dismissed):

jcstein:
> 
This is technically API breaking as someone could depend on the error, so we should mark it as so

Isn't this breaking? why were tags and ! removed?

jcstein:
makes sense, thanks! and in this feature, if there are no blobs, empty vector is returned? is there anything else that changes from a user perspective?

ramin (dismissed):

vgonkivs:
> if there are no blobs, empty vector is returned? 

Yeap

> is there anything else that changes from a user perspective?

Single `ErrNotFound` is returned in case blobs won't be found from multiple namespaces.

renaynay (dismissed):
Ok i am fine with this and solution looks cleaner anyway, but this would be WAY more simplified if we did not allow requests for multiple namespaces in a single request.

vgonkivs:
Hey @jcstein, we discussed internally with @renaynay and agreed to return the break label bc the behaviour has been slightly changed.

blob/service.go
```diff
@@ -168,42 +167,47 @@ func (s *Service) GetAll(ctx context.Context, height uint64, namespaces []share.
...
+				err = fmt.Errorf("getting blobs for namespace(%s): %w", namespace.String(), err)
+				errCh <- err
+				log.Debug(err)
 				return
 			}
 
 			log.Debugw("receiving blobs", "height", height, "total", len(blobs))
-			resultBlobs[i] = blobs
+			blobsCh <- blobs
 		}(i, namespace)
```
walldiss, line 189: Blobs were ordered before and now they are not. Could it be possible that some users depend on order of returned blobs? It would break API for such clients.

vgonkivs: Yes, but we've defined this PR as breaking. I'm also not sure that ordering can break smth.

Wondertan: The order must be preserved; imagine asking for data in one order and getting it back in a completely different one. I would complain a lot about such an API personally

vgonkivs: reworked

blob/service.go
```diff
@@ -170,40 +171,31 @@ func (s *Service) GetAll(ctx context.Context, height uint64, namespaces []share.
...
-			if err != nil {
-				resultErr[i] = fmt.Errorf("getting blobs for namespace(%s): %w", namespace.String(), err)
-				return
+			switch {
+			case err == nil:
+				log.Debugw("receiving blobs", "height", height, "total", len(blobs))
+				resultBlobs[i] = blobs
+			case errors.Is(err, ErrBlobNotFound):
+			default:
+				log.Debug("getting blobs for namespace(%s): %w", namespace.String(), err)
```
walldiss, line 189: DebugF

blob/service.go
```diff
@@ -170,40 +171,31 @@ func (s *Service) GetAll(ctx context.Context, height uint64, namespaces []share.
...
-		}
-	}
-
-	if len(blobs) == 0 {
-		resultErr = append(resultErr, ErrBlobNotFound)
-	}
-	return blobs, errors.Join(resultErr...)
+	blobs := slices.Concat(resultBlobs...)
+	err = errors.Join(resultErr...)
+	return blobs, err
```
walldiss, line 198: Would information of which blobs are returned and which resulted in error represented in final response? It should be there to ensure user will be able to retry missing blobs if needned.

vgonkivs: Errors contained namespaces previously and it turned into a mess. The user will have to parse multiple errors in one string which is pretty hard and it is much easier to parse the result blobs and find missing namespaces.

blob/service.go
```diff
@@ -170,40 +178,31 @@ func (s *Service) GetAll(ctx context.Context, height uint64, namespaces []share.
...
-				resultErr[i] = fmt.Errorf("getting blobs for namespace(%s): %w", namespace.String(), err)
-				return
+			switch {
+			case err == nil:
+				log.Debugw("receiving blobs", "height", height, "total", len(blobs))
+				resultBlobs[i] = blobs
+			case errors.Is(err, ErrBlobNotFound):
+			default:
+				log.Debugf("getting blobs for namespace(%s): %v", namespace.String(), err)
+				resultErr[i] = err
```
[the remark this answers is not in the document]
vgonkivs: Joining errors in multiple goroutines will lead to a race.

blob/service_test.go
```diff
@@ -293,17 +294,17 @@ func TestBlobService_Get(t *testing.T) {
 			},
 		},
 		{
-			name: "get all not found",
+			name: "empty result and err when blobs wer not found ",
 			doFn: func() (interface{}, error) {
-				namespace := share.Namespace(tmrand.Bytes(share.NamespaceSize))
-				return service.GetAll(ctx, 1, []share.Namespace{namespace})
+				nid, err := share.NewBlobNamespaceV0(tmrand.Bytes(7))
```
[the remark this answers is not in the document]
vgonkivs: NAmesapcedSize is 29, while share.NewBlobNamespaceV0 is accepting a byte slice that should be less then 10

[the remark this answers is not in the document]
vgonkivs: Used `NamespaceVersionZeroIDSize`

renaynay (dismissed):

blob/service.go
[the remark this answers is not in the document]
renaynay: ah yeah sorry forgot was async

blob/service.go
walldiss: >Errors contained namespaces previously and it turned into a mess. The user will have to parse multiple errors in one string which is pretty hard and it is much easier to parse the result blobs and find missing namespaces.

I agree. Aggregating this information into error is a recipe for disaster. 

Ideal solution would be to not request multiple namespaces as described here https://github.com/celestiaorg/celestia-node/issues/3390#issuecomment-2109879641

However if making API to work with single namespace is not feasible right now alternative approach can be used.  To allow parsable errors returned for result should be changed from:
```
[]*Blob, error
```
To
```
[]Response, error

type Response struct {
    []Blob
    Namespace
    error
}
```

Global error would be set only if call failed entirely. This approach will simplify error handling for user by a lot, while also providing granular control on what was retrieved and what was not and need to be retried.

vgonkivs: I agree that it could solve the problem but in this case, we should rework all our modules to make it consistent.

distractedm1nd (approved):

blob/service.go
```diff
@@ -160,7 +161,14 @@ func (s *Service) GetProof(
...
 // GetAll returns all blobs under the given namespaces at the given height.
-// GetAll can return blobs and an error in case if some requests failed.
+// If all blobs were found without any errors, the user will receive a list of blobs.
+// If the BlobService couldn't find any blobs under the requested namespaces,
+// the user will receive an empty list of blobs along with an empty error.
+// If some of the requested namespaces were not found, the user will receive all the found blobs and an empty error.
+// If there were internal errors during some of the requests,
+// the user will receive all found blobs along with a combined error message.
+//
+// All blobs will preserve the order of the namespaces that were requested.
```
Wondertan, line 171: We should have this comment on the interface

renaynay (dismissed):

walldiss (dismissed):

blob/service_test.go
```diff
@@ -292,17 +295,17 @@ func TestBlobService_Get(t *testing.T) {
...
+				require.NoError(t, err)
+				return service.GetAll(ctx, 1, []share.Namespace{nid})
 			},
 			expectedResult: func(i interface{}, err error) {
 				blobs, ok := i.([]*Blob)
 				require.True(t, ok)
 				assert.Empty(t, blobs)
-				require.Error(t, err)
-				require.ErrorIs(t, err, ErrBlobNotFound)
+				assert.Empty(t, err)
```
renaynay, line 308: nit but asserting nil is more readable

nodebuilder/blob/blob.go
```diff
@@ -19,7 +19,15 @@ type Module interface {
 	Submit(_ context.Context, _ []*blob.Blob, _ blob.GasPrice) (height uint64, _ error)
 	// Get retrieves the blob by commitment under the given namespace and height.
 	Get(_ context.Context, height uint64, _ share.Namespace, _ blob.Commitment) (*blob.Blob, error)
-	// GetAll returns all blobs at the given height under the given namespaces.
+	// GetAll returns all blobs under the given namespaces at the given height.
+	// If all blobs were found without any errors, the user will receive a list of blobs.
+	// If the BlobService couldn't find any blobs under the requested namespaces,
+	// the user will receive an empty list of blobs along with an empty error.
+	// If some of the requested namespaces were not found, the user will receive all the found blobs and an empty error.
+	// If there were internal errors during some of the requests,
```
renaynay, line 27: Do we have a test case for this ?

renaynay (dismissed):

blob/service.go
Wondertan: Another we could do is to return `[][]*Blob, []error` with the order matching the order of the namespaces given

renaynay (approved):

walldiss (approved):

distractedm1nd (approved):

ramin (dismissed):

blob/service_test.go
vgonkivs: done

nodebuilder/blob/blob.go
vgonkivs: added

blob/service_test.go
```diff
@@ -293,17 +298,17 @@ func TestBlobService_Get(t *testing.T) {
...
+				require.NoError(t, err)
+				return service.GetAll(ctx, 1, []share.Namespace{nid})
 			},
 			expectedResult: func(i interface{}, err error) {
 				blobs, ok := i.([]*Blob)
 				require.True(t, ok)
 				assert.Empty(t, blobs)
-				require.Error(t, err)
-				require.ErrorIs(t, err, ErrBlobNotFound)
+				assert.Nil(t, err)
```
cristaloleg, line 311: Still not addressed https://github.com/celestiaorg/celestia-node/pull/3223#discussion_r1656891153

ramin (approved):

blob/service.go
```diff
@@ -190,15 +190,14 @@ func (s *Service) GetAll(ctx context.Context, height uint64, namespaces []share.
...
 
 			blobs, err := s.getBlobs(ctx, namespace, header)
-			switch {
-			case err == nil:
-				log.Infow("retrieved blobs", "height", height, "total", len(blobs))
-				resultBlobs[i] = blobs
-			case errors.Is(err, ErrBlobNotFound):
-			default:
+			if err != nil && !errors.Is(err, ErrBlobNotFound) {
 				log.Debugf("getting blobs for namespace(%s): %v", namespace.String(), err)
```
distractedm1nd, line 194: this should be an error log no?

blob/service_test.go
```diff
@@ -328,6 +330,24 @@ func TestBlobService_Get(t *testing.T) {
 				require.NoError(t, proof.equal(*newProof))
 			},
 		},
+		{
+			name: "internal error",
```
renaynay, line 334: what about testing the case where there's a partial internal error?

e.g. some blobs returned for other namespaces, but an internal error for one of the namespaces?

blob/service_test.go
vgonkivs: wdym? I change `assert.Empty(t, err)` to `assert.Nil(t, err)`

vgonkivs: Ok, you meant to do the same with blobs.

blob/service_test.go
```diff
@@ -334,18 +334,29 @@ func TestBlobService_Get(t *testing.T) {
...
 			},
 			expectedResult: func(res interface{}, err error) {
 				blobs, ok := res.([]*Blob)
-				require.True(t, ok)
-				assert.Empty(t, blobs)
-				require.Error(t, err)
+				assert.True(t, ok)
+				assert.Error(t, err)
+				assert.Contains(t, err.Error(), "internal error")
+				assert.Equal(t, blobs[0].Namespace(), blobs1[0].Namespace())
```
renaynay, line 357: would be happier with more readable naming for the blob arrays here but 🤷🏻

renaynay (approved):

blob/service_test.go
cristaloleg: `assert.NoError(t, err)`

renaynay (approved):

Wondertan (approved):

PR #3227, opened 2024-03-01
ci(triggers): change release process from tag push to release publish

ramin:
Updates the CI workflow trigger to run specifically from a release publish, vs pushing a tag. This will separate any testing we do with with any pre-release or temporary tags from triggering the GitHub workflow to generate a release. Instead, only the release published event should trigger this.


Wondertan (dismissed):

walldiss (dismissed):

MSevey (changes requested):
minor clean up of the if condition.

walldiss (approved):

Wondertan (approved):

Issue #2780, opened 2023-09-29
light/availability: reevaluate light availability success condition


walldiss:
### Implementation ideas

Lets evaluate light sampling algorithm for withholding attack case, since there seems to be a confusion how this attack should be simulated. https://github.com/celestiaorg/celestia-node/pull/2697#discussion_r1340910152

1. Initial simulation script by @distractedm1nd assumes, that `k` amount of shares are selected for sampling. If any of selected shares are not available, light node will only those, that it was able to sample from initial selection. 

2. Another simulation made by me assumes, that light nodes stores `k` amount of shares, regardless if initial selection. In other words it tries to sample new shares until `k` amount is reached.

Current implementation of light availability is different and do not satisfy either of assumtuions. It select random 16 shares from eds and tries to sample them. It will store all successful samples to blockstore, but If any of samples fails, whole sampling operation will fail.  Having some samples unavailable have a high probability in withholding attack case.  Failed sampling operation causes to DASer to retry sampling attempt for the same height. New attempt will select new set of 16 random shares regardless of previous results. All successfully sampled shares from subsequent calls will also be stored as an addition to previous attempts. Since subsequent attempts will still have probabilistic nature of success/failures and can result in multiple retries. And on every attempt successfully amount of sampled shares will likely grow beyond initially preset sampling limit `k`. 

Lets check an example. For sake for simplicity some values are lowered from our defaults:
- sampling amount (k) is set to 4
- eds size 4 (16 shares)
- assume shares with index >9 are withheld 
```
Attempt 1 (fail): 
  selected to be sampled: 3,6,8,12
  sampled: 3,6,8
  failed: 12
  in store:  3,6,8 (3)
  
Attempt 2 (fail): 
  selected to be sampled: 1,4,3,15
  sampled: 1,4,3,15
  failed: 15
  in store:  1,3,4,6,8 (5)
  
Attempt 3 (success): 
  selected to be sampled: 2,3,6,9
  sampled: 2,3,6,9
  failed: 
  in store:  1,2,3,4,6,8,9 (7)
 ``` 
 
 In the example you can see, how more than `k` samples are stored by light node. 
 
Lets evaluate how many shares needs to be sampled in case of withholding attack and modify light availability accordingly if needed.

distractedm1nd:
Light nodes should clearly be doing (2), as it allows for the lowest `c` in a withholding scenario. If we simulate this case, how many light nodes are needed for BarelyRecoverable to hit 99%?

[...]

PR #3239, opened 2024-03-05
feat(share/availability): persist random samples selection in availability call

walldiss:
This PR introduces persistence for sample selection in random sampling. It addresses the issue by storing all failed samples into the datastore, allowing them to be reloaded on the next sampling attempt. This ensures that if the availability call fails fully or partially during the last sampling attempt, the sampling retry will use the same preselected random coordinates of shares.

Provided solution is backwards compatible with previously stored empty byte slice on sampling success, allowing the change to be non-breaking for existing storage. 

Additionally, this PR includes basic refactoring to simplify concurrency logic in availability. It also ensures that errors returned by the call are aligned with the interface declaration in [availability.go](https://github.com/celestiaorg/celestia-node/blob/main/share/availability.go) enhancing code consistency and maintainability.

Resolves https://github.com/celestiaorg/celestia-node/issues/2780


share/availability/light/availability.go
```diff
@@ -67,77 +66,88 @@ func (la *ShareAvailability) SharesAvailable(ctx context.Context, header *header
...
 	}
 
 	// indicate to the share.Getter that a blockservice session should be created. This
 	// functionality is optional and must be supported by the used share.Getter.
 	ctx = getters.WithSession(ctx)
 
+	result := struct {
+		sync.Mutex
+		failed []Sample
+	}{}
```
Wondertan, line 113: nit: this does not need to be a struct but two separate vars, which is more idiomatic

share/availability/light/availability.go
```diff
@@ -67,77 +66,88 @@ func (la *ShareAvailability) SharesAvailable(ctx context.Context, header *header
...
-			if errors.Is(err, context.Canceled) {
-				return err
-			}
-			log.Errorw("availability validation failed", "root", dah.String(), "err", err.Error())
-			if ipldFormat.IsNotFound(err) || errors.Is(err, context.DeadlineExceeded) {
-				return share.ErrNotAvailable
-			}
-			return err
-		}
+	if errors.Is(ctx.Err(), context.Canceled) {
```
Wondertan, line 133: Why only context.Canceled is respected?

Wondertan (approved):
Very well written test

renaynay (approved):
Actually looks nice, minimal changes. I won't be too finicky re serialisation bc i know this will be removed / handled with shwap

share/availability/light/availability.go
walldiss: I made it to align Availability interface according to: https://github.com/celestiaorg/celestia-node/blob/c93713f69b937a3d8bbba221e1c8026512e6d190/share/availability.go#L13

context.Canceled  is the only error that is not technically failed availability.

share/availability/light/sample.go
```diff
@@ -66,11 +68,37 @@ func (ss *squareSampler) samples() []Sample {
...
 
-	return int(n.Int64())
+	return uint16(n.Uint64())
+}
+
+// encodeSamples encodes a slice of samples into a byte slice using little endian encoding.
+func encodeSamples(samples []Sample) []byte {
+	bs := make([]byte, 0, len(samples)*4)
+	for _, s := range samples {
+		bs = binary.LittleEndian.AppendUint16(bs, s.Row)
```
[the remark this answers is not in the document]
walldiss: What do you think could be better alternative?

Wondertan (approved):

share/availability/light/sample.go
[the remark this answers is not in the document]
renaynay: Was just a question why little and not big? I thought you would use big in shwap

renaynay (approved):

PR #3246, opened 2024-03-07
feat(cmd/rpc): Automatically detect running node for RPC requests

### LINKED ISSUE #2859: cmd: command accessing RPC should detect running node
## Problem
We have to specify `node.store` flag every time we want to do an RPC request, but it should be possible to detect running instance automatically.

I am currently playing a lot with CLI, and it's a pretty good experience, however copying `node.store` flag all over again is annoying. In all Unix daemon apps, you don't have to specify the running instance all over again when you use the default setup.

## Implementation 
* Take the default network as NetworkType
* Detect running node type by looking at the directory.
	* If there is a lock, it means a node is running, so we can use it.
* Combine network type and detected node type and use that as `node.store`

mastergaurang94:
Closes #2859.

TL;DR Previously, the node.store flag had to be specified manually for each RPC request. This commit introduces automatic detection of the running node.

Assumptions:
- presence of lock indicates a running node
- specific order of networks (mainnet, mocha, arabica, private, custom) and type (bridge, full, light)
- 1 network will only have 1 running node type. multiple nodes of same network, same type are disallowed (receive `Error: node: store is in use`).
- auth token, other flags still retain prev behavior and have priority
- address, port read in from config
- skipAuth skips auth for trusted setups
- aligns with Unix daemon conventions
- non-default node store & cel-key still require config flags

**Sample Test Cases**
1. Node store set
```
❯ celestia blob get 1318129 0x42690c204d39600fddd3 0MFhYKQUi2BU+U1jxPzG7QY2BVV1lb3kiU+zAK7nUiY= --node.store=$NODE_STORE
{
  "result": {
    "data": "0x676d",
    "share_version": 0,
    "commitment": "0MFhYKQUi2BU+U1jxPzG7QY2BVV1lb3kiU+zAK7nUiY=",
    "index": 23
  }
}
```

2. Node store not set, but flag specified
```
❯ celestia blob get 1318129 0x42690c204d39600fddd3 0MFhYKQUi2BU+U1jxPzG7QY2BVV1lb3kiU+zAK7nUiY= --node.store=
Error: cant access the auth token: root directory was not specified: unable to load in config: open config.toml: no such file or directory
```

3. No node store flag specified, yay
```
❯ celestia blob get 1318129 0x42690c204d39600fddd3 0MFhYKQUi2BU+U1jxPzG7QY2BVV1lb3kiU+zAK7nUiY=
{ // valid result, same as 1. }
```

4. Multiple networks running, will go to mainnet before mocha
```
❯ celestia blob get 1318129 0x42690c204d39600fddd3 0MFhYKQUi2BU+U1jxPzG7QY2BVV1lb3kiU+zAK7nUiY=
{
  "result": "RPC client error: sendRequest failed: http status 401 Unauthorized unmarshaling response: EOF"
}
```

5. Multiple networks running, will go to mocha before arabica
 ```
❯ celestia blob get 1318129 0x42690c204d39600fddd3 0MFhYKQUi2BU+U1jxPzG7QY2BVV1lb3kiU+zAK7nUiY=
{
   "result": "header: given height is from the future: networkHeight: 802095, requestedHeight: 1318129"
}
```

6.  Run node with rpc config Address: 0.0.0.1 and Port: 25231 -- accurately directs request
```
❯ celestia blob get 1318129 0x42690c204d39600fddd3 0MFhYKQUi2BU+U1jxPzG7QY2BVV1lb3kiU+zAK7nUiY=
{
  "result": "RPC client error: sendRequest failed: Post \"http://0.0.0.1:25231\": dial tcp 0.0.0.1:25231: connect: no route to host"
}
```

7. Run node with --rpc.skip_auth=true and config with SkipAuth: true (prints log in rpc/server [here](https://github.com/celestiaorg/celestia-node/blob/main/api/rpc/server.go#L57))
```
❯ celestia blob get 1318129 0x42690c204d39600fddd3 0MFhYKQUi2BU+U1jxPzG7QY2BVV1lb3kiU+zAK7nUiY=
{ // valid result, same as 1. }
```

8. Run node with --rpc.skip_auth=true and config with SkipAuth: false (server skips auth, prints log in rpc/server [here](https://github.com/celestiaorg/celestia-node/blob/main/api/rpc/server.go#L57))
```
❯ celestia blob get 1318129 0x42690c204d39600fddd3 0MFhYKQUi2BU+U1jxPzG7QY2BVV1lb3kiU+zAK7nUiY=
{ // valid result, same as 1. }
```

### CODE REVIEW

File: libs/fslock/locker.go [hunk 1]
```diff
@@ -20,6 +20,15 @@ func Lock(path string) (*Locker, error) {
 	return l, nil
 }
 
+// IsLocked checks if a lock file exists at the given 'path'.
```
mastergaurang94, line 23: can just move this into rpc.go if preferred

Wondertan (changes requested):
Thank you for the pr!

Lets not complicate things and stick to the former but with full swaped with light in the order. 
>In the case that multiple node types are running on 1 network, the request is directed to 1 node in this order: bridge, light, full. Can change that ordering or alternatively, can detect many node types and ask user to clarify via CLI input. I like former approach since this use case seems uncommon.

libs/fslock/locker.go
```diff
@@ -20,6 +20,15 @@ func Lock(path string) (*Locker, error) {
...
+func IsLocked(path string) bool {
+	if _, err := os.Stat(path); errors.Is(err, os.ErrNotExist) {
+		return false
+	}
+
+	return true
+}
```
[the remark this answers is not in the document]
Wondertan: So this PR is blocked

cmd/rpc.go
```diff
@@ -69,6 +72,30 @@ func InitClient(cmd *cobra.Command, _ []string) error {
...
 }
 
+func getStorePath(cmd *cobra.Command) (string, error) {
+	// if node store flag is set, use it
+	if cmd.Flag(nodeStoreFlag).Changed {
+		return cmd.Flag(nodeStoreFlag).Value.String(), nil
+	}
+
+	// try to detect a running node by checking for a lock file
+	defaultNetwork := string(p2p.DefaultNetwork)
```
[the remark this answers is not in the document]
mastergaurang94: done, dropped p2p.defaultNetwork

cmd/rpc.go
```diff
@@ -69,6 +72,30 @@ func InitClient(cmd *cobra.Command, _ []string) error {
...
+	for _, t := range nodeTypes {
+		path, err := DefaultNodeStorePath(t.String(), defaultNetwork)
+		if err != nil {
+			return "", err
+		}
+
+		lockpath := filepath.Join(path, "lock")
+		if fslock.IsLocked(lockpath) {
+			return path, nil
+		}
```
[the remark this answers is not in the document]
mastergaurang94: moved to `store.go`, can add unit test if preferred

nodebuilder/store.go
```diff
@@ -150,6 +151,16 @@ type fsStore struct {
 	dirLock *flock.Flock // protects directory
 }
 
+// IsOpened checks if a directory is in use by checking if a lock file exists.
+func IsOpened(path string) bool {
+	lockPath := lockPath(path)
+	if _, err := os.Stat(lockPath); errors.Is(err, os.ErrNotExist) {
+		return false
+	}
```
Wondertan, line 159: This is not enough. We should check for actual locking besides the file using the flock library.

cmd/rpc.go
```diff
@@ -69,6 +71,31 @@ func InitClient(cmd *cobra.Command, _ []string) error {
...
+			path, err := DefaultNodeStorePath(t.String(), n.String())
+			if err != nil {
+				return "", err
+			}
+
+			if nodebuilder.IsOpened(path) {
+				return path, nil
+			}
+		}
+	}
```
Wondertan, line 94: Let's also move it next to `IsOpened` as `DiscoverOpened` returning the path and the error if not found.

Wondertan (changes requested):

nodebuilder/store.go
mastergaurang94: yup working on that, just pushed initial

mastergaurang94: added now

cmd/rpc.go
mastergaurang94: moved, had to move DefaultNodeStorePath as well

mastergaurang94:
@Wondertan updated!

also, note that flock now leaves .lock file in directory - https://github.com/gofrs/flock/blob/master/flock.go#L54

nodebuilder/store.go
```diff
@@ -150,6 +154,66 @@ type fsStore struct {
 	dirLock *flock.Flock // protects directory
 }
 
+// Finds a single running node and return its store path.
+// Node Type Order: Bridge, Full, Light
+// Network Order: Mainnet, Mocha, Arabica, Private
+func DiscoverOpened() (string, error) {
```
mastergaurang94, line 160: array of paths preferred?

Wondertan:
> also, note that flock now leaves .lock file in director

I am aware and that's ok. The locking functionality should be around the flock syscall, not the existence of the file.

nodebuilder/store.go
Wondertan, line 160:
```suggestion
// DiscoverOpened finds a path of an opened Node Store and returns its path.
// If multiple nodes are running, it only returns the path of the first found node.
// 
// Type preference order: Bridge, Full, Light
// Network preference order: Mainnet, Mocha, Arabica, Private
func DiscoverOpened() (string, error) {
```

nodebuilder/store.go
```diff
@@ -150,6 +154,66 @@ type fsStore struct {
...
+			}
+
+			ok, _ := IsOpened(path)
+			if ok {
+				return path, nil
+			}
+		}
+	}
+
+	return "", nil
```
Wondertan, line 178: I think we should error here and wrap it in getStorePath

Wondertan (changes requested):

nodebuilder/store.go
Wondertan: It's fine to return just the first one. Will extend it multiple once there is a need for that.

Wondertan: Mind writing a unit test for it?

nodebuilder/store.go
mastergaurang94: wrapped

nodebuilder/store.go
mastergaurang94: done, opted to favor network > node type, lmk

mastergaurang94: yup

cmd/rpc.go
```diff
@@ -47,11 +48,11 @@ func RPCFlags() *flag.FlagSet {
...
 	if authTokenFlag == "" {
-		storePath := ""
-		if !cmd.Flag(nodeStoreFlag).Changed {
-			return errors.New("cant get the access to the auth token: token/node-store flag was not specified")
+		storePath, err := getStorePath(cmd)
+		if err != nil {
+			return err
 		}
-		storePath = cmd.Flag(nodeStoreFlag).Value.String()
+
```
Wondertan, line 55: One more thing I realized. We now have a SkipAuth config in rpc module, which removes the need for the token. So we may check for that in node's config and skip getting token. Additionally, cli depends on the default rpc address and port, which may be configurable via config. The really finished goal here would be to cover both of these cases. 

* Discover running node
* Get its config with address and auth information
* Optionally get a token
* Do request.

Wondertan (commented):
Thanks for pushing this forward and writing good tests! 

One minor comment and one major comment about logic improvement

nodebuilder/store.go
```diff
@@ -150,6 +156,73 @@ type fsStore struct {
...
+			}
+		}
+	}
+
+	return "", ErrNoOpenStore
+}
+
+// DefaultNodeStorePath constructs the default node store path using the given
+// node type and network.
+var DefaultNodeStorePath = func(tp string, network string) (string, error) {
```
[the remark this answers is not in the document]
mastergaurang94: for the unit test to inject. Thoughts?

Wondertan:
Hey @mastergaurang94. Gentle ping. We would want to have this PR merged and wonder if you are going to finish it up

mastergaurang94:
@Wondertan yes. Been open too long! Will aim to merge this week. Thank you 🙏🏽

cmd/rpc.go
```diff
@@ -91,6 +108,15 @@ func getStorePath(cmd *cobra.Command) (string, error) {
 	return "", errors.New(errMsg)
 }
 
+func determineProtocol(address string) (protocol string, trimmedAddress string) {
```
mastergaurang94, line 111: needed or default http is okay? lmk

cmd/rpc.go
```diff
@@ -69,6 +87,36 @@ func InitClient(cmd *cobra.Command, _ []string) error {
...
+	}
+
+	if path != "" {
+		return path, nil
+	}
+
+	return "", errors.New(errMsg)
+}
+
+func determineProtocol(address string) (protocol string, trimmedAddress string) {
```
mastergaurang94, line 111: @Wondertan is this needed or http as default is fine? lmk.

Wondertan: You could add a little method to the RPC configuration struct that gives you the full address. There, you would encapsulate logic that builds an address. 

Also, to answer your question, the logic can trim any protocol prefix(by looking at '://') and use it as is. You don't need to detect if that HTTP or https.

nodebuilder/store.go
```diff
@@ -150,6 +156,73 @@ type fsStore struct {
...
+	flk := flock.New(lockPath(path))
+	ok, err := flk.TryLock()
+	if err != nil {
+		return false, fmt.Errorf("locking file: %w", err)
+	}
+
+	err = flk.Unlock()
+	if err != nil {
+		return false, fmt.Errorf("unlocking file: %w", err)
+	}
```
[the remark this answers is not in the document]
mastergaurang94: Don't think its necessary but happy to add (maybe if someone changes/upgrades library, etc): https://github.com/gofrs/flock/blob/master/flock_unix.go#L92-L94

cmd/rpc.go
```diff
@@ -47,16 +50,31 @@ func RPCFlags() *flag.FlagSet {
...
+
+		// load config to get the RPC url and auth requirements
+		cfg, err := nodebuilder.LoadConfig(filepath.Join(storePath, "config.toml"))
 		if err != nil {
-			return fmt.Errorf("cant get the access to the auth token: %v", err)
+			return fmt.Errorf("cant load in config: %v", err)
+		}
+
+		protocol, trimmedAddress := determineProtocol(cfg.RPC.Address)
+		requestURL = fmt.Sprintf("%s://%s:%s", protocol, trimmedAddress, cfg.RPC.Port)
```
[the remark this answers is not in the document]
mastergaurang94: Oops! Yup.. agreed!

mastergaurang94:
@Wondertan anything else?

Wondertan:
I think that's all. I already skimmed through the code and its LGTM. The next step is to get another review from a team member

nodebuilder/store.go
```diff
@@ -150,6 +156,73 @@ type fsStore struct {
...
 
+// DiscoverOpened finds a path of an opened Node Store and returns its path.
+// If multiple nodes are running, it only returns the path of the first found node.
+// Network is favored over node type.
+//
+// Network preference order: Mainnet, Mocha, Arabica, Private
+// Type preference order: Bridge, Full, Light
+func DiscoverOpened() (string, error) {
+	defaultNetwork := []p2p.Network{p2p.Mainnet, p2p.Mocha, p2p.Arabica, p2p.Private}
+	nodeTypes := []nodemod.Type{nodemod.Bridge, nodemod.Full, nodemod.Light}
```
vgonkivs, line 167: I'd recommend modifying `listProvidedNetworks` from `github.com/celestiaorg/celestia-node/nodebuilder/p2p`, making it public, and getting the list of networks altogether(the same with network types). Otherwise, we can forget to include them here, once we'll add new networks/types.

vgonkivs (changes requested):
Left one small comment, but generally LGTM 🚀 . Thanks @mastergaurang94

nodebuilder/store_test.go
```diff
@@ -160,6 +162,103 @@ func TestStoreRestart(t *testing.T) {
...
+				dirMap[key] = dir
+				storeMap[key] = store
+			}
+		}
+
+		mockDefaultNodeStorePath := func(tp string, n string) (string, error) {
+			key := n + "_" + tp
+			return dirMap[key], nil
+		}
+		DefaultNodeStorePath = mockDefaultNodeStorePath
```
walldiss, line 199: Interesting testing approach with replacing global func. Should mockDefaultNodeStorePath return some error in case when key is not present in the map, as we expect it to be there at that point?

cmd/rpc.go
```diff
@@ -69,6 +82,27 @@ func InitClient(cmd *cobra.Command, _ []string) error {
...
+	path, err := nodebuilder.DiscoverOpened()
+	if err != nil {
+		return "", fmt.Errorf("%s: %w", errMsg, err)
+	}
+
+	if path != "" {
+		return path, nil
+	}
+
+	return "", errors.New(errMsg)
```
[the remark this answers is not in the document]
mastergaurang94: yeah

nodebuilder/store_test.go
```diff
@@ -160,6 +162,103 @@ func TestStoreRestart(t *testing.T) {
...
+
+		// Store opened stores in a map (network + node -> dir/store)
+		dirMap := make(map[string]string)
+		storeMap := make(map[string]Store)
+		for _, network := range networks {
+			for _, tp := range nodeTypes {
+				store, dir := initAndOpenStore(t, tp)
+				key := network.String() + "_" + tp.String()
+				dirMap[key] = dir
+				storeMap[key] = store
```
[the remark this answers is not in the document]
mastergaurang94: hmm? https://github.com/celestiaorg/celestia-node/pull/3246/files/5c7b75eaa459023bc775b95d6af0e6ac369ab8e0#diff-ace4f36c5460f455029d810ee006a067e1e779687148c31b626d99c777431986R210

nodebuilder/store_test.go
mastergaurang94: yeah, figured this was simplest.. I added in the error return but shouldn't impact the test in its current form

nodebuilder/store.go
mastergaurang94: I added in a slice to handle the ordering since apparently, maps randomize order when iterating. Also, added support for custom networks now too.

nodebuilder/p2p/network.go
```diff
@@ -68,16 +69,30 @@ var networkAliases = map[string]Network{
...
+// GetOrderedNetworks provides a list of all known networks in order of priority.
+func GetOrderedNetworks() []Network {
+	return append([]Network(nil), orderedNetworks...)
+}
+
+// listAvailableNetworks provides a string listing all known long-standing networks for things like CLI hints.
+func listAvailableNetworks() string {
+    var networks []string
+    for _, net := range orderedNetworks {
+		// "private" networks are configured via env vars, so skip
```
walldiss, line 84: Nit: remove extra tab for nice alignment

walldiss (commented):
Looks good! Just one small nit on formatting

nodebuilder/p2p/network.go
mastergaurang94: think it was a spaces vs tabs thing.. thx

nodebuilder/p2p/network.go
```diff
@@ -68,16 +69,31 @@ var networkAliases = map[string]Network{
...
-// command hints.
-func listProvidedNetworks() string {
-	var networks string
-	for net := range networksList {
-		// "private" network isn't really a choosable option, so skip
+// orderedNetworks is a list of all known networks in order of priority.
+var orderedNetworks = []Network{Mainnet, Mocha, Arabica, Private}
+
+// GetOrderedNetworks provides a list of all known networks in order of priority.
+func GetOrderedNetworks() []Network {
```
Wondertan, line 76: last nit: let's remove the word ordered from the method. It's redundant as slice already implies an order.  You can keep the priority ordering in the comment though

mastergaurang94: Fair

Wondertan: doesn't matter. I mostly care about public methods

Wondertan (approved):
Thank you for this great contribution and for persistence on resolving multiple requested changes!

mastergaurang94:
Haven't seen the approval in a long time 🥹 Thank you for lending your time & energy to this review @Wondertan! Had fun

vgonkivs (approved):
Great work. Thank you 🚀

vgonkivs:
cc @jcstein for visibility

jcstein:
Cross-posting from slack:

This flag also appears:
1. in the cel-key utility: https://docs.celestia.org/developers/celestia-node-key#steps-for-generating-node-keys
2. for using a different directory than the standard dir: https://docs.celestia.org/nodes/celestia-node-troubleshooting#changing-the-location-of-your-node-store

these are the two that i found on a quick search

what impact does this have on `cel-key` utility or someone wishing to use a different directory than the default?

Wondertan:
> what impact does this have on cel-key utility or someone wishing to use a different directory than the default?

@jcstein
No impact on cel-key, and I think, as it stands, cel-key does not integrate automatic path resolution.

For non-default path users, they have to specify the path using the flag.

jcstein:
> For non-default path users, they have to specify the path using the flag.

Got it, thanks @Wondertan. If I were running with a non-default node store, making an RPC request, I would still use the node.store flag, correct?

Wondertan:
>  No impact on cel-key, and I think, as it stands, cel-key does not integrate automatic path resolution.

Correction after looking at the code. cel-key does have automatics but they aren't as smart as in this PR and this PR doesn't chang those

Wondertan:
> Got it, thanks @Wondertan. If I were running with a non-default node store, making an RPC request, I would still use the node.store flag, correct?

Yes

mastergaurang94:
Thanks @vgonkivs & @walldiss! Also, PR will need to be labeled by someone who has access for merge

jcstein:
gm @mastergaurang94 - I'm working on the docs for this. can you help me understand which network(s) you were running a node(s) for on for 4-6?

> 4. Multiple networks running, will go to mainnet before mocha
> ```bash
> ❯ celestia blob get 1318129 0x42690c204d39600fddd3 0MFhYKQUi2BU+U1jxPzG7QY2BVV1lb3kiU+zAK7nUiY=
> {
>  "result": "RPC client error: sendRequest failed: http status 401 Unauthorized unmarshaling response: EOF"
> }
I'm assuming you had a mocha node running, the request was sent to mainnet node, but looks like the call failed?

---

> 5. Multiple networks running, will go to mocha before arabica
> ```bash
> ❯ celestia blob get 1318129 0x42690c204d39600fddd3 0MFhYKQUi2BU+U1jxPzG7QY2BVV1lb3kiU+zAK7nUiY=
> {
>   "result": "header: given height is from the future: networkHeight: 802095, requestedHeight: 1318129"
> }
> ```

I'm guessing you had mocha and arabica running?

---

> 6. Run node with rpc config Address: 0.0.0.1 and Port: 25231 -- accurately directs request
> ```bash
> ❯ celestia blob get 1318129 0x42690c204d39600fddd3 0MFhYKQUi2BU+U1jxPzG7QY2BVV1lb3kiU+zAK7nUiY=
> {
>   "result": "RPC client error: sendRequest failed: Post \"http://0.0.0.1:25231\": dial tcp 0.0.0.1:25231: connect: no route to host"
> }
> ```

It looks like this request fails?

mastergaurang94:
hey @jcstein, sure:

4) mainnet full, mocha light. re: fail, I confirmed the request went to the right node, not the validity of the response. 401 showed up in the mainnet node logs (that blob nor namespace exists there)
5) yes, both mocha full & arabica light. re: header in future, I had deleted my node store so new one hadn't caught up yet.

main thing, node types are unique. these were the cases I listed but any variation will work

6) similar to above, the request directs to http://0.0.0.1:25231. failure because I wasn't running anything at that address, if I was, it would have to that node

jcstein:
awesome, thanks for the details here @mastergaurang94 

could you please take a look [at the corresponding docs PR](https://github.com/celestiaorg/docs/pull/1532)?

PR #3247, opened 2024-03-08
ci(lint): upgrade golangci-lint to 1.56 and fix revive errors that emerged

ramin:
with our `1.22` upgrade i got inadvertently upgraded to `golangci-lint` 1.56 locally and experiencing different results to what was in CI so figured, why not upgrade, i believe this upgrade makes 'revive' active and a replacement to golint, which introduced some better/stricter checks

- upgrades `golangci-lint` to 1.56
- fixes or marks `//nolint:revive` a few new unused variable errors that were triggering
- configures revive to ignore `_test` files
- CI passes with new checks

### DIFF
--- a/.github/workflows/ci_release.yml
+++ b/.github/workflows/ci_release.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - ci/ramin/lint
   release:
     types: [published]
   pull_request:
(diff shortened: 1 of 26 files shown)


mastergaurang94 (approved):
same

Wondertan (approved):

PR #3254, opened 2024-03-13
fix(nodebuilder/blob): use state module instead of the state accessor in the module construction

### LINKED ISSUE #3244: bug(nodebuilder/blob): Blob module needs to be able to take a stubbed state module as Submitter
Currently, if no core endpoint is provided and the blob module is used for blob submission, we get nasty errors from RPC: 

```
panic in rpc method 'blob.Submit': runtime error: invalid memory address or nil pointer dereference
```

It doesn't crash the node, but it's not clean. Use stubbed state module instead for cleaner err.

vgonkivs:
Resolves #3244


Wondertan (approved):

walldiss (approved):
The art of FX

PR #3267, opened 2024-03-24
fix bug in da.Validate api

tsite:


walldiss (approved):
Thank you for fixing this one!

Wondertan (commented):
Might need a review from @tzdybal as an author of the code

walldiss:
The auther per github is @distractedm1nd See:
https://github.com/celestiaorg/celestia-node/pull/3146

Wondertan:
Ryan just copy-pasted code from go-da instead of merging histort

Wondertan:
This is the original code https://github.com/rollkit/celestia-da/blob/main/celestia/celestia.go#L158-L178. 
Would be great to submit this fix there @tsite 🙏🏻

tsite:
created https://github.com/rollkit/celestia-da/pull/90

tsite:
I think the rollkit version is okay actually and the bug only exists in celestia-node

Wondertan:
That's right, @tsite. Appends on empty slice work correctly there. Seems like the code was slightly bugified during the copy process 😄. Thanks for catching this again!

Wondertan (approved):

PR #3280, opened 2024-03-26
fix(p2p/discovery): Unregister metrics callback on close

renaynay:
Self explanatory


Wondertan (approved):
Thanks! Iirc, the same issue is also in daser. Can you check?

walldiss (approved):

ramin (approved):

PR #3281, opened 2024-03-27
fix(metrics): Unregister callback on stop for several metrics implementations

renaynay:
Self explanatory


share/eds/cache/cache.go
```diff
@@ -38,6 +38,9 @@ type Cache interface {
 
 	// EnableMetrics enables metrics in Cache
 	EnableMetrics() error
+
+	// CloseMetrics closes metrics in Cache
+	CloseMetrics() error
```
walldiss, line 43: Modify Cache interface to return unreg func from EnableMetrics(). This will make unreg  to be available only when metrics are enabled.
```
EnableMetrics() (unreg func(), err error)
```

renaynay: done https://github.com/celestiaorg/celestia-node/pull/3281/commits/1c2188e5299497cb7053186afdc8eb80e931dd8b

share/p2p/discovery/discovery.go
```diff
@@ -114,7 +114,7 @@ func (d *Discovery) Start(context.Context) error {
 
 func (d *Discovery) Stop(context.Context) error {
 	d.cancel()
-	return nil
+	return d.metrics.close()
```
walldiss, line 117: Here error from from closing metrics is propagated from `Stop()`.But in In some modules error is only logged and not returned. Example:
```
if err := d.sampler.metrics.close(); err != nil {
	log.Errorw("closing metrics", "err", err)
}
```
Also in cases when metrics unregistered directly by fx.LifeCycle it is also propagated to fx. 
Lets make it consistent and not propagate, as it does not matter if unregister failed. Also log severity cpuld be lowered to WARN from ERROR for the same reason.

renaynay: https://github.com/celestiaorg/celestia-node/pull/3281/commits/3ffcdeb2644a3ef07a891c59a57dc7d87fdfd3e1

das/daser.go
```diff
@@ -132,6 +132,11 @@ func (d *DASer) Stop(ctx context.Context) error {
 	}
 
 	d.cancel()
+
+	if err := d.sampler.metrics.close(); err != nil {
+		log.Errorw("closing metrics", "err", err)
```
walldiss, line 137: Warn

share/p2p/peers/manager.go
```diff
@@ -166,6 +166,10 @@ func (m *Manager) Start(startCtx context.Context) error {
 func (m *Manager) Stop(ctx context.Context) error {
 	m.cancel()
 
+	if err := m.metrics.close(); err != nil {
+		log.Errorw("closing metrics", "err", err)
```
walldiss, line 170: Warn

walldiss (commented):
warning: few warns missed 😄

walldiss (approved):

vgonkivs (approved):

Wondertan (approved):

PR #3261, opened 2024-03-19
[WIP] Bridge node skips storing EDS if historic on sync

renaynay:
Based on #3188

PR #3283, opened 2024-03-28
feat(core): Bridge node short-circuits storing a historical EDS if pruning is enabled

renaynay:
This PR implements short-circuiting storing EDSes in the case that they are outside the sampling window in the case of a syncing bridge node if `--experimental-pruning` is enabled.

Overrides #3261


core/exchange.go
```diff
@@ -150,11 +155,14 @@ func (ce *Exchange) Get(ctx context.Context, hash libhead.Hash) (*header.Extende
 			&block.Height, hash, eh.Hash())
 	}
 
-	ctx = ipld.CtxWithProofsAdder(ctx, adder)
-	err = storeEDS(ctx, eh.DAH.Hash(), eds, ce.store)
+	ctx = ipld.CtxWithProofsAdder(ctx, adder) // TODO @renaynay: should we short-circuit this if pruning enabled
```
renaynay, line 158: cc @walldiss 

we can make this conditional on whether block is outside avail window

walldiss: Yes

core/listener_test.go
```diff
@@ -79,17 +80,11 @@ func TestListenerWithWrongChainRPC(t *testing.T) {
...
 	fetcher, _ := createCoreFetcher(t, cfg)
 	eds := createEdsPubSub(ctx, t)
 
 	store := createStore(t)
-	err := store.Start(ctx)
-	require.NoError(t, err)
-	t.Cleanup(func() {
-		err = store.Stop(ctx)
-		require.NoError(t, err)
-	})
```
Wondertan, line 92: Did this get removed as unnecessary?

Wondertan (approved):
LGTM and the proofs have to be excluded when outside of the window, as you state

core/listener_test.go
renaynay: Yes, `createStore` manages lifecycle

core/exchange.go
```diff
@@ -190,10 +195,33 @@ func (ce *Exchange) getExtendedHeaderByHeight(ctx context.Context, height *int64
...
+	eh *header.ExtendedHeader,
+	eds *rsmt2d.ExtendedDataSquare,
+	adder *ipld.ProofsAdder,
+) error {
+	if !pruner.IsWithinAvailabilityWindow(eh.Time(), ce.availabilityWindow) {
+		log.Debugw("skipping storage of historic block", "height", eh.Height())
+		return nil
+	}
+
+	ctx = ipld.CtxWithProofsAdder(ctx, adder) // TODO @renaynay: should we short-circuit this if pruning enabled
```
Wondertan, line 217: Comment seems irrelevant anymore

walldiss (approved):
LGTM.

vgonkivs (approved):

abstraking (approved):

Wondertan (approved):

vgonkivs (approved):

PR #3291, opened 2024-04-03
feat(ci): call reusable workflow to track the release deployments and notify devops

MSevey:
This PR makes use of the reusable workflow that creates an EPIC in the devops repo and notifies the devops team on slack to help with tracking release deployments on our testnets and mainnet infrastructure.

### DIFF
--- a/.github/workflows/create_release_tracking_epic.yml
+++ b/.github/workflows/create_release_tracking_epic.yml
@@ -1,0 +1,14 @@
+name: Create Release Tracking Epic
+
+# This workflow creates an EPIC in the devops repo and notifies the devops team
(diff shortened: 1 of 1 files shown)


Wondertan (approved):
utAck

ramin (approved):
looks good, we could potentially call this from our main release pipeline except being its own file? I'm not sure if that is better or not other than maybe having one less workflow file? I like this as is though yay

MSevey:
> looks good, we could potentially call this from our main release pipeline except being its own file? I'm not sure if that is better or not other than maybe having one less workflow file? I like this as is though yay

You could combine them. Just depends which is easier to maintain for the team. 
It is mostly preference.

PR #3297, opened 2024-04-05
fix(nodebuilder): return fx error if theres an error getting keystore or constructing signer

renaynay:
We never did anything with the errors provided to fx, we can just short-circuit and return quickly if either fails.

### DIFF
--- a/nodebuilder/module.go
+++ b/nodebuilder/module.go
@@ -26,11 +26,11 @@ func ConstructModule(tp node.Type, network p2p.Network, cfg *Config, store Store
 	log.Infow("Accessing keyring...")
 	ks, err := store.Keystore()
 	if err != nil {
-		fx.Error(err)
+		return fx.Error(err)
 	}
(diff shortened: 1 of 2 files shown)


vgonkivs (approved):

nodebuilder/module.go
```diff
@@ -26,11 +26,11 @@ func ConstructModule(tp node.Type, network p2p.Network, cfg *Config, store Store
 	log.Infow("Accessing keyring...")
 	ks, err := store.Keystore()
 	if err != nil {
-		fx.Error(err)
+		return fx.Error(err)
```
Wondertan, line 29: I presume it reports the error to FX and then returns it, correct?

Wondertan (approved):

nodebuilder/module.go
renaynay: provides an err to fx and fx checks that err before constructing app and if err != nil, return

ramin (approved):

walldiss (approved):

Issue #3296, opened 2024-04-04
[Feature Request]: async blob submissions


Bidon15:
### Implementation ideas

the api should have a way to let users submit async blob submission calls
should be delivered in tandem with https://github.com/celestiaorg/celestia-app/issues/5672

ref: https://github.com/celestiaorg/celestia-node/issues/3256

0xRampey:
Any plans to return result statuses through a channel on async blob submission? Like `INCLUDED`, `FINALIZED`, `CANCELLED` etc

renaynay:
@0xRampey yes something like that, we are considering adding an endpoint to blob module that would be able to submit a tx and query for its status, returning the result through a channel once included.

Wondertan:
The current Signer integration doesn't make blob submission async. Yes, it changed the tx submission mode from Block to Sync, but [this line](https://github.com/celestiaorg/celestia-app/blob/171a54769bd8b87e60b3d7107f3fc050767c3549/pkg/user/signer.go#L197) makes it equivalent to what we had before

PR #3298, opened 2024-04-05
feat(state): Integrate new Signer for concurrent tx submission

cmwaters:
Closes: #3296

This adds the [`Signer`](https://github.com/celestiaorg/celestia-app/blob/bcab06582c9802f1d49a2bee86820792258bb587/pkg/user/signer.go#L31) struct from celestia-app's user package. 

Signer keeps track of both the network and local sequence number allowing for multiple messages to be signed within a single block.

This also uses the `EstimateGas` function of the signer to estimate the gas, when the user provides 0 as the `gasLim`

TODO: we should probably introduce some gas multiplier (default: 1.1) to the `EstimateGas` as it seems to underestimate the value (cc @vgonkivs)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/BurntSushi/toml v1.3.2
 	github.com/alecthomas/jsonschema v0.0.0-20220216202328-9eeeec9d044b
 	github.com/benbjohnson/clock v1.3.5
-	github.com/celestiaorg/celestia-app v1.7.0
+	github.com/celestiaorg/celestia-app v1.8.0-rc0
 	github.com/celestiaorg/go-fraud v0.2.0
 	github.com/celestiaorg/go-header v0.6.0
 	github.com/celestiaorg/go-libp2p-messenger v0.2.0
--- a/nodebuilder/state/keyring.go
+++ b/nodebuilder/state/keyring.go
@@ -13,9 +13,6 @@ const DefaultAccountName = "my_celes_key"
 // as having keyring-backend set to `file` prompts user for password.
 func Keyring(cfg Config, ks keystore.Keystore) (kr.Keyring, string, error) {
 	ring := ks.Keyring()
-	if ring == nil {
-		panic("keyring is nil")
-	}
 	var info *kr.Record
 	// if custom keyringAccName provided, find key for that name
 	if cfg.KeyringAccName != "" {
--- a/nodebuilder/state/keyring.go
+++ b/nodebuilder/state/keyring.go
@@ -3,26 +3,26 @@ package state
 import (
 	kr "github.com/cosmos/cosmos-sdk/crypto/keyring"
 
-	apptypes "github.com/celestiaorg/celestia-app/x/blob/types"
-
 	"github.com/celestiaorg/celestia-node/libs/keystore"
-	"github.com/celestiaorg/celestia-node/nodebuilder/p2p"
 )
 
 const DefaultAccountName = "my_celes_key"
 
-// KeyringSigner constructs a new keyring signer.
-// NOTE: we construct keyring signer before constructing node for easier UX
+// Keyring constructs a new keyring.
+// NOTE: we construct keyring before constructing node for easier UX
 // as having keyring-backend set to `file` prompts user for password.
-func KeyringSigner(cfg Config, ks keystore.Keystore, net p2p.Network) (*apptypes.KeyringSigner, error) {
+func Keyring(cfg Config, ks keystore.Keystore) (kr.Keyring, string, error) {
 	ring := ks.Keyring()
+	if ring == nil {
+		panic("keyring is nil")
+	}
 	var info *kr.Record
 	// if custom keyringAccName provided, find key for that name
 	if cfg.KeyringAccName != "" {
 		keyInfo, err := ring.Key(cfg.KeyringAccName)
 		if err != nil {
 			log.Errorw("failed to find key by given name", "keyring.accname", cfg.KeyringAccName)
-			return nil, err
+			return nil, "", err
 		}
 		info = keyInfo
 	} else {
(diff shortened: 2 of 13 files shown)


renaynay:
There are a lot more relevant issues that this PR would close.

vgonkivs:
> TODO: we should probably introduce some gas multiplier (default: 1.1) to the EstimateGas as it seems to underestimate the value (cc @vgonkivs)

Agree, bc it seems that neither EstimateGas nor Simulate don't include tx options in the calculation and we are running out of gas. Should this multiplier be dynamically changed depending on the amount of options?
I'm also wondering, should the node be aware of this multiplier or it should be encapsulated in the EstimateGas?

cmwaters:
> Should this multiplier be dynamically changed depending on the amount of options?

I think it should be fixed. We should have a sane conservative default (like 1.2x) and add a field in the relevant config for users to change it if they need to. Fees are so cheap I don't think people will mind slightly overestimating the gas in exchange for better UX

renaynay:
After further discussion, @cmwaters and I agreed that it might make more sense to handle for the case that node is started with an unfunded account inside of node codebase rather than inside the signer as the signer is just a wrapper for writes to state which cannot happen when the account is not funded.

renaynay:
@TODO @renaynay @cmwaters: 

Make signer configurable to not need to wait for inclusion of blobs to return (general config for the signer)

renaynay:
Waiting on response regarding defaults for state module blob submission

renaynay:
> TODO: we should probably introduce some gas multiplier (default: 1.1) to the EstimateGas as it seems to underestimate the value (cc @vgonkivs)

Let's do this in a later PR

cmwaters (commented):
This looks good to me @renaynay 

I will cut v1.8.0 now

UPDATE: v1.8.0 has been released

state/core_access.go
```diff
@@ -157,6 +175,11 @@ func (ca *CoreAccessor) Start(ctx context.Context) error {
 	}
 	ca.rpcCli = cli
 
+	err = ca.setupSigner(ctx)
+	if err != nil {
+		log.Warnw("failed to set up signer, check if node's account is funded", "err", err)
+	}
```
Wondertan, line 181: [blocking]
Why isn't this a hard error?

I can see that `getSigner` can also construct signer lazily, but:
* It's racy - there is no synchronization around the signer struct
* There is no obvious reason to attempt to construct a Signer on every TX if signer construction fails on Start.

state/core_access.go
```diff
@@ -461,21 +469,26 @@ func (ca *CoreAccessor) BeginRedelegate(
...
-	signedTx, err := ca.constructSignedTx(ctx, msg, apptypes.SetGasLimit(gasLim), withFee(fee))
-	if err != nil {
-		return nil, err
+	msg := stakingtypes.NewMsgBeginRedelegate(signer.Address(), srcValAddr, dstValAddr, coins)
+	if gasLim == 0 {
+		var err error
+		gasLim, err = signer.EstimateGas(ctx, []sdktypes.Msg{msg})
+		if err != nil {
+			return nil, fmt.Errorf("estimating gas: %w", err)
+		}
```
Wondertan, line 488: Note for future refactoring, I guess. The gas estimation is very repetitive and something the Signer should do automatically rather than the user before sending every transaction.
cc @cmwaters

Wondertan (changes requested):
Is there anything that `SubmitTx` could benefit from using the Signer? Like retries? 
cc @cmwaters

state/core_access.go
cmwaters: I guess we're still iterating on the boundary line between what is the responsibility of the signer and the responsibility of signers caller

state/core_access.go
renaynay: @Wondertan yes it's a bit racey so I can put a mutex there for additional safety but the reason I decided to go for the lazy construction is because it should be possible for a node runner to run a node first, get the address of the new key that's generated, and then while the node is still running, go fund the account, then use state module once the account is funded. 

Signer is unable to be constructed before the account "exists" which is why I opted for lazy construction.

Wondertan: Ok, would be great to have this nuance on the pr description.

renaynay: https://github.com/celestiaorg/celestia-node/pull/3298/commits/24b8b2ff28b4eed8eac9fb3539cc116562a30803

state/core_access.go
renaynay: Pls put in issue or resolve @cmwaters @Wondertan

state/core_access.go
Wondertan: Let's add the comment, though. It's a detail that is easy to miss, and the lazy construction should be justified for anyone looking at the code in future.

nodebuilder/state/keyring.go
```diff
@@ -3,42 +3,36 @@ package state
...
-
 	"github.com/celestiaorg/celestia-node/libs/keystore"
-	"github.com/celestiaorg/celestia-node/nodebuilder/p2p"
 )
 
 const DefaultAccountName = "my_celes_key"
 
-// KeyringSigner constructs a new keyring signer.
-// NOTE: we construct keyring signer before constructing node for easier UX
+type AccName string
```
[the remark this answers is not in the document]
Wondertan: This needs to be resolved

state/core_access.go
renaynay: done

state/core_access.go
cmwaters: @Wondertan let me know if you want gas estimation to be on the signers side

Wondertan: I think it makes sense as it abstracts aways a detail that users have to do every time. If you agree with this, let's do it or park it as an app issue

nodebuilder/state/keyring.go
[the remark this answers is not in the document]
renaynay: https://github.com/celestiaorg/celestia-node/pull/3298/commits/3e1a15bed203112c014d4f7f1fef42c89d8f7e07

vgonkivs (approved):

go.mod
```diff
@@ -8,7 +8,7 @@ require (
 	github.com/BurntSushi/toml v1.3.2
 	github.com/alecthomas/jsonschema v0.0.0-20220216202328-9eeeec9d044b
 	github.com/benbjohnson/clock v1.3.5
-	github.com/celestiaorg/celestia-app v1.7.0
+	github.com/celestiaorg/celestia-app v1.8.0-rc0
```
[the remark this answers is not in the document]
renaynay: Yeah will update the dep shortly in a separate PR as part of v0.14.0-rc0 for us.

Issue #3256, opened 2024-03-13
[Feature Request]: Signer support for fee grant module for blob submission


Bidon15:
### Implementation ideas

ATM, we need something like a `--fee-account $account_address` for node's part. This is needed for blob submission

Bidon15:
https://gist.github.com/rootulp/c47606e177aad1f8959ea31fa077ca5e

Bidon15:
https://github.com/celestiaorg/celestia-app/issues/3221

Issue #3284, opened 2024-04-01
feegrant/state: Light node's celestia account can't pass fees to the granter


Bidon15:
### Celestia Node version

v0.13.1

### OS

linux/mac

### Steps to reproduce it

Go through this tutorial https://github.com/celestiaorg/docs/pull/1500

### Expected result

Fees are charged from the granter account

### Actual result

The light node's account is still paying for fees itself

jcstein:
`ERROR` in logs of successful feegrant transaction and revoke. [see comments](https://github.com/celestiaorg/docs/pull/1500/#issuecomment-2065551407)

ramin:
@vgonkivs is this still relevent given the work completed for #3256 ?

vgonkivs:
All changes were done in https://github.com/celestiaorg/celestia-node/issues/3256

This one can be closed.

PR #3304, opened 2024-04-08
feat(state): enable fee granting

vgonkivs:
Resolves https://github.com/celestiaorg/celestia-node/issues/3256

Tested on mocha:
1) Run funded light node and grant unlimited fee to the newly created FN:
`celestia state grant-fee celestia12psg90lwvxdktnqerr2p5mxuw2v3je497uv6tj 2000 1000000 --node.store ~/.celestia-light-mocha-4/`

2) Run FN and submitted blob:
`celestia blob submit 0x42690c204d39600fddd3 'gm' --node.store ~/.celestia-full-mocha-4/`
Result: blob was successfully submitted:
```
{
  "result": {
    "height": 1562820,
    "commitments": [
      "IXg+08HV5RsPF3Lle8PH+B2TUGsGUsBiseflxh6wB5E="
    ]
  }
}
```

3) Run revoke-fee cmd on the LN:
`celestia state revoke-grant-fee celestia12psg90lwvxdktnqerr2p5mxuw2v3je497uv6tj 2000 1000000 --node.store ~/.celestia-light-mocha-4/`

4) Submitted blob from the FN and got an error back:
```
{
  "result": ": insufficient funds"
}
```
5) Run grant fee with a very small limit:
`celestia state grant-fee celestia12psg90lwvxdktnqerr2p5mxuw2v3je497uv6tj 2000 1000000 --amount 10 --node.store ~/.celestia-light-mocha-4/`

6) Submitting blobs now returns an error:
```
{
  "result": ": fee limit exceeded"
}
```

### DIFF
--- a/nodebuilder/state/cmd/state.go
+++ b/nodebuilder/state/cmd/state.go
@@ -12,6 +12,10 @@ import (
 	"github.com/celestiaorg/celestia-node/state"
 )
 
+var (
+	amount uint64
+)
+
 func init() {
 	Cmd.AddCommand(
 		accountAddressCmd,
@@ -26,6 +30,16 @@ func init() {
 		queryDelegationCmd,
 		queryUnbondingCmd,
 		queryRedelegationCmd,
+		grantFeeCmd,
+		revokeGrantFeeCmd,
+	)
+
+	grantFeeCmd.PersistentFlags().Uint64Var(
+		&amount,
+		"amount",
+		0,
+		"specifies the spend limit(in utia) for the grantee.\n"+
+			"The default value is 0 which means the grantee does not have a spend limit.",
 	)
 }
 
@@ -399,6 +413,75 @@ var queryRedelegationCmd = &cobra.Command{
 			dstAddr.Address.(state.ValAddress),
 		)
 		return cmdnode.PrintOutput(response, err, nil)
+	},
+}
+
+var grantFeeCmd = &cobra.Command{
+	Use: "grant-fee [grantee] [fee] [gasLimit]",
+	Short: "Adds permission for Grantee to spend up to Allowance of fees from the account of Granter.\n" +
+		"Grantee could spend any amount of tokens in case if spend limit is not set.",
+	Args: cobra.ExactArgs(3),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		client, err := cmdnode.ParseClientFromCtx(cmd.Context())
+		if err != nil {
+			return err
+		}
+		defer client.Close()
+
+		granteeAddr, err := parseAddressFromString(args[0])
+		if err != nil {
+			return fmt.Errorf("error parsing an address:%v", err)
+		}
+
+		fee, err := strconv.ParseInt(args[1], 10, 64)
+		if err != nil {
+			return fmt.Errorf("error parsing a fee:%v", err)
+		}
+		gasLimit, err := strconv.ParseUint(args[2], 10, 64)
+		if err != nil {
+			return fmt.Errorf("error parsing a gas limit:%v", err)
+		}
+
+		txResponse, err := client.State.GrantFee(
+			cmd.Context(),
+			granteeAddr.Address.(state.AccAddress),
+			math.NewInt(int64(amount)), math.NewInt(fee), gasLimit,
+		)
+		return cmdnode.PrintOutput(txResponse, err, nil)
+	},
+}
+
+var revokeGrantFeeCmd = &cobra.Command{
+	Use:   "revoke-grant-fee [grantee] [fee] [gasLimit]",
+	Short: "Adds permission for Grantee to spend up to Allowance of fees from the account of Granter.",
+	Args:  cobra.ExactArgs(3),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		client, err := cmdnode.ParseClientFromCtx(cmd.Context())
+		if err != nil {
+			return err
+		}
+		defer client.Close()
+
+		granteeAddr, err := parseAddressFromString(args[0])
+		if err != nil {
+			return fmt.Errorf("error parsing an address:%v", err)
+		}
+
+		fee, err := strconv.ParseInt(args[1], 10, 64)
+		if err != nil {
+			return fmt.Errorf("error parsing a fee:%v", err)
+		}
+		gasLimit, err := strconv.ParseUint(args[2], 10, 64)
+		if err != nil {
+			return fmt.Errorf("error parsing a gas limit:%v", err)
+		}
+
+		txResponse, err := client.State.RevokeGrantFee(
+			cmd.Context(),
+			granteeAddr.Address.(state.AccAddress),
+			math.NewInt(fee), gasLimit,
+		)
+		return cmdnode.PrintOutput(txResponse, err, nil)
 	},
 }
 
--- a/nodebuilder/state/cmd/state.go
+++ b/nodebuilder/state/cmd/state.go
@@ -418,7 +418,7 @@ var queryRedelegationCmd = &cobra.Command{
 
 var grantFeeCmd = &cobra.Command{
 	Use: "grant-fee [granteeAddress] [fee] [gasLimit]",
-	Short: "Grant an allowance to a specified grantee account to pay the fees for their SubmitPFB transactions.\n" +
+	Short: "Grant an allowance to a specified grantee account to pay the fees for their transactions.\n" +
 		"Grantee can spend any amount of tokens in case the spend limit is not set.",
 	Args: cobra.ExactArgs(3),
 	RunE: func(cmd *cobra.Command, args []string) error {
--- a/nodebuilder/state/cmd/state.go
+++ b/nodebuilder/state/cmd/state.go
@@ -417,9 +417,9 @@ var queryRedelegationCmd = &cobra.Command{
 }
 
 var grantFeeCmd = &cobra.Command{
-	Use: "grant-fee [grantee] [fee] [gasLimit]",
-	Short: "Grant an allowance to a specified grantee account to pay the fees for their transactions.\n" +
-		"Grantee could spend any amount of tokens in case if spend limit is not set.",
+	Use: "grant-fee [granteeAddress] [fee] [gasLimit]",
+	Short: "Grant an allowance to a specified grantee account to pay the fees for their SubmitPFB transactions.\n" +
+		"Grantee can spend any amount of tokens in case the spend limit is not set.",
 	Args: cobra.ExactArgs(3),
 	RunE: func(cmd *cobra.Command, args []string) error {
 		client, err := cmdnode.ParseClientFromCtx(cmd.Context())
@@ -452,8 +452,8 @@ var grantFeeCmd = &cobra.Command{
 }
 
 var revokeGrantFeeCmd = &cobra.Command{
-	Use:   "revoke-grant-fee [grantee] [fee] [gasLimit]",
-	Short: "Adds permission for Grantee to spend up to Allowance of fees from the account of Granter.",
+	Use:   "revoke-grant-fee [granteeAddress] [fee] [gasLimit]",
+	Short: "Removes permission for grantee to submit PFB transactions which will be paid by granter.",
 	Args:  cobra.ExactArgs(3),
 	RunE: func(cmd *cobra.Command, args []string) error {
 		client, err := cmdnode.ParseClientFromCtx(cmd.Context())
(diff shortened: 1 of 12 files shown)


Bidon15:
Can you try this approach and see if this works? 🙏 
https://github.com/celestiaorg/celestia-node/issues/3284

vgonkivs:
Hey @Bidon15 

Please find all info below:

```
celestia-appd tx feegrant grant \
  celestia1ly7vmh26hffvau52ktnxff557t4u8thh0gwtyc celestia12psg90lwvxdktnqerr2p5mxuw2v3je497uv6tj \
   --home ~/.celestia-light-mocha-4/keys \
   --node public-celestia-mocha4-consensus.numia.xyz:26657 \
   --spend-limit 1000000utia \
   --allowed-messages "/cosmos.bank.v1beta1.MsgSend,/celestia.blob.v1.MsgPayForBlobs" \
   --chain-id mocha-4 \
   --keyring-backend test \
   --fees 20000utia \
   --broadcast-mode block \
   --yes
```
https://mocha-4.celenium.io/tx/4a317410cf3a0ccc86169a1774d796fc3f37a47f109f7a00664ea155fcc86e9b

https://mocha-4.celenium.io/tx/165865c66fa785bae7b9e73b21abe20b1c0cff97ad1a80dbde0b3a6ee29f4003

Снимок экрана 2024-04-09 в 17 07 34

nodebuilder/state/cmd/state.go
```diff
@@ -402,6 +416,75 @@ var queryRedelegationCmd = &cobra.Command{
 	},
 }
 
+var grantFeeCmd = &cobra.Command{
+	Use: "grant-fee [grantee] [fee] [gasLimit]",
+	Short: "Adds permission for Grantee to spend up to Allowance of fees from the account of Granter.\n" +
```
Wondertan, line 421: What is Allowance and where is it defined?

Wondertan: Let's clarify that grantee expected to be an address

nodebuilder/state/cmd/state.go
```diff
@@ -402,6 +416,75 @@ var queryRedelegationCmd = &cobra.Command{
...
+		"Grantee could spend any amount of tokens in case if spend limit is not set.",
```
Wondertan, line 422:
```suggestion
		"Grantee can spend any amount of tokens in case the spend limit is not set.",
```

Wondertan: Also, spend limit is a required arguemnt, so not sure how it can't be set

state/core_access.go
```diff
@@ -51,9 +57,10 @@ type CoreAccessor struct {
 	signer *apptypes.KeyringSigner
 	getter libhead.Head[*header.ExtendedHeader]
 
-	queryCli   banktypes.QueryClient
-	stakingCli stakingtypes.QueryClient
-	rpcCli     rpcclient.ABCIClient
+	queryCli    banktypes.QueryClient
+	stakingCli  stakingtypes.QueryClient
+	feeGrantCli feegrant.QueryClient
```
Wondertan, line 62: nit: cli naming is a bit confusing, I thought of CLI while reading the code, but its a short for Client. If you want to shorten it, let's use clnt

state/core_access.go
```diff
@@ -218,6 +226,14 @@ func (ca *CoreAccessor) SubmitPayForBlob(
 
 	minGasPrice := ca.getMinGasPrice()
 
+	feeGrant, err := ca.getGranter(ctx)
+	if err != nil {
+		log.Warn(err)
+	}
```
Wondertan, line 232: This adds "yet another roundtrip" to the tx submission. Is there a way to persisting (and pre-loading) that info locally instead of quering remote address?

cc @vgonkivs and @cmwaters

Wondertan: Also, shouldn't feeGranter be supported for all tx submissions and not just blobs?

cmwaters: Yeah I agree, if a user knows they have a grant then maybe they can simply initiate the client with the granter address (i.e. in the constructor) and we do one query at the start to see what messages have grants and even if it exists and then use that address for every message

cmwaters (commented):
Logically, these changes are sound and I only have a few adjacent comments.

While I was reviewing this, I was thinking about the broader picture. Will node API eventually support all transactions that can be submitted to the Celestia state machine or only a subset that is specifically relevant to rollup nodes? If we're taking a minimalist stance I don't see why the `grant-fee` and `revoke-grant-fee` cli needs to be supported. These already exist in the app cli so do we really need to duplicate it here.

My naive implementation of this would have been to added a field feeGranter in the coreAccessor struct that can be set in the constructor and then simply in every `SubmitPayForBlob`, check if it is set and then set it in the transaction that goes to the `Signer` - that way light nodes are now compliant with the fee granter protocol

Wondertan:
> Will node API eventually support all transactions that can be submitted to the Celestia state machine?

I would say so, considering merge and API unification efforts

cmwaters:
> Thoughts or objections?

I like the general principle of it doesn't matter which node you talk to, they all have the same API

state/core_access.go
vgonkivs: How do users know when the granter revokes the grant in this case? I guess we will anyway need to do a roundtrip.

nodebuilder/state/cmd/state.go
```diff
@@ -26,6 +30,16 @@ func init() {
...
+		revokeGrantFeeCmd,
+	)
+
+	grantFeeCmd.PersistentFlags().Uint64Var(
+		&amount,
+		"amount",
+		0,
+		"specifies the spend limit(in utia) for the grantee.\n"+
+			"The default value is 0 which means the grantee does not have a spend limit.",
 	)
```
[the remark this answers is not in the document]
vgonkivs: As we agreed, only SubmitPFBs can be paid.

state/core_access.go
Wondertan: Through an error. You can do it once you start the node, and if, during runtime, the granter revokes - the transaction will error. This works, and we can recommend that users restart the node to fix the error. Alternatively, we can spend some time making custom error-handling logic for this case that removes granting.

nodebuilder/state/flags.go
```diff
@@ -21,6 +23,8 @@ func Flags() *flag.FlagSet {
 	flags.String(keyringBackendFlag, defaultKeyringBackend, fmt.Sprintf("Directs node's keyring signer to use the given "+
 		"backend. Default is %s.", defaultKeyringBackend))
 
+	flags.String(granterAddressFlag, "", "External node's address that will pay for all transactions, submitted by "+
+		"the local node.")
```
renaynay, line 27:
```suggestion
	flags.String(granterAddressFlag, "", "Account address that will pay for all transactions submitted from the node.")
```

nodebuilder/state/module.go
```diff
@@ -24,9 +24,9 @@ func ConstructModule(tp node.Type, cfg *Config, coreCfg *core.Config) fx.Option
 
 	baseComponents := fx.Options(
 		fx.Supply(*cfg),
+		fx.Supply(cfg.GranterAddress),
```
renaynay, line 27: Isn't this just a string? How does this supply work? Can't you just read from the cfg wherever you need it?

state/core_access.go
```diff
@@ -214,6 +238,13 @@ func (ca *CoreAccessor) SubmitPayForBlob(
 
 	minGasPrice := ca.getMinGasPrice()
 
+	var feeGrant apptypes.TxBuilderOption
+
+	// set granter and update gasLimit in case node run in a grantee mode
```
renaynay, line 243: Why do we need the gas multiplier here?

state/core_access.go
```diff
@@ -223,14 +254,17 @@ func (ca *CoreAccessor) SubmitPayForBlob(
 
 	var lastErr error
 	for attempt := 0; attempt < maxRetries; attempt++ {
+		options := []apptypes.TxBuilderOption{apptypes.SetGasLimit(gasLim), withFee(fee)}
```
renaynay, line 257: maybe you can define the options a bit higher up and just do the append to opts inline instead of checking `feeGrant != nil`, 

a bit cleaner

state/core_access.go
```diff
@@ -256,6 +290,10 @@ func (ca *CoreAccessor) SubmitPayForBlob(
 		if response != nil && response.Code != 0 {
 			err = errors.Join(err, sdkErrors.ABCIError(response.Codespace, response.Code, response.Logs.String()))
 		}
+
+		if err != nil && errors.Is(err, sdkerrors.ErrNotFound) && !ca.granter.Empty() {
+			return response, errors.New("granter has revoked the grant")
```
renaynay, line 295: I wouldn't do a super complicated err check for this unless the error returned from app is *so* unclear that we need to do this check

state/core_access.go
```diff
@@ -81,19 +94,29 @@ func NewCoreAccessor(
 	coreIP,
 	rpcPort string,
 	grpcPort string,
-) *CoreAccessor {
+	granter string,
```
renaynay, line 97: Nit but I think this would've made more sense as an ...Option bc we'll end up breaking constructor more and more the more configurability we add, so we may as well do it now to add opts?

state/core_access.go
vgonkivs: we are getting back "not found" which is very confusing.

state/core_access.go
vgonkivs: gasLimit and fee options could be recalculated, so we will need to clear the slice(besides grantFee opt) on every attempt. That's why I decided to create a new slice

state/core_access.go
vgonkivs: gasLimit for tx with options is higher the without options, but app does not include in the calculations the options cost. So me and @cmwaters agreed to use a multiplier as a quick solution.

nodebuilder/state/module.go
vgonkivs: Yeah, it's a string and It works 🙂 . We don't pass a state cfg to the core accessor, so I can't do it directly. But, as you suggested, I will use an option pattern

state/core_access.go
vgonkivs: added options

nodebuilder/state/flags.go
```diff
@@ -36,5 +37,13 @@ func ParseFlags(cmd *cobra.Command, cfg *Config) {
 	}
 
 	cfg.KeyringBackend = cmd.Flag(keyringBackendFlag).Value.String()
-	cfg.GranterAddress = cmd.Flag(granterAddressFlag).Value.String()
+
+	addr := cmd.Flag(granterAddressFlag).Value.String()
+	if addr != "" {
+		sdkAddress, err := sdktypes.AccAddressFromBech32(addr)
+		if err != nil {
+			panic(err)
```
renaynay, line 44: We return error in parse flags instead of panic for other modules

nodebuilder/state/module.go
```diff
@@ -26,7 +30,14 @@ func ConstructModule(tp node.Type, cfg *Config, coreCfg *core.Config) fx.Option
 		fx.Supply(*cfg),
 		fx.Supply(cfg.GranterAddress),
 		fx.Error(cfgErr),
-		fxutil.ProvideIf(coreCfg.IsEndpointConfigured(), fx.Annotate(coreAccessor,
+		fxutil.ProvideIf(coreCfg.IsEndpointConfigured(), fx.Annotate(
```
renaynay, line 33: Fx cleanliness nit: you can also just provide a state.Option in case cfg.GranterAddr != nil and take []state.Option as param in module's core access constructor

nodebuilder/state/cmd/state.go
```diff
@@ -402,6 +416,75 @@ var queryRedelegationCmd = &cobra.Command{
 	},
 }
 
+var grantFeeCmd = &cobra.Command{
+	Use: "grant-fee [granteeAddress] [fee] [gasLimit]",
+	Short: "Grant an allowance to a specified grantee account to pay the fees for their transactions.\n" +
+		"Grantee can spend any amount of tokens in case the spend limit is not set.",
```
Wondertan, line 422: nit: for multi-line you could use ``

nodebuilder/state/cmd/state.go
```diff
@@ -402,6 +416,75 @@ var queryRedelegationCmd = &cobra.Command{
 	},
 }
 
+var grantFeeCmd = &cobra.Command{
+	Use: "grant-fee [granteeAddress] [fee] [gasLimit]",
+	Short: "Grant an allowance to a specified grantee account to pay the fees for their transactions.\n" +
```
Wondertan, line 421: We, by default, set allowance for all transactions, and it's not clear precisely what the term means, so I would remove allowance from the description.

Wondertan (approved):
LGTM.

nodebuilder/state/cmd/state.go
vgonkivs: Grant allowance == give a permission. So, I think we can leave it as it is for now

PR #3305, opened 2024-04-10
feat(share/eds): Store nmt nodes on failed reconstruction

walldiss:
Store repaired shares and calculated nmt proofs if reconstruction failed.

### DIFF
--- a/share/eds/retriever.go
+++ b/share/eds/retriever.go
@@ -128,11 +128,7 @@ type retrievalSession struct {
 func (r *Retriever) newSession(ctx context.Context, dah *da.DataAvailabilityHeader) (*retrievalSession, error) {
 	size := len(dah.RowRoots)
 
-	adder := ipld.NewNmtNodeAdder(
-		ctx,
-		r.bServ,
-		ipld.MaxSizeBatchOption(size),
-	)
+	adder := ipld.NewNmtNodeAdder(ctx, r.bServ, ipld.MaxSizeBatchOption(size))
 	proofsVisitor := ipld.ProofsAdderFromCtx(ctx).VisitFn()
 	visitor := func(hash []byte, children ...[]byte) {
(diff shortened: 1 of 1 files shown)


Wondertan (dismissed):
Looks correct. I feel like this needs a test, which will stay for after shwap times.

Wondertan (dismissed):

share/eds/retriever.go
```diff
@@ -123,13 +128,24 @@ type retrievalSession struct {
...
-		visitor := ipld.ProofsAdderFromCtx(ctx).VisitFn()
-		if visitor != nil {
-			opts = append(opts, nmt.NodeVisitor(visitor))
+
+		proofsAdder := ipld.ProofsAdderFromCtx(ctx)
+		visitor := func(hash []byte, children ...[]byte) {
+			if proofsAdder != nil {
+				proofsAdder.VisitFn()(hash, children...)
+			}
+			adder.Visit(hash, children...)
```
renaynay, line 146: so in case of proofsAdder in ctx, both retrieval session's adder + proofs adder from ctx will cache collected proofs ?

renaynay (approved):

share/eds/retriever.go
walldiss: yes. Not a big deal, because soon, when shwap is released, we would not need proofsAdder here

renaynay (approved):
Agree with hlib that this needs a test case -- I think we should make an issue or document some test cases somewhere that we need to implement.

walldiss:
> Agree with hlib that this needs a test case -- I think we should make an issue or document some test cases somewhere that we need to implement.

https://github.com/celestiaorg/celestia-node/issues/3314

Wondertan (approved):

vgonkivs (approved):

PR #3306, opened 2024-04-10
fix!(share/byzantine): use any available axis for befp nmt proofs construction

walldiss:
We should not restrict the BEFP constructor to collect proof only for orthogonal axes to ErrByzantine. This PR enables the constructor to attempt building both Row and Col proofs, irrespective of the ErrByzantine axis type. Additionally, it prevents the BEFP constructor from requesting proofs pr shars from the network by granting access exclusively to the local blockstore. There should be sufficient proofs and shares in the local blockstore at the time BEFP is detected.

Breaks befp message, by introducing proofAxis field, which is not supported by older befp subscribers.

Depends on 
- https://github.com/celestiaorg/celestia-node/pull/3305 to have enough data in blockstore.
- https://github.com/celestiaorg/rsmt2d/pull/310 to provide proper coordinates of verified shares


share/eds/byzantine/bad_encoding.go
```diff
@@ -184,19 +177,25 @@ func (p *BadEncodingProof) Validate(hdr *header.ExtendedHeader) error {
...
-		if ok := shr.Validate(ipld.MustCidFromNamespacedSha256(merkleRoots[index])); !ok {
+		var x, y int
+		if p.Axis == rsmt2d.Row {
+			x, y = index, int(p.Index)
+		} else {
+			x, y = int(p.Index), index
+		}
+		root := getRoot(hdr.DAH, shr.Axis, x, y)
+		rootCid := ipld.MustCidFromNamespacedSha256(root)
+		if ok := shr.Validate(rootCid, x, y, width); !ok {
```
Wondertan, line 194: I propose to pass the DAH to Validate and do this logic there instead.

share/eds/byzantine/byzantine.go
```diff
@@ -31,40 +31,47 @@ func (e *ErrByzantine) Error() string {
...
 
-		index := index
+		var x, y int
+		if errByz.Axis == rsmt2d.Row {
+			x = index
+			y = int(errByz.Index)
+		} else {
+			x = int(errByz.Index)
+			y = index
+		}
```
Wondertan, line 58: Similarly here. Feels like it should be coupled with `getSharesWithProof`

share/eds/byzantine/share_proof.go
```diff
@@ -76,18 +70,66 @@ func GetProofsForShares(
...
+	proof, err := getProofsAt(ctx, bGetter, rootCid, x, width)
+	if err == nil {
+		shareWithProof := &ShareWithProof{
+			Share: share,
+			Proof: &proof,
+			Axis:  rsmt2d.Row,
+		}
+		if shareWithProof.Validate(rootCid, x, y, width) {
+			return shareWithProof, nil
+		}
```
Wondertan, line 114: I think we trust our own local proofs, so feels unnecessary.

walldiss: This proof might contain bad encoded data which will make it invalid. So in this case we need orthogonal proof instead

Wondertan: From my understanding proof js already valid here as we collected it from local proof data with got and validated before

walldiss: We collected it from traversing the tree from top to the bottom. Validation goes up by recomputing hashes and validating namespacing rules. Collected proof might fail validation for bad encoded data here

share/eds/byzantine/bad_encoding.go
walldiss: Coordintes are defined by befp axis, not by proof axis. Those are not relateed. Refactored whole Validate thing to work over axisType,axisIdx,shrIdx instead of using x,y. Let me know how you like it

share/eds/byzantine/share_proof.go
```diff
@@ -96,20 +138,20 @@ func getProofsAt(
 	root cid.Cid,
 	index,
 	total int,
-) (*ShareWithProof, error) {
-	proof := make([]cid.Cid, 0)
-	// TODO(@vgonkivs): Combine GetLeafData and GetProof in one function as the are traversing the same
-	// tree. Add options that will control what data will be fetched.
-	node, err := ipld.GetLeaf(ctx, bGetter, root, index, total)
+) (nmt.Proof, error) {
+	proofPath := make([]cid.Cid, 0)
```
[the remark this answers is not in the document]
walldiss: added prealloc

share/eds/byzantine/share_proof.go
Wondertan: This validates the correctness of the proof of any data. This does not check for badly encoded erasure-coded data. That price part is done by rsmt2d, so this won't fail for bed erasure coding.

walldiss: Namespacing rules could have been violated.

Wondertan: They are validated earlier when we fetch proofs during reconstruction over IPLD

Wondertan: Anyway, it does not hurt to have additional validation, so let's keep it the way it is. We can justify it by disk corruption 😉

share/eds/byzantine/share_proof.go
```diff
@@ -66,28 +69,55 @@ func (s *ShareWithProof) ShareWithProofToProto() *pb.Share {
...
+		return nil, err
+	}
+	shareWithProof := &ShareWithProof{
+		Share: share,
+		Proof: &proof,
+		Axis:  rsmt2d.Col,
+	}
+	if shareWithProof.Validate(dah, axisType, axisIdx, shrIdx) {
+		return shareWithProof, nil
+	}
```
Wondertan, line 119: minor nit: we could dedup that, but that doesn't matter much

Wondertan (approved):
Great find, honestly. It should have become obvious to us after we realized that the server is responsible for selecting the proof when we sample and it did 👍🏻

vgonkivs (approved):

vgonkivs:
```
--- FAIL: TestBEFP_ValidateOutOfOrderShares (0.00s)
    bad_encoding_test.go:241: 
        	Error Trace:	/home/runner/work/celestia-node/celestia-node/share/eds/byzantine/bad_encoding_test.go:241
        	            	incorrect share received
        	Test:       	TestBEFP_ValidateOutOfOrderShares
FAIL
coverage: 13.2% of statements
```

walldiss:
I had to remove concurrent fetching of ShareWithProof code to resolve flakiness of unit tests. Also looks much simpler now. Performance downside is not important, as we collect shares from disk and it is still fast enough being sequential instead of concurrent.

Wondertan (approved):

vgonkivs (approved):

share/eds/byzantine/share_proof.go
```diff
@@ -24,31 +27,31 @@ type ShareWithProof struct {
...
-		node := ipld.NamespacedSha256FromCID(pathToLeaf[i])
-		rangeProofs = append(rangeProofs, node)
+// Validate validates inclusion of the share under the given root CID.
+func (s *ShareWithProof) Validate(dah *share.Root, axisType rsmt2d.Axis, axisIdx, shrIdx int) bool {
+	var rootHash []byte
+	switch axisType {
+	case rsmt2d.Row:
+		rootHash = rootHashForCoordinates(dah, s.Axis, shrIdx, axisIdx)
+	case rsmt2d.Col:
+		rootHash = rootHashForCoordinates(dah, s.Axis, axisIdx, shrIdx)
```
oblique, line 41: I'm evaluating your changes to port them into Rust's `celestia-types`. Correct me if I'm wrong, but isn't this the same as the following?

```go
	switch axisType {
	case rsmt2d.Row:
		rootHash = dah.RowRoots[axisIdx]
	case rsmt2d.Col:
		rootHash = dah.ColumnRoots[axisIdx]
```

oblique, line 41: I'm evaluating your changes to port them into Rust's `celestia-types`.

Correct me if I'm wrong but isn't this the same as the following?

```
	switch s.Axis {
	case rsmt2d.Row:
		rootHash = dah.RowRoots[axisIdx]
	case rsmt2d.Col:
		rootHash = dah.ColumnRoots[axisIdx]
```

PR #3318, opened 2024-04-18
feat(cmd): match upcoming node version with API

vgonkivs:
Resolves https://github.com/celestiaorg/celestia-node/issues/2549

`docgen` cmd moved to the root celestia cmd folder, so it will be a part of the c-node binary to get the correct Semantic version.


ramin (approved):

renaynay (approved):

Wondertan (approved):

jcstein:
will this PR make it possible to generate the spec json file on releases? to be stored as artifacts on the release as suggested in https://github.com/celestiaorg/celestia-node/issues/2611?

jcstein:
> Снимок экрана 2024-04-18 в 17 42 48

it looks like this just outputs the version info? but not the actual api specs? am i missing something here?

jcstein (changes requested):
now it is not possible to generate the api specs?

```bash
make openrpc-gen
--> Generating OpenRPC spec
stat /Users/joshstein/celestia-node/cmd/docgen: directory not found
make: *** [openrpc-gen] Error 1
```

however, the version (and no actual api specs) show when you run:

```bash
$ celestia docgen
{
    "openrpc": "1.2.6",
    "info": {
        "title": "Celestia Node API",
        "description": "The Celestia Node API is the collection of RPC methods that can be used to interact with the services provided by Celestia Data Availability Nodes.",
        "version": "v0.13.2-13-gc57e39bc"
    },
    "externalDocs": {
        "description": "Celestia Node GitHub",
        "url": "https://github.com/celestiaorg/celestia-node"
    },
    "methods": []
}%               
```

i am quite confused at what this PR accomplishes

jcstein:
flagging that I do not think this PR should be merged. it 
(1) makes it impossible to generate rpc docs, which is counterproductive
(2) introduces `celestia docgen` which outputs the description of the docs with the api version, which is sort of productive, except for all the missing api docs that should come along with this spec generation

vgonkivs:
Let me fix the spec generation. Sorry, did not test it properly

Wondertan (dismissed):
utACK

jcstein:
has this been tested?

vgonkivs:
Yeap. I have tested it and it worked for me

jcstein:
it works for me. but why do we need 2 commands to generate the spec? shouldn't `celestia docgen` handle this for us?

also, [in the output I get when testing `make openrpc-gen`](https://gist.github.com/jcstein/6a7e04bbe760bd5c386f8159c43213af), there is still `--> Generating OpenRPC spec` which will need to be filtered out to handle #2611

jcstein:
I understand that. I'm trying to understand if it helps accomplish that goal, or of it is going to add more work, that is all

jcstein (approved):
LGTM, also especially if it also helps to accomplish https://github.com/celestiaorg/celestia-node/issues/2611

jcstein:
my only feedback would be that the `celestia docgen` command is a little misleading, as it doesn't actually generate the docs. the command to do that is `make openrpc-gen` which AFAIU uses `celestia docgen`?

vgonkivs:
> my only feedback would be that the celestia docgen command is a little misleading, as it doesn't actually generate the docs. the command to do that is make openrpc-gen which AFAIU uses celestia docgen?

`make openrpc-gen` calls docgen under the hood, passing all available modules. so to make it work, you should call
`celestia docgen da fraud share` etc....

But I'd remove the parameters and generate docs for all modules. 

wdyt @jcstein @distractedm1nd , @renaynay, @Wondertan

distractedm1nd:
Tested, does not work on my machine?
```
celestia-node on  HEAD (b886feb) [$] via 🐹 v1.22.0 took 5s
➜ ./build/celestia docgen
[1]    64481 killed     ./build/celestia docgen
```

ramin (approved):

distractedm1nd (approved):

renaynay (approved):
Utack

walldiss (approved):

PR #3325, opened 2024-04-20
chore: use errors.New to replace fmt.Errorf with no parameters

ChengenH:
If you don't need to format the string, recommended to use error.New().


Wondertan:
iirc, there are equivalent in these cases
Please link to the recommendation of one over another

PR #3329, opened 2024-04-21
refactor: refactor core/listener/start to accept context parameter

minhngoc274:
Updated the Start function in the core/listener module to accept a context parameter. This change enhances flexibility and improves context management


walldiss:
Thank you for contributing attempt, but the change will introduce issue rather than solving any. Context provided to Start call is temporary context for startup and should not be used for lifelong context of module.

Issue #3322, opened 2024-04-19
bug: successful fee-grant and revoke-fee-grant show `ERROR` in logs and `  "result": "RPC client error: sendRequest failed: http status 200 OK unmarshaling response: EOF"` in CLI


jcstein:
### Celestia Node version

v0.13.2-17-g6382dc13 

### OS

mac

### Install tools

building on main at a point after https://github.com/celestiaorg/celestia-node/pull/3304, at `6382dc13b92d7b40077154aab823367500371cb8`

### Steps to reproduce it

make feegrant or revoke feegrant

### Expected result

no `ERROR` in logs, because the transaction was successful

### Actual result

`ERROR` in logs of celestia-node on successful fee-grant and in revoke-fee-grant and `RPC client error` in CLI, although `http status 200 OK`

1. [see comment with logs for successful grant-fee tx](https://github.com/celestiaorg/docs/pull/1500/#issuecomment-2065551407)
2. [see comment with logs for successful revoke-grant-fee tx](https://github.com/celestiaorg/docs/pull/1500/#issuecomment-2065556862)

liamsi:
I can confirm this is not working and this mainly seems to be a serialization issue:

```
[...]

PR #3336, opened 2024-04-23
fix(state/coreAccessor): fix txResponse

vgonkivs:
Resolves https://github.com/celestiaorg/celestia-node/issues/3322

Marshaling `TxResponse` fails because `TxResponse.Tx` is not empty but does not contain respective codec for encoding using the standard `json.MarshalJSON()`


state/core_access.go
```diff
@@ -699,3 +705,17 @@ func withFee(fee Int) user.TxOption {
...
+// THIS IS A TEMPORARY SOLUTION!!!
+// unsetTx helps to fix issue in TxResponse marshaling. Marshaling TxReponse
+// fails because `TxResponse.Tx` is not empty but does not contain respective codec
+// for encoding using the standard `json.MarshalJSON()`. It becomes empty when
+// https://github.com/celestiaorg/celestia-core/issues/1281 will be merged.
+// The `TxResponse.Tx` contains the transaction that is sent to the cosmos-sdk in the form
+// in which it is processed there, so the user should not be aware of it.
+func unsetTx(txResponse *TxResponse) *TxResponse {
+	if txResponse != nil && txResponse.Tx != nil {
+		txResponse.Tx = nil
```
liamsi, line 718: I would expect this to be resolvable differently. but this is a workable workaround if it's only short-lived.

vgonkivs: This is the simplest and the quickest way of resolving this issue from the node side. I also think it's not the node issue as we expect to get results that can be converted to both JSON and proto formats.

liamsi: If that is a contract between app and node (I might not be aware of that), then the app behaviour needs to change (cc @cmwaters )

state/core_access.go
```diff
@@ -624,7 +629,8 @@ func (ca *CoreAccessor) RevokeGrantFee(
 	granter := signer.Address()
 
 	msg := feegrant.NewMsgRevokeAllowance(granter, grantee)
-	return signer.SubmitTx(ctx, []sdktypes.Msg{&msg}, user.SetGasLimit(gasLim), withFee(fee))
+	resp, err := signer.SubmitTx(ctx, []sdktypes.Msg{&msg}, user.SetGasLimit(gasLim), withFee(fee))
+	return unsetTx(resp), err
```
walldiss, line 633: Could it be the case that problems comes from the fact, that error is not checked before returning, which might result in both `err` and `resp` being non-nil on return. OpenRPC has limitation to return either value or error, so it might lead to problems. Perhaps err != nil check would help?

```suggestion
	resp, err := signer.SubmitTx(ctx, []sdktypes.Msg{&msg}, user.SetGasLimit(gasLim), withFee(fee))
	if err != nil{
		return nil, err
	}	
	return resp, nil
```

jcstein (approved):

state/core_access.go
vgonkivs: But this is how it worked previously. unsetTx removes not the entire response but a single field inside.

state/core_access.go
liamsi: Can we break this out into a separate issue how to deal with this long term before merging? I also would appreciate some input from the app team there.

vgonkivs: After finishing https://github.com/celestiaorg/celestia-core/issues/1281 the Tx field won't be needed and it will be unset(or even removed) on the celestia-app side

liamsi: nice, thanks!

liamsi: One more question on this (before I would approve the PR): what is the _concrete_ type here that gets back from app (in txResponse.Tx)? I would be surprised if it did not come with a json marshaller implemented (it likely just lives in a different file which is not imported).

vgonkivs: > One more question on this (before I would approve the PR): what is the concrete type here that gets back from app (in txResponse.Tx)? I would be surprised if it did not come with a json marshaller implemented (it likely just lives in a different file which is not imported).

The `txResponse.Tx` type is [types.Any](https://github.com/celestiaorg/cosmos-sdk/blob/02de31436e7616f9215aa79404b00d17fe5c0e5a/codec/types/any.go#L11) that apparently does have default json unmarshaler as it's not set in the [c-tor](https://github.com/celestiaorg/cosmos-sdk/blob/02de31436e7616f9215aa79404b00d17fe5c0e5a/codec/types/any.go#L59). So, when it comes to the marshaling the `compat`(compatibility) field is empty, which cause an error. I have a theory that proto message should be wrapped somewhere by [ProtoJsonPacker](https://github.com/celestiaorg/cosmos-sdk/blob/02de31436e7616f9215aa79404b00d17fe5c0e5a/codec/types/compat.go#L186) but haven't found the confirmation yet.

liamsi: > The txResponse.Tx type is [types.Any](https://github.com/celestiaorg/cosmos-sdk/blob/02de31436e7616f9215aa79404b00d17fe5c0e5a/codec/types/any.go#L11) that apparently

Yeah, I thought the concrete Tx type that gets into the Any-typed field might have implemented the json marshaller directly somehow but it does not sound like it actually does.

liamsi (approved):
good as a temporary workaround

renaynay (approved):
I'm okay with this temporary fix. We need a corresponding issue that tracks the core/app issue to make sure we do not forget it in node.

jcstein:
> I'm okay with this temporary fix. We need a corresponding issue that tracks the core/app issue to make sure we do not forget it in node.

has this been made? or does it still need to be made?

PR #3349, opened 2024-04-30
feat!(state/core_access): add TxConfig

### LINKED ISSUE #3346: state | nodebuilder/state: Implement `TxOptions`
Implement `TxOptions` struct that will be applied by core accessor and use passed options.

vgonkivs:
Resolves #3346

What was done:
Implemented a new TxConfig struct
```go
// TxConfig specifies additional options that will be applied to the Tx.
type TxConfig struct {
	// Specifies the address from the keystore that will sign transactions.
	// NOTE: Only `accountAddress` or `KeyName` should be passed.
	// accountAddress is a primary options. This means If both the address and the key are specified,
	// the address field will take priority.
	signerAddress string
	// Specifies the key from the keystore associated with an account that
	// will be used to sign transactions.
	// NOTE: This `Account` must be available in the `Keystore`.
	keyName string
	// gasPrice represents the amount to be paid per gas unit.
	// Negative gasPrice means user want us to use the minGasPrice
	// defined in the node.
	gasPrice float64
	// since gasPrice can be 0, it is necessary to understand that user explicitly set it.
	isGasPriceSet bool
	// 0 gas means users want us to calculate it for them.
	gas uint64

	// Specifies the account that will pay for the transaction.
	// Input format Bech32.
	feeGranterAddress string
}
```

The keyName field allows the user to use a non-default account from the Keystore to pay fees.
The granter field is moved from the state config. The user will need to specify the granter address each time he wants to submit a tx with a granter.  The reason for moving the granter to the options struct is that previously user had to restart the node each time when he wanted to start/stop using `grantee` mode. 
Tested on mocha with several accounts:

``` 
celestia blob submit 0x42690c204d39600fddd3 'gm' --key.name testKey --gas 140000
{
  "result": {
    "height": 1383824,
    "commitments": [
      "IXg+08HV5RsPF3Lle8PH+B2TUGsGUsBiseflxh6wB5E="
    ]
  }
}
```

```
celestia blob submit 0x42690c204d39600fddd3 'gm' --gas.price 0.1 --gas 1400000 --granter.address celestia14s2aeemrzw34crdh79e6wj56kjzpu4ffmvxsyx
{
  "result": {
    "height": 1383809,
    "commitments": [
      "IXg+08HV5RsPF3Lle8PH+B2TUGsGUsBiseflxh6wB5E="
    ]
  }
}
```

```
celestia state transfer celestia14s2aeemrzw34crdh79e6wj56kjzpu4ffmvxsyx 10000 --gas 100000 --gas.price 0.1
{
  "result": {
    "height": 1383817,
    "txhash": "1F38E94AFAE8AA4C4819844100803B8D7C83016DB1C02148F332FBB8E8DBC857",
```

List of breaking changes:
* Added a TxConfig struct that is applied to all *write* transactions(transfer/submitPFB etc);
* Renamed `KeyringAccName` to `DefaultKeyName` and `KeyringBackend`  to `DefaultBackendName` in the config;
* Removed `blob.GasPrice` since `blob.Submit` requires `SubmitOptions` which are type alias of `TxConfig`;
* Reworked flags for tx submission through CLI;
* Added `state.Blob` which is an alias for `app.Blob`(State Module now requires blob type from app)(Was added to avoid circular dependency between node's blob Module and node's state Module);

### DIFF
--- a/api/docgen/examples.go
+++ b/api/docgen/examples.go
@@ -178,14 +178,14 @@ func init() {
 	}
 	addToExampleValues(libhead.Hash(hash))
 
-	txOptions := state.NewTxOptions(
+	txConfig := state.NewTxConfig(
 		state.WithGasPrice(0.002),
 		state.WithGas(142225),
 		state.WithKeyName("my_celes_key"),
 		state.WithSignerAddress("celestia1pjcmwj8w6hyr2c4wehakc5g8cfs36aysgucx66"),
 		state.WithFeeGranterAddress("celestia1hakc56ax66ypjcmwj8w6hyr2c4g8cfs3wesguc"),
 	)
-	addToExampleValues(txOptions)
+	addToExampleValues(txConfig)
 }
 
 func addToExampleValues(v interface{}) {
--- a/api/docgen/examples.go
+++ b/api/docgen/examples.go
@@ -61,7 +61,6 @@ var ExampleValues = map[reflect.Type]interface{}{
 	reflect.TypeOf(42):                       42,
 	reflect.TypeOf(byte(7)):                  byte(7),
 	reflect.TypeOf(float64(42)):              float64(42),
-	reflect.TypeOf(blob.GasPrice(0)):         blob.GasPrice(0.002),
 	reflect.TypeOf(true):                     true,
 	reflect.TypeOf([]byte{}):                 []byte("byte array"),
 	reflect.TypeOf(node.Full):                node.Full,
@@ -187,7 +186,7 @@ func init() {
 	txOptions := options.DefaultTxOptions()
 	txOptions.AccountKey = "my_celes_key"
 	txOptions.Gas = 142225
-	txOptions.SetFeeAmount(2_000)
+	txOptions.SetGasPrice(0.002)
 	txOptions.FeeGranterAddress = "celestia1hakc56ax66ypjcmwj8w6hyr2c4g8cfs3wesguc"
 	addToExampleValues(txOptions)
 }
--- a/api/docgen/examples.go
+++ b/api/docgen/examples.go
@@ -188,7 +188,7 @@ func init() {
 	txOptions.Account = "my_celes_key"
 	txOptions.Gas = 142225
 	txOptions.SetFeeAmount(2_000)
-	txOptions.Granter = "celestia1hakc56ax66ypjcmwj8w6hyr2c4g8cfs3wesguc"
+	txOptions.FeeGranterAddress = "celestia1hakc56ax66ypjcmwj8w6hyr2c4g8cfs3wesguc"
 	addToExampleValues(txOptions)
 }
 
--- a/api/docgen/examples.go
+++ b/api/docgen/examples.go
@@ -185,7 +185,7 @@ func init() {
 	addToExampleValues(blobProof)
 
 	txOptions := options.DefaultTxOptions()
-	txOptions.Account = "my_celes_key"
+	txOptions.AccountKey = "my_celes_key"
 	txOptions.Gas = 142225
 	txOptions.SetFeeAmount(2_000)
 	txOptions.FeeGranterAddress = "celestia1hakc56ax66ypjcmwj8w6hyr2c4g8cfs3wesguc"
--- a/api/docgen/examples.go
+++ b/api/docgen/examples.go
@@ -181,7 +181,8 @@ func init() {
 	txOptions := state.NewTxOptions(
 		state.WithGasPrice(0.002),
 		state.WithGas(142225),
-		state.WithAccountKey("my_celes_key"),
+		state.WithKeyName("my_celes_key"),
+		state.WithSignerAddress("celestia1pjcmwj8w6hyr2c4wehakc5g8cfs36aysgucx66"),
 		state.WithFeeGranterAddress("celestia1hakc56ax66ypjcmwj8w6hyr2c4g8cfs3wesguc"),
 	)
 	addToExampleValues(txOptions)
--- a/api/docgen/examples.go
+++ b/api/docgen/examples.go
@@ -31,7 +31,6 @@ import (
 	"github.com/celestiaorg/celestia-node/share"
 	"github.com/celestiaorg/celestia-node/share/eds/byzantine"
 	"github.com/celestiaorg/celestia-node/state"
-	"github.com/celestiaorg/celestia-node/state/options"
 )
 
 //go:embed "exampledata/extendedHeader.json"
@@ -183,11 +182,12 @@ func init() {
 	blobProof = &blob.Proof{&proof}
 	addToExampleValues(blobProof)
 
-	txOptions := options.DefaultTxOptions()
-	txOptions.AccountKey = "my_celes_key"
-	txOptions.Gas = 142225
-	txOptions.SetGasPrice(0.002)
-	txOptions.FeeGranterAddress = "celestia1hakc56ax66ypjcmwj8w6hyr2c4g8cfs3wesguc"
+	txOptions := state.NewTxOptions(
+		state.WithGasPrice(0.002),
+		state.WithGas(142225),
+		state.WithAccountKey("my_celes_key"),
+		state.WithFeeGranterAddress("celestia1hakc56ax66ypjcmwj8w6hyr2c4g8cfs3wesguc"),
+	)
 	addToExampleValues(txOptions)
 }
 
--- a/api/docgen/examples.go
+++ b/api/docgen/examples.go
@@ -186,7 +186,7 @@ func init() {
 
 	txOptions := options.DefaultTxOptions()
 	txOptions.Account = "my_celes_key"
-	txOptions.GasLimit = 142225
+	txOptions.Gas = 142225
 	txOptions.SetFeeAmount(2_000)
 	txOptions.Granter = "celestia1hakc56ax66ypjcmwj8w6hyr2c4g8cfs3wesguc"
 	addToExampleValues(txOptions)
--- a/api/docgen/examples.go
+++ b/api/docgen/examples.go
@@ -31,6 +31,7 @@ import (
 	"github.com/celestiaorg/celestia-node/share"
 	"github.com/celestiaorg/celestia-node/share/eds/byzantine"
 	"github.com/celestiaorg/celestia-node/state"
+	"github.com/celestiaorg/celestia-node/state/options"
 )
 
 //go:embed "exampledata/extendedHeader.json"
@@ -177,8 +178,18 @@ func init() {
 	if err != nil {
 		panic(err)
 	}
-
 	addToExampleValues(libhead.Hash(hash))
+
+	proof := nmt.NewInclusionProof(0, 4, [][]byte{[]byte("test")}, true)
+	blobProof = &blob.Proof{&proof}
+	addToExampleValues(blobProof)
+
+	txOptions := options.DefaultTxOptions()
+	txOptions.Account = "celestia1fcw3wrlhzk4pf5y68m4009pzyg97p2ftaezl6y"
+	txOptions.GasLimit = 142225
+	txOptions.SetFeeAmount(2_000)
+	txOptions.Granter = "celestia1hakc56ax66ypjcmwj8w6hyr2c4g8cfs3wesguc"
+	addToExampleValues(txOptions)
 }
 
 func addToExampleValues(v interface{}) {
--- a/api/docgen/examples.go
+++ b/api/docgen/examples.go
@@ -178,10 +178,6 @@ func init() {
 	}
 	addToExampleValues(libhead.Hash(hash))
 
-	proof := nmt.NewInclusionProof(0, 4, [][]byte{[]byte("test")}, true)
-	blobProof = &blob.Proof{&proof}
-	addToExampleValues(blobProof)
-
 	txOptions := state.NewTxOptions(
 		state.WithGasPrice(0.002),
 		state.WithGas(142225),
--- a/api/docgen/examples.go
+++ b/api/docgen/examples.go
@@ -185,7 +185,7 @@ func init() {
 	addToExampleValues(blobProof)
 
 	txOptions := options.DefaultTxOptions()
-	txOptions.Account = "celestia1fcw3wrlhzk4pf5y68m4009pzyg97p2ftaezl6y"
+	txOptions.Account = "my_celes_key"
 	txOptions.GasLimit = 142225
 	txOptions.SetFeeAmount(2_000)
 	txOptions.Granter = "celestia1hakc56ax66ypjcmwj8w6hyr2c4g8cfs3wesguc"
--- a/blob/service.go
+++ b/blob/service.go
@@ -68,21 +68,6 @@ func NewService(
 	}
 }
 
-// SubmitOptions contains the information about fee and gasLimit price in order to configure the
-// Submit request.
-type SubmitOptions struct {
-	Fee      int64
-	GasLimit uint64
-}
-
-// DefaultSubmitOptions creates a default fee and gas price values.
-func DefaultSubmitOptions() *SubmitOptions {
-	return &SubmitOptions{
-		Fee:      -1,
-		GasLimit: 0,
-	}
-}
-
 // Submit sends PFB transaction and reports the height at which it was included.
 // Allows sending multiple Blobs atomically synchronously.
 // Uses default wallet registered on the Node.
--- a/blob/service.go
+++ b/blob/service.go
@@ -31,7 +31,7 @@ var (
 )
 
 // GasPrice represents the amount to be paid per gas unit. Fee is set by
-// multiplying GasPrice by GasLimit, which is determined by the blob sizes.
+// multiplying GasPrice by Gas, which is determined by the blob sizes.
 type GasPrice float64
 
 // DefaultGasPrice returns the default gas price, letting node automatically
--- a/blob/service.go
+++ b/blob/service.go
@@ -34,7 +34,7 @@ var (
 // avoid a circular dependency between the blob and the state package, since the state package needs
 // the blob.Blob type for this signature.
 type Submitter interface {
-	SubmitPayForBlob(context.Context, []*state.Blob, state.Options) (*types.TxResponse, error)
+	SubmitPayForBlob(context.Context, []*state.Blob, *state.TxOptions) (*types.TxResponse, error)
 }
 
 type Service struct {
@@ -62,7 +62,7 @@ func NewService(
 // Allows sending multiple Blobs atomically synchronously.
 // Uses default wallet registered on the Node.
 // Handles gas estimation and fee calculation.
-func (s *Service) Submit(ctx context.Context, blobs []*Blob, txOptions state.Options) (uint64, error) {
+func (s *Service) Submit(ctx context.Context, blobs []*Blob, txOptions *state.TxOptions) (uint64, error) {
 	log.Debugw("submitting blobs", "amount", len(blobs))
 
 	appblobs := make([]*state.Blob, len(blobs))
--- a/blob/service.go
+++ b/blob/service.go
@@ -4,26 +4,22 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"math"
 	"slices"
 	"sync"
 
-	sdkmath "cosmossdk.io/math"
 	"github.com/cosmos/cosmos-sdk/types"
-	auth "github.com/cosmos/cosmos-sdk/x/auth/types"
 	logging "github.com/ipfs/go-log/v2"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/trace"
 
-	"github.com/celestiaorg/celestia-app/pkg/appconsts"
 	"github.com/celestiaorg/celestia-app/pkg/shares"
-	blobtypes "github.com/celestiaorg/celestia-app/x/blob/types"
 
 	"github.com/celestiaorg/celestia-node/header"
 	"github.com/celestiaorg/celestia-node/libs/utils"
 	"github.com/celestiaorg/celestia-node/share"
+	"github.com/celestiaorg/celestia-node/state/options"
 )
 
 var (
@@ -48,7 +44,7 @@ func DefaultGasPrice() GasPrice {
 // avoid a circular dependency between the blob and the state package, since the state package needs
 // the blob.Blob type for this signature.
 type Submitter interface {
-	SubmitPayForBlob(ctx context.Context, fee sdkmath.Int, gasLim uint64, blobs []*Blob) (*types.TxResponse, error)
+	SubmitPayForBlob(context.Context, []*Blob, *options.TxOptions) (*types.TxResponse, error)
 }
 
 type Service struct {
@@ -91,20 +87,10 @@ func DefaultSubmitOptions() *SubmitOptions {
 // Allows sending multiple Blobs atomically synchronously.
 // Uses default wallet registered on the Node.
 // Handles gas estimation and fee calculation.
-func (s *Service) Submit(ctx context.Context, blobs []*Blob, gasPrice GasPrice) (uint64, error) {
+func (s *Service) Submit(ctx context.Context, blobs []*Blob, txOptions *options.TxOptions) (uint64, error) {
 	log.Debugw("submitting blobs", "amount", len(blobs))
 
-	options := DefaultSubmitOptions()
-	if gasPrice >= 0 {
-		blobSizes := make([]uint32, len(blobs))
-		for i, blob := range blobs {
-			blobSizes[i] = uint32(len(blob.Data))
-		}
-		options.GasLimit = blobtypes.EstimateGas(blobSizes, appconsts.DefaultGasPerBlobByte, auth.DefaultTxSizeCostPerByte)
-		options.Fee = types.NewInt(int64(math.Ceil(float64(gasPrice) * float64(options.GasLimit)))).Int64()
-	}
-
-	resp, err := s.blobSubmitter.SubmitPayForBlob(ctx, types.NewInt(options.Fee), options.GasLimit, blobs)
+	resp, err := s.blobSubmitter.SubmitPayForBlob(ctx, blobs, txOptions)
 	if err != nil {
 		return 0, err
 	}
--- a/blob/service.go
+++ b/blob/service.go
@@ -30,6 +30,10 @@ var (
 	tracer = otel.Tracer("blob/service")
 )
 
+// SubmitOptions aliases TxOptions from state package allowing users
+// to specify options for SubmitPFB transaction.
+type SubmitOptions = state.TxConfig
+
 // Submitter is an interface that allows submitting blobs to the celestia-core. It is used to
 // avoid a circular dependency between the blob and the state package, since the state package needs
 // the blob.Blob type for this signature.
@@ -62,7 +66,7 @@ func NewService(
 // Allows sending multiple Blobs atomically synchronously.
 // Uses default wallet registered on the Node.
 // Handles gas estimation and fee calculation.
-func (s *Service) Submit(ctx context.Context, blobs []*Blob, txConfig *state.TxConfig) (uint64, error) {
+func (s *Service) Submit(ctx context.Context, blobs []*Blob, txConfig *SubmitOptions) (uint64, error) {
 	log.Debugw("submitting blobs", "amount", len(blobs))
 
 	appblobs := make([]*state.Blob, len(blobs))
--- a/blob/service.go
+++ b/blob/service.go
@@ -19,7 +19,7 @@ import (
 	"github.com/celestiaorg/celestia-node/header"
 	"github.com/celestiaorg/celestia-node/libs/utils"
 	"github.com/celestiaorg/celestia-node/share"
-	"github.com/celestiaorg/celestia-node/state/options"
+	"github.com/celestiaorg/celestia-node/state"
 )
 
 var (
@@ -34,7 +34,7 @@ var (
 // avoid a circular dependency between the blob and the state package, since the state package needs
 // the blob.Blob type for this signature.
 type Submitter interface {
-	SubmitPayForBlob(context.Context, []*Blob, *options.TxOptions) (*types.TxResponse, error)
+	SubmitPayForBlob(context.Context, []*state.Blob, state.Options) (*types.TxResponse, error)
 }
 
 type Service struct {
@@ -62,10 +62,18 @@ func NewService(
 // Allows sending multiple Blobs atomically synchronously.
 // Uses default wallet registered on the Node.
 // Handles gas estimation and fee calculation.
-func (s *Service) Submit(ctx context.Context, blobs []*Blob, txOptions *options.TxOptions) (uint64, error) {
+func (s *Service) Submit(ctx context.Context, blobs []*Blob, txOptions state.Options) (uint64, error) {
 	log.Debugw("submitting blobs", "amount", len(blobs))
 
-	resp, err := s.blobSubmitter.SubmitPayForBlob(ctx, blobs, txOptions)
+	appblobs := make([]*state.Blob, len(blobs))
+	for i := range blobs {
+		if err := blobs[i].Namespace().ValidateForBlob(); err != nil {
+			return 0, err
+		}
+		appblobs[i] = &blobs[i].Blob
+	}
+
+	resp, err := s.blobSubmitter.SubmitPayForBlob(ctx, appblobs, txOptions)
 	if err != nil {
 		return 0, err
 	}
--- a/blob/service.go
+++ b/blob/service.go
@@ -30,16 +30,6 @@ var (
 	tracer = otel.Tracer("blob/service")
 )
 
-// GasPrice represents the amount to be paid per gas unit. Fee is set by
-// multiplying GasPrice by Gas, which is determined by the blob sizes.
-type GasPrice float64
-
-// DefaultGasPrice returns the default gas price, letting node automatically
-// determine the Fee based on the passed blob sizes.
-func DefaultGasPrice() GasPrice {
-	return -1.0
-}
-
 // Submitter is an interface that allows submitting blobs to the celestia-core. It is used to
 // avoid a circular dependency between the blob and the state package, since the state package needs
 // the blob.Blob type for this signature.
--- a/blob/service.go
+++ b/blob/service.go
@@ -34,7 +34,7 @@ var (
 // avoid a circular dependency between the blob and the state package, since the state package needs
 // the blob.Blob type for this signature.
 type Submitter interface {
-	SubmitPayForBlob(context.Context, []*state.Blob, *state.TxOptions) (*types.TxResponse, error)
+	SubmitPayForBlob(context.Context, []*state.Blob, *state.TxConfig) (*types.TxResponse, error)
 }
 
 type Service struct {
@@ -62,7 +62,7 @@ func NewService(
 // Allows sending multiple Blobs atomically synchronously.
 // Uses default wallet registered on the Node.
 // Handles gas estimation and fee calculation.
-func (s *Service) Submit(ctx context.Context, blobs []*Blob, txOptions *state.TxOptions) (uint64, error) {
+func (s *Service) Submit(ctx context.Context, blobs []*Blob, txConfig *state.TxConfig) (uint64, error) {
 	log.Debugw("submitting blobs", "amount", len(blobs))
 
 	appblobs := make([]*state.Blob, len(blobs))
@@ -73,7 +73,7 @@ func (s *Service) Submit(ctx context.Context, blobs []*Blob, txOptions *state.Tx
 		appblobs[i] = &blobs[i].Blob
 	}
 
-	resp, err := s.blobSubmitter.SubmitPayForBlob(ctx, appblobs, txOptions)
+	resp, err := s.blobSubmitter.SubmitPayForBlob(ctx, appblobs, txConfig)
 	if err != nil {
 		return 0, err
 	}
--- a/nodebuilder/blob/cmd/blob.go
+++ b/nodebuilder/blob/cmd/blob.go
@@ -13,20 +13,13 @@ import (
 
 	"github.com/celestiaorg/celestia-node/blob"
 	cmdnode "github.com/celestiaorg/celestia-node/cmd"
+	state "github.com/celestiaorg/celestia-node/nodebuilder/state/cmd"
 	"github.com/celestiaorg/celestia-node/share"
 	"github.com/celestiaorg/celestia-node/state/options"
 )
 
 var (
 	base64Flag bool
-
-	fee int64
-
-	gasLimit uint64
-
-	account string
-
-	granter string
 
 	// flagFileInput allows the user to provide file path to the json file
 	// for submitting multiple blobs.
@@ -50,40 +43,7 @@ func init() {
 		"Printed blob's data and namespace as base64 strings",
 	)
 
-	submitCmd.PersistentFlags().Int64Var(
-		&fee,
-		"fee",
-		-1,
-		"Specifies fee(in utia) for blob submission.\n"+
-			"Fee will be set to default(-1) if no value is passed.",
-	)
-
-	submitCmd.PersistentFlags().Uint64Var(
-		&gasLimit,
-		"gas.limit",
-		0,
-		"Specifies gas limit (in utia) for blob submission.\n"+
-			"Gas Limit will be set to default(0) if no value is passed",
-	)
-
-	submitCmd.PersistentFlags().String(
-		account,
-		"",
-		"Specifies the signer address.\n"+
-			"Account address will be set to an empty string in case no value is passed.\n"+
-			"Note: Address should be passed as Bench32 address.\n"+
-			"Example: celestiaxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
-	)
-
-	submitCmd.PersistentFlags().String(
-		granter,
-		"",
-		"Specifies the address that can pay fees on behalf of the signer."+
-			"If no value is passed, the granter address will be set to an empty string."+
-			"The granter must submit the transaction to pay for the grantee's (signer's) transactions."+
-			"By default, this will be set to an empty string, meaning the signer will pay the fees."+
-			"Note: The granter should be provided as a Bech32 address.",
-	)
+	state.ApplyFlags(submitCmd)
 
 	submitCmd.PersistentFlags().String(flagFileInput, "", "Specifies the file input")
 }
@@ -246,10 +206,10 @@ var submitCmd = &cobra.Command{
 		}
 
 		opts := options.DefaultTxOptions()
-		opts.SetFeeAmount(fee)
-		opts.GasLimit = gasLimit
-		opts.Account = account
-		opts.Granter = granter
+		opts.SetFeeAmount(state.Fee)
+		opts.GasLimit = state.GasLimit
+		opts.Account = state.Account
+		opts.Granter = state.Granter
 		height, err := client.Blob.Submit(
 			cmd.Context(),
 			resultBlobs,
--- a/nodebuilder/blob/cmd/blob.go
+++ b/nodebuilder/blob/cmd/blob.go
@@ -209,7 +209,7 @@ var submitCmd = &cobra.Command{
 		opts.SetFeeAmount(state.Fee)
 		opts.Gas = state.Gas
 		opts.Account = state.Account
-		opts.Granter = state.Granter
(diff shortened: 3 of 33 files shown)


distractedm1nd (approved):

liamsi (commented):
This PR looks good to me. 
Just re-iterating that the most important feature/user-request is that users can submit blobs from different accounts. No other options should be added before that particular one. Having an options struct part of the JSON-RPC further more manifests that this current API is basically driven by implementation details (this is very go idiomatic). No blocker to merge this but sth I wanted to raise.

cc @renaynay

state/core_access.go
```diff
@@ -224,6 +224,19 @@ func (ca *CoreAccessor) SubmitPayForBlob(
 	fee Int,
 	gasLim uint64,
 	blobs []*blob.Blob,
+) (*TxResponse, error) {
+	return ca.SubmitPayForBlobWithOptions(ctx, fee, gasLim, blobs, nil)
+}
+
+// SubmitPayForBlobWithOptions builds, signs, and submits a MsgPayForBlob with an additional options defined
+// in `PayForBlobOptions`. If gasLim is set to 0, the method will automatically estimate the gas limit.
```
renaynay, line 232:
```suggestion
// SubmitPayForBlobWithOptions builds, signs, and submits a MsgPayForBlob with additional options defined
// in `PayForBlobOptions`. If gasLim is set to 0, the method will automatically estimate the gas limit.
```

state/core_access.go
```diff
@@ -224,6 +224,19 @@ func (ca *CoreAccessor) SubmitPayForBlob(
...
+// If the fee is negative, the method will use the nodes min gas price multiplied by the gas limit.
+func (ca *CoreAccessor) SubmitPayForBlobWithOptions(
+	ctx context.Context,
+	fee Int,
+	gasLim uint64,
```
renaynay, line 237: [discussion] - should gasLim be added as an option bc it's not necessary to set it? if the field is not set in options, it would just do the calculation similar to if a negative fee is passed.

cc @Wondertan

renaynay (commented):
LGTM - want to discuss the `gasLim` option there

Should we also update ADR 9 (API ADR)?

state/core_access.go
renaynay: @cmwaters

renaynay: @liamsi see the comment above regarding
> Having an options struct part of the JSON-RPC further more manifests that this current API is basically driven by implementation details

liamsi: Can we please first add the option to define which account to use for signing before adding more things to the options? That said, if gasLim is optional, yes it should go into options. Maybe we can remove it from this Pr instead though. Introduce the only important option first and then bikeshed about other options are to be added?

vgonkivs: My suggestion is to be iterative and to do everything step by step. I mean, let's merge this pr with empty options and then we can extend the struct whenever we need.

Wondertan: This has an issue making the main branch unreleasable. I think we should stick to an approach where we implement the whole feature fully and merge it in the main. We can either do this through the feature branch or through a single PR that has the whole feature

vgonkivs: Feature branch also works for me

nodebuilder/state/keyring.go
```diff
@@ -16,15 +16,19 @@ type AccountName string
...
 			return nil, "", err
 		}
-		info = keyInfo
-	} else {
+		if info == nil {
+			info = keyInfo
+		}
+	}
+
+	if info == nil {
```
walldiss, line 31: It looks like keyInfo variable gets overwritten on on every iteration over KeyringAccName. So only the last one will be used. What is the reason to have multiple accounts specified, if only one will be used? If there is some logic implied, lets add comment for it.

state/core_access.go
walldiss: Currently, gasLim is passed as a parameter, which would prevent moving it to options later without breaking the API. To avoid this, let remove gasLim from the parameters, allowing for optional future configuration.

nodebuilder/state/keyring.go
vgonkivs: it does not. info is set up on the first iteration of the loop or the default account is used in case loop is empty.

nodebuilder/state/flags.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 var (
-	keyringAccNameFlag = "keyring.accname"
+	keyringAccNameFlag = "keyring.accnames"
```
renaynay, line 12: i'd prefer not to break the name of flag pls

it's okay to just have it as accname and allow for a list as it preserves backward compat

nodebuilder/state/flags.go
```diff
@@ -19,7 +19,7 @@ var (
 func Flags() *flag.FlagSet {
 	flags := &flag.FlagSet{}
 
-	flags.String(keyringAccNameFlag, "", "Directs node's keyring signer to use the key prefixed with the "+
+	flags.StringSlice(keyringAccNameFlag, []string{}, "Directs node's keyring signer to use the key prefixed with the "+
```
renaynay, line 22: please add doc on how it can be used to specify multiple keys

state/state.go
```diff
@@ -58,4 +58,8 @@ func (a Address) MarshalJSON() ([]byte, error) {
 }
 
 // PayForBlobOptions contains additional options that will be applied to the `MsgPayForBlob`.
-type PayForBlobOptions struct{}
+type PayForBlobOptions struct {
+	// Specifies the key that should be used to sign transactions.
+	// NOTE: This `AccName` should be available in the `Keystore`.
+	AccName string
```
renaynay, line 64: I would call it `KeyName` so it's more clear what we are actually looking for.

renaynay: speaking of which, we should eventually change the name of the `--keyring.accname flag` to `--keyring.keyname`

renaynay: Should we break it in this release? @liamsi @Wondertan 

I feel like `--keyring.keyname` is more clear and it'd prevent us having to break in future

nodebuilder/state/keyring.go
renaynay: @vgonkivs default key should only be used in the case that no keys are specified (len(cfg.KeyringAccName) == 0)

state/state.go
renaynay: also would change the name of KeyringAccName in state config (which is breaking as well)

we shouldn't have used this terminology from sdk / comet to begin with.

nodebuilder/state/keyring.go
```diff
@@ -16,15 +16,19 @@ type AccountName string
...
 	ring := ks.Keyring()
 	var info *kr.Record
-	// if custom keyringAccName provided, find key for that name
-	if cfg.KeyringAccName != "" {
-		keyInfo, err := ring.Key(cfg.KeyringAccName)
+
+	for _, accName := range cfg.KeyringAccName {
+		keyInfo, err := ring.Key(accName)
 		if err != nil {
 			log.Errorw("failed to find key by given name", "keyring.accname", cfg.KeyringAccName)
```
renaynay, line 23: this log needs to be changed to log just the keyName that doesn't exist

nodebuilder/state/keyring.go
renaynay: also please document what is happening here (just checking that keys that are specified actually exist in keyring)

renaynay (changes requested):

nodebuilder/state/keyring.go
vgonkivs: > @vgonkivs default key should only be used in the case that no keys are specified (len(cfg.KeyringAccName) == 0)

if (len(cfg.KeyringAccName) == 0), then info == nil.

state/state.go
```diff
@@ -56,3 +56,10 @@ func (a Address) MarshalJSON() ([]byte, error) {
 	// The address is marshaled into a simple string value
 	return []byte("\"" + a.Address.String() + "\""), nil
 }
+
+// PayForBlobOptions contains additional options that will be applied to the `MsgPayForBlob`.
+type PayForBlobOptions struct {
```
distractedm1nd, line 61: I would tend to argue that a struct is the wrong approach for this. Once more options are added, it will mean that users must set ALL options

Above I say we could use variadic params, but that doesn't actually work because you cannot pass those types over RPC.

An idea that /might/ work is a map[string]interface{} (I know, ugly, but we can hide it behind a custom type with easy to use constructors and validation methods, possibly using actual custom variadic options). This sounds a bit complex, so we'd have to see it and try it before deciding if it makes sense. Especially because it means this custom type+methods would have to be implemented in the other languages' client libraries to keep good devex.

I don't really like that either, but I think a struct can't work, so we need to find some alternative

distractedm1nd (changes requested):

state/state.go
renaynay: We could ask for `[]Option` instead and users could pass `DefaultOptions` which is just some empty []opt list.

This is a way to work around variadics not being over RPC

distractedm1nd: The issue isn't directly variadics, it's the actual option type. How is it represented? And how does it get sent over the wire?

state/core_access.go
walldiss: Let's consolidate all parameters that can default to implicit values into the options structure in this PR.

nodebuilder/state/flags.go
```diff
@@ -19,8 +19,9 @@ var (
 func Flags() *flag.FlagSet {
 	flags := &flag.FlagSet{}
 
-	flags.String(keyringAccNameFlag, "", "Directs node's keyring signer to use the key prefixed with the "+
-		"given string.")
+	flags.StringSlice(keyringKeyNameFlag, []string{}, "Directs node's keyring signer to use the key prefixed with the "+
+		"given string. The first key in the list will be the default key that will be used to sign the transactions. "+
+		"All other keys can be used to sign transactions, given that the user specifies it in the transaction options")
```
renaynay, line 24:
```suggestion
	flags.StringSlice(keyringKeyNameFlag, []string{}, "Directs node's keyring signer to use the key by the "+
		"given name. The first key in the list will be the default key that will be used to sign the transactions. "+
		"All other keys can be used to sign transactions, given that the user specifies it in the transaction options.")
```

nodebuilder/state/flags.go
```diff
@@ -30,10 +31,11 @@ func Flags() *flag.FlagSet {
 
 // ParseFlags parses State flags from the given cmd and saves them to the passed config.
 func ParseFlags(cmd *cobra.Command, cfg *Config) error {
-	keyringAccName := cmd.Flag(keyringAccNameFlag).Value.String()
-	if keyringAccName != "" {
-		cfg.KeyringAccName = keyringAccName
+	keyringKeyName, err := cmd.Flags().GetStringSlice(keyringKeyNameFlag)
```
renaynay, line 34: nit for clarity
```suggestion
	keyringKeyNames, err := cmd.Flags().GetStringSlice(keyringKeyNameFlag)
```

nodebuilder/state/keyring.go
```diff
@@ -16,15 +16,20 @@ type AccountName string
...
-	// if custom keyringAccName provided, find key for that name
-	if cfg.KeyringAccName != "" {
-		keyInfo, err := ring.Key(cfg.KeyringAccName)
+
+	// go through all keys in the config and check their availability in the KeyStore.
+	for _, accName := range cfg.KeyringKeyName {
+		keyInfo, err := ring.Key(accName)
 		if err != nil {
-			log.Errorw("failed to find key by given name", "keyring.accname", cfg.KeyringAccName)
+			log.Errorw("keyname does not exist", "keyring.keyname", cfg.KeyringKeyName)
```
renaynay, line 24: Pls both log and return this error: 

```
key not found in keystore: %s
```

renaynay (commented):
Question for potential follow-up PR: 
should we break `SubmitPayForBlob` on the state level to not take gasLim or fee and just use defaults? And then we can make blob module use `SubmitPayForBlobWithOptions` under the hood if opts are specified?

nodebuilder/testing.go
```diff
@@ -49,10 +53,14 @@ func TestNodeWithConfig(t *testing.T, tp node.Type, cfg *Config, opts ...fx.Opti
...
 	kr := ks.Keyring()
+
 	// create a key in the keystore to be used by the core accessor
 	_, _, err = kr.NewMnemonic(TestKeyringName, keyring.English, "", "", hd.Secp256k1)
 	require.NoError(t, err)
-	cfg.State.KeyringAccName = TestKeyringName
+	_, _, err = kr.NewMnemonic(TestKeyringName1, keyring.English, "", "", hd.Secp256k1)
+	require.NoError(t, err)
+	cfg.State.KeyringKeyName = []string{TestKeyringName, TestKeyringName1}
+	_, _, _ = state.Keyring(cfg.State, ks)
```
renaynay, line 63: are you not going to check any return val here/

distractedm1nd (commented):
Are we going to add the optional parameters to the struct in this PR as discussed above?

distractedm1nd (dismissed):
LGTM

vgonkivs:
> should we break SubmitPayForBlob on the state level to not take gasLim or fee and just use defaults? And then we can make blob module use SubmitPayForBlobWithOptions under the hood if opts are specified?

Yes, I also thought about it.

liamsi:
It sounds like there is no consensus around if we should unify SubmitPayForBlob and SubmitPayForBlobWithOptions and fold them into one method. My opinion:
- if we were pre-launch and no-one was using the existing API, we should certainly not expose two very similar methods bu instead fold the functionality into one method.
- now that we are live, we should be very careful with introducing API breaking changes even though the API isn't v1 yet
- unless someone checks with external teams + solutions that they would welcome a cleaner but breaking change and are willing to upgrade their integrations for that asap, we should not break the API (@distractedm1nd and @Bidon15 to verify)

liamsi:
So currently we already have two methods to submit blobs:
- `Blob.Submit(blobs []*blob.Blob, gasPrice blob.GasPrice)uint64perms: write`
- `State.SubmitPayForBlob(fee state.Int, gasLim uint64, blobs []*blob.Blob)*state.TxResponseperms: write`

Not speak of the infamous `SubmitTx(tx state.Tx)*state.TxResponse` method which is slated to be removed. 

This PR introduces another method:
- `Blob.SubmitPayForBlobWithOptions

After talking with @renaynay: If we have significant proof that users are currently using the `State` method instead of the `Blob` method, then I am also in favour of breaking the current Blob.Submit instead of introducing a third method here!

distractedm1nd:
Alright, the optional params PR for `go-jsonrpc` isn’t really workable. It makes all parameters with pointers optional, and it doesn’t solve the problem that the generated client can’t actually mark the methods as optional.Basically, if you don’t send the optional params over the wire, the server will not complain, but the client methods won’t actually reflect that any params are optional.

And, variadics also wouldn’t work - as @vgonkivs pointed out internally, we cannot send our traditional options (which are just function pointers) over the wire.

Given that we want as many users as possible to use blob module, the cleanest way using go-jsonrpc will be to move TxOptions (-> SubmitOptions?) type to the blob module. This would replace the GasLimit parameter. We should define blob.DefaultSubmitOptions() *SubmitOptions  and methods like opts.WithGasLimit(...).WithFee()  for best devex.   

As for the state module, we also need to eventually add some sort of TxOptions , which will probably have a lot of overlap                                                                                   with SubmitOptions, but likely with even more parameters. This is not very urgent though. This should be added to all state methods in place of the current parameters. If everyone else agrees with this, I will open an issue and get started on it ASAP

state/options/tx_options.go
```diff
@@ -0,0 +1,201 @@
...
+	GasLimit uint64
+
+	// Specifies the key from the keystore associated with an account that
+	// will be used to sign transactions.
+	// NOTE: This `Account` must be available in the `Keystore`.
+	Account string
+	// Specifies the account that will pay for the transaction.
+	// Input format Bech32.
+	Granter string
+}
```
cmwaters, line 44: I think we can make these fields a little more verbose and clear for the user. I would propose:
- renaming `Granter` to `FeePayingAddress`
- renaming `Account` to `SigningAccount`

I've tried to only use the term `Address` when referring to the Bech32 encoded SDK address and `Account` to refer to the key name that links to the underlying `Record` with the users private & public keys.

state/options/tx_options.go
```diff
@@ -0,0 +1,201 @@
...
+	return nil
+}
+
+// SetFeeAmount sets fee for the transaction.
+func (options *TxOptions) SetFeeAmount(amount int64) {
+	if amount >= 0 {
+		options.fee.Amount = amount
+		options.fee.isSet = true
+	}
+}
```
cmwaters, line 89: This method is for go users right? Because if I'm submitting a JSON RPC request, then the `Fee` field is public and I would fill it out myself.

state/options/tx_options.go
```diff
@@ -0,0 +1,201 @@
...
+
+var (
+	errNoGasProvided     = errors.New("gas limit was not set")
+	errNoAddressProvided = errors.New("address was not set")
+)
+
+// TxOptions specifies additional options that will be applied to the Tx.
+type TxOptions struct {
+	// fee is private since it has to be set through `SetFeeAmount`
+	fee      *Fee
```
cmwaters, line 34: Perhaps now would be a good time to advocate that the user should specify the gasPrice and not the total fee since it is the gas price that determines priority and inclusion.

state/options/tx_options.go
```diff
@@ -0,0 +1,201 @@
...
+	GasLimit uint64
```
cmwaters, line 35: I've always had a problem with `GasLimit` because `Limit` implies the maximum that you can consume (in gas) whereas the Cosmos SDK will always consume that amount, regardless for how much you actually use. It may be better to simply call it `Gas`.

Would a `Gas` of zero indicate that the user wants the gas to be calculated for them?

cmwaters (commented):
I have some questions and ideas that more relate to the semantics of the API than the logic.

state/options/tx_options.go
vgonkivs: > Account to refer to the key name that links to the underlying Record with the users private & public keys.
The account refers to the key name in the key store.

> renaming Granter to FeePayingAddress
Account will also pay fees iirc, so it may be confusing for the user.

state/options/tx_options.go
vgonkivs: Yes. It will be handled in `UnmarshalJSON`.

state/options/tx_options.go
vgonkivs: > Would a Gas of zero indicate that the user wants the gas to be calculated for them?

Yes.

>  It may be better to simply call it Gas.

Agree

state/options/tx_options.go
vgonkivs: This is something we discussed in DM last week. It makes a lot of sense for me but since the implementation was done, I decided to push all changes w/o `minGasPrice` and discuss it here.
@distractedm1nd , @renaynay wdyt?

also @jcstein for visibility.

distractedm1nd: Either we only have fee (without GasLimit), or we have GasPrice+Gas. I definitely prefer GasPrice+Gas

state/options/tx_options.go
distractedm1nd: I would also prefer keeping it as Account + Granter

But maybe Account + GranterAddress would be clearer. But no strong opinions here

cmwaters: > The account refers to the key name in the key store.

Yes that's exactly what I was saying (`Record` comes from the Keyring)

> Account will also pay fees iirc, so it may be confusing for the user.

Then maybe be more specific in the comments

> I would also prefer keeping it as Account + Granter

I don't think `Granter` by itself is sufficient. Imagine users who have no prior context. At least `FeeGranter`

vgonkivs: FeeGranterAddress sounds good to me.

state/options/tx_options.go
```diff
@@ -0,0 +1,209 @@
+package options
```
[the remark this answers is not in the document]
vgonkivs: it was done to avoid circular dependency between blob and state, but I think i found a better solution: SubmitPFB will accept app.Blob instead of our custom(potentially it will be changed to go-square blob)

state/options/tx_options.go
```diff
@@ -0,0 +1,209 @@
...
+
+func (options *TxOptions) GetGasPrice() float64 {
+	if options.gasPrice == nil {
+		options.gasPrice = DefaultGasPrice()
+	}
+	return options.gasPrice.Price
+}
+
+// CalculateFee calculates fee amount based on the `minGasPrice` and `Gas`.
+// NOTE: Gas can't be 0.
```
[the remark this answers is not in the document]
vgonkivs: Gas can't be 0 during fee calculation, since `fee=gasPrice*gas`

state/options/tx_options.go
```diff
@@ -0,0 +1,209 @@
...
+	options.AccountKey = jsonOpts.AccountKey
+	options.FeeGranterAddress = jsonOpts.FeeGranterAddress
+	return nil
+}
+
+// SetGasPrice sets fee for the transaction.
+func (options *TxOptions) SetGasPrice(price float64) {
+	if options.gasPrice == nil {
+		options.gasPrice = DefaultGasPrice()
+	}
```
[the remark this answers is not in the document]
vgonkivs: this is a protection as we are accessing the fields of the structure further. will panic in case it haven't been instantiated

state/options/tx_options.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+// CalculateFee calculates fee amount based on the `minGasPrice` and `Gas`.
+// NOTE: Gas can't be 0.
+func (options *TxOptions) CalculateFee() (int64, error) {
+	if options.Gas == 0 {
+		return 0, errNoGasProvided
+	}
+	if options.gasPrice == nil {
+		options.gasPrice = DefaultGasPrice()
```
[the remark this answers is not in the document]
vgonkivs: `GasPrice` is a struct, not just a value. It will panic in case it haven't been instantiated

state/options/tx_options.go
```diff
@@ -0,0 +1,209 @@
...
+// to cover additional options of the Tx.
+func (options *TxOptions) EstimateGasForBlobs(blobSizes []uint32) {
+	if options.Gas == 0 {
+		gas := apptypes.DefaultEstimateGas(blobSizes)
+		options.Gas = uint64(float64(gas) * gasMultiplier)
+	}
+}
+
+// GetSigner retrieves the keystore by the provided account name and returns the account address.
+func (options *TxOptions) GetSigner(kr keyring.Keyring) (sdktypes.AccAddress, error) {
```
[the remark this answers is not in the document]
vgonkivs: It returns and address that will sign the tx

state/core_access.go
```diff
@@ -694,30 +643,46 @@ func (ca *CoreAccessor) queryMinimumGasPrice(
...
-func (ca *CoreAccessor) setupSigner(ctx context.Context) (*user.Signer, error) {
-	encCfg := encoding.MakeConfig(app.ModuleEncodingRegisters...)
-	return user.SetupSigner(ctx, ca.keyring, ca.coreConn, ca.addr, encCfg)
-}
+	fee, err := options.CalculateFee()
+	if err != nil {
+		return nil, err
+	}
+	if fee == 0 {
+		log.Warn("gas price wasn't provided. Transaction will be sent with 0 fee.")
```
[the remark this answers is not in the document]
vgonkivs: removed

state/core_access.go
```diff
@@ -242,59 +205,78 @@ func (ca *CoreAccessor) SubmitPayForBlob(
...
-	// set granter and update gasLimit in case node run in a grantee mode
-	if !ca.granter.Empty() {
-		feeGrant = user.SetFeeGranter(ca.granter)
-		gasLim = uint64(float64(gasLim) * gasMultiplier)
+	fee, err := options.CalculateFee()
+	if err != nil {
+		return nil, err
+	}
+	if fee == 0 {
+		log.Warn("gas price wasn't provided. Transaction will be sent with 0 fee.")
```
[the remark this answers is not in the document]
vgonkivs: removed

state/options/tx_options.go
```diff
@@ -0,0 +1,209 @@
...
+	errNoAddressProvided = errors.New("address was not set")
+)
+
+// TxOptions specifies additional options that will be applied to the Tx.
+type TxOptions struct {
+	// GasPrice represents the amount to be paid per gas unit.
+	// Negative GasPrice means user want us to use a minGasPrice,
+	// stored in node.
+	gasPrice *GasPrice
+	// 0 Gas means users want us to calculate it for them.
```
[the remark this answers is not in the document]
vgonkivs: GasPrice can be set to 0 and this is a valid input. From the other side, Gas can't be 0.

state/core_access.go
```diff
@@ -242,59 +205,78 @@ func (ca *CoreAccessor) SubmitPayForBlob(
...
 				return nil, fmt.Errorf("parsing insufficient min gas price error: %w", err)
 			}
-			ca.setMinGasPrice(minGasPrice)
+
+			// update minGasPrice in case `IsInsufficientMinGasPrice` is received
+			if gasPrice > ca.getMinGasPrice() {
+				ca.setMinGasPrice(gasPrice)
+			}
+
+			options.SetGasPrice(gasPrice)
```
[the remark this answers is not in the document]
vgonkivs: we don't need this logic anymore

state/options/tx_options.go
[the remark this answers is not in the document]
vgonkivs: moved options inside state package with some workaround

state/options/tx_options.go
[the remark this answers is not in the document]
vgonkivs: N/A

state/options/tx_options.go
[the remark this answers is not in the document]
vgonkivs: N/A

state/options/tx_options.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+}
+
+type jsonGasPrice struct {
+	Price float64 `json:"price,omitempty"`
+	IsSet bool    `json:"isSet,omitempty"`
+}
+
+func (g *GasPrice) MarshalJSON() ([]byte, error) {
+	fee := jsonGasPrice{
```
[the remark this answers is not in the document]
vgonkivs: N/A

nodebuilder/state/keyring.go
```diff
@@ -16,15 +18,20 @@ type AccountName string
...
+		keyInfo, err := ring.Key(accName)
 		if err != nil {
-			log.Errorw("failed to find key by given name", "keyring.accname", cfg.KeyringAccName)
+			err = fmt.Errorf("key not found in keystore: %s", accName)
 			return nil, "", err
 		}
-		info = keyInfo
-	} else {
+		if info == nil {
+			info = keyInfo
```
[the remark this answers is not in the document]
vgonkivs: cfg.KeyringKeyNames can contain multiple accounts.  We should check that all accounts are available in the Keystore. The first account will be the default node's account.

nodebuilder/state/cmd/state.go
```diff
@@ -466,3 +419,49 @@ func parseAddressFromString(addrStr string) (state.Address, error) {
 	}
 	return address, nil
 }
+
+func ApplyFlags(cmds ...*cobra.Command) {
+	for _, cmd := range cmds {
+		cmd.PersistentFlags().Float64Var(
+			&gasPrice,
+			"gas.price",
+			-1,
```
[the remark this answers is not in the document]
vgonkivs: Yes. Will fix

api/docgen/examples.go
```diff
@@ -52,7 +52,6 @@ var ExampleValues = map[reflect.Type]interface{}{
 	reflect.TypeOf(42):                       42,
 	reflect.TypeOf(byte(7)):                  byte(7),
 	reflect.TypeOf(float64(42)):              float64(42),
-	reflect.TypeOf(blob.GasPrice(0)):         blob.GasPrice(0.002),
```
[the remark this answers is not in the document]
vgonkivs: ` blob.GasPrice` was an alias to float64 in blob module. It was removed.

state/tx_options.go
```diff
@@ -0,0 +1,168 @@
...
+
+// TxOptions specifies additional options that will be applied to the Tx.
+// Implements `Option` interface.
+type TxOptions struct {
+	// gasPrice represents the amount to be paid per gas unit.
+	// Negative gasPrice means user want us to use the minGasPrice
+	// defined in the node.
+	gasPrice float64
+	// since gasPrice can be 0, it is necessary to understand that user explicitly set it.
+	isGasPriceSet bool
```
cmwaters, line 38: I'm guessing it's either negative`gasPrice` or `isGasPriceSet==false` which indicates automatic estimating of gas price

cmwaters (approved):
I'm happy with how this looks, just have a few minor comments

state/tx_options.go
vgonkivs: `isGasPriceSet` helps to detect whether the user explicitly sets 0 or instantiates an empty `TxOptions{}` struct. I agree that comments look confusing. Will change them accordingly

state/tx_options.go
```diff
@@ -0,0 +1,168 @@
...
+	return rec.GetAddress()
+}
+
+func parseAccAddressFromString(addrStr string) (sdktypes.AccAddress, error) {
+	return sdktypes.AccAddressFromBech32(addrStr)
+}
+
+// Attribute is the functional option that is applied to the TxOption instance
+// to configure parameters.
+type Attribute func(options *TxOptions)
```
renaynay, line 137: Why not just Option?

state/tx_options.go
```diff
@@ -0,0 +1,168 @@
...
+	gasPrice float64
+	// since gasPrice can be 0, it is necessary to understand that user explicitly set it.
+	isGasPriceSet bool
+	// 0 gas means users want us to calculate it for them.
+	gas uint64
+
+	// Specifies the key from the keystore associated with an account that
+	// will be used to sign transactions.
+	// NOTE: This `Account` must be available in the `Keystore`.
+	accountKey string
```
liamsi, line 45: Sorry for chiming in so late on that but in my opinion it should (also) be possible to specificy the key by account address. Rationale: account addr is already a unique identifier which will always work no matter how the keys are stored or not. If people feel strongly about this, we can keep it as is.

nodebuilder/state/cmd/state.go
```diff
@@ -11,7 +11,17 @@ import (
...
+var (
+	amount uint64
+
+	gasPrice float64
+
+	gas uint64
+
+	accountKey string
+
+	feeGranterAddress string
```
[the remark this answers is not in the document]
vgonkivs: yes

state/tx_options.go
vgonkivs: I agree with you. But from the user perspective imo it's easier to specify a key rather than deal with the address(imagine the user has 5-10 addresses). I'm fine with adding an alternative option `signerAddress` where users can specify the exact address. But let's agree on some cross cases:
in case the user provides both key + address and they do not match. We can:
1) return some misconfiguration error;
2) assume the address is a primary option;

cc @distractedm1nd

nodebuilder/blob/cmd/blob.go
```diff
@@ -32,25 +31,19 @@ func init() {
 		&base64Flag,
 		"base64",
 		false,
-		"printed blob's data a base64 string",
+		"Printed blob's data a base64 string",
```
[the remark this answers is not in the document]
vgonkivs: 🚔

vgonkivs:
ok, so I'm adding an AccountAddress option that allows to specify the signerAddress directly. In case both key and address were passed and they do not match, we return an error. any objections?
cc @liamsi @renaynay

renaynay:
> ok, so I'm adding an AccountAddress option that allows to specify the signerAddress directly. In case both key and address were passed and they do not match, we return an error. any objections? cc @liamsi @renaynay

I'm fine with this @vgonkivs , I think that's a good option. Can you grab the key from the keyring by its address?

vgonkivs:
> Can you grab the key from the keyring by its address?

I will check. This is relevant only if the user passes both key and address. Otherwise, it's not necessary to check the address in the keysotre, bc the request will anyway fail.

state/core_access.go
```diff
@@ -419,157 +379,130 @@ func (ca *CoreAccessor) SubmitTxWithBroadcastMode(
...
 	}
 
-	coins := sdktypes.NewCoin(app.BondDenom, amount)
-	msg := stakingtypes.NewMsgCancelUnbondingDelegation(signer.Address(), valAddr, height.Int64(), coins)
-	if gasLim == 0 {
-		var err error
-		gasLim, err = signer.EstimateGas(ctx, []sdktypes.Msg{msg})
+	signer := ca.defaultSignerAddress
+	var err error
+	if options.AccountKey() != "" {
```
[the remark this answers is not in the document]
vgonkivs: reworked

vgonkivs:
>  In case both key and address were passed and they do not match, we return an error.

Let's just use an address in case both were passed.

cristaloleg (approved):
SGTM !

state/tx_options.go
Wondertan: I would rename `TxOptions` then to something like `TxConfig/Params`

state/tx_options.go
```diff
@@ -0,0 +1,190 @@
...
+	apptypes "github.com/celestiaorg/celestia-app/x/blob/types"
+)
+
+const (
+	DefaultPrice float64 = -1.0
+	// gasMultiplier is used to increase gas limit in case if tx has additional options.
+	gasMultiplier = 1.1
+)
+
+func NewTxOptions(attributes ...Attribute) *TxOptions {
```
Wondertan: A little godoc wouldn't hurt

state/tx_options.go
```diff
@@ -0,0 +1,190 @@
...
+
+	"github.com/cosmos/cosmos-sdk/crypto/keyring"
+	sdktypes "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/celestiaorg/celestia-app/pkg/user"
+	apptypes "github.com/celestiaorg/celestia-app/x/blob/types"
+)
+
+const (
+	DefaultPrice float64 = -1.0
```
Wondertan: godoc, especially since it public

nodebuilder/state/state.go
```diff
@@ -215,52 +197,46 @@ func (api *API) SubmitTx(ctx context.Context, tx state.Tx) (*state.TxResponse, e
 
 func (api *API) SubmitPayForBlob(
 	ctx context.Context,
-	fee state.Int,
-	gasLim uint64,
-	blobs []*blob.Blob,
+	blobs []*state.Blob,
```
Wondertan, line 200: Specifying a list of things we break in this PR would be nice. This one is minor but very breaking. Some teams rely on this endpoint isntead of blob.Submit

state/tx_options.go
vgonkivs: let's call it `TxConfig`

blob/service.go
```diff
@@ -71,39 +57,22 @@ func NewService(
...
-		GasLimit: 0,
-	}
-}
-
 // Submit sends PFB transaction and reports the height at which it was included.
 // Allows sending multiple Blobs atomically synchronously.
 // Uses default wallet registered on the Node.
 // Handles gas estimation and fee calculation.
-func (s *Service) Submit(ctx context.Context, blobs []*Blob, gasPrice GasPrice) (uint64, error) {
+func (s *Service) Submit(ctx context.Context, blobs []*Blob, txOptions *state.TxOptions) (uint64, error) {
```
Wondertan: A minor suggestion. It would be cool not to depend on the `state` pkg here directly but to define `SubmitOptions` that embed `TxOptions` to have cleaner boundaries.

nodebuilder/state/keyring.go
```diff
@@ -16,15 +18,20 @@ type AccountName string
...
-	// if custom keyringAccName provided, find key for that name
-	if cfg.KeyringAccName != "" {
-		keyInfo, err := ring.Key(cfg.KeyringAccName)
+
+	// go through all keys in the config and check their availability in the KeyStore.
+	for _, accName := range cfg.KeyringKeyNames {
+		keyInfo, err := ring.Key(accName)
 		if err != nil {
-			log.Errorw("failed to find key by given name", "keyring.accname", cfg.KeyringAccName)
+			err = fmt.Errorf("key not found in keystore: %s", accName)
```
Wondertan, line 26: [blocking]
This might be a different error, not necessarily not found. We should wrap the inner one instead of replacing it. This is a little foot gun that may complicate debugging if something goes wrong with the keyring, especially considering there are multiple backends for them, e.g., Ledger wallet.

nodebuilder/state/config.go
```diff
@@ -2,25 +2,21 @@ package state
...
 
 var defaultKeyringBackend = keyring.BackendTest
 
 // Config contains configuration parameters for constructing
 // the node's keyring signer.
 type Config struct {
-	KeyringAccName string
-	KeyringBackend string
-	GranterAddress state.AccAddress
+	KeyringKeyNames []string
```
Wondertan, line 12: What's the goal of this? Is that a filter specifying a list of keys the API can use? Or what would be the reason for the user to add new key names here?

Wondertan: From the code it looks like we get the first one and ignore all the others. What's the idea?

Wondertan (commented):
(I didn't review comments, flag and comments description)

nodebuilder/state/config.go
vgonkivs: `// go through all keys in the config and check their availability in the KeyStore.`

state/tx_options.go
Wondertan: (note that's the reason to rename Attribute to more conventional Option)

nodebuilder/state/state.go
vgonkivs: I will modify PR description

nodebuilder/state/config.go
vgonkivs: I realized that we don't need a slice of keys. KeyringAccName should only specify the default signer. Will return everything back. Thanks

state/tx_config.go
```diff
@@ -0,0 +1,188 @@
...
+	return rec.GetAddress()
+}
+
+func parseAccAddressFromString(addrStr string) (sdktypes.AccAddress, error) {
+	return sdktypes.AccAddressFromBech32(addrStr)
+}
+
+// Attribute is the functional option that is applied to the TxOption instance
+// to configure parameters.
+type Attribute func(cfg *TxConfig)
```
Wondertan, line 144: If we have TxConfig, we can call it as Option conventially

nodebuilder/state/keyring.go
```diff
@@ -15,24 +15,10 @@ type AccountName string
...
-		keyInfo, err := ring.Key(DefaultAccountName)
-		if err != nil {
-			log.Errorw("could not access key in keyring", "name", DefaultAccountName)
-			return nil, "", err
-		}
-		info = keyInfo
+	keyInfo, err := ring.Key(cfg.KeyringKeyName)
+	if err != nil {
+		log.Errorw("could not access key in keyring", "keyring.keyname", cfg.KeyringKeyName)
+		return nil, "", err
```
Wondertan, line 21: You could still wrap it to add some context. You could make a new error to return and also pass it the log

nodebuilder/state/config.go
```diff
@@ -2,25 +2,21 @@ package state
...
-	"github.com/celestiaorg/celestia-node/state"
 )
 
 var defaultKeyringBackend = keyring.BackendTest
 
 // Config contains configuration parameters for constructing
 // the node's keyring signer.
 type Config struct {
-	KeyringAccName string
+	KeyringKeyName string
```
Wondertan, line 12: We may add prefix it with Default to imply others can be used as well

state/tx_config.go
cristaloleg: https://github.com/celestiaorg/celestia-node/pull/3349#discussion_r1653022544 ?

state/tx_config.go
```diff
@@ -0,0 +1,188 @@
...
+func NewTxConfig(attributes ...Attribute) *TxConfig {
+	options := &TxConfig{gasPrice: DefaultGasPrice}
+	for _, attr := range attributes {
+		attr(options)
+	}
+	return options
+}
+
+// TxConfig specifies additional options that will be applied to the Tx.
+// Implements `Option` interface.
```
ramin, line 35: what Option interface?

ramin (commented):
couple of things:

Options vs Config seem muddled and not clean, i see there's been a bunch of discussion but def feels unclear.

The stuff around signer, granter, keyName is difficult to understand. I guess we can assume that someone has read some docs and understands all these things, but i wonder if some of the cmd flag comments should be more explicit about where eachis sources, keystore, vs grant module, vs the new signer. It's a bit vague and i'm not sure what is what

state/tx_config.go
vgonkivs: ooops, it is a relic comment. Thanks 🙏

vgonkivs:
> Options vs Config

`Options` in our codebase is a functional pattern that allows you to pass any amount of options. It helps to avoid breaking changes in the future. In my opinion, the term Option(s) can't be applied to the entity that holds parameters, applied to the transactions. This is the main reason why I renamed it to the `TxConfig`. 

> If we have TxConfig, we can call it an Option conventially

I don't mind. Previously, it was named TxOption and It does not look good to me(Options applied to TxOptions 🙂).

nodebuilder/state/state.go
```diff
@@ -202,11 +185,10 @@ func (api *API) BalanceForAddress(ctx context.Context, addr state.Address) (*sta
 func (api *API) Transfer(
 	ctx context.Context,
 	to state.AccAddress,
-	amount,
-	fee state.Int,
-	gasLimit uint64,
+	amount state.Int,
+	options *state.TxConfig,
```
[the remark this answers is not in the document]
vgonkivs: True. I've mistakenly left it after renaming to TxConfig

nodebuilder/state/config.go
renaynay: `DefaultKeyName`

nodebuilder/state/flags.go
```diff
@@ -8,26 +8,25 @@ import (
...
 
 // Flags gives a set of hardcoded State flags.
 func Flags() *flag.FlagSet {
 	flags := &flag.FlagSet{}
 
-	flags.String(keyringAccNameFlag, "", "Directs node's keyring signer to use the key prefixed with the "+
-		"given string.")
-	flags.String(keyringBackendFlag, defaultKeyringBackend, fmt.Sprintf("Directs node's keyring signer to use the given "+
-		"backend. Default is %s.", defaultKeyringBackend))
+	flags.String(keyringKeyNameFlag, DefaultAccountName,
```
renaynay, line 19: let's call the var `DefaultKeyName`

renaynay (approved):

nodebuilder/state/config.go
vgonkivs: And `DefaultBackendName` for consistency

state/tx_config.go
Wondertan: I am aware @cristaloleg. The reasoning for not calling them Options was that making options for TxOptions is a tautology. Now it's TxConfig, so we should call Attributes as Options, as the original reason doesn't hold anymore.

vgonkivs: CoreAccessor already has declared Option type in state package, so we should either leave as `Attribute` or rename to `CfgOption`.

Wondertan: ConfigOption sgtm

Wondertan (approved):

ramin (approved):

Wondertan (approved):

renaynay (approved):

walldiss (approved):

distractedm1nd (approved):
LFG

Wondertan (approved):

liamsi (approved):
utACK

state/tx_config.go
```diff
@@ -0,0 +1,187 @@
...
+	// signerAddress is a primary cfg. This means If both the address and the key are specified,
+	// the address field will take priority.
+	signerAddress string
+	// Specifies the key from the keystore associated with an account that
+	// will be used to sign transactions.
+	// NOTE: This `Account` must be available in the `Keystore`.
+	keyName string
+	// gasPrice represents the amount to be paid per gas unit.
+	// Negative gasPrice means user want us to use the minGasPrice
+	// defined in the node.
```
rootulp, line 47: nit: minGasPrice is a bit ambiguous because in celestia-app version >= 2, consensus nodes have a local min gas price. There is also a global network min gas price which is a governance modifiable parameter so perhaps it is worth clarifying which gas price this defaults to: the local min gas price or the network min gas price.

PR #3350, opened 2024-04-30
ci: generate and store openrpc.json with the tagged release artifacts

ramin:
Adds a step in the ci_and_release pipeline to generate `openrpc.json` using `make openrpc-gen` and store that generated file with the release artifacts. This is set to run after go-releaser, though we could have made this its own workflow file potentially. 

I have tested that release upload works in a personal repo, but as with all things on github, we can only REALLY test and do this one live 🤞 

refs #2611


renaynay (approved):
utack

is there an accompanying PR from docs repo to point to this artifact?

ramin:
> utack
> 
> is there an accompanying PR from docs repo to point to this artifact?

i defer to @jcstein on what his plan is/was for updating this with each release once it starts getting generated

jcstein:
> is there an accompanying PR from docs repo to point to this artifact?

I will make this PR once this exists on a release

jcstein (approved):
utACK

renaynay (approved):
utack

walldiss (approved):

Issue #2439, opened 2023-07-06
Support SSL for connecting to consensus node API with --core.ip


musalbas:

JB273:
What we know works already:
For celestia-appd version 0.11.1 (mocha):
`celestia-appd query bank balances celestia1g486scr5t24jywkql8ld2uuw6clws2fckclhlj --node https://rpc-mocha.pops.one:443`

JB273:
It looks like the node client expects to be able to connect to both RPC and gRPC on the same hostname.
To add SSL with a proxy many are now using different hostnames and forward to the correct port, eg:

- `https://rpc-mocha.pops.one:443` forwards to local port 26657
- `https://grpc-mocha.pops.one` forwards to local port 9090

Currently we can only give one hostname through the `--core-ip` flag, this looks like to be a part of the issue

malise800:
From I could figure, I think the celestia-node CoreAccessor client doesn't support SSL. 

https://github.com/celestiaorg/celestia-node/blob/main/state/core_access.go#L85-L112

malise800:
now it also seems the RPC node provider will need to update their infrastructure to also support proxied grpc connection : https://www.nginx.com/blog/deploying-nginx-plus-as-an-api-gateway-part-3-publishing-grpc-services/

ramin:
closing in favor of https://github.com/celestiaorg/celestia-node/issues/2931

PR #3359, opened 2024-05-02
refactor(state): Remove need for core RPC conn and rely only on gRPC for state reads

### LINKED ISSUE #3357: improvement(state | nodebuilder/state): Use ABCIQuery over gRPC
This would remove the need for specifying a `--core.rpc.port` in addition to `--core.grpc.port` by using abci query over gRPC instead of over RPC

TODO

- [ ] Use https://github.com/celestiaorg/cosmos-sdk/blob/02244ae2a73e053234ca2de6e68a97974edb952a/client/grpc/tmservice/service.go#L221 in `CoreAccessor`
- [ ] remove requirement for `--core.rpc.port` for state module construction

renaynay:
Resolves #3357 

Blocked on patch release from app including https://github.com/celestiaorg/celestia-app/pull/3424

Corresponding docs PR https://github.com/celestiaorg/docs/pull/1558


state/core_access.go
```diff
@@ -367,8 +355,21 @@ func (ca *CoreAccessor) BalanceForAddress(ctx context.Context, addr Address) (*B
 		return nil, fmt.Errorf("cannot convert %s into sdktypes.Int", string(value))
 	}
 	// verify balance
+	var proofOps *crypto.ProofOps
```
renaynay, line 358: @tac0turtle anything prettier available here or do we just have to do this manually?

state/core_access.go
```diff
@@ -367,8 +355,21 @@ func (ca *CoreAccessor) BalanceForAddress(ctx context.Context, addr Address) (*B
...
+			Ops: make([]crypto.ProofOp, len(result.ProofOps.Ops)),
+		}
+		for i, proofOp := range result.ProofOps.Ops {
+			proofOps.Ops[i] = crypto.ProofOp{
+				Type: proofOp.Type,
+				Key:  proofOp.Key,
+				Data: proofOp.Data,
+			}
+		}
+	}
```
liamsi, line 370: Is there a test for this query or did you manually test it?

renaynay: There's a test in the integration suite that tests the `GetBalance` endpoint yes.

state/core_access.go
```diff
@@ -334,26 +327,21 @@ func (ca *CoreAccessor) BalanceForAddress(ctx context.Context, addr Address) (*B
...
 	}
-	opts := rpcclient.ABCIQueryOptions{
-		Height: abciReq.GetHeight(),
-		Prove:  abciReq.GetProve(),
-	}
-	result, err := ca.rpcCli.ABCIQueryWithOptions(ctx, abciReq.GetPath(), abciReq.GetData(), opts)
-	if err != nil {
+
+	result, err := ca.abciQueryCli.ABCIQuery(ctx, req)
+	if err != nil || result.Code != 0 {
```
Wondertan, line 339: If we hit result.Code != 0, the err might be nil, returning both values nil, when there is an error.
Is there err field on the result struct?

state/core_access.go
```diff
@@ -367,8 +355,21 @@ func (ca *CoreAccessor) BalanceForAddress(ctx context.Context, addr Address) (*B
 		return nil, fmt.Errorf("cannot convert %s into sdktypes.Int", string(value))
 	}
 	// verify balance
+	var proofOps *crypto.ProofOps
+	if result.GetProofOps() != nil {
```
Wondertan, line 359: So, is there a case where the server gives us no proof? Shouldn't we disallow such a case?

state/core_access.go
```diff
@@ -367,8 +355,21 @@ func (ca *CoreAccessor) BalanceForAddress(ctx context.Context, addr Address) (*B
...
+		proofOps = &crypto.ProofOps{
+			Ops: make([]crypto.ProofOp, len(result.ProofOps.Ops)),
+		}
+		for i, proofOp := range result.ProofOps.Ops {
+			proofOps.Ops[i] = crypto.ProofOp{
+				Type: proofOp.Type,
+				Key:  proofOp.Key,
+				Data: proofOp.Data,
+			}
+		}
```
Wondertan, line 369: Do we need this to convert one time of proof to another?

state/core_access.go
```diff
@@ -367,8 +355,21 @@ func (ca *CoreAccessor) BalanceForAddress(ctx context.Context, addr Address) (*B
...
+	}
 	err = ca.prt.VerifyValueFromKeys(
-		result.Response.GetProofOps(),
+		proofOps,
```
Wondertan, line 372: If `proofOps == nil` (possible with the if check above), how does this method behaves?

state/core_access.go
renaynay: Yes thanks to cosmos-sdk copying over types from tendermint rather than just using them directly

state/core_access.go
```diff
@@ -334,26 +327,22 @@ func (ca *CoreAccessor) BalanceForAddress(ctx context.Context, addr Address) (*B
...
-	opts := rpcclient.ABCIQueryOptions{
-		Height: abciReq.GetHeight(),
-		Prove:  abciReq.GetProve(),
-	}
-	result, err := ca.rpcCli.ABCIQueryWithOptions(ctx, abciReq.GetPath(), abciReq.GetData(), opts)
-	if err != nil {
+
+	result, err := ca.abciQueryCli.ABCIQuery(ctx, req)
+	if err != nil || result.Code != 0 {
+		err = fmt.Errorf("failed to query for balance: %w; result log: %s", err, result.Log)
```
Wondertan, line 340: Optional: add code to the erro

Wondertan (approved):

liamsi:
Is it in scope of this Pr to ensure TLS encryption works for gRPC? And for bridge nodes TLS for RPC to core? If not I assume this will be tackled shortly after?

vgonkivs (approved):

Wondertan:
> Is it in scope of this Pr to ensure TLS encryption works for gRPC? And for bridge nodes TLS for RPC to core? If not I assume this will be tackled shortly after?

It's good to have if for privacy reasons so that no MITM can see the addresses and balances you request. Worth having an issue

Wondertan (approved):
LGTM otherwise

liamsi:
> It's good to have if for privacy reasons so that no MITM can see the addresses and balances you request. Worth having an issue

There was an over 10 months old issue and it was closed... Rene opened a new one https://github.com/celestiaorg/celestia-node/issues/3377

state/core_access.go
```diff
@@ -136,7 +133,6 @@ func NewCoreAccessor(
 		addr:     addr,
 		getter:   getter,
 		coreIP:   coreIP,
```
liamsi, line 135: Can't core be accesses by URL? Does it have to be by IP? I wonder how this would work with TLS (where certs will be tied to a domain / URL).

liamsi (approved):

Wondertan:
> https://github.com/celestiaorg/celestia-node/issues/2439#issuecomment-1851729638

🤔

PR #3378, opened 2024-05-08
chore(nodebuilder/das | pruner): Privatise `samplingWindow` in DAS params, add Duration method to `AvailabilityWindow`

renaynay:
While this isn't really a bug fix, it's more of a "UX" fix - I'm removing `SamplingWindow` from the node's config so that it doesn't confuse users to see "0ms" as the config value even though in practice it's 30 days. 

The reason there's a discrepancy is because we force the light sampling window in the DASer for light nodes in nodebuilder anyway so that config value is ignored regardless, meaning it does not need to be exposed.

Also adding `Duration()` method to return time.Duration instead of having to cast everywhere.


pruner/window.go
```diff
@@ -6,12 +6,16 @@ import (
 
 type AvailabilityWindow time.Duration
 
+func (aw AvailabilityWindow) Duration() time.Duration {
+	return time.Duration(aw)
+}
```
Wondertan, line 11: q: Is there a reason to make AvailabilityWindow a new type or more like a feeling? I think that could be an alias that wouldn't need the method

nodebuilder/das/config.go
```diff
@@ -23,6 +24,7 @@ func DefaultConfig(tp node.Type) Config {
 	switch tp {
 	case node.Light:
 		cfg.SampleTimeout = modp2p.BlockTime * time.Duration(cfg.ConcurrencyLimit)
+		cfg.SamplingWindow = light.Window.Duration()
```
Wondertan, line 27: Do you wanna do the same for FN?

pruner/find.go
```diff
@@ -18,7 +18,7 @@ func (s *Service) findPruneableHeaders(
 	ctx context.Context,
 	lastPruned *header.ExtendedHeader,
 ) ([]*header.ExtendedHeader, error) {
-	pruneCutoff := time.Now().UTC().Add(time.Duration(-s.window))
+	pruneCutoff := time.Now().UTC().Add(-s.window.Duration())
```
ramin, line 21: They added a `.Sub` that can be easier to process in your head vs Add on a negative number if you fancy that

ramin (approved):

nodebuilder/das/config.go
walldiss: We should set sampling window for FN/BN to infinity, otherwise it will be 0

walldiss, line 27: Would it be possible to make DASer config field SamplingWindow to be type of pruner.AvailabilityWindow? It will ensure type safety and remove the need for conversions

walldiss:
I have a couple of suggestions that could simplify the implementation:
1. Could we incorporate the default value for the `SamplingWindow` directly within the `[DefaultParameters()](https://github.com/celestiaorg/celestia-node/blob/c04bbd251e7947180aadba793255e16e3126ef56/das/options.go#L56-L63)` function? Could be just one-line addition.

2. Given that the `SamplingWindow` value is always overwritten in our current setup, exposing this setting to users seems unnecessary and potentially misleading. We should consider making `SamplingWindow` config field non-public. This approach would help prevent any confusion and ensure our configuration remains streamlined and intuitive.

pruner/window.go
renaynay: Can address in a separate PR

das/options.go
```diff
@@ -42,10 +42,10 @@ type Parameters struct {
 	// ConcurrencyLimit.
 	SampleTimeout time.Duration
 
-	// SamplingWindow determines the time window that headers should fall into
+	// samplingWindow determines the time window that headers should fall into
 	// in order to be sampled. If set to 0, the sampling window will include
 	// all headers.
-	SamplingWindow time.Duration
+	samplingWindow time.Duration
```
walldiss, line 48: Perhaps changing samplingWindow type from time.Duration to pruner.AvailabilityWindow will simplify code a lot. I assume it was time.duration to avoid importing `pruner` as a dependency. But the dependancy inevitable, considering DASer has to rely on pruner logic to skip pruned heights. Using native type for window here can remove the need to cast it everywhere and for cast method itself.

renaynay: just did

renaynay: also in archival PR, daser will use pruner convenience method to check recency instead of its own

walldiss (approved):

vgonkivs (approved):

PR #3380, opened 2024-05-08
fix(das): Clarify logs for historic headers that are skipped

renaynay:
Fix is kind of ugly but there needs to be a way to bubble up to worker that the header is skipped vs being sampled successfully.

Resolves https://github.com/celestiaorg/celestia-node/issues/3194


das/daser.go
```diff
@@ -21,6 +21,11 @@ import (
 
 var log = logging.Logger("das")
 
+// errOutsideSamplingWindow is an error used to inform
+// the caller of Sample that the given header is outside
+// the sampling window.
+var errOutsideSamplingWindow = fmt.Errorf("skipping header outside of sampling window")
```
ramin, line 27: curious why not `errors.New` if no formatting?

ramin (approved):

walldiss (approved):

das/daser.go
jcstein, line 27: :nice: 😄

PR #3387, opened 2024-05-10
testing(state/core_accessor): add async submitPFB test case

vgonkivs:
Additional test case checks whether tx was included in the block after async submission

### DIFF
--- a/state/core_access_test.go
+++ b/state/core_access_test.go
@@ -11,13 +11,12 @@ import (
 	"time"
 
 	"cosmossdk.io/math"
-	sdktypes "github.com/cosmos/cosmos-sdk/types"
-	"github.com/stretchr/testify/require"
-
 	"github.com/celestiaorg/celestia-app/app"
 	"github.com/celestiaorg/celestia-app/pkg/appconsts"
 	"github.com/celestiaorg/celestia-app/test/util/testnode"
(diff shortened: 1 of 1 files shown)


ramin (approved):

walldiss (changes requested):
Lets improve the test, so in case when it breaks it is easier to identify the root cause.

state/core_access_test.go
```diff
@@ -89,6 +92,27 @@ func TestSubmitPayForBlob(t *testing.T) {
...
+				}()
+
+				ctx, cancel := context.WithTimeout(ctx, time.Second*3)
+				resp, err := ca.signer.ConfirmTx(ctx, <-hashCh)
+				cancel()
+				require.NoError(t, err)
+				require.EqualValues(t, 0, resp.Code)
+			})
+		}
+	}
```
walldiss, line 115: If `ca.SubmitPayForBlob` returns non nill error, the test will hang forever because of deadlock on reading from `<-hashCh` and also no meaningful error will be logged or returned. To fix this lets do:
- use select with `ctx.Done` for reading from `<-hashCh` so test can fail fast
- use `reqiure.NoError` for error produced by `ca.SubmitPayForBlob` to know what happen if it failed.

vgonkivs:
N/A

PR #3348, opened 2024-04-30
refactoring(cmd/state): remove submit-tx from cli

vgonkivs:
Addresses a TODO in #3345

Issue #3345, opened 2024-04-29
CLI/API: very unclear how to use `state submit-tx`


liamsi:
how is the user supposed to sign the tx in question? even weirder the tx seems to be expected to be [hex encoded](https://github.com/celestiaorg/celestia-node/blob/332739f9e68196333bd1dea76fbeaac0f12737f8/nodebuilder/state/cmd/state.go#L158-L165)? even if one used app or some code to generate and sign the tx, it is unclear why then go through the hassle to then submit it via node.

Bare minimum is that:
- [ ] someone familiar with node / celestia can look at docs and will know exactly how to create a Tx and sign to submit via SubmitTx (all encoding questions resolved + code snippet)

Optional/Nicetohave:
- [ ] clarify in the docs to the spec so someone who wants to do this e.g. in Rust has all info how to do this or add in another end-point for generating and signing (needs further discussion)

Additionally:
- [ ] consider removing submitx from the CLI (this does not seem useful as is right now)

renaynay:
@distractedm1nd will take care of updating docs repo with a snippet of how to use `SubmitTx` endpoint. Thank you

distractedm1nd:
To use SubmitTx programmatically, you ofc have to import tons of cosmossdk/app stuff, so at that point it doesn’t really make sense to use node at all…. 
I think it makes the most sense to have the sensible tx types as available endpoints, but maybe remove `SubmitTx` alltogether, even via the RPC.

I actually think our bases are covered here. We support the following transaction types in node:

- Transfer
- SubmitPayForBlob
- CancelUnbondingDelegation
- BeginRedelegate
- Undelegate
- Delegate
- GrantFee
- RevokeGrantFee

So, I would suggest deprecating `SubmitTx` endpoint and focusing on documenting how to use the transaction specific endpoints. The snippet for SubmitTx will look very similar to [this.](https://docs.celestia.org/developers/submit-data#api)

Whereas a snippet for a `Transfer` or `SubmitPayForBlob` would just be a few lines, way shorter than the above example.

distractedm1nd:
@Bidon15 is asking a few teams and will report back to me, but I am already confident that this endpoint is useless and would already prepare the PR in advance.

Wondertan:
> That was my exact reaction on this endpoint as well. @Wondertan and @renaynay kept mentioning that this was mainly a docs issue which might me doubt my intuition on this endpoint a bit. 

[...]

PR #3391, opened 2024-05-10
refactor(state/gateway)!: remove SubmitTx endpoint from RPC and Gateway

distractedm1nd:
This PR removes the `state.SubmitTx` endpoint from the RPC, and the `submit_tx` endpoint from the gateway. 

Related: #3348 
## Rationale
This endpoint was never fully documented, and no usage examples exist. 
When creating usage examples after a discussion in #3345 , we realized the endpoint is not very useful at all. In order to use it, you need to construct a transaction, importing app, tendermint... all to then proxy it to an app instance via node anyways. Here is a snippet from @jcstein in the docs tutorial for submitting data.

The example for `celestia-openrpc`/`celestia-node` would essentially be exactly the same, except we build a rpc connection to node instead of a grpc to app. I think it is clear from this example why SubmitTx is too much of a hassle to use, and if users do need this functionality for their use-case, it is likely a sign that they should be interacting with app directly instead anyways.

To verify this, @Bidon15 reached out to teams to get feedback on this decision. All teams confirmed that the endpoint is not being used and not needed.

We support enough transaction types in the state module. If users need more functionality, we should either add the granularity to the current methods or add specific methods for specific transaction types/use cases. Dropping down a level to SubmitTx should never be needed if our API is good enough.

### Example (using app conn)
```go
import (
    "context"
    "fmt"

    "github.com/celestiaorg/celestia-app/app"
    "github.com/celestiaorg/celestia-app/app/encoding"
    "github.com/celestiaorg/celestia-app/pkg/appconsts"
    "github.com/celestiaorg/celestia-app/pkg/namespace"
    "github.com/celestiaorg/celestia-app/pkg/user"
    blobtypes "github.com/celestiaorg/celestia-app/x/blob/types"
    "github.com/cosmos/cosmos-sdk/crypto/keyring"
    tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
    "google.golang.org/grpc"
    "google.golang.org/grpc/credentials/insecure"
)

// SubmitData is a demo function that shows how to use the signer to submit data
// to the blockchain directly via a celestia node. We can manage this keyring
// using the `celestia-appd keys` or `celestia keys` sub commands and load this
// keyring from a file and use it to programmatically sign transactions.
func DemoSubmitData(grpcAddr string, kr keyring.Keyring) error {
    // create an encoding config that can decode and encode all celestia-app
    // data structures.
    ecfg := encoding.MakeConfig(app.ModuleEncodingRegisters...)

    // create a connection to the grpc server on the consensus node.
    conn, err := grpc.Dial(grpcAddr, grpc.WithTransportCredentials(insecure.NewCredentials()))
    if err != nil {
        return err
    }
    defer conn.Close()

    // get the address of the account we want to use to sign transactions.
    rec, err := kr.Key("accountName")
    if err != nil {
        return err
    }

    addr, err := rec.GetAddress()
    if err != nil {
        return err
    }

    // Setup the signer. This function will automatically query the relevant
    // account information such as sequence (nonce) and account number.
    signer, err := user.SetupSigner(context.TODO(), kr, conn, addr, ecfg)
    if err != nil {
        return err
    }

    ns := namespace.MustNewV0([]byte("1234567890"))

    fmt.Println("namespace", len(ns.Bytes()))

    blob, err := blobtypes.NewBlob(ns, []byte("some data"), appconsts.ShareVersionZero)
    if err != nil {
        return err
    }

    gasLimit := blobtypes.DefaultEstimateGas([]uint32{uint32(len(blob.Data))})

    options := []user.TxOption{
        // here we're setting estimating the gas limit from the above estimated
        // function, and then setting the gas price to 0.1utia per unit of gas.
        user.SetGasLimitAndFee(gasLimit, 0.1),
    }

    // this function will submit the transaction and block until a timeout is
    // reached or the transaction is committed.
    resp, err := signer.SubmitPayForBlob(context.TODO(), []*tmproto.Blob{blob}, options...)
    if err != nil {
        return err
    }

    // check the response code to see if the transaction was successful.
    if resp.Code != 0 {
        // handle code
        fmt.Println(resp.Code, resp.Codespace, resp.RawLog)
    }

    // if we don't want to wait for the transaction to be confirmed, we can
    // manually sign and submit the transaction using the same package.
    blobTx, err := signer.CreatePayForBlob([]*tmproto.Blob{blob}, options...)
    if err != nil {
        return err
    }

    resp, err = signer.BroadcastTx(context.TODO(), blobTx)
    if err != nil {
        return err
    }

    // check the response code to see if the transaction was successful. Note
    // that this time we're not waiting for the transaction to be committed.
    // Therefore the code here is only from the consensus node's mempool.
    if resp.Code != 0 {
        // handle code
        fmt.Println(resp.Code, resp.Codespace, resp.RawLog)
    }

    return err
}
```

### DIFF
--- a/api/gateway/bindings.go
+++ b/api/gateway/bindings.go
@@ -14,12 +14,6 @@ func (h *Handler) RegisterEndpoints(rpc *Server) {
 	)
 
 	rpc.RegisterHandlerFunc(
-		submitTxEndpoint,
-		h.handleSubmitTx,
-		http.MethodPost,
-	)
-
-	rpc.RegisterHandlerFunc(
 		healthEndpoint,
 		h.handleHealthRequest,
 		http.MethodGet,
--- a/api/gateway/bindings_test.go
+++ b/api/gateway/bindings_test.go
@@ -28,12 +28,6 @@ func TestRegisterEndpoints(t *testing.T) {
 			expected: true,
 		},
 		{
-			name:     "Submit transaction endpoint",
-			path:     submitTxEndpoint,
-			method:   http.MethodPost,
-			expected: true,
-		},
-		{
 			name:     "Get namespaced shares by height endpoint",
 			path:     fmt.Sprintf("%s/{%s}/height/{%s}", namespacedSharesEndpoint, namespaceKey, heightKey),
 			method:   http.MethodGet,
--- a/api/gateway/state.go
+++ b/api/gateway/state.go
@@ -1,7 +1,6 @@
 package gateway
 
 import (
-	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"net/http"
@@ -13,8 +12,7 @@ import (
 )
 
 const (
-	balanceEndpoint  = "/balance"
-	submitTxEndpoint = "/submit_tx"
+	balanceEndpoint = "/balance"
 )
 
 const addrKey = "address"
@@ -23,11 +21,6 @@ var (
 	ErrInvalidAddressFormat = errors.New("address must be a valid account or validator address")
 	ErrMissingAddress       = errors.New("address not specified")
 )
-
-// submitTxRequest represents a request to submit a raw transaction
-type submitTxRequest struct {
-	Tx string `json:"tx"`
-}
 
 func (h *Handler) handleBalanceRequest(w http.ResponseWriter, r *http.Request) {
 	var (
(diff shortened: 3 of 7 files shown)


walldiss (dismissed):
My fav type of PRs - deletions only! 🚀 Less is more, right?

ramin (approved):

vgonkivs (dismissed):

distractedm1nd:
@walldiss 
Slash It

vgonkivs (approved):

ramin (approved):

PR #3404, opened 2024-05-15
fix(blob/service): fix handling of the padding shares

vgonkivs:
Fixes the issue with the missing blob. Initially, there was an incorrect assumption that only one padding share is possible between two blobs. So, the Blob service was skipping only one share instead of multiple. 

* Fixed parsing logic allowing to skip multiple padding shares;
* Added a test case that retrieves the EDS and finds the correct number of blobs;


blob/service_test.go
```diff
@@ -492,6 +496,66 @@ func TestService_GetAllWithoutPadding(t *testing.T) {
 	}
 }
 
+// TestRetrieveBlobsFromRealEds takes the real data, submitted to mocha-4 and
+// ensures that all blobs under `0x51cc1a73d65f95f6bde6` namespace will be retrieved successfully.
+// At the first step it retrieves all blobs under the given namespace:
+// the amount of the blobs is known and equal to 6. Then it ensures that each blob has a correct index inside the eds
+// by requesting share and comparing them.
+func TestRetrieveBlobsFromRealEds(t *testing.T) {
+	f, err := os.OpenFile("example/eds.json", os.O_RDONLY, os.ModePerm)
```
Wondertan, line 505: Example name will be confusing. People will think thats an example on how to use blob module there. It should be called smth like testdata instead.

blob/service_test.go
```diff
@@ -492,6 +496,66 @@ func TestService_GetAllWithoutPadding(t *testing.T) {
...
+	t.Cleanup(cancel)
+
+	recomputedEds, err := ipld.AddShares(ctx, ods, bServ)
+	require.NoError(t, err)
+	h := headertest.ExtendedHeaderFromEDS(t, 1, recomputedEds)
+	fn := func(ctx context.Context, u uint64) (*header.ExtendedHeader, error) {
+		return h, nil
+	}
+
+	ipldGetter := getters.NewIPLDGetter(bServ)
```
Wondertan, line 528: The fact that we need ipld getter to unit test this suggest deeper architectural issue. We need to make an issue to decouple data retrieval from blob parsing. This might be tackled together with BlobV2

blob/service_test.go
Wondertan, line 505: The rule of thumb is to use embed directive for tests. It allows to make all the tests compiled in the binary instead of reading this in runtime.

Wondertan (changes requested):
The fix LGTM. Only have two blocking on the test file and need to make an issue.

walldiss (commented):
Lets make test, that generates edses with multiple conditions of padding shares instead of using eds.json. 1 padding share, multiple padding shares between blobs, multiple padding shares between namespaces.

blob/parser.go
```diff
@@ -117,19 +117,21 @@ func (p *parser) skipPadding(shares []shares.Share) ([]shares.Share, error) {
...
+	for _, sh := range shares {
+		isPadding, err := sh.IsPadding()
+		if err != nil {
+			return nil, err
+		}
+		if !isPadding {
+			break
+		}
+		offset++
 	}
```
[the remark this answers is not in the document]
vgonkivs: > What if all shares are padding?
It's ok. We'll remove them, update the index, and move on to the next row. 

The case that you've described is possible. Imagine we have a blob with the length `[0;odsSize-2]`. To satisfy the non-interactive default rules a padding share will be added as the last share for this row. So, during parsing, we detect it, skip, update the index of the next blob and move on to the next row(or finish parsing)

blob/service_test.go
vgonkivs: removed this test with json at in four of generated eds with multiple padding shares

blob/service_test.go
vgonkivs: we don't need it. `shareGetter` helps to resolve this

blob/service_test.go
vgonkivs: removed this test

blob/parser.go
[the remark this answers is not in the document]
walldiss: Good candidate for test imo

walldiss (dismissed):

blob/service_test.go
Wondertan: Elaborate how that helps

blob/parser.go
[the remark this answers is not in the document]
vgonkivs: `TestGetAllWithoutPadding` tests it

blob/service_test.go
vgonkivs: Helps with what? I don't get it. It's just an extra way to check that all shares are in their place. And I don't see anything criminal accessing it in unit tests.

blob/parser.go
```diff
@@ -117,19 +117,21 @@ func (p *parser) skipPadding(shares []shares.Share) ([]shares.Share, error) {
...
 
-	// update blob index if we are going to skip one share
-	p.index++
-	if len(shares) > 1 {
-		return shares[1:], nil
+	p.index += offset
```
renaynay, line 132: is it possible p.index + offset can overflow len shares?

blob/service_test.go
Wondertan: > shareGetter helps to resolve this

This is what you said and I ask how it helps

vgonkivs: It helps me to fetch shares by row/col index and compare with shares that I've parsed from blob

Wondertan: There is nothing criminal, but the separation of concerns is violated causing complexities with testing and code readability. Rootul couldn't figure out how to test blob parsing after working on that for an hour and that's a bad sign.

Wondertan:
@renaynay, I propose to make a new patch for this fix, instead of 0.14.0 as its blocking Astria and in general a critical bug.

Wondertan (approved):

blob/parser.go
vgonkivs: no, skipAdding is called when the index is 0 and the offset can't be more than len(shares)

blob/service_test.go
vgonkivs: https://github.com/celestiaorg/celestia-node/issues/3409

Wondertan (approved):

walldiss (approved):

PR #3405, opened 2024-05-15
testing(api): Adding API equivalency test for celestia-openrpc

distractedm1nd:
Adds a test to ensure API equivalence to celestia-openrpc. It will not be used or added to CI in this repo, hence the build tag. celestia-openrpc will pull this repo, build with the build tag, and run the test itself.

The reason it shouldnt be added to the CI in celestia-node is because it would fail every time we change the API. It shouldn't fail here, it should fail in the other repo.

### DIFF
--- a/api/rpc_library_test.go
+++ b/api/rpc_library_test.go
@@ -1,5 +1,4 @@
-//go:build ignore
-// +build ignore
+//go:build rpcequivalence
 
 package api
 
--- a/api/rpc_library_test.go
+++ b/api/rpc_library_test.go
@@ -1,0 +1,123 @@
+//go:build ignore
+// +build ignore
+
+package api
+
+import (
+	"reflect"
+	"testing"
+
+	blobA "github.com/celestiaorg/celestia-openrpc/types/blob"
+	daA "github.com/celestiaorg/celestia-openrpc/types/da"
+	dasA "github.com/celestiaorg/celestia-openrpc/types/das"
+	fraudA "github.com/celestiaorg/celestia-openrpc/types/fraud"
+	headerA "github.com/celestiaorg/celestia-openrpc/types/header"
+	nodeA "github.com/celestiaorg/celestia-openrpc/types/node"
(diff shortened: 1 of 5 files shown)


api/rpc_library_test.go
```diff
@@ -0,0 +1,122 @@
...
+		compareStructs(t, libraryStructs[i], apiStructs[i], moduleNames[i])
+	}
+}
+
+func compareStructs(t *testing.T, structA, structB interface{}, moduleName string) {
+	typeA := reflect.TypeOf(structA)
+	typeB := reflect.TypeOf(structB)
+
+	if typeA.NumField() != typeB.NumField() {
+		t.Fatalf("%s API has a comflicting number of fields: %d != %d", moduleName, typeA.NumField(), typeB.NumField())
```
ramin, line 78: spelling: conflicting

ramin (commented):
I think it'd be nice to start to organize these into a top level test folder and intro a more reusable build tag, maybe

`/tests/conformance/api` and the tag being `conformance` then we can run ALL conformance tests, or only a subset by running go test inside `/tests/conformance/<type of test>`

open to another name than conformance

api/rpc_library_test.go
distractedm1nd: thx

distractedm1nd:
I like the name conformance a lot more, I will change the build tag - but I would maybe hold off on moving the test until we have more conformance test cases? WDYT

renaynay (approved):
utack

walldiss:
I like the idea to have this test. But why test for external client implementation is in node repo?  It seems wrong that node will depend on that client after this PR

ramin:
yeah actually i agree with @walldiss, this should live in the `celestia-openrpc` repo and run as a conformance test there

distractedm1nd:
It can't, that's why its here :( I agree ofc but openrpc can't import node because thats the whole reason it exists

renaynay:
@distractedm1nd make a conformance test repo and put it there.

PR #3425, opened 2024-05-22
feat(share/eds) Add Accessor test infra

walldiss:
Adds generalised testing framework for file interface implementaions

Also adds non-inclusion proofs handling in RowNamespaceDataFromShares and closes https://github.com/celestiaorg/celestia-node/issues/3428

### DIFF
--- a/share/new_eds/rsmt2d.go
+++ b/share/new_eds/rsmt2d.go
@@ -87,7 +87,7 @@ func (eds Rsmt2D) RowNamespaceData(
 	return shwap.RowNamespaceDataFromShares(shares, namespace, rowIdx)
 }
 
-// Shares returns data shares extracted from the EDS. It returns new copy of the shares each
+// Shares returns data (ODS) shares extracted from the EDS. It returns new copy of the shares each
 // time.
 func (eds Rsmt2D) Shares(_ context.Context) ([]share.Share, error) {
 	return eds.ExtendedDataSquare.FlattenedODS(), nil
--- a/share/new_eds/rsmt2d.go
+++ b/share/new_eds/rsmt2d.go
@@ -90,7 +90,7 @@ func (eds Rsmt2D) RowNamespaceData(
 // Shares returns data shares extracted from the EDS. It returns new copy of the shares each
 // time.
 func (eds Rsmt2D) Shares(_ context.Context) ([]share.Share, error) {
-	return eds.ExtendedDataSquare.Flattened(), nil
+	return eds.ExtendedDataSquare.FlattenedODS(), nil
 }
 
 func getAxis(eds *rsmt2d.ExtendedDataSquare, axisType rsmt2d.Axis, axisIdx int) []share.Share {
--- a/share/new_eds/testing.go
+++ b/share/new_eds/testing.go
@@ -193,7 +193,7 @@ func TestAccessorAxisHalf(
 	})
 }
 
-func TestAccessorEds(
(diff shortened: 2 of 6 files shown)


cristaloleg (approved):
SGTM

Wondertan:
Linter complains

walldiss:
Linter complains, because it is older version of linter in ci. Fixed to satisfy ci linter

share/new_eds/rsmt2d.go
```diff
@@ -90,7 +90,7 @@ func (eds Rsmt2D) RowNamespaceData(
 // Shares returns data shares extracted from the EDS. It returns new copy of the shares each
 // time.
 func (eds Rsmt2D) Shares(_ context.Context) ([]share.Share, error) {
-	return eds.ExtendedDataSquare.Flattened(), nil
+	return eds.ExtendedDataSquare.FlattenedODS(), nil
```
Wondertan, line 93: The comment says its EDS shares. We may want to update the name of the method as well to say its ODS

share/new_eds/testing.go
```diff
@@ -0,0 +1,276 @@
...
+
+	"github.com/celestiaorg/celestia-node/share"
+	"github.com/celestiaorg/celestia-node/share/eds/edstest"
+	"github.com/celestiaorg/celestia-node/share/sharetest"
+	"github.com/celestiaorg/celestia-node/share/shwap"
+)
+
+type createAccessor func(eds *rsmt2d.ExtendedDataSquare) Accessor
+
+func TestAccessorSample(
```
Wondertan, line 23: Do we need these testing funcs exported?

share/shwap/row_namespace_data.go
```diff
@@ -37,22 +66,22 @@ func RowNamespaceDataFromShares(
 			break
 		}
 	}
+
+	// if count is 0, then the namespace is not present in the shares. Return non-inclusion proof.
 	if count == 0 {
-		// FIXME: This should return Non-inclusion proofs instead. Need support in app wrapper to generate
-		// absence proofs.
-		return RowNamespaceData{}, fmt.Errorf("no shares found in the namespace for row %d", rowIndex)
```
Wondertan, line 43: So does it return non-inclusion proof now? If so, we may want to close the issue that we made for those in shwap context. Also, we should test that on Accessor if not yet

share/new_eds/testing.go
cristaloleg, line 1: Nit: when file doesn't have `_test.go` suffix it will be compiled with the regular code. Probably the biggest thing here is that code from `testing.go` file can be used with the regular code by mistake (harmless but doubtful).

share/new_eds/testing.go
walldiss: Yes, they can be used for any Accessor implementation externaly

share/shwap/row_namespace_data.go
walldiss: Yes, I've added fucntionality here as it was required and I didn't want to make too many PRs. Issue is linked already, check description

Wondertan: Do we test it ?

share/new_eds/testing.go
walldiss: Linker will strip unused functions. The motivation I had is that independent functions are more flexible. Lets make them private until it is needed. I've made a single suite and made all functions private.

Wondertan (approved):

PR #3427, opened 2024-05-22
fix(pruner): findPruneable short-circuits if estimated cutoff height is too close to last pruned

renaynay:
It can happen that a pruning job is triggered where no headers need pruning (if estimated cutoff height is one greater than or equal to the last pruned height). This would cause a trivial error to occur during a pruning job:

```
failed to get range from header store	{"from": 36352, "to": 36353, "error": "header/store: invalid range(36353,36352)"}
```

We should check to make sure that estimatedCutoffHeight <= lastPruned.Height()+1 instead of just equivalency because requesting `GetRangeByHeight(ctx, lastPruned, estimatedCutoffHeight)` where lastPruned is 500 and estimated cutoff is 501 can result in a range of (500:501] being requested (so the request once it reaches the header store is actually for [501:500]) since header store implementation actually requests the range [from+1 : to-1].

Also, I moved the tests specific to findPruneable into a separate file.

This is actually a bit ugly and we should ensure that all `GetRangeByHeight` returned ranges are consistent across `Getter` implementations.


pruner/find.go
```diff
@@ -31,8 +31,9 @@ func (s *Service) findPruneableHeaders(
 		return nil, err
 	}
 
-	if lastPruned.Height() == estimatedCutoffHeight {
-		// nothing left to prune
+	// estimatedCutoffHeight must be at least greater than lastPruned.Height()+1
+	// to continue, otherwise, there is nothing left to prune
+	if estimatedCutoffHeight <= lastPruned.Height()+1 {
```
walldiss, line 36: Why greater than lastPruned.Height() `+1` ? Seems like if it is lastPruned.Height()+1 it should be pruned and not skipped

renaynay: @walldiss let's say estimatedCutoffHeight == lastPruned.Height()+1 where lastPruned height is 5


The request down to the header store would look like: 

GetRangeByHeight(from:5, 6) -- it goes down to the header store as: 



```
func (s *Store[H]) GetRangeByHeight(
	ctx context.Context,
	from H,
	to uint64,
) ([]H, error) {
	headers, err := s.getRangeByHeight(ctx, from.Height()+1, to)
	if err != nil {
		return nil, err
	}
```

`getRangeByHeight( 6, 6)` where it hits



```
	if from > to-1 {
		return nil, fmt.Errorf("header/store: invalid range(%d,%d)", from, to-1)
	}
```

if 6 > 5 



It's just ugliness from inconsistent implementations across GetRangeByHeight implementations and what range exactly is expected as a return value.

walldiss: Request has to be from 6, as it is the height that needs to be pruned. GetRangeByHeight(from:6, 6)

walldiss: I think the issue is in header pkg. It should include `from` into requested interval

Wondertan: It shouldn't because otherwise you request data you already have

PR #3441, opened 2024-05-27
feat(share/store_v2) Update file interface to use shwap types

walldiss:


share/store/file/axis_half.go
```diff
@@ -0,0 +1,19 @@
+package file
+
+import (
+	"github.com/celestiaorg/celestia-node/share"
+	"github.com/celestiaorg/celestia-node/share/shwap"
+)
+
+type AxisHalf struct {
+	Shares   []share.Share
+	IsParity bool
```
Wondertan, line 10: Lets just the shwap sides. It its much clearer to have one way of doing that in the code base

share/ipld/get_shares.go
```diff
@@ -44,12 +44,13 @@ func GetShares(ctx context.Context, bg blockservice.BlockGetter, root cid.Cid, s
 func GetSharesByNamespace(
 	ctx context.Context,
 	bGetter blockservice.BlockGetter,
-	root cid.Cid,
+	root []byte,
```
Wondertan, line 47: Don't get why is that needed to be part of this PR

walldiss: because ipld.FilterRootByNamespace is gone and no longer returns cids. We use indexes and row roots instead now

share/store/file/axis_half.go
walldiss: What is left side of columnt?

share/ipld/get_shares.go
Wondertan: Yeah, but you could calll MustRootToCID before calling GetShareByNamespace

share/store/file/axis_half.go
Wondertan: Single type > semantical correctness(that can be documented)

share/ipld/get_shares.go
walldiss: ipld.GetSharesByNamespace is used in File pkg to build proofs from the tree. I don't think File needs to be aware of cids, so construction of cid moved into ipld pkg

share/store/file/axis_half.go
walldiss: It will be confusing to have Columns types to be names as shwap.Row and to have left and right sides. I would rather use file.AxisHalf type in shwap pkg for shwap.HalfRow, as it is the type that describes both rows and columns.

share/ipld/get_shares.go
Wondertan: kk

share/store/file/axis_half.go
Wondertan: But we can't use AxisHalf in shwap because its not part of the spec

Wondertan:
Needs rebase and PR description

walldiss:
Rebased

Wondertan (approved):

share/root.go
```diff
@@ -9,6 +15,30 @@ import (
...
+
+// IsEmptyRoot check whether DataHash corresponds to the root of an empty block EDS.
+func (dh DataHash) IsEmptyRoot() bool {
+	return bytes.Equal(EmptyRoot().Hash(), dh)
+}
+
+// NewSHA256Hasher returns a new instance of a SHA-256 hasher.
+func NewSHA256Hasher() hash.Hash {
+	return sha256.New()
+}
```
Wondertan, line 40: Is this used?

walldiss: Yes, it is used everywhere instead of sha256.New(). The only place missing is latest refactoring of shwap types. I'll fix in next commit to this PR.

PR #3446, opened 2024-05-28
chore: bump libp2p, go-header and app

Wondertan:
~Blocked on https://github.com/celestiaorg/cosmos-sdk/issues/400~

It needs to wait for v1.11.0


state/core_access.go
```diff
@@ -151,14 +152,20 @@ func (ca *CoreAccessor) Start(ctx context.Context) error {
...
 	if err != nil {
 		return err
 	}
+	// this ensures we can't start the node without core connection
+	client.Connect()
+	if !client.WaitForStateChange(ctx, connectivity.Ready) {
+		// hits the case when context is canceled
+		return fmt.Errorf("couldn't connect to core endpoint(%s): %w", endpoint, ctx.Err())
+	}
+
```
Wondertan, line 168: The options are:
* Make connections lazily during requests(new grpc now allows that)
* Ensure we setup connection on start

This change persists in existing behavior where a node cannot be started without a working connection, but we may reconsider that. WIth lazy users will see core connection issues(which happen surprisingly often) only during tx and state requests

nodebuilder/p2p/opts.go
```diff
@@ -35,8 +34,3 @@ func WithP2PKeyStr(key string) fx.Option {
 func WithHost(hst hst.Host) fx.Option {
 	return fxutil.ReplaceAs(hst, new(HostBase))
 }
-
-// WithBlockService allows to replace the default BlockService.
-func WithBlockService(bServ blockservice.BlockService) fx.Option {
-	return fxutil.ReplaceAs(bServ, new(blockservice.BlockService))
-}
```
Wondertan, line 42: Deprecation warning, but this code wasn't anymore anyway

cristaloleg (approved):
SGTM

state/core_access.go
walldiss: In the latest gRPC, it's now standard to establish connections only during requests because the underlying transport automatically attempts to reestablish connections if errors occur. Restarting the application due to failed connectivity doesn't add much value; it only delays the process without offering any benefits over internal retries. This is generally considered a bad practice. If there are compelling reasons to enforce connection establishment on startup, I'd be interested to hear them.

renaynay: repeating here but we have agreement: 

```
user passes --core.ip to node thinking core accessor was set up properly, then later on down the line, tries to use core accessor but grpc conn fails bc that endpoint is borked for whatever reason. I’d rather it fail on conn at start since user expects working state service
```

renaynay (dismissed):
👍🏻

walldiss (approved):

cristaloleg (approved):

vgonkivs (approved):

PR #3452, opened 2024-05-31
refactor(shwap): Extract eds interface

walldiss:
- Reintroduce file interface as eds interface. Change aims to allow usage of EDS interface outside of storage package and to be high level interface of EDS methods.
- Renames of eds interface methods to align with returned shwap types names
    - Share() -> Sample
    - Data -> Row data
- Extracts New<shwap_type_name>FromEDS functions to eds file methods
    - moves associated tests to eds pkg
    
**Additional refactoring:**
- **Change Interface Name**: Realized that 'EDS' is a terrible name for an interface. Renamed `eds.EDS` to `eds.Accessor` to more accurately reflect its functionality rather than its internal content.

- **Separate Closer**: Extracted `Closer` from `Accessor`. Now it is available in a new composite interface `AccessorCloser`.

- **Rename InMem**: Renamed `InMem` to `rsmt2d` to better align with its usage.

- **Decouple NamespacedData**: Separated `NamespacedData` from the `rsmt2d` implementation. It is now a standalone function.

- **Update EDS Method**: Replaced the `EDS()` method with `Flattened`, similar to `rsmt2d`. Considered introducing two separate methods, `Flattened` and `FlattenedODS`, with the latter to be potentially added later. Proposed to park this suggestion in an issue for future consideration.

### DIFF
--- a/share/new_eds/accessor.go
+++ b/share/new_eds/accessor.go
@@ -13,7 +13,7 @@ import (
 // Accessor is an interface for accessing extended data square data.
 type Accessor interface {
 	// Size returns square size of the Accessor.
-	Size() int
+	Size(ctx context.Context) int
 	// Sample returns share and corresponding proof for row and column indices. Implementation can
 	// choose which axis to use for proof. Chosen axis for proof should be indicated in the returned
 	// Sample.
@@ -23,8 +23,8 @@ type Accessor interface {
 	AxisHalf(ctx context.Context, axisType rsmt2d.Axis, axisIdx int) (AxisHalf, error)
 	// RowNamespaceData returns data for the given namespace and row index.
 	RowNamespaceData(ctx context.Context, namespace share.Namespace, rowIdx int) (shwap.RowNamespaceData, error)
-	// Flattened returns data shares extracted from the Accessor.
-	Flattened(ctx context.Context) ([]share.Share, error)
+	// Shares returns data shares extracted from the Accessor.
+	Shares(ctx context.Context) ([]share.Share, error)
 }
 
 // AccessorCloser is an interface that groups Accessor and io.Closer interfaces.
--- a/share/new_eds/accessor.go
+++ b/share/new_eds/accessor.go
@@ -10,10 +10,9 @@ import (
 	"github.com/celestiaorg/celestia-node/share/shwap"
 )
 
-// EDS is an interface for accessing extended data square data.
-type EDS interface {
-	io.Closer
-	// Size returns square size of the file.
+// Accessor is an interface for accessing extended data square data.
+type Accessor interface {
+	// Size returns square size of the Accessor.
 	Size() int
 	// Sample returns share and corresponding proof for row and column indices. Implementation can
 	// choose which axis to use for proof. Chosen axis for proof should be indicated in the returned
@@ -22,8 +21,14 @@ type EDS interface {
 	// AxisHalf returns half of shares axis of the given type and index. Side is determined by
 	// implementation. Implementations should indicate the side in the returned AxisHalf.
 	AxisHalf(ctx context.Context, axisType rsmt2d.Axis, axisIdx int) (AxisHalf, error)
-	// RowData returns data for the given namespace and row index.
-	RowData(ctx context.Context, namespace share.Namespace, rowIdx int) (shwap.RowNamespaceData, error)
-	// EDS returns extended data square stored in the file.
-	EDS(ctx context.Context) (*rsmt2d.ExtendedDataSquare, error)
+	// RowNamespaceData returns data for the given namespace and row index.
+	RowNamespaceData(ctx context.Context, namespace share.Namespace, rowIdx int) (shwap.RowNamespaceData, error)
+	// Flattened returns data shares extracted from the Accessor.
+	Flattened(ctx context.Context) ([]share.Share, error)
+}
+
+// AccessorCloser is an interface that groups Accessor and io.Closer interfaces.
+type AccessorCloser interface {
+	Accessor
+	io.Closer
 }
--- a/share/new_eds/eds.go
+++ b/share/new_eds/eds.go
@@ -10,6 +10,7 @@ import (
 	"github.com/celestiaorg/celestia-node/share/shwap"
 )
 
+// EDS is an interface for accessing extended data square data.
 type EDS interface {
 	io.Closer
 	// Size returns square size of the file.
@@ -20,6 +21,6 @@ type EDS interface {
 	AxisHalf(ctx context.Context, axisType rsmt2d.Axis, axisIdx int) (AxisHalf, error)
 	// Data returns data for the given namespace and row index.
 	Data(ctx context.Context, namespace share.Namespace, rowIdx int) (shwap.RowNamespaceData, error)
-	// InMem returns extended data square stored in the file.
+	// EDS returns extended data square stored in the file.
 	EDS(ctx context.Context) (*rsmt2d.ExtendedDataSquare, error)
 }
--- a/share/new_eds/eds.go
+++ b/share/new_eds/eds.go
@@ -15,9 +15,12 @@ type EDS interface {
 	io.Closer
 	// Size returns square size of the file.
 	Size() int
-	// Sample returns share and corresponding proof for the given axis and share index in this axis.
+	// Sample returns share and corresponding proof for row and column indices. Implementation can choose
+	// which axis to use for proof. Chosen axis for proof should be indicated in the returned Sample.
 	Sample(ctx context.Context, rowIdx, colIdx int) (shwap.Sample, error)
-	// AxisHalf returns Shares for the first half of the axis of the given type and index.
+	// AxisHalf returns half of shares axis of the given type and index. Side is determined by implementation
+	// and is not guaranteed to be the same for all implementations. Implementations should indicate the side
+	// in the returned AxisHalf.
 	AxisHalf(ctx context.Context, axisType rsmt2d.Axis, axisIdx int) (AxisHalf, error)
 	// RowData returns data for the given namespace and row index.
 	RowData(ctx context.Context, namespace share.Namespace, rowIdx int) (shwap.RowNamespaceData, error)
--- a/share/new_eds/eds.go
+++ b/share/new_eds/eds.go
@@ -15,12 +15,12 @@ type EDS interface {
 	io.Closer
 	// Size returns square size of the file.
 	Size() int
-	// Share returns share and corresponding proof for the given axis and share index in this axis.
-	Share(ctx context.Context, rowIdx, colIdx int) (*shwap.Sample, error)
+	// Sample returns share and corresponding proof for the given axis and share index in this axis.
+	Sample(ctx context.Context, rowIdx, colIdx int) (shwap.Sample, error)
 	// AxisHalf returns Shares for the first half of the axis of the given type and index.
 	AxisHalf(ctx context.Context, axisType rsmt2d.Axis, axisIdx int) (AxisHalf, error)
-	// Data returns data for the given namespace and row index.
-	Data(ctx context.Context, namespace share.Namespace, rowIdx int) (shwap.RowNamespaceData, error)
+	// RowData returns data for the given namespace and row index.
+	RowData(ctx context.Context, namespace share.Namespace, rowIdx int) (shwap.RowNamespaceData, error)
 	// EDS returns extended data square stored in the file.
 	EDS(ctx context.Context) (*rsmt2d.ExtendedDataSquare, error)
 }
--- a/share/new_eds/eds.go
+++ b/share/new_eds/eds.go
@@ -15,12 +15,12 @@ type EDS interface {
 	io.Closer
 	// Size returns square size of the file.
 	Size() int
-	// Sample returns share and corresponding proof for row and column indices. Implementation can choose
-	// which axis to use for proof. Chosen axis for proof should be indicated in the returned Sample.
+	// Sample returns share and corresponding proof for row and column indices. Implementation can
+	// choose which axis to use for proof. Chosen axis for proof should be indicated in the returned
+	// Sample.
 	Sample(ctx context.Context, rowIdx, colIdx int) (shwap.Sample, error)
-	// AxisHalf returns half of shares axis of the given type and index. Side is determined by implementation
-	// and is not guaranteed to be the same for all implementations. Implementations should indicate the side
-	// in the returned AxisHalf.
+	// AxisHalf returns half of shares axis of the given type and index. Side is determined by
+	// implementation. Implementations should indicate the side in the returned AxisHalf.
 	AxisHalf(ctx context.Context, axisType rsmt2d.Axis, axisIdx int) (AxisHalf, error)
 	// RowData returns data for the given namespace and row index.
 	RowData(ctx context.Context, namespace share.Namespace, rowIdx int) (shwap.RowNamespaceData, error)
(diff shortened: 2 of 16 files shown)


share/new_eds/in_mem.go
```diff
@@ -12,29 +12,30 @@ import (
...
 )
 
-var _ EdsFile = (*MemFile)(nil)
+var _ EDS = InMem{}
 
-type MemFile struct {
-	Eds *rsmt2d.ExtendedDataSquare
+// InMem is an in-memory implementation of EDS.
+type InMem struct {
+	*rsmt2d.ExtendedDataSquare
```
cristaloleg, line 19: Why embedded?

walldiss: To allow access to inherited type

walldiss: It is meant to extend imported  rsmt2d.ExtendedDataSquare type with few more methods

share/new_eds/in_mem.go
```diff
@@ -12,29 +12,30 @@ import (
 	"github.com/celestiaorg/celestia-node/share/shwap"
 )
 
-var _ EdsFile = (*MemFile)(nil)
+var _ EDS = InMem{}
 
-type MemFile struct {
-	Eds *rsmt2d.ExtendedDataSquare
+// InMem is an in-memory implementation of EDS.
+type InMem struct {
```
Wondertan, line 18: Just `Mem`?

Wondertan:
Drafting until we migrate some funcs over EDS interface

share/new_eds/eds.go
```diff
@@ -10,16 +10,17 @@ import (
...
+	Sample(ctx context.Context, rowIdx, colIdx int) (shwap.Sample, error)
 	// AxisHalf returns Shares for the first half of the axis of the given type and index.
 	AxisHalf(ctx context.Context, axisType rsmt2d.Axis, axisIdx int) (AxisHalf, error)
-	// Data returns data for the given namespace and row index.
-	Data(ctx context.Context, namespace share.Namespace, rowIdx int) (shwap.RowNamespaceData, error)
+	// RowData returns data for the given namespace and row index.
+	RowData(ctx context.Context, namespace share.Namespace, rowIdx int) (shwap.RowNamespaceData, error)
 	// EDS returns extended data square stored in the file.
 	EDS(ctx context.Context) (*rsmt2d.ExtendedDataSquare, error)
 }
```
Wondertan: * Most of the comments reference a file, and this is not a file anymore
* RowData -> RowNamespaceData
* The EDS method on the EDS interface makes me feel uneasy. I would rename either of those to something like `Square`, preferably the interface. Then, the EDS method would return the canonical rsmt2d struct.
 * This also made us wonder what the Getter interface should return: the actual EDS or the interface. I remember we had issues with serialization on rsmt2d side, which we could solve on our layer as well.

share/new_eds/in_mem.go
Wondertan: Bump

share/new_eds/eds.go
```diff
@@ -0,0 +1,29 @@
...
+
+	"github.com/celestiaorg/rsmt2d"
+
+	"github.com/celestiaorg/celestia-node/share"
+	"github.com/celestiaorg/celestia-node/share/shwap"
+)
+
+// EDS is an interface for accessing extended data square data.
+type EDS interface {
+	io.Closer
```
Wondertan, line 15: Do we want to enforce closing for all implementations? Generally I am fine with that, but we may also consider enforcing that on the storage/network layer only.

share/new_eds/eds.go
walldiss: > EDS(ctx context.Context) (*rsmt2d.ExtendedDataSquare, error)
Perhaps only method can be renamed to `ToRSMT2D`?

walldiss:
I've applied review suggestions in https://github.com/celestiaorg/celestia-node/pull/3452/commits/46daf1359200898f036760db5c0613e6857dcc34 and listed what changed in PR description in **Additional refactoring** section.

share/new_eds/rsmt2d_test.go
```diff
@@ -0,0 +1,74 @@
...
+				require.Equal(t, proofType, sample.ProofType)
+				require.NotNil(t, sample.Proof)
+				require.Equal(t, sample.Proof.End()-sample.Proof.Start(), 1)
+				require.Len(t, sample.Proof.Nodes(), 4)
+			}
+		}
+	}
+}
+
+func randRsmt2dAccsessor(t *testing.T, size int) (Rsmt2D, *share.Root) {
```
cristaloleg, line 69: testing.TB please

share/new_eds/accessor.go
```diff
@@ -0,0 +1,34 @@
...
+	// implementation. Implementations should indicate the side in the returned AxisHalf.
+	AxisHalf(ctx context.Context, axisType rsmt2d.Axis, axisIdx int) (AxisHalf, error)
+	// RowNamespaceData returns data for the given namespace and row index.
+	RowNamespaceData(ctx context.Context, namespace share.Namespace, rowIdx int) (shwap.RowNamespaceData, error)
+	// Flattened returns data shares extracted from the Accessor.
+	Flattened(ctx context.Context) ([]share.Share, error)
+}
+
+// AccessorCloser is an interface that groups Accessor and io.Closer interfaces.
+type AccessorCloser interface {
```
cristaloleg, line 31: Do we use this somewhere? If not, why not add `io.Closer` to `Accessor` and keep 1 type?

share/new_eds/nd.go
```diff
@@ -0,0 +1,30 @@
...
+	"context"
+	"fmt"
+
+	"github.com/celestiaorg/celestia-node/share"
+	"github.com/celestiaorg/celestia-node/share/shwap"
+)
+
+// NamespacedData extracts shares for a specific namespace from an EDS, considering
+// each row independently.
+func NamespacedData(
```
cristaloleg, line 13: Why this isn't a method from `Accessor`  but a standalone func? 🤔

share/new_eds/rsmt2d.go
```diff
@@ -0,0 +1,116 @@
...
+	namespace share.Namespace,
+	rowIdx int,
+) (shwap.RowNamespaceData, error) {
+	shares := eds.Row(uint(rowIdx))
+	return shwap.RowNamespaceDataFromShares(shares, namespace, rowIdx)
+}
+
+// Flattened returns data shares extracted from the EDS.
+func (eds Rsmt2D) Flattened(_ context.Context) ([]share.Share, error) {
+	// Flattened returns copy of shares, which is not efficient.
```
cristaloleg, line 103: `// TODO...` ? or just move it to method doc?

share/new_eds/nd.go
```diff
@@ -0,0 +1,30 @@
...
+	ctx context.Context,
+	root *share.Root,
```
ramin, line 15: prolly a silly question, but the comment doc suggests that I can call this with JUST and EDS, but it'd be good to explain (really NOT assume) that I know why I need, or even WHAT the share.Root is for and why it's necessary

share/new_eds/rsmt2d.go
```diff
@@ -0,0 +1,116 @@
...
+) (shwap.Sample, error) {
+	var axisIdx, shrIdx int
+	switch proofType {
+	case rsmt2d.Row:
+		axisIdx, shrIdx = rowIdx, colIdx
+	case rsmt2d.Col:
+		axisIdx, shrIdx = colIdx, rowIdx
+	default:
+		return shwap.Sample{}, fmt.Errorf("invalid proof type: %d", proofType)
+	}
```
ramin, line 53: this little switcher seems like something nice t have in it's own little func and test for, subtle but small like a nice place for a bug to hide

share/new_eds/rsmt2d.go
```diff
@@ -0,0 +1,116 @@
...
+		Share:     shrs[shrIdx],
+		Proof:     &prf,
+		ProofType: proofType,
+	}, nil
+}
+
+// AxisHalf returns Shares for the first half of the axis of the given type and index.
+func (eds Rsmt2D) AxisHalf(_ context.Context, axisType rsmt2d.Axis, axisIdx int) (AxisHalf, error) {
+	return AxisHalf{
+		Shares:   getAxis(eds.ExtendedDataSquare, axisType, axisIdx)[:eds.Size()/2],
```
ramin, line 79: very nit: personally i would assign / call this getAxis before the return then assign to AxisHalf.Shares, a little less cognitive overhead that its doing multiple things calling getaAxis, retiring index, AND also /. 

Also, is there a chance that we should check the `eds.Size()/2` and potentially return a bounds error with the error return (that is currently always nil)

share/shwap/row.go
```diff
@@ -35,14 +34,12 @@ func NewRow(halfShares []share.Share, side RowSide) Row {
...
 // RowFromEDS constructs a new Row from an Extended Data Square based on the specified index and
 // side.
-func RowFromEDS(square *rsmt2d.ExtendedDataSquare, idx int, side RowSide) Row {
-	sqrLn := int(square.Width())
-	shares := square.Row(uint(idx))
+func RowFromShares(shares []share.Share, side RowSide) Row {
 	var halfShares []share.Share
 	if side == Right {
-		halfShares = shares[sqrLn/2:] // Take the right half of the shares.
+		halfShares = shares[len(shares)/2:] // Take the right half of the shares.
```
ramin, line 40: I assume we can be confident these divides will never blow up?

ramin (commented):
had a very superficial look / review, the interfaces and methods are starting to make sense to me as a layman and someone that might need to use them 😸 

One question: is `share/new_eds` meant to survive? I wonder if having it organized more like

`share/eds/v2` or something would be better than `new_eds`?

share/new_eds/accessor.go
walldiss: Accessor does not necessarily implements Closer. For example Rsmt2d one does not.

share/new_eds/nd.go
walldiss: NamespacedData works using Accessor API and will be the same for all implementations. There is no reason to put it into interface now.

share/new_eds/rsmt2d.go
walldiss: More like todo for rsmt2d 🙂  Will move up to doc

share/shwap/row.go
walldiss: It is safe even for 0 len

share/new_eds/rsmt2d.go
walldiss: yeah, I'll decompile this line

getAxis will return exactly eds.Size() amount of shares, so it is always safe

share/new_eds/rsmt2d_test.go
walldiss: Why? It is not exported and there are no benchmarks

share/new_eds/accessor.go
walldiss: There will be cases, when we won't need closer

share/new_eds/nd.go
walldiss: doc improved
https://github.com/celestiaorg/celestia-node/pull/3452/commits/8cbd6b461597a0b4ea0a6e8f0e500f78018d6309

share/new_eds/rsmt2d.go
walldiss: extracted in https://github.com/celestiaorg/celestia-node/pull/3452/commits/8cbd6b461597a0b4ea0a6e8f0e500f78018d6309

share/new_eds/rsmt2d.go
walldiss: Adjusted comment and moved to method description
https://github.com/celestiaorg/celestia-node/pull/3452/commits/8cbd6b461597a0b4ea0a6e8f0e500f78018d6309

share/new_eds/rsmt2d.go
```diff
@@ -0,0 +1,121 @@
...
+
+// Rsmt2D is a rsmt2d based in-memory implementation of Accessor.
+type Rsmt2D struct {
+	*rsmt2d.ExtendedDataSquare
+}
+
+// Close does nothing.
+func (eds Rsmt2D) Close() error {
+	return nil
+}
```
Wondertan, line 24: Shouldn't we remove it?

share/new_eds/accessor.go
```diff
@@ -0,0 +1,34 @@
...
+	// choose which axis to use for proof. Chosen axis for proof should be indicated in the returned
+	// Sample.
+	Sample(ctx context.Context, rowIdx, colIdx int) (shwap.Sample, error)
+	// AxisHalf returns half of shares axis of the given type and index. Side is determined by
+	// implementation. Implementations should indicate the side in the returned AxisHalf.
+	AxisHalf(ctx context.Context, axisType rsmt2d.Axis, axisIdx int) (AxisHalf, error)
+	// RowNamespaceData returns data for the given namespace and row index.
+	RowNamespaceData(ctx context.Context, namespace share.Namespace, rowIdx int) (shwap.RowNamespaceData, error)
+	// Flattened returns data shares extracted from the Accessor.
+	Flattened(ctx context.Context) ([]share.Share, error)
```
Wondertan, line 27: We usually follow a different way of naming things. To be consistent with our naming, it should be something like `Shares`

share/shwap/row.go
Wondertan: Len of shares is guaranteed to be power of 2, so should not be an issue, however we should be very cautious with that

share/shwap/sample_test.go
```diff
@@ -1,6 +1,7 @@
-package shwap
+package shwap_test
```
Wondertan, line 1: It should stay shwap, no?

walldiss: It is necessary to resolve import cycle of testing pkg, that relies on eds for support functions.

share/new_eds/rsmt2d.go
```diff
@@ -0,0 +1,116 @@
...
+	"fmt"
+
+	"github.com/celestiaorg/celestia-app/pkg/wrapper"
+	"github.com/celestiaorg/rsmt2d"
+
+	"github.com/celestiaorg/celestia-node/share"
+	"github.com/celestiaorg/celestia-node/share/shwap"
+)
+
+var _ Accessor = Rsmt2D{}
```
Wondertan, line 14: That's what cool kids do, but we are not doing it in our codebase. Not sure it is worth it as our tests would anyway catch if the implementation doesn't follow the interface anymore

share/shwap/sample_test.go
```diff
@@ -54,55 +36,53 @@ func TestSampleValidate(t *testing.T) {
...
 	require.NoError(t, err)
 	sample.ProofType = rsmt2d.Col
-	included = sample.verifyInclusion(root, 0, 0)
-	require.False(t, included)
+	err = sample.Validate(root, 0, 0)
+	require.ErrorIs(t, err, shwap.ErrorFailedVerification)
 
 	// Corrupt the last root hash byte
-	sample, err = SampleFromEDS(eds, rsmt2d.Row, 0, 0)
+	sample, err = inMem.Sample(context.TODO(), 0, 0)
```
Wondertan, line 66: TODO hints that a real context should be added sometime in the future, and it was added during the times when std started adopting context, but in this case, we should just pas Background

share/new_eds/rsmt2d.go
walldiss: It is nice to have safety feature, that is free. Also it will be there of all other Accessor implementations

walldiss: It is useful to have this check to catch problems on compile time (or even by ide), instead of waiting for test runs

Issue #3448, opened 2024-05-28
share.GetSharesByNamespace: non-deterministic response


christopherbrumm:
### Celestia Node version

v0.13.4 - Commit: 13ac02fe9a659b8741c2f62fac256c4ef249b718

### OS

Ubuntu 22.04.4 LTS

### Install tools

- go1.22.0
- GNU Make 4.3

### Others

`celestia full start --core.ip public-celestia-grpc.numia.xyz --rpc.port 35001`

### Steps to reproduce it

1. Set up two different full nodes
2. Request the `share.GetSharesByNamespace` endpoint for a certain namespace for a range of heights (e.g. `1,000,000 - 1,500,000`).
3. Compare the results.

### Expected result

Deterministic responses.

### Actual result

We’ve experienced non-deterministic responses to the `share.GetSharesByNamespace` query. For certain heights (e.g., height 1400810, namespace `AAAAAAAAAAAAAAAAAAAAAAAAAAAADBuw7+PjGs8=`), different nodes respond with different values (`null` or `[]`), although they have the exact same  full-node setup.

### Relevant log output

```Shell
Will be submitted in the following days alongside with a script that can be used to reproduce the submitted issue.
```

walldiss:
Hey! Thank you for reporting the issue. It is very concerning and I'll try to reproduce. I have few question, that will help to identify whats is the root cause: 
- How do you call share.GetSharesByNamespace? Do you use JSON-RPC or maybe CLI?
- Can inconsistency be achieved by doing request to the same node?
  - If yes, do you run your nodes with same setup mentioned above?
- Does `null` result contain any error with it?

christopherbrumm:
> * How do you call share.GetSharesByNamespace? Do you use JSON-RPC or maybe CLI?
With JSON-RPC:
`await axios.post(this.config.rpc, {
        id: 1,
        jsonrpc: '2.0',
        method: 'share.GetSharesByNamespace',
        params: [
          header,
          namespace
        ]
      }, {
        headers: {
          'Content-Type': 'application/json',
          'Authorization': `Bearer ${process.env.CELESTIA_NODE_AUTH_TOKEN}`
        }
      });`

> * Can inconsistency be achieved by doing request to the same node?
No. After trying again, all nodes respond with `[]`.

> * Does `null` result contain any error with it?
No, it doesn't contain an error, although I haven't checked the nodes' logs. Respond:
`{
[...]

PR #3454, opened 2024-06-03
fix(share/getters)!: return nil shares from all getters in non-inclusion case

walldiss:
Make return of shares consistent accross tall getter, when encountering non-inclusion case
Improves unnecessary allocations of empty slice.

Resolves https://github.com/celestiaorg/celestia-node/issues/3448


cristaloleg (approved):
SGTM

Wondertan:
Technically breaking, so let's mark it as so.

Wondertan (approved):

PR #3455, opened 2024-06-03
fix(share/pruner) enable Light node pruner service

walldiss:
Fixes 2 issues that prevented Light pruner service from launching:
- pruning flag was not parsed by Light node
- Pruner service was not provided to Light node


nodebuilder/pruner/module.go
```diff
@@ -34,31 +33,42 @@ func ConstructModule(tp node.Type, cfg *Config) fx.Option {
...
-		opts := baseComponents
 		if cfg.EnableService {
-			opts = fullAndBridgeComponents
+			return fx.Module("prune",
+				baseComponents,
+				prunerService,
+				fx.Provide(func(store *eds.Store) pruner.Pruner {
+					return full.NewPruner(store)
+				}),
+			)
```
walldiss, line 61: I hate this duplication, but kept it this way to avoid major refactring of  code that is already partially tested

nodebuilder/pruner/module.go
```diff
@@ -34,31 +33,42 @@ func ConstructModule(tp node.Type, cfg *Config) fx.Option {
...
+		if cfg.EnableService {
+			return fx.Module("prune",
+				baseComponents,
+				prunerService,
+				fx.Provide(light.NewPruner),
+			)
+		}
 		return fx.Module("prune",
 			baseComponents,
 			fx.Provide(light.NewPruner),
```
renaynay, line 51: why provide light pruner if serv not enabled?

nodebuilder/pruner/module.go
```diff
@@ -34,31 +33,42 @@ func ConstructModule(tp node.Type, cfg *Config) fx.Option {
...
 	case node.Full:
-		opts := baseComponents
 		if cfg.EnableService {
-			opts = fullAndBridgeComponents
+			return fx.Module("prune",
+				baseComponents,
+				prunerService,
+				fx.Provide(func(store *eds.Store) pruner.Pruner {
+					return full.NewPruner(store)
+				}),
```
renaynay, line 60: I think you can just provide full.NewPruner since it takes eds Store

nodebuilder/pruner/module.go
```diff
@@ -34,31 +33,42 @@ func ConstructModule(tp node.Type, cfg *Config) fx.Option {
...
-		)
+		return fx.Module("prune", baseComponents)
 	case node.Bridge:
 		if cfg.EnableService {
 			return fx.Module("prune",
-				fullAndBridgeComponents,
+				baseComponents,
+				prunerService,
+				fx.Provide(func(store *eds.Store) pruner.Pruner {
+					return full.NewPruner(store)
```
renaynay, line 70: same as above

renaynay (dismissed):
Lord.

ramin (dismissed):

nodebuilder/pruner/module.go
walldiss: It is not gree, so it used to be this way 😆  Will remove

nodebuilder/pruner/module.go
walldiss: Well, it was set this way before and for reason. Pruner returned from full.NewPruner(store) is not an interface and needs to be casted. Anyway, updated to use ProvideAs

nodebuilder/pruner/module.go
Wondertan: The comment got moved or something. The comment is about if case above

ramin (dismissed):

cristaloleg (approved):
SGTM

nodebuilder/pruner/module.go
```diff
@@ -34,31 +33,37 @@ func ConstructModule(tp node.Type, cfg *Config) fx.Option {
...
 		// quirk in FX.
 		fx.Invoke(func(_ *pruner.Service) {}),
-		fx.Provide(func(store *eds.Store) pruner.Pruner {
-			return full.NewPruner(store)
-		}),
 	)
 
 	switch tp {
 	// TODO: Eventually, light nodes will be capable of pruning samples
 	//  in which case, this can be enabled.
```
Wondertan, line 40: You can remove this TODO. 

I think a rebase artifact could have led to some of the changes missing, like this TODO, because I remember removing it. This may also explain why you could not run the Service(as per PR description), but I definitely could.

cmd/util.go
```diff
@@ -119,20 +119,14 @@ func PersistentPreRunEnv(cmd *cobra.Command, nodeType node.Type, _ []string) err
...
 			return err
 		}
-	case node.Full:
-		err = header.ParseFlags(cmd, &cfg.Header)
-		if err != nil {
-			return err
-		}
-		pruner.ParseFlags(cmd, &cfg.Pruner)
 	case node.Bridge:
-		pruner.ParseFlags(cmd, &cfg.Pruner)
```
Wondertan, line 135: Do we disable it for the bridge?

walldiss: line 122, It is moved to be applied to all types of node

cristaloleg (approved):

renaynay (approved):

Wondertan (approved):

PR #3463, opened 2024-06-04
chore(swamp): remove always failing assert

cristaloleg:
From `context.Context.Err() error` documentation (https://pkg.go.dev/context#Context):

```
	// If Done is not yet closed, Err returns nil.
	// If Done is closed, Err returns a non-nil error explaining why:
	// Canceled if the context was canceled
	// or DeadlineExceeded if the context's deadline passed.
	// After Err returns a non-nil error, successive calls to Err return the same error.
	Err() error
```

In other words when we observe a cancelled context `Err` is always not nil https://go.dev/play/p/2N6F75a70lK


Wondertan (approved):
Wow

ramin (approved):

walldiss (approved):

PR #3465, opened 2024-06-04
fix(share/getters): Shrex getter routing should be based in fixed `light.Window`

renaynay:
Since it is a network constant, shrex getter request routing should be based on the hardcoded `light.Window` rather than the avail window that is passed to it.

This also fixes a bug where archival nodes would not route historic EDS requests to the archival pool due to `AvailabilityWindow` being 0.

### DIFF
--- a/nodebuilder/share/module.go
+++ b/nodebuilder/share/module.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/celestiaorg/celestia-node/nodebuilder/node"
 	modp2p "github.com/celestiaorg/celestia-node/nodebuilder/p2p"
-	"github.com/celestiaorg/celestia-node/pruner"
 	"github.com/celestiaorg/celestia-node/share"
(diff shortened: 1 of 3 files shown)


nodebuilder/share/module.go
```diff
@@ -92,17 +91,12 @@ func shrexComponents(tp node.Type, cfg *Config) fx.Option {
...
 				ndClient *shrexnd.Client,
 				managers map[string]*peers.Manager,
-				window pruner.AvailabilityWindow,
 			) *getters.ShrexGetter {
 				return getters.NewShrexGetter(
 					edsClient,
 					ndClient,
 					managers[fullNodesTag],
 					managers[archivalNodesTag],
-					// TODO @renaynay: Pruned FNs should pass `light.Window` to shrex getter
```
renaynay, line 102: This PR addresses this TODO

PR #3467, opened 2024-06-04
feat(share/pruner): disallow converting pruned node to archival

walldiss:
This PR add safety check, that prevents node runner from converting previously pruned node to archival, because it is not supported.

### DIFF
--- a/pruner/checkpoint.go
+++ b/pruner/checkpoint.go
@@ -13,14 +13,15 @@ import (
 )
 
 var (
-	storePrefix           = datastore.NewKey("pruner")
-	checkpointKey         = datastore.NewKey("checkpoint")
-	errCheckpointNotFound = errors.New("checkpoint not found")
 	// ErrDisallowRevertToArchival is returned when a node has been run with pruner enabled before and
 	// launching it with archival mode.
 	ErrDisallowRevertToArchival = errors.New(
 		"node has been run with pruner enabled before, it is not safe to convert to an archival" +
 			"Run with --experimental-pruning enabled or consider re-initializing the store")
+
+	storePrefix           = datastore.NewKey("pruner")
+	checkpointKey         = datastore.NewKey("checkpoint")
+	errCheckpointNotFound = errors.New("checkpoint not found")
 )
 
 // checkpoint contains information related to the state of the
--- a/pruner/checkpoint.go
+++ b/pruner/checkpoint.go
@@ -7,13 +7,20 @@ import (
 	"fmt"
 
 	"github.com/ipfs/go-datastore"
+	"github.com/ipfs/go-datastore/namespace"
 
 	"github.com/celestiaorg/celestia-node/header"
 )
 
 var (
-	storePrefix   = datastore.NewKey("pruner")
-	checkpointKey = datastore.NewKey("checkpoint")
+	storePrefix           = datastore.NewKey("pruner")
+	checkpointKey         = datastore.NewKey("checkpoint")
+	errCheckpointNotFound = errors.New("checkpoint not found")
+	// ErrDisallowRevertToArchival is returned when a node has been run with pruner enabled before and
+	// launching it with archival mode.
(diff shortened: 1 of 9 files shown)


share/eds/byzantine/share_proof.go
```diff
@@ -72,8 +72,8 @@ func (s *ShareWithProof) ShareWithProofToProto() *pb.Share {
 	}
 }
 
-// GetShareWithProof attempts to get a share with proof for the given share. It first tries to get a row proof
-// and if that fails or proof is invalid, it tries to get a column proof.
+// GetShareWithProof attempts to get a share with proof for the given share. It first tries to get
+// a row proof and if that fails or proof is invalid, it tries to get a column proof.
```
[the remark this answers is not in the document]
Wondertan: Just exclude those from this one :)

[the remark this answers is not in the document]
walldiss: It needs to be fixed. Exclude == another PR

pruner/checkpoint.go
```diff
@@ -7,13 +7,20 @@ import (
...
 	"github.com/celestiaorg/celestia-node/header"
 )
 
 var (
-	storePrefix   = datastore.NewKey("pruner")
-	checkpointKey = datastore.NewKey("checkpoint")
+	storePrefix           = datastore.NewKey("pruner")
+	checkpointKey         = datastore.NewKey("checkpoint")
+	errCheckpointNotFound = errors.New("checkpoint not found")
+	// ErrDisallowRevertToArchival is returned when a node has been run with pruner enabled before and
```
renaynay, line 19: Nit but can we list public var first

pruner/checkpoint.go
```diff
@@ -23,25 +30,60 @@ type checkpoint struct {
...
+	return ds.Put(ctx, checkpointKey, bin)
+}
+
+// getCheckpoint loads the last checkpoint from disk.
+func getCheckpoint(ctx context.Context, ds datastore.Datastore) (*checkpoint, error) {
+	bin, err := ds.Get(ctx, checkpointKey)
 	if err != nil {
 		if errors.Is(err, datastore.ErrNotFound) {
-			return s.initializeCheckpoint(ctx)
+			return nil, errCheckpointNotFound
```
renaynay, line 61: why not just check for datastore.ErrNotFound err in `DetectPreviousRun` ?

pruner/checkpoint.go
```diff
@@ -23,25 +30,60 @@ type checkpoint struct {
...
+
+// loadCheckpoint loads the last checkpoint from disk, initializing it if it does not already exist.
+func (s *Service) loadCheckpoint(ctx context.Context) error {
+	cp, err := getCheckpoint(ctx, s.ds)
+	if err != nil {
+		if errors.Is(err, errCheckpointNotFound) {
+			s.checkpoint = &checkpoint{
+				LastPrunedHeight: 1,
+				FailedHeaders:    map[uint64]struct{}{},
+			}
```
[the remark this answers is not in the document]
renaynay: @walldiss yeah this was a bug before but it's fixed inside of `find.go` in an ugly way 

```
	// ensures genesis block gets pruned
	if lastPruned.Height() == 1 {
		headers = append([]*header.ExtendedHeader{lastPruned}, headers...)
	}

```

If you have a prettier way to do this, let's do it in a diff PR

pruner/checkpoint.go
walldiss: the error is also used for init of checkpoint. I think it is easier to read via err name this way:
```
if err.Is(errCheckpointNotFound){
  initCheckpoint
}
```
instead of 
```
if err.Is(dagstpre.ErrNotFound){
  initCheckpoint
}
```

renaynay: fine with me

renaynay (approved):

cristaloleg (approved):
SGTM

Issue #3361, opened 2024-05-02
Blobstream API should be moved to node


liamsi:
> Proofs for Blobstream (DataRootInclusionProof and ProveShare endpoints on core) are not accessible through node API, so currently all integrations with Blobstream require one to have two endpoints, and two APIs.

related: https://github.com/celestiaorg/celestia-core/issues/1306

rach-id:
Referencing this PR here: https://github.com/celestiaorg/blobstream-node-api/pull/1

PR #3470, opened 2024-06-05
feat: add support for Blobstream API

rach-id:
Closes https://github.com/celestiaorg/celestia-node/issues/3361


go.mod
```diff
@@ -1,18 +1,18 @@
...
 	github.com/BurntSushi/toml v1.3.2
 	github.com/alecthomas/jsonschema v0.0.0-20220216202328-9eeeec9d044b
 	github.com/benbjohnson/clock v1.3.5
-	github.com/celestiaorg/celestia-app v1.9.0
+	github.com/celestiaorg/celestia-app v1.10.1
 	github.com/celestiaorg/go-fraud v0.2.1
 	github.com/celestiaorg/go-header v0.6.1
 	github.com/celestiaorg/go-libp2p-messenger v0.2.0
-	github.com/celestiaorg/nmt v0.21.0
+	github.com/celestiaorg/nmt v0.21.1-0.20240602221058-a81b748b6f51 // TODO replace with release when ready
```
rach-id, line 15: will be replaced once we release nmt

nodebuilder/blobstream/service.go
```diff
@@ -0,0 +1,534 @@
...
+		tuples = append(tuples, DataRootTuple{
+			height:   block.Height(),
+			dataRoot: *(*[32]byte)(block.DataHash),
+		})
+	}
+	return tuples, nil
+}
+
+// ProveShares generates a share proof for a share range.
+// Note: queries the whole EDS to generate the proof.
```
rach-id, line 313: should I create an issue for this?

nodebuilder/blobstream/service.go
```diff
@@ -0,0 +1,534 @@
...
+		return nil, err
+	}
+
+	log.Debugw("converting the blob to shares", "height", height, "commitment", shareCommitment, "namespace", namespace)
+	blobShares, err := blob.BlobsToShares(blb)
+	if err != nil {
+		return nil, err
+	}
+	if len(blobShares) == 0 {
+		// TODO we return the share commitment as hex or some other format?
```
rach-id, line 390: [note to reviewers]
should we return the share commitment as hex or some other format?

nodebuilder/blobstream/service.go
```diff
@@ -0,0 +1,534 @@
...
+			&nmtProof,
+		)
+	}
+
+	// compute the subtree roots of the blob shares
+	log.Debugw("computing the subtree roots", "height", height, "commitment", shareCommitment, "namespace", namespace)
+	var subtreeRoots [][]byte
+	var dataCursor int
+	for _, proof := range nmtProofs {
+		// TODO: do we want directly use the default subtree root threshold or want to allow specifying which version to use?
```
rach-id, line 447: [note to reviewers]
do we want directly use the default subtree root threshold or want to allow specifying which version to use?

nodebuilder/blobstream/service.go
```diff
@@ -0,0 +1,534 @@
...
+// ProveShareToSubtreeRoot generates a share to subtree root inclusion proof
+// Note: this method is not part of the API. It will not be served by any endpoint, however,
+// it can be called directly programmatically.
+func ProveShareToSubtreeRoot(shares [][]byte, shareIndex uint64) (*ResultShareToSubtreeRootProof, error) {
+	_, proofs := merkle.ProofsFromByteSlices(shares)
+	return &ResultShareToSubtreeRootProof{
+		ShareToSubtreeRootProof: ShareToSubtreeRootProof{
+			Proof: *proofs[shareIndex],
+		},
+	}, nil
```
rach-id, line 536: [note to reviewers]
these two methods are not used in the API but I see them as useful for people using the API. should we keep them here? move them somewhere special? create a `utils.go` or `helpers.go` and put them there?

nodebuilder/blobstream/service_test.go
```diff
@@ -0,0 +1,1087 @@
...
+	for subtreeRootIndex, actualProof := range actualProofs {
+		t.Run(fmt.Sprintf("subtreeRootIndex=%d", subtreeRootIndex), func(t *testing.T) {
+			valid, err := actualProof.SubtreeRootToCommitmentProof.Verify(shareCommitment, subtreeRoots[subtreeRootIndex])
+			assert.NoError(t, err)
+			assert.True(t, valid)
+			assert.Equal(t, *expectedProofs[subtreeRootIndex], actualProof.SubtreeRootToCommitmentProof.Proof)
+		})
+	}
+}
+
```
rach-id, line 918: instead of using the generated mock. I decided to implement a custom mock here so that I can test a lot of combinations without having to worry about the mocked p2p network etc.

If you check the `TestProveCommitmentAllCombinations` test, you will see that it's kind of fuzzing tons of combinations.

nodebuilder/blobstream/types.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+// ResultCommitmentProof is an API response that contains a CommitmentProof.
+// A commitment proof is a proof of a blob share commitment to the data root.
+type ResultCommitmentProof struct {
+	CommitmentProof CommitmentProof `json:"commitment_proof"`
+}
+
+// CommitmentProof is an inclusion proof of a commitment to the data root.
+// TODO: Ask reviewers if we need protobuf definitions for this
```
rach-id, line 44: [note to reviewers]
do we need to create protobuf definitions for these types? 
similarly, do we want json marshellers/unmarshallers?

nodebuilder/blobstream/service.go
```diff
@@ -0,0 +1,537 @@
...
+//	        "internalType":"structDataRootTuple",
+//	        "name":"_tuple",
+//	        "type":"tuple"
+//	     }
+//	  ]
+//	}
+//
+// padding the hex representation of the height padded to 32 bytes concatenated to the data root.
+// For more information, refer to:
+// https://github.com/celestiaorg/quantum-gravity-bridge/blob/master/src/DataRootTuple.sol
```
rach-id, line 166:
```suggestion
// https://github.com/celestiaorg/blobstream-contracts/blob/master/src/DataRootTuple.sol
```

nodebuilder/blobstream/service.go
```diff
@@ -0,0 +1,537 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &ResultShareProof{ShareProof: proof}, nil
+}
+
+// ProveCommitment generates a commitment proof for a share commitment.
+// It takes as argument the height of the block containing the blob of data, its
+// namespace and its share commitment.
+// Note: queries the whole EDS to generate the proof.
```
rach-id, line 368: similar, should we create an issue for this to track this improvement?

rach-id:
the remaining linting complains come from files that I didn't touch. I will open a separate PR to handle them

nodebuilder/blobstream/types.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+// ResultShareProof is the API response that contains a ShareProof.
+// A share proof is a proof of a set of shares to the data root.
+type ResultShareProof struct {
+	ShareProof types.ShareProof `json:"share_proof"`
+}
+
+// ResultCommitmentProof is an API response that contains a CommitmentProof.
+// A commitment proof is a proof of a blob share commitment to the data root.
```
damiannolan, line 39: 🙏🏻

nodebuilder/blobstream/types.go
damiannolan: I think protobuf definitions is a good call! Would be useful for consuming and sending to e.g. on-chain lightclient via std cosmos-sdk grpc apis for something like ibc. 

However, it might be good to consider where these protobufs are maintained. Similarly, I see this PRs consumes a number of pkgs from celestia-app. I think solving https://github.com/celestiaorg/go-square/issues/55 would be a step towards further reducing the dependency tree - as I assume ideally, node should not rely on app.

nodebuilder/blobstream/service.go
damiannolan: does this api essentially supersede the `blob.GetProof` api? I don't really see a need for it to exist if this is delivered. However, I'm unsure if `blobstream` if focused on some eth integration stuff or this is meant to be a general purpose proof api (limited understanding here, apologies!)

damiannolan (commented):
Thanks for sharing this PR @rach-id. I think this closes https://github.com/celestiaorg/celestia-node/issues/3473 as well. I left a couple of comments 🙏🏻 🚀

nodebuilder/blobstream/service.go
rach-id: `blob.GetProof` returns a different proof. It returns the proof of the namespace in the rows that contain the blob, which is proving shares >= the shares of the blob.

nodebuilder/blobstream/types.go
rach-id: makes sense 👍  ideally, we have the types defined in a single place

api/rpc_test.go
```diff
@@ -22,6 +24,7 @@ import (
 	"github.com/celestiaorg/celestia-node/nodebuilder"
 	"github.com/celestiaorg/celestia-node/nodebuilder/blob"
 	blobMock "github.com/celestiaorg/celestia-node/nodebuilder/blob/mocks"
+	BlobstreamMock "github.com/celestiaorg/celestia-node/nodebuilder/blobstream/mocks"
```
cristaloleg, line 27: Nit: packages starts with lower case

rach-id: e4c59107d64cee30b7d07b33688917708f3e191c

nodebuilder/blobstream/blobstream.go
```diff
@@ -0,0 +1,95 @@
...
+	// ProveCommitment generates a commitment proof for a share commitment.
+	ProveCommitment(
+		ctx context.Context,
+		height uint64,
+		namespace share.Namespace,
+		shareCommitment []byte,
+	) (*ResultCommitmentProof, error)
+}
+
+type Internal struct {
```
cristaloleg, line 39: In other files we keep `Interal` as an embedded struct field. Not sure how important it's, commenting just for the consistency 🤷

rach-id: [f8a3066](https://github.com/celestiaorg/celestia-node/pull/3470/commits/f8a30668432f6aae6254a04f9b552b7f3ef5a36a)

nodebuilder/blobstream/types.go
```diff
@@ -0,0 +1,195 @@
...
+type ResultCommitmentProof struct {
+	CommitmentProof CommitmentProof `json:"commitment_proof"`
+}
+
+// CommitmentProof is an inclusion proof of a commitment to the data root.
+// TODO: Ask reviewers if we need protobuf definitions for this
+type CommitmentProof struct {
```
insiderbyte, line 46: Feels like `CommitmentProof` could be pretty heavy it terms of memory usage. May be we should consider having a pointer to the receivers of its methods instead of value receivers?

nodebuilder/blobstream/blobstream.go
```diff
@@ -0,0 +1,93 @@
...
+
+var _ Module = (*API)(nil)
+
+// Module defines the API related to interacting with the proofs
+//
+//go:generate mockgen -destination=mocks/api.go -package=mocks . Module
+type Module interface {
+	// DataCommitment collects the data roots over a provided ordered range of blocks,
+	// and then creates a new Merkle root of those data roots. The range is end exclusive.
+	DataCommitment(ctx context.Context, start, end uint64) (*ResultDataCommitment, error)
```
insiderbyte, line 17:
```suggestion
	GetDataCommitment(ctx context.Context, start, end uint64) (*ResultDataCommitment, error)
```

nodebuilder/blobstream/blobstream.go
```diff
@@ -0,0 +1,93 @@
...
+
+	// DataRootInclusionProof creates an inclusion proof for the data root of block
+	// height `height` in the set of blocks defined by `start` and `end`. The range
+	// is end exclusive.
+	DataRootInclusionProof(
```
insiderbyte, line 22:
```suggestion
	GetDataRootInclusionProof(
```

insiderbyte (changes requested):
General comment: we are using nodebuilder folder only for creating Modules. Could you please create a blobstream package in the root and move all logic there, to follow node's style.

nodebuilder/blobstream/blobstream.go
vgonkivs, line 17:
```suggestion
	GetDataCommitment(ctx context.Context, start, end uint64) (*ResultDataCommitment, error)
```

nodebuilder/blobstream/blobstream.go
vgonkivs, line 22:
```suggestion
	GetDataRootInclusionProof(
```

nodebuilder/blobstream/types.go
vgonkivs, line 46: Can we make all its methods with pointers to a receiver? It feels like it could be pretty heavy.

liamsi:
> Any suggestions for better names @renaynay @adlerjohn @liamsi?

Yeah, I also raised this somewhere else. So ideally:

- anything shares related would be folded into the shares "module"
- hence, ProveShares and ProveCommitment could be moved into the shares module
- even better: shares.GetProof would actually accommodate ProveShares 
-  GetDataCommitment and DataRootInclusionProof are really very blobstream specific but they could live in a "blocks" module as they operate over a range of blocks/block heights. The closest we have for this is the header module.

nodebuilder/blobstream/service.go
```diff
@@ -0,0 +1,623 @@
...
+	if padErr != nil {
+		return nil, padErr
+	}
+	return paddedBytes, nil
+}
+
+// DataRootTuple contains the data that will be used to create the QGB commitments.
+// The commitments will be signed by orchestrators and submitted to an EVM chain via a relayer.
+// For more information: https://github.com/celestiaorg/quantum-gravity-bridge/blob/master/src/DataRootTuple.sol
+type DataRootTuple struct {
```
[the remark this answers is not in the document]
distractedm1nd: Ah I see, I misinterpreted the original godoc of `ResultDataCommitment`. So `GetDataCommitment` should only return the datahash part

Fulfilling the encoding/binary interface is still probably a more idiomatic way to do this though

nodebuilder/blobstream/service.go
Wondertan: The GetProof will stay as a canonical way to serve proofs for blobs long term. The GetProof is supposed to return the proof of just the blob, rather than the whole namespace. You can consider current behaviour a bug.

Wondertan: Fetching whole EDS is okay short term solution and warrants an issue.

api/rpc_test.go
```diff
@@ -8,6 +8,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/celestiaorg/celestia-node/nodebuilder/blobstream"
```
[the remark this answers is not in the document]
walldiss: `make fmt` will do the trick

nodebuilder/blobstream/service.go
Wondertan: Yes, that's the goal. The DataAvailbailityHeader is a detail of Celestia protocol that should have been abstracted away; similarly, the proofs our API provides should not prove to it, but to a single DataRoot hash.

api/rpc_test.go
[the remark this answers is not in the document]
rach-id: [a409722](https://github.com/celestiaorg/celestia-node/pull/3470/commits/a409722021cddc86063090ababb0cb3e14570b42)

api/rpc/client/client.go
```diff
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"net/http"
 
+	"github.com/celestiaorg/celestia-node/nodebuilder/blobstream"
```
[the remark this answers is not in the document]
rach-id: [a409722](https://github.com/celestiaorg/celestia-node/pull/3470/commits/a409722021cddc86063090ababb0cb3e14570b42)

nodebuilder/blobstream/blobstream.go
rach-id: [44f17a1](https://github.com/celestiaorg/celestia-node/pull/3470/commits/44f17a1dfc90532ef7805b3a4006541665735d76)

nodebuilder/blobstream/blobstream.go
rach-id: [5fc0203](https://github.com/celestiaorg/celestia-node/pull/3470/commits/5fc0203f80f8b243766303864a639192820dc5c0)

nodebuilder/blobstream/types.go
rach-id: [350d950](https://github.com/celestiaorg/celestia-node/pull/3470/commits/350d9503b3289cd9f939eab2551414686d5e115a)

nodebuilder/blobstream/types.go
rach-id: [411fc20](https://github.com/celestiaorg/celestia-node/pull/3470/commits/411fc209817aab54d2b0e53e40babd33547e5d7a)

nodebuilder/blobstream/types.go
```diff
@@ -0,0 +1,195 @@
...
+		return fmt.Errorf(
+			"the number of subtree roots %d should be bigger than the number of subtree root proofs %d",
+			len(commitmentProof.SubtreeRoots),
+			len(commitmentProof.SubtreeRootProofs),
+		)
+	}
+	if len(commitmentProof.SubtreeRootProofs) != len(commitmentProof.RowProof.Proofs) {
+		return fmt.Errorf(
+			"the number of subtree root proofs %d should be equal to the number of row root proofs %d",
+			len(commitmentProof.SubtreeRoots),
```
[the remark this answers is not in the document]
rach-id: [9e48e55](https://github.com/celestiaorg/celestia-node/pull/3470/commits/9e48e55ca81feb2b713fa5b6460e59ff86aee770)

nodebuilder/blobstream/blobstream.go
```diff
@@ -0,0 +1,93 @@
...
+		ctx context.Context,
+		height int64,
+		start, end uint64,
+	) (*ResultDataRootInclusionProof, error)
+
+	// ProveShares generates a share proof for a share range.
+	ProveShares(ctx context.Context, height, start, end uint64) (*ResultShareProof, error)
+	// ProveCommitment generates a commitment proof for a share commitment.
```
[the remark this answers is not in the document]
rach-id: [f24171e](https://github.com/celestiaorg/celestia-node/pull/3470/commits/f24171ecbd191595bdd6f085c7942835d621a47d)

nodebuilder/blobstream/service.go
Wondertan: @rach-id, agreed, let's make this in a separate PR, thank you!

nodebuilder/blobstream/service.go
```diff
@@ -0,0 +1,623 @@
...
+	return &Service{
+		blobServ:   blobMod,
+		headerServ: headerMod,
+		shareServ:  shareMod,
+	}
+}
+
+// DataCommitment collects the data roots over a provided ordered range of blocks,
+// and then creates a new Merkle root of those data roots. The range is end exclusive.
+func (s *Service) DataCommitment(ctx context.Context, start, end uint64) (*ResultDataCommitment, error) {
```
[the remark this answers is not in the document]
rach-id: [44f17a1](https://github.com/celestiaorg/celestia-node/pull/3470/commits/44f17a1dfc90532ef7805b3a4006541665735d76)

nodebuilder/blobstream/service.go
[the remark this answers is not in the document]
rach-id: what does this mean? what needs to be changed?

nodebuilder/blobstream/service.go
rach-id: https://github.com/celestiaorg/celestia-node/issues/3516

nodebuilder/blobstream/service.go
rach-id:
```suggestion
```

rach-id:
> Nodebuilder folder is using only for the modules creation. Maybe you should add a blobstream package and move all logic there?

@insiderbyte where do you mean I should put the blobstream package? in the root?

nodebuilder/blobstream/types.go
```diff
@@ -0,0 +1,195 @@
...
+// A share proof is a proof of a set of shares to the data root.
+type ResultShareProof struct {
+	ShareProof types.ShareProof `json:"share_proof"`
+}
+
+// ResultCommitmentProof is an API response that contains a CommitmentProof.
+// A commitment proof is a proof of a blob share commitment to the data root.
+type ResultCommitmentProof struct {
+	CommitmentProof CommitmentProof `json:"commitment_proof"`
+}
```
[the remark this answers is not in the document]
distractedm1nd: Hlib and I discussed this shortly, and we decided we would still prefer them to be unwrapped - we think backwards compatibility is liekly overkill here, or do you have a good example of something that may be added?

My thinking here is that these are actually pretty much low-level types that are processed directly in the smart contracts (as far as I imagine? haven't read them) and that they won't be doing any parsing out of extra additional data

nodebuilder/blobstream/service.go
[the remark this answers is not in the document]
distractedm1nd: This means implementing `MarshalBinary` and `UnmarshalBinary` onto `DataRootTuple` to fulfill the encoding/binary interface, to replace `EncodeDataRootTuple`. I can gladly help out here or send a diff if it helps!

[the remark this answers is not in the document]
rach-id: But we only need to encode it. do you mean fulfilling the interface and making the unmarshalling a no-op?

nodebuilder/blobstream/types.go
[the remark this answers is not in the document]
rach-id: Agreed, but I guess the same reasoning lead to having the blob proof to be `[]*nmt.Proof` and ended up needing to be a breaking change to change it to contain a proof to the data root. 

From my experience, it's always good to separate the problematical types from the API ones. But I am fine with any. Let me know if you prefer them to be unwrapped and I'll do it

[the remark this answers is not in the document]
ramin: i think consensus it to unwrap please @rach-id. @distractedm1nd and @Wondertan do you agree?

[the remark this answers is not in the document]
rach-id: cool, I'll unwrap as soon as I get confirmation.

ramin:
@rach-id so there is a general desire to follow @liamsi 's comment about decomposing this, such that instead of "blobstream" being a whole module, the new API functionality lives in each of their relative existing modules, with some functionality moved to another package (like the "blocks" one that he also suggested). There was a promise that everyone would chime in on reviews and suggestions of where to place things, I'll chase that - but we can work on together too if you want to suggest a diff layout...

liamsi:
I like header better than block.

distractedm1nd:
agreed

nodebuilder/blobstream/types.go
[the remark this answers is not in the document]
rach-id: types unwrapped: [02be8ac](https://github.com/celestiaorg/celestia-node/pull/3470/commits/02be8ac38d1bab9f8951ef730e578eac445291b8)

go.mod
rach-id: [3b2680b](https://github.com/celestiaorg/celestia-node/pull/3470/commits/3b2680b6bee005aa82fdc5a9ced69b0a77aa46ee)

nodebuilder/header/data_commitment.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+	if start >= end {
+		return fmt.Errorf("end block is smaller or equal to the start block")
+	}
+	heightsRange := end - start
+	if heightsRange > uint64(dataCommitmentBlocksLimit) {
+		return fmt.Errorf("the query exceeds the limit of allowed blocks %d", dataCommitmentBlocksLimit)
+	}
+
+	currentHeader, err := s.NetworkHead(ctx)
```
renaynay, line 123: We should only do this request against `LocalHead`, not `NetworkHead` please since we need to load the range and that range should already exist in the store.

nodebuilder/header/data_commitment.go
```diff
@@ -0,0 +1,242 @@
...
+	_, proofs := merkle.ProofsFromByteSlices(dataRootEncodedTuples)
+	return proofs[height-int64(tuples[0].height)], nil
+}
+
+// fetchDataRootTuples takes an end exclusive range of heights and fetches its
+// corresponding data root tuples.
+func (s *Service) fetchDataRootTuples(ctx context.Context, start, end uint64) ([]DataRootTuple, error) {
+	tuples := make([]DataRootTuple, 0, end-start)
+	for height := start; height < end; height++ {
+		block, err := s.GetByHeight(ctx, height)
```
renaynay, line 229: please use a `GetRangeByHeight` request.

nodebuilder/header/data_commitment.go
```diff
@@ -0,0 +1,242 @@
...
+	return root, nil
+}
+
+// validateDataRootInclusionProofRequest validates the request to generate a data root
+// inclusion proof.
+func (s *Service) validateDataRootInclusionProofRequest(
+	ctx context.Context,
+	height, start, end uint64,
+) error {
+	err := s.validateDataCommitmentRange(ctx, start, end)
```
renaynay, line 179: can you combine these two methods please, it's fine if it's a bit longer of a function.

nodebuilder/header/data_commitment.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+	return nil
+}
+
+// proveDataRootTuples returns the merkle inclusion proof for a height.
+func proveDataRootTuples(tuples []DataRootTuple, height int64) (*merkle.Proof, error) {
+	if len(tuples) == 0 {
+		return nil, fmt.Errorf("cannot prove an empty list of tuples")
+	}
+	if height < 0 {
```
renaynay, line 199: again let's extract out into an error `ErrZeroHeight` or something like that.

nodebuilder/header/data_commitment.go
```diff
@@ -0,0 +1,242 @@
...
+		return nil, fmt.Errorf("cannot prove a strictly negative height %d", height)
+	}
+	currentHeight := tuples[0].height - 1
+	for _, tuple := range tuples {
+		if tuple.height != currentHeight+1 {
```
renaynay, line 204: this function is being called after `fetchDataRootTuples`, which - if you use `GetRange..` request, guarantees a contiguous ascending range to be returned. 

You do not need to do this validation logic here, function can document that it assumes the tuples to be contiguous and complete.

nodebuilder/header/data_commitment.go
```diff
@@ -0,0 +1,242 @@
...
+	dataRootEncodedTuples := make([][]byte, 0, len(tuples))
+	for _, tuple := range tuples {
+		encodedTuple, err := EncodeDataRootTuple(
+			tuple.height,
+			tuple.dataRoot,
+		)
+		if err != nil {
+			return nil, err
+		}
+		dataRootEncodedTuples = append(dataRootEncodedTuples, encodedTuple)
```
renaynay, line 218: why can't this also be done in `fetchDataRootTuples` ? both callers of `fetchDataRootTuples` want an encoded version of the tuples.

nodebuilder/header/data_commitment.go
```diff
@@ -0,0 +1,242 @@
...
+//	        "name":"_tuple",
+//	        "type":"tuple"
+//	     }
+//	  ]
+//	}
+//
+// padding the hex representation of the height padded to 32 bytes concatenated to the data root.
+// For more information, refer to:
+// https://github.com/celestiaorg/blobstream-contracts/blob/master/src/DataRootTuple.sol
+func EncodeDataRootTuple(height uint64, dataRoot [32]byte) ([]byte, error) {
```
renaynay, line 96: why does this need to be exported? When would users need to deal with data root tuples directly?

renaynay: Similarly, why does DataRootTuple need to be exported?

renaynay (changes requested):
Generally, the only broader issue I have with the PR is that the additional methods on header service could just exist in a separate package and whatever construction you decide to make there would just take a header service. The methods feel too opinionated for me to feel like they belong on header module.

@Wondertan also has additional comments regarding `GetCommitmentProof` that he'll discuss with you separately and I agree to them.

share/eds/eds.go
```diff
@@ -271,3 +274,30 @@ func ReadEDS(ctx context.Context, r io.Reader, root share.DataHash) (eds *rsmt2d
 	}
 	return eds, nil
 }
+
+// ProveShares generates a share proof for a share range.
+// The share range, defined by start and end, is end-exclusive.
+func ProveShares(eds *rsmt2d.ExtendedDataSquare, start, end int) (*types.ShareProof, error) {
```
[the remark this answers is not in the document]
rach-id: I was discussing with Vlad and he suggested to add it in the EDS package. Ideally, we have it in the go-square, agreed

nodebuilder/header/data_commitment.go
```diff
@@ -0,0 +1,242 @@
...
+		return byt, nil
+	}
+	tmp := make([]byte, length)
+	copy(tmp[length-l:], byt)
+	return tmp, nil
+}
+
+// To32PaddedHexBytes takes a number and returns its hex representation padded to 32 bytes.
+// Used to mimic the result of `abi.encode(number)` in Ethereum.
+func To32PaddedHexBytes(number uint64) ([]byte, error) {
```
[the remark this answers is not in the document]
rach-id: this was added in code just for these endpoints. Once we add these endpoints in node, we will delete them in code. So, this code should be here

nodebuilder/header/data_commitment.go
```diff
@@ -0,0 +1,242 @@
...
+		return nil, padErr
+	}
+	return paddedBytes, nil
+}
+
+// DataRootTuple contains the data that will be used to create the QGB commitments.
+// The commitments will be signed by orchestrators and submitted to an EVM chain via a relayer.
+// For more information:
+// https://github.com/celestiaorg/quantum-gravity-bridge/blob/master/src/DataRootTuple.sol
+type DataRootTuple struct {
```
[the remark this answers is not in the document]
rach-id: https://github.com/celestiaorg/celestia-node/pull/3470#discussion_r1683215078

nodebuilder/header/service.go
```diff
@@ -141,3 +141,62 @@ func (s *Service) Subscribe(ctx context.Context) (<-chan *header.ExtendedHeader,
...
+	dataCommitment := DataCommitment(root)
+	return &dataCommitment, nil
+}
+
+// GetDataRootInclusionProof creates an inclusion proof for the data root of block
+// height `height` in the set of blocks defined by `start` and `end`. The range
+// is end exclusive.
+func (s *Service) GetDataRootInclusionProof(
+	ctx context.Context,
+	height int64,
```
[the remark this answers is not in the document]
rach-id: in code, heights are always int64

nodebuilder/header/data_commitment.go
```diff
@@ -0,0 +1,242 @@
...
+	paddedHeight, err := To32PaddedHexBytes(height)
+	if err != nil {
+		return nil, err
+	}
+	return append(paddedHeight, dataRoot[:]...), nil
+}
+
+// dataCommitmentBlocksLimit The maximum number of blocks to be used to create a data commitment.
+// It's a local parameter to protect the API from creating unnecessarily large commitments.
+const dataCommitmentBlocksLimit = 10_000 // ~33 hours of blocks assuming 12-second blocks.
```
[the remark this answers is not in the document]
rach-id: We just need to set some limit on the API for nodes not to be spammed with creating large proofs unnecessarily.
Also, we could have it as a parameter, but I don't think it's worth doing it now since no one will ask for proofs that are larger than 2000 blocks

nodebuilder/header/header.go
```diff
@@ -43,6 +43,19 @@ type Module interface {
...
 	Subscribe(ctx context.Context) (<-chan *header.ExtendedHeader, error)
+
+	// GetDataCommitment collects the data roots over a provided ordered range of blocks,
+	// and then creates a new Merkle root of those data roots. The range is end exclusive.
+	GetDataCommitment(ctx context.Context, start, end uint64) (*DataCommitment, error)
+
+	// GetDataRootInclusionProof creates an inclusion proof for the data root of block
+	// height `height` in the set of blocks defined by `start` and `end`. The range
+	// is end exclusive.
+	GetDataRootInclusionProof(
```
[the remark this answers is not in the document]
rach-id: where do you suggest to put the example? we have them documented in the docs and will update the docs to use node once we merge/release this

share/eds/eds.go
```diff
@@ -271,3 +274,30 @@ func ReadEDS(ctx context.Context, r io.Reader, root share.DataHash) (eds *rsmt2d
...
+	log.Debugw("proving share range", "start", start, "end", end)
+	if start == end {
+		return nil, fmt.Errorf("start share cannot be equal to end share")
+	}
+	if start > end {
+		return nil, fmt.Errorf("start share %d cannot be greater than end share %d", start, end)
+	}
```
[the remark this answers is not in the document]
rach-id: once https://github.com/celestiaorg/celestia-app/pull/3709 is merged, i'll make a release and remove these checks from here

nodebuilder/header/data_commitment.go
rach-id: [8c56c74](https://github.com/celestiaorg/celestia-node/pull/3470/commits/8c56c741f9ad6f64b2ea1a1ffef534267fb46968)

nodebuilder/header/data_commitment.go
```diff
@@ -0,0 +1,242 @@
...
+
+// validateDataCommitmentRange runs basic checks on the asc sorted list of
+// heights that will be used subsequently in generating data commitments over
+// the defined set of heights.
+func (s *Service) validateDataCommitmentRange(ctx context.Context, start, end uint64) error {
+	if start == 0 {
+		return fmt.Errorf("the start block is 0")
```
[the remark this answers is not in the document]
rach-id: [8c56c74](https://github.com/celestiaorg/celestia-node/pull/3470/commits/8c56c741f9ad6f64b2ea1a1ffef534267fb46968)

nodebuilder/share/share.go
```diff
@@ -87,8 +110,31 @@ func (api *API) GetSharesByNamespace(
...
 	share.Availability
+	hs headerServ.Module
 }
 
 func (m module) SharesAvailable(ctx context.Context, header *header.ExtendedHeader) error {
 	return m.Availability.SharesAvailable(ctx, header)
 }
+
+func (m module) GetRange(ctx context.Context, height uint64, start, end int) (*GetRangeResult, error) {
+	if height == 0 {
```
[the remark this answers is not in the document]
rach-id: [8c56c74](https://github.com/celestiaorg/celestia-node/pull/3470/commits/8c56c741f9ad6f64b2ea1a1ffef534267fb46968)

nodebuilder/header/data_commitment.go
rach-id: good catch, I will un-export these. They were exported because they were used by vanilla blobstream

nodebuilder/header/data_commitment.go
rach-id: you mean to reduce the duplicated code? I think it's fine since we use the information from the non encoded data root tuple to know the starting height of the range. Otherwise, we will need to pass it as a parameter.

Anyway, I am fine with both, let me know what you prefer

nodebuilder/header/data_commitment.go
rach-id: the validation is used twice. If we remove that function and put the logic here, we will have to maintain two validation logic. I think it's better in a separate method

nodebuilder/header/service.go
```diff
@@ -141,3 +141,62 @@ func (s *Service) Subscribe(ctx context.Context) (<-chan *header.ExtendedHeader,
...
+		return nil, err
+	}
+	// Create data commitment
+	dataCommitment := DataCommitment(root)
+	return &dataCommitment, nil
+}
+
+// GetDataRootInclusionProof creates an inclusion proof for the data root of block
+// height `height` in the set of blocks defined by `start` and `end`. The range
+// is end exclusive.
```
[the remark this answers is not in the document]
rach-id: 2c53cb0

share/eds/eds.go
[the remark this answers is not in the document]
renaynay: Would be nice to make a comment there in the code and @ yourself not to forget.

nodebuilder/header/service.go
```diff
@@ -13,6 +13,9 @@ import (
 	modfraud "github.com/celestiaorg/celestia-node/nodebuilder/fraud"
 )
 
+// ErrHeightNegative returned when the provided block height is <= 0.
+var ErrHeightNegative = errors.New("height is negative")
```
renaynay, line 17: Let's call it ErrInvalidHeight and say height is invalid when <= 0

rach-id: changed heights to uint64 to be consistent with your repo and switched to ErrHeightZero: [8bc6a66](https://github.com/celestiaorg/celestia-node/pull/3470/commits/8bc6a664b4c2c9ad1220d0b42abd9c71d6f82822)

nodebuilder/header/service.go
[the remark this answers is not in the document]
rach-id: switched them to uint64: [8bc6a66](https://github.com/celestiaorg/celestia-node/pull/3470/commits/8bc6a664b4c2c9ad1220d0b42abd9c71d6f82822)

nodebuilder/header/data_commitment.go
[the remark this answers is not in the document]
rach-id: and then: [8bc6a66](https://github.com/celestiaorg/celestia-node/pull/3470/commits/8bc6a664b4c2c9ad1220d0b42abd9c71d6f82822)

nodebuilder/header/data_commitment.go
```diff
@@ -0,0 +1,242 @@
...
+		return fmt.Errorf(
+			"end block %d is higher than current chain height %d",
+			end,
+			currentHeader.Height(),
+		)
+	}
+
+	currentLocalHeader, err := s.LocalHead(ctx)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
rach-id: [18452f6](https://github.com/celestiaorg/celestia-node/pull/3470/commits/18452f6da5f8c1313463ea03f6f1faf8ed7d0cb6)

nodebuilder/header/data_commitment.go
rach-id: [18452f6](https://github.com/celestiaorg/celestia-node/pull/3470/commits/18452f6da5f8c1313463ea03f6f1faf8ed7d0cb6)

nodebuilder/header/data_commitment.go
```diff
@@ -0,0 +1,242 @@
...
+	return append(paddedHeight, dataRoot[:]...), nil
+}
+
+// dataCommitmentBlocksLimit The maximum number of blocks to be used to create a data commitment.
+// It's a local parameter to protect the API from creating unnecessarily large commitments.
+const dataCommitmentBlocksLimit = 10_000 // ~33 hours of blocks assuming 12-second blocks.
+
+// validateDataCommitmentRange runs basic checks on the asc sorted list of
+// heights that will be used subsequently in generating data commitments over
+// the defined set of heights.
```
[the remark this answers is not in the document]
rach-id: [e5f6a05](https://github.com/celestiaorg/celestia-node/pull/3470/commits/e5f6a0541c7b8f92db5441678eee73efc79476ca)

nodebuilder/header/data_commitment.go
```diff
@@ -0,0 +1,242 @@
...
+func (s *Service) validateDataCommitmentRange(ctx context.Context, start, end uint64) error {
+	if start == 0 {
+		return fmt.Errorf("the start block is 0")
+	}
+	if start >= end {
+		return fmt.Errorf("end block is smaller or equal to the start block")
+	}
```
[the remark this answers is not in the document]
rach-id: [e5d1719](https://github.com/celestiaorg/celestia-node/pull/3470/commits/e5d1719947f91b7504f03f26862f5636458765a5)

nodebuilder/header/data_commitment.go
renaynay: I'd rather dedup + eliminate need for non-encoded tuple type, and pass start to `proveDataRootTuples`

nodebuilder/header/header.go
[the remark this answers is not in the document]
renaynay: That's fine

nodebuilder/header/data_commitment.go
rach-id: [9d46758](https://github.com/celestiaorg/celestia-node/pull/3470/commits/9d4675857994f89431985c3ea9191607a9712858)

nodebuilder/header/service.go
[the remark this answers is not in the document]
renaynay: yes I know tendermint uses int64 - we made a deliberate decision to use uint64 throughout node but i just saw a mix here of using both int64 and uint64 and was wondering why

nodebuilder/header/data_commitment.go
rach-id: [b9d7da2](https://github.com/celestiaorg/celestia-node/pull/3470/commits/b9d7da2abe995a162e0e9fcd7fc53981f5d261e7)

share/eds/eds.go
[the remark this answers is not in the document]
rach-id: [8276e5a](https://github.com/celestiaorg/celestia-node/pull/3470/commits/8276e5ac2b1989e30182f9b70eacb3c84b8dc1c8)

nodebuilder/blob/blob.go
```diff
@@ -34,15 +34,52 @@ type Module interface {
...
 	// given height and under the namespace.
 	Included(_ context.Context, height uint64, _ share.Namespace, _ *blob.Proof, _ blob.Commitment) (bool, error)
+	// GetCommitmentProof generates a commitment proof for a share commitment.
+	GetCommitmentProof(
+		ctx context.Context,
+		height uint64,
+		namespace share.Namespace,
+		shareCommitment []byte,
+	) (*blob.CommitmentProof, error)
 }
```
Wondertan, line 44: The comment Rene referred to is about combining this with `GetProof`. As we decided to integrate APIs Blobstream needs into the node nicely, we should also the same for `GetProof`. I am open to doing that in a separate PR though if you think that's better @rach-id, but as long we release these two PRs together.

Wondertan: The tricky part is unifying the new `CommitmentProof` structure with an existing `Proof` and making sure that `Included` method support it as well

rach-id: I prefer to do it in a separate PR. This PR has been open for so long, I don't think we should delay merging it any more. Hopefully, after handling the feedback above, we can go ahead and merge it. 

Created this issue for this: https://github.com/celestiaorg/celestia-node/issues/3582

rach-id:
@renaynay @Wondertan blobstream module re-introduced and endpoints removed from header module. Please review when you have time :D

nodebuilder/header/data_root_tuple_root.go
```diff
@@ -211,19 +204,21 @@ func proveDataRootTuples(tuples []dataRootTuple, height uint64) (*merkle.Proof,
...
+	headerRange, err := s.GetRangeByHeight(ctx, startHeader, end)
+	if err != nil {
+		return nil, err
+	}
+	for _, header := range headerRange {
 		tuples = append(tuples, dataRootTuple{
-			height:   block.Height(),
-			dataRoot: *(*[32]byte)(block.DataHash),
+			height:   header.Height(),
+			dataRoot: *(*[32]byte)(header.DataHash),
```
renaynay: FYI header @ `start` height will not be included in the tuples, is this intentional?

nodebuilder/header/data_commitment.go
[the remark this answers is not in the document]
renaynay: Yes i understand the reasoning for it, I just think you should use the parameter we defined to calculate the approximate num blocks (`p2p.BlockTime`)

nodebuilder/blobstream/data_root_tuple_root.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	heightsRange := end - start
+	if heightsRange > uint64(dataRootTupleRootBlocksLimit) {
+		return fmt.Errorf("the query exceeds the limit of allowed blocks %d", dataRootTupleRootBlocksLimit)
+	}
+
+	currentLocalHeader, err := s.headerServ.LocalHead(ctx)
+	if err != nil {
+		return fmt.Errorf("couldn't get the local head to validate the data root tuple root range%w", err)
```
renaynay, line 120:
```suggestion
		return fmt.Errorf("could not get the local head to validate the data root tuple root range: %w", err)
```

nodebuilder/header/data_root_tuple_root.go
rach-id: Thanks 🙏  
[644c279](https://github.com/celestiaorg/celestia-node/pull/3470/commits/644c2792f91991a83b4d372db019a5561b5b939c)

nodebuilder/header/data_commitment.go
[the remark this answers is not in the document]
rach-id: I see, the 10_000 is just a certain number of blocks that we think is enough. You mean using `p2p.BlockTime` to calculate the corresponding period in the comment?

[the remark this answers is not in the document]
rach-id: [b8d7a44](https://github.com/celestiaorg/celestia-node/pull/3470/commits/b8d7a444198905245439dc8743ac26e6f6f9a72f)

nodebuilder/blobstream/data_root_tuple_root.go
```diff
@@ -180,15 +181,21 @@ func proveDataRootTuples(encodedDataRootTuples [][]byte, rangeStartHeight, heigh
 // end is not included in the range.
 func (s *Service) fetchEncodedDataRootTuples(ctx context.Context, start, end uint64) ([][]byte, error) {
 	encodedDataRootTuples := make([][]byte, 0, end-start)
+	headers := make([]*nodeheader.ExtendedHeader, 0)
```
renaynay, line 185: nit but you can already cap the slice here bc you know the expected size

```
make([]*nodeheader.ExtendedHeader, 0, end-start)
```

renaynay (approved):

distractedm1nd (dismissed):

nodebuilder/blobstream/data_root_tuple_root.go
rach-id: [1183e46](https://github.com/celestiaorg/celestia-node/pull/3470/commits/1183e46a694e00b659e82fad3b7981ab20c8a957)

renaynay (approved):

distractedm1nd (approved):

cmwaters (commented):
This comment is probably a little late and I acknowledge that but I don't think this is generic and widespread enough to warrant being a part of the node API. This will probably only be run by a single machine. We don't need to bloat the repo with this added module. I feel like it can be implemented completely outside (in the blobstreamX relayer code) using the header module and constructing the merkle tree from the results.

We can still keep the blob inclusion proofs. Although I think we should not use the word `Commitment` publicly because it's not clear what it means. I would advocate `GetInclusionProof`

rach-id:
> This will probably only be run by a single machine

This will be run by zk-Blobstream provers, we now have:

- BlobstreamX with some operators
- Blobstream0: close to production

And also by users/validators in case of a fraud proof. For the commitment proof, it can also be used by sovereign rollups in their fraud proofs (still to be defined/discussed). So, there are a few parties that will use these endpoints.

> We don't need to bloat the repo with this added module. I feel like it can be implemented completely outside (in the blobstreamX relayer code) using the header module and constructing the merkle tree from the results

I don't think it's bloating the repo. The share proof and commitment proof belong to this repo. For the `GetDataRootTupleRoot` and `GetDataRootTupleInclusionProof`, they can be implemented somewhere else (I did the first implementation here: https://github.com/celestiaorg/blobstream-node-api/pull/1), but Blobstream is part of the Celestia protocol in the short-mid term, so it's fine to have its API as part of the official implementation of node. 

> Although I think we should not use the word Commitment publicly because it's not clear what it means. I would advocate GetInclusionProof

For this, it will replace the `blob.Proof` in here in a subsequent PR. We just didn't want to add more work to this PR as it's been open for such a long time. Check here for corresponding issue: https://github.com/celestiaorg/celestia-node/issues/3582

vgonkivs (approved):

walldiss (approved):

cmwaters:
>  For the commitment proof, it can also be used by sovereign rollups in their fraud proofs (still to be defined/discussed). 

I think the commitment proof (and share proof) makes sense. It's a generic use case that you want to prove inclusion of the blob (and shares). My original push back was to adding the blobstream module. Can't that logic be run by the "relayer", "prover" which makes the calls to the node's API rather than being inside the node itself.

> but Blobstream is part of the Celestia protocol in the short-mid term

I don't view Blobstream as core to the DA protocol. It's a service built "on top off" Celestia.

rach-id:
> Can't that logic be run by the "relayer", "prover" which makes the calls to the node's API rather than being inside the node itself.

you're right, it can. I just checked our current implementations of Blobstream, BlobstreamX and Blobstream0, and they both rewrite that logic, so they're not using these endpoints. So the main usage of those two endpoints would be when generating an inclusion proof to verify the inclusion of a blob. 

So we have two options:

- Provide this API with the proof and users can just use it + it's self-contained a can be removed from node anytime in the future
- Or, we delete it and write some specs for the endpoints. Then, whenever someone is building a rollup, we tell them to re-implement that using the specs.

I'll take this discussion to slack and we can get other's opinion.

Thanks for flagging this 👍

Wondertan:
Another future option is for when the node actually becomes pluggable. You will be able to make a module that lives outside the node but can be integrated into the node like its part of it. Basically, there will be enhanced nodes with some additional application-specific functionality depending on module configuration. So, we get seamless integration into the node without needing the node team to maintain all the nonstandard modules.

PR #3472, opened 2024-06-05
test(internal): add testutils package

cristaloleg:
Chances that we will get https://github.com/golang/go/issues/36532 is quite low, however this should make our tests a bit simpler.


cristaloleg:
Chances of https://github.com/golang/go/issues/36532 are high now. Closing.

PR #3475, opened 2024-06-06
feat(rpc): allowing all CORS origins

distractedm1nd:
Needed for API Playground

### DIFF
--- a/api/rpc/server.go
+++ b/api/rpc/server.go
@@ -45,14 +45,14 @@ func NewServer(address, port string, authDisabled bool, secret jwt.Signer) *Serv
 	srv.srv.Handler = &auth.Handler{
 		Verify: srv.verifyAuth,
 		Next: func(w http.ResponseWriter, r *http.Request) {
-			if origin := r.Header.Get("Origin"); origin != "" {
+			if origin := r.Header.Get("Origin"); origin != "" && authDisabled {
 				w.Header().Set("Access-Control-Allow-Origin", origin)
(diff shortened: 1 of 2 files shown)


cristaloleg (approved):
SGTM but I haven't looked deeply enough.

walldiss:
Coverted to draft, to prevent merge before it will get updated

api/rpc/server.go
```diff
@@ -44,7 +44,19 @@ func NewServer(address, port string, authDisabled bool, secret jwt.Signer) *Serv
 	}
 	srv.srv.Handler = &auth.Handler{
 		Verify: srv.verifyAuth,
-		Next:   rpc.ServeHTTP,
+		Next: func(w http.ResponseWriter, r *http.Request) {
+			if origin := r.Header.Get("Origin"); origin != "" {
+				w.Header().Set("Access-Control-Allow-Origin", origin)
```
Wondertan, line 49: Any reason you dont set simply a `*`? AFAIU, it eill have the same effect

Wondertan (commented):
Ideally we would add a config section for the rpc server http header where you would configure the playground node to allow all origins, instead of every node to allow that

walldiss:
After discussing with @distractedm1nd I've put into the issue the approach that will be implemented. Current one is irrelevant
https://github.com/celestiaorg/celestia-node/issues/3481

api/rpc/server.go
distractedm1nd: Doesn't work. Was my first thought too, but it had something to do with the OPTIONS request IIRC. I'll post a link tomorrow when i'm back at the computer

PR #3478, opened 2024-06-07
refactor(share): use atomic.Int64 and fix 'racy' bugs

cristaloleg:
Moving from raw `int64` and atomic operations to the `atomic.Int64` type. As a result fixing a few places where we were accessing atomic variables non-atomically. Which isn't a problem 'cause there were only reads BUT these places would reported by race detector and fail our tests for not a big reason.

### DIFF
--- a/share/ipld/namespace_data.go
+++ b/share/ipld/namespace_data.go
@@ -54,13 +54,15 @@ type NamespaceData struct {
 
 func NewNamespaceData(maxShares int, namespace share.Namespace, options ...Option) *NamespaceData {
 	data := &NamespaceData{
-		// we don't know where in the tree the leaves in the namespace are,
-		// so we keep track of the bounds to return the correct slice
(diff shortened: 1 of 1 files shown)


share/ipld/namespace_data.go
```diff
@@ -54,14 +54,16 @@ type NamespaceData struct {
...
-		bounds:    fetchedBounds{int64(maxShares), 0},
 		maxShares: maxShares,
 		namespace: namespace,
 	}
 
+	// we don't know where in the tree the leaves in the namespace are,
+	// so we keep track of the bounds to return the correct slice
+	// maxShares acts as a sentinel to know if we find any leaves
+	data.bounds.lowest.Store(int64(maxShares))
+	data.bounds.highest.Store(0)
```
cristaloleg, line 65: Just for clarity, by default it's already a zero.

Wondertan (approved):

walldiss (approved):

PR #3479, opened 2024-06-07
feat(docgen): improve examples

distractedm1nd:
Improved examples for blob related types to be consistent throughout (Commitment, Proof, Blob, Namespace). Previous blob.Proof and blob.Commitment were not valid values.

Also added a valid example value for libhead.Hash

### DIFF
--- a/api/docgen/examples.go
+++ b/api/docgen/examples.go
@@ -45,55 +45,11 @@ var exampleTxResponse string
 //go:embed "exampledata/resourceManagerStats.json"
 var exampleResourceMngrStats string
 
-var exampleBlob = `{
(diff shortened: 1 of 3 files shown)


api/docgen/examples.go
```diff
@@ -43,6 +45,56 @@ var exampleTxResponse string
 //go:embed "exampledata/resourceManagerStats.json"
 var exampleResourceMngrStats string
 
+var exampleBlob = `{
```
cristaloleg, line 48: This and big variable below can be 1) const (for clarity) but also 2) go:embed-ed

distractedm1nd: done

vgonkivs (approved):

cristaloleg (approved):
SGTM

ramin (approved):

PR #3482, opened 2024-06-10
feat(shwap): Add ODS file

walldiss:
This PR adds ods file, that implements eds.Accessor interface. The file stores ods part of eds on disk and lazily reads data upon request. If requested data is from Q4, it reads full ods in single read and stores it in-memory for later re-use.

Based on https://github.com/celestiaorg/celestia-node/pull/3425

### DIFF
--- a/share/shwap/row_namespace_data.go
+++ b/share/shwap/row_namespace_data.go
@@ -164,10 +164,14 @@ func (rnd RowNamespaceData) Validate(dah *share.Root, namespace share.Namespace,
 // verifyInclusion checks the inclusion of the row's shares in the provided root using NMT.
 func (rnd RowNamespaceData) verifyInclusion(rowRoot []byte, namespace share.Namespace) bool {
 	leaves := make([][]byte, 0, len(rnd.Shares))
-	for _, shr := range rnd.Shares {
-		namespaceBytes := share.GetNamespace(shr)
-		leaves = append(leaves, append(namespaceBytes, shr...))
+	for _, sh := range rnd.Shares {
+		namespaceBytes := share.GetNamespace(sh)
+		leave := make([]byte, len(sh)+len(namespaceBytes))
+		copy(leave, namespaceBytes)
+		copy(leave[len(namespaceBytes):], sh)
+		leaves = append(leaves, leave)
 	}
+
 	return rnd.Proof.VerifyNamespace(
 		share.NewSHA256Hasher(),
 		namespace.ToNMT(),
--- a/store/file/codec.go
+++ b/store/file/codec.go
@@ -1,0 +1,38 @@
+package file
+
+import (
+	"sync"
+
+	"github.com/klauspost/reedsolomon"
+)
+
+var codec Codec
+
+func init() {
+	codec = NewCodec()
+}
+
+type Codec interface {
+	Encoder(len int) (reedsolomon.Encoder, error)
+}
+
+type codecCache struct {
+	cache sync.Map
+}
+
+func NewCodec() Codec {
+	return &codecCache{}
+}
+
+func (l *codecCache) Encoder(len int) (reedsolomon.Encoder, error) {
+	enc, ok := l.cache.Load(len)
+	if !ok {
+		var err error
+		enc, err = reedsolomon.New(len/2, len/2, reedsolomon.WithLeopardGF(true))
+		if err != nil {
+			return nil, err
+		}
+		l.cache.Store(len, enc)
+	}
+	return enc.(reedsolomon.Encoder), nil
+}
--- a/store/file/codec.go
+++ b/store/file/codec.go
@@ -6,7 +6,6 @@ import (
 	"github.com/klauspost/reedsolomon"
 )
 
-// TODO: codec will be removed after support for reconstructSome is added to rsmt2d.
 var codec Codec
 
 func init() {
--- a/store/file/codec.go
+++ b/store/file/codec.go
@@ -6,6 +6,7 @@ import (
 	"github.com/klauspost/reedsolomon"
 )
 
+// TODO: codec will be removed after support for reconstructSome is added to rsmt2d.
 var codec Codec
 
 func init() {
--- a/store/file/codec_test.go
+++ b/store/file/codec_test.go
@@ -1,0 +1,83 @@
+package file
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/klauspost/reedsolomon"
+	"github.com/stretchr/testify/require"
+
+	"github.com/celestiaorg/celestia-node/share/sharetest"
+)
+
+func BenchmarkCodec(b *testing.B) {
+	minSize, maxSize := 32, 128
+
+	for size := minSize; size <= maxSize; size *= 2 {
+		// BenchmarkCodec/Leopard/size:32-10         					  409194	      2793 ns/op
+		// BenchmarkCodec/Leopard/size:64-10                         	  190969	      6170 ns/op
+		// BenchmarkCodec/Leopard/size:128-10                        	   82821	     14287 ns/op
+		b.Run(fmt.Sprintf("Leopard/size:%v", size), func(b *testing.B) {
+			enc, err := reedsolomon.New(size/2, size/2, reedsolomon.WithLeopardGF(true))
+			require.NoError(b, err)
+
+			shards := newShards(b, size, true)
+
+			b.ResetTimer()
+			for i := 0; i < b.N; i++ {
+				err = enc.Encode(shards)
+				require.NoError(b, err)
+			}
+		})
+
+		// BenchmarkCodec/default/size:32-10       					  	  222153	      5364 ns/op
+		// BenchmarkCodec/default/size:64-10                         	   58831	     20349 ns/op
+		// BenchmarkCodec/default/size:128-10                        	   14940	     80471 ns/op
+		b.Run(fmt.Sprintf("default/size:%v", size), func(b *testing.B) {
+			enc, err := reedsolomon.New(size/2, size/2, reedsolomon.WithLeopardGF(false))
+			require.NoError(b, err)
+
+			shards := newShards(b, size, true)
+
+			b.ResetTimer()
+			for i := 0; i < b.N; i++ {
+				err = enc.Encode(shards)
+				require.NoError(b, err)
+			}
+		})
+
+		// BenchmarkCodec/default-reconstructSome/size:32-10         	 1263585	       954.4 ns/op
+		// BenchmarkCodec/default-reconstructSome/size:64-10         	  762273	      1554 ns/op
+		// BenchmarkCodec/default-reconstructSome/size:128-10        	  429268	      2974 ns/op
+		b.Run(fmt.Sprintf("default-reconstructSome/size:%v", size), func(b *testing.B) {
+			enc, err := reedsolomon.New(size/2, size/2, reedsolomon.WithLeopardGF(false))
+			require.NoError(b, err)
+
+			shards := newShards(b, size, false)
+			targets := make([]bool, size)
+			target := size - 2
+			targets[target] = true
+
+			b.ResetTimer()
+			for i := 0; i < b.N; i++ {
+				err = enc.ReconstructSome(shards, targets)
+				require.NoError(b, err)
+				shards[target] = nil
+			}
+		})
+	}
+}
+
+func newShards(b require.TestingT, size int, fillParity bool) [][]byte {
+	shards := make([][]byte, size)
+	original := sharetest.RandShares(b, size/2)
+	copy(shards, original)
+
+	if fillParity {
+		// fill with parity empty Shares
+		for j := len(original); j < len(shards); j++ {
+			shards[j] = make([]byte, len(original[0]))
+		}
+	}
+	return shards
+}
--- a/store/file/header.go
+++ b/store/file/header.go
@@ -46,13 +46,14 @@ func readHeader(r io.Reader) (*headerV0, error) {
 		return nil, fmt.Errorf("readHeader: %w", err)
 	}
 
-	switch version { //nolint:gocritic // gocritic wants to convert it to if statement.
+	switch version {
 	case headerVersionV0:
 		h := &headerV0{}
 		_, err := h.ReadFrom(r)
 		return h, err
+	default:
+		return nil, fmt.Errorf("unsupported header fileVersion: %d", version)
 	}
-	return nil, fmt.Errorf("unsupported header fileVersion: %d", version)
 }
 
 func writeHeader(w io.Writer, h *headerV0) error {
@@ -68,6 +69,7 @@ func writeHeader(w io.Writer, h *headerV0) error {
 }
 
 func (h *headerV0) Size() int {
+	// header size + 1 byte for header fileVersion
 	return headerVOSize + 1
 }
 
@@ -86,7 +88,7 @@ func (h *headerV0) ReadFrom(r io.Reader) (int64, error) {
 	bytesHeader := make([]byte, headerVOSize)
 	n, err := io.ReadFull(r, bytesHeader)
 	if n != headerVOSize {
-		return 0, fmt.Errorf("readHeader: read %d bytes, expected %d", len(bytesHeader), headerVOSize)
+		return 0, fmt.Errorf("headerV0 ReadFrom: read %d bytes, expected %d", len(bytesHeader), headerVOSize)
 	}
 
 	h.fileVersion = fileVersion(bytesHeader[0])
--- a/store/file/header.go
+++ b/store/file/header.go
@@ -1,0 +1,94 @@
+package file
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"io"
+
+	"github.com/celestiaorg/celestia-node/share"
+)
+
+const headerSize = 64
+
+type header struct {
+	version  fileVersion
+	fileType fileType
+
+	// Taken directly from EDS
+	shareSize  uint16
+	squareSize uint16
+
+	datahash share.DataHash
+}
+
+type fileVersion uint8
+
+const (
+	fileV0 fileVersion = iota
+)
+
+type fileType uint8
+
+const (
+	ods fileType = iota
+	q1q4
+)
+
+func (h *header) WriteTo(w io.Writer) (int64, error) {
+	b := bytes.NewBuffer(make([]byte, 0, headerSize))
+	_ = b.WriteByte(byte(h.version))
+	_ = b.WriteByte(byte(h.fileType))
+	_ = binary.Write(b, binary.LittleEndian, h.shareSize)
+	_ = binary.Write(b, binary.LittleEndian, h.squareSize)
+	_, _ = b.Write(h.datahash)
+	// write padding
+	_, _ = b.Write(make([]byte, headerSize-b.Len()-1))
+	return writeLenEncoded(w, b.Bytes())
+}
+
+func readHeader(r io.Reader) (*header, error) {
+	bytesHeader, err := readLenEncoded(r)
+	if err != nil {
+		return nil, err
+	}
+	if len(bytesHeader) != headerSize-1 {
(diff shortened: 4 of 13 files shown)


walldiss:
Lint will be get fixed after linter upgrade https://github.com/celestiaorg/celestia-node/pull/3483

store/file/codec_test.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/klauspost/reedsolomon"
+	"github.com/stretchr/testify/require"
+
+	"github.com/celestiaorg/celestia-node/share/sharetest"
+)
+
+func BenchmarkCodec(b *testing.B) {
+	minSize, maxSize := 32, 128
+
+	for size := minSize; size <= maxSize; size *= 2 {
```
cristaloleg, line 16: IMO feels like this should be inside every (sub-) benchmark to make them isolated and more visible relation to `size` variable.

cristaloleg: Ah...we use it in benchmark name...so (probably) ignore my comment.

store/file/header.go
```diff
@@ -0,0 +1,94 @@
...
+		squareSize: binary.LittleEndian.Uint16(bytesHeader[4:6]),
+		datahash:   make([]byte, 32),
+	}
+
+	copy(h.datahash, bytesHeader[6:6+32])
+	return h, err
+}
+
+func writeLenEncoded(w io.Writer, data []byte) (int64, error) {
+	_, err := w.Write([]byte{byte(len(data))})
```
cristaloleg, line 71: Waaaait, are we writing just the lower 8bits of `len(data)` here?

store/file/square.go
```diff
@@ -0,0 +1,132 @@
...
+	axisType rsmt2d.Axis,
+	axisIdx int,
+) ([]share.Share, error) {
+	shares := make([]share.Share, s.size())
+
+	// extend opposite half of the square while collecting Shares for the first half of required axis
+	g, ctx := errgroup.WithContext(ctx)
+	opposite := oppositeAxis(axisType)
+	for i := 0; i < s.size(); i++ {
+		i := i
```
cristaloleg, line 95: Starting with Go 1.22 we don't need that https://go.dev/blog/loopvar-preview

store/file/ods_test.go
```diff
@@ -0,0 +1,146 @@
...
+// BenchmarkShareFromOdsFile/Size:128/Axis:col/squareHalf:first(original)-10        	    2114	    514642 ns/op
+// BenchmarkShareFromOdsFile/Size:128/Axis:col/squareHalf:second(extended)-10       	     373	   3068104 ns/op
+func BenchmarkShareFromOdsFile(b *testing.B) {
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	b.Cleanup(cancel)
+
+	minSize, maxSize := 32, 128
+	dir := b.TempDir()
+
+	newFile := func(size int) eds.Accessor {
```
cristaloleg, line 137: Looks like it's the same as above. Do we need to keep this as a param or just leave on eds testing funcs?

store/file/header.go
cristaloleg: Ok, looks like we always(?) write 64 bytes (`headerSize`), am I right?

store/file/header.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		return 0, err
+	}
+	return io.Copy(w, bytes.NewBuffer(data))
```
cristaloleg, line 75: `w.Write(data)` ?

store/file/ods.go
```diff
@@ -0,0 +1,269 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("writing ODS file: %w", err)
+	}
+
+	// TODO: fill ods field with data from eds
+	return &OdsFile{
+		path: path,
+		fl:   f,
+		hdr:  h,
+	}, f.Sync()
```
cristaloleg, line 76: I read this file 3 times and only on the last iteration I noticed this `f.Sync`, what about to move it right after `if` above?

store/file/header.go
walldiss: It writes len of data into first byte. Right now we read 64 bytes always no matter what. In future header struct might grow and new fields might be added. It would require old storages to be upgraded to bigger header or lenEncoded version of it. So lenEncoding is added to prevent future break of full file. Otherwise we would need to rewrite all files fully on file.header upgrade. 
TLDR: It doesn't do much as of now, but will be needed if header struct grow to keep backwards compatibility.

store/file/header.go
walldiss: I don't want to manage amount bytes written manually and prefer io.Copy to do the job

cristaloleg: In my example you don't manage how many bytes you're writing. If it will fail - it will.

store/file/header.go
cristaloleg: Sounds like a good place for a) write `int32` which will be enough for any sane version (also, it's header, I bet it will be 1Gb) or b) specify header version with byte counter.

Anyway, both of such solutions will make simpler our migration in the future (even if it will happen).

store/file/header.go
walldiss: For example, if you write 100 bytes with w.Write(data), it can write first 10 bytes out of 100 and return `10, nil`. While it is not an error, you need yo manually write the rest to the writer. I don't want to do it manually.

cristaloleg: As per `io.Writer` documentation (which sadly doesn't give any guarantees on the interface implementation, oof) it should work as: `It returns the number of bytes written from p (0 <= n <= len(p)) and any error encountered that caused the write to stop early.`

So, if it writer less than `len(p)` we should have an error otherwise everything should be written fully.

I don't see a good argument to continue this thread now, let's leave it as `io.Copy + bytes.Buffer`.

quasilyte: If one doesn't trust `Writer` inteface to return a non-nil error when n is less than `len(data)`, why one trusts `WriterTo` interface to be any better?

`io.Copy` will recognize that `bytes.Buffer` implements `io.WriterTo` method and forward the call directly:
https://cs.opensource.google/go/go/+/refs/tags/go1.22.4:src/io/io.go;l=411
https://cs.opensource.google/go/go/+/refs/tags/go1.22.4:src/bytes/buffer.go;l=258

Here is the WriterTo documentation:
https://pkg.go.dev/io#WriterTo
It doesn't give any contracts while io.Writer docs require this:
> Write must return a non-nil error if it returns n < len(p). Write must not modify the slice data, even temporarily. 

I would at least give it a second thought. The code right now is more fragile (it relies on something that is not guaranteed while avoiding something with a specified contract) plus it does more job (creates a new bytes buffer, for example) than necessary.

walldiss: `WriteTo` implementation of bytes.Buffer does give guarantee of full write, and in case when not full data has been written it will return io.ErrShortWrite. Nonetheless I agree, that in discussed snippet ordinary `Write` should be sufficient and will be simplier. I'll will update code accordingly

store/file/ods_test.go
```diff
@@ -0,0 +1,133 @@
+//nolint:goconst
```
walldiss, line 1: This will be ineffective after upgrade to 1.59 golangci-lint and will be removed

store/file/header.go
walldiss: I've removed length encoding in favour of headerVersion

Wondertan (commented):
First look, still have to go over the logic in square and the file

share/shwap/row_namespace_data.go
```diff
@@ -164,10 +164,14 @@ func (rnd RowNamespaceData) Validate(dah *share.Root, namespace share.Namespace,
...
 	leaves := make([][]byte, 0, len(rnd.Shares))
-	for _, shr := range rnd.Shares {
-		namespaceBytes := share.GetNamespace(shr)
-		leaves = append(leaves, append(namespaceBytes, shr...))
+	for _, sh := range rnd.Shares {
+		namespaceBytes := share.GetNamespace(sh)
+		leave := make([]byte, len(sh)+len(namespaceBytes))
+		copy(leave, namespaceBytes)
+		copy(leave[len(namespaceBytes):], sh)
+		leaves = append(leaves, leave)
```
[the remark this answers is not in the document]
walldiss: I like that you noticed. There was a very nasty mutations bug that took me few hours to fix.

store/file/codec.go
```diff
@@ -0,0 +1,38 @@
+package file
+
+import (
+	"sync"
+
+	"github.com/klauspost/reedsolomon"
+)
+
+var codec Codec
```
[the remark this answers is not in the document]
walldiss: Access to ReconstructSome

store/file/header.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+type headerVersion uint8
+
+const (
+	headerVersionV0 headerVersion = 1
+	headerVOSize    int           = 40
+)
+
+type headerV0 struct {
```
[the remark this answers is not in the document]
walldiss: tye field is fileType, not a header version

store/file/square.go
```diff
@@ -0,0 +1,131 @@
...
+
+	square := make(square, odsLn)
+	for i := range square {
+		square[i] = make([]share.Share, odsLn)
+		for j := range square[i] {
+			square[i][j] = make(share.Share, shareSize)
+		}
+	}
+
+	// TODO(@walldiss): run benchmark to find optimal size for this buffer
```
Wondertan, line 32: I would park this in the issue, so that some rando interested in low-level things could contribute

Wondertan: It might not be an issue to be created now, but we could make a little list of low-prior issues to be posted to GH. This one for example and the Size returning error

store/file/square.go
```diff
@@ -0,0 +1,131 @@
...
+	"io"
+
+	"golang.org/x/sync/errgroup"
+
+	"github.com/celestiaorg/rsmt2d"
+
+	"github.com/celestiaorg/celestia-node/share"
+)
+
+type square [][]share.Share
```
Wondertan, line 16: Is this PR used only for ODS? Should we call it so, then? Or will it be used for EDS as well?

Wondertan: Or at least the `readSquare` func because its hardcoded to read the ODS, but square mean both ODS and EDS. Need to clarify

store/file/square.go
```diff
@@ -0,0 +1,131 @@
...
+func readSquare(r io.Reader, shareSize, edsSize int) (square, error) {
+	odsLn := edsSize / 2
+
+	square := make(square, odsLn)
+	for i := range square {
+		square[i] = make([]share.Share, odsLn)
+		for j := range square[i] {
+			square[i][j] = make(share.Share, shareSize)
+		}
+	}
```
Wondertan, line 30: Don't we want to read the square by Row number of reads instead of Row*Col reads? Not sure why you preferred reading share-by-share instead of row-by-row

store/file/square.go
```diff
@@ -0,0 +1,131 @@
...
+
+func (s square) shares() ([]share.Share, error) {
+	shares := make([]share.Share, 0, s.size()*s.size())
+	for _, row := range s {
+		shares = append(shares, row...)
+	}
+	return shares, nil
+}
+
+func (s square) computeAxisHalf(
```
Wondertan, line 84: Isn't this a full axis?

Wondertan: Ohh, compute the opposite half, basically parity, because the square stores the ODS. This needs to be clarified in the comment or by a better function name, I believe.

store/file/square.go
```diff
@@ -0,0 +1,131 @@
...
+
+			target := make([]bool, s.size()*2)
+			target[axisIdx] = true
+
+			err = enc.ReconstructSome(shards, target)
+			if err != nil {
+				return fmt.Errorf("reconstruct some: %w", err)
+			}
+
+			shares[i] = shards[axisIdx]
```
Wondertan, line 117: Shouldn't we cache this on the square? Is this what the cache file is for?

store/file/square.go
```diff
@@ -0,0 +1,131 @@
...
+				return err
+			}
+
+			enc, err := codec.Encoder(s.size() * 2)
+			if err != nil {
+				return fmt.Errorf("encoder: %w", err)
+			}
+
+			shards := make([][]byte, s.size()*2)
+			copy(shards, original)
```
Wondertan, line 107: Why is copy necessary?

store/file/square.go
```diff
@@ -0,0 +1,131 @@
...
+
+			target := make([]bool, s.size()*2)
+			target[axisIdx] = true
+
+			err = enc.ReconstructSome(shards, target)
+			if err != nil {
+				return fmt.Errorf("reconstruct some: %w", err)
+			}
```
Wondertan, line 115: This actually doesn't need a reconstructSome, because it always does the Reconstruct operation as it recovers the full half

Wondertan: So can we avoid spending time on Some?

store/file/square.go
```diff
@@ -0,0 +1,131 @@
...
+	for i := 0; i < s.size(); i++ {
+		g.Go(func() error {
+			original, err := s.axisHalf(ctx, opposite, i)
+			if err != nil {
+				return err
+			}
+
+			enc, err := codec.Encoder(s.size() * 2)
+			if err != nil {
+				return fmt.Errorf("encoder: %w", err)
```
Wondertan, line 103:
```suggestion
				return fmt.Errorf("getting encoder: %w", err)
```

store/file/ods.go
```diff
@@ -0,0 +1,280 @@
...
+		return eds.AxisHalf{}, err
+	}
+
+	shares, err := f.ods.computeAxisHalf(ctx, axisType, axisIdx)
+	if err != nil {
+		return eds.AxisHalf{}, fmt.Errorf("computing axis half: %w", err)
+	}
+	return eds.AxisHalf{
+		Shares:   shares,
+		IsParity: false,
```
Wondertan, line 170: This should be true

store/file/ods.go
```diff
@@ -0,0 +1,280 @@
...
+	if ods != nil {
+		return f.ods.axisHalf(context.Background(), axisType, axisIdx)
+	}
+
+	switch axisType {
+	case rsmt2d.Col:
+		return f.readCol(axisIdx, 0)
+	case rsmt2d.Row:
+		return f.readRow(axisIdx)
+	}
```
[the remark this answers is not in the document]
Wondertan: Ah, I guess this is gonna be responsibility of the cache file

store/file/ods.go
```diff
@@ -0,0 +1,280 @@
...
+
+	shares, err := f.ods.computeAxisHalf(ctx, axisType, axisIdx)
+	if err != nil {
+		return eds.AxisHalf{}, fmt.Errorf("computing axis half: %w", err)
+	}
+	return eds.AxisHalf{
+		Shares:   shares,
+		IsParity: false,
+	}, nil
```
Wondertan, line 171: I am a bit concerned that this gonna recompute halfs twice:
* computeAxisHalf: square -> to parity half
* AxisHalf: parity half -> original
  * When original is already in the square

store/file/ods.go
```diff
@@ -0,0 +1,280 @@
...
+		hdr:  h,
+		fl:   f,
+	}, nil
+}
+
+// CreateOdsFile creates a new file. File has to be closed after usage.
+func CreateOdsFile(
+	path string,
+	datahash share.DataHash,
+	eds *rsmt2d.ExtendedDataSquare,
```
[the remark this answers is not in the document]
Wondertan: In sync: we don't need that right now, but we may return to this later if needed, e.g. storage operating over Accessors isntead of the concretion

store/file/ods.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+
+	return shwap.SampleFromShares(axis, axisType, axisIdx, shrIdx)
+}
+
+// AxisHalf returns half of shares axis of the given type and index. Side is determined by
+// implementation. Implementations should indicate the side in the returned AxisHalf.
+func (f *OdsFile) AxisHalf(ctx context.Context, axisType rsmt2d.Axis, axisIdx int) (eds.AxisHalf, error) {
+	// read axis from file if axisis row and from top half of the square or if axis is column and from
+	// left half of the square
```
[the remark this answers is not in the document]
Wondertan: nvm there is column case in the comment later as well

store/file/square.go
Wondertan: Sync: call it a `quadrant`

store/file/square.go
Wondertan: sync: comment with a picture

Wondertan: and generally more comments

store/file/square.go
Wondertan: Yes

store/file/square.go
Wondertan: Because we need to allocate the full row here for codec's API?

Wondertan: Needs more comments though

store/file/square.go
Wondertan: My comment was wrong. Some is needed to recompute particular opposite shares, so we better off keeping it

store/file/ods.go
Wondertan: That is not true. I got confused because objectively, this is parity, but relative to the axis in Q2, this is original data.

store/file/ods.go
Wondertan: This comment is also wrong and based on a misunderstanding of computeAxisHalf works

store/file/square.go
```diff
@@ -0,0 +1,132 @@
...
+	shares := make([]share.Share, 0, s.size()*s.size())
+	for _, row := range s {
+		shares = append(shares, row...)
+	}
+	return shares, nil
+}
+
+// TODO(@walldiss): make comments with diagram of computed axis. Add more comment on actual algo
+func (s square) computeAxisHalf(
+	ctx context.Context,
```
Wondertan, line 86: We don't actually need context here

store/file/square.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+	}
+	return square, nil
+}
+
+func (s square) size() int {
+	return len(s)
+}
+
+func (s square) axisHalf(_ context.Context, axisType rsmt2d.Axis, axisIdx int) ([]share.Share, error) {
```
Wondertan, line 54: Use AxisHalf type instead

store/file/square.go
```diff
@@ -0,0 +1,132 @@
...
+
+	// construct half column from row ordered square
+	col := make([]share.Share, s.size())
+	for i := 0; i < s.size(); i++ {
+		col[i] = s[i][axisIdx]
+	}
+	return col, nil
+}
+
+func (s square) shares() ([]share.Share, error) {
```
Wondertan, line 76: Put above axisHalf

store/file/ods.go
```diff
@@ -193,21 +187,29 @@ func (f *ODSFile) Shares(context.Context) ([]share.Share, error) {
...
 	case rsmt2d.Row:
-		return f.readRow(axisIdx)
+		row, err := f.readRow(axisIdx)
+		return eds.AxisHalf{
+			Shares:   row,
+			IsParity: false,
+		}, err
 	}
-	return nil, fmt.Errorf("unknown axis")
+	return eds.AxisHalf{}, fmt.Errorf("unknown axis")
```
Wondertan: uber optional nit: dedup

store/file/ods.go
```diff
@@ -233,7 +235,7 @@ func (f *ODSFile) readODS() error {
 
 func (f *ODSFile) readRow(idx int) ([]share.Share, error) {
 	shrLn := int(f.hdr.shareSize)
-	ODSLn := int(f.size()) / 2
+	ODSLn := f.size() / 2
```
Wondertan, line 238: The vars here aren't supposed to be capitalized

cristaloleg (approved):

PR #3484, opened 2024-06-10
feat(shwap): Add close once file

walldiss:
Adds middleware to protect AccessorCloser from multiple Closes.

Based on https://github.com/celestiaorg/celestia-node/pull/3425

### DIFF
--- a/share/new_eds/close_once.go
+++ b/share/new_eds/close_once.go
@@ -1,4 +1,4 @@
-package file
+package eds
 
 import (
 	"context"
@@ -8,20 +8,19 @@ import (
 	"github.com/celestiaorg/rsmt2d"
 
 	"github.com/celestiaorg/celestia-node/share"
-	eds "github.com/celestiaorg/celestia-node/share/new_eds"
 	"github.com/celestiaorg/celestia-node/share/shwap"
 )
 
-var _ eds.AccessorCloser = (*closeOnce)(nil)
+var _ AccessorCloser = (*closeOnce)(nil)
 
 var errAccessorClosed = errors.New("accessor is closed")
 
 type closeOnce struct {
-	f      eds.AccessorCloser
+	f      AccessorCloser
 	closed atomic.Bool
 }
 
-func WithClosedOnce(f eds.AccessorCloser) eds.AccessorCloser {
+func WithClosedOnce(f AccessorCloser) AccessorCloser {
 	return &closeOnce{f: f}
 }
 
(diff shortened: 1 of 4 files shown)


store/file/close_once.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/celestiaorg/celestia-node/share/shwap"
+)
+
+var _ eds.AccessorCloser = (*closeOnce)(nil)
+
+var errAccessorClosed = errors.New("accessor is closed")
+
+type closeOnce struct {
+	f      eds.AccessorCloser
+	closed atomic.Bool
```
cristaloleg, line 21: We've amazing https://pkg.go.dev/sync#OnceFunc for that

Wondertan:
Needs rebase

store/file/close_once.go
walldiss: closeOnce is doing something more here. The motivation behind closeOnce is to prevent call of any methods on inner Accessor if it was closed.

walldiss:
>Why ValidationAccessor would need to manage state of underlying accessor being closed. Doesnt seem like that should be it responsibility

ValidationAccessor is different PR

Wondertan:
It wasnt though

store/file/close_once.go
```diff
@@ -0,0 +1,79 @@
+package file
```
Wondertan, line 1: Should we move it closer to interface definition as a helper implementation?

walldiss: That is something I wanted to discuss. closeOnce and validation are generic middleware around Accessor, so naturally they better fit into `eds` pkg, where Accessor is defined. However those seems to be useful only for file implementations and serves to solve files related problems. I also envision `eds` pkg as something that we want to extract into common repo/lib and those middleware are not necessarily needs to be placed into that repo/lib.

share/new_eds/close_once.go
```diff
@@ -0,0 +1,78 @@
...
+	if !c.closed.Swap(true) {
+		err := c.f.Close()
+		// release reference to the accessor to allow GC to collect all resources associated with it
+		c.f = nil
+		return err
+	}
+	return nil
+}
+
+func (c *closeOnce) Size(ctx context.Context) int {
```
Wondertan, line 37: Random realization that we could fix in some other PR: if it takes context, it should return an error

share/new_eds/close_once.go
```diff
@@ -0,0 +1,78 @@
...
+	return &closeOnce{f: f}
+}
+
+func (c *closeOnce) Close() error {
+	if !c.closed.Swap(true) {
+		err := c.f.Close()
+		// release reference to the accessor to allow GC to collect all resources associated with it
+		c.f = nil
+		return err
+	}
```
Wondertan, line 33: Optional nit: avoid nesting

walldiss: done

PR #3498, opened 2024-06-17
feat(shwap): Add cache file

walldiss:
Add eds.Accessor to cache extended shares and proofs.


Wondertan:
Needs rebase

store/file/cache_file.go
```diff
@@ -0,0 +1,252 @@
...
+		axisCache: []map[int]inMemoryAxis{make(map[int]inMemoryAxis), make(map[int]inMemoryAxis)},
+	}
+}
+
+func (f *proofsCacheFile) Size(ctx context.Context) int {
+	f.lock.Lock()
+	defer f.lock.Unlock()
+	if f.size == 0 {
+		f.size = f.Accessor.Size(ctx)
+	}
```
Wondertan, line 63: Use read lock for size check and write lock only on write

store/file/cache_file.go
```diff
@@ -0,0 +1,252 @@
...
+	"github.com/celestiaorg/celestia-node/store/file/proofs"
+)
+
+var _ eds.Accessor = (*proofsCacheFile)(nil)
+
+// proofsCacheFile is eds accessor that caches proofs for rows and columns. It also caches extended axis Shares.
+// It is used to speed up the process of building proofs for rows and columns, reducing the number of reads from the
+// underlying file.
+type proofsCacheFile struct {
+	eds.Accessor
```
Wondertan, line 30: Sync: embedding here is a foot gun

store/file/proofs/proofs.go
```diff
@@ -0,0 +1,36 @@
...
+
+	"github.com/ipfs/boxo/blockservice"
+	"github.com/ipfs/go-cid"
+
+	"github.com/celestiaorg/nmt"
+
+	"github.com/celestiaorg/celestia-node/share/ipld"
+)
+
+func GetProofs(
```
Wondertan, line 15: sync: move to IPLD

store/file/cache_file.go
```diff
@@ -0,0 +1,252 @@
...
+	"github.com/celestiaorg/celestia-node/share/shwap"
+	"github.com/celestiaorg/celestia-node/store/file/proofs"
+)
+
+var _ eds.Accessor = (*proofsCacheFile)(nil)
+
+// proofsCacheFile is eds accessor that caches proofs for rows and columns. It also caches extended axis Shares.
+// It is used to speed up the process of building proofs for rows and columns, reducing the number of reads from the
+// underlying file.
+type proofsCacheFile struct {
```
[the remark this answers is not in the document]
Wondertan: Also, move closer to Interface definition

share/ipld/proofs.go
```diff
@@ -0,0 +1,74 @@
...
+	"context"
+	"math"
+
+	"github.com/ipfs/boxo/blockservice"
+	"github.com/ipfs/go-cid"
+
+	"github.com/celestiaorg/nmt"
+)
+
+func GetProof(
```
Wondertan, line 13: Little godoc, might copy the one from the `getProof`.

share/new_eds/proofs_cache.go
```diff
@@ -0,0 +1,276 @@
...
+// axis Shares. It is used to speed up the process of building proofs for rows and columns,
+// reducing the number of reads from the underlying accessor.
+type proofsCache struct {
+	inner Accessor
+
+	// lock protects axisCache
+	lock sync.RWMutex
+	// axisCache caches the axis Shares and proofs. Index in the slice corresponds to the axis type.
+	// The map key is the index of the axis.
+	axisCache []map[int]axisWithProofs
```
Wondertan, line 35: Is there a reason to make it a map with an int key instead of a slice? We could speed up reads by using a slice there.

In the future, if that becomes a contention point, we could make a lock per axis instead of the whole cache.

Wondertan (commented):
Much better documentation!

Wondertan:
And needs rebase

share/new_eds/proofs_cache.go
walldiss: Few minor advantages for map:
- it allows nice (val, ok) check. Otherwise we would need to make cache an array of pointers to check for nil. `[][]*axisWithProofs`
- it grows gradually with cache size, so there is no need to allocate array for all axis right away.

Wondertan: > Otherwise we would need to make cache an array of pointers to check for nil. 

That's fine, that would also allow gradual growth of the cache

walldiss: Doubt mutex it can become contention point, it locks on writing to map and there are at most 256 writes in lifespan of cached accessor.

PR #3522, opened 2024-06-24
feat(shwap): add Q1Q4 file

walldiss:
Q1Q4 file stores 2 quadrants and allow more efficient reads of Q4 samples compared to ODS file

### DIFF
--- a/store/file/ods.go
+++ b/store/file/ods.go
@@ -29,7 +29,8 @@ type ODSFile struct {
 	// repeated file reads. - Serving full ODS data by Shares().
 	// Storing the square in memory allows for efficient single-read operations, avoiding the need for
 	// piecemeal reads by rows or columns, and facilitates quick access to data for these operations.
-	ods square
+	ods          square
+	disableCache bool
 }
 
 // OpenODSFile opens an existing file. File has to be closed after usage.
@@ -153,12 +154,12 @@ func (f *ODSFile) AxisHalf(_ context.Context, axisType rsmt2d.Axis, axisIdx int)
 	}
 
 	// if axis is from the second half of the square, read full ODS and compute the axis half
-	err := f.readODS()
+	ods, err := f.readODS()
 	if err != nil {
 		return eds.AxisHalf{}, err
 	}
 
-	half, err := f.ods.computeAxisHalf(axisType, axisIdx)
+	half, err := ods.computeAxisHalf(axisType, axisIdx)
 	if err != nil {
 		return eds.AxisHalf{}, fmt.Errorf("computing axis half: %w", err)
 	}
@@ -180,11 +181,11 @@ func (f *ODSFile) RowNamespaceData(
 
 // Shares returns data shares extracted from the Accessor.
 func (f *ODSFile) Shares(context.Context) ([]share.Share, error) {
-	err := f.readODS()
+	ods, err := f.readODS()
 	if err != nil {
 		return nil, err
 	}
-	return f.ods.shares()
+	return ods.shares()
 }
 
 func (f *ODSFile) readAxisHalf(axisType rsmt2d.Axis, axisIdx int) (eds.AxisHalf, error) {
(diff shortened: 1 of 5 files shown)


store/file/ods.go
```diff
@@ -29,7 +29,8 @@ type ODSFile struct {
 	// repeated file reads. - Serving full ODS data by Shares().
 	// Storing the square in memory allows for efficient single-read operations, avoiding the need for
 	// piecemeal reads by rows or columns, and facilitates quick access to data for these operations.
-	ods square
+	ods          square
```
Wondertan, line 32: rename to quadrant

store/file/q1q4_file.go
```diff
@@ -0,0 +1,135 @@
...
+	half, err := f.AxisHalf(ctx, rsmt2d.Row, rowIdx)
+	if err != nil {
+		return shwap.Sample{}, fmt.Errorf("reading axis: %w", err)
+	}
+	shares, err := half.Extended()
+	if err != nil {
+		return shwap.Sample{}, fmt.Errorf("extending shares: %w", err)
+	}
+	return shwap.SampleFromShares(shares, rsmt2d.Row, rowIdx, colIdx)
+}
```
Wondertan, line 65: My understanding is that Sample won't ever be called if it is wrapped with CacheAccessor. The same is true for RowNamespaceData. This is because cache only calls AxisHalf of the underlying Accessor. Correct?

store/file/q1q4_file.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/celestiaorg/rsmt2d"
+
+	"github.com/celestiaorg/celestia-node/share"
+	eds "github.com/celestiaorg/celestia-node/share/new_eds"
+	"github.com/celestiaorg/celestia-node/share/shwap"
+)
+
+var _ eds.AccessorCloser = (*Q1Q4File)(nil)
+
+type Q1Q4File struct {
```
Wondertan, line 18: Primitive description would not hurt

store/file/ods.go
```diff
@@ -212,38 +213,43 @@ func (f *ODSFile) readAxisHalf(axisType rsmt2d.Axis, axisIdx int) (eds.AxisHalf,
...
 	square, err := readSquare(f.fl, share.Size, f.size())
 	if err != nil {
-		return fmt.Errorf("reading ODS: %w", err)
+		return nil, fmt.Errorf("reading ODS: %w", err)
 	}
-	f.ods = square
-	return nil
+
+	if !f.disableCache {
+		fmt.Println("STORE ODS")
```
Wondertan, line 235: relic

store/file/ods.go
```diff
@@ -29,7 +29,8 @@ type ODSFile struct {
 	// repeated file reads. - Serving full ODS data by Shares().
 	// Storing the square in memory allows for efficient single-read operations, avoiding the need for
 	// piecemeal reads by rows or columns, and facilitates quick access to data for these operations.
-	ods square
+	ods          square
+	disableCache bool
```
Wondertan, line 33: comment to say only in tests.

store/file/q1q4_file.go
```diff
@@ -0,0 +1,135 @@
...
+			_, err := w.Write(eds.GetCell(uint(x), uint(y)))
+			if err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func (f *Q1Q4File) readSharesFromQ4(axisType rsmt2d.Axis, axisIdx int) ([]share.Share, error) {
```
Wondertan, line 122: readAxisFromQ4

Wondertan (commented):
Pretty straightforward

store/file/q1q4_file.go
walldiss: Yes, but cache is decoupled, so Q1Q4 file works in the most optimal way in assumption of no cache. It also helps to have benchmarks of performance of cache-less Q1Q4 file.

Wondertan (approved):

store/file/q1q4_file.go
```diff
@@ -0,0 +1,146 @@
...
+			}
+		}
+	}
+	return nil
+}
+
+func (f *Q1Q4File) readAxisHalfFromQ4(axisType rsmt2d.Axis, axisIdx int) (eds.AxisHalf, error) {
+	q4idx := axisIdx - f.ods.size()/2
+	if q4idx < 0 {
+		return eds.AxisHalf{}, errors.New("invalid index requested from Q4")
```
cristaloleg, line 122: Should we report the index?

walldiss: Added index to the error

PR #3526, opened 2024-06-25
fix(nodebuilder/core): non default values from config.toml are written over by flag defaults in specific circumstances

### LINKED ISSUE #3523: nodebuilder: Config changes not seen by node.Start
@tac0turtle reported that changes to the config were not picked up by the node when started.

He changed the RPC and gRPC ports in the config and started the node, but the node still looked at the default ports. No port flags were passed via CLI

CLI flags should override config on start, but in their absence, config should be read.

This isn't confirmed to be an issue yet, so let's confirm it's reproducible first.

ramin:
identified by @tac0turtle (TYBG) whereby, in the specific case where you had edited you node store's config values to be non-default for `Core.RPCPort` or `Core.GRPCPort`, if you were to then provide a different value for Core.IP via the `--core.ip` flag at node start time, then the flag defaults for `core.rpc.port` and `core.grpc.port` would revert the values to defaults and override the custom values in config.toml. Fix is to check that the default value had been reset in the flag, and only use the default if no value had been set in the passed in config, which in most cases is already loaded as default or the config.toml values.

Took liberty to add some tests, and update some duplication of default values while in here.

fixes #3523

### DIFF
--- a/nodebuilder/core/flags.go
+++ b/nodebuilder/core/flags.go
@@ -50,14 +50,11 @@ func ParseFlags(
 		return nil
 	}
 
-	// in the case that the RPC port is not set, use the default port
-	// or in the case a value has been set via flag,
-	// set that as highest precedence
-	if cfg.RPCPort == "" || cmd.Flag(coreRPCFlag).Changed {
+	if cmd.Flag(coreRPCFlag).Changed {
 		rpc := cmd.Flag(coreRPCFlag).Value.String()
 		cfg.RPCPort = rpc
 	}
-	if cfg.GRPCPort == "" || cmd.Flag(coreGRPCFlag).Changed {
+	if cmd.Flag(coreGRPCFlag).Changed {
 		grpc := cmd.Flag(coreGRPCFlag).Value.String()
 		cfg.GRPCPort = grpc
 	}
--- a/nodebuilder/core/flags.go
+++ b/nodebuilder/core/flags.go
@@ -54,6 +54,7 @@ func ParseFlags(
 		rpc := cmd.Flag(coreRPCFlag).Value.String()
 		cfg.RPCPort = rpc
 	}
+
 	if cmd.Flag(coreGRPCFlag).Changed {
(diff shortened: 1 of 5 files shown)


nodebuilder/core/flags.go
```diff
@@ -50,11 +50,18 @@ func ParseFlags(
...
-	rpc := cmd.Flag(coreRPCFlag).Value.String()
-	grpc := cmd.Flag(coreGRPCFlag).Value.String()
+	// in the case that the RPC port is not set, use the default port
+	// or in the case a value has been set via flag,
+	// set that as highest precedence
+	if cfg.RPCPort == "" || cmd.Flag(coreRPCFlag).Changed {
+		rpc := cmd.Flag(coreRPCFlag).Value.String()
+		cfg.RPCPort = rpc
+	}
+	if cfg.GRPCPort == "" || cmd.Flag(coreGRPCFlag).Changed {
```
Wondertan, line 60: This allows an invalid configuration to be put in place with empty fields. WDYT about prohibiting that and failing the user with Validate? We could only detect that the flag has been changed and alter the config via `cmd.Flag(coreGRPCFlag).Change`

ramin: This actually prevents an empty config being accidentally preserved after ParseFlags if the default flag is not changed and an empty config  is passed in. We never get here because we always (I think) get NodeConfig with defaults. I only encountered from writing tests, feels like its worth a check

agree with adding empty checks to cfg.Validate() ...I'll add

Wondertan: > This actually prevents an empty config being accidentally preserved after ParseFlags if the default flag is not changed and an empty config is passed in.

That's true. My suggestion is to fail if an empty config fields were passed, as opposed to populating them. So we only override fields if flags change, allowing the empty fields case to fail during further validation in ParseFlags

Wondertan (approved):

renaynay (approved):
Thank you!

Issue #3335, opened 2024-04-23
Usability/DevUX Meta Issue


liamsi:
This issue collects smaller quirks with the current API/docs/CLI in one place. I will list them here and break them out into their own issues where it makes sense:

- [x] the comment here is wrong: https://github.com/celestiaorg/celestia-node/blob/f2b664f3abe5e76ed51e19305aa3489980a93692/blob/service.go#L113 (it only returns one blob; GetAll is what returns all blobs)
- [x] GetAll is confusing and edge-cases are under-specified: https://github.com/celestiaorg/celestia-node/issues/3292#issuecomment-2059313529
- [x] some examples in API docs are not representative; e.g. blob.Get returns a different commitment than requested (#3317 )
- [x] CLI: when looking at the for usage returned by `--help` encoding related guidance is often missing, eg. `celestia blob get [height] [namespace] [commitment] [flags]` should state which format the inputs are expected in (hex? base64?). The CLI help should be self-contained; currently one needs to e.g. look [here](https://docs.celestia.org/developers/node-tutorial) to know how to use this
- [x] CLI: namespaces are submitted as hex but returned base64 encoded (which is very confusing); example: 

``` 
celestia blob submit 0x42690c204d39600fddd3 0x676d                                                                                                                  
{
  "result": {
    "height": 1671475,
    "commitments": [
      "0MFhYKQUi2BU+U1jxPzG7QY2BVV1lb3kiU+zAK7nUiY="
    ]
  }
}
⋊> ~/G/celestia-node on main ⨯ celestia blob get-all 1671475  0x42690c204d39600fddd3                                                                                                               12:06:50
{
  "result": [
    {
      "namespace": "AAAAAAAAAAAAAAAAAAAAAAAAAEJpDCBNOWAP3dM=",
      "data": "0x676d",
      "share_version": 0,
      "commitment": "0MFhYKQUi2BU+U1jxPzG7QY2BVV1lb3kiU+zAK7nUiY=",
      "index": 531
    }
  ]
}
```

 - [x] CLI/API: the node daemon gets started with only one account and one can only submit PFBs from that single account; this does not reflect how people interact with chains using different wallets for different purposes (applies to e.g. `celestia blob submit` and a bunch of state related commands / rpcs) https://github.com/celestiaorg/celestia-node/issues/3295
 - [x] #3345
 - [ ] consumers of the API do not have sufficient control for retries on blob submission
[...]

PR #3530, opened 2024-06-26
docs(blob): update godoc comments to better match what Get and GetProof actually do

ramin:
Refs #3335 

Attempts to create more accurate doc/comments on Get and GetProof in `blob/service.go` to improve legibility and comprehension of how these methods behave to API consumers

specifically cc @vgonkivs

### DIFF
--- a/blob/service.go
+++ b/blob/service.go
@@ -138,8 +138,9 @@ func (s *Service) Get(
 	return
 }
 
-// GetProof returns an NMT inclusion proof for a blob within the specified namespace at the given
-// height, using the given commitment. It employs the same algorithm as service.Get() internally.
+// GetProof returns an NMT inclusion proof for a specified namespace to the respective row roots
+// on which the blob spans on at the given height, using the given commitment.
+// It employs the same algorithm as service.Get() internally.
 func (s *Service) GetProof(
 	ctx context.Context,
 	height uint64,
--- a/blob/service.go
+++ b/blob/service.go
@@ -110,9 +110,10 @@ func (s *Service) Submit(ctx context.Context, blobs []*Blob, gasPrice GasPrice) 
 	return uint64(resp.Height), nil
 }
 
-// Get retrieves all the blobs for given namespaces at the given height by commitment.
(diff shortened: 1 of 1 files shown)


blob/service.go
```diff
@@ -136,9 +137,9 @@ func (s *Service) Get(
 	return
 }
 
-// GetProof retrieves all blobs in the given namespaces at the given height by commitment
-// and returns their Proof. It collects all namespaced data from the EDS, constructs blobs
-// and compares commitments.
+// GetProof retrieves a blob in the given namespace at the given height by commitment
+// and returns their Proof. It collects all namespaced data from the EDS, constructs the blob
+// and compares the commitment argument.
```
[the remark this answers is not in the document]
ramin: @walldiss wanna make an inline suggestion of how it should look and we can accept based on that?

walldiss, line 142:
```suggestion
// GetProof returns an NMT inclusion proof for a blob within the specified namespace at the given
// height, using the given commitment. It employs the same algorithm as service.Get() internally.
```

renaynay (approved):

blob/service.go
```diff
@@ -137,9 +138,8 @@ func (s *Service) Get(
 	return
 }
 
-// GetProof retrieves all blobs in the given namespaces at the given height by commitment
-// and returns their Proof. It collects all namespaced data from the EDS, constructs blobs
-// and compares commitments.
+// GetProof returns an NMT inclusion proof for a blob within the specified namespace at the given
```
vgonkivs, line 141: Let's also specify that the Proof itself is not a Blob inclusion Proof but the namespace inclusion proof to the respective row roots.

ramin: @vgonkivs can you draft a suggested edit?

vgonkivs: `// GetProof returns an NMT inclusion proof for a specified namespace to the respective row roots on which the blob spans on`

vgonkivs (approved):

renaynay (approved):
thanks for clarification on the proof

PR #3531, opened 2024-06-26
feat(shwap):Add eds streaming

walldiss:
Add streaming: read/write for files. Read is implemented only for Rsmt2d file by `ReadFrom` method. There is no usecase so farm to directly read for ods file, so `ReadFrom` is not added to eds.Accessor interface


share/new_eds/rsmt2d.go
```diff
@@ -11,20 +12,20 @@ import (
 	"github.com/celestiaorg/celestia-node/share/shwap"
 )
 
-var _ Accessor = Rsmt2D{}
+// var _ Accessor = Rsmt2D{}
```
cristaloleg, line 15: Remove?

store/file/square.go
```diff
@@ -141,3 +150,34 @@ func oppositeAxis(axis rsmt2d.Axis) rsmt2d.Axis {
...
+}
+
+type odsWriter struct {
+	ods square
+	// current is the amount of Shares stored in square that have been written by odsWriter. When
+	// current reaches total, odsWriter will prevent further reads by returning io.EOF
+	current, total int
+}
+
+func (w *odsWriter) WriteTo(writer io.Writer, limit int) (int, error) {
```
cristaloleg, line 168: `WriteTo` is a method from a well-known interface https://pkg.go.dev/io#WriterTo

Quite often breaking such unspoken rules might result in a future incompatibilities between implementation and stdlib interfaces. Just saying.

share/new_eds/reader.go
```diff
@@ -0,0 +1,97 @@
...
+func NewSharesReader(odsSize int, getShare func(rowIdx, colIdx int) ([]byte, error)) *BufferedReader {
+	return &BufferedReader{
+		getShare: getShare,
+		buf:      bytes.NewBuffer(nil),
+		odsSize:  odsSize,
+		total:    odsSize * odsSize,
+	}
+}
+
+// BufferedReader will read Shares from inMemOds into the buffer.
```
Wondertan, line 21:
```suggestion
// BufferedReader will read Shares from getShare function into the buffer.
```

store/file/ods_test.go
```diff
@@ -62,7 +62,9 @@ func TestODSFile(t *testing.T) {
 	t.Cleanup(cancel)
 
 	ODSSize := 8
```
Wondertan, line 64: It is worth increasing the size so it's bigger than the buffer. Ideally, we would do specify MaxODS to the test suite so it tests everything for all the sizes until the max given. It will take more time, yes, but we have high assurance that all sizes work correctly.

store/file/ods.go
```diff
@@ -288,3 +306,24 @@ func (f *ODSFile) axis(ctx context.Context, axisType rsmt2d.Axis, axisIdx int) (
...
+		total:   int64(odsSize*odsSize*shareSize + offset),
+	}
+}
+
+func (w *fileReader) Read(p []byte) (int, error) {
+	if w.current >= w.total {
+		return 0, io.EOF
+	}
+
+	return w.r.ReadAt(p, w.current)
```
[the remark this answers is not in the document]
Wondertan: Well, nvm tests aren't working

share/new_eds/reader.go
```diff
@@ -0,0 +1,97 @@
...
+	if !errors.Is(err, io.ErrUnexpectedEOF) && !errors.Is(err, io.EOF) {
+		return n, fmt.Errorf("unexpected error reading from buf: %w", err)
+	}
+
+	written := n
+	for r.current < r.total {
+		rowIdx, colIdx := r.current/r.odsSize, r.current%r.odsSize
+		share, err := r.getShare(rowIdx, colIdx)
+		if err != nil {
+			return 0, fmt.Errorf("get share; %w", err)
```
Wondertan, line 52:
```suggestion
			return 0, fmt.Errorf("get share: %w", err)
```

share/new_eds/close_once.go
```diff
@@ -11,16 +12,16 @@ import (
...
 var errAccessorClosed = errors.New("accessor is closed")
 
 type closeOnce struct {
-	f      AccessorCloser
+	f      AccessorStreamer
 	closed atomic.Bool
 }
 
-func WithClosedOnce(f AccessorCloser) AccessorCloser {
+func WithClosedOnce(f AccessorStreamer) AccessorStreamer {
```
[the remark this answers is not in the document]
walldiss: Will be used later. It is meant for storage

share/new_eds/rsmt2d.go
```diff
@@ -89,10 +90,36 @@ func (eds Rsmt2D) RowNamespaceData(
...
+		return eds.GetCell(uint(rowIdx), uint(colIdx)), nil
+	}
+	odsSize := int(eds.Width() / 2)
+	reader := NewSharesReader(odsSize, getShare)
+	return reader, nil
+}
+
+// Rsmt2DFromShares reads shares from the provided reader and constructs an Extended Data Square.
+// Provided reader should contain shares in row-major order.
+func Rsmt2DFromShares(shares []share.Share, odsSize int) (Rsmt2D, error) {
```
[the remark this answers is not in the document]
walldiss: It doesn't need reader. It is jsut a constructor from share.

Comment updated

share/new_eds/accessor.go
```diff
@@ -27,17 +27,24 @@ type Accessor interface {
...
+// AccessorStreamer is an interface that groups Accessor and Streamer interfaces.
+type AccessorStreamer interface {
 	Accessor
+	Streamer
+}
+
+type Streamer interface {
+	// Reader returns binary reader for the file (ODS) shares. It should read the shares from the
+	// ODS part of the square row by row.
+	Reader() (io.Reader, error)
```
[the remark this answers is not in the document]
walldiss: Sync: decided to keep Reader as part of Streamer interface and use it only for file implementation for now.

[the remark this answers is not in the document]
walldiss: Sync: ctx not needed if we don't include it into Accessor interface

store/file/ods.go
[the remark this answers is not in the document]
walldiss: removed that code

share/new_eds/accessor.go
```diff
@@ -27,17 +27,24 @@ type Accessor interface {
...
-type accessorCloser struct {
+type accessorStreamer struct {
 	Accessor
-	io.Closer
+	Streamer
 }
 
-func WithCloser(a Accessor, c io.Closer) AccessorCloser {
-	return &accessorCloser{a, c}
+func AccessorAndStreamer(a Accessor, s Streamer) AccessorStreamer {
```
Wondertan, line 48: When we discussed the idiomatic approach, I think you wanted to find `iotest` pkg, which has things like [`TimeoutReader`](https://pkg.go.dev/testing/iotest#TimeoutReader)

share/new_eds/proofs_cache.go
```diff
@@ -211,21 +212,28 @@ func (c *proofsCache) Shares(ctx context.Context) ([]share.Share, error) {
...
-		if err != nil {
-			return nil, err
-		}
-		ax.half = half
+	half, err := c.AxisHalf(ctx, axisType, axisIdx)
+	if err != nil {
+		return nil, err
 	}
 
-	shares, err := ax.half.Extended()
```
Wondertan, line 228: What's the reason for this change? I don't follow.

Wondertan (approved):

share/new_eds/accessor.go
walldiss: // TimeoutReader returns [ErrTimeout](https://pkg.go.dev/testing/iotest#ErrTimeout) on the second read with no data. Subsequent calls to read succeed.

Not sure how it is relevant

share/new_eds/proofs_cache.go
walldiss: It caches half axis independently inside AxisHalf call

PR #3535, opened 2024-06-27
fix(shwap/bitswap): close acccessor in Blockstore

Wondertan:
Blockstore must cleanup the accesor

### DIFF
--- a/share/shwap/p2p/bitswap/block_fetch_test.go
+++ b/share/shwap/p2p/bitswap/block_fetch_test.go
@@ -117,7 +117,7 @@ func TestFetch_Duplicates(t *testing.T) {
 func newExchangeOverEDS(ctx context.Context, t *testing.T, rsmt2d *rsmt2d.ExtendedDataSquare) exchange.SessionExchange {
 	bstore := &Blockstore{
 		Getter: testAccessorGetter{
-			Accessor: &eds.Rsmt2D{ExtendedDataSquare: rsmt2d},
+			AccessorStreamer: &eds.Rsmt2D{ExtendedDataSquare: rsmt2d},
 		},
 	}
 	return newExchange(ctx, t, bstore)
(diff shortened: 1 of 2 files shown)


share/shwap/p2p/bitswap/block_fetch_test.go
```diff
@@ -117,7 +117,7 @@ func TestFetch_Duplicates(t *testing.T) {
 func newExchangeOverEDS(ctx context.Context, t *testing.T, rsmt2d *rsmt2d.ExtendedDataSquare) exchange.SessionExchange {
 	bstore := &Blockstore{
 		Getter: testAccessorGetter{
-			Accessor: eds.Rsmt2D{ExtendedDataSquare: rsmt2d},
+			AccessorCloser: eds.WithCloser(eds.Rsmt2D{ExtendedDataSquare: rsmt2d}, nopCloser{}),
```
Wondertan, line 120: That is exactly why I don't like the `eds.WithCloser` approach @walldiss

walldiss:
Unfortunately would need rebase after merge of streaming.

share/shwap/p2p/bitswap/block_store.go
```diff
@@ -38,6 +38,11 @@ func (b *Blockstore) getBlock(ctx context.Context, cid cid.Cid) (blocks.Block, e
 	if err != nil {
 		return nil, fmt.Errorf("getting EDS Accessor for height %v: %w", blk.Height(), err)
 	}
+	defer func() {
+		if err := eds.Close(); err != nil {
+			log.Warnf("failed to close EDS accessor for height %v: %s", blk.Height(), err)
+		}
+	}()
```
walldiss, line 45: We now have handy utils.CloseAndLog for that, as it becomes common code accross the repo

Wondertan: Its not merged to shwap yet and i dont find it useful honestly. I am gonna keep it as it is here

Wondertan: The log message should contain the context of where the warn/error happens. Using this util will remove any context

walldiss: It takes logger as first argument, so you can pass as much additional context information as you want within loggier fields.

walldiss: This is optional, so  would defer to you to decide.

walldiss (approved):

share/shwap/p2p/bitswap/block_store.go
Wondertan: i dont see any value so gonna keep the way it is

PR #3536, opened 2024-06-27
feat(shwap/bitswap): Shwap optimized Bitswap constructors

Wondertan:
This PR abstracts away Bitswap parameterization and provides simple constructors optimally configured for Shwap protocol composition.

### DIFF
--- a/share/shwap/p2p/bitswap/bitswap.go
+++ b/share/shwap/p2p/bitswap/bitswap.go
@@ -1,0 +1,136 @@
+package bitswap
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/ipfs/boxo/bitswap/client"
+	"github.com/ipfs/boxo/bitswap/network"
+	"github.com/ipfs/boxo/bitswap/server"
+	"github.com/ipfs/boxo/blockstore"
+	delay "github.com/ipfs/go-ipfs-delay"
+	routinghelpers "github.com/libp2p/go-libp2p-routing-helpers"
+	"github.com/libp2p/go-libp2p/core/host"
+	"github.com/libp2p/go-libp2p/core/protocol"
+)
+
+// Client constants
+const (
+	// simulateDontHaves emulates DONT_HAVE message from a peer after 5 second timeout.
+	// This protects us from unresponsive/slow peers.
+	// TODO(@Wondertan): PR to bitswap to make this timeout configurable
+	//  Higher timeout increases the probability of successful reconstruction
+	simulateDontHaves = true
+	// providerSearchDelay defines the initial delay before Bitswap client starts aggressive
+	// broadcasting of WANTs to all the peers. We offset this for longer than the default to minimize
+	// unnecessary broadcasting as in most cases we already have peers connected with needed data on
+	// a new request.
+	providerSearchDelay = time.Second * 10
+	// rebroadcastDelay is similar to the providerSearchDelay, but it targets DHT/ContentRouting
+	// peer discovery and a gentle broadcast of a single random live WANT to all connected peers.
+	// Considering no DHT usage and broadcasting configured by providerSearchDelay, we set
+	// rebroadcastDelay to max value, effectively disabling it
+	rebroadcastDelay = 1<<63 - 1
+)
+
+// Server constants
+const (
+	// providesEnabled dictates Bitswap Server not to provide content to DHT/ContentRouting as we don't use it
+	providesEnabled = false
+	// sendDontHaves prevents Bitswap Server from sending DONT_HAVEs while keeping peers on hold instead:
+	//  * Clients simulate DONT_HAVEs after timeout anyway
+	//  * Servers may not have data immediately and this gives an opportunity to subscribe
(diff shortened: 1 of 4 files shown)


share/shwap/p2p/bitswap/bitswap.go
```diff
@@ -0,0 +1,136 @@
...
+	// providerSearchDelay defines the initial delay before Bitswap client starts aggressive
+	// broadcasting of WANTs to all the peers. We offset this for longer than the default to minimize
+	// unnecessary broadcasting as in most cases we already have peers connected with needed data on
+	// a new request.
+	providerSearchDelay = time.Second * 10
+	// rebroadcastDelay is similar to the providerSearchDelay, but it targets DHT/ContentRouting
+	// peer discovery and a gentle broadcast of a single random live WANT to all connected peers.
+	// Considering no DHT usage and broadcasting configured by providerSearchDelay, we set
+	// rebroadcastDelay to max value, effectively disabling it
+	rebroadcastDelay = 1<<63 - 1
```
walldiss, line 34: Another reason to disable it is that because of erasure coding of EDS we dreconstruct block without specific data to be available. DHT iwant is unnecessary for reconstruction.

share/shwap/p2p/bitswap/bitswap.go
```diff
@@ -0,0 +1,136 @@
...
+
+// Server constants
+const (
+	// providesEnabled dictates Bitswap Server not to provide content to DHT/ContentRouting as we don't use it
+	providesEnabled = false
+	// sendDontHaves prevents Bitswap Server from sending DONT_HAVEs while keeping peers on hold instead:
+	//  * Clients simulate DONT_HAVEs after timeout anyway
+	//  * Servers may not have data immediately and this gives an opportunity to subscribe
+	//  * This is necessary for reconstruction. See https://github.com/celestiaorg/celestia-node/issues/732
+	sendDontHaves = false
```
walldiss, line 45: In other words, turning off don't haves turns bitswap into async/subscribe mod. It is important for cases when we fallback to reconstruction to allow FN to act as samples relay

share/shwap/p2p/bitswap/bitswap.go
```diff
@@ -0,0 +1,136 @@
...
+	// maxServerWantListsPerPeer defines the limit for maximum possible cached wants/requests per peer
+	// in the Bitswap. Exceeding this limit will cause Bitswap server to drop requested wants leaving
+	// client stuck for sometime.
+	// Thus, we make the limit a bit generous, so we minimize the chances of this happening.
+	// This is relevant until https://github.com/ipfs/boxo/pull/629#discussion_r1653362485 is fixed.
+	maxServerWantListsPerPeer = 8096
```
walldiss, line 51: Do you think It will affect the max size of blocks we can reconstruct? WIf yes, would be great to specify max eds size we can support with this value in use.

share/shwap/p2p/bitswap/bitswap.go
```diff
@@ -0,0 +1,136 @@
...
+//	due to a little bug. Bitswap allows setting custom protocols, but
+//	they have to be either one of the switch.
+//	https://github.com/ipfs/boxo/blob/dfd4a53ba828a368cec8d61c3fe12969ac6aa94c/bitswap/network/ipfs_impl.go#L250-L266
+var protocolID = network.ProtocolBitswap
+
+func shwapProtocolID(network protocol.ID) protocol.ID {
+	if network == "" {
+		return ""
+	}
+	return protocol.ID(fmt.Sprintf("/%s/shwap", network))
```
walldiss, line 135: Maybe we should add versioning in case protocol breaking changes?

share/shwap/p2p/bitswap/bitswap.go
Wondertan: Reconstruction will slow down for N*30 secs if the limit is reached, but it still should succeed. The N stands for how many times all the samples didn't fit into the want list.

walldiss: Where 30sec comes from? If the peer sends 16k Iwants to the server, looks like the server will just drop first 8k of requests.  which means it might never reply to those iwant requests. For example, 8096 will not fit 128*128 samples, which might be a potential limitation point for that setting value

share/shwap/p2p/bitswap/bitswap.go
walldiss: Another question I have about protocol ID is the idea we discussed to separate bitswap networks into samples and reconstruction related. Any plans to make constructors aware of network type in this PR? it seems like a PR those changes would fit well

share/shwap/p2p/bitswap/bitswap.go
Wondertan: 30 seconds comes from WANT resend timeout. Client sides resends the WANTs every 30 secs

share/shwap/p2p/bitswap/bitswap.go
Wondertan: I thought of including that in this PR, but decided against that as those constructors should land next to reconstruction retrieval.

walldiss (approved):

PR #3537, opened 2024-06-28
fix(cmd): fix namespace representation in cmd output

### LINKED ISSUE #3335: Usability/DevUX Meta Issue

vgonkivs:
Partly refs #3335

```
celestia blob get 1337536 0x42690c204d39600fddd3 IXg+08HV5RsPF3Lle8PH+B2TUGsGUsBiseflxh6wB5E=
{
  "result": {
    "namespace": "0x42690c204d39600fddd3",
    "data": "gm",
    "share_version": 0,
    "commitment": "IXg+08HV5RsPF3Lle8PH+B2TUGsGUsBiseflxh6wB5E=",
    "index": 1
  }
}
```
```
celestia blob get 1337536 0x42690c204d39600fddd3 IXg+08HV5RsPF3Lle8PH+B2TUGsGUsBiseflxh6wB5E= --base64=true
{
  "result": {
    "data": "Z20=",
    "share_version": 0,
    "commitment": "IXg+08HV5RsPF3Lle8PH+B2TUGsGUsBiseflxh6wB5E=",
    "index": 1
  }
}
```

Current PR fixes namespace representation, so it will be consistent with the passed one. Blob cmds require hex-based namespaces only now.

### DIFF
--- a/nodebuilder/blob/cmd/blob.go
+++ b/nodebuilder/blob/cmd/blob.go
@@ -77,7 +77,7 @@ var getCmd = &cobra.Command{
 			return fmt.Errorf("error parsing a height: %w", err)
 		}
 
-		if !strings.Contains(args[1], "0x") {
+		if !strings.HasPrefix(args[1], "0x") {
 			return fmt.Errorf("only hex namespace is supported")
 		}
 		namespace, err := cmdnode.ParseV0Namespace(args[1])
@@ -117,7 +117,7 @@ var getAllCmd = &cobra.Command{
 		}
 
 		namespace, err := cmdnode.ParseV0Namespace(args[1])
-		if !strings.Contains(args[1], "0x") {
+		if !strings.HasPrefix(args[1], "0x") {
 			return fmt.Errorf("only hex namespace is supported")
 		}
 		if err != nil {
--- a/nodebuilder/blob/cmd/blob.go
+++ b/nodebuilder/blob/cmd/blob.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 	"reflect"
 	"strconv"
+	"strings"
 
 	"github.com/spf13/cobra"
 
@@ -32,14 +33,14 @@ func init() {
 		&base64Flag,
 		"base64",
 		false,
-		"printed blob's data a base64 string",
+		"printed blob's data and namespace as base64 strings",
 	)
 
 	getAllCmd.PersistentFlags().BoolVar(
 		&base64Flag,
 		"base64",
 		false,
-		"printed blob's data as a base64 string",
+		"printed blob's data and namespace as base64 strings",
 	)
 
 	submitCmd.PersistentFlags().Float64Var(
@@ -76,6 +77,9 @@ var getCmd = &cobra.Command{
 			return fmt.Errorf("error parsing a height: %w", err)
 		}
 
+		if !strings.Contains(args[1], "0x") {
+			return fmt.Errorf("only hex namespace is supported")
+		}
 		namespace, err := cmdnode.ParseV0Namespace(args[1])
 		if err != nil {
 			return fmt.Errorf("error parsing a namespace: %w", err)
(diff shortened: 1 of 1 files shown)


ramin (approved):

nodebuilder/blob/cmd/blob.go
```diff
@@ -76,6 +77,9 @@ var getCmd = &cobra.Command{
 			return fmt.Errorf("error parsing a height: %w", err)
 		}
 
+		if !strings.Contains(args[1], "0x") {
```
cristaloleg, line 80: Nit(?): `strings.HasPrefix`

cristaloleg: Same below

cristaloleg (approved):

nodebuilder/blob/cmd/blob.go
vgonkivs: true. thanks

ramin (approved):

vgonkivs:
Yes, non-base64 looks weird and it does not work with get methods `!x\u003e\ufffd\ufffd\ufffd\ufffd\u001b\u000f\u0017r\ufffd{\ufffd\ufffd\ufffd\u001d\ufffdPk\u0006R\ufffdb\ufffd\ufffd\ufffd\ufffd\u001e\ufffd\u0007\ufffd`

walldiss (approved):

jcstein (approved):
utACK

PR #3539, opened 2024-06-28
feat(blob): blobsub

### LINKED ISSUE #2737: [Feature Request]: allow subscribing to new blocks

distractedm1nd:
Introduces `blob.Subscribe`. Takes a `share.Namespace`, returns a `<-chan *BlobsubResponse` containing a `height uint64` and `blobs []*blob.Blob`. 

I made the decision to not support subscribing to multiple namespaces with one call, because the complexity of the return type mitigates the benefits for the client. It is likely more convenient to have one subscription per namespace:

Imagine you are watching two namespaces for different things. On the client side, you would have a `select` statement reading from the two channels in two cases. This is cleaner than receiving everything in one case, but then having to manually parse and send to the respective handlers yourself.  

This closes #2737 , with the caveat that we are not taking the map approach (see rationale above)

### DIFF
--- a/blob/service.go
+++ b/blob/service.go
@@ -60,8 +60,6 @@ type Service struct {
 	headerGetter func(context.Context, uint64) (*header.ExtendedHeader, error)
 	// headerSub subscribes to new headers to supply to blob subscriptions.
 	headerSub func(ctx context.Context) (<-chan *header.ExtendedHeader, error)
-
-	mu sync.Mutex
 }
 
 func NewService(
@@ -95,7 +93,6 @@ type SubscriptionResponse struct {
 
 func (s *Service) Subscribe(ctx context.Context, ns share.Namespace) (<-chan *SubscriptionResponse, error) {
 	if s.ctx == nil {
-		s.mu.Unlock()
 		return nil, fmt.Errorf("service has not been started")
 	}
 
--- a/blob/service.go
+++ b/blob/service.go
@@ -63,6 +63,7 @@ type Service struct {
 
 	// activeSubscriptions tracks the number of active subscriptions
 	activeSubscriptions sync.WaitGroup
+	stopped             bool
 	mu                  sync.Mutex
 }
 
@@ -88,6 +89,7 @@ func (s *Service) Start(context.Context) error {
 func (s *Service) Stop(context.Context) error {
 	s.mu.Lock()
 	if s.cancel != nil {
+		s.stopped = true
 		s.cancel()
 	}
 	s.mu.Unlock()
@@ -103,6 +105,10 @@ type SubscriptionResponse struct {
 
 func (s *Service) Subscribe(ctx context.Context, ns share.Namespace) (<-chan *SubscriptionResponse, error) {
 	s.mu.Lock()
+	if s.stopped {
+		s.mu.Unlock()
+		return nil, fmt.Errorf("service has been stopped")
+	}
 	if s.ctx == nil {
 		s.mu.Unlock()
 		return nil, fmt.Errorf("service has not been started")
@@ -125,6 +131,7 @@ func (s *Service) Subscribe(ctx context.Context, ns share.Namespace) (<-chan *Su
 			select {
 			case header, ok := <-headerCh:
 				if ctx.Err() != nil {
+					log.Debug("blobsub: cancelling subscription due to user context closing")
 					return
 				}
 				if !ok {
@@ -134,6 +141,7 @@ func (s *Service) Subscribe(ctx context.Context, ns share.Namespace) (<-chan *Su
 				blobs, err := s.getAll(ctx, header, []share.Namespace{ns})
 				if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
 					// context canceled, continuing would lead to unexpected missed heights for the client
+					log.Debug("blobsub: cancelling subscription due to user context closing")
 					return
 				}
 				if err != nil {
@@ -143,12 +151,15 @@ func (s *Service) Subscribe(ctx context.Context, ns share.Namespace) (<-chan *Su
 
 				select {
 				case <-ctx.Done():
+					log.Debug("blobsub: cancelling subscription with pending response due to user context closing")
 					return
 				case blobCh <- &SubscriptionResponse{Blobs: blobs, Height: header.Height()}:
 				}
 			case <-ctx.Done():
+				log.Debug("blobsub: cancelling subscription due to user context closing")
 				return
 			case <-s.ctx.Done():
+				log.Debug("blobsub: cancelling subscription due to service context closing")
 				return
 			}
 		}
@@ -252,7 +263,7 @@ func (s *Service) GetAll(ctx context.Context, height uint64, namespaces []share.
 func (s *Service) getAll(
 	ctx context.Context,
 	header *header.ExtendedHeader,
-	namespaces []share.Namespace
+	namespaces []share.Namespace,
 ) ([]*Blob, error) {
 	height := header.Height()
 	var (
--- a/blob/service.go
+++ b/blob/service.go
@@ -55,18 +55,62 @@ type Service struct {
 	shareGetter share.Getter
 	// headerGetter fetches header by the provided height
 	headerGetter func(context.Context, uint64) (*header.ExtendedHeader, error)
+	headerSub    func(ctx context.Context) (<-chan *header.ExtendedHeader, error)
 }
 
 func NewService(
 	submitter Submitter,
 	getter share.Getter,
 	headerGetter func(context.Context, uint64) (*header.ExtendedHeader, error),
+	headerSub func(ctx context.Context) (<-chan *header.ExtendedHeader, error),
 ) *Service {
 	return &Service{
 		blobSubmitter: submitter,
 		shareGetter:   getter,
 		headerGetter:  headerGetter,
+		headerSub:     headerSub,
 	}
+}
+
+type BlobsubResponse struct {
+	blobs  []*Blob
+	height uint64
+}
+
+func (s *Service) Subscribe(ctx context.Context, ns share.Namespace) (<-chan *BlobsubResponse, error) {
+	headerCh, err := s.headerSub(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	blobCh := make(chan *BlobsubResponse)
+	go func() {
+		defer close(blobCh)
+
+		for {
+			select {
+			case header, ok := <-headerCh:
+				if !ok {
+					log.Errorw("header channel closed for blobsub", "namespace", ns.ID())
+					return
+				}
+				blobs, err := s.GetAll(ctx, header.Height(), []share.Namespace{ns})
+				if err != nil {
+					log.Errorw("failed to get blobs", "height", header.Height(), "namespace", ns.ID(), "err", err)
+					continue
+				}
+
+				select {
+				case <-ctx.Done():
+					return
+				case blobCh <- &BlobsubResponse{blobs: blobs, height: header.Height()}:
+				}
+			case <-ctx.Done():
+				return
+			}
+		}
+	}()
+	return blobCh, nil
 }
 
 // Submit sends PFB transaction and reports the height at which it was included.
--- a/blob/service.go
+++ b/blob/service.go
@@ -8,13 +8,14 @@ import (
 	"fmt"
 	"sync"
 
+	"slices"
+
 	"github.com/cosmos/cosmos-sdk/types"
 	logging "github.com/ipfs/go-log/v2"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/trace"
-	"slices"
 
 	"github.com/celestiaorg/celestia-app/pkg/appconsts"
 	appns "github.com/celestiaorg/celestia-app/pkg/namespace"
@@ -60,6 +61,10 @@ type Service struct {
 	headerGetter func(context.Context, uint64) (*header.ExtendedHeader, error)
 	// headerSub subscribes to new headers to supply to blob subscriptions.
 	headerSub func(ctx context.Context) (<-chan *header.ExtendedHeader, error)
+
+	// activeSubscriptions tracks the number of active subscriptions
+	activeSubscriptions sync.WaitGroup
+	mu                  sync.Mutex
 }
 
 func NewService(
@@ -82,7 +87,13 @@ func (s *Service) Start(context.Context) error {
 }
 
 func (s *Service) Stop(context.Context) error {
-	s.cancel()
+	s.mu.Lock()
+	if s.cancel != nil {
+		s.cancel()
+	}
+	s.mu.Unlock()
+
+	s.activeSubscriptions.Wait()
 	return nil
 }
 
@@ -92,17 +103,23 @@ type SubscriptionResponse struct {
 }
 
 func (s *Service) Subscribe(ctx context.Context, ns share.Namespace) (<-chan *SubscriptionResponse, error) {
+	s.mu.Lock()
 	if s.ctx == nil {
+		s.mu.Unlock()
 		return nil, fmt.Errorf("service has not been started")
 	}
+	s.activeSubscriptions.Add(1)
+	s.mu.Unlock()
 
 	headerCh, err := s.headerSub(ctx)
 	if err != nil {
+		s.activeSubscriptions.Done()
 		return nil, err
 	}
 
 	blobCh := make(chan *SubscriptionResponse, 3)
 	go func() {
+		defer s.activeSubscriptions.Done()
 		defer close(blobCh)
 
 		for {
--- a/blob/service.go
+++ b/blob/service.go
@@ -49,13 +49,17 @@ type Submitter interface {
 }
 
 type Service struct {
+	// ctx represents the Service's lifecycle context.
+	ctx    context.Context
+	cancel context.CancelFunc
 	// accessor dials the given celestia-core endpoint to submit blobs.
 	blobSubmitter Submitter
 	// shareGetter retrieves the EDS to fetch all shares from the requested header.
 	shareGetter share.Getter
 	// headerGetter fetches header by the provided height
 	headerGetter func(context.Context, uint64) (*header.ExtendedHeader, error)
-	headerSub    func(ctx context.Context) (<-chan *header.ExtendedHeader, error)
+	// headerSub subscribes to new headers to supply to blob subscriptions.
+	headerSub func(ctx context.Context) (<-chan *header.ExtendedHeader, error)
 }
 
 func NewService(
@@ -72,12 +76,26 @@ func NewService(
 	}
 }
 
+func (s *Service) Start(context.Context) error {
+	s.ctx, s.cancel = context.WithCancel(context.Background())
+	return nil
+}
+
+func (s *Service) Stop(context.Context) error {
+	s.cancel()
+	return nil
+}
+
 type SubscriptionResponse struct {
 	Blobs  []*Blob
 	Height uint64
 }
 
 func (s *Service) Subscribe(ctx context.Context, ns share.Namespace) (<-chan *SubscriptionResponse, error) {
+	if s.ctx != nil {
+		return nil, fmt.Errorf("service has not been started")
+	}
+
 	headerCh, err := s.headerSub(ctx)
 	if err != nil {
 		return nil, err
@@ -95,6 +113,10 @@ func (s *Service) Subscribe(ctx context.Context, ns share.Namespace) (<-chan *Su
 					return
 				}
 				blobs, err := s.GetAll(ctx, header.Height(), []share.Namespace{ns})
+				if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+					// context canceled, continuing would lead to unexpected missed heights for the client
+					return
+				}
 				if err != nil {
 					log.Errorw("failed to get blobs", "height", header.Height(), "namespace", ns.ID(), "err", err)
 					continue
@@ -106,6 +128,8 @@ func (s *Service) Subscribe(ctx context.Context, ns share.Namespace) (<-chan *Su
 				case blobCh <- &SubscriptionResponse{Blobs: blobs, Height: header.Height()}:
 				}
 			case <-ctx.Done():
+				return
+			case <-s.ctx.Done():
 				return
 			}
 		}
--- a/blob/service.go
+++ b/blob/service.go
@@ -6,7 +6,6 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
-	"slices"
 	"sync"
 
 	"github.com/cosmos/cosmos-sdk/types"
@@ -15,6 +14,7 @@ import (
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/trace"
+	"slices"
 
 	"github.com/celestiaorg/celestia-app/pkg/appconsts"
 	appns "github.com/celestiaorg/celestia-app/pkg/namespace"
@@ -92,7 +92,7 @@ type SubscriptionResponse struct {
 }
 
 func (s *Service) Subscribe(ctx context.Context, ns share.Namespace) (<-chan *SubscriptionResponse, error) {
-	if s.ctx != nil {
+	if s.ctx == nil {
 		return nil, fmt.Errorf("service has not been started")
 	}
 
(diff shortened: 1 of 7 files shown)


blob/service.go
```diff
@@ -86,6 +89,47 @@ func DefaultSubmitOptions() *SubmitOptions {
...
+		return nil, err
+	}
+
+	blobCh := make(chan *BlobsubResponse)
+	go func() {
+		defer close(blobCh)
+
+		for {
+			select {
+			case header, ok := <-headerCh:
```
Wondertan, line 109: There is an edge case possible here. The HeaderSub doesn't guarantee a contiguous stream of headers, and there might be gaps that we have to handle. It shouldn't be hard to handle those through the header store, though.

blob/service.go
```diff
@@ -86,6 +89,47 @@ func DefaultSubmitOptions() *SubmitOptions {
 	}
 }
 
+type BlobsubResponse struct {
+	blobs  []*Blob
+	height uint64
```
Wondertan, line 94: We need these public

blob/service.go
```diff
@@ -86,6 +89,47 @@ func DefaultSubmitOptions() *SubmitOptions {
...
+}
+
+func (s *Service) Subscribe(ctx context.Context, ns share.Namespace) (<-chan *BlobsubResponse, error) {
+	headerCh, err := s.headerSub(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	blobCh := make(chan *BlobsubResponse)
```
Wondertan, line 103: A little buffer wouldn't hurt and will lead to less context switching.

blob/service.go
```diff
@@ -86,6 +89,47 @@ func DefaultSubmitOptions() *SubmitOptions {
 	}
 }
 
+type BlobsubResponse struct {
+	blobs  []*Blob
+	height uint64
+}
+
+func (s *Service) Subscribe(ctx context.Context, ns share.Namespace) (<-chan *BlobsubResponse, error) {
```
Wondertan, line 97: These should follow GetAll semantics on the plurality of namespaces. We have yet to decide whether to keep GetAll with a single namespace or multiple; the subscription should be consistent with it, so we should decide on that first.

Wondertan (commented):
Lovely

blob/service.go
```diff
@@ -86,6 +89,47 @@ func DefaultSubmitOptions() *SubmitOptions {
 	}
 }
 
+type BlobsubResponse struct {
```
ramin, line 92: we should move away from these kind of names, it'll be kinder to everyone that uses these types by giving them descriptive names, if boring. This looks like a `SubscriptionResponse`, as it's in the blob module, we know it's the `blob.SubscriptionResponse`, that'd be nice to read and think about.

blob/service.go
```diff
@@ -86,6 +89,47 @@ func DefaultSubmitOptions() *SubmitOptions {
...
+
+	blobCh := make(chan *BlobsubResponse)
+	go func() {
+		defer close(blobCh)
+
+		for {
+			select {
+			case header, ok := <-headerCh:
+				if !ok {
+					log.Errorw("header channel closed for blobsub", "namespace", ns.ID())
```
ramin, line 111: i'd say similar here, blobsub sounds silly and shouldn't be in logs, everywhere else its something sensible like Subscription

blob/service.go
distractedm1nd: definitely, nice catch

blob/service.go
Wondertan: (I would also call it SubscribeAll)

vgonkivs: I don't think we should have multiple namespaces for subscription.  For GetAll it's fine since it is a single call, but for a long-term subscription, it adds complexity in handling on the user's side. So, if you insist on consistency, I'd rather make GetAll with a single namespace than SubscribeAll with multiple namespaces.

blob/service.go
```diff
@@ -86,6 +89,47 @@ func DefaultSubmitOptions() *SubmitOptions {
...
+				}
+				blobs, err := s.GetAll(ctx, header.Height(), []share.Namespace{ns})
+				if err != nil {
+					log.Errorw("failed to get blobs", "height", header.Height(), "namespace", ns.ID(), "err", err)
+					continue
+				}
+
+				select {
+				case <-ctx.Done():
+					return
```
vgonkivs, line 122: IMO you need also to respect the service context(which has not been used so far but we should add it with Start/Stop methods)

vgonkivs (changes requested):

blob/service.go
```diff
@@ -86,6 +89,47 @@ func DefaultSubmitOptions() *SubmitOptions {
...
+			select {
+			case header, ok := <-headerCh:
+				if !ok {
+					log.Errorw("header channel closed for blobsub", "namespace", ns.ID())
+					return
+				}
+				blobs, err := s.GetAll(ctx, header.Height(), []share.Namespace{ns})
+				if err != nil {
+					log.Errorw("failed to get blobs", "height", header.Height(), "namespace", ns.ID(), "err", err)
+					continue
```
cristaloleg, line 117: I think we should treat `context.Canceled` separately and do `return`.

blob/service.go
Wondertan: You would need to write two codepaths for the same blob handling logic, if these are inconsistent. As a dev, I would hate doing the same thing twice.

Wondertan: On the single vs multiple namespace. Try to imagine how the implementation of the protocol that relies on multiple namespaces would look like. If you do it sequentially, you add another roundtrip. If you do it concurrently, you bring complexity for devs to synchronise both.

distractedm1nd: For the reasons I stated in the description, I will not be extending this to add multiple namespaces. I also do not think consistency is so important here, these are two very different types of methods - and as I described, a single channel per namespace makes the most sense for multiple namespace handling anyways.

Changing GetAll is out of scope here and so I would defer that to a discussion. I don't think this feature should be blocked on this at all

blob/service.go
distractedm1nd: I wasn't aware that it doesn't guarantee a contiguous stream of headers - to me, this sounds like something that needs to be fixed in headersub itself. Otherwise, we need to make it public because that is not intuitive and it's how everybody is currently using it. I would be fine holding off on merging this until that is fixed, can we discuss it further?

vgonkivs: What if we will cache non-contiguous headers?

blob/service.go
vgonkivs: > Try to imagine how the implementation of the protocol that relies on multiple namespaces

Having multiple subscriptions and a generic logic that parses them they will directly know from which namespace the blob comes.

blob/service.go
distractedm1nd: done

blob/service.go
Wondertan: Agreed, we can move this issue separately. I think there might be a potential reason to fix this on go-header level with a Subscriber wrapper that does exactly gap filling. This way users relying on header subscriptions would not need to think of that edge case. Need to think of the implications of that tho

Wondertan: Mind parking this in an issue?

ramin (approved):

blob/service.go
```diff
@@ -86,6 +93,67 @@ func DefaultSubmitOptions() *SubmitOptions {
...
+		defer close(blobCh)
+
+		for {
+			select {
+			case header, ok := <-headerCh:
+				if !ok {
+					log.Errorw("header channel closed for subscription", "namespace", ns.ID())
+					return
+				}
+				blobs, err := s.GetAll(ctx, header.Height(), []share.Namespace{ns})
```
vgonkivs, line 132: There is one roundtrip here. You could break retrieve logic into getting header and getting shares. So you don't need to get the header one more time

blob/service.go
```diff
@@ -86,6 +93,67 @@ func DefaultSubmitOptions() *SubmitOptions {
...
 
+func (s *Service) Start(context.Context) error {
+	s.ctx, s.cancel = context.WithCancel(context.Background())
+	return nil
+}
+
+func (s *Service) Stop(context.Context) error {
+	s.cancel()
+	return nil
+}
```
walldiss, line 104: Would be great to have graceful subscribtions shutdown logic here. Stop() should wait for all all ongiong subscriptions to be termintaed before exiting. Could be implemented as separate PR if you don't have time for it right now.

blob/service.go
```diff
@@ -86,6 +93,67 @@ func DefaultSubmitOptions() *SubmitOptions {
...
+	if s.ctx != nil {
+		return nil, fmt.Errorf("service has not been started")
+	}
+
+	headerCh, err := s.headerSub(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	blobCh := make(chan *SubscriptionResponse, 3)
```
walldiss, line 121: Would it work without bufferization? If it is required, why buffer size 3 ?

blob/service.go
```diff
@@ -86,6 +93,67 @@ func DefaultSubmitOptions() *SubmitOptions {
...
+	go func() {
+		defer close(blobCh)
+
+		for {
+			select {
+			case header, ok := <-headerCh:
+				if !ok {
+					log.Errorw("header channel closed for subscription", "namespace", ns.ID())
+					return
+				}
```
walldiss, line 131: headerSub might be closed because context is canceled, which is valid closure of sub and don't need to be logged with ERROR.
Moving handling of ctx errors higher than `!ok`  case should fix that.

blob/service.go
distractedm1nd: nice catch

blob/service.go
```diff
@@ -86,6 +93,67 @@ func DefaultSubmitOptions() *SubmitOptions {
...
+
+type SubscriptionResponse struct {
+	Blobs  []*Blob
+	Height uint64
+}
+
+func (s *Service) Subscribe(ctx context.Context, ns share.Namespace) (<-chan *SubscriptionResponse, error) {
+	if s.ctx != nil {
+		return nil, fmt.Errorf("service has not been started")
+	}
```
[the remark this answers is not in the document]
distractedm1nd: correct, thanks

blob/service.go
distractedm1nd: probably works fine without bufferization, I didn't want to make any assumptions about how often the user tries to read from the channel, and thought about 3 blocktimes is enough "buffer" for this assumption. It's opinonated, and I'm not sure what the best devex would be. Do you have an opinion?

blob/service.go
distractedm1nd: true, nice

distractedm1nd: can you verify that the latest commit is what you had in mind?

blob/service.go
distractedm1nd: could you check the latest commit to verify?

blob/service.go
distractedm1nd: #3578

Wondertan:
Do we want to block this PR on absence proofs? The rough consensus from our call is that we should. Thoughts?

blob/service.go
walldiss: The sub bufferization will also happen on header p2p subscription level. If user is not responsive, we should notify him on being slow reader or even close subscription with error, because eventually it will lead to buffer overflow here or under the hood of go-header sub. 
Perhaps adding select with writing to channel to detect overflow is the best here and of overflow is there return error from sub. Would also bump buffer size to 16 just in case, the bigger size is not a deal breaker.

blob/service.go
```diff
@@ -42,26 +42,113 @@ type Submitter interface {
...
+func (s *Service) Start(context.Context) error {
+	s.ctx, s.cancel = context.WithCancel(context.Background())
+	return nil
+}
+
+func (s *Service) Stop(context.Context) error {
+	s.mu.Lock()
+	if s.cancel != nil {
+		s.cancel()
+	}
```
walldiss, line 85: Re: graceful shutdown. 
Need to set inner ctx to nil here or save some flag indicating that svc stopped, to prevent new subscription from launching

renaynay (dismissed):
lint.

blob/service.go
renaynay: +

distractedm1nd: Have to use flag because setting context to nil will result in panics for existing subs

blob/service.go
```diff
@@ -49,26 +50,131 @@ type Submitter interface {
...
+				if !ok {
+					log.Errorw("header channel closed for subscription", "namespace", ns.ID())
+					return
+				}
+				if len(blobCh) == cap(blobCh)-1 {
+					log.Warn("blobsub: slow consumer detected, closing subscription")
+					// still send an error to the client even though it seems like they've stopped consuming messages
+					blobCh <- &SubscriptionResponse{Error: errors.New("canceling subscription due to slow consumption")}
+					return
+				}
```
walldiss, line 151: There is no reason to send message to the channel, because it will be closed immediately after return and the messages will be gone.
 I would suggest to simplify it a bit by adding default case to the select where you write to blobCh.
```
default:
log.Warn("blobsub: slow consumer detected, closing subscription")
return
```

distractedm1nd: Buffered messages can still be read from a closed channel no?

walldiss: True

Wondertan: Literally was pending comment
image

blob/service.go
```diff
@@ -49,26 +50,131 @@ type Submitter interface {
...
+		return nil, err
+	}
+
+	blobCh := make(chan *SubscriptionResponse, 16)
+	go func() {
+		defer s.activeSubscriptions.Done()
+		defer close(blobCh)
+
+		ticker := time.NewTicker(time.Minute)
+		defer ticker.Stop()
```
Wondertan, line 133: This is not used anywhere

blob/service.go
```diff
@@ -49,26 +50,131 @@ type Submitter interface {
...
+
+func (s *Service) Stop(context.Context) error {
+	s.mu.Lock()
+	if s.cancel != nil {
+		s.stopped = true
+		s.cancel()
+	}
+	s.mu.Unlock()
+
+	s.activeSubscriptions.Wait()
```
Wondertan, line 98: What's the motivation behind this? A slightly retarded user may forget to close a subscription, which will block the node from stopping, and then he will start screaming about the node being wrong as it doesn't stop.

Wondertan: Basically, why do we deprioritize the node lifecycle yet bring more complexity?

blob/service.go
Wondertan, line 151: I don't get the problem we want to solve here. If the blob channel gets overflowed, it will block the header sub channel leaving the slow reader handling on pubsub, which just skips the channel if its getting blocked. I think that's fine because it will be a similar problem to non-contiguous headers. If we implement the solution for that, it can even work in a way that recovers from a slow reader. Once unblocked, the goroutine will detect gaps, read up headers from the store, and serve them to the subscription.

Maybe I miss something and how do you imagine the user handling that?

walldiss: This solution is targeting similar problem, by making it impossible to introduce non-adjustent messages from sub. It will shutdown subscription on errors or buffer overflow. Subscription is much simpler thus more predictable for user. The user of sub would need to cover the gap, if he needs to. Which is the case anyway, because subscription might not start from the point user wants it to.

blob/service.go
walldiss: Subscriptions will be terminated on canceling service context.

Wondertan: So what's the point for waitGroup?

blob/service.go
Wondertan: > The user of sub would need to cover the gap, if he needs to. Which is the case anyway, because subscription might not start from the point user wants it to.

Uncontigious headers is a bug, not a feature. We should fix it internally instead of making every user responsible for that. This is why @distractedm1nd  made: https://github.com/celestiaorg/celestia-node/issues/3578

walldiss: I'm talking about different case, where if user restarts subscription it is possible for the first message it returns to be not adjacent to last message of previous subscription.

Wondertan: Ok, and if we don't send this error message, there is no need to restart the subscription. Problem solved by doing less

blob/service.go
Wondertan: Hmm, I see. I am not sure why we need that tbh,

Header comes every 12 seconds. The service context gets canceled on Stop instantaneously, stopping all the subscriptions that, in common case, sit still. Then they finish the waitgroup unblocking the Stop. How is that different from being disgraceful?

blob/service.go
distractedm1nd: Oh yeah that was from a previous approach, thanks for catching

blob/service.go
distractedm1nd: I implemented this at the request of reviewers so I don't really know what to input here - I do not have a preference so once the debate is over I will go with whatever is chosen

Wondertan: Basically:
* A slow reader is not gonna read this error because its slow
* Even If they gonna read it, this means they recovered from being slow, and in such case, they benefit more from a subscription recovery rather than canceling.
* We get subscription recovery by not doing anything rn and fixing https://github.com/celestiaorg/celestia-node/issues/3578 later on.

walldiss: I think 
- we should at least close subscription on buffer overflow
- message is not necessary, but also seems to be not a big deal to have it

blob/service.go
walldiss: The forcefull shutdown is 4 more lines of code, I can help to add it.

blob/service.go
Wondertan: We synced on that and agreed to close the subscription on buffer overflow without sending any error message, so we can remove the error field

blob/service.go
Wondertan: We synced up on this and agreed to remove the graceful shutdown.

renaynay (commented):
Agree with hlib's comments re buffer + re shutdown

nodebuilder/blob/module.go
```diff
@@ -17,12 +17,31 @@ func ConstructModule() fx.Option {
...
+				return service.Subscribe
+			},
+		),
+		fx.Provide(fx.Annotate(
+			func(
+				state state.Module,
+				sGetter share.Getter,
+				getByHeightFn func(context.Context, uint64) (*header.ExtendedHeader, error),
+				subscribeFn func(context.Context) (<-chan *header.ExtendedHeader, error),
+			) *blob.Service {
```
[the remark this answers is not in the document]
distractedm1nd: need it for tests

blob/service.go
distractedm1nd: Not removing the error field because it could come from GetAll as well, but will do the rest

blob/service.go
```diff
@@ -49,26 +49,111 @@ type Submitter interface {
...
+				var blobs []*Blob
+				var err error
+				for {
+					blobs, err = s.getAll(ctx, header, []share.Namespace{ns})
+					if err == nil || ctx.Err() != nil {
+						// operation successful, break the loop
+						break
+					}
+				}
+				if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
```
renaynay, line 133: I don't understand why a context with a deadline would be used here but okay

renaynay (dismissed):

blob/service.go
```diff
@@ -49,26 +49,111 @@ type Submitter interface {
...
+					// context canceled, continuing would lead to unexpected missed heights for the client
+					log.Debug("blobsub: canceling subscription due to user context closing")
+					return
+				}
```
Wondertan, line 137: * Do context check once and return immediately
* Check for any ctx error, rather than two specific, as there might be more added.

```suggestion
				var blobs []*Blob
				var err error
				for {
					blobs, err = s.getAll(ctx, header, []share.Namespace{ns})
					if ctx.Err() != nil {
						// context canceled, continuing would lead to unexpected missed heights for the client
						log.Debug("blobsub: canceling subscription due to user context closing")
						return
					}
					if err == nil ||  {
						// operation successful, break the loop
						break
					}
				}
	
```

blob/service.go
```diff
@@ -49,26 +49,111 @@ type Submitter interface {
...
+				if !ok {
+					log.Errorw("header channel closed for subscription", "namespace", ns.ID())
+					return
+				}
+				// close subscription before buffer overflows
+				if len(blobCh) == cap(blobCh)-1 {
+					return
+				}
```
Wondertan, line 122: * Add log
* -1 is not needed
```suggestion
				if len(blobCh) == cap(blobCh) {
					log.Warnff("blob subscription cancelled: buffer overflow due to slow reader")
					return
				}
```

blob/service.go
```diff
@@ -49,26 +49,111 @@ type Submitter interface {
...
+	s.ctx, s.cancel = context.WithCancel(context.Background())
+	return nil
+}
+
+func (s *Service) Stop(context.Context) error {
+	s.cancel()
+	return nil
+}
+
+type SubscriptionResponse struct {
```
Wondertan, line 89: Godoc

blob/service.go
```diff
@@ -49,26 +49,111 @@ type Submitter interface {
...
+	Blobs  []*Blob
+	Height uint64
+}
+
+func (s *Service) Subscribe(ctx context.Context, ns share.Namespace) (<-chan *SubscriptionResponse, error) {
+	if s.ctx == nil {
+		return nil, fmt.Errorf("service has not been started")
+	}
```
walldiss, line 97: Actually a racy check. Write of `s.ctx` will be from another go-routine

state/core_access.go
```diff
@@ -184,9 +184,9 @@ func (ca *CoreAccessor) cancelCtx() {
 	ca.cancel = nil
 }
 
-// SubmitPayForBlob builds, signs, and synchronously submits a MsgPayForBlob with additional options defined
-// in `TxConfig`. It blocks until the transaction is committed and returns the TxResponse.
-// The user can specify additional options that can bee applied to the Tx.
+// SubmitPayForBlob builds, signs, and synchronously submits a MsgPayForBlob with additional
+// options defined in `TxConfig`. It blocks until the transaction is committed and returns the
+// TxResponse. The user can specify additional options that can bee applied to the Tx.
```
Wondertan, line 189: lets try to avoid unnecessary changes in this PR.

(I know this lint issue is annoying and we have to fix it in a separate PR)

distractedm1nd: No, unnecessary nit. I got asked to make the linter pass. I'm not going to continue subjecting to back and forths in a PR review

blob/service.go
distractedm1nd: Yes. If a user subscribes in the moment Start is being called in the Service on node startup, they will need to try again in a second. I don't think this is worth fixing for the sake of simplicity.

state/core_access.go
Wondertan: It's valid to ask not to make changes in irrelevant pkgs. The CI linter would pass without those, and the mismatch between ci and local has to be fixed anyway. I am not requesting changes with this comment

Wondertan: I fine to live that as it is, as an exception

blob/service.go
```diff
@@ -49,26 +49,119 @@ type Submitter interface {
...
+
+				var blobs []*Blob
+				var err error
+				for {
+					if ctx.Err() != nil {
+						// context canceled, continuing would lead to unexpected missed heights for the client
+						log.Debugw("blobsub: canceling subscription due to user ctx closing", "namespace", ns.ID())
+						return
+					}
+					blobs, err = s.getAll(ctx, header, []share.Namespace{ns})
```
Wondertan, line 140: I guess you meant to put it after
```suggestion
					blobs, err = s.getAll(ctx, header, []share.Namespace{ns})
							if ctx.Err() != nil {
						// context canceled, continuing would lead to unexpected missed heights for the client
						log.Debugw("blobsub: canceling subscription due to user ctx closing", "namespace", ns.ID())
						return
					}
```

blob/service.go
```diff
@@ -49,26 +49,119 @@ type Submitter interface {
...
+				if len(blobCh) == cap(blobCh) {
+					log.Debugw("blobsub: canceling subscription due to buffer overflow from slow reader", "namespace", ns.ID())
+					return
+				}
+
+				var blobs []*Blob
+				var err error
+				for {
+					blobs, err = s.getAll(ctx, header, []share.Namespace{ns})
+							if ctx.Err() != nil {
```
Wondertan, line 136: My suggestion broke linter
```suggestion
						if ctx.Err() != nil {
```

Wondertan: (i don't know if this fixes it, so better be verified locally)

Wondertan (approved):

vgonkivs (approved):

PR #3545, opened 2024-07-02
feat(shwap): Add eds store

walldiss:
Add eds store engine
Based on https://github.com/celestiaorg/celestia-node/pull/3543


share/new_eds/accessor.go
```diff
@@ -14,6 +14,8 @@ import (
 type Accessor interface {
 	// Size returns square size of the Accessor.
 	Size(ctx context.Context) int
+	// DataHash returns data hash of the Accessor.
+	DataHash(ctx context.Context) (share.DataHash, error)
```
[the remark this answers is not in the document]
Wondertan: But could also rename here as its a low effort thingy

store/metrics.go
```diff
@@ -0,0 +1,130 @@
...
+		return err
+	}
+
+	remove, err := meter.Float64Histogram("eds_store_remove_time_histogram",
+		metric.WithDescription("eds store remove time histogram(s)"))
+	if err != nil {
+		return err
+	}
+
+	if err = s.cache.EnableMetrics(); err != nil {
```
Wondertan, line 58: We should also call close somewhere

store/store.go
```diff
@@ -0,0 +1,519 @@
...
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, ErrNotFound
+		}
+		return nil, fmt.Errorf("opening ODS file: %w", err)
+	}
+	return wrappedFile(f), nil
+}
+
+func (s *Store) LinkHashToHeight(_ context.Context, datahash share.DataHash, height uint64) error {
```
Wondertan, line 224: Why do we need this public?

store/store.go
```diff
@@ -0,0 +1,519 @@
...
+	lock.Lock()
+	defer lock.Unlock()
+
+	tNow := time.Now()
+	err := s.remove(ctx, height)
+	s.metrics.observeRemove(ctx, time.Since(tNow), err != nil)
+	return err
+}
+
+func (s *Store) remove(ctx context.Context, height uint64) error {
```
Wondertan, line 347: The remove logic misses empty height clean up.

Wondertan: Honestly, I still feel like the map and a single file approach add complexity to the code for no reason. We need to iterate over all the heights in the file remove the unneeded one in place, and rewrite/reencode the whole thing.


While storing and checking for an empty file is as simple as this:

```go
// createEmptyFile ensures the file exists and is empty
func createEmptyFile(filePath string) error {
	// Create or truncate the file
	file, err := os.Create(filePath)
	if err != nil {
		return err
	}
	// Close the file after creation/truncation
	defer file.Close()
	return nil
}

// checkFileEmpty checks if the given file exists and is empty
func checkFileEmpty(filePath string) (exists bool, empty bool, err error) {
	// Get the file info
	fileInfo, err := os.Stat(filePath)
	if err != nil {
		if os.IsNotExist(err) {
			// File does not exist
			return false, false, nil
		}
		// Other error
		return false, false, err
	}

	// File exists, check if it is empty
	return true, fileInfo.Size() == 0, nil
}
```

Wondertan: In the end, the proposed solution will have much less code for empty heights handling, which will never be a bottleneck. Current solution feels overengineered for what we need

store/store.go
```diff
@@ -0,0 +1,519 @@
...
+func (s *Store) Close() error {
+	return s.storeEmptyHeights()
+}
+
+func (s *Store) Put(
+	ctx context.Context,
+	datahash share.DataHash,
+	height uint64,
+	square *rsmt2d.ExtendedDataSquare,
+) (eds.AccessorStreamer, error) {
```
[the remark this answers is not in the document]
walldiss: The biggest difficulty to do it here is adding an ability for Accessor to read Q4 for Q1Q4 file. Too many changes for already massive PR here. Will do as separate one

store/store.go
```diff
@@ -0,0 +1,519 @@
...
+func (s *Store) createFile(
+	filePath string,
+	datahash share.DataHash,
+	square *rsmt2d.ExtendedDataSquare,
+) (eds.AccessorStreamer, error) {
+	// check if file with the same hash already exists
+	f, err := s.getByHash(datahash)
+	if err == nil {
+		return f, nil
+	}
```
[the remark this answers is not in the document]
walldiss: File with same hash could exist on different height

store/store.go
walldiss: It will be used to short circuit add only link, if hash exists. Will be usefull in Availability, when we know datahash and heigh of DAH, so we can avoid doing network request and just add height link.

store/store.go
walldiss: In real world scenario we would barely have empty heights. The suggested solution would need to perform checkEmptyFile on every file read, even for non empty files.  I tried t avoid by having inmemory map lookup instead, which is much faster and easier, as it does not disk I/O.
Removal is intentionally not implemented. It might looks missing, but there is no actual value to make it. We don't expect empty height and each will take only 32 bytes on disk. Not worth to add complexity to delete this data

store/store.go
```diff
@@ -0,0 +1,519 @@
...
+	count, err := linksCount(hashPath)
+	if err != nil {
+		return fmt.Errorf("counting links: %w", err)
+	}
+	if count == 1 {
+		err = os.Remove(hashPath)
+		if err != nil {
+			return fmt.Errorf("removing by hash: %w", err)
+		}
+	}
```
Wondertan, line 392: We can simplify removal logic as that's impossible to make duplicate files unless its an empty file that we can handle differently

store/store.go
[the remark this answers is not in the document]
Wondertan: As discussed, this is impossible, so I would remove this additional IO lookup.

store/store.go
```diff
@@ -0,0 +1,415 @@
...
+	if datahash.IsEmptyRoot() {
+		err := s.addEmptyHeight(height)
+		return emptyAccessor, err
+	}
+
+	// short circuit if file exists
+	if has, _ := s.hasByHeight(height); has {
+		s.metrics.observePutExist(ctx)
+		return s.getByHeight(height)
+	}
```
Wondertan, line 128: optional optimization: 
we could do `getByHeight` here, and if it exists, return directly. In case of an error create a file. Otherwise we do duplicate work across both calls

store/store.go
Wondertan: Sync: decide to go with a single empty file and multiple links

store/store.go
```diff
@@ -0,0 +1,415 @@
...
+	_, err := s.cache.Get(height)
+	if err == nil {
+		return true, nil
+	}
+
+	path := s.basepath + heightsPath + fmt.Sprintf("%d", height)
+	return pathExists(path)
+}
+
+func (s *Store) Remove(ctx context.Context, height uint64) error {
```
Wondertan, line 272: We may have a slightly more optimized version of this method that takes both `height` and `hash.` The pruner who will be calling this has full access to the header.

store/store.go
```diff
@@ -0,0 +1,415 @@
...
+	if err != nil {
+		return err
+	}
+	if has {
+		return nil
+	}
+	return s.createHeightLink(share.EmptyRoot().Hash(), height)
+}
+
+func createEmptyFile(basepath string) error {
```
Wondertan, line 398:
```suggestion
func ensureEmptyFile(basepath string) error {
```

store/store.go
```diff
@@ -0,0 +1,415 @@
...
+func (s *Store) addEmptyHeight(height uint64) error {
+	// short circuit if link exists
+	has, err := s.hasByHeight(height)
+	if err != nil {
+		return err
+	}
+	if has {
+		return nil
+	}
+	return s.createHeightLink(share.EmptyRoot().Hash(), height)
```
Wondertan, line 395: We actually don't need to do `hasByHeight` here. We can check inside `s.createHeightLink` if the returned error form `os.Link` is `os.IsExist` and if so, return success.

store/store_options.go
```diff
@@ -0,0 +1,33 @@
...
+	}
+}
+
+func (p *Parameters) Validate() error {
+	if p.RecentBlocksCacheSize < 1 {
+		return errors.New("recent blocks cache size must be positive")
+	}
+
+	if p.AvailabilityCacheSize < 1 {
+		return errors.New("blockstore cache size must be positive")
```
Wondertan, line 30:
```suggestion
		return errors.New("availability cache size must be positive")
```

store/file/ods.go
```diff
@@ -16,6 +16,8 @@ import (
 
 var _ eds.AccessorStreamer = (*ODSFile)(nil)
 
+var ErrFileIsEmpty = fmt.Errorf("file is empty")
```
Wondertan, line 19: Nit: 'ErrEmptyFile'

state/core_access.go
```diff
@@ -184,9 +184,9 @@ func (ca *CoreAccessor) cancelCtx() {
 	ca.cancel = nil
 }
 
-// SubmitPayForBlob builds, signs, and synchronously submits a MsgPayForBlob with additional options defined
-// in `TxConfig`. It blocks until the transaction is committed and returns the TxResponse.
-// The user can specify additional options that can bee applied to the Tx.
+// SubmitPayForBlob builds, signs, and synchronously submits a MsgPayForBlob with additional
+// options defined in `TxConfig`. It blocks until the transaction is committed and returns the
+// TxResponse. The user can specify additional options that can bee applied to the Tx.
```
Wondertan, line 189: Do you know what happened here and other unrelated lint places?

store/file/ods.go
```diff
@@ -16,6 +16,8 @@ import (
 
 var _ eds.AccessorStreamer = (*ODSFile)(nil)
 
+var ErrEmptyFile = fmt.Errorf("file is empty")
```
Wondertan, line 19:
```suggestion
// ErrEmptyFile signals that the ODS file is empty.
// This helps avoid storing empty block EDSes.
var ErrEmptyFile = fmt.Errorf("file is empty")
```

store/store.go
```diff
@@ -0,0 +1,377 @@
...
+	tNow := time.Now()
+	f, err := s.getByHash(datahash)
+	s.metrics.observeGet(ctx, time.Since(tNow), err != nil)
+	return f, err
+}
+
+func (s *Store) getByHash(datahash share.DataHash) (eds.AccessorStreamer, error) {
+	if datahash.IsEmptyRoot() {
+		return emptyAccessor, nil
+	}
```
Wondertan, line 172: Duplicated on the 156 loc, so can be removed.

store/store.go
```diff
@@ -0,0 +1,377 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("creating empty file: %w", err)
+	}
+
+	recentEDSCache, err := cache.NewAccessorCache("recent", params.RecentBlocksCacheSize)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create recent eds cache: %w", err)
+	}
+
+	availabilityCache, err := cache.NewAccessorCache("blockstore", params.AvailabilityCacheSize)
```
Wondertan, line 80:
```suggestion
	availabilityCache, err := cache.NewAccessorCache("availability", params.AvailabilityCacheSize)
```

store/store.go
```diff
@@ -0,0 +1,377 @@
...
+func (s *Store) createFile(
+	datahash share.DataHash,
+	height uint64,
+	square *rsmt2d.ExtendedDataSquare,
+) (eds.AccessorStreamer, error) {
+	// create Q1Q4 file
+	filePath := s.basepath + blocksPath + datahash.String()
+	f, err := file.CreateQ1Q4File(filePath, datahash, square)
+	if err != nil {
+		return nil, fmt.Errorf("creating ODS file: %w", err)
```
Wondertan, line 144:
```suggestion
		return nil, fmt.Errorf("creating Q1Q4 file: %w", err)
```

store/store.go
```diff
@@ -0,0 +1,377 @@
...
+	tNow := time.Now()
+	exist, err := s.hasByHash(datahash)
+	s.metrics.observeHas(ctx, time.Since(tNow), err != nil)
+	return exist, err
+}
+
+func (s *Store) hasByHash(datahash share.DataHash) (bool, error) {
+	if datahash.IsEmptyRoot() {
+		return true, nil
+	}
```
Wondertan, line 225: Same here

store/store.go
```diff
@@ -0,0 +1,377 @@
...
+	return nil
+}
+
+func (s *Store) removeLink(height uint64) error {
+	if err := s.cache.Remove(height); err != nil {
+		return fmt.Errorf("removing from cache: %w", err)
+	}
+
+	// remove hard link by height
+	heightPath := s.basepath + heightsPath + fmt.Sprintf("%d", height)
```
Wondertan, line 295: nit: strconv seems to be more efficient.

Wondertan: we could also extract path building in a separate func and use it everywhere

store/store.go
```diff
@@ -0,0 +1,377 @@
...
+	defaultDirPerm = 0o755
+)
+
+var ErrNotFound = errors.New("eds not found in store")
+
+// Store is a storage for EDS files. It persists EDS files on disk in form of Q1Q4 files or ODS
+// files. It provides methods to put, get and remove EDS files. It has two caches: recent eds cache
+// and availability cache. Recent eds cache is used to cache recent blocks. Availability cache is
+// used to cache blocks that are accessed by sample requests. Store is thread-safe.
+type Store struct {
```
Wondertan, line 40: For archival support we will eventually need a method that converts Q1Q4 into ODS

state/core_access.go
walldiss: `make fmt` fixes

store/store.go
walldiss: We already support archival, but it will be part of archival pruning. It is planned as part of hardening.

state/core_access.go
Wondertan: Weird, would be nice to remove these

store/store_test.go
```diff
@@ -0,0 +1,336 @@
...
+func TestEDSStore(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
+	t.Cleanup(cancel)
+
+	edsStore, err := NewStore(DefaultParameters(), t.TempDir())
+	require.NoError(t, err)
+
+	// disable cache
+	edsStore.cache = cache.NewDoubleCache(cache.NoopCache{}, cache.NoopCache{})
+	height := atomic.NewUint64(100)
```
Wondertan, line 28: Let's use std atomic

store/store_test.go
```diff
@@ -0,0 +1,336 @@
...
+		edsStore, err := NewStore(DefaultParameters(), b.TempDir())
+		require.NoError(b, err)
+
+		// disable cache
+		edsStore.cache = cache.NewDoubleCache(cache.NoopCache{}, cache.NoopCache{})
+
+		dah, err := share.NewRoot(eds)
+		require.NoError(b, err)
+
+		height := uint64(1984)
```
Wondertan, line 316: lol

store/store_test.go
```diff
@@ -0,0 +1,336 @@
...
+
+		// check that cached file is the same eds
+		fromFile, err := f.Shares(ctx)
+		require.NoError(t, err)
+		require.NoError(t, f.Close())
+		expected := eds.FlattenedODS()
+		require.Equal(t, expected, fromFile)
+	})
+
+	t.Run("GetByDataHash", func(t *testing.T) {
```
Wondertan, line 108:
```suggestion
	t.Run("GetByHash", func(t *testing.T) {
```

Ну либо метод поменять

store/store.go
```diff
@@ -0,0 +1,371 @@
...
+	if err != nil {
+		s.metrics.observePut(ctx, time.Since(tNow), square.Width(), true)
+		return nil, fmt.Errorf("creating file: %w", err)
+	}
+	s.metrics.observePut(ctx, time.Since(tNow), square.Width(), false)
+
+	// put file in recent cache
+	f, err = s.cache.First().GetOrLoad(ctx, height, fileLoader(f))
+	if err != nil {
+		log.Warnf("failed to put file in recent cache: %s", err)
```
Wondertan, line 130: Looks like an error to me

store/store_test.go
Wondertan: do go mod tidy after

store/store.go
```diff
@@ -0,0 +1,356 @@
...
+func (s *Store) Close() error {
+	return s.metrics.close()
+}
+
+func (s *Store) Put(
+	ctx context.Context,
+	datahash share.DataHash,
+	height uint64,
+	square *rsmt2d.ExtendedDataSquare,
+) (eds.AccessorStreamer, error) {
```
Wondertan, line 102: How do we use Accessor on Put?

store/store.go
```diff
@@ -0,0 +1,356 @@
...
+	tNow := time.Now()
+	lock := s.stripLock.byDatahashAndHeight(datahash, height)
```
Wondertan, line 104: Why do we need to lock by both here? Isn't datahash enough already?

Wondertan: Also, if we need it here, why don't we do the same for the `Remove`? There, we do locks separately.

store/store.go
```diff
@@ -0,0 +1,356 @@
...
+	if datahash.IsEmptyRoot() {
+		err := s.ensureHeightLink(path, height)
+		return emptyAccessor, err
+	}
+
+	// ensure Q1Q4 file
+	f, existed, err := file.CreateOrOpenQ1Q4File(path, datahash, square)
+	if existed {
+		s.metrics.observePutExist(ctx)
+	}
```
Wondertan, line 118: It might be worth making `existed` an attribute on observePut, rather than separate counter

walldiss:
We both want to shwap sooner, but lets keep using review to discuss changes first. instead of committing your ideas to PR first and discussing afterwards.

Wondertan:
Yeah, doing in a diff PR pointing to this PR would be better

store/store.go
Wondertan: sync: we don't need locks theoretically and we could remove them, but needs further testing and edgecasing

Wondertan (approved):

PR #2956, opened 2023-11-24
changed user UserAgent name

ATREAY:
## Description
This pull request introduces modifications to the user agent/network name.

## Issues
fixes: #2932 , #2908

Issue #2932, opened 2023-11-17
modp2p: change user UserAgent name


Wondertan:
Currently, its `celestia-<network-name>`, which in mainnet it looks like `celestia-celestia`.We might just trim `celestia-` prefix.

Related #2908

ATREAY:
Hey @Wondertan, can I work on this issue if you are not currently working on it?

Wondertan:
Hey @ATREAY, yes, go for it! 

The file where this is defined is at nodebuilder/p2p/host.go. 
Would be great to fix #2908 as well

Wondertan:
The network version is encoded in the network string. p2p.Nerwork

ATREAY:
Hey @Wondertan, I've made the necessary changes. Could you please take a look whenever you have a moment and let me know if any changes are required?

PR #3547, opened 2024-07-03
fix(nodebuilder/p2p): remove prepended v against detected semantic version 

ramin:
It looks as though when the useragent name request of #2932 was addressed in https://github.com/celestiaorg/celestia-node/pull/2956 a 'v' was added to the string formatting with the `BuildInfo.SemanticVersion` which is added to the useragent. (see: https://github.com/celestiaorg/celestia-node/pull/2956/files#diff-78ab4bf0b5adebb93a427d5c0222a637919ca538457678126428146a3bf3c663R35) and carried over in the PR with tests and refactoring (https://github.com/celestiaorg/celestia-node/commit/86f4b2aa1dd54a390dd213d5011b48ab51aa7321#diff-4bbc044dfe14e79d9f168542d12302c631a91f5b5d9c17f55e5ef81ea7fb5611R35)

We actually set this SemanticVersion from git tags at build time [here](https://github.com/celestiaorg/celestia-node/blob/main/Makefile#L6) with ldflags and the tag already has `v` in the naming of the tag. This introduced our useragent having two v, like a va-va-voom, eg: `rcelestia-node/celestia/light/vv0.14.0/13439cc`

This removes the `v` added in the string and takes the tag and set SemanticVersion value as is, removing the vv-stutter

### DIFF
--- a/nodebuilder/node/buildInfo.go
+++ b/nodebuilder/node/buildInfo.go
@@ -32,7 +32,7 @@ func (b *BuildInfo) GetSemanticVersion() string {
 		return emptyValue
 	}
 
-	return fmt.Sprintf("v%s", b.SemanticVersion)
+	return fmt.Sprintf("%s", b.SemanticVersion)
 }
 
 func (b *BuildInfo) CommitShortSha() string {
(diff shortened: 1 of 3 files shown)


Wondertan (approved):
The quickest patch on the west 🔫

cristaloleg (approved):

PR #3551, opened 2024-07-03
feat(shwap): Separate verify and validate for shwap ids

walldiss:
PR separates Verify and validate functions for shwap id types. The change is crutial to allow server side application to do initial validation on shwapID, that server as requests in shwap, before accessing DAH. Verify is meant to be used only by server, after Validation passed and when there is access to the file.


Wondertan (approved):
Reviewed and pushed review changes

PR #3552, opened 2024-07-04
refactor(pruner): reset maxHeadersPerLoop properly

cristaloleg:
We were resetting `maxHeadersPerLoop` to a different value (`1024`) which has no relations to the default value (`512`). Which isn't that bad because it's a test but still we have a side-effect in the `TestPrune_LargeNumberOfBlocks` test.

Also, changing type of `maxHeadersPerLoop` to `int` 'cause in most places we use it as an `int` and not `int64`.

### DIFF
--- a/pruner/find.go
+++ b/pruner/find.go
@@ -10,7 +10,7 @@ import (
 // maxHeadersPerLoop is the maximum number of headers to fetch
 // for a prune loop (prevents fetching too many headers at a
 // time for nodes that have a large number of pruneable headers).
-var maxHeadersPerLoop = uint64(512)
+var maxHeadersPerLoop = 512
 
 // findPruneableHeaders returns all headers that are eligible for pruning
 // (outside the sampling window).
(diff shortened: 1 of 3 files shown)


pruner/find.go
```diff
@@ -10,7 +10,7 @@ import (
 // maxHeadersPerLoop is the maximum number of headers to fetch
 // for a prune loop (prevents fetching too many headers at a
 // time for nodes that have a large number of pruneable headers).
-var maxHeadersPerLoop = uint64(512)
```
vgonkivs, line 13: maybe const then?

cristaloleg: We change it's value in test. I don't have a good reason to redesign logic around this variable now. Unless someone is also against unexported global variable.

vgonkivs (approved):

renaynay (approved):

renaynay:
thank you!

PR #3560, opened 2024-07-11
feat(modp2p): websocket transport with TLS

vgonkivs:
Added support of the TLS certificates to satisfy conditions for the websocket connection. MinTLS version was set to 1.2 as a default version for now.(used lower versions but linter returned errors)

Screenshot 2024-07-11 at 12 01 05


ramin (approved):

renaynay:
@vgonkivs can you please convert to draft and wait for this to be tested? Let's wait to confirm that this actually makes lumina work on safari

Wondertan (approved):
Approving as we need this ASAP.

But generally I think this PR does more than needed. We don't need to store TLS config and we don't need a separate config for "tls" addresses. We only need a simple switch to enable WSS with the path to certificate and key, which could be an ENVVAR.

nodebuilder/p2p/tls.go
```diff
@@ -0,0 +1,38 @@
+package p2p
+
+import (
+	cfg "crypto/tls"
```
cristaloleg, line 4: Curious why it got aliased?

nodebuilder/p2p/tls.go
```diff
@@ -0,0 +1,38 @@
...
+func enableWss() (libp2p.Option, bool, error) {
+	path := os.Getenv(tlsPath)
+	exist := utils.Exists(path+cert) && utils.Exists(path+key)
+	if !exist {
+		return libp2p.Transport(ws.New), exist, nil
+	}
+
+	var certificates []cfg.Certificate
+	if path != "" {
+		cert, err := cfg.LoadX509KeyPair(path+cert, path+key)
```
cristaloleg, line 29: This should be `filepath.Join` (TBH same for `utils.Exist`)(also `cert` and `key` consts without `/`)

nodebuilder/p2p/tls.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+const (
+	cert = "/cert.pem"
+	key  = "/key.pem"
+)
+
+var tlsPath = "TLS_PATH"
+
+func enableWss() (libp2p.Option, bool, error) {
```
cristaloleg, line 20: Can we add a comment what `bool` means here, please?

state/core_access.go
```diff
@@ -594,7 +594,7 @@ func (ca *CoreAccessor) setupTxClient(ctx context.Context, keyName string) (*use
 	}
 	ca.defaultSignerAddress = addr
 	return user.SetupTxClient(ctx, ca.keyring, ca.coreConn, encCfg,
-		user.WithDefaultAccount(keyName), user.WithDefaultAddress(addr),
```
renaynay, line 597: why removing this in this PR? (and why in general?)

renaynay (dismissed):

nodebuilder/p2p/tls.go
vgonkivs: copy pasted from the previous version 🙂 where I have one more tls import. will fix. change

state/core_access.go
vgonkivs: It was a temporary solution, as I my node crashed multiple times(in the app code) on the previous version of app. Will revert changes since we've bumped it

nodebuilder/p2p/tls.go
```diff
@@ -0,0 +1,38 @@
...
+
+	"github.com/celestiaorg/celestia-node/libs/utils"
+)
+
+const (
+	cert = "/cert.pem"
+	key  = "/key.pem"
+)
+
+var tlsPath = "TLS_PATH"
```
Wondertan, line 18: We should prefix this with `CELESTIA_`. The node running our app may have other TLS specific configurations and we should ensure they don't interfere.

nodebuilder/p2p/host.go
```diff
@@ -79,9 +91,14 @@ func host(params hostParams) (HostBase, error) {
 		libp2p.DisableRelay(),
 		libp2p.BandwidthReporter(params.Bandwidth),
 		libp2p.ResourceManager(params.ResourceManager),
+		libp2p.ChainOptions(
+			libp2p.Transport(tcp.NewTCPTransport),
+			libp2p.Transport(quic.NewTransport),
+			libp2p.Transport(webtransport.New),
+			wss,
+		),
```
Wondertan, line 99: Let's still keep using defaults and chain them with wss

state/core_access.go
```diff
@@ -273,6 +273,7 @@ func (ca *CoreAccessor) SubmitPayForBlob(
 		if response != nil && response.Code != 0 {
 			err = errors.Join(err, sdkErrors.ABCIError(response.Codespace, response.Code, response.Logs.String()))
 		}
+		fmt.Println("TX HASH ", response.TxHash)
```
renaynay, line 276: remove

state/core_access.go
renaynay: let's revert it yea

nodebuilder/p2p/host.go
vgonkivs: it can't be replaced: `transports already registered for protocol(s): ws, wss`

nodebuilder/p2p/tls.go
```diff
@@ -0,0 +1,51 @@
...
+	key  = "key.pem"
+)
+
+var tlsPath = "CELESTIA_TLS_PATH"
+
+// tlsEnabled checks whether `tlsPath` is not empty and creates a certificate.
+// it returns the cfg itself, the bool flag that specifies whether the config was created
+// and an error.
+func tlsEnabled() (*tls.Config, bool, error) {
+	path := os.Getenv(tlsPath)
```
Wondertan, line 25: check if the string is empty here earlier
and add a debug log if that's the case
this should help debugging and confirm is env var was set correctly or not

nodebuilder/p2p/tls.go
```diff
@@ -0,0 +1,51 @@
...
+		return nil, false, nil
+	}
+
+	var certificates []tls.Certificate
+	cert, err := tls.LoadX509KeyPair(certPath, keyPath)
+	if err != nil {
+		return nil, false, err
+	}
+	certificates = append(certificates, cert)
+	return &tls.Config{MinVersion: tls.VersionTLS12, Certificates: certificates}, true, nil
```
Wondertan, line 40: uber nit:
```suggestion
	cert, err := tls.LoadX509KeyPair(certPath, keyPath)
	if err != nil {
		return nil, false, err
	}
	
	return &tls.Config{
	        MinVersion: tls.VersionTLS12, 
	        Certificates: []tls.Certificate{cert},
	}, true, nil
```

Wondertan (approved):

cristaloleg (approved):

renaynay (approved):

PR #3561, opened 2024-07-11
misc(share/p2p): reduce frequency of discovery retries

### LINKED ISSUE #3115: ci(unit tests): discovery tests, flakey test "did not discover peer on time"
- TestDiscovery

flakey tests intermittently with race detector

```
2024/01/18 13:37:23 failed to sufficiently increase receive buffer size (was: 208 kiB, wanted: 2048 kiB, got: 416 kiB). See https://github.com/quic-go/quic-go/wiki/UDP-Buffer-Sizes for details.
--- FAIL: TestDiscovery (20.00s)
    discovery_test.go:67: did not discover peer in time
```

https://github.com/celestiaorg/celestia-node/actions/runs/7570583591/job/20616824070

### LINKED ISSUE #3550: feat(discovery): discover triggered too frequently
In the case a node doesn't have its quota of peers, it will send a new `discover` request [every second](https://github.com/celestiaorg/celestia-node/blob/accb0589d2de42a44de43e0974610d3a0186976d/share/p2p/discovery/discovery.go#L32-L40). There are no guarantees that the `discover` request can complete within 1 second.

https://github.com/celestiaorg/celestia-node/blob/accb0589d2de42a44de43e0974610d3a0186976d/share/p2p/discovery/discovery.go#L225-L229

This interval seems too aggressive and should probably be increased (e.g to `1` or even `5` minutes?)

ramin:
Fixes #3550 as suggested by [guillaumemichel](https://github.com/guillaumemichel). In adjusting the "1 liner" realized there was some relatively funky stuff in tests where we set/reset a package level var for the interval, so moved both AdvertiserRetryTimeout and DiscoveryRetryTimeout to parameters and allowed customization where NewDiscovery is called. Also adjusted some of the intervals to be faster in tests so they execute within the test context Timeout and thus, hopefully will fix #3115

### DIFF
--- a/share/p2p/discovery/discovery.go
+++ b/share/p2p/discovery/discovery.go
@@ -28,17 +28,10 @@ const (
 	// findPeersTimeout limits the FindPeers operation in time
 	findPeersTimeout = time.Minute
 
-	// advertiseRetryTimeout defines time interval between advertise attempts.
-	advertiseRetryTimeout = time.Second * 1
-
 	// logInterval defines the time interval at which a warning message will be logged
 	// if the desired number of nodes is not detected.
 	logInterval = 5 * time.Minute
 )
-
-// discoveryRetryTimeout defines time interval between discovery attempts
-// this is set independently for tests in discover_test.go
-var DiscoveryRetryTimeout = advertiseRetryTimeout * 60
 
 // Discovery combines advertise and discover services and allows to store discovered nodes.
 // TODO: The code here gets horribly hairy, so we should refactor this at some point
@@ -183,7 +176,7 @@ func (d *Discovery) Advertise(ctx context.Context) {
 			// we don't want retry indefinitely in busy loop
 			// internal discovery mechanism may need some time before attempts
 			select {
-			case <-time.After(advertiseRetryTimeout):
+			case <-time.After(d.params.AdvertiseRetryTimeout):
 				if !timer.Stop() {
 					<-timer.C
 				}
@@ -210,7 +203,7 @@ func (d *Discovery) Advertise(ctx context.Context) {
 // It initiates peer discovery upon request and restarts the process until the soft limit is
 // reached.
 func (d *Discovery) discoveryLoop(ctx context.Context) {
-	t := time.NewTicker(DiscoveryRetryTimeout)
+	t := time.NewTicker(d.params.DiscoveryRetryTimeout)
 	defer t.Stop()
 
 	warnTicker := time.NewTicker(logInterval)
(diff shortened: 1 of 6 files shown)


share/p2p/discovery/options.go
```diff
@@ -7,6 +7,11 @@ import (
 	"github.com/libp2p/go-libp2p/core/peer"
 )
 
+const (
+	defaultAdvertiseRetryTimeout = time.Second
+	defaultDiscoveryRetryTimeout = defaultAdvertiseRetryTimeout * 60
+)
```
walldiss, line 13: Let put into values of default params struct, to make it more consistent with other values and other default params we have.

share/p2p/discovery/discovery.go
```diff
@@ -181,16 +175,13 @@ func (d *Discovery) Advertise(ctx context.Context) {
 
 			// we don't want retry indefinitely in busy loop
 			// internal discovery mechanism may need some time before attempts
-			errTimer := time.NewTimer(retryTimeout)
 			select {
-			case <-errTimer.C:
-				errTimer.Stop()
+			case <-time.After(d.params.AdvertiseRetryTimeout):
```
renaynay, line 179: why do we do this instead of just setting a short-lived context with params.AdvertiseRetryTimeout for the call to `Advertise` ?

share/p2p/discovery/options.go
```diff
@@ -33,13 +40,33 @@ type Option func(*options)
...
+
 // Validate validates the values in Parameters
 func (p *Parameters) Validate() error {
+	if p.AdvertiseRetryTimeout <= 0 {
+		return fmt.Errorf("discovery: advertise retry timeout cannot be zero or negative")
+	}
+
+	if p.DiscoveryRetryTimeout <= 0 {
+		return fmt.Errorf("discovery: discovery retry timeout cannot be zero or negative")
+	}
```
renaynay, line 68: Isn't this config breaking? If so, let's mark PR as such.

share/p2p/discovery/discovery.go
Wondertan, line 179: Generally, `time.After` should be avoided as it's leaky. 
(It won't be after 1.23 tho)

walldiss: The code waits until next retry after `Advertise` returned. Short context does not affect it

Wondertan:
There is some wisdom behind the metaphor in ca67fef50f89af5cbbc1b4ebfebc40f733ddae92, but I am not getting it

ramin:
> There is some wisdom behind the metaphor in [ca67fef](https://github.com/celestiaorg/celestia-node/commit/ca67fef50f89af5cbbc1b4ebfebc40f733ddae92), but I am not getting it

@Wondertan ah, I put it back how it was, so no unnecessary change :-)

renaynay:
Let's hold off on this til we get conf from v0.14.1

PR #3567, opened 2024-07-15
fix(.github): change workflow to only work on releases

renaynay:
Fixes an issue where RCs are turned into releases after some time.

### DIFF
--- a/.github/workflows/ci_release.yml
+++ b/.github/workflows/ci_release.yml
@@ -5,7 +5,7 @@ on:
     branches:
       - main
   release:
-    types: [published]
+    types: [released]
   pull_request:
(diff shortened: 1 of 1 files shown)


ramin (approved):

ramin:
@renaynay i think we may be able to avoid this change and still be able to get binaries and artifacts while notting setting every release as current by altering the prerelesae-suffix to be `-rc*` or something, i can have a look 

eg: [-rc0](https://github.com/celestiaorg/celestia-node/releases/tag/v0.15.0-rc0)

renaynay:
Closing in favour of @ramin changing to ignore rc suffix

PR #3568, opened 2024-07-16
misc(ci): detect breaking change to any .proto files or config fields in nodebuilder/**/config.go and add kind:break!

### LINKED ISSUE #1918: ci: Action to add `kind:break!` label if config fields or proto files are touched in PR
https://github.com/marketplace/actions/changed-files

ramin:
deep cut: fixes #1918

If we change a .proto file, or add/modify a field in nodebuilder/**/config.go (specifically a struct) then add a label `kind:break!` on PR creation

hard to test but you can see some of mine in the commit history. As it stands i've left this on PR open ONLY for now so if it does misbehave, we don't get stuck with a break label on sync/update. We can expand scope later.


walldiss:
Proto changes are not neccessary breaking, as there might be backwards compatibility support. Perhaps the motivation here is that it is easier to remember to remove unwanted breaking tag, than add one?

renaynay (approved):
utack, i agree with this flow

vgonkivs (approved):

walldiss (approved):

PR #3569, opened 2024-07-16
fix(ci/release): golreleaser pre-release behavior

ramin:
Changes goreleaser.yml config to ensure that tags/releases are not automatically promoted to latest by adding the `release.prerelease = auto` config (previously defaulted to false), also set a proper pre-release suffix for tag filtering etc. This allows us to create signed binaries for pre-releases/release candidates.

Tested on a private branch and ensured that with  `release.prerelease: auto` releases created in github and with the toggle "Set as a pre-release" set to true, goreleaser would NOT set them as latest upon publishing artifacts.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -65,5 +65,7 @@ changelog:
     exclude:
       - "^docs:"
       - "^test:"
+release:
+  prerelease: auto
 git:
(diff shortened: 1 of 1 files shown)


renaynay (approved):
utack

Wondertan (approved):

PR #3580, opened 2024-07-19
feat(cmd): adding timeout flag

### LINKED ISSUE #3575: improvement(cmd): add timeout flag
Our current RPC client, created for the CLI, does not contain any timeouts. We can extend our basic flags with `--timeout`, allowing the user to specify the maximum duration for a request.

It should be added [here](https://github.com/celestiaorg/celestia-node/blob/79295fc5560a0b14d110e02f6a424446ecc58807/cmd/rpc.go#L23), so non-empty timeout could be used [here](https://github.com/celestiaorg/celestia-node/blob/79295fc5560a0b14d110e02f6a424446ecc58807/cmd/rpc.go#L76)

sebasti810:
fixes #3575


jcstein:
does this time out the transaction in the mempool, or just in the client?

sebasti810:
just in the client. is it supposed to also set it for the mempool?

jcstein:
I'm thinking of a case where I send transaction with nonce x, it times out, and then when I go to send another transaction, also with nonce x, it errors because it has the same nonce as the tx that is already timed out on client, but in the mempool

vgonkivs:
But node does not support nonce so far. We agreed that's it is not necessary.

distractedm1nd:
> I'm thinking of a case where I send transaction with nonce x, it times out, and then when I go to send another transaction, also with nonce x, it errors because it has the same nonce as the tx that is already timed out on client, but in the mempool

This does not happen anymore - the nonce error was fixed by upgrading app :)

> But node does not support nonce so far. We agreed that's it is not necessary.

I don't think he's saying we need to support nonce, I think @jcstein was just describing that what happens with a nonce in the background was throwing errors to the user, and that a timeout for txs in the mempool could fix this. 

What if we add a timeout to `TxConfig` in a separate issue?

vgonkivs:
I'd rather pass it through the context.

distractedm1nd:
Fair, I guess we can't have a timeout in the transaction itself that gets submitted, right?

vgonkivs:
I don't want to grow the TxConfig. Timeout is smth we can configure through the context.
wdyt @distractedm1nd ?

jcstein:
> This does not happen anymore - the nonce error was fixed by upgrading app :)
😄 🚢

renaynay:
@jcstein Did you test this PR already?

renaynay (approved):

cristaloleg (approved):

walldiss (approved):

cmd/rpc.go
```diff
@@ -37,6 +39,13 @@ func RPCFlags() *flag.FlagSet {
 		"Authorization token",
 	)
 
+	fset.DurationVar(
+		&timeoutFlag,
+		"timeout",
+		0,
+		"Timeout for RPC requests (e.g. 30s, 1m)",
```
vgonkivs, line 46: Could you describe how the user should interact with the flag? In what units the timeout should be passed, bc duration represents time in nanoseconds.?

vgonkivs (changes requested):
Could you describe how the user should interact with the flag? In what units the timeout should be passed, bc duration represents time in nanoseconds.?

sebasti810:
@vgonkivs The flag docstring prints the usage, e.g. 30s, 1m. This is because pflag has a DurationVar option, which handles this parsing underneath. 

And it's tested, this is the output you get when the timeline is exceeded
image

jcstein:
> 
@jcstein Did you test this PR already?

@renaynay just saw this, no I haven't tested it yet

cmd/rpc.go
```diff
@@ -73,7 +82,12 @@ func InitClient(cmd *cobra.Command, _ []string) error {
 		}
 	}
 
-	client, err := rpc.NewClient(cmd.Context(), requestURL, authTokenFlag)
+	var opts []rpc.RPCClientOption
+	if timeoutFlag > 0 {
+		opts = append(opts, rpc.WithTimeout(timeoutFlag))
+	}
+
+	client, err := rpc.NewClient(cmd.Context(), requestURL, authTokenFlag, opts...)
```
Wondertan, line 90: There is a simpler way to achieve the same result.

We can pass through the timeout in the context. The timeout is respected by the RPC HTTP client performing the request, which avoids creating a new option for the client.

Wondertan (changes requested):
Hey @sebasti810! 

Thank you for the contribution

renaynay (dismissed):

Wondertan (dismissed):

cristaloleg (approved):

cmd/rpc.go
```diff
@@ -73,6 +82,12 @@ func InitClient(cmd *cobra.Command, _ []string) error {
 		}
 	}
 
+	if timeoutFlag > 0 {
+		// we don't cancel, because we want to keep this context alive outside the InitClient Function
+		ctx, _ := context.WithTimeout(cmd.Context(), timeoutFlag)
```
cristaloleg, line 87: Can you add `//nolint:govet` here please?

cristaloleg (approved):

walldiss (approved):

PR #3581, opened 2024-07-19
chore: bump libp2p

Wondertan:
There are quite a few important bug fixes in libp2p that cause leaks in various edge cases and have been causing issues for other networks, so we should use them as well.

https://github.com/libp2p/go-libp2p/releases


walldiss (approved):

cristaloleg (approved):

renaynay:
why so many integration tests failing?

Wondertan:
It was a flake. Tests failed with "bind: already in use" errors. Works now

PR #3583, opened 2024-07-22
refactor(shwap):  rename share.root to axisRoots

walldiss:
- rename share.Root to share.AxisRoots. The value contains set of RowRoots and ColRoots, which are easier to refer as axisRoots, as root might be subjective to context.
- rename empty Root to empty EDS. Adjusting to refer empty content of block, rather than root.
- rename dah to roots where appropriate
- replace da.NewAvailabilityHeader with share.NewAxisRoots where appropriate

### DIFF
--- a/share/empty.go
+++ b/share/empty.go
@@ -11,14 +11,14 @@ import (
 	"github.com/celestiaorg/rsmt2d"
 )
 
-// EmptyRoot returns Root of the empty block EDS.
-func EmptyRoot() *Root {
+// EmptyEDSRoot returns AxisRoots of the empty block EDS.
+func EmptyEDSRoot() *AxisRoots {
 	initEmpty()
 	return emptyBlockRoot
 }
 
-// EmptyExtendedDataSquare returns the EDS of the empty block data square.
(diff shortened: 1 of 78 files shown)


store/store.go
```diff
@@ -361,20 +361,20 @@ func pathExists(path string) (bool, error) {
 }
 
 func ensureEmptyFile(basepath string) error {
-	path := basepath + blocksPath + share.DataHash(share.EmptyRoot().Hash()).String()
+	path := basepath + blocksPath + share.DataHash(share.EmptyEDSRoot().Hash()).String()
```
cristaloleg, line 364: Nit: `filepath.Join`

share/empty.go
```diff
@@ -11,14 +11,14 @@ import (
 	"github.com/celestiaorg/rsmt2d"
 )
 
-// EmptyRoot returns Root of the empty block EDS.
-func EmptyRoot() *Root {
+// EmptyEDSRoot returns AxisRoots of the empty block EDS.
+func EmptyEDSRoot() *AxisRoots {
```
Wondertan, line 15: Make it plural - `EmptyEDSRoots`

PR #3584, opened 2024-07-22
feat(shwap/p2p): Bitswap Getter

Wondertan:
Implements new `Getter` over Bitswap p2p composition.

This PR does not intend to implement unit tests for the Getter itself. Adding them here is not trivial as we don't have the `Getter` test suite, which will soon be replaced with the `Accessor` interface. The current coverage of every `Block` is sufficient to prove every method works while `Getter` itself is simply an adapter. We need this `Getter` _temporary_ to accelerate testing of `Shwap` and replace it with an `Accesor` once we allocate time for refactorings.

### DIFF
--- a/share/shwap/p2p/bitswap/getter.go
+++ b/share/shwap/p2p/bitswap/getter.go
@@ -1,0 +1,172 @@
+package bitswap
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/ipfs/boxo/blockstore"
+	"github.com/ipfs/boxo/exchange"
+
+	"github.com/celestiaorg/celestia-app/pkg/wrapper"
+	"github.com/celestiaorg/rsmt2d"
+
+	"github.com/celestiaorg/celestia-node/header"
+	"github.com/celestiaorg/celestia-node/share"
+)
+
+// Getter implements share.Getter.
+type Getter struct {
+	exchange exchange.SessionExchange
+	bstore   blockstore.Blockstore
+	session  exchange.Fetcher
+	cancel   context.CancelFunc
+}
+
+// NewGetter constructs a new Getter.
+func NewGetter(exchange exchange.SessionExchange, bstore blockstore.Blockstore) *Getter {
+	return &Getter{exchange: exchange, bstore: bstore}
+}
+
+// Start kicks off internal fetching session.
+func (g *Getter) Start() {
+	ctx, cancel := context.WithCancel(context.Background())
+	g.session = g.exchange.NewSession(ctx)
+	g.cancel = cancel
+}
+
+// Stop shuts down Getter's internal fetching session.
+func (g *Getter) Stop() {
+	g.cancel()
+}
+
+// TODO(@Wondertan): Rework API to get coordinates as a single param to make it ergonomic.
+func (g *Getter) GetShares(
+	ctx context.Context,
+	hdr *header.ExtendedHeader,
+	rowIdxs, colIdxs []int,
+) ([]share.Share, error) {
+	if len(rowIdxs) != len(colIdxs) {
+		return nil, fmt.Errorf("row indecies and col indecies must be same length")
+	}
+
+	if len(rowIdxs) == 0 {
+		return nil, fmt.Errorf("empty coordinates")
+	}
+
+	blks := make([]Block, 0, len(rowIdxs))
+	for i, rowIdx := range rowIdxs {
+		sid, err := NewEmptySampleBlock(hdr.Height(), rowIdx, colIdxs[i], len(hdr.DAH.RowRoots))
+		if err != nil {
+			return nil, err
+		}
+
+		blks[i] = sid
+	}
+
+	err := Fetch(ctx, g.exchange, hdr.DAH, blks, WithStore(g.bstore))
+	if err != nil {
+		return nil, err
+	}
+
+	shares := make([]share.Share, len(blks))
+	for i, blk := range blks {
+		shares[i] = blk.(*SampleBlock).Container.Share
+	}
+
+	return shares, nil
+}
+
+func (g *Getter) GetShare(
+	ctx context.Context,
+	hdr *header.ExtendedHeader,
+	row, col int,
+) (share.Share, error) {
+	shrs, err := g.GetShares(ctx, hdr, []int{row}, []int{col})
+	if err != nil {
+		return nil, err
+	}
+
+	if len(shrs) != 1 {
+		return nil, fmt.Errorf("expected 1 share row, got %d", len(shrs))
+	}
+
+	return shrs[0], nil
+}
+
+func (g *Getter) GetEDS(
+	ctx context.Context,
+	hdr *header.ExtendedHeader,
+) (*rsmt2d.ExtendedDataSquare, error) {
+	sqrLn := len(hdr.DAH.RowRoots)
+	blks := make([]Block, sqrLn/2)
+	for i := 0; i < sqrLn/2; i++ {
+		blk, err := NewEmptyRowBlock(hdr.Height(), i, sqrLn)
+		if err != nil {
+			return nil, err
+		}
+
+		blks[i] = blk
+	}
+
+	err := Fetch(ctx, g.exchange, hdr.DAH, blks, WithFetcher(g.session))
+	if err != nil {
+		return nil, err
+	}
+
+	shrs := make([]share.Share, 0, sqrLn*sqrLn)
+	for _, row := range blks {
+		rowShrs, err := row.(*RowBlock).Container.Shares()
+		if err != nil {
+			return nil, fmt.Errorf("decoding Shares out of Row: %w", err)
+		}
+		shrs = append(shrs, rowShrs...)
+	}
+
+	square, err := rsmt2d.ComputeExtendedDataSquare(
+		shrs,
+		share.DefaultRSMT2DCodec(),
+		wrapper.NewConstructor(uint64(sqrLn/2)),
+	)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


cristaloleg (commented):
SGTM, maybe method docs?

share/shwap/p2p/bitswap/getter.go
```diff
@@ -0,0 +1,172 @@
...
+func NewGetter(exchange exchange.SessionExchange, bstore blockstore.Blockstore) *Getter {
+	return &Getter{exchange: exchange, bstore: bstore}
+}
+
+// Start kicks off internal fetching session.
+func (g *Getter) Start() {
+	ctx, cancel := context.WithCancel(context.Background())
+	g.session = g.exchange.NewSession(ctx)
+	g.cancel = cancel
+}
```
walldiss, line 35: Is it application lifetime specific session or session is expected per request?

share/shwap/p2p/bitswap/getter.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	blks := make([]Block, 0, len(rowIdxs))
+	for i, rowIdx := range rowIdxs {
+		sid, err := NewEmptySampleBlock(hdr.Height(), rowIdx, colIdxs[i], len(hdr.DAH.RowRoots))
+		if err != nil {
+			return nil, err
+		}
+
+		blks[i] = sid
```
walldiss, line 63: Should be append. Otherwise will panic on write to `blks` slice of len 0.

share/shwap/p2p/bitswap/getter.go
```diff
@@ -0,0 +1,172 @@
...
+// NewGetter constructs a new Getter.
+func NewGetter(exchange exchange.SessionExchange, bstore blockstore.Blockstore) *Getter {
+	return &Getter{exchange: exchange, bstore: bstore}
+}
+
+// Start kicks off internal fetching session.
+func (g *Getter) Start() {
+	ctx, cancel := context.WithCancel(context.Background())
+	g.session = g.exchange.NewSession(ctx)
+	g.cancel = cancel
```
walldiss, line 34: Would it be possible for Fetch to read from blockstore before requesting from network? It seems like it will always network exchange as fetcher here, when we would want to have local lookup from blockstore first.

share/shwap/p2p/bitswap/getter.go
```diff
@@ -0,0 +1,172 @@
...
+		return nil, err
+	}
+
+	shrs := make([]share.Share, 0, sqrLn*sqrLn)
+	for _, row := range blks {
+		rowShrs, err := row.(*RowBlock).Container.Shares()
+		if err != nil {
+			return nil, fmt.Errorf("decoding Shares out of Row: %w", err)
+		}
+		shrs = append(shrs, rowShrs...)
```
walldiss, line 122: Requests are sent only for ods part of the square, so should prealloc shares only for `sqrLn*sqrLn/2`

share/shwap/p2p/bitswap/getter.go
Wondertan: "session for life"

share/shwap/p2p/bitswap/getter.go
Wondertan: "This PR does not intend to implement any unit tests." 💀

share/shwap/p2p/bitswap/getter.go
Wondertan: I don't see when we would want that. We have store getter for that

share/shwap/p2p/bitswap/getter.go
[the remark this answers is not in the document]
Wondertan: There is no way to get them without decoding rn. It's fine for now to return them the way as they aren't gonna be decoded again during Compute.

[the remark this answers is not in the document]
Wondertan: Besides, they are going to be decoded already during the verification once we implement caching

Wondertan: `sqrLn/2*sqrLn/2`, but yes

share/shwap/p2p/bitswap/getter.go
walldiss: After failed reconstruction, next one (retry) should be able to read samples from blockstore that prevoius attempt has saved.

share/shwap/p2p/bitswap/getter.go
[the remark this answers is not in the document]
walldiss: I mean, you supply too many shares to EDSRecompute. It expects only original shares, but you also add parity ones to shares slice

share/shwap/p2p/bitswap/getter.go
Wondertan: This should be part of `Retriever` logic. This way we avoid hitting store(and disk) for when we don't need it

share/shwap/p2p/bitswap/getter.go
Wondertan: I think we talked about it. The session retains information about peers, their response time, etc, and involves some expensive bootstrapping that we can avoid. Agree that there should be a comment.

share/shwap/p2p/bitswap/getter.go
walldiss: Do I understand it that the blockstore provided to Fetch will be used for storing samples by Lights Nodes? Perhaps we can also store samples onside of Getter for LN. For example in Availability, same way Full nodes do it outside of getter

share/shwap/p2p/bitswap/getter.go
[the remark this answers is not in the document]
Wondertan: I thought one could supply EDS to compute, but maybe you are right. I will make a test for that.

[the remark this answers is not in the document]
Wondertan: If that's the case, we should use the `NewExtendedSquare` with the `Repair` method so we don't have to recompute twice.

share/shwap/p2p/bitswap/getter.go
Wondertan: This is likely to happen eventually

share/shwap/p2p/bitswap/getter.go
Wondertan: Another thing I realized. We need to split sessions for archival data and available data. These are two different sets of peers serving different data, so we should not mix them up

share/shwap/p2p/bitswap/getter.go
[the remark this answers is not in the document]
Wondertan: `NewExtendedSquare` doesn't work as it doesn't allow passing share slice, but it works with `ImportEDS`

share/shwap/p2p/bitswap/getter.go
```diff
@@ -0,0 +1,214 @@
...
+
+// GetShares uses [SampleBlock] and [Fetch] to get and verify samples for given coordinates.
+// TODO(@Wondertan): Rework API to get coordinates as a single param to make it ergonomic.
+func (g *Getter) GetShares(
+	ctx context.Context,
+	hdr *header.ExtendedHeader,
+	rowIdxs, colIdxs []int,
+) ([]share.Share, error) {
+	if len(rowIdxs) != len(colIdxs) {
+		return nil, fmt.Errorf("row indecies and col indecies must be same length")
```
ramin, line 68: nit typo: indecies -> indices

share/shwap/p2p/bitswap/getter.go
```diff
@@ -0,0 +1,214 @@
...
+	shrs := make([]share.Share, 0, sqrLn/2*sqrLn/2)
+	for i, row := range blks {
+		rowShrs, err := row.(*RowBlock).Container.Shares()
+		if err != nil {
+			return nil, fmt.Errorf("decoding Shares out of Row: %w", err)
+		}
+
+		for j, shr := range rowShrs {
+			shrs[i*j] = shr
+		}
```
walldiss, line 151: You still loop over full sized row, while trying to set shares. Only original part of row should be used here for recompute.

share/shwap/p2p/bitswap/getter.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	square, err := rsmt2d.ImportExtendedDataSquare(
+		shrs,
+		share.DefaultRSMT2DCodec(),
+		wrapper.NewConstructor(uint64(sqrLn/2)),
+	)
```
walldiss, line 158: Import would need all EDS shares, and you only requested half of rows. Consider using `ComputeExtendedDataSquare`

share/shwap/p2p/bitswap/getter.go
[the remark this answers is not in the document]
walldiss: You need to supply ODS shares to Compute. While you pass Q1+Q2

share/shwap/p2p/bitswap/getter.go
Wondertan: Import works and that's tested. Compute expects only ods, while import can take empty shares

share/shwap/p2p/bitswap/getter.go
Wondertan: Thats the goal. We set full rows that were computed previously during verification.

share/shwap/p2p/bitswap/getter.go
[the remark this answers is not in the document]
Wondertan: No, i pass q1+q2  purposefull and to impory instead of compute

walldiss (approved):
utAck

PR #3586, opened 2024-07-22
feat(shwap): Add axis roots to Accessor

walldiss:
Add axis roots to all Accessor implementations. It will allow server side to handle namespace requests without the need to know exact rowIdxes beforehand.

### DIFF
--- a/share/eds/edstest/testing.go
+++ b/share/eds/edstest/testing.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/stretchr/testify/require"
 
-	"github.com/celestiaorg/celestia-app/pkg/da"
 	"github.com/celestiaorg/celestia-app/pkg/wrapper"
 	"github.com/celestiaorg/nmt"
 	"github.com/celestiaorg/rsmt2d"
@@ -50,6 +49,7 @@ func RandEDSWithNamespace(
 	return eds, roots
 }
 
+// RandomAxisRoots generates random share.AxisRoots for the given eds size.
 func RandomAxisRoots(t testing.TB, size int) *share.AxisRoots {
 	roots := make([][]byte, size*2)
 	for i := range roots {
@@ -61,7 +61,7 @@ func RandomAxisRoots(t testing.TB, size int) *share.AxisRoots {
 
 	rows := roots[:size]
 	cols := roots[size:]
-	return &da.DataAvailabilityHeader{
+	return &share.AxisRoots{
 		RowRoots:    rows,
 		ColumnRoots: cols,
 	}
--- a/share/eds/edstest/testing.go
+++ b/share/eds/edstest/testing.go
@@ -1,10 +1,12 @@
 package edstest
 
 import (
+	"crypto/rand"
 	"testing"
 
 	"github.com/stretchr/testify/require"
 
+	"github.com/celestiaorg/celestia-app/pkg/da"
 	"github.com/celestiaorg/celestia-app/pkg/wrapper"
 	"github.com/celestiaorg/nmt"
 	"github.com/celestiaorg/rsmt2d"
@@ -46,4 +48,21 @@ func RandEDSWithNamespace(
 	roots, err := share.NewAxisRoots(eds)
 	require.NoError(t, err)
 	return eds, roots
+}
+
+func RandomAxisRoots(t testing.TB, size int) *share.AxisRoots {
+	roots := make([][]byte, size*2)
+	for i := range roots {
+		root := make([]byte, size)
+		_, err := rand.Read(root)
+		require.NoError(t, err)
+		roots[i] = root
+	}
+
+	rows := roots[:size]
+	cols := roots[size:]
+	return &da.DataAvailabilityHeader{
+		RowRoots:    rows,
+		ColumnRoots: cols,
+	}
 }
--- a/share/eds/edstest/testing.go
+++ b/share/eds/edstest/testing.go
@@ -50,17 +50,17 @@ func RandEDSWithNamespace(
 }
 
 // RandomAxisRoots generates random share.AxisRoots for the given eds size.
-func RandomAxisRoots(t testing.TB, size int) *share.AxisRoots {
-	roots := make([][]byte, size*2)
+func RandomAxisRoots(t testing.TB, edsSize int) *share.AxisRoots {
+	roots := make([][]byte, edsSize*2)
 	for i := range roots {
-		root := make([]byte, size)
+		root := make([]byte, edsSize)
 		_, err := rand.Read(root)
 		require.NoError(t, err)
 		roots[i] = root
 	}
 
-	rows := roots[:size]
-	cols := roots[size:]
+	rows := roots[:edsSize]
+	cols := roots[edsSize:]
 	return &share.AxisRoots{
 		RowRoots:    rows,
 		ColumnRoots: cols,
--- a/share/new_eds/accessor.go
+++ b/share/new_eds/accessor.go
@@ -16,7 +16,7 @@ type Accessor interface {
 	Size(ctx context.Context) int
 	// DataHash returns data hash of the Accessor.
 	DataHash(ctx context.Context) (share.DataHash, error)
-	// AxisRoots returns axis roots of the Accessor.
+	// AxisRoots returns share.AxisRoots (DataAvailabilityHeader) of the Accessor.
 	AxisRoots(ctx context.Context) (*share.AxisRoots, error)
 	// Sample returns share and corresponding proof for row and column indices. Implementation can
 	// choose which axis to use for proof. Chosen axis for proof should be indicated in the returned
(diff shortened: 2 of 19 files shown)


share/new_eds/close_once.go
```diff
@@ -49,6 +49,14 @@ func (c *closeOnce) DataHash(ctx context.Context) (share.DataHash, error) {
 	return c.f.DataHash(ctx)
 }
 
+// AxisRoots returns AxisRoots of Accessor's underlying EDS.
```
Wondertan, line 52: No need to duplicate the comment from the interface, especially when no other method has comment.

share/new_eds/accessor.go
```diff
@@ -16,6 +16,8 @@ type Accessor interface {
 	Size(ctx context.Context) int
 	// DataHash returns data hash of the Accessor.
 	DataHash(ctx context.Context) (share.DataHash, error)
+	// AxisRoots returns axis roots of the Accessor.
```
Wondertan, line 19: Worth adding a little comment that's also called DAH. This will be useful for people that don't have enough context on the new name.

share/new_eds/nd.go
```diff
@@ -13,13 +13,15 @@ import (
...
 func NamespacedData(
 	ctx context.Context,
-	root *share.AxisRoots,
 	eds Accessor,
 	namespace share.Namespace,
 ) (shwap.NamespacedData, error) {
-	rowIdxs := share.RowsWithNamespace(root, namespace)
+	roots, err := eds.AxisRoots(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get root: %w", err)
```
Wondertan, line 21:
```suggestion
		return nil, fmt.Errorf("failed to get AxisRoots: %w", err)
```

share/new_eds/proofs_cache.go
```diff
@@ -81,6 +82,25 @@ func (c *proofsCache) DataHash(ctx context.Context) (share.DataHash, error) {
...
+		return roots, nil
+	}
+
+	// if roots are not in cache, read them from the inner accessor
+	roots, err := c.inner.AxisRoots(ctx)
+	if err != nil {
+		return nil, err
+	}
+	c.lock.Lock()
+	defer c.lock.Unlock()
```
Wondertan, line 99: uber nit: no need for defer here

share/new_eds/testing.go
```diff
@@ -64,6 +72,40 @@ func TestStreamer(
 	})
 }
 
+func testAccessorDataRoot(
```
Wondertan, line 75: DataHash

share/new_eds/testing.go
```diff
@@ -35,6 +35,14 @@ func TestSuiteAccessor(
 		t.Errorf("minSize must be power of 2: %v", maxSize)
 	}
 	for size := minSize; size <= maxSize; size *= 2 {
+		t.Run(fmt.Sprintf("DataRoot:%d", size), func(t *testing.T) {
```
Wondertan, line 38: DataHash

share/root.go
```diff
@@ -11,6 +11,13 @@ import (
 	"github.com/celestiaorg/rsmt2d"
 )
 
+const (
+	// DataHashSize is the size of the DataHash.
+	DataHashSize = 32
+	// AxisRootsSize is the size of the single root in AxisRoots.
+	AxisRootsSize = 90
```
Wondertan, line 18:
```suggestion
	// AxisRootSize is the size of the single root in AxisRoots.
	AxisRootSize = 90
```

store/store.go
```diff
@@ -97,22 +97,23 @@ func (s *Store) Close() error {
...
 	tNow := time.Now()
+	datahash := share.DataHash(roots.Hash())
 	lock := s.stripLock.byDatahashAndHeight(datahash, height)
 	lock.lock()
 	defer lock.unlock()
 
-	path := filepath.Join(s.basepath, blocksPath, datahash.String())
 	if datahash.IsEmptyEDS() {
+		path := filepath.Join(s.basepath, blocksPath, roots.String())
 		err := s.ensureHeightLink(path, height)
```
Wondertan, line 112: would put path construction inside as well

store/store.go
```diff
@@ -134,12 +135,12 @@ func (s *Store) Put(
...
 func (s *Store) createFile(
-	path string,
-	datahash share.DataHash,
-	height uint64,
 	square *rsmt2d.ExtendedDataSquare,
+	roots *share.AxisRoots,
+	height uint64,
 ) (exists bool, err error) {
-	f, err := file.CreateQ1Q4File(path, datahash, square)
+	path := filepath.Join(s.basepath, blocksPath, roots.String())
```
Wondertan, line 142: once again suggest to hide path creation behind a method

Wondertan (commented):
first look, still need to go over new odsfile logic

share/eds/edstest/testing.go
```diff
@@ -47,3 +49,20 @@ func RandEDSWithNamespace(
...
+		require.NoError(t, err)
+		roots[i] = root
+	}
+
+	rows := roots[:size]
+	cols := roots[size:]
+	return &da.DataAvailabilityHeader{
+		RowRoots:    rows,
+		ColumnRoots: cols,
+	}
```
Wondertan, line 67: nit: use share.AxisRoots directly to avoid confusion and import

share/eds/edstest/testing.go
```diff
@@ -47,3 +49,20 @@ func RandEDSWithNamespace(
 	require.NoError(t, err)
 	return eds, roots
 }
+
+func RandomAxisRoots(t testing.TB, size int) *share.AxisRoots {
```
Wondertan, line 53: godoc

store/file/ods.go
```diff
@@ -243,6 +284,14 @@ func (f *ODSFile) readAxisHalf(axisType rsmt2d.Axis, axisIdx int) (eds.AxisHalf,
 	return eds.AxisHalf{}, fmt.Errorf("unknown axis")
 }
 
+func (f *ODSFile) sharesOffset() int {
+	return f.hdr.Size() + f.axisRootsSize()
+}
+
+func (f *ODSFile) axisRootsSize() int {
+	return share.AxisRootSize * 2 * f.size()
```
Wondertan, line 292: I would add where 2 is coming from.
```suggestion
	return share.AxisRootSize * 2 * f.size() // there are two Row and Col axises; thus we multiply by 2
```

store/file/ods.go
```diff
@@ -125,6 +142,29 @@ func (f *ODSFile) DataHash(context.Context) (share.DataHash, error) {
 	return f.hdr.datahash, nil
 }
 
+// AxisRoots returns axis roots of the Accessor.
```
Wondertan, line 145:
```suggestion
// AxisRoots reads AxisRoots stored in the file.
// AxisRoots are stored in between the header and actual Share data.
```

store/file/ods.go
```diff
@@ -97,15 +89,40 @@ func CreateODSFile(
...
+			return nil, fmt.Errorf("writing shares: %w", err)
+		}
+	}
+	return h, nil
+}
+
+func writeAxisRoots(w io.Writer, roots *share.AxisRoots) error {
+	for _, roots := range [][][]byte{roots.RowRoots, roots.ColumnRoots} {
+		for _, root := range roots {
+			if _, err := w.Write(root); err != nil {
```
Wondertan, line 123: Do we buffer writes? Feels like a lot of independent writes

store/file/ods.go
```diff
@@ -97,15 +89,40 @@ func CreateODSFile(
...
+		squareSize:  uint16(eds.Width()),
+		datahash:    axisRoots.Hash(),
+	}
 	err := writeHeader(w, h)
 	if err != nil {
-		return err
+		return nil, fmt.Errorf("writing header: %w", err)
+	}
+
+	// write axisRoots
```
Wondertan, line 106: helpful :)

share/new_eds/proofs_cache.go
```diff
@@ -29,8 +29,9 @@ var _ AccessorStreamer = (*proofsCache)(nil)
 type proofsCache struct {
 	inner AccessorStreamer
 
-	// lock protects axisCache
-	lock sync.RWMutex
+	// lock protects proofCache
+	lock       sync.RWMutex
```
Wondertan, line 33: Why share the lock between multiple _independent_ fields, instead of using separate for each?

walldiss: Not really a big problem, because it is blasingly fast lock and very rare. It locks on setting pointer of AxisRoots, which will happen once

store/file/ods.go
walldiss: I've tried buffering write of axisRoots there is barely any benefit. But this idea made me also try to buffering of shares when writing quadrants. It improved Put time by 3x. Will add in another PR.

share/new_eds/proofs_cache.go
Wondertan: I generally prefer to separate locks when possible. It is also better for readability.

Wondertan (approved):

PR #3453, opened 2024-05-31
chore(deps): upgrade to celestia-app v2.0.0

rootulp:
Update celestia-node to use celestia-app v2.0.0. A few changes that were needed:
- celestia-app v1.x had a shares package. celestia-app v2.x uses the shares package from go-square.
- celestia-app v1.x had a blob.types package with `CreateCommitment`. celestia-app v2.x uses `CreateCommitment` from the go-square inclusion package.
- I had to update extended header verification to allow `header.Version.App = 2`. Added unit tests.
- celestia-app v1.x had a lot of functionality included in the `signer`. celestia-app v2.x split a `txClient` from the `signer`. See: https://github.com/celestiaorg/celestia-app/pull/3433
- ~~I had to update `core_access.go` a lot. Mostly inspired by https://github.com/celestiaorg/celestia-node/pull/3451~~

## Testing

I ran a [script](https://gist.github.com/rootulp/73ee382b4d533cb9da27fc675e9047c0) with: celestia-app v2.0.0-rc2  and configured it to upgrade at block height 3. celestia-node (built from this PR) continued to work:

```
2024-07-09T18:13:27.040-0400	INFO	header/store	store/store.go:367	new head	{"height": 2, "hash": "8776AEAF4114BD7E88E8DEC38445720D0BD857335BED99649957A43BB845EC87"}
2024-07-09T18:13:38.065-0400	INFO	header/store	store/store.go:367	new head	{"height": 3, "hash": "63D5C64521A964290BD21658314DDF60146AE419FE99026003048F74D2886B35"}
2024-07-09T18:13:49.093-0400	INFO	header/store	store/store.go:367	new head	{"height": 4, "hash": "FC7900918E716697A7CD6D9A4865B261F3E71D181F366E765AA53CF475223F9A"}
```

PR #3589, opened 2024-07-23
refactor(header): `Validate` should check that the raw header's app version is not greater than latest app version

renaynay:
In preparation for V2, we should change the header validation logic to ensure that we support validation for V2 and below.

### DIFF
--- a/blob/commitment_proof.go
+++ b/blob/commitment_proof.go
@@ -4,9 +4,9 @@ import (
 	"bytes"
 	"fmt"
 
-	"github.com/celestiaorg/celestia-app/pkg/appconsts"
-	"github.com/celestiaorg/celestia-app/pkg/shares"
+	"github.com/celestiaorg/celestia-app/v2/pkg/appconsts"
 	"github.com/celestiaorg/celestia-app/v2/pkg/proof"
+	"github.com/celestiaorg/go-square/inclusion"
 	"github.com/celestiaorg/nmt"
 	"github.com/celestiaorg/nmt/namespace"
 
@@ -98,7 +98,7 @@ func (commitmentProof *CommitmentProof) Verify(root []byte, subtreeRootThreshold
 	// the subtree roots width is defined in ADR-013:
 	//
 	//https://github.com/celestiaorg/celestia-app/blob/main/docs/architecture/adr-013-non-interactive-default-rules-for-zero-padding.md
-	subtreeRootsWidth := shares.SubTreeWidth(numberOfShares, subtreeRootThreshold)
+	subtreeRootsWidth := inclusion.SubTreeWidth(numberOfShares, subtreeRootThreshold)
 
 	// verify the proof of the subtree roots
 	subtreeRootsCursor := 0
--- a/header/header.go
+++ b/header/header.go
@@ -11,7 +11,7 @@ import (
 	"github.com/tendermint/tendermint/light"
 	core "github.com/tendermint/tendermint/types"
 
-	"github.com/celestiaorg/celestia-app/pkg/appconsts"
+	"github.com/celestiaorg/celestia-app/v2/pkg/appconsts"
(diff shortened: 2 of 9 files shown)


renaynay (commented):
feel like we should also ensure app version is not 0 since that's not valid

header/header.go
```diff
@@ -114,9 +114,9 @@ func (eh *ExtendedHeader) Validate() error {
 		return fmt.Errorf("ValidateBasic error on RawHeader at height %d: %w", eh.Height(), err)
 	}
 
-	if eh.RawHeader.Version.App != appconsts.LatestVersion {
-		return fmt.Errorf("app version mismatch, expected: %d, got %d", appconsts.LatestVersion,
-			eh.RawHeader.Version.App)
+	if eh.RawHeader.Version.App > appconsts.LatestVersion {
+		return fmt.Errorf("app version must be less than the latest app version %d, got %d",
+			appconsts.LatestVersion, eh.RawHeader.Version.App)
```
cmwaters, line 119:
```suggestion
		return fmt.Errorf("header received at height %d has version %d, this node supports up to version %d. Please upgrade to support new version",
			eh.RawHeader.Height, eh.RawHeader.Version.App, appconsts.LatestVersion)
```

nit: I would have something a little more descriptive as this will likely get hit by all nodes that don't upgrade to v3 when the consensus network does

cmwaters (approved):

vgonkivs (approved):

cristaloleg (approved):

cristaloleg (dismissed):

cmwaters (approved):

Wondertan (dismissed):

walldiss (approved):

header/header.go
```diff
@@ -115,13 +113,10 @@ func (eh *ExtendedHeader) Validate() error {
...
 	}
 
-	if eh.RawHeader.Version.App != v1.Version && eh.RawHeader.Version.App != v2.Version {
-		return fmt.Errorf(
-			"app version mismatch, expected: %d or %d, got %d",
-			v1.Version,
-			v2.Version,
-			eh.RawHeader.Version.App,
-		)
+	if eh.RawHeader.Version.App > appconsts.LatestVersion || eh.RawHeader.Version.App == 0 {
```
cristaloleg, line 116: looks like `"github.com/celestiaorg/celestia-app/pkg/appconsts"` isn't imported

cristaloleg: Nit: keeping `eh.RawHeader.Version.App == 0` at the beginning will make `if` looks a bit simpler.

cristaloleg (dismissed):

vgonkivs (dismissed):

blob/commitment_proof.go
```diff
@@ -4,9 +4,9 @@ import (
 	"bytes"
 	"fmt"
 
-	"github.com/celestiaorg/celestia-app/pkg/appconsts"
-	"github.com/celestiaorg/celestia-app/pkg/shares"
```
rootulp, line 8: [note for reviewers] I'm not sure how this slipped through https://github.com/celestiaorg/celestia-node/pull/3453. Maybe because this file was created after that PR first started. Regardless, this repo should no longer import from `github.com/celestiaorg/celestia-app` and should instead import from `github.com/celestiaorg/celestia-app/v2`.

rootulp:
Fraud integration tests failed on CI but pass locally so I'm assuming it's a flake:

```
$ make test-integration TAGS=fraud
--> Running integrations tests  -tags=fraud
ok  	github.com/celestiaorg/celestia-node/nodebuilder/tests	64.643s
```

Wondertan (approved):

rootulp:
wow fully green CI 🟢

cristaloleg (approved):
Nice catch with v2!

PR #3591, opened 2024-07-24
fix(share/getters): set minTimeout

vgonkivs:
Set minTimeout in case the context hasn't contained it.


share/getters/cascade.go
```diff
@@ -119,13 +120,21 @@ func cascadeGetters[V any](
 		}
 	}()
 
+	minTimeout := time.Duration(0)
+	_, ok := ctx.Deadline()
+	if !ok {
+		// in this case minTimeout will be applied for all getters,so each of them
+		// will have 1 minute timeout.
+		minTimeout = time.Minute
+	}
```
walldiss, line 129: Well, we still may encounter case, when user set 1Hour timeout here.
Perhaphs limiting timeout to be strict max 3min would be simple and robust. 
```
ctx = WithTimeout(ctx,3min)
```

renaynay:
several integration tests failing @vgonkivs

vgonkivs:
I'm not sure if my changes caused that.

share/getters/cascade.go
vgonkivs: > Well, we still may encounter case, when user set 1Hour timeout here. 

True, but then the user explicitly wants to wait 1h. I wanted to cover the case where he hasn't specified anything at all.

cristaloleg:
Looks like 1min timeout is just too much:

> panic: test timed out after 10m0s

Maybe in tests we should keep 15sec or 30sec timeout, 2c.

Wondertan (approved):

cristaloleg (approved):

renaynay (approved):
utack

PR #3592, opened 2024-07-24
feat(shwap): Add bufferisation for writing ods/Q1Q4 files

walldiss:
Use buffered write of eds shares and axis roots to the disk. According to benchmark it improves write time for writing single block from 79ms -> 19ms (4x).

### DIFF
--- a/store/file/ods.go
+++ b/store/file/ods.go
@@ -108,21 +108,44 @@ func writeODSFile(w io.Writer, eds *rsmt2d.ExtendedDataSquare, axisRoots *share.
 		return nil, fmt.Errorf("writing axis roots: %w", err)
 	}
 
-	for _, shr := range eds.FlattenedODS() {
-		if _, err := w.Write(shr); err != nil {
-			return nil, fmt.Errorf("writing shares: %w", err)
-		}
+	// write quadrants
+	err = writeQuadrant(w, eds, 0)
+	if err != nil {
+		return nil, fmt.Errorf("writing Q1: %w", err)
 	}
 	return h, nil
 }
 
+func writeQuadrant(w io.Writer, eds *rsmt2d.ExtendedDataSquare, quadrantIdx int) error {
+	buf := make([]byte, 0, share.Size*eds.Width())
+	fromRow := quadrantIdx / 2 * int(eds.Width()) / 2
+	toRow := fromRow + int(eds.Width())/2
+
+	fromCol := quadrantIdx % 2 * int(eds.Width()) / 2
+	toCol := fromCol + int(eds.Width())/2
+
+	for rowIdx := fromRow; rowIdx < toRow; rowIdx++ {
+		row := eds.Row(uint(rowIdx))
+		for colIdx := fromCol; colIdx < toCol; colIdx++ {
+			buf = append(buf, row[colIdx]...)
+		}
+		if _, err := w.Write(buf); err != nil {
+			return fmt.Errorf("writing shares: %w", err)
+		}
+		buf = buf[:0]
+	}
+	return nil
+}
+
 func writeAxisRoots(w io.Writer, roots *share.AxisRoots) error {
+	buf := make([]byte, 0, share.AxisRootSize*len(roots.RowRoots))
 	for _, roots := range [][][]byte{roots.RowRoots, roots.ColumnRoots} {
 		for _, root := range roots {
-			if _, err := w.Write(root); err != nil {
-				return fmt.Errorf("writing axis root: %w", err)
-			}
+			buf = append(buf, root...)
 		}
+	}
+	if _, err := w.Write(buf); err != nil {
+		return fmt.Errorf("writing axis root: %w", err)
 	}
 	return nil
 }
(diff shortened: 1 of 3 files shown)


store/file/ods.go
```diff
@@ -108,22 +108,45 @@ func writeODSFile(w io.Writer, eds *rsmt2d.ExtendedDataSquare, axisRoots *share.
...
+	// write quadrants
+	err = writeQuadrant(w, eds, 0)
+	if err != nil {
+		return nil, fmt.Errorf("writing Q1: %w", err)
 	}
 	return h, nil
 }
 
+func writeQuadrant(w io.Writer, eds *rsmt2d.ExtendedDataSquare, quadrantIdx int) error {
+	buf := make([]byte, 0, share.Size*eds.Width())
```
Wondertan, line 120: share size from header?

store/file/ods.go
```diff
@@ -108,22 +108,45 @@ func writeODSFile(w io.Writer, eds *rsmt2d.ExtendedDataSquare, axisRoots *share.
...
+	for rowIdx := fromRow; rowIdx < toRow; rowIdx++ {
+		row := eds.Row(uint(rowIdx))
+		for colIdx := fromCol; colIdx < toCol; colIdx++ {
+			buf = append(buf, row[colIdx]...)
+		}
+		if _, err := w.Write(buf); err != nil {
+			return fmt.Errorf("writing shares: %w", err)
+		}
+		buf = buf[:0]
+	}
```
Wondertan, line 136: I don't follow. Why can't we just iterate over half of the rows and write every half of it? Why do we need columns involved at all?

store/file/ods.go
```diff
@@ -108,22 +108,45 @@ func writeODSFile(w io.Writer, eds *rsmt2d.ExtendedDataSquare, axisRoots *share.
...
 	for _, roots := range [][][]byte{roots.RowRoots, roots.ColumnRoots} {
 		for _, root := range roots {
-			if _, err := w.Write(root); err != nil {
-				return fmt.Errorf("writing axis root: %w", err)
-			}
+			buf = append(buf, root...)
 		}
 	}
+	if _, err := w.Write(buf); err != nil {
+		return fmt.Errorf("writing axis root: %w", err)
```
Wondertan, line 148:
```suggestion
		return fmt.Errorf("writing axis roots: %w", err)
```

store/file/ods.go
```diff
@@ -108,22 +108,45 @@ func writeODSFile(w io.Writer, eds *rsmt2d.ExtendedDataSquare, axisRoots *share.
...
-		}
+	// write quadrants
+	err = writeQuadrant(w, eds, 0)
+	if err != nil {
+		return nil, fmt.Errorf("writing Q1: %w", err)
 	}
 	return h, nil
 }
 
+func writeQuadrant(w io.Writer, eds *rsmt2d.ExtendedDataSquare, quadrantIdx int) error {
```
Wondertan, line 119: It would be nice to doc that it writes row by row with a buffer. with a size of a row

store/file/ods.go
Wondertan, line 120: I think on the read side we would benefit a lot by reading row by row, rather than share by share. To be proven with benchmarks

store/file/ods.go
Wondertan: Implementation would be much easier with `bufio.Writer` tbh.

store/file/ods.go
walldiss: It this the case, there is even wip draft PR for that and some memory optimisations for reading buffer as well
https://github.com/celestiaorg/celestia-node/pull/3512

store/file/ods.go
walldiss: We need both rows and columns to be able to write any quadrant, not only the first one.
I think basic slice with append is as simple as it can get (only 3 lines of code). Can replace those 3 lines with 3 lines of bufio.Writer, if you think it is easier to read

store/file/ods.go
```diff
@@ -108,22 +109,50 @@ func writeODSFile(w io.Writer, eds *rsmt2d.ExtendedDataSquare, axisRoots *share.
...
+	}
+	err := buf.Flush()
+	if err != nil {
+		return fmt.Errorf("flushing buffer: %w", err)
+	}
+	return nil
+}
+
 func writeAxisRoots(w io.Writer, roots *share.AxisRoots) error {
+	buf := make([]byte, 0, share.AxisRootSize*len(roots.RowRoots))
```
cristaloleg, line 147: TBH I don't follow why allocating buffer, filling it with data and then writing to `io.Writer` is faster/better. Feels like the most improvement comes from `writeQuadrant` which uses `bufio` rather then this func. WDYT ?

Wondertan (approved):

store/file/ods.go
walldiss: Thank you. Fixed in recent comments

Wondertan: We improved that logic in sync with Vlad on his repo and merged in here

PR #3593, opened 2024-07-24
use buffer to write to file

Wondertan:
A little simplification of #3592


Wondertan:
Done by https://github.com/walldiss/celestia-node/pull/3

Issue #2420, opened 2023-07-03
libs: Add expiry for JWT


renaynay:
Add expiration to JWT generation

Wondertan:
Besides, we should keep support for indefinite tokens

cristaloleg:
Do we want a fixed TTL for the token? or better to give a config param which configures this?

> Besides, we should keep support for indefinite tokens
In other words - token with no TTL is still a correct token (assuming signature is correct).

PR #3595, opened 2024-07-25
feat: add nonce to JWT tokens

cristaloleg:
Add nonce to JWT tokens to make them a bit more random.

### DIFF
--- a/api/rpc/perms/permissions.go
+++ b/api/rpc/perms/permissions.go
@@ -1,6 +1,7 @@
 package perms
 
 import (
+	"crypto/rand"
 	"encoding/json"
 
 	"github.com/cristalhq/jwt/v5"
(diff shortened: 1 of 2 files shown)


distractedm1nd:
Not understanding why this would be necessary - much more important would be to add an optional, disabled by default timeout.

cristaloleg:
Timeout == TTL ?

distractedm1nd:
Yeah I guess you could consider it a sort of TTL - I don't know what they call it with this library but I mean an expiry date, after which the token will be rejected. Its pretty standard for JWTs and actually frowned upon to not have it

cristaloleg:
It's called `ExpiresAt` https://github.com/cristalhq/jwt/blob/main/claims.go#L27  (NOTE: I'm the author of `cristalhq/jwt`)

This PR addresses @renaynay comment that we should add more randomness to the token. Adding TTL is a separate PR, which isn't planned for now, unless I'm wrong (CC: Rene).

distractedm1nd:
I guess I just dont understand why we need more randomness? If somebody is exploiting the randomness there we have much bigger issues anyways, and these nodes are 99% of the time running in isolated networks

Not going to block it, ill approve, im just curious

cristaloleg:
Yep, why not? Gonna close this one and continue #2420. Thanks everyone.

PR #3599, opened 2024-07-25
feat(shwap): Shwap cascade and mock getters

walldiss:
Update according to new shwap.Getter interface. Move to new pkg structure:
- cascase 
- mock getters

Remove:
- old store.Getter
- old ipld.Getter


Wondertan (approved):
🔪

PR #3554, opened 2024-07-06
feat(shwap): Integrate shrex into shwap

walldiss:
This PR add shrex integration with shwap. Key changes:
- Shrex uses shwap-id to request data
- Added shwap-id verification for bot client and server
- bumped shrex version to prevent old clients from hitting new protocol servers
- move shrex into shwap protocols folder (closer to bitswap)
- move shrex getter into shwap/shrex
- update shrex getter implementation and tests

**This PR is meant to be broken on lint and test CI. It will be fixed in subsequent PR, that will integrate shrex into nodebuilder.**

PR #3600, opened 2024-07-25
feat(shwap): Update availability to work with shwap

walldiss:
Updates Availability and related tests to be aligned with shwap underlying getter

Blocked by https://github.com/celestiaorg/celestia-node/pull/3554


share/shwap/p2p/bitswap/getter.go
```diff
@@ -183,10 +183,10 @@ func (g *Getter) GetSharesByNamespace(
 	}
 
 	// TODO(@Wondertan): this must use shwap types eventually
```
Wondertan, line 185: Can remove it as it does that with this PR.

share/availability/full/availability.go
```diff
@@ -40,8 +39,9 @@ func NewShareAvailability(
 // enough Shares from the network.
 func (fa *ShareAvailability) SharesAvailable(ctx context.Context, header *header.ExtendedHeader) error {
 	dah := header.DAH
-	// short-circuit if the given root is an empty data square, to avoid datastore hit
+	// if the data square is empty, we can safely link the header height in the store to an empty EDS.
 	if share.DataHash(dah.Hash()).IsEmptyEDS() {
+		fa.store.Put(ctx, dah, header.Height(), share.EmptyEDS())
```
Wondertan, line 44: should still check for an error

share/availability/full/reconstruction_test.go
```diff
@@ -1,284 +0,0 @@
...
-)
-
-func init() {
-	eds.RetrieveQuadrantTimeout = time.Millisecond * 100 // to speed up tests
-}
-
-// TestShareAvailable_OneFullNode asserts that a full node can ensure
-// data is available (reconstruct data square) while being connected to
-// light nodes only.
-func TestShareAvailable_OneFullNode(t *testing.T) {
```
Wondertan, line 28: We should keep those tests but comment instead, so we can quickly rewrite once we get to reconstruction

We discussed that we need them and agreed to keep them on the availability level.

Wondertan (commented):
Few comments, LFG, otherwise

Wondertan (approved):

PR #3606, opened 2024-07-30
feat(blob): implement length method

vgonkivs:
Added a `Length()` method to the Blob struct that helps to determine the number of the shares in the blob.

### DIFF
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -109,6 +109,7 @@ func (b *Blob) Index() int {
 	return b.index
 }
 
+// Length returns the number of shares in the blob.
 func (b *Blob) Length() (int, error) {
 	s, err := BlobsToShares(b)
 	if err != nil {
@@ -116,7 +117,7 @@ func (b *Blob) Length() (int, error) {
 	}
 
 	if len(s) == 0 {
-		return 0, errors.New("incorrect number of shares")
+		return 0, errors.New("blob with zero shares received")
 	}
 
 	appShare, err := shares.NewShare(s[0])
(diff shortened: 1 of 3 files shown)


blob/blob.go
```diff
@@ -108,6 +109,29 @@ func (b *Blob) Index() int {
 	return b.index
 }
 
+func (b *Blob) Length() (int, error) {
```
Wondertan, line 112: godoc

blob/blob.go
```diff
@@ -108,6 +109,29 @@ func (b *Blob) Index() int {
...
+	s, err := BlobsToShares(b)
+	if err != nil {
+		return 0, err
+	}
+
+	if len(s) == 0 {
+		return 0, errors.New("incorrect number of shares")
```
Wondertan, line 119: Could specify that its zero as the check concretely check for that

blob/blob.go
```diff
@@ -108,6 +109,29 @@ func (b *Blob) Index() int {
...
+	if err != nil {
+		return 0, err
+	}
+
+	length, err := appShare.SequenceLen()
+	if err != nil {
+		return 0, err
+	}
+
+	return shares.SparseSharesNeeded(length), nil
```
Wondertan, line 132: What is this call for and why returning length from SequenceLen is not enough?

vgonkivs: `SequenceLen` returns the sequence length(and not the exact number of shares). `SparseSharesNeeded` returns the number of shares needed to store the sequenceLen.

Wondertan (approved):

renaynay (approved):

PR #3609, opened 2024-07-31
chore(shwap): various improvements

Wondertan:
* Fixes ErrInvalidID
  * Comment
  * Name
  * Location
* Adds notes for EDS and NamespaceData containers
* Adds Read/Write methods on RowNamespaceData to be used by NamespaceData

### DIFF
--- a/share/shwap/row_namespace_data.go
+++ b/share/shwap/row_namespace_data.go
@@ -3,9 +3,11 @@ package shwap
 import (
 	"errors"
 	"fmt"
+	"io"
 
 	"github.com/celestiaorg/celestia-app/pkg/appconsts"
 	"github.com/celestiaorg/celestia-app/pkg/wrapper"
+	"github.com/celestiaorg/go-libp2p-messenger/serde"
 	"github.com/celestiaorg/nmt"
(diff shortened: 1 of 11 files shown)


share/shwap/row_namespace_data.go
```diff
@@ -122,6 +124,18 @@ func RowNamespaceDataFromProto(row *pb.RowNamespaceData) RowNamespaceData {
 	}
 }
 
+// ReadRowNamespaceData reads length-delimited protobuf representation of RowNamespaceData and
+// returns it with total number of bytes written.
+func ReadRowNamespaceData(reader io.Reader) (RowNamespaceData, int, error) {
```
Wondertan, line 129: Made it a func, so that:
* NamespaceData.ReadFrom remains atomic
* Goland is not going crazy about pointer vs value methods(optional)

walldiss: I find shwap pkg to be very heavy in terms of amount of function already. can be also RowNamespaceData.ReadFrom(). It will hide this func from pkg level func surface to method associated with type. 
Goland linting is not a big deal

walldiss (approved):

Issue #3582, opened 2024-07-19
Merge GetCommitmentProof with GetProof


rach-id:
The comment Rene referred to is about combining this with `GetProof`. The `GetProof` is originally intended to give blob commitment proof, instead of namespace proof, which is a bug. The namespace proof is already served by the Share module and we don't need to duplicate that.

I would do this change in this PR as we decided to integrate APIs Blobstream needs into the node _nicely_. I am open to doing that in a separate PR, though, if you think that's better, @rach-id, but as long we release these two PRs together.

_Originally posted by @Wondertan in https://github.com/celestiaorg/celestia-node/pull/3470#discussion_r1684388222_

rach-id:
Might also be related to: https://github.com/celestiaorg/celestia-node/issues/2289

Issue #2289, opened 2023-05-31
Store the first share of the Blob in the blob's proof.


vgonkivs:

nashqueue:
From the first share of the Blob you can get the Sequence Length. From the sequence Length, you can check the length of the blob. If i don't, i can create a correct inclusion proof for 3/4 of a blob that is actually 4 shares long. The malicious commitment would be PFB2
![image](https://github.com/celestiaorg/celestia-node/assets/99758629/d80ef142-0f54-4b0f-b340-c63abab2ec8c)

vgonkivs:
https://github.com/celestiaorg/celestia-node/pull/3610#issuecomment-2263610790

vgonkivs:
Should look closely into this issue during blob over shwap implementation.

PR #3610, opened 2024-07-31
fix(blob)!: update the blob.Proof to return a proof to the data root

rach-id:
Closes https://github.com/celestiaorg/celestia-node/issues/3582


blob/blob.go
```diff
@@ -18,42 +21,107 @@ import (
...
+				return ErrInvalidProof
+			}
+		}
 	}
 	return nil
 }
 
+// Verify takes a set of shares, a namespace and a data root, and verifies if the
+// provided shares are committed to by the data root.
+func (p *Proof) Verify(rawShares [][]byte, namespace share.Namespace, dataRoot []byte) (bool, error) {
```
rach-id, line 88: We will also add another method that will verify this proof but using the subtree roots

rach-id: https://github.com/celestiaorg/celestia-node/issues/3611

blob/blob.go
```diff
@@ -16,44 +18,104 @@ import (
 	"github.com/celestiaorg/celestia-node/share"
 )
 
+//nolint:unused
 var errEmptyShares = errors.New("empty shares")
```
Wondertan, line 22: Let's remove it if its unused

blob/parser.go
```diff
@@ -21,6 +21,8 @@ type parser struct {
 
 // set tries to find the first blob's share by skipping padding shares and
 // sets the metadata of the blob(index and length)
+//
+//nolint:unused
```
Wondertan, line 25: @vgonkivs, should we keep this code if it's unused now?

api/docgen/exampledata/blobProof.json
```diff
@@ -1,31 +1,53 @@
...
-    "nodes": [
-      "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAMJ/xGlNMdEAAAAAAAAAAAAAAAAAAAAAAAAAAAAAwn/EaU0x0UTO9HUGKjyjcv5U2gHeSjJ8S1rftqv6k8kxlVWW8e/7",
-      "/////////////////////////////////////////////////////////////////////////////wexh4khLQ9HQ2X6nh9wU5B+m6r+LWwPTEDTa5/CosDF"
-    ],
-    "is_max_namespace_ignored": true
+    "start_row": 1,
+    "end_row": 3
   }
-]
+}
```
Wondertan, line 53: new line

share/eds/edstest/testing.go
```diff
@@ -138,13 +139,22 @@ func createTestBlobTransaction(
 	ns := namespace.RandomBlobNamespace()
 	msg, blob := blobfactory.RandMsgPayForBlobsWithNamespaceAndSigner(addr.String(), ns, size)
 	require.NoError(t, err)
+	cTx := BuildCoreTx(t, signer, msg, blob)
+	return ns, msg, blob, cTx
+}
 
+func BuildCoreTx(
```
Wondertan, line 146: A little godoc description wouldn't hurt

blob/service_test.go
```diff
@@ -287,7 +281,9 @@ func TestBlobService_Get(t *testing.T) {
 				require.ErrorIs(t, err, ErrInvalidProof)
 				included, ok := res.(bool)
 				require.True(t, ok)
-				require.True(t, included)
+				// This is supposed to be false not true.
+				// we're querying blob[1]'s proof and verifying it against blob[0], no?
+				require.False(t, included)
```
Wondertan, line 286: cc @vgonkivs

blob/blob.go
rach-id: want to confirm with @vgonkivs  first

blob/parser.go
Wondertan: Actually, we are likely gonna use it later on, but on the server side. The only way for the server to serve blobs with proofs is by parsing the whole namespace/row and this code is gonna be useful there, so we might keep it for now

blob/service.go
```diff
@@ -351,123 +362,187 @@ func (s *Service) retrieve(
...
-			err = ErrBlobNotFound
+	// get all the shares of the rows containing the namespace
+	_, getSharesSpan := tracer.Start(ctx, "get-all-shares-in-namespace")
+	// store the ODS shares of the rows containing the blob
+	rowsShares := make([]share.Share, 0, (exclusiveNamespaceEndRowIndex-inclusiveNamespaceStartRowIndex)*squareSize)
+	// store the EDS shares of the rows containing the blob
+	rowsWithParityShares := make([][]shares.Share, exclusiveNamespaceEndRowIndex-inclusiveNamespaceStartRowIndex)
+	for rowIndex := inclusiveNamespaceStartRowIndex; rowIndex < exclusiveNamespaceEndRowIndex; rowIndex++ {
+		for colIndex := 0; colIndex < squareSize*2; colIndex++ {
+			share, err := s.shareGetter.GetShare(ctx, header, rowIndex, colIndex)
```
vgonkivs, line 397: having a blob that consists of 256 shares(or in case of GetAll request), you will make 256 independent network calls in case of the light node. Can we rework it by replacing with GetEDS instead?

blob/service.go
```diff
@@ -351,123 +362,187 @@ func (s *Service) retrieve(
...
+		// skip if it's a padding share
+		isPadding, err := currentShareApp.IsPadding()
 		if err != nil {
 			return nil, nil, err
 		}
+		if isPadding {
+			currentShareIndex++
+			continue
+		}
+		isCompactShare, err := currentShareApp.IsCompactShare()
```
vgonkivs, line 435: Is this something new? as I'm not aware about `compactShares`?

vgonkivs (commented):
Left some comments. Generally speaking the changes look very good but I still need time to understand `blob.Index `calculation. 


P.S. I believe [this](https://github.com/celestiaorg/celestia-node/issues/2289) issue is not relevant anymore: blob retrieval will download **all** shares anyway, so the downloaded commitment will differ from the passed one, in case smn wants to pass a commitment without 1 share.

blob/service.go
rach-id: those are transaction shares: https://celestiaorg.github.io/celestia-app/specs/shares.html#transaction-shares
Could be any Celestia transaction.

blob/blob.go
rach-id: afb2293718e9d4413e7761dfe2c697cb83726a77

api/docgen/exampledata/blobProof.json
rach-id: 052a79751d133df92da76bf51317e9764b4762cd

share/eds/edstest/testing.go
rach-id: a006f94a188ece637a3be15cdd8d14be8876fd2a

blob/service_test.go
rach-id: 698d90671df4045c5c4ce1bd59713d4826434206

blob/blob.go
```diff
@@ -16,44 +18,104 @@ import (
...
-		for i, node := range pNodes {
-			if !bytes.Equal(node, inputNodes[i]) {
+func (p *Proof) equal(input *Proof) error {
+	// compare the NMT proofs
+	for i, proof := range p.ShareToRowRootProof {
+		if proof.Start != input.ShareToRowRootProof[i].Start {
+			// TODO(@rach-id): should we define specific errors for each case? that's better
+			// to know which part is not equal.
+			// Also, this error should be ErrProofNotEqual or similar, and not ErrInvalidProof
+			return ErrInvalidProof
```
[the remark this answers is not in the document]
rach-id: ccb5799c7cd4575d78735f3f6244da44decfa9e2

blob/service.go
```diff
@@ -315,15 +318,23 @@ func (s *Service) Included(
...
 		return false, nil
 	default:
 		return false, err
 	}
-	return true, resProof.equal(*proof)
+	appShares, err := BlobsToShares(blob)
+	if err != nil {
+		return false, err
+	}
+	getter, err := s.headerGetter(ctx, height)
```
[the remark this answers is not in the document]
rach-id: 952b4add6b2560cd2e2c77b4a1b32516816ea445

blob/service.go
rach-id: We should then define a threshold that we're over it, we should just get the whole EDS. Otherwise, we will be querying the whole EDS for any single blob even if it's just 1 row...

Wondertan: Fetching the whole EDS will take more bandwidth, but it will be substantially faster. If we want to keep assuming the LN cannot afford that even for a few months from now, then we can do it by shares, but it will be slower and lead to more load on FN/BN

Note that this is based on our understanding and intuition of how the current protocol works, and we don't have a proper benchmark to prove it.

Wondertan: I assume there is no way to do it over `SharesByNamespace`, right?

rach-id: `SharesByNamespace` doesn't provide enough information to generate the proof.

We could set a limit to 100 shares, for example. If the queried shares are more, then we query the EDS directly, otherwise we query the shares separately. What do you think?

Later on, we can improve on the value based on the blobs sizes that are submitted to Celestia and how often their proofs are queried.

rach-id:
> P.S. I believe https://github.com/celestiaorg/celestia-node/issues/2289 issue is not relevant anymore: blob retrieval will download all shares anyway, so the downloaded commitment will differ from the passed one, in case smn wants to pass a commitment without 1 share.

Agreed. And in the case of a rollup referencing data using the share commitment, we can document somewhere the possibilities so that they're aware of this.

blob/service.go
vgonkivs: I'd any advocate for `GetEDS`. Assuming you have an EDS with blobs[b1,b2,b3,b4]. So in the worst case, you'd have to download the whole EDS to get b4, in the best case - only n shares of the first blob. I agree that GetEDS will take a lot more bandwidth in case of the best case, but imo it will be much much faster than the worst case. Also, GetEDS fits more for the GetAll request, since you will anyway have to fetch al the shares under the same namespace. We should also think about users that are actively using `GetAll`: getting shares one by one might take some time, so there might cases when the request will be timeouted.
Setting a threshold for N shares IMO is not the case(and the such logic will be harder to understand and maintain, which could lead a ton of bugs. As Blob Module is pretty hot and popular within the node, it will be painful to fix those bugs), since you don't know whether the share belongs to the blob that you are looking for(you will detect it once download all shares of the blob a create a commitment).

I personally think that should sacrifice the bandwidth in favour of speed.

rach-id: sure, as you prefer. This means that we don't need to improve the `GetShareProof` to only get the needed shares :D I'll update the implemenetation

rach-id: [2d209ad](https://github.com/celestiaorg/celestia-node/pull/3610/commits/2d209adef1e5cd662efe2994b60236a95ca76d4e)

vgonkivs:
Agreed in DM with @rach-id that the proof Verification logic should additionally check commitment: whether it belongs to the blob or not.

rach-id:
> Agreed in DM with @rach-id that the proof Verification logic should additionally check commitment: whether it belongs to the blob or not.

ae65e8d

blob/service.go
```diff
@@ -347,127 +354,204 @@ func (s *Service) retrieve(
...
+	}
+	if inclusiveNamespaceStartRowIndex == -1 {
+		return nil, nil, ErrBlobNotFound
+	}
+
+	// end exclusive index of the row root containing the namespace
+	exclusiveNamespaceEndRowIndex := inclusiveNamespaceStartRowIndex
+	for i, row := range header.DAH.RowRoots[inclusiveNamespaceStartRowIndex:] {
+		if namespace.IsOutsideRange(row, row) {
+			exclusiveNamespaceEndRowIndex = inclusiveNamespaceStartRowIndex + i
```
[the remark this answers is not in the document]
rach-id: [e473057](https://github.com/celestiaorg/celestia-node/pull/3610/commits/e473057d5f0e3a9486dce557a5565c7e7962c809)

blob/service.go
```diff
@@ -347,127 +354,204 @@ func (s *Service) retrieve(
...
+			if endShareIndex > len(rowsShares) {
+				// this blob spans to the next row which has a namespace > requested namespace.
+				// this means that we can stop.
+				return nil, nil, ErrBlobNotFound
+			}
+			// convert the blob shares to app shares
+			blobShares := rowsShares[currentShareIndex:endShareIndex]
+			appBlobShares, err := toAppShares(blobShares...)
+			if err != nil {
+				return nil, nil, err
```
vgonkivs, line 466: Can we just create a blob here? We already know the exact length of it and get all shares, so, let's just call `NewBlob`. It will simplify logic since you don't need using parser anymore.

blob/service.go
```diff
@@ -347,127 +354,204 @@ func (s *Service) retrieve(
...
 
-			shrs, isComplete = sharesParser.addShares(appShares)
-			// move to the next row if the blob is incomplete
+			// parse the blob
+			sharesParser.length = blobLen
+			_, isComplete := sharesParser.addShares(appBlobShares)
 			if !isComplete {
-				appShares = nil
-				break
+				return nil, nil, fmt.Errorf("expected the shares to construct a full blob")
```
vgonkivs, line 473: Can be removed. See comment above.

blob/service.go
```diff
@@ -347,127 +354,204 @@ func (s *Service) retrieve(
...
 			}
-			// otherwise construct blob
 			blob, err := sharesParser.parse()
 			if err != nil {
 				return nil, nil, err
 			}
```
vgonkivs, line 478: can be also removed

blob/service.go
```diff
@@ -347,127 +354,204 @@ func (s *Service) retrieve(
...
-					appShares = shrs
-				}
+			endShareIndex := currentShareIndex + blobLen
+			if endShareIndex > len(rowsShares) {
+				// this blob spans to the next row which has a namespace > requested namespace.
+				// this means that we can stop.
+				return nil, nil, ErrBlobNotFound
+			}
+			// convert the blob shares to app shares
+			blobShares := rowsShares[currentShareIndex:endShareIndex]
```
vgonkivs, line 463: should be exclusive, no? since `endShareIndex := currentShareIndex + blobLen`

blob/service.go
```diff
@@ -347,127 +354,204 @@ func (s *Service) retrieve(
...
+			sharesFromCurrentRowStart := currentShareIndex % squareSize
+			// setting the index manually since we didn't use the parser.set() method
+			blob.index = sharesFromSquareStartToNsStart +
+				sharesFromNsStartToCurrentRow +
+				sharesFromCurrentRowStart
+
+			if blob.Namespace().Equals(namespace) && sharesParser.verify(blob) {
+				// now that we found the requested blob, we will create
+				// its inclusion proof.
+				inclusiveBlobStartRowIndex := inclusiveNamespaceStartRowIndex + currentShareIndex/squareSize
```
vgonkivs, line 498: Do we need this calculation after we've calculated the `blob.Index`? can't it be calculated this way: `blob.Index/(squareSize*2)`?

blob/service.go
```diff
@@ -347,127 +354,204 @@ func (s *Service) retrieve(
...
-				"namespace: %s detected at %d: %w", namespace.String(), height, err)
-			log.Error(err)
-		}
-	}
-	return nil, nil, err
+// proveRowRootsToDataRoot creates a set of binary merkle proofs for all the
+// roots defined by the range [start, end).
+func proveRowRootsToDataRoot(roots [][]byte, start, end int) []*merkle.Proof {
+	_, proofs := merkle.ProofsFromByteSlices(roots)
+	return proofs[start:end]
```
vgonkivs, line 557: Can we move this function to the blob.go, where Proof is defined?

blob/service.go
```diff
@@ -347,127 +354,204 @@ func (s *Service) retrieve(
...
+			// setting the index manually since we didn't use the parser.set() method
+			blob.index = sharesFromSquareStartToNsStart +
+				sharesFromNsStartToCurrentRow +
+				sharesFromCurrentRowStart
+
+			if blob.Namespace().Equals(namespace) && sharesParser.verify(blob) {
+				// now that we found the requested blob, we will create
+				// its inclusion proof.
+				inclusiveBlobStartRowIndex := inclusiveNamespaceStartRowIndex + currentShareIndex/squareSize
+				exclusiveBlobEndRowIndex := inclusiveNamespaceStartRowIndex + endShareIndex/squareSize
```
vgonkivs, line 499: I would not call it exclusive as you include it in every range:
`range header.DAH.RowRoots[inclusiveBlobStartRowIndex:exclusiveBlobEndRowIndex]`

vgonkivs (commented):
Made one more review round 🙂. During it I noted all the variable that were used in PR with brief explanation 🤣. I have an impression that it will cause a lot confusion in future when we revisit this code. I want to propose to rename them now: since we are dealing with ods/eds shares, you include those into namings. It will definitely simplify the understanding.

blob/service.go
```diff
@@ -347,127 +354,204 @@ func (s *Service) retrieve(
...
+	_, getSharesSpan := tracer.Start(ctx, "get-all-shares-in-namespace")
+	// store the ODS shares of the rows containing the blob
+	rowsShares := make([]share.Share, 0, (exclusiveNamespaceEndRowIndex-inclusiveNamespaceStartRowIndex)*squareSize)
+	// store the EDS shares of the rows containing the blob
+	rowsWithParityShares := make([][]shares.Share, exclusiveNamespaceEndRowIndex-inclusiveNamespaceStartRowIndex)
+
+	for rowIndex := inclusiveNamespaceStartRowIndex; rowIndex < exclusiveNamespaceEndRowIndex; rowIndex++ {
+		for colIndex := 0; colIndex < squareSize*2; colIndex++ {
+			share := eds.GetCell(uint(rowIndex), uint(colIndex))
+			if colIndex < squareSize {
```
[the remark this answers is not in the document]
rach-id: [e5677a8](https://github.com/celestiaorg/celestia-node/pull/3610/commits/e5677a8d498f140555f7213e050ee972e97de866)

blob/service.go
```diff
@@ -347,127 +354,204 @@ func (s *Service) retrieve(
...
 
-	// collect shares for the requested namespace
-	namespacedShares, err := s.shareGetter.GetSharesByNamespace(getCtx, header, namespace)
-	if err != nil {
-		if errors.Is(err, share.ErrNotFound) {
-			err = ErrBlobNotFound
+	// get all the shares of the rows containing the namespace
+	_, getSharesSpan := tracer.Start(ctx, "get-all-shares-in-namespace")
+	// store the ODS shares of the rows containing the blob
+	rowsShares := make([]share.Share, 0, (exclusiveNamespaceEndRowIndex-inclusiveNamespaceStartRowIndex)*squareSize)
```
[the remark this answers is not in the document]
rach-id: [d7bb13b](https://github.com/celestiaorg/celestia-node/pull/3610/commits/d7bb13b4ff89bd00fab59635e8e49e0bb5f20bb7)

blob/service.go
```diff
@@ -347,127 +354,204 @@ func (s *Service) retrieve(
...
+	// store the EDS shares of the rows containing the blob
+	rowsWithParityShares := make([][]shares.Share, exclusiveNamespaceEndRowIndex-inclusiveNamespaceStartRowIndex)
```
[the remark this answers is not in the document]
rach-id: [11bfc48](https://github.com/celestiaorg/celestia-node/pull/3610/commits/11bfc485c48197427e951683d35e400ccea6416f)

blob/service.go
rach-id: I keep this logic because we call `sharesParser.verify()` which can have custom verification function and it's used by `getAll` etc...

blob/service.go
rach-id: https://github.com/celestiaorg/celestia-node/pull/3610#discussion_r1702092515

blob/service.go
rach-id: https://github.com/celestiaorg/celestia-node/pull/3610#discussion_r1702092515

blob/service.go
rach-id: yes, the `endShareIndex` is exclusive. Renamed in: 500f0b8

For example, we have the following row:

0 1 2 3 4 5 6 7

currentShareIndex := 2
blobLen := 3

since `currentShareIndex` is the first share of the blob, the end share of the blob is `4` : `{2, 3, 4}`

2 + 3 = 5 which is exclusive

blob/service.go
rach-id: yes, want me to update?

```suggestion
				inclusiveBlobStartRowIndex := blob.index / (squareSize * 2)
```

blob/service.go
rach-id: [11d425c](https://github.com/celestiaorg/celestia-node/pull/3610/commits/11d425c3fe02293c3229f14bad985d6a5a075f56)

blob/service.go
rach-id: the ranges when slicing go slices are exclusive, so it's exclusive, no?

blob/service.go
vgonkivs: we can rework retrieve by applying `verifyFn() `. But it's up to you. we can do it later

blob/service.go
vgonkivs: As it exclusive, shouldn't it be:
`blobShares := rowsShares[currentShareIndex:endShareIndex-1]`
currentShareIndex := 2
blobLen := 3
2+3 =5 
But the actual end share is 4

blob/service.go
vgonkivs: yes, please

blob/service.go
vgonkivs: true. But you should range over 
header.DAH.RowRoots[inclusiveBlobStartRowIndex:exclusiveBlobEndRowIndex-1] then, no?

blob/service.go
rach-id: yes, but didn't want to remove the share parser in this PR and have to change unrelated code too. Let's leave it for a separate PR

blob/service.go
rach-id: for example: `slice := []int{0, 1, 2, 3, 4}`
`slice[0:2] =  {0, 1}`.

The end range is not included in the go slice. Go slicing is end exclusive

blob/service.go
rach-id: https://github.com/celestiaorg/celestia-node/pull/3610#discussion_r1703865887

blob/blob.go
```diff
@@ -16,42 +19,70 @@ import (
...
-	if p.Len() != input.Len() {
-		return ErrInvalidProof
+// Verify takes a blob and a data root and verifies if the
+// provided blob was committed to by the data root.
+func (p *Proof) Verify(blob *Blob, dataRoot []byte) (bool, error) {
+	blobCommitment, err := types.CreateCommitment(ToAppBlobs(blob)[0])
+	if err != nil {
+		return false, err
+	}
+	if !blob.Commitment.Equal(blobCommitment) {
```
renaynay, line 42: Here we are checking if the blob commitment we generate using the given blob matches the commitment inside the given blob? 

What does this have to do with the proof then (re: error message)? wouldn't it just be a malformed blob?

blob/blob.go
```diff
@@ -16,42 +19,70 @@ import (
...
+	RowProof coretypes.RowProof
+}
 
-// equal is a temporary method that compares two proofs.
-// should be removed in BlobService V1.
-func (p Proof) equal(input Proof) error {
-	if p.Len() != input.Len() {
-		return ErrInvalidProof
+// Verify takes a blob and a data root and verifies if the
+// provided blob was committed to by the data root.
```
renaynay, line 36:
```suggestion
// Verify takes a blob and a data root and verifies if the
// provided blob was committed to the given data root.
```

renaynay (commented):
very difficult to verify the parsing logic here, so could benefit from more tests with malformed proofs + blobs

blob/blob.go
```diff
@@ -16,42 +18,63 @@ import (
...
-
-func (p Proof) Len() int { return len(p) }
+// Proof constructs the proof of a blob to the data root.
+type Proof struct {
+	// ShareToRowRootProof the proofs of the shares to the row roots they belong to.
+	// If the blob spans across multiple rows, then this will contain multiple proofs.
+	ShareToRowRootProof []*tmproto.NMTProof
+	// RowProof the proofs of the row roots containing the blob shares
+	// to the data root.
+	RowProof coretypes.RowProof
```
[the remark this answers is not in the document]
rach-id: [5b054bc](https://github.com/celestiaorg/celestia-node/pull/3610/commits/5b054bc56d12437886dbe2a9f681eb0d57c5c853)

blob/helper.go
```diff
@@ -39,7 +39,7 @@ func BlobsToShares(blobs ...*Blob) ([]share.Share, error) {
 
 // ToAppBlobs converts node's blob type to the blob type from celestia-app.
 func ToAppBlobs(blobs ...*Blob) []*apptypes.Blob {
-	appBlobs := make([]*apptypes.Blob, 0, len(blobs))
+	appBlobs := make([]*apptypes.Blob, len(blobs))
```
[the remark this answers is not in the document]
rach-id: was buggy before

blob/blob.go
rach-id: that's a malformed blob, yes. should I create a new error type?

renaynay: Not sure if a whole new error type is necessary unless we have multiple places where a malformed blob can be detected. regardless, we need the err msg to be accurate

blob/blob.go
```diff
@@ -16,42 +19,70 @@ import (
...
+	ns := append([]byte{namespace.Version()}, namespace.ID()...)
+	cursor := int32(0)
+	for i, proof := range p.ShareToRowRootProof {
+		nmtProof := nmt.NewInclusionProof(
+			int(proof.Start),
+			int(proof.End),
+			proof.Nodes,
+			true,
+		)
+		sharesUsed := proof.End - proof.Start
```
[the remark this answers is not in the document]
rach-id: [28cb03f](https://github.com/celestiaorg/celestia-node/pull/3610/commits/28cb03f036a42f938a2eff43d0886176a9575e78)

blob/blob.go
```diff
@@ -16,42 +19,70 @@ import (
 	"github.com/celestiaorg/celestia-node/share"
 )
 
+//nolint:unused
```
Wondertan, line 22: Weird that its still here, even there was a comment addressing it

Wondertan, line 22: Weird that its still here, even there was a comment addressing it

blob/service.go
```diff
@@ -347,127 +353,193 @@ func (s *Service) retrieve(
...
-		rowIndex++
-		if sharesParser.isEmpty() {
-			proofs = nil
-		}
-	}
+				// create the share to row root proofs
+				shareToRowRootProofs, _, err := pkgproof.CreateShareToRowRootProofs(
+					squareSize,
+					//nolint:lll
+					edsShares[inclusiveBlobStartRowIndex-inclusiveNamespaceStartRowIndex:exclusiveBlobEndRowIndex-inclusiveNamespaceStartRowIndex],
```
cristaloleg, line 514: Is there a bug? lower bound has 2 vars with `inclusive` and upped bound has 1 var with `exclusive` and 1 with `inclusive`. Looks sus.

Can we split this into 3 lines, please?
```
start := inclusiveBlobStartRowIndex-inclusiveNamespaceStartRowIndex
end := exclusiveBlobEndRowIndex-inclusiveNamespaceStartRowIndex
...
edsShares[start : end]
```

cristaloleg: Also not symmetrical `...StartRow - ...StartRow` BUT `...EndRow - ...StartRow`. Which might be ok but sus 🤷

blob/blob.go
rach-id: https://github.com/celestiaorg/celestia-node/pull/3610#discussion_r1700202243

blob/service.go
```diff
@@ -347,127 +353,193 @@ func (s *Service) retrieve(
...
+	exclusiveNamespaceEndRowIndex := inclusiveNamespaceStartRowIndex
+	for _, row := range header.DAH.RowRoots[inclusiveNamespaceStartRowIndex:] {
+		if namespace.IsOutsideRange(row, row) {
+			break
+		}
+		exclusiveNamespaceEndRowIndex++
+	}
+	if exclusiveNamespaceEndRowIndex == inclusiveNamespaceStartRowIndex {
+		return nil, nil, fmt.Errorf("couldn't find the row index of the namespace end")
+	}
```
cristaloleg, line 386: Q: so, it's ok when we just iterate over `header.DAH.RowRoots[inclusiveNamespaceStartRowIndex:]` till the end. This means that we cover all ranges, am I right?

blob/blob.go
Wondertan: That comment referred to a diff code and path tho

Wondertan: Although, it doesn't hurt to keep it if we keep some other deadcode anyway

rach-id: > That comment referred to a diff code and path tho

yes, but that's the only code using this. So the linter also detects this as unused.

blob/service.go
rach-id: yes, the namespaces are ordered

blob/service_test.go
```diff
@@ -1061,3 +1061,124 @@ func generateCommitmentProofFromBlock(
...
+			}(),
+			blob: func() Blob {
+				b := *blob
+				b.Commitment = []byte{0x1}
+				return b
+			}(),
+			expectErr: true,
+		},
+		{
+			name:     "mismatched number of share proofs and row proofs",
```
renaynay, line 1148: this is same test case as invalid proof

renaynay (commented):
one comment here otherwise thank u for tests :)

blob/service_test.go
rach-id: [857e3ed](https://github.com/celestiaorg/celestia-node/pull/3610/commits/857e3ed106e17a43ca88eff68fc2672ca1fddb7c)

vgonkivs:
Btw, unit test failed

blob/service.go
cristaloleg: No bug in indexes, right? Just asking for the confirmation.

rach-id:
Unit test fixed: [712c272](https://github.com/celestiaorg/celestia-node/pull/3610/commits/712c2728ea118f571019133bb09e28467a29d3d4)

was a good catch 👍

vgonkivs (approved):

blob/service.go
rach-id: yes, no bug. We're shifting the indexes by `inclusiveNamespaceStartRowIndex` because the EDS shares start from `[inclusiveNamespaceStartRowIndex: exclusiveNamespaceEndRowIndex]`

blob/blob.go
renaynay: Yes bc this error doesn't have anything to do with the proof, just the blob

renaynay (approved):
Thank you @rach-id let's get this over the line :)

blob/blob.go
```diff
@@ -16,42 +19,70 @@ import (
...
-
-func (p Proof) Len() int { return len(p) }
+// Proof constructs the proof of a blob to the data root.
+type Proof struct {
+	// ShareToRowRootProof the proofs of the shares to the row roots they belong to.
+	// If the blob spans across multiple rows, then this will contain multiple proofs.
+	ShareToRowRootProof []*tmproto.NMTProof
+	// RowToDataRootProof the proofs of the row roots containing the blob shares
+	// to the data root.
+	RowToDataRootProof coretypes.RowProof
```
Wondertan, line 32: Q: why do we call DataRoot if inside header it is called `DataHash`?

rach-id: That's an existential question. From the dawn of existence, the name `DataRoot` has echoed through the ages, a timeless whisper woven into the fabric of eternity.

That's a theory vs implementation issue... In the fraud proofs paper, it's referred to as `DataRoot`. In tendermint, it's called `DataHash`, and AFAIK we never really discussed choosing one over the other.

blob/blob.go
rach-id: [3630f33](https://github.com/celestiaorg/celestia-node/pull/3610/commits/3630f332e858c72ae0445e1abb9871b0fcc1e946)

PR #3612, opened 2024-08-01
refactor(pruner): modify pruner checkpoint to init at height 0

### LINKED ISSUE #3474: refactor(share/pruner): make pruner init to start from 0
### Implementation ideas

Pruner starts with initial checkpoint with height 1, but still needs to prune that height on first run. It is managed by 
```
if lastPruned.Height() == 1 {
		headers = append([]*header.ExtendedHeader{lastPruned}, headers...)
	}
```
Rework initial checkpoint and logic for first run to use 0 height at init.

najeal:
Closes #3474 

This changes the init of pruner checkpoint to start from height 0.

I moved one error in a common package as I cannot use direct reference to a injected dependency.

I modified logic to use possible nil lastPruned header as we have no pruned header at height 0. nothing looks broken regarding the tests but a special attention may be useful.
I am wondering if this change is correct at https://github.com/najeal/celestia-node/blob/df230c411f2f3f1f282806b3f3da4d6b12ad070e/pruner/find.go#L23-L27

### DIFF
--- a/header/header.go
+++ b/header/header.go
@@ -17,6 +17,9 @@ import (
 	"github.com/celestiaorg/rsmt2d"
 )
 
+// ErrHeightZero returned when the provided block height is equal to 0.
+var ErrHeightZero = errors.New("height is equal to 0")
+
 // ConstructFn aliases a function that creates an ExtendedHeader.
 type ConstructFn = func(
 	*core.Header,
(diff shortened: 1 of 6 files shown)


header/header.go
```diff
@@ -17,6 +17,9 @@ import (
 	"github.com/celestiaorg/rsmt2d"
 )
 
+// ErrHeightZero returned when the provided block height is equal to 0.
+var ErrHeightZero = errors.New("height is equal to 0")
```
renaynay, line 21: can you please move this error back to where it was originally? 🙏🏻

renaynay (commented):
Hey @najeal thank you for the contribution - there's a couple of lint errors there - do you mind resolving them and the issue I marked? It'll make it easier to review the PR :) thanks

renaynay:
Hey @najeal Thank you for the contribution, unfortunately it is now stale. Hope to see you around this repository again soon!

PR #3614, opened 2024-08-01
feat(shwap): update daser for shwap

walldiss:
Add shwap integration to DASer

### DIFF
--- a/das/daser_test.go
+++ b/das/daser_test.go
@@ -2,37 +2,27 @@ package das
 
 import (
 	"context"
+	"fmt"
 	"strconv"
 	"testing"
 	"time"
 
 	"github.com/golang/mock/gomock"
-	"github.com/ipfs/boxo/blockservice"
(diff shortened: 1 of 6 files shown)


Wondertan (approved):

das/daser_test.go
```diff
@@ -144,73 +136,76 @@ func TestDASer_Restart(t *testing.T) {
...
-	select {
-	case <-ctx.Done():
-		t.Fatal(ctx.Err())
-	case res := <-resultCh:
-		require.NoError(t, res)
-	}
-	// wait for manager to finish catchup
-	require.True(t, daser.running == 0)
-}
+// TODO(@walldiss): BEFP test will not work until BEFP-shwap integration
```
cristaloleg, line 139: Why not just `t.Skip()` ?

Wondertan: I presumed it doesn't compile

PR #3615, opened 2024-08-02
feat(shwap): Add caching to blockstore

walldiss:


store/cache/doublecache.go
```diff
@@ -29,6 +30,20 @@ func (mc *DoubleCache) Get(height uint64) (eds.AccessorStreamer, error) {
...
 }
 
+// GetOrLoad attempts to get an item from the both caches and, if not found, invokes
+// the provided loader function to load it into the first Cache.
+func (mc *DoubleCache) GetOrLoad(
+	ctx context.Context,
+	height uint64,
+	loader OpenAccessorFn,
+) (eds.AccessorStreamer, error) {
+	accessor, err := mc.second.Get(height)
```
cristaloleg, line 40: In `Get` we access `first` and then `second` cache (which sounds right). Why in `GetOrLoad` we accessing `second` and then `first` ?

walldiss: Read is optimistic (no tx gurantees). Order does not matter for read. `GetOrLoad` should load into first cache, so it checks `second` first. I'll add comments do clarify

cristaloleg: I didn't mention tx at all. But order indeed looks weird. Comment should be enough, thanks.

walldiss: Added comments

store/store_options.go
```diff
@@ -7,27 +7,18 @@ import (
 type Parameters struct {
 	// RecentBlocksCacheSize is the size of the cache for recent blocks.
 	RecentBlocksCacheSize int
-
-	// AvailabilityCacheSize is the size of the cache for accessors requested for serving availability
-	// samples.
-	AvailabilityCacheSize int
```
Wondertan, line 13: Another concern that I had, but forgot to bring up during our sync is this. Where do you see this configuration param moving with the approach in this PR.
IMO these cache configurations should be close together, as they were.

store/store_options.go
```diff
@@ -7,27 +7,18 @@ import (
...
 func (p *Parameters) Validate() error {
-	if p.RecentBlocksCacheSize < 1 {
-		return errors.New("recent eds cache size must be positive")
-	}
-
-	if p.AvailabilityCacheSize < 1 {
-		return errors.New("availability cache size must be positive")
+	if p.RecentBlocksCacheSize < 0 {
+		return errors.New("recent eds cache size cannot be negative")
 	}
```
Wondertan, line 22: If constructor now allows disabling cache by taking 0, we should remove this check.

Wondertan: Also, I recommend adopting simplification changes I made to tests in https://github.com/celestiaorg/celestia-node/pull/3619/files#diff-ad5b6c129e998456b5ae539dd2a5793f7d094b64ed8175edeeecade80c39e008R339-R344

store/metrics.go
```diff
@@ -57,19 +59,26 @@ func (s *Store) WithMetrics() error {
...
 		has:       has,
 		remove:    remove,
-		unreg:     unreg,
 	}
+	return s.metrics.addCacheMetrics(s.cache)
+}
+
+// addCacheMetrics adds cache metrics to store metrics
+func (m *metrics) addCacheMetrics(c cache.Cache) error {
+	if m == nil || c == nil {
```
Wondertan, line 74: is c ever nil? I can only be noop cache no?

store/store_cache.go
```diff
@@ -0,0 +1,50 @@
...
+type CachedStore struct {
+	inner *Store
+	cache *cache.DoubleCache
+}
+
+// WithCache creates new CachedStore with cache of the specified size.
+func (s *Store) WithCache(name string, size int) (*CachedStore, error) {
+	newCache, err := cache.NewAccessorCache(name, size)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create availability cache: %w", err)
```
Wondertan, line 22: name instead of availability

store/store_options.go
Wondertan: I guess we can keep it as it is and then use it nodebuilder where we construct `WithCache`

store/store_options.go
walldiss: This check prevents negative values of CacheSize to be passed

store/store_options.go
walldiss: Yes, it will be set externally.

Wondertan (approved):

PR #3616, opened 2024-08-02
feat(shwap): Add discovery version

walldiss:
Add versioning for discovery, so we can separate shwap network from older one. Metrics will still use tag instead of topic, so we can support multiple visions with same dashboard

### DIFF
--- a/nodebuilder/p2p/routing.go
+++ b/nodebuilder/p2p/routing.go
@@ -32,6 +32,12 @@ func newDHT(
 		return nil, fmt.Errorf("unsupported node type: %s", tp)
 	}
 
+	// no bootstrappers for a bootstrapper ¯\_(ツ)_/¯
+	// otherwise dht.Bootstrap(OnStart hook) will deadlock
+	if isBootstrapper() {
+		bootsrappers = nil
+	}
+
 	dht, err := discovery.NewDHT(ctx, network.String(), bootsrappers, host, dataStore, mode)
 	if err != nil {
 		return nil, err
--- a/nodebuilder/p2p/routing.go
+++ b/nodebuilder/p2p/routing.go
@@ -6,70 +6,46 @@ import (
 
 	"github.com/ipfs/go-datastore"
 	dht "github.com/libp2p/go-libp2p-kad-dht"
-	"github.com/libp2p/go-libp2p/core/protocol"
 	"github.com/libp2p/go-libp2p/core/routing"
 	"go.uber.org/fx"
 
 	"github.com/celestiaorg/celestia-node/nodebuilder/node"
+	"github.com/celestiaorg/celestia-node/share/shwap/p2p/discovery"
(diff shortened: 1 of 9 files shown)


share/shwap/p2p/discovery/discovery.go
```diff
@@ -34,6 +34,10 @@ const (
 	// logInterval defines the time interval at which a warning message will be logged
 	// if the desired number of nodes is not detected.
 	logInterval = 5 * time.Minute
+
+	// discovery version is a prefix for all tags used in discovery. It is bumped when
+	// there are protocol breaking changes.
+	version = "v1"
```
Wondertan, line 40: The breaking changes we have are not in discovery. The breaking changes are with shrex protocol or the protocol that uses the discovery tool. So instead of doing global bump for the tool, only ensuring tag given to discovery should be enough

walldiss: The change allows to decouple versioning from tags, which is important to keep other tag relateed logic, like metric tags. We dont want metric names to be bumped with every version

share/shwap/p2p/discovery/dht.go
```diff
@@ -0,0 +1,42 @@
...
+	ctx context.Context,
+	prefix string,
+	bootsrappers []peer.AddrInfo,
+	host host.Host,
+	dataStore datastore.Batching,
+	mode dht.ModeOpt,
+) (*dht.IpfsDHT, error) {
+	opts := []dht.Option{
+		dht.BootstrapPeers(bootsrappers...),
+		dht.ProtocolPrefix(protocol.ID(fmt.Sprintf("/celestia/%s/%s", prefix, version))),
```
Wondertan, line 35: We don't need the version here but on the topics! This won't allow nodes to share DHT.

Wondertan (changes requested):

share/shwap/p2p/discovery/discovery.go
Wondertan, line 40: sync: change to prefix and supply in constructor instead.

nodebuilder/p2p/routing.go
```diff
@@ -6,70 +6,46 @@ import (
...
-		dht.Datastore(params.DataStore),
-		dht.RoutingTableRefreshPeriod(cfg.RoutingTableRefreshPeriod),
-	}
-
-	if isBootstrapper() {
-		opts = append(opts,
-			dht.BootstrapPeers(), // no bootstrappers for a bootstrapper ¯\_(ツ)_/¯
-		)
-	}
-
```
Wondertan, line 37: We need this 😑 , otherwise bootstrapping(OnStart hook) will deadlock bootstrappers

Wondertan (changes requested):

Wondertan (approved):

PR #3617, opened 2024-08-02
feat(shwap): add shwap store getter

walldiss:
add store getter implementation of shwap.Getter.


Wondertan:
Agreed to cover GetEDS and GetND with tests

Wondertan (approved):

PR #136, opened 2021-10-14
node: add cache for bitswap

vgonkivs:
Cached blockstore was used instead of common.

CacheOpts contains 3 params:
* HasBloomFilterSize - size of bloom filter that will be used in blockStore (default value is 512 << 10);
* HasBloomFilterHashes - amount of hash functions (default value is 7);
* HasARCCacheSize - size of arc cache (default value is 64 << 10);

Closes #108

PR #3619, opened 2024-08-03
feat(shwap/store): availability cache

Wondertan:
A simpler alternative to #136 which also avoid DoubleCache

### DIFF
--- a/store/store.go
+++ b/store/store.go
@@ -37,16 +37,19 @@ var ErrNotFound = errors.New("eds not found in store")
 
 // Store is a storage for EDS files. It persists EDS files on disk in form of Q1Q4 files or ODS
 // files. It provides methods to put, get and remove EDS files. It has two caches: recent eds cache
-// and availability cache. Recent eds cache is used to cache recent blocks. Availability cache is
-// used to cache blocks that are accessed by sample requests. Store is thread-safe.
+// and availability cache. Recent eds cache is used to cache full EDSes on Put. Availability cache is
+// populated externally by protocols via Cache method. Store is thread-safe.
 type Store struct {
 	// basepath is the root directory of the store
 	basepath string
-	// cache is used to cache recent blocks and blocks that are accessed frequently
-	cache *cache.DoubleCache
+
+	recent       cache.Cache
+	availability cache.Cache
+
 	// stripedLocks is used to synchronize parallel operations
 	stripLock *striplock
-	metrics   *metrics
+
+	metrics *metrics
 }
 
 // NewStore creates a new EDS Store under the given basepath and datastore.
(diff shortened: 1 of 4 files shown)


store/store.go
```diff
@@ -266,8 +293,15 @@ func (s *Store) HasByHeight(ctx context.Context, height uint64) (bool, error) {
 }
 
 func (s *Store) hasByHeight(height uint64) (bool, error) {
-	_, err := s.cache.Get(height)
```
Wondertan, line 269: This was a bug because we didn't close the Accessor after Get. Warrants adding Has method to the cache.

store/store_test.go
```diff
@@ -152,7 +154,7 @@ func TestEDSStore(t *testing.T) {
 		require.NoError(t, err)
 
 		// file should be removed from cache
-		_, err = edsStore.cache.Get(height)
+		_, err = edsStore.recent.Get(height)
```
Wondertan, line 157: This was testing  on NoopCache. Fixed now

store/store.go
```diff
@@ -37,16 +37,19 @@ var ErrNotFound = errors.New("eds not found in store")
...
+// and availability cache. Recent eds cache is used to cache full EDSes on Put. Availability cache is
+// populated externally by protocols via Cache method. Store is thread-safe.
 type Store struct {
 	// basepath is the root directory of the store
 	basepath string
-	// cache is used to cache recent blocks and blocks that are accessed frequently
-	cache *cache.DoubleCache
+
+	recent       cache.Cache
+	availability cache.Cache
```
Wondertan, line 47: BUG: They don't share reference counting

* Put A -> Recent
* Get A <- Recent
* Put A -> Avail
* Close A -> Recent
* Get A <- Avail (kaput)

Wondertan: Technically this is a bug, but it would never happen in practice because we only put to recent rsmt2d accessor, which has Close as OOO. We should fix this anyway.

Wondertan: Once Put starts taking any general Accessor this becomes problematic, so has to be fixed anyway.

store/store.go
```diff
@@ -96,6 +106,17 @@ func (s *Store) Close() error {
 	return s.metrics.close()
 }
 
+func (s *Store) Cache(height uint64, acsr eds.AccessorStreamer) error {
```
walldiss, line 109: In order to get and cache accessor in blocks for you would need to open it twice.
1. Get accessor
2. Cache accessor
The first issue with this approach is that it is inefficient. The second, it is not concurrency safe, because parallel Get call may result in duplicate Put of same accessor in the cache. GetOrLoad method in cache allows to avoid both issues, if used directly by blockstore.

store/store.go
```diff
@@ -74,20 +77,27 @@ func NewStore(params *Parameters, basePath string) (*Store, error) {
...
+	if params.AvailabilityCacheSize > 0 {
+		availabilityCache, err = cache.NewAccessorCache("availability", params.AvailabilityCacheSize)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create availability cache: %w", err)
+		}
 	}
 
 	store := &Store{
-		basepath:  basePath,
-		cache:     cache.NewDoubleCache(recentEDSCache, availabilityCache),
```
walldiss, line 89: It looks like you removed DoubleCache, but still read from both caches every time. I wonder what is the reason to remove, because double cache did exactly the same, while also having advantage of having read logic in single place instead of multiple copies.

walldiss (commented):
This PR introduces cache() method, that needs to be called every time store user want to cache something. 
I’m generally leaning toward the idea of making cache a config / setting of the store, that you set up once on launch of store and no longer need to worry about afterwards. So would rather keep cache inside store, rather then making it external

store/store.go
Wondertan: Why is it inefficient?

Wondertan: There is no concurrency issue because Cache has height based locking and if two Puts happen simultaneously the other one will resolve gracefully with Get portion instead of Load

store/store.go
Wondertan: There is only one time where Get happens. There other time happens because there is no Has method on the Cache. Anyway, I will add private Get method to avoid duplication until we get Has.

walldiss: Why not return double cache that does exactly what is needed?

store/store.go
walldiss: It is inefficient because for every get you need to open file 2 times. When you don’t need to if only getorload is used.
Concurrency issue happens when 2 parallel get happens. They will load different accessors  with independent internal axis caches. Then you will load third one into cache by Cache method. Third accessor will be cached, while read will happen from first and second one. No axis cache proofs from first and second will be kept in third(chached) accessor

Wondertan:
In sync decided to go with #3615

PR #3620, opened 2024-08-03
feat(shwap/ods_file): tail padding trim

Wondertan:
Implements tail padding share trimming a.k.a. compression with supporting changes. The actual trimming logic turned out extremely simple and does not even require any form of indexing. (896bdd3211dc36ede3209a66b35888f996a5d8bc)

This PR is reviewable commit-by-commit and is supposed to merge as so.

Depends on #3643

### DIFF
--- a/share/new_eds/read.go
+++ b/share/new_eds/read.go
@@ -4,8 +4,9 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"github.com/celestiaorg/celestia-node/share"
 	"io"
+
+	"github.com/celestiaorg/celestia-node/share"
 )
 
 // ReadAccessor reads up EDS out of the io.Reader until io.EOF and provides.
--- a/share/new_eds/read.go
+++ b/share/new_eds/read.go
@@ -3,6 +3,7 @@ package eds
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 
@@ -37,19 +38,23 @@ func ReadAccessor(ctx context.Context, reader io.Reader, root *share.AxisRoots) 
 }
 
 // ReadShares reads shares from the provided io.Reader until EOF. If EOF is reached, the remaining shares
-// are populated as padding share. Provided reader must contain shares in row-major order.
+// are populated as tail padding shares. Provided reader must contain shares in row-major order.
 func ReadShares(r io.Reader, shareSize, odsSize int) ([]share.Share, error) {
 	shares := make([]share.Share, odsSize*odsSize)
 	var total int
 	for i := range shares {
 		shr := make(share.Share, shareSize)
 		n, err := io.ReadFull(r, shr)
+		if errors.Is(err, io.EOF) {
+			for ; i < len(shares); i++ {
+				shares[i] = share.TailPadding()
+			}
+			return shares, nil
+		}
 		if err != nil {
-			return nil, fmt.Errorf("reading share: %w, bytes read: %v", err, total+n)
+			return nil, fmt.Errorf("reading shares: %w, bytes read: %v", err, total+n)
 		}
-		if n != shareSize {
-			return nil, fmt.Errorf("share size mismatch: expected %v, got %v", shareSize, n)
-		}
+
 		shares[i] = shr
 		total += n
 	}
--- a/share/new_eds/read.go
+++ b/share/new_eds/read.go
@@ -1,0 +1,56 @@
+package eds
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"github.com/celestiaorg/celestia-node/share"
+	"io"
+)
+
+// ReadAccessor reads up EDS out of the io.Reader until io.EOF and provides.
+func ReadAccessor(ctx context.Context, reader io.Reader, root *share.AxisRoots) (*Rsmt2D, error) {
+	odsSize := len(root.RowRoots) / 2
+	shares, err := ReadShares(reader, share.Size, odsSize)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read eds from ods bytes: %w", err)
+	}
+
+	// verify that the EDS hash matches the expected hash
+	rsmt2d, err := Rsmt2DFromShares(shares, odsSize)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create rsmt2d from shares: %w", err)
+	}
+	datahash, err := rsmt2d.DataHash(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to calculate data hash: %w", err)
+	}
+	if !bytes.Equal(datahash, root.Hash()) {
+		return nil, fmt.Errorf(
+			"content integrity mismatch: imported root %s doesn't match expected root %s",
+			datahash,
+			root.Hash(),
+		)
+	}
+	return rsmt2d, nil
+}
+
+// ReadShares reads shares from the provided io.Reader until EOF. If EOF is reached, the remaining shares
+// are populated as padding share. Provided reader must contain shares in row-major order.
+func ReadShares(r io.Reader, shareSize, odsSize int) ([]share.Share, error) {
+	shares := make([]share.Share, odsSize*odsSize)
+	var total int
+	for i := range shares {
+		shr := make(share.Share, shareSize)
+		n, err := io.ReadFull(r, shr)
+		if err != nil {
+			return nil, fmt.Errorf("reading share: %w, bytes read: %v", err, total+n)
+		}
+		if n != shareSize {
+			return nil, fmt.Errorf("share size mismatch: expected %v, got %v", shareSize, n)
+		}
+		shares[i] = shr
+		total += n
+	}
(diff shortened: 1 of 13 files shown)


share/share.go
```diff
@@ -21,6 +22,9 @@ const (
 // rsmt2d.ExtendedDataSquare.
 var MaxSquareSize = appconsts.SquareSizeUpperBound(appconsts.LatestVersion)
 
+// TailPadding is a constant tail padding share exported for reuse
+var TailPadding Share
```
cristaloleg, line 26: What about make this as a func? `GetTailPadding() Share` by example.

store/file/ods.go
```diff
@@ -368,17 +375,25 @@ func readRow(fl io.ReaderAt, hdr *headerV0, sharesOffset, quadrantIdx, rowIdx in
 	offset := sharesOffset + quadrantOffset + rowOffset
 
 	axsData := make([]byte, odsLn*shrLn)
-	if _, err := fl.ReadAt(axsData, int64(offset)); err != nil {
+	n, err := fl.ReadAt(axsData, int64(offset))
+	if err != nil && !errors.Is(err, io.EOF) {
```
cristaloleg, line 379: Should we just `return nil, nil` on `io.EOF` ?

share/share.go
Wondertan: What would that give us?

cristaloleg: Global exported variable is worse than an exported func. You can change behaviour with less pain in the future. +variable gives you ability to change it from another package, `constant` in the comment is no-op.

store/file/ods.go
Wondertan: Continuing on EOF is intentional and the core of this change. If it reads beyond the file boundaries it feels the row shares with TailPadding shares.

share/share.go
Wondertan: Ah I see you point. I thought you also meant copying the share for every caller.

We do functions in a similar case with empty eds, for the reasons you mentioned. I am gonna do this here as well. Thanks for catching it

store/file/ods.go
Wondertan: So the write side ends writing earlier if it sees tail padding shares, and the read side adds it back if the file is less than the expected data.

share/new_eds/read.go
```diff
@@ -0,0 +1,65 @@
...
+func ReadShares(r io.Reader, shareSize, odsSize int) ([]share.Share, error) {
+	shares := make([]share.Share, odsSize*odsSize)
+	var total int
+	for i := range shares {
+		shr := make(share.Share, shareSize)
+		n, err := io.ReadFull(r, shr)
+		if errors.Is(err, io.EOF) {
+			shares[i] = share.TailPadding
+			continue
+		}
```
walldiss, line 54: There is small inefficiency. After EOF is reached, every remaining share can be set as TailPadding. But the loop will keep allocating new `shr` slice on each iteration, despite share is no longer needed to be read. Should be simple to make another loop to fill TailPadding without extra allocations.

share/new_eds/testing.go
```diff
@@ -35,28 +36,31 @@ func TestSuiteAccessor(
 		t.Errorf("minSize must be power of 2: %v", maxSize)
 	}
 	for size := minSize; size <= maxSize; size *= 2 {
+		padding := rand.IntN(size)
```
walldiss, line 39: This random will select amount of padding in range of  [0;rowSize) So there always will be less than full row of padding. Lets ste up tests for cases to test instead:
- no padding
- at least 1 row.  Could be achieved by random in range [rowSize; 3/4 * rowSize*rowSize].

share/new_eds/testing.go
```diff
@@ -67,18 +71,23 @@ func TestStreamer(
 	create createAccessorStreamer,
 	odsSize int,
 ) {
```
walldiss, line 73: Why this one is using odsSize still, when other tests are working over eds?

store/file/ods.go
```diff
@@ -389,9 +404,14 @@ func readCol(fl io.ReaderAt, hdr *headerV0, sharesOffset, quadrantIdx, colIdx in
...
+		n, err := fl.ReadAt(shr, int64(offset))
+		if err != nil && !errors.Is(err, io.EOF) {
 			return nil, err
 		}
+		if n == 0 {
+			shr = share.TailPadding
+		}
+
 		shares[i] = shr
 	}
```
walldiss, line 416: Same loop allocations problem as in eds.ReadShares

walldiss (commented):
Nice trick with io.EOF triggering padding!

store/file/ods.go
Wondertan: I am not sure we can solve it in this case.  ReadAt reads the whole row at once and we have to allocate it before we know its EOF. I don't think its a big deal tho as we allocate as much as we do in the regular case and with buffer pooling its gonna be even less noticable.

share/new_eds/testing.go
Wondertan: Because it is the only test that needs customized EDS with namespaces

store/file/ods.go
Wondertan: The above is for readRow. For readCol we can do that, yes

share/new_eds/read.go
```diff
@@ -0,0 +1,65 @@
...
+	"fmt"
+	"io"
+
+	"github.com/celestiaorg/rsmt2d"
+
+	"github.com/celestiaorg/celestia-node/share"
+)
+
+// ReadEDS reads up EDS out of the io.Reader until io.EOF.
+// TODO(@Wondertan): Should become ReadAccessor.
```
walldiss, line 16: Remove todo

share/new_eds/read.go
```diff
@@ -0,0 +1,65 @@
...
+	shares := make([]share.Share, odsSize*odsSize)
+	var total int
+	for i := range shares {
+		shr := make(share.Share, shareSize)
+		n, err := io.ReadFull(r, shr)
+		if errors.Is(err, io.EOF) {
+			for ; i < len(shares); i++ {
+				shares[i] = share.TailPadding()
+			}
+			return shares, nil
```
walldiss, line 55: Optional: Same iterator reused within loops. Extracting inner loop into small function should improve readability

share/new_eds/testing.go
```diff
@@ -35,29 +36,49 @@ func TestSuiteAccessor(
...
+				testAccessorShares(ctx, t, createAccessor, eds)
+			})
+		}
+	}
+}
+
+func testEDSes(t *testing.T, size int) map[string]*rsmt2d.ExtendedDataSquare {
+	fullEDS := edstest.RandEDS(t, size)
+
+	padding := rand.IntN(size * size)
```
walldiss, line 76: Should be at least 1 padding share

store/file/header.go
```diff
@@ -79,6 +79,20 @@ func (h *headerV0) Size() int {
 	return headerVOSize + 1
 }
 
+func (h *headerV0) RootsSize() int {
+	// axis roots are stored in two parts: row roots and column roots, each part has size equal to
+	// the square size. Thus, the total amount of roots is equal to the square size * 2.
+	return share.AxisRootSize * h.SquareSize() * 2
+}
```
walldiss, line 86: OPtinal: RootsSize and OffsetWithRoots is used only for ods file. Can be hidden away from whole header surface, if implemented as methods on `ods` file

store/file/ods.go
```diff
@@ -362,45 +350,69 @@ func (o *ODS) readODS() (square, error) {
...
+	n, err := r.ReadAt(axsData, int64(offset))
+	if err != nil && !errors.Is(err, io.EOF) {
+		// unknown error
 		return nil, err
 	}
 
+	shrsRead := n / hdr.ShareSize()
 	for i := range shares {
-		shares[i] = axsData[i*shrLn : (i+1)*shrLn]
+		if i > shrsRead-1 {
```
walldiss, line 381: Just curious, where is -1 comes from?

store/file/ods.go
```diff
@@ -100,11 +100,16 @@ func writeODSFile(f *os.File, axisRoots *share.AxisRoots, eds *rsmt2d.ExtendedDa
...
+// order. Write finishes once all the shares are written or on the first instance of tail padding share.
+// Tail padding share are constant and aren't stored.
 func writeODS(w io.Writer, eds *rsmt2d.ExtendedDataSquare) error {
 	for i := range eds.Width() / 2 {
 		for j := range eds.Width() / 2 {
 			shr := eds.GetCell(i, j) // TODO: Avoid copying inside GetCell
+			if share.GetNamespace(shr).Equals(share.TailPaddingNamespace) {
+				return nil
+			}
+
```
walldiss, line 112: Need test for ods file to ensure file store was actually trimmed

share/new_eds/testing.go
Wondertan: done

walldiss (approved):

Issue #3570, opened 2024-07-17
DNS TTL Not Respected by celestia-node Leading to Sync Issues


smuu:
**Description:**

We encountered an issue where changes to the DNS entries of DA nodes in Arabica caused light nodes to fail to sync. Restarting the light nodes resolved the issue, indicating that they resolve DNS once at startup and then use the same IP address indefinitely, ignoring DNS TTL.

**Steps to Reproduce:**
1. Change the DNS entries for nodes.
2. Observe light nodes failing to sync.
3. Restart the light nodes and observe they can sync.

**Suspected Cause:**
Light nodes resolve DNS entries only once at startup and continue using the same IP address without respecting the TTL. This affects both:
- DA nodes connecting to other DA nodes.
- DA Bridge nodes connecting to consensus nodes.

**Relevant Code:**
- `multiaddr` DNS resolution: I could not find the relevant code.
- `--core.ip` DNS resolution:
  https://github.com/celestiaorg/celestia-node/blob/f98d632818d566c7b4fd995b0f4bdc6443a7ed06/nodebuilder/core/config.go#L47

**Potential Fix:**
1. Periodically re-resolve DNS entries based on the TTL.
2. Update active connections if the resolved IP address changes.

**Repositories Potentially Needing Changes:**
- [`celestiaorg/celestia-node`](https://github.com/celestiaorg/celestia-node)
- [`libp2p/go-libp2p`](https://github.com/libp2p/go-libp2p)
- [`multiformats/go-multiaddr`](https://github.com/multiformats/go-multiaddr)

**Impact:**
Not respecting DNS TTL can lead to connectivity and sync issues, affecting network reliability.

**Request for Assistance:**
1. Identify where DNS resolution is handled in the codebase and dependencies.
2. Implement periodic DNS resolution based on TTL.
3. Test changes to ensure nodes dynamically update connections based on DNS updates.

smuu:
> i think simplest here is to remove https://github.com/celestiaorg/celestia-node/blob/main/libs/utils/address.go#L40 which resolves the IP a single time on start, instead letting clients use the domain as passed in, and relying on the infra of the internet to work
> 
> unless there was there good a reason we HAD to resolve IP?

If I understand the code correctly, this would only resolve one part of the issue: the connection between the DA BN and the consensus node.
From my understanding, this code is not called when resolving the DNS in a `multiaddr`.

renaynay:
What's status on this @ramin @smuu ?

PR #3624, opened 2024-08-05
fix(libs/utils): rely on DNS for rpc/grpc requests to core

ramin:
Refs #3570 and I believe solves the issue well enough

previously, on start, node would resolve DNS hostnames to IP when validating. This is somewhat legitimate and necessary when building multiaddr format, but the same code was being used for configuring http clients for RPC/gRPC connections. 

This meant, that in the event of some server administration or an incident, if DNS was updated for a "core" node (as configured via `--core.ip`) this meant a long running node would suddenly lose it's destination and get stuck failing and retrying. 

Instead, this seeks to separate validation, for validating --core.ip for rpc/grpc clients do NOT resolve to IP on start/validate, then should make requests over the DNS using hostname, meaning a core node moving, should eventually resolve again properly, preventing need for a restart.

We had to take care NOT to accidentally cause multiaddr to remain hostnames. The original issue suggests wanting to update all p2p networking to be host based, I feel that is probably unnecessary

### DIFF
--- a/libs/utils/address.go
+++ b/libs/utils/address.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"net/netip"
 	"strings"
 )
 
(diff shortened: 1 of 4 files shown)


libs/utils/address.go
```diff
@@ -37,9 +52,5 @@ func ValidateAddr(addr string) (string, error) {
 		return addr, nil
 	}
 
-	resolved, err := net.ResolveIPAddr("ip4", addr)
-	if err != nil {
-		return addr, err
-	}
-	return resolved.String(), nil
+	return addr, nil
```
cristaloleg, line 55: Looks like this and `return` above are identical. Probably we should return error in `return` above.

Also, we can use https://pkg.go.dev/net/netip#ParseAddr to get error message, `net.ParseIP` is an old and not friendly way to parse ip.

libs/utils/address.go
```diff
@@ -24,9 +24,24 @@ func SanitizeAddr(addr string) (string, error) {
...
-// sanitized address is returned.
+// ValidateAddr sanitizes the given address and verifies that it
+// is a valid IP or hostname. The sanitized address is returned.
 func ValidateAddr(addr string) (string, error) {
+	addr, err := ValidateAddrPreservingHostname(addr)
+	if err != nil {
+		return addr, err
+	}
+
+	resolved, err := net.ResolveIPAddr("ip4", addr)
```
cristaloleg, line 35: Also can be https://pkg.go.dev/net/netip#ParseAddr

renaynay (commented):
agree with @cristaloleg comments

cristaloleg (approved):
LGTM, thanks!

Wondertan (approved):
Agree with p2p networking reasoning. Thanks@

PR #3625, opened 2024-08-05
refactor(share): use sync.Once instead of mutex for init

cristaloleg:
We use `share.EmptyRoot()` in many places, concurrently. To get an empty root we need to take a mutex for a short time but why not to make it better by changing to a simple atomic check.

`sync.Once` exactly does this in a very lightweight way.


Wondertan (approved):
good call

PR #3626, opened 2024-08-05
fix(core_accessor): pass only default account for txClient

vgonkivs:
Previous version of app has a bug, so the defaultAddress had to be explicitly set. The issue is [fixed](https://github.com/celestiaorg/celestia-app/blob/4f4d0f7ff1a43b62b232726e52d1793616423df7/pkg/user/tx_client.go#L72) in the newer release.


renaynay (approved):
utack

Also @rach-id will have to rebase his PR on this

renaynay:
@vgonkivs actually pls wait for rachid to merge his PR, will be way way easier for u to rebase than him

Wondertan (dismissed):

Issue #3629, opened 2024-08-06
Improve the blob endpoints not to query the EDS on every get


rach-id:
As of https://github.com/celestiaorg/celestia-node/pull/3610, we query the EDS everytime when getting the proof for a blob.

This might lead to degraded performance. We can improve on only using that type of proof when `GetProof`. Otherwise, we keep just verifying the inclusion of the namespace.

PR #3633, opened 2024-08-07
feat: keep using the namespace proof for GetBlob and GetAll

rach-id:
Closes https://github.com/celestiaorg/celestia-node/issues/3629

### DIFF
--- a/blob/service.go
+++ b/blob/service.go
@@ -365,9 +365,9 @@ func (s *Service) retrieve(
 			break
 		}
 	}
-	if rowIndex == -1 {
-		return nil, nil, fmt.Errorf("couldn't find rows containing the required namespace")
-	}
+
+	// Note: there is no need to check whether the row index is different from -1
+	// because it will be handled at the end to return the correct error.
 
 	getCtx, getSharesSpan := tracer.Start(ctx, "get-shares-by-namespace")
 
--- a/blob/service.go
+++ b/blob/service.go
@@ -365,6 +365,9 @@ func (s *Service) retrieve(
 			break
 		}
 	}
+	if rowIndex == -1 {
+		return nil, nil, fmt.Errorf("couldn't find rows containing the required namespace")
+	}
 
 	getCtx, getSharesSpan := tracer.Start(ctx, "get-shares-by-namespace")
 
--- a/blob/service.go
+++ b/blob/service.go
@@ -524,6 +524,7 @@ func (s *Service) retrieveBlobProof(
 
 	eds, err := s.shareGetter.GetEDS(ctx, header)
 	if err != nil {
+		headerGetterSpan.SetStatus(codes.Error, err.Error())
 		return nil, nil, err
 	}
 	headerGetterSpan.SetStatus(codes.Ok, "")
--- a/blob/service.go
+++ b/blob/service.go
@@ -236,8 +236,8 @@ func (s *Service) GetProof(
 		return blob.compareCommitments(commitment)
 	}}
 
-	_, proof, err = s.retrieve(ctx, height, namespace, sharesParser)
-	return proof, nil
+	_, proof, err = s.retrieveBlobProof(ctx, height, namespace, sharesParser)
+	return proof, err
 }
 
 // GetAll returns all blobs under the given namespaces at the given height.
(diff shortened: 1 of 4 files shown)


blob/service.go
```diff
@@ -386,6 +522,14 @@ func (s *Service) retrieve(
 		return nil, nil, fmt.Errorf("couldn't find the row index of the namespace end")
 	}
 
+	eds, err := s.shareGetter.GetEDS(ctx, header)
+	if err != nil {
+		return nil, nil, err
```
cristaloleg, line 527: Should we record error in span here?

blob/service.go
```diff
@@ -236,8 +236,8 @@ func (s *Service) GetProof(
 		return blob.compareCommitments(commitment)
 	}}
 
-	_, proof, err = s.retrieve(ctx, height, namespace, sharesParser)
-	return proof, nil
+	_, proof, err = s.retrieveBlobProof(ctx, height, namespace, sharesParser)
+	return proof, err
```
cristaloleg, line 240: Nice catch!

blob/service.go
rach-id: [5534b54](https://github.com/celestiaorg/celestia-node/pull/3633/commits/5534b54173a1ceb124bc9770678d8c967ae96c06)

blob/service.go
```diff
@@ -354,14 +354,150 @@ func (s *Service) retrieve(
...
+	headerGetterSpan.AddEvent("received eds", trace.WithAttributes(
+		attribute.Int64("eds-size", int64(len(header.DAH.RowRoots)))))
+
+	rowIndex := -1
+	for i, row := range header.DAH.RowRoots {
+		if !namespace.IsOutsideRange(row, row) {
+			rowIndex = i
+			break
+		}
+	}
```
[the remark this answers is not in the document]
rach-id: can you explain more what you means?

blob/service.go
```diff
@@ -354,14 +354,150 @@ func (s *Service) retrieve(
...
+
```
cristaloleg, line 368: Should we verify here that `rowIndex` is not `-1` ?

blob/service.go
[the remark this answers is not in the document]
vgonkivs: My suggestion is to return `ErrBlobNotFound` in case rowIndex is -1 after the loop finishes.

blob/service.go
rach-id: [b241e7e](https://github.com/celestiaorg/celestia-node/pull/3633/commits/b241e7e8171ac1dc18dfea63c4a2b88fa905d722)

blob/service.go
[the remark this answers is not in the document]
rach-id: [b241e7e](https://github.com/celestiaorg/celestia-node/pull/3633/commits/b241e7e8171ac1dc18dfea63c4a2b88fa905d722)

blob/service.go
```diff
@@ -354,14 +354,153 @@ func (s *Service) retrieve(
 		return nil, nil, err
 	}
 
-	eds, err := s.shareGetter.GetEDS(ctx, header)
+	headerGetterSpan.SetStatus(codes.Ok, "")
```
Wondertan, line 357: tracing: Ok status should be set once on success. 

Currently, it's written always by default, while error cases do not overwrite it, so error op may show as success.

blob/service.go
```diff
@@ -354,14 +354,153 @@ func (s *Service) retrieve(
...
+	headerGetterSpan.AddEvent("received eds", trace.WithAttributes(
```
Wondertan, line 358:
```suggestion
	headerGetterSpan.AddEvent("received header", trace.WithAttributes(
```

blob/service.go
```diff
@@ -354,14 +354,153 @@ func (s *Service) retrieve(
...
+// retrieve retrieves blobs and their proofs by requesting the whole namespace and
+// comparing Commitments.
+// Retrieving is stopped once the `verify` condition in shareParser is met.
+func (s *Service) retrieveBlobProof(
+	ctx context.Context,
+	height uint64,
+	namespace share.Namespace,
+	sharesParser *parser,
+) (_ *Blob, _ *Proof, err error) {
+	log.Infow("requesting blob",
```
Wondertan, line 492:
```suggestion
	log.Infow("requesting blob proof",
```

because proof is now retrieved via `retrieve`

Wondertan (commented):
Mostly tracing nits

blob/service.go
Wondertan: Actually, it is correct. I missed the fact that there several spans in this method, which we agreed not to refactor in this PR with @rach-id

Wondertan (approved):

blob/service.go
rach-id: upon sync discussion, this is how it was previously, and a separate issue would take care of hiding the statuses in their corresponding methods.

PR #3635, opened 2024-08-08
test(nodebuilder): disable TestArchivalBlobSync

cristaloleg:
Disable this test for now due to its frequent failures. Will be revisited later.


vgonkivs (commented):
Could you create an issue and add it here above `t.Skip()`?

cristaloleg:
Done https://github.com/celestiaorg/celestia-node/issues/3636

walldiss (dismissed):

ramin:
@cristaloleg we had wrapped these skips in testing.Short() before so they don't vanish completely, and CI can be run with the short flag (assuming its some resource/bootstrap constraint in ci env)  but locally they still run, might be worth doing similar

renaynay (approved):

Wondertan (approved):

PR #3638, opened 2024-08-08
feat(shwap): Improve eds store cache refcounter forgot Close() protection

walldiss:
- Force close Accessor if not closed within timeout
- Close Accessor and log Error if ref was removed by GC instead of user
- close Accessor on has check hitting the cache


Wondertan (approved):
LGTM but would port the refCloser part and force closing to main as well

PR #3641, opened 2024-08-08
fix(p2p): couple bitswap client and server

walldiss:
Decoupled client was not able to notify server on newly available blocks.

### DIFF
--- a/nodebuilder/p2p/bitswap.go
+++ b/nodebuilder/p2p/bitswap.go
@@ -3,6 +3,7 @@ package p2p
 import (
 	"context"
 	"fmt"
+
 	"github.com/ipfs/boxo/bitswap"
 	"github.com/ipfs/boxo/bitswap/network"
 	"github.com/ipfs/boxo/blockstore"
--- a/nodebuilder/p2p/bitswap.go
+++ b/nodebuilder/p2p/bitswap.go
@@ -2,12 +2,11 @@ package p2p
 
 import (
 	"context"
-	"errors"
 	"fmt"
+	"slices"
 
-	"github.com/ipfs/boxo/bitswap/client"
+	"github.com/ipfs/boxo/bitswap"
 	"github.com/ipfs/boxo/bitswap/network"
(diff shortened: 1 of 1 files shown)


nodebuilder/p2p/bitswap.go
```diff
@@ -32,36 +31,27 @@ const (
...
+		bitswap.ProvideEnabled(false), // we don't provide blocks over DHT
 		// NOTE: These below are required for our protocol to work reliably.
 		// // See https://github.com/celestiaorg/celestia-node/issues/732
-		server.SetSendDontHaves(false),
-	)
+		bitswap.SetSendDontHaves(false),
+	}
+	clOpts := []bitswap.Option{
+		bitswap.SetSimulateDontHavesOnTimeout(false),
+		bitswap.WithoutDuplicatedBlockStats(),
```
Wondertan, line 43: You could also add them in a single var and separate client server via comments

Wondertan (approved):

Wondertan (approved):

nodebuilder/p2p/bitswap.go
```diff
@@ -32,36 +29,26 @@ const (
...
 		// // See https://github.com/celestiaorg/celestia-node/issues/732
-		server.SetSendDontHaves(false),
-	)
+		bitswap.SetSendDontHaves(false),
 
-	clnt := client.New(
-		params.Ctx,
-		net,
-		params.Bs,
-		client.WithBlockReceivedNotifier(srvr),
```
renaynay, line 49: so bc they're coupled again now, we don't need to specify this?

renaynay (dismissed):
utack

nodebuilder/p2p/bitswap.go
walldiss: yes. `client.WithBlockReceivedNotifier` is enabled inside bitswap.New constructor.

Wondertan (approved):

renaynay (approved):

PR #3642, opened 2024-08-08
fix(store/cache): force close Accessor if not closed within timeout

walldiss:
- Force close Accessor if not closed within timeout

### DIFF
--- a/share/eds/cache/accessor_cache.go
+++ b/share/eds/cache/accessor_cache.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"runtime"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -97,10 +98,12 @@ func (s *accessorWithBlockstore) close() error {
 	done := s.done
 	s.Unlock()
 
+	// wait until all references are released or timeout is reached. If timeout is reached, log an
+	// error and close the accessor forcefully.
 	select {
 	case <-done:
 	case <-time.After(defaultCloseTimeout):
-		return fmt.Errorf("closing accessor, some readers didn't close the accessor within timeout,"+
+		log.Errorf("closing accessor, some readers didn't close the accessor within timeout,"+
 			" amount left: %v", s.refs.Load())
 	}
 	if err := s.shardAccessor.Close(); err != nil {
@@ -235,7 +238,8 @@ func (bc *AccessorCache) EnableMetrics() (CloseMetricsFn, error) {
 // Close is called
 type refCloser struct {
 	*accessorWithBlockstore
-	closeFn func()
+	closed    atomic.Bool
+	removeRef func()
 }
 
 // newRefCloser creates new refCloser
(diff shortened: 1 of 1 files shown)


Wondertan (approved):

share/eds/cache/accessor_cache.go
```diff
@@ -244,17 +248,31 @@ func newRefCloser(abs *accessorWithBlockstore) (*refCloser, error) {
...
-		},
-	}, nil
+		removeRef:              abs.removeRef,
+	}
+	// Set finalizer to ensure that accessor is closed when refCloser is garbage collected.
+	// We expect that refCloser will be closed explicitly by the caller. If it is not closed,
+	// we log an error.
+	runtime.SetFinalizer(rf, func(rf *refCloser) {
+		if rf.close() {
+			log.Errorf("refCloser for accessor was garbage collected before Close was called")
```
renaynay, line 260:
```suggestion
			log.Error("refCloser for accessor was garbage collected before Close was called")
```

renaynay (approved):

share/eds/cache/accessor_cache.go
```diff
@@ -244,17 +248,31 @@ func newRefCloser(abs *accessorWithBlockstore) (*refCloser, error) {
...
-		closeFn: func() {
-			closeOnce.Do(abs.removeRef)
-		},
-	}, nil
+		removeRef:              abs.removeRef,
+	}
+	// Set finalizer to ensure that accessor is closed when refCloser is garbage collected.
+	// We expect that refCloser will be closed explicitly by the caller. If it is not closed,
+	// we log an error.
+	runtime.SetFinalizer(rf, func(rf *refCloser) {
```
cristaloleg, line 258: How many such invocations we will have? Approx.

walldiss: few(1-10) per sec at most

cristaloleg: upto 100 per **second** ??

cristaloleg: Why I'm asking: `runtime.SetFinalizer` behaviour is easy to misuse, see docs (https://pkg.go.dev/runtime#SetFinalizer) & proposal to deprecated it (https://github.com/golang/go/issues/67535)

Adding up to 100/sec looks possible but better to bring attention to this line and solution in general, my 2c.

walldiss: This is one of 2 safety solutions we use for preventing hanging ref from blocking Accessor eviction from cache. 
- Finaliser closing forgotten ref.
- Force closing forgotten refs on timeout.

Given the fact, it is safety only mechanics and not main case. And assuming the only client of cache library is us, we can get away without Finaliser. Considering implications and concerns about Finaliser I'm inclined to remove it in favour of remaining forced closing only. 

@Wondertan thoughts?

cristaloleg: Cool

Wondertan (approved):

vgonkivs (approved):

renaynay (approved):
Okay

PR #3643, opened 2024-08-11
refactor!(shwap/store): decouple Q1Q4 and ODS files

Wondertan:
The main goal of this PR is to decouple Q1Q4 and ODS files physically, so now there are two files `.ods` and `.q4`. This is needed to fix and unblock trimming in #3620. Besides, this PR ends up shaving off another ~27% of times it take for Put.

This PR is split into several commits that are independently reviewable and should be merged as so.

TODO: 
- [x] Doc new funcs
- [x] Lazy open


Wondertan:
One more optimization to do here is to open files lazily. However, that's not trivial, as we detect whether file is empty or not only while opening it and we can't open empty file lazily with ease.

store/file/q4.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+// RemoveQ4 removes Q4 file under the given FS path.
+func RemoveQ4(path string) error {
+	err := os.Remove(path)
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return fmt.Errorf("removing Q4 %s: %w", path, err)
+	}
+	return nil
+}
```
walldiss, line 133: Looks same as RemoveODS. Do we even need this file specific or might move it to store? Additionally it feel to belong to store, because it operates under assumption, that file is closed, which is controlled only on store level.

store/file/q4.go
```diff
@@ -0,0 +1,183 @@
...
+	"os"
+
+	"github.com/celestiaorg/rsmt2d"
+
+	"github.com/celestiaorg/celestia-node/share"
+	eds "github.com/celestiaorg/celestia-node/share/new_eds"
+	"github.com/celestiaorg/celestia-node/share/shwap"
+)
+
+var _ eds.AccessorStreamer = (*Q4)(nil)
```
walldiss, line 18: Do we really need it? Seems like Q4 implements only AxisHalf and Close. And is not going to be used as generic Accessor anywhere. Does not seem that we need Q4 to be Accessor and can remove non-implemented methods.

store/store.go
```diff
@@ -51,44 +50,41 @@ type Store struct {
...
-
-	// Ensure the heights folder exists or is created.
-	heightsFolderPath := filepath.Join(basePath, heightsPath)
-	if err := ensureFolder(heightsFolderPath); err != nil {
-		log.Errorf("Failed to ensure the existence of the heights folder at '%s': %s", heightsFolderPath, err)
-		return nil, fmt.Errorf("ensure heights folder '%s': %w", heightsFolderPath, err)
+	// ensure the blocks dir exists
+	blocksDir := filepath.Join(basePath, blocksPath)
+	if err := mkdir(blocksDir); err != nil {
+		return nil, fmt.Errorf("ensuring blocks directory: %w", err)
```
walldiss, line 61: ensureFolder feel closer than `mkdir` to what actually does: create only when missing. And it is similar to other functions like `ensureEmptyFile` and `ensureHeighLing`

store/store.go
```diff
@@ -172,19 +165,38 @@ func (s *Store) createFile(
...
+		link(pathOds, linktoOds),
+		link(pathQ4, linktoQ4),
+	)
+}
+
+func (s *Store) ensureEmptyFile() error {
+	roots, err := eds.EmptyAccessor.AxisRoots(context.Background())
+	if err != nil {
+		panic(fmt.Sprintf("error getting roots of an empty accessor: %s", err))
+	}
```
walldiss, line 183: why not simple share.EmptyEDSRoots?

store/file/ods_q4_test.go
```diff
@@ -0,0 +1,103 @@
...
+	require.NoError(t, odsq4.Close())
+}
+
+func TestODSQ4File(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+	t.Cleanup(cancel)
+
+	ODSSize := 16
+	eds.TestSuiteAccessor(ctx, t, createODSAccessor, ODSSize)
+}
```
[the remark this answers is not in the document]
Wondertan: It wasn't there, so I didn't do it. Althought I though of that and it simply proxies the call to ODS, so I don't see any value of doing this quite expensive test again.

store/file/q4.go
Wondertan: It's only used to cleanup file on failed create. Doesn't hurt to have it, but can remove. Feel indifferent.

store/store.go
Wondertan: It felt wrong to request the same thing from two different pkgs.

store/store.go
Wondertan: kk

store/file/q4.go
Wondertan: I use it as accessor type in ODSQ4

store/store.go
Wondertan: [39beb05](https://github.com/celestiaorg/celestia-node/pull/3643/commits/39beb053157377caaa21578e904c4f2d5cc7fe08)

walldiss (approved):

PR #3650, opened 2024-08-12
chore(state): remove duplicate function

### LINKED ISSUE #3515: improvement(state/core_access): remove defaultSignerAccount and defaultSignerAddress in favor and use app.ClientTx instead
We can get rid of these two fields inside core_access and ru-use clientTx. Since the node's version of the app contains a small bug in handling `DefaultAccountAddress()`, it should be done after https://github.com/celestiaorg/celestia-app/pull/3597 is included in app release.

Lets also use only `user.WithDefaultAccount(keyName)`  in [SetupTxClient](https://github.com/celestiaorg/celestia-node/blob/f98d632818d566c7b4fd995b0f4bdc6443a7ed06/state/core_access.go#L596)

kobakaku:
closes #3515 

- Fix to use only the `DefaultAccountAddress()` function as args of `SetupTxClient`.


state/core_access.go
```diff
@@ -99,13 +95,13 @@ func NewCoreAccessor(
...
-		coreIP:               coreIP,
-		grpcPort:             grpcPort,
-		prt:                  prt,
+		keyring:  keyring,
+		getter:   getter,
+		coreIP:   coreIP,
+		grpcPort: grpcPort,
+		prt:      prt,
 	}
+	_ = user.WithDefaultAccount(keyname)
```
kobakaku, line 104: The default account (& default address) needs to be updated inside `NewCoreAccessor `.

vgonkivs:
Hey @kobakaku, could you please resolve the merge conflict?

state/core_access.go
cristaloleg: Excuse me but why do we need this line?

PR #3656, opened 2024-08-13
test(nodebuilder): test fraud 4

cristaloleg:
DO NOT REVIEW (2ff58bf4)


nodebuilder/p2p/host.go
```diff
@@ -92,17 +92,17 @@ func host(params hostParams) (HostBase, error) {
...
 			libp2p.Transport(quic.NewTransport),
 			libp2p.Transport(webtransport.New),
 			libp2p.Transport(libp2pwebrtc.New),
-			wsTransport(tlsCfg),
 		),
-		// to clearly define what defaults we rely upon
-		libp2p.DefaultSecurity,
 		libp2p.DefaultMuxers,
 	}
+	wsTransport(tlsCfg)
```
vgonkivs, line 105: all transports will be replaced by the ws in this case. It should be wrapped by `libp2p.ChainOptions`

PR #3663, opened 2024-08-16
fix(nodebuilder/state): overwrite state flags only if they were passed explicitly

### LINKED ISSUE #3661: State.DefaultKeyName value from config .toml is not respected
after 0.15.0 upgrade node doesn't read correct wallet name from config.toml, it looks for `my_celes_key` name 
instead.

adding `--keyring.keyname my_different_key_name` flag resolves the issue.

vgonkivs:
Closes #3661


cristaloleg (approved):
SGTM

Wondertan (approved):

renaynay (approved):

PR #3668, opened 2024-08-20
ci: harder version regexp

cristaloleg:
I have found that our regexp isn't strict enough. See https://regex101.com/r/wr3QE8/1


vgonkivs (approved):

renaynay (approved):
utack

walldiss (approved):

PR #3675, opened 2024-08-22
feat!: Live, laugh, Shwap

Wondertan:
This PR aims to merge Shwap into the `main` branch.

Reworks ~60% of the codebase. 

Commits must be merged without squashing.

image


nodebuilder/tests/reconstruct_test.go
```diff
@@ -36,6 +36,7 @@ Steps:
 5. Check that a FN can retrieve shares from 1 to 20 blocks
 */
 func TestFullReconstructFromBridge(t *testing.T) {
+	t.Skip()
```
cristaloleg, line 39: Why these 3 tests?

share/availability/full/testing.go
```diff
@@ -1,56 +1,57 @@
...
-	err = store.Start(context.Background())
-	require.NoError(t, err)
-
-	t.Cleanup(func() {
-		err = store.Stop(context.Background())
-		require.NoError(t, err)
-	})
-	return NewShareAvailability(store, getter)
-}
+//
```
cristaloleg, line 3: Why commented?

share/availability/light/testing.go
```diff
@@ -1,107 +1,108 @@
...
-}
-
-func (m onceGetter) GetSharesByNamespace(
-	_ context.Context,
-	_ *header.ExtendedHeader,
-	_ share.Namespace,
-) (share.NamespacedShares, error) {
-	panic("not implemented")
-}
+//
```
cristaloleg, line 3: Why commented?

share/availability/full/testing.go
Wondertan: Because we decided to leave reconstruction out of the scope of the initial release and the tests are not compiling. We keep them around so we can quickly refactor them once we back at the reconstruction problem.

nodebuilder/tests/reconstruct_test.go
Wondertan: They are temporarily skipped until we come back to reconstruction. Agree though that adding a reason string to Skip would be helpful cc @walldiss

store/striplock.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+// TODO: move to utils
+type striplock struct {
+	heights    []*sync.RWMutex
+	datahashes []*sync.RWMutex
+}
+
+type multiLock struct {
+	mu []*sync.RWMutex
```
cristaloleg, line 16: Do we need `RWMutex` ? Looks like `Mutex` will be enough.

cristaloleg: Also, do we need a slice? looks like we only using 2 values at most

store/metrics.go
```diff
@@ -0,0 +1,150 @@
...
+	m.unreg = unreg
+	return nil
+}
+
+func (m *metrics) observePut(ctx context.Context, dur time.Duration, size uint, failed bool) {
+	if m == nil {
+		return
+	}
+	if ctx.Err() != nil {
+		ctx = context.Background()
```
cristaloleg, line 90: Same here, infinite ctx.

cristaloleg: See above the same

store/striplock.go
Wondertan: Read lock is used in the store itself

Wondertan: Yes, the slice is needed and there are actually 1024 entries in it

cristaloleg: Thx

walldiss (approved):
Looks harmless. Let’s ship

renaynay (approved):
с богом

PR #3677, opened 2024-08-22
chore(ci): add lint imports ci action

walldiss:
Adds lint imports check to CI to ensure imports order is correct. Found by failed lint locally, that was green in ci.

### DIFF
--- a/.github/workflows/go-ci.yml
+++ b/.github/workflows/go-ci.yml
@@ -49,7 +49,7 @@ jobs:
 
   lint-imports:
     needs: [setup]
-    name: Sort Imports
+    name: Lint Imports
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/go-ci.yml
+++ b/.github/workflows/go-ci.yml
@@ -47,6 +47,24 @@ jobs:
           version: v1.59
           skip-cache: true
 
+  lint-imports:
+    needs: [setup]
+    name: Sort Imports
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version: ${{ inputs.go-version }}
+
+      - name: install goimports-reviser
+        run: go install -v github.com/incu6us/goimports-reviser/v3@latest
+
+      - name: lint imports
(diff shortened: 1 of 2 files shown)


.github/workflows/go-ci.yml
```diff
@@ -47,6 +47,24 @@ jobs:
...
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version: ${{ inputs.go-version }}
+
+      - name: install goimports-reviser
+        run: go install -v github.com/incu6us/goimports-reviser/v3@latest
```
cristaloleg, line 63: Why this linter? Are we using it at all?

cristaloleg: Also, I see install but no start.

walldiss: >Why this linter? Are we using it at all?

It is a linter, that checks the `make sort-imports` command has been ran on the repo. And util is a part of `make lint` command, that is used to check lint rules locally. 

> Also, I see install but no start.

It is run in the following command: `make lint-imports`

walldiss: For the context: we use this util, because 
- golangci-lint imports order linter does not have `fix` option to quickly order of the imports the way we like. This util provides both `linter` and `fmt`
- it is a bit stricker than golangci-lint linter and groups imports into 4 groups: 
    - std
    - local repo
    -  same org
    -  external orgs

vgonkivs (approved):

.github/workflows/go-ci.yml
cristaloleg: >It is run in the following command: make lint-imports

```
lint: lint-imports
	@echo "--> Running linter"
	@golangci-lint run
	@markdownlint --config .markdownlint.yaml '**/*.md'
	@cfmt --m=120 ./...
.PHONY: lint
```

huh?

walldiss: First line embeds `lint-imports`:
```
lint: lint-imports
```

cristaloleg: Oh, thanks.

cristaloleg (approved):

PR #3678, opened 2024-08-22
chore(lint): bump golang-ci lint to 1.60

walldiss:
Bump linter and fix errors caused by upgrade.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -13,7 +13,7 @@ linters:
     # - funlen
     # - gochecknoglobals
     # - gochecknoinits
-    - exportloopref
+    - copyloopvar
     - gocheckcompilerdirectives
     - goconst
     - gocritic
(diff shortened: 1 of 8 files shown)


.golangci.yml
```diff
@@ -68,6 +67,9 @@ issues:
   max-same-issues: 50
 
 linters-settings:
+  gosec:
+    excludes:
+      - G115 # integer overflow conversion
```
cristaloleg, line 72: Is this a new check or what? Does it report something on our codebase?

cristaloleg (approved):

.golangci.yml
walldiss: This is a new rule that flags integer conversion overflow. We've got a lot of conversions that trigger it, but these are always false positives. I'm disabling it so we don't have to use nolint every time we convert between int and uint because:

- Our codebase never deals with values large enough to cause an int overflow.
- We're diligent enough in our reviews to catch incorrect conversions if they do happen.

Overall, this rule is more of a annoyance than a help.

vgonkivs (approved):

.golangci.yml
cristaloleg: Yeah, we also got a similar one https://github.com/go-critic/go-critic/pull/1431

Thanks!

PR #3679, opened 2024-08-23
misc(shwap/store): miscelaneous improvements 

walldiss:
Contains five commits with various bug fixes and improvements.

The PR is meant to be merged as 5 separate commits grouped under a common theme.

### DIFF
--- a/store/file/header.go
+++ b/store/file/header.go
@@ -8,16 +8,14 @@ import (
 	"github.com/celestiaorg/celestia-node/share"
 )
 
+const headerVOSize int = 40
+
 type headerVersion uint8
 
-const (
-	headerVersionV0 headerVersion = 1
-	headerVOSize    int           = 40
-)
+const headerVersionV0 headerVersion = iota + 1
 
 type headerV0 struct {
 	fileVersion fileVersion
-	fileType    fileType
 
 	// Taken directly from EDS
 	shareSize  uint16
@@ -30,13 +28,6 @@ type fileVersion uint8
 
 const (
 	fileV0 fileVersion = iota + 1
-)
-
-type fileType uint8
-
-const (
-	ods fileType = iota
-	q4
 )
 
 func readHeader(r io.Reader) (*headerV0, error) {
@@ -96,7 +87,6 @@ func (h *headerV0) OffsetWithRoots() int {
 func (h *headerV0) WriteTo(w io.Writer) (int64, error) {
 	buf := make([]byte, headerVOSize)
 	buf[0] = byte(h.fileVersion)
-	buf[1] = byte(h.fileType)
 	binary.LittleEndian.PutUint16(buf[4:6], h.shareSize)
 	binary.LittleEndian.PutUint16(buf[6:8], h.squareSize)
 	copy(buf[8:40], h.datahash)
@@ -112,7 +102,6 @@ func (h *headerV0) ReadFrom(r io.Reader) (int64, error) {
 	}
 
 	h.fileVersion = fileVersion(bytesHeader[0])
-	h.fileType = fileType(bytesHeader[1])
 	h.shareSize = binary.LittleEndian.Uint16(bytesHeader[4:6])
 	h.squareSize = binary.LittleEndian.Uint16(bytesHeader[6:8])
 	h.datahash = bytesHeader[8:40]
--- a/store/file/header.go
+++ b/store/file/header.go
@@ -78,12 +78,8 @@ func (h *headerV0) RootsSize() int {
 	return share.AxisRootSize * h.SquareSize() * 2
 }
 
-func (h *headerV0) Offset() int {
-	return h.Size()
-}
-
 func (h *headerV0) OffsetWithRoots() int {
-	return h.RootsSize() + h.Offset()
+	return h.RootsSize() + h.Size()
 }
 
 func (h *headerV0) WriteTo(w io.Writer) (int64, error) {
--- a/store/file/header.go
+++ b/store/file/header.go
@@ -8,7 +8,9 @@ import (
 	"github.com/celestiaorg/celestia-node/share"
 )
 
-const headerVOSize int = 40
+// headerVOSize is the size of the headerV0 in bytes. It has more space than the headerV0 struct
+// to allow for future extensions of the header without breaking compatibility.
+const headerVOSize int = 64
 
 type headerVersion uint8
 
@@ -87,9 +89,9 @@ func (h *headerV0) OffsetWithRoots() int {
 func (h *headerV0) WriteTo(w io.Writer) (int64, error) {
 	buf := make([]byte, headerVOSize)
 	buf[0] = byte(h.fileVersion)
-	binary.LittleEndian.PutUint16(buf[4:6], h.shareSize)
-	binary.LittleEndian.PutUint16(buf[6:8], h.squareSize)
-	copy(buf[8:40], h.datahash)
+	binary.LittleEndian.PutUint16(buf[28:30], h.shareSize)
+	binary.LittleEndian.PutUint16(buf[30:32], h.squareSize)
+	copy(buf[32:64], h.datahash)
 	n, err := w.Write(buf)
 	return int64(n), err
 }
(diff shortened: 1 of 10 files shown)


store/file/ods_q4.go
```diff
@@ -69,32 +70,35 @@ func OpenODSQ4(pathODS, pathQ4 string) (*ODSQ4, error) {
...
+
+	q4, err := OpenQ4(odsq4.pathQ4)
+	if errors.Is(err, os.ErrNotExist) {
+		return
 	}
 
-	q4, err := odsq4.q4Open()
 	if err != nil {
-		return nil, fmt.Errorf("failed to open Q4: %w", err)
+		log.Errorf("opening Q4 file %s: %s", odsq4.pathQ4, err)
```
Wondertan, line 89: shouldn't we handle those?

store/store.go
```diff
@@ -228,34 +223,42 @@ func (s *Store) GetByHeight(ctx context.Context, height uint64) (eds.AccessorStr
...
 
-// openODSQ4 opens ODSQ4 Accessor.
+// openFile opens ODSQ4 or ODS Accessor.
 // It opens ODS file first, reads up its DataHash and constructs the path for Q4
 // This done as Q4 is not indexed(hard-linked) and there is no other way to Q4 by height only.
-func (s *Store) openODSQ4(path string) (eds.AccessorStreamer, error) {
-	odsq4, err := file.OpenODSQ4(path+odsFileExt, path+q4FileExt)
-	switch {
-	case errors.Is(err, os.ErrNotExist):
+func (s *Store) openODSQ4(ctx context.Context, path string) (eds.AccessorStreamer, error) {
```
Wondertan, line 244: name and comment are inconsistent.

`openAcccesor`?

store/file/q4.go
```diff
@@ -102,17 +89,12 @@ func writeQ4(w io.Writer, eds *rsmt2d.ExtendedDataSquare) error {
 // of the File is read lazily.
 // If file is empty, the ErrEmptyFile is returned.
 // File must be closed after usage.
-func OpenQ4(path string) (*Q4, error) {
+func OpenQ4(path string, hdr *headerV0) (*Q4, error) {
```
Wondertan, line 92: It feels wrong to pass private header. I propose to either:
* Make Header public, doesn't hurt.
* Make OpenQ4 private
  * Then stop implementing Accessor for it. It's not needed anymore as its not usable on its own.

store/file/ods_q4.go
walldiss: If we fail to open Q4, we still allow reader to read from Q1

store/file/header.go
```diff
@@ -8,16 +8,14 @@ import (
 	"github.com/celestiaorg/celestia-node/share"
 )
 
+const headerVOSize int = 40
```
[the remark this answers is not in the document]
Wondertan: Sync: agreed to preallocate 64 bytes for future header additional fields

store/file/q4.go
walldiss: implemented option 2

store/file/header.go
```diff
@@ -8,16 +8,14 @@ import (
 	"github.com/celestiaorg/celestia-node/share"
 )
 
+const headerVOSize int = 64
```
Wondertan, line 11: There must be a note explaining why there is more data than needed

store/file/ods_q4.go
```diff
@@ -60,41 +61,36 @@ func CreateODSQ4(
...
 	if odsq4.q4Opened.Load() {
-		return odsq4.q4, nil
+		return
 	}
 
 	odsq4.q4Mu.Lock()
 	defer odsq4.q4Mu.Unlock()
-	if odsq4.q4Opened.Load() {
-		return odsq4.q4, nil
+	if loaded := odsq4.q4Opened.Swap(true); loaded {
```
Wondertan, line 79: Imo, this needs a note anyway on why its Swap and then we can avoid creating loaded

store/file/q4.go
```diff
@@ -141,30 +119,6 @@ func (q4 *Q4) AxisHalf(ctx context.Context, axisType rsmt2d.Axis, axisIdx int) (
 	}, nil
 }
 
-func (q4 *Q4) Size(context.Context) int {
+func (q4 *q4) Size(context.Context) int {
```
Wondertan, line 122: Just let it access directly from header method

store/file/q4.go
```diff
@@ -97,33 +80,28 @@ func writeQ4(w io.Writer, eds *rsmt2d.ExtendedDataSquare) error {
 	return nil
 }
 
-// OpenQ4 opens an existing Q4 file under given FS path.
+// openQ4 opens an existing Q4 file under given FS path.
```
Wondertan, line 83: Needs an update that it doesnt read the header

store/file/q4.go
```diff
@@ -10,29 +10,24 @@ import (
...
 
-	"github.com/celestiaorg/celestia-node/share"
 	eds "github.com/celestiaorg/celestia-node/share/new_eds"
-	"github.com/celestiaorg/celestia-node/share/shwap"
 )
 
-var _ eds.AccessorStreamer = (*Q4)(nil)
-
-// Q4 implements eds.Accessor as an FS file.
+// q4 implements eds.Accessor as an FS file.
```
Wondertan, line 16: It doesnt implement anymore

store/file/ods_q4.go
```diff
@@ -24,10 +25,10 @@ var _ eds.AccessorStreamer = (*ODSQ4)(nil)
 type ODSQ4 struct {
 	ods *ODS
 
-	q4Open   func() (*Q4, error)
 	q4Mu     sync.Mutex
 	q4Opened atomic.Bool
-	q4       *Q4
+	pathQ4   string
```
Wondertan, line 30: Uber nit: put above mutex, as mutex doesnt protect it.

store/file/ods_q4.go
```diff
@@ -125,22 +121,15 @@ func (odsq4 *ODSQ4) Sample(ctx context.Context, rowIdx, colIdx int) (shwap.Sampl
 func (odsq4 *ODSQ4) AxisHalf(ctx context.Context, axisType rsmt2d.Axis, axisIdx int) (eds.AxisHalf, error) {
 	size := odsq4.Size(ctx) // TODO(@Wondertan): Should return error.
 
-	var acsr eds.Accessor = odsq4.ods
 	if axisIdx >= size/2 {
-		q4, err := odsq4.getQ4()
-		if err != nil {
-			return eds.AxisHalf{}, err
+		// lazy load Q4 file and read axis from it if loaded
+		odsq4.tryLoadQ4()
```
Wondertan, line 126: Readability nit: i would still make it return q4, so it doesn't confuse readers with accessing Q4 field without synchronization, while it has it on the struct. Returning it makes it seem like the tryLoadq4 takes full responsibility on proper synchornization

store/file/q4.go
Wondertan: Actually, that size check isn't really needed there, as it happens on ODSQ4 and Q4 isnt general anymore

store/file/ods_q4.go
```diff
@@ -60,41 +63,39 @@ func CreateODSQ4(
...
 	if odsq4.q4Opened.Load() {
-		return odsq4.q4, nil
+		return odsq4.q4
 	}
 
 	odsq4.q4Mu.Lock()
 	defer odsq4.q4Mu.Unlock()
-	if odsq4.q4Opened.Load() {
-		return odsq4.q4, nil
+	// check again in case another goroutine opened the file before we acquired the lock.
```
Wondertan, line 82:
```suggestion
	// update bool to make sure we try opening only once
	// no matter if the try was successful or failed.
```

store/file/ods_q4.go
```diff
@@ -61,21 +63,15 @@ func CreateODSQ4(
...
 
-// OpenODSQ4 opens ODS file under the given FS path. The Q4 is opened lazily
-// on demand.
-func OpenODSQ4(pathODS, pathQ4 string) (*ODSQ4, error) {
-	ods, err := OpenODS(pathODS)
-	if err != nil {
-		return nil, fmt.Errorf("failed to open ODS: %w", err)
-	}
-
+// ODSWithQ4 returns ODSQ4 instance for the given FS paths. It opens Q4 file lazily.
```
Wondertan, line 66:
```suggestion
// ODSWithQ4 returns ODSQ4 instance over ODS. It opens Q4 file lazily under the given path.
```

Wondertan (commented):
Can also remove `Offset` method from header in the last commit

PR #3680, opened 2024-08-23
feat(shwap/store): Add q4 trimming support in store

walldiss:


store/metrics.go
```diff
@@ -130,15 +138,27 @@ func (m *metrics) observeHas(ctx context.Context, dur time.Duration, failed bool
 		attribute.Bool(failedKey, failed)))
 }
 
-func (m *metrics) observeRemove(ctx context.Context, dur time.Duration, failed bool) {
+func (m *metrics) observeRemoveAll(ctx context.Context, dur time.Duration, failed bool) {
+	if m == nil {
+		return
+	}
+	if ctx.Err() != nil {
+		ctx = context.Background()
```
cristaloleg, line 146: So, always blocking? Maybe `context.WithTimeout(context.Background(), time.Second)` ?

walldiss: Good that you pointed out. We use common function to reset context in metrics everywhere else. I'll change how we do it in another PR.

store/store.go
```diff
@@ -305,51 +304,76 @@ func (s *Store) hasByHeight(height uint64) (bool, error) {
 	return exists(pathODS)
 }
 
-func (s *Store) Remove(ctx context.Context, height uint64, datahash share.DataHash) error {
+func (s *Store) RemoveAll(ctx context.Context, height uint64, datahash share.DataHash) error {
```
Wondertan, line 307: Let's name to `RemoveODSQ4`. Its not clear what All stands for. 

Same for helpers methods and metrics.

store/store.go
```diff
@@ -33,7 +33,7 @@ const (
 var ErrNotFound = errors.New("eds not found in store")
 
 // Store is a storage for EDS files. It persists EDS files on disk in form of Q1Q4 files or ODS
-// files. It provides methods to put, get and remove EDS files. It has two caches: recent eds cache
+// files. It provides methods to put, get and removeAll EDS files. It has two caches: recent eds cache
```
Wondertan, line 36: rename relic

PR #3692, opened 2024-08-26
fix(ci): detect breaking ci by comparing with base branch instead of main

walldiss:
Refactor the action to:
- compare changes with target branch instead of main
- detect base branch changes to trigger ci
- don't apply kind:break! if breaking changes are detected, but there is label already
- apply kind:break! only if it is not present 

TODO: tokens need permissions to write to pull requests to allow this action to set up the label

### DIFF
--- a/share/p2p/shrexeds/pb/extended_data_square.proto
+++ b/share/p2p/shrexeds/pb/extended_data_square.proto
@@ -11,6 +11,7 @@ enum Status {
   INTERNAL = 3; // internal server error
 }
 
+FAKE BREAK
 message EDSResponse {
   Status status = 1;
 }


walldiss:
@MSevey could you please check the GitHub token access to ensure it has write privileges for pull requests?

.github/workflows/labels.yml
```diff
@@ -1,49 +1,83 @@
...
   pull_request:
-    types: [opened, labeled, unlabeled, synchronize]
+    types: [opened, labeled, unlabeled, synchronize, edited]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
 
 jobs:
-  label:
```
MSevey, line 12: fwiw this is now blocking the PR as this was the name of the required CI. 
So you'll need to update the branch protection rules to merge this PR.

.github/workflows/labels.yml
```diff
@@ -1,49 +1,83 @@
...
+          BASE_BRANCH: ${{ github.event.pull_request.base.ref }}
+        continue-on-error: true  # This allows the job to continue even if breaking changes are detected
         run: |
-          git fetch origin main
+          echo "Base Branch: $BASE_BRANCH"
+          git fetch origin "$BASE_BRANCH"
           make detect-breaking
 
-      - name: Add label if breaking changes detected
+      - name: Check if 'kind:break!' label is already present
```
MSevey, line 59: This might be useful https://github.com/egmacke/action-check-label

.github/workflows/labels.yml
```diff
@@ -1,49 +1,83 @@
...
+          else
+            echo "label_exists=false" >> $GITHUB_ENV
+          fi
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          GH_REPO: ${{ github.repository }}
-          NUMBER: ${{ github.event.pull_request.number }}
-          LABELS: kind:break!
+
+      - name: Add 'kind:break!' label if breaking changes detected
```
MSevey, line 73: this might be useful https://github.com/marketplace/actions/actions-ecosystem-add-labels

walldiss: I've tried to use it, but it didn't work for some reason. Perhaps, you might want to contribute to this PR?

.github/workflows/labels.yml
```diff
@@ -1,49 +1,83 @@
...
         if: failure()
-        run: gh issue edit "$NUMBER" --add-label "$LABELS"
+        run: |
+          LABEL_EXISTS=$(gh pr view ${{ github.event.pull_request.number }} --json labels --jq '.labels | map(.name) | contains(["kind:break!"])')
+          echo "Label exists: $LABEL_EXISTS"
+          if [[ "$LABEL_EXISTS" == "true" ]]; then
+            echo "label_exists=true" >> $GITHUB_ENV
+          else
+            echo "label_exists=false" >> $GITHUB_ENV
+          fi
```
MSevey, line 69: Did this not work?
```suggestion
         echo "label_exists=${{ $LABEL_EXISTS }}" >> $GITHUB_ENV
```

PR #756, opened 2022-05-24
Oblitirate test flakiness

Wondertan:
Fixes #683
Fixes #467 

Also, #736 did a good job for us and:
Closes https://github.com/celestiaorg/celestia-node/issues/582 
Closes https://github.com/celestiaorg/celestia-node/issues/608
Hopefully closes https://github.com/celestiaorg/celestia-node/issues/584
Hopefully closes https://github.com/celestiaorg/celestia-node/issues/579

Bonus: https://github.com/celestiaorg/celestia-node/issues/476
(it does not really fix this #476, but it fixes the flakeyness that was mixed up with #476, while #476 is another one that was not reproduced for a long time, and I assume it is now fixed)

PR #3699, opened 2024-08-28
fix(share/discovery): fix TestDiscovery test

walldiss:
TestDiscovery has been increasingly unstable lately. This fix aims to make it stable


Wondertan (approved):
Gives me #756 vibes

vgonkivs (approved):

PR #3700, opened 2024-08-29
chore(ci): run unit-tests in parallel with lint

walldiss:
Currently `unit-tests` ci is starting after `lint` succeeded. Given that both checks are `required` and  each of those steps takes around 3 min to pass, we can make them run in parallel to speed up ci time.


cristaloleg (approved):
SGTM

Wondertan (approved):

PR #3702, opened 2024-08-29
feat(metrics): export prometheus metrics to otel

walldiss:
Instead of running prometheus server, send prometheus collected metrics by otel exporter


Wondertan:
Did they finally implemented support for running prom metrics ovet otel??

PR #3709, opened 2024-09-03
[WIP]feat(shwap): add RangeNamespaceDataID along with the RangeNamespaceData

vgonkivs:


share/shwap/namespace_data.go
```diff
@@ -78,3 +80,12 @@ func (nd NamespaceData) WriteTo(writer io.Writer) (int64, error) {
 	}
 	return n, nil
 }
+
+// Proofs iterates over each row and collects proofs.
+func (nd NamespaceData) Proofs() []*nmt.Proof {
```
Wondertan, line 85: nit: move next to Flatten

share/shwap/range_namespace_data_id.go
```diff
@@ -0,0 +1,141 @@
...
+		return fmt.Errorf("RangeNamespaceDataID: sample id verification: %w", err)
+	}
+
+	endIndex := uint16(rngid.SampleID.ShareIndex) + rngid.Length
+
+	// check that end index is not bigger than amount of shares in the ODS.
+	if endIndex > uint16(edsSize) {
+		return errors.New("RangeNamespaceDataID: end index exceeds ODS length")
+	}
+	return nil
```
Wondertan, line 117: Verify should call Validate

share/shwap/range_namespace_data_id.go
```diff
@@ -0,0 +1,141 @@
...
+			"invalid RangeNamespaceDataID data length: expected %d, got %d", RangeNamespaceDataIDSize, len(data),
+		)
+	}
+
+	sid, err := SampleIDFromBinary(data[:SampleIDSize])
+	if err != nil {
+		return RangeNamespaceDataID{}, fmt.Errorf("converting SampleId from binary: %w", err)
+	}
+
+	return RangeNamespaceDataID{
```
Wondertan, line 62: Should call Validate as well

share/shwap/range_namespace_data.go
```diff
@@ -0,0 +1,160 @@
...
+	rowStart := req.RowIndex
+	for i, row := range rngdata.data {
+		verified := row.Proof.VerifyInclusion(
+			share.NewSHA256Hasher(),
+			req.RangeNamespace.ToNMT(),
+			row.Shares,
+			root.RowRoots[rowStart+i],
+		)
+		if !verified {
+			return fmt.Errorf("RangeNamespaceData: proof verification failed at row: %d", rowStart+i)
```
Wondertan, line 134: Should wrap `ErrVerificationFailed`

share/shwap/range_namespace_data_id.go
```diff
@@ -0,0 +1,141 @@
...
+const RangeNamespaceDataIDSize = SampleIDSize + share.NamespaceSize + 2 + 2
+
+// RangeNamespaceDataID identifies the continuous range of shares in the DataSquare(EDS),
+// starting from the given `SampleID` and contains `Length` number of shares.
+type RangeNamespaceDataID struct {
+	SampleID
+	//  RangeNamespace is a string representation of the namespace of the requested range.
+	RangeNamespace share.Namespace
+	// Length is a number of shares to of the requested range.
+	Length uint16
```
Wondertan, line 23: In other shwap types we use int in ID struct and convert to the wire type in marshal functions

Wondertan: Also, do you think uint16 is enough? Why not uint32?

share/shwap/range_namespace_data.go
```diff
@@ -0,0 +1,160 @@
...
+// rawShares is a list of shares relative to the data square needed to build the range;
+// namespace is the target namespace for the built range;
+// rowIndex is the starting index of the list relative to the data square;
+// from is the share index of the first share of the range;
+// to is an EXCLUSIVE index of the last share of the range.
+func RangedNamespaceDataFromShares(
+	rawShares [][]share.Share,
+	namespace share.Namespace,
+	rowIndex, from, to int,
+) (NamespaceData, error) {
```
Wondertan, line 24: Why not make it return RangeNamespaceData? To avoid handling proofs only in it?

Wondertan (commented):
First look. The thing which is still unclear to me is why we need to have Sample in the container and why NamespaceData isn't enough

share/shwap/row_namespace_data.go
```diff
@@ -84,16 +84,13 @@ func RowNamespaceDataFromShares(
 		}, nil
 	}
 
-	namespacedShares := make([]share.Share, count)
-	copy(namespacedShares, shares[from:from+count])
-
```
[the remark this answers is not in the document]
vgonkivs: Discussed internally with @walldiss. No reason for copying.

vgonkivs:
> First look. The thing which is still unclear to me is why we need to have Sample in the container and why NamespaceData isn't enough

We need to additionally check the `from` share, to ensure that the whole range is valid. Since the request contains SampleID, I decided to return Sample as well. The other way is to build a sub-range proof for the `from` share and verify(but the nmt hasn't supported it)

share/shwap/range_namespace_data_id.go
vgonkivs: I think uint32 is better. Thanks.

share/shwap/range_namespace_data.go
vgonkivs: Yes and to avoid passing SampleID as it's a part of `RangeNamespaceData`

PR #3723, opened 2024-09-12
chore: use sync/atomic types

### LINKED ISSUE #1048: all: swap all the `int` atomic accesses with atomics introduced in go1.19

cristaloleg:
Fixes #1048


Wondertan (dismissed):

das/daser.go
```diff
@@ -41,7 +41,7 @@ type DASer struct {
 
 	cancel         context.CancelFunc
 	subscriberDone chan struct{}
-	running        int32
```
walldiss, line 44: Do you mind to change this one to use atomic.Bool? It suites the usecase better here

cristaloleg: Pretty valid suggestion, fixed & thanks!

Wondertan (approved):

walldiss (approved):

PR #3724, opened 2024-09-13
feat(shwap): check ID equality

Wondertan:
We need to check that requested ID is equal to received ID.
This was the case, but implicitly in the Bitswap layer.
This change adds another explicit check for that, ensuring IDs are printed on error

### DIFF
--- a/share/shwap/p2p/bitswap/cid.go
+++ b/share/shwap/p2p/bitswap/cid.go
@@ -8,18 +8,18 @@ import (
 	mh "github.com/multiformats/go-multihash"
 )
 
-// extractCID retrieves Shwap ID out of the CID.
-func extractCID(cid cid.Cid) ([]byte, error) {
+// extractFromCID retrieves Shwap ID out of the CID.
+func extractFromCID(cid cid.Cid) ([]byte, error) {
 	if err := validateCID(cid); err != nil {
-		return nil, err
+		return nil, fmt.Errorf("invalid cid %s: %w", cid, err)
 	}
 	// mhPrefixSize is the size of the multihash prefix that used to cut it off.
 	const mhPrefixSize = 4
 	return cid.Hash()[mhPrefixSize:], nil
 }
 
-// encodeCID encodes Shwap ID into the CID.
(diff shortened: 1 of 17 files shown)


share/shwap/p2p/bitswap/sample_block.go
```diff
@@ -96,19 +96,28 @@ func (sb *SampleBlock) Populate(ctx context.Context, eds eds.Accessor) error {
 }
 
 func (sb *SampleBlock) UnmarshalFn(root *share.AxisRoots) UnmarshalFn {
-	return func(data []byte) error {
+	return func(cntrData, idData []byte) error {
 		if !sb.Container.IsEmpty() {
```
cristaloleg, line 100: Why it should be empty before unmarshaling? We don't want to overwrite the current data or what?

cristaloleg (approved):
SGTM

share/shwap/p2p/bitswap/sample_block.go
Wondertan: Yes, there are possible edge cases in tests where this function is called twice for the same data

cristaloleg: Is this really a problem?

share/shwap/p2p/bitswap/cid.go
```diff
@@ -8,18 +8,18 @@ import (
 	mh "github.com/multiformats/go-multihash"
 )
 
-// extractCID retrieves Shwap ID out of the CID.
-func extractCID(cid cid.Cid) ([]byte, error) {
+// extractFromCID retrieves Shwap ID out of the CID.
+func extractFromCID(cid cid.Cid) ([]byte, error) {
 	if err := validateCID(cid); err != nil {
-		return nil, err
+		return nil, fmt.Errorf("invalid cid %q: %w", cid, err)
```
walldiss, line 14: why %q?

walldiss (approved):

share/shwap/p2p/bitswap/cid.go
Wondertan: Good catch

PR #3736, opened 2024-09-17
fix(store/ods): fix full ods caching

walldiss:
There was a race condition for reseting the offset in file. This change fixes the race and removes the need for global offset by making it local to SectionReader


Wondertan (approved):
makes sense

PR #3739, opened 2024-09-17
feat(nodebuilder): Blockstore cache params

walldiss:
Expose Blockstore cache size to nodebuilder config.

### DIFF
--- a/nodebuilder/share/bitswap.go
+++ b/nodebuilder/share/bitswap.go
@@ -17,12 +17,6 @@ import (
 	"github.com/celestiaorg/celestia-node/store"
 )
 
-const (
-	// TODO(@walldiss): expose cache size to cfg
-	// default blockstore cache size
-	defaultBlockstoreCacheSize = 128
-)
-
 // dataExchange constructs Exchange(Bitswap Composition) for Shwap
 func dataExchange(tp node.Type, params bitSwapParams) exchange.SessionExchange {
 	prefix := protocolID(params.Net)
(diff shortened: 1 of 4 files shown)


nodebuilder/share/bitswap.go
```diff
@@ -58,8 +52,12 @@ func blockstoreFromDatastore(ds datastore.Batching) (blockstore.Blockstore, erro
...
 }
 
-func blockstoreFromEDSStore(store *store.Store) (blockstore.Blockstore, error) {
-	withCache, err := store.WithCache("blockstore", defaultBlockstoreCacheSize)
+func blockstoreFromEDSStore(store *store.Store, blockStoreCacheSize int) (blockstore.Blockstore, error) {
+	if blockStoreCacheSize == 0 {
+		// no cache, return plain blockstore
+		return &bitswap.Blockstore{Getter: store}, nil
+	}
+	withCache, err := store.WithCache("blockstore", blockStoreCacheSize)
```
Wondertan, line 60: Also, might be a special case on WithCache that returns itself.

Wondertan (approved):

nodebuilder/share/bitswap.go
Wondertan: Ok, CachedStore is not an interface, so won't work

walldiss: Good point. Perhaps user should not be allowed call WithCache with size <= 0. I've added the check for that

Wondertan (approved):

PR #3742, opened 2024-09-18
fix(blob/service): return error from GetProof if it appears

### LINKED ISSUE #3726: blob: Service.GetProof mistakenly swallows up error from Service.retrieve
If we examine this code https://github.com/celestiaorg/celestia-node/blob/1a1286f454c1dd670af856153a8fc8674cd91192/blob/service.go#L237-L238 we can see something peculiar there that err is set but down below, there is a `return proof, nil` of which due to the use of named returns in Go, even if there was a variable name clash. Please return that value and test that code more: if it was intended, please document it instead.

/cc @liamsi @rootulp @musalbas

vgonkivs:
Resolves #3726


blob/service.go
```diff
@@ -235,7 +235,7 @@ func (s *Service) GetProof(
 	}}
 
 	_, proof, err = s.retrieve(ctx, height, namespace, sharesParser)
-	return proof, nil
```
cristaloleg, line 238: nononon, no bare return please 😢

vgonkivs: ok

Wondertan (approved):

walldiss (approved):

PR #3744, opened 2024-09-18
feat(modrpc): list pubsub topics RPC

Wondertan:
Really handy for when you wanna debug GossipSub things.  Dunno why we didn't have it earlier


vgonkivs (approved):

nodebuilder/p2p/cmd/p2p.go
```diff
@@ -574,3 +575,27 @@ var pubsubPeersCmd = &cobra.Command{
 		return cmdnode.PrintOutput(peers, err, formatter)
 	},
 }
+
+var pubsubTopicsCmd = &cobra.Command{
+	Use:   "pubsub-topics ",
```
cristaloleg, line 580: space is intentional, right?

cristaloleg (approved):
SGTM

nodebuilder/p2p/cmd/p2p.go
Wondertan: No. But it works with it, suprisingly

PR #3747, opened 2024-09-18
fix(store): use symlinks for empty blocks heighs

walldiss:
There is limitation on max amount of hard links in ext4 filesystem of 65000. Some chains have more empty blocks (arabica). We will use symlinks for empty blocks to avoid this limitation, while keeping hard-links for non-empty blocks.

### DIFF
--- a/store/store.go
+++ b/store/store.go
@@ -174,7 +174,12 @@ func (s *Store) linkHeight(datahash share.DataHash, height uint64) error {
 	// create hard link with height as name
 	pathOds := s.hashToPath(datahash, odsFileExt)
 	linktoOds := s.heightToPath(height, odsFileExt)
-	return link(pathOds, linktoOds)
+	if datahash.IsEmptyEDS() {
+		// empty EDS is always symlinked, because there is limited number of hardlinks
+		// for the same file in some filesystems (ext4)
+		return symlink(pathOds, linktoOds)
+	}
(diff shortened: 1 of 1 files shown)


store/store.go
```diff
@@ -170,11 +170,18 @@ func (s *Store) createFile(
...
+func (s *Store) hardLinkHeight(datahash share.DataHash, height uint64) error {
 	// create hard link with height as name
 	pathOds := s.hashToPath(datahash, odsFileExt)
 	linktoOds := s.heightToPath(height, odsFileExt)
-	return link(pathOds, linktoOds)
+	return hardLink(pathOds, linktoOds)
+}
+
+func (s *Store) symLinkHeight(datahash share.DataHash, height uint64) error {
+	// create hard link with height as name
```
Wondertan, line 181: Update comment

store/store.go
```diff
@@ -170,11 +170,18 @@ func (s *Store) createFile(
 	return false, nil
 }
 
-func (s *Store) linkHeight(datahash share.DataHash, height uint64) error {
```
Wondertan, line 173: What about keeping this method usage as is, but change it to check for empty root and if so make it symlink with proper comment. In other words, decision on whether its symlink or hardlink stays an impl detail of linkHeight

Wondertan (approved):

PR #3750, opened 2024-09-19
feat(nodebuilder/da): implement go-da v0.6.1 changes

### LINKED ISSUE #3743: feat: Update to go-da v0.6.x
## Overview
New go-da release contains breaking changes and is not backward compatible. All details are tracked in the [epic](https://github.com/rollkit/go-da/issues/92).

The most important change from the celestia-node perspective is change of return type of `GetIDs` method (not backward compatible!) and addition of `SubmitWithOptions`. Both changes can be implemented by composition of existing functionality. 

### `GetIDs`
`GetIDs` implementation currently depends only on blob module - there is no direct way of accessing header. The most intuitive way of resolving this is to modify `da.Service` to inject header getter function. It can be then used to get block time information.

### `SubmitWithOptions`
`SubmitWithOption` was introduced to enable the usage of multi-account feature. Because of generic nature of go-da, the new `options` argument has to be parsed and interpreted by implementation. To keep things simple yet extensible, `options` value should be a string containing JSON object with keys corresponding to available options, like: `KeyName`, `SignerAddress`, `FeeGranterAddress`. For example:
```json
{"KeyName": "secret_key"}
```
Those options can be trivially added to `NewTxConfig` call in `da.Service`.

tzdybal:
Resolves #3743


nodebuilder/tests/da_test.go
```diff
@@ -105,11 +104,13 @@ func TestDaModule(t *testing.T) {
 		{
 			name: "GetIDs",
 			doFn: func(t *testing.T) {
-				t.Skip()
```
cristaloleg, line 108: I'm curious why it was skipped before. Anyone knows?

cristaloleg (dismissed):
SGTM

nodebuilder/tests/da_test.go
tzdybal: IIRC order of returned IDs was non-deterministic in the past.

vgonkivs (approved):
LGTM 🚀

nodebuilder/da/service.go
```diff
@@ -126,6 +152,36 @@ func (s *Service) Submit(
 	return ids, nil
 }
 
+func parseOptions(options []byte) ([]state.ConfigOption, error) {
+	var opts []state.ConfigOption
+
+	parsedOpts := struct {
```
renaynay, line 158: doesn't this need json tags?

tzdybal: No, we're using defaults, so it's just the name. 
See: https://go.dev/play/p/I4925joE12fhttps://go.dev/play/p/I4925joE12f

nodebuilder/da/service.go
```diff
@@ -126,6 +152,36 @@ func (s *Service) Submit(
...
+		KeyName           string
+		SignerAddress     string
+		FeeGranterAddress string
+	}{}
```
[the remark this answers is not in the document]
tzdybal: This options are implementation specific. It doesn't even have to be JSON in other implementations of go-da.
It's also up to implementation to specify supported options, which can be very different from the ones defined in celestia-node. That's why we don't define them in go-da.

We don't care about encoding to JSON as here we only parse (using reasonable defaults) - rollup operator will prepare actual value of `options` as a part of a rollkit config or start script.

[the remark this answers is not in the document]
walldiss: Does it imply, that client should be aware of implementation specific options when using generic go-da interface? 
There could be few problems with this approach:
- different name for same options in different implementations
- client will need to depend on implementation specific encoding of options instead of interface definition only.

If go-da does not define encoding, we should define it in node as separate struct with json annotation tags for fields. Otherwise other clients (non go-da) won't be able to properly encode options.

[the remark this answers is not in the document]
tzdybal: Let's clarify entire setup a bit.

1. Rollkit is a client of go-da API. Rollkit is DA implementation agnostic.
2. celestia-node is one of go-da implementations, acting as server of go-da API.
3. Each time a user configures rollkit to run a rollup, he needs to setup DA - he knows:
  - the type of DA (celestia-node, avail, bitcoin, etc)
  - endpoint address
  - options to use (for example: key name for this concrete instance of celestia-node that he wish to use) 

Typical real-world usage for `SubmitWithOption` is RaaS setup - RaaS provider runs a single celestia-node with multiple keys in keyring. Rollkit rollup is then configured to pass `{"KeyName": "value agreed with RaaS provider"}` as `--rollkit.da_submit_options`. 
Those options are not interpreted or parsed by Rollkit, it's just passed to go-da implementation - any parsing errors are returned by go-da server/implementation.

This change only applies to go-da implementation, so I don't expect "other clients".  But I agree that this should be documented somewhere - this will be definitely a part of rollkit doc, but if there is a good place to document this in celestia-node please let me know.

I can also refactor so the struct is exported and properly described in godoc.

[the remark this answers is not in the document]
renaynay: I'd prefer the latter (export + describe in godoc).

renaynay (dismissed):
I'm okay with this

nodebuilder/da/service.go
```diff
@@ -28,12 +29,32 @@ var log = logging.Logger("go-da")
...
+	// KeyName is the name of key from the keystore used to sign transactions.
+	KeyName string
+
+	// SignerAddress is the address that signs the transaction.
+	// This address must be stored locally in the key store.
+	SignerAddress string
+
+	// FeeGranterAddress specifies the account that will pay for the transaction fee.
+	FeeGranterAddress string
+}
```
walldiss, line 49: Pls add json tags, so we can use them in API docs and other langs can implement options support

tzdybal: Added JSON tags, and used `snake_case` for consistency.

renaynay (approved):

Wondertan (approved):
utACK

cristaloleg:
Congrats @tzdybal ☝️ 😂

PR #3751, opened 2024-09-19
feat(share/availability): Dont store q4 on archival

walldiss:
Avoid storing q4 for data outside of availability window

### DIFF
--- a/store/store.go
+++ b/store/store.go
@@ -90,11 +90,30 @@ func (s *Store) Stop(context.Context) error {
 	return s.metrics.close()
 }
 
-func (s *Store) Put(
+func (s *Store) PutODSQ4(
 	ctx context.Context,
 	roots *share.AxisRoots,
 	height uint64,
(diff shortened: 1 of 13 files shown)


store/store.go
```diff
@@ -170,6 +195,39 @@ func (s *Store) createFile(
 	return false, nil
 }
 
+func (s *Store) createODSFile(
```
Wondertan, line 198: let's avoid having duplicate code by passing create function. The `createODSFile` and `createODSQ4File` are identical otherwise

Wondertan (commented):
One duplication nit, lgtm otherwise

store/store.go
walldiss: Not really identical, they createODSFile and createODSQ4File have different revert operation used (removeODS <-> removeODSQ4).

Wondertan (approved):

PR #3759, opened 2024-09-21
fix(node): fixup CELESTIA_HOME

Wondertan:
This mainly fixes the usage of `CELESTIA_HOME` env var. Previously, it worked like `XDG_HOME` in the form of `$CELESTIA_HOME/.celestia-<network>-<type>`. This change makes `$CELESTIA_HOME` the actual path to which node data goes.

This is promoted by me logging into the pods on Robusta and being tired of constantly passing the `--node.store` flag while `CELESTIA_HOME` is already set. 

---

Besides, it contains a minor refactor to use respective types for node and network instead of strings.

### DIFF
--- a/cmd/cel-key/node_types.go
+++ b/cmd/cel-key/node_types.go
@@ -10,6 +10,7 @@ import (
 	flag "github.com/spf13/pflag"
 
 	"github.com/celestiaorg/celestia-node/nodebuilder"
+	nodemod "github.com/celestiaorg/celestia-node/nodebuilder/node"
 	"github.com/celestiaorg/celestia-node/nodebuilder/p2p"
 )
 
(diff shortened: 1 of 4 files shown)


vgonkivs (approved):

cmd/cel-key/node_types.go
```diff
@@ -42,32 +43,27 @@ func ParseDirectoryFlags(cmd *cobra.Command) error {
 		return nil
 	}
 
-	nodeType := cmd.Flag(nodeDirKey).Value.String()
-	if nodeType == "" {
-		return errors.New("no node type provided")
+	nodeTypeStr := cmd.Flag(nodeDirKey).Value.String()
+	nodeType := nodemod.ParseType(nodeTypeStr)
+	if !nodeType.IsValid() {
+		return errors.New("no or invalid node type provided")
```
walldiss, line 49: Nit: could return err from ParseType to specify the exact case:
- no node type 
- invalid node type

walldiss (approved):

cmd/cel-key/node_types.go
Wondertan: I agree, but its very low value to spend cycles on rn

PR #3761, opened 2024-09-23
Chore(fix): set asynctimerchan=1 on build

walldiss:
Set go:debug directive  "asynctimerchan=1" until https://github.com/golang/go/issues/69312 is fixed.

### DIFF
--- a/cmd/celestia/main.go
+++ b/cmd/celestia/main.go
@@ -1,4 +1,3 @@
-//go:debug asynctimerchan=1
 package main
 
 import (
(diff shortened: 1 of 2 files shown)


Wondertan:
Let's do it in go.mod instead as per (https://go.dev/doc/godebug). The main file also works, but it's buried somewhere deep instead of top level go.mod

Wondertan (dismissed):

Wondertan (approved):

cristaloleg (approved):

vgonkivs (approved):

PR #3763, opened 2024-09-23
fix(nodebuilder/p2p/metrics): enforce init order for bitswap metrics

walldiss:
Bitswap metrics are initialised inside constructors and require prometheus to be injected before the call. This PR enforces initialisation order:
 1. set prometheus global Registrer
 2. inject bitswap metrics fabric
 3. set bitswap component scope though ctx
 4. init bitswap components


Wondertan (dismissed):

renaynay (approved):
utack

renaynay (approved):

Wondertan (approved):

PR #3774, opened 2024-09-25
fxi(nodebuilder/swamp): ignore leveldb panic in swamp tests

walldiss:
LevelDB may panic after test is over. This happens due to leveldb being closed before node client is closed. This is temporary fix to make swamp tests more stable until swamp tests shutdown order is fixed.


walldiss:
Not needed

PR #3775, opened 2024-09-25
fix(core/listener): wait listener to shutdown before exit

walldiss:
Previously we observed panic with `leveldb: closed` in integration tests. It was caused by core.listener accessing already closed core block generator. The reason for it could be leaked listener subscriber routine after node is closed, that was trying to access generator. 

The PR introduces graceful shutdown for listener, that will wait for spawned routines to stop before exiting.

### DIFF
--- a/core/listener.go
+++ b/core/listener.go
@@ -49,6 +49,7 @@ type Listener struct {
 
 	listenerTimeout time.Duration
 	cancel          context.CancelFunc
+	closed          chan struct{}
 }
 
 func NewListener(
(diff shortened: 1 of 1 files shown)


Wondertan (commented):
Please explain how it fixes the flake either in the comment or PR description for future ref

renaynay (approved):
Nit: Set chan to nil when done ? And make chan on start ?

Wondertan (approved):

renaynay (approved):

PR #3780, opened 2024-09-27
feat(cel-shed/p2p): bootstrapper probe tool

smuu:
Based on @Wondertan PR: https://github.com/celestiaorg/celestia-node/pull/3511

This PR introduces a new `connect-bootstrappers` command to the `cel-shed` tool. The command allows users to connect to bootstrappers of a specified network. It includes options to handle connection failures by returning errors if at least one or all bootstrappers are unreachable.

This will be used by the @celestiaorg/devops team to create alerts if no Bootstrapper is reachable.

Closes https://github.com/celestiaorg/celestia-node/pull/3511


Wondertan (approved):
Thank you for cleaning up my PR!

* Linter seems to complain
* Did you test new flags work as expected?

smuu:
> Thank you for cleaning up and improving that PR!
> 
>     * Linter seems to complain
> 
>     * Did you test new flags work as expected?

I tried the arguments manually with these scenarios.
- All Bootstrappers are reachable
- All Bootstrappers are not reachable
- At least one Bootstrapper is reachable

Do you see a way to write tests for these?

cmd/cel-shed/p2p.go
```diff
@@ -75,3 +81,99 @@ var p2pPeerIDCmd = &cobra.Command{
...
+func init() {
+	p2pConnectBootstrappersCmd.Flags().BoolVar(
+		&errorOnAnyFailure, "err-any", false,
+		"Return error if at least one bootstrapper is not reachable",
+	)
+	p2pConnectBootstrappersCmd.Flags().BoolVar(
+		&errorOnAllFailure, "err-all", false,
+		"Return error if no bootstrapper is reachable",
+	)
+}
```
walldiss, line 180: You may want to provide explicit timeout for the util either per connection or per bootsrapper. Otherwise it can potentially get stuck on connection attempt.

cmd/cel-shed/p2p.go
```diff
@@ -75,3 +81,99 @@ var p2pPeerIDCmd = &cobra.Command{
...
+				failedConnections++
+				continue
+			}
+			fmt.Println("connected")
+			successfulConnections++
+		}
+
+		if failedConnections == len(bootstrappers) && errorOnAllFailure {
+			fmt.Println()
+			fmt.Printf("failed to connect to all bootstrappers\n")
```
MSevey, line 156: If you aren't formatting anything, just use `Println` as that has the `\n` included
```suggestion
			fmt.Println("failed to connect to all bootstrappers")
```

MSevey: if the desired out come was this
```

failed...

```
then you need an extra `\n` with the `Printf`

Wondertan:
> Do you see a way to write tests for these?

We don't write tests for cel-shed usually. In some way it's bad, but on the bright side it saves a lot of time, like in this case writing a proper test is not straightforward. Usually, manual testing is enough here and that's why I am checking.

Wondertan (dismissed):

walldiss (approved):

smuu:
> When running this in a container I get `bash: /bin/cel-shed: cannot execute binary file: Exec format error`. But I can run the `celestia` command in the container. `cel-key` fails for the same reason.
> 
> Any ideas what is going on here?
> 
> ```
> bootstrapper-probe-celestia-node-0:~$ celestia
> 
>             ____      __          __  _
>           / ____/__  / /__  _____/ /_(_)___ _
>          / /   / _ \/ / _ \/ ___/ __/ / __  /
>         / /___/  __/ /  __(__  ) /_/ / /_/ /
>         \____/\___/_/\___/____/\__/_/\__,_/
> 
> Usage:
>   celestia [command]
> 
> Available Commands:
>   blob        Allows to interact with the Blob Service via JSON-RPC
>   bridge      Manage your Bridge node
>   completion  Generate the autocompletion script for the specified shell
>   das         Allows to interact with the Daser via JSON-RPC
>   docgen      docgen generates the openrpc documentation for all Celestia Node packages
>   full        Manage your Full node
>   header      Allows interaction with the Header Module via JSON-RPC
>   light       Manage your Light node
>   node        Allows administrating running node.
>   p2p         Allows interaction with the P2P Module via JSON-RPC
>   share       Allows interaction with the Share Module via JSON-RPC
>   state       Allows interaction with the State Module via JSON-RPC
>   version     Show information about the current binary build
> 
> Flags:
>   -h, --help   help for celestia
> 
> Additional help topics:
>   celestia            
> 
> Use "celestia [command] --help" for more information about a command.
> bootstrapper-probe-celestia-node-0:~$ cel-shed
> bash: /bin/cel-shed: cannot execute binary file: Exec format error
> bootstrapper-probe-celestia-node-0:~$ cel-key
> bash: /bin/cel-key: cannot execute binary file: Exec format error
> ```

Fixed due to @tty47

Wondertan (approved):

MSevey (approved):

cmd/cel-shed/p2p.go
```diff
@@ -75,3 +85,124 @@ var p2pPeerIDCmd = &cobra.Command{
...
+		}
+
+		fmt.Printf("PeerID: %s\n", p2pInfo.ID)
+		for _, addr := range p2pInfo.Addrs {
+			fmt.Printf("Listening on: %s\n", addr.String())
+		}
+		fmt.Println()
+
+		successfulConnections := 0
+		failedConnections := 0
```
cristaloleg, line 153: This one should be `var failedConnections atomic.Int32` 'cause it's accessed in different goroutines. Same for `successfulConnections`.

cristaloleg: Ah...you're using a mutex for this, my bad 🙈 

Anyway! you can replace type to `atomic.Int32` and drop the mutex.

cristaloleg (approved):

Wondertan (dismissed):

Dockerfile
```diff
@@ -54,6 +56,9 @@ RUN uname -a &&\
 # Copy in the binary
 COPY --from=builder /src/build/celestia /bin/celestia
 COPY --from=builder /src/./cel-key /bin/cel-key
+COPY --from=builder /src/./cel-shed /bin/cel-shed
+
+RUN cel-shed --help
```
renaynay, line 61: why is this necessary to have in the dockerfile?

renaynay (dismissed):
utack

Dockerfile
smuu: `RUN cel-shed --help` was only for debugging, removed it

We use `cel-shed` in our alerting system to get notified if all hardcoded bootstrappers are down simultaneously.
We would like to use the official container to automatically have the hardcoded list.

renaynay (approved):
utack

Wondertan (approved):

Issue #3777, opened 2024-09-26
share/shwap/discovery: Reformat discovery tag


renaynay:
Currently it's `/v0.1.0/full` and other protocols we have the version at the end. 

Let's make it as such in discovery too please :pray:

PR #3781, opened 2024-09-27
fix(modshare): discovery leftover from #3777 

Wondertan:
I realized by seeing logs that the archival tag and version weren't rearranged in #3777


cristaloleg (approved):

walldiss (approved):

odeke-em:
Would be useful to add a test to protect against a regression here.

Wondertan:
Don't think its worth in this case

PR #3637, opened 2024-08-08
feat(shwap): shwap nodebuilder and cleanup

walldiss:
This PR:
- add integration of all shwap modules into nodebuilder 
- Full Pruner integration (2 lines)
- removes unused code
  - old storage related code and tests
  - old getter related code
- fixes linter errors
- fixes multiple tests
- fixes edsser util

There are still remaining unused code here and there that will be clean-up later. The goal of this PR is to make sure that
- build
- test 
- lint
- run 
Are running without errors

PR #3789, opened 2024-10-01
fix(modshare): ensure Bitswap Getter is started

Wondertan:
What title says.  Likely forgotten from #3637


walldiss (approved):

cristaloleg (approved):
SGTM

Wondertan:
Let's not merge this; I suspect this is actually detrimental, and yet it can work without Getter being started. Do you think we should disable that?

Wondertan:
Robusta test proved this change useful when used without #3813. #3813 on itself makes timeout issues more prevailing.

PR #3796, opened 2024-10-02
feat!: return proofs up to the data root

rach-id:
This PR is created for now just to agree on the existing changes and whether we need to make more changes.


vgonkivs:
closed in favor of https://github.com/celestiaorg/celestia-node/pull/3985

PR #3808, opened 2024-10-03
fix(docker): set addr flags to 0.0.0.0 in entrypoint file

MSevey:
When using port forwarding in docker, you need to use the `any` address interface `0.0.0.0` otherwise you would be able to communicate to the server inside the container. 

This PR updates the `entrypoint.sh` file to set the RPC and Gateway addresses to `0.0.0.0` so that developers can properly communicate to the node when exposing ports, for example `-p 26658:26658`.

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -4,7 +4,7 @@ set -e
 
 if [ "$1" = 'celestia' ]; then
     echo "Initializing Celestia Node with command:"
-    COMMAND="celestia "${NODE_TYPE}" init --p2p.network "${P2P_NETWORK}""
+    COMMAND="celestia "${NODE_TYPE}" init --p2p.network "${P2P_NETWORK}" --rpc.addr="0.0.0.0" --gateway.addr="0.0.0.0""
     if [[ -n "$NODE_STORE" ]]; then
         COMMAND=${COMMAND}" --node.store "${NODE_STORE}""
         echo $COMMAND
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -4,13 +4,14 @@ set -e
 
 if [ "$1" = 'celestia' ]; then
     echo "Initializing Celestia Node with command:"
-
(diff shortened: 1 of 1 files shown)


MSevey:
> apart from length, is there any other reason to wrap it in `COMMAND` ?

nope, just DRY'ing up the code a little.

MSevey:
is this a known NDF in the test suite? 
https://github.com/celestiaorg/celestia-node/actions/runs/11165146996/job/31036158882?pr=3808#step:4:402
```
--- FAIL: TestEDSStore (0.02s)
    --- FAIL: TestEDSStore/Second_Put_should_be_noop (0.00s)
        testing.go:1232: TempDir RemoveAll cleanup: unlinkat /tmp/TestEDSStoreSecond_Put_should_be_noop3464455727/001/blocks: directory not empty
FAIL
```

renaynay (approved):
utack

docker/entrypoint.sh
```diff
@@ -4,13 +4,14 @@ set -e
...
 if [ "$1" = 'celestia' ]; then
     echo "Initializing Celestia Node with command:"
-
+    COMMAND="celestia "${NODE_TYPE}" init --p2p.network "${P2P_NETWORK}" --rpc.addr="0.0.0.0" --gateway.addr="0.0.0.0""
     if [[ -n "$NODE_STORE" ]]; then
-        echo "celestia "${NODE_TYPE}" init --p2p.network "${P2P_NETWORK}" --node.store "${NODE_STORE}""
-        celestia "${NODE_TYPE}" init --p2p.network "${P2P_NETWORK}" --node.store "${NODE_STORE}"
+        COMMAND=${COMMAND}" --node.store "${NODE_STORE}""
+        echo $COMMAND
+        $COMMAND
```
cristaloleg, line 11: Nit: looks like these 2 lines can be after `fi` and `else` branch can be removed.

MSevey: true, is it ok to just merge though since it is approved and handle this nit as a good beginner issue? 
This is currently blocking some work on rollkit.

PR #3810, opened 2024-10-04
chore: remove some TODOs

cristaloleg:
Some of the TODOs are outdated or not that important (imo).


nodebuilder/header/service.go
```diff
@@ -81,8 +81,6 @@ func (s *Service) GetByHeight(ctx context.Context, height uint64) (*header.Exten
 			"networkHeight: %d, requestedHeight: %d", head.Height(), height)
 	}
 
-	// TODO(vgonkivs): remove after https://github.com/celestiaorg/go-header/issues/32 is
```
cristaloleg, line 84: This issue is fixed, save to remove?

nodebuilder/share/config.go
```diff
@@ -16,7 +16,6 @@ const (
 	defaultBlockstoreCacheSize = 128
 )
 
-// TODO: some params are pointers and other are not, Let's fix this.
```
cristaloleg, line 19: This change triggers api break ci. However, change is simple and safe.

walldiss (approved):

Wondertan (approved):

PR #3813, opened 2024-10-04
feat(shwap/bitswap): use new option for optimized Has check

Wondertan:
This option instructs Bitswap to use `Has` call instead of `GetSize` on `Blockstore` for `WANT_HAVE` message, which prevents `Store` from loading and caching eds files in memory. Instead, the `Bitswap` now simply check eds file presence on the disk.


cristaloleg (approved):
SGTM

Wondertan:
Keeping this draft until further testing. This change seems to make request timeout issues worth

Issue #3516, opened 2024-06-20
blob: Unify `GetProof` and `GetCommitmentProof`


rach-id:
One potential idea is to fix GetProof instead of introducing a new endpoint with equivalent purpose. We could copy @rach-id's implementation of proof compute and solve the EDS fetching inefficiency later on with Shwap.

_Originally posted by @Wondertan in https://github.com/celestiaorg/celestia-node/pull/3470#discussion_r1645211905_

renaynay:
Closing this issue as Proof now proves the blob up to the data root via #3470

Wondertan:
@renaynay. this wasn't fixed and this is a follow up issue from #3470. We still have two methods `GetProof` and `GetCommitmentProof` that are meant to do the same thing

renaynay:
Please rename the issue then so it's more clear @Wondertan

PR #3821, opened 2024-10-07
feat!: unify the blob.Proof and the CommitmentProof

rach-id:
Closes https://github.com/celestiaorg/celestia-node/issues/3516 and https://github.com/celestiaorg/celestia-node/issues/3582

Note: added couple notes for reviewers concerning the error handling


go.mod
```diff
@@ -11,7 +11,7 @@ require (
 	github.com/celestiaorg/go-fraud v0.2.1
 	github.com/celestiaorg/go-header v0.6.2
 	github.com/celestiaorg/go-libp2p-messenger v0.2.0
-	github.com/celestiaorg/go-square v1.1.0
+	github.com/celestiaorg/go-square v1.1.1-0.20241007084659-3f4910615eeb
```
rach-id, line 14: will be changed once we merge https://github.com/celestiaorg/go-square/pull/108 and make a new release

blob/service_test.go
```diff
@@ -298,20 +299,40 @@ func TestBlobService_Get(t *testing.T) {
...
+					ctx,
+					12345,
+					blobsWithDiffNamespaces[1].Namespace(),
+					proof,
+					blobsWithDiffNamespaces[1].Commitment,
+				)
 			},
 			expectedResult: func(res interface{}, err error) {
-				require.NoError(t, err)
+				// Question for reviwers: for the case of included, it's here made not to return an error
```
rach-id, line 329: cc @vgonkivs

blob/service_test.go
```diff
@@ -289,7 +288,9 @@ func TestBlobService_Get(t *testing.T) {
 			},
 			expectedResult: func(res interface{}, err error) {
 				require.Error(t, err)
-				require.ErrorIs(t, err, ErrInvalidProof)
+				// Question for reviewers: same for here, we're returning exactly, ErrInvalidProof.
```
rach-id, line 291: cc @vgonkivs

go.mod
rach-id: e9f6cf7

blob/service.go
```diff
@@ -300,6 +296,8 @@ func (s *Service) getAll(
 // To ensure that blob was included in a specific height, we need:
 // 1. verify the provided commitment by recomputing it;
 // 2. verify the provided Proof against subtree roots that were used in 1.;
+// Note: this method can be deprecated because it's doing processing that can
+// be done locally.
```
vgonkivs, line 300: Since our API operates with header's height, deprecating this method requires users to download the header to get the hash.
So, I agree that we can deprecate this method in favour of proof public `validate` methods but only after migration from heights to hashes.

blob/service_test.go
rach-id: so we can return the error now?

blob/service.go
rach-id: cool, we can create an issue for it

rach-id: https://github.com/celestiaorg/celestia-node/issues/3919

vgonkivs: I think we had somewhere such an issue or task inside the EPIC. I will double-check and close the issue in this case

blob/service_test.go
vgonkivs: Yes. This case is impossible, since verification does not require any network requests now.

rach-id: [fed1637](https://github.com/celestiaorg/celestia-node/pull/3821/commits/fed16378e5b055bdbc3637bd4125d943f142a0a1)

blob/service_test.go
rach-id: [6d2dc80](https://github.com/celestiaorg/celestia-node/pull/3821/commits/6d2dc800af26af882e494e83104bebcda50edc8c)

blob/blob.go
```diff
@@ -28,78 +26,150 @@ var errEmptyShares = errors.New("empty shares")
...
-		cursor += sharesUsed
+		subtreeRootsCursor += len(ranges)
 	}
+
 	return true, nil
 }
 
+// GenerateCommitment generates the share commitment corresponding
+// to the proof's subtree roots
+func (p *Proof) GenerateCommitment() []byte {
```
Wondertan, line 160: nit: if that's commitment, it should return the respective Commitment type we defined.

blob/blob.go
```diff
@@ -28,78 +26,150 @@ var errEmptyShares = errors.New("empty shares")
...
-	if err != nil {
-		return false, err
+// Commitment is a Merkle Root of the subtree built from shares of the Blob.
+// It is computed by splitting the blob into shares and building the Merkle subtree to be included
+// after Submit.
+type Commitment []byte
+
+// Verify takes a data root and verifies if the
+// provided proof's subtree roots were committed to the given data root.
+func (p *Proof) Verify(dataRoot []byte, subtreeRootThreshold int) (bool, error) {
```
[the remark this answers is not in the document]
Wondertan: For now we could something simple as `appconsts.SubtreeRootThreshold(appVersion)`

nodebuilder/share/share.go
```diff
@@ -17,11 +20,30 @@ import (
...
 
-// GetRangeResult wraps the return value of the GetRange endpoint
-// because Json-RPC doesn't support more than two return values.
-type GetRangeResult struct {
-	Shares []share.Share
-	Proof  *types.ShareProof
+// RangeResult wraps the return value of the GetRange endpoint.
+// It contains a set of shares along with their proof to
+// the data root.
+type RangeResult struct {
```
Wondertan, line 26: We should call it `Range`. The `Result` implies it's an API-specific struct, while it's not. Range is gonna become a shared type

blob/blob.go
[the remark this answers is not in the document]
rach-id: The reason this was left to be provided is to have its logic in a separate place.
Once we start having many values for the subtree root threshold, we will need the verify method to know which threshold it is depending on the block containing the data.

So, leaving it as an argument is clearer IMO

blob/blob.go
rach-id: c89bea467a5b00c01ce6e2c5977a3891ce352858

nodebuilder/share/share.go
rach-id: but it's currently the return value of the GetRange endpoint. You mean there is separate work that will make RangeResult used elsewher?

blob/blob.go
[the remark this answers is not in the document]
rach-id: a508263b83cd1517cf6a0b6475259743ea87afe2

nodebuilder/share/share.go
Wondertan: Let's keep it as it is.

Wondertan (approved):
Thanks for the work here and apologies for taking it that long with reviews!

rach-id:
Thanks for the reviews :pray:
 
Can you merge please? And after ifs merged, you will take care of opening a PR from the feature branch to main?

vgonkivs (approved):

PR #3828, opened 2024-10-08
fix(pruner): fix find algo so that it does not pass a malformed range into `GetRangeByHeight` call

### LINKED ISSUE #3823: ERROR  pruner/service
### Celestia Node version

0.18.2

### OS

ubuntu 22.04

### Install tools

_No response_

### Others

_No response_

### Steps to reproduce it

run with --experimental-pruning flag

### Expected result

n/a

### Actual result

getting these from time to time:

Oct 07 15:19:53 brd02-tia-t celestia[730210]: 2024-10-07T15:19:53.275Z        ERROR        pruner/service        pruner/find.go:44        failed to get range from header store        {"from": 1325056, "to": 1325057, "error": "header/store: invalid range(1325057,1325056)"}

### Relevant log output

_No response_

_No response_

### Notes

_No response_

renaynay:
Fixes #3823


vgonkivs (approved):

cristaloleg (approved):

pruner/find.go
```diff
@@ -31,7 +31,7 @@ func (s *Service) findPruneableHeaders(
 		return nil, err
 	}
 
-	if lastPruned.Height() == estimatedCutoffHeight {
```
walldiss, line 34: Could you please add explanation on why estimated height needs to be higher than lastPruned by at least 2? It seems odd, so comment will help

walldiss (commented):
It would benefit from unit test

pruner/find.go
renaynay: Ah yes sure.

renaynay: done https://github.com/celestiaorg/celestia-node/pull/3828/commits/150e9c900b3e6b5d79e5fb5d242207a8df70fc6e

vgonkivs (approved):

pruner/find.go
```diff
@@ -31,7 +31,8 @@ func (s *Service) findPruneableHeaders(
...
+	// note that GetRangeByHeight will fetch the range (lastPruned.Height(): estimatedCutoffHeight)
```
walldiss, line 34: Still had to look into go-header to figure out why, so perhaps this may clarify:
```suggestion
	// note that GetRangeByHeight will fetch the range, that does not include lastPruned height: [lastPruned.Height()+1:to).
```

pruner/find.go
```diff
@@ -31,7 +31,8 @@ func (s *Service) findPruneableHeaders(
...
+	if lastPruned.Height() >= estimatedCutoffHeight-1 {
```
walldiss, line 35: Simplification nit:
```suggestion
	if lastPruned.Height() > estimatedCutoffHeight {
```

walldiss: wrong

walldiss, line 35: The condition was making me feel uneasy, and I kept trying to figure out what was off until I had an "aha" moment. According to the `GetRangeByHeight` doc: [lastPruned.Height()+1:to), `to` is not included in the returned range. To prune it, we should request the range [lastPruned.Height():estimatedCutoffHeight+1) from [GetRangeByHeight here](https://github.com/celestiaorg/celestia-node/blob/0da16ddfcf25ea98d41602b59850f462f4df83c3/pruner/find.go#L42). So, to prune `estimatedCutoffHeight`, the condition should allow `lastPruned.Height() < estimatedCutoffHeight`. We can achieve this by removing the `-1` and modifying linked code.
```suggestion
	if lastPruned.Height() >= estimatedCutoffHeight {
```

pruner/find.go
renaynay: That's the same in notation --  `(` is non-inclusive

pruner/find.go
renaynay: We use the store implementation of `Getter` iface here which includes an additional check 

```
	// as the requested range is non-inclusive in the end[from;to), we need to compare
	// `from` with `to-1`
	if from > to-1 {
		return nil, fmt.Errorf("header/store: invalid range(%d,%d)", from, to-1)
	}

```

where `from` would be lastPruned.Height()+1 and `to` would be `estimatedCutoffHeight`

So if we have a situation where `lastPruned.Height()+1` = `estimatedCutoffHeight`, it would hit this condition and error out as `lastPruned.Height()+1 > estimatedCutoffHeight-1`.

I agree this is very ugly bc it forces the caller to be aware of the detail that `GetRangeByHeight` will error out if a range of 1 header is attempted. The fix should be in header `store` itself though where it should allow a range request of amount `1` to happen as that's still a valid range.

Until that fix is in go-header, however, we should fix this so pruning logs don't keep erroring. I can add a TODO here + make an issue.

renaynay: https://github.com/celestiaorg/go-header/pull/227

renaynay: https://github.com/celestiaorg/celestia-node/pull/3828/commits/e7fc9c5af0fb4a43392028373304099ce9250629

this is the better fix

cristaloleg: Looks like mentioned commit isn't a part of this PR, am I wrong?

walldiss (approved):

cristaloleg (approved):
SGTM

PR #3834, opened 2024-10-09
fix(.github/workflows): attempt to fix releaser workflow so it runs on tags/branches not just main

renaynay:
Attempt to fix CI release workflow 😬

### DIFF
--- a/.github/workflows/ci_release.yml
+++ b/.github/workflows/ci_release.yml
@@ -109,7 +109,7 @@ jobs:
 
   # Generate the release with goreleaser to include pre-built binaries
   goreleaser:
-    needs: [version_bump, setup]
+    needs: [setup]
     runs-on: ubuntu-latest
     if: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_release.yml
```diff
@@ -4,6 +4,9 @@ on:
   push:
     branches:
       - main
+      - '**' # This allows the workflow to trigger on any branch
+    tags:
+      - 'v*.*.*'
```
cristaloleg, line 9: I think you were looking for something else https://regex101.com/r/GO0uEm/1

Maybe just `v.*` ? as in https://github.com/celestiaorg/celestia-node/pull/3787 (shiiit, should be `v.*` in the mentioned PR 😬 )

cristaloleg (dismissed):
SGTM

cristaloleg (approved):

renaynay:
Closing in favuor of #3855

PR #3835, opened 2024-10-10
test(nodebuilder): fuzzer for fullClient.DataAvailability submission

odeke-em:
This change presents a fuzzer for testing out fullClient data availability submission, in which the inputs to .Submit are mutated then the responses tested out to check for logical conformation. I ran this fuzzer for about 8 hours on my machine, but ideally we should get this repository up on oss-fuzz for continuous fuzzing.


odeke-em:
Kindly cc-ing @Wondertan @liamsi

nodebuilder/tests/fuzz_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+var namespace, _ = share.NewBlobNamespaceV0([]byte("namespace"))
+
+type daModuleCorpus struct {
+	Blobs [][]byte `json:"blobs"`
+}
+
+func FuzzDaModule(f *testing.F) {
+	// 1. Retrieve the fuzz corpra.
+	corpraPath := filepath.Join("testdata", "bytes-corpra", "*.bytes")
```
cristaloleg, line 32: `bytes-submit` ?

odeke-em: Sorry I changed the path name while the fuzzer was running during its 8 hours. I'll update that, thanks!

PR #3838, opened 2024-10-11
feat(modp2p): close peer connections after successful blocking

### LINKED ISSUE #3791: modp2p: `BlockPeer` doesn't disconnect the blocked peer
Currently, we simply [tell ConnGater to block a peer](https://github.com/celestiaorg/celestia-node/blob/fc1dca3354d152151ef1bd87657f91ecd884422c/nodebuilder/p2p/p2p.go#L148). However, in case the peer is connected, that does not cause a disconnect from it.

Using `ClosePeer` method on the API, we can simply close right after blocking succeeds

notlelouch:
closes #3791


renaynay (commented):
Thank you!

cristaloleg (approved):

renaynay (approved):

renaynay (changes requested):
There is a broken unit test `TestP2PModule_ConnGater` that's related to this change, do you mind looking into it 🙏🏻

cristaloleg (approved):

nodebuilder/p2p/p2p.go
```diff
@@ -145,7 +146,15 @@ func (m *module) NATStatus(context.Context) (network.Reachability, error) {
...
 func (m *module) BlockPeer(_ context.Context, p peer.ID) error {
-	return m.connGater.BlockPeer(p)
+	if err := m.connGater.BlockPeer(p); err != nil {
+		return err
+	}
+	if m.host != nil {
+		if err := m.host.Network().ClosePeer(p); err != nil {
+			log.Warnf("failed to close connection to blocked peer %s: %v", p, err)
+		}
+	}
```
walldiss, line 156: The Host field should always be initialized, so there shouldn’t be a nil check for it. Could you please remove the if m.host != nil condition to reflect this?

Additionally, could you modify TestP2PModule_ConnGater to also test the newly introduced Disconnect call? This will help ensure the new logic is fully tested.

cristaloleg (approved):

cristaloleg:
Thanks, LGTM!

Wondertan (approved):

walldiss (approved):
Looks good! Thank you for contribution!

PR #3844, opened 2024-10-14
fix(store): check if link exists on write

walldiss:
In case of unexpected shutdown file might be partially written, while link is not. This PR will ensure we that put will result in both link and file on success.

File validation will be tackled in another PR

### DIFF
--- a/store/store.go
+++ b/store/store.go
@@ -202,7 +202,7 @@ func (s *Store) createODSFile(
 ) (bool, error) {
 	pathODS := s.hashToPath(roots.Hash(), odsFileExt)
 	err := file.CreateODS(pathODS, roots, square)
-	if err != nil {
+	if err != nil && !errors.Is(err, os.ErrExist) {
 		// ensure we don't have partial writes if any operation fails
 		removeErr := s.removeODS(height, roots.Hash())
 		return false, errors.Join(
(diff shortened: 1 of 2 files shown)


store/store.go
```diff
@@ -202,10 +202,6 @@ func (s *Store) createODSFile(
 ) (bool, error) {
 	pathODS := s.hashToPath(roots.Hash(), odsFileExt)
 	err := file.CreateODS(pathODS, roots, square)
-	if errors.Is(err, os.ErrExist) {
-		// TODO(@Wondertan): Should we verify that the exist file is correct?
-		return true, nil
-	}
 	if err != nil {
```
Wondertan, line 205:
```suggestion
	if err != nil && !errors.Is(err, os.ErrExist) {
```

Wondertan (commented):
It asks for a test

store/store_test.go
```diff
@@ -90,6 +90,30 @@ func TestEDSStore(t *testing.T) {
 		ensureAmountFileAndLinks(t, dir, 2, 1)
 	})
 
+	t.Run("Second Put after partial write", func(t *testing.T) {
+		dir := t.TempDir()
```
Wondertan, line 94: In case tests are run in parallel, do they have their own temp dir?

walldiss: yes

walldiss: Also test coverage is going to be greatly improved in following PR

Wondertan (approved):

Wondertan:
For context: shwap-robus network has a few blocks missed in the storage, and the hypothesis is that they were missed during OOM leaving the Store in an inconsistent state(file exists, but link for it does not). This covers this possibility and we are yet to prove this is an issue

cristaloleg (approved):
SGTM

PR #3845, opened 2024-10-14
chore(Dockerfile): bump go version

Wondertan:
We removed `godebug asynctimerchan=1` in https://github.com/celestiaorg/celestia-node/pull/3790 and forgot to update the Dockerfile to the latest version of Go, resulting in images being built with go1.23.0 reproducing the original bug.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM --platform=$BUILDPLATFORM docker.io/golang:1.23-alpine3.20 AS builder
+FROM --platform=$BUILDPLATFORM docker.io/golang:1.23.2-alpine3.20 AS builder
 
 ARG TARGETPLATFORM
 ARG BUILDPLATFORM


walldiss:
https://github.com/docker-library/golang/blob/89de06f6dd2c4edb29f33bb7270bdcc8000cf58a/1.23/alpine3.20/Dockerfile

Wondertan:
Hmm, looks right, then why was the go version was 1.23.0 🤔

walldiss:
Perhaps your image was build at time, when 1.23 was pointing to 1.23.0. If you rebuild with new one it should use 1.23.2

Wondertan:
https://github.com/docker-library/golang/commit/89de06f6dd2c4edb29f33bb7270bdcc8000cf58a

2 weeks ago, while image was build over the weekend

Issue #3840, opened 2024-10-11
docs: response examples should match API


jcstein:
# Problem

[From Slack thread](https://celestia-team.slack.com/archives/C0639CBCAF7/p1728632038656139?thread_ts=1728548434.440719&cid=C0639CBCAF7)

I generated specs using `celestia docgen` on v0.16.0 and v0.17.1 and added them to docs, but it is reported that the response examples are incorrect on docs still, although the API changed. API spec updates from v0.16.0 and v0.17.1 do not contain correct responses, for at least tx responses:

Source: https://github.com/celestiaorg/celestia-node/releases/tag/v0.16.0

> API Response Changes: The return type for SubmitPayForBlob remains unchanged, but most fields in the response will now be omitted. Only the Code, TxHash, and Height fields will be populated, with all other fields set to empty values.

# Solution

Modify the docs within celestia-node so that when docs are generated in the CI (building the openrpc.json specs) or manually in CLI, they contain the correct responses

jcstein:
Note that these were manually generated because CI did not build the specs for either of the mentioned releases.

zvolin:
I think the reason for that could be because the type hasn't changed on celestia-node side, just now celestia-app responds with the other type that only partially maps to the TxResponse. Docs probably just fill dummy values for the type they see

PR #3848, opened 2024-10-14
fix(docs) update txResponse in docs

walldiss:
Fixes https://github.com/celestiaorg/celestia-node/issues/3840


api/docgen/exampledata/txResponse.json
```diff
@@ -1,187 +1,5 @@
...
-      "attributes": [
-        {
-          "key": "bW9kdWxl",
-          "value": "YmFuaw==",
-          "index": true
-        }
-      ]
-    }
-  ]
+  "code": 0
```
walldiss, line 4: This field is actually optional and only set in case of errors.

vgonkivs (approved):

cristaloleg (approved):

renaynay (approved):

zvolin:
as per our async conversation, `"code": 0` will likely not be present as it's not serialized if default. Did that change?

zvolin:
ok I found a comment saying it's optional. I think putting that in docs is confusing and will likely make ppl try to get it anyway, it's not documented missing means 0. I'm a bit biased as I don't like such `default == missing` behaviors at all so take it with a grain of salt :p

PR #3859, opened 2024-10-17
feat(store): correct corrupted files on write

walldiss:
This PRs adds corruption detection to store. Corrupted files are detected by checking if their length is corresponding to expected file size. There is no content inspection for files verification yet, but it can be added via datahash compute. Added support for:
- ODSQ4 file
- ODS file

### DIFF
--- a/store/file/ods_q4.go
+++ b/store/file/ods_q4.go
@@ -43,23 +43,21 @@ func CreateODSQ4(
 	go func() {
 		// doing this async shaves off ~27% of time for 128 ODS
 		// for bigger ODSes the discrepancy is even bigger
-		err := createQ4(pathQ4, eds)
-		if err != nil {
-			err = fmt.Errorf("сreating Q4 file: %w", err)
-		}
-
-		errCh <- err
+		errCh <- createQ4(pathQ4, eds)
 	}()
 
-	if err := CreateODS(pathODS, roots, eds); err != nil {
+	err := CreateODS(pathODS, roots, eds)
+	q4Err := <-errCh
+
+	if err != nil && q4Err != nil {
+		return fmt.Errorf("creating ODS and Q4 files: %w", errors.Join(err, q4Err))
+	}
+	if err != nil {
 		return fmt.Errorf("creating ODS file: %w", err)
 	}
-
-	err := <-errCh
-	if err != nil {
-		return err
+	if q4Err != nil {
+		return fmt.Errorf("creating Q4 file: %w", q4Err)
 	}
-
 	return nil
 }
 


store/store.go
```diff
@@ -181,6 +181,22 @@ func (s *Store) createODSQ4File(
...
+		err = file.CheckODSQ4Size(pathODS, pathQ4, square)
+		if err != nil {
+			err = s.removeODSQ4(height, roots.Hash())
+			if err != nil {
+				return false, fmt.Errorf("removing corrupted ODSQ4 file: %w", err)
+			}
+			err = file.CreateODSQ4(pathODS, pathQ4, roots, square)
+			if err != nil {
+				return false, fmt.Errorf("recreating ODSQ4 file: %w", err)
+			}
```
Wondertan, line 196: Readbility nit: extract into a `validateAnRecoverX` method

store/file/ods.go
```diff
@@ -130,6 +130,31 @@ func writeAxisRoots(w io.Writer, roots *share.AxisRoots) error {
 	return nil
 }
 
+// CheckODSSize checks if the file under given FS path has the expected size.
+func CheckODSSize(path string, eds *rsmt2d.ExtendedDataSquare) error {
```
Wondertan, line 134: More like Validate rather than check

store/file/ods.go
```diff
@@ -130,6 +130,31 @@ func writeAxisRoots(w io.Writer, roots *share.AxisRoots) error {
...
+}
+
+func checkODSSize(path string, eds *rsmt2d.ExtendedDataSquare) (*headerV0, error) {
+	ods, err := OpenODS(path)
+	if err != nil {
+		return nil, fmt.Errorf("opening ODS file: %w", err)
+	}
+
+	shares := filledSharesAmount(eds)
+	expectedSize := ods.hdr.OffsetWithRoots() + shares*ods.hdr.ShareSize()
```
Wondertan, line 146: Shouldnt we verify header as well instead of trusting it? Whats if header also corrupted?

walldiss: It will fail to read the header in check and return the error (same results as corrupted)

walldiss: If it has corrupted values it is very unlikely it will match exactly expected file size, and will fail the check. Keep in mind this is only Size check, not a full content inspection. In case of full content integrity check it would be necessary to verify headers values as well as content of roots and data.

store/store.go
walldiss: I think it is much easier to understand what is going on here by not hiding details into smaller func. Function calls are self-explanatory here and `validateAnRecoverX` is not going to be reused anywhere

store/file/ods.go
walldiss: https://github.com/celestiaorg/celestia-node/pull/3859/commits/49adeb65eac2b9f6842329b772125c209f399bce

store/store.go
Wondertan: The recovery details have layers of nesting involved  with edcase logic unrelated to create func. It would be more readable for a novice to read what simple create does and jump on recovery logic only and only if its needed

Wondertan: Essentially recovery logic is a separate operation thus needs a separate method

store/file/ods.go
Wondertan: > If the header has corrupted values it is very unlikely it will match exactly expected file size

This is what concerns me, we don't know if it will match or not. I don't think trusting header is valid when we dont trust the file. We could simply get the file size from the eds, arent we?

walldiss: Share size is tricker than length, but I think I have a workaround. We can use the size of first share. So we don't have to trust header on disk at all. Will update the code

Wondertan (approved):
Thanks for addressing readability nits

vgonkivs (dismissed):

Wondertan (approved):

cristaloleg (approved):
SGTM

vgonkivs (approved):
LGTM

Issue #3858, opened 2024-10-17
CPU usage spikes reports


walldiss:
### Implementation ideas

Multiple users have reported spikes in CPU usage on v0.18.2 (shwap) Bridge node. Need to investigate root cause and potential fixes. Examples based on metrics:

![Screenshot 2024-10-17 at 11 37 21](https://github.com/user-attachments/assets/d273b52e-3db4-427a-8a3a-172870e51d06)

![Screenshot 2024-10-17 at 11 36 39](https://github.com/user-attachments/assets/b15595ad-b202-4343-ac2b-df138ec4ba56)

PR #3860, opened 2024-10-17
fix(eds/cache): Fix prealloc for proof nodes

walldiss:
Previous prealloc calc was incorrect resulting in higher allocation than needed. This is chart of amount of allocation we should have done and actually done:
```
eds size; needed alloc; real alloc
2 3 4
4 7 24
8 15 112
16 31 480
32 63 1984
64 127 8064
128 255 32512
256 511 130560
512 1023 523264
```

Fixes https://github.com/celestiaorg/celestia-node/issues/3858

### DIFF
--- a/share/ipld/nmt_adder.go
+++ b/share/ipld/nmt_adder.go
@@ -196,5 +196,5 @@ func (a *ProofsAdder) addProof(id cid.Cid, proof []byte) {
 
 // innerNodesAmount return amount of inner nodes in eds with given size
 func innerNodesAmount(squareSize int) int {
-	return 2 * (squareSize - 1) * squareSize
+	return 2*squareSize - 1
 }


Wondertan (approved):
Good self catch

renaynay (approved):
Oh wow

Wondertan:
Link https://github.com/celestiaorg/celestia-node/issues/3858

odeke-em:
Thank you for the fix @walldiss! I highly would encourage adding a test to lock this behavior as it is subtle and even if it changed from a revert (say if something unrelated broke) it would never be caught.

Wondertan:
Agree with @odeke-em here. If we had a test for the number of nodes, you wouldn't introduce this in the first place

Issue #3853, opened 2024-10-16
[Feature Request]: Improve [empty] block indexing by using an alternative to symlinks with absolute paths


easy2stake:
Hi,
```
Semantic version: v0.18.2-arabica
Commit: 4309c8349857638b033a2a278da0f8ab182fdb26
Build Date: Tue Oct  8 01:11:02 PM EEST 2024
System version: amd64/linux
Golang version: go1.23.2
```

### The "issue":
Right now, on celestia mocha there are `634247` links inside `blocks/heights` folder, all poiting to:
`/home/celestia-bridge-t-shwap/.celestia-bridge-mocha-4/blocks/3D96B7D238E7E0456F6AF8E7CDF0A67BD6CF9C2089ECB559C659DCAA1F880353.ods` 

The above is the "absolute" path in my filesystem. 
And this is the reason I'm asking for this "feature request", I didn't consider it a BUG, but it's problematic being "absolute".

**Why is this problematic:**
The links above are all using absolute paths towards the block hash `3D96B7D238E7E0456F6AF8E7CDF0A67BD6CF9C2089ECB559C659DCAA1F880353.ods` which is a "special empty block" accrodiing to @Wondertan 

**For node runners who run archival bridge nodes these links are:**
- limiting our ability to copy the database from one machine to another
- it is complicated to have backups / snapshots to restore from
- some cloning tools such as rclone, doesn't fully support symlinks when copying files
- it makes it harder to minimize downtime if, for some reason, you have to restore it
- you can't even move the datadir on the same machine  but another disk, such as a bigger partition when this will grow large enough

**Solution:**
The first thing that crossed my mind was, Why not use a **KVdb** for this?
And it seems it's overkill since this behavior only happens when empty blocks are created, and we don't expect to many empty blocks on mainnet.

**Then another idea:** 
Why not using the same symlinks but make them relative considering the directory structure. The current structure:
```
.celestia-node-datadir/blocks/3D96B7D238E7E0456F6AF8E7CDF0A67BD6CF9C2089ECB559C659DCAA1F880353.ods
.celestia-node-datadir/blocks/heights/here-are-the-symlinks
```

Therefore, each symlink could point to ```../3D96B7D238E7E0456F6AF8E7CDF0A67BD6CF9C2089ECB559C659DCAA1F880353.ods.```

Not sure if this could have other implications, but I hope we can elaborate on the subject.
[...]

PR #3861, opened 2024-10-17
feat(store): use use relative path for empty blocks symlinks

walldiss:
Resolves https://github.com/celestiaorg/celestia-node/issues/3853

### DIFF
--- a/store/store.go
+++ b/store/store.go
@@ -232,14 +232,15 @@ func (s *Store) createODSFile(
 }
 
 func (s *Store) linkHeight(datahash share.DataHash, height uint64) error {
-	// create hard link with height as name
-	pathOds := s.hashToPath(datahash, odsFileExt)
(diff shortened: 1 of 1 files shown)


store/store.go
```diff
@@ -444,6 +445,10 @@ func (s *Store) hashToPath(datahash share.DataHash, ext string) string {
 	return filepath.Join(s.basepath, blocksPath, datahash.String()) + ext
 }
 
+func (s *Store) hashToRelativePath(datahash share.DataHash, ext string) string {
```
Wondertan, line 448: Nit: doesnt need to be a method

walldiss: Made it as a method to be consistent with `hashToPath` and `heightToPath`

Wondertan (approved):

cristaloleg (approved):

PR #3865, opened 2024-10-21
fix(store/file): wait for q4 to be written before returning

walldiss:
TLDR: It was possible to return with error before reading from channel. Which could lead to leak of q4 writing go-routine.

Fixed a race condition when writing/removing the Q4 file. The main go-routine wasn’t waiting for the Q4 write to finish, causing the file to be removed before the write completed. This could leave unpredictable data on disk after the removal. These changes ensure the write operation finishes before any removal is attempted.

### DIFF
--- a/store/file/ods_q4.go
+++ b/store/file/ods_q4.go
@@ -43,23 +43,21 @@ func CreateODSQ4(
 	go func() {
 		// doing this async shaves off ~27% of time for 128 ODS
 		// for bigger ODSes the discrepancy is even bigger
-		err := createQ4(pathQ4, eds)
-		if err != nil {
-			err = fmt.Errorf("сreating Q4 file: %w", err)
-		}
-
-		errCh <- err
+		errCh <- createQ4(pathQ4, eds)
 	}()
 
-	if err := CreateODS(pathODS, roots, eds); err != nil {
+	err := CreateODS(pathODS, roots, eds)
+	q4Err := <-errCh
+
+	if err != nil && q4Err != nil {
+		return fmt.Errorf("creating ODS and Q4 files: %w", errors.Join(err, q4Err))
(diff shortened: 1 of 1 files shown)


store/file/ods_q4.go
```diff
@@ -43,23 +43,21 @@ func CreateODSQ4(
...
+		errCh <- createQ4(pathQ4, eds)
 	}()
 
-	if err := CreateODS(pathODS, roots, eds); err != nil {
-		return fmt.Errorf("creating ODS file: %w", err)
-	}
+	err := CreateODS(pathODS, roots, eds)
+	q4Err := <-errCh
 
-	err := <-errCh
```
Wondertan, line 58: Its waiting for the write in here. Where is the race exaclty?

walldiss: It was possible to return with error before reading from channel

Wondertan (approved):

store/file/ods_q4.go
Wondertan: Update the description to clarify that

cristaloleg: Is it possible to test that? I mean add a test.

cristaloleg (approved):
SGTM

vgonkivs (approved):

store/file/ods_q4.go
walldiss: Not sure how to reliably test it

walldiss: Unreliably the condition fails in tests added by https://github.com/celestiaorg/celestia-node/pull/3859

cristaloleg: Resolving 'cause not worth of additional testing. Thanks.

PR #3870, opened 2024-10-22
!refactor(api/share): use height in share module

walldiss:
Use height instead of header in share module API

### DIFF
--- a/nodebuilder/node_test.go
+++ b/nodebuilder/node_test.go
@@ -15,9 +15,7 @@ import (
 	collectormetricpb "go.opentelemetry.io/proto/otlp/collector/metrics/v1"
 	"google.golang.org/protobuf/proto"
 
-	"github.com/celestiaorg/celestia-node/header/headertest"
 	"github.com/celestiaorg/celestia-node/nodebuilder/node"
-	"github.com/celestiaorg/celestia-node/share"
 )
 
 func TestLifecycle(t *testing.T) {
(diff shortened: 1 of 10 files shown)


Wondertan (commented):
Swamp tests are broken as they still pass headers instead of height

nodebuilder/node_test.go
```diff
@@ -125,34 +123,3 @@ func StartMockOtelCollectorHTTPServer(t *testing.T) (string, func()) {
 	server.EnableHTTP2 = true
 	return server.URL, server.Close
 }
-
-func TestEmptyBlockExists(t *testing.T) {
```
Wondertan, line 129: That seems unrelated. Why? Even if it's not needed, let's keep PRs focused on their goal 🙏🏻

Wondertan (approved):

nodebuilder/node_test.go
walldiss: Unnecessary test, that is tricky to fix. Removal is the best solution here

Wondertan: Ok, agreed

cristaloleg (approved):

Wondertan:
conflicts (

renaynay (approved):
very quick review, thanks @walldiss

cristaloleg (approved):

vgonkivs:
FYI @jcstein

vgonkivs (approved):

renaynay (approved):

PR #3883, opened 2024-10-24
fix(availability): prevent parallel availability calls

walldiss:
Parallel availability calls could simultaneously select different set of uniq coordinates for block sampling. The selection would be overwritten by last call, which would result in more samples in store than needed. It would make light node to do more sampling work than needed and also would create problems for future pruning of additional samples. 

Provided unit explains the problem the best. With removed change, it would result in 176 samples in store instead of 16.


share/availability/light/availability.go
```diff
@@ -29,11 +29,9 @@ type ShareAvailability struct {
 	getter shwap.Getter
 	params Parameters
 
-	// TODO(@Wondertan): Once we come to parallelized DASer, this lock becomes a contention point
-	//  Related to #483
-	// TODO: Striped locks? :D
```
walldiss, line 34: TODO is no longer relevant

Wondertan (commented):
* When is this happening in practice? 
  * Besides someone else calling ShareAvailabe over RPC, shouldn't DASer ensure there is only one availability per height at a time?
* This seems like a broader problem for all availabilities and remind me of SyncGetter(:triggered:) we still missing. Shouldn't we do that for all availabilties instead of light?

share/availability/light/availability.go
```diff
@@ -66,6 +64,17 @@ func (la *ShareAvailability) SharesAvailable(ctx context.Context, header *header
...
+	lk, loaded := la.inProgress.LoadOrStore(header.Height(), &sync.Mutex{})
+	lk.(*sync.Mutex).Lock()
+	defer func() {
+		lk.(*sync.Mutex).Unlock()
+		// Delete the lock if it was created in this call
+		if !loaded {
+			la.inProgress.Delete(header.Height())
+		}
+	}()
+
```
Wondertan, line 77: If we ever make SharesAvailable blocking (we will), it would be great to have this respecting context. It shouldn't be hard to change Mutex to a channel. This will partially resemble `heightSub` logic and can be directly copied from there.

walldiss: Good idea. Reworked to use channels. After visiting `heightSub` it seems unnecessary complicated, so I've similar, but simpler solution. Perhaps it can be used in heightSub too?

renaynay (approved):
Am fine with this change - I do agree with hlib's comment re extending single flight protection to both calls (for full availability too) even though this circumstance only really happens from external RPC call.

One thing about it though is that we have an instance of this pattern already in go-header where we delegate the flight protection to the caller (syncer) rather than the component itself, but here it is delegated to the component. Just a thought, but we should probably try to choose one pattern.

walldiss:
>Am fine with this change - I do agree with hlib's comment re extending single flight protection to both calls (for full availability too) even though this circumstance only really happens from external RPC call.

Just to clarify, full availability doesn’t need this since the issue can’t happen there. In full availability, the state is kept in one place and updated all at once. But in light availability, the state is split between datastore and blockstore and updated separately.

>One thing about it though is that we have an instance of this pattern already in go-header where we delegate the flight protection to the caller (syncer) rather than the component itself, but here it is delegated to the component. Just a thought, but we should probably try to choose one pattern.

We can’t delegate to the caller here because there are two possible parallel callers from different components: DASer and RPC. Maybe for go-header, if there are potentially multiple callers, it should also handle state access internally instead of relying on the caller.

share/availability/light/availability.go
```diff
@@ -66,6 +64,22 @@ func (la *ShareAvailability) SharesAvailable(ctx context.Context, header *header
 		return nil
 	}
 
+	// Prevent multiple sampling sessions for the same header height
+	lockChan, loaded := la.activeHeights.LoadOrStore(header.Height(), make(chan struct{}))
+	if loaded {
+		select {
+		case <-lockChan.(chan struct{}):
```
Wondertan, line 71: So the waiter sits here until the first caller is done and then goes over the same flow again. As the state is stored, the waiter sees that the sampling operation finished earlier and short circuits. In case the original operation failed, the waiter retries. Do we want to retry for each waiter out there?

walldiss: Yes, next job will find result of previous job from store. It will retry if there is still time within provided ctx deadline. it is good to keep trying if caller is waiting.

renaynay (approved):

share/availability/light/availability.go
Wondertan: In case there 3 waiters, doesn't that mean all of them will do the same retrying?

Wondertan: @walldiss

walldiss: If all failed and all have time left to retry after previous one fail I see no issue. Availability supposed to succeed eventually.

walldiss: There was a bug causing second and further calls to start in parallel. Fixed in new commit

Wondertan (approved):

share/availability/light/availability.go
Wondertan: Yes, so this is a bug

renaynay (approved):
utack

PR #3887, opened 2024-10-28
feat(share/availability): Store success results for light availability

walldiss:
Store coordinates of successful sampling calls

### DIFF
--- a/share/availability/light/availability.go
+++ b/share/availability/light/availability.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"sync"
+	"time"
 
 	"github.com/ipfs/go-datastore"
 	"github.com/ipfs/go-datastore/autobatch"
@@ -63,6 +64,7 @@ func NewShareAvailability(
 // ExtendedHeader. This way SharesAvailable subjectively verifies that Shares are available.
 func (la *ShareAvailability) SharesAvailable(ctx context.Context, header *header.ExtendedHeader) error {
 	dah := header.DAH
+
 	// short-circuit if the given root is an empty data square
 	if share.DataHash(dah.Hash()).IsEmptyEDS() {
 		return nil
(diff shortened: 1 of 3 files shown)


Wondertan:
Why do we store successful sampling results? I know we wanted to do this, but I think we should wait to do it. My concern is mainly about not doing this together with the reconstruction and the new light node storage that would keep the bitmap-like state. If think you we need to do it earlier, then please explain.

walldiss:
>Why do we store successful sampling results? I know we wanted to do this, but I think we should wait to do it. My concern is mainly about not doing this together with the reconstruction and the new light node storage that would keep the bitmap-like state. If think you we need to do it earlier, then please explain.

We need to store list of sampled coordinates for pruning. We would need to know coordinates to delete samples from blockstore by cid

share/availability/light/availability.go
```diff
@@ -75,62 +77,81 @@ func (la *ShareAvailability) SharesAvailable(ctx context.Context, header *header
...
 		if err != nil {
 			return err
 		}
+		// Verify total samples count
+		totalSamples := len(samples.Remaining) + len(samples.Available)
+		if totalSamples != int(la.params.SampleAmount) {
+			return fmt.Errorf("invalid sampling result:"+
+				" expected %d samples, got %d", la.params.SampleAmount, totalSamples)
+		}
+	}
```
walldiss, line 105: This check failure not only verifies data stored, but also happen for users who didn't resync their node after upgrade. Which will help us to give them clear instructions for upgrade

Wondertan: I think they would get an error from Unmarshal if they don't resync

share/availability/light/availability.go
```diff
@@ -75,62 +77,80 @@ func (la *ShareAvailability) SharesAvailable(ctx context.Context, header *header
...
 	log.Debugw("starting sampling session", "height", header.Height())
-	var wg sync.WaitGroup
-	for _, s := range samples {
+
+	// remove one second from the deadline to ensure we have enough time to process the results
+	samplingCtx, cancel := context.WithCancel(ctx)
+	if deadline, ok := ctx.Deadline(); ok {
+		samplingCtx, cancel = context.WithDeadline(ctx, deadline.Add(-time.Second))
+	}
+	defer cancel()
```
Wondertan, line 124: ❤️

Wondertan: Alternatively, we could store results with an independent context to avoid making assumptions about the time it takes. While also running that in a separate routine to respect the real context and return as the user requested.

share/availability/light/availability.go
```diff
@@ -75,62 +77,80 @@ func (la *ShareAvailability) SharesAvailable(ctx context.Context, header *header
...
 				log.Debugw("error fetching share", "height", header.Height(), "row", s.Row, "col", s.Col)
-				failedSamplesLock.Lock()
 				failedSamples = append(failedSamples, s)
-				failedSamplesLock.Unlock()
+			} else {
+				samples.Available = append(samples.Available, s)
 			}
 		}(s)
 	}
 	wg.Wait()
```
Wondertan, line 142: @cristaloleg, this is the part we change in #3891 PR

share/availability/light/sample.go
```diff
@@ -7,12 +7,31 @@ import (
...
+}
+
 // Sample represents a coordinate in a 2D data square.
 type Sample struct {
 	Row int `json:"row"`
 	Col int `json:"col"`
 }
 
+// NewSamplingResult creates a new SamplingResult with randomly selected samples.
+func NewSamplingResult(squareSize, sampleCount int) *SamplingResult {
```
Wondertan, line 23: nit: value > ptr (in this case)

share/availability/light/sample.go
```diff
@@ -7,12 +7,31 @@ import (
...
 )
 
+// SamplingResult holds the available and remaining samples.
+type SamplingResult struct {
+	Available []Sample `json:"available"`
+	Remaining []Sample `json:"remaining"`
+}
+
 // Sample represents a coordinate in a 2D data square.
 type Sample struct {
```
Wondertan, line 17: I just realized that we would be better off using the `SampleIndex` defined in #3891. Ideally, having two types for the same thing should be avoided. If we end up postponing the patch release, I think we should stick to a single type.

@walldiss, @cristaloleg

share/availability/light/availability.go
walldiss: We must wait for store operation to finish before returning to maintain tx properties.

share/availability/light/sample.go
walldiss: Looks like it should be resolved in second PR, that introduces second type.

vgonkivs (approved):

share/availability/light/availability.go
```diff
@@ -75,62 +77,80 @@ func (la *ShareAvailability) SharesAvailable(ctx context.Context, header *header
...
+		// No previous results; create new samples
+		samples = NewSamplingResult(len(dah.RowRoots), int(la.params.SampleAmount))
+	} else {
+		err = json.Unmarshal(data, samples)
 		if err != nil {
 			return err
 		}
+		// Verify total samples count.
+		totalSamples := len(samples.Remaining) + len(samples.Available)
+		if totalSamples != int(la.params.SampleAmount) {
```
Wondertan, line 100: The fact that this param is configurable doesn't work with this error. We could reconsider the need for this param, but that's separate.

Wondertan (approved):
Appeoving to unblock. The error issue can be resolved async

share/availability/light/availability.go
renaynay, line 100: nit: alternatively could have a validation func on SamplingResult

non blockin

renaynay (approved):

PR #3889, opened 2024-10-28
fix(DASer): Log sampling errors in DASer

walldiss:
Log errors in DASer instead of availability


Wondertan (dismissed):

Wondertan (approved):

das/worker.go
```diff
@@ -127,18 +128,25 @@ func (w *worker) sample(ctx context.Context, timeout time.Duration, height uint6
...
-				"failed to sample header",
-				"type", w.state.jobType,
-				"height", h.Height(),
-				"hash", h.Hash(),
-				"square width", len(h.DAH.RowRoots),
-				"data root", h.DAH.String(),
-				"err", err,
-				"finished (s)", time.Since(start),
-			)
+		if errors.Is(err, context.Canceled) {
```
cristaloleg, line 131: Also `context.DeadlineExceeded` ?

cristaloleg (dismissed):

das/worker.go
renaynay: Deadline exceeded would be a failure to retrieve share in given time so it would be a `share ErrNotAvailable`

renaynay (dismissed):

das/worker.go
```diff
@@ -127,18 +127,20 @@ func (w *worker) sample(ctx context.Context, timeout time.Duration, height uint6
...
+			return err
 		}
+
+		log.Errorw(
```
Wondertan, line 134: Was about to comment that we need to see individual failures if we remove the logs from avail 👍🏻

share/availability/light/availability.go
```diff
@@ -97,7 +97,6 @@ func (la *ShareAvailability) SharesAvailable(ctx context.Context, header *header
 	}
 
 	if err := dah.ValidateBasic(); err != nil {
-		log.Errorw("DAH validation failed", "error", err)
 		return err
 	}
```
Wondertan, line 101: Btw, we may want to remove this check entirely. Headers are guaranteed to be valid at this point.

walldiss: Removed in light simplification PR: https://github.com/celestiaorg/celestia-node/pull/3895

renaynay (approved):

Wondertan (approved):

PR #3891, opened 2024-10-28
refactor(share): GetShare -> GetSamples

Wondertan:
Samples were requested individually rather than grouped. This was causing issues for Bitswap in the mode where each request is a distinct session, so the solution is to start grouping those as we originally planned. 

Mainly removes `GetShare` and introduces:
> GetSamples(ctx context.Context, header *header.ExtendedHeader, indices []SampleIndex) ([]Sample, error)

And introduces `SampleIndex` with helper funcs. 

The current state of the PR unblocks me and allows me to iterate further on Bitswap issues, while the remaining TODOs still need to be finished.

TODOs:
- [x] `eds.Accessor` should take the new `shwap.SampleIndex` as param in `Sample` method
- [x] !Add `GetSamples` to the ShareModule. 
  - Need to decide which release we want to put this breaking change in
- [x] Godoc new things
- [x] Fix remaining tests
- [x] Rebase
- [x] Decide on whether `SampleIndex` should be an index or a tuple.

--- 

🎱 mb blonks certified

### DIFF
--- a/nodebuilder/share/share.go
+++ b/nodebuilder/share/share.go
@@ -47,7 +47,7 @@ type Module interface {
 	// GetShare gets a Share by coordinates in EDS.
 	GetShare(ctx context.Context, height uint64, row, col int) (libshare.Share, error)
 	// GetSamples gets sample for given indices.
-	GetSamples(ctx context.Context, header *header.ExtendedHeader, indices []shwap.SampleIndex) ([]shwap.Sample, error)
+	GetSamples(ctx context.Context, header *header.ExtendedHeader, indices []shwap.SampleCoords) ([]shwap.Sample, error)
 	// GetEDS gets the full EDS identified by the given extended header.
 	GetEDS(ctx context.Context, height uint64) (*rsmt2d.ExtendedDataSquare, error)
 	// GetNamespaceData gets all shares from an EDS within the given namespace.
@@ -71,7 +71,7 @@ type API struct {
 		GetSamples func(
 			ctx context.Context,
 			header *header.ExtendedHeader,
-			indices []shwap.SampleIndex,
+			indices []shwap.SampleCoords,
 		) ([]shwap.Sample, error) `perm:"read"`
 		GetEDS func(
 			ctx context.Context,
@@ -99,7 +99,7 @@ func (api *API) GetShare(ctx context.Context, height uint64, row, col int) (libs
 }
 
 func (api *API) GetSamples(ctx context.Context, header *header.ExtendedHeader,
-	indices []shwap.SampleIndex,
+	indices []shwap.SampleCoords,
 ) ([]shwap.Sample, error) {
 	return api.Internal.GetSamples(ctx, header, indices)
 }
@@ -132,9 +132,9 @@ func (m module) GetShare(ctx context.Context, height uint64, row, col int) (libs
 		return libshare.Share{}, err
 	}
 
-	idx := shwap.SampleIndex{Row: row, Col: col}
+	idx := shwap.SampleCoords{Row: row, Col: col}
 
-	smpls, err := m.getter.GetSamples(ctx, header, []shwap.SampleIndex{idx})
+	smpls, err := m.getter.GetSamples(ctx, header, []shwap.SampleCoords{idx})
 	if err != nil {
 		return libshare.Share{}, err
 	}
@@ -143,7 +143,7 @@ func (m module) GetShare(ctx context.Context, height uint64, row, col int) (libs
 }
 
 func (m module) GetSamples(ctx context.Context, header *header.ExtendedHeader,
-	indices []shwap.SampleIndex,
+	indices []shwap.SampleCoords,
 ) ([]shwap.Sample, error) {
 	return m.getter.GetSamples(ctx, header, indices)
 }
--- a/nodebuilder/share/share.go
+++ b/nodebuilder/share/share.go
@@ -126,17 +126,13 @@ type module struct {
 	hs     headerServ.Module
 }
 
-// TODO(@Wondertan): break
 func (m module) GetShare(ctx context.Context, height uint64, row, col int) (libshare.Share, error) {
 	header, err := m.hs.GetByHeight(ctx, height)
 	if err != nil {
 		return libshare.Share{}, err
 	}
 
-	idx, err := shwap.SampleIndexFromCoordinates(row, col, len(header.DAH.RowRoots))
-	if err != nil {
-		return libshare.Share{}, err
-	}
+	idx := shwap.SampleIndex{Row: row, Col: col}
 
 	smpls, err := m.getter.GetSamples(ctx, header, []shwap.SampleIndex{idx})
 	if err != nil {
--- a/nodebuilder/share/share.go
+++ b/nodebuilder/share/share.go
@@ -8,6 +8,7 @@ import (
 	libshare "github.com/celestiaorg/go-square/v2/share"
 	"github.com/celestiaorg/rsmt2d"
 
+	"github.com/celestiaorg/celestia-node/header"
 	headerServ "github.com/celestiaorg/celestia-node/nodebuilder/header"
 	"github.com/celestiaorg/celestia-node/share"
 	"github.com/celestiaorg/celestia-node/share/eds"
@@ -45,6 +46,8 @@ type Module interface {
 	SharesAvailable(ctx context.Context, height uint64) error
 	// GetShare gets a Share by coordinates in EDS.
 	GetShare(ctx context.Context, height uint64, row, col int) (libshare.Share, error)
+	// GetSamples gets sample for given indices.
+	GetSamples(ctx context.Context, header *header.ExtendedHeader, indices []shwap.SampleIndex) ([]shwap.Sample, error)
 	// GetEDS gets the full EDS identified by the given extended header.
 	GetEDS(ctx context.Context, height uint64) (*rsmt2d.ExtendedDataSquare, error)
 	// GetNamespaceData gets all shares from an EDS within the given namespace.
@@ -65,6 +68,11 @@ type API struct {
 			height uint64,
 			row, col int,
 		) (libshare.Share, error) `perm:"read"`
+		GetSamples func(
+			ctx context.Context,
+			header *header.ExtendedHeader,
+			indices []shwap.SampleIndex,
+		) ([]shwap.Sample, error) `perm:"read"`
 		GetEDS func(
 			ctx context.Context,
 			height uint64,
@@ -88,6 +96,12 @@ func (api *API) SharesAvailable(ctx context.Context, height uint64) error {
 
 func (api *API) GetShare(ctx context.Context, height uint64, row, col int) (libshare.Share, error) {
 	return api.Internal.GetShare(ctx, height, row, col)
+}
+
+func (api *API) GetSamples(ctx context.Context, header *header.ExtendedHeader,
+	indices []shwap.SampleIndex,
+) ([]shwap.Sample, error) {
+	return api.Internal.GetSamples(ctx, header, indices)
 }
 
 func (api *API) GetEDS(ctx context.Context, height uint64) (*rsmt2d.ExtendedDataSquare, error) {
@@ -130,6 +144,12 @@ func (m module) GetShare(ctx context.Context, height uint64, row, col int) (libs
 	}
 
 	return smpls[0].Share, nil
+}
+
+func (m module) GetSamples(ctx context.Context, header *header.ExtendedHeader,
+	indices []shwap.SampleIndex,
+) ([]shwap.Sample, error) {
+	return m.getter.GetSamples(ctx, header, indices)
 }
 
 func (m module) GetEDS(ctx context.Context, height uint64) (*rsmt2d.ExtendedDataSquare, error) {
--- a/nodebuilder/share/share.go
+++ b/nodebuilder/share/share.go
@@ -112,12 +112,24 @@ type module struct {
 	hs     headerServ.Module
 }
 
+// TODO(@Wondertan): break
 func (m module) GetShare(ctx context.Context, height uint64, row, col int) (libshare.Share, error) {
 	header, err := m.hs.GetByHeight(ctx, height)
 	if err != nil {
 		return libshare.Share{}, err
 	}
-	return m.getter.GetShare(ctx, header, row, col)
+
+	idx, err := shwap.SampleIndexFromCoordinates(row, col, len(header.DAH.RowRoots))
+	if err != nil {
+		return libshare.Share{}, err
+	}
+
+	smpls, err := m.getter.GetSamples(ctx, header, []shwap.SampleIndex{idx})
+	if err != nil {
+		return libshare.Share{}, err
+	}
+
+	return smpls[0].Share, nil
 }
 
 func (m module) GetEDS(ctx context.Context, height uint64) (*rsmt2d.ExtendedDataSquare, error) {
--- a/share/availability/light/availability.go
+++ b/share/availability/light/availability.go
@@ -126,10 +126,10 @@ func (la *ShareAvailability) SharesAvailable(ctx context.Context, header *header
 	}
 
 	smpls, err := la.getter.GetSamples(ctx, header, idxs)
-	if errors.Is(ctx.Err(), context.Canceled) {
+	if errors.Is(err, context.Canceled) {
 		// Availability did not complete due to context cancellation, return context error instead of
 		// share.ErrNotAvailable
-		return ctx.Err()
+		return err
 	}
 	if len(smpls) == 0 {
 		return share.ErrNotAvailable
--- a/share/availability/light/availability.go
+++ b/share/availability/light/availability.go
@@ -117,12 +117,7 @@ func (la *ShareAvailability) SharesAvailable(ctx context.Context, header *header
 
 	idxs := make([]shwap.SampleIndex, len(samples.Remaining))
 	for i, s := range samples.Remaining {
-		idx, err := shwap.SampleIndexFromCoordinates(s.Row, s.Col, len(dah.RowRoots))
-		if err != nil {
-			return err
-		}
-
-		idxs[i] = idx
+		idxs[i] = shwap.SampleIndex{Row: s.Row, Col: s.Col}
 	}
 
 	smpls, err := la.getter.GetSamples(ctx, header, idxs)
@@ -138,15 +133,10 @@ func (la *ShareAvailability) SharesAvailable(ctx context.Context, header *header
 	var failedSamples []Sample
 
 	for i, smpl := range smpls {
-		row, col, err := idxs[i].Coordinates(len(dah.RowRoots))
-		if err != nil {
-			return err
-		}
-
 		if smpl.IsEmpty() {
-			failedSamples = append(failedSamples, Sample{Row: row, Col: col})
+			failedSamples = append(failedSamples, Sample{Row: idxs[i].Row, Col: idxs[i].Col})
 		} else {
-			samples.Available = append(samples.Available, Sample{Row: row, Col: col})
+			samples.Available = append(samples.Available, Sample{Row: idxs[i].Row, Col: idxs[i].Col})
 		}
 	}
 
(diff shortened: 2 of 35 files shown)


share/shwap/p2p/bitswap/getter.go
```diff
@@ -112,36 +111,34 @@ func (g *Getter) GetShares(
...
+		}
 		return nil, err
 	}
 
-	shares := make([]libshare.Share, len(blks))
+	smpls := make([]shwap.Sample, len(blks))
 	for i, blk := range blks {
-		shares[i] = blk.(*SampleBlock).Container.Share
+		c := blk.(*SampleBlock).Container
+		c.Proof = &nmt.Proof{} // TODO
```
cristaloleg, line 136: Without this line `TestPruneAll` & `TestPrunePartialFailed` will fail 'cause `smpl.IsEmpty()` will be true. Not sure how to fetch `.Proof` properly here, @Wondertan any ideas?

CC: @vgonkivs

share/availability/light/sample.go
```diff
@@ -50,8 +50,8 @@ func selectRandomSamples(squareSize, sampleCount int) []Sample {
 	return maps.Keys(samples)
 }
 
-func randInt(max int) int {
-	n, err := crand.Int(crand.Reader, big.NewInt(int64(max)))
+func randInt(m int) int {
```
cristaloleg, line 53: Linter suggestion

share/availability/light/availability_test.go
```diff
@@ -252,51 +260,49 @@ func TestParallelAvailability(t *testing.T) {
...
-func (g onceGetter) checkOnce(t *testing.T) {
+func (g successGetter) sampledList() []Sample {
 	g.Lock()
 	defer g.Unlock()
-	for s, count := range g.sampled {
-		if count > 1 {
-			t.Errorf("sample %v was called more than once", s)
-		}
-	}
+	return slices.Collect(maps.Keys(g.sampled))
```
cristaloleg, line 278: forgive me father i have sinned

nodebuilder/share/share.go
```diff
@@ -112,12 +126,30 @@ type module struct {
 	hs     headerServ.Module
 }
 
+// TODO(@Wondertan): break
```
Wondertan, line 129: I guess we can remove the TODO for now and park the break into the issue.

share/shwap/p2p/bitswap/getter.go
Wondertan: Looking into it now.

We definitely should find a different option, as this one basically erases the part of the response users may want.

Wondertan: Ok, there is a bug in the test itself. The blocks are always coming empty.

cc @walldiss.

Wondertan: Dunno how that worked previously, like it was succeeding for some other reason, like a false positive

Wondertan: Almost finished the fix

Wondertan: e13f2c24

share/availability/light/availability_test.go
```diff
@@ -359,31 +351,23 @@ func TestPruneAll(t *testing.T) {
...
 	getter.Start()
 	defer getter.Stop()
 
 	// Create a new ShareAvailability instance and sample the shares
 	sampleAmount := uint(20)
 	avail := NewShareAvailability(getter, ds, clientBs, WithSampleAmount(sampleAmount))
-	err = avail.SharesAvailable(ctx, h)
-	require.NoError(t, err)
+	err := avail.SharesAvailable(ctx, h)
+	require.Error(t, err)
```
Wondertan, line 370: This must have been error, because we fail half of requests and thus the error comes up

share/availability/light/availability_test.go
```diff
@@ -463,3 +446,55 @@ func countKeys(ctx context.Context, t *testing.T, bs blockstore.Blockstore) int
...
+
+	client := newClient(ctx, net.Hosts()[2], bstore)
+
+	err = net.ConnectAllButSelf()
+	require.NoError(t, err)
+	return client
+}
+
+func newServer(ctx context.Context, host host.Host, store blockstore.Blockstore) {
+	net := bitswap.NewNetwork(host, "test")
```
Wondertan, line 474: For `bitswap.Getter/Fetch` to work, requests must go through `hasher`; thus, real Bitswap should be used.

Wondertan: the bitswap setup is copied from bitswap tests and can be deduplicaated if we want to

Wondertan (commented):
Pointing out fixes made in e13f2c24

share/availability/light/availability.go
```diff
@@ -114,40 +113,43 @@ func (la *ShareAvailability) SharesAvailable(ctx context.Context, header *header
...
-	)
-
-	log.Debugw("starting sampling session", "height", header.Height())
-
-	// remove one second from the deadline to ensure we have enough time to process the results
-	samplingCtx, cancel := context.WithCancel(ctx)
-	if deadline, ok := ctx.Deadline(); ok {
-		samplingCtx, cancel = context.WithDeadline(ctx, deadline.Add(-time.Second))
-	}
-	defer cancel()
```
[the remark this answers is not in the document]
Wondertan: We should also modify the TestPrunePartialFailed to ensure that we save data on context cancellation

Wondertan:
So how would you name the structure? Would it be something like?

```go

type SampleCoord struct {
    Row, Col int
}

func NewSampleCoordFromIndex(index, edsSize int) SampleCoord

func (sc SampleCoord) Index(edsSize int) int 
```


Then we can also avoid defining the `Sample` struct in `availability/light` altogether.

walldiss:
> So how would you name the structure? Would it be something like?
>  type SampleCoord struct {
  >    Row, Col int
 > }

SampleCoords is good. Alternative can be just `Coordinates` 

> func NewSampleCoordFromIndex(index, edsSize int) SampleCoord
> func (sc SampleCoord) Index(edsSize int) int 

I don't follow what is `index` here. Is it 1D index? I assume we won't need 1D index with SampleCoord struct

> Then we can also avoid defining the Sample struct in availability/light altogether.

yes!  Sample struct can be replaced with new struct (keep marshalling for backwards compat)

Wondertan:
> I don't follow what is index here. Is it 1D index? I assume we won't need 1D index with SampleCoord struct. And won't need related methods either

Yes, it is a 1D index. I think it's worth having it for convenience, as there are still cases where 1D is used. It is not necessary internally but externally. Users, like Diego and Rollkit operated with index. E.g Rollkit fraud proofs point to 1D index, rather than a coord.

nodebuilder/share/share.go
cristaloleg: Done https://github.com/celestiaorg/celestia-node/issues/3951

walldiss:
> Yes, it is a 1D index. I think it's worth having it for convenience, as there are still cases where 1D is used. It is not necessary internally but externally. Users, like Diego and Rollkit operated with index. E.g Rollkit fraud proofs point to 1D index, rather than a coord.

Well, we don't have any 1D indexes in our code base internal API, not in external json-rpc API we have. Introduction of 1D index brings lots of conversion from 1D to 2D and back here and there as we can see in this PR. Which suggests that we don't need it anywhere. We don't really need to introduce the second way to express same thing. 

>  Diego and Rollkit operated with index

Do you know why they use it instead of 2D? Might be aswell be historical or random decision. Perhaps we should not introduce additional 1D index unless there is reason to have it

Wondertan:
We should keep 1D convenience methods, while we default coordinates tuple.

share/shwap/getter.go
```diff
@@ -29,8 +29,10 @@ var (
 //
 //go:generate mockgen -destination=getters/mock/getter.go -package=mock . Getter
 type Getter interface {
-	// GetShare gets a Share by coordinates in EDS.
-	GetShare(ctx context.Context, header *header.ExtendedHeader, row, col int) (libshare.Share, error)
+	// GetSamples gets samples by their indices.
+	// Returns Sample slice with requested number of samples in the requested order.
+	// May return partial response with some samples being empty if they weren't found.
+	GetSamples(ctx context.Context, header *header.ExtendedHeader, indices []SampleIndex) ([]Sample, error)
```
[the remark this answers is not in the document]
Wondertan: The sentence after may is supposed to address your question. 

Ideally we add a test suite for Getter that ensures we always return the same amount of samples as requested for all implementations.

nodebuilder/share/share.go
```diff
@@ -45,6 +46,8 @@ type Module interface {
 	SharesAvailable(ctx context.Context, height uint64) error
 	// GetShare gets a Share by coordinates in EDS.
 	GetShare(ctx context.Context, height uint64, row, col int) (libshare.Share, error)
```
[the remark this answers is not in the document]
cristaloleg: https://github.com/celestiaorg/celestia-node/pull/3891#discussion_r1849677926

share/shwap/getter.go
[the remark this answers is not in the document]
walldiss: I was asking about slightly different thing. If 7 out of 16 samples fail to be retrieved with various errors. Would method return any errors in such case along with successful samples?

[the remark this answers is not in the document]
Wondertan: Yes

share/availability/light/availability.go
```diff
@@ -114,40 +113,43 @@ func (la *ShareAvailability) SharesAvailable(ctx context.Context, header *header
...
+		}
+
+		idxs[i] = idx
+	}
+
+	smpls, err := la.getter.GetSamples(ctx, header, idxs)
+	if errors.Is(err, context.Canceled) {
+		// Availability did not complete due to context cancellation, return context error instead of
+		// share.ErrNotAvailable
+		return context.Canceled
```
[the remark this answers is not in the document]
cristaloleg: Fixed

share/availability/light/availability.go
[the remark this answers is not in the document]
cristaloleg: After all created a separate test.

Wondertan (commented):
Approving

share/availability/light/availability_test.go
```diff
@@ -399,38 +384,111 @@ func TestPrunePartialFailed(t *testing.T) {
 	require.False(t, exist)
 }
 
-var _ exchange.SessionExchange = (*fakeSessionExchange)(nil)
+func TestPruneWithCancelledContext(t *testing.T) {
```
walldiss, line 387: Addition of this test and related refactoring looks unrelated to PR, but ok

share/availability/light/availability_test.go
```diff
@@ -399,38 +384,111 @@ func TestPrunePartialFailed(t *testing.T) {
...
+		if err != nil {
+			return nil, err
+		}
+
+		out <- blk
 	}
-	return hfb.Blockstore.Get(ctx, c)
+
+	// sleep 1 second guarantees that we will exhaust context.
+	time.Sleep(time.Second)
```
walldiss, line 489: Why not just wait for context to be finished by reading from <- ctx.Done()?

share/shwap/p2p/bitswap/getter.go
```diff
@@ -74,27 +75,24 @@ func (g *Getter) Stop() {
...
 	ctx context.Context,
 	hdr *header.ExtendedHeader,
-	rowIdxs, colIdxs []int,
-) ([]libshare.Share, error) {
-	if len(rowIdxs) != len(colIdxs) {
-		return nil, fmt.Errorf("row indecies and col indices must be same length")
+	indices []shwap.SampleCoords,
+) ([]shwap.Sample, error) {
+	if len(indices) == 0 {
+		return nil, fmt.Errorf("no sample indicies to fetch")
```
walldiss, line 85: This error seems to be common to all getters, so might worth to make it part of Getter interface

share/shwap/getters/cascade.go
```diff
@@ -41,24 +41,17 @@ func NewCascadeGetter(getters []shwap.Getter) *CascadeGetter {
...
-		span.RecordError(err)
-		return libshare.Share{}, err
-	}
-	get := func(ctx context.Context, get shwap.Getter) (libshare.Share, error) {
-		return get.GetShare(ctx, header, row, col)
+	get := func(ctx context.Context, get shwap.Getter) ([]shwap.Sample, error) {
+		return get.GetSamples(ctx, hdr, indices)
 	}
 
 	return cascadeGetters(ctx, cg.getters, get)
```
walldiss, line 57: cascade getter will retry full operation if single sample failed

walldiss (commented):
There is a problem in cascade getter. If any of requested samples fail, it will retry with full sampleCoordinates set with next fallback getter, throwing away all found samples. There is no proper support added in this PR right now. It would technically work with current setup because bitswap getter is the last, but it will misbehaive in any other case. 
If left as is it may become a huge footgun and nightmare to debug later.

share/availability/light/availability.go
```diff
@@ -129,25 +128,21 @@ func (la *ShareAvailability) SharesAvailable(ctx context.Context, header *header
...
-			} else {
-				samples.Available = append(samples.Available, s)
-			}
-		}(s)
+	smpls, errGetSamples := la.getter.GetSamples(samplingCtx, header, idxs)
+	if len(smpls) == 0 {
+		return share.ErrNotAvailable
+	}
+
+	var failedSamples []Sample
```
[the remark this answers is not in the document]
cristaloleg: You mean `shwap.SampleCoords` ? Well, theoretically yeah, we can store only coords. Gonna update that.

share/availability/light/availability_test.go
cristaloleg: It's related cause we want be sure that cancelled context works too. (This PR isn't about refactoring from the start LOL).

share/shwap/getters/cascade.go
cristaloleg: So...what? I don't get it.

share/shwap/p2p/bitswap/getter.go
cristaloleg: > it part of Getter interface

you mean extract to a variable, right?

Smth like: `var ErrNoSampleIndicies = errors.New("no sample indicies to fetch")`

share/availability/light/availability.go
```diff
@@ -162,16 +157,18 @@ func (la *ShareAvailability) SharesAvailable(ctx context.Context, header *header
 		return fmt.Errorf("store sampling result: %w", err)
 	}
 
-	if errors.Is(ctx.Err(), context.Canceled) {
+	if errors.Is(errGetSamples, context.Canceled) ||
+		errors.Is(errGetSamples, context.DeadlineExceeded) {
```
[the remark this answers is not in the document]
cristaloleg: Okay, removed (will explain in detail in comment with 1sec sleep).

share/availability/light/availability_test.go
cristaloleg: Not that fast. We need to simulate `context.Cancel` error. To do this we need async cancel a context, to be sure that cancelling will happen before return from this method - sleep is needed.

This was the case to simulate `context.DeadlineExceeded` error but have updated to `context.Cancel`. Thanks.

share/availability/light/availability_test.go
```diff
@@ -208,7 +216,6 @@ func TestSharesAvailableFailed(t *testing.T) {
 		}
 
 		// onceGetter should have no more samples stored after the call
-		successfulGetter.checkOnce(t)
```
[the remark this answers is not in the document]
cristaloleg: We don't do this anymore in mock getter due to it's new behaviour. So, it got removed.

share/availability/light/availability_test.go
```diff
@@ -164,7 +163,16 @@ func TestSharesAvailableFailed(t *testing.T) {
...
+		failGetter := mock.NewMockGetter(gomock.NewController(t))
+		ds := datastore.NewMapDatastore()
+		avail := NewShareAvailability(failGetter, ds, nil)
+
+		// Getter doesn't have the eds, so it should fail for all samples
+		mockSamples := min(int(avail.params.SampleAmount), 2*len(tt.eh.DAH.RowRoots))
+		failGetter.EXPECT().
+			GetSamples(gomock.Any(), gomock.Any(), gomock.Any()).
+			Return(make([]shwap.Sample, mockSamples), shrex.ErrNotFound).
+			AnyTimes()
```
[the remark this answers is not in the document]
cristaloleg: WDYT ? `AnyTimes()` ?

share/availability/light/availability.go
```diff
@@ -129,25 +128,21 @@ func (la *ShareAvailability) SharesAvailable(ctx context.Context, header *header
...
-				log.Debugw("error fetching share", "height", header.Height(), "row", s.Row, "col", s.Col)
-				failedSamples = append(failedSamples, s)
-			} else {
-				samples.Available = append(samples.Available, s)
-			}
-		}(s)
+	smpls, errGetSamples := la.getter.GetSamples(samplingCtx, header, idxs)
+	if len(smpls) == 0 {
+		return share.ErrNotAvailable
+	}
```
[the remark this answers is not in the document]
cristaloleg: Looking at `shwap.Getter` doc-comment yes should be impossible. But we had this thing in tests so...looks like it's behaviour. I would like to ping Hlib here but that's not possible and useless now :(

Any desired behaviour?

share/availability/light/availability.go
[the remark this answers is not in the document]
cristaloleg: Hm, not that fast. We have 

```
// SamplingResult holds the available and remaining samples.
type SamplingResult struct {
	Available []Sample `json:"available"`
	Remaining []Sample `json:"remaining"`
}
```

Changing it's field to `[]shwap.SampleCoords` probably not what we want now.

share/availability/light/availability.go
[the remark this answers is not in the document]
walldiss: Looks like it should be enforced by interface definition. The only way to do it is to have tests for every getter, that ensusers it returns same amount of samples as requested. Or returns error, if requested amount os 0.

share/availability/light/availability.go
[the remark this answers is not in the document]
walldiss: Lets change for new type. Just make sure to maintain same encoding schema, so it does not break older nodes

share/shwap/getters/cascade.go
vgonkivs: Here is the full explanation of the issue @cristaloleg 
https://github.com/celestiaorg/celestia-node/pull/3891#pullrequestreview-2459066562

share/availability/light/availability.go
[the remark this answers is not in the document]
cristaloleg: done, thank you!

share/availability/light/availability_test.go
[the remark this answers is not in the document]
cristaloleg: Change looks green after rebase on the latest `main`.

share/shwap/p2p/bitswap/getter.go
walldiss: yes. And add it to description of `GetSamples` method of `Getter` interface

share/availability/light/availability_test.go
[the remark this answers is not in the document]
cristaloleg: And it's back ;)

share/availability/light/availability.go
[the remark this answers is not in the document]
walldiss: Since we have no bitswap.Getter tests we can only rely on availability pkg.

share/shwap/p2p/bitswap/getter.go
cristaloleg: Done

share/shwap/getters/cascade.go
walldiss: Turned into an issue https://github.com/celestiaorg/celestia-node/issues/3964

walldiss (approved):

vgonkivs (approved):

PR #3894, opened 2024-10-28
fix(shwap/bitswap): Blockstore.GetSize: getting size with no compute

Wondertan:
Bitswap prioritizes peers based on their active/pending work and the priority that peers set for requests(work) themselves. The prioritization happens on the `Get` operation of `Blockstore` not `GetSize`. However, we currently do the most expensive work in `GetSize` followed by `Get`. It is still part of the same WANT_BLOCK request(since #3813), and `GetSize` work is usually cached for `Get` to catch it. Still, the prioritization makes it so that the time between `GetSize` and `Get` can be pretty long, inducing cache misses and redoing the same work again, which is confirmed by profiles. Also, doing the expensive part in `GetSize` avoids Bitswap server's rate limiting. 

All this brings the need to make `GetSize` as lightweight as possible, leaving actual proof computed to `Get`. This PR achieves this by defining a constant max block size for every type block.

--- 

🎱 mb blonks certified


cristaloleg:
Lint fail https://github.com/celestiaorg/celestia-node/actions/runs/11556547784/job/32164506582?pr=3894#step:4:31

walldiss:
If GetSize is used only for prioritisation of bitswap server hanling of requests, perhaps don't really need to calculate sizes. Alternative could be to return const values depending on type of the cid and our estimate of proper priority / load of handling. For examples:
- samples: 100
- rows: 10000 (much heavier and less important than namespace data)
- namespace data: 5000 

This would allow us to remove the need for heavy reads to get the priority for serving of requests and will easy the pressure on cache.

Wondertan:
In sync, we agreed to keep the Size computation based on real blocksize to be safe. Making all sizes fixed to max block size may limit bitswap's throughput for small block sizes.

share/shwap/p2p/bitswap/sample_block.go
```diff
@@ -71,6 +74,12 @@ func (sb *SampleBlock) Height() uint64 {
 	return sb.ID.Height
 }
 
+func (sb *SampleBlock) Size(ctx context.Context, acc eds.Accessor) (int, error) {
+	squareSize := acc.Size(ctx)
+	sampleSize := libshare.ShareSize + share.AxisRootSize*int(math.Log2(float64(squareSize)))
```
renaynay, line 79: is there no helper util we can use to determine proof size for a share based on dah?

renaynay (commented):
Do you need additional tests for this? Are the changes already covered by some test suite?

share/shwap/p2p/bitswap/sample_block.go
Wondertan: I am not aware of such

Wondertan:
It is covered

walldiss (dismissed):

vgonkivs (dismissed):

Wondertan:
Disabling the automerge for now. I want to do another change here

renaynay (approved):
utack

walldiss (approved):

renaynay (approved):

walldiss (approved):

PR #3895, opened 2024-10-28
refactor(share/availability): simplify light availability

walldiss:
- simplification refactoring of Samples selection code
- remove squareSampler
- remove binary encode/decode in favour of json
- removal of no longer relevant TODOs
- rename "cache" naming artefacts from the past
- use height in logs instead of root hash


share/availability/light/availability.go
```diff
@@ -130,7 +126,10 @@ func (la *ShareAvailability) SharesAvailable(ctx context.Context, header *header
 	}
 
 	// store the result of the sampling session
-	bs := encodeSamples(failedSamples)
+	bs, err := json.Marshal(failedSamples)
```
Wondertan, line 129: Why do we break the serialization? We save a few loc for a much slower and bloated serialization. I don't think that's worth it at all.

share/availability/light/sample.go
```diff
@@ -1,104 +1,41 @@
...
-	"encoding/binary"
-	"errors"
 	"math/big"
+
+	"golang.org/x/exp/maps"
 )
 
-// Sample is a point in 2D space over square.
+// Sample represents a coordinate in a 2D data square.
 type Sample struct {
```
Wondertan, line 11: Let's make it private while we are on it. Can't recall it being used elsewhere. Please check

share/availability/light/availability.go
walldiss: json encoding speed is 100-1000mb/s. It would take for our ~100 bytes encoding around 1ns. Does not seems to be a performance hit to me.

share/availability/light/sample.go
walldiss: It would be used externally by `pruner` pkg very soon.

share/availability/light/availability.go
Wondertan: The point is, there is no good reason to do it in the first place. Also, we don't need the JSON bloat.

walldiss: As you pointed out in original comment, the benefit is simplified code. Custom encoding does not bring much of performance benefit here.

Wondertan: It does being storage overhead tho

renaynay (commented):
Additional request is to move dah validation check in `availability.go` earlier in the `SharesAvailability` function -- ideally in the beginning. It doesn't make sense where it currently is.

share/availability/light/sample.go
Wondertan: Kk

share/availability/light/availability.go
Wondertan: Discussed in sync that breaking is fine:
* We don't store successful samples, so the only way to prune those is by resyncing, which is equivalent to breaking. 
* We are still allowed to break

Wondertan: Another missed thing is that the store state is changing in https://github.com/celestiaorg/celestia-node/pull/3887 to contain successful samples. We could store two states separately to not break, but again, it still requires resyncing to cleanup old samples.

walldiss:
Had to rebase the PR on main, and commit history is lost. Here is what changed after last reviews
- Removed uneccesary dah validation from both light and full availability
- clean up remaining logs to use height.
- changed uint -> int for Row and Col in Sample

renaynay (dismissed):

Wondertan (approved):

renaynay (approved):

PR #3898, opened 2024-10-28
chore: remove unused code

walldiss:
Fine collection of skeletons


share/shwap/sample.go
```diff
@@ -14,9 +14,6 @@ import (
 	"github.com/celestiaorg/celestia-node/share/shwap/pb"
 )
 
-// SampleName is the name identifier for the sample container.
-const SampleName = "sample_v0"
-
```
Wondertan, line 19: Keep the names, for future shrex support

share/empty.go
```diff
@@ -28,18 +28,11 @@ func EmptyEDS() *rsmt2d.ExtendedDataSquare {
 	return emptyBlockEDS
 }
 
-// EmptyBlockShares returns the shares of the empty block.
-func EmptyBlockShares() []libshare.Share {
-	initEmpty()
-	return emptyBlockShares
-}
-
```
Wondertan, line 36: I would keep this one. You never know

header/serde.go
```diff
@@ -64,37 +64,6 @@ func UnmarshalExtendedHeader(data []byte) (*ExtendedHeader, error) {
...
-	pb.ValidatorSet = valSet
-	dah, err := eh.DAH.ToProto()
-	if err != nil {
-		return nil, err
-	}
-	pb.Dah = dah
-	return pb, nil
-}
-
-func ProtoToExtendedHeader(pb *header_pb.ExtendedHeader) (*ExtendedHeader, error) {
```
Wondertan, line 85: Wait, how do we serialize then?

share/empty.go
walldiss: Shares can be obtained from chained call EmptyEDS.FlattenedODS(), so there is not much value in additional global stored shares

header/serde.go
walldiss: We serialise by MarshalExtendedHeader/UnmarshalExtendedHeader. Those are converter fucntions, that are not really needed

share/empty.go
Wondertan: But its ods, not eds

walldiss: The idea is similar, there is `Flattened()` for all eds shares

Wondertan (approved):

cristaloleg (approved):

PR #3905, opened 2024-10-30
refactor(share): GetShare -> GetSamples

cristaloleg:
Continuation for https://github.com/celestiaorg/celestia-node/pull/3891


blob/service_test.go
```diff
@@ -904,7 +915,7 @@ func createService(ctx context.Context, t testing.TB, shares []libshare.Share) *
 		})
 	shareGetter.EXPECT().GetSamples(gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes().
 		DoAndReturn(func(ctx context.Context, h *header.ExtendedHeader, indices []shwap.SampleIndex) ([]shwap.Sample, error) {
-			return smpls, nil
+			return []shwap.Sample{}, nil
```
cristaloleg, line 918: No idea what should be here.

blob/service_test.go
```diff
@@ -580,10 +588,13 @@ func TestService_GetAllWithoutPadding(t *testing.T) {
...
 		row, col := calculateIndex(len(h.DAH.RowRoots), blob.index)
-		sh, err := service.shareGetter.GetShare(ctx, h, row, col)
+		idx, err := shwap.SampleIndexFromCoordinates(row, col, len(h.DAH.RowRoots))
+		require.NoError(t, err)
+
+		smpls, err := service.shareGetter.GetSamples(ctx, h, []shwap.SampleIndex{idx})
 		require.NoError(t, err)
 
-		assert.Equal(t, sh, resultShares[shareOffset])
+		assert.Equal(t, smpls[0].Share, resultShares[shareOffset])
```
cristaloleg, line 597: Should it be always `[0]` or somethings else? `idx` ?

blob/service_test.go
```diff
@@ -521,10 +526,13 @@ func TestService_Get(t *testing.T) {
...
 		row, col := calculateIndex(len(h.DAH.RowRoots), b.index)
-		sh, err := service.shareGetter.GetShare(ctx, h, row, col)
+		idx, err := shwap.SampleIndexFromCoordinates(row, col, len(h.DAH.RowRoots))
 		require.NoError(t, err)
 
-		assert.Equal(t, sh, resultShares[shareOffset], fmt.Sprintf("issue on %d attempt", i))
+		smpls, err := service.shareGetter.GetSamples(ctx, h, []shwap.SampleIndex{idx})
+		require.NoError(t, err)
+
+		assert.Equal(t, smpls[0].Share, resultShares[shareOffset], fmt.Sprintf("issue on %d attempt", i))
```
cristaloleg, line 535: Should it be always [0] or somethings else? idx ?

blob/service_test.go
```diff
@@ -487,10 +489,13 @@ func TestService_GetSingleBlobWithoutPadding(t *testing.T) {
...
 	row, col := calculateIndex(len(h.DAH.RowRoots), newBlob.index)
-	sh, err := service.shareGetter.GetShare(ctx, h, row, col)
+	idx, err := shwap.SampleIndexFromCoordinates(row, col, len(h.DAH.RowRoots))
+	require.NoError(t, err)
+
+	smpls, err := service.shareGetter.GetSamples(ctx, h, []shwap.SampleIndex{idx})
 	require.NoError(t, err)
 
-	assert.Equal(t, sh, resultShares[0])
+	assert.Equal(t, smpls[0].Share, resultShares[0])
```
cristaloleg, line 498: Should it be always [0] or somethings else? idx ?

blob/service_test.go
```diff
@@ -145,9 +145,11 @@ func TestBlobService_Get(t *testing.T) {
 				shareOffset := 0
 				for i := range blobs {
 					row, col := calculateIndex(len(h.DAH.RowRoots), blobs[i].index)
-					sh, err := service.shareGetter.GetShare(ctx, h, row, col)
+					idx, err := shwap.SampleIndexFromCoordinates(row, col, len(h.DAH.RowRoots))
 					require.NoError(t, err)
-					require.True(t, bytes.Equal(sh.ToBytes(), resultShares[shareOffset].ToBytes()),
+					smpls, err := service.shareGetter.GetSamples(ctx, h, []shwap.SampleIndex{idx})
+					require.NoError(t, err)
+					require.True(t, bytes.Equal(smpls[0].Share.ToBytes(), resultShares[shareOffset].ToBytes()),
```
cristaloleg, line 152: Should it be always [0] or somethings else? idx ?

nodebuilder/share/share.go
```diff
@@ -45,6 +45,7 @@ type Module interface {
 	SharesAvailable(ctx context.Context, height uint64) error
 	// GetShare gets a Share by coordinates in EDS.
 	GetShare(ctx context.Context, height uint64, row, col int) (libshare.Share, error)
+	GetSamples(ctx context.Context) (shwap.Sample, error)
```
cristaloleg, line 48: From parent PR #3891 

>  !Add GetSamples to the ShareModule.

Are there params for this method?

share/availability/light/availability.go
```diff
@@ -113,10 +113,10 @@ func (la *ShareAvailability) SharesAvailable(ctx context.Context, header *header
 	}
 
 	smpls, err := la.getter.GetSamples(ctx, header, idxs)
-	if errors.Is(ctx.Err(), context.Canceled) {
+	if errors.Is(err, context.Canceled) {
```
cristaloleg, line 116: Better to check `err` rather than `ctx.Err`

Wondertan: Btw, we will have to rebase on recent changes from Vlad here

share/eds/accessor.go
```diff
@@ -26,7 +26,7 @@ type Accessor interface {
 	// choose which axis to use for proof. Chosen axis for proof should be indicated in the returned
 	// Sample.
 	// TODO(@Wondertan): change to SampleIndex
-	Sample(ctx context.Context, rowIdx, colIdx int) (shwap.Sample, error)
+	Sample(ctx context.Context, rowIdx, colIdx int, sampleIdx shwap.SampleIndex) (shwap.Sample, error)
```
Wondertan, line 29: Btw, the idea is to use SampleIndex instead of rowIdx, colIdx

Wondertan (changes requested):

share/eds/accessor.go
cristaloleg: So, drop these 2 right?

Wondertan: The SampleIndex has Coordinates method that gives you the equivalent

cristaloleg: Done in most places. Something should be updated in tests.

share/availability/light/availability.go
cristaloleg: Well, this PR and branch are targeting your PR. So, firstly this PR should be merged into yours and then rebased on Vlad's changes.

share/shwap/p2p/bitswap/sample_block.go
```diff
@@ -86,7 +86,7 @@ func (sb *SampleBlock) Marshal() ([]byte, error) {
 }
 
 func (sb *SampleBlock) Populate(ctx context.Context, eds eds.Accessor) error {
-	smpl, err := eds.Sample(ctx, sb.ID.RowIndex, sb.ID.ShareIndex)
+	smpl, err := eds.Sample(ctx, shwap.SampleIndex(sb.ID.ShareIndex))
```
Wondertan, line 89: ShareIndex is only the column

share/eds/rsmt2d.go
```diff
@@ -46,8 +46,12 @@ func (eds *Rsmt2D) AxisRoots(context.Context) (*share.AxisRoots, error) {
...
 func (eds *Rsmt2D) Sample(
 	_ context.Context,
-	rowIdx, colIdx int,
+	idx shwap.SampleIndex,
 ) (shwap.Sample, error) {
+	rowIdx, colIdx, err := idx.Coordinates(shwap.EdsIDSize)
+	if err != nil {
+		return shwap.Sample{}, err
+	}
 	return eds.SampleForProofAxis(rowIdx, colIdx, rsmt2d.Row)
```
Wondertan, line 55: We should use the SampleIndex here as well. The objective is to use a single type everywhere

store/file/ods.go
```diff
@@ -228,28 +228,28 @@ func (o *ODS) Close() error {
...
 	// - For the fourth quadrant, it does not matter which axis we read because we need to read full ODS
 	//   to calculate the sample
-	axisType, axisIdx, shrIdx := rsmt2d.Row, rowIdx, colIdx
-	if colIdx < o.size()/2 && rowIdx >= o.size()/2 {
-		axisType, axisIdx, shrIdx = rsmt2d.Col, colIdx, rowIdx
-	}
+	axisType, axisIdx := rsmt2d.Row, 0 // rowIdx, colIdx
+	// if colIdx < o.size()/2 && rowIdx >= o.size()/2 {
+	// 	axisType, axisIdx, shrIdx = rsmt2d.Col, colIdx, rowIdx
+	// }
```
Wondertan, line 242: we still need to swap coordinates here.

share/eds/rsmt2d.go
cristaloleg: Ok, `SampleForProofAxis` is now:

```
func (eds *Rsmt2D) SampleForProofAxis(
	idx shwap.SampleIndex,
	proofType rsmt2d.Axis,
) (shwap.Sample, error) {
```

store/file/ods.go
cristaloleg: Looks like we don't need `rowIdx, colIdx` (`idx` is enough here). The only question I have is: how to get from `SampleIndex` which axis do we want?

share/availability/light/availability.go
Wondertan: I didn't mean this PR, but this code here

share/shwap/p2p/bitswap/sample_block.go
Wondertan: The data structure we work with is a table of Rows and Columns. Cells in the table are called shares. Every share can be addressed by coordinates or index. Coordinate in 2D space and index in 1D vector space.

In this case ShareIndex is ColumnIndex(I know it might be confusing in this context), so 
(RowIndex; ShareIndex) is a coordinate.

SampleIndex is an index in 1D space, so we cant use ShareIndex ad SampleIndex.

P.S. Sample is Share + Proof, so in the context of positiom Share and Sample can be used interchangeably.

store/file/ods.go
cristaloleg: Fixed

share/shwap/p2p/bitswap/sample_block.go
cristaloleg: Updated.

blob/service_test.go
Wondertan: If it needs only a single share, and it seems like that is the case, then yes.

blob/service_test.go
Wondertan: Same as above

blob/service_test.go
Wondertan: Same as above

store/file/ods_q4.go
```diff
@@ -122,9 +122,9 @@ func (odsq4 *ODSQ4) AxisRoots(ctx context.Context) (*share.AxisRoots, error) {
 	return odsq4.ods.AxisRoots(ctx)
 }
 
-func (odsq4 *ODSQ4) Sample(ctx context.Context, rowIdx, colIdx int) (shwap.Sample, error) {
+func (odsq4 *ODSQ4) Sample(ctx context.Context, idx shwap.SampleIndex) (shwap.Sample, error) {
 	// use native AxisHalf implementation, to read axis from q4 quadrant when possible
-	half, err := odsq4.AxisHalf(ctx, rsmt2d.Row, rowIdx)
+	half, err := odsq4.AxisHalf(ctx, rsmt2d.Row, 0)
```
Wondertan, line 127: This should still be rowIdx. Do tests pass with it being 0?

share/shwap/p2p/bitswap/sample_block.go
```diff
@@ -86,7 +86,8 @@ func (sb *SampleBlock) Marshal() ([]byte, error) {
 }
 
 func (sb *SampleBlock) Populate(ctx context.Context, eds eds.Accessor) error {
-	smpl, err := eds.Sample(ctx, sb.ID.RowIndex, sb.ID.ShareIndex)
+	width := eds.Size(ctx)
+	smpl, err := eds.Sample(ctx, shwap.SampleIndex(width*width-1))
```
Wondertan, line 90: `SampleIndexFromCoordinates(sb.ID.RowIndex, sb.ID.ShareIndex)`  instead

store/file/ods.go
```diff
@@ -246,7 +251,7 @@ func (o *ODS) Sample(ctx context.Context, rowIdx, colIdx int) (shwap.Sample, err
 		return shwap.Sample{}, fmt.Errorf("reading axis: %w", err)
 	}
 
-	idx, err := shwap.SampleIndexFromCoordinates(rowIdx, shrIdx, o.Size(ctx))
+	idx, err = shwap.SampleIndexFromCoordinates(rowIdx, shrIdx, o.Size(ctx))
```
Wondertan, line 254:
```suggestion
	idx, err = shwap.SampleIndexFromCoordinates(axisIdx, shrIdx, o.Size(ctx))
```

Wondertan (changes requested):

store/file/ods_q4.go
cristaloleg: Done

store/file/ods.go
cristaloleg: Nice catch, done!

share/eds/testing.go
```diff
@@ -195,9 +199,12 @@ func testSample(
...
 	acc Accessor,
 	roots *share.AxisRoots,
-	rowIdx, colIdx int,
+	idx shwap.SampleIndex,
 ) {
-	shr, err := acc.Sample(ctx, rowIdx, colIdx)
+	shr, err := acc.Sample(ctx, idx)
+	require.NoError(t, err)
+
+	rowIdx, colIdx, err := idx.Coordinates(shwap.EdsIDSize)
```
Wondertan, line 207: This is supposed to be the size of the actual square, so `acc.Size`

share/shwap/sample_test.go
```diff
@@ -25,10 +25,18 @@ func TestSampleValidate(t *testing.T) {
 	for _, proofType := range []rsmt2d.Axis{rsmt2d.Row, rsmt2d.Col} {
 		for rowIdx := 0; rowIdx < odsSize*2; rowIdx++ {
 			for colIdx := 0; colIdx < odsSize*2; colIdx++ {
-				sample, err := inMem.SampleForProofAxis(rowIdx, colIdx, proofType)
+				idx, err := shwap.SampleIndexFromCoordinates(rowIdx, colIdx, odsSize)
```
Wondertan, line 28: It should be `odsSize*2` which is equals to `edsSize`. which is what we get when we call `acc.Size(ctx)`

Wondertan: This should also fix this Error, NoError fix

Wondertan (changes requested):

share/shwap/sample_test.go
cristaloleg: Fixed, thanks.

share/eds/testing.go
cristaloleg: Fixed, thanks.

share/eds/validation_test.go
```diff
@@ -34,7 +34,9 @@ func TestValidation_Sample(t *testing.T) {
 			accessor := &Rsmt2D{ExtendedDataSquare: randEDS}
 			validation := WithValidation(AccessorAndStreamer(accessor, nil))
 
-			_, err := validation.Sample(context.Background(), tt.rowIdx, tt.colIdx)
+			idx, _ := shwap.SampleIndexFromCoordinates(tt.rowIdx, tt.colIdx, shwap.SampleIDSize)
```
Wondertan, line 37: Another one

PR #3906, opened 2024-10-30
refactor(share/availability): Move window and constants to share/availability pkg 

renaynay:
* moves the concept of availability window to the share/availability package where it belongs and changes the names to be more clear -- light.Window -> availability.RequestWindow ; full.Window -> availability.StorageWindow
* limits the use of `availability.Window` to the nodebuilder only, use `time.Duration` everywhere else
* removes archival window in favour of using explicit `time.Duration(0)` -- will be removed entirely in follow-up PRs.


share/availability/window.go
```diff
@@ -0,0 +1,24 @@
...
+const (
+	RequestWindow = 30 * 24 * time.Hour
+	StorageWindow = RequestWindow + time.Hour
+)
+
+type Window time.Duration
+
+func (w Window) Duration() time.Duration {
+	return time.Duration(w)
+}
```
walldiss, line 14: It feels like the only reason for availability.Window to be typed is nodebuilder. Perhaps it can be moved from `share` pkg to `nodebuilder/share` pkg. 

It also seems the type is used to pass not any availability.Window, but just StorageWindow, so if moved, it can be renamed to be share.StorageWindow to be more specific.

walldiss (commented):
Handling pruning windows just got a whole lot easier! It would be awesome to rename the variables so it's clear which window they're using. Right now, names like "window," "samplingWindow," and "availWindow" can be confusing inside the component. Being more specific about which of the two new windows is being used would help a lot.

Don’t forget the functions too, like changing WithAvailabilityWindow to WithStorageWindow. These tweaks would make the code easier to understand for everyone.

renaynay:
@walldiss 
>  Right now, names like "window," "samplingWindow," and "availWindow" can be confusing inside the component

I plan to move those parameters outside of those components entirely (with exception of core pkg n which case I will name it to `StorageWindow`)

share/availability/window.go
renaynay: this would cause an import cycle issue inside of nodebuilder, so would prefer not to spend time on this. Eventually, pruner will be moved into share pkg which will resolve the dependency issue.

walldiss: It is used inside nodebuilder, so it should not be a problem to move it there without cycle dependancy. Perhaps better package for it might be nodebuilder/pruner

share/availability/window.go
```diff
@@ -0,0 +1,24 @@
+package availability
+
+import "time"
+
+const (
+	RequestWindow = 30 * 24 * time.Hour
+	StorageWindow = RequestWindow + time.Hour
+)
+
+type Window time.Duration
```
cristaloleg, line 10: TBH I don't get it why do we need this type. Looks like `time.Duration` is more than enough in all cases. What I'm missing?

cristaloleg:
Should #3912 be closed?

share/availability/window.go
renaynay: nodebuilder :( I can move the definition to nodebuilder in this PR I guess which is what @walldiss wanted

renaynay:
no #3912 is based on this PR @cristaloleg

share/availability/window.go
renaynay: moved it in https://github.com/celestiaorg/celestia-node/pull/3906/commits/1e260c3179c88f0491885fe4dcee08034ff15047

nodebuilder/share/window.go
```diff
@@ -0,0 +1,13 @@
+package share
+
+import (
+	"time"
+)
+
+// Window is a type alias for time.Duration used in nodebuilder
+// to provide sampling windows to their relevant components
+type Window time.Duration
```
vgonkivs, line 9: It seems it is used only to configure fx in Nodebuilder. Can we get rid of this alias in favor of standard time.Duration? or is it, mainly used to improve the readability?

renaynay: https://github.com/celestiaorg/celestia-node/commit/1e260c3179c88f0491885fe4dcee08034ff15047

cristaloleg (approved):

walldiss (approved):

PR #3909, opened 2024-11-01
refactor(availability/light): remove samples amount settings

walldiss:
The setting was allowing users to change amount of samples for light availability. The ability to set the value only causes users to misuse the setting:
- setting value lower than 16 screws security guaranties of DASing
- setting value higher than default consumes more resources of light node, while gain of security is minimal. It would cause unnecessary overhead for light node.

Node can define hardcoded value, which should be part of protocol defined values for light nodes. From code perspective, hardcoded value and removal of settings  will simplify the code of light availability as well as configuration options.


Wondertan:
We need to discuss this before merging

PR #3915, opened 2024-11-04
feat!: adds support for gRPC streaming and removes dependency on RPC

rach-id:
The Celestia-node part of https://github.com/celestiaorg/celestia-core/pull/1513


rach-id:
PR still not fully ready, some tests are still failing. but would be good to get initial opinion + if someone can see if it's providing you with any performance increase

core/fetcher.go
```diff
@@ -33,6 +36,16 @@ func NewBlockFetcher(client Client) *BlockFetcher {
 	}
 }
 
+func (f *BlockFetcher) Stop(ctx context.Context) error {
+	f.cancel()
+	select {
+	case <-f.doneCh:
+		return nil
+	case <-ctx.Done():
+		return fmt.Errorf("fetcher: unsubscribe from new block events: %w", ctx.Err())
```
rach-id, line 45: [question for reviewers]
not sure if this is the right way to stop the resources in here. I guess we can delete this part. What do you think?

go.mod
```diff
@@ -354,10 +353,10 @@ require (
 )
 
 replace (
-	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.25.0-sdk-v0.46.16
+	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.25.0-sdk-v0.46.16.0.20241009080849-0cd36203b354
```
rach-id, line 356: [note for reviewers]
will be updated once we merge the PRs

go.mod
```diff
@@ -354,10 +353,10 @@ require (
...
 	github.com/filecoin-project/dagstore => github.com/celestiaorg/dagstore v0.0.0-20230824094345-537c012aa403
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
 	// broken goleveldb needs to be replaced for the cosmos-sdk and celestia-app
 	github.com/syndtr/goleveldb => github.com/syndtr/goleveldb v1.0.1-0.20210819022825-2ae1ddf74ef7
-	github.com/tendermint/tendermint => github.com/celestiaorg/celestia-core v1.43.0-tm-v0.34.35
+	github.com/tendermint/tendermint => github.com/celestiaorg/celestia-core v1.43.0-tm-v0.34.35.0.20241104132041-3d2b32b7ddf2
```
rach-id, line 361: [note for reviewers]
will be updated once we merge the PRs and make release

go.mod
vgonkivs: what is the status of those PRs? any timeline on when you are going to merge them and make a release?

go.mod
vgonkivs: the same here

go.mod
rach-id: yes, as soon as the prototype is validated, and this PR is ready to merge. I am leaving them open so that we do changes on them if the changes on node require any without having to make multiple releases

vgonkivs: ok. that makes sense. thank you. So once we agree on some solution here, then you'll merge those PRs and make a release.

rach-id: yes

core/fetcher.go
```diff
@@ -2,18 +2,27 @@ package core
...
+	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
+	coregrpc "github.com/tendermint/tendermint/rpc/grpc"
 	"github.com/tendermint/tendermint/types"
 
 	libhead "github.com/celestiaorg/go-header"
 )
 
 const newBlockSubscriber = "NewBlock/Events"
 
+type SignedBlock struct {
```
rach-id, line 19: this is the redefinition of the SignedBlock to fix the memory leak we had.
More details in: https://github.com/celestiaorg/celestia-node/pull/3926/files#r1833766092

vgonkivs:
conflicts cc @rach-id

rach-id:
@vgonkivs Fixed 🙏  Would be good to know if this PR is good to go so that we start merging the changes in core/comsos-sdk

nodebuilder/core/config.go
```diff
@@ -17,7 +16,6 @@ var MetricsEnabled bool
 // Config combines all configuration fields for managing the relationship with a Core node.
 type Config struct {
 	IP       string
-	RPCPort  string
 	GRPCPort string
```
vgonkivs, line 19: what about calling it "Port" since we assume that it will anyway be a grpc?

rach-id: sure

rach-id: fe20093b27c1dbf8ee1c3678b8e43aaf301fa238

rach-id: does it also make sense to change the flag 'core.grpc.port' to 'core.port'?

vgonkivs: let's keep it grpc for now. I'll update flags in https://github.com/celestiaorg/celestia-node/pull/3922

vgonkivs:
Tested locally. Both grpc clients works

vgonkivs (dismissed):

state/core_access.go
```diff
@@ -123,7 +123,7 @@ func (ca *CoreAccessor) Start(ctx context.Context) error {
 	ca.ctx, ca.cancel = context.WithCancel(context.Background())
 
 	// dial given celestia-core endpoint
-	endpoint := fmt.Sprintf("%s:%s", ca.coreIP, ca.grpcPort)
+	endpoint := fmt.Sprintf("%s:%s", ca.coreIP, ca.port)
```
cristaloleg, line 126: Nit: https://pkg.go.dev/net#JoinHostPort

core/fetcher.go
```diff
@@ -54,41 +73,86 @@ func (f *BlockFetcher) GetBlockInfo(ctx context.Context, height *int64) (*types.
...
+	return block, nil
+}
+
+// resolveHeight takes a height pointer and returns its value if it's not nil.
+// otherwise, returns the latest height.
+func (f *BlockFetcher) resolveHeight(ctx context.Context, height *int64) (int64, error) {
+	if height != nil {
+		return *height, nil
+	}
+	status, err := f.client.Status(ctx, &coregrpc.StatusRequest{})
```
Wondertan, line 116: Isn't there a way to request the latest height through existing methods? Let's avoid additional roundtrip for Head request if possible.

Wondertan: Like `0` on `BlockByHeightRequest` can represent latest height.

core/testing.go
```diff
@@ -50,37 +46,12 @@ func StartTestNode(t *testing.T) testnode.Context {
...
-	require.NoError(t, err)
-	t.Cleanup(func() {
-		err := client.Stop()
-		require.NoError(t, err)
-	})
-
-	cctx.WithClient(client)
+	// so we are as close to the real environment as possible,
+	// however, it might be useful to use a local tendermint client
+	// if you need to debug something inside it
```
Wondertan, line 51: Can we still keep the remote client?

core/client.go
```diff
@@ -3,24 +3,13 @@ package core
...
-	httpClient.RetryMax = 2
-	// suppress logging
-	httpClient.Logger = nil
-
-	return http.NewWithClient(
-		fmt.Sprintf("tcp://%s:%s", ip, port),
-		"/websocket",
-		httpClient.StandardClient(),
-	)
+	return coregrpc.StartBlockAPIGRPCClient(fmt.Sprintf("tcp://%s:%s", ip, port))
```
Wondertan, line 14: Can we reuse the client across `state` and `core` modules?

vgonkivs: Yes. I'm planning to do it in a subsequent PR

core/fetcher.go
```diff
@@ -145,45 +211,133 @@ func (f *BlockFetcher) SubscribeNewBlockEvent(ctx context.Context) (<-chan types
...
 	return resp.SyncInfo.CatchingUp, nil
 }
+
+func receiveBlockByHeight(streamer coregrpc.BlockAPI_BlockByHeightClient) (
+	*types.Block,
+	*types.BlockMeta,
+	*types.Commit,
+	*types.ValidatorSet,
+	error,
+) {
```
walldiss, line 260: Return of 5 values caught my eye. I can't find if BlockMeta returned from this method is used anywhere. If not, perhaps we can return just SignedBlock struct, which contains everything except Blockmeta?

core/fetcher.go
```diff
@@ -145,45 +211,133 @@ func (f *BlockFetcher) SubscribeNewBlockEvent(ctx context.Context) (<-chan types
...
+				if ctx.Err() == nil {
+					log.Errorw("fetcher: error receiving new height", "err", err.Error())
+				}
+				return
+			}
+			signedBlock, err := f.GetSignedBlock(ctx, &resp.Height)
+			if err != nil {
+				log.Errorw("fetcher: error receiving signed block", "height", resp.Height, "err", err.Error())
+				return
+			}
```
walldiss, line 226: Lets add per-call timeout. Otherwise we might get stuck on one of the calls indefinitely

Also would need to handle canceled context here

core/fetcher.go
renaynay: what about closing the underlying connection? What's responsible for that?

nodebuilder/core/module.go
```diff
@@ -76,12 +76,6 @@ func ConstructModule(tp node.Type, cfg *Config, options ...fx.Option) fx.Option
 			)),
 			fx.Provide(fx.Annotate(
 				remote,
-				fx.OnStart(func(_ context.Context, client core.Client) error {
-					return client.Start()
-				}),
-				fx.OnStop(func(_ context.Context, client core.Client) error {
-					return client.Stop()
-				}),
```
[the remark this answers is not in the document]
renaynay: Yes but we also need to close conn when node stops.

renaynay (commented):
Many thanks for your efforts here @rach-id !

This PR made me think about #3952 which would allow us to only rely on `GetSignedBlock` endpoint in fetcher and remove all the other unnecessary fluff (like `GetBlockInfo` etc.

core/fetcher.go
renaynay: That's why I recommended to just use 0

core/fetcher.go
rach-id: good catch. I can support closing the connection if we make the `Client` a struct. would that be something we want?

nodebuilder/core/module.go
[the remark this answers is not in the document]
rach-id: https://github.com/celestiaorg/celestia-node/pull/3915#discussion_r1850903408

state/core_access.go
rach-id: f371c33329d21daf3ee0ebc84d363fcfd778ea82

rach-id: f371c33329d21daf3ee0ebc84d363fcfd778ea82

core/fetcher.go
```diff
@@ -54,41 +67,86 @@ func (f *BlockFetcher) GetBlockInfo(ctx context.Context, height *int64) (*types.
...
+		return nil, err
 	}
 
-	return res.Block, nil
+	return block, nil
+}
+
+// resolveHeight takes a height pointer and returns its value if it's not nil.
+// otherwise, returns the latest height.
+func (f *BlockFetcher) resolveHeight(ctx context.Context, height *int64) (int64, error) {
```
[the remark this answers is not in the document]
rach-id: Yes, I tried to keep the same API. I don't like passing pointers to numbers too. Do you think this should be updated in this PR? or we create an issue and leave it later on for someone to pick it up at some point?

core/exchange.go
```diff
@@ -173,12 +173,12 @@ func (ce *Exchange) getExtendedHeaderByHeight(ctx context.Context, height *int64
 	}
 	log.Debugw("fetched signed block from core", "height", b.Header.Height)
 
-	eds, err := extendBlock(b.Data, b.Header.Version.App)
+	eds, err := extendBlock(*b.Data, b.Header.Version.App)
```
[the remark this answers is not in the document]
rach-id: this will be GC at some point. The thing is `extendBlock` also makes a couple other calls by value instead of reference. So multiple copies will be done regardless, this will be just one less. What do you think of creating an issue that checks wherever `Data` is used, it passes it by reference, even in app and maybe other repos so that we have everything fixed?

core/fetcher.go
```diff
@@ -97,45 +161,47 @@ func (f *BlockFetcher) Commit(ctx context.Context, height *int64) (*types.Commit
...
 // a new block event channel on success.
 func (f *BlockFetcher) SubscribeNewBlockEvent(ctx context.Context) (<-chan types.EventDataSignedBlock, error) {
-	// start the client if not started yet
-	if !f.client.IsRunning() {
-		return nil, errors.New("client not running")
-	}
-
 	ctx, cancel := context.WithCancel(ctx)
 	f.cancel = cancel
 	f.doneCh = make(chan struct{})
```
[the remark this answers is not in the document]
rach-id: I kept the same existing pattern. What do you mean by fool-proving?

core/fetcher.go
rach-id: Im fine with changing and removing the pointers. Is this something you want to do in this PR?

core/testing.go
rach-id: what do you mean keep it? the tests still use the client except that it's automatically created and no need to manually start it.

core/fetcher.go
rach-id: I am returning everything we receive in this. If you don't want block meta to be returned, I guess we could even stop sending it from core. What do you think?

core/fetcher.go
rach-id: 92c4f942f62503ef41cb40af50a835f463cea216

rach-id: 92c4f942f62503ef41cb40af50a835f463cea216

core/fetcher.go
```diff
@@ -54,41 +73,86 @@ func (f *BlockFetcher) GetBlockInfo(ctx context.Context, height *int64) (*types.
...
+		return nil, err
 	}
 
-	return res.Block, nil
+	return block, nil
+}
+
+// resolveHeight takes a height pointer and returns its value if it's not nil.
+// otherwise, returns the latest height.
+func (f *BlockFetcher) resolveHeight(ctx context.Context, height *int64) (int64, error) {
```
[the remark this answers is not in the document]
rach-id: https://github.com/celestiaorg/celestia-node/pull/3915#discussion_r1850910229

core/exchange.go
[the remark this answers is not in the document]
rach-id: because extendBlock expects the parameter to be passed by value. I didn't want to change that behavior since Data is already being copied in a lot of places, one more copy is fine. Related: https://github.com/celestiaorg/celestia-node/pull/3915#discussion_r1850917520

core/exchange_test.go
```diff
@@ -136,7 +142,9 @@ func fillBlocks(
 		}
 
 		_, err := cctx.FillBlock(16, cfg.Genesis.Accounts()[0].Name, flags.BroadcastBlock)
-		require.NoError(t, err)
+		if err != nil && ctx.Err() == nil {
+			require.NoError(t, err)
```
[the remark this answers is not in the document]
rach-id: there is a funny behavior happening there that I guess will be fixed if we support closing the connection manually. Related:  https://github.com/celestiaorg/celestia-node/pull/3915#discussion_r1850903408

core/fetcher.go
Wondertan: Yes pls, we would like to leave as minimal tech debt behind as possiblr

core/fetcher.go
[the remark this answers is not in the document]
Wondertan: I meant making erroring or panicking if subsvription is called again

core/fetcher.go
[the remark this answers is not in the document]
rach-id: 7a6257044bf23f4c88e6a243ba8bb31a5da6def4

core/fetcher.go
rach-id: 7a6257044bf23f4c88e6a243ba8bb31a5da6def4

core/fetcher.go
rach-id: a033feb98bbe57abad444a34d63e8f72369fd0cd

nodebuilder/core/module.go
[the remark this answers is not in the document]
rach-id: a033feb98bbe57abad444a34d63e8f72369fd0cd

core/fetcher.go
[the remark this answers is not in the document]
rach-id: 7a6257044bf23f4c88e6a243ba8bb31a5da6def4

core/exchange_test.go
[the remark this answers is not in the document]
cristaloleg: My 2c: feel like it can be `if ctx.Err() == nil`, in all other cases we want nil error.

core/fetcher.go
```diff
@@ -97,45 +159,55 @@ func (f *BlockFetcher) Commit(ctx context.Context, height *int64) (*types.Commit
...
 // SubscribeNewBlockEvent subscribes to new block events from Core, returning
 // a new block event channel on success.
 func (f *BlockFetcher) SubscribeNewBlockEvent(ctx context.Context) (<-chan types.EventDataSignedBlock, error) {
-	// start the client if not started yet
+	// return error if the client is still not started
 	if !f.client.IsRunning() {
 		return nil, errors.New("client not running")
 	}
-
+	if f.isListeningForBlocks {
```
cristaloleg, line 202: Feels like we have a race condition here. WDYT to change `isListeningForBlocks` to `atomic.Bool` ?

rach-id: 504591866243ad03a246957f8d61182192c1e5c2

core/fetcher.go
```diff
@@ -97,45 +159,55 @@ func (f *BlockFetcher) Commit(ctx context.Context, height *int64) (*types.Commit
...
-	for page := 1; len(vals) != total; page++ {
-		res, err := f.client.Validators(ctx, height, &page, &perPage)
-		if err != nil {
-			return nil, err
-		}
+// If the height is nil, use the latest height.
+func (f *BlockFetcher) ValidatorSet(ctx context.Context, height int64) (*types.ValidatorSet, error) {
+	// return error if the client is still not started
+	if !f.client.IsRunning() {
+		return nil, errors.New("client not running")
```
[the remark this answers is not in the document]
rach-id: 5634a4d3d6cc9cea623f1cca16554d92899e455c

core/exchange_test.go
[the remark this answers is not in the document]
rach-id: c5a9abdb815eb391258be86a6faecfbb872461b3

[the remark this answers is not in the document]
rach-id: removed the check as it's fixed when we started closing the connection explicitly

cristaloleg (commented):
LGTM from me

core/exchange.go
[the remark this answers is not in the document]
renaynay: That's also fine, but `extendBlock` used to accept data by ref, so why make this change here?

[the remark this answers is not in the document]
renaynay: Why not both? (pass by ref here and also open an issue to change other calls to accept data by ref)

core/fetcher.go
renaynay: So just to confirm, you've changed this on the tendermint gRPC server-side such that requesting height 0 is the same as requesting height == nil over the RPC client ( return latest block )?

rach-id: https://github.com/celestiaorg/celestia-core/pull/1513/commits/6291264239f8c3cb3f5add5da7a9e6742422ab85

core/fetcher.go
```diff
@@ -187,30 +218,41 @@ func (f *BlockFetcher) SubscribeNewBlockEvent(ctx context.Context) (<-chan types
...
-				ValidatorSet: *signedBlock.ValidatorSet,
-				Data:         *signedBlock.Data,
-			}:
 			case <-ctx.Done():
 				return
+			default:
+				resp, err := subscription.Recv()
+				if err != nil {
+					// case where the context was not canceled but still received an error
+					if ctx.Err() == nil {
```
renaynay, line 231: What's the point of the `ctx.Err` check right here?

api/docgen/exampledata/extendedHeader.json
```diff
@@ -46,7 +46,7 @@
     }
   },
   "commit": {
-    "height": 67374,
+    "height": "67374",
```
renaynay, line 49: why is this change necessary here btw?

core/fetcher.go
rach-id: removing block meta and using signed block as a return value: 45cb70f730f7ac34a35330cb540b8d6b5ae58aac

core/fetcher.go
```diff
@@ -56,6 +59,10 @@ func (f *BlockFetcher) Stop(ctx context.Context) error {
 
 // GetBlockInfo queries Core for additional block information, like Commit and ValidatorSet.
 func (f *BlockFetcher) GetBlockInfo(ctx context.Context, height int64) (*types.Commit, *types.ValidatorSet, error) {
+	// return error if the client is still not started
+	if !f.client.IsRunning() {
+		return nil, nil, errors.New("client not running")
```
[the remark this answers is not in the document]
rach-id: cd8092a5cd822a027fe1389fd6c12dfa64969e6f

core/fetcher.go
rach-id: `subscription.Recv()` is blocking, and I didn't want to overcomplicate things by adding a channel. So, if some error happens when the subscription is waiting to receive something, we check if it's just the context cancelled to avoid printing an error in that case.

api/docgen/exampledata/extendedHeader.json
rach-id: it just happened

rach-id: it's from some other commit that I don't know why it happened :D I'll revert

core/fetcher.go
renaynay: @rach-id this solution doesn't really resolve the issue you're trying to address, though. Can you make subscription.Recv() take a context and respect its cancellation in order to return? 

otherwise, we'd need to wrap that blocking call in a way that respects ctx cancellation.

rach-id: as per sync discussion, removing this part as it's unnecessary and confusing: 600b540576b5672a522d7aa502844cf9ad9780a6

vgonkivs (approved):

PR #3917, opened 2024-11-05
initial POC for grpc tls

jhawk28:
Initial concept for gRPC tls support

### DIFF
--- a/nodebuilder/core/config.go
+++ b/nodebuilder/core/config.go
@@ -16,9 +16,10 @@ var MetricsEnabled bool
 
 // Config combines all configuration fields for managing the relationship with a Core node.
 type Config struct {
-	IP       string
-	RPCPort  string
-	GRPCPort string
+	IP        string
+	RPCPort   string
+	GRPCPort  string
+	EnableTLS bool
 }
 
 // DefaultConfig returns default configuration for managing the
--- a/state/core_access.go
+++ b/state/core_access.go
@@ -2,11 +2,18 @@ package state
 
 import (
 	"context"
+	"crypto/tls"
 	"errors"
 	"fmt"
 	"sync"
 	"time"
 
+	"github.com/celestiaorg/celestia-app/v3/app"
+	"github.com/celestiaorg/celestia-app/v3/app/encoding"
+	apperrors "github.com/celestiaorg/celestia-app/v3/app/errors"
+	"github.com/celestiaorg/celestia-app/v3/pkg/user"
+	libhead "github.com/celestiaorg/go-header"
+	libshare "github.com/celestiaorg/go-square/v2/share"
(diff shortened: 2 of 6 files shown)


a26nine:
I was able to sync a light node with:
```
[Core]
  IP = "xxx.celestia-mainnet.quiknode.pro"
  RPCPort = "443"
  GRPCPort = "9090"
  EnableTLS = true
```
using this build.

state/core_access.go
```diff
@@ -124,9 +119,15 @@ func (ca *CoreAccessor) Start(ctx context.Context) error {
 
 	// dial given celestia-core endpoint
 	endpoint := fmt.Sprintf("%s:%s", ca.coreIP, ca.grpcPort)
+	grpcOpts := make([]grpc.DialOption, 0)
+	if ca.enableTLS {
+		grpcOpts = append(grpcOpts, grpc.WithTransportCredentials(credentials.NewTLS(&tls.Config{})))
```
vgonkivs, line 124: Without specifying valid certificates in the tls.Config object will not secure gRPC connection. By default, an empty tls.Config{} means that no certificate or private key is provided, which leads to an insecure setup.

Node already has TLS certificates handling logic [here](https://github.com/celestiaorg/celestia-node/blob/ace6840bb00b337c320c07be4dbe10e4d853271d/nodebuilder/p2p/tls.go#L24) for web socket, so you can reuse the certificate for the grpc connection.

nodebuilder/core/config.go
```diff
@@ -16,9 +16,10 @@ var MetricsEnabled bool
 
 // Config combines all configuration fields for managing the relationship with a Core node.
 type Config struct {
-	IP       string
-	RPCPort  string
-	GRPCPort string
+	IP        string
+	RPCPort   string
+	GRPCPort  string
+	EnableTLS bool
```
vgonkivs, line 22: let's avoid growing the node's config. We have an env variable that signals that TLS should be enabled.

vgonkivs (changes requested):

jcstein:
is it possible to set it up the same way that `rpc-mocha.pops.one` (mocha) works, for example? it uses the light node's default config and works by default (`celestia light start --p2p.network mocha --core.ip rpc-mocha.pops.one`)

state/core_access.go
jhawk28: This was a proof of concept to get the Light node working against a Quicknode celestia endpoint. QN doesn't support client certs. gRPC auth for a QN endpoint occurs over the `x-token` header

vgonkivs: One more point here is the solution is not generic. You may not support the certificates but if someone does, the connection won't be established. Also, it is recommended to manually specify the MinTLSVersion(linter confirmed it)

vgonkivs:
closed in favor of #3954

PR #3925, opened 2024-11-07
fix(header)!: use tmjson encoding for Commit field as well

### LINKED ISSUE #3918: Serialization inconsistency between celestia-node API and Tendermint API
### Celestia Node version

ace6840bb00b337c320c07be4dbe10e4d853271d

### OS

Arch Linux

### Install tools

_No response_

### Others

_No response_

### Steps to reproduce it

Send `header.GetByHeight` request and `/commit` request.

### Expected result

Both to be the same, preferably a string.

### Actual result

In the `header.GetByHeight` response the `Commit.Height` is a number and in `/commit` response it is a string.

### Relevant log output

_No response_

_No response_

### Notes

celestia-node API and Tendermint API have some types in common. One example is `Commit` type.
However we noticed that their JSON serialization is not the same. For example `Commit.Height` when it used from celestia-node API is serialized as a number and as a string when it is used from Tendermint API.

We managed to find the root cause: Tendermint API uses a custom serializer which [explicitly serializes `int64`/`uint64` as strings](https://github.com/celestiaorg/celestia-core/blob/746a1e15d6d730bc0db09bd510236ab269472da4/libs/json/encoder.go#L104-L111), but celestia-node API does it [only for `ExtendedHeader.RawHeader` and `ExtendedHeader.ValidatorSet`](https://github.com/celestiaorg/celestia-node/blob/ace6840bb00b337c320c07be4dbe10e4d853271d/header/header.go#L234-L241).

This inconsistency is a huge problem in Lumina because we can not control which serializer must be used in the [jsonrpsee](https://github.com/paritytech/jsonrpsee) level.

We believe celestia-node API could also use the same serializer as Tendermint API.

renaynay:
Fixes #3918 

**Please note: this PR is _BREAKING_** as header encoding has changed.

I don't see a reason we weren't doing this to begin with as we tmjson encode the other "tendermint" fields.

### DIFF
--- a/api/docgen/exampledata/extendedHeader.json
+++ b/api/docgen/exampledata/extendedHeader.json
@@ -46,7 +46,7 @@
     }
   },
   "commit": {
-    "height": 67374,
+    "height": "67374",
     "round": 0,
     "block_id": {
       "hash": "A7F6B1CF33313121539206754A73FDC22ADA48C4AA8C4BB4F707ED2E089E59D3",
--- a/header/header.go
+++ b/header/header.go
@@ -239,13 +239,19 @@ func (eh *ExtendedHeader) MarshalJSON() ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
+	commit, err := tmjson.Marshal(eh.Commit)
+	if err != nil {
+		return nil, err
+	}
 	return json.Marshal(&struct {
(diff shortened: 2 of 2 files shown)


Wondertan (changes requested):
We should tmjson the whole ExtendedHeader altogether, without aliasing and individual calls for each field

Issue #3916, opened 2024-11-05
[Feature Request]: Update Grant Amount in Celestia's FeeGrant Module


a26nine:
## Feature Request: Update Grant Amount in Celestia's FeeGrant Module

**Description:**

The [feeGrant module](https://docs.celestia.org/how-to-guides/feegrant-for-blobs) in celestia-node enables a Granter to cover the transaction fees of a Grantee's blob submission. This is facilitated by granting a fee allowance, which authorizes the Grantee to submit blobs on behalf of the Granter up to a specified threshold.

**Problem:**

Currently, there's no way to modify the "grant amount" once it's set. Attempting to update it using the same process results in the following error:

```
failed to execute message; message index: 0: fee allowance already exists: invalid request
```

This behavior is evident in [this transaction](https://celenium.io/tx/36bd5ab0246db250b6e6c83ac2f5ab2242687d1bc98e2e414563f41f433da280?tab=messages), which fails when attempting to increase the allowance for an existing grant.

**Proposed Solution:**

Implement a feature that allows updating an existing grant amount. This would improve flexibility by eliminating the need to revoke the previous grant and create a new one each time a threshold adjustment is necessary.

**Benefits:**

- Simplifies workflows for managing allowances
- Reduces redundant transactions related to revoking and re-granting allowances
- Enhances usability for users leveraging the feeGrant module

jcstein:
I agree that this would be useful. I think this also may need a corresponding issue to be made in https://github.com/celestia-app

PR #3926, opened 2024-11-08
hotfix(core): don't hold on Data

Wondertan:
The `Data` field of `core.ResultSignedBlock` was retained even after we constructed the respective header and moved on. This was due to the header consisting of pointers pointing to fields of `core.ResultSignedBlock`, retaining the whole structure, including the `Data` field. A BN, by default, has a header store cache of size 4096; thus, holding on `Data` with 8 MB blocks took around 32GiB of RAM.

Initially, we believed that the issue was with JSON unmarshalling. However, we were wrong about this being the whole story, as while profiles did confirm that the allocations did originate there, the allocated data wasn't cleaned up.  Kudos to @rach-id for helping us figuring this out!

This is a hotfix and is meant to be replaced with a better solution described on TODO. Also, the origin of this bug is yet to be confirmed by @rach-id by testing RAM usage with disabled header cache. 

---

🎱mb certified

### DIFF
--- a/core/exchange.go
+++ b/core/exchange.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/tendermint/tendermint/types"
 	"golang.org/x/sync/errgroup"
 
 	libhead "github.com/celestiaorg/go-header"
@@ -177,7 +178,11 @@ func (ce *Exchange) getExtendedHeaderByHeight(ctx context.Context, height *int64
 	if err != nil {
 		return nil, fmt.Errorf("extending block data for height %d: %w", b.Header.Height, err)
 	}
-	// create extended header
+
+	// TODO(@Wondertan): This is a hack to deref Data, allowing GC to pick it up.
+	//  The better footgun-less solution is to change core.ResultSignedBlock fields to be pointers instead of values.
+	b.Data = types.Data{}
+
 	eh, err := ce.construct(&b.Header, &b.Commit, &b.ValidatorSet, eds)
 	if err != nil {
 		panic(fmt.Errorf("constructing extended header for height %d: %w", b.Header.Height, err))


core/exchange.go
```diff
@@ -177,7 +178,11 @@ func (ce *Exchange) getExtendedHeaderByHeight(ctx context.Context, height *int64
 	if err != nil {
 		return nil, fmt.Errorf("extending block data for height %d: %w", b.Header.Height, err)
 	}
-	// create extended header
+
+	// TODO(@Wondertan): This is a hack to deref Data, allowing GC to pick it up.
+	//  The better footgun-less solution is to change core.ResultSignedBlock fields to be pointers instead of values.
+	b.Data = types.Data{}
+
 	eh, err := ce.construct(&b.Header, &b.Commit, &b.ValidatorSet, eds)
```
rach-id, line 186:
```suggestion

	// copying the commit and validator set so we break the link with
	// the signed header and the signed block is garbage collected
	commit := &types.Commit{
		Height:     b.Commit.Height,
		Round:      b.Commit.Round,
		BlockID:    b.Commit.BlockID,
		Signatures: b.Commit.Signatures,
	}
	vs := &types.ValidatorSet{
		Validators: b.ValidatorSet.Validators,
		Proposer:   b.ValidatorSet.Proposer,
	}

	// create extended header
	eh, err := ce.construct(&b.Header, commit, vs, eds)
```

This fixes the bug because the commit and validator set are used as they are in:
https://github.com/celestiaorg/celestia-node/blob/353141fb5a0a33c10a8678ac6e92a90b02f74dbd/header/header.go#L64-L70

Which results in GC not liberating that memory. 

After copying their values as per the suggestion, the signed block is garbage collected, memory freed, bridge node syncing in under 5GB RAM compared to 50GB RAM.

An alternative solution would be to change: 
https://github.com/celestiaorg/celestia-node/blob/353141fb5a0a33c10a8678ac6e92a90b02f74dbd/header/header.go#L44-L49

To:

```
func MakeExtendedHeader(
	h *core.Header,
	comm core.Commit,
	vals core.ValidatorSet,
	eds *rsmt2d.ExtendedDataSquare,
) (*ExtendedHeader, error) {
```

So the copy is made automatically but I didn't try it as I prefer the clear first approach.

Attached image showing a bridge node syncing a 6s full 8mb blocks network :D 

image

rach-id (commented):
I guess this PR can be closed in favor of https://github.com/celestiaorg/celestia-node/pull/3915/files#r1834055837

Wondertan:
The idea is to make hotfix for this in nearest patch. Releasing grpc in core and here is gonna take a bunch of time

core/exchange.go
Wondertan: In Slack we agreed to land the proper fix in #3916, while setting Data to be empty is sufficient as a hot fix.

rach-id:
> The idea is to make hotfix for this in a nearest patch, as releasing grpc in core and here is gonna take a bunch of time.

Makes sense 👍 Then I guess applying the change in https://github.com/celestiaorg/celestia-node/pull/3926#discussion_r1833766092 is better to avoid messing with memory ourselves. Instead, make the copy and let the GC do its job without us forcing the freeing of memory. Also, GC might collect additional objects that might be linked to SignedBlock instead of collecting just the Data.

renaynay (approved):
as per https://github.com/celestiaorg/celestia-node/pull/3926#discussion_r1834116277, I am okay with this hotfix until @rach-id 's solution lands.

renaynay (commented):
I am also okay with the explicit approach in https://github.com/celestiaorg/celestia-node/pull/3926#issuecomment-2464939713

Wondertan:
@rach-id, I applied your suggestion to get you as a coauthor on the PR

vgonkivs (dismissed):

Wondertan:
Well, the suggestion has a bug. It does not copy the `RawHeader`, so the problem preserves. 

Copying every field of `RawHeader` is gonna be bulky, so I will reverse to the old fix

rach-id:
You dont need to copy the raw header. Did you try running it and see the memory usage?

Wondertan:
You need to copy it as well, because it is also a value that we create pointer to, keeping the `ResultSignedBlock` around

rach-id:
If I remember correctly, I tried that change and ran a bridge node, and it didn't have the issue.

Could you double check please in case I missed something when running?

Wondertan:
Well, this is how memory works. We don't need to test that.

Wondertan:
But I also simply don't have time to run the test during the devcon, and we should go with the solution we know works well, which is setting empty data.

walldiss (approved):

cristaloleg (approved):

renaynay (approved):

PR #3928, opened 2024-11-11
fix(header)!: tendermint compatible json marshall

zvolin:
Fixes https://github.com/celestiaorg/celestia-node/issues/3918

**Please note: this PR is BREAKING** as header encoding has changed.

### DIFF
--- a/header/header.go
+++ b/header/header.go
@@ -2,7 +2,6 @@ package header
 
 import (
 	"bytes"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"time"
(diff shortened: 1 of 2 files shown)


header/header.go
```diff
@@ -227,56 +226,19 @@ func (eh *ExtendedHeader) UnmarshalBinary(data []byte) error {
 	return nil
 }
 
-// MarshalJSON marshals an ExtendedHeader to JSON. The ValidatorSet is wrapped with amino encoding,
-// to be able to unmarshal the crypto.PubKey type back from JSON.
+// MarshalJSON marshals an ExtendedHeader to JSON.
+// Uses tendermint encoder for tendermint compatiblity.
```
cristaloleg, line 230: Lint failure on `compatibility`

header/header.go
```diff
@@ -227,56 +226,19 @@ func (eh *ExtendedHeader) UnmarshalBinary(data []byte) error {
...
-	return json.Marshal(&struct {
-		RawHeader    json.RawMessage `json:"header"`
-		ValidatorSet json.RawMessage `json:"validator_set"`
-		*Alias
-	}{
-		ValidatorSet: validatorSet,
-		RawHeader:    rawHeader,
-		Alias:        (*Alias)(eh),
-	})
+	return tmjson.Marshal(*eh)
```
Wondertan, line 232: It shouldn't be necessary to pass the value unless there is some quirk in the `tmjson`. Have you tried without `*`?

zvolin: it looks like it was the same case with recursive call as in unmarshal. Just dereferencing hidden that. Switched to alias to be clear what issue is solved there

Wondertan: I see

Wondertan (approved):

renaynay (approved):

PR #3929, opened 2024-11-11
fix(blob)!: tendermint compatible commitment proof json marshall

zvolin:
Fixes https://github.com/celestiaorg/celestia-node/issues/3918

**Please note: this PR is BREAKING** as commitment proof encoding has changed.


rach-id (commented):
Thanks for opening this PR.

Do you think we will still need the JSON marshall/unmarshall after merging this https://github.com/celestiaorg/celestia-node/pull/3821 to main?

zvolin:
yes, it uses `coretypes.RowProof`. Everything that comes from core (tendermint fork) should be serialized with tendermint json encoder / decoder as it treats 64 bit integers differently (stringifies them)

rach-id:
thanks for your response @zvolin. And why is this change useful?

rach-id:
After second thinking about this, I guess you're right, this needs fixing

zvolin:
> And why is this change useful?

to have unified representation of objects regardless whether they are returned from celestia-node api or tendermint rest api. Moreover, tendermint implementations in different languages can have the rules how to encode types in different level than in go, namely [tendermint-rs bakes how the type is meant to be encoded on type level](https://github.com/informalsystems/tendermint-rs/blob/main/proto/src/prost/v0_34/tendermint.crypto.rs#L7), not encoder level, so in lumina we had to patch tendermint to achieve celestia-node compatible representation, but we could never be compatible with both tendermint and celestia-node.

tl;dr it originates from us wishing to drop tendermint-rs fork but I believe it's a good thing in general

vgonkivs:
@zvolin could you please fix lint?

zvolin:
switched to marshaling whole type as per https://github.com/celestiaorg/celestia-node/pull/3930#discussion_r1837870368

renaynay (approved):

vgonkivs:
Linter failed.

zvolin:
should be good now, sorry

renaynay (commented):
@zvolin some broken tests

vgonkivs:
` blob_fuzz_test.go:69: invalid 64-bit integer encoding "64", expected string`

vgonkivs (approved):
LGTM 🚀 . Thanks @zvolin

zvolin:
`make test-integration SHORT=true TAGS=sync` passes for me locally :thinking:

PR #3944, opened 2024-11-15
fix(share/availability/light): Allow retries on ODS size 1

### LINKED ISSUE #3943: share/availability/light: 16 samples attempted on a non-empty, single-share ODS
```
2024-11-15T15:58:13.262Z	ERROR	das	das/worker.go:134	failed to sample header	{"type": "retry", "height": 3089050, "hash": "F2D6B3C115227531BB89B525073451475ABA4D997445043622A7A336196218F8", "square width": 2, "data root": "8F86D0659F5D39A08F3064E0352F64C5EA1558AA8C607A428A8722108268F090", "err": "invalid sampling result: expected 16 samples, got 4", "finished (s)": 0.00136031}
2024-11-15T15:58:13.262Z	ERROR	das	das/worker.go:134	failed to sample header	{"type": "retry", "height": 3089855, "hash": "C6805A02F90C480BA5422D419BDFCA044427F6CE6A2BAF34EB0B6DA18190D035", "square width": 2, "data root": "DAD2887B6E3BC2FB4E27ED6ADBD350744CB1121AA3C9E2F60D87BE57522517E7", "err": "invalid sampling result: expected 16 samples, got 4", "finished (s)": 0.00147832}
```

with corresponding non-empty block: 
https://mocha.celenium.io/block/3089855?tab=transactions

This is because we use `Sample` as the key in the map for the generated random coordinates such that we don't duplicate coordinates - so we end up with a map of 4 keys and that throws the error we're seeing above once sampling is actually successfully completed.

renaynay:
Fixes #3943 

Another PR will come later which optimises the request for the single-share-square case :)

### DIFF
--- a/share/availability/light/availability.go
+++ b/share/availability/light/availability.go
@@ -103,7 +103,7 @@ func (la *ShareAvailability) SharesAvailable(ctx context.Context, header *header
 		}
 		// Verify total samples count.
 		totalSamples := len(samples.Remaining) + len(samples.Available)
-		if totalSamples != int(la.params.SampleAmount) {
+		if (totalSamples != int(la.params.SampleAmount)) && (totalSamples != len(dah.RowRoots)*len(dah.RowRoots)) {
 			return fmt.Errorf("invalid sampling result:"+
 				" expected %d samples, got %d", la.params.SampleAmount, totalSamples)
 		}
--- a/share/availability/light/availability_test.go
+++ b/share/availability/light/availability_test.go
@@ -135,9 +135,15 @@ func TestSharesAvailableFailed(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	getter := mock.NewMockGetter(gomock.NewController(t))
+	failGetter := mock.NewMockGetter(gomock.NewController(t))
+	// Getter doesn't have the eds, so it should fail for all samples
+	failGetter.EXPECT().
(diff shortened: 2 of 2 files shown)


share/availability/light/availability.go
```diff
@@ -103,7 +103,7 @@ func (la *ShareAvailability) SharesAvailable(ctx context.Context, header *header
 		}
 		// Verify total samples count.
 		totalSamples := len(samples.Remaining) + len(samples.Available)
-		if totalSamples != int(la.params.SampleAmount) {
+		if (totalSamples != int(la.params.SampleAmount)) && (totalSamples != len(dah.RowRoots)*len(dah.RowRoots)) {
```
walldiss, line 106: Imagine if sample la.params.SampleAmount is not 16. We should detect only when params value is higher when eds size. 
```suggestion
		if (totalSamples != int(la.params.SampleAmount)) && (la.params.SampleAmount > len(dah.RowRoots)*len(dah.RowRoots)) {
```

renaynay: I was thinking about this, but I also didn't think it was a problem if we sample *more* -- only if we sample less.

walldiss (approved):

vgonkivs:
UT failed
`Error: share/availability/light/availability.go:106:81: invalid operation: la.params.SampleAmount > len(dah.RowRoots) * len(dah.RowRoots) (mismatched types uint and int)`

walldiss (approved):

Wondertan (approved):

PR #3946, opened 2024-11-18
fix(shwap/bitswap): update boxo and config

Wondertan:
Yet another change to the bitswap configuration
* Adds `MinTimeout`. 
  * Robusta was having hard times with tail latency being too long, skewing timeout estimations, and causing cascade of simulated DONT_Haves that aren't real.
  * Basically only needed on low-latency envs like robusta.
* Reverts wantlist limit for peers to 4096
  * We observed limit overflow with FN syncing over bitswap. The overflow wasn't reached in any subsequent runs after the limit was raised back

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -362,4 +362,4 @@ replace (
 	github.com/tendermint/tendermint => github.com/celestiaorg/celestia-core v1.43.0-tm-v0.34.35
 )
 
-replace github.com/ipfs/boxo => github.com/celestiaorg/boxo v0.0.0-20241114103253-16dd306892a7
+replace github.com/ipfs/boxo => github.com/celestiaorg/boxo v0.0.0-20241118020536-b0a7e7e3f8b6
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -362,4 +362,4 @@ replace (
 	github.com/tendermint/tendermint => github.com/celestiaorg/celestia-core v1.43.0-tm-v0.34.35
 )
 
-replace github.com/ipfs/boxo => github.com/celestiaorg/boxo v0.0.0-20241114103253-16dd306892a7
+replace github.com/ipfs/boxo => github.com/celestiaorg/boxo v0.0.0-20241118020536-b0a7e7e3f8b6
```
cristaloleg, line 365: What is inside this update?

Wondertan: Yes

Wondertan: Btw kudos to @waldiss for the idea

cristaloleg (approved):

go.mod
walldiss: MinTimeout support in bitswap

cristaloleg: @Wondertan wrong `walldiss` 🤣

PR #3947, opened 2024-11-18
perf(shwap/bitswap): add session pools inside Getter

Wondertan:
In the Robusta environment this change:
* Improved load distribution across BN/FNs
* Sped up sampling

Originally, I was testing Bitswap with this change but decided against including it in the patch, as it is solely an optimization. However, without this change on the rc tag, load distribution was rather strange. One LN sampled at 180/s, while all 4 others at 4/s. Adding pooling fixed the issue and resulted in all 5 LNs at about 100/s.

### DIFF
--- a/share/shwap/p2p/bitswap/getter.go
+++ b/share/shwap/p2p/bitswap/getter.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/ipfs/boxo/blockstore"
@@ -31,8 +30,8 @@ type Getter struct {
 	bstore    blockstore.Blockstore
 	availWndw time.Duration
 
-	availablePool sync.Pool
-	archivalPool  sync.Pool
+	availablePool *pool
+	archivalPool  *pool
 
 	cancel context.CancelFunc
 }
@@ -43,7 +42,13 @@ func NewGetter(
 	bstore blockstore.Blockstore,
 	availWndw time.Duration,
 ) *Getter {
-	return &Getter{exchange: exchange, bstore: bstore, availWndw: availWndw}
+	return &Getter{
+		exchange:      exchange,
+		bstore:        bstore,
+		availWndw:     availWndw,
+		availablePool: newPool(exchange),
+		archivalPool:  newPool(exchange),
+	}
 }
 
 // Start kicks off internal fetching sessions.
@@ -60,19 +65,8 @@ func (g *Getter) Start() {
 	ctx, cancel := context.WithCancel(context.Background())
 	g.cancel = cancel
 
-	var availableID atomic.Uint64
-	g.availablePool.New = func() interface{} {
-		log.Debugw("creating new available window session", "id", availableID.Load())
-		availableID.Add(1)
-		return g.exchange.NewSession(ctx)
-	}
-
-	var archivalID atomic.Uint64
-	g.archivalPool.New = func() interface{} {
-		log.Debugw("creating new archival session", "id", archivalID.Load())
-		archivalID.Add(1)
-		return g.exchange.NewSession(ctx)
-	}
+	g.availablePool.ctx = ctx
+	g.availablePool.ctx = ctx
 }
 
 // Stop shuts down Getter's internal fetching getSession.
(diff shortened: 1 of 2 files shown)


share/shwap/p2p/bitswap/getter.go
```diff
@@ -57,9 +58,21 @@ func NewGetter(
...
 func (g *Getter) Start() {
 	ctx, cancel := context.WithCancel(context.Background())
-	g.availableSession = g.exchange.NewSession(ctx)
-	g.archivalSession = g.exchange.NewSession(ctx)
 	g.cancel = cancel
+
+	var availableID atomic.Uint64
+	g.availablePool.New = func() interface{} {
+		log.Debugw("creating new available window session", "id", availableID.Load())
+		defer availableID.Add(1)
```
cristaloleg, line 66: Why `defer` here?

share/shwap/p2p/bitswap/getter.go
```diff
@@ -230,17 +258,37 @@ func (g *Getter) GetNamespaceData(
...
+	v := g.availablePool.Get()
+	if v == nil {
+		panic("Getter must be started")
+	}
+
+	return v.(exchange.Fetcher)
+}
+
+// poolSession puts session back into the respective session pool
+func (g *Getter) poolSession(session exchange.Fetcher, isArchival bool) {
```
cristaloleg, line 286: Nit: more common naming will be `putSession`.

share/shwap/p2p/bitswap/getter.go
```diff
@@ -230,17 +258,37 @@ func (g *Getter) GetNamespaceData(
...
+// isArchival reports whether the header is for archival data
+func (g *Getter) isArchival(hdr *header.ExtendedHeader) bool {
+	return !availability.IsWithinWindow(hdr.Time(), g.availWndw)
+}
 
-	isWithinAvailability := availability.IsWithinWindow(hdr.Time(), g.availWndw)
-	if isWithinAvailability {
-		session = g.availableSession
+// session takes a session out of the respective session pool
+func (g *Getter) session(isArchival bool) exchange.Fetcher {
```
cristaloleg, line 267: Nit: same-ish here, why not `get-/acquire-`?

share/shwap/p2p/bitswap/getter.go
Wondertan: Sure

share/shwap/p2p/bitswap/getter.go
cristaloleg: So up to you.

Wondertan: Oh, you're right. Removed defer

share/shwap/p2p/bitswap/getter.go
```diff
@@ -230,17 +258,37 @@ func (g *Getter) GetNamespaceData(
 	return nsShrs, nil
 }
 
-// session decides which fetching session to use for the given header.
-func (g *Getter) session(ctx context.Context, hdr *header.ExtendedHeader) exchange.Fetcher {
-	session := g.archivalSession
+// isArchival reports whether the header is for archival data
+func (g *Getter) isArchival(hdr *header.ExtendedHeader) bool {
+	return !availability.IsWithinWindow(hdr.Time(), g.availWndw)
+}
```
Wondertan, line 264: btw, isArchival is much more readable, shorter and easier to work with then IsWithinWindow

cristaloleg (approved):
SGTM

share/shwap/p2p/bitswap/getter.go
```diff
@@ -57,12 +58,24 @@ func NewGetter(
...
 func (g *Getter) Start() {
 	ctx, cancel := context.WithCancel(context.Background())
-	g.availableSession = g.exchange.NewSession(ctx)
-	g.archivalSession = g.exchange.NewSession(ctx)
 	g.cancel = cancel
+
+	var availableID atomic.Uint64
+	g.availablePool.New = func() interface{} {
+		log.Debugw("creating new available window session", "id", availableID.Load())
+		availableID.Add(1)
```
walldiss, line 66: Loaded value might be overwriten between add, so logging might be confusing. Would suggest rather to call  atomic.Add() first and use value returned by Add() for logging. 
```suggestion
	g.availablePool.New = func() interface{} {
	        id := availableID.Add(1)
		log.Debugw("creating new available window session", "id", id)
		
```

Wondertan:
I didn't realize that `sync.Pool` can silently delete objects, so there is a bug - once the pool deletes a session, it's not cleaned up from the bitswap itself. Also, we don't need to create many more sessions but require a constant amount equal to a number of DASer workers, so instead, we should implement pooling over the map.

@walldiss is gonna take over that.

Wondertan (commented):
Thanks @walldiss. LGTM

walldiss (dismissed):
approving on behalf of @Wondertan

walldiss (approved):

vgonkivs (approved):

PR #3949, opened 2024-11-18
feat(availability): env var to disable pruning

Wondertan:
This trick allowed me to reuse an old 8mb snapshot, even though it is already far beyond the availability window. I think it is helpful to add it to main.

### DIFF
--- a/share/availability/window.go
+++ b/share/availability/window.go
@@ -17,6 +17,9 @@ func IsWithinWindow(t time.Time, window time.Duration) bool {
 	if alwaysAvailable {
 		return true
 	}
+	if windowOverride != time.Duration(0) {
+		window = windowOverride
+	}
 	if window == time.Duration(0) {
 		return true
 	}
@@ -25,3 +28,14 @@ func IsWithinWindow(t time.Time, window time.Duration) bool {
 
 // alwaysAvailable is a flag that disables the availability window.
 var alwaysAvailable = os.Getenv("CELESTIA_PRUNING_DISABLED") == "true"
+
+var windowOverride time.Duration
+
+func init() {
+	durationRaw := os.Getenv("CELESTIA_OVERRIDE_AVAILABILITY_WINDOW")
+	duration, err := time.ParseDuration(durationRaw)
+	if err != nil {
+		panic("failed to parse CELESTIA_OVERRIDE_AVAILABILITY_WINDOW: " + err.Error())
+	}
+	windowOverride = duration
+}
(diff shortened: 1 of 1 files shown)


share/availability/window.go
```diff
@@ -11,8 +14,14 @@ const (
...
+	if alwaysAvailable {
+		return true
+	}
 	if window == time.Duration(0) {
 		return true
 	}
 	return time.Since(t) <= window
 }
+
+// alwaysAvailable is a flag that disables the availability window.
```
renaynay, line 26:
```suggestion
// alwaysAvailable is a flag that disables the availability window. This flag is intended for 
// testing purposes only.
```

renaynay (approved):

share/availability/window.go
```diff
@@ -11,8 +14,31 @@ const (
...
+var windowOverride time.Duration
+
+func init() {
+	durationRaw := os.Getenv("CELESTIA_OVERRIDE_AVAILABILITY_WINDOW")
+	duration, err := time.ParseDuration(durationRaw)
+	if err != nil {
+		panic("failed to parse CELESTIA_OVERRIDE_AVAILABILITY_WINDOW: " + err.Error())
+	}
+	windowOverride = duration
+}
```
walldiss, line 44: Seems like if CELESTIA_OVERRIDE_AVAILABILITY_WINDOW is set to `0` it can act as alwaysAvailable. Perhaps you can remove CELESTIA_PRUNING_DISABLED in favour of single env var?

walldiss: os.LookupEnv()

renaynay (approved):

share/availability/window.go
```diff
@@ -11,8 +14,33 @@ const (
...
 	if window == time.Duration(0) {
 		return true
 	}
 	return time.Since(t) <= window
 }
+
+// windowOverride is a flag that overrides the availability window. This flag is intended for
+// testing purposes only.
+var (
+	windowOverride    bool
```
cristaloleg, line 29: Nit: not needed, `windowOverrideDur != 0` will be enough.

Wondertan: Zero window override is a valid case and the one i need. It disables the whole pruning

cristaloleg: a-ha

cristaloleg (approved):

share/availability/window.go
Wondertan: So if override is 0 and the window is 10, the window will stay 10, even if i need as 0

walldiss (approved):

PR #3922, opened 2024-11-06
feat!(core): enable tls for grpc connection

vgonkivs:
This PR includes all changes from #3917. 
The PR introduces 2 additional fields in the core config and 2 cli flags that allow to configure these fields.

PR #3954, opened 2024-11-20
feat(core): added support of the TLS certificates along with x-token authorisation token for the gRPC connection

vgonkivs:
The PR introduces changes to the Core config. It extends the config with 2 additional fields that allow for configuring a secure grpc connection:
* `TLSEnabled` - specifies whether the connection is secure or not;
* `XTokenPath` - the path to the directory with JSON file containing the X-Token for gRPC authentication.

Three additional flags have been added to configure these fields:
* `core.tls` - allows to configure a secure connection. By default: `false`
* `core.xtoken.path` - allows to set an authorisation token. Should be passed along with the `core.tls` flag

### DIFF
--- a/nodebuilder/core/config.go
+++ b/nodebuilder/core/config.go
@@ -15,11 +15,16 @@ var MetricsEnabled bool
 
 // Config combines all configuration fields for managing the relationship with a Core node.
 type Config struct {
-	IP        string
-	Port      string
-	GRPCPort  string
-	EnableTLS bool
-	TLSPath   string
+	IP   string
+	Port string
+	// TLSPath specifies the directory path where the TLS certificates are stored.
+	// It should not include file names('cert.pem' and 'key.pem').
+	// If left empty, the client will be configured for an insecure (non-TLS) connection.
+	TLSPath string
+	// XTokenPath specifies the file path to the JSON file containing the X-Token for gRPC authentication.
+	// The JSON file should have a key-value pair where the key is "x-token" and the value is the authentication token.
+	// If left empty, the client will not include the X-Token in its requests.
+	XTokenPath string
 }
 
 // DefaultConfig returns default configuration for managing the
--- a/nodebuilder/core/config.go
+++ b/nodebuilder/core/config.go
@@ -19,6 +19,7 @@ type Config struct {
 	Port      string
 	GRPCPort  string
 	EnableTLS bool
+	TLSPath   string
 }
 
 // DefaultConfig returns default configuration for managing the
--- a/nodebuilder/core/config.go
+++ b/nodebuilder/core/config.go
@@ -15,8 +15,10 @@ var MetricsEnabled bool
 
 // Config combines all configuration fields for managing the relationship with a Core node.
 type Config struct {
-	IP   string
-	Port string
+	IP        string
+	Port      string
+	GRPCPort  string
+	EnableTLS bool
 }
 
 // DefaultConfig returns default configuration for managing the
--- a/nodebuilder/core/config.go
+++ b/nodebuilder/core/config.go
@@ -21,7 +21,7 @@ type Config struct {
 	// It should not include file names('cert.pem' and 'key.pem').
 	// If left empty, the client will be configured for an insecure (non-TLS) connection.
 	TLSPath string
-	// XTokenPath specifies the file path to the JSON file containing the X-Token for gRPC authentication.
+	// XTokenPath specifies the path to the directory with JSON file containing the X-Token for gRPC authentication.
 	// The JSON file should have a key-value pair where the key is "x-token" and the value is the authentication token.
 	// If left empty, the client will not include the X-Token in its requests.
 	XTokenPath string
--- a/nodebuilder/core/config.go
+++ b/nodebuilder/core/config.go
@@ -17,6 +17,9 @@ var MetricsEnabled bool
 type Config struct {
 	IP   string
 	Port string
+	// TLSEnabled specifies whether the connection is secure or not.
+	// PLEASE NOTE: it should be set to true in order to handle TLSPath and/or XTokenPath.
+	TLSEnabled bool
 	// TLSPath specifies the directory path where the TLS certificates are stored.
 	// It should not include file names('cert.pem' and 'key.pem').
 	// If left empty, the client will be configured for an insecure (non-TLS) connection.
--- a/nodebuilder/core/config.go
+++ b/nodebuilder/core/config.go
@@ -18,12 +18,8 @@ type Config struct {
 	IP   string
 	Port string
 	// TLSEnabled specifies whether the connection is secure or not.
-	// PLEASE NOTE: it should be set to true in order to handle TLSPath and/or XTokenPath.
+	// PLEASE NOTE: it should be set to true in order to handle XTokenPath.
 	TLSEnabled bool
-	// TLSPath specifies the directory path where the TLS certificates are stored.
-	// It should not include file names('cert.pem' and 'key.pem').
-	// If left empty, the client will be configured for an insecure (non-TLS) connection.
-	TLSPath string
 	// XTokenPath specifies the path to the directory with JSON file containing the X-Token for gRPC authentication.
 	// The JSON file should have a key-value pair where the key is "x-token" and the value is the authentication token.
 	// If left empty, the client will not include the X-Token in its requests.
--- a/nodebuilder/core/flags.go
+++ b/nodebuilder/core/flags.go
@@ -11,8 +11,8 @@ var (
 	coreFlag           = "core.ip"
 	coreGRPCFlag       = "core.grpc.port"
 	coreTLS            = "core.tls"
-	coreTLSPathFlag    = "core.grpc.tls.path"
-	coreXTokenPathFlag = "core.grpc.xtoken.path" //nolint:gosec
+	coreTLSPathFlag    = "core.tls.path"
+	coreXTokenPathFlag = "core.xtoken.path" //nolint:gosec
 )
 
 // Flags gives a set of hardcoded Core flags.
--- a/nodebuilder/core/flags.go
+++ b/nodebuilder/core/flags.go
@@ -11,6 +11,7 @@ var (
 	coreFlag          = "core.ip"
 	coreGRPCFlag      = "core.grpc.port"
 	coreEnableTLSFlag = "core.grpc.tls"
+	coreTLSPAthFlag   = "core.grpc.tls.path"
 )
 
 // Flags gives a set of hardcoded Core flags.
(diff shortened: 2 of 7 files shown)


nodebuilder/core/tls.go
```diff
@@ -0,0 +1,85 @@
...
+// Parameters:
+// * tlsPath: The directory path where the TLS certificate ("cert.pem") and key ("key.pem") files are located.
+// Returns:
+// * A tls.Config structure configured with the provided certificate and key.
+// * An error if the certificate or key file does not exist, or if loading the key pair fails.
+func TLS(tlsPath string) (*tls.Config, error) {
+	certPath := filepath.Join(tlsPath, cert)
+	keyPath := filepath.Join(tlsPath, key)
+	exist := utils.Exists(certPath) && utils.Exists(keyPath)
+	if !exist {
```
vgonkivs, line 37: https://github.com/celestiaorg/celestia-node/pull/3922#discussion_r1850188745

nodebuilder/state/core.go
```diff
@@ -30,14 +33,31 @@ func coreAccessor(
...
+			tlsCfg = core.EmptyTLSConfig()
+		default:
+			return nil, nil, nil, err
+		}
+
+		xtoken, err := core.XToken(corecfg.XTokenPath)
+		if err != nil && !errors.Is(err, os.ErrNotExist) {
+			return nil, nil, nil, err
+		}
+		opts = append(opts, state.WithTLSConfig(tlsCfg), state.WithXToken(xtoken))
```
vgonkivs, line 51: https://github.com/celestiaorg/celestia-node/pull/3922#discussion_r1850192010

state/core_access.go
```diff
@@ -601,6 +604,50 @@ func (ca *CoreAccessor) setupTxClient(ctx context.Context, keyName string) (*use
...
+	if ca.xtoken != "" {
+		authInterceptor := func(ctx context.Context,
+			method string,
+			req, reply interface{},
+			cc *grpc.ClientConn,
+			invoker grpc.UnaryInvoker,
+			opts ...grpc.CallOption,
+		) error {
+			ctx = metadata.AppendToOutgoingContext(ctx, "x-token", ca.xtoken)
+			return invoker(ctx, method, req, reply, cc, opts...)
```
vgonkivs, line 627: https://github.com/celestiaorg/celestia-node/pull/3922#discussion_r1850196622

renaynay (commented):
@vgonkivs this isn't a breaking change bc those fields are purely additive - nothing about validation logic for config has changed so no worries to base it on api-breaks. You can re-base it on main actually.

nodebuilder/core/config.go
```diff
@@ -19,13 +19,23 @@ type Config struct {
 	IP       string
 	RPCPort  string
 	GRPCPort string
+	// TLSEnabled specifies whether the connection is secure or not.
+	// PLEASE NOTE: it should be set to true in order to handle TLSPath and/or XTokenPath.
+	TLSEnabled bool
```
renaynay, line 24: Do we need this if we just provide a TLSPath?

nodebuilder/core/flags.go
```diff
@@ -8,9 +8,12 @@ import (
 )
 
 var (
-	coreFlag     = "core.ip"
-	coreRPCFlag  = "core.rpc.port"
-	coreGRPCFlag = "core.grpc.port"
+	coreFlag           = "core.ip"
+	coreRPCFlag        = "core.rpc.port"
```
renaynay, line 12: if you rebase, you won't need this anymore and also won't need to specify `grpc` in the name of the flag

so `core.grpc.tls.path` --> `core.tls.path` and same for xtoken

nodebuilder/core/config.go
vgonkivs: Yes, we need it for a use case when we have to provide an empty config.

vgonkivs:
> @vgonkivs this isn't a breaking change bc those fields are purely additive - nothing about validation logic for config has changed so no worries to base it on api-breaks. You can re-base it on main actually.

I'd propose merging it inside api breaks as we have one more PR related to the grpc connection. So, imo it's better to keep everything in one place.

walldiss:
I was reviewing the PR and noticed that we're configuring client certificates, which seems to be for setting up mutual TLS (mTLS). My understanding is that we only need to verify the server's certificate since we'll be sending an x-token authorization header for client authentication. Do we really need client certificates on the client side in this case?

Also, I didn't see any setup for root CA certificates, which are necessary when dealing with self-signed or custom certificates—a pretty common use case. Without specifying the root CA, the client won't be able to validate the server's certificate.

Could we adjust the implementation to focus on server-side certificates and include support for custom root CAs? What are your thoughts?

vgonkivs:
We agreed with @walldiss in DM that we don't want to keep any certificates for now(as they were out of the scope of the initial request)

nodebuilder/core/tls.go
```diff
@@ -0,0 +1,58 @@
...
+	"errors"
+	"os"
+	"path/filepath"
+
+	"github.com/celestiaorg/celestia-node/libs/utils"
+)
+
+const (
+	cert   = "cert.pem"
+	key    = "key.pem"
```
walldiss, line 15: unused?

nodebuilder/core/tls.go
```diff
@@ -0,0 +1,58 @@
...
+	xtoken = "xtoken.json"
+)
+
+func EmptyTLSConfig() *tls.Config {
+	return &tls.Config{MinVersion: tls.VersionTLS12}
+}
```
walldiss, line 21: nit: don't need to hide details into func, would be easier to read if value used in place

nodebuilder/core/tls.go
```diff
@@ -0,0 +1,58 @@
...
+// It first constructs the full file path by joining the provided directory path with the token file name.
+// If the file does not exist, it returns an os.ErrNotExist error.
+// If the file exists, it reads the content and unmarshalls it.
+// If the field in the unmarshalled struct is empty, an error is returned indicating that the token is missing.
+// Parameters:
+// * xtokenPath: The directory path where the JSON file containing the X-Token is located.
+// Returns:
+// * The X-Token as a string if successfully retrieved.
+// * An error if the file does not exist, reading fails, unmarshalling fails, or the token is empty.
+func XToken(xtokenPath string) (string, error) {
```
walldiss, line 37: Feels a bit too much comments for such simple function. Would you consider rename to MustReadXToken, to indicate the read and requirement for non-empty value?

state/core_access.go
```diff
@@ -602,6 +605,41 @@ func (ca *CoreAccessor) setupTxClient(ctx context.Context, keyName string) (*use
...
+	endpoint := net.JoinHostPort(ca.coreIP, ca.port)
+	// By default, the gRPC client is configured to handle an insecure connection.
+	// If the TLS configuration is not empty, it will be applied to the client's options.
+	// If the TLS configuration is empty but the X-Token is provided,
+	// the X-Token will be applied as an interceptor along with an empty TLS configuration.
+	opts := []grpc.DialOption{grpc.WithTransportCredentials(insecure.NewCredentials())}
+	if ca.tls != nil {
+		opts = append(opts, grpc.WithTransportCredentials(credentials.NewTLS(ca.tls)))
+	}
+
```
walldiss, line 619: Should be if {...} else {...}, otherwise you will end up with 2 TransportCredentials options in slice when tls is enabled

state/core_access.go
```diff
@@ -602,6 +605,41 @@ func (ca *CoreAccessor) setupTxClient(ctx context.Context, keyName string) (*use
...
+	if interceptor := authInterceptor(ca.xtoken); interceptor != nil {
+		opts = append(opts, grpc.WithUnaryInterceptor(interceptor))
+	}
```
walldiss, line 622: nit:
Would be cleaner and would remove the check from interceptor constructor with:
if ca.xtoken != "" { add interceptor}

state/core_access.go
vgonkivs: Yes. That's the idea. TLS overrides insecure

nodebuilder/core/tls.go
vgonkivs: We are using `Must...` + panic inside. I do not want to break this pattern bc panic is not expected there.

nodebuilder/core/tls.go
vgonkivs: The idea was to keep the dependency inside core package and does not bloat all other places

nodebuilder/core/tls.go
vgonkivs: > Feels a bit too much comments for such simple function.

will reduce

state/core_access.go
vgonkivs: done

cristaloleg (commented):
SGTM

nodebuilder/core/flags.go
```diff
@@ -49,6 +64,18 @@ func ParseFlags(
 		cfg.Port = grpc
 	}
 
+	enabled, err := cmd.Flags().GetBool(coreTLS)
+	if err != nil {
+		panic(err)
```
renaynay, line 69: can we just return err?

nodebuilder/core/tls.go
```diff
@@ -0,0 +1,45 @@
...
+	"crypto/tls"
+	"encoding/json"
+	"errors"
+	"os"
+	"path/filepath"
+
+	"github.com/celestiaorg/celestia-node/libs/utils"
+)
+
+const xtoken = "xtoken.json"
```
renaynay, line 13: nit: xtokenFileName

nodebuilder/core/tls.go
```diff
@@ -0,0 +1,45 @@
...
+
+func EmptyTLSConfig() *tls.Config {
+	return &tls.Config{MinVersion: tls.VersionTLS12}
+}
+
+type AuthToken struct {
```
renaynay, line 19: we don't need this struct really, let's make it local to the function and just return string. No need for it to be separate from func or even exported.

nodebuilder/state/core.go
renaynay: So even if we have empty string for xtoken here, we still set state.WithXToken?

walldiss (approved):

nodebuilder/state/core.go
renaynay: I guess it's fine bc it's implicitly handled inside core access when it sets up the gRPC but it's still odd. It's fine tho just a nit, feel free to resolve.

renaynay (commented):
happy with it, just some nits.

walldiss:
Looks good and simple. Additionaly it might be worth to add grpc layer unit tests for:
 - tls enforced mod
 - x-token can be retrieved by server

nodebuilder/state/core.go
vgonkivs: > So even if we have empty string for xtoken here, we still set state.WithXToken?

It will set an empty string, which will be set anyway in core_accessor if we don't pass anything.

nodebuilder/core/tls.go
vgonkivs: Makes sense. Thank you

renaynay (approved):

PR #3957, opened 2024-11-22
refactor(share/availability/light | share/availability/full): Availability implementations are aware of sampling window, removed from DASer

renaynay:
This PR makes the concept of `archival` node explicit, splitting it away from sampling window. Now, both the node's pruning status (archival or not) and the sampling window will have to be provided to availability implementations for them to determine what to sync + what to store. This also applies to `core` package for BNs as that's their "availability implementation" technically.

### DIFF
--- a/core/option.go
+++ b/core/option.go
@@ -12,11 +12,13 @@ type params struct {
 	metrics            bool
 	chainID            string
 	availabilityWindow time.Duration
+	archival           bool
 }
 
 func defaultParams() params {
 	return params{
 		availabilityWindow: time.Duration(0),
+		archival:           false,
 	}
 }
 
@@ -37,5 +39,11 @@ func WithChainID(id p2p.Network) Option {
 func WithAvailabilityWindow(window time.Duration) Option {
 	return func(p *params) {
 		p.availabilityWindow = window
+	}
+}
+
+func WithArchivalMode() Option {
+	return func(p *params) {
+		p.archival = true
 	}
 }
--- a/share/availability/full/availability.go
+++ b/share/availability/full/availability.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"time"
 
 	logging "github.com/ipfs/go-log/v2"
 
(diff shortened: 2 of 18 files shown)


core/option.go
```diff
@@ -12,11 +12,13 @@ type params struct {
 	metrics            bool
 	chainID            string
 	availabilityWindow time.Duration
+	archival           bool
```
renaynay, line 15: Unfortunately, as `core` is technically the implicit `Availability` implementation for BN, it needs to be aware of storage status. 

FN + LN rely on their respective availability implementations to handle this.

share/availability/light/availability_test.go
```diff
@@ -517,6 +517,7 @@ func randEdsAndHeader(t *testing.T, size int) (*rsmt2d.ExtendedDataSquare, *head
 	h := &header.ExtendedHeader{
 		RawHeader: header.RawHeader{
 			Height: int64(height),
+			Time:   time.Now(),
```
cristaloleg, line 520: Why it worked before this change? 🤔

share/availability/full/availability.go
```diff
@@ -23,23 +24,44 @@ var log = logging.Logger("share/full")
...
 // SharesAvailable reconstructs the data committed to the given Root by requesting
 // enough Shares from the network.
 func (fa *ShareAvailability) SharesAvailable(ctx context.Context, header *header.ExtendedHeader) error {
 	dah := header.DAH
+
+	if !fa.archival {
+		// do not sync blocks outside of sampling window if not archival
+		if !availability.IsWithinWindow(header.Time(), fa.storageWindow) {
+			log.Debugw("skipping availability check for block outside sampling"+
+				" window", "height", header.Height(), "data hash", dah.String())
```
cristaloleg, line 60: Just curious why to split `window` word?

renaynay: Unintentional, it's just string

share/availability/full/options.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+}
+
+// WithStorageWindow is a functional option to set the storage window
+// to something other than the default for the full availability implementation
+func WithStorageWindow(window time.Duration) Option {
+	return func(p *params) {
+		p.storageWindow = window
+	}
+}
```
walldiss, line 41: This one looks like only be used in tests. And I think it should not be exposed to user, since we rely on its value on shrex clients. How about remove it from params and put on availability struct with hardcoded values (similar to light availability)?

renaynay: Yeah i was thinking to do that too - will do.

cristaloleg (commented):
SGTM

walldiss (approved):

cristaloleg (approved):

vgonkivs (approved):

Issue #3956, opened 2024-11-21
libshare.Blob json marshaling is broken, making `state.SubmitPayForBlob` unusable


zvolin:
### Celestia Node version

v0.20.2

### OS

nixos unstable with asahi

### Install tools

docker

### Steps to reproduce it

```shell
git clone https://github.com/celestiaorg/go-square && cd go-square 

patch -p1 --ignore-whitespace <<'EOF'
diff --git a/share/blob_test.go b/share/blob_test.go
index 771328e..fa58b0c 100644
--- a/share/blob_test.go
+++ b/share/blob_test.go
@@ -3,6 +3,8 @@ package share
 import (
        "bytes"
        "crypto/rand"
+       "encoding/json"
+       "fmt"
        "testing"

        v1 "github.com/celestiaorg/go-square/v2/proto/blob/v1"
@@ -22,6 +24,18 @@ func TestProtoEncoding(t *testing.T) {
        newBlob, err := UnmarshalBlob(blobBytes)
        require.NoError(t, err)

+       blobJson, err := json.Marshal(blob)
+       require.NoError(t, err)
+
+       fmt.Println("blob", *blob)
+       fmt.Println("json blob", string(blobJson))
+
+       blob = &Blob{}
+       err = json.Unmarshal(blobJson, blob)
+       require.NoError(t, err)
+
+       fmt.Println("blob", *blob)
+
        require.Equal(t, blob, newBlob)
 }

EOF

make test
```

### Expected result

`state.SubmitPayForBlob` to work

### Actual result

```
⇒ {
    "jsonrpc":"2.0",
    "id":1,
    "method":"state.SubmitPayForBlob",
    "params":[
        [{
            "namespace_id":"AAAAAAAAAAAAAAAAAAAAAAAAOUuI4+yg+3IeEw==",
[...]

PR #3958, opened 2024-11-22
chore(go.mod) bump go-square

vgonkivs:
Bump go-square version + add a test to ensure that the fix works.

Closes https://github.com/celestiaorg/celestia-node/issues/3956

Screenshot 2024-11-22 at 16 06 07


renaynay (commented):
Let's wait to do proper release on go-square?

cristaloleg (approved):

walldiss (dismissed):

nodebuilder/tests/api_test.go
```diff
@@ -168,3 +183,43 @@ func TestHeaderSubscription(t *testing.T) {
...
+	ctx, cancel := context.WithTimeout(context.Background(), swamp.DefaultTestTimeout)
+	t.Cleanup(cancel)
+
+	sw := swamp.NewSwamp(t, swamp.WithBlockTime(time.Second))
+	// start a bridge node
+	bridge := sw.NewBridgeNode()
+	err := bridge.Start(ctx)
+	require.NoError(t, err)
+
+	bridgeAddr := "http://" + bridge.RPCServer.ListenAddr()
```
cristaloleg, line 197: What about to move closer to place of 1st usage?

nodebuilder/tests/api_test.go
```diff
@@ -168,3 +183,43 @@ func TestHeaderSubscription(t *testing.T) {
...
+
+	f, err := os.Open("test_data/submitPFB.json")
+	require.NoError(t, err)
+	defer f.Close()
+
+	payload, err := io.ReadAll(f)
+	require.NoError(t, err)
+
+	req, err := http.NewRequest("POST", bridgeAddr, bytes.NewBuffer(payload))
+	if err != nil {
```
cristaloleg, line 211: Maybe `require.NoError` ?

nodebuilder/tests/api_test.go
```diff
@@ -168,3 +183,43 @@ func TestHeaderSubscription(t *testing.T) {
...
+
+	adminPerms := []auth.Permission{"public", "read", "write", "admin"}
+	jwt, err := bridge.AdminServ.AuthNew(ctx, adminPerms)
+	require.NoError(t, err)
+
+	f, err := os.Open("test_data/submitPFB.json")
+	require.NoError(t, err)
+	defer f.Close()
+
+	payload, err := io.ReadAll(f)
```
cristaloleg, line 207: We have https://pkg.go.dev/os#ReadFile

nodebuilder/tests/test_data/submitPFB.json
```diff
@@ -0,0 +1,14 @@
+{
```
cristaloleg, line 1: Directory should be `testdata` not `test_data`.

cristaloleg (commented):
Some things in tests

walldiss (dismissed):

renaynay (approved):
Thanks for adding test

nodebuilder/tests/api_test.go
```diff
@@ -168,3 +182,35 @@ func TestHeaderSubscription(t *testing.T) {
 	err = light.Stop(ctx)
 	require.NoError(t, err)
 }
+
+func TestSubmitBlobOverHTTP(t *testing.T) {
```
walldiss, line 186: This is the same test as  TestBlobRPC, but without usage of client lib, instead it uses raw json request. The downside of raw api tests is that every time api is changed they would need to be manually updated.

cristaloleg (approved):

walldiss (approved):

walldiss (approved):

nodebuilder/tests/api_test.go
renaynay: That's the point, we want this (for every method though)

walldiss: Approved bacause it is fine as long as there not many tests like that. Because it may become burden to maintain such tests. For future API work (Canonical api) we should rather do backwards compatibility tests

PR #3967, opened 2024-11-27
fix: jwt token nonce and expiration time

### LINKED ISSUE #2420: libs: Add expiry for JWT

cristaloleg:
Fixes #2420

### DIFF
--- a/api/rpc/perms/permissions.go
+++ b/api/rpc/perms/permissions.go
@@ -33,15 +33,20 @@ func (j *JWTPayload) MarshalBinary() (data []byte, err error) {
 // NewTokenWithPerms generates and signs a new JWT token with the given secret
 // and given permissions.
 func NewTokenWithPerms(signer jwt.Signer, perms []auth.Permission, ttl time.Duration) ([]byte, error) {
-	var nonce [32]byte
-	if _, err := rand.Read(nonce[:]); err != nil {
+	nonce := make([]byte, 32)
+	if _, err := rand.Read(nonce); err != nil {
 		return nil, err
+	}
+
+	var expiresAt time.Time
+	if ttl != 0 {
+		expiresAt = time.Now().UTC().Add(ttl)
 	}
 
 	p := &JWTPayload{
 		Allow:     perms,
 		Nonce:     nonce[:],
-		ExpiresAt: time.Now().UTC().Add(ttl),
+		ExpiresAt: expiresAt,
 	}
 	token, err := jwt.NewBuilder(signer).Build(p)
 	if err != nil {
--- a/api/rpc/perms/permissions.go
+++ b/api/rpc/perms/permissions.go
@@ -32,7 +32,26 @@ func (j *JWTPayload) MarshalBinary() (data []byte, err error) {
 
 // NewTokenWithPerms generates and signs a new JWT token with the given secret
 // and given permissions.
-func NewTokenWithPerms(signer jwt.Signer, perms []auth.Permission, ttl time.Duration) ([]byte, error) {
+func NewTokenWithPerms(signer jwt.Signer, perms []auth.Permission) ([]byte, error) {
+	nonce := make([]byte, 32)
+	if _, err := rand.Read(nonce); err != nil {
+		return nil, err
+	}
+
+	p := &JWTPayload{
+		Allow: perms,
+		Nonce: nonce,
+	}
+	token, err := jwt.NewBuilder(signer).Build(p)
+	if err != nil {
+		return nil, err
+	}
+	return token.Bytes(), nil
+}
+
+// NewTokenWithTTL generates and signs a new JWT token with the given secret
+// and given permissions and TTL.
+func NewTokenWithTTL(signer jwt.Signer, perms []auth.Permission, ttl time.Duration) ([]byte, error) {
 	nonce := make([]byte, 32)
 	if _, err := rand.Read(nonce); err != nil {
 		return nil, err
--- a/api/rpc/perms/permissions.go
+++ b/api/rpc/perms/permissions.go
@@ -33,20 +33,7 @@ func (j *JWTPayload) MarshalBinary() (data []byte, err error) {
 // NewTokenWithPerms generates and signs a new JWT token with the given secret
 // and given permissions.
 func NewTokenWithPerms(signer jwt.Signer, perms []auth.Permission) ([]byte, error) {
-	nonce := make([]byte, 32)
-	if _, err := rand.Read(nonce); err != nil {
-		return nil, err
-	}
-
-	p := &JWTPayload{
-		Allow: perms,
-		Nonce: nonce,
-	}
-	token, err := jwt.NewBuilder(signer).Build(p)
-	if err != nil {
-		return nil, err
-	}
-	return token.Bytes(), nil
+	return NewTokenWithTTL(signer, perms, 0)
 }
 
 // NewTokenWithTTL generates and signs a new JWT token with the given secret
--- a/api/rpc/perms/permissions.go
+++ b/api/rpc/perms/permissions.go
@@ -1,7 +1,9 @@
 package perms
 
 import (
+	"crypto/rand"
 	"encoding/json"
+	"time"
 
 	"github.com/cristalhq/jwt/v5"
 	"github.com/filecoin-project/go-jsonrpc/auth"
@@ -19,7 +21,9 @@ var AuthKey = "Authorization"
 // JWTPayload is a utility struct for marshaling/unmarshalling
 // permissions into for token signing/verifying.
 type JWTPayload struct {
-	Allow []auth.Permission
+	Allow     []auth.Permission
+	Nonce     []byte
+	ExpiresAt time.Time
 }
 
 func (j *JWTPayload) MarshalBinary() (data []byte, err error) {
@@ -28,9 +32,16 @@ func (j *JWTPayload) MarshalBinary() (data []byte, err error) {
 
 // NewTokenWithPerms generates and signs a new JWT token with the given secret
 // and given permissions.
-func NewTokenWithPerms(signer jwt.Signer, perms []auth.Permission) ([]byte, error) {
+func NewTokenWithPerms(signer jwt.Signer, perms []auth.Permission, ttl time.Duration) ([]byte, error) {
+	var nonce [32]byte
+	if _, err := rand.Read(nonce[:]); err != nil {
+		return nil, err
+	}
+
 	p := &JWTPayload{
-		Allow: perms,
+		Allow:     perms,
+		Nonce:     nonce[:],
+		ExpiresAt: time.Now().UTC().Add(ttl),
 	}
 	token, err := jwt.NewBuilder(signer).Build(p)
 	if err != nil {
--- a/api/rpc/perms/permissions.go
+++ b/api/rpc/perms/permissions.go
@@ -45,7 +45,7 @@ func NewTokenWithPerms(signer jwt.Signer, perms []auth.Permission, ttl time.Dura
 
 	p := &JWTPayload{
 		Allow:     perms,
-		Nonce:     nonce[:],
+		Nonce:     nonce,
 		ExpiresAt: expiresAt,
 	}
 	token, err := jwt.NewBuilder(signer).Build(p)
--- a/api/rpc_test.go
+++ b/api/rpc_test.go
@@ -59,7 +59,7 @@ func TestRPCCallsUnderlyingNode(t *testing.T) {
 	nd, server := setupNodeWithAuthedRPC(t, signer, verifier)
 	url := nd.RPCServer.ListenAddr()
 
-	adminToken, err := perms.NewTokenWithPerms(signer, perms.AllPerms, time.Minute)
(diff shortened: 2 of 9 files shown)


api/rpc/perms/permissions.go
```diff
@@ -28,9 +32,16 @@ func (j *JWTPayload) MarshalBinary() (data []byte, err error) {
...
+	var nonce [32]byte
+	if _, err := rand.Read(nonce[:]); err != nil {
+		return nil, err
+	}
+
 	p := &JWTPayload{
-		Allow: perms,
+		Allow:     perms,
+		Nonce:     nonce[:],
+		ExpiresAt: time.Now().UTC().Add(ttl),
```
cristaloleg, line 44: `.UTC()` to prevent timezone issues.

api/rpc_test.go
```diff
@@ -59,7 +59,7 @@ func TestRPCCallsUnderlyingNode(t *testing.T) {
 	nd, server := setupNodeWithAuthedRPC(t, signer, verifier)
 	url := nd.RPCServer.ListenAddr()
 
-	adminToken, err := perms.NewTokenWithPerms(signer, perms.AllPerms)
+	adminToken, err := perms.NewTokenWithPerms(signer, perms.AllPerms, time.Minute)
```
cristaloleg, line 62: Should be fine for the test.

cmd/auth.go
```diff
@@ -34,6 +35,11 @@ func AuthCmd(fsets ...*flag.FlagSet) *cobra.Command {
 				return err
 			}
 
+			ttl, err := cmd.Flags().GetDuration("ttl")
```
cristaloleg, line 38: Should we have a big default here or require a well-defined value all the time?

cmd/rpc.go
```diff
@@ -128,7 +128,7 @@ func getToken(path string) (string, error) {
 		fmt.Printf("error getting the JWT secret: %v", err)
 		return "", err
 	}
-	return buildJWTToken(key.Body, perms.AllPerms)
+	return buildJWTToken(key.Body, perms.AllPerms, time.Minute)
```
cristaloleg, line 131: Fine for rpc, I think?

nodebuilder/node/cmd/node.go
```diff
@@ -99,7 +99,12 @@ var authCmd = &cobra.Command{
 			perms[i] = (auth.Permission)(p)
 		}
 
-		result, err := client.Node.AuthNew(c.Context(), perms)
+		ttl, err := cmd.Flags().GetDuration("ttl")
```
cristaloleg, line 102: Same here, do we need a default value or require all the time?

nodebuilder/node/cmd/node.go
```diff
@@ -87,8 +87,8 @@ var authCmd = &cobra.Command{
 	Use:   "set-permissions",
 	Args:  cobra.MinimumNArgs(1),
 	Short: "Signs and returns a new token with the given permissions.",
-	RunE: func(c *cobra.Command, args []string) error {
-		client, err := cmdnode.ParseClientFromCtx(c.Context())
+	RunE: func(cmd *cobra.Command, args []string) error {
```
cristaloleg, line 90: Rename for consistency with other usages.

libs/authtoken/authtoken.go
```diff
@@ -17,17 +20,27 @@ func ExtractSignedPermissions(verifier jwt.Verifier, token string) ([]auth.Permi
 		return nil, err
 	}
 	p := new(perms.JWTPayload)
-	err = json.Unmarshal(tk.Claims(), p)
-	if err != nil {
+
+	if err := json.Unmarshal(tk.Claims(), p); err != nil {
 		return nil, err
 	}
+	if p.ExpiresAt.Before(time.Now().UTC()) {
```
cristaloleg, line 27: AFAIK that's the only place where we need to validate time.

cmd/auth.go
vgonkivs: Id short-circuit immediately if TTL is not passed. I don't see any reason why we should pick TTL for the user.

api/rpc/perms/permissions.go
renaynay: @cristaloleg let's please only add a TTL if one is specified. I think we should preserve the default behaviour which is token is eternal.

api/rpc_test.go
renaynay: Can you also add test where token is valid for some time, then is invalid after expiry? I know it's time-based so it's tricky but would be nice to have that coverage.

cmd/auth.go
renaynay: + to slava's comment. We should not set a TTL - our default should be that the key is eternal unless TTL is specified.

api/rpc/perms/permissions.go
```diff
@@ -28,9 +32,16 @@ func (j *JWTPayload) MarshalBinary() (data []byte, err error) {
 
 // NewTokenWithPerms generates and signs a new JWT token with the given secret
 // and given permissions.
-func NewTokenWithPerms(signer jwt.Signer, perms []auth.Permission) ([]byte, error) {
+func NewTokenWithPerms(signer jwt.Signer, perms []auth.Permission, ttl time.Duration) ([]byte, error) {
+	var nonce [32]byte
```
walldiss, line 36: Is there particular reason to use fixed sized array here? if slice is used instead of array it will remove convertions by [:]

cmd/rpc.go
renaynay: No actually, users usually use rpc cli locally or for debugging, I don't think auth token should expire for those purposes. Let's pls leave default behaviour.

nodebuilder/node/node.go
```diff
@@ -23,18 +24,18 @@ type Module interface {
 	// AuthVerify returns the permissions assigned to the given token.
 	AuthVerify(ctx context.Context, token string) ([]auth.Permission, error)
 	// AuthNew signs and returns a new token with the given permissions.
-	AuthNew(ctx context.Context, perms []auth.Permission) (string, error)
+	AuthNew(ctx context.Context, perms []auth.Permission, ttl time.Duration) (string, error)
```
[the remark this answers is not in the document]
renaynay: Let's pls avoid break in this

cmd/auth.go
walldiss: Force them to pass, so they know when to expect expiry

nodebuilder/node/node.go
```diff
@@ -23,18 +24,18 @@ type Module interface {
...
-		Info        func(context.Context) (Info, error)                                `perm:"admin"`
-		Ready       func(context.Context) (bool, error)                                `perm:"read"`
-		LogLevelSet func(ctx context.Context, name, level string) error                `perm:"admin"`
-		AuthVerify  func(ctx context.Context, token string) ([]auth.Permission, error) `perm:"admin"`
-		AuthNew     func(ctx context.Context, perms []auth.Permission) (string, error) `perm:"admin"`
+		Info        func(context.Context) (Info, error)                                                   `perm:"admin"`
+		Ready       func(context.Context) (bool, error)                                                   `perm:"read"`
+		LogLevelSet func(ctx context.Context, name, level string) error                                   `perm:"admin"`
+		AuthVerify  func(ctx context.Context, token string) ([]auth.Permission, error)                    `perm:"admin"`
+		AuthNew     func(ctx context.Context, perms []auth.Permission, ttl time.Duration) (string, error) `perm:"admin"`
```
walldiss, line 38: API break, need to update PR description

api/rpc/perms/permissions.go
cristaloleg: Done

libs/authtoken/authtoken.go
```diff
@@ -17,17 +20,27 @@ func ExtractSignedPermissions(verifier jwt.Verifier, token string) ([]auth.Permi
...
+	var nonce [32]byte
+	if _, err := rand.Read(nonce[:]); err != nil {
+		return "", err
+	}
+
 	token, err := jwt.NewBuilder(signer).Build(&perms.JWTPayload{
-		Allow: permissions,
+		Allow:     permissions,
+		Nonce:     nonce[:],
+		ExpiresAt: time.Now().UTC().Add(ttl),
```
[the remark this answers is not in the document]
cristaloleg: Done

libs/authtoken/authtoken.go
```diff
@@ -17,17 +20,27 @@ func ExtractSignedPermissions(verifier jwt.Verifier, token string) ([]auth.Permi
...
 	}
 	p := new(perms.JWTPayload)
-	err = json.Unmarshal(tk.Claims(), p)
-	if err != nil {
+
+	if err := json.Unmarshal(tk.Claims(), p); err != nil {
 		return nil, err
 	}
+	if p.ExpiresAt.Before(time.Now().UTC()) {
+		return nil, fmt.Errorf("token expired %s ago", time.Since(p.ExpiresAt))
```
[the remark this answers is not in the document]
cristaloleg: `time.Duration` has `.String()` method, are you talking about that? https://pkg.go.dev/time#Duration.String

nodebuilder/node/node.go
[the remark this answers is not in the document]
cristaloleg: Done, have added a new method

api/rpc/perms/permissions.go
cristaloleg: ok, fixed.

nodebuilder/node/node.go
cristaloleg: Reverted

api/rpc/perms/permissions.go
```diff
@@ -28,9 +32,21 @@ func (j *JWTPayload) MarshalBinary() (data []byte, err error) {
 
 // NewTokenWithPerms generates and signs a new JWT token with the given secret
 // and given permissions.
-func NewTokenWithPerms(signer jwt.Signer, perms []auth.Permission) ([]byte, error) {
+func NewTokenWithPerms(signer jwt.Signer, perms []auth.Permission, ttl time.Duration) ([]byte, error) {
```
renaynay, line 35: ok sorry missed this one but can you make a separate public method that allows for setting ttl and make this `NewTokenWithPerms` rely on it and not set a TTL ? 

so `NewTokenWithPerms` and `NewTokenWithTTL`

api/rpc_test.go
```diff
@@ -89,6 +89,42 @@ func TestRPCCallsUnderlyingNode(t *testing.T) {
...
+		if err == nil {
+			t.Cleanup(rpcClient.Close)
+			break
+		}
+	}
+	require.NotNil(t, rpcClient)
+	require.NoError(t, err)
+
+	_, err = rpcClient.State.Balance(ctx)
+	require.Error(t, err, err)
```
renaynay, line 125: pls assert err is == smth or contains somth

cmd/rpc.go
renaynay: needs addressing pls - let's not set expiry here.

libs/authtoken/authtoken.go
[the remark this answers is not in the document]
renaynay: Yes

cmd/auth.go
```diff
@@ -34,6 +35,8 @@ func AuthCmd(fsets ...*flag.FlagSet) *cobra.Command {
 				return err
 			}
 
+			ttl, _ := cmd.Flags().GetDuration("ttl")
```
renaynay, line 38: does this add `--ttl` to the flagset for the auth cmd? also if no `--ttl` is passed then the default for an unchanged/nil Duration == 0? Can we make explicit pls

api/rpc_test.go
cristaloleg: My bad, fixed.

api/rpc/perms/permissions.go
cristaloleg: Ah, okay. Now it clear. Have added a separate function with TTL.

cmd/auth.go
cristaloleg: I'm not `cobra` expert. If you know this part better, please make a push.

libs/authtoken/authtoken.go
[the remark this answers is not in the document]
cristaloleg: It will be used automatically due to `%s` so we are safe here.

api/rpc/perms/permissions.go
```diff
@@ -32,7 +32,26 @@ func (j *JWTPayload) MarshalBinary() (data []byte, err error) {
...
+
+	p := &JWTPayload{
+		Allow: perms,
+		Nonce: nonce,
+	}
+	token, err := jwt.NewBuilder(signer).Build(p)
+	if err != nil {
+		return nil, err
+	}
+	return token.Bytes(), nil
```
renaynay, line 49: nit but this can use NewTokenWithTTL and just not set TTL

renaynay (approved):
tested manually as well 

`auth` cmd only has `--ttl` flag so it's not a global flag. If it is not passed, ttl will be set to 0, disabling the TTL on the token (so it's eternal).

This PR preserves old behaviour but allows for TTLs on the tokens and also makes the token generated non-deterministic per secret.

vgonkivs (approved):

PR #3973, opened 2024-11-29
chore: make justfile for better arabica devX

Bidon15:
gm chat

As devs are lazy living beings, I propose we automate celestia light node in arabica-11 with a simple 1 command for better devX:

```sh
just light arabica up
```

Which encompasses these features: 

- initialising a new node
- querying the latest block height from arabica-11
- updates the `config.toml` to omit historical sync
- funds the local celestia account from the faucet backend
- skips auth token
- changes the rpc address to 0.0.0.0

Closes an elephant in the room for solutions team and many more execution environment partners. 
Co-authored with @sysrex 

Next steps: 

- [ ] Make it as part of CI automation job (@sysrex has context)
- [ ] Document into the ReadMe file if you, @Ferret-san @jcstein @S1nus @distractedm1nd  like this idea

### DIFF
--- a/justfile
+++ b/justfile
@@ -39,6 +39,7 @@ check-and-fund:
         echo "Waiting 10 seconds for transaction to process..."
         sleep 10
     fi
+
 # Reset node state and update config with latest block height
 reset-node:
     #!/usr/bin/env bash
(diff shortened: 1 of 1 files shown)


renaynay (approved):
utack

renaynay:
The only concern I'd have is hardcoded core.ip `            --core.ip validator-1.celestia-arabica-11.com \`

but otherwise fine

jcstein:
should this use make instead of just?

PR #3189, opened 2024-02-19
recompute hash

expertdicer:
Closes: #2959 
Recompute hash for the ExtendedHeader

PR #3975, opened 2024-12-02
feat(header): cache hash

### LINKED ISSUE #2959: header.go: `Hash()` should recompute / read from cached value of actually computed hash of header
Rather than taking it from the commit BlockID.

Also implement caching inside the header struct on computation.

cristaloleg:
Closes #2959 and Closes #3189

### DIFF
--- a/header/header.go
+++ b/header/header.go
@@ -40,7 +40,7 @@ type ExtendedHeader struct {
 	ValidatorSet *core.ValidatorSet         `json:"validator_set"`
 	DAH          *da.DataAvailabilityHeader `json:"dah"`
 
-	hashOnce sync.Once
+	hashOnce *sync.Once
 	hash     []byte
 }
 
--- a/header/header.go
+++ b/header/header.go
@@ -14,6 +14,7 @@ import (
 	"github.com/celestiaorg/celestia-app/v3/pkg/appconsts"
 	"github.com/celestiaorg/celestia-app/v3/pkg/da"
 	libhead "github.com/celestiaorg/go-header"
+	"github.com/celestiaorg/go-square/merkle"
 	"github.com/celestiaorg/rsmt2d"
 )
 
@@ -38,6 +39,8 @@ type ExtendedHeader struct {
 	Commit       *core.Commit               `json:"commit"`
 	ValidatorSet *core.ValidatorSet         `json:"validator_set"`
 	DAH          *da.DataAvailabilityHeader `json:"dah"`
+
+	hash []byte
 }
 
 // MakeExtendedHeader assembles new ExtendedHeader.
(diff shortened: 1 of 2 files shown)


header/header.go
```diff
@@ -94,7 +97,22 @@ func (eh *ExtendedHeader) Time() time.Time {
 // NOTE: It purposely overrides Hash method of RawHeader to get it directly from Commit without
 // recomputing.
```
cristaloleg, line 98: Well, practically we're recomputing it but on the other side we do this when it's not present. Any ideas for the best possible doc-comment here? (or just drop the part about recomputing, lol)

Wondertan: And added a fresh one

header/header.go
```diff
@@ -94,7 +97,22 @@ func (eh *ExtendedHeader) Time() time.Time {
...
 func (eh *ExtendedHeader) Hash() libhead.Hash {
-	return eh.Commit.BlockID.Hash.Bytes()
+	if len(eh.hash) == 0 {
+		bs := make([][]byte, len(eh.Commit.Signatures))
+		for i, commitSig := range eh.Commit.Signatures {
```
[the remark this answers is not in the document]
Wondertan: Commit.Hash != BlockID.Hash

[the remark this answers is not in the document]
Wondertan: So we should do `eh.RawHeader.Hash()` + hashing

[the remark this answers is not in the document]
cristaloleg: I blindly assumed that you approved previous PR, thankfully tests confirm that.

header/header.go
```diff
@@ -38,6 +39,8 @@ type ExtendedHeader struct {
 	Commit       *core.Commit               `json:"commit"`
 	ValidatorSet *core.ValidatorSet         `json:"validator_set"`
 	DAH          *da.DataAvailabilityHeader `json:"dah"`
+
+	hash []byte
```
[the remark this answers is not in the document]
cristaloleg: Fixed with `sync.Once.`

header/header.go
cristaloleg: Done

header/header.go
```diff
@@ -223,7 +229,7 @@ func (eh *ExtendedHeader) UnmarshalBinary(data []byte) error {
 		return err
 	}
 
-	*eh = *out
+	*eh = *out.Clone()
```
Wondertan, line 232: Why is this necessary?

cristaloleg: 'cause now we've `sync.Once` that cannot (should not) be copied. We've a few places where we do assignment which triggers `go vet`.

So, additional helper method to prevent such things. See failure: https://github.com/celestiaorg/celestia-node/actions/runs/12118920400/job/33784451925?pr=3975#step:4:34

Wondertan: Can we find a way to solve this without introducing additional copy?

cristaloleg: Well, ok, done.

header/header.go
```diff
@@ -223,7 +229,12 @@ func (eh *ExtendedHeader) UnmarshalBinary(data []byte) error {
 		return err
 	}
 
-	*eh = *out
+	*eh = ExtendedHeader{
+		RawHeader:    out.RawHeader,
+		Commit:       out.Commit,
+		ValidatorSet: out.ValidatorSet,
+		DAH:          out.DAH,
+	}
```
Wondertan, line 237: Isn't this an equivalent of what it was before, but without a method?

Wondertan (commented):
I carefully reviewed this again and realized that we actually don't need this change.

[Here](https://github.com/celestiaorg/celestia-node/blob/d726f1131c89536d967f754eb48b49e1cd0aa2f6/header/header.go#L150) we check that the result of local hashing is equal to what we have in the commit already. So when we return from commit we know its something we computed ourselves. Therefore, we keep returning hash from the commit and there is no need to add recompute value and synchronization around it.

header/header.go
Wondertan: (not relevant)

cristaloleg: Yes, it's. We did shallow copy before.

cristaloleg:
Closing as not needed (see Hlib's review comment)

PR #3979, opened 2024-12-02
fix(ci): drop outdated param

cristaloleg:
Leftover after #3963

### DIFF
--- a/.github/workflows/ci_release.yml
+++ b/.github/workflows/ci_release.yml
@@ -11,10 +11,6 @@ on:
     types: [published]
   pull_request:
   workflow_dispatch:
-    inputs:
-      tag-as:
-        description: 'Tag for snapshot release (optional)'
-        required: false
 
 jobs:
   # set up go version for use through pipelines, setting


.github/workflows/ci_release.yml
```diff
@@ -104,13 +108,27 @@ jobs:
...
       - uses: goreleaser/goreleaser-action@v6
+        if: github.event_name == 'release'
+        with:
+          distribution: goreleaser
+          version: latest
+          args: release --clean
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GPG_FINGERPRINT: ${{ steps.import_gpg.outputs.fingerprint }}
+          GORELEASER_CURRENT_TAG: ${{ github.event.release.tag_name }}
```
cristaloleg, line 119: Take the tag from release name. This should solve the problem with bad automatic tag fetch (due to `git` limited knowledge and `goreleaser` as a result).

.github/workflows/ci_release.yml
```diff
@@ -111,19 +123,35 @@ jobs:
...
+        if: github.event_name == 'release'
+        run: echo "LATEST_TAG=${{ github.event.release.tag_name }}" >> $GITHUB_ENV
+
+      - name: Set LATEST_TAG for workflow_dispatch
+        if: github.event_name == 'workflow_dispatch'
+        run: echo "LATEST_TAG=${{ github.event.inputs.run-on-tag }}" >> $GITHUB_ENV
+
+      - name: Checkout a given tag
+        if: github.event_name == 'workflow_dispatch'
+        run: git checkout ${{ inputs.run-on-tag }}
```
walldiss, line 151: Should checkout to LATEST_TAG instead?

cristaloleg: Doesn't matter 'cause it's the same as in `workflow_dispatch`.

walldiss (approved):

PR #3980, opened 2024-12-04
feat(nonce): Update core_access.go adjust network nonce

InoMurko:
Continuation of the discussion here: https://github.com/celestiaorg/celestia-app/pull/4067

This PR adds the ability for self-adjusting flow when clients submit blobs and the message returned yells that the nonce is out of sync.

Caldera has been using this in mocha-4 and celestia network for months now and it would be beneficial for both to close the gaps on changes we've implemented.

### DIFF
--- a/state/core_access.go
+++ b/state/core_access.go
@@ -6,6 +6,9 @@ import (
 	"fmt"
 	"sync"
 	"time"
+	"strings"
+	"regexp"
+	"strconv"
 
 	nodeservice "github.com/cosmos/cosmos-sdk/client/grpc/node"
 	"github.com/cosmos/cosmos-sdk/client/grpc/tmservice"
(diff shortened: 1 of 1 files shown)


Wondertan:
Shouldn't this be the responsibility of the Signer? We could add this as a hotfix if its a pressing issue, but the node is not responsible for nounce management and error recovery of tx sending. The Signer component is.

cc @cmwaters

InoMurko:
> Shouldn't this be the responsibility of the Signer? We could add this as a hotfix if its a pressing issue, but the node is not responsible for nounce management and error recovery of tx sending. The Signer component is.
> 
> 
> 
> cc @cmwaters

Where does the Signer component live in the architecture diagram? Let's imagine https://github.com/celestiaorg/nitro-das-celestia and Nitro deployment. Cc @ferret-san

Wondertan:
@InoMurko, the [Signer](https://github.com/celestiaorg/celestia-app/blob/e90888b20fbf78f6ca47ecffc483b917586ab92c/pkg/user/signer.go#L23) is a component sitting on the app side. It has smarts around nounce and error handling during tx submission, and it seems to have a bug.

Ferret-san:
> Where does the Signer component live in the architecture diagram? Let's imagine https://github.com/celestiaorg/nitro-das-celestia and Nitro deployment. Cc @Ferret-san

`nitro-das-celestia` is just a proxy between the Nitro codebase and a celestia node, so it will use whatever key you setup the celestia node to use.

walldiss:
Closing this because stale. Please open an issue if there is still a problem.

PR #3985, opened 2024-12-06
Feature branch blob proof

vgonkivs:
Merge Blob Proof feature branch to api/breaks


rach-id (dismissed):
🎉 🎉 
Great working fixing all the conflicts 🙏  Thanks a lot for helping with this

blob/blob.go
```diff
@@ -6,54 +6,162 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/tendermint/tendermint/crypto/merkle"
```
cristaloleg, line 9: Looks like it should be `"github.com/celestiaorg/go-square/merkle"` 😉

blob/blob.go
```diff
@@ -6,54 +6,162 @@ import (
...
+type Commitment []byte
 
-// equal is a temporary method that compares two proofs.
-// should be removed in BlobService V1.
-func (p Proof) equal(input Proof) error {
-	if p.Len() != input.Len() {
-		return ErrInvalidProof
+// Verify takes a data root and verifies if the
+// provided proof's subtree roots were committed to the given data root.
+func (p *Proof) Verify(dataRoot []byte) (bool, error) {
```
cristaloleg, line 44: Do we need a `bool` return param at all? Looks like it's only `true` when error is `nil`.

blob/blob.go
```diff
@@ -6,54 +6,162 @@ import (
...
+}
+
+// GenerateCommitment generates the share commitment corresponding
+// to the proof's subtree roots
+func (p *Proof) GenerateCommitment() Commitment {
+	return merkle.HashFromByteSlices(p.SubtreeRoots)
+}
+
+func (com Commitment) String() string {
+	return string(com)
```
cristaloleg, line 159: Should it be hex or base64 ?

blob/blob.go
cristaloleg: Same for other files.

cristaloleg: Hm, I see that in `go.mod` `github.com/celestiaorg/go-square/merkle` got removed. Also, in `go-square/v2` there is no `merkle` package anymore. So, I might be wrong about bad import.

blob/service_test.go
```diff
@@ -294,26 +290,41 @@ func TestBlobService_Get(t *testing.T) {
 				require.ErrorIs(t, err, ErrInvalidProof)
 				included, ok := res.(bool)
 				require.True(t, ok)
-				require.True(t, included)
+				require.False(t, included)
```
cristaloleg, line 293: Why it's inverted now?

rach-id:
btw, given that we have this PR, can we close this one https://github.com/celestiaorg/celestia-node/pull/3796?

vgonkivs:
> btw, given that we have this PR, can we close this one https://github.com/celestiaorg/celestia-node/pull/3796?

Yeah I think so

blob/service_test.go
vgonkivs: previously bool flag was incorrectly handled. Now it works correct

blob/blob.go
vgonkivs: You are right. Double-checked and it seems we don't have a case when we have to return` false, nil` or `true,error`

PR #3996, opened 2024-12-11
feat(observability): add start parameter to provide basic auth to pyr…

smuu:
…oscope

### DIFF
--- a/nodebuilder/settings.go
+++ b/nodebuilder/settings.go
@@ -49,10 +49,10 @@ func WithBootstrappers(peers p2p.Bootstrappers) fx.Option {
 }
 
 // WithPyroscope enables pyroscope profiling for the node.
-func WithPyroscope(endpoint string, nodeType node.Type) fx.Option {
+func WithPyroscope(endpoint string, basicAuthUser string, basicAuthPassword string, nodeType node.Type) fx.Option {
 	return fx.Options(
 		fx.Invoke(func(peerID peer.ID) error {
-			_, err := pyroscope.Start(pyroscope.Config{
+			config := pyroscope.Config{
 				UploadRate:      15 * time.Second,
 				ApplicationName: "celestia.da-node",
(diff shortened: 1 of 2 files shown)


nodebuilder/settings.go
```diff
@@ -69,7 +69,14 @@ func WithPyroscope(endpoint string, nodeType node.Type) fx.Option {
 					pyroscope.ProfileInuseSpace,
 					pyroscope.ProfileGoroutines,
 				},
-			})
+			}
+
+			if basicAuthUser != "" && basicAuthPassword != "" {
```
cristaloleg, line 74: Nit: not sure do we need this? By default `config.BasicAuthUser` and `config.BasicAuthPassword` are already `""`. Looks like we can do this without `if`.

smuu: Makes sense, changed that 👍

Wondertan (approved):

renaynay (commented):
this needs a nolint directive for the flags_misc file and also is it a problem if empty strings are passed for both user + password ?

smuu:
> this needs a nolint directive for the flags_misc file and also is it a problem if empty strings are passed for both user + password ?

- fixed the lint issue
- if the pyroscope user/password is empty, authentication is skipped

Wondertan (approved):

nodebuilder/settings.go
```diff
@@ -69,7 +69,11 @@ func WithPyroscope(endpoint string, nodeType node.Type) fx.Option {
 					pyroscope.ProfileInuseSpace,
 					pyroscope.ProfileGoroutines,
 				},
-			})
+				BasicAuthUser:     basicAuthUser,
+				BasicAuthPassword: basicAuthPassword,
```
renaynay, line 73: will it be a problem if password is set but not user? / and vice versa?

PR #4006, opened 2024-12-12
chore: use errors.New to replace fmt.Errorf with no parameters

RiceChuan:
use errors.New to replace fmt.Errorf with no parameters


Wondertan:
These are basically equivalent, and moving to `errors.New` prevents us from adding new fields in the future.

PR #4028, opened 2025-01-05
feat(share/availability/full): Introduce Q4 trimming for archival nodes

renaynay:
This PR introduces trimming of `.q4` files outside the sampling window for archival nodes.

It also updates the pruner checkpoint such that it's possible to convert from an archival node --> full pruning node but not the other way around (the node runner would need to resync the node in this case). 

All deletion logic now lives in either `share/availability/light` or `share/availability/full` but the pruner service is still contained in a separate package and now runs by default as all node types will have some form of pruning happening.


pruner/pruner.go
```diff
@@ -10,4 +10,5 @@ import (
 // from the node's datastore.
 type Pruner interface {
 	Prune(context.Context, *header.ExtendedHeader) error
+	Kind() string
```
renaynay, line 13: Okay so i hate this

we can implicitly do this by checking for the presence of a historical block (without its .q4 file) but I think that could be much more invasive and annoying to sort out.

renaynay: I think this is the cleanest option as of right now.

pruner/checkpoint.go
```diff
@@ -27,21 +27,44 @@ var (
...
+		PrunerKind:       prunerKind,
+		LastPrunedHeight: 1,
+		FailedHeaders:    map[uint64]struct{}{},
 	}
+}
+
+// DetectPreviousRun ensures that a node that has been run with "full" pruning
+// mode previously cannot revert back to an "archival" one. This check should
+// only be performed when a node is either a Full or Bridge node.
+func DetectPreviousRun(ctx context.Context, ds datastore.Datastore, expectedKind string) error {
```
cristaloleg, line 46: Technically it's not only "detect" but also "save". Don't know what will be a better name but at least this should be mentioned in a doc-comment.

pruner/pruner.go
walldiss: Meh, I don't like the leaking abstraction. Did you consider persisting implementation identifier inside Pruner implementation? Perhaps it can be part to implementation to detect unwanted migration

renaynay:
There's currently an annoying edge-case to the solution presented in https://github.com/celestiaorg/celestia-node/pull/4028/commits/6087e647cc861e19e4a06d7490516a7e9bd4880a that wouldn't be able to detect a node that has been previously run with pruner enabled (on older version where archival trimming was not present), and once upgraded to new version, runs without pruning enabled successfully even though it should not be allowed to revert to an archival run. 

I need to figure out how to prevent this without making a super ugly solution.

renaynay:
The solution to the above comment ^ is [f370346](https://github.com/celestiaorg/celestia-node/pull/4028/commits/f370346da2dc910e36a3c5ef160ae8e34cd7b393)

walldiss:
Looks good. It took me a while to understand migration. I know it is temporary code, so not insisting to change. Perhaps it might be simpler if migration was detected inside nodebuilder? This way It will be easier to remove later, wihtout the need for refactor. Smth like this:
```
	// nodebuilder/share
	migration(isarhival bool){
		// Should check internaly prunned <-> archival migration of new endpoints
		created, err := avail.CreateCheckpoint(isarhival)
		
		// temporary method, can be removed later with migration() func
		wasPrunned := pruner.HasCheckpoint()
		
		if isarhival && wasPrunned && !created{
			// unwanted migration detected
		}
	}
```

Another thing to consider is lack of archival -> pruned conversion support in the PR. We should reset prunner checkpoint to allow it to Remove Q1 files.

cristaloleg (approved):
SGTM

renaynay:
@walldiss 

re https://github.com/celestiaorg/celestia-node/pull/4028#issuecomment-2595333248

I can move the logic to nodebuilder, yes.

share/availability/full/utils.go
```diff
@@ -0,0 +1,46 @@
...
+	storePrefix     = datastore.NewKey("full_avail")
+	previousModeKey = datastore.NewKey("previous_run")
+	pruned          = []byte("pruned")
+	archival        = []byte("archival")
+)
+
+// ConvertFromArchivalToPruned ensures that a node has not been run with pruning enabled before
+// cannot revert to archival mode. It returns true only if the node is converting to
+// pruned mode for the first time.
+func ConvertFromArchivalToPruned(ctx context.Context, ds datastore.Datastore, isArchival bool) (bool, error) {
```
walldiss, line 22: (nit) Would it be better to name like updatePruneMode, since it handles both archival -> pruned and pruned -> arhival cases now?

walldiss (dismissed):
Thank you for dedication for making it better! Happy that we have temporary logic extracted to migration_utils.go. It has its flaws of reimplementing logic for storing previous run, but I don't think it is a big deal, since it will be remove relatively soon. 

Added Few optional nits, otherwise Looks good to me!

share/availability/full/utils.go
renaynay: I think `Convert` is more descriptive here, so I will leave it

walldiss (approved):

vgonkivs (dismissed):

walldiss (approved):

vgonkivs (approved):

cristaloleg (approved):

PR #4041, opened 2025-01-13
(Discussion): canonical api draft

walldiss:
New gRPC API Design Document and Proto Definitions

This PR introduces a comprehensive design document (readme.md) and proto definitions for the new gRPC-based API for Celestia Node. The design focuses on improving security, simplicity, and developer experience while enabling smooth migration from the current JSON-RPC implementation.

Key Features:
- Unified status codes and error handling across all services
- Enhanced data availability verification
- Support for both row root (V0) and data root (V1) verification approaches
- Simplified blob submission and management
- Clear separation between different service domains (blobs, shares, proofs)

Looking for feedback on:
- Overall API structure and design decisions
- Security considerations
- Migration approach (V0/V1)
- Error handling strategy
- Any missing functionality

TODO:
[ ] Header service
[ ] http annotation for http gateway support

Note: The proto definitions and design doc should be reviewed together as they complement each other.


api/V1/readme.md
```diff
@@ -0,0 +1,279 @@
...
+- Rich error information
+- Easy to extend
+- Supports both technical and user-friendly messages
+
+Some error codes already listed in the draft, however the list might be not full and would need to be revisited with each particular usecase in mind
+
+Standardized Status Codes
+```protobuf
+// Common status codes used across all services
+enum StatusCode {
```
liamsi, line 106: I do think we should have the possibility for a dev to know if an error stems from node or e.g. somewhere from app. Maybe that is already possible with the list somehow (e.g. not by the code but via the error string or sth?).

api/V1/readme.md
```diff
@@ -0,0 +1,279 @@
...
+
+1. **Enhanced ShareAvailable Verification**:
+   The ShareAvailable call will verify the continuous chain of blocks from the oldest block in the availability window up to the requested height. The call will return false if any block within this range is unavailable.
+
+2. **Integrated Availability Verification**:
+   A new `verify_available` option will be added to all data fetching methods (blobs, shares, etc). When enabled:
+  - The API will verify availability for both the requested block and all preceding blocks within the sampling window
+  - Data will only be returned if the entire chain segment is verified as available
+  - Failed availability checks will return a specific `data_not_available` error code
+
```
liamsi, line 76: Could this slow anything down from an app perspective?

api/V1/common.proto
```diff
@@ -0,0 +1,41 @@
...
+enum StatusCode {
+  // Reserve 0 for unspecified
+  STATUS_UNSPECIFIED = 0;
+  // Success codes (1-99)
+  STATUS_OK = 1;
+  // Error codes (100+)
+  // ERROR_DATA_NOT_AVAILABLE is returned if data availability sampling gurantees was required for request
+  // and failed. Keep in mind that data availability sampling will be ensured for all blocks in
+  // sampling window till the requested height
+  ERROR_DATA_NOT_AVAILABLE = 100;
```
liamsi, line 17: Will the caller be able to differentiate between e.g. sampling timing out, and other cases?

api/V1/readme.md
```diff
@@ -0,0 +1,279 @@
...
+### ShareService
+Has support for requesting all common sets of shares:
+```protobuf
+// ShareService - handles data share operations
+service ShareService {
+  // Core share operations
+  rpc GetShare(GetShareRequest) returns (GetShareResponse);
+  rpc GetEDS(GetEDSRequest) returns (GetEDSResponse);
+
+  // Range operations - support both V0 (row roots) and V1 (data root) responses
```
liamsi, line 178: Both responses below return `V0` though?

liamsi: Ah, in the protos I see there is also `GetRangeResponseV1` etc. Maybe mention that we only list v0 here to keep the readme conscise.

api/V1/blob_service.proto
```diff
@@ -0,0 +1,100 @@
...
+  bytes namespace = 1;
+  repeated bytes data = 2;
+  SubmitOptions options = 3;
+}
+
+message SubmitDataResponse {
+  celestia.node.v1.common.Status status = 1;
+  uint64 height = 2;
+  repeated bytes commitments = 3;
+}
```
vgonkivs, line 63: Let's include the tx hash here. iirc, some teams are using SubmitPFB in the state module only bc it additionally provides a tx hash.

api/V1/blob_service.proto
```diff
@@ -0,0 +1,100 @@
...
+  celestia.node.v1.common.Status status = 1;
+  repeated celestia.node.v1.types.BlobWithProofV0 blobs = 2;
+}
+
+message GetBlobsResponseV1 {
+  celestia.node.v1.common.Status status = 1;
+  repeated celestia.node.v1.types.BlobWithProofV1 blobs = 2;
+}
+
+message SubmitDataRequest {
```
vgonkivs, line 53: What about submitting data under multiple namespaces?

I'd suggest to add one more method:

```proto
message SubmitBlobsRequest {
  repeated Blobs blobs = 1; // blobs from different namespaces can be set
 SubmitOptions options = 2;
}
```

api/V1/blob_service.proto
```diff
@@ -0,0 +1,100 @@
...
+  uint64 height = 2;
+  repeated celestia.node.v1.types.BlobWithProofV1 blobs = 3;
+}
+
+message BlobsSubscribeRequest {
+  bytes namespaces = 1;
+  uint64 from_height = 2;
+  repeated bytes commitment = 3;
+  optional celestia.node.v1.common.Options options = 4;
+}
```
vgonkivs, line 82: Such a method could be useful in case we support async blob submission. but iirc, it's not feasible rn as error handling in this case might be tricky.

api/V1/types.proto
```diff
@@ -0,0 +1,43 @@
...
+  uint32 start = 1;
+  uint32 end = 2;
+}
+
+message Blob {
+  bytes data = 1;
+  optional bytes signer = 2;
+  bytes commitment = 3;
+}
+
```
vgonkivs, line 25: Let's define BlobProof([reference](https://github.com/celestiaorg/celestia-node/pull/3985/files#diff-e0265c8a8ef591c5dac3ccd849803c7c79932ed3b87bff256fdf9f291611ace4R21-R31) here)

api/V1/blob_service.proto
```diff
@@ -0,0 +1,100 @@
...
+message BlobsSubscribeResponse {
+  bytes commitment = 1;
+  celestia.node.v1.types.BlobWithProofV1 blob = 2;
+  uint64 height = 3;
+}
+
+// BlobService - handles blob operations
+service BlobService {
+  // Core operations
+  rpc GetBlob(GetBlobRequest) returns (GetBlobResponseV0);
```
vgonkivs, line 93: what if the submitted blob is v1,v2...? Maybe it's better to have a generic BlobResponse and extend the blob's message with additional `optional` fields?

api/V1/types.proto
```diff
@@ -0,0 +1,43 @@
...
+message SquareRange {
+  uint32 start = 1;
+  uint32 end = 2;
+}
+
+message Blob {
+  bytes data = 1;
+  optional bytes signer = 2;
+  bytes commitment = 3;
+}
```
vgonkivs, line 24: I've got your point the Blob is part of the namespace, and by submitting/getting the blob, the user already has it but I'd anyway make it part of the message for consistency.

api/V1/proofs.proto
```diff
@@ -0,0 +1,39 @@
...
+  repeated bytes nodes = 3;  // NMT proof nodes (48 bytes each: namespace bounds + hash)
+  bytes leaf_hash = 4;       // Used for non-inclusion proofs (40 bytes: namespace + hash)
+}
+
+message DataRootProof {
+  ProofType type = 1;
+  Proof proof = 2;
+}
+
+message RowRootProof {
```
vgonkivs, line 36: As we discussed with @walldiss, he does not want to import data structures from different repo. So, let's copy past it from the tendermint repo

ramin (commented):
i hope my comments are welcome. I've added some inline thoughts both from perspective as a keen observer and API user, high level it feels like this first pass is a literal "proto-fication" of the existing APIs vs a clean re-design considering dev ergonomics. 

As a user:

- i'd appreciate some aggressive design iterations on naming, there's lots of name stuttering inside files, services, messages
- i think we could consider modularizing more, adding more files (ie: blob/v1 etc) so you could compose / import / use these protos as a developer building against node piecemeal, will then make gradual expansion easier
- some of the proto's defined as `*_service.proto` should (i think) be just `<name>.proto`, they can then for example, define a Service, thus reading cleaner, like  `shares.Service`, vs `share_service.ShareService`. Then, various types could be included there, and even nested in sub folder protos
- i think another sweep or two of usability considerations should be made, that support end user/developer usage (ie: namespace being a messagehelpers etc) being supported in the proto messages to be more complex types, we could also consider `protoc-gen-validate` to define length and validation of these types etc in the canonical API, vs "needing to know" to submit a good request

that's all for now, thank you 😸

api/V1/types.proto
```diff
@@ -0,0 +1,43 @@
...
+  celestia.node.v1.proof.DataRootProof proof = 2;
+  uint64 index = 3;  // index of the first share of the blob in the EDS
+}
+
+// SharesWithProof is used in V0 verification approach with row roots
+message SharesWithProof {
+  // Shares should be from the same row
+  repeated celestia.node.v1.types.Share shares = 1;
+  celestia.node.v1.proof.RowRootProof row_proof = 2;      // NMT proof to the row root
+}
```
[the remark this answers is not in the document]
vgonkivs: I think it makes sense to move definitions to the places where they are used.

api/V1/fraud_service.proto
```diff
@@ -0,0 +1,46 @@
...
+  BadEncodingFraudProof fraud_proofs = 1;
+}
+
+message VerificationResponse {
+  bool valid = 1;
+  string error = 2;
+}
+
+// BadEncodingFraudProofService - handles operations related to Bad Encoding Fraud Proofs
+service BadEncodingFraudProofService {
```
cmwaters, line 37: What's the plan around supporting BEFPs? I've heard various things like we're not supporting them at the moment but perhaps plan to.

api/V1/common.proto
cmwaters, line 17: Is this only returned by the `SharesAvailable` method?

api/V1/common.proto
```diff
@@ -0,0 +1,41 @@
...
+  // Fraud was detected in the chain before requested block
+  ERROR_FRAUD_DETECTED = 101;
```
cmwaters, line 19: What would cause this error? (i.e. more specifically)

api/V1/common.proto
```diff
@@ -0,0 +1,41 @@
...
+  ERROR_TIMEOUT = 102;
```
cmwaters, line 20: Are all timeouts handed by gRPC itself i.e. contexts in the case of go. Do they need to be included here?

api/V1/share_service.proto
```diff
@@ -0,0 +1,120 @@
...
+}
+
+message SharesAvailableResponse {
+  celestia.node.v1.common.Status status = 1;
+  bool is_available = 2;
+  uint64 highest_available_height = 3;
+}
+
+// ShareService - handles data share operations
+service ShareService {
```
cmwaters, line 106: We have subscription options for both blob and fraud servies, why not share services?

I can imagine most base rollups (or rollup full nodes in general) will subscribe to all data in one or more namespaces.

api/V1/share_service.proto
```diff
@@ -0,0 +1,120 @@
...
+import "common.proto";
+import "types.proto";
+import "proof.proto";
+
+message GetShareRequest {
+  uint64 height = 1;
+  uint32 row = 2;
+  uint32 col = 3;
+  optional celestia.node.v1.common.Options options = 4;
+}
```
cmwaters, line 16: Will you also be able to reference a share based on an index (rather than rol and col). Some of the responses return the starting index of a blob.

cmwaters (commented):
Great work. Just tried to do a first pass of everything.

A more general topic regarding the proof API. What do you think about prioritising the data root based proving system first before releasing this API?

api/V1/blob_service.proto
```diff
@@ -0,0 +1,100 @@
...
+option go_package = "github.com/celestiaorg/celestia-node/api/v1/blob";
+
+import "common.proto";
+import "types.proto";
+import "proof.proto";
+
+message SubmitOptions {
+  double gas_price = 1;    // Amount to be paid per gas unit
+  uint64 gas = 2;         // Gas amount
+  string key_name = 3;    // Keystore key for signing
```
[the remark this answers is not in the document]
adlerjohn: https://protobuf.dev/programming-guides/proto3/#field-labels

It looks like they're not exactly `optional` by default if they're not messages. It might be good to explicitly use the keyword `optional` in all places.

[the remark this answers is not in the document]
sezna: I did some digging and it looks like, from what I can tell, `optional` will tell the protoc generators to generate `has_` methods, which check for presence versus just using default values. They accomplish this by tracking a "set" versus "unset" property in addition to the value. I _think_ this is actually an extra bit that is being used, and not just a "is this equal to the default value?" method. So, yes, it seems like `optional` is better in most non-primitive-type fields unless there's a good reason to defaults

adlerjohn (commented):
The current draft for the canonical API couples several APIs into a single API, which should be separate components:
1. Light node API (i.e. availability).
2. Full node API (i.e. retrievability).

The availability API is what light nodes expose, and should only contain the methods that are useful if the user is running a local light node. Since there's no point in using a hosted light node. For example, the status of the light node's sampling (i.e. is a block available or not).

The retrievability API is what full nodes expose, and contains all other methods. This API will be useful regardless of if the user is running their own full node or using a hosted full node.

Note that I'm not distinguishing here between a consensus full node and DA full node. This is because that the spec should not be prescriptive on where each API lives. This also means this should not be the celestia-node canonical API, but rather the Celestia-specific canonical API. And it should be namespaced in the protobuf as such.

E.g. rather than `celestia.node...` it should be `celestia.availability...` and `celestia.retrievability...`

Wondertan:
I would like to point that light node should still expose the retrievability API, s.t. LN runners can retrieve data over p2p network with discovery.

This is what celestia-node currently provides and this should stay so.

Wondertan:
The current API is _intentionally_ designed to be consistent across the node types, so no matter which node you run you have the same functionality. 

This is a **local-first** design build around caching locality and layering. If you run a FN - it simply means you cache everything, LN - nothing, PN - certain namespaces.

sezna:
> LN runners can retrieve data over p2p network with discovery.

> This is what celestia-node currently provides and this should stay so.

Just for my own clarity -- what is the use case of a LN providing a full data retrieval API if it just has to query other nodes over p2p anyway?

walldiss:
> The availability API is what light nodes expose, and should only contain the methods that are useful if the user is running a local light node. Since there's no point in using a hosted light node. For example, the status of the light node's sampling (i.e. is a block available or not). 
The retrievability API is what full nodes expose, and contains all other methods. This API will be useful regardless of if the user is running their own full node or using a hosted full node.

I agree with @Wondertan and think all nodes (light DA, full DA) should axpose both. All data can be obtained from DA network and Light nodes can utilise efficient protocol for obtaining data, which is already there. 

Consensus nodes can expose it later. I don't think it is improtant for now, because the way they expose it is by running Bridge node, which is kinda a proxy with node API. 

> Note that I'm not distinguishing here between a consensus full node and DA full node. This is because that the spec should not be prescriptive on where each API lives. This also means this should not be the celestia-node canonical API, but rather the Celestia-specific canonical API. And it should be namespaced in the protobuf as such.
E.g. rather than celestia.node... it should be celestia.availability... and celestia.retrievability...

Agree, the APi will not be part of node repo and we would need to update protos accordingly.

walldiss:
> Just for my own clarity -- what is the use case of a LN providing a full data retrieval API if it just has to query other nodes over p2p anyway?

LN is smart when quering p2p network and just easier to most of the users. It has peers and content discovery, which makes it more efficient and easier to fetch the data.

Wondertan:
To add to @walldiss said and to address future potential @adlerjohn's concern, the p2p network doesn't necessary imply altruistic data sources for recent and archival data. LNs can still be configured to retrieve data using external payed data sources for both. 

The downstream effects of it are:
* Data retrieval API is standard for Celestia, with an ecosystem of backends that need follow the standard to integrate with us. Users are happy to work with a single evolving API, instead of zoo of API per each retrievability service.
* The canonical way of retrieving data requires running any Celestia node, with LN at minimum. Assuming running LNs becomes more practical over time, we end up maximizing availability.

Wondertan:
> Also, what does local first mean in this context -- does it mean that the mental model around the node types prioritizes only talking to a locally run node?

It means in this context that all the operation you do with Celestia are going through you personal tiny cute node, which you can trust. It takes care of all the verification necessary, signing, caching and finding source for the data you need on demand. Of course with customization to enable alternative behaviors if requested.

walldiss:
Migrating this PR to new repo https://github.com/celestiaorg/canonical-api/pull/1 Pls continue the discussion there.

api/V1/share_service.proto
adlerjohn: The blob subscription allows you to subscribe to a namespace too, so it handles that case.

api/V1/blob_service.proto
```diff
@@ -0,0 +1,100 @@
...
+
+message SubmitDataResponse {
+  celestia.node.v1.common.Status status = 1;
+  uint64 height = 2;
+  repeated bytes commitments = 3;
+}
+
+message SubscribeAllRequest {
+  bytes namespaces = 1;
+  uint64 from_height = 2;
```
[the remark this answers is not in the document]
adlerjohn: Fixed in https://github.com/celestiaorg/canonical-api/pull/2/commits/fe5aa19a52ff7c3a3baf6fcd25f0b4a1e9768b34

api/V1/proof_service.proto
```diff
@@ -0,0 +1,69 @@
...
+  celestia.node.v1.common.Status status = 1;
+  bool valid = 2;
+}
+
+// ProofsService - verification service that supports both V0 and V1 approaches
+service ProofsService {
+  // V0 verification of inclusion of shares. Requires row root to perform verification.
+  // Does not verifies namespace rules of provided shares and should be used only
+  // to verify inclusion of data spread over multiple namespaces
+  rpc VerifySharesV0(VerifySharesRequestV0) returns (VerifyProofResponse);
```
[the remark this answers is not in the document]
adlerjohn: These do in fact verify the proof, they don't get it. Removed this feature in https://github.com/celestiaorg/canonical-api/pull/2

api/V1/fraud_service.proto
adlerjohn: BEFPs are removed and there's a good chance they'll never be supported again before we get ZK proofs.

api/V1/common.proto
adlerjohn: Removed in https://github.com/celestiaorg/canonical-api/pull/2

PR #4046, opened 2025-01-15
refactor: move striplock to utils package

crStiv:
This commit moves the striplock implementation to the utils package to make it
more reusable across the codebase. Changes include:
- Created new libs/utils/striplock.go with improved implementation
- Added comprehensive documentation
- Left deprecated type aliases in store package for backward compatibility

### DIFF
--- a/libs/utils/striplock.go
+++ b/libs/utils/striplock.go
@@ -1,0 +1,64 @@
+package utils
+
+import (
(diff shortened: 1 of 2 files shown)


cristaloleg:
Thanks for the PR. However, we are not planning such refactoring in the near future.

PR #4048, opened 2025-01-17
docs: adr-13 - automatic gas price estimation

cmwaters:
First pass of the gas price estimation ADR. Comments are welcome!

### DIFF
--- a/docs/adr/adr-013-automatic-gas-price-estimation.md
+++ b/docs/adr/adr-013-automatic-gas-price-estimation.md
@@ -59,4 +59,4 @@ There has been some discussion of a pricing mechanism built in to the protocol. 
 
 ## References
 
-- [CIP 18](https://github.com/celestiaorg/CIPs/blob/main/cips/cip-18.md) 
+- [CIP 18](https://github.com/celestiaorg/CIPs/blob/main/cips/cip-18.md)


docs/adr/adr-013-automatic-gas-price-estimation.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Consequences
+
+### Positive
+
+- Light nodes should 
+
+### Negative
+
+- Extra round trip with either the consensus node or someone offering the `GasEstimator` service before the transaction is submitted
```
vgonkivs, line 59: Would it make sense to subscribe to the `GasEstimator` and store the latest value?

cmwaters: Hmm it's an option. But most light clients running will seldom be sending transactions, so imagine if a server had to constantly send updates of the gas price to these light clients which were only querying the state of their rollup/s and not actually submitting any transactions

docs/adr/adr-013-automatic-gas-price-estimation.md
```diff
@@ -0,0 +1,65 @@
...
+
+Depending on the message types submitted, the node will decide whether to also query for the amount of gas that is required to submit the transaction. In the case of the following messages, the gas can be reliably estimated using a simple linear model:
+
+- `MsgPayForBlob`
+
+For all other messages (or combination of messages), the node will call the `GasEstimator` service.
+
+### Setting guardrails
+
+A new field in the `TxConfig` will be introduced allowing for a user to  use the gas price estimation but setting a maximum gas price as a cap that the user is willing to pay. If the price returned by the estimator is greater than the cap, likely due to a price spike, an error will be returned informing the user that the gas price required is currently greater than the user is willing to pay
```
renaynay, line 37: What will be the default cap in case max isn't specified?

docs/adr/adr-013-automatic-gas-price-estimation.md
```diff
@@ -0,0 +1,65 @@
...
+
+### Defaults
+
+By default, the same consensus node that the node is connected with, should be used as the `GasEstimator`. During construction, a user can optionally provide a grpc address that can be used instead.
```
renaynay, line 41: Unclear what you mean by "a grpc address that can be used instead" 

Node has one entrypoint for tx submission and it's through a core node's gRPC. So you're saying the GasEstimator will also be available through exactly the same gRPC conn as other state-related queries we perform on the core node?

docs/adr/adr-013-automatic-gas-price-estimation.md
```diff
@@ -0,0 +1,65 @@
...
+
+### Fallback
+
+If there are any errors with interacting with the `GasEstimator` service, the node should log an error and fallback to using the minimum gas price.
```
renaynay, line 45: Should you propagate this error back to the user somehow? Otherwise caller will have no idea why their txs might be taking relatively long.

docs/adr/adr-013-automatic-gas-price-estimation.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Alternative Approaches
+
+There has been some discussion of a pricing mechanism built in to the protocol. This would be more complex but would remove the reliance on a trusted third party. This may be explored in the future but given resource constraints, the `GasEstimator` service has been prioritized
+
+## Consequences
+
+### Positive
+
+- Light nodes should
```
renaynay, line 55: should what

docs/adr/adr-013-automatic-gas-price-estimation.md
renaynay: @cmwaters the core accessor (state service) in node is not only used by LNs but also FNs and BNs, so anyone using node to submit txs and has specified the default to use GasEstimator will have to go through this roundtrip. So maybe LNs will be seldom for submission, but FNs + BNs will be frequent.

Something to consider.

docs/adr/adr-013-automatic-gas-price-estimation.md
Wondertan: My understanding is that we should provide new config field/flag for this service and try core GRPC if not specified.

docs/adr/adr-013-automatic-gas-price-estimation.md
Wondertan: Are there negative financial consequences for doing this fallback? Is it possible that by looping in fallback we spend much more gas then expected? If that's possible, I think we should error back to user.

docs/adr/adr-013-automatic-gas-price-estimation.md
Wondertan: If its seldom, this can be optimized and disable susbcribtion after certain period. 

Although, I am not sure that's worth the complexity of subscriptions for this case:
* This service is a short term solution
* We don't have evidence the additional rountrip is harming our users and we can always subscription once its  really needed. 

P.S. In other world where this Service is a long term solution, I would push for the subscription,)

Wondertan (commented):
Is this ADR for node only or is there additional expected for the core? If not, I think we should mention here info about relation to core and how its gonna integrate with the Signer.

docs/adr/adr-013-automatic-gas-price-estimation.md
cmwaters: We can choose either:
a) no cap - the user will pay whatever price is picked
b) Something 100x or 1000x the default price so 0.2 or 2 utia

docs/adr/adr-013-automatic-gas-price-estimation.md
cmwaters: yes by default the consensus node will provide the `GasEstimator` service in the same gRPC connection. However users should be able to override that with a gRPC address of their own choosing which will be used just for gas/pricing estimation

docs/adr/adr-013-automatic-gas-price-estimation.md
cmwaters: It's not an error to take a long time. There are other ways a user can view the progress of their transaction and with something like replace-by-fee in the future, they will even be able to boost the gas price of their transaction

docs/adr/adr-013-automatic-gas-price-estimation.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Context
+
+[CIP 18](https://github.com/celestiaorg/CIPs/blob/main/cips/cip-18.md) defines an interface that a light node can use for transaction submission to retrieve an estimate of the gas price and amount of gas required to include their transaction on Celestia. This predominantly is to allow for greater dynamism during times of congestions where gas prices become greater than the global minimum.
+
+This ADR defines how the go light node will interact with the service as a client. Note this does not cover how the server will actually estimate the gas prices and gas required.
+
+## Decision
+
+Allow users of Celestia nodes to opt in to automatic gas price and gas usage estimation for every write method
```
rach-id, line 19:
```suggestion
Allow users of Celestia nodes to opt in to automatic gas price and gas usage estimation for every write method.
```

docs/adr/adr-013-automatic-gas-price-estimation.md
cmwaters: I get it's used by all node types. I would keep it simple for now and if it becomes problematic we can come up with a few optimizations or even change it to a subscription model.

cmwaters:
> Is this ADR for node only or is there additional expected for the core? If not, I think we should mention here info about relation to core and how its gonna integrate with the Signer.

It's for node only. There is no plan to integrate with signer

docs/adr/adr-013-automatic-gas-price-estimation.md
renaynay: But utilising the fallback is unexpected behaviour. I guess maybe if it's well documented enough how the estimator works then no error returned is fine.

docs/adr/adr-013-automatic-gas-price-estimation.md
renaynay: So an entirely new config field in the `state` config that can specify addr of estimator service. 

This should be included in ADR

docs/adr/adr-013-automatic-gas-price-estimation.md
renaynay: In this case, we should definitely have some spans inside submission process to measure total time as a metric so we can have some data from the get-go.

Wondertan:
> It's for node only. There is no plan to integrate with signer

Anything in particular about why not? The app provides blob submission as well and should support gas estimation as well, imo

docs/adr/adr-013-automatic-gas-price-estimation.md
cmwaters: I'm just going to go for b) here

docs/adr/adr-013-automatic-gas-price-estimation.md
cmwaters: Yeah I don't think it's unexpected if it's specified but I was conscious that it may be better to fail loudly if the estimation service is unavailable.

An idea could be, in the manual submission path, allow for someone to specify the minimum that way users can easily encode that fallback.

docs/adr/adr-013-automatic-gas-price-estimation.md
cmwaters: lol good catch

cmwaters:
> Anything in particular about why not? The app provides blob submission as well and should support gas estimation as well, imo

My mental model is that the Signer should just be about signing and be quite manual and anything automated is an abstraction above it, that way people can always import the signer with their own automation logic (if they don't like ours)

docs/adr/adr-013-automatic-gas-price-estimation.md
cmwaters: Yeah I wasn't sure how exactly it would be implemented. Should I just add that line

docs/adr/adr-013-automatic-gas-price-estimation.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Detailed Design
+
+### When to estimate the gas price
+
+Users can opt in to automatic price estimation by using the existing `TxConfig.isGasPriceSet`. Previously, if this was false, the `DefaultGasPrice` was used, now if a user isn't manually setting the gas price, the node will query it from the `GasEstimator` service.
```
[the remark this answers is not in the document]
cmwaters: Can't full nodes also be connected to consensus node (or whatever the user defines as the grpc service)

Wondertan:
I see you point on Signer and can agree with it. But my point is about support in app broadly within Signer or not.

Also, looking at the ADR again, its has no connection to node, besides minimal TxConfig field implementation details. I start to believe this document is misplaced.

cmwaters:
> I see you point on Signer and can agree with it. But my point is about support in app broadly within Signer or not.

App won't use the API, at least not initially.

> Also, looking at the ADR again, its has no connection to node, besides minimal TxConfig field implementation details. I start to believe this document is misplaced.

It's not just the config, it's also the usage of the API itself, it's whether we should estimate the gas locally or rely on the API, it's the guardrails in place and possibly a fallback. If not in node, where would you like this document to be placed

renaynay:
I guess it's fine to document the behaviour here as the expected behaviour of the default will change.

docs/adr/adr-013-automatic-gas-price-estimation.md
```diff
@@ -0,0 +1,66 @@
...
+
+Depending on the message types submitted, the node will decide whether to also query for the amount of gas that is required to submit the transaction. In the case of the following messages, the gas can be reliably estimated using a simple linear model:
+
+- `MsgPayForBlob`
+
+For all other messages (or combination of messages), the node will call the `GasEstimator` service.
+
+### Setting guardrails
+
+A new field in the `TxConfig` will be introduced allowing for a user to  use the gas price estimation but setting a maximum gas price as a cap that the user is willing to pay. If the price returned by the estimator is greater than the cap, likely due to a price spike, an error will be returned informing the user that the gas price required is currently greater than the user is willing to pay. If the user doesn't specify a cap, a hardcoded value that is 100x the default min gas price will be used (i.e. 0.2 utia).
```
vgonkivs, line 37: Can we assume that [tx priority ](https://github.com/celestiaorg/celestia-app/blob/0e4b98433b084a9a1d05bc8bceec7d6f8b4619e3/proto/celestia/core/v1/gas_estimation/gas_estimator.proto#L36)is medium in case the user wants us to use gas estimator?

vgonkivs: or we can configure the node during the startup to use some priority by default.

Wondertan:
>  It's not just the config, it's also the usage of the API itself, it's whether we should estimate the gas locally or rely on the API, it's the guardrails in place and possibly a fallback. If not in node, where would you like this document to be placed

Yes, `TxConfig` is part of the API and I am in favor of documenting the change. However,  the ADR process, AFAIU works differently. There should have been an State/Core API ADR in the first place where this ( arguably tiny API change) would land on top with an entry in changelog section of the ADR. Others, more app specific changes would land together in the app repo.

Let's agree how we would this ideally and then see if we can go that path. If not, happy to hear short-term options to unblock this.

docs/adr/adr-013-automatic-gas-price-estimation.md
vgonkivs: Q: why do we need to have any limits if user haven't specified them?

docs/adr/adr-013-automatic-gas-price-estimation.md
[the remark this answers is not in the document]
vgonkivs: should we still use `gasMultiplier`? Or It assumes that results will be more accurate?

vgonkivs:
In my opinion, most of the logic from this ADR should be placed on the TxClient, since it is responsible for submitting the txs and it also has gas estimation logic. It turns we have the same logic(estimation) in both `core_accessor` and `TxClient`. I see core_accessor as a component that receives data Requests from user proxy it, converting types where it is necessary. All operations for re-submitting(blobs), gas estimation, should live inside TxClient and should be hidden(but it could be configurable by core_accessor - e.g. passing a custom estimator connection instead of the default one).

Core accessor over complicated with a lot of unneeded logic ([example](https://github.com/celestiaorg/celestia-app/issues/4273)) and I propose to define and divide responsibilities between core_accessor and txClient.

docs/adr/adr-013-automatic-gas-price-estimation.md
cmwaters: The default is medium priority - users don't have to specify it if they don't want to. I guess it should also be present in the Node API as part of the config

cmwaters: > Q: why do we need to have any limits if the user hasn't specified them?

Out of an abundance of caution (i.e. limit the damage this can do if the GasEstimation service is faulty)

cmwaters:
> There should have been an State/Core API ADR in the first place where this ( arguably tiny API change) would land on top with an entry in changelog section of the ADR

Sure I would agree with that

docs/adr/adr-013-automatic-gas-price-estimation.md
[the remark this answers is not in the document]
cmwaters: Good point. Tbh, I think we should remove the `gasMultiplier` and assume that is taken care of on the server side. What do you think?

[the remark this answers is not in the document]
vgonkivs: We can use it only in case of a fallback solution when requesting `EstimateGas` which simulates the transaction.

docs/adr/adr-013-automatic-gas-price-estimation.md
vgonkivs: Is there any chance that connection with this endpoint will be secured and we'll need a tls config or xToken to communicate with?

docs/adr/adr-013-automatic-gas-price-estimation.md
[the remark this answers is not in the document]
vgonkivs: > Tbh, I think we should remove the gasMultiplier and assume that is taken care of on the server side. What do you think?

Double checked and the test failed with `out of gas`. So I'll keep it.

docs/adr/adr-013-automatic-gas-price-estimation.md
vgonkivs: What if the error occurs after submitting the transaction( e.g `out of gas`)? Should also fall back to a default solution with an additional warning to the user(in case he provides his own estimator service address)?

docs/adr/adr-013-automatic-gas-price-estimation.md
vgonkivs: I think it is worth adding that we should not use queried gas price that is less than minGasPrice

docs/adr/adr-013-automatic-gas-price-estimation.md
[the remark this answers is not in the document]
cmwaters: I do think we should remove it and have it in the server side. We can just multiply the result by 1.1 on the server side

docs/adr/adr-013-automatic-gas-price-estimation.md
renaynay: Let's clarify this please @cmwaters

renaynay (changes requested):

docs/adr/adr-013-automatic-gas-price-estimation.md
cmwaters: I've come back around on this and think that we should remove the fallback. It should just error to the user in the case that their service is unavailable.

docs/adr/adr-013-automatic-gas-price-estimation.md
```diff
@@ -0,0 +1,62 @@
...
+
+- 17.01.2024: Initial design
+
+## Status
+
+Proposed
+
+## Context
+
+[CIP 18](https://github.com/celestiaorg/CIPs/blob/main/cips/cip-18.md) defines an interface that a light node can use for transaction submission to retrieve an estimate of the gas price and amount of gas required to include their transaction on Celestia. This predominantly is to allow for greater dynamism during times of congestions where gas prices become greater than the global minimum.
```
fl0rek, line 13:
```suggestion
[CIP 18](https://github.com/celestiaorg/CIPs/blob/main/cips/cip-018.md) defines an interface that a light node can use for transaction submission to retrieve an estimate of the gas price and amount of gas required to include their transaction on Celestia. This predominantly is to allow for greater dynamism during times of congestions where gas prices become greater than the global minimum.
```
CIP filenames got updated to `%03d` format

docs/adr/adr-013-automatic-gas-price-estimation.md
```diff
@@ -0,0 +1,62 @@
...
+### Negative
+
+- Extra round trip with either the consensus node or someone offering the `GasEstimator` service before the transaction is submitted
+- Reliance on a third party service which could become unavailable.
+
+### Neutral
+
+## References
+
+- [CIP 18](https://github.com/celestiaorg/CIPs/blob/main/cips/cip-18.md)
```
fl0rek, line 62:
```suggestion
- [CIP 18](https://github.com/celestiaorg/CIPs/blob/main/cips/cip-018.md)
```
CIPs filenames got updated to `%03d` format

cmwaters:
Going to close this as stale 😞

PR #4051, opened 2025-01-20
[Draft] New Block Reconstruction Protocol Proposal

walldiss:
This PR introduces a draft proposal for a new block reconstruction protocol aimed at improving network efficiency and scalability.

Key improvements:
- Reduces duplicate requests and bandwidth usage
- Structured bitmap sharing using Roaring Bitmaps
- Clear separation of client/server responsibilities
- Robust failure handling with peer penalties
- Optimized proof packaging and sharing

The document serves as a starting point for discussion and will be later split into CIP and ADR.

TODO:
- [ ] Add testing plan
- [ ] Add false positive triggered reconstruction impact scenario. Potential harm and mitigation

### DIFF
--- a/docs/adr/reconstruction.md
+++ b/docs/adr/reconstruction.md
@@ -68,6 +68,21 @@ If a Full Node (FN) cannot retrieve a block via the shrex protocol, it initiates
 
 ---
 
+## Sampling Protocol
+
+The current implementation of the sampling protocol relies on [bitswap](https://github.com/ipfs/go-bitswap) to request samples from Full (or Bridge) nodes. Under normal assumptions—when data is reliably available—bitswap works well. It provides a robust system for fetching data by content identifier (CID), along with advanced features for load distribution and prioritization.
+
+However, bitswap has a critical limitation: **it lacks content discovery**. When content discovery is absent, a node must attempt requests with each peer until it finds one that actually has the data. This becomes especially problematic during reconstruction. In that scenario, only the attacker nodes hold the block data initially; honest Full Nodes do not. As a result:
+
+1. Light Nodes (LNs) repeatedly attempt to fetch samples from honest Full Nodes, which cannot serve these requests because they don’t yet have the block data.
+2. This generates excessive “spam” requests directed at nodes that are in the process of reconstructing the block, ultimately degrading performance.
+
+**Replace bitswap with shrex-based Samples protocol.**
+
+To resolve this issue, the protocol must incorporate a more effective content discovery mechanism. Fortunately, the **shrex-based Samples protocol**, already required for reconstruction, can also serve LN sampling needs. By adopting this unified solution, we can eventually phase out bitswap for **all** data retrieval tasks, removing it as a dependency and streamlining the system’s architecture.
+
+---
+
 ## Reconstruction Flow
 
 Below is a step-by-step illustration of the reconstruction flow, with accompanying diagrams.
@@ -134,6 +149,26 @@ Below is a step-by-step illustration of the reconstruction flow, with accompanyi
 
 ---
 
+### Protocol Diagrams
+
+Below is an outline of the proposed protocols. Full flow diagrams appear at the end of this document.
+
+```
+1. Bitmap Subscription 
+   Client (FN)                            Server (FN)
+      |---- Subscribe to bitmap -------------->|
+      |<---- Initial bitmap -------------------|
+      |<---- Updates  -------------------------|
+      |<---- End updates ----------------------|
+
+2. GetSamples
+   Client (FN)                     Server (FN/LN)
+      |---- Request(bitmap) ----------->|
+      |<---- [Samples + Proof] parts ---|
+```
+
+---
+
 ### No Bitmap Subscription From Light Nodes: Why?
 
 Subscribing to bitmaps from Light Nodes would allow more precise deduplication of requested samples, but it also introduces additional complexity and round-trip overhead. Since each Light Node holds relatively few samples, the probability of overlap is low. Instead, the Full Node can send an inverse “have” bitmap in its request to indicate which samples are still needed.
@@ -163,26 +198,6 @@ The data indicates that overhead is negligible for larger block sizes. To keep t
 
 ---
 
-### Protocol Diagrams
-
-Below is an outline of the proposed protocols. Full flow diagrams appear at the end of this document.
-
-```
-1. Bitmap Subscription 
-   Client (FN)                            Server (FN)
-      |---- Subscribe to bitmap -------------->|
-      |<---- Initial bitmap -------------------|
-      |<---- Updates  -------------------------|
-      |<---- End updates ----------------------|
-
-2. GetSamples
-   Client (FN)                     Server (FN/LN)
-      |---- Request(bitmap) ----------->|
-      |<---- [Samples + Proof] parts ---|
-```
-
----
-
 ## Core Components
 
 ### List of Core Components
@@ -193,6 +208,7 @@ Below is an outline of the proposed protocols. Full flow diagrams appear at the 
 4. Samples retrieval protocol
 5. Samples store (new file format)
 6. Peer identification
+7. Light node sampling protocol
 
 ---
 
@@ -399,6 +415,12 @@ Peer identification ensures Full Nodes can distinguish Full Nodes from Light Nod
 - Light Nodes: `GetSamples`
 
 Peer information can be inferred from the host (e.g., user agent or a `libp2p.Identity` protocol).
+
+---
+
+### 7. Light node sampling protocol
+
+A new protocol for Light Nodes to acquire block samples directly from Full Nodes, **based on the Samples Retrieval Protocol (item #4)**. By adopting this approach, we can phase out bitswap and unify the mechanisms used for both normal sampling and reconstruction. Key advantages include:
 
 ---
 
--- a/docs/adr/reconstruction.md
+++ b/docs/adr/reconstruction.md
@@ -1,95 +1,101 @@
 ## Abstract
 
-This document proposes a new block reconstruction protocol that addresses key bottlenecks in the current implementation, specifically targeting duplicate request reduction and bandwidth optimization. The protocol introduces a structured approach to data retrieval with an explicit focus on network resource efficiency and scalability.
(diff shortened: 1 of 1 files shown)


musalbas:
Looks generally fine to me directionally. Networking protocol should probably be in a CIP as it's relevant to other clients like lumina etc.

docs/adr/reconstruction.md
```diff
@@ -0,0 +1,286 @@
...
+
+
+### Core Protocol Components
+
+#### 1. Connection Management
+
+Handshakes, Node Type Identification
+- Node type must be declared during connection handshake by each of connection nodes
+- Active connection state maintenance on full nodes. List should be maintained by node type
+- Connection state should allow subscription for updated
```
Wondertan, line 98: Whats the rationale behind sharing your node type? Is it possible to avoid sharing this information with peers?

There is a privacy concern we discussed a while ago with @musalbas and @adlerjohn. FNs ideally should be indistinguishable from LNs on p2p network, so they validators don't detect them.

If we don't think that's important anymore or sharing type is unavoidable, I also propose to cut the handshake roundtrip and rely on libp2p toolset for that.

musalbas: That is indeed a goal, but the problem here is that it's implicit in this protocol that only full nodes request bitmaps from others, regardless of how you identify in the handshake.

FNs pretending to be LNs (to reconstruct) would only work if the FNs spam sample requests to the block producer, which isn't something implemented yet afaik. (See https://forum.celestia.org/t/security-levels-for-data-availability-for-light-nodes/919#level-3-data-availability-sampling-without-an-honest-minority-of-light-nodes-6)

Wondertan: > if the FNs spam sample requests to the block producer, which isn't something implemented yet afaik. 

Do you think this should be in scope of current iteration of reconstruction protocol?

docs/adr/reconstruction.md
```diff
@@ -0,0 +1,286 @@
...
+- Penalty system for delayed updates
+
+#### 3. Reconstruction State Management
+
+3.1. Global State Components
+```go
+type RemoteState struct {
+    coords [][]peers    // Peer lists by coordinates
+    rows []peers        // Peers with row data
+    cols []peers        // Peers with column data
```
renaynay, line 124: how is this actually going to be structured? 

you keep a []peer.ID per which row they claim to have? So index 0 []peer.ID will be peers that claim to have row 0?

docs/adr/reconstruction.md
```diff
@@ -0,0 +1,286 @@
...
+3.2. State Query Interface
+
+Remote state:
+- GetPeersWithSample(Coords) -> []peer.ID
+- GetPeersWithAxis(axisIdx, axisType) -> []peer.ID
+- Available() -> bitmap
+
+Local state:
+- InProgress() -> bitmap. Tracks ongoing fetch sessions to prevent requesting duplicates
+- Have() -> bitmap.  Tracks
```
renaynay, line 142:
```suggestion
- Have() -> bitmap.  Tracks node's locally available samples
```

docs/adr/reconstruction.md
```diff
@@ -0,0 +1,286 @@
...
+4.1. Request Initiation may have multiple strategies:
+- Immediate request of all missing samples bitmap receipt
+- Optional: Delayed start for bandwidth optimization
+    - Wait for X% peer responses
+    - Fixed time delay
+    - Complete EDS availability
+    - Pre-confirmation threshold
+    - Combination of conditions
+
+4.1.1 what and were (to request) biggest question.
```
renaynay, line 158: what and where ?

docs/adr/reconstruction.md
```diff
@@ -0,0 +1,286 @@
...
+- Parallel request distribution to reduce network load on single peers
+- Request from peers with smaller latency
+
+
+4.2. Sample Request Protocol
+
+4.2.1. Request 
+
+- Update InProgress bitmap before request initiation
+- Use shrex for data retrieval. Send bitmap for data request. Shrex has built-in support for requesting rows and samplies, however bitmap-based data retrieval is more general and would be easier to support in future, because it allows requesting multiples pieces of data with single request instead of multiple requests
```
renaynay, line 178:
```suggestion
- Use shrex for data retrieval. Send bitmap for data request. Shrex has built-in support for requesting rows and samples, however bitmap-based data retrieval is more general and would be easier to support in future, because it allows requesting multiples pieces of data with single request instead of multiple requests
```

docs/adr/reconstruction.md
```diff
@@ -0,0 +1,286 @@
...
+```
+
+
+4.2.2. Response 
+- (base version) Server should respond with samples with proofs
+
+Optiomisations:
+- Server may split response into multiple parts
+Each part contains packed samples in Range format.
+- Each part of shoold have specified message prefix. It would allow client to identify sample container type, which would help to maintain backwards compatibility in case of breaking changes in packing algorithm
```
renaynay, line 192:
```suggestion
- Each part of should have specified message prefix. It would allow client to identify sample container type, which would help to maintain backwards compatibility in case of breaking changes in packing algorithm
```

docs/adr/reconstruction.md
```diff
@@ -0,0 +1,286 @@
...
+type SampleRequest struct {
+    Samples *roaring.Bitmap
+}
+```
+
+
+4.2.2. Response 
+- (base version) Server should respond with samples with proofs
+
+Optiomisations:
```
renaynay, line 189:
```suggestion
Optimisations:
```

docs/adr/reconstruction.md
```diff
@@ -0,0 +1,286 @@
...
+- Client should verify proofs. If proof is invalid, peer should be penalized
+- Verified samples should stored and added to local state `Have`
+- Clean up InProgress bitmap
+   - Clean up information from remote state to free up memory
+- If reconstruction is complete clean up local state, shut down reconstruction process and all subscriptions
+
+### Server-Side Implementation
+
+#### 1. Storage Interface
+New storage format needs to be implemented for efficient storage of sampels proofs. The format will be used
```
renaynay, line 208:
```suggestion
New storage format needs to be implemented for efficient storage of sample proofs. The format will be used
```

docs/adr/reconstruction.md
```diff
@@ -0,0 +1,286 @@
...
+initially used for storing ongoing reconstruction process and later can be used for light node storage.
```
renaynay, line 209:
```suggestion
initially for storing ongoing reconstruction process and later can be used for light node storage.
```

renaynay (commented):
Would this proposal also require some modifications in discovery + peer management as it currently stands? If so, can you please outline what that would look like?

docs/adr/reconstruction.md
musalbas: I think not in this iteration, because it seems we're doing pretty good on increasing number of light nodes, so we could just focus directly on having a robust construction protocol and achieving Level 4 light nodes. It might be nice to have later as an option, though, but not high priority right now.

walldiss: Good point, handshake is not necessary for initial basic implementation. It will be needed in peer scoring optimisation. When FN would need to know the type of peer for prioritisation of requests. I'll mark handshake protocol as future optimisation.

docs/adr/reconstruction.md
walldiss: Yes. The idea is to have efficient look up for peers who have full row or full col.

docs/adr/reconstruction.md
```diff
@@ -0,0 +1,317 @@
...
+
+## Backwards Compatibility
+
+In the lifespan of the protocol, it may require a coordinated network upgrade. Implementation should allow for:
+1. Version negotiation
+2. Transition period support
+3. Fallback mechanisms
+
+
+
```
walldiss, line 245: Add testing plan / requirements

walldiss (commented):
Feedback notes from sync call

docs/adr/reconstruction.md
```diff
@@ -0,0 +1,488 @@
...
+   Client (FN)                     Server (FN/LN)
+      |---- Request(bitmap) ----------->|
+      |<---- [Samples + Proof] parts ---|
+```
+
+---
+
+### No Bitmap Subscription From Light Nodes: Why?
+
+Subscribing to bitmaps from Light Nodes would allow more precise deduplication of requested samples, but it also introduces additional complexity and round-trip overhead. Since each Light Node holds relatively few samples, the probability of overlap is low. Instead, the Full Node can send an inverse “have” bitmap in its request to indicate which samples are still needed.
```
musalbas, line 174: Does this mean that the FN needs to regularly send the bitmap (or a diff of the bitmap) again to LNs as the FN collects more samples? Or is it not necessary as most LNs only have a few samples anyway?

walldiss: Good question! The short answer is that **updates are not needed.**

Because each Light Node typically holds only a few samples, a single request round-trip to the LN is usually sufficient to retrieve everything it has. If the LN sampled only some of 16 coordinates it can provide none. So the FN can retry later with an increasing backoff. This approach allows the FN to determine whether a particular LN has fully sampled the block and no additional bitmap exchanges are required. Essentially, once the LN reports having all samples (or delivers all it has), the FN can safely move on.

musalbas: How does the FN know if the LN has all the samples though? SamplesRequest struct doesn't provide for this.

musalbas: I think this will result cases where the block cannot be reconstructed even if LNs have enough shares to do so -- for example if 1/4 of the block is withheld, most LN sampling will fail, til reconstruction happens. It's a chicken and egg situation...

walldiss: Yes, that’s correct: if 1/4 of the block is withheld, LNs can’t finish sampling and therefore won’t serve any shares. 99% of LNs will treat the block as unavailable. Although it may seem like a “chicken and egg” issue, the protocol is actually protecting LNs from finalizing incomplete blocks. The block could be recovered if the missing shares eventually surface, but until then, it remains withheld. The assumption here is that protocol’s main goal is to prevent LNs from accepting blocks that aren’t fully available.

Do you think this scenario is a viable attack vector? In given case random 1% of LN will get into thinking that block is available, which is the same as the chance of false positive sampling

musalbas: My point is that in such a scenario, LNs should be sharing their partial samples with FNs, in case FNs can reconstruct the missing parts of the block, and complete the rest of their samples.

PR #4053, opened 2025-01-20
misc(state/core_accessor): add retry logic for creating a txClient

vgonkivs:
Create a txClient in runtime if it was initialized during the startup


state/core_access.go
```diff
@@ -55,7 +55,6 @@ type CoreAccessor struct {
 	keyring keyring.Keyring
 	client  *user.TxClient
 
-	// TODO: remove in scope of https://github.com/celestiaorg/celestia-node/issues/3515
```
cristaloleg, line 58: IMO better to remove when 3515 will be closed.

state/core_access.go
```diff
@@ -156,17 +165,15 @@ func (ca *CoreAccessor) Start(ctx context.Context) error {
 		return fmt.Errorf("wrong network in core.ip endpoint, expected %s, got %s", ca.network, defaultNetwork)
 	}
 
-	// set up signer to handle tx submission
-	ca.client, err = ca.setupTxClient(ctx, ca.defaultSignerAccount)
+	err = ca.setupTxClient(ctx)
 	if err != nil {
-		log.Warnw("failed to set up signer, check if node's account is funded", "err", err)
+		log.Warn(err)
```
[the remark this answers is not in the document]
vgonkivs: I've wrapped the error inside setupTxClient.

state/core_access.go
vgonkivs: 3515 is not relevant anymore. I will close it once this pr is merged with the explanation.
TLDR: you can't rely on the txClient if it is not initialized(so you can't get the default address and account).

renaynay (commented):
why not just try to set it up in `getSigner` if it's == nil ?

vgonkivs:
`getSigner` parses the TxConfig(signer is not required to be a defaultAccount) which is not the same as creating a client.

renaynay:
My point is `getSigner` is called everywhere where the signer is needed, so why can't we lazy re-init in there if signer == nil ?

renaynay:
@vgonkivs I mean why can't we have the same logic (`ca.getTxClient()`) to either return already-initialised client or try to lazy re-init the client? 

so instead of on #L122 instead of calling `setupTxClient`, there's a fucntion

`getTxClient()` which checks if ca.client !=nil, returns the client, but if nil, then it tries to set it up.

vgonkivs:
Ah, ok. I did not get your initial idea. Sorry. will rework it.

Wondertan (approved):

cristaloleg (approved):
SGTM

renaynay (approved):

PR #4067, opened 2025-01-25
fix(shrex): improve resource management

DeVikingMark:
### Prevent resource leaks in shrex recovery middleware:
   - Add stream closure in panic recovery handler
   - Ensures proper cleanup of network resources in case of panics
   - Prevents potential resource leaks in error scenarios


walldiss:
Steam will be closed by handler

PR #4068, opened 2025-01-25
fix(p2p): improve error handling

DeVikingMark:
Remove redundant error handling in p2p/resources.go:
   - Remove unnecessary error check and panic() call in traceReporter
   - NewStatsTraceReporter() never returns an error, making the check redundant
   - Makes the code cleaner and more maintainable


cristaloleg:
Thank you for the PR but this change is invalid and not gonna compile. Closing as not planned.

PR #4087, opened 2025-02-04
feat(state/core_accessor): add fee estimator

vgonkivs:
This PR enables a third-party fee estimator.

TODO:
* add description;
* add tests;


nodebuilder/state/core.go
```diff
@@ -23,13 +24,14 @@ func coreAccessor(
 	fraudServ libfraud.Service[*header.ExtendedHeader],
 	network p2p.Network,
 	client *grpc.ClientConn,
+	address core.EstimatorAddress,
```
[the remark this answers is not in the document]
vgonkivs: We are using core.EstimatorAddress only from the core config:

```
type Config struct {
	IP   string
	Port string
	TLSEnabled bool
	XTokenPath string
	FeeEstimatorAddress EstimatorAddress
}
```

Wondertan:
Did we tested this manually? Like run a node with an estimator submitting a real blob?

vgonkivs:
Submit blobs is using a linear formula to calculate gas. I tested it with `Transfer`

Wondertan:
Oh, i see, didnt know that, but you made it with Transfer - to confirm?

vgonkivs:
It worked fine last month when I was testing it iirc. I'm getting an error from the estimation endpoint now. Trying to clarify with the app team.

vgonkivs:
> BTW, are we sure that estimation is only for Transfers and not blobs?
Yes, as per ADR blobs should be estimated using formula. All other types of the Txs have to use estimation service

vgonkivs:
Tested on `Trasfer` transaction with numia public endpoint.

state/core_access.go
```diff
@@ -35,9 +35,9 @@ const (
 )
 
 var (
-	ErrInvalidAmount = errors.New("state: amount must be greater than zero")
-
-	log = logging.Logger("state")
+	ErrInvalidAmount        = errors.New("state: amount must be greater than zero")
+	errGasPriceExceedsLimit = errors.New("state: estimated gasPrice exceeds max gasPrice")
```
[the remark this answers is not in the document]
vgonkivs: Made exported

state/estimator.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	if e.defaultClientConn == nil || e.defaultClientConn.GetState() == connectivity.Shutdown {
+		return 0, 0, errors.New("connection with the consensus node is dropped")
+	}
+
+	gasPrice, gas, err := signer.QueryGasUsedAndPrice(ctx, e.defaultClientConn, priority.toApp(), rawTx)
+	if err != nil {
+		log.Warn("state: failed to query gas used and price from the default endpoint", "err", err)
+	}
```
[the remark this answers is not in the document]
vgonkivs: simplified

Wondertan (approved):
Skipping review of semantical nits. 
If it works - it works

state/core_access.go
```diff
@@ -136,12 +138,17 @@ func (ca *CoreAccessor) Start(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("querying minimum gas price: %w", err)
 	}
+
+	err = ca.estimator.Start(ctx)
+	if err != nil {
+		log.Warn("state: failed to connect to estimator endpoint", "err", err)
+	}
```
cristaloleg, line 145: should we `return err` here? if no - comment why it's ok to skip it.

renaynay: we should actually return an error here bc the only time err is returned from estimator.Start is if grpc conn fails to be created which is an actual err.

state/estimator.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func (e *estimator) estimate(
+	ctx context.Context,
+	cfg *TxConfig,
+	client *user.TxClient,
+	msg sdktypes.Msg,
+) (float64, uint64, error) {
+	if cfg.GasPrice() != DefaultGasPrice && cfg.GasLimit() != 0 {
+		return cfg.GasPrice(), cfg.GasLimit(), nil
```
renaynay, line 65: so if both GasPrice and GasLimit are set in the config, GasPrice sanity check (checking if it exceeds the MaxGasPrice) can be bypassed? Is the idea here that if the user sets gas themselves that they wouldn't set a max gas price, and if they set a max gas price, that means they probably didn't specify gas and wants the node to calculate for them (within some upper bound)?

We should document this behaviour here

renaynay: comments throughout this function would be nice.

state/estimator.go
```diff
@@ -0,0 +1,147 @@
...
+	return DefaultGasPrice, gas, nil
+}
+
+func (e *estimator) queryGasUsedAndPrice(
+	ctx context.Context,
+	signer *user.Signer,
+	priority TxPriority,
+	rawTx []byte,
+) (gasPrice float64, gas uint64, err error) {
+	conns := []*grpc.ClientConn{e.estimatorConn, e.defaultClientConn}
```
renaynay, line 117: So the idea here is that there will /can be an estimator service that will perform this function, but if not, the default validator you're using to submit txs will also be able to estimate.

cristaloleg: From what I understand your comment is just informational, nothing to change, right?

state/estimator.go
cristaloleg: Done. Not sure about "comments throughout this function", I don't see a good reason for that, added Go-doc only.

cristaloleg:
Closing in favour of #4168 

Thank you @vgonkivs for your time and effort 🫡

state/tx_config.go
```diff
@@ -187,3 +202,20 @@ func WithFeeGranterAddress(granter string) ConfigOption {
...
+func WithMaxGasPrice(gasPrice float64) ConfigOption {
+	return func(cfg *TxConfig) {
+		cfg.maxGasPrice = gasPrice
+	}
+}
+
+// WithTxPriority is an option that allows you to specify a priority of the tx.
+func WithTxPriority(priority int) ConfigOption {
+	return func(cfg *TxConfig) {
+		if priority < 0 || priority > 3 {
```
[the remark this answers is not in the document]
cristaloleg: Done

state/core_access.go
cristaloleg: Done

PR #4089, opened 2025-02-04
refactor: move striplock to libs/utils package

VolodymyrBg:
Move the striplock implementation from store package to libs/utils
to make it more reusable across the codebase. This change:

- Creates new StripLock and MultiLock types in libs/utils package
- Adds proper documentation for all exported types and methods
- Maintains backward compatibility through type aliases
- Marks old package-level functions as deprecated

The implementation remains functionally identical, but is now
properly placed in the utils package as suggested by the TODO comment.

Resolves TODO: move striplock to utils


cristaloleg:
Sounds like https://github.com/celestiaorg/celestia-node/pull/4046

PR #4093, opened 2025-02-06
fix(core): add retry logic for the grpc subscription

### LINKED ISSUE #4090: Bridge Node Stuck After Archive RPC Restart
### Celestia Node version

v0.21.3-mocha

### Celestia Consensus Node version

3.3.0-mocha

### OS

22.04.5 LTS (Jammy Jellyfish)

### Steps to reproduce it

Restart the `archive RPC` while the bridge node is running

### Expected result

The bridge node should be able to recover from an archive RPC restart without requiring a manual service restart.

### Actual result

The bridge node became unresponsive. It repeatedly logged fetcher and listener errors. The issue persisted until a manual restart was performed.

### Relevant log output

```Shell
2025-02-05T07:28:54.515Z        INFO    header/store    store/store.go:367      new head        {"height": 4529796, "hash": "4881C54F1993FBC1BEDC574CBBA71A798210B88C17572DBE970B0673E83FC110"}
2025-02-05T07:28:59.391Z        INFO    header/store    store/store.go:367      new head        {"height": 4529797, "hash": "7953D784D1982641504D365BF30BF8AF3ADE797EB722B87996296C87FAB85BF1"}
2025-02-05T07:29:04.389Z        INFO    header/store    store/store.go:367      new head        {"height": 4529798, "hash": "3A798993AD62B133BB73EA63FEBDF29B1D06B21AA650335B9ADAEB05A38CF754"}
2025-02-05T07:29:09.274Z        INFO    header/store    store/store.go:367      new head        {"height": 4529799, "hash": "B6F80329D1DA76F3602C1D5FB91B27C04835E4AA142F6221E7A175A01015E55B"}
2025-02-05T07:29:14.243Z        INFO    header/store    store/store.go:367      new head        {"height": 4529800, "hash": "BCE184F68E3C165FF60C2BB569CA1DB37C78F2356B34039FFD4261557A049DAF"}
2025-02-05T07:29:17.963Z        INFO    bitswap/server/decision decision/engine.go:733  handling wantlist overflow      {"local": "12D3KooWKdtKbb6KHYnnEjxdMgBT3eVKmuvT8ykjgbajcSKz89LP", "from": "12D3KooWDHfTaYvQE1z4DBKRtH5rKnJc5RrBpjrYhM1EFYttyCkt", "wantlistSize": 4093, "overflowSize": 3}
2025-02-05T07:29:19.217Z        INFO    header/store    store/store.go:367      new head        {"height": 4529801, "hash": "ACA69B7C5EFD9F98372B8D1E18255FBD30B77AD7D76F4299C486B3391278E01C"}
2025-02-05T07:29:24.286Z        INFO    header/store    store/store.go:367      new head        {"height": 4529802, "hash": "91B489EC15D4F1D5CFC0D7B1BD8AC8DF37D5A6FA6159836FB9F09A8E2F8C91D0"}
2025-02-05T07:29:29.497Z        INFO    header/store    store/store.go:367      new head        {"height": 4529803, "hash": "3530B64936561099C82A9E3DF4E4BDBF6359BEC8D196C299A9A155755453ED88"}
2025-02-05T07:29:31.733Z        ERROR   core    core/fetcher.go:191     fetcher: error receiving new height     {"err": "rpc error: code = Unavailable desc = error reading from server: EOF"}
2025-02-05T07:29:31.734Z        ERROR   core    core/fetcher.go:191     fetcher: error receiving new height     {"err": "rpc error: code = Unavailable desc = error reading from server: EOF"}
2025-02-05T07:29:31.734Z        ERROR   core    core/fetcher.go:191     fetcher: error receiving new height     {"err": "rpc error: code = Unavailable desc = error reading from server: EOF"}
2025-02-05T07:29:31.734Z        ERROR   core    core/fetcher.go:191     fetcher: error receiving new height     {"err": "rpc error: code = Unavailable desc = error reading from server: EOF"}
2025-02-05T07:29:31.734Z        ERROR   core    core/fetcher.go:191     fetcher: error receiving new height     {"err": "rpc error: code = Unavailable desc = error reading from server: EOF"}
...
2025-02-05T07:30:09.864Z        ERROR   core    core/fetcher.go:191     fetcher: error receiving new height     {"err": "rpc error: code = Unavailable desc = error reading from server: EOF"}
2025-02-05T07:30:09.864Z        ERROR   core    core/fetcher.go:191     fetcher: error receiving new height     {"err": "rpc error: code = Unavailable desc = error reading from server: EOF"}
2025-02-05T07:30:09.864Z        ERROR   core    core/fetcher.go:191     fetcher: error receiving new height     {"err": "rpc error: code = Unavailable desc = error reading from server: EOF"}
2025-02-05T07:30:09.864Z        ERROR   core    core/fetcher.go:191     fetcher: error receiving new height     {"err": "rpc error: code = Unavailable desc = error reading from server: EOF"}
2025-02-05T07:30:09.864Z        ERROR   core    core/fetcher.go:191     fetcher: error receiving new height     {"err": "rpc error: code = Unavailable desc = error reading from server: EOF"}
Suppressed 1363430 messages from celestia-bridge.service
2025-02-05T07:30:39.499Z        ERROR   core    core/listener.go:171    listener: resubscribe   {"err": "already subscribed to new blocks"}
2025-02-05T07:30:44.499Z        ERROR   core    core/listener.go:171    listener: resubscribe   {"err": "already subscribed to new blocks"}
2025-02-05T07:30:47.994Z        INFO    bitswap/server/decision decision/engine.go:733  handling wantlist overflow      {"local": "12D3KooWKdtKbb6KHYnnEjxdMgBT3eVKmuvT8ykjgbajcSKz89LP", "from": "12D3KooWDHfTaYvQE1z4DBKRtH5rKnJc5RrBpjrYhM1EFYttyCkt", "wantlistSize": 4094, "overflowSize": 2}
2025-02-05T07:30:49.498Z        ERROR   core    core/listener.go:171    listener: resubscribe   {"err": "already subscribed to new blocks"}
2025-02-05T07:30:54.498Z        ERROR   core    core/listener.go:171    listener: resubscribe   {"err": "already subscribed to new blocks"}
2025-02-05T07:30:59.499Z        ERROR   core    core/listener.go:171    listener: resubscribe   {"err": "already subscribed to new blocks"}
```

Yes

### Notes

We encountered an issue today on our testnet bridge node. After restarting the archive RPC, our bridge node got stuck and could not recover on its own. However, a manual service restart resolved the issue.

vgonkivs:
Closes #4090 
Added retry logic for the grpc subscription.  In case, the server goes down for some reason, block fetcher will attempt to create a new subscription every 5 seconds. 

Tested manually.

To perform unit/swamp tests we need https://github.com/celestiaorg/celestia-app/issues/4304


core/exchange_test.go
```diff
@@ -62,7 +61,6 @@ func TestCoreExchange_RequestHeaders(t *testing.T) {
 		require.NoError(t, err)
 		assert.True(t, has)
 	}
-	require.NoError(t, fetcher.Stop(ctx))
```
cristaloleg, line 65: Why?

core/fetcher.go
```diff
@@ -160,64 +171,82 @@ func (f *BlockFetcher) ValidatorSet(ctx context.Context, height int64) (*types.V
...
 			select {
-			case <-ctx.Done():
+			case <-f.ctx.Done():
+				return
+			default:
+			}
+
+			subscription, err = f.client.SubscribeNewHeights(f.ctx, &coregrpc.SubscribeNewHeightsRequest{})
+			switch {
+			case err == nil:
```
cristaloleg, line 196: Nit: add a comment like `// ok` to know that nothing is missed here.

core/fetcher.go
```diff
@@ -160,64 +171,82 @@ func (f *BlockFetcher) ValidatorSet(ctx context.Context, height int64) (*types.V
...
+	for {
+		resp, err := subscription.Recv()
+		if err != nil {
+			return err
+		}
+
+		withTimeout, ctxCancel := context.WithTimeout(ctx, 10*time.Second)
+		signedBlock, err := f.GetSignedBlock(withTimeout, resp.Height)
+		ctxCancel()
+		if err != nil {
```
cristaloleg, line 230: Should we exclude `ctx.Err()` here?

core/fetcher.go
```diff
@@ -160,64 +171,82 @@ func (f *BlockFetcher) ValidatorSet(ctx context.Context, height int64) (*types.V
...
-	}
-	ctx, cancel := context.WithCancel(ctx)
-	f.cancel = cancel
-	f.doneCh = make(chan struct{})
-	f.isListeningForBlocks.Store(true)
+func (f *BlockFetcher) runSubscriber() (chan types.EventDataSignedBlock, error) {
+	var (
+		subscription coregrpc.BlockAPI_SubscribeNewHeightsClient
+		err          error
+	)
```
walldiss, line 178: subscription and error do not need to be predeclared here, since it only propagated down the stack from for loop

core/fetcher.go
walldiss, line 196: grpc dialer will use exponential backoff retry by default. what is the point of timeout timer here? I think it can be removed and the code will be much simpler

core/fetcher.go
```diff
@@ -160,64 +171,82 @@ func (f *BlockFetcher) ValidatorSet(ctx context.Context, height int64) (*types.V
...
+				log.Errorw("fetcher: failed to subscribe to new block events", "err", err)
+				timeout.Reset(retrySubscriptionDelay)
+				<-timeout.C
+				continue
+			}
+
+			err = f.receive(f.ctx, signedBlockCh, subscription)
+			if err != nil {
+				log.Errorw("fetcher: error receiving new height", "err", err.Error())
+				continue
```
walldiss, line 209: if ctx.Cancel need to return here too.

core/fetcher.go
```diff
@@ -160,64 +171,82 @@ func (f *BlockFetcher) ValidatorSet(ctx context.Context, height int64) (*types.V
...
+		}
+
+		withTimeout, ctxCancel := context.WithTimeout(ctx, 10*time.Second)
+		signedBlock, err := f.GetSignedBlock(withTimeout, resp.Height)
+		ctxCancel()
+		if err != nil {
+			log.Errorw("fetcher: error receiving signed block", "height", resp.Height, "err", err.Error())
+			// sleeping a bit to avoid retrying instantly and give time for the gRPC connection
+			// to recover automatically.
+			time.Sleep(time.Second)
```
walldiss, line 234: Why would we need sleep, with backoff in dialer?

core/fetcher.go
```diff
@@ -160,64 +171,82 @@ func (f *BlockFetcher) ValidatorSet(ctx context.Context, height int64) (*types.V
...
-					Commit:       *signedBlock.Commit,
-					ValidatorSet: *signedBlock.ValidatorSet,
-					Data:         *signedBlock.Data,
-				}:
-				case <-ctx.Done():
-					return
-				}
+				log.Errorw("fetcher: failed to subscribe to new block events", "err", err)
+				timeout.Reset(retrySubscriptionDelay)
+				<-timeout.C
```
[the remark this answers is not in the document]
vgonkivs: But it is not recommended to use time.After in the loop as it creates a new timer on every new iteration.

core/fetcher.go
vgonkivs: I guess backoff related to the grcp.Dial and not to the calling client's method.(removed timeout and starts spamming)

core/fetcher.go
vgonkivs: but it will be handled in the next iteration.

core/fetcher.go
vgonkivs: did not touch this code. moved it from the previous version.

core/fetcher.go
vgonkivs: Timer removed

core/listener.go
```diff
@@ -132,70 +132,26 @@ func (cl *Listener) Stop(ctx context.Context) error {
...
+				log.Error("underlying subscription was closed")
+				return
 			}
 
 			if cl.chainID != "" && b.Header.ChainID != cl.chainID {
-				log.Errorf("listener: received block with unexpected chain ID: expected %s,"+
-					" received %s", cl.chainID, b.Header.ChainID)
-				return errInvalidSubscription
+				// stop node if there is a critical issue with the block subscription
+				panic(fmt.Sprintf("listener: received block with unexpected chain ID: expected %s,"+
```
renaynay, line 153: log.Fatalw pls

core/listener.go
```diff
@@ -132,70 +132,26 @@ func (cl *Listener) Stop(ctx context.Context) error {
...
 	for {
 		select {
 		case b, ok := <-sub:
 			if !ok {
-				return errors.New("underlying subscription was closed")
+				log.Error("underlying subscription was closed")
+				return
 			}
 
 			if cl.chainID != "" && b.Header.ChainID != cl.chainID {
```
renaynay, line 151: when would cl.chainID ever be empty? Come to think of it.

core/listener.go
renaynay: and with more details about the block (height, hash)

core/listener.go
```diff
@@ -211,13 +167,13 @@ func (cl *Listener) listen(ctx context.Context, sub <-chan types.EventDataSigned
...
-			timeout.Reset(cl.listenerTimeout)
 		case <-timeout.C:
 			cl.metrics.subscriptionStuck(ctx)
-			return errors.New("underlying subscription is stuck")
+			log.Error("underlying subscription is stuck")
 		case <-ctx.Done():
-			return ctx.Err()
+			return
 		}
+		timeout.Reset(cl.listenerTimeout)
```
renaynay, line 176: won't get hit

core/listener_test.go
```diff
@@ -90,11 +90,11 @@ func TestListenerWithWrongChainRPC(t *testing.T) {
...
 	// create Listener and start listening
 	cl := createListener(ctx, t, fetcher, ps0, eds, store, "wrong-chain-rpc")
-	sub, err := cl.fetcher.SubscribeNewBlockEvent(ctx)
+	require.NoError(t, cl.fetcher.Start(ctx))
+	sub, err := cl.fetcher.runSubscriber()
 	require.NoError(t, err)
 
-	err = cl.listen(ctx, sub)
-	assert.ErrorIs(t, err, errInvalidSubscription)
+	assert.Panics(t, func() { cl.listen(ctx, sub) })
```
renaynay, line 97: It might be annoying to test the log.Fatal bc it's an os exit status 1 

I guess maybe we can just keep it as a panic for now.

nodebuilder/core/constructors.go
```diff
@@ -30,13 +33,30 @@ func grpcClient(lc fx.Lifecycle, cfg Config) (*grpc.ClientConn, error) {
...
+	)
+	retryStreamInterceptor := grpc_retry.StreamClientInterceptor(
+		grpc_retry.WithMax(5),
+		grpc_retry.WithCodes(codes.Unavailable),
+		grpc_retry.WithBackoff(
+			grpc_retry.BackoffExponentialWithJitter(time.Second, 2.0)),
+	)
+
+	opts = append(opts, grpc.WithUnaryInterceptor(retryInterceptor))
+	opts = append(opts, grpc.WithStreamInterceptor(retryStreamInterceptor))
```
renaynay, line 51: can you combine

core/listener.go
```diff
@@ -211,13 +167,13 @@ func (cl *Listener) listen(ctx context.Context, sub <-chan types.EventDataSigned
 			if !timeout.Stop() {
 				<-timeout.C
 			}
-			timeout.Reset(cl.listenerTimeout)
 		case <-timeout.C:
```
renaynay, line 170: can we just track this in fetcher and remove the timer from listener entirely?

vgonkivs: along with the metric(subscription is stuck)?

core/listener.go
vgonkivs: the main reason why panic is because of log.Fatal calls os.Exit which is pretty tricky to catch and test.

core/listener.go
renaynay: I understand the point of the metric. I didn't say we should remove the metric - i'm saying we should track the metric in fetcher where it occurs and remove the concept of the `timer` from the listener listen loop.

core/listener.go
renaynay: ??

core/listener.go
vgonkivs: I don't think the fetcher should know about any limitations of waiting for new blocks. IMO, the fetcher should only wait for them and pass to the listener which responsible for the verification, storing and broadcasting them.

core/listener.go
vgonkivs: It will after select finishes.

core/fetcher.go
```diff
@@ -33,11 +32,12 @@ var (
 )
 
 type BlockFetcher struct {
+	ctx    context.Context
```
walldiss, line 35: Nit: Don't need to store ctx, it is anti-pattern. It can be passed to all routines from Start call

core/fetcher.go
```diff
@@ -47,16 +47,28 @@ func NewBlockFetcher(conn *grpc.ClientConn) (*BlockFetcher, error) {
 	}, nil
 }
 
+func (f *BlockFetcher) Start(ctx context.Context) error {
+	f.ctx, f.cancel = context.WithCancel(ctx)
```
Wondertan, line 51: Is this Start supposed to be used in fx? If so, the context given to it short-lived

core/listener.go
```diff
@@ -101,11 +96,16 @@ func (cl *Listener) Start(context.Context) error {
 	cl.cancel = cancel
 	cl.closed = make(chan struct{})
 
-	sub, err := cl.fetcher.SubscribeNewBlockEvent(ctx)
+	err := cl.fetcher.Start(ctx)
+	if err != nil {
+		return err
+	}
+	subs, err := cl.fetcher.runSubscriber()
```
walldiss, line 103: nit suggestion: listener can pass subscriber context to fetcher. It will allow listener to manage lifecycle of subscription. fetcher on other hand would not have lifecycle at all.

core/fetcher.go
```diff
@@ -161,73 +172,81 @@ func (f *BlockFetcher) ValidatorSet(ctx context.Context, height int64) (*types.V
...
 		defer close(f.doneCh)
 		defer close(signedBlockCh)
-		defer func() { f.isListeningForBlocks.Store(false) }()
 		for {
 			select {
-			case <-ctx.Done():
+			case <-f.ctx.Done():
+				return
+			default:
+			}
```
walldiss, line 186: Is it necessary? Looks like next call (SubscribeNewHeights) can exit on canceled context

core/fetcher.go
```diff
@@ -161,73 +172,81 @@ func (f *BlockFetcher) ValidatorSet(ctx context.Context, height int64) (*types.V
...
-					Data:         *signedBlock.Data,
-				}:
-				case <-ctx.Done():
-					return
-				}
+				// try re-subscribe in case of any errors that can come during subscription. gRPC
+				// retry mechanism has a back off on retries, so we don't need timers anymore.
+				log.Errorw("fetcher: failed to subscribe to new block events", "err", err)
+				continue
+			}
```
walldiss, line 199: nit: would be eaier to read with return fast approch:
if err.Is(ctx.Cancel)

core/fetcher.go
walldiss: You would log.Error here, which is unwanted

core/fetcher.go
```diff
@@ -161,73 +172,81 @@ func (f *BlockFetcher) ValidatorSet(ctx context.Context, height int64) (*types.V
...
+	signedBlockCh chan types.EventDataSignedBlock,
+	subscription coregrpc.BlockAPI_SubscribeNewHeightsClient,
+) error {
+	log.Error("fetcher: started listening for new blocks")
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+		}
```
walldiss, line 222: Is it necessary? Perhaps subscription already tracks context, since it was provided to it during construction.

core/fetcher.go
```diff
@@ -161,73 +172,81 @@ func (f *BlockFetcher) ValidatorSet(ctx context.Context, height int64) (*types.V
...
+
+		resp, err := subscription.Recv()
+		if err != nil {
+			return err
+		}
+
+		withTimeout, ctxCancel := context.WithTimeout(ctx, 10*time.Second)
```
walldiss, line 229: It is nice to have a timeout here, but the value of 10 seconds seems to be low. It might fail due to some networking temporary problems. Would also be nice to make it configurable.

core/listener.go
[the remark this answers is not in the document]
vgonkivs: I'm starting to think that block fetcher should be divided into the BlockFetcher and BlockSubscriber since start and stop methods are related to the subscription only. All other methods will work even after the stop is called which is confusing. so the listener will keep both - the subscriber and fetcher independently, and from the other side core/exchange will not have access to the subscription logic

core/fetcher.go
```diff
@@ -161,73 +172,81 @@ func (f *BlockFetcher) ValidatorSet(ctx context.Context, height int64) (*types.V
...
+		signedBlock, err := f.GetSignedBlock(withTimeout, resp.Height)
+		ctxCancel()
+		if err != nil {
+			log.Errorw("fetcher: error receiving signed block", "height", resp.Height, "err", err.Error())
+			continue
+		}
```
walldiss, line 235: What if it error? Can it cause not downloading and storing a block failed block? If not, would be nice if you can add comment here, since it is unclear how missed blocks will handled from this code.

core/listener.go
vgonkivs: I like the idea of not managing the lifecycle of the fetcher.

core/fetcher.go
vgonkivs: This is the idea. It can be canceled only if context was canceled(in `listener.Stop`)

core/fetcher.go
vgonkivs: Added a TODO

core/listener.go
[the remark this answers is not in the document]
vgonkivs: Start and Stop hooks are removed.

core/fetcher.go
vgonkivs: No Start/Stop hooks anymore

core/fetcher.go
vgonkivs: They will be handled through `core.Exchange`

walldiss (approved):
less code, more tests - love it! Much simpler and easier than it used to be!

renaynay (approved):
utack - thank you for cleanup!

PR #4101, opened 2025-02-10
fix: Add deterministic seed for random share removal in tests

avorylli:
Initializes random number generator with time-based seed in removeRandShares() test helper to ensure proper randomization of test data and prevent potential test flakiness.


cristaloleg (changes requested):
This doesn't prevent "potential test flakiness" and does "proper randomization of test data" at all.

PR #4116, opened 2025-02-13
chore(ci): fail openrpc-gen if error

### LINKED ISSUE #4112: [Feature Request]: prevent CI from passing for openrpc.json if there is a failure within spec generation
### Implementation ideas

In the case of "failed to retrieve example value for type: network.Direction on parent 'p2p.ConnectionState')", this should fail in CI.

cristaloleg:
Fixes #4112


vgonkivs (approved):

renaynay (approved):
utack

PR #4117, opened 2025-02-13
fix(share/peer-manager): fix potential nil dereference in peer manager

walldiss:
If pool was removed during request it could lead to unexpected panic in client.


cristaloleg (approved):
SGTM

renaynay (approved):

PR #4131, opened 2025-02-18
fix(cmd/celestia): Split subcommands into full subcommands and auxiliary

### LINKED ISSUE #4063: fix(module/pruner): WARN log for disabled full pruning service should only pop up on `start` if `--experimental-pruning` is not passed.
Suppress WARN log from `module/pruner` that pruning is disabled when using CLI for anything other than `start` subcommand for **full** and **bridge** node types that are not running with full pruning enabled, via the `--experimental-pruning` flag.

renaynay:
All subcommands were inheriting the same flagsets + parsing logic even though 4/6 subcommands only need access to `--node.store` and `--p2p.network`. 

This splits logic such that only the necessary flags + parsing logic are added to auxiliary subcommands.

This also resolves #4063


cmd/util.go
```diff
@@ -75,13 +75,31 @@ func DecodeToBytes(param string) ([]byte, error) {
...
+func ParseMinimumFlags(cmd *cobra.Command) error {
+	ctx := cmd.Context()
+
+	net, err := p2p.ParseNetwork(cmd)
+	if err != nil {
+		return err
+	}
+	ctx = WithNetwork(cmd.Context(), net)
+
+	ctx, err = ParseNodeStore(cmd.Context(), cmd, net)
```
cristaloleg, line 87: `s/cmd.Context()/ctx` ?

PR #4132, opened 2025-02-21
chore: bump go version to match the go.mod

tty47:
helloo team! 👋 

small one, just bumping the docker versions to match the [go.mod](https://github.com/celestiaorg/celestia-node/blob/main/go.mod#L3)

best!


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM --platform=$BUILDPLATFORM docker.io/golang:1.23-alpine3.20 AS builder
```
walldiss, line 1: 1.23 uses latest golang within 1.23 family, the image is updated in docker registry on each patch release an right now points to 1.23.6

https://hub.docker.com/_/golang

walldiss:
outdated

PR #4134, opened 2025-02-25
docs: added badges and links to twitter and discord

sashaphmn:


cristaloleg:
Thanks for the idea but we are okay with current badges.

sashaphmn:
Ok

PR #4136, opened 2025-02-26
fix: use a more straightforward return value

looklose:
Since there is already a check for err != nil and a return statement, err must be nil at this point.


cristaloleg (approved):
Nice catch, thanks!

looklose:
> Nice catch, thanks!

My pleasure.

looklose:
Thanks for your review.  Do I need to rebase the latest code? 

If there's anything you need me to do, please let me know; I'm very happy to do it.

walldiss (approved):

cristaloleg (approved):

PR #4142, opened 2025-03-03
refactor: use errors.New to replace fmt.Errorf with no parameters

LesCyber:
Replaced fmt.Errorf with errors.New in cases where formatting is not required. This reduces unnecessary function calls, leading to slightly improved performance and cleaner code.


Wondertan:
Hello. Its unlikely there would be any performance gain from using errors.New in the error code path and cleanless argument is similalrly debatable. We prefer using fmt.Errorf when inlined in a function and errors.New only when we define a comparable error type

PR #4145, opened 2025-03-05
fix(ci): Add fetch-depth:0 to upload-docs checkout action

crStiv:
Fix potential issues with git tags during release process by adding fetch-depth:0 to actions/checkout in upload-docs job.

### DIFF
--- a/.github/workflows/ci_release.yml
+++ b/.github/workflows/ci_release.yml
@@ -135,6 +135,8 @@ jobs:
       contents: write
     steps:
       - uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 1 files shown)


cristaloleg:
What does it solve exactly?

crStiv:
> What does it solve exactly?

@cristaloleg 

The current setup can fail when trying to work with git tags during releases because of the default shallow clone. Adding fetch-depth:0 makes sure we have access to the full git history, so the workflow can properly handle tags when uploading docs.

walldiss:
We don't use shallow clones

PR #4146, opened 2025-03-06
feat(cel-shed): datastore: erase samples

Wondertan:
This new cel-shed command allows erasing samples and relevant daser/availability/pruning states. I needed this to resample, but without wasting hours on resyncing headers, and I believe this is gonna be useful for us.

Besides, I added a command to delete a single arbitrary namespace as opposed to a constant group of namespace.

### DIFF
--- a/cmd/cel-shed/datastore.go
+++ b/cmd/cel-shed/datastore.go
@@ -1,0 +1,132 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"github.com/celestiaorg/celestia-node/nodebuilder"
+	"github.com/ipfs/boxo/blockstore"
+	ds "github.com/ipfs/go-datastore"
+	"github.com/ipfs/go-datastore/namespace"
+	dsq "github.com/ipfs/go-datastore/query"
+	"github.com/spf13/cobra"
+	"golang.org/x/sync/errgroup"
+)
+
+func init() {
+	datastoreCmd.AddCommand(eraseCmd, eraseSamplesCmd)
+}
+
+var datastoreCmd = &cobra.Command{
+	Use:   "datastore [subcommand]",
+	Short: "Collection of datastore related utilities",
+}
+
+var eraseCmd = &cobra.Command{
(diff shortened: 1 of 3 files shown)


cmd/cel-shed/datastore.go
```diff
@@ -0,0 +1,134 @@
...
+	datastoreCmd.AddCommand(eraseCmd, eraseSamplesCmd)
+}
+
+var datastoreCmd = &cobra.Command{
+	Use:   "datastore [subcommand]",
+	Short: "Collection of datastore related utilities",
+}
+
+var eraseCmd = &cobra.Command{
+	Use:   "erase <ds_key>",
```
renaynay, line 28: Why need the discover stopped? Why can't they pass either `--node.store` or `--p2p.network` to make it more explicit and versatile cmd?

Wondertan: It should have both, so you can be lazy without specifying anything, but also can override of non default path

walldiss: I personally have multiple types of nodes for multiple networks. Would appreciate an option to specify network name + type of the node in this command. And only network for command of erasing samples

Wondertan: Yes, that's TBD

cmd/cel-shed/datastore.go
```diff
@@ -0,0 +1,134 @@
...
+		fmt.Printf("Erased %s\n", key)
+		return nil
+	},
+	Args: cobra.ExactArgs(1),
+}
+
+var sampleDataKeys = []ds.Key{
+	ds.NewKey("sampling_result"),
+	ds.NewKey("das"),
+	ds.NewKey("pruner"),
```
Wondertan, line 65: TODO: Make those importable

mcrakhman (approved):
Basically I would have the same comment as other reviewers that to make this behavior predictable, by default it should work if you have only one store, if you have multiple: you should specify a single named store otherwise you should stop.  But other than that seems fine.

PR #4149, opened 2025-03-06
perf(header): MsgID func only needs to unmarshal Commit from ExtendedHeader

renaynay:
`header.MsgID` now only unmarshals commit from header instead of whole thing as we only need to take hash of commit for msg ID. 

Note, MsgID is only calculated on BN side when it publishes headers, so this optimisation is only helpful for BNs.

4.6x faster execution
memory reduced by 55%
heap allocs cut by 70% 

I've included 2 tests: 

* benchmark for MsgID
* equivalency check to ensure MsgID calculated now is the same as was calculated with prior behaviour


header/serde.go
```diff
@@ -64,18 +64,30 @@ func UnmarshalExtendedHeader(data []byte) (*ExtendedHeader, error) {
 	return out, nil
 }
 
-// msgID computes an id for a pubsub message
+// unmarhsalCommit // TODO @renaynay
```
Wondertan: ?

header/serde.go
```diff
@@ -64,18 +64,30 @@ func UnmarshalExtendedHeader(data []byte) (*ExtendedHeader, error) {
...
+	err := in.Unmarshal(data)
+	if err != nil {
+		return nil, err
+	}
+
+	return core.CommitFromProto(in.Commit)
+}
+
+// MsgID computes an id for a pubsub message
 // TODO(@Wondertan): This cause additional allocations per each recvd message in the topic
```
cristaloleg, line 81: Already outdated? (assuming this and this changes https://github.com/celestiaorg/go-header/pull/251)

Wondertan: There are two redundant allocs. The one in go-header fixes one and this is another one. The one here is not fixable entirely without large protocol breaking changes, so we only minimize its impact

Wondertan (approved):

header/headertest/serde_test.go
```diff
@@ -30,6 +33,51 @@ func TestMarshalUnmarshalExtendedHeader(t *testing.T) {
 	equalExtendedHeader(t, in, out)
 }
 
+func TestMsgIDEquivalency(t *testing.T) {
+	randHeader := RandExtendedHeader(t)
+	bin, err := randHeader.MarshalBinary()
+	require.NoError(t, err)
+
+	oldMsgIDFunc := func(message *pubsub_pb.Message) string {
```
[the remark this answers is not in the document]
cristaloleg: Resolving as per DM discussion.

cristaloleg (approved):

PR #4150, opened 2025-03-06
fix(shwap/discovery): increase DHT refresh interval to default 10mins

### LINKED ISSUE #4147: discovery: decrease routing table refresh interval
Current value is [one minute](https://github.com/celestiaorg/celestia-node/blob/9c1d27a9c41cf16f7b20e96a556d9601b29b5b3a/share/shwap/p2p/discovery/dht.go#L16). This means that every minute every node performs DHT queries connecting to new peers. 

The side effect of this behavior is that libp2p's connection manager becomes ineffective and worse we create a high node churn. It is set to drop connections after one minute(grace period). So in we end, we get situation where the connection manager drops connections every minute and DHT rediscovers them back. 

The default value in DHT itself is 10 minutes, so we may just go for it.

Wondertan:
Closes #4147

### DIFF
--- a/share/shwap/p2p/discovery/dht.go
+++ b/share/shwap/p2p/discovery/dht.go
@@ -3,17 +3,12 @@ package discovery
 import (
 	"context"
 	"fmt"
-	"time"
 
 	"github.com/ipfs/go-datastore"
 	dht "github.com/libp2p/go-libp2p-kad-dht"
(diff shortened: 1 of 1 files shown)


vgonkivs (approved):

renaynay (approved):
okay les go

cristaloleg (approved):

PR #4154, opened 2025-03-07
fix(das): correct catch-up detection condition for light nodes

smuu:
### Problem
The condition in `checkDone()` was incorrectly checking if `s.next > s.networkHead` instead of `s.next >= s.networkHead+1`. This condition was too strict for light nodes, which process blocks differently than full nodes.

### Solution
- Changed the condition from `s.next > s.networkHead` to `s.next >= s.networkHead+1`

### Impact
Light nodes will now correctly report when they have caught up to the network head, which improves user experience and monitoring capabilities.

### DIFF
--- a/das/state.go
+++ b/das/state.go
@@ -283,7 +283,7 @@ func (s *coordinatorState) unsafeStats() SamplingStats {
 }
 
 func (s *coordinatorState) checkDone() {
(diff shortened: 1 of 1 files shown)


smuu:
It seems that the issue I observe has a different origin. I will continue to investigate the issue.

walldiss:
Can it be closed?

PR #4156, opened 2025-03-07
feat!(share/shwap): implement get range request over shwap

vgonkivs:
This PR enables GetRangeRequests over the shwap by introducing the GetRangeNamespaceData container. It allows users to retrieve namespace data using ODS coordinates [from; to]. The user must specify the type of response they want:
Data + Proofs;
ProofsOnly;
The returned proof can be verified against the data root.


share/shwap/p2p/shrex/shrex_getter/shrex.go
```diff
@@ -306,6 +306,16 @@ func (sg *Getter) GetNamespaceData(
...
 
+func (sg *Getter) GetRangeNamespaceData(
+	_ context.Context,
+	_ *header.ExtendedHeader,
+	_ libshare.Namespace,
+	_, _ shwap.SampleCoords,
+	_ bool,
+) (shwap.RangeNamespaceData, error) {
+	return shwap.RangeNamespaceData{}, errors.New("not supported")
+}
```
walldiss, line 317: Why not implement in shrex, but implement in bitswap? We planing to remove bitswap soon, so perhaps makes sense to do the opposite

share/shwap/proof.go
```diff
@@ -0,0 +1,220 @@
...
+// committed data root.
+// It consists of multiple components to support verification:
+// * shareProof: A nmt proof that verifies the inclusion of a specific data share within a row of the datasquare.
+// * rowRootProof: A Merkle proof that verifies the inclusion of the row root within the final data root.
+// * root: The row root against which the proof is verified
+type Proof struct {
+	shareProof   *nmt.Proof
+	rowRootProof *merkle.Proof
+	root         []byte
+}
```
walldiss, line 28: This proof currently uses a two-step inclusion verification:

- Inclusion of a row into the row root.
- Inclusion of row roots into the data root.

This proof structure isn't optimal because a range request would require multiple row-root proofs, which is unnecessary. It should be possible to optimize proofs for `GetRange` over shwap, as we already have the necessary code implemented in `types.Proof` (from the app repo).

It's important to address this now to avoid introducing breaking changes in a future P2P release, which would otherwise require a network migration and significant additional effort.

walldiss:
Discussed with the team. Below are clarifications addressing previously raised concerns and the subsequent next steps for this PR:

1. We will retain the current Bitswap implementation. There is no need to implement `GetRange` over ShrEx in this PR, as support for ShrEx will be introduced separately with generalized ShrEx. No changes required here.

2. Implement the following TODO:
https://github.com/vgonkivs/celestia-node/blob/4093946715f8d1cd4588a20e651d66c0ca12476a/share/shwap/range_namespace_data.go#L28
```
TODO: Proof collection can be simplified to store only incomplete rows (the first and the last), as proofs for all complete rows can be recomputed.
```

3. Remove the row root hash from the DataRoot proof structure. Currently, verification occurs in two steps:
   - Proof to row root.
   - Proof of row root inclusion in the DataRoot.

It is possible to verify inclusion directly against the DataRoot, bypassing the first step entirely. This adjustment will eliminate the need to store the row root hash within the proof structure. Achieving this will require minimal modifications in the NMT repository to support constructing the row root from Proof + Data. Once NMT support is available, this repository will also require updates to adopt the new verification algorithm.

walldiss:
NMT repo feature required for this PR:
https://github.com/celestiaorg/nmt/issues/290

go.mod
```diff
@@ -2,6 +2,10 @@ module github.com/celestiaorg/celestia-node
 
 go 1.23.5
 
+replace github.com/celestiaorg/nmt => github.com/gupadhyaya/nmt v0.0.0-20250428044855-61fbcbfa18b7
+
+///Users/gupadhyaya/go/src/github.com/gupadhyaya/nmt
```
gupadhyaya, line 7: will remove in next update.

share/shwap/proof.go
```diff
@@ -49,64 +46,141 @@ func (p *Proof) RowRootProof() *merkle.Proof {
...
+		p.shareProof.IsMaxNamespaceIDIgnored(),
+	)
+
+	nid := namespace.ID()
+	leaves := libshare.ToBytes(shares)
+
+	// If the proof is empty and no leaves exist, it's a valid proof
+	if p.shareProof.IsEmptyProof() && len(leaves) == 0 {
+		return nil
 	}
```
vgonkivs, line 61: empty proof is not valid proof. the proof is empty, then we must throw an error back to the user. There is no sense for the user to request a range of [0;0]. Could you please double-check 
`func (rngdata *RangeNamespaceData) VerifyShares`. There are conditions for checking Empty and AbsenseProof. I'd propose making these checks inside `VerifyInclusion` under a `ValidateBasic` method. Requesting a range user must exactly know the coordinates of this range, so if he missed smth we have to notify him IMO.

cc @walldiss @Wondertan

vgonkivs (commented):
Focused more on the Creation/Verification part as I assume all other parts are  untouched. General comment: we need an additional `proot_test.go` with unit test coverage of the new proof structure.

share/shwap/proof.go
```diff
@@ -0,0 +1,271 @@
...
+		return fmt.Errorf("failed to compute leaf hashes: %w", err)
+	}
+
+	// Validate the proof structure
+	if err := p.shareProof.ValidateProofStructure(nth, nid, hashes); err != nil {
+		return fmt.Errorf("invalid proof structure: %w", err)
+	}
+
+	// Compute the root from proof and leaf hashes
+	root, err := p.shareProof.ComputeRoot(nth, hashes)
```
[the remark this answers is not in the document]
gupadhyaya: added ComputeRootWithBasicValidation function

share/shwap/range_namespace_data.go
```diff
@@ -0,0 +1,298 @@
...
+// RangedNamespaceDataFromShares builds a range of namespaced data for the given coordinates:
+// shares is a list of shares (grouped by the rows) relative to the data square and
+// needed to build the range;
+// namespace is the target namespace for the built range;
+// from is the coordinates of the first share of the range within the EDS.
+// to is the coordinates of the last inclusive share of the range within the EDS.
+func RangedNamespaceDataFromShares(
+	shares [][]libshare.Share,
+	namespace libshare.Namespace,
+	rowRootProofs []*merkle.Proof,
```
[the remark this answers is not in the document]
gupadhyaya: > I'd create a data root proof as the last step in this function.

what does this mean? i just passed the AxisRoots and computed rowRootProofs locally. Are you suggesting anything beyond that?

share/shwap/proof.go
```diff
@@ -0,0 +1,271 @@
...
+	leaves := libshare.ToBytes(shares)
+
+	// Compute leaf hashes
+	hashes, err := nmt.ComputePrefixedLeafHashes(nth, nid, leaves)
+	if err != nil {
+		return fmt.Errorf("failed to compute leaf hashes: %w", err)
+	}
+
+	// Validate the proof structure
+	if err := p.shareProof.ValidateProofStructure(nth, nid, hashes); err != nil {
```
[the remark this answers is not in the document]
gupadhyaya: can you check if the updated code looks good?

share/shwap/range_namespace_data.go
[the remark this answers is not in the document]
vgonkivs: no, just AxisRoots instead of []*merkle.Proof.

share/shwap/range_namespace_data.go
```diff
@@ -0,0 +1,323 @@
...
+	shares [][]libshare.Share,
+	namespace libshare.Namespace,
+	from SampleCoords,
+	to SampleCoords,
+	dataHash []byte,
+) error {
+	proofs := rngdata.Proof
+	// compute the proofs for the complete rows
+	for i, row := 0, from.Row; i < len(shares); i++ {
+		// skip the incomplete rows
```
vgonkivs, line 152: you can additionally check that row root proof is not empty here

share/eds/proofs_cache_test.go
```diff
@@ -10,7 +10,7 @@ import (
 
 func TestCache(t *testing.T) {
 	ODSSize := 16
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second*15)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
```
vgonkivs, line 13: lets return 15 seconds back

share/shwap/range_namespace_data.go
gupadhyaya: whats the purpose for adding the additional check?

share/shwap/range_namespace_data.go
```diff
@@ -0,0 +1,323 @@
...
+		if !proofs[i].shareProof.IsEmptyProof() {
+			continue
+		}
+
+		extendedRowShares, err := share.ExtendShares(shares[i])
```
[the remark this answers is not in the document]
gupadhyaya: loop will not be entered right?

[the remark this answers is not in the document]
gupadhyaya: instead added this check to protect panic for `proofs[i]`
```
if len(shares) != len(proofs) {
		return fmt.Errorf("shares and proofs length mismatch")
}
```

share/shwap/p2p/bitswap/range_namespace_data_block.go
```diff
@@ -0,0 +1,149 @@
...
+
+		var rnd shwappb.RangeNamespaceData
+		if err := rnd.Unmarshal(cntrData); err != nil {
+			return fmt.Errorf("unmarshaling RangeNamespaceData for %+v: %w", rndb.ID, err)
+		}
+
+		rangeNsData, err := shwap.RangeNamespaceDataFromProto(&rnd)
+		if err != nil {
+			return fmt.Errorf("unmarshaling RangeNamespaceData for %+v: %w", rndb.ID, err)
+		}
```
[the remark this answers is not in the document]
gupadhyaya: why do we need that check?

share/shwap/p2p/bitswap/range_namespace_data_block.go
```diff
@@ -0,0 +1,149 @@
...
+	if rndb.ID.ProofsOnly {
+		rnd.CleanupData()
+	}
+	rndb.Container = rnd
+	return nil
+}
+
+func (rndb *RangeNamespaceDataBlock) UnmarshalFn(root *share.AxisRoots) UnmarshalFn {
+	return func(cntrData, idData []byte) error {
+		if empty := rndb.Container.IsEmpty(); !empty {
```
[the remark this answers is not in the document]
gupadhyaya: you mean return error instead of nil?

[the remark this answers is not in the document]
vgonkivs: `log.Warn()`

share/shwap/proof.go
```diff
@@ -0,0 +1,235 @@
...
+	return p.shareProof
+}
+
+// RowRootProof returns a merkle proof of the row root to the data root.
+func (p *Proof) RowRootProof() *merkle.Proof {
+	return p.rowRootProof
+}
+
+// VerifyInclusion verifies the inclusion of the shares to the data root.
+func (p *Proof) VerifyInclusion(shares []libshare.Share, namespace libshare.Namespace, dataRoot []byte) error {
```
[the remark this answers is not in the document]
vgonkivs: It is called inside `RngData.Verify`  but the Proof field if exported and VerifyInclusion is exported too, so we are not protected if someone calls it directly.

share/shwap/range_namespace_data.go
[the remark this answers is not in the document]
vgonkivs: yes, true. I misunderstood the brackets in the github's ui

share/shwap/range_namespace_data.go
vgonkivs: you are doing additional computation here that takes some time and memory. but they won't make any sense if rowRootProofs were not set correctly as it will fail on this step. Actually, I meant before the loop

share/shwap/p2p/bitswap/range_namespace_data_block.go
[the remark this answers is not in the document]
gupadhyaya: they can still provide a non-empty, but junk proof to get through right? unless we verify, we don't know.

share/shwap/range_namespace_data.go
gupadhyaya: actually, we have to check it inside the loop for every proof[i], the corresponding row root proof is set.

share/shwap/p2p/bitswap/range_namespace_data_block.go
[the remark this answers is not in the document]
vgonkivs: But we can't verify empty Data Root proof and verification of the (even junk) data does not make any sense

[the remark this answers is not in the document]
gupadhyaya: ok. will add the non-empty proof check.

share/shwap/p2p/bitswap/range_namespace_data_block.go
```diff
@@ -0,0 +1,153 @@
...
+		}
+		if len(rangeNsData.Proof) == 0 {
+			return fmt.Errorf("unmarshaling RangeNamespaceData for %+v: proof is empty", rndb.ID)
+		}
+		if !rndb.ID.ProofsOnly {
+			if err := rangeNsData.Verify(rndid.DataNamespace, rndid.From, rndid.To, root.Hash()); err != nil {
+				return fmt.Errorf("validating RangeNamespaceData for %+v: %w", rndb.ID, err)
+			}
+		}
+		rndb.Container = *rangeNsData
```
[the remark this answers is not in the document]
vgonkivs: yes, assuming the user already has the data and can verify it on his end.

share/shwap/range_namespace_data_id.go
```diff
@@ -0,0 +1,184 @@
...
+}
+
+// ReadFrom reads the binary form of RangeNamespaceDataID from the provided reader.
+func (rngid *RangeNamespaceDataID) ReadFrom(r io.Reader) (int64, error) {
+	data := make([]byte, RangeNamespaceDataIDSize)
+	n, err := io.ReadFull(r, data)
+	if err != nil {
+		return int64(n), err
+	}
+	if n != RangeNamespaceDataIDSize {
```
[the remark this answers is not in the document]
gupadhyaya: could read more than `RangeNamespaceDataIDSize` bytes and still have err == nil ?

share/shwap/range_namespace_data_id.go
```diff
@@ -0,0 +1,184 @@
...
+	err := rngid.Verify(edsSize)
+	if err != nil {
+		return RangeNamespaceDataID{}, fmt.Errorf("verifying range id: %w", err)
+	}
+	return rngid, nil
+}
+
+// Verify validates the RangeNamespaceDataID fields and verifies that number of the requested shares
+// does not exceed the number of shares inside the ODS.
+func (rngid RangeNamespaceDataID) Verify(edsSize int) error {
```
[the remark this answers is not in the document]
gupadhyaya: good catch.

share/shwap/range_namespace_data.go
```diff
@@ -0,0 +1,329 @@
...
+
+	toInclusive := odsIndex + length - 1
+	toCoords, err := SampleCoordsFrom1DIndex(toInclusive, edsSize/2)
+	if err != nil {
+		return SampleCoords{}, SampleCoords{}, err
+	}
+	return from, toCoords, nil
+}
+
+func generateSharesProofs(
```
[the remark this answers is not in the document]
gupadhyaya: we have similar code in RowNamespaceDataFromShares, but not exactly same (in terms of return values and additional row specific checks).

share/shwap/range_namespace_data.go
```diff
@@ -0,0 +1,329 @@
...
+//   - namespace: The target namespace ID to filter and extract data for.
+//   - axisRoots: The AxisRoots (row and column roots) associated with the EDS.
+//   - from: The starting coordinate (row, column) of the range within the EDS.
+//   - to: The ending coordinate (row, column) of the range (inclusive).
+//
+// Returns:
+//   - A RangeNamespaceData containing the shares in the given namespace and Merkle proofs
+//     needed to verify them.
+//   - An error if the range is invalid or extraction fails.
+func RangedNamespaceDataFromShares(
```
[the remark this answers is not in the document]
gupadhyaya: range_namespace_data_test.go

share/shwap/proof.go
```diff
@@ -0,0 +1,241 @@
...
+
+	leaves := libshare.ToBytes(shares)
+
+	// Compute leaf hashes
+	hashes, err := nmt.ComputePrefixedLeafHashes(nth, nid, leaves)
+	if err != nil {
+		return fmt.Errorf("failed to compute leaf hashes: %w", err)
+	}
+
+	root, err := p.shareProof.ComputeRootWithBasicValidation(nth, nid, hashes, false)
```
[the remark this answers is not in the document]
gupadhyaya: in verify inclusion there is no requirement for namespace completeness, also we just insert the namespace id to leaves ourself right above, so no point validating the namespace again.

share/shwap/getter.go
```diff
@@ -47,4 +47,13 @@ type Getter interface {
...
 	// Verify method.
 	GetNamespaceData(context.Context, *header.ExtendedHeader, libshare.Namespace) (NamespaceData, error)
+	// GetNamespaceData gets all namespaced shares from an EDS within a given range.
+	// Shares are returned in a row-by-row order if the range spans multiple rows.
+	GetRangeNamespaceData(
+		_ context.Context,
+		_ *header.ExtendedHeader,
+		_ libshare.Namespace,
+		from, to SampleCoords,
+		_ bool,
```
Wondertan: I guess the goal is to keep parameters named so that they are self-documenting for `SampleCoords`. On that basis, I would also name the boolean parameter, otherwise its not clear what is it for. Also, the comment does not have any mentioning of a proof only mode

share/shwap/getters/cascade.go
```diff
@@ -102,6 +103,30 @@ func (cg *CascadeGetter) GetNamespaceData(
...
+) (shwap.RangeNamespaceData, error) {
+	ctx, span := tracer.Start(ctx, "cascade/get-shares-range", trace.WithAttributes(
+		attribute.String("namespace", namespace.String()),
+	))
+	defer span.End()
+
+	if from.Row > to.Row {
+		return shwap.RangeNamespaceData{},
+			fmt.Errorf("start row must not be bigger to end row: %d-%d", from.Row, to.Row)
+	}
```
Wondertan, line 129: optional: ideally, we would have here the same validation as we do in the validation accessor. The DAH in the extended header allows us to know the size of the eds, and thus we can check the boundaries 

its optional because relying on DAH right now feels wrong considering all the discussions about removing it.

share/shwap/getters/cascade.go
```diff
@@ -102,6 +103,30 @@ func (cg *CascadeGetter) GetNamespaceData(
...
 
+func (cg *CascadeGetter) GetRangeNamespaceData(
+	ctx context.Context,
+	header *header.ExtendedHeader,
+	namespace libshare.Namespace,
+	from, to shwap.SampleCoords,
+	proofsOnly bool,
+) (shwap.RangeNamespaceData, error) {
+	ctx, span := tracer.Start(ctx, "cascade/get-shares-range", trace.WithAttributes(
+		attribute.String("namespace", namespace.String()),
```
Wondertan: Shouldn't we add coordinates too? Maybe it's somewhere else in codepath already tho and there is not much value duplicating it

share/eds/proofs_cache_test.go
gupadhyaya: we are adding an additional test:

```
t.Run(fmt.Sprintf("RangeNamespaceData:%s", name), func(t *testing.T) {
				t.Parallel()
				testRangeNamespaceData(ctx, t, createAccessor, size)
			})
```
it does make sense that we will need a higher timeout?

nodebuilder/share/share.go
```diff
@@ -58,7 +48,9 @@ type Module interface {
 		ctx context.Context, height uint64, namespace libshare.Namespace,
 	) (shwap.NamespaceData, error)
 	// GetRange gets a list of shares and their corresponding proof.
-	GetRange(ctx context.Context, height uint64, start, end int) (*GetRangeResult, error)
+	GetRange(
```
Wondertan, line 71: It would be great to have more documentation for the call, as this is our top-level API call. It's fine if it just copies the comment on the Getter once it is more descriptive as well

share/shwap/getter.go
```diff
@@ -47,4 +47,13 @@ type Getter interface {
...
+	) (RangeNamespaceData, error)
```
Wondertan, line 86: Any particular reason to make it a value instead of a pointer?

share/shwap/range_namespace_data.go
```diff
@@ -0,0 +1,323 @@
...
+) (RangeNamespaceData, error) {
+	if len(shares) == 0 {
+		return RangeNamespaceData{}, fmt.Errorf("empty share list")
+	}
+
+	roots := append(axisRoots.RowRoots, axisRoots.ColumnRoots...) //nolint: gocritic
+	_, rowRootProofs := merkle.ProofsFromByteSlices(roots)
+
+	numRows := len(shares)
+	odsSize := len(shares[0]) / 2
```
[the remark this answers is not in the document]
renaynay: @gupadhyaya code cleanliness - let's use axis roots as that's more explicit what we're measuring.

share/shwap/proof.go
```diff
@@ -0,0 +1,241 @@
...
+	outside, err := share.IsOutsideRange(namespace, root, root)
+	if err != nil {
+		return fmt.Errorf("failed to check namespace range: %w", err)
+	}
+	if outside {
+		return fmt.Errorf("namespace %x is outside the root range", namespace.ID())
+	}
+
+	// Verify against the data root
+	if err := p.rowRootProof.Verify(dataRoot, root); err != nil {
```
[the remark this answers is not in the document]
renaynay: yes ignore, I don't know what i thought when i was writing this

share/shwap/getter.go
gupadhyaya: i think just following GetNamespaceData, GetRow, etc. @vgonkivs

nodebuilder/share/share.go
gupadhyaya: You mean improve the description? because all other Get* have oneline desc. We should improve across the board, not just GetRange

share/shwap/getters/cascade.go
gupadhyaya: adding coordinates and other parameters to the tracer

share/shwap/range_namespace_data.go
```diff
@@ -0,0 +1,329 @@
...
+// Flatten combines all shares from all rows within the namespace into a single slice.
+func (rngdata *RangeNamespaceData) Flatten() []libshare.Share {
+	var shares []libshare.Share
+	for _, shrs := range rngdata.Shares {
+		shares = append(shares, shrs...)
+	}
+	return shares
+}
+
+func (rngdata *RangeNamespaceData) IsEmpty() bool {
```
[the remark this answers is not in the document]
renaynay: @gupadhyaya that's preciesely what we do not want to happen - we should not panic if IsEmpty is called on a nil range namespace data - we should make sure that cannot be the case.

share/shwap/range_namespace_data.go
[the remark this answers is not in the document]
renaynay: can you make sure to either deduplicate or clarify with better docs the difference?

share/shwap/range_namespace_data_id.go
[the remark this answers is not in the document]
renaynay: it cannot read more, you've allocated the buf

renaynay (commented):
reviewed and closed some resolved comments.

nodebuilder/share/share.go
Wondertan: Agreed

share/shwap/getter.go
vgonkivs: was just following  `GetNamespaceData, GetRow`, keeping in mind to change it in the future. Good catch. Thanks.

share/shwap/range_namespace_data.go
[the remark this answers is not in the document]
gupadhyaya: used RowNamespaceDataFromShares in generateSharesProofs to remove code duplication.

share/shwap/getter.go
renaynay: If we change in future, it will break. We need to document in an issue so we don't forget what we want to do here.

share/shwap/range_namespace_data.go
```diff
@@ -182,6 +182,9 @@ func (rngdata *RangeNamespaceData) Verify(
 	to SampleCoords,
 	dataHash []byte,
 ) error {
+	if rngdata.IsEmpty() {
+		return errors.New("empty RangeNamespaceData")
```
renaynay: last nit, but makes sense to extract this into a err var.

share/shwap/range_namespace_data.go
```diff
@@ -291,6 +294,9 @@ func (rngdata *RangeNamespaceData) VerifyShares(
 
 // Flatten combines all shares from all rows within the namespace into a single slice.
 func (rngdata *RangeNamespaceData) Flatten() []libshare.Share {
+	if rngdata.IsEmpty() {
+		return nil
```
renaynay, line 304: would at least document this possible return value in godoc

share/shwap/range_namespace_data.go
```diff
@@ -299,10 +305,16 @@ func (rngdata *RangeNamespaceData) Flatten() []libshare.Share {
 }
 
 func (rngdata *RangeNamespaceData) IsEmpty() bool {
+	if rngdata == nil {
```
renaynay: can combine this with checks on the line below

share/shwap/range_namespace_data.go
```diff
@@ -299,10 +305,16 @@ func (rngdata *RangeNamespaceData) Flatten() []libshare.Share {
...
+		return true
+	}
 	return len(rngdata.Shares) == 0 && len(rngdata.Proof) == 0
 }
 
 func (rngdata *RangeNamespaceData) ToProto() *pb.RangeNamespaceData {
+	if rngdata.IsEmpty() {
```
renaynay: not sure we need this here/

share/shwap/range_namespace_data.go
```diff
@@ -375,26 +390,23 @@ func generateSharesProofs(
...
-
-	for i, shr := range rowShares {
-		if err := tree.Push(shr.ToBytes()); err != nil {
-			return nil, fmt.Errorf("failed to build tree at share index %d (row %d): %w", i, row, err)
-		}
-	}
-
-	root, err := tree.Root()
+	// Create RowNamespaceData for the entire row first
+	rowData, err := RowNamespaceDataFromShares(rowShares, namespace, row)
```
renaynay: wait why are we doing this here? This feels excessive just to check for NS.

gupadhyaya: yes, it doesn't make sense to do this refactor. reverted it back.
in terms of the difference, the RowNamespaceDataFromShares derives the prove range and also does implicit handling of the non-inclusion proof.

renaynay (approved):
Okay w me; pls recheck latest commits w @walldiss

share/shwap/range_namespace_data_id.go
```diff
@@ -0,0 +1,202 @@
...
+	return rngID, rngID.Validate()
+}
+
+// MarshalBinary encodes RangeNamespaceDataID into binary form.
+func (rngid RangeNamespaceDataID) MarshalBinary() ([]byte, error) {
+	data := make([]byte, 0, RangeNamespaceDataIDSize)
+	return rngid.appendTo(data)
+}
+
+// appendTo helps in constructing the binary representation  of RangeNamespaceDataID
```
Wondertan:
```suggestion
// appendTo helps in constructing the binary representation of RangeNamespaceDataID
```

share/shwap/p2p/shrex/shrex_getter/shrex.go
```diff
@@ -306,6 +306,16 @@ func (sg *Getter) GetNamespaceData(
...
 }
 
+func (sg *Getter) GetRangeNamespaceData(
+	_ context.Context,
+	_ *header.ExtendedHeader,
+	_ libshare.Namespace,
+	_, _ shwap.SampleCoords,
+	_ bool,
+) (shwap.RangeNamespaceData, error) {
+	return shwap.RangeNamespaceData{}, errors.New("not supported")
```
Wondertan: We should use `shwap.ErrOperationNotSupported`; otherwise, the cascade getter won't work properly

Wondertan (dismissed):
All the patterns we've set are followed. Documentation is great. Lot's of edge cases were thought through. Attention to details is clear. Really is a good job for the first major PR in node 👏🏻 

One more non-blocking thing todo here is to finish [the CIP](https://github.com/celestiaorg/CIPs/pull/197). I believe it should be straightforward after having full implementation

renaynay (approved):

PR #4163, opened 2025-03-10
ci: bump golangci-lint

cristaloleg:


.github/workflows/go-ci.yml
```diff
@@ -41,10 +41,10 @@ jobs:
           go-version: ${{ inputs.go-version }}
 
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@v6.1.0
+        uses: golangci/golangci-lint-action@v6.5.0
         with:
           args: --timeout 10m
-          version: v1.60
+          version: latest
```
cristaloleg, line 47: Anyone against that?

.golangci.yml
```diff
@@ -72,10 +72,6 @@ linters-settings:
       - G115 # integer overflow conversion
   dogsled:
     max-blank-identifiers: 3
-  golint:
```
cristaloleg, line 75: Not supported anymore and config check fails with it.

.github/workflows/go-ci.yml
Wondertan: I am all for

Wondertan (approved):

walldiss (approved):

renaynay (approved):

Issue #4065, opened 2025-01-24
feat(nodebuilder/p2p): Add startup connectivity check for Full and Bridge nodes


walldiss:
### Implementation ideas

Use the same Reachability logic already in the p2p packages of nodebuilder to verify whether a node can accept inbound connections at startup. If the node is unreachable (e.g., due to NAT or firewall rules), display a clear warning and guidance.

yiannisbot:
Just adding some more details here to make clearer where the need for startup connectivity check comes from.

As of 2025-01-10, ProbeLab's LN monitoring infra included support for `wss` and could therefore see `lumina` nodes. See: https://probelab.io/celestia/dht/2025-02/#celestia-client-agents-overall-stacked-plot.

One thing we noticed is that we're seeing different numbers of nodes for `celestia-node/bridge`, `celestia-celestia` and `celestia-node/full` when we crawl the network with Nebula (https://probelab.io/celestia/dht/2025-02/#celestia-agents-overall-stacked-plot) and when we observe peers with our `ants` deployment  (https://probelab.io/celestia/dht/2025-02/#celestia-client-agents-overall-stacked-plot).

The discrepancies are pretty substantial. From a brief look, we found:
- `celestia-node/bridge` count is ~64 from Nebula crawler and 110 from ants observers
- `celestia-celestia` count is 3 from Nebula crawler and ~30 from ants observers
- `celestia-node/full` count is ~11 from Nebula crawler and 16 from ants observers.

The difference between the two approaches (Nebula vs ants) is that Nebula is trying to establish (outgoing) connections with remote peers, whereas ants are sitting there accepting (incoming) connections from other peers. This further means that if we can't see as many nodes with Nebula (outgoing connections), these peers we can't see are likely facing (incoming) connectivity issues, e.g., due to firewalls etc.

Assuming that incoming connectivity is important for those nodes, it's worth investigating further and a good first step is to add a startup connectivity check.

A good check point for whether the connectivity improves over time is to monitor these two plots referenced above on probelab.io:
- Node count with Nebula crawler: https://probelab.io/celestia/dht/2025-05/#celestia-agents-overall-stacked-plot (latest report from Week 5)
- Node count with `ants` deployment: https://probelab.io/celestia/dht/2025-05/#celestia-client-agents-overall-stacked-plot (latest report from Week 5)

PR #4164, opened 2025-03-10
feat(modp2p): reachability check

Wondertan:
Every ten second scream in logs that node is not reachable until it, only if its FN/BN

Closes https://github.com/celestiaorg/celestia-node/issues/4065


nodebuilder/p2p/reachability.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+	autoNAT := getter.GetAutoNat()
+
+	go func() {
+		timer := time.NewTimer(reachabilityCheckTick)
+		defer timer.Stop()
+
+		for {
+			select {
+			case <-timer.C:
```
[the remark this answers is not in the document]
Wondertan: Yes, good catch. I meant Ticker, thanks

cristaloleg (approved):
SGTM

nodebuilder/p2p/reachability.go
```diff
@@ -0,0 +1,37 @@
...
+					return
+				}
+
+				log.Error("Host is not reachable from the public network")
+			case <-ctx.Done():
+				return
+			}
+		}
+	}()
+}
```
walldiss, line 37: Why go routine is needed here?

Wondertan: > Every ten second scream in logs that node is not reachable, only if its FN/BN

cristaloleg (approved):

nodebuilder/p2p/reachability.go
walldiss: It seems the context provided to the call is construction time context. So it will only do the check while other components are being launched. Which implies that ctx.Done may happen before the first check even performed. 

Would it better to do it in sync mode (without go-routine)? Also would you consider to terminate the node if check failed?

Wondertan: > It seems the context provided to the call is construction time context.

This is likely confusing, but its not `Start`-like context. Its a context that is alive for as long as node lives.

Wondertan: > Also would you consider to terminate the node if check failed?

This is possible but tricky to do right. It's not the most secure reachability check protocol(AutoNATv1) and can be abused. I am also not confident that the results of the check are always gonna be correct. It also needs sometime to run, you need to connect and send requests to peers(re sync)

I believe the current solution is simple enough and should show allow us to test this properly and after we can consider to block on it if this solution gonna be more reliable(likely with AutoNATv2 which is in development)

walldiss:
LGTM. Waiting for a decision regarding error clarification before approving.

nodebuilder/p2p/reachability.go
```diff
@@ -0,0 +1,37 @@
+package p2p
+
+import (
+	"context"
+	"time"
+
+	"github.com/libp2p/go-libp2p/core/network"
+)
+
+const reachabilityCheckTick = 10 * time.Second
```
renaynay, line 10: Is this annoying enough to notice?

renaynay (approved):

renaynay (approved):

PR #4166, opened 2025-03-11
docs: replaced the shield and links in the documentation

Lubov66:


walldiss:
CodeQL and Go-CI are different workflows. The current link is correct.

PR #4168, opened 2025-03-12
feat(state/core_accessor): add fee estimator

cristaloleg:
Continuation of #4087 

Implementation of the https://github.com/celestiaorg/celestia-node/pull/4048

This PR enables third-party estimation for the submit transactions in the node(besides PFB). By default, fee estimation relies on the consensus node to which the node is connected. But the user can now provide a non-default endpoint, so the node will query gas price and gas from it. It can be provided via cli(using "core.estimator.address"). The address will be added to the node's config as `FeeEstimatorAddress`.

Additionally, the user can now provide a `maxGasPrice` for every submit tx(via `max.gas.price` flag)- the max price that the user is willing to pay for the transaction. The transaction will not be submitted in case when estimated gas price exceeds the `maxGasPrice` (default: `minGasPrice * 100`)


state/core_access.go
```diff
@@ -142,6 +142,7 @@ func (ca *CoreAccessor) Start(ctx context.Context) error {
 	err = ca.estimator.Start(ctx)
 	if err != nil {
 		log.Warn("state: failed to connect to estimator endpoint", "err", err)
+		return err
```
renaynay, line 145: we can maybe wrap the err but nit

state/estimator.go
```diff
@@ -55,6 +55,9 @@ func (e *estimator) Stop(context.Context) error {
 	return nil
 }
 
+// estimate the gas price and limit.
+// if gas price is not default and gas limit is set then
+// this 2 values going to be returned (skipping all other checks).
```
renaynay, line 60: if gas price and gas limit are both set, then these two values will be returned, bypassing other checks.

renaynay (dismissed):

state/estimator.go
cristaloleg: Done

renaynay (dismissed):

Wondertan (dismissed):
Make to add @vgonkivs as a coauthor in squashed commit

vgonkivs (commented):
Please make sure the fee estimation works by running app node locally and provide its address as estimator address.

renaynay (approved):
🥳

Wondertan (approved):
LFG

jcstein:
this doesn't impact PFBs??

> This PR enables third-party estimation for the submit transactions in the node(besides PFB).

PR #4170, opened 2025-03-13
fix(nodebuilder/p2p): fix autonat == nil panic in test

renaynay:
Fixes a panic that can occur in tests if autonat is nil

Ref https://github.com/celestiaorg/celestia-node/actions/runs/13833242772/job/38702557881?pr=4149#step:4:449

### DIFF
--- a/nodebuilder/p2p/reachability.go
+++ b/nodebuilder/p2p/reachability.go
@@ -23,6 +23,11 @@ func reachabilityCheck(ctx context.Context, host HostBase) {
 		for {
 			select {
 			case <-ticker.C:
+				if autoNAT == nil {
+					log.Error("autoNAT is nil on host")
+					continue
+				}
+
 				reachability := autoNAT.Status()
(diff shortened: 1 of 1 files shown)


nodebuilder/p2p/reachability.go
```diff
@@ -23,6 +23,11 @@ func reachabilityCheck(ctx context.Context, host HostBase) {
 		for {
 			select {
 			case <-ticker.C:
+				if autoNAT == nil {
```
cristaloleg, line 26: should it be before `for` ???

cristaloleg: this variable is not gonna change anyway, so we're going loop for no reason and spamming with this message (not sure is this the expected behaviour or not).

cristaloleg (approved):

Wondertan (approved):

walldiss (approved):

PR #4190, opened 2025-03-31
Update README.md

leopardracer:


walldiss (commented):
What is the reason for the changes?

cristaloleg:
> rename arabica to mocha

that's a fact not the motivation

Wondertan:
Arabica is a public developer network for people to play around with the node. It's intentionally an arabica here rather then mocha

PR #4191, opened 2025-04-02
chore: fix incorrect logging in `updateHead`

santamasa:
I noticed that `s.networkHead` was being updated **after** the log statement, causing `"from_height"` to always match `"to_height"`.
Now, the old value is stored before updating, ensuring the log accurately reflects the height change.

### DIFF
--- a/das/state.go
+++ b/das/state.go
@@ -146,8 +146,10 @@ func (s *coordinatorState) updateHead(newHead uint64) {
 		log.Infow("found first header, starting sampling")
 	}
 
+	oldHead := s.networkHead
 	s.networkHead = newHead
-	log.Debugw("updated head", "from_height", s.networkHead, "to_height", newHead)
+	
+	log.Debugw("updated head", "from_height", oldHead, "to_height", newHead)
 	s.checkDone()
 }
 


cristaloleg (dismissed):
Nice

renaynay (approved):

santamasa:
> Could you run `make fmt` and `make lint` locally? and commit changes, please.

[cristaloleg](https://github.com/cristaloleg) done.

cristaloleg:
You've added lots of dots? why?

santamasa:
> You've added lots of dots? why?

This is a logical conclusion to the sentence.

cristaloleg:
> This is a logical conclusion to the sentence.

Please revert these dots. 1) that's not the scope of PR, 2) that's adds additional changes in git history for not a big reason.

Thanks.

santamasa:
> > This is a logical conclusion to the sentence.
> 
> Please revert these dots. 1) that's not the scope of PR, 2) that's adds additional changes in git history for not a big reason.
> 
> Thanks.

[cristaloleg](https://github.com/cristaloleg), np! updated.

cristaloleg (approved):

das/state.go
```diff
@@ -146,8 +153,10 @@ func (s *coordinatorState) updateHead(newHead uint64) {
 		log.Infow("found first header, starting sampling")
 	}
 
+	oldHead := s.networkHead
 	s.networkHead = newHead
-	log.Debugw("updated head", "from_height", s.networkHead, "to_height", newHead)
+
+	log.Debugw("updated head", "from_height", oldHead, "to_height", newHead)
```
walldiss, line 159: How about just swap order of lines instead of introducing new var?

walldiss, line 159: Also looks like  keys for values in the log are a bit off. Do you mind to update them for something better while we are on it? Like "old_head" and "new_head" or similar.

walldiss (commented):
Nice find! Would appreciate if you can do small good to haves before we merge

walldiss:
@santamasa Do you mind to take a look at suggested changes?

Wondertan (approved):

PR #4192, opened 2025-04-02
chore: fix formatting

cristaloleg:
I noticed that some files are not formatted with `make fmt`.


walldiss (approved):

nodebuilder/tests/blob_test.go
```diff
@@ -111,7 +111,6 @@ func TestBlobModule(t *testing.T) {
 				assert.Equal(t, v1[0].Commitment, blobV1.Commitment)
 				assert.NotNil(t, blobV1.Signer())
 				assert.Equal(t, blobV1.Signer(), v1[0].Signer())
-
```
Wondertan, line 114: Likely not intentional

Wondertan (approved):

nodebuilder/tests/blob_test.go
cristaloleg: Formatter fixed that. That's why it's committed.

PR #4194, opened 2025-04-03
fix(pruner): change header -> blocks in logs

renaynay:
It would confuse users when we have header pruning

### DIFF
--- a/pruner/service.go
+++ b/pruner/service.go
@@ -117,7 +117,7 @@ func (s *Service) run() {
 
 	lastPrunedHeader, err := s.lastPruned(s.ctx)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pruner/service.go
```diff
@@ -117,7 +117,7 @@ func (s *Service) run() {
 
 	lastPrunedHeader, err := s.lastPruned(s.ctx)
 	if err != nil {
-		log.Errorw("failed to get last pruned header", "height", s.checkpoint.LastPrunedHeight,
+		log.Errorw("failed to get last pruned block", "height", s.checkpoint.LastPrunedHeight,
```
cristaloleg, line 120: But `lastPrunedHeader` is a header

Wondertan (approved):

PR #4197, opened 2025-04-04
fix(rpc/client): handle empty auth token

### LINKED ISSUE #4196: Empty token is not handled in RPC client
This line doesn't properly handle the case of empty token that is useful if node is running with `skip-auth` flag enabled.
https://github.com/celestiaorg/celestia-node/blob/5f486732e68ca670e25a9ae4a145ef8c77cf89d1/api/rpc/client/client.go#L69

This should be handled similar to the logic in `celestia-openrpc`:

https://github.com/celestiaorg/celestia-openrpc/blob/089b92870cb65f8e190ad6d7215b645b942e43b6/client.go#L44-L47

eshaan7:
Fixes #4196.


walldiss (approved):
Nice catch!

renaynay (approved):

PR #4203, opened 2025-04-08
fix(nodebuilder): TLS key/cert name variables

koenmtb1:
Depending on how one generates the TLS certificates it might use different names for the key files.
These are currently hard-coded to `cert.pem` and `key.pem`, but for example letsencrypt by default uses `keychain.pem` which does load correctly into Go TLS.

To prevent having to manually create a symlink using an env variable will resolve that.

One outstanding question that I still have is that if TLS certs are renewed, does the node pick it up automatically? As far as I can tell it currently doesn't, happy to work on that too but a pointer of would be helpful.


cristaloleg:
Please run `make fmt`

smuu (dismissed):

cristaloleg (dismissed):
SGTM

vgonkivs (changes requested):
Generally  LGMT. But it makes me think that we don't need TLS_PATH anymore and we can have `TLS_KEY_PATH` and `TLS_CERT_PATH` instead. It also covers a case when key and certs can be set in different folders.

UPD: we can do much more with the tls stuff in future, so maybe we can have a separate structure in the node's config for it?
Thoughts @renaynay @Wondertan @walldiss @cristaloleg

koenmtb1:
The thought behind keeping `TLS_KEY_PATH` is to not make this a breaking change and retain backwards compatibility, happy to update if deemed helpful.
Happy to update node config for this too, if you think thats a better location.

I've also noticed there's no docs for this (or any requirement for having TLS at all). Is that something worth exploring?

vgonkivs:
> I've also noticed there's no docs for this (or any requirement for having TLS at all). Is that something worth exploring?


do you mean these docs https://docs.celestia.org/how-to-guides/light-node#optional-start-light-node-with-core-endpoint-with-authentication ?

Wondertan:
> do you mean these docs https://docs.celestia.org/how-to-guides/light-node#optional-start-light-node-with-core-endpoint-with-authentication ?

this TLS config is for libp2p's websocket, not for core

Wondertan:
Hey @koenmtb1, aside from this PR, you may wanna take a look at https://blog.libp2p.io/autotls/. We haven't integrated it yet, but it's targeting this particular struggle of managing TLS certs. This is gonna be super helpful for bootstrapper infra by automating cert issuing. We would appreciate if you can take a look into it and help us and other bootstrappers with integration of this tool.

koenmtb1:
> Hey @koenmtb1, aside from this PR, you may wanna take a look at https://blog.libp2p.io/autotls/. We haven't integrated it yet, but it's targeting this particular struggle of managing TLS certs. This is gonna be super helpful for bootstrapper infra by automating cert issuing. We would appreciate if you can take a look into it and help us and other bootstrappers with integration of this tool.

This is interesting, would be open to adding support for this. Probably a separate PR and keep support for self serving certificates?
Any thoughts on using their default `libp2p.direct` or setting up a custom service?

Wondertan:
> Probably a separate PR and keep support for self serving certificates?

Yes

>  Any thoughts on using their default libp2p.direct or setting up a custom service?

We can start with the default one for simplicity

vgonkivs:
> The thought behind keeping TLS_KEY_PATH is to not make this a breaking change and retain backwards compatibility, happy to update if deemed helpful.
Happy to update node config for this too, if you think thats a better location.


Ok, my suggestion is to:
1) add "CELESTIA_TLS_KEY_PATH" and "CELESTIA_TLS_CERT_PATH" as full paths to the key and certs (instead of "NAME") 
2) use "CELESTIA_TLS_PATH" with the default namings if CELESTIA_TLS_KEY_PATH/CELESTIA_TLS_CERT_PATH are empty. 

It will help us to deprecate and remove CELESTIA_TLS_PATH in the future.

PR #4211, opened 2025-04-11
fix!(store/metrics): fix metrics prefix

vgonkivs:
Self-explanatory


renaynay (approved):
Utack

renaynay (commented):
This could be technically breaking btw even if it's a fix

vgonkivs:
I'll add the breaking label then

Wondertan (approved):

cristaloleg (approved):

Wondertan (approved):
wonder if there is a linter which could check this

PR #4212, opened 2025-04-11
feat(p2p): support autotls using IP shipyard

koenmtb1:
Adding support for autotls as discussed in #4203 

Some work to be done
- [ ] Add tests
- [ ] Align on approach
- [ ] Write docs


Wondertan:
Hey @koenmtb1, what's the status of this PR? Do you plan to finish it? We would like to have it, and are facing TLS misconfiguration issues (#4319 ) that your PR could fix once and for all

PR #4219, opened 2025-04-16
deps: Bump SDK to version 0.50 and Celestia App to V4

chatton:
This PR bumps celestia-app to use v4 and the sdk to be on v0.50

Marking as R4R but shouldn't be merged unless we are happy to do a follow up to update celestia-app to a final release tag.

The only failing integration test is one that looks like it is already failing on main.

~~NOTE: this PR is still pinning celestia-app to an in-development build from this PR https://github.com/celestiaorg/celestia-app/pull/4580 so nothing should be merged until there is a tag that includes this fix. This is required for the test code to correctly start the `BlockApi`~~


core/fetcher.go
```diff
@@ -6,11 +6,11 @@ import (
 	"io"
 	"time"
 
+	tmproto "github.com/cometbft/cometbft/proto/tendermint/types"
```
Wondertan, line 9: Should we start directly importing celestia-core instead of replacing comet? If we anyway on changing imports we can resolve a long standing dependency issue. Core is never going to be in paar with comet with core needed changes upstreamed. We then can simply node usage by removing one more annoying replace, as well as allow having core and comet simultaneously in a dependency graph.

chatton: I think this is not possible at the moment because node depends on app, which imports two versions of core.

`github.com/cometbft/cometbft => github.com/celestiaorg/celestia-core v1.52.0-tm-v0.38.17`
`github.com/tendermint/tendermint => github.com/celestiaorg/celestia-core v1.45.0-tm-v0.34.35`

This is required in app in order to make the multiplexer work across different versions of comet (34 and 38)

Wondertan: I see, thanks for explaining. Hopefully after v5 can resolve this

cmd/cel-key/main.go
```diff
@@ -12,11 +12,11 @@ import (
...
 	"github.com/spf13/cobra"
 
-	"github.com/celestiaorg/celestia-app/v3/app"
-	"github.com/celestiaorg/celestia-app/v3/app/encoding"
+	"github.com/celestiaorg/celestia-app/v4/app"
+	"github.com/celestiaorg/celestia-app/v4/app/encoding"
 )
 
-var encodingConfig = encoding.MakeConfig(app.ModuleEncodingRegisters...)
+var encodingConfig = encoding.MakeTestConfig(app.ModuleEncodingRegisters...)
```
chatton, line 19: this might look unusual because of the naming. In `celestia-app` the `TestConfig` is an encoding config that has all types registered (what I assume is desired here) while `MakeConfig` constructs a config that is later updated with the different types based on the sdk app module life cycle functions.

cmd/cel-key/main.go
```diff
@@ -25,11 +25,11 @@ var initClientCtx = client.Context{}.
 	WithLegacyAmino(encodingConfig.Amino).
 	WithInput(os.Stdin).
 	WithAccountRetriever(types.AccountRetriever{}).
-	WithBroadcastMode(flags.BroadcastBlock).
+	WithBroadcastMode(flags.BroadcastSync).
```
chatton, line 28: `BroadcastBlock` has been removed.

core/client_test.go
```diff
@@ -23,8 +23,16 @@ func TestRemoteClient_StartBlockSubscription_And_GetBlock(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second*15)
 	t.Cleanup(cancel)
 
-	client := StartTestNode(t).Client
-	eventChan, err := client.Subscribe(ctx, newBlockSubscriber, newDataSignedBlockQuery)
+	node, wsClient := StartTestNodeWithConfigAndClient(t)
+	client := node.Client
```
chatton, line 27: the websocket client has been separated out from the existing client.

core/eds.go
```diff
@@ -21,19 +20,25 @@ import (
 	"github.com/celestiaorg/celestia-node/store"
 )
 
+// isEmptyBlockRef returns true if the application considers the given block data
+// empty at a given version.
+func isEmptyBlockRef(data *coretypes.Data) bool {
+	return len(data.Txs) == 0
+}
```
chatton, line 27: this function was removed from celestia-app, it's small so I just duplicated it here.

core/eds.go
```diff
@@ -21,19 +20,25 @@ import (
...
 		return share.EmptyEDS(), nil
 	}
 
 	// Construct the data square from the block's transactions
 	square, err := libsquare.Construct(
 		data.Txs.ToSliceOfBytes(),
-		appconsts.SquareSizeUpperBound(appVersion),
-		appconsts.SubtreeRootThreshold(appVersion),
+		appconsts.SquareSizeUpperBound,
+		appconsts.SubtreeRootThreshold,
```
chatton, line 41: since `celestia-app` is always just the `latest` version, the functions that return different values based on version no longer exist, so the constants can be used directly.

core/eds_test.go
```diff
@@ -38,13 +35,14 @@ func TestEmptySquareWithZeroTxs(t *testing.T) {
...
 
-	eds, err := extendBlock(&data, appconsts.LatestVersion)
+	eds, err := extendBlock(&data)
 	require.NoError(t, err)
 	require.True(t, eds.Equals(share.EmptyEDS()))
 
 	// force extend the square using an empty block and compare with the min DAH
-	eds, err = app.ExtendBlock(data, appconsts.LatestVersion)
-	require.NoError(t, err)
+	// TODO(chatton): app.ExtendBlock was removed, can we delete?
```
chatton, line 43: I'm unsure of what to do with this, `app.ExtendBlock` no longer exists.

core/exchange_test.go
```diff
@@ -44,6 +44,9 @@ func TestCoreExchange_RequestHeaders(t *testing.T) {
 	expectedLastHeightInRange := to - 1
 	expectedLenHeaders := to - expectedFirstHeightInRange
 
+	_, err = cctx.WaitForHeightWithTimeout(30, 10*time.Second)
+	require.NoError(t, err)
```
chatton, line 48: I think we need to wait for some blocks due to the `BroadcastModeBlock` being removed.

core/exchange_test.go
```diff
@@ -156,10 +165,12 @@ func TestExchange_StoreHistoricIfArchival(t *testing.T) {
...
 		assert.True(t, has)
 
-		// ensure .q4 file was not stored
+		// ensure .q4 file was not stored if not IsEmptyEDS
+		// TODO(chatton): verify if this is the correct behavior. Does the added WaitForHeight
+		// make it so there are some headers that are not empty and so a different code path is followed?
 		has, err = store.HasQ4ByHash(ctx, h.DAH.Hash())
 		require.NoError(t, err)
-		assert.False(t, has)
+		assert.Equal(t, has, !share.DataHash(h.DataHash).IsEmptyEDS())
```
chatton, line 173: ❗ I want to flag this ❗ 

I don't fully understand the expected behaviour based on test inputs. I don't want to make this change just so the test passes without making sure it is okay.

core/testing.go
```diff
@@ -40,9 +42,9 @@ func DefaultTestConfig() *testnode.Config {
 	tmConfig.Consensus.TimeoutCommit = time.Millisecond * 200
 
 	return testnode.DefaultConfig().
-		WithChainID(chainID).
-		WithFundedAccounts(generateRandomAccounts(10)...). // 10 usually is enough for testing
 		WithGenesis(genesis).
+		WithFundedAccounts(generateRandomAccounts(10)...). // 10 usually is enough for testing
+		WithChainID(chainID).
```
chatton, line 47: this is a super subtle change, the chainID field is stored on the genesis type, so in the old version of the code, it set a chainID on a genesis, the overwrote the genesis. Internally the chainID is used in two different places now and so it must be propagated correctly by assigning the genesis, then modifying the chainID if required.

header/headertest/testing.go
```diff
@@ -125,11 +131,32 @@ func MakeCommit(
...
+		commitSigs[i] = tmproto.CommitSig{
+			BlockIdFlag:      blockIDFlag,
+			ValidatorAddress: vote.ValidatorAddress,
+			Timestamp:        vote.Timestamp,
+			Signature:        vote.Signature,
 		}
 	}
 
-	return voteSet.MakeCommit(), nil
+	protoCommit := &tmproto.Commit{
```
chatton, line 152: `MakeCommit` has been removed.

header/headertest/testing.go
```diff
@@ -219,10 +247,28 @@ func (s *TestSuite) Commit(h *header.RawHeader) *types.Commit {
...
+		sigs[i] = tmproto.CommitSig{
+			BlockIdFlag:      tmproto.BlockIDFlag(commitSig.BlockIDFlag),
+			ValidatorAddress: commitSig.ValidatorAddress,
+			Timestamp:        commitSig.Timestamp,
+			Signature:        commitSig.Signature,
+		}
 	}
 
-	return types.NewCommit(h.Height, round, bid, comms)
+	// Create a proto.Commit manually
```
chatton, line 259: `NewCommit` has also been removed.

nodebuilder/tests/swamp/swamp_tx.go
```diff
@@ -19,7 +21,14 @@ func FillBlocks(ctx context.Context, cctx testnode.Context, account string, bsiz
...
 		for i := 0; i < blocks; i++ {
-			_, err = cctx.FillBlock(bsize, account, flags.BroadcastBlock)
+			var resp *sdk.TxResponse
+			resp, err = cctx.FillBlock(bsize, account, flags.BroadcastSync)
+			if err != nil {
+				break
+			}
+
+			// ensure each tx is included before moving on to avoid account sequence mismatch errors in tests.
+			err = waitForTxResponse(cctx, resp.TxHash, time.Second*1)
```
chatton, line 31: this is another significant change. With `BroadcastBlock` there was no need to wait for anything else, we now need to explicitly wait for the txHash to be included in a block as the `resp` we get back contains the tx hash. We poll until it's included in a block.

state/core_access.go
```diff
@@ -219,7 +222,7 @@ func (ca *CoreAccessor) SubmitPayForBlob(
 		return convertToSdkTxResponse(response), nil
 	}
 	// TODO @renaynay: use new rachid named func
-	if apperrors.IsInsufficientMinGasPrice(err) {
+	if apperrors.IsInsufficientFee(err) {
```
chatton, line 225: just a rename.

go.mod
```diff
@@ -3,19 +3,23 @@ module github.com/celestiaorg/celestia-node
...
-	cosmossdk.io/math v1.5.0
+	cosmossdk.io/log v1.5.1
+	cosmossdk.io/math v1.5.3
+	cosmossdk.io/store v1.1.2
+	cosmossdk.io/x/feegrant v0.1.1
 	github.com/BurntSushi/toml v1.4.1-0.20240526193622-a339e1f7089c
 	github.com/alecthomas/jsonschema v0.0.0-20220216202328-9eeeec9d044b
 	github.com/benbjohnson/clock v1.3.5
-	github.com/celestiaorg/celestia-app/v3 v3.8.0-arabica
+	github.com/celestiaorg/celestia-app/v4 v4.0.0-20250415151309-5138040df8a7
```
chatton, line 13: should not be merged until there is a tag, unless we are okay with a follow up PR to do this.

core/eds_test.go
damiannolan: should be able to just do it in line

```go
dataSquare, err := square.Construct(
	data.Txs.ToSliceOfBytes(),
	appconsts.SquareSizeUpperBound,
	appconsts.SubtreeRootThreshold,
)

eds, err := da.ExtendSquare(share.ToBytes(dataSquare))
```

chatton: that's basically already what happens in the unexported `extendBlock` fn, I would imagine there was an intentional reason to use both in the same test fn 🤔

cmd/cel-key/main.go
damiannolan: You will have to update this to `encoding.MakeConfig(...)` when you update to latest of v4. I PR'd a refactor that managed to get rid of this separate test helper yesterday

chatton: yeah I saw that! Should just be a one liner right?

header/headertest/testing.go
```diff
@@ -111,14 +111,12 @@ func MakeCommit(
 	validators []types.PrivValidator,
 	now time.Time,
 ) (*types.Commit, error) {
-	commitSigs := make([]tmproto.CommitSig, len(validators))
-
-	for i := range validators {
+	// all sign
+	for i := 0; i < len(validators); i++ {
```
cristaloleg, line 115: Just curious what was the point to change `for i := range validators {` to 3-clause `for` ?

chatton: it looks like it's like this one main 🤔 , this might be showing up as a diff since I made a change then changed it back based on @damiannolan mentioning I can use the `MakeExtendedCommit`

rootulp:
Created https://github.com/celestiaorg/celestia-app/releases/tag/v4.0.0-rc0 for usage here

walldiss:
Generally looks good! Need to resolve conflicts to run ci and proceed to tests

evan-forbes (dismissed):

rootulp (approved):

core/exchange_test.go
chatton: Yes the `assert.False(t, has)` fails. It looks like we are hitting [this](https://github.com/celestiaorg/celestia-node/blob/69487249b1e276a4aba58e644e281401acb67924/core/eds.go#L80) codepath for some reason.

I'm wondering if it has something to do with the `cctx.WaitForHeight(30)` I added above. I initially added the wait for height as I encountered this error

```
fetching signed block at height 28 from core: rpc error: code = Unknown desc = nil block meta for height 28
```

from the `ce.GetRangeByHeight(ctx, genHeader, 30)` call.

core/eds.go
evan-forbes: good point, for new blocks it shouldn't but for historical it will as we added authored blobs

can we get this in a flup? out of curiousity, how far do nodes sync?

header/pb/extended_header.pb.go
```diff
@@ -5,9 +5,9 @@ package header_pb
 
 import (
 	fmt "fmt"
-	da "github.com/celestiaorg/celestia-app/v3/proto/celestia/core/v1/da"
+	da "github.com/celestiaorg/celestia-app/v4/proto/celestia/core/v1/da"
 	proto "github.com/gogo/protobuf/proto"
-	types "github.com/tendermint/tendermint/proto/tendermint/types"
+	types "github.com/cometbft/cometbft/proto/tendermint/types"
```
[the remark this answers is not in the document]
chatton: shouldn't cause any problems I don't think since it's just the import being changed and no proto files themselves.

cmd/cel-key/main.go
```diff
@@ -25,11 +25,11 @@ var initClientCtx = client.Context{}.
...
 	WithHomeDir(app.DefaultNodeHome).
 	WithViper("CELESTIA")
 
-var rootCmd = keys.Commands("~")
+var rootCmd = keys.Commands()
```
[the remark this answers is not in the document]
chatton: the keys command doesn't require the directory in sdk 50 I don't have context on why, we can look for the PR it was removed in if you want.

core/eds_test.go
chatton: do you have any suggestions about how to do this? I'm not familiar enough with the code base yet to write this without some help. I can do in a follow up if that's okay?

cmd/cel-key/main.go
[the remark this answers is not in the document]
damiannolan: its added in default keyring flags here: https://github.com/celestiaorg/cosmos-sdk/blob/release/v0.50.x-celestia/client/flags/flags.go#L152-L155

core/exchange_test.go
chatton: @renaynay  So digging into the code a little bit more, it looks like things pass if we just use a smaller number of in the `GetRangeByHeight`. I passed 5 and it works, it seems that if you pass in more (20+) we start entering [this codepath](https://github.com/celestiaorg/celestia-node/blob/87bc6c891949c80d1a68389af45c8bae87588597/store/store.go#L132) i.e. there is a datahash that is not empty.

I'm not sure how to debug this further, do you have any suggestions?

core/exchange_test.go
```diff
@@ -121,6 +121,7 @@ func TestExchange_DoNotStoreHistoric(t *testing.T) {
 func TestExchange_StoreHistoricIfArchival(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
 	t.Cleanup(cancel)
+	t.Setenv("CELESTIA_OVERRIDE_AVAILABILITY_WINDOW", "1ns") // all blocks will be "historic"
```
renaynay, line 124: So this works too but what was not working about the previous approach where we had the `WithAvailabilityWindow(time.Nanosecond)` ?

renaynay: Ah wow yes correct. This test is actually so stupid 😂  Thanks for the catch @chatton !

This test is a false positive. We try to generate non-empty blocks, but non-empty block generation only really begins from height 50 onwards -- we request range (1:30], all are empty blocks, so of course we do not store the Q4. 

This test needs to be completely refactored so that we have a deterministic way to ensure we are querying a range where there are several non-empty blocks included. 

Would you like to do this or should I and I can commit directly to the branch?

chatton: hey @renaynay thanks for verifying, feel free to push straight to the branch! I think you will have a better idea of how to make sure this is testing the correct behaviour.

core/eds.go
rootulp: celestia-node lost the ability to sync from genesis with [this PR](https://github.com/celestiaorg/celestia-node/pull/3782) specifically [here](https://github.com/celestiaorg/celestia-node/pull/3782/files#diff-ee38e3a0bb3c3b606b88f964c0f2732b3308868262112dd8406788deb598b213R33) because blocks produced with celestia-app app version 1 or 2 used go-square v1 and blocks produced with celestia-app app version 3 use go-square v2. If celestia-node wants to support sync from genesis that it needs something like [these lines](https://github.com/celestiaorg/celestia-app/blob/f373e30c395ba3f20df22c7f3a423e0d0208f3ab/app/process_proposal.go#L140-L158) to call go-square v1 for app version 1 and 2 and go-square v2 for app version 3 and above.

IMO adding sync from genesis to celestia-node is a new feature and shouldn't block this PR because this PR preserves the same behavior as celestia-node `main`. In other words, celestia-node `main` will remain compatible with v3 and v4 blocks.

walldiss: Does it mean the node not only doesn't support syncing from genesis, but also doesn't support syncing any v1 or v2 blocks? I'm a bit confused about how the v2 to v3 migration happened then, given that any node trying to sync v2 blocks would fail.

rootulp: > Does it mean the node not only doesn't support syncing from genesis, but also doesn't support syncing any v1 or v2 blocks?

Yes AFAICT current node releases don't support syncing v1 or v2 blocks. 

> I'm a bit confused about how the v2 to v3 migration happened then, given that any node trying to sync v2 blocks would fail.

Good question, I'm not sure. We could try to sync v2 and v3 blocks with https://github.com/celestiaorg/celestia-node/pull/3782 to see what happens.

rootulp: We just had a short call [notes](https://docs.google.com/document/d/1BPIhNc2gqeV0Xx5Hz6S_4chG2QUHcnqw_AqCNoxZyo8/edit?usp=sharing) to discuss this. Our current hypothesis is that _square construction_ did not change between go-square v1 and v2 so although the routing logic exists in celestia-app v3, it seems unnecessary.

If square construction didn't break between go-square v1 and v2, then the implementation in this PR is sufficient for celestia-node to sync historical blocks all the way from genesis to tip.

renaynay (approved):
all fine w me, we will FLUP w app team later re: thing we discussed in sync.

rootulp:
Are ya'll waiting to merge this until after Slava + Sysrex's manual testing of celestia-node <> celestia-app compatibility?

walldiss (approved):

walldiss (approved):

PR #4225, opened 2025-04-21
docs(cmd/rpccli): improve auth token documentation and CLI UX

### LINKED ISSUE #3974: docs(cmd/rpccli): ui/ux improvements
Originally posted by @rach-id 

Rachid has shared his main pain points when interacting with our API/docs:

```
* getting the token docs need to be more appearent in the docs. I always struggle to find the right commands
`Error: cant access the auth token: token/node-store flag was not specified: no opened Node Store found (no node is running)` 

I don't understand what this error means. The bridge node is running but still it complains. Should I run a light/full node? also, if the fix is always the same, would it make sense to add it in the error message?

--url string     Request URL what is this URL? what does it mean?
--token string    Authorization token  I would add hints on how to get it. something like: if you still haven't generated one, use celestia <node_type> admin auth to get it.

account-address             Retrieves the address of the node's account/signer.
  balance                     Retrieves the Celestia coin balance for the node's account/signer and verifies it against the corresponding block's AppHash.
  balance-for-address         Retrieves the Celestia coin balance for the given address and verifies the returned balance against the corresponding block's AppHash.


I don't exacly get what they do, the difference and I guess the verifies it against the corresponding block's AppHash is too specific to be exposed to users.
I guess that setting the CELESTIA_CUSTOM to run against a custom network needs to be documented in a clear place in the docs. I struggled and had to ask node to know how to run it for a custom node. Alternatively, you could provide a CLI flag --trusted.hash -chain.id  that need to be set when setting p2p.network custom
I propose to stick to the same names in comsos-sdk for subcommands and even stick to the tx|query subcommand to organise the subcommands. For example, begin-redelegate can be redelegate simply and even copy the description as it's clear, concise and anyone in the space is already familiar with.
```

I propose to discuss here how we can improve it 🙂

DeVikingMark:
Closes #3974.

Changes:
- Improved flag descriptions for --url and --token flags
- Enhanced error messages with clear instructions on generating tokens
- Simplified command descriptions for better readability
- Renamed begin-redelegate command to redelegate to match Cosmos SDK convention
- Added comprehensive auth token documentation (docs/cli-auth-tokens.md)
- Improved auth command output with usage examples


docs/cli-auth-tokens.md
```diff
@@ -0,0 +1,75 @@
+# Using Authentication Tokens with Celestia Node CLI
```
gupadhyaya, line 1: why not make a PR to update the celestia docs: https://github.com/celestiaorg/docs/blob/main/tutorials/node-tutorial.md#auth-token--auth-token
instead of trying to add a new page here?

nodebuilder/state/cmd/state.go
```diff
@@ -30,7 +30,7 @@ func init() {
 		balanceForAddressCmd,
 		transferCmd,
 		cancelUnbondingDelegationCmd,
-		beginRedelegateCmd,
+		redelegateCmd,
```
gupadhyaya, line 33: i would not combine this change with others. you can make a separate PR.

gupadhyaya:
we plan to make a mass lint and fix of descriptions and other texts throughout the repo, hence closing this. thanks for your attempt to contribute.

PR #4231, opened 2025-04-22
fix(p2p/discovery,core/listener): fix timer deadlock

### LINKED ISSUE #4045: Celestia-bridge stuck 2 times during 24 hours
### Summary of Bug

Hi,

Our Celestia-bridge node has encountered two instances of being stuck within the past 24 hours.

1. First Occurrence Jan 13 at 13:33:27 GMT:

The service logs indicated:
```
"underlying subscription is stuck"
``` 

More logs are available on [gist](https://gist.github.com/Sebby83/26d906be6e9ea79b7d742cc76a5de648).
The only way to restore the service was by deleting the database, after which the Celestia bridge took a couple of hours to resync.

2. Second Occurrence:

This issue started again today at approximately 20:15 GMT.

Common log messages included:

```
    "listener: subscriber error, resubscribing..."
    "server: request height not found"
``` 

The full logs for this event are available [gist](https://gist.github.com/Sebby83/957d519a9ea2d46efcb9524faf348767).

We've ruled out an RPC issue, as the node successfully received block height updates.
Proof of this is available [here](https://gist.github.com/Sebby83/544c2a74e758dffb9bab6c50f08f28ab)

We suspect this may be related to database corruption, as the server has sufficient free space on the NVMe drive and the storage appears to be healthy.

Additional Context:
We made a configuration change two days ago, increasing the max receive message size with the following command, but we are not sure if this is related:

```
sed -i.bak 's/^max-recv-msg-size = "10485760"/max-recv-msg-size = "20971520"/' ~/.celestia-app/config/app.toml
``` 

Let us know if you need further information 

### Version

Semantic version: v0.20.4
Commit: 51b79431533576a5cddf1235bdf3751e4c014212
Build Date: Thu Dec 19 19:25:05 GMT 2024
System version: amd64/linux
Golang version: go1.23.3

### Steps to Reproduce

The issue manifests intermittently, and unfortunately, I was unable to reproduce it consistently.

### LINKED ISSUE #4228: bug(p2p/discovery): DHT Discovery unable to find all nodes
### Implementation ideas

**Description:**  
Network crawler scripts indicate that DHT discovery is finding only ~13-14 nodes. When investigated locally, my node discovered even fewer (7-8 nodes). However, Shrexsub notifications (only from full or bridge nodes) suggest there are around 100 active full/bridge nodes. This discrepancy clearly indicates an issue with DHT announcements. Additionally, the advertised peer count per DHT namespace seems incomplete.

**Expected behavior:**
DHT announcements should reflect the actual number of connected full/bridge nodes.

**Current behavior:**
- Network crawler scripts report ~13-14 nodes.
- Local tests find 7-8 nodes.
- Shrexsub metrics show approximately 100 active nodes.

**Action:**
Investigate why DHT announcements fail to reflect the full network state and ensure accuracy in peer advertising.

walldiss:
**Go 1.23 Timer Behavior Change**

Go 1.23 introduced changes to the behavior of `Timer` that affect how channels should be handled:

1. **No need to drain the channel before `Reset`**  
   ```go
   // For a chan-based timer created with NewTimer, as of Go 1.23,
   // any receive from t.C after Reset has returned is guaranteed not
   // to receive a time value corresponding to the previous timer settings;
   // if the program has not received from t.C already and the timer is
   // running, Reset is guaranteed to return true.
   // Before Go 1.23, the only safe way to use Reset was to [Stop] and
   // explicitly drain the timer first.
   ```

2. **Draining after `Stop` will never read a value**  
   ```go
   // For a chan-based timer created with NewTimer(d), as of Go 1.23,
   // any receive from t.C after Stop has returned is guaranteed to block
   // rather than receive a stale time value from before the Stop;
   // if the program has not received from t.C already and the timer is
   // running, Stop is guaranteed to return true.
   // Before Go 1.23, the only safe way to use Stop was to insert an extra
   // <-t.C if Stop returned false to drain a potential stale value.
   // See the [NewTimer] documentation for more details.
   ```

These changes caused our code to enter a deadlock state when reading from the channel, which in turn prevented the discovery module from continuing announcements. The same issue also affected the core listener, which could potentially deadlock under similar circumstances.

In particular, If `Timer.Stop` is called after the `Timer` expires, we deadlock. This can happen if advertisement call or new block handling takes _longer_ then the set deadline on the `Timer`, both are rare but possible case in production environment.

Fixes #4228
Fixes #4045

### DIFF
--- a/share/shwap/p2p/discovery/discovery.go
+++ b/share/shwap/p2p/discovery/discovery.go
@@ -189,24 +189,16 @@ func (d *Discovery) Advertise(ctx context.Context) {
 
 			// we don't want retry indefinitely in busy loop
 			// internal discovery mechanism may need some time before attempts
-			errTimer := time.NewTimer(retryTimeout)
+			errTimer := time.After(retryTimeout)
 			select {
-			case <-errTimer.C:
-				errTimer.Stop()
-				if !timer.Stop() {
-					<-timer.C
-				}
+			case <-errTimer:
 				continue
 			case <-ctx.Done():
-				errTimer.Stop()
 				return
 			}
 		}
 
 		log.Infof("successfully advertised to topic %s", d.topic)
-		if !timer.Stop() {
-			<-timer.C
-		}
 		timer.Reset(d.params.AdvertiseInterval)
 		select {
 		case <-timer.C:
--- a/core/listener.go
+++ b/core/listener.go
@@ -156,10 +156,6 @@ func (cl *Listener) listen(ctx context.Context, sub <-chan types.EventDataSigned
 					"hash", b.Header.Hash().String(),
 					"err", err)
 			}
-
-			if !timeout.Stop() {
-				<-timeout.C
-			}
 		case <-timeout.C:
 			cl.metrics.subscriptionStuck(ctx)
 			log.Error("underlying subscription is stuck")


renaynay (approved):

share/shwap/p2p/discovery/discovery.go
```diff
@@ -189,24 +189,16 @@ func (d *Discovery) Advertise(ctx context.Context) {
...
-				errTimer.Stop()
 				return
 			}
 		}
 
 		log.Infof("successfully advertised to topic %s", d.topic)
-		if !timer.Stop() {
-			<-timer.C
-		}
 		timer.Reset(d.params.AdvertiseInterval)
```
Wondertan, line 202: Basing off 
```
// if the program has not received from t.C already and the timer is
// running, Stop is guaranteed to return true.
```

I think it should never get into the if case. I also believe they wouldn't break all the millions of loc written that were handling for this timer edge case this way.

cristaloleg (approved):

Wondertan (commented):
Reading the timer API changes I am not convinced this actually fixes it. Yes, we can remove some code not to handle channel draining post Stop, but unless I misread, the deadlock shouldn't ever

share/shwap/p2p/discovery/discovery.go
walldiss: It will get to this case if timer didn't fire by the time Stop() was called.

walldiss:
The issue was that `Stop()` now always drains the channel. So, reading from the channel *after* calling `Stop()` results in a deadlock. 

Additionally, `Reset()` no longer requires `Stop()` to be called beforehand. It seems like these two changes together may have caused some confusion.

To clarify, the change involves two key updates:
1. **Remove draining the channel**—it's no longer necessary after calling `Stop()`.  
2. **Remove unnecessary `Stop()` before `Reset()`**—`Reset()` now handles draining the channel on its own.

Wondertan:
> 
    1. Remove draining the channel—it's no longer necessary after calling Stop().
    2. Remove unnecessary Stop() before Reset()—Reset() now handles draining the channel on its own.

I get how this two are no longer necessary. The thing I still struggle to grasp is how the deadlock can happen here. I made a bet that the issue is somewhere in the Timer lol and it looks its the case, but I can't believe go team would just break us this way with all the forward compatibility promises.

walldiss:
This line from the golang docs (listed in Pr description):
```
any receive from t.C after Stop has returned is guaranteed to block
```

Wondertan:
> any receive from t.C after Stop has returned is guaranteed to block

Sure, but t.C is only read from only if Stop returns false and basically the question when can this happen

share/shwap/p2p/discovery/discovery.go
Wondertan: I think its the opposite actually:
>  It returns true if the call stops the timer, false if the timer has already
 expired or been stopped.
 

So in case `d.disc.Advertise(ctx, d.topic)` call takes longer than timer, meaning the timer ha fired by the time the call finishes.

Wondertan (approved):
Approve basing on this thread https://github.com/celestiaorg/celestia-node/pull/4231/files#r2053999292

Also, I would clarify in the description with the following summary:
"If `Timer.Stop` is called after the `Timer` expires, we deadlock post go1.23. This can happen if advertisement or new block handling takes _longer_ then the set deadline on the timer."

Kinda worrying to be betrayed by the go team this way

share/shwap/p2p/discovery/discovery.go
walldiss: You right on correction for comment. But It was in PR description from the beginning:
```
In particular, If Timer.Stop is called after the Timer expires, we deadlock. This can happen if advertisement call or new block handling takes longer then the set deadline on the Timer, both are rare but possible case in production environment.
```

Wondertan: Dude, that's the line I added, see the edit history of the PR

PR #4238, opened 2025-04-28
refactor: use slices.Clone

careworry:
replacing append([]T(nil), s...) by slices.Clone(s) added in go1.21.  https://pkg.go.dev/slices#Clone


cristaloleg:
Will be addressed in https://github.com/celestiaorg/celestia-node/pull/4239

PR #4244, opened 2025-04-28
fix(core): `storeEDS` respects window param

renaynay:
This fixes a bug where `storeEDS` would bypass the check to see if the node is a pruned node that it should not store historic blocks. While this isn't critical, because pruner would clean them up eventually anyway, this helps avoid unnecessary writes during the syncing process for pruned BNs.

Thanks to @chatton for unlocking this find!

### DIFF
--- a/core/option.go
+++ b/core/option.go
@@ -4,6 +4,7 @@ import (
 	"time"
 
 	"github.com/celestiaorg/celestia-node/nodebuilder/p2p"
+	"github.com/celestiaorg/celestia-node/share/availability"
 )
 
 type Option func(*params)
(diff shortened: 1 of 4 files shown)


core/option.go
```diff
@@ -17,7 +18,7 @@ type params struct {
 
 func defaultParams() params {
 	return params{
-		availabilityWindow: time.Duration(0),
+		availabilityWindow: availability.StorageWindow,
```
renaynay, line 21: this is the most important change - we were not passing the availability window properly to the listener + to the core fetcher, making this bug possible.

Wondertan (approved):
Nice find

walldiss (approved):

cristaloleg (approved):

Issue #4218, opened 2025-04-15
feat(p2p/shrex): Generalize and Unify the Shrex Protocol


walldiss:
### Summary

The current `shrex` protocol implementation in Celestia Node supports direct P2P downloads for specific data types like EDS and `NamespaceData`. However, each data type requires implementing its own client and server logic, which leads to duplicated effort, disparities in functionality, and increased likelihood of bugs.

This issue proposes generalizing and unifying the `shrex` protocol into a single flexible mechanism that can support all data types conforming to a well-defined interface. The result will be a streamlined and extensible P2P download protocol.

---

### 📌 Motivation

- **Eliminate Redundant Code**: Avoid having to write a new client/server for each data type.
- **Consistency**: Reduces implementation disparity and corresponding bugs.
- **Extensibility**: Simplifies support for new data types in the future.
- **Modularity**: Moves shared logic (peer discovery, scoring, retry) out of `shrex` to avoid tight coupling.

---

#### Client API

Introduce a generic client method:

```go
shrex.Get(id id, c container, peerID peer.ID) error
```

- `id`: identifies the data being requested (e.g., EDSID, NamespaceDataID, SampleID etc)
- `container`: the target for streamed data (e.g., an EDS container or Sample etc)
- `peerID`: target peer to fetch the data from

Under the hood, `Get()` will:
- Serialize the `id` and send it as a request
- Receive a stream of data into the `container`
- Abstract all data-specific logic behind interfaces

---

#### ID and Container Interfaces

These interfaces would allow shrex to be type-agnostic:

```go
type id interface {
	Name() string                      // Unique name for route registration in server constructor
	WriteTo(io.Writer) (int64, error)  // Encode request
	ReadFrom(io.Reader) (int64, error) // Decode on server
	Validate() error                   // Validation logic
	Reader(acc eds.AccessorStreamer) io.Reader // Stream from storage
}

type container interface {
	io.ReaderFrom // Target for incoming streamed data
}
```

`shwap.ID` and `shwap.Container` types will need minor adjustments to satisfy these interfaces.

---

#### Server Handler Registration

Servers should be able to register handler logic per `shwap` type:

```go
RegisterHandlers([]id)
```

- By default, a server supports all known `shwap` types.
- If a user specifies a subset, only matching endpoints will be active.
- Enables fine-grained control over protocol exposure.

---

### 🔧 Refactoring Plan

- [ ] Abstract current shrex clients and servers behind interfaces.
[...]

PR #4249, opened 2025-04-29
shwap!(shrex): unify shrex-nd and shrex-eds

vgonkivs:
This PR unifies shrex client and server to support all kinds of registered requests. The PR breaks the node's config, dividing shrex section into `Client` and `Server`. 

Closes https://github.com/celestiaorg/celestia-node/issues/4218

The PR brings multiple breaks node's metrics and config

Metrics:
* `shrex_eds_client_total_requests`/`shrex_nd_client_total_requests` were changed to `shrex_client_total_requests`;
* `shrex_eds_server_total_requests`/`shrex_nd_server_total_requests` were changed to `shrex_server_total_requests`;
Newly added shrex metrics have an attribute "protocol" that specifies the request name.

Config:
* ShrexEDSParams/ ShrexNDParams were changed to ShrexClient/ShrexServer;

```go
type ClientParams struct {
	// ReadTimeout sets the timeout for reading messages from the stream.
	ReadTimeout time.Duration

	// WriteTimeout sets the timeout for writing messages to the stream.
	WriteTimeout time.Duration

	// networkID is prepended to the protocolID and represents the network the protocol is
	// running on.
	networkID string
}

type ServerParams struct {
	// ReadTimeout sets the timeout for reading messages from the stream.
	ReadTimeout time.Duration

	// WriteTimeout sets the timeout for writing messages to the stream.
	WriteTimeout time.Duration

	// networkID is prepended to the protocolID and represents the network the protocol is
	// running on.
	networkID string

	// HandleRequestTimeout defines the deadline for handling request.
	HandleRequestTimeout time.Duration

	// ConcurrencyLimit is the maximum number of concurrently handled streams
	ConcurrencyLimit int
}
```

client.ReadTimeout default value was changed to 2 Minutes(server.WriteTimeout+server.HandleTimeout)


share/shwap/row.go
```diff
@@ -240,3 +240,70 @@ func toRowSide(s string) RowSide {
 		panic("invalid row side")
 	}
 }
+
+var codec = share.DefaultRSMT2DCodec()
+
```
walldiss, line 245: Any changes or just copy? Can't see the diff in gh

share/shwap/p2p/shrex/shrexnd/id.go
```diff
@@ -0,0 +1,52 @@
...
+	io.WriterTo
+	io.ReaderFrom
+
+	Name() string
+	// Target reports the target height of the shwap container
+	Target() uint64
+	// Validate performs a basic validation of the request.
+	Validate() error
+
+	// FetchContainerReader gets Accessor and wrap all requested data with `io.Reader`
```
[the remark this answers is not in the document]
vgonkivs: ContainerDataReader sounds good to me

nodebuilder/config_test.go
```diff
@@ -97,17 +97,12 @@ var outdatedConfig = `
...
-    ServerWriteTimeout = "1m0s"
-    HandleRequestTimeout = "1m0s"
-    ConcurrencyLimit = 10
-    BufferSize = 32768
   [Share.ShrExNDParams]
     ServerReadTimeout = "5s"
     ServerWriteTimeout = "2m35s"
     HandleRequestTimeout = "1m0s"
     ConcurrencyLimit = 10
+ 	BufferSize = 32768
```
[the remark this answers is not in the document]
vgonkivs: buffer size is not needed at all now. Since we are using io.Copy that does `stream.ReadFrom(eds)` under the hood

share/shwap/p2p/shrex/shrexnd/server.go
```diff
@@ -39,158 +38,169 @@ type Server struct {
...
 	if err != nil {
 		logger.Warnw("closing read side of the stream", "err", err)
 	}
 
-	return ndid, nil
+	return nil
 }
 
-func (srv *Server) getNamespaceData(
+func (srv *Server) getData(
```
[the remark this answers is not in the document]
vgonkivs: it still needs to be extracted since there is a general error and status handling logic after it(since get data performs 2 calls that can lead to errors)

share/shwap/p2p/shrex/shrexnd/server.go
```diff
@@ -39,158 +38,169 @@ type Server struct {
...
-	srv.handler = withRecovery
-	return srv, nil
+	return &Server{
+		ctx:                ctx,
+		cancel:             cancel,
+		store:              store,
+		host:               host,
+		params:             params,
+		supportedProtocols: supportedProtocols,
+		middleware:         shrex.NewMiddleware(params.ConcurrencyLimit),
```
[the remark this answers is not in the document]
walldiss: I meant that it should support:
- no protocol provided by user. In this case use default protocols list
- incorrect protocol name provided to constrator should validated and failed during construction time. Right now validation happens in Start() call

share/shwap/row.go
walldiss: Slightly prefer to keep it as separate file as it used to be.

vgonkivs: shwap folder contains containers so I'm afraid that the axis half could be treated as a separate container.

walldiss: Right now It is in row container file, which is more confusing and way harder to find

share/shwap/p2p/shrex/shrexnd/server.go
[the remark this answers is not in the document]
vgonkivs: unwrapping both will make code harder imo, bc the logic of status responding is valid for both calls inside get data(in case of an error). So, I'd leave in separate method and we can refactor it in the future if needed

share/shwap/p2p/shrex/server.go
```diff
@@ -0,0 +1,239 @@
...
+		logger.Warnw("closing read side of the stream", "err", err)
+	}
+
+	return nil
+}
+
+func (srv *Server) getData(
+	ctx context.Context,
+	id id,
+) (io.Reader, shrexpb.Status, error) {
```
walldiss, line 178: Imo we should remove this func. It makes it complicated to track down what error is returned and what is observed

share/shwap/p2p/shrex/shrex_getter/shrex.go
```diff
@@ -247,6 +259,12 @@ func (sg *Getter) GetNamespaceData(
 		return shwap.NamespaceData{}, nil
 	}
 
+	request, err := shwap.NewNamespaceDataID(header.Height(), namespace)
+	if err != nil {
+		return nil, err
+	}
+	response := shwap.NamespaceData{}
```
walldiss, line 266: better be pointer? &shwap.NamespaceData{}

share/shwap/p2p/shrex/shrex_getter/shrex.go
```diff
@@ -187,13 +191,21 @@ func (sg *Getter) GetEDS(ctx context.Context, header *header.ExtendedHeader) (*r
 
 		reqStart := time.Now()
 		reqCtx, cancel := utils.CtxWithSplitTimeout(ctx, sg.minAttemptsCount-attempt+1, sg.minRequestTimeout)
-		eds, getErr := sg.edsClient.RequestEDS(reqCtx, header.DAH, header.Height(), peer)
+		getErr = sg.ndClient.Get(reqCtx, &request, response, peer)
 		cancel()
 		switch {
```
walldiss, line 196: `&request` why we need pointer here? Would it work jsut fine with value because all `id` have value receiver methods?

walldiss (commented):
Massive rework! 🚀  There are bunch of nits here and there

share/shwap/p2p/shrex/id.go
```diff
@@ -0,0 +1,56 @@
...
+	Validate() error
+
+	// ContainerDataReader returns io.Reader that reads data from the Accessor.
+	ContainerDataReader(ctx context.Context, acc shwap.Accessor) (io.Reader, error)
+}
+
+// represents compatible generalised interface type for shwap responses
+type container interface {
+	io.ReaderFrom
+}
```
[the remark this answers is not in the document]
vgonkivs: I also thought about it but decided to leave it for the consistency.

share/shwap/p2p/shrex/shrex_getter/shrex.go
vgonkivs: its a type alias of `[]RowNamespaceData`

share/shwap/p2p/shrex/shrex_getter/shrex.go
vgonkivs: Since IDs have both pointer and value receivers. Receviers will be fixed subsequently.

vgonkivs: `NewEDSID{}` creates a raw value and not a pointer.

Wondertan:
Lint borked

share/shwap/p2p/shrex/request.go
```diff
@@ -0,0 +1,57 @@
...
+	req := newRequestID()
+	protocolNames := make([]string, 0, len(req))
+	for name := range maps.Keys(req) {
+		protocolNames = append(protocolNames, name)
+	}
+	return protocolNames
+}
+
+// request represents compatible generalised interface type for shwap requests.
+type request interface {
```
Wondertan: I believe this should be:
* public
* defined in shwap pkg as ID interface
  * could be additionally aliased/defined as new Request type in shrex pkg

The comment on the interface hints at the semantic issue I want to raise. The comment says "shwap request", but it's shrex request. The problem here is that we are introducing ambiguity into our terms, and we are even starting to confuse them. 

Bitswap composition defines a Block interface, too, but there is a difference. Block combines both the ID and the Container.  The composition also defines the respective block types for each ID/Container pair.

The solution here is to define everything **once** in the shwap pkg as ID and Container interfaces, and then have shrex import it. This way, we ensure the initial intended separation, with shwap defining just the data structures and shrex defining the actual protocol communication rules among peers for shwap data structures.

This comment points to a violation of semantics and thus it's purely semantical, so I don't want us to block on it, but I would love to make things more straightforward.

share/shwap/p2p/shrex/request.go
```diff
@@ -0,0 +1,57 @@
...
+	io.ReaderFrom
+
+	Name() string
+	// Height reports the target height of the shwap container.
+	Height() uint64
+	// Validate performs a basic validation of the request.
+	Validate() error
+
+	// ContainerDataReader returns io.Reader that reads data from the Accessor.
+	ContainerDataReader(ctx context.Context, acc shwap.Accessor) (io.Reader, error)
```
Wondertan: This builds on top of the previous comment. If this is the ContainerReader, then the response type should be called Container, not response.

share/shwap/p2p/shrex/request.go
```diff
@@ -0,0 +1,57 @@
...
+
+	request := make(map[string]newRequest)
+	request[nsDataInitID().Name()] = nsDataInitID
+	request[edsInitID().Name()] = edsInitID
+	return request
+}
+
+// SupportedProtocols returns  a slice of protocol names that
+// the client and server support by default.
+func SupportedProtocols() []string {
```
Wondertan: non-blocking: If we refer to those strings as protocol, it would be clearer to use the specific protocol type for them

share/shwap/p2p/shrex/metrics.go
```diff
@@ -11,63 +11,92 @@ import (
...
+			attribute.String("request.name", requestName),
+		))
 }
 
-func InitClientMetrics(protocol string) (*Metrics, error) {
+func InitClientMetrics() (*Metrics, error) {
 	totalRequestCounter, err := meter.Int64Counter(
-		fmt.Sprintf("shrex_%s_client_total_requests", protocol),
-		metric.WithDescription(fmt.Sprintf("Total count of sent shrex/%s requests", protocol)),
+		"shrex_client_total_requests",
```
Wondertan, line 70: Making sure that we are aware this can break dashboards

share/shwap/p2p/shrex/client.go
```diff
@@ -0,0 +1,174 @@
...
+
+	switch statusResp.Status {
+	case shrexpb.Status_OK:
+	case shrexpb.Status_NOT_FOUND:
+		return statusNotFound, ErrNotFound
+	case shrexpb.Status_INTERNAL:
+		return statusInternalErr, ErrInternalServer
+	default:
+		return statusReadRespErr, ErrInvalidResponse
+	}
```
Wondertan, line 117: One important thing to note. Libp2p now has [error codes](https://github.com/libp2p/go-libp2p/blob/fb1d9512e87309cbbe0667d4f1dd130b9ed66e67/core/network/stream.go#L31-L33) that are guaranteed to be reliably delivered. As we're on breaking shrex, I believe it's a good opportunity to take advantage of this feature. It has been well tested and is currently being used in other production protocols.

We also have some weirdness with error codes that needs to be cleaned up, e.g. EOF is used to signal rate limiting, which is usually a signal of successful read.

share/shwap/namespace_data_id.go
```diff
@@ -130,3 +138,30 @@ func (ndid NamespaceDataID) AppendBinary(data []byte) ([]byte, error) {
 	}
 	return append(data, ndid.DataNamespace.Bytes()...), nil
 }
+
+func (ndid NamespaceDataID) ContainerDataReader(ctx context.Context, acc Accessor) (io.Reader, error) {
```
Wondertan: performance: while  returning reader works, it forces the creation of this buffer that copies data for no good reason. Instead we could receive a Writer(stream) that we write directly to. For the of eds it's gonna be a simple `io.Copy(acc.Reader(), stream)`

share/shwap/p2p/shrex/client.go
```diff
@@ -0,0 +1,174 @@
...
+func (c *Client) Get(
+	ctx context.Context,
+	req request,
+	resp response,
+	peer peer.ID,
+) error {
+	requestTime := time.Now()
+	status, err := c.doRequest(ctx, req, resp, peer)
+	if err != nil {
+		log.Warnw("shrex/client: requesting data from peer failed",
```
Wondertan: nit: make it "source": "client" instead or make server follow the same pattern

share/shwap/axis_half.go
```diff
@@ -1,12 +1,11 @@
-package eds
+package shwap
```
Wondertan, line 1: AxisHalf is not technicall a shwap type, as it has no network representation, but i am ok with moving it if helps with cycle imports or similar

share/shwap/accessor.go
```diff
@@ -0,0 +1,17 @@
...
+	"context"
+	"io"
+
+	libshare "github.com/celestiaorg/go-square/v2/share"
+
+	"github.com/celestiaorg/celestia-node/share"
+)
+
+// Accessor is used to access the data from the shwap containers.
+type Accessor interface {
```
Wondertan, line 13: Can you explain the reason to define another accessor interface?

share/shwap/p2p/shrex/request.go
vgonkivs: it has explicitly been added inside the shrex protocol as I want to keep it private and make it shrex-specific types. Re comment: this is a relic from the previous version.  The logic is pretty straightforward: shrex now works with any type of requests and responses that implement its private interfaces.

share/shwap/accessor.go
vgonkivs: it also helped to avoid dependency hell. I've moved only a part that is needed for the `ContainerDataReader`, but the idea is to move the whole interface under the shwap package(in a subsuquent PR).

share/shwap/namespace_data_id.go
vgonkivs: You are right. Sorry, I havent added a comment, explaining why I did this. Once GetRange+DataRootPRoofs will be merged, we can move from RowNamespaceData as a basic underlying type to RangeNamespaceData(that will implement a reader interface). I've opened one more PR - enabling GetSamples/GetRow over shrex, where I've already implemented a reader interface for both types.

vgonkivs: So, this is a simple and fast solution to make it work.

share/shwap/p2p/shrex/client.go
vgonkivs: I will take a look. Thanks

share/shwap/p2p/shrex/metrics.go
vgonkivs: I will update all dashboards during the tests on the private network.

share/shwap/p2p/shrex/request.go
vgonkivs: in case we'll agree on making the response a private interface, then it won't work

share/shwap/p2p/shrex/request.go
Wondertan: > The logic is pretty straightforward: shrex now works with any type of requests and responses that implement its private interfaces.

Do you see there will be other usages besides shwap types?

vgonkivs: everything is possible in this world

Wondertan: Well, that's not the best argument to introduce confusion between the shrex and shwap layers.

vgonkivs: It was a joke. But to be serious, I don't want them to explicitly interact with each other. I mean, I don't want shrex know about any shwap related functionality. That's why I renamed it from `id` to `request` and from `container` to `response`

walldiss: The  `request` / `response` interfaces are local to consumer (shrex) and does not need to be exposed. They are defined by local implementation of shrex client/server only. There is no need to define interfaces on provider (shwap) side.

share/shwap/p2p/shrex/client.go
walldiss: This sounds like a nice feature, please create an issue for future to group it with shrex breaking release. We should not break shrex in this PR. And there are no plans to break shrex in near future

share/shwap/p2p/shrex/request.go
Wondertan: I understand that, its just semantically incorrect

share/shwap/p2p/shrex/client.go
Wondertan: How is this PR not breaking?

share/shwap/namespace_data_id.go
walldiss: Serde lib has limitation that does not allow reader to be created with prefixed bytes. It requires to pass data through serder.Writer to make it prefixed and creates unwanted copy in the process. Which indeed not optimal for performance

Extra copy is not great. The limitation of Serde should be solved in follow up PR. Please leave it in the issue

share/shwap/p2p/shrex/request.go
Wondertan: Whats the goal with keeping it private?

share/shwap/namespace_data_id.go
Wondertan: This doesn't require to be working through serde and thus shouldn't be neccesary blocked on it

share/shwap/p2p/shrex/request.go
vgonkivs: https://github.com/celestiaorg/celestia-node/pull/4249#discussion_r2132100502

Wondertan: "I want" is not a good reason

vgonkivs: @walldiss explained a motivation why we should keep it private. `ContainerDataReader` will be renamed.

vgonkivs: > "I want" is not a good reason

Where do you see "I want" in this thread?

vgonkivs: > ContainerDataReader will be renamed.

The suggested naming is `ResponseReader`/`DataReader`. Is it semantically correct now?

share/shwap/p2p/shrex/client.go
vgonkivs: > We should not break shrex in this PR

It won't break. There is a note somewhere in the code that it behaves as `Reset()` in case client does not support it

vgonkivs: looks like it will anyway lead to breaking the shrex, which we did not want to do in current PR. mock net(that we are actively using for unit tests does not support these codes, so it will take some time to also adapt our tests)

nodebuilder/config_test.go
```diff
@@ -97,15 +97,12 @@ var outdatedConfig = `
...
-    ServerReadTimeout = "5s"
-    ServerWriteTimeout = "1m0s"
-    HandleRequestTimeout = "1m0s"
-    ConcurrencyLimit = 10
-    BufferSize = 32768
-  [Share.ShrExNDParams]
-    ServerReadTimeout = "5s"
-    ServerWriteTimeout = "2m35s"
+ [Share.ShrexClient]
+    ReadTimeout = "5s"
```
renaynay: Reverse these defaults please, client-side reads should be longer than client-side writes

nodebuilder/share/module.go
```diff
@@ -185,32 +174,17 @@ func shrexComponents(tp node.Type, cfg *Config) fx.Option {
...
 		fx.Provide(fx.Annotate(
 			func(
 				host host.Host,
 				store *store.Store,
 				network modp2p.Network,
-			) (*shrexnd.Server, error) {
-				cfg.ShrExNDParams.WithNetworkID(network.String())
-				return shrexnd.NewServer(cfg.ShrExNDParams, host, store)
+			) (*shrex.Server, error) {
+				cfg.ShrexServer.WithNetworkID(network.String())
```
renaynay, line 184: It's weird that some components have options on the struct and some on the config, I feel like we should clean that up. I prefer passing options to constructor explicitly.

share/shwap/p2p/shrex/shrex_getter/shrex.go
```diff
@@ -187,13 +191,21 @@ func (sg *Getter) GetEDS(ctx context.Context, header *header.ExtendedHeader) (*r
...
 		case getErr == nil:
 			setStatus(peers.ResultNoop)
 			sg.metrics.recordEDSAttempt(ctx, attempt, true)
-			return eds, nil
+			eds, buildErr := eds.ReadAccessor(ctx, response, header.DAH)
+			if buildErr != nil {
+				getErr = buildErr
+				setStatus(peers.ResultBlacklistPeer)
```
renaynay, line 203: can we please not blacklist any peers in this iteration? Let's introduce that when we can prove we do not get false positives.

share/shwap/eds_id.go
```diff
@@ -12,13 +13,13 @@ const EdsIDSize = 8
 // EdsID represents a unique identifier for a row, using the height of the block
 // to identify the data square in the chain.
 type EdsID struct {
-	Height uint64 // Height specifies the block height.
+	height uint64 // Height specifies the block height.
```
renaynay, line 16: what was the point of doing this in this PR?

share/shwap/p2p/shrex/request.go
```diff
@@ -0,0 +1,63 @@
...
+	io.ReaderFrom
+
+	Name() string
+	// Height reports the target height of the shwap container.
+	Height() uint64
+	// Validate performs a basic validation of the request.
+	Validate() error
+
+	// DataReader returns io.Reader that reads data from the Accessor.
+	DataReader(ctx context.Context, acc shwap.Accessor) (io.Reader, error)
```
renaynay: I would want this to be `ResponseReader` to make it exactly clear

share/shwap/p2p/shrex/request.go
```diff
@@ -0,0 +1,63 @@
...
+
+type SupportedProtocolName string
+
+func (s SupportedProtocolName) String() string {
+	return string(s)
+}
+
+// SupportedProtocols returns  a slice of protocol names that
+// the client and server support by default.
+func SupportedProtocols() []SupportedProtocolName {
```
renaynay: why is this a function? Can we not have this be a var that's initialised on start?

Also `newRequestID` actually returns more like a "registry" of what kind of requests are supported at runtime. Can we clean up these names and make them more clear what they're doing?

share/shwap/p2p/shrex/request.go
```diff
@@ -0,0 +1,63 @@
...
+	"context"
+	"io"
+	"maps"
+
+	"github.com/celestiaorg/celestia-node/share/shwap"
+)
+
+type newRequest func() request
+
+func newRequestID() map[string]newRequest {
```
renaynay: refer to registry comment below

share/shwap/p2p/shrex/server.go
```diff
@@ -0,0 +1,215 @@
...
+			return nil, fmt.Errorf("shrex/server: %w: %s", ErrUnsupportedProtocol,
+				ProtocolID(params.NetworkID(), protocolName.String()),
+			)
+		}
+		handler := srv.streamHandler(srv.ctx, id)
+		withRateLimit := srv.middleware.rateLimitHandler(ctx, handler, id().Name())
+		withRecovery := RecoveryMiddleware(withRateLimit)
+		srv.SetHandler(ProtocolID(srv.params.NetworkID(), protocolName.String()), withRecovery)
+	}
+	return srv, nil
```
renaynay: Why are we allowing custom supported protocols here if you'd still need to hardcode the request ids in `newRequestID` such that they actually exist there to be supported during runtime? It doesn't make sense to me right now as a feature. Can we introduce this later on in a more extensible way? I don't think this needs to be a feature we support for this iteration.

share/shwap/p2p/shrex/params.go
```diff
@@ -4,53 +4,85 @@ import (
...
-		ServerWriteTimeout:   time.Minute, // based on max observed sample time for 256 blocks (~50s)
+func DefaultServerParameters() *ServerParams {
+	return &ServerParams{
+		Parameters:           defaultParameters(),
 		HandleRequestTimeout: time.Minute,
 		ConcurrencyLimit:     10,
 	}
 }
 
+func defaultParameters() *Parameters {
```
renaynay: these defaults don't apply to both client and server - client should have reverse defaults of server

client's read deadline == server's write deadline and vice versa

share/shwap/p2p/shrex/client.go
```diff
@@ -0,0 +1,175 @@
...
+	// if deadline not set, client write deadline defaults to server read deadline
+	if c.params.ReadTimeout != 0 {
+		err := stream.SetWriteDeadline(time.Now().Add(c.params.ReadTimeout))
+		if err != nil {
+			logger.Debugw("set write deadline", "err", err)
+		}
+	}
+}
+
+func (c *Client) convertToTimeoutError(ctx context.Context, err error) error {
```
walldiss: isNetTimeoutError

share/shwap/p2p/shrex/client.go
```diff
@@ -0,0 +1,175 @@
...
+		return statusSuccess, nil
+	}
+
+	err = c.convertToTimeoutError(ctx, err)
+	if errors.Is(err, context.DeadlineExceeded) || errors.Is(err, context.Canceled) {
+		return statusTimeout, err
+	}
+	if !errors.Is(err, ErrNotFound) && errors.Is(err, ErrRateLimited) {
+		return statusRateLimited, err
+	}
```
walldiss: Rate limited should be impossible at this stage

share/shwap/p2p/shrex/shrex_getter/shrex.go
renaynay: I know we don't actually "blacklist" but i feel like this metric / status is useless bc we don't do anything w it.

share/shwap/p2p/shrex/client.go
```diff
@@ -0,0 +1,175 @@
...
+		"source", "client",
+		"name", req.Name(),
+		"peer", peer.String(),
+	)
+	requestTime := time.Now()
+	status, err := c.doRequest(ctx, logger, req, resp, peer)
+	if err != nil {
+		logger.Warnw("requesting data from peer failed", "error", err)
+	}
+	c.metrics.observeRequests(ctx, 1, req.Name(), status, time.Since(requestTime))
```
renaynay, line 66: should we also debug log this ?

share/shwap/p2p/shrex/client.go
renaynay: I'm pretty sure this entire PR is breaking - in metrics and in config. 

What does our internal error handling warrant a special PR for?

share/shwap/p2p/shrex/client.go
```diff
@@ -0,0 +1,175 @@
...
+	_, err = resp.ReadFrom(stream)
+	if err == nil {
+		return statusSuccess, nil
+	}
+
+	err = c.convertToTimeoutError(ctx, err)
+	if errors.Is(err, context.DeadlineExceeded) || errors.Is(err, context.Canceled) {
+		return statusTimeout, err
+	}
+	if !errors.Is(err, ErrNotFound) && errors.Is(err, ErrRateLimited) {
```
renaynay: is it actually possible to get rate limited all the way down here? we just need to handle for ctx deadline exceeded

share/shwap/p2p/shrex/server.go
```diff
@@ -0,0 +1,215 @@
...
+	srv.metrics = metrics
+	return nil
+}
+
+func (srv *Server) streamHandler(ctx context.Context, id newRequest) network.StreamHandler {
+	return func(s network.Stream) {
+		requestID := id()
+		handleTime := time.Now()
+		status := srv.handleDataRequest(ctx, requestID, s)
+		srv.metrics.observeRequests(ctx, 1, requestID.Name(), status, time.Since(handleTime))
```
renaynay, line 110: same here, maybe we should debug log

renaynay (commented):
we need to mark this PR as breaking

share/shwap/p2p/shrex/client.go
vgonkivs: Yes, it is breaking for metrics and config. But it's not a big deal. It does not break anything on p2p level(which is more important rn)

share/shwap/p2p/shrex/client.go
vgonkivs: No, this statement will be deleted. RateLimited comes from the error on read resp.

share/shwap/eds_id.go
vgonkivs: Request interface requires a method that returns height.

share/shwap/p2p/shrex/request.go
vgonkivs: I am ok with calling it `ResponseReader`

nodebuilder/share/module.go
vgonkivs: It looks like a refactoring. I'd not do it in this PR, but scope out all similar places and refactor everything within a single PR. Do you want me to create a tracking issue ?

share/shwap/p2p/shrex/server.go
vgonkivs: Agree. Let's use all default protocols instead.

share/shwap/p2p/shrex/shrex_getter/shrex.go
vgonkivs: This is how it worked previously in shrex-nd/eds. Your suggestion makes sense, but let me do it in a subsequent PR.

share/shwap/p2p/shrex/request.go
vgonkivs: we can create it internally.

share/shwap/p2p/shrex/params.go
vgonkivs: added defaultClient/Server params functions with vice versa numbers.

share/shwap/p2p/shrex/client.go
vgonkivs: > What does our internal error handling warrant a special PR for?

it not about the error handling, but about the error codes that are sent by the libp2p.  The nodes with the newer version will parse them while the older ones won't.

nodebuilder/share/opts.go
```diff
@@ -32,22 +31,12 @@ func WithDiscoveryMetrics(discs []*discovery.Discovery) error {
...
 }
 
-func WithShrexClientMetrics(edsClient *shrexeds.Client, ndClient *shrexnd.Client) error {
-	err := edsClient.WithMetrics()
-	if err != nil {
-		return err
-	}
-
-	return ndClient.WithMetrics()
+func WithShrexClientMetrics(client *shrex.Client) error {
```
walldiss, line 34: useless wrapper func

share/eds/accessor.go
```diff
@@ -11,6 +11,8 @@ import (
 	"github.com/celestiaorg/celestia-node/share/shwap"
 )
 
+var _ shwap.Accessor = (AccessorStreamer)(nil)
```
walldiss: does not make sense to test external implementation to satisfy local interface. Suggest to remove

share/shwap/p2p/shrex/client.go
```diff
@@ -0,0 +1,184 @@
...
+	ctx context.Context,
+	req request,
+	resp response,
+	peer peer.ID,
+) error {
+	logger := log.With(
+		"source", "client",
+		"name", req.Name(),
+		"peer", peer.String(),
+	)
```
walldiss: Can we add fileds from request here? Otherwise it is hard to understand which request has failed

share/shwap/p2p/shrex/client.go
```diff
@@ -0,0 +1,184 @@
...
+	peer peer.ID,
+) (status, error) {
+	streamOpenCtx, cancel := context.WithTimeout(ctx, c.params.ReadTimeout)
+	defer cancel()
+
+	stream, err := c.host.NewStream(streamOpenCtx, peer, ProtocolID(c.params.NetworkID(), req.Name()))
+	if err != nil {
+		if timeoutErr := asTimeoutError(ctx, err); timeoutErr != nil {
+			return statusTimeout, timeoutErr
+		}
```
walldiss: It is impossible to understand which timeout happened. Other timeout cases are possible in code below. Sometimes they are converted, sometimes not (after WriteTo). 

Not sure if it is worth separating it from other errors. Would you think removing conversion will make code simplier?

share/shwap/p2p/shrex/metrics.go
```diff
@@ -11,63 +11,92 @@ import (
...
+	statusBadRequest  status = "bad_request"
+	statusSendRespErr status = "send_resp_err"
+	statusSendReqErr  status = "send_req_err"
+	statusReadRespErr status = "read_resp_err"
+	statusReadReqErr  status = "read_req_err"
+	statusInternalErr status = "internal_err"
+	statusNotFound    status = "not_found"
+	statusTimeout     status = "timeout"
+	statusSuccess     status = "success"
+	statusRateLimited status = "rate_limited"
```
walldiss: Will be easier if we split into 2 group: client and server statuses

share/shwap/p2p/shrex/request.go
```diff
@@ -0,0 +1,48 @@
...
+		return &shwap.NamespaceDataID{}
+	}
+	edsInitID := func() request {
+		return &shwap.EdsID{}
+	}
+
+	registry = make(map[string]newRequestID)
+	registry[nsDataInitID().Name()] = nsDataInitID
+	registry[edsInitID().Name()] = edsInitID
+}
```
walldiss: worth to add comment here, since it might be unclear for first time reader what is going on here

share/shwap/p2p/shrex/middleware.go
```diff
@@ -1,41 +1,50 @@
...
+
+	rateLimiterCounter metric.Int64Counter
 }
 
-func NewMiddleware(concurrencyLimit int) *Middleware {
-	return &Middleware{
-		concurrencyLimit: int64(concurrencyLimit),
+func NewMiddleware(concurrencyLimit int) (*Middleware, error) {
+	rateLimiter, err := meter.Int64Counter("shrex_rate_limit_counter",
+		metric.WithDescription("concurrency limit of the shrex server"),
```
walldiss: This needs to be set with WithMetrics, otherwise it can be initied before metrics collector and will send all data to dev/null

share/shwap/p2p/shrex/middleware.go
```diff
@@ -1,41 +1,50 @@
...
-	// numRateLimited is the number of requests that were rate limited.
-	numRateLimited atomic.Int64
+
+	rateLimiterCounter metric.Int64Counter
 }
 
-func NewMiddleware(concurrencyLimit int) *Middleware {
-	return &Middleware{
-		concurrencyLimit: int64(concurrencyLimit),
+func NewMiddleware(concurrencyLimit int) (*Middleware, error) {
```
walldiss: Nit: does not need to be exported anymore

share/shwap/p2p/shrex/server.go
```diff
@@ -0,0 +1,216 @@
...
+	index := 0
+	for protocolName, id := range registry {
+		handler := srv.streamHandler(srv.ctx, id)
+		withRateLimit := srv.middleware.rateLimitHandler(ctx, handler, id().Name())
+		withRecovery := RecoveryMiddleware(withRateLimit)
+
+		protocols[index] = ProtocolID(srv.params.NetworkID(), protocolName)
+		srv.SetHandler(protocols[index], withRecovery)
+		index++
+	}
```
walldiss: Set of protocols is not configurable in server constructor right now. Are there plans to make it configurable? If not, registry is not needed and can loop over hardcoded set

Also what is the purpose if  `protocols` array and index var? It sets value to only pass it to SetHandler right away. Looks like some refactoring artifact that is no longer needed.

share/shwap/p2p/shrex/server.go
```diff
@@ -0,0 +1,216 @@
...
+	srv.cancel()
+	for _, protocol := range srv.protocols {
+		srv.host.RemoveStreamHandler(protocol)
+	}
+	return nil
+}
+
+func (srv *Server) Protocols() []protocol.ID {
+	return srv.protocols
+}
```
walldiss: Should it be removed? It is only used in test, where it seems not needed

share/shwap/p2p/shrex/server.go
```diff
@@ -0,0 +1,216 @@
...
+		logger.Debugw("setting read deadline", "err", err)
+	}
+
+	_, err = requestID.ReadFrom(stream)
+	if err != nil {
+		logger.Warnf("reading request: %w", err)
+		return statusReadReqErr
+	}
+
+	err = stream.CloseRead()
```
walldiss: Add request info to logger after reading it

share/shwap/p2p/shrex/server.go
```diff
@@ -0,0 +1,216 @@
...
+	file, err := srv.store.GetByHeight(ctx, requestID.Height())
+	if err == nil {
+		defer utils.CloseAndLog(log, "file", file)
+		r, err = requestID.ResponseReader(ctx, file)
+	}
+
+	deadlineErr := stream.SetWriteDeadline(time.Now().Add(srv.params.WriteTimeout))
+	if deadlineErr != nil {
+		logger.Debugw("setting write deadline", "err", err)
+	}
```
walldiss: Maybe set it inside respondStatus func?

share/shwap/p2p/shrex/server.go
```diff
@@ -0,0 +1,216 @@
...
+	logger.Debugw("new request")
+	ctx, cancel := context.WithTimeout(ctx, srv.params.HandleRequestTimeout)
+	defer cancel()
+
+	var r io.Reader
+	file, err := srv.store.GetByHeight(ctx, requestID.Height())
+	if err == nil {
+		defer utils.CloseAndLog(log, "file", file)
+		r, err = requestID.ResponseReader(ctx, file)
+	}
```
walldiss: Should this be after switch case (line 177)? Will be much easier to read if return fast statements moved above.

share/shwap/p2p/shrex/client.go
vgonkivs: Only the height field is left. Will add it

share/shwap/p2p/shrex/metrics.go
vgonkivs: Not sure if it's worth it. The statuses are self-described.

share/shwap/p2p/shrex/client.go
vgonkivs: Do you mean to entirely remove the casting to net error?

share/shwap/p2p/shrex/server.go
vgonkivs: yes

nodebuilder/share/opts.go
vgonkivs: Why useless? we have a similar approach for other components.

share/shwap/p2p/shrex/server.go
vgonkivs: But it does not refer to respondStatus only. So, imo, it's cleaner to keep it inside handler.

share/shwap/p2p/shrex/server.go
vgonkivs: No, because the switch covers errors from `srv.store.GetByHeight` and ` requestID.ResponseReader`.

nodebuilder/share/opts.go
walldiss: ah, it is fx quirk. Ok

share/shwap/p2p/shrex/client.go
walldiss: not only height. Request can have height, namespace, commitment etc. Depends on request

share/shwap/p2p/shrex/client.go
walldiss: I mean casting and handling statusTimeout entirely. Does not make much sense to separate timeout related errors imo

share/shwap/p2p/shrex/metrics.go
walldiss: They are giant piece of statuses, it is hard to read. Server and client statuses are out of order

share/shwap/p2p/shrex/server.go
vgonkivs: will rework it

walldiss: Consider handle them separately? combining this handling makes code very unusual and hard to read

share/shwap/p2p/shrex/server.go
walldiss: It does set timeout to write response. So makes sense to set it inside write response function

share/shwap/p2p/shrex/server.go
vgonkivs: I've reworked it

share/shwap/p2p/shrex/server.go
vgonkivs: write response - not write status

share/shwap/p2p/shrex/client.go
vgonkivs: request is an interface w/o access to these fields. These fields can be added in getter

share/shwap/p2p/shrex/client.go
vgonkivs: statusTimeout helps to detect high latency cases

share/shwap/p2p/shrex/metrics.go
vgonkivs: let me sort them

share/shwap/p2p/shrex/request.go
vgonkivs: removed entirely

share/shwap/p2p/shrex/middleware.go
vgonkivs: done

share/shwap/p2p/shrex/middleware.go
vgonkivs: added to metrics

share/shwap/p2p/shrex/shrex_getter/shrex.go
vgonkivs: > can we please not blacklist any peers in this iteration?

there is a metric that is updated inside setStatus. we can explicitly call observeMetrics as a temp workaround. wdyt?

share/shwap/p2p/shrex/params.go
vgonkivs: reworked client.Read param. It includes both server Write+Handle timeout.

nodebuilder/config_test.go
```diff
@@ -97,15 +97,12 @@ var outdatedConfig = `
...
-    BufferSize = 32768
-  [Share.ShrExNDParams]
-    ServerReadTimeout = "5s"
-    ServerWriteTimeout = "2m35s"
+ [Share.ShrexClient]
+    ReadTimeout = "2m0s"
+    WriteTimeout = "5s"
+  [Share.ShrexServer]   
+    ReadTimeout = "5s"
+    WriteTimeout = "1m0s"
```
renaynay, line 100: What's the rationale for decreasing the write timeout generally?

nodebuilder/share/module.go
renaynay: Yeah please do

share/shwap/eds_id.go
renaynay: okay

nodebuilder/share/module.go
```diff
@@ -185,32 +174,20 @@ func shrexComponents(tp node.Type, cfg *Config) fx.Option {
...
-				return shrexeds.NewServer(cfg.ShrExEDSParams, host, store)
-			},
-			fx.OnStart(func(ctx context.Context, server *shrexeds.Server) error {
-				return server.Start(ctx)
-			}),
-			fx.OnStop(func(ctx context.Context, server *shrexeds.Server) error {
-				return server.Stop(ctx)
-			}),
-		)),
+		fx.Invoke(func(_ *shrex.Server) {}),
```
walldiss, line 177: Is it needed?

share/shwap/p2p/shrex/client.go
renaynay: @vgonkivs in the case here, it has nothing to do with the req/resp but rather just opening the initial stream to the server.

I agree w vlad here.

it's more important to check for timeouts when it comes to reads of requests / responses

share/shwap/p2p/shrex/client.go
renaynay: I'm fine w making this a FLUP

share/shwap/p2p/shrex/client.go
walldiss: Latency observed separately in ms in `requestDuration`. there are 2 places where timeout could have happened and it is impossible which one actually was. Suggestion: remove `statusTimeout` and make `statusOpenStream` instead for this case

renaynay: error handling here could definitely use a cleanup

share/shwap/p2p/shrex/params.go
```diff
@@ -4,53 +4,92 @@ import (
...
-		ServerWriteTimeout:   time.Minute, // based on max observed sample time for 256 blocks (~50s)
+func DefaultServerParameters() *ServerParams {
+	return &ServerParams{
+		Parameters:           defaultServerParameters(),
 		HandleRequestTimeout: time.Minute,
 		ConcurrencyLimit:     10,
 	}
 }
 
+func defaultClientParameters() *Parameters {
```
renaynay: what's the point of the private funcs

share/shwap/p2p/shrex/server.go
```diff
@@ -0,0 +1,221 @@
...
+		logger.Debugw("setting read deadline", "err", err)
+	}
+
+	_, err = requestID.ReadFrom(stream)
+	if err != nil {
+		logger.Warnf("reading request: %w", err)
+		return statusReadReqErr
+	}
+
+	logger.Debug("new request")
```
renaynay: not useful

share/shwap/p2p/shrex/server.go
```diff
@@ -0,0 +1,221 @@
...
+		logger.Warnw("closing read side of the stream", "err", err)
+	}
+
+	err = requestID.Validate()
+	if err != nil {
+		logger.Warnw("validate request", "err", err)
+		return statusBadRequest
+	}
+
+	logger.Debugw("new request")
```
renaynay: here too

share/shwap/p2p/shrex/server.go
```diff
@@ -0,0 +1,221 @@
...
+	file, err := srv.store.GetByHeight(ctx, requestID.Height())
+
+	deadlineErr := stream.SetWriteDeadline(time.Now().Add(srv.params.WriteTimeout))
+	if deadlineErr != nil {
+		logger.Debugw("setting write deadline", "err", err)
+	}
+
+	if err != nil {
+		if errors.Is(err, store.ErrNotFound) {
+			logger.Errorf("file not found")
```
renaynay: please more details here, like requested height, same for err below

share/shwap/p2p/shrex/server.go
```diff
@@ -0,0 +1,221 @@
...
+			return respondStatus(logger, shrexpb.Status_NOT_FOUND, stream)
+		}
+		logger.Errorf("getting header %w", err)
+		return respondStatus(logger, shrexpb.Status_INTERNAL, stream)
+	}
+
+	defer utils.CloseAndLog(log, "file", file)
+	r, err := requestID.ResponseReader(ctx, file)
+	if err != nil {
+		logger.Errorf("getting the data %w", err)
```
renaynay:
```suggestion
		logger.Errorf("getting data from response reader: %w", err)
```

share/shwap/p2p/shrex/server.go
```diff
@@ -0,0 +1,221 @@
...
+	if status != statusSuccess {
+		return status
+	}
+
+	_, err = io.Copy(stream, r)
+	if err != nil {
+		logger.Errorw("send data", "err", err)
+		return statusSendRespErr
+	}
+	logger.Debugw("sent the data to the client")
```
renaynay, line 197: more details to make this useful debug.

nodebuilder/config_test.go
vgonkivs: ServerTimeout stays the same. From the client perspective it should not take more than ServerRead timeout(that is 5s). Also note, the request is significantly lightweight than the response, so 5s is more than enough time to send it, otherwise we deal with the poor connection that will anyway fail on response.

share/shwap/p2p/shrex/params.go
vgonkivs: we do not need them anymore. left from previous iterations. 
Thank you

share/shwap/p2p/shrex/server.go
vgonkivs: not sure whether this log will be useful at all

nodebuilder/share/module.go
vgonkivs: Yes, otherwise fx does not instantiate and start server(the similar approach we have for pruner service)

share/shwap/p2p/shrex/server.go
vgonkivs: agree. this one is unneeded

share/shwap/p2p/shrex/server.go
vgonkivs: It is actually useful as it contains the whole info wrapped with the new logger instance:

```
	logger := log.With(
		"source", "server",
		"name", requestID.Name(),
		"height", requestID.Height(),
		"peer", stream.Conn().RemotePeer().String(),
	)
```

share/shwap/p2p/shrex/server.go
vgonkivs: logger already contains all necessary info

share/shwap/p2p/shrex/server.go
vgonkivs: The same. please refer to 

```
	logger := log.With(
		"source", "server",
		"name", requestID.Name(),
		"height", requestID.Height(),
		"peer", stream.Conn().RemotePeer().String(),
	)
```

share/shwap/p2p/shrex/server.go
walldiss: It logged with Debug level twice: here and inside `streamHandler` function

share/shwap/p2p/shrex/client.go
vgonkivs: removed statusTimeout and introduce multiple new statuses:
client side:
`statusOpenStreamErr status = "open_stream_err"`
`statusReadStatusErr status = "read_status_err"`

server side:
`statusSendStatusErr status = "send_status_err"`

share/shwap/p2p/shrex/server.go
vgonkivs: `logger` contains all info. Also, fixed issue with logging empty params from requestID(before data is stored)

share/shwap/p2p/shrex/params.go
vgonkivs: reworked

share/shwap/p2p/shrex/server.go
vgonkivs: `streamHandler` logs final results of the request:
```
log.Debugw("server: handling request",
			"name", requestID.Name(),
			"status", status,
			"duration", time.Since(handleTime),
		)
```

vgonkivs: removed this line

share/shwap/p2p/shrex/server.go
vgonkivs: removed

renaynay (dismissed):

vgonkivs:
Tested conifg-update tool: the node was able to update to a new config and has been successfully started.

walldiss (approved):

PR #4256, opened 2025-05-05
feat(api): Implement Celestia Client Library v0

walldiss:

## Implementation Details

### Key Components

### Usage Example

```go
// Initialize keyring
    KeyName:     "my_key",
}, "/path/to/keys")

// Configure client
cfg := Config{
    ReadConfig: ReadConfig{
        DAAuthToken:  "token",
    },
    SubmitConfig: SubmitConfig{
        DefaultKeyName: "my_key",
        Network:        "private",
            Addr: "consensus-node:9090",
        },
    },
}

// Submit a blob

// Retrieve a blob
```

# TODOs for Further Development / Known Limitations

## Decision Points

- **Configuration Options**:

### DIFF
--- a/api/client/readme.md
+++ b/api/client/readme.md
@@ -13,8 +13,8 @@ This client package provides a simplified interface for interacting with Celesti
 - **State Management**: Check balances, execute transactions
 - **Keyring Integration**: Manage keys for transaction signing
 - **Flexible Client Modes**:
-    - Read-only mode (Bridge node connection only)
-    - Full client mode (Bridge + Core node connections)
+  - Read-only mode (Bridge node connection only)
+  - Full client mode (Bridge + Core node connections)
 
 ## Installation
 


jcstein:
is there a CLI?

cmwaters:
> Can be solved by moving client to a separate package and decoupling all RPCs

What does decoupling all RPCs entail?

jcstein:
can you please provide an example of how to use this on Mocha?

walldiss:
> is there a CLI?

This is go pkg that can be imported and used as client to celestia

> What does decoupling all RPCs entail?

Potentially move celestia-client to separate repo and redeclare RPC there. Will clarify in description. 

> so user needs a bridge node and the auth token for the bridge node to use this? what is DAAuthToken?

User needs Bridge node addr and access token to this bridge node (if any)

> can you please provide an example of how to use this on Mocha?

Example is in description. Just put mocha rpcs into it

api/client/client.go
```diff
@@ -0,0 +1,178 @@
...
+
+// TODO:
+// Major limitations:
+// We still carry deps for cosmos-sdk and celestia-app replaces.
+// - can be solved by moving client to a separate package and decoupling all rpcs.
+// Client needs to be reinitialized on every key change due to txClient being stateful and
+// maintaining copy of keys.	- can be reworked by wrapping keyring and updating txClient internal
+// state on key change. Complicated Core accessor requires KeyName to be set and exist in the
+// keyring. Which is not necessary imo. - require refactor of core_accessor/txClient
+// Can't make submit only client, because submition is coupled with reads in blob service
```
cmwaters, line 32: I don't think that's necessary and if someone really wants to submit without reading they can just use the TxClient

walldiss:
> so this is not a node's auth token like you would get from a running node, but a auth token from the RPC provider?
This is auth token you get from node. For the RPC provider it depends whether it uses same auth schema as node. If provider schema is different, we can add that as option too.

api/client/client.go
walldiss: Generally agree. There some implementation details that are not yet available in TxClient like fee estiamtion.

api/client/keyring.go
```diff
@@ -0,0 +1,73 @@
...
+	return KeyringConfig{
+		KeyName:     "my_celes_key",
+		BackendName: keyring.BackendTest,
+	}
+}
+
+// NewDefaultRing is a helper function for easy keyring initialization. It initializes a keyring
+// with the given config and path. The returned keyring is ready to use and can be used to sign
+// transactions.
+func NewDefaultRing(cfg KeyringConfig, path string) (keyring.Keyring, error) {
```
adlerjohn, line 34: As a follow-up, how about removing keyring management from `celestia-node` entirely? Especially if this client can be accessed from the command line.

api/client/grpc.go
```diff
@@ -0,0 +1,107 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/metadata"
+
+	"github.com/celestiaorg/celestia-node/libs/utils"
+)
+
+// GRPCConfig combines all configuration fields for managing the relationship with a Core node.
+type GRPCConfig struct {
```
adlerjohn, line 19: I would avoid calling this just "gRPC," since that's a transport protocol and not the target. Or rather, instead of using GRPC and Core interchangeably, Core should use GRPC under the hood, and never expose this detail.

We may have other stuff using gRPC in the future. Same with RPC.

api/client/wrappers.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	// TODO: this is a hack to allow nil options, because it's not exported by the blob package
+	if options == nil {
+		options = &blob.SubmitOptions{}
+	}
+	return api.submitter.Submit(ctx, blobs, options)
+}
+
+// readOnlyStateAPI returns an error on all state client methods if consensus connection is not
+// configured
```
adlerjohn, line 50: wait this doesn't look like what the code below does

jcstein:
> Example is in description. Just put mocha rpcs into it

can you please provide a full, working example for mocha though? as I understand it I don't know how to set this up.

for some endpoint like quicknode consensus endpoint (which is both bridge + consensus in 1 endpoint), how do you provide the auth token to this? i don't see that it is possible in current implementation. i.e. how do i use this endpoint? `https://clean-wiser-glitter.celestia-mocha.quiknode.pro/myapikey/` in both ReadConfig and SubmitConfig? I suggest using some variable other than "DAAuthToken" here and use something like "WriteRPCAuth" and "ReadRPCAuth"

```
SubmitConfig: SubmitConfig{
        DefaultKeyName: "my_key",
        Network:        "private",
            Addr: "consensus-node:9090",
        },
    },
```

based on "User needs Bridge node addr and access token to this bridge node (if any)" and "This is auth token you get from node. For the RPC provider it depends whether it uses same auth schema as node. If provider schema is different, we can add that as option too." this won't work with an RPC provider, so this whole setup requires someone to run a bridge node or have one hosted somewhere and have the auth token to it?

---

why isn't this a separate library/sdk repo? it doesn't seem like it accomplishes the goal of "minimal-overhead blob submission and retrieval" if celestia-node is still required at the end of the day, in addition to having a bridge node somewhere with auth token access

> Keyring Management: Decide whether to keep keyring inside the client

if this was taken out, how would keys be managed?

> Determine if we need logging configuration inside the client

there's currently no logging? how does user know read or write is successful?

jcstein:
this is how you use celestia-node with x-token from quicknode currently, for ref https://docs.celestia.org/how-to-guides/light-node#optional-start-light-node-with-consensus-node-endpoint-with-authentication

api/client/keyring.go
jcstein: isn't this why cel-key is a separate utility?

walldiss: If we remove key management from `celestia-node` it no longer will be able to sign tx (and submit blobs). Do you suggest remove submission from node or maybe i misunderstood smth?

walldiss:
> can you please provide a full, working example for mocha though? as I understand it I don't know how to set this up / don't think it is possible to setup with a quicknode endpoint in the current implementation.

All you need is to set 2 rpc endpoints in config similar to example. I'll provide you example shortly

> based on "User needs Bridge node addr and access token to this bridge node (if any)" and "This is auth token you get from node. For the RPC provider it depends whether it uses same auth schema as node. If provider schema is different, we can add that as option too." this won't work with an RPC provider, so this whole setup requires someone to run a bridge node or have one hosted somewhere and have the auth token to it?

Solution should work with RPC provider that supports default node auth schema. If schema is different please provide it here so I can add auth logic to client.

> why isn't this a separate library/sdk repo? it doesn't seem like it accomplishes the goal of "minimal-overhead blob submission and retrieval" if celestia-node is still required at the end of the day, in addition to having a bridge node somewhere with auth token access

Required in dependancy graph != required to run. You don't need to run node to use this client. The goal was to have 0 overhead of node and client has 0 overhead. As listed in limitations dependancies are still inherited from node. Removal of node dependancy is desirable property and listed in limitation and can be worked on later. 


> if this was taken out, how would keys be managed?

Keys can be managed by using keyring. Which initialised first in the example.

> there's currently no logging? how does user know read or write is successful?

Client returns error from call, user can handle error or log it on caller side

walldiss:
@jcstein Perhaps confusion with auth tokens comes from the fact that in example auth token is not set for grpc connection. However it is possible to setup:
https://github.com/walldiss/celestia-node/blob/15857edd96380d856793de5d845a8a02fa68958b/api/client/grpc.go#L26

I'm not familier with quicknode auth. If they use same endpoint and same token for both Bridge and Consensus, then just set Each Addr and Auth token to same value.

jcstein:
I think the endpoints have tls which is why you need to do this to set it up https://docs.celestia.org/how-to-guides/light-node#optional-start-light-node-with-consensus-node-endpoint-with-authentication so based on this https://github.com/walldiss/celestia-node/blob/15857edd96380d856793de5d845a8a02fa68958b/api/client/grpc.go#L25 it wouldn't work?

the endpoint format is like https://clean-wiser-glitter.celestia-mocha.quiknode.pro/api-key/

jcstein:
> All you need is to set 2 rpc endpoints in config similar to example. I'll provide you example shortly

i'm pointing out that if you use a quicknode endpoint as a test, it probably won't work because there's only a way to provide the node's auth token, not an auth token to an endpoint

jcstein:
after slack chat i think this is roughly what setup would look like for using qn endpoint with x-token + tls:

```
// Initialize keyring
    KeyName:     "my_key",
}, "/PATH/TO/KEYS")

// Configure client
cfg := Config{
    ReadConfig: ReadConfig{
        BridgeDAAddr: "https://YOUR-QN-PROJECT-ID.celestia-mocha.quiknode.pro:26658",
        DAAuthToken:  "YOUR-AUTH-TOKEN-GOES-HERE", # this is the x-token, a.k.a. API Key, from quicknode
    },
    SubmitConfig: SubmitConfig{
        DefaultKeyName: "my_key",
        Network:        "private",
            Addr: "https://YOUR-QN-PROJECT-ID.celestia-mocha.quiknode.pro:9090",
            AuthToken: "YOUR-AUTH-TOKEN-GOES-HERE" # this is the x-token, a.k.a. API Key, from quicknode
        },
    },
}

// Submit a blob
namespace := libshare.MustNewV0Namespace([]byte("YOUR-NAMESPACE"))
blob, _ := blob.NewBlob(libshare.ShareVersionZero, namespace, []byte("YOUR-DATA"), nil)

// Retrieve a blob
```

adlerjohn:
As a follow-up, add support for https://github.com/celestiaorg/celestia-app/pull/4628

walldiss:
All TODOs and development plans will be addressed in future PRs.

PR #4259, opened 2025-05-07
chore: magic env variable doc update

damiannolan:
One would normally assume that `genesisHash` means the hash of a genesis.json file, perhaps...
Hope this saves somebody sometime.


nodebuilder/p2p/flags.go
```diff
@@ -92,7 +92,7 @@ func ParseNetwork(cmd *cobra.Command) (Network, error) {
 func parseNetworkFromEnv() (Network, error) {
 	var network Network
 	// check if custom network option set
-	// format: CELESTIA_CUSTOM=<netID>:<genesisHash>:<bootstrapPeerList>
+	// format: CELESTIA_CUSTOM=<netID>:<genesisBlockHash>:<bootstrapPeerList>
```
damiannolan, line 95: Can the node not query this from the consensus network via:

consensus-node:26657/status - "earliest_block_hash"

Wondertan:
Related to https://github.com/celestiaorg/celestia-node/discussions/3524

nodebuilder/p2p/flags.go
Wondertan: It can, if we wanna move on making node always dependent on core which wasn't the case.

Wondertan (approved):

damiannolan:
Can we merge this?

vgonkivs (approved):

PR #4263, opened 2025-05-11
fix: refresh account cache when funded after node start

### LINKED ISSUE #4261: bug: node starts with 0 balance, then funded, but can't submit blobs
### Celestia Node version

v0.22.2

### OS

macOS & ubuntu-WSL2

### Install tools

docs

### Others

_No response_

### Steps to reproduce it

1. create new key with: `cel-key add empty --node.type light --p2p.network mocha`
2. start node `celestia light start --p2p.network mocha --core.ip rpc-mocha.pops.one --core.port 9090 --keyring.keyname empty`
3. check balance
```
celestia state balance
{
  "result": {
    "denom": "utia",
    "amount": "0"
  }
}
```
4. use faucet to fund wallet
5. check balance again
```
celestia state balance
{
  "result": {
    "denom": "utia",
    "amount": "1000000"
  }
}
```
6. while node is still running, try and submit blob
`celestia blob submit 676d 676d`
result shows signer not found and unable to submit blob
```
{
  "result": "account for signer celestia1jv2s0flrfw9wt9ue0qtnkgwufxspfznaejnqlq not found"
}
```

I know that this means account isn't funded, but the error itself isn't clear and the [go code](https://github.com/celestiaorg/celestia-node/blob/b056780200b39160c87aaee8d0e7ed01a4884c0c/state/core_access.go#L201) means that this couldn't find the signer, not that the account isn't funded. 

### Expected result

1. signer is found even if 0 balance,
2. if 0 balance, error should return clearly that there isn't a balance, not that the signer wasn't found
3. if balance, blob should submit, even if node started with 0 balance

### Actual result

error returns saying signer isn't found, when node account is both funded and signer is there

### Relevant log output

```Shell
2025-05-08T12:53:14.090-0400	WARN	rpc	go-jsonrpc@v0.6.0/handler.go:474	error in RPC call to 'blob.Submit': account for signer celestia1jv2s0flrfw9wt9ue0qtnkgwufxspfznaejnqlq not found
```

_No response_

### Notes

restarting the node and then submitting works, so this seems to be some issue with signer accessibility and fetching balance from core endpoint

avorylli:
Fixes issue #4261 where a node can't submit blobs after being funded post-startup.

Problem:
- Account cache not updated when funding received after node start
- Blob submissions fail with "account not found" error

Solution:
- Check for balance when account not found in cache
- Force TxClient refresh when account has balance but isn't cached
- Improve error messages

Fixes #4261

### DIFF
--- a/state/core_access.go
+++ b/state/core_access.go
@@ -200,7 +200,29 @@ func (ca *CoreAccessor) SubmitPayForBlob(
 	}
 	account := ca.client.AccountByAddress(author)
 	if account == nil {
-		return nil, fmt.Errorf("account for signer %s not found", author)
+		// Check if account has balance but is not cached in the tx client
+		balance, err := ca.BalanceForAddress(ctx, Address{author})
+		if err != nil {
+			return nil, fmt.Errorf("failed to check balance for account %s: %w", author, err)
+		}
+
+		if !balance.Amount.IsZero() {
+			// Account has balance but isn't cached in tx client, force refresh by recreating client
+			log.Debugw("account has balance but not found in tx client cache, refreshing client", "address", author.String())
+			ca.client = nil
+			client, err = ca.getTxClient(ctx)
+			if err != nil {
+				return nil, err
+			}
+
+			// Try to get account again
+			account = ca.client.AccountByAddress(author)
+			if account == nil {
+				return nil, fmt.Errorf("account for signer %s still not found after refresh", author)
+			}
+		} else {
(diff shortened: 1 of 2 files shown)


renaynay:
Hey @avorylli 

Thank you for this PR but I would refrain from putting this kind of logic in celestia-node. We are currently in the process of trying to shift most state-related logic to celestia-app's `TxClient` - you're welcome to open a PR there and I'd be happy to review it.

renaynay:
@avorylli 

https://github.com/celestiaorg/celestia-app/issues/4803

PR #4265, opened 2025-05-12
feat(api): Implement Celestia Client Library v0

walldiss:
Reopening PR from feature branch instead of fork. Previous discussion here https://github.com/celestiaorg/celestia-node/pull/4256

--------

## Implementation Details

### Key Components

### Usage Example

QuickNode example by @jcstein :https://github.com/celestiaorg/celestia-node/pull/4256#issuecomment-2852085985

```go
// Initialize keyring
    KeyName:     "my_key",
}, "/path/to/keys")

// Configure client
cfg := Config{
    ReadConfig: ReadConfig{
        DAAuthToken:  "token",
    },
    SubmitConfig: SubmitConfig{
        DefaultKeyName: "my_key",
        Network:        "private",
            Addr: "consensus-node:9090",
        },
    },
}

// Submit a blob

// Retrieve a blob
```

# TODOs for Further Development / Known Limitations (will be resolved in future PRs)

- **Multiple core endpoints**: will be added after https://github.com/celestiaorg/celestia-app/pull/4628 support added to the node core-accessor. 

## Decision Points

- **Configuration Options**:


api/client/grpc.go
```diff
@@ -0,0 +1,108 @@
...
+
+	"github.com/celestiaorg/celestia-node/libs/utils"
+)
+
+// CoreGRPCConfig is the configuration for the core gRPC client.
+type CoreGRPCConfig struct {
+	// Addr is the address of the core gRPC server.
+	Addr string
+	// TLSEnabled specifies whether the connection is secure or not.
+	TLSEnabled bool
```
vgonkivs, line 23: What about TLSPath? Not in this PR but worth adding a todo.

api/client/client.go
```diff
@@ -0,0 +1,156 @@
...
+
+var log = logging.Logger("celestia-client")
+
+// Config holds configuration for the Client.
+type Config struct {
+	// TODO: Do we need backwards compatibility tracking? APIVersion will check version of the API and
+	// fail if it doesn't match
+	// APIVersion        string
+
+	// TODO: do we need logging inside the client?
```
gupadhyaya: logging does not make sense as client is just a wrapper with other modules. not sure if you can log anything meaningful.

api/client/client.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/celestiaorg/celestia-node/state"
+)
+
+var log = logging.Logger("celestia-client")
+
+// Config holds configuration for the Client.
+type Config struct {
+	// TODO: Do we need backwards compatibility tracking? APIVersion will check version of the API and
+	// fail if it doesn't match
+	// APIVersion        string
```
gupadhyaya: i would also not do any API versioning here as the underlying api's themselves not have versioning and just versioning this client does not make sense.

api/client/grpc.go
walldiss: What do you mean by TLSPath for grpc client? Perhaps you refer to `XTokenPath`? In this case library expose it as config value instead of path to file.

vgonkivs: was confused with p2p layer. We haven't added support for tls certificates with the core/app endpoint

gupadhyaya (approved):

PR #4266, opened 2025-05-13
feat(api): allow configurable CORS policy

### LINKED ISSUE #3481: feat(api): allow configurable CORS policy
### Implementation ideas

We need to allow users to configure CORS policy. It should accomodate both secure and easy to configure way to set up CORS for different usecases.
- if skip-auth is enabled CORS should allow any origin
- default CORS setting should be strict mode(no cors headers)
- have a config field to add more to Access-Control-Allow-Origin manually

TheRealSibasishBehera:
Fixes #3481
kind:feat

Enforces CORS configuration rules and adds middleware handling.

- Introduces CORS flags (`--rpc.cors-enabled`, `--rpc.cors-allowed-origins`, `--rpc.cors-allowed-methods`, `--rpc.cors-allowed-headers`)

- Requires `--rpc.cors-enabled=true` to use any CORS-related flags

- Using `--rpc.skip-auth` allows all CORS settings (equivalent to * wildcard)

- Provides default values when partial configuration is given

- Fully overwrites (does not extend) settings when all configurations are provided, giving users complete control


TheRealSibasishBehera:
@walldiss Could you please review this PR? Let me know if anything needs further changes. Many thanks 🙏

nodebuilder/rpc/config.go
```diff
@@ -20,6 +29,16 @@ func DefaultConfig() Config {
...
+		CORS:     DefaultCORSConfig(),
+	}
+}
+
+func DefaultCORSConfig() CORSConfig {
+	return CORSConfig{
+		Enabled:        false,
+		AllowedOrigins: []string{},
+		AllowedHeaders: []string{},
+		AllowedMethods: []string{},
```
walldiss, line 41: Seems to miss defaults from the task

nodebuilder/rpc/config.go
```diff
@@ -44,5 +63,21 @@ func (cfg *Config) Validate() error {
...
+	return nil
+}
+
+// Add validation to reject wildcards
+func validateSecureCORS(config CORSConfig) error {
+	if utils.ContainsWildcard(config.AllowedOrigins) ||
+		utils.ContainsWildcard(config.AllowedMethods) ||
+		utils.ContainsWildcard(config.AllowedHeaders) {
+		return errors.New("wildcards not allowed in secure mode")
+	}
```
walldiss, line 81: Curious what is your intuition behind disallowing users to set wildcards?

walldiss (commented):
Sorry for delayed review. Thank you for this contribution! Really nice one!

To ensure that middleware works and sets http headers as expected we would need unit test. Would you have time to implement it?

api/rpc/server.go
```diff
@@ -15,42 +15,69 @@ import (
...
-	started atomic.Bool
+	started    atomic.Bool
+	corsConfig CORSConfig
 
 	signer   jwt.Signer
 	verifier jwt.Verifier
 }
 
-func NewServer(address, port string, authDisabled bool, signer jwt.Signer, verifier jwt.Verifier) *Server {
+func NewServer(address, port string, authDisabled, corsEnabled bool, corsAllowedHeaders, corsAllowedOrigins, corsAllowedMethods []string, signer jwt.Signer, verifier jwt.Verifier) *Server {
```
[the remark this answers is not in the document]
TheRealSibasishBehera: will do

api/rpc/server.go
```diff
@@ -15,42 +15,69 @@ import (
...
+		Handler: srv.NewHandlerStack(rpc),
+		// the amount of time allowed to read request headers. set to the default 2 seconds
+		ReadHeaderTimeout: 2 * time.Second,
 	}
+
 	return srv
 }
 
+// NewHTTPHandlerStack returns wrapped rpc related handlers
+func (s *Server) NewHandlerStack(core http.Handler) http.Handler {
```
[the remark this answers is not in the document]
TheRealSibasishBehera: will do

nodebuilder/rpc/config.go
TheRealSibasishBehera: You mentioned this in the issue comment [here](https://github.com/celestiaorg/celestia-node/issues/3481#issuecomment-2866956626) that `--skip-auth` would implicitly mean wildcard for all CORS settings if I understood correctly. 

I feel the same way, but there’s a valid use case where a user might want to set only a specific CORS setting to "*".

nodebuilder/rpc/config.go
TheRealSibasishBehera: The default configuration for CORS is that CORS is disabled. So the other parameters are set to empty as the default values.

The case where cors is enabled without a possible subsetting, i.e., `--rpc.cors-allowed-origins` or `--rpc.cors-allowed-methods` or `--rpc.cors-allowed-headers`  , in that case, default values are filled in while parsing the flags for each of the subsetting. 

Do you think this should be moved someplace outside the `ParseFlags` in `nodebuilder/rpc/flags.go/` ?

api/rpc/server.go
```diff
@@ -61,6 +88,46 @@ func (s *Server) verifyAuth(_ context.Context, token string) ([]auth.Permission,
 	return authtoken.ExtractSignedPermissions(s.verifier, token)
 }
 
+// authHandler wraps the handler with authentication.
+func (s *Server) authHandler(next http.Handler) http.Handler {
+	if s.authDisabled {
+		return next
+	}
```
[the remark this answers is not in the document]
TheRealSibasishBehera: agreed,  I will change that

nodebuilder/rpc/config.go
walldiss: The validation will not allow user to set any wildcards explicitly right now.  But I think he should be able to

api/rpc/server.go
```diff
@@ -15,42 +15,65 @@ import (
...
+		return s.corsAny(core)
+	}
+
+	if s.corsConfig.Enabled {
+		return s.corsWithConfig(s.authHandler(core))
+	}
+
+	return s.authHandler(core)
+}
+
```
walldiss, line 82: Beautiful code right there! 🚀

Wondertan (approved):

PR #4267, opened 2025-05-14
test(e2e): add `TestE2EMsgPayForBlob`

chatton:
### Overview

This PR adds an E2E test using the new [e2e testing library](https://github.com/celestiaorg/tastora) adapted from [interchaintest](https://github.com/strangelove-ventures/interchaintest).

A single E2E which does the following is added in this PR.

- Creates a single validator chain
- Creates ca bridge node
- Submits a MsgPayForBlobs
- Creates a light node
- Verifies the blob can be queried from the light node and that the bytes are correct and in the specificed namespace.

This test should catch a lot of errors where basic compatibility gets broken.

Right now I have hard coded an app version of the latest RC, however the version of node that gets tested is built on every PR so we can tell if changes in any given PR break the basic flow.

I included this test under a new go.mod to avoid adding test library dependencies to the main go.mod.

You can run this test locally by just running `make test-e2e test=TestE2EMsgPayForBlob`.


.github/workflows/go-ci.yml
```diff
@@ -149,3 +149,37 @@ jobs:
...
+        uses: actions/setup-go@v4
+        with:
+          go-version-file: go.mod
+
+      - name: Generate image tag
+        id: tag
+        run: echo "value=$(echo $GITHUB_SHA | cut -c1-8)" >> $GITHUB_OUTPUT
+
+      - name: Build Docker image locally
+        run: docker build -t "celestia-node:${{ steps.tag.outputs.value }}" .
```
chatton, line 176: for now I am intentionally not pushing to ghcr.io/celestia-node as the existing re-usable docker workflow does that, and don't want to clash with anything there.

This adds a bit of time to each run, and can be re-considered later if it is a problem.

e2e/e2e_test.go
```diff
@@ -0,0 +1,150 @@
...
+	return overrides
+}
+
+func (s *CelestiaTestSuite) CreateDockerProvider() celestiatypes.Provider {
+	numValidators := 1
+	numFullNodes := 0
+
+	enc := testutil.MakeTestEncodingConfig(app.ModuleEncodingRegisters...)
+
+	cfg := celestiadockertypes.Config{
```
chatton, line 69: a lot of this configuration is pretty boiler plate and can probably be done in the testing library. 

All configuration for the docker chains/bridge/light nodes is done here.

e2e/e2e_test.go
```diff
@@ -0,0 +1,150 @@
...
+	// required to query tx by hash when broadcasting transactions.
+	overrides := make(toml.Toml)
+	txIndexConfig := make(toml.Toml)
+	txIndexConfig["indexer"] = "kv"
+	overrides["tx_index"] = txIndexConfig
+	return overrides
+}
+
+func (s *CelestiaTestSuite) CreateDockerProvider() celestiatypes.Provider {
+	numValidators := 1
```
chatton, line 64: since this is a correctness test, we can just run a single node, if there is a need/desire to increase this, it is always an option.

e2e/go.mod
```diff
@@ -0,0 +1,285 @@
+module github.com/celestiaorg/celestia-node/e2e
```
chatton, line 1: added go mod to not introduce extra deps in the main go.mod

e2e/Makefile
```diff
@@ -0,0 +1,8 @@
+## e2e-test: Run end to end tests via docker.
+e2e-test:
```
cristaloleg, line 2: IMO this should be a part of main `Makefile` (in root directory).

Makefile
```diff
@@ -160,6 +160,16 @@ test-integration-race:
...
 .PHONY: test-integration-race
 
+## test-e2e: Run an end to end test via docker.
+test-e2e:
+	@if [ -z "$(test)" ]; then \
+		echo "ERROR: 'test' variable is required. Usage: make test-e2e test=TestE2EMsgPayForBlob"; \
+		exit 1; \
+	fi
+	@echo "--> Running: TestCelestiaTestSuite/$(test)"
+	cd e2e && go test -v -run ^TestCelestiaTestSuite/$(test)$$ ./...
```
cristaloleg, line 170: `go test -v -run ^TestCelestiaTestSuite/$(test)$$ ./e2e` should work as well.

BTW, why do we need to specify test all the time? Why not run all the test by default unless specific test(s?) is specified? Just curious.

e2e/Makefile
chatton: moved it 🫡

chatton:
investigating the CI failure, it looks like it is an environmental issue 👀

e2e/go.mod
```diff
@@ -0,0 +1,279 @@
...
+
+go 1.23.6
```
chatton, line 3: I explicitly chose this version due to a breakage if bumped to v1.24

`link: github.com/bytedance/sonic/loader: invalid reference to runtime.lastmoduledatap`

It looks like the sonic library is pulled in by the cosmos sdk and depends on some go internals that changed between these versions.

PR #4189, opened 2025-03-29
chore: upgrade to golangci-lint v2

quantix9:
migrate ref: https://golangci-lint.run/product/migration-guide/

PR #4280, opened 2025-05-22
chore: upgrade to golangci-lint v2

Wondertan:
Have to reopen #4189  due to perms issue


core/header_test.go
```diff
@@ -57,9 +57,9 @@ func TestMismatchedDataHash_ComputedRoot(t *testing.T) {
 
 func TestBadAppVersion(t *testing.T) {
 	header := headertest.RandExtendedHeader(t)
-	header.RawHeader.Version.App = appconsts.LatestVersion + 1
+	header.Version.App = appconsts.LatestVersion + 1
```
walldiss, line 60: Why struct fields are no longer accessed by field name?

Wondertan: Linter rule I guess

walldiss (approved):

Issue #4240, opened 2025-04-28
Pruning Integration


Wondertan:
This is a meta/epic issue for pruning integration based on https://github.com/celestiaorg/go-header/issues/260 WIP release.

Wondertan:
Initially I thought #4042 is the feature we need, but it is actually targeting a different problem

Wondertan:
We found a way to integrate pruning via https://github.com/celestiaorg/go-header/issues/301 without breaking storage, making #4243 and #4335 obsolete, so closing them.

PR #4283, opened 2025-05-26
test: Tastora framework integration with test for blob module

chatton:
### Overview

This PR adds 2 tests using the new [tastora](https://github.com/celestiaorg/tastora) library. 

This PR specifies a version of celestia-app that has [these changes](https://github.com/celestiaorg/celestia-app/pull/4854) as `TxClient` is not correctly handling accounts without it. 

This can stay as a reference PR until the changes linked above make their way into an RC.

One thing to note: the test sutie type in `e2e/e2e_test.go` will liekly end up moving directly into the tastora library and be importable directly into celestia-app and celestia-node tests, as a lot of the functionality here could be useful in both repos.


nodebuilder/tests/tastora/framework.go
```diff
@@ -0,0 +1,417 @@
...
+	celestia, err := f.provider.GetChain(ctx)
+	require.NoError(f.t, err, "failed to get chain")
+
+	err = celestia.Start(ctx)
+	require.NoError(f.t, err)
+
+	// verify the chain is producing blocks
+	require.NoError(f.t, wait.ForBlocks(ctx, 2, celestia))
+	return celestia
+}
```
chatton, line 398: based on the work in the refactor I think just this function will need to change (not now), it will end up building the chain a different way, this is great that it's scoped to a small function so the impact should be minimal 👍

nodebuilder/tests/tastora/framework.go
```diff
@@ -0,0 +1,417 @@
...
+
+	// Cleanup function
+	t.Cleanup(f.cleanup)
+
+	return f
+}
+
+// SetupNetwork initializes the complete network infrastructure.
+// This includes starting the Celestia chain and all DA nodes.
+func (f *Framework) SetupNetwork(ctx context.Context) error {
```
chatton, line 86: just an observation, it looks like the `Framework` type is implementing a lot of lifecycle type functions that exist in `suite.Suite`

did you consider making `Framework` a `suite.Suite` and implementing these as first class `SetupTest` / `SetupSuite` type functions?

```go
type Framework struct {
  suite.Suite
  // ....
}
```

would give access to all the `f.T().Require()` fns as an added bonus.

nodebuilder/tests/tastora/framework.go
```diff
@@ -0,0 +1,417 @@
...
+	genesisHash := block.Block.Header.Hash().String()
+	require.NotEmpty(f.t, genesisHash, "genesis hash is empty")
+	return genesisHash
+}
+
+// cleanup frees up all resources.
+func (f *Framework) cleanup() {
+	f.logger.Info("Cleaning up Tastora framework")
+	// Cleanup is handled by the Tastora framework automatically
+}
```
chatton: not sure if this provides any value?

chatton (commented):
Can't approve as I opened the original PR, adding ACK for usage of the tastora lib, everything looks good to me 👍

nodebuilder/tests/tastora/framework.go
gupadhyaya: it is better to keep the framework and testsuites not tightly coupled.
Framework = Infrastructure management (Docker, nodes, wallets)
TestSuite = Test orchestration and assertions
note that, testsuites extend suite.Suite

Makefile
```diff
@@ -160,6 +160,16 @@ test-integration-race:
...
 
+## test-blob: Run blob module tests via Tastora framework.
+test-blob:
+	@echo "--> Running blob module tests"
+	cd nodebuilder/tests/tastora && go test -v -run TestBlobTestSuite ./...
+
+## test-tastora: Run all Tastora framework tests.
+test-tastora: test-blob
+	@echo "--> Running all Tastora tests"
+	cd nodebuilder/tests/tastora && go test -v ./...
```
Wondertan: This will run the blob test twice. Is this a desired behavior? Seems unlikely.

nodebuilder/tests/tastora/config.go
```diff
@@ -0,0 +1,60 @@
...
+	return func(c *Config) {
+		c.LightNodeCount = count
+	}
+}
+
+// WithBlockTime sets the block time for the chain (for sync testing).
+func WithBlockTime(blockTime int) Option {
+	return func(c *Config) {
+		// This would be used in chain configuration
+		// Implementation depends on how block time is configured in Tastora
```
Wondertan: Do we track support for this somewhere?

nodebuilder/tests/tastora/framework.go
Wondertan: If there is nothing to clean up, let's just remove it.

To double-check, does every test artifact get removed? By docker?

nodebuilder/tests/tastora/framework.go
```diff
@@ -0,0 +1,542 @@
...
+
+	// Cleanup function
+	t.Cleanup(f.cleanup)
+
+	return f
+}
+
+// SetupNetwork initializes the complete network infrastructure.
+// This includes starting the Celestia chain and all DA nodes.
+func (f *Framework) SetupNetwork(ctx context.Context) error {
```
Wondertan, line 86: devx: What's the reason to _not_ run nodes lazily? I mean, calling `GetBridgeNode` could create a new node in-place, rather than requiring you to specify everything in advance in the config.

It may be a taste-dependent comment, but I find querying the node slice, which is manually indexed, to retrieve the needed node, quite inconvenient. Also, as a person who wrote a bunch of tests for the node with swamp, I can say you don't always get the correct number of nodes right before you start writing test, and simple ad-hoc calling of `NewLightNode`  when you need it is a minor, but still a devx win.

nodebuilder/tests/tastora/framework.go
```diff
@@ -0,0 +1,542 @@
...
+	return f.lightNodes
+}
+
+// GetCelestiaChain returns the Celestia chain instance.
+func (f *Framework) GetCelestiaChain() tastoratypes.Chain {
+	return f.celestia
+}
+
+// GetNodeRPCClient retrieves an RPC client for the provided DA node.
+func (f *Framework) GetNodeRPCClient(ctx context.Context, daNode tastoratypes.DANode) *rpcclient.Client {
```
Wondertan, line 220: devx: This could be inlined with getting a node. I mean, we need an RPC client at 95% for every node, and you could just add the method on DANode type that returns an RPC. Also, looking at DANode iface, for some reason it defines a bunch of methods copied from the client, which could be reduced.

nodebuilder/tests/tastora/framework.go
```diff
@@ -0,0 +1,542 @@
...
+				"P2P_NETWORK":     testChainID,
+			},
+		),
+	)
+	require.NoError(f.t, err, "failed to start light node")
+	return lightNode
+}
+
+// getGenesisHash returns the genesis hash of the chain.
+func (f *Framework) getGenesisHash(ctx context.Context, chain tastoratypes.Chain) string {
```
Wondertan, line 482: Note to myself that we can remove this after #4240 is merged.

nodebuilder/tests/tastora/blob_test.go
```diff
@@ -0,0 +1,405 @@
...
+	libBlobV1, err := blobtypes.NewV1Blob(namespace, dataV1, walletAddr)
+	s.Require().NoError(err)
+
+	// Submit mixed blobs via chain
+	signerStr := testWallet.GetFormattedAddress()
+	msg, err := blobtypes.NewMsgPayForBlobs(signerStr, appconsts.LatestVersion, libBlobV0, libBlobV1)
+	s.Require().NoError(err)
+
+	chain := s.framework.GetCelestiaChain()
+	resp, err := chain.BroadcastBlobMessage(ctx, testWallet, msg, libBlobV0, libBlobV1)
```
Wondertan: Do we send it through chain because you can't send the older version with node or what?

nodebuilder/tests/tastora/framework.go
```diff
@@ -0,0 +1,542 @@
...
+// getNodeImage returns the Celestia node image.
+func getNodeImage() string {
+	if tag := os.Getenv("CELESTIA_NODE_IMAGE"); tag != "" {
+		return tag
+	}
+	return nodeImage
+}
+
+// GenerateTestBlobs generates test blobs for blob operations
+func (f *Framework) GenerateTestBlobs(count int, dataSize int) ([]*blob.Blob, error) {
```
Wondertan: Shouldn't be a method, and part of this file. It doesn't use any fields on the Framework and is quite irrelevant to its responsibility. I would move it to test utilities next to the blob pkg or at least a separate utilities file. Same as other generator methods.

nodebuilder/tests/tastora/framework.go
```diff
@@ -0,0 +1,542 @@
...
+
+		blobs[i] = testBlob
+	}
+
+	return blobs, nil
+}
+
+// FillBlocks produces the given amount of contiguous blocks with customizable size.
+// This is needed for sync testing scenarios. Returns a channel that reports when done.
+func (f *Framework) FillBlocks(ctx context.Context, account string, blockSize, numBlocks int) (<-chan error, error) {
```
Wondertan: This method seems completely useless. It only simulates block filling, but also does it quite inaccurately with a fixed 100ms duration per block. Instead, we have a handy WaitForHeight client method that renders this method useless.

cc @gupadhyaya to confirm if this is used in swamp ports to tastora.

nodebuilder/tests/tastora/framework.go
```diff
@@ -0,0 +1,542 @@
...
+		}
+		errCh <- nil
+	}()
+
+	return errCh, nil
+}
+
+// WaitTillHeight waits until the chain reaches the specified height.
+// This is useful for sync testing scenarios.
+func (f *Framework) WaitTillHeight(ctx context.Context, height uint64) error {
```
Wondertan: This also seems redundant. Inaccurate and have a better alternative on the node client. If its not used for the test setup, I think this can be safely deleted

again, cc @gupadhyaya  to confirm

nodebuilder/tests/tastora/framework.go
```diff
@@ -0,0 +1,542 @@
...
+	celestiaAppImage   = "ghcr.io/celestiaorg/celestia-app"
+	defaultCelestiaTag = "v4.0.0-rc6"
+	nodeImage          = "ghcr.io/celestiaorg/celestia-node"
+	defaultNodeTag     = "v0.23.2-mocha"
+	testChainID        = "test"
+)
+
+// DefaultTestTimeout should be used as the default timeout on all Tastora tests.
+// It's generously set to 10 minutes to give enough time for CI.
+const DefaultTestTimeout = 10 * time.Minute
```
Wondertan: Is this even used?

nodebuilder/tests/tastora/framework.go
```diff
@@ -0,0 +1,542 @@
...
+	// wait for blocks to ensure the funds are available
+	require.NoError(f.t, wait.ForBlocks(ctx, 2, f.celestia))
+
+	bal := f.queryBalance(ctx, toAddr.String())
+	require.GreaterOrEqual(f.t, bal.Amount.Int64(), amount, "balance is not sufficient")
+}
+
+// FundNodeAccount funds a DA node's account address from the provided wallet.
+// This is a consolidated method that can be used across all test modules.
+func (f *Framework) FundNodeAccount(ctx context.Context, fromWallet tastoratypes.Wallet, daNode tastoratypes.DANode, amount int64) sdk.AccAddress {
```
Wondertan: devx: Any reason not to do it automatically for every node?

Wondertan (commented):
Reviewed mostly from devx persective.

nodebuilder/tests/tastora/framework.go
chatton: yeah cleanup should happen automatically. There are some issues to explicitly bypass this if we want resources to hang around for debugging purposes, but right now it should just delete everything so yeah this function doesn't need to exist.

nodebuilder/tests/tastora/config.go
gupadhyaya: this will be addressed in the followup PR that adds sync tests.

Makefile
gupadhyaya: when the next set of tests gets added, it will looks more like:
```
test-tastora: test-blob test-sync ...
```

nodebuilder/tests/tastora/framework.go
gupadhyaya: its used in the sync and das tests. let's revisit while review them. i prefer to keep it for now if ok?

Makefile
Wondertan: ok, but then we don't need to run `go test` manually.

nodebuilder/tests/tastora/framework.go
Wondertan: I mean #4330

Makefile
gupadhyaya: keeping go test and removing test-blob.

nodebuilder/tests/tastora/framework.go
gupadhyaya: we will revisit this in the next PR: https://github.com/celestiaorg/tastora/issues/76

nodebuilder/tests/tastora/framework.go
gupadhyaya: not needed as part of this PR, so removing it and will revise it in the next PR while adding the sync tests.

nodebuilder/tests/tastora/blob_test.go
gupadhyaya: i think mainly to test the alternative approach of sending blobs. we can still use node api.

nodebuilder/tests/tastora/framework.go
gupadhyaya: made it demand-based now.

nodebuilder/tests/tastora/blob_test.go
Wondertan: Hmm, not sure we should be testing orthogonal tools here, but the node

nodebuilder/tests/tastora/framework.go
gupadhyaya: ok. removed as it is not required in this PR. will add back in the relevant test in the next PRs.

nodebuilder/tests/tastora/blob_test.go
gupadhyaya: makes sense.

gupadhyaya (approved):

PR #4288, opened 2025-05-31
enable getSamples and getRow in shrex

vgonkivs:
What was done:
* Implemented 3 new shrex operations: `GetSamples`, `GetRow`, and `GetRangeNamespaceData` (previously returned `ErrOperationNotSupported`)
* Unified metrics system: Replaced separate EDS/namespace metrics with a single `getters_shrex_attempts_per_request` metric that uses `request_type` labels
* Refactored request handling: Introduced a unified `executeRequest` method that handles retry logic, peer management, and verification for all request types.

### DIFF
--- a/share/shwap/p2p/shrex/shrex_getter/shrex.go
+++ b/share/shwap/p2p/shrex/shrex_getter/shrex.go
@@ -146,6 +146,10 @@ func (sg *Getter) GetSamples(
 		})
 	}
 
+	// Wait for all sample requests to complete. On partial failures, we return
+	// whatever samples were successfully retrieved along with the error.
+	// Callers should check both the returned samples slice and error - a non-nil
+	// error with non-empty samples indicates partial success.
 	err = errGroup.Wait()
 	// Delete empty entries if samples were not found or any other error occurred and return partial response
 	samples = slices.DeleteFunc(samples, func(sample shwap.Sample) bool {
@@ -228,6 +232,7 @@ func (sg *Getter) GetEDS(ctx context.Context, header *header.ExtendedHeader) (*r
 	)
 
 	req := func(ctx context.Context, peer libpeer.ID) error {
+		buff.Reset()
 		return sg.client.Get(ctx, &request, buff, peer)
 	}
 
@@ -319,6 +324,11 @@ func (sg *Getter) GetRangeNamespaceData(
 	defer func() {
 		utils.SetStatusAndEnd(span, err)
 	}()
+
+	// short circuit if the data root is empty
+	if header.DAH.Equals(share.EmptyEDSRoots()) {
+		return shwap.RangeNamespaceData{}, nil
+	}
 
 	edsID, err := shwap.NewEdsID(header.Height())
 	if err != nil {
@@ -445,7 +455,6 @@ func (sg *Getter) executeRequest(
 				break
 			}
 			sg.metrics.recordAttempts(ctx, reqType, attempt, true)
-			setStatus(peers.ResultNoop)
 			return nil
 		case errors.Is(getErr, context.DeadlineExceeded),
 			errors.Is(getErr, context.Canceled):
--- a/share/shwap/range_namespace_data_id.go
+++ b/share/shwap/range_namespace_data_id.go
@@ -64,21 +64,12 @@ func (rngid RangeNamespaceDataID) Name() string {
 // Verify validates the RangeNamespaceDataID fields and verifies that number of the requested shares
 // does not exceed the number of shares inside the ODS.
 func (rngid RangeNamespaceDataID) Verify(odsSize int) error {
-	err := rngid.EdsID.Validate()
+	err := rngid.Validate()
 	if err != nil {
-		return fmt.Errorf("invalid EdsID: %w", err)
+		return err
 	}
 
 	sharesAmount := odsSize * odsSize
-	if rngid.From < 0 {
-		return fmt.Errorf("from must be greater than or equal to 0: %d", rngid.From)
-	}
-	if rngid.To <= 0 {
-		return fmt.Errorf("to must be greater than 0: %d", rngid.To)
-	}
-	if rngid.From >= rngid.To {
-		return fmt.Errorf("invalid range: from %d to %d", rngid.From, rngid.To)
-	}
 	if rngid.From >= sharesAmount {
 		return fmt.Errorf("invalid start index: from %d >= size: %d", rngid.From, odsSize)
 	}
@@ -90,6 +81,19 @@ func (rngid RangeNamespaceDataID) Verify(odsSize int) error {
 
 // Validate performs basic fields validation.
 func (rngid RangeNamespaceDataID) Validate() error {
+	err := rngid.EdsID.Validate()
+	if err != nil {
+		return fmt.Errorf("invalid EdsID: %w", err)
+	}
+	if rngid.From < 0 {
+		return fmt.Errorf("%w: From: %d < 0", ErrInvalidID, rngid.From)
+	}
+	if rngid.To <= 0 {
+		return fmt.Errorf("%w: To: %d <= 0", ErrInvalidID, rngid.To)
+	}
+	if rngid.From >= rngid.To {
+		return fmt.Errorf("invalid range: from %d to %d", rngid.From, rngid.To)
+	}
 	return nil
 }
 
@@ -111,6 +115,10 @@ func (rngid *RangeNamespaceDataID) ReadFrom(r io.Reader) (int64, error) {
 
 // WriteTo writes the binary form of RangeNamespaceDataID to the provided writer.
 func (rngid RangeNamespaceDataID) WriteTo(w io.Writer) (int64, error) {
+	if err := rngid.Validate(); err != nil {
+		return int64(0), err
+	}
+
 	data, err := rngid.MarshalBinary()
 	if err != nil {
 		return 0, err
--- a/share/shwap/row.go
+++ b/share/shwap/row.go
@@ -2,7 +2,6 @@ package shwap
 
 import (
 	"bytes"
-	"context"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -265,19 +264,4 @@ func (r *Row) ReadFrom(reader io.Reader) (int64, error) {
 
 	*r, err = RowFromProto(&rowpb)
 	return int64(n), err
-}
-
-func (rid RowID) ResponseReader(ctx context.Context, acc Accessor) (io.Reader, error) {
-	halfRow, err := acc.AxisHalf(ctx, rsmt2d.Row, rid.RowIndex)
-	if err != nil {
-		return nil, fmt.Errorf("getting half row from accessor: %w", err)
-	}
-
-	row := halfRow.ToRow()
-	buf := &bytes.Buffer{}
-	_, err = row.WriteTo(buf)
-	if err != nil {
-		return nil, fmt.Errorf("writing row: %w", err)
-	}
-	return buf, nil
 }
--- a/share/shwap/row_id.go
+++ b/share/shwap/row_id.go
@@ -139,17 +139,17 @@ func (rid RowID) AppendBinary(data []byte) ([]byte, error) {
 	return binary.BigEndian.AppendUint16(data, uint16(rid.RowIndex)), nil
 }
 
-func (rid RowID) ResponseDataReader(ctx context.Context, acc Accessor) (io.Reader, error) {
-	axisHalf, err := acc.AxisHalf(ctx, rsmt2d.Row, rid.RowIndex)
+func (rid RowID) ResponseReader(ctx context.Context, acc Accessor) (io.Reader, error) {
+	halfRow, err := acc.AxisHalf(ctx, rsmt2d.Row, rid.RowIndex)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("getting half row from accessor: %w", err)
 	}
 
-	r := axisHalf.ToRow()
+	row := halfRow.ToRow()
 	buf := &bytes.Buffer{}
-	_, err = r.WriteTo(buf)
+	_, err = row.WriteTo(buf)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("writing row: %w", err)
 	}
 	return buf, nil
 }
--- a/share/shwap/row_namespace_data.go
+++ b/share/shwap/row_namespace_data.go
@@ -16,9 +16,6 @@ import (
 	"github.com/celestiaorg/celestia-node/share/shwap/pb"
 )
 
-// RowNamespaceDataName is the name identifier for the row namespace data container.
-const RowNamespaceDataName = "rnd_v0"
-
 // ErrNamespaceOutsideRange is returned by RowNamespaceDataFromShares when the target namespace is
 // outside of the namespace range for the given row. In this case, the implementation cannot return
 // the non-inclusion proof and will return ErrNamespaceOutsideRange.


share/shwap/namespace_data_id.go
```diff
@@ -41,6 +42,33 @@ func NewNamespaceDataID(height uint64, namespace libshare.Namespace) (NamespaceD
...
+
+	err := json.Unmarshal(data, &jsonNdID)
+	if err != nil {
+		return err
+	}
+	ndid.height = jsonNdID.Height
+	ndid.DataNamespace = jsonNdID.DataNamespace
+	return nil
+}
+
```
walldiss: Did we change namespace data to use single request?

share/shwap/p2p/shrex/shrex_getter/metrics.go
```diff
@@ -0,0 +1,70 @@
...
+	case rowRequest:
+		return "rowRequest"
+	case samplesRequest:
+		return "samplesRequest"
+	case rangeRequest:
+		return "rangeRequest"
+	default:
+		panic(fmt.Sprintf("unknown shrex getter requestType: %v", int(r)))
+	}
+}
```
walldiss: This seem to duplicate shwap type native Name method. Can metrics use Name() instead of defining request string tags and type?

share/shwap/p2p/shrex/shrex_getter/shrex.go
```diff
@@ -143,12 +94,101 @@ func (sg *Getter) Stop(ctx context.Context) error {
...
+			if samples[i].IsEmpty() {
+				return errors.New("nil response")
+			}
+			return samples[i].Verify(header.DAH, request.RowIndex, request.ShareIndex)
+		}
+		err = sg.executeRequest(ctx, logger, header, samplesRequest, req, verify)
+		if err != nil {
+			return nil, err
+		}
+	}
```
walldiss, line 147: Lets request in parallel. There could be 16 samples and they would be requested sequentially

share/shwap/p2p/shrex/shrex_getter/shrex_test.go
```diff
@@ -240,6 +240,133 @@ func TestShrexGetter(t *testing.T) {
...
+
+		coords := []shwap.SampleCoords{
+			{Row: 0, Col: 0},
+			{Row: 0, Col: 1},
+			{Row: 0, Col: 2},
+			{Row: 0, Col: 3},
+		}
+		got, err := getter.GetSamples(ctx, eh, coords)
+		require.NoError(t, err)
+		assert.Len(t, got, len(coords))
```
walldiss, line 276: assert response to be what you expect

share/shwap/p2p/shrex/shrex_getter/shrex_test.go
```diff
@@ -240,6 +240,133 @@ func TestShrexGetter(t *testing.T) {
...
+			DataHash: roots.Hash(),
+			Height:   height,
+		})
+
+		coords := []shwap.SampleCoords{
+			{Row: 0, Col: 10},
+		}
+
+		_, err := getter.GetSamples(ctx, eh, coords)
+		require.Error(t, err)
```
walldiss, line 307: check for specific error

share/shwap/p2p/shrex/shrex_getter/shrex_test.go
```diff
@@ -240,6 +240,133 @@ func TestShrexGetter(t *testing.T) {
...
+			Height:   height,
+		})
+
+		coords := []shwap.SampleCoords{
+			{Row: 0, Col: 5},
+			{Row: 0, Col: 10},
+		}
+
+		_, err := getter.GetSamples(ctx, eh, coords)
+		require.Error(t, err)
```
walldiss: need to asset if partial response

share/shwap/range_namespace_data_id.go
```diff
@@ -155,3 +195,18 @@ func (rngid RangeNamespaceDataID) appendTo(data []byte) ([]byte, error) {
...
 	return data, nil
 }
+
+func (rngid RangeNamespaceDataID) ResponseReader(ctx context.Context, acc Accessor) (io.Reader, error) {
+	rngdata, err := acc.RangeNamespaceData(ctx, rngid.From, rngid.To)
+	if err != nil {
+		return nil, fmt.Errorf("getting rngdata from accessor: %w", err)
+	}
+
+	// TODO(@vgonkivs): This is a temporary solution that will be reworked
```
walldiss: Reworked to what? What makes it remporary?

share/shwap/range_namespace_data_id.go
```diff
@@ -11,6 +14,8 @@ import (
 // for the start and end ODS indexes of share of the range.
 const RangeNamespaceDataIDSize = EdsIDSize + 4
 
+const rangeNamespaceData = "rangeNamespaceData_v0"
```
walldiss: rangeNamespaceDataName

share/shwap/p2p/shrex/shrex_getter/shrex.go
```diff
@@ -143,12 +94,101 @@ func (sg *Getter) Stop(ctx context.Context) error {
...
 
-func (sg *Getter) GetSamples(context.Context, *header.ExtendedHeader, []shwap.SampleCoords) ([]shwap.Sample, error) {
-	return nil, fmt.Errorf("getter/shrex: GetShare %w", shwap.ErrOperationNotSupported)
+func (sg *Getter) GetSamples(
+	ctx context.Context,
+	header *header.ExtendedHeader,
+	coords []shwap.SampleCoords,
+) ([]shwap.Sample, error) {
+	var err error
+	ctx, span := tracer.Start(ctx, "shrex/get-eds")
```
renaynay:
```suggestion
	ctx, span := tracer.Start(ctx, "shrex/get-samples")
```

share/shwap/p2p/shrex/shrex_getter/shrex.go
```diff
@@ -143,12 +94,101 @@ func (sg *Getter) Stop(ctx context.Context) error {
...
+	defer func() {
+		utils.SetStatusAndEnd(span, err)
+	}()
+	// short circuit if the data root is empty
+	if header.DAH.Equals(share.EmptyEDSRoots()) {
```
renaynay, line 105: start span after this

share/shwap/p2p/shrex/shrex_getter/shrex.go
renaynay: this could have been a bottleneck btw to the request throughput :)

share/shwap/p2p/shrex/shrex_getter/shrex.go
```diff
@@ -143,12 +94,101 @@ func (sg *Getter) Stop(ctx context.Context) error {
...
+	}
+
+	return samples, nil
 }
 
-func (sg *Getter) GetRow(_ context.Context, _ *header.ExtendedHeader, _ int) (shwap.Row, error) {
-	return shwap.Row{}, fmt.Errorf("getter/shrex: GetRow %w", shwap.ErrOperationNotSupported)
+func (sg *Getter) GetRow(ctx context.Context, header *header.ExtendedHeader, rowIndex int) (shwap.Row, error) {
+	var err error
+	ctx, span := tracer.Start(ctx, "shrex/get-eds")
```
renaynay:
```suggestion
	ctx, span := tracer.Start(ctx, "shrex/get-row")
```

share/shwap/p2p/shrex/shrex_getter/shrex.go
```diff
@@ -143,12 +94,101 @@ func (sg *Getter) Stop(ctx context.Context) error {
...
+	defer func() {
+		utils.SetStatusAndEnd(span, err)
+	}()
+	// short circuit if the data root is empty
+	if header.DAH.Equals(share.EmptyEDSRoots()) {
```
renaynay, line 163: same here

share/shwap/p2p/shrex/shrex_getter/shrex.go
```diff
@@ -265,40 +271,175 @@ func (sg *Getter) GetNamespaceData(
...
+		"hash", dah.String(),
+		"namespace", namespace.String(),
+	)
+
+	req := func(ctx context.Context, peer libpeer.ID) (int64, error) {
+		return sg.client.Get(ctx, &request, &response, peer)
+	}
+
+	verify := func() error {
+		if response == nil {
```
renaynay: response.IsEmpty?

share/shwap/p2p/shrex/shrex_getter/shrex_test.go
```diff
@@ -240,6 +240,133 @@ func TestShrexGetter(t *testing.T) {
...
+
+		err = edsStore.PutODSQ4(ctx, roots, height, randEDS)
+		require.NoError(t, err)
+		fullPeerManager.Validate(ctx, srvHost.ID(), shrexsub.Notification{
+			DataHash: roots.Hash(),
+			Height:   height,
+		})
+
+		_, err := getter.GetRow(ctx, eh, 20)
+		require.Error(t, err)
```
renaynay, line 390: yeah also assert which err here + to vlad comment

share/shwap/p2p/shrex/client.go
```diff
@@ -100,26 +100,28 @@ func (c *Client) doRequest(
...
+		return int64(statusLength),
+			statusReadStatusErr,
+			fmt.Errorf("unexpected error during reading the status from stream: %w", err)
 	}
 
 	switch statusResp.Status {
 	case shrexpb.Status_OK:
 	case shrexpb.Status_NOT_FOUND:
-		return statusNotFound, ErrNotFound
+		return 0, statusNotFound, ErrNotFound
```
renaynay: still return statusLen for metrics  (for all)

share/shwap/p2p/shrex/client.go
```diff
@@ -100,26 +100,28 @@ func (c *Client) doRequest(
...
 	default:
-		return statusReadRespErr, ErrInvalidResponse
+		return 0, statusReadRespErr, ErrInvalidResponse
 	}
 
-	_, err = resp.ReadFrom(stream)
+	dataLength, err := resp.ReadFrom(stream)
 	if err != nil {
-		return statusReadRespErr, fmt.Errorf("%w: %w", ErrInvalidResponse, err)
+		return 0, statusReadRespErr, fmt.Errorf("%w: %w", ErrInvalidResponse, err)
```
renaynay: dataLen + statusLen

share/shwap/accessor.go
```diff
@@ -13,5 +14,11 @@ import (
 type Accessor interface {
 	AxisRoots(context.Context) (*share.AxisRoots, error)
 	RowNamespaceData(context.Context, libshare.Namespace, int) (RowNamespaceData, error)
+	Sample(ctx context.Context, idx SampleCoords) (Sample, error)
```
renaynay, line 17: technically break but it's internal iface 🤷🏻 I don't know

share/shwap/namespace_data_id.go
renaynay: ?

walldiss: nvm, this is left over before we discussed how namespace data is sent to client.

share/shwap/p2p/shrex/shrex_getter/shrex_test.go
vgonkivs: current implementation does not support both `samples, error`. I think, such response are very confusing for the users. Also, error means, that we've requested ALL(or almost all) peers from the peer manager and they have responded with the error.

Made a mistake, naming this test case. My original idea, was to cover the case when request contains both valid and invalid cords, so the getter returns error only

share/shwap/range_namespace_data_id.go
vgonkivs: Custom reader implementation per each container instead of a buffer

share/shwap/p2p/shrex/shrex_getter/shrex.go
vgonkivs: good catch

share/shwap/p2p/shrex/shrex_getter/metrics.go
vgonkivs: removed

share/shwap/accessor.go
renaynay: yeah let's just label as a break for safe measure.

share/shwap/p2p/shrex/shrex_getter/shrex.go
```diff
@@ -143,105 +95,168 @@ func (sg *Getter) Stop(ctx context.Context) error {
...
+				if samples[i].IsEmpty() {
+					return errors.New("nil response")
+				}
+				return samples[i].Verify(header.DAH, request.RowIndex, request.ShareIndex)
+			}
+			err = sg.executeRequest(ctx, logger, header, request.Name(), req, verify)
+			if err != nil {
+				return err
+			}
+			return nil
```
renaynay:
```suggestion
			return sg.executeRequest(ctx, logger, header, request.Name(), req, verify)
```

renaynay (approved):

share/shwap/eds_id.go
```diff
@@ -24,6 +25,30 @@ func NewEdsID(height uint64) (EdsID, error) {
...
+		Height uint64 `json:"height"`
+	}{}
+
+	err := json.Unmarshal(data, &jsonEdsID)
+	if err != nil {
+		return err
+	}
+	eid.height = jsonEdsID.Height
+	return nil
+}
```
walldiss: Why do we need json encoding for id types? Shrex used to send binary encoded ids in and we should not add json encoding if we don't need to

share/shwap/p2p/shrex/client.go
```diff
@@ -50,21 +50,21 @@ func (c *Client) Get(
...
 	requestTime := time.Now()
-	status, err := c.doRequest(ctx, logger, req, resp, peer)
+	n, status, err := c.doRequest(ctx, logger, req, resp, peer)
 	if err != nil {
 		logger.Warnw("requesting data from peer failed", "error", err)
 	}
 	c.metrics.observeRequest(ctx, req.Name(), status, time.Since(requestTime))
 	logger.Debugw("requested data", "status", status, "duration", time.Since(requestTime))
-	return err
+	return n, err
```
walldiss: The returned amount of bytes (n) is unused by the caller. observe it here and don't return it further.

share/shwap/p2p/shrex/shrex_getter/shrex.go
```diff
@@ -143,105 +95,164 @@ func (sg *Getter) Stop(ctx context.Context) error {
...
+				}
+				return samples[i].Verify(header.DAH, request.RowIndex, request.ShareIndex)
+			}
+			return sg.executeRequest(ctx, logger, header, request.Name(), req, verify)
+		})
+	}
+	if err = errGroup.Wait(); err != nil {
+		return nil, err
+	}
+	return samples, nil
```
walldiss: Is it intended that this will not return partial response for sampling? We used to returned successful samples so next retry attempt don't have to re-request them

share/shwap/p2p/shrex/shrex_getter/shrex.go
```diff
@@ -143,105 +95,164 @@ func (sg *Getter) Stop(ctx context.Context) error {
...
+		return []shwap.Sample{}, nil
+	}
 
-func (sg *Getter) GetRow(_ context.Context, _ *header.ExtendedHeader, _ int) (shwap.Row, error) {
-	return shwap.Row{}, fmt.Errorf("getter/shrex: GetRow %w", shwap.ErrOperationNotSupported)
+	var err error
+	ctx, span := tracer.Start(ctx, "shrex/get-samples")
+	defer func() {
+		utils.SetStatusAndEnd(span, err)
+	}()
```
walldiss: There will be issues because you shadow `err` var once and in `sg.executeRequest(ctx, logger, header, request.Name(), req, verify)` call you don't assign returned error to this var

share/shwap/p2p/shrex/shrex_getter/shrex.go
```diff
@@ -143,105 +95,164 @@ func (sg *Getter) Stop(ctx context.Context) error {
...
+		errGroup.Go(func() error {
+			req := func(ctx context.Context, peer libpeer.ID) (int64, error) {
+				return sg.client.Get(ctx, &request, &samples[i], peer)
+			}
+			verify := func() error {
+				if samples[i].IsEmpty() {
+					return errors.New("nil response")
+				}
+				return samples[i].Verify(header.DAH, request.RowIndex, request.ShareIndex)
+			}
```
walldiss, line 144: captured `i` from the loop. Golang should be smart enough now to copy the var, but still worth to double check in tests

share/shwap/p2p/shrex/shrex_getter/shrex.go
```diff
@@ -143,105 +95,164 @@ func (sg *Getter) Stop(ctx context.Context) error {
...
-			setStatus(peers.ResultBlacklistPeer)
-		default:
-			setStatus(peers.ResultCooldownPeer)
+	build := func() error {
+		if buff.Len() == 0 {
+			return errors.New("nil response")
 		}
+		response, err = eds.ReadAccessor(ctx, buff, header.DAH)
+		return err
+	}
```
walldiss, line 245: Just to double check: build will verify internally hash equivalence?

share/shwap/p2p/shrex/shrex_getter/shrex.go
```diff
@@ -265,40 +276,175 @@ func (sg *Getter) GetNamespaceData(
...
+		toCoords, err := shwap.SampleCoordsFrom1DIndex(to-1, len(header.DAH.RowRoots)/2)
+		if err != nil {
+			return err
+		}
+
+		return response.VerifyInclusion(
+			fromCoords,
+			toCoords,
+			len(header.DAH.RowRoots)/2,
+			header.DAH.RowRoots[fromCoords.Row:toCoords.Row+1])
```
walldiss, line 376: Noticed there is `to-1` and `toCoord.Row+1` conversions. Do we have different schema for range coordinates on different levels? I suggest we unify it everywhere to be half open interval [from;to)

share/shwap/p2p/shrex/shrex_getter/shrex.go
```diff
@@ -265,40 +276,175 @@ func (sg *Getter) GetNamespaceData(
...
+		return p, df, err
+	}
+	return sg.fullPeerManager.Peer(ctx, header.DAH.Hash(), header.Height())
+}
+
+// requestFn defines a function type that wraps the actual request logic as a closure.
+// The closure captures additional request parameters and then executes
+// the request with the provided context and peer ID.
+// It returns the length of data that was read and an error
+type requestFn func(context.Context, libpeer.ID) (int64, error)
```
walldiss: Length of data returned is unused. I suggest we track it on shrex client side

share/shwap/p2p/shrex/shrex_getter/shrex_test.go
```diff
@@ -240,6 +240,141 @@ func TestShrexGetter(t *testing.T) {
...
+			DataHash: roots.Hash(),
+			Height:   height,
+		})
+
+		coords := []shwap.SampleCoords{
+			{Row: 0, Col: 0},
+			{Row: 0, Col: 1},
+			{Row: 0, Col: 2},
+			{Row: 0, Col: 3},
+		}
```
walldiss: nit: can just loop over all samples to make sure there are no edge cases.

share/shwap/p2p/shrex/shrex_getter/shrex_test.go
```diff
@@ -240,6 +240,141 @@ func TestShrexGetter(t *testing.T) {
...
+		err = edsStore.PutODSQ4(ctx, roots, height, randEDS)
+		require.NoError(t, err)
+		fullPeerManager.Validate(ctx, srvHost.ID(), shrexsub.Notification{
+			DataHash: roots.Hash(),
+			Height:   height,
+		})
+
+		coords := []shwap.SampleCoords{
+			{Row: 0, Col: 10},
+		}
```
walldiss, line 304: nit: -check that requesting samples with ods_size+1 return error. It will help to capture edge cases with test

share/shwap/p2p/shrex/shrex_getter/shrex_test.go
walldiss: You retry all samples at once and if one failed the attempt will retry all again. Instead client should retry only one failed sample and if one sample is unavailable return those that are available so on next attempts they not longer requested

share/shwap/p2p/shrex/shrex_getter/shrex_test.go
```diff
@@ -240,6 +240,141 @@ func TestShrexGetter(t *testing.T) {
...
+			Height:   height,
+		})
+
+		row, err := getter.GetRow(ctx, eh, 0)
+		require.NoError(t, err)
+		shrs, err := row.Shares()
+		require.NoError(t, err)
+		edsRow := randEDS.Row(0)
+		for i, shr := range shrs {
+			require.Equal(t, edsRow[i], shr.ToBytes())
```
walldiss: Same here. 
- Loop over all rows
- `ods_size+1` edge cases

share/shwap/range_namespace_data_id.go
walldiss: Maybe rewrite this comment or remove it , because it does not provide any information for unaware reader

share/shwap/p2p/shrex/shrex_getter/shrex_test.go
vgonkivs: > You retry all samples at once and if one failed the attempt will retry all again

if you are talking about the client, then NO bc execute request contains an infinite for loop that will try to request a sample from an available peer in PeerManager. 
So, the request(of a single sample) fails only if context was canceled or deadline exceeded. Do you want me to return a partial response in this case?

share/shwap/eds_id.go
vgonkivs: removed

share/shwap/p2p/shrex/client.go
vgonkivs: added additional metric for total payload requested
cc @renaynay

share/shwap/p2p/shrex/shrex_getter/shrex.go
vgonkivs: Responded above. `sg.executeRequest` will try to request a single sample from all peers that are available in PM. It can be failed with context.Canceled or context.DeadlineExceeded.

share/shwap/p2p/shrex/shrex_getter/shrex.go
vgonkivs: go now creates a copy for every iteration(1.22+).

share/shwap/p2p/shrex/shrex_getter/shrex.go
vgonkivs:
```
	if err = errGroup.Wait(); err != nil {
		return nil, err
	}
```

But it should be assigned here

share/shwap/p2p/shrex/shrex_getter/shrex.go
vgonkivs: Yes, ReadAccessor works in two steps:
1) read all shares from the provided reader until eof
2) verify that the EDS hash matches the expected hash

share/shwap/p2p/shrex/shrex_getter/shrex.go
vgonkivs: `to` is an index and `toCoords` is a coordinate. Since we have to have an exact coordinate(not half opened), `to-1` was used. If I looks confusing, then I can change namings

share/shwap/p2p/shrex/shrex_getter/shrex_test.go
vgonkivs: I don't think we should limit users of requesting ods shares only. SampleID contructor was designed to verify against EDS, so I'd suggest to keep it.

share/shwap/p2p/shrex/shrex_getter/shrex_test.go
vgonkivs: `ods_size+1 edge cases `-> the use should still requests parity data if he wants to

vgonkivs: And we already have `ErrOutofBounds` test case

share/shwap/p2p/shrex/shrex_getter/shrex_test.go
walldiss: yes. need to return succesful samples similar to existing bitswap getter implementation

share/shwap/p2p/shrex/shrex_getter/metrics.go
```diff
@@ -0,0 +1,42 @@
...
+
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+
+	"github.com/celestiaorg/celestia-node/libs/utils"
+)
+
+type metrics struct {
+	requestAttempt metric.Int64Histogram
+}
```
walldiss, line 14: This will be not great for visialisation and outside of scope of the PR. Lets remove

share/shwap/range_namespace_data.go
```diff
@@ -471,3 +472,31 @@ func GenerateSharesProofs(
...
+func (rngdata *RangeNamespaceData) WriteTo(writer io.Writer) (int64, error) {
+	pbrngData := rngdata.ToProto()
+
+	b, err := pbrngData.Marshal()
+	if err != nil {
+		return 0, err
+	}
+	l, err := writer.Write(b)
+	return int64(l), err
+}
```
walldiss, line 493: From private discussion: Need to split into rows and sent them one by one similar to ND data.

share/shwap/p2p/shrex/shrex_getter/shrex_test.go
walldiss: you are right. I mean eds_size+1 which is edge case

store/store.go
```diff
@@ -32,6 +32,15 @@ const (
 
 var ErrNotFound = errors.New("eds not found in store")
 
+// AccessorGetter abstracts storage system that indexes and manages multiple eds.AccessorGetter by
+// network height.
+type AccessorGetter interface {
```
renaynay, line 37: I don't love this iface for one implementation but i'm also not gonna die on hill

vgonkivs: I know but I've added it to be able to properly test some edge cases

share/shwap/range_namespace_data.go
vgonkivs: it requires changing in proto definition that will break the protocol. Current approach also works. Also NS data is a []RowNsData and RangeNsData contains a set of shares + 2 proofs(that can be empty)

share/shwap/p2p/shrex/shrex_getter/metrics.go
vgonkivs: removed

PR #4298, opened 2025-06-04
fix(test): extendBlock tests

### LINKED ISSUE #4252: Unit test for `extendBlock` on historical mainnet blocks
## Context

https://github.com/celestiaorg/celestia-node/pull/4219/files#r2046888052

## Proposal

It would increase confidence if celestia-node had a test whereby a few historical mainnet blocks were saved and stored in `testdata/` and then `extendBlock` was invoked on them to verify that the generated EDS data root matches what was produced on mainnet for that block. It's worth including a few sample blocks that span app versions (v1, v2, v3) and span a few different types of tx data (regular blobs, authored blobs, empty block).

moldis:
- closes #4252
- Added unit-tests for extendBlock function, the tests data grabbed from main-net for app_version 1, 2 and 3.
- kind:testing

### DIFF
--- a/core/eds_test.go
+++ b/core/eds_test.go
@@ -180,4 +180,3 @@ func TestExtendBlock_MainnetBlocks(t *testing.T) {
 		})
 	}
 }
-


renaynay:
Thank you @moldis for making this PR - it's great and I think it would make the most sense once we do https://github.com/celestiaorg/celestia-app/issues/5063

renaynay (approved):

Issue #4305, opened 2025-06-04
feat(share): Implement Proof to Dataroot Structure


walldiss:
Replace the current per-row proof list (implemented only for Range container right now) with an dataroot proof structure that combines row root proofs into a single Merkle proof and minimizes NMT proofs to only edge boundaries.

### Current Structure Issues
- Redundant proof for each row independently
- Unnecessary NMT proofs for complete rows
- Inefficient transmission size

### New Optimized Structure
```go
type DatarootProof struct {
    // Single Merkle proof for all contiguous row roots
    RowRootsProof *merkle.Proof `json:"row_roots_proof"`

    // NMT proof only for left boundary (if partial row)
    LeftEdgeProof *nmt.Proof `json:"left_edge_proof,omitempty"`

    // NMT proof only for right boundary (if partial row)
    RightEdgeProof *nmt.Proof `json:"right_edge_proof,omitempty"`
}
```

Implement the verification algorithm for the new dataroot proof structure that rebuilds row roots and verifies inclusion in the data root.

### Verification Steps
1. Build row roots for each row in the range:
   - For complete rows: Calculate directly from share data
   - For partial rows: Use NMT edge proofs to verify and build roots
2. Verify inclusion of all row roots using single Merkle proof to data root

### Acceptance Criteria

- [ ] Add comprehensive unit tests for all edge cases
- [ ] Performance benchmarks showing verification time improvement

### Acceptance Criteria
- [ ] mplement `DatarootProof` structure
- [ ] Implement logic to determine when edge proofs are needed (full rows don't need edge proof)
- [ ] Implement `Verify()` method for `DatarootProof`
  - [ ] Support verification of complete rows without NMT proofs
  - [ ] Support verification of partial rows using edge proofs
  - [ ] Implement row root calculation from share data
- [ ] Document proof structure and verification algorithm

PR #4314, opened 2025-06-05
refactor(share/proof): optimise  data root inclusion proof for ranges

vgonkivs:
Closes https://github.com/celestiaorg/celestia-node/issues/4305

The PR can be easily merged as the new version of the proof is not used on the client side to avoid breaking the existing API. It will be enabled only after adding a new version of the node's API that serves data root proofs.


vgonkivs:
Agreed with @walldiss to split this PR into 2:
1) Data Root proofs
2) integration in RangeNSData
Moving to draft until 1. is merged

PR #4318, opened 2025-06-09
feat(api): Celestia client lib (v4 migration)

walldiss:
Forward port of https://github.com/celestiaorg/celestia-node/pull/4265

--------

## Implementation Details

### Key Components

### Usage Example

QuickNode example by @jcstein :https://github.com/celestiaorg/celestia-node/pull/4256#issuecomment-2852085985

```go
// Initialize keyring
    KeyName:     "my_key",
}, "/path/to/keys")

// Configure client
cfg := Config{
    ReadConfig: ReadConfig{
        DAAuthToken:  "token",
    },
    SubmitConfig: SubmitConfig{
        DefaultKeyName: "my_key",
        Network:        "private",
            Addr: "consensus-node:9090",
        },
    },
}

// Submit a blob

// Retrieve a blob
```

# TODOs for Further Development / Known Limitations (will be resolved in future PRs)

- **Multiple core endpoints**: will be added after https://github.com/celestiaorg/celestia-app/pull/4628 support added to the node core-accessor. 

## Decision Points

- **Configuration Options**:


vgonkivs (dismissed):

api/client/read_client.go
```diff
@@ -0,0 +1,113 @@
...
+		return nil, fmt.Errorf("failed to initialize fraud client: %w", err)
+	}
+
+	// Initialize blob read client
+	blobAPI := blobapi.API{}
+	blobCloser, err := jsonrpc.NewClient(
+		ctx, cfg.BridgeDAAddr, "blob", &blobAPI.Internal, cfg.HTTPHeader)
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize blob client: %w", err)
+	}
```
Wondertan, line 103: The need to initialize them individually stems from the requirement to stub certain APIs, right? The approach we have in the RPC could work here too, where we just have a map of modules that we loop over, creating default impls, and then we could override certain modules with a stub depending on the use case.

api/client/read_client.go
```diff
@@ -0,0 +1,113 @@
...
+// Bridge node json-rpc connection config
+type ReadConfig struct {
+	// BridgeDAAddr is the address of the bridge node
+	BridgeDAAddr string
+	// DAAuthToken sets the value for Authorization http header
+	DAAuthToken string
+	// HTTPHeader contains custom headers that will be sent with each request
+	HTTPHeader http.Header
+	// EnableDATLS enables TLS for bridge node
+	EnableDATLS bool
```
Wondertan, line 27: Is this a stub? I couldn't find this being used anywhere

api/client/client.go
```diff
@@ -0,0 +1,148 @@
...
+	blobstreamapi "github.com/celestiaorg/celestia-node/nodebuilder/blobstream"
+	fraudapi "github.com/celestiaorg/celestia-node/nodebuilder/fraud"
+	headerapi "github.com/celestiaorg/celestia-node/nodebuilder/header"
+	"github.com/celestiaorg/celestia-node/nodebuilder/p2p"
+	shareapi "github.com/celestiaorg/celestia-node/nodebuilder/share"
+	stateapi "github.com/celestiaorg/celestia-node/nodebuilder/state"
+	"github.com/celestiaorg/celestia-node/state"
+)
+
+var log = logging.Logger("celestia-client")
```
Wondertan, line 18: Is this only used to log that TLSes are not enabled?

api/client/client.go
```diff
@@ -0,0 +1,148 @@
...
+	Fraud      fraudapi.Module
+	Blobstream blobstreamapi.Module
+
+	chainCloser func() error
+}
+
+// New initializes the Celestia client. It connects to the Celestia consensus nodes and Bridge
+// nodes. Any changes to the keyring are not visible to the client. The client needs to be
+// reinitialized to pick up new keys.
+func New(ctx context.Context, cfg Config, kr keyring.Keyring) (*Client, error) {
```
Wondertan, line 57: optional nit: mention that calling Close after usage is a must

Wondertan (commented):
Mostly questions. One more general question I have is regarding the WalletAPI that we discussed as an option to be rolled out with the client lib. What your thinking about it now?

Also, the Read mode reminds me of the existing rpc client. Couldn't we just compose over it?

api/client/read_client.go
walldiss: Does not require complication with generalised solution for now

api/client/read_client.go
walldiss: Not allowed by BN server side.

api/client/wrappers.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	return api.submitter.Submit(ctx, blobs, options)
+}
+
+// disabledStateAPI returns an error on all state client methods if consensus connection is not
+// configured
+type disabledStateAPI struct{}
+
+func (r *disabledStateAPI) AccountAddress(context.Context) (state.Address, error) {
+	return state.Address{}, ErrReadOnlyMode
```
renaynay: why are we returning read only mode err for completely disabled API ? Err would be confusing, let's make new err for misconfigured / not configured consensus connection?

api/client/keyring.go
```diff
@@ -0,0 +1,65 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+
+	"github.com/celestiaorg/celestia-app/v4/app"
+	"github.com/celestiaorg/celestia-app/v4/app/encoding"
+)
+
+// KeyringConfig contains configuration parameters for constructing
+// the node's keyring signer.
+type KeyringConfig struct {
```
renaynay, line 19: would be nice if this were a subconfig of state config so we could dedup

api/client/keyring.go
```diff
@@ -0,0 +1,65 @@
...
+	KeyName     string
+	BackendName string
+}
+
+// KeyringWithNewKey is a helper function for easy keyring initialization. It initializes a keyring
+// with the given config and path. It creates a new key if no key is found in the keyring store.
+func KeyringWithNewKey(cfg KeyringConfig, path string) (keyring.Keyring, error) {
```
renaynay, line 26: any way to dedup this with other key generation / check code?

renaynay (commented):
generally fine with me and seems pretty small and minimal. Some duplicated logic but guess that's okay for first iteration. Not gonna nit too hard on that.

api/client/wrappers.go
walldiss: Good point. This is artefact of having single constructor for both read and read+submit clients. ReadClient has dedicated constructor now, so there are better ways to prevent State module access. Disabled it by removing State module from ReadClient here: https://github.com/celestiaorg/celestia-node/pull/4318/commits/adfe4d38164d1e67b22410543b8514377b9e286d

api/client/keyring.go
walldiss: I want to have it as sufficient starter pkg for now. We also might need it as copy here if celestia lib would be decoupled from node and its dependancies.

PR #4324, opened 2025-06-13
feat(share/shrex): add context cancellation to shrexeds and shrexnd clients

### LINKED ISSUE #3480: feat(share/shrex): shrex client should abort requesting on ctx.Cancel()
### Implementation ideas

Currently shrex clients (shrex_eds,shrex_nd) consumes parent context, but if parent context is canceled, both clients will ignore canceleation and continue to reqest data. Need to add support for fast return in case of canceled parent context

TheRealSibasishBehera:
This PR implements context cancellation support for shrex clients to fix hanging requests when parent
  context is canceled.

  Closes #3480

  ## Changes

  - **shrexeds client**: Added `readEDSWithContext()` method with goroutine+channel pattern for context cancellation
  - **shrexnd client**: Added `readNamespaceDataWithContext()` method with similar cancellation support
  - **share/eds**: Updated `ReadShares()` to accept context and check for cancellation in reading loop
  - **store/file**: Added context propagation through ODS and square reading functions
  - **tests**: Added tests for quick server response cancellation and data reading cancellation cases

### DIFF
--- a/share/shwap/p2p/shrex/shrexeds/client.go
+++ b/share/shwap/p2p/shrex/shrexeds/client.go
@@ -136,8 +136,7 @@ func (c *Client) doRequest(
 	case shrexpb.Status_OK:
 		// reset stream deadlines to original values, since read deadline was changed during status read
 		c.setStreamDeadlines(ctx, stream)
-		// use header and ODS bytes to construct EDS and verify it against dataHash
-		eds, err := eds.ReadAccessor(ctx, stream, root)
+		eds, err := c.readEDSWithContext(ctx, stream, root)
 		if err != nil {
 			return nil, fmt.Errorf("read eds from stream: %w", err)
 		}
@@ -154,6 +153,30 @@ func (c *Client) doRequest(
 	default:
 		c.metrics.ObserveRequests(ctx, 1, shrex.StatusInternalErr)
 		return nil, shrex.ErrInvalidResponse
+	}
+}
+
+// readEDSWithContext reads EDS data with context cancellation support.
+func (c *Client) readEDSWithContext(
+	ctx context.Context, stream network.Stream, root *share.AxisRoots,
+) (*eds.Rsmt2D, error) {
+	type result struct {
+		eds *eds.Rsmt2D
+		err error
+	}
+	resultCh := make(chan result, 1)
+
+	go func() {
+		eds, err := eds.ReadAccessor(ctx, stream, root)
+		resultCh <- result{eds: eds, err: err}
+	}()
+
+	select {
+	case res := <-resultCh:
+		return res.eds, res.err
+	case <-ctx.Done():
+		stream.Reset() //nolint:errcheck
+		return nil, ctx.Err()
 	}
 }
 
--- a/share/shwap/p2p/shrex/shrexnd/client.go
+++ b/share/shwap/p2p/shrex/shrexnd/client.go
@@ -117,13 +117,41 @@ func (c *Client) doRequest(
 		return nil, err
 	}
 
-	nd := shwap.NamespaceData{}
-	_, err = nd.ReadFrom(stream)
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	default:
+	}
+
+	nd, err := c.readNamespaceDataWithContext(ctx, stream)
 	if err != nil {
 		c.metrics.ObserveRequests(ctx, 1, shrex.StatusReadRespErr)
 		return nil, err
 	}
 	return nd, nil
+}
+
+// readNamespaceDataWithContext reads namespace data with context cancellation support.
+func (c *Client) readNamespaceDataWithContext(ctx context.Context, stream network.Stream) (shwap.NamespaceData, error) {
+	type result struct {
+		nd  shwap.NamespaceData
+		err error
+	}
+	resultCh := make(chan result, 1)
+
+	go func() {
+		nd := shwap.NamespaceData{}
+		_, err := nd.ReadFrom(stream)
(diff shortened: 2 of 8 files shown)


share/shwap/p2p/shrex/shrexnd/client.go
```diff
@@ -117,15 +117,43 @@ func (c *Client) doRequest(
...
+		nd := shwap.NamespaceData{}
+		_, err := nd.ReadFrom(stream)
+		resultCh <- result{nd: nd, err: err}
+	}()
+
+	select {
+	case res := <-resultCh:
+		return res.nd, res.err
+	case <-ctx.Done():
+		stream.Reset() //nolint:errcheck
```
Wondertan: Lets put reset in the readNamespaceData error handling path. We should reset it for any error we see and we actually dont do it. A good opportunity to fix it

share/shwap/p2p/shrex/shrexnd/client.go
```diff
@@ -117,15 +117,43 @@ func (c *Client) doRequest(
...
+	nd, err := c.readNamespaceDataWithContext(ctx, stream)
 	if err != nil {
 		c.metrics.ObserveRequests(ctx, 1, shrex.StatusReadRespErr)
 		return nil, err
 	}
 	return nd, nil
 }
 
+// readNamespaceDataWithContext reads namespace data with context cancellation support.
+func (c *Client) readNamespaceDataWithContext(ctx context.Context, stream network.Stream) (shwap.NamespaceData, error) {
```
Wondertan: nit: avoid WithContext, it doesn't actually add anything but bloay

share/shwap/p2p/shrex/shrexnd/client.go
```diff
@@ -117,15 +117,43 @@ func (c *Client) doRequest(
 		return nil, err
 	}
 
-	nd := shwap.NamespaceData{}
-	_, err = nd.ReadFrom(stream)
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	default:
+	}
```
Wondertan: Lets move it inside the new func. Then it will go through the same error handling path with reset

share/shwap/p2p/shrex/shrexeds/client.go
```diff
@@ -157,6 +156,30 @@ func (c *Client) doRequest(
...
+	go func() {
+		eds, err := eds.ReadAccessor(ctx, stream, root)
+		resultCh <- result{eds: eds, err: err}
+	}()
+
+	select {
+	case res := <-resultCh:
+		return res.eds, res.err
+	case <-ctx.Done():
+		stream.Reset() //nolint:errcheck
```
Wondertan: Same here: move reset outside

share/shwap/p2p/shrex/shrexnd/client.go
TheRealSibasishBehera: removed it completely in the refactor , as the duration between two context cancellation checks would have been too small to check

share/shwap/p2p/shrex/shrexeds/client.go
```diff
@@ -157,6 +157,30 @@ func (c *Client) doRequest(
...
 
+// readEDS reads EDS data with context cancellation support.
+func (c *Client) readEDS(
+	ctx context.Context, stream network.Stream, root *share.AxisRoots,
+) (*eds.Rsmt2D, error) {
+	type result struct {
+		eds *eds.Rsmt2D
+		err error
+	}
+	resultCh := make(chan result, 1)
```
Wondertan, line 168: buffer is not needed

share/shwap/p2p/shrex/shrexeds/client.go
```diff
@@ -157,6 +157,30 @@ func (c *Client) doRequest(
...
+
+	go func() {
+		defer close(resultCh)
```
Wondertan, line 171: nit: not necessary

share/shwap/p2p/shrex/shrexeds/client.go
TheRealSibasishBehera: I am thinking of the case 
t= 0 goroutine is fired
t=2 context is cancelled  and function has returned
t =4 it tries to send using the ch
select is not waiting to receive the value from ch. In this case goroutine is blocked I think

Wondertan: Right, handling context on write side is missing

Wondertan:
Streams already have a deadline set from the context! What's missing is handling the cancellation from manual `cancel`, and there is no way to do it without a goroutine.

TheRealSibasishBehera:
I can see that the read deadline is reset to a server parameter value after write gets completed , due to this the parent context wont be respected

Is it perhaps this part that the issue originally aims to solve rather than the manual`cancel` call ? 
```
share/shwap/p2p/shrex/shrexeds/client.go

	// read and parse status from peer
	resp := new(shrexpb.Response)
	err = stream.SetReadDeadline(time.Now().Add(c.params.ServerReadTimeout))
	if err != nil {
		log.Debugw("client: failed to set read deadline for reading status", "err", err)
	}
```

vgonkivs:
Hey @TheRealSibasishBehera. We have unified shrexnd and shrexeds components and replaced them with shrex client/server parts that implement generic logic for handling all types of shwap requests. Do you still want to continue working on this issue and prefer rebase on main(fixing all conflicts) or open a new PR?

TheRealSibasishBehera:
Thanks for the update @vgonkivs
I will start working on a fresh PR from the main branch, as it will be easier given the refactor

PR #4325, opened 2025-06-13
refactor!(share/shwap): rework range namespace data

vgonkivs:
This PR refactors the range namespace data implementation, optimizing how proofs are handled for incomplete rows in namespace data retrieval. The changes focus on reducing redundant proof data.

closes https://github.com/celestiaorg/celestia-node/issues/4339


share/shwap/pb/shwap.proto
```diff
@@ -35,9 +36,10 @@ message RowShares {
 }
 
 message RangeNamespaceData {
-    int32 start = 1;
-    repeated RowShares shares = 2;
-    repeated Proof proofs = 3;
+    int64 StartRow  = 1;
+    int64 EndRow  = 2;
+    repeated RowShares shares = 3;
+    share.proof.IncompleteRowsProof proof = 4;
```
walldiss: Can we use StartRow and EndRow from request? If not, what would be reason to send them back?

walldiss: We don't need dedicated IncompleteRowsProof type. Just use 2 fields of nmt.Proof

share/eds/accessor.go
```diff
@@ -37,9 +37,8 @@ type Accessor interface {
 	// from the Accessor. Response might have only proofs.
 	RangeNamespaceData(
 		ctx context.Context,
-		ns libshare.Namespace,
 		from, to shwap.SampleCoords,
-	) (shwap.RangeNamespaceData, error)
+	) (*shwap.RangeNamespaceData, error)
```
walldiss: All other methods return values and not the pointer. Should this method be returning value aswell?

vgonkivs: No, `shwap.RangeNamespaceData` can be huge, we discussed this in previous PR. So let's keep it as pointer

share/shwap/pb/shwap.proto
vgonkivs: I know. But IMO it looks much cleaner and clear if we put it inside the separate structure.

vgonkivs: it helps during the validation. we can verify that returned range corresponds to the request.

vgonkivs: we reduced range size a lot, so I don't think it's a big problem for keeping 2 integers

share/eds/accessor.go
walldiss: It contains few slices, which are small pointer based structs. Same size as namespaceData

share/shwap/pb/shwap.proto
walldiss: It creates additional abstraction where we can do just fine without it. It will be easier to read without extra abstraction

walldiss: We don't need those fields if the only reason we have them is just to validate them on client side. Proof verification is the only way we can know the response is valid

vgonkivs: reworked

vgonkivs: reworked

share/eds/accessor.go
vgonkivs: reverted back to value type

share/eds/accessor.go
```diff
@@ -37,7 +37,6 @@ type Accessor interface {
 	// from the Accessor. Response might have only proofs.
 	RangeNamespaceData(
 		ctx context.Context,
-		ns libshare.Namespace,
```
renaynay, line 40: so we are removing ns because we want to make it possible to request range from inside square regardless of ns? Just purely coords based?

share/shwap/range_namespace_data_id.go
```diff
@@ -127,7 +117,7 @@ func (rngid RangeNamespaceDataID) WriteTo(w io.Writer) (int64, error) {
 
 // Equals checks equality of RangeNamespaceDataID.
 func (rngid *RangeNamespaceDataID) Equals(other RangeNamespaceDataID) bool {
-	return rngid.DataNamespace.Equals(other.DataNamespace) && rngid.From == other.From &&
+	return rngid.EdsID == other.EdsID && rngid.From == other.From &&
```
renaynay: some nice finds here

share/shwap/sample_id.go
```diff
@@ -15,6 +15,10 @@ type SampleCoords struct {
 	Col int `json:"col"`
 }
 
+func (s SampleCoords) String() string {
+	return fmt.Sprintf("[%d:%d]", s.Row, s.Col)
```
renaynay:
```suggestion
	return fmt.Sprintf("(%d:%d)", s.Row, s.Col)
```
nit

renaynay (commented):
* pls mark PR as breaking as it breaks accessor + API
* godocs for `shwap.RangeNamespaceDataFromShares` need to be updated 
* godocs for `RangeNamespaceDataIDSize` type need to be updated
* side note but the name of the container now doesn't make sense as we've omitted ns -- since we are breaking anyway, should we shorten it? `RangeData`
* If we are doing coords-based range only, then lets change method names to omit Namespace as it’s confusing for user
* Should we add opts to range request at API level so we can add WithProofOnly later without breaking?


some more comments here because I couldn't do it while offline (sorry for the inconvenience):



* What is an “incomplete row proof” ? And does it apply to EVERY start index / end index in the row? What if I range request one row (completely)? 
get_range_result.go

* Wait I'm confused now, do we allow range containers for pure range (coords-based only) or do we only allow range that is within bounds of one ns? 
* Use `odsSize` on L65 L66
* Should we sanity cross-check dah root against merkle root calculated via `ProofsFromByteSlices`? Seems like cheap check since we already gen proofs there
* Double proof verification in GetRangeResult.Verify (both in verify explicit call and Validate) 

range_namespace_data.go

* L291: flip expected and got values
* Where is VerifyNamespace used? Can we simplify VerifyShares then and reduce methods in this file? VerifyInclusion and VerifyNamespace can call verify inclusion with additional namespace completeness check in the future if we need it. Very confusing, all the diff methods with diff naming - hard to follow quickly what’s doing what
* Fix godoc for RangeNamespaceDataFromShares
* isMultiRow on L73 can check if numRows > 1
* What Is the point of isSingleRow ? just use isMultiRow or just use one of them lol

share/eds/accessor.go
vgonkivs: > Just purely coords based?

Yes, assuming the user knows the namespace(similarly how he knows the indexes)

nodebuilder/share/types/get_range_result.go
```diff
@@ -0,0 +1,126 @@
...
+		)
+		if err != nil {
+			return nil, err
+		}
+		nmtProofs[i] = proof
+	}
+	coreProofs := toCoreNMTProof(nmtProofs)
+
+	root, allProofs := merkle.ProofsFromByteSlices(append(dah.RowRoots, dah.ColumnRoots...))
+	if !bytes.Equal(root, dah.Hash()) {
```
gupadhyaya: this check feels unnecessary as both root and dah.Hash are based on dah. only seems to be validating the correctness of ProofsFromByteSlices everytime?

gupadhyaya: is there a possibility of dah being malformed?

nodebuilder/share/types/get_range_result.go
```diff
@@ -0,0 +1,126 @@
...
+		return nil, errors.New("recomputed data hash does not match expected data hash")
+	}
+
+	rowProofs := make([]*merkle.Proof, endRow-startRow+1)
```
gupadhyaya: numRows :=  endRow-startRow+1

nodebuilder/share/types/get_range_result.go
```diff
@@ -0,0 +1,126 @@
...
+			Total:    allProofs[i].Total,
+			Index:    allProofs[i].Index,
+			LeafHash: allProofs[i].LeafHash,
+			Aunts:    allProofs[i].Aunts,
+		}
+		rowRoots[i-startRow] = dah.RowRoots[i]
+	}
+
+	sharesProof := &types.ShareProof{
+		Data:             libshare.ToBytes(rngdata.Flatten()),
```
gupadhyaya: [nit] could pull out `rngdata.Flatten()` to a variable to avoid repeated computation.

nodebuilder/share/types/get_range_result.go
```diff
@@ -0,0 +1,126 @@
...
+		NamespaceVersion: uint32(ns.Version()),
+		RowProof: types.RowProof{
+			RowRoots: rowRoots,
+			Proofs:   rowProofs,
+			StartRow: uint32(startRow),
+			EndRow:   uint32(endRow),
+		},
+	}
+	return &GetRangeResult{
+		Shares: rngdata.Flatten(),
```
gupadhyaya: reuse the flattened shares from above.

share/shwap/p2p/bitswap/range_namespace_data_block_test.go
```diff
@@ -81,7 +81,7 @@ func TestRangeNamespaceData_FetchRoundtrip(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			blk, err := NewEmptyRangeNamespaceDataBlock(1, tc.ns, tc.from, tc.to, 16, false)
+			blk, err := NewEmptyRangeNamespaceDataBlock(1, tc.from, tc.to, 16/2)
```
gupadhyaya: [nit] just use 8 instead of 16/2

share/shwap/range_namespace_data.go
```diff
@@ -1,219 +1,186 @@
...
+//   - A RangeNamespaceData containing the shares and Merkle proofs for the incomplete rows.
 //   - An error if the range is invalid or extraction fails.
-func RangeNamespaceDataFromShares(
-	shares [][]libshare.Share,
-	namespace libshare.Namespace,
-	axisRoots *share.AxisRoots,
-	from, to SampleCoords,
-) (RangeNamespaceData, error) {
+func RangeNamespaceDataFromShares(shares [][]libshare.Share, from, to SampleCoords) (RangeNamespaceData, error) {
 	if len(shares) == 0 {
```
gupadhyaya:
```
if len(shares) == 0 || len(shares[0]) == 0 {
	return RangeNamespaceData{}, errors.New("shares contain no rows or empty row")
}
```

share/shwap/range_namespace_data.go
```diff
@@ -96,73 +84,70 @@ func RangeNamespaceDataFromShares(
...
+			endCol,
+			odsSize,
+			shares[0],
+		)
 		if err != nil {
-			return RangeNamespaceData{}, fmt.Errorf("failed to generate proof for row %d: %w", from.Row, err)
+			return nil, fmt.Errorf("failed to generate proof for row %d: %w", from.Row, err)
 		}
-		rngData.Proof[0].sharesProof = sharesProofs
+		shares[0] = shares[0][from.Col:endCol]
```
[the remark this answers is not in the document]
gupadhyaya: may be a good idea to leave a comment in the godoc that input shares will be modified with ....

share/shwap/range_namespace_data.go
```diff
@@ -1,219 +1,186 @@
...
-func (rngdata *RangeNamespaceData) Verify(
-	namespace libshare.Namespace,
+// VerifyInclusion checks whether the shares stored within the RangeNamespaceData (`rngdata.Shares`)
+// are valid and provably included in the provided rowRoots.
+func (rngdata *RangeNamespaceData) VerifyInclusion(
 	from SampleCoords,
 	to SampleCoords,
-	dataHash []byte,
+	odsSize int,
+	roots [][]byte,
```
gupadhyaya, line 148: so the expectation now changed for verification, that verify does have access to data roots? could you elaborate as to why?

gupadhyaya: not just datahash

share/shwap/range_namespace_data.go
```diff
@@ -1,219 +1,186 @@
...
+func (rngdata *RangeNamespaceData) VerifyNamespace(
+	from SampleCoords,
+	to SampleCoords,
+	odsSize int,
+	roots [][]byte,
+) error {
+	return rngdata.verifyShares(rngdata.Shares, from, to, odsSize, roots, true)
 }
 
 // VerifyShares verifies that the provided 2D slice of shares is valid and correctly
```
gupadhyaya: you can get rid of this godoc as the function is private. may be leave a shorter highlight of the function.

share/shwap/range_namespace_data_id_test.go
```diff
@@ -26,14 +25,11 @@ func TestNewRangeNamespaceDataID(t *testing.T) {
...
 		t.Run(tc.name, func(t *testing.T) {
-			ns := libshare.RandomNamespace()
 			rngid, err := NewRangeNamespaceDataID(
 				EdsID{1},
-				ns,
 				tc.from,
 				tc.to,
-				tc.total,
-				true,
+				tc.total/2,
```
gupadhyaya, line 31: [nit] you can have total changed to 8

nodebuilder/share/types/get_range_result.go
vgonkivs: Just a sanity check 🙂

share/shwap/range_namespace_data.go
vgonkivs: Yes, the reason is that current structure will be used only on the P2P level, which means that we don't need to verify up to the data root. The client is requesting a range and verify proofs(since he has access to the DAH). But then we transform the structure of proof being able to be proven up to the data root. 
Basically, VerifyInlusion/Namespace will be used by our client after getting the data from server.

share/shwap/range_namespace_data.go
vgonkivs: will revisit the go doc and simplify it. Thanks.

share/shwap/range_namespace_data.go
```diff
@@ -390,9 +350,115 @@ func RangeCoordsFromIdx(edsIndex, length, edsSize int) (SampleCoords, SampleCoor
...
+	}
+
+	if endShare-startShare != sharesAmount {
+		return libshare.Namespace{}, fmt.Errorf(
+			"mismatch shares amount expected %d got %d", endShare-startShare, sharesAmount,
+		)
+	}
+
+	startShareNs := rawShares[0][0].Namespace()
+	for i, rowShares := range rawShares {
```
gupadhyaya, line 426: do you have to validate here? this seems expensive to do and also computeRoot as part of completeness validates it anyways.

gupadhyaya: i would totally get rid of ParseNamespace function all together and simply take it from the first share and rely on computeRoot to validate it.

share/shwap/range_namespace_data_id_test.go
vgonkivs: we can add a case with the bigger size and it fails in this case. let's avoid magic number

share/shwap/range_namespace_data.go
vgonkivs: It contains some basic validation logic of the range. I'll try to extract it somewhere(ParseNamespace is also used in get_range_result.go)

vgonkivs: but from the other side - bytes.Equal does not seem that expensive.

vgonkivs: I'll leave it as an additional level of verification if you are ok with this.

nodebuilder/share/types/get_range_result.go
vgonkivs: reused

nodebuilder/share/types/get_range_result.go
renaynay: I'm okay to leave it or keep it, no big deal since it gets calculated anyway from merklisation.

vgonkivs: Let's remove then.

nodebuilder/share/types/get_range_result.go
```diff
@@ -0,0 +1,124 @@
+package share_types //nolint:revive
```
renaynay: one general comment re types made purely for the API -- maybe we should have package name include `api` to differentiate there? I'm not really sure what the best thing to do is but I feel like it should be differentiated from the genuine types we have defined in node.

renaynay (approved):

share/shwap/range_namespace_data.go
gupadhyaya: ok for now.

nodebuilder/share/types/get_range_result.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/celestiaorg/celestia-node/share/shwap"
+)
+
+// GetRangeResult wraps the return value of the GetRange endpoint
+// because Json-RPC doesn't support more than two return values.
+type GetRangeResult struct {
+	Shares []libshare.Share
+	Proof  *types.ShareProof
+}
+
```
gupadhyaya: don't you need a godoc here?

share/shwap/range_namespace_data.go
```diff
@@ -390,9 +326,115 @@ func RangeCoordsFromIdx(edsIndex, length, edsSize int) (SampleCoords, SampleCoor
...
+				return libshare.Namespace{}, fmt.Errorf(
+					"shares range contain different namespaces at index [%d:%d] %v ", i, j, ns,
+				)
+			}
+		}
+	}
+	return startShareNs, nil
+}
+
+func GenerateSharesProofs(
```
gupadhyaya, line 456: [nit] godoc

renaynay (dismissed):

gupadhyaya (dismissed):

nodebuilder/share/share.go
```diff
@@ -197,16 +192,30 @@ func (m module) SharesAvailable(ctx context.Context, height uint64) error {
...
+		return nil, err
+	}
+
+	// build ODS coordinates from the provided indexes
+	from, err := shwap.SampleCoordsFrom1DIndex(start, len(hdr.DAH.RowRoots)/2)
+	if err != nil {
+		return nil, err
+	}
+	// end is exclusive in the API but an inclusive for coords calculation
+	to, err := shwap.SampleCoordsFrom1DIndex(end-1, len(hdr.DAH.RowRoots)/2)
```
walldiss: Why `end-1`? Could you please document external API if it uses half open interval or full interval. 

Lets make internal (Getter) and external (json-rpc) API consistent and use same interval schema.

nodebuilder/share/types/get_range_result.go
```diff
@@ -0,0 +1,126 @@
+package share_types //nolint:revive
```
walldiss: `GetRangeResult` was in the same pkg as API. In order for this be PR to be revert of breaking changes we should not extract it into another subpkg. The `GetRangeResult` used to be in `nodebuilder/share/share.go`

vgonkivs: let me at least extract it into a separate file within `nodebuilder/share` folder

nodebuilder/share/share.go
vgonkivs: `end-1` end is exlclusive in json-rpc api.

> Lets make internal (Getter) and external (json-rpc) API consistent and use same interval schema.

We can't make it inclusive bc the API will be breaking in this case. From the other hand, making it exclusive on the p2p level requires additional handling(since we are sending coords on the P2P level) of edge cases, which I would like to avoid. I'm planning to move to inclusive indexes when we'll break the top level API.

The edge case happens when the *EXCLUSIVE* index is the first share of the row, so during calculating coords, we must take this into account and reduce row by 1.

I think it's ok to convert exclusive to inclusive for now.

vgonkivs: reworked to apply indexes instead of coordinates up to Accessor level.
In future, I think we can rework RangeNamespaceDataFromShares to apply indexes too(but not in scope of this PR)

store/file/ods.go
```diff
@@ -324,12 +324,25 @@ func (o *ODS) Reader() (io.Reader, error) {
...
+	if err != nil {
+		return shwap.RangeNamespaceData{}, err
+	}
+	// to is an exclusive index.
+	toCoords, err := shwap.SampleCoordsFrom1DIndex(to-1, odsSize)
+	if err != nil {
+		return shwap.RangeNamespaceData{}, err
+	}
+
+	shares := make([][]libshare.Share, fromCoords.Row-toCoords.Row+1)
```
gupadhyaya: should this be `toCoords.Row-fromCoords.Row+1` instead?

vgonkivs: true. Thanks

nodebuilder/share/get_range_result.go
```diff
@@ -0,0 +1,126 @@
...
+// GetRangeResult wraps the return value of the GetRange endpoint
+// because Json-RPC doesn't support more than two return values.
+type GetRangeResult struct {
+	// Shares contains the data shares retrieved from the specified range.
+	Shares []libshare.Share
+	// Proof proves that shares were included in the data root.
+	Proof *types.ShareProof
+}
+
+func NewGetRangeResult(
```
gupadhyaya: [nit]godoc plz

share/shwap/p2p/bitswap/range_namespace_data_block.go
```diff
@@ -143,15 +130,25 @@ func (rndb *RangeNamespaceDataBlock) UnmarshalFn(root *share.AxisRoots) Unmarsha
 		if err != nil {
 			return fmt.Errorf("unmarshaling RangeNamespaceData for %+v: %w", rndb.ID, err)
 		}
-		if len(rangeNsData.Proof) == 0 {
-			return fmt.Errorf("unmarshaling RangeNamespaceData for %+v: proof is empty", rndb.ID)
+
+		from, err := shwap.SampleCoordsFrom1DIndex(rndid.From, len(root.RowRoots)/2)
```
gupadhyaya: [nit] pull `len(root.RowRoots) / 2` out to `odsSize` variable.

gupadhyaya: everywhere you can do this. helps with reading code.

share/shwap/range_namespace_data_id.go
```diff
@@ -70,33 +55,34 @@ func NewRangeNamespaceDataID(
 
 // Verify validates the RangeNamespaceDataID fields and verifies that number of the requested shares
 // does not exceed the number of shares inside the ODS.
-func (rngid RangeNamespaceDataID) Verify(edsSize int) error {
+func (rngid RangeNamespaceDataID) Verify(size int) error {
```
gupadhyaya: either change the name to odsSize or add a godoc. it will be helpful for consumers of Verify

share/shwap/range_namespace_data.go
```diff
@@ -390,9 +326,132 @@ func RangeCoordsFromIdx(edsIndex, length, edsSize int) (SampleCoords, SampleCoor
...
+	// Compute leaf hashes for the namespace merkle tree
+	hashes, err := nmt.ComputePrefixedLeafHashes(nth, ns.Bytes(), libshare.ToBytes(shares))
+	if err != nil {
+		return nil, fmt.Errorf("failed to compute leaf hashes: %w", err)
+	}
+	// Compute the row root using the namespace merkle tree proof
+	return proof.ComputeRootWithBasicValidation(nth, ns.Bytes(), hashes, nsCompleteness)
+}
+
+func buildTreeRootFromLeaves(shares [][]byte, index uint) ([]byte, error) {
```
gupadhyaya: i prefer to have the parameter named more accurately especially when it is being used to compute somethings. e.g., `extendedShares` here is later tree size is determined as `len(shares)/2)`.
same applies in other places. i would leave it `shares` if no such assumptions (of being extended / original) is made in the function.

share/shwap/range_namespace_data.go
```diff
@@ -1,295 +1,249 @@
...
+	if err != nil {
+		return err
+	}
+
+	computedRoots := make([][]byte, len(shares))
+	firstRoot, err := computeRoot(shares[0], ns, rngdata.FirstIncompleteRowProof, nsCompleteness)
+	if err != nil {
+		return err
+	}
+	lastRoot, err := computeRoot(shares[len(shares)-1], ns, rngdata.LastIncompleteRowProof, nsCompleteness)
```
gupadhyaya: should you only computeRoot if `len(shares) > 1` here? otherwise, you may return err? i would just compute and assign together:
```
if len(shares) > 1 {
                // computeRoot
		computedRoots[len(shares)-1] = lastRoot
}
```

nodebuilder/share/get_range_result.go
vgonkivs: added a godoc + made it internal.

share/shwap/range_namespace_data.go
vgonkivs: could you please elaborate? What error do you mean?

Computing last root happens only if range spans across multiple ranges, otherwise(len(shares) == 1) - `computedRoots[0] == computedRoots[len(shares)-1)]`, so we'll overwrite the correct proof.

gupadhyaya: you are right. but why even overwrite if you can skip it?

vgonkivs: Skip assigning the proof? in what case?

vgonkivs:
```
	// this is an expected behavior.
	if proof == nil {
		return nil, nil
	}
```

There is an if condition in the beginning of the function that returns nil.

gupadhyaya: what i mean is,

```
computedRoots := make([][]byte, len(shares))

firstIncompleteRoot, err := computeRoot(
    shares[0],
    ns,
    rngdata.FirstIncompleteRowProof,
    nsCompleteness,
)
if err != nil {
    return err
}
computedRoots[0] = firstIncompleteRoot

if len(shares) > 1 {
    lastIdx := len(shares) - 1
    lastIncompleteRoot, err := computeRoot(
        shares[lastIdx],
        ns,
        rngdata.LastIncompleteRowProof,
        nsCompleteness,
    )
    if err != nil {
        return err
    }
    computedRoots[lastIdx] = lastIncompleteRoot
}
```

vgonkivs: I did the same by adding an if condition inside compute proof
```
if proof == nil {
		return nil, nil
	}
```

gupadhyaya: works.

renaynay (approved):
approving but why was eds/proof.go deleted?

walldiss (approved):

PR #4330, opened 2025-06-14
feat: integrate pruning

### LINKED ISSUE #4240: Pruning Integration

### LINKED ISSUE #4241: refactor(pruner): prune from Tail
Instead of current `pruner.Service` [starting from genesis](https://github.com/celestiaorg/celestia-node/blob/7af0c47a25ab892d0f3da867c912496ebf6db137/pruner/checkpoint.go#L29), the pruner should start from header.Store.Tail. 

The checkpointing with core prunable header searching algorithm should stay as it is Pruner still needs to maintain its own state of what has been pruned.

### LINKED ISSUE #4242: refactor(das): prune from Tail
Instead of current `das.Daser` [starting from genesis](https://github.com/celestiaorg/celestia-node/blob/2447fbe6eea788ed8e9f982c01dea069d164622a/das/options.go#L55), the daser should start from header.Store.Tail instead.

The checkpointing should stay as it is, but DASer should not expose `SampleFrom` configuration anymore and always use Tail instead. 

One tricky thing to look out for is the order of component starts. If Daser depends on Tail, the Tail should exist on the store. This may not happen as smoothly as we would want on the first initialization. Potentially, retries could help here for `Tail` or we ensure that DASer starts only after header.Syncer. The solution there is up to the implementer.

Wondertan:
This PR integrates [header pruning](https://github.com/celestiaorg/go-header/issues/260) into the node. The work of [dozens of PRs in go-header](https://github.com/celestiaorg/go-header/pulls?q=is%3Apr+sort%3Aupdated-desc++label%3Apruning+) finally lands with this PR.

## New
While this feature is quite broad, affecting most layers of the node's stack, for the user it is _mainly_ manifested with a few additions in the config (and significant reduction of disk usage consumption and synchronization/sampling speed)

```toml
  [Header.Syncer]
    # PruningWindow defines the duration within which headers are retained before being pruned.
    # Default is 337 hours.
    PruningWindow = "337h0m0s"
    # SyncFromHash is the hash of the header from which the syncer should start syncing.
    # Zero value to disable. Value updates up and down the chain are gracefully handled by Syncer.
    #
    # By default, Syncer maintains PruningWindow number of headers. SyncFromHash overrides this default,
    # allowing any user to specify a custom starting point.
    #
    # SyncFromHash has higher priority than SyncFromHeight.
    SyncFromHash = ""
    # SyncFromHeight is the height of the header from which the syncer should start syncing.
    # Zero value to disable. Value updates up and down the chain are gracefully handled by Syncer.
    #
    # By default, Syncer maintains PruningWindow number of headers. SyncFromHeight overrides this default,
    # allowing any user to specify a custom starting point.
    #
    # SyncFromHeight has lower priority than SyncFromHash.
    SyncFromHeight = 0
```

## Gotchas
Header pruning aims to be minimally breaking for the existing node runners. However, there are a few gotchas:

#### LNs no longer support historical queries
By default,  node maintains a sliding window of headers, bounded by Tail and Head headers. Requests with height below the Tail are rejected (This is, however, temporary, and lazy header fetching will be available with Backward Sync)

To retain the ability to request older queries with LNs, use new configuration fields to set an absolute header that the node will sync from, either the SyncFromHeight or the SyncFromHash field.

#### LNs only
Pruning for Bridge and Full nodes is disabled. Currently, the p2p protocol for headers cannot distinguish between pruned and non-pruned serving node types. Therefore, to keep *stable* synchronization of historical headers, we disable pruned servers.

#### Prolonged Starts 
* If your LN stores a significant portion of the history of headers, you may need to increase the `StartTimeout` value in the config, as LN prunes on start. If the starting times out, e.g., with the default 20s, the node will attempt to persist the progress, but storage state corruptions may still occur.  The complete mainnet header history pruning on moderate hardware takes under 100s, so update your config accordingly.
* If you move the tail down the chain (e.g., SyncByHeight 150 -> 100), you may need to increase the `StartTimeout` value in the config, as LN syncs the difference during start. (This is, however, temporary, and asynchronous syncing that doesn't block start will be available with Backward Sync).
  * __WARNING__: This feature wasn't tested extensively and may break in unknown ways. Use at your own risk!

#### Config Breaks
* `Daser.SampleFrom` field is no longer supported. Use `Header.Syncer.SyncFromHeight` instead.
* `Header.TrustedHash` field is no longer supported. Use `Header.Syncer.SyncFromHash` instead.

#### Replace
Replace for go-datastore was added, which may break users depending directly on the node's go.mod.
Do the following in your project's root to fix:
`go mod edit -replace github.com/ipfs/go-datastore=github.com/celestiaorg/go-datastore@48a63ae && go mod tidy`. It's gonna be removed in future releases.

## Manual Tests
- [x] Sync LN with a default config 
  - assert the tail is moving on every new head
  - assert it triggers block data pruning
  - assert Daser starts from Tail and syncs everything up to Head
  - assert storage usage stays relatively the same
- [x] Sync full header history by setting SyncFromHeight to 1
  - assert Daser and Pruner iterate over all headers, but skip most of them, respecting DPW
  - assert tail is absolute and constant
- [x] Update LN config to move the tail up (by changing window, height, or hash)
  - assert tail moves on Start and with reasonable speed
    - deleting headers 
    - triggering block data pruning
  - assert Daser doesn't break after the move
- [ ] Updated LN config to move the tail down (by changing window, height, or hash)
  - assert tail moves on Start by syncing missing headers

Some of the listed tests may become new or merged into an existing e2e test.

## TODO
 - [x] Disable header pruning for PFN. We decided to disable it as there is no way to distinguish between pruned and archival on the header exchange protocol. 
   - [x] Task in go-header to implement it
 - [x] Update windows to match recent CIPs
   - [x] TrustingPeriod (was missed in #4424)
   - Others will be done by https://github.com/celestiaorg/celestia-node/pull/4427
 - [x] Forbid DPW > HPW
 - [x] Recommend users to increase the start timeout
 - [ ] Docs in docs repo
 - [ ] E2e test covering some of the manual scenarios
 - [ ] Integration/functional test asserting interactions between the header store and pruner
 - [ ] Under the default settings where DPW == HPW, the data Pruner routine is always front-run by Syncer and only eats resources. Maybe think of a way to suspend it
 
 ## Relations
 Closes #4240
 Closes #4241
 Closes #4242


pruner/params.go
```diff
@@ -22,7 +22,7 @@ func (p *Params) Validate() error {
 
 func DefaultParams() Params {
 	return Params{
-		pruneCycle: time.Minute * 5,
+		pruneCycle: time.Second * 30,
```
Wondertan, line 25: I don't see why we would do it every 5 mins. Longer time prefers a tiny bit of CPU over disk usage, and I  don't think that's a tradeoff we wanna go

pruner/service.go
```diff
@@ -115,17 +126,8 @@ func (s *Service) run() {
...
-	if err != nil {
-		log.Errorw("failed to get last pruned header", "height", s.checkpoint.LastPrunedHeight,
-			"err", err)
-		log.Warn("exiting pruner service!")
-
-		s.cancel()
-	}
-
 	for {
-		lastPrunedHeader = s.prune(s.ctx, lastPrunedHeader)
```
Wondertan, line 128: Keeping the last height from the last round doesn't work anymore. Instead, we should dynamically ask `lastPruned` as `onHeaderDelete` may kick in between rounds.

pruner/service.go
```diff
@@ -139,32 +141,50 @@ func (s *Service) run() {
...
+
 	// prioritize retrying previously-failed headers
 	s.retryFailed(s.ctx)
 
 	now := time.Now()
+	successful, failed := 0, 0
 	log.Debug("pruning round start")
 	defer func() {
-		log.Debugw("pruning round finished", "took", time.Since(now))
+		log.Infow("pruning round finished", "done", successful, "failed", failed, "took", time.Since(now))
```
Wondertan, line 170: Doesn't hurt to announce that pruning has been executed and the results of it

nodebuilder/header/config.go
```diff
@@ -21,9 +19,6 @@ var MetricsEnabled = false
 
 // Config contains configuration parameters for header retrieval and management.
 type Config struct {
-	// TrustedHash is the Block/Header hash that Nodes use as starting point for header synchronization.
-	// Only affects the node once on initial sync.
-	TrustedHash string
```
Wondertan, line 26: We may reintroduce trustedHash at some point, as originally intended - a point of trust rather than a particular point to sync from.

Currently, we only support the automatic trusted peers based one, but as per mus, someone may want to configure the trust point themselves.

nodebuilder/header/module.go
```diff
@@ -41,6 +42,13 @@ func ConstructModule[H libhead.Header[H]](tp node.Type, cfg *Config) fx.Option {
 				ctx context.Context,
 				breaker *modfraud.ServiceBreaker[*sync.Syncer[H], H],
 			) error {
+				// TODO(@Wondertan): This fix flakes in e2e tests
+				//  This is coming from the store asynchronity.
+				//  Previously, we would request genesis during initialization
+				//  but now we request it during Syncer start and given to the Store.
+				//  However, the Store doesn't makes it immediately available causing flakes
+				//  The proper fix will be in a follow up release after pruning.
```
Wondertan, line 50: The proper fix is to use Syncer.Tail rather then Store.Tail in node's codebase. SyncerTail guarantees to return the most correct tail at any point of time. However, SyncerTail is not currently exposed but will be

pruner/find.go
```diff
@@ -18,7 +18,14 @@ func (s *Service) findPruneableHeaders(
 	ctx context.Context,
 	lastPruned *header.ExtendedHeader,
 ) ([]*header.ExtendedHeader, error) {
-	pruneCutoff := time.Now().UTC().Add(-s.window)
```
Wondertan, line 21: The goal here is to align the logic with the header pruning one as described in the respective CIP.

nodebuilder/pruner/flags.go
```diff
@@ -40,5 +40,6 @@ func ParseFlags(cmd *cobra.Command, cfg *Config, tp node.Type) {
 	}
 
 	// If node operator does not explicitly pass `--archival` flag and pruner service is disabled, disallow
-	log.Fatal("ARCHIVAL MODE ENABLED BY ACCIDENT!!! Please explicitly pass `--archival` to enable archival mode!")
+	log.Fatal(`ARCHIVAL MODE ENABLED BY ACCIDENT!!!
+		Please enable pruning service in config or explicitly pass "--archival" to enable archival mode`)
```
Wondertan: This "ask" makes the EnablePruning configuration option effectively noop, so we should remove it if we __really__ want to make archival only via flags.

However, I think the idea of enabling it __through flags only__ wouldn't hold any criticism

nodebuilder/header/config.go
```diff
@@ -82,34 +82,44 @@ func (cfg *Config) trustedPeers(bpeers p2p.Bootstrappers) (infos []peer.AddrInfo
...
-	if err != nil {
-		return fmt.Errorf("module/header: misconfiguration of syncer: %w", err)
-	}
-
 	err = cfg.Server.Validate()
 	if err != nil {
 		return fmt.Errorf("module/header: misconfiguration of p2p exchange server: %w", err)
 	}
 
+	if cfg.Syncer.TrustingPeriod != trustingPeriod {
```
Wondertan, line 98: I think we shouldn't expose this parameter, the same way we don't expose the sampling window

nodebuilder/header/config.go
```diff
@@ -82,34 +83,44 @@ func (cfg *Config) trustedPeers(bpeers p2p.Bootstrappers) (infos []peer.AddrInfo
...
+			)
+		}
+	case node.Light:
+		err = cfg.Syncer.Validate()
+		if err != nil {
+			return fmt.Errorf("module/header: misconfiguration of syncer: %w", err)
+		}
+
+		if cfg.Syncer.PruningWindow < availability.StorageWindow {
+			// TODO(@Wondertan): Technically, a LN may break this restriction by setting SyncFromHeight/Hash to a header
```
renaynay, line 117: but it's only for a brief period right? until pruning window worth of blocks produced since SyncFromHeight/Hash

nodebuilder/header/flags.go
```diff
@@ -1,34 +1,25 @@
...
 	"github.com/multiformats/go-multiaddr"
 	"github.com/spf13/cobra"
 	flag "github.com/spf13/pflag"
 )
 
-var (
-	headersTrustedHashFlag  = "headers.trusted-hash"
-	headersTrustedPeersFlag = "headers.trusted-peers"
-)
+var headersTrustedPeersFlag = "headers.trusted-peers"
```
renaynay, line 11: so we will only expose header pruning params via config ?

nodebuilder/header/module.go
renaynay: could we WaitStoreCompleteInit? I'd rather not deal with syncer in node borderline period.

nodebuilder/pruner/constructors.go
```diff
@@ -4,17 +4,23 @@ import (
...
 
 	"github.com/celestiaorg/celestia-node/header"
+	modfraud "github.com/celestiaorg/celestia-node/nodebuilder/fraud"
 	"github.com/celestiaorg/celestia-node/nodebuilder/p2p"
 	modshare "github.com/celestiaorg/celestia-node/nodebuilder/share"
 	"github.com/celestiaorg/celestia-node/pruner"
 )
 
+// ensures Pruner always starts after Syncer
+type syncerAnchor = modfraud.ServiceBreaker[*headsync.Syncer[*header.ExtendedHeader], *header.ExtendedHeader]
```
renaynay, line 17: oh fx <3

nodebuilder/pruner/flags.go
renaynay: Yeah you're right.

renaynay: we could potentially do this before release is fully public

pruner/params.go
renaynay: i'm fine w this

das/daser_test.go
```diff
@@ -54,13 +52,6 @@ func TestDASerLifecycle(t *testing.T) {
 		require.EqualValues(t, 30, checkpoint.SampleFrom-1)
 	}()
 
-	// wait for mock to indicate that catchup is done
```
renaynay, line 57: why is it now possible to remove these?

pruner/checkpoint.go
```diff
@@ -91,5 +107,23 @@ func (s *Service) updateCheckpoint(
...
+	if err != nil {
+		return nil, err
+	}
+
+	lastPruned := s.checkpoint.LastPrunedHeight
+	if tail.Height() < lastPruned {
+		return s.hstore.GetByHeight(ctx, lastPruned)
+	}
+
+	s.checkpoint.LastPrunedHeight = tail.Height()
```
renaynay, line 120: when would this case happen / how often could this happen?

pruner/checkpoint.go
```diff
@@ -91,5 +107,23 @@ func (s *Service) updateCheckpoint(
...
+	for height := range s.checkpoint.FailedHeaders {
+		if height < tail.Height() {
+			delete(s.checkpoint.FailedHeaders, height)
+			log.Warnw("unfailing unavailable header; some block data may not be pruned", "height", height)
```
renaynay, line 124: this is confusing message to user, maybe we can elaborate like 
```suggestion
			log.Warnw("attempt to prune block data from unavailable header; some block data may not be pruned", "height", height)
```

pruner/find.go
```diff
@@ -18,7 +18,14 @@ func (s *Service) findPruneableHeaders(
...
+	// TODO(@Wondertan): This should be SyncerHead to ensure the most accurate Head
+	//  but its ok for now and will be fixed in a future release.
+	head, err := s.hstore.Head(ctx)
```
renaynay, line 23: There's no danger here really other than estimating a pruneCutoff that's actually **too old** rather than **too new** so I think we're fine here.

renaynay: Eventually w bsync, syncer head will be == store head, no?

pruner/service.go
```diff
@@ -88,12 +92,19 @@ func (s *Service) Stop(ctx context.Context) error {
...
+	if err != nil {
+		return fmt.Errorf("pruner: saving checkpoint on stop: %w", err)
+	}
+
+	return nil
 }
 
+// TODO: Exposed but accesses checkpoint without synchronization
 func (s *Service) LastPruned(ctx context.Context) (uint64, error) {
 	err := s.loadCheckpoint(ctx)
```
renaynay, line 116: should we lock checkpoint mux in load checkpoint and where accessing last pruned header? feels like needs read protection no?

pruner/service.go
```diff
@@ -139,69 +141,88 @@ func (s *Service) run() {
...
+	}()
+
+	ctx, done := s.withWriteBatch(ctx)
+	defer func() {
+		if err := done(); err != nil {
+			log.Errorw("committing batch", "err", err)
+		}
 	}()
 
+	ctx, doneTxn := s.withReadTransaction(ctx)
```
renaynay, line 180: can you explain a bit about why we need this here too?

nodebuilder/header/module.go
Wondertan: @renaynay, I don't see why would we hack around this if we can rely on logic that guarantees correct Tail or Head at any point of time.

I also don't see why would you want to avoid "messing" with Syncer "period". Such strong opinions needs some backing

nodebuilder/header/config.go
Wondertan: Right and its unclear if we want to allow it. As currently it doesnt happen, i left it as is but with a todo. Doesnt hurt to cover it too

nodebuilder/header/flags.go
Wondertan: We can add flags too

das/daser_test.go
Wondertan: Because it uses mocks provided by headertest pkg

pruner/find.go
Wondertan: Agree its not a biggy, thus i kept it as is.

Yes, with bsync store will have the most recent head, unlike now where it needs to sync up there.
Although, header sub may fall behind and syncer head can cover that

pruner/service.go
Wondertan: It is required if LastPruned called after Pruner.Start

pruner/service.go
Wondertan: Its an optional optimization

pruner/checkpoint.go
Wondertan: Syncer starts first and deletes up to the new tail. The pruner starts later, meaning it will always get the latest tail, and the tail will always be higher or equal than LastPrunedHeight

nodebuilder/header/module.go
renaynay: I don’t see this as a hack. What feels wrong is that an internal detail (syncer) of go-header is so exposed that we, as users, have to decide whether to use the syncer’s tail or the store’s tail.

Right now, we’re the only “power user” of go-header, but our usage depends on a lot of implicit knowledge about its internals to get the most optimized behavior. Ideally, that wouldn’t be necessary.

From our perspective, go-header should just provide a contiguous chain without us needing to understand or handle its internal states. The syncer’s job is to maintain that chain / ensure its completeness— but consumers shouldn’t need to know how it does so in order to use the library correctly.

I know this is an ideal that we can move towards and doesn't need to happen in this PR, but this is my backing for my "strong opinion".

nodebuilder/header/flags.go
renaynay: Maybe we should do this for consistency in this PR.

pruner/service.go
renaynay: If it's optional, does it need to be in this PR or can it come later?

pruner/service.go
renaynay: Is there a reason that this method isn't protected and relies on the callers to protect it though? Like what's the point of leaving it unprotected?

nodebuilder/header/module.go
Wondertan: Hmm, we should not assume Store as the top-level interface for users and the canonical place for retrieving Tail or Head. The store is exactly "an internal detail" that is so exposed that we have to decide which one to use. Syncer is a higher-level concern; although it couples the syncing logic and head determination together, it is still a canonical place currently that knows what the correct Tail and Head are.

What should help resolve this confusion is introducing a Service on the go-header level. At its base, it will be a Service as it is right now in the header module - a service that gives the full CRUD for headers in a verified manner. This Service will take some of the Syncer's responsibilities, like Tail and Head determination, leaving Syncer to do precisely one thing - syncing. Yet, it will be a single top-level API for users like us and RPC to interact with. 

The Store should gradually become a low-level concern that users do not interact with at all. The Service would then be able to get data from the store itself or lazy-load it (or spit an error that data is not available).

Wondertan: One thing I can propose here is to start going the Service path from the get-go, rather than temporarily tying to Syncer. Yes, we need to correct head or tail logic here, but it doesn't mean we can't start by extracting that logic into a better place than Syncer.

pruner/service.go
Wondertan: It could come later, obviously, but I don't see this being a big deal to copy and paste here. This has proven to be a substantial optimization that has minor maintenance overhead

pruner/service.go
Wondertan: I am not the person who implemented this method to answer this. The synchronization was supposed to be there before I added header pruning

nodebuilder/header/module.go
renaynay: I agree with top-level Service approach - this makes go-header cleaner to use and makes us (maintainers of go-header) the only ones who need to understand internal components in order to make top-level Service as clean and accurate as possible.

> One thing I can propose here is to start going the Service path from the get-go, rather than temporarily tying to Syncer. 

Does this mean you'll do this as an immediate FLUP in go-header or what's the proposed plan of action here concretely?

renaynay: > this Service will take some of the Syncer's responsibilities, like Tail and Head determination, leaving Syncer to do precisely one thing - syncing. 

Why does this have to happen? I still think it is within syncer scope to do head and tail determination. It's the coordinator.

pruner/params.go
vgonkivs: Have you run any benchmarks to test this? I mean, why 30seconds, not 20 or 40

nodebuilder/header/module.go
Wondertan: > Does this mean you'll do this as an immediate FLUP in go-header or what's the proposed plan of action here concretely?

Not necessary to block this PR immediately, but a tech debt FLUP to be picked up later on. The immediate thing here would be opening up an issue

Wondertan: > Why does this have to happen? I still think it is within syncer scope to do head and tail determination. It's the coordinator.

The simplest thing to do is to expose Service.Tail|Head that will call Syncer's respective methods, but I think there is merit in making Syncer focused solely on syncing.

pruner/service.go
```diff
@@ -214,3 +235,83 @@ func (s *Service) retryFailed(ctx context.Context) {
...
+// There is a possible race between Syncer and Pruner, however, it is gracefully resolved.
+// * If Syncer prunes a header first - pruneOnHeaderDelete gets called and data is deleted.
+//   - Pruner then ignores the header based on updated checkpoint state
+//
+// * If Pruner prunes a header first - Syncer only removes the header
+//   - pruneOnHeaderDelete ignores the header based on updated checkpoint state
+func (s *Service) pruneOnHeaderDelete(ctx context.Context, height uint64) error {
+	s.checkpointMu.Lock()
+	if err := s.loadCheckpoint(ctx); err != nil {
+		return err
```
vgonkivs: deadlock?

pruner/params.go
Wondertan: The pruning cycle is on the order of hundreds of ms in avg case. There is no point in making this run every 5 mins. Ideally, we would run in place for every header without any delay even.

pruner/service.go
Wondertan: while extremly rare, it is still possible

pruner/service.go
Wondertan: I made sure synchronization is everywhere now

pruner/service.go
Wondertan: fixed

vgonkivs (approved):

PR #4341, opened 2025-06-19
feat(nodebuilder/node)!: More useful node info (type is string and report network ID)

renaynay:
New reported result from `node.Info`

```
{
  "result": {
    "type": "Bridge",
    "api_version": "v0.23.0-arabica-12-g855eec41d-dev",
    "network_id": "arabica-11"
  }
}
```

Also adds a `NetworkID` API endpoint.


nodebuilder/node/module.go
```diff
@@ -5,11 +5,11 @@ import (
 	"go.uber.org/fx"
 )
 
-func ConstructModule(tp Type) fx.Option {
+func ConstructModule(tp Type, network string) fx.Option {
 	return fx.Module(
 		"node",
 		fx.Provide(func(signer jwt.Signer, verifier jwt.Verifier) Module {
-			return newModule(tp, signer, verifier)
+			return newModule(tp, signer, verifier, network)
```
Wondertan: There is a p2p.Network type already supplied to FX. You can just add it as a param in the provide func instead.

Wondertan: Also would be good to use this more specific type everywhere instead of a string

renaynay: @Wondertan cyclical deps so I can't, that's why I did this.

Wondertan: Add pls a comment somewhere so during nodebuilder refactoring this doesnt get forgotten

Wondertan: Actually, if network is defined on p2p, i think we should just add this endpoint to p2p module

renaynay: You don't think it's useful to return as part of node info? 

I guess I could move it bc of the flag anyway. Node runners are used to passing `--p2p.network`

renaynay: check new commit @Wondertan

Wondertan: Why not to use respective types now? Therr should not be any cycled anymore

vgonkivs: changed to Typed network

rach-id (approved):

PR #4363, opened 2025-06-25
deps(go.mod): Bump app

renaynay:
bump app


vgonkivs (dismissed):

walldiss (dismissed):

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM --platform=$BUILDPLATFORM docker.io/golang:1.24-alpine3.20 AS builder
+FROM --platform=$BUILDPLATFORM docker.io/golang:1.24.4-alpine3.22 AS builder
```
walldiss: why this changed to point to minor versions? Golang 1.24 pulls latest

renaynay: Needed to otherwise docker build would fail.

walldiss (approved):

PR #4380, opened 2025-07-02
feat(nodebuilder/core | state): Multi-core endpoint tx submission

### LINKED ISSUE #3837: feat(api): Support Multiple core.Ip Addresses with Retry for Data Posting
### Implementation ideas

Currently, celestia-node requires a single core.Ip for posting data. If that IP is unavailable, the node can't post data, leading to failure.

Suggestion:
Allow users to specify multiple core.Ip addresses.
Implement a retry mechanism: if the first IP fails, retry with the next available IP in the list until successful or all attempts fail.
Benefits:
Improves reliability by switching to another IP if the primary is unreachable.
Reduces downtime caused by network or node issues.

renaynay:
Allow additional core endpoints to be configured for transaction submission via Core config.

I deliberately excluded flags from CLI for this configuration as the flags could add additional confusion / parsing logic and IMO configuring via config is sufficient for UX. Additionally, I chose not to limit the amount of additional endpoints we allow as the tx client will cap additional conns at a max of 3.

_Does not break config -- purely additive._

Closes #3837

```toml
[Core]
IP = "192.155.1.1"
Port = "9090"
TLSEnabled = false
XTokenPath = ""

[[Core.AdditionalCoreEndpoints]]
IP = "192.168.1.10"
Port = "8080"
TLSEnabled = true
XTokenPath = "/path/to/token"

[[Core.AdditionalCoreEndpoints]]
IP = "192.168.1.11"
Port = "8081"
TLSEnabled = false
XTokenPath = "/another/token/path"

```


nodebuilder/core/config.go
```diff
@@ -15,6 +15,13 @@ var MetricsEnabled bool
 
 // Config combines all configuration fields for managing the relationship with a Core node.
 type Config struct {
+	EndpointConfig
+	// AdditionalCoreEndpoints is a list of additional Celestia-Core endpoints to be used for
+	// transaction submission. Must be provided as `host:port` pairs.
+	AdditionalCoreEndpoints []EndpointConfig
+}
```
walldiss, line 22: What is the point of AdditionalCoreEndpoints in case of Bridge node core sync? Does it use additional endpoints?

state/core_access.go
```diff
@@ -122,10 +122,10 @@ func NewCoreAccessor(
 func (ca *CoreAccessor) Start(ctx context.Context) error {
 	ca.ctx, ca.cancel = context.WithCancel(context.Background())
 	// create the staking query client
-	ca.stakingCli = stakingtypes.NewQueryClient(ca.coreConn)
-	ca.feeGrantCli = feegrant.NewQueryClient(ca.coreConn)
+	ca.stakingCli = stakingtypes.NewQueryClient(ca.coreConns[0])
+	ca.feeGrantCli = feegrant.NewQueryClient(ca.coreConns[0])
 	// create ABCI query client
-	ca.abciQueryCli = tmservice.NewServiceClient(ca.coreConn)
+	ca.abciQueryCli = tmservice.NewServiceClient(ca.coreConns[0])
```
vgonkivs, line 128: We should be able to re-init clients in case `ca.coreConns[0]` is down.

nodebuilder/core/module.go
```diff
@@ -24,8 +24,11 @@ func ConstructModule(tp node.Type, cfg *Config, options ...fx.Option) fx.Option
 
 	baseComponents := fx.Options(
 		fx.Supply(*cfg),
+		fx.Supply(cfg.EndpointConfig),
 		fx.Error(cfgErr),
 		fx.Provide(grpcClient),
+		fx.Provide(additionalCoreEndpointGrpcClients),
+		fx.Provide(),
```
vgonkivs: empty `fx.Provide()`?

state/core_access.go
```diff
@@ -108,7 +108,7 @@ func NewCoreAccessor(
 		defaultSignerAddress: addr,
 		getter:               getter,
 		prt:                  prt,
-		coreConn:             conn,
+		coreConns:            []*grpc.ClientConn{conn},
```
vgonkivs, line 111: Are we ok with keeping multiple connections(some of them won't be used at all)? I mean tx client already stores them.

nodebuilder/core/config.go
renaynay: No, BNs don't need it and don't accept additional endpoints. Historically, we've had all core-related setup in core config so i elected to keep it there for this feature as well.

nodebuilder/core/module.go
renaynay: nice catch

state/core_access.go
renaynay: I'm not sure what you mean by the question. If we have multiple connections, we will pass them to tx client which will utilise them accordingly.

state/core_access.go
renaynay: Isn't there logic in grpc conn to keep alive / retry, etc? What additional logic do we need to add here? I would not want to bloat.

vgonkivs: If the connection is down, then we can't use some of our functionality:  our staking API is not working.
btw, I've checked` feeGrantCli  feegrant.QueryClient` and it looks like it is not used in our code at all.

state/core_access.go
vgonkivs: I mean that both CoreAccessor and TxClient store them. What is the point of keeping them inside core accessor?

vgonkivs: Maybe we can add `Conns() `to the TxClient to fetch them if needed

nodebuilder/core/config.go
```diff
@@ -15,6 +15,13 @@ var MetricsEnabled bool
 
 // Config combines all configuration fields for managing the relationship with a Core node.
 type Config struct {
+	EndpointConfig
```
Wondertan, line 18: Original endpoint could become a slice or is the idea to not break the config?

nodebuilder/core/constructors.go
```diff
@@ -35,9 +35,11 @@ const (
 	xtokenFileName = "xtoken.json"
 )
 
+type AdditionalCoreEndpoints []*grpc.ClientConn
```
Wondertan: A pedantic note: the word endpoint is used to refer to an actual established connection and to an IP/port pair. You can't equate all these meanings.

nodebuilder/core/module.go
```diff
@@ -24,8 +24,10 @@ func ConstructModule(tp node.Type, cfg *Config, options ...fx.Option) fx.Option
 
 	baseComponents := fx.Options(
 		fx.Supply(*cfg),
+		fx.Supply(cfg.EndpointConfig),
```
Wondertan, line 27: where is being used to be supplied?

nodebuilder/core/constructors.go
Wondertan: Also, make the type private if its not supposed to be used outside. It looks like it doesn't if you didn't make the option rely on it

nodebuilder/core/config.go
renaynay: idea was not to break config but to make it additive change - og endpoint cfg would be untouched.

nodebuilder/core/module.go
renaynay: it's being used in `grpcClient` construction (for main connection to core)

nodebuilder/core/constructors.go
renaynay: Check [38bbb9c37](https://github.com/celestiaorg/celestia-node/pull/4380/commits/38bbb9c374d3a0742eb7372be67fc6940e1f2e03)

i can't privatise i need in state mod constructors

walldiss (approved):

state/core_access.go
renaynay: @vgonkivs the conns aren't managed by the txclient, the core accessor manages them. The txclient only uses them.

PR #4381, opened 2025-07-02
chore(node/tests): fix and cleanup swamp tests

Wondertan:
While working on #4330, I had to fix swamp tests s.t. I know pruning doesn't break anything. While diving again into swamp, I realized we have a bunch of garbage and unnecessary code that I removed + unified node setup and connections for all the tests. Also, now the integration tests should be **GREEN**.

Also runs these tests in parallel. On my machine, time goes from 122s -> 86s. There is one more test time optimization in #4330 that reduces the number to ~60s

### DIFF
--- a/nodebuilder/tests/prune_test.go
+++ b/nodebuilder/tests/prune_test.go
@@ -41,6 +41,8 @@ import (
 // spin up 3 pruning FNs, connect
 // spin up 1 LN that syncs historic blobs
 func TestArchivalBlobSync(t *testing.T) {
+	t.Parallel()
+
 	const (
 		blocks = 10
 		btime  = time.Millisecond * 300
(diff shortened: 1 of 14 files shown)


renaynay (dismissed):
Thank you 👑

nodebuilder/tests/prune_test.go
```diff
@@ -169,6 +166,18 @@ func TestArchivalBlobSync(t *testing.T) {
...
+	//  With following pruning intergration PR it just works
+	//  and I don't have anymore time to figure this one out
+	//  Its something with subscriptions and headers are not
+	//  delivered by LN
+	go func() {
+		for {
+			ln.HeaderServ.NetworkHead(ctx)
+			time.Sleep(time.Second)
+		}
+	}()
```
walldiss, line 179: This is very concerning. Why do you think calling NetworkHead repeatedly would be required for test to pass?

Wondertan: Because it sees that the current head is outdated and requests a newer one

Wondertan: Again, i remove it in the further pr and it works withiut this

PR #4417, opened 2025-07-21
fix: skip file permission checks on Windows

avorylli:
Skip Unix-style file permission validation on Windows as the filesystem permission model differs as this might fix TestFSKeystore failures on Windows platforms while maintaining security on Unix systems


Wondertan:
@avorylli, we don't fully support windows. Ideally, we would check permissions there based on the Windows format if you can do it

PR #4421, opened 2025-07-23
test: use tastora docker types instead of interface types

gupadhyaya:
instead of using https://github.com/celestiaorg/tastora/blob/main/framework/types/da_node.go#L72, directly use the docker types: https://github.com/celestiaorg/tastora/blob/main/framework/docker/da_node.go#L61 as docker is the only anticipated platform that will be supported and the generic types will be removed as per: https://github.com/celestiaorg/tastora/issues/53


chatton:
> Wouldn't this drop support for in memory nodes that were discussed? Having an ability to (a) run tests locally without running docker and (b) being able to debug with multiple nodes in the same process where you can step through is invaluable.

It should still be possible to support this use case with by attaching a debugger to the running node processes in the containers, but yes it would eliminate the "swappable backend" idea.

PR #4423, opened 2025-07-23
test: header module tastora tests

gupadhyaya:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -170,6 +170,11 @@ test-share:
 	@echo "Running share module tests..."
 	@cd nodebuilder/tests/tastora && go test -v -run="TestShareTestSuite" -timeout=15m
 
+## test-header: Run header module tests via Tastora framework.
+test-header:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -170,6 +170,11 @@ test-share:
 	@echo "Running share module tests..."
 	@cd nodebuilder/tests/tastora && go test -v -run="TestShareTestSuite" -timeout=15m
 
+## test-header: Run header module tests via Tastora framework.
+test-header:
+	@echo "Running header module tests..."
+	@cd nodebuilder/tests/tastora && go test -v -run="TestHeaderTestSuite" -timeout=15m
+
```
chatton: I think this is still missing the build tag, I don't think this will run the test

gupadhyaya: yes, it needs `-tags=integration`

gupadhyaya:
will replace this with sync tests that covers better integration scenarios.

PR #4437, opened 2025-07-26
fix(pruner): restrict log output to 'start' command in ParseFlags

### LINKED ISSUE #4063: fix(module/pruner): WARN log for disabled full pruning service should only pop up on `start` if `--experimental-pruning` is not passed.

crStiv:
Updated ParseFlags to log pruning and archival mode messages only when the start command is used. This prevents unnecessary log noise during non-runtime commands like status or config.

closes #4063


renaynay:
I believe this was resolved by https://github.com/celestiaorg/celestia-node/pull/4472 but thank you @crStiv

Issue #4125, opened 2025-02-17
bug: make light-arabica-up missing grpc flag


jcstein:
### Celestia Node version

v0.21.5

### OS

mac

### Install tools

docs

### Steps to reproduce it

`make light-arabica-up`

### Expected result

node starts after being funded

### Actual result

```
Error: nodebuilder/core: grpc port is not set
make: *** [light-arabica-up] Error 1
```

### Relevant log output

```Shell

```

PR #4440, opened 2025-07-28
fix: ensure default core.port is always set via flag

DeVikingMark:
Fixes an issue #4125 where the core.port flag value (default or user-defined) was not consistently applied when --core.port was omitted from the CLI args. Now, the port is always derived from the flag value, improving configuration consistency.

Includes additional test coverage to validate default behavior when only --core.ip is provided.


gupadhyaya (dismissed):

walldiss (approved):

DeVikingMark:
@walldiss @gupadhyaya @Wondertan 

Gentlemen, what are your thoughts about this approach

Wondertan (approved):

PR #4460, opened 2025-08-03
test: tastora blob module tests updated to cover integration scenarios

gupadhyaya:
## Test Infrastructure
- **Network**: 1 Validator + 1 Bridge Node + 2 Light Nodes (Tastora Docker-based)
- **Auto-funding**: All nodes funded with 10B utia automatically

### 1. `TestCrossNodeBlobAvailability` 
Validates that blobs submitted on one node become available for retrieval on all other nodes in the network.
Tests cross-node data propagation and ensures blob data integrity across bridge and light nodes.

### 2. `TestMultiNamespaceBlobIsolation`
Validates that blobs in different namespaces are properly isolated and can be queried independently.
Tests namespace separation by ensuring queries for one namespace don't return data from another namespace.

### 3. `TestConcurrentBlobOperations`
Validates that multiple concurrent blob submissions work correctly without conflicts using goroutines.
Tests 3 simultaneous blob submissions with unique namespaces and verifies all can be retrieved from all nodes.

### 4. `TestBlobProofConsistency` 
Validates that blob inclusion proofs are consistent and identical across different node types.
Tests that bridge and light nodes generate the same proofs and can cross-verify each other's proofs.

### 5. `TestLightNodeBlobAccess` 
Validates that light nodes can successfully access blob data without storing full block data.
Tests light node capabilities: blob retrieval, namespace querying, and proof generation for large blobs (~1.15KB).

### DIFF
--- a/nodebuilder/tests/tastora/blob_test.go
+++ b/nodebuilder/tests/tastora/blob_test.go
@@ -5,6 +5,8 @@ package tastora
 import (
 	"bytes"
 	"context"
+	"fmt"
+	"sync"
 	"testing"
 	"time"
 
@@ -12,12 +14,13 @@ import (
 
 	"github.com/celestiaorg/go-square/v2/share"
 
+	"github.com/celestiaorg/celestia-node/api/rpc/client"
 	nodeblob "github.com/celestiaorg/celestia-node/blob"
 	"github.com/celestiaorg/celestia-node/state"
 )
 
-// BlobTestSuite provides comprehensive testing of the Blob module APIs.
-// Tests blob submission, retrieval, and validation functionality.
+// BlobTestSuite provides integration testing of blob operations across multiple nodes.
+// Focuses on cross-node data availability, network coordination, and real-world blob scenarios.
 type BlobTestSuite struct {
 	suite.Suite
 	framework *Framework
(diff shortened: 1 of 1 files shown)


gupadhyaya:
recommend not looking at the diff as most of the previous tests were unit test kind and the updated tests are integration scenarios. looking at the blob_test file itself to see if the tests make sense is better.

nodebuilder/tests/tastora/blob_test.go
```diff
@@ -31,319 +34,432 @@ func TestBlobTestSuite(t *testing.T) {
...
+	// Wait for block inclusion and propagation
+	_, err = bridgeClient.Header.WaitForHeight(ctx, height)
 	s.Require().NoError(err)
-	s.Require().NotZero(height)
 
-	// Wait for inclusion
-	_, err = client.Header.WaitForHeight(ctx, height)
-	s.Require().NoError(err)
+	// Allow time for data propagation across the network
+	time.Sleep(10 * time.Second)
```
renaynay, line 78: This sleep is not necessary - rather wait for both LNs to reach same header height which you are doing below.

renaynay: Same for rest of tests.

nodebuilder/tests/tastora/blob_test.go
```diff
@@ -31,319 +34,432 @@ func TestBlobTestSuite(t *testing.T) {
...
+		})
+	}
 }
 
-// TestBlobSubmit_MultipleBlobs tests blob submission API with multiple blobs
-func (s *BlobTestSuite) TestBlobSubmit_MultipleBlobs() {
-	ctx, cancel := context.WithTimeout(context.Background(), 120*time.Second)
+// TestMultiNamespaceBlobIsolation validates that blobs in different namespaces
+// are properly isolated and can be queried independently across nodes
+func (s *BlobTestSuite) TestMultiNamespaceBlobIsolation() {
```
renaynay, line 109: This test really doesn't belong in e2e test suite IMO - should be more at the blob module level or even share level.

nodebuilder/tests/tastora/blob_test.go
```diff
@@ -31,319 +34,432 @@ func TestBlobTestSuite(t *testing.T) {
...
+		})
 	}
 }
 
-// TestBlobGet_ExistingBlob tests blob retrieval API for existing blobs
-func (s *BlobTestSuite) TestBlobGet_ExistingBlob() {
-	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+// TestConcurrentBlobOperations validates that multiple concurrent blob operations
+// work correctly across different nodes without conflicts
+func (s *BlobTestSuite) TestConcurrentBlobOperations() {
```
renaynay, line 199: The value in this test lies in the logic that tests whether BN can submit blobs concurrently. 

let's limit this test to that + also include a test case for LN to be able to do the same (not like they're any different from a state operations perspective, but better to test both just to be able to catch regressions)

nodebuilder/tests/tastora/blob_test.go
```diff
@@ -31,319 +34,432 @@ func TestBlobTestSuite(t *testing.T) {
...
+			namespace, err := share.NewV0Namespace(namespaceBytes)
+			if err != nil {
+				s.T().Errorf("Submission %d: failed to create namespace: %v", submissionID, err)
+				return
+			}
+
+			data := []byte(fmt.Sprintf("Concurrent submission %d data", submissionID))
+			nodeBlobs := s.createBlobsForSubmission(ctx, bridgeClient, namespace, data)
+
+			// Submit with different gas to avoid nonce conflicts
```
renaynay, line 233: The `txClient` should be able to handle nonce conflicts internally at some point. I'd prefer if we had a test case either now (which we can expect to flake) or later on once we have confirmed better nonce mgmt for concurrent txs that will submit txs with the defaults and ensure they are committed if valid.

nodebuilder/tests/tastora/blob_test.go
```diff
@@ -31,319 +34,432 @@ func TestBlobTestSuite(t *testing.T) {
...
 	s.Require().NoError(err)
 
-	// Submit multiple blobs to same namespace
-	data1 := []byte("GetAll test data 1")
-	data2 := []byte("GetAll test data 2")
+	blobData := []byte("Proof consistency test data")
+	nodeBlobs := s.createBlobsForSubmission(ctx, bridgeClient, namespace, blobData)
 
-	libBlob1, err := share.NewV1Blob(namespace, data1, nodeAddr.Bytes())
+	txConfig := state.NewTxConfig(state.WithGas(250_000), state.WithGasPrice(5000))
```
renaynay, line 333: let's not do config manipulation if we do not need it in the context of the test - keep test as simple as possible.

nodebuilder/tests/tastora/blob_test.go
```diff
@@ -31,319 +34,432 @@ func TestBlobTestSuite(t *testing.T) {
...
-	height, err := client.Blob.Submit(ctx, nodeBlobs, txConfig)
+	// Get proofs from both nodes
+	bridgeProof, err := bridgeClient.Blob.GetProof(ctx, height, namespace, nodeBlobs[0].Commitment)
+	s.Require().NoError(err, "bridge should generate proof")
+
+	lightProof, err := lightClient.Blob.GetProof(ctx, height, namespace, nodeBlobs[0].Commitment)
+	s.Require().NoError(err, "light should generate proof")
+
+	// Verify proof consistency
+	s.Assert().Equal(len(*bridgeProof), len(*lightProof), "proof lengths should match")
```
renaynay, line 356: why can't you do a deep equal on it?

nodebuilder/tests/tastora/blob_test.go
```diff
@@ -31,319 +34,432 @@ func TestBlobTestSuite(t *testing.T) {
...
 
-	// Wait for inclusion
-	_, err = client.Header.WaitForHeight(ctx, height)
+	lightIncluded, err := lightClient.Blob.Included(ctx, height, namespace, lightProof, nodeBlobs[0].Commitment)
 	s.Require().NoError(err)
+	s.Assert().True(lightIncluded, "light should verify inclusion")
 
-	// Test GetAll API
-	retrievedBlobs, err := client.Blob.GetAll(ctx, height, []share.Namespace{namespace})
+	// Cross-verification: bridge proof should work on light node
```
renaynay, line 374: then you wouldn't need this.

nodebuilder/tests/tastora/blob_test.go
```diff
@@ -31,319 +34,432 @@ func TestBlobTestSuite(t *testing.T) {
...
-	for _, blob := range retrievedBlobs {
-		s.Assert().True(blob.Namespace().Equals(namespace))
-	}
+	s.T().Logf("✅ Blob proof consistency verified across nodes")
 }
 
-// TestBlobGetProof_ValidBlob tests GetProof API functionality
-func (s *BlobTestSuite) TestBlobGetProof_ValidBlob() {
-	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+// TestLightNodeBlobAccess validates that light nodes can successfully access
```
renaynay, line 382: I'm okay with this test I guess - but there's coverage for all of these cases via other tests.

I'd move it higher up then I guess as a plain sanity check. Also pls rename and change the description as it's confusing.

renaynay (changes requested):

PR #4464, opened 2025-08-04
feat(cmd/share): add range cmd

vgonkivs:
Add GetShareRange cmd.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,13 +2,18 @@ SHELL=/usr/bin/env bash
 PROJECTNAME=$(shell basename "$(PWD)")
 DIR_FULLPATH=$(shell pwd)
 versioningPath := github.com/celestiaorg/celestia-node/nodebuilder/node
+tastoraPath := github.com/celestiaorg/celestia-node/nodebuilder/tests/tastora
 OS := $(shell uname -s)
 LDFLAGS = -ldflags="-X $(versioningPath).buildTime=$(shell date -u +%Y-%m-%dT%H:%M:%SZ) \
                     -X $(versioningPath).lastCommit=$(shell git rev-parse HEAD) \
                     -X $(versioningPath).semanticVersion=$(shell \
                       git name-rev --name-only --tags --no-undefined HEAD 2>/dev/null || \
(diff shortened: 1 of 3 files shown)


nodebuilder/share/cmd/share.go
```diff
@@ -156,3 +158,36 @@ var getEDS = &cobra.Command{
...
+		if err != nil {
+			return err
+		}
+		end, err := strconv.ParseUint(args[2], 10, 64)
+		if err != nil {
+			return err
+		}
+
+		if start >= end {
+			return errors.New("start must be greater than end")
```
gupadhyaya: error message is confusing. should be `errors.New("start index must be less than end index")`

gupadhyaya (dismissed):

share/shwap/getter.go
```diff
@@ -72,10 +72,10 @@ type Getter interface {
 	) (NamespaceData, error)
 
 	// GetRangeNamespaceData retrieves a range of shares within a specific namespace in the
-	// Extended Data Square (EDS). The range is defined by from and to coordinates, which
-	// specify the start and end points of the desired range. The coordinates are inclusive,
-	// meaning shares at both from and to positions are included in the result. The shares are
-	// returned in a row-by-row order if the range spans multiple rows.
+	// Extended Data Square (EDS). The range is defined by from and to indexes, which
+	// specify the start and end points of the desired range, where `from` is inclusive
+	// and `to` is exclusive indexes. The shares are returned in a row-by-row order
```
renaynay, line 77: Okay just want to double check here that this is the case and the previous docs were inaccurate?

vgonkivs: Yes, was checking the docs and found that previous version was inaccurate. At some point GetRangeNamespaceData was requiring shwap.Coords that both were inclusive. Then I've changed coords to ods indexes.

nodebuilder/share/cmd/share.go
vgonkivs: updated

renaynay (approved):
utack

PR #4465, opened 2025-08-05
test: cleanup tastora framework

gupadhyaya:
Improvements
* bridge nodes and light nodes in framework with default bridge node
* automatic funding
* simple GetBridgeNode and GetLightNode apis for using in tests
* made helper functions private

### DIFF
--- a/nodebuilder/tests/tastora/framework.go
+++ b/nodebuilder/tests/tastora/framework.go
@@ -66,7 +66,7 @@ func NewFramework(t *testing.T, options ...Option) *Framework {
 	f := &Framework{
 		t:                    t,
 		logger:               zaptest.NewLogger(t),
-		defaultFundingAmount: 3_000_000_000, // 3 billion utia - default funding amount
+		defaultFundingAmount: 10_000_000_000, // 10 billion utia - enough for multiple blob transactions
 	}
 
 	// Apply configuration options
--- a/nodebuilder/tests/tastora/framework.go
+++ b/nodebuilder/tests/tastora/framework.go
@@ -30,9 +30,9 @@ import (
 
 const (
 	celestiaAppImage   = "ghcr.io/celestiaorg/celestia-app"
-	defaultCelestiaTag = "v4.0.0-rc6"
+	defaultCelestiaTag = "v5.0.1"
 	nodeImage          = "ghcr.io/celestiaorg/celestia-node"
-	defaultNodeTag     = "v0.23.2-mocha"
+	defaultNodeTag     = "5b96c43" // update this to tag after the release
 	testChainID        = "test"
 )
 
--- a/nodebuilder/tests/tastora/framework.go
+++ b/nodebuilder/tests/tastora/framework.go
@@ -45,15 +45,18 @@ type Framework struct {
 	client  *client.Client
 	network string
 
-	provider   tastoratypes.Provider
-	daNetwork  tastoratypes.DataAvailabilityNetwork
-	bridgeNode *tastoradockertypes.DANode
-	lightNodes []*tastoradockertypes.DANode
-	celestia   tastoratypes.Chain
+	provider  tastoratypes.Provider
+	daNetwork tastoratypes.DataAvailabilityNetwork
 
-	// Default wallet for automatic node funding
-	defaultWallet tastoratypes.Wallet
-	// Default funding amount for nodes (3 billion utia)
+	// Main DA network infrastructure
+	celestia tastoratypes.Chain
+
+	// Node topology (simplified)
+	bridgeNodes []*tastoradockertypes.DANode // Bridge nodes (bridgeNodes[0] created by default)
+	lightNodes  []*tastoradockertypes.DANode // Light nodes
+
+	// Private funding infrastructure (not exposed to tests)
+	fundingWallet        tastoratypes.Wallet
 	defaultFundingAmount int64
 }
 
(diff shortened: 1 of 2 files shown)


nodebuilder/tests/tastora/framework.go
```diff
@@ -30,9 +30,9 @@ import (
 
 const (
 	celestiaAppImage   = "ghcr.io/celestiaorg/celestia-app"
-	defaultCelestiaTag = "v4.0.0-rc6"
+	defaultCelestiaTag = "v5.0.1"
 	nodeImage          = "ghcr.io/celestiaorg/celestia-node"
-	defaultNodeTag     = "v0.23.2-mocha"
+	defaultNodeTag     = "5b96c43" // update this to tag after the release
```
renaynay: There should be a way to pull this in from ldflags btw

renaynay: we absolutely need to do this otherwise we will miss this and will be PITA

nodebuilder/tests/tastora/framework.go
```diff
@@ -80,90 +83,61 @@ func NewFramework(t *testing.T, options ...Option) *Framework {
...
-// The bridge node is automatically funded with the default amount for transaction operations.
-func (f *Framework) GetOrCreateBridgeNode(ctx context.Context) *tastoradockertypes.DANode {
-	if f.bridgeNode == nil {
-		f.bridgeNode = f.startBridgeNode(ctx, f.celestia)
-
-		// Automatically fund the bridge node
-		defaultWallet := f.getOrCreateDefaultWallet(ctx)
-		f.FundNodeAccount(ctx, defaultWallet, f.bridgeNode, f.defaultFundingAmount)
-		f.t.Logf("Bridge node automatically funded with %d utia", f.defaultFundingAmount)
+// ============ CLEAN API ============
```
renaynay: please clean up PR of these comments

nodebuilder/tests/tastora/framework.go
```diff
@@ -80,90 +83,61 @@ func NewFramework(t *testing.T, options ...Option) *Framework {
...
 
-		// Wait a moment to ensure funds are available
-		time.Sleep(2 * time.Second)
+// GetBridgeNode returns a bridge node by index, creating it if it doesn't exist.
+func (f *Framework) GetBridgeNode(ctx context.Context, index int) *tastoradockertypes.DANode {
```
renaynay: Let's just do simple 

`GetBridgeNodes` --> returns f.bridgeNodes

`GetLightNodes` --> returns f.lightNodes

`NewBridgeNode` --> creates, starts + appends BN
`NewLightNode` --> creates, starts + appends LN

gupadhyaya: does this mean, the tests now will do two calls? also, tests will have to be aware of how many exists and have to add logic to make sure the number of nodes (bridge and light) that they require is created, etc. I was trying to hide them behind the Getters.

gupadhyaya: is it not easier for tests to just call `GetBridgeNode(index)` and `GetLightNode(index)` to get as many bridge/light nodes as they want, without having to deal with checking/ensuring nodes are created, etc? i feel this is better ux for tests than separating two calls. 
also, if every test does NewBridge/NewLight node, without Get, they cannot reuse the already created nodes and containers. they will have to add this unnecessary logic.

gupadhyaya: for now, changed the apis as suggested.

nodebuilder/tests/tastora/framework.go
```diff
@@ -104,40 +104,42 @@ func (f *Framework) SetupNetwork(ctx context.Context) error {
...
+	f.bridgeNodes = append(f.bridgeNodes, bridgeNode)
+	return bridgeNode
+}
 
-		f.lightNodes = append(f.lightNodes, lightNode)
+// NewLightNode creates, starts and appends a new light node.
+func (f *Framework) NewLightNode(ctx context.Context) *tastoradockertypes.DANode {
+	// Get the next available light node from the DA network
+	allLightNodes := f.daNetwork.GetLightNodes()
+	if len(f.lightNodes) >= len(allLightNodes) {
```
renaynay: I still don't really understand how this works - you have to preallocate the amount of LNs you're allowed to run in tastora before running a test? That doesn't seem right to me.

> they cannot reuse the already created nodes and containers

I understand the usefulness of being able to reuse a container for another test without having to spin down/up new containers on every test if not needed, but why can't a test spin up 1 additional container for an LN without it already being preallocated?

renaynay (commented):
The way I imagine the tests to go is: 

`api_test.go`
-> suite setup: spin up and start 1 BN and 1 LN
-> `TestHeaderAPICalls` 
-> `TestShareAPICalls`
-> etc etc etc
-> spin down suite

`e2e_sanity.go`
-> suite setup: 
-> all e2e sanity tests
-> spin down suite

etc. etc.

nodebuilder/tests/tastora/framework.go
gupadhyaya: yes, this is just a config setting that you pass when u initialize the DA network and say something like let's say 10 light nodes max (based on resource limitations). the tests can still dynamically create and destroy the nodes as they require. its not a limitation for our tests, we can simply set this number high.

gupadhyaya: @chatton any reason behind having the static topology via `DataAvailabilityNetworkConfig`?

renaynay (approved):
yeah this is better for now thank you. 

Utack on ldflags but looks like it should work.

nodebuilder/tests/tastora/framework.go
chatton: The initial design made some assumptions that you would deploy a DA network and then perform some operations, and then make some assertions.

Dynamic creation/removal of the network nodes themselves wasn't taken into consideration. 

I think since this is something that is definitely desirable we can update the tastora API to be more flexible and allow for dynamic addition / removal of nodes to a given DA network. 

We should be able to provide a more minimal config to the `DataAvailabilityNetwork` type itself, and then dynamically create or remove nodes as required.

The way the celestia-app chain is built also change a bit from the original design (done through a builder now). It would be nice to align the APIs a bit so construction of resources is more consistent. I think this can be done later I think.

gupadhyaya: yes, we are dynamically adjusting in the tests while still specifying an upper bound in the config.

chatton: Do you want to create an issue in tastora to support this usecase? I can do it in a bit otherwise!

gupadhyaya: i will let you. thanks.

chatton: https://github.com/celestiaorg/tastora/issues/92

PR #4472, opened 2025-08-07
fix(cmd/celestia): parse only store determination flags for auxiliary commands

renaynay:
_Please review this PR carefuly_

---------------------------------

I've wanted to do this for a while - we do not need to parse the entire set of flags available for `Init` and `Start` for all other auxiliary commands as they're irrelevant.

This PR decouples this parsing logic into 
`ParseStoreDeterminationFlags` and `ParseAllFlags`

Now, all subcommands will run their own necessary parsing logic. 

This PR fixes a bug found by @vgonkivs where a LN would fatal out on `config-update` subcommand due to unrelated: 
```
celestia light config-update --p2p.network=mocha
2025-08-07T16:23:27.396+0300	FATAL	module/pruner	pruner/flags.go:43	ARCHIVAL MODE ENABLED BY ACCIDENT!!! Please explicitly pass `--archival` to enable archival mode!
```

This PR also only parses pruning flags for the `Bridge` node -- LNs all prune by default and don't have an archival option anyway.

Supersedes #4131

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -12,8 +12,14 @@ func RemoveConfigCmd(fsets ...*flag.FlagSet) *cobra.Command {
 		Use:   "config-remove",
 		Short: "Deletes the node's config",
 		Args:  cobra.NoArgs,
-		RunE: func(cmd *cobra.Command, _ []string) error {
+		RunE: func(cmd *cobra.Command, args []string) error {
+			err := ParseStoreDeterminationFlags(cmd, NodeType(cmd.Context()), args)
+			if err != nil {
+				return err
+			}
+
 			ctx := cmd.Context()
+
 			return nodebuilder.RemoveConfig(StorePath(ctx))
 		},
 	}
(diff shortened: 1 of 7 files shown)


cmd/config.go
```diff
@@ -12,8 +12,14 @@ func RemoveConfigCmd(fsets ...*flag.FlagSet) *cobra.Command {
 		Use:   "config-remove",
 		Short: "Deletes the node's config",
 		Args:  cobra.NoArgs,
-		RunE: func(cmd *cobra.Command, _ []string) error {
+		RunE: func(cmd *cobra.Command, args []string) error {
+			err := ParseStoreDeterminationFlags(cmd, NodeType(cmd.Context()), args)
+			if err != nil {
+				return err
+			}
+
```
vgonkivs, line 20: What is the reason of doing this for remove cmd?

renaynay: I changed all auxiliary subcommands to only parse for store determination.

vgonkivs (approved):

PR #4476, opened 2025-08-10
fix(node): Harden keystore filesystem permissions (0700 dir; validate key modes)

anim001k:
### Summary
Tighten filesystem permissions for the keystore: create the directory with 0700 and validate key file modes via FileInfo.Mode().

### Rationale
Prevent group/other access to private keys; previous 0755 could expose sensitive material.

### Changes
- Create keystore directory with 0700 instead of 0755.
- Use DirEntry.Info().Mode() for permission checks.


renaynay (approved):
Actually my apologies, I thought this was for the entire data store, not just keystore. For keystore, makes sense.

PR #4482, opened 2025-08-13
fix(nodebuilder/pruner)!: Remove pruner config from nodes config

### LINKED ISSUE #4477: nodebuilder/pruner/config.go: Remove pruner config entirely and require user to explicitly pass `--archival` to enable archival mode
Ref https://github.com/celestiaorg/celestia-node/pull/4330/files#r2250190110

renaynay:
The pruner config field has been removed from the config as it was rendered ineffectual by https://github.com/celestiaorg/celestia-node/pull/4303. 

**Now, users MUST PASS `--archival` when running their node to disable pruning (and ensure the node runs in archival mode)**

Resolves #4477

### DIFF
--- a/nodebuilder/pruner/module.go
+++ b/nodebuilder/pruner/module.go
@@ -19,7 +19,7 @@ import (
 
 var log = logging.Logger("module/pruner")
 
-func ConstructModule(tp node.Type, cfg *Config) fx.Option {
+func ConstructModule(tp node.Type) fx.Option {
 	prunerService := fx.Options(
 		fx.Provide(fx.Annotate(
 			newPrunerService,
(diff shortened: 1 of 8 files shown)


renaynay:
@Wondertan https://github.com/celestiaorg/celestia-node/pull/4330/files#r2250190110
your own comment

walldiss:
> Why would we want to do this flag only?

Enabling pruning setting is not really configurable property. Pruning can be enabled, but not disabled after that. Using flag only is stronger UX enforcement than cfg field.

nodebuilder/pruner/module.go
```diff
@@ -36,9 +36,13 @@ func ConstructModule(tp node.Type, cfg *Config) fx.Option {
 	)
 
 	baseComponents := fx.Options(
-		fx.Supply(cfg),
+		// supply the default config, which can only be overridden by
+		// passing the `--archival` flag
+		fx.Supply(&Config{
+			EnableService: true,
+		}),
```
walldiss: nit: Default config would look better as `func DefaultConfig() config` and would not need comments

walldiss (approved):
Would be nice to have specific flags usage tested in tastora tests in future

Wondertan:
My comment says config doesnt work, but it doesnt  suggest removing it.

The argument that it cant be switched back off is fine.

vgonkivs (approved):

walldiss (approved):

PR #4503, opened 2025-08-26
chore: bump celestia-app to v6

evan-forbes:
bumps app to v6


evan-forbes:
notably, I didn't do anything to increase the square size in tests but ya'll might want to try that out. max square size ods is now 512

walldiss (commented):
Any breaking changes inherited from app? Can we expect v6 node to run fine against v5 app?

evan-forbes:
> Any breaking changes inherited from app? Can we expect v6 node to run fine against v5 app?

max ods square size increase to 512 is the main thing. The rest should be the same. Bug fixes in for the state machine and TxClient.

walldiss (approved):

nodebuilder/testing.go
```diff
@@ -75,10 +75,17 @@ func TestNodeWithConfig(t *testing.T, tp node.Type, cfg *Config, opts ...fx.Opti
...
+		if ip == "" {
+			tn := core.StartTestNode(t)
+			var err error
+			ip, port, err = net.SplitHostPort(tn.GRPCClient.Target())
+			require.NoError(t, err)
+		}
+
 		con, err := grpc.NewClient(
-			core.StartTestNode(t).GRPCClient.Target(),
+			net.JoinHostPort(ip, port),
```
evan-forbes, line 88: not sure if this was on purpose or not, but it looks like this was spinning up two different consensus networks?

nodebuilder/node_test.go
```diff
@@ -33,7 +33,8 @@ func TestLifecycle(t *testing.T) {
 		t.Run(strconv.Itoa(i), func(t *testing.T) {
 			// we're also creating a test node because the gRPC connection
 			// is started automatically when starting the node.
-			host, port, err := net.SplitHostPort(core.StartTestNode(t).GRPCClient.Target())
+			consNode := core.StartTestNode(t)
+			host, port, err := net.SplitHostPort(consNode.GRPCClient.Target())
 			require.NoError(t, err)
```
evan-forbes, line 38: combined with the below TestnodeWithConfig call, we were spinning up three different consensus networks afaiu

maybe I'm missing something simple

nodebuilder/testing.go
evan-forbes: I don't like this "solution" as it feels really hacky. I didn't want to break the api to fix however. imo the best approach here is to be explicit and force the passing of a core node for bridge nodes or not handle that edge case in that function.

also if there are more than one bridge node and the core ip is not specified, then it will still spin up multiple consensus networks

evan-forbes:
converted to a draft as this is no longer using the rc tag, but a similar tag to what the arabica release will be

hopefully getting that out ASAP, the this PR should be good. The only macos test failing now is

```
--- FAIL: TestFraudProofValidation (20.14s)
    --- FAIL: TestFraudProofValidation/ods_size:128 (8.09s)
        retriever_test.go:127: 
            	Error Trace:	/Users/runner/work/celestia-node/celestia-node/share/eds/retriever_test.go:127
            	            				/Users/runner/work/celestia-node/celestia-node/share/eds/retriever_test.go:108
            	            	before batch commit: context deadline exceeded
            	Test:       	TestFraudProofValidation/ods_size:128
FAIL
```

testnode restarts now with different ports now if it hits the port already in use error. I still don't fully understand why that test fails consistently at the same port without restarting.

renaynay:
@evan-forbes is the Fraud proof validation test failure a flake or a consistent failure?

state/core_access_test.go
```diff
@@ -226,53 +225,20 @@ func buildAccessor(t *testing.T, opts ...Option) (*CoreAccessor, []string) {
...
+		WithFundedAccounts(accounts...).
+		WithTimeoutCommit(time.Millisecond * 1)
 
 	cctx, _, grpcAddr := testnode.NewNetwork(t, config)
 
 	conn, err := grpc.NewClient(grpcAddr, grpc.WithTransportCredentials(insecure.NewCredentials()))
 	require.NoError(t, err)
-	ca, err := NewCoreAccessor(cctx.Keyring, accounts[0].Name, nil, conn, chainID, opts...)
+	ca, err := NewCoreAccessor(cctx.Keyring, accounts[0], nil, conn, chainID, opts...)
 	require.NoError(t, err)
```
evan-forbes, line 242: I don't think this is strictly required, but certain things such as setting the timeout commit in the config don't actually do anything since the timeout commit is determined by the application. The rest are already the defaults in testnode.

therefore  I cleaned some of the things up here, and hopefully making them less brittle as testnode should rely on the defaults set in app as much as possible.

evan-forbes:
> is the Fraud proof validation test failure a flake or a consistent failure?

the mac fraud proof failure is technically a flake as it doesn't _always_ occur, only like 85% of the time :innocent:

vgonkivs (approved):

Issue #4489, opened 2025-08-18
test: yet another tastora reorganization plan


gupadhyaya:
# Test Reorganization Proposal for Tastora

### **Proposed Solution**

This reorganization introduces a **four-tier testing strategy** that provides:

1. **API Tests** (< 30s): Pure contract validation for rapid feedback
2. **E2E Sanity** (1-2 min): Basic functionality flows for core validation  
3. **E2E P2P Sanity** (2-4 min): Network topology and multi-node coordination
4. **E2E P2P Complex** (3-5 min): Advanced scenarios and edge cases

### **Scope**

This reorganization covers all major Celestia node modules:
- **Share Module**: DAS operations, namespace isolation, sampling
- **Header Module**: Sync operations, network head management  
- **Blob Module**: Submission, retrieval, proof generation
- **State Module**: Account management, delegation operations
- **P2P Module**: Peer management, bandwidth monitoring
- **Node Module**: Administrative operations, authentication
- **Blobstream Module**: Data root tuple root operations
- **Core Module**: Bridge node core chain integration
- **Pruner Module**: Pruning coordination and archival operations

### 1. **API Tests** (Pure API Compliance)
**Setup**: 1 Bridge Node + 1 Light Node  
**Focus**: API contract validation, RPC compliance, parameter validation  
**Execution Time**: Fast (< 30 seconds per test)

- **TestRPCEndpointCompliance**: Raw curl tests against both BN/LN RPC ports
- **TestShareAPIContract**: Validate Share module API responses match expected schema
  - SharesAvailable, GetShare, GetSamples, GetEDS, GetRow, GetNamespaceData, GetRange
- **TestHeaderAPIContract**: Validate Header module API responses
  - LocalHead, GetByHash, GetByHeight, WaitForHeight, GetRangeByHeight, SyncState, SyncWait, NetworkHead, Subscribe
- **TestBlobAPIContract**: Validate Blob module API responses
  - Submit, Get, GetAll, GetProof, Included, GetCommitmentProof, Subscribe
- **TestStateAPIContract**: Validate State module API responses
  - AccountAddress, Balance, BalanceForAddress, Transfer, SubmitPayForBlob, Delegate, Undelegate, QueryDelegation
- **TestP2PAPIContract**: Validate P2P module API responses
  - Info, Peers, PeerInfo, Connect, ClosePeer, Connectedness, NATStatus, BandwidthStats, ResourceState
[...]

PR #4508, opened 2025-08-28
test: e2e sanity test with blob, das, header sync and p2p

gupadhyaya:
This PR covers part of category 2 from #4489: e2e sanity tests and includes:
* blow e2e flow using bridge node
* das e2e flow using bridge and light node
* header sync
* p2p connectivity

also includes a bunch of changes to framework for robustness

### DIFF
--- a/nodebuilder/core/config.go
+++ b/nodebuilder/core/config.go
@@ -24,11 +24,11 @@ type Config struct {
 type EndpointConfig struct {
 	IP   string
 	Port string
-	// TLSEnabled specifies whether the connection is secure or not.
-	// PLEASE NOTE: it should be set to true in order to handle XTokenPath.
+	// TLSEnabled specifies whether the connection is secure.
+	// Must be set to true if XTokenPath is provided.
 	TLSEnabled bool
-	// XTokenPath specifies the path to the directory with JSON file containing the X-Token for gRPC authentication.
-	// The JSON file should have a key-value pair where the key is "x-token" and the value is the authentication token.
+	// XTokenPath specifies the path to the directory that contains a JSON file with the X-Token for gRPC authentication.
+	// The JSON file must contain a key "x-token" with the authentication token.
 	// If left empty, the client will not include the X-Token in its requests.
 	XTokenPath string
 }
@@ -77,6 +77,15 @@ func (cfg *EndpointConfig) validate() error {
 	_, err = strconv.Atoi(cfg.Port)
 	if err != nil {
 		return fmt.Errorf("nodebuilder/core: invalid grpc port: %s", err.Error())
+	}
+
+	if cfg.XTokenPath != "" {
+		if !cfg.TLSEnabled {
+			return fmt.Errorf("nodebuilder/core: TLSEnabled must be true when XTokenPath is set")
+		}
+		if !utils.Exists(cfg.XTokenPath) {
+			return fmt.Errorf("nodebuilder/core: XTokenPath does not exist: %s", cfg.XTokenPath)
+		}
 	}
 
 	return nil


nodebuilder/tests/tastora/framework.go
```diff
@@ -215,8 +215,20 @@ func (f *Framework) queryBalance(ctx context.Context, addr string) sdk.Coin {
...
 
-// FundWallet sends funds from one wallet to another address.
-func (f *Framework) FundWallet(ctx context.Context, fromWallet tastoratypes.Wallet, toAddr sdk.AccAddress, amount int64) {
+// newBridgeNode creates and starts a bridge node.
+func (f *Framework) newBridgeNode(ctx context.Context) *tastoradockertypes.DANode {
+	bridgeNode := f.startBridgeNode(ctx, f.celestia)
+	f.fundNodeAccount(ctx, bridgeNode, f.defaultFundingAmount)
+	f.t.Logf("Bridge node created and funded with %d utia", f.defaultFundingAmount)
+
+	// Wait a moment to ensure funds are available
```
renaynay: isn't there a more deterministic way to do this by querying the chain for its current height?

nodebuilder/tests/tastora/framework.go
```diff
@@ -387,19 +411,45 @@ func (f *Framework) startLightNode(ctx context.Context, bridgeNode *tastoradocke
...
+					"P2P_NETWORK":     testChainID,
+				},
+			),
+		)
+		if err == nil {
+			return lightNode
+		}
+		lastErr = err
+		// Retry only for transient docker/container races
+		if strings.Contains(err.Error(), "No such container") {
```
renaynay: nit: does docker go sdk expose errors?

nodebuilder/tests/tastora/framework.go
```diff
@@ -459,13 +509,129 @@ func getNodeImage() string {
...
+		f.fundingWallet = f.CreateTestWallet(ctx, 50_000_000_000) // 50 billion utia
+		f.t.Logf("Created funding wallet for automatic node funding")
+	}
+	return f.fundingWallet
+}
+
+// Robust waiting and P2P connectivity helpers
+
+// ConnectLightToBridge ensures the light node is directly connected to the bridge peer.
+func (f *Framework) ConnectLightToBridge(ctx context.Context, bridgeClient, lightClient *rpcclient.Client, label string) {
```
renaynay: why can't this be more generic -- just "ConnectNodes" and it can take any number of nodes and connect them?

nodebuilder/tests/tastora/framework.go
```diff
@@ -459,13 +509,129 @@ func getNodeImage() string {
...
+		time.Sleep(1 * time.Second)
+	}
+}
+
+// WaitSharesAvailable waits until SharesAvailable succeeds for the given client and height.
+func (f *Framework) WaitSharesAvailable(ctx context.Context, client *rpcclient.Client, height uint64) error {
+	deadline := time.Now().Add(30 * time.Second)
+	for {
+		reqCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
+		err := client.Share.SharesAvailable(reqCtx, height)
```
renaynay: why can't you just call this once with a longer ctx deadline ?

nodebuilder/tests/tastora/framework.go
```diff
@@ -459,13 +509,129 @@ func getNodeImage() string {
...
+		time.Sleep(1 * time.Second)
+	}
+}
+
+// WaitBlobAvailable waits until Blob.Get succeeds for the given parameters.
+func (f *Framework) WaitBlobAvailable(ctx context.Context, client *rpcclient.Client, height uint64, namespace share.Namespace, commitment []byte) (*nodeblob.Blob, error) {
+	deadline := time.Now().Add(30 * time.Second)
+	for {
+		reqCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
+		blob, err := client.Blob.Get(reqCtx, height, namespace, commitment)
```
renaynay: same w this

nodebuilder/tests/tastora/framework.go
```diff
@@ -459,13 +509,129 @@ func getNodeImage() string {
...
+		time.Sleep(1 * time.Second)
+	}
+}
+
+// WaitNamespaceDataAvailable waits until GetNamespaceData succeeds for the given parameters.
+func (f *Framework) WaitNamespaceDataAvailable(ctx context.Context, client *rpcclient.Client, height uint64, namespace share.Namespace) (interface{}, error) {
+	deadline := time.Now().Add(30 * time.Second)
+	for {
+		reqCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
+		data, err := client.Share.GetNamespaceData(reqCtx, height, namespace)
```
renaynay: same

nodebuilder/tests/tastora/framework.go
```diff
@@ -459,13 +509,129 @@ func getNodeImage() string {
...
+		}
+		if time.Now().After(deadline) {
+			return nil, err
+		}
+		time.Sleep(1 * time.Second)
+	}
+}
+
+// SetupP2PConnectivity establishes P2P connections between bridge and light nodes.
+func (f *Framework) SetupP2PConnectivity(ctx context.Context, bridgeClient, lightClient *rpcclient.Client, label string) {
```
renaynay: this feels like a weird method - not sure what purpose it will serve across tests.

nodebuilder/tests/tastora/e2e_sanity_test.go
```diff
@@ -0,0 +1,360 @@
...
+	if testing.Short() {
+		t.Skip("skipping E2E sanity integration tests in short mode")
+	}
+	suite.Run(t, &E2ESanityTestSuite{})
+}
+
+func (s *E2ESanityTestSuite) SetupSuite() {
+	// Setup with minimal topology: 1 Bridge Node + 1 Light Node
+	s.framework = NewFramework(s.T(), WithValidators(1), WithBridgeNodes(1), WithLightNodes(1))
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
```
renaynay: test suite timeout and individual timeouts don't really align. Maybe this needs larger timeout

nodebuilder/tests/tastora/e2e_sanity_test.go
```diff
@@ -0,0 +1,360 @@
...
+	defer cancel()
+	s.Require().NoError(s.framework.SetupNetwork(ctx))
+
+	// Start the light node after network setup
+	s.framework.NewLightNode(ctx)
+}
+
+// TestBasicDASFlow validates the complete DAS workflow: submit blob → verify availability → retrieve data with cross-node coordination
+func (s *E2ESanityTestSuite) TestBasicDASFlow() {
```
renaynay, line 62: this test does much more than just DAS - can you please split this up into either test cases or completely separate tests?

nodebuilder/tests/tastora/e2e_sanity_test.go
```diff
@@ -0,0 +1,360 @@
...
+
+	// Step 4: Verify cross-node coordination
+	bridgeNamespaceData, err := bridgeClient.Share.GetNamespaceData(ctx, height, namespace)
+	s.Require().NoError(err, "bridge node should be able to get namespace data")
+
+	lightNamespaceData, err := lightClient.Share.GetNamespaceData(ctx, height, namespace)
+	s.Require().NoError(err, "light node should be able to get namespace data")
+
+	// Verify both nodes return the same data
+	bridgeShares := bridgeNamespaceData.Flatten()
```
renaynay: just bytes.equal here, no need for more LOCs

nodebuilder/tests/tastora/framework.go
gupadhyaya: we can add a query balance with multiple attempts as deterministic logic instead of waiting with delay, however this will add additional complexity, is that ok?

nodebuilder/tests/tastora/framework.go
gupadhyaya: could do:

```
import
"github.com/containerd/errdefs"
...

errdefs.IsNotFound(err)
```

nodebuilder/tests/tastora/framework.go
gupadhyaya: good suggestion.

nodebuilder/tests/tastora/framework.go
gupadhyaya: yes, we can do that.

nodebuilder/tests/tastora/framework.go
gupadhyaya: removed

nodebuilder/tests/tastora/e2e_sanity_test.go
gupadhyaya: increased to cover overall tests. whole test suite was taking around 1.5 mins only.

nodebuilder/tests/tastora/e2e_sanity_test.go
gupadhyaya: refactored to keep single node das test in e2e sanity and moved the rest to e2e p2p sanity (complex cross-node das tests)

nodebuilder/tests/tastora/framework.go
```diff
@@ -464,9 +529,88 @@ func getNodeImage() string {
...
+		}
+		nodeInfos[label] = info
+	}
+
+	labels := make([]string, 0, len(nodes))
+	for label := range nodes {
+		labels = append(labels, label)
+	}
+
+	for i, sourceLabel := range labels {
```
renaynay: what's all this?

nodebuilder/tests/tastora/framework.go
```diff
@@ -464,9 +529,88 @@ func getNodeImage() string {
...
+				f.t.Logf("✓ %s ↔ %s", sourceLabel, targetLabel)
+			} else {
+				f.t.Logf("⚠ %s failed to connect to %s (state: %v)", sourceLabel, targetLabel, state)
+			}
+		}
+	}
+}
+
+// WaitPeersReady waits until the node reports at least one peer.
+func (f *Framework) WaitPeersReady(ctx context.Context, client *rpcclient.Client, label string, timeout time.Duration) {
```
renaynay: do we need this ?

renaynay (commented):
* test timeouts are very long - more sane timeout would be MAX 30 seconds for the entire test to run.
* to make things more readable, you can set a context at the beginning of the test (as you've been doing) and use that context for all of the calls throughout the test  - unless you explicitly expect a very short deadline on a certain operation but for now, we do not define a test like this).

nodebuilder/tests/tastora/framework.go
gupadhyaya: many to many connection between peers. simplified a little bit. but needed.

nodebuilder/tests/tastora/framework.go
gupadhyaya: removed.

nodebuilder/tests/tastora/e2e_sanity_test.go
```diff
@@ -0,0 +1,257 @@
...
+
+	rpcclient "github.com/celestiaorg/celestia-node/api/rpc/client"
+	nodeblob "github.com/celestiaorg/celestia-node/blob"
+	"github.com/celestiaorg/celestia-node/header"
+	"github.com/celestiaorg/celestia-node/state"
+)
+
+// E2ESanityTestSuite provides E2E sanity testing of basic functionality flows.
+// Focuses on core functionality validation with minimal setup (1 BN + 1 LN).
+type E2ESanityTestSuite struct {
```
Wondertan, line 23: optional naming nit: I have no idea why we call it a sanity test, and why we call it so, but it may be fine. From my experience, this type of tests are called ["smoke"](https://en.wikipedia.org/wiki/Smoke_testing_(software)) or e2e tests. 

P.S. Looking at the wiki link, it does call smoke test a sanity test too

nodebuilder/tests/tastora/go.mod
```diff
@@ -380,6 +379,8 @@ replace (
 	github.com/tendermint/tendermint => github.com/celestiaorg/celestia-core v1.55.0-tm-v0.34.35
 )
 
-replace github.com/ipfs/boxo => github.com/celestiaorg/boxo v0.29.0-fork
+replace github.com/ipfs/boxo => github.com/celestiaorg/boxo v0.29.0-fork-4
 
 replace github.com/ipfs/go-datastore => github.com/celestiaorg/go-datastore v0.0.0-20250801131506-48a63ae531e4
+
+replace github.com/celestiaorg/celestia-node => ../../..
```
Wondertan, line 389: good call, then we don't need to update node depency all the time

nodebuilder/tests/tastora/framework.go
```diff
@@ -464,9 +529,57 @@ func getNodeImage() string {
...
+		return
+	}
+
+	// Get P2P info for all nodes
+	nodeInfos := make(map[string]peer.AddrInfo)
+	infoCtx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
+	for label, client := range nodes {
+		if info, err := client.P2P.Info(infoCtx); err != nil {
```
Wondertan: Timeout seems to be for the whole operation, but its is used only for Info request which is unlikely to ever be an issue.

nodebuilder/tests/tastora/framework.go
```diff
@@ -464,9 +529,57 @@ func getNodeImage() string {
...
+//
+// Example usage:
+//
+//	nodes := map[string]*rpcclient.Client{
+//	    "bridge": bridgeClient,
+//	    "light1": lightClient1,
+//	    "light2": lightClient2,
+//	}
+//	f.ConnectNodes(ctx, nodes, 30*time.Second)
+func (f *Framework) ConnectNodes(ctx context.Context, nodes map[string]*rpcclient.Client, timeout time.Duration) {
```
Wondertan: optional nit: it's cute that we pass timeout, but it shouldn't be a required argument IMO. Users can control the timeout by wrapping the context if they need to, and internally, we could set a constant deadline if its not set just in case. Although internally, those calls have timeouts already.

nodebuilder/tests/tastora/framework.go
```diff
@@ -464,9 +529,57 @@ func getNodeImage() string {
...
+	// Connect each node to all others
+	for sourceLabel, sourceClient := range nodes {
+		for targetLabel, targetInfo := range nodeInfos {
+			if sourceLabel == targetLabel || targetInfo.ID == "" {
+				continue
+			}
+
+			_ = sourceClient.P2P.Connect(ctx, targetInfo)
+
+			if state, _ := sourceClient.P2P.Connectedness(ctx, targetInfo.ID); state == network.Connected {
```
Wondertan: nit: this call is unnecessary, success on Connect guarantees Connectenness to be Connected fyi

nodebuilder/tests/tastora/e2e_sanity_test.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+
+	s.Assert().Equal(len(bridgeHeaders), len(lightHeaders), "should have same number of headers")
+	for i := 0; i < len(bridgeHeaders); i++ {
+		s.Assert().Equal(bridgeHeaders[i].Hash(), lightHeaders[i].Hash(), "headers at index %d should match", i)
+	}
+}
+
+// TestP2PConnectivity tests end-to-end P2P network connectivity workflow
+func (s *E2ESanityTestSuite) TestP2PConnectivity() {
```
Wondertan, line 228: This should do Disconnect and Connect API calls.

nodebuilder/tests/tastora/framework.go
Wondertan: While reviewing, I couldn't find this the new method being used anywhere. Is this expected?

nodebuilder/tests/tastora/e2e_sanity_test.go
```diff
@@ -0,0 +1,257 @@
+//go:build integration
```
Wondertan, line 1: Should we change the tag to be aligned with how we call the tests? sanity or smoke

nodebuilder/tests/tastora/framework.go
gupadhyaya: yea. will remove. this is needed in the follow-up PR. someone that code sneaked into this PR.

nodebuilder/tests/tastora/e2e_sanity_test.go
```diff
@@ -35,7 +35,7 @@ func TestE2ESanityTestSuite(t *testing.T) {
 func (s *E2ESanityTestSuite) SetupSuite() {
 	// Setup with minimal topology: 1 Bridge Node + 1 Light Node
 	s.framework = NewFramework(s.T(), WithValidators(1), WithBridgeNodes(1), WithLightNodes(1))
-	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
```
renaynay: let's extract timeout into a default for the suite and it can be a param on the suite

nodebuilder/tests/tastora/e2e_sanity_test.go
gupadhyaya: currently we use integration as default tag throughout. i would avoid having to maintain multiple tags. we can come back to this when we have different tastora tests and can clearly separate them into smoke vs api vs integration tests.

nodebuilder/tests/tastora/framework.go
gupadhyaya: for now removed ConnectNodes as its not used in this test, but will propagate this in the future tests.

nodebuilder/tests/tastora/framework.go
gupadhyaya: removed ConnectNodes, but will propagate this change in the future tests.

nodebuilder/tests/tastora/framework.go
gupadhyaya: removed ConnectNodes.

Wondertan (approved):

PR #4516, opened 2025-09-02
docs: add discovery spec for shrex

vgonkivs:
Added a specification for the peer discovery in shrex.

[Rendered](https://github.com/vgonkivs/celestia-node/tree/add_discovery_spec/specs/src/shrex/discovery.md)

### DIFF
--- a/specs/src/shrex/discovery.md
+++ b/specs/src/shrex/discovery.md
@@ -34,11 +34,6 @@ This specification defines the peer discovery mechanism for the SHREX protocol i
 The discovery service REQUIRES the libp2p DHT (Distributed Hash Table) to enable nodes to find peers capable of serving specific data types. The service handles both peer advertisement (announcing capabilities) and peer discovery (finding peers with required capabilities) while maintaining an active set of connected peers.
 The discovery service maintains a set in which all peers MUST have active connections with the host service. When connectedness changes from active to any other state, the peer MUST be immediately dropped and replaced through new discovery. The dropped peer SHOULD be added to a cache with backoff, so it will not be prematurely re-connected to through discovery.
 
-DHT Configuration: The kad-dht is configured with default libp2p parameters and operates in different modes based on node type:
-
-Light Nodes: dht.ModeClient - consume DHT services without serving DHT queries
-Bridge/Full Nodes: dht.ModeServer - both consume and serve DHT queries to support network infrastructure
-
 ## Sequence Diagrams
 
 ### Advertisement
@@ -167,6 +162,12 @@ The discovery service operates with the following configurable parameters:
 - **Default**: 10 Minutes
 - **Purpose**: Interval between peer connection attempts
 - **Rationale**: Prevents wasting resources by repeatedly trying to connect to a peer that's likely experiencing persistent issues, giving it time to recover while avoiding connection storms across the network.
+
+#### DHT Parameters
+The kad-dht is configured with default libp2p parameters and operates in different modes based on node type:
+ - **dht.ModeClient** - MUST be configured for LN(consume DHT services without serving DHT queries)
+ - **dht.ModeServer** MUST be configured for BN/FN(both consume and serve DHT queries to support network infrastructure)
+
 
 ### Discovery Operations
 


specs/src/shrex/discovery.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Terminology
+
+- **Discovery**: The main service that handles peer discovery operations
+- **Tag**: A string identifier that categorizes peers by their capabilities (e.g., "full", "archival")
+- **Limited Set**: A bounded collection that maintains discovered peers with configurable size limits
+- **Advertisement**: The process of announcing peer presence under a specific tag
+- **Parameters**: Configuration structure containing discovery settings
+
+## Overview
```
Wondertan: One comment I expect from the team is that FN is deprecated. While I still disagree with the decision to remove it, we should mention that in the spec, but not remove the FN tag entirely, as it is still part of the protocol.

specs/src/shrex/discovery.md
```diff
@@ -0,0 +1,208 @@
...
+### Core Components
+
+1. **Discovery**: Main service that handles advertisement and peer discovery operations
+2. **Limited Set**: Bounded collection that maintains discovered peers (accessed via `set.Size()`)
+3. **Parameters**: Configuration including `PeersLimit`, `AdvertiseInterval`
+4. **Peer Update Callbacks**: Event handlers for peer addition/removal notifications via `WithOnPeersUpdate`
+
+## Protocol Specification
+
+### Discovery Operations
```
Wondertan: Do you plan to fill them up?

specs/src/shrex/discovery.md
```diff
@@ -0,0 +1,208 @@
...
+### Tag-Based Organization
+
+- Nodes create separate Discovery instances for each tag they manage
+- Tags identify node capabilities
+- Discovery instances operate independently per tag
+
+## References
+
+1. **Celestia Node**: <https://github.com/celestiaorg/celestia-node>
+2. **libp2p Discovery**: <https://docs.libp2p.io/concepts/protocols/#peer-discovery>
```
Wondertan: We should also reference libp2p's dht specification

Wondertan: somewhere in libp2p/specs

specs/src/shrex/discovery.md
```diff
@@ -0,0 +1,208 @@
...
+This specification covers:
+
+- Peer discovery mechanisms for DA network
+- Tag-based peer advertisement and lookup
+- Integration with node types
+- Bounded peer set management
+
+## Requirements Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119.
```
Wondertan: While you mention this, none of the words are actually being used. E.g., LNs MUST NOT advertise themselves.

specs/src/shrex/discovery.md
```diff
@@ -0,0 +1,208 @@
...
+    // Size returns the current number of peers in the limited set
+    Size() int
+
+    // Peers provides a list of discovered peers in the given topic
+    // If Discovery hasn't found any peers, it blocks until at least one peer is found
+    func Peers(ctx context.Context) ([]peer.ID, error)
+
+    // Discard removes the peer from the peer set and rediscovers more if soft peer limit is not
+    // reached. Reports whether peer was removed with bool.
+    func Discard(id peer.ID) bool
```
Wondertan: Remove func

specs/src/shrex/discovery.md
```diff
@@ -0,0 +1,208 @@
...
+2. **Limited Set Management**: Discovered peers are added to bounded peer set
+3. **Automatic Updates**: Peer additions/removals trigger update notifications
+4. **Connection Tracking**: Monitors peer connectivity state
+
+**Typical Usage:**
+
+- **Light nodes** discover Full and Bridge nodes
+- **Full and Bridge nodes** may discover each other
+
+## API Reference
```
Wondertan: spec nit: Generally, specifications should aim to be language agnostic. If you still want to include code there, ensure the language is specified as well.

specs/src/shrex/discovery.md
```diff
@@ -0,0 +1,208 @@
...
+type Discovery interface {
+    // Advertise announces peer presence under the configured tag (FN/BN only)
+    Advertise(ctx context.Context) error
+
+    // Start begins the discovery service and launches the loop that performs
+    // peer discovery under specific topic.
+    Start(ctx context.Context) error
+
+    // Stop shuts down the discovery service
+    Stop(ctx context.Context) error
```
Wondertan: This is a very low-level detail and mplementation specific so I would remove it

specs/src/shrex/discovery.md
```diff
@@ -0,0 +1,208 @@
...
+
+// DefaultParameters returns the default Parameters' configuration values
+// for the Discovery module
+func DefaultParameters() *Parameters {
+    return &Parameters{
+        PeersLimit:        5, 
+        AdvertiseInterval: time.Hour,
+    }
+}
+```
```
Wondertan: Instead of the code, there should be a separate section on parameters. You already mentioned them, but expand on this with words, not code, together with defaults and some rationale behind them

specs/src/shrex/discovery.md
```diff
@@ -0,0 +1,208 @@
...
+### Tags
+
+```go
+const (
+    // fullNodesTag is the tag used to identify full nodes in the discovery service.
+    fullNodesTag = "full"
+    // archivalNodesTag is the tag used to identify archival nodes in the  discovery service.
+    archivalNodesTag = "archival"
+)
+```
```
Wondertan: Same here, these are implementation independent yet a vital part of the specification that should be described with words and MUST language.

Wondertan (commented):
First pass. Please add a link to the rendered markdown to see the rendered diagram.

Directionally, the spec is alright. My main concern is that implementation details are mixed into the part of specification. Particularly, parts that have to be specified in main body of the doc with words are somewhere hidden in implementation details in code, like tags, params and rationale for them.

specs/src/shrex/discovery.md
vgonkivs: Yeah, FN is deprecated but not deleted so far. I also thought whether add it or not to the doc, but chose adding it.

specs/src/shrex/discovery.md
vgonkivs: So, I guees i can leave `Advertise() `only

specs/src/shrex/discovery.md
vgonkivs: This is a section that consists of smaller sections.

specs/src/shrex/discovery.md
vgonkivs: Updated according to your comment

specs/src/shrex/discovery.md
vgonkivs: reworked

specs/src/shrex/discovery.md
```diff
@@ -0,0 +1,210 @@
...
+#### Advertisement Process
+
+1. Nodes MUST determine their appropriate tags based on capabilities
+2. Nodes SHOULD periodically advertise their presence under relevant tags
+3. Advertisement includes node's peer ID and network addresses
+
+#### Peer Discovery Process
+
+1. Nodes MUST query the DHT for peers under specific tags
+2. Discovery service maintains a limited set of discovered peers
```
renaynay: one thing is we need to mention connectedness status and that the peers in the limited set must *active connections* - otherwise they must drop the peer for which connectedness changes to anything but connected and discover a new active connection.

renaynay (commented):
We should make it very _clear_ that from a discovery standpoint - full and bridge nodes are identical in the services they can provide to light nodes and that is the rationale for both advertising under `full` tag. I think even though we specify that full nodes are deprecated, it's a bit confusing to follow in this doc.

Also mention the connectedness of the peers and what *kind* of connectedness we expect peers to have that we maintain in our limited peer set. Make sure to specify that the criteria for _staying_ in the limited peer set is that it is actively connected to the node at all times, and once that changes, that we kick peer + discover new one that meets connectedness criteria.

specs/src/shrex/discovery.md
```diff
@@ -23,10 +24,15 @@ This specification defines the peer discovery mechanism for the SHREX protocol i
...
+- **Limited Peer Set**: A bounded collection of actively connected peers maintained by the discovery service
+- **Active Connection**: A live, bidirectional connection between two peers that can carry protocol messages
+- **Connectedness**: The state of connectivity between two peers (connected, disconnected, connecting, etc.)
+- **Backoff**: The delay between connection attempts with a single peer.
 
 ## Overview
 
-The discovery service REQUIRES the libp2p DHT (Distributed Hash Table) to enable nodes to find peers capable of serving specific data types. The service handles both peer advertisement (announcing capabilities) and peer discovery (finding peers with required capabilities).
+The discovery service REQUIRES the libp2p DHT (Distributed Hash Table) to enable nodes to find peers capable of serving specific data types. The service handles both peer advertisement (announcing capabilities) and peer discovery (finding peers with required capabilities). The service handles both peer advertisement (announcing capabilities) and peer discovery (finding peers with required capabilities) while maintaining a limited set of actively connected peers.
```
renaynay: sentence here is duplicated

specs/src/shrex/discovery.md
```diff
@@ -23,10 +24,15 @@ This specification defines the peer discovery mechanism for the SHREX protocol i
...
+The discovery service maintains a limited peer set where all peers MUST have only active connections. When connectedness changes from connected to any other state, the peer MUST be immediately dropped and replaced through new discovery. The dropped peer is added to a cache with the backoff, so it won't be connected through the discovery in case backoff period is not ended.
```
renaynay: The discovery service maintains a limited peer set in which all peers MUST have active connections with the host service. When connectedness changes from active to any other state, the peer MUST be immediately dropped and replaced through new discovery. The dropped peer SHOULD be added to a cache with backoff so it will not be prematurely re-connected to through discovery.

specs/src/shrex/discovery.md
```diff
@@ -162,6 +177,26 @@ The discovery service operates with the following configurable parameters:
...
 
+### Connection Management
+
+The discovery service implements strict connection management to ensure reliable data availability:
+
+#### Active Connection Requirements
+
+- **Peer Set Criteria**: Only peers with active connections are maintained in the limited peer set
+- **Connection Monitoring**: The service MUST continuously monitor the connectedness status of all peers in the limited set
+- **Immediate Removal**: When a peer's connectedness changes from connected to any other state (disconnected, connecting, etc.), the peer MUST be immediately dropped from the limited set
```
renaynay: we can use `active` here to be clear, as we use it elsewhere.

renaynay (commented):
Besides these comments, I think it's pretty good coverage of discovery!

specs/src/shrex/discovery.md
vgonkivs: removed a limited set everywhere and changed to **Active Set** in glossary

walldiss:
Need to add network level params for dht configuration (per node type etc)

specs/src/shrex/discovery.md
```diff
@@ -0,0 +1,256 @@
...
+
+## Overview
+
+The discovery service REQUIRES the libp2p DHT (Distributed Hash Table) to enable nodes to find peers capable of serving specific data types. The service handles both peer advertisement (announcing capabilities) and peer discovery (finding peers with required capabilities) while maintaining an active set of connected peers.
+The discovery service maintains a set in which all peers MUST have active connections with the host service. When connectedness changes from active to any other state, the peer MUST be immediately dropped and replaced through new discovery. The dropped peer SHOULD be added to a cache with backoff, so it will not be prematurely re-connected to through discovery.
+
+DHT Configuration: The kad-dht is configured with default libp2p parameters and operates in different modes based on node type:
+
+Light Nodes: dht.ModeClient - consume DHT services without serving DHT queries
+Bridge/Full Nodes: dht.ModeServer - both consume and serve DHT queries to support network infrastructure
```
Wondertan: When rendered, this looks like a single paragraph. Also the overview is not section to put DHT configuration and modes.

PR #4524, opened 2025-09-04
config(nodebuilder/header): bumps go-header, removes trustingPeriod

renaynay:


Wondertan (commented):
Loc 105 also mentions an issue that was fixed and we could cleanup some code

nodebuilder/header/config.go
```diff
@@ -108,6 +106,9 @@ func (cfg *Config) Validate(tp node.Type) error {
 			)
 		}
 	case node.Light:
+		// enforce trusting period
+		sync.WithTrustingPeriod(trustingPeriod)(&cfg.Syncer)
+
```
walldiss: Is it necessary here? Seems off to overwrite cfg value inside Validate call

renaynay: Yes this is the fuglyness I was talking about and it's necessary bc of the private value in config --> hitting validation logic before the syncer constructs. So we have empty config value for trustingPeriod until syncer is constructed but we validate config before we even construct any modules.

walldiss: Would creating nodebuilder.DefaultSyncerCfg() with setted trusted window fix this issue?

nodebuilder/header/config.go
```diff
@@ -36,14 +36,16 @@ type Config struct {
 }
 
 func DefaultConfig(tp node.Type) Config {
+	syncerCfg := sync.DefaultParameters()
+	sync.WithTrustingPeriod(trustingPeriod)(&syncerCfg) // enforce default trusting period
```
walldiss: Should the default be enforced only for LN in this func?

renaynay: removed in latest commit.

nodebuilder/header/config.go
renaynay: It's now fixed in latest commit pls check

Wondertan (approved):
red diff 🍬

vgonkivs (approved):

PR #4526, opened 2025-09-05
fix(discovery): remove panic in switch statement

vgonkivs:
IMO, previous version had a hidden bug:
1) we cannot guarantee that status libp2p sends us that leads to unexpected panic(if we receive `Limited` connection).
2) libp2p may introduce a new status that can also lead to a panic.


share/shwap/p2p/discovery/metrics.go
```diff
@@ -15,13 +15,14 @@ import (
...
-	handlePeerConnErr     handlePeerResult = "conn_err"
-	handlePeerInSet       handlePeerResult = "in_set"
+	handlePeerResultKey                      = "result"
+	handlePeerSkipSelf      handlePeerResult = "skip_self"
+	handlePeerEnoughPeers   handlePeerResult = "skip_enough_peers"
+	handlePeerBackoff       handlePeerResult = "skip_backoff"
+	handlePeerConnected     handlePeerResult = "connected"
+	handlePeerConnErr       handlePeerResult = "conn_err"
+	handlePeerInSet         handlePeerResult = "in_set"
+	hanldePeerUnknownStatus handlePeerResult = "unknown_status"
```
renaynay: typo

renaynay (approved):
👍🏻

PR #4539, opened 2025-09-09
chore(blob|header|state): add and cleanup tracing and logs

### LINKED ISSUE #4530: blob: cover submission and retrieval with traces
We have traces integrated into the node, but we lack coverage on the most critical paths, which are blob submission and retrieval. Having them will provide introspection into the individual calls, allowing us to debug rare hiccups

Wondertan:
- Closes #4530
- Adds error and info logging for valuable endpoints

### DIFF
--- a/blob/service.go
+++ b/blob/service.go
@@ -176,12 +176,23 @@ func (s *Service) Subscribe(ctx context.Context, ns libshare.Namespace) (<-chan 
 // Uses default wallet registered on the Node.
 // Handles gas estimation and fee calculation.
 func (s *Service) Submit(ctx context.Context, blobs []*Blob, txConfig *SubmitOptions) (_ uint64, err error) {
+	ctx, span := tracer.Start(ctx, "blob/submit")
+	defer func() {
+		utils.SetStatusAndEnd(span, err)
+		if err != nil {
+			log.Errorw("submitting blobs failed", "err", err,
+				"err", err,
+			)
+		}
+	}()
+
 	libBlobs := make([]*libshare.Blob, len(blobs))
 	for i := range blobs {
 		libBlobs[i] = blobs[i].Blob
 	}
 
-	resp, err := s.blobSubmitter.SubmitPayForBlob(ctx, libBlobs, txConfig)
+	spanCtx := trace.ContextWithSpan(ctx, span)
+	resp, err := s.blobSubmitter.SubmitPayForBlob(spanCtx, libBlobs, txConfig)
 	if err != nil {
 		return 0, err
 	}
@@ -200,7 +211,7 @@ func (s *Service) Get(
 	commitment Commitment,
 ) (blob *Blob, err error) {
 	start := time.Now()
-	ctx, span := tracer.Start(ctx, "get")
+	ctx, span := tracer.Start(ctx, "blob/get")
 	defer func() {
 		utils.SetStatusAndEnd(span, err)
 		s.metrics.observeRetrieval(ctx, time.Since(start), err)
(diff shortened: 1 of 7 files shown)


state/core_access.go
```diff
@@ -163,13 +168,63 @@ func (ca *CoreAccessor) Stop(_ context.Context) error {
...
+		"amount", len(blobs),
+		"namespaces", ids,
+		"data-lengths", dataLengths,
+		"signer", cfg.SignerAddress(),
+		"key-name", cfg.KeyName(),
+		"gas-price", cfg.GasPrice(),
+		"gas-limit", cfg.GasLimit(),
+		"fee-granter", cfg.FeeGranterAddress(),
+		"tx-priority", int(cfg.TxPriority()),
+	)
```
Wondertan, line 253: I think it's helpful to find out whenever an endpoint was triggered. It might end up being spammier than before, but logging will not be limited to syncing only, and will have more value for node runners

Wondertan: Note that this is the case for every call

Wondertan:
We can also cover other state APIs if we think it's valuable

blob/service.go
```diff
@@ -262,29 +305,24 @@ func (s *Service) getAll(
...
-			blobs, err := s.getBlobs(ctx, namespace, header)
-			if err != nil && !errors.Is(err, ErrBlobNotFound) {
-				log.Errorf("getting blobs for namespaceID(%s): %v", hex.EncodeToString(namespace.ID()), err)
-				resultErr[i] = err
-			}
-			if len(blobs) > 0 {
-				log.Infow("retrieved blobs", "height", height, "total", len(blobs))
-				resultBlobs[i] = blobs
-			}
+			resultBlobs[i], resultErr[i] = s.getBlobs(ctx, namespace, header)
```
vgonkivs, line 340: `ErrBlobNotFound` should not  be set in a resultErr in getAll(https://github.com/celestiaorg/celestia-node/blob/4439d556591ceed99e68432f71bbc177f7315173/blob/service.go#L246)

Wondertan: It'ss still not set like it should, as I moved that code in getBlobs

Wondertan: Its much cleaner to have getAll focus on parallelization, while getBlobs focuses on nuances of getting blobs like handling this error.  Function responsibilities were a bit messy before.

blob/service.go
```diff
@@ -167,22 +170,17 @@ func (s *Service) Subscribe(ctx context.Context, ns libshare.Namespace) (<-chan
...
 // Handles gas estimation and fee calculation.
-func (s *Service) Submit(ctx context.Context, blobs []*Blob, txConfig *SubmitOptions) (uint64, error) {
-	log.Debugw("submitting blobs", "amount", len(blobs))
-
+func (s *Service) Submit(ctx context.Context, blobs []*Blob, txConfig *SubmitOptions) (_ uint64, err error) {
 	libBlobs := make([]*libshare.Blob, len(blobs))
 	for i := range blobs {
-		if err := blobs[i].Namespace().ValidateForBlob(); err != nil {
-			return 0, fmt.Errorf("not allowed namespace %s were used to build the blob", blobs[i].Namespace().ID())
-		}
```
Wondertan, line 182: I moved this level down to state.submit. I dont know why we did we check only here, but I have a feeling that there could have been something. Will revert if we find the reason

state/core_access.go
vgonkivs: maybe we should be make it debug? so node runners can optionally turn it on? It can be very spamming.

blob/service.go
vgonkivs: The signer verifies blobs during `NewMsgPayForBlobs` construction, so I guess we can skip this check at all.

state/core_access.go
Wondertan: Regular rollups submit once per celestia block which shouldnt be that spammy i believe

Wondertan: I am thinking that we can try info and if people or we complain we revert. Generally we should aim to show more in logs to optimize debugging of issues that may arrise with our partners

vgonkivs: Agree

vgonkivs:
> pls use utils.SetStatusend span

I've double checked and looks `utils.SetStatusAndEnd` is used everywhere

> and u can pass span through to core_access calls

It requires changes in the API as well.

PR #4541, opened 2025-09-10
doc: add shrex-sub spec

vgonkivs:
Added spec for the shree-sub sub-component

Co-Authored-By: Wondertan <hlibwondertan@gmail.com>


specs/src/shrex/shrex-sub.md
```diff
@@ -0,0 +1,178 @@
+# ShrEx/Sub Protocol Specification
+
+## Abstract
+
+**ShrEx/Sub** (Share Exchange/Subscribe) is a push-based notification protocol implementing the publish-subscribe pattern for Extended Data Square (EDS) hash distribution in the Celestia data availability network. The protocol enables efficient dissemination of new EDS availability notifications across different node types.
```
Wondertan: I wish we had a place to point to what EDS is

Wondertan: Maybe core specs?

specs/src/shrex/shrex-sub.md
```diff
@@ -0,0 +1,178 @@
...
+
+### PubSub Component
+
+The core publish-subscribe functionality for message distribution:
+
+**Interface Methods:**
+
+```text
+// Publish publishes data hash and height to the topic
+Publish(context, dataHash []byte, height uint64) -> error
```
Wondertan: Lets aim to be consistent with notations. Discovery detached from golang, even in implementation section, but here we start explaining the protocol directly with interfaces.

Wondertan: I think its fine to use golang in the spec generally, but in the some for of implementation detail section in the end.

specs/src/shrex/shrex-sub.md
```diff
@@ -0,0 +1,178 @@
...
+1. PubSub receives message from network
+2. Subscription processes message through validation
+3. Validated messages sent to registered listeners
+```
+
+**Lifecycle Management:**
+
+- All components must be started before message processing
+- Proper shutdown ensures cleanup of resources and connections
+
```
Wondertan: Is this AI generated?

Wondertan (commented):
* Rendered link would help even with no diagram
* Misses some form of motivation section where we explain the purpose behind this protocol. Yes, it delivers notifications, but what's the point?

specs/src/shrex/shrex-sub.md
vgonkivs: It was editing small parts and helping me to fix misspelling.

vgonkivs: Will remove this section.

specs/src/shrex/shrex-sub.md
vgonkivs: I reworked discovery. Please double check. No golang notations anymore.

Wondertan: That's what I mean here by it's not consistent with the discovery spec
I just don't think stepping away from golang is necessary in the implementation section. Up to you tho, as long as it consistent in our specs

vgonkivs: I'd leave language agnostic if you are ok with it.

specs/src/shrex/shrex-sub.md
vgonkivs: I did not find anything specific. But probably we can use https://github.com/celestiaorg/celestia-app/blob/6ec2d1cab99a130cbfd9a2ddecf5d59e20084fa3/specs/src/data_structures.md the Erasure Coding Section..

walldiss: I think this is a bit misleading to say that it distributes eds notifications. It is just new blocks being available. Notification mainly carries the height and eds root hash only to verify notification

specs/src/shrex/shrex-sub.md
```diff
@@ -0,0 +1,183 @@
...
+## Message Schema
+
+The notification message MUST contain the EDS data hash and block height:
+
+```text
+Notification {
+    data_hash: bytes[32]  // EDS root hash
+    height: uint64        // Block height
+}
+```
```
walldiss, line 56: need to specify encoding schema (json / proto )

specs/src/shrex/shrex-sub.md
```diff
@@ -0,0 +1,183 @@
...
+- [Additional Notes](#additional-notes)
+- [Links](#links)
+
+## Overview
+
+ShrEx/Sub is built on libp2p's FloodSub router with the following characteristics:
+
+- **Topic ID**: `/eds-sub/0.0.1`
+- **Message Distribution**: Flood-based (sends to all connected peers)
+- **Overlay**: No mesh topology (unlike GossipSub)
```
walldiss: Those two points looks like one. floodsub is no mesh

specs/src/shrex/shrex-sub.md
```diff
@@ -0,0 +1,183 @@
...
+Notification {
+    data_hash: bytes[32]  // EDS root hash
+    height: uint64        // Block height
+}
+```
+
+**Properties:**
+
+- Messages MUST have a fixed 40-byte payload (32 bytes hash + 8 bytes height)
+- Serialization overhead SHOULD be minimal
```
walldiss, line 63: Serialisation must be defined in spec

specs/src/shrex/shrex-sub.md
```diff
@@ -0,0 +1,183 @@
...
+- Each EDS at a specific height MUST generate only a single notification
```
walldiss: This needs is not a MUST right now. Currently it is ok if some message is not send or duplicates are also acceptable. So for current state of implementation lets replace MUST with SHOULD.

specs/src/shrex/shrex-sub.md
```diff
@@ -0,0 +1,183 @@
...
+## Message Verification
+
+### Validation Interface
+
+**MessageValidator Interface:**
+
+```text
+// Validate validates a message from a peer
+Validate(context, peerID PeerID, message []byte) -> ValidationResult
+```
```
walldiss, line 115: Need to add explanation why this is exposed and how it will be used by consumer of Validate (peer-manager)

specs/src/shrex/shrex-sub.md
Wondertan: Otherwise its like spec saying to itself that it should have a minimal overhead 😁

Wondertan:
> Misses some form of motivation section where we explain the purpose behind this protocol. Yes, it delivers notifications, but what's the point?


Motivation section is still missing

specs/src/shrex/shrex-sub.md
walldiss: I would also double check grammar here. "generate" might need to be to replaced with "produce" or "send"

vgonkivs: **produce** sgtm

specs/src/shrex/shrex-sub.md
```diff
@@ -0,0 +1,194 @@
...
+## Overview
+
+ShrEx/Sub is built on libp2p's FloodSub router with the following characteristics:
+
+- **Topic ID**: `/eds-sub/0.0.1`
+- **Message Distribution**: Flood-based (sends to all connected peers)
+
+### Node Roles
+
+- **Light Nodes (LN)**: Subscribers that receive EDS notifications for data availability sampling
```
renaynay:
```suggestion
- **Light Nodes (LN)**: Subscribers that receive EDS notifications for data availability sampling, blob retrieval, and other share-related requests (e.g. row retrieval).
```

specs/src/shrex/shrex-sub.md
```diff
@@ -0,0 +1,194 @@
...
+}
+```
+
+**Encoding**: Protocol Buffers (protobuf) serialization
+
+**Properties:**
+
+- Messages MUST have a fixed 40-byte payload (32 bytes hash + 8 bytes height)
+- Serialization overhead SHOULD be minimal
+- Each EDS at a specific height SHOULD produce only a single notification
```
renaynay, line 64: why SHOULD here?

specs/src/shrex/shrex-sub.md
renaynay: if we are going language agnostic then context is not needed here or in other signature definitions.

specs/src/shrex/shrex-sub.md
```diff
@@ -0,0 +1,194 @@
...
+- `REJECT`: Message is invalid and MUST be discarded
+- `IGNORE`: Message is valid but duplicate/stale and SHOULD be ignored
+
+The validation interface is exposed specifically for integration with the Peer Manager, which implements this interface to validate incoming ShrEx/Sub notifications. The Peer Manager uses validation results to add the announcing peer to the appropriate data hash pool, making them available for future data requests
+
+### Validation Pipeline
+
+ShrEx/Sub implementations MUST implement the following validation process:
+
+**Format Validation Requirements:**
```
renaynay: we need to add here that the header corresponding to the given height + datahash must be fully verified + included as part of implementation's subjective chain in order for notification to be considered valid. 

Maybe this was implied by 
> Height MUST be a valid block height

but we should be more concrete about what that means.

You could optionally also split this up into sanity checks vs verification check to make it more clear.

specs/src/shrex/shrex-sub.md
```diff
@@ -0,0 +1,194 @@
...
+
+FloodSub, on the other hand, sends messages to every "physical" connection without overlay mesh of logical links, which solves the problem with the cost of higher message duplication factor on the network. Although, a moderate amount of duplicates from different peers are helpful in this case. If the primary message sender peer does not serve data, the senders of duplicates are requested instead.
+
+**Trade-offs:**
+
+- **Higher Message Duplication**: Each message sent to every connection
+- **Bandwidth Cost**: ~40 bytes × peer_count per notification (32 bytes hash + 8 bytes height)
+- **Acceptable**: Given small message size and notification frequency
+
+## Additional Notes
```
renaynay, line 184: not ssure this is necessary to add but sure.

specs/src/shrex/shrex-sub.md
vgonkivs: Because we don't have any restrictions about that. The message will be ignored in this case and the peer won't be added to the pool.

specs/src/shrex/shrex-sub.md
```diff
@@ -0,0 +1,200 @@
...
+```text
+// Validate validates a message from a peer
+Validate(peerID PeerID, message []byte) -> ValidationResult
+```
+
+**ValidationResult Values:**
+
+- `ACCEPT`: Message is valid and MUST be processed
+- `REJECT`: Message is invalid and MUST be discarded
+- `IGNORE`: Message is valid but duplicate/stale and SHOULD be ignored
```
renaynay: we should *only* ignore a message if sender has already been seen for the exact same message.

specs/src/shrex/shrex-sub.md
```diff
@@ -0,0 +1,200 @@
...
+1. PubSub receives message from network
+2. Subscription processes message through validation
+3. Validated messages sent to registered listeners
+```
+
+## Protocol Behavior
+
+### Publisher Behavior (BN)
+
+- **Trigger**: Publishers MUST publish when a new EDS becomes available locally
```
renaynay: please distinguish "new EDS becomes available locally" from a syncing BN

publishers should only publish message if subjectively it understands that it is synced up and that this EDS is from a height at the tip of the network

specs/src/shrex/shrex-sub.md
```diff
@@ -0,0 +1,200 @@
...
+- **Action**: Publishers MUST publish EDS hash and height to `/eds-sub/0.0.1` topic
+- **Frequency**: Each EDS MUST be published exactly once per height
+- **Validation**: Publishers MUST only publish hashes for locally validated and available EDS
```
renaynay: should we define "available" at top of file?

specs/src/shrex/shrex-sub.md
```diff
@@ -0,0 +1,200 @@
...
+
+### Subscriber Behavior (LN)
+
+- **Subscription**: Subscribers MUST maintain an active subscription to `/eds-sub/0.0.1`
+- **Processing**: Subscribers MUST validate received hash format and height
+- **Action**: Subscribers MUST process notifications through registered listeners
+- **Deduplication**: Subscribers MUST ignore duplicate notifications
```
renaynay: need to specify what a duplicate is

also this feels like a `SHOULD`

specs/src/shrex/shrex-sub.md
vgonkivs: Peer manager detects duplication. Shrex-sub does not use `IGNORE` at all. Mistakenly added deduplication here.

specs/src/shrex/shrex-sub.md
vgonkivs: rephrased

specs/src/shrex/shrex-sub.md
vgonkivs: Removed available as it is already mentioned in `Trigger`

specs/src/shrex/shrex-sub.md
vgonkivs: removed deduplication section

specs/src/shrex/shrex-sub.md
```diff
@@ -118,7 +118,6 @@ Validate(peerID PeerID, message []byte) -> ValidationResult
 
 - `ACCEPT`: Message is valid and MUST be processed
 - `REJECT`: Message is invalid and MUST be discarded
-- `IGNORE`: Message is valid but duplicate/stale and SHOULD be ignored
```
renaynay: why was this removed? it's still a validation result that can/should be returned

specs/src/shrex/shrex-sub.md
```diff
@@ -151,17 +150,16 @@ ShrEx/Sub implementations MUST implement the following validation process:
...
 - **Frequency**: Each EDS MUST be published exactly once per height
-- **Validation**: Publishers MUST only publish hashes for locally validated and available EDS
+- **Validation**: Publishers MUST only publish hashes for locally validated EDS
 
 ### Subscriber Behavior (LN)
 
 - **Subscription**: Subscribers MUST maintain an active subscription to `/eds-sub/0.0.1`
 - **Processing**: Subscribers MUST validate received hash format and height
 - **Action**: Subscribers MUST process notifications through registered listeners
-- **Deduplication**: Subscribers MUST ignore duplicate notifications
```
renaynay: we should leave this and say `SHOULD`

PR #4546, opened 2025-09-11
fix(pruner): check nil context

### LINKED ISSUE #4543: bug(pruner): `onDelete` causes go-header to panic
Error: node: failed to start: error getting latest head during Start: subjective tail for head 8138303: moving tail from 8004805 to 8014206: deleting headers up to newly configured tail(8014206): header/store: delete to height 8014206: on delete handler for 8004805: header/store: user provided onDelete panicked on 8004805 with: runtime error: invalid memory address or nil pointer dereference

```
2025-09-11T11:55:15.787Z ERROR header/store store/store_delete.go:117 partial delete with error {"from_height": 8004805, "expected_to_height": 8014279, "actual_to_height": 8004805, "hdrs_not_found": 0, "took(s)": 0.000045947, "err": "on delete handler for 8004805: header/store: user provided onDelete panicked on 8004805 with: runtime error: invalid memory address or nil pointer dereference"}
```

ref: https://github.com/celestiaorg/go-header/issues/348

Wondertan:
Closes #4543 

This may happen when OnDelete is called before Pruner starts

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/benbjohnson/clock v1.3.5
 	github.com/celestiaorg/celestia-app/v5 v5.0.1
 	github.com/celestiaorg/go-fraud v0.2.3
-	github.com/celestiaorg/go-header v0.7.1
+	github.com/celestiaorg/go-header v0.7.2
 	github.com/celestiaorg/go-libp2p-messenger v0.2.2
 	github.com/celestiaorg/go-square/merkle v0.0.0-20240117232118-fd78256df076
 	github.com/celestiaorg/go-square/v2 v2.3.2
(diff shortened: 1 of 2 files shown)


pruner/service.go
```diff
@@ -254,7 +254,7 @@ func (s *Service) retryFailed(ctx context.Context) {
 // * If Pruner prunes a header first - Syncer only removes the header
 //   - pruneOnHeaderDelete ignores the header based on updated checkpoint state
 func (s *Service) pruneOnHeaderDelete(ctx context.Context, height uint64) error {
-	if s.ctx.Err() != nil {
+	if s.ctx != nil && s.ctx.Err() != nil {
```
walldiss, line 257: Should it error on nil context?
```suggestion
	if s.ctx == nil || s.ctx.Err() != nil {
```

PR #4549, opened 2025-09-12
feat(blob|state): Blob submission metrics

### LINKED ISSUE #4538: feat(blob/state): add metrics to submission
### Implementation ideas

Add metrics to cover major steps in blob submission path of blob and state module

gupadhyaya:
# Add Comprehensive Metrics for Blob Submission Path

## Overview

This PR adds comprehensive OpenTelemetry metrics to track performance and behavior of the blob submission path in both the `blob` and `state` modules. The metrics provide detailed insights into submission timing, error rates, gas estimation performance, and resource usage.

### Blob Module Metrics (`blob/metrics.go`)
- **Retrieval tracking**: Get operations, timing, and error rates  
- **Proof tracking**: Proof generation timing and success rates
- **Error categorization**: Detailed error tracking with context

**Note**: Submission metrics are handled by the state module to avoid duplication.

### State Module Metrics (`state/metrics.go`)
- **PFB submission tracking**: PayForBlob transaction metrics with gas details
- **Gas estimation metrics**: Timing and success rates for gas calculations
- **Gas price estimation**: Performance tracking for price estimation
- **Account query metrics**: Account lookup timing and error rates

### Blob Metrics
```
blob_retrieval_total
blob_retrieval_duration_seconds
blob_proof_total
blob_proof_duration_seconds
```

**Note**: Blob submission metrics are tracked in the state module to avoid duplication.

### State Metrics
```
state_pfb_submission_total{blob_count, total_size_bytes, gas_price_utia}
state_pfb_submission_duration_seconds
state_pfb_gas_estimation_duration_seconds
state_pfb_gas_price_estimation
state_gas_estimation_total
state_account_query_total
```

### 1. Enable Metrics in Node Configuration

```bash
# Start node with OTLP metrics enabled
./celestia bridge start --metrics --metrics.endpoint localhost:4318 --metrics.tls=false
./celestia light start --metrics --metrics.endpoint localhost:4318 --metrics.tls=false
```

### 2. Set Up OTLP Collector (Required for Grafana)

Create `otel-collector-config.yaml`:
```yaml
receivers:
  otlp:
    protocols:
      http:
        endpoint: 0.0.0.0:4318

exporters:
  prometheus:
    endpoint: "0.0.0.0:8889"
    namespace: "celestia"

service:
  pipelines:
    metrics:
      receivers: [otlp]
      exporters: [prometheus]
```

Run OTLP collector:
```bash
docker run -p 4318:4318 -p 8889:8889 -v $(pwd)/otel-collector-config.yaml:/etc/otel-collector-config.yaml otel/opentelemetry-collector-contrib:latest --config=/etc/otel-collector-config.yaml
```

### 3. Query Metrics Endpoint

```bash
# Get all metrics from OTLP collector
curl http://localhost:8889/metrics

# Filter blob metrics
curl http://localhost:8889/metrics | grep "blob_"

# Filter state metrics  
curl http://localhost:8889/metrics | grep "state_"
```

### 4. Monitor Specific Operations

```bash
# Monitor blob retrieval operations
curl -s http://localhost:8889/metrics | grep "blob_retrieval_total"

# Monitor blob proof generation
curl -s http://localhost:8889/metrics | grep "blob_proof_total"

# Monitor gas estimation performance
curl -s http://localhost:8889/metrics | grep "state_pfb_gas_estimation"
```

### 5. Prometheus Integration

Add to your `prometheus.yml`:
```yaml
scrape_configs:
  - job_name: 'celestia-node-otlp'
    static_configs:
      - targets: ['localhost:8889']  # OTLP collector's Prometheus endpoint
    metrics_path: '/metrics'
    scrape_interval: 15s
```

### 6. Grafana Dashboard

Configure Grafana to use Prometheus datasource pointing to `http://localhost:8889`

Create dashboards to visualize:
- Blob retrieval and proof generation performance
- State PFB submission performance  
- Error rates and types
- Gas estimation metrics
- Account query performance

### 7. Architecture Overview

```
Celestia Node (OpenTelemetry) → OTLP Collector → Prometheus Format → Grafana
```

- **Celestia Node**: Exports OpenTelemetry metrics to OTLP collector
- **OTLP Collector**: Converts OpenTelemetry metrics to Prometheus format
- **Grafana**: Visualizes metrics from Prometheus endpoint

## Example Metrics Output

```
# State PFB metrics (from OTLP collector)
celestia_pfb_count_total{instance="12D3KooW...",job="test/Bridge",label1="value1",otel_scope_name="state"} 5
celestia_last_pfb_timestamp_total{instance="12D3KooW...",job="test/Bridge",label1="value1",otel_scope_name="state"} 1.758100573267e+12

# Blob metrics
celestia_blob_retrieval_observable_total{instance="12D3KooW...",job="test/Bridge",label1="value1",otel_scope_name="blob"} 0
celestia_blob_proof_observable_total{instance="12D3KooW...",job="test/Bridge",label1="value1",otel_scope_name="blob"} 0

# Discovery metrics
celestia_archival_discovery_amount_of_peers{instance="12D3KooW...",job="test/Bridge",label1="value1",otel_scope_name="share_discovery"} 0
celestia_archival_discovery_find_peers_result_total{enough_peers="false",instance="12D3KooW...",job="test/Bridge",label1="value1",otel_scope_name="share_discovery"} 65
```

**Note**: Metrics are prefixed with `celestia_` and include OpenTelemetry scope information when exported through the OTLP collector.

Fixes #4538


blob/metrics.go
```diff
@@ -0,0 +1,297 @@
...
+var meter = otel.Meter("blob")
+
+// Metrics tracks blob-related metrics
+type Metrics struct {
+	// Submission metrics
+	submissionCounter   metric.Int64Counter
+	submissionDuration  metric.Float64Histogram
+	submissionErrors    metric.Int64Counter
+	submissionBlobCount metric.Int64Counter
+	submissionBlobSize  metric.Int64Counter
```
Wondertan: Duplicates state module metrics. Do we want this? Doesn't seem like

Wondertan (commented):
Solid. Mainly one comment about metrics duplication. Also wonder if you tested this on Graphana, so we don't accidentally get nil ptr or a metrics that doesn work

blob/metrics.go
gupadhyaya: you are right. don't need them in blob. hence, removed.

share/shwap/p2p/bitswap/blockstore_metrics.go
```diff
@@ -202,3 +202,8 @@ func (w *BlockstoreWithMetrics) AllKeysChan(ctx context.Context) (<-chan cid.Cid
 	))
 	return ch, err
 }
+
+func (w *BlockstoreWithMetrics) HashOnRead(enabled bool) {
```
Wondertan: This should not be needed. Recent boxo bump removed this from the interface.

state/metrics.go
```diff
@@ -2,15 +2,246 @@ package state
...
+	totalPfbSubmissionErrors      atomic.Int64
+	totalGasEstimations           atomic.Int64
+	totalGasEstimationErrors      atomic.Int64
+	totalGasPriceEstimations      atomic.Int64
+	totalGasPriceEstimationErrors atomic.Int64
+	totalAccountQueries           atomic.Int64
+	totalAccountQueryErrors       atomic.Int64
+}
+
+func WithMetrics(lc fx.Lifecycle, ca *CoreAccessor) (*Metrics, error) {
```
Wondertan: We usually make this a method on a struct that has/needs metrics. Its not the best approach, but it avoid the need to provide nil Metrics struct to FX which is quite ugly

nodebuilder/tests/tastora/framework.go
```diff
@@ -360,7 +360,7 @@ func (f *Framework) startBridgeNode(ctx context.Context, chain tastoratypes.Chai
 
 	err = bridgeNode.Start(ctx,
 		tastoratypes.WithChainID(testChainID),
-		tastoratypes.WithAdditionalStartArguments("--p2p.network", testChainID, "--core.ip", hostname, "--rpc.addr", "0.0.0.0"),
+		tastoratypes.WithAdditionalStartArguments("--p2p.network", testChainID, "--core.ip", hostname, "--rpc.addr", "0.0.0.0", "--metrics", "--p2p.metrics", "--metrics.endpoint", "host.docker.internal:4318", "--metrics.tls=false"),
```
Wondertan: This is part of the testing and doesnt need to be merged or?

gupadhyaya: correct.

state/metrics.go
gupadhyaya: changed to struct method. but we still need to provide nil metrics in case WithMetrics is not called right?

Wondertan: You don't need to provide nil Metrics if the Metrics struct is not part of the dependency graph.

Wondertan: I would suggest making the metrics struct private even. There is really no value in making it exposed. It's never gonna be used standalone

gupadhyaya: removed.

state/metrics.go
```diff
@@ -2,15 +2,247 @@ package state
...
+	totalGasEstimations           atomic.Int64
+	totalGasEstimationErrors      atomic.Int64
+	totalGasPriceEstimations      atomic.Int64
+	totalGasPriceEstimationErrors atomic.Int64
+	totalAccountQueries           atomic.Int64
+	totalAccountQueryErrors       atomic.Int64
+}
+
+// WithMetrics initializes metrics for the CoreAccessor
+func (ca *CoreAccessor) WithMetrics(lc fx.Lifecycle) (*Metrics, error) {
```
Wondertan: Let's not make individual components depend on FX. It's only a concern of the pkg that assembles everything together. You can achieve the same effect by exposing Start and Stop on the CoreAccessor struct and ensuring they are called in the module where everything is assembled.

gupadhyaya: added

gupadhyaya: removed fx and added stop method to unregister

blob/metrics.go
```diff
@@ -0,0 +1,196 @@
...
+
+	// Internal counters (thread-safe)
+	totalRetrievals      atomic.Int64
+	totalRetrievalErrors atomic.Int64
+	totalProofs          atomic.Int64
+	totalProofErrors     atomic.Int64
+}
+
+// WithMetrics registers blob metrics
+func WithMetrics(lc fx.Lifecycle) (*Metrics, error) {
```
Wondertan: Well, we need to do the same here...

state/core_access.go
```diff
@@ -155,6 +160,13 @@ func (ca *CoreAccessor) Stop(_ context.Context) error {
 		ca.estimatorConn = nil
 	}
 
+	// Stop metrics if they exist
+	if ca.metrics != nil {
```
Wondertan: You dont need to check it here if its already checked on the Stop method itself

state/core_access.go
```diff
@@ -501,13 +543,53 @@ func (ca *CoreAccessor) PayForBlobCount() int64 {
 	return ca.payForBlobCount
 }
 
+// SetMetrics updates the metrics for the CoreAccessor
+func (ca *CoreAccessor) SetMetrics(metrics *metrics) {
```
Wondertan: Lets remove this from public api as its cant and not going to be used. I think we can directly assign to metrics field in place without the method and locking.

blob/metrics.go
gupadhyaya: done.

state/core_access.go
gupadhyaya: right

state/core_access.go
gupadhyaya: done

state/core_access.go
```diff
@@ -508,6 +548,39 @@ func (ca *CoreAccessor) markSuccessfulPFB() {
...
+
+func (ca *CoreAccessor) observeAccountQuery(ctx context.Context, duration time.Duration, err error) {
+	if ca.metrics != nil {
+		ca.metrics.ObserveAccountQuery(ctx, duration, err)
+	}
+}
+
+func (ca *CoreAccessor) observeGasPriceEstimation(ctx context.Context, duration time.Duration, err error) {
+	if ca.metrics != nil {
+		ca.metrics.ObserveGasPriceEstimation(ctx, duration, err)
```
Wondertan: This is all redundant too as nil for metrics is already checked internally

nodebuilder/settings.go
```diff
@@ -86,11 +87,23 @@ func WithMetrics(metricOpts []otlpmetrichttp.Option, nodeType node.Type) fx.Opti
...
-			state.WithMetrics(lc, ca)
+			err := ca.WithMetrics()
+			if err != nil {
+				log.Warnf("failed to initialize state metrics: %v", err)
+			}
+		}),
+		fx.Invoke(func(serv *blob.Service) {
+			if serv == nil {
+				return
+			}
```
Wondertan: serv is never nil, so can remove

nodebuilder/settings.go
```diff
@@ -86,11 +87,23 @@ func WithMetrics(metricOpts []otlpmetrichttp.Option, nodeType node.Type) fx.Opti
...
+			err := serv.WithMetrics()
+			if err != nil {
+				log.Warnf("failed to initialize blob metrics: %v", err)
```
Wondertan: We should not start the node if metrics were requested but were not built

Wondertan (commented):
almost there

state/core_access.go
```diff
@@ -166,45 +176,78 @@ func (ca *CoreAccessor) SubmitPayForBlob(
...
 		}
 		gas = ca.estimateGasForBlobs(blobSizes)
+		gasEstimationDuration = time.Since(gasEstimationStart)
+		ca.metrics.ObserveGasEstimation(ctx, gasEstimationDuration, nil)
 	}
 
 	// get tx signer account name
 	author, err := ca.getTxAuthorAccAddress(cfg)
 	if err != nil {
+		finalErr = err
```
Wondertan: option micro nit: 
the finalErr is not necessary if you make named return var

Wondertan (dismissed):

blob/metrics.go
```diff
@@ -0,0 +1,188 @@
...
+		m.totalRetrievalErrors.Add(1)
+	}
+
+	// Record metrics with error type enum to avoid cardinality explosion
+	attrs := []attribute.KeyValue{}
+	if err != nil {
+		errorType := "unknown"
+		switch {
+		case errors.Is(err, ErrBlobNotFound):
+			errorType = "not_found"
```
Wondertan: Can we make these magic strings to be defined as constants?

state/metrics.go
```diff
@@ -2,41 +2,402 @@ package state
...
+	if err != nil {
+		m.totalPfbSubmissionErrors.Add(1)
+	}
+
+	// Record metrics
+	attrs := []attribute.KeyValue{
+		attribute.Int("blob_count", blobCount),
+		attribute.Int64("total_size_bytes", totalSize),
+		attribute.Float64("gas_price_utia", gasPrice),
+		attribute.Bool("success", err == nil),
```
Wondertan: Ideally we would define some error types here too, but that can be done as follow up once we enumarate all of them

Wondertan (dismissed):

Wondertan (dismissed):

Wondertan (dismissed):

blob/metrics.go
```diff
@@ -0,0 +1,197 @@
...
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+)
+
+var meter = otel.Meter("blob")
+
+// Error type constants for metrics labels
+const (
+	errorTypeNone     = "none"
```
renaynay: why do we even add err attribute if success

blob/service.go
```diff
@@ -61,19 +62,23 @@ type Service struct {
 	headerGetter func(context.Context, uint64) (*header.ExtendedHeader, error)
 	// headerSub subscribes to new headers to supply to blob subscriptions.
 	headerSub func(ctx context.Context) (<-chan *header.ExtendedHeader, error)
+	// metrics tracks blob-related metrics
+	metrics *metrics
```
renaynay, line 66: why do we pass it in as param if we need to configure it post-construction ?

state/core_access.go
```diff
@@ -184,26 +208,39 @@ func (ca *CoreAccessor) SubmitPayForBlob(
...
 	gas := cfg.GasLimit()
 	if gas == 0 {
+		gasEstimationStart := time.Now()
 		blobSizes := make([]uint32, len(libBlobs))
 		for i, blob := range libBlobs {
 			blobSizes[i] = uint32(len(blob.Data()))
 		}
 		gas = ca.estimateGasForBlobs(blobSizes)
+		gasEstimationDuration = time.Since(gasEstimationStart)
+		ca.metrics.ObserveGasEstimation(ctx, gasEstimationDuration, nil)
```
renaynay: why do we observe gas estimation separately + also as part of PFB submission?

state/metrics.go
```diff
@@ -2,41 +2,410 @@ package state
...
+	pfbGasEstimationDuration metric.Float64Histogram
+	pfbGasPriceEstimation    metric.Float64Histogram
+
+	// Gas estimation metrics
+	gasEstimationCounter  metric.Int64Counter
+	gasEstimationDuration metric.Float64Histogram
+	gasEstimationErrors   metric.Int64Counter
+
+	// Gas price estimation metrics
+	gasPriceEstimationCounter  metric.Int64Counter
```
renaynay: what's the point of the counter, we just need duration 

every time we calculate duration, we know gas price estimation event occurred.

state/metrics.go
```diff
@@ -2,41 +2,410 @@ package state
...
+	gasPriceEstimationDuration metric.Float64Histogram
+	gasPriceEstimationErrors   metric.Int64Counter
+
+	// Account operations metrics
+	accountQueryCounter  metric.Int64Counter
```
renaynay: same here

renaynay (changes requested):
* pls make sure go mods for our internal deps (celestiaorg) match what's on main branch (like don't update go-square)
* I'd prefer to do metric observations (time it took) in the helper methods themselves instead of inline in the submission code path as the submission code path is already extremely long and convoluted

renaynay (commented):
* Lint errors that need resolving
* thank you for adding grafana dashboard, I UTACK that
* might need to be rebased
* bunch of root level files that do not belong at root
* some of my comments re necessity of some of the metrics not addressed - pls let's not overbloat w unnecessary metrics.

blob/metrics.go
gupadhyaya: fixed

renaynay (dismissed):

gupadhyaya:
closing in favor of https://github.com/celestiaorg/celestia-node/pull/4664

PR #4550, opened 2025-09-12
chore: bump tastora version

chatton:
### Overview

This PR bumps tastora to [v0.7.0](https://github.com/celestiaorg/tastora/releases/tag/v0.7.0). 

The main changes are the removal of the provider pattern and moving everything to builders. Typed configuration for app configs (not yet node configs) and a more standarised way of accessing internal/external hostnames and ports.

I ran into some issues with using the pinned mocha tag of node in the tastora go.mod, in this PR I have a local pin. If that is note desirable we can figure out what makes the most sense for the tests.

Note: the functionality to dynamically add nodes has been added but I didn't want to add support for that at the same time, that can be done in a follow up.


nodebuilder/tests/tastora/go.mod
```diff
@@ -1,29 +1,34 @@
 module github.com/celestiaorg/celestia-node/nodebuilder/tests/tastora
 
-go 1.24.4
+go 1.24.6
+
+toolchain go1.24.7
+
+replace github.com/celestiaorg/celestia-node => ../../..
```
chatton: this was the main problem, there were dependency conflicts when using `v0.23.2-mocha`, I'm not sure if it's desired to be pinned to a specific node version from these tests, or if the tests should use whatever is on the current branch.

nodebuilder/tests/tastora/framework.go
```diff
@@ -275,76 +293,69 @@ func (f *Framework) fundNodeAccount(ctx context.Context, daNode *tastoradockerty
...
+			"--timeout-commit", "1s",
+		).
+		WithPostInit(func(ctx context.Context, node *cosmos.ChainNode) error {
+			if err := config.Modify(ctx, node, "config/config.toml", func(cfg *cometcfg.Config) {
+				cfg.TxIndex.Indexer = "kv"
+			}); err != nil {
+				return err
+			}
+			return config.Modify(ctx, node, "config/app.toml", func(cfg *servercfg.Config) {
+				cfg.GRPC.Enable = true
```
gupadhyaya, line 363: previously we had both app and config overrides set `tx_index`. does this change here?

nodebuilder/tests/tastora/framework.go
```diff
@@ -275,76 +293,69 @@ func (f *Framework) fundNodeAccount(ctx context.Context, daNode *tastoradockerty
...
+		Build()
 
-	// verify the chain is producing blocks
-	require.NoError(f.t, wait.ForBlocks(ctx, 2, celestia))
-	return celestia
+	daNetworkBuilder := dataavailability.NewNetworkBuilderWithTestName(f.t, f.t.Name()).
+		WithDockerClient(f.client).
+		WithDockerNetworkID(f.network).
+		WithImage(daImage).
+		WithNodes(bridgeNodeConfig)
```
gupadhyaya, line 389: did we remove the static topology and we no longer need the bridge and light node counts to be passed over?

nodebuilder/tests/tastora/framework.go
chatton: There is actually no tx_index field in app.toml, I think that originally came from some AI hallucination, the field only exists in `config.toml`, I think it ended up in the celestia-app one too but it shouldn't be required.

nodebuilder/tests/tastora/framework.go
chatton: yes you no longer need to specify the number of nodes up front and can dynamically add them with `daNetworkBuilder.WithNode(someCfg)` before calling `daNetworkBuilder.Build`

You can also call `daNetwork.AddNodes(...)` on an already created da network to add additional nodes

gupadhyaya (approved):

nodebuilder/tests/tastora/go.mod
```diff
@@ -1,16 +1,21 @@
 module github.com/celestiaorg/celestia-node/nodebuilder/tests/tastora
 
-go 1.24.4
+go 1.24.6
+
+toolchain go1.24.7
```
Wondertan, line 5: Do we need this?

Wondertan (dismissed):

PR #4551, opened 2025-09-12
Bump app in v0.27 0

vgonkivs:
Ports main commits into release branch + bumps app and square deps to fix square construction.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -378,6 +378,7 @@ require (
 
 replace (
 	cosmossdk.io/x/upgrade => github.com/celestiaorg/cosmos-sdk/x/upgrade v0.2.0
+	github.com/celestiaorg/celestia-app/v6 => github.com/celestiaorg/celestia-app/v6 v6.0.1-arabica.0.20250915095140-6d59b05df60a
 	github.com/cometbft/cometbft => github.com/celestiaorg/celestia-core v0.39.3
 	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v0.51.0
 	github.com/cosmos/ibc-go/v8 => github.com/celestiaorg/ibc-go/v8 v8.7.2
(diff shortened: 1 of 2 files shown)


state/core_access.go
```diff
@@ -203,6 +194,14 @@ func (ca *CoreAccessor) SubmitPayForBlob(
 		return nil, fmt.Errorf("account for signer %s not found", author)
 	}
 
+	gas := cfg.GasLimit()
+	if gas == 0 {
+		gas, err = ca.estimateGasForBlobs(account.Address().String(), libBlobs)
```
renaynay, line 199: this path looks like it gets overwritten anyway in `BroadcastPayForBlobWithAccount` -- which is interesting because AFAIU it would also write any custom value set in txconfig..

renaynay: It would be nice if we had some sort of integration test that would ensure that values set in config were actually the values with which txclient builds + signs the tx.

vgonkivs: Their gas estimation is prepended to the passed options. So, our values will overwrite their.

state/estimate.go
```diff
@@ -66,6 +67,10 @@ func (ca *CoreAccessor) estimateGasPriceAndUsage(
 }
 
 // estimateGasForBlobs returns a gas limit that can be applied to the `MsgPayForBlob` transactions.
-func (ca *CoreAccessor) estimateGasForBlobs(blobSizes []uint32) uint64 {
-	return apptypes.DefaultEstimateGas(blobSizes)
+func (ca *CoreAccessor) estimateGasForBlobs(signer string, blobs []*libshare.Blob) (uint64, error) {
```
renaynay, line 70: Please note this logic needs to be deduplicated between node and `TxClient` but we can consider it part of `TxClient` fixussy effort.

PR #4566, opened 2025-09-17
fix(cel-shed): start the store

Wondertan:
Otherwise, the commands are stuck forever.

Not sure how this was missed, maybe during some rebase, because I used this cmds myself


renaynay (approved):
utack

PR #4567, opened 2025-09-18
shwap(shrex/server): fix logging on server side

vgonkivs:
Changed log level on server side in case if header is not found in store. It's not necessary bloating logs with errors since this is a normal case when header is not found.

### DIFF
--- a/share/shwap/p2p/shrex/server.go
+++ b/share/shwap/p2p/shrex/server.go
@@ -169,7 +169,7 @@ func (srv *Server) handleDataRequest(ctx context.Context, requestID request, str
 
 	if err != nil {
 		if errors.Is(err, store.ErrNotFound) {
-			logger.Errorf("file not found in store")
+			logger.Warn("file not found in store")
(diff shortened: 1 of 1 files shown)


share/shwap/p2p/shrex/server.go
```diff
@@ -169,7 +169,7 @@ func (srv *Server) handleDataRequest(ctx context.Context, requestID request, str
 
 	if err != nil {
 		if errors.Is(err, store.ErrNotFound) {
-			logger.Errorf("file not found in store")
+			logger.Warn("file not found in store")
```
Wondertan: Lets use EDS instead of file

Wondertan:
```suggestion
			logger.Warn("EDS not found in local store")
```

vgonkivs: sure

Wondertan (approved):

PR #4568, opened 2025-09-18
fix(share/shwap): mismatched row roots error formatting

sashass1315:
Replace printing a slice with its length in the error message within RangeNamespaceData.verifyShares. Previously, the code passed the slice expectedRoots to a %d formatter, producing malformed output (e.g., %!d(MISSING=...)) instead of the intended count. This change aligns the message with the rest of the codebase, improves debuggability, and does not affect runtime logic.


renaynay (approved):

walldiss (approved):
Good find!

PR #4578, opened 2025-09-23
deps(go.mod): bump app to arabica release

renaynay:


cmwaters:
App just cut v5.0.6-arabica and will cut the -mocha version soon. We should be good to upgrade to that

PR #4595, opened 2025-09-26
feat: add the ability to submit blobs in parallel

evan-forbes:
This is a draft until we get an official release from app, but it would be super useful for me to get reviews from node folks as I do not know what I'm doing in this repo :sweat_smile: 

Currently, node automatically submits blobs asynchronously. Meaning we can submit multiple per block, but the ordering isn't guaranteed.

This adds the ability to submit blobs in parallel and the ability to submit blobs synchronously.

 Synchronous (workers == 1) has a single worker that has a queue of blobs. It submits the blob, waits for it to be confirmed, and then submits the next blob in the queue. This doesn't guarantee order, as its possible a tx fails or gets evicted and the next blob in the queue does not. However most importantly avoids sequence mismatches since we're only ever submitting a single blob from an account at once. The account that is submitting the blob here is the one passed to TxClient.
 
Parallel submission works in the exact same way, except there are more than one worker. In order to manage these accounts, it creates new ones and grants them a feegrant by submitting a single transaction for all accounts that don't already have this. This ofc means that different accounts will submit blobs. All accounts are in the same keyring that has the main account.

### DIFF
--- a/docs/adr/adr-009-public-api.md
+++ b/docs/adr/adr-009-public-api.md
@@ -412,6 +412,14 @@ type BankModule interface {
 }
 ```
 
+Nodes can enable parallel PayForBlob submission by configuring `state.WorkerAccounts`.
+The default value of `1` retains the existing single-account behaviour, while setting
+`0` disables the parallel pool entirely. Values greater than `1` cause the node to
+automatically create `parallel-worker-*` accounts and issue fee grants so the default
+signer pays their transaction fees. Parallel submission currently only applies when
+the default signer and fee granter are used; specifying alternate values in
+`TxConfig` falls back to the single-account path.
+
 ##### Staking (same as pre-mainnet staking module)
 
 ```go
--- a/nodebuilder/state/config.go
+++ b/nodebuilder/state/config.go
@@ -1,6 +1,8 @@
 package state
 
 import (
+	"fmt"
+
 	"github.com/cosmos/cosmos-sdk/crypto/keyring"
 
 	"github.com/celestiaorg/celestia-node/libs/utils"
@@ -19,6 +21,17 @@ type Config struct {
 	// EnableEstimatorTLS specifies whether to use TLS for the gRPC connection to the
 	// estimator service
 	EnableEstimatorTLS bool
+	// WorkerAccounts defines how many accounts the TxClient should manage for
+	// PayForBlob submissions. A value of 0 disables queued submission entirely, which
+	// results in submitting blobs immediately without waiting for previous blobs to be
+	// confirmed. This is not reccomended at this time. Setting the value to a value of
+	// 1 enables queued submission, which means blobs are added to a queue and submitted
+	// one after another. No additional accounts are initialized. Values greater than 1
+	// enable automatic creation and management of additional worker accounts for
+	// parallel submissions. This means that blobs can be submitted by multiple different
+	// signers, and that blobs will not be submitted on chain in the original sending order.
+	// This is highly reccomended for high throughput chains.
+	WorkerAccounts int
 }
 
 func DefaultConfig() Config {
@@ -26,11 +39,16 @@ func DefaultConfig() Config {
 		DefaultKeyName:     DefaultKeyName,
 		DefaultBackendName: defaultBackendName,
 		EstimatorAddress:   "",
+		WorkerAccounts:     1,
 	}
 }
 
 // Validate performs basic validation of the config.
 func (cfg *Config) Validate() error {
+	if cfg.WorkerAccounts < 0 {
+		return fmt.Errorf("worker accounts must be zero or positive")
+	}
+
 	if cfg.EstimatorAddress == "" {
 		return nil
 	}
--- a/nodebuilder/state/flags.go
+++ b/nodebuilder/state/flags.go
@@ -2,6 +2,7 @@ package state
 
 import (
 	"fmt"
+	"strconv"
 
 	"github.com/spf13/cobra"
 	flag "github.com/spf13/pflag"
@@ -12,6 +13,7 @@ var (
 	keyringBackendFlag          = "keyring.backend"
 	estimatorServiceAddressFlag = "estimator.service.address"
 	estimatorServiceTLSFlag     = "estimator.service.tls"
+	txWorkerAccountsFlag        = "tx.worker.accounts"
 )
 
 // Flags gives a set of hardcoded State flags.
@@ -36,6 +38,12 @@ func Flags() *flag.FlagSet {
 		false,
 		"enables TLS for the estimator service gRPC connection",
 	)
+	flags.Int(
+		txWorkerAccountsFlag,
+		1,
+		"specifies how many accounts the TxClient manages for PayForBlob submission.\n"+
+			"0 disables parallel submission. Values greater than 1 automatically create \"parallel-worker-*\" accounts and grant them fees using the default signer.",
+	)
 
 	return flags
 }
@@ -55,5 +63,12 @@ func ParseFlags(cmd *cobra.Command, cfg *Config) {
 
 	if cmd.Flag(estimatorServiceTLSFlag).Changed {
 		cfg.EnableEstimatorTLS = true
+	}
+
+	if cmd.Flag(txWorkerAccountsFlag).Changed {
+		value, err := strconv.Atoi(cmd.Flag(txWorkerAccountsFlag).Value.String())
+		if err == nil {
+			cfg.WorkerAccounts = value
+		}
 	}
 }
--- a/state/core_access.go
+++ b/state/core_access.go
@@ -51,6 +51,12 @@ type CoreAccessor struct {
 
 	keyring keyring.Keyring
 	client  *user.TxClient
+	// workerAccounts defines how many accounts the tx client manages for
+	// PayForBlob submissions. A value of zero keeps async submission.
+	// Value of 1 uses synchonous submission. Value of > 1 uses parallel
+	// submission. Parallel and async submission are not guaranteed to
+	// include blobs in the same order as they were submitted.
+	workerAccounts int
 
 	// TODO @renaynay: clean this up -- only one!!!!
 	defaultSignerAccount string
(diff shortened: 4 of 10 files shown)


docs/adr/adr-009-public-api.md
```diff
@@ -101,15 +101,15 @@ GetByHeight(ctx context.Context, height uint64) (*header.ExtendedHeader, error)
 // by the node's header store or until context deadline is exceeded.
 WaitForHeight(ctx context.Context, height uint64) (*header.ExtendedHeader, error)
 // GetRangeByHeight returns the given range (from:to) of ExtendedHeaders
-// from the node's header store and verifies that the returned headers are 
+// from the node's header store and verifies that the returned headers are
```
evan-forbes, line 104: all of the changes to this file are I think due to ai slop auto correcting, but hey why not avoid a few airdrop farming PRs

state/core_access.go
```diff
@@ -213,7 +225,7 @@ func (ca *CoreAccessor) SubmitPayForBlob(
 		opts = append(opts, feeGrant)
 	}
 
-	response, err := client.SubmitPayForBlobWithAccount(ctx, account.Name(), libBlobs, opts...)
+	response, err := client.SubmitPayForBlob(ctx, libBlobs, opts...)
```
evan-forbes: this is the bulk of the change

nodebuilder/state/config.go
```diff
@@ -19,18 +21,34 @@ type Config struct {
...
+	// PayForBlob submissions. A value of 0 disables queued submission entirely, which
+	// results in submitting blobs immediately without waiting for previous blobs to be
+	// confirmed. This is not recommended at this time. Setting the value to a value of
+	// 1 enables queued submission, which means blobs are added to a queue and submitted
+	// one after another. No additional accounts are initialized. Values greater than 1
+	// enable automatic creation and management of additional worker accounts for
+	// parallel submissions. This means that blobs can be submitted by multiple different
+	// signers, and that blobs will not be submitted on chain in the original sending order.
+	// This is highly recommended for high throughput chains.
+	WorkerAccounts int
```
renaynay:
```suggestion
	TxWorkerAccounts int
```

nodebuilder/state/flags.go
```diff
@@ -56,4 +65,11 @@ func ParseFlags(cmd *cobra.Command, cfg *Config) {
 	if cmd.Flag(estimatorServiceTLSFlag).Changed {
 		cfg.EnableEstimatorTLS = true
 	}
+
+	if cmd.Flag(txWorkerAccountsFlag).Changed {
+		value, err := strconv.Atoi(cmd.Flag(txWorkerAccountsFlag).Value.String())
```
renaynay, line 70: we can fatal out on err here

state/core_access.go
```diff
@@ -524,6 +530,10 @@ func (ca *CoreAccessor) setupTxClient(ctx context.Context) error {
 		ca.estimatorConn = estimatorConn
 	}
 
+	if ca.workerAccounts > 0 {
```
renaynay: maybe > 1 since default is 1

state/option.go
```diff
@@ -28,3 +28,12 @@ func WithAdditionalCoreEndpoints(conns []*grpc.ClientConn) Option {
 		ca.coreConns = append(ca.coreConns, conns...)
 	}
 }
+
+// WithWorkerAccounts configures the CoreAccessor to manage the provided number of
+// worker accounts via the TxClient. A value of zero leaves parallel
+// submission disabled.
+func WithWorkerAccounts(workerAccounts int) Option {
```
renaynay:
```suggestion
func WithTxWorkerAccounts(workerAccounts int) Option {
```

renaynay (commented):
mostly good w me, mostly about namings

state/core_access.go
```diff
@@ -213,15 +219,23 @@ func (ca *CoreAccessor) SubmitPayForBlob(
 		opts = append(opts, feeGrant)
 	}
 
-	response, err := client.SubmitPayForBlobWithAccount(ctx, account.Name(), libBlobs, opts...)
+	var response *user.TxResponse
+	if account.Name() == ca.defaultSignerAccount {
```
renaynay: Note that this is ugly, but until we have TxClient refactor, it will remain ugly.

Also, once we get `Signer` field in `TxResponse`, we should implement unit test for this. For now, manually tested.

renaynay:
# For documentation purposes (cc @jcstein )

This PR introduces a new feature called **parallel transaction submission lanes**, introduced in https://github.com/celestiaorg/celestia-app/pull/5776. 

## New feature - Parallel Transaction Submission Lanes

To use this feature, node runners must specify `TxWorkerAccounts` in the `state` config. 

```toml
  [State]
    DefaultKeyName = "my_celes_key"
    DefaultBackendName = "test"
    EstimatorAddress = ""
    EnableEstimatorTLS = false
    TxWorkerAccounts = 8
```

The number of `TxWorkerAccounts` specified will indicate to `TxClient` to initialise the given number of "parallel transaction submission lanes" -- _subaccounts funded by the "master" default account that will submit PayForBlob transactions on behalf of the submitter in order to bypass the limitations around account sequences, **allowing for high throughput PayForBlob submission (e.g. at minimum the # of TxWorkerAccounts amount of PayForBlob transactions per block)**._

So, if the user specifies `TxWorkerAccounts = 8`, that means there will be 7 additional subaccounts generated + funded by the default account that will be available to submit blobs, so the user will be able to reliably submit _at least_ 8 separate PayForBlob transactions per block.

### ⚠️  VERY IMPORTANT NOTES ⚠️ 

* This feature CANNOT be used by implementations that need sequential ordering of transactions. 
* This feature can ONLY be used for implementations that can allow unordered transactions.
* If an account other than the default account is specified in the `TxConfig`, parallel transaction submission will be **bypassed.** This feature only works for PayForBlob transactions submitted from the default configured account.

## Default behaviour

The default `TxWorkerAccounts` value is 1. 

A value of 1 would indicate to the `TxClient` **not** to enable parallel tx submission and all PayForBlob transactions would be submitted through the default account on the `TxClient` (unless another account was specified in the `TxConfig`). 

This preserves old behaviour with the caveat that now, users will still be able to call `SubmitPayForBlob` multiple times, but every call to this endpoint will queue up transactions to be submitted by the default account. Subsequent transactions in the queue will be processed only once the current transaction attains a definitive status (Evicted, Confirmed, Rejected).

**This could result in a submission cadence of 1 PayForBlob transaction per block.** But it preserves sequential ordering of transactions and ensures that the default signer will always be the author of the PayForBlob transaction (unless otherwise specified in the `TxConfig`).

However, if the user specifies an account other than the default account in the `TxConfig`, the transaction queue will be bypassed and the transactions can be submitted directly into the mempool without blocking on a definitive transaction status for the previously submitted transaction.

evan-forbes (commented):
I can't approve my own PR but I approve the changes

nodebuilder/state/config.go
```diff
@@ -19,18 +21,32 @@ type Config struct {
...
 	EnableEstimatorTLS bool
+	// TxWorkerAccounts defines how many accounts the tx client manages for
+	// PayForBlob submissions.
+	//   - Value of 0 submits transactions immediately (without a submission queue).
+	//   - Value of 1 uses synchronous submission (submission queue with default
+	//     signer as author of transactions).
+	//   - Value of > 1 uses parallel submission (submission queue with several accounts
+	//     submitting blobs). Parallel submission is not guaranteed to include blobs
+	//     in the same order as they were submitted.
+	TxWorkerAccounts int
```
cmwaters, line 32: nit:
```suggestion
	// TxWorkerAccounts is used for queued submission. It defines how many accounts the 
	// tx client uses for PayForBlob submissions.
	//   - Value of 0 submits transactions immediately (without a submission queue).
	//   - Value of 1 uses synchronous submission (submission queue with default
	//     signer as author of transactions).
	//   - Value of > 1 uses parallel submission (submission queue with several accounts
	//     submitting blobs). Parallel submission is not guaranteed to include blobs
	//     in the same order as they were submitted.
	TxWorkerAccounts int
```

nodebuilder/state/flags.go
```diff
@@ -36,6 +38,13 @@ func Flags() *flag.FlagSet {
 		false,
 		"enables TLS for the estimator service gRPC connection",
 	)
+	flags.Int(
+		txWorkerAccountsFlag,
+		1,
+		"specifies how many accounts the TxClient manages for PayForBlob submission.\n"+
+			"0 disables parallel submission. Values greater than 1 automatically create "+
+			"\"parallel-worker-*\" accounts and grant them fees using the default signer.",
```
cmwaters, line 46: nit: specify behaviour of when this is set to 1
```suggestion
		"is used for queued submission. It defines how many accounts the TxClient uses for PayForBlob submission.\n"+
			"0 disables queued submission. 1 uses the default account. Values greater than 1 automatically create "+
			"\"parallel-worker-*\" accounts and grant them fees using the default signer.",
```

nodebuilder/state/flags.go
```diff
@@ -36,6 +38,13 @@ func Flags() *flag.FlagSet {
 		false,
 		"enables TLS for the estimator service gRPC connection",
 	)
+	flags.Int(
+		txWorkerAccountsFlag,
+		1,
```
cmwaters, line 43: Is this the default? Don't we want it to be 0?

state/core_access.go
```diff
@@ -51,6 +51,15 @@ type CoreAccessor struct {
...
 	client  *user.TxClient
+	// workerAccounts defines how many accounts the tx client manages for
+	// PayForBlob submissions.
+	//   - Value of 0 submits transactions immediately (without a submission queue).
+	//   - Value of 1 uses synchronous submission (submission queue with default
+	//     signer as author of transactions).
+	//   - Value of > 1 uses parallel submission (submission queue with several accounts
+	//     submitting blobs). Parallel submission is not guaranteed to include blobs
+	//     in the same order as they were submitted.
+	workerAccounts int
```
cmwaters, line 62: nit: for consistency
```suggestion
	txWorkerAccounts int
```

cmwaters (approved):
Change looks good. I only have a few nits

PR #4596, opened 2025-09-26
fix(share/shwap/p2p/shrex): defer stream close in doRequest

renaynay:
Fixes stream leak in shrex

Found by @vgonkivs

### DIFF
--- a/share/shwap/p2p/shrex/client.go
+++ b/share/shwap/p2p/shrex/client.go
@@ -84,6 +84,12 @@ func (c *Client) doRequest(
 	if err != nil {
 		return statusOpenStreamErr, fmt.Errorf("open stream: %w", err)
 	}
+	defer func() {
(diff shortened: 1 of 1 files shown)


share/shwap/p2p/shrex/client.go
```diff
@@ -84,6 +84,12 @@ func (c *Client) doRequest(
 	if err != nil {
 		return statusOpenStreamErr, fmt.Errorf("open stream: %w", err)
 	}
+	defer func() {
+		err = stream.Close()
+		if err != nil {
+			log.Errorw("closing stream", "err", err, "peer", peer.String())
```
Wondertan: Let's make it warning, its not critical

vgonkivs:
You can use utils.CloseAndLog instead

vgonkivs (approved):

PR #4599, opened 2025-09-26
feat(core): allow configurable concurrencyLimit for exchange

renaynay:
Breaks config - requires `config-update`

### DIFF
--- a/core/exchange.go
+++ b/core/exchange.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"runtime"
 	"time"
 
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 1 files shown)


nodebuilder/core/config.go
```diff
@@ -42,11 +47,17 @@ func DefaultConfig() Config {
...
+		ConcurrencyLimit:        16,
 	}
 }
 
 // Validate performs basic validation of the config.
 func (cfg *Config) Validate() error {
+	if cfg.ConcurrencyLimit < core.DefaultConcurrencyLimit {
+		// override anything below the default limit
+		cfg.ConcurrencyLimit = core.DefaultConcurrencyLimit
+	}
```
Wondertan: This is shady. We shouldn't silently override the configurations users explicitly set to be non-default. Like, if I set it 1 to like save resources, it would still blow up. 

We should error out if we don't allow setting it below the default, or only dissalow setting 0.

renaynay: @Wondertan I agree it's "shady" to override but it's not breaking out current behaviour. Currently we just spin up 16 routines and our resource requirements for BNs are sufficient to cover that. Are you saying we should allow less now?

Wondertan: No, I don't mind restricting it to be more than 16 only if you think this is better, but by erroring out and not overriding

renaynay: Sure 🤝

PR #4608, opened 2025-10-01
fix(state/core_accessor): fix gas estimation logic

vgonkivs:
fix panic in estimateGasForBlobs

### DIFF
--- a/state/core_access.go
+++ b/state/core_access.go
@@ -32,6 +32,7 @@ import (
 	apperrors "github.com/celestiaorg/celestia-app/v6/app/errors"
 	"github.com/celestiaorg/celestia-app/v6/pkg/appconsts"
 	"github.com/celestiaorg/celestia-app/v6/pkg/user"
+	apptypes "github.com/celestiaorg/celestia-app/v6/x/blob/types"
 	libhead "github.com/celestiaorg/go-header"
 	libshare "github.com/celestiaorg/go-square/v3/share"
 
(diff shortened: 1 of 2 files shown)


state/estimate.go
```diff
@@ -66,9 +66,6 @@ func (ca *CoreAccessor) estimateGasPriceAndUsage(
 }
 
 // estimateGasForBlobs returns a gas limit that can be applied to the `MsgPayForBlob` transactions.
-func (ca *CoreAccessor) estimateGasForBlobs(blobSizes []uint32, shareVersion uint32) uint64 {
-	return apptypes.DefaultEstimateGas(&apptypes.MsgPayForBlobs{
-		BlobSizes:     blobSizes,
-		ShareVersions: []uint32{shareVersion},
-	})
+func (ca *CoreAccessor) estimateGasForBlobs(msg *apptypes.MsgPayForBlobs) uint64 {
```
vgonkivs: intentionally left it as a separate method

state/core_access.go
```diff
@@ -203,6 +195,15 @@ func (ca *CoreAccessor) SubmitPayForBlob(
 		return nil, fmt.Errorf("account for signer %s not found", author)
 	}
 
+	gas := cfg.GasLimit()
+	if gas == 0 {
+		msg, err := apptypes.NewMsgPayForBlobs(author.String(), appconsts.Version, libBlobs...)
```
renaynay: you don't need to pass app const version here, value is ignored anyway in function (❓ ) but passing this would just make it confusing.

state/core_access.go
```diff
@@ -203,6 +195,15 @@ func (ca *CoreAccessor) SubmitPayForBlob(
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to create msg pay-for-blobs: %w", err)
+		}
+		gas = ca.estimateGasForBlobs(msg)
```
renaynay: it would be nice if our utility method would. just accept the author + blobs and do the msg construction there instead of inline here as this method is already super long.

state/core_access.go
vgonkivs: I've checked the code and the app team is passing Version or `0`. I thought `0` is more confusing.

renaynay: wtf lol

renaynay (approved):
kk

PR #4637, opened 2025-10-14
refactor(blob): refactor blob's commitment proof verification method

vgonkivs:
Refactored `commitmentProof.Verify` method:
1) removed subtree root threshold from the input params(the node is using a const threshold in all places of the code)
2) added commitment to the input params
3) remove bool value from the returned values

### DIFF
--- a/blob/service_test.go
+++ b/blob/service_test.go
@@ -19,14 +19,9 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
-	"github.com/celestiaorg/celestia-app/v6/pkg/appconsts"
-	pkgproof "github.com/celestiaorg/celestia-app/v6/pkg/proof"
 	"github.com/celestiaorg/celestia-app/v6/pkg/wrapper"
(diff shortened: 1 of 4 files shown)


blob/service_test.go
```diff
@@ -985,111 +985,21 @@ func proveAndVerifyShareCommitments(t *testing.T, blobSize int) {
...
-			actualCommitment, _ := merkle.ProofsFromByteSlices(actualCommitmentProof.SubtreeRoots)
-			assert.Equal(t, msg.ShareCommitments[0], actualCommitment)
+			require.Error(t, err)
 		})
 	}
 }
 
-// generateCommitmentProofFromBlock takes a block and a PFB index and generates the commitment proof
-// using the traditional way of doing, instead of using the API.
-func generateCommitmentProofFromBlock(
```
renaynay, line 1017: why do we no longer need this test?

vgonkivs: it was not a test but another "trusted" way of generating the proof in test, so we can compare proofs. But we don't need it anymore.

PR #4639, opened 2025-10-16
WIP - parallel tx workers tastora test

renaynay:


nodebuilder/tests/tastora/framework.go
```diff
@@ -229,6 +229,13 @@ func (f *Framework) queryBalance(ctx context.Context, addr string) sdk.Coin {
 // newBridgeNode creates and starts a bridge node.
 func (f *Framework) newBridgeNode(ctx context.Context) *dataavailability.Node {
 	bridgeNode := f.startBridgeNode(ctx, f.celestia)
+
+	// Add a delay to allow the node to fully initialize, especially when using worker accounts
+	if f.config.TxWorkerAccounts > 0 {
+		f.t.Logf("Waiting for bridge node with %d worker accounts to initialize...", f.config.TxWorkerAccounts)
+		time.Sleep(10 * time.Second)
```
renaynay, line 204: this doesn't do anything

nodebuilder/tests/tastora/framework.go
```diff
@@ -246,9 +253,23 @@ func (f *Framework) fundWallet(ctx context.Context, fromWallet *types.Wallet, to
 	f.t.Logf("sending funds from %s to %s", fromAddr.String(), toAddr.String())
 
 	bankSend := banktypes.NewMsgSend(fromAddr, toAddr.Bytes(), sdk.NewCoins(sdk.NewCoin("utia", sdkmath.NewInt(amount))))
-	resp, err := f.celestia.BroadcastMessages(ctx, fromWallet, bankSend)
-	require.NoError(f.t, err)
-	require.Equal(f.t, resp.Code, uint32(0), "resp: %v", resp)
+
+	// Retry transaction broadcasting with exponential backoff
+	var resp sdk.TxResponse
+	broadcastRetries := 3
```
renaynay: general comment here is we should try to use TxClient inside of our framework here so we don't have to touch app code. Even if that means building a standalone txclient using the validator's signer.

PR #4649, opened 2025-10-21
test: framework improvements for robustness

gupadhyaya:
Improved:
* Robust transaction funding: Replaced unreliable block-based waiting with explicit transaction confirmation using waitForTransactionInclusion
* Simplified error handling: Streamlined fundWallet function to fail fast with clear error messages instead of silent retries
* Improved test stability: Fixed intermittent setup failures by ensuring funding transactions are confirmed before proceeding
* Better resource management: Enhanced cleanup and reduced unnecessary complexity in framework initialization

### DIFF
--- a/nodebuilder/tests/tastora/framework.go
+++ b/nodebuilder/tests/tastora/framework.go
@@ -39,14 +39,14 @@ const (
 	celestiaAppImage      = "ghcr.io/celestiaorg/celestia-app"
 	// defaultNodeTag can be overridden at build time using ldflags
 	// Example: go build -ldflags "-X github.com/celestiaorg/celestia-node/nodebuilder/tests/tastora.defaultNodeTag=v1.2.3"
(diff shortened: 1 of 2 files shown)


nodebuilder/tests/tastora/framework.go
```diff
@@ -647,8 +618,54 @@ func (f *Framework) Stop(ctx context.Context) error {
...
 
+	// Use Docker's built-in cleanup commands for comprehensive cleanup
+	f.dockerCleanup()
+
 	f.logger.Info("Framework cleanup completed")
 	return nil
 }
+
+// dockerCleanup uses Docker's built-in commands to clean up unused resources
+func (f *Framework) dockerCleanup() {
```
renaynay, line 640: ideally tastora framework would be able to provide something like `CleanupDAContainers` but i guess this is fine for now.

PR #4650, opened 2025-10-22
chore: minor improvement for docs

spuradage:
minor improvement for docs

### DIFF
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -17,8 +17,8 @@ var errEmptyShares = errors.New("empty shares")
 
 var subtreeRootThreshold = appconsts.SubtreeRootThreshold
 
-// The Proof is a set of nmt proofs that can be verified only through
-// the included method (due to limitation of the nmt https://github.com/celestiaorg/nmt/issues/218).
+// Proof is a set of NMT proofs that can be verified only through
+// the Included method (due to limitation of the NMT https://github.com/celestiaorg/nmt/issues/218).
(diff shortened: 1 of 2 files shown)


blob/blob.go
```diff
@@ -17,8 +17,8 @@ var errEmptyShares = errors.New("empty shares")
 
 var subtreeRootThreshold = appconsts.SubtreeRootThreshold
 
-// The Proof is a set of nmt proofs that can be verified only through
-// the included method (due to limitation of the nmt https://github.com/celestiaorg/nmt/issues/218).
+// Proof is a set of NMT proofs that can be verified only through
+// the Included method (due to limitation of the NMT https://github.com/celestiaorg/nmt/issues/218).
```
rootulp, line 21: TIL `Included` is a method: https://github.com/rootulp/celestia-node/blob/e6a940b6c222a9f73590462329fa4a1a3c1e2d3e/blob/service.go#L299

rootulp (approved):
LGTM but this PR needs reviews from the celestia-node subteam

blob/blob.go
spuradage: Thanks for your reply and review.  Yes. Included here is a specific method.

spuradage:
> LGTM but this PR needs reviews from the celestia-node subteam

@walldiss @Wondertan Hello, would you mind reviewing this PR when you have a moment? Thanks.

PR #4651, opened 2025-10-22
refactor(core): Better metrics for exchange and add metric to fetcher

renaynay:
Adds to exchange:

	downloadDuration        metric.Float64Histogram
	edsConstructionDuration metric.Float64Histogram
	edsStorageDuration      metric.Float64Histogram
	
Adds to fetcher: 

 	receiveBlockDuration metric.Float64Histogram

### DIFF
--- a/api/client/grpc.go
+++ b/api/client/grpc.go
@@ -3,6 +3,7 @@ package client
 import (
 	"context"
 	"crypto/tls"
+	"net"
 	"time"
 
 	grpc_retry "github.com/grpc-ecosystem/go-grpc-middleware/retry"
@@ -27,9 +28,9 @@ type CoreGRPCConfig struct {
 	AuthToken string
 }
 
-// Validate performs basic validation of the config.
 func (cfg *CoreGRPCConfig) Validate() error {
-	_, err := utils.SanitizeAddr(cfg.Addr)
+	normalized := utils.NormalizeGRPCAddress(cfg.Addr)
+	_, _, err := net.SplitHostPort(normalized)
 	return err
 }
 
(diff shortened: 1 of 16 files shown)


core/fetcher.go
```diff
@@ -251,10 +255,14 @@ func receiveBlockByHeight(streamer coregrpc.BlockAPI_BlockByHeightClient) (
 		parts = append(parts, resp.BlockPart)
 		isLast = resp.IsLast
 	}
+
+	f.metrics.observeReceiveBlock(ctx, time.Since(start), len(parts))
```
renaynay: BTW - the majority of download time (exchange metric) is actually spent here so I'm not 100% sure we need both exchange block download time and fetcher receive block time.

core/fetcher_metrics.go
```diff
@@ -0,0 +1,54 @@
...
+			metric.WithAttributes(numPartsAttribute(numParts)))
+	})
+}
+
+// numPartsAttribute creates an attribute for the number of parts in a block.
+// This can be used to approximate the block size (with each part being roughly
+// 64KB).
+func numPartsAttribute(numParts int) attribute.KeyValue {
+	return attribute.Int("num_parts", numParts)
+}
```
walldiss: Seems a bit too much to have a function for that. It adds extra nesting

renaynay: Oh it's a relic where i was actually trying to do some math there to reduce cardinality but it's fine to just do numParts  as int attribute i think.

core/fetcher.go
renaynay: it would be good to keep it actually to observe departure in case it happens.

walldiss (commented):
Given that otel does no support proper buckets setup, those might not bring as much value as expected. Like all time events might get recorded as same bucket. I would suggest to try to add traces for same events and compare avarages, to verify that metrics are accurate here

renaynay:
@walldiss 
Why wouldn't you be able to use a query like
```
rate(core_ex_total_blocks_processed{eds_size="64"}[1m])
```

to view metrics by `eds_size` ? I'm not sure what you mean by otel doesn't support proper buckets setup.

PR #4658, opened 2025-10-24
chore(share/shwap/p2p/shrex): Add shrex server successful payload served metric

renaynay:
This metric will be able to help us track BN shrex serving throughput by request type (`nd`, `eds`, and soon will be sample).

In grafana, we can visualise serving throughput in MB/s with a panel via the following query: 

`shrex-nd`
```
rate(shrex_payload_served_bytes_sum{protocol="nd_v0"}[1m]) / 1024 / 1024
```

`shrex-eds`
```
rate(shrex_payload_served_bytes_sum{protocol="nd_v0"}[1m]) / 1024 / 1024
```

### DIFF
--- a/share/shwap/p2p/shrex/metrics.go
+++ b/share/shwap/p2p/shrex/metrics.go
@@ -88,6 +88,14 @@ func (m *Metrics) observePayloadServed(
 }
 
 func InitClientMetrics() (*Metrics, error) {
+	totalRequestCounter, err := meter.Int64Counter(
+		"shrex_server_total_responses",
+		metric.WithDescription("Total count of sent shrex responses"),
+	)
+	if err != nil {
+		return nil, err
+	}
+
 	requestDuration, err := meter.Float64Histogram(
 		"shrex_client_request_duration",
 		metric.WithDescription("Time taken to complete a shrex client request"),
@@ -97,7 +105,8 @@ func InitClientMetrics() (*Metrics, error) {
 	}
 
 	return &Metrics{
-		requestDuration: requestDuration,
+		requestDuration:     requestDuration,
+		totalRequestCounter: totalRequestCounter,
 	}, nil
 }
 


share/shwap/p2p/shrex/metrics.go
```diff
@@ -39,11 +39,14 @@ type Metrics struct {
 	totalRequestCounter metric.Int64Counter
 	rateLimiterCounter  metric.Int64Counter
 	requestDuration     metric.Float64Histogram
+	// payloadServed will aggregate the total payload served
+	// by the shrex server
+	payloadServed metric.Int64Histogram
```
walldiss: Should be Counter of bytes.

share/shwap/p2p/shrex/server.go
```diff
@@ -104,10 +104,13 @@ func (srv *Server) WithMetrics() error {
...
-		handleTime := time.Now()
-		status := srv.handleDataRequest(ctx, requestID, s)
-		srv.metrics.observeRequests(ctx, 1, requestID.Name(), status, time.Since(handleTime))
+		requestID, handleTime := id(), time.Now()
+
+		status, size := srv.handleDataRequest(ctx, requestID, s)
+
+		srv.metrics.observeRequest(ctx, 1, requestID.Name(), status, time.Since(handleTime))
+		srv.metrics.observePayloadServed(ctx, requestID.Name(), status, size)
+
```
walldiss: Can combine tracking of requests / sec with bytes/ sec. 

Some side refactoring question. Not sure why count is sent to `observeRequest` function, to me it seems it will always should be 1 and we can remove it.

share/shwap/p2p/shrex/server.go
```diff
@@ -170,32 +175,32 @@ func (srv *Server) handleDataRequest(ctx context.Context, requestID request, str
...
-		return status
+		return status, 0
 	}
 
-	_, err = io.Copy(stream, r)
+	written, err := io.Copy(stream, r)
 	if err != nil {
 		logger.Errorw("send data", "err", err)
-		return statusSendRespErr
+		return statusSendRespErr, 0
```
walldiss: Could be partial write, with some bytes written

share/shwap/p2p/shrex/server.go
```diff
@@ -170,32 +175,32 @@ func (srv *Server) handleDataRequest(ctx context.Context, requestID request, str
...
 
 	defer utils.CloseAndLog(log, "file", file)
 	r, err := requestID.ResponseReader(ctx, file)
 	if err != nil {
 		logger.Errorf("getting data from response reader %w", err)
-		return respondStatus(logger, shrexpb.Status_INTERNAL, stream)
+		return respondStatus(logger, shrexpb.Status_INTERNAL, stream), 0
 	}
 
 	status := respondStatus(logger, shrexpb.Status_OK, stream)
```
walldiss: We should count respond status bytes too. In respondStatus func can return bytes written as well

share/shwap/p2p/shrex/server.go
renaynay: @walldiss count is a relic. I wanted to refactor things in here too but i decided against doing it in this PR, but can do it anyway. 

> Can combine tracking of requests / sec with bytes/ sec.

I want this to be a server-side metric only and this metric was for both client + server. So I would be against combining it (as it's nit anyway)

share/shwap/p2p/shrex/server.go
renaynay: I only want to track successful / completed responses in this metric as to not pollute throughput metric with incomplete requests. 

We could maybe track this some other way if we find it important? Or do you think this partial write should count towards throughput somehow?

share/shwap/p2p/shrex/server.go
renaynay: @walldiss we know how much respondStatus bytes are (1 byte), so I feel it's kind of negligible to add this byte. If we have a successful request, we can assume respondStatus adds 1 byte to overall payload. If you feel strongly, we can add this.

share/shwap/p2p/shrex/metrics.go
renaynay: Histogram does the same thing as counter would, since it provides metric as `sum`. Both do the same thing, it's just about how to build the metricsql query such that it displays in the way we want to visualise the data. 

If you feel strongly, I can change to counter + change the query.

walldiss: We just don't need hist for this

share/shwap/p2p/shrex/server.go
walldiss: make me wonder why client and server are combined in the first place tbh. Could be refactored too (optionally in this PR)

share/shwap/p2p/shrex/server.go
walldiss: WE should track all bandwidth or it will be quite tricky since we won't see half of the picture. Metric can have additional label for error: true/false. Should be similar to req/sec counter

share/shwap/p2p/shrex/server.go
walldiss: It would give us better picture to see consumption of whole protocol without doing calculations in the head. I think we should count all bytes served

renaynay: https://github.com/celestiaorg/celestia-node/pull/4658/commits/be8f95dd91affe3d6c01590ac122279da66d8884

share/shwap/p2p/shrex/server.go
renaynay: https://github.com/celestiaorg/celestia-node/pull/4658/commits/be8f95dd91affe3d6c01590ac122279da66d8884

share/shwap/p2p/shrex/server.go
renaynay: Did small refactor but don't want to scope creep it https://github.com/celestiaorg/celestia-node/pull/4658/commits/450899dc57620f2c5cdbd85eee45694223dfbf1f

share/shwap/p2p/shrex/metrics.go
renaynay: https://github.com/celestiaorg/celestia-node/commit/be8f95dd91affe3d6c01590ac122279da66d8884

share/shwap/p2p/shrex/metrics.go
```diff
@@ -36,66 +36,71 @@ const (
 )
 
 type Metrics struct {
-	totalRequestCounter metric.Int64Counter
-	rateLimiterCounter  metric.Int64Counter
-	requestDuration     metric.Float64Histogram
+	rateLimiterCounter metric.Int64Counter
```
walldiss: wasn't rate-limiter removed?

share/shwap/p2p/shrex/server.go
```diff
@@ -183,35 +188,36 @@ func (srv *Server) handleDataRequest(ctx context.Context, requestID request, str
...
-		return status
+		return status, 0
 	}
 
-	_, err = io.Copy(stream, r)
+	writtenResponse, err := io.Copy(stream, r)
 	if err != nil {
 		logger.Errorw("send data", "err", err)
-		return statusSendRespErr
+		return statusSendRespErr, 0
```
walldiss: Metrics It is still reporting 0 for written status and in case of partial write, which is not fully representing served bytes

share/shwap/p2p/shrex/metrics.go
renaynay: needs rebase yea

share/shwap/p2p/shrex/server.go
```diff
@@ -174,35 +179,36 @@ func (srv *Server) handleDataRequest(ctx context.Context, requestID request, str
...
 
-func respondStatus(log *zap.SugaredLogger, status shrexpb.Status, stream network.Stream) status {
-	_, err := serde.Write(stream, &shrexpb.Response{Status: status})
+// respondStatus returns the status written to stream and the size of the response.
+func respondStatus(log *zap.SugaredLogger, status shrexpb.Status, stream network.Stream) (status, int) {
+	written, err := serde.Write(stream, &shrexpb.Response{Status: status})
 	if err != nil {
 		log.Errorw("sending response status", "err", err)
-		return statusSendStatusErr
+		return statusSendStatusErr, 0
```
walldiss: return written on error too

renaynay: done https://github.com/celestiaorg/celestia-node/pull/4658/commits/d827ff5d3f8b3b8bfabb8db92b691bcba868ab4b

share/shwap/p2p/shrex/metrics.go
```diff
@@ -37,41 +37,55 @@ const (
...
 	if m == nil {
 		return
 	}
+
 	ctx = utils.ResetContextOnError(ctx)
-	m.totalRequestCounter.Add(ctx, count,
-		metric.WithAttributes(
-			attribute.String("protocol", requestName),
-			attribute.String("status", string(status)),
-		))
```
vgonkivs, line 59: Can we return it back?

renaynay: Oh yes good catch

vgonkivs (dismissed):

PR #4663, opened 2025-10-27
fix(ci): Fix unit flakes from `testnode` by stopping calling the stop functions multiple times

evan-forbes:
trying a hack in CI

### DIFF
--- a/nodebuilder/tests/helpers_test.go
+++ b/nodebuilder/tests/helpers_test.go
@@ -26,8 +26,9 @@ func getAdminClient(ctx context.Context, nd *nodebuilder.Node, t *testing.T) *cl
 	jwt, err := authtoken.NewSignedJWT(signer, []auth.Permission{"public", "read", "write", "admin"}, time.Minute)
 	require.NoError(t, err)
 
-	client, err := client.NewClient(ctx, listenAddr, jwt)
+	client, err := client.NewClient(context.WithoutCancel(ctx), listenAddr, jwt)
 	require.NoError(t, err)
+	t.Cleanup(client.Close)
 
 	return client
 }


core/testing.go
```diff
@@ -117,6 +119,9 @@ type Network struct {
 	stopNode func() error
 	stopGRPC func() error
 	stopAPI  func() error
+
+	stopOnce sync.Once
+	stopErr  error
 }
```
evan-forbes, line 125: this is a hack just to demonstrate the fix

a better solution is to not wrap testnode and rely on it adding to the cleanup functions to the `testing.T` we pass it. if you all need new functionality happy to prioritize adding it to testnode so you all don't need this wrapper.

Its unclear to me what that functionality is however I did see this issue https://github.com/celestiaorg/celestia-app/issues/4304

share/eds/retriever_test.go
```diff
@@ -44,17 +44,16 @@ func TestRetriever_Retrieve(t *testing.T) {
 		{"128x128(max)", share.MaxSquareSize},
 	}
 	for _, tc := range tests {
+		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
 			// generate EDS
 			shares, err := libshare.RandShares(tc.squareSize * tc.squareSize)
 			require.NoError(t, err)
+			ctx, cancel := context.WithTimeout(baseCtx, time.Minute*5) // generous timeout for large squares
+			t.Cleanup(cancel)
```
evan-forbes, line 54: this test isn't related to testnode afaiu, but trying this fix as well to

1) be helpful 2) get more runs of the other fix

walldiss (approved):

PR #4664, opened 2025-10-28
feat(blob|state): Blob submission metrics

gupadhyaya:
Copies https://github.com/celestiaorg/celestia-node/pull/4549 as previous one is unable to merge due to fork docker issue. No review required, just approvals

### DIFF
--- a/pruner/metrics.go
+++ b/pruner/metrics.go
@@ -39,8 +39,12 @@ func (s *Service) WithMetrics() error {
 	}
 
 	callback := func(_ context.Context, observer metric.Observer) error {
-		observer.ObserveInt64(lastPruned, int64(s.checkpoint.LastPrunedHeight))
-		observer.ObserveInt64(failedPrunes, int64(len(s.checkpoint.FailedHeaders)))
+		s.checkpointMu.Lock()
+		if s.checkpoint != nil {
+			observer.ObserveInt64(lastPruned, int64(s.checkpoint.LastPrunedHeight))
+			observer.ObserveInt64(failedPrunes, int64(len(s.checkpoint.FailedHeaders)))
+		}
+		s.checkpointMu.Unlock()
 		return nil
 	}
 


blob/metrics.go
```diff
@@ -0,0 +1,192 @@
...
+	errorTypeTimeout  = "timeout"
+	errorTypeCanceled = "canceled"
+	errorTypeUnknown  = "unknown"
+)
+
+// metrics tracks blob-related metrics
+type metrics struct {
+	// Retrieval metrics
+	retrievalCounter  metric.Int64Counter
+	retrievalDuration metric.Float64Histogram
```
renaynay, line 26: Histograms include a count, you can derive count from this metric, no need for duplicate Counter.

Remove retrievalCounter and keep retrievalDuration

blob/metrics.go
```diff
@@ -0,0 +1,192 @@
...
+
+	// Proof metrics
+	proofCounter  metric.Int64Counter
+	proofDuration metric.Float64Histogram
```
renaynay, line 30: same

blob/metrics.go
```diff
@@ -0,0 +1,192 @@
...
+
+	// Internal counters (thread-safe)
+	totalRetrievals      atomic.Int64
```
renaynay: add err as attribute, no need to count errors in sep counter.

blob/metrics.go
```diff
@@ -0,0 +1,192 @@
...
+	totalRetrievalErrors atomic.Int64
+	totalProofs          atomic.Int64
+	totalProofErrors     atomic.Int64
```
renaynay: same

blob/service.go
```diff
@@ -222,9 +235,12 @@ func (s *Service) GetProof(
 	namespace libshare.Namespace,
 	commitment Commitment,
 ) (proof *Proof, err error) {
+	start := time.Now()
 	ctx, span := tracer.Start(ctx, "get-proof")
 	defer func() {
 		utils.SetStatusAndEnd(span, err)
+		// Record metrics
```
renaynay: nit: unnecessary comments left in code, pls remove these

blob/metrics.go
```diff
@@ -0,0 +1,192 @@
...
+// Stop cleans up metrics resources
+func (m *metrics) Stop() error {
+	if m == nil || m.clientReg == nil {
+		return nil
+	}
+	return m.clientReg.Unregister()
+}
+
+// ObserveRetrieval records blob retrieval metrics
+func (m *metrics) ObserveRetrieval(ctx context.Context, duration time.Duration, err error) {
```
renaynay: please make all metrics' methods private if they are not used outside of pkg. same for state/metrics

blob/metrics.go
```diff
@@ -119,25 +97,22 @@ func (s *Service) WithMetrics() error {
...
-// ObserveRetrieval records blob retrieval metrics
-func (m *metrics) ObserveRetrieval(ctx context.Context, duration time.Duration, err error) {
+// observeRetrieval records blob retrieval metrics
+func (m *metrics) observeRetrieval(ctx context.Context, duration time.Duration, err error) {
 	if m == nil {
 		return
 	}
 
-	// Update counters
+	// Update counter
```
renaynay: 😭

blob/metrics.go
```diff
@@ -155,24 +130,18 @@ func (m *metrics) ObserveRetrieval(ctx context.Context, duration time.Duration,
...
-// ObserveProof records blob proof metrics
-func (m *metrics) ObserveProof(ctx context.Context, duration time.Duration, err error) {
+// observeProof records blob proof metrics
+func (m *metrics) observeProof(ctx context.Context, duration time.Duration, err error) {
 	if m == nil {
 		return
 	}
 
-	// Update counters
+	// Update counter
```
renaynay: 😭

renaynay (dismissed):

blob/metrics.go
```diff
@@ -0,0 +1,157 @@
...
+	)
+	if err != nil {
+		return err
+	}
+
+	callback := func(_ context.Context, observer metric.Observer) error {
+		observer.ObserveInt64(retrievalTotal, m.totalRetrievals.Load())
+		observer.ObserveInt64(proofTotal, m.totalProofs.Load())
+		return nil
+	}
```
walldiss: Normal counter would be much better here than async + atomic int. Current logic has flaw of recounting older retrievals and proofs in subsequent callback calls

state/metrics.go
```diff
@@ -2,41 +2,285 @@ package state
...
+	if m == nil {
+		return
+	}
+
+	m.totalPfbSubmissions.Add(1)
+
+	attrs := []attribute.KeyValue{
+		attribute.Int(attrBlobCount, blobCount),
+		attribute.Int64(attrTotalSize, totalSize),
+		attribute.Float64(attrGasPrice, gasPrice),
```
walldiss: None of those can be an attribute or we will have high cardinality problems. I would suggest to
- track each blob in a counter as 1. 
- Blob sizes can be tracked as another counter (bytes_uploaded)
- gas price can be another counter (gas_spent)

state/metrics.go
```diff
@@ -2,41 +2,277 @@ package state
...
+	pfbSubmissionBlobCount   metric.Int64Counter
+	pfbSubmissionBlobSize    metric.Int64Counter
+	pfbGasEstimationDuration metric.Float64Histogram
+	pfbGasPriceEstimation    metric.Float64Histogram
+	pfbSubmissionTotal       metric.Int64Counter
+
+	// Gas estimation metrics
+	gasEstimationDuration metric.Float64Histogram
+	gasEstimationTotal    metric.Int64Counter
+
```
walldiss, line 40: What is the diff between pfbGasEstimationDuration and gasEstimationDuration?

gupadhyaya: pfb vs any other tx

walldiss (approved):

PR #4665, opened 2025-10-28
Shrex client server specification

vgonkivs:
[Rendered](https://github.com/vgonkivs/celestia-node/blob/shrex_client_server/specs/src/shrex/client-server.md)

### DIFF
--- a/specs/src/shrex/peer_manager.md
+++ b/specs/src/shrex/peer_manager.md
@@ -1,0 +1,208 @@
+# Peer Manager Specification for SHREX Protocol
+
+## Abstract
+
+This specification defines the Peer Manager component for the SHREX protocol in the Celestia network. The Peer Manager collects, organizes, and provides peers for efficient data retrieval based on data availability notifications and peer discovery.
+
+## Overview
+
+The Peer Manager serves as the central coordination point for peer selection in the SHREX protocol. It aggregates peers from two primary sources:
+
+1. **ShrEx/Sub notifications**: Peers announcing specific data availability through pubsub
+2. **Discovery service**: Peers found through DHT-based discovery
+
+The manager maintains hash pools by creating a pool for each data hash when first announced via ShrEx/Sub, then validates these pools by matching them against headers received from its header subscription (representing the node's subjective view of the chain). Once a pool is validated through header matching, all peers in that pool are promoted to the discovered nodes pool. Pools that are not validated within a timeout are garbage collected. This validation mechanism ensures the manager provides the best request routing possible
+
+## Flowchart: Adding Peers to the Manager
+
+```text
+                    ┌─────────────────────────────┐
+                    │   Peer Announcement Source  │
+                    └──────────┬──────────────────┘
+                               │
+              ┌────────────────┴────────────────┐
+              │                                 │
+              ▼                                 ▼
+   ┌─────────────────────┐         ┌─────────────────────┐
+   │  ShrEx/Sub Protocol │         │  Discovery Service  │
+   │  Notification        │        │  Event              │
+   └──────────┬───────────┘        └──────────┬──────────┘
+              │                               │
+              │                               │
+              ▼                               ▼
+   ┌─────────────────────┐         ┌─────────────────────┐
+   │  Validator Function │         │  Add directly to    │
+   │  - Self check       │         │  Discovered Nodes   │
+   │  - Deduplication    │         │  Pool               │
+   │  - Filtering        │         └─────────────────────┘
+   └──────────┬───────────┘
+              │
+              ▼
+      ┌───────────────┐
+      │  Valid?       │
+      └───┬───────────┘
+          │
+    ┌─────┴─────┐
+    │           │
+   No          Yes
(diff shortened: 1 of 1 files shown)


specs/src/shrex/client-server.md
```diff
@@ -0,0 +1,142 @@
...
+```
+
+## Server
+
+- Accepting incoming connections
+- Reading and validating requests
+- Retrieving data from local storage
+- Sending responses back to clients
+
+### Request Handling
```
renaynay, line 188: in certain cases (ND), not found comes with data too (non incl proof)

specs/src/shrex/client-server.md
```diff
@@ -0,0 +1,142 @@
...
+```
+
+## Response Format
+
+All server responses follow a two-phase pattern:
+
+1. **Status Message**: Indicates whether the request can be fulfilled
+    - **OK**: Data is available and will follow
+    - **READ REQUEST ERROR** - error happened during reading a request from the stream
+    - **SEND STATUS ERROR** - errod during sending a status back the client
```
renaynay:
```suggestion
    - **SEND STATUS ERROR** - error while sending a status back the client
```

specs/src/shrex/client-server.md
```diff
@@ -0,0 +1,142 @@
...
+    - **SEND RESPONSE ERROR** - errod during sending a status back the client
```
renaynay:
```suggestion
    - **SEND RESPONSE ERROR** - error during sending a status back the client
```

specs/src/shrex/client-server.md
```diff
@@ -0,0 +1,142 @@
...
+    - **BAD REQUEST** - request did not pass validation
+    - **NOT FOUND**: Requested data is not available
+    - **INTERNAL ERROR**: Server encountered an error
+
+2. **Data (if OK)**: The actual requested data
```
renaynay: also include non-incl proof if ND not found

specs/src/shrex/client-server.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Error Handling
+
+### Client Errors
+
+The client reports errors to its caller:
+
+- Connection failures
+- Timeouts
+- Invalid data
```
renaynay: meaning invalid data sent back from server?

vgonkivs: relic

vgonkivs:
> reference to shwap spec defining encoding and payload of requests and responses

I haven't added this spec yet. My idea was to add it when I will be merging all specs into a single. 

> peer manager integration, reference to peer manager spec

Client-Server does not interact with peer manager, but Getter does.


I will update spec with 
```
transport protocol spec (libp2p streaming)
message flow. Status message definition and proto. Error model in status message
```

Additionally, I can add a list of available endpoints

specs/src/shrex/client-server.md
```diff
@@ -0,0 +1,388 @@
...
+- **Response Messages**: Protocol Buffers (protobuf) encoded
+
+### Request Messages
+
+Request messages are binary encoded.
+
+The request is serialized to binary format and written directly to the stream. Each request type includes:
+
+- Height: Block height for the requested data
+- Type-specific parameters
```
walldiss, line 62: Requests use shwap id and associated encoding. We can be more specific here and link to the shwap spec

specs/src/shrex/client-server.md
```diff
@@ -0,0 +1,388 @@
...
+   INVALID = 0;      // Invalid/unknown status
+   OK = 1;           // Data found and will be sent
+   NOT_FOUND = 2;    // Requested data not found
+   INTERNAL = 3;     // Internal server error
+}
+
+message Response {
+   Status status = 1;
+}
+```
```
walldiss, line 81: After handling requests server must send Status message indicating result of handling.

specs/src/shrex/client-server.md
```diff
@@ -0,0 +1,388 @@
...
+message Share {
+   bytes data = 1;
+}
+
+enum AxisType {
+   ROW = 0;
+   COL = 1;
+}
+```
+
```
walldiss: This should list shwap IDs and associated shawp.Containers. Encoding is defined in shwap spec, so no need to make copy of it here. But make sure to make to link them here. Formatting as table might look better:
| protocol-id | request (shwap.id) | response payload (shwap.container)

specs/src/shrex/client-server.md
walldiss: Actually false. Server sends "OK" along with non-inclusion proof

specs/src/shrex/client-server.md
vgonkivs: added

specs/src/shrex/client-server.md
vgonkivs: added

specs/src/shrex/client-server.md
vgonkivs: reworked

Issue #4675, opened 2025-10-30
[Feature Request]: Allow both `x-token` and `xtoken` key names (and filenames) for authentication token — to prevent breaking integrations and external docs


jcstein:
### **Description**

Running a Celestia light node with QuickNode (or any endpoint that requires auth) is confusing and inconsistent when it comes to how the authentication token is handled.

The [Celestia docs](https://docs.celestia.org/how-to-guides/light-node#optional-start-light-node-with-consensus-node-endpoint-with-authentication), the [QuickNode guide](https://www.quicknode.com/guides/infrastructure/node-setup/run-a-celestia-light-node), and the actual node implementation **used to all disagree** — making it look like something was broken when it was actually just naming inconsistency.

Specifically:

* The node expects the JSON key to be `"x-token"`, but older versions of Quicknode's guides used `"token"`. When I wrote the Celestia guide, I expected `xtoken`.
* The node expects a file named `x-token.json`, but the CLI (`--core.xtoken.path`)existing setups (and muscle memory from this CLI) use `xtoken.json`.
* The error message (`x-token is empty`) gives zero hint that the problem is just inconsistent naming.

QuickNode **followed normal conventions** and wrote their guide accordingly — only to find out that our implementation was the one that didn’t make sense. They’ve since fixed their guide, but the real problem remains: the node itself shouldn’t be this brittle.

---

### **Steps to Reproduce**

Follow the QuickNode guide (before it was fixed):
[https://www.quicknode.com/guides/infrastructure/node-setup/run-a-celestia-light-node](https://www.quicknode.com/guides/infrastructure/node-setup/run-a-celestia-light-node)

```bash
cat ~/.celestia-auth/xtoken.json
{
  "token": "hunter2"
}

celestia light start \
  --core.ip a-b-c.celestia-mocha.quiknode.pro \
  --core.port 9090 \
  --core.tls \
  --core.xtoken.path /home/ubuntu/.celestia-auth/ \
  --p2p.network mocha
```

#### **Result**

```
Error: could not build *grpc.ClientConn: x-token is empty. Please setup a token or cleanup xtokenPath
```

#### **Expected Behavior**

It should just work if there’s a valid token — regardless of whether it’s named `x-token` or `xtoken`, or whether the file is `x-token.json` or `xtoken.json`.

---

### **Actual Behavior**

Fails with a useless error message about an “empty x-token,” when the real issue is just a naming mismatch between our implementation and what most developers would assume from the docs or guides.

[...]

PR #4676, opened 2025-10-31
fix(nodebuilder/core): xtoken discrepancies

gupadhyaya:
Fixes [ #4676](https://github.com/celestiaorg/celestia-node/issues/4675)

### DIFF
--- a/nodebuilder/core/constructors.go
+++ b/nodebuilder/core/constructors.go
@@ -191,7 +191,8 @@ func parseTokenPath(xtokenPath string) (string, error) {
 		if val, ok := tokenData[key]; ok && val != "" {
 			tokenValue = val
 			if i > 0 {
-				log.Warnf("Using alternate JSON key '%s' in file '%s'. Consider using '%s' for consistency.", key, tokenFilePath, xtokenJSONKeys[0])
+				log.Warnf("Using alternate JSON key '%s' in file '%s'. Consider using '%s' for consistency.",
+					key, tokenFilePath, xtokenJSONKeys[0])
 			}
 			break
 		}


nodebuilder/core/constructors.go
```diff
@@ -22,6 +24,8 @@ import (
 	"github.com/celestiaorg/celestia-node/libs/utils"
 )
 
+var log = logging.Logger("core")
```
renaynay:
```suggestion
var log = logging.Logger("nodebuilder/core")
```

nodebuilder/core/constructors.go
```diff
@@ -32,7 +36,8 @@ const (
 	// TODO(@vgonkivs): Revisit this constant once the block size reaches 64MB.
 	defaultGRPCMessageSize = 64 * 1024 * 1024 // 64Mb
 
-	xtokenFileName = "xtoken.json"
+	xtokenFileName    = "xtoken.json"
```
renaynay: can u just array and loop?

nodebuilder/core/constructors.go
```diff
@@ -144,28 +149,68 @@ func authStreamInterceptor(xtoken string) grpc.StreamClientInterceptor {
...
+	token, err := os.ReadFile(tokenFilePath)
 	if err != nil {
-		return "", err
+		return "", fmt.Errorf("failed to read token file '%s': %w", tokenFilePath, err)
 	}
 
+	// Support "x-token" (preferred), "xtoken", and "token" JSON keys for maximum compatibility
 	auth := struct {
-		Token string `json:"x-token"`
+		XToken    string `json:"x-token"`
```
renaynay: this can be simplified

Just construct the auth token when you successfully read from one of the expected file names.

gupadhyaya: wdym? i am reading from the expected file names, but the token name can be either of these three (x-token | xtoken | token) hence the struct and unmarshal. can u clarify?

nodebuilder/core/constructors.go
gupadhyaya: why? there are only two file names: xtoken.json or x-token.json

gupadhyaya: using array now.

nodebuilder/core/constructors.go
gupadhyaya: changed this to unmarshall into a map on successful file read and then extract token.

gupadhyaya:
manually tested all combinations using quicknode and confirm that fix is working. @walldiss

PR #4680, opened 2025-11-04
test: cross version api compatibility test using tastroa

gupadhyaya:
## Summary

Adds bidirectional cross-version API compatibility tests using tastora to detect breaking API changes between versions.

## How It Works

1. **Current client → Old server**: Uses current codebase's RPC client to connect to an old server version (via Docker image)
2. **Old client → Current server**: Compiles client code from an old version inside Docker and connects to current server (built from current branch)

Both directions must pass to ensure API compatibility.

## Test Coverage

Tests all major API modules: Node, Header, State, P2P, Share, DAS, and Blob APIs across both bridge and light nodes.

## Changes

- Added cross-version compatibility test suite (`nodebuilder/tests/tastora/api/cross_version_client_test.go`)
- Enhanced tastora framework with version-aware node creation (`NewBridgeNodeWithVersion`, `NewLightNodeWithVersion`)
- Added `buildCurrentNodeImage()` to build Docker images from current codebase
- Added GitHub Actions workflow for CI integration
- Reverted local changes to `transaction_test.go` to match upstream

## Running Tests

```bash
cd nodebuilder/tests/tastora/api
go test -tags=integration -v -run TestCrossVersionClientTestSuite/TestCrossVersionBidirectional -timeout 30m
```


api/client/client_test.go
```diff
@@ -258,7 +258,9 @@ func TestSubmission_QueuedSubmission(t *testing.T) {
 	// Spam blobs from default key parallel
 	wg := sync.WaitGroup{}
 	for range 5 {
-		wg.Go(func() {
+		wg.Add(1)
```
renaynay: pls revert and use wg.Go

nodebuilder/p2p/p2p.go
```diff
@@ -210,13 +211,28 @@ func (m *module) BandwidthForProtocol(_ context.Context, proto protocol.ID) (met
...
+func (m *module) ResourceState(_ context.Context) (rcmgr.ResourceManagerStat, error) {
 	rms, ok := m.rm.(rcmgr.ResourceManagerState)
 	if !ok {
 		return rcmgr.ResourceManagerStat{}, fmt.Errorf("network.resourceManager does not implement " +
 			"rcmgr.ResourceManagerState")
 	}
-	return rms.Stat(), nil
+	stat := rms.Stat()
+
+	// Sanitize peer IDs to ensure backward compatibility with old clients that cannot decode certain peer ID formats.
```
renaynay, line 222: wait what happened here?

nodebuilder/tests/tastora/api/cross_version_client_test.go
```diff
@@ -0,0 +1,760 @@
...
+	if s.f != nil {
+		s.f.Stop(context.Background())
+	}
+}
+
+func (s *CrossVersionClientTestSuite) TestCrossVersionBidirectional() {
+	// Use a longer timeout to handle Docker builds and all test combinations
+	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Minute)
+	defer cancel()
+	oldVersion := "v0.28.3-arabica"
```
renaynay: It would be nice if we could grab this from somewhere (inject from CI for example) bc otherwise we'll have to hardcode it constantly. This can be a FLUP though - but pls track it.

nodebuilder/tests/tastora/api/cross_version_client_test.go
```diff
@@ -0,0 +1,760 @@
...
+		serverRPC := "http://" + rpcAddr
+
+		err = s.runClientTestFromDockerImage(testCtx, oldVersion, serverRPC, true)
+		require.NoError(s.T(), err)
+	})
+}
+
+// waitForNodeReadyAndSynced waits for a node to be fully ready and synced before testing APIs.
+// This ensures APIs will work without needing individual timeouts.
+func (s *CrossVersionClientTestSuite) waitForNodeReadyAndSynced(ctx context.Context, client *rpcclient.Client, nodeName string, timeout time.Duration) {
```
renaynay: this should be a method on the framework itself 

and it can definitely be cleaned up: instead of polling, just use `SyncWait`. 

Also shares available not a necessary pre-check for this test.

please also remove all the AI-generated comments from the code 😭

nodebuilder/tests/tastora/api/cross_version_client_test.go
```diff
@@ -0,0 +1,760 @@
...
+		s.T().Logf("Share.GetNamespaceData: data not available (known limitation with old light servers): %v", err)
+	} else {
+		require.NoError(s.T(), err)
+	}
+	_, err = client.Share.GetEDS(ctx, head.Height())
+	if err != nil && strings.Contains(err.Error(), "data not available") {
+		s.T().Logf("Share.GetEDS: data not available (known limitation with old light servers): %v", err)
+	} else {
+		require.NoError(s.T(), err)
+	}
```
renaynay: Here we should test for expected success case and failure case (errors expected for example)

instead of allowing silent failure.

nodebuilder/tests/tastora/api/cross_version_client_test.go
```diff
@@ -0,0 +1,760 @@
...
+
+	dasCtx, dasCancel := context.WithTimeout(ctx, 15*time.Second)
+	_, err = client.DAS.SamplingStats(dasCtx)
+	dasCancel()
+	if err != nil && !strings.Contains(err.Error(), "stubbed") && !strings.Contains(err.Error(), "deadline exceeded") {
+		require.NoError(s.T(), err)
+	}
+
+	waitCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	err = client.DAS.WaitCatchUp(waitCtx)
```
renaynay: IMO we don't need to test the wait methods bc they're just blocking methods that return err in case ctx deadline exceeded.

nodebuilder/tests/tastora/api/cross_version_client_test.go
```diff
@@ -0,0 +1,760 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to get/create Go module cache volume: %w", err)
+	}
+	gocacheVol, err := s.getOrCreateVolume(ctx, dockerClient, "tastora-gocache")
+	if err != nil {
+		return fmt.Errorf("failed to get/create Go build cache volume: %w", err)
+	}
+
+	tmpDir := s.T().TempDir()
+	testProgram := s.generateTestClientProgram(serverRPCAddr, skipGetRow)
```
renaynay: what's this?

renaynay (commented):
Few comments here. The code in cross_version_client is a bit confusing and i'd like to reduce it to the bare minimum necessary to run the tests. 

The tests definitely pass locally?

chatton (changes requested):
I'm mainly just reviewing the approach of dynamic code generation and image building in this PR in cross_version_client_test.go

If my understanding is correct, what we are attempting to achieve here is creating a set of tests that allows us to use **old** client code with a **new** server version, and the other way around.

I don't think the current approach is the way this problem should be solved.
- It adds a lot of docker logic that isn't really relevant for the scope of the test
- the dynamic go code is kind of hard to follow and we don't get any type safety with it.
- it's also a bit weird and unintuitive.

In my opinion, a cleaner approach would be to create a separate binary that is a simple CLI application that performs these checks. It would be built and tagged on every PR/tag. (This would need backported and be built for previous versions, but that would be a one time thing )

It could be run like `docker run ghcr.io/celestiaorg/celestia-node-compat-test:v6.1.2-arabica compat-test --rpc-url <some-url>`

With this `--rpc-url` command, all you need is 2 tags, one for your celestia-node and one for the compat-test container.

In this PR, there could be a type added that embeds the tastora `*container.Node`. This is a type that allows easy access to all the docker volume / exec functionality so none of that needs to be implemented here.

The go code in this PR, can just create an instance of `CompatTester` or whatever the struct is called. Run it with the `--rpc-url` of whatever node version you want, and then just check the exit code and stderr of the container. If all assertions pass it will be exit code 0 and the test passes, otherwise you get an error message to display and the go test will fail.

nodebuilder/tests/tastora/api/cross_version_client_test.go
gupadhyaya: removed.

nodebuilder/tests/tastora/api/cross_version_client_test.go
```diff
@@ -0,0 +1,392 @@
...
+
+	_, _, err := dockerClient.ImageInspectWithRaw(ctx, imageName)
+	if err != nil {
+		s.T().Logf("Pulling Docker image %s (this may take a while)...", imageName)
+		pullResp, pullErr := dockerClient.ImagePull(ctx, imageName, image.PullOptions{})
+		if pullErr != nil {
+			s.T().Logf("Failed to pull image %s: %v", imageName, pullErr)
+			s.T().Logf("Attempting to build image locally as fallback...")
+
+			if buildErr := s.buildCompatTestImageLocally(ctx, clientVersion, imageName); buildErr != nil {
```
chatton, line 231: for compat tests, I would imagine we never actually want to build the image locally, it should always be something that has been built from a tag, or this PR (prior to this go test running, so as a previous step in the workflow)

nodebuilder/tests/tastora/api/cross_version_client_test.go
```diff
@@ -0,0 +1,392 @@
...
+	config := &container.Config{
+		Image: imageName,
+		Cmd:   args,
+	}
+
+	hostConfig := &container.HostConfig{
+		NetworkMode: container.NetworkMode(networkID),
+	}
+
+	createResp, err := dockerClient.ContainerCreate(ctx, config, hostConfig, nil, nil, containerName)
```
chatton, line 260: rather than doing a raw container creation, take a look at how the [hermes](https://github.com/celestiaorg/tastora/blob/7defa8b2923b3e8461e9a3e15e65cbc170c1f940/framework/docker/ibc/relayer/hermes.go#L33) struct is implemented, you can create a type embedding *container.Node and get lots of functionality out of the box

cmd/compat-test/main.go
```diff
@@ -0,0 +1,192 @@
+package main
```
chatton, line 1: yes this is much nicer! I would recommend however that you pull this out into a separate PR, so that it is easy to backport to all the relevant release branches.

nodebuilder/tests/tastora/api/cross_version_client_test.go
```diff
@@ -0,0 +1,392 @@
...
+		err = s.runCompatTest(testCtx, oldVersion, serverRPC, true)
+		require.NoError(s.T(), err)
+	})
+}
+
+func (s *CrossVersionClientTestSuite) testAllAPIs(ctx context.Context, client *rpcclient.Client) {
+	s.testAllAPIsWithOptions(ctx, client, false)
+}
+
+func (s *CrossVersionClientTestSuite) testAllAPIsWithOptions(ctx context.Context, client *rpcclient.Client, skipGetRow bool) {
```
chatton, line 115: I think we don't need this anymore because we can replace it with using the container of whatever this PR is, that way all this compat logic is just in one place.

nodebuilder/tests/tastora/api/cross_version_client_test.go
```diff
@@ -0,0 +1,392 @@
...
+		".",
+	)
+	buildCmd.Dir = repoRoot
+	output, err := buildCmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("docker build failed: %w\nOutput: %s", err, string(output))
+	}
+
+	return nil
+}
```
chatton, line 370: I don't think we need this at all, we shouldn't need to execute any docker/git commands via cli like this, IMO the cleanest approach is just making sure that the images exist before the test runs.

chatton (commented):
Looking a lot better! Left a few suggestions for additional improvements

nodebuilder/tests/tastora/api/cross_version_client_test.go
gupadhyaya: if we intend to run compatibility test for every PR then we may need to build image locally right? PR branch and the branch on top of which it is based on? not necessarily always against a release version.

PR #4684, opened 2025-11-10
chore: bump app and core for v6.2.5

evan-forbes:
bumps app and core for v6.2.3


rach-id:
can we update this one to v6.2.4 instead of creating a new PR?

renaynay (approved):

PR #4688, opened 2025-11-11
chore: bump tastora to v0.8.0

chatton:
### Overview

The [v0.8.0](https://github.com/celestiaorg/tastora/releases/tag/v0.8.0) release includes fixes which prevent resources from being left over after a tests depending on where the docker client was initialized.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -153,7 +153,6 @@ require (
 	github.com/cockroachdb/redact v1.1.5 // indirect
 	github.com/cockroachdb/tokenbucket v0.0.0-20230807174530-cc333fc44b06 // indirect
 	github.com/cometbft/cometbft-db v1.0.4 // indirect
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -394,6 +394,9 @@ replace (
 	github.com/tendermint/tendermint => github.com/celestiaorg/celestia-core v1.55.0-tm-v0.34.35
 )
 
+// align replacement with tastora so the go mod parity script correctly identifies a match
+replace github.com/consensys/gnark-crypto => github.com/consensys/gnark-crypto v0.18.0
```
chatton, line 398: the parity script didn't seem to respect purely replacing the version, it needed the lines to be identical. 

e.g. this go.mod was already using v0.18.0, but I needed to add this line since I added a replace in the tastora go.mod.

I didn't want to get into modifying that script in this PR, and this should be functionally equivalent.

Issue #4673, opened 2025-10-30
[Feature Request]: make URL handling between BridgeDAAddr and Addr in Go Tx Client consistent


jcstein:
### Implementation ideas

Using the Celestia Go client with QuickNode is confusing and inconsistent when it comes to configuring addresses. The documentation and actual behavior of the client differ in ways that make it extremely difficult to get running without trial and error.

Specifically:

- The client expects `BridgeDAAddr` to include `https://`
- But Addr must not include https://, or you’ll get a colon parsing error.

This inconsistency is undocumented and counterintuitive.

**Steps to reproduce**

1. Configure the client as follows:
```
cfg := client.Config{
    ReadConfig: client.ReadConfig{
        BridgeDAAddr: "https://your-quicknode-url.celestia-mocha.quiknode.pro/<token>",
        EnableDATLS:  true,
    },
    Addr:    "https://your-quicknode-url.celestia-mocha.quiknode.pro:9090", // following what seems logical
    EnableTLS: true,
    AuthToken: "<token>",
}
```

2. run the client
3. observe the error `failed to get node info: rpc error: code = Unavailable desc = name resolver error: produced zero addresses`
4. if you remove `https://` from `Addr` it works
```
Addr: "your-quicknode-url.celestia-mocha.quiknode.pro:9090",
```

**Expected behavior**
- Either:
    - Both addresses should accept full URLs including `https://` (the intuitive behavior), or
    - The client should normalize the input (strip scheme, add default ports, etc.) so users don’t have to guess.

**Actual behavior**
- `BridgeDAAddr` expects a full HTTPS URL.
- `Addr` expects only `hostname:port` — adding `https://` breaks it with “too many colons in address”.

**Environment**

Go client version: v0.28.2-mocha
Node provider: QuickNode, with auth (x-token)
TLS: enabled or disabled doesn’t change the behavior

**Additional context**

This caused me confusion while writing the guide initially and @Ferret-san confusion when trying to connect the Celestia client to QuickNode. The [docs](https://docs.celestia.org/how-to-guides/client/go#_1-set-environment-variables) are accurate, but an experienced dev will assume you can just substitute URLs directly, but the client treats them differently internally. I've added a callout in an attempt to make it abundantly clear that it doesn't make sense right now: https://github.com/celestiaorg/docs/pull/2292

It might help to:
- update the client to handle URLs more consistently or normalize them automatically.

**Proposed fix**

[...]

PR #4689, opened 2025-11-11
fix: normalize grpc addr

gupadhyaya:
Fixes https://github.com/celestiaorg/celestia-node/issues/4673


walldiss:
Could you please confirm that it was tested against quicknode endpoints and works with setup either with and without https prefix?

PR #4695, opened 2025-11-14
fix(e2e-sanity-test): fix tastora DA provisioning

mojtaba-esk:
This PR proposes a fix the nil dereference when starting the Celestia chain and ensure the Tastora DA network pre-provisions light nodes based on the test config. Without this, integration suites that request a light node crash at setup.

Here are the two failures worth highlighting:
1. Initial panic from the framework:
```sh
panic.go:262: test panicked: runtime error: invalid memory address or nil pointer dereference
...
github.com/celestiaorg/celestia-node/nodebuilder/tests/tastora.(*...
```

2. After fixing that, the suites failed during setup:
```sh
framework.go:152: Cannot create more light nodes: already have 0, max is 0
```

### DIFF
--- a/nodebuilder/tests/tastora/framework.go
+++ b/nodebuilder/tests/tastora/framework.go
@@ -376,16 +376,26 @@ func (f *Framework) createBuilders(cfg *Config) (*cosmos.ChainBuilder, *dataavai
 		UIDGID:     "10001:10001",
 	}
 
-	// always have at least one bridge node.
-	bridgeNodeConfig := dataavailability.NewNodeBuilder().
-		WithNodeType(types.BridgeNode).
-		Build()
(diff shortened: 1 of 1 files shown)


nodebuilder/tests/tastora/framework.go
```diff
@@ -394,6 +404,7 @@ func (f *Framework) createBuilders(cfg *Config) (*cosmos.ChainBuilder, *dataavai
 func (f *Framework) createAndStartCelestiaChain(ctx context.Context) *cosmos.Chain {
 	celestia, err := f.chainBuilder.Build(ctx)
 	require.NoError(f.t, err, "failed to build celestia chain")
+	f.celestia = celestia
```
mojtaba-esk, line 407: This fixes the nil dereference panic.

PR #4696, opened 2025-11-17
fix!(shwap/rangeNamespaceDataID): fix RangeNamespaceDataID fields type

vgonkivs:
Updated indexes From, To because passing huge numbers could overflow 16 bit integer (in case of big blocks)


Wondertan (approved):

share/shwap/range_namespace_data_id.go
```diff
@@ -132,8 +134,8 @@ func RangeNamespaceDataIDFromBinary(data []byte) (RangeNamespaceDataID, error) {
 
 	rngID := RangeNamespaceDataID{
 		EdsID: edsID,
-		From:  int(binary.BigEndian.Uint16(data[EdsIDSize : EdsIDSize+2])),
-		To:    int(binary.BigEndian.Uint16(data[EdsIDSize+2 : EdsIDSize+4])),
```
walldiss, line 136: This is protocol breaking change. Lets add versioning and use 16 bits version in bitswap implementation and 32 bits in new shrex version. So we can deprecate older uint with protocol migration

Wondertan: Considering that the original is not widely deployed and used, i would just skip the versioning hussle(issueing version, backwards compatibility, migration). This more of a oversight quick fix, rather then a new version entirely.

walldiss: I will break every single client for this request on the network. We don't need fancy migration. Bitswap will use v0 while shrex can use new u32 based encoding

renaynay: so the suggestion is to do a v1 ?

share/shwap/range_namesapce_data_v0.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	edsID, err := EdsIDFromBinary(data[:EdsIDSize])
+	if err != nil {
+		return RangeNamespaceDataIDV0{}, err
+	}
+
+	rngID := RangeNamespaceDataID{
+		EdsID: edsID,
+		From:  int(binary.BigEndian.Uint16(data[EdsIDSize : EdsIDSize+2])),
```
renaynay: wait why are we using uint16 here?

vgonkivs:
We need RangeNamespaceDataIDV0 to be able to deprecate it along with the bitswap. Bitswap won't be deprecated after sampling over shrex is landed. It will live for some time to ensure that we have a fallback solution. When the time come, we will delete `RangeNamespaceDataIDV0` and that's it. We dont have to deal with versioning in future. Also, I dont feel it's necessary to bump the whole version to v1 just because the encoding changed.

This appraoch is imo less painful, allowing to delete all at once

walldiss (dismissed):

Issue #4296, opened 2025-06-03
[Feature Request]: Allow disabling p2p for storage-only node


adlerjohn:
### Implementation ideas

Currently, users have no way of easily disabling p2p. There's a class of users that don't care for p2p, and would prefer instead just using node for storage and RPC serving. For example, hosted RPC providers.

PR #4701, opened 2025-11-18
feat: Add storage-only mode for Bridge nodes with P2P disabled and ODS-only storage

gupadhyaya:
## Summary
Adds support for storage-only Bridge nodes that run without P2P networking, enabling hosted RPC providers to operate with minimal resource usage.

## Changes
- **P2P Disabled Mode**: New `p2p.disabled` config flag (CLI: `--p2p.disabled`) that disables all P2P components (host, DHT, Gossip, Shrex, Bitswap) for Bridge nodes
- **ODS-Only Storage**: New `share.store_ods_only` config flag that stores only ODS instead of ODSQ4, reducing storage requirements
- **Cross-Module Validation**: Enforces that `p2p.disabled` requires `share.store_ods_only` to be enabled
- **Stub Implementations**: Provides no-op implementations for P2P-dependent services (fraud, header subscriber/broadcaster) when P2P is disabled

## Testing
- Config validation tests for both P2P and Share modules
- Integration tests verifying storage-only mode behavior
- ODS-only storage behavior tests

## Use Case
Enables Bridge nodes to operate as storage-only RPC providers without P2P overhead, ideal for hosted infrastructure providers. #4296

### DIFF
--- a/core/eds.go
+++ b/core/eds.go
@@ -19,6 +19,7 @@ func storeEDS(
 	store *store.Store,
 	window time.Duration,
 	archival bool,
+	odsOnly bool,
 ) error {
 	if !archival && !availability.IsWithinWindow(eh.Time(), window) {
 		log.Debugw("skipping storage of historic block", "height", eh.Height())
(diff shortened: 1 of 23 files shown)


walldiss:
Outdated, will be reworked later when becomes a priority

PR #4719, opened 2025-11-29
fix: remove redundant nil check and fix comment

Bashmunta:
The nil-check for leaves in GetSharesByNamespace was unnecessary because CollectLeavesByNamespace returns an error on any retrieval failure, and the absent-namespace path exits before building shares. This could silently skip elements if retained. Also corrected the function comment to reflect actual behavior: errors short-circuit with no shares, while absent namespace returns no shares and a proof with err == nil.

### DIFF
--- a/share/ipld/get_shares.go
+++ b/share/ipld/get_shares.go
@@ -42,8 +42,8 @@ func GetShares(ctx context.Context, bg blockservice.BlockGetter, root cid.Cid, s
 }
 
 // GetSharesByNamespace walks the tree of a given root and returns its shares within the given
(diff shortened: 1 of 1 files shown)


vgonkivs:
There it is — line 243: n.addLeaf(j.sharePos, nil). When GetNode fails, it adds a nil leaf to the namespace data. The error is stored via singleErr.Do but processing continues — other leaves may succeed.
Then at line 225, retrievalErr is returned but Leaves() still contains nil entries from failed retrievals. So the nil check in GetSharesByNamespace is not redundant — it protects against panicking on these nil leaves.

PR #4721, opened 2025-12-01
test: introduce cel-compat binary

chatton:
This PR adds the compatibility testing binary that was being introduced in this PR #4680

By adding it in a single PR, we can more easily backport it to any branch that needs it.

For more context on its purpose see https://github.com/celestiaorg/celestia-node/pull/4680#pullrequestreview-3493287952


.github/workflows/docker-build-publish.yml
```diff
@@ -26,3 +27,14 @@ jobs:
       dockerfile: Dockerfile
       checkout_ref: ${{ github.event.inputs.ref }}
     secrets: inherit
+
+  # build the compatibility image for use in cross compatibility tests.
+  docker-compat-build:
```
chatton, line 32: we might actually need to move this into the same workflow as the test, so that the tests can depend on this image being built more easily, but that can happen in a followup, I don't want to re-structure workflows too much in this same PR.

cmd/cel-compat/main.go
```diff
@@ -0,0 +1,213 @@
...
+	client, err := rpcclient.NewClient(ctx, serverAddr, args.token)
+	if err != nil {
+		return fmt.Errorf("failed to create client: %w", err)
+	}
+	defer client.Close()
+
+	return runCompatibilityTests(ctx, client, args.skipGetRow)
+}
+
+func runCompatibilityTests(ctx context.Context, client *rpcclient.Client, skipGetRow bool) error {
```
chatton, line 84: I didn't make modifications to this, if anyone on the node team has more compatibility tests in mind, feel free to push to this branch or let me know what else needs to be tested. I was mainly focusing on how the image will be built and then how it will be utilized in future PRs.

chatton:
looks like bot closed this, @renaynay - is this still work that we want to do?

renaynay:
Hey @chatton I think it's been deprioritised but I will raise with the team and ask.

Issue #4729, opened 2025-12-08
service OOM during the synchronization of a full archival bridge node


gluckzhang:
### Celestia Node version

0.28.4-mocha

### OS

Ubuntu 24.04.3 LTS

### Steps to reproduce it

Init a bridge node and start syncing it from block 0.

### Expected result

The service runs continually to sync the blocks with a stable memory usage.

### Actual result

The bridge node service gets killed every now and then because of OOM. We have configured auto-restart for this systemd service unit, so after a restart, the service continues with the synchronization but it gets killed again due to OOM after several hours.

<img width="1488" height="538" alt="Image" src="https://github.com/user-attachments/assets/c94b42ad-ebdb-46ca-8a40-2db4b566bd63" />

### Relevant log output

```Shell
# kernel logs:

Dec 08 05:40:58 celestia-testnet kernel: oom-kill:constraint=CONSTRAINT_NONE,nodemask=(null),cpuset=/,mems_allowed=0,global_oom,task_memcg=/system.slice/celestia-bridge.service,task=celestia,pid=855114,uid=1002
Dec 08 05:40:58 celestia-testnet kernel: Out of memory: Killed process 855114 (celestia) total-vm:208342688kB, anon-rss:112867252kB, file-rss:24728kB, shmem-rss:0kB, UID:1002 pgtables:271968kB oom_score_adj:0
Dec 08 05:41:03 celestia-testnet kernel: oom_reaper: reaped process 855114 (celestia), now anon-rss:0kB, file-rss:780kB, shmem-rss:0kB

# service logs: see the attachment
```

### Notes

Here are the service logs captured around an OOM event. I have not found any errors there though.

[bridge-node-logs-OOM.log](https://github.com/user-attachments/files/24028643/bridge-node-logs-OOM.log)

Here is the systemd service config:

```
# /etc/systemd/system/celestia-bridge.service
[Unit]
Description=celestia-bridge Cosmos daemon
After=network-online.target

[Service]
User=celestia
ExecStart=/usr/local/bin/celestia bridge start --metrics.tls=false --metrics --metrics.endpoint localhost:4318 --core.ip <rpc_endpoint> --core.port 9090 --p2p.network mocha --node.store /celestia/bridge/.celestia-bridge-mocha-4 --archival
Restart=on-failure
RestartSec=3
LimitNOFILE=1400000

[Install]
WantedBy=multi-user.target
```

Here is the bridge node config:

```yaml
[Node]
[...]

PR #4733, opened 2025-12-08
fix(listener): fix memory leak in core.listener

walldiss:
Release pointer to data field before caching the Header.
Similar to https://github.com/celestiaorg/celestia-node/pull/3926/files. This time instead of exchange it fixes the listener to avoid capturing the block pointer by header cache

Resolves https://github.com/celestiaorg/celestia-node/issues/4729

### DIFF
--- a/core/fetcher_test.go
+++ b/core/fetcher_test.go
@@ -32,10 +32,10 @@ func TestBlockFetcher_GetBlock_and_SubscribeNewBlockEvent(t *testing.T) {
 			h := newBlockFromChan.Header.Height
 			block, err := fetcher.GetSignedBlock(ctx, h)
 			require.NoError(t, err)
-			assert.Equal(t, newBlockFromChan.Data, *block.Data)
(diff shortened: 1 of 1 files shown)


core/listener.go
```diff
@@ -182,8 +182,14 @@ func (cl *Listener) handleNewSignedBlock(ctx context.Context, b types.EventDataS
 		return fmt.Errorf("extending block data: %w", err)
 	}
 
+	// TODO(walldiss): Header pkg will capture pointers to those pointers in constructed extended header.
+	// copy to avoid keeping reference to whole block in memory after this function returns.
+	header := &b.Header
+	commit := &b.Commit
+	valSet := &b.ValidatorSet
```
Wondertan: Pass `SignedBlock` block in this func

also, why todo?

walldiss: thank you for suggestion, updated

Wondertan:
Tests are still borked

Wondertan (approved):

vgonkivs:
please check go mod in tastora

PR #4734, opened 2025-12-09
fix: treat CELESTIA_CUSTOM parse errors as fatal in ParseNetwork

yyhrnk:
ParseNetwork previously ignored errors returned by parseNetworkFromEnv when CELESTIA_CUSTOM was set but invalid, and silently fell back to the CLI network flag or default network. This could cause a node to join an unintended network if the custom env var was misconfigured. Now ParseNetwork returns any error from parseNetworkFromEnv and only falls back to flags when CELESTIA_CUSTOM is unset, making custom network configuration failures explicit instead of silent.


vgonkivs:
It does not silently fell back to the CLI network flag if invalid network is provided. `parseNetworkFromEnv` returns an empty string in case any error occurred. It fells down only if `EnvCustomNetwork` contains a valid network id or it's empty.

PR #4738, opened 2025-12-12
fix(core): Return partial response for headers if one fails

renaynay:
Fixes an issue where we would discard all headers in case one failed. Now we can at least return the ones we synced successfully (as long as they're contiguous from `from`) so we don't need to do duplicate work.

### DIFF
--- a/core/exchange.go
+++ b/core/exchange.go
@@ -135,6 +135,15 @@ func (ce *Exchange) getRangeByHeight(ctx context.Context, from, amount uint64) (
 	}
 
 	if err := errGroup.Wait(); err != nil {
+		// return partial range if possible
+		for i, h := range headers {
+			if h == nil || h.IsZero() {
+				if i > 0 {
+					return headers[:i], nil
+				}
+				break
+			}
+		}
 		return nil, err
 	}
 	log.Debugw("received headers", "from", from, "to", from+amount, "after", time.Since(start))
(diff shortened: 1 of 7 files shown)


walldiss (commented):
A few concerns:

**Syncer assumes full range on nil error.** In go-header's `sync/syncer.go:373`, `requestHeaders` does `amount -= size` (the *requested* size, not `len(headers)`) and advances `fromHead` to the last returned header. If this returns fewer headers with nil error, the syncer will silently skip the gap between what it got and what it thinks it got. There's also a panic if the partial result is empty (`headers[len(headers)-1]`).

**Concurrent fill makes nil-scan unreliable.** `errgroup.WithContext` cancels remaining goroutines on first error. Earlier slots may still be nil if those goroutines hadn't written yet. The contiguous prefix assumption only holds reliably when failures are at the tail (e.g. beyond chain tip), not for mid-range transient errors.

**`h.IsZero()` is redundant** — it's just `eh == nil`, same as the nil check before it.

Suggestion: either return the error alongside partial results (new sentinel like `ErrPartialRange` so callers can opt in), or fix the syncer in go-header to use `amount -= uint64(len(headers))` before landing this.

vgonkivs:
Moving it to draft until go-header pr merged and tested. @renaynay do you want to finish it then or should I take it over ?

vgonkivs (approved):
LGTM. I'd suggest to mark it as breaking. 

@renaynay do you want me to add a test case to tastora ?

renaynay:
Yes please add a test case to tastora @vgonkivs but how is this PR breaking?

PR #4752, opened 2026-01-09
!feat(api):Add headers to subscription response

alpe:
The `SubscriptionResponse` contains the height at which the blobs were included but no the timestamp. 
In our use case, the timestamp matters for delivery guarantees and deterministic behaviour.

### DIFF
--- a/blob/service.go
+++ b/blob/service.go
@@ -97,6 +97,7 @@ func (s *Service) Stop(context.Context) error {
 // If the Blobs slice is empty, it means that no blobs were included at the given height.
 type SubscriptionResponse struct {
 	Blobs  []*Blob
+	Height uint64 // Deprecated: use Header.Height() instead. Kept for backwards compatibility.
 	Header *header.RawHeader
 }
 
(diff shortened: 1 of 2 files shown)


blob/service.go
```diff
@@ -93,11 +93,12 @@ func (s *Service) Stop(context.Context) error {
 }
 
 // SubscriptionResponse is the response type for the Subscribe method.
-// It contains the blobs and the height at which they were included.
+// It contains the blobs, the height and time at which they were included.
 // If the Blobs slice is empty, it means that no blobs were included at the given height.
 type SubscriptionResponse struct {
 	Blobs  []*Blob
 	Height uint64
+	Time   time.Time
```
Wondertan: Calling it just Time is too vague. It should be HeaderTime.

On another note, what if we included the whole Header there? To future-proof for potentially other fields that may become needed.

blob/service.go
```diff
@@ -93,11 +93,11 @@ func (s *Service) Stop(context.Context) error {
 }
 
 // SubscriptionResponse is the response type for the Subscribe method.
-// It contains the blobs and the height at which they were included.
+// It contains the blobs and the header at which they were included.
 // If the Blobs slice is empty, it means that no blobs were included at the given height.
 type SubscriptionResponse struct {
 	Blobs  []*Blob
-	Height uint64
+	header *header.ExtendedHeader
```
Wondertan: Don't you need it public?

alpe: argh, I was in a rush. Should be public of course. Thanks!

Issue #4653, opened 2025-10-22
feat(core): Allow BN to sync headers from network outside of pruning window


walldiss:
### Implementation ideas

The current implementation of pruned BN syncs full blocks from the consensus starting from genesis. When prunIng is enabled it downloads the entire block but only stores the header outside of pruning window

To optimize this inefficiency, we should allow bridge nodes to sync headers from the network outside the pruning window. This mode is much more efficient due to reduced bandwidth usage and should become the default behavior.

However, we should still keep the option to sync BN exclusively from the consensus node.

PR #4758, opened 2026-01-14
feat: implement routing exchange for bridge nodes

vgonkivs:
This PR implements a RoutingExchange for Bridge nodes that optimizes header syncing by routing requests between core and P2P exchanges based on the availability window:

  - Inside availability window: Routes to core exchange (fetches full blocks from consensus) 
  - Outside availability window: Routes to P2P exchange (fetches headers only)

fixes https://github.com/celestiaorg/celestia-node/issues/4653

### DIFF
--- a/core/routing_exchange.go
+++ b/core/routing_exchange.go
@@ -31,14 +31,18 @@ func NewRoutingExchange(
 	p2pEx libhead.Exchange[*header.ExtendedHeader],
 	window time.Duration,
 	blockTime time.Duration,
-) *RoutingExchange {
+) (*RoutingExchange, error) {
+	if window == 0 || blockTime == 0 {
+		return nil, errors.New("window and blockTime can't be 0")
+	}
+
 	h := &RoutingExchange{
 		core:      coreEx,
 		p2p:       p2pEx,
 		window:    window,
 		blockTime: blockTime,
 	}
-	return h
+	return h, nil
 }
 
 // Head returns the latest header from the core exchange.
@@ -91,10 +95,7 @@ func (h *RoutingExchange) GetRangeByHeight(
 		return nil, nil
 	}
 
-	cutoff, err := h.calculateCutoffHeight(from)
-	if err != nil {
-		return nil, err
-	}
+	cutoff := h.calculateCutoffHeight(from)
 
 	// All headers are within window - use core
 	if startHeight > cutoff {
@@ -142,21 +143,17 @@ func (h *RoutingExchange) GetRangeByHeight(
 	return headers, nil
 }
 
-func (h *RoutingExchange) calculateCutoffHeight(from *header.ExtendedHeader) (uint64, error) {
-	if h.window == 0 || h.blockTime == 0 {
-		return 0, errors.New("window and blockTime can't be 0")
-	}
-
+func (h *RoutingExchange) calculateCutoffHeight(from *header.ExtendedHeader) uint64 {
 	// Cutoff time: headers with time >= cutoffTime are inside window
-	cutoffTime := time.Now().Add(-h.window)
+	cutoffTime := time.Now().UTC().Add(-h.window)
 
 	// If from is already inside window, all subsequent headers are too
 	if !from.Time().Before(cutoffTime) {
-		return 0, nil
+		return 0
 	}
 
 	// Estimate blocks from 'from' until cutoffTime
-	timeToCutoff := cutoffTime.Sub(from.Time())
+	timeToCutoff := cutoffTime.Sub(from.Time().UTC())
 	blocksToCutoff := uint64(timeToCutoff / h.blockTime)
-	return from.Height() + blocksToCutoff, nil
+	return from.Height() + blocksToCutoff
 }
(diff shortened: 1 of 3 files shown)


core/routing_exchange.go
```diff
@@ -0,0 +1,175 @@
...
+// Splits the range based on the availability window cutoff.
+func (h *RoutingExchange) GetRangeByHeight(
+	ctx context.Context,
+	from *header.ExtendedHeader,
+	to uint64,
+) ([]*header.ExtendedHeader, error) {
+	if from.Height() == to {
+		return []*header.ExtendedHeader{}, errors.New("empty range requested")
+	}
+	if from.Height() > to {
```
renaynay, line 89: funny we didn't do these sanity checks on core exchange

core/routing_exchange.go
```diff
@@ -0,0 +1,175 @@
...
+	if err != nil {
+		return nil, err
+	}
+	headers = append(headers, p2pHeaders...)
+
+	// Get recent headers from core (inside window)
+	if len(headers) > 0 {
+		from = headers[len(headers)-1]
+	}
+	coreHeaders, err := h.core.GetRangeByHeight(ctx, from, to)
```
renaynay, line 137: shouldn't this be cutoff, to as params? 

otherwise aren't we just trying to fetch full range from core exchange anyway

vgonkivs: from is updated at L160

core/routing_exchange.go
```diff
@@ -0,0 +1,175 @@
...
+	startHeight := from.Height() + 1
+	if startHeight > to {
+		return nil, nil
+	}
+
+	cutoff := h.windowCutoffHeight()
+
+	// All headers are within window - use core
+	if startHeight > cutoff {
+		return h.core.GetRangeByHeight(ctx, from, to)
```
walldiss: Can you wrap errors within this function. There are a lot of similar unwrapped calls and we might have trouble to find which one produced error.

core/routing_exchange.go
```diff
@@ -0,0 +1,175 @@
...
+	// Update cached cutoff height
+	h.updateCutoffHeight(head.Height())
+	return head, nil
+}
+
+// updateCutoffHeight recalculates and caches the cutoff height based on head height.
+func (h *RoutingExchange) updateCutoffHeight(headHeight uint64) {
+	if h.window == 0 || h.blockTime == 0 {
+		return
+	}
```
walldiss: Should we force those params to be non-zero? What is the purpose of optionality for those params?

core/routing_exchange.go
```diff
@@ -0,0 +1,175 @@
...
+	cutoff := h.windowCutoffHeight()
+	if height > cutoff {
+		return h.core.GetByHeight(ctx, height)
+	}
+	return h.p2p.GetByHeight(ctx, height)
+}
+
+// GetRangeByHeight retrieves a range of headers.
+// Splits the range based on the availability window cutoff.
+func (h *RoutingExchange) GetRangeByHeight(
```
walldiss, line 81: It seems it is required cutoff height to be initiallised before other methods (GetRAngeByHeight and GetByHeight) can be called. How can we ensure Head() is called first? Or can we ensure the RoutingExchange is initialised first before allowing other methods to be called

core/routing_exchange.go
vgonkivs: avoiding a lot of input params in constructor.

core/routing_exchange.go
vgonkivs: `Head()` is called on every startup, so cut-off will be calculated earlier. Or maybe I misunderstood your question?

vgonkivs: reworked the logic

walldiss (approved):

core/routing_exchange.go
```diff
@@ -0,0 +1,162 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("requesting partial range from consensus node: %w", err)
+	}
+	headers = append(headers, coreHeaders...)
+
+	return headers, nil
+}
+
+func (h *RoutingExchange) calculateCutoffHeight(from *header.ExtendedHeader) (uint64, error) {
+	if h.window == 0 || h.blockTime == 0 {
```
walldiss: Should it be construction time check?

core/routing_exchange.go
```diff
@@ -0,0 +1,162 @@
...
+		return 0, errors.New("window and blockTime can't be 0")
+	}
+
+	// Cutoff time: headers with time >= cutoffTime are inside window
+	cutoffTime := time.Now().Add(-h.window)
```
walldiss: utc()

core/routing_exchange.go
```diff
@@ -0,0 +1,162 @@
...
+
+	// If from is already inside window, all subsequent headers are too
+	if !from.Time().Before(cutoffTime) {
```
walldiss, line 151: utc()

core/routing_exchange.go
vgonkivs: yes

core/routing_exchange.go
vgonkivs: changed

core/routing_exchange.go
vgonkivs: changed

PR #4759, opened 2026-01-14
chore(lifecycle): disable service breaker

vgonkivs:
Disable service breaker as we don't really need this logic anymore.


renaynay (approved):
<3

renaynay:
but lint pls

PR #4761, opened 2026-01-15
chore(befp): improve validation

vgonkivs:
Improve validation

### DIFF
--- a/share/eds/byzantine/bad_encoding.go
+++ b/share/eds/byzantine/bad_encoding.go
@@ -178,7 +178,7 @@ func (p *BadEncodingProof) Validate(hdr *header.ExtendedHeader) error {
 		return errIncorrectAmountOfShares
 	}
 
-	indexes := make([]int, 0) // nolint:prealoc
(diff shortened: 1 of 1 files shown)


share/eds/byzantine/bad_encoding.go
```diff
@@ -189,6 +191,14 @@ func (p *BadEncodingProof) Validate(hdr *header.ExtendedHeader) error {
...
 		}
 		shares[index] = shr.ToBytes()
+		indexes = append(indexes, shr.Proof.Start())
+	}
+
+	isSorted := slices.IsSorted(indexes)
+	if !isSorted {
+		err := errors.New("provided share list is not sorted")
+		log.Debugf("%v", err)
+		return err
```
Wondertan: optional simplification nit: i think we could check if its sorted by simply looking if the previous share.proof equals to current index

vgonkivs: we can build a befp against row, using Shares with Proof built against Col. In this case Proof.Start() will be equal for all shares

PR #4763, opened 2026-01-16
Implement subscribe from height

vgonkivs:
Implement `SubscribeFrom` that allows to sync blobs from historical headers.

### DIFF
--- a/blob/service.go
+++ b/blob/service.go
@@ -116,7 +116,7 @@ func (s *Service) Subscribe(ctx context.Context, ns libshare.Namespace) (<-chan 
 		return nil, err
 	}
 
-	blobCh := make(chan *SubscriptionResponse, 16)
+	blobCh := make(chan *SubscriptionResponse, 32)
 	go s.handleSubscription(ctx, headerCh, blobCh, ns)
 	return blobCh, nil
 }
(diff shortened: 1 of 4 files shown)


blob/service.go
```diff
@@ -109,65 +109,71 @@ func (s *Service) Subscribe(ctx context.Context, ns libshare.Namespace) (<-chan
...
+
+	// wait for current network head
+	var hdr *header.ExtendedHeader
+	select {
+	case hdr = <-headerCh:
+		if hdr.Height() < height {
+			return nil, fmt.Errorf("network height %d is less than requested height %d", hdr.Height(), height)
 		}
-	}()
+	case <-ctx.Done():
```
renaynay: nit this should be first case

renaynay (commented):
I don't have any problems w/ adding this method but we have to be mindful of gossipsub buffer - IIRC we have 32 msg buffer w/ the headersub, gives us ~3 min to fetch all blobs between height --> network head before we start the live blobsub. So if we have a significant distance between height --> network head that comes thru headersub, we may have a problem of dropped headers.

vgonkivs:
closing until there will be a specific request for it

PR #4765, opened 2026-01-22
feat(shwap/blob): add blob to shwap

vgonkivs:
The PR adds a support of Blob type request and response for the shrex/protocol. Please note that the PR is breaking as it breaks the proof verification logic: instead of proving the whole namespace within the Row, blob Proof now tied to the specific blob.


share/shwap/getters/cascade.go
```diff
@@ -145,6 +145,49 @@ func (cg *CascadeGetter) GetRangeNamespaceData(
...
+func (cg *CascadeGetter) GetBlob(
+	ctx context.Context,
+	header *header.ExtendedHeader,
+	ns libshare.Namespace,
+	commitment []byte,
+) (*shwap.Blob, error) {
+	ctx, span := tracer.Start(
+		ctx,
+		"cascade/get-blob",
+		trace.WithAttributes(
```
renaynay, line 158: add commitment here

share/eds/rsmt2d.go
```diff
@@ -177,6 +177,44 @@ func (eds *Rsmt2D) Reader() (io.Reader, error) {
...
+		sh, err := libshare.FromBytes(rawShare)
+		if err != nil {
+			return nil, err
+		}
+		extendedRowShares[index] = sh
+	}
+	return shwap.BlobFromShares(extendedRowShares, namespace, commitment, odsSize)
+}
+
+func (eds *Rsmt2D) Blobs(ctx context.Context, namespace libshare.Namespace) ([]*shwap.Blob, error) {
```
renaynay: where are unit tests for these methods

share/shwap/p2p/shrex/shrex_getter/shrex.go
```diff
@@ -384,6 +385,106 @@ func (sg *Getter) GetRangeNamespaceData(
...
 }
 
+func (sg *Getter) GetBlob(
+	ctx context.Context,
+	header *header.ExtendedHeader,
+	ns libshare.Namespace,
+	commitment []byte,
+) (*shwap.Blob, error) {
+	var err error
+	ctx, span := tracer.Start(ctx, "shrex/get-blob", trace.WithAttributes(
```
renaynay, line 395: we need to pull span from the trace that's passed through cascade getter rather than starting new one

share/shwap/p2p/shrex/shrex_getter/shrex.go
```diff
@@ -384,6 +385,106 @@ func (sg *Getter) GetRangeNamespaceData(
...
+	return &response, nil
+}
+
+func (sg *Getter) GetBlobs(
+	ctx context.Context,
+	header *header.ExtendedHeader,
+	ns libshare.Namespace,
+) ([]*shwap.Blob, error) {
+	var err error
+	ctx, span := tracer.Start(ctx, "shrex/get-blobs", trace.WithAttributes(
```
renaynay, line 442: same

share/shwap/p2p/shrex/shrex_getter/shrex.go
```diff
@@ -384,6 +385,106 @@ func (sg *Getter) GetRangeNamespaceData(
...
+	req := func(ctx context.Context, peer libpeer.ID) error {
+		return sg.client.Get(ctx, &request, &response, peer)
+	}
+
+	verify := func() error {
+		if response == nil {
+			return errors.New("nil response")
+		}
+		for _, blob := range response {
+			err = blob.VerifyInclusion(header.DAH.RowRoots)
```
renaynay, line 472: why are we only calling VerifyInclusion here instead of top-level Verify?

share/shwap/blob.go
```diff
@@ -0,0 +1,333 @@
...
+		if err != nil {
+			return nil, err
+		}
+		endIndex := fromIndex + sharesAmount - 1
+		to, err := SampleCoordsFrom1DIndex(endIndex, odsSize)
+		if err != nil {
+			return nil, err
+		}
+
+		rngShares := libShares[rowStart : to.Row+1]
```
renaynay: This feels vulnerable to dos from server - can pass malicious overflowing start index as uint64 + give you an out of bounds on the row check.

Solution would be to sanity check lower bound (0) in `SampleCoordsFrom1DIndex` or fix in BlobFromProto to insure start index does not overflow max int64. Or both.

renaynay: would also be ideal to write a test for this case.

share/shwap/blob.go
```diff
@@ -0,0 +1,333 @@
...
+	if err != nil {
+		return Blob{}, err
+	}
+	return Blob{
+		RangeNamespaceData: &rngData,
+		StartIndex:         int(pbBlob.StartIndex),
+	}, nil
+}
+
+func BlobsFromShares(
```
renaynay: Was all of this code moved from the blob pkg down here?

share/shwap/blob_id.go
```diff
@@ -0,0 +1,169 @@
...
+	"context"
+	"fmt"
+	"io"
+
+	libshare "github.com/celestiaorg/go-square/v3/share"
+)
+
+// BlobIDSize defines the total size of a BlobID in bytes, combining the
+// size of a EdsID, the size of a Namespace and the commitment size
+const BlobIDSize = EdsIDSize + libshare.NamespaceSize + commitmentSize
```
renaynay, line 14: is there any test coverage for this file?

share/shwap/blob_id.go
```diff
@@ -0,0 +1,169 @@
...
+		return nil, err
+	}
+	data = append(data, blbID.DataNamespace.Bytes()...)
+	return append(data, blbID.Commitment...), nil
+}
+
+func (blbID *BlobID) ResponseReader(ctx context.Context, acc Accessor) (io.Reader, error) {
+	buf := &bytes.Buffer{}
+
+	if !bytes.Equal(blbID.Commitment, emptyCommitment) {
```
renaynay, line 143: Switch case on here to make more readable?

share/shwap/range_namespace_data.go
```diff
@@ -68,7 +68,12 @@ func RangeNamespaceDataFromShares(
 	}
 
 	namespace := extendedRowShares[0][from.Col].Namespace()
-	odsSize := len(extendedRowShares[0]) / 2
+
+	// Make a copy of the outer slice to avoid modifying the caller's slice
+	// when we reassign row elements below.
+	shares := make([][]libshare.Share, len(extendedRowShares))
```
renaynay, line 72: can we extract into separate PR as this is a big feature PR and would rather not bundle other stuff into

store/file/ods.go
```diff
@@ -358,6 +358,50 @@ func (o *ODS) RangeNamespaceData(
 	return shwap.RangeNamespaceDataFromShares(shares, fromCoords, toCoords)
 }
 
+func (o *ODS) Blob(ctx context.Context, namespace libshare.Namespace, commitment []byte) (*shwap.Blob, error) {
```
renaynay: where are units?

renaynay (commented):
Some questions:
1. what is the point of `GetBlobs` to be implemented over blob over shwap -- why wouldn't it just route to namespace data over shwap? Is this just an optimisation not to send padding if it exists? If so, then what's the point of shares by namespace over shwap (besides namespace non-inclusion)?
2. similarly, `GetBlobs()` feels to me like it should be a message at shwap protocol level for fetching multiple blobs by commitment rather than just `all` -- right now, we can only fetch 1 blob by commitment over shwap or fetching ALL blobs in that namespace. Is there a case that we would need that? fetching multiple blobs by commitment from the same namespace? And not to bikeshed, but similarly, would there be a need to fetch blobs from different namespaces but at the same `height` such that it would be more efficient to batch it in one request?
3. Can we DRY between Blob()/Blobs() methods between rsmt2d.go and ods.go
4. Why do we need two methods on accessor? Blob + Blobs? Can we not just have 1 (Blobs) and pass in 1 or multiple there? Right now, this is probably b/c we have the binary 1 blob or ALL blobs.
5. feels like missing lots of unit tests

share/shwap/p2p/shrex/shrex_getter/shrex.go
vgonkivs: This is the problem of all methods in shrex. Probably, I missed this part during shrex refactoring. I'll prepare a PR to fix all at once.

vgonkivs:
> what is the point of GetBlobs to be implemented over blob over shwap -- why wouldn't it just route to namespace data over shwap? Is this just an optimisation not to send padding if it exists? If so, then what's the point of shares by namespace over shwap (besides namespace non-inclusion)?

The point is to have a proof, associated with blob and not for the whole namespace in the row.


> similarly, GetBlobs() feels to me like it should be a message at shwap protocol level for fetching multiple blobs by commitment rather than just all -- right now.

You can request multiple blobs by commitment from different peers(as GetSamples) and make requests parallel. We can add this if it will be requested.
The point of `give all blobs from this namespace` is to support blob subscription.

> And not to bikeshed, but similarly, would there be a need to fetch blobs from different namespaces but at the same height such that it would be more efficient to batch it in one request? 

We are not requesting samples as batches from the security standpoint. Maybe we should follow this approach with blobs as well?


> Can we DRY between Blob()/Blobs() methods between rsmt2d.go and ods.go

Yes

> Why do we need two methods on accessor? Blob + Blobs? Can we not just have 1 (Blobs) and pass in 1 or multiple there? Right now, this is probably b/c we have the binary 1 blob or ALL blobs.

Right. Thank you.

> feels like missing lots of unit tests

Already working on them.

store/file/ods.go
vgonkivs: already working on them

share/shwap/p2p/shrex/shrex_getter/shrex.go
vgonkivs: We don't know the commitments during GetBlobs. So we are interested only in inclusion.

share/shwap/blob.go
vgonkivs: no

share/shwap/range_namespace_data.go
vgonkivs: This change is necessary and imo should be a part of this PR. Previously, a single range was requested which allows us to modify the outer slice. In case of blob's retrieval it will lead to an error with indexing, out of bounds errors etc.

share/eds/rsmt2d.go
vgonkivs: added

nodebuilder/da/service.go
```diff
@@ -120,7 +121,7 @@ func (s *Service) GetIDs(ctx context.Context, height uint64, namespace da.Namesp
 	blobs, err := s.blobServ.GetAll(ctx, height, []libshare.Namespace{ns})
 	log.Debugw("got ids", "height", height, "namespace", ns)
 	if err != nil {
-		if strings.Contains(err.Error(), blob.ErrBlobNotFound.Error()) {
+		if strings.Contains(err.Error(), shwap.ErrBlobNotFound.Error()) {
 			return nil, nil //nolint:nilnil
 		}
```
walldiss, line 126: hmm, why don't we return non-inclusion proof?

share/shwap/getter.go
```diff
@@ -81,4 +81,12 @@ type Getter interface {
...
 		from, to int,
 	) (RangeNamespaceData, error)
+
+	// GetBlob retrieves a single blob identified by its namespace and commitment
+	// from the Extended Data Square (EDS) referenced by the given header.
+	GetBlob(_ context.Context, _ *header.ExtendedHeader, _ libshare.Namespace, commitment []byte) (*Blob, error)
+
+	// GetBlobs retrieves all blobs belonging to the given namespace from the
+	// Extended Data Square (EDS) referenced by the given header.
+	GetBlobs(_ context.Context, _ *header.ExtendedHeader, _ libshare.Namespace) ([]*Blob, error)
```
walldiss, line 91: Why you added this method? It seems to duplicate blobs parsing from blob service but leaking into reader level for some reason.

share/eds/accessor.go
```diff
@@ -39,6 +39,11 @@ type Accessor interface {
 		ctx context.Context,
 		from, to int,
 	) (shwap.RangeNamespaceData, error)
+
+	// Blobs retrieves blobs belonging to the given namespace from the data square.
+	// When commitments are provided, only blobs matching those commitments are returned.
+	// When no commitments are provided, all blobs in the namespace are returned.
+	Blobs(context.Context, libshare.Namespace, ...[]byte) ([]*shwap.Blob, error)
```
walldiss, line 46: It is a bit confusing why some APIs and interfaces are requesting single blobs while others are requesting multiple. We used to have ability to request only one blob by commitment. Are there some fundamentl changes required to make this PR? Imo ability to request multiple blobs by commitment at the same time is out of scope of this change and should be discussed / implemented separetly.

share/eds/rsmt2d.go
```diff
@@ -177,6 +177,37 @@ func (eds *Rsmt2D) Reader() (io.Reader, error) {
...
+	for index := 0; index < odsSize; index++ {
+		rawShare := eds.Row(uint(index))
+		sh, err := libshare.FromBytes(rawShare)
+		if err != nil {
+			return nil, 0, err
+		}
+		shares[index] = sh
+	}
+	return shares, odsSize, nil
+}
```
walldiss, line 209: It is duplication. You can get extended row from just Row method.

share/shwap/blob.go
vgonkivs: done

share/eds/accessor.go
vgonkivs: Top API level contains 2 independent methods(GetBlob/GetBlobs), while accessor has only one, since the logic is pretty the same and @renaynay suggested to have only one method.

nodebuilder/da/service.go
vgonkivs: We havent returned it previously

vgonkivs:
will come back to in when this feature will be requested

Issue #4755, opened 2026-01-13
Consider enabling pruning for BNs


Wondertan:
The requirement to run an archival consensus node to run BN causes significant pain. One of the easiest things to do is simply allow header pruning for BN, which it is already capable of, but was artificially limited to ensure architectural header sync guarantees on the network.

Given that we don't want to run archival nodes long-term, I would check whether we are fine with losing headers forever and irrecoverably. If so, we may ignore the [p2p routing issue](https://github.com/celestiaorg/go-header/issues/333) and accept that archival sync or sync from a very far-in-the-past height may break. 

To be more precise, archival sync may not work even when an archival node is on the network, but if every archival sync becomes necessary, we can always come back and fix https://github.com/celestiaorg/go-header/issues/333. The only risk here would be that archival headers may be lost, given that everyone will be incentivized to migrate to pruned nodes.

The alternative is to do #4653, which speeds up the sync, similarly allows pruned consensus nodes instead of archival, yet keeps the same guarantee about header availability by not allowing BNs to be pruned, which is IMO is still a preferred option for a general node runner. Some customized builds may allow pruned BN when applicable instead.

PR #4768, opened 2026-01-26
feat(pruning): allow pruning for Bridge Nodes

vgonkivs:
Closes https://github.com/celestiaorg/celestia-node/issues/4755

What were done:
Bridge Nodes can now prune headers and start syncing from a specific height or hash instead of syncing all headers from genesis. The `--archival` flag disables pruning and syncs from genesis.

Changes:
* Prunning is enabled by default for Bridge Nodes;
* The user should explicitly set `--archival` to run an archival node;
* SyncFromHeight / SyncFromHash can be set for Bridge Nodes;

### DIFF
--- a/nodebuilder/header/constructors.go
+++ b/nodebuilder/header/constructors.go
@@ -28,7 +28,7 @@ func newP2PExchange[H libhead.Header[H]](
 	host host.Host,
 	conngater *conngater.BasicConnectionGater,
 	pidstore p2p.PeerIDStore,
-) (libhead.Exchange[H], error) {
+) (*p2p.Exchange[H], error) {
 	peers, err := cfg.trustedPeers(bpeers)
 	if err != nil {
 		return nil, err
--- a/core/routing_exchange.go
+++ b/core/routing_exchange.go
@@ -1,0 +1,159 @@
+package core
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	libhead "github.com/celestiaorg/go-header"
+
+	"github.com/celestiaorg/celestia-node/header"
+)
+
+// RoutingExchange combines core and P2P exchanges to optimize header syncing for Bridge nodes.
+// It routes requests based on whether headers fall within the availability window:
+//   - Inside window: uses core exchange (fetches full blocks from consensus)
+//   - Outside window: uses P2P exchange (fetches headers only from network)
+type RoutingExchange struct {
+	core libhead.Exchange[*header.ExtendedHeader]
+	p2p  libhead.Exchange[*header.ExtendedHeader]
+
+	// window is the storage window duration used to calculate the cutoff height.
+	window time.Duration
+	// blockTime is the expected block production time.
+	blockTime time.Duration
+}
+
+// NewRoutingExchange creates a new RoutingExchange that wraps core and P2P exchanges.
+func NewRoutingExchange(
+	coreEx libhead.Exchange[*header.ExtendedHeader],
+	p2pEx libhead.Exchange[*header.ExtendedHeader],
+	window time.Duration,
+	blockTime time.Duration,
+) (*RoutingExchange, error) {
+	if window == 0 || blockTime == 0 {
+		return nil, errors.New("window and blockTime can't be 0")
+	}
+
+	h := &RoutingExchange{
+		core:      coreEx,
+		p2p:       p2pEx,
+		window:    window,
+		blockTime: blockTime,
+	}
+	return h, nil
+}
+
+// Head returns the latest header from the core exchange.
+// We always use core for Head() to get the authoritative latest block.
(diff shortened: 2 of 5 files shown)


nodebuilder/header/config.go
```diff
@@ -85,18 +89,25 @@ func (cfg *Config) trustedPeers(bpeers p2p.Bootstrappers) (infos []peer.AddrInfo
...
 func (cfg *Config) Validate(tp node.Type) error {
 	switch tp {
-	case node.Full, node.Bridge:
+	case node.Full:
+		// Full nodes must sync all headers from genesis
 		if cfg.Syncer.SyncFromHash != "" || cfg.Syncer.SyncFromHeight != 0 || cfg.Syncer.PruningWindow != 0 {
 			return fmt.Errorf(
-				"module/header: Syncer.SyncFromHash/Syncer.SyncFromHeight/Syncer.PruningWindow must not be set for FN/BN nodes" +
-					"until https://github.com/celestiaorg/go-header/issues/333 is completed. Full and Bridge must sync all the headers until then",
+				"module/header: Syncer.SyncFromHash/Syncer.SyncFromHeight/Syncer.PruningWindow must not be set for FN nodes " +
```
renaynay, line 96: we don't have full nodes anymore so idek if warning is necessary.

nodebuilder/header/constructors.go
```diff
@@ -72,18 +72,36 @@ func newSyncer[H libhead.Header[H]](
...
+		if isArchival {
+			// Archival mode: disable header pruning and sync from genesis
+			genesis, err := modp2p.GenesisFor(net)
+			if err != nil {
+				return nil, err
+			}
+			cfg.Syncer.PruningWindow = 0
+			cfg.Syncer.SyncFromHash = genesis
+			if genesis == "" {
+				cfg.Syncer.SyncFromHeight = 1
```
renaynay, line 99: Can we just do this in config since we've already introduced the the type anyway.

nodebuilder/pruner/module.go
```diff
@@ -39,6 +39,9 @@ func ConstructModule(tp node.Type) fx.Option {
 		// supply the default config, which can only be overridden by
 		// passing the `--archival` flag
 		fx.Supply(DefaultConfig()),
+		fx.Provide(func(cfg *Config) node.ArchivalMode {
```
renaynay, line 43: can this be supply

nodebuilder/header/constructors.go
vgonkivs: Yes, we actually can set it in syncer's config. But I'd suggest to go with these rn and after releasing go-header with config changes, I will update this part

nodebuilder/pruner/module.go
vgonkivs: I tried supply, but fails with a bug(where I can't run an archival node at all). It took me some time to realise that:
1) fx.Supply - provides a pre-computed value at module construction time
2) fx.Provide - provides a factory function that runs during dependency injection

So in my case Supply stored a default value.

nodebuilder/header/config.go
vgonkivs: lets wait until full removal

nodebuilder/pruner/module.go
renaynay: But isn't cfg.EnableService already injected during construction time? Or does it literally just take the default value of that field?

renaynay (approved):
lgtm

nodebuilder/header/constructors.go
```diff
@@ -72,18 +72,36 @@ func newSyncer[H libhead.Header[H]](
...
+		// Bridge nodes: check if archival mode is enabled via --archival flag
+		if isArchival {
+			// Archival mode: disable header pruning and sync from genesis
+			genesis, err := modp2p.GenesisFor(net)
+			if err != nil {
+				return nil, err
+			}
+			cfg.Syncer.PruningWindow = 0
+			cfg.Syncer.SyncFromHash = genesis
+			if genesis == "" {
```
renaynay, line 98: I still don't fully understand why this check is necessary - if we do not get a genesis hash && node isArchival then we should short-circuit as genesis hash is necessary for fetching first block and technically *should* be defined (even if passed via CUSTOM)

nodebuilder/pruner/module.go
vgonkivs: looks like at this particular moment is not overwritten and the default value is used(wasn't able to run an archival node at all)

nodebuilder/header/constructors.go
vgonkivs: oh, I see the problem now. Yeah, lets leave hash only

vgonkivs: updated

walldiss (approved):

PR #4773, opened 2026-02-05
refactor: Delete option to run full node

walldiss:
Remove deprecated full nodes from codebase


liamsi (dismissed):
Code changes look good and complete. Will we have some migration notice in the changelog or somewhere? I am not sure how many people runn full nodes (if there any at all) but it would be good to document the deprecation and how to get the same functionality (bridge nodes) prominently.

renaynay (dismissed):
goodnight sweet prince

Issue #4770, opened 2026-02-03
Bridge node should fallback to other bridge nodes when the consensus node doesn't have the block


cmwaters:
If the consensus node has pruned the height, the bridge node should then look for other bridge nodes for the data. As far as I understand this is an existing functionality that just needs to be re-enabled

PR #4776, opened 2026-02-11
feat(core): allow BN to BN sync

walldiss:
Solution is leveraging the existing syncer to download headers only from p2p in cases where the consensus node has pruned the corresponding heights.

In parallel, we can rely on the catch-up DASer to backfill the missing EDSes in the background. Since the DASer short-circuits when an EDS is already present in the store, this approach should avoid double work.

This keeps the trust model clean:

- Headers are synced and verified via the normal p2p header sync process.
- EDSes are retrieved and verified through DAS.
- No need to reintroduce the old “trusted parallel block sync” behavior.
- No need to maintain two separate full syncing pipelines inside the bridge node.

This approach reuses existing components (syncer + catch-up DASer), minimizes architectural changes, and preserves performance characteristics.

Resolves https://github.com/celestiaorg/celestia-node/issues/4770

### DIFF
--- a/api/rpc/server.go
+++ b/api/rpc/server.go
@@ -1,4 +1,3 @@
-// Package rpc provides RPC server implementation
 package rpc
 
 import (
--- a/nodebuilder/rpc/defaults.go
+++ b/nodebuilder/rpc/defaults.go
@@ -1,4 +1,3 @@
-// Package rpc provides RPC configuration defaults
 package rpc
 
 const (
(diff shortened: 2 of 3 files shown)


core/exchange.go
```diff
@@ -127,6 +131,11 @@ func (ce *Exchange) Get(ctx context.Context, hash libhead.Hash) (*header.Extende
 	log.Debugw("requesting header", "hash", hash.String())
 	block, err := ce.fetcher.GetBlockByHash(ctx, hash)
 	if err != nil {
+		log.Debugw("failed to fetch block by hash from core, trying fallback", "hash", hash.String(), "error", err)
```
renaynay, line 134: nit move debug log into ce.p2pExchange nest

PR #4781, opened 2026-02-13
feat(core): verify core endpoint network on startup

walldiss:
## Summary
  - Add `BlockFetcher.ChainID()` method that queries the core endpoint's Status RPC to retrieve the network name
  - Verify the chain ID on `Listener.Start()` before subscribing to block events — if the core endpoint is on a different network than configured, startup fails with a clear error
  - Add chain ID validation to `Exchange` for runtime safety (panics on mismatch, matching existing Listener behavior)
  - Pass `chainID` option to `Exchange` in the module wiring

  ## Motivation
  Previously, a chain ID mismatch was only detected at runtime when the first block arrived, causing a panic. This change catches the misconfiguration immediately at startup with a descriptive error message, making it much easier to diagnose wrong core endpoint configurations.

  ## Test plan
  - [x] Updated `TestListenerWithWrongChainRPC` to verify that `Start()` returns an error (instead of testing the runtime panic path)
  - [ ] Manual test: configure a node with a core endpoint pointing to a different network and confirm startup fails with `"core endpoint network mismatch"` error

Closes: https://linear.app/celestia/issue/PROTOCO-1349/featcore-verify-core-endpoint-network-on-startup


core/exchange.go
```diff
@@ -139,6 +142,15 @@ func (ce *Exchange) Get(ctx context.Context, hash libhead.Hash) (*header.Extende
 		return nil, fmt.Errorf("fetching block by hash %s: %w", hash.String(), err)
 	}
 
+	if ce.chainID != "" && block.ChainID != ce.chainID {
```
mcrakhman: [nit] mb extract into method, used twice

nodebuilder/testing.go
```diff
@@ -52,6 +50,15 @@ func TestNodeWithConfig(t *testing.T, tp node.Type, cfg *Config, opts ...fx.Opti
...
 	cfg.Header.TrustedPeers = []string{"/ip4/1.2.3.4/tcp/12345/p2p/12D3KooWNaJ1y1Yio3fFJEXCZyd1Cat3jmrPdgkYCrHfKD3Ce21p"}
 
+	// Bridge node requires a core endpoint. Set config IP/Port so the
+	// production grpcClient provider creates a properly configured connection.
+	if tp == node.Bridge && cfg.Core.IP == "" {
+		tn := core.StartTestNode(t)
+		var err error
+		cfg.Core.IP, cfg.Core.Port, err = net.SplitHostPort(tn.GRPCClient.Target())
+		require.NoError(t, err)
+	}
```
Wondertan, line 60: 👍

core/listener.go
```diff
@@ -86,11 +86,15 @@ func NewListener(
 }
 
 // Start kicks off the Listener listener loop.
-func (cl *Listener) Start(context.Context) error {
+func (cl *Listener) Start(ctx context.Context) error {
 	if cl.cancel != nil {
 		return fmt.Errorf("listener: already started")
 	}
 
+	if err := cl.verifyChainID(ctx); err != nil {
```
Wondertan, line 94: Do we really need to fetch chain ID from core here? Cant we use chain id on the incoming data, like in exchange?

walldiss: I want to do Start-up time check so it can error out properly. When incoming data is processed it is already too late for errors and it have to panic. This check would catch incorrect chainid before panic is the only option and can return with error gracefuly

Wondertan (approved):

PR #4793, opened 2026-02-18
fix(nodebuilder/core): fix inverted gRPC connection readiness check

walldiss:
WaitForStateChange(ctx, connectivity.Ready) waits for the state to change from Ready. Since a local test connection reaches Ready almost instantly after conn.Connect(), the call then just blocks until the context expires (15s test deadline), returning false → "couldn't connect to core endpoint"

Closes: https://linear.app/celestia/issue/PROTOCO-1341/fixnodebuildercore-fix-inverted-grpc-connection-readiness-check

### DIFF
--- a/.github/workflows/ci_release.yml
+++ b/.github/workflows/ci_release.yml
@@ -65,7 +65,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: actions/setup-node@v5
+      - uses: actions/setup-node@v6
         with:
           node-version: 18
       - run: |
--- a/cmd/cel-shed/ods_verify.go
+++ b/cmd/cel-shed/ods_verify.go
@@ -1,0 +1,469 @@
+package main
+
+import (
+	"bytes"
+	"context"
+	"errors"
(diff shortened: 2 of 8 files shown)


Wondertan (approved):
Could be a single if statement but fine

nodebuilder/core/constructors.go
```diff
@@ -97,10 +97,15 @@ func grpcClient(lc fx.Lifecycle, cfg EndpointConfig) (*grpc.ClientConn, error) {
...
-			if !conn.WaitForStateChange(ctx, connectivity.Ready) {
-				return errors.New("couldn't connect to core endpoint")
+			for {
+				state := conn.GetState()
+				if state == connectivity.Ready {
+					return nil
+				}
+				if !conn.WaitForStateChange(ctx, state) {
+					return errors.New("couldn't connect to core endpoint")
+				}
```
Wondertan, line 107:
```suggestion
			for conn.GetState() != connectivity.Ready {
				if !conn.WaitForStateChange(ctx, state) {
					return errors.New("couldn't connect to core endpoint")
				}
```

PR #4815, opened 2026-03-02
ci: selective unit testing based on changed packages

walldiss:
## Summary
- Adds `scripts/affected-packages.sh`: computes which Go packages need testing based on what files changed in the PR
- On `pull_request` events: only packages that changed, plus any package that (transitively) imports a changed package, are tested
- On push to `main`: full test suite runs unchanged
- Automatic fallback to full suite when `go.mod`/`go.sum` changes

## How it works
1. Collect changed `.go` files vs `github.base_ref`
2. Map each file to its Go package
3. BFS over the import graph to find all packages that depend on any changed package (reverse transitive closure)
4. Run `go test` on that set only

## Fallback conditions (runs all packages)
- `go.mod` or `go.sum` changed — dependency graph may have shifted
- No Go files changed — non-code PR, safest to run everything
- Any push to `main` — always validate the full suite

Untested

---

<a href="https://app.devin.ai/review/celestiaorg/celestia-node/pull/4815" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

Closes: https://linear.app/celestia/issue/PROTOCO-1307/ci-selective-unit-testing-based-on-changed-packages

### DIFF
--- a/scripts/affected-packages.sh
+++ b/scripts/affected-packages.sh
@@ -1,0 +1,77 @@
+#!/usr/bin/env bash
+# Prints the set of Go packages affected by changes relative to a base ref.
+# Falls back to all packages if dependencies changed or no Go files changed.
+#
+# Usage: ./scripts/affected-packages.sh <base-ref> [exclude-pattern]
+# Example: ./scripts/affected-packages.sh origin/main nodebuilder/tests
+
+set -euo pipefail
+
+BASE_REF=${1:-origin/main}
+EXCLUDE=${2:-nodebuilder/tests}
+
+ALL_PKGS=$(go list ./... | grep -v "$EXCLUDE")
+
+# If go.mod/go.sum changed, dependency graph may have changed — run everything
+if git diff --name-only "${BASE_REF}..HEAD" | grep -qE '^go\.(mod|sum)$'; then
+  echo "go.mod/go.sum changed: running all packages" >&2
+  echo "$ALL_PKGS"
+  exit 0
+fi
+
+# Get changed .go files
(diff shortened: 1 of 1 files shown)


scripts/affected-packages.sh
```diff
@@ -0,0 +1,78 @@
...
+fi
+
+# Get changed .go files
+CHANGED_GO=$(git diff --name-only "${BASE_REF}..HEAD" -- '*.go')
+
+if [ -z "$CHANGED_GO" ]; then
+  echo "No Go files changed: running all packages" >&2
+  echo "$ALL_PKGS"
+  exit 0
+fi
```
rach-id, line 28: if no go file changes, shouldn't this not run any package?

scripts/affected-packages.sh
```diff
@@ -0,0 +1,78 @@
...
+  | while read -r dir; do
+      go list "./$dir" 2>/dev/null || true
+    done \
+  | sort -u)
+
+if [ -z "$CHANGED_PKGS" ]; then
+  echo "No Go packages resolved: running all packages" >&2
+  echo "$ALL_PKGS"
+  exit 0
+fi
```
rach-id, line 42: same

Wondertan:
One worrying thing is that a dependant package outside could break too if the current packaged happened to not cover some case

scripts/affected-packages.sh
walldiss: True, good catch and it was intentional. Not to run any if no changes to go files.

walldiss: fixed

PR #4820, opened 2026-03-04
fix(das): prevent bridge DASer from receiving gossip headers before EDS is stored

### LINKED ISSUE #4802: fix(das): bridge node DASer races p2p gossip against core listener EDS storage
### Implementation ideas

  Bridge nodes log a flood of shrex WARN messages at every new block height:

  ```WARN shrex shrex/client.go:63 requesting data from peer failed ... ```

  ~590 occurrences per node per ~1.5 hours, repeating at live tip height every ~6 seconds.

Root cause

  PR #4776 enabled the DASer on bridge nodes by removing the tp == node.Bridge guard in nodebuilder/das/module.go. It also added a P2P exchange
  fallback to core.Exchange so the bridge can fetch headers from p2p when the core endpoint is unavailable.

  The DASer subscribes to libhead.Subscriber[*header.ExtendedHeader], which on both light and bridge nodes is backed by the p2p.Subscriber — it receives headers from the gossipsub network and also from local core listener that publishes headers with hook to broadcaster.

  The core Listener flow is:
  1. Receive block from core gRPC
  2. Construct EDS + extended header
  3. Store EDS locally
  4. Broadcast header to gossipsub network (with hook)

Remote bridge nodes receive the gossip broadcast and re-propagate it. The local DASer can receive the header from a remote peer before step 3 completes locally. Which creates race condition on the store due to 2 instances of header.Subscription being not synced. At that point HasByHeight returns false, so SharesAvailable falls through to getter.GetEDS, which goes through the shrex getter and logs the WARN.

walldiss:
## Summary

Fixes #4802

Bridge nodes subscribe to the p2p header gossip topic. Remote peers can gossip the same header **before the local `core.Listener` finishes storing EDS**, causing the DASer to look up a header it just received, call `HasByHeight`, fall through to `GetEDS` over shrex, and log a flood of WARN messages at every block height:

```
WARN	das	requesting data from peer failed	...
```

**Fix:** Add `p2p.WithTopicOpts(pubsub.FanoutOnly())` to the bridge node's `p2p.Subscriber`. With `FanoutOnly`, the subscriber can still **broadcast** (gossip headers out), but it **never receives** incoming gossip messages. The DASer only processes headers that the local `core.Listener` produced, guaranteeing EDS is already in the store.

Implementation notes:
- The `fx.Decorate` for the bridge's `p2p.Subscriber` wraps with `fx.Annotate(..., fx.OnStart, fx.OnStop)` — without explicit lifecycle hooks the decorated instance's `Start()` is never called, leaving `s.topic = nil` and causing a panic in `Broadcast`.
- Adds `replace github.com/libp2p/go-libp2p-pubsub => github.com/celestiaorg/go-libp2p-pubsub` required for `pubsub.FanoutOnly()` support.
- Replaces mocknet-based pubsub tests with real libp2p hosts for more realistic coverage.

Closes: https://linear.app/celestia/issue/DA-1072

---

<a href="https://app.devin.ai/review/celestiaorg/celestia-node/pull/4820" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/nodebuilder/header/module.go
+++ b/nodebuilder/header/module.go
@@ -22,6 +22,33 @@ import (
 
 var log = logging.Logger("module/header")
 
+// newSubscriber constructs the fx.Option for the p2p.Subscriber component.
+// Bridge nodes use FanoutOnly mode to prevent receiving gossip headers before
+// the local core.Listener has finished storing the corresponding EDS.
+func newSubscriber[H libhead.Header[H]](tp node.Type) fx.Option {
+	return fx.Provide(fx.Annotate(
+		func(ps *pubsub.PubSub, network modp2p.Network) (*p2p.Subscriber[H], error) {
+			opts := []p2p.SubscriberOption{p2p.WithSubscriberNetworkID(network.String())}
+			// Bridge nodes must not receive headers via p2p gossip: a remote peer
+			// can gossip the same header before the local core.Listener finishes
+			// storing EDS, causing the DASer to access EDS prematurely.
+			if tp == node.Bridge {
+				opts = append(opts, p2p.WithTopicOpts(pubsub.FanoutOnly()))
+			}
+			if MetricsEnabled {
+				opts = append(opts, p2p.WithSubscriberMetrics())
+			}
+			return p2p.NewSubscriber[H](ps, header.MsgID, opts...)
+		},
+		fx.OnStart(func(ctx context.Context, sub *p2p.Subscriber[H]) error {
+			return sub.Start(ctx)
+		}),
+		fx.OnStop(func(ctx context.Context, sub *p2p.Subscriber[H]) error {
+			return sub.Stop(ctx)
+		}),
+	))
+}
+
 func ConstructModule[H libhead.Header[H]](tp node.Type, cfg *Config) fx.Option {
 	// sanitize config values before constructing module
 	cfgErr := cfg.Validate(tp)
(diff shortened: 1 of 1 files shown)


nodebuilder/header/module.go
```diff
@@ -114,6 +114,29 @@ func ConstructModule[H libhead.Header[H]](tp node.Type, cfg *Config) fx.Option {
 		return fx.Module(
 			"header",
 			baseComponents,
+			// Bridge nodes only broadcast headers; they must not receive headers via
+			// p2p gossip because a remote peer can gossip the same header before the
+			// local core Listener has finished storing EDS. FanoutOnly prevents any
+			// incoming gossip messages so the DASer only sees headers it produced
+			// locally, avoiding premature EDS access.
+			fx.Decorate(fx.Annotate(
```
Wondertan: While decorate works, it seems rather complicated when we could do a single if around the subscriber construction.

nodebuilder/header/module.go
```diff
@@ -59,6 +59,12 @@ func ConstructModule[H libhead.Header[H]](tp node.Type, cfg *Config) fx.Option {
 		fx.Provide(fx.Annotate(
 			func(ps *pubsub.PubSub, network modp2p.Network) (*p2p.Subscriber[H], error) {
 				opts := []p2p.SubscriberOption{p2p.WithSubscriberNetworkID(network.String())}
+				// Bridge nodes must not receive headers via p2p gossip: a remote peer
+				// can gossip the same header before the local core.Listener finishes
+				// storing EDS, causing the DASer to access EDS prematurely.
+				if tp == node.Bridge {
+					opts = append(opts, p2p.WithTopicOpts(pubsub.FanoutOnly()))
+				}
```
Wondertan: i would also extract that in a separate func that returns fx.Option and take node type, but non blocking

vgonkivs (approved):

PR #4843, opened 2026-03-11
Add ADR for Fibre API design

cmwaters:
[Rendered](https://github.com/celestiaorg/celestia-node/blob/cal/fibre-api-adr/docs/adr/adr-013-fibre-api.md)

Summary
- Document the hybrid Fibre API proposal that extends the existing blob module while adding a dedicated Fibre module for escrow and data-plane workflows
- Define behaviour changes for blob Submit/GetAll/Subscribe and introduce new Fibre module types for upload, submission, retrieval, and escrow account management
- Capture rationale, consequences, and alternatives to provide context for future implementation

Testing
- Not run (not requested)

---

<a href="https://app.devin.ai/review/celestiaorg/celestia-node/pull/4843" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


docs/adr/adr-013-fibre-api.md
```diff
@@ -0,0 +1,336 @@
...
+`1`, and `2`:
+
+```go
+GetAll(ctx, height, namespaces)
+```
+
+For share version `2`, the blob module returns the on-chain system-level Fibre
+blob, because that is what is actually committed into the data square. This
+keeps `GetAll` consistent with its current role as a namespaced on-chain blob
+enumeration API.
```
vgonkivs, line 115: The ADR should specify extending the Blob type with IsFibreBlob() and FibreCommitment() methods, so callers of GetAll/Subscribe can identify fibre blobs and pass the commitment to fibre.Get without manually parsing the system blob layout.

cmwaters: Feel free to add that as helper methods. The user should be able to identify the fibre blob based on the share version. 

@rootulp correct me if I am wrong, but there are methods to extract the meta data from the fibre system level blob

docs/adr/adr-013-fibre-api.md
```diff
@@ -0,0 +1,336 @@
...
+The blob module should absorb Fibre support where callers are still working with
+chain-visible blobs.
+
+#### Submit
+
+`blob.Submit` continues to accept a list of blobs and routes behavior by share
+version:
+
+```go
+Submit(ctx, blobs, opts)
```
walldiss, line 85: Need to define returned values. Depending on what would be returned in payload current Submit may or may not fit. Things to consider as returns: PFF, PFF inclusion proof, RLC, UploadResult

docs/adr/adr-013-fibre-api.md
```diff
@@ -0,0 +1,336 @@
...
+For share version `2`, the blob module returns the on-chain system-level Fibre
+blob, because that is what is actually committed into the data square. This
+keeps `GetAll` consistent with its current role as a namespaced on-chain blob
+enumeration API.
+
+#### Subscribe
+
+`blob.Subscribe` should also support mixed namespaces and emit share version `2`
+blobs when a `MsgPayForFibre` results in a Fibre system blob being included:
+
```
walldiss, line 121: Alternative could be SubscribeFibre, which filters internally and returns only fibre blobs. General Subscribe is better for clients who need both fibre and non-fibre blobs from same subscription, while more specialised one is easier to use for those who knows they need only fibre. 

Same logic can be applied to GetAll, which may also have filtering option. 

I personally would like to hide share versionning detail through API lense, since it is yet another detail users need to learn about protocol internals, and would be great to hide it

docs/adr/adr-013-fibre-api.md
```diff
@@ -0,0 +1,336 @@
...
+    ValsetHeight      uint64
+    CreationTimestamp time.Time
+    Signature         []byte
+}
+
+type Module interface {
+    Upload(ctx context.Context, ns Namespace, data []byte) (UploadResult, error)
+    Submit(ctx context.Context, ns Namespace, data []byte) (SubmitResult, error)
+    Get(ctx context.Context, ns Namespace, commitment Commitment) ([]byte, error)
+    Account() AccountModule
```
walldiss, line 191: I'm slightly confused whether user is supposed to Upload through fibre.Upload, fibre.Submit and blob.Submit or both? It looks like you suggest blob.Submit to be compatible too. Imo we should keep only one

cmwaters: fibre.Upload and fibre.Submit are semantically quite different and target different use cases so we can't collapse those.

fibre.Submit and blob.Submit are intentionally named the same because they do the same thing. It makes it easier for the user to intuit what fibre.Submit does. We support submission through blob.Submit because it's plausible for users to expect that submitting a v2 blob should work and it's trivial to support.

The docs would only mention using the fibre methods

docs/adr/adr-013-fibre-api.md
cmwaters: I don't think the filter is necessary. If I am running a centralised sequencer that's just submitting fibre blobs then there should just be fibre blobs. If we were going to filter, I would suggest filtering by author but that's separate. In practice, I think it's more likely that users switch between different blob types depending on criteria (how much they need to submit and how frequently)

docs/adr/adr-013-fibre-api.md
cmwaters: My understanding is that changing the return values, breaks the go client and thus I hadn't planned to change the API at all

docs/adr/adr-013-fibre-api.md
rootulp: Correct for `FibreCommitment()` and `FibreBlobVersion()`

Ref:
- https://github.com/celestiaorg/go-square/blob/a9f9a81b671fbf42a0a4ef381e1993bf2c3a38ab/share/blob.go#L205-L211
- https://github.com/celestiaorg/go-square/blob/a9f9a81b671fbf42a0a4ef381e1993bf2c3a38ab/share/blob.go#L197-L203

I don't see a method in that file for `IsFibreBlob` so created https://github.com/celestiaorg/go-square/pull/256

rootulp:
I noticed integration tests failed on this PR even though it's only modifying a Markdown file so created https://github.com/celestiaorg/celestia-node/issues/4845

docs/adr/adr-013-fibre-api.md
walldiss: How do you see existing Submit being used for Fibre? It does not provide most of the fields native fibre moduless can.

docs/adr/adr-013-fibre-api.md
walldiss: I agree and thats why I don't think users should run existing subsription, because it forces them to do check share_ver == 2 on every blob returned from it. But we can just give them SubscribeFibre and they would no longer need to think about concept of share version

docs/adr/adr-013-fibre-api.md
walldiss: I don't like idea of implementing something and not documenting it. Imo we should try to reduce the scope of Fibre API as much as possible. Undocumented behaviour is bad, so I would rather suggest the opposite -  blocking fibre submits from blob module

walldiss: Submit seems to be actually both Upload + SubmitPFF. And there is no SubmitPFF API. Which makes me think api suggests to allow users to submit to fibre blobs without posting PFF. Why would we want this to be valid usecase? Invisible fibre blobs?

vgonkivs: What if instead of having two separate blob-like modules (blob + fibre), we have a single blob module that acts as an orchestrator over two internal clients — DABlobClient for v0/v1 PFB flow and FibreBlobClient for v2 fibre flow? `blob.Submit` routes by share version to the appropriate client and returns an extended SubmitResult struct where fibre-specific fields (Commitment, ValidatorSignatures, PaymentPromise, etc.) are nil for legacy blobs. Each client  encapsulates its own submission logic independently, while the orchestrator provides a single user-facing API. Fibre-unique operations like Upload (off-chain only), Get by commitment, and escrow management would be exposed directly on the blob module interface and routed to the fibre client internally. This will break the existing blob module API (return types change), but on the other hand it provides a much clearer UX and cleaner architecture going forward.

```go
  type SubmitResult struct {
      Height uint64
      TxHash string
      // Fibre-specific fields, nil for v0/v1 blobs
      Fibre  *FibreResult
  }

  type FibreResult struct {
      Commitment          Commitment
      ValidatorSignatures []ValidatorSignature
      PaymentPromise      *PaymentPromise
      RetentionUntil      *time.Time
  }
```

docs/adr/adr-013-fibre-api.md
```diff
@@ -0,0 +1,336 @@
...
+    RowVersion        uint32
+    ValsetHeight      uint64
+    CreationTimestamp time.Time
+    Signature         []byte
+}
+
+type Module interface {
+    Upload(ctx context.Context, ns Namespace, data []byte) (UploadResult, error)
+    Submit(ctx context.Context, ns Namespace, data []byte) (SubmitResult, error)
+    Get(ctx context.Context, ns Namespace, commitment Commitment) ([]byte, error)
```
vgonkivs, line 190: should return `fibre.Blob` I guess

docs/adr/adr-013-fibre-api.md
cmwaters: If you don't care about the return info beyond the height (or will derive it locally), then a user may still opt to using blob.Submit. There are quite a few that just fire and forget.

docs/adr/adr-013-fibre-api.md
cmwaters: So you're proposing having both blob.Subscribe and fibre.Subscribe?

docs/adr/adr-013-fibre-api.md
cmwaters: What extra information would you want in the blob? the signer?

docs/adr/adr-013-fibre-api.md
cmwaters: To @vgonkivs' proposal, is it really that hard for a user that wants to do both v2 (fibre) and v0 and v1 blobs that they can switch between fibre.Submit and blob.Submit? 

> I don't like idea of implementing something and not documenting it

It can also be documented then. The point was to avoid having this foot gun - people assuming bumping the share version should be valid then it errors and in the ideal case allowing users to use a single API for different versions, which would have been fine if more info than just the height was returned.

> Why would we want this to be valid usecase? Invisible fibre blobs?

The rollup wants to get ordering of their transactions using another mechanism that's not Celestia consensus - like Ethereum consensus or if we manage it, getting ordering from Fibre only and thus not needing to submit to Celestia

docs/adr/adr-013-fibre-api.md
walldiss: yes. And expect fibre users to use only the fibre.Subscribe.

docs/adr/adr-013-fibre-api.md
walldiss: Imo it is unnecessary functionality given there are will be native fibre methods with better UX.

cmwaters: If you are switching between different share versions, it's better UX to use the same method than have to switch between both.

If you want more info about your submission, it's better UX to use the fibre specific methods

It's also bad UX to have a protocol support a share version but blob.Submit doesn't. It's counter intuitive and a possible footgun

docs/adr/adr-013-fibre-api.md
cmwaters: > because it forces them to do check share_ver == 2 on every blob returned from it

But why would they enforce fibre only? At the end of the day, they just care about the bytes (and sometimes the signer)

cmwaters: I don't think any team would want to force fibre (especially given the minimum size is 256KB), but if you think this is valuable we can add it

docs/adr/adr-013-fibre-api.md
vgonkivs: I meant, `Get` should return the blob itself and not `[]byte`, no?

cmwaters: I understand that, but you already know the namespace and the commitment. What extra information should be returned to you that you don't already have?

PR #4864, opened 2026-03-18
feat(api) start blob sub from height instead of tip of the chain

### LINKED ISSUE #4716: api: start blob sub from height instead of tip of the chain
### Implementation ideas

Current blob.Subscribe does not allow starting from a specific height, which makes it hard for users to start from that height or resume after the subscription breaks.

The implementation should be additive, without breaking node API

ninabarbakadze:
fixes https://github.com/celestiaorg/celestia-node/issues/4716
fixes https://linear.app/celestia/issue/PROTOCO-931/api-start-blob-sub-from-height-instead-of-tip-of-the-chain

---

<a href="https://app.devin.ai/review/celestiaorg/celestia-node/pull/4864" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


blob/service.go
```diff
@@ -121,61 +121,204 @@ func (s *Service) Subscribe(ctx context.Context, ns libshare.Namespace) (<-chan
...
+// SubscribeWithStartHeight returns a channel that will receive SubscriptionResponse objects
+// starting from the given startHeight. It first replays all historical blobs from startHeight
+// up to the current network head (catchup), then seamlessly transitions to streaming
+// new blobs as they arrive (live).
+//
+// Retrieval errors during catchup (e.g. pruned data) will close the subscription.
+// During the live phase, retrieval errors are retried until successful.
+// The channel will be closed when the context is canceled or the service is stopped.
+// Not reading from the returned channel will cause the stream to close after 16 messages.
+func (s *Service) SubscribeWithStartHeight(
```
Wondertan: SubsribeFrom

ninabarbakadze: still in draft sir pls

Wondertan: doesn't mean i cant comment

Wondertan: see, another comment

blob/service.go
```diff
@@ -121,61 +121,204 @@ func (s *Service) Subscribe(ctx context.Context, ns libshare.Namespace) (<-chan
...
-					Blobs:  blobs,
-					Height: header.Height(),
-					Header: &header.RawHeader,
-				}:
+	log.Infow("subscribing for blobs with start height",
+		"namespace", ns.String(),
+		"startHeight", startHeight,
+	)
+
+	blobCh := make(chan *SubscriptionResponse, 16)
```
ninabarbakadze: we might want to increase this buffer size as the catchup blobs will arrive much faster than live subscription. I'll defer to the reviewer on this.

blob/service.go
```diff
@@ -111,71 +115,208 @@ func (s *Service) Subscribe(ctx context.Context, ns libshare.Namespace) (<-chan
...
 						return
 					}
-					if err == nil {
-						// operation successful, break the loop
-						break
+					gapHeader, err := s.headerGetter(ctx, expectedNextHeight)
+					if err != nil {
+						log.Errorw("blobsub: failed to fill gap between catchup and live",
+							"namespace", ns.ID(), "height", expectedNextHeight, "err", err)
+						return
```
walldiss: **Bug: gap-fill treats all `headerGetter` errors as fatal, but "syncing in progress" is transient.**

Scenario:
1. Catchup processes heights 1..100, height 101 returns "syncing in progress" (store head is 100, syncer head is 200)
2. `isBeyondTip` returns true, catchup ends with `expectedNextHeight = 101`
3. `headerSub` starts, first live header arrives at height 200
4. Gap-fill calls `headerGetter(ctx, 101)` — store still hasn't synced to 101
5. Returns "syncing in progress" again → subscription permanently closed

This needs a retry with backoff when `isBeyondTip(err)` is true, similar to:
```go
gapHeader, err := s.headerGetter(ctx, expectedNextHeight)
if err != nil {
    if isBeyondTip(err) {
        // store hasn't caught up yet, wait and retry
        select {
        case <-ctx.Done(): return
        case <-time.After(time.Second): continue
        }
    }
    // ...
}
```
Alternatively, consider using a `WaitForHeight`-style call that blocks until the header is available.

blob/service.go
```diff
@@ -111,71 +115,208 @@ func (s *Service) Subscribe(ctx context.Context, ns libshare.Namespace) (<-chan
...
+				log.Errorw("blobsub: failed to retrieve blobs during catchup",
+					"namespace", ns.ID(), "height", height, "err", err)
+				return
+			}
+			select {
+			case <-ctx.Done():
+				return
+			case <-s.ctx.Done():
+				return
+			case blobCh <- &SubscriptionResponse{Blobs: blobs, Height: header.Height(), Header: &header.RawHeader}:
```
walldiss: **No overflow check during catchup sends — inconsistent with live phase.**

During live streaming (line 263), the subscription is closed when the buffer is full. But here during catchup, the send is blocking — a slow reader will stall the goroutine indefinitely rather than closing.

This might be intentional (catchup data is already available, so blocking is reasonable), but the behavior difference should be documented in the godoc. Currently the doc says "Not reading from the returned channel during live subscription will cause the stream to close after 16 messages" — add a note that during catchup, sends are blocking.

blob/service.go
```diff
@@ -111,71 +115,208 @@ func (s *Service) Subscribe(ctx context.Context, ns libshare.Namespace) (<-chan
...
+// SubscribeFromStartHeight returns a channel that will receive SubscriptionResponse objects
+// starting from the given startHeight. It first replays all historical blobs from startHeight
+// up to the current network head (catchup), then transitions to streaming new blobs as they
+// arrive (live).
+//
+// Retrieval errors during catchup (e.g. pruned data) will close the subscription.
+// During the live phase, retrieval errors are retried until successful.
+// The channel will be closed when the context is canceled or the service is stopped.
+// Not reading from the returned channel during live subscription will cause the stream
+// to close after 16 messages.
```
walldiss: **Godoc should document recovery pattern.**

The docs explain the overflow behavior but not how users should handle it. Since `SubscriptionResponse` includes `Height`, clients can track the last received height and re-subscribe from `lastHeight + 1`. Worth adding a note like:

```
// If the channel is closed due to buffer overflow, clients can re-subscribe
// from the last received height + 1 to resume without gaps.
```

blob/service.go
```diff
@@ -111,71 +115,208 @@ func (s *Service) Subscribe(ctx context.Context, ns libshare.Namespace) (<-chan
...
 		}
 	}()
 	return blobCh, nil
 }
 
+// isBeyondTip returns true if the error from headerGetter indicates the requested
+// height is beyond the current network or local head.
+func isBeyondTip(err error) bool {
+	msg := err.Error()
+	return strings.Contains(msg, "from the future") || strings.Contains(msg, "syncing in progress")
```
walldiss: **Fragile string matching — will silently break if error messages change.**

This is coupled to specific error strings in `nodebuilder/header/service.go`. If those messages are ever changed, this breaks silently and catchup will treat "beyond tip" as a fatal error.

Ideal fix: export sentinel errors or error types from the header package that can be matched with `errors.Is`/`errors.As`.

Minimal fix: add a comment noting the coupling, e.g.:
```go
// NOTE: coupled to error messages in nodebuilder/header/service.go:107,132
```

blob/service.go
```diff
@@ -111,71 +115,208 @@ func (s *Service) Subscribe(ctx context.Context, ns libshare.Namespace) (<-chan
...
+// and sends the response on blobCh. Returns false if the caller should stop.
+func (s *Service) fetchAndSendBlobs(
+	ctx context.Context,
+	blobCh chan *SubscriptionResponse,
+	header *header.ExtendedHeader,
+	ns libshare.Namespace,
+) bool {
+	var blobs []*Blob
+	var err error
+	for {
```
walldiss, line 231: **Tight infinite retry loop with no backoff or logging.**

If `getAll` consistently fails (e.g. network partition), this spins at 100% CPU with no visibility. Consider:
1. Logging the error (at least at Debug level)
2. Adding a short backoff (`time.After(100ms)` or exponential)

This is pre-existing behavior from the original `Subscribe`, but worth fixing now that it's extracted into a shared helper.

blob/service.go
```diff
@@ -111,71 +115,208 @@ func (s *Service) Subscribe(ctx context.Context, ns libshare.Namespace) (<-chan
...
-					if err == nil {
-						// operation successful, break the loop
-						break
+					gapHeader, err := s.headerGetter(ctx, expectedNextHeight)
+					if err != nil {
+						log.Errorw("blobsub: failed to fill gap between catchup and live",
+							"namespace", ns.ID(), "height", expectedNextHeight, "err", err)
+						return
 					}
+					if !s.fetchAndSendBlobs(ctx, blobCh, gapHeader, ns) {
```
walldiss: **Gap-fill sends via `fetchAndSendBlobs` bypass the buffer overflow check.**

The overflow check at line 263 runs once per live header, but this loop can emit multiple gap-fill responses before reaching it. If the gap is large and the reader is slow, the buffer could fill beyond the check. Consider adding an overflow check inside this loop as well.

walldiss (changes requested):
Nice work overall — the catchup→live design is solid and the code is well-structured. A few issues to address, one of which is a bug in the gap-fill path.

blob/service.go
ninabarbakadze: Yes, this was intentional because catchup will be much faster than live subscription so I'll document it.

blob/service.go
ninabarbakadze: yes! I was going to ask if we wanted to ad an error type directly in the header package(IIRC this is where headerGetter lives)

blob/service.go
ninabarbakadze: Overflow check being that we disconnect if buffer is full? I wouldn't disconnect during gap-fill because similar to catch-up the buffer might fill up quickly because of fast writes and slower reads.

blob/service.go
ninabarbakadze: Resolving this since we moved away form catchup design.

blob/service.go
ninabarbakadze: Resolving this since we moved away form catchup design.

blob/service.go
ninabarbakadze: Resolving this since we moved away form catchup design.

blob/service.go
ninabarbakadze: Resolving because this method was removed.

blob/service.go
ninabarbakadze: backoff + debug https://github.com/celestiaorg/celestia-node/pull/4864/changes/67e27344429ab5c98f578d09467987c38ab8728a

walldiss:
Hey @ninabarbakadze , I owe you a real apology. I knew #4864 (this PR)  was open, I was supposed to be the main reviewer, and we'd talked through it plenty of times. Last week I needed Subscribe-from-height for the Fibre work on a tight timeline and just shipped #4962 to `feature/fibre` without coming back to your PR first.

Reading both side by side now, your design is the one we should land. I want to revert the breaking `Subscribe` change I made and adopt your additive `SubscribeFromStartHeight` shape, that's the right call for an RPC API.

Could you open a new PR against `feature/fibre` that:
- restores the original `Subscribe(ctx, ns)` (no `fromHeight` arg)
- adds `SubscribeFromStartHeight(ctx, ns, startHeight)` as in your original PR
- brings forward the `HeaderService` interface, upfront startHeight validation, retry backoff, and your test suite (including the atomic-tip mock — that path has zero tests right now)

One small thing worth keeping from #4962: the cross-context cancel using `context.AfterFunc` is cleaner than the extra goroutine. Otherwise it's your design.

I'll handle the matching update on the ev-node side. Once your fibre PR lands, close #4864 with a pointer to it. I'll prioritize the review this time.

ninabarbakadze:
> Hey @ninabarbakadze , I owe you a real apology. I knew #4864 (this PR) was open, I was supposed to be the main reviewer, and we'd talked through it plenty of times. Last week I needed Subscribe-from-height for the Fibre work on a tight timeline and just shipped #4962 to `feature/fibre` without coming back to your PR first.
> 
> Reading both side by side now, your design is the one we should land. I want to revert the breaking `Subscribe` change I made and adopt your additive `SubscribeFromStartHeight` shape, that's the right call for an RPC API.
> 
> Could you open a new PR against `feature/fibre` that:
> 
> * restores the original `Subscribe(ctx, ns)` (no `fromHeight` arg)
> * adds `SubscribeFromStartHeight(ctx, ns, startHeight)` as in your original PR
> * brings forward the `HeaderService` interface, upfront startHeight validation, retry backoff, and your test suite (including the atomic-tip mock — that path has zero tests right now)
> 
> One small thing worth keeping from #4962: the cross-context cancel using `context.AfterFunc` is cleaner than the extra goroutine. Otherwise it's your design.
> 
> I'll handle the matching update on the ev-node side. Once your fibre PR lands, close #4864 with a pointer to it. I'll prioritize the review this time.

I just saw your comment. I'll create a new PR against the Fiber branch but do we want to merge this to main?

ninabarbakadze:
closing this in favour of https://github.com/celestiaorg/celestia-node/pull/4990 as per async discussion w/ @walldiss

PR #4892, opened 2026-03-25
feat(fibre): implement Fibre Module

vgonkivs:
Resolves https://linear.app/celestia/issue/PROTOCO-1332/implement-fibre-module

---

<a href="https://app.devin.ai/review/celestiaorg/celestia-node/pull/4892" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/fibre/service.go
+++ b/fibre/service.go
@@ -47,7 +47,11 @@ func (s *Service) Submit(
 	ns libshare.Namespace,
 	data []byte,
 	options *txclient.TxConfig,
-) (_ *user.TxResponse, _ *appfibre.SignedPaymentPromise, _ appfibre.BlobID, err error) {
+) (_ *user.TxResponse, _ *appfibre.SignedPaymentPromise, err error) {
+	if s == nil {
+		return nil, nil, ErrClientNotAvailable
+	}
+
 	start := time.Now()
 	defer func() {
 		s.metrics.observeSubmit(ctx, time.Since(start), len(data), err)
@@ -57,23 +61,23 @@ func (s *Service) Submit(
 
 	blob, err := appfibre.NewBlob(data, appfibre.DefaultBlobConfigV0())
 	if err != nil {
-		return nil, nil, nil, err
+		return nil, nil, err
 	}
 
 	promise, err := s.upload(ctx, ns, blob)
 	if err != nil {
 		log.Errorw("uploading blob", "err", err, "namespace", ns.ID())
-		return nil, nil, nil, err
+		return nil, nil, err
 	}
 
 	protoPromise, err := promise.ToProto()
 	if err != nil {
-		return nil, nil, nil, err
+		return nil, nil, err
 	}
 
 	signer, err := s.txClient.GetTxAuthorAccAddress(options)
 	if err != nil {
-		return nil, nil, nil, fmt.Errorf("getting signer address: %w", err)
+		return nil, nil, fmt.Errorf("getting signer address: %w", err)
 	}
 
 	msg := &fibretypes.MsgPayForFibre{
@@ -84,7 +88,7 @@ func (s *Service) Submit(
 	resp, err := s.txClient.SubmitMessage(ctx, msg, options)
 	if err != nil {
 		log.Errorw("submitting blob", "err", err, "namespace", promise.Namespace)
-		return nil, nil, nil, err
+		return nil, nil, err
 	}
 
 	// go does not allow slicing a function return value directly (e.g., f()[:]).
@@ -96,7 +100,7 @@ func (s *Service) Submit(
 		"tx-hash", resp.TxHash,
 		"signatures", len(promise.ValidatorSignatures),
 	)
-	return resp, promise, blob.ID(), nil
+	return resp, promise, nil
 }
 
 func (s *Service) Upload(
@@ -105,6 +109,10 @@ func (s *Service) Upload(
 	data []byte,
 	_ *txclient.TxConfig,
 ) (_ *appfibre.SignedPaymentPromise, _ appfibre.BlobID, err error) {
+	if s == nil {
+		return nil, nil, ErrClientNotAvailable
+	}
+
 	start := time.Now()
 	defer func() {
 		s.metrics.observeUpload(ctx, time.Since(start), len(data), err)
@@ -125,15 +133,17 @@ func (s *Service) Upload(
 	return promise, blob.ID(), nil
 }
 
-func (s *Service) Get(ctx context.Context, blobID []byte) (*appfibre.Blob, error) {
-	blbID := appfibre.BlobID(blobID)
-	err := blbID.Validate()
+func (s *Service) Download(ctx context.Context, blobID appfibre.BlobID) (*appfibre.Blob, error) {
+	if s == nil {
+		return nil, ErrClientNotAvailable
+	}
+
+	err := blobID.Validate()
 	if err != nil {
 		return nil, fmt.Errorf("invalid blob ID: %w", err)
 	}
-	log.Debugw("downloading fibre blob", "commitment", blbID.Commitment())
 
-	return s.fibreClient.Download(ctx, blbID)
+	return s.fibreClient.Download(ctx, blobID)
 }
 
 func (s *Service) upload(
--- a/nodebuilder/fibre/fibre.go
+++ b/nodebuilder/fibre/fibre.go
@@ -5,6 +5,7 @@ import (
 
 	sdktypes "github.com/cosmos/cosmos-sdk/types"
 
+	appfibre "github.com/celestiaorg/celestia-app/v8/fibre"
 	libshare "github.com/celestiaorg/go-square/v4/share"
 
 	"github.com/celestiaorg/celestia-node/fibre"
@@ -20,32 +21,34 @@ var _ Module = (*API)(nil)
 //
 //go:generate mockgen -destination=mocks/api.go -package=mocks . Module
 type Module interface {
-	// Submit submits a blob via the Fibre network.
+	// Submit submits a v0 Fibre blob via the Fibre network.
 	// It performs the full Fibre flow: uploads blob data to FSPs, aggregates validator
 	// availability signatures, and submits MsgPayForFibre on-chain.
 	// Returns the submission result including the on-chain height and transaction hash.
 	// Requires the node to be connected to a core endpoint with Fibre support.
-	Submit(_ context.Context, _ libshare.Namespace, _ []byte, _ *txclient.TxConfig) (*SubmitResult, error)
-	// Upload performs the off-chain portion of Fibre blob submission only.
+	// NOTE: Currently only v0 Fibre blobs are supported.
+	Submit(context.Context, libshare.Namespace, []byte, *txclient.TxConfig) (*SubmitResult, error)
+	// Upload performs the off-chain portion of v0 Fibre blob submission only.
 	// It encodes the blob, constructs a payment promise, uploads encoded rows to FSPs,
 	// and aggregates validator availability signatures. It does NOT submit MsgPayForFibre on-chain.
-	// Use blob.SubmitFibreBlob for the full submit flow.
-	Upload(_ context.Context, _ libshare.Namespace, _ []byte, _ *txclient.TxConfig) (*UploadResult, error)
-	// Get retrieves a Fibre blob from FSPs by blobID.
+	// Use fibre.Submit for the full submit flow.
+	// NOTE: Currently only v0 Fibre blobs are supported.
+	Upload(context.Context, libshare.Namespace, []byte, *txclient.TxConfig) (*UploadResult, error)
+	// Download retrieves a Fibre blob from FSPs by blobID.
 	// It reconstructs the original blob data from the encoded rows stored off-chain.
-	Get(ctx context.Context, blobID []byte) (*GetBlobResult, error)
+	Download(context.Context, appfibre.BlobID) (*GetBlobResult, error)
 	// QueryEscrowAccount returns the escrow account details for the given signer address,
 	// including total balance and available (spendable) balance.
-	QueryEscrowAccount(ctx context.Context, signer string) (*fibre.EscrowAccount, error)
+	QueryEscrowAccount(_ context.Context, signer string) (*fibre.EscrowAccount, error)
 	// Deposit adds funds to the node's Fibre escrow account.
 	// The signer is resolved from cfg (SignerAddress or KeyName) or the node's default account.
-	Deposit(ctx context.Context, amount sdktypes.Coin, cfg *txclient.TxConfig) error
+	Deposit(context.Context, sdktypes.Coin, *txclient.TxConfig) error
 	// Withdraw requests a withdrawal from the node's Fibre escrow account.
 	// The signer is resolved from cfg (SignerAddress or KeyName) or the node's default account.
 	// The withdrawal enters an unbonding period before funds become claimable.
-	Withdraw(ctx context.Context, amount sdktypes.Coin, cfg *txclient.TxConfig) error
+	Withdraw(context.Context, sdktypes.Coin, *txclient.TxConfig) error
 	// PendingWithdrawals returns all pending (not yet claimable) withdrawals for the given signer.
-	PendingWithdrawals(ctx context.Context, signer string) ([]fibre.PendingWithdrawal, error)
+	PendingWithdrawals(_ context.Context, signer string) ([]fibre.PendingWithdrawal, error)
 }
 
 // API is a wrapper around Module for the RPC.
@@ -63,9 +66,9 @@ type API struct {
 			data []byte,
 			config *txclient.TxConfig,
 		) (*UploadResult, error) `perm:"write"`
-		Get func(
+		Download func(
 			ctx context.Context,
-			blobID []byte,
+			blobID appfibre.BlobID,
 		) (*GetBlobResult, error) `perm:"read"`
 		QueryEscrowAccount func(
 			ctx context.Context,
@@ -106,8 +109,8 @@ func (api *API) Upload(
 	return api.Internal.Upload(ctx, ns, data, options)
 }
 
-func (api *API) Get(ctx context.Context, blobID []byte) (*GetBlobResult, error) {
-	return api.Internal.Get(ctx, blobID)
+func (api *API) Download(ctx context.Context, blobID appfibre.BlobID) (*GetBlobResult, error) {
+	return api.Internal.Download(ctx, blobID)
 }
 
 func (api *API) QueryEscrowAccount(ctx context.Context, signer string) (*fibre.EscrowAccount, error) {
--- a/nodebuilder/fibre/types.go
+++ b/nodebuilder/fibre/types.go
@@ -25,16 +25,11 @@ type UploadResult struct {
 // SubmitResult is the result of a full Fibre blob submission including on-chain MsgPayForFibre.
 // It extends UploadResult with chain inclusion details (height and transaction hash).
 type SubmitResult struct {
-	// BlobID is the Fibre blob identifier for the submitted blob.
-	BlobID appfibre.BlobID `json:"blob_id"`
-	// ValidatorSignatures are attestations from validators confirming blob availability.
-	ValidatorSignatures []ValidatorSignature `json:"validator_signatures"`
+	UploadResult
 	// Height is the block height at which MsgPayForFibre was included on-chain.
 	Height uint64 `json:"height"`
 	// TxHash is the transaction hash of the on-chain MsgPayForFibre.
 	TxHash string `json:"tx_hash"`
-	// PaymentPromise is the signed promise used for on-chain fee settlement.
-	PaymentPromise *PaymentPromise `json:"payment_promise,omitempty"`
 }
 
 // PaymentPromise represents the signed payment promise between a user and the Fibre network.
--- a/api/docgen/examples.go
+++ b/api/docgen/examples.go
@@ -23,6 +23,7 @@ import (
 	"golang.org/x/text/language"
 
 	appfibre "github.com/celestiaorg/celestia-app/v8/fibre"
+	libhead "github.com/celestiaorg/go-header"
 	libshare "github.com/celestiaorg/go-square/v4/share"
 	"github.com/celestiaorg/rsmt2d"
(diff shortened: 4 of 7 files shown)


nodebuilder/fibre/fibre.go
```diff
@@ -0,0 +1,101 @@
...
+// payment settlement. Full blob submission (upload + on-chain MsgPayForFibre) is available
+// through blob.SubmitFibreBlob; this module exposes upload-only, retrieval, and escrow operations.
+//
+//go:generate mockgen -destination=mocks/api.go -package=mocks . Module
+type Module interface {
+	// Upload performs the off-chain portion of Fibre blob submission only.
+	// It encodes the blob, constructs a payment promise, uploads encoded rows to FSPs,
+	// and aggregates validator availability signatures. It does NOT submit MsgPayForFibre on-chain.
+	// Use blob.SubmitFibreBlob for the full submit flow.
+	Upload(ctx context.Context, ns libshare.Namespace, data []byte) (*fibre.UploadResult, error)
```
Wondertan: interface definition should be together with types it uses for modules in case those types are API level heler types(Result/Response), instead of internal types(Blob)

Wondertan: This is UploadResult, but GetBlobResponse? Consistencty?

Wondertan (changes requested):
### There are two competing fibre blob submission paths, and one of them is incomplete

* `modfibre.Upload`

The newly introduced API, however, is incomplete as it doesn't send the PFF transactions. The comment mentions that users should use `modblob.SubmitFibreBlob` to send TXs; however, doing so requires reuploading the entire blob over the API, and `modblob.SubmitFibreBlob` reuploads it again. 

_Essentially, there is no way for fibre users of `Upload` to pay for their uploads_

* `modblob.SubmitFibreBlob`

The purpose of this method is unclear. It duplicates the entire fibre blob submission on the old blob service, however, the intention was clearly to introduce a new API as seen in the new fibre module, so why duplicate? If the intention was to allow sending PFFs through this module, then it is unclear why `blob` module would be responsible for that and not `fibre`.

### Too many layers of indirections 

With this PR we have:
* modfibre.Module
* `fibre.Client`
* `fibre.client`
* txclient fibre methods
* `appfibre.Client`

This is extremely messy and a pure pain to review. We can easily squash a bunch of them with no repercussions into:
* modfibre.Module
* `fibre.Service`
* `appfibre.Client`.

Besides, for whatever reason, out of all those layers, the txclient turned out to be responsible for actually uploading. TxClient, Carl! It is supposed to send transactions and not call `appfibre.Client.Upload`.

### Subscriptions

There is no way to subscribe for blobs from fibre in the order they'be landed through consensus.

fibre/metrics.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+func (m *blobMetrics) observeSubmit(ctx context.Context, dur time.Duration, blobSize int, err error) {
+	if m == nil {
+		return
+	}
+	m.submitDuration.Record(ctx, dur.Seconds(), blobAttrs(blobSize, err))
+}
+
+func (m *blobMetrics) observeGet(ctx context.Context, dur time.Duration, err error) {
```
Wondertan: Do we really need those metrics if they are already part of the `appfibre.Client`, besides the Submit, but that's just tx submission metric?

state/txclient/fibre.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	cl, err := appfibre.NewClient(c.keyring, appfibre.DefaultClientConfig())
+	if err != nil {
+		return fmt.Errorf("failed to setup fibre client: %w", err)
+	}
+	if err := cl.Start(c.ctx); err != nil {
+		return fmt.Errorf("failed to start fibre client: %w", err)
+	}
+	c.fibreClient = cl
```
Wondertan: Please provide fibre client as a component to DI with full lifecycling. Its Start method is blocking and is meant to called while node starts and break node start if there is misconfiguration, rather then being called in runtime

vgonkivs:
This PR was done in accordance to the ADR, made by @cmwaters. 

> There are two competing fibre blob submission paths, and one of them is incomplete
modfibre.Upload
The newly introduced API, however, is incomplete as it doesn't send the PFF transactions. The comment mentions that users should use modblob.SubmitFibreBlob to send TXs; however, doing so requires reuploading the entire blob over the API, and modblob.SubmitFibreBlob reuploads it again.

[Link1](https://github.com/celestiaorg/celestia-node/blob/e8c4f70dbea021a8a8253858427c51b134a3073c/docs/adr/adr-013-fibre-api.md?plain=1#L201)

[Link2](https://github.com/celestiaorg/celestia-node/pull/4843#discussion_r2925954123)


> Subscriptions
There is no way to subscribe for blobs from fibre in the order they'be landed through consensus.

 Auto-fetching full fibre data from FSPs needs design work first. Feel free to open an ADR update with a proposal

fibre/metrics.go
vgonkivs: These metrics track different layers. fibre/metrics.go measures end-to-end node-level latency (upload, submit, get),  txclient measures tx submission and gas estimation, and appfibre.Client tracks FSP networking internals. They complement each other — knowing that a submit took 5s total but only 200ms on tx submission tells you the bottleneck is in the upload phase.

Wondertan:
@vgonkivs, I acknowledge that the PR implements the ADR. However, it does not change the fact that users can't pay for the uploads they make and we should fix this.

> Auto-fetching full fibre data from FSPs needs design work first. Feel free to open an ADR update with a proposal

The subscription does not imply listening for data from FSPs, but listening to new fibre-blobs in the square and fetching respective fibre blobs by commitment.

fibre/fibre.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/celestiaorg/celestia-node/state"
+)
+
+// TxConfig aliases TxOptions from state package allowing users
+// to specify options for SubmitPFB transaction.
+type TxConfig = state.TxConfig
+
+// Commitment is a 32-byte Fibre blob commitment derived from the blob data.
+type Commitment [appfibre.CommitmentSize]byte
```
Wondertan: Let's aim to reuse the existing commitment type defined in appfibre, by extending to support Marshal methods. With this Commitment, there will be now 3 commitment types.

nodebuilder/fibre/types.go
```diff
@@ -0,0 +1,87 @@
...
+	TxHash string `json:"tx_hash"`
+	// PaymentPromise is the signed promise used for on-chain fee settlement.
+	PaymentPromise *PaymentPromise `json:"payment_promise,omitempty"`
+	// RetentionUntil is the time until which FSPs will retain the blob data.
+	RetentionUntil *time.Time `json:"retention_until,omitempty"`
+}
+
+// PaymentPromise represents the signed payment promise between a user and the Fibre network.
+// It is constructed during upload and submitted on-chain as part of MsgPayForFibre.
+type PaymentPromise struct {
```
Wondertan, line 37: Do we need to copy this type?

nodebuilder/fibre/types.go
```diff
@@ -0,0 +1,87 @@
...
+	// ValidatorSignatures are attestations from validators confirming blob availability.
+	ValidatorSignatures []ValidatorSignature `json:"validator_signatures"`
+	// Height is the block height at which MsgPayForFibre was included on-chain.
+	Height uint64 `json:"height"`
+	// TxHash is the transaction hash of the on-chain MsgPayForFibre.
+	TxHash string `json:"tx_hash"`
+	// PaymentPromise is the signed promise used for on-chain fee settlement.
+	PaymentPromise *PaymentPromise `json:"payment_promise,omitempty"`
+	// RetentionUntil is the time until which FSPs will retain the blob data.
+	RetentionUntil *time.Time `json:"retention_until,omitempty"`
```
Wondertan: We should mark somehow that this is not implemented and has no efffect or better remove entirely until this functionaility is implemented. IIRC, we don't include this feature for the mainnet cc @mcrakhman

Wondertan (commented):
Much, much better, cleaner, and less code.

fibre/metrics.go
Wondertan: For Submit this is true; however, the comment points to Get, which just calls Download, which is already covered with a metric.

fibre/fibre.go
Wondertan: I know this requires changes in the app, so can be a flup

nodebuilder/fibre/types.go
vgonkivs: we can remove it for now and include it in future.  Additive changes are not breaking so we should be good

nodebuilder/fibre/types.go
vgonkivs: no, lets keep the original type. I was wondering whether we want to hide something specific from the user but I guess the answer is no.

nodebuilder/fibre/constructor.go
```diff
@@ -0,0 +1,56 @@
...
+		fxutil.ProvideIf(!coreCfg.IsEndpointConfigured(), func() (*fibre.Service, Module) {
+			return nil, &stubbedFibreModule{}
+		}),
+	)
+}
+
+func newAppFibreClient(keyring keyring.Keyring, conn *grpc.ClientConn) (*appfibre.Client, error) {
+	cfg := appfibre.DefaultClientConfig()
+	cfg.StateAddress = conn.Target()
+	return appfibre.NewClient(keyring, cfg)
```
Wondertan, line 51: One thing I just realized is that this uses the default `state.Client`, which trusts the grpc endpoint entirely. The trust there is for (a) the validator sets and (b) the state of `HostRegistry`. 

This works and will work, but it goes against celestia-node's design as a trust-minimized light client/node. I would even argue there is no point in integrating _the fibre client_ into the node without making it trust minimized. If people want to use a trusted client, they can just use `appfibre.Client` directly.

To make this trust minimized, we would need to supply a different `StateClientFn` with a custom implementation of `state.Client` that (a) provides `validator.SetGetter` implementation over the header service(it would simply get `ExtendedHeader` by height and return valset in it) and (b) implementation of `validator.HostRegistry` that verifies merkle inclusion of retrieved state of validator addresses. `VerifyPromise`/`Start`/`Stop` would be left noop,

An additional reason for this is saving one round-trip. The default  `validator.SetGetter`  always makes a request for a val set to the remote GRPC app server, while the node keeps them locally with smart caching and follows/subscribes to new headers/valsets, unlike the default.

If I were the implementer, I would include this as part of the integration PR; however, I don't mind this being a flup as well, however, i do feel strongly that fibre integration should follow the light client header chain implemented within the node

vgonkivs: Your proposal makes sense for me but the PR is already quite big, so I'd rather implement it as flup

nodebuilder/fibre/types.go
vgonkivs: I have double checked the original PaymentPromise type and it has SignerKey `*secp256k1.PubKey` which is a cosmos-sdk protobuf type — it doesn't serialize cleanly to JSON and leaks cosmos dependencies into the RPC API contract. The wrapper lets us keep the API clean for  non-Go clients. wdyt @Wondertan ?

vgonkivs: In dm we agreed to keep the wrapper type bc we don't want to expose cosmos-sdk protobuf types in our external api.

fibre/metrics.go
vgonkivs: done

fibre/fibre.go
vgonkivs: removed custom Commitment type in favour of `appfibre.Commitment`

nodebuilder/fibre/types.go
vgonkivs: removed

nodebuilder/fibre/constructor.go
Wondertan: Let's park then the comment into issue, as well as, the upload leftover

vgonkivs: done

Wondertan: Which issue

vgonkivs: In linear.

fibre/service.go
```diff
@@ -0,0 +1,161 @@
...
+	return promise, nil
+}
+
+func (s *Service) Get(ctx context.Context, ns libshare.Namespace, commitment []byte) (*appfibre.Blob, error) {
+	log.Debugw("getting blob", "namespace", ns.ID(), "commitment", hex.EncodeToString(commitment))
+	if len(commitment) != appfibre.CommitmentSize {
+		return nil, fmt.Errorf("commitment size does not match. want:%d got:%d", appfibre.CommitmentSize, len(commitment))
+	}
+
+	blobID := appfibre.NewBlobID(ns.Version(), appfibre.Commitment(commitment))
```
Wondertan: The version of namespace != blob version.

We need to return BlobID in Upload/SubmitResult with the respective blob version to users, and change Get to have BlobId instead of commitment.

Wondertan (changes requested):

fibre/service.go
vgonkivs: yeah. right. nice catch. thank you

vgonkivs: So, one more thing worth mentioning here is the height. Do we want to get height from the user along with the blobID? Asking bc, with the regular blobs we had complaints from the rollup users that that have to track a pair `commitment -> height `internally(most of them do not care about height)

Wondertan: I don't have a preference. Passing height is optional to download but its not error prone without it. It's also ambigious API wise, because the expected height is not the one where tx landed and returned in SubmitResult, but the height on the payment promise.

mcrakhman: Ok, so speaking purely logically if the height is changed then new val set can have different nodes from previous one (note: if the set of nodes is the same, even if they have different stakes, then it is fine, because we will try to download from all of them in the end). 

So in case the nodes differ, depending on the extend of the change, one of the nodes excluded from the set can be the last one needed to reconstruct the blob (e.g. if we have 1/3 malicious, and then after val set change we can have less than 1/3 shards available in the val set). This is mostly theoretical, so I don't have a strong opinion why the height should be included, provided we state our assumptions as defined below. Also if that becomes critical, it seems we can always introduce it later and older clients can provide some sentinel parameter for height, meaning that we should just use current one (also retention period is quite small, so it shouldn't be a problem anyways). 

But we should still add the reasoning as to why we don't include it. I am not sure about exact wording, but it can be something along the lines: "we assume that the validator set won't change to the extent that will affect the ability to download the blobs uploaded before the val set change during the respective retention period".

vgonkivs: I guess the explanation should be added in the ADR. I'm planing to update it in the end when everything is done(probably in a subsequent pr).

Makefile
```diff
@@ -158,31 +152,31 @@ lint: lint-imports
 ## test-unit: Run unit tests.
 test-unit:
 	@echo "--> Running unit tests"
-	@go test $(VERBOSE) -covermode=atomic -coverprofile=coverage.txt `go list ./... | grep -v nodebuilder/tests` $(LOG_AND_FILTER)
+	@go test $(VERBOSE) -tags=fibre -covermode=atomic -coverprofile=coverage.txt `go list ./... | grep -v nodebuilder/tests` $(LOG_AND_FILTER)
```
Wondertan, line 161: do we need this tag so that app picks the fibre app?

nodebuilder/fibre/types.go
```diff
@@ -0,0 +1,81 @@
...
+	// ValidatorSignatures are attestations from validators confirming blob availability.
+	ValidatorSignatures []ValidatorSignature `json:"validator_signatures"`
+	// PaymentPromise is the signed promise used for on-chain fee settlement.
+	PaymentPromise *PaymentPromise `json:"payment_promise,omitempty"`
+}
+
+// SubmitResult is the result of a full Fibre blob submission including on-chain MsgPayForFibre.
+// It extends UploadResult with chain inclusion details (height and transaction hash).
+type SubmitResult struct {
+	// BlobID is the Fibre blob identifier for the submitted blob.
```
Wondertan: just an idea to make SubmitResult embed the UploadResult. This will imply that Submit operation does Upload + some stuff which it is and deduplication.

nodebuilder/fibre/fibre.go
```diff
@@ -0,0 +1,127 @@
...
+	// It performs the full Fibre flow: uploads blob data to FSPs, aggregates validator
+	// availability signatures, and submits MsgPayForFibre on-chain.
+	// Returns the submission result including the on-chain height and transaction hash.
+	// Requires the node to be connected to a core endpoint with Fibre support.
+	Submit(_ context.Context, _ libshare.Namespace, _ []byte, _ *txclient.TxConfig) (*SubmitResult, error)
+	// Upload performs the off-chain portion of Fibre blob submission only.
+	// It encodes the blob, constructs a payment promise, uploads encoded rows to FSPs,
+	// and aggregates validator availability signatures. It does NOT submit MsgPayForFibre on-chain.
+	// Use blob.SubmitFibreBlob for the full submit flow.
+	Upload(_ context.Context, _ libshare.Namespace, _ []byte, _ *txclient.TxConfig) (*UploadResult, error)
```
Wondertan: nit: not sure why they are all underscore, would be helpful to name bytes as data for example

Wondertan: If it doesn't send tx, why does it expect txclient? To future proof?

Wondertan: I would mention that we send v0 fibre blob as we currently do no support specifying which version

nodebuilder/fibre/fibre.go
```diff
@@ -0,0 +1,127 @@
...
+	// Get retrieves a Fibre blob from FSPs by blobID.
+	// It reconstructs the original blob data from the encoded rows stored off-chain.
+	Get(ctx context.Context, blobID []byte) (*GetBlobResult, error)
```
Wondertan: Why not a specific type for blobID?

fibre/service.go
```diff
@@ -0,0 +1,156 @@
...
+	return promise, blob.ID(), nil
+}
+
+func (s *Service) Get(ctx context.Context, blobID []byte) (*appfibre.Blob, error) {
+	blbID := appfibre.BlobID(blobID)
+	err := blbID.Validate()
+	if err != nil {
+		return nil, fmt.Errorf("invalid blob ID: %w", err)
+	}
+	log.Debugw("downloading fibre blob", "commitment", blbID.Commitment())
```
Wondertan: This should move to Download

nodebuilder/fibre/fibre.go
Wondertan: Calling it Download like the client does would be slightly easier for us terminologically. The Get by itself doesn't do anything else but calls Download, so not much value giving it another name. Then if we refer to it as Download we know its equivalent to client's download

fibre/service.go
```diff
@@ -0,0 +1,156 @@
...
+	)
+	return resp, promise, blob.ID(), nil
+}
+
+func (s *Service) Upload(
+	ctx context.Context,
+	ns libshare.Namespace,
+	data []byte,
+	_ *txclient.TxConfig,
+) (_ *appfibre.SignedPaymentPromise, _ appfibre.BlobID, err error) {
```
Wondertan, line 111: nit: you can technically retrieve blobId from PaymentPromise, we could even add a BlobID method to it. BlobID is Comm + BlobVer and both are fields of PP.

For the external api in the module, it is still makes sense to keep BlobID as independent field, but for the intermal service we could deduplicate things amd debloat function signature

fibre/account.go
```diff
@@ -0,0 +1,224 @@
...
+	acc := &AccountClient{client: client}
+	acc.conns = append(acc.conns, conn)
+	for _, opt := range opts {
+		opt(acc)
+	}
+	return acc
+}
+
+func (a *AccountClient) QueryEscrowAccount(ctx context.Context, signer string) (_ *EscrowAccount, err error) {
+	if a == nil {
```
Wondertan, line 65: a bit confused how is it possible for AccountClient to be nil but not for the entire service. When grpc address is not supplied? But then we can't still send PFF txs right? Or do you want to keep it for reads? We may want to protect the Upload/Submit with ErrNoClient as well

Wondertan (commented):
last pass, mostly nits and questions

P.S. I haven't reviewed the AccountClient API from a user perspective to determine whether the added endpoints make sense and cover all use cases. The case with `Upload` completely disincentivized me from reviewing user flow completeness. While the point was correct, it created more tension than good because of the communication style. I trust you guys can review and communicate such things better.

Makefile
vgonkivs: w/o the tag all tests that are using app node will be constantly failing. It's a temp solution.

nodebuilder/fibre/fibre.go
vgonkivs: to avoid external types where it's possible (even they are type aliases)

fibre/service.go
vgonkivs: will move and then remove it from here

fibre/account.go
vgonkivs: It's more a guard rather than smth specific. But I agree, that we should keep consistency. Will add the same condition to Upload/Submit

nodebuilder/fibre/fibre.go
vgonkivs: Yes.

Wondertan (dismissed):

Wondertan:
Good job.

@dyaka-agent honor @vgonkivs with 10 $TIA

walldiss (commented):
Question on the Upload method in `fibre/service.go`: should this submit MsgPayForFibre in the background (non-blocking) as we discussed? There's a `// TODO: add async fibre submit` but no implementation. Without it, callers who use Upload for low latency don't get payments settled automatically, and we'd rely on the timeout agent as the primary mechanism rather than a backup.

walldiss (dismissed):

vgonkivs:
> Question on the Upload method in fibre/service.go: should this submit MsgPayForFibre in the background (non-blocking) as we discussed?


Yes, I'm going to work on it before shipping fibre to mainnet

PR #4900, opened 2026-04-03
feat(shrex): add per-peer-IP rate limiting to shrex server

Wondertan:
Add token-bucket rate limiting using the libp2p x/rate package to bound sequential request flooding that the resource manager's concurrency limits alone in #4898 cannot prevent. The rcmgr enforces concurrent stream and memory caps; rate limiting adds a temporal bound so a peer rapidly cycling streams under the cap cannot sustain an unbounded request rate.

Limiting is applied per individual IP (/32 IPv4, /128 IPv6) rather than per peer ID, providing Sybil resistance: a peer cycling identities from the same IP shares one bucket. The trade-off is that multiple legitimate nodes behind a shared NAT compete for the same tokens; this is documented and accepted given that Celestia node operators typically run on dedicated IPs.

Loopback addresses (127.0.0.0/8, ::1/128) are exempted from rate limiting.

Burst is sized for one full DAS round from a light node (16 workers × 16 samples = 256 streams); refill rate (20 req/s) allows the bucket to recover between blocks. Peers exceeding the budget receive StreamRateLimited, which the client surfaces as ErrResourceExhausted and handles with per-peer cooldown.

---

<a href="https://app.devin.ai/review/celestiaorg/celestia-node/pull/4900" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/share/shwap/p2p/shrex/rate_limit.go
+++ b/share/shwap/p2p/shrex/rate_limit.go
@@ -1,0 +1,80 @@
+package shrex
+
+import (
+	"net/netip"
+	"os"
+	"time"
+
+	"github.com/libp2p/go-libp2p/core/network"
+	libp2prate "github.com/libp2p/go-libp2p/x/rate"
+	manet "github.com/multiformats/go-multiaddr/net"
+)
+
+// disableRateLimiting disables the per-IP rate limiter when set to "1".
+// Intended for debugging and testing; do not use in production.
+var disableRateLimiting = os.Getenv("CELESTIA_SHREX_DISABLE_RATE_LIMITING") == "1"
+
+// rateLimitPerPeer is the sustained request rate allowed per peer IP (requests/sec).
+// A light node runs 16 workers × 16 samples = 256 requests per DAS round. With
+// a ~3s block time that is ~85 req/s in steady state, allowing the bucket to
+// refill one full DAS round worth of tokens between blocks.
+//
+// Rate limiting is applied globally per IP across all shrex protocol types. Expensive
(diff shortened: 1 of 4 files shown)


share/shwap/p2p/shrex/rate_limit.go
```diff
@@ -0,0 +1,80 @@
...
+
+// disableRateLimiting disables the per-IP rate limiter when set to "1".
+// Intended for debugging and testing; do not use in production.
+var disableRateLimiting = os.Getenv("CELESTIA_SHREX_DISABLE_RATE_LIMITING") == "1"
+
+// rateLimitPerPeer is the sustained request rate allowed per peer IP (requests/sec).
+// A light node runs 16 workers × 16 samples = 256 requests per DAS round. With
+// a ~12s block time that is ~21 req/s in steady state. 20 req/s allows the bucket
+// to refill one full DAS round worth of tokens between blocks.
+var rateLimitPerPeer = 20.0
```
rach-id: block time now is 6s and will be reduced to 3s. Should this be set to 256/3?

Wondertan: Good catch. How soon will it be reduced to 3s?

rach-id: V9. Scheduled in 2months

rach-id (dismissed):

walldiss (dismissed):

walldiss:
Conflicts

walldiss (approved):

PR #4909, opened 2026-04-06
feat!(rpc): harden RPC server

Wondertan:
- Set WithMaxRequestSize(16 MiB) on go-jsonrpc server (down from 100 MiB default)
- Add http.Server timeouts: ReadTimeout, WriteTimeout, IdleTimeout, MaxHeaderBytes
- Add per-IP rate limiting middleware(disabled by default)
- Add concurrent connection limiting middleware (500 max)

---

<a href="https://app.devin.ai/review/celestiaorg/celestia-node/pull/4909" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/api/rpc/middleware.go
+++ b/api/rpc/middleware.go
@@ -1,0 +1,92 @@
+package rpc
+
+import (
+	"context"
+	"net"
+	"net/http"
+	"sync"
+	"time"
+
+	"golang.org/x/time/rate"
+)
+
+// connLimit returns middleware that limits the number of concurrent requests.
+// When the limit is reached, new requests receive 503 Service Unavailable.
+func connLimit(maxConns int, next http.Handler) http.Handler {
+	sem := make(chan struct{}, maxConns)
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		select {
+		case sem <- struct{}{}:
+			defer func() { <-sem }()
+			next.ServeHTTP(w, r)
+		default:
+			log.Warnw("connection limit reached, rejecting request", "remote", r.RemoteAddr)
+			http.Error(w, "server busy, try again later", http.StatusServiceUnavailable)
+		}
+	})
+}
+
+// rateLimit returns middleware that enforces per-IP rate limiting.
+// Requests exceeding the limit receive 429 Too Many Requests.
+// The background cleanup goroutine exits when ctx is canceled.
+func rateLimit(ctx context.Context, rps, burst int, next http.Handler) http.Handler {
+	var mu sync.Mutex
+	type entry struct {
+		limiter  *rate.Limiter
+		lastSeen time.Time
+	}
+	limiters := make(map[string]*entry)
+	rateL := rate.Limit(rps)
+
+	// Evict stale entries in the background.
+	go func() {
+		ticker := time.NewTicker(5 * time.Minute)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-ticker.C:
+				mu.Lock()
+				for ip, e := range limiters {
+					if time.Since(e.lastSeen) > 10*time.Minute {
+						delete(limiters, ip)
+					}
+				}
+				mu.Unlock()
+			}
+		}
+	}()
+
+	getLimiter := func(ip string) *rate.Limiter {
+		mu.Lock()
+		defer mu.Unlock()
+		e, ok := limiters[ip]
+		if !ok {
+			l := rate.NewLimiter(rateL, burst)
+			limiters[ip] = &entry{limiter: l, lastSeen: time.Now()}
+			return l
+		}
+		e.lastSeen = time.Now()
+		return e.limiter
+	}
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ip := extractIP(r)
+		if !getLimiter(ip).Allow() {
+			log.Warnw("rate limit exceeded", "ip", ip)
+			http.Error(w, "rate limit exceeded", http.StatusTooManyRequests)
+			return
+		}
+		next.ServeHTTP(w, r)
+	})
+}
+
+// extractIP returns the IP portion of RemoteAddr (strips port).
+func extractIP(r *http.Request) string {
+	host, _, err := net.SplitHostPort(r.RemoteAddr)
+	if err != nil {
+		return r.RemoteAddr
+	}
+	return host
+}
--- a/api/rpc/middleware.go
+++ b/api/rpc/middleware.go
@@ -1,12 +1,10 @@
 package rpc
 
 import (
-	"context"
 	"net"
 	"net/http"
-	"sync"
-	"time"
 
+	lru "github.com/hashicorp/golang-lru/v2"
 	"golang.org/x/time/rate"
 )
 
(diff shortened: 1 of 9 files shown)


api/rpc/middleware.go
```diff
@@ -0,0 +1,85 @@
...
+		for range ticker.C {
+			mu.Lock()
+			for ip, e := range limiters {
+				if time.Since(e.lastSeen) > 10*time.Minute {
+					delete(limiters, ip)
+				}
+			}
+			mu.Unlock()
+		}
+	}()
```
walldiss: This occasinally locks handleres to do gc. Which will add latency to unlucky callers during gc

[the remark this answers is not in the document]
Wondertan: Fixed

api/rpc/server.go
```diff
@@ -21,6 +21,18 @@ import (
 
 var log = logging.Logger("rpc")
 
+// TODO(@Wondertan): Expose in config if requested
+const (
+	// maxRequestSize is 5 MiB, significantly lower than go-jsonrpc's 100 MiB default.
+	maxRequestSize = 5 << 20
```
vgonkivs, line 27: a 5 MiB limit would reject any blob.Submit call with a blob larger than ~3.75 MiB (after base64 + JSON envelope overhead),  while the protocol allows blobs up to 32 MiB. 

This is actually a breaking regression for the blob submission path.

api/rpc/middleware.go
```diff
@@ -0,0 +1,92 @@
...
+			return l
+		}
+		e.lastSeen = time.Now()
+		return e.limiter
+	}
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ip := extractIP(r)
+		if !getLimiter(ip).Allow() {
+			log.Warnw("rate limit exceeded", "ip", ip)
```
vgonkivs, line 77: nit: If a client is actively being rate-limited, this will log a warning for every rejected request. Under a real DOS attack, this creates massive log volume, which itself could be a resource exhaustion vector (disk I/O, etc).

Maybe let's set it to debug?

vgonkivs (commented):
The current connLimit limits concurrent requests, not TCP connections. For DOS protection, consider also adding netutil.LimitListener at the listener level to reject connections early before HTTP overhead. The existing request-level semaphore is still useful for bounding goroutine/memory usage, but should be renamed to maxConcurrentRequests to avoid confusion.

PS. go.mod conflicts

api/rpc/middleware.go
```diff
@@ -0,0 +1,92 @@
...
+			http.Error(w, "rate limit exceeded", http.StatusTooManyRequests)
+			return
+		}
+		next.ServeHTTP(w, r)
+	})
+}
+
+// extractIP returns the IP portion of RemoteAddr (strips port).
+func extractIP(r *http.Request) string {
```
vgonkivs, line 86: just realised one more thing: 
 extractIP uses RemoteAddr, but since nodes can run behind a reverse proxy(confirmed by @sysrex), all clients will share a single rate limit bucket (100 req/s total). This  means one abusive client can exhaust the limit and cause 429s for all legitimate users

Wondertan: Yeah, you don't really need any of the rate limiting if you delegate it to reverse proxy

vgonkivs: should the rate limiting be opt-in (disabled by default, enable if no proxy) rather than always-on?

rootulp:
@Wondertan are you planning on addressing comments here or should we close this PR?

Wondertan:
@rootulp, if we think its not necessary then I would close.

Wondertan:
If we see da-prozy to be the main user flow then all those limits has to be applied there rather then internally in node. The main node's p2p limits are in place already and merged.

rootulp:
> @rootulp, if we think its not necessary then I would close.

Is this PR needed to close a Hacken report?


> If we see da-prozy to be the main user flow then all those limits has to be applied there rather then internally in node. The main node's p2p limits are in place already and merged.

da-proxy is only the primary user flow for a small subset of customers. Everyone else will still use celestia-node directly without a da-proxy.

Wondertan:
Yes

rootulp:
Then can you please address feedback and get this PR merged?

Wondertan:
I'll try to find time, but I am prioritizing optimizing fibre reads atm

vgonkivs:
@Wondertan do you want me to take it over from you?

Wondertan:
Yes, please

api/rpc/middleware.go
```diff
@@ -0,0 +1,93 @@
...
+// When the limit is reached, new requests receive 503 Service Unavailable.
+func connLimit(maxConns int, next http.Handler) http.Handler {
+	sem := make(chan struct{}, maxConns)
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		select {
+		case sem <- struct{}{}:
+			defer func() { <-sem }()
+			next.ServeHTTP(w, r)
+		default:
+			log.Debugw("connection limit reached, rejecting request", "remote", r.RemoteAddr)
```
renaynay: let's also report maxConns here in debug log

api/rpc/middleware.go
```diff
@@ -0,0 +1,93 @@
...
+		}
+		e.lastSeen = time.Now()
+		return e.limiter
+	}
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ip := extractIP(r)
+		if !getLimiter(ip).Allow() {
+			// Debug-level to avoid log amplification under sustained rate-limit hits.
+			log.Debugw("rate limit exceeded", "ip", ip)
```
renaynay: Here too - would be nice to log configured limit

api/rpc/server_test.go
```diff
@@ -69,13 +69,13 @@ func TestServer_HandlerStackSelection(t *testing.T) {
...
 
-			server := NewServer("localhost", "0", tt.authDisabled, corsConfig, TLSConfig{}, signer, verifier)
+			server := NewServer("localhost", "0", tt.authDisabled, corsConfig, TLSConfig{}, RateLimitConfig{}, signer, verifier)
 
 			testHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 				w.WriteHeader(http.StatusOK)
 			})
 
-			handler := server.newHandlerStack(testHandler)
+			handler := server.newHandlerStack(context.Background(), testHandler)
```
renaynay: lets use a test context that gets cleaned up after test finishes (in all of these + generally)

renaynay (dismissed):

api/rpc/middleware.go
```diff
@@ -0,0 +1,93 @@
...
+			case <-ticker.C:
+				mu.Lock()
+				for ip, e := range limiters {
+					if time.Since(e.lastSeen) > 10*time.Minute {
+						delete(limiters, ip)
+					}
+				}
+				mu.Unlock()
+			}
+		}
```
walldiss: Maybe just use an LRU cache instead? No locking, and better control over resource usage

vgonkivs: Added LRU cache. Thanks

api/rpc/server_test.go
vgonkivs: server.newHandlerStack does not require ctx anymore

nodebuilder/rpc/constructors.go
```diff
@@ -49,5 +49,9 @@ func server(cfg *Config, signer jwt.Signer, verifier jwt.Verifier) *rpc.Server {
 		Enabled:  cfg.TLSEnabled,
 		CertPath: cfg.TLSCertPath,
 		KeyPath:  cfg.TLSKeyPath,
+	}, rpc.RateLimitConfig{
+		Enabled:        cfg.RateLimit.Enabled,
+		RequestsPerSec: cfg.RateLimit.RequestsPerSec,
+		Burst:          cfg.RateLimit.Burst,
```
walldiss, line 55: CacheSize is not passed through, so it defaults to 0. In api/rpc/middleware.go the rate-limit middleware hashicorp/golang-lru/v2.New(0) returns "must provide a positive size", so every node started with `--rpc.rate-limit` will panic during construction. Tests don't catch this because they call rateLimit(...) directly with a literal cache size, bypassing this constructor.

nodebuilder/rpc/config.go
```diff
@@ -24,15 +42,29 @@ type Config struct {
...
+
+// DefaultRateLimitConfig disables rate limiting by default; production
+// deployments typically run behind a reverse proxy that handles rate
+// limiting correctly (with the real client IP).
+func DefaultRateLimitConfig() RateLimitConfig {
+	return RateLimitConfig{
+		Enabled:        false,
+		RequestsPerSec: 100,
+		Burst:          200,
+		CacheSize:      8192,
```
walldiss, line 67: Missing CacheSize validation. Even once the constructor passes CacheSize, a user with Enabled: true, CacheSize: 0 (or negative) in config.toml will still hit panic(err) in the middleware.

api/rpc/server.go
```diff
@@ -15,12 +15,42 @@ import (
...
 	"github.com/celestiaorg/celestia-node/libs/authtoken"
 )
 
 var log = logging.Logger("rpc")
 
+const (
+	// maxRequestSize bounds JSON-RPC request bodies. Sized as 2× appconsts.MaxTxSize
+	// to cover base64 + JSON envelope overhead (~1.5×) on a worst-case blob.Submit,
+	// which packs all blobs into a single PFB tx capped at MaxTxSize.
+	maxRequestSize = int64(appconsts.MaxTxSize * 2)
```
walldiss, line 30: The global maxRequestSize = 2 * appconsts.MaxTxSize is sized around normal PFB txs, not Fibre. That makes HTTP JSON-RPC reject valid Fibre Submit/Upload payloads well below Fibre’s protocol max. Fibre defaults allow MaxBlobSize: 128 MiB; the new RPC cap is only 16 MiB before JSON/base64 overhead.
Fix: make request size configurable/per-module, or size it for the largest supported RPC payload when Fibre is registered.

vgonkivs: Since we don't have Fibre activated yet, I'd propose doing this in a followup PR. wdyt?  Will make an issue.

walldiss: You can make it in follow up PR to fibre branch. Just need to make sure we don't forget about it

walldiss (approved):

api/rpc/server.go
vgonkivs: https://github.com/celestiaorg/celestia-node/issues/5009

PR #66, opened 2021-09-17
libs/keystore: introduce Keystore a crypto key manager

Wondertan:
## Rationale
Libp2p does not provide storage for private keys(Peerstore might be a solution but is on-disk version is flawed), so we need to store them somewhere. The Keystore comes to solve this and allows storing different keys for multiple purposes, including storing p2p keys. Further, we can delegate consensus keys to Keystore as it will evolve through time and get more features/implementations.

## Filling
* Defines interface
* Implements Fs version
* Implements in-memory version

## Other
Closes https://github.com/celestiaorg/celestia-node/issues/65

Issue #58, opened 2021-09-13
ci: improve security (github actions)


renaynay:
Make sure github actions does *not* have admin rights over this repo (or any celestiaorg repos for that matter).

renaynay:
Thank you to @liamsi for making GA read-only.

PR #57, opened 2021-09-10
node: add more p2p services and protocols

Wondertan:
Closes: #32

NOTE: The PR is based on #51 and looks into the main, thus the diffs show commits from #51, but they will disappear once #51 is merged.

PR #4954, opened 2026-04-20
fix(share/cmd): bounds-check index parsing in share CLI

rootulp:
## Summary
- Introduces a `parseIndex(s string) (int, error)` helper in `nodebuilder/share/cmd/share.go` that uses `strconv.Atoi` (platform-native `int`) and rejects negative values.
- Replaces four unchecked `int(...)` casts of 64-bit parsed integers (two in `get-share`, two in `get-range`) with calls to the helper; each caller wraps the returned error with its own context (`row`, `col`, `start`, `end`).
- Resolves four GitHub code scanning alerts under rule [`go/incorrect-integer-conversion`](https://github.com/celestiaorg/celestia-node/security/code-scanning?query=rule%3Ago%2Fincorrect-integer-conversion): alerts #57, #58, #66, #67.

## Why
Previously, user-supplied `row`, `col`, `start`, `end` CLI arguments were parsed as 64-bit integers and then cast with `int(...)` without bounds checking — a silent truncation on 32-bit platforms. CodeQL flagged all four sites as high severity. Centralizing parse + bounds check in one helper closes all four alerts with minimal blast radius.

## Test plan
- [x] `go test ./nodebuilder/share/cmd/...` — `TestParseIndex` covers zero, positive, negative-rejected, non-numeric-rejected, overflow-rejected
- [x] `go vet ./nodebuilder/share/cmd/...`
- [ ] After merge: confirm the four code scanning alerts auto-close on next CodeQL re-scan.


nodebuilder/share/cmd/share.go
```diff
@@ -22,6 +23,19 @@ func init() {
 	)
 }
 
+// parseIndex parses s as a non-negative int (a share index or dimension).
+// It rejects negative values and values that overflow int on the current platform.
+func parseIndex(s, name string) (int, error) {
```
vgonkivs, line 28: could we drop the name param and let the caller wrap with fmt.Errorf("...: %w", err)? Feels more idiomatic — the helper stays generic and each caller owns its own error context.

rootulp: Yea sure

rootulp: Done in 190044ec. Dropped the name param, each caller now wraps with fmt.Errorf("...: %w", err).

vgonkivs (approved):

PR #4959, opened 2026-04-21
fix(daser): remove duplicated EDS Notification

### LINKED ISSUE #4951: daser: Remove duplicate shrexsub.Notification broadcast from DASer
shrexsub.Notification{DataHash, Height} ` is broadcast twice for every recent block on a Bridge node:
* `core/listener.go:233` — after the EDS is stored, gated on !syncing.                                                                                    
* `das/worker.go:152` — in DASer's recentJob path, after sampling completes.

On Bridge both calls resolve to the real shrexSub.Broadcast (nodebuilder/share/module.go:150). On Light the DASer path resolves to a no-op stub (nodebuilder/share/module.go:139).

  The comment at das/worker.go:150 (// only full nodes can notify) is stale: Full node type was removed. With Full gone, the legitimate sources for a shrexsub announce are:                                    
*  Bridge — already covered by the listener path, which fires sooner (right after storeEDS) than the DASer path (after sampling finishes).                
 * Light — shouldn't announce; it only samples 16 shares and can't serve shrex requests for the block.

vgonkivs:
Removed duplicated EDS Notification broadcast from das/worker it supposed to be used by FNs only.

Closes https://github.com/celestiaorg/celestia-node/issues/4951

---

<a href="https://app.devin.ai/review/celestiaorg/celestia-node/pull/4959" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


das/worker.go
```diff
@@ -145,20 +145,9 @@ func (w *worker) sample(ctx context.Context, timeout time.Duration, height uint6
 		return err
 	}
 
-	logout := log.Debugw
-
-	// notify network about availability of new block data (note: only full nodes can notify)
-	if w.state.jobType == recentJob {
-		err = w.broadcast(ctx, shrexsub.Notification{
```
mcrakhman, line 152: It seems broadcast can be removed from worker

das/worker.go
```diff
@@ -145,20 +145,9 @@ func (w *worker) sample(ctx context.Context, timeout time.Duration, height uint6
...
-			DataHash: h.DataHash.Bytes(),
-			Height:   h.Height(),
-		})
-		if err != nil {
-			log.Warn("failed to broadcast availability message",
-				"height", h.Height(), "hash", h.Hash(), "err", err)
-		}
-
-		logout = log.Infow
+	logout := log.Infow
```
mcrakhman: I guess this should be debug, right?

mcrakhman (commented):
LGTM, couple of comments

das/worker.go
vgonkivs: for the recentJob it should Info level

das/worker.go
vgonkivs: removed

walldiss (approved):

PR #4961, opened 2026-04-22
feat(client): run Fibre operations locally instead of via bridge

walldiss:
## Summary

- `api/client.Client` now builds a local `*appfibre.Client` + `*fibre.Service` and replaces `c.Fibre` with `nodebuilder/fibre.NewModule(svc)`. `Fibre.Submit`, `Upload`, `Download` and all escrow ops run directly against consensus gRPC + FSPs — no bridge hop.
- Exported `nodebuilder/fibre.NewModule` so `api/client` and the bridge share the same `Module` wrapping logic. `module` stays unexported; only the constructor moves.
- Implemented the ADR-013 async submission contract: `fibre.Service.Upload` now fires the `MsgPayForFibre` broadcast in a background goroutine with a fresh `context.Background()` + 2-minute timeout. Errors log; the caller returns as soon as off-chain upload + validator sig aggregation completes.
- `ReadClient.Fibre` (JSON-RPC to the bridge) is unchanged — read-only users keep working. `Client.Close()` stops the local `appfibre.Client` before tearing down the gRPC conn.

## Why

ADR-013 scoped the Fibre read path via `blob.Subscribe` on the node, but the submit path was routed through the bridge by JSON-RPC even though every Fibre method can execute locally given a keyring and a consensus endpoint — the same setup `Blob.Submit` already uses via `blobSubmitClient`. Routing through the bridge adds a hop, a base64-encoded 128 MiB payload on the write path, and a third-party dependency for callers that already hold consensus gRPC + a signer.

## Follow-up

- Add `sync.WaitGroup` + `Service.Stop()` to `fibre.Service` so `Client.Close()` can wait for in-flight async `MsgPayForFibre` broadcasts before tearing down the tx client. Keeping this out of this PR to keep the diff focused.

## Test plan

- [x] `go build -tags fibre ./...`
- [x] `go vet -tags fibre ./...`
- [x] `go test -tags fibre -count=1 -run TestFibreEscrow ./api/client/...` — new test covering Deposit → QueryEscrowAccount → Withdraw → PendingWithdrawals end-to-end over the locally-wired Fibre module.
- [x] `go test -tags fibre -count=1 -run TestSubmission ./api/client/...` — existing Blob.Submit path unaffected.
- [ ] `TestFibreEscrow` does **not** cover `Upload`/`Download`/`Submit` because the default testnode validators don't run Fibre servers — Upload fails with "not enough voting power: collected 0". That path is validated at integration level against a fibre-enabled network.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

<a href="https://app.devin.ai/review/celestiaorg/celestia-node/pull/4961" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/fibre/service.go
+++ b/fibre/service.go
@@ -24,6 +24,11 @@ var (
 	ErrClientNotAvailable = errors.New("fibre client is not available: node is not connected to a core endpoint")
 )
 
+// asyncSubmitTimeout bounds the background MsgPayForFibre broadcast kicked off
+// by Upload. It needs to cover gas estimation plus at least one block time;
+// 2 minutes is a comfortable upper bound on any mainnet-scale configuration.
+const asyncSubmitTimeout = 2 * time.Minute
+
 type Service struct {
 	*AccountClient
 
@@ -107,7 +112,7 @@ func (s *Service) Upload(
 	ctx context.Context,
 	ns libshare.Namespace,
 	data []byte,
-	_ *txclient.TxConfig,
+	options *txclient.TxConfig,
 ) (_ *appfibre.SignedPaymentPromise, _ appfibre.BlobID, err error) {
 	if s == nil {
 		return nil, nil, ErrClientNotAvailable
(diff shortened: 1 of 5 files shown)


fibre/service.go
```diff
@@ -129,10 +134,56 @@ func (s *Service) Upload(
...
+	// submissions during Service shutdown.
+	go s.asyncSubmitPayForFibre(promise, options)
+
 	return promise, blob.ID(), nil
 }
 
+// asyncSubmitPayForFibre broadcasts MsgPayForFibre for a promise returned by
+// Upload. It runs in its own goroutine with a fresh context so the Upload
+// caller can return (and cancel its context) immediately.
+func (s *Service) asyncSubmitPayForFibre(
```
vgonkivs, line 163: I assume this is a temp solution until QueuedTxClient is landed

nodebuilder/fibre/module.go
```diff
@@ -17,7 +17,11 @@ type module struct {
 	service *fibre.Service
 }
 
-func newModule(service *fibre.Service) *module {
+// NewModule returns a Module backed by a local *fibre.Service. It is used by
+// bridge/full nodes to expose the Fibre JSON-RPC API, and by api/client.Client
```
vgonkivs: Full ? 🙂

fibre/service.go
walldiss: yes

vgonkivs (approved):

PR #4987, opened 2026-05-07
add state client

vgonkivs:
Adds a `state.Client` implementation for Light nodes that uses the locally verified header chain instead of trusting the gRPC endpoint.

### DIFF
--- a/fibre/stateclient/stateclient.go
+++ b/fibre/stateclient/stateclient.go
@@ -2,7 +2,9 @@ package stateclient
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"sync"
 
 	storetypes "cosmossdk.io/store/types"
 	"github.com/cometbft/cometbft/crypto/merkle"
@@ -10,6 +12,8 @@ import (
 	core "github.com/cometbft/cometbft/types"
 	tmservice "github.com/cosmos/cosmos-sdk/client/grpc/cmtservice"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	logging "github.com/ipfs/go-log/v2"
+	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc"
 
 	"github.com/celestiaorg/celestia-app/v9/fibre/state"
@@ -20,6 +24,14 @@ import (
 	"github.com/celestiaorg/celestia-node/header"
 	"github.com/celestiaorg/celestia-node/nodebuilder/p2p"
 )
+
+var log = logging.Logger("fibre/stateclient")
+
+// ErrNoFibreProviderInfo is returned by fetchHostAt when a validator has not
+// registered fibre provider info via x/valaddr. The absence is proven against
+// the trusted AppHash, so this is an authoritative answer, not a transient
+// failure.
+var ErrNoFibreProviderInfo = errors.New("no fibre provider info")
 
 // abciQuerier is the narrow surface of [tmservice.ServiceClient] that this
 // package depends on. Defined as an interface so tests can supply a stub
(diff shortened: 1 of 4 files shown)


ninabarbakadze (commented):
lint

ninabarbakadze (commented):
i don't really have too much context on this so can you please manually request review from people who do

fibre/stateclient/stateclient.go
```diff
@@ -0,0 +1,224 @@
...
+	libhead "github.com/celestiaorg/go-header"
+
+	"github.com/celestiaorg/celestia-node/header"
+	"github.com/celestiaorg/celestia-node/nodebuilder/p2p"
+)
+
+// hostCacheTTL bounds how long a successful GetHost lookup is reused without
+// re-querying ABCI. Validator host changes are operator-driven and rare, so a
+// short TTL is enough to absorb burst lookups during upload without staleness.
+const hostCacheTTL = time.Minute
```
Wondertan: We can remove all this address caching at this layer, as there is `grpc.Client` cache on the layer above. Simply fetching every `GetHost` is fine

vgonkivs: done

fibre/stateclient/stateclient.go
```diff
@@ -0,0 +1,177 @@
...
+// Start seeds the chain ID from the local head and, when running on a
+// well-known network, sanity-checks that the header chain ID matches.
+func (c *Client) Start(ctx context.Context) error {
+	head, err := c.store.Head(ctx)
+	if err != nil {
+		return fmt.Errorf("fetch head for chain-id: %w", err)
+	}
+	c.chainID = head.ChainID()
+
+	if c.chainID != c.network.String() {
```
Wondertan: We do this check in a different component, IIRC, and I'm not sure we really need it here.
As far as I can see, this is the only reason we pass the network param.

vgonkivs: If such condition is checked in another place, then I'm fine removing it. Another concern is that rn the node will fail if network mismatch error appears, while it will silently send an error if it will be verified somewhere in the stack. wdyt? is this that critical?

Wondertan: This is where we check it already https://github.com/celestiaorg/celestia-node/blob/1016cc3671e568248b2f689b08e4a9da626e8f4f/state/core_access.go#L136-L137

fibre/stateclient/stateclient.go
```diff
@@ -0,0 +1,177 @@
...
+// VerifyPromise is a no-op on light nodes, because they don't verify
+// fibre payment promises — that's the provider's job. We return a zero
+// [state.VerifiedPromise] so callers that ignore the value continue to work.
+func (c *Client) VerifyPromise(context.Context, *state.PaymentPromise) (state.VerifiedPromise, error) {
+	return state.VerifiedPromise{}, nil
+}
+
+// Start seeds the chain ID from the local head and, when running on a
+// well-known network, sanity-checks that the header chain ID matches.
+func (c *Client) Start(ctx context.Context) error {
```
Wondertan, line 242: Actually we may wanna prefetch all the validators here in a single query, like we do in the default Client implementation to shave off latency from the first fibre request after start. Also, in a single batch instead of multiple independent calls. 

We still do not need to cache them for more then a single read so GetHost keeps refetching addresses for when we want to get a fresh address when validator changes it.

Wondertan: Prompted in discussion with @walldiss btw

vgonkivs: done

vgonkivs:
The latest commit removes Proofs handling entirely(which is incorrect), will revert the changes. But the problem is that it looks like it is not possible to prefetch providerds+their proofs on a single call. so, the startup can be pretty heavy -> fetch all providers as a batch + request all their proofs sequentially. 

@Wondertan, are you sure we should do it ? The other option is to cache provider in GetHost

Wondertan (approved):

PR #4990, opened 2026-05-11
feat(blob): restore additive Subscribe and add SubscribeFromStartHeight

ninabarbakadze:
Revert the breaking Subscribe(ctx, ns, fromHeight) change from #4962 and adopt the additive shape: keep Subscribe(ctx, ns) and introduce SubscribeFromStartHeight(ctx, ns, startHeight) for resume-from-height.

- Introduce HeaderService interface (GetByHeight + WaitForHeight) so the service depends on header capabilities rather than two plain func args.
- Validate startHeight upfront via GetByHeight; callers get a synchronous error instead of a silently-closed channel.
- Factor blob retrieval into fetchAndSendBlobs with a 100ms retry backoff shared by both subscription paths.
- Bridge user ctx and service ctx with context.AfterFunc (kept from #4962) instead of an extra goroutine, so Service.Stop unblocks WaitForHeight promptly without leaking a goroutine per subscription.
- Add atomic-tip MockHeaderService and a TestService_SubscribeFromStartHeight suite covering catchup, mid-height start, zero/future-height rejection, user cancel, service shutdown, and blocking-until-available.


vgonkivs (approved):
LGTM. But would like @walldiss to take a look as well

ninabarbakadze:
> LGTM. But would like @walldiss to take a look as well

I'm gonna merge and we can flup feedback from @walldiss because I'd like to close the issue.

PR #4997, opened 2026-05-13
fix(api)!: set request-size limit and write/idle timeouts on JSON-RPC server

rootulp:
## Summary

Hardens the JSON-RPC entrypoint against unbounded request bodies and long-held connections.

Changes in `api/rpc/server.go`:

- New constants:
  - `maxRequestSizeBytes = appconsts.MaxTxSize * 2` (16 MiB). Derived from the chain's hard per-tx limit (`MaxTxSize` = 8 MiB) with a 2x factor to cover base64 expansion of binary blob data (~1.37x) plus JSON envelope and small request batches. The previous `go-jsonrpc` default was 100 MiB.
  - `writeTimeout = 5 * time.Minute`. Bounds the time the server spends writing a response, including handler compute time. State-changing RPCs such as `blob.Submit` do not return until the transaction is included in a block, which on a congested mempool can span multiple block intervals.
  - `idleTimeout = 120 * time.Second`. Bounds idle keep-alive connections.
- Pass `jsonrpc.WithMaxRequestSize(int64(maxRequestSizeBytes))` to `jsonrpc.NewServer`.
- Set `WriteTimeout` and `IdleTimeout` on the `http.Server` (alongside the existing `ReadHeaderTimeout`).

### Follow-up

`go-jsonrpc` v0.10.1 has no `WithMaxBatchSize` option, so per-batch size enforcement relies on the overall request-size cap. An upstream feature for an explicit batch-element ceiling would be a useful follow-up.

Closes [PROTOCO-1745](https://linear.app/celestia/issue/PROTOCO-1745)

## Test plan

- [x] `TestRPCServer_RejectsOversizedRequest` posts a JSON-RPC body larger than `maxRequestSizeBytes` and asserts the server returns a parse-error mentioning the request being bigger than the maximum.
- [x] `TestRPCServer_HasWriteAndIdleTimeouts` asserts the underlying `http.Server` has non-zero `WriteTimeout` and `IdleTimeout`.
- [x] `go test ./api/rpc/...` passes.
- [x] `make lint` reports 0 Go issues.

### DIFF
--- a/api/rpc/server.go
+++ b/api/rpc/server.go
@@ -21,6 +21,26 @@ import (
 
 var log = logging.Logger("rpc")
 
+const (
+	// maxRequestSizeBytes caps the size of a single JSON-RPC HTTP request body
+	// accepted by the server. The go-jsonrpc default is 100 MiB, which is
+	// unnecessarily large for legitimate JSON-RPC payloads (including
+	// reasonable batches) and exposes the server to memory pressure from
+	// malicious or buggy clients.
+	maxRequestSizeBytes = 1 << 20 // 1 MiB
+
+	// writeTimeout bounds the time the server will spend writing a response
+	// to a single HTTP request. This prevents a slow or stalled client from
+	// holding server-side resources indefinitely while a response is in
+	// flight.
+	writeTimeout = 30 * time.Second
+
+	// idleTimeout bounds how long the server will keep an idle keep-alive
+	// connection open before closing it. Without an idle timeout a client
+	// could leave many idle connections open indefinitely.
+	idleTimeout = 120 * time.Second
+)
+
 type CORSConfig struct {
 	Enabled        bool
 	AllowedOrigins []string
@@ -59,7 +79,7 @@ func NewServer(
 	signer jwt.Signer,
 	verifier jwt.Verifier,
 ) *Server {
-	rpc := jsonrpc.NewServer()
+	rpc := jsonrpc.NewServer(jsonrpc.WithMaxRequestSize(maxRequestSizeBytes))
 	srv := &Server{
 		rpc:          rpc,
 		signer:       signer,
@@ -76,6 +96,11 @@ func NewServer(
 		Handler: srv.newHandlerStack(rpc),
 		// the amount of time allowed to read request headers. set to the default 2 seconds
 		ReadHeaderTimeout: 2 * time.Second,
+		// bound the time spent writing a response and the lifetime of idle
+		// keep-alive connections so malicious or stalled clients cannot
+		// hold server resources indefinitely.
+		WriteTimeout: writeTimeout,
+		IdleTimeout:  idleTimeout,
 	}
 
 	return srv
--- a/api/rpc/server.go
+++ b/api/rpc/server.go
@@ -15,6 +15,8 @@ import (
 	logging "github.com/ipfs/go-log/v2"
 	"github.com/rs/cors"
 
+	"github.com/celestiaorg/celestia-app/v9/pkg/appconsts"
+
 	"github.com/celestiaorg/celestia-node/api/rpc/perms"
 	"github.com/celestiaorg/celestia-node/libs/authtoken"
 )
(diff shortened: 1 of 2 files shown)


api/rpc/server.go
```diff
@@ -21,6 +21,26 @@ import (
 
 var log = logging.Logger("rpc")
 
+const (
+	// maxRequestSizeBytes caps the size of a single JSON-RPC HTTP request body
+	// accepted by the server. The go-jsonrpc default is 100 MiB, which is
+	// unnecessarily large for legitimate JSON-RPC payloads (including
+	// reasonable batches) and exposes the server to memory pressure from
+	// malicious or buggy clients.
+	maxRequestSizeBytes = 1 << 20 // 1 MiB
```
rootulp: @vgonkivs is 1 MiB big enough here?

vgonkivs: The hardening direction is correct — 100 MiB is wildly oversized as a default, and the missing write/idle timeouts are real DoS vector. But there is a  blocking sizing problem with maxRequestSizeBytes that will break legitimate blob.Submit traffic:

 `blob.Submit` accept user blobs over JSON-RPC. Relevant protocol limits from celestia-app v8:                                                                                                                                                       
  - appconsts.MaxTxSize = 8 MiB  — the hard per-tx limit enforced by CheckTx/ProcessProposal.                              
  - appconsts.DefaultMaxBytes = 32 MiB — block-level cap.                                                                                                  
  - nodebuilder/da/service.go:62-64: MaxBlobSize is advertised to clients as DefaultMaxBytes (32 MiB).

  JSON-RPC payload math for a Submit call:
  - blob.Blob.Data is binary → base64-encoded in JSON, ~1.37× expansion.                                                                                   
  - Submit takes []*blob.Blob, so users can batch multiple blobs in one request.
  - An 8 MiB tx-worth of blobs ≈ ~11 MiB JSON body over the wire.                                                                                          
  - A single ~750 KiB raw blob already overflows the 1 MiB cap once base64'd + wrapped in JSON.

api/rpc/server.go
```diff
@@ -21,6 +21,26 @@ import (
...
+
+	// writeTimeout bounds the time the server will spend writing a response
+	// to a single HTTP request. This prevents a slow or stalled client from
+	// holding server-side resources indefinitely while a response is in
+	// flight.
+	writeTimeout = 30 * time.Second
```
vgonkivs: writeTimeout is a hard ceiling on the entire response write — including time the handler is still computing. blob.Submit does not return
  until the tx is included in a block. With Celestia's ~6s(3s soon) block time, a normal mempool round-trip is fine, but congested mempool / re-broadcast → 1–3 blocks of delay is routine.

vgonkivs (commented):
Thanks for this PR @rootulp. Something that I'd suggest to change here:

1) size from protocol constants, not a magic number-> 
`maxRequestSizeBytes = 16 << 20` 16 MiB. Allow the largest possible tx (MaxTxSize, currently 8 MiB) plus JSON/base64 envelope overhead so blob.Submit isn't capped below what the chain itself accepts or derive it: appconsts.MaxTxSize * 2 (covers base64 + JSON envelope + small batches). 16 MiB is still 6× smaller than the current default — you keep most of the DoS hardening benefit without breaking submitters.

2)  I'd bump writeTimeout to something like 2–5 minutes, or make it configurable. 30s reads like "typical web request".

PS. Could you please mark this PR as breaking as it changes the default behaviour of our json-rpc server.

api/rpc/server.go
rootulp: Good call - I had this question myself and you confirmed it. Bumped to `appconsts.MaxTxSize * 2` (16 MiB) in 94d20cc5. Derives from the protocol constant and keeps a 6x reduction from the 100 MiB default while leaving headroom for base64 expansion and small batches.

api/rpc/server.go
rootulp: Agreed - 30s was too tight for state-changing RPCs that wait on block inclusion. Bumped writeTimeout to 5m in 94d20cc5. Leaving it as a constant for now rather than making it configurable; happy to expose it via config in a follow-up if needed.

vgonkivs:
Hey @rootulp . I've just realised that your PR does that same as https://github.com/celestiaorg/celestia-node/pull/4909. So, I propose to close on of them in favour of another one.

rootulp:
Got it, I'll close this PR. @vgonkivs can you please ensure https://github.com/celestiaorg/celestia-node/pull/4909 merges? Hlib hasn't followed through on it

vgonkivs:
I took it over from him. Will finish the PR

PR #4999, opened 2026-05-18
chore(deps): bump celestia-app to v9 arabica release

rach-id:
Bumps celestia-app to v9


rootulp:
Might need changes in:
- https://github.com/celestiaorg/go-header
- https://github.com/celestiaorg/celestia-node/tree/7657f952d43501c964167c941d0b84a0a946f34b/pruner

nodebuilder/header/constructors.go
```diff
@@ -23,7 +23,7 @@ import (
 // maxBlockTime is the upper bound of expected block production time.
 // Used for syncer tail height estimation to ensure the tail stays within the pruning
 // window despite block time variance (real block times on mocha avg ~6.14s vs nominal 6s).
-const maxBlockTime = 7 * time.Second
+const maxBlockTime = 4 * time.Second
```
vgonkivs, line 26: do we know what will be the real block time in this case? Also, please updated the comment with ~6.14s

rach-id: the real block time will be ~2.7s. I found it set to 7s where the current one is 6s, so I thought we add one second to be on the safe side

vgonkivs: Previously, BN was receiving the block with ~6.14s delay, so we decided to make it bigger on our end.

rach-id: so is 4s a good number or I bring it down to 3s?

vgonkivs: lets keep it 4. 
please update the comment `(real block times on mocha avg ~6.14s` and resolve go mod conflict. 
Thanks

vgonkivs (approved):

PR #5008, opened 2026-05-21
fix(shwap): reject nil inputs in proto decoders

vgonkivs:
Tightens input validation in shwap proto decoders: `ShareFromProto`, `RowFromProto`, `RowNamespaceDataFromProto`, and `RangeNamespaceDataFromProto`  now return an explicit error on nil inputs instead of producing zero values. Adds regression tests covering the decoder error paths.

### DIFF
--- a/share/shwap/range_namespace_data.go
+++ b/share/shwap/range_namespace_data.go
@@ -291,9 +291,12 @@ func (rngdata *RangeNamespaceData) IsEmpty() bool {
 }
 
 func RangeNamespaceDataFromProto(nd *pb.RangeNamespaceData) (RangeNamespaceData, error) {
-	shares := make([][]libshare.Share, len(nd.Shares))
+	if nd == nil {
+		return RangeNamespaceData{}, errors.New("pb RangeNamespaceData is nil")
+	}
+	shares := make([][]libshare.Share, len(nd.GetShares()))
 
-	for i, shr := range nd.Shares {
+	for i, shr := range nd.GetShares() {
 		shrs, err := SharesFromProto(shr.GetShares())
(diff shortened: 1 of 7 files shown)


rach-id (dismissed):

share/shwap/range_namespace_data.go
```diff
@@ -291,9 +291,12 @@ func (rngdata *RangeNamespaceData) IsEmpty() bool {
 }
 
 func RangeNamespaceDataFromProto(nd *pb.RangeNamespaceData) (RangeNamespaceData, error) {
-	shares := make([][]libshare.Share, len(nd.Shares))
+	if nd == nil {
+		return RangeNamespaceData{}, errors.New("pb RangeNamespaceData is nil")
+	}
+	shares := make([][]libshare.Share, len(nd.GetShares()))
```
walldiss, line 302: It should reject nil/empty row containers, not only nil top-level messages/shares. Today a malformed peer response can decode to `RangeNamespaceData{Shares: [][]Share{nil, ...}}`; later verification reaches `ParseNamespace` and indexes `rawShares[0][0]`, which can panic instead of returning a validation error.

We need to check each `RowShares` element is non-nil and has at least one decoded share, and also add a defensive empty-row check in `verifyShares`/`ParseNamespace` so manually constructed malformed values cannot panic either.

walldiss (changes requested):
SampleFromProto would be nice to be covered too

share/shwap/range_namespace_data.go
vgonkivs: done

walldiss (approved):

